PR #146, opened 2026-05-15
Add destructuring and pattern matching

egonSchiele:
## Summary

Adds a unified pattern language to Agency supporting destructuring and structural pattern matching across `let` / `const` declarations, `match` blocks, `if` / `while` conditions (via the new `is` operator), and `for` loops.

Built as a **pre-typecheck lowering pass**, so the rest of the pipeline (typechecker, TypeScriptBuilder, preprocessor, LSP) requires zero changes — it sees only existing AST nodes.

## What you can write

```ts
// Array & object destructuring
let [a, b, ...rest]       = items
let { name, age }         = person
let { name, ...others }   = person
let { coords: [x, y] }    = nested

// for-loop destructuring
for ([k, v] in entries) { print(k, v) }
for ({ name, age } in users) { print(name, age) }

// is operator: boolean, or with bindings inside if/while
let isShow = step is { type: "showPolicy" }
if (step is { type: "showPolicy", policy }) {
    print(policy.name)
}

// Pattern arms in match blocks (with optional guards)
match (event) {
    { type: "click",  x, y } if (x > 0) => handleClick(x, y)
    { type: "scroll", delta }            => handleScroll(delta)
    _                                    => ignore()
}

// Destructure-once + guard arms
match (req is { user, role }) {
    role == "admin"  => grantAll(user)
    role == "editor" => grantEditing(user)
    _                => grantReadOnly(user)
}
```

## How it works

```
parse → lowerPatterns → SymbolTable → typechecker → preprocessor → builder → printTs
                                          ▲
                              format path skips lowerPatterns
                              so AgencyGenerator sees patterns
```

- Two pattern parsers by context: `bindingPatternParser` (binders only) for `let` / `const` and `for`, `matchPatternParser` (binders + literals) for `is` and match arms.
- The `is` operator is wrapped at atom level rather than added to the precedence table, since its right operand is a pattern, not an expression. `IsExpression` is its own AST node.
- Lowering recursively walks expressions: any `isExpression` outside `if` / `while` / `match` head becomes a boolean condition; binders in pure-boolean position are a compile error.
- Object rest is emitted as a synthetic `__objectRest` call that the TS builder rewrites to a native-JS IIFE — no runtime helper required.
- Match arm LHS uses a small lookahead to disambiguate pattern vs. guard expression so `match(x is pat) { v > 5 => ... }` parses correctly.
- AgencyGenerator gains `formatPattern` and renders patterns / guards back to source so files round-trip cleanly.

## Tests

- New AST tests: `lib/types/pattern.ts`
- New parser tests: `lib/parsers/pattern.test.ts`
- New lowering tests: `lib/lowering/patternLowering.test.ts`
- New end-to-end agency tests under `tests/agency/`:
  - `destructuring.agency` — array/object/nested/rest destructuring
  - `forDestructuring.agency` — array & object pattern itemVar
  - `isOperator.agency` — boolean is, is + if + bindings, composed with `&&`
  - `patternMatch.agency` — pattern match arms with binders + default
  - `matchIsForm.agency` — `match(x is pattern)` form with guard expressions

All 3349 unit tests pass; spot-checked existing match / for / assignment fixtures pass; `make fixtures` regenerates clean.

## Docs

- New guide: [`docs/site/guide/pattern-matching.md`](packages/agency-lang/docs/site/guide/pattern-matching.md)
- Brief teaser added to `basic-syntax.md`
- New "Adding a new pattern form" section in `docs/dev/adding-features.md`

## Out of scope

- No exhaustiveness check on match blocks (same as today).
- No new fixture files under `tests/typescriptGenerator/` — the new agency tests already exercise the codegen path end-to-end.


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2217,6 +2217,35 @@ export class TypeScriptBuilder {
...
+      // Compiler-internal call emitted by the pattern lowering pass for
+      // `let { a, b, ...rest } = obj`. Args are [source, ["a", "b"]].
+      // Emit a native-JS IIFE that destructures the source and returns the
+      // rest object. No runtime helper involved.
+      //
+      //   __objectRest(source, ["a", "b"])
+      //   → (({ a: __a, b: __b, ...__r }) => __r)(<resolved source>)
+      const [sourceArg, keysArg] = node.arguments;
+      const sourceJs = this.str(this.processNode(sourceArg as Expression));
+      const keys =
```
egonSchiele: This is pretty hard to read, please write it as an if statement instead.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2217,6 +2217,35 @@ export class TypeScriptBuilder {
...
+        keysArg && (keysArg as AgencyArray).type === "agencyArray"
+          ? (keysArg as AgencyArray).items
```
egonSchiele: Are all these `as` statements really necessary?

packages/agency-lang/lib/typeChecker/builtins.ts
```diff
@@ -103,16 +86,16 @@ export const BUILTIN_FUNCTION_TYPES: Record<string, BuiltinSignature> = {
...
-  // No TS implementation; pure codegen. Intentionally NOT in BUILTIN_FUNCTIONS
-  // (lib/config.ts) since that registry is for runtime helper bindings.
-  getContext: { params: [], returnType: contextType },
+  checkpoint:    { params: [],         returnType: number },
+  getCheckpoint: { params: [number],   returnType: "any" },
+
+  // --- Compiler-internal: emitted by the pattern lowering pass for
+  // `let { a, b, ...rest } = obj`. The TS builder rewrites this call into
+  // a native-JS IIFE; there is no runtime function with this name. ---
+  __objectRest: { params: ["any", anyArray], returnType: "any" },
```
egonSchiele, line 101: This is no longer needed, right?

packages/agency-lang/tests/agency/destructuring/destructuring.agency
```diff
@@ -0,0 +1,19 @@
+node main() {
```
egonSchiele, line 1: Please create a new directory under tests/agency/ to put all the destructuring tests in so that they are grouped together.

packages/agency-lang/tests/agency/pattern-matching/isOperator.agency
```diff
@@ -0,0 +1,18 @@
+node main() {
```
egonSchiele, line 1: Similarly, all the "is" tests can go into a new pattern-matching directory under tests/agency/

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3252,3 +3379,226 @@ const _classParserInner: Parser<ClassDefinition> = (input: string) => {
...
+    if (!result.success) return result;
+    enforceRestAtEnd(result.result.elements);
+    return result;
+  },
+);
+
+const _matchPropertyWithValueParser: Parser<ObjectPatternProperty> = (
+  input: string,
+) => {
+  const parser = seqC(
```
egonSchiele: I see some code duplication here. For example, doesn't this method have a lot in common with _bindingPropertyWithValueParser?

packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -0,0 +1,734 @@
...
+            : this.lowerExpression(it as Expression),
+        ),
+      };
+    }
+    if (expr.type === "agencyObject") {
+      return {
+        ...expr,
+        entries: expr.entries.map((e) =>
+          "type" in e && e.type === "splat"
+            ? { ...e, value: this.lowerExpression(e.value) }
```
egonSchiele: Not sure I see the difference in the two arms of this ternary statement.

PR #151, opened 2026-05-16
memory: overhaul recall, cache, and runtime context plumbing

egonSchiele:
## Summary

Three intertwined memory-layer changes:

### 1. `MemoryCacheEntry` refactor (`lib/runtime/memory/cacheEntry.ts`)

Factor the in-process memory cache into one class that keeps the
knowledge graph, embeddings, and reverse index in sync on every
mutation. Replaces the previous scattered cache-update sites that
could drift apart and gives the new retrieval funnel a single
substrate to read from.

### 2. 3-tier recall funnel (`lib/runtime/memory/retrieval.ts`)

Replace the LLM-as-generator recall path with a layered
candidate-selection pipeline:

| Tier | Role |
| --- | --- |
| 1 | Bidirectional structured lookup on entity / relation indexes |
| 2 | Contextualized embedding similarity over the cache |
| 3 | LLM filters the candidate set — never generates ground truth |

### 3. Context-injected builtins (`lib/codegenBuiltins/contextInjected.ts`)

Replace the `getContext()` builder macro with a codegen-driven
mechanism that injects `__ctx` into a registered set of internal
builtins at compile time. User code never sees or holds a reference
to the runtime context.

- New `__internal_*` naming convention, routed through the existing
  `__`-prefixed direct-call codepath in `typescriptBuilder.ts`.
- Codegen always emits the registry's import block from
  `agency-lang/stdlib-lib/memory.js` — esbuild tree-shakes anything
  the call site didn't use.
- Typechecker rejects referencing these names as values
  (`let f = __internal_recall`) and flags any `__internal_*` name not
  in the registry as an unknown internal builtin.
- Drift-safeguard test in `lib/codegenBuiltins/contextInjected.test.ts`
  asserts each registry entry has a matching TS impl with
  `length === 1 + params.length`.
- Removes `lib/runtime/publicContext.ts` and the four `getContext`
  test fixtures.
- Eliminates a runtime bug where binding `getContext()` to a `const`
  caused the next runner step to be skipped.

## Why

- The previous module-level `currentContext` singleton raced across
  concurrent `runNode` calls in the same Node.js process. The
  `getContext()` builtin solved the race but exposed the runtime
  context to user code, recreating the same singleton risk in a
  different shape.
- Context-injection threads `__ctx` purely at codegen time, so user
  code can never hold a reference to it — no singletons, no races,
  no leaking internal fields.

## Docs

- Adds a "Reserved names" note to `docs/site/guide/basic-syntax.md`
  advising against user names starting with `__`.
- Rewrites the runtime-state section of
  `docs/dev/adding-a-module-to-the-agency-stdlib.md` to describe the
  context-injected pattern.

## Test plan

- `pnpm tsc --noEmit` — clean.
- `pnpm exec vitest run` — 3224 passed / 23 skipped, 0 failures.
- `make` — rebuilds stdlib, templates, and generated docs.
- Spot-checked the generated `stdlib/memory.js`: zero `getContext`
  references; every memory call is `await __internal_x(__ctx, ...)`.

## Plans

- `docs/superpowers/plans/2026-05-16-memory-recall-tiers.md`
- `docs/superpowers/plans/2026-05-17-context-injected-builtins.md`


packages/agency-lang/lib/runtime/memory/manager.ts
```diff
@@ -536,24 +549,25 @@ export class MemoryManager {
       model: this.model(),
     });
 
-    if (entry.summary) {
+    const prevSummary = entry.getSummary();
```
egonSchiele, line 570: What level of granularity is one memory entry at? Because I noticed that the entry has information about entities and observations, but each entry also seems to only have one summary. Is it fair to say that one memory entry equals message thread in agency? Could you please add a comment to the entry file explaining how entries map to agency programs? What happens if an agency program has multiple different threads? Would they all share the same memory entry?

packages/agency-lang/lib/runtime/memory/manager.ts
```diff
@@ -565,80 +579,73 @@ export class MemoryManager {
...
   private async generateEmbeddings(
-    entry: CacheEntry,
+    entry: MemoryCacheEntry,
     observationIds: string[]
   ): Promise<void> {
     for (const obsId of observationIds) {
-      const obsContent = findObservationContent(entry.graph, obsId);
-      if (!obsContent) continue;
+      const embedText = buildEmbedText(entry, obsId);
+      if (!embedText) continue;
```
egonSchiele: should we log an error?

packages/agency-lang/lib/runtime/memory/manager.ts
```diff
@@ -659,35 +666,53 @@ export class MemoryManager {
...
-    const entityNames = parseStringArray(response);
-    if (!entityNames) return [];
-    const ids: string[] = [];
-    for (const name of entityNames) {
-      const e = entry.graph.findEntityByName(name);
-      if (e && !ids.includes(e.id)) ids.push(e.id);
+    if (candidateIds.length === 0) return [];
+    const candidates = this.buildCandidateIndex(entry.getGraph(), candidateIds);
+    const prompt = retrievalTemplate({ candidates, query });
+    const response = await this._text(prompt, { model });
```
egonSchiele: needs a structured output json schema to guarantee the response format

packages/agency-lang/lib/runtime/memory/manager.ts
```diff
@@ -659,35 +666,53 @@ export class MemoryManager {
...
+    const ids = parseStringArray(response);
+    if (!ids) return [];
```
egonSchiele: log an error! it's likely that the LLM returned the response that didn't match the output we expected, and parseStringArray doesn't log an error if it fails to parse the data in the specified format. We need to log an error somewhere!

packages/agency-lang/lib/runtime/memory/manager.ts
```diff
@@ -659,35 +666,53 @@ export class MemoryManager {
...
+    // Hallucination guard: only keep ids we actually offered. Set
+    // lookup is O(1) per check; we also de-dupe in case the LLM
+    // returned the same id twice.
+    const offered = new Set(candidateIds);
```
egonSchiele: set intersection would be cleaner

packages/agency-lang/lib/runtime/memory/manager.ts
egonSchiele: use StringArraySchema.toJSONSchema()

packages/agency-lang/lib/runtime/memory/retrieval.ts
```diff
@@ -1,31 +1,91 @@
...
+        containsToken(lower, nameLower) ||
+        containsToken(nameLower, lower)
+      ) {
+        return true;
+      }
+    }
 
-    // Match by type (exact, case-insensitive)
+    // Type match — exact, case-insensitive. Types are short
+    // categorical labels ("person", "place"), so word-boundary
```
egonSchiele, line 78: do we normalize categorical labels in some way? For example, converting all plurals to singular?

packages/agency-lang/lib/runtime/memory/types.ts
```diff
@@ -44,7 +44,21 @@ export type EmbeddingEntry = {
...
+/**
+ * `formatVersion` distinguishes embeddings built from different source
+ * texts. Bumped to 2 when we started feeding the embedder a
+ * contextualized "{name} ({type}): {content}" string per observation
+ * instead of the bare observation content. Legacy files (no version)
+ * are treated as v1 and discarded on load — the next write rebuilds
+ * them in the current format.
+ *
+ * If the source-text shape changes again, bump this constant and
+ * update the load-side guard in `MemoryManager.getEntry`.
```
egonSchiele: What does the load side guard do if it finds a memory with an older version? Does it have a graceful fallback? Does it log an error?

packages/agency-lang/lib/codegenBuiltins/contextInjected.ts
```diff
@@ -0,0 +1,115 @@
...
+  /** If set, calls with at least `minParams` and at most `params.length`
+   *  args are valid (i.e. trailing optionals). Same semantics as
+   *  `BuiltinSignature.minParams`. */
+  minParams?: number;
+  /** If set, accepts unlimited extra args of this type after the fixed params. */
+  restParam?: VariableType | "any";
+  returnType: VariableType | "any";
+};
+
+export const CONTEXT_INJECTED_BUILTINS: Record<string, ContextInjectedBuiltin> = {
```
egonSchiele, line 58: There are some subtleties with this.
1. all of our other built-ins right now are something that has implemented in the agency runtime, or as part of the TS IR, builder or in the mustache template. These are the only ones that are implemented in TypeScript code.
2. And the function name is the name of the function in the TypeScript code.
3. And it's not in the agency runtime. It's in the standard library.

 I think that is a little surprising. What pattern do we expect future built-ins to follow? Would it make sense to keep all built-ins in the same file? And if not, would it make sense to group built-ins? 
eg

```
export const CONTEXT_INJECTED_BUILTINS = {
  "stdlib/memory.ts" => MEMORY_BUILTINS
}
```

I am just wondering how we can generalize some of this, instead of special casing logic in a few places. Similarly, I noticed that the builder generates an import statement for the relevant memory functions. But the logic of what file those functions are in, and so what import statement should be generated, is pretty far removed from the definition of what these functions actually are. So it would be nice to keep all of that information in one place.

PR #153, opened 2026-05-17
Per-branch span stacks via AsyncLocalStorage

egonSchiele:
## Summary

Replaces the fork-depth gate in `StatelogClient` with per-async-context span stacks via Node's `AsyncLocalStorage`, so spans (and the events they own) are correctly attributed inside concurrent fork/race branches.

## Background

`StatelogClient` is shared across an entire agent run. Fork/race branches run concurrently in JavaScript (`Promise.all` / `Promise.race`), so a single mutable `spanStack: SpanContext[]` cannot serve all branches at once — parallel `startSpan` / `endSpan` calls from different branches would pop each other's spans.

The previous workaround was to disable span tracking whenever `forkDepth > 0`. Functionally safe, but events emitted from inside a branch all inherited the outer span. The StateLog viewer therefore could not show "branch 0 did X while branch 1 did Y" — both branches looked identical from the trace's perspective.

## Solution

A per-async-context span stack via `AsyncLocalStorage`:

- `StatelogClient` now keeps:
  - `rootStack: SpanContext[]` — used outside branches.
  - `spanStorage: AsyncLocalStorage<SpanContext[]>` — used inside branches.
- `currentStack()` returns the ALS store if present, otherwise the root stack. All span-related methods (`startSpan`, `endSpan`, `currentSpan`, and the event-emission `parent_span_id` lookup) go through it.
- New `runInBranchContext(parentStack, fn)` runs `fn` inside `spanStorage.run([...parentStack], fn)`. Each call creates an independent ALS context with a defensive copy of the parent stack, so sibling branches can never see or mutate each other's pushes.
- New `snapshotStack()` returns a shallow copy of the current stack — used by the runner to seed each branch.
- Removed `enterFork()` / `exitFork()` / `forkDepth`.

In the runner:

- `runForkAll`, `runRace`, and `resumeRaceWinner` snapshot the current span stack once before scheduling branches and wrap every branch's `blockFn` in `runInBranchContext(snapshot, ...)`.
- Race winner-abort + branch-end emission is extracted into `abortLoserBranchesAndEmitEnds` (helper) to keep `runRace` under the 100-line structural limit.

## Interrupt / resume

Spans are intentionally ephemeral (not serialized in checkpoints). When a branch interrupts, its ALS context simply ends with the unwinding promise — nothing to persist. On resume the runtime re-enters the fork; each branch (re-running or returning a cached result) gets a fresh ALS context. This matches the existing behavior of opening a fresh `agentRun` span on resume.

## Tests

- Deleted the old `forkDepth gating` tests (gate no longer exists).
- Added a `branch span isolation (AsyncLocalStorage)` suite covering:
  - `snapshotStack` returns a decoupled copy.
  - Pushes inside `runInBranchContext` are invisible outside.
  - Concurrent branches see independent stacks across `await` boundaries.
  - Events emitted inside a branch carry the branch's local span as `parent_span_id`.
  - A branch ending a span it never opened is a no-op on the parent stack.

All 3452 unit tests pass. All 8 statelog integration scenarios pass.

## Dependency

This change builds on the StateLog infrastructure introduced in #147. It should land after #147 (or be rebased onto main once #147 merges).

### DIFF
--- a/packages/agency-lang/lib/runtime/runner.ts
+++ b/packages/agency-lang/lib/runtime/runner.ts
@@ -620,12 +620,11 @@ export class Runner {
       this.frame.popBranches();
     } finally {
       this.path.pop();
-      // Each branch is responsible for balancing its own
-      // enterFork/exitFork call via .finally/.then-or-catch handlers.
-      // We deliberately do NOT decrement here: in a race, losing branches
-      // may still be running, and decrementing prematurely would corrupt
-      // a parent fork's depth tracking and let nested code start logging
-      // spans into the shared stack while branches are still in flight.
+      // Each branch ran inside its own AsyncLocalStorage span context, so
+      // its pushes/pops are isolated from the parent. We can safely emit
+      // forkEnd and pop the fork span here even while loser race branches
+      // are still draining in the background — their stacks live in
+      // independent ALS contexts and cannot touch the parent's stack.
       this.ctx.statelogClient.forkEnd({
         forkId,
         mode,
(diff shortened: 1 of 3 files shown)


packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -768,6 +771,7 @@ export class Runner {
     ) => Promise<any>,
     stateStack: StateStack,
     branchStartTimes: number[],
+    parentSpanStack: import("../statelogClient.js").SpanContext[],
```
egonSchiele: no dynamic imports

PR #155, opened 2026-05-17
Fix ERR_MODULE_NOT_FOUND when running compiled output from a global agency-lang install

egonSchiele:
Fix `ERR_MODULE_NOT_FOUND` when running compiled output from a global `agency-lang` install

## Problem

When users install `agency-lang` globally (`npm i -g agency-lang`) and run a `.agency` file from a directory without a local install, the compile step succeeds but the spawned `node` process fails with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'agency-lang' imported from /Users/.../test.js
```

This is because `npm i -g` puts the package under npm's global prefix (e.g. `/opt/homebrew/lib/node_modules/agency-lang`) and symlinks its `bin` entry onto `$PATH`. The `agency` CLI works from anywhere because `$PATH` finds the binary, but **Node's module resolver never consults the global prefix** — for bare specifiers like `import "agency-lang"`, it only walks `node_modules` upward from the importing file. So the compiled output cannot find the package it was compiled against.

## Fix

In `lib/cli/commands.ts`, before spawning `node` on the compiled output:

1. **Probe** whether `agency-lang` resolves from the compiled file's directory using `createRequire(...).resolve("agency-lang")`. This matches Node's actual resolution behavior for bare specifiers, so a `true` result is a reliable proxy for "the spawned process will be able to import it."
2. If the probe fails, **locate the running CLI's own package root** by walking up from `import.meta.url` until we find a `package.json` with `"name": "agency-lang"`. This works for global installs, `npm link`, and local dev checkouts equally well — no dependency on `npm config` or shelling out to `npm root -g`.
3. **Inject `NODE_PATH`** on the spawned process's env, pointing at:
   - the `node_modules` directory containing the package (so `agency-lang` itself plus any hoisted siblings like `zod` resolve), and
   - the package's own `node_modules` (so non-hoisted transitive deps resolve — npm typically nests deps for `-g` installs).
4. **Print a warning** telling the user the proper fix is `npm install agency-lang zod` in their project directory.

The fallback is silent and inert when not needed: in a project that has `agency-lang` installed locally (or inside the monorepo, or via `npm link`), the probe succeeds and `NODE_PATH` is left untouched.

## Files changed

- `lib/cli/commands.ts` — two new helpers (`isAgencyLangResolvableFrom`, `findGlobalAgencyLangSearchPaths`) and the `NODE_PATH` injection in `run()`. The existing `AGENCY_RESUME_FILE` env handling is preserved.

## Manual verification

```bash
# 1. Global install scenario (the original bug)
cd ~              # no node_modules here
agency run test.agency
# → prints fallback warning, runs successfully

# 2. Local install scenario
cd ~/myproject
npm install agency-lang zod
agency run test.agency
# → no warning, NODE_PATH untouched

# 3. Monorepo dev scenario
cd packages/agency-lang
pnpm run agency examples/helloWorld.agency
# → no warning, no behavior change
```

`pnpm exec tsc --noEmit` is clean.

## Notes / follow-ups

- No new unit tests yet. Both helpers are easy to cover with tmpdir fixtures (`findGlobalAgencyLangSearchPaths` against a fake `node_modules/agency-lang/package.json` layout, `isAgencyLangResolvableFrom` against a dir with and without a local install). Happy to add these if reviewers want.
- The warning intentionally points users at the proper fix (`npm install` locally) rather than silently papering over the issue — the `NODE_PATH` route is a convenience, not a recommendation.

### DIFF
--- a/packages/agency-lang/lib/cli/commands.ts
+++ b/packages/agency-lang/lib/cli/commands.ts
@@ -9,7 +9,9 @@ import { formatErrors, typeCheck } from "@/typeChecker/index.js";
 import { spawn } from "child_process";
 import { transformSync } from "esbuild";
 import * as fs from "fs";
+import { createRequire } from "module";
 import * as path from "path";
+import { fileURLToPath } from "url";
 
 import {
   getStdlibDir,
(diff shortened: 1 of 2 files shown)


packages/agency-lang/lib/cli/commands.ts
```diff
@@ -276,6 +278,73 @@ export async function formatFile(
...
+            paths.push(parent);
+          }
+          const ownNodeModules = path.join(dir, "node_modules");
+          if (fs.existsSync(ownNodeModules)) {
+            paths.push(ownNodeModules);
+          }
+          return paths;
+        }
+      } catch {
+        // Malformed package.json — keep walking up.
```
egonSchiele, line 339: this can lead to an infinite loop

packages/agency-lang/lib/cli/commands.ts
```diff
@@ -293,9 +362,32 @@ export function run(
...
+    const fallbackPaths = findGlobalAgencyLangSearchPaths();
+    if (fallbackPaths.length > 0) {
+      const existing = env.NODE_PATH ? env.NODE_PATH.split(path.delimiter) : [];
+      env.NODE_PATH = [...fallbackPaths, ...existing].join(path.delimiter);
+      console.warn(
+        `\nWarning: 'agency-lang' is not installed locally in ${outputDir}.\n` +
+          `Falling back to the globally installed copy at:\n` +
+          fallbackPaths.map((p) => `  ${p}`).join("\n") +
+          `\nFor a more reliable setup, install it in your project:\n` +
+          `  npm install agency-lang zod\n`,
```
egonSchiele, line 387: remove `zod`, no longer needed

PR #157, opened 2026-05-17
Fix agency run under global install, add agency pack for portable scripts

egonSchiele:
## Motivation

When a user installs the agency CLI globally (`npm install -g agency-lang`) and runs `agency compile foo.agency && node foo.js`, they hit:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'agency-lang' imported from /path/to/foo.js
```

This is a classic Node footgun: global `npm install` and Node's module resolver are separate mechanisms. Node does not look in the global `node_modules` for bare imports. Three related improvements address this:

1. Make the obvious workflow work.
2. Tell the user something useful instead of silently leaving them with a confusing stack trace.
3. Give them a way to ship a `.agency` program as a self-contained file.

### 1. Detect install kind (`lib/cli/installLocation.ts`)

Small helper that classifies the CLI install as `global`, `local`, or `workspace` from `import.meta.url`. Used by the next two changes.

### 2. Fix `agency run` for global installs

The compiled output is ESM and Node's ESM resolver **does not honor `NODE_PATH`** (only CommonJS does). The fix is an ESM loader hook: two tiny `.mjs` files are shipped in `dist/lib/cli/runShim/`, and `agency run` spawns the child Node with `--import=<file-url-to-register.mjs>`. When the default resolver can't find a bare specifier, the hook retries with `parentURL` set to a file inside agency-lang's package — letting Node walk that package's own `node_modules`.

`agency compile` now also prints a one-paragraph note (to stderr) when invoked from a global install, steering the user to `agency run` or `agency pack` instead of running the compiled output directly with `node`.

### 3. `agency pack`

```bash
agency pack foo.agency -o foo.js   # produces a portable single-file script
./foo.js                            # runs anywhere Node is installed
```

Uses the already-bundled `esbuild` to inline the agency runtime, smoltalk, zod, and any user imports into one ESM file with a `#!/usr/bin/env node` shebang and mode `0o755`. Only Node built-ins are kept external.

Two subtleties:
- esbuild's CJS-in-ESM shim throws `Dynamic require of "child_process"` for transitive CJS deps; a banner installs a real `createRequire` so it works.
- `lib/importPaths.ts` previously called `findPackageRoot(__dirname)` at module-load time, which crashed packed binaries that lived in directory trees with no `package.json` above them. Made lazy — only fires if/when the user code actually needs stdlib paths.

`--target=node` only for now. `--target=sea` (Node single-executable application) is a planned follow-up.

## Verification

All three behaviors verified end-to-end against a real `npm install -g`:

| Behavior | Result |
|---|---|
| Global install detected → friendly warning from `agency compile` | ✅ |
| `agency run` works under global install (was previously broken) | ✅ |
| `agency pack` produces 3MB portable .js that runs from a directory with no `node_modules` | ✅ |

Test suite: **209 files, 3641 tests, all passing** (baseline 206/3626; this PR adds 3 test files and 15 tests).

## Commits

1. `feat(cli): add install-location helpers`
2. `fix(cli): make agency run work for global installs` *(NODE_PATH attempt + warning)*
3. `fix(cli): use ESM loader hook instead of NODE_PATH` *(real fix once I confirmed ESM ignores NODE_PATH)*
4. `feat(cli): add agency pack for portable single-file scripts`
5. `docs(guide): add page on running agents, the global-install gotcha, and pack`

## Out of scope (follow-ups)

- `agency pack --target=sea` for Node single-executable applications, and `--target=bun-compile` for a `bun build --compile` binary.
- Inlining the `.agency` stdlib for packed programs that need it at runtime (most don't — the stdlib is compile-time consumed).
- A test-mode `agency pack` that emits the bundle next to a `package.json` with `"type": "module"` so users with `.js` outputs don't see Node's `MODULE_TYPELESS_PACKAGE_JSON` warning.

### DIFF
--- a/packages/agency-lang/lib/cli/pack.ts
+++ b/packages/agency-lang/lib/cli/pack.ts
@@ -1,0 +1,145 @@
+import * as fs from "fs";
+import * as path from "path";
+import { fileURLToPath } from "url";
+import { build } from "esbuild";
+import { AgencyConfig } from "@/config.js";
+import { compile } from "./commands.js";
+
+// Locate this package's install root by walking up from this module's
+// location until we find a package.json whose name is "agency-lang".
+// Works for both the built `dist/lib/cli/pack.js` and the dev-tree
+// `lib/cli/pack.ts` paths.
+function agencyLangInstallRoot(): string {
+  let dir = path.dirname(fileURLToPath(import.meta.url));
+  while (true) {
+    const pkgJson = path.join(dir, "package.json");
+    if (fs.existsSync(pkgJson)) {
+      try {
+        const parsed = JSON.parse(fs.readFileSync(pkgJson, "utf-8"));
+        if (parsed.name === "agency-lang") return dir;
+      } catch {
+        /* keep walking */
+      }
+    }
+    const parent = path.dirname(dir);
+    if (parent === dir) {
+      throw new Error("Could not locate agency-lang package root");
+    }
+    dir = parent;
+  }
+}
+
+export type PackTarget = "node";
+
+export type PackOptions = {
+  config: AgencyConfig;
+  inputFile: string;
+  outputFile: string;
+  target: PackTarget;
+};
+
+const SHEBANG = "#!/usr/bin/env node\n";
+
+// Only Node built-ins are kept external; everything else (agency-lang,
+// smoltalk, zod, user .agency imports, etc.) is bundled into the output
+// so the result runs anywhere Node is available without an install.
+const NODE_BUILTINS = [
+  "assert",
+  "async_hooks",
+  "buffer",
+  "child_process",
+  "cluster",
+  "console",
+  "constants",
+  "crypto",
+  "dgram",
+  "dns",
+  "events",
+  "fs",
+  "http",
+  "http2",
+  "https",
+  "inspector",
+  "module",
+  "net",
+  "os",
+  "path",
+  "perf_hooks",
+  "process",
+  "punycode",
+  "querystring",
+  "readline",
+  "repl",
+  "stream",
+  "string_decoder",
+  "timers",
+  "tls",
+  "trace_events",
+  "tty",
+  "url",
+  "util",
+  "v8",
+  "vm",
+  "worker_threads",
+  "zlib",
+  "node:*",
+];
+
+export async function pack(opts: PackOptions): Promise<void> {
+  // 1. Compile the .agency file to .js using the existing pipeline.
+  //    We write to a sibling temp file so relative paths inside the user's
+  //    source dir continue to resolve the same way they would for `compile`.
+  const tmpJs = opts.inputFile.replace(/\.agency$/, ".__pack__.js");
+  const compiled = compile(opts.config, opts.inputFile, tmpJs);
+  if (!compiled) {
+    throw new Error(`compile failed for ${opts.inputFile}`);
+  }
+
+  try {
+    // 2. Bundle with esbuild — inline agency-lang, smoltalk, zod, and any
+    //    transitive .js imports into one ESM file. The user's working
+    //    directory may have no node_modules, so explicitly tell esbuild to
+    //    resolve bare specifiers from agency-lang's install location.
+    const installRoot = agencyLangInstallRoot();
+    const result = await build({
+      entryPoints: [compiled],
+      bundle: true,
+      platform: "node",
+      format: "esm",
+      target: "node20",
+      write: false,
+      outfile: opts.outputFile,
+      external: NODE_BUILTINS,
+      nodePaths: [
+        path.dirname(installRoot), // the directory containing agency-lang/
+        path.join(installRoot, "node_modules"), // its own deps
+      ],
+      // Some transitive CJS deps call `require("child_process")` etc.
+      // esbuild's bundler can't statically rewrite those for ESM output,
+      // so it emits a runtime shim that throws "Dynamic require ...".
+      // Replace that with a real createRequire so they work.
+      banner: {
+        js: [
+          "// Built by `agency pack`",
+          'import { createRequire as __pack_createRequire } from "node:module";',
+          "const require = __pack_createRequire(import.meta.url);",
+        ].join("\n"),
+      },
+      logLevel: "silent",
+    });
+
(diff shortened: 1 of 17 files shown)


packages/agency-lang/lib/cli/runShim/resolver.mjs
```diff
@@ -0,0 +1,49 @@
...
+    if (err && err.code === "ERR_MODULE_NOT_FOUND" && isAgencyLangSpecifier(specifier)) {
+      // Retry only for agency-lang specifiers. If this also fails,
+      // fall through and rethrow the original error.
+      try {
+        return await nextResolve(specifier, {
+          ...context,
+          parentURL: fakeParentURL,
+        });
+      } catch {
+        /* fall through */
```
egonSchiele: should you log a message here?

packages/agency-lang/lib/cli/pack.ts
```diff
@@ -0,0 +1,249 @@
...
+import { build, type Plugin } from "esbuild";
+import { AgencyConfig } from "@/config.js";
+import { SymbolTable } from "@/symbolTable.js";
+import { compile } from "./commands.js";
+
+// Locate this package's install root by walking up from this module's
+// location until we find a package.json whose name is "agency-lang".
+// Works for both the built `dist/lib/cli/pack.js` and the dev-tree
+// `lib/cli/pack.ts` paths.
+function agencyLangInstallRoot(): string {
```
egonSchiele, line 15: seems similar to findPackageRoot in lib/importPaths.ts

packages/agency-lang/lib/cli/pack.ts
```diff
@@ -0,0 +1,249 @@
...
+  outputFile: string;
+  target: PackTarget;
+};
+
+const SHEBANG = "#!/usr/bin/env node\n";
+
+// Only Node built-ins are kept external; everything else (agency-lang,
+// smoltalk, zod, user .agency imports, etc.) is bundled into the output
+// so the result runs anywhere Node is available without an install.
+const NODE_BUILTINS = [
```
egonSchiele: how do we keep this list up to date?

packages/agency-lang/lib/cli/pack.ts
```diff
@@ -0,0 +1,249 @@
...
+    //    agency-lang package and its transitive deps inline. The
+    //    resolver plugin handles `agency-lang*` specifiers using
+    //    agency-lang's own require resolution — no nodePaths, so we
+    //    cannot accidentally pull in globally-installed user packages.
+    const installRoot = agencyLangInstallRoot();
+    const result = await build({
+      entryPoints: [compiled],
+      bundle: true,
+      platform: "node",
+      format: "esm",
```
egonSchiele: Instead of hard coding these, do we want to expose these values in the config so the users can pick a different format and target if they want, and we can set some default options?

packages/agency-lang/lib/cli/pack.ts
```diff
@@ -0,0 +1,249 @@
...
+      // emits a runtime shim that throws "Dynamic require ...".
+      // Install a real createRequire so they work.
+      banner: {
+        js: [
+          "// Built by `agency pack`",
+          'import { createRequire as __pack_createRequire } from "node:module";',
+          "const require = __pack_createRequire(import.meta.url);",
+        ].join("\n"),
+      },
+      logLevel: "silent",
```
egonSchiele: If the user has verbose set to true, should we emit logging?

packages/agency-lang/lib/cli/pack.ts
```diff
@@ -0,0 +1,249 @@
...
+    });
+
+    if (result.outputFiles.length !== 1) {
+      throw new Error(
+        `agency pack: expected exactly one output file from esbuild, got ${result.outputFiles.length}`,
```
egonSchiele: Also print out exactly what output files the user passed in for easier debugging.

packages/agency-lang/lib/cli/pack.ts
```diff
@@ -0,0 +1,249 @@
...
+      );
+    }
+
+    // 5. Write the bundle. mode on writeFileSync only takes effect when
+    //    the file is created — if the user re-packs over an existing
+    //    file, Node preserves the old permissions. chmod explicitly so
+    //    the output is always executable.
+    const outDir = path.dirname(path.resolve(opts.outputFile));
+    fs.mkdirSync(outDir, { recursive: true });
+    fs.writeFileSync(opts.outputFile, SHEBANG + result.outputFiles[0].text, {
```
egonSchiele, line 188: The comment says we preserve the old permissions, but this code seems to always change the mode to 755.

packages/agency-lang/lib/cli/pack.ts
```diff
@@ -0,0 +1,249 @@
...
+      //  - the pack output itself (it might coincidentally share a name)
+      if (
+        preExistingJs.has(p) ||
+        path.resolve(p) === absOutput ||
+        !fs.existsSync(p)
+      ) {
+        continue;
+      }
+      try {
+        fs.unlinkSync(p);
```
egonSchiele: Check if it is a directory and if it is, dont delete.

packages/agency-lang/lib/cli/pack.ts
```diff
@@ -0,0 +1,249 @@
...
+      } catch {
+        /* best-effort */
```
egonSchiele: print an error

packages/agency-lang/package.json
```diff
@@ -8,7 +8,7 @@
     "test:run": "vitest run",
     "test:agency": "time node ./dist/scripts/agency.js test tests/agency -p 12",
     "test:agency-js": "time node ./dist/scripts/agency.js test js tests/agency-js -p 12",
-    "build": "rm -rf dist/ && tsc && tsc-alias && cp -r lib/agents dist/lib",
+    "build": "rm -rf dist/ && tsc && tsc-alias && cp -r lib/agents dist/lib && mkdir -p dist/lib/cli/runShim && cp lib/cli/runShim/*.mjs dist/lib/cli/runShim/",
```
egonSchiele: The length of this command is getting unwieldy, anything we can do?

PR #168, opened 2026-05-19
feat: result pattern matching (success/failure in is and match)

egonSchiele:
# Result Pattern Matching

Adds `success` and `failure` as pattern keywords in `is` and `match` positions for ergonomic Result-type unwrapping.

## Summary

Today, unwrapping a Result is verbose:

```agency
if (isSuccess(result)) {
  let value = result.value
  print(value)
}
```

After this PR you can write:

```agency
if (result is success(value)) {
  print(value)
}
```

## New syntax

**Boolean test:**
```agency
let ok    = r is success
let bad   = r is failure
```

**Binding in `if` / `while`:**
```agency
if (r is success(v))  { print(v) }
if (r is failure(e))  { print(e) }
```

**Match arms:**
```agency
match (r) {
  success(v) => print("got ${v}")
  failure(e) => print("err: ${e}")
}
```

**Combined with `match(expr is pattern)`:**
```agency
match (r is success(v)) {
  v > 10 => print("big")
  _      => print("small")
}
```

**Nested inside other patterns:**
```agency
match (pair) {
  [success(v), _] => print(v)
  _               => print("none")
}
```

`success(v)` binds `result.value` to `v`; `failure(e)` binds `result.error` to `e`. Bare `is success` / `is failure` lowers to a call to `isSuccess(...)` / `isFailure(...)`.

## Architecture

- **New AST node** `ResultPattern` in `lib/types/pattern.ts`, added to the `MatchPattern` union (and the `AgencyNode` union for the formatter).
- **Parser** intercepts `success` / `failure` in pattern position (`_matchPatternParser`), placed before `variableNameParser`. Empty parens (`success()`) and non-identifier bindings (`success(123)`) throw committed parse errors so the outer `or` cannot fall through to `variableNameParser`.
- **Lowering** (`patternLowering.ts`) does all desugaring:
  - `patternToCondition` emits `isSuccess(source)` / `isFailure(source)`.
  - `extractBindings` emits `const <binding> = <source>.value` / `.error`.
  - `assertNoBindersInBoolIs` rejects `let x = r is success(v)`.
  - `hasPatternArms` recognises `resultPattern` as triggering match-arm lowering.
- **Formatter** (`agencyGenerator.ts`) adds a case so `pnpm run fmt` round-trips result patterns.
- No changes required in the typechecker, preprocessor, or TS builder. `isSuccess` / `isFailure` were already registered builtins.

## What this does NOT include

(deliberately, per the spec)

- No guard clauses on result-pattern arms (use `match(expr is success(v))` form).
- No destructuring in `let`/`const` or `for` positions.
- No nested patterns *inside* result patterns (`success({ name })` is not supported — bind the value then destructure separately).
- No empty-paren forms.
- `failure(e)` exposes only the error string. For `checkpoint` / `functionName` / `args`, use the traditional `if (isFailure(result))` form and access fields directly.

## Tests

- **Parser** — 13 new tests in `lib/parsers/pattern.test.ts` covering bare, binding, empty-paren rejection, non-identifier rejection, `successful` boundary, nested in array, nested in object, and the `is`-expression context.
- **Lowering** — 9 new tests in `lib/lowering/patternLowering.test.ts` covering bare boolean form, `if`-binding, `while`-binding, nested-in-array lowering, pure-boolean-context binders rejected, and match-arm if/else-if chain.
- **Formatter** — 6 new tests in `lib/backends/agencyGenerator.test.ts` covering all four shapes plus nested.
- **End-to-end** — 5 new Agency execution tests in `tests/agency/pattern-matching/`:
  - `resultPatternIs` — `if (r is success(v))` + `if (r is failure(e))`
  - `resultPatternMatch` — `match` with success/failure arms
  - `resultPatternMatchIs` — `match (r is success(v))` form with guards
  - `resultPatternBoolIs` — bare `is success` / `is failure`
  - `resultPatternNested` — `[success(v), _]` arms

**Verification:** all 3893 unit tests pass; all 10 pattern-matching execution tests pass; `pnpm run lint:structure` clean; full build clean.

## Plan & Spec

- Plan: `packages/agency-lang/docs/superpowers/plans/2026-05-18-result-pattern-matching.md`
- Spec: `packages/agency-lang/docs/superpowers/specs/2026-05-18-result-pattern-matching-design.md`


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3524,6 +3525,85 @@ export const objectBindingPatternParser: Parser<ObjectPattern> = label(
...
+
+    // Boundary check — keyword must not be followed by an identifier char,
+    // so `successful` does NOT match.
+    const boundary = not(varNameChar)(kwResult.rest);
+    if (!boundary.success) {
+      return fail("not a result pattern keyword boundary")(input);
+    }
+
+    const rest = kwResult.rest;
+    if (rest.length > 0 && rest[0] === "(") {
```
egonSchiele: I think some of this could be simplified by using the `parseError` combinator

PR #173, opened 2026-05-19
feat: doc string interpolation with type-checker safety

egonSchiele:
## Summary

Adds `${expr}` interpolation in `"""..."""` doc strings, with a type-checker error when users try to interpolate function/node parameters.

Spec: [`docs/superpowers/plans/2026-05-19-docstring-interpolation.md`](docs/superpowers/plans/2026-05-19-docstring-interpolation.md)
Plan: [`docs/superpowers/plans/2026-05-19-docstring-interpolation-impl.md`](docs/superpowers/plans/2026-05-19-docstring-interpolation-impl.md)

## Motivating example

```agency
const toolVersion = "2.0"

def versionedGreet(name: string): string {
  """
  Greets someone. Tool version: ${toolVersion}.
  """
  return "Hello, ${name}"
}
```

The generated tool description, sent to the LLM, becomes `"Greets someone. Tool version: 2.0."` — the global is interpolated at module-load time. By contrast:

```agency
def greet(name: string) {
  """Greets ${name}."""   // ← compile error
}
```

…fails with a clear type-checker error pointing at `${name}`, because parameter values are not bound when the tool description is constructed.

## Approach

- **AST**: Collapse the parser's `DocString` type into the existing `MultiLineStringLiteral`. Add `loc` to `InterpolationSegment` so the new error can point at the offending `${...}` rather than the whole doc string.
- **Parser**: `docStringParser` delegates to `multiLineStringParser`, preserving historic leading/trailing-whitespace trimming on edge text segments (inner whitespace is preserved).
- **Preprocessor**: New pass walks function/node doc-string interpolation expressions and resolves scope (global / imported / functionRef), matching how the same nodes are resolved at module top-level.
- **TS builder**: Description emission uses `generateStringLiteralNode`. The runtime preamble gains two new lines so doc-string interpolations can be evaluated at module load time:
  - `const __ctx = __globalCtx;`
  - `__initializeGlobals(__globalCtx);` (hoisted function call — safe forward reference)
- **Agency generator (formatter)**: Reconstructs `"""..."""` from segments so `${expr}` round-trips faithfully.
- **Doc CLI + LSP**: Share a new `docStringText` helper that renders the doc string as plain text, reconstructing interpolation segments as `${expr}` via the agency generator's expression printer.
- **Type checker**: Rejects top-level `${param}` interpolation in doc strings with a clear error explaining the lifecycle.

## Architectural change worth flagging

The two-line runtime preamble addition (`__ctx` alias + eager `__initializeGlobals`) is **why every `tests/typescriptGenerator/*.mjs` and `tests/typescriptBuilder/*.mjs` fixture changed**. The change is benign — all 4050 tests pass, no behavior regressions — but it's a real architectural addition. Without it, doc-string interpolation of globals would produce `ReferenceError: __ctx is not defined` at module load time.

The eager init is fire-and-forget (the function is `async` but for pure-literal global initializers all side-effects run synchronously). Re-initialization is short-circuited by the `markInitialized` flag.

## Verification

- `npx tsc --noEmit`: clean
- `npx vitest run` (full suite): **4050 tests pass, 0 failures**

## New tests

- `lib/typeChecker/docstringParamInterpolation.test.ts` — 4 cases: function param errors, node param errors, global allowed, no-interpolation allowed
- `lib/backends/docstringInterpolationRuntime.test.ts` — 2 cases: compile-output check + runtime check that `toolRegistry["versionedGreet"].toolDefinition.description === "Greets someone. Tool version: 2.0."`
- 5 new parser fixtures in `lib/parsers/function.test.ts` covering leading/trailing interpolation and inner-whitespace preservation
- `tests/agency/docstring-interpolation.{agency,test.json}` — runtime smoke test
- Roundtrip case appended to `tests/formatter/roundtrip.agency`

## Files touched

- `lib/types/{literals,function,graphNode}.ts`
- `lib/parsers/parsers.ts`, `lib/parsers/function.test.ts`
- `lib/backends/{typescriptBuilder,agencyGenerator}.ts`
- `lib/preprocessors/typescriptPreprocessor.ts`
- `lib/typeChecker/index.ts`
- `lib/cli/doc.ts`, `lib/lsp/completion.ts`
- `lib/utils/docStringText.ts` (new)
- `tests/typescriptGenerator/docstrings.{agency,mjs}` (interpolation fixture)
- `tests/formatter/roundtrip.agency`
- All `tests/typescriptGenerator/*.mjs`, `tests/typescriptBuilder/*.mjs`, `tests/typescriptPreprocessor/*.json`, `tests/debugger/*.ts` — regenerated to pick up the new preamble lines and scope pass.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -1262,6 +1262,29 @@ export class TypeScriptBuilder {
   }
 
   /**
+   * Returns true if any function or graph node in the compilation unit
+   * has a doc string with at least one interpolation segment.
+   */
+  private hasDocStringInterpolation(): boolean {
+    const fns = Object.values(this.compilationUnit.functionDefinitions);
+    for (const fn of fns) {
+      if (
+        fn.docString?.segments.some((s) => s.type === "interpolation")
+      ) {
+        return true;
+      }
+    }
+    for (const node of this.compilationUnit.graphNodes) {
+      if (
+        node.docString?.segments.some((s) => s.type === "interpolation")
+      ) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
    * Generate __toolRegistry as an empty object. AgencyFunction.create() calls
    * register local functions into it. Imported and builtin tools are registered
    * here directly. The reviver is bound at the end.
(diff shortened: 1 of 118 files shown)


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2959,6 +2983,26 @@ export class TypeScriptBuilder {
...
+    // shadow this with their own `__ctx2` binding), and trigger global
+    // initialization eagerly so any interpolated reads resolve. This is
+    // gated to modules that actually use the feature, so other modules
+    // see no preamble change.
+    //
+    // Caveat: `__initializeGlobals` is async; for modules with
+    // `static` declarations or async global initializers, top-level
+    // interpolation may still see uninitialized values. Synchronous
+    // literal globals are populated before the function returns.
+    if (this.hasDocStringInterpolation()) {
```
egonSchiele, line 3006: Is this the right approach? It seems like a lot of work. Would it make more sense to evaluate the description of a function lazily when needed?

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2724,22 +2723,38 @@ export const modifiedAssignmentParser: Parser<Assignment> = withLoc((input: stri
...
-      str('"""'),
-      capture(map(many1Till(str('"""')), trim), "value"),
-      str('"""'),
-    ),
-  );
-  const result = parser(input);
-  if (result.success) {
-    result.result.value = stripSentinels(result.result.value);
+export const docStringParser: Parser<MultiLineStringLiteral> = (input: string) => {
+  const result = multiLineStringParser(input);
```
egonSchiele: To be honest, I'm not sure that any of this is needed. I think you could just use the multiLineStringParser. Why do we need the trimming, etc?

packages/agency-lang/lib/typeChecker/index.ts
```diff
@@ -241,6 +241,36 @@ export class TypeChecker {
       checkValidatedParamReturn(name, def);
     }
 
+    // 1f. Doc strings must not interpolate function/node parameters.
+    // The description is built when the tool object is constructed at module
+    // load time — long before the function is ever called — so parameter
+    // values are simply not bound yet. Catch the obvious case here:
+    // `${param}` as a top-level interpolation expression.
+    const checkDocStringParams = (
```
egonSchiele, line 249: just curious, do we also type check interpolation segments in other string literals?

PR #174, opened 2026-05-19
feat: @validate and @jsonSchema type annotations + std::validators/schemas/types

egonSchiele:
## Summary

Adds two type-level annotations and a small standard library that uses them.

### `@validate(fn1, fn2, ...)`

Attaches an async validator chain that runs **after** the Zod parse on any
`!`-validated value. Validators are transform-capable — `success(value)`
threads the new value into the next validator — and short-circuit on the
first failure. Plain JS async functions and Agency `def` functions
(`AgencyFunction` instances) are both accepted.

```ts
import { isEmail } from "std::validators"

@validate(isEmail)
type Email = string

def main() {
  const e: Email! = "user@example.com"   // validated
}
```

### `@jsonSchema({...})`

Emits a Zod `.meta(...)` call on the generated schema, so structured-output
LLM responses and `toJSONSchema(...)` output pick up custom JSON Schema
fields (`format`, `description`, `minimum`, etc.).

```ts
@validate(isEmail)
@jsonSchema({ format: "email", description: "work address" })
type WorkEmail = string
```

Validation only runs when `!` is present — same rule as the existing
`schemas` feature — so users have one mental model for "when does
validation fire".

### New stdlib modules

- `std::validators` — `isEmail`, `isUrl`, `isUuid`, `isInt`,
  `isPositive`, `isNegative`.
- `std::schemas` — `emailFormat`, `urlFormat`, `uuidFormat`,
  `dateTimeFormat`, `dateFormat`, `ipv4Format`, `ipv6Format`. Plain object
  constants intended for spread into `@jsonSchema({...})`.
- `std::types` — pre-baked `Email`, `Url`, `Uuid` aliases combining the
  matching validator + format helper.

## Implementation notes

- `Tag.arguments` widened from `string[]` to `Expression[]`. The tag parser
  now accepts a restricted expression subset (literals, identifiers,
  function calls, object literals with spread). Existing `@goal` /
  `@optimize` round-trip identically.
- Preprocessor `attachTags` extended to `typeAlias`; object-type parser
  grows tag accumulation for `ObjectProperty`.
- `TypeAliasEntry` carries `tags`; `resolveType` propagates them through
  both generic and non-generic alias references. `mergeTagSets` gives
  alias-tags-first concat for `@validate` and right-merge for
  `@jsonSchema`.
- New runtime helpers `__validateChain` and `__validateChainRecursive`
  run async validator chains after the sync Zod parse, recurse into
  nested types via a structural descriptor, dispatch unions to the
  matching branch only, skip inner validators on `null` / `undefined`,
  and bail past a depth cap of 64 (configurable).
- Type aliases whose body carries `@validate` emit a co-located
  `(Foo as any).__agency_descriptor = ...` so consumer modules get the
  full validation tree without needing to re-import the validator or
  schema-helper identifiers themselves.
- `importResolver` flows static-`const` exports through Agency imports
  so `import { emailFormat } from "std::schemas"` actually arrives in
  the generated TS.
- `agencyGenerator` round-trips alias-level tags so `pnpm run fmt`
  doesn't drop annotations.

## Verification

- `pnpm exec tsc --noEmit`: clean
- `pnpm test:run`: 4086 / 4086 pass
- `pnpm run agency test tests/agency/validation -p 12`: 39 / 39 pass
  (including the three new end-to-end tests added in this PR)
- `make` (full build + stdlib compile + docs): clean

## Spec and plan

- Spec: `packages/agency-lang/docs/superpowers/specs/2026-05-19-type-validation-and-json-schema-annotations-design.md`
- Plan: `packages/agency-lang/docs/superpowers/plans/2026-05-19-type-validation-and-json-schema-annotations-impl.md`

## Fast-follow items (not in this PR)

- Hook `jsonSchemaArgValidator` into the type-check pass so the
  static-expression restriction surfaces as a diagnostic. The helper
  and its unit tests are already in place.
- Property-level tag round-trip in `agencyGenerator` (alias-level done;
  object-property emit-back TBD).
- Remaining cells of the end-to-end test matrix (3 of ~15 done).
- User-facing docs (`docs/site/guide/annotations.md`).
- The legacy `# description` codemod and parser removal, always
  intended as a separate PR.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptGenerator/validationDescriptor.ts
+++ b/packages/agency-lang/lib/backends/typescriptGenerator/validationDescriptor.ts
@@ -1,13 +1,15 @@
 import type { Tag, VariableType, TypeAliasEntry } from "../../types.js";
+import type { TsNode, TsObjectEntry } from "@/ir/tsIR.js";
+import { ts } from "@/ir/builders.js";
 import { mapTypeToValidationSchema } from "./typeToZodSchema.js";
 import { tagArgToTs } from "./tagArgToTs.js";
 import { mergeTagSets } from "@/typeChecker/mergeTags.js";
 
 /**
- * Render a TS source expression that evaluates to a runtime
- * `TypeValidationDescriptor`. Mirrors `mapTypeToValidationSchema`'s
- * structural recursion but threads validator lists collected from
- * `@validate(...)` tags at each level.
+ * Build a TS IR node that evaluates to a runtime `TypeValidationDescriptor`.
+ *
+ * Mirrors `mapTypeToValidationSchema`'s structural recursion but threads
+ * validator lists collected from `@validate(...)` tags at each level.
  *
  * Use this only when the type carries at least one `@validate` tag
  * somewhere — callers should check via `hasAnyValidateTag` first.
@@ -19,11 +21,11 @@ import { mergeTagSets } from "@/typeChecker/mergeTags.js";
  * tags reachable only through a non-generic alias reference would be
  * silently dropped.
  */
-export function buildValidationDescriptorTs(
+export function buildValidationDescriptor(
   variableType: VariableType,
   typeAliases: Record<string, VariableType>,
   typeAliasesFull?: Record<string, TypeAliasEntry>,
-): string {
+): TsNode {
   return descriptor(variableType, typeAliases, typeAliasesFull ?? {});
 }
 
@@ -94,13 +96,19 @@ function hasAliasValidate(
   return hasAnyValidateTag(entry.body, typeAliasesFull);
 }
 
-function collectValidators(tags: Tag[] | undefined): string[] {
+/**
+ * Each validator referenced inside `@validate(...)` is itself an Agency
+ * expression. We don't have a TS IR builder for arbitrary Agency
+ * expressions, so we delegate to `tagArgToTs` (which returns a TS
+ * source string) and wrap with `ts.raw(...)`.
+ */
+function validatorNodes(tags: Tag[] | undefined): TsNode[] {
   if (!tags) return [];
-  const out: string[] = [];
+  const out: TsNode[] = [];
   for (const t of tags) {
     if (t.name !== "validate") continue;
     for (const arg of t.arguments) {
-      out.push(tagArgToTs(arg));
+      out.push(ts.raw(tagArgToTs(arg)));
     }
   }
   return out;
@@ -115,12 +123,24 @@ function isNullableType(t: VariableType): boolean {
   );
 }
 
+/**
+ * Build the (string-built) Zod schema for `t` as a `ts.raw` node.
+ * `mapTypeToValidationSchema` predates the TS IR; rather than rewriting
+ * it, we keep it returning strings and wrap here.
+ */
+function schemaNode(
+  t: VariableType,
+  typeAliases: Record<string, VariableType>,
+): TsNode {
+  return ts.raw(mapTypeToValidationSchema(t, typeAliases));
+}
+
 function descriptor(
   variableType: VariableType,
   typeAliases: Record<string, VariableType>,
   typeAliasesFull: Record<string, TypeAliasEntry>,
   seen: Set<string> = new Set(),
-): string {
+): TsNode {
   // For an alias reference whose alias body has any `@validate(...)` tag,
   // reference the alias's runtime descriptor (attached as
   // `(Alias as any).__agency_descriptor` by `processTypeAlias`) rather than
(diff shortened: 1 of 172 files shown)


packages/agency-lang/lib/backends/typescriptGenerator/validationDescriptor.ts
```diff
@@ -0,0 +1,216 @@
...
+  // not the consumer's. Use-site validators are concatenated on top of the
+  // alias's via a runtime helper.
+  if (variableType.type === "typeAliasVariable") {
+    const entry = typeAliasesFull[variableType.aliasName];
+    const useSiteValidators = collectValidators(variableType.tags);
+    if (entry && hasAliasValidate(entry, typeAliasesFull)) {
+      const ref = `(${variableType.aliasName} as any).__agency_descriptor`;
+      if (useSiteValidators.length === 0) return ref;
+      // Append use-site validators to the alias's chain.
+      return `{ ...${ref}, validators: [...(${ref}?.validators ?? []), ${useSiteValidators.join(", ")}] }`;
```
egonSchiele: Would it make sense to use TSIR builders here?

egonSchiele: Or maybe add some builder functions specifically for JSON schemas, unrelated to TS IR?

packages/agency-lang/lib/runtime/validateChain.ts
```diff
@@ -0,0 +1,209 @@
...
+async function walk(
+  value: unknown,
+  descriptor: TypeValidationDescriptor,
+  ctx: unknown,
+  depth: number,
+  maxDepth: number,
+): Promise<ResultValue> {
+  if (depth > maxDepth) {
+    return failure(
+      `validation recursion depth exceeded (limit ${maxDepth})`,
```
egonSchiele: might be helpful to also return the value and maybe the descriptor, so users can figure out exactly what value caused the recursion depth to be exceeded

packages/agency-lang/lib/typeChecker/mergeTags.ts
```diff
@@ -0,0 +1,111 @@
...
+      const merged = mergeJsonSchemaArgs(all);
+      combined.push({
+        type: "tag",
+        name: "jsonSchema",
+        arguments: merged,
+        loc: all[0].loc,
+      });
+    } else {
+      // Other tags: concat verbatim.
+      for (const t of all) combined.push(t);
```
egonSchiele: other tags shouldn't be handled here, right? They may be for a totally different purpose.

packages/agency-lang/lib/typeChecker/mergeTags.ts
```diff
@@ -0,0 +1,111 @@
...
+  // argument, preserving the order: alias entries first, use-site entries
+  // last. For literal keys present in both, the later (use-site) wins.
+  // Splats are kept verbatim and stay in source order (their runtime
+  // evaluation is what produces the final merged object).
+  type Entry = AgencyObjectKV | SplatExpression;
+  const entries: Entry[] = [];
+  for (const t of tags) {
+    const arg = t.arguments[0];
+    if (!arg || arg.type !== "agencyObject") {
+      // Not a well-formed @jsonSchema(...) — leave the tag's args alone.
```
egonSchiele: should we error in this case?

packages/agency-lang/lib/types/tag.ts
```diff
@@ -1,7 +1,35 @@
...
+import type { Expression } from "../types.js";
 
 export type Tag = BaseNode & {
   type: "tag";
   name: string;
-  arguments: string[];
+  arguments: Expression[];
 };
+
+/**
```
egonSchiele: no need to support backwards compatibility unless not doing so would bloat this pr

packages/agency-lang/stdlib/types.agency
```diff
@@ -0,0 +1,35 @@
...
+
+/** An http:// or https:// URL. */
+@validate(isUrl)
+@jsonSchema({ ...urlFormat })
+export type Url = string
+
+/** A canonical 8-4-4-4-12 hex UUID. */
+@validate(isUuid)
+@jsonSchema({ ...uuidFormat })
+export type Uuid = string
```
egonSchiele: prefer all uppercase

packages/agency-lang/stdlib/validators.agency
```diff
@@ -0,0 +1,54 @@
...
+  def main() {
+    const e: Email! = "foo@example.com"   // validated
+  }
+  ```
+
+  Validators are only run for types annotated with `!` (the bang operator).
+  See the [validation annotations guide](../guide/annotations.md) for details.
+*/
+
+/** Returns success if `value` is a syntactically valid email address. */
```
egonSchiele: Instead of doc comments, please add these as doc strings for the functions, so if they're ever used as tools, the tools will have a description.

PR #175, opened 2026-05-20
feat: validation annotations follow-ups

egonSchiele:
Fast follows for `@validate` / `@jsonSchema` annotations (see #174).

## Track A / D — User-facing docs
- New guide: `docs/site/guide/annotations.md` covering `@validate` and `@jsonSchema` usage, propagation, parameterized validators, and TS function validators.
- VitePress sidebar updated to include **Annotations**.
- `stdlib/validators.agency` doc link now points to the new guide.
- `docs/dev/validation-annotations.md` updated to reference the guide and documents parameterized validators / `__factory`.

## Track B — Missing Agency coverage
New Agency tests under `tests/agency/validation/`:
- `validateOnUnionMembers`
- `validateOnGeneric`
- `jsonSchemaOnResultSuccess`
- `validateRecursiveTypeShallow`

## Track C — Parameterized validators
- Runtime factories added to `lib/stdlib/validators.ts`: `_min`, `_max`, `_minLength`, `_maxLength`, `_matches`.
- Exported from `stdlib/validators.agency` as `min`, `max`, `minLength`, `maxLength`, `matches`.
- Unit tests in `lib/stdlib/__tests__/validators.test.ts`.
- E2E test `tests/agency/validation/validateParameterized`.

## Runtime + codegen plumbing
While wiring up parameterized validators, `min(0)` evaluated before its imported binding was initialized (because `export static const min = _min` went through async `__initializeStatic`). Two fixes:

1. **Lazy factory wrapping** — new `__factory` helper in `lib/runtime/validateChain.ts` (exported from `lib/runtime/index.ts`), imported via TS generator import templates. `lib/backends/typescriptGenerator/validationDescriptor.ts` wraps function-call tag args (`@validate(min(0))`) as `__factory(() => min(0))`.
2. **Eager static aliases of plain JS imports** — `lib/backends/typescriptBuilder/sectionAssembler.ts` pre-scans imports and emits `static const x = importedJsName` as a direct top-level const when the import is from a non-Agency module, so stdlib factory exports are immediately usable.

## Validation
- `pnpm exec tsc --noEmit` ✓
- `pnpm run lint:structure` ✓
- `pnpm test:run` ✓ (247 files / 4147 tests)
- `pnpm run agency test tests/agency/validation -p 12` ✓ (56 / 56)
- `make` ✓ and fixtures regenerated.

Generator fixture and debugger snapshot diffs are due to the added `__factory` import line and are expected.


packages/agency-lang/docs/site/guide/annotations.md
```diff
@@ -0,0 +1,246 @@
+# Type Annotations: `@validate` and `@jsonSchema`
```
egonSchiele: I think "Type Validation" would be a more descriptive name. Also change file name to type-validation.md

packages/agency-lang/stdlib/validators.agency
```diff
@@ -81,3 +81,24 @@ export def isNegative(value: number): Result {
...
+// load (when the descriptor is constructed), so the captured parameters
+// are baked into the resulting validator.
+//
+// Note: these are re-exports of the raw JS factories rather than agency
+// `def` wrappers. They must remain plain callable JS functions because
+// tags compile to direct calls at module-load time (e.g. `min(0)` runs
+// once when the descriptor is built). Wrapping them in AgencyFunction
+// would break this — see the comment in `lib/stdlib/validators.ts`.
+//
+export static const min = _min
```
egonSchiele: I don't think this is the right approach here. I would have expected `min` to be a function so that it can be used in other contexts too. I think having it be a constant that is a TypeScript function is actually pretty unintuitive. I would suggest making it a function, and giving it a docstring to explain what it does.

Then I would suggest letting users use PFA to construct a new function out of min. That function is then what runs when the validators are called. I think that change would also mean you don't need to add all the code for factories, etc.

https://agency-lang.com/guide/partial-application.html

PR #176, opened 2026-05-20
Add value-parameterized type aliases

egonSchiele:
## What

Implements the "dependent types" feature deferred from the [validation annotations spec](docs/superpowers/specs/2026-05-19-type-validation-and-json-schema-annotations-design.md) — value-parameterized type aliases. Users can now write:

```agency
@validate(min.partial(n: low), max.partial(n: high))
@jsonSchema({ minimum: low, maximum: high })
type NumberInRange(low: number, high: number) = number

type User = {
  age: NumberInRange(0, 150)
  score: NumberInRange(1, 100)
}
```

Each use-site instantiates the alias with literal / static-const args that get substituted into the tag expressions at compile time. The substituted tags then flow through the existing `mergeTagSets` / `appendMeta` / descriptor pipeline unchanged.

## Highlights

- **Parser** — value params use `()`, type params use `<>`. Combined form: `type BoundedList<T>(n: number) = T[]`. Type params first.
- **Tag-arg restriction** — bare function calls are now forbidden in `@validate` / `@jsonSchema` args; use PFA syntax (`min.partial(n: 0)`) instead. A shared `staticTagArgParser` enforces the same restriction for value-arg use sites and default expressions.
- **Substitution pass** — new `lib/typeChecker/valueParamSubstitution.ts` with `applyValueArgs` as the canonical entry point. Resolution order is type-param sub first, then value-arg sub. Substitution walks both tags **and the alias body** so wrapper aliases that forward value params (`type EvenInRange(low, high) = NumberInRange(low, high)`) work.
- **Codegen** — schemas + descriptors are inlined at each use site for value-parameterized instantiations. No `__agency_descriptor` side-channel for these; no top-level schema const emitted for the alias.
- **`agency docs`** — renders value params alongside type params.
- **Stdlib** — `NumberInRange`, `StringWithLength`, `MatchesPattern`, `BoundedArray<T>`.
- **Docs** — user guide section, dev internals (`docs/dev/validation-annotations.md`), spec promoted from "future work" to body.

## Plan + spec

- Plan: [`docs/superpowers/plans/2026-05-20-value-parameterized-type-aliases-impl.md`](docs/superpowers/plans/2026-05-20-value-parameterized-type-aliases-impl.md)
- Spec: [`docs/superpowers/specs/2026-05-19-type-validation-and-json-schema-annotations-design.md`](docs/superpowers/specs/2026-05-19-type-validation-and-json-schema-annotations-design.md) (now has the value-parameterized aliases section in the body)

## Testing

- 7 new e2e Agency tests under `tests/agency/validation/`: `valueParamSimple`, `valueParamMultiArg`, `valueParamDefaults`, `valueParamStaticConst`, `valueParamInJsonSchema`, `valueParamWithGeneric`, `valueParamWrappingAlias`.
- 35 new unit tests across substitution, `applyValueArgs`, `resolveType` + valueArgs, jsonSchema/PFA validation, parser, and codegen.
- `pnpm exec tsc --noEmit` — clean
- `pnpm test:run` — **249 files, 4201 tests passing** (was 4168 before)
- `make` — clean (stdlib + docs regenerated)

## Future work (documented in the plan)

Two items intentionally out of scope, captured for follow-up:

1. **Static refinement / dependent-style checking** — today value args are inert from the checker's perspective; two `BoundedArray(0, 100)` and `BoundedArray(0, 50)` are mutually assignable because both bottom out at `T[]`. A future pass could promote value args to first-class participants in assignability with a predicate registry for stdlib validators.

2. **Phantom-type / type-state soundness gap** — confirmed via a small Agency test: `Thread<Anonymous>` is silently accepted where `Thread<Authenticated>` is expected because the type checker unfolds the alias and compares structurally. Reproducer and fix sketch (compare type-arg lists pairwise before falling back to structural body comparison) are in the plan.

## Design decisions worth flagging

- **Value-param aliases are nominally distinct but mutually assignment-compatible** — `NumberInRange(0, 100)` and `NumberInRange(1, 50)` are interchangeable as `number`. Validators are independent and run at `!` sites only. This is the v1 simple-is-better-than-clever stance; refinement subtyping is in future work.
- **No `__agency_descriptor` for instantiations** — there's no single descriptor for a parameterized alias because the descriptor would need to encode specific args. Codegen inlines at each use site.
- **Top-level schema const skipped for parameterized aliases** — there's no useful single Zod schema for `NumberInRange` without args. If a value-parameterized alias is `export`ed, importing the name fails at the TS level (option (a) in the plan); the type-side import still works because use-sites inline the substituted schema.
- **Substitution walks the alias body**, not just its tags. Discovered during G7 (`valueParamWrappingAlias`) — without this, wrapper aliases would silently emit unbound identifiers into generated TS.

### DIFF
--- a/packages/agency-lang/docs/dev/validation-annotations.md
+++ b/packages/agency-lang/docs/dev/validation-annotations.md
@@ -394,3 +394,101 @@ since `.meta()`'s semantics only matter at runtime.
    threads `success.value` into the next validator. Tests in
    `tests/agency/validation/` cover the boundary cases (array element,
    nested object, nullable, transform).
+
+---
+
+## Value parameters and substitution
+
+Value-parameterized type aliases — `type NumberInRange(low: number, high: number) = number`
+— let users lift a tag's numeric (or string) literals out to the alias
+declaration and re-supply them at every use site. Substitution is purely
+compile-time; the runtime walker, descriptor format, and `__validateChain`
+helpers are unchanged.
+
+### The substitution pass
+
+`lib/typeChecker/valueParamSubstitution.ts` is the single source of
+truth. It exports:
+
+- `substituteValueArgsInExpression(expr, bindings)` — walks the
+  restricted tag-arg expression subtree (literals, identifiers, object
+  literals + spreads, `valueAccess` chains incl. PFA method-call args)
+  and replaces any `variableName` whose `value` matches a key in
+  `bindings` with a structural clone of the bound `Expression`. Nodes
+  outside the restricted subset pass through unchanged. The input tree
+  is never mutated.
+- `substituteValueArgsInTag(tag, bindings)` — convenience wrapper that
+  maps the substitution over `tag.arguments`.
+- `substituteValueArgsInType(vt, bindings)` — walks a `VariableType`
+  tree and rewrites any inner `typeAliasVariable` / `genericType` whose
+  own `valueArgs` reference a name in `bindings`. Required so wrapping
+  aliases (`type EvenInRange(low, high) = NumberInRange(low, high)`)
+  can forward their value parameters to inner instantiations.
+- `applyValueArgs(entry, valueArgs, aliasName)` — the canonical entry
+  point: validates arity (filling missing tail args from
+  `valueParams[i].default`, throwing on missing-required or
+  too-many-args), builds the bindings map, then returns a fresh
+  `TypeAliasEntry` whose `tags` and `body` are both substituted. Best-
+  effort literal type-check on each argument; non-literal args are
+  accepted and deferred.
+
+### Where it fires
+
+In `lib/typeChecker/assignability.ts`, `resolveType` calls
+`applyValueArgs` whenever it resolves a `typeAliasVariable` or
+`genericType` whose entry has `valueParams`. Type-param substitution
+happens first (existing path); value-arg substitution runs on the
+result. The substituted entry's tags then flow through the existing
+`attachAliasTags` + `mergeTagSets` chain unchanged.
+
+### The shared `staticTagArgParser`
+
+A single combinator backs every place that needs the restricted
+expression subset:
+
+1. Tag arguments inside `@validate(...)` and `@jsonSchema(...)`.
+2. Default-value expressions for value params (`= 0`).
+3. Use-site value-arg expressions (`Age(18)`).
+
+Centralizing the rule keeps "static consts and value-param identifiers
+are OK; bare function calls are not" defined exactly once. PFA
+(`min.partial(n: low)`) stays allowed because it's a method call on an
+identifier, not a bare function call.
+
+### Codegen divergence
+
+Value-parameterized aliases are deliberately *not* emitted as top-level
+Zod schemas:
+
+- **No top-level `const AliasName = z....` emission.**
+  `typescriptBuilder.ts` skips the const for aliases with
+  `valueParams`. There's no single schema for `NumberInRange` without
+  arguments.
+- **No `(AliasName as any).__agency_descriptor` side-channel.** The
+  descriptor would have to encode the specific value args; instead,
+  the descriptor is constructed *inline at the use site* with the
+  substituted tags.
+- **Use-site inlining.** `typeToZodSchema.ts` and
+  `validationDescriptor.ts` both check for `variableType.valueArgs` on
+  a `typeAliasVariable` / `genericType` reference, call
+  `applyValueArgs`, and recursively map the substituted entry. The
+  substituted tags drive `appendMeta` and `validatorNodes`.
+
+### `tagArgToTs` safety net
+
+After substitution, the TypeScript printer should never see a
+value-param identifier. If it does (i.e. someone added a code path that
+skipped `applyValueArgs`), `tagArgToTs` throws
+`value param 'X' left unsubstituted` so the bug surfaces immediately.
+
+### Files of record (value-parameterized additions)
+
+| File | Role |
+|------|------|
+| `lib/typeChecker/valueParamSubstitution.ts` | The substitution pass + `applyValueArgs` |
+| `lib/typeChecker/valueParamSubstitution.test.ts` | Unit tests for every branch |
+| `lib/typeChecker/assignability.ts` | Calls `applyValueArgs` during `resolveType` |
+| `lib/backends/typescriptGenerator/typeToZodSchema.ts` | Inlines substituted schemas at use sites |
+| `lib/backends/typescriptGenerator/validationDescriptor.ts` | Inlines substituted descriptors at use sites |
+| `lib/backends/typescriptBuilder.ts` | Skips top-level emission for value-parameterized aliases |
+| `stdlib/types.agency` | Pre-baked `NumberInRange`, `StringWithLength`, `MatchesPattern`, `BoundedArray<T>` |
--- a/packages/agency-lang/docs/dev/validation-annotations.md
+++ b/packages/agency-lang/docs/dev/validation-annotations.md
@@ -469,17 +469,70 @@ Zod schemas:
   the descriptor is constructed *inline at the use site* with the
   substituted tags.
 - **Use-site inlining.** `typeToZodSchema.ts` and
-  `validationDescriptor.ts` both check for `variableType.valueArgs` on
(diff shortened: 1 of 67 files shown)


packages/agency-lang/docs/dev/validation-annotations.md
```diff
@@ -394,3 +394,101 @@ since `.meta()`'s semantics only matter at runtime.
...
+Value-parameterized aliases are deliberately *not* emitted as top-level
+Zod schemas:
+
+- **No top-level `const AliasName = z....` emission.**
+  `typescriptBuilder.ts` skips the const for aliases with
+  `valueParams`. There's no single schema for `NumberInRange` without
+  arguments.
+- **No `(AliasName as any).__agency_descriptor` side-channel.** The
+  descriptor would have to encode the specific value args; instead,
+  the descriptor is constructed *inline at the use site* with the
```
egonSchiele, line 469: What are the pros and cons of doing it this way versus creating a function?

`const Age = (value:number) => z....`

packages/agency-lang/lib/backends/typescriptGenerator/validationDescriptor.ts
```diff
@@ -158,6 +166,23 @@ function descriptor(
   if (variableType.type === "typeAliasVariable") {
     const entry = typeAliasesFull[variableType.aliasName];
     const useSiteValidators = validatorNodes(variableType.tags);
+    // Value-parameterized alias instantiation: there is NO
```
egonSchiele, line 172: This distinction is not very clear to me. Does this mean there is a lot of branching logic everywhere? What do we expect the maintenance cost of this additional complexity to be?

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -689,10 +690,33 @@ export const primitiveTypeParser: Parser<PrimitiveType> = trace(
...
-    capture(many1WithJoin(varNameChar), "aliasName"),
-  ),
+  (input: string): ParserResult<TypeAliasVariable> => {
+    const parser = seqC(
+      set("type", "typeAliasVariable"),
+      capture(many1WithJoin(varNameChar), "aliasName"),
+      // Optional `(arg1, arg2, ...)` value-arg suffix for
+      // value-parameterized aliases. Each arg is a statically-
+      // restricted expression (literal, identifier, PFA, object literal).
+      optional(
```
egonSchiele: these parser changes duplicate logic

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -1360,32 +1452,51 @@ export const skillParser = (input: string) => {
...
+//   1. `@validate(...)` / `@jsonSchema(...)` tag arguments
+//   2. value-parameter default expressions (`= 0`)
+//   3. value-arg expressions at use sites (`Age(18)`)
 //
 // Allowed: string / number / boolean / null literals, identifiers,
-// function calls, object literals (including spread). NOT allowed:
-// ternaries, binary ops, pipes, member access, template strings,
-// array literals.
+// object literals (including spread), and PFA expressions
+// (e.g. `min.partial(n: 0)`). NOT allowed: bare function calls
```
egonSchiele: remind me again why function calls, bin ops, and array literals aren't allowed? Also, on "template strings" -- All strings in agency allow string interpolation.

packages/agency-lang/lib/typeChecker/valueParamSubstitution.ts
```diff
@@ -0,0 +1,427 @@
...
+
+/**
+ * Structural deep clone of an Expression. We never share Expression
+ * nodes between an alias declaration and a use-site instantiation
+ * (separate substitutions must not see each other's mutations).
+ *
+ * Implemented via `JSON.parse(JSON.stringify(...))` because the
+ * Expression subset accepted in tag args is plain data: no functions,
+ * no cycles. Suffices for the cases we care about (literals, object
+ * literals, PFA expressions).
```
egonSchiele, line 362: are you sure PFAs would work here?

packages/agency-lang/lib/typeChecker/valueParamSubstitution.ts
```diff
@@ -0,0 +1,427 @@
...
+    case "blockType":
+      return {
+        ...vt,
+        params: vt.params.map((p) => ({
+          ...p,
+          typeAnnotation: substituteValueArgsInType(p.typeAnnotation, bindings),
+        })),
+        returnType: substituteValueArgsInType(vt.returnType, bindings),
+      };
+    default:
```
egonSchiele: In the future, if users add a new kind of type, how will they remember that they need to add it here as well? Is this documented somewhere?

packages/agency-lang/tests/agency/validation/valueParamDefaults.agency
```diff
@@ -0,0 +1,23 @@
...
+
+import { min } from "std::validators"
+
+@validate(min.partial(n: low))
+type Age(low: number = 0) = number
+
+node main() {
+  const adult: Age(18)! = 25
+  const childButOk: Age()! = 5
+  const negative: Age()! = -1
```
egonSchiele: What about an example with no parens? Like

```
  const childButOk: Age! = 5
```

packages/agency-lang/tests/agency/validation/valueParamMultiArg.agency
```diff
@@ -0,0 +1,33 @@
...
+          if (isFailure(tightTooLow)) {
+            if (isFailure(tightTooHigh)) {
+              return "ok"
+            }
+          }
+        }
+      }
+    }
+  }
+  return "failed"
```
egonSchiele: makes it hard to know which one failed

packages/agency-lang/tests/agency/validation/valueParamWrappingAlias.agency
```diff
@@ -0,0 +1,31 @@
...
+// at the inner reference site, which itself is substituted at the
+// outer use site.
+
+import { min, max } from "std::validators"
+
+@validate(min.partial(n: low), max.partial(n: high))
+@jsonSchema({ minimum: low, maximum: high })
+type NumberInRange(low: number, high: number) = number
+
+type EvenInRange(low: number, high: number) = NumberInRange(low, high)
```
egonSchiele, line 13: What about an example where only some of the parameters are forwarded?

packages/agency-lang/tests/agency/validation/valueParamWrappingAlias.agency
```diff
@@ -0,0 +1,31 @@
...
+
+node main() {
+  const s = schema(EvenInRange(0, 100))
+  const js = s.toJSONSchema()
+
+  const okMid: EvenInRange(0, 100)! = 50
```
egonSchiele, line 19: You're missing tests for value parameterized type aliases that are objects or unions. Also what about examples where these are used as types for function args, or as return types?

What about an example where they are used to specify the structured output for an llm?

What about this:

```
  const goodAdult: Age(18)! = 20
  const anyPositive: Age(0)! = 5

const foo = goodAdult + anyPositive
return schema(foo).zodSchema.toJSONSchema()
```

What do we expect this to return?

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -516,6 +517,81 @@ export const simpleStringParser: Parser<StringLiteral> = (input: string) => {
...
+// String literal that allows identifier-only `${name}` interpolation but
+// nothing else. Used by `staticTagArgParser` so e.g.
+// `@jsonSchema({ description: "divisible by ${divisor}" })` is accepted
+// when `divisor` is a value-param name (or a top-level static const),
+// while `"${a + b}"`, `"${foo()}"`, `"${obj.field}"` are rejected by
+// the parser before reaching the type checker / generator.
+export const staticInterpolatedStringParser: Parser<StringLiteral> = (
+  input: string,
+) => {
+  const open = oneOf('"`')(input);
```
egonSchiele: oof, this is... not pretty. see the combinators provided by tarsec (like `quotedString`): https://egonschiele.github.io/tarsec/ also see other parsers in this file for how to write this more cleanly.

PR #180, opened 2026-05-22
Replace callback keyword with scoped callback() stdlib function

egonSchiele:
## Summary

Replaces Agency's `callback` keyword with a `callback(name, fn)` stdlib function whose registrations are scoped to the dynamic extent of the calling function/node. Registrations are stored on stack frames, so they clean up automatically on frame pop and survive interrupts via the existing frame serialization.

Implements the plan at `docs/superpowers/plans/2026-05-20-scoped-callbacks.md`.

## What changed

**Runtime / storage**
- `State.scopedCallbacks` field with `addScopedCallback(name, fn)`, serialized through `toJSON` / `fromJSON`
- `StateStack.callerFrame()` and `StateStack.collectScopedCallbacks(name)`
- `RuntimeContext.topLevelCallbacks` for callbacks registered during `__initializeGlobals` (when no real caller frame exists)
- `_callback` runtime function routes registrations to either the caller frame or `topLevelCallbacks`

**Hook execution**
- `callHook` rewritten to gather callbacks from: scoped (innermost → outermost), then top-level, then TS-passed
- Return values from callbacks are discarded — the message-override capability of `onLLMCallStart` / `onLLMCallEnd` is removed
- Per-instance recursion guard prevents a callback from invoking itself
- Errors are caught and logged; interrupts are re-thrown so enclosing `handle` blocks see them

**Language / backends**
- `callback` keyword removed from the parser; `FunctionDefinition.callback` field removed
- `_registeredCallbacks` and `installRegisteredCallbacks` removed from `RuntimeContext` and from call sites in `node.ts`, `interrupts.ts`, `rewind.ts`
- Generated TypeScript now passes `ctx: __ctx` to `callHook`
- `blockSetup.mustache` updated so top-level callback blocks created in `__initializeGlobals` no longer crash trying to reference `__threads`

**stdlib**
- `std::agency` exports a new public `callback(name, fn)` function that wraps `_callback`

**Tests**
- Migrated existing callback-keyword usage to the new function form (`tests/agency/callback-basic.agency`, `tests/agency/memory/llm-injection.agency`, `tests/agency/builtins/emit.agency`, `lib/agents/policy/agent.agency`)
- New Agency execution tests: `callback-scoped`, `callback-nested`, `callback-toplevel`, `callback-recursion`
- New unit tests: `lib/runtime/hooks.test.ts` (10 tests) and additional cases in `lib/runtime/state/stateStack.test.ts`

**Docs**
- `docs/site/appendix/callbacks.md` rewritten to use the new function form and explain scoping
- `docs/misc/lifecycleHooks.md` "Modifying Messages in LLM Hooks" section removed; `onLLMCallStart` / `onLLMCallEnd` descriptions updated

## Verification

- `pnpm run typecheck` ✓
- `pnpm run lint:structure` ✓
- `pnpm test:run` — all 4347 tests pass
- `pnpm run a test` on each new fixture (`callback-scoped`, `callback-nested`, `callback-toplevel`, `callback-recursion`) ✓
- `make` builds cleanly

### DIFF
--- a/packages/agency-lang/docs/site/stdlib/agency.md
+++ b/packages/agency-lang/docs/site/stdlib/agency.md
@@ -60,7 +60,7 @@ type FilterImportsResult = {
 }
 ```
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/agency.agency#L122))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/agency.agency#L134))
 
 ## Functions
 
@@ -82,6 +82,28 @@ Compile Agency source code. Returns a CompiledProgram on success, or a failure w
 **Returns:** `Result`
 
 ([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/agency.agency#L28))
+
+### callback
+
+```ts
+callback(name: string, fn: any)
+```
+
+Register a scoped callback for the dynamic extent of the calling function or node.
+  When the caller returns, the callback is automatically removed. Top-level callbacks
+  (registered outside any function or node) are active for the whole run.
+
+  @param name - One of the Agency callback hook names (e.g. "onNodeStart", "onFunctionStart", "onLLMCallEnd")
+  @param fn - A function that receives the event data
+
+**Parameters:**
+
+| Name | Type | Default |
+|---|---|---|
+| name | `string` |  |
+| fn | `any` |  |
+
+([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/agency.agency#L36))
 
 ### run
 
@@ -117,7 +139,7 @@ Execute a compiled Agency program in a subprocess. The parent's handler chain ex
 
 **Throws:** `std::run`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/agency.agency#L36))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/agency.agency#L48))
 
 ### runFile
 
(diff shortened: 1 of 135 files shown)


packages/agency-lang/docs/site/stdlib/agency.md
```diff
@@ -83,6 +83,28 @@ Compile Agency source code. Returns a CompiledProgram on success, or a failure w
 
 ([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/agency.agency#L28))
 
+### callback
```
egonSchiele, line 86: not sure this is the right place, actually... most of this file is about manipulating agency *code*. Where else would make sense?

packages/agency-lang/lib/runtime/hooks.ts
```diff
@@ -93,41 +93,74 @@ export function registerGlobalHook<K extends keyof CallbackMap>(
...
+  ctx: RuntimeContext<any>,
+  errorLabel: string,
+): Promise<void> {
+  const key = fn as object;
+  if (_activeCallbacks.has(key)) return;
+  _activeCallbacks.add(key);
+  try {
+    await invokeCallback(fn, data, ctx);
+  } catch (error) {
+    if (isAgencyInterrupt(error)) throw error;
```
egonSchiele: Just to clarify, agency interrupts aren't actually thrown. They are returned. Please see how interrupts are handled in other places. I am not sure that the code you have here is correct.

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -14,6 +14,68 @@ import {
...
+  if (!VALID_CALLBACK_NAME_SET.has(name)) {
+    throw new Error(
+      `Unknown callback '${name}'. Valid: ${VALID_CALLBACK_NAMES.join(", ")}`,
+    );
+  }
+  const ctx = __state.ctx;
+  // Top-level: we're inside __initializeGlobals. The only frame on the stack
+  // is `callback`'s own (or none, defensively). There is no caller frame
+  // that survives past init, so route to ctx.topLevelCallbacks.
+  if (ctx.stateStack.stack.length <= 1) {
```
egonSchiele: This doesn't feel like a very clear way to check that we are on the top level. Why not add a method to the state stack like `isGlobalContext` or something similar?

PR #184, opened 2026-05-22
Phase 1: codegen-emitted callback hook sites via Runner.hook

egonSchiele:
## Summary

Phase 1 of callback interrupts: propagates interrupts raised inside callback bodies for codegen-emitted hook sites (`onFunctionStart`, `onEmit`, `onNodeStart`, `onNodeEnd`). When a callback halts with `Interrupt[]`, the surrounding compiled function/node halts via the new `Runner.hook` specialized step type and returns the interrupts up the stack so `respondToInterrupts` works end-to-end.

Builds on Phase 0 (`callHook` returns `Interrupt[] | undefined`, `callHookAndDrop` exists for sites that can't propagate).

## What's in this PR

1. **`Runner.hook(id, hookName, data)`** — new specialized step type in `lib/runtime/runner.ts` that mirrors the `Runner.handle` / `Runner.thread` / `Runner.pipe` skeleton (substep counter, debug hook, coverage). On callback interrupt it halts the runner without advancing the counter, so on resume the hook re-fires and the callback frame is restored via its own checkpoint.

2. **`ts.runnerHook` IR node + builder + prettyPrint** — codegen emits a single `await runner.hook(<id>, "hookName", <data>);` line per hook site.

3. **Codegen migration for 4 of the 5 hook sites** in `lib/backends/typescriptBuilder.ts`:
   - `onFunctionStart` (substep id 0 in the function try-body)
   - `onEmit` (current step id from the body-parts loop)
   - `onNodeStart` (substep id 0 in the node try-body) — Runner now constructed BEFORE the hook so it can fire as a resumable substep
   - `onNodeEnd` (substep id N+1, success-path only)

   `onFunctionEnd` is intentionally **NOT migrated** — it fires from a `finally` block, and interrupts thrown from `finally` cannot be propagated cleanly. A comment in `buildFunctionBody` documents this limitation. Moving it out of `finally` is a follow-up that requires changing function-return lowering away from immediately-halt semantics.

4. **`onAgentStart` / `onAgentEnd` reject interrupts** in `lib/runtime/hooks.ts`. These two hooks fire outside any agency frame — no Runner to halt, no checkpoint to stamp, nowhere for the user to respond from. `callHook` now throws loudly if a callback for these hooks raises interrupts, instead of silently dropping them via `callHookAndDrop`.

5. **End-to-end resume test** `tests/agency/callback-interrupt-resume-onfunctionstart.{agency,test.json}` covering the full interrupt→respond→resume cycle for a callback-raised interrupt.

6. **Unrelated build fix** in `lib/preprocessors/typescriptPreprocessor.core.test.ts` (pre-existing on `main`): `walkNodes()` returns a `Generator` which has no `.map`, so use the existing `walkNodesArray` wrapper. Included to unblock `make build`.

## Important: Runner.hook does NOT stamp its own checkpoint

The first iteration of `Runner.hook` stamped a pinned checkpoint via `createPinned` and overwrote `intr.checkpointId` on every returned interrupt. That turned out to be both unused and incorrect:

- `respondToInterrupts` reads `intr.checkpoint` first, falling back to `ctx.checkpoints.get(intr.checkpointId)` only if `intr.checkpoint` is undefined. We never overwrote `intr.checkpoint`, so the callback's own checkpoint (stamped inside its interrupt step before its `finally` popped the frame) is what gets used on resume.
- The callback's own checkpoint is the one we actually want: it captures the full stack including the callback frame with its substep counters and saved `__interruptId_N`. On resume the callback's `setupFunction` reuses the saved frame from the deserialize queue, the body's increment step is skipped (counter already past), and only the interrupt step re-runs — finding the user's response without re-running earlier side effects.

So `Runner.hook` is intentionally minimal: it only halts the runner. The unit test was updated accordingly.

## Semantic worth knowing

A callback that raises an interrupt fires **exactly once total**: substep counters inside the callback frame survive across the resume via the callback's own checkpoint. The `calls = calls + 1` line in the example test does NOT re-run on resume.

## Validation

- `npx tsc` ✅
- 38 `Runner` unit tests pass
- 31 `hooks` unit tests pass
- All 18 `tests/agency/callback-*.agency` tests pass (17 pre-existing + 1 new resume test)
- Full agency suite not run locally per workflow; CI covers it.

## Not in this PR (deferred / out of scope)

- Resume tests for `onNodeEnd`, `onEmit`, multi-callback batching — easy to add once you confirm the approach.
- `make fixtures` regen — pending.
- `onFunctionEnd` migration.
- LLM/tool hook propagation — covered by Phase 2 (the `runPrompt` work).
- Removing `callHookAndDrop` from agent-start/end sites (those fundamentally can't resume).


egonSchiele:
## Phase 1 follow-up work pushed

Three new commits on top of the initial PR:

- `37235613` — end-to-end resume test for callback interrupt at `emit()` site (`onEmit`)
- `301c9afc` — docs: `docs/dev/callback-hooks.md` and `docs/dev/interrupts.md` updated to describe `ts.runnerHook` / `Runner.hook`, resume semantics, and known limitations
- `7723d071` — regenerated codegen fixtures (84 files) for the `runner.hook` migration

### Corrected resume semantics

Earlier in this PR I had `Runner.hook` stamp/overwrite a pinned checkpoint to drive resume. After re-reading `docs/dev/interrupts.md`, `docs/dev/concurrent-interrupts.md`, the execution model docs, and the state stack code, the correct model is:

- The callback's **own interrupt step** stamps the checkpoint **before** its frame is popped.
- `respondToInterrupts` uses `interrupt.checkpoint` first, so the callback-stamped checkpoint drives resume.
- The callback frame and substep counters survive resume; already-completed substeps inside the callback are not re-run.
- On resume, only the interrupt substep re-fires and consumes the user response.
- Net effect: the callback fires **exactly once total** across an interrupt/resume cycle.

`Runner.hook` no longer stamps its own checkpoint; it just halts with the returned `Interrupt[]` and lets the callback's checkpoint drive resume. Tests were updated accordingly.

### Known limitations (documented in `docs/dev/callback-hooks.md`)

1. **`onFunctionEnd` is not migrated to `runner.hook`.** It still emits raw `callHook` in the generated `finally`. Returning interrupts from `onFunctionEnd` is not supported in this PR.
2. **`onNodeEnd` only fires for void-returning nodes.** For nodes with an explicit `return`, the node body halts the runner before the generated `onNodeEnd` hook site is reached. This is pre-existing behavior, not introduced by this PR.
3. **Multiple interrupting callbacks on the same hook are not resumable.** When two callbacks on the same hook both throw interrupts, each interrupt carries its own checkpoint capturing its own callback frame. `respondToInterrupts` resumes from `interrupts[0].checkpoint`, so only the first callback frame is restored; on resume the second callback is invoked fresh and throws a new interrupt. This is analogous to the concurrent-interrupts/branching problem and is out of scope for Phase 1.
4. **`onAgentStart` / `onAgentEnd` reject interrupts** (validated in codegen). They run outside any runner step and have no resume semantics.

### Validation

- `make build` ✅
- `pnpm test:run lib/runtime/runner.test.ts lib/runtime/hooks.test.ts` ✅ (69 tests)
- `node ./dist/scripts/agency.js test tests/agency/callback-interrupt-resume-onfunctionstart.agency` ✅
- `node ./dist/scripts/agency.js test tests/agency/callback-interrupt-resume-onemit.agency` ✅
- Full agency suite deferred to CI.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -1540,12 +1533,33 @@ export class TypeScriptBuilder {
...
+        hookName: "onFunctionStart",
+        data: {
+          functionName: ts.str(functionName),
+          args: ts.obj(argsObj),
+          isBuiltin: ts.bool(false),
+          moduleId: ts.str(this.moduleId),
+        },
+      }),
+      ts.raw(
+        "if (runner.halted) { if (isFailure(runner.haltResult)) { runner.haltResult.retryable = runner.haltResult.retryable && __self.__retryable; } return runner.haltResult; }",
```
egonSchiele: what does this do? only tool calls are retryable.

packages/agency-lang/lib/runtime/hooks.ts
```diff
@@ -190,6 +190,24 @@ export async function callHook<K extends keyof CallbackMap>(args: {
...
+  // Defensive: onAgentStart / onAgentEnd fire outside any agency frame so
+  // there is no Runner to halt and nowhere for the user to respond from.
+  // Reject loudly so a developer who accidentally registers an
+  // interrupt-raising callback for these hooks sees a clear error
+  // instead of silent failure (callHookAndDrop would log + drop them).
+  if (
+    collected.length > 0 &&
+    (name === "onAgentStart" || name === "onAgentEnd")
+  ) {
+    throw new Error(
```
egonSchiele, line 202: could we also catch this through the typechecker

packages/agency-lang/tests/agency/callback-interrupt-resume-onemit.agency
```diff
@@ -0,0 +1,27 @@
...
+// with the callback frame's substep counters intact. The hook
+// re-fires on the emit substep, the callback's setupFunction reuses
+// the saved callback frame, the increment step is already past, and
+// only the interrupt step re-runs — finding the user's response.
+// The hook returns no interrupts the second time, the emit substep
+// advances, and the rest of main runs.
+callback("onEmit") as _data {
+  cbRuns = cbRuns + 1
+  if (cbRuns == 1) {
+    const _y = interrupt("pause")
```
egonSchiele: test that code before interrupt doesn't re-fire. Test that code after interrupt in the callback fires on resume.

packages/agency-lang/tests/agency/callback-interrupt-resume-onfunctionstart.agency
```diff
@@ -0,0 +1,36 @@
...
+// without re-throwing. doWork runs, main returns.
+//
+// Expected post-resume value: calls == 1 (the callback's increment
+// substep was already past at checkpoint time, so it does NOT re-run
+// on resume).
+callback("onFunctionStart") as data {
+  if (data.functionName == "doWork") {
+    calls = calls + 1
+    if (calls == 1) {
+      const _y = interrupt("pause")
```
egonSchiele: what's the point of this assignment

PR #186, opened 2026-05-23
feat: runBatch concurrent-interrupt primitive + fork/race/runPrompt migrations

egonSchiele:
## Summary

Introduces a single `runBatch` runtime primitive that owns the concurrent-interrupt orchestration that today every site hand-rolls (Promise.allSettled/race, abort composition, shared checkpoint stamping + `intr.checkpoint` overwrite, branch lifecycle). Migrates the three existing call sites — `Runner.runForkAll`, `Runner.runRace` (+ deletes `resumeRaceWinner`), and `runPrompt`'s tool loop — to use the primitive.

This is **Tasks 1–4** of the runBatch plan at [`docs/superpowers/plans/2026-05-22-runbatch-concurrent-interrupt-primitive.md`](docs/superpowers/plans/2026-05-22-runbatch-concurrent-interrupt-primitive.md). Tasks 5–9 (callback slice-rule fixes, multi-callback hook resume, `_activeCallbacks` per-stack guard, docs, supersede markers) are not in this PR.

## What `runBatch` owns

- per-child `getOrCreateBranch` + cached-result short-circuit
- per-child `AbortController` composition with the parent stack's signal
- ALS-isolated `runInBranchContext` invocation
- three modes:
  - `"all"` — `Promise.allSettled`, every child runs concurrently
  - `"sequential"` — `for…of`, strict ordering (today's `callHook` semantics; lets Task 6 preserve them)
  - `"race"` — `Promise.race` + abort losers, with **resume dispatch folded in** (re-runs only the recorded winner when `__race_winner_<id>` is persisted)
- collect/settle, `setInterruptOnBranch` with the leaf-stamped checkpoint, shared batch-level checkpoint stamp, and `intr.checkpoint`/`checkpointId` overwrite (intentional per c72b9c1574 — all interrupts in a batch deliberately share one resume point)
- cost seed/propagate hooks (asymmetric pair for race: `propagateLoserCost` eagerly, `propagateWinnerCost` deferred)
- defensive guards: duplicate-key check, mode-flip mismatch assert

## What `runBatch` deliberately does NOT touch

Per commit c72b9c1574 (which removed the buggy `isForked` approach that broke nested-fork composition):

- the leaf `interruptReturn` template still stamps a per-leaf checkpoint exactly as today; `runBatch` just reads it off each surfaced `Interrupt` and stores it on `BranchState.checkpoint`
- handler (`handle` block) bookkeeping; per-branch handler chains stay as today
- the `__race_winner_<id>` key shape (Task 3 keeps it for in-flight-checkpoint compatibility)

## Commits in this PR

1. **feat: add runBatch primitive** — `lib/runtime/runBatch.ts` + 19 unit tests in `lib/runtime/runBatch.test.ts`. No migrations yet.
2. **refactor: runForkAll uses runBatch primitive** — ~110 lines of hand-rolled `Promise.allSettled` plumbing collapses into a thin adapter.
3. **refactor: runRace uses runBatch primitive (deletes resumeRaceWinner)** — ~280 lines (`runRace` + `buildRaceBranchPromise` + `abortLoserBranchesAndEmitEnds` + `resumeRaceWinner`) collapses into one adapter. `Runner.fork` no longer dispatches between first-run and resume — the adapter handles it internally.
4. **refactor: runPrompt tool loop uses runBatch primitive** — `PromptRunner.parallel` becomes a thin adapter; gains a `keyFor` arg so it and the branchFn body allocate the same branch; returns `RunBatchResult` instead of throwing `PromptBailout`. Adds `recordBranchOutcomes: false` option to `runBatch` so the tool-loop body can keep managing branch state itself (otherwise `runBatch`'s `setResultOnBranch(key, undefined)` would destroy the real tool result).

## Net code change

- `lib/runtime/runner.ts`: **−392 lines** (-110 from fork, −280 from race + helpers, +adapters)
- `lib/runtime/promptRunner.ts`: cleaner, gains tagged-union return
- `lib/runtime/prompt.ts`: minor diff (call-site pattern-match instead of try/catch on `PromptBailout`)
- `lib/runtime/runBatch.ts`: new ~510 lines including doc comments
- `lib/runtime/runBatch.test.ts`: new ~530 lines, 19 unit tests

## Tests

- `lib/runtime/runBatch.test.ts` (19 tests) — pure-JS unit coverage of all three modes, cached short-circuit, abort propagation, race resume dispatch (pending + cached winner), cost-hook asymmetry, empty/duplicate/rejection edge cases, mode-flip defensive assert.
- `lib/runtime/promptRunner.test.ts` (20 tests) — updated to use real `State`/`StateStack` and the new `keyFor` arg, asserts against `RunBatchResult` instead of catching `PromptBailout`.
- `pnpm test:run` — full unit suite (4441 tests, 258 files) passes locally after every commit.
- `pnpm run agency test tests/agency/fork` — all 67 tests pass (covers `fork-multi-interrupt`, `fork-multi-cycle-interrupt`, `fork-nested-interrupt`, `nested/three-levels-deep`, `nested/mixed-completion-nested`, `fork-llm-tool-nested`, `fork-llm-deep-loop`, `fork-stress`, all of `tests/agency/fork/race/*`, all of `tests/agency/fork/llm-tools/*`).
- Full `tests/agency` regression sweep deferred to CI per the project's "don't run the full suite locally — it's slow" guideline.

## Constraints honored

- Everything serializable. The only live-only field on `BranchState` is still `abortController`.
- Handlers (`handle` blocks) — not touched. Per-branch handler chains stay as today.
- Per-branch cost/token propagation preserved (`seedBranchCost` / `propagateBranchCost` hooks).
- Race cost-propagation asymmetry preserved: losers eagerly at race time, winner deferred until winner branch finally pops in a no-interrupt resume.
- `__race_winner_<id>` key shape unchanged — in-flight serialized checkpoints stay compatible.
- `runBatch.invoke` no-throw contract documented at the top of `runBatch.ts`; the only `Interrupt[]`-bearing throw in the runPrompt path (`PromptBailout` from `PromptRunner.parallel`) is converted to a return in Task 4. The audit findings are recorded in the file header.

- Task 5 (thread `branchStack` through `callHook`)
- Task 6 (multi-callback hook resume via `runBatch({ mode: "sequential" })`)
- Task 7 (per-stack `_activeCallbacks` guard)
- Tasks 8 & 9 (docs + supersede plan markers)

### DIFF
--- a/packages/agency-lang/docs/dev/concurrent-interrupts.md
+++ b/packages/agency-lang/docs/dev/concurrent-interrupts.md
@@ -1,34 +1,47 @@
 # Concurrent Interrupts
 
-This doc covers how Agency handles multiple interrupts that arise from concurrent execution paths. There are two user-facing constructs that produce concurrent interrupts:
+This doc covers how Agency handles multiple interrupts that arise from concurrent execution paths.
+
+There are three user-facing constructs that produce concurrent execution:
 
 - **`fork(items) as item { ... }`** — runs the block in parallel for each item. Each item is a separate "branch." If multiple branches interrupt, all interrupts are batched together and returned to the caller in a single `Interrupt[]`.
-- **`llm(prompt, { tools })`** — when the LLM calls multiple tools in a single response, those tool invocations run as parallel branches inside `runPrompt`. If multiple tools interrupt, those interrupts are also batched.
-
-A third construct — **`race(items) as item { ... }`** — also creates parallel branches, but only one wins; interrupts from losers are cancelled rather than batched. See the race section below.
+- **`race(items) as item { ... }`** — same shape as fork, but only one branch wins; interrupts from losers are cancelled.
+- **`llm(prompt, { tools })`** — when the LLM calls multiple tools in a single response, those tool invocations run as parallel branches inside `runPrompt`. If multiple tools interrupt, those interrupts are batched too.
 
 The user responds to a batch via `respondToInterrupts(interrupts, responses)`. Resume reconstructs the full execution state, replays only the work that hadn't completed, and continues.
+
+**As of the runBatch refactor (PR #186):** all three of those call sites delegate to a single runtime primitive, [`runBatch`](../../lib/runtime/runBatch.ts), that owns the concurrent-interrupt orchestration. Previously every site hand-rolled the same `Promise.allSettled` / `Promise.race` boilerplate, made the same subtle mistakes, and was hard to keep in lock-step. After the refactor there is **one** place where branch lifecycle, abort composition, checkpoint stamping, and `intr.checkpoint` overwrite live, and three thin adapter functions that wire it up.
+
+The rest of this document is split into:
+
+1. The shared mental model and data types (unchanged by the refactor).
+2. The `runBatch` primitive itself.
+3. **Implementation details**: how each call site uses `runBatch`.
+4. **Usage guide**: how to add a new concurrent-interrupt site without re-introducing the same bugs.
+5. The slice-rule, abort signals, resume mechanics, invariants — preserved from the pre-refactor doc because they are still load-bearing.
+
+---
 
 ## Quick mental model
 
 Think of each parallel branch as having its own miniature state stack. The parent frame holds a map of branches under its `branches` field. The runtime serialises this tree on interrupt, restores it on resume, and tracks which branches have completed (cached result) versus interrupted (saved checkpoint) versus never-finished (just a stack reference).
 
-```
-main_frame
-├─ branches:
-│  ├─ "fork_1_0": BranchState { stack, interruptId?, result?, abortController? }
-│  ├─ "fork_1_1": BranchState { ... }
-│  └─ "fork_1_2": BranchState { ... }
+```diagram
+╭─────────────────────────╮
+│ main_frame              │
+│   branches:             │
+│   ├─ "fork_1_0": ─┐     │      ╭───────────────────────────╮
+│   ├─ "fork_1_1": ─┼────▶│ BranchState                       │
+│   └─ "fork_1_2": ─┘     │ ├ stack          (own StateStack) │
+╰─────────────────────────╯ ├ interruptId?  (set if halted)  │
+                            │ interruptData?                  │
+                            │ checkpoint?    (leaf cp)        │
+                            │ result?        ({ result })     │
+                            │ abortController? (live-only)    │
+                            ╰───────────────────────────────────╯
 ```
 
-Same shape applies to tool-call branches inside `runPrompt`, just keyed differently:
-
-```
-runPrompt_frame
-├─ branches:
-│  ├─ "tool_call_abc123": BranchState { ... }
-│  └─ "tool_call_def456": BranchState { ... }
-```
+Same shape applies to tool-call branches inside `runPrompt`, just keyed differently (`tool_<toolCallId>`).
 
 ## Core data types
 
@@ -39,7 +52,7 @@ export type BranchState = {
   stack: StateStack;            // this branch's own isolated stack
   interruptId?: string;         // set if this branch is paused on an interrupt
   interruptData?: any;
-  checkpoint?: Checkpoint;      // the inner checkpoint this branch produced when it interrupted
+  checkpoint?: Checkpoint;      // the LEAF checkpoint this branch produced when it interrupted
   result?: { result: any };     // set if this branch completed normally
   abortController?: AbortController;  // live, not serialised — used by race
 };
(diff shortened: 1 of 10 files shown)


packages/agency-lang/docs/dev/concurrent-interrupts.md
```diff
@@ -39,184 +52,409 @@ export type BranchState = {
...
 
-1. Build a `branchKey` from the step path and item index, e.g. `fork_1.2_0`.
-2. Call `frame.getOrCreateBranch(branchKey)` — returns the existing branch on resume, or creates a fresh `BranchState` with an empty `StateStack` on first run.
-3. **If `existing.result !== undefined`** → this branch finished in a previous cycle. Resolve immediately with the cached value. Don't call `blockFn`.
-4. **Otherwise** → create an `AbortController` (composed with parent's signal for nested fork support, see below), invoke `blockFn(item, i, existing.stack)`. The block receives the branch's stack so its `setupFunction` pushes frames onto it, isolated from siblings.
+### Signature
 
-`Promise.allSettled` waits for every branch (resolved value or interrupt array). The runner walks the settled list:
+```ts
+export async function runBatch<T>(opts: RunBatchOpts<T>): Promise<RunBatchResult<T>>;
```
egonSchiele, line 79: Adding the full code here seems unnecessary and feels like something you'll have to keep in sync with the actual code. Why not just add a small relevant snippet in here and point readers to the actual code file for the rest?

packages/agency-lang/docs/dev/concurrent-interrupts.md
```diff
@@ -39,184 +52,409 @@ export type BranchState = {
...
 
-### Multi-cycle resumes
+### Inherited invariants
 
-Branches can interrupt across multiple respond/resume cycles. Common pattern: thread A interrupts and gets approved, then later interrupts again. Meanwhile sibling threads B and C may have completed in cycle 1 and shouldn't replay.
+These behaviours come from the pre-refactor code; `runBatch` preserves them exactly.
 
-The mechanism: `BranchState.result` and `BranchState.interruptId` survive across cycles via `BranchStateJSON`. On any resume:
+- **Errors win over interrupts.** If any child rejects (throws a non-interrupt error), `runBatch` rethrows that error and abandons any interrupts that sibling branches successfully halted with. Matches today's `runForkAll` / `runRace`. Callers that need both must catch inside their `invoke`.
```
egonSchiele: This doesn't make a lot of sense because agency doesn't have exceptions. It has failures: https://agency-lang.com/guide/error-handling.html. I think the only place this would make sense is if an exception is being thrown that is somehow internal to the agency runtime itself. Otherwise, if it's an exception being thrown by some type script code that got called in the branch, what we do in other places like function calls is the entire invocation gets wrapped in a try catch and exceptions get converted into failures. I would check out the code for how a function is defined and do the same kind of wrapping but perhaps excluding the same exceptions that the function definitions exclude.

packages/agency-lang/docs/dev/concurrent-interrupts.md
```diff
@@ -39,184 +52,409 @@ export type BranchState = {
...
+- **Race cost asymmetry.** Losers' cost propagates eagerly at race time (so their partial LLM spend is billed). The winner's cost propagates only when the winner's branch finally completes via a no-interrupt resume.
```
egonSchiele: What does this mean? I would have expected that the cost for the losers gets propagated when their threads are aborted. Is that not the case?

packages/agency-lang/docs/dev/concurrent-interrupts.md
```diff
@@ -39,184 +52,409 @@ export type BranchState = {
...
+- Hooks: `seedBranchCost`, **asymmetric** `propagateLoserCost` / `propagateWinnerCost` (both delegating to the same `propagateBranchCost` helper; only the timing differs). `onBranchEnd → forkBranchEnd` (`"aborted"` for losers, `"interrupted"`/`"success"` for winner, `"failure"` for the first rejecting branch). `onCheckpoint → checkpointCreated` with `reason: "race"`.
+- Returns: `result.kind === "interrupts" ? result.interrupts : result.values[0]` (race always has exactly one winner).
 
-If the winner interrupts a second time after being approved, `resumeRaceWinner` is called again. It still sees the same `winnerIndex`, finds the branch still pending (no `result`), re-invokes the block. The inner code advances past the first interrupt site (cached in frame locals), hits the second interrupt site, mints a new `interruptId`, halts. We stamp another fork-level checkpoint and return.
+What that replaces: ~280 lines (`runRace` + `buildRaceBranchPromise` + `abortLoserBranchesAndEmitEnds` + `resumeRaceWinner`) plus the dispatch in `Runner.fork`. `Runner.fork` no longer needs to distinguish "first run" from "resume winner" — `runBatch`'s mode-`"race"` resume-dispatch path handles both.
```
egonSchiele: These lines about what gets replaced may be useful in a PR description, but I don't think they are useful in a document like this that is meant to be a guide for future software engineers. I don't think they are going to care how many lines of code you replaced in your refactor.

packages/agency-lang/docs/dev/concurrent-interrupts.md
```diff
@@ -39,184 +52,409 @@ export type BranchState = {
...
+```
+
+Use the tagged union. Do not do `if (Array.isArray(result) && result[0]?.type === "interrupt") ...` — that was the pre-refactor pattern and it was easy to get wrong with mixed shapes.
+
+### 8. Test the four cycles, not just the happy path
+
+Any new site needs at least these tests:
+
+1. All children succeed → returns `kind: "values"`.
+2. One child interrupts, others succeed → returns `kind: "interrupts"`; on resume after `respondToInterrupts`, the previously-successful children short-circuit (their `setResultOnBranch` survived), the previously-interrupted child runs once more and succeeds; final result matches the all-succeed case.
```
egonSchiele, line 312: Probably also want a test case where multiple children interrupt, right?

packages/agency-lang/docs/dev/concurrent-interrupts.md
```diff
@@ -299,9 +551,15 @@ If you change anything in this area, make sure these still hold:
...
+- **The runPrompt tool loop's `recordBranchOutcomes: false` arrangement is the only current site that opts out of `runBatch`'s branch-state recording.** If a future adopter needs the same arrangement, follow the same pattern (and re-read the comment in `runBatch.ts` explaining why the cached-branch short-circuit is also disabled in that mode).
+- **Subprocess interrupts (Plan 3, `docs/superpowers/plans/2026-05-10-subprocess-propagation-and-resume.md`) have the same _logical_ shape as a single-child `runBatch`** but the `invoke` crosses an IPC boundary (JSON serialisation, separate process lifetime, separate checkpoint store). Modeling subprocess as `runBatch({ children: [{ key: "sub", invoke: ipcCall }] })` is correct in shape but the IPC layer still needs to (a) reconstruct serialised checkpoints, (b) translate the subprocess's internal `Interrupt[]` JSON into runtime `Interrupt` objects with `.checkpoint` reconstructed via `Checkpoint.fromJSON`, (c) handle subprocess lifecycle. `runBatch` provides the surrounding shape; the IPC layer is separate work.
```
egonSchiele: Again, this may be useful as part of a PR description, but does not belong in a long-running doc.

PR #187, opened 2026-05-23
runBatch Tasks 5-8: per-tool callback slice-rule + multi-callback hook resume + ALS recursion guard

egonSchiele:
Implements Tasks 5–8 of the runBatch concurrent-interrupt primitive plan, extending PR #186's per-call-site `runBatch` migration to cover the per-tool callback firings and the multi-callback hook resume case.

### Task 5 — Thread `branchStack` through `callHook` (slice-rule fix)

The per-tool `onToolCallStart` / `onToolCallEnd` callbacks in `runPrompt` fired on `ctx.stateStack` rather than the tool branch's own stack. Any interrupt raised inside such a callback captured a checkpoint with the wrong slice, silently corrupting the slice-rule discipline that `runBatch` enforces everywhere else.

- Add `invokeCallbacks(...)` in `hooks.ts` accepting an optional `stateStack` override; `callHook` becomes a thin wrapper that omits it (today's top-level behaviour).
- `prompt.ts` per-tool `.start` / `.end` `b.step`s now use `invokeCallbacks({ ..., stateStack: branchStack })`.
- Eight new fixtures under `tests/agency/fork/llm-tools/multi-tool-callback-interrupts-*` exercise the parallel callback-interrupt batch/resume across both/one/start/end/mixed/three/reject/handler-caught/asymmetric-failure shapes.

### Task 6 — Multi-callback hook resume via runBatch

Before this PR, `Runner.hook` called `callHook` directly. Two or more callbacks raising interrupts on the same hook would have ONE of them resume correctly (via the leaf checkpoint) but the others re-execute from the top on every resume cycle, breaking side-effect-once semantics.

- `Runner.hook` now delegates each gathered callback to its own `runBatch` child (`mode: "sequential"`). The shared batch checkpoint preserves all interrupt branches; cached-branch short-circuit + per-leaf checkpoint together give each callback its own resume point.
- `gatherCallbacks` takes an explicit `stack` parameter so scoped-callback discovery walks the branch's own stack when fired inside a fork branch (latent bug fix).
- New fixture `tests/agency/callback-multi-interrupt-resume.agency` proves each pre- and post-interrupt counter increments exactly once across resume with two top-level callbacks.

### Task 7 — Per-async-context recursion guard

With Task 5 / Task 6 firing the same callback across parallel branches, the module-level `_activeCallbacks` WeakSet was incorrectly suppressing legitimate sibling invocations. A per-stack WeakSet doesn't help because each per-callback `runBatch` allocates a new branch stack, so recursive fires don't see the outer entry.

- `fireWithGuard` now uses an `AsyncLocalStorage`-scoped Set. The "currently firing" set is inherited through `await` boundaries and nested sync calls — but a fork branch with its own ALS context (via `runInBranchContext`) starts with the inherited copy and any sibling branches each get their own fresh copy. Recursion is still detected (`tests/agency/callback-recursion` passes); parallel branches no longer drop each other.

### Task 8 — Documentation

- New `docs/dev/runBatch.md` (~100 lines): signature, three modes (sequential is required for callback hooks — `all` would silently race ordered side effects), slice rule, invoke no-throw contract, defensive guards, `recordBranchOutcomes`, subprocess-shape future note.
- `docs/dev/concurrent-interrupts.md` lightly updated to point at the new doc and add `Runner.hook` as the fourth call-site adopter (PR-#186 trimming preserved).
- `docs/dev/callback-hooks.md` "multiple callbacks on the same hook" section rewritten for the per-callback runBatch model; new sections for parallel-branch callbacks and the ALS recursion guard.
- `docs/site/appendix/callbacks.md`: per-hook table updated — `onNodeStart` / `onNodeEnd` / `onFunctionStart` / `onEmit` move from "⚠️ Batched, but resume is partial" to "✅ Batched". `onFunctionEnd` unchanged (still raw `callHook` in `finally`). Added a "Checkpoint format compatibility" note (no migration shim per project's "breaking changes acceptable" stance).
- `CLAUDE.md` adds `runBatch.md` to the deeper-docs list.

## Notable deviations from the original plan

1. **Slice-rule audit assertion dropped.** Plan Task 6 Step 2 specified `frame.hasBranches() && !this.stack` defensive check in `Runner.hook`. It false-positives on resume because the hook's own per-callback `runBatch` leaves branches on the frame until the hook completes. A more precise check would couple to the `hook_*` key scheme; left as a comment in `runner.ts` instead.
2. **Recursion guard uses AsyncLocalStorage, not per-stack WeakSet.** The plan permitted "skip the guard when stateStack is set" as a fallback. That broke `callback-recursion` because the recursive fire goes through a new branch stack. ALS naturally inherits the set through nested calls while letting sibling branches start fresh.

## Validation

- All 587 `lib/runtime/*` vitest tests pass.
- All 20 `tests/agency/callback*.agency` tests pass (including `callback-recursion` and the new `callback-multi-interrupt-resume`).
- All 6 `tests/agency-js/callback-interrupts-*` tests pass.
- Task 5's eight LLM-tool fixtures committed with `retry: 2`; LLM-flaky on first attempt locally but expected to settle on CI.

## Constraints honoured (from PR #186 review)

- The leaf checkpoint mechanism (`interruptReturn` template, leaf stamping in `interruptWithHandlers`) is untouched. The `isForked`-bypass approach removed in `c72b9c1574` is NOT reintroduced.
- `intr.checkpoint` / `intr.checkpointId` overwrite in `runBatch` is intentional and preserved.
- `__race_winner_<id>` key shape unchanged.
- Per-branch handler chains (`handle` blocks) untouched.
- `runBatch.invoke` no-throw contract enforced (callbacks return `Interrupt[]`, never throw).
- All new state is serialisable; only `AbortController` and the live-only ALS Set are non-serialised, matching today's `BranchState.abortController` pattern.

### DIFF
--- a/packages/agency-lang/docs/site/appendix/callbacks.md
+++ b/packages/agency-lang/docs/site/appendix/callbacks.md
@@ -81,21 +81,6 @@ Callbacks fire in declared order. Sequential mode is required, not a
 preference — using `mode: "all"` would silently turn ordered side
 effects into a concurrent race.
 
-#### Checkpoint format compatibility
-
-Callback-hook checkpoints written by versions BEFORE the per-callback
-`runBatch` migration (Task 6 of the runBatch concurrent-interrupt
-plan) are **NOT forward-compatible**. The previous format stored the
-hook's state under the runner's substep counter only; the new format
-adds per-callback branch state on the hook's frame. Resuming an
-old-format checkpoint will fail to find the per-callback branches and
-mis-attribute interrupt responses.
-
-No migration shim is provided — consistent with the project's small
-user base and "breaking changes acceptable across versions" stance.
-Run any pending interrupt cycles to completion on the old version
-before upgrading.
-
 ### Concurrent firing across fork branches
 
 A separate sense of "concurrent" is when the *same hook* fires from
--- a/packages/agency-lang/docs/site/appendix/callbacks.md
+++ b/packages/agency-lang/docs/site/appendix/callbacks.md
@@ -53,11 +53,11 @@ The table below summarizes the current state per hook. Three columns:
 |-------------------|------------------------------|------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `onAgentStart`    | ❌ Compile + runtime error    | N/A                                | Fires outside any agency frame — there is no stack to checkpoint and nowhere to resume from. The typechecker errors at build; `callHook` also throws at runtime as a defense-in-depth.                                                         |
 | `onAgentEnd`      | ❌ Compile + runtime error    | N/A                                | Same as `onAgentStart`.                                                                                                                                                                                                                        |
-| `onNodeStart`     | ✅ Yes                        | ⚠️ Batched, but resume is partial   | Fires inside the node's runner as a resumable substep. Single-callback resume works cleanly (exactly-once semantics). Multi-callback: see the **single-callback constraint** below.                                                            |
-| `onNodeEnd`       | ⚠️ Only on void-returning nodes | ⚠️ Same as `onNodeStart`           | Nodes that exit via `return value` halt the runner first; the post-body halted check short-circuits the End hook, so it never fires. Use a void node body if you need `onNodeEnd` to fire.                                                     |
-| `onFunctionStart` | ✅ Yes                        | ⚠️ Batched, but resume is partial   | Fires inside the function's runner as a resumable substep. Same as `onNodeStart`.                                                                                                                                                              |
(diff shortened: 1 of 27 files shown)


packages/agency-lang/docs/site/appendix/callbacks.md
```diff
@@ -53,37 +53,48 @@ The table below summarizes the current state per hook. Three columns:
...
+plus the leaf checkpoint's preserved substep counters give each
+callback its own resume point.
+
+Callbacks fire in declared order. Sequential mode is required, not a
+preference — using `mode: "all"` would silently turn ordered side
+effects into a concurrent race.
+
+#### Checkpoint format compatibility
+
+Callback-hook checkpoints written by versions BEFORE the per-callback
```
egonSchiele: no need to add this info on earlier checkpoint versions

PR #190, opened 2026-05-24
Builtin cost guards: guard(cost: $X) as { ... } in std::thread

egonSchiele:
Adds `guard(cost: $X) as { ... }` to `std::thread` — a stdlib function that aborts its block when the cumulative LLM cost exceeds the limit, returning a `Result` so callers can decide what to do.

## Links

- Plan: [`docs/superpowers/plans/2026-05-23-builtin-cost-guards.md`](./packages/agency-lang/docs/superpowers/plans/2026-05-23-builtin-cost-guards.md)
- Spec: [`docs/superpowers/specs/2026-05-20-cost-and-guard-tracking-design.md`](./packages/agency-lang/docs/superpowers/specs/2026-05-20-cost-and-guard-tracking-design.md)
- User guide: [`docs/site/guide/cost-guards.md`](./packages/agency-lang/docs/site/guide/cost-guards.md)

## Surface syntax

```ts
import { guard } from "std::thread"

node main() {
  const result = guard(cost: $2.0) as {
    const a = llm("classify this email")
    const b = llm("draft a reply")
    return a + " — " + b
  }
  if (isFailure(result)) {
    return "over budget: spent " + result.error.actualCost
  }
  return result.value
}
```

`$2.0` is the existing unit literal that compiles to a plain `number` — `guard(cost: $2.0)` is equivalent to `guard(cost: 2.0)`.

## Failure shape

When a guard trips, `result.error` carries:

```ts
type GuardFailureData = {
  type: "guardFailure",
  maxCost: number,
  actualCost: number,
}
```

The `type` field is a `string` rather than a literal so a future `"timeoutFailure"` variant can be added without breaking V1 consumers.

## How it works

1. The agency `guard` function calls `__internal_pushGuard(cost)` which records `{ costLimit, costAtPush: stack.localCost }` on the `StateStack`.
2. The block runs. Inside `prompt.ts`, after each LLM call's cost is added to `targetStack.localCost`, the active guards are walked innermost-first.
3. If any guard's delta (`localCost - costAtPush`) exceeds its limit, `prompt.ts` throws a `GuardExceededError`.
4. The error propagates through the user's code. The function-body auto-wrap (`functionCatchFailure.mustache` + `typescriptBuilder.ts`) re-throws `GuardExceededError` (alongside `RestoreSignal`) so it cannot be silently converted to a generic Failure.
5. The stdlib `guard` function's `try block()` catches the error. `__tryCall` detects the `GuardExceededError` and emits a Failure with the structured `GuardFailureData` shape.
6. `__internal_popGuard()` removes the guard from the stack.

Guards are serialized as part of `StateStack` so they survive checkpoint/restore. Branch seeding (`Runner.seedBranchCost`) clones the parent's guards onto each child stack.

## Files added

- `lib/runtime/guard.ts` + `guard.test.ts` — `GuardEntry`, `GuardExceededError`, `isGuardExceededError`.
- `tests/agency/guards/guard-cost-trip.{agency,test.json}` — tiny budget trips on the first LLM call.
- `tests/agency/guards/guard-cost-no-trip.{agency,test.json}` — generous budget; block runs to completion.
- `tests/agency/guards/guard-nested.{agency,test.json}` — inner trip doesn't trip outer.
- `docs/site/guide/cost-guards.md` — user-facing guide.

## Files modified

- `lib/runtime/state/stateStack.ts` — `guards: GuardEntry[]` + `pushGuard`/`popGuard`, serialized in `toJSON`/`fromJSON`.
- `lib/runtime/runner.ts` — `seedBranchCost` clones parent guards onto child branch stacks.
- `lib/runtime/prompt.ts` — innermost-first guard check after cost accumulation.
- `lib/runtime/result.ts` — `__tryCall` preserves `GuardExceededError` data as structured failure.
- `lib/runtime/index.ts` — exports `GuardExceededError`, `isGuardExceededError`, `GuardEntry`.
- `lib/codegenBuiltins/contextInjected.ts` — registers `__internal_pushGuard` / `__internal_popGuard`.
- `lib/stdlib/thread.ts` — TS implementations for the two builtins.
- `stdlib/thread.agency` — `guard` function + `GuardFailureData` type.
- `lib/templates/backends/typescriptGenerator/{imports,functionCatchFailure,classMethod}.mustache` (+ regenerated `.ts`) — adds `GuardExceededError` to the auto-wrap allow-list.
- `lib/backends/typescriptBuilder.ts` — programmatic auto-wrap also re-throws `GuardExceededError`.
- Drive-by: fixes the trailing-block-with-named-args bug (`CallType.named` gains `blockArg`; `resolveNamed` binds it to the last parameter).
- Fixture regenerations across `tests/typescriptGenerator/` and `tests/typescriptBuilder/` (import block now includes the two new context-injected builtins).
- `CHANGELOG.md` — Unreleased section.

## V1 limitations (documented)

- **Forks don't trip outer guards mid-flight.** Branch costs only roll up to the parent at branch completion via `propagateBranchCost`. An outer guard wrapping a `fork` cannot pre-empt branches in-flight; the recommended pattern is to put the guard *inside* each branch instead. After the fork completes, the next LLM call inside the outer guard will observe the rolled-up spend and trip.
- **Memory layer LLM calls bypass the guard.** `memory.text` / `memory.embed` extraction prompts currently don't flow through the same accumulator path. Tracked as a follow-up.
- **No dimensional unit checking.** `$2.0` and `2.0` are the same `number` at the type level; the unit literal is a notation aid.

## Test results

- `lib/runtime/guard.test.ts`: 5/5 passed
- `lib/runtime/state/stateStack.test.ts`: 4452/4452 passed (5 new guard cases added)
- `pnpm test:run -- guards thread prompt fork stateStack contextInjected`: 4462/4462 passed
- `node ./dist/scripts/agency.js test tests/agency/guards tests/agency/thread`: 7/7 passed (3 new + 4 existing)
- `make`: clean

## Notes for the reviewer

A `guard-cost-fork.agency` fixture was attempted but had to be removed: putting an `llm()` call inside a `fork` branch (with or without a guard) triggers a pre-existing codegen bug — `Error: Unknown scope type: block` from `scopeManager.ts#returnType()`. This is unrelated to guards; the fork-related behavior is covered by the documentation in `cost-guards.md` and by the existing branch-cost-propagation tests in `tests/agency/thread/cost-*`. Flagging in case you want me to file a separate issue for the scope-manager bug.

The trailing-block-with-named-args fix was necessary to make the plan's locked syntax (`guard(cost: $2.0) as { ... }`) work. Without it, named-arg calls silently dropped the trailing block, causing the runtime to report "Missing required argument 'block' in call to 'guard'". The fix adds a `blockArg` field to the named CallType variant, which `resolveNamed` binds to the last non-variadic unbound parameter. This was a small, focused change and made the design intent achievable; let me know if you'd rather split it into its own PR.

The `try` keyword interaction with `GuardExceededError` shook out cleanly: rather than parsing string error messages in agency code, I taught `__tryCall` in `lib/runtime/result.ts` to detect a `GuardExceededError` and emit the structured `GuardFailureData` directly. This keeps the agency-side `guard` function trivial (just `try block()`) and means any future code that uses `try` will also get the structured shape for guard errors.

### DIFF
--- a/packages/agency-lang/lib/runtime/agencyFunction.ts
+++ b/packages/agency-lang/lib/runtime/agencyFunction.ts
@@ -20,11 +20,17 @@ export type CallType =
       positionalArgs: unknown[];
       namedArgs: Record<string, unknown>;
       /** Trailing-block argument from `f(name: val) as { ... }` syntax.
-       *  Filled into the last unfilled non-variadic parameter (by
-       *  convention the block param) during resolveNamed. Separate
-       *  from positionalArgs because the block always binds to the
-       *  trailing block parameter, regardless of which earlier params
-       *  named args have filled. */
+       *
+       *  In the positional case the block is simply appended to `args`
+       *  (it's always the next positional slot), so no separate field
+       *  is needed. In the named case earlier params may be filled by
+       *  name and intermediate params may need UNSET padding, so the
+       *  block must be bound by NAME to the last non-variadic
+       *  parameter — appending it as positional[N] would mis-fill
+       *  parameter N. resolveNamed synthesizes it into namedArgs under
+       *  the last param's name so the rest of fill logic treats it
+       *  uniformly. Without this, `f(name: val) as { ... }` silently
+       *  dropped the block. */
       blockArg?: unknown;
     };
 
--- a/packages/agency-lang/lib/runtime/agencyFunction.ts
+++ b/packages/agency-lang/lib/runtime/agencyFunction.ts
@@ -15,7 +15,18 @@ export type FuncParam = {
 
 export type CallType =
   | { type: "positional"; args: unknown[] }
-  | { type: "named"; positionalArgs: unknown[]; namedArgs: Record<string, unknown> };
+  | {
+      type: "named";
+      positionalArgs: unknown[];
+      namedArgs: Record<string, unknown>;
+      /** Trailing-block argument from `f(name: val) as { ... }` syntax.
+       *  Filled into the last unfilled non-variadic parameter (by
+       *  convention the block param) during resolveNamed. Separate
+       *  from positionalArgs because the block always binds to the
+       *  trailing block parameter, regardless of which earlier params
+       *  named args have filled. */
+      blockArg?: unknown;
+    };
 
 export type ToolDefinition = {
   name: string;
@@ -212,7 +223,11 @@ export class AgencyFunction {
     if (descriptor.type === "positional") {
       return this.resolvePositional(descriptor.args);
     }
-    return this.resolveNamed(descriptor.positionalArgs, descriptor.namedArgs);
+    return this.resolveNamed(
+      descriptor.positionalArgs,
+      descriptor.namedArgs,
+      descriptor.blockArg,
+    );
   }
 
   private resolvePositional(args: unknown[]): unknown[] {
@@ -243,7 +258,11 @@ export class AgencyFunction {
     return result;
   }
 
-  private resolveNamed(positionalArgs: unknown[], namedArgs: Record<string, unknown>): unknown[] {
+  private resolveNamed(
+    positionalArgs: unknown[],
+    namedArgs: Record<string, unknown>,
+    blockArg?: unknown,
+  ): unknown[] {
     // Validate named args: no unknowns, no conflicts with positional
     for (const name of Object.keys(namedArgs)) {
       const paramIdx = this._nonVariadicUnbound.findIndex(p => p.name === name);
@@ -257,6 +276,26 @@ export class AgencyFunction {
           `Named argument '${name}' conflicts with positional argument at position ${paramIdx + 1} in call to '${this.name}'`,
         );
       }
+    }
+
+    // A trailing block argument (from `f(name: val) as { ... }` syntax)
+    // binds to the LAST non-variadic unbound parameter. By convention
+    // that's the block-typed param. We synthesize it into namedArgs
+    // here so the rest of the fill logic treats it uniformly.
+    const hasBlock = blockArg !== undefined;
+    if (hasBlock) {
+      const lastParam = this._nonVariadicUnbound[this._nonVariadicUnbound.length - 1];
+      if (!lastParam) {
+        throw new Error(
+          `Call to '${this.name}' passed a trailing block but the function takes no parameters`,
+        );
+      }
+      if (Object.hasOwn(namedArgs, lastParam.name)) {
+        throw new Error(
+          `Trailing block conflicts with named argument '${lastParam.name}' in call to '${this.name}'`,
+        );
+      }
+      namedArgs = { ...namedArgs, [lastParam.name]: blockArg };
     }
 
     // Build result: positional args first, then fill from named args in parameter order
--- a/packages/agency-lang/lib/templates/backends/typescriptGenerator/functionCatchFailure.ts
+++ b/packages/agency-lang/lib/templates/backends/typescriptGenerator/functionCatchFailure.ts
@@ -6,6 +6,14 @@ import { apply } from "typestache";
 export const template = `if (__error instanceof RestoreSignal) {
   throw __error;
 }
+// GuardExceededError must propagate up to the stdlib \`guard\`
+// function's try/catch (in lib/runtime/result.ts via \`try block()\`).
+// If we converted it to a Failure here, the guard would never see
+// the trip and every guarded block would appear to succeed even
+// over budget. See lib/runtime/guard.ts.
+if (__error instanceof GuardExceededError) {
+  throw __error;
+}
 return failure(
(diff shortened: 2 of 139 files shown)


packages/agency-lang/lib/runtime/agencyFunction.ts
```diff
@@ -15,7 +15,18 @@ export type FuncParam = {
...
+      type: "named";
+      positionalArgs: unknown[];
+      namedArgs: Record<string, unknown>;
+      /** Trailing-block argument from `f(name: val) as { ... }` syntax.
+       *  Filled into the last unfilled non-variadic parameter (by
+       *  convention the block param) during resolveNamed. Separate
+       *  from positionalArgs because the block always binds to the
+       *  trailing block parameter, regardless of which earlier params
+       *  named args have filled. */
+      blockArg?: unknown;
```
egonSchiele, line 34: I am just not sure what this is needed. Blocks syntax currently works... I notice that the positional call type doesn't have a block arg either, why does block syntax work there?

packages/agency-lang/lib/templates/backends/typescriptGenerator/functionCatchFailure.ts
```diff
@@ -6,6 +6,14 @@ import { apply } from "typestache";
 export const template = `if (__error instanceof RestoreSignal) {
   throw __error;
 }
+// GuardExceededError must propagate up to the stdlib \`guard\`
```
egonSchiele, line 9: Suppose that a type script function that is running inside of a guard block throws an exception. What would happen? Would the guard trigger?

packages/agency-lang/lib/runtime/agencyFunction.ts
egonSchiele: Good question — I expanded the doc comment in `d672b01e` to make this concrete.

The asymmetry is in how blocks bind:

- **Positional** (`f(1, 2) as { ... }`): the codegen literally appends the block to `args`, so the descriptor is `{ type: "positional", args: [1, 2, <block>] }`. The block is already in the right slot (the next positional one) and `resolvePositional` consumes it like any other arg. No separate field needed.
- **Named** (`f(name: val) as { ... }`): the block always binds to the LAST non-variadic param, but earlier params may be filled by name (or even left to default + UNSET-padded). If we just stuffed the block into `positionalArgs[0]`, it would bind to param #0 — wrong. Pre-fix, the codegen omitted the block entirely from the named descriptor, which is why `f(name: val) as { ... }` silently dropped the block at runtime.

So `blockArg` is the named-case-only piece of metadata that says "this value binds by NAME to the trailing block param, regardless of which other slots got filled by name vs left to default." `resolveNamed` synthesizes it into `namedArgs[lastParam.name]` so the rest of the fill logic stays uniform.

The regression fixture at `tests/agency/named-arg-with-trailing-block.agency` covers the three shapes that were broken (single named arg + block, mixed positional/named + block, named in reverse declaration order + block).

packages/agency-lang/lib/templates/backends/typescriptGenerator/functionCatchFailure.ts
egonSchiele: Good question. Short answer: **no, the guard does not trigger on a TypeScript-side throw — the user gets a regular Failure, and the guard cleanup still runs.**

Tracing it through:

1. A TS function inside `guard(cost: $X) as { ... }` throws `Error("boom")`.
2. If the enclosing Agency `def` is auto-wrapped (the normal case), the wrapper hits the snippet on this line — and the `instanceof GuardExceededError` check rethrows only `GuardExceededError`. A plain `Error` falls through to the `return failure(__error.message, ...)` path, so the inner Agency call returns a regular Failure.
3. The block body then bubbles that Failure up (or, if the throw escapes Agency entirely, `try block()` in `stdlib::guard` catches it and converts it to a Failure too — same shape, different path).
4. Either way, control reaches `__internal_popGuard()` in `stdlib::guard`, so the guard entry comes off the stack cleanly.
5. The user sees `Failure { error: "boom", ... }` — not a `GuardFailureData` Failure. `isFailure(result)` is true but `result.error.type` is undefined.

So "guard trips" is reserved exclusively for the LLM-cost check in `prompt.ts` doing `cost > limit`. Any other error path produces a plain Failure with the original message. That seemed right because:

- A guard tripping should mean "I spent too much money", not "some code I called blew up".
- The user can still distinguish the two by inspecting `result.error` shape (`GuardFailureData` has `{ type, maxCost, actualCost }`; an ordinary failure has just a string message).

One subtlety worth flagging: an LLM call that adds cost and **then** throws (e.g. network error after a successful tool call) still updates `localCost` before throwing, so the next LLM call inside the guard would see the accumulated spend — the throw doesn't "un-bill" the call. That matches reality (the call really happened, we really paid for it) and matches what `getCost()` would report.

PR #192, opened 2026-05-24
fix: support `return llm(...)` inside any block body

egonSchiele:
## What

Fixes a codegen bug where `return llm(...)` inside any block body
(fork branch, guard block, any `as { ... }` body) crashed
compilation with `Unknown scope type: block`.

## Why it was broken

Two interacting bugs:

1. `ScopeManager.returnType()` did not handle block scope. It only
   recognised `global`/`function`/`node` and threw on `block`. Any
   path that called `returnType()` from inside a block (e.g.
   `processLlmCall` looking up the type hint for structured output)
   crashed.
2. Even if you got past #1, `processReturnStatement`'s block branch
   ran `processNode(node.value)` and wrapped the result in
   `runner.halt(...)`. For `llm(...)` that produces a statements
   list, so codegen emitted invalid TypeScript like
   `runner.halt(stmt; stmt; ...);`.

## What changed

- `lib/backends/typescriptBuilder/scopeManager.ts` — `returnType()`
  now returns `undefined` for block scope. Documented why:
  propagating the block's declared return type into the builder
  would require threading the block-parameter type through every
  block-scope push site, which is non-trivial.
- `lib/backends/typescriptBuilder.ts` — `processReturnStatement`'s
  block branch now has a dedicated `return llm(...)` path that
  hoists `processLlmCall`'s statements and halts the runner with
  the `__prompt` self-ref. Uses scope `"block"` so the assignment
  lands at `__bstack.locals.__prompt`, which is what `ts.self(...)`
  resolves to inside a block frame.
- `docs/site/guide/llm.md` — documents the limitation that
  `return llm(...)` from a block always uses a string-typed schema,
  with the workaround (assign to a typed local, then return it).
- `tests/agency/block-return-llm.{agency,test.json}` — regression
  fixture.
- `tests/agency/guards/guard-cost-fork.agency` — removes the
  obsolete "pre-existing codegen bug" workaround note.

## Known limitation (documented, not in scope)

A bare `return llm(...)` from a block always uses a `string`
structured-output schema, even when the surrounding block's
declared return type is something else. Users who want a typed
object should assign first:

```ts
const reply: MyType = llm("...")
return reply
```

Propagating the declared block return type into the builder would
require either threading the type through every `scopes.push({
type: "block", ... })` site or looking it up via the typechecker —
substantial enough to be its own PR.

## Validation

- `pnpm test:run -- typeChecker typescriptBuilder typescriptGenerator scopeManager`
  → 4443 / 4443 pass.
- `pnpm run agency test tests/agency/block-return-llm.agency` → pass.

### DIFF
--- a/packages/agency-lang/docs/site/guide/llm.md
+++ b/packages/agency-lang/docs/site/guide/llm.md
@@ -18,6 +18,37 @@ const response: Response = llm("What is the capital of France?")
 print(response)
 ```
 
+### Limitation: structured output inside block bodies
+
+When you write `return llm(...)` as the last statement of a block
+body (a `fork(...) as item { ... }` branch, a `guard(cost: $X) as { ... }`
+body, or any user-defined function that takes a block parameter),
+Agency currently has no way to know the block's declared return type
+at codegen time — so the structured-output schema falls back to
+`string`. The block still runs and returns the LLM's text reply, but
+you cannot get a typed object out of it.
+
+```ts
+type Response = { capital: string }
+
+// ❌ The annotation on `result` is NOT propagated into the block.
+// Each branch returns a plain string from the LLM.
+const result: Response[] = fork(["France", "Spain"]) as country {
+  return llm(`What is the capital of ${country}?`)
+}
+```
+
+The workaround is to assign the LLM call to a typed local first, then
+return it:
+
+```ts
+// ✅ The annotation on `reply` controls the LLM's structured output.
+const result: Response[] = fork(["France", "Spain"]) as country {
+  const reply: Response = llm(`What is the capital of ${country}?`)
+  return reply
+}
+```
+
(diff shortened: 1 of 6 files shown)


packages/agency-lang/docs/site/guide/llm.md
```diff
@@ -18,6 +18,37 @@ const response: Response = llm("What is the capital of France?")
 print(response)
 ```
 
+### Limitation: structured output inside block bodies
+
+When you write `return llm(...)` as the last statement of a block
```
egonSchiele: A return statement can show up anywhere in the block, not just as the last statement.

egonSchiele: Good catch — you're right, `return` can appear anywhere in a block (early returns, conditional returns). My codegen fix in `processReturnStatement` already handles all positions; only the doc wording was wrong. Reworded in `480fdc21` to say "anywhere in the block, not just at the end."

PR #193, opened 2026-05-24
Timeout guards: guard(time: ...) and guard(cost:, time:)

egonSchiele:
Extends the existing `guard(...) as { ... }` stdlib function with an optional `time:` parameter. When the wall-clock time spent *executing* inside the block exceeds the limit, the block aborts and `guard` returns a `Failure` carrying `GuardFailureData` with a `"timeoutFailure"` variant.

Plan: [`docs/superpowers/plans/2026-05-24-timeout-guards.md`](docs/superpowers/plans/2026-05-24-timeout-guards.md)

## What's new

```ts
import { guard } from "std::thread"

const r = guard(time: 30s) as {
  return planAndExecute()
}
if (isFailure(r)) {
  // r.error.type === "timeoutFailure"
  // r.error.maxTime, r.error.actualTime (ms)
}

// Combine with cost — whichever fires first wins
const r = guard(cost: $5, time: 60s) as { ... }
```

Time semantics are **compute-time**: the timer only ticks while a runner is actively executing inside the guarded scope. Time paused on an interrupt does not count; on resume the timer is re-armed with the remaining budget. In-flight LLM HTTP requests are cancelled via the composed `AbortSignal`; Agency tool bodies stop at the next runner step boundary.

## How it works

The existing `GuardEntry` POJO is refactored into a polymorphic `Guard` interface in `lib/runtime/guard.ts`. `StateStack` and `Runner` only ever talk to the interface — no per-variant `instanceof` checks.

Lifecycle:
| Hook | When | CostGuard | TimeGuard |
| --- | --- | --- | --- |
| `install(stack)` | on push | capture `costAtPush` | compose AbortController into `stack.abortSignal`, schedule `setTimeout` |
| `pause()` | on `Runner.halt` | no-op | charge in-flight delta to `elapsedMs`, clear timer |
| `resume(stack)` | at every step entry | no-op | re-arm `setTimeout(timeLimit - elapsedMs)` |
| `check(stack)` | at sync points | return error if `spent > limit` | return error if `tripped && !consumed` |
| `uninstall(stack)` | on pop | no-op | clear timer, restore previous abortSignal |
| `cloneForBranch(parent, child)` | fork seed | fresh clone with rebased `costAtPush` | `undefined` (parent's timer is single source of truth) |

Idempotency lives on the guard itself (TimeGuard's `state` field), so multiple Runners halting/stepping in the same JS tick don't double-charge or double-arm.

Fork/race: CostGuards clone per-branch (each branch independently tracks cost-since-push). TimeGuards do not clone — the parent's timer fires once and the abort cascades to every branch's `stack.abortSignal` via `composeBranchAbortSignal`; branches halt silently in their own `Runner.shouldSkip`, control returns to the parent, and the parent's next sync point produces the `timeoutFailure`.

Checkpoint/restore round-trips work: `toJSON` serializes only persistent state (cost baselines, `elapsedMs`, `timeLimit`); `resume()` re-establishes the runtime plumbing (AbortControllers, timers) on the first runner step after deserialization.

## Three subtle fixes that fall out of integration testing

1. **Block runners didn't receive `__stateStack`** — the surrounding `guard(...) as { ... }`'s guards were invisible inside the block body. Fixed in `blockSetup.mustache` by passing `stack: __ctx.stateStack`.

2. **`TimeGuard.check()` reported `spent: 0`** when called mid-window (timer fired but no pause has happened to charge `elapsedMs`). Now charges the in-flight delta.

3. **Trip re-thrown after `try` caught it** — the still-aborted signal made `Runner.shouldSkip` re-throw on `popGuard`'s step. Added a `consumed` flag (TimeGuard's `check()` returns null after the first trip) plus an `isTripped()` method on the `Guard` interface; `shouldSkip` uses it to distinguish guard-owned aborts (don't silent-halt cleanup) from race-loser cancels.

## Limitations (documented)

- **JS-bodied tool calls cannot be aborted mid-execution.** They run to completion in the background; only the next runner step sees the trip and abandons their result. Agency tool bodies, LLM HTTP requests, and the runner itself observe the abort signal promptly.
- **Cost deltas from a fork only roll up to an outer cost guard at fork completion** — same V1 limitation as cost guards alone. Time guards do *not* have this limitation: the parent's timer is wall-clock.
- Memory-layer LLM calls still bypass guard cost accounting (unchanged from V1).

## Tests

- **24 unit tests** in `lib/runtime/guard.test.ts` covering install/uninstall/pause/resume idempotency, JSON round-trip, `cloneForBranch`, in-flight charging, consumed semantics, `isTripped`, and `CostGuard.isTripped`.
- **4 new agency E2E tests** under `tests/agency/guards/`: `guard-time-no-trip`, `guard-time-trip`, `guard-time-and-cost`, `guard-time-nested`.
- **All 9 existing cost-guard tests pass with zero changes** — that was the contract of the refactor.
- Full unit suite green (4458 tests), `pnpm run lint:structure` clean, `make` clean.

## Docs

`docs/site/guide/cost-guards.md` → `docs/site/guide/guards.md`, with new sections for Timeout, Combining cost and time, Forks / parallel / message threads, and updated Limitations.


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -1507,7 +1507,7 @@ export class TypeScriptBuilder {
     // Create runner for step execution
     setupStmts.push(
       ts.raw(
-        `const runner = new Runner(__ctx, __stack, { state: __stack, moduleId: ${JSON.stringify(this.moduleId)}, scopeName: ${JSON.stringify(functionName)} });`,
+        `const runner = new Runner(__ctx, __stack, { state: __stack, moduleId: ${JSON.stringify(this.moduleId)}, scopeName: ${JSON.stringify(functionName)}, stack: __stateStack });`,
```
egonSchiele, line 1510: this is a lot of raw code. Consider using ts ir builders or a mustache template.

packages/agency-lang/lib/codegenBuiltins/contextInjected.ts
```diff
@@ -157,16 +157,21 @@ export const CONTEXT_INJECTED_BUILTINS: Record<string, ContextInjectedBuiltin> =
     params: [],
     returnType: number,
   },
+  // Push 0 or more guards (cost, time, or both) and return the count
+  // pushed so the caller knows how many to pop. Param types are "any"
+  // because each can be `number | null` (null = "no limit on this
```
egonSchiele, line 162: why not create a `number | null` type then for better type checking?

packages/agency-lang/lib/codegenBuiltins/contextInjected.ts
```diff
@@ -157,16 +157,21 @@ export const CONTEXT_INJECTED_BUILTINS: Record<string, ContextInjectedBuiltin> =
...
+  // dimension"). The stdlib `guard` function threads the count to
+  // `__internal_popGuard(count)`.
   __internal_pushGuard: {
     name: "__internal_pushGuard",
     from: THREAD_FROM,
-    params: [number],
-    returnType: voidT,
+    params: [ANY_T, ANY_T],
+    returnType: number,
```
egonSchiele, line 169: what number gets returned?

packages/agency-lang/lib/runtime/guard.ts
```diff
@@ -1,37 +1,350 @@
...
+  install(stack: StateStack): void {
+    this.costAtPush = stack.localCost;
+  }
+
+  uninstall(_stack: StateStack): void {
+    /* nothing */
+  }
+
+  pause(): void {
+    /* nothing — cost is checked at sync points */
```
egonSchiele, line 94: What happens if an interrupt is thrown inside of a cost guard? Am I right in thinking we don't need to do any special tracking on the guard, because the cost is being tracked on the state stack?

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -106,10 +123,34 @@ export class Runner {
...
+   *      (caught by `try`). The popGuard cleanup steps still need to
+   *      run, so don't halt; fall through.
+   *   3. No guard returns an error and none isTripped() → external
+   *      abort (race-loser branch cancel). Halt silently as before. */
   private shouldSkip(): boolean {
     if (this.stack?.abortSignal?.aborted && !this.halted) {
-      this.halt(undefined);
+      // Walk innermost-first so the deepest guard reports its trip
+      // first. Mirrors the order in prompt.ts's cost-check loop.
+      for (let i = this.stack.guards.length - 1; i >= 0; i--) {
```
egonSchiele, line 146: I am seeing this pattern in a few different places, would it make sense to add a method on the state stack that encapsulates this logic?

packages/agency-lang/tests/agency/guards/guard-time-and-cost.agency
```diff
@@ -0,0 +1,19 @@
...
+//   3. the time variant produces "timeoutFailure" with maxTime set
+//      and the cost fields null.
+node main() {
+  const result = guard(cost: $1.0, time: 20ms) as {
+    sleep(150ms)
+    sleep(1ms)
+    return "did not trip"
+  }
+  if (isFailure(result)) {
+    return "tripped:${result.error.type}:maxTime=${result.error.maxTime}:maxCost=${result.error.maxCost}"
```
egonSchiele, line 16: This is a pretty unreadable format. Why not return an object instead?

packages/agency-lang/tests/agency/guards/guard-time-trip.agency
```diff
@@ -0,0 +1,22 @@
...
+  const result = guard(time: 20ms) as {
+    sleep(150ms)
+    sleep(1ms)
+    return "did not trip"
+  }
+  if (isFailure(result)) {
+    // actualTime > maxTime catches a bug where actualTime is just set
+    // to the limit instead of the real elapsed; actualTime > 0 alone
+    // would not.
+    return "tripped:${result.error.type}:${result.error.maxTime}:${result.error.actualTime > result.error.maxTime}"
```
egonSchiele, line 19: same, unreadable

PR #194, opened 2026-05-25
Shared cost guards: real-time mid-fork enforcement

egonSchiele:
## Summary

Today each fork branch gets an isolated clone of the parent `CostGuard`, so an outer `guard(cost: $X)` budget can be silently exceeded mid-fork (two branches at $1.25 each under a $2.00 cap — neither clone trips, and the cap only fires after the fork settles).

This PR replaces per-branch cloning with a **shared in-memory CostGuard reference**, so any branch charging the shared counter mutates what the parent and all siblings see in real time. The next post-call check from any descendant returns the trip the moment cumulative spend exceeds the budget.

Plan: [`docs/superpowers/plans/2026-05-24-shared-cost-guards.md`](packages/agency-lang/docs/superpowers/plans/2026-05-24-shared-cost-guards.md)

## Key mechanics

- `CostGuard` owns its own `spent` counter (no more derived from `stack.localCost - costAtPush`). Every LLM-cost site walks `stack.guards` and calls `guard.charge(cost)` on each one.
- `CostGuard.cloneForBranch` returns `this` — the same JS object — so every branch's `stack.guards` array holds a reference to the parent's `CostGuard`. Single-threaded JS makes the increment race-free.
- `StateStack.inheritedGuardCount` records how many of a child's guards are parent-owned (refs at indices `< count`). `toJSON` serializes only branch-owned guards (slice past `count`), so a shared parent guard is serialized **once** on the parent's snapshot.
- `runBatch.rehydrateInheritedGuards` re-prepends the parent's live guard references onto each child stack on resume, restoring real-time sharing across the checkpoint cycle. Idempotent via a live-only `rehydrated` flag on `BranchState`.
- **Pre-call gate** in `prompt.ts`: before issuing an LLM request, walk the guard stack. If a sibling already pushed us over budget, refuse the call before it hits the wire.

`CostGuard` deliberately does NOT install an `AbortController` (unlike `TimeGuard`). A sibling branch mid-LLM-call won't be cancelled the instant another branch trips the shared guard — its smoltalk request runs to completion, then its post-call check sees the over-limit counter and throws. This preserves the same-stack nested-guard "innermost-first" reporting that `guard-nested-iteration-order` guarantees.

## Test coverage

New E2E tests in `tests/agency/guards/`:

| Test | Verifies |
|---|---|
| `guard-cost-shared-outer-trips-mid-fork` | Motivating example: outer trips mid-fork before any branch finishes. |
| `guard-cost-shared-survives-interrupt` | 3 branches interrupt under budget. On resume, rehydrated shared OUTER trips. |
| `guard-cost-shared-inner-still-isolated` | Inner guards a branch pushes after fork-open stay branch-local; one branch's inner trip does not cascade to siblings. |
| `guard-cost-pre-call-gate` | Engineers OUTER over its limit without tripping (via an inner-guard-catches-trip pattern); next `llm()` is refused by the pre-call gate; proven by providing EXACTLY 4 mocks (a 5th would throw "too many calls"). |
| `guard-cost-shared-nested-fork` | Depth-2 fork verifies inheritance propagates through TWO `seedBranchCost` hops. |

Updated `guard-cost-fork` to assert `tripped-mid-fork:true` (was `tripped-after-fork:true` under the V1 deferred-rollup behavior).

## Docs

- [`docs/site/guide/guards.md`](packages/agency-lang/docs/site/guide/guards.md) — documents the shared-guard model, the pre-call gate, and the inner-stays-local rule. Removed the V1 "forks don't trip outer cost guards mid-flight" limitation.

## Verification

- `pnpm run agency test tests/agency/guards/` — **20/20 pass**
- `pnpm test:run` — **4464/4464 pass**

### DIFF
--- a/packages/agency-lang/lib/runtime/runBatch.ts
+++ b/packages/agency-lang/lib/runtime/runBatch.ts
@@ -195,38 +195,19 @@ function composeBranchAbortSignal(
   return composed;
 }
 
-/** Populate a child branch's `stack.guards` with the parent's inherited
- *  guards (via each guard's `cloneForBranch`), and stamp
- *  `inheritedGuardCount` so future serialization slices them off. This
- *  must run before any code reads `branch.stack.guards` or before the
- *  branch is `invoke`d.
- *
- *  Handles both fresh and resumed branches:
- *   - Fresh branch (first run): `branch.stack.guards` is empty;
- *     prepended refs become the entire array.
- *   - Resumed branch (post-deserialize): `branch.stack.guards` holds
- *     only the branch-owned guards (slice from `toJSON`); prepended
- *     refs go in front of them.
- *
- *  Idempotent via `branch.guardsRehydrated` (a live-only flag — reset
- *  to undefined automatically on deserialize, so a fresh execution
- *  always re-runs this once per branch).
- *
- *  `CostGuard.cloneForBranch` returns `this` (shared reference);
- *  `TimeGuard.cloneForBranch` returns `undefined` (parent's timer is
- *  single source of truth via abort cascade). Other future guards
- *  pick their own semantic. */
+/** Thin per-execution idempotency wrapper around
+ *  `StateStack.rehydrateInheritedGuardsFrom`. The inner method handles
+ *  the slice/prepend invariant and the resume validation; this wrapper
+ *  ensures we only do it once per BranchState per execution (the flag
+ *  is live-only — automatically reset on deserialize). Must run before
+ *  any code reads `branch.stack.guards` or before the branch is
+ *  `invoke`d. */
 function rehydrateInheritedGuards(
   branch: BranchState,
   parentStack: StateStack,
 ): void {
   if (branch.guardsRehydrated) return;
-  const child = branch.stack;
-  const inheritedRefs = parentStack.guards
-    .map((g) => g.cloneForBranch(parentStack, child))
-    .filter((g): g is NonNullable<typeof g> => g !== undefined);
-  child.guards = [...inheritedRefs, ...child.guards];
-  child.inheritedGuardCount = inheritedRefs.length;
+  branch.stack.rehydrateInheritedGuardsFrom(parentStack);
   branch.guardsRehydrated = true;
 }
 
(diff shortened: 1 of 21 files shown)


packages/agency-lang/lib/runtime/runBatch.ts
```diff
@@ -195,6 +195,41 @@ function composeBranchAbortSignal(
...
+ *
+ *  `CostGuard.cloneForBranch` returns `this` (shared reference);
+ *  `TimeGuard.cloneForBranch` returns `undefined` (parent's timer is
+ *  single source of truth via abort cascade). Other future guards
+ *  pick their own semantic. */
+function rehydrateInheritedGuards(
+  branch: BranchState,
+  parentStack: StateStack,
+): void {
+  if (branch.guardsRehydrated) return;
```
egonSchiele, line 209: This is a leaky abstraction. The statestack should instead have a function on it that can be called to rehydrate guards.

packages/agency-lang/tests/agency/guards/guard-cost-pre-call-gate.agency
```diff
@@ -0,0 +1,54 @@
...
+// OUTER's `spent` is now over its budget.
+//
+// After the fork returns, OUTER's body issues llm("after"). The
+// pre-call gate walks the stack — OUTER's check returns
+// GuardExceededError ($0.000008 > $0.000005) — the call is REFUSED
+// before smoltalk.complete() runs. The mock for "after" is never
+// consumed.
+//
+// llmMocks provides EXACTLY 4 mocks (a1, b1, a2, b2). If the
+// pre-call gate is broken, "after" would be sent and the
```
egonSchiele: Wouldn't the behavior be exactly the same if the pre-call check was broken but the post-call check still worked?

packages/agency-lang/tests/agency/guards/guard-cost-shared-inner-still-isolated.agency
```diff
@@ -0,0 +1,55 @@
...
+  const outer = guard(cost: 0.001) as {
+    let results = fork([1, 2, 3]) as n {
+      if (n == 2) {
+        return branchWork(n, 2)
+      }
+      return branchWork(n, 1)
+    }
+    return results
+  }
+  if (isFailure(outer)) {
```
egonSchiele, line 51: do you also want to check the values of all the inner threads? Only the 2nd should be a failure. What should the other two's values be?

PR #196, opened 2026-05-25
feat: make stdlib sleep, input, prompt, exec, bash, oauth, browserUse, speech, system abortable

egonSchiele:
Follow-up to #195. Promotes the rest of the high-priority stdlib functions (plus `speech` and `system`) to the abortable / context-injected pattern, so a Ctrl-C, race-loser cleanup, or time-guard abort actually tears down their in-flight work instead of leaving subprocesses, timers, fetches, or readline interfaces hanging.

## What is now abortable

| Function | What gets torn down on abort |
| --- | --- |
| `sleep(ms)` | `setTimeout` is cleared and the promise rejects with `AgencyCancelledError` |
| `input(prompt)` | The `readline` interface is closed (releases stdin) |
| `prompt(question)` (`std::ui`) | Same as above, plus the fixed input bar state is reset |
| `exec(cmd, args, ...)`, `bash(cmd, ...)` (`std::shell`) | The child process gets `SIGTERM` |
| `openUrl(url)`, `screenshot(...)` (`std::system`) | The `open` / `screencapture` / `import` child process gets `SIGTERM` |
| `speak(text, ...)`, `record(...)`, `transcribe(file, ...)` (`std::speech`) | `say` / `rec` get `SIGTERM` (and `rec` restores stdin out of raw mode); the Whisper upload fetch is cancelled |
| `browserUse(task, ...)` (`std::browser`) | Session-creation fetch, polling fetch, and inter-poll sleep all see the abort signal so we stop hitting the remote API |
| `authorize(name, ...)`, `getAccessToken(name)` (`std::oauth`) | Local OAuth callback server is closed (otherwise it would sit on the port for the full 5-minute timeout) and the token-exchange / refresh-token fetch is cancelled |

## Mechanism

Same context-injected builtin recipe as #195:

1. Register an entry in `lib/codegenBuiltins/contextInjected.ts`.
2. Rename `_foo` to `__internal_foo` and take `(ctx, stack, threads, ...userParams)`.
3. Use `ctx.getAbortSignal(stack)` to get a composite signal that fires on either global or per-branch abort.
4. Thread it into whatever API supports it (`fetch`'s `signal:`, `spawn`'s `signal` option, `addEventListener("abort", ...)` for things that don't).
5. Translate the resulting `AbortError` to `AgencyCancelledError` so `__tryCall` re-throws (instead of silently turning cancellation into a `Failure` value).

## Code reuse

To keep the pattern from getting more boilerplate-heavy than it already is:

- New `lib/stdlib/abortable.ts` exposes three shared helpers used across `shell.ts`, `system.ts`, `speech.ts`, `builtins.ts`, and `browserUse.ts`:
  - `abortableSpawn(command, args, opts)` — spawn that pipes stdout/stderr and `SIGTERM`s the child on abort
  - `abortableExec(command, args, signal)` — same for fire-and-forget children with no output (so e.g. `say` doesn't buffer audio output indefinitely)
  - `abortableSleep(ms, signal)` — used by both `__internal_sleep` and the browser-use polling loop
- `runHttp()` from `http.ts` is now exported and reused by `oauth.ts`, `speech.ts`, and `browserUse.ts` for their `fetch` calls, so the `AbortError` -> `AgencyCancelledError` translation lives in one place.

## OAuth + calendar

`oauth.ts` still exports non-cancellable `_authorize` and `_getAccessToken` wrappers because `calendar.ts` calls them internally and isn't being migrated in this PR. The core implementations are factored as private `authorizeImpl(name, config, signal?)` / `getAccessTokenImpl(name, signal?)`, so there is no duplicated logic — just two thin wrappers around the same body.

## On the pattern itself

The current "context-injected builtin" mechanism is starting to feel ugly. Each abortable function needs (1) a registry entry, (2) the `__internal_` name prefix, (3) codegen interaction, and (4) the impl must take `ctx, stack, threads` as its first three params. Two PRs in and we already have ~20 entries.

How other languages handle this:

- **Go**: same as us — pass `context.Context` as the first param of every function. Honest but verbose.
- **Python**: `contextvars.ContextVar` — set once at the top of execution, any function reads it via `var.get()`. asyncio propagates them across `await`.
- **Ruby**: `Thread.current[:x]` / `Fiber[:x]`. Rails wraps this as `ActiveSupport::CurrentAttributes`.
- **Node**: `AsyncLocalStorage` from `async_hooks`. Same primitive: context that flows across async boundaries. Used by Fastify, OpenTelemetry, etc.

I think the right long-term direction is `AsyncLocalStorage`. Set ctx/stack/threads at the top of every run; refresh the `stack` slot when entering a fork/race branch; have stdlib functions just call `getRuntimeContext().getAbortSignal()`. That eliminates the registry, the `__internal_` prefix, and the codegen rewrite. Stdlib functions become ordinary TS — no more "this stdlib function is special, sometimes."

That refactor is bigger than this PR; happy to plan it as a follow-up if you agree on the direction.

## Tests

- All 4570 existing tests still pass.
- New file `lib/stdlib/abortable.test.ts` covers the critical paths: `abortableSleep` cancels mid-sleep, `abortableSpawn` `SIGTERM`s the child, `__internal_sleep` wakes on `ctx.cancel()`, `__internal_input` releases stdin on abort, `__internal_exec` and `__internal_bash` kill their subprocesses on abort, plus a happy-path test that `abortableSpawn` runs `printf` to completion when no abort fires.
- Fixtures in `tests/typescriptGenerator/` and `tests/debugger/` were regenerated by `make fixtures` — they include additional `__internal_*` imports in every generated file because the registry grew.


packages/agency-lang/stdlib/speech.agency
```diff
@@ -26,24 +32,28 @@ export def record(outputFile: string = "", silenceTimeout: number = 2000): strin
...
 
+  Cancellation: an in-progress recording is stopped on Ctrl-C, race-loser, or time-guard abort, surfacing as an AgencyCancelledError.
+
   @param outputFile - File path to save audio to (auto-generated if empty)
   @param silenceTimeout - Silence before auto-stopping in ms (0 to disable)
   """
   return interrupt std::record("Allow microphone recording?", {})
 
-  return _record(outputFile, silenceTimeout)
+  return __internal_record(outputFile, silenceTimeout)
```
egonSchiele, line 42: do these need a `try` prefixed? Can they throw an error?

egonSchiele: Note that using `try` will change the return type to `Result`, so may need to update tests

PR #198, opened 2026-05-25
drop per-call-site context plumbing (ALS migration follow-up)

egonSchiele:
## Summary

Follow-up to the ALS migration (#1XX, merged at d39103cc). Stops emitting the `{ ctx: __ctx, threads: __threads, stateStack: __stateStack }` context-bag at every Agency function call site, `runPrompt` invocation, `callHook` invocation, and `Runner` construction. Runtime helpers now read those three values from the active `agencyStore` frame instead of accepting them as positional/named arguments.

Net effect across the 4 commits: **105 files changed, +362 / −1269**. The bulk of the deletion is fixture diff — the generated `.mjs` files in `tests/typescriptGenerator/` and elsewhere shrink because the bag-emission boilerplate is gone.

### Phase 1 — `__call` / `__callMethod` (`627ae953`)
- Runtime: both helpers now read `ctx` / `threads` / `stateStack` from `agencyStore` and merge with caller-provided extras (the `{moduleId, scopeName, stepPath}` location bag for `checkpoint`/`getCheckpoint`/`restore` sites and the `{ctx}` bag from inside `__initializeGlobals` are the only surviving call-site extras).
- Codegen: `buildStateConfig()` returns `undefined` for the common case. Emission sites (`processValueAccess`, `emitRuntimeDispatchCall`, `buildHandlerArrow`, `assignmentEmitter`, `pipeChainEmitter`) skip the third arg when undefined.
- Net: 67 fixture files lose ~5 lines per `__call`/`__callMethod` site.

### Phase 2 — `runPrompt` (`97be24b2`)
- Runtime: `runPrompt({prompt, messages, clientConfig, ...})` — `ctx` and `stateStack` are no longer accepted; both come from `getRuntimeContext()`.
- Codegen: drops the `ctx: __ctx` and `stateStack: __stateStack` entries from the emitted runPrompt config object.

### Phase 3 — Runner constructor and `callHook` (`5fc6c445`)
- Runner: `stack` and `threads` opts are no longer emitted by codegen. The constructor reads both from ALS as a fallback so `runner.step/hook/pipe/fork`'s `runInScope` still re-seeds the per-scope frame correctly. Direct test usages (`new Runner(ctx, frame)`) keep working unchanged.
- `callHook({name, data})`: `ctx` is now optional; missing → resolved from ALS. Codegen stops emitting `ctx: __ctx`.
- Templates `blockSetup.mustache` and `forkBlockSetup.mustache` lose the `stack` / `threads` keys (regenerated via `pnpm run templates`).
- **External entry points get an ALS seed**: `respondToInterrupts` (via `runResumeLoop`) and `rewindFrom` now wrap their `graph.run(...)` call in `agencyStore.run(...)`. Without this, the new ALS-driven `callHook` and the existing ALS-driven stdlib helpers would throw `getRuntimeContext()` errors on every resume / rewind path. This restored 6 debugger driver tests that broke when `callHook` became ALS-only.

### Docs (`63f490a4`)
- New "Codegen contract" section in [docs/dev/async-context.md](docs/dev/async-context.md) describes the post-migration call-site shape and points back to both plans.

## What's NOT in this PR

The plan's Phase 4 (drop `__ctx`/`__threads`/`__stateStack` as locals from generated node/function bodies) is left for a follow-up. The pre-condition for Phase 4 — that callers no longer need to thread these through — is satisfied by this PR. But the body-local cleanup is genuinely the trickier half: replacing every `__ctx.checkpoints.create(__stateStack, ...)`, `__ctx.pendingPromises.awaitAll()`, `__threads.createSubthread()` etc. with either ALS lookups or per-feature accessor helpers needs more code-surgery than the per-call-site bag drop, and on the resume path `__threads` from `setupNode` can be a DIFFERENT instance from `ALS.threads` (the resume helpers seed a placeholder ThreadStore that's replaced by the deserialized one inside `setupNode`), so a naive substitution would silently break resume.

Phase 5/6 (cleanup) are partially done: the `respondToInterrupts`/`rewindFrom` ALS wrapping in Phase 3 covers most of Phase 5's "wrap external entry points" goal, and the docs commit covers the visible piece of Phase 6.

## Verification

- `make` and `make fixtures` clean every phase.
- `pnpm test:run` — **260 test files, 4418 tests, all passing** at HEAD.
- Hand-checked one fixture diff per phase to confirm the deletion is exactly the expected bag boilerplate.

## Reviewer note

The fixture diff is large and mechanical (multiple thousand lines removed). Focus on:
- `lib/runtime/call.ts` — new ALS-merging behaviour in `__call` / `__callMethod`.
- `lib/runtime/prompt.ts` — `runPrompt` signature change.
- `lib/runtime/runner.ts` — constructor ALS fallback.
- `lib/runtime/hooks.ts` — `callHook` ctx optional.
- `lib/runtime/interrupts.ts` and `lib/runtime/rewind.ts` — outer `agencyStore.run` wrap around the resume/rewind loop.
- `lib/backends/typescriptBuilder.ts` and `lib/backends/typescriptBuilder/{assignmentEmitter,pipeChainEmitter}.ts` — codegen-side drops.
- Templates `lib/templates/backends/typescriptGenerator/{blockSetup,forkBlockSetup}.mustache` (mechanical).

Refs `docs/superpowers/plans/2026-05-25-drop-per-call-context-plumbing.md`.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -821,16 +821,26 @@ export class TypeScriptBuilder {
           const fnCall = element.functionCall;
           const descriptor = this.buildCallDescriptor(fnCall);
           const configObj = this.buildStateConfig();
-          const callArgs: TsNode[] = [result, ts.str(fnCall.functionName), descriptor, configObj];
-          if (element.optional) callArgs.push(ts.bool(true));
+          const callArgs: TsNode[] = [result, ts.str(fnCall.functionName), descriptor];
+          if (element.optional) {
+            callArgs.push(configObj ?? ts.id("undefined"));
+            callArgs.push(ts.bool(true));
+          } else if (configObj) {
+            callArgs.push(configObj);
+          }
           result = this.awaitChainCall(ts.call(ts.id("__callMethod"), callArgs), element === node.chain[node.chain.length - 1]);
           break;
         }
         case "call": {
           const descriptor = this.buildCallDescriptor(element);
           const configObj = this.buildStateConfig();
-          const callArgs: TsNode[] = [result, descriptor, configObj];
-          if (element.optional) callArgs.push(ts.bool(true));
+          const callArgs: TsNode[] = [result, descriptor];
+          if (element.optional) {
+            callArgs.push(configObj ?? ts.id("undefined"));
+            callArgs.push(ts.bool(true));
+          } else if (configObj) {
+            callArgs.push(configObj);
+          }
           result = this.awaitChainCall(ts.call(ts.id("__call"), callArgs), element === node.chain[node.chain.length - 1]);
           break;
         }
@@ -962,23 +972,33 @@ export class TypeScriptBuilder {
   }
 
   /**
-   * Build the standard state config for __call/__callMethod dispatch.
-   * During global init, only ctx is available; otherwise includes threads
-   * and stateStack.
+   * Build the state config passed as the third argument to `__call` /
+   * `__callMethod`.
+   *
+   * After the ALS migration, the runtime helpers read `ctx`/`threads`/
+   * `stateStack` from `agencyStore` (see lib/runtime/call.ts), so call
+   * sites no longer need to thread those locals. This helper returns
+   * `undefined` for the common case — the caller MUST omit the arg
+   * entirely so the emitted code stays minimal.
+   *
+   * It still returns a real object in two situations:
+   *  - `insideGlobalInit`: globals run before the ALS frame is
+   *    installed, so the bag must carry `{ ctx }` explicitly.
+   *  - When `opts.extra` is provided (currently the `checkpoint` /
+   *    `getCheckpoint` / `restore` location info — `moduleId`,
+   *    `scopeName`, `stepPath`): these extras are merged into the
+   *    runtime-built bag by `__call`.
    */
   private buildStateConfig(opts?: {
-    stateStack?: TsNode;
     extra?: Record<string, TsNode>;
-  }): TsNode {
+  }): TsNode | undefined {
     if (this.insideGlobalInit) {
       return ts.functionCallConfig({ ctx: ts.runtime.ctx });
     }
-    return ts.functionCallConfig({
-      ctx: ts.runtime.ctx,
-      threads: ts.runtime.threads,
-      stateStack: opts?.stateStack ?? ts.id("__stateStack"),
-      ...opts?.extra,
-    });
+    if (opts?.extra && Object.keys(opts.extra).length > 0) {
+      return ts.obj(opts.extra);
+    }
+    return undefined;
   }
 
 
@@ -1909,7 +1929,7 @@ export class TypeScriptBuilder {
     node: FunctionCall,
     functionName: string,
     shouldAwait: boolean,
-    options?: { stateStack?: TsNode },
+    _options?: { stateStack?: TsNode },
   ): TsNode {
     const descriptor = this.buildCallDescriptor(node);
 
@@ -1919,7 +1939,6 @@ export class TypeScriptBuilder {
       stepPath: ts.str(this.steps.joined()),
     } : undefined;
     const configObj = this.buildStateConfig({
-      stateStack: options?.stateStack,
       extra: locationOpts,
     });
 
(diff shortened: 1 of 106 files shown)


packages/agency-lang/lib/backends/typescriptBuilder/pipeChainEmitter.ts
```diff
@@ -250,14 +262,17 @@ export class PipeChainEmitter {
           : null;
       if (propName) {
         const descriptor = ts.obj({ type: ts.str("positional"), args: ts.arr([pipeArg]) });
-        return ts.await(ts.call(
-          ts.id("__callMethod"),
-          [receiver, ts.str(propName), descriptor, this.deps.buildStateConfig()],
-        ));
+        const config = this.deps.buildStateConfig();
```
egonSchiele, line 265: repeated pattern; worth adding a new TS IR builder for this?

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -1503,7 +1523,7 @@ export class TypeScriptBuilder {
     // Create runner for step execution
     setupStmts.push(
       ts.raw(
-        `const runner = new Runner(__ctx, __stack, { state: __stack, moduleId: ${JSON.stringify(this.moduleId)}, scopeName: ${JSON.stringify(functionName)}, stack: __stateStack, threads: __threads });`,
+        `const runner = new Runner(__ctx, __stack, { state: __stack, moduleId: ${JSON.stringify(this.moduleId)}, scopeName: ${JSON.stringify(functionName)} });`,
```
egonSchiele, line 1526: worth creating a new TS IR builder?

packages/agency-lang/lib/runtime/hooks.ts
```diff
@@ -218,14 +219,19 @@ export function gatherCallbacks<K extends keyof CallbackMap>(
...
  *  per-tool `onToolCallStart` / `onToolCallEnd` in `prompt.ts`). The
- *  public `callHook` is now a thin wrapper that omits `stateStack`. */
+ *  public `callHook` is now a thin wrapper that omits `stateStack`.
+ *
+ *  `ctx` is optional — when omitted (which is now the case for every
+ *  codegen-emitted `callHook(...)` site), it's resolved from the active
+ *  ALS frame via `getRuntimeContext()`. */
 export async function invokeCallbacks<K extends keyof CallbackMap>(args: {
-  ctx: RuntimeContext<any>;
+  ctx?: RuntimeContext<any>;
```
egonSchiele, line 228: where does ctx still need to get passed in?

packages/agency-lang/lib/runtime/interrupts.ts
```diff
@@ -286,15 +288,26 @@ async function runResumeLoop(
   agentStartTime: number,
 ): Promise<any> {
   let nodeName = startNodeName;
+  // Seed an ALS frame so stdlib helpers and `callHook` reads (which
```
egonSchiele, line 291: what are the side effects of doing this? we could also move this callback or remove it entirely if this is going to cause other bugs

packages/agency-lang/lib/runtime/node.ts
```diff
@@ -144,16 +144,31 @@ export async function runNode({
   }
 
   const execCtx = await ctx.createExecutionContext(runId);
+  // initializeGlobals + registerTopLevelCallbacks both invoke Agency
+  // code that goes through `__call` — and `__call` reads `ctx` /
+  // `threads` / `stateStack` from the ALS frame after the
+  // drop-per-call-context-plumbing migration. Without an ALS frame
+  // installed here, calls to user-defined stdlib helpers (e.g. the
```
egonSchiele, line 151: could you explain the consequences of doing this a bit more?

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -252,33 +253,35 @@ async function _runPrompt({
...
-    state: args.stateStack
-      ? {
-        stateStack: args.stateStack,
-        ctx: args.ctx,
-        threads: new ThreadStore(),
-      }
-      : undefined,
+    state: {
+      stateStack: alsStack,
+      ctx,
```
egonSchiele, line 282: ctx or `alsCtx`?

PR #204, opened 2026-05-26
codegen: remove class syntax from Agency (keep new Foo())

egonSchiele:
## Summary

Drops Agency's `class Foo { ... }` definition syntax — fields, methods, `extends`, auto-generated constructors, the typechecker / preprocessor branches that handled them, and the codegen / templates that emitted them. **`new Foo(args)` expression syntax is kept** for instantiating JS classes imported from TypeScript (e.g. `new ThreadStore()`, `new RuntimeContext()`).

## What got removed

- `ClassDefinition`, `ClassField`, `ClassMethod` AST types.
- `classParser` and friends in `lib/parsers/parsers.ts`.
- `lib/backends/typescriptBuilder/classEmitter.ts`.
- `lib/templates/backends/typescriptGenerator/classDefinition.{mustache,ts}` and `classMethod.{mustache,ts}`.
- `compilationUnit.classDefinitions` + the build-loop branch.
- Class case branches in the typechecker (`shadowing.ts`, `resolveVariable.ts`, `resolveCall.ts`, `undefinedFunctionDiagnostic.ts`, `undefinedVariableDiagnostic.ts`, `checker.ts`) and the preprocessors (`injectSchemaArgs.ts`, `liftCallbacks.ts`, `parallelDesugar.ts`, `typescriptPreprocessor.ts`).
- LSP class-symbol handling (`semantics.ts`, `documentSymbol.ts`, `foldingRange.ts`, `completion.ts`, `workspaceSymbol.ts`).
- `lib/runtime/classReviver.ts` and the `RuntimeContext.classRegistry` / `registerClass` plumbing. The only remaining caller (`RuntimeContext.restoreState`) now uses a small inline `reviveNative` JSON round-trip.
- `SymbolInfo`'s `ClassSymbol` variant and the `kind === "class"` re-export plumbing in `symbolTable.ts` / `resolveReExports.ts`.
- `isClassKeyword` (`this` / `super`) bypasses in the typescript preprocessor and codegen.
- `tests/agency/classes/` (20 .agency + .test.json triplets) and two `tests/typescriptGenerator/class-*` fixtures.
- `docs/site/guide/classes.md`.

## What got added / kept

- New `lib/types/newExpression.ts` houses the `NewExpression` AST type that used to be co-located with `ClassDefinition`. All importers updated.
- New `reservedClassParser` at the statement level fires a fatal-ish parse failure when it sees `class <Ident>` — so users migrating off the feature see something actionable instead of `class` silently being parsed as an identifier and producing a tangential error.
- `newExpressionParser` + the `processNewExpression` codegen path are unchanged — `new ThreadStore()` etc. still work.

## Validation

- `pnpm tsc --noEmit` — clean.
- `pnpm test:run` — all 4404 unit tests pass.
- `make` — stdlib + agents recompile cleanly.
- Smoke test: `node main(): number { let store = new ThreadStore(); return 1 }` compiles and runs.
- Tests that previously used Agency `class` syntax incidentally (e.g. `pipeReceiverCodegen.test.ts`) were rewritten to use object literals + `def` callables. Tests that exercised class-specific behavior (`Class Definitions` describe in `agencyGenerator.test.ts`, the class-method safe-flag case in `typescriptBuilder.integration.test.ts`, the class-shadowing case in `jsGlobalsSig.test.ts`, the `new MyClass()` unresolved-name case in `undefinedVariableDiagnostic.test.ts`, the class-related cases in `compilationUnit.test.ts` / `injectSchemaArgs.test.ts` / `function.test.ts`) were dropped.

## Rationale

The user-facing class feature was almost entirely unused inside the project — zero hits in `stdlib/`, `lib/agents/`, `examples/`, or `docs/site/stdlib/`. Every piece of the rest of the codebase (typechecker, preprocessor, codegen, runtime) carried a parallel branch for it. Removing it simplifies the upcoming `__ctx` ALS accessor migration: the `classMethod` template carried its own `const __ctx = __state?.ctx || __globalCtx;` rebind that was the largest complication blocking that work.

## Plan

`docs/superpowers/plans/2026-05-26-remove-class-syntax.md`.


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3641,134 +3641,28 @@ export const graphNodeParser: Parser<GraphNodeDefinition> = label("a node defini
...
-      "_extends",
-    ),
-    char("{"),
-    optionalSpacesOrNewline,
-    capture(many(classBodyMemberParser), "members"),
-    optionalSpacesOrNewline,
-    char("}"),
-    optionalSpacesOrNewline,
+  return failure(
+    "`class` definitions are no longer supported in Agency. " +
```
egonSchiele: "no longer supported" – implies that users will have knowledge of this change... that some users have used this before and are no longer able to use it. Agency doesn't have any users right now, so you don't need to give this historical context. You can simply say "are not supported".

PR #205, opened 2026-05-26
codegen: migrate __ctx to ALS accessor (Phase 4 follow-up)

egonSchiele:
## Summary

Migrates the `__ctx` codegen from a per-scope local const to the `__ctx()` accessor that reads from the active `agencyStore` ALS frame. This mirrors the prior `__threads()` and `__stateStack()` migrations completed in earlier Phase 4 PRs.

## Approach (Approach B from the design discussion)

- `ts.runtime.ctx` is now a `TsRaw` accessor call: `__ctx()`
- Pre-wrap, parameter-context, and seed sites use `ts.id("__ctx")` explicitly (the lexical local emitted by `setupEnv`)
- Inside-wrap dereferences go through the strict `getRuntimeContext().ctx` accessor — a missing frame throws a dedicated error pointing to `runInTestContext` instead of a cryptic `Cannot read 'X' of undefined`

## Template migrations

- `interruptReturn.mustache`, `interruptAssignment.mustache`, `functionCatchFailure.mustache` — `__ctx.X` derefs migrated to `getRuntimeContext().ctx.X`
- `blockSetup.mustache` / `forkBlockSetup.mustache` — kept the closure-captured lexical `__ctx` for the Runner ctor + stateStack push/pop symmetry (the block's arrow function closes over the outer function's `__ctx2`)
- `resultCheckpointSetup.mustache` — kept the lexical because this template is emitted **before** the per-scope `withAlsFrame` wrap. Using the strict accessor here would throw at function entry
- `debugger.mustache` — no changes (only passes `__ctx` as an arg, no derefs)

## Docstring interpolation

Tool descriptions like `"This tool is version ${toolVersion}."` are evaluated eagerly at module load, before any `agencyStore.run(...)` frame is installed. The codegen now flips a new `topLevel: true` flag on every `TsScopedVar` in the description subtree via `markTopLevelScopedVars`; the pretty-printer routes those to `__globalCtx.globals.get(...)` instead of the strict accessor.

Fixed a bug where the marker walker didn't recurse into `TsTemplatePart` carriers (which lack a `kind` field), so descriptions with template-literal interpolations were silently falling back to the strict accessor and crashing at module load.

Removed the top-level `const __ctx = __globalCtx;` rebind that used to support docstring interpolation — it would now shadow the `__ctx()` runtime import and break every accessor call.

## Other touch-ups

- `globalGet` / `globalSet` got an optional `ctxRef` override so `sectionAssembler.ts` can pass `ts.id("__ctx")` for the `__initializeGlobals(__ctx)` parameter context
- `withHandler` (top-level handler wrap) uses `getRuntimeContext().ctx` — it always runs under `runInBootstrapFrame`, and handlers are safety infrastructure so failing loud here is preferable to silent skipping
- `typescriptPreprocessor.ts` `awaitPending` emission uses `getRuntimeContext().ctx.pendingPromises.awaitPending(...)`

## Verification

- All 4408/4408 unit tests pass
- Fixtures regenerated (`make fixtures`)
- Smoke tests verified end-to-end:
  - Fork-with-interrupt (`tests/agency/fork/fork-deep-call-interrupt.agency`) — exercises the migrated `interruptReturn` template + ALS frame propagation across fork branches
  - Docstring interpolation runtime test — confirms eager-evaluated descriptions read through `__globalCtx`
- `make` builds the project + stdlib without errors

## Files

- IR / pretty printer: `lib/ir/builders.ts`, `lib/ir/prettyPrint.ts`, `lib/ir/tsIR.ts`
- Codegen: `lib/backends/typescriptBuilder.ts`, `lib/backends/typescriptBuilder/sectionAssembler.ts`, `lib/preprocessors/typescriptPreprocessor.ts`
- Templates: `lib/templates/backends/typescriptGenerator/*.mustache` (+ regenerated `.ts`)
- Imports: added `__ctx` to `imports.mustache`
- Tests: updated `prettyPrint.test.ts` and `docstringInterpolationRuntime.test.ts` expectations
- Fixtures: 100+ generator/builder/debugger fixtures regenerated


egonSchiele:
Great catch by Copilot — this was a real bug, not a low-confidence one.

The `__ctx()` accessor call (emitted by `ts.runtime.ctx` for `goToNode`'s `ctx:` field) was getting shadowed by the `const __ctx = __state?.ctx || __globalCtx;` local emitted by setupEnv. esbuild's TS transform renames the local AND every `__ctx` reference in the same scope (including the import-bound accessor call) to a non-colliding name like `__ctx2` — so `__ctx()` became `__ctx2()`, and since `__ctx2` is a RuntimeContext object (not a function), it crashed with `__ctx2 is not a function`.

Caught the bug by running `tests/agency/goto.agency` which exercises `goToNode` — failed with exactly this error.

Fixed in fc7467d:
- `ts.runtime.ctx` now emits `getRuntimeContext().ctx` instead of `__ctx()` (avoids the shadow entirely)
- Removed the now-unused `__ctx` import from `imports.mustache`

All 4408/4408 unit tests pass and the previously broken goto test now passes.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -142,6 +142,48 @@ const COMPOUND_RUNNER_KINDS: ReadonlySet<TsNode["kind"]> = new Set([
...
+ * any `agencyStore.run(...)` wrap installs an ALS frame. The pretty
+ * printer reads `topLevel` to emit `__globalCtx.globals.get(...)`
+ * instead of `getRuntimeContext().ctx.globals.get(...)` — the strict
+ * accessor would throw at module load.
+ *
+ * Generic walker: visits every object-valued field whose value carries
+ * a `kind` discriminator, plus arrays of such values. Returns a
+ * structurally cloned tree (does NOT mutate the input).
+ */
+function markTopLevelScopedVars<T extends TsNode>(node: T): T {
```
egonSchiele, line 160: This belongs in the preprocessor

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -934,7 +979,10 @@ export class TypeScriptBuilder {
     if (scope.type === "function") {
       args.push(
         ts.obj({
-          checkpoint: ts.raw("__ctx.getResultCheckpoint()"),
+          // Inside the function body's withAlsFrame wrap — strict
+          // accessor so a missing frame throws cleanly instead of a
+          // generic "Cannot read 'getResultCheckpoint' of undefined".
+          checkpoint: ts.raw("getRuntimeContext().ctx.getResultCheckpoint()"),
```
egonSchiele, line 985: lots of `ts.raw` references to `getRuntimeContext().ctx`, create a ts ir builder for this pattern instead

PR #206, opened 2026-05-27
codegen: drop trailing __state param from generated functions

egonSchiele:
Eliminates the legacy `__state: InternalFunctionState | undefined` trailing positional from every generated Agency function. After this PR, function signatures are just `async function __foo_impl(a, b)` — `ctx` / `threads` / `stateStack` flow through the active `agencyStore` AsyncLocalStorage frame seeded by the parent scope.

Last PR in the three-part series:
1. `2026-05-26-remove-class-syntax.md` (merged)
2. `2026-05-26-als-migration-ctx-accessor.md` (merged as #205)
3. `2026-05-26-drop-state-param.md` ← this PR

### Codegen / templates

- `lib/backends/typescriptBuilder.ts`
  - Function-decl emission drops the trailing `__state: InternalFunctionState | undefined = undefined` param.
  - `setupFunction()` is invoked with **no arguments** — it reads `ctx` / `threads` from `getRuntimeContext()`.
  - The setup-block `ctx:` arg switches from `__state?.ctx || __globalCtx` to `getRuntimeContext().ctx` for both function bodies and graph-node bodies.
  - `setupNode` still receives `__state: GraphState` because the node-arrow signature carries `data` / `isResume` / `messages` from the simplemachine — but `ctx` is read from ALS, not from `state.ctx`.
- `lib/templates/backends/typescriptGenerator/blockSetup.mustache` — no longer plumbs `state: { ctx, threads }` into `setupFunction`.
- `lib/templates/backends/typescriptGenerator/debugger.mustache` — drops the `__state` arg from the `debugStep(__ctx, …)` call.
- `lib/templates/backends/typescriptGenerator/imports.mustache` — drops the now-dead `InternalFunctionState` import.

### Runtime

- `lib/runtime/call.ts` — `__call` / `__callMethod` stop merging an ALS bag into state extras for every call. Caller-passed extras are forwarded to `AgencyFunction.invoke(...)` as-is. **Two narrow special cases remain (Option C):**
  1. **Checkpoint location info** — codegen emits `{ moduleId, scopeName, stepPath }` at every `checkpoint()` site so the created checkpoint records the caller's source location.
  2. **Async-fork `stateStack` override** — `__call` installs a per-branch ALS frame whose `stack` is the forked branch's stack so the callee resolves stack-reads to the branch stack. The async-unassigned operator is slated for removal in favor of `fork` / `parallel`, which carry their own per-branch frames via `runBatch.runInBranchAlsFrame`.
- `lib/runtime/checkpoint.ts` — `checkpoint()` / `getCheckpoint()` / `restore()` read `ctx` / `stateStack` from `getRuntimeContext()`. `checkpoint()` keeps a trailing `__state?` positional that now carries ONLY the location fields. `getCheckpoint` / `restore` drop the trailing state arg entirely (`AgencyFunction.invoke` still silently passes one through; it's ignored).
- `lib/runtime/ipc.ts` — `_run` reads `ctx` and the per-scope `stateStack` from ALS.
- `lib/runtime/debugger.ts` — `debugStep` drops its previously-unused `state` param.
- `lib/stdlib/agency.ts` — `_callbackImpl` reads `ctx` from ALS.
- `lib/runtime/node.ts`
  - `setupFunction()` takes no args. The previous "called as a tool by the LLM with no state" fallback (fresh `StateStack` + empty `ThreadStore`) is no longer reachable because tool dispatch always runs inside the issuing `runner.step` ALS frame.
  - `setupNode` reads `ctx` from ALS instead of `state.ctx`.
- `lib/runtime/prompt.ts`, `lib/runtime/runner.ts` — internal callers updated to the new no-arg signatures.

### Type-system cleanup

- `InternalFunctionState` type deleted from `lib/runtime/types.ts` along with its export from `lib/runtime/index.ts`. Dead `SimpleMachine` / `StatelogClient` imports in `types.ts` removed.

### Tests

Direct-call test sites that exercise runtime helpers without going through `runNode` now wrap their invocations in `runInTestContext(...)` so `getRuntimeContext()` finds an active ALS frame:

- `lib/runtime/__tests__/node.test.ts`
- `lib/runtime/checkpoint.test.ts`
- `lib/runtime/debugger.test.ts`
- `lib/runtime/state/stateStack.test.ts`

### Fixtures

`make fixtures` regenerated every `tests/typescriptGenerator/*.mjs`, `tests/typescriptBuilder/*.mjs`, and `tests/debugger/*.ts` fixture to match the new codegen output.

## End-state shape

```ts
// Before
async function __foo_impl(
  a: number, b: number,
  __state: InternalFunctionState | undefined = undefined,
) {
  const __setupData = setupFunction({ state: __state });
  // ...
  const __ctx = __state?.ctx || __globalCtx;
  // ...
}

// After
async function __foo_impl(a: number, b: number) {
  const __setupData = setupFunction();
  // ...
  const __ctx = getRuntimeContext().ctx;
  // ...
}
```

## Verification

- `pnpm tsc --noEmit` — clean
- `pnpm test:run` — **4408 / 4408** vitest tests pass
- Smoke tests (`pnpm run agency test ...`):
  - `tests/agency/fork/fork-interrupt.test.json` — 1/1
  - `tests/agency/interrupts/interrupt.test.json` — 3/3
  - `tests/agency/threads/messages.test.json` — 1/1
- CLI integration: `npm pack` + `node tests/integration/cli/test.mjs ./agency-lang-*.tgz` — all 4 scenarios pass

## Breaking changes for external integrations

- Anyone calling `compiled.__foo_impl(args, customState)` directly from JS gets a runtime "called outside an Agency execution frame…" error unless they wrap in `runInTestContext(ctx, stack, threads, () => compiled.__foo_impl(args))`.
- `setupFunction(...)` and `setupNode(...)` API change — callers outside generated code must install an ALS frame first.
- Anyone importing `InternalFunctionState` from `agency-lang/runtime` will see a missing-symbol error; the type is gone.

## Open caveats

- The legacy `async helper()` unassigned-async operator's branch-stack isolation now relies on `__call`'s per-call ALS-frame install (not on a positional override). Single-helper async-call sites still work. Per the plan, this operator is slated for deletion in favor of `fork` / `parallel`.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -1500,23 +1500,23 @@ export class TypeScriptBuilder {
       argsObj[arg] = ts.id(arg);
     }
 
-    // Setup block
+    // Setup block. `setupFunction()` reads `ctx` / `threads` from the
+    // active `agencyStore` ALS frame seeded by the caller (a
+    // `runner.step` body, `runNode`'s top-level frame, or
+    // `runBatch.runInBranchAlsFrame`). Tool dispatch by an LLM also runs
+    // inside the issuing `runner.step` frame, so a frame is always
+    // active here.
     const setupStmts: TsNode[] = [
       ts.constDecl(
         "__setupData",
-        $(ts.id("setupFunction"))
-          .call([ts.obj({ state: ts.runtime.state })])
-          .done(),
+        $(ts.id("setupFunction")).call([]).done(),
       ),
 
-      ts.comment(
-        "__state will be undefined if this function is being called as a tool by an llm",
-      ),
       ts.setupEnv({
         stack: $(ts.id("__setupData")).prop("stack").done(),
         step: $(ts.id("__setupData")).prop("step").done(),
         self: $(ts.id("__setupData")).prop("self").done(),
-        ctx: ts.raw("__state?.ctx || __globalCtx"),
+        ctx: ts.raw("getRuntimeContext().ctx"),
       }),
 
       // Ensure this module's globals are initialized on the current ctx.
@@ -1562,8 +1562,8 @@ export class TypeScriptBuilder {
     );
 
     // Create runner for step execution. `threads` is read directly from
-    // the setup-function result (a fresh ThreadStore when called outside
-    // any node, or the per-node store threaded in via the InternalFunctionState).
+    // the setup-function result, which now resolves it from the active
+    // ALS frame instead of an `__state` positional.
     setupStmts.push(
       ts.raw(
         `const runner = new Runner(__ctx, __stack, { state: __stack, moduleId: ${JSON.stringify(this.moduleId)}, scopeName: ${JSON.stringify(functionName)}, threads: __setupData.threads });`,
@@ -1694,7 +1694,9 @@ export class TypeScriptBuilder {
     this.scopes.inSafeFunction = prevSafe;
     this.scopes.pop();
 
-    // Build function params: typed args + __state
+    // Build function params from the source signature. The legacy
+    // trailing `__state: InternalFunctionState | undefined` positional
+    // is gone — `ctx` / `threads` / `stateStack` flow through ALS.
     const fnParams: TsParam[] = parameters.map((p) => {
       const baseType = p.typeHint ? formatTypeHintTs(p.typeHint) : "any";
       if (p.defaultValue) {
@@ -1705,11 +1707,6 @@ export class TypeScriptBuilder {
         };
       }
       return { name: p.name, typeAnnotation: baseType };
-    });
-    fnParams.push({
-      name: "__state",
-      typeAnnotation: "InternalFunctionState | undefined",
-      defaultValue: ts.id("undefined"),
     });
 
     const implName = `__${functionName}_impl`;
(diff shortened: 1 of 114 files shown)


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2270,7 +2267,11 @@ export class TypeScriptBuilder {
         stack: $(ts.id("__setupData")).prop("stack").done(),
         step: $(ts.id("__setupData")).prop("step").done(),
         self: $(ts.id("__setupData")).prop("self").done(),
-        ctx: $(ts.runtime.state).prop("ctx").done(),
+        // `runNode` (and the resume / rewind variants) install the
+        // top-level `agencyStore` frame before `graph.run` invokes
+        // this node body. Read `ctx` from ALS so the local matches
+        // the same per-run context that `setupNode` just used.
+        ctx: ts.raw("getRuntimeContext().ctx"),
```
egonSchiele, line 2274: Lots of `ts.raw` references to `getRuntimeContext().ctx`. `ts.raw` is a last-ditch option, overuse of it is an anti-pattern.

packages/agency-lang/lib/runtime/call.ts
```diff
@@ -1,31 +1,47 @@
...
 import { agencyStore } from "./asyncContext.js";
+import type { StateStack } from "./state/stateStack.js";
 
 /**
- * Construct the `__state` bag that flows into AgencyFunction.invoke() from
- * the active ALS frame, merged with any caller-provided extras (e.g.
- * `moduleId`/`scopeName`/`stepPath` for checkpoint sites, or `{ ctx }`
- * when called from inside `__initializeGlobals` which runs before the
- * ALS frame is installed).
+ * Post-`__state`-drop migration: generated functions read `ctx` /
```
egonSchiele, line 7: Please look through these large comments you are adding. Is this information that will be useful in the future? Or are these essentially just comments you are adding that are only useful context for this specific PR? When you add comments like

```
`threads` / `stateStack` from the active `agencyStore` frame and no longer accept a trailing `__state` positional. The runtime call
```

...great, so they won't accept a trailing `__state` positional. People reading this in the future won't even know what the `__state` positional was, so this comment is not relevant information for them.

PR #207, opened 2026-05-27
runtime: move checkpoint location to ALS + drop trailing state arg from invoke/__call

egonSchiele:
## What

Moves the per-call-site checkpoint location into the `agencyStore` ALS frame, then drops the trailing `state` positional from `AgencyFunction.invoke`, `_invokeInner`, `_fn`, `__call`, and `__callMethod`.

Closes the loop on the PR #206 migration: with this change, the only way to plumb a value into a called Agency function is the ALS frame. The "extra positional arg = magic runtime context" pattern that a future editor might extend is gone from the public-ish surfaces.

## Why

Before this PR, `AgencyFunction.invoke(descriptor, state?)` carried a `state?: unknown` trailing arg whose only surviving consumer was the per-call-site checkpoint location `{moduleId, scopeName, stepPath}` that codegen emitted for `checkpoint()`. PR #206 had already cut the codegen's dependency on `__state` inside generated function bodies; the invoke-level positional was vestigial.

Leaving it in would invite the next person reading `state?: unknown` to assume "extra positional arg = runtime plumbing pattern" and extend it. Removing it forces every future caller through the ALS frame, which is the right answer.

## How

Three tasks, all in one commit per the plan's "code + templates" / "fixtures" split:

1. **Task 1 — ALS plumbing** (`asyncContext.ts`, `runner.ts`):
   - New `CallsiteLocation` type + `callsite?` slot on `AgencyStore`.
   - `withCallsite(loc, fn)` overrides the active frame's callsite.
   - `withPushedHandler(ctx, handler, fn)` encapsulates the push/try/finally shared by `AgencyFunction.invoke()`'s preapprove branch and (in a future PR) `agency.withHandler`. Co-locating it here avoids a future circular import between `agencyFunction.ts` and the namespace module.
   - `Runner.runInScope` seeds `callsite: { moduleId, scopeName, stepPath: this.path.join(".") }` on every per-step ALS frame.

2. **Task 2 — checkpoint reads from ALS** (`checkpoint.ts`, `typescriptBuilder.ts`):
   - `checkpoint()` reads `{ moduleId, scopeName, stepPath }` from `getRuntimeContext().callsite` instead of accepting a trailing arg. Missing-callsite paths (bootstrap scope) fall back to the empty `""::""::""` location — same as today's `__state == null` branch.
   - Codegen drops the `checkpoint`-only emission of `{ moduleId, scopeName, stepPath }` extras.

3. **Task 3 — drop the trailing state arg** (`agencyFunction.ts`, `call.ts`, `typescriptBuilder.ts`, plus all in-tree TS callers):
   - `AgencyFunction.invoke()` collapses from a manual `ctx.pushHandler`/try/finally block to a one-line `withPushedHandler` wrap.
   - `_invokeInner()` drops the trailing `state` param and the `_fn(...args, state)` spread.
   - `__call(target, descriptor, optional?)` / `__callMethod(obj, prop, descriptor, optional?)` drop the `stateExtras` positional; the unused `maybeRunWithForkedStack` helper is deleted.
   - Codegen drops every `extras` argument from emitted `__call` / `__callMethod` invocations, including the `insideGlobalInit` branch's `{ ctx: __ctx }` (bootstrap frame seeds ALS already).

### Async-fork operator decision

The `{ stateStack: __forked }` extras path that codegen emits for `async helper()` unassigned calls was the only place where `__call`'s extras did meaningful work at runtime. Rather than carrying the special case across the cleanup, the codegen now wraps that call site in an inline `agencyStore.run` frame:

```ts
// before
await __call(target, descr, { stateStack: __forked });

// after
await agencyStore.run(
  { ...getRuntimeContext(), stack: __forked },
  () => __call(target, descr),
);
```

The callee picks up the branch's isolated stack via `getRuntimeContext()` — same end result, but the ALS frame is the only mechanism.

### Other TS call sites that historically passed `{ ctx, stateStack }` to `invoke()`

- `hooks.ts` `invokeCallback`: when `stateStack` is provided, the callback now runs inside an `agencyStore.run` frame that overrides `stack`.
- `prompt.ts` per-tool-call invocation: wraps in an `agencyStore.run` frame seeded with the per-tool `ThreadStore` and the branch `StateStack`.
- `validateChain.ts`: drops the `{ ctx }` arg entirely — invoke no longer reads it.

## Risk surface — PR #206 regression class

The class of regression that bit PR #206 was preapprove handlers + parallel/fork branches: a bug that skips the handler push (or pop) leaves the stack in an inconsistent state that only surfaces under load. To pin that contract:

- `agencyFunction.test.ts` — three new tests covering `withPushedHandler` correctness for preapproved tools (handler installed during call, popped on normal return, popped on throw, no-op for non-preapproved fns).
- `asyncContext.test.ts` — `withPushedHandler` pops in both finally paths.
- `runner.test.ts` — load-bearing test that `Runner.runInScope` actually seeds the ALS `callsite` slot. Without this, a future regression that deletes the `callsite: { ... }` block from `runInScope` would silently degrade every checkpoint to `""::""::""` and every `withCallsite` test would still pass.

Smoke tests run locally:

- `pnpm run agency test tests/agency/preapprove.test.json` — 5/5 ✓
- `pnpm run agency test tests/agency/fork/` — 67/67 ✓ across 65 files
- `pnpm run agency test tests/agency/interrupts/` — 16/16 ✓ across 11 files

## Validation

- `pnpm tsc --noEmit` ✓
- `pnpm run lint:structure` ✓
- `pnpm test:run` — 4425/4425 ✓
- `make` builds cleanly ✓
- Fixture spot-check: no `__call(..., { moduleId: ... })` and no `__call(..., { stateStack: ... })` remain — confirmed via `grep -nE "__call\(.+, \{" tests/typescriptGenerator/`.

## CHANGELOG note

External callers that imported `__call` / `__callMethod` from `agency-lang/runtime` should drop the `stateExtras` positional from their call sites. The arg was always ignored by `__call` after the ALS migration; this PR removes the type-level signature.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -1993,9 +1993,14 @@ export class TypeScriptBuilder {
     // inline at the call site that overrides `stack`; the callee picks
     // it up via `getRuntimeContext()`.
     if (options?.stateStack) {
-      const wrapped = ts.raw(
-        `agencyStore.run({ ...getRuntimeContext(), stack: ${this.str(options.stateStack)} }, () => ${this.str(callExpr)})`,
-      );
+      const frame = ts.obj([
+        ts.setSpread(ts.call(ts.id("getRuntimeContext"))),
+        ts.set("stack", options.stateStack),
+      ]);
+      const wrapped = ts.call(ts.prop(ts.id("agencyStore"), "run"), [
+        frame,
+        ts.arrowFn([], callExpr),
+      ]);
       return shouldAwait ? ts.await(wrapped) : wrapped;
     }
 
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -849,27 +849,15 @@ export class TypeScriptBuilder {
         case "methodCall": {
           const fnCall = element.functionCall;
           const descriptor = this.buildCallDescriptor(fnCall);
-          const configObj = this.buildStateConfig();
           const callArgs: TsNode[] = [result, ts.str(fnCall.functionName), descriptor];
-          if (element.optional) {
-            callArgs.push(configObj ?? ts.id("undefined"));
-            callArgs.push(ts.bool(true));
-          } else if (configObj) {
-            callArgs.push(configObj);
-          }
+          if (element.optional) callArgs.push(ts.bool(true));
           result = this.awaitChainCall(ts.call(ts.id("__callMethod"), callArgs), element === node.chain[node.chain.length - 1]);
           break;
         }
         case "call": {
           const descriptor = this.buildCallDescriptor(element);
-          const configObj = this.buildStateConfig();
           const callArgs: TsNode[] = [result, descriptor];
-          if (element.optional) {
-            callArgs.push(configObj ?? ts.id("undefined"));
-            callArgs.push(ts.bool(true));
-          } else if (configObj) {
-            callArgs.push(configObj);
-          }
+          if (element.optional) callArgs.push(ts.bool(true));
           result = this.awaitChainCall(ts.call(ts.id("__call"), callArgs), element === node.chain[node.chain.length - 1]);
           break;
         }
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -1004,36 +1004,20 @@ export class TypeScriptBuilder {
   }
 
   /**
-   * Build the state config passed as the third argument to `__call` /
-   * `__callMethod`.
+   * After the ALS migration AND the trailing-`state`-arg drop, every
+   * Agency call site reads `ctx` / `stack` / `threads` / `callsite`
+   * from the active `agencyStore` frame. `__call` / `__callMethod` no
+   * longer accept a state-extras object, and the codegen never emits
+   * one. This helper is kept as a stable seam in case future codegen
+   * needs to attach per-call metadata, but for now it always returns
+   * `undefined` — callers should simply omit the third positional.
    *
-   * After the ALS migration, the runtime helpers read `ctx`/`threads`/
-   * `stateStack` from `agencyStore` (see lib/runtime/call.ts), so call
-   * sites no longer need to thread those locals. This helper returns
-   * `undefined` for the common case — the caller MUST omit the arg
-   * entirely so the emitted code stays minimal.
-   *
-   * It still returns a real object in two situations:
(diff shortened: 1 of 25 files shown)


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2009,28 +1993,24 @@ export class TypeScriptBuilder {
...
+    const callExpr = ts.call(ts.id("__call"), [callee, descriptor]);
+
+    // Async-fork sites need the branch's isolated stack visible to the
+    // callee (so its checkpoints/handlers/etc. push/pop on the branch
+    // stack rather than the parent's). Install a fresh ALS frame
+    // inline at the call site that overrides `stack`; the callee picks
+    // it up via `getRuntimeContext()`.
+    if (options?.stateStack) {
+      const wrapped = ts.raw(
+        `agencyStore.run({ ...getRuntimeContext(), stack: ${this.str(options.stateStack)} }, () => ${this.str(callExpr)})`,
```
egonSchiele: use builder if possible

packages/agency-lang/lib/runtime/agencyFunction.ts
```diff
@@ -120,36 +120,29 @@ export class AgencyFunction {
...
+  async invoke(descriptor: CallType): Promise<unknown> {
+    // Preapprove tools install a handler that auto-approves any
+    // `request_approval` interrupt raised while the tool body runs.
+    // The handler must be pushed onto the same `ctx.handlers` chain
+    // the invocation consults, so we read `ctx` from the live ALS
+    // frame (seeded by the caller's `runner.step` body — or by the
+    // bootstrap frame for module-init paths). When no frame is
+    // installed (the function was called from a non-Agency context),
+    // skip the push/pop dance — no preapprove machinery to wire up.
+    const ctx = this._isPreapproved ? agencyStore.getStore()?.ctx : undefined;
```
egonSchiele: lets just remove the preapprove functionality, I don't like this code branching

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -475,9 +475,18 @@ export async function runPrompt(args: {
       try {
         const toolThreads = new ThreadStore();
         toolThreads.setStatelogClient(ctx.statelogClient);
-        toolResult = await handler.invoke(
-          { type: "named", positionalArgs: [], namedArgs },
-          { ctx, threads: toolThreads, stateStack: branchStack },
+        // Install a per-tool ALS frame with the tool's isolated
+        // ThreadStore and the branch's StateStack. The handler body
+        // reads these via `getRuntimeContext()`; without the wrap, it
+        // would see whatever frame the prompt loop is running in.
```
egonSchiele: "would see whatever frame the prompt loop is running in" -- isn't that the correct behavior?

PR #208, opened 2026-05-27
runtime: introduce public agency.* namespace

egonSchiele:
## Summary

Introduces a public `agency.*` namespace exported from
`agency-lang/runtime` — the canonical entry point for TS code that
wants to participate in an Agency run. Purely additive: no fixture
changes, no behavior changes.

This is the first of three planned API PRs for the "TS helpers as
first-class participants" effort, and combines Plan 2 (namespace)
with the Task 4 cleanup from Plan 1. The Task 4 cleanup grep
returned empty (`CheckpointLocation`, `maybeRunWithForkedStack`,
`stateExtras`, `state?: unknown`) — Plan 1's PR already removed
them — so the diff here is purely the namespace.

## Surface

```ts

// context reads
agency.ctx() / ctxMaybe()
agency.callsite()
agency.global<T>(name, moduleId?)

// thread subnamespace
agency.thread.current()
agency.thread.user(content) / system(content) / assistant(content)
agency.thread.store() / storeMaybe()
agency.thread.with(threadId, fn)

// checkpoints
agency.checkpoint()
agency.getCheckpoint(id)
agency.restore(id | cp, opts?)
agency.withCallsite(loc, fn)

// handlers / guards
agency.withHandler(handler, fn)
agency.withCostGuard(maxCost, fn)

// test helpers
agency.withTestContext({ctx,stack,threads}, fn)
```

## Design decisions

- **Thread subnamespace from day one.** `agency.thread.*` keeps the
  top-level surface lean and avoids a breaking re-namespacing later
  once the cluster grows (`current` / `user` / `system` /
  `assistant` / `store` / `storeMaybe` / `with` is already 7
  entries). Adding a top-level `agency.userMessage` and migrating
  to `agency.thread.user` later would be a breaking change.

- **No `tool` helper in v1.** `smoltalk.toolMessage(...)` requires a
  `tool_call_id`, which TS helpers rarely have on hand (those are
  emitted by LLM responses). Authors who really need to push a tool
  message can do `agency.thread.current().push(smoltalk.toolMessage(...))`
  directly. Easy to add later if real demand shows up.

- **No `agency.llm` or `agency.withResumableScope` yet.** Those
  land in their own PRs (Plans 3 and 4). Adding them here would
  inflate review surface and conflict with the resumable-scope
  design that is still in flight.

- **Existing exports stay.** `getRuntimeContext`, `agencyStore`,
  `__threads`, `__stateStack`, `__ctx`, `runInTestContext` are still
  re-exported from `lib/runtime/index.ts` because every generated
  `prog.ts` imports them by name. They are now marked in JSDoc as
  the codegen-internal surface — TS helper authors should prefer
  `agency.*`. A separate follow-up could regenerate fixtures to
  drop these from the public API entirely, but that is not "purely
  additive" and out of scope here.

- **`withHandler` delegates to the existing `withPushedHandler`**
  primitive in `asyncContext.ts` — no duplicated push/try/finally
  dance, same combinator as `AgencyFunction.preapprove()`'s factory.

## Tests

`lib/runtime/agency.test.ts` — 24 tests, one per method, including
throw-path restoration for `withHandler` / `withCostGuard` /
`thread.with` and a checkpoint round-trip.

## Validation

- `pnpm tsc --noEmit` clean
- `pnpm test:run` — 4449 / 4449 passing (4425 baseline + 24 new)
- `git status` — only `lib/runtime/agency.ts`, `lib/runtime/agency.test.ts`,
  `lib/runtime/index.ts` changed; no fixtures touched

## Naming bikeshedding

The spec uses `agency`, which overloads with "Agency the language".
Alternatives considered: `runtime`, `lang`, `rt`. Happy to rename if
you'd prefer something else now — retrofitting is breaking, so it's
worth deciding before merge.

### DIFF
--- a/packages/agency-lang/lib/runtime/agency.ts
+++ b/packages/agency-lang/lib/runtime/agency.ts
@@ -1,0 +1,204 @@
+/**
+ * Public `agency.*` namespace — the canonical entry point for TS code
+ * that wants to participate in an Agency run (read context, push
+ * thread messages, install handlers, create checkpoints, etc.).
+ *
+ * Every method is a thin wrapper over an existing runtime function;
+ * no new behavior is added here. The namespace exists for
+ * discoverability and to give docs/IDEs a single surface to point at.
+ *
+ * Naming conventions:
+ *  - No underscore prefix. Read APIs throw on missing frame by
+ *    default; `*Maybe` variants return `undefined` instead.
+ *  - `with*` prefix for scope-installing helpers (run a callback with
+ *    a temporary modification to the active frame).
+ *  - Thread-related operations live under `agency.thread.*` to keep
+ *    the top-level surface lean; e.g. `agency.thread.user("hi")` and
+ *    `agency.thread.current()`.
+ *  - Codegen-emitted internals (`getRuntimeContext`, `agencyStore`,
+ *    `__threads`, `__stateStack`, `__call`, `__callMethod`,
+ *    `runInTestContext`) keep their existing names and are still
+ *    exported from `agency-lang/runtime` because generated code
+ *    imports them directly. TS helper authors should prefer
+ *    `agency.*`.
+ *
+ * See docs/site/guide/ts-helpers.md (PR #5) for the long-form guide.
+ */
+import * as smoltalk from "smoltalk";
+import {
+  agencyStore,
+  getRuntimeContext,
+  runInTestContext,
+  withCallsite as _withCallsite,
+  withPushedHandler,
+  type CallsiteLocation,
+} from "./asyncContext.js";
+import {
+  checkpoint as _checkpoint,
+  getCheckpoint as _getCheckpoint,
+  restore as _restore,
+} from "./checkpoint.js";
+import type { RestoreOptions } from "./errors.js";
+import { CostGuard } from "./guard.js";
+import type { Checkpoint } from "./state/checkpointStore.js";
+import type { RuntimeContext } from "./state/context.js";
+import type { StateStack } from "./state/stateStack.js";
+import type { MessageThread } from "./state/messageThread.js";
+import type { ThreadStore } from "./state/threadStore.js";
+import type { HandlerFn } from "./types.js";
+
+// ---- Context reads -----------------------------------------------------
+
+/** Read the active `RuntimeContext`. Throws when called outside any
+ *  `agencyStore.run(...)` frame (i.e. from non-Agency code). Tests
+ *  that need a frame can use `agency.withTestContext({ctx,stack,threads}, fn)`. */
+const ctx = (): RuntimeContext<any> => getRuntimeContext().ctx;
+
+/** Lax variant of `agency.ctx()`. Returns `undefined` outside any frame. */
+const ctxMaybe = (): RuntimeContext<any> | undefined => agencyStore.getStore()?.ctx;
+
+/** Per-call-site source location seeded by `Runner.runInScope` for
+ *  every step body. Returns `undefined` outside any frame or in
+ *  frames where no callsite was installed (bootstrap scope). */
+const callsite = (): CallsiteLocation | undefined => agencyStore.getStore()?.callsite;
+
+/** Read a module-scoped global. Same semantics as the Agency-level
+ *  `globals.get(moduleId, name)`. `moduleId` defaults to `""`
+ *  (the bare/anonymous module). */
+const global_ = <T = unknown>(name: string, moduleId = ""): T =>
+  ctx().globals.get(moduleId, name) as T;
+
+// ---- Thread subnamespace ----------------------------------------------
+
+/** Active `MessageThread`, creating one if none is active yet. */
+const threadCurrent = (): MessageThread =>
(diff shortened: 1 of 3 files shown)


packages/agency-lang/lib/runtime/agency.ts
```diff
@@ -0,0 +1,204 @@
...
+/** Restore execution to a prior checkpoint. Throws `RestoreSignal`;
+ *  the surrounding runtime catches it and rewinds. */
+const restore = (
+  idOrCp: number | Checkpoint,
+  opts: RestoreOptions = {},
+): void => _restore(idOrCp, opts);
+
+/** Run `fn` with a custom `callsite` installed on the active frame.
+ *  Throws when called outside any agency frame. */
+const withCallsite = <T>(loc: CallsiteLocation, fn: () => T): T =>
```
egonSchiele, line 148: what is a call site? When would users use this?

packages/agency-lang/lib/runtime/agency.ts
```diff
@@ -0,0 +1,204 @@
...
+/** Push `handler` onto `ctx.handlers` for the duration of `fn`; pop in
+ *  finally. Thin wrapper over the shared `withPushedHandler` primitive
+ *  in `asyncContext.ts` so user code and `AgencyFunction`'s preapprove
+ *  factory go through the same encapsulated combinator. */
+const withHandler = <T>(
+  handler: HandlerFn,
+  fn: () => Promise<T>,
+): Promise<T> => withPushedHandler(ctx(), handler, fn);
+
+/** Install a `CostGuard(maxCost)` on the active branch's
```
egonSchiele: are you also planning to add `withTimeGuard`? Or that not needed?

packages/agency-lang/lib/runtime/agency.ts
```diff
@@ -0,0 +1,204 @@
...
+  }
+};
+
+// ---- Test helpers ------------------------------------------------------
+
+/** Install an ALS frame from explicit `{ctx, stack, threads}` for
+ *  tests that exercise stdlib helpers directly. Mirrors
+ *  `runInTestContext` with an object-arg signature so it composes
+ *  with the rest of the namespace. */
+const withTestContext = <T>(
```
egonSchiele, line 225: Would this ever be used by users, or only in tests? If it will be used by users, I think it needs a different name, to make it clear that this is not just to be used in a test context. If it will not be used by users, for any of these functions that are only for internal functionality, we should document them as internal when you write the documentation.

packages/agency-lang/lib/runtime/agency.ts
```diff
@@ -0,0 +1,204 @@
...
+    with: threadWith,
+  },
+
+  checkpoint,
+  getCheckpoint,
+  restore,
+  withCallsite,
+
+  withHandler,
+  withCostGuard,
```
egonSchiele, line 260: I think `addCost` would be another useful function, for users who use their own LLM function instead of the one built into agency, but still want to track the cost and have the cost guard fire if the cost is too high.

egonSchiele:
Pushed c4a0079a addressing all review feedback.

**Bug fix (Copilot, withCostGuard L156):** real bug — switched to `getRuntimeContext().stack` (ALS-resolved per-branch stack) so the guard cannot leak across fork/race siblings. Added a per-branch isolation test that constructs a distinct `branchStack` and asserts the guard lands there, not on `ctx.stateStack`. Tests updated accordingly (Copilot L251).

**Naming convention (Copilot L17):** kept `callsite()` returning `undefined` (callsite is intrinsically optional — bootstrap frames omit it even when there's a valid frame, so `Maybe` would over-promise). The header JSDoc now calls out `callsite()` as the lone exception so the convention is no longer silently violated.

**Dead doc link (Copilot L25):** removed the `docs/site/guide/ts-helpers.md` reference.

**`thread.with` sync widening (Copilot L110):** widened to `fn: () => T | Promise<T>` returning `Promise<T>`.

**`withCallsite` docs (Egon L131):** rewrote the JSDoc to explain what a callsite is (the source location `Runner.runInScope` seeds per step to attribute checkpoints), when users would need to override it (rarely — only TS helpers subdividing their own work into substeps for debugger / trace attribution), and added a small example. Most TS helpers should never touch it.

**`withTimeGuard` (Egon L145):** added. Same shape as `withCostGuard`, same ALS-stack fix.

**`withTestContext` naming (Egon L166):** marked `@internal` with a JSDoc note saying it's a test-only escape hatch and shouldn't be used in production TS-helper code (those frames are seeded by `Runner.runInScope`).

**`addCost` (Egon L201):** added. Bumps `localCost`, calls `chargeGuards(amount)`, then `enforceGuards()` — the same three-step sequence `lib/runtime/prompt.ts` runs after every built-in `llm()` completion. So a TS helper that wraps its own paid call site can route through `agency.addCost(...)` and get free participation in `getCost()` / `withCostGuard` / `withTimeGuard`.

29/29 tests pass (added 5 new: per-branch isolation, TimeGuard install + throw-path, addCost accumulation + guard-trip). Full suite 4454/4454 (4425 baseline + 29).

PR #209, opened 2026-05-27
runtime: add agency.withResumableScope for resumable TS workflows

egonSchiele:
## Summary

Adds `agency.withResumableScope({name, ...}, async (s) => {...})` — a TS façade that wraps a body in the same substep-counter + serialized-frame machinery generated Agency function bodies use. Interrupts that happen inside a step body re-enter exactly where they left off on resume: already-completed steps are skipped, the deepest in-flight step re-runs from scratch.

Builds on the `agency.*` namespace from PR #208 and the ALS callsite seeding from PR #207.

## Example

```ts

export async function processOrder(id: string) {
  return agency.withResumableScope({ name: "processOrder" }, async (s) => {
    const order     = await s.step(1, () => fetchOrder(id));
    const validated = await s.step(2, () => validate(order));
    const stored    = await s.step(3, () => writeToDb(id, validated));
    return stored;
  });
}
```

## Scope API (v1)

| Method | Purpose |
|---|---|
| `step(id, fn)` | Resumable step; cached return value persists across resume in `frame.locals[__scope_step_${id}]`. |
| `hook(id, fn)` | Substep-counter-idempotent fire-and-forget hook for once-per-scope side effects. |
| `local(key, init)` / `setLocal` / `getLocal` | Frame-local storage that survives resume; lives under `frame.locals.__userLocals` to avoid collision with codegen-emitted keys. |
| `halt(result)` | Sets the underlying Runner's halt flag. Subsequent `s.step(...)` calls short-circuit; `withResumableScope` returns `result`. Does NOT throw. |

Options: `name` (required), `moduleId` (defaults to `"<ts-helper>"`), `pinResultCheckpoint` (defaults to `true`).

## Design decisions

- **Reuse Runner's halt machinery — don't duplicate it.** All halt state lives on the underlying `Runner` (`halted` / `haltResult`). The scope facade is stateless: no closure-local `halted`, no `HALT_SENTINEL`, no try/catch dance. After `runner.halt(...)`, every subsequent `runner.step(...)` short-circuits via `shouldSkip()`, so the body's remaining `await s.step(...)` calls are no-ops and `body(scope)` resolves naturally.
- **Step return values via `frame.locals`.** `runner.step(id, callback)` returns `void` and persists step state in `frame.locals`. To support `const x = await s.step(id, () => fetchValue())`, the wrapper stores `await fn()` under `frame.locals[__scope_step_${id}]` and returns the cached value. On resume, when the substep counter short-circuits the callback, the cached value is returned — matching what the original execution observed.
- **Type renamed `Scope` → `ResumableScope`.** `Scope` was already used in `lib/types.ts` (symbol-table scopes). Renamed at the public surface to `ResumableScope` / `ResumableScopeOpts` so the import doesn't shadow.
- **`stateStack.pop()` unconditional in `finally`.** Mirrors the generated function-body emission in `typescriptBuilder.ts`. If a regression ever requires a conditional pop, the branch logic should live on `Runner` (as a `shouldPopOnFinally()` method) — not in the wrapper.
- **`pinResultCheckpoint: true` by default.** Matches Agency function-body behaviour so `result.retry()` semantics work the same way for resumable-scope-wrapped helpers.

## Determinism contract

Step bodies must be pure with respect to inputs. On resume, the in-flight step re-runs from scratch — any non-pure work (`Date.now`, `Math.random`, cumulative I/O) will diverge from the pre-interrupt run. All "I/O that should run exactly once" must live inside a `s.step(...)` body. Documented in the file's header JSDoc.

## Tests

**Unit tests (17 passing):** `lib/runtime/resumableScope.test.ts`
- Basic flow, non-interrupt error propagation, `stateStack.pop` in finally on completion + halt + throw
- Halt contract: return value, no callback execution after halt, undefined cached result after halt
- Frame locals: init runs once, setLocal/getLocal contract
- Substep idempotence on re-execution (cached-value contract)
- Callsite defaults + override; pinResultCheckpoint true/false
- Hooks fire once
- Nested scopes: callsite independence, substep counter isolation

**Integration tests (3 passing):** `tests/agency-js/resumable-scope/`
- `basic` — Agency `def main()` invokes a TS helper that uses `withResumableScope` with three sequential steps; asserts final return value bubbles back through Agency.
- `halt-contract` — TS helper calls `s.halt(value)` inside step 2; asserts steps 3 + 4 do NOT execute and `withResumableScope` returns the halt value (not the body's trailing return).
- `nested-scopes` — outer `withResumableScope` invokes inner `withResumableScope` from inside an outer step; asserts both scopes run their steps in order and the inner result is composed into the outer return.

## Deferred to follow-ups

The plan listed 8 integration tests; 5 are deferred because they require a TS-facing interrupt helper that the runtime does not expose today:

- `halt-interrupt`, `local-persistence`, `fork-branch-with-scope`, `pin-result-checkpoint-retry`, `hook-idempotence` (across resume).

Each of these needs a public `agency.interrupt(...)` (or similar) helper that wraps the `interruptWithHandlers` + `getInterruptResponse` + checkpoint attachment dance that generated Agency code currently inlines. That helper lands with `agency.llm` (Plan 4); these integration tests follow there.

The unit tests already cover the load-bearing scope semantics in isolation (halt contract, substep idempotence, frame locals, callsite, nested scopes), and the three integration tests prove the end-to-end Agency → TS helper → `withResumableScope` path is wired correctly. The deferred tests verify the interrupt round-trip; pushing them to PR 4 keeps this PR focused.

## Validation

- `pnpm tsc --noEmit` clean
- `pnpm test:run` — 4471 / 4471 (4454 baseline + 17 new)
- All 3 integration tests under `tests/agency-js/resumable-scope/` pass
- No fixture changes (purely additive)


packages/agency-lang/lib/runtime/resumableScope.ts
```diff
@@ -0,0 +1,162 @@
...
+   *  first time the key is referenced. */
+  local<T>(key: string, init: () => T): T;
+  setLocal<T>(key: string, value: T): void;
+  getLocal<T>(key: string): T | undefined;
+
+  /** Halt the scope; `withResumableScope` will return `result`.
+   *  Used to bubble interrupt responses out of a step body. Sets a
+   *  flag on the underlying Runner — every subsequent `s.step(...)`
+   *  short-circuits without invoking its callback. Does NOT throw. */
+  halt(result: unknown): void;
```
egonSchiele, line 100: is this how ts code would throw an interrupt?

packages/agency-lang/lib/runtime/resumableScope.ts
```diff
@@ -0,0 +1,162 @@
...
+};
+
+const STEP_RESULT_PREFIX = "__scope_step_";
+const USER_LOCALS_KEY = "__userLocals";
+
+export async function withResumableScope<T>(
+  opts: ResumableScopeOpts,
+  body: (s: ResumableScope) => Promise<T>,
+): Promise<T> {
+  const moduleId = opts.moduleId ?? "<ts-helper>";
```
egonSchiele, line 110: Should this be required? What happens if multiple different TypeScript files have the same module ID?

packages/agency-lang/lib/runtime/resumableScope.ts
```diff
@@ -0,0 +1,162 @@
...
+  const pin = opts.pinResultCheckpoint ?? true;
```
egonSchiele, line 111: What is this for? the debugger?

packages/agency-lang/lib/runtime/resumableScope.ts
```diff
@@ -0,0 +1,162 @@
...
+  const runner = new Runner(ctx, stack, {
+    state: stack,
+    moduleId,
+    scopeName: opts.name,
+    stack: stateStack,
+    threads,
+  });
+
+  const scope: ResumableScope = {
+    step: async <U>(id: number, fn: () => U | Promise<U>): Promise<U> => {
```
egonSchiele: Does ID always have to be a number, or can it be a string? If it always has to be a number, and a monotonically increasing number, why not just leave it out and compute the step number for the user?

packages/agency-lang/lib/runtime/resumableScope.ts
```diff
@@ -0,0 +1,162 @@
...
+      await runner.step(id, async () => {
+        stack.locals[key] = await fn();
+      });
+      // After a no-op short-circuit (halted, replayed substep), the
+      // cached value persists from the previous execution. Returns
+      // `undefined as U` for steps that never ran (e.g. after halt).
+      return stack.locals[key] as U;
+    },
+
+    hook: (id, fn) => runner.hook(id, async () => { await fn(); }),
```
egonSchiele: not sure I understand what hook is for

packages/agency-lang/lib/runtime/resumableScope.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    local: <U>(key: string, init: () => U): U => {
```
egonSchiele: When would it be useful to use this function instead of getLocal and setLocal?

packages/agency-lang/tests/agency-js/resumable-scope/nested-scopes/test.js
```diff
@@ -0,0 +1,10 @@
+import { main } from "./agent.js";
+import { calls } from "./helpers.js";
```
egonSchiele, line 2: Did you forget to check some files in? Not sure what you're testing here. Can you please make sure you have a test that tests the interrupt functionality and make sure it works? Make sure you have a test where one TypeScript file with a runner calls into another TypeScript file with a runner, and the second one returns an interrupt. Test that the interrupt bubbles up correctly across two TypeScript files, and then on resume, correctly resumes execution inside the nested TypeScript file.

egonSchiele:
Thanks for the review! Answers and fixes inline. All landed in commits cdfde91a and 2ba92637; integration tests green locally.

### Copilot

- **Missing `helpers.js` in `basic/`, `halt-contract/`, `nested-scopes/`** — real bug. The files existed locally but were silently ignored by `**/*.js` in `.gitignore`. Force-added them, matching the established pattern in `tests/agency-js/tool-retry/tools.js`. CI will now resolve the imports.

- **`pinResultCheckpoint` using `label: null`** — also a real bug. `RuntimeContext.getResultCheckpoint()` filters on `label === RESULT_ENTRY_LABEL`, so the pin never participated in `result.retry()` despite what the JSDoc claimed. Now imports `RESULT_ENTRY_LABEL` and uses it; added a unit test that walks `ctx.checkpoints.getSorted()` looking for a pinned result-entry to lock the contract in.

### Egon

- **Line 74 — "is this how ts code would throw an interrupt?"** No — `halt()` only sets `runner.halted = true` so the remaining `s.step()` calls short-circuit and the scope resolves with the halt value. Actually throwing an interrupt requires the full codegen dance (interruptWithHandlers → checkpoint create → attach checkpointId → persist interruptId → halt with `{messages, data: [intr]}`); that's what `agency.interrupt()` will encapsulate in Plan 4. Rewrote the JSDoc to be explicit on both counts.

- **Line 84 — moduleId required / collisions** — Made it explicit in the JSDoc: it's purely a label for grouping in trace files and the debugger. Execution is unaffected by collisions because each scope still gets its own isolated frame on the stack at runtime. Not requiring it because that's a footgun for casual TS helpers; default `"<ts-helper>"` lets users opt in to labels.

- **Line 85 — `pinResultCheckpoint` what for / debugger?** — It's for `result.retry()` rewinding to scope entry, same as it would for a generated function body. Not for the debugger. JSDoc rewritten to say this directly, and combined with the `RESULT_ENTRY_LABEL` fix above the pin now actually works.

- **Line 112 — explicit step id** — Good call. Auto-incremented now: `s.step(fn)` with no id, counter lives on the scope facade and maps to a stable substep slot. The determinism contract (stable step-call order across runs+resumes) is documented at the top of the file and is what keeps the counter aligned on resume. Step ids must still be numbers under the hood because they slot into `__substep_${id}` on the frame, but users no longer see them.

- **Line 123 — `hook`** — Dropped from the public API. The only difference from `step` was skipping the debugger-pause hook (useful for codegen-emitted callback sites, not for user TS code). If we find a real use case later, easy to add back.

- **Line 125 — `local` vs `getLocal`/`setLocal`** — Dropped `local`; kept only `getLocal`/`setLocal` per your direction.

- **Nested-scopes / cross-TS-file interrupt test** — The fixture files are now in (they were just gitignored before, same bug Copilot caught). On the cross-TS-file interrupt test you asked for: I want to do this right and it depends on a public `agency.interrupt()` helper that wraps the codegen dance. I'm proposing to build that as Plan 4's first deliverable so the test exercises the real public API instead of a one-off inline helper. If you'd rather have the test in this PR, I'll build a minimal `agency.interrupt()` helper here and add the test; otherwise it slots in immediately after this lands.

### Summary of commits

- `cdfde91a` — force-add the three `helpers.js` fixtures
- `2ba92637` — `resumableScope.ts` API and JSDoc changes, unit tests updated, `RESULT_ENTRY_LABEL` wired through

PR #210, opened 2026-05-27
runtime: add agency.llm() for TS callers

egonSchiele:
Plan 4 of the 5-PR sequence to make TypeScript helpers first-class participants in agent runs. Adds `agency.llm` — a TS façade over `runPrompt` so user TS code can issue LLM calls with full participation in the surrounding agent run (cost tracking, the active LLM client, thread accumulation, trace events, checkpoint integration).

The codegen emission for Agency-source `llm(...)` is unchanged — it continues to call `runPrompt` directly. Both paths converge inside `runPrompt`.

### API

```ts

const text = await agency.llm("Summarize: " + doc);
const obj  = await agency.llm("Extract", { schema: ExtractSchema });
```

```ts
export type LlmOpts = {
  /** Override the model for THIS call only. Does NOT mutate the
   *  active client config. */
  model?: string;
  /** Zod schema → runPrompt's responseFormat. */
  schema?: z.ZodSchema;
  /** Override thread (default: active thread). */
  thread?: MessageThread;
};
```

### Design decisions

- **No `tools` field.** The `__toolRegistry` is per-Agency-module and codegen-only; exposing it to TS leaks codegen internals into the public surface. If a TS helper needs LLM-driven tool dispatch, define it as an Agency `def` (which gets the registry automatically) and invoke that `def` from TS. Pinned with a `@ts-expect-error` test so accidentally adding `tools?` later breaks the build.
- **No `removedTools`, no `maxToolCallRounds`.** Those are codegen-internal and only meaningful when tools are in play.
- **`opts.model` is per-call.** It is NOT a sticky setter on the active client config; subsequent `agency.llm` calls without `opts.model` fall back to the client default. Pinned with a regression test (`opts.model overrides the model for THIS call only`).
- **No `opts.systemMessage` in v1.** Trivially done with `agency.thread.system(msg)` before the call — keeps the surface minimal. Easy add if real demand emerges.

### Tests

10 unit tests in `lib/runtime/agencyLlm.test.ts`:

**Basic behavior:**
1. Returns the assistant string response.
2. Prompt + response land in the active thread (`[user, assistant]`).
3. Cost tracking: `ctx.stateStack.localCost` increments after the call.
4. Structured output via Zod schema parses correctly.

**Options mapping (load-bearing — pin the façade ↔ runPrompt contract):**
5. `opts.thread` override routes the prompt + response to the override thread; active thread unaffected.
6. `opts.model` is per-call: a second call with no override reverts to the client default.
7. `opts.schema` flows through as `responseFormat`.
8. `agency.callsite()` populates `checkpointInfo` (Plan 1 ↔ Plan 4 seam).

**Surface lock:**
9. `LlmOpts` rejects `tools`, `removedTools`, `maxToolCallRounds` at compile time (three `@ts-expect-error` assertions).

**Frame requirement:**
10. Throws when called outside any agency frame (`getRuntimeContext()`).

### Validation

- `pnpm tsc --noEmit` clean
- `pnpm vitest run lib/runtime/agencyLlm.test.ts` — 10/10 pass
- Full `pnpm test:run` shows 58 failures vs 68 on main; the 10-test delta comes from the new file. No pre-existing test regressed.

### Deferred to a follow-up commit / PR

Task 3 of the plan (`tests/agency-js/agency-llm/*`) — six end-to-end integration tests covering cost attribution through TS, model-override semantics in a live run, checkpoint-after-llm, structured output, thread override, and the cost-guard-interrupt-through-TS path. These exercise the live `pnpm run agency test js ...` harness and are best added as a follow-up so this PR stays focused on the façade itself. Happy to do them in this PR if you'd rather.

### DIFF
--- a/packages/agency-lang/lib/runtime/agency.ts
+++ b/packages/agency-lang/lib/runtime/agency.ts
@@ -34,6 +34,7 @@ import {
   withPushedHandler,
   type CallsiteLocation,
 } from "./asyncContext.js";
+import { llm as _llm } from "./agencyLlm.js";
 import {
   checkpoint as _checkpoint,
   getCheckpoint as _getCheckpoint,
@@ -267,6 +268,8 @@ export const agency = {
   addCost,
 
   withResumableScope: _withResumableScope,
+
+  llm: _llm,
 
   withTestContext,
 };
--- a/packages/agency-lang/lib/runtime/agencyLlm.checkpointInfo.test.ts
+++ b/packages/agency-lang/lib/runtime/agencyLlm.checkpointInfo.test.ts
@@ -1,0 +1,82 @@
+import { describe, it, expect, vi } from "vitest";
+
+/**
+ * Pin the contract that `agency.llm` forwards `agency.callsite()` to
+ * `runPrompt` as `checkpointInfo`. A regression that drops that
+ * forwarding would silently disable interrupt-checkpoint location
+ * attribution for TS-issued LLM calls — `getResultCheckpoint()` would
+ * still find the result-entry pin, but the runtime checkpoint
+ * `PromptRunner` writes when the LLM call interrupts would lose its
+ * `moduleId / scopeName / stepPath`. That makes the rewind UI show
+ * "unknown location" and breaks `step in / over` after a resumed
+ * interrupt. We can't observe `checkpointInfo` via the LLM client
+ * (it's consumed by `PromptRunner`, not by the client), so we mock
+ * `runPrompt` itself and read the field off the captured args.
+ *
+ * Lives in its own file because `vi.mock` is module-scoped and we
+ * don't want it to replace `runPrompt` for every other agencyLlm
+ * test in this directory.
+ */
+vi.mock("./prompt.js", () => ({
+  runPrompt: vi.fn(async () => "spy-response"),
+}));
+
+import { agency } from "./agency.js";
+import { RuntimeContext } from "./state/context.js";
+import { ThreadStore } from "./state/threadStore.js";
+import { runPrompt } from "./prompt.js";
+
+function makeCtx(): RuntimeContext<any> {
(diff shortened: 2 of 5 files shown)


egonSchiele:
Thanks Copilot — both addressed.

**#1 — `Promise<any>` return type.** Valid. Replaced with a generic + overload pair in commit `07fce325`:

```ts
export function llm<S extends z.ZodSchema>(
  prompt: string,
  opts: LlmOpts<S> & { schema: S },
): Promise<z.infer<S>>;
export function llm(prompt: string, opts?: LlmOpts): Promise<string>;
```

Schema overload comes first so TS picks it over the no-schema fallback when both could match. Added two type-level tests:
- `agency.llm("p")` — asserts `const s: string = r` so a regression that loses the no-schema branch becomes a TS error.
- `agency.llm("p", { schema })` — asserts typed field access and pins a `@ts-expect-error` on a wrong field name.

**#2 — checkpointInfo coverage.** Already addressed in commit `baf6be25` (you reviewed against `30e0af29`). The new file [`agencyLlm.checkpointInfo.test.ts`](https://github.com/egonSchiele/agency-lang/blob/agency-llm/packages/agency-lang/lib/runtime/agencyLlm.checkpointInfo.test.ts) mocks `runPrompt` via `vi.mock` and asserts the captured args:

```ts
expect(spy.mock.calls[0][0]).toMatchObject({
  checkpointInfo: { moduleId: "M", scopeName: "S", stepPath: "1.2" },
});
```

Plus a companion test that no callsite installed → `checkpointInfo` is forwarded as `undefined`. Lives in its own file because `vi.mock` is module-scoped and we don't want it to replace `runPrompt` for the other agencyLlm tests. The weak callsite-only test was dropped from the main file and replaced with a pointer comment.

13/13 unit tests passing at HEAD `07fce325`.

PR #212, opened 2026-05-28
runtime: add agency.interrupt() for TS callers

egonSchiele:
## Why

After the 5-PR series that made TS helpers first-class participants in agent runs (#206–#211), TS code can install handlers, call `agency.llm`, run inside `agency.withResumableScope`, and observe interrupts — but it cannot **raise** an interrupt. The codegen `interrupt(...)` emission is ~35 lines per call site (see `lib/templates/backends/typescriptGenerator/interruptAssignment.mustache` and `interruptReturn.mustache`) and that dance has no public encapsulation.

That gap blocks both follow-up plans for testing infrastructure:
- `docs/superpowers/plans/2026-05-27-runtime-integration-harness.md`
- `docs/superpowers/plans/2026-05-27-missing-integration-tests.md`

Both require TS-triggered interrupts to simulate round-trips in a `runScenario(...)` vitest harness without needing a `.agency` file per test.

## What

A single TS helper, `agency.interrupt({ kind, message, data })`, that runs the same handler-chain + resume-idempotent halt-with-checkpoint protocol as the codegen path. Surface is intentionally minimal (just the three options the user requested — no convenience knobs).

```ts
const resp = await agency.interrupt({
  kind: "needs-approval",
  message: "delete the production database?",
  data: { table: "users" },
});
if (resp.type === "approve") { /* go */ } else { /* abort */ }
```

## Mechanism

Mirrors the codegen template ordering exactly:

1. **Lookup persisted id**: read `frame.locals[\`__interrupt_${callsite.stepPath}\`]`. If a response is already recorded for that id on the context, return it immediately. This is what makes the helper resume-idempotent — the second pass through the same step body sees the stamp and short-circuits before re-firing handlers or creating a duplicate checkpoint.
2. **Handler chain**: call `interruptWithHandlers(...)` (the same routine the codegen path uses). Approve / reject outcomes flow straight back to the caller.
3. **No handler → propagate**: persist the new interrupt id on `frame.locals` **before** creating the checkpoint (so the snapshot captures it — same ordering as the template). Attach the resulting `checkpointId` + checkpoint snapshot onto the propagated interrupt. Halt the runner with `{messages, data: [intr]}` in a node body, raw `[intr]` in a function/scope body.
4. Throw `HaltSignal`, which `Runner.step` absorbs once `runner.halted` is set. This is the only way to keep post-interrupt code from running on the first pass — TS callers can't be forced to `return` the way generated code does.

## Plumbing

- `AgencyStore` gains an optional `runner` slot. `Runner.runInScope` now threads `runner: this` into the ALS frame so TS helpers can call `runner.halt(...)` without taking a runner argument.
- `Runner` exposes a public `isNodeContext` getter so `agency.interrupt` can produce the right halt payload shape (`{messages, data}` vs raw `data`).
- `HaltSignal` lives in its own module (`lib/runtime/haltSignal.ts`) to break the would-be import cycle between `runner.ts` (which catches it) and `agencyInterrupt.ts` (which throws it).

## Interrupt-id key allocation strategy

The codegen path keys persisted ids off a compile-time-stable `N` per call site (`__interruptId_${N}`). At runtime we don't have `N`, so I went with the **stepPath-as-key** approach the user's note flagged as "clean but constraining":

> Key = `__interrupt_${callsite.stepPath}` → one `agency.interrupt(...)` per step body.

Inside `agency.withResumableScope`, each `s.step(...)` already gets a unique substep path, so the natural pattern is "wrap each interrupt in its own `s.step(...)`". When called from a TS function invoked by generated code, the surrounding step's `stepPath` is the key. Code that needs multiple sequential interrupts splits them across multiple `s.step(...)` calls — a small constraint that aligns with how `withResumableScope` already encourages users to chunk work.

The alternative (counter on the ALS frame) works everywhere but adds resume-determinism foot-guns. Going with the simpler design for v1 and documenting the constraint in the helper's docstring; we can revisit if real usage shows it's too restrictive.

## Tests

`lib/runtime/agencyInterrupt.test.ts` (7 cases, all the user's explicit requests):

1. **Handler approves** → `agency.interrupt` returns `approve(value)` without halting.
2. **Handler rejects** → returns `reject(value)` without halting.
3. **No handler — halt shape** → scope returns `[intr]` with `checkpointId`, snapshot attached, and the checkpoint resolvable from `ctx.checkpoints`.
4. **No handler — no post-interrupt code** → confirms the throw-and-absorb mechanism keeps the rest of the step body from running on the first pass.
5. **No handler — no later steps** → confirms `runner.halted` short-circuits subsequent `s.step(...)` bodies (existing `withResumableScope` semantics, regression-guarded).
6. **Resume idempotency** → stamps a user response for the persisted id, re-enters the scope with the same id pre-seeded on the new frame, confirms `agency.interrupt` returns the user's response without consulting handlers.
7. **Frame requirement** → throws a clear error when called without a Runner in the ALS frame.

`makeMockCtx()` (`lib/runtime/__tests__/testHelpers.ts`) picks up the small surface `interruptWithHandlers` + `respondToInterrupts` exercise: `isCancelled`, `enterToolCall` / `exitToolCall`, `setInterruptResponses` / `getInterruptResponse`, and the missing `statelogClient` event methods.

All 698 tests in `lib/runtime/` pass; no regressions in adjacent files (`resumableScope`, `agency`, `runner`, `interrupts`, `checkpoint`, `asyncContext`, `agencyLlm`, `hooks`).

## Unblocks

- `docs/superpowers/plans/2026-05-27-runtime-integration-harness.md` (the `runScenario` harness needs TS-triggered interrupts to exercise the round-trip).
- `docs/superpowers/plans/2026-05-27-missing-integration-tests.md` (7 specific integration tests that close the gaps from PR #211's deferred-tests review).

## Validation

- `pnpm tsc --noEmit` — clean
- `pnpm vitest run lib/runtime/` — 698 passed

### DIFF
--- a/packages/agency-lang/lib/runtime/agency.ts
+++ b/packages/agency-lang/lib/runtime/agency.ts
@@ -35,7 +35,7 @@ import {
   type CallsiteLocation,
 } from "./asyncContext.js";
 import {
-  interrupt as _interrupt,
+  interrupt,
   type InterruptOpts,
 } from "./agencyInterrupt.js";
 import { llm as _llm } from "./agencyLlm.js";
@@ -275,7 +275,7 @@ export const agency = {
 
   llm: _llm,
 
-  interrupt: _interrupt,
+  interrupt,
 
   withTestContext,
 };
--- a/packages/agency-lang/lib/runtime/agency.ts
+++ b/packages/agency-lang/lib/runtime/agency.ts
@@ -34,6 +34,10 @@ import {
   withPushedHandler,
   type CallsiteLocation,
 } from "./asyncContext.js";
+import {
+  interrupt as _interrupt,
+  type InterruptOpts,
+} from "./agencyInterrupt.js";
 import { llm as _llm } from "./agencyLlm.js";
 import {
   checkpoint as _checkpoint,
@@ -271,7 +275,9 @@ export const agency = {
 
   llm: _llm,
 
+  interrupt: _interrupt,
+
   withTestContext,
 };
 
-export type { ResumableScope, ResumableScopeOpts };
+export type { InterruptOpts, ResumableScope, ResumableScopeOpts };
--- a/packages/agency-lang/lib/runtime/agencyInterrupt.ts
+++ b/packages/agency-lang/lib/runtime/agencyInterrupt.ts
@@ -1,0 +1,179 @@
+/**
+ * `agency.interrupt` — user-facing TS encapsulation of the codegen
+ * `interrupt(...)` dance.
+ *
+ * Generated Agency code today emits ~35 lines per `interrupt(...)`
+ * call site (see lib/templates/backends/typescriptGenerator/
+ * interruptAssignment.mustache and interruptReturn.mustache). This
+ * helper mirrors that emission so TS code can raise interrupts the
+ * same way an Agency `interrupt` statement would, fully participating
+ * in handlers, checkpointing, and the resume protocol.
+ *
+ * # Mechanism
+ *
+ * Per call:
+ *  1. Read the persisted interrupt id off the active state frame's
+ *     `locals` (keyed by the surrounding step's path). If a response
+ *     for that id is already recorded on the context, we are on the
+ *     resume side — return it immediately. This is the resume-
+ *     idempotency mechanism: the second execution of the same step
+ *     body sees the stamp and short-circuits without re-firing
+ *     handlers or creating a duplicate checkpoint.
+ *  2. Otherwise, consult `interruptWithHandlers` (the same routine the
+ *     codegen path uses). Approved / rejected outcomes flow straight
+ *     back to the caller.
+ *  3. If no handler intercepts, this is a brand-new propagation:
+ *     - Persist the new interrupt id on `frame.locals` BEFORE
+ *       creating the checkpoint so the id is captured in the
+ *       snapshot. (Same ordering as the codegen template — the post-
+ *       restore replay relies on reading the id from the restored
+ *       frame.)
+ *     - Create a checkpoint via `ctx.checkpoints.create` with the
+ *       location attached to the active ALS callsite, attach the
+ *       resulting id + checkpoint object onto the propagated
+ *       interrupt, and `runner.halt(...)` with the interrupt array.
+ *     - Throw `HaltSignal`, which the surrounding `Runner.step`
+ *       absorbs once the runner is halted. This stops the caller
+ *       from executing post-interrupt code on the first pass; on
+ *       resume the same call site reaches step 1 and returns the
+ *       user's response without re-tripping.
+ *
+ * # Interrupt-id key allocation
+ *
+ * The codegen path keys persisted ids off a compile-time-stable N per
+ * call site (`__interruptId_${N}`). At runtime we don't have N, so
+ * we key off the active callsite's `stepPath` instead:
+ * `__interrupt_${stepPath}`. This makes the constraint explicit:
+ *
+ *   **One `agency.interrupt(...)` per step body.**
+ *
+ * Inside `agency.withResumableScope`, each `s.step(...)` body gets its
+ * own substep path, so the natural pattern is "wrap each interrupt in
+ * its own `s.step(...)`". When called from a TS function invoked from
+ * generated code, the surrounding generated step's `stepPath` is the
+ * key — so likewise, raise at most one interrupt per generated step
+ * call site. Code that needs multiple sequential interrupts should
+ * split them across multiple `s.step(...)` calls.
+ *
+ * # Halting and HaltSignal
+ *
+ * After `runner.halt(...)`, this function throws `HaltSignal` rather
+ * than returning a synthetic "halted" value. The codegen template
+ * stops execution with an explicit `return` after `runner.halt`; from
+ * inside a TS helper we cannot force the caller to return, so the
+ * throw is the only way to keep post-interrupt code from running on
+ * the first pass. `Runner.step` catches `HaltSignal` when
+ * `this.halted` is set; the user never sees it.
+ *
+ * # Required ALS frame
+ *
+ * Throws if called outside a frame seeded with a `Runner`. In practice
+ * this means: inside `s.step(...)` of `agency.withResumableScope`, or
+ * inside a TS function called from a generated Agency step body. Calls
+ * from the top-level `withResumableScope` body (outside any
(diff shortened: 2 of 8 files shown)


packages/agency-lang/lib/runtime/agency.ts
```diff
@@ -34,6 +34,10 @@ import {
   withPushedHandler,
   type CallsiteLocation,
 } from "./asyncContext.js";
+import {
+  interrupt as _interrupt,
```
egonSchiele: why the rename?

packages/agency-lang/lib/runtime/agencyInterrupt.ts
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * Per call:
+ *  1. Read the persisted interrupt id off the active state frame's
+ *     `locals` (keyed by the surrounding step's path). If a response
+ *     for that id is already recorded on the context, we are on the
+ *     resume side — return it immediately. This is the resume-
+ *     idempotency mechanism: the second execution of the same step
+ *     body sees the stamp and short-circuits without re-firing
+ *     handlers or creating a duplicate checkpoint.
+ *  2. Otherwise, consult `interruptWithHandlers` (the same routine the
```
egonSchiele, line 22: so a TS function, called in agency code, could have its interrupt caught by handlers defined in the agency code?

packages/agency-lang/lib/runtime/agencyInterrupt.ts
```diff
@@ -0,0 +1,179 @@
...
+ *       user's response without re-tripping.
+ *
+ * # Interrupt-id key allocation
+ *
+ * The codegen path keys persisted ids off a compile-time-stable N per
+ * call site (`__interruptId_${N}`). At runtime we don't have N, so
+ * we key off the active callsite's `stepPath` instead:
+ * `__interrupt_${stepPath}`. This makes the constraint explicit:
+ *
+ *   **One `agency.interrupt(...)` per step body.**
```
egonSchiele, line 55: please document this.

packages/agency-lang/lib/runtime/agencyInterrupt.ts
```diff
@@ -0,0 +1,179 @@
...
+  isRejected,
+  type Interrupt,
+  type InterruptResponse,
+} from "./interrupts.js";
+
+export type InterruptOpts<T = unknown> = {
+  /** Stable identifier for the interrupt kind. Mirrors the kind a
+   *  generated `interrupt foo` statement would emit. Used by handlers
+   *  to decide whether to intercept. */
+  kind: string;
```
egonSchiele: optional, set to `unknown` if missing

packages/agency-lang/lib/runtime/agencyInterrupt.ts
```diff
@@ -0,0 +1,179 @@
...
+  /** Human-readable description shown to the user / dashboards when
+   *  the interrupt propagates. */
+  message: string;
+  /** Arbitrary structured data attached to the interrupt. Forwarded
+   *  to handlers and persisted on the propagated `Interrupt`. */
+  data: T;
```
egonSchiele: optional

packages/agency-lang/lib/runtime/agencyInterrupt.ts
```diff
@@ -0,0 +1,179 @@
...
+    );
+  }
+
+  const key = `__interrupt_${callsite.stepPath}`;
+
+  // Resume path: if we already persisted an interrupt id at this
+  // location AND the user has responded, return the response without
+  // re-running the handler chain or creating another checkpoint. The
+  // codegen template does exactly this with `__self.__interruptId_N`.
+  const persistedId = frame.locals[key] as string | undefined;
```
egonSchiele: Is the `as` cast needed?

packages/agency-lang/lib/runtime/agency.ts
egonSchiele: Good catch — there was no local `interrupt` to shadow it. Dropped the alias in 2c49224. The renamed-import pattern in this file (`_checkpoint`, `_getCheckpoint`, `_restore`) is only needed where a same-named local is also declared.

packages/agency-lang/lib/runtime/agencyInterrupt.ts
egonSchiele: Yes, exactly — `ctx.handlers` is a single stack shared with generated Agency code, so handlers defined in an Agency `handle` block catch interrupts raised by a TS function called from that Agency code, and conversely an Agency `interrupt` statement nested inside a TS-installed `agency.withHandler(...)` is caught by that TS handler. Spelled this out in the module docstring in 2c49224.

packages/agency-lang/lib/runtime/agencyInterrupt.ts
egonSchiele: Added a paragraph in the module docstring documenting what happens if you violate this constraint: two `agency.interrupt(...)` calls with the same `stepPath` write to the same `frame.locals` key, the second overwrites the first's persisted id, so the resume path only ever looks up the second's response — the first's handlers fire from scratch on resume instead of short-circuiting. Symptom is silent rather than a thrown error. Done in 2c49224.

packages/agency-lang/lib/runtime/agencyInterrupt.ts
egonSchiele: Done in 2c49224 — `kind?: string` with `"unknown"` default in the implementation. Added a test (`agency.interrupt — option defaults > defaults kind to "unknown" when omitted`) verifying a handler sees `kind === "unknown"` when the caller omits it.

packages/agency-lang/lib/runtime/agencyInterrupt.ts
egonSchiele: Done in 2c49224 — `data?: T` (no default; propagates as `undefined`). Added a test (`agency.interrupt — option defaults > propagates with undefined data when data is omitted`) confirming the interrupt payload carries `data === undefined`.

packages/agency-lang/lib/runtime/agencyInterrupt.ts
egonSchiele: No — `frame.locals` is `Record<string, any>` so the lookup already comes back as `any`, and the subsequent `persistedId !== undefined` narrows correctly without the annotation. Dropped the cast in 2c49224.

PR #213, opened 2026-05-28
tests: integration coverage for agency.* TS helpers

egonSchiele:
Adds 7 integration tests under `tests/agency/ts-helpers/` pinning the load-bearing behaviors of the `agency.*` TS-helpers surface (`agency.withResumableScope`, `agency.interrupt`, `agency.withHandler`, `agency.withCostGuard`, `agency.addCost`). The per-call unit tests added in PRs #207–#212 cover individual contracts; this PR closes the integration-coverage gap surfaced by the post-PR-211 self-review.

## Why agency tests (not agency-js, not a new vitest harness)

The question we are testing is *"does TS code calling `agency.*` participate in agent runs correctly?"* — i.e. **agency code calling JS**. That's the natural domain of `tests/agency/`: the `.agency` file is the entry, it imports a `.js` helper, the helper uses `agency.*`. The runner already drives interrupt round-trips declaratively via `interruptHandlers` in `.test.json`. No new test infrastructure was needed.

## Tests

Each test is a self-contained triplet (`agent.agency` + `helper.js` + `agent.test.json`) with assertions in `.test.json`:

| Test | Load-bearing behavior pinned |
| --- | --- |
| `resumable-scope-resume` | `withResumableScope` skips completed `s.step(...)` bodies on resume and re-runs the in-flight step from scratch, where it finds the user's response on the second pass. |
| `interrupt-handler-approve` | `agency.withHandler` returning `approve(value)` makes `agency.interrupt` resolve to that outcome with no halt and no resume. |
| `interrupt-handler-reject` | Same shape as approve but with `reject(value)`. |
| `interrupt-resume-idempotency` | When a handler propagates, `agency.interrupt` halts with a checkpoint; on resume, the persisted interrupt id short-circuits the second pass so the handler chain does NOT re-fire (counter stays at 1). |
| `halt-trycatch-cleanup` | `s.halt(value)` does not throw — surrounding `try/finally` still runs while subsequent `s.step(...)` calls short-circuit. |
| `cost-guard-trips` | `agency.addCost` over budget throws `GuardExceededError` synchronously, catchable inside the helper for structured failure reporting. |
| `fork-branch-isolation` | Per-branch `agency.addCost` contributions are attributed to each branch's own stack and do not leak across siblings. |

## Validation

- `pnpm tsc --noEmit` clean
- `pnpm run agency test tests/agency/ts-helpers/` — 7/7 pass
- No production code changes; this PR is tests-only
- No JS-side `expect` / `assert` calls — assertions live in `.test.json` (declarative)

## Follow-ups

This plan supersedes both `docs/superpowers/plans/2026-05-27-runtime-integration-harness.md` and `docs/superpowers/plans/2026-05-27-missing-integration-tests.md`. Both can be deleted in a follow-up cleanup.

### DIFF
--- a/packages/agency-lang/tests/agency/ts-helpers/interrupt-handler-reject/agent.agency
+++ b/packages/agency-lang/tests/agency/ts-helpers/interrupt-handler-reject/agent.agency
@@ -1,0 +1,6 @@
+import { run } from "./helper.js"
+
+node main() {
+  const result = run()
+  return result
+}
--- a/packages/agency-lang/docs/superpowers/plans/2026-05-27-ts-helper-integration-tests.md
+++ b/packages/agency-lang/docs/superpowers/plans/2026-05-27-ts-helper-integration-tests.md
@@ -1,0 +1,250 @@
+# TS-helper integration tests (as agency tests)
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans. Steps use checkbox (`- [ ]`) syntax.
+
+**Goal:** close the integration-coverage gap surfaced by the post-PR-211 self-review of the TS-helpers surface (`agency.*` namespace, `withResumableScope`, `agency.interrupt`, `agency.withCostGuard`, `agency.withHandler`). Unit tests cover per-call contracts; **load-bearing behaviors** (resumability across interrupts, handler interception, cost-guard trips, fork-branch isolation) are not exercised end-to-end.
+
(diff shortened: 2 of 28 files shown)


packages/agency-lang/tests/agency/ts-helpers/interrupt-handler-reject/agent.agency
```diff
@@ -0,0 +1,6 @@
+import { run } from "./helper.js"
+
+node main() {
+  const result = run()
```
egonSchiele, line 4: also add tests where the handlers are defined in agency

PR #215, opened 2026-05-28
memory: charge internal LLM and embed spend against the active cost guard

egonSchiele:
## What

Memory's internal LLM and embedding calls (extraction, compaction, tier-3 recall filter, observation embeddings, query embeddings) now charge their spend against the surrounding branch's `withCostGuard` budget via `agency.addCost(...)`. Previously these calls bypassed cost tracking entirely.

This is Proposal A from the design discussion around the [memory-llm-via-async-context spec](docs/superpowers/specs/2026-05-25-memory-llm-via-async-context-design.md) — the smallest possible step that buys correct cost semantics without restructuring the memory layer's LLM client plumbing. Future PRs (B/C) can switch memory's text calls to `agency.llm` and drop the constructor's `llmClient`/`smoltalkDefaults`/`statelogClient` params; this PR explicitly does not do that.

## Why

The memory layer maintains its own reference to the runtime's `LLMClient` (captured at `MemoryManager` construction time). Its calls reach the provider successfully, but the `cost.totalCost` they return never flowed into `stack.localCost`, so a user wrapping their agent in `withCostGuard($X)` could under-count actual spend whenever memory was active. The cost-guard spec calls this out as a hole.

The `agency.addCost(amount)` helper (#212 area) is the exact primitive `prompt.ts` uses post-completion for agency-side `llm()` calls — it does `stack.localCost += amount; chargeGuards; enforceGuards`. Calling it from memory's `_text`/`_embed` after a successful provider response gives memory calls the same cost-tracking semantics as native `llm()`.

## How

- `lib/runtime/memory/manager.ts`
  - Import `agency` and `agencyStore`.
  - Add a one-line `chargeCostIfInFrame(amount)` helper that calls `agency.addCost(amount)` only when an Agency frame is installed (so direct-construction unit tests don't blow up).
  - Call it inside `_text` (using `result.value.cost?.totalCost ?? 0`) and `_embed` (using `result.value.costEstimate?.totalCost ?? 0`, since `EmbedResult` uses the `costEstimate` field name).
- `lib/runtime/memory/manager.test.ts`
  - New `cost attribution` describe block with 4 tests:
    1. `_text` spend goes into `stack.localCost`.
    2. `_text` + `_embed` both charge when extraction yields observations.
    3. Direct-construction (no frame) no-ops cleanly.
    4. `withCostGuard($0.50)` trips when a $1.00 memory call runs.
- `CHANGELOG.md`
  - Added a `Changed` section under unreleased calling out the user-visible behavior change.

## Why this is safe to ship as a single PR

- Zero changes to memory's constructor surface — `MemoryManagerOptions` is untouched.
- Zero changes to which LLMClient is used or how — same code path, just one extra call after success.
- Frame check on the addCost path means direct-construction tests continue to work unchanged (all 14 existing tests pass).
- Reverting is one diff hunk per file if cost semantics surprise anyone.

## Validation

```
$ npx tsc --noEmit             # clean
$ npx vitest run lib/runtime/memory/manager.test.ts  # 18/18 pass (14 existing + 4 new)
$ npx vitest run lib/runtime/agency.test.ts          # 29/29 pass
$ pnpm run lint:structure       # clean
```

## Out of scope

- Switching `_text` to `agency.llm` (Proposal B): saves boilerplate and gives memory the test-client swap, but requires updating the existing statelog event assertions; deferred.
- Switching `_embed` to a new `agency.embed` helper (Proposal C): mirrors `agency.llm` but doesn't exist yet; deferred until B lands.
- Per-source cost breakdown in `tokenStats` so users can see "memory cost $X separately from agent cost $Y": noted in the spec as v1.x follow-up.
- Recursive-depth guard on `agency.addCost`/`agency.llm`: orthogonal to this PR, called out in the spec.

### DIFF
--- a/packages/agency-lang/lib/runtime/memory/manager.ts
+++ b/packages/agency-lang/lib/runtime/memory/manager.ts
@@ -31,6 +31,24 @@ import type { StatelogClient } from "../../statelogClient.js";
 import forgetTemplate from "../../templates/prompts/memory/forget.js";
 import retrievalTemplate from "../../templates/prompts/memory/retrieval.js";
 import type { LLMClient } from "../llmClient.js";
+import { agency } from "../agency.js";
+import { agencyStore } from "../asyncContext.js";
+
+/**
+ * Charge `amount` against the active branch's `withCostGuard` budget
+ * if (and only if) we're inside an Agency execution frame. Memory's
+ * text + embed calls run from inside `runPrompt`'s post-completion
+ * hook or from stdlib agency calls — both reach this code with an
+ * `agencyStore` frame already installed, so production paths always
+ * charge correctly. The frame check is for direct-construction unit
+ * tests that exercise `MemoryManager` without going through the
+ * runner.
+ */
+function chargeCostIfInFrame(amount: number): void {
+  if (amount <= 0) return;
+  if (!agencyStore.getStore()) return;
+  agency.addCost(amount);
+}
 
 // Cap on the `inputPreview` slice we put on every embedCompletion
 // event. Short enough that a transcript fragment in stderr or in a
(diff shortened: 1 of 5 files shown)


packages/agency-lang/lib/runtime/memory/manager.ts
```diff
@@ -31,6 +31,24 @@ import type { StatelogClient } from "../../statelogClient.js";
...
+ * text + embed calls run from inside `runPrompt`'s post-completion
+ * hook or from stdlib agency calls — both reach this code with an
+ * `agencyStore` frame already installed, so production paths always
+ * charge correctly. The frame check is for direct-construction unit
+ * tests that exercise `MemoryManager` without going through the
+ * runner.
+ */
+function chargeCostIfInFrame(amount: number): void {
+  if (amount <= 0) return;
+  if (!agencyStore.getStore()) return;
```
egonSchiele, line 49: should this log an error? when would this happen?

PR #220, opened 2026-05-29
feat(agent): Pi-inspired improvements to Agency Agent

egonSchiele:
Nine Pi-inspired improvements to make the Agency Agent in
`lib/agents/agency-agent/agent.agency` more capable, better-grounded,
and safer to run. Most changes are in the stdlib so any Agency author
can opt in; agent-specific glue lives in the agent file itself.

## Stdlib improvements

1. **`skillsDir` emits `<available_skills>` XML.** Matches Pi's skills
   format so recent coding models recognize it consistently. Each entry
   becomes a `<skill>` block with `<name>`, `<description>`,
   `<location>`.
2. **`read` gains `offset`/`limit` pagination + a 2000-line soft cap.**
   Previously every `read` dumped the whole file into context. The cap
   produces a clear truncation note pointing the model at
   `offset`/`limit`. `0` is the unset sentinel because Agency lacks
   undefined arguments. New `_read` unit tests.
3. **`ls`/`grep`/`glob` docstrings now name their truncation limits.**
   The tools already cap results internally (200 for grep, 500 for
   glob, 5 MB grep file-skip); now the LLM knows when to narrow scope.
4. **`edit` + `multiedit` consolidated into one `edit(filename, edits[])`**
   tool. Single surface, batches nearby changes naturally. `_edit`
   removed from `lib/stdlib/fs.ts`. Breaking change — every in-tree
   caller (3 agency-js fixtures, the stdlib-sandbox integration agent,
   the coding-agent example) is migrated in the same commit.

## Agent improvements

5. **`Current date:` + `Current working directory:` appended to the
   system prompt** so the LLM doesn't have to ask for grounding.
6. **`AGENTS.md` auto-loaded into a `<project_context>` block** if the
   workspace has one. Aligns with Anthropic / Pi / Aider conventions.
7. **`workspace.bash` added to the bundle** and the tools list, plus a
   workflow step in the system prompt. `cwd` is partial-applied to the
   bundle's anchor dir.
8. **`openDir(dir, allowedPaths?)` binds `allowedPaths` defense-in-depth**
   on `ls`/`glob`/`grep`/`bash`. Defaults to `[dir]` so the bundle
   refuses to escape its anchor even if downstream args are tampered
   with. `read`/`write`/`edit` remain sandboxed via `resolvePath`
   (filename containment within `dir`).

## Runtime addition

9. **Per-tool `promptGuidelines` + `toolGuidelines()` helper.** Pi's
   per-tool advice ("edit: prefer batching", "write: only for new
   files") rides with each tool's metadata via
   `AgencyFunction.withGuidelines(...)` and is composed into the system
   prompt by `std::agency.toolGuidelines(tools)`. The agent attaches
   guidelines at the tools-list call site for `read`, `write`, `edit`,
   `bash`, and `typecheck` and prepends the resulting block to its
   system message. `FunctionRefReviver` round-trips
   `promptGuidelines` alongside `.describe()` metadata; empty arrays
   are stripped on the wire. Unit tests cover `withGuidelines`,
   reviver round-trip, and the `_toolGuidelines` formatter.

## Verification

- `make` clean on a fresh worktree.
- `pnpm run agency typecheck lib/agents/agency-agent/agent.agency` →
  No type errors found.
- All targeted unit tests pass locally:
  `pnpm test:run lib/stdlib/builtins.test.ts lib/runtime/agencyFunction.test.ts lib/runtime/revivers/functionRefReviver.test.ts lib/stdlib/agency.test.ts`
  → 130 passed / 130.
- The three migrated agency-js fs fixtures (`std-fs-edit`,
  `std-fs-edit-reject`, `std-fs-multiedit`) all pass via
  `pnpm run agency test js …` after the edit-consolidation change.
- Full Agency suite deferred to CI per repo convention.


packages/agency-lang/docs/site/stdlib/agency.md
```diff
@@ -469,3 +469,23 @@ Type-check an Agency file on disk. The file is read from dir/filename, and relat
...
 
 ([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/agency.agency#L284))
+
+### toolGuidelines
+
+```ts
+toolGuidelines(tools: any[]): string
+```
+
+Collect every tool's `promptGuidelines` (set via `.withGuidelines(...)`) into a single prompt-ready block. Use this when assembling a system prompt to surface per-tool advice (e.g. "edit: prefer one call with multiple entries") to the LLM. Returns an empty string when no tool has guidelines, so it composes cleanly with `sysPrompt + toolGuidelines(tools)`.
```
egonSchiele: How do tool guidelines differ from tool descriptions?

packages/agency-lang/docs/site/stdlib/fs.md
```diff
@@ -74,57 +63,29 @@ export type Workspace = {
...
 
 ### edit
 
 ```ts
-edit(filename: string, oldText: string, newText: string, replaceAll: boolean, dir: string): Result
+edit(filename: string, edits: Edit[], dir: string): Result
 ```
 
-Edit a file by replacing oldText with newText. By default oldText must match exactly once in the file; pass replaceAll=true to replace every occurrence. Fails if oldText is not found or appears multiple times (unless replaceAll is set).
+Edit a single file by applying one or more text replacements atomically. Each edit has `oldText`, `newText`, and `replaceAll`. Every edit's `oldText` must match a unique, non-overlapping region of the original file (matches are looked up against the file as it stands when the edit runs, after earlier edits). Set `replaceAll: true` on an edit to replace every occurrence. When any edit fails, nothing is written.
```
egonSchiele, line 84: I'd like to print a nice diff to the console for each edit. Could we have a `printDiff` param in this function? Could you add a new function to print a diff in the stdlib?

packages/agency-lang/docs/site/stdlib/index.md
```diff
@@ -113,20 +113,30 @@ A tool for rounding a number to a specified number of decimal places.
...
+  When called with no offset/limit, returns the full file if it has at
+  most 2000 lines; otherwise returns the first 2000 lines and appends a
+  truncation note showing the total line count. Pass `offset` (1-indexed)
+  and/or `limit` to paginate larger files. `0` for either argument means
+  "unset" — Agency does not have undefined arguments.
+
   @param filename - The file to read
   @param dir - The directory to resolve the filename against (defaults to ".")
+  @param offset - 1-indexed line to start at (0 means start of file)
+  @param limit - Maximum number of lines to return (0 means default 2000)
```
egonSchiele: How would users specify reading the full file? I think the default should be to read the full file

packages/agency-lang/docs/site/stdlib/shell.md
```diff
@@ -160,6 +162,8 @@ Search for a regex pattern in files under a directory. Returns matches with file
 
   Relative `dir` is resolved against the calling Agency module's directory (same policy as `read`/`write`/`glob`/`ls`), and returned `file` values are relative to `dir` — so results compose with `read(file, dir)` directly.
 
+  Returns at most `maxResults` matches (default 200). Files larger than 5 MB are skipped silently. If you suspect matches were truncated, narrow `dir` or refine `pattern`.
```
egonSchiele:
```
Files larger than 5 MB are skipped silently
```

This is surprising behavior, remove

packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -56,22 +58,40 @@ let policy: Policy = {}
...
 
+def loadAgentsMd(dir: string): string {
+  """
+  Read `AGENTS.md` from `dir` if it exists and return it wrapped in a
+  <project_context> block ready to splice into a system prompt. Returns
+  an empty string if there is no AGENTS.md or it cannot be read. The
+  read is auto-approved — the user implicitly opted in to scanning the
+  project directory by running the agent there.
+  """
+  if (!exists("${dir}/AGENTS.md")) {
```
egonSchiele: should exists take a dir param too to mirror the other funcs?

PR #222, opened 2026-05-29
Add ~ expansion to stdlib path resolution

egonSchiele:
Add ~ expansion to stdlib path resolution

Lets users write `~/notes.md`, `~/.agency/memory`, etc. in any
filesystem-touching stdlib function and have `~` expand to the user's
home directory. Previously `~` was treated as a literal directory
name, silently doing the wrong thing.

The expansion policy lives in one resolver, not at call sites:

- `lib/stdlib/expandPath.ts` — pure string transform, single owner of
  user-path-shorthand policy. Today only handles leading `~`; future
  rules (env vars, normalization) land here.
- `lib/stdlib/resolveDir.ts` — the "dir-only" resolver shared by
  every stdlib site that previously wrote `path.resolve(moduleDir, dir)`.
  Runs `expandPath`, resolves against the module dir, asserts
  containment.
- `lib/stdlib/resolvePath.ts` — delegates its dir step to
  `resolveDir`, so `read`/`write`/`edit` get `~` support for free.
- `lib/stdlib/assertContained.ts` — expands `~` in each `allowedPaths`
  entry, so policies that say `allowedPaths: ["~/.agency"]` match.

Call sites in `shell.ts`, `fs.ts`, and `builtins.ts` collapse from
two-line resolve+assert pairs to a single `resolveDir(dir, allowed)`
call.

Documents the convention in `docs/dev/coding-standards.md`: any new
stdlib function taking a `dir` / `cwd` / `path` arg MUST resolve it
via `resolveDir` (or `resolvePath`).

Plan: docs/superpowers/plans/2026-05-29-tilde-expansion-in-stdlib.md

### DIFF
--- a/packages/agency-lang/docs/superpowers/plans/2026-05-29-tilde-expansion-in-stdlib.md
+++ b/packages/agency-lang/docs/superpowers/plans/2026-05-29-tilde-expansion-in-stdlib.md
@@ -1,0 +1,418 @@
+# Home-Directory (`~`) Expansion in Stdlib Path Resolution — Implementation Plan
+
+## Goal
+
+Let users write `~/notes.md`, `~/.agency/memory`, etc. in any
+filesystem-touching stdlib function and have `~` expand to the user's
+home directory. Today the stdlib treats `~` as a literal directory
+name, which silently does the wrong thing.
+
+**Stronger requirement:** any path-resolving stdlib function — present
+or future — should pick up `~` (and any later path-expansion rules)
+**by going through one resolver, not by re-implementing the policy at
+the call site.** This is the part of the plan that pays for itself
+the next time we add a path rule (env-var expansion, normalization,
+allow-list overlay, etc.).
+
+## Verification before writing code
+
+> Grep `lib/` for any existing path-expansion utility before adding a
+> new one. As of writing:
+>
+> - `lib/stdlib/path.ts` exposes only thin wrappers around node `path`
+>   (`_join`, `_resolve`, `_basename`, `_dirname`, `_extname`,
+>   `_relative`, `_isAbsolute`) — no expansion.
+> - `lib/stdlib/resolvePath.ts` resolves `(dir, filename)` against the
+>   module directory with traversal / symlink-escape checks. No `~`
+>   handling.
+> - `os.homedir()` is used ad-hoc in `lib/stdlib/oauth.ts`,
+>   `lib/cli/schedule/backends/{launchd,systemd}.ts`,
+>   `lib/cli/schedule/index.ts`, `lib/mcp/setup.ts`,
+>   `lib/serve/policyStore.ts`. None of those are a reusable helper.
+>
+> If something has been added since, prefer reusing it over creating
+> a new one.
+
+## Scope: which APIs
+
+Every stdlib entry point that resolves a filesystem path. Concretely:
+
+**Built-ins ([`lib/stdlib/builtins.ts`](../../../lib/stdlib/builtins.ts)):**
+- `_read`, `_write`, `_readImage` — `dir` arg (filename stays relative).
+
+**Shell tools ([`lib/stdlib/shell.ts`](../../../lib/stdlib/shell.ts)):**
+- `_ls`, `_grep`, `_glob`, `_stat`, `_exists` — `dir` arg.
+- `_exec`, `_bash` — `cwd` arg.
+- `_which` — scans PATH, no user dir input; skip.
+
+**Filesystem tools ([`lib/stdlib/fs.ts`](../../../lib/stdlib/fs.ts)):**
+- `_copy`, `_move`, `_remove`, `_mkdir`, `_edit`, `_applyPatch`,
+  etc. — every `src` / `dest` / `path` / `dir` arg.
+
+**Allow-list policy fields ([`lib/stdlib/assertContained.ts`](../../../lib/stdlib/assertContained.ts)):**
+- `allowedPaths: ["~/.agency", ...]` must expand the same way, or
+  policies break for users who write `~` in them.
+
+**Memory store path** (after the memory-config PR lands):
+- `dir` passed to `enableMemory({ dir: "~/.agency/memory" })`. Comes
+  for free because the memory plan resolves through this same helper.
+
+## Design: one resolver, no inlined policy
+
+### Anti-pattern this plan refuses
+
+The naive approach is to insert `expandHome(dir)` at every existing
+`path.resolve(moduleDir, dir)` site. That gives `~` expansion but
+encodes the path-resolution policy ("expand, then resolve against
+module dir") at ~15 call sites. The next rule we add (env vars, NFC
+normalization, allow-list overlay) has to be inserted at all 15
+sites again — and any *new* stdlib function added in the future has
+to remember to do the same dance. That is the
+[`inconsistent patterns`](../../dev/anti-patterns.md#inconsistent-patterns)
+anti-pattern about to be born.
+
+### What this plan does instead
+
+Two layered helpers; call sites consume the second.
+
+**Layer 1 — `expandPath(p)` in `lib/stdlib/expandPath.ts`** — pure
+string transform. Single source of truth for "user-typed path string →
+expanded string." Today it only handles leading `~`; the function
+exists so the next expansion rule (env vars, NFC, anything) lands in
+one place.
+
+```ts
+import os from "os";
+import path from "path";
+
+/**
+ * Expand user-shorthand prefixes in a path string. Currently:
+ *
+ * - `~` alone → $HOME
+ * - `~/foo` (or `~\foo` on Windows) → $HOME/foo
+ * - `~user/...` throws (POSIX-only, platform complexity not worth it).
+ * - everything else returns unchanged.
+ *
+ * This is the single owner of path-shorthand policy for the stdlib.
+ * Future rules (env-var expansion, NFC normalization, etc.) land
+ * here, not at call sites.
+ *
+ * Does NOT resolve to an absolute path — callers still pass the
+ * result through path.resolve / resolvePath / resolveDir.
+ */
+export function expandPath(p: string): string;
+```
+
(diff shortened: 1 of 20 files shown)


packages/agency-lang/lib/stdlib/expandPath.ts
```diff
@@ -0,0 +1,53 @@
...
+ * the policy locally.
+ *
+ * Does NOT resolve to an absolute path — callers still pass the
+ * result through `path.resolve` / `resolvePath` / `resolveDir`.
+ *
+ * `expandPath("")` returns `""` unchanged so empty-string sentinels
+ * pass through (callers that disallow empty dir handle that
+ * themselves with a clearer error).
+ */
+export function expandPath(p: string): string {
```
egonSchiele, line 37: why not write the parser for this using tarsec https://egonschiele.github.io/tarsec/ (what the agency lang parser uses)?

packages/agency-lang/lib/stdlib/expandPath.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const home = os.homedir();
+  if (!home) {
+    throw new Error(
+      "expandPath: cannot expand `~` because os.homedir() returned no value (no HOME env var on POSIX, no USERPROFILE on Windows).",
+    );
+  }
+
+  // p is now either `~` exactly, or `~` + separator + tail.
+  if (p === "~") return home;
```
egonSchiele, line 58: `~/` should work too

packages/agency-lang/lib/stdlib/expandPath.ts
egonSchiele, line 37: should this handle relative paths too? have all the logic in one place?

packages/agency-lang/lib/stdlib/fs.ts
```diff
@@ -247,10 +246,8 @@ export async function _move(
   dest: string,
   allowedPaths?: string[],
 ): Promise<void> {
-  const srcFull = path.resolve(process.cwd(), src);
-  const destFull = path.resolve(process.cwd(), dest);
-  await assertContained(srcFull, allowedPaths ?? []);
-  await assertContained(destFull, allowedPaths ?? []);
+  const srcFull = await resolveDir(src, allowedPaths ?? [], "cwd");
```
egonSchiele, line 248: Do we no longer need `assertContained` for these?

packages/agency-lang/lib/stdlib/expandPath.ts
```diff
@@ -0,0 +1,53 @@
...
+  // Strip the leading `~` plus the separator character (length 2).
+  const tail = p.slice(2);
+  return path.join(home, tail);
```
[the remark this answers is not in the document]
egonSchiele: Verified this is a false positive — Copilot is conflating `path.join` with `path.resolve`. `path.resolve(home, "/foo")` does drop the home prefix (the second arg becomes absolute), but `path.join(home, "/foo")` returns `home/foo`. Tested directly:

```
> path.join("/home/user", "/foo")
'/home/user/foo'
```

Added an explicit `~//foo` test case to lock in the behavior in case someone later swaps `join` for `resolve` (fixed in 6e3b6bb).

packages/agency-lang/lib/stdlib/resolvePath.ts
```diff
@@ -1,6 +1,7 @@
 import fs from "fs/promises";
 import path from "path";
-import { getModuleDir } from "../runtime/asyncContext.js";
+import { resolveDir } from "./resolveDir.js";
+import { expandPath } from "./expandPath.js";
```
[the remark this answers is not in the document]
egonSchiele: Good catch — fixed in 6e3b6bb. The expansion happens inside `resolveDir`, so the import here was leftover.

packages/agency-lang/lib/stdlib/resolvePath.ts
```diff
@@ -11,20 +12,29 @@ import { getModuleDir } from "../runtime/asyncContext.js";
...
  * test invocations.
  */
 export async function resolvePath(dir: string, filename: string): Promise<string> {
   if (!dir) {
     throw new Error(`dir must not be empty. Use "." for the current directory.`);
   }
-  if (path.isAbsolute(filename)) {
+  if (path.isAbsolute(filename) || filename.startsWith("~")) {
     throw new Error(`Filename must not be absolute when dir is specified (got "${filename}").`);
   }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 6e3b6bb. Message now reads `Filename must not be absolute or start with "~" when dir is specified (...)`.

packages/agency-lang/lib/stdlib/fs.ts
```diff
@@ -4,7 +4,9 @@ import path from "path";
 import process from "process";
 import diff_match_patch from "diff-match-patch";
 import { resolvePath } from "./resolvePath.js";
+import { resolveDir } from "./resolveDir.js";
 import { assertContained } from "./assertContained.js";
+import { expandPath } from "./expandPath.js";
 import { formatDiff } from "../utils/diff.js";
```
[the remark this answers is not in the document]
egonSchiele: Confirmed unused — fixed in 6e3b6bb. All allow-list enforcement in this file now flows through `resolveDir`, which calls `assertContained` internally.

packages/agency-lang/docs/superpowers/plans/2026-05-29-tilde-expansion-in-stdlib.md
```diff
@@ -0,0 +1,418 @@
...
+- `_read`, `_write`, `_readImage` — `dir` arg (filename stays relative).
+
+**Shell tools ([`lib/stdlib/shell.ts`](../../../lib/stdlib/shell.ts)):**
+- `_ls`, `_grep`, `_glob`, `_stat`, `_exists` — `dir` arg.
+- `_exec`, `_bash` — `cwd` arg.
+- `_which` — scans PATH, no user dir input; skip.
+
+**Filesystem tools ([`lib/stdlib/fs.ts`](../../../lib/stdlib/fs.ts)):**
+- `_copy`, `_move`, `_remove`, `_mkdir`, `_edit`, `_applyPatch`,
+  etc. — every `src` / `dest` / `path` / `dir` arg.
```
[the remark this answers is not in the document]
egonSchiele: Good call — extended the scope to actually match the plan rather than narrowing the words. Migrated in 6e3b6bb:

- `system.ts` — `_screenshot` (filepath)
- `speech.ts` — `_speak` (outputFile), `_record` (outputFile), `_transcribe` (filepath)
- `policy.ts` — `_writePolicyFile` (filePath)

Each was a `path.resolve(process.cwd(), x)` + `assertContained` pair that collapsed to a single `resolveDir(x, allowedPaths, "cwd")` call. The structural lint mentioned in the plan would have caught these — still worth filing as a follow-up.

packages/agency-lang/tests/agency/tilde-paths/tilde.agency
```diff
@@ -0,0 +1,23 @@
...
+// to exist for any user) so the test is side-effect-free and never
+// writes under the user's home.
+node main(): boolean {
+  const homeStat = stat(`~`, ``)
+  if (!homeStat.exists) {
+    return false
+  }
+  if (homeStat.type != `dir`) {
+    return false
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 6e3b6bb — the assertion now accepts `dir` or `symlink`, with a comment explaining why (`_stat` uses `lstat`, and `$HOME` resolves through a symlink on macOS).

packages/agency-lang/lib/stdlib/expandPath.ts
egonSchiele: Tarsec would be a great fit once we have actual grammar to parse — `$VAR`, `${VAR}`, mixed-in literals, etc. Today the whole policy is `startsWith("~")` + a separator probe, so a combinator parser would be ~10 lines of plumbing wrapping one branch.

I'd suggest reaching for tarsec when we add env-var expansion. Happy to file that as a follow-up if you want to track it.

packages/agency-lang/lib/stdlib/expandPath.ts
egonSchiele: Verified — `~/` already works (falls through to `path.join(home, "") === home`), but it wasn't covered by a test, which is probably why it looked uncertain. Added an explicit test in 6e3b6bb:

```ts
it("expands ~/ (trailing slash only) to os.homedir()", () => {
  expect(expandPath("~/")).toBe(os.homedir());
});
```

packages/agency-lang/lib/stdlib/expandPath.ts
egonSchiele: Deliberate two-layer split:

- **`expandPath`** — pure string transform. No async, no ALS, no base directory. Trivially testable.
- **`resolveDir`** — picks the base (`moduleDir` for stdlib I/O that ships next to the `.agency` file, `cwd` for `fs.ts`/`shell.ts` ops that should mean "the user's working directory"), then asserts containment.

Collapsing relative-path resolution into `expandPath` would force it to reach into runtime context and pick a base for callers that have different opinions about which base is right. The "one place" for the overall policy is `resolveDir` — `expandPath` is the first stage inside it.

Tightened the docstring on `expandPath` in 6e3b6bb to make the layering explicit so future readers see it.

packages/agency-lang/lib/stdlib/fs.ts
egonSchiele: Right — `resolveDir` runs `assertContained(root, allowedPaths, baseDir)` internally on every call ([resolveDir.ts:38](https://github.com/egonSchiele/agency-lang/blob/feature/tilde-expansion/packages/agency-lang/lib/stdlib/resolveDir.ts#L38)), so the call sites get containment-checking for free.

The now-stale `assertContained` import on line 8 (also flagged by Copilot) was dropped in 6e3b6bb.

PR #13, opened 2026-04-15
Add first wave of stdlib coding-agent tools

egonSchiele:
## Summary

Adds the primitives an Agency-based coding agent needs to navigate and edit codebases. Modeled on the tool set used by OpenCode / Claude Code, organized into dedicated stdlib modules so users can import by namespace (e.g. `import { edit } from "std::fs"`).

### New modules
- **`std::fs`** — `edit` (unique-match search/replace), `multiedit` (atomic batch), `applyPatch` (unified diff).
- **`std::shell`** — renamed from `std::process`. Adds `ls`, `grep`, `glob`, `stat`, `exists` alongside `bash`. Also exports `bash` (was previously unexported — preexisting oversight).
- **`std::path`** — `join`, `resolve`, `basename`, `dirname`, `extname`, `relative`, `isAbsolute`.
- **`std::http`** — `fetch`, `fetchJSON`, `webfetch` (HTML→markdown for LLM consumption).
- **`std::agent`** — `todoWrite`, `todoList`, `question` (interrupt-based prompt, distinct from stdin-based `input`).

### Compatibility
`read`, `write`, `readImage`, `fetch`, `fetchJSON` remain in `index.agency` (auto-imported), so existing `.agency` code is untouched. The new modules are purely additive.

### Notes
- `bash`'s signature simplified (dropped `env`/`maxBuffer`) because Agency's parser does not currently support `param?: type`; the original signature never actually parsed.
- Inspection tools (`ls`, `grep`, `glob`) gate with `interrupt({...})` for permission prompts; cheap metadata tools (`stat`, `exists`, all `path` helpers) are `safe` and ungated.

## Test plan
- [x] `make stdlib` — all new TS files compile
- [x] `pnpm run compile` — sanity `.agency` importing from every new module compiles clean
- [x] `pnpm run agency` — read-only tools (`stat`, `exists`, path helpers) run end-to-end
- [x] `pnpm run test:run` — full suite: 1811 passing, 21 skipped, no regressions
- [ ] Manual: try `edit`/`multiedit`/`applyPatch` on a real file (interrupt-gated, so needs an approval handler)
- [ ] Manual: try `webfetch` on an HTML page

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #223, opened 2026-05-29
Add memory configuration in Agency code

egonSchiele:
Add memory configuration in Agency code

Lets Agency code enable, configure, and disable memory directly,
without requiring a `memory` block in `agency.json`. Supports per-thread
/ per-fork memory by moving memory configuration onto the stateStack.

New stdlib API in `stdlib/memory.agency`:

  enableMemory({ dir, model?, autoExtract?, compaction?, embeddings? })
  disableMemory()
  memory({ ... }) as { ... }

All three are also valid tools an LLM can be passed.

Architecture:

- `StateStack` carries a stack of memory frames, hidden behind
  `pushMemoryFrame` / `popMemoryFrame` / `topMemoryFrame`. No caller
  touches the array. Lives in `stack.other` so it serializes for free
  and forks correctly.
- Single source of truth: JSON config is just the bottom frame. At
  `createExecutionContext`, the JSON config is pushed onto the new
  execCtx's stateStack. After that, the stack is the only thing
  anyone consults. "Code wins over JSON" is structural, not coded.
- Process-wide store registry in `lib/runtime/memory/registry.ts`,
  keyed by absolute realpath'd dir. Multiple call sites with the
  same dir share storage.
- Active manager resolved via single
  `ctx.getActiveMemoryManager()` — top of stack or undefined, no
  fallback branch.

`agency.memory.*` namespace added for TS helpers so they have
symmetric access to what Agency code can do.

Dir resolution mirrors `agency.json`: relative `dir` resolves against
`process.cwd()`. `~` expansion lands separately in
docs/superpowers/plans/2026-05-29-tilde-expansion-in-stdlib.md.

Plan: docs/superpowers/plans/2026-05-29-memory-config-in-code.md

### DIFF
--- a/packages/agency-lang/docs/site/guide/memory.md
+++ b/packages/agency-lang/docs/site/guide/memory.md
@@ -158,18 +158,19 @@ Because each fork branch has its own state stack, frames pushed in
 one branch never bleed into siblings:
 
 ```ts
-fork {
-  branch {
-    memory({ dir: "./mem-a" }) as { remember("alice fact") }
-  }
-  branch {
-    memory({ dir: "./mem-b" }) as { remember("bob fact") }
+const dirs = ["./mem-a", "./mem-b"]
+fork(dirs) as dir {
+  memory({ dir: dir }) as {
+    remember("a fact scoped to this branch")
   }
 }
 ```
 
-Both branches share the process-wide store cache, so two branches
-pushing the same `dir` share one `MemoryManager` instance.
+Both branches share the process-wide *store* cache, so two branches
+that resolve to the same physical `dir` share one `FileMemoryStore`
+instance. Each execCtx still wraps that store in its own
+`MemoryManager` (with its own statelog client and log level), so the
+sharing is at the on-disk layer only.
 
 ### `setMemoryId` is orthogonal
 
--- a/packages/agency-lang/docs/site/guide/memory.md
+++ b/packages/agency-lang/docs/site/guide/memory.md
@@ -84,6 +84,102 @@ All other fields are optional with sensible defaults:
 | `compaction.threshold` | Threshold above which compaction runs. |
 | `embeddings.model` | Embedding model name (forwarded to smoltalk's `embed`). |
 
+## Configuring memory in code
+
+You can also enable, reconfigure, or disable memory directly from
+Agency code. This is useful for multi-tenant agents that want a
+different store per user, library helpers that want a side store
+without affecting the caller, and fork branches that each want their
+own memory.
+
+Three functions live in `std::memory`:
+
+```ts
+import { enableMemory, disableMemory, memory } from "std::memory"
+```
+
+### `enableMemory(config)`
+
+Push a memory frame onto the active branch. Subsequent
+`remember`/`recall`/`forget` calls in this branch use the new
+config.
+
+```ts
+enableMemory({ dir: "./mem-user-alice" })
+remember("alice's favourite colour is blue")
+```
+
+- `dir` is resolved against the **current working directory** (the
+  same as `agency.json`'s `memory.dir`, deliberately NOT the module
+  dir like `read`/`write`).
+- The directory is auto-created.
+- Storage is shared process-wide by absolute dir, so multiple calls
+  (or multiple runs) using the same `dir` share one underlying
+  store.
+- Pushing the same `dir` as the current top is a no-op (safe to call
+  in `main()` even when a `static const _ = enableMemory({...})`
+  already ran).
+- Pushing a different `dir` stacks the new frame on top.
+
+### `disableMemory()`
+
+Pop the top memory frame from the active branch. Frame-scoped: a
+`disableMemory()` inside a fork branch only affects that branch.
+
+> **Warning:** `disableMemory()` pops whatever is on top, including
+> the JSON-seeded bottom frame from `agency.json`. Library authors
+> should not call this casually — they will shadow the caller's
+> configured memory. Prefer the block form below for lexical scoping.
+
+### `memory(config) as { ... }`
+
+The block form. Pushes a frame on entry, pops it on exit (including
+on throw / interrupt).
+
+```ts
+memory({ dir: "./mem-user-alice" }) as {
+  remember("alice's favourite colour is blue")
+  print(recall("alice"))
+}
+// Frame is popped here — the next remember/recall sees whatever was
+// active before this block.
+```
+
+### Precedence
+
+Code wins over `agency.json`. The JSON config is seeded as the
+bottom-of-stack frame on every run; any `enableMemory(...)` or
+`memory({...}) as { ... }` pushes on top. The active frame is always
+the top of the stack.
+
+### Per-fork memory
+
+Because each fork branch has its own state stack, frames pushed in
+one branch never bleed into siblings:
+
+```ts
+fork {
+  branch {
+    memory({ dir: "./mem-a" }) as { remember("alice fact") }
+  }
+  branch {
+    memory({ dir: "./mem-b" }) as { remember("bob fact") }
+  }
+}
+```
+
+Both branches share the process-wide store cache, so two branches
+pushing the same `dir` share one `MemoryManager` instance.
+
+### `setMemoryId` is orthogonal
+
+`setMemoryId(...)` updates the active scope id (lives on the
+state stack's `other.memoryId`) and is *independent* of which frame
+is on top. A library helper that opens a side store does not
+clobber the caller's `setMemoryId`. If you want a fresh scope when
+switching stores, call `setMemoryId(...)` explicitly inside the
+block.
+
 ## The `std::memory` module
 
 Import the functions you need from `std::memory`:
--- a/packages/agency-lang/docs/site/stdlib/memory.md
+++ b/packages/agency-lang/docs/site/stdlib/memory.md
@@ -31,13 +31,13 @@ type ForgetResult = {
 export type MemoryConfig = {
   dir: string;
   model?: string;
-  autoExtract?: { interval: number };
-  compaction?: { trigger: string; threshold: number };
-  embeddings?: { model: string }
+  autoExtract?: { interval: number | undefined };
+  compaction?: { trigger: "token" | "messages" | undefined; threshold: number | undefined };
+  embeddings?: { model: string | undefined }
 }
 ```
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/memory.agency#L20))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/memory.agency#L22))
 
 ## Functions
 
@@ -67,7 +67,7 @@ Set the memory scope for this agent run. Call this before other
 |---|---|---|
 | id | `string` |  |
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/memory.agency#L45))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/memory.agency#L48))
 
 ### enableMemory
 
@@ -104,7 +104,7 @@ Turn memory on for the current execution branch using `config`.
 |---|---|---|
 | config | [MemoryConfig](#memoryconfig) |  |
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/memory.agency#L64))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/memory.agency#L67))
 
 ### disableMemory
 
@@ -123,22 +123,31 @@ Pop the top memory frame from the current branch's stateStack.
   Prefer the block form `memory({...}) as { ... }` for lexical
   scoping, which restores the previous frame on exit.
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/memory.agency#L92))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/memory.agency#L95))
 
 ### memory
 
 ```ts
-memory(config: MemoryConfig, block: () => any): any
+memory(config: MemoryConfig, block: () => any): Result
 ```
 
 Run `block` with `config` pushed as the active memory frame; pop
-  the frame when the block returns (or throws).
+  the frame when the block returns (or fails). Mirrors the `guard()`
+  pattern in `std::thread` so push/pop balancing flows through the
+  same `try` expression — Agency has no `finally`, so the agency-side
+  body uses `try block()` to capture the result and runs `pop` after.
 
-  The block form of `enableMemory` / `disableMemory`. Prefer this
-  whenever the memory frame should be lexically scoped.
+  The pop only fires when the push actually happened: `_pushMemoryFrame`
+  returns `false` if the new frame's dir matches the current top
+  (dedup), and the matching no-pop preserves the caller's frame
+  instead of unbalancing it. This fixes the prior `_memoryBlock`
+  unbalanced-pop bug.
+
+  Returns a `Result` (success holds the block's return value, failure
+  holds an error from inside the block). Same convention as `guard()`.
 
   Example:
-    memory({ dir: "./mem-user-a" }) as {
+    const r = memory({ dir: "./mem-user-a" }) as {
       remember("alice's favorite color is blue")
     }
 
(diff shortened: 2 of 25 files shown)


packages/agency-lang/docs/site/guide/memory.md
```diff
@@ -84,6 +84,102 @@ All other fields are optional with sensible defaults:
...
+the top of the stack.
+
+### Per-fork memory
+
+Because each fork branch has its own state stack, frames pushed in
+one branch never bleed into siblings:
+
+```ts
+fork {
+  branch {
```
egonSchiele: This isn't how fork works

packages/agency-lang/docs/site/stdlib/memory.md
```diff
@@ -25,6 +25,20 @@ type ForgetResult = {
...
 ([source](https://github.com/egonSchiele/agency-lang/tree/main/stdlib/memory.agency#L12))
 
+### MemoryConfig
+
+```ts
+export type MemoryConfig = {
+  dir: string;
+  model?: string;
+  autoExtract?: { interval: number };
+  compaction?: { trigger: string; threshold: number };
```
egonSchiele: should `trigger` be an enum?

packages/agency-lang/lib/runtime/memory/frame.ts
```diff
@@ -0,0 +1,50 @@
...
+ *    so the same string in JSON and in code points at the same
+ *    place).
+ *  - Auto-creates the directory tree if missing.
+ *  - Resolves symlinks via `fs.realpathSync` so two different paths
+ *    that point at the same physical dir dedupe through the
+ *    process-wide store registry.
+ *
+ * Throws if `config.dir` is empty.
+ */
+export function normalizeMemoryFrame(config: MemoryConfig): MemoryFrame {
```
egonSchiele: IMO MemoryFrame should be a class, and this func should be its constructor

packages/agency-lang/lib/runtime/memory/registry.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Keys are absolute paths from `normalizeMemoryFrame` (which mkdirs
+ * and resolves against `process.cwd()`); callers must NOT pass raw
+ * user input here. Tests can call `_resetStoreRegistry()` between
+ * cases to start clean.
+ */
+import { FileMemoryStore } from "./store.js";
+import type { LogLevel } from "../../logger.js";
+
+const stores: Record<string, FileMemoryStore> = {};
```
egonSchiele, line 52: Does this work with the execution model? What happens if two different agent runs use the same absDir? Is the absDir related to the memory id? https://agency-lang.com/guide/execution-model.html

packages/agency-lang/lib/runtime/state/context.ts
```diff
@@ -265,44 +274,107 @@ export class RuntimeContext<T> {
...
+   * for the "pop back to A returns A's manager" semantics.
+   */
+  getActiveMemoryManager(): MemoryManager | undefined {
+    if (!this.stateStack) return undefined;
+    let frame = this.stateStack.topMemoryFrame();
+    if (!frame && this.jsonMemoryConfig) {
+      // Old-checkpoint back-compat: stateStack restored without any
+      // memoryFrames at all. Re-seed the JSON bottom frame so
+      // resume behaves like a fresh run.
+      const memoryFramesArr = this.stateStack.other.memoryFrames as
```
egonSchiele: breaks encapsulation, also isn't this the same as `this.stateStack.topMemoryFrame()`?

egonSchiele: Also do we actually need to add a frame? Or can we just create a memory manager with the json config, but not pop a frame?

packages/agency-lang/lib/runtime/state/stateStack.ts
```diff
@@ -329,6 +330,49 @@ export class StateStack {
...
+  popMemoryFrame(): MemoryFrame | undefined {
+    const existing = this.other.memoryFrames as MemoryFrame[] | undefined;
+    if (!existing || existing.length === 0) return undefined;
+    const popped = existing[existing.length - 1];
+    this.other.memoryFrames = existing.slice(0, -1);
+    return popped;
+  }
+
+  /** The active memory frame, or `undefined` if memory is currently off. */
+  topMemoryFrame(): MemoryFrame | undefined {
```
egonSchiele: would `activeMemoryFrame` be more descriptive?

packages/agency-lang/lib/runtime/state/stateStack.ts
```diff
@@ -329,6 +330,49 @@ export class StateStack {
...
+    const frames = this.other.memoryFrames as MemoryFrame[] | undefined;
+    if (!frames || frames.length === 0) return undefined;
```
egonSchiele: `frames.at(-1)` is more concise

packages/agency-lang/lib/runtime/state/stateStack.ts
```diff
@@ -329,6 +330,49 @@ export class StateStack {
...
+   *  When the last frame is popped the array stays present (but
+   *  empty) so `topMemoryFrame()` returns `undefined` and the
+   *  old-checkpoint re-seed in `getActiveMemoryManager()` knows the
+   *  user explicitly disabled memory rather than restored an
+   *  ancient snapshot. */
+  popMemoryFrame(): MemoryFrame | undefined {
+    const existing = this.other.memoryFrames as MemoryFrame[] | undefined;
+    if (!existing || existing.length === 0) return undefined;
+    const popped = existing[existing.length - 1];
+    this.other.memoryFrames = existing.slice(0, -1);
```
egonSchiele: why not just do `return existing?.pop()`

packages/agency-lang/lib/runtime/state/stateStack.ts
```diff
@@ -329,6 +330,49 @@ export class StateStack {
...
+   *
+   * Same-`configKey` as top is a no-op so the common pattern
+   * `static const _ = enableMemory(...)` + an `enableMemory(...)`
+   * in `main()` doesn't double-stack the same frame. Different-dir
+   * pushes stack on top; pop with `popMemoryFrame()` or use the
+   * `memory(){}` block for lexical scoping.
+   */
+  pushMemoryFrame(frame: MemoryFrame): void {
+    const top = this.topMemoryFrame();
+    if (top?.configKey === frame.configKey) return;
```
egonSchiele: If memory frame becomes a class, it could get a .equals() func

packages/agency-lang/lib/runtime/state/stateStack.ts
```diff
@@ -329,6 +330,49 @@ export class StateStack {
...
+    const existing = (this.other.memoryFrames as MemoryFrame[] | undefined) ?? [];
+    this.other.memoryFrames = [...existing, frame];
```
egonSchiele: why not just push to the existing array?

packages/agency-lang/lib/stdlib/memory.ts
```diff
@@ -108,62 +121,139 @@ export async function __internal_forget(
...
+ * `enableMemory` / `disableMemory` for lexical scoping.
+ *
+ * The push/pop is implemented in TS rather than Agency `try/finally`
+ * because Agency has no `finally`; an explicit TS wrapper is the
+ * smallest correct construct. `block` is the AgencyFunction the
+ * compiler synthesises for `memory({...}) as { ... }` — invoke it
+ * through `AgencyFunction.invoke` so its frame, callsite, and
+ * interrupt machinery hook up correctly.
+ */
+export async function _memoryBlock(
```
egonSchiele: extremely unnecessary. Why not just define the block func entirely in agency, and not do it in TS?

packages/agency-lang/tests/agency/memory-code/enable.agency
```diff
@@ -0,0 +1,14 @@
+import { enableMemory, recall, setMemoryId } from "std::memory"
+
+// Exercises the new code-side memory configuration. enableMemory
+// pushes a frame onto the stateStack; setMemoryId scopes it.
+// Recall on a fresh, empty graph returns the empty string without
+// any LLM call (the manager short-circuits on graph.entities=0),
+// which lets this test prove the new wiring works end-to-end without
+// requiring mocks.
+node main(): boolean {
```
egonSchiele, line 11: all these tests would pass even if the memory functionality was broken

packages/agency-lang/lib/stdlib/memory.ts
```diff
@@ -108,62 +121,139 @@ export async function __internal_forget(
...
+    }
+    return Promise.resolve((block as () => unknown)());
+  };
+  const { stack } = getRuntimeContext();
+  if (!stack) return await callBlock();
+  stack.pushMemoryFrame(normalizeMemoryFrame(config));
+  try {
+    return await callBlock();
+  } finally {
+    stack.popMemoryFrame();
```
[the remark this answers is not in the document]
egonSchiele: Good catch — fixed in cf7dcd8a. `_memoryBlock` is gone; the block is now defined in Agency (`stdlib/memory.agency::memory`), and `_pushMemoryFrame` returns a boolean. The Agency body only calls `_popMemoryFrame` if the push actually happened, so a dedup-no-op push no longer eats the caller's frame. Added a unit test for the boolean and a stronger integration test (see comment #16).

packages/agency-lang/stdlib/memory.agency
```diff
@@ -14,6 +14,17 @@ type ForgetResult = {
...
 
+// User-facing memory config shape. Mirrors the `memory:` block in
+// `agency.json` so the same fields work whether you configure memory
+// declaratively or in code.
+export type MemoryConfig = {
+  dir: string,
+  model?: string,
+  autoExtract?: { interval: number },
+  compaction?: { trigger: string, threshold: number },
+  embeddings?: { model: string }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in cf7dcd8a — every nested field on `MemoryConfig` in `stdlib/memory.agency` is now optional, matching `MemoryConfig` in `lib/runtime/memory/types.ts`. `compaction.trigger` is now a literal union `"token" | "messages"` (see #5).

packages/agency-lang/docs/site/guide/memory.md
```diff
@@ -84,6 +84,102 @@ All other fields are optional with sensible defaults:
...
+    memory({ dir: "./mem-a" }) as { remember("alice fact") }
+  }
+  branch {
+    memory({ dir: "./mem-b" }) as { remember("bob fact") }
+  }
+}
+```
+
+Both branches share the process-wide store cache, so two branches
+pushing the same `dir` share one `MemoryManager` instance.
```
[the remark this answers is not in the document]
egonSchiele: Fixed in cf7dcd8a. Updated `docs/site/guide/memory.md` to say the **store** is shared process-wide; each execCtx still wraps that store in its own `MemoryManager` with its own statelog client + log level. Also added a longer note to `lib/runtime/memory/registry.ts` covering the same distinction (see #7).

packages/agency-lang/docs/site/guide/memory.md
egonSchiele: Fixed in cf7dcd8a — rewrote the fork example in `docs/site/guide/memory.md` to use the correct `fork(items) as item { ... }` form.

packages/agency-lang/docs/site/stdlib/memory.md
egonSchiele: Yep, fixed in cf7dcd8a. `compaction.trigger` is now `"token" | "messages"` in `stdlib/memory.agency`.

packages/agency-lang/lib/runtime/memory/frame.ts
egonSchiele: Done in cf7dcd8a. `MemoryFrame` is now a class; the resolve/mkdir/realpath policy lives in the constructor. `normalizeMemoryFrame` is gone — all callers (`_enableMemory`, `RuntimeContext.createExecutionContext`) do `new MemoryFrame(config)`. Also added `MemoryFrame.equals` (static) per #13.

packages/agency-lang/lib/runtime/memory/registry.ts
egonSchiele: Sharing a `FileMemoryStore` across runs that point at the same `absDir` is intentional — that is what makes file-backed memory persist across runs and be visible to concurrent agents working on the same workspace. The store is shared *state*, not a deterministic-replay participant. `absDir` (where the store lives) is orthogonal to `memoryId` (which scope inside it). I expanded the top-of-file doc in `lib/runtime/memory/registry.ts` (cf7dcd8a) covering all three points and linking the execution-model guide.

packages/agency-lang/lib/runtime/state/context.ts
egonSchiele: Encapsulation: agreed, fixed in cf7dcd8a. Added `StateStack.hasMemoryFrameStack()`; `getActiveMemoryManager` now calls that instead of poking `other.memoryFrames` directly. The two checks aren't equivalent though — `activeMemoryFrame()` returns the top frame or `undefined`, whereas `hasMemoryFrameStack()` is true even when the array has been popped to empty (the "user explicitly disabled memory" case). Distinguishing those is what keeps the old-checkpoint re-seed from resurrecting memory the user just turned off.

egonSchiele: I went back and forth on this and kept the frame for two reasons:
1. `disableMemory()` semantics. If the JSON config isn't a frame, the user calling `disableMemory()` either (a) can't turn JSON-configured memory off at all, or (b) we need a separate "JSON override" flag. Treating JSON as the bottom frame means `disableMemory()` is just "pop one" with one rule, no special cases.
2. Symmetry. `enableMemory(...)` in code and `"memory": {...}` in `agency.json` should be interchangeable — same dir resolves to the same `FileMemoryStore`, same lifecycle, same priority. Making one a frame and the other a side-channel manager would re-introduce the asymmetry the feature was designed to remove.
The sentinel-array trick (`hasMemoryFrameStack()`) is what keeps "user popped the JSON frame" distinguishable from "old checkpoint without any frames at all," so old-checkpoint behavior is preserved without resurrecting a frame the user explicitly disabled.

packages/agency-lang/lib/runtime/state/stateStack.ts
egonSchiele: Renamed in cf7dcd8a — `topMemoryFrame` → `activeMemoryFrame` everywhere (StateStack, tests, callers, comments).

packages/agency-lang/lib/runtime/state/stateStack.ts
egonSchiele: Done in cf7dcd8a — `activeMemoryFrame()` now just returns `this.memoryFramesArr()?.at(-1)`.

packages/agency-lang/lib/runtime/state/stateStack.ts
egonSchiele: Done in cf7dcd8a — `popMemoryFrame` is now `return this.memoryFramesArr()?.pop()`. `.pop()` already returns `undefined` on an empty array, so no extra guard needed.

packages/agency-lang/lib/runtime/state/stateStack.ts
egonSchiele: Done in cf7dcd8a. `MemoryFrame.equals(a, b)` is static (not an instance method) because frames survive a JSON round-trip through StateStack serialization, which drops the class prototype — a restored plain `{configKey, config}` shape would `TypeError` on `frame.equals(other)`. The static form works on both fresh instances and restored shapes. Added an explicit unit test for the JSON-roundtrip case.

packages/agency-lang/lib/runtime/state/stateStack.ts
egonSchiele: Done in cf7dcd8a — `pushMemoryFrame` now mutates with `existing.push(frame)` instead of cloning. The only initialization is the first-push case where the array doesn't exist yet; once it does, we mutate. (That first-push assignment also doubles as the sentinel that flips `hasMemoryFrameStack()` to true — see #8.)

packages/agency-lang/lib/stdlib/memory.ts
egonSchiele: Done in cf7dcd8a. The block now lives entirely in `stdlib/memory.agency::memory` and uses `try block()` to capture the result, mirroring the `guard()` pattern from `std::thread`. TS only exposes `_pushMemoryFrame` (returns a boolean for dedup) and `_popMemoryFrame`; the Agency body skips the pop when the push was a no-op (fixes the Copilot-flagged unbalanced-pop bug in #1).

packages/agency-lang/tests/agency/memory-code/enable.agency
egonSchiele: Fair point — fixed in cf7dcd8a. Added a public `isMemoryActive()` to `std::memory` (wraps the existing `_shouldRunMemory`) and rewrote both integration tests to use it. `block.agency` now asserts `isMemoryActive()` is false before the block, true inside, false after — that fails if the push silently no-op'd, if the pop didn't fire, or if both did. `enable.agency` similarly asserts the flag before `enableMemory`, after `enableMemory`, and after `disableMemory`. The recall-returns-empty check is still there as a manager-wiring smoke test, but it's no longer the *only* signal.

PR #224, opened 2026-05-29
Add `agency literate weave` command

egonSchiele:
## Summary

Adds a new `agency literate weave` command that renders Agency source as
literate-programming markdown. Block comments (`/* */` and `/** */`)
become prose; everything else stays in a code fence.

```
agency literate weave <input> [-o <outputDir>] [--ignore <dirs...>] [--lang <name>]
```

- `<input>` is a file or directory. Directory input mirrors the relative
  tree under `outputDir`.
- Output dir defaults to `literate`.
- Code-fence language tag defaults to `agency`; pass `--lang typescript`
  for syntax highlighting on renderers that do not yet know `agency`.

This is intentionally distinct from `agency doc`, which produces API
reference output reorganized by symbol kind. `literate` is a narrative
reproduction of the file in source order.

## Design notes

- **Declarative classifier.** `lib/cli/literate.ts` is built as a
  three-stage pipeline (`kindOf` -> `resolveEmpty` -> `groupRuns`) and
  then mapped to segments. No mutable scratch state, no order-dependent
  flush callbacks.
- **Skip the preprocessor.** The literate weaver calls `parse()` only;
  running `TypescriptPreprocessor` would mutate the AST (attach doc
  comments to declarations, etc.) and break source-order rendering.
- **Skip the parser template.** `applyTemplate: false` is passed to
  `parse()` so the implicit `import { ... } from "std::index"` prelude
  does not show up in every weaved file.
- **JSDoc star-stripping.** Multi-line block comments have their
  leading ` * ` decoration removed so prose renders cleanly. Single-line
  comments and inner `*` characters are untouched, so markdown bullets
  and arithmetic survive.

## Side-improvement: `codeFence` auto-escalation

`lib/utils/markdown.ts:codeFence` now picks the fence delimiter as
`max(3, longest backtick run in code + 1)` instead of hardcoding three
backticks. Backward-compatible; every existing caller becomes correct
for inputs that happen to contain triple-backticks.

## Tests

- `lib/cli/literate.test.ts` (14 cases): block/doc/line comments,
  adjacent block merging, `--lang` flow-through, directory tree mirror,
  empty input, fence escalation (3-, 4-, 5-backtick), JSDoc star
  stripping, markdown-bullet preservation, single-line passthrough, and
  a real-file smoke test against `stdlib/markdown.agency`.
- `lib/utils/markdown.test.ts` (4 new cases): default + escalation +
  trim behavior for `codeFence`.
- `tests/integration/cli/test.mjs` (Test 5): end-to-end smoke through
  the published CLI.

## Out of scope

- `agency literate tangle` (markdown -> agency).
- Round-trip / idempotency.
- Local-import linking.
- A `literate?` config key in `agency.json` (defaults are inline on the
  commander options; YAGNI for now).

### DIFF
--- a/packages/agency-lang/lib/cli/literate.ts
+++ b/packages/agency-lang/lib/cli/literate.ts
@@ -113,7 +113,14 @@ function classify(program: AgencyProgram): Segment[] {
 
 function renderSegment(seg: Segment, lang: string): string {
   if (seg.kind === "prose") return seg.text;
-  const code = generateAgency({ type: "agencyProgram", nodes: seg.nodes });
+  // `preserveOrder: true` keeps imports in their source position rather
+  // than hoisting them to the top of the segment and sorting them. We
+  // need this for literate output to faithfully reproduce the file in
+  // source order.
+  const code = generateAgency(
+    { type: "agencyProgram", nodes: seg.nodes },
+    { preserveOrder: true },
+  );
   return codeFence(code, lang); // codeFence handles trimEnd + fence escalation
 }
 
--- a/packages/agency-lang/lib/cli/literate.ts
+++ b/packages/agency-lang/lib/cli/literate.ts
@@ -1,0 +1,165 @@
+import { AgencyConfig } from "@/config.js";
+import { generateAgency } from "@/backends/agencyGenerator.js";
+import { parse, readFile } from "./commands.js";
+import { findRecursively } from "./util.js";
+import { AgencyNode, AgencyProgram, AgencyMultiLineComment } from "@/types.js";
+import { codeFence } from "@/utils/markdown.js";
+import * as fs from "fs";
+import * as path from "path";
+
+type Kind = "prose" | "empty" | "code";
+type Tagged = { kind: "prose" | "code"; node: AgencyNode };
+type Segment =
+  | { kind: "prose"; text: string }
+  | { kind: "code"; nodes: AgencyNode[] };
+
+// --- classification pipeline (declarative) -------------------------------
+
+/** Base kind of a single node, ignoring context. */
+function kindOf(node: AgencyNode): Kind {
+  if (node.type === "multiLineComment") return "prose"; // /* */ and /** */
+  if (node.type === "newLine") return "empty";
+  return "code";
+}
+
+/** Nearest non-empty neighbor in `dir` direction. */
+function neighbor(
+  nodes: AgencyNode[],
+  from: number,
+  dir: 1 | -1,
+): AgencyNode | undefined {
+  for (let i = from + dir; i >= 0 && i < nodes.length; i += dir) {
+    if (kindOf(nodes[i]) !== "empty") return nodes[i];
+  }
+  return undefined;
+}
+
+/**
+ * Promote `empty` nodes to `code` when both neighbors are code (blank lines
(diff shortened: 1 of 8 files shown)


packages/agency-lang/lib/utils/markdown.ts
```diff
@@ -3,7 +3,10 @@ export function heading(level: number, text: string): string {
 }
 
 export function codeFence(code: string, lang: string = "ts"): string {
-  return `\`\`\`${lang}\n${code.trimEnd()}\n\`\`\``;
+  const matches = code.match(/`+/g) ?? [];
```
egonSchiele, line 6: Is this an aesthetic choice? Only 3 backticks are needed, right?

egonSchiele: Correctness, not aesthetics. The CommonMark closing-fence rule says a fenced code block ends at the first line that consists of nothing but backticks (>= the opening count). So if the fenced content contains a *line* that is just ` ``` ` — possible in Agency code via a doc-string or multi-line string that embeds markdown — three backticks would terminate the outer fence prematurely and the rest of the code would leak into the markdown body.

Inline backticks within a line (`` `foo` ``) do *not* terminate a fence, so most Agency code is fine. The escalation only ever fires when needed; for everything else the output is unchanged (`max(3, longest + 1)` returns 3).

Happy to revert to a hardcoded `"```"` if you'd rather; the cost of being wrong is a broken markdown render on contrived inputs, which is rare. Your call.

packages/agency-lang/lib/cli/literate.ts
```diff
@@ -0,0 +1,165 @@
...
+  });
+}
+
+// --- rendering ------------------------------------------------------------
+
+function renderSegment(seg: Segment, lang: string): string {
+  if (seg.kind === "prose") return seg.text;
+  const code = generateAgency({ type: "agencyProgram", nodes: seg.nodes });
+  return codeFence(code, lang); // codeFence handles trimEnd + fence escalation
+}
```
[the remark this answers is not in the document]
egonSchiele: Good catch — verified and fixed in db47e10d.

Verification (before the fix):

    import { zebra } from "z"
    import { apple } from "a"

    def hello(): string { ... }

Wove as `apple` before `zebra` (alphabetical) and dropped the blank line between imports and the function.

Went with option 2 from the suggestion (generator mode) rather than option 1 (slicing original source via `loc` ranges) because `importStatement` and `multiLineComment` nodes don't carry `loc`, so slicing isn't workable today.

Fix:
- Added `preserveOrder?: boolean` to `AgencyGenerator` and `generateAgency()`.
- When set: skip `partitionImports`, skip `sortAndRenderImports`, and render imports in-place via `processImportStatement` / `processImportNodeStatement` from inside `processNodeInner`.
- Default is `false`, so `agency doc`, `agency fmt`, and the round-trip tests are unchanged. `lib/cli/literate.ts` opts in.

Tests added:
- `lib/backends/agencyGenerator.test.ts` — two cases pinning that `preserveOrder` does not sort and does not hoist mid-file imports.
- `lib/cli/literate.test.ts` — two cases covering the same at the literate level (alphabetical-sort regression + mid-file import regression).

PR #17, opened 2026-04-15
Add regex literals and =~ / !~ match operators

egonSchiele:
## Summary
- Adds regex literal syntax (`/pattern/flags`) with a new `RegexLiteral` AST node and parser
- Adds Ruby-style `=~` and `!~` binary operators for regex matching
- `foo =~ /bar/` compiles to `/bar/.test(foo)`, `foo !~ /bar/` compiles to `!(/bar/.test(foo))`
- Also includes underscore separators in number literals (e.g. `1_000_000`)

## Test plan
- [x] Parser tests for regex literals (patterns, flags, escaped slashes, empty regex)
- [x] Parser tests for `=~` and `!~` as binary operators
- [x] End-to-end Agency test (`tests/agency/regexMatch`) with 3 test cases covering both operators
- [x] All 1846 existing + new tests pass
- [x] Build passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11, opened 2026-04-14
Add OOP classes to Agency language

egonSchiele:
## Summary

- Adds `class` keyword with fields, constructors, methods, single inheritance (`extends`), and method overriding
- Classes compile to TypeScript classes with full runtime machinery (steps, interrupts, async support in methods)
- Auto-generated `toJSON()`/`fromJSON()` for interrupt serialization, with class registry and JSON reviver for round-trip support
- `this` and `super()` work naturally (compile directly to TypeScript equivalents)
- Default constructors generated from field declarations when none provided

## What's included

- **AST types**: `ClassDefinition`, `NewExpression`, `ClassField`, `ClassMethod`, `ClassConstructor`
- **Parsers**: `class` definitions, `new ClassName(...)`, `this`/`super()` (latter two work via existing parsers)
- **Builder**: Full method compilation with `setupFunction`/Runner/steps, `__state` injection on all method calls
- **Preprocessor**: `walkBody` and `walkNodes` traverse class bodies; `this`/`super` skip scope resolution
- **Formatter**: `classDefinition` and `newExpression` in agency generator
- **Runtime**: Class registry on `RuntimeContext` (shared across execution contexts), serialization reviver in `restoreState` and `resumeFromState`
- **Symbol table + ProgramInfo**: `"class"` kind, `classDefinitions` map

## Not yet implemented

- Zod schema generation for classes as tool parameter types
- Tool parameter instance reconstruction (`fromJSON` on LLM results)
- Type checker support for classes

## Test plan

- [x] 3 Agency execution tests passing (basic, inheritance, interrupt)
- [x] 1803 existing unit tests passing (1 pre-existing failure unrelated)
- [ ] Manual: `pnpm run agency test tests/agency/classes/`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #225, opened 2026-05-30
Cross-Thread Context Sharing: std::threads registry + resumable thread blocks

egonSchiele:
## Cross-Thread Context Sharing — `std::threads` registry + resumable `thread {}` blocks

Implements the plan in `docs/superpowers/plans/2026-05-29-thread-registry.md`. The marquee use case is the **categorize-and-route** pattern: a router node sends each user message to a specialized agent node, and the third message that re-visits a category (e.g. "back to the coding task") resumes the *original* thread instead of starting fresh.

### User-visible API

New stdlib module `std::threads`:

```agency
import { listThreads, getThread, currentThreadId } from "std::threads"

listThreads()                       // ThreadInfo[]  (id, label, summary, parentId, messageCount, isActive, ...)
getThread(id, offset?, limit?)      // ThreadMessage[]
currentThreadId()                   // slug-form id of the active thread, or ""
```

`thread {}` gains four optional named args:

```agency
thread(label: "coding task", summarize: true) { ... }      // tag + opt-in eager summary
thread(continue: priorId) { ... }                          // resume a closed thread
thread(session: "coding", label: "coding task") { ... }    // session-keyed auto-resume
```

`continue` and `session` are mutually exclusive (parser-rejected). `session` is sugar over `continue` with a runtime-owned name→id map; first entry creates, subsequent entries resume.

### Architecture

Shipped as **six small extensibility primitives + one stdlib module**, so any user can replicate (or extend) the registry without touching core:

1. `agency.threads.{list, get, current}` — TS namespace mirroring the stdlib surface.
2. `onThreadStart` / `onThreadEnd` lifecycle hooks (slug ids, label, isResumption, message snapshot on End).
3. `thread(label, summarize, continue, session)` named-arg syntax.
4. Cross-node `ThreadStore` persistence — required zero production code; added pinning test (`lib/runtime/__tests__/node.test.ts`).
5. `ThreadStore.resumeExisting(id)` — explicit continuation primitive.
6. `ThreadStore.openSession(name)` — single owner of the "create if absent, resume if present" rule.

The stdlib module is *the same code a user would write themselves* given those primitives. The thread metadata cache lives in `lib/stdlib/threads.ts` because Agency `static const` variables are immutable.

### Implementation notes

- **Subthreads cannot be resumed.** Both `resumeExisting` and `openSession` reject subthreads — a subthread's identity is tied to its parent's context at the time of creation, so resuming one out of context would be misleading.
- **`MessageThread.parentId`** now round-trips through JSON so `ThreadInfo` surfaces parent-child linkage and the resume guards survive interrupt resume.
- **`hasHook(ctx, name)`** fast-path: `Runner.thread` skips the `await invokeCallbacks(...)` microtask boundary when no listeners are registered, preserving debugger checkpoint timing in the common case. Scoped callbacks are intentionally not consulted (false is a fast-path hint only; callers are always safe to call invokeCallbacks anyway).
- **Slug ids** (`t1`, `t2`, ...) wrap the existing per-store counter. LLM-friendly; the "wrong but valid integer" risk is bounded by `listThreads()` returning the slug→summary mapping for disambiguation.

### Out of scope (v1)

- Auto-context injection by label (`thread(label: ..., includePriorSummaries: true)`).
- Cross-run persistence (registry is run-scoped; future work could fold this into the memory layer).

### Tests

- Agency integration: `tests/agency/threads-registry/{basic,label,pagination,eager-vs-lazy,continue,session}.agency`.
- Unit: `ThreadStore.resumeExisting` / `openSession` in `lib/runtime/state/threadStore.test.ts`.
- Cross-node persistence pinning test in `lib/runtime/__tests__/node.test.ts`.
- Hook firing tests in the existing hooks test suite.

### Local validation

- `make fixtures` — regenerated (`tests/typescriptPreprocessor/threads.json` includes the four new optional named-arg fields).

Full unit + agency suites will run in CI per repo policy.

### Docs

- New guide: `docs/site/guide/cross-thread-context.md` (full API + categorize-and-route worked example + memory-scoping-for-routed-agents section + sessions subsection).
- Linked from `docs/site/guide/message-history-and-threads.md` and from the site nav (`docs/site/.vitepress/config.mts`).
- CHANGELOG entry under `## Unreleased`.

### Plan + history

- Plan: `docs/superpowers/plans/2026-05-29-thread-registry.md`
- Six commits, one per task:
  - `766d33c8` Task 1: `agency.threads.*` namespace + `parentId` on MessageThread
  - `f23cd3d3` Task 2: `onThreadStart`/`onThreadEnd` hooks + `ThreadStore.resumeExisting`/`openSession`
  - `f2071aef` Task 3: `thread(label, summarize, continue, session)` parser + codegen
  - `260bd4c7` Tasks 4-5: `stdlib/threads.agency` module + cross-node persistence test
  - `0a515c7b` Tasks 6+7: `thread(continue)` + `thread(session)` integration tests
  - `0d6806a2` Task 9: docs + changelog + lint cleanup + hasHook fast path

### DIFF
--- a/packages/agency-lang/CHANGELOG.md
+++ b/packages/agency-lang/CHANGELOG.md
@@ -1,3 +1,23 @@
+## Unreleased
+
+### Cross-Thread Context Sharing
+
+- New `std::threads` stdlib module: `listThreads()`, `getThread(id, offset, limit)`, `currentThreadId()`. Inspect every thread in the run (active + closed) and paginated-read any thread's messages. The marquee use case is the categorize-and-route pattern (one router, many specialized agent nodes, per-category thread continuity).
+- `agency.threads.{list, get, current}` TS namespace mirrors the stdlib surface for TypeScript helpers and external integrations.
+- New `onThreadStart` / `onThreadEnd` lifecycle hooks, alongside the existing `onNodeStart/End`. Payloads carry slug-form thread ids, `label`, `isResumption`, and the final message snapshot at close.
+- `thread {}` block grows four optional named args: `thread(label, summarize, continue, session) { ... }`.
+  - `label` — template-level intent ("coding task"); surfaces on `ThreadInfo`.
+  - `summarize` — opt-in eager summarization hook payload.
+  - `continue: id` — resume a previously-closed thread; new messages append.
+  - `session: "name"` — runtime-owned name → id map; first entry creates, subsequent entries resume. Sugar over `continue` that erases the "did I save the id?" footgun.
+- `ThreadStore.resumeExisting()` and `ThreadStore.openSession()` runtime primitives; both reject subthreads (a subthread's identity is tied to its parent's context at the time it was created).
+- `MessageThread.parentId` round-trips through JSON, so the registry surfaces parent-child linkage on `ThreadInfo` and the resume guards can recognize subthreads after an interrupt resume.
+- **Cross-node persistence required no production code.** The existing `state.messages` mechanism (`runNode` creates one `ThreadStore` per agent run; `setupNode` reuses it on every node hop) already does the right thing. We added a unit test (`lib/runtime/__tests__/node.test.ts`) pinning the contract so future refactors of `setupNode` can't silently break the registry.
+
+See [`docs/site/guide/cross-thread-context.md`](docs/site/guide/cross-thread-context.md) for the full guide and the categorize-and-route worked example.
+
+---
+
 ## May 28 2026 — v0.3
 
 ### Language
--- a/packages/agency-lang/CHANGELOG.md
+++ b/packages/agency-lang/CHANGELOG.md
@@ -1,23 +1,3 @@
-## Unreleased
-
-### Cross-Thread Context Sharing
-
-- New `std::threads` stdlib module: `listThreads()`, `getThread(id, offset, limit)`, `currentThreadId()`. Inspect every thread in the run (active + closed) and paginated-read any thread's messages. The marquee use case is the categorize-and-route pattern (one router, many specialized agent nodes, per-category thread continuity).
-- `agency.threads.{list, get, current}` TS namespace mirrors the stdlib surface for TypeScript helpers and external integrations.
-- New `onThreadStart` / `onThreadEnd` lifecycle hooks, alongside the existing `onNodeStart/End`. Payloads carry slug-form thread ids, `label`, `isResumption`, and the final message snapshot at close.
-- `thread {}` block grows four optional named args: `thread(label, summarize, continue, session) { ... }`.
-  - `label` — template-level intent ("coding task"); surfaces on `ThreadInfo`.
-  - `summarize` — opt-in eager summarization hook payload.
-  - `continue: id` — resume a previously-closed thread; new messages append.
-  - `session: "name"` — runtime-owned name → id map; first entry creates, subsequent entries resume. Sugar over `continue` that erases the "did I save the id?" footgun.
-- `ThreadStore.resumeExisting()` and `ThreadStore.openSession()` runtime primitives; both reject subthreads (a subthread's identity is tied to its parent's context at the time it was created).
-- `MessageThread.parentId` round-trips through JSON, so the registry surfaces parent-child linkage on `ThreadInfo` and the resume guards can recognize subthreads after an interrupt resume.
-- **Cross-node persistence required no production code.** The existing `state.messages` mechanism (`runNode` creates one `ThreadStore` per agent run; `setupNode` reuses it on every node hop) already does the right thing. We added a unit test (`lib/runtime/__tests__/node.test.ts`) pinning the contract so future refactors of `setupNode` can't silently break the registry.
-
-See [`docs/site/guide/cross-thread-context.md`](docs/site/guide/cross-thread-context.md) for the full guide and the categorize-and-route worked example.
-
----
-
 ## May 28 2026 — v0.3
 
 ### Language
--- a/packages/agency-lang/docs/site/.vitepress/config.mts
+++ b/packages/agency-lang/docs/site/.vitepress/config.mts
@@ -47,10 +47,6 @@ export default defineConfig({
               text: "Message History and Threads",
               link: "/guide/message-history-and-threads",
             },
-            {
-              text: "Cross-Thread Context Sharing",
-              link: "/guide/cross-thread-context",
-            },
             { text: "Error Handling", link: "/guide/error-handling" },
             { text: "Blocks", link: "/guide/blocks" },
             { text: "Execution Model", link: "/guide/execution-model" },
@@ -67,6 +63,10 @@ export default defineConfig({
             { text: "Type Validation", link: "/guide/type-validation" },
             { text: "Checkpointing", link: "/guide/checkpointing" },
             { text: "Memory", link: "/guide/memory" },
+            {
+              text: "Cross-Thread Context Sharing",
+              link: "/guide/cross-thread-context",
+            },
             { text: "MCP", link: "/guide/mcp" },
           ],
         },
@@ -116,10 +116,6 @@ export default defineConfig({
             {
               text: "Message History and Threads",
               link: "/guide/message-history-and-threads",
-            },
-            {
-              text: "Cross-Thread Context Sharing",
-              link: "/guide/cross-thread-context",
             },
             { text: "Error Handling", link: "/guide/error-handling" },
             { text: "Debugger", link: "/guide/debugger" },
--- a/packages/agency-lang/docs/site/.vitepress/config.mts
+++ b/packages/agency-lang/docs/site/.vitepress/config.mts
@@ -47,6 +47,10 @@ export default defineConfig({
               text: "Message History and Threads",
               link: "/guide/message-history-and-threads",
             },
+            {
+              text: "Cross-Thread Context Sharing",
+              link: "/guide/cross-thread-context",
+            },
             { text: "Error Handling", link: "/guide/error-handling" },
             { text: "Blocks", link: "/guide/blocks" },
             { text: "Execution Model", link: "/guide/execution-model" },
@@ -112,6 +116,10 @@ export default defineConfig({
             {
               text: "Message History and Threads",
               link: "/guide/message-history-and-threads",
+            },
+            {
+              text: "Cross-Thread Context Sharing",
+              link: "/guide/cross-thread-context",
             },
             { text: "Error Handling", link: "/guide/error-handling" },
             { text: "Debugger", link: "/guide/debugger" },
--- a/packages/agency-lang/docs/site/guide/cross-thread-context.md
+++ b/packages/agency-lang/docs/site/guide/cross-thread-context.md
@@ -1,0 +1,166 @@
+# Cross-Thread Context Sharing
+
+The cross-thread context-sharing feature lets one agent inspect, read, and resume **sibling `thread {}` blocks** in the same run. The marquee use case is the *categorize-and-route* pattern: a top-level router sends each user message to a specialized agent node, each agent owns its own thread for isolation, and when the user returns to an earlier topic the router resumes the *original* thread instead of starting fresh.
+
+```agency
+import { listThreads, getThread, currentThreadId } from "std::threads"
+```
+
+Three exports cover the user-facing API:
+
+- `listThreads(): ThreadInfo[]` — every thread in the run (active + closed) with optional `label` and lazily-generated `summary`.
+- `getThread(id, offset, limit): ThreadMessage[]` — paginated read of any thread's messages.
+- `currentThreadId(): string` — slug of the active thread, useful when you want to capture an id at the moment it was active so you can `thread(continue: id)` later.
+
+And the `thread {}` block grows four optional named arguments:
+
+```agency
+thread(label: "coding task", summarize: true) { /* ... */ }
+thread(continue: priorId) { /* ... */ }
+thread(session: "coding", label: "coding") { /* ... */ }
+```
+
+`continue` and `session` are mutually exclusive — pick one. `session` is sugar over `continue` plus a runtime-owned name → id map.
+
+---
+
+## Why `label` *and* `summary`?
+
+A `thread {}` block inside a loop runs many times with very different content. The user-supplied `label` captures the *template intent* ("this is a coding task"). The auto-generated `summary` captures *what actually happened in this instance* ("refactored auth, added JWT validation"). `listThreads()` returns both so an LLM searching for "the auth one" can match on summary while one looking for "any coding task" can match on label.
+
+## Slug ids (`t1`, `t2`, ...) vs. nanoids
+
+`ThreadStore` uses a per-store integer counter internally; the registry exposes it as a slug (`t${counter}`). LLMs handle short ordinal strings far better than random nanoids, and the wrong-but-valid-integer risk is bounded by `listThreads()` returning the summaries — the LLM picks by reading the list, not by guessing an integer.
+
+---
+
+## Resuming a thread (`continue`)
+
+`thread(continue: id)` re-enters a previously-closed thread. New messages append to its existing history rather than starting fresh:
+
+```agency
+import { getThread, currentThreadId } from "std::threads"
+import { userMessage } from "std::thread"
+
+node main() {
+  let codingId: string = ""
+  thread(label: "coding") {
+    userMessage("first coding turn")
+    codingId = currentThreadId()
+  }
+  // ... work on something else ...
+  thread(continue: codingId) {
+    userMessage("back to coding")
+  }
+  // The coding thread now holds BOTH messages, in order.
+}
+```
+
+**v1 limitation: subthreads cannot be resumed.** A subthread inherits its parent's message history at the moment it was created; resuming one outside that context would surface confusing ordering. If you need to continue work inside a subthread, resume the parent thread and open a fresh `subthread {}` block.
+
+---
+
+## Routing with sessions
+
+`thread(session: "name")` is the bread-and-butter primitive for routers. The runtime owns a `session-name → thread-id` map; first entry creates a thread, subsequent entries auto-resume it — no id bookkeeping at the call site:
+
+```agency
+import { listThreads } from "std::threads"
+import { userMessage } from "std::thread"
+
+node main() {
+  // First "coding" entry creates a new thread.
+  thread(session: "coding", label: "coding") {
+    userMessage("first coding")
+  }
+  // Distinct session — new thread.
+  thread(session: "weather", label: "weather") {
+    userMessage("weather request")
+  }
+  // Second "coding" entry RESUMES the original — message
+  // history persists across the weather hop.
+  thread(session: "coding", label: "coding") {
+    userMessage("back to coding")
+  }
+}
+```
+
+Sessions are always **top-level threads** — they cannot map to subthreads. (If you want a subthread inside a session, open the session normally and put a `subthread {}` block inside.)
+
+---
+
+## Memory scoping for routed agents
+
+When a router fans messages out to per-category agents, each agent should have its own *memory* scope so the coding agent's facts don't bleed into the weather agent. `setMemoryId(category)` from `std::memory` is already per-branch — call it at the top of each routed node:
+
+```agency
+import { setMemoryId } from "std::memory"
+
+node coding() {
+  setMemoryId("coding")
+  // ... coding-specific recall() and remember() calls ...
+}
+
+node weather() {
+  setMemoryId("weather")
+  // ... weather-specific recall() and remember() calls ...
+}
+```
+
+If you have facts that everyone should read — a workspace overview, the user's name — keep one well-known id and switch to it for cross-cutting writes:
+
+```agency
+node coding() {
+  setMemoryId("workspace")
+  remember("the user prefers Rust")
+  setMemoryId("coding")
+  // ... back to per-category scope ...
+}
+```
+
+This pairs with `thread(session: category)` to give each routed agent its own thread *and* its own memory scope:
+
+```agency
+import { listThreads } from "std::threads"
+import { setMemoryId } from "std::memory"
+
+node routedAgent(category: string, message: string) {
+  setMemoryId(category)
+  thread(session: category, label: category) {
+    // ... category-specific work ...
+  }
+}
+```
+
+---
+
+## Cross-node persistence
+
+The thread registry persists across node transitions automatically. Internally a single `ThreadStore` is created once per `runNode` call and threaded through `state.messages` to every step — so threads created in node `a` remain queryable from node `b`. Subthreads are normal entries in the same registry with `parentId` set.
+
+On interrupt resume the store is rebuilt from `stack.threads` (checkpoint serialization). The registry survives the interrupt and in-flight `thread {}` blocks resume mid-flight via the existing substep mechanism. See `docs/dev/threads.md` for the underlying mechanism.
+
+---
+
+## How this is built
+
+The whole feature is six small primitives plus a stdlib module:
+
+1. `agency.threads.{list, get, current}` — TS namespace, parallel to `agency.memory.*`.
+2. `onThreadStart` / `onThreadEnd` lifecycle hooks, in the same `CallbackMap` as `onNodeStart`/`onNodeEnd`.
+3. `thread(label, summarize)` named args.
+4. Cross-node persistence (no new code — `ThreadStore` already flows through `state.messages` on every node transition).
+5. `thread(continue: id)` — `ThreadStore.resumeExisting()`, rejects subthreads.
+6. `thread(session: "name")` — `ThreadStore.openSession()`, sugar over `continue`.
+
+The `stdlib/threads.agency` module is *the same code a user could write* on top of those primitives. If you want to ship a "tag threads by topic" feature, or a "diff two threads" inspector, or any other variant, the same surface is open to you — `agency.threads.*` plus the lifecycle hooks plus `thread(label / summarize / continue / session)` cover the whole feature set.
+
+---
+
+## Out of scope for v1
+
+- **Auto-context injection by label.** Users can replicate it in three lines with `listThreads().filter(...)`.
+- **Cross-run persistence.** The registry is run-scoped — restarting the agent starts fresh.
+- **Filtering / search in `listThreads`.** Returns everything; the LLM filters.
+- **Summary regeneration.** Once cached, a summary is final. Eager summarize on close (`thread(summarize: true)`) is not yet wired through the v1 stdlib hook — it remains on the wire for users who register their own `callback("onThreadEnd")`.
+- **A router primitive.** User-space already expresses categorize-and-route in three lines (see `tests/agency/categorize.agency`); adding `route { ... }` would be redundant complexity.
--- a/packages/agency-lang/docs/site/guide/cross-thread-context.md
+++ b/packages/agency-lang/docs/site/guide/cross-thread-context.md
@@ -18,10 +18,11 @@ The `std::threads` module gives you both:
 import { listThreads, getThread, currentThreadId } from "std::threads"
 ```
 
-And the `thread {}` block grows four optional named arguments:
+And the `thread {}` block grows five optional named arguments:
 
 ```ts
 thread(label: "coding") { /* ... */ }                        // tag it
+thread(summarize: true) { /* ... */ }                         // pre-mark for eager summary
 thread(continue: priorId) { /* ... */ }                       // resume by id
 thread(session: "coding") { /* ... */ }                       // resume by name (sugar)
 thread(label: "coding", hidden: true) { /* ... */ }           // exclude from listThreads()
--- a/packages/agency-lang/docs/site/guide/cross-thread-context.md
+++ b/packages/agency-lang/docs/site/guide/cross-thread-context.md
@@ -1,44 +1,87 @@
-# Cross-Thread Context Sharing
+---
+title: Cross-thread context sharing
+description: Inspect, read, and resume sibling `thread {}` blocks in the same run. Covers `listThreads()`, `getThread()`, and the `thread(label, summarize, continue, session)` named arguments, plus the categorize-and-route worked example.
+---
 
-The cross-thread context-sharing feature lets one agent inspect, read, and resume **sibling `thread {}` blocks** in the same run. The marquee use case is the *categorize-and-route* pattern: a top-level router sends each user message to a specialized agent node, each agent owns its own thread for isolation, and when the user returns to an earlier topic the router resumes the *original* thread instead of starting fresh.
+# Cross-thread context sharing
 
-```agency
+Here's the problem. You've built a router: a top-level node reads each user message, picks a category, and dispatches to one of several specialized agents. Each agent runs in its own `thread {}` block so its context stays clean. Great — until the user comes back to a topic they were discussing five minutes ago, and the agent has no memory of it. Each `thread {}` block starts fresh, so "back to coding" walks into a `coding` thread with no prior turns.
+
+You need two things to fix this:
+
+1. A way to *see* the other threads in the run.
+2. A way to *resume* one when the user circles back.
+
+The `std::threads` module gives you both:
+
+```ts
 import { listThreads, getThread, currentThreadId } from "std::threads"
 ```
 
-Three exports cover the user-facing API:
-
-- `listThreads(): ThreadInfo[]` — every thread in the run (active + closed) with optional `label` and lazily-generated `summary`.
-- `getThread(id, offset, limit): ThreadMessage[]` — paginated read of any thread's messages.
-- `currentThreadId(): string` — slug of the active thread, useful when you want to capture an id at the moment it was active so you can `thread(continue: id)` later.
-
 And the `thread {}` block grows four optional named arguments:
 
-```agency
-thread(label: "coding task", summarize: true) { /* ... */ }
-thread(continue: priorId) { /* ... */ }
-thread(session: "coding", label: "coding") { /* ... */ }
+```ts
+thread(label: "coding") { /* ... */ }                        // tag it
+thread(continue: priorId) { /* ... */ }                       // resume by id
+thread(session: "coding") { /* ... */ }                       // resume by name (sugar)
+thread(label: "coding", hidden: true) { /* ... */ }           // exclude from listThreads()
 ```
 
-`continue` and `session` are mutually exclusive — pick one. `session` is sugar over `continue` plus a runtime-owned name → id map.
+`continue` and `session` are mutually exclusive — pick one. We'll cover each below.
 
----
+## Listing threads
 
-## Why `label` *and* `summary`?
+`listThreads()` returns every thread in the run, including the active one. It returns a `Result` (see [error handling](./error-handling)) — success holds a `ThreadInfo[]`, failure holds an error message (e.g. when called from outside a node body).
 
-A `thread {}` block inside a loop runs many times with very different content. The user-supplied `label` captures the *template intent* ("this is a coding task"). The auto-generated `summary` captures *what actually happened in this instance* ("refactored auth, added JWT validation"). `listThreads()` returns both so an LLM searching for "the auth one" can match on summary while one looking for "any coding task" can match on label.
+```ts
+const all = listThreads()
+if (isSuccess(all)) {
+  for (t in all.value) {
+    print("[" + t.id + "] " + t.label + " — " + t.summary)
+  }
+}
+```
 
-## Slug ids (`t1`, `t2`, ...) vs. nanoids
+Each `ThreadInfo` has:
 
-`ThreadStore` uses a per-store integer counter internally; the registry exposes it as a slug (`t${counter}`). LLMs handle short ordinal strings far better than random nanoids, and the wrong-but-valid-integer risk is bounded by `listThreads()` returning the summaries — the LLM picks by reading the list, not by guessing an integer.
+```ts
+type ThreadInfo = {
+  id: string             // slug form: "t0", "t1", ...
+  label?: string         // from thread(label: "...") { }
+  summary?: string       // LLM-generated, lazily on first listThreads() call
+  parentId?: string      // set when this is a subthread
+  threadType: string     // "thread" | "subthread"
+  messageCount: number
+  isActive: boolean
+}
+```
 
----
+A few things worth knowing:
 
-## Resuming a thread (`continue`)
+- **Slug ids (`t1`, `t2`, ...) instead of nanoids.** LLMs handle short ordinal strings far better than random ids. When the LLM picks a thread to resume, it picks by reading `listThreads()` and matching on the summary — not by guessing an integer.
+- **Summaries are lazy.** The first time `listThreads()` runs against a closed thread without a cached summary, it makes one LLM call to summarize and stashes the result on the thread itself. Subsequent calls read the cache. Active threads are skipped so we don't summarize a half-finished conversation.
+- **`label` vs `summary`.** A `thread {}` block inside a loop runs many times with very different content. `label` captures the *template intent* ("coding task"). `summary` captures *what actually happened this time* ("refactored auth, added JWT validation"). An LLM searching for "the auth one" can match on summary; one searching for "any coding task" can match on label.
 
-`thread(continue: id)` re-enters a previously-closed thread. New messages append to its existing history rather than starting fresh:
+## Reading a thread's messages
 
-```agency
+`getThread(id, offset, limit)` returns a slice of a thread's history. Also a `Result`:
+
+```ts
+const msgs = getThread("t1", 0, 20)
+if (isSuccess(msgs)) {
+  for (m in msgs.value) {
+    print("[" + m.role + "] " + m.content)
+  }
+}
+```
+
+`offset` is 0-indexed; `limit` defaults to 50. Pass larger limits for a full-thread read.
+
+## Resuming a thread by id
+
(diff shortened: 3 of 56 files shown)


packages/agency-lang/docs/site/.vitepress/config.mts
```diff
@@ -113,6 +117,10 @@ export default defineConfig({
               text: "Message History and Threads",
               link: "/guide/message-history-and-threads",
             },
+            {
+              text: "Cross-Thread Context Sharing",
```
egonSchiele: Put this in the advanced section

packages/agency-lang/docs/site/.vitepress/config.mts
```diff
@@ -47,6 +47,10 @@ export default defineConfig({
...
```
egonSchiele: Why two links for the same thing?

packages/agency-lang/docs/site/guide/cross-thread-context.md
```diff
@@ -0,0 +1,166 @@
+# Cross-Thread Context Sharing
```
egonSchiele: Try to imitate the more informal style of other parts of the guide

packages/agency-lang/docs/site/guide/message-history-and-threads.md
```diff
@@ -81,4 +81,8 @@ They do **not** work in these "bootstrap" scopes:
 
 If you call a thread builtin from one of these scopes — `systemMessage`, `userMessage`, `llm`, `thread { ... }`, `subthread { ... }` — you'll get a runtime error like *"Message threads are not available in this scope."* with a pointer to this guide. Move the code inside a `node` or `def` body to fix it.
 
-`onAgentEnd` is the exception: it fires *after* the run completes, so it has access to the final conversation. You can inspect or log the thread state from there.
\ No newline at end of file
+`onAgentEnd` is the exception: it fires *after* the run completes, so it has access to the final conversation. You can inspect or log the thread state from there.
+
+## See also
+
+- [Cross-Thread Context Sharing](./cross-thread-context.md) — `listThreads()`, `getThread()`, and the `thread(label, summarize, continue, session)` named arguments for inspecting and resuming sibling threads. The marquee use case is the categorize-and-route pattern (one router, many specialized agents, per-category thread continuity).
```
egonSchiele: wrong url type, since .md files get compiled

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2791,7 +2791,33 @@ export class TypeScriptBuilder {
       );
     }
 
-    return ts.runnerThread({ id, method, body: bodyNodes });
+    // Optional named args: thread(label, summarize, continue, session).
+    // Each is an Expression in the AST — process to TsNode so codegen
+    // can splice it into the opts arg. Codegen-time mutual exclusion
+    // check defends parse-time guard.
+    if (node.continueExpr && node.sessionExpr) {
```
egonSchiele, line 2802: This feels like the wrong way to do it. Could we implement the thread block as an agency function that takes a block instead (https://agency-lang.com/guide/blocks.html)? Then we could get all of this functionality for free, they would just be other parameters to the function. How big of a change would that be? Is it useful to keep the `thread` block as its own separate type in the agency AST?

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3089,12 +3089,77 @@ export const bodyParser = (input: string): ParserResult<AgencyNode[]> => {
...
+  );
+  const r = inner(input);
+  if (!r.success) return r as ParserResult<ThreadNamedArgs>;
+  const args: NamedArgument[] = r.result.arguments;
+  const allowed = ["label", "summarize", "continue", "session"];
+  const seen: Record<string, Expression> = {};
+  for (const arg of args) {
+    if (!allowed.includes(arg.name)) {
+      return failure(
+        `Unknown thread argument: ${arg.name}. Allowed: label, summarize, continue, session`,
```
egonSchiele: Some of these failures feel like they should actually be type checker failures, not parser failures. Again, fine for v1, but this is another thing we would get for free if the thread block was a function defined in agency instead.

packages/agency-lang/lib/runtime/agency.ts
```diff
@@ -132,6 +132,97 @@ const threadWith = async <T>(
...
+  id: string;
+  parentId: string | null;
+  threadType: "thread" | "subthread";
+  messageCount: number;
+  isActive: boolean;
+  messages: ThreadMessageTS[];
+};
+
+/** Plain record describing a single message inside a thread. */
+export type ThreadMessageTS = {
```
egonSchiele: We already have a Message type used in llmclient, runPrompt, etc. Any reason not to reuse that type?

packages/agency-lang/lib/runtime/agency.ts
```diff
@@ -132,6 +132,97 @@ const threadWith = async <T>(
...
+/** Read a slice of a thread's messages. Returns `[]` for unknown ids
+ *  (silent — callers can probe). `offset` and `limit` default to 0 and
+ *  50, matching the Agency-side `getThread` signature. */
+const threadsGet = (
+  id: string,
+  offset = 0,
+  limit = 50,
+): ThreadMessageTS[] => {
+  const store = threadStoreMaybe();
+  if (!store) return [];
```
egonSchiele: Should we throw an error if no store was found?

packages/agency-lang/lib/runtime/agency.ts
```diff
@@ -132,6 +132,97 @@ const threadWith = async <T>(
...
+  const rawId = fromSlug(id);
+  const thread = store.threads[rawId];
+  if (!thread) return [];
```
egonSchiele, line 230: Same, throw error?

egonSchiele: Both errors should get converted to Failures in agency https://agency-lang.com/guide/error-handling.html

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -466,8 +486,21 @@ export class Runner {
   async thread(
     id: number,
     method: "create" | "createSubthread",
-    callback: (runner: Runner) => Promise<void>,
+    optsOrCallback:
+      | ThreadStepOpts
+      | ((runner: Runner) => Promise<void>),
+    maybeCallback?: (runner: Runner) => Promise<void>,
   ): Promise<void> {
+    // Two-form signature so legacy call sites
```
egonSchiele: no need to have backwards compatibility, make a clean, breaking change

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -491,27 +524,121 @@ export class Runner {
...
-      threads.popActive();
+      // Snapshot messages BEFORE popping the active stack so the
+      // onThreadEnd payload sees the just-closed thread's final
+      // message list. Use the active id (which is `tid`) to look it
+      // up regardless of double-push avoidance.
+      const closingThread = threads.get(tid);
+      const messagesSnapshot = closingThread
+        ? closingThread.messages.map((m) => m.toJSON())
+        : [];
+      if (!alreadyActive) {
```
egonSchiele: both branches of if statement are the same

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -491,27 +524,121 @@ export class Runner {
...
+          });
+        }
+      } catch (e) {
+        // Swallow hook errors in finally to avoid masking the
+        // primary exception. `fireWithGuard` inside invokeCallbacks
+        // already logs JS errors; this catch is belt-and-braces for
+        // unexpected throws from the dispatcher itself.
+        if (e instanceof RestoreSignal) throw e;
+        // eslint-disable-next-line no-console
+        console.error("[agency] onThreadEnd dispatcher error:", e);
```
egonSchiele: maybe log to statelog too?

packages/agency-lang/stdlib/threads.agency
```diff
@@ -0,0 +1,152 @@
+/**
```
egonSchiele: Use `/** @module */` for module-level docstrings

packages/agency-lang/stdlib/threads.agency
```diff
@@ -0,0 +1,152 @@
...
+  id: string,
+  label?: string,
+  summary?: string,
+  parentId?: string,
+  threadType: string,
+  messageCount: number,
+  isActive: boolean,
+}
+
+// One-shot LLM summarization used by the lazy + eager paths. Wrapped
```
egonSchiele: Use doc comment so this info gets put into docs

packages/agency-lang/stdlib/threads.agency
```diff
@@ -0,0 +1,152 @@
...
+// in a `thread {}` block so the summarizer prompt runs on an isolated
+// message history and doesn't pollute the agent's main conversation.
+def summarize(messages: ThreadMessage[]): string {
```
egonSchiele, line 58: Add doc string, which is used as a tool description if this func is given as a tool to an LLM

packages/agency-lang/stdlib/threads.agency
```diff
@@ -0,0 +1,152 @@
...
+  let transcript = ""
+  for (m in messages) {
+    transcript = transcript + "[" + m.role + "] " + m.content + "\n"
+  }
+  let result = ""
+  thread {
+    result = llm("Summarize this conversation in 1-2 sentences:\n" + transcript)
```
egonSchiele: The LLM may return a response that includes extraneous text like "Here is the summary you asked for." I would recommend setting a structured output format to prevent this (https://agency-lang.com/guide/llm.html):

```
type Result = {
  summary: string
}
result: Result = llm("Summarize this conversation in 1-2 sentences:\n" + transcript)
```

packages/agency-lang/stdlib/threads.agency
```diff
@@ -0,0 +1,152 @@
...
+  }
+  if (messages == null) {
+    return null
+  }
+  // Narrow: messages is non-null at this point. Re-bind via a
+  // `const` of the non-null variant so the typechecker's flow
+  // analysis sees a `ThreadMessage[]` for the summarize() call.
+  const concrete: ThreadMessage[] = messages
+  const s = summarize(concrete)
+  _remember(id, { summary: s })
```
egonSchiele: name is the same as one of the funcs for the memory layer, which makes it confusing as the context is similar

packages/agency-lang/stdlib/threads.agency
```diff
@@ -0,0 +1,152 @@
...
+    result = llm("Summarize this conversation in 1-2 sentences:\n" + transcript)
+  }
+  return result
+}
+
+// Single owner of the "return cached summary, else LLM-summarize and
+// cache" rule. Active threads pass `null` so the LLM call is skipped
+// while the conversation is still in flight.
+def summaryFor(id: string, messages: (ThreadMessage[] | null) = null): (string | null) {
+  const c = _getCached(id)
```
egonSchiele: what is the cache? Does it get shared across all calls to the agent, because we don't want that: https://agency-lang.com/guide/execution-model.html

Create it in agency if you want isolated state per agent call

packages/agency-lang/stdlib/threads.agency
```diff
@@ -0,0 +1,152 @@
...
+  // analysis sees a `ThreadMessage[]` for the summarize() call.
+  const concrete: ThreadMessage[] = messages
+  const s = summarize(concrete)
+  _remember(id, { summary: s })
+  return s
+}
+
+// Lifecycle hook is registered TS-side in lib/stdlib/threads.ts via
+// `registerGlobalHook("onThreadEnd", ...)`. A top-level Agency-side
+// `callback(...)` only fires when its module is the entry point of
```
egonSchiele: Are you sure about this? That is unexpected, I'd expect it to fire always if its part of the same agent.

packages/agency-lang/CHANGELOG.md
```diff
@@ -1,3 +1,23 @@
+## Unreleased
+
+### Cross-Thread Context Sharing
```
egonSchiele: don't update changelog

packages/agency-lang/lib/runtime/hooks.ts
```diff
@@ -133,6 +146,24 @@ export function registerGlobalHook<K extends keyof CallbackMap>(
...
+export function hasHook<K extends keyof CallbackMap>(
+  ctx: RuntimeContext<any>,
+  name: K,
+): boolean {
+  if ((_globalHooks[name] ?? []).length > 0) return true;
+  if (ctx.callbacks && ctx.callbacks[name]) return true;
+  const topLevel = ctx.topLevelCallbacks ?? [];
+  for (const cb of topLevel) if (cb.name === name) return true;
+  return false;
+}
```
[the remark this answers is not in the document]
egonSchiele: Dropped `hasHook` entirely in 9dcc7389. The original justification (extra microtask boundary perturbing debugger checkpoint timing) didn't hold up: re-ran the four substep / debugger tests that wrap `thread {}` blocks (`checkpoint-in-thread`, `interrupt-in-thread`, `interrupt-in-nested-thread`, `interrupt-deep-nesting`) and all four pass exact-match without `hasHook`. `Runner.thread` now always awaits `invokeCallbacks` for both `onThreadStart` and `onThreadEnd`.

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -12,6 +14,24 @@ import { StateStack } from "./state/stateStack.js";
...
+  continueId?: string;
+  /** Session name; runtime maps it to a thread id via openSession. */
+  session?: string;
+};
+
+/** Strip the leading `t` from a public thread slug to recover the
+ *  internal counter-string id. Defensive: accepts a raw id unchanged. */
+function stripSlug(slug: string): string {
+  return slug.startsWith("t") ? slug.slice(1) : slug;
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 9dcc7389: `stripSlug` is now `/^t\d+$/.test(slug) ? slug.slice(1) : slug` so it only strips when the `t` is followed by digits.

packages/agency-lang/stdlib/threads.agency
```diff
@@ -0,0 +1,152 @@
...
+def summarize(messages: ThreadMessage[]): string {
+  let transcript = ""
+  for (m in messages) {
+    transcript = transcript + "[" + m.role + "] " + m.content + "\n"
+  }
+  let result = ""
+  thread {
+    result = llm("Summarize this conversation in 1-2 sentences:\n" + transcript)
+  }
+  return result
```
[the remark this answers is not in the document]
egonSchiele: Added a `hidden: bool` named arg to `thread {}` / `subthread {}` in 9dcc7389 and used it inside `summarize()` (commit 6620bc9c). Hidden threads are filtered out of `agency.threads.list()` / `listThreads()`. Plumbed through parser → AST → IR → codegen → runtime → `MessageThread` (with JSON round-trip). New integration test at `tests/agency/threads-registry/hidden-arg.agency` exercises the filter.

packages/agency-lang/stdlib/threads.agency
```diff
@@ -0,0 +1,152 @@
...
+  // Inspect the run's other threads:
+  const info = listThreads()
+
+  // Read a slice of a prior thread's messages:
+  const lines = getThread("t1", 0, 20)
+  ```
+
+`label` is template-level (set via `thread(label: "...") { ... }`);
+`summary` is per-instance and generated lazily on first `listThreads()`
+call unless the user opts into `thread(summarize: true) { ... }`.
```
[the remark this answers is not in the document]
egonSchiele: The module header was updated in 6620bc9c to match actual v1 behavior: lazy only. Eager-summarize wiring (firing `summarize()` from the `onThreadEnd` hook) needs a clean way to invoke an Agency function from TS-side hook code; deferred to a follow-up so the parser/IR/codegen retain the `summarize: true` arg (callbacks can still observe `evt.eagerSummarize`) without shipping a half-wired path. The `@module` docstring and the guide's "Out of scope" section both call this out now.

packages/agency-lang/stdlib/threads.agency
```diff
@@ -0,0 +1,152 @@
...
+  Return every thread in the current run, including the active one.
+
+  Lazy summary generation: a thread that doesn't yet have a cached
+  summary triggers exactly one LLM round-trip the first time
+  `listThreads()` is called on it. Active threads are skipped so the
+  in-flight conversation is not summarized mid-stream.
+
+  Eager summarization (one summary call per `thread {}` close instead
+  of all-at-once on first `listThreads()`) is opt-in:
+  `thread(summarize: true) { ... }`.
```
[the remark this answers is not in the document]
egonSchiele: Docstring updated in 6620bc9c to match the actual lazy-only behavior (see also #1).

packages/agency-lang/tests/agency/threads-registry/basic.test.json
```diff
@@ -0,0 +1,15 @@
+{
+  "tests": [
+    {
+      "nodeName": "main",
+      "input": "",
+      "expectedOutput": "[\"first\",\"second\"]",
+      "evaluationCriteria": [{ "type": "exact" }],
+      "description": "stdlib/threads listThreads() surfaces labels from thread(label) blocks; eager summaries are cached at close.",
```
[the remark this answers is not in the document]
egonSchiele: Updated in 6620bc9c to describe the lazy-population behavior actually shipping.

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -491,27 +524,121 @@ export class Runner {
...
+      if (opts.continueId !== undefined) {
+        const rawId = stripSlug(opts.continueId);
+        threads.resumeExisting(rawId);
+        tid = rawId;
+        isResumption = true;
+      } else if (opts.session !== undefined) {
+        const { id: openedId, existed } = threads.openSession(opts.session);
+        tid = openedId;
+        isResumption = existed;
+      } else {
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 9dcc7389: `subthread(continue: ...)` and `subthread(session: ...)` are now rejected at codegen time in `processMessageThread` with a clear error message. Tests added under `tests/agency/threads-registry/`.

packages/agency-lang/docs/site/.vitepress/config.mts
egonSchiele: Done in 6620bc9c: the Cross-Thread Context Sharing entry now lives only in the "Advanced" sidebar section (duplicate top-level nav entry removed — see #22).

packages/agency-lang/docs/site/.vitepress/config.mts
egonSchiele: Removed the duplicate in 6620bc9c; the single remaining entry sits in the "Advanced" sidebar section.

packages/agency-lang/docs/site/guide/cross-thread-context.md
egonSchiele: Rewrote in 6620bc9c. Read `message-history-and-threads.md` and `error-handling.md` first to absorb the tone; new version leads with the problem, drops the bullet-heavy Architecture section, and keeps the categorize-and-route worked example.

packages/agency-lang/docs/site/guide/message-history-and-threads.md
egonSchiele: Fixed in 6620bc9c: `./cross-thread-context.md` → `./cross-thread-context`.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Agreed — captured as a follow-up spec at `docs/superpowers/specs/2026-05-30-thread-as-agency-function.md` (commit 3e5c1aec). The spec covers the parser/IR/codegen fan-out, the `__internal_thread(opts, block)` proposal, migration path, tradeoffs, and rough effort estimate (3–4 dev-days, sliceable into 3 PRs). Also dropped a TODO comment above `processMessageThread` in `typescriptBuilder.ts` pointing at the spec. Not part of this PR.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Same spec covers this — under the proposed `__internal_thread(opts, block)` refactor, the "unknown thread argument" parser error becomes a regular type error on the `ThreadOpts` type. See `docs/superpowers/specs/2026-05-30-thread-as-agency-function.md` (commit 3e5c1aec). Following up post-merge.

packages/agency-lang/lib/runtime/agency.ts
egonSchiele: Done in 6620bc9c: `ThreadInfoTS.messages` is now `smoltalk.MessageJSON[]` (was the bespoke `ThreadMessageTS[]`). `messageToThreadMessage` collapsed to `m.toJSON()`. The Agency-visible `ThreadMessage = { role, content: string }` type stayed put for the public API (smoltalk's `content` can be structured); the boundary coerces non-string content at `_getThread` and `_listThreadsRaw`.

packages/agency-lang/lib/runtime/agency.ts
egonSchiele: Throws now (9dcc7389). The Agency-side `stdlib/threads.agency::listThreads()` wraps the throw in a `try` and returns a `Result` per the error-handling guide (6620bc9c). See also #17/#18.

packages/agency-lang/lib/runtime/agency.ts
egonSchiele: Done in 9dcc7389 (same as #16): `threadsGet` throws on missing store; the Agency wrapper returns a `Result`.

egonSchiele: Done in 6620bc9c. `stdlib/threads.agency`'s `listThreads()` and `getThread()` now return `Result` per https://agency-lang.com/guide/error-handling.html. Followed the wrap pattern from `stdlib/memory.agency::memory`. Updated the four integration tests under `tests/agency/threads-registry/` to unwrap via `if (isSuccess(...))`.

packages/agency-lang/lib/runtime/runner.ts
egonSchiele: Done in 6620bc9c: `Runner.thread` is now strictly `(id, method, opts, callback)`. Removed the dual-signature `optsOrCallback` branch. IR emitter already always emits an opts arg, so no codegen change needed; updated `runner.test.ts` for the new signature.

packages/agency-lang/lib/runtime/runner.ts
egonSchiele: Collapsed in 9dcc7389 to a single unconditional `threads.popActive();`. The `alreadyActive` distinction was meaningless — pop runs the same way regardless of whether we pushed or whether `resumeExisting`/`openSession` pushed for us.

packages/agency-lang/lib/runtime/runner.ts
egonSchiele: Done in 6620bc9c. Added a `threadEndHookError({ threadId, error })` method to `lib/statelogClient.ts` (mirroring `threadResumed`) and `Runner.thread`'s `finally`/catch now dispatches via `this.ctx.statelogClient?.threadEndHookError?.(...)` instead of `console.error`.

packages/agency-lang/stdlib/threads.agency
egonSchiele: Done in 6620bc9c: opening block now starts `/** @module Cross-thread context sharing — ...`.

packages/agency-lang/stdlib/threads.agency
egonSchiele: Done in 6620bc9c — `summarize()` now has a triple-quoted Agency docstring so it propagates to docs / tool descriptions. See #5a and #5c for the related parts of this comment thread.

packages/agency-lang/stdlib/threads.agency
egonSchiele: Done in 6620bc9c. `summarize()` carries a triple-quoted Agency docstring covering its purpose and the thread-isolation rationale.

packages/agency-lang/stdlib/threads.agency
egonSchiele: Done in 6620bc9c: switched to structured output. `summarize()` now declares `type SummaryResult = { summary: string }` and binds the LLM call to that type, then returns `result.summary`. No more extraneous "Here is the summary..." prefixes leaking through.

packages/agency-lang/stdlib/threads.agency
egonSchiele: Resolved as part of fix #6 (6620bc9c): the cache was deleted entirely (label and summary live on `MessageThread` directly now), which removed the `_remember` / `_getCached` imports. No name collision with `std::memory.remember` left.

packages/agency-lang/stdlib/threads.agency
egonSchiele: Yes — shared across runs in a long-lived process was a real concern. Fixed in 6620bc9c by deleting the cache entirely and storing `label` and `summary` directly on `MessageThread` (with `toJSON`/`fromJSON` round-trip). MessageThreads live inside `ThreadStore`, which is per-run, so there's no shared state across calls anymore. The TS-side `registerGlobalHook("onThreadEnd", ...)` callback also went away (label is captured at thread *create* time now).

packages/agency-lang/stdlib/threads.agency
egonSchiele: Wrote a regression test for this — `tests/agency-js/threads-fix-callback-propagation/` imports `std::threads` and a `mod.agency` whose top-level `callback("onThreadEnd")` writes to a shared mutable. The test confirms `importedCallbackFired: false`: a top-level `callback(...)` in an imported (non-entry) Agency module does NOT fire. So we kept the TS-side `registerGlobalHook("onThreadEnd")` for now (with the comment updated to point at the test). Happy to file a separate issue if you want top-level callbacks to propagate cross-module — would unblock a clean Agency-only `std::threads`.

packages/agency-lang/CHANGELOG.md
egonSchiele: Reverted in 6620bc9c — the `## Unreleased` section is removed.

packages/agency-lang/stdlib/threads.agency
```diff
@@ -0,0 +1,165 @@
...
+  back on a known field instead of free-form completion text.
+
+  @param messages - The thread's messages to summarize
+  """
+  let transcript = ""
+  for (m in messages) {
+    transcript = transcript + "[" + m.role + "] " + m.content + "\n"
+  }
+  let summary = ""
+  thread {
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. `summarize()` now opens `thread(hidden: true) { ... }` so the summarizer prompt's thread is filtered out of `listThreads()`. Without this fix every lazy summary call would have leaked a visible "summarize this conversation in 1-2 sentences..." thread into the registry. New regression test would land via the eager-summarize wiring follow-up; in the meantime `tests/agency/threads-registry/hidden-arg.agency` exercises the `hidden` plumbing.

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -491,27 +521,127 @@ export class Runner {
...
+    await invokeCallbacks({
+      ctx: this.ctx,
+      name: "onThreadStart",
+      data: {
+        threadId: slug,
+        threadType: parentRaw ? "subthread" : threadType,
+        parentThreadId: parentRaw ? `t${parentRaw}` : undefined,
+        label: opts.label,
+        isResumption,
+      },
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. `onThreadStart` now sources `label` from `threads.get(tid).label` first, falling back to `opts.label`. Resumed threads via `thread(continue: ...)` / `thread(session: ...)` emit the persisted label even when the call site omits the arg.

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -491,27 +521,127 @@ export class Runner {
...
+      try {
+        await invokeCallbacks({
+          ctx: this.ctx,
+          name: "onThreadEnd",
+          data: {
+            threadId: slug,
+            label: opts.label,
+            eagerSummarize: opts.summarize === true,
+            messages: messagesSnapshot,
+          },
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. `onThreadEnd` now prefers `closingThread.label` (the persisted MessageThread label) over `opts.label`, so resumed threads close with stable label payloads. Mirrors the onThreadStart fix.

packages/agency-lang/lib/runtime/agency.ts
```diff
@@ -132,6 +132,113 @@ const threadWith = async <T>(
...
+/** Convert an internal counter string id (`"0"`, `"1"`, ...) to the
+ *  public slug form (`"t0"`, `"t1"`, ...). */
+const toSlug = (rawId: string): string => `t${rawId}`;
+
+/** Strip the leading `t` from a public slug to get the internal
+ *  counter string id. Returns the input unchanged if it does not
+ *  start with `t` — defensive only; codegen always emits the slug
+ *  form. */
+const fromSlug = (slug: string): string =>
+  slug.startsWith("t") ? slug.slice(1) : slug;
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. `fromSlug` now uses the same `/^t\d+$/` tightening as `stripSlug` in runner.ts. `_setThreadSummary` in `lib/stdlib/threads.ts` picked up the same change. All slug-strip call sites now treat slugs consistently.

packages/agency-lang/lib/stdlib/threads.ts
```diff
@@ -0,0 +1,64 @@
...
+  limit: number = 50,
+): Array<{ role: string; content: string }> {
+  const raw = agency.threads.get(id, offset, limit);
+  return raw.map((m) => ({
+    role: String(m.role),
+    content:
+      typeof m.content === "string"
+        ? m.content
+        : JSON.stringify(m.content ?? ""),
+  }));
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. Now coerces `null`/`undefined` content to `""` explicitly before falling back to `JSON.stringify(...)` for genuinely structured content. Tool-call assistant messages no longer surface as the literal string `'""'`.

packages/agency-lang/docs/superpowers/specs/2026-05-30-thread-as-agency-function.md
```diff
@@ -0,0 +1,109 @@
...
+> ⚠️ This document is a **follow-up proposal** captured during PR #225 review. It is intentionally out of scope for that PR. Filing it here so the idea isn't lost.
+
+---
+
+## Problem
+
+The current `thread {}` block is parsed by a hand-rolled parser ([`_threadNamedArgsParser` in `lib/parsers/parsers.ts:3103`](file:///Users/adityabhargava/agency-lang/packages/agency-lang/lib/parsers/parsers.ts#L3103)) and lowered through a bespoke pipeline. As a result:
+
+- Bad argument names produce a parse-time error message that lives inside the parser ("Unknown thread argument: foo. Allowed: label, summarize, ..."). That string is awkward to maintain and drifts every time the allowed-arg list changes.
+- The AST node ([`MessageThread` in `lib/types/messageThread.ts`](file:///Users/adityabhargava/agency-lang/packages/agency-lang/lib/types/messageThread.ts)) carries five+ optional `Expression` fields (`label`, `summarize`, `continueExpr`, `sessionExpr`, and — as of PR #225 — `hidden`) that have to be plumbed through:
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. Replaced all `file:///Users/...` links with repo-relative paths (e.g. `../../../lib/parsers/parsers.ts`).

packages/agency-lang/lib/ir/tsIR.ts
```diff
@@ -320,12 +320,21 @@ export interface TsRunnerStep {
   body: TsNode[];
 }
 
-/** runner.thread(id, method, async (runner) => { body }) */
+/** runner.thread(id, method, opts, async (runner) => { body }).
+ *  `label`, `summarize`, `continueExpr`, `sessionExpr` are optional
+ *  expressions plumbed in by Tasks 3 / 6 / 7. When all are absent the
+ *  emitter elides the opts arg and falls back to the legacy
+ *  `runner.thread(id, method, async ...)` shape. */
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. The comment now correctly says the emitter always emits an opts object (`{}` when empty) and `Runner.thread` relies on the uniform 4-arg signature. Matches the dual-signature drop from comment #11.

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -491,27 +521,127 @@ export class Runner {
...
+    const parentRaw = threads.get(tid)?.parentId ?? undefined;
+    await invokeCallbacks({
+      ctx: this.ctx,
+      name: "onThreadStart",
+      data: {
+        threadId: slug,
+        threadType: parentRaw ? "subthread" : threadType,
+        parentThreadId: parentRaw ? `t${parentRaw}` : undefined,
+        label: opts.label,
+        isResumption,
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af (same as #3328134829 — labelled the second occurrence too). `onThreadStart` payload now prefers persisted label over `opts.label`.

packages/agency-lang/lib/runtime/runner.ts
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af (same as #3328134833). `onThreadEnd` payload now prefers persisted label over `opts.label`.

packages/agency-lang/docs/superpowers/specs/2026-05-30-thread-as-agency-function.md
```diff
@@ -0,0 +1,109 @@
...
+So every new named argument is a five-file edit. **Concrete example: the `hidden` arg added in PR #225** required touching exactly that fan-out — the most recent demonstration of the cost.
+
+---
+
+## Proposal
+
+`thread {}` becomes pure syntactic sugar for a stdlib function call. The source:
+
+```agency
+thread label="x", hidden=true {
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. Spec now shows the actual current syntax: `thread(label: "x", hidden: true) { ... }`.

packages/agency-lang/docs/site/guide/cross-thread-context.md
```diff
@@ -0,0 +1,199 @@
...
+1. A way to *see* the other threads in the run.
+2. A way to *resume* one when the user circles back.
+
+The `std::threads` module gives you both:
+
+```ts
+import { listThreads, getThread, currentThreadId } from "std::threads"
+```
+
+And the `thread {}` block grows four optional named arguments:
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. Updated count from "four" to "five" optional named arguments to reflect the `hidden` addition.

packages/agency-lang/docs/site/guide/cross-thread-context.md
```diff
@@ -0,0 +1,199 @@
...
+
+```ts
+thread(label: "coding") { /* ... */ }                        // tag it
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. The quick-list code block now includes `thread(summarize: true) { /* ... */ }` between the `label` and `continue` examples.

packages/agency-lang/lib/runtime/state/threadStore.ts
```diff
@@ -108,6 +118,51 @@ export class ThreadStore {
...
+  }
+
+  /** Open a named session. Returns `{id, existed}` — `existed` is
+   *  `true` when this name already mapped to a thread. Always leaves
+   *  the session's thread on top of `activeStack`. First entry
+   *  creates a top-level thread (never a subthread); later entries
+   *  resume via `resumeExisting` (which already rejects subthreads —
+   *  sessions can only map to top-level threads). */
+  openSession(name: string): { id: MessageThreadID; existed: boolean } {
+    const existing = this.sessions[name];
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. `ThreadStore.sessions` is now backed by `Object.create(null)` (both at construction and during `fromJSON` rebuild, which iterates `Object.entries(...)` into a fresh null-proto bag). Session names like `"__proto__"` or `"constructor"` can no longer mutate the Object prototype.

packages/agency-lang/docs/superpowers/specs/2026-05-30-thread-as-agency-function.md
```diff
@@ -0,0 +1,109 @@
...
+
+- Bad argument names produce a parse-time error message that lives inside the parser ("Unknown thread argument: foo. Allowed: label, summarize, ..."). That string is awkward to maintain and drifts every time the allowed-arg list changes.
+- The AST node ([`MessageThread` in `lib/types/messageThread.ts`](file:///Users/adityabhargava/agency-lang/packages/agency-lang/lib/types/messageThread.ts)) carries five+ optional `Expression` fields (`label`, `summarize`, `continueExpr`, `sessionExpr`, and — as of PR #225 — `hidden`) that have to be plumbed through:
+  1. the parser (`_threadNamedArgsParser`),
+  2. the AST type (`MessageThread`),
+  3. the IR (`TsRunnerThread`),
+  4. the codegen emitter ([`processMessageThread` in `lib/backends/typescriptBuilder.ts:2767`](file:///Users/adityabhargava/agency-lang/packages/agency-lang/lib/backends/typescriptBuilder.ts#L2767)),
+  5. the IR pretty-printer ([`prettyPrint.ts:300`](file:///Users/adityabhargava/agency-lang/packages/agency-lang/lib/backends/prettyPrint.ts#L300)), and
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 363050af. Local `file:///Users/...` links replaced with repo-relative paths; `lib/backends/prettyPrint.ts` → `lib/ir/prettyPrint.ts` (the correct location).

packages/agency-lang/lib/parsers/messageThread.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  it("parses thread(session: \"coding\") { ... }", () => {
+    const input = 'thread(session: "coding") {\n  foo()\n}';
+    const r = messageThreadParser(normalizeCode(input));
+    expect(r.success).toBe(true);
+    if (r.success) {
+      expect(r.result.sessionExpr).not.toBeNull();
+      expect(r.result.continueExpr).toBeNull();
+    }
+  });
+
```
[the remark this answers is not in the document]
egonSchiele: Added in 363050af. Three new cases in `messageThread.test.ts`: `thread(hidden: true)`, combined `label` + `hidden`, and `subthread(hidden: true)`. Also tightened the no-args case to assert `r.result.hidden` is null. 11 tests total.

packages/agency-lang/lib/runtime/state/threadStore.ts
```diff
@@ -108,6 +121,51 @@ export class ThreadStore {
...
+   *  v1: rejects subthreads. A subthread's identity is tied to its
+   *  parent's context at the time it was created; resuming one
+   *  outside that context could surface confusing message ordering.
+   *  If a user wants to continue inside a subthread, they should
+   *  resume the parent thread and open a fresh `subthread {}` block.
+   */
+  resumeExisting(id: MessageThreadID): void {
+    const thread = this.threads[id];
+    if (!thread) {
+      throw new Error(`Cannot resume unknown thread id: t${id}`);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in d92caa49. `resumeExisting` now only prepends `t` when the id matches `/^\d+$/`. Non-numeric ids (slugs accidentally passed in, test mocks) pass through unchanged so we no longer surface `tt1` in the error message. Added regression tests in `threadStore.test.ts` covering both the numeric (`"42"` → `t42`) and slug (`"t1"` → `t1`, not `tt1`) paths.

packages/agency-lang/lib/utils/node.ts
```diff
@@ -327,6 +327,18 @@ export function* walkNodes(
...
+      // resolver populates the `scope` field on any variable
+      // references inside them — otherwise codegen emits bare
+      // identifiers instead of `__stack.locals.foo`.
+      if (node.type === "messageThread") {
+        if (node.label) yield* walkNodes([node.label as AgencyNode], [...ancestors, node], scopes);
+        if (node.summarize) yield* walkNodes([node.summarize as AgencyNode], [...ancestors, node], scopes);
+        if (node.continueExpr) yield* walkNodes([node.continueExpr as AgencyNode], [...ancestors, node], scopes);
+        if (node.sessionExpr) yield* walkNodes([node.sessionExpr as AgencyNode], [...ancestors, node], scopes);
+      }
       yield* walkNodes(node.body, [...ancestors, node], scopes);
```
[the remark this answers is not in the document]
egonSchiele: Real bug, fixed in d92caa49. `walkNodes` now visits `node.hidden` alongside the other named-arg expressions on messageThread nodes. Added end-to-end regression test at `tests/agency/threads-registry/hidden-arg-variable.agency` that uses `const shouldHide: boolean = true` inside `thread(hidden: shouldHide) { ... }` — before the walker fix this would crash at runtime with "shouldHide is not defined".

PR #226, opened 2026-05-30
stdlib/threads: wire `thread(summarize: true)` to an eager onThreadEnd hook

egonSchiele:
Wires `thread(summarize: true) { ... }` end-to-end. Until now the
`eagerSummarize: true` payload on `onThreadEnd` was emitted by
`Runner.thread` but nothing consumed it; summaries were always
computed lazily on the first `listThreads()` call.

## What changed

- **`lib/stdlib/threads.ts`** — registers a global `onThreadEnd` hook
  (`_eagerSummarizeIfNeeded`) that fires a one-shot summarize LLM
  call on a standalone `MessageThread` (so the summarizer prompt
  does not pollute the closed thread or the active thread) and
  writes the result back via `_setThreadSummary`. Errors are
  swallowed — the lazy `summaryFor()` path in
  `stdlib/threads.agency` is the correctness fallback.
- **Why a TS-side hook (not Agency-side)** — a top-level
  `callback("onThreadEnd")` inside an Agency stdlib module only
  fires when that module is the run entry-point (see the regression
  test at `tests/agency-js/threads-fix-callback-propagation`). A
  global TS hook bypasses that, so any program that imports
  `std::threads` gets eager summarize for free.
- **`stdlib/threads.agency`** — module docstring and
  `listThreads()` docstring rewritten to describe the actual
  eager-vs-lazy behaviour instead of the prior "v1 stdlib does not
  yet act on it" placeholder.

## Tests

- `lib/stdlib/threads.test.ts` — focused unit coverage for the hook
  (mocks `runPrompt`): writes summary on success, uses a standalone
  `MessageThread`, no-ops on `eagerSummarize:false` / empty messages
  / already-summarized threads, swallows LLM errors, tolerates
  non-canonical (non-slug) thread ids.
- `tests/agency/threads-registry/eager-summarize.agency` — end-to-end
  integration test that `thread(summarize: true)` populates the
  cache at close while a sibling `thread { ... }` block still goes
  through the lazy path on the next `listThreads()`.

## Verification

- `pnpm vitest run lib/stdlib/threads.test.ts` → 7/7 pass.
- `pnpm run agency test tests/agency/threads-registry/eager-summarize.agency` → 1/1 pass.
- `make` builds cleanly (stdlib regen + docs regen).

## Follow-up

Unblocks the multi-thread refactor of `lib/agents/agency-agent/agent.agency`
(see `docs/superpowers/plans/2026-05-29-agency-agent-multi-thread.md`),
which relies on `thread(session: ..., summarize: true)` for snappy
`/sessions` output.

### DIFF
--- a/packages/agency-lang/lib/stdlib/threads.ts
+++ b/packages/agency-lang/lib/stdlib/threads.ts
@@ -28,6 +28,20 @@ import * as smoltalk from "smoltalk";
 import { agency, type ThreadInfoTS } from "../runtime/agency.js";
 import { registerGlobalHook } from "../runtime/hooks.js";
 import { MessageThread } from "../runtime/state/messageThread.js";
+import { createLogger } from "../logger.js";
+
+/** Coerce a `smoltalk` message's `content` to a flat string. Non-string
+ *  content (tool-call / structured) is JSON-stringified; nullish
+ *  content (common on tool-call assistant messages where the LLM
+ *  emitted tool calls but no text) maps to `""` instead of the literal
+ *  JSON-encoded `'""'`. Shared by `_getThread` (the Agency-facing
+ *  reader) and `_buildSummaryTranscript` (the eager summarizer
+ *  prompt) so both surfaces agree on the same coercion rule. */
+function _contentToString(content: smoltalk.MessageJSON["content"]): string {
+  if (typeof content === "string") return content;
+  if (content == null) return "";
+  return JSON.stringify(content);
+}
 
 /** Pass-through to `agency.threads.list()`. */
 export function _listThreadsRaw(): ThreadInfoTS[] {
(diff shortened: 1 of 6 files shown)


packages/agency-lang/lib/stdlib/threads.ts
```diff
@@ -71,3 +84,88 @@ export function _setThreadSummary(id: string, summary: string): void {
...
+
+/** Build the transcript string used by both the lazy and eager
+ *  summarize prompts. Keeps the shape symmetric with the Agency-side
+ *  `summarize()` def. Non-string content gets coerced the same way
+ *  `_getThread()` does to keep the prompt deterministic. */
+function _buildSummaryTranscript(messages: smoltalk.MessageJSON[]): string {
+  let transcript = "";
+  for (const m of messages) {
+    const content =
+      typeof m.content === "string"
```
egonSchiele: hard to read ternary

packages/agency-lang/lib/stdlib/threads.ts
```diff
@@ -71,3 +84,88 @@ export function _setThreadSummary(id: string, summary: string): void {
...
+        // Standalone MessageThread — not in the active ThreadStore,
+        // so listThreads() never sees the summarizer prompt. Same
+        // isolation guarantee as `thread(hidden: true) { ... }` on
+        // the Agency side.
+        thread: new MessageThread(),
+      },
+    );
+    _setThreadSummary(evt.threadId, result.summary);
+  } catch {
+    // Best-effort — lazy summarize will retry on next listThreads().
```
egonSchiele, line 172: still log error. Use logger.debug. Also log to statelog

packages/agency-lang/tests/agency/threads-registry/eager-summarize.agency
```diff
@@ -0,0 +1,33 @@
+import { listThreads } from "std::threads"
```
egonSchiele, line 1: if eager summaries broke, would this test fail?

PR #228, opened 2026-05-30
agency-agent: split into code + research subagents with handoff routing

egonSchiele:
Refactors the Agency Agent from a single 23-tool REPL into a multi-thread architecture where each specialist subagent has its own file, system prompt, tool set, and thread session.

## Architecture

```diagram
╭────────────────╮      handoff()       ╭───────────────────╮
│  code agent    │ ◀──────────────────▶ │  research agent   │
│                │                      │                   │
│ file ops       │                      │ docSkill          │
│ shell          │                      │ cliSkill          │
│ typecheck      │                      │ appendixSkill     │
│ highlight      │                      │ wikipedia*        │
│ remember/recall│                      │ fetch / fetchJSON │
│ print          │                      │ browserUse        │
│                │                      │ remember/recall   │
│ thread:"code"  │                      │ thread:"research" │
│ memory:"code"  │                      │ memory:"research" │
╰────────────────╯                      ╰───────────────────╯
        ▲                                       ▲
        │                                       │
        ╰────────────── routing ────────────────╯
                  │
                  ▼
        ╭────────────────────╮
        │ agent.agency       │
        │   REPL + handler   │
        │   handoff loop     │
        ╰────────────────────╯
```

## Files

- **`lib/agents/agency-agent/shared.agency`** — `Category` type, `AgentContext`, `handoff(category, reason)` tool + `consumeHandoff()` plumbing.
- **`lib/agents/agency-agent/code.agency`** — code specialist (file ops, shell, typecheck, syntax highlight). Owns the `workspace` state + `setCwd` tool. NO docs / web access.
- **`lib/agents/agency-agent/research.agency`** — research specialist (docs Skills, Wikipedia, fetch, browserUse). NO write / edit / bash.
- **`lib/agents/agency-agent/agent.agency`** — REPL loop, interrupt handler, policy state, routing loop with hop limit.
- **`docs/superpowers/ideas/2026-05-30-routing-stickiness.md`** — design rationale and a catalogue of alternative routing techniques.

## Routing design

Rather than categorizing every user message (which over-eagerly re-routes ambiguous follow-ups and strands the new specialist with no conversation context), each subagent gets a `handoff(category, reason)` tool. If the current agent can handle the message, it just answers. If clearly out of scope, it calls `handoff()` and the main loop re-runs the user's original message in the chosen specialist's thread.

The current agent is the most-informed judge of whether a follow-up message is in-scope — it has the full thread history. This biases hard toward "stay" without any extra LLM calls. The full design space (sticky-default categorizer, probability distributions, hysteresis, short-message bypass, embedding similarity, in-session calibration, auto-context-injection on switch) is documented in the new ideas file for future experiments.

## Per-thread continuity

Every routed turn runs inside `thread(session: category, label: category, summarize: true) { ... }`. Re-entering a category resumes its prior conversation via the session mechanism. Closed sessions are eagerly summarized via the hook in `lib/stdlib/threads.ts` so future `listThreads()` calls don't pay a lazy round-trip.

## Memory scoping

Each subagent calls `setMemoryId("code")` / `setMemoryId("research")` so each agent's knowledge graph stays separate. A "this user prefers tabs" fact in the code thread doesn't pollute the research thread, and vice versa.

## Verification

End-to-end smoke tests against the real LLM:

- **Stay** — `hello` → defaults to code, gets a coding-assistant reply, no handoff.
- **Handoff** — `search wikipedia for 'Alan Turing'...` → code agent calls `handoff({"category":"research","reason":"User requested a summary from Wikipedia."})`, main loop prints `→ handing off to research`, research agent picks up and starts using docSkill.

Both compile clean (`make` + `pnpm run lint:structure` both green; no type warnings on the new files).

## Out of scope (per plan edits)

- Two subagents (code + research) instead of the originally-planned four. More specialists are easy follow-ups now that the infrastructure is in place.
- `/sessions` REPL shortcut deferred — `listThreads`/`getThread` are still available to add to a tool set later when we want LLM-side session introspection.
- Docs guide update and CHANGELOG entry deferred.

## Follow-up work tracked in the ideas doc

The routing-stickiness ideas file lists 11 alternative routing techniques. The recommended stack for a future "best-effort" implementation is short-message bypass + hysteresis + embedding similarity + context-aware categorizer + auto-context-injection — each is documented with cost / effort / failure-mode analysis.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -1,37 +1,41 @@
-import { typecheck } from "std::agency"
-import { browserUse } from "std::browser"
 import { today } from "std::date"
-import { Workspace, openDir } from "std::fs"
-import { setMemoryId, remember, recall } from "std::memory"
-import { fetch, fetchJSON, fetchMarkdown } from "std::http"
 import { Policy, PolicyRule, checkPolicy } from "std::policy"
 import { exists } from "std::shell"
-import { skillsDir } from "std::skills"
-import { highlight } from "std::syntax"
 import { cwd } from "std::system"
-import { systemMessage } from "std::thread"
-// Wikipedia exports a bare `search`; alias it so the LLM sees a
-// self-describing tool name in its tool list.
-import {
-  search as wikipediaSearch,
-  summary as wikipediaSummary,
-  article as wikipediaArticle,
-} from "std::wikipedia"
+import { AgentContext, consumeHandoff } from "./shared.agency"
+import { runCode } from "./code.agency"
+import { runResearch } from "./research.agency"
+
 /*
  * Agency Agent — helps users create and modify Agency code.
  *
- * Features:
- * - Doc-as-skills: bundles `docs/site/guide` (copied to
- *   `dist/lib/agents/docs/guide` at build time, see makefile) and
- *   exposes each Markdown file as a skill the LLM can read on demand.
- * - Typechecking: the LLM can call `typecheck` on any source it
- *   generates to verify the code before showing it to the user.
+ * Multi-thread architecture: the agent splits work across two
+ * specialists, each with its own thread / system prompt / tool set:
+ *   - `code.agency`     — writes, edits, typechecks, runs code
+ *   - `research.agency` — looks up docs, fetches URLs, browses web
+ *
+ * Routing: we don't categorize on every turn. The current
+ * specialist's LLM gets a `handoff(category, reason)` tool. If it
+ * can handle the message, it just answers. If clearly out of scope,
+ * it calls `handoff()` and this loop re-runs the user's message in
+ * the new category. See `shared.agency` for the design rationale and
+ * `docs/superpowers/ideas/2026-05-30-routing-stickiness.md` for the
+ * alternatives we deliberately deferred.
+ *
+ * Other features:
+ * - Doc-as-skills: bundles `docs/site/guide` etc. (copied to
+ *   `dist/lib/agents/docs/` at build time, see makefile) and
+ *   exposes each Markdown file as a skill the research agent can
+ *   read on demand.
+ * - Typechecking: the code agent can call `typecheck` on any
+ *   source it generates to verify the code before showing it.
  * - File I/O tools: `read`, `write`, `edit` for actually creating /
  *   editing `.agency` files.
  * - Interactive policy: every interrupt the LLM raises pops a
  *   prompt with approve / reject / always-approve / always-reject.
- *   "Always" decisions are recorded in a per-run Policy and replayed
- *   automatically for the rest of the session via `checkPolicy`.
+ *   "Always" decisions are recorded in a per-run Policy and
+ *   replayed automatically for the rest of the session via
+ *   `checkPolicy`.
  *
  * Run: `pnpm run agency run dist/lib/agents/agency-agent/agent.agency`
  *      (run from `dist/` so the bundled docs at `../docs/guide` resolve.)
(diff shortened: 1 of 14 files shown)


packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -357,143 +372,114 @@ def defaultHandler(intr) {
...
+          print(color.cyan("→ handing off to ${target}"))
+          category = target
+          hop = hop + 1
+        }
+      }
+      if (!done) {
+        print(
+          color.yellow("(handoff hop limit reached; answering in ${category})"),
+        )
+        const finalReply = runForCategory(category, userMsg, ctx)
```
egonSchiele: Do we want to pass a flag here that removes the handoff tool as an option, to make sure the agent responds?

packages/agency-lang/lib/agents/agency-agent/shared.agency
```diff
@@ -0,0 +1,97 @@
...
+// — the `Category` union does not contain `""`, so the sentinel is
+// unambiguous. The `handoff` tool only ever writes a valid Category.
+let _handoffTarget: string = ""
+
+export safe def handoff(category: Category, reason: string): string {
+  """
+  Re-route the user's current message to a different specialist. Use
+  this ONLY when the user's message is clearly outside your scope:
+
+  - Call `handoff("code", reason)` for write/edit/typecheck/run
```
egonSchiele: This introduces the risk that the coding agent may call `handoff("code", reason)`, and similar for the research agent. A couple suggestions:
- Use PFA (https://agency-lang.com/guide/partial-application.html) to limit the valid handoff category values for each agent. This will catch issues like the coding agent trying to handoff to itself. The handoff function can then return an error message to the agent.
- Use .describe() to set a better description of the handoff tool that is specific to each agent. For the coding agent, you would only tell it to handoff to the research agent, and for the research agent, you would only tell it to handoff to the coding agent.

packages/agency-lang/lib/agents/agency-agent/research.agency
```diff
@@ -0,0 +1,148 @@
...
+      web search or a page that requires JavaScript / a login flow.
+      It is the heaviest option and costs money per call.
+  Cite the source URL in your reply whenever you used one of these.
+
+- You have a persistent knowledge graph scoped to research work.
+  Call `recall(query)` to retrieve anything the user told you in a
+  previous session. Call `remember(content)` to persist a fact
+  worth keeping across sessions — user preferences ("prefers
+  brief replies"), recurring topics, decisions made earlier. Don't
+  `remember` transient details. Relevant facts are also
```
egonSchiele, line 97: ``
Relevant facts are also
  auto-injected before each turn, so often you don't need to call
  `recall` explicitly
```

Are you sure? Where does the injection happen?

File: packages/agency-lang/lib/agents/agency-agent/research.agency [hunk 4]
```diff
@@ -0,0 +1,148 @@
...
+
+  Returns the LLM's text reply. If the LLM called `handoff()` mid-
+  turn, the reply is the LLM's post-handoff filler text and the
+  caller MUST check `consumeHandoff()` to discover whether to
+  re-route.
+
+  @param userMsg - The user's input message for this turn
+  @param ctx - Per-call invariants (env, projectContext) for the system message
+  """
+  setMemoryId("research")
```
egonSchiele, line 140: Also call enableMemory and store memories in `~/.agency-agent/memory/`

packages/agency-lang/lib/agents/agency-agent/research.agency
```diff
@@ -0,0 +1,148 @@
...
+  // First-entry detection — see the matching comment in `code.agency`
+  // for the rationale.
+  const t = getThread(currentThreadId())
```
egonSchiele: There has to be a better way to do it. Why not simply keep a counter of how many times the research agent has been called, or a Boolean flag like `firstTimeCalled` that gets it to false.

PR #236, opened 2026-05-31
stdlib: lift policy handler and multi-agent router into the standard library

egonSchiele:
## Summary

Lifts the agency-agent's interactive policy handler and multi-agent
routing/handoff machinery into the standard library, then refactors
the agency-agent to use them. The agent collapses from ~750 lines of
agent-local plumbing into a ~150-line wrapper that wires two
specialists into one `route()` call.

Specs / plan:
- [Spec](docs/superpowers/specs/2026-05-30-stdlib-policy-and-router-design.md)
- [Plan](docs/superpowers/plans/2026-05-30-stdlib-policy-and-router.md)

### `std::policy`

- Types: `Decision`, `FieldSpec`, `AlwaysFields`
- Pure primitives: `recordRule`, `buildScopedMatch`,
  `recordScopedRule`, `parsePolicyFile`
- CLI sugar: `cliPolicyHandler({file, fields})` — installable on a
  `handle` block, owns load + save + prompt + record. Uses the
  flip-flag-first pattern so the policy file's own `std::read`
  doesn't recurse.
- `flushPolicy()` — force-flush between turns (the auto-flush only
  fires on the next interrupt).

### `std::agent`

- Types: `AgentSpec`, `RouterConfig`
- `route(config, userMsg)` — multi-specialist hop loop. Each
  specialist gets a PFA-bound `handoff` tool injected; when the
  hop limit is reached, the last call strips `handoff` so the LLM
  is forced to answer.
- `handoff(category, reason, validCategories)` and
  `consumeHandoff()` — the underlying signal primitives. Module-
  level state lives in the GlobalStore so concurrent runs in
  separate RuntimeContexts don't collide.

### Codegen fix

The generator previously emitted bare `__call(handlerName, ...)`
for `handle ... with <handler>`, ignoring scope. Locally-bound
handlers like `const h = cliPolicyHandler(...); handle { } with h`
crashed at runtime. Fixed in `lib/types/handleBlock.ts`,
`lib/preprocessors/typescriptPreprocessor.ts`,
`lib/backends/typescriptBuilder.ts` — the preprocessor now
resolves the handler ref's scope and the builder emits
`scopedVar(name, scope, moduleId)` when scope is present.

### Agency-agent refactor (acceptance test)

`lib/agents/agency-agent/*` collapses onto the new stdlib. Deleted:
`defaultHandler`, `loadPolicy`, `savePolicy`, `ensurePolicyLoaded`,
`recordAlways`, `recordAlwaysScoped`, `buildScopedMatch`,
`describeScopedMatch`, `askUser`, `runForCategory`, `MAX_HOPS`,
`runCode`, `runResearch`, `_handoffTarget`, `handoff`,
`consumeHandoff`, `Decision`/`FieldSpec`/`AgentContext` types.

The new `agent.agency` is just the banner, REPL loop, the per-kind
`ALWAYS_FIELDS` map, and a single `route(...)` call inside `handle
{ ... } with handler`.

## Tests

All 4698 existing unit tests still pass.

New tests:
- `tests/agency/policy-record-rule/`
- `tests/agency/policy-record-scoped-rule/`
- `tests/agency/policy-build-scoped-match/`
- `tests/agency-js/policy-parse-file/`
- `tests/agency-js/cli-policy-handler/` — full turn-1 records,
  turn-2 auto-approves flow
- `tests/agency/router-basic/`
- `tests/agency/router-handoff/`
- `tests/agency/router-fallback/`

Manual smoke test: the agency-agent boots, prints its banner,
accepts input, and exits cleanly. The interactive paths (handoff
to research, "aa"/"ap" prompts, policy persistence to
`~/.agency-agent/policy.json`) are not exercised by CI and need a
manual session to confirm end-to-end.

## Diff stat

```
stdlib/policy.agency           +200 lines
stdlib/agent.agency            +130 lines
lib/agents/agency-agent/*      -630 lines net
```

Roughly matches the plan's predicted shape, with the agent
collapse coming in a bit deeper than the predicted -300.

## Known caveat

`codeTools` and `researchTools` are now `static const` arrays whose
`workspace.read`/etc. references are captured once at module load.
The old `runCode` rebuilt the array on every call, so an LLM-driven
`setCwd` re-anchored future tools. With the static array, `setCwd`
updates the `workspace` `let` but the cached method references stay
bound to the original dir. If this surfaces in practice, the fix is
to switch `codeTools` to a `def` and call it from the router-config
site, or to mutate the existing workspace object in `setCwd` instead
of rebinding the variable.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -2985,7 +2985,10 @@ export class TypeScriptBuilder {
     }
   }
 
-  private buildHandlerArrow(handlerName: string): TsNode {
+  private buildHandlerArrow(
+    handlerName: string,
+    handlerScope?: ScopeType,
+  ): TsNode {
     if (this.names.isDirectCallFunction(handlerName)) {
       // Built-in handler (approve/reject/propagate): call with no args
       return ts.arrowFn(
@@ -2997,13 +3000,19 @@ export class TypeScriptBuilder {
 
     const args = [ts.id("__data")];
 
-    // User-defined function handler: use __call
+    // User-defined function handler: use __call. Resolve the callee
+    // through scopedVar when scope info is available so locals
+    // (`__stack.locals.NAME`), globals, statics, etc. are dereferenced
+    // correctly instead of being emitted as bare JS identifiers.
     const descriptor = ts.obj({
       type: ts.str("positional"),
       args: ts.arr(args),
     });
     const configObj = this.buildStateConfig();
-    const callArgs: TsNode[] = [ts.id(handlerName), descriptor];
+    const callee = handlerScope
+      ? ts.scopedVar(handlerName, handlerScope, this.moduleId)
+      : ts.id(handlerName);
+    const callArgs: TsNode[] = [callee, descriptor];
     if (configObj) callArgs.push(configObj);
     const callExpr = ts.call(ts.id("__call"), callArgs);
     return ts.arrowFn(
@@ -3036,7 +3045,10 @@ export class TypeScriptBuilder {
         { async: true },
       );
     } else {
-      handler = this.buildHandlerArrow(node.handler.functionName);
+      handler = this.buildHandlerArrow(
+        node.handler.functionName,
+        node.handler.scope,
+      );
     }
 
     // Body: process each statement with substep tracking
--- a/packages/agency-lang/stdlib/policy.agency
+++ b/packages/agency-lang/stdlib/policy.agency
@@ -3,6 +3,8 @@ import {
   _validatePolicy,
   _writePolicyFile,
  } from "agency-lang/stdlib-lib/policy.js"
+import { exists } from "std::shell"
+import { dirname, basename } from "std::path"
 
 export type InterruptDataKey = string
 
@@ -16,6 +18,27 @@ export type PolicyRule = {
 }
 
 export type Policy = Record<InterruptKind, PolicyRule[]>
+
+export type Decision =
+  | "approve"
+  | "reject"
+  | "approve-always"
+  | "approve-always-here"
+  | "reject-always"
+
+export type FieldSpec = {
+  field: string;
+  wildcardSubpaths: boolean
+}
+
+export type AlwaysFields = Record<InterruptKind, FieldSpec[]>
+
+// Internal state for `cliPolicyHandler`. Singleton — calling
+// `cliPolicyHandler` more than once silently overwrites these.
+let _opts: { file: string, fields: AlwaysFields } = { file: "", fields: {} }
+let _policy: Policy = {}
+let _loaded: boolean = false
+let _pendingSave: boolean = false
 
 export def checkPolicy(
   policy: Record<string, any>,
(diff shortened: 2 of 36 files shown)


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2997,13 +3000,19 @@ export class TypeScriptBuilder {
...
+    // through scopedVar when scope info is available so locals
+    // (`__stack.locals.NAME`), globals, statics, etc. are dereferenced
+    // correctly instead of being emitted as bare JS identifiers.
     const descriptor = ts.obj({
       type: ts.str("positional"),
       args: ts.arr(args),
     });
     const configObj = this.buildStateConfig();
-    const callArgs: TsNode[] = [ts.id(handlerName), descriptor];
+    const callee = handlerScope
```
egonSchiele, line 3012: what is this change for? It seems unrelated to the other changes

packages/agency-lang/stdlib/policy.agency
```diff
@@ -44,3 +157,149 @@ export def writePolicyFile(path: string, policy: Policy, allowedPaths: string[]
...
+    else if (answer == "rr") { result = "reject-always"; valid = true }
+    else if (answer == "ap" && scopedAvailable) { result = "approve-always-here"; valid = true }
+    else if (answer == "ap") {
+      print("(ap) isn't available for ${intr.kind} — pick one of (a)/(r)/(aa)/(rr).")
+    }
+  }
+  return result
+}
+
+export def _handler(intr: any): any {
```
egonSchiele, line 543: leading underscore implies this function is internal, but it is exported. Which is it?

packages/agency-lang/stdlib/policy.agency
```diff
@@ -44,3 +157,149 @@ export def writePolicyFile(path: string, policy: Policy, allowedPaths: string[]
...
+  // Interactive reject: ask for an optional reason so the LLM sees
+  // concrete feedback as the tool's return value.
+  const reason = input("Why are you rejecting? (press enter to skip) ")
+  if (reason == "") { return reject() }
+  return reject(reason)
+}
+
+export def cliPolicyHandler(opts: {
+  file: string,
+  fields: AlwaysFields,
```
egonSchiele: should we provide some sensible defaults for fields?

egonSchiele:
Thanks for the review — agreed on all points. Pushed `dd1f6168` addressing the feedback.

## Module-level docs with usage examples

Both `std::policy` and `std::agent` now open with `/** @module */` blocks (rendered into `docs/site/stdlib/{policy,agent}.md` via `make doc`). They cover:

- What the module is for, organized by feature group (pure primitives vs CLI sugar for policy; router vs question/todos for agent).
- A "Usage: the CLI handler" worked example that's the happy path for most users — `cliPolicyHandler` + `ScopedRuleFields` + `handle ... with handler`.
- A "Usage: writing your own handler" example for non-CLI agents (Slack, web, CI) that composes `checkPolicy` + `recordRule` directly.
- A "How matches work" section explaining glob patterns and what's actually inside a `PolicyRule.match` map.
- A "Precedence trap" section spelling out the foot-gun you raised — see below.
- For the router: a usage example, an explanation of how handoff works under the hood, and what `maxHops` / per-specialist state actually do.

## Renames

- `AlwaysFields` → `ScopedRuleFields`
- `FieldSpec` → `ScopedField`
- `wildcardSubpaths` → `matchSubpaths`

The new names tie back to the concept of "scoped rules" (rules pinned to specific data fields, as opposed to the catch-all `recordRule` produces). All callsites (agency-agent, tests, internal usage) updated.

## Doc comments on public types and functions

Added `/** ... */` doc comments — distinct from the LLM-facing `"""..."""` docstrings — explaining what each type's fields mean, when to use each function, and how they compose. Highlights:

- `ScopedField` now documents `matchSubpaths` with "pass `true` for directory-like fields (so approving `/tmp/x` also approves `/tmp/x/sub/file.txt`); pass `false` for opaque identifiers (commands, IDs, env names) that shouldn't fan out."
- `buildScopedMatch` explains what a "match" is in terms of glob patterns and `PolicyRule.match`, with a code snippet showing how the returned object plugs into a rule.
- `cliPolicyHandler` documents the singleton-state caveat, the bind-to-variable requirement, and the four-step lifecycle.
- `_handler` is marked internal — exposed only because codegen needs to resolve the function reference cross-module.

## Precedence trap on `recordRule`

You're right that the existing docs let the user walk into a shadowing bug. Both the doc comment and the LLM-facing docstring now call this out explicitly:

```ts
/**
 * Return a new policy with a catch-all rule (`{ action }` with no
 * `match`) for `kind` appended. Pure — does not mutate the input.
 *
 * ## Precedence trap
 *
 * Evaluation is first-match-wins, so **append order matters**. A
 * second call for the same kind with a different action is dead
 * code:
 *
 * ```ts
 * let p = recordRule({}, "std::read", "reject")
 * p = recordRule(p, "std::read", "approve")  // never reached
 * ```
 *
 * If you're flipping a previously-recorded decision, decide
 * explicitly: either reset the kind's rules first
 * (`{ ...policy, "std::read": [] }` and re-record), or hand-edit
 * `policy[kind]` to replace the offending rule. This function does
 * not try to detect or warn about shadowing on your behalf.
 */
```

I considered making `recordRule` itself detect and warn on shadowing, but chose to keep it a pure data primitive — auto-warning would require deciding *where* to warn (a returned `Result`? `print`?) and would surprise users composing it programmatically. The docs now make the shape of the bug obvious, and `recordScopedRule` (which is the more user-friendly path) deliberately **prepends** so it always wins over an older catch-all, which is also documented.

## Verification

- All 4698 unit tests still pass.
- The policy primitives tests (`policy-build-scoped-match`, `policy-record-rule`, `policy-record-scoped-rule`) all still pass against the renamed types.
- The agency-agent still compiles + boots.

Let me know if anything is still unclear — happy to push more passes.

PR #237, opened 2026-05-31
Fix cross-module static/global init order (#232)

### LINKED ISSUE #232: Cross-module import of `static const` reads `undefined` if importer's init runs first
## Summary

When module A imports a `static const` X from module B, A's `__initializeGlobals` (and any module-init expressions in A) can read X **before** B's static initializer has run. The binding holds `undefined`, which silently corrupts any computed values that depend on it.

## Repro

`shared.agency`:
```agency
def computeAgentDir(): string {
  const home = env("HOME")
  if (home == null) { return "./.agency-agent" }
  return "${home}/.agency-agent"
}

export static const AGENCY_AGENT_DIR = computeAgentDir()
```

`agent.agency`:
```agency
import { AGENCY_AGENT_DIR } from "./shared.agency"

const POLICY_FILE = "policy.json"
const POLICY_DIR = AGENCY_AGENT_DIR                       // captures undefined
const POLICY_PATH = "${POLICY_DIR}/${POLICY_FILE}"        // "undefined/policy.json"
```

At runtime any code that uses `POLICY_PATH` produces:

```
Could not save policy to undefined/policy.json: ENOENT: no such file or directory, open '/Users/me/work/undefined/policy.json'
```

## What the compiler emits

`shared.js`:
```js
let AGENCY_AGENT_DIR;
let __staticInitPromise = null;
async function __initializeStatic(__ctx) {
  if (__staticInitPromise) return __staticInitPromise;
  __staticInitPromise = (async () => {
    AGENCY_AGENT_DIR = __deepFreeze(await __call(computeAgentDir, ...));
    ...
  })();
  return __staticInitPromise;
}
async function __initializeGlobals(__ctx) {
  __ctx.globals.markInitialized("...shared.agency");
  await __initializeStatic(__ctx);
}
export { AGENCY_AGENT_DIR, ... };
```

`agent.js`:
```js
import { AGENCY_AGENT_DIR } from "./shared.js";   // ← live binding, currently undefined
...
async function __initializeGlobals(__ctx) {
  __ctx.globals.markInitialized("...agent.agency");
  __ctx.globals.set("...", "POLICY_DIR", AGENCY_AGENT_DIR);          // undefined!
  __ctx.globals.set("...", "POLICY_PATH",
    `${...get("POLICY_DIR")}/${...get("POLICY_FILE")}`);              // "undefined/policy.json"
  // ...continues without ever awaiting shared.js's __initializeStatic
}
```

`agent.js`'s `__initializeGlobals` never awaits `shared.js`'s `__initializeStatic`, so the imported binding can still be `undefined` when it's read into `POLICY_DIR`.

It's not just module-init time either: until *something* from `shared.js` is called (anywhere in the program), the binding stays `undefined` for any consumer that re-reads it. So even deferring the read to runtime doesn't fully help — if the first read happens before the first call to any function exported by `shared.js`, you still see `undefined`.

## Why this is bad

1. **Silent corruption.** No throw, no warning. The program just writes to a path called `undefined/policy.json` (or whatever the downstream computation produces).
2. **Hard to debug.** The symptom (ENOENT on a weird path) doesn't point at the cross-module init ordering — see [this debugging session](https://ampcode.com/threads/T-019e7a3a-edfa-74d6-917a-255c31bf8491) where the same project had to inline `computeAgentDir`'s logic locally to dodge the bug.
3. **Spooky action at a distance.** Whether the bug fires depends on whether *some unrelated code path* has happened to invoke a function from the imported module first. That's a load-order foot-gun.

## Suggested fix

When module A imports anything (especially a `static const`) from module B, A's `__initializeGlobals` should `await B.__initializeStatic(__ctx)` (or A's `__initializeGlobals` should be guaranteed to run after B's) before reading B's exports.

A few sketches:

**Option 1 (most targeted):** For every `static const` A imports from B, the codegen prepends `await __ensureModuleInitialized(B)` to A's `__initializeGlobals`. The runtime helper resolves to B's `__initializeStatic`/`__initializeGlobals`. Pure additive change.

**Option 2 (broader):** Treat every cross-module import as a transitive init dependency. A's `__initializeGlobals` awaits the init of every module it imports from. Slightly more conservative — initializes more eagerly than strictly needed — but eliminates the entire class of cross-module ordering bugs.

**Option 3 (read-side):** Replace direct `import { X }` of `static const`s with a function accessor `import { __getX }` whose first call triggers init. Avoids changing module-init order at all; just makes the read self-init. More invasive on the import-site codegen.

I'd lean toward Option 1: minimal, targeted, doesn't risk regressing init order for things that work today.

## Workaround for users

Don't rely on cross-module `static const` imports for values you'll read during your own module's init. Either:
- Re-compute the value locally (what we ended up doing — see the linked debugging session).
- Wrap the value in a `def` and only read it from inside other `def`s, AND ensure something from the source module has already been invoked before the first read.

Both are unsatisfying because the bug is silent — users won't discover the workaround until they hit the symptom.

## Context

Discovered while debugging the persistent-policy storage in the agency-agent: every "always approve here" decision appeared to succeed (no error in the user-visible flow) but the rule was never persisted because `POLICY_PATH` was `undefined/policy.json`.

Related thread: https://ampcode.com/threads/T-019e7a3a-edfa-74d6-917a-255c31bf8491

egonSchiele:
Fixes #232.

## Problem

When module A imports a `static const` (or top-level `let`/`const`) from module B, A's initializer previously read the live ES binding before B's `__initializeStatic` ran, silently producing `undefined` and downstream corruption like `"undefined/policy.json"`.

The agency-agent had been working around this by inlining a duplicate `policyDir` def instead of importing `AGENCY_AGENT_DIR` from `shared.agency`.

## Solution

Each top-level `static const` (and top-level global) compiles to a memoized async **getter** that initializes the variable on first invocation. Compute bodies read other top-level static/global values via getter calls, so the topological ordering of initialization falls out of the runtime call chain — no compile-time DAG, no topo sort.

Cycle detection happens at runtime via a synchronous `_running` flag that fires on re-entry before the recursive await suspends. The thrown error names the variable and the JS stack trace shows the dependency chain.

Two-phase entry orchestration preserves the "all values populated before any imperative runs anywhere" invariant:
1. **Phase 1:** every reachable module's `__initializeStatic` fires.
2. **Phase 2:** every reachable module's `__runImperatives` fires in import order.

`__initializeGlobals` is preserved as a backward-compat shim for lazy first-call init, resume/rewind, and doc-string interpolation eager eval.

## Generated code shape

```js
// shared.js
let AGENCY_AGENT_DIR;
async function __init_AGENCY_AGENT_DIR_compute(__ctx) {
  AGENCY_AGENT_DIR = __deepFreeze(await computeAgentDir());
  return AGENCY_AGENT_DIR;
}
const __init_AGENCY_AGENT_DIR = __initVar(
  "shared.agency:AGENCY_AGENT_DIR",
  __init_AGENCY_AGENT_DIR_compute,
);
const __MY_INIT_GETTERS = [__init_AGENCY_AGENT_DIR];
export { AGENCY_AGENT_DIR, __init_AGENCY_AGENT_DIR, __MY_INIT_GETTERS, __initializeStatic };

// agent.js
import { __init_AGENCY_AGENT_DIR } from "./shared.js";
let POLICY_PATH;
async function __init_POLICY_PATH_compute(__ctx) {
  POLICY_PATH = __deepFreeze(`${await __init_AGENCY_AGENT_DIR(__ctx)}/policy.json`);
  return POLICY_PATH;
}
const __init_POLICY_PATH = __initVar("agent.agency:POLICY_PATH", __init_POLICY_PATH_compute);
```

The reference to `__init_AGENCY_AGENT_DIR` inside `__init_POLICY_PATH`'s compute body IS the dep declaration AND the runtime mechanism that enforces ordering.

## Files

**Runtime:**
- `lib/runtime/initVar.ts` (new) — memoized async getter helper with cycle detection (permanent-failure semantics).
- `lib/runtime/initOrchestrator.ts` (new) — module registry + two-phase orchestrator.
- `lib/runtime/index.ts` — re-exports.
- `lib/templates/backends/typescriptGenerator/imports.mustache` — imports new runtime helpers into every generated module.

**Codegen:**
- `lib/backends/typescriptBuilder.ts` — rewrites init-context reads to `await __init_X(__ctx)`; tracks which imports need getter wrapping.
- `lib/backends/typescriptBuilder/sectionAssembler.ts` — partitions init vars into spec objects; emits per-var named compute fns + getters; splits init into static + imperative phases.
- `lib/compilationUnit.ts` — tracks `importedConstants`.

**Tests:**
- `lib/runtime/initVar.test.ts` — 9 unit tests (memoization, diamond, linear cascade, self-cycle, mutual cycle, permanent failure, sync exception).
- `lib/backends/static-init-cycle.test.ts` — end-to-end cycle error assertion.
- `tests/agency/static-init-cross-module/` — regression for #232.
- `tests/agency/static-init-cross-module-diamond/` — shared dep through diamond, initialized exactly once.
- `tests/agency/static-init-function-mediated/` — `static const X = someFn()` where the importing module reads X transitively.
- `tests/agency/static-init-cycle/cycle-same-module.agency` — runtime cycle fixture.

**Workaround removed:**
- `lib/agents/agency-agent/agent.agency` — restored canonical import of `AGENCY_AGENT_DIR`; added `export static const POLICY_PATH`.

**Docs:**
- `docs/site/guide/execution-model.md` — new section covering cross-module init order, two-phase guarantee, and cycle error UX.

**Fixture updates:**
- Snapshot tests under `tests/typescriptGenerator/` and `tests/typescriptBuilder/` updated to reflect the new per-module `__init_*` / `__MY_INIT_GETTERS` / `__initializeStatic` / `__runImperatives` emit shape.

## Verification

- `make` clean build ✓
- `pnpm run lint:structure` ✓
- New unit tests: 10/10 ✓
- New integration tests: 3/3 ✓
- Cycle test: throws `Init cycle on cycle-same-module.agency:a` with `__init_a` and `__init_b` frames in stack ✓
- All `lib/backends/` + `lib/compilationUnit.test.ts` tests: 252/252 ✓
- Agency-agent compiles and `agent.js` correctly emits `await __init_AGENCY_AGENT_DIR(__ctx)` in the `POLICY_PATH` compute body ✓

The full agency execution test suite will run in CI.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -7,6 +7,7 @@ import { cwd, env } from "std::system"
 
 import { codeSysPrompt, codeTools } from "./code.agency"
 import { researchSysPrompt, researchTools } from "./research.agency"
+import { AGENCY_AGENT_DIR } from "./shared.agency"
 
 
 /*
@@ -62,19 +63,11 @@ callback("onToolCallEnd") as data {
   }
 }
 
-// On-disk policy at `$HOME/.agency-agent/policy.json`. We duplicate the
-// home-dir computation here instead of importing `AGENCY_AGENT_DIR` from
-// `shared.agency` because Agency doesn't guarantee imported static-const
-// init runs before the importer reads it (upstream #232).
-const POLICY_FILE = "policy.json"
-def policyDir(): string {
-  const home = env("HOME")
-  if (home == null) {
-    return "./.agency-agent"
-  }
-  return "${home}/.agency-agent"
-}
-def policyPath(): string { return "${policyDir()}/${POLICY_FILE}" }
+// On-disk policy at `$HOME/.agency-agent/policy.json`. We compose the
+// path from `shared.agency`'s `AGENCY_AGENT_DIR` static const; the
+// cross-module init-order fix (#232) guarantees the value is populated
+// before this static const evaluates.
+export static const POLICY_PATH = "${AGENCY_AGENT_DIR}/policy.json"
 
 // Per-kind config driving the "approve-always-here" option. For each
 // kind we list the data fields the scoped rule should pin, plus a
@@ -141,7 +134,7 @@ node main() {
   // Bind the handler to a local var so `with handler` parses (the
   // `with` clause only accepts an identifier, not a call expression).
   const handler = cliPolicyHandler({
-    file: policyPath(),
+    file: POLICY_PATH,
     fields: ALWAYS_FIELDS,
   })
 
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -7,7 +7,6 @@ import { cwd, env } from "std::system"
 
 import { codeSysPrompt, codeTools } from "./code.agency"
 import { researchSysPrompt, researchTools } from "./research.agency"
-import { AGENCY_AGENT_DIR } from "./shared.agency"
 
 
 /*
@@ -63,11 +62,19 @@ callback("onToolCallEnd") as data {
   }
 }
 
-// On-disk policy at `$HOME/.agency-agent/policy.json`. We compose the
-// path from `shared.agency`'s `AGENCY_AGENT_DIR` static const; the
-// cross-module init-order fix (#232) guarantees the value is populated
-// before this static const evaluates.
-export static const POLICY_PATH = "${AGENCY_AGENT_DIR}/policy.json"
+// On-disk policy at `$HOME/.agency-agent/policy.json`. We duplicate the
+// home-dir computation here instead of importing `AGENCY_AGENT_DIR` from
+// `shared.agency` because Agency doesn't guarantee imported static-const
+// init runs before the importer reads it (upstream #232).
+const POLICY_FILE = "policy.json"
+def policyDir(): string {
+  const home = env("HOME")
+  if (home == null) {
+    return "./.agency-agent"
+  }
+  return "${home}/.agency-agent"
+}
+def policyPath(): string { return "${policyDir()}/${POLICY_FILE}" }
 
 // Per-kind config driving the "approve-always-here" option. For each
 // kind we list the data fields the scoped rule should pin, plus a
@@ -134,7 +141,7 @@ node main() {
   // Bind the handler to a local var so `with handler` parses (the
   // `with` clause only accepts an identifier, not a call expression).
   const handler = cliPolicyHandler({
-    file: POLICY_PATH,
+    file: policyPath(),
     fields: ALWAYS_FIELDS,
   })
 
--- a/packages/agency-lang/tests/agency/global-let-cross-module/bar.agency
+++ b/packages/agency-lang/tests/agency/global-let-cross-module/bar.agency
@@ -1,0 +1,6 @@
+// Source-module side of the global-let cross-module test. Exports a
+// `static const` (the only top-level value Agency permits exporting).
+// The importing module's top-level `const Y = X` is a *global* on
+// the importer side — codegen routes Y through __ctx.globals.set/get,
+// distinct from X's module-level `let` binding.
+export static const X = "/global/x"
--- a/packages/agency-lang/docs/dev/cross-module-init.md
+++ b/packages/agency-lang/docs/dev/cross-module-init.md
@@ -1,0 +1,417 @@
+# Cross-module Static / Global Init Order
+
+This document explains how Agency guarantees that imported `static const` (and top-level `let` / `const`) values are populated before any importer reads them. It is the internals doc for the subsystem that landed as PR #237 and fixed issue #232.
+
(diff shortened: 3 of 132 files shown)


packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -7,6 +7,7 @@ import { cwd, env } from "std::system"
 
 import { codeSysPrompt, codeTools } from "./code.agency"
 import { researchSysPrompt, researchTools } from "./research.agency"
+import { AGENCY_AGENT_DIR } from "./shared.agency"
```
egonSchiele: don't make changes to this file in this pr

packages/agency-lang/tests/agency/global-let-cross-module/bar.agency
```diff
@@ -0,0 +1,6 @@
+// Source-module side of the global-let cross-module test. Exports a
+// `static const` (the only top-level value Agency permits exporting).
```
egonSchiele, line 2: huh. So regular global vars can't be exported? Only static ones?

PR #242, opened 2026-05-31
std::ui PR 3/3: migration + cleanup

egonSchiele:
## std::ui redesign — PR 3 of 3: Migration + cleanup

Third and final PR of the std::ui redesign series. Builds on:
- PR 1 (#TODO) — TS engine prep (`tickMs`, `ui-region.ts`, bridge primitives)
- PR 2 (#TODO) — Agency-side declarative API + `repl()` widget + tests

This PR migrates the rest of the codebase to the new API and removes the old imperative surface.

### What's in this PR

**Policy integration:**
- `_routePrompt` in `stdlib/ui.agency` routes prompts through the active REPL when one is running, falling back to raw stdin otherwise.
- `std::policy` interactive interrupts now use `_routePrompt` so they integrate cleanly with `repl()`.

**Agency-agent rewrite:**
- The agency-agent REPL loop (`lib/agents/agency-agent/agent.agency`) is collapsed onto the new `repl()` widget — replacing ~50 lines of imperative input/output handling with a declarative `repl()` call that gets history, command palette, and live status for free.

**Cleanup / breaking changes:**
- Removed the old imperative `std::ui` API from both `stdlib/ui.agency` and `lib/stdlib/ui.ts`: `initUI`, `destroyUI`, `log`, `status`, `chat`, `code`, `diff`, `separator`, `startSpinner`, `stopSpinner`, `prompt`, `getConfirmation`, `emptyLine`. No compat shims (per spec).
- Removed `examples/coding-agent.agency` (relied entirely on the old API; `lib/agents/agency-agent` is now the canonical example).
- Removed the `tests/integration/stdlib-sandbox/ui` smoke test (exercised the removed imperative functions).
- Added a `/** @module */` doc block to `stdlib/ui.agency` with Layer 1 and Layer 2 usage examples.
- Regenerated `docs/site/stdlib/ui.md`.
- CHANGELOG entry under Unreleased.

**Tests:**
- New scripted `agency-agent` smoke test (`tests/agency-js/agency-agent-smoke/`) confirms the rewritten REPL integrates with the agent plumbing (slash commands, output flow, exit).

### Verification

- `make` — builds cleanly
- `pnpm vitest run lib/stdlib/ui.test.ts lib/tui/test/screen.test.ts` — 7/7 pass
- `pnpm run agency test js tests/agency-js/agency-agent-smoke` — passes
- Manual smoke test of `lib/agents/agency-agent` — banner, input bar, status, history, palette, `/exit` all work.

### Notes for reviewers

The `fixtureTypeCheck.integration.test.ts` flags an "Unknown named argument 'block'" warning on `column` / `row` / `box` calls. This is a pre-existing typechecker quirk around callback-typed parameters and is not introduced by this PR; it does not affect runtime.


packages/agency-lang/lib/stdlib/ui.ts
```diff
@@ -629,13 +156,97 @@ async function callBridgeFn<T>(fn: unknown, ...args: unknown[]): Promise<T> {
   return (await __call(fn, { type: "positional", args })) as T;
 }
 
+export function _activateReplState(state: ActiveReplState): void {
+  activeReplState = state;
```
egonSchiele: if this is a local var in TS, this means that the repl code can now only be used in one process. If this was in Agency instead, it would automatically support concurrent usage:
https://agency-lang.com/guide/execution-model.html

This is probably not a big concern with a repl, where we assume the agent will only be used by one process, but it is not a lot of work, it would still be good to move this into agency so that it can be used concurrently, if needed. That is generally something we try to support in all funcs in the agency stdlib. The other option is to use agency ts helpers, and create a global var:
https://agency-lang.com/appendix/ts-helpers.html

The global var will be isolated per-agent-run.

packages/agency-lang/lib/stdlib/ui.ts
```diff
@@ -629,13 +156,97 @@ async function callBridgeFn<T>(fn: unknown, ...args: unknown[]): Promise<T> {
...
+}
+
+export function _deactivateReplState(): void {
```
egonSchiele: to be clear – are these functions only used for tests?

packages/agency-lang/stdlib/policy.agency
```diff
@@ -507,31 +508,32 @@ def askUser(intr: any): Decision {
...
   }
-  return result
+  menuLines.push("  (rr) reject always")
+  const menuText = menuLines.join("\n")
+
+  let choices: string[] = ["a", "r", "aa", "rr"]
+  if (scopedAvailable) { choices = ["a", "r", "aa", "ap", "rr"] }
+
+  const answer = _routePrompt(menuText, choices)
+  if (answer == "a")  { return "approve" }
```
egonSchiele: match block would be more idiomatic

packages/agency-lang/stdlib/ui.agency
```diff
@@ -249,19 +267,33 @@ export def list(
  */
 export def textInput(
   value: string = "",
-  flex?: number,
-  width?: number,
-  height?: number,
+  flex: number = null,
```
egonSchiele: Does this type check? The type is `number` but the default is `null`? The older code `flex?: number` was correct, why replace it?

packages/agency-lang/stdlib/ui.agency
```diff
@@ -276,40 +308,62 @@ export def textInput(
...
-  if (width != null)        { style.width = width }
-  if (height != null)       { style.height = height }
-  if (padding != null)      { style.padding = padding }
-  if (border)               { style.border = true }
-  if (borderColor != "")    { style.borderColor = borderColor }
-  if (label != "")          { style.label = label }
-  if (bg != "")             { style.bg = bg }
-  if (fg != "")             { style.fg = fg }
-  if (!visible)             { style.visible = false }
+  if (flexDirection != "") {
```
egonSchiele: This pattern is super ugly and it is duplicated over and over. Please refactor into a reusable function.

packages/agency-lang/stdlib/ui.agency
```diff
@@ -601,274 +738,499 @@ export def setQuitAfterMs(ms: number) {
...
+// view and closing it returns them.
 const _PALETTE_ROWS: number = 5
 
-type ReplState = {
+/**
+ * Private — used by other stdlib modules (notably `std::policy`)
+ * to route a prompt through the active REPL when one exists, or
+ * fall back to raw `print` + `input` otherwise.
+ *
+ * Not part of the public API; the underscore prefix is convention,
```
egonSchiele: Remove "Not part of the public API; the underscore prefix is convention,
 * but the export is required so cross-module imports work."

packages/agency-lang/stdlib/ui.agency
```diff
@@ -601,274 +738,499 @@ export def setQuitAfterMs(ms: number) {
...
+ * but the export is required so cross-module imports work.
+ *
+ * @param text - The menu/question text shown to the user
```
egonSchiele: These should go in a docstring https://agency-lang.com/guide/partial-application.html

packages/agency-lang/stdlib/ui.agency
```diff
@@ -601,274 +738,499 @@ export def setQuitAfterMs(ms: number) {
...
+  input: ReplInputState;
+  palette: ReplPaletteState;
+  transcript: ReplTranscriptState;
+  submit: ReplSubmitState;
+  config: ReplConfigState;
+  done: boolean
+}
+
+/**
+ * Append a styled message to the active `repl()` transcript.
```
egonSchiele: All these functions could be used as tools. Tools should have descriptions saying what they're for and how to use them. In order to give each tool a description, they need to have a docstring. That's where the description goes. None of the functions you have written have doc strings, so no LLM will understand how to use any of them. Please add doc strings.
Doc strings are different from doc comments, which is what you have here. A doc string contains a description of a tool that will tell an LLM how to use the tool. A doc comment contains any other extra information that will be useful for a human to know.

packages/agency-lang/stdlib/ui.agency
```diff
@@ -601,274 +738,499 @@ export def setQuitAfterMs(ms: number) {
...
+  return _recordedFrameTexts()
+}
+
+// Returns the keys of palette commands whose name contains the
+// lowercased filter substring. Lists in insertion order.
+def _filteredPaletteKeys(state: ReplState): string[] {
+  const paletteCommands = state.palette.commands
+  let allCommandNames: string[] = []
+  for (entry in entries(paletteCommands)) {
+    allCommandNames.push(entry.key)
```
egonSchiele: Agency's stdlib also has `map`:
https://agency-lang.com/stdlib/array.html

packages/agency-lang/stdlib/ui.agency
```diff
@@ -601,274 +738,499 @@ export def setQuitAfterMs(ms: number) {
...
+    return allCommandNames
   }
-  if (s.paletteFilter == "") { return allKeys }
-  let out: string[] = []
-  const filt = s.paletteFilter.toLowerCase()
-  for (k in allKeys) {
-    if (k.toLowerCase().includes(filt)) { out.push(k) }
+  let matchingCommandNames: string[] = []
+  const filterText = state.palette.filter.toLowerCase()
+  for (commandName in allCommandNames) {
```
egonSchiele: and filter

packages/agency-lang/stdlib/ui.agency
```diff
@@ -601,274 +738,499 @@ export def setQuitAfterMs(ms: number) {
...
-// native scroll region; we append new tails via `_writeScrollLine`
-// inside `_replReduce` rather than rendering them into this tree.
-def _replView(s: ReplState): Element {
-  const stat = s.status()
-  const palKeys = _filteredPaletteKeys(s)
-  let palItems: string[] = []
-  for (k in palKeys) {
-    const desc = s.paletteCommands[k]
-    palItems.push("${k}  - ${desc}")
+// View: the full terminal. `state.transcript.messages` is the single
```
egonSchiele: use a doc comment if you want this info to show up in the docs.

packages/agency-lang/tests/agency-js/agency-agent-smoke/agent.agency
```diff
@@ -0,0 +1,39 @@
...
+  return true
+}
+
+node main(): string {
+  // Sequence:
+  //   "h","i",enter        → submit "hi" → lines becomes ["echo: hi"]
+  //   "/","e","x","i","t"  → palette opens, filter narrows to "/exit"
+  //   enter                → palette selects "/exit" into buffer
+  //   enter                → submit "/exit" → onSubmit returns false, exit
+  setScriptedKeys([
```
egonSchiele: ...where is this function coming from? I don't see it defined or imported anywhere

packages/agency-lang/tests/agency/ui-repl-clear-messages/main.agency
```diff
@@ -0,0 +1,37 @@
+import { clearMessages, pushMessage, recordedFrameTexts, repl, setScriptedKeys } from "std::ui"
```
egonSchiele: These tests aren't really in the spirit of agency tests. Agency tests are supposed to be integration tests that actually test the functionality. These tests seem to mock out a lot of stuff. If that is the case, just write vitest integration tests, the way the debugger does. Is there any value in keeping these as agency tests?

PR #244, opened 2026-05-31
Agent init PR 1 of 6: runtime read-before-init trap

egonSchiele:
PR 1 of 6 in the agent-initialization redesign series. Replaces silent `undefined` reads of uninitialized static variables with loud, actionable runtime errors. Establishes the safety net that subsequent PRs rely on.

Design doc: `agent-init-design.md`
Plan: `docs/superpowers/plans/2026-05-31-agent-init-pr1-read-before-init-trap.md`

## What changes

Today, when one module reads a `static const` from another module before that module's `__initializeStatic` has run, the read silently returns `undefined`, which propagates through expressions like `fooStatic = barStatic + "!"` as `"undefined!"`. This PR turns that silent failure into a loud one.

**Runtime** (`lib/runtime/staticInit.ts`):
- New `__UNINIT_STATIC` sentinel — a unique Symbol that cannot collide with any user value.
- New `__readStatic(value, name, moduleId)` wrapper — returns the value unchanged, except when the value is the sentinel, in which case it throws a clear error naming the variable and suggesting common fixes (circular imports, indirect-through-function-call reads).

**Codegen**:
- Every `static const X = ...` now compiles to `let X = __UNINIT_STATIC;` instead of `let X;`. The assignment inside `__initializeStatic` replaces the sentinel with the real value.
- The IR pretty-printer wraps every read of a static-scoped `scopedVar` with `__readStatic(X, "X", "module.agency")`.
- `variableName` references with scope "imported" that resolve to agency imports are also wrapped — closes the cross-module gap where the sentinel would otherwise propagate into expressions and surface as a confusing JS engine error ("Cannot convert a Symbol value to a string").

The wrap is transparent for initialized values — `__readStatic(x, ...)` returns `x` unchanged — so binary operations, template interpolations, indexing, spreads, and property access all continue to work without any other change. For non-static agency imports (functions, nodes), the wrap is a no-op because the binding never equals the sentinel.

Compiler-emitted boilerplate (e.g. `__registerTool(barStatic)`) builds its source with `ts.raw` and does not flow through the wrapped paths, so module-load-time tool registration is unaffected.

## Caveat

For cross-module reads, the error message currently shows an empty `moduleId` because the codegen site for imported names does not easily know the source module's id. PR 2's cross-module dep graph will fill it in. The variable name is correct, which makes the error actionable.

## Tests

- `lib/runtime/staticInit.test.ts` — 5 unit tests covering plain values, sentinel throws, error-message content, transparency in binary ops / templates / indexing, and Symbol uniqueness.
- `lib/runtime/staticInit.crossModule.test.ts` — end-to-end vitest that compiles a two-file program exhibiting the silent-undefined bug, runs it, and asserts the friendly error fires.
- `tests/agency/static-binary-ops.agency` — Agency integration test exercising statics through `+`, template literals, indexing, property access, and arithmetic; all return correct values, confirming wrap transparency.
- Regenerated `tests/typescriptGenerator/` (83 fixtures), `tests/typescriptBuilder/` (3 fixtures), and `tests/debugger/` (6 fixtures) — almost all changes are just the added import line.

Full `lib/` suite: 4713 pass / 1 fail. The 1 failure is pre-existing on main (stdlib/ui named-arg issue, unrelated to this PR).

## Coming up

This unblocks the rest of the series:
- PR 2 — Per-variable topological sort across the import closure (eliminates most ordering bugs that the trap would otherwise catch; populates the real source moduleId for cross-module wraps).
- PR 3 — `static` prefix on bare top-level statements.
- PR 4 — Compile-time validations for static initializers.
- PR 5 — `agency explain-init` CLI + trace events.
- PR 6 — Docs rewrite.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -773,10 +773,12 @@ export class TypeScriptBuilder {
           // never statics. Compiler-emitted boilerplate (e.g.
           // `__registerTool(name)`) builds its source with `ts.raw`, not
           // through this case, so it is unaffected.
+          //
+          // `scope === "imported"` is mutually exclusive with builtin
+          // and loop-var names at the source level, so we only need to
+          // check the import side.
           if (
             literal.scope === "imported" &&
-            !isBuiltinVar &&
-            !isLoopVar &&
             this.names.isAgencyImport(literal.value)
           ) {
             return ts.call(ts.id("__readStatic"), [
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -761,6 +761,30 @@ export class TypeScriptBuilder {
         const isBuiltinVar = BUILTIN_VARIABLES.includes(literal.value);
         const isLoopVar = this.loopVars.includes(literal.value);
         if (importedOrUnknownScope || isBuiltinVar || isLoopVar) {
+          // Agency imports may resolve to a `static const` in the source
+          // module, whose value is `__UNINIT_STATIC` until its initializer
+          // has run. Wrap reads in `__readStatic` so an early access
+          // throws a clear error instead of silently propagating the
+          // sentinel (which would surface as a confusing JS engine error
+          // like "Cannot convert a Symbol value to a string"). For
+          // non-static agency imports (functions, nodes) the wrap is a
+          // no-op because the binding never equals the sentinel. JS
+          // imports, builtins, and loop vars are not wrapped — they're
+          // never statics. Compiler-emitted boilerplate (e.g.
+          // `__registerTool(name)`) builds its source with `ts.raw`, not
+          // through this case, so it is unaffected.
+          if (
+            literal.scope === "imported" &&
+            !isBuiltinVar &&
+            !isLoopVar &&
+            this.names.isAgencyImport(literal.value)
+          ) {
+            return ts.call(ts.id("__readStatic"), [
+              ts.id(literal.value),
+              ts.str(literal.value),
+              ts.str(""),
+            ]);
+          }
           return ts.id(literal.value);
         }
         return ts.scopedVar(literal.value, literal.scope!, this.moduleId);
--- a/packages/agency-lang/lib/ir/prettyPrint.ts
+++ b/packages/agency-lang/lib/ir/prettyPrint.ts
@@ -282,6 +282,22 @@ export function printTs(node: TsNode, indent = 0): string {
         const ctxRef = node.topLevel ? "__globalCtx" : "getRuntimeContext().ctx";
         return `${ctxRef}.globals.get(${JSON.stringify(node.moduleId)}, ${JSON.stringify(node.name)})`;
       }
+      if (node.scope === "static") {
+        // Wrap static reads in `__readStatic` so that reading a static
+        // before its initializer has run throws a clear, actionable
+        // error (the `let` binding holds the sentinel
+        // `__UNINIT_STATIC` until then). The wrapper is transparent
+        // for initialized values — `__readStatic(x, ...)` returns `x`
+        // unchanged — so binary operations, template interpolations,
+        // indexing, etc. continue to work without any other change.
+        //
+        // `moduleId` may be undefined for static-scoped scopedVars
+        // produced in contexts where it wasn't propagated; fall back
+        // to the empty string in that case so the error message still
+        // points at the variable name.
+        const moduleIdLit = JSON.stringify(node.moduleId ?? "");
+        return `__readStatic(${node.name}, ${JSON.stringify(node.name)}, ${moduleIdLit})`;
+      }
       const prefix = scopeToPrefix(node.scope);
       if (prefix === "") return node.name;
       return `${prefix}.${node.name}`;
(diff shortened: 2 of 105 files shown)


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -761,6 +761,30 @@ export class TypeScriptBuilder {
...
+          // like "Cannot convert a Symbol value to a string"). For
+          // non-static agency imports (functions, nodes) the wrap is a
+          // no-op because the binding never equals the sentinel. JS
+          // imports, builtins, and loop vars are not wrapped — they're
+          // never statics. Compiler-emitted boilerplate (e.g.
+          // `__registerTool(name)`) builds its source with `ts.raw`, not
+          // through this case, so it is unaffected.
+          if (
+            literal.scope === "imported" &&
+            !isBuiltinVar &&
```
egonSchiele: if it is imported, how can it be a builtin var or a loop var?

packages/agency-lang/lib/ir/prettyPrint.ts
```diff
@@ -282,6 +282,22 @@ export function printTs(node: TsNode, indent = 0): string {
         const ctxRef = node.topLevel ? "__globalCtx" : "getRuntimeContext().ctx";
         return `${ctxRef}.globals.get(${JSON.stringify(node.moduleId)}, ${JSON.stringify(node.name)})`;
       }
+      if (node.scope === "static") {
```
egonSchiele, line 285: is this going to add a ton of boilerplate code around static vars?

packages/agency-lang/lib/runtime/staticInit.ts
```diff
@@ -0,0 +1,73 @@
...
+ * To turn that silent failure into a loud one:
+ *
+ *   1. Codegen initializes every `static const` binding to
+ *      `__UNINIT_STATIC` (a unique sentinel value).
+ *   2. Codegen wraps every *read* of a static name with
+ *      `__readStatic(value, name, moduleId)`.
+ *   3. `__readStatic` returns the value unchanged if it's not the
+ *      sentinel, or throws a clear error if it is.
+ *
+ * The wrapper is transparent — `__readStatic(x, ...)` evaluates to the
```
egonSchiele, line 21: are there places where a var name would be allowed but a func call wouldn't be?

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Good catch — they cant. `scope === "imported"` is mutually exclusive with builtin and loop-var names at the source level (builtins arent imported and loop vars dont have an `imported` scope), so the `!isBuiltinVar && !isLoopVar` guards were dead weight. Removed in the follow-up commit; the inner condition is now just `literal.scope === "imported" && this.names.isAgencyImport(literal.value)`.

packages/agency-lang/lib/ir/prettyPrint.ts
egonSchiele: Yes — every read of a `static const` name becomes `__readStatic(name, "name", "moduleId")`. It does inflate the generated source for code that touches statics a lot, but in practice static vars are a small fraction of identifiers, and the wrap is a single function call with a cheap pointer-equality check on the sentinel — no allocation, no branch in the hot path once initialized. The tradeoff is intentional: this is the safety net the rest of the agent-init series leans on, so we want it to be loud and uniform rather than rare and easy to bypass. If volume becomes a real problem we could later limit wraps to only the statics the per-variable dep graph (PR 2) flags as cycle participants, but until we have evidence of cost I'd rather keep the rule simple.

packages/agency-lang/lib/runtime/staticInit.ts
egonSchiele: Syntactically yes: a bare name can appear in positions where a function call cannot — assignment LHS (`x = 5`), `typeof x` / `delete x` on possibly-undeclared names, destructuring patterns (`const {x} = ...`), import specifiers, function/class declaration names, etc. The codegen however never emits a `static const` read in any of those positions — statics are always emitted as RHS expressions in a read context (binary ops, calls, template interpolations, indexing, property access). So in practice the wrap is a drop-in replacement for the identifier and the doc examples are accurate for every position the codegen actually produces.

egonSchiele:
Addressed review feedback in 20a9e747:

**egonSchiele:**
- `typescriptBuilder.ts` (redundant guards): Removed `!isBuiltinVar && !isLoopVar` — `scope === "imported"` is mutually exclusive with both at the source level.
- `prettyPrint.ts` (boilerplate volume): Kept as-is; reply on the thread explains the tradeoff (every static read becomes `__readStatic(...)`, but its a pointer-equality check, statics are a small fraction of identifiers, and uniformity matters for the safety net).
- `staticInit.ts` (var-vs-funccall positions): Reply on the thread — statics are only ever emitted as RHS expressions in read context, so the wrap is a drop-in replacement in every position the codegen actually produces.

**Copilot:**
- `staticInit.ts` empty `moduleId`: Now substitutes `<unknown module>` so the error message stays readable until PR 2 threads source module ids through every wrap site.
- `staticInit.crossModule.test.ts`: Rewrote to use in-process `compile()` + `pathToFileURL`, drops dependence on `dist/scripts/agency.js`, calls `resetCompilationCache()`, and writes fixtures under `.agency-tmp/` (gitignored) instead of shelling out.

PR #245, opened 2026-06-01
PR 2 of 6 — Per-variable topological sort + centralized init

egonSchiele:
## PR 2 of 6 — Per-variable topological sort + centralized init

Implements the second PR from [agent-init-design.md](../blob/main/packages/agency-lang/agent-init-design.md), following the plan in [docs/superpowers/plans/2026-05-31-agent-init-pr2-per-var-topsort.md](../blob/agent-init-pr2-per-var-topsort/packages/agency-lang/docs/superpowers/plans/2026-05-31-agent-init-pr2-per-var-topsort.md).

Built on top of PR 1's runtime read-before-init trap.

### What changes

Replaces per-module lazy `__initializeStatic` / `__initializeGlobals` with a centralized, topologically-sorted init across the full import closure. The walk now happens at the entry point (`compileClosure`), once, with two independent per-variable dep graphs:

- **Phase A — static graph**, run once per process, ordered by topsort.
- **Phase B — global graph**, run on every invocation, ordered by topsort.

Both graphs share an `ImportAliasResolver` that also feeds PR 1's `__readStatic` trap message with the real source `moduleId`.

Cycles in either graph become compile-time errors that name the offending decls + their source locations. A `static` initializer that references a `global` is a separate compile-time error since the global doesn't yet exist at Phase A time.

### Key pieces

- `lib/compiler/initDepGraph.ts` — two-phase per-variable dep graph builder. Uses `walkNodes` + `SymbolTable.resolveImport`; closure walking is owned by the caller. One-hop alias resolution so re-export wrappers naturally cascade.
- `lib/compiler/topSortInitGraph.ts` — Kahn topsort keyed on each node's precomputed `sequenceHint` (file-import depth × 1e6 + source line). Cycle errors report a full traced loop.
- `lib/compiler/compileClosure.ts` — single owner of multi-file compilation. Parses every reachable file once via BFS, runs `resolveReExports` on each, builds both graphs + plans, and exposes the resolver for downstream codegen.
- `lib/backends/typescriptBuilder/sectionAssembler.ts` — emits the `__awaitStaticInit(...)` / `__awaitGlobalsInit(...)` prelude based on the per-module plan.
- `lib/backends/typescriptBuilder.ts` — threads the resolver into `__readStatic` so trap messages show the real source module instead of `<unknown>`.
- `lib/runtime/initRegistry.ts` — process-global registry of per-module init functions. Modules register on JS-load; closure-derived awaits look them up at agent-run time.
- `lib/runtime/initPlanWorkedExamples.test.ts` — end-to-end integration tests covering the eight worked examples from the design doc plus re-export chains and multi-entry-point.

### Notable subtleties

- **Global → static cross-phase reads.** Statics are fully initialized before any global runs, but the global graph drops static refs as edges (cycle detection lives per-phase). `globalPhasePlanFor` walks each local global's free identifiers and adds modules of any cross-module statics it reads to the plan's `awaitModules` — without this a `const g = importedStatic` pattern hit the PR-1 trap.
- **Re-export chains.** `resolveReExports` synthesizes wrapper statics in each intermediate re-exporter that need their own `__initializeStatic`. Running `resolveReExports` inside `parseClosure` puts the wrappers in front of the dep graph; one-hop resolution then lets the topsort cascade through the entire chain (a → b → c) automatically.
- **Closure walking and `resolveReExports` order.** The walker needs the raw `exportFromStatement` nodes that `resolveReExports` strips, so the BFS runs over raw parse output, and the expand-with-wrappers pass runs once the closure is fully discovered.

### Tests

- 4748 / 4749 `lib/` tests pass. The remaining failure is the pre-existing `stdlib: 'ui'` typecheck called out as out of scope in the plan.
- New tests: 8 worked-example integration cases, re-export chain end-to-end, compileClosure cycle / static-references-global / re-export plan tests, dep graph re-export cascade test, topsort 50-module perf smoke (< 100ms).

### What's next

- PR 3: `static` keyword for bare top-level statements.
- PR 4: compile-time validations on what's allowed inside static initializers.
- PR 5: `agency explain-init` CLI.
- PR 6: out-of-scope cleanup items collected during PR 2 review.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder/sectionAssembler.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder/sectionAssembler.ts
@@ -200,10 +200,13 @@ export function partitionProgram(
 
 /**
  * Apply the topsort plan's `localOrder` to tagged init statements.
- * Named statements (statics, globals) are pulled out and re-emitted in
- * plan order; unnamed bare statements are interleaved at the position
- * they originally appeared, preserving source-order semantics for
- * side-effecty calls that have no var to depend on.
+ *
+ * Bare (unnamed) statements are anchored to their source position —
+ * each occupies exactly the slot where it originally appeared. Named
+ * statements are reordered per the plan, with the k-th name in
+ * `localOrder` filling the k-th named slot in source order. This
+ * preserves side-effect ordering for patterns like `foo(); const x =
+ * ...; bar();` while still letting topsort sequence the named decls.
  *
  * No plan → source order, unchanged.
  */
@@ -218,15 +221,20 @@ function reorderTagged(
   for (const { varName, node } of tagged) {
     if (varName) byName[varName] = node;
   }
+  // Walk tagged once. Bare slots emit their own node immediately.
+  // Named slots are placeholders to be filled from `order` in
+  // top-to-bottom order — only names that have a matching tagged
+  // node count, so cross-module-only names in `order` are skipped.
+  const planNamesInTagged = order.filter((n) => byName[n]);
   const out: TsNode[] = [];
-  // Named statements first, in plan order.
-  for (const name of order) {
-    if (byName[name]) out.push(byName[name]);
-  }
-  // Then bare statements (no varName) in their original source order
-  // so any side effect ordering relative to other bares is preserved.
+  let namedSlotIdx = 0;
   for (const { varName, node } of tagged) {
-    if (!varName) out.push(node);
+    if (varName) {
+      const fillName = planNamesInTagged[namedSlotIdx++];
+      if (fillName) out.push(byName[fillName]!);
+    } else {
+      out.push(node);
+    }
   }
   return out;
 }
--- a/packages/agency-lang/lib/backends/typescriptBuilder/sectionAssembler.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder/sectionAssembler.ts
@@ -1,6 +1,24 @@
+import * as path from "path";
 import type { AgencyNode, AgencyProgram } from "../../types.js";
 import type { TsNode } from "../../ir/tsIR.js";
 import { ts } from "../../ir/builders.js";
+
+/**
+ * Convert an absolute module path to one rooted at the codegen cwd
+ * for readability in the generated registry-key string literals.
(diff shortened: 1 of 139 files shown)


packages/agency-lang/lib/backends/typescriptBuilder/sectionAssembler.ts
```diff
@@ -161,20 +181,56 @@ export function partitionProgram(
...
+/**
+ * Apply the topsort plan's `localOrder` to tagged init statements.
+ * Named statements (statics, globals) are pulled out and re-emitted in
+ * plan order; unnamed bare statements are interleaved at the position
+ * they originally appeared, preserving source-order semantics for
+ * side-effecty calls that have no var to depend on.
+ *
+ * No plan → source order, unchanged.
+ */
+function reorderTagged(
```
egonSchiele, line 231: I assumed topsort would be for ordering the init of imported vars, I didn't think it would also be for ordering the init of vars in the file. It makes sense, but if the user has ordered statements in the current file in a specific order, changing that ordering is surprising. I'd rather print an error saying "X is being used before it is defined" or similar.

packages/agency-lang/lib/compiler/compileClosure.ts
```diff
@@ -0,0 +1,406 @@
...
+    const target = agencyImportTarget(node);
+    if (!target) continue;
+    if (isStdlibImport(target) || isPkgImport(target)) continue;
+    if (!isAgencyImport(target)) continue;
+    out.push(resolveAgencyImportPath(target, moduleId));
+  }
+  return out;
+}
+
+function agencyImportTarget(node: AgencyNode): string | null {
```
egonSchiele, line 227: IMO there should be a reusable func that parses a file and gives you the symbol table so you can get the imports, instead of rewriting this logic here

packages/agency-lang/lib/runtime/initRegistry.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * Process-global registry of per-module init functions, used by
```
egonSchiele, line 2: we have a lot of things called "registry" in the code ... how do we differentiate between them all?

PR #246, opened 2026-06-01
Agent-init PR 2 stragglers: rename, cycle tracing, loadModule, use-before-def, doc fix

egonSchiele:
Closes out the four review-cycle items raised on PR #245 that did not
block merge, plus the one-line doc fix flagged after PR 2 landed.

## What is in this PR

Four loosely-coupled cleanups + a doc fix, one per commit so the
design-change commit can be reverted independently:

1. **Rename `initRegistry` → `crossModuleInitRegistry`** (egon review
   item). The name collided in conversation with each module's own
   `__initializeStatic` / `__initializeGlobals` closures. Pure file
   rename; the one consumer (`lib/runtime/index.ts`) updated. No
   symbols change names, no template changes needed (the templates
   import from `agency-lang/runtime`, not by file path), no fixture
   diffs.

2. **Cycle tracing uses post-Kahn `remaining`** (Copilot review item).
   `kahn` now returns `{ order, remaining }`. `traceCycleFrom` walks
   the post-Kahn `remaining` map for both the starter pick and the
   next-step selection, so the trace never wanders out of the cycle
   into already-drained DAG portions. Added a unit test
   (`c1 → d1 → d2` tail + `c1 ↔ c2` cycle) that returns
   `[c1, d1, d2]` as a "cycle" pre-fix and `[c1, c2]` post-fix.

3. **Extract `loadModule` helper from `parseClosure`** (egon review
   item). Module-private helper that reads + parses one `.agency`
   file and returns `{ program, importTargets }`. PR 5
   (`agency explain-init`) is the planned second caller; it stays
   private until that caller drives the signature. Pure refactor —
   no behavior change.

4. **Intra-file use-before-def is now a compile error** (egon review
   item — design change confirmed during PR 245 review). Before this
   change, the section assembler silently reordered same-file named
   decls when topsort disagreed with source order; after this change,
   the same fixture errors with both source lines and a "reorder so
   X appears before Y" suggestion. Cross-module reorder is still
   allowed (no canonical "source order" across files). Re-export
   wrappers and bare statements are excluded. Driven by two new
   fixtures under `tests/agency/topsort/use-before-def/` and two
   cases in the existing `lib/runtime/topsortCycleErrors.test.ts`.

5. **Doc fix in `init-topsort.md`** — drops the "Known gap: non-static
   export const" subsection. In Agency, globals cannot be exported
   at all (only statics can), so the resolver coverage is complete
   for the supported export surface; the section was misleading.

## Validation

- `pnpm test:run lib/compiler/ lib/runtime/ lib/backends/` — 1095/1095 pass
- `pnpm run agency test tests/agency/topsort/` — 6/6 pass (stale `.js` cleaned first)
- `make fixtures` — no fixture diffs

The pre-existing stdlib `ui` typecheck failure under
`lib/typeChecker/fixtureTypeCheck.integration.test.ts` is unrelated
to this PR and known-failing on `main`.

## Plan + traceability

- Plan: `docs/superpowers/plans/2026-06-01-agent-init-pr2-stragglers.md`
- Roadmap context: `agent-init-roadmap.md`
- Predecessor: PR #245 (per-variable topsort + centralized init)


packages/agency-lang/tests/agency/topsort/use-before-def/README.md
```diff
@@ -0,0 +1,14 @@
+# Intra-file use-before-def fixtures
+
+Each subdirectory is an Agency program that should fail compilation with
+a use-before-def error because a same-file decl references another
+same-file decl that appears later in source order.
+
+The Agency test framework (`agency test`) cannot express "expected
```
egonSchiele: Ooo this is really ugly. If a file isn't run as an agency test, it does NOT belong in tests/agency/.

egonSchiele:
Addressed all three review comments in dfc07e7f:

- **egon — fixtures don't belong under `tests/agency/`:** moved the use-before-def fixtures inline as strings into `lib/runtime/topsortCycleErrors.test.ts`, materialized at run-time into `.agency-tmp/use-before-def/<case>/` (gitignored), and wiped per-test in `afterEach`. Matches the pattern already in `lib/runtime/staticInit.crossModule.test.ts`. The whole `tests/agency/topsort/use-before-def/` directory is gone.
- **Copilot — `_reexport_` prefix isn't reserved:** replaced the shape check in `isReExportWrapper` with a `SymbolTable.getFile(...).reExportedFrom` lookup. SymbolTable.build sets `reExportedFrom` authoritatively for symbols that came in via `export ... from "..."`, so it identifies exactly the wrappers `resolveReExports` later synthesizes — no overlap with user code.
- **Copilot — broken relative link in README:** resolved by the README being deleted (point 1).

Validation: `pnpm test:run lib/compiler/ lib/runtime/ lib/backends/` still 1095/1095. `pnpm run agency test tests/agency/topsort/` still 6/6 (the use-before-def cases are now vitest-only).

PR #247, opened 2026-06-02
PR 2.5: depth-1 function-body analysis for init dep edges

egonSchiele:
## Summary

Implements [PR 2.5 from the agent-init roadmap](docs/superpowers/plans/2026-06-01-agent-init-pr2.5-depth1-callgraph.md): extend the per-variable init dep graph so that when an initializer calls a top-level Agency function, the graph walks that function's body **one level deep** for free identifiers and adds the resolutions as additional edges.

This catches the 90% case where an initializer reads a value through a single function call:

```
import { getBar } from "./bar.agency"
static const foo = getBar()  // bar.barStatic now counts as a dep edge
```

The boundary stops at one function-body hop. Depth-2+ chains, function values stored in variables, methods on user objects, and stdlib calls all remain covered by the PR 1 runtime read-before-init trap.

## Why depth-1 and not transitive

The boundary stays predictable: users can answer "will the compiler catch my case?" by asking "is this a single direct function call?" Transitive analysis introduces a fuzzy boundary that grows over time; the runtime trap handles the long tail with a clear error message either way.

## What this fixes

1. `static const foo = getBarStatic()` where `getBarStatic` reads a static in another module. Previously fired the runtime trap on first call; now compiles into a plan that awaits the source module's static init before `foo`.
2. `static const foo = getBarGlobal()` where `getBarGlobal` reads a global. Previously compiled cleanly even though statics can't read globals — now `rejectStaticReferencesGlobal` catches it at compile time.

## Mechanism

- `FunctionDefLookup` (`lib/compiler/initDepGraph.ts`) resolves bare or namespace-prefixed call names to their `(moduleId, FunctionDefinition)` source.
- `collectDirectCalls` walks an init expression for call sites (`functionCall` nodes + `valueAccess.methodCall` chain entries). `collectFunctionBodyFreeRefs` walks each matched function's body via the existing `collectFreeIdentifiers`, which already skips nested `function` / `graphNode` bodies — that gives the depth-1 boundary for free.
- Expansion is applied identically in `depsFor`, `rejectStaticReferencesGlobal`, and `globalPhasePlanFor` so static edges, the static-references-global compile error, and cross-phase `awaitModules` all stay in lockstep.

## Tests

- 9 new unit cases in `lib/compiler/initDepGraph.test.ts` cover direct expansion, namespace expansion, cross-phase rejection, depth-2 cap, function-value not chased, stdlib silent skip, self-recursion, and block bodies inside the called function.
- 2 new vitest cases in `lib/runtime/topsortCycleErrors.test.ts` cover compile-error and no-trap cases via inline fixtures.
- 2 new agency end-to-end fixtures under `tests/agency/topsort/depth1-callgraph/` (static-via-function, bare-stmt-via-function).
- The existing `runtime-trap` fixture is reworked to use a 2-hop chain (`a -> outerReadB -> innerReadB -> b`) so the runtime trap still has something to catch.

## Verification

- `pnpm test:run lib/compiler/ lib/runtime/ lib/backends/` → 1106/1106 pass
- `pnpm run agency test tests/agency/topsort/` → 8/8 pass (includes the two new depth-1 fixtures)
- `pnpm run lint:structure` → clean
- `make fixtures` → no diffs beyond the intentional `runtime-trap` rework

## Docs

- `docs/site/guide/imports-and-packages.md` — new user-facing "Initializer dependencies through function calls" section with examples + boundary list + safety-net note.
- `docs/dev/init-topsort.md` — new "Depth-1 function-body expansion" subsection, updated safety-net wording for the new boundary, file-map note for `FunctionDefLookup`.

## Plan + design references

- Plan: `docs/superpowers/plans/2026-06-01-agent-init-pr2.5-depth1-callgraph.md`
- Roadmap: `agent-init-roadmap.md`
- Builds on PR 1 (read-before-init trap) and PR 2 (per-variable topsort).

### DIFF
--- a/packages/agency-lang/tests/agency/topsort/depth1-callgraph/static-via-function/main.agency
+++ b/packages/agency-lang/tests/agency/topsort/depth1-callgraph/static-via-function/main.agency
@@ -1,0 +1,16 @@
+// PR-2.5 depth-1 expansion: `fooStatic`'s initializer calls
+// `getBarStatic` in `helper.agency`, whose body reads `barStatic`.
+// Without depth-1 expansion, the static dep graph would have no edge
+// from `fooStatic` to `barStatic`, helper.__initializeStatic would not
+// run before main's, and the PR-1 runtime trap would fire when
+// `getBarStatic` accessed `barStatic` during main's static init.
+//
+// With depth-1, the dep graph sees the cross-module read, schedules
+// helper.barStatic first, and the assertion below passes.
+import { getBarStatic } from "./helper.agency"
+
+static const fooStatic = getBarStatic() + "!"
+
+node main() {
+  return fooStatic
+}
--- a/packages/agency-lang/docs/dev/init-topsort.md
+++ b/packages/agency-lang/docs/dev/init-topsort.md
@@ -100,10 +100,55 @@ references**. The edge set is derived by walking the initializer
 expression for free identifiers and resolving each one through the
 shared `ImportAliasResolver`.
 
-**Functions are never edges.** Only values participate. A static
-initializer that calls a function defined elsewhere produces no edge
-to that function's locals; the runtime trap (PR 1) is the safety
-net for unset statics read indirectly.
+**Functions are never edges themselves.** Only values are nodes in
+the graph. A bare reference to a function name never produces an
+edge to a function node — function defs aren't tracked.
+
+### Depth-1 function-body expansion
+
+When a free identifier in an init expression is a **direct call** of
+a top-level Agency function — bare (`getBar()`) or namespace-prefixed
+(`bar.getBar()`) — the dep graph walks that function's body once and
+treats every top-level value the body reads as an additional
+dependency of the enclosing init node. Inner references are resolved
+in the function's home module, not the caller's, so an import the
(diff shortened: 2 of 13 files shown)


packages/agency-lang/tests/agency/topsort/depth1-callgraph/static-via-function/main.agency
```diff
@@ -0,0 +1,16 @@
...
+// Without depth-1 expansion, the static dep graph would have no edge
+// from `fooStatic` to `barStatic`, helper.__initializeStatic would not
+// run before main's, and the PR-1 runtime trap would fire when
+// `getBarStatic` accessed `barStatic` during main's static init.
+//
+// With depth-1, the dep graph sees the cross-module read, schedules
+// helper.barStatic first, and the assertion below passes.
+import { getBarStatic } from "./helper.agency"
+
+static const fooStatic = getBarStatic() + "!"
```
egonSchiele, line 12: For some of these tests, if they failed, would they fail consistently or only intermittently? I am worried that some of these tests will be flaky. I am not sure what to do about that though.

PR #255, opened 2026-06-03
Add std::layout module for declarative text layout

egonSchiele:
## Summary

Adds `std::layout`, a declarative module for composing terminal text
layouts — bordered panels, multi-column splash screens, summary
tables. Two sharply separated layers:

- **Agency-side** (`stdlib/layout.agency`): pure data. Builder /
  constructor calls produce an immutable `LayoutNode` tree.
- **TS-side** (`lib/stdlib/layout.ts`): pure render. Consumes a tree,
  returns an ANSI-styled multi-line string.

Construction works two ways from the same surface:

```ts
// Trailing-block style — ergonomic for Agency authors
const panel = box(title: "Hello", padding: 1) as b {
  b.text("Welcome!", bold: true)
  b.text("How are you?")
}

// Children-array style — suitable for LLM tool calls / JSON
const panel = box(
  title: "Hello",
  padding: 1,
  children: [
    text("Welcome!", bold: true),
    text("How are you?"),
  ],
)

print(render(panel))
```

Both styles produce structurally identical trees and rendered
output (verified by `children-style.agency` against `banner.agency`).

### What's in the box

- `Block` primitive + five free operators (`beside`, `above`, `pad`,
  `styled`, `bordered`). All ANSI awareness funnelled through a
  single `visualWidth` helper.
- Border styles: `rounded` (default), `heavy`, `double`, `light`,
  with embedded titles that auto-grow the box (no truncation) and a
  separate `titleColor` from `borderColor`.
- Node renderers: `text`, `raw`, `space`, `hline`, `vline`, `row`,
  `column`, `box`.
- Stretchy lines: `vline()` inside a row spans the row's height,
  `hline()` inside a column spans the column's width.
- `space(n)` + container `gap` are additive on the same axis.
- `render(node, color: "auto" | true | false)` — `auto` checks
  `process.stdout.isTTY`; `false` strips all SGR for plain-ASCII
  consumers; `true` always emits them.
- Compile-time rejection of unknown `borderStyle` via Agency string
  literal union types.

`std::ui` and `std::layout` both export `box`, `row`, `column` —
each module's docstring points at the other and recommends aliasing
on import if you need both.

### Drive-by parser fix

This branch also includes a small fix to the inline-block argument
parser so `column() as col { ... }` binds the trailing block to a
function's `block:` parameter (matched by name + function type)
rather than to the first positional param. The matching `block-skips-
optional-params` test fixture was added.

## Test Plan

- [x] `pnpm test:run lib/stdlib/layout.test.ts` — 78 tests pass
- [x] `pnpm run agency test tests/agency/layout/` — 18 tests pass
  across 5 files (builders, nested, banner, children-style,
  color-modes)
- [x] `make stdlib` compiles `stdlib/layout.agency` clean
- [x] `make doc` regenerates `docs/site/stdlib/layout.md`
- [x] `pnpm run agency examples/layoutDemo.agency` renders four
  sample panels without visual glitches
- [ ] CI: full agency test suite


packages/agency-lang/examples/layoutDemo.agency
```diff
@@ -0,0 +1,58 @@
...
+        right.text("Ctrl-C — interrupt")
+        right.text("Ctrl-D — submit")
+      }
+    }
+  }
+  print(render(twoCol))
+  print("")
+
+  // 3) Column with an hline separator and a heavy-bordered footer.
+  const sections = column(gap: 0) as c {
```
egonSchiele, line 33: where is the heavy-bordered footer?

packages/agency-lang/tests/typescriptGenerator/inlineBlockParams.mjs
```diff
@@ -347,8 +347,10 @@ __stack.locals.items = [1, 2, 3];
       });
       await runner.step(2, async (runner) => {
 __stack.locals.doubled = await __call(mapItems, {
-          type: "positional",
-          args: [__stack.locals.items, __AgencyFunction.create({ name: "__block_0", module: "inlineBlockParams.agency", fn: async (x: any) => {
+          type: "named",
```
egonSchiele, line 350: Is this change correct? What is going from a positional to a named arg?

PR #258, opened 2026-06-04
Support standard SKILL.md skill format in stdlib/skills

egonSchiele:
## Summary

- `stdlib/skills.agency`: `skillsDir(dir, layout: "flat" | "standard")` now supports the [Claude Code / agentskills.io](https://code.claude.com/docs/en/skills.md) standard format — each subdirectory under `dir` is one skill with a `SKILL.md` entry point and YAML `name` + `description` frontmatter. The legacy flat-markdown behavior (one `.md` per skill) is preserved as `"flat"` and remains the layout the bundled agency-agent docs use.
- `agency doc` emits `title:` + `name:` YAML frontmatter on every generated stdlib markdown file, and the handwritten docs under `docs/site/guide/` and `docs/site/appendix/` get a matching `name:` field. VitePress continues to read `title`; downstream skill tooling reads `name`.
- `AgencyFunction` gets a `get description()` getter — symmetric counterpart to the existing `.describe(str)` setter. Reading `tool.description` from Agency code now returns the description string instead of `undefined`.
- All four internal `skillsDir(...)` call sites in `lib/agents/agency-agent/` pass the explicit `"flat"` layout, matching what their on-disk directories actually contain.

## Test plan

- [x] `pnpm test:run lib/cli/doc.test.ts` — 17/17 pass (3 new tests assert the frontmatter shape)
- [x] `pnpm run agency test tests/agency/skills-dir.test.json` — 8/8 pass (new fixture-backed agency execution test covers both layouts, frontmatter fallbacks, XML escaping, and the empty-directory case)
- [x] `make` runs clean end-to-end
- [ ] Verify the agency-agent's bundled `docSkill` / `cliSkill` / `appendixSkill` / `superpowersSkill` still list real filenames when the agent runs interactively
- [ ] CI green

### DIFF
--- a/packages/agency-lang/lib/cli/doc.ts
+++ b/packages/agency-lang/lib/cli/doc.ts
@@ -162,7 +162,11 @@ function generateDocForFile(
   const constants = collectExportedConstants(program);
 
   const title = path.basename(filePath).replace(/\.agency$/, "");
-  const sections: string[] = [heading(1, title)];
+  if (/["\\\n]/.test(title)) {
+    throw new Error(`Cannot generate doc for ${filePath}: title ${JSON.stringify(title)} contains characters unsafe for YAML/Markdown.`);
+  }
+  const frontmatter = `---\ntitle: "${title}"\nname: "${title}"\n---`;
+  const sections: string[] = [frontmatter, heading(1, title)];
 
   // Page-level "View source" link
   if (ctx.baseUrl && ctx.sourceRelPath) {
--- a/packages/agency-lang/lib/cli/doc.ts
+++ b/packages/agency-lang/lib/cli/doc.ts
@@ -162,10 +162,8 @@ function generateDocForFile(
   const constants = collectExportedConstants(program);
 
   const title = path.basename(filePath).replace(/\.agency$/, "");
-  if (/["\\\n]/.test(title)) {
-    throw new Error(`Cannot generate doc for ${filePath}: title ${JSON.stringify(title)} contains characters unsafe for YAML/Markdown.`);
-  }
-  const frontmatter = `---\ntitle: "${title}"\nname: "${title}"\n---`;
+  const safeName = title.replace(/["\\\n]/g, "");
+  const frontmatter = `---\nname: "${safeName}"\n---`;
   const sections: string[] = [frontmatter, heading(1, title)];
 
   // Page-level "View source" link
--- a/packages/agency-lang/stdlib/skills.agency
+++ b/packages/agency-lang/stdlib/skills.agency
@@ -26,50 +26,55 @@ def xmlEscape(s: string): string {
 }
 
 def renderEntry(entry: SkillEntry): string {
+  """
+  Render one SkillEntry as a `<skill>` XML block. All string fields are
+  XML-escaped because they originate from user-authored frontmatter or
+  filenames, and an unescaped `<` or `&` would corrupt the surrounding
+  `<available_skills>` markup.
+  """
   const nameX = xmlEscape(entry.name)
   const descX = xmlEscape(entry.description)
   const locX = xmlEscape(entry.location)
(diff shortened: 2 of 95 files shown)


packages/agency-lang/stdlib/skills.agency
```diff
@@ -27,63 +25,80 @@ def xmlEscape(s: string): string {
   return out
 }
 
-def describeSkill(filename: string, fm: Result): string {
+def renderEntry(entry: SkillEntry): string {
+  const nameX = xmlEscape(entry.name)
+  const descX = xmlEscape(entry.description)
+  const locX = xmlEscape(entry.location)
+  return "  <skill>\n    <name>${nameX}</name>\n    <description>${descX}</description>\n    <location>${locX}</location>\n  </skill>"
```
egonSchiele: use triple-quote string, which allows newlines in it

packages/agency-lang/stdlib/skills.agency
```diff
@@ -27,63 +25,80 @@ def xmlEscape(s: string): string {
...
+}
+
+def flatEntry(filename: string, fm: Result): SkillEntry {
```
egonSchiele: what is `fm`? These names need to be descriptive to an llm understands what they mean

egonSchiele: also, `Result` makes no sense as a param type. Catch the failure at the call site and pass in empty front matter or something

packages/agency-lang/stdlib/skills.agency
```diff
@@ -27,63 +25,80 @@ def xmlEscape(s: string): string {
...
+}
+
+def standardEntry(skillPath: string, fm: Result): SkillEntry {
+  """
+  Build a SkillEntry for one skill in the standard SKILL.md layout.
+  `skillPath` is "<skill-dir>/SKILL.md" relative to the skills root;
+  `name` defaults to the skill's directory name when not in frontmatter.
+  """
+  const skillDir = skillPath.replace("/SKILL.md", "")
+  if (isFailure(fm)) {
```
egonSchiele: use idiomatic pattern matching https://agency-lang.com/guide/pattern-matching.html

packages/agency-lang/stdlib/skills.agency
```diff
@@ -27,63 +25,80 @@ def xmlEscape(s: string): string {
...
   if (pathsResult is success(paths)) {
-    lines = map(paths) as filename {
-      // `glob` now returns paths relative to `dir`, so `filename` is
-      // ready to hand straight to `read(filename, dir)` — no basename
-      // step needed.
-      const fm = read(filename, dir) |> frontmatter
-      return describeSkill(filename, fm)
+    lines = map(paths) as path {
+      const fm = read(path, dir) |> frontmatter
+      let entry = flatEntry(path, fm)
```
egonSchiele: could use `catch` to unwrap or use a default val

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -162,7 +162,11 @@ function generateDocForFile(
   const constants = collectExportedConstants(program);
 
   const title = path.basename(filePath).replace(/\.agency$/, "");
-  const sections: string[] = [heading(1, title)];
+  if (/["\\\n]/.test(title)) {
+    throw new Error(`Cannot generate doc for ${filePath}: title ${JSON.stringify(title)} contains characters unsafe for YAML/Markdown.`);
```
egonSchiele: strip the chars then, don't throw an error.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -162,7 +162,11 @@ function generateDocForFile(
...
+  }
+  const frontmatter = `---\ntitle: "${title}"\nname: "${title}"\n---`;
```
egonSchiele: actually, just emit `name`. Emitting both is ugly. How could users set a description? Maybe a `@skillDescription` doc comment, like `@module`?

PR #262, opened 2026-06-05
Per-branch state isolation for fork/parallel/race (with shared opt-out)

egonSchiele:
## What

Each branch of `fork`, `parallel`, and `race` now gets its own snapshot of the parent's globals and a fresh active-thread pointer by default. Writes inside a branch are isolated from siblings and the parent; only branch return values cross the join boundary. This matches the same isolation property that runs already have.

Users can opt back into pointer-sharing the parent's state at the call site with `shared: true`:

```ts
parallel(shared: true) { workerA(); workerB() }
fork(items, shared: true) as item { ... }
race(items, shared: true) as item { ... }
```

## Why

Previously, branches pointer-shared the parent's globals. That meant wrapping two existing agents in `parallel` could silently corrupt their internal state if they happened to use the same global names. The new default makes parallel-composition safe by default and reserves cross-branch sharing as an explicit opt-in.

The fork-loop pitfall that users coming from JS would reach for is now caught by the default:

```ts
let count = 0
for (i in [0, 1, 2]) {
  fork([1]) as _ { count = count + 1 }  // writes to branch-local copy
}
print(count)   // 0 — branch writes were discarded
```

The idiomatic replacement is "return the contribution and reduce":

```ts
const contributions = fork([0, 1, 2]) as i { return 1 }
const count = sum(contributions)
```

## How (stages)

1. **Stage 1 — ALS-frame globals access.** Generated user code reads/writes globals through `__globals()` (an AsyncLocalStorage accessor) instead of `__ctx.globals` directly. `__initializeGlobals` and tool-description docstrings keep the old direct-context path (they have nuanced bootstrap requirements).

2. **Stage 2 — per-branch isolation.** `runInBranchAlsFrame` (in `lib/runtime/runBatch.ts`) installs a per-branch ALS frame seeded with `parent.globals.clone()` and `parent.threads.forkBranchView()`. Branches read and write their own copies; the parent and siblings are invisible.

3. **Stage 3 — interrupt resume.** Per-branch state is captured to `BranchState.globalsJSON` and `BranchState.activeStack` on every body settle (including settling-as-`Interrupt[]`). On resume, `runInBranchAlsFrame` restores from those snapshots so the body sees its pre-interrupt state. Both fields are wired through `StateStack.toJSON` / `fromJSON`.

4. **`shared: true` syntax.** Parser (`parallelBlockParser`, `_parallelNamedArgsParser`), typechecker (`acceptsNamedArgs` on `BuiltinSignature`), preprocessor (`parallelDesugar`), and codegen (`typescriptBuilder.processForkCall`) all plumb a new optional `shared` named arg through to the `runBatch` call site as `isolateState: false`.

5. **`runPrompt` tool dispatch.** `runPrompt`'s parallel tool-call loop is forced to `isolateState: false` because tool calls are conceptually sequential function invocations and any global state they touch (counters, retry budgets, dedup caches) should accumulate across calls. Same effect as `shared: true`.

6. **Stage 4 — docs.** Updated `docs/site/guide/concurrency.md`, `execution-model.md`, `global-vs-static.md`, `cross-thread-context.md`, `docs/dev/globalstore.md`, `docs/dev/runBatch.md`.

## Tests

Regression tests added under `tests/agency/fork/` and `tests/agency/parallel/`:

- isolation: branch writes do not leak to parent or siblings
- shared: opt-in pointer-sharing works for cooperative-worker patterns
- interrupt/resume survival: per-branch globals + active-thread pointer round-trip through `BranchState` serialization across interrupts

## Notes

- The message-thread active pointer is **always** branch-local regardless of `shared` — concurrent branches pushing/popping the same active thread would corrupt the conversation. The thread *registry* is always shared, so `thread(session: ...)` and `thread(continue: id)` keep working under both modes.
- Removed dead-code `StateStack.branchToJSON`.


packages/agency-lang/lib/runtime/state/threadStore.ts
```diff
@@ -14,29 +14,169 @@ export type ThreadStoreJSON = {
...
+    if (parentActiveId !== undefined) {
+      // Inline the subthread-creation steps from `createSubthread` so
+      // we write the new id onto the *branch's* activeStack instead
+      // of the parent's. Goes through the shared registry, so the
+      // subthread is visible to `listThreads()` from any branch and
+      // to the parent after join.
+      const id = (this.registry.counter++).toString();
+      const parentThread = this.registry.threads[parentActiveId];
+      const child = parentThread.newSubthreadChild();
+      child.parentId = parentActiveId;
```
egonSchiele: shouldn't the `newSubthreadChild` func set this id?

packages/agency-lang/lib/runtime/state/threadStore.ts
```diff
@@ -14,29 +14,169 @@ export type ThreadStoreJSON = {
...
+      this.registry.threads[id] = child;
```
egonSchiele: I am surprised that we need to write all this code for thread creation here. Is this code not needed when we create threads elsewhere? Since we are using existing functions to create the thread, I would assume that any other thread creation related functionality would already be in those existing functions.

packages/agency-lang/tests/agency/fork/fork-globals-isolated.agency
```diff
@@ -0,0 +1,17 @@
+global let counter = 0
```
egonSchiele: I'm not sure agency has the `global` keyword? All vars in global state are global (unless marked static)

PR #266, opened 2026-06-05
Add slash commands for the agency agent

egonSchiele:
## What

Adds `commandsDir(dir)` to `stdlib/skills.agency` and wires it into the agency agent. Users can drop Claude Code-format `.claude/commands/foo.md` files into a project (or `~/.claude/commands/`) and invoke them as `/foo` in the agent — both in the interactive REPL and via the one-shot `--print` / piped mode.

CC merged commands and skills into one file format; we already had `skillsDir` for the model-invoked side, this PR adds the user-typed side. Both now live in `std::skills`.

## Why

The agent currently hard-codes `/exit`, `/clear`, `/help` and offers nothing extensible. Users with existing CC command files have to retype their workflows on every turn.

## What v1 deliberately does NOT do

v1 is a pure prompt template: the rendered body becomes a user message, exactly as if the user had typed the text. The following are explicit non-goals, tracked as follow-ups in the [spec](packages/agency-lang/docs/superpowers/plans/2026-06-04-slash-commands.md):

- `` !`cmd` `` shell injection
- `@<path>` file references
- `allowed-tools` pre-approval
- `model:` / `effort:` / `context: fork` overrides
- Live reload on file changes
- Configurable arg delimiters (cf. anthropics/claude-code#8406)

This keeps the security model trivial: commands cannot do anything the user couldn't do by typing. When shell injection is added later, it will route through `cliPolicyHandler` (Agency's universal handler boundary), not through static frontmatter pre-approval.

## API

```agency
type Command = {
  name: string;          // "foo" or "ns:bar" (no leading slash)
  description: string;
  argHint: string;
  body: string
}

type CommandSet = {
  entries: Command[];
  dispatch: any          // (string) => Result<string>
}

export def commandsDir(dir: string): CommandSet
```

`dispatch(input)`:
- `success(renderedBody)` on `/name` match (with optional whitespace + args)
- `failure("no-match")` otherwise

## Body preprocessing (matches CC)

- `$ARGUMENTS` → raw arg string (quotes preserved)
- `$ARGUMENTS[N]` and `$N` → 1-indexed tokenized arg
- Tokenizer: whitespace splits + `"..."` grouping only. No single quotes, no escapes, no env expansion.
- If body has no `$ARGUMENTS` ref and args were passed, appends `\n\nARGUMENTS: <raw>` (matches CC).

## File layout (matches CC)

- `<dir>/foo.md` → `/foo`
[...]

PR #267, opened 2026-06-05
Add slash commands to the agency agent (v2)

egonSchiele:
## Summary

Adds Claude-Code-style slash commands to the agency agent.

- `commandsDir(dir)` (stdlib) — discovers `.md` files in `dir`, returns `{ name, description, argHint, body }[]`.
- `expandSlash(msg, commands)` (stdlib) — substitutes `$ARGUMENTS` in the matched command body; passes unknown `/foo` through unchanged.
- Wired into the agent's REPL **and** one-shot path. Project-local `.claude/commands/foo.md` becomes `/foo`. Built-ins (`/exit`, `/clear`, `/help`) win over file commands.

## Scope

**In v1**
- Read `.md` files under `<project>/.claude/commands/`
- Substitute `$ARGUMENTS` (only placeholder)
- Append `\n\nARGUMENTS: <raw>` when the body has no `$ARGUMENTS` token but args were passed (matches CC)
- Surface entries in the `/` palette

**Deferred (each is a follow-up PR if anyone asks)**
- `$N` / `$ARGUMENTS[N]` positional refs
- Namespaced commands (`/ns:scoped`)
- `.markdown` extension (just `.md` for now)
- `` !`cmd` `` shell injection
- `@<path>` file references
- `allowed-tools` / `model` / `effort` / any frontmatter field besides `description` / `argument-hint`
- User-level commands (`~/.claude/commands/`)
- Live reload

## Tests

20 execution tests, every assertion exact-equality (`==`), all green:

- 7 discovery tests (entries count, description/argHint defaults, missing dir, undefined-leak contract)
- 8 expansion tests (substitution, raw-string preservation, suffix fallback, unknown / non-slash / empty pass-through)
- 4 whitespace-tolerance tests (trailing newline, leading/trailing ws, tab as delimiter)
- 1 defensive contract test (`description` / `argHint` are always strings, never `null`/`undefined`)

## Notes

- Replaces #266 (closed unmerged). The earlier PR over-engineered this with a hand-rolled tokenizer, a `Command`/`CommandSet` type, a `.partial`-based `dispatch` closure, and CC's full arg grammar. v2 is a thin wrapper on the existing `skillsDir` pattern (~75 lines of production code) — see `docs/superpowers/plans/2026-06-05-slash-commands-v2.md` for the full reasoning.
- Fixed a subtle `undefined`-leak bug in the process: in Agency `undefined == null` evaluates to `false`, so the original `if (fm["argument-hint"] != null)` guard let `undefined` through and stringified it as `"undefined"`. Switched to `?? ""` coalescing, which treats both nullish.

## Manual smoke

```bash
mkdir -p /tmp/cmd-smoke/.claude/commands
cat > /tmp/cmd-smoke/.claude/commands/echo.md <<'EOF'
---
description: Echo back the args
---
Please repeat this verbatim: $ARGUMENTS
EOF
cd /tmp/cmd-smoke
pnpm --prefix ~/agency-lang/packages/agency-lang run agent -p '/echo hello'
```

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -227,17 +227,23 @@ def mergedPalette(): Record<string, string> {
 // Slash-command + user-message dispatcher invoked by `repl()` for
 // every Enter keystroke. Return `false` to terminate the loop.
 def _runTurn(msg: string): boolean {
-  if (msg == "") {
+  // Trim once for built-in matching so `"/help\n"` (piped) or
+  // `"  /clear"` behave the same as `"/clear"`. `expandSlash` is
+  // whitespace-tolerant in its own right, but it still receives the
+  // raw `msg` so the LLM sees the exact text the user typed when
+  // nothing matches.
+  const trimmed = msg.trim()
+  if (trimmed == "") {
     return true
   }
-  if (msg == "/exit" || msg == "/quit") {
+  if (trimmed == "/exit" || trimmed == "/quit") {
     return false
   }
-  if (msg == "/clear") {
+  if (trimmed == "/clear") {
     clearMessages()
     return true
   }
-  if (msg == "/help") {
+  if (trimmed == "/help") {
     pushMessage("Commands: /exit, /clear, /help")
     return true
   }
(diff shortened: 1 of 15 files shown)


packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -191,6 +192,38 @@ def loadAgentsMd(dir: string): string {
 // LLM doesn't have to re-ground each turn. Initialized in `main()`.
 let _context: string = ""
 
+// Project-local slash commands. `cwd()` works at module-init time so
+// we anchor at the user's project root (not the agent's source dir).
+// `with approve` auto-approves the `glob`/`read` during init — the
+// user implicitly opted in by running the agent in this directory.
+static const projectCommands = commandsDir("${cwd()}/.claude/commands") with approve
```
egonSchiele, line 199: use path.join

PR #268, opened 2026-06-06
Block-type: accept named params and `->` arrow

egonSchiele:
## Summary

Two related changes to Agency's type grammar for function-typed values:

1. **Block-type params may now be named.** `(userMsg: string) -> string` is
   accepted alongside the legacy bare-type form `(string) -> string`.
   Names are optional and per-param; mixed lists work
   (`(a: string, number, c: boolean) -> any`).
2. **`->` is the preferred arrow for block-types**, matching inline-block
   lambda syntax (`\x -> x + 1`). `=>` keeps parsing as a permanent
   alias; `pnpm run fmt` silently rewrites `=>` → `->` on next save.

Closes the gap that prevented `stdlib/agent.agency:139`'s `AgentSpec`
from parsing:

```agency
export type AgentSpec = {
  agent: (userMsg: string) -> string
}
```

## Why this is small

`BlockType.params` in `lib/types/typeHints.ts:120-125` already had
`{ name: string; typeAnnotation: VariableType }`. The parser just
hard-coded `name: ""` and discarded whatever it parsed. Adding name
support is purely additive — no AST change, no downstream consumer
update needed (`typescriptPreprocessor` and `typescriptBuilder` only
read `.typeAnnotation`).

## What changed

| File | Change |
|------|--------|
| `lib/parsers/parsers.ts` | New `blockTypeParam` parser: tries `name: type` first, falls back to bare type. Arrow accepts `or(str("->"), str("=>"))`. |
| `lib/utils/formatType.ts` | Branches on output dialect: Agency source → `(name: T) -> R`, TS codegen → `(T) => R` (TS-syntax, names dropped to avoid double-naming downstream). |
| `lib/backends/typescriptGenerator/typeToString.ts` | Same dialect split, keyed on existing `forFormatting` boolean. |
| `lib/parsers/typeHints.test.ts` | 15 new parser tests: both arrows, named/unnamed/mixed params, zero-arg, disambiguation against `parenthesizedTypeParser`. |
| `lib/formatter.test.ts` | New test locks in the `=>` → `->` migration and named-param round-trip. |

Plan with full rationale + risk register:
`docs/superpowers/plans/2026-06-05-block-type-named-params.md`.

## Migration

Commit 2 (`fmt: migrate block-type arrows from `=>` to `->`)`) is a
mechanical sweep: `pnpm run fmt -i` over every `.agency` file under
`stdlib/`, `lib/`, `tests/`, `docs/`, and `examples/` that used the
legacy `=>` arrow in a block-type. 50 files. Three files
(`pfa-bound-block-invoked.agency`,
`optional-block-dropped-from-schema.agency`) had to be hand-edited via
`sed` because `fmt` crashes on them (pre-existing `agencyGenerator.ts`
bug "Unhandled Agency node type: blockArgument", confirmed present on
`main`). Behavior unchanged.

## Tests

- **Parser**: 1424/1424 in `lib/parsers/` (15 new in `typeHints.test.ts`)
- **Backends**: 265/265 in `lib/backends/`
- **Formatter**: 34/34 in `lib/formatter.test.ts` (1 new)
- **Full unit suite**: 5293/5294 — the one failure is a pre-existing
  `stdlib: 'agent'` typecheck (`Type 'number' is not assignable to type
  'boolean' (condition)`), confirmed present on `main` and unrelated to
  this change.
- **Slash-commands integration**: 25/25 in `tests/agency/commands-dir.agency`.
- **End-to-end smoke**: hand-written probe compiled, ran, AST verified
  to show `name: "userMsg"` in the `blockType.params[]` entry. Deleted
  before commit.

## Out of scope (per plan)

- Default-valued block-type params (`(x: number = 0) -> ret`)
- Param destructuring in block-types
- Renaming the AST node from `blockType` to `functionType`
- A deprecation warning for `=>` (user opted for silent fmt migration)
- Surfacing names in TS codegen output

## Risk notes

- Pre-existing `agencyGenerator.ts` cannot format files that use
  `blockArgument` as a function-call argument. Two test fixtures
  affected — hand-migrated. Not blocking, but worth a separate
  follow-up.
- `parenthesizedTypeParser` disambiguation: `(name: string)` outside a
  block-type now produces a parse error (it always should have). Test
  guard added.

### DIFF
--- a/packages/agency-lang/docs/site/stdlib/thread.md
+++ b/packages/agency-lang/docs/site/stdlib/thread.md
@@ -98,9 +98,9 @@ Get the cumulative cost (USD, floating point) of all LLM calls
   happened and cost real money).
 
   To measure a specific section, capture getCost() before and after:
-    const before = getCost()
-    // ... do work ...
-    const sectionCost = getCost() - before
+  const before = getCost()
+  // ... do work ...
+  const sectionCost = getCost() - before
 
 **Returns:** `number`
 
@@ -133,11 +133,11 @@ Run a block under a cost limit, a time limit, or both. The block
   variables are scoped to the block — only the block's return value
   is observable from the caller. Use isFailure(result) to branch on
   the trip; inspect `result.error.type` to distinguish:
-    - "guardFailure" — cumulative LLM cost exceeded `cost`. Read
-      `result.error.maxCost` and `result.error.actualCost`.
-    - "timeoutFailure" — compute time inside the block exceeded
-      `time`. Read `result.error.maxTime` and `result.error.actualTime`
-      (milliseconds).
+  - "guardFailure" — cumulative LLM cost exceeded `cost`. Read
+  `result.error.maxCost` and `result.error.actualCost`.
+  - "timeoutFailure" — compute time inside the block exceeded
+  `time`. Read `result.error.maxTime` and `result.error.actualTime`
+  (milliseconds).
 
   Time semantics are compute-time: wall clock only ticks while a
   Runner is actively executing inside the guarded scope. Time spent
@@ -167,16 +167,16 @@ Run a block under a cost limit, a time limit, or both. The block
   @param block - The work to run under the guard.
 
   Example:
-    const result = guard(cost: $2.0, time: 30s) as {
-      const a = llm("step 1")
-      const b = llm("step 2")
-      return a + b
-    }
-    if (isFailure(result)) {
-      print("Guard tripped: " + result.error.type)
-    } else {
-      print(result.value)
-    }
+  const result = guard(cost: $2.0, time: 30s) as {
+  const a = llm("step 1")
+  const b = llm("step 2")
+  return a + b
+  }
+  if (isFailure(result)) {
+  print("Guard tripped: " + result.error.type)
+  } else {
+  print(result.value)
+  }
 
 **Parameters:**
 
--- a/packages/agency-lang/docs/site/stdlib/thread.md
+++ b/packages/agency-lang/docs/site/stdlib/thread.md
@@ -167,16 +167,19 @@ Run a block under a cost limit, a time limit, or both. The block
   @param block - The work to run under the guard.
 
   Example:
+
+  ```ts
   const result = guard(cost: $2.0, time: 30s) as {
-  const a = llm("step 1")
-  const b = llm("step 2")
-  return a + b
+    const a = llm("step 1")
+    const b = llm("step 2")
+    return a + b
   }
   if (isFailure(result)) {
-  print("Guard tripped: " + result.error.type)
+    print("Guard tripped: " + result.error.type)
   } else {
-  print(result.value)
+    print(result.value)
   }
+  ```
 
 **Parameters:**
 
--- a/packages/agency-lang/lib/backends/typescriptGenerator/typeToString.ts
+++ b/packages/agency-lang/lib/backends/typescriptGenerator/typeToString.ts
@@ -144,9 +144,9 @@ export function variableTypeToString(
   } else if (variableType.type === "typeAliasVariable") {
     return `${variableType.aliasName}${formatValueArgs(variableType.valueArgs)}`;
   } else if (variableType.type === "blockType") {
-    // Source-level Agency output uses `->` and surfaces param names.
-    // TS codegen uses `=>` (TS syntax) and drops names; downstream
-    // typescriptBuilder.ts already injects names where it needs them.
+    // Dialect-keyed arrow: `->` for Agency source, `=>` for TypeScript
+    // codegen. Param names are surfaced in both dialects when present
+    // (TS function types accept named params).
     const arrow = forFormatting ? "->" : "=>";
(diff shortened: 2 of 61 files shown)


packages/agency-lang/docs/site/stdlib/thread.md
```diff
@@ -167,16 +167,16 @@ Run a block under a cost limit, a time limit, or both. The block
...
-      const a = llm("step 1")
-      const b = llm("step 2")
-      return a + b
-    }
-    if (isFailure(result)) {
-      print("Guard tripped: " + result.error.type)
-    } else {
-      print(result.value)
-    }
+  const result = guard(cost: $2.0, time: 30s) as {
```
egonSchiele, line 172: put example inside code fence (language: ts for syntax highlighting)

packages/agency-lang/lib/backends/typescriptGenerator/typeToString.ts
```diff
@@ -144,11 +144,22 @@ export function variableTypeToString(
...
+    const arrow = forFormatting ? "->" : "=>";
     const params = variableType.params
-      .map((p) => variableTypeToString(p.typeAnnotation, typeAliases, forFormatting))
+      .map((p) => {
+        const t = variableTypeToString(
+          p.typeAnnotation,
+          typeAliases,
+          forFormatting,
+        );
+        return forFormatting && p.name ? `${p.name}: ${t}` : t;
```
egonSchiele: would adding the name in the generated TypeScript type be as easy as replacing this ternary with a single statement? Because if so, I'd say let's do it.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -1243,8 +1243,35 @@ export const unionTypeParser: Parser<UnionType> = memo(
...
+// Block type: () -> string, (number) -> any, (string, number) -> boolean
+// Params may be named or unnamed: (userMsg: string) -> string, (string) -> string.
+// Both `->` (preferred, matches inline-block lambda syntax) and `=>` (legacy)
+// are accepted; the formatter rewrites `=>` to `->` on next save.
+const blockTypeParam: Parser<{ name: string; typeAnnotation: VariableType }> =
+  or(
+    // Named param: `ident : type`. Try this alternative first — once we
+    // see `ident :` we're committed (no other grammar produces that shape
+    // inside a block-type param list).
+    map(
```
egonSchiele: You don't need the `map` call here. The `seqC` plus `capture` combination is generating this exact object for you.

packages/agency-lang/lib/utils/formatType.ts
```diff
@@ -35,8 +35,22 @@ export function formatTypeHint(
...
-      const params = vt.params.map((p) => recurse(p.typeAnnotation)).join(", ");
-      return `(${params}) => ${recurse(vt.returnType)}`;
+      // Use `primitiveAliases` (defined ⇒ codegen target) as the signal
+      // for which dialect to emit:
+      //   - Agency source (no aliases): `(name: T) -> R`, names surfaced.
+      //   - TypeScript codegen (aliases present): `(T) => R`, names
+      //     dropped. TS does accept names, but `typescriptBuilder.ts`
+      //     synthesizes parameter names downstream and emitting them
+      //     here would risk double-naming.
+      const isTsTarget = primitiveAliases !== undefined;
```
egonSchiele: Is this a durable way to detect whether target is ts?

packages/agency-lang/stdlib/thread.agency
```diff
@@ -130,16 +134,16 @@ export def guard(cost: (number | null) = null, time: (number | null) = null, blo
...
-      const a = llm("step 1")
-      const b = llm("step 2")
-      return a + b
-    }
-    if (isFailure(result)) {
-      print("Guard tripped: " + result.error.type)
-    } else {
-      print(result.value)
-    }
+  const result = guard(cost: $2.0, time: 30s) as {
```
egonSchiele, line 139: You'll need to add a code fence here (and intent correctly) to have the code fence show up in the docs. After indenting, please run the formatter again and see if it strips the indentation. If it does, we should fix that.

packages/agency-lang/tests/agency/parens-as-access-base.agency
```diff
@@ -10,8 +10,8 @@
 // same — only the receiver flavor changed.
 
 type Counter = {
-  n: number,
-  bump: () => number
+  n: number;
```
egonSchiele: I think you probably want to undo most of the changes to this file.

packages/agency-lang/tests/integration/cli-main/fixtures/project/src/fmt-input.agency
```diff
@@ -1,9 +1,37 @@
 /**Module docs for formatter fixture.*/
 // leading comment
-type Person={name:string,meta:{active:boolean}}
+type Person = {
```
egonSchiele: I assume this should stay as-is, I think this is a fixture for a test of the formatter.

PR #269, opened 2026-06-06
ui: prompts-backed line-mode chooseOption + select/autocomplete/prompt/confirm primitives

egonSchiele:
## Summary

Replace the hand-rolled line-mode `chooseOption` fallback with calls to the [`prompts`](https://www.npmjs.com/package/prompts) package (already a dependency), and expose `select` / `autocomplete` / `prompt` / `confirm` as new `std::ui` primitives.

Implements [docs/superpowers/plans/2026-06-05-prompts-package-line-mode.md](../tree/prompts-line-mode/packages/agency-lang/docs/superpowers/plans/2026-06-05-prompts-package-line-mode.md).

## What changed

- **Four new TS bridges** in `lib/stdlib/ui.ts` wrap `prompts.{autocomplete,select,text,toggle}`. Each guards on `process.stdout.isTTY` and `_hasActiveScreen()` (throws if invoked inside an active `repl()`). All return `Result<value, "cancelled">` via a shared `_runPrompt` helper.
- **Four new Agency primitives** in `stdlib/ui.agency`: `select`, `autocomplete`, `prompt` (free-form text — renamed from `text` to avoid colliding with the Layer 1 `text` element builder), `confirm`.
- **`chooseOption`'s line-mode branch** now delegates to `autocomplete()`. The loop only retries on user cancellation (Escape / Ctrl+C) to preserve the must-answer contract that safety-critical callers like `std::policy` depend on; non-cancel failures (e.g. non-TTY) propagate via `throw()` so the runtime wraps them as `Result.failure`.
- **Deleted** `_printChoicePromptFallback` (the styled banner renderer) and its banner comment — no longer used.

## Spec deviations

- Renamed `text` → `prompt` to avoid collision with the existing Layer 1 `text` element builder. The plan's snippets used `text`; the implementation, docs, and tests use `prompt`.
- `chooseOption`'s loop uses `throw(result.error)` on non-cancel failures rather than letting `autocomplete()`'s exception bubble through unconverted. This means non-TTY callers see a `Result.failure` with the original error message rather than an unhandled exception.
- `_runPrompt` defensively treats a `null`/`undefined` value as cancellation in addition to the `cancelled` flag set by `onCancel` — `prompts.inject([null])` resolves with `{value: null}` (not `{}`), so the original `!("value" in answer)` check wasn't sufficient.

## Codegen bug fix (bundled)

While wiring up Task 5, I hit a TypeScript builder bug: `throw("msg")` lowers to a `throw new Error(...)` statement, but `processFunctionCallAsStatement` was wrapping every call result in `const __funcResult = ...` for the interrupt-result check. The resulting `const __funcResult = throw new Error(...)` is invalid TS and esbuild rejects it.

Fix: bail out early for `throw` in `processFunctionCallAsStatement` (mirroring the existing `_emit` short-circuit) and register `throw` in `BUILTIN_FUNCTION_TYPES` so the undefined-function diagnostic doesn't warn on legitimate usage.

## Test plan

- `pnpm exec vitest run lib/stdlib/ui-smoke.test.ts` — 36/36 green, including:
  - bridge guard tests (non-TTY throws, active-repl throws) for `_promptsAutocomplete`
  - happy-path tests for all four bridges (success/cancel/free-text)
  - `chooseOption` line-mode tests (clean select, free-text, cancel-re-prompt loop, non-TTY surfaces as failure)
  - all pre-existing REPL state-machine tests still pass
- `pnpm exec vitest run lib/backends/typescriptBuilder` — 45/45 green (codegen change)
- `pnpm exec vitest run lib/typeChecker` — 764/764 green (builtins change)
- `pnpm run agency test tests/agency/policy-build-scoped-match` — passes (spot-check that `policy.agency` still works)
- `make` succeeds.

## Manual smoke test (not yet performed)

Per Task 10 of the plan, the TTY-driven `prompts` UX cannot be exercised by vitest. The author will drive it by hand before merging:

- `pnpm run agency lib/agents/agency-agent/agent.agency`, trigger a policy interrupt, verify the autocomplete UI renders correctly, arrow/filter/Enter work, free-text reason capture works, Escape re-opens the prompt (must-answer contract).
- Same agent piped through `cat` to strip the TTY — expect the run to die with `autocomplete requires a TTY (stdout is not a TTY)`.

## Out of scope

- TUI repl modal path (`_openChoicePrompt`) — untouched.
- Migrating any other prompt usage. `chooseOption` is the only consumer of the line-mode fallback.
- Adding a non-TTY graceful fallback — explicitly rejected in the spec (throw instead).


packages/agency-lang/stdlib/ui.agency
```diff
@@ -1006,37 +1153,35 @@ export def chooseOption(
...
+    // Only re-prompt on an explicit user-cancel; anything else
+    // (e.g. "requires a TTY") gets surfaced via throw so chooseOption
+    // doesn't loop forever in non-interactive environments.
+    if (result.error != "cancelled") {
+      throw(result.error)
     }
+    print("(answer required)")
   }
-  return answer
+  return ""  // unreachable; satisfies the type checker
```
egonSchiele: in agency,  comments have to be on their own line.

PR #271, opened 2026-06-06
Parser: anchor squiggles at the real error position, and target the params-separated-by-semicolon message

egonSchiele:
Two fixes that compound. They are both surfaced by inputs like:

```agency
def greet(name: string; greeting: string) { ... }
```

(`;` instead of `,` between parameters)

### 1. Squiggle position

For parse failures that don't go through `parseError(...)`, the LSP was anchoring its red squiggle at line 0, col 0 even when the underlying message embedded the correct location (`"Line X, col Y: ..."`).

`parseAgency` now extracts the tarsec rightmost-failure offset and converts it to editor-coordinate line/col via `buildLineTable` + `offsetToPosition`, populating `errorData` the same way the `TarsecError` branch does.

tarsec's own `getDiagnostics` also undercounts `\n` separators when walking its line table, so the `TarsecError` branch is overridden with the same rightmost-offset computation to avoid column drift once the template wrapper is applied.

### 2. Targeted message for `def NAME (... ; ...)`

Once `def NAME (` plus parameters is consumed, we're committed to a function definition. The parameter list's close-paren is now wrapped in `parseError(...)` so anything other than `,` (another param) or `)` (end of list) throws a `TarsecError` with:

> expected `` ` ``,`` ` `` between parameters or `` ` ``)`` ` `` to close the parameter list (note: parameters are separated by `` ` ``,`` ` ``, not `` ` ``;`` ` ``)

instead of letting `or(...)` backtrack to the top-level alternatives and dump every possible statement form the user might have meant (the prior message listed ~30 things: `"break", "continue", "import", "export", a node definition, ...`).

### Test impact

The `parseError` change reaches three existing function-parser unit tests (`def bad(x:)`, `def bad(: number)`, `def bad(x: number y: string)`) which previously returned a recoverable failure; they now throw and are marked `throws: true`. Three new tests pin the structured `errorData` line/col and the targeted message.

All 5283 unit tests pass.

### Before / after

```
$ cat foo.agency
import { highlight } from "std::syntax"

def greet(name: string; greeting: string) {
  print("${greeting}, ${name}!")
}
...

$ # BEFORE
Failed to parse Agency program: Line 5, col 10: expected whitespace, "break",
"continue", "import", "export", a node definition, "class", a type alias, an if
statement, a for loop, a while loop, a match block, "thread", "subthread",
"handle", "debugger", "skills", "skill", a function definition, a return
statement, a goto statement, an interrupt statement, a tag, "static", an
assignment, a function call, an expression, a boolean, "async", "sync",
"await", one of " 	", "/*", "//", "", "...", "...", or end of input

$ # AFTER
Failed to parse Agency program: expected `,` between parameters or `)` to
close the parameter list (note: parameters are separated by `,`, not `;`)
```

In the LSP, the squiggle now anchors on the `;` (line 2, col 22 in the source above) instead of line 0, col 0.

### DIFF
--- a/packages/agency-lang/lib/parser.ts
+++ b/packages/agency-lang/lib/parser.ts
@@ -1,10 +1,12 @@
 import {
+  buildLineTable,
   capture,
   eof,
-  failure,
   getErrorMessage,
+  getRightmostFailure,
   many,
   map,
+  offsetToPosition,
   or,
   Parser,
   ParserResult,
@@ -126,10 +128,20 @@ export function replaceBlankLines(input: string): string {
   );
 }
 
+/** Failure result enriched with the tarsec rightmost-failure position so
+ *  the LSP can surface a real squiggle range instead of falling back to
+ *  line 0, col 0. `pos` is a byte offset into the normalized input. */
+export type ParserFailureWithPos = {
+  success: false;
+  message: string;
+  rest: string;
+  rightmostPos?: number;
+};
+
 export function _parseAgency(
   input: string,
   config: AgencyConfig = {},
-): ParserResult<AgencyProgram> {
+): ParserResult<AgencyProgram> | ParserFailureWithPos {
   const normalized = normalizeCode(input);
   if (normalized.trim().length === 0) {
     return success(
@@ -151,8 +163,22 @@ export function _parseAgency(
   const result = agencyParser(normalized);
   if (!result.success) {
     const betterMessage = getErrorMessage();
+    const rightmost = getRightmostFailure();
     if (betterMessage) {
-      return failure(betterMessage, normalized);
+      return {
+        success: false,
+        message: betterMessage,
+        rest: normalized,
+        ...(rightmost ? { rightmostPos: rightmost.pos } : {}),
+      };
+    }
+    if (rightmost) {
+      return {
+        success: false,
+        message: result.message ?? "Parse error",
+        rest: normalized,
+        rightmostPos: rightmost.pos,
+      };
     }
   }
   return result;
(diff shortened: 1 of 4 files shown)


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3890,7 +3890,18 @@ const _baseFunctionParser: Parser<any> = memo(
...
+    // Once we've consumed `def NAME (` plus parameters, we're committed
+    // to a function definition. Anything other than `,` (another param)
+    // or `)` (end of list) is a user mistake — most commonly using `;`
+    // to separate params. Surface a targeted error via `parseError`
+    // (which throws a TarsecError, bypassing `or(...)` backtracking and
+    // `label(...)` suppression) instead of falling through to the
+    // top-level alternatives list and dumping every possible statement
+    // form the user might have meant.
+    parseError(
+      "expected `,` between parameters or `)` to close the parameter list (note: parameters are separated by `,`, not `;`)",
```
egonSchiele: remove `(note: parameters are separated by `,`, not `;`)`

packages/agency-lang/lib/parser.ts
```diff
@@ -187,22 +213,66 @@ export function parseAgency(
...
+    return { success: false, message: result.message, rest: result.rest };
   } catch (error) {
     if (error instanceof TarsecError) {
+      // Prefer the rightmost-failure offset over tarsec's own
+      // line/column. tarsec's `getDiagnostics` line/col computation
+      // ignores `\n` separators when walking the line table, which
+      // makes columns drift right by the number of preceding
+      // newlines once the template is applied. Recomputing from
+      // the absolute offset avoids that drift entirely.
+      const rightmost = getRightmostFailure();
```
egonSchiele, line 311: What is the difference between _parseAgency and parseAgency? because it seems like there is a lot of code duplication between the two. Would it make sense, instead of having two functions, to have a single function that calls the other one?

PR #274, opened 2026-06-08
Add explicit eval input and output annotations

egonSchiele:
## Summary
- Add `std::statelog.evalInput` / `evalOutput` wrappers and statelog wire events for explicit eval annotations.
- Change eval records to version 2 with `evalInputs[]` / `evalOutputs[]`, preserving `value`, `threadId`, `tMs`, and `truncated` metadata.
- Update `agency eval extract` to prefer explicit annotations, fall back to existing promptCompletion heuristics with actionable warnings, and cap oversized eval values at extraction time.

## Test Plan
- `make`
- `pnpm run ast stdlib/statelog.agency`
- `pnpm exec vitest run lib/stdlib/statelog.test.ts lib/eval/extract.test.ts`
- `pnpm run typecheck`
- `pnpm run lint:structure`
- CLI smoke: run an Agency file that calls `evalInput` / `evalOutput`, extract its statelog, and assert `evalInputRecorded` / `evalOutputRecorded` plus populated `evalInputs` / `evalOutputs` with no warnings

### DIFF
--- a/packages/agency-lang/lib/stdlib/statelog.ts
+++ b/packages/agency-lang/lib/stdlib/statelog.ts
@@ -1,0 +1,32 @@
+import { agencyStore } from "../runtime/asyncContext.js";
+import type { StatelogClient } from "../statelogClient.js";
+
+/** Pick the statelog client method to call for this eval event. */
+type EvalEmit = (
+  client: StatelogClient,
+  payload: { value: unknown; threadId: string | null },
+) => Promise<void>;
+
+/**
+ * std::statelog TS impls. Called from the agency-side wrappers in
+ * stdlib/statelog.agency, which pass through the user's value
+ * argument. Each function reads the active AgencyStore from
+ * AsyncLocalStorage and emits the corresponding wire event.
+ *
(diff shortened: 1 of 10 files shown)


packages/agency-lang/lib/stdlib/statelog.ts
```diff
@@ -0,0 +1,32 @@
...
+ * No-op when called outside an Agency execution frame (e.g. a tool
+ * function invoked directly from a test). This is the lenient pattern
+ * used by the generated-code accessors in lib/runtime/asyncContext.ts.
+ */
+async function emitEvalEvent(emit: EvalEmit, value: unknown): Promise<void> {
+  const frame = agencyStore.getStore();
+  if (!frame) return;
+  const safeValue = JSON.parse(JSON.stringify(value ?? null));
+  const threadId = frame.threads.activeId() ?? null;
+  await emit(frame.ctx.statelogClient, { value: safeValue, threadId });
```
egonSchiele: where is the `emit` func coming from, and why not call the statelog func directly?

PR #276, opened 2026-06-09
Add eval run command

egonSchiele:
## Summary
- Add `agency eval run` with task loading, run artifacts, per-task workdirs/statelogs, and eval-record extraction.
- Add `std::agency/eval.evalRun` backed by shared artifact helpers while preserving subprocess handler safety through `std::agency.run`.
- Add subprocess `cwd` and config override plumbing, plus IPC hardening for interrupt decision correlation and payload byte limits.

## Test Plan
- [x] `pnpm test:run lib/eval/loadTasks.test.ts lib/eval/runArtifacts.test.ts lib/runtime/ipc.configOverrides.test.ts lib/runtime/configOverrides.test.ts lib/cli/eval/run.test.ts lib/stdlib/agencyEval.test.ts lib/config.test.ts`
- [x] `pnpm run agency test tests/agency/subprocess/run-cwd.agency tests/agency/subprocess/run-log-file.agency tests/agency/subprocess/eval-run-basic.agency tests/agency/subprocess/limit-ipc-payload.agency`
- [x] `pnpm run typecheck`
- [x] `pnpm run build && make stdlib`


packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -162,13 +196,29 @@ export async function sendInterruptToParent(
...
     type: "interrupt",
+    interruptId: nanoid(),
     interrupt: interruptData,
     subprocessVotes: votes,
   } satisfies IpcInterruptMessage;
+  const serialized = serializedByteLength(outMsg);
+  if (!serialized.ok) {
+    const value = `Failed to serialize interrupt payload: ${serialized.error}`;
+    process.send!(buildIpcPayloadLimitError(subprocessIpcPayloadLimit, Infinity, value));
+    return { type: "reject", value };
```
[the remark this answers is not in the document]
egonSchiele: Fixed. The non-serializable interrupt path now sends a plain `type: "error"` IPC message instead of reusing the `limit_exceeded` payload shape, so it no longer reports misleading `Infinity bytes` diagnostics. Added a regression test in `lib/runtime/ipc.configOverrides.test.ts`.

packages/agency-lang/lib/eval/runArtifacts.ts
```diff
@@ -0,0 +1,138 @@
...
+  runsDir: string;
+  agent: string;
+  tasksSource: string;
+  tasks: EvalRunTask[];
+  continueOnError: boolean;
+  startedAt: Date;
+}): EvalRunState {
+  const runDir = path.resolve(args.runsDir, args.runId);
+  const tasksDir = path.join(runDir, "tasks");
+  fs.mkdirSync(tasksDir, { recursive: true });
```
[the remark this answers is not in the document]
egonSchiele: Fixed. `initializeEvalRun` now validates `runId` against the same restricted artifact-id character set we already use for task loading, so stdlib callers cannot escape `runsDir` with `..`, slashes, or absolute paths. Added a regression test in `lib/eval/runArtifacts.test.ts`.

packages/agency-lang/lib/eval/runArtifacts.ts
```diff
@@ -0,0 +1,138 @@
...
+
+export function prepareEvalRunTask(state: EvalRunState, task: EvalRunTask): PreparedEvalRunTask {
+  const taskDir = path.join(state.tasksDir, task.task_id);
+  const workdirPath = path.join(taskDir, "workdir");
+  fs.mkdirSync(taskDir, { recursive: true });
+  if (task.working_dir) {
+    fs.cpSync(task.working_dir, workdirPath, { recursive: true });
+  } else {
+    fs.mkdirSync(workdirPath, { recursive: true });
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed. `prepareEvalRunTask` now validates `task.task_id` before constructing the task directory, and it checks that `working_dir` is a directory before copying it into the per-task `workdir`. Added regression coverage for both cases in `lib/eval/runArtifacts.test.ts`.

packages/agency-lang/lib/cli/eval/run.ts
```diff
@@ -0,0 +1,194 @@
...
+  tasks?: string;
+  goal?: string;
+  runId?: string;
+  runsDir?: string;
+  continueOnError?: boolean;
+  verbose?: boolean;
+  config?: AgencyConfig;
+};
+
+export type EvalRunCliDependencies = {
```
egonSchiele: Is this for dependency injection or something? I am finding this pattern really ugly. Why not create an interface and then create classes that match that interface if you really need this?

packages/agency-lang/lib/cli/eval/run.ts
```diff
@@ -0,0 +1,194 @@
...
+  },
+};
+
+async function runCompiledAgent(args: {
+  compiledPath: string;
+  node: string;
+  args: Record<string, any>;
+  cwd: string;
+  statelogPath: string;
+}): Promise<{ ok: true } | { ok: false; errorMessage: string }> {
```
egonSchiele, line 201: terrible formatting, leave some blank lines, why is the `limits` obj defined on one line?

packages/agency-lang/lib/cli/eval/run.ts
```diff
@@ -0,0 +1,194 @@
...
+    let settled = false;
+    const settle = (value: { ok: true } | { ok: false; errorMessage: string }) => {
+      if (settled) return;
+      settled = true;
+      resolve(value);
+    };
+    child.stdout?.pipe(process.stdout);
+    child.stderr?.pipe(process.stderr);
+    child.on("message", (msg: any) => {
+      if (msg?.type === "result") settle({ ok: true });
```
egonSchiele: Are you being charged by the line? I don't understand why everything is so scrunched up.

packages/agency-lang/lib/eval/runArtifacts.ts
```diff
@@ -0,0 +1,145 @@
...
+  const prepared = {
+    task,
+    taskDir,
+    taskJsonPath: path.join(taskDir, "task.json"),
+    statelogPath: path.join(taskDir, "statelog.jsonl"),
+    evalRecordPath: path.join(taskDir, "eval-record.json"),
+    workdirPath,
+    errorPath: path.join(taskDir, "error.txt"),
+  };
+  fs.rmSync(prepared.statelogPath, { force: true });
```
egonSchiele: use safeDeleteFile and safeDeleteDirectory from lib/utils.ts to delete things.

egonSchiele: unless we think these paths could ever be outside a project dir (see checkInsideProject func)

packages/agency-lang/lib/eval/runArtifacts.ts
```diff
@@ -0,0 +1,145 @@
...
+  fs.rmSync(prepared.evalRecordPath, { force: true });
+  fs.rmSync(prepared.errorPath, { force: true });
+  writeJson(prepared.taskJsonPath, task);
+  return prepared;
+}
+
+export function recordEvalRunTaskError(prepared: PreparedEvalRunTask | EvalRunTask, errorMessage: string): EvalRunTaskResult {
+  if ("taskDir" in prepared) {
```
egonSchiele: You have a lot of types and objects all over this code that are almost the same but not quite. This is a really gross pattern because it's not immediately clear to the reader what the difference between these objects is. Readers have to go through each key one by one and compare and see which keys are in both types and which keys are specific to one type. Please clean up this logic.

packages/agency-lang/lib/runtime/configOverrides.ts
```diff
@@ -0,0 +1,49 @@
...
+export function applyRuntimeConfigOverridesToContextArgs(
+  args: RuntimeContextConstructorArgs,
+  overrides: Partial<AgencyConfig> | undefined = activeRuntimeConfigOverrides,
+): RuntimeContextConstructorArgs {
+  if (!overrides) return args;
+  return {
+    ...args,
+    statelogConfig: {
+      ...args.statelogConfig,
+      ...(overrides.observability !== undefined ? { observability: overrides.observability } : {}),
```
egonSchiele: wow this is incredibly ugly code.

PR #278, opened 2026-06-10
Add `withLock` primitive for per-run, cross-process mutual exclusion.

egonSchiele:
## Summary
- Add a per-run `withLock` mutex primitive with diagnostics, timeout support, and non-reentrant owner checks.
- Coordinate `withLock` across `std::agency.run` subprocesses via parent-side IPC arbitration and session cleanup.
- Expose `withLock` from `std::agency` and wrap CLI policy prompts in `withLock("std::tty")` to prevent interleaved interactive prompts.

## Design notes
The lock state lives on `RuntimeContext` as named promise-chain mutexes with owner/waiter/releaser diagnostics. In subprocess mode, children send `lockAcquire`/`lockRelease` messages to the parent; the parent arbitrates against the same local lock chain and cleans up any held locks when the child session settles or exits.

## Test Plan
- [x] `make`
- [x] `pnpm test:run lib/runtime/lock.test.ts lib/runtime/agency.test.ts lib/runtime/ipc.configOverrides.test.ts`
- [x] `pnpm run agency test tests/agency/lock/basic.test.json`
- [x] `pnpm run agency test tests/agency/subprocess/lock-cross-process.test.json`
- [x] `pnpm run typecheck`
- [x] `pnpm run lint:structure`

### DIFF
--- a/packages/agency-lang/lib/runtime/lock.ts
+++ b/packages/agency-lang/lib/runtime/lock.ts
@@ -27,10 +27,19 @@ function createTimeout(name: string, ms: number): Promise<never> {
   });
 }
 
-function startWarnTimer(name: string, ownerId: string, warnAfterMs: number): NodeJS.Timeout | null {
+function startWarnTimer(
+  ctx: RuntimeContext<any>,
+  name: string,
+  waiterId: string,
+  warnAfterMs: number,
+): NodeJS.Timeout | null {
   if (warnAfterMs <= 0) return null;
   return setTimeout(() => {
-    console.warn(`Still waiting for lock '${name}' after ${warnAfterMs}ms (owner ${ownerId})`);
+    const currentOwner = ctx.lockOwners[name] ?? "<none>";
+    console.warn(
+      `Still waiting for lock '${name}' after ${warnAfterMs}ms ` +
+        `(waiter ${waiterId}, current owner ${currentOwner})`,
+    );
   }, warnAfterMs);
 }
 
@@ -77,7 +86,6 @@ export async function acquireLocalLock(
   ctx.locks[name] = chain;
   rememberWaiter(ctx, name, ownerId);
 
-  let timedOut = false;
   let released = false;
   const release = () => {
     if (released) return;
@@ -90,6 +98,7 @@ export async function acquireLocalLock(
   };
 
   const warnTimer = startWarnTimer(
+    ctx,
     name,
     ownerId,
     opts.warnAfterMs ?? DEFAULT_WARN_AFTER_MS,
@@ -103,17 +112,11 @@ export async function acquireLocalLock(
       await waitForPrevious;
     }
   } catch (err) {
-    timedOut = true;
     forgetWaiter(ctx, name, ownerId);
     previous.finally(release);
     throw err;
   } finally {
     if (warnTimer) clearTimeout(warnTimer);
-  }
-
-  if (timedOut) {
-    release();
-    throw new Error(`Timed out waiting for lock '${name}'`);
   }
 
   ctx.lockOwners[name] = ownerId;
--- a/packages/agency-lang/lib/runtime/lock.ts
+++ b/packages/agency-lang/lib/runtime/lock.ts
@@ -1,0 +1,151 @@
+import { nanoid } from "nanoid";
+import type { RuntimeContext } from "./state/context.js";
+
+const DEFAULT_WARN_AFTER_MS = 30_000;
+
+export type WithLockOptions = {
+  ownerId?: string;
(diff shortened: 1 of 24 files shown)


packages/agency-lang/lib/runtime/lock.ts
```diff
@@ -0,0 +1,151 @@
+import { nanoid } from "nanoid";
+import type { RuntimeContext } from "./state/context.js";
+
+const DEFAULT_WARN_AFTER_MS = 30_000;
```
egonSchiele, line 4: what's the right way to make this configurable? key in `agency.json`?

packages/agency-lang/stdlib/agency.agency
```diff
@@ -141,6 +142,30 @@ export def runFile(
   )
 }
 
+export def withLock(
```
egonSchiele: this is not the correct module for this. Create a new `std::mutex` or `std::concurrency` model

packages/agency-lang/lib/runtime/state/context.ts
```diff
@@ -268,6 +276,10 @@ export class RuntimeContext<T> {
     execCtx.callbacks = {};
     execCtx.topLevelCallbacks = [];
     execCtx.onStreamLock = false;
+    execCtx.locks = {};
```
egonSchiele, line 279: what happens if there's a `fork` call inside a `withLock`? Do we need to copy these onto branches?

packages/agency-lang/lib/runtime/lock.ts
egonSchiele: Also, is this the only way we catch deadlocks?

egonSchiele: Keeping this out of agency.json for this PR. Current behavior is per-call: warnAfterMs controls diagnostics and timeoutMs controls bounded waits. I documented the MVP deadlock story in docs/dev/lock.md: same-owner reentrancy is detected immediately; multi-lock cycles are not detected automatically.

packages/agency-lang/lib/runtime/lock.ts
```diff
@@ -0,0 +1,151 @@
...
+      reject(new Error(`Timed out waiting for lock '${name}' after ${ms}ms`));
+    }, ms);
+  });
+}
+
+function startWarnTimer(name: string, ownerId: string, warnAfterMs: number): NodeJS.Timeout | null {
+  if (warnAfterMs <= 0) return null;
+  return setTimeout(() => {
+    console.warn(`Still waiting for lock '${name}' after ${warnAfterMs}ms (owner ${ownerId})`);
+  }, warnAfterMs);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in da244a87. The warning now identifies the waiting owner as waiter and includes the current lock owner separately.

packages/agency-lang/lib/runtime/lock.ts
```diff
@@ -0,0 +1,151 @@
...
+  const current = new Promise<void>((resolve) => {
+    releaseCurrent = resolve;
+  });
+  const chain = previous.catch(() => undefined).then(() => current);
+  ctx.locks[name] = chain;
+  rememberWaiter(ctx, name, ownerId);
+
+  let timedOut = false;
+  let released = false;
+  const release = () => {
```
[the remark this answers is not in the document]
egonSchiele: Fixed in da244a87. Removed the timedOut flag and unreachable branch so timeout handling is in the catch path only.

packages/agency-lang/stdlib/agency.agency
egonSchiele: Moved the Agency-facing API to a new std::concurrency module in da244a87. std::agency no longer exports withLock; policy/tests import from std::concurrency.

packages/agency-lang/lib/runtime/state/context.ts
egonSchiele: The lock queues stay on RuntimeContext and are intentionally shared across fork branches; copying them onto branches would split the mutex. I added docs and a fixture showing fork branches can run inside withLock when they do not reacquire the same lock. Reacquiring the same lock inside a fork spawned under that lock is documented as a deadlock pattern to avoid.

packages/agency-lang/lib/runtime/lock.ts
egonSchiele: Documented in da244a87. This PR catches immediate same-owner reentrancy, provides stalled-wait warnings, and supports timeoutMs for bounded waits. It does not implement full wait-for-graph cycle detection for multi-lock deadlocks.

PR #279, opened 2026-06-10
Add eval optimize command

egonSchiele:
## Summary
- Add the core eval-driven optimize loop and wire it to `agency eval optimize`
- Add stdlib `agency.eval.optimize(...)` without installing CLI-only auto-approval handlers
- Add `eval.optimizeRunsDir` config support and document CLI/std-lib usage

## Test Plan
- `pnpm test:run lib/optimize/*.test.ts lib/cli/eval/optimize.test.ts lib/stdlib/agencyEval.test.ts scripts/agency.test.ts`
- `make`


packages/agency-lang/lib/optimize/ast.ts
```diff
@@ -0,0 +1,105 @@
...
+  return "";
+}
+
+export function updatePrompt(target: OptimizeTarget, newPrompt: string): void {
+  if (target.llmCall && target.llmCall.type === "functionCall" && target.llmCall.arguments[0]?.type === "string") {
+    target.llmCall.arguments[0].segments = parsePromptToSegments(newPrompt);
+  }
+}
+
+export function parsePromptToSegments(prompt: string): PromptSegment[] {
```
egonSchiele, line 24: Doesn't agency already have a parser for this? I'm wondering why you are writing one from scratch.

packages/agency-lang/lib/optimize/ast.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+  }
+  return targets;
+}
+
+function collectOptimizeTargets(
+  body: AgencyNode[],
+  targets: OptimizeTarget[],
+): void {
+  for (const node of body) {
```
egonSchiele: This would miss any optimize targets nested inside a thread block, for example. Please use the walkNodesArray function instead, which already has the logic to recurse inside every single node body.

packages/agency-lang/lib/optimize/ast.ts
```diff
@@ -0,0 +1,105 @@
...
+      } else if (node.type === "functionCall" && node.functionName === "llm") {
+        llmCall = node;
+      }
+
+      const target: OptimizeTarget = { node, llmCall, tag: optimizeTag };
+      target.promptValue = getPromptValue(target);
+      if (optimizeTag.arguments.length === 0) {
+        target.configKeys = ["prompt"];
+      } else {
+        target.configKeys = optimizeTag.arguments.map((argument) => {
```
egonSchiele: put this in a function with a clear name so its clear what it does

packages/agency-lang/lib/optimize/mutator.ts
```diff
@@ -0,0 +1,88 @@
...
+  prompt: z.string().min(1),
+  rationale: z.string().min(1),
+});
+
+export type MutatorModelCaller = (args: {
+  message: string;
+  model: string;
+}) => Promise<unknown>;
+
+export function buildMutatorMessage(args: {
```
egonSchiele, line 28: Since you are making an LLM call with structured output here, I would suggest writing this code in agency. It will be much easier to write.

packages/agency-lang/lib/stdlib/agencyEval.ts
```diff
@@ -159,3 +161,42 @@ export async function _evalJudge(
 ): Promise<PairwiseVerdict> {
   return judgePairwise(rubric, recordPathA, recordPathB);
 }
+
+/**
+ * Stdlib binding for `agency.eval.optimize`. This deliberately does not
+ * install any approval handler; callers decide which handlers are in scope.
+ */
+export async function _optimize(
```
egonSchiele, line 169: When you have this many arguments to a function, please use named arguments by asking users to pass in an object instead. If users call this with positional arguments, it will be very hard to know what each argument is for, and it will be very easy to mess up and make a subtle mistake by passing an argument in the wrong position.

egonSchiele: Oh, I see. If it is just a function that will get called from agency code, that is fine, because agency supports named arguments by default.

packages/agency-lang/lib/optimize/types.ts
```diff
@@ -0,0 +1,83 @@
...
+export type MutationProposal = {
+  prompt: string;
+  rationale: string;
+};
+
+export type ValidationResult =
+  | { ok: true }
+  | { ok: false; reason: string };
+
+export type OptimizeJudgeSample = {
```
egonSchiele, line 16: This many types, all with a similar shape, is usually a code smell. Even though I don't see a lot of overlap in the fields for each object, it is funny that we have multiple verdict objects and result objects, and they all have such a long list of keys. Maybe it is okay here, but usually this would be a code smell.

packages/agency-lang/lib/optimize/ast.ts
egonSchiele: Fixed: optimize target discovery now uses walkNodesArray, with a regression test for @optimize inside a thread block.

packages/agency-lang/lib/optimize/ast.ts
```diff
@@ -0,0 +1,114 @@
...
+          if (argument.type !== "variableName") {
+            const where = argument.loc ? ` (line ${argument.loc.line}, col ${argument.loc.col})` : "";
+            throw new Error(
+              `@optimize(...) arguments must be plain config-key identifiers${where}; got ${argument.type}.`,
+            );
+          }
```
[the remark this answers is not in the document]
egonSchiele: Fixed: user-facing @optimize argument locations now display 1-indexed line/column values, with a regression test for the error message.

packages/agency-lang/lib/optimize/ast.ts
egonSchiele: Fixed: replaced the custom interpolation regex with the existing Agency string parser and added coverage for nested braces in interpolation expressions.

packages/agency-lang/docs/site/cli/eval.md
```diff
@@ -65,6 +66,67 @@ runs/<run-id>/
...
+- `--tasks <file|dir>` — required task suite file or directory. Unlike `eval run`, optimizer tasks must come from a suite; `--goal` is reserved for the optimization objective.
+- `--goal <text>` — required plain-English objective used by the prompt mutator.
+- `--iterations <n>` — maximum candidate iterations after the baseline. Defaults to `5`.
+- `--judge-samples <n>` — pairwise judge samples per task. Defaults to `3`.
+- `--accept-threshold <n>` — accept when confident candidate wins minus losses is greater than this value. Defaults to `0`.
+- `--mutator-model <model>` — optional model override for proposing prompt mutations.
+- `--run-id <id>` — output run id. Defaults to a generated id.
+- `--runs-dir <path>` — optimizer output root. Defaults to `eval.optimizeRunsDir` in `agency.json`, or `eval.runsDir/optimize`, or `runs/optimize`.
+
+Only judge samples with confidence at least `50` count toward the win/loss margin. Pairwise judge confidence is an integer from `0` to `100`.
```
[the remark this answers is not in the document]
egonSchiele: Fixed: docs now say the confidence threshold applies to task-level aggregated verdicts, matching the implementation.

packages/agency-lang/lib/optimize/types.ts
egonSchiele: Adjusted the biggest many-key object: OptimizeLoopConfig is now grouped into runtime, target, policy, and artifacts. I left the verdict/result shapes flat because those are artifact/API payloads rather than call-site argument bags.

packages/agency-lang/lib/optimize/ast.ts
egonSchiele: Fixed: split the tag handling into named helpers, including readOptimizeConfigKeys and formatSourceLocation.

packages/agency-lang/lib/optimize/mutator.ts
egonSchiele: Fixed: prompt mutation now runs through bundled Agency code in lib/agents/mutatePrompt.agency. The TS wrapper calls that node and has regression coverage for the bundled execution path.

PR #282, opened 2026-06-12
Optimize declaration modifiers and target discovery

egonSchiele:
## Summary
- Add the `optimize` declaration modifier to parsing, AST, formatting, and TypeScript generation as a runtime no-op.
- Discover optimized variable declarations across local Agency import trees with stable `file:scope:name` target IDs.
- Remove the legacy top-level interactive optimize command and update eval optimize stdlib/CLI plumbing to use entry files while preserving nested paths.

## Follow-up work
- `lib/optimize/loop.ts` still uses the legacy tag-based mutation path. The declarative source mutator and refactored eval pipeline plans will rewire the loop in follow-up PRs.

## Test Plan
- `pnpm test:run lib/parsers/assignment.test.ts lib/parsers/exportConst.test.ts lib/backends/agencyGenerator.test.ts lib/backends/typescriptBuilder.integration.test.ts lib/optimize/targets.test.ts lib/cli/eval/optimize.test.ts lib/stdlib/agencyEval.test.ts`
- `make`
- `git diff --check -- lib/cli/eval/optimize.ts lib/cli/eval/optimize.test.ts lib/stdlib/agencyEval.ts lib/stdlib/agencyEval.test.ts`


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3318,6 +3344,7 @@ const _bodyNodeParser: Parser<AgencyNode> = memo("bodyNodeParser", or(
   multiLineCommentParser,
   commentParser,
   skillParser,
+  modifiedAssignmentParser,
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 28cd4a12. Body parsing now uses a body-scoped modified-assignment parser, so only `optimize` declarations are accepted in function/node bodies. `static` and `export` declaration modifiers in bodies now commit to a parse error instead of being parsed as declarations.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3161,11 +3178,19 @@ export const modifiedAssignmentParser: Parser<Assignment> = withLoc((input: stri
   }
 
   // If no modifiers found, this parser doesn't match
-  if (!isExported && !isStatic) return failure("expected 'export' or 'static'", input);
+  if (!isExported && !isStatic && !isOptimize) return failure("expected 'export', 'static', or 'optimize'", input);
+
+  if (isOptimize && isExported) {
+    return failure("export optimize declarations are unsupported in v1", input);
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 28cd4a12. `export optimize ...` is detected immediately after parsing `export`, so the intended unsupported-declaration diagnostic is reachable and covered by parser tests.

packages/agency-lang/lib/cli/eval/optimize.ts
```diff
@@ -89,3 +92,83 @@ function buildOptimizeLoopConfig(
...
+}
+
+function isInsideOrSame(candidate: string, parent: string): boolean {
+  const relative = path.relative(parent, path.resolve(candidate));
+  return relative === "" || (!relative.startsWith("..") && !path.isAbsolute(relative));
+}
+
+function relativeAgencyPath(baseDir: string, absoluteFile: string): string {
+  return path.relative(baseDir, absoluteFile).split(path.sep).join("/");
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 28cd4a12. CLI relative agent path generation now validates containment before returning `agentFilename`; tests cover absolute agents outside the invocation directory being materialized under their own working directory without `..` path segments.

packages/agency-lang/lib/stdlib/agencyEval.ts
```diff
@@ -201,17 +203,22 @@ export async function _optimize(
...
     policy: { goal, iterations, judgeSamples, acceptThreshold, mutatorModel },
     artifacts: { runsDir, runId: runId || nanoid() },
   });
 }
 
 type AgencyConfigLike = OptimizeLoopConfig["runtime"]["config"];
+
+function relativeAgencyPath(baseDir: string, absoluteFile: string): string {
+  return path.relative(baseDir, absoluteFile).split(path.sep).join("/");
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 28cd4a12. The stdlib `_optimize` binding now rejects entry files that resolve outside the explicit working directory before constructing `agentFilename`, with regression coverage.

packages/agency-lang/lib/cli/eval/optimize.ts
```diff
@@ -89,3 +92,87 @@ function buildOptimizeLoopConfig(
...
+  if (!parseResult.success) {
+    throw new Error(`Failed to parse optimize agent file ${canonicalFile}: ${parseResult.message ?? "parse error"}`);
+  }
+
+  for (const importPath of localAgencyImports(parseResult.result)) {
+    visitLocalAgencyFile(path.resolve(path.dirname(filePath), importPath), visited, files);
+  }
+}
+
+function localAgencyImports(program: AgencyProgram): string[] {
```
egonSchiele: There is a lot of duplication with your logic to detect imports and the logic in these files:

lib/importPaths.ts
lib/compilationUnit.ts
lib/symbolTable.ts

packages/agency-lang/lib/optimize/targets.ts
```diff
@@ -0,0 +1,203 @@
...
+  entryFile: string;
+  files: Record<string, OptimizeSourceFile>;
+  targets: OptimizeTarget[];
+};
+
+export type DiscoverOptimizeTargetsOptions = {
+  baseDir?: string;
+};
+
+export function discoverOptimizeTargets(
```
egonSchiele, line 38: wonder if this should get tracked in lib/symbolTable.ts? Feels like you're adding a lot of code here to find the targets, but files like the symbol table and compilation unit already do most of this for you. So it would be easier to add a quick new function to find optimize targets in there

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3133,19 +3133,39 @@ const staticKeywordParser: Parser<boolean> = or(
   succeed(false),
 );
 
+const optimizeKeywordParser: Parser<boolean> = (input: string): ParserResult<boolean> => {
+  const result = seqC(str("optimize"), spaces)(input);
+  if (!result.success) return success(false, input);
+  const nextToken = result.rest.trimStart();
+  if (/^(const|let|static)\b/.test(nextToken)) {
```
egonSchiele: Tarsec has a built-in `peek` parser that you could use for this instead, and use it right in the `seqC` call

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3178,9 +3206,58 @@ export const modifiedAssignmentParser: Parser<Assignment> = withLoc((input: stri
...
 });
 
+const bodyOptimizeAssignmentParser: Parser<Assignment> = (input: string) => {
+  const result = modifiedAssignmentParser(input);
+  if (!result.success) return result;
+  if (result.result.optimize) return result;
+  return failure("expected optimize assignment", input);
+};
+
+const BODY_RESERVED_MODIFIER_MESSAGE =
```
egonSchiele, line 3228: To me, it feels like you're trying to catch every invalid combination of these modifiers separately in your code. There is no need to add this sort of imperative logic, just write a parser in tarsec to declaratively say, what this statement should look like, and if users don't have it written correctly, the parser will complain. Is there any reason you are also trying to handle every single incorrect combination manually here?

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3178,9 +3206,58 @@ export const modifiedAssignmentParser: Parser<Assignment> = withLoc((input: stri
...
+    or(str("const"), str("let"), str("static"), str("optimize")),
+  );
+  if (!probe(input).success) return failure("", input);
+  return parseError(
+    BODY_RESERVED_MODIFIER_MESSAGE,
+    fail("body declaration modifier"),
+  )(input) as ParserResult<never>;
+};
+
+const bodyWithModifierParser: Parser<WithModifier> = withLoc((input: string) => {
```
egonSchiele: Again, it's not clear to me why you needed to add a parser for `with approve`. It feels unrelated to the change we are making.

PR #283, opened 2026-06-12
Declarative optimize mutator and refactored eval optimize pipeline

egonSchiele:
Implements the two optimize plans (`docs/superpowers/plans/2026-06-11-declarative-optimize-mutator.md` and `2026-06-11-optimize-refactored-eval-pipeline.md`) plus a logging/UX round, completing the migration of `agency eval optimize` from the legacy `@optimize(prompt)` tag pipeline to declaration-target discovery and a declarative source mutator.

## Declarative source mutator

- New `lib/optimize/sourceMutator.ts`: `OptimizeSourceMutator` validates declarative operation records (`{ target, kind, op, value, expected?, rationale? }`) against the discovered target catalog and applies them via parse → replace AST node → `AgencyGenerator` render — never by patching source text. Batches are atomic: one invalid operation means no files are produced.
- Validation produces structured diagnostics (unknown target, kind mismatch, unsupported op, duplicate target, stale `expected` compared against the decoded value, invalid Agency expression syntax, non-string values, interpolation-multiset mismatches, reserved `type` ops in v1) that feed the LLM mutator's retry prompt.
- `preview()` parses **only the files a batch touches**, from discovery-time sources, and returns the full candidate file set, target-level changes, an uncolored diff, and the **updated target set** — so the loop never re-discovers or re-parses unchanged files. Parse cost is one pass over the import closure at startup plus O(files touched) per iteration.
- `proposeMutation()` now emits `{ operations, rationale }`, prompted with sorted target ids/kinds/current values, suite goals in task-id order, mutation history, and prior rejected-preview diagnostics. The bundled `mutatePrompt.agency` schema matches.

## Refactored eval optimize pipeline

- CLI: positional `<agent>` target, exactly one of `--goal`/`--tasks` (goal desugars through the shared `taskFromGoal()`; fails upfront if the node requires args), shared judge flags (`--samples`, `--confidence-threshold`, `--margin-threshold`), `--no-writeback`, `--silent`.
- The loop is orchestration only: discover targets once at startup (the working-dir rule folded into discovery — one closure walk), run baseline and candidates through `evalRunLoadedTasks()`, judge champion (side A) vs candidate (side B) with `judgeSuite()`, accept iff `winner === "B"`. The champion is a file set; an accepted candidate's preview target set becomes the next mutation base.
- **Baseline fail-fast**: any baseline task failure aborts immediately, naming the failing tasks — a failure before any mutation means the program or task suite is broken.
- Multi-file writeback verifies every discovered file's discovery-time sha256 and aborts entirely on any mismatch.
- Artifact layout: `targets.json`, `iter-N/{agent/,workspace/,eval-run/,mutation.json,mutation.md,diff.txt,verdict.json}` (verdict is the raw `SuiteVerdict`), `champion/agent/` file set, `summary.json`; run-dir collisions throw.
- Deleted legacy machinery: `lib/optimize/ast.ts`, `sampling.ts`, `verdict.ts`, the `acceptThreshold`/`judgeSamples` knobs, and the single-prompt mutation path. The stdlib `optimize()` wrapper mirrors the CLI semantics and calls the same `optimizeLoop` core.

## Reporting

- New `OptimizeReporter` boundary (`lib/optimize/reporter.ts`): the loop emits semantic events; rendering decisions live in the reporter. Default mode prints the discovered-target gut check at startup, per-iteration phases and decisions with colored line diffs of each proposed value (dimmed when a proposal left a value unchanged), the mutator's rationale, champion-vs-candidate diffs of each task's final response, and a closing start → end summary of every optimized variable. `--silent` prints nothing; artifacts are always written.
- Internal compile progress lines (`x.agency → x.js (in Nms)`) are suppressed via a `quiet` option threaded through `compile()` / `executeNodeAsync` for candidate, mutator-agent, and judge-agent compiles. `agency compile` and `agency eval run` output is unchanged.

## Verification

- 2334 tests passing across the optimize, eval CLI, stdlib, and judge suites (≈90 new); typecheck and structural lint clean; `make` clean with stdlib + docs regenerated.
- Docs updated: `docs/site/cli/eval.md` (new surface, acceptance rule, artifact layout, output behavior) and regenerated `docs/site/stdlib/agency/eval.md`.

Note for reviewers: the tracked sample `packages/agency-lang/foo.agency` still uses the legacy `@optimize(prompt)` tag on main and is deliberately not touched by this PR; it needs a follow-up migration to the `optimize` modifier since the legacy tag is now rejected.


packages/agency-lang/lib/optimize/loop.ts
```diff
@@ -102,141 +103,245 @@ export async function optimizeLoop(
...
-): Promise<MutationProposal & { ok: true } | { ok: false; attemptedPrompt: string; rationale?: string; error: string }> {
-  let validationFailure: string | undefined;
+  champion: ChampionState,
+  history: MutationHistoryEntry[],
+): Promise<MutationOutcome> {
+  const sourceMutator = new OptimizeSourceMutator({ targetSet: champion.targetSet });
+  let diagnostics: OptimizeMutationDiagnostic[] | undefined;
+  let lastProposal: MutationProposal | null = null;
+
   for (let attempt = 0; attempt < 2; attempt += 1) {
```
egonSchiele: 2 is a magic number

packages/agency-lang/lib/optimize/loop.ts
```diff
@@ -102,141 +103,245 @@ export async function optimizeLoop(
...
+    agent: path.join(workspace.workspaceDir, config.target.entryFile),
+    tasks,
+    runsDir: path.join(artifacts.runDir, `iter-${iter}`),
+    runId: "eval-run",
+    config: config.runtime.config,
+    pipeAgentOutput: deps.pipeAgentOutput ?? false,
+  });
 }
 
 async function defaultEvalRun(args: {
```
egonSchiele, line 358: I think this function would be clearer if you just spread `args` in the parameter to `evalRunLoadedTasks` and then added any other arguments

packages/agency-lang/lib/optimize/loop.ts
```diff
@@ -102,141 +103,245 @@ export async function optimizeLoop(
...
-  runId: string,
-  runsDir: string,
 ): Promise<EvalRunResult> {
-  return (deps.evalRun ?? defaultEvalRun)({ agent, tasks, runsDir, runId, config: config.runtime.config });
+  const workspace = artifacts.writeIterationWorkspace(iter, files);
+  return (deps.evalRun ?? defaultEvalRun)({
+    agent: path.join(workspace.workspaceDir, config.target.entryFile),
+    tasks,
+    runsDir: path.join(artifacts.runDir, `iter-${iter}`),
+    runId: "eval-run",
```
egonSchiele: not sure why this has a hardcoded run id?

packages/agency-lang/lib/optimize/mutator.ts
```diff
@@ -7,78 +7,124 @@ import { z } from "zod";
 
 import type { AgencyConfig } from "@/config.js";
 import { executeNodeAsync } from "@/cli/util.js";
+import type { EvalTask } from "@/eval/runTypes.js";
```
egonSchiele, line 10: Given how much LLM-related functionality we need in this code – making an LLM call, getting structured output, et cetera – sure feels like this would be much simpler to write as an agency agent, as you would get a lot of that for free.

packages/agency-lang/lib/optimize/sourceMutator.ts
```diff
@@ -0,0 +1,374 @@
...
+  collectTargets,
+  promptSegmentsToString,
+  sha256Text,
+  type OptimizeSourceFile,
+  type OptimizeTarget,
+  type OptimizeTargetSet,
+} from "./targets.js";
+import { validateOptimizedStringValue } from "./validation.js";
+
+export type ReplaceVariableInitializerOperation = {
```
egonSchiele, line 30: Could you add short comments on what all these types are for, and an example of what an object of this type would look like? Nothing too long

packages/agency-lang/lib/optimize/sourceMutator.ts
```diff
@@ -0,0 +1,374 @@
...
+    return this.resolveReplacementValue(operation as ReplaceVariableInitializerOperation, target);
+  }
+
+  private resolveReplacementValue(
+    operation: ReplaceVariableInitializerOperation,
+    target: OptimizeTarget,
+  ): ValidationOutcome {
+    const parsed = exprParser(operation.value);
+    if (!parsed.success || parsed.rest.trim() !== "") {
+      return diagnostic(operation, "invalid-replacement-syntax", `Replacement value for ${operation.target} is not a single valid Agency expression: ${JSON.stringify(operation.value)}`);
```
egonSchiele: Let's have diagnostic take named arguments, so this sort of code is easier to read.

packages/agency-lang/lib/optimize/reporter.ts
```diff
@@ -0,0 +1,167 @@
...
+  if (!recordPath || !fs.existsSync(recordPath)) return null;
+  try {
+    const record = JSON.parse(fs.readFileSync(recordPath, "utf8"));
+    const response = selectFinalResponse(record);
+    return response.missing ? null : response.text;
+  } catch (error) {
+    console.error(`failed to read eval record ${recordPath}:`, error);
+    return null;
+  }
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 0143708b — readResponse() now takes the injected logger and reports read failures through it, so captured/redirected reporter output never sees stray stderr writes.

packages/agency-lang/lib/optimize/artifacts.ts
```diff
@@ -156,27 +176,36 @@ function isInsideOrSame(candidate: string, parent: string): boolean {
...
+      "Old value:",
+      "```",
+      change.oldValue,
+      "```",
+      "",
+      "New value:",
+      "```",
+      change.newValue,
+      "```",
+    ]),
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 0143708b — fences are now chosen dynamically (one backtick longer than the longest run in the value, minimum three), so prompts containing ``` keep mutation.md well-formed.

packages/agency-lang/lib/optimize/loop.ts
egonSchiele: Fixed in 0143708b — named MUTATION_PROPOSAL_ATTEMPTS = 2 with a comment explaining the shape: one initial proposal plus one retry with the rejected preview's diagnostics fed back.

packages/agency-lang/lib/optimize/loop.ts
egonSchiele: Fixed in 0143708b — defaultEvalRun now spreads args and adds only the optimize-specific options (tasksSource, continueOnError, quietCompile).

packages/agency-lang/lib/optimize/loop.ts
egonSchiele: It's a directory-name convention rather than an id: each iteration's eval run lives inside its own iter-N/ directory, so runsDir=iter-N plus runId="eval-run" materializes the fixed path iter-N/eval-run/. Named it EVAL_RUN_DIR_NAME with a comment saying exactly that (0143708b).

packages/agency-lang/lib/optimize/sourceMutator.ts
egonSchiele: Done in 0143708b — each public type now has a short doc comment with an example object.

packages/agency-lang/lib/optimize/sourceMutator.ts
egonSchiele: Done in 0143708b — diagnostic() now takes { operation, code, message } as named arguments.

packages/agency-lang/lib/optimize/mutator.ts
egonSchiele: Good instinct — and it's actually already mostly true: the LLM call and structured output do run through an Agency agent (lib/agents/mutatePrompt.agency, invoked by defaultCallModel via executeNodeAsync), which is where the schema enforcement comes for free. What this file had wrong was a duplicated copy of the agent's prompt skeleton in TS (buildMutatorMessage), kept around for the injectable test caller — a real drift risk. Fixed in 0143708b: the agent is now the single owner of the prompt; TS keeps only (1) deterministic data-section rendering (sorted targets, goals, history, diagnostics — unit-testable without an LLM), (2) the callModel seam the loop tests inject, and (3) zod validation at the process boundary since executeNodeAsync returns untyped JSON. Pushing the section rendering into the agent too is possible, but string-munging (sorting, quoting, escaping) is what TS does well and tests cheaply — happy to go further if you'd prefer it fully agent-side.

PR #285, opened 2026-06-14
Part A: layout handler-per-type refactor (no behavior change)

egonSchiele:
## Part A of 3 — layout handler-per-type refactor (no behavior change)

This is the first of three stacked PRs implementing the `std::chart` bar chart module. It is a **pure refactor of `std::layout`'s render pipeline with no behavior change** — every existing layout test passes unchanged.

Design docs (spec + plan) for the whole series are included here for context.

### What changed

`std::layout` previously kept each node type's two behaviors in two parallel tables: `SIZERS` (phase 1, width resolution) and `RENDERERS` (phase 2, paint to `Block`), with the sizers centralized in `render.ts` and the renderers spread across the per-concern files. This PR collapses them into one `HANDLERS` table and co-locates each node type's `size` + `render` in its own file.

- **New `sizing.ts`** holds the shared width helpers (`resolveOwnWidth`, `resolveContainer`, `innerWidthAfterChrome`, `nonNegativeInteger`, `setAttr`) plus the `SizingContext` and `NodeHandler` types.
- **`box.ts` / `axis.ts` / `nodes.ts` / `table.ts`** each now export a `NodeHandler` (`{ size, render }`) and own their sizer next to their renderer.
- **`render.ts`** slims to the orchestrator: it builds `HANDLERS` from the per-file handlers and dispatches `resolveNode` / `renderNode` through it. `RENDERERS` and `SIZERS` remain as derived views so the test surface and any external readers keep working.

The benign import cycle (per-concern files import `renderNode`/`resolveNode` from `render.ts`; `render.ts` imports their handlers) is the same one the codebase already documented for `renderNode` — all uses are inside function bodies, so module init completes before any call.

### Why

Adding a new first-class node type (the bar chart, in Part C) becomes a one-entry change instead of touching two dispatch tables in different files. It also makes each per-concern file the single source of truth for its node type.

### Verification

- `pnpm exec vitest run lib/stdlib/layout.test.ts` → **184 passed** (unchanged).
- `pnpm exec tsc --noEmit` → clean.
- `pnpm run lint:structure` → clean.

### Next in series

- **Part B:** extract the table constructor surface into a `std::table` module (renderer stays a registered handler here).
- **Part C:** add the horizontal bar chart as one new `barchart` handler + a `std::chart` module.

### DIFF
--- a/packages/agency-lang/docs/superpowers/plans/2026-06-13-chart-module.md
+++ b/packages/agency-lang/docs/superpowers/plans/2026-06-13-chart-module.md
@@ -1,0 +1,1449 @@
+# std::chart — Bar Chart Module Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:executing-plans to implement this plan task-by-task (the user prefers inline execution in the main session — do **not** use subagent-driven development). Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a `std::chart` stdlib module that renders horizontal bar charts (grouped + stacked, keyed series with auto color/symbol, legend, negative values via a zero baseline) as a first-class layout node.
+
+**Architecture:** Three phases. (A) Refactor `std::layout`'s render pipeline so each node type's size+render live together in one `HANDLERS` table — pure refactor, no behavior change. (B) Extract the table constructor surface into its own `std::table` Agency module (hard move; the renderer stays a registered handler). (C) Add the bar chart as one new handler (`barchart.ts`) plus a `std::chart` Agency module that mirrors `std::table`.
+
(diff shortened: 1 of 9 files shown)


egonSchiele:
Thanks — both addressed in f8a23d5f:

1. **resolveNode unknown-type guard:** extracted a shared `handlerFor(type)` helper so the sizing pass throws the same clear `unknown node type` error as the render pass (and DRYs the guard across both).
2. **Stale module-map comment:** updated to describe the `HANDLERS` dispatch table (with `RENDERERS`/`SIZERS` derived) and added the new `sizing.ts` entry.

Verified: 184 layout tests pass, `tsc --noEmit` and `lint:structure` clean.

PR #286, opened 2026-06-14
Part B: extract std::table from std::layout

egonSchiele:
## Part B of 3 — extract `std::table`

Second of three PRs for the `std::chart` work. This **moves the table constructor surface out of `std::layout` into a new `std::table` module**. It's a hard move (no back-compat re-export); the table *renderer* stays a registered handler in the layout pipeline, so rendered output is byte-identical.

### What changed

- **New `stdlib/table.agency`** (`std::table`): the `Cell` / `CellRow` / `ColumnSpec` / `TableBuilder` types, the table builder helpers, and the `table()` constructor — moved verbatim from `layout.agency`. `table()` now carries an in-function `"""…"""` docstring (with `@param` lines for PFA-aware tool descriptions). It imports the shared types (`LayoutNode` / `Width` / `Alignment` / `BorderStyle`) from `std::layout`.
- **`stdlib/layout.agency`**: table types/builders/`table()` removed; `@module` doc points readers to `std::table`. `std::layout` is now just the structural primitives (`box` / `row` / `column` / `text` / `raw` / `space` / `hline` / `vline` + `render`).
- **Repointed all 7 in-repo consumers** to `import { table } from "std::table"`: `stdlib/policy.agency`, `examples/layoutDemo.agency`, and the five table tests under `tests/agency/layout/`. Each keeps importing `render` / `text` from `std::layout`.
- **Regenerated `docs/site/stdlib`**: adds `table.md`, removes table from `layout.md`, and updates `policy.md` source line-refs.

### Note: no `render` re-export

The design called for `std::table` to re-export `render` for one-import convenience. In practice the preprocessor hoists `export { … } from "std::layout"` up with the imports, landing it *before* the `@module` doc comment and failing the "module doc must precede code" check. Since every consumer already imports `render` from `std::layout` directly, I dropped the re-export rather than work around the hoist. (Same approach will apply to `std::chart` in Part C.)

### Verification

- `make stdlib` → all modules compile (including `table.agency`, the trimmed `layout.agency`, and `policy.agency`).
- All table agency tests pass with **exact byte-identical output**: `table-data-form`, `table-block-form`, `table-render`, `table-no-mutate`, `table-validation`, `width-sizing` (14 assertions total).
- `pnpm run build` (`tsc`) clean; `make doc` regenerates cleanly.

### Next in series

- **Part C:** add the horizontal bar chart as a `barchart` handler + the `std::chart` module.


egonSchiele:
Thanks — all five addressed in 618a8640:

- **render import in examples (table.agency `@module` + generated table.md):** examples now `import { table } from "std::table"` and `import { render } from "std::layout"` separately, since `std::table` doesn't re-export `render`.
- **`ColumnSpec.width` "full" (both doc occurrences + the .md duplicates):** the doc said `"full"` is not allowed at the column level, but `table.ts` treats it as percent-like (`isPercentLike` + `full → 100`). Updated the doc to "`"full"` is treated as `"100%"`" to match behavior.

Regenerated `docs/site/stdlib/table.md`. table.agency recompiles; table tests still pass byte-identically.

Side note: PR #287 makes the `render` re-export work above an `@module` doc, so a future change could restore `export { render } from "std::layout"` here for one-import convenience — left out of this PR to keep it focused.

PR #287, opened 2026-06-14
Fix: allow re-exports above an @module doc comment

egonSchiele:
### Problem

A module that both re-exports a symbol **and** has an `@module` doc comment fails to compile:

```ts
import { x } from "std::foo"

/** @module
  ...
*/

export { x } from "std::foo"
```

```
Error: @module doc comment must appear before any code. Move it to the top
of the file or right after the imports.
```

This surfaced while extracting `std::table` (Part B of the chart work): `std::table` wanted to re-export `render` from `std::layout` for one-import convenience, but couldn't while also carrying an `@module` doc.

### Root cause

The check itself is fine — the problem is upstream. `resolveReExports` (which runs before the preprocessor) **desugars** a function/constant/type re-export into a synthesized `import` **plus a wrapper** (`buildFunctionWrapper` returns a `FunctionDefinition`, constants a `static const`, types a `type`). It then placed every synthesized node at the absolute front:

```ts
return { ...program, nodes: [...synthesized, ...kept] };
```

The wrapper is non-preamble code, so it landed *before* the `@module` comment in `kept`, tripping the preprocessor's "module doc must precede code" check. (The literal `exportFromStatement` is already gone by then, which is why relaxing the check to allow it wouldn't have worked.)

### Fix

Insert the synthesized nodes **after the file's leading preamble** (comments incl. `@module`, blank lines, imports) rather than at the absolute front. Imports stay at the top and `@module` keeps documenting the module:

```ts
let head = 0;
while (head < kept.length && LEADING_PREAMBLE_TYPES.has(kept[head].type)) head++;
return { ...program, nodes: [...kept.slice(0, head), ...synthesized, ...kept.slice(head)] };
```

### Verification

- **New unit test** in `resolveReExports.test.ts`: a leading `@module` comment + imports stay ahead of the synthesized import/wrapper.
- `resolveReExports.test.ts` (9) + `typescriptPreprocessor.core.test.ts` (22) → **31 pass**.
- **End-to-end**: the repro above now compiles cleanly.
- **No regression**: `tests/agency/reexport-node` (3/3) and `tests/agency/topsort/reexport-main` (1/1) still pass; full `make stdlib` compiles (including `cli.agency`'s re-export).
- `tsc --noEmit` and `lint:structure` clean.

### Follow-up

With this in, `std::table` (and `std::chart` in Part C) can re-export `render` from `std::layout` again if desired — though consumers importing `render` directly from `std::layout` also works fine.

### DIFF
--- a/packages/agency-lang/lib/preprocessors/resolveReExports.test.ts
+++ b/packages/agency-lang/lib/preprocessors/resolveReExports.test.ts
@@ -450,7 +450,7 @@ describe("resolveReExports: leading preamble ordering", () => {
     } as unknown as AgencyNode;
     const moduleDoc = {
       type: "multiLineComment",
-      value: "@module\n  docs",
+      content: "\n  module docs\n",
       isDoc: true,
       isModuleDoc: true,
     } as unknown as AgencyNode;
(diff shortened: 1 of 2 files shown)


egonSchiele:
Thanks — both addressed in fa35a901:

1. **Doc comment vs. leading preamble:** you're right, treating every `multiLineComment` as preamble could splice synthesized nodes between a regular `/** doc */` and the declaration it documents. The scan now stops at a declaration doc comment — only the `@module` doc (`isModuleDoc`) and plain non-doc block comments (`!isDoc`) are kept ahead; a regular doc comment ends the preamble. Added a regression test, and verified end-to-end that a file with a doc-commented function + a re-export still attaches the doc to the right declaration in `agency doc` output.
2. **Test fixture fidelity:** fixed the `@module` node to use the real `AgencyMultiLineComment` shape — `content` (not `value`), with `isDoc`/`isModuleDoc` flags and no literal `@module` in `content` (the parser strips the tag).

Verified: `resolveReExports` + preprocessor core tests pass (32), `tsc` and `lint:structure` clean.

PR #288, opened 2026-06-14
Part C: std::chart horizontal bar charts

egonSchiele:
## Part C of 3 — `std::chart` horizontal bar charts

The feature itself: a new `std::chart` module that renders **horizontal bar charts** in the terminal as a first-class layout node, building on the Part A handler refactor and the Part B `std::table` split.

```ts
import { barChart } from "std::chart"
import { render } from "std::layout"

const c = barChart(
  title: "Revenue by quarter",
  mode: "stacked",
  keys: [{ name: "web", color: "blue" }, { name: "app", color: "green" }],
  data: [
    { label: "Q1", values: [120, 80] },
    { label: "Q2", values: [98, 90] },
  ],
)
print(render(c))
```

### What's included

- **`lib/stdlib/layout/barchart.ts`** — a new `barchart` node handler registered in the layout `HANDLERS` table. Grouped + stacked modes, keyed series with auto color/symbol assignment, a legend, value labels, and **negative values via a zero baseline** (uniform-sign stacks enforced). Value→cell scaling + largest-remainder rounding adapted from chartscii (MIT). Coloring uses `lib/utils/termcolors.ts`.
- **`stdlib/chart.agency`** — the `std::chart` module: `barChart(...)` with both the data-param form (LLM-callable) and the block form (`as ch { ch.key(...); ch.bar(...) }`), plus an in-function docstring with `@param` lines for PFA-aware tool descriptions.
- Wiring: `barchart` added to the `NodeType` union and `HANDLERS`; chart helpers exposed on `_internal`.

### Render assembly (declarative)

The renderer keeps the chart-specific math in small pure helpers (`barCells`, `baselineColumn`, `stackSegments`, `drawBar`, `drawStack`) and assembles the chart by mapping data into rows, then combining three column `Block`s via the existing `pad`/`beside`/`above` algebra — the same approach `composeRow`/`composeTable` use, rather than hand-rolling padding.

### Notable fix along the way

The block builder exposed a runtime quirk: Agency leaves omitted optional params (e.g. `color`/`symbol` in `ch.key("web")`) as an internal sentinel rather than the declared `""` default, which is invisible to `printJSON` but a non-string in memory. The renderer now treats only a non-empty **string** as an explicit color/symbol and auto-assigns otherwise, so the block and data forms render identically. (No `render` re-export from `std::chart` — consumers import `render` from `std::layout`, since the re-export-above-`@module` fix isn't on `main` yet.)

### Verification

- **TS unit + render tests** (`barchart.test.ts`, 16): scaling, baseline, largest-remainder sum, color/symbol auto-assignment, legend, error cases (length mismatch, mixed-sign stack), and `_render` smoke tests. Full layout suite stays green (**200 passed**).
- **Agency execution test** (`tests/agency/chart/basic.agency`): block form and data-param form produce **byte-identical** output (2/2 exact-match).
- `tsc --noEmit`, `lint:structure`, and `make doc` (adds `chart.md`) all clean.

### DIFF
--- a/packages/agency-lang/docs/site/stdlib/chart.md
+++ b/packages/agency-lang/docs/site/stdlib/chart.md
@@ -1,0 +1,142 @@
+---
+name: "chart"
+---
+
+# chart
+
+## Module: std::chart
+
+  Horizontal bar charts for terminal output. A `barChart(...)` returns a
+  layout node; render it with `render` from `std::layout`, so a chart
+  nests inside `box` / `row` / `column`.
+
+  Two construction styles, same result:
+
+  - **Data form (LLM-callable, JSON-friendly):** pass `keys` and `data`
+    arrays. `data[i].values` aligns to `keys` by index.
+
+    ```ts
+    import { barChart } from "std::chart"
+    import { render } from "std::layout"
+
+    const c = barChart(
+      title: "Revenue by quarter",
+      mode: "stacked",
+      keys: [{ name: "web", color: "blue" }, { name: "app", color: "green" }],
+      data: [
+        { label: "Q1", values: [120, 80] },
+        { label: "Q2", values: [98, 90] },
+      ],
+    )
+    print(render(c))
+    ```
+
+  - **Block form (Agency-author ergonomics):**
+
+    ```ts
+    const c = barChart(title: "Revenue", mode: "stacked") as ch {
+      ch.key("web", color: "blue")
+      ch.key("app", color: "green")
+      ch.bar("Q1", 120, 80)
+      ch.bar("Q2", 98, 90)
+    }
+    ```
+
+  Keys get a distinct color and fill symbol automatically when omitted,
+  so charts stay readable even when color is disabled. Negative values
+  draw left of a zero baseline; stacked bars must be uniform-sign.
+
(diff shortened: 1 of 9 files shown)


egonSchiele:
Thanks — all three addressed in bfd004e1:

1. **`barArea` overflow:** now clamps to `Math.max(0, …)` instead of `1`, so an explicit `width` is never exceeded — a chart with no remaining room renders an empty bar region.
2. **`max` semantics:** a positive `max` now fixes the axis maximum regardless of the data's own max; values beyond it **saturate** at a full bar (clamped in both `drawBar` and the new `drawStack` clamp, so stacked bars don't overflow either). Negative / non-finite `max` is rejected early; `0`/unset still derives from the data. Docstring updated.
3. **`showValues` in grouped mode:** the renderer now prints **each** grouped bar's own value (one per key row) rather than a single max-magnitude value per category, matching the "beside each bar" docstring.

Verified: 200 layout/chart TS tests pass, the chart agency test stays byte-identical (2/2 exact-match), `tsc`/`lint`/`make doc` clean. Smoke-tested `max=60` saturation and per-bar grouped values.

PR #289, opened 2026-06-14
Add std::syntax diff and patch functions; remove std::fs::printDiff

egonSchiele:
## Summary

Adds two pure, string-returning functions to `std::syntax`, sharing one line-level hunk engine in `lib/utils/diff.ts`:

- **`diff(oldText, newText, ...)`** — a flexible human-readable diff. Line-aware (a single logical line is never split across multiple output lines), with inline word-highlighting on replacements. Flat named-arg options: `context`, `lineNumbers`, `colored`, `oldLabel`/`newLabel`, `ignoreWhitespace`, `hunkHeaders`, `summary`.
- **`patch(oldText, newText, filename, ...)`** — a standard unified diff that `std::fs::applyPatch` (or `git apply`) can apply. Supports file creation (`/dev/null`), deletion, and renames (`newFilename`).

Under the hood `lib/utils/diff.ts` is refactored into `computeHunks` → `renderDiff` (display) / `renderPatch` (unified diff). `formatDiff` stays as a thin shim, so the optimizer/test-runner/sourceMutator callers are unchanged.

Also: **removes the `std::fs::printDiff` tool** (and its `_printDiff` helper); callers now use `print(diff(old, new, colored: true))`. `edit`'s `printDiff: true` side effect now renders through the shared engine. `edit` remains the single public file-editing tool.

## Test Plan

- [x] Unit tests for `computeHunks`/`renderDiff`/`renderPatch` and back-compat `formatDiff` (`lib/utils/diff.test.ts`, 21 tests).
- [x] Existing optimizer callers unchanged (`reporter`/`sourceMutator`/`artifacts` tests green).
- [x] Pure Agency execution tests for `diff` (named-arg skipping) and `patch` output.
- [x] Agency-js round-trip: `applyPatch(patch(old, new, file))` reproduces `new` for single- and multi-hunk edits.
- [x] `std-fs-edit` / `std-fs-multiedit` still pass after the `printDiff` removal.
- [x] `pnpm run typecheck` clean.

### DIFF
--- a/packages/agency-lang/CHANGELOG.md
+++ b/packages/agency-lang/CHANGELOG.md
@@ -19,9 +19,6 @@
 - `std::args` for parsing command-line args.
 - `std::skills` understands the standard SKILL.md format.
 - Eval helpers (`evalJudge`, `evalExtract`), a `StatelogParser`, and `runAgencyAgent()` for running the built-in agents.
-- `std::syntax::diff(old, new, ...)` returns a flexible human-readable diff string (line numbers, context, color, labels, hunk headers, summary, ignore-whitespace).
-- `std::syntax::patch(old, new, filename, ...)` returns a standard unified diff that `std::fs::applyPatch` can apply (supports file creation, deletion, and renames).
-- Removed `std::fs::printDiff`; use `print(diff(old, new, colored: true))` instead.
 
 ### Agency Agent
 - Re-architected from a handoff model to a coordinator pattern.
--- a/packages/agency-lang/CHANGELOG.md
+++ b/packages/agency-lang/CHANGELOG.md
@@ -19,6 +19,9 @@
 - `std::args` for parsing command-line args.
 - `std::skills` understands the standard SKILL.md format.
 - Eval helpers (`evalJudge`, `evalExtract`), a `StatelogParser`, and `runAgencyAgent()` for running the built-in agents.
+- `std::syntax::diff(old, new, ...)` returns a flexible human-readable diff string (line numbers, context, color, labels, hunk headers, summary, ignore-whitespace).
+- `std::syntax::patch(old, new, filename, ...)` returns a standard unified diff that `std::fs::applyPatch` can apply (supports file creation, deletion, and renames).
+- Removed `std::fs::printDiff`; use `print(diff(old, new, colored: true))` instead.
 
 ### Agency Agent
 - Re-architected from a handoff model to a coordinator pattern.
--- a/packages/agency-lang/lib/stdlib/fs.ts
+++ b/packages/agency-lang/lib/stdlib/fs.ts
@@ -6,19 +6,9 @@ import diff_match_patch from "diff-match-patch";
 import { resolvePath } from "./resolvePath.js";
 import { resolveDir } from "./resolveDir.js";
 import { expandPath } from "./expandPath.js";
-import { formatDiff } from "../utils/diff.js";
+import { computeHunks, renderDiff } from "../utils/diff.js";
 
 export { resolvePath } from "./resolvePath.js";
-
-/**
- * Print a colored, line-based diff of `oldText` vs `newText` to
- * stdout. Deletions are red, insertions are green, unchanged context
- * is dimmed. Used both as a standalone Agency tool (`std::fs::printDiff`)
- * and as the side effect of `edit(..., printDiff: true)`.
- */
-export function _printDiff(oldText: string, newText: string): void {
-  console.log(formatDiff(oldText, newText));
-}
 
 export type MultiEdit = {
   oldText: string;
@@ -83,7 +73,7 @@ export async function _multiedit(
 
   await fs.writeFile(full, contents, "utf8");
   if (printDiff && original !== contents) {
-    _printDiff(original, contents);
+    console.log(renderDiff(computeHunks(original, contents, -1, false), { colored: true }));
   }
   return { replacements: total, path: filename, edits: edits.length };
 }
--- a/packages/agency-lang/lib/stdlib/fs.ts
+++ b/packages/agency-lang/lib/stdlib/fs.ts
@@ -6,7 +6,6 @@ import diff_match_patch from "diff-match-patch";
 import { resolvePath } from "./resolvePath.js";
 import { resolveDir } from "./resolveDir.js";
 import { expandPath } from "./expandPath.js";
-import { computeHunks, renderDiff } from "../utils/diff.js";
 
 export { resolvePath } from "./resolvePath.js";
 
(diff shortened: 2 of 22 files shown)


packages/agency-lang/CHANGELOG.md
```diff
@@ -19,6 +19,9 @@
 - `std::args` for parsing command-line args.
 - `std::skills` understands the standard SKILL.md format.
 - Eval helpers (`evalJudge`, `evalExtract`), a `StatelogParser`, and `runAgencyAgent()` for running the built-in agents.
+- `std::syntax::diff(old, new, ...)` returns a flexible human-readable diff string (line numbers, context, color, labels, hunk headers, summary, ignore-whitespace).
```
egonSchiele: don't update the changelog

packages/agency-lang/stdlib/syntax.agency
```diff
@@ -18,3 +20,58 @@ export def highlight(
...
+  """
+  Produce a human-readable diff of two strings and return it as a string.
+  By default returns a plain (uncolored) inline diff showing the full text
+  with changed words highlighted via `-`/`+` lines.
+
+  @param oldText - The original text
+  @param newText - The updated text
+  @param context - Unchanged lines to keep around each change; -1 means show the full text
+  @param lineNumbers - Prefix each line with its line number
+  @param colored - Emit ANSI colors (red deletions, green insertions) with inline word highlighting
```
egonSchiele: Check out some of the other standard library functions in agency that print colors. Is the parameter name `color` or `colored`? Let's keep it consistent.

packages/agency-lang/lib/stdlib/fs.ts
```diff
@@ -83,7 +73,7 @@ export async function _multiedit(
 
   await fs.writeFile(full, contents, "utf8");
   if (printDiff && original !== contents) {
-    _printDiff(original, contents);
+    console.log(renderDiff(computeHunks(original, contents, -1, false), { colored: true }));
```
egonSchiele: Let's remove the printDiff parameter from the edit function. Instead, the interrupt for edit should return the full before and after text as part of its data, and then the user can decide whether they want to use that data and call the `diff` function themselves. I think that will be a more flexible solution.

packages/agency-lang/lib/stdlib/syntax.ts
```diff
@@ -157,4 +158,41 @@ function mapMarkdownBlock(block: Block): Block {
...
+  newLabel: string,
+  ignoreWhitespace: boolean,
+  hunkHeaders: boolean,
+  summary: boolean,
+): string {
+  const hunks = computeHunks(oldText, newText, context, ignoreWhitespace);
+  return renderDiff(hunks, {
+    lineNumbers,
+    colored,
+    oldLabel: oldLabel || undefined,
```
egonSchiele: `|| undefined` isn't needed

packages/agency-lang/lib/utils/diff.ts
```diff
@@ -7,38 +7,286 @@ const DIFF_EQUAL = 0;
 
 const dmp = new DiffMatchPatch();
 
+type Diff = [number, string];
+
+export type DiffLine = {
+  kind: "context" | "delete" | "insert";
+  text: string;
+  oldNo: number | null;
```
egonSchiele: Use `oldNum` and `newNum` instead

packages/agency-lang/lib/utils/diff.ts
```diff
@@ -7,38 +7,286 @@ const DIFF_EQUAL = 0;
...
+function encodeLines(lines: string[], map: Map<string, number>, arr: string[]): string {
+  let s = "";
+  for (const line of lines) {
+    let idx = map.get(line);
+    if (idx === undefined) {
+      idx = arr.length;
+      arr.push(line);
+      map.set(line, idx);
+    }
+    s += String.fromCharCode(idx);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in ef810bf5. encodeLines now throws a clear error once unique lines exceed 0xFFFF instead of letting fromCharCode wrap. Added a unit test asserting computeHunks throws on >65535 unique lines.

packages/agency-lang/lib/utils/diff.ts
```diff
@@ -7,38 +7,286 @@ const DIFF_EQUAL = 0;
...
+}
+
+function makeHunk(lines: DiffLine[]): Hunk {
+  const oldNos = lines.filter((l) => l.oldNo !== null).map((l) => l.oldNo as number);
+  const newNos = lines.filter((l) => l.newNo !== null).map((l) => l.newNo as number);
+  return {
+    oldStart: oldNos.length ? oldNos[0] : 0,
+    oldLines: oldNos.length,
+    newStart: newNos.length ? newNos[0] : 0,
+    newLines: newNos.length,
```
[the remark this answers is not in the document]
egonSchiele: Fixed in ef810bf5. Insert-only and delete-only hunks now anchor their start to the preceding numbered line (new anchorsBefore helper) instead of defaulting to 0. A context:0 insertion now yields @@ -1,0 +2,1 @@ rather than @@ -0,0 ... @@. Added unit tests for both insert-only and delete-only anchoring.

packages/agency-lang/CHANGELOG.md
egonSchiele: Reverted in ef810bf5 -- CHANGELOG.md is back to the main version.

packages/agency-lang/stdlib/syntax.agency
egonSchiele: Renamed to color, typed auto | boolean defaulting to auto, matching std::layout::render. Shared the auto/TTY resolver as termcolors.autoUseColor and pointed render.ts at it too. Done in ef810bf5.

packages/agency-lang/lib/stdlib/fs.ts
egonSchiele: Done in ef810bf5. Removed the printDiff parameter from edit. The std::edit interrupt now carries the full before and after file contents (computed by a best-effort _previewEdit) so a handler can render a diff itself, e.g. print(diff(data.before, data.after, color: true)). Authoritative validation and the actual write stay in _multiedit after approval, so reject and error paths are unchanged.

packages/agency-lang/lib/stdlib/syntax.ts
egonSchiele: Removed in ef810bf5 -- empty strings are already falsy, so renderDiff skips the label without the guard.

packages/agency-lang/lib/utils/diff.ts
egonSchiele: Renamed oldNo/newNo to oldNum/newNum in ef810bf5.

PR #291, opened 2026-06-14
Generalize StatelogParser into a shared model; refactor logs viewer to use it

egonSchiele:
## Summary

Extracts the statelog parser out of `lib/eval/` into a general-purpose data layer at `lib/statelogParser.ts` (the MVC "model"), and refactors `agency logs view` to consume it through a `TreeNode` class that hides the parser.

- **`StatelogParser` (model)** — `StatelogParser.fromFile(path)` / `fromString(jsonl)`. Tolerant parsing collects `parseErrors()` instead of throwing; eval methods stay strict. Multi-trace with `trace(id)` / `onlyTrace()` scoping and a `TraceView` class. Owns the trace→span→event hierarchy (`getNodeById`, rolled-up metrics) ported from the viewer's `buildForest`. Query API: `llmCalls()` / `toolCalls()` (via `wireAccessors`), `events()` (Iterable), `lines()`. Payloads are fetched lazily via `eventOf(id)` so a future indexed/byte-offset backend can drop in unchanged (see `docs/dev/statelog-parser-memory-model.md`).
- **`TreeNode` (view)** — now a class. `TreeNode.forestFromLog(path)` / `forestFromString(jsonl)` build the display tree and **hide the parser entirely**; `node.event()` fetches a leaf's payload lazily. This replaces the old threaded `getEvent` accessor and the `parser.reload()` method (follow mode just re-calls the builder).
- **Viewer internals preserved** — the render loop, the pure input reducer, scroll math, and search match-logic are unchanged. `parse.ts` + `buildForest` are deleted (folded into the parser); plain-text summaries moved to `lib/statelog/summarize.ts` (shared with the viewer's styled variants).
- **Eval compat** — the 4 eval consumers switch from `new StatelogParser(path)` to `StatelogParser.fromFile(path)` and otherwise keep calling `.evalRecord()` / `.evalInputs()` etc.

Design + plan: `docs/superpowers/specs/2026-06-14-statelog-parser-refactor-design.md`, `docs/superpowers/plans/2026-06-14-statelog-parser-refactor.md`.

Note: node ids use hyphen form (`trace-<id>`, `evt-<lineNo>`, `<span_id>`) rather than the colon form in the spec, because the viewer's synthetic-row ids (`<leafId>:convo:…`) are parsed by splitting on the first colon — a colon in the leaf id would break that.

## Test Plan

- [x] `pnpm test:run` for the affected areas (`lib/logsViewer`, `lib/statelogParser.test.ts`, `lib/statelog`, `lib/eval`, `lib/stdlib/statelog.test.ts`, `lib/statelogClient.test.ts`) — 309 passing
- [x] `pnpm run typecheck` — clean
- [x] `pnpm run lint:structure` — clean
- [x] `forestFromLog` exercised against a real temp file; viewer integration (`run.test.ts`) exercises render/navigate/expand-JSON/search/parse-errors through the new `TreeNode` path
- [ ] CI runs the full agency test suite


packages/agency-lang/lib/logsViewer/treeNode.ts
```diff
@@ -0,0 +1,143 @@
...
+  static syntheticLine(
+    parent: TreeNode,
+    id: string,
+    nodeKind: NodeKind,
+    summary: string,
+  ): TreeNode {
+    return new TreeNode({ id, traceId: parent.traceId, parentId: parent.id, nodeKind, summary });
+  }
+
+  // The "raw data" toggle delegates event() back to the leaf it was spawned
```
egonSchiele, line 120: Tell me more about this. Does this save memory in some way? Otherwise, it feels like it would be better to mutate the current tree node instead of returning a new one -- maybe have an instance variable like `rawToggle`, which is a Boolean, and we lazily load the raw data on demand.

packages/agency-lang/lib/statelog/summarize.ts
```diff
@@ -0,0 +1,203 @@
...
+// *styled* (color-tagged) variants and imports the shared formatters here.
+import type { EventEnvelope } from "./wireTypes.js";
+
+export type SummaryMetrics = {
+  durationMs?: number;
+  tokens?: number;
+  cost?: number;
+};
+
+export function summarizeEvent(evt: EventEnvelope): string {
```
egonSchiele, line 14: why not have the TreeNode class own this logic?

egonSchiele: ...or at least expose this logic from there (all helper methods can stay in this file and out of the class)?

packages/agency-lang/lib/statelogParser.ts
```diff
@@ -0,0 +1,588 @@
...
+  }
+
+  normalized(): Normalized {
+    if (!this.normalizedCache) {
+      const events = this.parsed.events.map((p) => p.event);
+      assertSingleTrace(events);
+      this.normalizedCache = normalize(events);
+    }
+    return this.normalizedCache;
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 70a1ed1d. `normalized()` now calls a shared `assertNoParseErrors()` and throws the same `Malformed statelog on line N` error, matching `evalRecord()` and the pre-refactor strict behavior. Added a test (`normalized also throws on malformed input`).

packages/agency-lang/lib/statelogParser.ts
```diff
@@ -0,0 +1,588 @@
...
+
+export type ParseError = {
+  line: number;
+  kind: "invalid_json" | "missing_fields" | "unsupported_version";
+  detail: string;
+};
+
+// One kept event plus its 1-based source line. The line number is the stable
+// identity for event nodes (`evt:<lineNo>`) and is what `lines()` yields.
+type ParsedEvent = { event: EventEnvelope; lineNo: number };
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 70a1ed1d — corrected the comment to `evt-<lineNo>` to match the implementation and the `StatelogNode.id` doc.

packages/agency-lang/lib/statelog/summarize.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+export function fmtCost(c?: number): string {
+  if (c === undefined) return "?";
+  return `$${c.toFixed(3)}`;
+}
+
+function shortId(id?: string): string {
+  return (String(id) ?? "").slice(0, 6);
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 70a1ed1d. `shortId()` now returns `"?"` when the id is null/undefined (consistent with the other `?` placeholders in this file), so e.g. a `checkpointCreated` with no id reads `#?` instead of `#undefi`. Added a test.

packages/agency-lang/lib/logsViewer/treeNode.ts
egonSchiele: The memory angle is marginal, but there is a real reason it stays a separate node rather than a boolean on the leaf:

- The synthetic rows (`convoLine`, `jsonLine`, `rawDataToggle`) are all **ephemeral** — `flattenVisibleRows` regenerates them on every flatten and they are never stored in the persistent forest. So returning a fresh instance is not an allocation we are holding onto; it is transient by design, and `rawDataToggle` is consistent with how `convoLine`/`jsonLine` already work.
- `rawDataToggle` is a distinct, navigable **row**: the cursor can land on the `▶ raw data` line, and its expand state lives in `state.expanded` keyed by the row id — independently of whether the leaf is expanded. A boolean `rawToggle` on the leaf would collapse two independently-expandable rows into one and break that per-row cursor/expand model.
- The memory-relevant part of this change is actually `source` vs the old `event: leaf.event`: the toggle no longer copies the payload onto itself; it re-fetches lazily via the leaf (`this.source.event()`). So if anything this reduced what the toggle holds.

This synthetic-row design predates the PR (I only swapped the payload-carrying mechanism). Happy to file a follow-up if you want to rethink how expansion rows are modeled, but it felt out of scope for this refactor.

packages/agency-lang/lib/statelog/summarize.ts
egonSchiele: Layering: `summarize.ts` is consumed by **both** the model (`lib/statelogParser.ts` builds `node.summary` at parse time via `summarizeEvent`/`summarizeSpanText`/`summarizeTraceText`) and the viewer (`summary.ts` styled variants). `TreeNode` is a *view* class in `lib/logsViewer/`. If `TreeNode` owned this logic, the model would have to import a viewer class — inverting the model→view dependency this whole PR sets up. So it lives in `lib/statelog/` as a peer of `wireTypes.ts`/`wireAccessors.ts`, below both consumers.

egonSchiele: That part is already the case — the logic is a set of free functions exported from `summarize.ts` (`summarizeEvent`, `summarizeSpanText`, `summarizeTraceText`, `spanLabelOf`, …), not methods on any class. The only difference from your suggestion is the file lives in `lib/statelog/` rather than alongside `TreeNode`, for the dependency-direction reason in the sibling comment. `TreeNode` and `StatelogParser` both just import these free functions.

PR #293, opened 2026-06-15
Add /paste multi-line input to the line-mode REPL

egonSchiele:
## What

Adds a `/paste` command to the **line-mode REPL** (`std::cli`) for multi-line input. Type `/paste` (or pick it from the `/` palette) to open a multi-line editor:

```
> /paste
── paste mode · Enter: newline · Ctrl+D: submit · Ctrl+C: cancel ──
… first line
… second line
```

- **Enter** inserts a newline (does not submit)
- **Ctrl+D** submits the whole buffer as one message
- **Ctrl+C / Esc** cancels
- **Pasting** a multi-line block lands intact — since Enter no longer submits in this mode, the paste's interior newlines just become buffer lines

The single-line common case is unchanged.

## Why

Node `readline` is single-line: every Enter submits, and a pasted block's interior newlines each fire a `line` event, so multi-line input fragments into one turn per line. `/paste` is modeled on Node's REPL `.editor` (Ctrl+D to finish), which sidesteps this cleanly.

## How

Hijacks readline's `_ttyWrite` for the duration of the entry — the same seam the existing `/` slash-trigger uses — accumulating keystrokes into its own buffer, and restores the previous handler on exit. No bracketed-paste machinery needed (Enter-as-newline makes paste "just work").

It's a `std::cli` built-in, so any line-mode agent benefits; the agency agent surfaces it in its palette and `/help`.

## Files

- `lib/stdlib/cli.ts` — pure buffer ops (`pasteChar`/`pasteText`/`pasteBackspace`/`pasteJoin`), key classifier, and the `readMultiline` editor; `/paste` wired into the `_runLineRepl` loop (TTY-only).
- `lib/stdlib/cli.test.ts` — new: 18 tests (buffer ops incl. `\r\n` normalization, key classification, and `readMultiline` integration via simulated keystrokes).
- `stdlib/cli.agency` + `docs/site/stdlib/cli.md` — docstring note.
- `lib/agents/agency-agent/agent.agency` — `/paste` in palette + `/help`.

## Limitations (intentional v1)

- Append-only: backspace edits the current line but does not merge back across a newline (no cross-line cursor).
- TTY only: under piped input there are no keystrokes to drive it, so `/paste` falls through verbatim (non-interactive runs use one-shot mode anyway).

## Testing

`pnpm exec vitest run lib/stdlib/cli.test.ts` → 18 pass. Full `make` green; agent starts and `/paste` is present in the bundle.

### DIFF
--- a/packages/agency-lang/docs/site/stdlib/cli.md
+++ b/packages/agency-lang/docs/site/stdlib/cli.md
@@ -26,6 +26,12 @@ Line-mode REPL. Same call signature as `std::ui.repl` so swapping
   Exits on `onSubmit` returning `false`, on Ctrl+D (EOF) at the
   prompt, or on Ctrl+C at an idle prompt.
 
+  Built-in `/paste` (TTY only): opens a multi-line editor (à la Node's
+  REPL `.editor`). Enter inserts a newline, Ctrl+D submits the whole
+  buffer as one message, and Ctrl+C / Esc cancels. Because Enter no
+  longer submits while in this mode, pasting a multi-line block lands
+  intact instead of firing one turn per line.
+
   Round-one limitations (intentional, see the spec):
   - `status` is accepted for signature parity but not yet rendered.
   WL2 will turn it into a per-turn footer.
@@ -64,7 +70,7 @@ Line-mode REPL. Same call signature as `std::ui.repl` so swapping
 clearScreen()
 ```
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/cli.agency#L115))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/cli.agency#L121))
 
 ### termWidth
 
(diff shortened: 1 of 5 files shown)


egonSchiele:
Addressed the review comments in 011d828b:

- **`/paste` TTY gate** — now also requires `process.stdin.isTTY`, so a piped stdin (even with a TTY stdout) falls through and reaches the agent verbatim instead of entering the editor and hanging on absent keystrokes.
- **Public API surface** — the paste helpers (`pasteChar`/`pasteText`/`pasteBackspace`/`pasteJoin`/`classifyPasteKey`/`readMultiline`/`EMPTY_PASTE`) are no longer exported individually; they're now behind a single `export const _internal = { ... }` (matching `layout.ts`), and the test imports from `_internal`.
- **Leaking stdout spy** — added `afterEach(() => vi.restoreAllMocks())` so the `process.stdout.write` spy is always restored and can't leak into other tests.

All 19 unit tests pass; verified `/paste` + Ctrl+D still works against the real agent in a PTY.

PR #295, opened 2026-06-16
Effect sets and raises declarations (Phase 1)

egonSchiele:
## What

Phase 1 of making Agency's interrupt model statically checkable: functions and nodes can **declare the interrupt effects they may raise**, and the typechecker **verifies** the declaration. The goal is fast compile-time feedback (especially for LLM-written code) so effect mistakes surface before runtime rather than after.

## Highlights

- **`raises <...>` clause** on `def`s, `node`s, and function types. It's an upper bound: the checker infers every effect raised (directly or transitively) and reports any that exceed the declaration, with bespoke effect-aware messages.
  - `raises <>` → raises nothing (enforced). `raises <*>` / omitted → no bound.
  - Local handling does **not** exempt declaration — every handler in the chain still observes the effect (decision A).
- **`effectSet X = <...>`** declarations — reusable, composable (spread), and importable across modules. They lower onto string-literal **union types**, so import/export and assignability reuse existing machinery rather than new plumbing.
- **`raise` statement** — raise an interrupt as a statement (continues on approve, bails with a `failure` on reject); a clearer spelling of the `return interrupt(...)` idiom. Deliberately distinct from the JS-error `throw(...)` builtin.
- **Bare effects everywhere** — `raise deploy(...)` / `interrupt deploy(...)` no longer require `::` namespacing.

## Implementation notes

- New `resolveEffectSet` resolver and `checkRaisesDeclarations` diagnostic (wired into `TypeChecker.check()`), sitting alongside the existing interrupt diagnostics.
- `raise` lowers to the existing `interruptStatement` node, so codegen and effect inference are reused unchanged (a `viaRaise` marker only drives formatter output).
- `isEffectSet` threaded through `TypeAlias` → `TypeSymbol` → `TypeAliasEntry` so imported effect sets are recognized cross-module.
- Formatter emits `effectSet`, `raises`, `raise`, and `<...>` via a single shared helper.

## Deferred (documented, not bugs)

- Function-type `raises` is parsed/formatted but callback-raises compatibility is not yet checked (pairs with row polymorphism, deferred).
- A typo'd bare effect-set reference degrades to a literal effect label (cost of not requiring namespaces).

## Testing

- New unit tests for the literal/declaration/clause/`raise` parsers, the resolver, the subset diagnostic (incl. transitive, node-level, decision-A local handling, `<>`/`<*>`, multiple effects, false-positive guard), and cross-module import.
- `raise` runtime behavior covered by an Agency execution test (continues on approve, bails on reject).
- Full vitest unit suite green (**5794 passed**, including the 118-fixture typecheck integration — no false positives), `tsc --noEmit` clean.

Spec and plan included under `docs/superpowers/`; user-facing docs in `docs/site/guide/effects-and-raises.md`.


packages/agency-lang/CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
+## Unreleased
+
+### Language / Typechecker
+- **Effect sets and `raises` declarations.** Functions and nodes can now
```
egonSchiele: don't update changelog

PR #297, opened 2026-06-16
Characterize memory/setMemoryId behavior inside forks (+ getMemoryId)

egonSchiele:
## Summary

Characterization tests that pin down how memory behaves across `fork` branches today. We were unsure of the actual semantics (forks share `ctx` but get their own stacks, and memory helpers split between the two), so these tests establish ground truth before any change. **They document current behavior — including the surprising parts — they do not endorse it.**

Also adds `getMemoryId()` to `std::memory` (symmetric with `setMemoryId`) as the no-LLM observable the `setMemoryId` tests need.

## What the tests show

| Test | Result | Meaning |
|---|---|---|
| `enable-inherits` | `[true,true,true]` | A branch **inherits** top-level `enableMemory` (`isMemoryActive()` true in each branch). |
| `branch-enable-invisible` | `[false,false,false,false]` | ⚠️ `enableMemory` **inside** a branch is **not observable** in that branch — the frame lands on the branch's own stack, but `isMemoryActive`/manager lookup reads the top-level stack. |
| `branch-enable-no-leak` | `false` | A branch-local enable does not leak to the parent after join. |
| `setid-inherits` | `["base","base","base"]` | Branches **inherit** the top-level memory id. |
| `setid-leaks` | `"branch"` | ⚠️ `setMemoryId` **inside** a branch **leaks** to the parent (it writes the shared top-level scope; with many branches it also races). |

The two ⚠️ rows are the surprising behaviors: memory frame/id reads resolve through the top-level stack (`ctx.stateStack`), so a branch can't establish its own memory scope, and a branch-local `setMemoryId` mutates shared state.

## Why

Captured ahead of a planned move to **branch-scoped** memory/model config (so `setModel`/`setMemoryId`/`memory{}` inside a fork affect only that branch, inheriting the parent's default at fork time). These tests will flip from "documents the quirk" to "guards the fix" when that lands.

## Testing

- 5 new agency execution tests under `tests/agency/memory-fork/` (no LLM calls).
- Existing memory tests (`tests/agency/memory-code/*`) and 97 memory unit tests still pass.
- `getMemoryId` reference doc regenerated via `agency doc`.

### DIFF
--- a/packages/agency-lang/docs/site/stdlib/memory.md
+++ b/packages/agency-lang/docs/site/stdlib/memory.md
@@ -18,7 +18,7 @@ export type MemoryConfig = {
 }
 ```
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/memory.agency#L39))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/memory.agency#L40))
 
 ## Functions
 
@@ -40,7 +40,7 @@ Return `true` iff there is an active memory frame on the current
 
 **Returns:** `boolean`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/memory.agency#L47))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/memory.agency#L48))
 
 ### setMemoryId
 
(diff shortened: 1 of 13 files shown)


egonSchiele:
Superseded by #298, which contains these characterization tests plus the actual branch-scoping fix (the two 'broken' tests are flipped there to assert the corrected behavior).

PR #298, opened 2026-06-16
Branch-scoped memory: fix setMemoryId/enableMemory/memory in forks (+ characterization tests)

egonSchiele:
## Summary

Makes memory **branch-scoped** in `fork`/`race`/llm-tool branches, the way globals and threads already are. A branch inherits the memory config active at fork time, but its own `enableMemory`/`disableMemory`/`setMemoryId` affect only that branch — no leakage to siblings or the parent, and no cross-branch races.

Supersedes #297 (which only characterized the old behavior). This PR contains those characterization tests **plus** the fix, with the two "broken" tests flipped to assert the corrected behavior.

## The bug (previously)

Memory frame/id reads resolved through `ctx.stateStack` (the top-level stack), which is never swapped per branch — so a branch couldn't establish its own scope:

- `enableMemory` inside a branch was **invisible** to that branch (the frame landed on the branch's own stack, but reads went to the top-level stack).
- `setMemoryId` inside a branch **leaked** to the shared top-level scope (and raced across siblings).

Meanwhile `enableMemory`/`setMemoryId` *writes* already targeted the branch stack — so reads and writes disagreed inside a branch.

## The fix

Mirrors how guards/cost already inherit into branches:

- `getActiveMemoryManager` and the `memoryIdRef` resolve against the **active branch stack** (`agencyStore.getStore()?.stack ?? this.stateStack`) instead of `ctx.stateStack`. The id ref reads the active stack dynamically per get/set, so the cached, per-dir manager serves each concurrent branch its own id.
- `StateStack.inheritMemoryFrom` seeds a fresh branch's memory id + frame stack from the parent (string copy + shallow frame-array copy, so branch push/pop/setId can't mutate the parent's).
- `runBatch.inheritBranchMemory` runs that for every branch with value-based fresh detection (mirrors `seedBranchCost`), so a resumed branch keeps its serialized memory rather than being re-seeded.

Also adds `getMemoryId()` to `std::memory` (symmetric with `setMemoryId`).

## Behavior

| Scenario | before | after |
|---|---|---|
| top-level enable → branch | inherits | inherits |
| `enableMemory` inside branch | invisible | visible in that branch |
| branch enable → parent | no leak | no leak |
| top-level `setMemoryId` → branch | inherits | inherits |
| `setMemoryId` inside branch | **leaks to parent** | branch-local |
| `fork as i { setMemoryId(i) }` | racy/shared | each branch its own |

## Testing

- 6 agency tests under `tests/agency/memory-fork/` (inherit, branch-scoped enable, no-leak, per-branch id).
- An agency-js resume test: branch-local memory survives a real interrupt/resume (does not get re-seeded from the parent).
- Full unit suite (5811) passes; fork isolation/resume/race and memory suites still green.

## Follow-up

`setModel`/`setLlmOptions` (the user-selectable-models work) will reuse this infrastructure (`stack.other.*` + active-stack reads + runBatch seeding) to be branch-scoped and resume-durable, instead of mutating `ctx.smoltalkDefaults`.


packages/agency-lang/lib/runtime/state/context.ts
```diff
@@ -349,17 +350,24 @@ export class RuntimeContext<T> {
...
+    // Resolve memory against the ACTIVE branch stack, not the top-level
+    // `this.stateStack`. Inside a fork/race/tool branch the active stack
+    // is that branch's own slice (seeded from the parent at fork time via
+    // `inheritMemoryFrom`), so `enableMemory`/`setMemoryId` inside a branch
+    // are visible to that branch and don't leak to siblings/parent. At the
+    // top level (and outside any ALS frame) this is `this.stateStack`.
+    const stack = agencyStore.getStore()?.stack ?? this.stateStack;
+    if (!stack) return undefined;
+    let frame = stack.activeMemoryFrame();
+    if (!frame && this.jsonMemoryConfig && !stack.hasMemoryFrameStack()) {
```
egonSchiele: when would this trigger? This feels like it could lead to subtle bugs.

egonSchiele: Good catch — it's back-compat for resuming a checkpoint written before memory frames existed: `StateStack.fromJSON` has no `other.memoryFrames` key and `restoreState` doesn't re-seed, so this re-establishes the agency.json bottom frame on resume. That's a TOP-LEVEL concern, and you're right that running it against the active (possibly branch) stack was a subtle-bug risk.

Fixed in 49bd8f06: guarded with `stack === this.stateStack` (true at top level / bootstrap, false in any branch — see node.ts:78-85), so it can never push onto a branch stack. Branches get memory via inheritMemoryFrom at fork time, so a branch with jsonMemoryConfig set already has a frame array and wouldn't reach it anyway; the guard makes that impossible by construction. Top-level behavior is unchanged (110 memory unit tests + the fork/resume tests still pass).

packages/agency-lang/lib/runtime/state/context.ts
```diff
@@ -381,17 +403,21 @@ export class RuntimeContext<T> {
...
-        // by `restoreState`).
+        // on `<stack>.other.memoryId` and persists across frame
+        // pushes/pops. Read the ACTIVE branch stack dynamically on each
+        // access (not a captured one): this manager is cached per
+        // configKey and shared across concurrent branches, so each
+        // branch's get/set must resolve to ITS own stack. Falls back to
+        // `this.stateStack` outside any ALS frame.
         get: () => {
-          const id = this.stateStack?.other?.memoryId;
+          const s = agencyStore.getStore()?.stack ?? this.stateStack;
```
egonSchiele, line 413: why not create a helper func for this pattern?

PR #301, opened 2026-06-16
User-selectable models: std::llm (setModel/setLlmOptions/pickProvider) + agent CLI flags

egonSchiele:
## Summary

Lets users pick which model `llm()` calls use — for the Agency agent and any Agency program — via a new `std::llm` module. Model defaults are **branch-scoped** (a `fork`/`race` branch inherits the run-wide default, but its own `setModel` stays local) and **resume-durable** (they ride the serialized `stack.other`). Builds on the branch-scoped-memory work (#298).

## `std::llm`

- **`setLlmOptions(opts)` / `setModel(name)`** — write the active branch stack's `other.llmDefaults`. `runPrompt` merges these **between** the baked `smoltalkDefaults` and per-call `llm({...})` options (precedence: baked `agency.json` < runtime `setLlmOptions` < per-call), and into the tool-result-cap resolution.
- **`pickProvider(order)`** — returns a `Result` naming the first provider whose API-key env var is set (`anthropic`→`ANTHROPIC_API_KEY`, `google`→`GEMINI_API_KEY`, `openai`→`OPENAI_API_KEY`), else a `failure` listing the checked vars.

Branch inheritance/isolation reuses #298's seam: `StateStack.inheritMemoryFrom` now also seeds `other.llmDefaults`, and `runBatch.inheritBranchMemory`'s fresh-detection includes it (resume-safe).

## Config

`anthropicApiKey` plumbed through `agency.json` (`client.anthropicApiKey`) and emitted into `smoltalkDefaults` (literal when set, else `ANTHROPIC_API_KEY` env), parallel to OpenAI/Google.

## Agent

- `--model` / `--fastmodel` / `--slowmodel` flags.
- A provider→models table + `setSlowModel`/`getSlowModel`/`getSlowProvider` in `shared.agency`.
- `main()` resolves: explicit flags win; otherwise `pickProvider` detects a provider and uses its defaults. The fast model becomes the run-wide default via `setModel`; the slow model is read by the oracle/explorer deep-reasoning calls (replacing hardcoded `gpt-5.5`/`openai-responses`).

## Testing

- Unit: `lib/stdlib/llm.test.ts` (setLlmOptions merge), `lib/backends/anthropicApiKey.codegen.test.ts` (baked / env fallback).
- agency-js: `llm-pick-provider` (env detection + failure), `llm-set-model` (e2e — dispatched model = `setModel` value, via a capturing client), `llm-fork-scoped` (each branch dispatches its own model; the parent's post-fork call is unaffected — no leak/race).
- Manual: `--help` lists the flags; no-key run errors cleanly via `pickProvider`; `--model X` alone skips provider detection.
- No regressions across the memory-fork and runtime unit suites.

## Notes for review

- **Flag names are `--fastmodel`/`--slowmodel`, not `--fast-model`/`--fastModel`.** Agency object literals reject hyphenated/quoted keys and `std::args` rejects uppercase flag names, so kebab/camel forms are unexpressible; a single lowercase token is the closest valid form. Easy to switch to `--fast`/`--slow` if preferred.
- Spotted an agency-semantics gotcha while wiring `main()`: an absent string flag reads as `undefined`, and `undefined == null` is **false** in agency (though `undefined ?? x` does fall back). The resolution normalizes flags via `?? ""` and compares strings to avoid it.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -16,7 +16,9 @@ import { highlight, diff } from "std::syntax"
 import { cwd, env, isTTY, readStdin, setTitle } from "std::system"
 import { getCost, getTokens, systemMessage } from "std::thread"
 import { chooseOption, ChoiceItem } from "std::ui"
+import { setModel, pickProvider } from "std::llm"
 
+import { MODELS, setSlowModel } from "./shared.agency"
 import { figs, title } from "./images/images.agency"
 import { POLICY_PATH, HISTORY_PATH, ALWAYS_FIELDS } from "./lib/config.agency"
 import {
@@ -588,6 +590,18 @@ node main() {
       verbose: {
         type: "boolean",
         description: "Echo tool-call starts to stdout in non-interactive mode"
+      },
+      model: {
+        type: "string",
+        description: "Model for all LLM calls (overrides provider auto-detection)"
+      },
+      fastmodel: {
+        type: "string",
+        description: "Model for ordinary work (default: per detected provider)"
+      },
+      slowmodel: {
+        type: "string",
+        description: "Model for deep reasoning, e.g. the oracle/explorer subagents"
       }
     }
   },
@@ -598,6 +612,48 @@ node main() {
   }
   if (args.flags.verbose) {
     VERBOSE = true
+  }
+
+  // Resolve which models to use. Explicit flags win; otherwise detect a
+  // provider from API-key env vars and use that provider's defaults. The
+  // fast model becomes the run-wide default via `setModel`; the slow
+  // model is read by the deep-reasoning subagents (oracle, explorer).
+  let provider = "openai"
+  // Absent string flags read as `undefined` (NOT null), and agency's
+  // `== null` does not match `undefined`. Normalize via `??` (which does
+  // treat `undefined` as nullish) to "" and compare strings.
+  const modelFlag = args.flags.model ?? ""
+  const fastFlag = args.flags.fastmodel ?? ""
+  const slowFlag = args.flags.slowmodel ?? ""
+  // A provider is only needed when fast OR slow can't be derived from
+  // flags (so `--model X` alone never requires an API key / detection).
+  const needFast = fastFlag == "" && modelFlag == ""
+  const needSlow = slowFlag == "" && modelFlag == ""
+  if (needFast || needSlow) {
+    const picked = pickProvider()
+    if (isSuccess(picked)) {
+      provider = picked.value
+    } else {
+      print(picked.error)
+      process.exit(1)
+    }
+  }
+  let fast = MODELS[provider].fast
+  if (fastFlag != "") {
+    fast = fastFlag
+  } else if (modelFlag != "") {
+    fast = modelFlag
+  }
+  setModel(fast)
+  // Slow model: an explicit flag wins (with empty provider so smoltalk
+  // resolves the provider from the model name); otherwise use the
+  // detected provider's slow model + its pinned provider.
+  if (slowFlag != "") {
+    setSlowModel(slowFlag, "")
+  } else if (modelFlag != "") {
+    setSlowModel(modelFlag, "")
+  } else {
+    setSlowModel(MODELS[provider].slow, MODELS[provider].slowProvider)
   }
 
   // Positional args (everything after flags) are joined with spaces to
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -16,9 +16,8 @@ import { highlight, diff } from "std::syntax"
 import { cwd, env, isTTY, readStdin, setTitle } from "std::system"
 import { getCost, getTokens, systemMessage } from "std::thread"
 import { chooseOption, ChoiceItem } from "std::ui"
-import { setModel, pickProvider } from "std::llm"
 
-import { MODELS, setSlowModel } from "./shared.agency"
+import { configureModels } from "./shared.agency"
 import { figs, title } from "./images/images.agency"
 import { POLICY_PATH, HISTORY_PATH, ALWAYS_FIELDS } from "./lib/config.agency"
 import {
@@ -602,6 +601,10 @@ node main() {
       slowmodel: {
         type: "string",
         description: "Model for deep reasoning, e.g. the oracle/explorer subagents"
+      },
+      provider: {
+        type: "string",
+        description: "Force the LLM provider (use when a model name doesn't imply one)"
       }
     }
   },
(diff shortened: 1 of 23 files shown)


packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -600,6 +614,48 @@ node main() {
     VERBOSE = true
   }
 
+  // Resolve which models to use. Explicit flags win; otherwise detect a
```
egonSchiele, line 620: move this logic to a shared module too

packages/agency-lang/lib/agents/agency-agent/shared.agency
```diff
@@ -51,3 +51,36 @@ static const _memoryReady = enableMemory({
...
+  anthropic: { fast: "claude-haiku-4-5-20251001", slow: "claude-opus-4-8", slowProvider: "anthropic" },
+  google: { fast: "gemini-2.5-flash", slow: "gemini-2.5-pro", slowProvider: "google" }
+}
+
+// The slow ("deep reasoning") model + its provider, set once in main()
+// from the CLI flags / detected provider. The deep subagents (oracle,
+// explorer) read these instead of hardcoding a model. An empty
+// slowProvider means "let smoltalk resolve the provider from the model
+// name" — used when the user overrides --slow-model.
+let _slowModel: string = "gpt-5.5"
```
egonSchiele: no need to prepend underscore for these vars

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -385,15 +385,35 @@ export async function runPrompt(args: {
...
   };
-  const clientConfig = ctx.getSmoltalkConfig(restClientConfig);
+
+  // Run-wide LLM defaults set at runtime via `std::llm`
+  // (setLlmOptions/setModel) live on the ACTIVE branch stack's
+  // `other.llmDefaults` (branch-scoped, serialized; seeded from the
+  // parent at fork time). Layer them BETWEEN the baked
+  // `smoltalkDefaults` and the per-call options, so precedence is
+  // baked agency.json < stack defaults < per-call `llm({...})`.
+  const stackDefaults = (stateStack?.other?.llmDefaults ?? {}) as {
```
egonSchiele: is the `as` cast needed here? Can you just do `Partial<whatever>`?

packages/agency-lang/stdlib/llm.agency
```diff
@@ -0,0 +1,91 @@
...
+import { _setLlmOptions } from "agency-lang/stdlib-lib/llm.js"
+import { env } from "std::system"
+
+// LLM defaults that can be set at runtime. Mirrors a subset of the
+// per-call `llm()` options. Every field is optional; only provided
+// fields are changed.
+export type LlmDefaults = {
+  model?: string;
+  temperature?: number;
+  reasoningEffort?: string;
```
egonSchiele: should this be an enum?

packages/agency-lang/stdlib/llm.agency
```diff
@@ -0,0 +1,91 @@
...
+  maxTokens?: number;
+  maxToolResultChars?: number
+}
+
+export def setLlmOptions(opts: LlmDefaults) {
```
egonSchiele, line 29: use doc comments and module doc comments. Doc string values are sent as tool descriptions to alms; does that change what you add here?  eg `Branch-scoped: a `fork`/`race` branch inherits the defaults active at fork time` -- an LLM doesn't need to know these details, just what the func is for, how to use it, what the params are.

packages/agency-lang/stdlib/llm.agency
```diff
@@ -0,0 +1,91 @@
...
+  if (provider == "google") {
+    return "GEMINI_API_KEY"
+  }
+  if (provider == "openai") {
+    return "OPENAI_API_KEY"
+  }
+  return ""
+}
+
+export def pickProvider(order: string[] = ["anthropic", "google", "openai"]): Result {
```
egonSchiele: returns `Result<string, string>`

packages/agency-lang/tests/agency-js/llm-fork-scoped/agent.agency
```diff
@@ -0,0 +1,14 @@
+import { setModel } from "std::llm"
+
+// Each branch sets its own model and dispatches with it; the parent's
+// model is unaffected after the fork joins (no leak, no cross-branch
+// races). The parent's llm() call runs after the fork, so it is the
+// last LLM call.
+node main() {
+  setModel("base-model")
+  fork(["a", "b"]) as item {
+    setModel("model-${item}")
```
egonSchiele, line 10: we should also allow users to set a provider. Useful in cases where we can't derive the provider name from the model.

egonSchiele:
Addressed all review comments in f2758d06:

**@egonSchiele**
- `agent.agency` "move this logic to a shared module" — moved the whole resolution into `shared.configureModels(model, fast, slow, provider)`; `main()` now just normalizes the flags (`?? ""`) and calls it.
- `shared.agency` "no need to prepend underscore" — `_slowModel`/`_slowProvider` → `slowModel`/`slowProvider`.
- `prompt.ts` "is the `as` cast needed?" — dropped it; now `const stackDefaults: Partial<LlmDefaults> = stateStack?.other?.llmDefaults ?? {}`.
- `llm.agency` "should this be an enum?" — `reasoningEffort` is now `"low" | "medium" | "high"` (in both the agency type and the TS `LlmDefaults`).
- `llm.agency` "use doc comments / docstrings go to LLMs" — added a `@module` doc comment and `/** */` doc comments for maintainer detail; trimmed the `"""docstrings"""` to just what/how/params (removed the branch-scoping internals an LLM doesn't need).
- `llm.agency` "returns `Result<string, string>`" — done.
- `llm-fork-scoped` "allow users to set a provider" — added `provider` to `LlmDefaults` (so `setLlmOptions({ provider })` works) and a `--provider` agent flag; `configureModels` forces it for both the fast and slow model when given (empty string → smoltalk derives from the model name).

**Copilot**
- `pickProvider` empty/all-unrecognized `order` — now returns an actionable failure ("no recognized providers given (expected anthropic, google, or openai)") instead of `Set one of: `; added an `emptyOrder` case to the agency-js test.
- stale `--fast-model`/`--slow-model` comments — updated to `--fastmodel`/`--slowmodel`.
- `llm.test.ts` `runInTestContext` suggestion — kept the lightweight `agencyStore.run({ stack })` seeding: `_setLlmOptions` only reads `stack`, and `runInTestContext` requires a full `RuntimeContext` (it dereferences `ctx.globals`), which would add construction overhead the test doesn't exercise. Happy to switch if you'd prefer the uniformity.

Note on flag naming: the flags are `--model`/`--fastmodel`/`--slowmodel`/`--provider` (single lowercase tokens) because agency object literals reject hyphenated/quoted keys and `std::args` rejects uppercase — so `--fast-model`/`--fastModel` can't be expressed. Let me know if you'd rather have `--fast`/`--slow`.

Verified: `make` clean; 6 unit + 4 agency-js tests pass; `--help` lists all four flags; no-key run errors cleanly.

PR #305, opened 2026-06-18
Pluggable optimizer framework — Phase 3A (agent-running infra)

egonSchiele:
## Summary

Phase 3A of the pluggable optimizer framework: the **agent-running infrastructure**, added as
additive, fully-tested services. **Nothing is wired into the live optimizer yet — zero behavior
change** (greedy still uses the Phase 1 pairwise path). The greedy → pointwise migration is
Phase 3B.

- **Refactor `runAgencyNode`** (`lib/cli/util.ts`) — extracts the general "run an `.agency` node" core (compile/resolve → render evaluate script → spawn `node` → parse results) from `executeNodeAsync`. The test-LLM coupling (`AGENCY_USE_TEST_LLM_PROVIDER`/`useTestLLMProvider` → `AGENCY_LLM_MOCKS`) now lives only in `executeNodeAsync`, which delegates to the core. Existing callers are unchanged.
- **`AgencyRunner`** (class) — `run()` executes the agent under test; `runStructured()` runs a judge/proposer node and zod-validates its return. A `runNode` seam makes it unit-testable.
- **`GraderInput += runAgency`** — graders can now invoke a judge `.agency` file.
- **`LlmJudge`** grader — scalar or binary verdicts from a judge `.agency` file.
- **`EvalCache`** — memoizes one `AgentRun` per `(workspace, input)` (the "run once, grade many" guarantee).
- **`WorkspaceManager`** — per-iteration workspace dirs (fork/read/write/applyFiles).

## Test Plan

- [x] `pnpm test:run lib/optimize lib/cli/util.test.ts lib/cli/eval` — 194 passed across 24 files.
- [x] `pnpm typecheck` — clean.
- [x] `pnpm run lint:structure` — clean.
- [x] Zero behavior change: the Phase 1 pairwise-greedy CLI/loop tests are unchanged and green; the refactor's consumers (mutator, eval runner) still pass through `executeNodeAsync` → `runAgencyNode`.

### DIFF
--- a/packages/agency-lang/lib/optimize/evalCache.ts
+++ b/packages/agency-lang/lib/optimize/evalCache.ts
@@ -1,14 +1,19 @@
 import type { AgentRun } from "./grading/types.js";
 
-/** Memoizes one AgentRun per (workspaceKey, inputId). Null-prototype: keys derive from user inputs. */
+/**
+ * Memoizes one AgentRun per (workspaceKey, inputId). A nested store avoids any
+ * separator-collision between the two components (a flat `${ws} ${id}` or
+ * `${ws}-${id}` key can collide, e.g. workspace keys are themselves `ws-N`).
+ * Null-prototype maps since both keys can derive from user input.
+ */
 export class EvalCache {
-  private readonly runs: Record<string, Promise<AgentRun>> = Object.create(null);
+  private readonly runs: Record<string, Record<string, Promise<AgentRun>>> = Object.create(null);
 
   get(workspaceKey: string, inputId: string, produce: () => Promise<AgentRun>): Promise<AgentRun> {
-    const key = `${workspaceKey} ${inputId}`;
-    if (!Object.hasOwn(this.runs, key)) {
-      this.runs[key] = produce();
+    const byInput = (this.runs[workspaceKey] ??= Object.create(null));
+    if (!Object.hasOwn(byInput, inputId)) {
+      byInput[inputId] = produce();
     }
-    return this.runs[key];
+    return byInput[inputId];
   }
 }
--- a/packages/agency-lang/lib/optimize/evalCache.ts
+++ b/packages/agency-lang/lib/optimize/evalCache.ts
@@ -1,0 +1,14 @@
+import type { AgentRun } from "./grading/types.js";
+
+/** Memoizes one AgentRun per (workspaceKey, inputId). Null-prototype: keys derive from user inputs. */
+export class EvalCache {
+  private readonly runs: Record<string, Promise<AgentRun>> = Object.create(null);
+
+  get(workspaceKey: string, inputId: string, produce: () => Promise<AgentRun>): Promise<AgentRun> {
+    const key = `${workspaceKey} ${inputId}`;
+    if (!Object.hasOwn(this.runs, key)) {
+      this.runs[key] = produce();
+    }
+    return this.runs[key];
+  }
+}
--- a/packages/agency-lang/lib/optimize/grading/agencyRunner.ts
+++ b/packages/agency-lang/lib/optimize/grading/agencyRunner.ts
@@ -33,14 +33,14 @@ export class AgencyRunner {
     private readonly runNode: NodeRunner = defaultRunner,
   ) {}
 
-  /** Run an agent node and return its raw value. */
-  async run(agencyFile: string, nodeName: string, args: Record<string, JSON>): Promise<JSON> {
+  /** Run an agent node and return its raw value. `args` are positional, in node-parameter order. */
+  async run(agencyFile: string, nodeName: string, args: JSON[]): Promise<JSON> {
     const { data } = await this.exec(agencyFile, nodeName, args);
     return data as JSON;
   }
 
   /** Run a judge/proposer node and validate its structured return against a schema. */
-  async runStructured<T>(agencyFile: string, nodeName: string, args: Record<string, JSON>, schema: ZodSchema<T>): Promise<T> {
+  async runStructured<T>(agencyFile: string, nodeName: string, args: JSON[], schema: ZodSchema<T>): Promise<T> {
     const { data } = await this.exec(agencyFile, nodeName, args);
     const parsed = schema.safeParse(data);
     if (!parsed.success) {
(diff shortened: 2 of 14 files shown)


packages/agency-lang/lib/optimize/grading/llmJudge.test.ts
```diff
@@ -0,0 +1,26 @@
+import { describe, expect, it } from "vitest";
+
+import { AgencyRunner } from "./agencyRunner.js";
+import { LlmJudge } from "./llmJudge.js";
+import type { GraderInput, Input } from "./types.js";
+
+const gi = (verdict: { score?: number; pass?: boolean; reasoning: string }): GraderInput => {
```
egonSchiele: `gi` is a terrible name

packages/agency-lang/lib/optimize/grading/llmJudge.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/** Grades an output by running a judge .agency file and reading its structured verdict. */
+export class LlmJudge extends BaseGrader {
+  protected readonly defaultName = "llm-judge";
+  constructor(protected readonly options: LlmJudgeOptions) {
+    super(options);
+  }
+
+  protected async _run({ input, run, runAgency }: GraderInput): Promise<Grade> {
+    const goal = String(getPath(input, this.options.goalPath ?? ["metadata", "goal"]) ?? "");
```
egonSchiele: I'm curious how effective you think an LLM judge would be with no goal. What do you expect the LLM judge to do in that scenario?

packages/agency-lang/lib/optimize/evalCache.ts
```diff
@@ -0,0 +1,14 @@
+import type { AgentRun } from "./grading/types.js";
+
+/** Memoizes one AgentRun per (workspaceKey, inputId). Null-prototype: keys derive from user inputs. */
+export class EvalCache {
+  private readonly runs: Record<string, Promise<AgentRun>> = Object.create(null);
+
+  get(workspaceKey: string, inputId: string, produce: () => Promise<AgentRun>): Promise<AgentRun> {
+    const key = `${workspaceKey} ${inputId}`;
```
egonSchiele: replace the space with a hyphen

packages/agency-lang/lib/optimize/grading/llmJudge.ts
egonSchiele: You're right — it would do nothing useful. With an empty goal the judge agent is asked to grade the output against no criterion, so its verdict is meaningless (and worse, it looks like a real score). I treated a missing goal as silent-success-ish, which is exactly the wrong default.

Fixed in bf0d4d65: `_run` now throws a clear error when the goal doesn't resolve on the input (undefined/null/empty-after-trim) — `"<name>: no goal found at <path> on input <id>; an LLM judge needs a goal."` — with a test. So a misconfigured judge fails loudly at the input rather than emitting a bogus score.

packages/agency-lang/lib/optimize/grading/llmJudge.test.ts
egonSchiele: Renamed the gi helper to graderInput across the grading tests in bf0d4d65.

packages/agency-lang/lib/optimize/workspace.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  read(ws: Workspace, relPath: string): string {
+    return fs.readFileSync(path.join(ws.dir, relPath), "utf8");
+  }
+
+  write(ws: Workspace, relPath: string, content: string): void {
+    const abs = path.join(ws.dir, relPath);
+    fs.mkdirSync(path.dirname(abs), { recursive: true });
+    fs.writeFileSync(abs, content);
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in bf0d4d65: read/write now resolve via a private resolveWithin() that throws if the resolved path escapes the workspace root (covers .. and absolute paths). Added a traversal test.

packages/agency-lang/lib/optimize/evalCache.ts
```diff
@@ -0,0 +1,14 @@
...
+    if (!Object.hasOwn(this.runs, key)) {
+      this.runs[key] = produce();
+    }
+    return this.runs[key];
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in bf0d4d65: EvalCache now uses a nested store keyed by workspace then input — no separator at all, so the two components cannot collide. Added a regression test for the boundary case.

packages/agency-lang/lib/optimize/grading/agencyRunner.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  private async exec(agencyFile: string, nodeName: string, args: Record<string, JSON>): Promise<{ data: unknown }> {
+    const scratchDir = fs.mkdtempSync(path.join(os.tmpdir(), "agency-runner-"));
+    try {
+      const config = { ...this.config };
+      delete config.distDir;
+      const argsString = Object.values(args).map((v) => globalThis.JSON.stringify(v)).join(", ");
+      return await this.runNode({ config, agencyFile, nodeName, hasArgs: argsString.length > 0, argsString, scratchDir, quietCompile: true });
+    } finally {
```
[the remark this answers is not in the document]
egonSchiele: Fixed in bf0d4d65: run() and runStructured() now take positional args (JSON[]) instead of Object.values(Record), so call order is explicit. (The named->positional mapping for the agent-under-test is handled in BaseOptimizer.runAgent in 3B, where the node param order is known.)

packages/agency-lang/lib/optimize/evalCache.ts
egonSchiele: Went with a nested map rather than a hyphen: a hyphen still collides because workspace keys are themselves ws-N (already contain hyphens), e.g. (ws-1, 2-x) vs (ws-1-2, x) both flatten to ws-1-2-x. Nesting by workspace then input removes the separator entirely. (bf0d4d65)

PR #312, opened 2026-06-18
Custom graders, validation sets, and optimize DX

egonSchiele:
## Summary

Adds composition-based grading, validation sets, and a legible report to `agency eval optimize`. Motivated by the reward-hacking failure where the optimizer gamed a soft LLM judge (objective climbed to 0.767) instead of solving the task.

Implements `docs/superpowers/specs/2026-06-18-custom-graders-and-optimize-dx-design.md` per `docs/superpowers/plans/2026-06-18-custom-graders-and-optimize-dx.md` (both included here).

## What changed

**Custom graders (the core).** Users can pass `--graders ./grading.ts` (or set `eval.optimize.graders`). The module default-exports one grader or an array; a "grader" is a metric function `({ output, input, judge }) => number | boolean | Grade`, a `grader(fn, opts)`-wrapped function carrying policy, a configured built-in (`ExactMatch`/`Contains`/`Similarity`/`LlmJudge`), or a `BaseGrader` subclass. The surface ships at `agency-lang/optimize`. Functions are adapted into `BaseGrader` so the existing Scorecard/gating/weighting pipeline is unchanged. The module is loaded by esbuild-transpile then `import()` (the established CLI-layer pattern). Per-input data lives under `input.metadata`, so input files can carry any fields graders need. When a grading module is present, a per-input `goal` is optional.

**Default behavior unchanged.** With no module, a single built-in `LlmJudge` (now with an inline `goal` option + default bundled judge file) grades against `--goal` / per-input `goal`. `--goal` and `--inputs` now combine (`--goal` is the overall-goal default).

**Validation sets.** `--validation-inputs` (or seeded `--validation-split`). Search runs on train; with `greedy` the writeback champion is the best **validation** objective, and `report.md` shows train-vs-validation. `gepa`/`example` report a validation objective but select on train and say so (no silent caps).

**DX / run dir.** Startup echo of the resolved grading setup; fail-fast `validateInput` pre-flight against the first input; a headline `report.md` (setup, per-iteration table, champion per-input grade breakdown) and `champion/grades.json` — the breakdown is the lens that makes a gamed metric obvious.

## Notable decisions discovered during implementation

- `toGrader` duck-types `BaseGrader` instances so a grader loaded from a user module in a different realm is still accepted (genuine dual-package safety; also unblocks the in-repo test).
- `requireGoal` is relaxed when an overall `--goal` is set, not just when a grading module is present.
- The §5.4 run-dir reorg is descoped (the live `BaseOptimizer` path writes no `iter-N/` pile — that belongs to the out-of-scope stdlib path). See the plan's Known Limitations.

## Test plan

- [x] Unit tests for every new unit: `goalJudgeFile`/`asJudgeText`, `FunctionGrader`/`grader`/`toGrader`, `loadGradingModule` (incl. a real `agency-lang/optimize` import), loader `requireGoal`/metadata, `MatchGrader` pre-flight, startup echo + fail-fast, `breakdown`, `renderReport`, `splitInputs`, `scoreFiles`, greedy champion-by-validation.
- [x] `pnpm run build` (tsc + make) clean; `pnpm run lint:structure` clean.
- [x] 392 tests pass across `lib/optimize`, `lib/cli/eval`, `lib/eval`; `lib/stdlib/agencyEval` green.
- [ ] Full agency execution suite — deferred to CI.


packages/agency-lang/lib/optimize/report.ts
```diff
@@ -0,0 +1,67 @@
...
+  return ["## Iterations", "", "| iter | decision | detail |", "| --- | --- | --- |", ...rows].join("\n");
+}
+
+function championGrades(breakdown?: InputBreakdown[]): string {
+  if (!breakdown || breakdown.length === 0) return "";
+  const rows = breakdown.flatMap((b) =>
+    b.grades.map((g) => {
+      const score = g.kind === "scalar" ? g.value.toFixed(3) : g.pass ? "pass" : "fail";
+      const out = String(b.output).replace(/\|/g, "\\|").replace(/\n/g, " ").slice(0, 80);
+      return `| ${b.inputId} | ${g.grader} | ${score} | ${(g.feedback ?? "").replace(/\|/g, "\\|").replace(/\n/g, " ")} | ${out} |`;
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 409e0478. `championGrades` now uses `asJudgeText(b.output)` (strings pass through, everything else is JSON-stringified), and `asJudgeText` falls back to "" when `JSON.stringify` returns `undefined` — which also hardens the `llmJudge`/`humanGrader` call sites.

packages/agency-lang/lib/optimize/validationSplit.ts
```diff
@@ -0,0 +1,16 @@
...
+import { makeRng, sampleWithoutReplacement } from "./rng.js";
+
+export type Split = { train: Input[]; validation: Input[] };
+
+/** Hold out a fraction of inputs for validation, deterministically by seed.
+ *  Always keeps at least one training input. */
+export function splitInputs(inputs: Input[], ratio: number, seed = 0): Split {
+  const clamped = Math.max(0, Math.min(1, ratio));
+  const maxHoldout = Math.max(0, inputs.length - 1);          // always keep ≥1 for training
+  const holdout = Math.min(maxHoldout, Math.floor(clamped * inputs.length));
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 409e0478. `splitInputs` now throws `--validation-split must be a finite number between 0 and 1` on a non-finite ratio, with a regression test. (Out-of-range finite values stay clamped, keeping at least one training input.)

packages/agency-lang/docs/site/cli/eval.md
```diff
@@ -9,7 +9,7 @@ description: Documents the `agency eval extract` command for converting a captur
 
 ```
 agency eval run --agent <file>[:<node>] (--inputs <file|dir> | --goal <text>)
-agency eval optimize <file>[:<node>] (--inputs <file|dir> | --goal <text>)
+agency eval optimize <file>[:<node>] [--inputs <file|dir>] [--goal <text>] [--graders <file>] [--validation-inputs <file|dir>]
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 409e0478. The synopsis now shows `[--validation-inputs <file|dir> | --validation-split <ratio>]`.

packages/agency-lang/lib/cli/eval/optimize.ts
```diff
@@ -41,45 +47,109 @@ export type EvalOptimizeDeps = {
...
   }
   return result;
 }
 
+/** Optimize allows --inputs and --goal together (--inputs = data, --goal =
+ *  overall-goal default). Only --goal → a single synthetic input. */
+function optimizeInputSelection(opts: EvalOptimizeOptions): "inputs" | "goal" {
+  if (opts.inputs) return "inputs";
+  if (opts.goal) return "goal";
+  throw new Error("Provide --inputs (optionally with --goal as the overall goal), or --goal");
```
egonSchiele, line 87: should this allow both inputs and goal?

packages/agency-lang/lib/cli/eval/optimize.ts
```diff
@@ -41,45 +47,109 @@ export type EvalOptimizeDeps = {
...
+    seed: opts.seed ?? 0,
+  };
+}
+
+/** Fill in the default node and overall goal for inputs that omit them. */
+function withDefaults(inputs: Input[], node: string, goal?: string): Input[] {
+  return inputs.map((input) => ({
+    ...input,
+    node: input.node ?? node,
+    ...(input.goal === undefined && goal !== undefined ? { goal } : {}),
```
egonSchiele: unreadable

packages/agency-lang/lib/cli/eval/optimize.ts
```diff
@@ -41,45 +47,109 @@ export type EvalOptimizeDeps = {
...
-  const loaded = loadInputs(path.resolve(opts.inputs ?? ""), deps.makeId ?? nanoid);
-  const inputs: Input[] = loaded.map((input) => ({ ...input, node: input.node ?? resolved.node }));
-  return { agent: opts.agent, inputs };
+  if (optimizeInputSelection(opts) === "goal") return goalTarget(opts, resolved);
+  const s = resolveOptimizeSettings(opts);
+  // A per-input goal is required only when nothing else supplies one: no custom
+  // grading module AND no overall --goal default to fall back on.
+  const requireGoal = !s.gradersPath && s.goal === undefined;
+  const { inputs, validationInputs } = provisionInputs(s, resolved.node, requireGoal, deps);
+  return { agent: opts.agent, inputs, ...(validationInputs ? { validationInputs } : {}) };
```
egonSchiele: unreadable

packages/agency-lang/lib/eval/loadInputs.ts
```diff
@@ -71,13 +79,18 @@ function normalizeInput(raw: unknown, baseDir: string, makeId: MakeId): Input {
   if (spec.working_dir !== undefined && typeof spec.working_dir !== "string") {
     throw new Error("Eval input working_dir must be a string when provided");
   }
+  if (spec.metadata !== undefined && (!spec.metadata || typeof spec.metadata !== "object" || Array.isArray(spec.metadata))) {
```
egonSchiele: is this supposed to be a logic puzzle? If not it is hard to parse

packages/agency-lang/lib/eval/loadInputs.ts
```diff
@@ -71,13 +79,18 @@ function normalizeInput(raw: unknown, baseDir: string, makeId: MakeId): Input {
...
+    throw new Error("Eval input metadata must be an object when provided");
+  }
   const out: Input = {
     id: typeof spec.id === "string" ? spec.id : makeId(),
-    goal: spec.goal,
     args: (spec.args ?? {}) as Record<string, any>,
   };
+  if (typeof spec.goal === "string") out.goal = spec.goal;
+  if (spec.expected !== undefined) out.expected = spec.expected;
```
egonSchiele: some of the logic around spec.expected seems unnecessary. What would happen if you didn't have the `if` condition around this

packages/agency-lang/lib/optimize/grading/functionGrader.ts
```diff
@@ -0,0 +1,69 @@
...
+  // realm (its own resolved copy of agency-lang), so `instanceof` can miss it.
+  // Duck-type the BaseGrader public surface to accept it across the boundary.
+  if (isGraderLike(spec)) return spec as BaseGrader;
+  if (typeof spec === "function") return new FunctionGrader(spec);
+  throw new Error(
+    `Invalid grader: expected a grader function or grader instance, got ${spec === null ? "null" : typeof spec}.`,
+  );
+}
+
+function isGraderLike(spec: unknown): spec is BaseGrader {
```
egonSchiele, line 81: use a zod schema instead

packages/agency-lang/lib/optimize/goalJudgeFile.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/** Structured verdict shape the goal judge returns (0..1 score + reasoning). */
+export const ScalarVerdict = z.object({ score: z.number(), reasoning: z.string() });
+
+/** Render an agent output as the string a judge reads: strings pass through,
+ *  everything else is JSON so it reads as data rather than "[object Object]".
+ *  Top-level `undefined` (where JSON.stringify returns undefined) becomes "". */
+export function asJudgeText(output: unknown): string {
+  if (typeof output === "string") return output;
+  return globalThis.JSON.stringify(output) ?? "";
```
egonSchiele: Why is `globalThis` needed?

packages/agency-lang/lib/optimize/report.ts
```diff
@@ -0,0 +1,68 @@
...
+/** Render a human-readable Markdown report for an optimize run. Pure. */
+export function renderReport(result: OptimizeResult, meta: ReportMeta): string {
+  return [
+    metaBlock(result, meta),
+    iterationTable(result),
+    championGrades(result.championBreakdown),
+  ].filter((s) => s.length > 0).join("\n\n") + "\n";
+}
+
+function metaBlock(result: OptimizeResult, meta: ReportMeta): string {
```
egonSchiele: I'd suggest using a typestache template to make some of this more readable (see lib/templates/). Docs https://github.com/egonSchiele/typestache

egonSchiele: Done in 3680d6c1 — renderReport now uses a typestache template (lib/templates/cli/optimizeReport.mustache) for the document skeleton + iterations table, following the coverageReport.mustache pattern. The dynamic blocks (rows/meta lines) are precomputed strings because each table cell needs pipe/newline escaping (a cell() helper), which has to live in TS.

packages/agency-lang/lib/optimize/goalJudgeFile.ts
egonSchiele: Not needed — removed in 3680d6c1 (now plain JSON.stringify). The JSON *type* import elsewhere lives in a separate namespace from the global JSON *value*, so it never actually shadowed it.

packages/agency-lang/lib/optimize/grading/functionGrader.ts
egonSchiele: Done in 3680d6c1. isGraderLike now uses a zod GraderLikeSchema ({ run, name, mustPass } as functions), and I also added a zod GradeSchema to validate the metric return shape in coerce().

packages/agency-lang/lib/eval/loadInputs.ts
egonSchiele: You're right — removed the guard in 3680d6c1. expected is now assigned directly in the literal; when absent it is undefined, which consumers (getPath / `!== undefined` checks) and toEqual treat identically to a missing key. The guard was pure noise.

packages/agency-lang/lib/eval/loadInputs.ts
egonSchiele: Agreed, that condition was a puzzle. Extracted an isPlainObject(value) helper in 3680d6c1 and used it for both the args and metadata checks.

packages/agency-lang/lib/cli/eval/optimize.ts
egonSchiele: Fixed in 3680d6c1 — buildTarget now builds the target and adds validationInputs with a plain if, instead of the conditional spread.

packages/agency-lang/lib/cli/eval/optimize.ts
egonSchiele: Fixed in 3680d6c1 — withDefaults now builds each input and sets the goal default with a plain if (the spread-ternary is gone).

packages/agency-lang/lib/cli/eval/optimize.ts
egonSchiele: Yes — both are allowed: --inputs is the data and --goal fills in as the overall-goal default for inputs that omit one. Added a doc comment on optimizeInputSelection in 3680d6c1 spelling that out.

packages/agency-lang/lib/optimize/grading/functionGrader.ts
```diff
@@ -0,0 +1,69 @@
+import { BaseGrader } from "./baseGrader.js";
+import { asJudgeText, goalJudgeFile, ScalarVerdict } from "../goalJudgeFile.js";
+import type { Grade, GraderInput, GraderOptions, Input, JSON } from "./types.js";
+
+/** What a metric function receives. `input` is the typed Input; per-input grading
+ *  data (an expected answer, tags) lives under `input.metadata`. */
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 3680d6c1 — the GraderContext doc now says the gold answer is input.expected, with extra per-input data under input.metadata.

packages/agency-lang/docs/superpowers/specs/2026-06-18-custom-graders-and-optimize-dx-design.md
```diff
@@ -0,0 +1,234 @@
...
+
+**Where custom per-input data lives.** `Input` is a typed shape
+(`id`, `goal?`, `args`, `node?`, `working_dir?`, `metadata?`); the loader reads
+those fields and any unrecognized top-level keys are not part of `Input`.
+Grader-specific data (an expected answer, tags, a rubric) therefore lives under
+`metadata`, and graders read it from `ctx.input.metadata`. `args` (and optional
+`node`) drive the agent run; `metadata` is freeform and grader-defined. This
+keeps one typed contract for an input rather than reintroducing a
+"tasks can be any shape" free-for-all — the cost is one level of nesting
+(`input.metadata.expectedCapital`, not `input.expectedCapital`).
```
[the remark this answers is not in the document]
egonSchiele: Addressed in 3680d6c1 — added a 'Superseded' note to this section: a first-class Input.expected field landed in a follow-up; only *additional* grader data lives under metadata now. (The spec is a dated design doc; the shipped shape is in docs/site/cli/optimize.md and docs/dev/writing-optimizers.md.)

packages/agency-lang/docs/superpowers/specs/2026-06-18-custom-graders-and-optimize-dx-design.md
```diff
@@ -0,0 +1,234 @@
...
+any of four forms:
+
+```ts
+import { grader, ExactMatch, LlmJudge, type Grader } from "agency-lang/optimize";
+
+// (a) a metric function: ctx = { output, input, judge }
+//     `input` is the typed Input (id, goal?, args, node?, metadata). Your own
+//     per-input fields live under `metadata` — there is no magic top-level key.
+const exact: Grader = ({ output, input }) =>
+  output === input.metadata?.expectedCapital ? 1 : 0;  // number (0..1), boolean, or {score, feedback}
```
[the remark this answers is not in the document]
egonSchiele: Same as above — this spec predates the first-class Input.expected decision (made in a follow-up). Added a Superseded note pointing to the current docs rather than rewriting the historical design record.

PR #314, opened 2026-06-20
Fix duplicate LLM tool names; add .rename() and agent smoke tests

egonSchiele:
## Problem

Running the agency agent crashed on its first turn:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"tools: Tool names must be unique."}}
```

The code subagent passed the LLM **five tools all named `read`**. `skillsDir()` returns `read.partial(dir).describe(...)`, and `.partial()` / `.describe()` preserve the base function's name — so `superpowersSkill`, `docSkill`, `cliSkill`, `appendixSkill` were all named `read` (alongside `workspace.read`). Anthropic rejects duplicate tool names with a 400.

The smoke tests added here surfaced a **second instance** of the same class: the research subagent had two tools named `search`, because `import { search as wikipediaSearch }` renames the binding but not the tool name.

## Changes

**Fix the bug**
- Add `AgencyFunction.rename(name)` — sets both `name` and `toolDefinition.name`. Wired into `__callMethod` dispatch and the typechecker; documented in the partial-application guide.
- `skillsDir()` now renames each tool after its directory (`skills_docs_guide`, ...).
- Research subagent renames the wikipedia tools (`wikipedia_search`, ...).

**Catch it earlier (#1)**
- Runtime backstop `assertUniqueToolNames()` in `prompt.ts` throws *before* the request hits the wire, naming the collision and pointing at `.rename()`. A fully general compile-time check isn't feasible (the duplicate name is produced inside `skillsDir`, invisible to the type system), so this runs next to the existing `validateForLLM()` pre-flight.

**Statelog gap (#3)**
- Dispatch failures now emit an `llmError` event carrying the tool list. Previously only the success path (`promptCompletion`) logged the request + tools, so a 400 left no record of what was sent.

**Agent tests + CI (#4)**
- New smoke-test pattern for built-in agents: a colocated harness `lib/agents/<agent>/<name>.agency` (the production `main` is an interactive REPL, not testable) + `<name>.test.json`, run deterministically with `useTestLLMProvider` + `llmMocks` (no API key).
- `toolWiring.agency` pings each subagent's exported tool list; `test:agents` script + a "Built-in agent tests" CI step in `test.yml`.
- This harness is what caught the research-subagent collision.

## Testing

- Full unit suite green (6019 passed); structural lint clean.
- New tests: `rename()` (runtime + typechecker), `assertUniqueToolNames`, statelog `tools` field, and the 4-node agent smoke test — all pass.
- Verified end-to-end: duplicate names trip the backstop with a clear message; renamed tools pass; all four subagent tool lists are now unique.

- The other built-in agents (`eval`, `optimize`, `policy`, `review`) still need smoke harnesses.
- A `promptStart` statelog event (logging every request, not just failures) was considered but deferred in favor of the lighter error-path enrichment.


egonSchiele:
Pushed `f83ad04a` addressing the review.

**Copilot inline comments**
- **prompt.ts (best-effort logging):** the dispatch-error statelog emit is now wrapped in try/catch and swallows any failure, then rethrows the original LLM/transport error unchanged — logging can never mask the real error.
- **call.ts (.rename type):** `.rename()` now runtime-checks the argument is a string (guards `any`/JS callers the typechecker can't see); throws a clear message otherwise.
- **skills.agency (Windows paths):** `skillToolName()` now sanitizes with a single `re/[^a-zA-Z0-9]+/g` pass, so backslash separators and the drive-letter colon (`C:\docs\guide`) collapse to underscores. (Bonus: this avoids a literal backslash string, which the template-literal codegen currently mis-emits — filing that separately.)

**Review feedback**
- **Declarative method typing:** `describe`/`preapprove`/`rename` now live in a table (`AGENCY_FUNCTION_METHOD_TYPES` in `builtins.ts`) validated generically via the existing `validatePrimitiveMethodCall`, instead of bespoke branches in `synthesizer.ts`. Adding a new such method is now a one-line signature. (`partial` stays special — it validates named args against the base function's parameters.)
- **Agency test for rename:** `tests/agency/tool-rename.agency` passes two copies of `add` to one `llm()` call, renames each, and asserts the call succeeds; a companion node confirms the un-renamed case is rejected by the backstop.
- **Test file placement:** the agent smoke test moved to `lib/agents/agency-agent/tests/` so it no longer reads as a source file.
- **Real LLM on main:** added a "Built-in agent tests (real LLM)" step to `test-with-llm.yml`; dropped the forced `useTestLLMProvider` so the env picks the provider (deterministic on PRs, real on merge to main). The harness rejects any tool interrupt so a real model calling a tool stays side-effect-free while the call still completes. Verified both paths locally (deterministic + real OpenAI).

Full unit suite (6021), lint, and both agent-test modes are green locally.

PR #316, opened 2026-06-20
Agent follow-ups: testable agent fn, basic-agency policy, REPL echo + tool-loss fixes

egonSchiele:
Follow-ups on the agency agent (4 items from review). Debugging notes for the two bugs are in the commit; summary below.

## #1 — Separate the REPL from the agent
Extracted `agentReply(userMsg)` in `agent.agency` (slash-expansion + `mainAgent`) and routed both the REPL (`_runTurn`) and the one-shot path through it. The agent's per-message logic is now callable with just a prompt — no terminal, no user input, no policy handler. New `tests/agentTurn.agency` drives it with the deterministic provider.

## #2 — Basic agency commands without prompting
- **(a)** Added `parseAST` (parse) as a code-agent tool; `typecheck` was already one. Both take a source string and are pure (no interrupt), so they need no approval.
- **(b)** Added an exec-based `agencyCli(args)` tool. Per the discussion, this uses `exec()` (structured args, **no shell → no command chaining**) rather than globbing the `bash` shell string. Both default policies now auto-approve `std::exec` for the read-only `agency parse`/`ast`/`typecheck` subcommands (added to `minimal`; `recommended` inherits via spread). Other subcommands still prompt. New `tests/execPolicy.agency` verifies the policy decisions.

> Note: existing users with a saved `~/.agency-agent/policy.json` won't get the new rules until that file is regenerated (the defaults are only written on first run). New users get them automatically.

## #3 — Duplicated prompt echo after approving an interrupt (bug)
Root cause: the interrupt menu runs through the `prompts` library in `lib/stdlib/ui.ts :: _runPrompt`, which left `process.stdin` in cooked mode. The line-mode REPL then double-echoed input, and it persisted ("from then on") because nothing re-asserted raw mode. Fix: snapshot `stdin.isRaw` on entry and restore it in the `finally`, mirroring the existing `wasPaused` handling. (Terminal behavior can't be unit-tested in CI — needs a manual confirm.)

## #4 — Code agent loses its tools after the first turn (bug)
Root cause: in `subagents/code.agency` the retry loop was gated on the **module-level** `_first` (`while (hasErrors || _first)`). It primes the loop only on the very first `codeAgent` call in the process; on every later call `_first` is already `false` and `hasErrors` starts `false`, so the loop body never runs — the tool-enabled `llm()` is skipped and only the tool-less format pass executes, so the model reports it can't run the command. Fix: a per-call `firstIteration` primer; `_first` now only gates the one-time system message. Also declared `reply` per-call instead of as an implicit module global.

## Testing
- Full unit suite green; structural lint clean.
- Agent tests: 9 across 3 files (`toolWiring`, `agentTurn`, `execPolicy`) pass deterministically; `agentTurn` is pinned to the deterministic provider (exact-match), `toolWiring` stays real-LLM-safe, `execPolicy` makes no LLM calls. All auto-discovered by the existing `test:agents` CI steps.
- #3 and the interactive side of #4 should be confirmed manually in the REPL.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
@@ -1,3 +1,7 @@
+// `std::exec` rules give basic agency functionality: auto-approve the
+// read-only `agency` subcommands the code agent runs via the exec-based
+// `agencyCli` tool (matched on command + subcommand — no shell, so no
+// chaining). Other subcommands (run, compile, ...) still prompt.
 export static const minimalAutoApprovePolicy = {
   "std::memory::remember": [{
     action: "approve"
@@ -13,7 +17,30 @@ export static const minimalAutoApprovePolicy = {
   }],
   "std::memory::disableMemory": [{
     action: "approve"
-  }]
+  }],
+  "std::exec": [
+    {
+    match: {
+      command: "agency",
+      subcommand: "parse"
+    },
+    action: "approve"
+  },
+    {
+    match: {
+      command: "agency",
+      subcommand: "ast"
+    },
+    action: "approve"
+  },
+    {
+    match: {
+      command: "agency",
+      subcommand: "typecheck"
+    },
+    action: "approve"
+  },
+  ]
 }
 
 export static const recommendedAutoApprovePolicy = {
--- a/packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
@@ -1,7 +1,34 @@
-// `std::exec` rules give basic agency functionality: auto-approve the
-// read-only `agency` subcommands the code agent runs via the exec-based
-// `agencyCli` tool (matched on command + subcommand — no shell, so no
-// chaining). Other subcommands (run, compile, ...) still prompt.
+import { map } from "std::array"
+
+// Basic agency functionality: auto-approve the read-only `agency`
+// subcommands the code agent runs via the exec-based `agencyCli` tool.
+// Matched on command + subcommand (no shell, so no chaining); other
+// subcommands (run, compile, ...) still prompt.
+static const AGENCY_SAFE_SUBCOMMANDS = [
+  "parse",
+  "ast",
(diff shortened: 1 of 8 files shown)


packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
```diff
@@ -13,7 +17,30 @@ export static const minimalAutoApprovePolicy = {
...
+  },
+    {
+    match: {
+      command: "agency",
+      subcommand: "ast"
+    },
+    action: "approve"
+  },
+    {
+    match: {
```
egonSchiele: also add to policy these commands: `help`, `tc`, `preprocess`, `definition`, `diagnostics`, `doc`, `literate`

packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
```diff
@@ -64,6 +65,23 @@ export def setCwd(dir: string): string {
...
+  to the executable (no shell interpretation), so they cannot be chained
+  or redirected. The default agent policies auto-approve the read-only
+  `parse`, `ast`, and `typecheck` subcommands; other subcommands still
+  prompt for approval. Prefer this over `bash` for running agency.
+
+  @param args - CLI arguments, e.g. ["typecheck", "foo.agency"]. The
+    first element is the subcommand.
+  """
+  return exec("agency", args)
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 7b969000. `agencyCli` now returns `any` (same as the `bash` tool, since `ExecResult` isn't exported from std::shell) and the docstring says it returns an object with `stdout`, `stderr`, and `exitCode`.

packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
egonSchiele: Done in 7b969000. Added help, tc, preprocess, definition, diagnostics, doc, and literate to the auto-approve list (alongside parse/ast/typecheck). They're now driven from a single `AGENCY_SAFE_SUBCOMMANDS` list, so adding more later is one line. Extended the execPolicy test to cover one of the new ones.

PR #320, opened 2026-06-21
Render std::edit interrupts as a diff (prompt + auto-approve)

egonSchiele:
## What

Edit interrupts (`std::edit`) now render as a colored diff instead of a raw key/value dump of the before/after content — both in the approval prompt and when an edit is auto-approved.

## How

Two coordinated changes:

**`stdlib/policy.agency`**
- `prettyPrintInterruptData` special-cases the `std::edit` effect: it renders a metadata table (just `dir` — `before`/`after`/`edits`/`filename` are excluded as bulky or redundant) followed by a unified diff with a `⏺ Edit: <filename>` header.
- It now takes the interrupt's `effect` as a second arg (the call site passes `intr.effect`), since it receives `intr.data` (the payload) and the payload alone can't tell you the effect.
- `renderObjAsTable` returns `""` when no rows remain, so an all-excluded edit payload yields just the diff with no empty table.
- `renderEditDiff` is hardened against a missing `before`/`after` (`|| ""`) so a malformed/partial edit interrupt renders cleanly instead of crashing `diff()`.
- New `willAutoApprove(intr): boolean` — reports whether the active policy would auto-approve an interrupt without prompting. Reads the already-loaded policy (no hot-path I/O).

**`lib/agents/agency-agent/agent.agency`**
- The handle blocks now print the edit diff (`renderEditDiff`) **only when the edit will be auto-approved** (`willAutoApprove(data)`). Auto-approved edits skip the approval prompt — which is where the diff would otherwise show — so without this the user would have no idea what changed. Prompted edits get their diff from the prompt itself, so there's no double-print.

```text
edit interrupt
 ├─ policy auto-approves → agent prints diff (prompt skipped)
 └─ policy prompts       → modal body shows diff (prettyPrintInterruptData)
```

## Testing

- All policy tests pass: agency-js (`policy-prompt-array-data`, `cli-policy-handler`, `policy-autoapprove-existing-file`, `ui-policy-integration`), agency (`policy-build-scoped-match`, `policy-record-rule`, `policy-record-scoped-rule`), and `lib/runtime/policy.test.ts` (20).
- `policy-prompt-array-data` surfaced a real crash (an edit with no `before`/`after` blew up `diff()`); the `|| ""` guard fixes it and the test now renders cleanly.
- Interactive verification of the in-terminal diff rendering is still worth a manual look (terminal output), as with other agent UI changes.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -10,11 +10,10 @@ import {
   cliPolicyHandler,
   parsePolicyFile,
   setPolicy,
-  willAutoApprove,
  } from "std::policy"
 import { exists } from "std::shell"
 import { commandsDir, expandSlash } from "std::skills"
-import { highlight, diff } from "std::syntax"
+import { highlight } from "std::syntax"
 import { cwd, env, isTTY, readStdin, setTitle } from "std::system"
 import { getCost, getTokens, systemMessage } from "std::thread"
 import { chooseOption, ChoiceItem } from "std::ui"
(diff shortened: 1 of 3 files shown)


packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -657,7 +666,12 @@ def startInteractive(handler: any, seedTarget: string, seedPrompt: string) {
       paletteCommands: mergedPalette(),
     )
   } with (data) {
-    renderEditDiff(data)
+    // Auto-approved edits skip the approval prompt (which would show the
+    // diff), so surface it here. Prompted edits get their diff from the
+    // prompt itself — don't double-print.
+    if (data.effect == "std::edit" && willAutoApprove(data)) {
```
egonSchiele: you don't really need `willAutoApprove`, you could just check the return value of the `handler` func

egonSchiele: Good call on dropping willAutoApprove — done in 3c6456fa. One wrinkle with literally checking the handler's return value, though: `_handler` returns `approve()` for *both* the auto-approve path (policy.agency:818) and every prompt-approve path (828/833/841), so the return alone can't tell them apart. And in line mode `chooseOption` prints the prompt body (the diff) persistently (ui.agency:1187), so an `isApproved(result)` check in the handle block would double-print the diff for prompted edits.

So instead of checking the return, I moved the diff into the handler where the decision is actually made: `_handler` prints `renderEditDiff(intr.data)` on its auto-approve branch. That drops willAutoApprove, avoids the double-print, and makes it generic — any `std::policy` consumer now gets auto-approve diffs, not just the agent. `agent.agency` is back to a plain `return handler(data)` (also removed its now-unused `renderEditDiff` + `diff` import). Verified end-to-end: an auto-approved std::edit prints the diff, and prompted edits still get theirs from the prompt.

PR #323, opened 2026-06-21
Roll subagent (tool-dispatch) LLM cost up to the parent stack

egonSchiele:
## What

The agent footer's cost (and `/cost`) under-reported any turn that used a subagent — e.g. the oracle/explorer running on a pricier model. Those costs are now included.

## Root cause

`getCost()` / `getTokens()` read the active stack's per-branch `localCost` / `localTokens`, which is meant to be cumulative across a stack **and its child branches**. `fork`/`parallel`/`race` make that true by wiring `seedBranchCost` + `propagateBranchCost` (via the node `Runner`), so a branch's spend is seeded from the parent and its delta propagates back on join.

But **LLM tool dispatches** run each tool in its own branch stack via `PromptRunner.parallel` (`runBatch.ts:358`), and that call site passed **neither** hook. So a subagent's spend accumulated only on its branch stack and was discarded at join — never reaching `getCost()`/`getTokens()`. Since subagents are invoked as tools, all of their cost was missing from the totals.

## Fix

- Extract `seedBranchCost` / `propagateBranchCost` into shared functions in `state/stateStack.ts`.
- `Runner` now delegates to them (no behavior change; removes the duplicated logic).
- Wire both into `PromptRunner.parallel`'s `runBatch` call, so tool branches seed + propagate exactly like fork/race. The seed is idempotent across the approval-interrupt/resume cycle, and propagate only fires on the no-interrupt completion path, so the per-branch delta math stays correct across resumes.

## Testing

- 117 unit tests pass across `runner` / `runBatch` / `stateStack` (these exercise the now-shared seed/propagate logic for fork/race).
- New deterministic regression test (`tests/agency/subagent-cost-propagation.*`): an outer `llm()` calls a tool that itself calls `llm()` (3 calls total). With the fix `getTokens()` reports **6** synthetic tokens (all three calls); before it reported **4** (the tool branch's spend was dropped).

Note: `getCost()` remains cumulative (per the design intent) — it now reflects the coordinator's calls plus every subagent.

### DIFF
--- a/packages/agency-lang/lib/runtime/promptRunner.ts
+++ b/packages/agency-lang/lib/runtime/promptRunner.ts
@@ -3,7 +3,12 @@ import { hasInterrupts, type Interrupt } from "./interrupts.js";
 import { runBatch, type RunBatchResult } from "./runBatch.js";
 import type { SourceLocationOpts } from "./state/checkpointStore.js";
 import type { RuntimeContext } from "./state/context.js";
-import type { State, StateStack } from "./state/stateStack.js";
+import {
+  seedBranchCost,
+  propagateBranchCost,
+  type State,
+  type StateStack,
+} from "./state/stateStack.js";
 
 /**
  * Thrown by {@link PromptRunner.step} when a step body returns interrupts.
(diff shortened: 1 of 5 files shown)


packages/agency-lang/lib/runtime/promptRunner.ts
```diff
@@ -205,6 +210,18 @@ export class PromptRunner {
...
       hooks: {
+        // Roll each tool branch's LLM cost/tokens up to the parent stack,
+        // just like fork/parallel/race do (see Runner). Subagents run as
+        // tools in their own branch stacks; without these hooks their
+        // spend (e.g. an oracle on a pricier model) never propagated back,
+        // so `getCost()`/`getTokens()` under-reported the turn. seed is
+        // idempotent across interrupt/resume; propagate fires only on the
+        // no-interrupt completion path (runBatch.ts), and the delta math
+        // is correct across resume because the seed baseline is per-branch.
+        seedBranchCost: (childStack, parentStack) =>
```
egonSchiele, line 221: You could just write `seedBranchCost` instead of `seedBranchCost: (childStack, parentStack) => seedBranchCost: (childStack, parentStack)` right?

PR #324, opened 2026-06-21
Per-model cost attribution in footer + /cost

egonSchiele:
## What

Surface **which models** a turn actually used, with per-model cost attribution.

The REPL footer showed the model name from a single `lastModel` scalar that `updateTokenStats` overwrote on *every* LLM call. Because the coordinator makes a final wrap-up call after a subagent returns, the last writer almost always won — so the footer showed the coordinator's model and quietly hid that a turn delegated to, say, the oracle on `opus-4.8`. The cost was already accumulated correctly (#323); only the *attribution* was lost.

## Changes

- **Data layer** — `__tokenStats.models[model]` now accumulates `inputTokens` / `outputTokens` / `cost` on every LLM call in `updateTokenStats`. Tool/subagent branches pointer-share `__tokenStats` (via `shareGlobals`), so their spend lands here automatically — no branch roll-up wiring needed. Plain JSON, so it survives checkpoint/restore.
- **Footer (Option A)** — lists the distinct models that did work *this turn*, ordered by spend descending (`… · opus-4.8, gpt-5-mini`). A single-model turn reads exactly as before. Computed by diffing a before/after per-model snapshot (`modelsUsedThisTurn`).
- **`/cost`** — gains a per-model breakdown:
  ```
  Cost:   $0.0410
  Tokens: 1340

  By model:
    opus-4.8     ↑800 ↓200  $0.0300
    gpt-5-mini   ↑400 ↓140  $0.0110
  ```
  backed by a new `getModelCosts()` builtin in `std::thread` (process-wide totals, sorted by cost descending).
- **Removed** the now-redundant `lastModel` scalar (it had two sources of truth and only the footer read it).

## DX decisions

- Footer shows **names only** by default; the full per-model breakdown is on demand via `/cost`.
- Multiple models in the footer are ordered **by spend descending** so the pricey one catches the eye.

## Tests

- `lib/stdlib/cli.test.ts` — `modelsUsedThisTurn`: empty turn, single model, multi-model sort by spend, delta-not-cumulative ordering, excludes a model used only in a prior turn, cost-tie tie-break by name.
- `lib/runtime/utils.test.ts` — `updateTokenStats` per-model: new model, repeated accumulation, multiple models tracked separately, no entry without a model name, backfill for token-stats restored from older checkpoints.
- `tests/agency/model-cost-breakdown.agency` — end-to-end: the breakdown is keyed by the provider's model name, groups by model, and **aggregates tool-branch spend** into the per-model totals (matches `getTokens()`).

All unit + new Agency tests pass; structural linter clean; existing `subagent-cost-propagation` Agency test still passes.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -15,7 +15,7 @@ import { exists } from "std::shell"
 import { commandsDir, expandSlash } from "std::skills"
 import { highlight } from "std::syntax"
 import { cwd, env, isTTY, readStdin, setTitle } from "std::system"
-import { getCost, getTokens, getModelCosts, systemMessage } from "std::thread"
+import { getCost, getModelCosts, systemMessage } from "std::thread"
 import { chooseOption, ChoiceItem } from "std::ui"
 
 import { figs, title } from "./images/images.agency"
@@ -232,12 +232,22 @@ def _runTurn(msg: string): boolean {
     return true
   }
   if (trimmed == "/cost") {
-    pushMessage("Cost:   $${getCost().toFixed(4)}")
-    pushMessage("Tokens: ${getTokens()}")
-    // Per-model breakdown so a turn that delegated to a subagent on a
-    // pricier model (e.g. the oracle on opus-4.8) shows where the spend
-    // went. Sorted by cost descending by getModelCosts().
+    // Derive BOTH the header total and the per-model breakdown from the
+    // same process-wide accumulator (getModelCosts), so the `Cost:` /
+    // `Tokens:` lines and the rows below them can never visibly disagree.
+    // (getCost()/getTokens() read the per-branch accumulator, a different
+    // scope — adjacent in the output, they could in principle diverge.)
+    // Rows are sorted by cost descending by getModelCosts(); a subagent
+    // on a pricier model (e.g. the oracle on opus-4.8) shows up here.
     const byModel = getModelCosts()
+    let totalCost = 0.0
+    let totalTokens = 0
+    for (m in byModel) {
+      totalCost = totalCost + m.cost
+      totalTokens = totalTokens + m.inputTokens + m.outputTokens
+    }
+    pushMessage("Cost:   $${totalCost.toFixed(4)}")
+    pushMessage("Tokens: ${totalTokens}")
     if (byModel.length > 0) {
       pushMessage("")
       pushMessage("By model:")
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -15,7 +15,7 @@ import { exists } from "std::shell"
 import { commandsDir, expandSlash } from "std::skills"
 import { highlight } from "std::syntax"
 import { cwd, env, isTTY, readStdin, setTitle } from "std::system"
-import { getCost, getTokens, systemMessage } from "std::thread"
+import { getCost, getTokens, getModelCosts, systemMessage } from "std::thread"
 import { chooseOption, ChoiceItem } from "std::ui"
 
 import { figs, title } from "./images/images.agency"
@@ -234,6 +234,17 @@ def _runTurn(msg: string): boolean {
   if (trimmed == "/cost") {
     pushMessage("Cost:   $${getCost().toFixed(4)}")
     pushMessage("Tokens: ${getTokens()}")
+    // Per-model breakdown so a turn that delegated to a subagent on a
+    // pricier model (e.g. the oracle on opus-4.8) shows where the spend
+    // went. Sorted by cost descending by getModelCosts().
+    const byModel = getModelCosts()
+    if (byModel.length > 0) {
+      pushMessage("")
+      pushMessage("By model:")
+      for (m in byModel) {
+        pushMessage("  ${m.model}  ↑${m.inputTokens} ↓${m.outputTokens}  $${m.cost.toFixed(4)}")
+      }
+    }
     return true
   }
   // Hand the message to the agent. `agentReply` does the slash-command
--- a/packages/agency-lang/lib/stdlib/cli.ts
+++ b/packages/agency-lang/lib/stdlib/cli.ts
@@ -480,42 +480,71 @@ type TurnStats = {
   elapsedMs: number;
   inputTokens: number;
   outputTokens: number;
-  model: string;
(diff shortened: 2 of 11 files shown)


packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -234,6 +234,17 @@ def _runTurn(msg: string): boolean {
   if (trimmed == "/cost") {
     pushMessage("Cost:   $${getCost().toFixed(4)}")
     pushMessage("Tokens: ${getTokens()}")
+    // Per-model breakdown so a turn that delegated to a subagent on a
+    // pricier model (e.g. the oracle on opus-4.8) shows where the spend
+    // went. Sorted by cost descending by getModelCosts().
+    const byModel = getModelCosts()
```
egonSchiele, line 242: **Scope mismatch worth confirming.** The header above reads `getCost()`/`getTokens()` (both *per-branch*), while this breakdown reads `getModelCosts()` (*process-wide*). The agency test confirms they reconcile in the single-branch + tool case, but if a turn delegates to a subagent on a different model, the per-model rows could sum to more than the `Tokens:` line shows (depending on whether token roll-up to the main branch matches cost roll-up). They sit adjacent in the output, so a visible disagreement would be confusing. Can you confirm the totals can't diverge? If they can, consider sourcing the `Cost:`/`Tokens:` header from the same process-wide accumulator for internal consistency.

packages/agency-lang/lib/stdlib/cli.ts
```diff
@@ -480,42 +480,71 @@ type TurnStats = {
...
+
+/** Distinct models whose token count grew between two snapshots, ordered
+ *  by cost spent during the turn (descending), name as tiebreak. This is
+ *  what the footer lists after the per-turn token/time stats. */
+function modelsUsedThisTurn(before: TokenSnapshot, after: TokenSnapshot): string[] {
+  const used: { name: string; cost: number }[] = [];
+  for (const [name, a] of Object.entries(after.models)) {
+    const b = before.models[name];
+    const tokenDelta = a.tokens - (b?.tokens ?? 0);
+    if (tokenDelta > 0) used.push({ name, cost: a.cost - (b?.cost ?? 0) });
```
egonSchiele: Detection uses `tokenDelta > 0` but ordering (below) uses the *cost* delta. A model with positive cost but zero token delta would be silently dropped. Can't happen with current providers (every call adds tokens), so this is only a latent edge case — maybe worth a one-line comment noting the assumption.

packages/agency-lang/lib/stdlib/thread.ts
```diff
@@ -100,6 +100,40 @@ export async function _getTokens(): Promise<number> {
...
+/**
+ * Per-model usage breakdown from the global `__tokenStats.models`
+ * accumulator (populated by `updateTokenStats` on every LLM call,
+ * including subagent/tool branches that pointer-share it). Returned
+ * sorted by cost descending so callers can show the priciest model
+ * first. Unlike `_getCost`/`_getTokens`, this reads the process-wide
+ * total (not the per-branch accumulator), which is the right scope for
+ * a `/cost` summary that attributes spend across every model used.
+ */
+export async function _getModelCosts(): Promise<ModelCost[]> {
```
egonSchiele, line 115: Minor duplication: the "read `__tokenStats.models`, narrow each field with `typeof`, sort by cost-desc then name" logic lives here and again in `readTokenSnapshot` (cli.ts). Not blocking, but a shared helper would keep the defensive narrowing + sort in one place.

packages/agency-lang/lib/stdlib/cli.ts
```diff
@@ -562,7 +591,7 @@ async function printFooter(
   // model is appended only when known.
   if (turn) {
     let stats = `↑${fmtTokens(turn.inputTokens)} ↓${fmtTokens(turn.outputTokens)} ${fmtElapsed(turn.elapsedMs)}`;
-    if (turn.model.length > 0) stats += ` · ${turn.model}`;
+    if (turn.models.length > 0) stats += ` · ${turn.models.join(", ")}`;
```
egonSchiele: Nit: `turn.models.join(", ")` is unbounded — a turn touching many models could produce a long footer line. Typically 1–2 models so this is fine; just noting it isn't capped.

egonSchiele (commented):
Solid, well-scoped fix for a real attribution bug, and the test coverage is excellent — unit tests at both layers plus an e2e agency test that pins the subtle delta-vs-cumulative and tool-branch-aggregation cases.

Verified locally: `lastModel` is fully removed (no lingering readers), the deterministic test client tags calls as `"deterministic"` (matches the agency test expectations), and the builtin is wired through `thread.agency` → `_getModelCosts`.

A few small things below — only the first is worth confirming before merge; the rest are optional polish.

packages/agency-lang/lib/agents/agency-agent/agent.agency
egonSchiele: Good catch — rather than confirm they reconcile, I made it impossible for them to disagree: `/cost` now sums the same process-wide `getModelCosts()` breakdown for its `Cost:`/`Tokens:` header, so header == sum-of-rows by construction. (Dropped the per-branch `getTokens()` read here; `getCost()` stays for the footer status at line 494.) Fixed in 13905f6f500d36204a0416c232e329c695a15d12.

packages/agency-lang/lib/stdlib/cli.ts
egonSchiele: Done — detection now triggers on `tokenDelta > 0 || costDelta > 0`, with a comment noting that current providers always add tokens so the cost check is just belt-and-suspenders. Added a unit test for the cost-grew-but-tokens-didn't case. 13905f6f500d36204a0416c232e329c695a15d12

packages/agency-lang/lib/stdlib/thread.ts
egonSchiele: Extracted into `normalizeModelUsage` in `runtime/utils.ts` (defensive narrowing + cost-desc/name sort in one place); both `_getModelCosts` and `readTokenSnapshot` now call it. 13905f6f500d36204a0416c232e329c695a15d12

packages/agency-lang/lib/stdlib/cli.ts
egonSchiele: Capped — new `fmtModels` shows the top 3 (by spend, since the list is already cost-sorted) and collapses the rest to `+N more`; full breakdown stays available via `/cost`. Tested. 13905f6f500d36204a0416c232e329c695a15d12

PR #327, opened 2026-06-21
Exclude human approval time from the footer's turn time

egonSchiele:
## What

The footer's "time taken" now measures **LLM + tool + agent-compute time only** — it no longer includes the time a human spends responding to a mid-turn approval prompt.

## The bug

The footer reported `Date.now() - turnStartMs` — naive wall-clock spanning the whole `onSubmit(line)` call. Approval interrupts run **inline** (`runHandlerChain`), and their menu (`chooseOption` → `_runPrompt` → `await prompts(...)`) blocks on the human *inside* that window. So if the agent paused for an approval and you took 30s to decide, those 30s were in the reported time. (The time you spend typing the prompt itself was already excluded — `turnStartMs` is set after readline returns.)

## The fix

A process-global **human-wait clock** (`runtime/utils.ts`):

- Every line-mode human prompt funnels through `ui.ts`'s `_runPrompt`, which now wraps its body in `measureHumanWait(...)` to charge time-blocked-on-the-human to the clock (on both resolve and throw paths, so a cancelled approval still doesn't count).
- The footer snapshots the clock before/after the turn — exactly as it already does for tokens — and subtracts the delta from wall-clock, floored at 0 (`computeTurnElapsedMs`).

This generalizes cleanly: **any** wait on a human-input primitive (approval menus, `prompt()`, `askUser`, reject free-text) is excluded; nothing else is.

## Scope

The line-mode REPL footer (what the agency-agent uses). Out of scope by construction:
- Top-level message composition (`askLine`, `/paste`) — outside the turn window already.
- The alt-screen TUI — separate status bar, not this footer.

## Tests

- `lib/runtime/utils.test.ts` — human-wait clock accumulation, ignores non-positive durations, and `measureHumanWait` charges the blocked time on both the resolve and throw paths.
- `lib/stdlib/cli.test.ts` — `computeTurnElapsedMs` subtracts the human-wait delta, no-ops when there was none, and floors at 0.

Coverage boundary: the in-`_runLineRepl` wiring (snapshot + subtract) and the one-line `_runPrompt` wrap aren't unit-tested directly (that path needs a PTY), but both building blocks they compose are. All unit tests pass; existing `ui.test.ts` (28) still passes; structural linter clean.

### DIFF
--- a/packages/agency-lang/lib/runtime/utils.ts
+++ b/packages/agency-lang/lib/runtime/utils.ts
@@ -171,3 +171,39 @@ export function updateTokenStats(args: {
   tokenStats.cost.outputCost += cost.outputCost || 0;
   tokenStats.cost.totalCost += cost.totalCost || 0;
 }
+
+// ── Human-wait clock ──────────────────────────────────────────────────
+// Cumulative milliseconds the process has spent BLOCKED on human input —
+// interrupt approval menus, prompt(), askUser, reject free-text. Every
+// line-mode human prompt funnels through ui.ts's `_runPrompt`, which wraps
+// its body in `measureHumanWait`. The REPL footer snapshots this counter
+// before/after a turn and subtracts the delta from wall-clock elapsed, so
+// the reported "time taken" reflects LLM + tool + agent compute only and
+// never the seconds a human spent deciding. Monotonic and process-global
+// (the single-process REPL serves one human at a time); `performance.now`
+// keeps it immune to wall-clock adjustments.
+let __humanWaitMs = 0;
+
+/** Add a measured human-input blocking duration (ms) to the cumulative
+ *  clock. Non-positive durations are ignored. */
+export function recordHumanWaitMs(ms: number): void {
+  if (ms > 0) __humanWaitMs += ms;
+}
+
+/** Read the cumulative human-wait clock (ms). Snapshot before/after a turn
+ *  and subtract the delta from wall-clock elapsed. */
+export function readHumanWaitMs(): number {
+  return __humanWaitMs;
+}
+
+/** Run `fn`, charging the time it spends blocked to the human-wait clock.
+ *  Records on both the resolve and throw paths (so a cancelled prompt
+ *  still doesn't count human deliberation as compute time). */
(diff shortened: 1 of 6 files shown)


packages/agency-lang/lib/runtime/utils.ts
```diff
@@ -171,3 +171,39 @@ export function updateTokenStats(args: {
...
+// Cumulative milliseconds the process has spent BLOCKED on human input —
+// interrupt approval menus, prompt(), askUser, reject free-text. Every
+// line-mode human prompt funnels through ui.ts's `_runPrompt`, which wraps
+// its body in `measureHumanWait`. The REPL footer snapshots this counter
+// before/after a turn and subtracts the delta from wall-clock elapsed, so
+// the reported "time taken" reflects LLM + tool + agent compute only and
+// never the seconds a human spent deciding. Monotonic and process-global
+// (the single-process REPL serves one human at a time); `performance.now`
+// keeps it immune to wall-clock adjustments.
+let __humanWaitMs = 0;
```
egonSchiele: NEVER keep global state in typescript like this. If you track global state in agency instead, it will create a copy of that global state for each execution of that agent, enabling no-effort state isolation:
https://agency-lang.com/guide/execution-model.html

The way you have it set up here, this code can only ever be used by a single run of the agent concurrently. I actually don't see any reason to put any of this logic in typescript at all. I think all of this could simply be logic inside agency.

egonSchiele: Good call — fixed in 65b1c257836a04eb6fff9a4022eb4bf482ade6ef. Moved the counter into the per-execution `GlobalStore` `__internal` blob (next to the token stats the footer already reads), so each agent run gets its own isolated counter — no shared module state. `recordHumanWaitMs(globals, ms)`/`readHumanWaitMs(globals)` now read/write `__tokenStats.humanWaitMs`; `withTokenStats()` seeds it (and it serializes/restores with the blob for free). The new unit tests assert two stores never share a counter. The remaining TS is just the measurement at the single prompt chokepoint (`_runPrompt`) and the footer subtraction — both inherently TS-side (the prompt blocks in `ui.ts`, the footer prints in `_runLineRepl`) — but the state is fully in the agency global store now.

PR #328, opened 2026-06-21
Grouped, flattened LLM-call view in the logs viewer

egonSchiele:
## Summary

A set of logs-viewer improvements that make `agency logs view` read the way you'd expect, plus the runtime changes that feed it. Built iteratively while reviewing real `foo.agency` / `bar.agency` traces.

Design spec: `packages/agency-lang/docs/superpowers/specs/2026-06-21-llm-call-grouped-view-design.md`.

## What's in it

**1. Grouped, flattened LLM-call view** (`04f51dba`)
One `llm()` call now renders as **one** node showing a single continuous conversation, with tool executions spliced inline and nested `llm()` calls flattened recursively. Runtime: the per-tool-loop-round `llmCall` spans collapse into one span per `llm()` call. Viewer: an `llmCall` span renders the flattened transcript (from the final round, which holds the full thread) with each `toolExecution` spliced after the assistant message that triggered it. Also relabels tool-execution spans `toolExecution` (they were mislabeled `toolCallStart`).

**2. Array tool results render as JSON** (`b12903e1`)
A tool message whose content is the tool's raw return value (e.g. `[0,1,1,2,3]`) was shown as an empty `[tool: …]` because `stringifyContent` treated every array as multimodal content-parts. Now falls back to JSON for non-content-part arrays. (Pre-existing bug, surfaced by the flatten.)

**3. `forkBranchEnd` carries the branch return value** (`0082234f`)
The event recorded only timing/outcome; now it also carries the branch's `value` on success, serialized defensively (caps, drops unserializable/undefined, never throws). The viewer shows `forkBranchEnd #0 (success, 653ms) → 0`.

**4. Identifying detail in collapsed span summaries** (`e80313de`)
Collapsed spans showed only metrics, so parallel calls were indistinguishable. Now each pulls its identity from its characteristic child event: `nodeExecution "agent"`, `agentRun "main"`, `toolExecution fibonacciNumbers`, `forkAll 5 branches`, `embedding recall-query · 1536d`, and `llmCall gpt-4o-mini · "prompt…" → outcome`.

**5. Wall-clock envelope durations** (`d337bb4d`)
Span duration summed `timeTaken` across the whole subtree, so an `llmCall` that forked N parallel LLM calls reported ~N× its real time (17.2s vs 5.6s) — longer than its own node. Replaced with a wall-clock envelope (earliest start to latest end, start = `timestamp − timeTaken`): correct single-round latency, true multi-event wall-clock, no parallel summing, and parent ⊇ child holds.

**6. Test coverage** (`c253298b`)
Unit tests for `safeStatelogValue` (the defensive serializer's edge cases) and a new `render.e2e.test.ts` that drives a realistic `llm → forking tool` log through the whole pipeline (`buildForest → flattenVisibleRows → renderRowText`), asserting all of the above together.

## Before / After (bar.agency)

Before: two/many sibling `llmCall` nodes per call, blank tool results, indistinguishable parallel calls, a 17.2s `llmCall` inside a 5.6s node.

After:
```
nodeExecution "agent" (5.6s, 565 tok)
  llmCall gpt-4o-mini · "Use the fibonacciNumbers tool…" → {"response":[0,1,1,2,3]} (5.6s, 565 tok)
    toolExecution fibonacciNumbers (3.0s, 310 tok)
      forkAll 5 branches (3.0s, 310 tok)
        llmCall gpt-4o-mini · "Calculate the 0th Fibonacci…" → {"response":0} (2.8s, 62 tok)
        … (each parallel call distinct)
        forkBranchEnd #0 (success, 3.0s) → 0
    [tool: fibonacciNumbers] [0,1,1,2,3]
    [assistant] {"response":[0,1,1,2,3]}
```

## Compatibility

No backward compatibility for old logs (per the spec) — they degrade to roughly the prior multi-node look; no `STATELOG_FORMAT_VERSION` bump. Regenerate logs to get the grouped view.

## Testing

~1040 runtime + logsViewer + statelog unit tests pass, plus end-to-end agency-js tests (`llm-call-single-span`, `fork-branch-value`) and the new `render.e2e.test.ts`. Verified end-to-end against real `foo.agency` / `bar.agency` traces.

### DIFF
--- a/packages/agency-lang/lib/logsViewer/render.e2e.test.ts
+++ b/packages/agency-lang/lib/logsViewer/render.e2e.test.ts
@@ -1,0 +1,175 @@
+import { describe, it, expect } from "vitest";
+import { buildForest } from "./tree.js";
+import { flattenVisibleRows, renderRowText } from "./render.js";
+import { EventEnvelope, TreeNode, ViewerState } from "./types.js";
+
+// End-to-end: feed a realistic statelog through the WHOLE viewer
+// pipeline (buildForest → flattenVisibleRows → renderRowText) and assert
+// the composed behavior of every recent change at once:
+//   - the grouped/flattened llmCall view with the tool execution
+//     spliced inline,
+//   - a non-string (array) tool result rendered as JSON,
+//   - identifying detail in span summaries (node/tool names, branch
+//     count, model + prompt + outcome),
+//   - forkBranchEnd values,
+//   - wall-clock envelope durations with parent ⊇ child.
+//
+// Scenario (mirrors bar.agency, shrunk to 2 fork branches):
+//   agent → llm("use fibonacciNumbers") → tool forks 2 parallel llms.
+
+const E = (
+  spanId: string | null,
+  parentSpanId: string | null,
+  data: any,
+): EventEnvelope => ({
+  format_version: 1,
+  trace_id: "t1",
+  project_id: "p",
+  span_id: spanId,
+  parent_span_id: parentSpanId,
+  data,
+});
+
+const TS = (sec: string) => `2026-06-21T00:00:0${sec}Z`;
+const TOOLCALL = { id: "c1", name: "fibonacciNumbers", arguments: { num: 2 } };
+
+const events: EventEnvelope[] = [
+  E("A", null, { type: "agentStart", timestamp: TS("0.000"), entryNode: "main" }),
+  E("N", "A", { type: "enterNode", timestamp: TS("0.100"), nodeId: "agent" }),
+  // Round 1: the model asks for the tool.
+  E("L", "N", {
+    type: "promptCompletion",
+    timestamp: TS("1.000"),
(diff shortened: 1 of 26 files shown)


packages/agency-lang/lib/logsViewer/render.e2e.test.ts
```diff
@@ -0,0 +1,175 @@
...
+//   - a non-string (array) tool result rendered as JSON,
+//   - identifying detail in span summaries (node/tool names, branch
+//     count, model + prompt + outcome),
+//   - forkBranchEnd values,
+//   - wall-clock envelope durations with parent ⊇ child.
+//
+// Scenario (mirrors bar.agency, shrunk to 2 fork branches):
+//   agent → llm("use fibonacciNumbers") → tool forks 2 parallel llms.
+
+const E = (
```
egonSchiele, line 20: more descriptive var name please

PR #329, opened 2026-06-22
fix(guards): carry structured AbortCause so time-guard trips don't crash CI

egonSchiele:
## What & why

Three time-guard fixtures have been **crashing on CI** ever since Esc-cancel landed:
`guard-time-trip`, `guard-time-nested`, `guard-time-and-cost` ([failing run](https://github.com/egonSchiele/agency-lang/actions/runs/27907963302/job/82579909174)).

**Root cause:** a `TimeGuard` timer aborts an in-flight `sleep`, which rejected with a **bare** `AgencyCancelledError` carrying no intent. Before Esc-cancel the generated catch ladders swallowed that into a `Failure`; after Esc-cancel they *re-throw* aborts (so user cancels propagate cleanly), so the bare cancel escaped the guarded block as an **unhandled rejection and killed the process**.

This is Increment 1 of a larger design ([spec](packages/agency-lang/docs/superpowers/specs/2026-06-21-abort-taxonomy-design.md), [plan](packages/agency-lang/docs/superpowers/plans/2026-06-22-abort-taxonomy-increment1.md)): **carry intent on the abort instead of re-deriving it at every catch site.**

## The fix

- **`AbortCause` carrier** (`errors.ts`): a tagged union (`userInterrupt | userKill | guardTrip | raceLoser | cleanup`) carried on the `AgencyCancelledError`'s new `agencyCause` field; `readCause()` reads it back from a signal or an error. `signal.reason` stays an `Error` (so `runBatch`'s `throw signal.reason` is unaffected).
- **Time-guard trips are recognizable:** `TimeGuard` aborts with a `guardTrip` cause; the leaf `sleep` propagates it; **`__tryCall` converts a `guardTrip` cause to a `Failure` — checked BEFORE the blanket `isAbortError → throw`** (the actual fix: `isAbortError` is `true` for the cause-carrying cancel too, so order matters).
- **Two-path de-dup:** implementation surfaced a *second* delivery path — the runner's `shouldSkip` throws `GuardExceededError` at the guard's own `_popGuard` step. A shared mutable `delivered` flag on the cause makes the leaf-op and runner paths cooperate so each trip is delivered exactly once. (Documented in spec §4.1.1.)
- **Foundation (additive):** leaf ops propagate the cause; `ctx.cancel()`/`cleanup()` and the REPL Esc handler carry `userKill`/`userInterrupt`/`cleanup` causes.

## Scope

- **No codegen/template change, no fixture regeneration.** `AgencyCancelledError` and `GuardExceededError` stay distinct — collapsing them into one `AgencyAbort` is **Increment 2** (separate PR).
- **Deferred (documented):** `prompt.ts` cause-driven thread-repair gating + normalization (changes the agent-cancel cleanup path; wants PTY coverage), and explicit `raceLoser` tagging (current untagged behavior is preserved).

## Testing

- +7 unit tests (`errors.test.ts` cause round-trips; `guard.test.ts` time-guard emits a `guardTrip` cause).
- Full vitest suite **6086 passed**; the 3 previously-failing fixtures + full guards dir (20) + race fixtures green; `lint:structure` clean.
- Verified the original crash is gone (no unhandled rejection).

### DIFF
--- a/packages/agency-lang/lib/runtime/errors.ts
+++ b/packages/agency-lang/lib/runtime/errors.ts
@@ -41,10 +41,87 @@ export class ConcurrentInterruptError extends Error {
   }
 }
 
+/**
+ * Structured reason carried by every abort — on `AbortController.abort(cause)`
+ * (so it surfaces as `signal.reason`) and on the `agencyCause` field of a
+ * thrown `AgencyCancelledError`. The `kind` discriminant lets every
+ * catch/boundary READ the intent instead of re-deriving it (which is what the
+ * runner's guard-sniff and the leaf bare-throws used to do). See
+ * docs/superpowers/specs/2026-06-21-abort-taxonomy-design.md.
+ */
+export type AbortCause =
+  | { kind: "userInterrupt" }
+  | { kind: "userKill"; reason?: string }
+  | {
+      kind: "guardTrip";
+      dimension: "cost" | "time";
+      limit: number;
+      spent: number;
+      guardId: string;
+      /**
+       * Mutable de-dup flag. A time-guard trip can be delivered by EITHER
+       * an aborted leaf op (→ `__tryCall` converts to a Failure) OR the
+       * runner's `shouldSkip` (→ throws `GuardExceededError`). Whichever
+       * fires first sets this; the other then knows the trip is already
+       * handled and must not re-deliver. The cause object has stable
+       * identity across the composed signal (`AbortSignal.any` adopts the
+       * source's reason object), so this single flag is visible to both
+       * paths. See docs/superpowers/specs/2026-06-21-abort-taxonomy-design.md §3.4.
+       */
+      delivered?: boolean;
+    }
+  | { kind: "raceLoser" }
+  | { kind: "cleanup" };
+
+/** Brand so a plain object on `signal.reason` is recognizable as ours. */
+const ABORT_CAUSE_BRAND = "__agencyAbortCause";
+
+export function makeAbortCause(
+  cause: AbortCause,
+): AbortCause & { [ABORT_CAUSE_BRAND]: true } {
+  return { ...cause, [ABORT_CAUSE_BRAND]: true } as AbortCause & {
+    [ABORT_CAUSE_BRAND]: true;
+  };
+}
+
+function isAbortCause(value: unknown): value is AbortCause {
+  return (
+    value != null &&
+    typeof value === "object" &&
+    (value as Record<string, unknown>)[ABORT_CAUSE_BRAND] === true &&
+    typeof (value as Record<string, unknown>).kind === "string"
+  );
+}
+
+/**
+ * Read the structured `AbortCause` off an `AbortSignal` (its `reason`), a
+ * thrown `AgencyCancelledError` (its `agencyCause`), or a bare cause value.
+ * Returns `undefined` when no structured cause is present — callers keep
+ * their existing heuristics as a fallback for that case.
+ */
+export function readCause(source: unknown): AbortCause | undefined {
+  if (source == null) return undefined;
+  if (typeof AbortSignal !== "undefined" && source instanceof AbortSignal) {
+    // `signal.reason` is kept as an Error (so `throw signal.reason` in
+    // runBatch stays an Error), with the structured cause carried on it —
+    // but a bare branded cause is also accepted for forward-compat.
+    return readCause(source.reason);
+  }
+  if (source instanceof AgencyCancelledError && source.agencyCause) {
+    return source.agencyCause;
+  }
+  if (isAbortCause(source)) return source;
+  return undefined;
+}
+
 export class AgencyCancelledError extends Error {
-  constructor(reason?: string) {
+  /** Structured intent for this abort, when known. */
+  readonly agencyCause?: AbortCause;
+
+  constructor(reason?: string, cause?: AbortCause) {
     super(reason ?? "Agent execution was cancelled");
     this.name = "AgencyCancelledError";
+    this.agencyCause = cause;
   }
 }
 
--- a/packages/agency-lang/lib/runtime/errors.ts
+++ b/packages/agency-lang/lib/runtime/errors.ts
@@ -98,6 +98,15 @@ function isAbortCause(value: unknown): value is AbortCause {
  * thrown `AgencyCancelledError` (its `agencyCause`), or a bare cause value.
  * Returns `undefined` when no structured cause is present — callers keep
  * their existing heuristics as a fallback for that case.
+ *
+ * Note: this uses `instanceof AgencyCancelledError`, NOT the name-based
+ * fallback that `isAbortError` adds for errors that cross module instances
+ * (the resolver shim — see below). In Increment 1 every guard-trip
+ * producer and consumer lives in one runtime module instance, so
+ * `instanceof` holds. If a cause ever needs reading across that boundary,
+ * `readCause` returns `undefined` where `isAbortError` would still
+ * recognize the error — the in-process sibling of the subprocess
+ * cause-payload risk noted in the abort-taxonomy spec.
  */
 export function readCause(source: unknown): AbortCause | undefined {
   if (source == null) return undefined;
--- a/packages/agency-lang/lib/runtime/guard.ts
+++ b/packages/agency-lang/lib/runtime/guard.ts
@@ -3,9 +3,14 @@ import { AgencyCancelledError, makeAbortCause } from "./errors.js";
 
 /** Monotonic source of stable per-guard ids. Threaded into the
  *  `guardTrip` AbortCause a TimeGuard emits so boundaries can identify
- *  which guard tripped (diagnostics + future cross-checks; structural
- *  call-stack nesting is what actually routes a trip to its owning
- *  `try`, exactly as it does for cost guards today). */
+ *  which guard tripped. NOTE: in Increment 1 no boundary actually
+ *  MATCHES on this id — `__tryCall` converts any guardTrip cause it
+ *  catches, relying on structural call-stack nesting to route a trip to
+ *  its owning `try` (exactly as cost guards do today). Real id-matching
+ *  (which would fix outer-tighter-than-inner mis-attribution; see
+ *  tests/agency/guards/guard-time-nested-outer-tighter) needs threading
+ *  the id into `__tryCall` — a codegen change deferred to Increment 2.
+ *  The id is emitted now so the cause shape is stable for that work. */
 let __guardIdCounter = 0;
 export function nextGuardId(): string {
   __guardIdCounter += 1;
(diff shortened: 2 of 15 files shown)


packages/agency-lang/lib/runtime/guard.ts
```diff
@@ -250,6 +262,9 @@ export class TimeGuard implements Guard {
    *  steps (popGuard) on its way out. */
   private consumed: boolean = false;
 
+  /** Stable id threaded into the emitted `guardTrip` cause. */
+  readonly guardId: string = nextGuardId();
```
egonSchiele, line 271: **`guardId` is emitted but never matched anywhere** (medium). `__tryCall` — the only guard `try` boundary (it's the codegen for `as`/`try` at `typescriptBuilder.ts:1066`) — receives no guardId and converts *any* `guardTrip` cause it catches. So this field is currently dead weight.

The spec's Concern-3 resolution says guardId matching "must still be implemented, not assumed," but the comment here is honest that structural nesting does the routing "exactly as cost guards do today." Real matching would need threading the id into `__tryCall` — a codegen change Increment 1 deliberately avoids, so deferring is defensible. But please reconcile: update the spec to say matching is deferred (and why) rather than leaving a documented-but-unfulfilled invariant + an unused field.

Behavioral consequence to lock down: the nested fixture only tests the *inner* (tight) guard tripping, which the innermost `__tryCall` catches correctly. The untested case is **outer trips while inside the inner block with an in-flight leaf op** — there the inner `__tryCall` catches the outer's trip and converts it into the *inner* guard's Failure (carrying the outer's numbers), so the outer block sees a normal return instead of tripping. Not a regression (it crashed before), but now silently mis-attributed. Add a fixture (outer tighter than inner) to pin the intended behavior.

packages/agency-lang/lib/runtime/result.ts
```diff
@@ -73,6 +107,27 @@ export async function __tryCall(fn: () => any, opts?: FailureOpts): Promise<Resu
...
+    // the cancel first, the conversion never runs, and the bare cancel
+    // escapes the guarded block as an unhandled rejection (the bug this
+    // fixes). See docs/superpowers/specs/2026-06-21-abort-taxonomy-design.md.
+    const guardCause = readCause(error);
+    if (guardCause?.kind === "guardTrip") {
+      // Mark the trip delivered so the runner's `shouldSkip` (which may
+      // step the guard's own `_popGuard` next, while the signal is still
+      // aborted) does NOT re-throw a GuardExceededError for the same
+      // trip. The cause object is shared by identity with `signal.reason`.
+      guardCause.delivered = true;
```
egonSchiele, line 125: The `delivered` two-path de-dup is the subtlest new logic in the PR (it's the *second* crash you discovered), yet it has no direct unit test — neither `result.ts` nor `runner.ts` got a test file, so it's exercised only through the three agency fixtures.

Those fixtures are a fine end-to-end regression lock, but a deterministic unit test (fake timers: time guard + in-flight `sleep` → assert `timeoutFailure` Result, assert no unhandled rejection, and assert the guard's own `_popGuard` step doesn't re-throw) would lock the exact CI crash at the unit level — which is what the spec's acceptance criteria called for — and would specifically protect this `delivered` cooperation.

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -268,6 +268,17 @@ export class Runner {
...
     if (this.stack?.abortSignal?.aborted && !this.halted) {
+      // If the abort carries a guard trip that was ALREADY delivered via
+      // the leaf-op path (an in-flight sleep/fetch aborted, then __tryCall
+      // converted it to a Failure), the trip is handled. Do NOT re-throw
+      // it here — this `shouldSkip` may be gating the guard's own
+      // `_popGuard` cleanup step, and throwing would surface an unhandled
+      // GuardExceededError for a trip the user already saw as a Failure.
+      // Fall through so cleanup runs. See the abort-taxonomy spec.
+      const cause = readCause(this.stack.abortSignal);
+      if (cause?.kind === "guardTrip" && cause.delivered) {
```
egonSchiele, line 279: Narrower than spec §4.1 step 5, which wanted a full cause-switch here (`userInterrupt`/`userKill` → propagate, no silent halt). This only adds the `delivered`-guardTrip early-return; user causes still fall through to the existing `halt(undefined)` below. That's fine for Increment 1 (Esc propagates via the leaf/catch-ladder path, not via `shouldSkip`), just flagging that the runner's user-cancel behavior is unchanged despite the spec describing a change.

packages/agency-lang/lib/runtime/errors.ts
```diff
@@ -41,10 +41,87 @@ export class ConcurrentInterruptError extends Error {
...
+ */
+export function readCause(source: unknown): AbortCause | undefined {
+  if (source == null) return undefined;
+  if (typeof AbortSignal !== "undefined" && source instanceof AbortSignal) {
+    // `signal.reason` is kept as an Error (so `throw signal.reason` in
+    // runBatch stays an Error), with the structured cause carried on it —
+    // but a bare branded cause is also accepted for forward-compat.
+    return readCause(source.reason);
+  }
+  if (source instanceof AgencyCancelledError && source.agencyCause) {
```
egonSchiele, line 119: Minor asymmetry: `readCause` uses `instanceof AgencyCancelledError`, but `isAbortError` deliberately adds a name-based fallback (`errors.ts:84`) for errors that cross module instances (the resolver shim). For Increment 1 every guard-trip producer/consumer is in one runtime module instance so `instanceof` holds and this is moot — but if a cause ever needs reading across that boundary, `readCause` will silently return `undefined` where `isAbortError` would still recognize the error. This is the in-process sibling of the subprocess-payload case already noted in the spec's risks; one sentence in the code or risks would close the loop.

egonSchiele (commented):
## Review: structured AbortCause (Increment 1)

Solid, ship-able Increment 1. It faithfully implements the plan and resolves all four concerns from the spec review — most importantly the `__tryCall` ordering (guardTrip check now sits *above* `isAbortError → throw`, with a comment explaining why). The carrier is clean and additive, and keeping `signal.reason` an `Error` so `runBatch`'s `throw signal.reason` is unaffected is the right call.

The findings below are about gaps and doc/impl drift, not blockers. Net: **approve** with two recommendations — reconcile the `guardId` story + add an outer-trips-during-inner fixture, and add a unit-level regression test for the `delivered` de-dup.

### Done well
- The ordering fix is correct and well-commented — the one thing that had to be right.
- Identity-based `delivered` de-dup is sound: `readCause` never clones, so both the leaf-op and runner paths mutate the same branded object on `signal.reason`'s `agencyCause` (identity holds through `AbortSignal.any`).
- `guardFailureData` extraction removes the duplicated Failure-shape logic cleanly.
- Scope discipline is genuinely good — no codegen/fixture churn, exactly as promised.

### Cross-cutting note: implementation is narrower than spec §4
The plan's "Deferred" section is honest, but the spec §4 body still lists as Increment 1 several things that didn't land (and don't need to to turn CI green):
- `runner.shouldSkip` full cause-switch — spec step 5 says `userInterrupt`/`userKill` should "propagate (no silent halt)"; the actual change only adds the `delivered`-guardTrip early-return, and user causes still fall through to the existing silent `halt`. Fine (Esc still works via the leaf/catch-ladder path), but the spec oversells what shipped.
- `prompt.ts` thread-repair gating + normalization (steps 5–6) and `raceLoser` tagging (step 2) are deferred — safe (untagged → `readCause` undefined → existing behavior), documented in the plan, just out of sync with the spec body.

Worth a one-line pointer in the spec so a future reader doesn't think §4 fully shipped.

Inline comments below for the specific items.

packages/agency-lang/lib/runtime/guard.ts
egonSchiele: Reconciled in 49aa99ac. Confirmed your behavioral claim empirically (outer 20ms / inner 5s → \`inner:timeoutFailure:20\`, mis-attributed) and pinned it as `tests/agency/guards/guard-time-nested-outer-tighter` with a docstring naming it a known Increment-1 limitation. Spec §4 now states guardId matching is deferred to Increment 2 (it needs threading the id into `__tryCall` = a codegen change this increment avoids); the `guard.ts` comment says the same. Kept the field emitted so the cause shape is stable for that work.

packages/agency-lang/lib/runtime/result.ts
egonSchiele: Added in 49aa99ac. `result.test.ts` now unit-tests the ordering fix directly (a guardTrip-cause-carrying abort → `timeoutFailure`/`guardFailure` rather than a throw; sets `delivered`; non-guard + bare aborts still re-throw). `runner.test.ts` covers the de-dup: `shouldSkip` throws `GuardExceededError` for an UNdelivered trip but does NOT re-throw an already-`delivered` one (fake timers, real `TimeGuard`) — locking the exact second crash at the unit level.

packages/agency-lang/lib/runtime/runner.ts
egonSchiele: Agreed — flagged explicitly in 49aa99ac. Spec §4 now has a "what shipped vs deferred" note: the shipped `shouldSkip` only adds the `delivered`-guardTrip early-return; `userInterrupt`/`userKill` still fall through to the existing silent `halt`, and Esc propagates via the leaf/catch-ladder path, not `shouldSkip`. The runner user-cancel behavior is intentionally unchanged for Increment 1.

packages/agency-lang/lib/runtime/errors.ts
egonSchiele: Documented in 49aa99ac — added a note on `readCause` that it is `instanceof`-based and (unlike `isAbortError`) has no name-based cross-module fallback, so it returns `undefined` for a cause that crossed a module instance. Called out as the in-process sibling of the subprocess cause-payload risk already in the spec. Left as a doc note since in Increment 1 every guard-trip producer/consumer is in one module instance.

egonSchiele:
Review addressed in 49aa99ac (tests + docs only, no production-logic change from the prior commit):

- **delivered de-dup / ordering** now have direct unit tests (`result.test.ts`, `runner.test.ts`) — the exact CI crash + its second variant are locked at the unit level, not just via fixtures.
- **guardId / nested mis-attribution**: empirically confirmed and pinned as `guard-time-nested-outer-tighter`; spec + `guard.ts` now state guardId matching is deferred to Increment 2 (codegen change).
- **Cross-cutting "spec §4 oversells what shipped"**: added a "what shipped vs deferred" callout at the top of spec §4 (partial `shouldSkip` cause-switch, deferred guardId matching, deferred `prompt.ts`/`raceLoser`). The plan's Deferred section remains authoritative on scope.

Full vitest 6093 green; guards dir 21; lint clean.

PR #330, opened 2026-06-22
Run served functions inside an Agency execution frame

egonSchiele:
## Problem

`agency serve http foo.agency`, then `POST /function/greet` (or the MCP equivalent), failed for **every** exported function:

```
ERROR function greet threw: getRuntimeContext() called outside an Agency execution frame.
```

Generated function bodies assume an ambient Agency execution frame — their first lines read `getRuntimeContext().ctx`, the base `StateStack`/`ThreadStore` via `setupFunction()`, and globals via `__globals()`. In normal execution a function only runs inside a node body (or an LLM tool-call that is itself inside a node), so the frame is already installed. **Nodes** work over `serve` because each generated node export installs the frame via `runNode`; the serve adapters invoked **functions** cold with `AgencyFunction.invoke()`, so no frame existed.

This affected both the HTTP and MCP adapters. The existing adapter unit tests didn't catch it because they construct fake `AgencyFunction.create({ fn: (a,b) => a+b })` JS bodies that never touch the runtime context.

## Fix

Route function calls through a node-grade frame:

- **runtime** (`node.ts`): new `runExportedFunction` mirrors `runNode`'s bootstrap (cross-module init → globals → top-level callbacks) against a fresh execution context, then runs the call inside `agencyStore.run({ctx, stack, threads, globals}, …)` with a real `ThreadStore` (so functions using `llm()` / message threads work). No `agentStart`/`agentEnd` — a bare function call is a stateless RPC.
- **codegen** (`imports.mustache`): emit a per-module `__invokeFunction` bound to that helper — the same pattern as the existing `respondToInterrupts` / `rewindFrom` exports.
- **serve** (`discovery.ts`, adapters): discovery attaches a frame-aware `invoke` to each exported function, falling back to the bare call for modules compiled before `__invokeFunction` existed; both adapters now call `fn.invoke(args)`.

## Testing

- New `lib/serve/http/functionFrame.integration.test.ts` compiles a real module (with a `static const` read, so the body depends on bootstrap init having run) and drives it through the actual HTTP handler — the only way to exercise the regression, since the adapter unit tests use fakes.
- Verified end-to-end by hand: `POST /function/greet {"name":"foo"}` → `{"success":true,"value":"Hello, foo!"}`.
- Full vitest suite green (6098 passed); structural lint clean. Generator/builder/debugger fixtures regenerated for the one-line preamble addition.

### DIFF
--- a/packages/agency-lang/lib/runtime/node.ts
+++ b/packages/agency-lang/lib/runtime/node.ts
@@ -89,6 +89,105 @@ export function setupFunction(): {
 }
 
 /**
+ * Run the fresh-run bootstrap on a freshly created execution context:
+ * cross-module statics/globals, then this module's globals, then top-level
+ * callback registration — each inside a bootstrap ALS frame.
+ *
+ * Shared by the two fresh-run entry points (`runNode`, `runExportedFunction`).
+ * The resume family (`respondToInterrupts`, `rewindFrom`) does NOT use this:
+ * it restores statics/globals from a checkpoint and only re-registers
+ * top-level callbacks.
+ */
+async function initFreshExecCtx(
+  execCtx: RuntimeContext<GraphState>,
+  opts: {
+    initializeGlobals?: (ctx: RuntimeContext<GraphState>) => void | Promise<void>;
+    registerTopLevelCallbacks?: (ctx: RuntimeContext<GraphState>) => void | Promise<void>;
+    moduleDir?: string;
+  },
+): Promise<void> {
+  const { initializeGlobals, registerTopLevelCallbacks, moduleDir } = opts;
+
+  // initializeGlobals + registerTopLevelCallbacks both invoke Agency
+  // code that goes through `__call` — and `__call` reads `ctx` /
+  // `threads` / `stateStack` from the ALS frame after the
+  // drop-per-call-context-plumbing migration. Without an ALS frame
+  // installed here, calls to user-defined stdlib helpers (e.g. the
+  // `callback(...)` wrapper that the codegen emits inside
+  // `__registerTopLevelCallbacks`) would invoke `_callbackImpl(name,
+  // fn, undefined)` and crash on `__state.ctx`.
+  //
+  // Consequences worth knowing:
+  //  - `stack` in the bootstrap frame is `execCtx.stateStack` — the
+  //    bare global stack with no node/function frames pushed. Globals
+  //    must not push node frames, so this is correct.
+  //  - `threads` is a `BootstrapThreadStore` sentinel. Any agency code
+  //    in global-init scope that reaches for a thread/message builtin
+  //    (e.g. `systemMessage("…")` at module top-level) now throws with
+  //    a clear error instead of silently writing into a placeholder
+  //    that this function discards on return.
+  //  - The `insideGlobalInit` codegen branch still emits an explicit
+  //    `{ ctx }` bag on `__call`, and `__call`'s merge prefers extras
+  //    over the ALS-read fields — so `ctx` resolution inside generated
+  //    global-init code does not depend on this frame's `ctx` either.
+  //    The frame is mostly here to satisfy the "every helper must see
+  //    *some* frame" contract.
+  //
+  // Closure-wide bootstrap (`__initAllRegistered`) MUST run BEFORE the
+  // entry's own `initializeGlobals`:
+  //   • A global-init expression in the entry can call a function /
+  //     node whose body reads an imported `static const`. Those
+  //     function-body reads don't show up in the compile-time
+  //     per-variable dep graph (which only walks initializer
+  //     expressions), so the foreign module's `__initializeGlobals`
+  //     would not have been pulled in by the dep-driven prelude. If
+  //     we ran the entry's globals init first, the read could hit
+  //     `__UNINIT_STATIC` and throw — the exact failure mode this
+  //     ordering exists to prevent.
+  //   • Running `__initAllRegistered` first guarantees every JS-loaded
+  //     module's statics and globals have been initialized before any
+  //     user code (including global-init expressions) runs. The
+  //     subsequent `initializeGlobals(execCtx)` call is then a no-op
+  //     for current codegen (per-execCtx early-return guard) and
+  //     double-call protected by the registry-level
+  //     `globals.isInitialized` check baked into `__initAllRegistered`.
+  await runInBootstrapFrame(execCtx, () => __initAllRegistered(execCtx), { moduleDir });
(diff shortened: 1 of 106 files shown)


packages/agency-lang/lib/runtime/node.ts
```diff
@@ -87,6 +88,86 @@ export function setupFunction(): {
...
+        // promises) before returning, mirroring runNode's awaitAll.
+        await execCtx.pendingPromises.awaitAll();
+        return result;
+      },
+    );
+  } finally {
+    // Best-effort telemetry flush; never fail the call on a telemetry
+    // problem.
+    await execCtx.statelogClient.flush();
+    execCtx.cleanup();
```
egonSchiele: **Minor parity gap:** this omits `runNode`'s MemoryManager save-loop. `runNode`'s `finally` iterates `execCtx.getAllCachedMemoryManagers()` and calls `manager.save()`; here we only flush statelog + `cleanup()`.

Low risk in practice — `remember`/`forget`/extraction all `persist()` eagerly (manager.ts:518, 842), so a served function that writes memory won't silently lose data. But any manager state that relies solely on the end-of-run flush would be dropped for served functions. Worth either mirroring the loop, or a one-line comment noting the divergence is intentional.

packages/agency-lang/lib/runtime/node.ts
```diff
@@ -87,6 +88,86 @@ export function setupFunction(): {
...
+  await runInBootstrapFrame(execCtx, () => __initAllRegistered(execCtx), { moduleDir });
+  if (initializeGlobals) {
+    await runInBootstrapFrame(execCtx, () => initializeGlobals(execCtx), { moduleDir });
+  }
+  if (registerTopLevelCallbacks) {
+    await runInBootstrapFrame(execCtx, () => registerTopLevelCallbacks(execCtx), { moduleDir });
+  }
+
+  const threadStore = ThreadStore.withDefaultActive(execCtx.statelogClient);
+  try {
```
egonSchiele, line 243: **Minor observability gap:** no `catch` arm here, so a throwing served function emits no statelog `error` event (and no partial-token `agentEnd`), unlike `runNode`. The error still reaches the client via the adapter's `logger.error` + error response, but tracing/telemetry never records the failure.

packages/agency-lang/lib/runtime/node.ts
```diff
@@ -87,6 +88,86 @@ export function setupFunction(): {
...
+    return await agencyStore.run(
+      {
+        ctx: execCtx,
+        stack: execCtx.stateStack,
+        threads: threadStore,
+        globals: execCtx.globals,
+        moduleDir,
+      },
+      async () => {
+        const result = await fn.invoke({
```
egonSchiele, line 253: **Confirm intent:** no `agentEnd`/token roll-up means LLM spend inside a served function isn't attributed (cost dashboards / `getCost`). The doc comment calls this intentional ("stateless RPC"), just flagging since served functions *can* call `llm()` and the cost will be invisible.

packages/agency-lang/lib/serve/discovery.ts
```diff
@@ -11,17 +11,50 @@ export type DiscoverOptions = {
...
+ * already-compiled bundles.
+ */
+function makeInvoker(
+  fn: AgencyFunction,
+  moduleInvoke: ModuleInvokeFunction | undefined,
+): (namedArgs: Record<string, unknown>) => Promise<unknown> {
+  if (moduleInvoke) {
+    return (namedArgs) => moduleInvoke(fn, namedArgs);
+  }
+  return (namedArgs) => fn.invoke({ type: "named", positionalArgs: [], namedArgs });
```
egonSchiele, line 49: **Informational:** this fallback silently reintroduces the exact `getRuntimeContext() called outside an Agency execution frame` error for modules compiled before `__invokeFunction` existed — stale bundles keep throwing until recompiled. Reasonable graceful-degradation choice; noting it's a re-expose rather than a fix for old output.

egonSchiele (commented):
Clean, well-targeted fix with genuinely good test coverage — the new compile-and-serve integration tests exercise the real regression that the fake-body unit tests couldn't. I verified concurrency isolation (`createExecutionContext` builds a fresh isolated context per request), bootstrap-ordering parity with `runNode`, the fallback path, and that the regenerated fixtures match the template change. No high-severity correctness bug.

The inline notes below are all **minor parity gaps** between `runExportedFunction` and `runNode`'s `finally`/`catch` — none blocking. The memory save-loop divergence is the one most worth addressing (or explicitly documenting).

(For context: the integration test's `await import(pathToFileURL(...))` is *not* a CLAUDE.md "no dynamic imports" violation — loading a freshly-compiled module is the established convention across the codebase, including `lib/cli/serve.ts` itself.)

packages/agency-lang/lib/runtime/node.ts
egonSchiele: Mirrored runNode's teardown loop in `1bc463b8`: the `finally` now iterates `getAllCachedMemoryManagers()` and `save()`s each (best-effort — logged, never thrown) before the statelog flush. Contract is now identical for both entry points.

packages/agency-lang/lib/runtime/node.ts
egonSchiele: Left the catch out deliberately and documented why (`1bc463b8`). The generated function body already emits the statelog `error` event and converts exceptions to a `Failure`, so per-call failures are traced and reach the client. The only throws that escape the body are control signals (`RestoreSignal`/abort/`GuardExceeded`); a catch here would risk swallowing them, and runNode handles those as part of its restore loop — which doesn't apply to a one-shot RPC.

packages/agency-lang/lib/runtime/node.ts
egonSchiele: Confirmed intentional, now spelled out in the doc comment (`1bc463b8`): no agentStart/agentEnd span or run-level token roll-up, so `llm()` spend inside a served function isn't attributed to `getCost`/dashboards. Leaving it as a documented limitation of the stateless-RPC model rather than synthesizing a run envelope — happy to add token roll-up if you want serve calls to show up in cost tracking.

packages/agency-lang/lib/serve/discovery.ts
egonSchiele: Agreed — it's a re-expose, not a fix for old output. Expanded the comment (`1bc463b8`) to say so: a pre-`__invokeFunction` module keeps throwing the frame error on function calls until recompiled; the fallback only avoids a hard crash and keeps the plain-JS bodies used in the adapter unit tests working.

PR #332, opened 2026-06-23
Value-parameterized validator factory

egonSchiele:
## Summary

Makes `@validate` on a value-parameterized type alias (e.g. `NumberInRange(low, high)` in `std::types`) resolve its validator functions in the alias's **defining** module. Importing the alias from another module now validates correctly without injecting anything into the consumer's scope.

### The problem

Value-parameterized validated aliases previously emitted **nothing** at their definition site and inlined a fresh validator chain at every use site — leaking bare `min`/`max` identifiers into the consumer's scope. This caused:
- a crash (`ReferenceError`) if the consumer didn't import the validators, and
- a **silent wrong-answer** if the consumer defined its own same-named `min`/`max` (validation would bind to the wrong function).

### The fix

Emit a **descriptor factory function** in the alias's own module, where the validators are already in scope:

```js
export function NumberInRange(low, high) {
  return { kind: "leaf", schema: z.number().meta({ minimum: low, maximum: high }),
           validators: [min.partial({ n: low }), max.partial({ n: high })] };
}
```

Use sites just **call** it: `__validateChainRecursive(v, NumberInRange(1, 10))`. This mirrors how bare validated aliases already work (their `__agency_descriptor` const lives in the defining module).

## Changes

- **Task 1** — `processTypeAlias` emits an (exported) descriptor factory for validated value-param aliases, with value-param **defaults baked into the factory signature** (`function Age(low = 0)`) so omitted args still resolve.
- **Task 2** — use sites of validated value-param aliases reference the factory by **call** instead of inlining; use-site `@validate` validators concatenate on top.
- **Task 4** — validated value-param aliases are now **importable** (factory is exported); they are never registered as LLM tools.
- **Bug fix found during execution** — building a *forwarding* alias's factory body (`type AdultAge(high) = NumberInRange(18, high)`) must **not** deep-resolve/substitute the body. Doing so tripped `applyValueArgs`'s unsubstituted-param guard when a forwarded name collided with the inner param name (`high` → `high`), and would have re-introduced the cross-module leak. The body is now passed unresolved so nested value-param refs become nested factory **calls**. `hasAnyValidateTag` no longer calls `applyValueArgs` (presence of `@validate` is substitution-invariant). Surfaced by the `valueParamPartialForwarding` / `valueParamWrappingAlias` fixtures.

## Tests

New agency execution fixtures under `tests/agency/validation/`:
- `valueParamFactoryLocal` — local value-param validated alias validates end-to-end.
- `stdTypesValueParamImport` — importing `NumberInRange`/`StringWithLength` from `std::types` validates correctly; no `min`/`max` leak.
- `valueParamImportLocalShadow` — a consumer-defined `min` cannot hijack an imported alias's validation (the core motivation; `tooLow:false`).
- `valueParamForwardImported` — a forwarding alias over an imported value-param alias validates via nested factory calls.

New unit tests in `typescriptBuilder.integration.test.ts` cover factory emission, default-in-signature, schema-identifier survival (`@jsonSchema`), and use-site call.

## Verification

- `pnpm test:run` — **6104 passed**, 0 failed.
- Full `tests/agency/validation` agency suite — **83 passed**, 0 failed.
- `make` — builds clean (stdlib recompiled with the new factories).

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -91,7 +91,9 @@ import {
 import {
   buildValidationDescriptor,
   hasAnyValidateTag,
+  hasAliasValidate,
 } from "./typescriptGenerator/validationDescriptor.js";
+import { tagArgToTs } from "./typescriptGenerator/tagArgToTs.js";
 import { resolveTypeDeep } from "../typeChecker/assignability.js";
 import { isFunctionTyped } from "../typeChecker/utils.js";
 
@@ -677,12 +679,48 @@ export class TypeScriptBuilder {
       if (!node.exported) return ts.empty();
       return ts.raw(`export const ${node.aliasName} = undefined;`);
     }
-    // Value-parameterized aliases (e.g. `type NumberInRange(low, high) = ...`)
-    // have no single schema — every use-site inlines a fresh substituted
-    // schema. Emit nothing at the declaration site; importers should never
-    // reference the bare name (use-sites are resolved before codegen).
+    // Value-parameterized aliases have no single schema — every use-site
+    // depends on the value args. For a VALIDATED one we emit a descriptor
+    // FACTORY here, in the alias's own module, so the validators it
+    // references stay in scope (exactly like a bare alias's
+    // `__agency_descriptor`). Use sites call `AliasName(args)` (see
+    // validationDescriptor.ts). A NON-validated value-param alias has no
+    // validators to leak, so it keeps emitting nothing and its schema is
+    // inlined at the use site.
     if (node.valueParams && node.valueParams.length > 0) {
-      return ts.empty();
+      const vpAliasesFull = this.scopes.visibleTypeAliasesFull();
+      const vpAliasedWithTags: VariableType = node.tags
+        ? {
+            ...node.aliasedType,
+            tags: [...(node.aliasedType.tags ?? []), ...node.tags],
+          }
+        : node.aliasedType;
+      if (!hasAnyValidateTag(vpAliasedWithTags, vpAliasesFull)) {
+        return ts.empty();
+      }
+      const vpResolved = resolveTypeDeep(vpAliasedWithTags, vpAliasesFull);
+      const vpDescriptor = buildValidationDescriptor(
+        vpResolved,
+        this.scopes.visibleTypeAliases(),
+        vpAliasesFull,
+      );
+      // Bake value-param DEFAULTS into the factory signature so an omitted
+      // use-site arg (e.g. `Age()!` or bare `Age!`) resolves the same default
+      // the old inline path got via `applyValueArgs` (which fills
+      // `bindings[p.name] = p.default`). Without this, `function Age(low)`
+      // called as `Age()` leaves `low` undefined and
+      // `min.partial({ n: undefined })` silently mis-validates.
+      const vpParams = node.valueParams
+        .map((p) =>
+          p.default !== undefined
+            ? `${p.name} = ${tagArgToTs(p.default)}`
+            : p.name,
+        )
+        .join(", ");
+      const vpExportPrefix = node.exported ? "export " : "";
+      return ts.raw(
+        `${vpExportPrefix}function ${node.aliasName}(${vpParams}) { return ${printTs(vpDescriptor)}; }`,
+      );
     }
     const exportPrefix = node.exported ? "export " : "";
     // Thread alias-level @validate / @jsonSchema tags onto the body type so
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -721,16 +721,16 @@ export class TypeScriptBuilder {
       // `bindings[p.name] = p.default`). Without this, `function Age(low)`
       // called as `Age()` leaves `low` undefined and
       // `min.partial({ n: undefined })` silently mis-validates.
-      const vpParams = node.valueParams
-        .map((p) =>
-          p.default !== undefined
-            ? `${p.name} = ${tagArgToTs(p.default)}`
-            : p.name,
-        )
-        .join(", ");
-      const vpExportPrefix = node.exported ? "export " : "";
-      return ts.raw(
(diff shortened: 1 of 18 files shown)


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -677,12 +679,59 @@ export class TypeScriptBuilder {
...
+      // `min.partial({ n: undefined })` silently mis-validates.
+      const vpParams = node.valueParams
+        .map((p) =>
+          p.default !== undefined
+            ? `${p.name} = ${tagArgToTs(p.default)}`
+            : p.name,
+        )
+        .join(", ");
+      const vpExportPrefix = node.exported ? "export " : "";
+      return ts.raw(
```
egonSchiele: should this use a different builder instead of ts.raw?

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -677,12 +679,59 @@ export class TypeScriptBuilder {
...
-    // reference the bare name (use-sites are resolved before codegen).
+    // Value-parameterized aliases have no single schema — every use-site
+    // depends on the value args. For a VALIDATED one we emit a descriptor
+    // FACTORY here, in the alias's own module, so the validators it
+    // references stay in scope (exactly like a bare alias's
+    // `__agency_descriptor`). Use sites call `AliasName(args)` (see
+    // validationDescriptor.ts). A NON-validated value-param alias has no
+    // validators to leak, so it keeps emitting nothing and its schema is
+    // inlined at the use site.
     if (node.valueParams && node.valueParams.length > 0) {
```
egonSchiele, line 690: A **validated alias that combines type params and value params** (`@validate(...) type Foo<T>(n) = ...`) skips this block entirely and crashes at use sites.

The `node.typeParams.length > 0` early-return higher up in `processTypeAlias` runs *before* this value-param factory block, so such an alias emits `export const Foo = undefined` and never a factory. But it still has `valueParams`, so at a `!` use site `isValueParamInstantiation` is true and `valueParamDescriptor` emits a factory **call** `Foo(args)` → `undefined(args)` → `TypeError: Foo is not a function` at runtime. The import path agrees with the crash (`isInlinedAlias` returns false for `typeParams > 0`, so it imports the `undefined` const).

The combined construct is real and tested (`valueParamWithGeneric.agency`, `valueParamGenericWrappingAlias.agency` use `type BoundedList<T>(n) = T[]`), but those use only `@jsonSchema` so they take the inline path. There's no fixture adding `@validate` to a generic value-param alias — that's the gap. Note this is also a *regression* for the local case: pre-PR such an alias inlined its validators and worked locally; post-PR it routes to a non-existent factory.

Niche (requires `@validate` + generics + value-params together), but worth either handling generics in the factory branch or explicitly rejecting the combination with a clear error.

packages/agency-lang/lib/backends/typescriptGenerator/validationDescriptor.ts
```diff
@@ -153,6 +155,69 @@ function schemaNode(
...
+): TsNode {
+  if (useSiteValidators.length === 0) return base;
+  const existingValidators = ts.binOp(
+    ts.prop(base, "validators", { optional: true }),
+    "??",
+    ts.arr([]),
+    { parenLeft: true },
+  );
+  return ts.obj([
+    ts.setSpread(base),
```
egonSchiele: The factory call is **evaluated twice** when use-site validators are present. When `base` is a factory call (`ts.raw("NumberInRange(1, 10)")`), this helper both spreads it here (`ts.setSpread(base)`) and reads `base.validators` above (`ts.prop(base, ...)`), so the printed output is:

```js
{ ...NumberInRange(1, 10), validators: [...(NumberInRange(1, 10)?.validators ?? []), ...useSite] }
```

Correct (the factory is pure), but it rebuilds the whole descriptor — including the `min.partial(...)`/`max.partial(...)` `AgencyFunction` allocations — twice per validation site. The old `__agency_descriptor` path shared this helper too, but there `base` was a cheap property read; promoting it to a function call makes the double-evaluation costlier. Binding the call to a local (or wrapping in a memoizing IIFE) would avoid it. Minor efficiency only.

egonSchiele (commented):
Two findings from a review of the factory change. Both are low-to-moderate severity; the PR is well-structured and the new fixtures cover the core motivation directly.

egonSchiele:
Thanks for the review! All three findings addressed in 708035b1.

**Finding 1 (generic + value-param + `@validate` crash).** Confirmed and fixed. The direct use site actually already worked (`resolveTypeDeep` flattens it before codegen), but the *forwarded/nested* case did crash — and the actual failure was a "Unresolved generic type at codegen" in the plain-schema path, because `hasAnyValidateTag`'s `genericType` case only inspected `typeArgs` and missed the aliased generic's own `@validate`, routing it away from the descriptor path. Three coordinated fixes:
- `hasAnyValidateTag` now looks up the generic alias entry and checks its tags/body (mirrors the `typeAliasVariable` case).
- `descriptor()` / `valueParamDescriptor` resolve-and-inline a user-defined generic alias instead of trying to factory-call it (generics can't be runtime factories — the schema depends on the type arg).
- `applyValueArgs`'s guard no longer false-positives when a forwarded identifier collides in name with an inner param (e.g. `Wrapped(high) = Ranked<string>(1, high)`).

New fixture `valueParamGenericValidate.agency` covers both the direct and forwarded cases (with the name collision). The forwarding alias now inlines `min.partial({ n: 1 }), max.partial({ n: high })` — no `undefined(...)` call.

**Finding 2 (factory call evaluated twice).** Fixed. `withUseSiteValidators` now binds the base to a local via an IIFE — `((__d) => ({ ...__d, validators: [...(__d?.validators ?? []), ...useSite] }))(base)` — so the factory (and its `.partial(...)` allocations) is evaluated once. Added a unit test asserting the single-eval form.

**Finding 3 (`ts.raw` for the factory decl).** Switched to the structured `ts.functionDecl` builder; value-param defaults go through `TsParam.defaultValue`. Generated output is unchanged.

Verification: `pnpm test:run` 6106 passed, full `tests/agency/validation` 84 passed, `lint:structure` clean.

PR #333, opened 2026-06-23
Finish & harden optimizers: expected-aware default judge, breakdown + validation parity

egonSchiele:
## Summary

A focused "finish & harden" pass that makes the three shipped optimizer features (custom graders, validation sets, the DX report) behave **consistently across all built-in optimizers**, and closes a surprising gap in the default judge.

## What changed

1. **Default judge grades against `expected`.** Previously, setting `input.expected` without a custom grader was silently ignored — the built-in `LlmJudge` only used `goal`. Now the bundled `goalJudge` agent takes an optional `expected` param and `LlmJudge` passes the input's `expected` (default `expectedPath: ["expected"]`), so providing a gold answer tightens grading even with no custom grader. (Empty when absent; a custom judge node that ignores it is unaffected.)
2. **Champion grade breakdown for `gepa` and `example`.** Only `greedy` set `result.championBreakdown`, so `report.md`/`champion/grades.json` (the reward-hacking lens) were empty for the others. Now all three populate it.
3. **`gepa`/`example` select the champion by validation.** They previously reported a validation objective but still selected on train. Champion selection is now hoisted into a shared `BaseOptimizer.pickValidationChampion` helper that `greedy`, `gepa`, and `example` all use, so validation-based selection is uniform.

Supporting cleanup: the report's "validation provided but unused" note is now derived from whether a validation objective was produced (robust for custom optimizers that ignore validation), replacing the name-based heuristic. Docs (`optimize.md`, `writing-optimizers.md`) updated.

## Test plan

- [x] Unit: `LlmJudge` passes `expected` as the third judge arg; `gepa`/`example` emit `championBreakdown` and a `validationObjective` when a validation set is given; greedy's existing validation-selection test still passes through the shared helper.
- [x] `pnpm run build` (tsc + make, incl. recompiled `goalJudge`) clean; `pnpm run lint:structure` clean; 310 tests pass across `lib/optimize` + `lib/cli/eval`.
- [ ] Full agency execution suite — deferred to CI.

### DIFF
--- a/packages/agency-lang/docs/site/cli/optimize.md
+++ b/packages/agency-lang/docs/site/cli/optimize.md
@@ -140,7 +140,7 @@ Also returns a binary pass/fail like exact match, but this one checks to see if 
 Calculates the levenshtein distance and returns a score between 0 and 1 (0 = no match, 1 = perfect match).
 
 #### LLM Judge
-Asks an LLM to return a score between 0 and 1 (0 = no match, 1 = perfect match) for how well the response matches the expected output.
+Asks an LLM to return a score between 0 and 1 (0 = no match, 1 = perfect match) for how well the response matches the goal — and, when an input sets `expected`, grades against that gold answer too (so `expected` tightens the default judge even without a custom grader).
 
 This is the default grader.
 
@@ -235,7 +235,7 @@ Every grade counts: a number contributes its value (0..1), and a boolean / `Exac
 
 ## Validation sets
 
-Pass `--validation-inputs <file|dir>` to grade the champion against held-out inputs, or `--validation-split <ratio>` to hold out a seeded fraction of `--inputs`. Search and candidate acceptance run on the **training** inputs; with the default `greedy` optimizer the champion written back is the one with the best **validation** objective, and `report.md` shows train-vs-validation side by side so an overfit prompt (high train, flat validation) is visible.
+Pass `--validation-inputs <file|dir>` to grade the champion against held-out inputs, or `--validation-split <ratio>` to hold out a seeded fraction of `--inputs`. Search and candidate acceptance run on the **training** inputs; the champion written back is the one with the best **validation** objective, and `report.md` shows train-vs-validation side by side so an overfit prompt (high train, flat validation) is visible. All built-in optimizers (`greedy`, `gepa`, `example`) select by validation; a custom optimizer that doesn't will say so in the report.
 
 ## Configuration
 
--- a/packages/agency-lang/lib/optimize/optimizers/example.ts
+++ b/packages/agency-lang/lib/optimize/optimizers/example.ts
@@ -1,10 +1,20 @@
 import { BaseOptimizer, type BaseOptimizerDeps } from "../baseOptimizer.js";
+import { breakdown } from "../gradeBreakdown.js";
+import type { Scorecard } from "../grading/scorecard.js";
 import type { Input } from "../grading/types.js";
 import { proposeMutation, type ProposeMutationArgs } from "../mutator.js";
 import type { BaseOptimizerConfig } from "../optimizer.js";
 import { defaultPreview, type OptimizeMutationOperation, type OptimizeMutationPreview } from "../sourceMutator.js";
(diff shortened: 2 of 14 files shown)


packages/agency-lang/docs/site/cli/optimize.md
```diff
@@ -235,7 +235,7 @@ Every grade counts: a number contributes its value (0..1), and a boolean / `Exac
 
 ## Validation sets
 
-Pass `--validation-inputs <file|dir>` to grade the champion against held-out inputs, or `--validation-split <ratio>` to hold out a seeded fraction of `--inputs`. Search and candidate acceptance run on the **training** inputs; with the default `greedy` optimizer the champion written back is the one with the best **validation** objective, and `report.md` shows train-vs-validation side by side so an overfit prompt (high train, flat validation) is visible.
+Pass `--validation-inputs <file|dir>` to grade the champion against held-out inputs, or `--validation-split <ratio>` to hold out a seeded fraction of `--inputs`. Search and candidate acceptance run on the **training** inputs; the champion written back is the one with the best **validation** objective, and `report.md` shows train-vs-validation side by side so an overfit prompt (high train, flat validation) is visible. All built-in optimizers (`greedy`, `gepa`, `example`) select by validation; a custom optimizer that doesn't will say so in the report.
```
egonSchiele, line 238: what does "select by validation" mean?

packages/agency-lang/lib/optimize/optimizers/example.ts
```diff
@@ -74,36 +82,37 @@ export class ExampleOptimizer extends BaseOptimizer {
       (operations) => (this.exampleDeps.preview ?? defaultPreview)(source, operations),
     );
 
-    // 3. If we got a valid proposal, score it and keep it only if it beats the baseline.
+    // 3. Keep the candidate only if it beats the baseline on the training objective.
```
egonSchiele: this is really awful imperative code.

egonSchiele:
## Review notes

Clean refactor overall — `pickValidationChampion`'s candidate lists are sound (`pool.best()` is always in `[baseline, ...accepted]`), the `validationConfiguredButUnused` derivation correctly keys off `validationObjective`, and adding an optional positional `expected` param to the judge node is backward-compatible (the evaluate shim calls the compiled JS function, so extra args are harmless for custom 2-arg judges). One parity gap and two notes.

### 1. `lib/optimize/grading/functionGrader.ts:31-32` — `judge` helper not updated for the expected-aware judge

`FunctionGrader._run` invokes the *same* bundled `goalJudge.agency`, but only passes `[goal, output]` — never `expected`:

```ts
const judge = ({ goal, output }: { goal: string; output?: JSON }) =>
  runAgency.runStructured(goalJudgeFile(), "main", [goal, asJudgeText(output ?? run.output)], ScalarVerdict);
```

Its `judge` signature (`{ goal, output? }`) doesn't even allow passing `expected`, yet the `GraderContext` doc comment three lines up states *"the gold answer is `input.expected`"* and `input` is in scope.

**Consequence:** a user who sets `input.expected` and writes a custom metric that calls `ctx.judge({ goal })` gets goal-only grading, while the default `LlmJudge` on the same input now grades against the gold answer — the two judge paths disagree on exactly the feature this PR set out to make consistent.

**Fix:** forward `asJudgeText(input.expected ?? "")` as the third arg (and optionally expose it on the `judge` context).

### Lower-severity notes (not blocking)

- **Behavior change for existing configs (intended, but worth a changelog line):** any current eval that sets `input.expected` while relying on the default `LlmJudge` will now have its grades change — outputs that don't state the expected answer score near 0.0 where before `expected` was inert. Some users may have populated `expected` for other graders/documentation and will see optimization results shift silently.

- **Minor `trainObjective` inconsistency:** `example.ts:313` reports `objectiveOf(champion.scorecard)` (gate-aware, 0 on a failed gate) while `gepa.ts:115` and `greedyReflective.ts:85` use raw `champion.scorecard.objective()`. In practice the written-back champion always passes gates (baseline via `requireBaselineGatesPass`; accepted candidates passed their gate check), so they agree today — but the divergence is a latent trap if that invariant is ever relaxed.

PR #334, opened 2026-06-23
Abort taxonomy: unification + guardId matching (Increment 2) + deferred carrier pieces

egonSchiele:
## Abort taxonomy — Increment 2 (unification + guardId matching) + deferred Increment-1 carrier pieces

Finishes the abort-taxonomy work from `docs/superpowers/specs/2026-06-21-abort-taxonomy-design.md`. Plan: `docs/superpowers/plans/2026-06-22-abort-unification-and-deferred.md`.

### Phase A — deferred Increment-1 carrier pieces
- **A1:** both `runBatch` abort sites tag losers with a `raceLoser` cause (signal.reason stays an Error, now `readCause`-able).
- **A2:** `markThreadCancelled` is now **non-destructive** — it stubs only the trailing assistant's unanswered `tool_calls` (preserving earlier rounds + the dangling assistant's text) instead of truncating to the last user message. Repair is gated behind `needsThreadRepair(cause)` so it fires only for user-initiated cancels, never a guard trip / race loser. Dispatch normalization now prefers the structured cause.

### Phase B — unification: one `AgencyAbort`, one catch rung
- **B1:** `AgencyAbort extends Error { agencyCause }` is the single abort base; `AgencyCancelledError` and `GuardExceededError` become subclasses. `isAbortError`/`readCause` updated (incl. cross-module name fallback).
- **B2/B3:** the codegen function- and node-level catch ladders collapse from two rungs (`GuardExceededError` + `__isAbortError`) to one `instanceof AgencyAbort` rung. Tracked golden snapshots regenerated; a new cross-function-boundary fixture backstops it. (Compiled `.agency`->`.js` fixtures are gitignored and recompiled per run, so there is no 126-file churn commit.)
- **B4:** deletes the now-redundant `isGuardExceededError` branch in `__tryCall`; `hooks.ts` callback catch widened to `AgencyAbort` (a guard trip in a callback was previously swallowed); `http.ts` carries the cause through the fetch boundary.

### Phase C — `guardId` matching (fixes nested mis-attribution)
- **C1:** every guard carries a `guardId`; both `CostGuard.check` and `TimeGuard.check` thread it into the trip; `_pushGuard` returns the ids.
- **C2:** `__tryCall` gains `ownedGuardIds` — a `guardTrip` converts only when its `guardId` is owned, else re-throws to its real owner. The stdlib `guard` runs its block via `_runGuarded` (invoking through `__call` and forwarding the same `FailureOpts` the codegen `try` injected). Fixes outer-tighter-than-inner mis-attribution for **both time and cost**. `guardId` is serialized in `GuardJSON` so the match survives interrupt/resume.

### Notable findings during implementation
- Pre-existing `tsc` errors in `functionGrader.test.ts` blocked the build tree-wide; fixed in an isolated first commit (approved).
- The plan's `markThreadCancelled` snippet had `smoltalk.toolMessage` args reversed and used `.tool_calls` instead of `.toolCalls`; corrected against the real smoltalk API.
- `_runGuarded` must invoke the block via `__call` (not a plain `block()`) or the guard-trip cause is lost — caught by `guard-time-trip` failing.
- `guardId` was not serialized, so it changed across interrupt/resume and broke `ownedGuardIds` matching (uncaught escape) — caught by `guard-cost-shared-survives-interrupt`; fixed by serializing it.
- The agency-js thread-repair contrast test (A2 Step 6) was **intentionally skipped** per the plan's fallback: the harness has no primitive to inject a `userInterrupt` mid-tool-call and inspecting thread state after `main()` throws has no precedent. The gating is locked instead by the `needsThreadRepair` predicate unit test (all 6 cause kinds) + the four `markThreadCancelled` repair-shape unit tests.

### Verification
- `make` clean; full unit suite **6134 passed / 0 failed**.
- Agency fixtures: `guards` (25, incl. flipped + 3 new), `fork/race` (6), `handlers` (51), `interrupts` (16) — all green.
- `lint:structure` clean; freshly compiled fixture shows the single `instanceof AgencyAbort` rung.

Spec `§4`/`§5` updated to mark unification + guardId matching + the deferred carrier pieces as shipped; `connectionLost` / `callTimeout` remain future work.


packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -371,24 +381,69 @@ async function _runPrompt({
...
+/** Thread repair (stubbing a dangling assistant tool-call turn) is only
+ *  needed for a user-initiated cancel mid-LLM-turn. A guard trip or a
+ *  race-loser abort doesn't leave the thread mid-tool-call in a way the
+ *  next user turn would choke on, and repairing then would discard a turn
+ *  the guard's Failure path still wants intact. Unknown / absent causes
+ *  default to repair (conservative: matches pre-cause behavior). This is
+ *  the single source of truth for the repair policy — future cause variants
+ *  (connectionLost, callTimeout) decide their policy HERE, not at the catch
+ *  site. */
+function needsThreadRepair(cause: AbortCause | undefined): boolean {
```
egonSchiele: This file is already so big. If needsThreadRepair and markThreadCancelled are pure functions and don't need to be in this file, I would suggest putting them in a separate file.

packages/agency-lang/lib/stdlib/thread.ts
```diff
@@ -189,13 +196,47 @@ export async function __internal_popGuard(
...
+ * THIS guard()'s guards (an outer guard's trip re-throws past it).
+ *
+ * Replaces the stdlib `guard`'s former agency-level `try block()`. That `try`
+ * lowered (in a function scope) to `__tryCall(() => block(), { checkpoint,
+ * functionName, args })`; we MUST forward the same FailureOpts so a guard
+ * failure keeps its checkpoint / functionName / args (retry + reporting
+ * depend on them) — only `ownedGuardIds` is added. See
+ * processTryExpression in lib/backends/typescriptBuilder.ts.
+ */
+export async function _runGuarded(
```
egonSchiele, line 233: Does this code need to be in TypeScript? Or can it just live in agency? There is no requirement that every agency function needs to be backed by a TypeScript function,  and this code to me looks like it could live purely in agency.

egonSchiele: `_runGuarded` has to be TS because the routing it adds is `__tryCall(() => __call(block, ...), { ownedGuardIds: ids, checkpoint, functionName, args })`, and `ownedGuardIds` is what makes a guard convert only *its own* trip. The agency `try block()` expression lowers (in a function scope) to `__tryCall(() => block(), { checkpoint, functionName, args })` — there's no syntax to pass `ownedGuardIds` through it.

Putting the owned ids on the stack frame instead and having `__tryCall` read them globally doesn't work either: a plain `try` nested *inside* the guarded block must own nothing (so it re-throws the guard's trip rather than swallowing it — see `guard-trip-not-swallowed-by-inner-try`). The owned set therefore has to be scoped to this specific `try` boundary, not read from the frame.

So the options are (a) this small TS seam, or (b) extend the `try` codegen to carry owned-guard-ids — a bigger codegen change I kept out of this PR. Happy to do (b) as a follow-up if you'd rather `guard()` stay pure agency; otherwise `_runGuarded` is the minimal seam. The push/pop themselves stay in agency.

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: Done in d6bd4c39 — moved `markThreadCancelled` + `needsThreadRepair` to a new `lib/runtime/threadRepair.ts` (with `threadRepair.test.ts`); `prompt.ts` imports them now. Pure move, no behavior change.

packages/agency-lang/lib/runtime/errors.ts
```diff
@@ -160,15 +182,26 @@ export class HandlerRecursionError extends Error {
...
+  // `name` so abort identity survives even though `agencyCause` may not
+  // survive serialization (spec §9). Note the `name === X || name === Y`
+  // form — a chained `X || "Y" || "Z"` would always be truthy.
+  if (error instanceof Error) {
+    return (
+      error.name === "AgencyAbort" ||
+      error.name === "AgencyCancelledError" ||
+      error.name === "AbortError"
+    );
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in d6bd4c39 — added `"GuardExceededError"` to `isAbortError`'s cross-module name fallback (a guard trip crossing the subprocess shim loses its prototype chain but keeps its name), plus a test for a cross-realm trip.

packages/agency-lang/lib/runtime/errors.test.ts
```diff
@@ -2,12 +2,55 @@ import { describe, it, expect } from "vitest";
...
+    expect(isAbortError(e)).toBe(true);
+    expect(readCause(e)).toBe(cause);
+  });
+
+  it("AgencyCancelledError and GuardExceededError are AgencyAbort subclasses", () => {
+    expect(new AgencyCancelledError()).toBeInstanceOf(AgencyAbort);
+    const g = new GuardExceededError("time", 20, 21);
+    expect(g).toBeInstanceOf(AgencyAbort);
+    expect(isGuardExceededError(g)).toBe(true);
+    expect(g.type).toBe("time");
```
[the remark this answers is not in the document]
egonSchiele: Applied in d6bd4c39 — the test now constructs the `GuardExceededError` with an explicit `"g-test"` id and asserts it round-trips on `agencyCause.guardId`.

packages/agency-lang/lib/stdlib/http.ts
```diff
@@ -84,7 +88,12 @@ export async function runHttp<T>(fn: () => Promise<T>, url: string): Promise<T>
     return await fn();
   } catch (e) {
     if (isAbortError(e)) {
-      throw new AgencyCancelledError(`fetch ${url} cancelled`);
+      // Carry the structured cause off the rejection (it IS the signal's
+      // reason). A guard trip that aborted this fetch surfaces as an
+      // AgencyCancelledError whose guardTrip cause must survive so the
+      // owning guard's `try` converts it instead of letting a bare cancel
```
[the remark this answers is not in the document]
egonSchiele: Hardened in d6bd4c39 — `runHttp` now falls back to reading the cause off the active runtime abort signal when the rejection itself carries none. One correction to the premise: I checked empirically and current Node `fetch` rejects with `signal.reason` directly (not a fresh `DOMException`), so for a guard-aborted fetch the `guardTrip` cause is already on `e` via the primary `readCause(e)` path. The fallback covers the bare-`DOMException`-with-no-reason edge.

PR #336, opened 2026-06-24
Resilient LLM calls: retries, per-call timeouts, and notification hooks

egonSchiele:
## Resilient LLM calls — retries, per-call timeouts, notification hooks

Transient LLM failures (dropped connections, `429`/`5xx`/`529`, or a call that just hangs) are common and usually self-heal. This makes `llm()` retry them automatically with exponential backoff and an optional per-call deadline — entirely in the backend, so the happy path is unchanged and authors don't have to check every call.

Spec: `docs/superpowers/specs/2026-06-23-resilient-llm-calls-design.md`. Plan: `docs/superpowers/plans/2026-06-23-resilient-llm-calls.md`. Depends on smoltalk 0.4.2 (`status`/`headers`/`cause` on `SmolError`).

### What you get

```ts
llm("Summarize this", {
  retries: 2,                                  // default 2; 0 disables
  timeout: 30s,                                // per-attempt deadline; default 10min; 0 disables
  backoff: { initial: 500ms, factor: 2, max: 10s },   // exponential, capped (defaults)
})
```

- **Retried:** connection drops, `5xx`, `429` (honoring server `retry-after`), `529` overloaded.
- **Not retried:** `400`/auth/content-policy/context-window (terminal); user cancels and guard trips (propagate immediately — never swallowed). A `guard(time:)` keeps ticking through backoff and still wins.
- **After exhaustion / `retries: 0`:** surfaces a normal `Failure` you can handle with `try`/`isFailure` — it does **not** abort the run.
- **Esc during a backoff wait** cancels the whole loop.
- Set branch-wide defaults with `setLlmOptions({ retries, timeout, backoff })`. New `onLLMRetry` / `onLLMTimeout` callbacks for surfacing progress.

### How it's built

- **Provider-neutral classification.** The LLM client is swappable, so retry policy never imports a provider SDK: a new `NormalizedLLMError` (`{ status?, retryAfterMs?, kind?, message }`) is produced by the client adapter (`SmoltalkClient.normalizeError` is the only place that knows smoltalk's error classes). `classifyLlmError` reads only the neutral shape; a custom client without `normalizeError` degrades to message-matching.
- **Pure policy, thin loop.** `classifyLlmError` (3-way: retryable / terminal / abort) and `decideRetry` are pure and exhaustively unit-tested; `runWithRetry` in `prompt.ts` is a thin driver.
- **Per-call timeout** = a call-scoped mini-`TimeGuard` (`armCallTimeout`): `setTimeout → controller.abort(callTimeout)` composed onto the call's signal.
- **Config** resolved in `runPrompt` from the per-call options (on `clientConfig`) + the branch-scoped `stack.other.llmDefaults` bag (the same path `model`/`temperature` use) + built-in defaults.
- No TS module globals — per-execution isolation preserved.

### Notable deviations from the plan (found during execution)

1. **Residual failure is a `Failure`, not an abort.** The plan/first cut threw the exhausted-retry failure as an `AgencyAbort` carrying an `llmFailure` cause — but the catch ladders re-throw any `AgencyAbort` (propagate-never-swallow), so a network blip surviving retries would have **aborted the whole run** instead of becoming a handleable `Failure`. Fixed: surface a plain `Error` (classified in its message) that the catch ladder converts to a `Failure`, exactly as a transient LLM error is today. `llmFailure` was therefore removed from the `AbortCause` union (it was never an abort); `callTimeout` stays (it's the cause on the call's `AbortController` while retrying).
2. **Config flows through `runPrompt`, not `agencyLlm.llm()`.** Agency-source `llm()` compiles to a *direct* `runPrompt()` call (codegen), bypassing the TS `agency.llm` wrapper — so per-call options must be resolved in `runPrompt` (from `clientConfig`). Caught by the `llm-timeout` e2e (it hung on the default 10-min timeout until fixed).
3. **`agency.json` `llm` defaults dropped from this PR.** Seeding them needs the *generated* `RuntimeContext` construction to thread `config.llm` (a codegen change) — left as a follow-up. `setLlmOptions` already provides global config from agency code, so the capability isn't lost.

### Verification

- `make` clean; full unit suite **6174 passed**; `lint:structure` clean.
- Fixtures: `guards` (25), `fork/race` (6) green.
- New deterministic e2e: `tests/agency-js/llm-retry` (survives 2 ECONNRESETs → "pong") and `tests/agency-js/llm-timeout` (hung call hits its 50ms deadline → catchable Failure).


packages/agency-lang/lib/runtime/agencyLlm.ts
```diff
@@ -69,8 +76,14 @@ export async function llm(prompt: string, opts: LlmOpts = {}): Promise<any> {
   // Passing `{ model: undefined }` would still let `runPrompt`'s merge
   // with smoltalkDefaults pick up the default, but being explicit keeps
   // the contract obvious: omit means "don't touch the model".
-  const clientConfig: { model?: string } = {};
+  // Build clientConfig with `model` only when explicitly overridden, plus the
+  // resilience options — runPrompt reads retries/timeout/backoff off clientConfig
+  // (the same path the llm() codegen uses) and resolves the effective policy.
+  const clientConfig: RetryConfig & { model?: string } = {};
   if (opts.model !== undefined) clientConfig.model = opts.model;
+  if (opts.retries !== undefined) clientConfig.retries = opts.retries;
```
egonSchiele: What would happen if you had this exact same code, but without the if condition? Would the behavior be any different?

packages/agency-lang/lib/runtime/llmClient.ts
```diff
@@ -107,3 +158,20 @@ export class SmoltalkClient implements LLMClient {
     } as Omit<SmolConfig, "stream">;
   }
 }
+
+function parseRetryAfter(
```
egonSchiele: I thought smoltalk added this functionality. Did you check the last couple commits?

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -148,13 +152,186 @@ function assertUniqueToolNames(tools: { name: string }[]): void {
...
+ * lives in the pure `decideRetry`; this loop is the thin driver.
+ */
+async function runWithRetry<T>(
+  dispatch: (signal: AbortSignal | undefined) => Promise<T>,
+  policy: RetryPolicy,
+  parentSignal: AbortSignal | undefined,
+  hooks: RetryHooks,
+  normalizeError: (err: unknown) => NormalizedLLMError,
+): Promise<T> {
+  for (let attempt = 0; ; attempt += 1) {
```
egonSchiele: maybe a while loop would be a better fit?

PR #340, opened 2026-06-25
Local-model support via custom provider modules

egonSchiele:
## Local-model support via custom provider modules

Lets any Agency program (and the agent) run against a custom or local LLM provider — e.g. a local model via the globally-installed `smoltalk-llama-cpp` — without `agency-lang` depending on that package and without editing the prebuilt binary. This is the final piece (requirement #5) of the user-selectable-models effort; model *selection* already shipped, so this PR adds only the load-and-register seam.

### How it works
A user writes a tiny ES module exporting `register({ registerProvider })`:

```js
// llama-setup.mjs
import { LlamaCPP } from "smoltalk-llama-cpp";
export function register({ registerProvider }) {
  registerProvider("llama-cpp", LlamaCPP);
}
```

Agency discovers module paths from `agency.json` (`client.providerModules`) and the `AGENCY_PROVIDER_MODULES` env var (merged, de-duped), then — in the shared run **and** resume bootstrap, before any `llm()` call — dynamically `import()`s each module and calls its `register` with **Agency's own** `smoltalk.registerProvider`. Injecting our `registerProvider` (rather than letting the module import it from smoltalk) guarantees the provider lands in the instance the runtime resolves against, which matters because smoltalk is a *peer* dependency of provider packages and a global install can carry a second copy. A process-level guard makes the load idempotent (load-bearing for long-running `serve`).

### What's in scope
- `client.providerModules` config (type + zod) and codegen bake into `RuntimeContext`.
- `loadProviderModules` (config+env merge, cwd-relative resolution, the single sanctioned dynamic `import()` behind one lint-excepted helper, `register` injection, fail-loud errors).
- Wired into both the fresh-run bootstrap (`initFreshExecCtx`, covering `runNode` + `serve`) and the resume path (`respondToInterrupts`).

### Out of scope (already works)
- Model selection (`--provider`/`--model`, `setModel`, `setLlmOptions`).
- Passing model config like the `.gguf` dir — flows through per-call `llm({ metadata: { llamaCppModelDir } })`.

### Compatibility
- **`serve`**: hooks the shared `initFreshExecCtx`; registers once per server process.
- **`pack`**: esbuild bundles our smoltalk (so registration lands correctly in the artifact) and leaves the runtime-computed `import()` as a real dynamic import (the user's module stays external). Covered by a pack smoke test.

### Tests
- Unit: `loadProviderModules` (success, env merge, dedupe, 3 fail-loud error modes).
- Codegen: `agency.json providerModules` → baked `RuntimeContext` args.
- agency-js end-to-end: `agency.json` → baked → loaded during a real compiled run → provider registered (asserts registration, so it is independent of the harness's deterministic-LLM mode).
- `pack` smoke test.
- Full unit suite (6196) green; lint clean.

Includes the design spec and implementation plan under `docs/superpowers/`, and a user guide at `docs/site/guide/custom-providers.md`.


packages/agency-lang/docs/site/.vitepress/config.mts
```diff
@@ -23,6 +23,10 @@ export default defineConfig({
             { text: "Types", link: "/guide/types" },
             { text: "Nodes", link: "/guide/nodes" },
             { text: "LLMs", link: "/guide/llm" },
+            {
+              text: "Custom & local model providers",
```
egonSchiele, line 27: why is this in here twice?

PR #344, opened 2026-06-27
feat(local-model): --local-model flag + agency local CLI suite

egonSchiele:
## Summary

Ships a complete local-model toolkit:

- `agency local` CLI subtree: `list`, `download`, `remove`, `resolve`, `alias add/list/remove`.
- `agency agent --local-model <name>`: one-flag download-and-run; sets both fast AND slow model to the local one so deep subagents stay local too. Overrides `--model`/`--provider`/`--fastmodel`/`--slowmodel`.
- `std::agency/local` agency module exposing the same operations programmatically.
- General `std::llm registerProviderModule(path)` for runtime provider registration (extracted shared loader, no behavior change for bootstrap).
- Curated catalog of 16 GGUF models spanning tiny -> XL plus coding, reasoning, and embedding specialists. Selected via short name, alias, `hf:` URI, or `.gguf` path.
- Aliases stored under `client.modelAliases` in the nearest `agency.json` (walks up from cwd, falls back to `~/agency.json`); the resolved file is printed on every mutation.

## Architecture

All logic lives in [packages/agency-lang/lib/stdlib/localModels.ts](packages/agency-lang/lib/stdlib/localModels.ts), exposed as:

- Agency functions in [packages/agency-lang/stdlib/agency/local.agency](packages/agency-lang/stdlib/agency/local.agency).
- CLI handlers in [packages/agency-lang/lib/cli/local.ts](packages/agency-lang/lib/cli/local.ts), wired in [packages/agency-lang/scripts/agency.ts](packages/agency-lang/scripts/agency.ts).

Downloads use `node-llama-cpp`'s `resolveModelFile` via a bundled, load-on-demand `lib/stdlib/providers/llama-cpp.mjs`, so `smoltalk-llama-cpp` stays a non-dependency. The Makefile copies the .mjs into `dist/` at build time.

## Why `smoltalk-llama-cpp` is not in `package.json`

Adding it would pull `node-llama-cpp`'s 30-100 MB prebuilt binary on every `pnpm install`, even for users who never run a local model. Instead, the post-merge CI workflow installs it explicitly with `--save=false` at a pinned version; the unit and agency-js suites use the `AGENCY_LLAMA_PROVIDER_MODULE` env override to inject a fake provider.

## Tests

| Suite | Status |
|---|---|
| New unit tests across 7 files | 95 / 95 |
| End-to-end agency-js (`tests/agency-js/local-model/`) | 1 / 1 |
| ESLint structural lint | clean |
| Build (`make`) | clean |
| Integration suite (gated, real download + inference) | 5 tests, skipped without `AGENCY_LLM_INTEGRATION=1`; runs post-merge via `.github/workflows/local-model.yml` |

## Notable refactors

- `findFileUp(startDir, filename, accept?)`: generic walk-up helper extracted from `findPackageRoot` in [packages/agency-lang/lib/importPaths.ts](packages/agency-lang/lib/importPaths.ts). Reused by both `findPackageRoot` and the new alias-config resolver.
- `loadProviderModuleByPath(raw)`: single-module loader extracted from `loadProviderModules` in [packages/agency-lang/lib/runtime/providerModules.ts](packages/agency-lang/lib/runtime/providerModules.ts); shared with the new `std::llm registerProviderModule`.

## Config

- `client.modelAliases` added to `lib/config.ts` schema + zod (read at runtime, not compile-time baked).
- Bumped `vitest.config.ts` `exclude` from blanket `'tests'` to per-subdir patterns so the new `tests/integration/` suite is reachable.

## Security (integration CI)

- Workflow triggers on `push: branches: [main]` only - no PR triggers, no `pull_request_target`, no label triggers.
- `permissions: contents: read`, no secrets.
- Downloaded GGUF is SHA256-verified against a checked-in `EXPECTED_SHA256` (tamper canary for HF account compromise / mirror MITM).
- Third-party actions pinned to SHA matching the existing `test-with-llm.yml` pins.
- `smoltalk-llama-cpp` is installed at a pinned version via `--save=false`, keeping `package.json` + lockfile clean.

## Plan

The full implementation plan (incl. spec coverage, anti-pattern audit, and resolved review issues) lives at [packages/agency-lang/docs/superpowers/plans/2026-06-25-local-model-flag.md](packages/agency-lang/docs/superpowers/plans/2026-06-25-local-model-flag.md).

## Heads up

`EXPECTED_SHA256` in [packages/agency-lang/tests/integration/local-model/smoltest.test.ts](packages/agency-lang/tests/integration/local-model/smoltest.test.ts) is currently the placeholder `"<fill-in-from-hf-download>"`. The plan calls for capturing it by running the suite once with a relaxed assertion, then pasting the printed hash. That step has not been performed (no real download was triggered locally), so the first post-merge run will fail loudly until the hash is filled in.

## Test plan

- `make` - clean.
- `pnpm test:run lib/stdlib/localModels.test.ts lib/stdlib/providers/llamaModelConfig.test.ts lib/stdlib/llm.registerProviderModule.test.ts lib/cli/local.test.ts lib/runtime/providerModules.test.ts lib/config.modelAliases.test.ts lib/importPaths.test.ts` - 95 / 95 green.
- `pnpm run agency test js tests/agency-js/local-model` - passes.
- `pnpm run agency local alias list` - prints all 16 curated models.
- `pnpm run agency agent --help` - shows `--local-model` flag.
- `pnpm test:run tests/integration/local-model` (gated) - 5 tests skipped without `AGENCY_LLM_INTEGRATION=1`.

### DIFF
--- a/packages/agency-lang/docs/site/cli/local.md
+++ b/packages/agency-lang/docs/site/cli/local.md
@@ -31,33 +31,34 @@ agency agent --local-model qwen3.5-2b     # download (if needed) + run locally
 | command | purpose |
 |---|---|
 | `agency local list` | List downloaded `.gguf` files with sizes and a total. |
-| `agency local download <value>` | Download a model if not already cached; prints the resolved local path. `<value>` may be a curated short name, an alias, an `hf:` URI, or an existing `.gguf` path. |
+| `agency local download <value>` | Download a model if not already cached; prints the source it resolved to and the local path. `<value>` may be a curated short name, an alias, an `hf:` URI, or an existing `.gguf` path. |
 | `agency local remove <name>` | Delete a downloaded `.gguf` from the cache. |
 | `agency local resolve <value>` | Show what a name/alias maps to, without downloading. |
-| `agency local alias list` | List usable short names: curated built-ins and your aliases. |
+| `agency local alias list` | List usable short names. Curated entries show params, category, size, context window, and license (with the description on the next line); your aliases show their target. |
 | `agency local alias add <name> <uri>` | Add a short-name alias. Prints the `agency.json` path that was edited. |
 | `agency local alias remove <name>` | Remove a short-name alias. Prints the `agency.json` path that was inspected (the file is left untouched if the alias wasn't present). |
 
 ### Where things live
 
-- **Cache dir**: `~/.agency-agent/models` by default; override with `AGENCY_MODELS_DIR`.
+- **Cache dir**: `AGENCY_MODELS_DIR` env var, else `client.modelsDir` in the nearest `agency.json`, else `~/.agency-agent/models`. The default is shared with `agency agent --local-model`, so a `local download` pre-populates what the agent reuses.
 - **Aliases**: written to the nearest `agency.json` walking up from the current directory; if none is found, `~/agency.json` is used. The CLI prints which file it edited on every add/remove.
+- **Curated catalog**: permissive licenses only (apache-2.0 / mit); restrictively-licensed weights (gemma, llama) are intentionally excluded.
 
 ### Config
 
-Aliases are stored under `client.modelAliases`:
+Aliases and the models cache dir live under `client` and are read at runtime,
+so edits take effect on the next call:
 
 ```jsonc
 {
   "client": {
     "modelAliases": {
       "my7b": "hf:Qwen/Qwen2.5-7B-Instruct-GGUF:Q4_K_M"
-    }
+    },
+    "modelsDir": "/data/agency-models"
   }
 }
 ```
-
-Read at runtime, so `agency local alias add/remove` edits take effect on the next call.
 
 ### See also
 
--- a/packages/agency-lang/docs/site/cli/local.md
+++ b/packages/agency-lang/docs/site/cli/local.md
@@ -1,0 +1,65 @@
+---
+title: local
+description: Documents the `agency local` command, which downloads, lists, aliases, and removes local GGUF models used by the `llama-cpp` provider.
+---
+
+# local
+
+Use this to manage and run local models. Backed by `smoltalk-llama-cpp` + `node-llama-cpp`; install once with `npm i -g smoltalk-llama-cpp` before any subcommand that downloads/inspects models.
+
+```bash
+agency local download qwen3.5-2b          # curated name, alias, hf: URI, or .gguf path
+agency local list                         # downloaded models + sizes
+agency local remove qwen3.5-2b            # delete a downloaded model
+agency local resolve my7b                 # show what a name/alias maps to
+
+agency local alias add my7b hf:Qwen/Qwen2.5-7B-Instruct-GGUF:Q4_K_M
+agency local alias list                   # curated + your aliases
+agency local alias remove my7b
+```
+
+The agent has a shortcut for the common case:
+
+```bash
+agency agent --local-model qwen3.5-2b     # download (if needed) + run locally
+```
+
+`--local-model` runs the local model as both the fast and slow model, so the deep subagents stay local too; it ignores `--model`/`--provider`/`--fastmodel`/`--slowmodel`.
+
+### Subcommands
+
+| command | purpose |
+|---|---|
+| `agency local list` | List downloaded `.gguf` files with sizes and a total. |
+| `agency local download <value>` | Download a model if not already cached; prints the resolved local path. `<value>` may be a curated short name, an alias, an `hf:` URI, or an existing `.gguf` path. |
+| `agency local remove <name>` | Delete a downloaded `.gguf` from the cache. |
+| `agency local resolve <value>` | Show what a name/alias maps to, without downloading. |
+| `agency local alias list` | List usable short names: curated built-ins and your aliases. |
+| `agency local alias add <name> <uri>` | Add a short-name alias. Prints the `agency.json` path that was edited. |
+| `agency local alias remove <name>` | Remove a short-name alias. Prints the `agency.json` path that was inspected (the file is left untouched if the alias wasn't present). |
+
+### Where things live
+
+- **Cache dir**: `~/.agency-agent/models` by default; override with `AGENCY_MODELS_DIR`.
+- **Aliases**: written to the nearest `agency.json` walking up from the current directory; if none is found, `~/agency.json` is used. The CLI prints which file it edited on every add/remove.
+
+### Config
+
+Aliases are stored under `client.modelAliases`:
+
+```jsonc
+{
+  "client": {
+    "modelAliases": {
+      "my7b": "hf:Qwen/Qwen2.5-7B-Instruct-GGUF:Q4_K_M"
+    }
+  }
+}
+```
+
+Read at runtime, so `agency local alias add/remove` edits take effect on the next call.
+
+### See also
+
+- [`agency agent --local-model`](./agent) — the easy button that composes the local-model primitives.
+- [Custom providers guide](../guide/custom-providers) — for using any non-llama provider.
--- a/packages/agency-lang/lib/cli/local.ts
+++ b/packages/agency-lang/lib/cli/local.ts
@@ -1,0 +1,92 @@
+import {
+  _localModelsSupported,
+  _resolveModelName,
+  _downloadModel,
+  _listDownloadedModels,
+  _listModelNames,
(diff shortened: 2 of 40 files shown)


egonSchiele:
Addressed in 1bc103a6:

1. **`EXPECTED_SHA256` placeholder** — Changed to `null` with a format-only fallback (asserts 64-char hex regex) that loudly logs the captured hash. The first integration run will print the actual SHA so it can be pasted in to enable strict tamper-canary matching. Workflow no longer red-lights on first run.

2. **`beforeAll`/`afterAll` running unconditionally** — Both hooks now early-return when `enabled === false`. Added an `AGENCY_INTEGRATION_USE_REAL_HOME=1` opt-out so CI can hit the real `~/.agency-agent/models` that `actions/cache` targets; the workflow now sets this env var.

3. **`AGENCY_LLAMA_PROVIDER_MODULE` relative path inconsistency** — `bundledLlamaModule()` now normalizes the override via `path.resolve(process.cwd(), …)` so both `_registerLocalProvider` and `_downloadModel` see an absolute path. Empty-string is treated as unset.

4. **`gate()` ignored the override** — Added an exported `hasLocalModelSupport()` that mirrors `requireSupport()`'s logic (skip the install check when the env override is set). The CLI uses it.

5. **Misleading "Removed alias" message** — `_unaliasModel` now returns `{ file, removed }`; CLI prints either `Removed alias "X" from <file>` or `Alias "X" not present in <file>; nothing changed`. Added a unit test for the new shape.

6. **`actions/cache@v4` and `actions/upload-artifact@v4` unpinned** — Pinned to:
   - `actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830` # v4.3.0
   - `actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02` # v4.6.2

Tests: 23 pass + 5 integration tests properly skipped; ESLint clean.

packages/agency-lang/docs/site/cli/local.md
```diff
@@ -0,0 +1,65 @@
...
+description: Documents the `agency local` command, which downloads, lists, aliases, and removes local GGUF models used by the `llama-cpp` provider.
+---
+
+# local
+
+Use this to manage and run local models. Backed by `smoltalk-llama-cpp` + `node-llama-cpp`; install once with `npm i -g smoltalk-llama-cpp` before any subcommand that downloads/inspects models.
+
+```bash
+agency local download qwen3.5-2b          # curated name, alias, hf: URI, or .gguf path
+agency local list                         # downloaded models + sizes
```
egonSchiele, line 12: should also print path to model

packages/agency-lang/docs/site/cli/local.md
```diff
@@ -0,0 +1,65 @@
...
+agency local remove qwen3.5-2b            # delete a downloaded model
+agency local resolve my7b                 # show what a name/alias maps to
+
+agency local alias add my7b hf:Qwen/Qwen2.5-7B-Instruct-GGUF:Q4_K_M
```
egonSchiele, line 16: would this add info like size too?

packages/agency-lang/docs/site/cli/local.md
egonSchiele: maybe the url it was downloaded from too?

packages/agency-lang/docs/site/cli/local.md
```diff
@@ -0,0 +1,65 @@
...
+| `agency local download <value>` | Download a model if not already cached; prints the resolved local path. `<value>` may be a curated short name, an alias, an `hf:` URI, or an existing `.gguf` path. |
+| `agency local remove <name>` | Delete a downloaded `.gguf` from the cache. |
+| `agency local resolve <value>` | Show what a name/alias maps to, without downloading. |
+| `agency local alias list` | List usable short names: curated built-ins and your aliases. |
+| `agency local alias add <name> <uri>` | Add a short-name alias. Prints the `agency.json` path that was edited. |
+| `agency local alias remove <name>` | Remove a short-name alias. Prints the `agency.json` path that was inspected (the file is left untouched if the alias wasn't present). |
+
+### Where things live
+
+- **Cache dir**: `~/.agency-agent/models` by default; override with `AGENCY_MODELS_DIR`.
```
egonSchiele: makes sense for agency agent, not sure it does for the `local download` command?

packages/agency-lang/lib/cli/local.ts
```diff
@@ -0,0 +1,100 @@
...
+  console.log(_resolveModelName(value));
+}
+
+export function runAliasList(): void {
+  // Curated entries show name, params, category, size, and description.
+  // User aliases show only what they have (name + target).
+  for (const m of _listModelNames()) {
+    if (m.source === "curated") {
+      const size = m.sizeBytes ? formatGB(m.sizeBytes) : "?";
+      console.log(`${m.name}\t${m.params}\t${m.category}\t${size}\t${m.description}`);
```
egonSchiele: consider using the TS code backing the `std::table` agency module to print this as a nice table. Optional, don't do if its a big change or ugly code

packages/agency-lang/lib/stdlib/localModels.ts
```diff
@@ -0,0 +1,482 @@
...
+import { createRequire } from "node:module";
+import { execFileSync } from "node:child_process";
+import { fileURLToPath, pathToFileURL } from "node:url";
+import { findFileUp } from "../importPaths.js";
+import { loadProviderModuleByPath } from "../runtime/providerModules.js";
+
+/** Categories let the CLI / future tooling group + filter without inspecting
+ *  the description. Kept as a small string union so the set is self-documenting. */
+export type ModelCategory =
+  | "tiny"        // < 1 GB, sub-1B parameters; smoke tests, edge devices
```
egonSchiele: isn't this conflicting size with other attrs (like coding/reasoning)?

packages/agency-lang/lib/stdlib/localModels.ts
```diff
@@ -0,0 +1,482 @@
...
+  /** Approximate Q4_K_M download size in bytes. */
+  sizeBytes: number;
+  /** Bucket for filtering / display. */
+  category: ModelCategory;
+  /** One-line "what is it good for" — shown by `agency local alias list`. */
+  description: string;
+  /** Native context window in tokens. */
+  contextWindow: number;
+  /** License identifier (SPDX-ish): "apache-2.0", "llama3.1", "gemma", "mit", etc. */
+  license: string;
```
egonSchiele, line 35: imported: curated set should only include models with a permissive license. Print license, context window in the `local alias list` command -- which may mean not printing that data as a table, or putting the description on a separate line

packages/agency-lang/lib/stdlib/localModels.ts
```diff
@@ -0,0 +1,482 @@
...
+  "nomic-embed-text": {
+    uri: "hf:nomic-ai/nomic-embed-text-v1.5-GGUF:Q4_K_M",
+    params: "137M", sizeBytes: 89_000_000, category: "embedding",
+    contextWindow: 8192, license: "apache-2.0",
+    description: "Returns 768-dim embeddings; pair with a chat model for RAG.",
+  },
+};
+
+export function defaultCacheDir(): string {
+  return process.env.AGENCY_MODELS_DIR ?? path.join(os.homedir(), ".agency-agent", "models");
```
egonSchiele: should we allow users to set this in agency.json, or as a flag?

egonSchiele: Maybe it would be hard to pass around that value?

packages/agency-lang/stdlib/agency/local.agency
```diff
@@ -0,0 +1,85 @@
...
+  """True if smoltalk-llama-cpp is installed."""
+  return _localModelsSupported()
+}
+
+export def resolveModelName(value: string): string {
+  """Map a curated short name or alias to its Hugging Face URI; pass URIs and .gguf paths through. @param value - name, alias, hf: URI, or .gguf path"""
+  return _resolveModelName(value)
+}
+
+export def downloadModel(value: string, cacheDir: string = ""): string {
```
egonSchiele, line 52: `@param`s need to go on their own lines, matching format needed for PFA https://agency-lang.com/guide/partial-application.html#why-not-just-use-wrapper-functions

PR #346, opened 2026-06-27
Improve local-model agent DX: rich catalog table + optional-value --local-model

egonSchiele:
## Summary

Two DX follow-ups to the local-model agent support.

### 1. Readable model list
`agency local alias list` now renders an aligned table instead of a dense, unaligned parenthetical:

```
NAME                          PARAMS     CATEGORY       SIZE   CTX  LICENSE
smollm2-135m                  135M       general     0.10 GB    8K  apache-2.0
    Smallest practical chat model; used by our integration tests, runs anywhere.

qwen3.5-2b                    2B         general     1.28 GB  128K  apache-2.0
    Most popular modern small general model; runs on CPU comfortably.
```

Columns are computed to fit the data; SIZE/CTX right-align; the description sits on a dimmed line; blank lines separate entries. Header is bold on a TTY, plain when piped.

### 2. Discoverable `--local-model`
`agency agent --local-model` with **no value** now prints that catalog so users can see what to pick, instead of failing with a terse "missing value" parser error. A given value runs locally as before; absent uses the normal model resolution.

## How it works

Rather than a one-off CLI intercept, the bare-flag behavior is modeled in the parsing lib:

- **New `optionalValue` flag type in `std::args`** (string flags only): a bare `--flag` yields `""` instead of a missing-value error. Implemented as an argv pre-pass that rewrites `--flag` to `--flag=` before any other stage, so preScan, Node's tokenizer, and coercion are untouched. `--help` shows the value as optional (`[<string>]`).
- The agent declares `--local-model` as `optionalValue` and handles the tri-state directly (absent / `""` -> catalog + exit / value -> run local), keeping behavior next to the flag declaration.
- The catalog formatter is extracted into `localModels.ts` (`formatModelCatalog` / `_printLocalCatalog`, exposed as `printLocalCatalog()` in `std::agency/local`) so the CLI and the agent render an identical table.

## Testing

- 10 new `optionalValue` cases; full 140-test `std::args` suite green.
- Existing `cli/local` tests green; structural linter clean.
- Manually verified: bare flag lists + exits 0, bare-before-flag lists, valued runs a local model, absent uses normal models, `--help` shows `[<string>]`, piped output has no escape codes.

Note: two pre-existing `resolveSmoltalkLlamaCppFromRoots` test failures are environmental (the package is resolvable in this workspace) and reproduce on `main` independent of this change.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -711,7 +711,7 @@ node main() {
       },
       "local-model": {
         type: "string",
-        optionalValue: true,
+        optional: true,
         description: "Run a local model (downloads if needed): a curated short name, an hf: URI, or a .gguf path. Run without a value to list available models. Sets fast+slow to the local model and ignores --model/--provider/--fastmodel/--slowmodel. Requires: npm i -g smoltalk-llama-cpp"
       }
     }
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -25,6 +25,7 @@ import {
   recommendedAutoApprovePolicy,
  } from "./lib/defaultPolicy.agency"
 import { truncate, formatArgs, formatToolResponse } from "./lib/utils.agency"
+import { printLocalCatalog } from "std::agency/local"
 import { configureModels, configureLocalModel } from "./shared.agency"
 import { codeAgent } from "./subagents/code.agency"
 import { explorerAgent } from "./subagents/explorer.agency"
@@ -710,7 +711,8 @@ node main() {
       },
       "local-model": {
         type: "string",
-        description: "Run a local model (downloads if needed): a curated short name, an hf: URI, or a .gguf path. Sets fast+slow to the local model and ignores --model/--provider/--fastmodel/--slowmodel. Requires: npm i -g smoltalk-llama-cpp"
+        optionalValue: true,
+        description: "Run a local model (downloads if needed): a curated short name, an hf: URI, or a .gguf path. Run without a value to list available models. Sets fast+slow to the local model and ignores --model/--provider/--fastmodel/--slowmodel. Requires: npm i -g smoltalk-llama-cpp"
       }
     }
   },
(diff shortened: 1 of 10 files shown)


packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -710,7 +711,8 @@ node main() {
       },
       "local-model": {
         type: "string",
-        description: "Run a local model (downloads if needed): a curated short name, an hf: URI, or a .gguf path. Sets fast+slow to the local model and ignores --model/--provider/--fastmodel/--slowmodel. Requires: npm i -g smoltalk-llama-cpp"
+        optionalValue: true,
```
egonSchiele: use `optional` instead of `optionalValue`

packages/agency-lang/lib/stdlib/args.ts
```diff
@@ -442,6 +466,45 @@ const ARGV_RULES: ArgvRule[] = [
...
+// at end-of-args or followed by another option throws "argument missing".
+// We rewrite a bare optionalValue flag `--name` to `--name=` (empty value)
+// BEFORE any other stage runs, so the rest of the pipeline — preScan's
+// greedy/duplicate rules, Node's tokenizer, coercion — sees a normal
+// valued token and never has to special-case optionality. A flag is "bare"
+// when it's the last token or the next token is another option (`-`/`--`
+// prefixed); a `--name=...` (incl. empty `--name=`) already carries a value
+// and is left alone. Tokens after a standalone `--` are positionals and are
+// copied verbatim.
+export function expandOptionalFlags(
```
egonSchiele, line 484: does this duplicate flag parsing logic? How are other flags parsed? I'd suggest using existing logic if it makes sense, otherwise writing a parser using tarsec (see lib/parsers/parsers.ts for examples)

packages/agency-lang/lib/agents/agency-agent/agent.agency
egonSchiele: Done — renamed the flag-spec field `optionalValue` → `optional` across `std::args`, the `stdlib/args.agency` type, the agent, and the tests (77c49a6f).

packages/agency-lang/lib/stdlib/args.ts
egonSchiele: Good call — it did duplicate `--name` classification. Reworked `expandOptionalFlags` to reuse the existing `flagNameOf` helper (which already resolves `--name`/`--name=v`/`-n`/`-nv` → long name) instead of re-deriving it, plus a small `tokenHasAttachedValue` check. I went with reusing the existing classifier rather than a tarsec parser, since the lib intentionally delegates tokenization to Node's `parseArgs` and `flagNameOf` is the established classifier — a hand-rolled parser would be a much larger surface. Bonus: reusing `flagNameOf` also fixed the short-alias gap Copilot flagged below. (77c49a6f)

packages/agency-lang/lib/stdlib/args.ts
```diff
@@ -442,6 +466,45 @@ const ARGV_RULES: ArgvRule[] = [
...
+      continue;
+    }
+    const flag = token.startsWith("--") ? schema.flagsByName[token.slice(2)] : undefined;
+    if (flag?.optionalValue === true) {
+      const next = argv[i + 1];
+      const bare = next === undefined || next === "--" || next.startsWith("-");
+      out.push(bare ? `${token}=` : token);
+    } else {
+      out.push(token);
+    }
```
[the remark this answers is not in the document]
egonSchiele: Fixed by the `flagNameOf` refactor above: a bare short alias (`-m` at end or before another option) now rewrites to the long `--name=` form and yields "", same as `--name`. Attached short values (`-mVALUE`) and separate values (`-m val`) pass through unchanged. Added short-form tests. (77c49a6f)

packages/agency-lang/lib/stdlib/localModels.ts
```diff
@@ -471,3 +472,94 @@ export async function _registerLocalModel(value: string, cacheDir: string = ""):
...
+        w.category,
+      )}  ${size.padStart(w.size)}  ${ctx.padStart(w.ctx)}  ${license}`;
+
+    // LICENSE is the last column, so it needs no trailing pad.
+    lines.push(ttyColor.bold(row("NAME", "PARAMS", "CATEGORY", "SIZE", "CTX", "LICENSE")));
+    for (const r of rows) {
+      lines.push(row(r.name, r.params, r.category, r.size, r.ctx, r.license));
+      if (r.description) lines.push(ttyColor.dim(`    ${r.description}`));
+      lines.push("");
+    }
```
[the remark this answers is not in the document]
egonSchiele: Fixed — `formatModelCatalog` now emits a blank line only *between* models (and once before the ALIASES section), so the returned string no longer ends in a newline and `console.log` adds exactly one. Updated the doc comment to match and added a test asserting no trailing newline. (77c49a6f)

PR #347, opened 2026-06-27
Add 'agency local refresh': refreshable local-model catalog

egonSchiele:
## Summary

Adds `agency local refresh [url]`, which fetches a GitHub-hosted JSON catalog of recommended local models and layers it onto the local model list — so new model recommendations ship by editing one file on `main`, no package upgrade needed.

Implements the approved spec (`docs/superpowers/specs/2026-06-27-model-catalog-refresh-design.md`).

## How it works

- **Generalized aliases.** A `client.modelAliases` value is now `string | AliasObject`; the object form carries a `uri` plus optional metadata (`params`/`sizeBytes`/`category`/…) and a `source: "remote"` tag. `CURATED_LOCAL_MODELS` stays as the hardcoded base layer. `_listModelNames` dedupes by name (alias wins); `formatModelCatalog` shows any metadata-bearing entry (built-ins + rich aliases) in the table.
- **`agency local refresh`.** Resolves the catalog URL (CLI arg → `AGENCY_MODEL_CATALOG_URL` → `client.modelCatalogUrl` → built-in default), fetches over HTTPS (timeout + 5 MB cap), validates (`version` gate; strict `uri`, lenient metadata). It then drops the previously-managed (`source:"remote"`) entries and writes the catalog's models back as tagged aliases — **skipping any name you've aliased yourself** (your value wins; the skipped remote value is printed). Fetch/parse/validation failure leaves `agency.json` untouched.
- **Diff reporting.** Each run reports `added / updated / unchanged / removed / skipped`; a no-op re-fetch reports everything `unchanged`.
- **Seed catalog.** `packages/agency-lang/data/model-catalog.json` is seeded from `CURATED_LOCAL_MODELS` and is what the default URL serves.

## Output example

```
Skipped "smollm2-135m": kept your alias (hf:mine/keep:Q4_K_M);
  remote would have set hf:unsloth/SmolLM2-135M-Instruct-GGUF:Q4_K_M
Refreshed 13 models from <url> → <agency.json>
  (12 added, 0 updated, 0 unchanged, 0 removed, 1 skipped)
```

## Testing

- New unit tests for object-form alias resolve/list/dedupe, rich-alias rendering, URL precedence, `parseCatalog` (valid / bad JSON / bad version / non-object models / bad-uri skip), the `_refreshCatalog` engine (added/updated/unchanged/removed, user-collision skip, non-destructive on invalid blob), and `formatRefreshOutput`.
- Full suites green: `localModels`, `cli/local`, `args` (184 pass; the 2 `resolveSmoltalkLlamaCppFromRoots` failures are pre-existing/environmental and reproduce on `main`).
- `tsc --noEmit` clean, structural linter clean.
- End-to-end smoke: the committed seed file flows through `_refreshCatalog` (13 models, user collision preserved, remote entries written with full metadata).

## Notes

- The default URL points at the seed file on `main`, so a **live** `agency local refresh` only works once this merges. Build/tests don't depend on it (they inject a fetcher).
- Implemented TDD per the plan (`docs/superpowers/plans/2026-06-27-model-catalog-refresh.md`), one commit per task.

### DIFF
--- a/packages/agency-lang/data/model-catalog.json
+++ b/packages/agency-lang/data/model-catalog.json
@@ -1,0 +1,122 @@
+{
+  "version": 1,
+  "models": {
+    "smollm2-135m": {
+      "uri": "hf:unsloth/SmolLM2-135M-Instruct-GGUF:Q4_K_M",
+      "params": "135M",
+      "sizeBytes": 105000000,
+      "category": "general",
+      "contextWindow": 8192,
+      "license": "apache-2.0",
+      "description": "Smallest practical chat model; used by our integration tests, runs anywhere."
+    },
+    "qwen3.5-0.8b": {
+      "uri": "hf:unsloth/Qwen3.5-0.8B-GGUF:Q4_K_M",
+      "params": "0.8B",
+      "sizeBytes": 500000000,
+      "category": "general",
+      "contextWindow": 131072,
+      "license": "apache-2.0",
+      "description": "Tiny model from Alibaba's current generation; good edge-device default."
+    },
+    "qwen3.5-2b": {
+      "uri": "hf:unsloth/Qwen3.5-2B-GGUF:Q4_K_M",
+      "params": "2B",
+      "sizeBytes": 1280000000,
+      "category": "general",
+      "contextWindow": 131072,
+      "license": "apache-2.0",
+      "description": "Most popular modern small general model; runs on CPU comfortably."
+    },
+    "qwen3.5-4b": {
+      "uri": "hf:unsloth/Qwen3.5-4B-GGUF:Q4_K_M",
+      "params": "4B",
+      "sizeBytes": 2400000000,
+      "category": "general",
+      "contextWindow": 131072,
+      "license": "apache-2.0",
+      "description": "Strong multilingual small general workhorse from Alibaba."
+    },
+    "qwen3.5-9b": {
+      "uri": "hf:unsloth/Qwen3.5-9B-GGUF:Q4_K_M",
+      "params": "9B",
+      "sizeBytes": 5500000000,
+      "category": "general",
+      "contextWindow": 131072,
+      "license": "apache-2.0",
+      "description": "Modern medium general model with strong tool use; 128K context."
+    },
+    "gpt-oss-20b": {
+      "uri": "hf:unsloth/gpt-oss-20b-GGUF:Q4_K_M",
+      "params": "20B",
+      "sizeBytes": 12000000000,
+      "category": "general",
+      "contextWindow": 131072,
+      "license": "apache-2.0",
+      "description": "OpenAI's open-weights release; balanced general model for ~16 GB machines."
+    },
+    "mistral-small-3.1": {
+      "uri": "hf:unsloth/Mistral-Small-3.1-24B-Instruct-2503-GGUF:Q4_K_M",
+      "params": "24B",
+      "sizeBytes": 14000000000,
+      "category": "general",
+      "contextWindow": 131072,
+      "license": "apache-2.0",
+      "description": "Mistral's general 24B base model (also Devstral's foundation); broad utility."
+    },
+    "qwen3.5-27b": {
+      "uri": "hf:unsloth/Qwen3.5-27B-GGUF:Q4_K_M",
+      "params": "27B",
+      "sizeBytes": 16000000000,
+      "category": "general",
+      "contextWindow": 131072,
+      "license": "apache-2.0",
+      "description": "Modern dense general 27B; the practical ceiling for most workstations."
+    },
+    "deepseek-r1-distill-llama-8b": {
+      "uri": "hf:unsloth/DeepSeek-R1-Distill-Llama-8B-GGUF:Q4_K_M",
+      "params": "8B",
+      "sizeBytes": 4920000000,
+      "category": "reasoning",
+      "contextWindow": 131072,
+      "license": "mit",
+      "description": "Chain-of-thought distill into Llama-8B; best small reasoning model."
+    },
+    "phi-4-reasoning": {
+      "uri": "hf:unsloth/Phi-4-reasoning-GGUF:Q4_K_M",
+      "params": "14B",
+      "sizeBytes": 9050000000,
+      "category": "reasoning",
+      "contextWindow": 32768,
+      "license": "mit",
+      "description": "Microsoft's reasoning-tuned 14B; competitive with much larger models on math/logic."
+    },
+    "devstral-small-2507": {
+      "uri": "hf:mistralai/Devstral-Small-2507_gguf:Q4_K_M",
+      "params": "24B",
+      "sizeBytes": 14300000000,
+      "category": "coding",
+      "contextWindow": 131072,
+      "license": "apache-2.0",
+      "description": "Mistral's official coding-agent GGUF; #1 open-source on SWE-Bench at release."
+    },
+    "qwen3-coder-30b-a3b": {
+      "uri": "hf:unsloth/Qwen3-Coder-30B-A3B-Instruct-GGUF:Q4_K_M",
+      "params": "30B (A3B)",
+      "sizeBytes": 19000000000,
+      "category": "coding",
+      "contextWindow": 262144,
+      "license": "apache-2.0",
+      "description": "Qwen's MoE coder (3.3B active); strong agentic coding + 256K context."
+    },
+    "nomic-embed-text": {
+      "uri": "hf:nomic-ai/nomic-embed-text-v1.5-GGUF:Q4_K_M",
+      "params": "137M",
+      "sizeBytes": 89000000,
+      "category": "embedding",
+      "contextWindow": 8192,
+      "license": "apache-2.0",
+      "description": "Returns 768-dim embeddings; pair with a chat model for RAG."
+    }
+  }
+}
--- a/packages/agency-lang/lib/cli/local.ts
+++ b/packages/agency-lang/lib/cli/local.ts
@@ -9,6 +9,8 @@ import {
   hasLocalModelSupport,
   formatGB,
   formatModelCatalog,
+  _refreshCatalog,
+  type RefreshResult,
 } from "../stdlib/localModels.js";
 
 /** Install-gate for I/O commands. Honors the AGENCY_LLAMA_PROVIDER_MODULE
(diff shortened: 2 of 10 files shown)


packages/agency-lang/data/model-catalog.json
```diff
@@ -0,0 +1,122 @@
...
+      "uri": "hf:unsloth/SmolLM2-135M-Instruct-GGUF:Q4_K_M",
+      "params": "135M",
+      "sizeBytes": 105000000,
+      "category": "general",
+      "contextWindow": 8192,
+      "license": "apache-2.0",
+      "description": "Smallest practical chat model; used by our integration tests, runs anywhere."
+    },
+    "qwen3.5-0.8b": {
+      "uri": "hf:unsloth/Qwen3.5-0.8B-GGUF:Q4_K_M",
```
egonSchiele, line 14: Is this the most specific URL or is there an even more specific one that locks to a specific version? I'm wondering if there is any way to make downloading these models even safer.

packages/agency-lang/lib/cli/local.ts
```diff
@@ -96,3 +98,33 @@ export function runAliasAdd(name: string, uri: string): void {
...
 }
+
+/** Format the lines `runRefresh` prints. Pure so it can be unit-tested without
+ *  a network call. `r.modelCount` is the size of the catalog blob; the
+ *  breakdown line accounts for every entry exactly once (added + updated +
+ *  unchanged + skipped = modelCount). */
+export function formatRefreshOutput(r: RefreshResult): string[] {
+  const lines: string[] = [];
+  for (const s of r.skipped) {
+    lines.push(`Skipped "${s.name}": kept your alias (${s.keptUri});`);
```
egonSchiele: print model names/url/file/other attrs in color.

packages/agency-lang/lib/stdlib/localModels.ts
```diff
@@ -240,24 +263,43 @@ export function _resolveModelName(value: string, file: string = ""): string {
...
+  ModelNameEntry,
+  "params" | "sizeBytes" | "category" | "description" | "contextWindow" | "license"
+>;
+
+/** Project the optional display-metadata fields off any source shape
+ *  (`ModelInfo` or `AliasObject`). Returns only defined fields so the
+ *  spread below doesn't introduce stray `undefined` keys. */
+function metaFrom(src: Partial<EntryMeta>): EntryMeta {
+  const out: EntryMeta = {};
+  if (src.params !== undefined) out.params = src.params;
```
egonSchiele, line 292: What would happen if you didn't have the conditional? Would the behavior change?

packages/agency-lang/lib/stdlib/localModels.ts
```diff
@@ -311,6 +353,294 @@ export function _removeModel(name: string, cacheDir: string = ""): boolean {
...
+function compact<T extends Record<string, unknown>>(o: T): Partial<T> {
+  return Object.fromEntries(Object.entries(o).filter(([, v]) => v !== undefined)) as Partial<T>;
+}
+
+/** Parse + validate the catalog JSON. Throws on blob-level problems (bad JSON,
+ *  unsupported version, `models` not an object) so refresh aborts without
+ *  touching agency.json. Skips an individual entry (with a warning) when its
+ *  `uri` is missing/invalid; metadata fields with the wrong type are dropped
+ *  but the entry is kept. */
+export function parseCatalog(text: string): Record<string, CatalogModel> {
```
egonSchiele, line 446: use a zod schema instead

packages/agency-lang/lib/stdlib/localModels.ts
```diff
@@ -311,6 +353,294 @@ export function _removeModel(name: string, cacheDir: string = ""): boolean {
...
+      }),
+    };
+    return [name, model];
+  });
+  return Object.fromEntries(entries.filter((e): e is [string, CatalogModel] => e !== null));
+}
+
+/** Default fetcher: HTTPS-only GET with a bounded timeout and body cap. */
+export async function fetchCatalog(url: string): Promise<string> {
+  if (!url.startsWith("https://")) {
```
egonSchiele, line 556: I'm trying to think of good ways to test this. I wonder if it would make sense to support the URL being a path to a file, because that would let us write an integration test for the merging logic. It wouldn't test fetching the data from the URL but it would test the majority of the refresh logic.

packages/agency-lang/lib/stdlib/localModels.ts
```diff
@@ -311,6 +353,294 @@ export function _removeModel(name: string, cacheDir: string = ""): boolean {
...
+
+/** Classify one catalog entry against the previous state. Pure: no I/O, no
+ *  mutation, deterministic. Tested via `_refreshCatalog`'s end-to-end tests. */
+function classifyEntry(
+  name: string,
+  model: CatalogModel,
+  userAliases: Record<string, AliasValue>,
+  oldManaged: Record<string, AliasObject>,
+): Classification {
+  if (name in userAliases) {
```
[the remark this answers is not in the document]
egonSchiele: Fixed (a3700c59). `classifyEntry` now uses own-property checks (`Object.prototype.hasOwnProperty.call`) for both the user-collision test and the previous-managed-value lookup, so a catalog model named `toString`/`__proto__`/etc. isn't a phantom collision or phantom prior value. Added a test: a model named `toString` is added, not skipped.

packages/agency-lang/lib/stdlib/localModels.ts
```diff
@@ -311,6 +353,294 @@ export function _removeModel(name: string, cacheDir: string = ""): boolean {
...
+  // `null` pairs, leaving only the valid models.
+  const entries: ([string, CatalogModel] | null)[] = Object.entries(
+    raw.models as Record<string, any>,
+  ).map(([name, entry]) => {
+    if (typeof entry !== "object" || entry === null) {
+      console.warn(`[catalog] skipping "${name}": entry is not an object`);
+      return null;
+    }
+    const uri = entry.uri;
+    if (typeof uri !== "string" || !(isModelUri(uri) || isGgufPath(uri))) {
```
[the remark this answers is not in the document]
egonSchiele: Fixed (a3700c59). Catalog entries are now validated with a stricter `isCatalogUri`: any `scheme://…` URL must be `https://`, so `http://` is rejected — including `http://…/x.gguf` (the `.gguf` suffix no longer sneaks an insecure URL through). `hf:` and local `.gguf` paths still pass. Test added.

packages/agency-lang/lib/stdlib/localModels.ts
```diff
@@ -311,6 +353,294 @@ export function _removeModel(name: string, cacheDir: string = ""): boolean {
...
+    if (text.length > CATALOG_MAX_BYTES) {
+      throw new Error(
+        `catalog too large (${text.length} bytes; cap ${CATALOG_MAX_BYTES} bytes)`,
+      );
+    }
+    return text;
+  } finally {
+    clearTimeout(timer);
+  }
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed (a3700c59). `fetchCatalog` now streams the body via `res.body.getReader()` and enforces the 5 MB cap on accumulated `byteLength` as chunks arrive (cancelling the reader on overflow), mirroring the capped reader in `lib/stdlib/http.ts` — no more buffering the whole response or using UTF-16 `length` as a byte count.

packages/agency-lang/lib/cli/local.ts
egonSchiele: Done (a3700c59). `formatRefreshOutput` now colorizes the model count, url, file, and the add/remove/skip counts via `ttyColor`, which is a no-op when stdout isn't a TTY (so piped output and the unit tests stay plain).

packages/agency-lang/lib/stdlib/localModels.ts
egonSchiele: Good catch — the conditional was redundant. `metaFrom(string)` returns `{}` (a string has none of the metadata fields), so behavior is identical with or without it; the `typeof` guard only existed for the type. I moved the string-guard into `metaFrom` itself (`if (typeof src === "string") return {}`) so the alias projection now calls `metaFrom(value)` unconditionally (a3700c59).

packages/agency-lang/lib/stdlib/localModels.ts
egonSchiele: Done (a3700c59). `parseCatalog` now validates with zod: `CatalogTopSchema` for { version, models } and `CatalogModelSchema` per entry (`uri` refined by `isCatalogUri`; metadata fields `.optional().catch(undefined)` so a wrongly-typed field is dropped, not fatal). Entries are still validated individually so one bad entry is skipped-and-warned rather than failing the whole blob, and I kept explicit top-level error messages (mapped from zod's first issue) so the 'version'/'models'/'valid JSON' messages stay precise. The hand-rolled `pick*` narrowers are gone.

packages/agency-lang/lib/stdlib/localModels.ts
egonSchiele: Done (a3700c59). `fetchCatalog` now treats a non-URL value (or a `file://` URL) as a local file path and reads it from disk (with the same byte cap), so `agency local refresh ./catalog.json` works and the merge logic is integration-testable without network. Added a test that runs the full default fetcher → parse → merge through a temp file, and verified the newly-exposed `refreshCatalog()` stdlib function end-to-end against a local file too.

packages/agency-lang/data/model-catalog.json
egonSchiele: Good question — and a real safety lever, but it needs a decision from you, so I held off on guessing.

Two separable things can be pinned:

1. **The catalog URL** (default points at `main`). I deliberately left this on a moving ref: pinning it to a commit SHA would freeze the catalog forever and defeat `refresh`. The integrity guarantee here is HTTPS + (now) the streamed byte cap. If you want more, a signed/checksummed catalog is the real answer — happy to add that as a follow-up.

2. **The model `hf:` URIs** (this line). These resolve to the *latest* revision of each HF repo, so a repo owner force-pushing a bad GGUF would change what users download. node-llama-cpp’s `hf:` URI does support pinning to a revision/commit, which would make downloads reproducible. The catch: it needs (a) confirming the exact `hf:user/repo@<rev>:quant` syntax against the installed node-llama-cpp version, and (b) looking up a specific commit SHA per model (13 of them) — and those go stale as better quants ship.

I did not want to hand-write 13 guessed SHAs into the seed file. My recommendation: keep the URIs unpinned for now (matches how `--local-model` already resolves them), and if you want pinning, I’ll (1) verify the revision syntax end-to-end with a real download, (2) pin the curated set to specific revisions, and (3) note in the catalog format that an entry *may* carry a pinned revision. Want me to take that on? If so, do you want all 13 pinned, or just the ones we recommend by default?

PR #349, opened 2026-06-28
feat(lowering): preserve match structure as Assignment.matchSource

egonSchiele:
## What

Pattern lowering rewrites a pattern-arm `match` into a synthetic scrutinee
binding + an `if`-chain, discarding the original `MatchBlock` before the type
checker ever runs. That makes match exhaustiveness checking impossible
downstream, because the arm structure is gone.

This change carries the arm structure forward by tagging the lowered scrutinee
assignment with `matchSource?: MatchArmMeta[]` — a **slim, deep-cloned,
body-free** snapshot of the arms (`{ caseValue, guard }[]`).

- `lib/types/matchBlock.ts` — new `MatchArmMeta` type (`{ caseValue, guard }`).
- `lib/types.ts` — add `matchSource?: MatchArmMeta[]` to `Assignment`.
- `lib/lowering/patternLowering.ts` — set `matchSource` to a `structuredClone`d
  arm snapshot in `lowerMatchBlock`'s pattern-arm path.
- `lib/lowering/patternLowering.test.ts` — shape test, survives-lifting test,
  nested-`is` regression test.

This is **Part A** — the prerequisite for `match-exhaustiveness-spec.md` (B1/B2)
and `handler-narrowing-spec.md`. It ships and is reviewable on its own.

## Updated after review

- **Slim deep clone, not the live node.** `matchSource` originally held the live
  `MatchBlock`, which (a) retained every un-lowered case body alive through
  codegen and (b) aliased AST that scope resolution mutates in place
  (`matchSource.expression`/`guard` were the same objects as the emitted
  if-chain's). It now stores a `structuredClone`d `{ caseValue, guard }[]`, which
  severs both hazards. Safe to change the shape — nothing consumes it yet.
- **Nested-`is` codegen crash fixed.** The scrutinee value is now lowered via
  `lowerExpression` instead of stored raw. Previously a nested `is` in a compound
  match head (e.g. `match ((r is success) && y) { … }`) survived unlowered into
  codegen, which has no `isExpression` handler → `Unhandled Agency node type:
  isExpression`. This aligns the scrutinee binding with how `lowerAssignment`
  already lowers every other assignment value. Pre-existing bug, on the touched
  line.

## Scope notes

- Only the **pattern-arm** path is tagged. The literal-passthrough match stays a
  `matchBlock` node (read directly) and the `is`-form is guard-based — both
  intentionally untagged. Documented on the field; the future consumer must
  handle all three shapes.

## No behavior change (common path)

The field is additive, optional, and read by **nobody** yet — codegen and every
downstream pass ignore it. The only generated-output change is the nested-`is`
fix, which turns a hard crash into correct output.

## Verification

- `npx vitest run lib/lowering/patternLowering.test.ts` — 34 passed.
- `npx vitest run lib/preprocessors/ lib/lowering/ lib/typeChecker/ lib/backends/`
  — 1140 passed. The single failure is a pre-existing fresh-worktree env issue
  (`agency-lang/stdlib` subpath not resolvable in a worktree checkout); it
  reproduces on the base commit without this change and passes in a clean/CI
  build.
- `npx tsc --noEmit` — clean.
- Manual: `match ((r is success) && y) { … }` compiles instead of crashing.

### DIFF
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -280,9 +280,21 @@ class PatternLowerer {
       type: "assignment",
       variableName: scrutineeName,
       declKind: "const",
-      value: node.expression,
+      // Lower the scrutinee like any other assignment value, so a nested
+      // `is` in a compound head (e.g. `match ((x is A) && y) { … }`) is
+      // rewritten to its boolean condition instead of surviving raw into
+      // codegen, which has no isExpression handler.
+      value: this.lowerExpression(node.expression),
       loc: node.loc,
-      matchSource: node,
+      // Carry a slim, deep-cloned, body-free snapshot of the arms for a later
+      // exhaustiveness pass. Cloned so we neither retain the un-lowered case
+      // bodies nor alias live AST that scope resolution mutates in place. See
+      // Assignment.matchSource.
+      matchSource: structuredClone(
+        node.cases
+          .filter((c): c is MatchBlockCase => c.type === "matchBlockCase")
+          .map((c) => ({ caseValue: c.caseValue, guard: c.guard })),
+      ),
     };
     const scrutineeRef = varRef(scrutineeName, node.loc);
 
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -282,6 +282,7 @@ class PatternLowerer {
       declKind: "const",
       value: node.expression,
       loc: node.loc,
+      matchSource: node,
     };
     const scrutineeRef = varRef(scrutineeName, node.loc);
 
--- a/packages/agency-lang/docs/superpowers/plans/2026-06-27-match-metadata-preservation.md
+++ b/packages/agency-lang/docs/superpowers/plans/2026-06-27-match-metadata-preservation.md
@@ -1,0 +1,198 @@
+# Match Metadata Preservation (Part A) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Stop discarding the original `match` structure during pattern lowering — carry it on the lowered scrutinee assignment as an optional `matchSource` field — so a later type-checker pass can do exhaustiveness checking. No behavior change.
+
+**Architecture:** `lowerMatchBlock` rewrites a pattern-arm `match` into `[scrutineeAssign, ifChain]` and currently drops the `MatchBlock`. Add `matchSource?: MatchBlock` to the `Assignment` type and set it on the synthetic scrutinee assignment. Codegen and every downstream pass ignore the field; it exists only for the type checker.
+
+**Scope note (intentional):** Only the **pattern-arm** path of `lowerMatchBlock` is tagged. The separate `match (x is …)` form has its own lowering path (`lowerMatchIsForm`, `lib/lowering/patternLowering.ts:292`) and is **deliberately not** tagged here — Part A targets pattern-arm exhaustiveness. A future consumer must not assume the is-form scrutinee carries `matchSource`.
+
+**Tech Stack:** TypeScript, vitest. This is the prerequisite for both `match-exhaustiveness-spec.md` (B1/B2) and `handler-narrowing-spec.md` — but it ships and is reviewable on its own.
+
+## Global Constraints
+
+- NEVER use dynamic imports. Use `type` aliases not `interface`. (CLAUDE.md)
+- This is a vitest **unit** change: `npx vitest run <path>`, NOT the agency execution runner.
+- **No behavior change.** Codegen output must be byte-identical; the field is additive and read by nobody yet. Verify the full lowering + typecheck suites stay green.
+- **Verified facts (do not re-derive):**
+  - `Assignment` is at `lib/types.ts:173`; `MatchBlock` is **already imported** at `lib/types.ts:20` (no new import / no new cycle).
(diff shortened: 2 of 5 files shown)


packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -282,6 +282,7 @@ class PatternLowerer {
       declKind: "const",
       value: node.expression,
```
egonSchiele: **[CONFIRMED] Nested `is` in a compound match head is never lowered → codegen crash.**

`value: node.expression` stores the match head raw. The is-form early-return earlier in `lowerMatchBlock` only catches a *top-level* `isExpression`; a nested one — e.g. `match ((x is A) && y) { {a} => … }` — survives unlowered into this scrutinee binding and reaches codegen, which has no `isExpression` handler → compile crash / wrong output.

Fix: `value: this.lowerExpression(node.expression),`.

Note: this is pre-existing (the line predates the `matchSource` change), but it's in the touched function and the fix is trivial. Worth confirming the exact repro parses before acting.

packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -282,6 +282,7 @@ class PatternLowerer {
...
       loc: node.loc,
+      matchSource: node,
```
egonSchiele: **[CONFIRMED + PLAUSIBLE] Store a slim clone, not the live `node`.** Two issues, one fix:

1. **Memory (CONFIRMED):** `matchSource: node` pins the entire un-lowered `MatchBlock` — every case body — alive through the preprocessor and codegen, even though only the arm *patterns* are ever read. The lowered if-chain already contains lowered copies of those bodies, so each pattern-arm match roughly doubles its retained AST during compilation.

2. **Aliasing (PLAUSIBLE):** `matchSource.expression` is the *same object* as `scrutineeAssign.value`, and scope resolution mutates that subtree in place (`walkNodes` descends into `.value` and sets `.scope` on the refs). So the "pristine source" the doc-comment promises is actually mutated post-lowering — a future consumer reads scope-annotated / diverged AST.

Both vanish if you store a **slim cloned** object — just `{ caseValue, guard }[]` per arm plus the scrutinee expression — instead of `node`. Bonus: no un-lowered case bodies, so the "consumers must only read caseValue, never bodies" caveat goes away too.

egonSchiele: **[PLAUSIBLE] Metadata is non-uniform across the three lowering paths.**

Only the pattern-arm path tags `matchSource`; `lowerMatchIsForm` and the literal-passthrough path don't. This is *partly intentional* per the spec (literal matches stay as a `matchBlock` node read directly; the `is`-form is deliberately untagged because it's guard-based). But it means the eventual exhaustiveness/narrowing consumer must explicitly recognize all three representations of "this was a match," or it'll silently skip `is`-form and literal matches. Worth a clear note wherever that consumer lands so the three shapes are handled deliberately, not by accident.

egonSchiele (commented):
Automated high-effort review (8 finder angles + an independent verifier per candidate). This PR implements the Match Part A plan (the `matchSource` metadata tag). Four findings survived verification — see the inline comments.

**Key recommendation:** store a **slim, cloned** arm-metadata object instead of the live `node`. That single change fixes both the memory retention (it currently pins every un-lowered case body alive through codegen) **and** the live-AST aliasing/mutation hazard (`matchSource.expression` is the same object scope-resolution mutates in place), and it removes the "un-lowered bodies" caveat entirely.

**Finding #1 (scrutinee not lowered)** is pre-existing in this function — not introduced by the `matchSource` change — but it's in the touched code and the fix is a one-liner, so worth doing here.

**Finding #4 (non-uniform across lowering paths)** is partly intentional per the spec (literal matches stay as `matchBlock` read directly; the `is`-form is deliberately untagged), but the eventual consumer must explicitly handle all three shapes.

Nothing here blocks the core idea — the additive metadata tag is the right approach; these are about *what* to store and a latent pre-existing bug.

packages/agency-lang/lib/lowering/patternLowering.ts
egonSchiele: Confirmed and fixed in fe316676. Reproduced the crash first (`match ((r is success) && y) { … }` → "Unhandled Agency node type: isExpression" at codegen). Fix is `value: this.lowerExpression(node.expression)`, which is exactly how `lowerAssignment` already lowers every other assignment value — the scrutinee binding was the inconsistent one. Added a lowering regression test asserting no `isExpression` survives. Note: a nested `is` with a bare binder (e.g. `x is foo`) now surfaces the existing clear "binder in pure-boolean is" error instead of crashing codegen, which is the right outcome.

packages/agency-lang/lib/lowering/patternLowering.ts
egonSchiele: Fixed in fe316676. `matchSource` now stores a slim, deep-cloned, body-free snapshot — `MatchArmMeta[]` = `{ caseValue, guard }[]` (defined next to `MatchBlockCase` in `types/matchBlock.ts`) via `structuredClone`. Confirmed both issues: the live `node` retained every un-lowered case body through codegen, and `matchSource.expression` was the same object as `scrutineeAssign.value` (and `foldArms` reuses `arm.guard` directly in the if-chain), so scope resolution mutated the "pristine" copy in place. The deep clone severs both, and the "consumers must not read bodies" caveat is gone. Safe to change the shape now since nothing consumes it yet.

egonSchiele: Acknowledged. Kept the asymmetry (it's intentional per the spec) but made it explicit in fe316676: the `Assignment.matchSource` doc comment now states that only the pattern-arm path is tagged, that the literal-passthrough match stays as a `matchBlock` node and the `is`-form is guard-based, and that the eventual exhaustiveness/narrowing consumer must recognize all three shapes deliberately.

PR #354, opened 2026-06-28
feat(typechecker): discriminated-union narrowing (D1 + D2)

egonSchiele:
## What

Narrow a union-typed variable when a branch tests its literal discriminant:
`if (r.kind == "answer") { … }` narrows `r` to the matching member(s) in the
then-branch and the complement in the else-branch. Foundation for the rest of
the narrowing program. Type-checker-only — no lowering/codegen change.

### Task 1 — tagged `Refine` + declarative dispatch (behavior-preserving)
`NarrowCandidate` becomes `{ variableName, refine }` where `refine` is a tagged
union (`resultBranch | discriminant`), applied via a `narrowers` table keyed by
`refine.kind`. Splits "what to narrow to" from the "how" (child scope,
reassignment gate, `declareLocal`). Existing `isSuccess`/`isFailure` Result
narrowing is unchanged — its full e2e suite is the gate.

### Task 2 — discriminant engine
- `analyzeCondition` recognizes `v.prop == literal` / `!= literal` over a bare
  variable (either operand order; string/number/boolean literals).
- `narrowUnionByDiscriminant` filters the union's members. **Sound /
  false-negative-only**: drops only provably-excluded members (a member whose
  discriminant prop isn't a matching-kind literal type can't be proven disjoint,
  so it's kept), and never narrows to `never` or to the full set.
- Composes with `&&` / `||` / `!` / early-return / `while`.
- New shared `literalToType` (`literalType.ts`); `synthType`'s **string** case
  now delegates to it.

### Task 3 — match arms are free
`match (r) { { kind: "answer", data } => … }` lowers to
`const __s = r; if (__s.kind == "answer") { const data = __s.data; … }`, so
Task 2 narrows the temp and the bound field types per arm — no match-specific
code. Witnessed with a same-field/different-type union.

## Notes / decisions from review

- **`synthType` DRY scope (corrected during review):** `synthType` does *not*
  produce numeric/boolean literal types today (`number`→`NUMBER_T`,
  `boolean`→`BOOLEAN_T`); only `string` yields a literal type. So only the
  string case delegates to `literalToType` — routing number/boolean through it
  would change inference (`let x = 42` → literal `42`) and break the
  no-behavior-change guarantee. `literalToType` still returns all three literal
  types; the numeric/boolean ones are consumed only by the recognizer, and the
  union members' discriminant literal types come from the **type-annotation
  parser** (verified), not `synthType`.
- **Agency has no `!(expr)` syntax** (the parser rejects `!` before a
  parenthesized group), so `!`-with-discriminant is covered by an
  `analyzeCondition` unit test rather than e2e.
- **Match-arm bodies** are single statements (no block body), so the witness
  uses `let n: number = data`; the `(assignment to 'x')` suffix is only emitted
  for the first lowered arm, so the test asserts by substring.
- Two documented limitations: only *bound fields* narrow (the source scrutinee
  var isn't re-typed, only the lowered temp); a mixed union with a non-literal
  discriminant member doesn't narrow. The `is`-form match stays untagged.
- The deferred "bind it in the pattern" hint (spec D2) is **not** in this PR;
  rationale retained in the plan.

## Verification

- `lib/typeChecker/` — **717 passed**, `npx tsc --noEmit` clean.
- `lib/backends/` + `lib/preprocessors/` — **414 passed** (the lone failure is a
  pre-existing fresh-worktree env issue: `agency-lang/stdlib` subpath not
  resolvable in a worktree checkout; it reproduces on `main` and passes in CI's
  clean build).
- Result-narrowing e2e suite unchanged (Task 1 behavior-preserving gate).

### DIFF
--- a/packages/agency-lang/lib/typeChecker/narrowing.ts
+++ b/packages/agency-lang/lib/typeChecker/narrowing.ts
@@ -9,12 +9,13 @@ import { Scope } from "./scope.js";
 import { walkNodes } from "../utils/node.js";
 import { safeResolveType } from "./assignability.js";
 import { literalToType } from "./literalType.js";
+import { unescapeStringLiteralValue } from "../parsers/parsers.js";
 
 /**
  * What a candidate narrows to. Tagged so a new narrowing form slots in as one
  * `narrowers`-table entry without touching the apply loop. `resultBranch` is
- * the `isSuccess`/`isFailure` Result narrowing; `discriminant` (wired in the
- * next increment) filters a union by `v.prop == literal`.
+ * the `isSuccess`/`isFailure` Result narrowing; `discriminant` filters a union
+ * by `v.prop == literal`.
  */
 export type Refine =
   | { kind: "resultBranch"; branch: "success" | "failure" }
@@ -146,7 +147,31 @@ function literalTypeMatches(
 ): "yes" | "no" | "unknown" {
   const r = safeResolveType(t, aliases);
   if (r.type !== literal.type) return "unknown"; // non-literal prop, literal union, or kind mismatch
-  return r.value === literal.value ? "yes" : "no";
+  return literalValuesEqual(r, literal) ? "yes" : "no";
+}
+
+/**
+ * Compare two same-kind literal-type values, normalizing for the fact that the
+ * member's type-position value (`a`) is parsed differently from the
+ * expression-position discriminant (`b`):
+ *  - strings: the type parser captures escapes raw (`a\tb`) while the
+ *    expression parser unescapes them (`a⇥b`) — unescape `a` before comparing,
+ *    or an escaped discriminant tag would wrongly drop the matching member.
+ *  - numbers: compare numerically so `1` and `1.0` (both valid literal tags)
+ *    are equal despite different source text.
+ *  - booleans: already canonical (`"true"`/`"false"`).
+ */
+function literalValuesEqual(
+  a: StringLiteralType | NumberLiteralType | BooleanLiteralType,
+  b: StringLiteralType | NumberLiteralType | BooleanLiteralType,
+): boolean {
+  if (a.type === "stringLiteralType") {
+    return unescapeStringLiteralValue(a.value) === b.value;
+  }
+  if (a.type === "numberLiteralType") {
+    return Number(a.value) === Number(b.value);
+  }
+  return a.value === b.value;
 }
 
 /**
@@ -242,15 +267,26 @@ export function applyNarrowing(
   for (const cand of candidates) {
     const current = childScope.lookup(cand.variableName);
     if (!current || current === "any") continue;
-    // Soundness gate: a branch that reassigns the variable may change its type
-    // mid-branch, so don't narrow it (whole-body scan, conservative).
-    if (isReassignedIn(branchBody, cand.variableName)) continue;
     // "what to narrow to" is delegated to the refine's narrower. Each narrower
     // resolves through type-alias variables itself (mirrors synthValueAccess) so
     // alias-typed scrutinees still narrow; null means "leave the type as-is".
-    const narrow = narrowers[cand.refine.kind];
-    const narrowed = narrow(cand.refine as never, current, typeAliases);
-    if (narrowed !== null) childScope.declareLocal(cand.variableName, narrowed);
+    // Run it first: most `v.prop == literal` candidates aren't unions and bail
+    // here, so we skip the (more expensive) whole-body reassignment walk.
+    // The cast widens the table entry — indexing by `refine.kind` can't be
+    // statically correlated with the matching `refine` variant — but the call
+    // is sound: `narrowers[k]` is only ever invoked with the variant whose
+    // `kind` is `k`.
+    const narrow = narrowers[cand.refine.kind] as (
+      refine: Refine,
+      current: VariableType,
+      aliases: Record<string, TypeAliasEntry>,
+    ) => VariableType | null;
+    const narrowed = narrow(cand.refine, current, typeAliases);
+    if (narrowed === null) continue;
+    // Soundness gate: a branch that reassigns the variable may change its type
+    // mid-branch, so don't narrow it (whole-body scan, conservative).
+    if (isReassignedIn(branchBody, cand.variableName)) continue;
+    childScope.declareLocal(cand.variableName, narrowed);
   }
 }
 
--- a/packages/agency-lang/lib/typeChecker/narrowing.ts
+++ b/packages/agency-lang/lib/typeChecker/narrowing.ts
@@ -8,6 +8,7 @@ import type {
 import { Scope } from "./scope.js";
 import { walkNodes } from "../utils/node.js";
 import { safeResolveType } from "./assignability.js";
+import { literalToType } from "./literalType.js";
 
 /**
  * What a candidate narrows to. Tagged so a new narrowing form slots in as one
@@ -42,9 +43,26 @@ const narrowers: {
     const resolved = safeResolveType(current, aliases);
     return resolved.type === "resultType" ? narrowToBranch(resolved, r.branch) : null;
   },
-  // discriminant added in the next increment.
-  discriminant: () => null,
+  discriminant: (r, current, aliases) =>
+    narrowUnionByDiscriminant(current, r.prop, r.literal, r.keep, aliases),
 };
+
+/**
+ * Recognize a single `v.prop` member access over a *bare variable* (exactly one
+ * property hop). Returns null for anything else — nested access (`a.b.c`), a
+ * non-variable base, index/call chains, etc. Variable-keyed only, so the
+ * narrowed scrutinee is statically the same binding at the access site.
+ */
+function asDiscriminantAccess(
+  e: Expression,
+): { variableName: string; prop: string } | null {
+  if (e.type !== "valueAccess") return null;
+  if (e.base.type !== "variableName") return null;
+  if (e.chain.length !== 1) return null;
+  const el = e.chain[0];
+  if (el.kind !== "property") return null;
+  return { variableName: e.base.value, prop: el.name };
+}
 
 /**
  * Inspect a (post-lowering) boolean condition and report the narrowing
@@ -77,6 +95,20 @@ export function analyzeCondition(condition: Expression): ConditionFacts {
       const r = analyzeCondition(condition.right);
       return { then: [], else: [...l.else, ...r.else] };
     }
+    if (condition.operator === "==" || condition.operator === "!=") {
+      // `v.prop == literal` / `!= literal` over a bare variable. Either operand
+      // order is accepted. then-branch keeps the matching member(s) for `==`
+      // (and the complement for `!=`); the else-branch is the inverse.
+      const acc = asDiscriminantAccess(condition.left) ?? asDiscriminantAccess(condition.right);
+      const lit = literalToType(condition.right) ?? literalToType(condition.left);
+      if (!acc || !lit) return NO_FACTS;
+      const keepThen = condition.operator === "==";
+      const mk = (keep: boolean): NarrowCandidate => ({
+        variableName: acc.variableName,
+        refine: { kind: "discriminant", prop: acc.prop, literal: lit, keep },
+      });
+      return { then: [mk(keepThen)], else: [mk(!keepThen)] };
+    }
     return NO_FACTS;
   }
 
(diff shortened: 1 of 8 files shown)


packages/agency-lang/lib/typeChecker/narrowing.ts
```diff
@@ -63,6 +132,51 @@ export function narrowToBranch(rt: ResultType, branch: "success" | "failure"): R
...
+ * soundness.
+ */
+function literalTypeMatches(
+  t: VariableType,
+  literal: StringLiteralType | NumberLiteralType | BooleanLiteralType,
+  aliases: Record<string, TypeAliasEntry>,
+): "yes" | "no" | "unknown" {
+  const r = safeResolveType(t, aliases);
+  if (r.type !== literal.type) return "unknown"; // non-literal prop, literal union, or kind mismatch
+  return r.value === literal.value ? "yes" : "no";
```
egonSchiele: **[CONFIRMED — soundness] Literal comparison diverges between expression and type positions.**

`literalTypeMatches` compares `r.value === literal.value`, but those values are parsed differently: the **expression** parser *unescapes* string literals (`\t`→tab, `\n`→newline — parsers.ts ~343-363), while `stringLiteralTypeParser` captures **raw** chars (`many1Till(char('"'))`, no escape handling). So for a tag with an escape — `type A = { kind: "a\tb", … } | …` — the member's `r.value` (raw `a\tb`) ≠ the expression's `literal.value` (unescaped `a⇥b`); `literalTypeMatches` returns `"no"` and `if (v.kind == "a\tb")` wrongly **drops member A**. The then-body is then checked against the wrong members (unsound) while at runtime `v` really is A → the missing-property crash the checker promised to prevent. Same class for non-canonical numerics (`1` vs `1.0` → `"1" !== "1.0"`).

Narrow trigger (escapes / odd numeric literals in discriminants), but a real soundness hole. Fix: normalize both sides to one representation before comparing (unescape the type-position value, or route through the shared literal-equality logic) rather than raw `===`.

packages/agency-lang/lib/typeChecker/narrowing.ts
```diff
@@ -40,6 +95,20 @@ export function analyzeCondition(condition: Expression): ConditionFacts {
       const r = analyzeCondition(condition.right);
       return { then: [], else: [...l.else, ...r.else] };
     }
+    if (condition.operator === "==" || condition.operator === "!=") {
```
egonSchiele, line 99: **[Intended behavior change — migration heads-up.]** Before this PR, `if (v.kind == "click") { v.key }` type-checked because `synthValueAccess` accepts a property present on *any* union member. Now `==` narrows `v` to the `click` member, so `v.key` errors if `key` isn't on `click`. That's the *correct* behavior and the whole point — but it's a change that can turn green builds red on upgrade (only for typechecker-enabled projects, since the checker is opt-in). Worth a changelog note + a fixture sweep, and it dovetails with the result-as-union strict-access rollout.

packages/agency-lang/lib/typeChecker/narrowing.ts
```diff
@@ -128,15 +242,15 @@ export function applyNarrowing(
...
-    // Resolve through type-alias variables (`let r: R = …` where
-    // `type R = Result<…>` is stored as `typeAliasVariable`, not `resultType`).
-    // Mirrors `synthValueAccess`, which also resolves before its Result check —
-    // without this, alias-typed Results silently never narrow, leaving them
-    // unfixable under Increment 3's hard-error flip.
-    const resolved = safeResolveType(current, typeAliases);
-    if (resolved.type !== "resultType") continue;
+    // Soundness gate: a branch that reassigns the variable may change its type
+    // mid-branch, so don't narrow it (whole-body scan, conservative).
     if (isReassignedIn(branchBody, cand.variableName)) continue;
```
egonSchiele: **[CONFIRMED — efficiency] Run the cheap narrower before the body walk.** Every `v.prop ==/!= literal` now emits a candidate, so `applyNarrowing` runs `isReassignedIn` (a full branch-body `walkNodes`) for *all* of them — including the common case where `v` isn't a union and `narrowUnionByDiscriminant` returns `null`. Reorder so the structural narrower runs first; the reassignment gate still runs before `declareLocal`, so soundness is unchanged:

```ts
const narrowed = narrowers[cand.refine.kind](cand.refine as never, current, typeAliases);
if (narrowed === null) continue;
if (isReassignedIn(branchBody, cand.variableName)) continue;
childScope.declareLocal(cand.variableName, narrowed);
```

packages/agency-lang/lib/typeChecker/narrowing.ts
```diff
@@ -63,6 +132,51 @@ export function narrowToBranch(rt: ResultType, branch: "success" | "failure"): R
...
+  const kept = members.filter((m) => {
+    const rm = safeResolveType(m, aliases);
+    const propType =
+      rm.type === "objectType"
+        ? rm.properties.find((p) => p.key === prop)?.value
+        : undefined;
+    const match = propType ? literalTypeMatches(propType, literal, aliases) : "unknown";
+    return keep ? match !== "no" : match !== "yes";
+  });
+  if (kept.length === members.length || kept.length === 0) return null;
```
egonSchiele, line 201: **[Enhancement] Impossible discriminant comparison narrows to nothing, silently.** When every member is excluded (`kept.length === 0`) the helper returns `null` (no narrowing) — sound, but a typo'd tag like `if (r.kind == "answr")` against `{kind:"answer"} | {kind:"error"}` then type-checks the (dead) branch against the full union with no diagnostic, so a genuine programming error passes. Consider a warning on the provably-empty case (the branch can never run). Fine to defer if keeping D1 minimal — but it's a cheap, high-value diagnostic.

packages/agency-lang/lib/typeChecker/narrowing.ts
```diff
@@ -63,6 +132,51 @@ export function narrowToBranch(rt: ResultType, branch: "success" | "failure"): R
...
+  aliases: Record<string, TypeAliasEntry>,
+): VariableType | null {
+  const resolved = safeResolveType(type, aliases);
+  if (resolved.type !== "unionType") return null;
+  const members = resolved.types;
+  const kept = members.filter((m) => {
+    const rm = safeResolveType(m, aliases);
+    const propType =
+      rm.type === "objectType"
+        ? rm.properties.find((p) => p.key === prop)?.value
```
egonSchiele, line 196: **[Cleanup] Duplicated union-member lookup.** This member-iterate → `safeResolveType` → `properties.find(p => p.key === …)` → collapse-to-single-or-union mirrors `synthValueAccess` (synthesizer.ts ~648-665). A future change to union-member shape handling (nested unions, `resultType` members, intersections) would have to land in both spots or silently diverge narrowing from access typing. Worth extracting a shared `collapseUnion(types)` (and maybe `unionMemberProperty(member, key, aliases)`).

packages/agency-lang/lib/typeChecker/narrowing.ts
```diff
@@ -1,14 +1,69 @@
...
 import { Scope } from "./scope.js";
 import { walkNodes } from "../utils/node.js";
 import { safeResolveType } from "./assignability.js";
+import { literalToType } from "./literalType.js";
 
-export type NarrowCandidate = { variableName: string; branch: "success" | "failure" };
+/**
+ * What a candidate narrows to. Tagged so a new narrowing form slots in as one
+ * `narrowers`-table entry without touching the apply loop. `resultBranch` is
+ * the `isSuccess`/`isFailure` Result narrowing; `discriminant` (wired in the
```
egonSchiele: **[Trivial] Stale comment.** `discriminant` is described as "(wired in the next increment)", but it's fully wired in *this* PR (the `==`/`!=` branch in `analyzeCondition` + `narrowers.discriminant`). Drop the parenthetical so a future reader doesn't think it's inert.

egonSchiele (commented):
Automated high-effort review (8 finder angles + an independent verifier per candidate). This is the D1 discriminated-union narrowing. Findings inline, most-severe first:

1. **Soundness (edge):** literal comparison diverges across expression vs type-position parsing (escape/numeric tags) — can drop the member that actually matches.
2. **Intended behavior change:** narrowing makes previously-lenient cross-member union access an error — a migration heads-up.
3. **Efficiency:** the reassignment body-walk runs before the cheap narrowability check on every `==`/`!=`.
4. **Diagnostic gap:** a provably-impossible discriminant comparison narrows to nothing, silently.
5. **Cleanup:** duplicated union-member lookup vs `synthValueAccess`.
6. **Trivial:** stale "wired in the next increment" comment.

Only #1 is a correctness defect (and narrow-trigger); the rest are a heads-up + cleanups. Worth fixing #1 before result-as-union makes union access strict.

packages/agency-lang/lib/typeChecker/narrowing.ts
egonSchiele: Confirmed and fixed in 174cee26. Reproduced the divergence: the type-position tag `{ kind: "a\tb" }` parses raw (`[97,92,116,98]`, 4 chars) while the expression `== "a\tb"` unescapes (`[97,9,98]`, tab) — so `===` returned "no" and dropped the matching member. Fix: extracted `unescapeStringLiteralValue` from the string parser (and refactored `stringTextSegmentParserFor` to share the escape table, so there's one source of truth), and a new `literalValuesEqual` normalizes — unescape the type-side string; numeric `Number()` compare so `1`/`1.0` agree; booleans stay canonical. Two regression tests added (escaped tag + `1.0`); verified both fail on raw `===` and pass after.

packages/agency-lang/lib/typeChecker/narrowing.ts
egonSchiele: Fixed in 174cee26. Reordered `applyNarrowing`: the narrower runs first and `continue`s on `null`, so the common non-union `==`/`!=` candidate skips the whole-body `isReassignedIn` walk. The reassignment gate still runs before `declareLocal`, so soundness is unchanged.

packages/agency-lang/lib/typeChecker/narrowing.ts
egonSchiele: Fixed in 174cee26 — dropped the parenthetical on both the `Refine` doc comment and the `narrowers.discriminant` entry; it's wired in this PR.

packages/agency-lang/lib/typeChecker/narrowing.ts
```diff
@@ -128,15 +242,15 @@ export function applyNarrowing(
...
+    // Soundness gate: a branch that reassigns the variable may change its type
+    // mid-branch, so don't narrow it (whole-body scan, conservative).
     if (isReassignedIn(branchBody, cand.variableName)) continue;
-    childScope.declareLocal(cand.variableName, narrowToBranch(resolved, cand.branch));
+    // "what to narrow to" is delegated to the refine's narrower. Each narrower
+    // resolves through type-alias variables itself (mirrors synthValueAccess) so
+    // alias-typed scrutinees still narrow; null means "leave the type as-is".
+    const narrow = narrowers[cand.refine.kind];
+    const narrowed = narrow(cand.refine as never, current, typeAliases);
+    if (narrowed !== null) childScope.declareLocal(cand.variableName, narrowed);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 174cee26. Replaced `cand.refine as never` with an honest function-type widening cast (`narrowers[k] as (refine: Refine, …) => …`) plus a comment explaining it: indexing by `refine.kind` can't be statically correlated with the matching variant, but the call is sound because `narrowers[k]` is only ever invoked with the variant whose kind is `k`. No longer claims unreachability.

packages/agency-lang/lib/typeChecker/narrowing.ts
egonSchiele: Acknowledged — added an Unreleased CHANGELOG entry in 174cee26 calling out the migration: for typechecker-enabled projects, accessing a property that only exists on a different member inside such a guard is now an error (intended). On the fixture sweep: the full lib/typeChecker + backends/preprocessors/lowering suites stay green, and CI runs the agency fixtures; I'll watch for any that relied on the old lenient access.

packages/agency-lang/lib/typeChecker/narrowing.ts
egonSchiele: Deferred (you flagged it as deferrable for a minimal D1). It needs threading a diagnostic sink through `applyNarrowing` → narrowers (the helper has no `ctx.errors`), which is more than D1 should carry. Captured as a follow-up: warn on the provably-empty (`kept.length === 0`) case — a typo'd tag like `== "answr"`.

packages/agency-lang/lib/typeChecker/narrowing.ts
egonSchiele: Deferred. The overlap with `synthValueAccess` is real, but the two also differ (it filters members; access collects+collapses prop types), so a good shared `collapseUnion`/`unionMemberProperty` extraction is its own change touching the synthesizer — better isolated from this feature PR. Noted as a follow-up so narrowing and access don't silently diverge on future union-shape changes.

packages/agency-lang/lib/typeChecker/narrowing.ts
```diff
@@ -1,14 +1,69 @@
...
 import { safeResolveType } from "./assignability.js";
+import { literalToType } from "./literalType.js";
 
-export type NarrowCandidate = { variableName: string; branch: "success" | "failure" };
+/**
+ * What a candidate narrows to. Tagged so a new narrowing form slots in as one
+ * `narrowers`-table entry without touching the apply loop. `resultBranch` is
+ * the `isSuccess`/`isFailure` Result narrowing; `discriminant` (wired in the
+ * next increment) filters a union by `v.prop == literal`.
+ */
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 174cee26 — updated both the `Refine` doc comment and the `narrowers.discriminant` entry; discriminant narrowing is wired in this PR.

PR #356, opened 2026-06-28
Add Tavily web search to std::search

egonSchiele:
## What

Adds `tavilySearch` to `std::search` — a web search backed by [Tavily](https://tavily.com), which has a **free tier with no credit card**. This gives the stdlib (and, later, the agency-agent) a search option that doesn't require a Brave key.

Motivation: Brave **discontinued its free Search API tier in Feb 2026** and now requires a credit card + metered billing, so the existing Brave-backed `search` is no longer a low-friction default.

## Changes

- **`_tavilySearch`** (`lib/stdlib/search.ts`): POSTs to the Tavily API with a `Bearer` token using the global `fetch` — **no new npm dependency**. Reads `TAVILY_API_KEY` (with an `apiKey` override), and maps Tavily's `content` field onto `description` so the result shape is identical to Brave's `SearchResult[]` — a drop-in alternative.
- **`tavilySearch`** (`stdlib/search.agency`): raises the same `std::search` interrupt effect as `search`, so it inherits the existing approval prompt and the recommended-policy auto-approve.
- **Unit tests**: 10 new tests mirroring the Brave `_search` suite (URL/body, auth header, env fallback, mapping, errors).
- **Integration tests** (see CI note below):
  - `tests/integration/stdlib-sandbox/tavily.agency` — compile/parse smoke that skips without a key (mirrors `search.agency`), wired into the per-PR sandbox runner.
  - `tests/integration/stdlib-sandbox/credential/tavily.agency` — **actually performs a live Tavily search** and asserts on the results, wired into `credential/run.mjs`.

## CI

The live search runs in the existing **`integration-credentials`** job, which only runs on **push to main** (never on PRs, so forks can't read secrets). This PR adds `TAVILY_API_KEY` to that job's `env`.

⚠️ **Action required:** add a `TAVILY_API_KEY` secret to the **`ci-credentials`** environment for the live test to run (a free Tavily key works). Until then the credential step will fail the request; everything else (unit tests, per-PR smoke) runs without it.

## Usage

```
import { tavilySearch } from "std::search"

node main() {
  const results = tavilySearch("agency language", count: 5)
}
```


packages/agency-lang/tests/integration/stdlib-sandbox/credential/tavily.agency
```diff
@@ -0,0 +1,20 @@
+import { tavilySearch } from "std::search"
+
+node main() {
+  // Live Tavily search. Runs only in the integration-credentials CI job,
+  // which provides TAVILY_API_KEY. The `with approve` auto-approves the
+  // std::search interrupt so the request actually goes out.
+  handle {
+    const results = tavilySearch("agency language programming", count: 3)
```
egonSchiele: I'm flattered, but not sure how many results you'll get for this search query

packages/agency-lang/tests/integration/stdlib-sandbox/tavily.agency
```diff
@@ -0,0 +1,23 @@
+import { tavilySearch } from "std::search"
+import { env } from "std::system"
+
+node main() {
```
egonSchiele: why is there another test with much the same test body?

packages/agency-lang/tests/integration/stdlib-sandbox/credential/tavily.agency
egonSchiele: Good catch — swapped the query to a generic "TypeScript programming language" so it reliably returns results. Fixed in ca23b0d7.

packages/agency-lang/tests/integration/stdlib-sandbox/tavily.agency
egonSchiele: Agreed, it was redundant. Removed this sandbox fixture (and its run.mjs wiring) — the credential test already exercises a real `tavilySearch`, and `search.agency` is compiled on every PR by the stdlib build/coverage step, so the smoke copy added nothing. Done in ca23b0d7.

packages/agency-lang/lib/stdlib/search.ts
```diff
@@ -59,3 +60,56 @@ export async function _search(
...
+  const data = await response.json();
+  const results = data?.results ?? [];
+
+  // Map Tavily's `content` (the extracted snippet) onto `description` so the
+  // result shape matches Brave's SearchResult exactly — a drop-in alternative.
+  return results.map((r: Record<string, unknown>) => ({
+    title: (r.title as string) ?? "",
+    url: (r.url as string) ?? "",
+    description: (r.content as string) ?? "",
+  }));
```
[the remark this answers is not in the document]
egonSchiele: Fixed — guarded with `Array.isArray(data?.results) ? data.results : []` and added a unit test for a non-array `results` field. ca23b0d7.

PR #358, opened 2026-06-29
Nullish unification: one nothing-value (null)

egonSchiele:
## Nullish Unification (Project 1)

Makes Agency have **one nothing-value: `null`.** `undefined` stays unwritable as a value and is *absorbed* into `null` wherever it arises from the JS runtime or TypeScript interop. The type-level representation of optionality flips from `T | undefined` to `T | null`, so the type system finally matches the `null` the codegen already emits.

Design rationale (cross-language survey, alternatives, why not a `Maybe` type): `packages/agency-lang/docs/dev/null-and-undefined.md`. Spec: `docs/superpowers/specs/2026-06-28-nullish-unification-design.md`. Plan: `docs/superpowers/plans/2026-06-28-nullish-unification.md`.

### What changed

- **Runtime `__eq` helper** — `a === b || (a == null && b == null)`. `null` and `undefined` compare equal; identical to `===` for non-nullish values.
- **Equality codegen** — `==`, `===`, `!=`, `!==` all lower to `__eq`. No strict escape hatch; `===`/`!==` are stylistic aliases. This is what lets a single `== null` check catch a runtime `undefined` from interop or a missing key.
- **Parser** — `key?: T` desugars to `T | null`; the `undefined` type keyword is accepted (e.g. from imported TS types) but normalized to `null` at parse time; optional params `x?: T` widen to `T | null`.
- **Type checker** — `UNDEFINED_T` retired for `NULL_T`; `isOptionalType`, built-in member return types, and the `optional` helper all key on `null`.
- **Formatter** — renders `T | null` properties with the canonical `key?: T` shorthand (round-trips, and cleans up generated stdlib docs).
- **Schema feature** — `schema(T).parse({})` now coalesces a missing optional key to `null` (recursively), so `schema({ foo?: string }).parse({})` → `{ foo: null }`. The LLM structured-output path is unchanged (keys stay required + nullable per OpenAI strict mode), via a threaded `optionalKeyMode`.

### Out of scope (deferred to Project 2 / the narrowing roadmap)

Strict null checking (`strictMemberAccess` on the null member), null narrowing (retarget `null-truthiness-narrowing-spec` to `null`), equality-operand type-checking, the `null`-literal-synthesizes-to-`null` change, `delete`, and a nominal `Option`/`Maybe` type. See the spec's Non-goals.

### Follow-up

- The untracked `null-truthiness-narrowing-spec.md` working file should be retargeted from `undefined` to `null` before it is implemented (the requirement is recorded in spec §5/§6). It isn't a tracked repo file, so it's not part of this PR.

### Testing

New tests: `__eq` truth table; `tests/agency/nullish-eq` (a `== null`/`=== null` check catches an absent/undefined key); `tests/agency/validation/optionalKey{Coalesce,Nested,Recursive}` (parse coalescing). Updated parser/formatter/tool-schema unit tests.

Verified locally (CI runs the full suite): `lib/parsers` (1480), `lib/typeChecker` (721), `lib/backends` (300), `lib/runtime/eq`, the codegen integration fixtures (30), `tsc --noEmit`, and the structural linter — all green; the named Agency execution tests pass.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -1063,10 +1063,21 @@ export class TypeScriptBuilder {
     }
     const leftNode = this.processNode(node.left);
     const rightNode = this.processNode(node.right);
-    // Agency uses strict equality/inequality: == → ===, != → !==
-    const emitOp = node.operator === "==" ? "===" :
-      node.operator === "!=" ? "!==" : node.operator;
-    return ts.binOp(leftNode, emitOp, rightNode, {
+    // All equality operators use unified nullish equality via the `__eq`
+    // runtime helper (null and undefined compare equal). There is no strict
+    // escape hatch: `===`/`!==` are stylistic aliases that compile identically
+    // to `==`/`!=`. See docs/dev/null-and-undefined.md.
+    if (
+      node.operator === "==" ||
+      node.operator === "===" ||
+      node.operator === "!=" ||
+      node.operator === "!=="
+    ) {
+      const eq = ts.call(ts.id("__eq"), [leftNode, rightNode]);
+      const negated = node.operator === "!=" || node.operator === "!==";
+      return negated ? ts.not(eq) : eq;
+    }
+    return ts.binOp(leftNode, node.operator, rightNode, {
       parenLeft: this.needsParensLeft(node.left, node.operator),
       parenRight: this.needsParensRight(node.right, node.operator),
     });
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -1540,7 +1540,16 @@ export class TypeScriptBuilder {
     let zod = this.zodSchemaFor(typeHint);
     if (param.defaultValue) {
       const defaultStr = expressionToString(param.defaultValue);
-      zod += `.nullable().describe(${JSON.stringify("Default: " + defaultStr)})`;
+      // A widened optional param (`x?: T` → `T | null`) is already nullable;
+      // only add `.nullable()` for params whose type is not already nullable
+      // (e.g. an explicit default `x: T = 5`), so the LLM may omit them.
+      const alreadyNullable =
+        typeHint.type === "unionType" &&
+        typeHint.types.some(
+          (t) => t.type === "primitiveType" && t.value === "null",
+        );
+      if (!alreadyNullable) zod += ".nullable()";
+      zod += `.describe(${JSON.stringify("Default: " + defaultStr)})`;
     }
     return { kind: "scalar", zod };
   }
--- a/packages/agency-lang/lib/parsers/parsers.ts
+++ b/packages/agency-lang/lib/parsers/parsers.ts
@@ -1047,7 +1047,7 @@ export const objectPropertyParser: Parser<ObjectProperty> = memo(
     }
 
     if (value.type === "unionType") {
-      // If it's already a union, just add undefined to the list of types
+      // If it's already a union, just add null to the list of types
       return success(
         {
           key,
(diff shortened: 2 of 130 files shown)


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -1529,7 +1540,16 @@ export class TypeScriptBuilder {
...
+      // A widened optional param (`x?: T` → `T | null`) is already nullable;
+      // only add `.nullable()` for params whose type is not already nullable
+      // (e.g. an explicit default `x: T = 5`), so the LLM may omit them.
+      const alreadyNullable =
+        typeHint.type === "unionType" &&
+        typeHint.types.some(
+          (t) => t.type === "primitiveType" && t.value === "null",
+        );
+      if (!alreadyNullable) zod += ".nullable()";
+      zod += `.describe(${JSON.stringify("Default: " + defaultStr)})`;
```
egonSchiele, line 1552: would this conflict with any custom descriptions users set with the `# foo` trailing description syntax? The `.describe` doesn't seem necessary?

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4164,8 +4171,20 @@ export const functionParameterParser: Parser<FunctionParameter> = memo(
...
+      // An optional param with no explicit default resolves to null when the
+      // arg is omitted. Inject that default AND widen the static type to
+      // `T | null` so the type matches the runtime value (nullish unification).
+      // When an explicit default is present (e.g. `x?: string = "hi"`), the
+      // value is never null, so neither the default nor the widening applies.
       if (__optional && !rest.defaultValue) {
         rest.defaultValue = { type: "null" };
+        if (rest.typeHint) {
+          const nullT = { type: "primitiveType", value: "null" };
+          rest.typeHint =
```
egonSchiele: hard to read nested ternary

egonSchiele:
Pushed `e9fed449` addressing the CI failures and review feedback.

**CI failures (full `pnpm test:run` is now green locally — 6464 passed, 0 failed):**
- `lib/typeChecker.test.ts` — two optional-property tests built the AST directly with the old `T | undefined` shape; retargeted to `T | null` (renamed the local `undef` → `nullish`). My earlier straggler sweep excluded `.test.ts` files, which is how these slipped through.
- `lib/preprocessors/parallelDesugar.test.ts` — the arm-dispatch equality now lowers to `__eq`, so the snapshot regex was updated to match `__eq(__arm_0, \`arm_0\`)`.

**Review findings:**
- **`.meta()` ordering (typeToZodSchema.ts) — real bug, fixed.** Good catch. `.optional().default(null)` is now applied *before* `appendMeta`, so any `.meta(...)` stays the final call and JSON-schema metadata is preserved for tagged optional keys. Added regression tests (meta-last for a tagged optional key; LLM path stays required + no coalescing).
- **`null-and-undefined.md` escape-hatch wording — fixed.** Corrected the D3 section: all four operators (`==`/`===`/`!=`/`!==`) lower to `__eq`; there is no strict escape hatch (it would reintroduce the distinction we remove and is a footgun). I'd updated the spec when dropping the hatch but missed the rationale doc.
- **`elementOrUndef` → `elementOrNull` — renamed** in `synthesizer.ts` and `primitiveMembers.ts` (kind, table entry, and doc).
- **Nested ternary readability — refactored** the optional-param type-widening into an `existingMembers` variable.

**On the `.describe("Default: …")` question (typescriptBuilder.ts):** that `.describe` predates this PR — the line was already `.nullable().describe("Default: …")`; I only dedup'd the redundant `.nullable()` for already-nullable optional params. It surfaces the default value to the LLM. Any interaction with the `# ` trailing-description syntax is pre-existing behavior and orthogonal to nullish unification, so I left it as-is — happy to look at it in a separate change if you'd like.

PR #362, opened 2026-06-29
feat(agency-agent): real web search via hosted tools + Tavily/Brave backends

egonSchiele:
## Summary

Adds real web search to the built-in agency agent. One backend is active at a time (`hosted` / `tavily` / `brave` / `off`), chosen at startup based on env and provider, persisted in `~/.agency-agent/settings.json`, and switchable in-session via the new `/search` slash command.

### Runtime / type system
- `hostedTools` is now a first-class, **typed** field on the `llm()` builtin and the pluggable `LLMClient` `PromptConfig` contract — provider-hosted web search reaches smoltalk explicitly instead of via the metadata side-channel.
- OpenAI is routed through the **Responses API** (`provider: "openai-responses"`) for both fast and slow models so hosted `web_search` works on the agent's default OpenAI models.

### Settings persistence
- New `~/.agency-agent/settings.json` persistence layer (`loadSettings` / `saveSettings`). Tolerant of missing or corrupt files (returns `{}` and prints a warning for malformed JSON; never crashes startup). Stores user preferences only; never secrets.

### Search-backend module (`lib/agents/agency-agent/lib/search.agency`)
- Single `BACKENDS` data table is the source of truth for backend availability rules (`envKey`, `cloudOnly`, `unavailableMessage`, `firstRunNote`). One `isBackendAvailable(b, ctx)` helper drives every callsite (`planSaved`, `planSearchSetup` first-run branches, `availableBackendItems`) — no per-backend `if` chains anywhere.
- **Pure decision** (`planSearchSetup`) split from **imperative shell** (`configureSearch`) so the branchy resolution is unit-testable without TTY or LLM.
- Single `SearchSession` record holds all session state (`isLocal` + `activeBackend`), set once at startup — closed the order-dependent-mutable-state hole that an earlier draft had with two parallel module vars.
- Single `setActiveBackend(backend, persist: boolean)` setter; the persist flag is in the signature rather than encoded by which of two setter names was picked. `warnDisabled` uses `persist: false` so re-adding a missing key restores the saved choice next run.
- `withWebSearch(opts)` is the **one declarative seam** every subagent uses. It dispatches via two lookup tables (`TOOL_BACKENDS`, `HOSTED_BACKENDS`) instead of an `if`/`else-if` chain on backend name. `off` naturally falls through (not in either table).
- Wired into `research`, `explorer`, `oracle`, and `code` subagents.
- Tavily is renamed to `"web_search"` via `tavilySearch.rename("web_search")` so the advertised tool name is stable across backends and distinct from `std::wikipedia`'s `"wikipedia_search"`.

### UX
- New `/search` slash command — picker over currently-available backends, prints the chosen one, persists.
- `/help` updated; `configureSearch(isLocal)` called in both cloud and `--local-model` branches of `main()`.

## Behaviour matrix

| Provider source                       | First-run plan                                              |
| ------------------------------------- | ----------------------------------------------------------- |
| Cloud (any provider)                  | `hosted` — silent, one-line note                            |
| `--local-model`, TTY, both keys       | `askWhich` — `select` Tavily vs Brave                       |
| `--local-model`, TTY, one key only    | `confirm` — Y/n on that backend                             |
| `--local-model`, non-TTY, key present | `useWithNote` — first usable backend (Tavily first by order)|
| `--local-model`, no keys              | `promptMissing` — note how to enable                        |

Saved preference flow: load from settings → if backend known + available, use it; if unknown or unavailable, print the descriptor's own `unavailableMessage` and disable for the session **without persisting** (so the saved choice survives).

## Verification

- **30/30** agency-agent execution tests pass
- **969/969** runtime unit tests pass
- `make` builds cleanly

(Full agency test suite intentionally not run locally; CI will run it on this PR.)

## Files

```
new   lib/agents/agency-agent/lib/search.agency        # backend selection module
new   lib/agents/agency-agent/lib/settings.agency      # settings.json persistence
new   lib/agents/agency-agent/tests/{search,settings,models}.agency + .test.json
new   lib/runtime/llmClient.hostedTools.test.ts
edit  lib/runtime/llmClient.ts                         # hostedTools field
edit  lib/typeChecker/builtins.ts                      # hostedTools in llm() type
edit  lib/agents/agency-agent/agent.agency             # /search command, /help, main()
edit  lib/agents/agency-agent/lib/{config,models}.agency
edit  lib/agents/agency-agent/subagents/{research,explorer,oracle,code}.agency
```

## Follow-up

A parallel design for **embeddings** (analogous backend selection over `hosted` / `openai` / `local` / `off` for memory's Tier-2 semantic recall) is being reviewed separately; it stacks on top of the new `settings.agency` and `BACKENDS`-table pattern introduced here.


packages/agency-lang/lib/agents/agency-agent/lib/settings.agency
```diff
@@ -0,0 +1,81 @@
...
+*/
+export def saveSettings(settings: Settings) {
+  """
+  Write settings.json with the given Settings object, creating
+  ~/.agency-agent/ if needed. Logs a warning on write failure but does
+  not throw.
+
+  @param settings - The Settings object to persist.
+  """
+  mkdir(POLICY_DIR)
```
egonSchiele: append `with approve`, Otherwise, this fails with an unhandled interrupt error.

packages/agency-lang/lib/agents/agency-agent/lib/search.agency
```diff
@@ -0,0 +1,451 @@
...
+
+export type SearchContext = {
+  isLocal: boolean;
+  isTty: boolean;
+  hasTavily: boolean;
+  hasBrave: boolean;
+  savedBackend: string;
+}
+
+// `kind` is one of: "use" | "useWithNote" | "askWhich" | "confirm" |
```
egonSchiele: why not type it as a union then?

packages/agency-lang/lib/agents/agency-agent/lib/search.agency
```diff
@@ -0,0 +1,451 @@
...
+  if (b.envKey == "TAVILY_API_KEY" && !ctx.hasTavily) {
+    return false
+  }
+  if (b.envKey == "BRAVE_API_KEY" && !ctx.hasBrave) {
+    return false
+  }
+  return true
+}
+
+def makePlan(kind: string, backend: string, note: string, reason: string): SearchPlan {
```
egonSchiele: What a fun function! What is its purpose?

packages/agency-lang/lib/agents/agency-agent/lib/search.agency
```diff
@@ -0,0 +1,451 @@
...
+      continue
+    }
+    if (isBackendAvailable(b, ctx)) {
+      result.push(b)
+    }
+  }
+  return result
+}
+
+/*
```
egonSchiele, line 212: Just confirming because I'm seeing you use a million different comment types but no doc comments. Do you intend for any of these comments to be visible in the documentation?

packages/agency-lang/lib/agents/agency-agent/lib/search.agency
```diff
@@ -0,0 +1,451 @@
...
+
+// The runtime source of truth for the session. Initialized to a sentinel
+// (off / not-local); configureSearch overwrites both fields at startup.
+let session: SearchSession = { isLocal: false, activeBackend: "off" }
+
+/*
+  Read the active web-search backend. Subagent calls go through
+  `withWebSearch`, which calls this internally — direct callers are rare.
+*/
+export def getSearchBackend(): string {
```
egonSchiele: Function type says it returns a string, but docstring indicates it returns an enum. Tighten the type.

packages/agency-lang/lib/agents/agency-agent/lib/search.agency
```diff
@@ -0,0 +1,451 @@
...
+// hostedTools list (hosted) OR nothing (off). Two lookup tables capture
+// that data so `withWebSearch` is a single table dispatch — no per-
+// backend if/else-if chain. Adding a new tool-backend means one entry in
+// TOOL_BACKENDS; adding a new hosted capability means one entry in
+// HOSTED_BACKENDS; nothing else changes.
+//
+// Tavily is renamed to "web_search" so the advertised tool name is
+// stable across backends and distinct from std::wikipedia's "search".
+
+static const TOOL_BACKENDS: Record<string, any> = {
```
egonSchiele: This variable name doesn't make sense. If you're trying to contrast it with "hosted backends", I don't think "tool backends" is clearly the opposite of "hosted backends". If you're trying to just use a descriptive name, "tool backends" still isn't good because these aren't just any tools, these are search specific tools.

packages/agency-lang/lib/agents/agency-agent/lib/search.agency
```diff
@@ -0,0 +1,451 @@
...
+    appendTool(opts, tool)
+  }
+  const hosted = HOSTED_BACKENDS[name]
+  if (hosted != null && hosted != undefined) {
+    opts.hostedTools = hosted
+  }
+  return opts
+}
+
+// --- Picker helper (Task 8: /search slash command) ----------------------
```
egonSchiele: "Task 8" – no need to have plan-specific numbers in the code. When users read this in the future, they aren't going to be reading it with the plan open in the next tab. "Task 8" is meaningless to everyone.

packages/agency-lang/lib/agents/agency-agent/tests/models.agency
```diff
@@ -0,0 +1,21 @@
+/*
+ * Tests for planModels' provider-routing rule: OpenAI is always promoted to
+ * the Responses API (both fast and slow) so hosted web_search works on the
+ * agent's default models. Other providers are unchanged.
+ */
+import { planModels } from "../lib/models.agency"
+
+node openaiDetectedRoutesBothToResponses(): string {
+  const plan = planModels("", "", "", "", "openai")
+  return "${plan.fastProvider}|${plan.slowProvider}|${plan.fastModel}"
```
egonSchiele: You can return structured data like an object or an array. You don't have to return a string.

egonSchiele:
Thanks for the careful pass — all 14 comments addressed in `8ba94019`.

### Copilot

| # | File | Fix |
| - | ---- | --- |
| 1 | `agent.agency:31` | Removed duplicate `chooseOption` import from `std::ui` (it's already on line 19). |
| 2 | `builtins.ts:60` | `hostedTools` is now typed as `string[]` via a new `stringArray` constant — the runtime expects capability-name strings, so `any[]` was strictly looser. |
| 3 | `settings.agency:50` | `read(...) with approve` — matches `parsePolicyFile`'s pattern; the user opted in to settings persistence by running the agent. |
| 4 | `settings.agency:59` | Added a defensive `typeof != "object" || Array.isArray(...)` check after `parseJSON` with a printed warning, so a valid-but-non-object JSON value (`null`, `42`, `[1,2]`) is treated as empty settings instead of being returned and crashing the next `settings.search = ...`. |
| 5 | `settings.agency:77` | `with approve` on both `mkdir` and `write`. |
| 6 | `search.agency:304` | Imported `search as braveSearch` and renamed both to `"web_search"` so the advertised tool name is stable across both client-tool backends. |

### @egonSchiele

| # | Line | Response |
| - | ---- | -------- |
| 7 | `settings.agency:75` | Done — `with approve` added (same fix as Copilot #5). |
| 8 | `search.agency:43` | Tightened to a `PlanKind` string-literal union (`"use" \| "useWithNote" \| ...`). Also added a `SearchBackend` union for `getSearchBackend()`'s return type (your #11). |
| 9 | `search.agency:159` | Added a `/* */` doc-comment above `makePlan` explaining it's the single constructor for a `SearchPlan` record (every planner branch is now a one-line `return makePlan(...)`; a future field addition is a one-spot change). |
| 10 | `search.agency:196` | Confirming: yes, the `/* */` block comments above each exported function (and `/** @module ... */` at the top) **are** the doc-comments per the convention in [AGENTS.md](https://github.com/egonSchiele/agency-lang/blob/main/packages/agency-lang/AGENTS.md#documentation). Matches `stdlib/calendar.agency` and `stdlib/search.agency`. The `//` line comments are implementation notes that intentionally don't render in generated docs (and this module is agent-internal anyway — no `agency doc` page is generated for it). Triple-quoted `"""..."""` docstrings inside each exported function body provide the LLM-facing tool description. |
| 11 | `search.agency:246` | Tightened return type to `SearchBackend`. |
| 12 | `search.agency:301` | Renamed `TOOL_BACKENDS` → `CLIENT_SEARCH_TOOLS_BY_BACKEND` and `HOSTED_BACKENDS` → `HOSTED_SEARCH_CAPS_BY_BACKEND`. Mirrors the `EMBED_MODEL_BY_PROVIDER` naming pattern in the runtime; makes clear one table holds search-specific client tools and the other holds hosted capabilities. |
| 13 | `search.agency:352` | Removed the `Task 8` reference; section header is now self-describing (`Picker helper for the /search slash command`). Also caught one more stale `Task 5` in the same module. |
| 14 | `tests/models.agency:10` | Tests now return the full `ModelPlan` record; expected outputs are JSON-encoded objects in the `.test.json`. 3/3 pass. |

### Verification (after the fix)
- `make` clean
- 14/14 search agency tests
- 3/3 settings agency tests
- 3/3 models agency tests (structured-output)
- 4/4 agency-agent toolWiring tests
- 7/7 hostedTools + llmClient TS tests
- 962/962 typeChecker tests (including 121 fixture type-checks that now exercise `hostedTools: string[]`)

PR #363, opened 2026-06-29
feat(typechecker): flow-aware synthType — route variable resolution through typeAt (flow-checker PR 2)

egonSchiele:
## Summary

Makes `synthType` consult the flow graph: when a flow node is attached to a variable reference, its type resolves through `typeAt` instead of `scope.lookup`. This makes narrowing **consistent across every pass** — killing the Phase A / Phase B disagreement that blocked enforced Result safety. **PR 2** of the flow-typed checker; the first behavior-*changing* PR.

## How it works

- One change in `synthType`'s `variableName` case: if `ctx.flowEnv` has a flow node for the reference, return `typeAt({ variable, chain: [] }, flow, ctx.flowEnv)`; otherwise `scope.lookup`. Only narrows a *variable* (where `scope.lookup` found one) — function/node/import-ref resolution is untouched.
- `ctx.flowEnv` is populated by `buildFlowGraphs`, which runs *after* `buildScopes`, so the flow path is naturally active only during `checkScopes` (**Phase B**). Phase A keeps its existing child-scope narrowing (PR 3 retires it). No `walkScopeBody` change here.
- Routing the `variableName` case (not `synthValueAccess`) covers member-access bases *and* bare refs (function args, returns, `&&`/`||` RHS) — which is what actually fixes Phase B.

**Observable change:** spurious Phase B errors disappear. `checkFunctionCallsInScope` / `checkReturnTypesInScope` synth args/returns against the flat scope today, so a narrowed member access used as an argument (`takesString(r.v)` inside `if (r.kind == "a")`) currently mis-reports "not assignable". Now `r` narrows and the type is precise.

## Included fix: narrowed Result member → Result type (discovered in triage)

Routing returns/args through `typeAt` surfaced a real false positive: a `Result` narrowed to its success/failure **member object** (via `resultToObjectUnion`) was no longer assignable to the declared `Result` type. `stdlib/policy.agency`'s `return parsed` — where `parsed` was narrowed to `{ success: true, value: any }` by an `if (isFailure(parsed)) return` guard — newly mis-errored against `Result<Policy, ParsePolicyFailure>`. A narrowed member *is* a valid Result. Fix in `assignability.ts`: when the target is a `Result` and the source is not, expand the target via `resultToObjectUnion` and check structurally (mirrors how narrowing already views Result as a union). This is a small scope addition beyond the literal plan, but required — returning/passing a narrowed Result is extremely common.

## Load-bearing invariant

`flowOf` is a `WeakMap<AgencyNode, FlowNode>` keyed on AST node identity. `synthType` looks up by the node it receives during `checkScopes`, so **no pass between `buildFlowGraphs` and `checkScopes` may clone/rewrap variable-reference nodes** (pattern lowering runs before the checker, so the AST is stable today). A new guard test asserts the parsed `r` node has a flow entry, so a future rewrite fails loudly instead of silently losing narrowing.

## Testing

- New `flowNarrowing.test.ts`: cross-pass narrowing (arg / return / `&&` RHS — each was red before with `'string | number' not assignable to 'string'`), the AST-identity guard, and a reassigned-variable pin (assign nodes carry the final declared type — flow isn't per-position yet; revisit in PR 3).
- New `assignability.test.ts` cases: narrowed Result member ⊆ Result type; unrelated object ⊄ Result.
- `make` clean; `npx tsc --noEmit` clean; `pnpm run lint:structure` clean.
- `pnpm test:run` — **6516 passed, 0 failed**. The only suite change was the `stdlib/policy` fixture, fixed at the root (assignability) rather than by editing the fixture.

## Out of scope (PR 3)

Retire the Phase A child-scope mechanism (`walkWithNarrowing`/`applyNarrowing`/`narrowers`/`postGuardFacts`) — requires making Phase A flow-aware too. Between PR 2 and PR 3 the two narrowing impls run side-by-side and must agree; PR 3 should add a Phase-A-vs-Phase-B agreement test before deleting the child-scope path. The reassigned-variable precision question also stands.

### DIFF
--- a/packages/agency-lang/lib/typeChecker/synthesizer.ts
+++ b/packages/agency-lang/lib/typeChecker/synthesizer.ts
@@ -108,7 +108,16 @@ export function synthType(
         if (ctx.flowEnv) {
           const flow = ctx.flowEnv.flowOf.get(expr);
           if (flow) {
-            return typeAt({ variable: expr.value, chain: [] }, flow, ctx.flowEnv);
+            // Resolve against the CURRENT scope's type aliases
+            // (ctx.getTypeAliases() tracks ctx.withScope), not the global
+            // snapshot captured once in buildFlowGraphs — a scope that overrides
+            // an alias must narrow against its own set. Safe with the shared
+            // memo: each flow node belongs to one scope, so it is only ever
+            // queried under that scope's alias context.
+            return typeAt({ variable: expr.value, chain: [] }, flow, {
+              ...ctx.flowEnv,
+              typeAliases: ctx.getTypeAliases(),
+            });
           }
         }
         return scopeType;
--- a/packages/agency-lang/lib/typeChecker/synthesizer.ts
+++ b/packages/agency-lang/lib/typeChecker/synthesizer.ts
@@ -7,6 +7,7 @@ import type {
 import { formatTypeHint } from "../utils/formatType.js";
 import { BUILTIN_FUNCTION_TYPES, AGENCY_FUNCTION_METHOD_TYPES } from "./builtins.js";
 import { isAssignable, isNever, safeResolveType } from "./assignability.js";
+import { typeAt } from "./flow.js";
 import { literalToType } from "./literalType.js";
 import { resultTypeForValidation } from "./validation.js";
 import { TypeCheckerContext } from "./types.js";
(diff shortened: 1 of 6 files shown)


packages/agency-lang/lib/typeChecker/types.ts
```diff
@@ -29,6 +29,10 @@ export type TypeCheckResult = {
   scopes: ScopeInfo[];
   interruptEffectsByFunction: Record<string, InterruptEffect[]>;
   interruptCallGraph: InterruptCallGraph;
+  /** The flow graph built during the check (PR 1b). Exposed for tests/tooling
+   *  (identity guard now; LSP hover / PR 4 later). Undefined if the check threw
+   *  before buildFlowGraphs. */
+  flowEnv?: import("./flow.js").FlowEnvironment;
```
egonSchiele: no dynamic imports!

egonSchiele: Fixed in 52e7b632. Replaced the inline `import("./flow.js")` with a top-level `import type { FlowEnvironment } from "./flow.js"` (no cycle — flow.ts doesn't import ./types.js). Also fixed the matching inline import on TypeCheckerContext.flowEnv (carried over from 1b), so both are clean now. `lint:structure` passes.

packages/agency-lang/lib/typeChecker/synthesizer.ts
```diff
@@ -95,7 +96,23 @@ export function synthType(
...
+        // WeakMap is keyed on the same node object the synthesizer receives. No
+        // flow node (buildScopes / inference, before flowEnv exists, or a
+        // synthetic node) → the declared scope type, unchanged. Only narrows a
+        // *variable* (scopeType found); the ref-resolution below is untouched.
+        if (ctx.flowEnv) {
+          const flow = ctx.flowEnv.flowOf.get(expr);
+          if (flow) {
+            return typeAt({ variable: expr.value, chain: [] }, flow, ctx.flowEnv);
+          }
+        }
```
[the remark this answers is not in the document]
egonSchiele: Good catch — fixed in 52e7b632. buildFlowGraphs captured ctx.getTypeAliases() once at the top level (global aliases), so a scope overriding an alias would narrow against the wrong set. synthType now passes ctx.getTypeAliases() (scope-aware via ctx.withScope) to typeAt, overriding the snapshot. Memo-safe: each flow node roots in exactly one scope's graph and is only queried while checking that scope, so the shared memo never sees two alias contexts for the same node. Full suite green (top-level-alias fixtures resolve identically; this only changes scope-local-alias resolution).

PR #371, opened 2026-06-30
feat(typechecker): narrow single-hop member-path scrutinees

egonSchiele:
## What

Flow narrowing for **single-hop member-path scrutinees** (M1). Before this PR, only a *bare variable* could be narrowed by a guard. Now a one-hop property path narrows too:

```agency
type Box = { r: Result<number, string> }
def f(b: Box): void {
  if (isSuccess(b.r)) {
    let n: number = b.r.value   // narrows — no strict-access error
  }
}
```

Covers all three guard families on a `base.prop` path:
- **Result** — `isSuccess(b.r)` / `isFailure(b.r)` (then-, else-, and post-guard-return)
- **Discriminant** — `e.payload.kind == "click"` filters the union member
- **Presence** — `c.timeout != null` strips the `null` member

This also makes **object-nested Result patterns** narrow, since `{ result: success(v) }` lowers to `if (isSuccess(env.result)) { const v = env.result.value }` — a member-path access that previously hit a spurious strict-access error.

## How

The flow `Reference.chain` (reserved for exactly this) is now lit up across the pipeline:
- **`typeAt`** resolves a property-path `Reference` via a new diagnostic-free `resolvePath` (objectType + `Record<K,V>`, alias-aware, `any` on any unresolvable hop).
- **Recognizers** (`asPathReference`, `asDiscriminantAccess`, `asPresenceTest`, isSuccess/isFailure) emit single-hop path `Reference`s.
- **`synthValueAccess`** resolves a narrowed one-hop prefix through `typeAt` and continues the structural walk from the next hop, behind a strict `flowHasNarrowFor` gate (only short-circuits when a `narrow` for that exact ref genuinely applies — so un-guarded access still raises strict-member-access).

## Soundness

Reassigning the path **or any prefix of it** drops the narrowing:
- `typeAt`'s `assign` case re-resolves `ref` through `resolvePath` when `isPrefixOf(at.ref, ref)` (e.g. `box = ...` invalidates `box.r`); the `loop` back-edge case does the same when the body widened the base var.
- A **sibling** assignment (`box.q = ...`) is not a prefix, so it correctly leaves `box.r` narrowed.
- `flowHasNarrowFor` applies the same prefix rule, so a re-bound `box.r.value` still hits the diagnostic.

## Latent bug fixed

Untyped member-path bindings surfaced a pre-existing flow-sensitivity bug: strict-member-access was emitted during the **pre-flow** inference passes (return-type inference; scope-building, where an untyped `let v = b.r.value` synthesizes its RHS to declare `v`) — a **false positive on narrowed access**. Strict access is flow-sensitive, so `strictMemberAccessSeverity` now returns `silent` when `ctx.flowEnv` is unset. The flow-aware `checkScopes` pass re-synthesizes every value access and is the single source of the diagnostic — verified it still fires on genuinely un-guarded access (the gate does not over-suppress).

## Scope

Multi-hop (`a.b.c`) and index (`arr[0]`) paths are a follow-on (M2) — `Reference.chain` only encodes one property hop today, and index segments aren't yet representable. Bare-variable narrowing is unchanged.

## Tests

- `memberPathNarrowing.test.ts` (new, 11 e2e) — Result/discriminant/presence narrowing, the **CRITICAL un-narrowed-still-errors** gate, soundness-on-reassign, untyped-binding regression (both directions), bare-variable unchanged.
- `flow.test.ts` — property-path `typeAt`, prefix-invalidation (exact-key/base/sibling/disjoint/multi-hop/loop), `flowHasNarrowFor`, `isPrefixOf`.
- `flowBuilder.test.ts` — `.property` valueAccess gets a `flowOf` entry (gate-degradation guard).
- `narrowing.test.ts` — path-ref recognizer positives/negatives.

Full suite green (6663 passed); `tsc`, `lint:structure` clean.

### DIFF
--- a/packages/agency-lang/docs/dev/typechecker/narrowing/README.md
+++ b/packages/agency-lang/docs/dev/typechecker/narrowing/README.md
@@ -28,14 +28,16 @@ flow-typed checker work and the four narrowing specs at the repo root.
 |---|---|---|
 | `isSuccess(r)` / `isFailure(r)` on a bare variable | ✅ | then- and else-branch; `r.value` / `r.error` type precisely |
 | `r.prop == literal` / `!= literal` on a bare variable | ✅ | discriminated-union member filter (either operand order; string/number/boolean literals) |
+| Same guards on a **single-hop member-path** scrutinee (`b.r`, `e.payload`) | ✅ | M1 — `isSuccess(b.r)`, `b.r.kind == "x"`, `b.r != null` narrow the path; `b.r.value` then types precisely. One hop only (`base.prop`); see the nested-access row for multi-hop |
 | `match` arm **bound fields** | ✅ | free via lowering to a `__s` temp — no match-specific code |
 | `!c`, `a && b`, `a || b` combinators | ✅ | `!` swaps then/else; `&&` unions then-facts; `\|\|` unions else-facts |
 | Post-guard / early-return (`if (isFailure(r)) { return }` ⇒ `r` is Success after) | ✅ | `alwaysExits` counts only `return` (conservative) |
-| Nested-access scrutinee (`a.b.c`, `obj.field`) | ❌ | single-hop **bare variable** only |
+| Single-hop member-path scrutinee (`obj.field`) | ✅ | M1 — see the member-path row above |
+| Multi-hop / index nested scrutinee (`a.b.c`, `arr[0]`) | ❌ | M2 follow-on — `Reference.chain` only encodes one property hop today (`asPathReference`'s one-hop ceiling); index segments aren't yet representable |
 | The scrutinee *variable* in a `match` arm (vs a bound field) | ❌ | only bound fields narrow; the source var isn't re-typed |
 | Mixed union with a non-literal discriminant member | ❌ | by design — the `string` member can't be proven disjoint |
 | Narrowing to `never` (dead-branch detection) | ❌ | suppressed today; **planned** with the flow model + `never` |
-| `null` / truthiness (`if (x != null)`, `if (x == null)`, `if (x)`) | ✅ | strips/keeps the `null` member of a `T \| null` optional; bare-variable scrutinee only. `x != null` / `x == null` are exact and narrow **both** branches. Bare `if (x)` narrows **only the then-branch** to non-null: the runtime uses JS truthiness, so a falsy `x` may be `""`/`0`/`false` (not just `null`), so the else-branch (and the post-`while` region) is left unnarrowed — narrowing it to `null` would be unsound. `if (x)` is accepted as a condition for optionals (an opt-in carve-out from the boolean-only condition rule — see `checkConditionType`). |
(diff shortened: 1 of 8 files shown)


packages/agency-lang/lib/typeChecker/synthesizer.ts
```diff
@@ -769,7 +779,25 @@ export function synthValueAccess(
...
-  for (const element of expr.chain) {
+  // Flow-sensitive path narrowing (M1): when the first hop forms a narrowed
+  // single-hop path `base.prop` (e.g. `b.r` inside `if (isSuccess(b.r))`),
+  // resolve THAT prefix through typeAt and continue the structural walk from the
+  // next hop — so `b.r.value` sees the narrowed `b.r`. (Bare-base narrowing,
+  // `r.value`, already flows through `synthType(expr.base)` above.) Gate on
+  // flowHasNarrowFor so we only short-circuit the prefix when narrowing genuinely
+  // applies; otherwise the structural walk's diagnostics (strict member access)
+  // still run on un-narrowed access.
+  let chainStart = 0;
```
egonSchiele: Would it make sense to write this in a more declarative manner?

packages/agency-lang/docs/dev/typechecker/narrowing/README.md
```diff
@@ -28,14 +28,16 @@ flow-typed checker work and the four narrowing specs at the repo root.
 |---|---|---|
 | `isSuccess(r)` / `isFailure(r)` on a bare variable | ✅ | then- and else-branch; `r.value` / `r.error` type precisely |
 | `r.prop == literal` / `!= literal` on a bare variable | ✅ | discriminated-union member filter (either operand order; string/number/boolean literals) |
+| Same guards on a **single-hop member-path** scrutinee (`b.r`, `e.payload`) | ✅ | M1 — `isSuccess(b.r)`, `b.r.kind == "x"`, `b.r != null` narrow the path; `b.r.value` then types precisely. One hop only (`base.prop`); see the nested-access row for multi-hop |
 | `match` arm **bound fields** | ✅ | free via lowering to a `__s` temp — no match-specific code |
 | `!c`, `a && b`, `a || b` combinators | ✅ | `!` swaps then/else; `&&` unions then-facts; `\|\|` unions else-facts |
 | Post-guard / early-return (`if (isFailure(r)) { return }` ⇒ `r` is Success after) | ✅ | `alwaysExits` counts only `return` (conservative) |
-| Nested-access scrutinee (`a.b.c`, `obj.field`) | ❌ | single-hop **bare variable** only |
+| Single-hop member-path scrutinee (`obj.field`) | ✅ | M1 — see the member-path row above |
```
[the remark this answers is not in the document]
egonSchiele: Removed the duplicate row in 1b211ab0 — the guards row ($`isSuccess(b.r)$`/discriminant/presence on a single-hop member path) already covers it; kept the multi-hop/index row as the M2 follow-on marker.

packages/agency-lang/lib/typeChecker/synthesizer.ts
egonSchiele: Done in 1b211ab0 — factored it into a `narrowedPathPrefix(expr, ctx)` helper with guard-clause early returns that returns the narrowed prefix type or `null`. The access site now reads declaratively: `currentType = prefix ?? synthType(base); chainStart = prefix === null ? 0 : 1`, no interleaved mutation.

PR #375, opened 2026-07-01
feat(typechecker): match exhaustiveness for discriminated unions + boolean (B2)

egonSchiele:
## What

Extends the compile-time **"match is not exhaustive"** diagnostic (B1 shipped Result + closed literal/value unions) to:

- **Discriminated object/tagged unions** — `{ kind: "a" } | { kind: "b" }`, where a common property is typed as a distinct literal in each member. A `{ kind: "a" }` arm covers that member; an uncovered member is reported.
- **Bare `boolean` scrutinees** — `match(b) { true => …; false => … }` (previously hard-coded open).

And **flips the default `silent → warn`** so it actually fires.

```agency
type Ev = { kind: "click", x: number } | { kind: "scroll", d: number }
match (e) {
    { kind: "click" } => handleClick(e)
    // warning: match is not exhaustive: missing `{ kind: "scroll" }`
}
```

## How (4 commits)

1. **`decomposeCases`** — new `findDiscriminant` detects a common literal-typed, distinct-valued tag across all object members and tags each member case with `{ prop, value }`; a bare `boolean` enumerates as `true | false`. `literalCase` refactored onto a shared `literalValue`.
2. **`matchExhaustiveness`** — an `objectPattern` arm that pins the discriminant to a literal covers that member; `missingCases` reports the uncovered ones. `armLiteral` + `objectPatternDiscriminantValue` share a single `asStaticLiteral`.
3. **B2b flip** — default `silent → warn` (a single read-site `??`; the zod schema is `.partial()` with no default, so there's one place, not two).
4. **Docs** — replace the guide's stale *"No exhaustiveness check is performed"* line with the current behavior.

## Soundness — every uncertainty path bails (no false positives)

Non-discriminated union · shared tag value · non-object member · nested-union member → not discriminated → no diagnostic. An un-guarded `objectPattern` arm that doesn't pin the discriminant (`{ x }`, shorthand `{ kind }`, interpolated `{ kind: "x${y}" }`) → could match several members → bail. Guarded arms never cover; `_`/bare-binding is a catch-all. `match(x is …)` is never checked (no `matchSource`).

## The B2b sweep

Measured at `warn` across the full suite + `agency doc stdlib`: **zero** genuine non-exhaustive matches in stdlib or fixtures — no source changes needed. The only tests touched were the three that asserted the *silent default* (now assert `warn`).

## Boolean scrutinees came free

Boolean coverage rides B1's existing `armLiteral` (a `true`/`false` arm is already a literal case), so Task 1's `decomposeCases` enumeration was sufficient — no `matchExhaustiveness.ts` change for it.

## Tests

- `typeCases.test.ts` — `findDiscriminant`/`decomposeCases` units: string/number/boolean tags, 3-member, **candidate-loop-advances** (shared `version:1` then discriminating `kind`), non-object/shared-value/nested-union negatives, type-alias unwrap, boolean enumeration.
- `matchExhaustiveness.test.ts` — discriminated e2e: missing member, multiple-missing (names all), covered/`_`/guarded, ambiguous/shorthand/interpolated bail, rest-binding covers, alias; boolean-scrutinee e2e; the flipped default tests.

Full gate green: `make`, `tsc`, `lint:structure` clean; full vitest **6737 passed**.

## Scope

`error` default, multi-field/nested discriminants, and non-discriminated object unions remain out of scope (documented follow-ons).

### DIFF
--- a/packages/agency-lang/docs/site/guide/pattern-matching.md
+++ b/packages/agency-lang/docs/site/guide/pattern-matching.md
@@ -193,5 +193,31 @@ for ({ name, age } in users) {
 Destructuring relies on the underlying JavaScript runtime: reading a
 property of `null` or `undefined` (e.g. `let { name } = null`) throws a
 `TypeError`, which Agency captures and surfaces as a `failure` Result.
-No exhaustiveness check is performed on match blocks — a match without
-a `_` arm that matches no other arm silently does nothing.
+At runtime, a match without a `_` arm that matches no other arm does
+nothing and returns `null`.
+
+## Exhaustiveness checking
+
+The type checker reports (by default, a **warning**) when a `match` over a
+*closed* type doesn't cover every case and has no `_` arm:
+
+- a **Result** (`success` / `failure`),
+- a **closed literal or value union** (`"a" | "b"`, `1 | 2`),
+- a **discriminated object union** (`{ kind: "a" } | { kind: "b" }` — a common
+  property typed as a distinct literal in each member),
+- a bare **`boolean`** (`true` / `false`).
+
+```agency
+type Ev = { kind: "click", x: number } | { kind: "scroll", d: number }
+match (e) {
+    { kind: "click" } => handleClick(e)
+    // warning: match is not exhaustive: missing `{ kind: "scroll" }`
+}
+```
+
+Adding the missing arm — or a `_` catch-all — clears it. A guarded arm
+(`… if (…) => …`) never counts toward coverage. Open types (`string`,
+`number`, arbitrary object unions) and the `match(x is …)` form are never
+required to be exhaustive. Control the severity with
+`typechecker.matchExhaustiveness` in `agency.json` (`"silent"` / `"warn"` /
+`"error"`; default `"warn"`).
--- a/packages/agency-lang/docs/site/guide/pattern-matching.md
+++ b/packages/agency-lang/docs/site/guide/pattern-matching.md
@@ -193,8 +193,8 @@ for ({ name, age } in users) {
 Destructuring relies on the underlying JavaScript runtime: reading a
 property of `null` or `undefined` (e.g. `let { name } = null`) throws a
 `TypeError`, which Agency captures and surfaces as a `failure` Result.
-At runtime, a match without a `_` arm that matches no other arm does
-nothing and returns `null`.
+At runtime, a match without a `_` arm that matches no other arm is a
+no-op — no branch runs.
 
 ## Exhaustiveness checking
 
--- a/packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
+++ b/packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
@@ -81,10 +81,30 @@ function coveredCases(arms: NormalizedArm[]): TypeCase[] {
   return covered;
 }
 
+type LiteralValue = string | number | boolean;
+
+// Case identity keys. A value is tagged with its `typeof` so a numeric `1` and a
+// string "1" never collide. An arm and a decomposed case are the "same case"
+// iff their keys are equal.
+function literalKey(value: LiteralValue): string {
+  return `literal:${typeof value}:${String(value)}`;
+}
+function discriminatedMemberKey(prop: string, value: LiteralValue): string {
+  return `member:${prop}:${typeof value}:${String(value)}`;
+}
+
 function caseKey(c: TypeCase): string {
-  if (c.kind === "literal") return `literal:${typeof c.value}:${String(c.value)}`;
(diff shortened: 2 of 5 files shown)


packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
```diff
@@ -59,7 +82,9 @@ function coveredCases(arms: NormalizedArm[]): TypeCase[] {
 }
 
 function caseKey(c: TypeCase): string {
-  return c.kind === "literal" ? `literal:${typeof c.value}:${String(c.value)}` : c.kind;
+  if (c.kind === "literal") return `literal:${typeof c.value}:${String(c.value)}`;
```
egonSchiele: this is unreadable code. We're not playing code golf.

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
```diff
@@ -80,13 +105,45 @@ function describeCase(c: TypeCase): string {
...
+ * discriminated, or if any un-guarded, non-catch-all arm is an `objectPattern`
+ * that does NOT pin the discriminant (it could match several members → we can't
+ * prove non-coverage). Guarded arms never cover.
  */
 function missingCases(arms: NormalizedArm[], caseSet: CaseSet): TypeCase[] {
   if (!caseSet.closed) return [];
-  if (caseSet.cases.some((c) => c.kind === "member")) return [];
   if (arms.some(isCatchAll)) return [];
+
+  const memberCases = caseSet.cases.filter((c) => c.kind === "member");
```
egonSchiele, line 179: Can you make this logic more declarative so it's easier to follow?

packages/agency-lang/lib/typeChecker/typeCases.ts
```diff
@@ -5,20 +5,64 @@ import { unescapeStringLiteralValue } from "../parsers/parsers.js";
...
+
+/**
+ * A union is DISCRIMINATED on property `p` iff every member resolves to an
+ * objectType with `p` typed as a string/number/boolean LITERAL, and those
+ * literals are pairwise DISTINCT (a value uniquely selects a member). Returns
+ * the first such property in member-0's declaration order (advancing past
+ * candidates whose values collide), or null. Exhaustiveness must DETECT the
+ * discriminant (narrowing is told it by the condition).
+ */
+function findDiscriminant(
```
egonSchiele, line 44: Can you make this logic more declarative so it's easier to follow? And name your vars better? This is unreadable right now

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
egonSchiele: Fixed in 0f5fd4c0. `caseKey` is now a `switch` over the case kinds with named `literalKey` / `discriminatedMemberKey` helpers instead of inline `${typeof}:${String}` templates (the helpers are also shared with the covered-set so the keys cannot drift).

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
egonSchiele: Fixed in 0f5fd4c0. `missingCases` now reads as three named questions: `sharedDiscriminant(memberCases)` (the common tag, or null → un-checkable), `hasArmWithUnpinnedDiscriminant(arms, prop)` (an arm that could match several members → bail), and `coveredMemberKeys(arms, prop)`. The discriminated path is now four short lines.

packages/agency-lang/lib/typeChecker/typeCases.ts
egonSchiele: Fixed in 0f5fd4c0. Extracted `memberLiteralValues` (each member's literal for a prop, or null) and `allDistinct`; renamed `cand`/`rm`/`pv`/`lv`/`ok` → `prop`/`resolved`/`propType`/`value`/(gone). The main function is now: compute `candidateProps`, then return the first whose `memberLiteralValues` are all distinct.

packages/agency-lang/docs/site/guide/pattern-matching.md
```diff
@@ -193,5 +193,31 @@ for ({ name, age } in users) {
 Destructuring relies on the underlying JavaScript runtime: reading a
 property of `null` or `undefined` (e.g. `let { name } = null`) throws a
 `TypeError`, which Agency captures and surfaces as a `failure` Result.
-No exhaustiveness check is performed on match blocks — a match without
-a `_` arm that matches no other arm silently does nothing.
+At runtime, a match without a `_` arm that matches no other arm does
+nothing and returns `null`.
```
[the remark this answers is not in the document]
egonSchiele: Good catch — fixed in 0f5fd4c0. `matchBlock` is a statement, so codegen emits an if/else that simply runs no branch. Changed the guide to "is a no-op — no branch runs" (dropped the incorrect "returns null").

PR #378, opened 2026-07-01
feat(llm): accept named option arguments on llm(...)

egonSchiele:
## What

Allow named option arguments on `llm(...)`:

```agency
llm("Summarize this", model: "gpt-4o-mini", maxTokens: 200)
llm("Do the math", tools: [add])
```

as an ergonomic alternative to the options-object form. The object form
still works unchanged:

```agency
llm("Summarize this", { model: "gpt-4o-mini", maxTokens: 200 })
```

## Why

Named args are nicer to write and read than a nested object, and they're
already the norm for other Agency functions. Previously `llm("hi", model: "x")`
emitted a warning (`Named arguments can only be used with Agency-defined
functions, not 'llm'`) **and miscompiled** — the first named arg's value became
the entire client config (`clientConfig: \`x\``), silently dropping everything
else.

## How

`llm` stays a compiler primitive; no wrapper, stdlib function, or new
machinery. Two small changes:

1. **Checker** (`builtins.ts`): the `llm` builtin gets an `acceptsNamedArgs`
   allowlist derived from the existing `llmOptions` field set (one source of
   truth, so the two never drift). Named option args are now validated
   per-field — unknown keys and wrong value types are errors, e.g.
   `llm("hi", modle: "x")` → `'llm' does not accept the named argument 'modle'`,
   and `llm("hi", maxTokens: "big")` → type error. This reuses the existing
   `acceptsNamedArgs` mechanism already used by `fork`/`race`.

2. **Codegen** (`processLlmCall`): folds named option args into the
   `clientConfig` object. If a positional options object is also present it's
   spread in first, so named args win on conflict. The no-named-args path is
   byte-identical to before (verified against the golden fixtures).

## Testing

- Checker: `lib/typeChecker/builtinNamedArgs.test.ts` — accepts known options,
  rejects unknown names and wrong types.
- Codegen: `lib/backends/llmNamedArgsCodegen.test.ts` — named args fold into a
  `clientConfig` object; positional object form preserved.
- Execution: `tests/agency/llm-named-args.agency` — a named-arg call runs
  end-to-end (deterministic mock).
- Full `lib/typeChecker/` + `lib/backends/` sweep (1308 tests) green;
  `typecheck` and `lint:structure` clean.

### DIFF
--- a/packages/agency-lang/lib/backends/llmNamedArgsCodegen.test.ts
+++ b/packages/agency-lang/lib/backends/llmNamedArgsCodegen.test.ts
@@ -1,0 +1,32 @@
+import { describe, it, expect } from "vitest";
+import { parseAgency } from "../parser.js";
+import { generateTypeScript } from "./typescriptGenerator.js";
+
+function gen(src: string): string {
+  const r = parseAgency(src, {}, false);
+  if (!r.success) throw new Error("parse failed: " + r.message);
+  return generateTypeScript(r.result, undefined, undefined, "test.agency");
+}
+
+describe("llm named-arg option folding", () => {
+  it("folds named options into the clientConfig object", () => {
+    const out = gen(
+      `node main() {\n const r = llm("hi", model: "gpt-4o-mini", maxTokens: 5)\n print(r)\n}`,
+    );
+    // Named args must fold into a clientConfig OBJECT.
+    expect(out).toMatch(/clientConfig:\s*\{/);
+    expect(out).toContain('"model":');
+    expect(out).toContain('"maxTokens":');
+    // Regression: the first named arg's value must not become the whole
+    // clientConfig (the pre-fix miscompile produced `clientConfig: `gpt-4o-mini``).
+    expect(out).not.toMatch(/clientConfig:\s*`gpt-4o-mini`/);
+  });
+
+  it("preserves the positional options-object form", () => {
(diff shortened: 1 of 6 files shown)


egonSchiele:
Thanks for the review — addressed all three Copilot comments in 82bdb8fb:

1. **Prototype pollution** (`builtins.ts`): `llmNamedOptions` is now a null-prototype object, so `in`-based membership no longer matches inherited names. `llm("hi", __proto__: "x")` and `llm("hi", constructor: "x")` are now rejected (tests added).
2. **Splat double-spread** (`typescriptBuilder.ts`): splat args are handled explicitly and spread once — confirmed `llm(prompt, ...opts, model: "x")` now emits `{ ...opts, "model": "x" }` and compiles (previously emitted invalid `......`).
3. **Mixed-form coverage**: added codegen tests for positional-object-`+`-named (spread first, named wins) and for the splat form.

Full `lib/typeChecker/` + `lib/backends/` sweep is green (1312 tests), typecheck + lint clean.

Re: the failing **build** check — it's a pre-existing OOM on `main` (the full `test:run` exhausts the runner's ~4GB heap; `FATAL ERROR: heap out of memory`, exit 134). `main`'s own Node.js CI fails identically, so it's not introduced by this PR. The `lint` check passes here.

PR #385, opened 2026-07-01
feat: multimodal image/file attachments for llm() and userMessage()

egonSchiele:
## What

Adds multimodal **image/file attachments** to Agency, backed by smoltalk 0.7.1's multimodal user-message content. `llm()`'s first argument and `std::thread.userMessage()` now accept an array mixing text strings with attachments; smoltalk does all the I/O (reading paths, fetching URLs, MIME inference, size caps) at send time — Agency only constructs the plain attachment objects.

```ts
import { image, file } from "std::thread"

node main() {
  const answer = llm([
    "What's in this image, and how does it relate to the report?",
    image("./diagram.png"),               // local path
    image("https://example.com/a.jpg"),   // remote URL (auto-detected)
    file("./report.pdf"),                 // local PDF / file
  ])
  print(answer)
}
```

`image(source, mimeType?, base64?)` / `file(source, filename?, mimeType?, base64?)` classify `source` as a local path, `http(s)` URL, `data:` URI, or raw base64 (`base64: true`, which requires a `mimeType`).

## How

- **`std::thread`** — new `image()` / `file()` builders (TS-backed `_imageAttachment` / `_fileAttachment` that own classification + validation, since Agency has no user-level `throw`). New `Attachment` / `AttachmentSource` types; `userMessage()` widened to the same array form.
- **Typechecker** — `llm()` first param tightened from `any` to `string | (string | Attachment)[]` via a structural mirror of the `.agency` types. A new accept/reject/inference test guards the mirror against drift **and** proves heterogeneous-array-literal inference works.
- **Runtime** — `prompt` now threads `string | UserContentInput`; a single `promptText()` flattener feeds string-only consumers (memory recall). Statelog is made attachment-safe: `redactAttachments` at all four message/prompt sites (`onLLMCallStart`, `promptCompletion`, `onLLMCallEnd`, streaming `debug`) so base64 blobs never reach the log.
- **Codegen** — unchanged (an array literal already flows to `runPrompt.prompt`); locked by a guard test.
- **Docs** — new guide section; regenerated stdlib reference from the builder docstrings.

## Scope / non-goals

- `assistantMessage` / `systemMessage` stay string-only (providers don't take attachments on those roles).
- Source kinds limited to path / url / base64 (no `bytes` / `providerFile`); no `timeoutMs` / `detail` knobs in v1.

## Testing

- New: builder units (12), typechecker accept/reject/inference (6), codegen guard (2), runtime `promptText`/redaction units (4), and an end-to-end agency-js fixture asserting attachment parts land on the thread.
- Regression sweep green (llm named-args codegen, builtin named-args, agencyLlm, deterministic client, existing `stdlib-thread-messages` agency-js test).
- `tsc --noEmit` clean · `lint:structure` clean.
- Full agency suite left to CI per repo convention.

Design spec, review, and implementation plan are included under `docs/superpowers/`.

### DIFF
--- a/packages/agency-lang/docs/site/guide/llm.md
+++ b/packages/agency-lang/docs/site/guide/llm.md
@@ -52,7 +52,8 @@ Both builders take optional arguments:
 - `mimeType` — set an explicit MIME type (overrides inference).
 - `base64: true` — treat `source` as raw base64 data. A `mimeType` is then
   required, e.g. `image(data, mimeType: "image/png", base64: true)`. This is
-  handy with `readImage()`, which returns raw base64.
+  handy when you already hold base64 in memory — for example the base64 string
+  inside the `Result` that `readImage()` returns.
 - `file(source, filename)` — the filename shown to the model; it defaults to the
   source's basename.
 
--- a/packages/agency-lang/docs/site/guide/llm.md
+++ b/packages/agency-lang/docs/site/guide/llm.md
@@ -22,6 +22,49 @@ const response: Response = llm("What is the capital of France?")
 print(response)
 ```
 
+## Attachments (images & files)
+
+`llm()`'s first argument can also be an **array** that mixes text with image and
+file attachments. Bare strings in the array are text; use the `image()` and
+`file()` builders from `std::thread` for attachments:
+
+```ts
+import { image, file } from "std::thread"
+
+node main() {
+  const answer = llm([
+    "What's in this image, and how does it relate to the report?",
+    image("./diagram.png"),               // local path
+    image("https://example.com/a.jpg"),   // remote URL (auto-detected)
+    file("./report.pdf"),                 // local PDF / file
+  ])
+  print(answer)
+}
+```
+
+`image(source)` and `file(source)` accept a local path, an `http(s)` URL, or a
+`data:` URI, and figure out which it is. Smoltalk reads the file, fetches the
+URL, infers the MIME type, and enforces size limits when the call is sent — you
+don't do any file I/O yourself.
+
+Both builders take optional arguments:
+
+- `mimeType` — set an explicit MIME type (overrides inference).
+- `base64: true` — treat `source` as raw base64 data. A `mimeType` is then
+  required, e.g. `image(data, mimeType: "image/png", base64: true)`. This is
+  handy with `readImage()`, which returns raw base64.
+- `file(source, filename)` — the filename shown to the model; it defaults to the
+  source's basename.
+
+The same array form works with `userMessage()` from `std::thread`, so you can
+seed a multimodal user turn into the conversation history:
+
+```ts
+import { userMessage, image } from "std::thread"
+
+userMessage(["Here's the screenshot I mentioned", image("./screenshot.png")])
+```
+
 ### Limitation: structured output inside block bodies
 
 Any `return llm(...)` inside a block body — anywhere in the block,
--- a/packages/agency-lang/docs/site/stdlib/thread.md
+++ b/packages/agency-lang/docs/site/stdlib/thread.md
@@ -17,7 +17,7 @@ export type AttachmentSource =
   | { kind: "base64"; base64: string; mimeType: string }
 ```
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/thread.agency#L25))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/thread.agency#L27))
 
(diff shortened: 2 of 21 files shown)


packages/agency-lang/lib/typeChecker/builtins.ts
```diff
@@ -110,6 +110,77 @@ const llmNamedOptions: Record<string, VariableType> = Object.assign(
...
+      properties: [
+        { key: "kind", value: { type: "stringLiteralType", value: "base64" } },
+        { key: "base64", value: string },
+        { key: "mimeType", value: string },
+      ],
+    },
+  ],
+};
+
+const attachment: VariableType = {
```
egonSchiele: Why are these built-in types, as opposed to types defined in agency?

egonSchiele: I also don't see any support for this anywhere besides in this file. If it's a built-in type, should it have parser level support?

egonSchiele: Reworked this. The types ARE defined in Agency — `Attachment` / `AttachmentSource` live in `stdlib/thread.agency`. The `llm` builtin signature now references the `Attachment` alias by name (a `typeAliasVariable`) and resolves it through the same type-alias table any user type uses, so there's no restated structure in `builtins.ts` anymore. Single source of truth. (82615bb8)

egonSchiele: It isn't a real "built-in type" after the refactor — no parser support needed. `Attachment` is an ordinary stdlib type alias (defined + exported from `std::thread`, parsed like any other type), and `image()`/`file()` return it. The only builtin-specific thing is `llm` itself (a compiler primitive), and its first-param type just references the `Attachment` alias by name via the existing alias resolution. So the type has support in exactly one place — the stdlib — and `builtins.ts` no longer restates it. (82615bb8)

packages/agency-lang/docs/site/guide/llm.md
```diff
@@ -22,6 +22,49 @@ const response: Response = llm("What is the capital of France?")
...
+`data:` URI, and figure out which it is. Smoltalk reads the file, fetches the
+URL, infers the MIME type, and enforces size limits when the call is sent — you
+don't do any file I/O yourself.
+
+Both builders take optional arguments:
+
+- `mimeType` — set an explicit MIME type (overrides inference).
+- `base64: true` — treat `source` as raw base64 data. A `mimeType` is then
+  required, e.g. `image(data, mimeType: "image/png", base64: true)`. This is
+  handy with `readImage()`, which returns raw base64.
```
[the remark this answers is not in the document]
egonSchiele: Reworded. `readImage()` returns a `Result`; the guide now points to "the base64 string inside the `Result` that `readImage()` returns" rather than implying it returns raw base64. (82615bb8)

packages/agency-lang/docs/site/stdlib/thread.md
```diff
@@ -53,27 +77,84 @@ Add a system message to the current thread's message history.
...
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/thread.agency#L19))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/thread.agency#L34))
 
 ### userMessage
 
 ```ts
-userMessage(msg: string)
+userMessage(msg: string | string | Attachment[])
 ```
```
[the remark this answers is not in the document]
egonSchiele: Fixed in the doc type printer (`typeToString.ts`): a union array element is now parenthesized, so the signature renders as `userMessage(msg: string | (string | Attachment)[])`. (82615bb8)

packages/agency-lang/docs/site/stdlib/thread.md
```diff
@@ -6,6 +6,30 @@ name: "thread"
...
 
+### AttachmentSource
+
+```ts
+export type AttachmentSource =
+  | { kind: "path"; path: string; mimeType: string
+  | null }
+  | { kind: "url"; url: string; mimeType: string
+  | null }
+  | { kind: "base64"; base64: string; mimeType: string }
```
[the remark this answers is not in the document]
egonSchiele: The paren-drop is fixed — union array elements are now parenthesized in generated docs. The remaining `mimeType: string | null` form and mid-union line wrapping are pre-existing generator behavior (Agency models optional as `| null` rather than `?`/undefined, and long unions get wrapped), which applies to every stdlib type, so I've kept that out of scope for this PR.

PR #387, opened 2026-07-01
fix(runtime): key parallel tool branches by position, not tool-call id

egonSchiele:
## Summary

Fixes `runBatch: duplicate child key "tool_"` when running on **Google Gemini** with parallel tool calls.

Gemini's function-calling protocol returns **no tool-call id** — it matches responses to calls by function name + position, not by id — so smoltalk's Google client defaults the missing id to `""` (`clients/google.js`, `id: z.string().default("")`). runPrompt's parallel tool-dispatch loop keyed each concurrent branch off `tool_${toolCall.id}`, so when Gemini emitted **two tool calls in one round** (parallel function calling), both keys became `"tool_"` and `runBatch` threw:

```
ERROR Function codeAgent threw an exception (converted to Failure): runBatch: duplicate child key "tool_"
    at runBatch (.../runBatch.js:243)
    at PromptRunner.parallel (.../promptRunner.js:125)
    at runPrompt (.../prompt.js:825)
```

The `No handler found for tool call: codeAgent` line was a secondary symptom after the round was abandoned.

## Fix

Derive the branch key, **every per-tool resume step path, and the per-tool timing key** from the tool call's **position index** in the round (`${index}_${toolCall.id}`) instead of the id alone — exactly how Gemini itself disambiguates parallel calls. This also closes a latent second bug: the id-less step paths were colliding too, which would silently break interrupt/resume idempotency (`completedSteps`) for id-less providers.

The real (possibly empty) `toolCall.id` is left **untouched** in message `tool_call_id` fields, so provider pairing and `threadRepair` are byte-for-byte unchanged.

Not done by mutating the id: at runtime the tool calls are `ToolCall` instances (getter, **no setter**) on a fresh run but plain JSON on resume, so mutation is unreliable and would also touch message content.

## Changes

- `lib/runtime/prompt.ts` — branch key / step paths / timing key use a position-based slug (`callSlug`)
- `lib/runtime/promptRunner.ts` — `pr.parallel` threads the item index into the branch body
- `lib/runtime/deterministicClient.ts` — `MultiToolCallMock` accepts an optional per-call `id` so tests can simulate an id-less provider (pass `id: ""`)
- `tests/agency-js/parallel-tools-no-ids/` — regression test: two id-less parallel tool calls in one round must not collide

## Testing

- New regression test reproduces the exact error without the fix and passes with it.
- Unit tests pass: `promptRunner`, `deterministicClient`, `prompt`, `agencyLlm`, `agencyLlm.checkpointInfo`, `agencyInterrupt`, `interrupts` (94 tests).
- Structural lint clean.
- The pre-existing `parallel-tools` timing assertion (`ranInParallel`) is unaffected by this change — its `data` result stays `"both done"`; only its wall-clock threshold is machine-load sensitive.

### DIFF
--- a/packages/agency-lang/lib/runtime/deterministicClient.test.ts
+++ b/packages/agency-lang/lib/runtime/deterministicClient.test.ts
@@ -59,6 +59,37 @@ describe("DeterministicClient", () => {
     }
   });
 
+  it("returns multiple tool calls with synthetic ids by default", async () => {
+    const client = new DeterministicClient([
+      { toolCalls: [{ name: "a" }, { name: "b" }] },
+    ]);
+    const result = await client.text(baseConfig);
+    expect(result.success).toBe(true);
+    if (result.success) {
+      expect(result.value.toolCalls).toHaveLength(2);
+      // Distinct, non-empty synthetic ids so branch keying never collides.
+      expect(result.value.toolCalls[0].id).toBe("mock-tool-1-0");
+      expect(result.value.toolCalls[1].id).toBe("mock-tool-1-1");
+    }
+  });
+
+  it("honors an explicit empty tool-call id (id-less provider like Gemini)", async () => {
+    // Google Gemini returns no tool-call id; smoltalk defaults the missing
+    // id to "". Reproducing that requires the mock to emit "" verbatim
+    // (nullish default, not `||`), which is what the agency-js
+    // parallel-tools-no-ids regression relies on.
+    const client = new DeterministicClient([
+      { toolCalls: [{ name: "a", id: "" }, { name: "b", id: "" }] },
+    ]);
+    const result = await client.text(baseConfig);
+    expect(result.success).toBe(true);
+    if (result.success) {
+      expect(result.value.toolCalls).toHaveLength(2);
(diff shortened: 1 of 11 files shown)


egonSchiele:
Added the suggested coverage in 1ec30c70:

- **`tests/agency-js/parallel-tools-no-ids-resume`** — two id-less (`""`) parallel tool calls in one round, one interrupts. After `respondToInterrupts`, the second LLM call must carry **both** tool responses exactly once. I verified this test catches a step-path-only regression: patching `dist` to keep the branch key unique but revert the step paths to the empty id makes `interruptTool`'s `.invoke` step get marked done by `succeedTool` and skipped on resume — the run still returns `"all done"` with 2 LLM calls, but `INTERRUPT_RESULT` is silently dropped, and the fixture assertion fails. A weaker check on `finalData` alone would have missed it.
- **`deterministicClient.test.ts`** — locks in the `MultiToolCallMock` id behavior the fixtures rely on (distinct synthetic ids by default; explicit `""` preserved via nullish default, not `||`). Verified it fails if the passthrough regresses.

PR #395, opened 2026-07-02
Agent image generation + auto-attached user images/PDFs

egonSchiele:
## Summary

Wires the merged multimodal capability into the built-in agent, per the spec (`docs/superpowers/specs/2026-07-02-agent-images-and-attachments-design.md`, included in this PR with its plan and plan-review):

1. **`generateImageFile` coordinator tool** — generates (or, with input paths in `images`, edits) an image via `std::image` and saves it to disk through the `writeBinary` interrupt gate, resolving paths against the agent cwd. Returns the saved path — never base64 — and a failure message when generation or the write fails (never a false success).
2. **Auto-attached user files** — `agentReplyVia` detects image/PDF paths in the user's message (drag-drop quoting, escaped spaces, `~/`, trailing punctuation), inlines the bytes as base64 attachments, and passes a multimodal array to `mainAgent` (widened to `string | (string | Attachment)[]`). Every attach/skip prints a visible `📎` line. Direct-to-subagent routes stay text-only by design, pinned by a test.

Supporting changes:

- **`std::llm.modelSupportsInput(model, "image" | "pdf")`** — tri-state probe over the same smoltalk predicate the send-time modality gate uses; the wiring drops attachments only on an explicit catalog `false` (unknown models attach optimistically, matching the send gate).
- **Thread transcripts render attachment placeholders** — `_contentToString` (backs the summarizer transcript and `getThread`) used to `JSON.stringify` multimodal content, which would dump base64 into the summarize prompt; parts now render as `[image attachment]` / `[file attachment: name]`.
- **Compiler fix: raw backslashes in emitted template literals** — the parser fully unescapes string literals (`"\\"` → one backslash char) but the template printer passed part text through raw, so a lone backslash corrupted the generated code. The printer now escapes backslashes; fixtures regenerated (68 files change by one banner line each).

## Design decisions (from the spec)

- Attachments are **inlined as base64 at attach time**, not stored as paths: smoltalk re-resolves path sources on every send of the persistent `session: "main"` thread, so a later-deleted file would poison every subsequent turn, an edited file would silently rewrite history, and the read would bypass the `std::readBinary` policy gate.
- Over-limit files (> 20 MB, smoltalk's cap) are **skipped at detection with a visible reason** — reaching the send-time gate would fail the user's whole turn.
- Text-only turns keep passing a bare string to `llm()` so the stored thread shape (and serialized sessions/fixtures) is unchanged for the 99% path.

## Testing

- 22 execution tests for `detectAttachments`/`modalityFilter` (quoting, escapes, tilde, dedupe, cap, size-skip, all six MIME extensions, case-insensitivity, order preservation, directory/nonexistent/code-path rejection, modality drops for image AND pdf).
- 6 execution tests for `generateImageFile` (agent-cwd vs process-cwd negative assertion, absolute paths, write-failure reporting, edit-input resolution, cost accrual).
- 4 turn-integration tests via the `agentTurn` harness (multimodal array reaches the thread — asserted through the new placeholder rendering — plain-string turns unchanged, modality skip end-to-end, subagent route text-only).
- 8 unit tests for `_contentToString`, 5 for `_modelSupportsInput`, 3 new printer tests for the backslash fix.
- Full local verification: `lint:structure` clean, full agent suite 120/120, full vitest unit suite 6858 passed — the 19 failures present are all pre-existing on main at e4ab093b (effect-declaration tests; verified by running them on the untouched main checkout).

Known coverage gaps are named in the plan (generation-failure branch not exercisable with the always-succeeding deterministic image client; `📎` print lines have no capture surface; read-failure skip branch).

## Behavior note: string-escape semantics in generated code (post-review)

The backslash printer fix changes what escape sequences inside `"""` multiline strings render to at runtime. The new behavior matches the documented semantics (`basic-syntax.md`: `"""` strings do not interpret backslash escapes) — previously the raw pass-through let the emitted JS template literal interpret them. Any user program that relied on the old accidental interpretation (e.g. `\"` in a `"""` string arriving as `"`) will see its output change on upgrade; the two in-repo prompts written against the old behavior (`judgePairwise.agency`, the code agent's markdown guidance) are updated in this PR.

## Review follow-ups applied

All 8 confirmed/plausible findings + the 4 minors from review: media-path scanning moved to a single-pass TS scanner (`lib/stdlib/mediaPathScan.ts`) with boundary-gated quote handling (apostrophes in prose can no longer defeat detection), `std::shell.stat` gained `followSymlinks` (symlinked media now attaches), cap overflow gets a visible `📎 skipped` entry, CR is escaped in emitted templates, URL/data-URI edit inputs pass through unresolved, unknown source-carrying thread parts render as placeholders instead of JSON, `modalityFilter` early-returns on text-only turns, and `DetectedContent` dropped its redundant `text` field.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -30,10 +30,11 @@ import { exists } from "std::shell"
 import { commandsDir, expandSlash } from "std::skills"
 import { highlight } from "std::syntax"
 import { cwd, env, isTTY, readStdin, setTitle } from "std::system"
-import { getCost, getModelCosts, systemMessage } from "std::thread"
+import { Attachment, getCost, getModelCosts, systemMessage } from "std::thread"
 import { chooseOption, ChoiceItem } from "std::ui"
 
 import { figs, title } from "./images/images.agency"
+import { detectAttachments, modalityFilter } from "./lib/attachments.agency"
 import { POLICY_PATH, HISTORY_PATH, ALWAYS_FIELDS } from "./lib/config.agency"
 import {
   minimalAutoApprovePolicy,
@@ -867,7 +868,7 @@ static const mainAgentTools = [
   generateImageFile,
 ]
 
-def mainAgent(prompt: string): string {
+def mainAgent(prompt: string | (string | Attachment)[]): string {
   thread(label: "main", summarize: true, session: "main") {
     setMemoryId("main")
     if (first) {
@@ -907,7 +908,28 @@ export def agentReplyVia(target: string, userMsg: string): string {
   if (target == "review") {
     return reviewAgent(expanded, false)
   }
-  return mainAgent(expanded)
+  // Auto-attach image/PDF files the user referenced (drag-drop or typed
+  // path), filtered to what the resolved model accepts. Never silent:
+  // every attach and every skip prints a visible line.
+  const detected = modalityFilter(detectAttachments(expanded))
+  for (skippedFile in detected.skipped) {
+    pushMessage(color.yellow("📎 skipped ${skippedFile.label} (${skippedFile.reason})"))
+  }
+  if (detected.attached.length == 0) {
+    // Deliberate special case: a text-only turn stays a bare string. A
+    // one-element array is API-equivalent, but the stored thread message
+    // shape (string vs parts array) — and with it every serialized
+    // session and message fixture — would change for EVERY turn.
+    return mainAgent(expanded)
+  }
+  for (item in detected.attached) {
+    pushMessage(color.dim("📎 attached ${item.label}"))
+  }
+  const attachments = map(detected.attached) as item {
+    return item.attachment
+  }
+  const parts: (string | Attachment)[] = [expanded, ...attachments]
+  return mainAgent(parts)
 }
 
 // The agent's core turn, decoupled from the terminal: expand any project
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -5,6 +5,7 @@ import {
   listModelNames,
  } from "std::agency/local"
 import { parseArgs } from "std::args"
+import { map } from "std::array"
 import {
   clearMessages,
   pushMessage,
@@ -13,7 +14,9 @@ import {
   clearHistory,
  } from "std::ui/cli"
 import { today } from "std::date"
-import { setAgentCwd } from "std::index"
+import { generateImage } from "std::image"
+import { applyAgentCwd, setAgentCwd } from "std::index"
+import { basename, dirname, isAbsolute } from "std::path"
 import { box, render } from "std::ui/layout"
 import { listHostedModels, hostedModelInfo } from "std::llm"
 import { setMemoryId, disableMemory } from "std::memory"
@@ -639,8 +642,8 @@ static const mainAgentSystemPrompt = """
 You are the top-level coordinator of an Agency-language assistant. You
 receive every user message and decide how to respond.
 
-You have five subagent tools, each running in its own isolated
-context:
+You have five subagent tools (each running in its own isolated
+context) and one direct tool:
 
 - `codeAgent(userMsg)` — anything that touches code or the filesystem:
   reading, writing, editing, typechecking, running shell commands,
@@ -657,6 +660,11 @@ context:
 - `explorerAgent(userMsg)` — a read-only researcher that produces
   broad, synthesizing answers about the codebase or bundled docs.
   Read the **Explorer** section below.
+- `generateImageFile(prompt, path, size, images)` — generate an image
+  from a text prompt (or modify existing images by passing their paths
+  in `images`) and save it to `path`. Call it directly whenever the
+  user asks you to create, draw, edit, or restyle an image — do NOT
+  route image generation to `codeAgent`.
 
 **Answer directly (no tool call) when** the message is conversational,
 a clarifying question, or something you can answer from context alone.
@@ -799,12 +807,64 @@ unless the user has clearly asked for an action ("do X", "fix Y",
 with them — don't sprint to implementation.
 """
 
+// Exported for tests: agent-cwd resolution of edit inputs is a spec
+// invariant with no other observable (the deterministic image client
+// ignores its images argument).
+export def resolveEditInputs(images: string[]): string[] {
+  return map(images) as inputPath {
+    return applyAgentCwd(inputPath)
+  }
+}
+
+// Direct image generation for the coordinator. Saves to disk and
+// returns the path — never base64, which would flood the LLM context.
+// The write rides the std::writeBinary interrupt gate (writes are not
+// auto-approved by policy) and generation cost accrues to the agent's
+// cost tracking like any llm() call.
+export def generateImageFile(
+  prompt: string,
+  path: string,
+  size: string = "",
+  images: string[] = [],
+): string {
+  """
+  Generate an image from a text prompt and save it to `path`. To MODIFY
+  existing images (edit / variation), pass their paths in `images`.
+  Returns the saved path on success. Use this when the user asks to
+  create, draw, edit, or restyle an image.
+
+  @param prompt - What to generate, or how to modify the input images.
+  @param path - Where to save the resulting image (e.g. "diagram.png").
+  @param size - Optional size like "1024x1024".
+  @param images - Optional input image paths to edit / vary.
+  """
+  // LLM-supplied paths resolve against the agent cwd, like every other
+  // agent file tool.
+  const inputs = resolveEditInputs(images)
+  const generated = generateImage(prompt, size: size, images: inputs)
+  if (isFailure(generated)) {
+    return "Image generation failed: ${generated.error}"
+  }
+  // writeBinary rejects an absolute filename when dir is set, so split
+  // path into (dir, name) and resolve the dir against the agent cwd.
+  let dir = dirname(path)
+  if (!isAbsolute(dir)) {
+    dir = applyAgentCwd(dir)
+  }
+  const written = writeBinary(basename(path), generated.value.base64, dir)
+  if (isFailure(written)) {
+    return "Generated the image, but saving to ${path} failed: ${written.error}"
+  }
+  return "Saved image to ${path}"
+}
+
 static const mainAgentTools = [
   researchAgent.partial(allowHandoff: false),
   codeAgent.partial(allowHandoff: false),
   reviewAgent.partial(allowHandoff: false),
   oracleAgent.partial(allowHandoff: false),
   explorerAgent.partial(allowHandoff: false),
+  generateImageFile,
 ]
 
 def mainAgent(prompt: string): string {
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -16,7 +16,7 @@ import {
 import { today } from "std::date"
 import { generateImage } from "std::image"
 import { applyAgentCwd, setAgentCwd } from "std::index"
-import { basename, dirname, isAbsolute } from "std::path"
+import { basename, dirname } from "std::path"
 import { box, render } from "std::ui/layout"
 import { listHostedModels, hostedModelInfo } from "std::llm"
 import { setMemoryId, disableMemory } from "std::memory"
@@ -810,9 +810,15 @@ with them — don't sprint to implementation.
 
 // Exported for tests: agent-cwd resolution of edit inputs is a spec
 // invariant with no other observable (the deterministic image client
-// ignores its images argument).
+// ignores its images argument). generateImage accepts path / URL /
+// data-URI strings, so only local paths get agent-cwd resolution —
+// running a URL through applyAgentCwd would mangle it into
+// "<agentCwd>/https:/…".
 export def resolveEditInputs(images: string[]): string[] {
   return map(images) as inputPath {
+    if (inputPath.startsWith("http://") || inputPath.startsWith("https://") || inputPath.startsWith("data:")) {
+      return inputPath
+    }
     return applyAgentCwd(inputPath)
   }
 }
@@ -847,12 +853,9 @@ export def generateImageFile(
     return "Image generation failed: ${generated.error}"
   }
   // writeBinary rejects an absolute filename when dir is set, so split
-  // path into (dir, name) and resolve the dir against the agent cwd.
-  let dir = dirname(path)
-  if (!isAbsolute(dir)) {
-    dir = applyAgentCwd(dir)
-  }
-  const written = writeBinary(basename(path), generated.value.base64, dir)
+  // path into (dir, name); useAgentCwd resolves a relative dir against
+  // the agent cwd (absolute dirs pass through untouched).
+  const written = writeBinary(basename(path), generated.value.base64, dirname(path), useAgentCwd: true)
   if (isFailure(written)) {
     return "Generated the image, but saving to ${path} failed: ${written.error}"
   }
--- a/packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
@@ -1,0 +1,155 @@
+import { applyAgentCwd } from "std::index"
+import { stat } from "std::shell"
+import { env } from "std::system"
+import { basename, extname, dirname, isAbsolute } from "std::path"
+import { Attachment, image, file } from "std::thread"
+
+// Detects image/PDF file paths in a user message (terminal drag-drop
+// inserts a quoted/escaped path; typed mentions are plain tokens) and
+// returns them as inline-base64 attachments for the coordinator turn.
+// Media-only by design: code/text paths stay with the read tool.
+
+export type DetectedAttachment = { attachment: Attachment, label: string }
+export type SkippedFile = { label: string, reason: string }
+
+export type DetectedContent = {
+  text: string,
+  attached: DetectedAttachment[],
+  skipped: SkippedFile[]
+}
+
+// A token that survived the pure (no-I/O) gates: normalized, media
+// extension, resolved to an absolute path.
+type Candidate = { abs: string, mime: string }
+
+// extension (lowercase, with dot) -> MIME type. Media-only by design.
+static const MIME_TYPES: Record<string, string> = {
+  ".png": "image/png",
+  ".jpg": "image/jpeg",
+  ".jpeg": "image/jpeg",
+  ".gif": "image/gif",
+  ".webp": "image/webp",
+  ".pdf": "application/pdf"
+}
+
+static const MAX_ATTACHMENTS = 10
+
+// Split msg into candidate tokens. Shell-quoted spans ('…' / "…")
+// become single tokens (drag-drop quotes paths with spaces), and a
+// backslash-escaped space is unescaped into the current token.
+// Hand-written scanner — order-dependent state is fine here (parsers
+// are exempt per docs/dev/anti-patterns.md).
+def tokenize(msg: string): string[] {
+  let tokens: string[] = []
+  let current = ""
+  let quote = ""
+  let index = 0
+  while (index < msg.length) {
+    const ch = msg.slice(index, index + 1)
+    if (quote != "") {
+      if (ch == quote) {
+        quote = ""
+      } else {
+        current = current + ch
+      }
+    } else if (ch == "'" || ch == "\"") {
+      quote = ch
+    } else if (ch == "\\" && index + 1 < msg.length && msg.slice(index + 1, index + 2) == " ") {
+      current = current + " "
+      index = index + 1
+    } else if (ch == " " || ch == "\t" || ch == "\n") {
+      if (current != "") {
+        tokens.push(current)
+        current = ""
+      }
+    } else {
+      current = current + ch
+    }
+    index = index + 1
+  }
+  if (current != "") {
+    tokens.push(current)
+  }
+  return tokens
+}
+
+// Trim trailing punctuation typed after a path, and expand a leading ~/
+// (drag-drop inserts absolute paths, but typed mentions are often ~/…).
+def normalizeToken(token: string): string {
+  let normalized = token
+  while (normalized.length > 0 && (normalized.endsWith(",") || normalized.endsWith(".") || normalized.endsWith("?") || normalized.endsWith(":"))) {
+    normalized = normalized.slice(0, normalized.length - 1)
+  }
+  if (normalized.startsWith("~/")) {
+    const home = env("HOME")
+    if (home != null && home != "") {
+      normalized = home + normalized.slice(1)
+    }
+  }
+  return normalized
+}
+
+// Pure stage (spec steps 2-3 + extension gate): normalize, keep only
+// media extensions, resolve to an absolute path. Returns null for
+// tokens that are not candidate media paths. No filesystem access.
+def tokenToCandidate(token: string): Candidate | null {
+  const normalized = normalizeToken(token)
+  const ext = extname(normalized).toLowerCase()
+  const mime = MIME_TYPES[ext]
+  if (mime == null) {
+    return null
+  }
+  if (isAbsolute(normalized)) {
+    return { abs: normalized, mime: mime }
+  }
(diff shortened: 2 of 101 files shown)


packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
```diff
@@ -0,0 +1,189 @@
...
+  ".jpg": "image/jpeg",
+  ".jpeg": "image/jpeg",
+  ".gif": "image/gif",
+  ".webp": "image/webp",
+  ".pdf": "application/pdf"
+}
+
+static const MAX_ATTACHMENTS = 10
+
+// Split msg into candidate tokens. Shell-quoted spans ('…' / "…")
```
egonSchiele: What are you trying to parse here? I imagine this code will be quite slow to parse. I would suggest writing a parser in TypeScript using Tarsec instead and importing it and using it here. Lots of examples of Tarsec parsers in lib/parsers.ts.

packages/agency-lang/lib/agents/agency-agent/tests/imageTool.agency
```diff
@@ -0,0 +1,43 @@
+import { generateImageFile, resolveEditInputs } from "../agent.agency"
+import { setAgentCwd } from "std::index"
+import { cwd } from "std::system"
+import { getCost } from "std::thread"
+
+node savesToAgentCwd(): string {
+  setAgentCwd("/tmp")
+  const msg = generateImageFile("a red bicycle", "it-gencwd.png") with approve
```
egonSchiele, line 8: Is there a way to mock this? Generating an image takes a really long time, so this will severely slow down the tests. I know that tests use the deterministic client on CI, but that is only for PRs... when code gets pushed to main, CI runs the actual agent, and if it has to generate images, that will take a really long time. I would rather we just not test the image generation code, or only test it once.

egonSchiele:
How are attachments counted in terms of tokens when sent in user messages? I attached an image to a message, and I noticed that the token count for my message was just five tokens, which doesn't seem right.

```
─── ↑5 ↓326 19.9s · claude-sonnet-4-6 · $0.0324 ───
```

egonSchiele (commented):
Automated review (8 finder angles + adversarial verification per finding). Verdicts: **CONFIRMED** = reproduced or traced end-to-end with quoted code; **PLAUSIBLE** = mechanism real, trigger uncertain.

The core implementation is solid and well-tested. Findings are posted inline, most-severe first — the top cluster is around the "never silent" attach contract, plus the multiline-string consequence of the printer fix. Four minor cleanup notes are also inline.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
```diff
@@ -0,0 +1,189 @@
...
+  let index = 0
+  while (index < msg.length) {
+    const ch = msg.slice(index, index + 1)
+    if (quote != "") {
+      if (ch == quote) {
+        quote = ""
+      } else {
+        current = current + ch
+      }
+    } else if (ch == "'" || ch == "\"") {
```
egonSchiele: **[1 — CONFIRMED, reproduced]** `tokenize()` treats any apostrophe in ordinary prose as opening a shell quote, so a contraction silently defeats detection.

Reproduced against the built dist: `here's /tmp/pic.png` (file exists) → 0 attached, 0 skipped; the apostrophe-free control attaches. The `'` in "here's" opens quote mode, spaces stop splitting, and the end-of-string fallthrough emits the whole remainder as one token (`heres /tmp/pic.png`), which resolves to a nonexistent cwd-relative path and is dropped with no `skipped` entry — the user believes the model saw the image but it never did. An odd quote count also inverts quoting for the rest of the message, so even a correctly quoted drag-drop path later in the same message mis-tokenizes.

The tests cover only balanced quotes (`quotedPathWithSpaces`, `doubleQuotedPath`); there is no apostrophe-in-prose test. Consider treating quote chars as quoting only when they delimit a path-shaped span, or falling back to unquoted tokenization when a quote never closes.

packages/agency-lang/lib/ir/prettyPrint.ts
```diff
@@ -144,15 +144,18 @@ export function printTs(node: TsNode, indent = 0): string {
...
+      //                    unterminated template; must run FIRST so it
+      //                    doesn't double the escapes added below)
       //   `` ` `` → `\``  (otherwise closes the template)
       //   `${`    → `\${` (otherwise starts an interpolation)
-      // We deliberately do NOT escape `\` — agency strings pass backslash
-      // sequences through to the JS template literal so `\n`, `\t`, etc.
-      // get interpreted as escape sequences at runtime.
       const escapeForTemplate = (text: string): string =>
-        text.replace(/`/g, "\\`").replace(/\$\{/g, "\\${");
+        text.replace(/\\/g, "\\\\").replace(/`/g, "\\`").replace(/\$\{/g, "\\${");
```
egonSchiele: **[2 — CONFIRMED mechanism; intent defensible but two shipped prompts are now garbled]** This fix changes what `\n`/`\"` inside `"""` multiline strings render to at runtime, and the PR doesn't update the in-repo victims.

The new behavior actually *matches* documented semantics — `basic-syntax.md` says `"""` strings do **not** interpret backslash escapes, and the old printer violated that by letting JS interpret them at runtime. But two files were written against the old behavior and are untouched by this PR:

- `lib/agents/eval/judgePairwise.agency:29` — `Pick \"A\", \"B\", or \"tie\"` used to reach the judge LLM as `Pick "A", "B", or "tie"`; it now arrives with literal backslashes before every quote.
- `lib/agents/agency-agent/subagents/code.agency:272-273` — the `## Heading\\n\\n- item` markdown examples now render with doubled backslashes in every code-agent prompt.

Both prompts should be updated in this PR, and the multiline semantic flip deserves a line in the PR description — any user program relying on the old (undocumented) escape interpretation in `"""` strings silently changes output on upgrade.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
```diff
@@ -0,0 +1,189 @@
...
+  let seen: string[] = []
+  // Positive `if` nesting rather than `continue` guards: flow narrowing
+  // (candidate != null, isSuccess) is only tracked into guarded blocks.
+  for (token in tokenize(msg)) {
+    const candidate = tokenToCandidate(token)
+    if (candidate != null) {
+      if (attached.length < MAX_ATTACHMENTS && !seen.includes(candidate.abs)) {
+        seen.push(candidate.abs)
+        const info = stat(candidate.abs)
+        if (info.type == "file") {
```
egonSchiele: **[3 — CONFIRMED]** Symlinks to images/PDFs are silently dropped.

`std::shell` `stat` → `_stat` (`lib/stdlib/shell.ts:506`) uses `fs.lstat` and checks `isSymbolicLink()` first, so a symlink to a regular PNG reports `type: "symlink"`, never `"file"`. This `if (info.type == "file")` gate has no else branch, so e.g. `~/latest.png → <newest screenshot>` is neither attached nor reported in `skipped` — the model answers about an image it never received.

Either follow symlinks here (stat the target) or push a `skipped` entry so the drop is visible.

packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -799,15 +808,67 @@ unless the user has clearly asked for an action ("do X", "fix Y",
 with them — don't sprint to implementation.
 """
 
+// Exported for tests: agent-cwd resolution of edit inputs is a spec
+// invariant with no other observable (the deterministic image client
+// ignores its images argument).
+export def resolveEditInputs(images: string[]): string[] {
+  return map(images) as inputPath {
+    return applyAgentCwd(inputPath)
```
egonSchiele, line 822: **[4 — CONFIRMED]** `resolveEditInputs` mangles URL/data-URI edit inputs.

`std::image.generateImage` documents and implements `images` as path / URL / data-URI strings (`classifySource` in `lib/stdlib/thread.ts` handles `https://` and `data:`), but this maps every entry through `applyAgentCwd` unconditionally, turning `https://example.com/cat.png` into `<agentCwd>/https:/example.com/cat.png`. The tool docstring says "input image paths", but when a user pastes an image URL the LLM's natural move is to forward it, and the failure is a confusing local-file error.

Skip resolution for entries starting with `data:` / `http://` / `https://` (mirroring `classifySource`).

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
```diff
@@ -0,0 +1,189 @@
...
+  """
+  let attached: DetectedAttachment[] = []
+  let skipped: SkippedFile[] = []
+  let seen: string[] = []
+  // Positive `if` nesting rather than `continue` guards: flow narrowing
+  // (candidate != null, isSuccess) is only tracked into guarded blocks.
+  for (token in tokenize(msg)) {
+    const candidate = tokenToCandidate(token)
+    if (candidate != null) {
+      if (attached.length < MAX_ATTACHMENTS && !seen.includes(candidate.abs)) {
```
egonSchiele: **[5 — CONFIRMED]** Files beyond the `MAX_ATTACHMENTS` cap are silently discarded, contradicting the "Never silent: every attach and every skip prints a visible line" contract at the `agentReplyVia` call site.

This guard has no else pushing a skip reason (unlike the size and read failures below, which do). A user dragging 12 photos sees 10 `📎 attached` lines and nothing for the other 2, and the model confidently answers about images it never saw. The `capsAtTen` test asserts only the attached count, not `skipped`.

Trivial fix: an else branch pushing a `"too many attachments"` skip entry.

packages/agency-lang/lib/ir/prettyPrint.ts
egonSchiele: **[6 — CONFIRMED, pre-existing but the new comment overclaims completeness]** `escapeForTemplate` misses carriage return.

The parser unescapes `\r` to a real CR (`parsers.ts` `case "r":`), this function emits it raw, and ECMAScript template-literal semantics normalize a raw CR (and CRLF) in source to LF — so `"a\rb"` compiles to a template literal whose runtime value is `"a\nb"` (verified end-to-end by compiling in the worktree). The gap predates this PR, but the rewritten comment now claims exactly three sequences need escaping; CR is a fourth — it doesn't corrupt the template syntactically, it silently changes the value.

Fix is one more `.replace(/\r/g, "\\r")`.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
```diff
@@ -0,0 +1,189 @@
...
+}
+
+static const MAX_ATTACHMENTS = 10
+
+// Split msg into candidate tokens. Shell-quoted spans ('…' / "…")
+// become single tokens (drag-drop quotes paths with spaces), and a
+// backslash-escaped space is unescaped into the current token.
+// Hand-written scanner — order-dependent state is fine here (parsers
+// are exempt per docs/dev/anti-patterns.md).
+def tokenize(msg: string): string[] {
```
egonSchiele: **[7 — efficiency]** Detection runs a per-character Agency `while` loop through the Runner step machinery on **every** user message, with no cheap bail-out.

Each character costs two `runInScope` async closures plus locals-scan bookkeeping (`runner.ts` `whileLoop`); pasting a 100KB log into the agent burns ~100k stepped iterations before the LLM call, on a path where >99% of messages contain no media path. Related: `modalityFilter` pays `getResolvedSlots` + two `modelSupportsInput` catalog probes on every text-only turn even when `attached` is empty.

Cheap wins: bail out of `detectAttachments` unless the raw string contains a media extension (single regex/`includes` check), and early-return from `modalityFilter` when `detected.attached.length == 0`. Longer-term, the scanner could be a plain-TS stdlib helper so it's one JS loop instead of Runner-stepped Agency code.

packages/agency-lang/lib/stdlib/threads.ts
```diff
@@ -30,16 +30,42 @@ import { registerGlobalHook } from "../runtime/hooks.js";
...
+          return String(typedPart.text ?? "");
+        }
+        if (typedPart.type === "image") {
+          return "[image attachment]";
+        }
+        if (typedPart.type === "file") {
+          return typedPart.filename ? `[file attachment: ${typedPart.filename}]` : "[file attachment]";
+        }
+      }
+      return JSON.stringify(part);
```
egonSchiele, line 72: **[8 — PLAUSIBLE, defense-in-depth]** The unknown-part-kind fallback is `JSON.stringify(part)`, which would dump a future smoltalk part's base64 payload into transcripts and the summarizer prompt — the exact leak class these placeholders exist to prevent. The "falls back to JSON for unknown part types" test pins the leaking behavior as correct.

A payload-safe fallback closes it: any object part with a `source` field renders as `[<type> attachment]` (or truncate the stringified output), so a new smoltalk part kind can't silently regress into megabytes-into-an-LLM-call.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// Trim trailing punctuation typed after a path, and expand a leading ~/
+// (drag-drop inserts absolute paths, but typed mentions are often ~/…).
+def normalizeToken(token: string): string {
+  let normalized = token
+  while (normalized.length > 0 && (normalized.endsWith(",") || normalized.endsWith(".") || normalized.endsWith("?") || normalized.endsWith(":"))) {
+    normalized = normalized.slice(0, normalized.length - 1)
+  }
+  if (normalized.startsWith("~/")) {
```
egonSchiele: **[minor — reuse]** `normalizeToken` re-implements tilde expansion locally, but `lib/stdlib/expandPath.ts` declares itself "the single owner of path-shorthand policy for the stdlib... never re-implement the policy locally". The local copy already diverges: `env("HOME")` instead of `os.homedir()` (silently leaves `~/x.png` undetected when HOME is unset, no Windows `USERPROFILE` fallback), no bare `~`, no `~user` rejection. Consider exposing `expandPath` as a std::path bridge (like `_resolve`/`_basename`) and calling it here.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
```diff
@@ -0,0 +1,189 @@
...
+// media extensions, resolve to an absolute path. Returns null for
+// tokens that are not candidate media paths. No filesystem access.
+def tokenToCandidate(token: string): Candidate | null {
+  const normalized = normalizeToken(token)
+  const ext = extname(normalized).toLowerCase()
+  const mime = MIME_TYPES[ext]
+  if (mime == null) {
+    return null
+  }
+  if (isAbsolute(normalized)) {
```
egonSchiele: **[minor — simplification]** This `isAbsolute` branch duplicates `applyAgentCwd`'s own documented contract ("`_resolve` short-circuits on an absolute `dir`" — and `_resolve` is `path.resolve`; the PR's own `resolveEditInputs` test pins absolute pass-through). A single `return { abs: applyAgentCwd(normalized), mime: mime }` is equivalent and drops the duplicated object literal and the `isAbsolute` import.

packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -799,15 +808,67 @@ unless the user has clearly asked for an action ("do X", "fix Y",
...
+  // agent file tool.
+  const inputs = resolveEditInputs(images)
+  const generated = generateImage(prompt, size: size, images: inputs)
+  if (isFailure(generated)) {
+    return "Image generation failed: ${generated.error}"
+  }
+  // writeBinary rejects an absolute filename when dir is set, so split
+  // path into (dir, name) and resolve the dir against the agent cwd.
+  let dir = dirname(path)
+  if (!isAbsolute(dir)) {
```
egonSchiele: **[minor — simplification]** The `isAbsolute` guard is redundant — `applyAgentCwd` already short-circuits on absolute dirs (its docstring: "`_resolve` short-circuits on an absolute `dir`"), so `const dir = applyAgentCwd(dirname(path))` does the same job. Alternatively, `writeBinary` already has `useAgentCwd: true` for exactly this resolution (the established `.partial(useAgentCwd: true)` pattern in the subagents), which would drop the manual `applyAgentCwd` call entirely — the split into (dir, name) still stands per the comment above.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
```diff
@@ -0,0 +1,189 @@
...
+
+// Detects image/PDF file paths in a user message (terminal drag-drop
+// inserts a quoted/escaped path; typed mentions are plain tokens) and
+// returns them as inline-base64 attachments for the coordinator turn.
+// Media-only by design: code/text paths stay with the read tool.
+
+export type DetectedAttachment = { attachment: Attachment, label: string }
+export type SkippedFile = { label: string, reason: string }
+
+export type DetectedContent = {
```
egonSchiele, line 24: **[minor — simplification]** `DetectedContent.text` is redundant state: it's always the unmodified input (`agentReplyVia` uses `expanded` directly; only two test nodes echo it back), yet every producer/transformer must thread it through. Dropping the field so `detectAttachments`/`modalityFilter` return `{ attached, skipped }` shrinks the type with zero behavior change.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
egonSchiele: Moved to TS in 5dcb179a: `lib/stdlib/mediaPathScan.ts`, imported via `agency-lang/stdlib-lib/mediaPathScan.js`. It is one JS pass with a cheap regex bail-out for text-only messages, so the per-message cost is a single regex test in the common case (the Agency side keeps only the stateful stages: agent-cwd resolution, dedupe, stat, read).

I did look at doing it with tarsec, but went with plain TS + anchored regexes: this is span-extraction from arbitrary prose rather than a structured parse, and the quote handling needs boundary + content guards (an apostrophe in "here's" must not open a span; a quoted span only counts when it is its own whitespace-delimited token AND ends in a media extension — that guard also fixes the confirmed apostrophe bug below). Expressing fail-and-backtrack on those semantic guards in tarsec would mean a custom parser function anyway. Happy to convert it to a tarsec grammar if you'd still prefer that shape — the scanner is isolated behind `_scanMediaPaths` with 15 unit tests, so swapping the internals is safe.

packages/agency-lang/lib/agents/agency-agent/tests/imageTool.agency
egonSchiele: These tests never hit a real image provider, including on push-to-main: `"useTestLLMProvider": true` is set on every test in this file, and per the field's docs in `lib/cli/test.ts` it forces the deterministic client "even when the suite is not launched with AGENCY_USE_TEST_LLM_PROVIDER=1 ... Without this flag such tests would silently fall through to the real OpenAI client on push-to-main CI". The deterministic client's `image()` returns a fixed 1x1 PNG with a fixed $0.04 cost and no network call — same pattern as the existing `tests/agency-js/image-generation` suite. So the suite stays fast everywhere; the trade-off (noted in the plan's known gaps) is that no test exercises a real generation.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
egonSchiele: Fixed in 5dcb179a. The TS scanner only treats a quote as opening a span when it sits at a token boundary AND the span content ends in a media extension, so apostrophes in prose fall through to word tokenization. Both reproduced cases are pinned: `apostropheInProse` ("here's /tmp/pic.png" → attached) and `apostropheThenQuotedPath` ("here's '/tmp/da b.png'" → the later quoted path still attaches), plus scanner-level unit tests in mediaPathScan.test.ts.

packages/agency-lang/lib/ir/prettyPrint.ts
egonSchiele: Fixed in 7be65b9b — both prompts updated (judgePairwise's quoted A/B/tie, plus a second \" instance at line 36 the sweep found; code agent's markdown \\n notation now written as \n). Swept all .agency sources for other backslash escapes: those four lines were the only victims. Added the semantic-flip note to the PR description.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
egonSchiele: Fixed in 5dcb179a: `std::shell.stat` gained a `followSymlinks: boolean = false` param (fs.stat vs lstat; a broken link reports "missing"), and detectAttachments passes `followSymlinks: true`, so the target's type and size gate the attach. Pinned by the `symlinkFollowed` execution test (ln -s to a real PNG → attached).

packages/agency-lang/lib/agents/agency-agent/agent.agency
egonSchiele: Fixed in a1f8a94d: entries starting with `http://` / `https://` / `data:` pass through resolveEditInputs untouched, mirroring classifySource. Pinned by `editInputsPassUrlsThrough`.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
egonSchiele: Fixed in 5dcb179a: files beyond the cap now push a `too many attachments (limit 10)` skip entry. The cap check moved after the existence gate so nonexistent mentions don't generate noise. `capsAtTenWithVisibleSkip` asserts both the attached count and the skip reason.

packages/agency-lang/lib/ir/prettyPrint.ts
egonSchiele: Fixed in 7be65b9b: `.replace(/\r/g, "\\r")` added (after the backslash rule, before backtick/dollar-brace), comment updated to four sequences, and a round-trip test pins `a\rb` surviving eval.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
egonSchiele: Fixed in 5dcb179a. The scanner is now one JS pass in `lib/stdlib/mediaPathScan.ts` with a regex bail-out, so a text-only message costs a single regex test instead of Runner-stepped per-character iterations. `modalityFilter` also early-returns when `attached` is empty — no slot lookup or catalog probes on text-only turns.

packages/agency-lang/lib/stdlib/threads.ts
egonSchiele: Fixed in 9a1e10b2: an unknown part kind carrying a `source` field renders as `[<type> attachment]`; payload-free unknown parts keep the JSON fallback. Test pins that a hypothetical video part's base64 never reaches the transcript.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
egonSchiele: Fixed in 5dcb179a — tilde expansion now goes through `expandPath` in the TS scanner (os.homedir, Windows fallback, bare `~`). One deliberate divergence: `expandPath` throws on `~user/...`, and a media mention must never crash detection, so the scanner only calls it for `~` / `~/...` tokens and lets `~user` paths die quietly at the stat gate (unit-tested).

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
egonSchiele: Fixed in 5dcb179a — the branch collapsed to a single `applyAgentCwd(candidate.path)` with a comment citing the short-circuit; `isAbsolute` import dropped.

packages/agency-lang/lib/agents/agency-agent/agent.agency
egonSchiele: Fixed in a1f8a94d — went with the second option: `writeBinary(basename(path), ..., dirname(path), useAgentCwd: true)`, matching the established agent-tool pattern; the manual applyAgentCwd/isAbsolute dance is gone.

packages/agency-lang/lib/agents/agency-agent/lib/attachments.agency
egonSchiele: Fixed in 5dcb179a — `text` dropped from `DetectedContent`; producers, filter, tests, and the in-repo spec updated.

PR #398, opened 2026-07-02
Subprocess pause/resume: surface unhandled interrupts, resume from checkpoint, unblock nesting

egonSchiele:
## What this does

Unhandled subprocess interrupts now **surface to the user and resume**, closing the biggest hole in the subprocess safety story: previously they were auto-rejected (`"slow-path not yet supported"`). A subprocess whose interrupt nobody handles checkpoints itself, exits, and its interrupts surface exactly like in-process interrupts; `respondToInterrupts` resumes it in a fresh process from exactly where it left off — including concurrent (fork) interrupts, multi-cycle pauses, and full durability across parent process restarts.

Spec: `packages/agency-lang/docs/superpowers/specs/2026-07-01-subprocess-pause-resume-design.md` (untracked by repo convention). Plan + review: `packages/agency-lang/docs/superpowers/plans/2026-07-01-subprocess-pause-resume{,-review}.md`. Dev docs rewritten: `docs/dev/subprocess-ipc.md`.

## How it works

- **Distributed handler chain — the parent reports, the child decides.** The parent replies to a relayed interrupt with its `HandlerChainOutcome` (approved/rejected/propagated/noResponse), never a verdict; the child merges with single-process precedence (`mergeChainOutcomes`) and renders the verdict locally. Child-local reject stays final (fail-fast). Child interrupt ids travel verbatim end-to-end.
- **Pause = the child's own normal propagate path.** A propagate verdict returns `Interrupt[]` through the child's existing runBatch/checkpoint machinery (siblings settle and cache; one shared checkpoint), and the bootstrap converts the final batch into a new `interrupted` terminal message.
- **`_run` is the fourth runBatch adopter** (single child): stores the child checkpoint as an opaque frame-local payload (never spliced into the parent's state tree), returns rehydrated interrupts for runBatch to stamp with the parent-side shared checkpoint. Composes with parent-level fork (two paused subprocesses batch into one respond). Abort signal now kills the child (parent cancellation / race loss / time guards).
- **Resume**: the parent re-forks with a `resume` instruction; the bootstrap hands the batch to the compiled module's own `respondToInterrupts` export. Handler re-registration on both sides is covered by dedicated tests.
- **`CompiledProgram` is now `{ moduleId, code }`** — compile() no longer writes disk; `_run` materializes per fork, deletes per settle. Parent checkpoints are fully self-contained: the durability test JSON-round-trips the interrupts, wipes `.agency-tmp`, and resumes in a fresh Node process.
- **Nested subprocesses are unblocked.** Every `run()` on every hop is gated by its `std::run` interrupt (data now includes the prospective `depth`) flowing through the distributed chain, with a `maxDepth` cap (default 5, hard ceiling 10, tightest ancestor wins) as the runaway backstop. Consultation, pause (Russian-doll checkpoints), resume, and lock brokering (relay to the ROOT lock domain) all compose recursively.
- **Statelog**: the child inherits the parent runId (one trace across processes and segments), the parent wraps each segment in a new `subprocessRun` span whose id the child adopts as its span root, and a stable `subprocessSessionId` correlates segments. Concurrent parent-side handler chains get span isolation.

## Behavior changes

1. **Vote-combining fixes**: child-approve + parent-silent is now an approve (was wrongly rejected); a propagate vote or total silence now surfaces to the user (was rejected).
2. **`compile()` no longer writes to `.agency-tmp/`**; the compiled value carries the code text.
3. **Nested `run()` no longer hard-errors** — gated + depth-capped instead (`nested-blocked` test replaced by `nested-gate-unapproved`).
4. **`__tryCall` passes a bubbled `Interrupt[]` through** instead of wrapping it in `success(...)` — a general fix any interrupting callee under `try` needed.
5. `run()` gains `maxDepth`; `std::run` interrupt data gains `depth`; `RuntimeContext` gains `subprocessDepth` (readable via `agency.ctx()`).

## Spec invariants → tests

1. One logical chain, child decides → `mergeChainOutcomes` units, `vote-child-approve-parent-silent`, `handler-reject`
2. Parent always replies explicitly → outcome-on-error path in `handleInterruptMessage`; `run-multiple-interrupts`
3. Child-local reject is final → `vote-child-reject-parent-approve`, `nested-reject-middle`
4. Interrupt ids preserved verbatim → id-correlation unit test; resume routing in `subprocess-pause-basic`
5. Child checkpoint opaque in parent → payload accessors; `pause-two-subprocesses`
6. Fully self-contained checkpoints → `subprocess-durable-resume` (fresh process + artifact wipe)
7. invoke returns `Interrupt[]`, never throws → `invokeSubprocess` + `__tryCall` fix + serializer round-trip unit
8. Handler re-registration before any resolve → `pause-then-child-handler`, `pause-then-parent-handler`
9. Paused subprocess holds no live resources → child exits on pause; `pause-limit-wallclock-resets`
10. Depth tracked, propagated, capped → `nested-basic` (per-level gate data), `nested-depth-boundary`, `resolveDepthCap` units

## Verification

- All 40 `tests/agency/subprocess/*` execution tests pass; all 4 `tests/agency-js/subprocess-*` pass.
- Full unit suite: 6859 passed; the 19 failures (4 files in the effect-payload/handler-typing area) **fail identically on untouched main** — pre-existing, unrelated.
- `pnpm run lint:structure` clean; `make` + `make fixtures` clean (86 fixtures changed by exactly the one-line `_run` descriptor gaining `maxDepth`).


packages/agency-lang/lib/runtime/result.ts
```diff
@@ -101,6 +101,13 @@ export function isFailure(result: unknown): result is ResultFailure {
 export async function __tryCall(fn: () => any, opts?: FailureOpts): Promise<ResultValue> {
   try {
     const value = await fn();
+    // Interrupts are control flow, not values: a callee that paused on an
+    // unresolved interrupt bubbles the Interrupt[] up as its return value,
+    // and `try` must pass it through untouched — wrapping it in success()
+    // would make the batch look like a completed Result and strand the
+    // paused state. (First hit by `try _run(...)` when a subprocess
+    // pauses; applies to any interrupting callee under `try`.)
+    if (hasInterrupts(value)) return value as any;
```
egonSchiele, line 110: **Assignment-position `try` strands the paused subprocess.** This passthrough is correct, but the compiler only emits the `hasInterrupts` halt check after plain function-call assignments — in `lib/backends/typescriptBuilder.ts` the assignment dispatch checks `value.type === "functionCall"`, and a `tryExpression` RHS falls through to the bare `scopedAssign` branch with no check.

Verified by compiling: `let r = mightPause(x)` generates `if (hasInterrupts(__stack.locals.r)) { ... runner.halt(...); return; }`, but `let r = try mightPause(x)` generates only the `__tryCall` assignment and keeps executing. So `let r = try run(compiled: c, node: "main")` on a pausing subprocess assigns the raw `Interrupt[]` to `r`: `isSuccess(r)` and `isFailure(r)` are both false, the user's failure branch is skipped, the interrupts never surface, and the paused child can never be resumed. Return-position `try` (the stdlib's own `return try _run(...)`) only works because the *caller's* plain-call check catches the batch.

Suggest emitting the same interrupt-halt check for tryExpression assignments (and a regression test with `let r = try <pausing fn>()`).

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -493,8 +673,15 @@ function settleWithLimitFailure(
   extras: Record<string, any> = {},
 ): void {
   if (s.settled) return;
-  try { s.child.kill("SIGKILL"); } catch (_) { /* already gone */ }
-  settle(s, s.resolvePromise, makeLimitFailure(limit, threshold, value, extras));
+  try {
+    s.child.kill("SIGKILL");
```
egonSchiele, line 688: **SIGKILL reaps only the immediate child — nested grandchildren are orphaned.** Now that nesting is unblocked, a mid-tier process can have live children of its own. Both kill sites (here and the abort path at ~line 929) signal only the direct child; `buildForkOptions` sets no `detached`/process-group, there's no tree-kill walk, and the bootstrap registers no `disconnect` handler that would make a child exit when its parent dies (SIGKILL is untrappable, so the mid process can't clean up either).

Concretely: `run(child, wallClock: 2s)` where the child spawns a grandchild doing long work — at 2s the child is SIGKILLed, its wall-clock timer for the grandchild dies with it, and the grandchild keeps running (burning CPU/LLM spend) outside every limit until its event loop happens to drain. Pre-PR this was structurally impossible (`Nested subprocess execution is not supported`).

A grandchild-side `process.on("disconnect", () => process.exit(...))` watchdog in the bootstrap would close most of this cheaply; process groups / tree-kill would close it completely.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -290,19 +360,15 @@ export async function sendInterruptToParent(
...
-      if (msg.type === "decision" && msg.interruptId === outMsg.interruptId) {
+      if (msg.type === "decision" && msg.interruptId === interruptId) {
         process.removeListener("message", handler);
         ipcLog("recv", msg);
-        if (msg.approved) {
-          resolve({ type: "approve", value: msg.value });
-        } else {
-          resolve({ type: "reject", value: msg.value });
-        }
+        resolve(msg.outcome as HandlerChainOutcome);
```
egonSchiele, line 371: **No `disconnect` handling while awaiting the parent's outcome** — `sendLockAcquireToParent` in this same file guards with `process.once("disconnect", ...)` and rejects with "IPC channel closed while waiting for lock", but this promise only ever resolves on a `decision` message.

If the parent dies mid-consult (e.g. root SIGKILLs the mid process on a limit while a grandchild is awaiting its decision), this promise never settles. Empirically a *bare* child then exits 0 silently (the pending listener doesn't ref the drained event loop) — i.e. a mid-pause child vanishes with neither a result nor an `interrupted` message; if anything else keeps the loop alive (statelog sockets, timers, its own children), it hangs instead. Either way the failure is unobservable.

Suggest mirroring the lock path: on `disconnect`, treat it as a rejected/aborted outcome (or `process.exit` with a distinct code) so the child never outlives its parent silently.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -435,6 +604,18 @@ export function buildForkOptions(args: { limits: RunLimits; cwd?: string }): For
...
 
+/** Write the compiled JS to a fresh .agency-tmp/<nanoid>/ dir (under cwd so
+ * Node resolves agency-lang package imports via the project's node_modules)
+ * and return the script path. Called at every fork — initial run and resume
+ * alike — and paired with cleanupTempDir when the session settles. */
+export function materializeCompiledScript(compiled: { moduleId: string; code: string }): string {
+  const tempDir = path.join(process.cwd(), ".agency-tmp", nanoid());
+  mkdirSync(tempDir, { recursive: true });
+  const scriptPath = path.join(tempDir, `${compiled.moduleId}.js`);
+  writeFileSync(scriptPath, compiled.code, "utf-8");
```
egonSchiele, line 622: **No validation that `compiled.code` exists → opaque fs error.** The user-facing Agency type is `CompiledProgram = { moduleId: string }` (`code` is deliberately undeclared), so `let c: CompiledProgram = { moduleId: "x" }` typechecks, and passing it to `run()` reaches this `writeFileSync(scriptPath, undefined)`. Reproduced end-to-end: the user gets a Failure whose message is `The "data" argument must be of type string or an instance of Buffer...` — nothing pointing at the actual problem. Old-shape `{ moduleId, path }` values persisted pre-upgrade hit the same path.

A one-line guard here (or at `_run` entry) with a clear message — e.g. `CompiledProgram has no code; obtain it from compile()` — would make this diagnosable.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -672,23 +886,206 @@ function attachSessionHandlers(s: RunSession, node: string, args: Record<string,
...
+        } catch (err) {
+          ipcLog("send", { type: "kill_failed", detail: err instanceof Error ? err.message : String(err) });
+        }
+        settle(session, rejectPromise, new AgencyCancelledError());
+      };
+      if (opts.abortSignal.aborted) {
+        onAbort();
+        return;
+      }
+      opts.abortSignal.addEventListener("abort", onAbort, { once: true });
```
egonSchiele: **Abort listener is never removed when the session settles normally**, and `settle()` has no `removeEventListener` (there's none anywhere in this file). The signal from runBatch is a fresh `AbortSignal.any([parentSig, branch signal])` per invoke, but that doesn't save you: a composed signal whose parent source is still alive is pinned by its unremoved abort listener (verified empirically — 0/500 GC'd vs 500/500 once removed), so whenever a parent signal exists (any `run()` under fork/race or a TimeGuard), each execution segment leaks its `onAbort` closure — RunSession, ctx, ChildProcess — for the lifetime of the parent signal. Multi-cycle pause/resume multiplies this per segment.

Fix: stash the listener ref on the session and `removeEventListener` in `settle()`. (The stale-kill concern is a non-issue — `child.kill` on an exited child returns false harmlessly, and `settle` early-returns on `s.settled`.)

packages/agency-lang/lib/runtime/subprocess-bootstrap.ts
```diff
@@ -64,17 +66,71 @@ async function sendResultOrLimitError(msg: IpcResultMessage): Promise<void> {
...
+  if (typeof mod.respondToInterrupts !== "function") {
+    await sendOrDie({
+      type: "error",
+      error: `respondToInterrupts export not found in compiled module. The module may have been compiled with an incompatible version.`,
+    });
+    process.exit(1);
+  }
+  const interrupts = msg.interrupts.map((intr) => ({ ...intr, checkpoint: msg.checkpoint }));
+  ipcLog("send", { type: "log", detail: `resuming node ${msg.node} with ${msg.responses.length} responses` });
+  return mod.respondToInterrupts(interrupts, msg.responses);
```
egonSchiele, line 133: **Duplicate `interruptResolved` per pause/resume cycle in the shared trace.** The parent's `respondToInterrupts` already emits `interruptResolved{resolvedBy:"user"}` for each interrupt (interrupts.ts ~539), and this call re-enters the same function in the resumed child — same preserved `interruptId`, same inherited runId/traceId — emitting the event again. Meanwhile `interruptThrown` fired exactly once (only `renderVerdict` emits it, in the child). So one thrown event pairs with two terminal resolutions, and each nesting hop adds another (N+1).

The statelog docs frame thrown↔resolved as a pairable lifecycle; consumers doing that pairing will double-count. Suggest suppressing the user-resolution emission on the resume-replay side (e.g. skip when running under a resume instruction / when `subprocessRunInfo` marks this a resumed segment), so the parent's emission is the single terminal event.

packages/agency-lang/lib/runtime/interrupts.ts
```diff
@@ -339,6 +358,41 @@ export async function interruptWithHandlers<T = any>(
...
+  // statelog events correlate to the same interrupt.
+  if (isIpcMode()) {
+    const parentOutcome = await sendInterruptToParent(interruptObj, interruptId);
+    const merged = mergeChainOutcomes(local, parentOutcome);
+    // A propagated/noResponse merge renders as an Interrupt[] — the child
+    // pauses itself through its NORMAL propagate path (batching with
+    // concurrent siblings, stamping the shared checkpoint), and the
+    // bootstrap converts the final Interrupt[] into an `interrupted`
+    // terminal message for the parent to surface to the user.
+    return renderVerdict(merged, ctx, interruptId, interruptObj, "ipc");
```
egonSchiele: **`resolvedBy: "ipc"` misattributes locally-handled interrupts.** Under the new child-decides protocol, `mergeChainOutcomes` returns the child's own local approval when the parent replies `noResponse` — but this line logs every IPC-mode resolution as `"ipc"`. So an interrupt approved purely by the subprocess's own handler is attributed to the parent hop, while a local *reject* in the same process is emitted as `"handler"` (inside `runHandlerChain`, which returns before this branch) — inconsistent within one process. `resolvedBy` is a consumed field (`logsViewer/summary.ts`, `eval/extract.ts`).

Under the old protocol the constant label was accurate (the parent's reply *was* the verdict); now the tag is knowable from the merge — e.g. pass `"ipc"` only when the parent's outcome actually decided the merged verdict, else `"handler"`.

packages/agency-lang/lib/runtime/interrupts.ts
```diff
@@ -339,6 +358,41 @@ export async function interruptWithHandlers<T = any>(
...
+  effect: string,
+  message: string,
+  data: T,
+  origin: string,
+  ctx: RuntimeContext<any>,
+  stack?: StateStack,
+): Promise<Interrupt<T>[] | Approved | Rejected> {
+  const interruptObj = { effect, message, data, origin };
+  const interruptId = nanoid();
+  const local = await runHandlerChain(ctx, stack, interruptId, interruptObj);
```
egonSchiele: **This is `gatherChainOutcome` inlined.** Lines 371–384 repeat its exact four-step walk (runHandlerChain → local-reject short-circuit → sendInterruptToParent → mergeChainOutcomes) — and the two copies are the two halves of the same distributed chain: the origin interrupt goes through this copy while every relay hop goes through `gatherChainOutcome`, so any future change to consult/merge policy applied to one but not the other makes hops at depth ≥2 render different verdicts for the same interrupt. Three independent review angles flagged this.

Consolidation looks mechanical: `renderVerdict(await gatherChainOutcome(...), ..., isIpcMode() ? "ipc" : "handler")`, threading a resolved-locally flag through for the local-reject early return (needed only because `runHandlerChain` already emitted the statelog event — a naive peek-then-delegate would run handlers twice). While there: `gatherChainOutcome`'s `interruptId = nanoid()` default is dead — its only caller (`handleInterruptMessage`) always passes `msg.interruptId`.

egonSchiele (commented):
Reviewed with a multi-angle pass (8 finder angles, every candidate independently verified against the head commit; 4 candidates were refuted and dropped). 8 findings below as inline comments, most severe first. The top one (assignment-position `try` strands a paused subprocess) is the only one I'd consider blocking.

Minor items not worth their own thread:
- `tests/agency-js/subprocess-statelog-runid/test.js:30` uses `new Set(...)` — CLAUDE.md says "Use arrays instead of sets."
- `docs/dev/runBatch.md:4` still says "Three callers delegate to it" and lists only runForkAll/runRace/PromptRunner.parallel — the PR itself calls `_run` the fourth adopter, and that doc is the slice-rule audit list, so please add `_run` to it.
- `compiler/typecheck.ts:57`'s comment still points at `compileAndPersist`, which this PR deleted.

Verified-but-intentional (no action needed, noting for the record): `mergeChainOutcomes`' `outer.value ?? innerValue` genuinely diverges from the in-process chain (which overwrites unconditionally, so a parent approve-with-no-value can't clear a child's injected value) — but it's tested, documented in the design doc, and partly forced by JSON not distinguishing absent from undefined; the inline comment claiming it "matches how a single-process chain overwrites approvedValue" overclaims, though. The `collectSubprocessResponses` all-responses-required throw and wallClock counting consult time are both documented contracts.

packages/agency-lang/lib/runtime/result.ts
egonSchiele: Fixed in fd7ad459. Confirmed: tryExpression assignments fell through to the bare scopedAssign branch. The sync halt check is now extracted into a shared assignmentInterruptGuard helper used by both the functionCall and new tryExpression branches, so the two cannot drift. Regression test: tests/agency/interrupts/tryAssignmentInterrupt.agency (in-process interrupting callee under assignment-position try — the identical generated guard covers the subprocess case; verified failing before the fix, passing after).

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: Fixed in fd7ad459 with the watchdog you suggested: the bootstrap registers process.on(disconnect) before anything else and exits. Because every process in a nested tree is a bootstrap child, a mid-tier death disconnects its children in turn, so the subtree chain-reaps without process-group plumbing. Process groups would still be the complete answer for a child that deliberately detaches from the IPC channel, but nothing in the runtime does that today.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: Covered by the same disconnect watchdog (fd7ad459): a child parked in sendInterruptToParent exits when the channel drops rather than vanishing silently or hanging. I chose exit-on-disconnect over fabricating a rejected outcome — with the parent dead there is nobody to receive a terminal message, and continuing to execute (LLM spend) after a rejected verdict would be worse than stopping.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: Fixed in fd7ad459 — materializeCompiledScript now validates compiled.code and throws: CompiledProgram has no code; obtain it from compile() (or compileFile()). Values from older Agency versions carried a file path instead and must be recompiled.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: Fixed in fd7ad459 — the listener detach is stored as session.detachAbortListener and runs in settle(), which every terminal path funnels through (result, interrupted, error, limits, close, abort itself).

packages/agency-lang/lib/runtime/subprocess-bootstrap.ts
egonSchiele: Fixed in fd7ad459 — respondToInterrupts suppresses the resolvedBy=user emission when running in IPC mode, so in a nested tree only the root emits the single terminal resolution. The subprocess-statelog-runid E2E now asserts the child log contains zero resolvedBy=user events across the pause/resume cycle.

packages/agency-lang/lib/runtime/interrupts.ts
egonSchiele: Fixed in fd7ad459 — gatherChainOutcome now reports whether the parent outcome participated in the verdict (parentDecided = parent replied anything but noResponse), and interruptWithHandlers tags resolvedBy accordingly: handler for verdicts settled purely locally, ipc when the parent hop decided or contributed. Two unit tests pin both attributions with a mocked parent reply.

packages/agency-lang/lib/runtime/interrupts.ts
egonSchiele: Consolidated in fd7ad459 along the lines you sketched: interruptWithHandlers now delegates to gatherChainOutcome, which returns { outcome, parentDecided, localRejected } — localRejected carries the already-emitted-statelog-event fact through the early return (no double emission, handlers run once), and parentDecided feeds the resolvedBy fix in the sibling thread. The dead nanoid default is gone; interruptId is a required parameter.

packages/agency-lang/lib/stdlib/agency.compile.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  it("returns code text and writes nothing to disk", () => {
+    const tmpRoot = join(process.cwd(), ".agency-tmp");
+    const before = existsSync(tmpRoot) ? readdirSync(tmpRoot) : [];
+    const result = _compile("node main() { return 42 }");
+    const after = existsSync(tmpRoot) ? readdirSync(tmpRoot) : [];
+    expect(typeof result.moduleId).toBe("string");
+    expect(result.code).toContain("main"); // transpiled JS text
+    expect((result as any).path).toBeUndefined(); // no file reference
+    expect(after).toEqual(before); // no temp-dir writes at compile time
+  });
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2e659845 — the test now asserts no module-specific artifact exists under .agency-tmp instead of whole-directory before/after equality, so concurrent tests writing their own temp dirs cannot flake it.

egonSchiele:
All 8 findings addressed in fd7ad459 (+ 2e659845 for the compile-test flakiness) — replies with details on each thread. The minor items from the review body are also in:

- `subprocess-statelog-runid/test.js` Set usage → array-based dedupe.
- `docs/dev/runBatch.md` (and the cross-reference in `concurrent-interrupts.md`) now list `_run` as the fourth adopter, including in the slice-rule audit list.
- The stale `compileAndPersist` comment in `compiler/typecheck.ts` now points at `materializeCompiledScript`.
- The `mergeChainOutcomes` comment no longer claims value parity with the in-process chain — it now states the `??` divergence and the JSON absent-vs-undefined reason.

On the blocking finding: the fix is a shared `assignmentInterruptGuard` helper used by both the functionCall and tryExpression assignment branches (so they cannot drift), with an in-process regression test that exercises the identical generated guard the `try run(...)` case hits. Verified failing before, passing after.

Verification after the fixes: 1213 runtime/fixture unit tests, the full subprocess execution sweep, all 4 agency-js subprocess tests, structural lint, `make fixtures` (one fixture line: the post-merge `matchExpression` fixture catching up to the `maxDepth` descriptor).

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2964,39 +2998,35 @@ export class TypeScriptBuilder {
...
-              ]),
-            ),
-          );
-        }
+      } else {
+        const guard = this.assignmentInterruptGuard(varRef);
+        if (guard) stmts.push(guard);
       }
       return ts.statements(stmts);
+    } else if (value.type === "tryExpression") {
```
egonSchiele: **The guard misses binOp-wrapped `try` — `const r = try f() catch fallback` still strands the pause.** This branch fires only when the RHS is *exactly* a `tryExpression`, but `try f() catch v` (and `try f() || x`) parse as a `binOpExpression` whose `left` is the tryExpression, so they fall through to the unguarded `scopedAssign` else-branch. Verified against this head: the AST for `const r = try mightPause() catch "x"` is `binOpExpression(catch)`, `processCatchExpression` emits a bare `await __catchResult(...)` with no `hasInterrupts` check, and `__catchResult` passes a raw `Interrupt[]` through unchanged (it fails `resultValueSchema`) — so `r` gets the interrupt array and execution continues, exactly the stranded-pause bug this commit fixed for bare `try`.

This isn't an exotic form: `docs/site/guide/mcp.md` uses `const tools = mcp("filesystem") catch []` and five more like it. Adding `catch ...` to your own new `tryAssignmentInterrupt.agency` test would reproduce it.

The underlying issue is that guard emission dispatches on the RHS's *top-level* node type (now two special cases: `functionCall`, `tryExpression`). A predicate like "RHS contains an interruptible call" feeding the one `assignmentInterruptGuard` emitter would cover `catch`, `||`, and the next syntactic form by construction — or at minimum, add a `binOpExpression`-containing-try case plus a catch-form regression test.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -541,34 +748,56 @@ function attachStdoutForwarder(
...
+    // The child's interruptId is relayed into the chain walk so this
+    // process's statelog events correlate with the originating interrupt.
+    // (gatherChainOutcome itself recurses to OUR parent when this process
+    // is also a subprocess — that is what makes nesting compose.)
+    //
+    // Span isolation: concurrent child interrupts each run their own
+    // handler chain here; without a branch-local span stack their
+    // handlerChain span pushes/pops would interleave on the shared stack
+    // (same discipline runBatch applies to its children).
+    const { outcome } = await s.ctx.statelogClient.runInBranchContext(
```
egonSchiele, line 752: **The relay hop discards `localRejected`, so a parent-side handler reject still double-emits `interruptResolved`.** When a parent handler rejects a child's relayed interrupt: `gatherChainOutcome` passes the child's `interruptId` into `runHandlerChain`, whose reject branch emits `interruptResolved{outcome:"rejected", resolvedBy:"handler"}` here in the parent (interrupts.ts:240); the rejected outcome is then replied to the child, where `localRejected` is false, so the `interrupts.ts:402` suppression doesn't fire and `renderVerdict` emits a second `interruptResolved{same id, resolvedBy:"ipc"}` — both into the same inherited trace. This is precisely the duplicate-terminal-event class the rest of this commit suppresses (origin-local rejects, user-resolution replays); rejects at relay hops are the one remaining path (one extra event per rejecting ancestor in nested trees). Approvals are fine — `runHandlerChain`'s approve branch emits only `handlerDecision`.

A deeper fix would also delete the flag: make the origin's `renderVerdict` the sole emitter of terminal `interruptResolved` events and have `runHandlerChain` emit only `handlerDecision` on reject — relay hops then emit nothing by construction, and `localRejected` (which is anyway derivable as `outcome.kind === "rejected" && !parentDecided`) disappears.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2889,6 +2889,40 @@ export class TypeScriptBuilder {
     return result;
   }
 
+  /** The interrupt guard emitted after a sync assignment whose RHS may
+   * bubble an Interrupt[]: halt the scope (awaitAll first) so the batch
+   * propagates to the caller, or — inside a handler body — throw, since
+   * handlers must not raise interrupts. Returns null at global scope,
+   * where there is no runner to halt. */
+  private assignmentInterruptGuard(varRef: TsNode): TsNode | null {
```
egonSchiele, line 2888: Nice extraction — but there's a third, byte-similar copy of this guard still inlined in `processFunctionCallAsStatement`'s sync path (~lines 2138–2178): same handler-body throw (same message), same node-vs-function `haltValue` shape, same `awaitAll` → `runner.halt` → `return` sequence, same global-scope skip (via the outer `scope.type !== "global"` gate). Since it already binds the result to a temp const, it could keep the `constDecl` and outer `shouldHandleInterrupts` gating and just append `this.assignmentInterruptGuard(ts.id(tempVar))` — the tryExpression site above already demonstrates the assign-then-guard pattern. Otherwise the next halt-semantics change (like the `awaitAll` this guard once gained) gets made in the helper and silently missed for statement-position calls.

packages/agency-lang/lib/runtime/subprocess-bootstrap.ts
```diff
@@ -11,19 +11,34 @@
...
 
+// Parent-death watchdog. If the parent process dies (SIGKILL on a limit,
+// crash, anything), our IPC channel disconnects — exit immediately instead
+// of running on outside every limit. This also reaps nested trees: when a
+// mid-tier process dies this way, its own children's channels disconnect
+// in turn, so the whole subtree tears itself down without process-group
+// plumbing. It equally covers a child parked in sendInterruptToParent
+// awaiting a decision that will never come. Registered before anything
+// else so no window exists where the parent can die unobserved.
+process.on("disconnect", () => {
```
egonSchiele, line 30: This watchdog makes `sendLockAcquireToParent`'s per-call disconnect handling (ipc.ts ~441) dead code in every real child: this listener registers at module load — before any lock call — Node fires same-event listeners in registration order, and `process.exit(1)` here prevents the later-registered `onDisconnect` from ever running (verified with a scratch script). So the `"IPC channel closed while waiting for lock"` rejection path is unreachable in bootstrapped children, and a maintainer debugging a lock hang will look for that error and never see it. No test relies on it either (the in-process unit tests never emit `disconnect`). Suggest deleting that handler and its `removeListener` bookkeeping — this watchdog is the single disconnect authority, which is exactly how `sendInterruptToParent` already treats it per the comment above — or documenting at that site why it stays.

egonSchiele (commented):
Re-reviewed the fix commits (`fd7ad459`, `2e659845`) with the same multi-angle + adversarial-verify process. First: all 8 previous findings were verified as genuinely addressed — the abort-listener detach covers every settle path, the disconnect watchdog correctly tears down nested trees on normal shutdown orderings, `compiled.code` validation sits before any fork side-effect, the resolvedBy attribution is right in all four merge cases, the `new Set` test violation is fixed, and the consolidation onto `gatherChainOutcome` preserves behavior. Nice turnaround.

Four new findings as inline comments (most severe first), plus one that has no diff line to anchor on:

**CI breakage — stale fixture `tests/typescriptGenerator/matchBlockBlockArms.mjs:142`.** It's the only `.mjs` fixture at this head whose `_run` descriptor still has the old 9 params (ends at `cwd`); the imports template now emits 10 (`maxDepth` — compare the regenerated `matchExpression.mjs:142`). The fixture landed on main in `7eefd7c1` after your `make fixtures` run, so the byte-comparison integration test fails at this head. One `make fixtures` should fix it.

Minor notes, no action required:
- `mergeChainOutcomes`' `outer.value ?? innerValue`: the documented rationale (JSON can't carry `undefined`) doesn't cover an explicit `null` — a parent `approve(null)` meaning "clear the child's injected value" gets swallowed and the child's value wins. Fine to leave, but worth a comment line if `null` is meant to be a real approve value.
- The disconnect watchdog is cooperative: a grandchild stuck in synchronous work (tight loop) never processes the `disconnect` event, and its wall-clock enforcement died with its parent. Only process groups close that completely — worth a sentence in `subprocess-ipc.md`'s limitations.
- The user-resolution suppression keys on `isIpcMode()` ("a subprocess never emits user resolutions"). Correct today because the bootstrap resume replay is the only in-subprocess caller of `respondToInterrupts`; if an in-subprocess responder ever becomes legitimate (e.g. agency code answering its own nested child's interrupts), its terminal events will be silently swallowed. An explicit `replayed` flag from `executeResume` would encode the rule where the replay actually happens.
- Two small duplications from round 1 still stand (not previously posted inline, so just noting): the identical `try { child.kill("SIGKILL") } catch { ipcLog(kill_failed) }` block in `settleWithLimitFailure` and `onAbort`, and the hand-rolled `.agency-tmp` dir construction in both `materializeCompiledScript` and `typecheck.ts` `withSourcePath`.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Fixed in bcae97ab with the predicate approach: rhsMayBubbleInterrupts walks binOp trees for a tryExpression anywhere inside, so catch, ||, and future wrappers hit the one assignmentInterruptGuard emitter by construction. Reproduced your catch case first — with one wrinkle worth noting: a test that does a bare return r self-heals, because returning the raw Interrupt[] re-bubbles it at node level; the regression test now consumes the value in a string template so stranding is actually detectable.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Fixed in bcae97ab — processFunctionCallAsStatement sync path (both the handler-body and normal variants) now binds the temp and appends this.assignmentInterruptGuard(ts.id(tempVar)); the third copy is gone. Fixture regeneration confirmed byte-identical generated output for statement-position calls.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: Took the deeper fix in bcae97ab: runHandlerChain emits only handlerDecision on reject, the origin renderVerdict is the sole terminal interruptResolved emitter, and localRejected is deleted (gatherChainOutcome returns { outcome, parentDecided }). Relay hops now emit nothing terminal by construction. Two unit tests pin it: a local reject in IPC mode emits exactly one terminal event (resolvedBy handler, parent never consulted), and a relay-hop reject via gatherChainOutcome emits zero.

packages/agency-lang/lib/runtime/subprocess-bootstrap.ts
egonSchiele: Deleted in bcae97ab — the onDisconnect/fail/cleanup bookkeeping is gone, with a comment at the send site naming the bootstrap watchdog as the single disconnect authority (same contract sendInterruptToParent documents). Verified the lock paths still pass (lock-cross-process, nested-lock-relay).

PR #399, opened 2026-07-02
Match-aware definite-return checking + while(true) divergence

egonSchiele:
## Summary

Definite-return checking now covers **match-containing functions** (removing the #386 `containsMatch` safe-subset skip) and treats **`while (true)` with no reachable `break` as diverging** (remaining-work §5b). Config default stays `warn`.

### Why removing the skip is sound now

The hard problem #386 deferred — "does a match-ending function return on every path?" depends on exhaustiveness, which the flow graph can't see — was **dissolved by match expressions (7eefd7c1)**. Since `return` in an arm yields to the match instead of returning from the function:

- A **statement-position** arm `return` is a compile error → statement matches are effect-only and can never make the function return. The flow graph threading through them without `exit` is now *exact*, not conservative.
- An **expression-arm** `return` lowers to a `matchYield` (passThrough to the flow builder) → can never fake a function `exit`.
- `return match(...)` lowers to the match region **plus a real trailing `returnStatement`** → genuine `exit`.
- `const x = match(...)` lowers to the region plus a plain assignment → genuine non-exit.

So the flow terminal is exact for every match shape and the fix is simply deleting the skip — no exhaustiveness oracle, no reachability walk, no lowering metadata. The old false-positive driver (the idiomatic arm-returning Result match as a function tail) is now a compile error whose fixit (`return match(...)`) is exactly the form the checker handles correctly.

### while(true) divergence

The flow builder's `whileLoop` rule yields `{kind: "exit"}` for a literal-`true` loop with no `break` that can exit *this* loop (`hasReachableBreak`: break binds to the nearest loop; nested loops/definitions excluded; anything else — e.g. a break inside a `handle` body — conservatively counts as reachable). Also makes post-infinite-loop statements dead code to all flow consumers.

### Measurement

- Blast-radius sweep at `definiteReturns: "error"` over **855** execution-test programs (`tests/agency`, `tests/agency-js`, `tests/integration`, including the new match-expression suite): **0 hits, 0 crashes**.
- Full unit suite at exact parity with base `main`: 6980 passed. Two `memberPathNarrowing` fixtures were statement matches with expression arms inside `def f(): number` — genuine fall-throughs under the new semantics — modernized to `return match(...)` (narrowing shape unchanged).
- Known degenerate limitation (documented, not engineered around): a statement match whose every arm ends in `while (true)` is still flagged, because the `matchBlock` flow rule discards arm end-flows.

### Handler safety

Type-checker only: zero codegen/runtime/handler-registration impact.

### Note: pre-existing failures on main

Base `main` (01c7b02c) currently has **21 failing unit tests** unrelated to this change (14 `effectPayloadCheck`, 2 `compile`, 2 `typescriptGenerator.integration`, 2 `effectDeclCollection`, 1 `handlerBodyInterrupts` — e.g. `Effect 'std::read' data field 'filename' is missing`). All 21 reproduce identically on a clean checkout of the base commit; this branch adds none.

### Docs

- `docs/misc/config.md`: `definiteReturns` bullet rewritten for the new semantics (match coverage, `while (true)`, how to promote to `error`).
- `docs/dev/typechecker/definite-returns-remaining-work.md` (new): the shipped state, the history of how the match problem dissolved, and what remains (§ flip warn→error — last sweep 855/0; § trailing-raise refinement; § dead-code detection).

### DIFF
--- a/packages/agency-lang/lib/typeChecker/flowBuilder.ts
+++ b/packages/agency-lang/lib/typeChecker/flowBuilder.ts
@@ -114,6 +114,36 @@ export function assignedNames(body: AgencyNode[]): string[] {
   return names;
 }
 
+/** The literal condition `true` — no constant folding (`1 == 1` doesn't count). */
+function isLiteralTrue(cond: AgencyNode): boolean {
+  return cond.type === "boolean" && cond.value === true;
+}
+
+/** A `break` that can exit the loop whose body this is: found anywhere in the
+ *  body except inside a nested loop (break binds to the nearest loop) or a
+ *  nested function/node definition (same ancestor filter as assignedNames).
+ *  Conservative toward the loop being escapable: a break inside e.g. a handle
+ *  block still counts as reachable. Traversal note (verified): walkNodes
+ *  descends into every statement-bearing construct that could carry a
+ *  loop-bound break — ifElse, loops, handleBlock body + inline handler body,
+ *  thread/parallel/seq blocks, match arm bodies, withModifier — the only
+ *  non-descended wrapper is staticStatement, which cannot meaningfully
+ *  contain a break. */
+function hasReachableBreak(body: AgencyNode[]): boolean {
+  for (const { node, ancestors } of walkNodes(body)) {
+    if (node.type !== "keyword" || node.value !== "break") continue;
+    const bindsElsewhere = ancestors.some(
+      (a) =>
+        a.type === "whileLoop" ||
+        a.type === "forLoop" ||
+        a.type === "function" ||
+        a.type === "graphNode",
+    );
+    if (!bindsElsewhere) return true;
+  }
+  return false;
+}
+
 /** statement kind → its flow transformation. The "what". */
 type StatementRuleTable = {
   [K in AgencyNode["type"]]?: (
@@ -185,6 +215,19 @@ const statementRules: StatementRuleTable = {
     attachExpressionsToFlow(node.condition as AgencyNode, flow, env);
     const facts = analyzeCondition(node.condition);
     const bodyEnd = buildFlowGraph(node.body, wrapFacts(flow, facts.then), env);
+    // `while (true)` with no break that can exit THIS loop never falls
+    // through: the post-loop flow is unreachable, so the loop diverges like a
+    // `return` (definite-return §5b; also makes trailing statements dead code
+    // to downstream consumers). The body was still built above so its refs
+    // keep their flowOf entries; skipping widenAtLoopBackEdge here is
+    // deliberate — widening only shapes the POST-loop flow, which does not
+    // exist for a diverging loop (do not "restore" it). Scope note: only the
+    // literal-true condition counts; `while (cond) { raise ... }` is out of
+    // scope (raise is passThrough / may resume — same convention as
+    // alwaysExits).
+    if (isLiteralTrue(node.condition as AgencyNode) && !hasReachableBreak(node.body)) {
+      return { kind: "exit" };
+    }
     return wrapFacts(
       widenAtLoopBackEdge(flow, bodyEnd, assignedNames(node.body), env),
       facts.else,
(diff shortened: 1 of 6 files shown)


packages/agency-lang/lib/typeChecker/flowBuilder.ts
```diff
@@ -114,6 +114,36 @@ export function assignedNames(body: AgencyNode[]): string[] {
...
+ *  block still counts as reachable. Traversal note (verified): walkNodes
+ *  descends into every statement-bearing construct that could carry a
+ *  loop-bound break — ifElse, loops, handleBlock body + inline handler body,
+ *  thread/parallel/seq blocks, match arm bodies, withModifier — the only
+ *  non-descended wrapper is staticStatement, which cannot meaningfully
+ *  contain a break. */
+function hasReachableBreak(body: AgencyNode[]): boolean {
+  for (const { node, ancestors } of walkNodes(body)) {
+    if (node.type !== "keyword" || node.value !== "break") continue;
+    const bindsElsewhere = ancestors.some(
```
egonSchiele, line 153: Are there any other node types to consider here?

packages/agency-lang/lib/typeChecker/flowBuilder.ts
```diff
@@ -114,6 +114,36 @@ export function assignedNames(body: AgencyNode[]): string[] {
...
+      (a) =>
+        a.type === "whileLoop" ||
+        a.type === "forLoop" ||
+        a.type === "function" ||
+        a.type === "graphNode",
+    );
+    if (!bindsElsewhere) return true;
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
egonSchiele: Verified and fixed in fc7a64c5. Confirmed against `processKeyword` (typescriptBuilder.ts:604-619): `insideHandlerBody` emits a bare `break`, and inline handler bodies compile to separate arrow functions, so a break there can never reach the enclosing Agency loop. `hasReachableBreak` now collects breaks under inline handler bodies (by node identity) and skips them, while still descending into the guarded `handle { ... }` body — a break there compiles to `runner.breakLoop()` and genuinely escapes. Both directions are pinned by tests ("counts a break inside a guarded handle body as reachable" / "a break inside an inline HANDLER body cannot escape the loop").

packages/agency-lang/lib/typeChecker/flowBuilder.ts
egonSchiele: Good question — I went through the node kinds with codegen in hand (fc7a64c5 folds in the one real gap). The full accounting:

**Excluded (break cannot bind to this loop):**
- `whileLoop`/`forLoop` — break binds to the nearest loop.
- `function`/`graphNode` — nested definitions are separate scopes (same filter `assignedNames` uses).
- Inline handler bodies (`with (e) { ... }`) — NEW in fc7a64c5, from the Copilot thread: handler bodies codegen as separate arrow functions where `processKeyword` emits a bare `break` (typescriptBuilder.ts:608-609), so it can never reach the enclosing Agency loop.

**Deliberately counted (conservative toward "escapable"):**
- The guarded `handle { ... }` body, `thread`/`parallel`/`seq` blocks, match arm bodies, `with`-modifier statements, and callback blocks — inside a stepped loop, `break` in these compiles to `runner.breakLoop()` + return, which sets the `_break` flag the loop's `shouldSkip()` honors, so it genuinely breaks the loop when it runs during an iteration. Where the runtime story is murkier (e.g. a break inside a parallel branch), counting stays sound for this check: the worst case is we fail to prove divergence and emit a warn-level definite-return diagnostic, never a missed fall-through.

**Non-issues:** `staticStatement` isn't descended by walkNodes but can't meaningfully carry a break; `continue` never exits the loop; `goto` is node-only and transfers out of the body entirely (post-loop statements stay dead either way), and nodes are exempt from the diagnostic.

PR #401, opened 2026-07-02
Fix for-loop record iteration + allow match expressions in handler bodies

egonSchiele:
Two independent fixes, one commit each.

## 1. For-loop over records binds the value, not the index

`for (key, value in record)` bound the second variable to the numeric iteration index instead of the value at that key, contradicting the docs. `for (val, i in array)` still binds the index. Both are now supported.

- **Runtime (`Runner.loop`)** — pass the value (`items[key]`) as the second callback argument for records/objects; keep the index (`i`) for arrays.
- **Type checker** — type the second loop variable as the value type for `Record<K, V>` and object literals, and as `number` (index) for arrays. Also accept structural object-literal types (`objectType`) as iterables; they were previously rejected because only `Record` generics were recognized, so iterating a plain `{ a: 1, b: 2 }` errored with *"For-loop iterable must be an array or Record"*.

Tests: runner unit tests (record → `[key, value]`, array → `[item, index]`), type-checker tests (value typing + object-literal iterable), and an execution test.

## 2. Match expressions inside handler bodies

`return match(...)` / `const x = match(...)` were rejected inside `with` handler bodies with *"match expressions are not supported inside handler bodies."*

**Why they were rejected:** stepped code unwinds a match arm's value via `runner.exitMatch` plus the owning `runner.ifElse` clearing the `_matchExit` flag. Handler bodies compile to plain JS with no owning `runner.ifElse` to clear it, so the flag would leak out of the handler and silently skip the rest of the enclosing node.

**The fix:** handler bodies now compile an expression-position match to a self-contained async IIFE whose arms `return` their value, storing the result in the `__matchval_<id>` frame local the consumer reads:

```js
__stack.locals.__matchval_1 = await (async () => {
  if (data.effect === "std::run") { return await approve(); }
  return await reject();
})();
return __stack.locals.__matchval_1;
```

This never touches `_matchExit`. Lowering is unchanged (only TS emission diverges), so the type checker is unaffected. Covers both the literal-arm (`matchBlock`) and pattern-arm (`ifElse`) lowered forms.

Tests: execution tests for return-position, assignment-position, an object-pattern arm, and a multi-path block arm (proving the IIFE's `return` short-circuits — a naive plain-assignment lowering would fall through and overwrite the value). Obsolete "is rejected" lowering tests were rewritten to assert correct lowering.

## Verification

- New execution tests pass (`forLoopRecordValue`, `match-expr-in-handler`).
- Unit suites green: `lib/lowering/`, `lib/runtime/runner.test.ts`, `lib/typeChecker.test.ts`.
- Structural linter clean.
- Existing handler/interrupt tests unaffected (statement matches and plain `if/else` in handlers still route through plain mode unchanged — the divert only applies to expression-position matches).

Note: two pre-existing fixture failures (`matchExpression`, `matchBlockBlockArms`) exist on `main` independent of this change — they fail on a stale `console.error` epilogue and are node-body matches on the stepped path this PR does not touch.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -233,6 +233,11 @@ export class TypeScriptBuilder {
   // checks. This counter gives each one a unique name.
   private handlerFuncResultCounter: number = 0;
   private insideGlobalInit: boolean = false;
+  // Set while emitting the body of a plain-mode (handler) match EXPRESSION,
+  // whose arms are wrapped in an async IIFE. Inside it, a `matchYield` compiles
+  // to a real `return <value>` (exiting the IIFE) instead of the stepped
+  // `runner.exitMatch`, so the handler never touches the `_matchExit` flag.
+  private insidePlainMatchExpr: boolean = false;
 
   /*
   We break up every function and node body into steps,
@@ -526,7 +531,9 @@ export class TypeScriptBuilder {
         return ts.empty();
       case "matchBlock":
         return this.insideHandlerBody
-          ? this.processBlockPlain(node)
+          ? node.matchExprId !== undefined
+            ? this.processMatchExpressionPlain(node)
+            : this.processBlockPlain(node)
           : this.processMatchBlockWithSteps(node);
       case "number":
       case "unitLiteral":
@@ -564,7 +571,9 @@ export class TypeScriptBuilder {
           : this.processWhileLoopWithSteps(node);
       case "ifElse":
         return this.insideHandlerBody
-          ? this.processBlockPlain(node)
+          ? node.matchExprId !== undefined
+            ? this.processMatchExpressionPlain(node)
+            : this.processBlockPlain(node)
           : this.processIfElseWithSteps(node);
       case "newLine":
         return ts.empty();
@@ -1360,7 +1369,43 @@ export class TypeScriptBuilder {
       );
     }
     const value = node.value ? this.processNode(node.value) : ts.id("undefined");
+    // Plain-mode (handler) match expressions wrap their arms in an async IIFE
+    // (see processMatchExpressionPlain): a yield is a real `return` out of that
+    // IIFE, not the stepped `runner.exitMatch` unwind — so the handler never
+    // sets the `_matchExit` flag that it has no `runner.ifElse` to clear.
+    if (this.insidePlainMatchExpr) {
+      return ts.return(value);
+    }
     return ts.runnerExitMatch({ matchId: node.matchId, value });
+  }
+
+  /**
+   * Compile an expression-position `match` inside a handler body (plain mode).
+   *
+   * Stepped code unwinds a match arm's value via `runner.exitMatch` + the owning
+   * `runner.ifElse` clearing `_matchExit`. A handler body compiles to plain JS
+   * with no owning `runner.ifElse`, so that protocol would leak the flag and
+   * silently skip the rest of the enclosing node. Instead we emit the arm
+   * dispatch as an async IIFE whose arms `return` their value, and store the
+   * result in the `__matchval_<id>` frame local the consumer reads:
+   *
+   *   __stack.locals.__matchval_<id> = await (async () => { <if-chain> })();
+   *
+   * `node` is either the pass-through `matchBlock` (literal arms) or the lowered
+   * `ifElse` chain (pattern arms) — both carry `matchExprId`. `processBlockPlain`
+   * builds the plain if/else for either; `insidePlainMatchExpr` makes the arms'
+   * `matchYield` nodes emit real returns out of the IIFE.
+   */
+  private processMatchExpressionPlain(node: MatchBlock | IfElse): TsNode {
+    const matchId = node.matchExprId!;
+    const prev = this.insidePlainMatchExpr;
+    this.insidePlainMatchExpr = true;
+    const ifChain = this.processBlockPlain(node);
+    this.insidePlainMatchExpr = prev;
+    return ts.assign(
+      ts.scopedVar(`__matchval_${matchId}`, "local", this.moduleId),
+      ts.await(ts.iife({ async: true, body: [ifChain] })),
+    );
   }
 
   private processImportStatement(node: ImportStatement): TsNode {
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -74,6 +74,7 @@ import {
 } from "../types/importStatement.js";
 import { MatchBlock, MatchBlockCase } from "../types/matchBlock.js";
 import { MatchYield } from "../types/matchYield.js";
+import { matchValName, isMatchValName } from "../matchVal.js";
 import { ReturnStatement } from "../types/returnStatement.js";
 import { GotoStatement } from "../types/gotoStatement.js";
 import { WhileLoop } from "../types/whileLoop.js";
@@ -233,10 +234,26 @@ export class TypeScriptBuilder {
   // checks. This counter gives each one a unique name.
   private handlerFuncResultCounter: number = 0;
   private insideGlobalInit: boolean = false;
+  // Gives each plain-mode (handler) for-loop unique names for its normalization
+  // temps, so nested loops in the same block don't collide.
+  private plainForLoopCounter: number = 0;
   // Set while emitting the body of a plain-mode (handler) match EXPRESSION,
   // whose arms are wrapped in an async IIFE. Inside it, a `matchYield` compiles
   // to a real `return <value>` (exiting the IIFE) instead of the stepped
   // `runner.exitMatch`, so the handler never touches the `_matchExit` flag.
+  //
+  // INVARIANT: this must be true ONLY where a `return` lands directly in the
+  // IIFE. Today that holds because (a) `matchYield` nodes exist only inside
+  // expression-match regions, every one of which routes through
+  // `processMatchExpressionPlain` in a handler, and (b) the builder never
+  // switches OUT of plain mode within the IIFE — it does not reset
+  // `insideHandlerBody` at block-argument / nested-frame boundaries, so nothing
+  // stepped compiles inside a handler. At every yield site this flag therefore
+  // currently equals `insideHandlerBody`. If a future change starts compiling
+  // stepped constructs inside a handler (e.g. resetting `insideHandlerBody` per
+  // frame), reset THIS flag at those same boundaries too — otherwise a stepped
+  // match nested in a plain arm would emit a bare `return` inside a step
+  // callback (value discarded, `_matchExit` never set: a silent wrong answer).
   private insidePlainMatchExpr: boolean = false;
 
   /*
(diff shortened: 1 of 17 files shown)


packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -854,7 +857,12 @@ export class Runner {
       this._continue = false;
       this.path.push(id);
       try {
-        await this.runInScope(() => callback(iterable[i], i, this));
+        // `for (item, second in x)`: for arrays the second variable is the
+        // numeric index; for records it is the value at the current key. The
+        // first callback arg is the element (array) or the key (record).
+        const item = iterable[i];
+        const second = isRecord ? (items as Record<string, any>)[item] : i;
```
egonSchiele, line 865: **Record iteration is fixed here (stepped mode) and in the type checker, but not in plain-mode codegen — so the feature silently no-ops inside handler bodies.**

For-loop semantics live in three places: `Runner.loop` (stepped), the type checker's `forLoop` case, and `processBlockPlain`'s for-loop emission in `typescriptBuilder.ts` (~3379–3408), which `with` handler bodies compile through. This PR updates the first two but not the third. A handler body containing `for (name, score in scores)` over a record compiles to:

```js
const __iter_name = __stack.locals.scores;
for (let score = 0; score < __iter_name.length; score++) {
  const name = __iter_name[score];
```

`record.length` is `undefined`, so the loop body executes **zero times, silently** — while the type checker now explicitly types `score` as the record's value type. A handler is arguably the most natural place to loop over interrupt data, so this combination (well-typed, does nothing, no error) is worth fixing in the same PR. The one-variable form `for (k in record)` in a handler emits `for...of` over a plain object and throws at runtime — pre-existing, same root cause.

Suggest desugaring plain-mode for-loops through the same array-vs-record normalization used here (e.g. iterate `Object.keys` when the iterable isn't an array), so all three implementations share one definition of record iteration.

packages/agency-lang/lib/typeChecker/scopes.ts
```diff
@@ -380,26 +381,52 @@ export function walkScopeBody(
...
           iterableType.type === "genericType" &&
           iterableType.name === "Record"
         ) {
-          // for (k in record): iteration variable is the key type.
-          scope.declare(node.itemVar as string, iterableType.typeArgs[0]);
+          // for (k, v in record): key is Record's key type, value is its
+          // value type.
+          itemType = iterableType.typeArgs[0];
+          secondType = iterableType.typeArgs[1];
+        } else if (iterableType.type === "objectType") {
```
egonSchiele: This branch fixes objectType iteration for *this* consumer, but the other Record-special-casing consumers still disagree about what an object literal's value type is: index access in `synthesizer.ts` (~923–931) and path resolution in `flow.ts` (~50–58) both return `any` for an `objectType` receiver. So inside the very loop this types, `for (k, v in obj) { obj[k] }` gives `v` a precise union but `obj[k]` is `any` — the checker's answer depends on which syntax you use.

Not necessarily a blocker for this PR, but a small shared helper (e.g. `recordLikeKeyValue(t): {key, value} | undefined` next to the existing objectType→Record assignability logic in `assignability.ts`) used by all three sites would make the consumers agree by construction, and would prevent the next objectType branch from picking its own value-type rule (forgetting `unionTypes`, or the empty-object → `any` case).

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -233,6 +233,11 @@ export class TypeScriptBuilder {
   // checks. This counter gives each one a unique name.
   private handlerFuncResultCounter: number = 0;
   private insideGlobalInit: boolean = false;
+  // Set while emitting the body of a plain-mode (handler) match EXPRESSION,
+  // whose arms are wrapped in an async IIFE. Inside it, a `matchYield` compiles
+  // to a real `return <value>` (exiting the IIFE) instead of the stepped
+  // `runner.exitMatch`, so the handler never touches the `_matchExit` flag.
+  private insidePlainMatchExpr: boolean = false;
```
egonSchiele, line 257: Two related observations on this flag:

1. **At its only read site (the `matchYield` case), `insidePlainMatchExpr === insideHandlerBody`.** `matchYield` nodes are only created by expression-match lowering, and every `matchExprId`-tagged region inside a handler routes through `processMatchExpressionPlain`, so in plain mode a `matchYield` is always lexically inside the IIFE. You could delete the field and the save/restore and just check `insideHandlerBody` — or keep it as the more defensive/structural encoding, but then the comment should say it currently always equals `insideHandlerBody` at yield sites.

2. **If kept, its correctness rests on an undocumented invariant:** the builder never resets `insideHandlerBody` (or this flag) at block-argument / nested-frame boundaries (`processBlockArgument`, `processBlockAsExpression`), so nothing stepped can compile within the IIFE's dynamic extent. The lowering pass this PR deleted *did* model those as new frames (`lowerNewFrameBody`); the builder never mirrored that, and now nothing documents it. If someone later adds the natural frame-boundary reset for `insideHandlerBody`, a stepped match inside a plain match arm would see this flag still set and emit a bare `return` inside a step callback — value discarded, `_matchExit` never set, silent wrong answers. Worth resetting this flag wherever a new compilation frame begins, or noting the invariant in the field comment.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -526,7 +531,9 @@ export class TypeScriptBuilder {
         return ts.empty();
       case "matchBlock":
         return this.insideHandlerBody
-          ? this.processBlockPlain(node)
+          ? node.matchExprId !== undefined
+            ? this.processMatchExpressionPlain(node)
```
egonSchiele: This nested ternary is duplicated verbatim in the `ifElse` case below (the two lowered shapes of a match expression: literal arms → `matchBlock`, pattern arms → `ifElse`). If the plain-mode routing rule ever changes — or a future match feature lowers to a third shape — one site can be updated and the other silently keeps the old behavior, and a miss doesn't error: the region compiles as a plain statement match and the consumer reads a never-assigned `__matchval_N` as `undefined`.

Suggest extracting one helper, e.g.:

```ts
private processPlainBranching(node: MatchBlock | IfElse): TsNode {
  return node.matchExprId !== undefined
    ? this.processMatchExpressionPlain(node)
    : this.processBlockPlain(node);
}
```

so both cases become `this.insideHandlerBody ? this.processPlainBranching(node) : this.processXWithSteps(node)`.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -1360,9 +1369,45 @@ export class TypeScriptBuilder {
...
+   * `matchYield` nodes emit real returns out of the IIFE.
+   */
+  private processMatchExpressionPlain(node: MatchBlock | IfElse): TsNode {
+    const matchId = node.matchExprId!;
+    const prev = this.insidePlainMatchExpr;
+    this.insidePlainMatchExpr = true;
+    const ifChain = this.processBlockPlain(node);
+    this.insidePlainMatchExpr = prev;
+    return ts.assign(
+      ts.scopedVar(`__matchval_${matchId}`, "local", this.moduleId),
```
egonSchiele: Nit (partly pre-existing debt): this is now the fourth hand-spelled site of the `__matchval_<id>` naming contract — the runner writes `frame.locals[`__matchval_${matchId}`]` (runner.ts:262), lowering builds the varRef (patternLowering.ts:427), and three regex consumers parse the prefix (this file ~928, checker.ts ~800, synthesizer.ts ~253). A typo at any one site produces a silently-`undefined` match value rather than an error. Since this PR adds a producer, it might be the moment to extract a shared `matchValName(id)` helper instead of deepening the duplication.

egonSchiele (commented):
Reviewed both fixes. The handler-body match IIFE design is sound (the short-circuit test coverage is nice), and the lowering simplification checks out — no leftover references to the deleted flags. One significant gap in the for-loop fix, plus a few smaller robustness/cleanup notes, all inline.

**Main concern:** the record key→value fix landed in the stepped runtime and the type checker, but not in the plain-mode codegen that handler bodies use — so the newly-blessed `for (key, value in record)` silently no-ops inside a `with` handler. Details on the `runner.ts` comment.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -1360,9 +1369,45 @@ export class TypeScriptBuilder {
...
+      ts.await(ts.iife({ async: true, body: [ifChain] })),
+    );
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed — `processMatchExpressionPlain` now restores the flag in a `finally`, so an emission error inside an arm can't leak it into a later sibling `matchYield`. (d93bbeaa)

packages/agency-lang/lib/runtime/runner.ts
egonSchiele: Fixed. Plain-mode for-loops now go through `processPlainForLoop`, which applies the same normalization as `Runner.loop`: `Array.isArray` → iterate the array (element, index); otherwise `Object.keys` → iterate a record (key, value); non-objects → empty. So `for (k, v in record)` binds the value, `for (k in record)` iterates keys, and arrays still bind the index — all three implementations now agree. Added `tests/agency/for-loop-in-handler.agency` covering the two-var record, one-var record, and array forms in a handler body. (d93bbeaa)

packages/agency-lang/lib/typeChecker/scopes.ts
egonSchiele: Addressed. Added `recordLikeKeyValue` in `lib/typeChecker/recordLike.ts` (maps both `Record<K, V>` and `objectType` to `{key, value}`) and used it in both this for-loop typer and index-access synthesis in `synthesizer.ts` — so `obj[k]` now yields the property value type instead of `any`, and the two agree by construction. Left `flow.ts` narrowing out of scope as a separate concern. Ran the full type-checker suite: no new failures beyond the pre-existing `effectPayloadCheck`/`handlerBodyInterrupts` ones. Added an index-access test asserting the tightened type. (d93bbeaa)

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Kept the flag rather than folding it into `insideHandlerBody`: `matchYield → return` is only correct inside the IIFE, and a distinct flag encodes that directly instead of leaning on the (currently-true) equivalence. Documented the invariant in the field comment — that it currently equals `insideHandlerBody` at yield sites, and that if stepped constructs ever start compiling inside a handler, this flag must be reset at those same frame boundaries or a nested stepped match would emit a bare `return` in a step callback. Also added the try/finally from the Copilot note so it can't leak. (d93bbeaa)

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Fixed — extracted `processPlainBranching(node)`; both the `matchBlock` and `ifElse` dispatch cases now call it, so the plain-mode routing rule lives in one place. (d93bbeaa)

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Done. Centralized the contract in `lib/matchVal.ts`: `matchValName(id)` (the runtime writer, lowering's varRef, and the new plain-mode producer), `isMatchValName` (the builder's read-resolution), and `parseMatchValId` (the checker and synthesizer consumers). (d93bbeaa)

PR #402, opened 2026-07-02
Tool reply attachments: tools can show images to the model

egonSchiele:
## Summary

Tools can now hand images back to the model — the missing half of generate → view → refine, and the channel the future `viewAttachment` tool will reuse. Spec, plan, and plan-review are included (`docs/superpowers/specs|plans/2026-07-02-tool-reply-attachments*`).

A base64 string in a tool result is just text to the model, and most providers don't accept image parts in tool results (Anthropic does; OpenAI chat/Gemini don't). So agency routes tool-produced images through **one labeled user message injected after the complete tool round** — the only placement valid under every provider's tool-call adjacency rules — uniformly across providers, with no smoltalk changes.

## How it works

1. **`std::thread.attachToReply(attachment)`** — a tool calls it during its invocation. The queue lands on the calling invocation's **branch-local `stack.other`**: parallel tool calls each have their own branch stack (no cross-talk), and branch state serializes, so a mid-round interrupt cannot drop a queued attachment. Outside a tool invocation it drops with a statelog error, never a throw.
2. **Harvest at invocation completion** (`runInvokeStep`, inside the idempotent invoke `b.step` — exactly once per tool call across interrupt/resume): a pure gate chain (modality tri-state via smoltalk's own send-gate predicate, missing file, 20 MB, 10-per-call cap) assigns persistent `img_N` ids and appends a model-facing marker to that tool's result, e.g. `[attached img_1 — delivered in the user message following these tool results]`. Skips become honest markers, never failed turns. Survivors buffer in `self.runnerState` (per-`llm()`-call, serialized, fork-safe).
3. **Injection after the round** (`pr.step`, resume-idempotent via the existing `messagesJSON` snapshot/restore mechanism): one user message with a label text part before each image part — `[img_1 — image output of tool generateImageFile]`. Path sources are inlined to base64 at build time so the persistent thread never re-reads a deletable file.

First consumer: `generateImageFile` attaches the image it just saved, so the agent finally sees what it generated.

## Design notes

- **Agency-side by design, not smoltalk**: native image-in-tool-result is Anthropic-only, so a smoltalk-level feature would be a leaky unification; the injected message is an ordinary user message, so attachment resolution, `validateModalities`, statelog redaction, `[image attachment]` transcript rendering, and thread persistence all apply with zero new code — and custom `LLMClient` implementations get the behavior for free.
- **Marker strings are model-facing API** — pinned verbatim by both the vitest units and the execution tests; documented in `docs/dev/reply-attachments.md`.
- On Anthropic the injected message rides as a consecutive user message after the tool_result user message (the API combines same-role messages); the planned smoltalk consecutive-user merge is belt-and-suspenders, not a dependency.
- No e2e *fork* isolation test on purpose: the deterministic client's mock queue is flat and module-keyed, so concurrent branches consume nondeterministically. The isolation property is structural (one `runnerState` per `runPrompt`) and pinned by a vitest unit plus a sequential-calls execution test.

## Testing

- 17 unit tests for the queue/harvest/build module (gates, ids, isolation, inlining, never-throw) + 3 for the bridge.
- 6 execution tests driving real tool-loop rounds via deterministic `{toolCalls}` mocks: single attach, parallel-tool attribution, **mid-round interrupt survival (two-trip resume)**, text-only-model skip, outside-tool drop, sequential-call isolation.
- Agent integration: the coordinator's `generateImageFile` turn shows the image to the model (asserted via the thread reader's `[image attachment]` placeholder).
- Full local verification: lint clean, `lib/runtime` suite 1041/1041, agent suite 125/125.

### DIFF
--- a/packages/agency-lang/docs/superpowers/specs/2026-07-02-tool-reply-attachments-design.md
+++ b/packages/agency-lang/docs/superpowers/specs/2026-07-02-tool-reply-attachments-design.md
@@ -41,7 +41,7 @@ export safe def attachToReply(attachment: Attachment) {
   as a tool inside an llm() call: the tool's text result is delivered
   as usual, and the queued attachments follow it as a single user
   message the model can see. Outside a tool invocation, the attachment
-  is dropped with a statelog warning.
+  is dropped with a statelog error.
 
   @param attachment - The attachment to show the model (from image() / file())
   """
@@ -125,7 +125,7 @@ The model's next turn sees the tool result (`Saved image to … [attached img_1 
 - `viewAttachment` / `listAttachments` and the attachment store — separate track; they become the second consumer of `attachToReply` unchanged.
 - Native Anthropic `tool_result` image blocks — revisit only if providers converge; the agency-side channel is forward-compatible (swap the injection for native parts without touching `attachToReply` or its callers).
 - Refinement prompting (teaching the agent to critique-and-regenerate).
-- Attachments from handler bodies or non-tool contexts (dropped with a statelog warning).
+- Attachments from handler bodies or non-tool contexts (dropped with a statelog error).
 
 ## Testing
 
--- a/packages/agency-lang/docs/superpowers/specs/2026-07-02-tool-reply-attachments-design.md
+++ b/packages/agency-lang/docs/superpowers/specs/2026-07-02-tool-reply-attachments-design.md
@@ -1,0 +1,141 @@
+---
+name: Tool reply attachments (images from tool calls)
+description: Let a tool call hand images back to the model — attachToReply() queues attachments during a tool invocation; the LLM tool loop injects them as one labeled user message after the tool round, uniformly across providers. Unlocks generate→view→refine and the future viewAttachment tool.
+status: implemented
+date: 2026-07-02
+---
+
+# Tool reply attachments
+
+## Summary
+
+A tool cannot "return" an image today: tool results are text, and a base64 string in a tool result is just tokens — it never reaches the model's vision encoder. This spec adds the missing channel, **entirely on the agency side**:
+
+1. **`std::thread.attachToReply(attachment)`** — a tool calls this during its invocation to queue an image (or PDF) for the model.
+2. **The LLM tool loop** (`lib/runtime/prompt.ts`) harvests queued attachments per tool invocation, appends an ID-bearing marker to that tool's result text, and — after ALL tool results for the round — injects **one labeled user message** carrying the image parts, then re-prompts.
+
+First consumer: `generateImageFile` attaches the image it just saved, so the agent can see what it generated. The future `viewAttachment` tool (attachment-store track) becomes the second consumer with no new machinery.
+
+## Background: why this shape
+
+Investigated during brainstorming (2026-07-02), grounded in provider wire formats:
+
+- **Native image-in-tool-result support is not unifiable.** Anthropic's `tool_result.content` accepts image blocks natively; OpenAI chat-completions tool messages are text-only; Gemini `functionResponse` is JSON-only; OpenAI Responses needs verification. smoltalk's value is a unified provider API — a feature that is native on one provider and silently rewrites message history on the others is a leaky unification. **Rejected: teaching smoltalk `ToolMessage` to carry image parts.**
+- **The user-message shim is the portable channel.** Every image-capable provider accepts image parts in a user message, and the modality gate already governs which models those are.
+- **Agency owns the shim, not smoltalk.** The rewrite is conversation *choreography*, not wire serialization. Doing it in the agency tool loop means: (a) users with custom `LLMClient` implementations get it for free; (b) everything downstream — attachment resolution, the 20 MB cap, `validateModalities`, statelog redaction, `[image attachment]` transcript rendering, thread persistence — applies automatically because the injected message is an ordinary user message; (c) it is pinned by agency execution tests with the deterministic client. **Rejected: a smoltalk-internal tool-message→user-message rewrite.**
+- **Adjacency rules dictate the injection point.** Providers require the assistant's tool calls to be answered by ALL their tool results before any other message (OpenAI: every `tool_call_id` must get a `role:"tool"` message first; Anthropic: `tool_result` blocks in the immediately-following user message; Gemini: `functionResponse` follows `functionCall`). Therefore: **inject after the complete tool round, never between results.** `assistant(tool_calls) → tool… → user(images) → assistant` is valid everywhere.
+
+### Prerequisite (smoltalk, owner: Aditya)
+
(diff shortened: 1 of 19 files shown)


packages/agency-lang/lib/runtime/replyAttachments.ts
```diff
@@ -0,0 +1,222 @@
...
+ * a marker to that tool's result text; after the full tool round the loop
+ * injects ONE labeled user message built here. See
+ * docs/superpowers/specs/2026-07-02-tool-reply-attachments-design.md and
+ * docs/dev/reply-attachments.md.
+ *
+ * The marker strings below are MODEL-FACING API — tests pin them; do not
+ * reword casually. (They contain em-dashes, not hyphens.)
+ */
+
+export type ReplyAttachmentPart = {
```
egonSchiele: is there a type we should reuse here?

packages/agency-lang/lib/runtime/replyAttachments.ts
```diff
@@ -0,0 +1,222 @@
...
+  filename?: string;
+};
+
+export type HarvestedReplyAttachment = {
+  id: string;
+  toolName: string;
+  part: ReplyAttachmentPart;
+};
+
+export const MAX_REPLY_ATTACHMENTS_PER_CALL = 10;
```
egonSchiele: Is this an arbitrary cap or does it come from the LLM API? I wonder if it would be useful to make this configurable, or at least to surface it as a constant in lib/config.ts instead of burying it in code here?

packages/agency-lang/lib/runtime/replyAttachments.ts
```diff
@@ -0,0 +1,222 @@
...
+export const MAX_REPLY_ATTACHMENT_BYTES = 20 * 1024 * 1024; // smoltalk's cap
+
+// Callers hand us raw `stack.other` bags rather than StateStack instances:
+// it keeps this module free of a state-layer dependency and trivially
+// unit-testable with `{}` literals. The storage keys stay private here.
+const QUEUE_KEY = "pendingReplyAttachments";
+const BUFFER_KEY = "replyAttachments";
+const COUNTER_KEY = "replyAttachmentCounter";
+
+const EXT_TO_MIME: Record<string, string> = {
```
egonSchiele: Similarly, I feel like we have already defined this map elsewhere.

packages/agency-lang/lib/runtime/replyAttachments.ts
```diff
@@ -0,0 +1,222 @@
...
+      // Intentional swallow: a missing file is a NORMAL outcome here —
+      // the file-not-found gate reports it as a skip marker; this
+      // function only answers "how big, if knowable".
+      return null;
+    }
+  }
+  return null;
+}
+
+type GateResult = { ok: true } | { ok: false; reason: string };
```
egonSchiele: Why not use the Result type that is used everywhere else in agency to denote success or failure? https://agency-lang.com/guide/error-handling.html

packages/agency-lang/lib/stdlib/thread.ts
```diff
@@ -170,6 +174,34 @@ export function _fileAttachment(
...
+    frame.ctx?.statelogClient?.error({
+      errorType: "toolError",
+      message:
+        "attachToReply called outside a tool invocation; attachment dropped",
+      functionName: "attachToReply",
+      retryable: false,
+    });
+    return;
+  }
+  queueReplyAttachment(
```
egonSchiele: This is pretty darn imperative. Would it make sense to add a method to the stack instead that lets people add attachments? Ideally, it would be clear from looking at the stack class all the things you can do with the stack, and putting an `addAttachment` or similar function on there would make it more discoverable and reusable.

packages/agency-lang/docs/superpowers/specs/2026-07-02-tool-reply-attachments-design.md
```diff
@@ -0,0 +1,141 @@
...
+
+```
+export safe def attachToReply(attachment: Attachment) {
+  """
+  Queue an attachment (see image()/file()) to be shown to the model
+  after the current tool call completes. Only meaningful while running
+  as a tool inside an llm() call: the tool's text result is delivered
+  as usual, and the queued attachments follow it as a single user
+  message the model can see. Outside a tool invocation, the attachment
+  is dropped with a statelog warning.
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 1291421c — the spec now says statelog error everywhere, matching the implementation (statelogClient has no warn-level method, so error is what the bridge actually emits).

packages/agency-lang/lib/runtime/replyAttachments.ts
egonSchiele: Yes — fixed in 1291421c. `ReplyAttachmentPart` is now `smoltalk.ImagePart | smoltalk.FilePart` (reachable from the root export via types.js → contentParts.js), which are exactly the parts `image()`/`file()` build. The local redefinition is gone; the source-kind narrowing (`path`/`url`/`base64`) works unchanged against smoltalk's `AttachmentSource`, and provider-file refs pass through like URLs.

packages/agency-lang/lib/runtime/replyAttachments.ts
egonSchiele: It comes from smoltalk: it mirrors `DEFAULT_MAX_ATTACHMENT_BYTES` (20 MB), which smoltalk enforces again at send time — attachments over it become a Failure of the whole call, so this pre-check exists to turn that into a visible skip marker instead. I could not import it directly: smoltalk's package exports map only exposes `.` and `./testing`, and the constant lives in `clients/resolveAttachments.js` without an index re-export. Done in 1291421c: both caps moved to `lib/config.ts` with a comment documenting the smoltalk provenance. Two follow-ups for your call: (a) export `DEFAULT_MAX_ATTACHMENT_BYTES` from smoltalk's index and I'll switch to importing it; (b) making the count cap an AgencyConfig knob is easy if anyone ever needs more than 10 per call — left out for YAGNI.

packages/agency-lang/lib/runtime/replyAttachments.ts
egonSchiele: Correct — it duplicated `MIME_TYPES` in `lib/stdlib/mediaPathScan.ts` (from the detection PR). Fixed in 1291421c: that map is now exported as the canonical media ext→MIME table and replyAttachments imports it (runtime→stdlib imports have precedent: prompt.ts already pulls `abortable` from stdlib).

packages/agency-lang/lib/runtime/replyAttachments.ts
egonSchiele: Done in 1291421c — the gate returns the runtime `ResultValue` via `success()`/`failure()` and the loop checks `isSuccess(gate)`, with the failure's `error` carrying the skip reason. (I'd originally avoided it because the runtime Result carries checkpoint/retryable baggage a pure gate never uses, but consistency with the rest of the codebase wins.)

packages/agency-lang/lib/stdlib/thread.ts
egonSchiele: Agreed and done in 1291421c: `StateStack` now exposes `queueReplyAttachment(part)` and `drainPendingReplyAttachments()` with docstrings explaining the branch-locality and serialization guarantees — the bridge and the tool loop call the methods, and the `pendingReplyAttachments` storage key is private to the class. The harvest function correspondingly takes the drained array (`queued: ReplyAttachmentPart[]`) instead of reaching into a bag, which also made its unit tests plainer.

PR #404, opened 2026-07-02
Cost guards meter subprocess spend via per-call telemetry

egonSchiele:
Closes the cost half of subprocess guard safety (the abort/kill half shipped with pause/resume, PR #398): a parent `guard(cost: $X) { run(...) }` was blind to subprocess LLM spend until the terminal message — after the money was spent. With nesting unblocked, an agent-written child (or grandchild) could make unbounded LLM calls outside every cost guard.

Design doc: `packages/agency-lang/docs/superpowers/specs/2026-07-02-subprocess-cost-telemetry-design.md` (committed with this PR).

## How it works

**Child side — one emission line at the choke point.** Every paid site (llm via `prompt.ts`, `addCost` — which memory and image generation pay through) funnels through `StateStack.chargeGuards`. That method now fire-and-forgets `{ type: "telemetry", costUsd }` to the parent via a new dependency-free leaf module (`costTelemetry.ts`, the `subprocessRunInfo.ts` layering pattern — `stateStack.ts` must not import `ipc.ts`). No-op outside IPC mode; never blocks; never throws; a dead channel is swallowed (the disconnect watchdog is about to reap the process anyway).

**Parent side — one new message case.** `handleTelemetryMessage` bills the charge to the `run()` call-site stack and enforces. On a trip it SIGKILLs the child and rejects the session with the guard-trip abort itself, which propagates through runBatch (errors win over interrupts) → the stdlib `run()`'s plain `try` re-throws trips → the user's owning `guard(cost:)` boundary converts it to the standard cost-limit Failure. No new failure shapes.

**Nested relay is automatic.** The mid-tier's handler charges its own `chargeGuards`, which re-emits upward because that process is itself in IPC mode. Grandchild spend reaches the root with zero explicit relay code.

**`billCharge` extraction.** The billing pair (`localCost += x; chargeGuards(x)`) previously existed inline in `prompt.ts` and `cost.ts`; the telemetry handler would have been a third copy. It is now a named `StateStack.billCharge` that all three sites call (enforcement stays at call sites — the handler skips it post-settle, the others always enforce).

## Decisions and semantics

- **getCost() includes child spend live** — telemetry updates `localCost` alongside the guard charge. Nothing folded child cost into parent accumulators at terminal before, so there is no double-count.
- **Per-call cadence** (owner-approved): paid calls are seconds apart; a telemetry message costs microseconds. Detection latency is at most one paid call, matching in-process CostGuard semantics.
- **Trip kills; it does not pause.** A busy child cannot be checkpointed on demand, so raise-the-budget-and-resume is not possible; partial child work is lost.
- **No double-charge across pause/resume**: charges are per-call incremental and a resumed child's replay skips completed LLM steps (pinned by E2E test).
- **Handler must stay synchronous** (documented invariant): `handleChildMessage` void-invokes its dispatch, so telemetry-before-terminal ordering holds only while the telemetry path contains no awaits.
- **Known getCost() edge** (documented, deliberate): telemetry arriving after a kill still charges budgets via the shared guard references, but can be invisible to `getCost()` if the owning fork branch already joined. Budgets never undercount; `getCost()` may, on abnormal termination only.
- **Tokens are a non-goal**: telemetry is cost-only; cumulative child tokens still arrive terminally via `result.tokens`.

## Tests

Unit (`costTelemetry.test.ts`, `ipc.test.ts`): sender gating/validation/dead-channel, `chargeGuards` emits exactly once per charge even with zero guards (the relay invariant), `billCharge` really charges guards, handler charge/trip/post-settle/malformed paths — the trip test asserts the rejection is the `GuardExceededError` itself (identity matters for `ownedGuardIds` conversion).

E2E (`tests/agency/subprocess/cost-*`), all against the deterministic client's synthetic cost:
1. `cost-guard-trips-on-child-spend` — parent guard trips mid-run; child killed; standard `guardFailure` at the guard boundary.
2. `cost-child-spend-in-getcost` — under budget: run succeeds AND parent `getCost()` shows the spend.
3. `cost-nested-relay-trips-root` — grandchild spend relays through a guardless mid-tier and trips the ROOT guard.
4. `cost-two-children-share-budget` — each child alone is under the cap; together they must trip the ONE shared budget (the counterexample to shipping guards into children).
5. `cost-no-double-charge-across-pause` — one paid call before a pause, one after resume; total is exactly two calls' worth.

Regression: subprocess suite spot checks (run-basic, pause-multi-cycle, concurrent-handled, nested-pause-resume), guard-cost-trip, full `lib/runtime` unit suite (1043 tests), tsc, structural lint — all green.

### DIFF
--- a/packages/agency-lang/lib/runtime/cost.ts
+++ b/packages/agency-lang/lib/runtime/cost.ts
@@ -8,8 +8,7 @@ import { getRuntimeContext } from "./asyncContext.js";
  *  Throws (a guard-trip) if enforcement fails — callers must not swallow it. */
 export function addCost(amount: number): void {
   const stack = getRuntimeContext().stack;
-  stack.localCost += amount;
-  stack.chargeGuards(amount);
+  stack.billCharge(amount);
   stack.enforceGuards();
 }
 
--- a/packages/agency-lang/lib/runtime/costTelemetry.test.ts
+++ b/packages/agency-lang/lib/runtime/costTelemetry.test.ts
@@ -1,0 +1,105 @@
+import { describe, it, expect, vi, afterEach } from "vitest";
+import { isPayableCost, sendCostTelemetryToParent } from "./costTelemetry.js";
+import { StateStack } from "./state/stateStack.js";
+import { CostGuard } from "./guard.js";
+
+describe("isPayableCost", () => {
+  it("accepts positive finite numbers only", () => {
+    expect(isPayableCost(0.5)).toBe(true);
+    expect(isPayableCost(0)).toBe(false);
+    expect(isPayableCost(-1)).toBe(false);
+    expect(isPayableCost(NaN)).toBe(false);
+    expect(isPayableCost(Infinity)).toBe(false);
+    expect(isPayableCost("0.5")).toBe(false);
+    expect(isPayableCost(undefined)).toBe(false);
+  });
+});
+
+describe("sendCostTelemetryToParent", () => {
+  const originalSend = process.send;
+  const originalIpc = process.env.AGENCY_IPC;
+
+  afterEach(() => {
+    process.send = originalSend;
+    if (originalIpc === undefined) delete process.env.AGENCY_IPC;
+    else process.env.AGENCY_IPC = originalIpc;
+    vi.restoreAllMocks();
+  });
+
+  it("sends the telemetry message when in IPC mode", () => {
+    process.env.AGENCY_IPC = "1";
+    const send = vi.fn(() => true);
+    process.send = send as any;
+    sendCostTelemetryToParent(0.5);
+    expect(send).toHaveBeenCalledExactlyOnceWith({ type: "telemetry", costUsd: 0.5 });
+  });
+
+  it("no-ops outside IPC mode", () => {
+    delete process.env.AGENCY_IPC;
+    const send = vi.fn(() => true);
+    process.send = send as any;
+    sendCostTelemetryToParent(0.5);
+    expect(send).not.toHaveBeenCalled();
+  });
+
+  it("no-ops for zero, negative, and non-finite cost", () => {
+    process.env.AGENCY_IPC = "1";
+    const send = vi.fn(() => true);
+    process.send = send as any;
+    sendCostTelemetryToParent(0);
+    sendCostTelemetryToParent(-1);
+    sendCostTelemetryToParent(NaN);
+    expect(send).not.toHaveBeenCalled();
+  });
+
+  it("swallows a dead-channel send error", () => {
+    process.env.AGENCY_IPC = "1";
+    process.send = vi.fn(() => { throw new Error("channel closed"); }) as any;
+    expect(() => sendCostTelemetryToParent(0.5)).not.toThrow();
+  });
+});
+
+describe("StateStack.chargeGuards emission", () => {
+  const originalSend = process.send;
+  const originalIpc = process.env.AGENCY_IPC;
+
+  afterEach(() => {
+    process.send = originalSend;
+    if (originalIpc === undefined) delete process.env.AGENCY_IPC;
+    else process.env.AGENCY_IPC = originalIpc;
+    vi.restoreAllMocks();
+  });
+
+  it("emits exactly once per charge, even with zero guards installed", () => {
+    // Emission must be unconditional on guards being present: a mid-tier
+    // relay process may have NO local guards but must still forward the
+    // grandchild spend upward.
+    process.env.AGENCY_IPC = "1";
+    const send = vi.fn(() => true);
+    process.send = send as any;
+    const stack = new StateStack();
+    stack.chargeGuards(0.25);
+    expect(send).toHaveBeenCalledExactlyOnceWith({ type: "telemetry", costUsd: 0.25 });
+  });
+
+  it("does not emit for a zero charge", () => {
+    process.env.AGENCY_IPC = "1";
+    const send = vi.fn(() => true);
+    process.send = send as any;
+    new StateStack().chargeGuards(0);
+    expect(send).not.toHaveBeenCalled();
+  });
+});
+
+describe("StateStack.billCharge", () => {
+  it("accumulates localCost and really charges guards in one call", () => {
+    const stack = new StateStack();
+    const guard = new CostGuard(0.1);
+    stack.guards.push(guard);
+    stack.billCharge(0.25);
+    expect(stack.localCost).toBe(0.25);
+    // 0.25 > 0.1: check() reporting a trip proves the guard was charged,
+    // not just localCost.
+    expect(guard.check(stack)).not.toBeNull();
+  });
+});
--- a/packages/agency-lang/lib/runtime/costTelemetry.test.ts
+++ b/packages/agency-lang/lib/runtime/costTelemetry.test.ts
@@ -3,6 +3,17 @@ import { isPayableCost, sendCostTelemetryToParent } from "./costTelemetry.js";
 import { StateStack } from "./state/stateStack.js";
 import { CostGuard } from "./guard.js";
 
+// process.send has no vi.stubEnv equivalent — save/restore it manually.
+// Env vars are stubbed per test and reset via vi.unstubAllEnvs(): a leaked
+// AGENCY_IPC=1 would make every billCharge in later tests emit telemetry.
+const originalSend = process.send;
+
+afterEach(() => {
+  process.send = originalSend;
+  vi.unstubAllEnvs();
+  vi.restoreAllMocks();
+});
+
 describe("isPayableCost", () => {
   it("accepts positive finite numbers only", () => {
     expect(isPayableCost(0.5)).toBe(true);
@@ -16,18 +27,8 @@ describe("isPayableCost", () => {
 });
 
 describe("sendCostTelemetryToParent", () => {
-  const originalSend = process.send;
-  const originalIpc = process.env.AGENCY_IPC;
-
-  afterEach(() => {
-    process.send = originalSend;
-    if (originalIpc === undefined) delete process.env.AGENCY_IPC;
-    else process.env.AGENCY_IPC = originalIpc;
-    vi.restoreAllMocks();
-  });
-
   it("sends the telemetry message when in IPC mode", () => {
-    process.env.AGENCY_IPC = "1";
+    vi.stubEnv("AGENCY_IPC", "1");
     const send = vi.fn(() => true);
     process.send = send as any;
     sendCostTelemetryToParent(0.5);
@@ -35,7 +36,6 @@ describe("sendCostTelemetryToParent", () => {
   });
 
   it("no-ops outside IPC mode", () => {
-    delete process.env.AGENCY_IPC;
     const send = vi.fn(() => true);
     process.send = send as any;
     sendCostTelemetryToParent(0.5);
@@ -43,7 +43,7 @@ describe("sendCostTelemetryToParent", () => {
   });
 
   it("no-ops for zero, negative, and non-finite cost", () => {
-    process.env.AGENCY_IPC = "1";
+    vi.stubEnv("AGENCY_IPC", "1");
     const send = vi.fn(() => true);
     process.send = send as any;
     sendCostTelemetryToParent(0);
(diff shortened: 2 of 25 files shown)


packages/agency-lang/lib/runtime/cost.ts
```diff
@@ -8,8 +8,7 @@ import { getRuntimeContext } from "./asyncContext.js";
  *  Throws (a guard-trip) if enforcement fails — callers must not swallow it. */
 export function addCost(amount: number): void {
   const stack = getRuntimeContext().stack;
-  stack.localCost += amount;
-  stack.chargeGuards(amount);
+  stack.billCharge(amount);
```
egonSchiele, line 11: **Blocking: this refactor breaks 4 pre-existing unit tests.** `lib/stdlib/image.test.ts`'s mock stack (`makeStack`) only stubs `localCost`/`chargeGuards`/`enforceGuards`, so `_generateImage → addCost → stack.billCharge` throws `stack.billCharge is not a function`. Verified by running `pnpm vitest run lib/stdlib/image.test.ts` on this branch: 4 failed, 2 passed. Fix is to add `billCharge` to the mock (or have the mock delegate `billCharge` to the real pair) — worth also grepping for any other hand-rolled stack mocks that stub `chargeGuards`.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -831,6 +835,47 @@ function handleErrorMessage(s: RunSession, msg: any): void {
...
+  if (s.settled) return;
+  try {
+    s.stateStack.enforceGuards();
+  } catch (err) {
+    try {
+      s.child.kill("SIGKILL");
+    } catch (killErr) {
+      ipcLog("send", { type: "kill_failed", detail: killErr instanceof Error ? killErr.message : String(killErr) });
+    }
+    settle(s, s.rejectPromise, err);
```
egonSchiele, line 885: **Trip-time race worth a guard while you're here: a pending relayed lock-acquire survives this settle and can mint a permanently-held lock.** `handleLockAcquireMessage` awaits `acquireLocalLock` and then calls `registerSessionLock` with no `s.settled` re-check. This trip path settles the session (running `cleanupSessionLocks`) while such an acquire can still be pending — and if the dying child itself held the contended lock, the settle-time cleanup is exactly what resolves the pending acquire. The late-granted lock is then registered on the dead session and nothing ever releases it: every future `withLock` on that name blocks forever (or times out). The hole is pre-existing (wall-clock and abort kills have the same window), but a cost trip is a much likelier trigger — a child spending in one branch while another branch waits on a contended lock is normal shape. Suggest: after the `await acquireLocalLock(...)` in `handleLockAcquireMessage`, if `s.settled`, call `release()` immediately instead of registering.

Same class, lower stakes, no action needed beyond awareness: an in-flight `handleInterruptMessage` consult (e.g. a user-facing approval prompt for a child interrupt) keeps running after this kill; the eventual decision is absorbed by `trySendDecision`'s `connected` check — orphaned prompt, no crash.

packages/agency-lang/lib/runtime/state/stateStack.ts
```diff
@@ -535,6 +550,11 @@ export class StateStack {
    */
   chargeGuards(amount: number): void {
     for (const g of this.guards) g.charge(amount);
+    // Subprocess: forward this charge to the parent so ITS cost guards
+    // see the spend live. Emission is per paid call and unconditional on
+    // local guards existing: a mid-tier relay may have none but must
+    // still forward grandchild spend upward. No-op outside IPC mode.
+    sendCostTelemetryToParent(amount);
```
egonSchiele: **The emission hook is one level too deep.** After this PR, `billCharge` is `chargeGuards`'s only production caller, and the semantic you want is "emit once per fresh paid charge", not "once per guard-accumulator update". Hoisting `sendCostTelemetryToParent(amount)` into `billCharge` is behavior-identical today (including the automatic grandchild relay, which rides on the parent handler calling `billCharge`) and removes a landmine: any future caller that reaches for `chargeGuards` directly — say a checkpoint/branch reconciliation path that re-applies serialized guard spend — would silently emit duplicate telemetry inside a subprocess and double-bill the parent's budget, with nothing at the call site hinting at IPC. Relatedly, consider privatizing or inlining `chargeGuards` so the bypassable half of the pair (`chargeGuards` without `localCost`) stops being a public entry point — the exact drift `billCharge` was extracted to prevent.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -831,6 +835,47 @@ function handleErrorMessage(s: RunSession, msg: any): void {
...
+ * never undercount; do not "fix" this by skipping post-settle billing. */
+export function handleTelemetryMessage(s: RunSession, msg: IpcTelemetryMessage): void {
+  if (!isPayableCost(msg.costUsd)) return;
+  s.stateStack.billCharge(msg.costUsd);
+  if (s.settled) return;
+  try {
+    s.stateStack.enforceGuards();
+  } catch (err) {
+    try {
+      s.child.kill("SIGKILL");
```
egonSchiele: This `try { s.child.kill("SIGKILL") } catch { ipcLog("kill_failed", ...) }` block is now the third verbatim copy in this file (`settleWithLimitFailure` ~line 692 and the abort `onAbort` path have the identical lines). The resolve-vs-reject difference between the sites is legitimate (a guard trip must propagate as the error object itself for `ownedGuardIds` matching), but the kill mechanics are pure duplication — a future teardown change (SIGTERM-then-SIGKILL grace, sessionId in the log, detaching forwarders) will get applied to one site and missed on another, giving guard-tripped children different termination semantics than limit-tripped ones. Suggest one `killChildSafely(s)` helper used by all three.

packages/agency-lang/lib/runtime/costTelemetry.ts
```diff
@@ -0,0 +1,44 @@
...
+/** The wire contract for a billable cost: a positive finite number.
+ * Shared by the sender and the parent-side handler so both ends of the
+ * channel enforce the same rule (the receiving side matters more — the
+ * child is the less-trusted party). */
+export function isPayableCost(costUsd: unknown): costUsd is number {
+  return typeof costUsd === "number" && Number.isFinite(costUsd) && costUsd > 0;
+}
+
+export function sendCostTelemetryToParent(costUsd: number): void {
+  if (process.env.AGENCY_IPC !== "1" || typeof process.send !== "function") return;
```
egonSchiele: Two small duplications of ipc.ts internals in this leaf, both fixable with the same move this PR already made for `IpcTelemetryMessage` (define in the leaf, re-export from ipc.ts):

1. The inline `process.env.AGENCY_IPC !== "1"` duplicates `isIpcMode()`. If the mode signal ever changes shape, ipc.ts gets updated and this copy is missed — and because sends are fire-and-forget by design, children would silently stop reporting spend with no error anywhere. Moving `isIpcMode()` into this leaf (or `subprocessRunInfo.ts`) and re-exporting from ipc.ts removes the second encoding.

2. The debug log below re-implements `ipcLog`'s gating and has already drifted: ipc.ts caches `AGENCY_IPC_DEBUG` once at module load while this reads it per call, and the `[ipc:telemetry]` prefix differs from `ipcLog`'s `[ipc:child]`/`[ipc:parent]` + timestamp format, so a grep for the standard prefix misses telemetry failures.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -18,6 +18,8 @@ import { runBatch } from "./runBatch.js";
 import { AgencyCancelledError } from "./errors.js";
 import type { State, StateStack } from "./state/stateStack.js";
 import { getSubprocessRunInfo, setSubprocessRunInfo, type SubprocessRunInfo } from "./subprocessRunInfo.js";
+import { isPayableCost, type IpcTelemetryMessage } from "./costTelemetry.js";
+export { type IpcTelemetryMessage };
```
egonSchiele: Dead code: nothing imports `IpcTelemetryMessage` from `ipc.js` (grep across lib/ and tests/ — the union member and the handler signature both use the direct import on the line above). Two import paths for one type just means future code splits between them; suggest deleting this re-export and the `costTelemetry.ts` docstring sentence that references it.

packages/agency-lang/lib/runtime/ipc.test.ts
```diff
@@ -262,3 +265,80 @@ describe("wall-clock timer is per execution segment", () => {
     }
   });
 });
+
+describe("handleTelemetryMessage", () => {
+  const makeSession = (stack: StateStack) => {
```
egonSchiele: Test-harness duplication, worth a small pass: (1) this is the second `makeSession` in this file (the wall-clock suite at line 183 builds the same 13-field `RunSession` literal) — when `RunSession` gains a field the code reads, one of the two `any`-typed mocks silently goes stale; hoist one parameterized `makeSession(overrides)`. (2) In `costTelemetry.test.ts`, the `originalSend`/`originalIpc` save-restore boilerplate is copy-pasted per describe and hand-rolls the undefined-vs-set env restore branch — `vi.stubEnv("AGENCY_IPC", "1")` + `vi.unstubAllEnvs()` in `afterEach` does this correctly for free, and a leaked `AGENCY_IPC=1` is nastier than usual now that `chargeGuards` emits on every charge in whatever test runs next.

egonSchiele (commented):
Reviewed with an 8-angle pass (3 correctness, 3 cleanup, altitude, conventions), every candidate independently verified against the head commit; 7 findings below as inline comments, most severe first. The first (broken `image.test.ts` mocks) is the only blocker — verified by running the suite.

What holds up well, verified rather than assumed: every removed inline billing pair is exactly re-established by `billCharge` (order and amounts, with `localTokens` untouched); fork-join and race cost propagation use raw `localCost +=` so joins never re-emit telemetry; checkpoint restore never calls `charge()` so resume cannot re-emit; the telemetry path is fully synchronous end-to-end so the FIFO/void-dispatch ordering invariant genuinely holds; `runBatch` rethrows the trip identity-preserved so `ownedGuardIds` conversion at the guard boundary works; and the generated stdlib docs match a real `agency doc` regeneration.

Dropped after verification, for the record: env-var read caching in the sender (matches the codebase's per-call `isIpcMode()` style), a telemetry fast-path before the defensive serialize in `handleChildMessage` (sub-microsecond), and a `let`-vs-`const` nit inside an embedded Agency source string (invisible to the linter).

packages/agency-lang/lib/runtime/costTelemetry.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    else process.env.AGENCY_IPC = originalIpc;
+    vi.restoreAllMocks();
+  });
+
+  it("sends the telemetry message when in IPC mode", () => {
+    process.env.AGENCY_IPC = "1";
+    const send = vi.fn(() => true);
+    process.send = send as any;
+    sendCostTelemetryToParent(0.5);
+    expect(send).toHaveBeenCalledExactlyOnceWith({ type: "telemetry", costUsd: 0.5 });
```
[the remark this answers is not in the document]
egonSchiele: Verified against the branch before changing anything: this repo is on vitest ^4.0.16, and toHaveBeenCalledExactlyOnceWith is a built-in Vitest matcher since 3.0 (not a setupFiles custom matcher). Both files pass locally on this branch with the matcher exercised (costTelemetry.test.ts + ipc.test.ts, exit 0). Leaving as is.

packages/agency-lang/lib/runtime/costTelemetry.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  it("emits exactly once per charge, even with zero guards installed", () => {
+    // Emission must be unconditional on guards being present: a mid-tier
+    // relay process may have NO local guards but must still forward the
+    // grandchild spend upward.
+    process.env.AGENCY_IPC = "1";
+    const send = vi.fn(() => true);
+    process.send = send as any;
+    const stack = new StateStack();
+    stack.chargeGuards(0.25);
+    expect(send).toHaveBeenCalledExactlyOnceWith({ type: "telemetry", costUsd: 0.25 });
```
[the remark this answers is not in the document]
egonSchiele: Same as the sibling comment: vitest ^4.0.16 ships this matcher natively (added in 3.0); the test passes on this branch. No change.

packages/agency-lang/lib/runtime/cost.ts
egonSchiele: Fixed in fd2f8828. The mock now has a billCharge that mirrors the real pair (localCost += amount, then chargeGuards(amount)), so all four assertions keep observing the same piece-wise effects, including the chargeGuards-before-enforceGuards ordering test. image.test.ts is 6/6 green. Also grepped for other hand-rolled stack mocks stubbing chargeGuards: the only other hit is stateStack.test.ts, which exercises the real StateStack class — no other mocks to fix.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: Implemented in fd2f8828: after the await in handleLockAcquireMessage, an s.settled re-check releases the late-granted lock immediately and returns instead of registering it on the dead session. lock-cross-process and nested-lock-relay E2E rerun green. The orphaned-consult observation is noted in passing — agreed no action, trySendDecision absorbs it.

packages/agency-lang/lib/runtime/state/stateStack.ts
egonSchiele: Hoisted in fd2f8828: emission now lives in billCharge (the fresh-paid-charge semantic), and chargeGuards is guard-accumulator-only with a comment pointing paid sites at billCharge. A new unit test pins the property you called out: a direct chargeGuards call does NOT emit, so a future restore/reconciliation path re-applying guard spend cannot double-bill the parent. Kept chargeGuards public rather than privatizing: stateStack.test.ts exercises its shared-reference semantics directly, and with emission moved out, the bypassable half is now inert (no localCost, no telemetry) rather than a landmine.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: Implemented in fd2f8828: one killChildSafely(s) helper used by settleWithLimitFailure, the abort path, and the telemetry trip path. The resolve-vs-reject split at each call site is untouched.

packages/agency-lang/lib/runtime/costTelemetry.ts
egonSchiele: Both addressed in fd2f8828. (1) isIpcMode moved to subprocessRunInfo.ts (the dependency-free module), imported by both ipc.ts and the leaf; ipc.ts re-exports it so interrupts.ts/agency.ts imports are unchanged — one encoding of the mode signal. (2) The failure line now matches ipcLog format: [ipc:child] prefix (this sender only runs in a child) + the same HH:MM:SS.mmm timestamp, message telemetry_send_failed — greppable with the standard prefix. Env gating stays per-call in the leaf rather than module-load-cached; divergence is deliberate (keeps the leaf trivially testable) and harmless since the flag never changes mid-process.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: Deleted in fd2f8828, along with the docstring sentence that referenced it. (ipc.ts now re-exports isIpcMode instead — that one has existing consumers importing from ipc.js, per the suggestion on the costTelemetry thread.)

packages/agency-lang/lib/runtime/ipc.test.ts
egonSchiele: Both done in fd2f8828. (1) One canonical makeSession(overrides) at file scope; the wall-clock suite layers via makeSegmentSession and the telemetry suite via makeTelemetrySession — one 13-field literal to keep in sync with RunSession. (2) costTelemetry.test.ts now uses vi.stubEnv per test + a single vi.unstubAllEnvs()/restore afterEach; the hand-rolled undefined-vs-set env restore is gone (process.send save/restore stays manual — vitest has no stub API for it).

PR #416, opened 2026-07-03
Fix streaming onStream to honor callback() builtin registrations

egonSchiele:
## Problem

The documented streaming example doesn't work:

```ts
node main() {
  callback("onStream") as data {
    printJSON(data)
  }
  const response = llm("What is the capital of India?", stream: true)
}
```

It prints `No onStream callback provided for streaming response, returning response synchronously` and never delivers any chunk.

## Root cause

`handleStreamingResponse` resolved the callback by reading `ctx.callbacks.onStream` directly. That object is populated by **only one path** — the programmatic host API (`node.ts` does `Object.assign(execCtx.callbacks, callbacks)` with invocation-time callbacks). The Agency-level `callback("onStream")` builtin does **not** write there: it routes through `_callbackImpl` to either a **scoped** frame callback (registration inside a node body) or `ctx.topLevelCallbacks` (module scope).

Every *other* callback fires through `gatherCallbacks`, which merges all three sources (scoped stack-frame callbacks + top-level registrations + the TS-passed slot). Streaming bypassed that entirely, so a `callback("onStream")` registration was never seen and streaming fell back to synchronous.

## Fix

Route streaming through the same callback machinery as everything else:

- **`hooks.ts`**: add `hasCallbackConsumer(ctx, name, stack)` — reports whether any consumer is registered across all sources a fire would reach (global hooks + `gatherCallbacks`).
- **`streaming.ts`**: decide streaming-vs-sync via `hasCallbackConsumer`, and emit each chunk via `invokeCallbacks` so `AgencyFunction` callbacks are invoked correctly. Emits are `await`ed, which also applies natural backpressure.
- **`prompt.ts`**: thread the branch-local `stateStack` from `_runPrompt` → `dispatchWithRetry` → `dispatchLLMRequest` → `handleStreamingResponse`, so scoped callbacks inside a `fork`/`race` branch resolve on the right frame chain.

## Behavior-change note (corrected from an earlier version of this description)

Routing emits through `invokeCallbacks` **does** change one thing for host-passed `onStream` callbacks: a callback that throws an ordinary JS error is now logged and swallowed by `fireWithGuard` instead of throwing out of `handleStreamingResponse` and aborting the stream. This makes streaming **consistent with every other callback** (they all fire through `fireWithGuard`, which swallows). Intentional — streaming was the lone fail-fast outlier. Guard trips / cancellations still propagate (`fireWithGuard` re-throws `AgencyAbort` / `RestoreSignal`). (Thanks to the review for catching that my first description wrongly claimed "no behavior change.")

## Also fixed in this PR (review point #4)

The streaming branch only returned inside the `done` case; a provider that ended the stream on an `error` chunk (or ended without `done`) fell through to `undefined`, which `dispatchLLMRequest` turned into a generic `No completion returned...` throw — masking the real error. This path is more reachable now that scoped-callback streams route through the streaming branch. Fixed: the branch now surfaces the provider's real error. New test `streaming-error-chunk` pins it.

## Deferred to a follow-up ticket (review points #2 and #3)

The `ctx.onStreamLock` boolean does not provide real mutual exclusion across concurrent branches (serializes independent consumers; force-acquires after a 60s busy-wait; and, now that emits are awaited, is reentrant-hostile if a callback body itself streams). This predates the PR but is newly *reachable* now that multiple branches can register their own scoped `onStream`. Fixing it is a lock redesign (per-consumer async mutex via the existing `withLock`), out of scope for this focused bugfix. Captured as a ticket: **concurrent streaming serializes on a global boolean lock and degrades to interleaving after 60s** (`docs/superpowers/ideas/2026-07-03-concurrent-streaming-serializes-on-global-lock.md`).

## Tests

- `streaming-scoped-onstream` — registers a scoped `callback("onStream")` inside a node and asserts it receives `text,text,done,`. Fails without the fix (`chunks` empty + the exact `No onStream callback provided` message), passes with it.
- `streaming-error-chunk` — a stream ending on an `error` chunk surfaces the provider's real error (`provider exploded`), not the generic message.

## Verification

- Both new regression tests pass; the first fails without the fix.
- `lib/runtime/hooks.test.ts`: 11/11.
- Related agency-js tests (`block-callback-tool-resume`, `entry-module-callback-fires`, `llm-fork-scoped`, `threads-fix-callback-propagation`, `llm-timeout`): pass.
- Pure-Agency callback tests (`callback-basic`, `callback-recursion`, `callback-cleanup`, `callback-multi-same-hook`): pass.
- Full `tsc` build: clean.

### DIFF
--- a/packages/agency-lang/lib/runtime/streaming.ts
+++ b/packages/agency-lang/lib/runtime/streaming.ts
@@ -8,8 +8,10 @@ import {
 } from "smoltalk";
 import { builtinSleep } from "./builtins.js";
 import type { RuntimeContext } from "./state/context.js";
+import type { StateStack } from "./state/stateStack.js";
 import { GraphState } from "./types.js";
 import { isAbortError } from "./errors.js";
+import { hasCallbackConsumer, invokeCallbacks, type CallbackMap } from "./hooks.js";
 
 export function isGenerator(variable: any): boolean {
   const toString = Object.prototype.toString.call(variable);
@@ -22,13 +24,21 @@ export async function handleStreamingResponse(args: {
   ctx: RuntimeContext<GraphState>;
   completion: AsyncGenerator<StreamChunk>;
   prompt: string | UserContentInput;
+  /** The branch-local stack, if this call runs inside a fork/race branch.
+   *  Used to discover scoped `callback("onStream")` registrations on the
+   *  right frame chain — passed straight through to the callback machinery. */
+  stateStack?: StateStack;
 }): Promise<
   Result<{ completion: PromptResult; toolCalls: ToolCallJSON[] }> | undefined
 > {
-  const { ctx, completion, prompt } = args;
+  const { ctx, completion, prompt, stateStack } = args;
   const toolCalls: ToolCallJSON[] = [];
 
-  if (!ctx.callbacks.onStream) {
+  // Resolve `onStream` through the shared callback machinery so a callback
+  // registered via the Agency `callback("onStream")` builtin — scoped inside a
+  // node or at module top level — is found, not just a TS-passed one on
+  // `ctx.callbacks`. See gatherCallbacks / invokeCallbacks in hooks.ts.
+  if (!hasCallbackConsumer(ctx, "onStream", stateStack)) {
     console.log(
       "No onStream callback provided for streaming response, returning response synchronously",
     );
@@ -63,7 +73,12 @@ export async function handleStreamingResponse(args: {
     }
     return { success: true, value: { completion: syncResult!, toolCalls } };
   } else {
-    const onStream = ctx.callbacks.onStream;
+    // Fire every registered onStream consumer for this chunk. Awaited because
+    // Agency-registered callbacks are AgencyFunctions invoked through the
+    // runtime (async); this also applies natural backpressure — we finish
+    // delivering a chunk before pulling the next.
+    const emit = (data: CallbackMap["onStream"]) =>
+      invokeCallbacks({ ctx, name: "onStream", data, stateStack });
     // try to acquire lock
     let count = 0;
     // wait 60 seconds to acquire lock
@@ -80,23 +95,23 @@ export async function handleStreamingResponse(args: {
       for await (const chunk of completion) {
         switch (chunk.type) {
           case "text":
-            onStream({ type: "text", text: chunk.text });
+            await emit({ type: "text", text: chunk.text });
             break;
           case "tool_call":
             toolCalls.push(chunk.toolCall);
-            onStream({
+            await emit({
               type: "tool_call",
               toolCall: chunk.toolCall,
             });
             break;
           case "done":
-            onStream({ type: "done", result: chunk.result });
+            await emit({ type: "done", result: chunk.result });
             return {
               success: true,
               value: { completion: chunk.result, toolCalls },
             };
           case "error":
-            onStream({ type: "error", error: chunk.error });
+            await emit({ type: "error", error: chunk.error });
             break;
         }
       }
--- a/packages/agency-lang/lib/runtime/streaming.ts
+++ b/packages/agency-lang/lib/runtime/streaming.ts
@@ -92,6 +92,7 @@ export async function handleStreamingResponse(args: {
     ctx.onStreamLock = true;
 
     try {
+      let streamError: unknown = undefined;
       for await (const chunk of completion) {
         switch (chunk.type) {
           case "text":
@@ -111,10 +112,23 @@ export async function handleStreamingResponse(args: {
               value: { completion: chunk.result, toolCalls },
             };
           case "error":
+            streamError = chunk.error;
             await emit({ type: "error", error: chunk.error });
             break;
         }
       }
+      // Stream ended without a `done` chunk. If the provider signaled an
+      // `error` chunk, surface THAT so the real failure reaches the caller
+      // (dispatchLLMRequest turns a failure Result into a throw carrying the
+      // message) instead of a generic "No completion returned". Mirrors the
+      // sync-fallback branch, which never masks the error either.
+      return {
+        success: false,
+        error:
+          streamError !== undefined
+            ? String(streamError)
+            : "Streaming response ended without a completion.",
+      };
     } catch (error) {
       if (isAbortError(error)) throw error;
       console.error("Unexpected error consuming LLM stream:", error);
--- a/packages/agency-lang/lib/runtime/hooks.ts
+++ b/packages/agency-lang/lib/runtime/hooks.ts
@@ -244,6 +244,22 @@ export function gatherCallbacks<K extends keyof CallbackMap>(
   return out;
 }
 
+/** Whether any consumer is registered for `name` across all sources a fire
+ *  would reach: global hooks, scoped callbacks on `stack`, top-level
+ *  registrations, and a TS-passed callback. Mirrors what `invokeCallbacks`
+ *  actually fires, so callers can branch on "is anyone listening?" without
+ *  reaching into `ctx.callbacks` directly (which only sees the TS-passed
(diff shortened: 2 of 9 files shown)


packages/agency-lang/lib/runtime/streaming.ts
```diff
@@ -80,23 +95,23 @@ export async function handleStreamingResponse(args: {
       for await (const chunk of completion) {
         switch (chunk.type) {
           case "text":
-            onStream({ type: "text", text: chunk.text });
+            await emit({ type: "text", text: chunk.text });
```
egonSchiele, line 99: Concern about the `onStreamLock` acquire block just above (lines 85–92): `ctx.onStreamLock` is a single boolean on the exec context (`state/context.ts:79`), shared across all fork/race branches. This PR's purpose is to let each branch register its own scoped `callback("onStream")` — but before the fix those branches took the lock-free sync path (never delivering), whereas now they all funnel through this locked `else` branch and contend on that one boolean.

With two branches streaming concurrently (the exact fork/race case): branch A acquires the lock; branch B spins in the wait loop for up to 60s, then line 92 **unconditionally** sets `onStreamLock = true` even though A never released it, and B's `finally` (line 123) later sets it back to `false` while A is still mid-stream. So after 60s the lock provides no mutual exclusion, and before 60s it needlessly serializes branches that each have their own independent consumer. Either way delivery across concurrent branches is wrong.

This busy-wait lock predates the PR, but the PR is what makes concurrent per-branch streaming actually reach it. Worth considering a per-consumer / per-branch guard (or dropping the shared lock) rather than one global boolean now that multiple branches can stream at once.

packages/agency-lang/lib/runtime/streaming.ts
```diff
@@ -63,7 +73,12 @@ export async function handleStreamingResponse(args: {
     }
     return { success: true, value: { completion: syncResult!, toolCalls } };
   } else {
-    const onStream = ctx.callbacks.onStream;
+    // Fire every registered onStream consumer for this chunk. Awaited because
+    // Agency-registered callbacks are AgencyFunctions invoked through the
+    // runtime (async); this also applies natural backpressure — we finish
+    // delivering a chunk before pulling the next.
+    const emit = (data: CallbackMap["onStream"]) =>
```
egonSchiele, line 80: `emit` is now awaited (lines 98–114), so it runs arbitrary Agency callback bodies inline while `ctx.onStreamLock` is held (line 92). Before this PR, `onStream` was a fire-and-forget host JS function, so a callback could never re-enter the streaming critical section. Now a scoped `callback("onStream")` body that issues another `llm(..., stream: true)` — directly or transitively — will re-enter `handleStreamingResponse`, hit the held lock, spin the 60s busy-wait, then force-acquire and on completion clear the *outer* stream's lock.

Probably an edge case, but since the emit is now awaited and runs user Agency code, worth confirming this reentrancy is either impossible (is a streaming `llm()` even permitted inside a callback body?) or handled.

egonSchiele (commented):
Reviewed the streaming `onStream` fix. The core change is right: routing streaming through `hasCallbackConsumer` + `invokeCallbacks` correctly makes scoped/top-level `callback("onStream")` registrations visible, the `stateStack` threading through `_runPrompt → dispatchWithRetry → dispatchLLMRequest → handleStreamingResponse` is complete and correct, and the regression test asserts the exact reported scenario. Nice, well-explained PR.

My one substantive concern is that the fix now routes concurrent fork/race branch streaming (its whole reason for existing) through the single process-wide `ctx.onStreamLock` boolean, which doesn't actually provide mutual exclusion. Details inline. There's also a reentrancy angle now that emits are awaited. Neither is a regression against `main` for the TS-passed-callback path, but both are newly *reachable* by the scenarios this PR enables, so worth a look before merge.

Minor (not blocking): the `else` branch returns `undefined` if the stream ends without a `done` chunk (e.g. a provider that terminates on an `error` chunk), which `dispatchLLMRequest` turns into a generic `No completion returned from streaming LLM call!` throw. The sync-fallback branch above handles the no-`done` case gracefully and returns success. Now that scoped-callback streams route through the `else` branch, that asymmetry is more reachable and would mask the real stream error. Pre-existing in the `else` branch, but flagging since this PR widens who hits it.

packages/agency-lang/lib/runtime/streaming.ts
```diff
@@ -63,7 +73,12 @@ export async function handleStreamingResponse(args: {
...
+      invokeCallbacks({ ctx, name: "onStream", data, stateStack });
     // try to acquire lock
```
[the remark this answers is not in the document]
egonSchiele: Verified — correct, and my PR description was wrong to claim "no behavior change." The old direct `ctx.callbacks.onStream(...)` call threw out of `handleStreamingResponse` and aborted the stream; `invokeCallbacks` → `fireWithGuard` logs and swallows ordinary JS errors.

Keeping the swallow: it makes streaming consistent with every other callback (they all fire through `fireWithGuard`) — streaming was the lone fail-fast outlier. Guard trips and cancellations still propagate, since `fireWithGuard` re-throws `AgencyAbort` / `RestoreSignal`. Not adding a strict/propagateErrors option since no caller needs fail-fast today (YAGNI). Corrected the PR description to document the change.

packages/agency-lang/lib/runtime/streaming.ts
egonSchiele: Confirmed — it isn't real mutual exclusion. Before 60s it needlessly serializes branches that each have an independent consumer; at 60s line 92 force-acquires while A still holds the lock, and B's `finally` then clears A's lock mid-stream. Not a regression vs `main` (the TS-passed path was effectively single-consumer), but this PR is what makes concurrent per-branch streaming actually reach it.

Fixing it is a lock redesign, so I've kept it out of this focused bugfix and filed a ticket: `docs/superpowers/ideas/2026-07-03-concurrent-streaming-serializes-on-global-lock.md`. Proposed direction: replace the global boolean with a per-consumer key on the existing `withLock` async mutex (queued, released in `finally`, no 60s force-acquire), so disjoint consumers stream concurrently and only same-sink streams serialize.

packages/agency-lang/lib/runtime/streaming.ts
egonSchiele: Right — the awaited emit runs the callback body inline while the lock is held, so a callback that issues `llm(..., stream: true)` re-enters, busy-waits the 60s, force-acquires, and its `finally` clears the outer stream's lock. Folded into the same lock-redesign ticket.

Open question I haven't yet confirmed: whether a streaming `llm()` is even permitted inside a callback body — callback bodies are barred from raising interrupts by the typechecker, but a plain streaming `llm()` may still slip through. The ticket proposes either statically disallowing streaming `llm()` in callback bodies (preferred) or keying the mutex so a nested stream can't self-deadlock.

egonSchiele:
Pushed follow-ups addressing the review:

- **Minor (no-`done` masks the real error):** fixed. The streaming branch now captures an `error` chunk and, when the stream ends without `done`, returns a failure Result carrying the provider's real error instead of falling through to `undefined` (which `dispatchLLMRequest` turned into the generic "No completion returned"). New test `streaming-error-chunk` pins it: `try llm(..., stream: true)` against a client that ends on an error chunk now surfaces `provider exploded`.
- **Error-swallowing (Copilot):** correct — replied inline. Kept the swallow for consistency with all other callbacks; corrected the PR description, which wrongly claimed no behavior change.
- **`onStreamLock` mutual exclusion + awaited-emit reentrancy:** confirmed, replied inline, and deferred to a lock-redesign ticket (`docs/superpowers/ideas/2026-07-03-concurrent-streaming-serializes-on-global-lock.md`) rather than balloon this bugfix — the redesign moves to a per-consumer `withLock` mutex and drops the 60s force-acquire.

PR #417, opened 2026-07-04
Subprocess callback forwarding

egonSchiele:
## What

Forwards a subprocess's lifecycle callbacks to its parent over IPC, so a parent's registered `AgencyCallbacks` fire for events that happen inside a `std::agency run()` child (and, recursively, grandchildren).

Mirrors the shipped cost-telemetry design (PR #404): a single choke point (`invokeCallbacks` in `hooks.ts`) fire-and-forgets every event upward via a dependency-light leaf module; the parent re-fires its own callbacks by re-entering `invokeCallbacks`, which makes nested (grandchild) relay automatic.

## How it works

- **Child side** (`callbackForwarding.ts`, new leaf): `invokeCallbacks` calls `sendCallbackToParent(name, data)` on every event. JSON-serialized once (strips functions like `onAgentStart.cancel`; unserializable payloads degrade to a dropped event). `hooks.ts` must not import `ipc.ts`, so the wire type + sender live in the leaf.
- **Parent side** (`ipc.ts`): a new `{ type: "callback", name, data }` message on `SubprocessToParent`; `handleCallbackMessage` (synchronous, like `handleTelemetryMessage`) validates the name against `VALID_CALLBACK_NAMES`, drops post-settle events, and `void invokeCallbacks(...)` fire-and-forget.
- **Nested relay is automatic** — `invokeCallbacks` re-emits when THIS process is itself a subprocess.
- **`onAgentStart.cancel` reconstructed** parent-side to a real (best-effort, async) kill + settle-cancelled.
- **Observational, never fatal** — an oversize/unserializable `callback` message is dropped, not settled (`isObservationalMessage` carve-out in `handleChildMessage`).
- **Not forwarded:** `onStream` and `onOAuthRequired` (denylisted — carry functions/Promises needing a live channel); `onTrace` (never dispatched today).

## Notable fix (found by the E2E tests)

The E2E tests exposed a real bug the unit tests missed (they used plain-JS callbacks via `ctx.callbacks`, bypassing both the scoped-callback stack walk and the `AgencyFunction` body path):

1. `handleCallbackMessage` walked the `run()` call-site slice (`s.stateStack`), so a node-level `callback("onNodeStart")` registered on an ancestor frame was never gathered. Fix: fire with `callHook` semantics (walk the full `ctx.stateStack`), matching in-process firing.
2. It fired from the event-loop message handler with no `agencyStore` frame, so an `AgencyFunction` callback body threw on `__globals()` (undefined store) and was swallowed by `fireWithGuard`. Fix: capture the parent's ALS store frame at `run()` time (`RunSession.parentStore`, threaded through `invokeSubprocess` → `runSubprocessSession`) and re-establish it around the fire.

A scoped-callback regression unit test now covers this (fails on the old slice-walk).

## Tests

- Unit: `callbackForwarding.test.ts` (sender: IPC gate, function-stripping, oversize/unserializable drop, dead-channel, denylist); `hooks.test.ts` (choke-point emit + additive firing); `ipc.test.ts` (parent handler, scoped-callback regression, guard discrimination, post-settle drop, `onAgentStart.cancel` reconstruction, observational drop for oversize + unserializable, `handleChildMessage` dispatch routing).
- E2E (`tests/agency/subprocess/`): `callback-forwarding-child-events` (parent callback fires for a child node entry), `callback-forwarding-nested-relay` (root callback fires for a grandchild event through a callbackless mid-tier).
- Ride-along: hoisted `ipcChildDebug` into `subprocessRunInfo.ts` and refactored `costTelemetry.ts` to share it.

Local: 63 unit tests (4 files) + both E2E pass; `tsc --noEmit` and `lint:structure` clean. The full agency suite runs in CI.

## Docs

`docs/dev/subprocess-ipc.md` (new "Callback forwarding" section, removed the limitation bullet, added both E2E rows) and `docs/misc/lifecycleHooks.md` (subprocess-forwarding note).

### DIFF
--- a/packages/agency-lang/lib/runtime/callbackForwarding.ts
+++ b/packages/agency-lang/lib/runtime/callbackForwarding.ts
@@ -1,8 +1,8 @@
 /**
  * Fire-and-forget forwarding of lifecycle callbacks from a subprocess to its
  * parent, so a parent's registered AgencyCallbacks fire for events that happen
- * inside a std::agency run() child (see
- * docs/superpowers/specs/2026-07-02-subprocess-callback-forwarding-design.md).
+ * inside a std::agency run() child (see the "Callback forwarding" section of
+ * docs/dev/subprocess-ipc.md).
  *
  * Dependency-light leaf (mirrors costTelemetry.ts): the only runtime import is
  * subprocessRunInfo.ts (for isIpcMode + the shared ipcChildDebug). invokeCallbacks
@@ -43,8 +43,10 @@ export const CALLBACK_PAYLOAD_LIMIT = 1024 * 1024 * 1024; // 1gb
  * defensive guard against a future refactor routing them through the choke
  * point. onTrace is function-free and simply never fires today, so it is
  * excluded by non-existence rather than listed here. See the plan's
- * "Scope & design tradeoffs" section. */
-const NON_FORWARDABLE_CALLBACKS: readonly CallbackName[] = ["onStream", "onOAuthRequired"];
+ * "Scope & design tradeoffs" section. Exported so the parent-side handler
+ * (ipc.ts) rejects the same names — the guard should match its intent even under
+ * parent/child version skew. */
+export const NON_FORWARDABLE_CALLBACKS: readonly CallbackName[] = ["onStream", "onOAuthRequired"];
 
 /** Forward one lifecycle event to the parent. No-op unless this process is a
  * forked Agency subprocess with a live IPC channel. `maxBytes` is overridable
@@ -72,9 +74,13 @@ export function sendCallbackToParent(
     return;
   }
   try {
-    // Re-parse so the sent object is guaranteed function-free regardless of the
-    // channel's serialization mode.
-    process.send(JSON.parse(serialized) as IpcCallbackMessage);
+    // Send the object directly: the fork uses the default ("json") IPC
+    // serialization (buildForkOptions sets no `serialization` option), so
+    // process.send JSON-serializes internally — stripping function fields and
+    // producing the SAME wire bytes we just measured. The up-front JSON.stringify
+    // stays only to validate serializability + measure size; a redundant
+    // parse-then-re-serialize round-trip would triple the hot-path cost.
+    process.send({ type: "callback", name, data } as IpcCallbackMessage);
   } catch (err) {
     // Channel gone — parent died; the watchdog will reap this process.
     ipcChildDebug(`callback_send_failed ${name} ${err instanceof Error ? err.message : String(err)}`);
--- a/packages/agency-lang/lib/runtime/callbackForwarding.ts
+++ b/packages/agency-lang/lib/runtime/callbackForwarding.ts
@@ -1,0 +1,82 @@
+/**
+ * Fire-and-forget forwarding of lifecycle callbacks from a subprocess to its
+ * parent, so a parent's registered AgencyCallbacks fire for events that happen
+ * inside a std::agency run() child (see
+ * docs/superpowers/specs/2026-07-02-subprocess-callback-forwarding-design.md).
+ *
+ * Dependency-light leaf (mirrors costTelemetry.ts): the only runtime import is
+ * subprocessRunInfo.ts (for isIpcMode + the shared ipcChildDebug). invokeCallbacks
+ * (hooks.ts) calls sendCallbackToParent on every event; hooks.ts must not import
+ * ipc.ts, so the wire type + sender live here.
+ *
+ * Never blocks, never throws, and never kills the run: no reply, no listener; a
+ * dead channel or an over-limit / unserializable payload is swallowed — the
+ * event is observational, so dropping it is always safe.
+ *
+ * Forwarding is UNCONDITIONAL: the child cannot know which callbacks the parent
+ * registered, so every event (except the NON_FORWARDABLE_CALLBACKS denylist) is
+ * serialized and sent on every occurrence, even when no parent callback exists.
+ * Payloads can be large (full messages arrays, run results). Accepted v1
+ * tradeoff — see the plan's "Scope & design tradeoffs".
+ */
+
+import { isIpcMode, ipcChildDebug } from "./subprocessRunInfo.js";
+import type { CallbackName } from "../types/function.js";
+
+export type IpcCallbackMessage = {
+  type: "callback";
+  name: CallbackName;
+  // JSON-sanitized payload; function fields (e.g. onAgentStart.cancel) dropped.
+  data: any;
+};
+
+/** Skip threshold for a forwarded callback payload. Matches the default
+ * ipcPayload limit (ipc.ts DEFAULT_LIMITS) so the child's skip and the parent's
+ * drop agree. Purely defensive: real callback payloads are KB-MB, far under it. */
+export const CALLBACK_PAYLOAD_LIMIT = 1024 * 1024 * 1024; // 1gb
+
+/** Callbacks that MUST NOT be forwarded. onStream and onOAuthRequired carry
+ * function / Promise fields (cancel, complete) whose semantics require a live
+ * local channel; JSON forwarding would strip them and fire a broken callback in
+ * the parent. Neither currently reaches this function (onStream dispatches
+ * outside invokeCallbacks; onOAuthRequired is never fired), so this is a
+ * defensive guard against a future refactor routing them through the choke
+ * point. onTrace is function-free and simply never fires today, so it is
+ * excluded by non-existence rather than listed here. See the plan's
+ * "Scope & design tradeoffs" section. */
+const NON_FORWARDABLE_CALLBACKS: readonly CallbackName[] = ["onStream", "onOAuthRequired"];
+
+/** Forward one lifecycle event to the parent. No-op unless this process is a
+ * forked Agency subprocess with a live IPC channel. `maxBytes` is overridable
+ * only so tests can exercise the oversize-skip without a gigabyte payload. */
+export function sendCallbackToParent(
+  name: CallbackName,
+  data: unknown,
+  maxBytes: number = CALLBACK_PAYLOAD_LIMIT,
+): void {
+  if (!isIpcMode() || typeof process.send !== "function") return;
+  if (NON_FORWARDABLE_CALLBACKS.includes(name)) return; // functions/Promises can't survive JSON
+  // JSON.stringify drops function-valued fields (e.g. onAgentStart.cancel) and
+  // throws on circular refs / BigInt. Serialize ONCE up front so an
+  // un-serializable payload degrades to a dropped event instead of a throw
+  // inside process.send, and so what we measure equals what we send.
+  let serialized: string;
+  try {
+    serialized = JSON.stringify({ type: "callback", name, data });
+  } catch (err) {
+    ipcChildDebug(`callback_unserializable ${name} ${err instanceof Error ? err.message : String(err)}`);
+    return;
+  }
+  if (Buffer.byteLength(serialized, "utf8") > maxBytes) {
+    ipcChildDebug(`callback_dropped_oversize ${name}`);
+    return;
+  }
+  try {
+    // Re-parse so the sent object is guaranteed function-free regardless of the
+    // channel's serialization mode.
+    process.send(JSON.parse(serialized) as IpcCallbackMessage);
+  } catch (err) {
+    // Channel gone — parent died; the watchdog will reap this process.
+    ipcChildDebug(`callback_send_failed ${name} ${err instanceof Error ? err.message : String(err)}`);
+  }
+}
--- a/packages/agency-lang/lib/runtime/ipc.ts
+++ b/packages/agency-lang/lib/runtime/ipc.ts
@@ -15,11 +15,11 @@ import type { AgencyConfig } from "../config.js";
 import { getRuntimeContext, agencyStore } from "./asyncContext.js";
 import { gatherChainOutcome, type HandlerChainOutcome, type Interrupt } from "./interrupts.js";
 import { runBatch } from "./runBatch.js";
-import { AgencyCancelledError } from "./errors.js";
+import { AgencyAbort, AgencyCancelledError } from "./errors.js";
 import type { State, StateStack } from "./state/stateStack.js";
 import { getSubprocessRunInfo, setSubprocessRunInfo, isIpcMode, type SubprocessRunInfo } from "./subprocessRunInfo.js";
 import { isPayableCost, type IpcTelemetryMessage } from "./costTelemetry.js";
-import { type IpcCallbackMessage } from "./callbackForwarding.js";
+import { type IpcCallbackMessage, NON_FORWARDABLE_CALLBACKS } from "./callbackForwarding.js";
 import { invokeCallbacks } from "./hooks.js";
 import { VALID_CALLBACK_NAMES, type CallbackName } from "../types/function.js";
 // isIpcMode lives in subprocessRunInfo.ts (dependency-free, so the telemetry
@@ -897,7 +897,13 @@ export function handleTelemetryMessage(s: RunSession, msg: IpcTelemetryMessage):
 }
 
 function isForwardableCallbackName(name: unknown): name is CallbackName {
-  return typeof name === "string" && (VALID_CALLBACK_NAMES as readonly string[]).includes(name);
+  return typeof name === "string"
+    && (VALID_CALLBACK_NAMES as readonly string[]).includes(name)
+    // Reject the same names the child-side sender denylists. The child never
+    // forwards these, but a version-skewed or future-refactored child might; a
+    // JSON-stripped onOAuthRequired/onStream would fire a broken (function-less)
+    // parent callback, so the guard must match its name and exclude them.
+    && !(NON_FORWARDABLE_CALLBACKS as readonly string[]).includes(name);
 }
 
 /**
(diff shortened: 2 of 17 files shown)


packages/agency-lang/lib/runtime/callbackForwarding.ts
```diff
@@ -0,0 +1,82 @@
...
+/** Callbacks that MUST NOT be forwarded. onStream and onOAuthRequired carry
+ * function / Promise fields (cancel, complete) whose semantics require a live
+ * local channel; JSON forwarding would strip them and fire a broken callback in
+ * the parent. Neither currently reaches this function (onStream dispatches
+ * outside invokeCallbacks; onOAuthRequired is never fired), so this is a
+ * defensive guard against a future refactor routing them through the choke
+ * point. onTrace is function-free and simply never fires today, so it is
+ * excluded by non-existence rather than listed here. See the plan's
+ * "Scope & design tradeoffs" section. */
+const NON_FORWARDABLE_CALLBACKS: readonly CallbackName[] = ["onStream", "onOAuthRequired"];
```
egonSchiele: do we have a ticket to make onStream work in the future?

packages/agency-lang/lib/runtime/callbackForwarding.ts
```diff
@@ -0,0 +1,82 @@
...
+    ipcChildDebug(`callback_dropped_oversize ${name}`);
+    return;
+  }
+  try {
+    // Re-parse so the sent object is guaranteed function-free regardless of the
+    // channel's serialization mode.
+    process.send(JSON.parse(serialized) as IpcCallbackMessage);
+  } catch (err) {
+    // Channel gone — parent died; the watchdog will reap this process.
+    ipcChildDebug(`callback_send_failed ${name} ${err instanceof Error ? err.message : String(err)}`);
```
egonSchiele, line 86: Do these debugs get logged to statelog?

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -447,6 +451,12 @@ export type RunSession = {
   limits: RunLimits;
   ctx: any;
   stateStack: any;
+  /** The parent's full ALS store frame captured at run() time. Forwarded
```
egonSchiele, line 454: does this break serialization/deserialization?

packages/agency-lang/tests/agency/subprocess/callback-forwarding-nested-relay.agency
```diff
@@ -0,0 +1,59 @@
+import { compile, run } from "std::agency"
+
+// The grandchild fires onNodeStart for "grandNode". The mid-tier child has no
```
egonSchiele, line 3: Will this test what it's meant to test? Any other test cases you need to add?

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -886,19 +896,109 @@ export function handleTelemetryMessage(s: RunSession, msg: IpcTelemetryMessage):
...
+  if (s.settled) return; // drop post-settle events
+  if (!isForwardableCallbackName(msg.name)) return; // child is less-trusted
+
+  const data = msg.name === "onAgentStart" ? withParentCancel(s, msg.data) : msg.data;
+  // Fire within the parent's captured ALS frame so an AgencyFunction callback
+  // body resolves __globals()/__threads() against the parent's real state (we
+  // run from the event-loop message handler, outside any agencyStore frame).
+  // The synchronous `void invokeCallbacks` starts inside agencyStore.run, so its
+  // async continuations inherit the frame while FIFO ordering is preserved.
+  const fire = () => void invokeCallbacks({ ctx: s.ctx, name: msg.name, data });
```
egonSchiele: **`void invokeCallbacks` can produce an unhandled rejection and silently swallows a guard trip.**

`fireWithGuard` (hooks.ts:215-216) deliberately re-throws `RestoreSignal` and `AgencyAbort`:

```js
if (error instanceof RestoreSignal) throw error;
if (error instanceof AgencyAbort) throw error;
```

with the explicit intent (per its comment) that "a guard trip raised inside a callback must propagate to its owning guard, not be logged + dropped." So `invokeCallbacks` **can reject** — and `AgencyAbort` covers both cancellation and cost-guard trips.

Here the parent fires forwarded callbacks fire-and-forget via `void invokeCallbacks(...)` with no `.catch()`. If a parent callback (fired for a *child* event) does paid work that trips a cost guard, or is hit by cancellation during the fire, the thrown `AgencyAbort`/`RestoreSignal`:

1. becomes an **unhandled promise rejection** in the parent (Node terminates the process on unhandledRejection by default), and
2. is **never delivered to its owning guard** — defeating exactly the propagation `fireWithGuard`'s re-throw exists to preserve. In the in-process path this works because the caller `await`s `invokeCallbacks`; here the `void` orphans it.

The existing tests don't catch this — the `onAgentStart.cancel` test settles the promise rather than throwing. Suggest at least attaching a `.catch()` that routes `AgencyAbort` to `settle(s, s.rejectPromise, err)` (mirroring how `handleTelemetryMessage` handles a guard trip) and logs/ignores the rest, so a throwing forwarded callback can't crash the parent or lose a trip.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -886,19 +896,109 @@ export function handleTelemetryMessage(s: RunSession, msg: IpcTelemetryMessage):
   }
 }
 
-async function handleChildMessage(s: RunSession, msg: any): Promise<void> {
+function isForwardableCallbackName(name: unknown): name is CallbackName {
```
egonSchiele, line 899: Minor / defense-in-depth: `isForwardableCallbackName` validates against `VALID_CALLBACK_NAMES` (all names), but the name implies it excludes the non-forwardable ones. The `onStream`/`onOAuthRequired` denylist lives only child-side (`NON_FORWARDABLE_CALLBACKS` in callbackForwarding.ts). So if such a message ever reaches the parent — version skew between parent/child, or the very future refactor that routes them through the choke point that the child-side denylist is guarding against — the parent will accept it and fire the callback with a JSON-stripped (function-less) payload. `onOAuthRequired`'s stripped Promise/functions would fire a broken oauth callback. Consider having the parent reject the same denylist (share the constant), so the guard actually matches its name.

packages/agency-lang/lib/runtime/callbackForwarding.ts
```diff
@@ -0,0 +1,82 @@
...
+    return;
+  }
+  if (Buffer.byteLength(serialized, "utf8") > maxBytes) {
+    ipcChildDebug(`callback_dropped_oversize ${name}`);
+    return;
+  }
+  try {
+    // Re-parse so the sent object is guaranteed function-free regardless of the
+    // channel's serialization mode.
+    process.send(JSON.parse(serialized) as IpcCallbackMessage);
```
egonSchiele: Minor efficiency (the PR already acknowledges "unconditional + heavy" as a v1 tradeoff, so just flagging the incremental part): each event now serializes the payload three times on the hot path — `JSON.stringify`, then `JSON.parse`, then `process.send` re-serializes the parsed object. The stringify+parse round-trip only exists to guarantee a function-free object; with the default (JSON) child_process serialization, `process.send` already strips functions, so the re-parse is redundant unless advanced/structured-clone serialization is enabled. For large `messages` payloads on frequent events (onToolCall*/onNode*), this roughly triples serialization cost. If default serialization is in use, sending the payload object directly (letting `process.send` do the one serialization) would be equivalent and cheaper.

egonSchiele (commented):
Reviewed the callback-forwarding change. The layering (leaf sender + choke-point emit), the observational carve-out, and the scoped-callback / ALS-frame fixes all look right, and the `parentStore` capture is correct (`getRuntimeContext()` returns `agencyStore.getStore()`, so re-establishing it is sound). One correctness concern on the parent-side fire, plus two smaller robustness notes below.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -886,19 +896,109 @@ export function handleTelemetryMessage(s: RunSession, msg: IpcTelemetryMessage):
...
+function withParentCancel(s: RunSession, data: unknown): unknown {
+  if (typeof data !== "object" || data === null) return data;
+  return {
+    ...(data as Record<string, unknown>),
+    cancel: (reason?: string) => {
+      killChildSafely(s);
+      settle(s, s.rejectPromise, new AgencyCancelledError(reason));
+    },
+  };
+}
```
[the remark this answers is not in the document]
egonSchiele: Verified this doesn't pollute the prototype: object spread copies own enumerable properties via `CreateDataPropertyOrThrow` (a *define*), not `[[Set]]`, so a `__proto__` key in the incoming payload lands as a plain own data property on the result rather than reaching `Object.prototype`:

```
> const d = JSON.parse('{"__proto__":{"polluted":1},"a":2}');
> const r = {...d, cancel(){}};
> r.polluted                       // undefined
> ({}).polluted                    // undefined   (Object.prototype clean)
> Object.getOwnPropertyNames(r)    // ['__proto__','a','cancel']
```

Prototype pollution needs assignment (`obj[k] = v`) or a recursive merge that uses `[[Set]]` — neither of which object spread does. Leaving as-is.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -886,19 +896,109 @@ export function handleTelemetryMessage(s: RunSession, msg: IpcTelemetryMessage):
...
+
+/** Message types whose delivery is observational — an oversize or
+ * unserializable one is DROPPED, never fatal, because it cannot affect the run
+ * outcome (unlike result/interrupt/error/lock messages, which must settle or
+ * kill the run). Extend this list when adding another fire-and-forget message. */
+const OBSERVATIONAL_MESSAGE_TYPES: readonly string[] = ["callback"];
+
+function isObservationalMessage(msg: { type?: string }): boolean {
+  return typeof msg.type === "string" && OBSERVATIONAL_MESSAGE_TYPES.includes(msg.type);
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 207347a8. `isObservationalMessage` now reads `.type` null-safely, so a malformed `undefined`/`null` child message settles via the serialize-error path instead of throwing — a throw there would escape the void-invoked handler and leave the session unsettled (a hung run). Added a regression test that sends `undefined` and asserts it settles rather than throws.

packages/agency-lang/lib/runtime/callbackForwarding.ts
```diff
@@ -0,0 +1,82 @@
+/**
+ * Fire-and-forget forwarding of lifecycle callbacks from a subprocess to its
+ * parent, so a parent's registered AgencyCallbacks fire for events that happen
+ * inside a std::agency run() child (see
+ * docs/superpowers/specs/2026-07-02-subprocess-callback-forwarding-design.md).
+ *
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 207347a8. The design spec lives under `docs/superpowers/`, which is gitignored, so it never ships in-repo. Repointed the header to the "Callback forwarding" section of `docs/dev/subprocess-ipc.md`.

packages/agency-lang/lib/runtime/callbackForwarding.ts
egonSchiele: No ticket yet. The inclusion cost is captured in the plan and the docs: it's moderate — `onStream` bypasses the `invokeCallbacks` choke point (it's dispatched directly at the `streaming.ts` call site), and the child only enters the streaming branch when it has its own `onStream` callback, so forwarding to a parent would need (a) an explicit forward at that call site, (b) a way for the child to stream even without a local callback, and (c) accepting per-chunk IPC volume. Happy to file a tracking issue if you want one.

packages/agency-lang/lib/runtime/callbackForwarding.ts
egonSchiele: No — `ipcChildDebug` writes straight to `process.stderr`, gated on `AGENCY_IPC_DEBUG=1`; it never touches statelog. It's the child-side mirror of `ipcLog`, which is unreachable from these dependency-light leaf modules without breaking the layering rule (leaves must not import `ipc.ts`). Same behavior as the pre-existing cost-telemetry leaf — this PR just hoists the one shared implementation into `subprocessRunInfo.ts`.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: No. `RunSession` is a live in-process object — it holds the child process handle, the resolve/reject callbacks, and timers — and is never serialized or checkpointed. `parentStore` holds the parent's live ALS frame (ctx / globals / threads / stack) by reference, also never serialized. Nothing in the checkpoint/state-restoration path touches `RunSession`, so there's no serialization impact.

packages/agency-lang/tests/agency/subprocess/callback-forwarding-nested-relay.agency
egonSchiele: Yes. If the relay were broken, the grandchild's `onNodeStart` never reaches the root, `seen` stays 0, and the exact-match on `"grandchildSeen:true"` fails. The mid-tier registers no callbacks of its own, so a pass proves the relay is *automatic* — its `invokeCallbacks` re-emits upward because that process is itself in IPC mode — rather than the mid-tier forwarding on the root's behalf.

For the edge behaviors I added unit tests in 207347a8: guard-trip (`AgencyAbort`) routing, parent-side denylist rejection, and malformed-message null-safety, plus the existing scoped-callback and dispatch-route regression tests. The `onAgentStart.cancel` path stays unit-only (it surfaces as a propagating abort with no deterministic Agency output string to assert on E2E). If you'd like an E2E where the mid-tier *also* has its own callback (both fire, no interference), I can add that.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: Fixed in 207347a8. `handleCallbackMessage` now attaches `.catch` to the fire: an `AgencyAbort` (guard trip / cancellation — which `fireWithGuard` deliberately re-throws) routes to `killChildSafely(s)` + `settle(s, s.rejectPromise, err)`, mirroring how `handleTelemetryMessage` handles a guard trip; anything else is logged via `ipcLog`. Attaching the handler is synchronous, so the FIFO arrival-order guarantee still holds. Added a test that throws an `AgencyAbort` from a parent callback and asserts the child is killed and the session settles with that error.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: Fixed in 207347a8. Exported `NON_FORWARDABLE_CALLBACKS` and `isForwardableCallbackName` now excludes it, so the parent rejects `onStream`/`onOAuthRequired` too — not just the child sender. Added a test asserting a denylisted name doesn't fire even when the parent registered a handler under it.

packages/agency-lang/lib/runtime/callbackForwarding.ts
egonSchiele: Fixed in 207347a8. Confirmed `buildForkOptions` sets no `serialization` option, so the fork uses the default `"json"` mode — `process.send` JSON-serializes internally and strips functions. Now sending the object directly and dropping the redundant `JSON.parse` round-trip; the up-front `JSON.stringify` stays only to validate serializability and measure size. The code comment ties this to the default-serialization assumption, and I updated the sender's unit-test mock to JSON round-trip so it faithfully reflects the real channel's wire payload (that's where function-stripping now happens).

packages/agency-lang/lib/runtime/callbackForwarding.ts
egonSchiele: Done in b1256f1d. `ipcChildDebug` now also posts a best-effort statelog `debug` event (resolved from the active ALS frame's `ctx.statelogClient`), in addition to the `AGENCY_IPC_DEBUG=1` stderr sink — so the child-side IPC diagnostics (`callback_send_failed`, `callback_dropped_oversize`, `callback_unserializable`, `telemetry_send_failed`) show up in the trace when observability is on. It no-ops with no active frame/client and never throws (the diagnostic sink must not affect the run). The new `subprocessRunInfo -> asyncContext` edge is cycle-safe — asyncContext's value chain (`bootstrapThreadStore -> threadStore -> statelogClient / messageThread`) never imports back into this module. Added `subprocessRunInfo.test.ts`.

packages/agency-lang/lib/runtime/callbackForwarding.ts
egonSchiele: Filed: #418. Captures why it's excluded today and the three things forwarding it would take (explicit forward at the `streaming.ts` call site, a way for the child to stream without a local callback, and accepting per-chunk IPC volume).

packages/agency-lang/tests/agency/subprocess/callback-forwarding-nested-relay.agency
egonSchiele: Added coverage in b1256f1d, but as a deterministic **unit** test rather than an E2E — a clean E2E turns out to be blocked by a pre-existing limitation, not the feature:

- A forwarded-event callback fires on the `parentStore` frame, so it can only touch **globals** (via `__globals()`), not node-locals of the registering scope.
- A `run()`-compiled child module doesn't initialize top-level globals — confirmed with a minimal repro (`let counter = 7; node main(){ counter = counter+1 }` in a compiled child → `counter is not defined`, no callbacks involved).

So a compiled mid-tier has no observable mutable state to record its own callback firing. The new unit test (`handleCallbackMessage` … "re-fires the mid-tier's own callback AND re-forwards upward") models the mid-tier exactly: on a forwarded grandchild event it BOTH fires its own registered callback AND re-forwards upward (because it's in IPC mode). Combined with the existing `nested-relay` E2E (which proves the re-forwarded event reaches the root through the mid), that covers "both fire independently." Happy to file the compiled-child-globals limitation separately if it's worth tracking.

egonSchiele: Correction — my earlier "compiled children can't hold top-level globals" diagnosis was wrong. Compiled children **do** hold top-level globals (verified: a compiled child with `let counter = 7; … return counter` returns 8), and a forwarded-event callback in a mid-tier reaches them.

The real blocker was a separate codegen bug: a triple-quoted **raw** string interpolates `${...}` at the enclosing level, because the string body is emitted into a JS template literal without escaping `${`. So `${midSeen}` inside my nested child source was interpolated by the *outer* program (`ReferenceError: midSeen is not defined`), and the child never ran. Isolated repro + details in #419.

With `${...}` kept out of the nested source, the E2E works — added in 9476a600 as `callback-forwarding-relay-both-fire.agency`: the mid-tier's own `onNodeStart` callback fires for the forwarded grandchild event (updating the mid's global) AND re-forwards to the root, whose callback also fires → `"root:1|mid:1"`, with each callback's `grandNode` filter respected. The deterministic unit test remains as a fast complement (comment corrected).

PR #420, opened 2026-07-04
llm(): accept per-provider object for per-call apiKey

### LINKED ISSUE #405: llm() apiKey option should accept a per-provider object, not just a string
## Summary

The per-call `apiKey` option on `llm()` only accepts a bare `string`, which is treated as the OpenAI key. It should accept a per-provider object so callers can supply keys for non-OpenAI providers on a per-call basis.

## Current behavior

- The typechecker types `apiKey` as `optional(string)` — see [`lib/typeChecker/builtins.ts`](../../lib/typeChecker/builtins.ts) (`llmOptionProperties`, the `apiKey` entry).
- The runtime only honors the string form: `toSmolConfig` maps a per-call string into `apiKey.openAi` and ignores any other shape — see `lib/runtime/llmClient.ts` (`toSmolConfig`, the `typeof apiKey === "string"` branch).
- As a result, a caller who wants to pass, say, an Anthropic or Google key per call cannot do so through `llm(..., { apiKey })`; they must fall back to `agency.json` `client.apiKey` (which *is* a per-provider object) or environment variables.

## Desired behavior

`apiKey` on `llm()` should accept the same nested per-provider object that `SmolConfig["apiKey"]` and `agency.json` `client.apiKey` already use:

```ts
type ApiKey = {
  openAi?: string
  google?: string
  anthropic?: string
  ollama?: string
  openRouter?: string
  deepInfra?: string
  liteLlm?: string
  openAiCompat?: string
}
```

A bare `string` should remain accepted for convenience (treated as the OpenAI key) for backward compatibility.

## Scope of change

- Broaden the `apiKey` type in `lib/typeChecker/builtins.ts` from `string` to `string | ApiKey`.
- Update `toSmolConfig` in `lib/runtime/llmClient.ts` to merge an object-form `apiKey` into the smoltalk config (in addition to the existing string → `openAi` shorthand).
- Update the `llm()` options table in `docs/site/guide/llm.md` once the object form is supported (the doc currently documents `apiKey` as `string` because that's all that works today).

## Context

Found while updating `docs/site/guide/llm.md`. The docs originally described `apiKey` as a per-provider object, but that only matches the config-file / smoltalk layer — not the actual per-call `llm()` option, which is still string-only. This issue tracks closing that gap.

egonSchiele:
## Summary

Closes #405.

The per-call `apiKey` option on `llm()` only accepted a bare `string` (treated as the OpenAI key), so a caller who wanted to pass an Anthropic/Google/etc. key on a per-call basis couldn't — they had to fall back to `agency.json` `client.apiKey` or environment variables. This broadens `apiKey` to also accept the same nested per-provider object that `SmolConfig["apiKey"]` and `agency.json` `client.apiKey` already use.

## Changes

- **Typechecker** (`lib/typeChecker/builtins.ts`): `apiKey` is now typed `string | { openAi?, google?, anthropic?, ollama?, openRouter?, deepInfra?, liteLlm?, openAiCompat? } | null`.
- **Runtime** (`lib/runtime/llmClient.ts`): `PromptConfig.apiKey` broadened to `string | SmolConfig["apiKey"]`; `toSmolConfig` now merges an object-form key over the baked-in metadata map (per-provider override). The string form keeps its existing `openAi`-shorthand behavior.
- **Docs** (`docs/site/guide/llm-part-2.md`): documented the object form in the `llm()` options table.

The bare-string shorthand remains accepted for backward compatibility.

## Testing

- `tsc --noEmit` clean.
- Added a unit test to `lib/runtime/llmClient.test.ts` covering object-form merge semantics (baked-in `openAi` preserved, `anthropic` overridden, `google` added). All 7 tests in the file pass.
- Compiled a real Agency program: both `apiKey: { anthropic: "..." }` and `apiKey: "sk-..."` type-check and forward correctly to the runtime; a wrong shape (`anthropic: 123`) is rejected with a clear union-type diagnostic.

### DIFF
--- a/packages/agency-lang/docs/site/guide/llm-part-2.md
+++ b/packages/agency-lang/docs/site/guide/llm-part-2.md
@@ -69,14 +69,14 @@ You can pass an options object as the second parameter, or use named arguments. 
 |---|---|
 | `model` | `string` |
 | `provider` | `string` |
-| `apiKey` | `string \| { openAi?, google?, anthropic?, ollama?, openRouter?, deepInfra?, liteLlm?, openAiCompat? }` |
+| `apiKey` | `{ openAi?, google?, anthropic?, ollama?, openRouter?, deepInfra?, liteLlm?, openAiCompat? }` |
 | `maxTokens` | `number` |
 | `temperature` | `number` |
 | `reasoningEffort` | `"low" \| "medium" \| "high"` |
 | `thinking` | `{ enabled: boolean, budgetTokens?: number }` |
 | `stream` | `boolean` |
 
-`apiKey` sets the API key(s) for this call. A bare `string` is a shorthand for the OpenAI key. To supply keys for other providers per call, pass the per-provider object form — e.g. `apiKey: { anthropic: "sk-ant-..." }` — which uses the same shape as `agency.json`'s `client.apiKey`. Keys you don't set still fall back to `agency.json` `client.apiKey` and the `OPENAI_API_KEY` / provider-specific environment variables.
+`apiKey` sets the API key(s) for this call as a per-provider object — e.g. `apiKey: { anthropic: "sk-ant-..." }` — using the same shape as `agency.json`'s `client.apiKey`. A bare string is intentionally **not** accepted: it would have to guess a provider slot, silently misrouting (say) an Anthropic key into the OpenAI slot. Name the provider explicitly. Keys you don't set still fall back to `agency.json` `client.apiKey` and the `OPENAI_API_KEY` / provider-specific environment variables.
 
 ### Tools & context
 
--- a/packages/agency-lang/docs/site/guide/llm-part-2.md
+++ b/packages/agency-lang/docs/site/guide/llm-part-2.md
@@ -69,14 +69,14 @@ You can pass an options object as the second parameter, or use named arguments. 
 |---|---|
 | `model` | `string` |
 | `provider` | `string` |
-| `apiKey` | `string` |
+| `apiKey` | `string \| { openAi?, google?, anthropic?, ollama?, openRouter?, deepInfra?, liteLlm?, openAiCompat? }` |
 | `maxTokens` | `number` |
 | `temperature` | `number` |
 | `reasoningEffort` | `"low" \| "medium" \| "high"` |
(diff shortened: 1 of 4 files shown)


packages/agency-lang/docs/site/guide/llm-part-2.md
```diff
@@ -69,14 +69,14 @@ You can pass an options object as the second parameter, or use named arguments.
...
-| `apiKey` | `string` |
+| `apiKey` | `string \| { openAi?, google?, anthropic?, ollama?, openRouter?, deepInfra?, liteLlm?, openAiCompat? }` |
 | `maxTokens` | `number` |
 | `temperature` | `number` |
 | `reasoningEffort` | `"low" \| "medium" \| "high"` |
 | `thinking` | `{ enabled: boolean, budgetTokens?: number }` |
 | `stream` | `boolean` |
 
-`apiKey` is the OpenAI API key for this call. To configure keys for other providers, set them in `agency.json` under `client.apiKey` (a per-provider object) or via the `OPENAI_API_KEY` / provider-specific environment variables.
+`apiKey` sets the API key(s) for this call. A bare `string` is a shorthand for the OpenAI key. To supply keys for other providers per call, pass the per-provider object form — e.g. `apiKey: { anthropic: "sk-ant-..." }` — which uses the same shape as `agency.json`'s `client.apiKey`. Keys you don't set still fall back to `agency.json` `client.apiKey` and the `OPENAI_API_KEY` / provider-specific environment variables.
```
egonSchiele: Would it be possible to remove the shorthand entirely? Or derive from the specified provider, instead of defaulting to openai? A breaking change is fine

egonSchiele:
Went with removing the shorthand entirely (pushed in 65505f61). `apiKey` now accepts only the per-provider object, e.g. `apiKey: { anthropic: "sk-ant-..." }`; a bare string is a type error.

I looked at "derive from the specified provider" too, but it has an unavoidable gap: when `provider` isn't set on the call, smoltalk infers it from the model name *inside* smoltalk — that inference isn't available at agency's `toSmolConfig` layer. So a bare string with no explicit provider would still be ambiguous (either an `openAi` fallback — the footgun — or replicating smoltalk's model→provider inference). Object-only sidesteps that entirely and, per your note, the breaking change is fine.

PR #421, opened 2026-07-04
Support ${ escape in triple-quoted strings

### LINKED ISSUE #419: Triple-quoted raw strings interpolate ${...} (codegen emits into a JS template literal without escaping)
### Summary

A triple-quoted **raw** string interpolates `${...}` instead of keeping it literal. The docs describe `"""` as a raw multi-line string ("do **not** interpret backslash escapes"), but `${...}` inside one is still evaluated — because the string body is code-generated into a JavaScript **template literal** (backticks) without escaping `${` (or backticks), so JS re-interprets it at runtime.

### Minimal repro (no subprocess)

```
node main() {
  const x: number = 5
  const s = """literal dollar-brace ${x} should stay text"""
  return s
}
```

- Expected: `"literal dollar-brace ${x} should stay text"` (raw/literal).
- Actual: `"literal dollar-brace 5 should stay text"` — the `${x}` was interpolated.

Escaping doesn't help either: Agency `\${x}` compiles to JS `\\${x}` (a literal backslash followed by a JS interpolation), so there is currently no way to place a literal `${` in the string.

### Codegen evidence

`const s = """value ${counter} here"""` emits:

```js
__stack.locals.s = `value ${counter} here`;
```

The `${counter}` is passed through verbatim into a JS template literal, so at runtime JavaScript evaluates it — resolving a bare identifier (`ReferenceError: counter is not defined` when it maps to nothing, or the wrong value otherwise).

### Impact

Any triple-quoted string used to embed text/code that legitimately contains `${` breaks: shell snippets, JS/TS templates, and — how I hit it — **embedding Agency source for `std::agency run()`**. A nested program returning `"count:${n}"` inside a `"""..."""` block is interpolated by the *outer* program instead of reaching the child.

Workaround today: keep `${...}` out of triple-quoted string bodies (e.g. return a number and interpolate in the caller).

### Fix direction

When emitting an Agency raw (triple-quoted) string as a JS template literal, escape `${` → `\${` and backtick → `` \` `` in the literal content (or emit raw strings as a non-template quoted string). Interpolating quote types (`"`/`'`/`` ` ``) should keep escaping literal-`${` runs while still compiling genuine interpolations.

### Notes

Found while implementing subprocess callback forwarding (#417). It caused a misdiagnosis there — I initially attributed a nested-relay test failure to "compiled children can't hold top-level globals," which is **not** true (they can; verified). The actual cause was this raw-string interpolation.

egonSchiele:
## What

Adds `\${` → literal `${` as an escape inside triple-quoted (`"""..."""`) strings, so you can write a literal `${` in a raw string.

## Why

Triple-quoted strings support `${...}` interpolation but intentionally treat backslash as literal (so `\n`, Windows paths, and regex stay raw). That combination left **no way** to write a literal `${` inside one — which breaks embedding any text/code that legitimately contains `${`: shell/JS templates, or Agency source passed to `std::agency run()`. (This is the bug from #419, found while implementing subprocess callback forwarding.)

The docs claimed triple-quoted strings were raw with no interpolation — that was wrong; interpolation is intentional. The real gap was the missing escape.

## How

- **Parser** (`multiLineStringTextSegmentParser`): decode `\${` → literal `${`; every other backslash stays verbatim. This is the single text-segment parser shared by expression, tag-arg, and doc-string triple-quoted literals, so the fix lands uniformly. The JS printer already escapes `${` in text segments, so no JS-codegen change was needed.
- **Formatter** (`AgencyGenerator`): it emitted triple-quoted text raw, so after the parser change `fmt` would silently turn a parsed literal `${x}` back into a live interpolation. Now re-escapes `${` → `\${` when printing triple-quoted text and doc-string text (`escapeMultiLineText`).
- **Docs**: corrected `basic-syntax.md` — triple-quoted strings interpolate, and `\${` is the one escape.

## Behavior change

`\${` in a triple-quoted string now means a literal `${` (previously: a literal backslash followed by an interpolation). In-repo blast radius is **zero** — no `.agency` file across `lib/`, `tests/`, or `docs/` contains `\${` in a triple-quoted string. Regular strings (`"`/`'`/`` ` ``) are unchanged — they already supported `\${`.

Known edge: `\\${x}` in a triple-quoted string is a literal `\` followed by a literal `${x}` (only `\${` is special; `\\` isn't processed in raw strings). Rare.

## Tests

- Parser: escaped-only, raw `\n` stays raw alongside an escape, and real interpolation next to an escaped `${}`.
- Formatter: emit → re-parse round-trip for the same three shapes (guards the `fmt` regression above).
- Execution: `"""hi ${name} and literal \${skip}"""` → `hi world and literal ${skip}`, through parser → codegen → runtime.

Full unit suite (7119 tests), typecheck, and `lint:structure` pass; `fmt` round-trips the execution fixture unchanged.

Closes #419.

### DIFF
--- a/packages/agency-lang/lib/parsers/parsers.ts
+++ b/packages/agency-lang/lib/parsers/parsers.ts
@@ -450,13 +450,31 @@ const stringTextSegmentParserFor = (delim: '"' | "'" | "`"): Parser<TextSegment>
     return success({ type: "text" as const, value }, input.slice(i));
   };
 
-export const multiLineStringTextSegmentParser: Parser<TextSegment> = map(
-  many1Till(or(str('"""'), str("${"))),
-  (text) => ({
-    type: "text",
-    value: text,
-  }),
-);
+// Text segment inside a triple-quoted string. Raw by design — backslash escapes
+// like `\n` are NOT interpreted (a literal backslash then `n`) — with ONE
+// exception: `\${` decodes to a literal `${` so a raw string can contain literal
+// interpolation syntax (e.g. embedded Agency/shell/JS source) without opening an
+// interpolation. A lone backslash, and every other `\x`, stays verbatim. Stops
+// at the closing `"""` or an unescaped `${` (which starts interpolation).
+export const multiLineStringTextSegmentParser: Parser<TextSegment> = (
+  input: string,
+): ParserResult<TextSegment> => {
+  let i = 0;
+  let value = "";
+  while (i < input.length) {
+    if (input.startsWith('"""', i)) break;
+    if (input[i] === "\\" && input[i + 1] === "$" && input[i + 2] === "{") {
+      value += "${";
+      i += 3;
+      continue;
+    }
+    if (input[i] === "$" && input[i + 1] === "{") break;
+    value += input[i];
+    i++;
+  }
+  if (i === 0) return failure("expected string text", input);
+  return success({ type: "text" as const, value }, input.slice(i));
+};
 
 export const interpolationSegmentParser: Parser<InterpolationSegment> = withLoc((
   input: string,
--- a/packages/agency-lang/lib/parsers/parsers.ts
+++ b/packages/agency-lang/lib/parsers/parsers.ts
@@ -450,31 +450,25 @@ const stringTextSegmentParserFor = (delim: '"' | "'" | "`"): Parser<TextSegment>
     return success({ type: "text" as const, value }, input.slice(i));
   };
 
+// One character of raw multi-line text: any char that does not begin the close
+// delimiter (`"""`) or an interpolation (`${`). `not(...)` is a zero-width
+// negative lookahead; `capture(anyChar)` then consumes the single char.
+const multiLineRawChar: Parser<string> = map(
+  seqC(not(or(str('"""'), str("${"))), capture(anyChar, "c")),
+  (r) => r.c,
+);
+
 // Text segment inside a triple-quoted string. Raw by design — backslash escapes
 // like `\n` are NOT interpreted (a literal backslash then `n`) — with ONE
-// exception: `\${` decodes to a literal `${` so a raw string can contain literal
-// interpolation syntax (e.g. embedded Agency/shell/JS source) without opening an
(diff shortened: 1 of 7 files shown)


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -450,13 +450,31 @@ const stringTextSegmentParserFor = (delim: '"' | "'" | "`"): Parser<TextSegment>
...
+// exception: `\${` decodes to a literal `${` so a raw string can contain literal
+// interpolation syntax (e.g. embedded Agency/shell/JS source) without opening an
+// interpolation. A lone backslash, and every other `\x`, stays verbatim. Stops
+// at the closing `"""` or an unescaped `${` (which starts interpolation).
+export const multiLineStringTextSegmentParser: Parser<TextSegment> = (
+  input: string,
+): ParserResult<TextSegment> => {
+  let i = 0;
+  let value = "";
+  while (i < input.length) {
```
egonSchiele: This parser is poorly written. It is not going to be performative and is using a completely different style from all the other parsers in this file. All the other parsers use the Tarsec library and write their parsers in a declarative style. Please see how those parsers are written and write your parser accordingly.

egonSchiele: Rewritten declaratively in 2c5b285d:

```ts
const multiLineRawChar: Parser<string> = map(
  seqC(not(or(str('"""'), str("${"))), capture(anyChar, "c")),
  (r) => r.c,
);

export const multiLineStringTextSegmentParser: Parser<TextSegment> = map(
  many1WithJoin(or(dollarBraceEscape, multiLineRawChar)),
  (value) => ({ type: "text", value }),
);
```

`dollarBraceEscape` already existed (it's what regular strings use for `\${`), and the `not(...) + capture(anyChar)` idiom for "one char that isn't a terminator" matches `hexByteEscape` in the same file. Per-char terminator lookahead is the same shape as the original `many1Till(or(str('"""'), str("${")))` it grew out of. Parser, generator round-trip, and execution tests are unchanged and green (full parser suite: 1513).

PR #422, opened 2026-07-04
Fix #317: allow comments between array/object literal entries

### LINKED ISSUE #317: Parser: // comment between object-literal entries fails to parse (and mis-reports the location)
## Summary

A `//` line comment placed **between entries of an object literal** is a parse error. The same comment is fine above the enclosing statement, and `//` comments work everywhere else, so this is surprising. The diagnostic also points at the wrong location (the start of the declaration), which makes it hard to find.

## Reproduction

```ts
// repro.agency
export static const policy = {
  "a": [{ action: "approve" }],
  // this comment between entries breaks the parse
  "b": [{ action: "reject" }]
}
```

Compiling fails:

```
$ pnpm run agency compile repro.agency
Failed to parse .../repro.agency: Line 1, col 46: expected whitespace, "break",
"continue", "import", "export", a node definition, ... , "//", ... or end of input
```

Note the reported location (`Line 1`) is the `const` declaration, not the comment on line 3 — the parser appears to terminate the object literal early and then report a top-level-statement error far from the real site.

## Workaround

Move the comment above the enclosing declaration (or onto its own line outside the `{ ... }`):

```ts
// this comment is fine here
export static const policy = {
  "a": [{ action: "approve" }],
  "b": [{ action: "reject" }]
}
```

## Expected

`//` (and presumably `/* */`) comments should be allowed between object-literal entries, the same as between statements — and if a comment genuinely isn't allowed somewhere, the error should point at the comment, not the start of the declaration.

## Notes

- Found while editing `lib/agents/agency-agent/lib/defaultPolicy.agency` (PR #316). Worked around by hoisting the comment.
- Comments between array elements may have the same issue — worth checking as part of a fix.

egonSchiele:
Fixes #317.

## Problem

A `//` or `/* */` comment placed **between entries of an array or object literal** was a parse error, even though comments work between statements and between object *type* properties:

```
const policy = {
  "a": [{ action: "approve" }],
  // this comment between entries breaks the parse
  "b": [{ action: "reject" }]
}
```

### Root cause

The literal parsers separated entries with `sepBy(commaWithNewline, …)`, and the whitespace helpers (`optionalSpacesOrNewline`, `commaWithNewline`) consume only whitespace — never comments. A comment stalled the separator, terminated the literal early, and re-surfaced as a confusing top-level statement error far from the comment.

Object *types* already supported inline comments via an `objectTrivia` mechanism; this PR brings the same capability to literals.

## Fix (lossless — comments are preserved, not dropped)

- **AST** (`lib/types/dataStructures.ts`): new `Trivia` type; optional `trivia?: Trivia[]` on `AgencyArray` / `AgencyObject`. Omitted when there are no comments, so existing consumers are unaffected.
- **Parser** (`lib/parsers/parsers.ts`): replaced `sepBy` with an interleaved `many(or(literalTrivia, literalItemEntry(…)))` plus a shared `splitLiteralEntries` that anchors each comment to the index of the entry it precedes (trailing trivia anchored at `length`). Whitespace-around-comma behavior matches the old `commaWithNewline` exactly.
- **Formatter** (`lib/backends/agencyGenerator.ts`): extracted a shared `emitTriviaAt` helper — now used by object types **and** both literals — and taught `processAgencyArray` / `processAgencyObject` to emit trivia. Comment-free arrays keep the compact single-line formatting.

Arrays had the identical bug (the issue's "worth checking" note) and are fixed too.

## Testing

- **TDD**: failing parser + formatter tests written first, confirmed to fail on the parse error, then made to pass.
- New tests: 6 parser (`dataStructures.test.ts`) + 4 formatter incl. an **idempotency** fixed-point test (`agencyGenerator.test.ts`).
- Full vitest unit suite passes (7160 tests).
- End-to-end: the exact issue repro compiles and `fmt` preserves the comment; nested array-of-objects, comment-inside-inner-array, splat + comment, block comments, and trailing commas all round-trip losslessly and idempotently.
- `tsc` and structural linter clean.

### DIFF
--- a/packages/agency-lang/lib/parsers/parsers.ts
+++ b/packages/agency-lang/lib/parsers/parsers.ts
@@ -127,6 +127,7 @@ import {
   AgencyObject,
   AgencyObjectKV,
   SplatExpression,
+  Trivia,
 } from "../types/dataStructures.js";
 import {
   DefaultImport,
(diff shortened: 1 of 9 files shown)


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2019,31 +2020,111 @@ export const splatParser: Parser<SplatExpression> = seqC(
   capture(lazy(() => exprParser), "value"),
 );
 
+// Trivia (blank line / line comment / block comment) that may appear between
+// items of an array or object literal, plus any trailing whitespace so the
+// cursor lands on the next item. Mirrors `objectTrivia` (used by object
+// *types*) so literals preserve comments the same way and `fmt` round-trips.
+type TriviaNode = AgencyComment | AgencyMultiLineComment | NewLine;
```
egonSchiele, line 2027: Is this the right place to put these types or do they need to go into one of the types files?

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2019,31 +2020,111 @@ export const splatParser: Parser<SplatExpression> = seqC(
...
+  input: string,
+): ParserResult<{ kind: "trivia"; node: TriviaNode }> => {
+  const parser = seqC(
+    capture(or(blankLineParser, commentParser, multiLineCommentParser), "node"),
+    optionalSpacesOrNewline,
+  );
+  const result = parser(input);
+  if (!result.success) return result;
+  return success(
+    { kind: "trivia" as const, node: result.result.node as TriviaNode },
```
egonSchiele, line 2043: You could use the `set` parser to set the kind, and then you wouldn't need to unwrap the result and check it at all.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2019,31 +2020,111 @@ export const splatParser: Parser<SplatExpression> = seqC(
...
+    return success(
+      { kind: "item" as const, item: result.result.item as T },
+      result.rest,
+    );
+  };
+
+// Split an interleaved item/trivia list into the item array plus
+// anchor-indexed trivia. Trivia is anchored to the index of the item that
+// follows it; trailing trivia is anchored at `items.length`.
+function splitLiteralEntries<T>(entries: LiteralEntry<T>[]): {
```
egonSchiele, line 2074: Do you need this or can you use one of the combinators like `many1`? Or is there a reusable function you can use or write?

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2089,21 +2170,31 @@ export const agencyObjectKVParser: Parser<AgencyObjectKV> = (
...
-  ),
-  optional(char(",")),
-  optionalSpacesOrNewline,
-  char("}"),
-);
+    optionalSpacesOrNewline,
+    char("}"),
+  );
+
+  const result = parser(input);
```
egonSchiele, line 2189: Again, this code is mostly good, and I'm glad you're using the Tarsec library. But you have some imperative code in here that I think becomes unnecessary if you use Tarsec's combinators properly.

packages/agency-lang/lib/runtime/llmClient.ts
```diff
@@ -236,6 +239,8 @@ export function toSmolConfig(config: PromptConfig): Omit<SmolConfig, "stream"> {
     hostedTools,
     ...(typeof apiKey === "string"
       ? { apiKey: { ...metaApiKey, openAi: apiKey } }
+      : apiKey
```
egonSchiele, line 242: Note that you have changes from another PR mixed into this PR.

egonSchiele:
Superseded by #423, which is based on a clean `main`.

This branch was accidentally stacked on #420's commit (`llm(): accept per-provider object for per-call apiKey`), which is why those changes showed up here. Rather than force-push, I moved the fix to a fresh branch off `main`.

All review feedback is addressed in #423:
- **`set` parser / no manual unwrapping**: `triviaEntry` and the `itemEntry(...)` wrapper build tagged entries with `set(...)`; the literal parsers use `map(...)` instead of hand-rolled `parser(input); if (!success)` plumbing.
- **Reusable function instead of custom imperative split**: extracted a single `partitionTrivia` fold now shared by object literals, array literals, and `objectTypeParser` (replacing its duplicated split loop).
- **Types placement**: `TriviaNode` moved to `types/dataStructures.ts`; parse-time tagging types stay local to the parser.
- **Unrelated changes**: gone — #423 touches only the 5 parser/formatter/type/test files.

Both `//` and `/* */` comment styles are supported and covered by tests.

PR #424, opened 2026-07-04
Allow `goto` in match arms for node dispatch (#415)

### LINKED ISSUE #415: Support goto match(...) for node dispatch (mirroring return match(...))
## Summary

Today a `match` expression arm cannot yield a graph-node call — a node call compiles to a control-flow transition (goto/halt), not a value, so it can't be a match-expression's value. The current workaround is an `if`/`else` chain for node dispatch. See `docs/site/guide/pattern-matching.md` ("v1 restrictions").

We should let users put `goto` in front of a `match` to dispatch to a node, exactly the way `return` in front of a `match` returns the matched value from a function:

```ts
goto match(x) {
    "next" => next
    _      => end
}
```

This desugars to "evaluate the match to pick a node target, then transition to it" — the node-dispatch analogue of `return match(...)`.

## Motivation

- **Symmetry with `return match(...)`.** Agency already supports `return match(...)` (put `return` in front of the match; each arm yields the returned value). `goto match(...)` is the same shape for control-flow transitions. Users will reasonably expect it to work.
- **Removes boilerplate.** Node dispatch currently requires an `if`/`else` chain; a `match` is more concise and gets exhaustiveness checking.

## Why it's blocked today

The restriction exists because a bare node call inside a match *arm* compiles to a goto/halt transition, which is not a value the match can yield. The proposal sidesteps that: the `goto` prefix form makes the whole match a **transition target selector**, not a value-producing expression — the arms name node targets, and the single leading `goto` performs the transition once a target is chosen. So arms don't need to "yield a node as a value"; they select which node the outer `goto` jumps to.

## Design questions

1. **Arm surface syntax.** Arms name a node target (`next`, `end`). Should this reuse the bare-node-reference form, or require `goto next` inside each arm? The prefix form (`goto match(...) { "next" => next }`) reads best and keeps a single transition point.
2. **Exhaustiveness.** Like a `match` expression, a `goto match(...)` must always pick a target, so it should require exhaustive arms or a `_` catch-all (a fall-through with no transition would be a bug).
3. **Interaction with the existing "`goto` is statically rejected inside any block" rule** (`docs/site/guide/guards.md`). The prefix `goto match` is at statement position in a node body, not nested inside a guard/other block, so it should be fine — but confirm the check keys off block context, not the presence of `goto` near a match.
4. **`match(x is pattern)` form** is statement-only and would compose naturally here; confirm it's in scope or explicitly deferred.

## Related

- `docs/site/guide/pattern-matching.md` ("v1 restrictions") — the bullet to remove/relax once this lands.
- Part of the broader `match` expression feature work.

egonSchiele:
## Summary

Closes #415 (per-arm form).

A `match` arm can now `goto` a graph node directly, so node dispatch can be written as a `match` instead of an `if`/`else` chain:

```ts
node dispatch(x: string) {
  match(x) {
    "a" => goto landA()        // bare arm
    "b" => {                    // block arm — mix freely
      prepare()
      goto landB()
    }
    _   => goto landC()
  }
}
```

Both the **bare** (`"a" => goto landA()`) and **block** (`"b" => { goto landB() }`) forms are supported and can be mixed. Each arm performs a real node transition, not a value yield.

## Why this is small

The block form already parsed and compiled correctly; only the *bare* `goto` arm was rejected, because the non-braced arm RHS accepted only return/assignment/expression. The change is one line: add `gotoStatementParser` to that alternative set. A bare `goto` arm then produces the same AST as today's block arm, and everything downstream (statement-match lowering → nested `if`/`else` emitting `goToNode(...)`, adjacency/`conditionalEdge` registration) is reused unchanged. No lowering, codegen, or exhaustiveness changes.

## Design decisions

- **Not exhaustive.** A `goto`-dispatching match is an ordinary statement match, so an unmatched value falls through and execution continues after the `match` (matches the discussion on #415: exhaustiveness would only be forced by a `goto match(...)` prefix form).
- **Prefix `goto match(...)` deliberately skipped.** It is brittle under change (one arm needing a block forces "yield a node target" semantics or a full rewrite) and offers no expressiveness the per-arm form lacks. The per-arm form scales locally — wrap an arm in braces and the `goto` comes along. Docs updated to describe the per-arm form and mark only the prefix form as unsupported.

## Testing

- **Parser unit test** (`lib/parsers/matchBlock.test.ts`): a bare `x => goto next()` arm parses to a `gotoStatement` body. Watched it fail before the fix, pass after.
- **Execution test** (`tests/agency/goto-match-arm.agency`): a `dispatch` node routes to three different target nodes via a bare arm, a block arm, and the wildcard arm; each asserts the correct terminal node is reached. All 3 pass.
- Full unit suite: **7130 passed, 0 failures** (incl. the 121 typecheck fixtures).

### DIFF
--- a/packages/agency-lang/docs/site/guide/pattern-matching.md
+++ b/packages/agency-lang/docs/site/guide/pattern-matching.md
@@ -271,36 +271,10 @@ Here are the places you can't use match blocks right now:
 - At the module level (outside of a function or node).
 - Inside `parallel`, `seq`, `thread`, and `subthread` blocks.
 
-### Dispatching to a node from a match arm
-
-Inside a node, a match arm can `goto` another node. Each arm performs a
-real node transition — it does not yield a value — so a `goto`-dispatching
-match must be a statement, not the value of a `return` or assignment. Arms
-may be written bare or as a block, and the two forms can be mixed:
+You can't use `goto` with match blocks:
 
 ```ts
-node dispatch(x: string) {
-  match(x) {
-    "a" => goto landA()
-    "b" => {
-      prepare()
-      goto landB()
-    }
-    _   => goto landC()
-  }
-}
-```
-
-Because it is an ordinary statement match, a `goto`-dispatching match is
-not required to be exhaustive: an unmatched value simply falls through and
-execution continues after the `match`.
-
-The **prefix** form, where a single `goto` precedes the whole `match` and
-arms name a bare node target, is not supported — put the `goto` inside each
-arm instead:
-
-```ts
-// not allowed — no `goto match(...)` prefix form
+// not allowed
 goto match(x) {
   "next" => next
   _ => end
--- a/packages/agency-lang/docs/site/guide/pattern-matching.md
+++ b/packages/agency-lang/docs/site/guide/pattern-matching.md
@@ -271,10 +271,36 @@ Here are the places you can't use match blocks right now:
 - At the module level (outside of a function or node).
 - Inside `parallel`, `seq`, `thread`, and `subthread` blocks.
 
-You can't use `goto` with match blocks:
+### Dispatching to a node from a match arm
+
+Inside a node, a match arm can `goto` another node. Each arm performs a
+real node transition — it does not yield a value — so a `goto`-dispatching
+match must be a statement, not the value of a `return` or assignment. Arms
+may be written bare or as a block, and the two forms can be mixed:
 
 ```ts
-// not allowed
+node dispatch(x: string) {
+  match(x) {
+    "a" => goto landA()
+    "b" => {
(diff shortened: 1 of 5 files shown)


packages/agency-lang/docs/site/guide/pattern-matching.md
```diff
@@ -271,10 +271,36 @@ Here are the places you can't use match blocks right now:
 - At the module level (outside of a function or node).
 - Inside `parallel`, `seq`, `thread`, and `subthread` blocks.
 
-You can't use `goto` with match blocks:
+### Dispatching to a node from a match arm
```
egonSchiele: undo this docs change please

Issue #413, opened 2026-07-03
Investigate: can a match-arm return work inside a parallel block?


egonSchiele:
## Summary

Investigate whether a `return` inside a `parallel { }` block should be allowed to yield a value from an enclosing `match` expression arm (today it is a hard compile error).

## Background

- A `return` inside a `fork(...) { }` branch already works — it returns from the branch, and `fork` collects the per-branch results. `parallel` desugars to `fork` calls before lowering (`lib/preprocessors/parallelDesugar.ts`), so the underlying branch-return machinery is already exercised.
- The match-yield unwind, however, is a **per-`Runner` scalar flag** (`_matchExit` in `lib/runtime/runner.ts`). The flag's own doc comment warns it "must never be set from a parallel/fork child ... this scalar would race." That is the crux of why `parallel` is currently forbidden.

## Questions to answer

1. **Semantics.** `parallel` runs N branches concurrently. If more than one branch does `return X`, which value becomes the match result? Is a match-yield from a parallel branch even coherent, or should `return` there keep its "branch produces X" meaning (as in `fork`) and simply *not* be a match-yield?
2. **Race safety.** The scalar `_matchExit` is set on a per-runner basis. Can a branch-local exit be collected and reconciled after the `parallel`/`fork` join without racing? (Contrast the `thread` case, where there is exactly one inline child and no race — see companion `thread` ticket.)
3. **Is this even desirable,** or is the `fork`-style "branch return, not match-yield" behavior the right and sufficient answer for `parallel`?

## Related

- Companion: "can `seq` support a match-arm return?"
- Companion (already actionable): allow `return` to yield a match value from `thread`/`subthread` — safe because thread is not concurrent.
- `docs/site/guide/pattern-matching.md` ("v1 restrictions").

PR #425, opened 2026-07-04
Allow match-arm return to yield from standalone seq and thread/subthread (#414)

### LINKED ISSUE #414: Investigate: can a match-arm return work inside a seq block?
## Summary

Investigate whether a `return` inside a `seq { }` block should be allowed to yield a value from an enclosing `match` expression arm (today it is a hard compile error).

## Background

- `seq` desugars to `fork` calls before lowering (`lib/preprocessors/parallelDesugar.ts`), and a `return` inside a `fork` branch already works as a branch return. So the branch-return machinery `seq` builds on already permits returns.
- The blocker is the match-yield unwind: it is a **per-`Runner` scalar flag** (`_matchExit` in `lib/runtime/runner.ts`) consumed by the owning frame between steps. A `seq` branch body runs under a child runner, so a `return` lowered to the child's `exitMatch` would set the flag on the child, and the parent match owner would never see it.

## Why `seq` may be more tractable than `parallel`

`seq` runs its branches **sequentially**, not concurrently. So the "scalar `_matchExit` would race" concern that rules out `parallel` is weaker or absent — there is a well-defined order and only one branch active at a time. The open questions are:

1. **Semantics.** In a `seq` of N branches, does a `return` in an earlier branch yield the match value (and short-circuit the rest of the `seq`), or does it just end that branch? What is the least-surprising rule?
2. **Mechanism.** Since branches run one at a time, can the child runner's `_matchExit` be forwarded to the parent after each branch — similar to the fix proposed for `thread`/`subthread` — without a race?
3. **Consistency.** Should `seq` behave like `thread` (allowed, inline forward) or like `parallel` (kept as a branch-return, not a match-yield)?

## Related

- Companion: "can `parallel` support a match-arm return?"
- Companion (already actionable): allow `return` to yield a match value from `thread`/`subthread`.
- `docs/site/guide/pattern-matching.md` ("v1 restrictions").

egonSchiele:
## Summary

Closes #414 and folds in the `thread`/`subthread` companion.

A `return` inside a match **expression** arm can now yield the match value when it sits in a **standalone `seq { }`** block or a **`thread` / `subthread`** block:

```ts
node dispatch(x: string): string {
  const r = match(x) {
    "seq"    => { seq    { let v = "from-seq";    return v } }
    "thread" => { thread    { return "from-thread" } }
    "sub"    => { subthread { return "from-subthread" } }
    _        => "default"
  }
  return "got: ${r}"
}
```

Previously all three were a hard compile error: *"cannot return from a match arm inside a parallel, fork, race, seq, or thread block."*

## Why these are safe (and `parallel` is not)

The whole thing turns on one fact: `_matchExit` is a **per-`Runner` scalar**, so the yield unwind only works when `exitMatch(matchId)` and the owning `ifElse({matchId})` run on the **same Runner**.

- **Standalone `seq`** desugars to *nothing* — its body is spliced inline into the enclosing body (`parallelDesugar.ts`), so there is no child Runner. A `return` there is exactly as safe as in a plain block arm.
- **`thread` / `subthread`** bodies run **inline on the parent Runner** (`runner.thread` → `callback(this)`), so again the flag is set and consumed on one Runner — no race.
- **`parallel`** lifts each branch into a separate `fork` frame with its **own child Runner**, and branches run concurrently — the scalar would race. It stays rejected (message narrowed to "parallel or fork block"). A `return` inside a `seq` arm *of* a `parallel` is unaffected: it remains a branch-local `fork` result, not a match-yield.

## Implementation

One file of logic (`lib/lowering/patternLowering.ts`):
- `rewriteReturnsToYields` recurses into `seqBlock` / `messageThread` bodies (rewriting `return` → `matchYield`) instead of rejecting them; the reject branch is narrowed to `parallelBlock`.
- `alwaysYields` sees through `seq`/`thread`/`subthread` for the all-paths-yield check (they run unconditionally and inline).
- Removed the now-unused `isConcurrencyBlock` / `concurrencyBlockBody` helpers; annotated the shared `returnFlowBodies` boundary to note the carve-out is applied explicitly, not there.

The **statement-arm** path is deliberately untouched — a `return` inside a thread block in a *statement* match arm still errors loudly (there the arm value is discarded, so it would silently exit the enclosing function).

## Testing

- **Lowering unit tests** (`patternLowering.matchExpr.test.ts`): standalone `seq`, `thread`, and `subthread` each rewrite their `return` to a `matchYield` in the right block; a non-yielding `seq` arm still errors; `parallel` and `seq`-in-`parallel` stay rejected. Watched the new cases fail before the change, pass after.
- **Execution test** (`tests/agency/match-yield-seq-thread.agency`): a `dispatch` node yields through a standalone `seq`, a `thread`, and a `subthread`, plus the wildcard arm — all 4 assert the right runtime value. This is the real proof the `_matchExit` unwind reaches its owner across the thread callback boundary.
- Full unit suite: **7136 passed, 0 failures**.

## Related

- #413 (`parallel`) — recommend closing as by-design; this PR keeps that rejection and narrows its message. Happy to open a follow-up that improves the error text / docs if wanted.
- A short guide note documenting the newly-allowed forms could be added to `pattern-matching.md`; left out here so you can decide the wording.

### DIFF
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -491,12 +491,30 @@ class PatternLowerer {
         out.push(mapBodies(stmt, (b) => this.rewriteReturnsToYields(b, matchId)));
         continue;
       }
-      // Opaque node. A `return` inside a concurrency block would otherwise
-      // survive as a raw function return the `_matchExit` unwind can't reach, so
-      // reject it explicitly.
-      if (isConcurrencyBlock(stmt) && containsReturn(concurrencyBlockBody(stmt))) {
+      // A standalone `seq` inlines into the enclosing body, and `thread` /
+      // `subthread` bodies run inline on this same runner, so a `return` inside
+      // them yields into THIS match — the `_matchExit` lands on the arm's own
+      // runner. Recurse the rewrite into their bodies. (A `seq` that is an arm
+      // of a `parallel` is a concurrent `fork` branch and is never reached here:
+      // we never descend into the `parallelBlock` below.)
+      if (stmt.type === "seqBlock" || stmt.type === "messageThread") {
+        out.push({
+          ...stmt,
+          body: this.rewriteReturnsToYields(
+            (stmt as { body: AgencyNode[] }).body,
+            matchId,
+          ),
+        });
+        continue;
+      }
+      // A `parallel` branch is lifted into a separate `fork` frame whose child
+      // runner the `_matchExit` unwind can't reach, and concurrent branches
+      // would race the scalar flag — so a `return` that would yield this match
+      // is rejected. (A `return` inside a `seq` arm of the parallel is a
+      // branch-local result, not seen by `containsReturn`, and stays legal.)
+      if (stmt.type === "parallelBlock" && containsReturn(stmt.body)) {
         throw new LoweringError(
-          "cannot return from a match arm inside a parallel, fork, race, seq, or thread block",
+          "cannot return from a match arm inside a parallel or fork block",
           stmt.loc,
         );
       }
@@ -513,6 +531,15 @@ class PatternLowerer {
   private alwaysYields(body: AgencyNode[]): boolean {
     for (const stmt of body) {
       if (stmt.type === "matchYield") return true;
+      // A standalone `seq` and a `thread`/`subthread` body run unconditionally
+      // and inline on this runner, so the arm yields on every path iff their
+      // body does.
+      if (
+        (stmt.type === "seqBlock" || stmt.type === "messageThread") &&
+        this.alwaysYields((stmt as { body: AgencyNode[] }).body)
+      ) {
+        return true;
+      }
       if (
         stmt.type === "ifElse" &&
         stmt.elseBody &&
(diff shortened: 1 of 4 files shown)


packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -491,12 +491,30 @@ class PatternLowerer {
...
-      // survive as a raw function return the `_matchExit` unwind can't reach, so
-      // reject it explicitly.
-      if (isConcurrencyBlock(stmt) && containsReturn(concurrencyBlockBody(stmt))) {
+      // A standalone `seq` inlines into the enclosing body, and `thread` /
+      // `subthread` bodies run inline on this same runner, so a `return` inside
+      // them yields into THIS match — the `_matchExit` lands on the arm's own
+      // runner. Recurse the rewrite into their bodies. (A `seq` that is an arm
+      // of a `parallel` is a concurrent `fork` branch and is never reached here:
+      // we never descend into the `parallelBlock` below.)
+      if (stmt.type === "seqBlock" || stmt.type === "messageThread") {
```
egonSchiele, line 500: what if there's a nested thread block, or a seq nested in a thread block, or vice versa?

packages/agency-lang/tests/agency/match-yield-seq-thread.agency
```diff
@@ -0,0 +1,25 @@
+// A `return` inside a standalone `seq` block or a `thread`/`subthread` block
+// yields the value of an enclosing `match` expression arm. These blocks run
+// inline on the arm's own runner, so the match-exit unwind reaches its owner.
+node dispatch(x: string): string {
```
egonSchiele, line 6: add tests for nested cases

egonSchiele: We should also have tests for other block types if we don't already, such as while loops, for loops, if statements and nested statements. What about a match nested inside another match?

egonSchiele:
Addressed in dbffa1d5 — added both lowering unit tests and execution tests (12 cases now).

**On nesting correctness (re `patternLowering.ts:500`, "nested thread, seq-in-thread or vice versa"):** `rewriteReturnsToYields` recurses into `seqBlock`/`messageThread` bodies *unconditionally*, so arbitrarily deep nesting of `seq`/`thread`/`subthread` is handled — each level is descended and the innermost `return` becomes the `matchYield`. It stays safe because none of these introduces a child `Runner`: a standalone `seq` inlines away and thread bodies run inline via `callback(this)`, so no matter how deep, the `exitMatch` lands on the arm's own runner. `parallel` remains opaque at every level (we never descend into a `parallelBlock`), so a `parallel` nested inside a `seq` is still rejected — there's a test for that too.

**Coverage added:**
- Nested concurrency: `thread`-in-`thread`, `seq`-in-`thread`, `thread`-in-`seq`, `seq`-in-`seq` (unit asserts the `matchYield` lands in the innermost body; execution asserts the runtime value).
- Control flow: a `seq`/`thread` return inside `if`, `for`, and `while` — including that the yield correctly unwinds *through* the loop (the loop breaks on `_matchExit`, the trailing statement is skipped).
- Nested match: a `thread` return in an **inner** match arm yields the inner match, which is then returned to the outer match.
- Negative: a `parallel` nested inside a standalone `seq` still errors.

All 12 execution cases pass and the full unit suite is green (7140).

PR #426, opened 2026-07-04
Allow match expressions in module-level initializers (#410)

### LINKED ISSUE #410: Allow match expressions in module-level const initializers
## Summary

Module-level `const x = match(...) { ... }` initializers are currently a compile error. You have to wrap the match in a `def` and call it from the initializer instead:

```ts
// not allowed today
const label = match(env) {
    "prod"  => "Production"
    "stage" => "Staging"
    _       => "Local"
}

// current workaround
def computeLabel(): string {
    return match(env) {
        "prod"  => "Production"
        "stage" => "Staging"
        _       => "Local"
    }
}
const label = computeLabel()
```

`match` is an expression and works fine as a `const` RHS inside functions, so requiring this wrapper at module level is an ergonomic wart. We should allow match expressions directly in module-level initializers.

## Why it's currently disallowed

Per `docs/site/guide/pattern-matching.md` and the init machinery: module-level initializers are planned one expression per variable by the init-topsort machinery, but a lowered match region is multiple statements. The lowering, not the language semantics, is the blocker.

## Proposed

Allow `const x = match(...)` at module level. Likely approaches:

- Teach the init-topsort machinery to accept a lowered multi-statement region for a single variable's initializer (run the region, then bind the yielded value), or
- Auto-hoist the match into a synthesized initializer function transparently (i.e. do the `def` workaround automatically) so the dependency graph still sees one expression per variable.

## Note (not the same as `if`)

For reference: this is specifically a `match` limitation. Agency has no `if`-expression or ternary (see `docs/site/guide/basic-syntax.md`), so `const x = if (...) { ... }` is invalid everywhere, not just at module level. This ticket is only about lifting the module-level restriction on `match` expressions.

egonSchiele:
## Summary

Closes #410. `const x = match(...) { ... }` at module level (and `let` / `static const`) is now allowed instead of a compile error.

```ts
const env = "prod"

const label = match(env) {       // previously an error
    "prod"  => "Production"
    "stage" => "Staging"
    _       => "Local"
}
```

## Approach: auto-hoist (the algorithm needs no change)

A `match` compiles to *several* statements, but the module-init machinery carries exactly **one expression per variable**. Rather than teach that machinery to hold a statement region (invasive: three hard-wired spots + dependency re-plumbing), pattern lowering rewrites the initializer into the manual workaround people write today:

```ts
const x = match(E) { ... }
  =>  def matchInit$N() { <lowered match region>; return __matchval_N }
      const x = matchInit$N()
```

Now `const x = matchInit$N()` is an ordinary single-expression initializer. Crucially, the init dep graph's existing **depth-1 call expansion** already walks a called function's body to attribute its reads as dependencies of the initializer — so `x` still correctly depends on whatever the scrutinee and arms reference. The topsort, dependency graph, and emit are all untouched. This is the exact shape of the already-tested `tests/agency/topsort/depth1-callgraph/static-via-function/` fixture.

## Two implementation details worth calling out

- **Synthesized name = `matchInit$N`.** The `$` can't appear in an Agency identifier (`[A-Za-z0-9_]` only), so it's collision-proof against user code. And it must **not** be `__`-prefixed: the codegen emits any `__`-prefixed call as a plain direct JS call (`matchInit$N()`), but a `def` compiles to an `AgencyFunction` *object*, so a `__`-name would crash with `... is not a function`. A non-`__` name routes through `__call` dispatch. (Found this the hard way while validating.)
- **`returnType: null`** — verified a `def` with no return type infers fine, so no return-type synthesis is needed.

## Correctness: every init case behaves like the equivalent non-match code

Because the transform emits the hand-written workaround, it inherits the exact init semantics. Verified end-to-end:

- ✅ **const = match**, **let = match**, **static const = match** — all run (`tests/agency/module-level-match.agency`, 4 cases).
- ✅ **A global depending on a match-initialized global** (`banner` ← `label` ← `env`) — ordered correctly.
- ✅ **Use-before-def** (match reads a later-declared var) → correct *"declared later in the same file"* error.
- ✅ **Two module-level matches that depend on each other** → correct *"Circular global dependency"* error.
- ✅ **static match reading a global** → correct static→global cross-phase error.
- ✅ **`match(x is …)`** at module level → still the correct "cannot be used as an expression" error.

The three negative cases live in `topsortCycleErrors.test.ts` (proving the hoist preserves dependency detection); the positive ones are an `agency test` fixture.

## Testing

- Lowering unit tests: module-level `const`/`let` match hoists to a synthesized fn + call; name is collision-proof and non-`__`.
- Init-graph tests: use-before-def, mutual cycle, and cross-phase errors all still fire through the hoist.
- Execution fixture: 4 runtime cases (const, dependent-ordering, let, static).
- Full unit suite: **7145 passed, 0 failures**.

## Note on docs

`docs/site/guide/pattern-matching.md`'s v1-restrictions list still says match can't be used *"At the module level (outside of a function or node)"* — that line is now inaccurate. I left the guide untouched so you can decide the wording; happy to update it in this PR if you'd like.

### DIFF
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -16,6 +16,7 @@ import type {
   Expression,
   ForLoop,
   FunctionCall,
+  FunctionDefinition,
   HandleBlock,
   IfElse,
   MatchBlock,
@@ -239,13 +240,45 @@ class PatternLowerer {
     // hoisted region to a self-contained async IIFE there (plain mode), which
     // never sets the runner's `_matchExit` flag.
     if (node.value && (node.value as AgencyNode).type === "matchBlock") {
-      if (this.atModuleLevel) {
-        throw new LoweringError(
-          "match expressions are not supported in module-level initializers",
-          node.loc,
-        );
-      }
       const region = this.lowerMatchExpressionCore(node.value as MatchBlock, node.loc);
+      if (this.atModuleLevel) {
+        // A module-level initializer has no execution frame for the match's
+        // `_matchExit` unwind, so we can't splice the region inline the way we
+        // do inside a function. Instead hoist the region into a synthesized
+        // init function and rewrite the initializer to call it. The result is
+        // exactly the manual `def`-wrapper workaround:
+        //
+        //   const x = match(E) { ... }
+        //     =>  def matchInit$N() { <region>; return __matchval_N }
+        //         const x = matchInit$N()
+        //
+        // `const x = matchInit$N()` is then an ordinary single-expression
+        // initializer that the init-topsort machinery already handles: its
+        // depth-1 call expansion walks the synthesized function's body, so `x`
+        // still depends on whatever the scrutinee and arms read. The `$` cannot
+        // appear in a user identifier (Agency names are [A-Za-z0-9_]), so the
+        // name never collides; and NOT starting with `__` routes the call
+        // through the runtime dispatch that makes an AgencyFunction callable.
+        const fnName = `matchInit$${++this.counter}`;
+        const synthFn: FunctionDefinition = {
+          type: "function",
+          functionName: fnName,
+          parameters: [],
+          returnType: null,
+          body: [
+            ...region.statements,
+            { type: "returnStatement", value: region.valueRef, loc: node.loc },
+          ],
+          loc: node.loc,
+        };
+        const call: FunctionCall = {
+          type: "functionCall",
+          functionName: fnName,
+          arguments: [],
+          loc: node.loc,
+        };
+        return [synthFn, { ...node, value: call }];
+      }
       return [
         ...region.statements,
         { ...node, value: region.valueRef, matchExprSource: { matchId: region.matchId } },
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -277,7 +277,10 @@ class PatternLowerer {
           arguments: [],
           loc: node.loc,
         };
-        return [synthFn, { ...node, value: call }];
+        return [
+          synthFn,
+          { ...node, value: call, matchExprSource: { matchId: region.matchId } },
+        ];
       }
       return [
         ...region.statements,
--- a/packages/agency-lang/docs/site/guide/pattern-matching.md
+++ b/packages/agency-lang/docs/site/guide/pattern-matching.md
@@ -268,7 +268,6 @@ const greeting = match(name) {
 
 Here are the places you can't use match blocks right now:
 
-- At the module level (outside of a function or node).
 - Inside `parallel`, `seq`, `thread`, and `subthread` blocks.
 
 You can't use `goto` with match blocks:
--- a/packages/agency-lang/lib/lowering/patternLowering.matchExpr.test.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.matchExpr.test.ts
@@ -186,8 +186,67 @@ describe("expression match lowering errors", () => {
   });
   it("match(x is ...) in expression position errors", () =>
     expectError(`node main(x: any) {\n  const val = match(x is { k }) {\n    _ => 2\n  }\n  return val\n}`, /cannot be used as an expression/i));
-  it("module-level match expression errors", () =>
-    expectError(`const g = match("a") {\n  "a" => 1\n  _ => 2\n}\nnode main() { return g }`, /module-level|top-level/i));
+});
+
+describe("module-level match expression hoisting", () => {
+  it("module-level match hoists to a synthesized function + a call", () => {
+    const parsed = parseAgency(`const label = match("a") {
+  "a" => "A"
+  _ => "other"
+}
+node main(): string { return label }`);
+    expect(parsed.success).toBe(true);
+    if (!parsed.success) return;
+    const nodes = parsed.result.nodes as any[];
+    // The const now calls a synthesized function instead of holding the match.
+    const decl = nodes.find(
+      (n) => n.type === "assignment" && n.variableName === "label",
+    );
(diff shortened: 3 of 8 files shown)


packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -239,13 +240,45 @@ class PatternLowerer {
...
+          ],
+          loc: node.loc,
+        };
+        const call: FunctionCall = {
+          type: "functionCall",
+          functionName: fnName,
+          arguments: [],
+          loc: node.loc,
+        };
+        return [synthFn, { ...node, value: call }];
```
egonSchiele: 🐛 **Module-level match consumer is typed `any`, silently escaping type checks the in-function form enforces.**

This branch drops `matchExprSource` (only the non-module path below attaches it). Expression-match result-typing is gated on that tag: `matchExprTypes.ts:80` patches the consumer variable to the match's union, and `scopes.ts:201` runs `checkMatchExprYields`, **only** when `matchExprSource` is set.

So the consumer is now typed from the synthesized function's inferred return type — which resolves to `any`: `inferReturnTypes` (`index.ts:296`) runs *before* `computeMatchExprTypes` (`index.ts:329`) populates `ctx.matchExprTypes`, so the `__matchval_N` ref in the synth fn's `return` synths to `any` (`synthesizer.ts:261`, `?? "any"`).

Result: type mismatches caught everywhere else are silently accepted at module level. Confirmed on the branch:

```
const env: string = "prod"
const label = match(env) { "prod" => "Production"; _ => "Local" }
const n: number = label   // accepted at module level; `label` is `any`
```

The identical code inside a `def` correctly errors `Type 'string' is not assignable to type 'number'`. Consider deriving the synth fn's `returnType` from `matchExprTypes[matchId]`, or otherwise re-establishing the `matchExprSource` typing link for the consumer.

egonSchiele (commented):
## Review: module-level `match` initializers

Nice, well-scoped change. The auto-hoist genuinely reuses the existing depth-1 init machinery rather than re-plumbing it, and the three `topsortCycleErrors` cases really do prove dependency detection survives the indirection (they would flip to `success` if the `matchInit$N` edge were missed). I checked out the branch and confirmed the happy paths — const / let / static const / dependent-ordering — all compile and run. Counter naming, `returnStatement` shape, exhaustiveness plumbing, and the `$`-in-name → dep-graph resolution all check out clean.

Two correctness issues and two doc/comment nits below.

### 2. ⚠️ Call-valued scrutinee reading a later global → silent wrong value (depth-2 gap)
`lib/compiler/initDepGraph.ts` (depth-1 boundary) + `lib/runtime/state/globalStore.ts` (no uninit trap for globals).

Depth-1 expansion walks only one hop, so when the scrutinee is itself a call the real read is depth-2 and is missed. Confirmed on the branch:

```
def getEnv(): string { return realEnv }
const label = match(getEnv()) { "prod" => "Production"; _ => "Local" }
const realEnv: string = "prod"
```

This compiles and runs, silently producing `"Local"`. The **direct** equivalent `const label = getEnv()` is caught at compile time (`Global 'label' references 'realEnv' ... which is declared later`). So refactoring `const x = getEnv()` → `const x = match(getEnv()) {...}` turns a clear compile error into a silent wrong value. It's not a regression (module-level match was a hard error before) and it matches the documented depth-1 boundary — but this PR makes the exposure ergonomic. Note statics are safe (the `__readStatic` sentinel throws); only globals are silent. Worth at least a caveat in `docs/dev/init-topsort.md`.

### 3. 📄 Stale doc line (you flagged this)
`docs/site/guide/pattern-matching.md:271` still lists *"At the module level (outside of a function or node)"* under v1 restrictions — now inaccurate, and the only remaining such statement in `docs/site/guide/`. Suggest deleting that bullet in this PR.

### 4. 📝 Stale method comment now contradicts the code
`lib/lowering/patternLowering.ts:238-239` — the `lowerAssignment` docstring still says module-level initializers *"are rejected."* The new inline comment inside the `if (this.atModuleLevel)` branch is correct, but this outer method comment should be updated to say they're hoisted into a synthesized init function.

### Minor: test coverage
The execution fixture exercises only single-expression literal arms at module level. A module-level match with a multi-statement arm that `return`s (yields) would exercise the synthesized function's `_matchExit` frame more directly. Low priority (that mechanism is shared with the in-function path), but one added case would raise confidence.

egonSchiele:
Great catch — confirmed and fixed in 0feb8882. Your diagnosis was exactly right: the consumer was untagged, so its type came from the synth function's inferred return type, which resolves to `any` because `inferReturnTypes` runs before `computeMatchExprTypes`.

I went with *both* levers you suggested, because each fixes a different half:

- **Tag the rewritten consumer with `matchExprSource`.** This alone fixed the *annotated* path — `checkMatchExprYields` reads the global `matchExprYieldTypes` table, so a `const label: number = match(...) { "prod" => "Production" }` now errors per-arm.
- **But the *un*annotated path stayed broken** (`const label = match(...)` → `const n: number = label` still silently accepted). The consumer-patch in `computeMatchExprTypes` runs per-scope inside the compute loop, and the module scope is processed before the synth fn's scope — so `ctx.matchExprTypes[matchId]` wasn't populated yet when `label` was patched. Deriving the synth fn's `returnType` wasn't viable either, since `inferReturnTypes` is locked in at step 2 (before scopes/flow exist). So I split `computeMatchExprTypes` into two passes: compute all scopes' types, then patch all consumers. Now the cross-scope consumer resolves regardless of order; same-scope behavior is unchanged.

Verified on your exact repro:

```
const env: string = "prod"
const label = match(env) { "prod" => "Production"  _ => "Local" }
const n: number = label   // now: Type 'string' is not assignable to type 'number'
```

Added three regression tests in `matchExpression.test.ts` (unannotated union flows downstream; annotated mismatch caught per-arm; compatible annotation clean). Full suite green (7148).

PR #427, opened 2026-07-04
Add default maxCallDepth guard to catch runaway recursion before OOM (#243)

### LINKED ISSUE #243: Add default maxCallDepth to catch runaway async recursion before OOM
## Problem

Agency has no default protection against runaway function-call recursion. The three failure modes today:

1. **Synchronous recursion** trips V8's stack limit and throws `RangeError: Maximum call stack size exceeded` after ~10k-50k frames. This is bounded and produces an actionable error.
2. **Asynchronous recursion** does NOT blow the stack — each `await` flattens the call frame — so an infinite async cycle (e.g. `foo()` calls `bar()` calls `foo()` where both are async, which most Agency functions are) consumes memory as the promise chain grows and eventually OOMs the process. Slow failure, multi-GB before crashing, no useful diagnostic.
3. **Cost guards** (`CostGuard`, `TimeGuard` in `lib/runtime/guard.ts`) can catch runaway computation by token budget or wall-clock, but they are opt-in and a typical user has none installed. They are also fundamentally orthogonal to the cycle problem — you can spin infinitely without making any LLM call.

Since most Agency functions are async (they take `__ctx`, return promises), case #2 is the realistic failure mode and it's bad: minutes of unresponsiveness followed by OOM.

## Today's repro

```agency
def foo() { return bar() }
def bar() { return foo() }

node main() { return foo() }
```

Run this and watch memory climb until the process gets killed. No diagnostic about what went wrong.

## Existing precedent in the codebase

There are several places we already do bounded recursion / depth limiting:

- `lib/runtime/initVar.ts` — explicit cycle detection for static-init cycles via the `running` flag. Throws a clean error listing the participating `__init_*` frames.
- `lib/runtime/interrupts.ts` — small built-in limit on handler-call depth. Catches runaway handler recursion at ~50 deep with a clear error.
- `lib/runtime/validateChain.ts` — `maxDepth: number = 64` on validation recursion. Throws a clean error at the limit.

A default `maxCallDepth` on user function-call depth fits this pattern: bounded, opt-out, actionable error.

## Proposal

Add a default `maxCallDepth` (suggested: 1000 or 10000 — pick something safely larger than typical realistic call depth, smaller than where OOM bites) on the runtime context. Increment on function-call entry, decrement on exit. If a call would exceed the limit, throw a `RuntimeError` (or `GuardExceededError`-like type) with:

- The current call-stack frame names (just the top N — full stack would be huge).
- The limit and observed depth.
- A pointer to the override config knob.

The user can raise the limit via runtime config (e.g. `agency.json` or a runtime API) for legitimate deep-recursion use cases, just like `maxDepth` is overridable in validation.

## Why this is a quick(ish) win

The state stack (`lib/runtime/state/stateStack.ts`) already tracks per-call frames, so the depth counter is mechanical to add. The error class already has precedent. The override knob slot already exists in `AgencyConfig`. Implementation is probably ~50 lines runtime + a test or two.

## Open questions

1. **What's the right default?** Too low and legitimate recursion (e.g. tree traversal of a deeply nested structure) breaks. Too high and OOM still wins. 1000 is conservative-aggressive; 10000 is permissive. Plain JS V8 stack limit is ~10k for thin frames so 1000-2500 is probably the sweet spot.
2. **Per-async-chain vs per-sync-stack?** Each `await` resets V8's stack, but the LOGICAL call depth (what we'd track on `__ctx.stateStack`) keeps growing. Want to track the logical depth, not V8's physical stack.
3. **Counting LLM tool calls vs regular def calls?** Should an LLM-orchestrated tool-loop count against the same budget? Probably yes (a runaway tool-call loop is also a real concern), but the LLM loop has its own iteration count, so doubling up may be redundant.
4. **Interaction with `runBatch` and other parallelism primitives?** Each branch has its own depth; the limit should be per-stack, not aggregated across branches.

## Followup from

PR #237 review discussion about what Agency does (and doesn't) catch at runtime.

egonSchiele:
## Summary

Fixes #243 — Agency had no default protection against runaway function-call recursion.

The dangerous case is **async** recursion. Most Agency functions are async (they `await`), and each `await` flattens V8's call frame — so an infinite async cycle never throws `RangeError`. Instead it grows the promise chain until the process **OOMs, minutes later, with no useful diagnostic**. (Synchronous recursion already trips V8's stack with an actionable error; this closes the async gap.)

## What this adds

A logical call-depth guard at the single chokepoint every Agency call passes through — `AgencyFunction.invoke()`.

- **`lib/runtime/callDepth.ts`** — tracks the logical nesting depth of the active async *lineage* in an `AsyncLocalStorage` (mirrors `handlerChainDepthALS` in interrupts.ts). Because depth is a property of the async call **tree**, not a global count:
  - concurrent `fork`/`parallel`/tool-call fan-out each inherit the parent depth and descend independently — a wide fan-out is never mistaken for deep recursion;
  - it's never serialized, so there's nothing to reset across checkpoint/resume.

  Frames are a linked list (O(1) per call; names are walked only on the cold overflow path). Default ceiling **2048** — well above realistic nesting (deep tree walks, recursive descent), far below where async recursion exhausts memory.

- **`CallDepthExceededError extends AgencyAbort`** with a new `callDepthExceeded` `AbortCause`, modeled exactly like `GuardExceededError`. As an abort it propagates untouched through every generated catch rung and **halts the run**, rather than being converted to a `Failure` that could silently re-descend into the recursion. The message names the limit, the observed depth, the recent call chain, and points at the config knob:

  ```
  Maximum call depth exceeded (2049 > 2048). This usually means unbounded
  recursion. Recent calls: bar → foo → bar → foo → bar → foo → bar → foo.
  If your program legitimately recurses this deeply, raise the limit via the
  `maxCallDepth` config option.
  ```

- **`maxCallDepth` config option** (top-level in `AgencyConfig`, zod-validated), plumbed through codegen → `RuntimeContext` args → field → execCtx exactly like `maxRestores`, and honored as a subprocess config override. Documented in `docs/dev/config.md`.

## Testing

- `lib/runtime/callDepth.test.ts` — guard logic: trips over the limit, allows up to it, per-lineage (not global) counting, abort cause + message contents.
- `lib/runtime/agencyFunction.test.ts` — `invoke()` integration: mutual recursion and async-self recursion trip; bounded recursion succeeds. (Written TDD — without the guard these crash the vitest worker with a stack blowup.)
- `tests/agency/call-depth-bounded.agency` — e2e regression guard that deep-but-bounded recursion (depth 300) still runs to completion.
- Full `lib/runtime` + `lib/config` unit suite (1092 tests) and `lib/backends` codegen suite (339 tests) pass; structural linter clean; `tsc --noEmit` clean.
- **Manual e2e**: the issue's exact async-recursion repro now trips at 2049 > 2048 with the diagnostic and exits code 1 instead of OOMing; a low `maxCallDepth` config trips earlier; `map` over 5000 elements does not false-trip; an interrupt suspend/resume test still passes through the new wrapper.

## Notes / decisions

- **Why hook `invoke()`** rather than `__call`: `invoke()` is the one point all Agency calls funnel through (direct calls, method calls, and LLM tool-loop dispatch), so a single guard covers every path.
- **Why an `AgencyAbort`** rather than a plain `Error`: generated `def`/`node` bodies convert thrown `Error`s to `Failure`. A plain error would be swallowed per-frame and could let a degenerate loop re-descend; an abort propagates untouched and halts, matching how guard trips already behave.
- The 2048 default is deliberately conservative-aggressive (the issue suggested 1000–2500). It's fully overridable for legitimate deep-recursion workloads.

### DIFF
--- a/packages/agency-lang/lib/config.ts
+++ b/packages/agency-lang/lib/config.ts
@@ -475,7 +475,10 @@ export const AgencyConfigSchema = z
     debugger: z.boolean(),
     instrument: z.boolean(),
     checkpoints: z.object({ maxRestores: z.number() }).partial(),
-    maxCallDepth: z.number(),
+    // A positive integer. The guard trips when depth > limit and the first
+    // call is depth 1, so a value < 1 (or a float/NaN) would make every call
+    // throw — reject it at config-load rather than bricking the program.
+    maxCallDepth: z.number().int().positive(),
     trace: z.boolean(),
     traceFile: z.string(),
     traceDir: z.string(),
--- a/packages/agency-lang/lib/config.ts
+++ b/packages/agency-lang/lib/config.ts
@@ -250,6 +250,13 @@ export interface AgencyConfig {
     maxRestores?: number;
   };
 
+  /** Maximum logical function-call nesting depth before the runaway-recursion
+   * guard throws CallDepthExceededError. Catches unbounded recursion — most
+   * importantly the async kind, which grows the promise chain until the process
+   * OOMs with no useful diagnostic — before it exhausts memory. Raise this for
+   * programs that legitimately recurse very deeply. Default: 2048. */
+  maxCallDepth?: number;
+
   /** Enable execution tracing — writes checkpoints to a .trace file */
   trace?: boolean;
 
@@ -468,6 +475,7 @@ export const AgencyConfigSchema = z
     debugger: z.boolean(),
     instrument: z.boolean(),
     checkpoints: z.object({ maxRestores: z.number() }).partial(),
+    maxCallDepth: z.number(),
     trace: z.boolean(),
     traceFile: z.string(),
     traceDir: z.string(),
--- a/packages/agency-lang/lib/runtime/agencyFunction.ts
+++ b/packages/agency-lang/lib/runtime/agencyFunction.ts
@@ -2,6 +2,7 @@ import { z } from "zod";
 import { stripBoundParams } from "./stripBoundParams.js";
 import { approve } from "./interrupts.js";
 import { agencyStore, withPushedHandler } from "./asyncContext.js";
+import { withCallDepth, DEFAULT_MAX_CALL_DEPTH } from "./callDepth.js";
 import { formatRequiredUnboundRuntimeError } from "./toolBlockDiagnostics.js";
 
 export const UNSET: unique symbol = Symbol("UNSET");
@@ -136,10 +137,20 @@ export class AgencyFunction {
   }
 
   async invoke(descriptor: CallType): Promise<unknown> {
-    const args = this._isBound
-      ? this.mergeWithBound(this.resolveArgs(descriptor))
-      : this.resolveArgs(descriptor);
-    return this._fn(...args);
+    // Guard every Agency call against runaway recursion. `invoke()` is the
+    // single chokepoint all calls pass through, so counting logical nesting
+    // here catches the async-recursion case that never trips V8's stack but
+    // grows the promise chain until the process OOMs. Limit is read from the
+    // active execution context (config-overridable), falling back to the
+    // default when no context is installed. See lib/runtime/callDepth.ts.
+    const limit =
+      agencyStore.getStore()?.ctx?.maxCallDepth ?? DEFAULT_MAX_CALL_DEPTH;
+    return withCallDepth(this.name, limit, () => {
+      const args = this._isBound
+        ? this.mergeWithBound(this.resolveArgs(descriptor))
+        : this.resolveArgs(descriptor);
+      return this._fn(...args);
+    });
   }
 
   partial(bindings: Record<string, unknown>): AgencyFunction {
--- a/packages/agency-lang/lib/runtime/agencyFunction.ts
+++ b/packages/agency-lang/lib/runtime/agencyFunction.ts
@@ -2,7 +2,7 @@ import { z } from "zod";
 import { stripBoundParams } from "./stripBoundParams.js";
 import { approve } from "./interrupts.js";
 import { agencyStore, withPushedHandler } from "./asyncContext.js";
-import { withCallDepth, DEFAULT_MAX_CALL_DEPTH } from "./callDepth.js";
+import { withCallDepth } from "./callDepth.js";
 import { formatRequiredUnboundRuntimeError } from "./toolBlockDiagnostics.js";
 
 export const UNSET: unique symbol = Symbol("UNSET");
@@ -140,12 +140,10 @@ export class AgencyFunction {
     // Guard every Agency call against runaway recursion. `invoke()` is the
     // single chokepoint all calls pass through, so counting logical nesting
     // here catches the async-recursion case that never trips V8's stack but
-    // grows the promise chain until the process OOMs. Limit is read from the
-    // active execution context (config-overridable), falling back to the
-    // default when no context is installed. See lib/runtime/callDepth.ts.
-    const limit =
-      agencyStore.getStore()?.ctx?.maxCallDepth ?? DEFAULT_MAX_CALL_DEPTH;
-    return withCallDepth(this.name, limit, () => {
+    // grows the promise chain until the process OOMs. The limit (config-
+    // overridable `maxCallDepth`) is resolved from the active execution context
+    // inside `withCallDepth`, once per lineage. See lib/runtime/callDepth.ts.
+    return withCallDepth(this.name, () => {
       const args = this._isBound
         ? this.mergeWithBound(this.resolveArgs(descriptor))
         : this.resolveArgs(descriptor);
--- a/packages/agency-lang/lib/runtime/callDepth.ts
+++ b/packages/agency-lang/lib/runtime/callDepth.ts
@@ -1,4 +1,5 @@
 import { AsyncLocalStorage } from "node:async_hooks";
+import { agencyStore } from "./asyncContext.js";
 import { CallDepthExceededError } from "./errors.js";
 
 /**
@@ -13,15 +14,29 @@ import { CallDepthExceededError } from "./errors.js";
  * infinite async cycle never throws `RangeError` — it just grows the promise
  * chain until the heap is exhausted, minutes later, with no diagnostic. We
  * track the LOGICAL depth of the async call tree instead.
+ *
+ * Budget accuracy caveat: EVERY Agency call — including the stdlib higher-order
+ * functions `map`/`filter`/`reduce`/`flatMap` — consumes one depth level, and a
+ * HOF also dispatches its callback through another call. So recursion written as
+ * `walk(n) { children.map(walk) }` burns ~2 levels per user level (more when
+ * HOFs nest), i.e. the effective budget is roughly halved versus the same walk
+ * written with a `for` loop, which costs 1 level per user level. The default is
+ * generous enough that this rarely matters; raise `maxCallDepth` for very deep
+ * HOF-style recursion. (Counting is by call, not by heap growth, so a bounded
+ * fire-and-forget spawn like `async produce(n-1)` also climbs the depth and can
+ * trip even though it holds only a bounded promise queue — a conservative false
+ * positive, acceptable for a safety net.)
  */
 export const DEFAULT_MAX_CALL_DEPTH = 2048;
 
 /** A node in the active async lineage's call chain. Linked (not an array) so
  *  entering a call is O(1) with a single small allocation — the frame names
- *  are only walked on the cold overflow path. */
+ *  are only walked on the cold overflow path. `limit` is resolved once at the
+ *  root of the lineage and carried down, so nested calls never re-read it. */
 type CallFrame = {
   readonly name: string;
   readonly depth: number;
+  readonly limit: number;
   readonly parent: CallFrame | null;
 };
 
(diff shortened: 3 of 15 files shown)


packages/agency-lang/lib/runtime/agencyFunction.ts
```diff
@@ -136,10 +137,20 @@ export class AgencyFunction {
...
-      ? this.mergeWithBound(this.resolveArgs(descriptor))
-      : this.resolveArgs(descriptor);
-    return this._fn(...args);
+    // Guard every Agency call against runaway recursion. `invoke()` is the
+    // single chokepoint all calls pass through, so counting logical nesting
+    // here catches the async-recursion case that never trips V8's stack but
+    // grows the promise chain until the process OOMs. Limit is read from the
+    // active execution context (config-overridable), falling back to the
+    // default when no context is installed. See lib/runtime/callDepth.ts.
+    const limit =
```
egonSchiele: Is there a perf impact to this? either adding a `getStore()` call on every function call, or adding a `withCallDepth` call?

egonSchiele (commented):
## Review: `maxCallDepth` runaway-recursion guard (#243)

Solid, well-tested change. I checked out the branch and independently verified the parts that matter most for a guard sitting at the call chokepoint:

- **Coverage holds.** Every recursion path that can actually grow the promise chain routes through `AgencyFunction.invoke()` → `withCallDepth`: direct and method calls (`__call`/`__callMethod` dispatch AgencyFunctions through `invoke`), `fork`/`parallel` branch bodies, async calls, and LLM tool dispatch (`prompt.ts:949`). `goto`/node-to-node loops run iteratively in simplemachine and are correctly **not** counted (no OOM risk there). No bypass found.
- **Handler / interrupt / checkpoint safety holds.** `CallDepthExceededError extends AgencyAbort`, so it propagates untouched through every catch rung (`functionCatchFailure.mustache:11`, `__tryCall` at `result.ts:151`, `prompt.ts:967`, `hooks.ts:217`) and halts rather than converting to a `Failure` — confirmed by running an unbounded recursion e2e (halts with the diagnostic, exit 1; `try` did not swallow it). The wrapper adds a second, independent ALS and never touches `ctx.handlers`, so no handler is skipped and no `RestoreSignal`/abort is swallowed. Depth resets to 0 across checkpoint/resume (ALS isn't serialized) — a deliberate, safe under-count.

Findings below, most actionable first. None are safety holes — they're budget-accuracy and config-robustness issues.

### 3. Fire-and-forget async recursion false-trips even though it can't OOM
`produce(n) { let x = async produce(n - 1); return 1 }` accumulates logical depth (each `async f()` invokes synchronously within the spawning lineage) and aborts at 2049 while the pending-promise queue drains — even though the pattern holds only a bounded queue and O(1) stack and never approaches heap exhaustion. That's a false positive relative to the guard's stated purpose (turn an *OOM* into an error). Narrow (recursive spawning is the only way to reach it), but worth a note.

### 5. Minor cleanup
- `lib/runtime/configOverrides.ts:38-42` — the function's doc comment ("other `AgencyConfig` fields are ignored…") is now contradicted by the new `maxCallDepth` merge block below it; update it.
- `lib/runtime/callDepth.ts:59` — `currentCallDepth` is exported but has no caller outside `callDepth.test.ts` and isn't re-exported from `runtime/index.ts`. Wire it into observability or drop the export.

Inline comments cover findings 1, 2, and 4.

packages/agency-lang/lib/runtime/callDepth.ts
```diff
@@ -0,0 +1,81 @@
+import { AsyncLocalStorage } from "node:async_hooks";
+import { CallDepthExceededError } from "./errors.js";
+
+/**
+ * Default ceiling on logical function-call nesting depth. Sits well above any
+ * realistic call nesting (deep tree walks, recursive descent) yet far below
```
egonSchiele, line 7: ⚠️ **The effective budget is ~half of 2048 for the "recursive descent" this comment cites as safe.**

`map`/`filter`/`reduce`/`flatMap` are themselves `AgencyFunction`s (`stdlib/index.js` → `__AgencyFunction.create`), so each consumes one `invoke` depth frame **and** dispatches its callback through `__call` (`__map_impl` does `await __call(__stack.args.func, …)`) — a second frame. So recursion written the idiomatic way, `walk(n) { children.map(walk) }`, burns **2** depth levels per user level and trips at ~1024, not 2048. Nested HOFs (`map` inside `filter` inside `reduce`) multiply the divisor further. Empirically, recursion via `map` succeeds near depth 1000 and trips near 1100.

The same walk written as `for (c in children) { walk(c) }` gets the full 2048 — so the effective budget silently depends on which iteration form the user picked. Consider either not counting stdlib-HOF frames (and/or their callback dispatch) toward depth, or documenting here that the budget is roughly halved when recursing through `map`/`filter`/`reduce`.

packages/agency-lang/lib/config.ts
```diff
@@ -468,6 +475,7 @@ export const AgencyConfigSchema = z
     debugger: z.boolean(),
     instrument: z.boolean(),
     checkpoints: z.object({ maxRestores: z.number() }).partial(),
+    maxCallDepth: z.number(),
```
egonSchiele: 🐛 **`z.number()` is too loose — `maxCallDepth: 0` (or any value < 1, a float, or `NaN`) passes validation and bricks the program.**

The guard trips when `depth > limit`, and the very first `invoke()` is `depth === 1`. So `maxCallDepth: 0` makes *every* function call throw `CallDepthExceededError` — with the misleading message "This usually means unbounded recursion" — at the very first call, even though config validated fine. Suggest `.int().positive()` (the codebase already uses refinements elsewhere, e.g. `coverage.threshold: z.number().min(0).max(100)`).

(`maxRestores` is equally loose, but its blast radius — a rarely-hit `>` check on checkpoint restores — is far smaller, so the looseness matters much more here.)

packages/agency-lang/lib/runtime/agencyFunction.ts
```diff
@@ -136,10 +137,20 @@ export class AgencyFunction {
...
+      agencyStore.getStore()?.ctx?.maxCallDepth ?? DEFAULT_MAX_CALL_DEPTH;
```
egonSchiele: 📝 **A user-configured `maxCallDepth` is silently ignored on this path when there's no active `agencyStore` frame.**

When `invoke()` runs outside any node body (e.g. a function invoked directly as a tool before a context is installed), `agencyStore.getStore()?.ctx` is undefined and the limit falls back to `DEFAULT_MAX_CALL_DEPTH` (2048), ignoring a configured value. Safety is unaffected (the default still guards), but a program that lowered `maxCallDepth` won't have it honored on this path. Minor — flagging for awareness.

packages/agency-lang/lib/runtime/errors.ts
```diff
@@ -161,6 +166,32 @@ export class AgencyCancelledError extends AgencyAbort {
...
+    const chain = recentFrames.join(" → ");
+    super(
+      `Maximum call depth exceeded (${observed} > ${limit}). This usually ` +
+        `means unbounded recursion. Recent calls: ${chain}. If your program ` +
+        `legitimately recurses this deeply, raise the limit via the ` +
+        `\`maxCallDepth\` config option.`,
+      makeAbortCause({ kind: "callDepthExceeded", limit, observed }),
+    );
+    this.name = "CallDepthExceededError";
+    this.limit = limit;
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 013e9521 — added `"CallDepthExceededError"` to the name-based fallback in `isAbortError`, alongside the existing `GuardExceededError` entry, with a matching cross-realm test in `errors.test.ts`. A trip reconstructed across the subprocess shim now classifies as an abort and propagates instead of converting to a Failure.

packages/agency-lang/lib/config.ts
```diff
@@ -468,6 +475,7 @@ export const AgencyConfigSchema = z
...
     trace: z.boolean(),
     traceFile: z.string(),
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 013e9521 — `maxCallDepth` is now `z.number().int().positive()`, so `0`/negative/float/NaN are rejected at config load (verified: `{ "maxCallDepth": 0 }` now fails with "Too small: expected number to be >0"). Added config tests.

packages/agency-lang/lib/runtime/agencyFunction.ts
egonSchiele: Measured it: the guard adds ~240ns/call in isolation, dominated by the `AsyncLocalStorage.run` that per-lineage depth tracking requires (a plain ctx counter can't be used — it would count concurrent fork/parallel fan-out breadth as depth). Against a real `invoke()` — which already allocates a `Runner`, resolves args, runs an `agencyStore` ALS frame via `runInScope`, and emits statelog — that's ~4% of per-call cost, negligible for LLM-bound work.

Trimmed what I could in 013e9521: the limit is now resolved from the context once at the root of each lineage and carried in the frame, so a deep recursion pays a single `agencyStore` lookup instead of one per call (`invoke()` no longer reads ctx itself). The remaining `als.run` is irreducible for correct per-lineage tracking.

packages/agency-lang/lib/runtime/callDepth.ts
egonSchiele: Confirmed the 2x — `map`/`filter`/`reduce`/`flatMap` are ordinary stdlib defs (`stdlib/index.agency`), so each burns its own `invoke` frame plus the `__call` callback dispatch.

I went with documentation (013e9521) rather than not-counting stdlib frames: the only reliable signal for "is this stdlib" is the `module` string prefix (`stdlib/*.agency` / `__runtime`), and keying runtime control flow on a stringly-typed path is fragile (user file paths could collide; silent behavior change if the naming scheme moves). Since the default stays generous even halved and this is a safety net rather than a precise budget, I documented the ~2-levels-per-HOF-level cost in both the `callDepth.ts` header and `docs/dev/config.md`. Happy to do the frame-skipping version instead if you'd prefer it — it'd need an explicit `builtin` flag on `AgencyFunction` to avoid the module-string coupling.

packages/agency-lang/lib/config.ts
egonSchiele: Fixed in 013e9521 — `z.number().int().positive()`, matching the `coverage.threshold` refinement precedent you cited. `maxCallDepth: 0` now fails config validation up front instead of throwing "unbounded recursion" on the first call. (Left `maxRestores` as-is per your note that its blast radius is small; can tighten it in a follow-up if you want.)

packages/agency-lang/lib/runtime/agencyFunction.ts
egonSchiele: Correct, and it's by design: outside any node body there is no execution context, so there is no configured `maxCallDepth` to read — the fallback to `DEFAULT_MAX_CALL_DEPTH` is the only sane value there, and safety still holds (the default guards). Honoring a lowered limit on that path would require caching the config in a module global, which the codebase avoids for isolation reasons. Left as-is; flagging that I evaluated it.

egonSchiele:
Addressed the review in 013e9521. Inline replies on each thread; summary of the review-body findings:

**Fixed**
- **Config robustness (#2 / Copilot):** `maxCallDepth` → `z.number().int().positive()`; `0`/negative/float/NaN now rejected at config load. + tests.
- **Cross-realm classification (Copilot):** `"CallDepthExceededError"` added to the `isAbortError` name fallback. + test.
- **Finding 5 cleanup:** updated the `configOverrides` doc comment (it now forwards `maxCallDepth`); removed the unused `currentCallDepth` export.
- **Perf (inline):** limit is resolved from the context once per lineage and carried in the frame, so deep recursion pays one `agencyStore` lookup, not one per call. Measured overhead ~240ns/call (dominated by the irreducible `als.run`), ~4% of a real `invoke`.

**Documented (rather than changed behavior)**
- **Finding 3 (fire-and-forget false-trip):** noted in the `callDepth.ts` header — a bounded `async produce(n-1)` spawn climbs logical depth and can trip conservatively even though it can't OOM. Acceptable for a safety net; called out so it's not surprising.
- **HOF budget halving (inline #4):** documented in both `callDepth.ts` and `docs/dev/config.md` that map/filter/reduce/flatMap cost ~2 levels per user level. Chose docs over frame-skipping because the only signal is the `module` string prefix and keying control flow on that is fragile — offered the explicit-`builtin`-flag version as a follow-up if preferred.

**Acknowledged as by-design**
- **No-ctx fallback (inline):** outside a node body there is no execution context to read a configured limit from, so the default is the only sane value; safety still holds.

All runtime + config unit suites green (1082 + config), lint + tsc clean; re-verified e2e that the async-recursion repro trips at 2049>2048 (exit 1) and `maxCallDepth: 0` is rejected at load.

PR #428, opened 2026-07-05
Add `if … then … else` conditional expressions (#411)

### LINKED ISSUE #411: Add if expressions (assignable/returnable, mirroring match expressions)
## Summary

Agency has `match` expressions (assignable / returnable) but `if` is statement-only, and there is no ternary (dropped deliberately, see `docs/site/guide/basic-syntax.md`). That leaves a gap: there's no lightweight way to pick a value based on a boolean condition without either a `match` on the boolean or a mutable `let` + `if` statement.

We should add **`if` expressions**, so this works:

```ts
const label = if (isProd) { "Production" } else { "Local" }
```

## Motivation

- Fills the hole left by removing ternaries, without reintroducing `? :`.
- `match` already proves the machinery exists for lowering a control-flow construct into a value-yielding expression; `if` expressions can reuse it.
- Removes the `match(cond) { true => ..., false => ... }` and `let x; if (...) { x = ... } else { x = ... }` boilerplate.

## Proposed semantics (mirror `match` expressions)

- **`else` is mandatory in expression position.** Like a `match` expression, an `if` expression must always produce a value, so a missing `else` (which could yield nothing) is a hard compile error. `else if` chains are fine as long as they terminate in an `else`.
- **Type:** in checked position each branch is checked against the annotated type; in synthesis position the type is the union of all branch types (same rule as match arms).
- **Single-expression vs block branches:** a braced branch that is a single expression yields it implicitly; a multi-statement block branch uses an explicit `return` that returns from the `if` expression (not the enclosing function) — identical to match block arms.
- **Same v1 restrictions as match expressions:** limited to assignment RHS and `return` operands; `return` inside a branch cannot cross a `parallel`/`seq`/`fork`/`race`/`thread` boundary; cannot yield a graph-node call; cannot appear inside a `with` handler body.

## Open questions

- Should module-level `const x = if (...) { ... } else { ... }` be allowed from day one, or share the same module-level restriction `match` currently has? (See companion issue on lifting the module-level `match` restriction.)
- Do we require braces on both branches in expression position (recommended, for consistency with the rest of the language), or allow brace-less single expressions?

## Related

- Companion: "Allow match expressions in module-level const initializers".
- `if` as a plain statement is unchanged; this only adds the expression form.

egonSchiele:
## Summary

Closes #411. Adds `if ... then ... else` as a conditional **expression**:

```ts
const label = if isProd then "Production" else "Local"

def describe(n: number): string {
  return if n > 100 then "big" else "small"
}
```

## The important part: it's sugar for the *stepped* `if`, so interrupts work

Reviewing the tests surfaced that an earlier ternary-based version **silently swallowed interrupts** thrown from a branch. Since interrupts must work everywhere, this version lowers `if c then a else b` to the **same `runner.ifElse` machinery a `match` expression uses** — a `matchExprId`-tagged `ifElse` whose branches `matchYield` — **not** a ternary. A fixture (`withInterrupt`) locks in that an interrupt from a branch pauses and resumes correctly.

Each branch is lowered as a statement-position temp binding (`const __tmp = <expr>`) then a yield of the temp, so a branch that is a function call is compiled at statement position where codegen emits interrupt/checkpoint propagation. (A bare `matchYield(await __call(f))` — the shape `match`'s own single-expression arms still generate — loses the interrupt.)

## The tradeoff (agreed with owner)

Like `match` expressions, an `if` expression is only valid as a `const`/`let` value or a `return` — **not** nested as an object field or argument. You can't both be a nested value *and* participate in stepping, because stepping needs a statement frame. Interrupts win; nested-value use was optional.

## What's in the diff

- **Parser**: `ifExpressionParser` produces an `IfElse` with single-expression branch bodies, wired only into the assignment-RHS / `return` positions (not `exprParser`, so it can't be a nested value). Guarded so a branch/condition can't start with `if` (no nesting, no `else if`); `else` is required (committed parse error).
- **Lowering**: `lowerIfExpressionCore`, sharing the `match` hoist / module-level machinery via `expressionRegion`. `is`-pattern conditions bind into the then-branch, like a statement `if`.
- **Formatter**: prints the `then` form for an `ifElse` in value position (round-trips).
- **Codegen / runtime / typechecker**: unchanged — rides `matchExprId` + `matchYield` + `matchExprSource`; the condition is checked as `boolean` via the existing `ifElse` path.

## Testing

- **Execution** (`tests/agency/if-expression.agency`, 8): then/else, `return` form, module-level, **interrupt in a branch (pause/resume)**, `is`-pattern condition binding.
- **Lowering** (3): pins the stepped shape (matchExprId `ifElse` + `matchYield` + `matchExprSource`), and the statement-position temp binding that makes interrupts propagate.
- **Parser** (8): AST shape; valid in return position; rejected as an object value / argument / nested / `else if` / missing-`else`; formatter round-trip.
- **Typecheck** (4): union typing, mismatch, downstream flow, non-boolean condition.
- Full unit suite: **7163 passed, 0 failures**. Lint clean.


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2499,6 +2500,31 @@ const unaryVoidParser = unaryKeywordParser("void");
...
+  set("type", "ifExpression"),
+  str("if"),
+  not(varNameChar),
+  optionalSpacesOrNewline,
+  capture(lazy(() => exprParser), "condition"),
+  optionalSpacesOrNewline,
+  str("then"),
+  not(varNameChar),
+  optionalSpacesOrNewline,
+  capture(lazy(() => exprParser), "thenExpr"),
```
egonSchiele: what counts as an expression? Do all of these count?

- `try foo()`
- `foo catch "default"`
- `read("file") with approve`

egonSchiele: Would pipe operator chains be allowed?

egonSchiele:
Fixed the CI lint failure and the three Copilot findings in e5879057; and here are answers to your questions.

## Review fixes
- **Condition typing (Copilot).** The `if`-expression condition is now type-checked as `boolean`, exactly like a statement `if`/`while` (same `checkConditionType`, same optional-presence-test exception). `if 123 then ...` now errors.
- **Lint.** The `check()` method went over the 150-line limit; extracted the `if`-expression restrictions and the doc-string interpolation check into their own methods.
- **Comments.** Fixed the parser comment (now points at `checkIfExpressionRestrictions`) and the codegen comment (flatness is enforced by the type checker, not parsing).

## "What counts as an expression?" — your questions

Branches are parsed with the **full expression parser**, so anything that *is* an expression is allowed. I verified each:

| branch | allowed? | why |
|---|---|---|
| `try foo()` | ✅ | `try` is an expression atom |
| `foo() catch "default"` | ✅ | `catch` is a binary operator |
| `a \|> b` (pipe chains) | ✅ | `\|>` is a binary operator |
| `read("file") with approve` | ❌ | `with` is a statement modifier, **not** in the `Expression` union — it doesn't parse as a branch |

So `try`/`catch`/`|>` work (e.g. `if ok then try fetch() else fallback`); `with approve` doesn't, by nature.

**My recommendation: keep branches as full expressions.** The readability problem you're guarding against is *nested conditionals*, and that's already forbidden (no nested `if`, no `else if`, no conditional spread). `try`/`catch`/`|>` in a branch are ordinary expressions, not the ternary-soup pattern — and restricting to some "simple expression" subset would be hard to define and inconsistent with how expressions compose everywhere else in the language.

That said, it's your call — if you'd rather branches be limited (say, no `|>`/`catch`, only "atomic" values), I can add that to `checkIfExpressionRestrictions` the same way the nesting/spread guards work. Just let me know where you'd draw the line.

egonSchiele:
Reworked per the interrupt discussion (6ae4a7cf). Reviewing the tests uncovered that the ternary-based form **silently swallowed interrupts** thrown from a branch — unacceptable, since interrupts must work everywhere.

So this is now sugar for the **stepped `if`**: `if c then a else b` lowers to the same `runner.ifElse` machinery a `match` expression uses (not a ternary), and a fixture locks in that an interrupt from a branch pauses and resumes correctly. The agreed tradeoff: like `match` expressions, it's only valid as a `const`/`let` value or a `return` — not as a nested object value / argument (you can't both be a nested value and be steppable, since stepping needs a statement frame).

I also updated the PR description. One byproduct worth noting: I confirmed `match` **expressions** have the same latent bug — a called-interrupt in a single-expression arm (`"go" => confirm()`) loses the interrupt, because it compiles to `exitMatch(N, await __call(confirm))`. My `if` lowering sidesteps it by binding each branch to a temp at statement position first; `match` could use the same fix in a follow-up. Want me to file that separately?

egonSchiele:
Status after the stepped-form rework (6ae4a7cf):

- **Copilot re-reviewed all 9 files and generated no new comments.** The three earlier Copilot findings were on the ternary code, which the rework replaced:
  - *Condition not type-checked* → the stepped form is an `ifElse`, so its condition is checked as `boolean` via the existing `ifElse` path (verified: `if 123 then …` errors).
  - The two *stale-comment* nits → those comments are gone (that code was reverted).

- **Your branch-expression questions still hold, unchanged.** Branches are now parsed by `ifBranchExprParser` → `exprParser` (guarded so a branch can't *start* with `if`), so it's still "any expression": `try foo()`, `foo() catch "default"`, and `a |> b` all compile; `read("file") with approve` is still rejected (`with` isn't an expression). Recommendation unchanged — keep full expressions (nesting/`else if`/spread are the things that hurt readability, and those are forbidden). Happy to restrict `|>`/`catch` if you'd prefer; just say where to draw the line. (The old inline threads point at ternary-era lines that no longer exist.)

- Filed the latent **`match` single-expression-arm interrupt bug** as #430 (the `if` lowering already sidesteps it via the statement-position temp binding).

CI: lint green; builds running.

PR #429, opened 2026-07-05
Make code agent exec-first, quarantine bash, auto-approve read-only git

egonSchiele:
## What

Makes the agency **code agent exec-first**, quarantines raw `bash` into a dedicated subagent, and **auto-approves read-only git** in the recommended policy.

The trigger was the request to auto-allow read-related git actions. That can only be done safely if git flows through the structured `exec` path: `exec()` raises `std::exec` with a `command` + synthetic `subcommand` (`args[0]`) and runs with no shell, so the policy matcher can pin a rule to `(command, subcommand)`. `bash()` raises `std::bash` with only a raw shell string that can chain (`git log; rm -rf x`), so it can't be reasoned about. This is the same mechanism the existing `AGENCY_SAFE_SUBCOMMANDS` list already uses for the `agency` CLI.

## Changes

- **`defaultPolicy.agency`** — add a conservative read-only git safelist + `gitExecApproveRules()` to `recommendedAutoApprovePolicy`. The matcher only sees `args[0]`, so the list contains **only** subcommands that are read-only regardless of any flag (`status`, `log`, `diff`, `show`, `blame`, `rev-parse`, `ls-files`, ...). Deliberately excluded (they have destructive flag forms, so subcommand-alone would over-approve): `branch` (`-D`), `tag`, `config`, `remote`, `stash`, `worktree`, `reflog`, `symbolic-ref`, plus all mutators. A global-option prefix like `git -C dir log` lands with subcommand `-C` and also still prompts.
- **`code.agency`** — expose `exec` as a tool; remove `bash` from the code agent; add `shellAgent`; rewrite the prompt to be exec-first with documented composition patterns for shell idioms (redirect → `write`, pipe → `exec` `stdin`, chain → sequential `exec`, `cwd` param, glob args).
- **`shell.agency`** (new) — a quarantined shell subagent that owns `bash`, called only when real shell features are needed. `bash` still raises its interrupt and is in no auto-approve policy, so **every command still prompts the user** — the subagent isolates the capability, it does not pre-authorize it.
- **`execPolicy` tests** — git read-only approvals + must-prompt cases (`push`, `branch`, `git -C`, minimal policy).

No new stdlib tools; an `env` param on `exec` is deferred (YAGNI).

## Testing

- `execPolicy` — **12/12 pass** (3 new git-approve, 4 must-prompt).
- `toolWiring` (deterministic provider) — **4/4 pass**; the new toolset builds a valid, unique-named `llm` request.
- `shell.agency` typechecks clean; `make` builds clean.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
@@ -29,6 +29,55 @@ def agencyExecApproveRules(): any[] {
   }
 }
 
+// Read-only `git` subcommands the code agent runs via `exec`. Same
+// (command + subcommand) matching as the agency rules above — the
+// matcher only sees `args[0]`, never later flags, so this list may
+// contain ONLY subcommands that are read-only regardless of any flag.
+// Flag-dependent ones are deliberately excluded (they would over-
+// approve a destructive form): `branch` (-D deletes), `tag`, `config`
+// (writes without --get), `remote` (add), `stash` (bare = push),
+// `worktree` (add), `reflog` (expire/delete), `symbolic-ref` (writes
+// with two args), and all mutators (commit/push/reset/checkout/...).
+// A global-option prefix (`git -C dir log`, `git -c x=y commit`) lands
+// with subcommand "-C"/"-c" and correctly fails to match, so it still
+// prompts.
+static const GIT_SAFE_SUBCOMMANDS = [
+  "status",
+  "log",
+  "diff",
+  "show",
+  "blame",
+  "describe",
+  "shortlog",
+  "rev-parse",
+  "rev-list",
+  "ls-files",
+  "ls-tree",
+  "cat-file",
+  "for-each-ref",
+  "merge-base",
+  "name-rev",
+  "diff-tree",
+  "diff-index",
+  "diff-files",
+  "whatchanged",
+  "grep",
+  "cherry",
+  "count-objects",
+]
+
+def gitExecApproveRules(): any[] {
+  return map(GIT_SAFE_SUBCOMMANDS) as sub {
+    return {
+      match: {
+        command: "git",
+        subcommand: sub
+      },
+      action: "approve"
+    }
+  }
+}
+
 export static const minimalAutoApprovePolicy = {
   "std::memory::remember": [{
     action: "approve"
@@ -50,6 +99,9 @@ export static const minimalAutoApprovePolicy = {
 
 export static const recommendedAutoApprovePolicy = {
   ...minimalAutoApprovePolicy,
+  // Re-declare `std::exec` (the spread above set it to the agency rules
+  // only) so it carries BOTH the agency and read-only git safelists.
+  "std::exec": [...agencyExecApproveRules(), ...gitExecApproveRules()],
   "std::read": [{
     action: "approve"
   }],
--- a/packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
@@ -17,18 +17,19 @@ import { filter } from "std::array"
 import { edit } from "std::fs"
 import { setAgentCwd, getAgentCwd } from "std::index"
 import { remember, recall, setMemoryId } from "std::memory"
-import { exec, ls, glob, grep, bash } from "std::shell"
+import { exec, ls, glob, grep } from "std::shell"
 import { skillsDir } from "std::skills"
 import { systemMessage } from "std::thread"
 
 import { agentNames } from "../lib/config.agency"
 import { withWebSearch } from "../lib/search.agency"
 import { oracleAgent } from "./oracle.agency"
+import { shellAgent } from "./shell.agency"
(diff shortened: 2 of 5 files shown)


packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
```diff
@@ -29,6 +29,55 @@ def agencyExecApproveRules(): any[] {
...
+  "ls-tree",
+  "cat-file",
+  "for-each-ref",
+  "merge-base",
+  "name-rev",
+  "diff-tree",
+  "diff-index",
+  "diff-files",
+  "whatchanged",
+  "grep",
```
egonSchiele, line 64: **`git grep` is not read-only regardless of flags — it enables arbitrary command execution.**

`git grep -O<pager>` / `--open-files-in-pager[=<pager>]` runs the given pager as a program (via the shell) instead of printing grep output. Since the matcher only sees `args[0]` (`"grep"`), a call like:

```
exec("git", ["grep", "-Otouch /tmp/pwned #", "def ", "--", "CLAUDE.md"])
```

lands with `subcommand = "grep"`, matches this safelist, and is **auto-approved with no user prompt** — executing an arbitrary command. (The auto-mode classifier in this very repo blocked my proof-of-concept as "arbitrary command execution," which is a good sign of how real this is.)

This directly violates the invariant stated in the comment above ("ONLY subcommands that are read-only regardless of any flag"). `grep` has a destructive/exec flag form just like the deliberately-excluded `branch`/`stash`/etc. Recommend removing `grep` from `GIT_SAFE_SUBCOMMANDS`.

packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
```diff
@@ -29,6 +29,55 @@ def agencyExecApproveRules(): any[] {
...
+// (writes without --get), `remote` (add), `stash` (bare = push),
+// `worktree` (add), `reflog` (expire/delete), `symbolic-ref` (writes
+// with two args), and all mutators (commit/push/reset/checkout/...).
+// A global-option prefix (`git -C dir log`, `git -c x=y commit`) lands
+// with subcommand "-C"/"-c" and correctly fails to match, so it still
+// prompts.
+static const GIT_SAFE_SUBCOMMANDS = [
+  "status",
+  "log",
+  "diff",
```
egonSchiele, line 47: **The diff-family subcommands write to arbitrary files via `--output`, so they aren't read-only regardless of flags either.**

`git diff`, `git log`, `git show`, `git whatchanged` (and the `diff-tree`/`diff-index`/`diff-files` plumbing) all accept `--output=<file>`, which writes the diff to that path instead of stdout, silently overwriting whatever is there:

```
exec("git", ["diff", "--output=/Users/me/.bashrc", "HEAD~5"])
```

`subcommand = "diff"` matches this safelist and is auto-approved with no prompt, clobbering `~/.bashrc`. This breaks the same "read-only regardless of any flag" invariant the comment claims to enforce.

The core issue: the exclusion analysis in the comment above only covers *mutating subcommands* (`branch -D`, `stash`, etc.) but the *included* read subcommands also have flag forms that write files (`--output`) or exec programs (`grep -O`). Matching on `args[0]` alone can't distinguish `git diff HEAD` from `git diff --output=<file>`. Worth reconsidering whether these can be safelisted at all without also constraining the flags.

egonSchiele:
Reviewed the exec-first / bash-quarantine / read-only-git-safelist change. The structure is clean and the `shell.agency` quarantine + prompt rewrite read well; the exec `stdin`/`cwd` params the new prompt teaches are all real (verified against `stdlib/shell.agency`), and the spread-then-redeclare of `std::exec` in `recommendedAutoApprovePolicy` is correct and tested.

My concern is with the central safety claim of the PR — that `GIT_SAFE_SUBCOMMANDS` contains "ONLY subcommands that are read-only regardless of any flag." Two of the safelisted subcommands have flag forms that break that invariant (arbitrary command execution and arbitrary file writes), and both would be auto-approved with no prompt. Details inline.

packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
egonSchiele: This may be an argument for creating specific tools for git diff / branch

packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
```diff
@@ -224,12 +225,31 @@ live in `.agency` files you can `read` directly.
    more than one change to make to the same file. The user will be
    prompted to approve every write.
 6. After every change, run `typecheck` (rule 3 above) and any
-   relevant tests via `bash`.
-7. Use `bash` to run project-level commands (tests, git, formatters,
-   package managers). Prefer the dedicated file tools above for file
-   operations; use `bash` for everything else. The user is asked to
-   approve each command.
-8. You have a persistent knowledge graph scoped to coding work. Call
+   relevant tests via `exec`.
```
egonSchiele, line 228: Try to keep this prompt short please, so we don't blow up the context

PR #431, opened 2026-07-05
Fix interrupt swallowed in single-expression match arm (#430)

### LINKED ISSUE #430: Interrupt from a call in a single-expression match arm is silently swallowed
## Summary

A function call in a **single-expression `match` arm** that throws an interrupt has its interrupt **silently swallowed** — the program continues as if no interrupt occurred, instead of pausing and resuming.

## Repro

```ts
def confirm(): string {
  interrupt("proceed?")
  return "approved"
}

node main(x: string): string {
  const val = match(x) {
    "go" => confirm()
    _    => "skip"
  }
  return "got:${val}"
}
```

Running `main("go")` and approving the interrupt should yield `"got:approved"`. Instead the interrupt never fires (0 interrupts occur) and the result is wrong.

## Root cause

Interrupts propagate as a **value carrying interrupt markers**. Statement-position calls get a `hasInterrupts(...)` check that halts and propagates:

```js
__stack.locals.val = await __call(confirm, {...});
if (hasInterrupts(__stack.locals.val)) { runner2.halt({ ..., data: __stack.locals.val }); return; }
```

But a single-expression arm compiles the call straight into `exitMatch` as an argument — with **no** check in between:

```js
// single-expression arm — interrupt result handed straight to exitMatch:
runner3.exitMatch(1, await __call(confirm, {...}));
```

So `exitMatch` stores the interrupt value as the match's yielded value and the interrupt is consumed.

A **block arm** works, because the call is at statement position:

```ts
"go" => {
  const tmp = confirm()   // statement → gets the hasInterrupts check
  return tmp
}
```

## Suggested fix

Lower a single-expression arm the way a block arm's temp is generated: bind the arm expression to a temp at statement position (`const __tmp = <expr>`) and yield the temp, so the call gets the interrupt-propagation check. This is exactly the fix used for `if ... then ... else` expressions in #428 (`lowerIfExpressionCore` in `lib/lowering/patternLowering.ts`); the same shape could be applied to `rewriteArmForYield`'s single-expression path.

## Discovered in

PR #428 (if-expressions) — while reviewing interrupt behavior in branches. The `if`-expression lowering sidesteps this; `match` single-expression arms still have it.

egonSchiele:
## Summary

Closes #430. An interrupt thrown from a call in a **single-expression** `match` arm was silently swallowed.

```ts
def confirm(): string {
  interrupt("proceed?")
  return "approved"
}

node main(x: string): string {
  const val = match(x) {
    "go" => confirm()   // interrupt from here was lost
    _    => "skip"
  }
  return "got:${val}"
}
```

## Root cause

A single-expression arm lowered directly to `matchYield(<expr>)`, which codegen turns into `runner.exitMatch(id, await __call(confirm))`. The interrupt-carrying value was handed straight to `exitMatch` with **no `hasInterrupts` check**, so the interrupt never propagated — `main("go")` returned `"got:approved"` without ever pausing.

Block arms never had this bug: a `return` inside `{ ... }` lowers to a statement-position temp, which *does* get the check. Single-expression arms were the gap.

## Fix: hoist every single-expression arm uniformly

Lowering now hoists **every** single-expression arm's value to a temp at statement position before yielding it:

```
"go" => confirm()   ⟶   const __armval = confirm()   // statement position → gets hasInterrupts check
                        matchYield(__armval)
```

An arm value therefore behaves **exactly like the same expression written as `const t = <expr>`** — it inherits whatever interrupt propagation a normal assignment has. (An earlier revision tried to *detect* which arm values could interrupt and hoist only those; that was brittle — a missed case silently reintroduces the swallow — so per review it was dropped in favor of unconditional hoisting.)

### Preserving type-checker behavior

Two things the checker got from the old bare-yield shape are preserved:

- **Literal precision + per-arm discriminant narrowing.** The checker types each arm from the yielded value; a temp ref would widen `"a"` to `string` and drop narrowing. `MatchYield` now carries a `typeSource` field with the arm's original expression — the checker types the arm from that, while codegen uses the temp ref in `value`.
- **Graph-node-transition guard.** `"a" => target()` (a node call) is control flow, not a value. The temp binding is tagged `matchArmValueTemp`, and codegen re-applies the guard to it, so a node call in a match arm stays a compile error rather than silently emitting a `goToNode` transition inside the arm.

### Known limitation (pre-existing, not introduced here)

A method call whose callback interrupts (`x.run()`, `items.map(f)`) is not guarded — but this is a **language-wide** gap: a plain `const r = items.map(f)` in ordinary code is unguarded too. Uniform hoisting makes arms behave identically to statement-position assignments; closing the general method-call gap is separate from #430.

## Testing

- **Execution** (`tests/agency/substeps/interrupt-in-match-arm-call.agency`, 4): a matched-arm call and a wildcard-arm call both pause on the interrupt and resume; the non-interrupting paths return their literals.
- **Lowering** (unit): every arm kind (call, literal, variable ref, method call) hoists to a tagged temp + yields it, with `typeSource` carrying the original expression.
- **Typecheck**: literal-union yields, discriminant-narrowed field-access yields, and per-arm error anchoring all still pass (via `typeSource`).
- **Guard**: a node-call arm still throws `match arm cannot return a graph node transition`.
- **Fixture**: `matchExpression.mjs` updated for the uniform-hoist shape.
- Full unit suite: **7180 passed**, 0 failures. Lint clean.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -1367,20 +1367,32 @@ export class TypeScriptBuilder {
    *  the match result via `runner.exitMatch` and control unwinds to the owning
    *  ifElse rather than the enclosing function. Never produced by the parser —
    *  only by pattern lowering (Task 6). */
-  private processMatchYield(node: MatchYield): TsNode {
-    // A graph-node call compiles to a goto/halt transition statement (see
-    // generateNodeCallExpression), which is control flow — it cannot serve as
-    // the value argument of `runner.exitMatch(id, <value>)`. Reject it with a
-    // clear compile error rather than emitting invalid TypeScript.
+  /** A graph-node call is a control-flow transition, not a value, so it cannot
+   *  be the result of a match/if arm. `processMatchYield` catches the common
+   *  case where the arm value is yielded directly; but a single-expression arm
+   *  whose value may interrupt is hoisted to a temp binding first (#430), which
+   *  hides the call from that check — so `_processAssignmentInner` re-runs this
+   *  guard on the marked binding. Shared here to keep one error message. */
+  private assertMatchArmValueNotGraphNode(
+    value: Expression | MessageThread | undefined,
+  ): void {
     if (
-      node.value?.type === "functionCall" &&
-      this.names.isGraphNode(node.value.functionName)
+      value?.type === "functionCall" &&
+      this.names.isGraphNode(value.functionName)
     ) {
       throw new Error(
         "a match arm cannot return a graph node transition; a node call is " +
           "control flow, not a value — use if/else statements for node dispatch",
       );
     }
+  }
+
+  private processMatchYield(node: MatchYield): TsNode {
+    // A graph-node call compiles to a goto/halt transition statement (see
+    // generateNodeCallExpression), which is control flow — it cannot serve as
+    // the value argument of `runner.exitMatch(id, <value>)`. Reject it with a
+    // clear compile error rather than emitting invalid TypeScript.
+    this.assertMatchArmValueNotGraphNode(node.value);
     const value = node.value ? this.processNode(node.value) : ts.id("undefined");
     // Plain-mode (handler) match expressions wrap their arms in an async IIFE
     // (see processMatchExpressionPlain): a yield is a real `return` out of that
@@ -2995,6 +3007,13 @@ export class TypeScriptBuilder {
 
   private _processAssignmentInner(node: Assignment): TsNode {
     const { variableName, typeHint, value } = node;
+
+    // A single-expression match/if arm whose value may interrupt is lowered to
+    // this temp binding so the call sits at statement position (#430). The node
+    // call is now hidden from `processMatchYield`'s guard, so re-apply it here.
+    if (node.matchArmValueTemp) {
+      this.assertMatchArmValueNotGraphNode(value);
+    }
 
     if (value.type === "functionCall" && value.functionName === "llm") {
       return this.processLlmCall(variableName, typeHint, value, node.scope!);
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -541,7 +541,34 @@ class PatternLowerer {
     arm: MatchBlockCase,
   ): AgencyNode[] {
     if (body.length === 1 && isExpressionNode(body[0])) {
-      return [{ type: "matchYield", matchId, value: body[0] as Expression, loc: body[0].loc }];
+      const expr = body[0] as Expression;
+      // A call arm (`"go" => confirm()`) must bind its value to a temp at
+      // STATEMENT position, then yield the temp (#430): a bare
+      // `matchYield(await __call(f))` hands the call result straight to
+      // `exitMatch` with no interrupt-propagation check, so an interrupt is
+      // swallowed. Non-call values (literals, refs, ...) stay a bare yield —
+      // they can't interrupt, and keeping them direct preserves the per-arm
+      // unwidened literal check and the graph-node-transition guard, both of
+      // which read the yielded value.
+      if (armValueMayInterrupt(expr)) {
+        const tmp = this.freshName("armval");
+        const binding: Assignment = {
+          type: "assignment",
+          variableName: tmp,
+          declKind: "const",
+          value: expr,
+          loc: expr.loc,
+          matchArmValueTemp: true,
+        };
+        const yielded: MatchYield = {
+          type: "matchYield",
+          matchId,
+          value: varRef(tmp, expr.loc),
+          loc: expr.loc,
+        };
+        return [binding, yielded];
+      }
+      return [{ type: "matchYield", matchId, value: expr, loc: expr.loc }];
     }
     const rewritten = this.rewriteReturnsToYields(body, matchId);
     if (!this.alwaysYields(rewritten)) {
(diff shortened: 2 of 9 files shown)


packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -1098,6 +1125,36 @@ function isExpr(v: unknown): v is Expression {
...
+/**
+ * True when a single-expression match-arm value involves a function/method call
+ * — so it may return an interrupt that must propagate rather than be swallowed
+ * by `exitMatch` (#430). Such arms are bound to a temp at statement position
+ * before yielding; everything else (literals, variable refs, plain field/index
+ * access) stays a bare yield, which cannot interrupt and keeps the yielded value
+ * direct for the per-arm unwidened-literal check and the graph-node-transition
+ * guard. `try`/`raise`/binop calls surface as one of these node types too.
+ */
+function armValueMayInterrupt(expr: Expression): boolean {
```
egonSchiele: This is brittle and liable to break. Why not just do this always instead of trying to detect whether the arm value may interrupt?

packages/agency-lang/tests/agency/substeps/interrupt-in-match-arm-call.agency
```diff
@@ -0,0 +1,16 @@
+// Regression for #430: an interrupt thrown by a function call in a
+// SINGLE-EXPRESSION match arm must pause/resume, not be swallowed. The arm
+// value is bound to a temp at statement position (like a block arm), so the
+// call gets the interrupt-propagation check.
+def confirm(): string {
+  interrupt("proceed?")
+  return "approved"
+}
+
+node main(x: string): string {
```
egonSchiele: Notice that in your code above you're trying to detect whether the arm value may interrupt, by listing out some cases. But you're not even testing all the cases you have listed out. So how can it possibly be confident that for the cases you haven't listed out, interrupts will not break?
This is exactly why the interrupt detection logic you have is brittle and hard to rely on. I would suggest removing it and doing the lowering always. I'd also suggest adding more test cases here.

packages/agency-lang/lib/lowering/patternLowering.ts
egonSchiele: Agreed — done in 913da24d. Removed the `armValueMayInterrupt` heuristic entirely; every single-expression arm now hoists to a temp unconditionally. An arm value now behaves exactly like the same expression written as `const t = <expr>` at statement position, so it inherits whatever interrupt propagation a normal assignment has — no guessing which values can interrupt.

Two things the type checker previously got from the bare-yield shape had to be preserved to make always-hoist correct (the reason I'd originally reached for detection):

1. **Literal precision + per-arm discriminant narrowing.** The checker types each arm from the yielded value; a temp ref would widen `"a"` to `string` and drop narrowing (this broke 5 typechecker tests when I first flipped to always-hoist). Fixed by giving `MatchYield` a `typeSource` field holding the arm's original expression — the checker types the arm from that, codegen still uses the temp ref in `value`. Restores the exact prior typing behavior.
2. **Graph-node-transition guard.** The temp binding is tagged `matchArmValueTemp` and codegen re-applies the guard to it.

Full suite green (7180 passed), and I enriched the execution test to cover a matched-arm call and a wildcard-arm call both interrupting/resuming.

packages/agency-lang/tests/agency/substeps/interrupt-in-match-arm-call.agency
egonSchiele: Done — added those. The execution test now exercises a matched-arm call and a wildcard-arm call both interrupting and resuming (4 cases), and since the lowering is now uniform (every single-expression arm hoists, no case list) there's no longer a set of "listed cases" to under-test. See the reply on the lowering comment for the details.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2996,6 +3008,13 @@ export class TypeScriptBuilder {
...
     const { variableName, typeHint, value } = node;
 
+    // A single-expression match/if arm whose value may interrupt is lowered to
+    // this temp binding so the call sits at statement position (#430). The node
+    // call is now hidden from `processMatchYield`'s guard, so re-apply it here.
+    if (node.matchArmValueTemp) {
+      this.assertMatchArmValueNotGraphNode(value);
+    }
+
     if (value.type === "functionCall" && value.functionName === "llm") {
```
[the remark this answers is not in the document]
egonSchiele: Good catch, and it pushed me the right way. You're correct that `_processAssignmentInner` only emits the interrupt guard for a direct `functionCall` (and `try`/binOp via `rhsMayBubbleInterrupts`), not for a `valueAccess` with a method/call in its chain.

But this is a **pre-existing, language-wide** gap, not something the arm hoisting introduces. A plain `const r = items.map(confirm)` in an ordinary node body compiles to `__callMethod(...)` with no `hasInterrupts` guard either — I verified this against the current compiler. So a method-call callback that interrupts is unguarded everywhere in the language today, independent of match arms.

That actually points to the right invariant, which the latest commit adopts: instead of trying to detect which arm values can interrupt (brittle, and exactly the false-coverage you're describing — my method-call case was hoisted but never guarded), every single-expression arm is hoisted unconditionally, so an arm value behaves **identically to the same expression at statement position**. Direct-call arms (this issue's actual repro) now propagate correctly; method-call arms behave exactly like a method-call assignment elsewhere — no better, no worse.

Closing the general method-call-callback guard gap is worth doing, but it's a separate change (it touches normal assignment codegen, not match lowering) and I'd rather scope it to its own issue than fold it into #430.

PR #442, opened 2026-07-07
feat(stdlib): add std::data/finance evidence connectors (GDELT, FRED, EDGAR, DBnomics)

egonSchiele:
## What

Adds four pure-Agency standard-library **evidence connectors** under a new `std::data/finance/*` namespace — the foundation for a future calibrated event-forecasting agent (research → probability), which is the LLM-forecasting approach with the strongest real-world track record (Preseen / FutureSearch).

| Module | Source | Kind | Auth |
|---|---|---|---|
| `std::data/finance/gdelt` | [GDELT DOC 2.0](https://blog.gdeltproject.org/gdelt-doc-2-0-api-debuts/) | news articles | none |
| `std::data/finance/fred` | [FRED](https://fred.stlouisfed.org/docs/api/fred/) | US macro time-series | free key (`FRED_API_KEY`) |
| `std::data/finance/edgar` | [SEC EDGAR](https://www.sec.gov/search-filings/edgar-application-programming-interfaces) | company filings | none (`User-Agent`) |
| `std::data/finance/dbnomics` | [DBnomics](https://db.nomics.world/) | world macro time-series | none |

Mnemonic baked into each module's docstring: **GDELT for the narrative, EDGAR for company facts, FRED for US macro, DBnomics for the rest of the world's macro.**

## How

Each connector is a single `.agency` file, no TypeScript backing, following the `weather`/`wikipedia` pattern:

- One module-level constrained client: `fetchJSON.partial(baseUrl, allowedDomains).preapprove()` — so exactly one *source-specific* approval prompt fires (the inner `std::http::fetchJSON` prompt is preapproved).
- A source-specific `effect` (`std::gdelt` / `std::fred` / `std::edgar` / `std::dbnomics`), registered in the `Network` effect set in `std::capabilities`.
- Logic split into **pure, independently testable functions**: `buildXPath` (URL/query construction, `encodeURIComponent`-escaped), `parseX` (JSON→typed reshaping, total on malformed input), and `xFinalize` (turns the fetch `Result` into the final `Result`). The public function is a thin orchestrator.
- Faithful, exported per-source return types (`NewsArticle`, `FredSeries`, `Filing`, `DbnomicsSeries`, …) — no premature unified "evidence" envelope (that belongs to the forecaster layer).

## Testing

- **Pure-function tests** (agency-js, no network, no LLM): build/parse/finalize incl. edge cases — URL encoding, FRED `"."`→`null`, EDGAR columnar zip + form filter + `limit` + unknown ticker, DBnomics ragged arrays + missing `series`, sparse-record totality.
- **Offline interrupt/effect tests** per connector: assert each raises its own effect with the right message/payload, then `reject()` (also exercises the orchestrator's pre-fetch half). Guards the safety-relevant surface — a typo'd effect label would break user policies silently otherwise.
- Sample fixtures are **real trimmed captures** (Apple EDGAR submissions, BLS/CPI DBnomics, UNRATE FRED).
- Opt-in **live smoke** for GDELT (env-gated, `AGENCY_LIVE_TESTS`); its handler rejects any non-`std::gdelt` interrupt, so a dropped `.preapprove()` fails it.

## Known limitation: FRED key in statelog

FRED v1 accepts the API key only as a URL query parameter, so with `observability` enabled the key appears in statelog `fetchJSON` interrupt payloads (the URL is logged). This is a **general property of any secret-in-URL `fetchJSON` use**, not FRED-specific, and it is gated behind observability being on (off by default).

FRED v2 offers `Authorization: Bearer` header auth (headers are *not* logged), but it turned out to be a differently-structured, less-documented, rate-limiting REST API — not a safe base for a stdlib connector. So this PR ships stable **v1**, documents the caveat in the module docs, and defers a **general URL-secret redaction for statelog** to a follow-up (filed separately).

## Notes

- Spec + implementation plan included under `docs/superpowers/`.
- Parser gotcha found & worked around: you can't chain a method directly onto a string literal (`"${x}".padStart(...)`) — bind to a variable first.
- Generated stdlib reference pages + a new `data` sidebar group.


egonSchiele:
Follow-up for the FRED statelog key-leak (general URL-secret redaction): #443

packages/agency-lang/stdlib/data/finance/fred.agency
```diff
@@ -0,0 +1,199 @@
...
+  @param observationEnd - Optional latest date, "YYYY-MM-DD" (empty for no bound)
+  @param frequency - Optional frequency aggregation, e.g. "m", "q", "a" (empty for native)
+  @param units - Optional units transform, e.g. "pch" for percent change (empty for "lin")
+  @param limit - Optional max observations from the oldest, 0 means no limit
+  """
+  const apiKey = env("FRED_API_KEY") ?? ""
+  if (apiKey == "") {
+    return failure("FRED_API_KEY is not set. Get a free key at https://fred.stlouisfed.org/docs/api/api_key.html")
+  }
+  return interrupt std::fred("Fetch this FRED series?", { seriesId: seriesId })
```
egonSchiele, line 192: **FRED key leak — disclosed, but the promised regression test is missing.** Good that the interrupt payload carries only `seriesId`, and the URL-in-statelog leak is documented as an observability-gated, off-by-default limitation with a follow-up filed — acceptable to ship.

However, the implementation plan (Task 2, Step 7) specified committing `tests/agency/data-finance-fred-leak.agency` as a sentinel-key regression guard, and that file isn't in the diff (there's no `fred`/`leak` test under `tests/agency/`). The payload-only-carries-`seriesId` invariant is exactly the safety-relevant surface — worth adding the guard so a future change can't silently widen the leak, or explicitly deferring it in the follow-up ticket.

packages/agency-lang/stdlib/data/finance/fred.agency
```diff
@@ -0,0 +1,199 @@
...
+/** Append "&key=value" only when value is non-empty (value is URL-encoded). Pure. */
+def appendParam(path: string, key: string, value: string): string {
+  if (value == "") {
+    return path
+  }
+  return "${path}&${key}=${encodeURIComponent(value)}"
+}
+
+/** Build the series/observations query path. Pure — no network. */
+export safe def buildFredObservationsPath(seriesId: string, apiKey: string, observationStart: string, observationEnd: string, frequency: string, units: string, limit: number): string {
```
egonSchiele, line 90: **Encoding inconsistency.** `seriesId` is `encodeURIComponent`-escaped but `api_key` is interpolated raw (`&api_key=${apiKey}`). Fine in practice (keys are URL-safe), but it's the one user-supplied value not run through `appendParam`'s encoder, which slightly contradicts the plan's stated "always encode user input" principle and makes an injection audit reason case-by-case.

packages/agency-lang/stdlib/data/finance/fred.agency
```diff
@@ -0,0 +1,199 @@
...
+  return appendParam(path, "limit", limitStr)
+}
+
+/** Build the series (metadata) query path. Pure — no network. */
+export safe def buildFredSeriesPath(seriesId: string, apiKey: string): string {
+  return "series?series_id=${encodeURIComponent(seriesId)}&api_key=${apiKey}&file_type=json"
+}
+
+/** Convert a FRED value string to a number, or null for the missing marker ".". Pure. */
+export safe def toFredValue(raw: string): number | null {
```
egonSchiele, line 109: **Minor:** `Number(raw)` maps a non-numeric, non-`"."` value to `NaN`, which serializes to JSON `null` — indistinguishable from a genuine missing value. Acceptable given FRED's contract, but a defensive `isNaN` → `null` would document the intent explicitly.

packages/agency-lang/stdlib/data/finance/edgar.agency
```diff
@@ -0,0 +1,171 @@
...
+  @param formType - Optional filing form to filter by, e.g. "10-K" (empty for all forms)
+  @param limit - Maximum filings to return (default 20)
+  """
+  const cik10 = cik.padStart(CIK_WIDTH, "0")
+  return interrupt std::edgar("Fetch SEC filings for this company?", { company: cik })
+  const path = buildSubmissionsPath(cik10)
+  return edgarSubmissionsFinalize(cik10, formType, limit, submissionsClient(path: path))
+}
+
+export safe def edgarFilings(ticker: string, formType: string = "", limit: number = 20): Result {
```
egonSchiele: **Performance:** `edgarFilings(ticker, ...)` re-downloads the full `company_tickers.json` (multi-MB, ~10k entries) on every call, then linear-scans it in `parseTickerMap`. For the intended consumer — an LLM agent calling this as a tool repeatedly — that's significant egress/latency per lookup. Caching is an explicit v1 non-goal so not blocking, but consider noting in the docstring that `edgarFilingsByCik` avoids the ticker fetch when the CIK is already known.

packages/agency-lang/stdlib/data/finance/gdelt.agency
```diff
@@ -0,0 +1,95 @@
...
+// ends at ".../doc" (not ".../doc/doc") because resolveUrl inserts a "/" before the
+// path, and buildGdeltPath returns a path beginning "doc?..." (see resolveUrl in
+// lib/stdlib/http.ts).
+const gdeltClient = fetchJSON.partial(
+  baseUrl: "https://api.gdeltproject.org/api/v2/doc",
+  allowedDomains: ["api.gdeltproject.org"]
+).preapprove()
+
+/** Build the GDELT DOC query path (query is URL-encoded). Pure — no network. */
+export safe def buildGdeltPath(query: string, maxRecords: number, timespan: string): string {
```
egonSchiele, line 52: **Minor consistency:** `query` is encoded but `timespan` is interpolated raw (`&timespan=${timespan}`). Low risk given the closed value set (`"3d"`, `"1w"`, …), but for consistency with the encode-all principle either `encodeURIComponent` it or add a comment that its value space is closed.

egonSchiele (commented):
Code review — solid, approvable work. The pure/orchestrator split is idiomatic, parsers are total on malformed input, and the safety surface (per-source effects in `Network`, per-client `allowedDomains`, secret-free interrupt payloads) is right. The pattern correctly matches `weather.agency`. Inline comments below cover one disclosed security limitation, a performance note, and several minor consistency/coverage items — none blocking.

packages/agency-lang/docs/site/stdlib/data/finance/fred.md
```diff
@@ -0,0 +1,302 @@
...
+
+### FredObservation
+
+A single FRED observation. `value` is null when FRED reports the value as missing (".").
+
+```ts
+/** A single FRED observation. `value` is null when FRED reports the value as missing ("."). */
+export type FredObservation = {
+  date: string;
+  value?: number
```
egonSchiele, line 54: **Docs accuracy (pre-existing generator behavior, surfacing here):** the source declares `value: number | null`, but the generated page renders it as `value?: number` — misrepresenting an explicit `null` as an absent/optional field. Same for `DbnomicsObservation`. Not a bug in this PR, but the new pages misstate the null contract; may warrant a separate `agency doc` fix.

packages/agency-lang/tests/agency-js/data-finance-gdelt-live/test.js
```diff
@@ -0,0 +1,13 @@
+import { liveGdelt } from "./agent.js";
+import { writeFileSync } from "node:fs";
+
+// Opt-in: only hits the live GDELT API when AGENCY_LIVE_TESTS is set. The handler
+// rejects any non-std::gdelt interrupt, so a dropped .preapprove() (which would surface
```
egonSchiele, line 5: **Coverage note (self-disclosed):** the regression that matters most for safety — a dropped `.preapprove()` re-surfacing the inner `std::http::fetchJSON` prompt — is only caught by this opt-in, env-gated live GDELT smoke. FRED/EDGAR/DBnomics have no equivalent guard until the fetch mock lands. Flagging so it isn't forgotten.

egonSchiele:
Thanks both — pushed 2e57544 addressing the review. Summary:

**Correctness (fixed):**
- **Parsers total on null root** (Copilot ×6): `parseGdelt`, `parseFredObservations`, `parseFredInfo`, `parseDbnomics`, `parseTickerMap`, `parseSubmissions` now guard the root with `?? {}` (and `Object.values(raw ?? {})`). Added an explicit null-input test per parser so the "total" claim is enforced.
- **`dbnomics` period fallback**: `period: periods[i] ?? ""` so `DbnomicsObservation.period` stays a string.
- **`toFredValue` NaN → null** (isNaN guard): a non-numeric, non-`"."` value now maps to `null` explicitly rather than a `NaN` that JSON-serializes to `null` by accident.

**Consistency (fixed):**
- URL-encode `timespan` (gdelt) and `api_key` (fred, both paths) — every user-supplied value now goes through `encodeURIComponent`, so an injection audit no longer has to reason case-by-case.

**FRED key leak (fixed the missing guard):**
- Instead of the plan's statelog-probe file (which needed `observability` + network and would be flaky in CI), I strengthened the offline FRED interrupt test to assert the invariant directly: with a sentinel key set, the `std::fred` interrupt payload contains **only** `seriesId` and never the key. That's the safety-relevant surface, and it now fails CI if a future change widens it.

**EDGAR perf note (documented):**
- `edgarFilings` docstring now tells callers to prefer `edgarFilingsByCik` when the CIK is known, to skip the multi-MB ticker-map download. Caching stays a v1 non-goal.

**Deferred (not this PR):**
- **Generated docs render `T | null` as `field?: T`** — confirmed pre-existing `agency doc` behavior affecting every nullable field repo-wide, so it doesn't belong in a connector PR. Filed #444.
- **preapprove regression only caught by the live GDELT smoke** for now — full CI coverage of "exactly one prompt per call" awaits the fetch mock; noted so it isn't forgotten.
- **FRED URL-secret-in-statelog** — the general redaction is #443 (from before).

packages/agency-lang/stdlib/data/finance/edgar.agency
```diff
@@ -0,0 +1,173 @@
...
+}
+
+const CIK_WIDTH = 10
+
+// SEC requires a descriptive User-Agent with contact info. Read once at module init;
+// setting SEC_USER_AGENT after the process starts will not take effect.
+const secUserAgent = env("SEC_USER_AGENT") ?? "agency-lang admin@agency-lang.com"
+
+// One preapproved client per host (fetchJSON.partial locks a single baseUrl).
+const tickerClient = fetchJSON.partial(
```
egonSchiele: don't preapprove anything

packages/agency-lang/stdlib/data/finance/edgar.agency
```diff
@@ -0,0 +1,173 @@
...
+  a direct URL to each primary document. Optionally filter by form type (e.g. "10-K", "8-K").
+  This downloads the SEC ticker->CIK map on each call; if you already know the CIK, prefer
+  edgarFilingsByCik to skip that fetch.
+
+  @param ticker - The company's stock ticker, e.g. "AAPL"
+  @param formType - Optional filing form to filter by, e.g. "10-K" (empty for all forms)
+  @param limit - Maximum filings to return (default 20)
+  """
+  return interrupt std::edgar("Fetch SEC filings for this company?", { company: ticker })
+  return match (tickerClient(path: "files/company_tickers.json")) {
```
egonSchiele: if the *condition* in a `match` block throws an interrupt, is Agency able to resume from that? See 
- https://agency-lang.com/guide/interrupts.html
- https://agency-lang.com/guide/handlers.html
- packages/agency-lang/docs/dev/interrupts.md

Check `if` statements and parameters to function calls while you're at it. Anything else that we should check? Match arm condition? `if else` condition?

packages/agency-lang/stdlib/data/finance/fred.agency
```diff
@@ -0,0 +1,204 @@
...
+      if (list.length == 0) {
+        return failure("FRED returned no series metadata (check the series id)")
+      }
+      return success(parseFredInfo(body))
+    }
+    failure(err) => failure("FRED request failed: ${err}")
+  }
+}
+
+export safe def fredSeries(seriesId: string, observationStart: string = "", observationEnd: string = "", frequency: string = "", units: string = "", limit: number = 0): Result {
```
egonSchiele: Add `raises` annotations to your functions. Any effect sets you should create?

https://agency-lang.com/guide/effects-and-raises.html

egonSchiele: Should `"frequency"` be a union of specific strings? What other types do you need to tighten in this PR?

egonSchiele:
Addressed the latest round in 5f43230b.

**"Don't preapprove" + "raise both effects" (edgar:52):** removed `.preapprove()` from all four connectors. Each function now raises **both** its source-specific effect and `std::http::fetchJSON`, so the network call is user-gateable rather than silently auto-approved. Verified in the tests: rejecting the domain interrupt short-circuits *before* any fetch; approving it resumes into the `std::http::fetchJSON` interrupt.

**Interrupt-in-condition resumption (edgar:162):** this was the real risk once preapprove is gone. I audited and restructured so **no interrupt sits in a `match` scrutinee, `if`/`while` condition, or call argument** — every fetch is now bound to its own `const` statement, then finalized. `edgarFilings` in particular went from `match (tickerClient(...))` (fetch in the scrutinee) + a nested-arm fetch to two flat statement-level fetches. Audit grep confirms zero client calls inside conditions/args across the four files. The offline tests exercise the actual resume path (approve → resumes to the fetch interrupt) with no network.

**raises + effect sets (fred:166):** added `raises <std::X, std::http::fetchJSON>` to every public function, and exported a `DataFinance = <std::gdelt, std::fred, std::edgar, std::dbnomics>` effect set in `std::capabilities` so callers can gate the whole family in one clause.
- Note: the generated `Throws:` line only shows the source effect, not `std::http::fetchJSON` — the doc generator's static analysis can't trace through a `.partial()` client, so it under-reports the transitive effect. The `raises` clause and runtime behavior are correct (the tests prove both fire); flagging the generator gap.

**Type tightening (fred:166):**
- Public functions now return `Result<NewsArticle[]>` / `Result<FredSeries>` / `Result<FredSeriesInfo>` / `Result<Filing[]>` / `Result<DbnomicsSeries>`.
- On `frequency`/`units` as unions: I tried it and hit a hard limitation — **`""` cannot be a member of a string-literal union** (`type X = "" | "lin"` is a parse error), and `""` is exactly the empty-string sentinel these optional params use for "unset" (Agency has no `undefined`). So a union conflicts with the optional-param pattern and I kept them `string`. The response-side `FredSeriesInfo.frequency` I also kept `string`: FRED's real values include open-ended variants ("Weekly, Ending Friday", "Daily, 7-Day", …), so a literal union would be incomplete and would misrepresent real responses. Happy to revisit if you'd rather I model the closed subset with a fallback.

egonSchiele:
Follow-up on the union question: used your `null`-sentinel idea — `fredSeries` now takes `units: FredUnits | null` and `frequency: FredFrequency | null` (default `null`, mapped to `""` internally). This gets the validated closed-set unions without the `""`-can't-be-a-union-member parse limitation; e.g. `fredSeries(…, "BADFREQ", …)` is now a compile-time type error. Also fixed the CI failure (it was `pnpm test:run`'s strict fixtureTypeCheck, which the earlier docs-build failure had masked): null-guards were typing intermediates as empty `{}`, and the `const client = fetchJSON.partial(...)` pattern warned 'not defined' — switched to direct `fetchJSON(...)` calls with config consts. Full suite green locally.

PR #445, opened 2026-07-07
Type-constrained optimizer mutations: typed optimize const targets

egonSchiele:
## What

`optimize const` targets can now be any literal type — string unions, numbers, booleans, records, arrays — and every value the mutator proposes is validated against the declared type before it is ever written into a candidate.

```ts
type Strategy = "fast" | "cheap" | "thorough"

optimize const strategy: Strategy = "fast"     // proposals must be one of the three
optimize const maxRetries: number = 2          // "three" is rejected, 3 is accepted
optimize const style = "Be concise."           // freeform strings unchanged
```

## How: the language is the judge

There is no hand-written shape checker. Validation is a **one-line probe program**: the proposal is spliced into `const __optimizeProbe: <Type> = <value>` and run through the public `typeCheck()` with a minimal compilation unit carrying the closure's type-alias registry (built by the compiler's own `SymbolTable.build` + `buildCompilationUnit` at discovery). The probe is parsed and typechecked in memory — never executed — and the checker's first diagnostic doubles as the retry feedback the mutator LLM sees. This buys the real record semantics for free: union membership through aliases (including imported ones), nullable-field omission, explicit `null`, and unknown-field rejection.

Design decisions (from the reviewed plan, rev 3):

- **Annotation required in v1.** Non-string literal targets must be annotated; there is no inference. `optimize const n = 5` errors with a fix-it message.
- **Baseline self-test.** If a target's *own initializer* does not pass the probe against its annotation (unresolvable alias, exotic type, pre-existing warning), the target is left unconstrained instead of blocked — the optimizer can never reject a value shaped like the one the user already wrote. This one rule replaces all per-type-variant special-casing.
- **`constraintText: null` is disambiguated by `valueKind`**: freeform for text targets (today's behavior, byte-identical path), unconstrained for literal targets.
- Two gates the probe cannot provide (verified empirically): interpolated strings are rejected in typed values (the undefined-variable pass does not descend into interpolation segments), and proposals must be literals (an unknown bare identifier synthesizes to `any` and typechecks clean) — with bare-word recovery, so an unquoted `fail` is requoted rather than rejected.
- Literal target values carry formatter-exact source text via the public `AgencyGenerator.processNode` (initializer nodes have no `loc` offsets to slice by, and `expressionToString` drops string quotes).

## Known limitation (inherited, pinned as a tripwire test)

A record proposal containing an explicit `null` field skips the deep field check: `synthObject` degrades the literal to `any` when any field synthesizes `any`, and a bare `null` does (it is a `variableName` the scope cannot resolve). This fails open — it can admit an imperfect record, never reject a valid one. A one-line fix in `synthType` (special-casing `variableName "null"` to the null type) would tighten this for the whole language; happy to file it separately.

## Guarantee, end to end

`greedyReflective.test.ts` drives the real pipeline with a fake proposer that insists on an out-of-shape value: `propose` is retried to the attempt cap with the `type-mismatch` reason fed back into each retry's inputs, the iteration lands validation-failed, the champion stays at baseline, and no candidate agent file is ever materialized.

## Testing

- 308 optimizer unit tests green (28 new across `constraint.test.ts`, `targets.test.ts`, `sourceMutator.test.ts`, `mutator.test.ts`, `greedyReflective.test.ts`), full repo vitest suite green, `make` green, `lint:structure` green.
- Freeform string behavior is regression-gated by the pre-existing mutator tests (untouched and passing).

### DIFF
--- a/packages/agency-lang/lib/optimize/constraint.ts
+++ b/packages/agency-lang/lib/optimize/constraint.ts
@@ -1,20 +1,10 @@
-import { AgencyGenerator } from "@/backends/agencyGenerator.js";
 import { GLOBAL_SCOPE_KEY, ScopedTypeAliases, type CompilationUnit } from "@/compilationUnit.js";
 import { parseAgency } from "@/parser.js";
 import { typeCheck } from "@/typeChecker/index.js";
-import type { Expression, TypeAliasEntry, VariableType } from "@/types.js";
+import type { TypeAliasEntry, VariableType } from "@/types.js";
 import { formatTypeHint } from "@/utils/formatType.js";
 
 export type ShapeResult = { ok: true } | { ok: false; reason: string };
-
-/**
- * The parser represents `null` as a variableName node — there is no
- * `{ type: "null" }` in parser output. Single point of truth for the check
- * so no caller ever tests `expr.type === "null"` (dead code).
- */
-export function isNullLiteral(expr: Expression): boolean {
-  return expr.type === "variableName" && expr.value === "null";
-}
 
 /**
  * A minimal CompilationUnit carrying only the alias registry, for
@@ -34,22 +24,12 @@ export function makeProbeUnit(typeAliases: Record<string, TypeAliasEntry>): Comp
   };
 }
 
-const PROBE_VARIABLE = "__optimizeProbe";
+// Chosen to read acceptably if it ever leaks into a diagnostic shown to the
+// mutator LLM ("assignment to 'proposedValue'" is self-explanatory).
+const PROBE_VARIABLE = "proposedValue";
 
 /**
- * Exact source rendering of a literal expression, via the Agency formatter
- * (`processNode` is the same renderer interpolations use, so string quotes
- * and escapes are preserved). NOT `expressionToString`, which drops string
- * quotes and makes `{a: "1"}` and `{a: 1}` render identically. Needed
- * because parsed initializer nodes carry no loc offsets, so the exact text
- * cannot be sliced from the source.
- */
-export function renderLiteralSource(expr: Expression): string {
-  return new AgencyGenerator({}).processNode(expr).trim();
-}
-
-/**
- * Does `valueText` fit the type written as `constraintText`? Decided by the
+ * Does `valueText` fit the type written as `declaredType`? Decided by the
  * language's own typechecker on a one-line probe program — parsed and
  * typechecked in memory, never executed. Any diagnostic (assignability,
  * unknown property, undefined variable inside an interpolation, parse
@@ -62,64 +42,26 @@ export function renderLiteralSource(expr: Expression): string {
  * right direction (a valid value is never wrongly rejected).
  */
 export function checkProposal(
-  constraintText: string,
+  declaredType: string,
   valueText: string,
   typeAliases: Record<string, TypeAliasEntry>,
 ): ShapeResult {
-  const probe = parseAgency(`const ${PROBE_VARIABLE}: ${constraintText} = ${valueText}`, {}, false);
+  const probe = parseAgency(`const ${PROBE_VARIABLE}: ${declaredType} = ${valueText}`, {}, false);
   if (!probe.success) {
+    // The parse failure may be the value OR the rendered type — surface the
+    // parser's own message so retry feedback (and baseline self-test
+    // fallbacks) are debuggable.
     return {
       ok: false,
-      reason: `value does not parse as an Agency expression of type ${constraintText}`,
+      reason: `value does not parse as an Agency expression of type ${declaredType}${probe.message ? `: ${probe.message}` : ""}`,
     };
   }
   const result = typeCheck(probe.result, {}, makeProbeUnit(typeAliases));
   if (result.errors.length === 0) return { ok: true };
   return {
     ok: false,
-    reason: result.errors[0].message.replace(` (assignment to '${PROBE_VARIABLE}')`, ""),
+    reason: result.errors[0].message.replaceAll(` (assignment to '${PROBE_VARIABLE}')`, ""),
   };
-}
-
-const LITERAL_EXPRESSION_TYPES = ["string", "multiLineString", "number", "boolean", "agencyObject", "agencyArray"];
-
-/**
- * True when the expression is a literal an optimize target/proposal may
- * carry: string, number, boolean, null, object, or array. The single gate
- * shared by discovery and the mutator — the probe alone cannot enforce
- * this, because an unknown bare identifier synthesizes to `any` and passes
- * typechecking unflagged.
- */
-export function isLiteralExpression(expr: Expression): boolean {
-  return LITERAL_EXPRESSION_TYPES.includes(expr.type) || isNullLiteral(expr);
-}
-
-/**
- * True when a literal expression contains a string interpolation anywhere —
- * at top level or nested inside object/array literals. Typed replacement
- * values must be self-contained: the mutator cannot know what identifiers
- * exist at the target's declaration site, and the probe's undefined-variable
- * pass does not descend into interpolation segments. Structural traversal
- * only — no type semantics.
- */
-export function hasInterpolation(expr: Expression): boolean {
-  switch (expr.type) {
-    case "string":
-    case "multiLineString":
-      return expr.segments.some((segment) => segment.type !== "text");
-    case "agencyArray":
-      return expr.items.some((item) =>
-        item.type === "splat" ? hasInterpolation(item.value) : hasInterpolation(item),
-      );
-    case "agencyObject":
-      return expr.entries.some((entry) => {
-        if ("type" in entry && entry.type === "splat") return hasInterpolation(entry.value);
-        const kv = entry as { value: Expression };
-        return hasInterpolation(kv.value);
-      });
-    default:
-      return false;
-  }
 }
 
 /**
@@ -128,21 +70,21 @@ export function hasInterpolation(expr: Expression): boolean {
  * round-trip: a rendered type that does not re-parse, or does not accept the
  * target's own original value, leaves the target unconstrained.
  */
-export function renderConstraintText(typeHint: VariableType): string {
+export function renderDeclaredType(typeHint: VariableType): string {
   return formatTypeHint(typeHint);
 }
 
 /**
  * One-line description of a target's allowed values for the mutator prompt.
- * `constraintText === null` means freeform for text targets and
+ * `declaredType === null` means freeform for text targets and
  * unconstrained for literal targets — see the valueKind-first invariant in
  * `OptimizeTarget`.
  */
 export function describeConstraint(target: {
-  constraintText: string | null;
+  declaredType: string | null;
   valueKind: string;
 }): string {
-  if (target.constraintText !== null) return target.constraintText;
+  if (target.declaredType !== null) return target.declaredType;
   if (target.valueKind === "literal") return "any literal value";
   return "free text (any string; keep every ${...} placeholder)";
 }
--- a/packages/agency-lang/lib/optimize/constraint.ts
+++ b/packages/agency-lang/lib/optimize/constraint.ts
@@ -1,0 +1,122 @@
+import { GLOBAL_SCOPE_KEY, ScopedTypeAliases, type CompilationUnit } from "@/compilationUnit.js";
+import { parseAgency } from "@/parser.js";
+import { typeCheck } from "@/typeChecker/index.js";
+import type { Expression, TypeAliasEntry, VariableType } from "@/types.js";
+import { formatTypeHint } from "@/utils/formatType.js";
+
+export type ShapeResult = { ok: true } | { ok: false; reason: string };
+
+/**
+ * The parser represents `null` as a variableName node — there is no
+ * `{ type: "null" }` in parser output. Single point of truth for the check
+ * so no caller ever tests `expr.type === "null"` (dead code).
+ */
+export function isNullLiteral(expr: Expression): boolean {
+  return expr.type === "variableName" && expr.value === "null";
+}
+
+/**
+ * A minimal CompilationUnit carrying only the alias registry, for
+ * typechecking probe programs. CompilationUnit is a plain data type (no
+ * closures — unlike TypeCheckerContext), so this construction is stable.
+ */
+export function makeProbeUnit(typeAliases: Record<string, TypeAliasEntry>): CompilationUnit {
+  return {
+    functionDefinitions: {},
+    typeAliases: new ScopedTypeAliases({ [GLOBAL_SCOPE_KEY]: typeAliases }),
+    graphNodes: [],
+    importedNodes: [],
+    importStatements: [],
+    safeFunctions: {},
+    importedFunctions: {},
+    jsImportedNames: {},
+  };
+}
+
+const PROBE_VARIABLE = "__optimizeProbe";
+
+/**
+ * Does `valueText` fit the type written as `constraintText`? Decided by the
+ * language's own typechecker on a one-line probe program — parsed and
+ * typechecked in memory, never executed. Any diagnostic (assignability,
+ * unknown property, undefined variable inside an interpolation, parse
+ * failure) is a rejection; the first message is the retry feedback for the
+ * mutator LLM.
+ *
+ * Known fail-open, inherited from the language: an object literal containing
+ * an explicit `null` field synthesizes to `any`, which skips the assignment
+ * check entirely — such proposals are accepted unchecked. Fail-open is the
+ * right direction (a valid value is never wrongly rejected).
+ */
+export function checkProposal(
+  constraintText: string,
+  valueText: string,
+  typeAliases: Record<string, TypeAliasEntry>,
+): ShapeResult {
+  const probe = parseAgency(`const ${PROBE_VARIABLE}: ${constraintText} = ${valueText}`, {}, false);
+  if (!probe.success) {
+    return {
+      ok: false,
+      reason: `value does not parse as an Agency expression of type ${constraintText}`,
+    };
+  }
+  const result = typeCheck(probe.result, {}, makeProbeUnit(typeAliases));
+  if (result.errors.length === 0) return { ok: true };
+  return {
+    ok: false,
+    reason: result.errors[0].message.replace(` (assignment to '${PROBE_VARIABLE}')`, ""),
+  };
+}
+
+/**
+ * True when a literal expression contains a string interpolation anywhere —
+ * at top level or nested inside object/array literals. Typed replacement
+ * values must be self-contained: the mutator cannot know what identifiers
+ * exist at the target's declaration site, and the probe's undefined-variable
+ * pass does not descend into interpolation segments. Structural traversal
+ * only — no type semantics.
+ */
+export function hasInterpolation(expr: Expression): boolean {
+  switch (expr.type) {
+    case "string":
+    case "multiLineString":
+      return expr.segments.some((segment) => segment.type !== "text");
+    case "agencyArray":
+      return expr.items.some((item) =>
+        item.type === "splat" ? hasInterpolation(item.value) : hasInterpolation(item),
+      );
+    case "agencyObject":
+      return expr.entries.some((entry) => {
+        if ("type" in entry && entry.type === "splat") return hasInterpolation(entry.value);
+        const kv = entry as { value: Expression };
+        return hasInterpolation(kv.value);
+      });
+    default:
+      return false;
+  }
+}
+
+/**
+ * Render an annotation back to parseable source text for probing and for the
+ * mutator prompt. The baseline self-test in discovery validates the
+ * round-trip: a rendered type that does not re-parse, or does not accept the
+ * target's own original value, leaves the target unconstrained.
+ */
+export function renderConstraintText(typeHint: VariableType): string {
+  return formatTypeHint(typeHint);
+}
+
+/**
+ * One-line description of a target's allowed values for the mutator prompt.
+ * `constraintText === null` means freeform for text targets and
+ * unconstrained for literal targets — see the valueKind-first invariant in
+ * `OptimizeTarget`.
+ */
+export function describeConstraint(target: {
+  constraintText: string | null;
+  valueKind: string;
+}): string {
+  if (target.constraintText !== null) return target.constraintText;
+  if (target.valueKind === "literal") return "any literal value";
+  return "free text (any string; keep every ${...} placeholder)";
+}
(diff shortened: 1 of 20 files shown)


packages/agency-lang/lib/optimize/constraint.ts
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Exact source rendering of a literal expression, via the Agency formatter
+ * (`processNode` is the same renderer interpolations use, so string quotes
+ * and escapes are preserved). NOT `expressionToString`, which drops string
+ * quotes and makes `{a: "1"}` and `{a: 1}` render identically. Needed
+ * because parsed initializer nodes carry no loc offsets, so the exact text
+ * cannot be sliced from the source.
+ */
+export function renderLiteralSource(expr: Expression): string {
+  return new AgencyGenerator({}).processNode(expr).trim();
```
egonSchiele: would it make sense to use `generateAgency` from lib/backends/agencyGenerator.ts instead?

packages/agency-lang/lib/optimize/constraint.ts
```diff
@@ -0,0 +1,148 @@
...
+ * an explicit `null` field synthesizes to `any`, which skips the assignment
+ * check entirely — such proposals are accepted unchecked. Fail-open is the
+ * right direction (a valid value is never wrongly rejected).
+ */
+export function checkProposal(
+  constraintText: string,
+  valueText: string,
+  typeAliases: Record<string, TypeAliasEntry>,
+): ShapeResult {
+  const probe = parseAgency(`const ${PROBE_VARIABLE}: ${constraintText} = ${valueText}`, {}, false);
```
egonSchiele: what if the type contains aliases?

packages/agency-lang/lib/optimize/constraint.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+  const result = typeCheck(probe.result, {}, makeProbeUnit(typeAliases));
+  if (result.errors.length === 0) return { ok: true };
+  return {
+    ok: false,
+    reason: result.errors[0].message.replace(` (assignment to '${PROBE_VARIABLE}')`, ""),
+  };
+}
+
+const LITERAL_EXPRESSION_TYPES = ["string", "multiLineString", "number", "boolean", "agencyObject", "agencyArray"];
```
egonSchiele: some of these consts/functions feel like they should be in a util file instead, so other places can use them too

packages/agency-lang/lib/optimize/targets.ts
```diff
@@ -13,8 +16,16 @@ export type OptimizeTarget = {
   absoluteFile: string;
   scope: string;
   name: string;
-  valueKind: "string" | "multilineString";
+  valueKind: "string" | "multilineString" | "literal";
+  /** Decoded text for text targets (interpolations rendered as `${...}` —
+   *  the `expected`-guard contract); the EXACT source slice, quotes intact,
+   *  for literal targets. */
   value: string;
+  /** Parseable type text proposals are probed against (constraint.ts).
```
egonSchiele, line 25: if this is a type def, why not name is something like `typeDefinition`? `constraintText` is not descriptive

packages/agency-lang/lib/optimize/sourceMutator.ts
```diff
@@ -353,29 +355,70 @@ export class OptimizeSourceMutator {
...
       const wrapped = exprParser(`${quote}${operation.value}${quote}`);
-      if (!wrapped.success || wrapped.rest.trim() !== "") {
-        return diagnostic({ operation, code: "invalid-replacement-syntax", message: `Replacement value for ${operation.target} must be a quoted Agency string literal (e.g. "new prompt"), but it did not parse as one even after wrapping it in quotes. Received: ${JSON.stringify(operation.value)}` });
+      if (wrapped.success && wrapped.rest.trim() === "") parsed = wrapped;
+    }
+    if (!parsed.success || parsed.rest.trim() !== "") {
+      return diagnostic({ operation, code: "invalid-replacement-syntax", message: `Replacement value for ${operation.target} did not parse as an Agency expression. Received: ${JSON.stringify(operation.value)}` });
+    }
+    let replacement = parsed.result;
+    let proposalText = operation.value.trim();
```
egonSchiele: 🔴 **Critical: `checkProposal` probes a different value than what gets written — false-accepts out-of-shape values.**

After the top-level quote-wrap recovery (just above) swaps `parsed` to the wrapped string literal, `proposalText` is still set here to the raw *unquoted* `operation.value.trim()`. So the check at line 411 probes the unquoted source, while the value actually written comes from `replacement` (the recovered string). Checked ≠ written.

**Repro (verified end-to-end on this branch):** `type Status = "pass" | "fail"; optimize const status: Status = "pass"`; model proposes the unquoted multi-word value `totally bogus`. `exprParser("totally bogus")` leaves rest `" bogus"`, so wrap-recovery makes `replacement` the string `"totally bogus"` — but `proposalText` stays `totally bogus`. `checkProposal("Status", "totally bogus")` builds the probe `const __optimizeProbe: Status = totally bogus`, which parses as `= totally` (undefined ident → `any`) **plus a stray `bogus` statement**, typechecks clean, and returns `ok: true`. The mutator writes `optimize const status: Status = "totally bogus"` with **zero diagnostics** — an invalid union member accepted. (Mirror case: a valid multi-word member that parses as a keyword statement is falsely *rejected*.)

**Fix:** update `proposalText` to the recovered/quoted source whenever wrap-recovery fires — e.g. probe `checkProposal` with `renderLiteralSource(replacement)` so the checked value is always exactly the written value.

packages/agency-lang/lib/optimize/sourceMutator.ts
```diff
@@ -353,29 +355,70 @@ export class OptimizeSourceMutator {
...
-    const newValue = promptSegmentsToString(replacement.segments);
-    const validation = validateOptimizedStringValue(target.value, newValue);
-    if (!validation.ok) {
-      return diagnostic({ operation, code: "interpolation-mismatch", message: `Replacement value for ${operation.target} is invalid: ${validation.reason}` });
+
+    // Literal gate. The probe cannot enforce this — an unknown bare
+    // identifier synthesizes to `any` and typechecks clean — so a bare-word
+    // proposal (`fail` instead of `"fail"`) is recovered by quoting, and
+    // anything else non-literal is rejected outright.
+    if (!isLiteralExpression(replacement)) {
```
egonSchiele, line 411: 🟠 **High: the literal gate is top-level-only, so nested non-literals bypass validation.**

`isLiteralExpression(replacement)` inspects only `replacement.type` (`agencyArray`/`agencyObject`), not the contents. A literal container holding bare identifiers passes the gate, `hasInterpolation` doesn't flag them, and `checkProposal` accepts because unknown identifiers synthesize to `any` (and `synthObject` degrades the whole object to `any`).

**Repro (verified):**
- `optimize const nums: number[] = [1, 2]`, propose `[x, y]` → accepted, writes `= [x, y]`.
- `type P = { n: number }; optimize const p: P = { n: 1 }`, propose `{ n: someVar }` → accepted, writes `= { n: someVar }`.

Both write candidates that reference undefined variables and fail downstream at compile/eval, instead of a clean `type-mismatch` with useful feedback. This is broader than the documented explicit-`null` fail-open (framed as "imperfect but valid record") — it admits **broken source**.

**Fix:** recurse `isLiteralExpression` into array items / object entry values, or have `checkProposal` reject the `any`-degradation instead of accepting it.

packages/agency-lang/lib/optimize/targets.ts
```diff
@@ -203,10 +260,31 @@ function buildTarget(
...
+  const value = assignment.value;
+  if (value.type === "messageThread" || !isLiteralExpression(value)) {
     throw new Error(
-      `Unsupported optimize target ${file}:${scope}:${assignment.variableName}. Only string and multiline string initializers are supported today.`,
+      `Unsupported optimize target ${file}:${scope}:${assignment.variableName}. Its value must be a literal (string, number, boolean, null, object, or array).`,
+    );
+  }
+
+  const isText = value.type === "string" || value.type === "multiLineString";
+  const valueKind: OptimizeTarget["valueKind"] = value.type === "multiLineString"
```
egonSchiele: 🟡 **Medium: `valueKind` is re-derived from the champion value on every refresh, so an unconstrained literal target can flip to freeform mid-run.**

`valueKind` comes from the *current initializer's* expression type, not the declaration. For an unconstrained literal target (a non-string literal whose annotation failed the baseline self-test → `constraintText: null`, `valueKind: "literal"`), a string-valued champion flips `valueKind` to `"string"` on the next `collectTargets` refresh, and with `constraintText` still `null` the target silently becomes **freeform**.

**Scenario:** `optimize const x: NotDefined = 5` (baseline self-test fails → unconstrained literal). Iter 1 proposes `"hi"` → accepted. Refresh re-derives `valueKind: "string"` → now freeform. Iter 2 proposes `7` → rejected ("must be a string or multiline string"), though the same declaration originally accepted `5`. Same declaration, contradictory rules depending on the champion's value type.

**Fix:** decide the literal-vs-text identity from the declaration once at discovery and carry it, rather than recomputing from each champion value.

packages/agency-lang/lib/optimize/optimizers/greedyReflective.test.ts
```diff
@@ -203,3 +205,70 @@ describe("GreedyReflective (pointwise)", () => {
...
+    expect(result.validationFailedCount).toBe(1);
+    expect(result.championIter).toBe("baseline");
+
+    // The rejection reason flows back into the retry prompt inputs.
+    expect(proposeCalls[1]?.diagnostics?.[0]?.code).toBe("type-mismatch");
+    expect(proposeCalls[1]?.diagnostics?.[0]?.message).toContain("exploded");
+
+    // No candidate agent file was ever materialized with the bad value.
+    const iterAgent = path.join(root, "typed-e2e", "iter-1", "agent");
+    expect(fs.existsSync(iterAgent)).toBe(false);
```
egonSchiele: 🟡 **Medium (test): this assertion is vacuous.**

`iter-N/agent` dirs are only ever written by the legacy `loop.ts` (`createOptimizeArtifacts.writeIterationAgent`); the pointwise/greedy path (`BaseOptimizer` / `GreedyReflective` / `reporter`) never references `artifacts` at all (verified: no matches). So `expect(fs.existsSync(iterAgent)).toBe(false)` is unconditionally true and would still pass even if a regression materialized a rejected candidate to disk. The real atomic-no-file guarantee is proven only by the sibling test's `expect(bad.files).toEqual({})`.

The other assertions here are solid (retry count = 3, `validationFailedCount === 1`, champion stays baseline, `type-mismatch` reason threaded into `proposeCalls[1].diagnostics`). Consider asserting against a path the pointwise path actually writes, or drop this leg.

packages/agency-lang/lib/optimize/sourceMutator.ts
```diff
@@ -353,29 +355,70 @@ export class OptimizeSourceMutator {
...
       // expression (e.g. embedded quotes, or newlines under single-quote style)
       // fall through to the diagnostic, so we never emit broken source.
       const quote = target.valueKind === "multilineString" ? `"""` : `"`;
       const wrapped = exprParser(`${quote}${operation.value}${quote}`);
-      if (!wrapped.success || wrapped.rest.trim() !== "") {
-        return diagnostic({ operation, code: "invalid-replacement-syntax", message: `Replacement value for ${operation.target} must be a quoted Agency string literal (e.g. "new prompt"), but it did not parse as one even after wrapping it in quotes. Received: ${JSON.stringify(operation.value)}` });
+      if (wrapped.success && wrapped.rest.trim() === "") parsed = wrapped;
+    }
+    if (!parsed.success || parsed.rest.trim() !== "") {
+      return diagnostic({ operation, code: "invalid-replacement-syntax", message: `Replacement value for ${operation.target} did not parse as an Agency expression. Received: ${JSON.stringify(operation.value)}` });
```
egonSchiele: 🔵 Minor: this unified message drops the old actionable hint. When a freeform string proposal fails even after quote-wrapping, the retry used to say "must be a quoted Agency string literal (e.g. \`"new prompt"\`)…", telling the LLM exactly what to send; it now just says "did not parse". Accept/reject is byte-identical, but the feedback the mutator sees is less actionable.

packages/agency-lang/lib/optimize/constraint.ts
```diff
@@ -0,0 +1,148 @@
...
+    return {
+      ok: false,
+      reason: `value does not parse as an Agency expression of type ${constraintText}`,
+    };
+  }
+  const result = typeCheck(probe.result, {}, makeProbeUnit(typeAliases));
+  if (result.errors.length === 0) return { ok: true };
+  return {
+    ok: false,
+    reason: result.errors[0].message.replace(` (assignment to '${PROBE_VARIABLE}')`, ""),
```
egonSchiele: 🔵 Minor: this `.replace()` only strips the suffix for one exact message format. If the first diagnostic doesn't end in ` (assignment to '__optimizeProbe')` — a parse error, an unknown-property error, or any future message-format change — the internal probe variable `__optimizeProbe` leaks into the retry feedback shown to the mutator LLM. Consider a probe name that reads acceptably if it leaks, or strip more robustly.

egonSchiele (commented):
## Review — typed optimize-const targets

Nice work — and the altitude is right: validating proposals by splicing them into a one-line probe program and running the real `typeCheck()`, rather than hand-rolling a shape checker, is the correct call. The end-to-end greedy test genuinely exercises the retry-feedback + champion-stays-baseline path, and the `null`-as-`variableName` handling and baseline self-test are well reasoned.

Two correctness issues undercut the core "reject out-of-shape values" guarantee — **both reproduced against this branch**, not inferred:

- 🔴 **Critical** — after quote-wrap recovery, `checkProposal` probes the *unquoted* source while a *quoted* string is what actually gets written, so an unquoted multi-word proposal for a string-union target is silently accepted (`sourceMutator.ts:374`/`411`).
- 🟠 **High** — the literal gate is top-level-only, so array/record proposals containing bare identifiers slip through and are written as candidates that reference undefined variables (`sourceMutator.ts:400`).

Plus a mid-run consistency bug (`valueKind` flip, `targets.ts:273`), a vacuous test assertion (`greedyReflective.test.ts:272`), and two minor nits — all inline with repros and suggested fixes.

Also worth noting (non-blocking):
- **Perf** — `renderFile` re-collects every target twice per `preview()`, and `buildTarget` runs the baseline self-test (`parseAgency` + `typeCheck`) for *every* typed target each time. So each preview re-typechecks all typed baselines, not just the changed one (N·M·2 redundant typechecks). Consider computing constraints once at discovery and not re-running the self-test on refresh.
- **Doc drift** — `OptimizeTarget.value` now stores an exact source slice for literal targets, but the `expected`-guard doc on `ReplaceVariableInitializerOperation` still says "current decoded value"; a latent `expected-mismatch` for API / custom-optimizer callers that honor that contract.

The design is sound; none of these need a rearchitecture.

packages/agency-lang/lib/optimize/sourceMutator.ts
egonSchiele: Fixed in 441c850b. The typed path now probes `generateExpression(replacement)` — the formatter rendering of the (possibly quote-recovered) value — so checked == written by construction, and the multi-statement probe hazard is gone entirely. Both directions pinned: `totally bogus` against `"pass" | "fail"` is rejected with `type-mismatch`, and the mirror case (unquoted `totally bogus` where it IS a member) recovers and accepts.

packages/agency-lang/lib/optimize/sourceMutator.ts
egonSchiele: Fixed in 441c850b. `isLiteralExpression` now recurses through array items, object entry values, and splat values, so `[x, y]` and `{ n: someVar }` are rejected as `unsupported-value-domain` before probing. I went with the recursive gate rather than rejecting `any`-degradation in `checkProposal`: rejecting the degradation would flip the documented explicit-`null` fail-open into fail-closed (a `{ ..., priority: null }` record would start bouncing), and the null case is a valid value whereas bare identifiers are broken source — different categories.

packages/agency-lang/lib/optimize/targets.ts
egonSchiele: Fixed in 441c850b. `collectTargets` takes the prior targets by id; refresh collects reuse the prior literal-vs-text identity and `declaredType` instead of re-deriving them from the champion value, so the unconstrained-literal → freeform flip is gone (pinned by a two-iteration test). Within text targets the string/multiline sub-flavor still tracks the current value — it selects the quote style recovery wraps with, and two pre-existing tests pin that flip as intended. This also resolves your perf note: the baseline self-test no longer re-runs on any refresh, only once at discovery.

packages/agency-lang/lib/optimize/optimizers/greedyReflective.test.ts
egonSchiele: Verified — `writeIterationAgent` is only called from the legacy `loop.ts`, so that path never fires here. Replaced in 441c850b with a recursive scan asserting no `.agency` file anywhere under the run root contains the rejected value, which covers whatever the pointwise path materializes (workspaces included) rather than a specific directory layout.

packages/agency-lang/lib/optimize/sourceMutator.ts
egonSchiele: Restored in 441c850b — the parse-failure message branches on target kind: text targets get the prescriptive "must be a quoted Agency string literal (e.g. \"new prompt\")…" wording back, literal targets get the expression-oriented one.

packages/agency-lang/lib/optimize/constraint.ts
egonSchiele: Fixed in 441c850b, both halves: the probe variable is renamed to `proposedValue` (reads as plain English if it ever leaks — "assignment to 'proposedValue'") and the strip is `replaceAll`. Parse failures now also append the parser's own message per the Copilot comment.

packages/agency-lang/lib/optimize/constraint.ts
egonSchiele: `generateAgency` is program-level (imports sorting, statement layout), so a bare expression would need wrapping in a synthetic program and unwrapping the output. Instead 441c850b exports `generateExpression` from `agencyGenerator.ts` — a thin public wrapper over the class's `processNode`, the same renderer interpolations use — and the optimizer consumes that. Same module owns it; no class construction in optimize code.

packages/agency-lang/lib/optimize/constraint.ts
egonSchiele: Aliases inside the annotation work without any eager expansion: `formatTypeHint` renders alias references by name (e.g. `{ status: Status }` stays `{ status: Status }`), and the probe resolves names lazily against the closure registry at comparison time (`safeResolveType` inside `isAssignable`), one hop at a time with a cycle guard. 441c850b adds tests pinning `{ status: Status }` and `Status | null` annotations end-to-end. Aliases the registry cannot see (e.g. from an unresolvable package) fall to the baseline self-test and the target degrades to unconstrained rather than over-rejecting.

packages/agency-lang/lib/optimize/constraint.ts
egonSchiele: Moved in 441c850b: `isNullLiteral`, `isLiteralExpression`, and `hasInterpolation` are pure AST predicates and now live in `lib/utils/node.ts`; `generateExpression` is exported from `agencyGenerator.ts`. `constraint.ts` keeps only the probe machinery (`makeProbeUnit`, `checkProposal`, `renderDeclaredType`, `describeConstraint`), which is optimizer-specific.

packages/agency-lang/lib/optimize/targets.ts
egonSchiele: Renamed to `declaredType` in 441c850b. I avoided `typeDefinition` because the field holds the *annotation* rendered to source text (`Status`, `number | null`), not a definition body (`type X = ...`) — "declared type" is what it literally is. Happy to switch if you prefer the other name.

packages/agency-lang/lib/optimize/targets.ts
```diff
@@ -107,9 +127,42 @@ function buildTargetSet(
...
+    const symbolTable = SymbolTable.build(absoluteEntryFile);
+    const unit = buildCompilationUnit(entry.program, symbolTable, entry.absoluteFile, entry.source);
+    return unit.typeAliases.visibleIn(GLOBAL_SCOPE_KEY);
+  } catch (error) {
+    console.warn(
+      `optimize discovery: could not build the type-alias registry for ${absoluteEntryFile} (typed targets will be unconstrained): ${error instanceof Error ? error.message : String(error)}`,
+    );
+    const unit = buildCompilationUnit(entry.program);
+    return unit.typeAliases.visibleIn(GLOBAL_SCOPE_KEY);
+  }
```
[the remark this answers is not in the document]
egonSchiele: Adopted in 441c850b — the registry is copied onto a null-prototype object at build time (`nullPrototyped()`, same pattern as `registry.ts` / `evalCache.ts`), so user type names like `__proto__` land as plain own properties.

packages/agency-lang/lib/optimize/constraint.ts
```diff
@@ -0,0 +1,148 @@
...
+  valueText: string,
+  typeAliases: Record<string, TypeAliasEntry>,
+): ShapeResult {
+  const probe = parseAgency(`const ${PROBE_VARIABLE}: ${constraintText} = ${valueText}`, {}, false);
+  if (!probe.success) {
+    return {
+      ok: false,
+      reason: `value does not parse as an Agency expression of type ${constraintText}`,
+    };
+  }
```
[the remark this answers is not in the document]
egonSchiele: Included in 441c850b — parse-failure reasons now append `probe.message`, so both mutator retries and baseline self-test fallbacks show the parser's actual complaint (which may indeed be the rendered type rather than the value).

egonSchiele:
Review addressed in 441c850b — all findings reproduced first, then fixed test-first:

- 🔴 **Checked ≠ written**: typed proposals are now probed as `generateExpression(replacement)` — the exact text that gets written — killing both the false-accept (`totally bogus`) and the multi-statement probe hazard. Mirror case (valid multi-word member, unquoted) recovers correctly.
- 🟠 **Nested literal gate**: `isLiteralExpression` recurses through containers; `[x, y]` / `{ n: someVar }` are rejected up front instead of landing as broken source. (Kept as a gate rather than rejecting `any`-degradation in the probe, which would have flipped the explicit-`null` fail-open to fail-closed.)
- 🟡 **`valueKind` flip**: literal-vs-text identity and `declaredType` are decided once at discovery and carried through refreshes via prior targets — which also deletes the redundant per-preview baseline self-tests from your perf note. The string↔multiline sub-flavor still tracks the champion (it selects quote-recovery style; two pre-existing tests pin that).
- 🟡 **Vacuous assertion**: replaced with a recursive scan proving no `.agency` file under the run root ever contains the rejected value.
- Naming/structure: `constraintText` → `declaredType`; AST predicates moved to `lib/utils/node.ts`; `generateExpression` exported from `agencyGenerator.ts`; probe variable renamed `proposedValue` + robust strip; parser messages surfaced; prescriptive freeform hint restored; null-prototype registry; `expected`-guard docs updated; alias-in-annotation tests added.

314 optimizer tests (10 new since review), full repo suite 7275 green, tsc/lint clean.

PR #446, opened 2026-07-07
feat(littlesis): add std::data/people/littlesis connector

egonSchiele:
## Summary

Adds `std::data/people/littlesis` — a keyless entity/power-network stdlib connector over the [LittleSis](https://littlesis.org/api/) REST API. Four thin primitives:

- `littlesisSearch(name, page?)` — resolve a name to matching entities
- `littlesisEntity(id)` — full profile by id
- `littlesisRelationships(id, category?, sort?)` — typed edges (entity IDs, amounts, dates, roles)
- `littlesisConnections(id, category?)` — neighbor entities **by name**, with the linking relationship

Follows the shipped `std::data/finance/*` recipe (pure `build*Path` URL builders, `parse*` reshapers, `*Finalize` Result handlers).

## Design decisions

- **Interrupt gating "Option C":** each function raises one `std::littlesis` prompt, then calls `std::http`'s `fetchJSON` inside an internal handler that approves the resulting `std::http::fetchJSON`. A plain caller sees a **single** prompt; an outer egress/audit handler **still receives** the fetch and can `reject()`/`propagate()` it (those beat the internal approve). `allowedDomains: ["littlesis.org"]` pins the host regardless. Net: single-prompt ergonomics without an egress-governance blind spot.
- **Tagged-union effect payload** `{ op, query }` so a handler or `std::policy` rule can gate individual operations (`match: { op: "relationships" }`) without splitting into four effects.
- **Fail-fast category validation** before the interrupt (better UX — don't prompt for an operation that will immediately fail on bad input; also makes the bad-category path offline-testable).
- Registers `std::littlesis` in the `Network` capability set.

## Testing

Offline agency-js tests (no LLM, no network): category mapping, URL builders, JSON:API parsers, finalize error paths, and interrupt behavior — gating (rejecting short-circuits before any fetch), single-prompt, governance (an outer `propagate()` surfaces the fetch and the test asserts the exact built URL offline), and offline bad-category. Plus an opt-in live test (`AGENCY_LIVE_TESTS`) run once end-to-end against the real API (search returned and parsed ≥1 entity).

Generated stdlib reference page + VitePress nav entry.

## Notes / follow-ups

- The `sample-*.json` fixtures backing the parser tests are hand-authored. The live run validated the **search** path against real data; the connections/relationships parser shapes are currently only checked against the samples (could be hardened by capturing real bodies and/or exercising the two-step recipe in the live test).

### DIFF
--- a/packages/agency-lang/stdlib/data/people/littlesis.agency
+++ b/packages/agency-lang/stdlib/data/people/littlesis.agency
@@ -38,11 +38,11 @@ import { fetchJSON } from "std::http"
 
 effect std::littlesis { op: string, query: string }
 
-const LITTLESIS_BASE = "https://littlesis.org/api"
-const LITTLESIS_DOMAINS = ["littlesis.org"]
+static const LITTLESIS_BASE = "https://littlesis.org/api"
+static const LITTLESIS_DOMAINS = ["littlesis.org"]
 
 // Index 0 is unused ("" sentinel); indices 1–12 are LittleSis category_id values.
-const CATEGORY_NAMES = ["", "position", "education", "membership", "family", "donation", "transaction", "lobbying", "social", "professional", "ownership", "hierarchy", "generic"]
+static const CATEGORY_NAMES = ["", "position", "education", "membership", "family", "donation", "transaction", "lobbying", "social", "professional", "ownership", "hierarchy", "generic"]
 
 /** A LittleSis person or organization. `type` is "Person" or "Org" (from primary_ext).
     `relationshipId`/`relationshipCategory` are populated only on connections results. */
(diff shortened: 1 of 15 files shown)


packages/agency-lang/stdlib/data/people/littlesis.agency
```diff
@@ -0,0 +1,328 @@
...
+      const linked = littlesisConnections(hits[0].id, "ownership") catch []
+      for (e in linked) { print("${e.type}  ${e.name}") }
+    }
+  }
+  ```
+*/
+
+effect std::littlesis { op: string, query: string }
+
+const LITTLESIS_BASE = "https://littlesis.org/api"
```
egonSchiele: mark consts static when possible

packages/agency-lang/stdlib/data/people/littlesis.agency
```diff
@@ -0,0 +1,328 @@
...
+const LITTLESIS_DOMAINS = ["littlesis.org"]
+
+// Index 0 is unused ("" sentinel); indices 1–12 are LittleSis category_id values.
```
egonSchiele, line 44: would it make sense to use `null` if category name is not given?

packages/agency-lang/stdlib/data/people/littlesis.agency
```diff
@@ -0,0 +1,328 @@
...
+export safe def parseEntities(raw: any): Entity[] {
+  const r: any = raw ?? {}
+  const data = r.data ?? []
+  return map(data) as node {
+    return parseEntity(node)
+  }
+}
+
+/** Reshape a LittleSis relationships body ({ data: [rel] }) into Relationship[]. Pure/total. */
+export safe def parseRelationships(raw: any): Relationship[] {
```
egonSchiele, line 179: many of these don't have doc strings. Why is that?

packages/agency-lang/stdlib/data/people/littlesis.agency
```diff
@@ -0,0 +1,328 @@
...
+    return success(0)
+  }
+  const id = categoryNameToId(category)
+  if (id < 0) {
+    return failure("unknown category '${category}'; valid: ${categoryNamesList()}")
+  }
+  return success(id)
+}
+
+/** Fetch a LittleSis path and return the parsed-JSON Result. Option C: this raises
```
egonSchiele: `Option C` is an implementation detail. Remove.

packages/agency-lang/stdlib/data/people/littlesis.agency
```diff
@@ -0,0 +1,328 @@
...
+        return failure("LittleSis returned no entity")
+      }
+      return success(parseEntity(bodyObj.data))
+    }
+    failure(err) => failure(littlesisError(err))
+  }
+}
+
+/** Turn a fetchJSON Result into a relationships Result. Pure. */
+export safe def relationshipsFinalize(fetchResult: any): Result {
```
egonSchiele: You can give a more specific type than `Result`:
https://agency-lang.com/guide/error-handling.html#result-type-parameters

PR #447, opened 2026-07-07
Add std::tag module with value tags and statelog redaction

egonSchiele:
## Summary

Adds a `std::tag` module for attaching arbitrary tags to values, plus a built-in `redact` tag that keeps secrets (API keys, tokens) out of state logs. Motivating case: an API key passed into a `fetch` should never appear in statelog.

```ts
import { redact } from "std::tag"

def callApi(apiKey: string) {
  redact(apiKey)   // apiKey shows as "[REDACTED]" in state logs
  return fetch("https://api.example.com", { headers: { key: apiKey } })
}
```

## Design

- **Two tag stores on `GlobalStore`** (no boxing of values, so TS interop is untouched):
  - **Primitives** keyed by **value** (`Map` under `__internal`) — durable: clones into fork branches and survives interrupt/resume via the existing `MapReviver`.
  - **Objects/arrays** keyed by **reference** (`WeakMap`) — branch-local by design (object identity does not survive the `toJSON`/`fromJSON` round-trip), reset on `clone()`.
- **`redact` semantics owned by `GlobalStore`** (`markRedacted`/`isRedacted`/`REDACT_TAG`) so the tag shape lives in exactly one place.
- **Redaction at the single `StatelogClient.post()` chokepoint** as a `JSON.stringify` replacer, so it rides the one serialization pass, preserves `Date`/`URL`/`toJSON` values, and covers all ~40 event types at one seam. Gated on `hasAnyTags()` so tag-free programs pay nothing.
- **Scoped to the `data` payload only** (envelope and payload are stringified separately and spliced), so a value-collision can never corrupt infra fields like `format_version`.

## Limits (v1, documented)

- Whole-value redaction only — a secret embedded inside a larger logged string is not scrubbed.
- Statelog only — does not affect `print()`; not a general secrecy/information-flow guarantee.
- Primitives are keyed by value, so tagging a low-entropy value (e.g. `1`) redacts every occurrence — intended for high-entropy secrets.

Provenance tracking (`response.provenance`) is explicitly out of scope; the spec sketches it as future work on the same mechanism.

## Testing

- Unit: `GlobalStore` tag storage (value/reference keying, `1` vs `"1"` distinctness, clone/serialize durability, `hasAnyTags`); the redaction replacer (native-type preservation, object-node redaction, substring boundary); `std::tag` TS helpers.
- Integration: redaction end-to-end through `post()` via the stdout sink, including an untagged-`Date` guard, envelope-immunity, and a fork-style globals-clone durability test.
- Execution (`tests/agency/tag.agency`): value keying, `redact` writing the tag, object-by-reference, and a `fork` branch inheriting a primitive tag.
- Full `make` build (tsc) passes.

Design spec and plan under `docs/superpowers/`.

### DIFF
--- a/packages/agency-lang/docs/site/.vitepress/config.mts
+++ b/packages/agency-lang/docs/site/.vitepress/config.mts
@@ -128,6 +128,7 @@ export default defineConfig({
             { text: "Developer Tools", link: "/guide/developer-tools" },
             { text: "Debugging", link: "/guide/debugging" },
             { text: "Observability", link: "/guide/observability" },
+            { text: "Tags and Redaction", link: "/guide/tags" },
           ],
         },
 
--- a/packages/agency-lang/docs/site/.vitepress/config.mts
+++ b/packages/agency-lang/docs/site/.vitepress/config.mts
@@ -128,7 +128,6 @@ export default defineConfig({
             { text: "Developer Tools", link: "/guide/developer-tools" },
             { text: "Debugging", link: "/guide/debugging" },
             { text: "Observability", link: "/guide/observability" },
-            { text: "Tags and Redaction", link: "/guide/tags" },
           ],
         },
 
@@ -158,6 +157,7 @@ export default defineConfig({
               link: "/guide/value-parameterized-types",
             },
             { text: "Memory", link: "/guide/memory" },
+            { text: "Tags and Redaction", link: "/guide/tags" },
           ],
         },
         {
--- a/packages/agency-lang/docs/site/stdlib/tag.md
+++ b/packages/agency-lang/docs/site/stdlib/tag.md
@@ -35,11 +35,11 @@ Note: object/array tags are branch-local — they do not survive `fork`/`race`/
 ### tag
 
 ```ts
-tag(value: any, key: string, val: any)
+tag(value: any, key: string, val: any): any
 ```
 
 Attach a tag to a value. Primitives are keyed by value; objects by
-  reference. `val` defaults to `true`.
+  reference. `val` defaults to `true`. Returns the value's current tags.
 
 **Parameters:**
 
@@ -49,7 +49,29 @@ Attach a tag to a value. Primitives are keyed by value; objects by
 | key | `string` |  |
 | val | `any` | true |
 
+**Returns:** `any`
+
 ([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/tag.agency#L30))
+
+### setTags
+
+```ts
+setTags(value: any, tags: any): any
+```
+
+Attach multiple tags at once from a `{ key: value }` object. Returns the
+  value's current tags.
+
+**Parameters:**
+
+| Name | Type | Default |
+|---|---|---|
+| value | `any` |  |
+| tags | `any` |  |
+
+**Returns:** `any`
+
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/tag.agency#L36))
 
 ### getTags
 
@@ -57,7 +79,8 @@ Attach a tag to a value. Primitives are keyed by value; objects by
 getTags(value: any): any
 ```
 
-Return all tags on a value as an object, or an empty object if none.
+Return all tags on a value as an object, or an empty object if none. The
+  result is a shallow copy; nested tag values are shared references.
 
 **Parameters:**
 
@@ -67,16 +90,16 @@ Return all tags on a value as an object, or an empty object if none.
 
 **Returns:** `any`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/tag.agency#L36))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/tag.agency#L42))
 
 ### redact
 
 ```ts
-redact(value: any)
+redact(value: any): any
 ```
 
 Mark a value so it is replaced with "[REDACTED]" in state logs. Shorthand
-  for tag(value, "redact", true).
+  for tag(value, "redact", true). Returns the value's current tags.
 
 **Parameters:**
 
@@ -84,4 +107,43 @@ Mark a value so it is replaced with "[REDACTED]" in state logs. Shorthand
 |---|---|---|
 | value | `any` |  |
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/tag.agency#L41))
+**Returns:** `any`
+
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/tag.agency#L48))
+
+### removeTag
+
+```ts
+removeTag(value: any, key: string): any
+```
+
+Remove a single tag from a value. Returns the value's remaining tags.
+
+**Parameters:**
+
+| Name | Type | Default |
+|---|---|---|
+| value | `any` |  |
+| key | `string` |  |
+
+**Returns:** `any`
+
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/tag.agency#L54))
+
+### removeAllTags
+
+```ts
+removeAllTags(value: any): any
+```
+
+Remove all tags from a value. Returns an empty object.
+
+**Parameters:**
+
+| Name | Type | Default |
+|---|---|---|
+| value | `any` |  |
+
+**Returns:** `any`
+
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/tag.agency#L59))
--- a/packages/agency-lang/docs/site/stdlib/tag.md
+++ b/packages/agency-lang/docs/site/stdlib/tag.md
@@ -1,0 +1,87 @@
+---
+name: "tag"
+---
+
+# tag
+
+Attach arbitrary tags to values and read them back anywhere in
+your program.
+
+Tags are stored in a side table, so nothing is attached to the value itself
+and TypeScript interop is unaffected. Two semantics, by value kind:
+
+- **Primitives** (string, number, boolean) are keyed by **value**: tagging one
+  copy of `"secret"` tags every equal `"secret"` in the current branch.
+- **Objects and arrays** are keyed by **reference**: tagging one object does
+  not tag a structurally-equal but distinct object.
+
+The built-in `redact` tag marks a value so it is replaced with `"[REDACTED]"`
+in state logs — useful for API keys and other secrets.
+
+  ```ts
+  import { redact } from "std::tag"
+
+  def callApi(apiKey: string) {
+    redact(apiKey)          // apiKey shows as "[REDACTED]" in state logs
+    return fetch("https://api.example.com", { headers: { key: apiKey } })
+  }
+  ```
+
+Note: object/array tags are branch-local — they do not survive `fork`/`race`/
+`parallel` branches or interrupt/resume. Primitive (value) tags survive both.
+
+## Functions
+
+### tag
+
+```ts
+tag(value: any, key: string, val: any)
+```
+
+Attach a tag to a value. Primitives are keyed by value; objects by
+  reference. `val` defaults to `true`.
+
+**Parameters:**
+
+| Name | Type | Default |
+|---|---|---|
+| value | `any` |  |
+| key | `string` |  |
+| val | `any` | true |
+
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/tag.agency#L30))
+
+### getTags
+
+```ts
+getTags(value: any): any
+```
+
+Return all tags on a value as an object, or an empty object if none.
+
+**Parameters:**
+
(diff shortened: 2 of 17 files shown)


packages/agency-lang/lib/statelogClient.ts
```diff
@@ -1163,14 +1165,37 @@ export class StatelogClient {
...
-      data: { ...body, timestamp: new Date().toISOString() },
     });
+    const dataJson = JSON.stringify(
+      { ...body, timestamp: new Date().toISOString() },
+      replacer,
+    );
+    // Splice: drop the envelope's closing brace and append the data field. The
+    // envelope always carries format_version, so envelopeJson is never "{}" and
+    // the leading comma is always valid.
+    const postBody = `${envelopeJson.slice(0, -1)},"data":${dataJson}}`;
```
egonSchiele: The scoping refinement here is good — splicing so the replacer only touches `data` means a pathological `redact(1)` can't blank out `format_version: 1`, and the output is byte-identical to the previous single-`stringify` when nothing is redacted.

One robustness note: this is string surgery that relies on `envelopeJson` always being a non-empty object ending in `}` (true today because `format_version` is always present). The comment calls it out, so it's fine — but since a wrong assumption produces *invalid JSON* silently rather than a loud failure, a one-line guard (e.g. assert `envelopeJson.endsWith("}")` / length > 2) would make the invariant enforced rather than documented.

packages/agency-lang/lib/runtime/redactForStatelog.ts
```diff
@@ -0,0 +1,42 @@
...
+    // Recover the raw, pre-toJSON value for the tag lookup.
+    let raw: unknown;
+    if (typeof value === "object" && value !== null) {
+      raw = value;
+    } else if (key === "") {
+      raw = value;
+    } else {
+      raw = (this as Record<string, unknown>)[key];
+    }
+    if (globals.isRedacted(raw)) return REDACTED;
```
egonSchiele, line 37: Worth surfacing a limitation inherent to the replacer approach: it redacts **values only, never object keys**. A `JSON.stringify` replacer gets `(key, value)` and can rewrite the value but not the key — so a secret used as a key (e.g. a map keyed by API key, `{ "sk-...": {...} }`) still appears in the log verbatim. Almost certainly acceptable for v1 (secrets-as-keys is unusual and the design targets values), but it belongs in the 'Limits' list in `docs/site/guide/tags.md` next to the whole-value/substring caveat so nobody assumes key coverage.

packages/agency-lang/lib/stdlib/tag.ts
```diff
@@ -0,0 +1,23 @@
...
+export function _tag(value: unknown, key: string, val: unknown): void {
+  const g = __globals();
+  if (!g) return;
+  g.setTag(value, key, val);
+}
+
+/** Return a shallow copy of a value's tags, or {} if none. */
+export function _getTags(value: unknown): Record<string, unknown> {
+  const t = __globals()?.getTagsFor(value);
+  return t ? { ...t } : {};
```
egonSchiele: Minor: this copy is shallow, so if a tag value is itself an object (`tag(x, "meta", { ... })`), the returned `.meta` is the same reference stored in the tag record — a caller mutating it mutates the store, despite the "returns a copy" contract. Tag values are usually primitives so impact is low; just flagging the docstring promises more isolation than the code gives. (The `_getTags returns a copy` test only exercises a primitive value, so it wouldn't catch this.)

egonSchiele (commented):
Strong PR — the implementation tracks the plan closely and the test coverage is genuinely load-bearing (each unit test is discriminating, not a happy-path rubber stamp). A few things I verified and liked:

- **Data-scoped splice is a real improvement over the plan.** Running the redaction replacer over only the `data` payload (stringify envelope + payload separately, splice) makes infra fields structurally immune to a value-collision, and it's byte-identical to the old single-`stringify` output when nothing is redacted — I confirmed the existing `statelogClient.test.ts` assertions (`payload.data.type`, `payload.trace_id`, `format_version: 1`) still hold, so no wire-format regression.
- **The native-type regression is properly guarded.** The Date / custom-`toJSON` tests (both at the replacer level and end-to-end through `post()`) pin the exact bug the replacer-over-walker rework was meant to avoid.
- **Fork durability is covered by composition:** the `.agency` `forkInheritsPrimitiveTag` test proves a real fork branch's `__globals()` sees the inherited primitive tag, and the integration test proves `post()` redacts from whatever `__globals()` returns — together that covers the real-fork → redaction seam.
- Nice catch on the `==` → `===` / `undefined`-not-`null` gotcha in `objectByReference`; the comment there will save the next person the same debugging.

Three minor, non-blocking notes inline (values-vs-keys redaction limitation, the splice invariant, and a shallow-copy nuance in `_getTags`). None of these need to hold up the merge.

packages/agency-lang/docs/site/.vitepress/config.mts
```diff
@@ -128,6 +128,7 @@ export default defineConfig({
             { text: "Developer Tools", link: "/guide/developer-tools" },
             { text: "Debugging", link: "/guide/debugging" },
             { text: "Observability", link: "/guide/observability" },
+            { text: "Tags and Redaction", link: "/guide/tags" },
```
egonSchiele: put this in the advanced section

packages/agency-lang/docs/site/stdlib/tag.md
```diff
@@ -0,0 +1,87 @@
...
+
+Attach arbitrary tags to values and read them back anywhere in
+your program.
+
+Tags are stored in a side table, so nothing is attached to the value itself
+and TypeScript interop is unaffected. Two semantics, by value kind:
+
+- **Primitives** (string, number, boolean) are keyed by **value**: tagging one
+  copy of `"secret"` tags every equal `"secret"` in the current branch.
+- **Objects and arrays** are keyed by **reference**: tagging one object does
```
egonSchiele, line 15: Just a thought, what if for objects and arrays (things where we're tracking tags by reference), we *do* store the tags on the object instead of a side table? I believe there is a way to do this in JavaScript. where you can make certain keys not get enumerated by functions like `Object.keys`. We could put all the tags in one object, under a single key like `__tags`. This would solve the problem where tags on objects and arrays don't work correctly after an interrupt.

packages/agency-lang/docs/site/stdlib/tag.md
```diff
@@ -0,0 +1,87 @@
...
+
+Note: object/array tags are branch-local — they do not survive `fork`/`race`/
+`parallel` branches or interrupt/resume. Primitive (value) tags survive both.
+
+## Functions
+
+### tag
+
+```ts
+tag(value: any, key: string, val: any)
```
egonSchiele: add `setTags` too which takes a `tags` object. Also add `removeTag` and `removeAllTags`

packages/agency-lang/lib/runtime/state/globalStore.ts
```diff
@@ -9,6 +9,111 @@ export class GlobalStore {
...
+    if (this.isRef(value)) {
+      let t = this.objectTags.get(value);
+      if (!t && create) {
+        t = {};
+        this.objectTags.set(value, t);
+        this.objectTagsPresent = true;
+      }
+      return t;
+    }
+    const m = create
```
egonSchiele: No single character variable names.

packages/agency-lang/lib/stdlib/tag.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export function _tag(value: unknown, key: string, val: unknown): void {
+  const g = __globals();
+  if (!g) return;
+  g.setTag(value, key, val);
+}
+
+/** Return a shallow copy of a value's tags, or {} if none. */
+export function _getTags(value: unknown): Record<string, unknown> {
+  const t = __globals()?.getTagsFor(value);
```
egonSchiele: Why are we storing tags on globals instead of on the current branch? you need to be using the TypeScript helpers.
https://agency-lang.com/guide/ts-helpers.html

packages/agency-lang/lib/statelogClient.ts
```diff
@@ -1163,14 +1165,37 @@ export class StatelogClient {
...
+    // trace_id, span ids). This keeps redaction single-pass (no extra deep
+    // copy) and preserves Date/URL/toJSON values, while making envelope fields
+    // structurally immune to a value-collision — e.g. a pathological
+    // `redact(1)` can't blank out `format_version: 1`. The replacer reads the
+    // caller's branch tag store via __globals(), synchronously before any
+    // detached (noWait) send, so it sees the correct branch. hasAnyTags() skips
+    // the replacer entirely when nothing is tagged, so tag-free programs pay
+    // nothing. Events posted outside an ALS frame (__globals() undefined) are
+    // not redacted — a documented boundary, not a secrecy guarantee.
+    const globals = __globals();
```
egonSchiele, line 1187: How does this work with the agency's state isolation?

packages/agency-lang/lib/statelogClient.ts
egonSchiele: This is horrible. Can you not edit the object when it is an object in JavaScript so you don't have to do this weird string splicing stuff?

packages/agency-lang/stdlib/tag.agency
```diff
@@ -0,0 +1,45 @@
...
+  }
+  ```
+
+Note: object/array tags are branch-local — they do not survive `fork`/`race`/
+`parallel` branches or interrupt/resume. Primitive (value) tags survive both.
+*/
+
+import { _tag, _getTags, _redact } from "agency-lang/stdlib-lib/tag.js"
+
+export def tag(value: any, key: string, val: any = true) {
```
egonSchiele: If your functions don't return any value, have it return a string like "value tagged successfully". This is because if an LLM uses this function as a tool, it will look for the return value to give it confirmation that the function ran correctly.

egonSchiele: You could also return all the current tags on this value. I think that would be useful.

packages/agency-lang/tests/agency/tag.agency
```diff
@@ -0,0 +1,39 @@
+import { tag, getTags, redact } from "std::tag"
+
+node valueKeyed(): string {
```
egonSchiele, line 3: Please review the tests you've written. Will the tests actually test what they're meant to? If the code breaks, will the test fail? Are there any missing test cases?

egonSchiele:
Thanks for the review — pushed three commits (`0ddd684a`, `c2f7a9bb`, `5d01fc26`) addressing everything. Point-by-point:

**Correctness**
- **Replacer `toJSON` leak (Copilot):** real bug. `JSON.stringify` runs `toJSON` before the replacer, so a tagged object with a custom `toJSON` returning an object was checked post-transform. Now the tag lookup always uses `this[key]` (the original). Added a regression test.
- **Prototype pollution (Copilot):** tag records are now `Object.create(null)`, so a user/LLM-supplied `__proto__` key is stored as data. Test added.
- **bigint/symbol keys (Copilot):** `setTag` now ignores non-serializable primitives (they'd throw in `MapReviver`/JSON and break `clone()`/interrupt save). Test added.

**Storage & isolation**
- **"How does this work with state isolation?" / "why globals not the current branch?"** `__globals()` (and `getRuntimeContext().globals`) resolve to the *branch-local* GlobalStore clone — per `docs/dev/globalstore.md`, the frame's `globals` slot is replaced by a per-branch clone inside fork/parallel/race. So tags are already per-branch and primitive tags are interrupt-durable. I switched the stdlib helpers to the documented `getRuntimeContext().globals` idiom (matches `thread.ts`); `post()` keeps the lenient `__globals()` since it can fire outside a frame. Added a code comment citing the isolation doc.
- **Store object tags on the object (non-enumerable `__tags`):** deferred to a follow-up as agreed. Noted in the spec why it needs a custom replacer/reviver rather than a bare non-enumerable key (non-enumerable is dropped by `JSON.stringify`; enumerable leaks into object shape and logs).

**Statelog wiring**
- **"This is horrible" string splice:** gone. `post()` now redacts `data` via `JSON.parse(JSON.stringify(data, replacer))` and builds the envelope as an ordinary object — same data-scoped immunity, no string surgery. The 43 existing `statelogClient.test.ts` assertions still pass (no wire-format change on the no-tag path).

**API**
- Added `setTags`, `removeTag`, `removeAllTags`.
- `tag`/`redact`/etc. now return the value's current tags (LLM-tool confirmation).
- `_getTags` shallow-copy nuance documented in the guide + docstring.

**Docs & tests**
- Guide: added the "values, never keys" redaction limit and the new API; moved the nav entry to **09. Advanced**.
- Reviewed the `.agency` tests and added compiled coverage for the new API (`mutateTags`, `tagReturnsCurrentTags`). Note: redaction-through-statelog is asserted at the TS integration level (`statelogClient.redaction.test.ts`, incl. the fork-clone durability case), since the `.test.json` harness checks node return values, not sink output; the compiled→`post()` seam is covered by composition (the `.agency` fork test proves a branch's tags are what `getTags` reads, and the integration test proves `post()` redacts from that same store).

All feature tests green: 114 TS unit/integration + 6 compiled execution, plus a clean `make` build.

PR #448, opened 2026-07-07
Deterministic fetch mocking for the agency test framework

egonSchiele:
## What

Deterministic HTTP `fetch` mocking for the agency test framework — the fetch analogue of the existing `llmMocks`. Tests declare canned responses and every `fetch` in the agent subprocess is served from them, with no real network.

## Why

Tests that exercise stdlib modules which fetch (weather, wikipedia, web-search, oauth, browserUse, memory embeddings) couldn't run deterministically in CI — they were skipped, sandbox-gated, or allowed to flake. There was no way to write a hermetic test for an agent whose behavior depends on an HTTP response. This turns the network off in tests the same way `AGENCY_USE_TEST_LLM_PROVIDER` turns the LLM off.

## How it works

Every consumer resolves the same `globalThis.fetch` at call time, so a **single swap** at the top of the agent subprocess intercepts them all — agency `fetch`/`fetchJSON`/`fetchMarkdown`, internal stdlib TS, and interop TS alike. The shim returns a real `Response`, so consumers that read `response.body.getReader()` (std::http) or `.json()` all work off one shim.

### Authoring

`fetchMocks` in a `.test.json` (file-level and/or per test), or a `fetchMocks.json` in an agency-js test dir — parallel to `llmMocks`:

```json
"fetchMocks": [
  { "url": "https://api.weather.com/v1/*", "return": { "tempF": 72 } },
  { "urlPattern": "wikipedia\\.org", "returnFile": "responses/dog.html",
    "headers": { "content-type": "text/html" } },
  { "url": "https://api/x", "method": "POST", "body": { "q": "cats" }, "return": "..." }
]
```

- **Match:** one of `url` (exact / `*` glob) or `urlPattern` (regex); optional `method`; optional request-body match via `body` (exact string / JSON-subset) or `bodyPattern` (regex). All ANDed; first match wins, serves unlimited calls.
- **Response:** one of `return` (inline) or `returnFile` (relative path, for large HTML/JSON); optional `status`/`headers`.
- **Unmatched → throws**, so a forgotten mock fails loudly instead of dialing out.

## Key design decisions

- **No activation gate, no bypass toggle.** Unlike the LLM system's two env vars (a gate + a data channel), fetch mocks are *always applied when present*. To hit the real API, just don't declare mocks; there's nothing to toggle.
- **File transport, not inline env.** `returnFile` targets large bodies, which would blow the `exec` `ARG_MAX` (~1 MB) ceiling if inlined into an env value. The runner writes resolved mocks to a temp file and passes its path via `AGENCY_FETCH_MOCKS_FILE`; the shim stays a pure in-memory matcher.
- **Below the language, not through `__call`.** The shim swaps the network primitive, so it does not interact with the interrupt/handler system — a rejected pre-fetch interrupt still stops the call (the mock never runs), and no mock can punch through a `with reject`. It is a test-only network switch, not a capability guard.

## Testing

- **25 unit tests** (`fetchMock.test.ts`, `fetchMockResolve.test.ts`): URL/method/body matching, first-match-wins + reuse, real-`Response` `getReader` fidelity, abort, config validation, `returnFile` inlining + merge order.
- **Agency e2e** (`tests/agency/fetchMock.test.json`): std::http `getReader` + content-type → markdown, inline mock, `returnFile`, and an unmatched fetch propagating as a failure — all through a real compiled agent.
- **Agency-js e2e** (`tests/agency-js/fetch-mock/`): inline + `returnFile` with no LLM flag set, proving independence from LLM deterministic mode.

Note: the stdlib `fetch`/`fetchJSON`/`fetchMarkdown` raise an approval interrupt before fetching, so a test node approves it (`fetch(url, path) with approve`) for the mock to be reached — see the e2e fixtures and the new TESTING.md section.

Record mode (`agency test fixtures --record-fetch`) to capture real responses into `returnFile` mocks is designed as Increment 2 and will land separately.


packages/agency-lang/lib/cli/test.ts
```diff
@@ -1040,6 +1071,10 @@ async function runTsTestDir(
       exitIfSignal(e);
       log(color.red(`  ✗ Test script execution failed: ${e}`));
       return { success: false, dir };
+    } finally {
+      if (fetchMocksTmpDir) {
+        fs.rmSync(fetchMocksTmpDir, { recursive: true, force: true });
```
egonSchiele: We have a safe delete function somewhere that checks to make sure you're not accidentally deleting the root directory or any directory outside of the current repo. Please use that function to delete files.

packages/agency-lang/lib/runtime/fetchMock.ts
```diff
@@ -0,0 +1,205 @@
...
+  }
+  if (Array.isArray(actual)) {
+    return false;
+  }
+  return Object.entries(subset as Record<string, unknown>).every(([k, v]) =>
+    deepSubset(v, (actual as Record<string, unknown>)[k]),
+  );
+}
+
+function buildBodyMatcher(m: FetchMock): ((raw: string) => boolean) | undefined {
```
egonSchiele: No single character variable names.

packages/agency-lang/tests/agency-js/fetch-mock/agent.agency
```diff
@@ -0,0 +1,17 @@
+import { fetch, fetchJSON } from "std::http"
```
egonSchiele, line 1: You should be able to write this as an agency test instead of an agency JS test, correct? Agency tests are much easier to write.

packages/agency-lang/lib/cli/test.ts
egonSchiele: Switched to a new writeFetchMocksTempFile helper that writes the mocks under the in-project .agency-tmp dir and deletes via safeDeleteDirectory (same pattern as typecheck.ts). Raw rmSync on an os.tmpdir dir could not use safeDeleteDirectory because it refuses paths outside the project root; moving the temp file in-project resolves that.

packages/agency-lang/lib/runtime/fetchMock.ts
egonSchiele: Renamed all single-character variables in fetchMock.ts and fetchMockResolve.ts.

packages/agency-lang/lib/cli/util.ts
```diff
@@ -358,14 +365,30 @@ export async function runAgencyNode({
...
     ? { AGENCY_LLM_MOCKS: JSON.stringify(llmMocks ?? []) }
     : {};
-  return runAgencyNode({ ...rest, env });
+
+  let fetchMocksDir: string | undefined;
+  if (fetchMocks && fetchMocks.length > 0) {
+    fetchMocksDir = fs.mkdtempSync(path.join(os.tmpdir(), "agency-fetchmocks-"));
+    const mocksFile = path.join(fetchMocksDir, "mocks.json");
+    fs.writeFileSync(mocksFile, JSON.stringify(fetchMocks));
+    env.AGENCY_FETCH_MOCKS_FILE = mocksFile;
```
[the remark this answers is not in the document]
egonSchiele: Fixed. The shim now installs whenever fetchMocks is defined (including an empty array), and the runner preserves the distinction: not-declared (undefined) leaves the shim off, declared-empty ([]) installs it so every fetch throws. Added an agency e2e (fetchMockEmpty) proving [] blocks fetches instead of falling through to real network.

packages/agency-lang/lib/cli/fetchMockResolve.ts
```diff
@@ -0,0 +1,33 @@
...
+    throw new Error(`fetchMock[${i}]: set only one of "return" or "returnFile".`);
+  }
+  const filePath = path.resolve(baseDir, m.returnFile);
+  if (!fs.existsSync(filePath)) {
+    throw new Error(`fetchMock[${i}]: returnFile not found: ${filePath}`);
+  }
+  const contents = fs.readFileSync(filePath, "utf-8");
+  const { returnFile, ...rest } = m;
+  return { ...rest, return: contents };
+}
```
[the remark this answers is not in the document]
egonSchiele: Added the validation to resolveFetchMocks: an entry with neither return nor returnFile now throws pointing at the test file, instead of deferring to runtime. Unit-tested.

packages/agency-lang/lib/runtime/fetchMock.ts
```diff
@@ -0,0 +1,205 @@
...
+  let raw = init?.method;
+  if (raw === undefined && isRequestInput(input)) {
+    raw = input.method;
+  }
+  return String(raw ?? "GET").toUpperCase();
+}
+
+export function installFetchMock(mocks: FetchMock[]): () => void {
+  const compiled = mocks.map(compileMock);
+  const real = globalThis.fetch;
```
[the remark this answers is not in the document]
egonSchiele: Added an Array.isArray guard in installFetchMock with a clear message. Unit-tested.

packages/agency-lang/tests/agency-js/fetch-mock/agent.agency
egonSchiele: Keeping it as an agency-js test: it is the only coverage of the agency-js runner's fetchMocks.json path (reading the file, temp-file wiring, cleanup), which is a separate code path from the .test.json runner. A regular agency test would leave that wiring untested in CI.

PR #451, opened 2026-07-08
Durable object tags: plain object/array tags survive fork and interrupt/resume

egonSchiele:
## Summary

Makes `std::tag` tags on **plain objects and arrays durable**: they now survive `fork`/`race`/`parallel` branches and interrupt/resume, the same way primitive (value) tags already do. This closes the one limitation deferred from the base value-tags feature (PR #447).

- **Design spec:** `docs/superpowers/specs/2026-07-07-durable-object-tags-design.md` (+ review)
- **Plan:** `docs/superpowers/plans/2026-07-07-durable-object-tags.md` (+ review)

## How it works

- A tagged plain object/array carries its tag record as a **non-enumerable `Symbol` property** (`lib/runtime/state/tagSymbol.ts`) — invisible to `Object.keys`, spread, and `JSON.stringify`.
- A new **`TaggedReviver`** in the shared `nativeTypeReplacer`/`nativeTypeReviver` array preserves that property through every state round-trip (`deepClone`, state-stack serialization, `GlobalStore` clone/`toJSON`/`fromJSON`).
- **Frozen/sealed and native-typed objects** (`Date`, `Map`, class instances, …) can't carry the marker; they keep the branch-local `WeakMap` fallback, unchanged.
- `GlobalStore.tagsRecordFor` resolves an **existing record first** (on-object, then WeakMap) and only chooses storage by capability for brand-new tags — so an object frozen *after* tagging still reads, writes, and clears its durable record.

## Redaction-gate hardening (two mechanisms, on purpose)

`statelogClient.post` skips the redaction pass entirely when `hasAnyTags()` is false, so a tag whose *presence signal* doesn't reach a store is a leak. Two guards keep the gate correct:

1. **Locally-true flag on adopt-write:** the serialized `__hasDurableObjectTags` flag is set whenever a store creates *or adopts* (writes to) a durable record — even one another store created — so the gate never depends on propagation ordering.
2. **Join OR-in:** `runBatch` ORs a settling branch's flag into the parent store in a `finally`, covering value, interrupt, **and throw** settles (a thrown error can reference a branch-tagged object the parent logs).

The WeakMap path keeps its separate in-memory bit that resets on clone, so one frozen-object tag doesn't permanently over-approximate descendants.

## Discovered while executing the plan (not in the spec)

Running the new end-to-end fixture surfaced two additional leak paths that defeated redaction *regardless* of the gate, both fixed here:

- **`safeStatelogValue` (forkBranchEnd value):** the telemetry copy used a plain JSON round-trip, which stripped the on-object tag before the redaction replacer ran. It now clones through the shared replacer/reviver pair (behavior otherwise unchanged; still never throws).
- **Out-of-frame `agentEnd` posts:** the result-bearing `agentEnd` event fires *after* the run's ALS frame ends, where `post()` previously skipped redaction entirely ("documented boundary"). The client now redacts against a fallback top-level-store accessor wired by the execution context (a getter, since checkpoint restore reassigns `execCtx.globals`). This also fixes the same pre-existing gap for **primitive** tags.

## Semantics changes

- Plain-object/array tags survive `fork`/`race`/`parallel` and interrupt/resume (guide updated; branch-local caveat narrowed to frozen/native objects).
- A spread/structural copy (`{...obj}`) yields an **untagged** object — durability follows the same object, not copies.
- `removeAllTags` on a durably-tagged object clears the record's keys in place, so `getTagsFor` returns `{}` (not `undefined`) afterward — deleting the property would throw on frozen-after-tag objects. `isRedacted` (`=== true`) is unaffected.
- Two existing tests that pinned "plain-object tags are branch-local" now pin that contract for frozen objects only.

## Testing

- Unit: `tagSymbol.test.ts` (new), `taggedReviver.test.ts` (new), `globalStore.tags.test.ts` (+6 incl. adopt-flag and freeze-after-tag), `runBatch.test.ts` (+3 incl. throw-path propagation), `runner.test.ts` (+2 for `safeStatelogValue`), `statelogClient.redaction.test.ts` (+2: durable round-trip, out-of-frame fallback).
- End-to-end: `tests/agency-js/tag-fork-redaction/` — a fork branch `redact()`s an object and returns it; the parent statelog must show `[REDACTED]` and never the raw value.
- Execution: `tests/agency/tag.agency` `forkInheritsObjectTag` (compiled fork path).
- Full unit suite: 7367 passed / 0 failed. `lint:structure` clean. `make` (tsc + fixtures + docs regen) clean; stdlib reference regen produced no diff.

### DIFF
--- a/packages/agency-lang/lib/runtime/revivers/taggedReviver.ts
+++ b/packages/agency-lang/lib/runtime/revivers/taggedReviver.ts
@@ -1,0 +1,40 @@
+import { BaseReviver } from "./baseReviver.js";
+import {
+  isPlainObjectOrArray,
+  attachTag,
+  readTag,
+} from "../state/tagSymbol.js";
+
+// Preserves a plain object's / array's durable tag (a non-enumerable TAG_SYMBOL
+// property) across the shared JSON round-trip. Plain objects/arrays match no
+// other reviver, so registration order is irrelevant.
+export class TaggedReviver implements BaseReviver<object> {
+  nativeTypeName(): string {
+    return "Tagged";
+  }
+
+  isInstance(value: unknown): value is object {
+    // Cheap symbol read first, then confirm it's a plain object/array.
+    return readTag(value) !== undefined && isPlainObjectOrArray(value);
+  }
+
+  serialize(value: object): Record<string, unknown> {
+    // Spread drops the non-enumerable symbol, so `v` is tag-free and recursing
+    // into it can't re-match this reviver (no loop). Nested natives and nested
+    // tagged values inside `v` recurse normally.
+    const v = Array.isArray(value) ? [...value] : { ...value };
+    return { __nativeType: this.nativeTypeName(), tags: readTag(value), v };
+  }
+
+  validate(value: Record<string, unknown>): boolean {
+    return "tags" in value && "v" in value;
+  }
+
+  revive(value: Record<string, unknown>): object {
+    // `v` is already revived (JSON.parse revivers run bottom-up). attachTag
+    // re-hides the tag and restores the null prototype the round-trip stripped.
+    const target = value.v as object;
+    attachTag(target, value.tags as Record<string, unknown>);
+    return target;
+  }
+}
--- a/packages/agency-lang/docs/site/guide/tags.md
+++ b/packages/agency-lang/docs/site/guide/tags.md
@@ -32,9 +32,12 @@ How a tag is stored depends on the kind of value:
 - **Objects and arrays** are keyed by **reference**. Tagging one object does
   *not* tag a structurally-equal but distinct object.
 
-> Object and array tags are branch-local: they do not survive `fork`, `race`,
-> or `parallel` branches, or an interrupt/resume. Primitive (value) tags
-> survive both.
+> Tags on **plain objects and arrays** survive `fork`, `race`, `parallel`,
+> and interrupt/resume, the same as primitive (value) tags. A **spread or
+> structural copy** (`{...obj}`) produces an *untagged* new object — durability
+> follows the *same* object, not its copies. Tags on **frozen/sealed objects**
+> and **native-typed objects** (`Date`, `Map`, `Set`, …) are branch-local
+> (best-effort) — they can't carry the durable marker.
 
 ## Redaction
 
--- a/packages/agency-lang/docs/superpowers/plans/2026-07-07-durable-object-tags-review.md
+++ b/packages/agency-lang/docs/superpowers/plans/2026-07-07-durable-object-tags-review.md
@@ -1,0 +1,125 @@
+# Review: Durable Object Tags Implementation Plan (2026-07-07)
+
+**Plan:** `2026-07-07-durable-object-tags.md`
+**Spec:** `docs/superpowers/specs/2026-07-07-durable-object-tags-design.md`
+**Spec review:** `docs/superpowers/specs/2026-07-07-durable-object-tags-design-review.md`
+**Verdict:** Approve with revisions — Finding A will make Task 4's fixture fail
+as written and must be fixed before execution; Findings B and C are one-line
+hardenings to the flag machinery; Finding D needs a documenting sentence.
+
+All harness/helper/syntax claims below were verified against the worktree code,
+not just read off the plan.
+
+## What checks out (verified against code)
+
+- **Spec-review findings are all folded in correctly.** Finding 1 → Task 4
+  (join OR-in), Finding 2 → Task 3 (`removeAllTags` clears keys, never
+  `delete`s), Finding 3 → Task 1 (`attachTag` forces null prototype, exercised
+  by Task 2's revive test), split-flag suggestion → the two-flag Global
+  Constraint + Task 3, `tagsRecordFor` naming → Task 3 edits the unified helper
+  rather than forking per-method.
+- **The read-existing-first dispatch is better than the spec.** Locating an
+  existing record before choosing a storage path makes `setTag` *and*
+  `removeAllTags` on a tagged-then-frozen object correct (the spec only handled
+  serialize/read after freeze). Task 3's freeze-after-tag test pins it.
+- **`runBatch` insertion point is right.** `shareGlobals`, `parent`, and
+  `branchGlobals` are all in scope in the settle closure
+  (`lib/runtime/runBatch.ts:336-388`), and both value and interrupt settles
+  return through that line.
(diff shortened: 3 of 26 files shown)


egonSchiele:
## Review — Durable object tags (PR #451)

Reviewed the diff end-to-end (not local files — this is a separate branch). Overall: **strong, ship-worthy PR.** It implements the plan faithfully and the two leak paths discovered during execution are exactly the kind of thing the plan couldn't have predicted. One finding I'd fix before merge, plus a couple of minors.

### What checks out
- **`tagSymbol.ts` + `TaggedReviver`** are correct: non-enumerable Symbol (dropped by spread/JSON), spread-in-`serialize` drops the marker so there's no re-match loop, bottom-up `revive` re-attaches, and `attachTag` restores the null prototype the round-trip strips (proto-safety preserved). Registration order genuinely doesn't matter (plain objects/arrays match no other reviver).
- **`GlobalStore` dispatch** — read-existing-first (durable → WeakMap) before consulting capability correctly handles *freeze-after-tag*; the **adopt-write sets the local flag** refinement (beyond the plan) is a nice touch — the gate is locally true whenever a store writes a durable record, so it never depends on join ordering.
- **`runBatch` join in a `finally`** is better than the plan's placement: it covers value, interrupt, **and throw** settles with one idempotent OR-in. Good call.
- **Fallback-globals fix** (`context.ts` → `setFallbackGlobals(() => this.globals)`) is right to use a *getter* (restore reassigns `globals`) and to keep `__globals()` winning inside branches. Cleanly closes the out-of-frame `agentEnd` gap (incl. the pre-existing primitive-tag case).
- Path separation stays intact (log path never touches the Tagged reviver), and unit/e2e coverage is thorough.

### Finding 1 (should-fix): `safeStatelogValue` truncation branch still leaks a redacted value
`lib/runtime/runner.ts`, `safeStatelogValue` — the clone path was fixed to preserve the tag (good, and tested), but the **oversized branch is unprotected**:

```
if (json.length > FORK_VALUE_CHAR_CAP) {
  return json.slice(0, FORK_VALUE_CHAR_CAP) + "…[truncated]";
}
```

For a durably-tagged (redacted) object, `json` is `{"__nativeType":"Tagged","tags":{"redact":true},"v":{…raw fields…}}`. The slice includes `v` — the real field values — and is returned as a **plain string**. At `post()` time the redaction replacer only redacts *tagged values*, and a string leaf isn't one, so an oversized `redact`ed object returned from a `fork`/`race`/`parallel` branch **leaks its first 4000 chars into the `forkBranchEnd` event**. This is exactly the data redaction promises to suppress; narrow path, but a real leak, and there's no test for it (the new `runner.test.ts` case only exercises the small/clone path).

Suggested direction: truncation needs to happen *after* redaction, not before. Since `safeStatelogValue` has no `globals` access, either (a) return the object clone and move the size cap to `post()` after the redaction pass, or (b) have the fork telemetry redact the clone before measuring/truncating. A regression test: a branch returns a `redact`ed object whose serialization exceeds `FORK_VALUE_CHAR_CAP`; assert the raw value never appears in statelog.

### Minor
- **`removeAllTags` leaves the object permanently "Tagged"-wrapped.** Clearing keys in place (correct — `delete` throws on frozen-after-tag) leaves an empty record attached, so `TaggedReviver.isInstance` keeps matching and wraps the object with `tags:{}` on every future serialization, and `hasDurableObjectTagFlag` stays true forever. Harmless, but for the common non-frozen case you could `delete` the property (guarded by `isExtensible`) and only fall back to clear-keys when frozen — avoids the perpetual wrap. The `getTagsFor` returns `{}` vs `undefined` asymmetry is well-documented.
- **Null-proto tagged object → `Object.prototype` after round-trip.** `serialize` uses `{...value}`, so a tagged `Object.create(null)` value comes back with a normal prototype. Rare and low-impact; noting for completeness.

### Verdict
Approve once Finding 1 is addressed (or explicitly accepted as a known narrow gap with a tracking note). Everything else is solid.

packages/agency-lang/lib/runtime/revivers/taggedReviver.ts
```diff
@@ -0,0 +1,40 @@
...
+    const v = Array.isArray(value) ? [...value] : { ...value };
+    return { __nativeType: this.nativeTypeName(), tags: readTag(value), v };
+  }
+
+  validate(value: Record<string, unknown>): boolean {
+    return "tags" in value && "v" in value;
+  }
+
+  revive(value: Record<string, unknown>): object {
+    // `v` is already revived (JSON.parse revivers run bottom-up). attachTag
```
egonSchiele, line 34: `v` would be very very hard to grep for, if we ever wanted to find all mentions of it

egonSchiele:
Addressed in 0c2767de.

**Finding 1 (truncation leak) — fixed.** Redaction now happens *during* the `safeStatelogValue` stringify instead of relying on `post()`: the redact check (`makeRedactReplacer` over the caller's ALS store, read leniently) composes with the shared `nativeTypeReplacer`, so the truncated string is already scrubbed before slicing. The small path is wire-identical to before (the clone now carries `[REDACTED]` in place of the value rather than a tag that `post()` would redact anyway). With no ALS frame it degrades to the previous tag-preserving clone — and both real call sites run at the fork join inside the parent frame. Regression tests: an oversized payload with a nested `redact`ed object asserts the truncated string contains `[REDACTED]` and never the raw value, plus a small-path-with-frame case.

**Minor 1 (perpetual "Tagged" wrap) — taken.** `removeAllTags` on an extensible target now detaches the symbol property outright (`detachTag` in `tagSymbol.ts` — `isExtensible` is a safe proxy for "delete won't throw", since our property is defined configurable and only freeze/seal flips that, both of which also make the object non-extensible). The object stops matching the `TaggedReviver` and `getTagsFor` returns `undefined`, consistent with the other paths. Only a frozen-after-tag target falls back to clear-keys-in-place and keeps the `{}` asymmetry (now narrowly scoped and documented at both the impl and the test). `hasDurableObjectTagFlag` stays monotonic by design.

**Minor 2 (null-proto value loses its prototype) — no change, with reasoning.** This is a property of the shared JSON round-trip, not of the `TaggedReviver`: `deepClone` already normalizes *every* null-proto plain object to `Object.prototype` (JSON.parse only creates ordinary objects), tagged or not. Preserving it would need a proto marker in the wire format for all plain objects, which is out of scope here. Note the *tag record's* null prototype (the proto-pollution-relevant one) is restored on revive.

Verification: full unit suite 7370 passed / 0 failed, e2e `tag-fork-redaction` fixture and 7/7 tag execution tests pass, `make` + `lint:structure` clean.

PR #452, opened 2026-07-08
Test-only imports: import test { ... } for non-exported helpers

egonSchiele:
## Summary

Adds test-only imports: `import test { ... }` lets the test harness import non-exported Agency helpers, so stdlib connectors no longer have to `export` internal parse/build helpers just for their tests. Internal helpers now stay out of the generated docs and are unreachable from normal programs and the `run()` subprocess sandbox.

Design spec and plan: `docs/superpowers/` on the `worktree-test-only-imports` branch.

### What `export` used to conflate

The Agency `export` keyword previously controlled three things at once: doc visibility, import eligibility, and the JS `export` of the compiled wrapper. Connector tests needed the third one, which forced the first two. This PR decouples them:

1. **Syntax** — `import test { ... }` parses into `ImportStatement.testOnly` (key absent on normal imports, so AST comparisons and `agency ast` JSON stay clean). A `peek({)` lookahead keeps `import test from "./x.ts"` and `import test, { bar } from "./x.ts"` parsing as default imports named `test`.
2. **Codegen** — the compiled function wrapper `const` is now JS-exported unconditionally; the `exported:` metadata on `AgencyFunction.create` still records true Agency visibility. Agency-level visibility is enforced at import resolution and in the docs generator.
3. **Enforcement** — a single `assertImportable` chokepoint in `resolveImports`: test-only imports skip the export check, but only under the test harness (`allowTestImports`), and only for first-party (`std::` + local) modules — `pkg::` is always rejected.

### Trust boundary

`allowTestImports` is deliberately NOT an `AgencyConfig` field, so agent-authored source can never enable it. Default is deny everywhere; it is set `true` only by:

- the test runner (`lib/cli/test.ts`, all four compile/execute sites), and
- the two analysis-only paths (`typeCheckSource`, LSP diagnostics) which never execute anything — this keeps migrated test files fully working in the editor instead of dying on a single 0:0 error.

`compileSource` (the `run()` subprocess sandbox) passes an explicit `false`, with a regression-guard test that imports a REAL non-exported stdlib symbol — if the privilege ever leaks, compilation would succeed and the guard fails unambiguously.

### Migrations

- **littlesis**: 13 internal helpers + the `CategoryFilter` type lose `export`; the docs page drops 257 lines and now lists only the 4 public functions + `Entity`/`Relationship`.
- **finance (gdelt, fred, edgar, dbnomics)**: 18 internal helpers lose `export`; docs drop 346 lines. Public entry points and their signature types stay exported.
- Live-test variants keep plain `import` (they only use public symbols), demonstrating the normal path still works.

### Also in this PR

- `pnpm run fmt` round-trips `import test` (AgencyGenerator emits the keyword; without this, formatting a migrated test file would silently break it).
- New `tests/agency/test-only-imports/` execution test covering the async runner threading path and local-module (non-stdlib) test imports.
- User docs: guide section in `imports-and-packages.md` + a note in `docs/misc/TESTING.md`.
- Regenerated TS fixtures (the `export const` widening) and debugger fixtures (which also picked up pending fetch-mock template drift from main).

## Test plan

- [x] `pnpm test:run` — 7356 passed, 0 failed
- [x] `pnpm run lint:structure` — clean
- [x] `tsc --noEmit` — clean
- [x] `pnpm run agency test tests/agency/test-only-imports/main.agency` — pass
- [x] `pnpm run compile tests/agency/test-only-imports/main.agency` — correctly rejected ("only allowed under the test harness")
- [x] agency-js: `data-people-littlesis`, `data-finance-{gdelt,fred,edgar,dbnomics}` — all pass
- [ ] CI: full agency test suite


packages/agency-lang/lib/preprocessors/importResolver.ts
```diff
@@ -41,6 +58,19 @@ export function resolveImports(
       continue;
     }
 
+    // Per-statement gate for test-only imports. Must run before symbol
+    // lookup so a pkg:: rejection does not depend on the package resolving.
+    if (node.testOnly) {
```
egonSchiele, line 66: **Contract gap: `import test` on a non-Agency path silently bypasses this whole gate.**

This gate runs only *after* the `if (node.type !== "importStatement" || !isAgencyImport(node.modulePath))` short-circuit a few lines above, which pushes any non-Agency import through untouched. `isAgencyImport` is true only for `.agency` / `std::` / `pkg::` paths — so `import test { x } from "./foo.ts"` or `import test { x } from "lodash"` (raw JS/TS or bare npm) never reaches this block.

Consequences:
- In a **normal** (non-harness) program, `import test { x } from "./foo.ts"` is **not** rejected — it contradicts the "`import test` is only allowed under the test harness" contract.
- The `test` keyword becomes a silent no-op there (parsed, then ignored), which is confusing rather than either working or erroring.

This is not a capability escalation — raw JS/TS imports have no export check to skip, so `import test` from them behaves identically to a plain import (already unrestricted). But it's a hole in the enforcement surface. Consider erroring on `node.testOnly && !isAgencyImport(node.modulePath)` (e.g. "`import test` is only for first-party `std::` and local `.agency` modules") so the keyword never silently degrades. The cleanest spot is to check `testOnly` before this short-circuit.

packages/agency-lang/lib/compiler/typecheck.ts
```diff
@@ -103,7 +103,12 @@ export function typeCheckSource(
   return withSourcePath(source, sourcePath, (syntheticPath) => {
     const symbolTable = SymbolTable.build(syntheticPath, {});
     const reExported = resolveReExports(program, symbolTable, syntheticPath);
-    const resolved = resolveImports(reExported, symbolTable, syntheticPath);
+    // Analysis-only path (never executes anything): honor `import test` so
+    // test files can be type-checked. The execution trust boundary lives in
+    // compileSource / compile(), which default to deny.
+    const resolved = resolveImports(reExported, symbolTable, syntheticPath, {
+      allowTestImports: true,
```
egonSchiele: **`typeCheckSource` is agent-reachable, not just the LSP — worth confirming this is intended.**

Allowing `import test` here is sound for editor DX, but `typeCheckSource` is also exposed to Agency programs through `std::agency`'s typecheck (`lib/stdlib/agency.ts:173` and `:181`), not only the LSP. So agent-callable typecheck now reports `import test` code as **valid**, even though the same code hits `allowTestImports: false` in `compileSource`/`compile()` and is rejected when actually run.

That's a typecheck-vs-execution mismatch visible to agent code (typecheck says OK, `run()` refuses). It's not a security issue — typecheck grants no execution — but if you'd prefer typecheck to agree with what will actually run, `typeCheckSource` could pass `allowTestImports: false` while the LSP (`diagnostics.ts`) keeps `true` for editor support. The two paths don't have to share the decision. Flagging for an explicit call rather than recommending a change.

egonSchiele (commented):
Reviewed the implementation against the design/plan and the surrounding code. Overall this is faithful to the plan and clean: the `assertImportable` chokepoint, the entrypoint-only `allowTestImports` (never on `AgencyConfig`, default-deny, threaded through **both** the sync `executeNode` and async `runAgencyNode` compile sites plus all four `test.ts` sites), the unconditional wrapper export with `exported:` metadata preserved, the `testOnly`-key-absent parser normalization, and the fmt round-trip are all correct. I also confirmed no non-test first-party module imports any of the de-exported helpers (only the test files and the doc-comment examples reference these modules, and the examples use only still-exported public symbols), and the leak-guard + LSP tests use real symbols/modules so they fail for the right reason.

Two non-blocking findings below — both low severity (neither is a capability escalation), but each is a small gap between the stated contract and the behavior. Worth a look before merge; neither blocks it.

packages/agency-lang/docs/site/guide/imports-and-packages.md
```diff
@@ -98,3 +98,30 @@ export { safe search } from "std::wikipedia"
 export { safe search, fetch } from "std::wikipedia"
 ```
 
+## Test-only imports
```
egonSchiele: remove the docs change please

packages/agency-lang/lib/preprocessors/importResolver.ts
egonSchiele: Fixed in 729c8605. The `testOnly` gate now runs before the non-Agency short-circuit, and a non-first-party path (`.ts` file or bare npm) is rejected with "`import test` cannot be used with TypeScript or npm imports; it is only for first-party (std:: and local) modules." — unconditionally, even outside test mode, so the keyword can never silently degrade to a no-op. Added two resolver tests pinning both cases (relative `.ts` in test mode, bare npm outside it).

packages/agency-lang/lib/compiler/typecheck.ts
egonSchiele: Good catch — I had missed that `typeCheckSource` is agent-reachable through `std::agency`. Made the call in 729c8605: `typeCheckSource` now passes `allowTestImports: false` so agent-visible typecheck agrees with what `run()`/`compileSource` will actually do, and only the LSP (`diagnostics.ts`) keeps `true` for editor support. Added a test pinning it: `typeCheckSource` on `import test` source throws "only allowed under the test harness".

packages/agency-lang/docs/site/guide/imports-and-packages.md
egonSchiele: Removed in 729c8605 — the guide page is back to what main has. I kept the short note in `docs/misc/TESTING.md` since that is the dev-facing testing doc rather than the published guide; say the word if you want that gone too.

PR #455, opened 2026-07-08
Fix std::http fetch functions to fail on non-2xx HTTP status

egonSchiele:
## What

`std::http`'s `fetch`, `fetchJSON`, and `fetchMarkdown` return a `Result`, but their implementations never checked the HTTP response status. A `4xx`/`5xx` with a parseable body was returned as `success` — e.g. `fetchJSON` against an API that returns a JSON error body on failure gave `success({error: ...})`, and `fetch`/`fetchMarkdown` returned 404 error pages as success text.

## Fix

Add a shared `ensureOkStatus(result, url, body)` helper in `lib/stdlib/http.ts` and call it in all three impls after the (size-capped) body is read, before parse/convert. It throws on `!result.ok` (status outside `[200, 300)`) with a message carrying the status and a body snippet. The existing `return try _fetch*(...)` wrappers in `stdlib/http.agency` turn the throw into a `Result` failure — no agency-side or `Result`-shape change.

## Tests

`tests/agency/httpStatus.{agency,test.json}` (fetch-mock based):
- `fetchJSON` 404 with a valid JSON body → failure (the status check fires before `JSON.parse`; the message carries the status + body snippet)
- `fetchJSON` 500, `fetch` 404, `fetchMarkdown` 503 → failure
- `fetchJSON` 200 → success, body parsed (regression guard)

No regressions: existing `fetchMock` / `fetchMockEmpty` agency tests, the agency-js `fetch-mock` test, and 27 runtime fetch-mock unit tests all pass.

## Blast radius

Audited: no existing test sets a non-200 mock status, and no connector reads error-body fields. Existing connectors only improve (FRED bad key → 400, EDGAR missing User-Agent → 403, LittleSis rate limit → 503 now surface as clean failures); GDELT/DBnomics use 200 bodies and are unchanged.

### DIFF
--- a/packages/agency-lang/lib/stdlib/http.ts
+++ b/packages/agency-lang/lib/stdlib/http.ts
@@ -67,6 +67,21 @@ function validateUrl(
 }
 
 /**
+ * Throw on a non-2xx response so std::http's `try`-wrapped callers surface a
+ * `failure` instead of treating an error page / JSON error body as success.
+ * `Response.ok` is true iff the status is in [200, 300). The body has already
+ * been read (and size-capped) by the caller, so include a snippet — most APIs
+ * put the failure reason in the body.
+ */
+function ensureOkStatus(result: Response, url: string, body: string): void {
+  if (!result.ok) {
+    throw new Error(
+      `HTTP ${result.status} ${result.statusText} from ${url}: ${body.slice(0, 500)}`,
+    );
+  }
+}
+
+/**
  * Run an HTTP request, translating any abort-shaped error into the
  * runtime's `AgencyCancelledError`. Node's `fetch` surfaces an
  * aborted request as a `DOMException` with `name === "AbortError"`,
@@ -122,12 +137,15 @@ async function fetchImpl(
   const signal = ctx.getAbortSignal(stack);
   return await runHttp(async () => {
     const result = await fetch(url, { headers, signal });
+    let body: string;
     try {
-      return await readBodyCapped(result, url, signal);
+      body = await readBodyCapped(result, url, signal);
     } catch (e) {
       if (isAbortError(e)) throw e;
       throw new Error(`Failed to get text from ${url}: ${e}`);
     }
+    ensureOkStatus(result, url, body);
+    return body;
   }, url);
 }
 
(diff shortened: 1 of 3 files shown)


packages/agency-lang/lib/stdlib/http.ts
```diff
@@ -66,6 +66,21 @@ function validateUrl(
...
+/**
+ * Throw on a non-2xx response so std::http's `try`-wrapped callers surface a
+ * `failure` instead of treating an error page / JSON error body as success.
+ * `Response.ok` is true iff the status is in [200, 300). The body has already
+ * been read (and size-capped) by the caller, so include a snippet — most APIs
+ * put the failure reason in the body.
+ */
+function ensureOkStatus(result: Response, url: string, body: string): void {
+  if (!result.ok) {
+    throw new Error(
```
egonSchiele: would be nicer to instead return a `failure` with structured fields for status, statusText etc

egonSchiele: You can import the `failure` func from `agency-lang` in ts

egonSchiele: Done in 892a3cec. Non-2xx now returns failure({ status, statusText, url, body, message }) instead of throwing a flat string, so callers can branch on e.status directly. The impls return the failure and __tryCall passes a returned Result through, so the agency-side `try _fetch*()` wrappers need no change.

egonSchiele: Done — imported `{ failure, type ResultFailure }` from `../runtime/result.js` (the same module image.ts uses).

packages/agency-lang/lib/stdlib/http.ts
```diff
@@ -66,6 +66,21 @@ function validateUrl(
...
+      `HTTP ${result.status} ${result.statusText} from ${url}: ${body.slice(0, 500)}`,
+    );
+  }
+}
```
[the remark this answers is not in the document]
egonSchiele: Addressed in 892a3cec: statusText is omitted when empty (no double space), the body snippet is whitespace-collapsed (no stray newlines) and gets an ellipsis when truncated. The status/statusText/body are also available as structured fields on the failure error object now, so callers don't have to parse the message.

PR #462, opened 2026-07-08
Add --policy flag with workspace-scoped built-in policies

egonSchiele:
## What

Adds a `--policy` flag to the agency agent and two workspace-scoped built-in approval policies, so a run can approve writes without a human while staying confined to the working directory.

### `--policy <name-or-path>`
Selects the approval policy for a single run. Accepts a **built-in name** or a **path to a policy JSON file**. It overrides the saved `~/.agency-agent/policy.json` for that run only and is **not persisted**.

### New built-ins (4 total: `minimal`, `recommended`, `with-writes`, `with-bash`)
- **`with-writes`** — `recommended` + auto-approve file writes (`write`/`writeBinary`/`edit`/`mkdir`/`remove`/`copy`/`move`) and git-write commands, **scoped to the launch directory and its children**.
- **`with-bash`** — `with-writes` + `std::bash`, also cwd-scoped.

Scoping uses policy `match` rules on each effect's path field (`dir` / `target` / `src`+`dest` / `cwd`) with the same `{base,base/**}` glob convention as the interactive "approve-always-here" rule.

### Safety
- **File writes are genuinely confined**: `write`/`edit`/`writeBinary` run `filename` through `resolvePath` (rejects `..`, absolute, `~`, symlink escapes), and the scope glob won't match a `..`-containing path (picomatch `dot:false`). An absolute path outside the workspace simply doesn't match and falls through to prompting.
- **`std::bash` confinement is best-effort** — a command can still reach outside via an absolute path or `cd` inside the command string (the policy matches the declared `cwd`, not the shell's actual effects). So `with-bash` is intended for disposable/sandbox runs.

### `cliPolicyHandler` in-memory policy (std::policy)
Adds an optional `policy` parameter to `cliPolicyHandler`. When provided, the handler starts from that in-memory policy and **skips the startup file read** (no load-time `std::read`, no dependency on the file existing); new "always" decisions still persist to `file`. Built-in `--policy` overrides use this with a session-file path, so a cwd-scoped rule can't clobber the saved policy or leak into a later run from a different directory. A `--policy <file>` points the handler directly at that file (normal load/persist).

## Why
Running the agency agent non-interactively (e.g. benchmarking on Terminal-Bench) needs writes approved without a human — but scoped to the workspace, not blanket-approved everywhere.

## Testing
- Verified scoping via `checkPolicy` over synthetic `std::write` / `std::bash` / `std::git::commit` interrupts (no LLM): in-cwd approved, out-of-cwd prompts; `bash` only under `with-bash`.
- Verified the in-memory handler path approves an in-scope write while never reading a (nonexistent) policy file.
- `--policy <bad-name>` exits 1 with a clear message before any LLM call.
- All four changed `.agency` files typecheck clean; `make` builds.

## Notes
- Generated `docs/site/stdlib/policy.md` is included (regenerated from the `cliPolicyHandler` docstring).
- `agent.agency` also shows formatter churn (import reordering / line-wrapping) from a `fmt` pass; functionally unrelated to this change.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
@@ -132,24 +132,12 @@ def dirScope(baseDir: string): string {
 }
 
 // `recommended` plus auto-approval of the mutating file-system and
-// git-write effects, each SCOPED to `baseDir` and its children. Writes
-// (or git actions) whose path lands outside `baseDir` still prompt.
-//
-// The scope is matched on each effect's path field: `dir` for
-// write/edit/mkdir, `target` for remove, `src`+`dest` for copy/move, and
-// `cwd` for the git commands (which resolve an empty cwd to the agent
-// working directory before raising the interrupt). This mirrors the
-// fields `ALWAYS_FIELDS` pins for the interactive "here" rule.
-//
-// This is a genuine confinement for file writes, guarded on two layers:
-// (1) `std::write`/`std::edit`/`std::writeBinary` run their `filename`
-// through `resolvePath`, which rejects `..`, absolute paths, `~`, and
-// symlink escapes at runtime — so `dir` matching `baseDir` really does
-// pin the write inside it; and (2) the scope glob itself won't match a
-// value containing `..` (picomatch, dot:false), so effects matched on a
-// whole-path field (remove/copy/move/mkdir) can't be approved for a
-// `..`-escaping path either. Absolute paths outside `baseDir` simply
-// don't match and fall through to prompting.
+// git-write effects, each scoped to `baseDir` and its children; anything
+// outside still prompts. Matched on each effect's path field: `dir`
+// (write/edit/mkdir), `target` (remove), `src`+`dest` (copy/move), `cwd`
+// (git). Confinement is real: write/edit run `filename` through
+// `resolvePath` (rejects `..`/absolute/`~`/symlink) and the scope glob
+// won't match a `..`-containing path.
 export def withWritesPolicy(baseDir: string) {
   const scope = dirScope(baseDir)
   const dirRule = [{ match: { dir: scope }, action: "approve" }]
@@ -193,8 +181,8 @@ export def withBashPolicy(baseDir: string) {
 // Built-in policy names accepted by the `--policy` flag (anything else
 // is treated as a path to a policy JSON file).
 export static const BUILTIN_POLICY_NAMES = [
-  "minimal",
   "recommended",
+  "minimal",
   "with-writes",
   "with-bash",
 ]
--- a/packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
@@ -178,14 +178,35 @@ export def withBashPolicy(baseDir: string) {
   }
 }
 
-// Built-in policy names accepted by the `--policy` flag (anything else
-// is treated as a path to a policy JSON file).
-export static const BUILTIN_POLICY_NAMES = [
-  "recommended",
-  "minimal",
-  "with-writes",
-  "with-bash",
+// Built-in policies accepted by the `--policy` flag (anything else is
+// treated as a path to a policy JSON file), in display order with a
+// one-line description each. Single source of truth for the flag's help
+// text and its unknown-value error.
+export static const BUILTIN_POLICIES = [
+  {
+    name: "recommended",
+    description: "Auto-approve reads and web/search; prompt for writes, shell, and git changes. (Default.)"
+  },
+  {
+    name: "minimal",
+    description: "Auto-approve only memory and the safe read-only agency subcommands; prompt for everything else."
+  },
+  {
+    name: "with-writes",
+    description: "recommended + auto-approve file writes and git changes, scoped to the current directory and its children."
+  },
+  {
+    name: "with-bash",
+    description: "with-writes + auto-approve shell commands, scoped to the current directory. Best-effort confinement — use in disposable/sandbox runs."
+  },
 ]
+
+// Just the built-in names, e.g. for the unknown-value error message.
+export def builtinPolicyNames(): string[] {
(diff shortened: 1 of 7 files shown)


packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
```diff
@@ -120,3 +120,100 @@ export static const recommendedAutoApprovePolicy = {
...
 }
+
+// Glob that matches a directory and everything under it. This is the
+// same convention `std::policy`'s `buildScopedMatch` uses for the
+// interactive "approve-always-here" rule: the exact path OR any nested
+// path. `baseDir` must be absolute (the agent pins it to the launch cwd
+// via `setAgentCwd(cwd())`), so relative paths the model passes resolve
+// into this range and outside paths fall through to prompting.
+def dirScope(baseDir: string): string {
+  return "{${baseDir},${baseDir}/**}"
```
egonSchiele, line 131: Please confirm that this will actually work correctly with glob. Do you need this or just `baseDir`? Confirm instead of guessing.

packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
```diff
@@ -120,3 +120,100 @@ export static const recommendedAutoApprovePolicy = {
...
+}
+
+// `recommended` plus auto-approval of the mutating file-system and
```
egonSchiele, line 134: Is this super long comment adding any useful information? Will this be helpful for readers of this code in the future, or is this just you taking notes for yourself as you work through the plan?

packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
```diff
@@ -120,3 +120,100 @@ export static const recommendedAutoApprovePolicy = {
...
+  const scope = dirScope(baseDir)
+  return {
+    ...withWritesPolicy(baseDir),
+    "std::bash": [{ match: { cwd: scope }, action: "approve" }]
+  }
+}
+
+// Built-in policy names accepted by the `--policy` flag (anything else
+// is treated as a path to a policy JSON file).
+export static const BUILTIN_POLICY_NAMES = [
```
egonSchiele: Please put recommended first in the array so it shows as the default option for users

packages/agency-lang/scripts/agency.ts
```diff
@@ -78,6 +78,16 @@ import { startMcpServer } from "@/mcp/server.js";
 import { pathToFileURL } from "url";
 import { serveMcp, serveHttp } from "@/cli/serve.js";
 
+const MIN_NODE_MAJOR = 20;
```
egonSchiele: Undo the change to this file and package.json, I have already committed this to main in a different commit.

packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
egonSchiele: Confirmed empirically — the `{baseDir,baseDir/**}` union is required; `baseDir` alone isn't enough:

- `baseDir`        → matches `/work` but NOT `/work/sub/b.txt`
- `baseDir/**`     → matches `/work/sub/b.txt` but NOT `/work` itself
- `{baseDir,baseDir/**}` → matches both; rejects `/etc`, `/work/../etc` (picomatch, dot:false)

It's also the exact form `std::policy.buildScopedMatch` builds (policy.agency:247), so it stays consistent with the interactive approve-always-here rule.

packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
egonSchiele: Fair — trimmed to 7 lines. Kept the path-field mapping and the one load-bearing fact (writes are confined via `resolvePath` + the glob rejects `..`); dropped the rest.

packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
```diff
@@ -120,3 +120,100 @@ export static const recommendedAutoApprovePolicy = {
...
+
+// Glob that matches a directory and everything under it. This is the
+// same convention `std::policy`'s `buildScopedMatch` uses for the
+// interactive "approve-always-here" rule: the exact path OR any nested
+// path. `baseDir` must be absolute (the agent pins it to the launch cwd
+// via `setAgentCwd(cwd())`), so relative paths the model passes resolve
+// into this range and outside paths fall through to prompting.
+def dirScope(baseDir: string): string {
+  return "{${baseDir},${baseDir}/**}"
+}
```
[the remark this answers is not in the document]
egonSchiele: Real gap, but it's a property of the shared matcher rather than this code. `lib/runtime/policy.ts` runs picomatch on the raw value, and the existing `buildScopedMatch` (interactive "approve-always-here") builds the identical `{path,path/**}` globs — so both have this Windows-`\` limitation today, not just `with-writes`/`with-bash`. Normalizing only `dirScope`'s pattern wouldn't fix it, since the matched value (`data.dir`, from `path.resolve`) still contains `\`; the fix has to normalize both sides in the matcher. Better as a separate cross-platform fix to `matchesRule` than expanding this PR's scope. (The agent is currently used on macOS/Linux.)

packages/agency-lang/lib/agents/agency-agent/lib/defaultPolicy.agency
egonSchiele: Done — `recommended` is now first in `BUILTIN_POLICY_NAMES`.

packages/agency-lang/scripts/agency.ts
egonSchiele: Reverted (commit `90257009`) — the net PR diff for `scripts/agency.ts` and `package.json` is now empty and they've dropped out of the PR's file list. The commit had landed on this branch, so I used a revert rather than dropping it, to respect the repo's no-force-push rule; if you'd rather it be gone from history entirely, I can force-push with your OK.

PR #463, opened 2026-07-09
Incremental builds Stage C: tsc incremental, make clean, single-process agency compile

egonSchiele:
## What

Stage C of the incremental-builds work (spec: `docs/superpowers/specs/2026-07-08-incremental-builds-design.md`; plan: `docs/superpowers/plans/2026-07-08-incremental-builds-stage-c.md`, both reviewed). Local-`make` iteration speedup with no new invalidation machinery:

- **tsc incremental**: flag enabled, buildinfo at `.agency-build/tsc.tsbuildinfo` — deliberately outside `dist/`, which ships wholesale in the npm tarball. Measured tsc: cold 6.9s → no-change 1.9s.
- **`make clean` + clean publish**: the default build no longer wipes `dist/`; `make clean` resets dist, scratch state, AND the `stdlib/**/*.js` siblings (also tarball-shipped), and `make publish` routes through it — an orphaned output from a deleted/renamed source can never ship. Publish deliberately runs no tests (CI gates merges) and may leave regenerated `docs/site/stdlib/` modified.
- **Single-process agency compile**: stdlib + all five agent dirs compile in ONE `node` invocation (the `compile` CLI was already variadic) — node boot and compiler import paid once, stdlib prelude parsed once via the parse cache from #457. Agent staging is factored into `AGENT_DIRS` + canned recipes.
- **Fixes riding along**: the pre-existing nested `dist/lib/agents/docs/guide/guide` artifact (repeated `cp -r` onto an existing dir) no longer ships in tarballs; and `tsc`/`tsc-alias` now resolve from `./node_modules/.bin` explicitly — bare `tsc` in the Makefile silently used whatever global install existed whenever make was not invoked through pnpm.

## Measured (M-series laptop)

| Scenario | Before | After |
|---|---|---|
| warm `make`, no changes | 16.8s | **6.9s** |
| one `.ts` edit | 16.8s | 7.0s |
| one `.agency` edit | 16.8s | 7.2s |
| cold (`make clean && make`) | 16.8s | 12.0s |

The plan's stretch target was 5–6s warm; the residual ~3.9s is irreducible per-file agency compile work (94 files, parse-once + walk + emit), which Stage A (content-hash manifest, skip unchanged files) addresses. Flagged per the plan's spec gate rather than silently shipped.

## Verification

- Stale-buildinfo guard exercised: hand-deleted `dist/` + surviving buildinfo rebuilds a full dist.
- `make agents` standalone, `.js`-sibling walker over all six roots, no `guide/guide` nesting.
- Incremental-vs-clean equivalence: post-edit-matrix clean rebuild, then `make fixtures` — byte-identical (`git status -- tests` clean).
- Orphan test: scratch stdlib module compiled, source deleted, publish-path rebuild → orphan gone.
- Runtime smoke: 72 execution-test files over `tests/agency/blocks` + `tests/agency/stdlib` pass against the consolidated-compiled stdlib.
- `npm pack --dry-run`: no buildinfo, no nested docs.
- CI unaffected: CI calls bare `make` (= `all`), behavior-equivalent, always cold on fresh checkouts.

### DIFF
--- a/packages/agency-lang/makefile
+++ b/packages/agency-lang/makefile
@@ -1,41 +1,98 @@
-.PHONY: all test stdlib agents doc packages-doc refresh-action-pins build
+.PHONY: all test stdlib agents doc packages-doc refresh-action-pins build clean publish compile-agency
+
+# Resolve tsc/tsc-alias from the workspace, not from whatever happens to
+# be global. Historically `make build` was only ever reached via
+# `pnpm run build`, which put node_modules/.bin on PATH; `all` now invokes
+# it directly ($(MAKE) build), so the Makefile must point at the tools
+# itself. Explicit paths (not a PATH export) because Apple's GNU make 3.81
+# does not reliably apply an exported PATH to its exec fast path. Also
+# pins tool versions to the repo lockfile.
+TSC := ./node_modules/.bin/tsc
+TSC_ALIAS := ./node_modules/.bin/tsc-alias
+
+# The bundled agent dirs. Adding an agent = add it here, nowhere else.
+AGENT_DIRS := review policy agency-agent eval optimize
+AGENT_PATHS := $(addprefix dist/lib/agents/,$(AGENT_DIRS))
+
+# Stage agent sources into dist. Wipe-then-copy, NOT overlay: with dist
+# persisting across builds, an overlay would keep outputs of agent sources
+# that were deleted from lib/agents. Safe to wipe: lib/agents contains no
+# .ts, so tsc emits nothing under dist/lib/agents — everything there is
+# agency-compiled from the sources this recipe just copied.
+define stage-agent-sources
+	rm -rf dist/lib/agents
+	cp -r lib/agents dist/lib
+endef
+
+# Stage the doc markdown the agents read at runtime. rm before cp -r:
+# copying onto an existing dir NESTS (guide/guide) — that artifact exists
+# in pre-Stage-C tarballs; these copies must be idempotent.
+define stage-agent-docs
+	mkdir -p dist/lib/agents/docs
+	rm -rf dist/lib/agents/docs/guide dist/lib/agents/docs/cli
+	cp -r docs/site/guide dist/lib/agents/docs/guide
+	cp -r docs/site/cli dist/lib/agents/docs/cli
+endef
 
 all:
-	pnpm run templates && pnpm run build && $(MAKE) stdlib && $(MAKE) agents && $(MAKE) doc
+	pnpm run templates && $(MAKE) build && $(MAKE) compile-agency && $(MAKE) doc
 
 # Build the dist/ tree. Invoked by `pnpm run build`. Kept here (rather
 # than as a chain in package.json) so the steps remain readable and easy
 # to extend without inflating the package.json scripts section.
 build:
-	rm -rf dist/
-	tsc
-	tsc-alias
+	# Incremental: dist/ is NOT wiped here (use `make clean` for a full
+	# reset; publish always goes through clean). Because the buildinfo
+	# lives OUTSIDE dist/ (it must not ship in the tarball), dist and its
+	# incremental state no longer share a lifetime: if dist/ was deleted
+	# by hand but the buildinfo survived, tsc would consider everything
+	# up-to-date, emit nothing, and leave an empty dist. Drop the stale
+	# buildinfo in that case.
+	@if [ ! -d dist/lib ]; then rm -f .agency-build/tsc.tsbuildinfo; fi
+	$(TSC)
+	$(TSC_ALIAS)
 	mkdir -p dist/lib/cli/runShim
 	cp lib/cli/runShim/*.mjs dist/lib/cli/runShim/
 	mkdir -p dist/lib/stdlib/providers
 	cp lib/stdlib/providers/*.mjs dist/lib/stdlib/providers/
-	mkdir -p dist/lib/agents/docs
-	cp -r docs/site/guide dist/lib/agents/docs/guide
-	cp -r docs/site/cli dist/lib/agents/docs/cli
 
+# Full reset: compiled TS, incremental scratch state, and the gitignored
+# stdlib .js siblings. Both dist/ AND stdlib/**/*.js ship in the npm
+# tarball, so publish MUST come through here to guarantee no orphaned
+# output (from a deleted/renamed source) can ship. Safe: stdlib contains
+# no .ts and no tracked/.mjs .js — every stdlib .js is agency-generated.
+clean:
+	rm -rf dist/ .agency-build/
+	find stdlib -name '*.js' -type f -delete
+
+# ONE compiler process for stdlib + all agent dirs: node boot and the
+# compiler import are paid once, and the process-wide parse cache means
+# the stdlib prelude (imported by every one of the 94 files) is parsed
+# once instead of once per invocation. This is the load-bearing lever for
+# the Stage C warm-make target.
+compile-agency:
+	$(stage-agent-sources)
+	node ./dist/scripts/agency.js compile stdlib/ $(AGENT_PATHS)
+	$(stage-agent-docs)
+
+# Selective-use aliases (same recipes, one root set each).
 stdlib:
-	pnpm run agency compile stdlib/
+	node ./dist/scripts/agency.js compile stdlib/
 
 agents:
-	cp -r lib/agents dist/lib && \
-	pnpm run agency compile dist/lib/agents/review/ && \
-	pnpm run agency compile dist/lib/agents/policy/ && \
-	pnpm run agency compile dist/lib/agents/agency-agent/ && \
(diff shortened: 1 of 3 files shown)


packages/agency-lang/makefile
```diff
@@ -1,49 +1,106 @@
...
-	tsc
-	tsc-alias
+	# Incremental: dist/ is NOT wiped here (use `make clean` for a full
+	# reset; publish always goes through clean). Because the buildinfo
+	# lives OUTSIDE dist/ (it must not ship in the tarball), dist and its
+	# incremental state no longer share a lifetime: if dist/ was deleted
+	# by hand but the buildinfo survived, tsc would consider everything
+	# up-to-date, emit nothing, and leave an empty dist. Drop the stale
+	# buildinfo in that case.
+	@if [ ! -d dist/lib ]; then rm -f .agency-build/tsc.tsbuildinfo; fi
```
egonSchiele, line 61: What does this do?

egonSchiele, line 61: Minor: the guard is a heuristic keyed on `dist/lib` as a proxy for "dist meaningfully exists." It correctly covers the documented case (hand-deleted `dist/` with a surviving buildinfo). But a *partial* deletion elsewhere isn't covered — e.g. `rm -rf dist/scripts` while `dist/lib` survives leaves the guard un-fired, tsc considers everything up-to-date and re-emits nothing, and then `compile-agency`/`doc`'s `node ./dist/scripts/agency.js ...` fails against a missing script dir. Edge case (who deletes `dist/scripts` specifically?), so I'd just leave a note rather than broaden it — flagging so the heuristic is a conscious choice.

egonSchiele (commented):
Reviewed the diff and validated it against the working tree (I dry-ran every target from the PR's makefile, checked the canned-recipe expansion, and verified the safety of each destructive command). This is a clean, well-factored implementation and it addresses the anti-pattern feedback from the plan review — the `AGENT_DIRS` variable plus the `stage-agent-sources`/`stage-agent-docs` canned recipes remove the duplication that was smeared across three targets. Nice.

**Verified good:**
- Canned recipes expand correctly (`make -n compile-agency`/`agents`/`build`/`all`/`publish` all render to the right commands; `$(AGENT_PATHS)` expands to all five dirs; `all` orchestrates templates → build → compile-agency → doc; `publish` routes clean → all → publish).
- `rm -rf dist/lib/agents` is safe — `lib/agents` contains zero `.ts`, so tsc emits nothing there; everything under `dist/lib/agents` is agency-compiled from the sources the recipe re-copies.
- `find stdlib -name '*.js' -delete` in `clean` is safe — no `.ts`, no tracked `.js`, no `.mjs` under `stdlib`; every `stdlib` `.js` is agency-generated.
- The stale-buildinfo guard logic is correct, and `.agency-build/` is genuinely ignored (`git check-ignore -v` -> the repo-root `.*` rule).
- Honest about the 6.9s-vs-5-6s target in the body rather than quietly shipping a miss.

**One residual risk worth naming (not a blocker):** the behaviors this PR introduces live entirely on the *warm/incremental* path — the stale-buildinfo guard, the `rm`-before-`cp` nesting idempotency (only bites on the 2nd+ build), and incremental-vs-clean correctness. But CI calls bare `make` on fresh checkouts, so **CI is always cold and never exercises any of them**. They were validated once manually per the Verification section, but there's no regression guard: a future makefile edit could silently reintroduce the `guide/guide` nesting or break the guard and every CI run would still pass. If cheap, a tiny CI smoke — run `make` twice, then assert no `dist/lib/agents/docs/guide/guide` and a non-empty `dist/scripts/agency.js` — would lock in exactly the two things cold CI can't see. Otherwise fine to accept as documented manual-verify.

Inline notes below are minor.

packages/agency-lang/tsconfig.json
```diff
@@ -2,9 +2,9 @@
   "compilerOptions": {
     /* Visit https://aka.ms/tsconfig to read more about this file */
     /* Projects */
-    // "incremental": true,                              /* Save .tsbuildinfo files to allow for incremental compilation of projects. */
+    "incremental": true /* Speeds up rebuilds; buildinfo lives in .agency-build/ (gitignored, wiped by `make clean`) so it can never ship in the npm tarball, which packs ./dist wholesale. */,
     // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
-    // "tsBuildInfoFile": "./.tsbuildinfo",              /* Specify the path to .tsbuildinfo incremental compilation file. */
+    "tsBuildInfoFile": "./.agency-build/tsc.tsbuildinfo",
```
egonSchiele, line 7: Heads up: `.agency-build/` is ignored *only* via the repo-root `.gitignore`'s broad `.*` catch-all (confirmed: `git check-ignore -v` -> `.gitignore:3:.*`), not an explicit entry — the plan had called for adding `.agency-build/` to `packages/agency-lang/.gitignore`, and that step didn't land. It works today, but (a) a dev grepping the local `.gitignore` for the buildinfo finds nothing, and (b) if that repo-root `.*` rule is ever narrowed, the buildinfo silently becomes trackable and could get committed. A one-line explicit `.agency-build/` in the package `.gitignore` would make the intent discoverable and decouple it from that broad rule. Low priority.

packages/agency-lang/makefile
egonSchiele: This is the recovery for a footgun the incremental build creates: the tsc buildinfo now lives OUTSIDE dist/ (it must not ship in the tarball, which packs ./dist wholesale). That splits their lifetimes — if you hand-delete dist/ but the buildinfo survives, tsc believes everything is up to date, emits nothing, and you get an empty dist. The line detects that state (no dist/lib but a buildinfo may exist) and drops the buildinfo so tsc does a full emit. Also fixed in fdc5e373: these explanation comments were tab-indented inside the recipe, so the shell echoed them into every build — they now live above the target, and the recipe runs quietly.

packages/agency-lang/makefile
```diff
@@ -1,49 +1,106 @@
...
+
+# Stage agent sources into dist. Wipe-then-copy, NOT overlay: with dist
+# persisting across builds, an overlay would keep outputs of agent sources
+# that were deleted from lib/agents. Safe to wipe: lib/agents contains no
+# .ts, so tsc emits nothing under dist/lib/agents — everything there is
+# agency-compiled from the sources this recipe just copied.
+define stage-agent-sources
+	rm -rf dist/lib/agents
+	cp -r lib/agents dist/lib
+endef
```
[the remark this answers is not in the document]
egonSchiele: Fixed in fdc5e373 — stage-agent-sources now begins with mkdir -p dist/lib. Note standalone make agents right after make clean still cannot succeed end-to-end (it needs the compiled compiler at dist/scripts/agency.js), but the copy no longer fails confusingly against a missing parent.

packages/agency-lang/makefile
```diff
@@ -1,49 +1,106 @@
...
+	# by hand but the buildinfo survived, tsc would consider everything
+	# up-to-date, emit nothing, and leave an empty dist. Drop the stale
+	# buildinfo in that case.
+	@if [ ! -d dist/lib ]; then rm -f .agency-build/tsc.tsbuildinfo; fi
+	$(TSC)
+	$(TSC_ALIAS)
 	mkdir -p dist/lib/cli/runShim
 	cp lib/cli/runShim/*.mjs dist/lib/cli/runShim/
 	mkdir -p dist/lib/stdlib/providers
 	cp lib/stdlib/providers/*.mjs dist/lib/stdlib/providers/
```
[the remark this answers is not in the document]
egonSchiele: Fixed in fdc5e373, with one deliberate narrowing: the wipe targets *.mjs only, not the whole dirs — dist/lib/cli/runShim and dist/lib/stdlib/providers also hold tsc-emitted .js/.d.ts (e.g. resolver.test.js, llamaModelConfig.js), and incremental tsc would NOT re-emit those if we deleted them behind its back. Verified: a planted stale .mjs in dist is removed by the next make build.

packages/agency-lang/makefile
egonSchiele: Agreed on leaving it a heuristic. fdc5e373 documents exactly that in the comment block above the target: dist/lib is a proxy for dist-meaningfully-exists, partial surgical deletions elsewhere are not covered, and make clean is the recovery for any damaged dist.

packages/agency-lang/tsconfig.json
egonSchiele: Good call, with one twist discovered implementing it: packages/agency-lang/.gitignore cannot carry the entry — that file is ITSELF ignored by the root .* rule (git check-ignore confirms), so committing it would need a force-add, and your local copy holds personal entries that would then become tracked. fdc5e373 adds the explicit .agency-build/ line to the tracked root .gitignore instead, right under the .* rule with a comment, so the intent is greppable and survives any future narrowing of the dot-rule.

PR #464, opened 2026-07-09
Config cleanup: remove dead options, add per-run + agent debug flags

egonSchiele:
## Summary

Config rethink & cleanup. Removes five obsolete `AgencyConfig` options and adds per-run / built-in-agent flags for execution tracing and statelog. Net **+573 / −891** across 20 files.

Spec: `docs/superpowers/specs/2026-07-08-config-cleanup-design.md` · Plan: `docs/superpowers/plans/2026-07-08-config-cleanup.md` (both in-tree on the parent branch).

## What changed

**1 — Delete dead options.** `excludeNodeTypes`, `excludeBuiltinFunctions`, `allowedFetchDomains`, `disallowedFetchDomains` were compile-time, string-literal-only, trivially bypassable, and fully superseded by the runtime effect/policy/interrupt system. Also `tarsecTraceHost`, whose value was never read (the host was hardcoded) — removing it drops the parse-cache bypass that existed solely for it, so the cache is now unconditional. The schema is `.passthrough()`, so stale keys in a user's `agency.json` load without error.

**2 — Per-run flags on `run`/`compile`.** A single exported `applyCliFlagsToConfig` helper owns the flag→config mapping (previously inline per command). Adds `--log-file` / `--observability` / `--strict` to `agency run` (and the hidden default command), `--strict` to `agency compile`. `--strict` sets `typechecker.strict` (+`strictTypes`) — the compile-path gate never runs the checker on `strictTypes` alone.

**3 — Trace + statelog for the built-in agents.** The precompiled agents can't be reconfigured at compile time, so `DEBUG_ENV_VARS` + `applyEnvOverridesToContextArgs` read `AGENCY_TRACE_FILE` / `AGENCY_TRACE_DIR` / `AGENCY_LOG_FILE` in the `RuntimeContext` constructor (mirroring the existing coverage env pattern; statelog half delegates to the existing override merge). `runBundledAgent` translates `--trace` / `--log-file` to those env vars using the same token rules as `std::args` (`--` terminator, `--flag=value`, single-dash detection). Flags declared on `agency agent` and forwarded through `agency doctor`.

**4 — `agency config show`** — prints the resolved, merged config as JSON (the `tsc --showConfig` equivalent).

Docs updated (`docs/misc/config.md`, `docs/dev/config.md`); Whitelist/Blacklist wording dropped from help output.

## Testing

- tsc clean; **176 unit/integration tests pass** across the touched files.
- New coverage includes the three wiring points: a `RuntimeContext`-constructed-with-env test, a mocked-`spawn` env assertion, and a writer→reader contract test tying `agentDebugFlagsToEnv` to `applyEnvOverridesToContextArgs` via the shared `DEBUG_ENV_VARS`.
- Real end-to-end smoke: `agency agent --print --trace … --log-file … "say hi"` produced a 2.9 MB trace + 41 KB statelog.

## Notable decisions

- **Doctor in scope:** `doctor()` builds a fixed args array and didn't forward user flags, so it gets its own `--trace`/`--log-file` (appended before the `--` terminator).
- **Nested-process trace semantics (documented, not yet enforced):** `AGENCY_LOG_FILE` is intentionally shared across a spawned agent tree (one statelog); bare `--trace` (per-runId dir mode) is the safe form under nesting; a fixed `--trace <file>` shared across concurrent nested runs can interleave — a consume-and-clear refinement at the spawn boundary is noted as a follow-up.
- **Breaking change** (acceptable now per owner): removed config keys are silently ignored, not errored.

Deferred (owner): `settings.json` layering and config versioning.


egonSchiele (commented):
## Review

**Overview:** Removes five dead `AgencyConfig` options (the four compile-time "security" filters + `tarsecTraceHost` with its parse-cache bypass), centralizes the per-run flag→config mapping in `applyCliFlagsToConfig`, adds `--log-file`/`--observability`/`--strict` to `run` (+ hidden default) and `--strict` to `compile`, wires trace/statelog for the precompiled agents via `DEBUG_ENV_VARS` env overrides read at `RuntimeContext` construction, forwards the flags through `doctor`, and adds `agency config show`.

**Strengths worth calling out:**
- The parse-cache replacement test actually discriminates old vs new code (stale `tarsecTraceHost` key: old bypass = 0 hits/0 misses, new = 1/1) instead of locking behavior that was already true.
- All three wiring points are covered: `RuntimeContext`-constructed-with-env, mocked-`spawn` env assertion, and the writer→reader contract test — plus `DEBUG_ENV_VARS` as a shared constant so the env names can't drift.
- `agentDebugFlagsToEnv` matches `std::args` token rules (`--` terminator, `=` form, single-dash detection) and cites the exact lines it mirrors.
- `--strict` correctly sets both `strict` and `strictTypes` (the compile-path gate never runs on `strictTypes` alone), with the reason documented at the helper.
- `buildDoctorArgs` extracted pure with the `--`-terminator ordering tested.
- Docs and help output updated for the deletions; the `.passthrough()` claim is pinned by a test that includes a wrong-typed stale key.

**Findings (inline comments, most severe first):**
1. `--trace=` (empty attached value) silently produces no trace, while `std::args` treats it identically to bare `--trace` (which produces one) — divergence in the one place the helper exists to prevent divergence. (`runBundledAgent.ts`)
2. "Env wins over baked args" doesn't hold cross-field: `AGENCY_TRACE_DIR` loses to a baked `traceConfig.traceFile` because `resolveTraceFilePath` prefers `traceFile`. Unreachable for the bundled agents (built with default config) but reachable for `agency run`-compiled programs. (`configOverrides.ts`)
3. The two new integration tests hard-depend on `dist/scripts/agency.js` existing — a fresh checkout running the unit suite now fails with a confusing `MODULE_NOT_FOUND` rejection (and the `--strict` assertion half-passes for the wrong reason). (`scripts/agency.test.ts`)
4. `agency config show` prints `log.apiKey` / nested statelog `apiKey` values verbatim — this output is exactly what users paste into bug reports; consider redacting key-like fields. (`scripts/agency.ts`)
5. Nit: `applyCliFlagsToConfig` with `trace: true` and no `input` yields `traceFile: ""`. (`scripts/agency.ts`)

**Not blocking, for follow-up:** the new user-facing surfaces (`--log-file`, `--observability`, `--strict`, agent/doctor `--trace`/`--log-file`, `config show`) exist only in `--help` — `docs/misc/config.md`/`docs/dev/config.md` were updated for deletions only. And the consume-and-clear refinement for nested-process `AGENCY_TRACE_FILE` interleaving is sensibly deferred but worth tracking as an issue so it doesn't evaporate.

Verdict: solid execution — findings are small and none undermine the design. 1 and 2 are the only behavior bugs, both edge-case-sized.

packages/agency-lang/lib/cli/runBundledAgent.ts
```diff
@@ -1,16 +1,69 @@
...
+
+    const eq = token.indexOf("=");
+    const name = eq === -1 ? token.slice(2) : token.slice(2, eq);
+    const attached = eq === -1 ? undefined : token.slice(eq + 1);
+    const next = args[i + 1];
+    const nextIsValue = next !== undefined && next !== "--" && !next.startsWith("-");
+
+    if (name === "trace") {
+      if (attached !== undefined) {
+        env[DEBUG_ENV_VARS.traceFile] = attached;
```
egonSchiele: **`--trace=` diverges from bare `--trace` — the one divergence class this helper exists to prevent.**

`std::args` normalizes bare `--trace` to `--trace=` before parsing (`lib/stdlib/args.ts:503` pushes `--${name}=`), so to the agent's parser the two are *identical* (both yield `""`). Here they behave differently: bare `--trace` → `AGENCY_TRACE_DIR="."` (trace produced), but `--trace=` → `attached === ""` → `AGENCY_TRACE_FILE=""`, which `applyEnvOverridesToContextArgs` drops as falsy (`configOverrides.ts:117`) → **no trace, silently** — the worst failure mode for a debugging flag.

Suggested fix: treat an empty attached value as bare:
```ts
if (attached !== undefined && attached !== "") {
  env[DEBUG_ENV_VARS.traceFile] = attached;
} else if (attached === undefined && nextIsValue) {
  ...
} else {
  env[DEBUG_ENV_VARS.traceDir] = ".";
}
```
(`--log-file=` is fine as-is: both sides no-op consistently since the flag isn't `optional`.)

While here, micro-nit: once `token === "--"` you can `break` instead of setting a flag and `continue`-scanning the rest.

packages/agency-lang/lib/runtime/configOverrides.ts
```diff
@@ -84,3 +84,54 @@ export function applyRuntimeConfigOverridesToContextArgs(
...
+      })
+    : args;
+
+  if (traceFile || traceDir) {
+    merged = {
+      ...merged,
+      traceConfig: {
+        ...(merged.traceConfig ?? {}),
+        ...(traceFile ? { traceFile } : {}),
+        ...(traceDir ? { traceDir } : {}),
```
egonSchiele: **"Env wins over baked args" doesn't hold across fields: `AGENCY_TRACE_DIR` loses to a baked `traceFile`.**

This merge keeps a baked `traceConfig.traceFile` when only `traceDir` comes from env, and `resolveTraceFilePath` (`traceWriter.ts`) prefers `traceFile` over `traceDir`. So for a program compiled with `traceFile` in its `agency.json`, `AGENCY_TRACE_DIR=/x` (or bare `--trace`) is silently ignored and the run keeps writing to the baked fixed file — the opposite of the per-run override the user asked for.

Not reachable for the bundled agents (built with default config, no baked `traceFile`), but the docstring sells this as generic ("also makes `agency run`-produced binaries and any packed bundle traceable via env"), and that's exactly the path that can have a baked `traceFile`.

Suggested fix: when env supplies `traceDir` without `traceFile`, drop the baked `traceFile` so the env dir actually takes effect:
```ts
const traceOverride = traceFile
  ? { traceFile }
  : { traceFile: undefined, traceDir };
```
(plus a one-case test: baked `traceFile` + env `traceDir` → per-runId file in the dir).

packages/agency-lang/scripts/agency.test.ts
```diff
@@ -92,3 +96,66 @@ describe("agency CLI command tree", () => {
...
+    expect(input).toEqual({});
+  });
+});
+
+describe("compile --strict (integration)", () => {
+  it("exits non-zero on a type error with --strict, zero without", async () => {
+    const dir = fs.mkdtempSync(path.join(os.tmpdir(), "strict-"));
+    const f = path.join(dir, "bad.agency");
+    fs.writeFileSync(f, 'node main() {\n  let x: number = "hello"\n}\n');
+    const cli = path.resolve("dist/scripts/agency.js");
```
egonSchiele: **These two integration tests hard-depend on a prior build (`dist/scripts/agency.js`).**

On a fresh checkout (or after `make clean`), `pnpm test:run scripts/agency.test.ts` now fails with a `MODULE_NOT_FOUND` rejection rather than anything indicating "build first". Worse, the failure is half-camouflaged: the `--strict` assertion on line 141 expects a rejection, so a missing CLI makes that half *pass* for the wrong reason while line 140 fails.

The subprocess approach itself is right (`compile` calls `process.exit(1)` in-process, which would kill the vitest worker — this is the correct way to observe the exit code). Suggest a fail-fast guard so the unmet precondition is legible:
```ts
const cli = path.resolve("dist/scripts/agency.js");
if (!fs.existsSync(cli)) {
  throw new Error(`${cli} not built — run \`make\` before the CLI integration tests`);
}
```
and a note in the test names (e.g. `compile --strict (integration, requires build)`) or a check that CI's test job orders after the build job.

packages/agency-lang/scripts/agency.ts
```diff
@@ -913,9 +958,31 @@ export function createProgram(deps: CliDependencies = {}): Command {
...
+
+  const configCmd = program
+    .command("config")
+    .description("Inspect Agency configuration");
+
+  configCmd
+    .command("show")
+    .description("Print the resolved, merged agency.json config as JSON")
+    .action(() => {
+      console.log(JSON.stringify(getConfig(), null, 2));
```
egonSchiele: **`config show` prints secrets verbatim.**

`getConfig()` can contain `log.apiKey` and `client.statelog.apiKey` straight from the user's `agency.json`, and this command's whole purpose is producing output people paste into issues and bug reports ("what config am I actually running?"). `tsc --showConfig` never had this problem because tsconfig has no secrets.

Suggest masking key-like values before printing — e.g. a small replacer that maps any `apiKey` field to `"•••" + last4` — with the raw value available only behind an explicit `--show-secrets` if ever needed. Cheap now, painful to retrofit after the output shape is relied on.

packages/agency-lang/scripts/agency.ts
```diff
@@ -78,7 +78,46 @@ import { startMcpServer } from "@/mcp/server.js";
...
+  options: RunOptions,
+  input?: string,
+): AgencyConfig {
+  const next: AgencyConfig = { ...config };
+  if (options.trace) {
+    next.trace = true;
+    next.traceFile =
+      typeof options.trace === "string"
+        ? options.trace
+        : (input ?? "").replace(/\.agency$/, ".trace");
```
egonSchiele: Nit: with `trace: true` and no `input`, this yields `traceFile: ""` — currently unreachable from the CLI (run/default always pass `input`; compile never passes `trace`), but the helper is exported, so the odd state is one call away. Either require `input` when `trace` is set (overload or runtime assert) or skip setting `traceFile` when there's no input to derive it from.

packages/agency-lang/lib/cli/doctor.ts
```diff
@@ -30,22 +30,34 @@ export function buildDoctorPrompt(file: string, symptom?: string): string {
...
+// flags go BEFORE the `--` terminator (tokens after `--` are positionals to
+// the agent's std::args parser); the prompt is the sole positional after it.
+export function buildDoctorArgs(opts: {
+  file: string;
+  symptom?: string;
+  trace?: string | true;
+  logFile?: string;
+}): string[] {
+  const prompt = buildDoctorPrompt(opts.file, opts.symptom);
+  const debug: string[] = [];
```
egonSchiele, line 44: If you're checking for these flags here, does that mean that every time we add a new agency agent, it will have to, again, write code to check for these flags? It would be great to handle these flags in a single place and make them available to every bundled agent, if possible. If not, we should at least make a pre-process function that every built-in agent calls and put all this logic there and share it across all the bundled agents.

packages/agency-lang/lib/cli/runBundledAgent.ts
```diff
@@ -1,16 +1,69 @@
...
+ * standalone `--`; accept `--flag=value`; treat the next token as a value only
+ * if it is defined, not `--`, and does not start with `-`. Pure + exported.
+ * The env var — not the flag — does the wiring (RuntimeContext is built before
+ * the agent's main() parses flags), so the agent also DECLARES these flags for
+ * --help; both sides sharing DEBUG_ENV_VARS keeps them in sync.
+ */
+export function agentDebugFlagsToEnv(args: string[]): Record<string, string> {
+  const env: Record<string, string> = {};
+  let afterDoubleDash = false;
+  for (let i = 0; i < args.length; i++) {
```
egonSchiele: This code is unreadable.

egonSchiele:
Reworked this to address the scattered/imperative config logic (commit 1a05a206).

**Single source of truth — `lib/config.ts`.** A documented "Config resolution" section now declares the three config sources (agency.json → CLI flags → `AGENCY_CONFIG_OVERRIDES` env), their precedence, and where each is applied. Concretely:
- `applyCliFlags(config, flags, input?)` — the one definition of what each debug flag *means* in config terms, used by `agency run`/`compile` **and** the bundled agents.
- `CONFIG_OVERRIDES_ENV` + `readConfigOverrides` / `serializeConfigOverrides` — **one** JSON env var carrying a `Partial<AgencyConfig>`, replacing the three bespoke `AGENCY_TRACE_*`/`AGENCY_LOG_FILE` vars. It's the env-transport twin of the subprocess IPC `configOverrides` message that already existed.

**One runtime merge.** `applyRuntimeConfigOverridesToContextArgs` (already used for subprocess IPC) is extended to handle trace and now serves both transports; the `RuntimeContext` ctor uses IPC overrides when present, else the env. Deleted the parallel `applyEnvOverridesToContextArgs` + `DEBUG_ENV_VARS`.

**One reusable bundled-agent runner (re: doctor.ts / runBundledAgent.ts).** `runBundledAgent` now extracts the debug flags **once** — via `node:util.parseArgs`, not the hand-rolled `token.indexOf("=")` scanner — into a config override and serializes it. Every bundled agent gets `--trace`/`--log-file` for free; no agent writes flag code. The unreadable parser is gone.

**Bug fixes from the review:**
- `--trace=` now behaves like bare `--trace` (per-run file) instead of `traceFile="" → silent no-op` — verified end-to-end.
- An override `traceDir` clears a baked `traceFile`, so the per-run dir actually takes effect.
- `agency config show` masks `apiKey` fields by default; `--show-secrets` to reveal (also covers the Copilot comment).
- CLI integration tests fail fast with a legible "run `make` first" message instead of `MODULE_NOT_FOUND`, and no longer half-pass for the wrong reason.
- `applyCliFlags` no longer yields `traceFile=""` for bare `--trace` with no input.

tsc clean; 179 unit/integration tests pass.

packages/agency-lang/lib/cli/runBundledAgent.ts
```diff
@@ -1,16 +1,55 @@
...
+ *
+ * Parsing is delegated to `node:util.parseArgs` (the standard library parser
+ * `std::args` itself is built on), so `--flag=value`, the `--` terminator, and
+ * last-wins-on-repeat all fall out correctly. `strict: false` ignores the
+ * agent's own flags. A bare or empty `--trace` maps to a per-run trace file in
+ * cwd; the flag→config meaning lives in `applyCliFlags` (config.ts), shared
+ * with `agency run`/`compile`.
+ */
+export function agentConfigOverride(args: string[]): Partial<AgencyConfig> {
+  const { values } = parseArgs({
```
egonSchiele, line 60: **Regression: non-strict `parseArgs` consumes a following flag as the `--trace` value — the divergence the previous commit fixed, reintroduced.**

Verified on Node 20: `["--trace", "--print", "hi"]` → `{ trace: "--print" }` and `["--trace", "-p"]` → `{ trace: "-p" }`. The agent's own parser treats both as *bare* `--trace` because `std::args` pre-normalizes a bare optional flag to `--flag=` when the next token starts with `-` (`lib/stdlib/args.ts:499-505`) — a workaround for exactly this `parseArgs` behavior. Raw `parseArgs` without that pass diverges: `agency agent --trace --print "hi"` runs one-shot with a bare trace on the agent side, while the child gets `traceFile: "--print"` and writes the trace to a file named `./--print`.

The docstring's rationale ("the standard library parser `std::args` itself is built on, so … all fall out correctly") is therefore not quite right — `=`-form, `--`, and last-wins do fall out, but dash-next-token does not.

Suggested fix — replicate the same normalization before parsing (this is the whole remaining delta between raw `parseArgs` and `std::args` tokenization for these two flags):
```ts
const normalized = args.map((tok, i) => {
  if (tok !== "--trace") return tok;
  const next = args[i + 1];
  const bare = next === undefined || next.startsWith("-");
  return bare ? "--trace=" : tok;
});
// stop mapping at the first "--" (or slice the array there first)
```
or use `tokens: true` and take only inline/attached values for `trace`.

And please restore the two deleted guard cases in `runBundledAgent.test.ts`:
```ts
expect(agentConfigOverride(["--trace", "-p"])).toEqual({ trace: true, traceDir: "." });
expect(agentConfigOverride(["--trace", "--print", "hi"])).toEqual({ trace: true, traceDir: "." });
```
Their removal is what let this regress silently — the first review's test suite had the `-p` case.

packages/agency-lang/lib/runtime/state/context.ts
```diff
@@ -205,7 +209,13 @@ export class RuntimeContext<T> {
      *  match the established no-debug-by-default behavior. */
     logLevel?: LogLevel;
   }) {
-    args = applyRuntimeConfigOverridesToContextArgs(args);
+    // One runtime merge, two transports: subprocess overrides (set via IPC)
+    // win; otherwise fall back to AGENCY_CONFIG_OVERRIDES from the environment
+    // (how the precompiled bundled agents receive per-run trace/statelog).
+    args = applyRuntimeConfigOverridesToContextArgs(
+      args,
+      getRuntimeConfigOverrides() ?? readConfigOverrides(),
```
egonSchiele: **`??` makes the two transports mutually exclusive — an IPC override silently masks the env override for that child.**

Any subprocess launched with explicit `configOverrides` — eval runs do this unconditionally (`lib/cli/eval/run.ts:308`), and user programs can via `_run(..., configOverrides: ...)` — gets `getRuntimeConfigOverrides() !== undefined`, so `readConfigOverrides()` is never consulted in that child. Net effect: `agency agent --log-file tree.jsonl` captures the agent and plain children, but any child spawned with explicit IPC overrides drops out of the statelog — a partial regression of the "one statelog across the spawned tree" behavior the PR description commits to (the pre-refactor code applied IPC overrides *and then* env on top).

Since `applyRuntimeConfigOverridesToContextArgs` is idempotent-shaped and cheap, the old layering is easy to keep with the new transport:
```ts
args = applyRuntimeConfigOverridesToContextArgs(args, readConfigOverrides());
args = applyRuntimeConfigOverridesToContextArgs(args, getRuntimeConfigOverrides());
```
(sequential application also avoids the shallow-spread problem a single `{...env, ...ipc}` merge would have: both sides carrying `log` objects would clobber each other's keys instead of layering). If IPC-masks-env is actually the intended semantics, it contradicts the PR description and deserves a comment here + a line in the config-resolution note in `config.ts` saying source 3 is *conditional*.

packages/agency-lang/lib/config.ts
```diff
@@ -586,3 +548,135 @@ export function findProjectRoot(startPath: string): string | null {
...
+ *  brick startup. */
+export function readConfigOverrides(
+  env: NodeJS.ProcessEnv = process.env,
+): Partial<AgencyConfig> {
+  const raw = env[CONFIG_OVERRIDES_ENV];
+  if (!raw) return {};
+  try {
+    const result = AgencyConfigSchema.safeParse(JSON.parse(raw));
+    return result.success ? (result.data as Partial<AgencyConfig>) : {};
+  } catch {
```
egonSchiele: **Silent catch — a malformed `AGENCY_CONFIG_OVERRIDES` makes the debug flags silently do nothing.**

`docs/dev/anti-patterns.md` calls out exactly this shape ("try-catch block without logging anything in the catch block"), and the failure mode here is the one this PR fixed twice already in other spots: someone sets the env var by hand (the config-resolution comment above explicitly advertises it for packed bundles), typos the JSON or a field type, and their `--trace`/`--log-file` just… doesn't happen, with zero signal. Same for the schema-failure branch on line 654, which discards `result.error` without a trace.

Not bricking startup is right — but say something:
```ts
if (!result.success) {
  console.error(`Ignoring invalid ${CONFIG_OVERRIDES_ENV}: ${result.error.issues.map(i => i.path.join(".")).join(", ")}`);
  return {};
}
...
} catch (e) {
  console.error(`Ignoring unparseable ${CONFIG_OVERRIDES_ENV}: ${e instanceof Error ? e.message : String(e)}`);
  return {};
}
```
Worth a test case each (invalid JSON, schema-violating field) asserting `{}` is returned — the round-trip test only covers the happy path.

egonSchiele (commented):
## Re-review (after 1a05a206)

All five findings from the first review are genuinely fixed, each with a test: `--trace=` now behaves like bare `--trace`; an override `traceDir` clears a baked `traceFile`; the integration tests fail fast with a legible message when `dist/` isn't built; `config show` masks `apiKey` fields by default behind `--show-secrets` (with a masked-vs-raw round-trip test); and the bare-`--trace`-without-input case now maps to `traceDir: "."`. The consolidation is also a real improvement: the config-resolution comment block in `config.ts` declaring the three sources and their precedence, `applyCliFlags` as the one flag→config definition shared by `run`/`compile`/bundled agents, one runtime merge serving both transports, and the helper tests relocated next to the helpers.

However, the refactor from the hand-rolled scanner to raw non-strict `util.parseArgs` **reintroduces the dash-token divergence the previous commit fixed — and deletes the test that guarded it.** Verified empirically (Node 20):

```
parseArgs({args: ["--trace", "--print", "hi"], options: {trace: {type: "string"}}, strict: false})
→ { trace: "--print" }
parseArgs({args: ["--trace", "-p"], ...})
→ { trace: "-p" }
```

Non-strict `parseArgs` greedily consumes the next token as a string option's value **even when it's another flag**. `std::args` avoids this only because it pre-normalizes bare optional flags to `--flag=` before calling `parseArgs` (`lib/stdlib/args.ts:499-505` — that normalization exists precisely because of this parseArgs behavior). So the docstring's premise — correctness "falls out" of using the parser `std::args` is built on — doesn't hold: `std::args` is that parser *plus* the normalization this refactor dropped.

Concretely: `agency agent --trace --print "hi"` → the agent parses bare `--trace` + `--print` (one-shot mode), but the child is configured with `traceFile: "--print"` — the trace lands in a file literally named `./--print`. The first review's test for exactly this (`--trace followed by a single-dash token is bare`) was removed in this commit rather than ported.

Details and two smaller issues (silent catch, transport exclusivity) inline. Everything else in the commit checks out.

packages/agency-lang/lib/cli/runBundledAgent.ts
egonSchiele: Fixed in dda6c760. You're right — I removed the guard test and the divergence came back. `agentConfigOverride` now normalizes a bare debug flag whose next token starts with `-` to `--flag=` before parseArgs (the same thing std::args does at args.ts:499-505), so `--trace --print` and `--trace -p` are bare, not a trace file named `--print`/`-p`. Restored both guard cases plus bare `--log-file` cases, and verified end-to-end: `agency agent --trace --print "say hi"` writes a per-run `<runId>.agencytrace` and no `./--print`.

packages/agency-lang/lib/runtime/state/context.ts
egonSchiele: Fixed in dda6c760. Switched to your layering — env first, then IPC on top — so a subprocess with explicit IPC overrides still inherits the tree-wide env `--log-file`; IPC wins per field. Added a regression test (env logFile survives when an IPC override is also set).

packages/agency-lang/lib/config.ts
egonSchiele: Fixed in dda6c760. `readConfigOverrides` now `console.error`s on both the schema-failure and JSON-parse-failure branches (still returns `{}` so startup never bricks). Added a test for each (unparseable JSON, schema-violating field) asserting `{}` + the warning.

packages/agency-lang/lib/runtime/configOverrides.ts
```diff
@@ -33,23 +33,23 @@ export function getRuntimeConfigOverrides(): Partial<AgencyConfig> | undefined {
...
- * (absolutized) provider-module paths here so a subprocess loads the same
- * custom/local providers the parent has, regardless of how the child was
- * compiled. They are merged with any baked-in `providerModules` on `args`;
- * `loadProviderModules` de-duplicates by resolved path at load time.
- *
- * `maxCallDepth` is honored so a subprocess inherits the parent's runaway-
- * recursion ceiling rather than falling back to the default.
+ * Fields honored (others are ignored — the runtime has its own pathways):
+ *   • `log.*` + `observability` → statelogConfig
+ *   • `trace` / `traceFile` / `traceDir` → traceConfig. An override always
```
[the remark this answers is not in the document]
egonSchiele: Fixed in dda6c760 — updated the docstring to say `traceFile`/`traceDir` drive traceConfig and the `trace` boolean is only a marker set alongside them by applyCliFlags, matching the implementation.

packages/agency-lang/scripts/agency.test.ts
```diff
@@ -92,3 +108,38 @@ describe("agency CLI command tree", () => {
       .toBe("function");
   });
 });
+
+describe("compile --strict (integration, requires build)", () => {
+  it("exits non-zero on a type error with --strict, zero without", async () => {
```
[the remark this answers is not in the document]
egonSchiele: Fixed in dda6c760 — these `describe` blocks are now `describe.skipIf(!HAS_BUILT_CLI)` with a visible "run make/build first" warning, so `pnpm test` stays green on a clean checkout while still running in CI and after a build.

packages/agency-lang/scripts/agency.test.ts
```diff
@@ -92,3 +108,38 @@ describe("agency CLI command tree", () => {
...
+    const f = path.join(dir, "bad.agency");
+    fs.writeFileSync(f, 'node main() {\n  let x: number = "hello"\n}\n');
+    await expect(execFileAsync("node", [cli, "compile", f])).resolves.toBeTruthy();
+    await expect(execFileAsync("node", [cli, "compile", "--strict", f])).rejects.toBeTruthy();
+    fs.rmSync(dir, { recursive: true, force: true });
+  });
+});
+
+describe("config show (integration, requires build)", () => {
+  it("prints the resolved, merged config as JSON, with secrets masked by default", async () => {
```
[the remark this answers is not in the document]
egonSchiele: Fixed in dda6c760 — see above: gated with `describe.skipIf(!HAS_BUILT_CLI)`.

egonSchiele (commented):
## Re-review (after dda6c760)

All three second-round findings are fixed, and I verified each rather than trusting the commit message:

**1. Dash-token regression — fixed correctly.** `normalizeBareDebugFlags` replicates the exact `std::args` pre-pass (`lib/stdlib/args.ts:499-505`: bare when next is absent/`--`/starts with `-`, stop at the terminator) before handing off to `parseArgs`. I re-ran the full divergence matrix against the new pipeline on Node 20:

```
["--trace","--print","hi"]        → trace:"" (bare → traceDir ".")   ✓
["--trace","-p"]                  → trace:"" (bare)                  ✓
["--trace"] / ["--trace="]        → trace:"" (bare)                  ✓
["--trace","t.trace"]             → trace:"t.trace"                  ✓
["--log-file"] / [.., "--print"]  → "" → ignored                     ✓
["--","--trace","x"]              → {} (terminator)                  ✓
["--trace","a","--trace"]         → last-wins bare                   ✓
attached = forms                   → literal values                   ✓
```

The deleted guard tests are restored and extended (`--trace --print`, `--trace -p`, bare/flag-followed `--log-file`), so this can't silently regress a second time. The post-normalization invariant (`values.trace` is always a string) is stated in a comment and matches the code.

**2. Transport layering — fixed.** Env is applied first, IPC on top; a child launched with explicit IPC overrides (the eval-run case) no longer drops out of a tree-wide `--log-file`, and the new regression test pins exactly that scenario (IPC `maxCallDepth` + env `logFile`, both surviving). The precedence *flip* relative to the original code (IPC now wins per-field where env used to) is deliberate, argued in the comment (explicit per-child beats ambient), and layered rather than clobbered — sequential application handles the nested `log` objects correctly where a single spread wouldn't. One micro-note, no action needed: passing `getRuntimeConfigOverrides()` explicitly still triggers the parameter default when it's `undefined` — which happens to be the same value, so it's benign.

**3. Silent catch — fixed.** Both failure branches (unparseable JSON, schema-violating field) now warn with the env var name and offending paths, with a test per branch plus a no-warning-when-absent case.

On the `skipIf` change for the CLI integration tests (my suggestion was fail-fast): I checked `.github/workflows/test.yml` — the `build` job runs `make` before `pnpm test:run`, so `dist/` always exists in CI and these tests genuinely run there; the skip only ever fires on a clean local checkout, with a visible warning. That's a reasonable trade and arguably better DX than my suggestion.

Remaining known items are all documented follow-ups, not blockers: consume-and-clear for a fixed `--trace <file>` under nested concurrent runs (noted in the merge docstring), and the nested-process semantics note in the PR description.

**Verdict: this would be an approval** (GitHub blocks approving your own PR, so this lands as a comment). At the time of review, `build (23.x)` and `lint` are green; the 22.x and agency-tests jobs are still pending — worth letting them finish before merge.)

PR #466, opened 2026-07-09
BuildSession: consolidate compile pipeline behind one declarative module (Stage A, PR 1)

egonSchiele:
## What

Stage A, PR 1 of the incremental-builds work (spec now committed at `docs/superpowers/specs/2026-07-08-incremental-builds-design.md`; plan at `docs/superpowers/plans/2026-07-08-buildsession-consolidation.md`, both review-hardened): a **pure consolidation with byte-identical outputs**. `lib/compiler/buildSession.ts` becomes the single owner of compile-pipeline caching and orchestration:

- The `commands.ts` module globals (`compiledFiles`, `currentClosure`), `ensureCompiledClosure`, the directory union-closure branch, and `compileMany` move into a `BuildSession` object (verbatim move; state threaded via a `SessionState` param).
- The cross-config machinery moves in from `precompile.ts`: `CompileGroup`, the single-slot conflict assert, and — new — `configKey` derivation happens INSIDE the session (an integrity mechanism, not caller input). `precompile.ts` shrinks to a test-runner adapter on a fresh session.
- `commands.ts` keeps thin delegates (`compile`/`compileMany`/`resetCompilationCache`) over one lazy `defaultSession`, so every caller (watch mode, runAgencyNode, coverage, tests) is untouched.
- One allowed shape change: a private `setClosure()` pairs every closure replacement with the dedupe-set clear — previously a convention repeated at three sites, structural before PR 2 adds more writers.

PR 2 adds the content-hash manifest and `freshness` policy on this interface.

## Found during execution

- **Tree-shaking trap:** an eager top-level `createBuildSession()` in `commands.ts` dragged the whole codegen subtree (~16k lines) into every `agency pack` artifact — packed programs import `commands.js` for small helpers, and the call was an unshakeable side effect. Caught by `pack.test.ts`; fixed with lazy initialization. Main was safe only because its module globals were inert literals.
- **Pre-existing, not fixed here:** `tests/debugger/loop-test.ts` (a tracked fixture) is stale against current codegen — `lib/debugger/driver.test.ts`'s beforeAll rewrites it to the newer lazy-iterable loop form on pristine main too. It will show as a dirty file after any full local vitest run. Worth a follow-up: regenerate and commit it, or untrack it.

## Verification

- **Byte-identity vs main:** all 94 compiled stdlib + agent `.js` files hash-identical to a baseline built from untouched main before any change; `make fixtures` diff-clean.
- Full vitest: 7481 passed (includes 24 new buildSession/delegate tests: session isolation, dedupe, reset, union-closure emit-counts, directory entries, throw contract, allowTestImports gating, outDir, cross-config assert, zod key-order tripwire).
- Runtime slices: 74 agency execution test files (incl. the local-config memory dir through `compileGroups`), agency-js debugger dir, full agents suite (48s — exercises the fresh-session precompile delta at scale).
- Watch-mode smoke: initial compile + rebuild-after-edit through the `resetCompilationCache` delegate.
- Declared behavior delta (reviewed): precompile uses a fresh session, so the default session no longer inherits its leftover closure — safe because test cases run `preferCompiled`; the old inherited state was a latent staleness hazard.

### DIFF
--- a/packages/agency-lang/lib/cli/commands.ts
+++ b/packages/agency-lang/lib/cli/commands.ts
@@ -1,45 +1,27 @@
 import { generateAgency } from "@/backends/agencyGenerator.js";
 import { AgencyConfig, loadConfigSafe } from "@/config.js";
-import { AgencyProgram, generateTypeScript } from "@/index.js";
-import { initPlanForModule } from "@/backends/typescriptGenerator.js";
-import { resolveImports } from "@/preprocessors/importResolver.js";
-import { resolveReExports } from "@/preprocessors/resolveReExports.js";
-import { liftCallbackBlocks } from "@/preprocessors/liftCallbacks.js";
-import { buildCompilationUnit, CompilationUnit } from "@/compilationUnit.js";
-import { SymbolTable } from "@/symbolTable.js";
-import { formatErrors, typeCheck } from "@/typeChecker/index.js";
-import {
-  buildCompiledClosure,
-  CompileClosureError,
-  type CompiledClosure,
-} from "@/compiler/compileClosure.js";
+import { AgencyProgram } from "@/index.js";
+import { CompileClosureError } from "@/compiler/compileClosure.js";
 import { spawn } from "child_process";
-import { transformSync } from "esbuild";
 import * as fs from "fs";
 import { createRequire } from "module";
 import * as path from "path";
 
-import {
-  getStdlibDir,
-  isNonTemplatedStdlib,
-  isPkgImport,
-  isStdlibImport,
-  resolveAgencyImportPath,
-} from "../importPaths.js";
-import {
-  CompileStrategy,
-  RunStrategy,
-  type ImportStrategy,
-} from "../importStrategy.js";
+import { RunStrategy } from "../importStrategy.js";
 import { parseAgency, replaceBlankLines } from "../parser.js";
-import { parseAgencyFileCached } from "../parseCache.js";
 import { fileURLToPath, pathToFileURL } from "url";
 import {
   classifyInstall,
   installDirFromUrl,
   type InstallKind,
 } from "./installLocation.js";
-import { findRecursively, getImports } from "./util.js";
+import { findRecursively } from "./util.js";
+import {
+  createBuildSession,
+  readFile,
+  type BuildSession,
+  type CompileOptions,
+} from "../compiler/buildSession.js";
 
 // Returns the file:// URL of the ESM loader-register shim shipped with the
 // agency-lang package. Passing this to `node --import=<url>` causes Node to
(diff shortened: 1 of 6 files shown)


packages/agency-lang/lib/cli/commands.ts
```diff
@@ -1,45 +1,27 @@
...
-import { buildCompilationUnit, CompilationUnit } from "@/compilationUnit.js";
-import { SymbolTable } from "@/symbolTable.js";
-import { formatErrors, typeCheck } from "@/typeChecker/index.js";
-import {
-  buildCompiledClosure,
-  CompileClosureError,
-  type CompiledClosure,
-} from "@/compiler/compileClosure.js";
+import { AgencyProgram } from "@/index.js";
+import { CompileClosureError } from "@/compiler/compileClosure.js";
```
egonSchiele: `CompileClosureError` is now an unused import — after the move, the only remaining reference in this file is the mention in `compileMany`'s doc comment (which doesn't need the import). Safe to drop the import line.

packages/agency-lang/lib/compiler/buildSession.ts
```diff
@@ -0,0 +1,541 @@
...
+  for (const { group, closure } of withClosures) {
+    compileManyImpl(state, group.config, group.files, {
+      closure,
+      quiet: options?.quiet,
+      allowTestImports: options?.allowTestImports,
+    });
+  }
+}
+
+export function readFile(inputFile: string): string {
```
egonSchiele: Non-blocking observation, two parts:

1. `readFile` is CLI file-IO, not compile-pipeline logic — its semantic home is closer to `lib/cli/util.ts` than the heaviest module in the compiler. Packed programs reach it via the `commands.js` re-export, so its module's import weight is only kept out of pack artifacts by esbuild successfully shaking this file's codegen imports (which `pack.test.ts` guards — fine today).
2. This PR lengthens a pre-existing import cycle: `commands.ts ↔ cli/util.ts` already existed on main (`util.ts:25` imports `compile`), and now it runs `commands → buildSession → cli/util → commands`. It works because everything crossing the cycle is a hoisted function, but each added edge makes a future top-level-binding load-order bug easier to introduce. Not worth restructuring in this PR — just flagging so PR 2 doesn't add a fourth edge (e.g. the manifest module importing from `cli/`).

packages/agency-lang/lib/compiler/buildSession.test.ts
```diff
@@ -0,0 +1,223 @@
...
+  });
+
+  test("allowTestImports gates import test compilation", () => {
+    const dir = writeTempDir({
+      "lib.agency": 'def secret(): string {\n  return "s"\n}\n',
+      "main.agency":
+        'import test { secret } from "./lib.agency"\n\nnode main() {\n  return secret()\n}\n',
+    });
+    const files = [path.join(dir, "main.agency")];
+    expect(() => createBuildSession().compileMany({}, files, { quiet: true })).toThrow();
```
egonSchiele: The bare `.toThrow()` would also pass on the wrong error — e.g. a typo'd fixture path throwing `CompileClosureError`, which would make the gate test green while the gate itself is broken. The rejection has a stable message (`importResolver.ts`: `` `import test` is only allowed under the test harness. ``), so this can be pinned:

```ts
).toThrow(/only allowed under the test harness/);
```

egonSchiele (commented):
Reviewed the full diff against main with the moved code compared line-by-line. **Verdict: this is a clean consolidation — no correctness issues found.** The verbatim move is faithful (I checked `compileEntry`, `compileManyImpl`, `ensureCompiledClosure`, and `compileDirectory` against the originals; the `setClosure` pairing is behavior-identical at all three former call sites, including the double-clear in `ensureCompiledClosure`, which is a no-op on an already-empty set). Findings are three non-blocking comments inline.

**What is notably right:**
- The byte-identity verification now proves main-equivalence (baseline hashed from untouched main), not just determinism.
- The lazy `getDefaultSession()` with the pack/tree-shaking rationale written at the declaration site — the comment explains a constraint the code cannot show, which is exactly when a comment earns its place. `pack.test.ts` remains the tripwire if a future top-level side effect re-breaks shaking.
- `configKey` derived inside the session and the prebuilt-closure handoff made internal-only: both remove caller-carried integrity contracts from the public type.
- The test suite tests the right properties: emit-count assertions actually distinguish union-closure-once from closure-per-entry; `fs.utimesSync` back-dating makes the mtime assertions granularity-independent; the fresh-session precompile delta is declared in a comment with its safety argument; the reframed zod key-order test is now an end-to-end tripwire on the assert instead of a string comparison.
- I ran `pnpm run lint:structure` on the branch: clean.

**One forward-looking note for PR 2:** `compileEntry` sits just under the 150-line eslint cap. The manifest skip-check cannot be added inline there — plan it as helpers (`shouldSkip(state, manifest, file)` etc.) from the start.

Also noting the committed spec reflects all seven findings from the second design review, including the corrected pkg:: closure claim and the agent-staging decision.

packages/agency-lang/lib/compiler/buildSession.ts
```diff
@@ -0,0 +1,541 @@
...
+// `compiledFiles` entries are only meaningful under the current closure,
+// so replacing the closure MUST clear the set. Previously that pairing was
+// enforced by hand at three separate call sites; this makes it structural
+// (and protects the additional writers PR 2 introduces).
+function setClosure(state: SessionState, closure: CompiledClosure | null): void {
+  state.currentClosure = closure;
+  state.compiledFiles.clear();
+}
+
+export function createBuildSession(): BuildSession {
```
egonSchiele: please create a class instead of an object with functions

egonSchiele: Done in 11a187f7 — BuildSession is now a class with private compiledFiles/currentClosure fields and private pipeline methods (compileEntry/compileDirectory/ensureCompiledClosure/compileManyImpl). createBuildSession() stays as the factory so no caller changed. Stateless helpers (readFile, parseFileOrExit, runTypecheck, deriveConfigKey, findCrossConfigConflicts) remain module functions. Byte-identity vs the main baseline re-verified after the refactor.

packages/agency-lang/lib/cli/commands.ts
egonSchiele: Fixed in 11a187f7. You were right — my unused-import scan counted the doc-comment mention as a use, which is why it survived the cleanup sweep.

packages/agency-lang/lib/compiler/buildSession.ts
egonSchiele: Agreed on both points, leaving as-is in this PR per your note. Recorded as a hard constraint for PR 2 planning: the manifest module must import only downward (buildSession -> manifest, never manifest -> commands or manifest -> cli/util), so the cycle stays at three edges. On readFile: moving it to cli/util would put commands <-> util imports in BOTH directions of the pre-existing 2-cycle, so if we relocate it later, the right move is probably a new leaf module (or fixing the cycle first).

packages/agency-lang/lib/compiler/buildSession.test.ts
egonSchiele: Fixed in 11a187f7 — pinned to /only allowed under the test harness/ (the stable message in importResolver.ts). Good catch: the fixture paths in that test are session-generated temp dirs, so a typo there really would have thrown CompileClosureError and greened the gate test.

packages/agency-lang/lib/cli/precompile.test.ts
```diff
@@ -110,28 +78,32 @@ describe("groupTestSources", () => {
...
+        "agency.json": '{"verbose": false, "observability": true}',
       },
       b: {
-        "main.agency": TRIVIAL,
+        "main.agency": IMPORTS_HELPER,
         "main.test.json": TEST_JSON,
-        "agency.json": '{"observability": false, "verbose": true}',
+        "agency.json": '{"observability": true, "verbose": false}',
       },
     });
```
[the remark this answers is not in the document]
egonSchiele: Checked — the premise does not match this file. precompile.test.ts defines IMPORTS_HELPER as importing `../shared/helper.agency` (lines 9-10), which from a/ and b/ resolves exactly to the shared/helper.agency this tree creates. The `./helper.agency` version is buildSession.test.ts, a different file whose same-named constant pairs with FLAT fixtures (helper in the same dir) — correct there too. The test also proves its own resolution structurally: a missing import makes buildCompiledClosure throw CompileClosureError inside compileGroups, which the .not.toThrow() assertion would catch — and this test passes locally and on both CI legs. No change needed.

PR #468, opened 2026-07-09
Incremental agency builds: content-hash manifest + freshness policy (Stage A, PR 2)

egonSchiele:
## What

Stage A, PR 2 of the incremental-builds work (spec: `docs/superpowers/specs/2026-07-08-incremental-builds-design.md`; plan: `docs/superpowers/plans/2026-07-09-incremental-manifest.md`, both review-hardened): the content-hash **build manifest**. The compiler now skips recompiling `.agency` files whose inputs are unchanged, with correctness carried by conservative invalidation fields that can only over-rebuild:

- `lib/compiler/buildManifest.ts` (leaf module, no new import-cycle edge): manifest types, atomic IO, hashing, the shared `computeDepsHash`, and `isEntryFresh` — freshness of a module requires its source, every recorded transitive dep (source AND compiled output AND manifest entry), `stdlibHash`, `compilerStamp`, `configKey`, and its own output to all check out.
- `lib/compiler/manifestTracker.ts`: freshness policy is interpreted in exactly ONE place — the tracker factory. "always" (test runner, `--ts`, any caller-supplied importStrategy — the run/coverage paths, whose RunStrategy shapes emitted bytes) gets a shared no-op tracker; `--force` gets reads-off/writes-on. Session call sites are unconditional one-liners.
- `BuildSession` public interface collapses to one entries-based `compile(config, { entries, ... })` with the fully-clean fast path in the single dispatch: a clean entry set performs zero parses (proven by parse-cache-stats tests).
- `agency compile --force` is the escape hatch; `make clean` remains the full reset.
- `make` integration: agent staging becomes a sync script (`scripts/stage-agents.mjs`) that stops deleting compiled agent outputs (the old `rm -rf` made agent skips impossible), and `make doc` gets a stamp keyed on stdlibHash + compilerStamp so doc-generator edits regenerate tracked docs.
- Dev doc: `docs/dev/incremental-builds.md` (linked from CLAUDE.md).

## Measured (M-series laptop)

| Scenario | Result |
|---|---|
| warm `make`, no changes | **2.5s** (was 6.9s; 16.8s before Stage C) — all 96 modules skip, zero parses |
| cold (`make clean && make`) | 11.9s |
| touch one agent `.agency` | recompiles exactly 1 file |
| touch any stdlib file | rebuilds the world (96) via `stdlibHash` — required: `resolveReExports` bakes stdlib paths into output |
| edit `lib/runtime/*.ts` | zero agency recompiles (stamp exclusion) |
| real codegen edit | rebuilds all 96 AND regenerates docs |
| whitespace-only compiler edit | zero rebuilds (content-hash stamp sees identical emitted `.js`) |
| `touch stdlib/*.agency` (mtime churn) | zero rebuilds |

**Purity gate**: after the full edit matrix, the incremental tree is hash-identical to a from-scratch `make clean && make` across all 96 compiled files — a skipped module's bytes never depend on which closure it was originally compiled under.

## Verification

Full vitest (7537), lint, fixtures diff-clean, 74-file agency execution slice, agents suite, and the boundary proofs: the test runner leaves no manifest entries for test sources; strategy compiles never skip in either direction; `export { x } from "pkg::…"` modules are never skipped (detection shares the closure walker across all three import-edge node types).

## Found during execution (pre-existing, not fixed here)

The built-in agent settings tests (`lib/agents/agency-agent/tests/settings.agency`) read and write the REAL `~/.agency-agent/settings.json` — every local `test:agents` run clobbers the user's actual settings, and under `-p 12` the shared file races (one such race failed a run of this verification; the suite passes in isolation and on rerun). Needs a sandboxed settings path — issue to follow.

### DIFF
--- a/packages/agency-lang/lib/compiler/buildManifest.ts
+++ b/packages/agency-lang/lib/compiler/buildManifest.ts
@@ -1,0 +1,224 @@
+/**
+ * The incremental-build manifest: content-hash records that let a
+ * BuildSession skip recompiling unchanged modules WITHOUT parsing anything.
+ *
+ * Leaf module by design (PR #466 review): imports only node built-ins and
+ * config.js. It must never import from lib/cli/ — the commands →
+ * buildSession → cli/util → commands cycle stays at three edges. That
+ * leaf-ness is also why walkFiles below deliberately duplicates the
+ * spirit of lib/cli/util.ts findRecursively: importing it would add the
+ * fourth cycle edge. Reuse traded for isolation, on purpose.
+ *
+ * Invalidation fields (spec "The manifest"): each can only over-rebuild.
+ *  - sourceHash: the module's own bytes. Unchanged source ⇒ unchanged
+ *    import list ⇒ the recorded `deps` are still the true deps (the
+ *    load-bearing soundness invariant — imports are part of the source).
+ *  - deps + depsHash: recorded transitive agency imports, re-hashed from
+ *    the recorded paths at check time via computeDepsHash — the ONE
+ *    construction both writer and checker share. Missing dep = stale.
+ *    Freshness ALSO requires every recorded dep to have a manifest entry
+ *    whose OUTPUT exists: a skip never recurses into deps, so a deleted
+ *    dep .js would otherwise survive the skip and ship a broken import.
+ *  - stdlibHash: the closure walker EXCLUDES std:: imports, so no depsHash
+ *    can see a stdlib edit — yet stdlib content genuinely shapes emitted
+ *    output (resolveReExports bakes resolved stdlib paths in). Any stdlib
+ *    edit rebuilds the world.
+ *  - hasPkgImports: pkg:: imports are likewise closure-invisible and shape
+ *    emitted imports; modules touching pkg:: are NEVER skipped.
+ *  - compilerStamp: content hash of the compiled compiler (dist/lib minus
+ *    runtime/ and agents/). runtime/ because generated TEXT does not
+ *    depend on runtime internals; agents/ because those are the agency
+ *    compiler's OWN OUTPUT — including them would make every build
+ *    invalidate the next (self-invalidation loop). Content, not mtimes:
+ *    tsc-alias rewrites the whole outDir every build.
+ *  - configKey: compiled output bakes config in.
+ *
+ * The MANIFEST decides whether to compile; parseCache remains an
+ * intra-process memo consulted only once compilation is already happening.
+ */
+import * as crypto from "crypto";
+import * as fs from "fs";
+import * as path from "path";
+import { findProjectRoot } from "@/config.js";
+
+export type ManifestEntry = {
+  sourceHash: string;
+  /** Transitive agency-import paths, manifest-dir-relative, sorted. */
+  deps: string[];
+  depsHash: string;
+  stdlibHash: string;
+  hasPkgImports: boolean;
+  configKey: string;
+  compilerStamp: string;
+  /** Manifest-dir-relative output path. */
+  outputPath: string;
+};
+
+export type BuildManifest = {
+  version: 1;
+  /** Keyed by manifest-dir-relative module path. */
+  entries: Record<string, ManifestEntry>;
+};
+
+export const MANIFEST_DIR_NAME = ".agency-build";
+const MANIFEST_FILE = "manifest.json";
+
+export function manifestDirFor(entryFile: string): string {
+  const root = findProjectRoot(path.dirname(path.resolve(entryFile)));
+  return root ?? path.dirname(path.resolve(entryFile));
+}
+
+function emptyManifest(): BuildManifest {
+  return { version: 1, entries: Object.create(null) };
+}
+
+export function loadManifest(manifestDir: string): BuildManifest {
+  const file = path.join(manifestDir, MANIFEST_DIR_NAME, MANIFEST_FILE);
+  if (!fs.existsSync(file)) {
+    return emptyManifest();
+  }
+  try {
+    const parsed = JSON.parse(fs.readFileSync(file, "utf-8"));
+    if (parsed?.version !== 1 || typeof parsed.entries !== "object" || parsed.entries === null) {
+      return emptyManifest();
+    }
+    return { version: 1, entries: Object.assign(Object.create(null), parsed.entries) };
+  } catch (e) {
+    // A corrupt manifest only costs a full rebuild; log for traceability.
+    console.warn(`agency: ignoring corrupt build manifest at ${file}: ${e}`);
+    return emptyManifest();
+  }
+}
+
+export function saveManifest(manifestDir: string, manifest: BuildManifest): void {
+  const dir = path.join(manifestDir, MANIFEST_DIR_NAME);
+  fs.mkdirSync(dir, { recursive: true });
+  const file = path.join(dir, MANIFEST_FILE);
+  // Atomic write: concurrent compiles get last-writer-wins, never a torn file.
+  const tmpFile = `${file}.${process.pid}.tmp`;
+  fs.writeFileSync(tmpFile, JSON.stringify(manifest, null, 2));
+  fs.renameSync(tmpFile, file);
+}
+
+export function hashBytes(data: string | Buffer): string {
+  return crypto.createHash("sha256").update(data).digest("hex");
+}
+
+export function hashFile(absPath: string): string | null {
+  if (!fs.existsSync(absPath)) {
+    return null;
+  }
+  return hashBytes(fs.readFileSync(absPath));
+}
+
+/** THE construction of depsHash. Writer (manifestTracker) and checker
+ *  (isEntryFresh) both call this; never inline the expression. */
+export function computeDepsHash(depHashes: string[]): string {
+  return hashBytes(JSON.stringify(depHashes));
+}
+
+// Deliberate near-duplicate of lib/cli/util.ts findRecursively — see the
+// module doc comment (leaf-ness beats reuse here).
+function walkFiles(dir: string, extension: string, skipDirs: string[]): string[] {
+  const out: string[] = [];
+  const walk = (current: string) => {
+    for (const entry of fs.readdirSync(current, { withFileTypes: true })) {
+      const child = path.join(current, entry.name);
+      if (entry.isDirectory()) {
+        if (!skipDirs.includes(entry.name)) {
+          walk(child);
+        }
+      } else if (child.endsWith(extension)) {
+        out.push(child);
+      }
+    }
+  };
+  if (fs.existsSync(dir)) {
+    walk(dir);
+  }
+  return out.sort();
+}
+
+// NUL separators between path and content and between files: without a
+// delimiter, path/content boundaries are ambiguous in principle.
+function hashTree(dir: string, extension: string, skipDirs: string[]): string {
+  const hash = crypto.createHash("sha256");
+  for (const file of walkFiles(dir, extension, skipDirs)) {
+    hash.update(path.relative(dir, file));
+    hash.update("\0");
+    hash.update(fs.readFileSync(file));
+    hash.update("\0");
+  }
+  return hash.digest("hex");
+}
+
+export function computeStdlibHash(stdlibDir: string): string {
+  return hashTree(stdlibDir, ".agency", []);
+}
+
+export function computeCompilerStamp(distLibDir: string): string {
+  return hashTree(distLibDir, ".js", ["runtime", "agents"]);
+}
+
+export type FreshnessContext = {
+  manifestDir: string;
+  stdlibHash: string;
+  compilerStamp: string;
+  configKey: string;
+};
+
+/**
+ * The skip algorithm, from the manifest alone — no parsing:
+ * 1. sourceHash matches (⇒ recorded deps are still the true deps);
+ * 2. every recorded dep source exists and the recomputed depsHash matches;
+ * 3. every recorded dep has a manifest entry whose OUTPUT exists — a skip
+ *    never recurses into deps, so a deleted dep .js would otherwise
+ *    survive the skip and ship a broken import;
+ * 4. stdlibHash / compilerStamp / configKey match; hasPkgImports is false;
+ * 5. the module's own recorded output exists.
+ */
+export function isEntryFresh(
+  moduleRel: string,
+  manifest: BuildManifest,
+  ctx: FreshnessContext,
+): boolean {
+  const entry = manifest.entries[moduleRel];
+  if (!entry) {
+    return false;
+  }
+  if (entry.hasPkgImports) {
+    return false;
+  }
+  if (entry.stdlibHash !== ctx.stdlibHash) {
+    return false;
+  }
+  if (entry.compilerStamp !== ctx.compilerStamp) {
+    return false;
+  }
+  if (entry.configKey !== ctx.configKey) {
+    return false;
+  }
+  const sourceHash = hashFile(path.join(ctx.manifestDir, moduleRel));
+  if (sourceHash === null || sourceHash !== entry.sourceHash) {
+    return false;
+  }
+  const depHashes: string[] = [];
+  for (const dep of entry.deps) {
+    const depHash = hashFile(path.join(ctx.manifestDir, dep));
+    if (depHash === null) {
+      return false;
+    }
+    depHashes.push(depHash);
+    const depEntry = manifest.entries[dep];
+    if (!depEntry) {
+      return false;
+    }
+    if (!fs.existsSync(path.join(ctx.manifestDir, depEntry.outputPath))) {
+      return false;
+    }
+  }
+  if (computeDepsHash(depHashes) !== entry.depsHash) {
+    return false;
+  }
+  return fs.existsSync(path.join(ctx.manifestDir, entry.outputPath));
+}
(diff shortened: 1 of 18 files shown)


packages/agency-lang/lib/compiler/buildSession.ts
```diff
@@ -422,6 +538,13 @@ export class BuildSession {
       fs.writeFileSync(outputFile, result.code, "utf-8");
       const esbuildEndTime = performance.now();
       logTime({ label: `Transformed code for ${absoluteInputFile} with esbuild`, start: esbuildStartTime, end: esbuildEndTime, verbose });
+      const transitiveDeps = this.transitiveDeps(absoluteInputFile);
```
egonSchiele: **Should block merge: `serve.ts` records poisoned entries with empty deps → stale skips that outlive serve.**

`compileForServe` (`lib/cli/serve.ts:43`) calls `compile(config, file, undefined, { symbolTable, quiet })`. Trace what happens here:

1. No `importStrategy` → `resolveFreshness` = `"incremental"` → real tracker.
2. `options.symbolTable` set → `ensureCompiledClosure` treats the call as non-outermost and builds **no closure**.
3. `transitiveDeps` finds the module absent from `currentClosure?.programs` → returns `[]`; `subtreeHasPkgImport` → `false`.
4. This line records the entry with `deps: []` for a user module that may have real imports.

Now the poison persists: `agency serve app.agency` (app imports helper), edit `helper.agency`, then ANY plain `agency compile app.agency` — sourceHash unchanged, deps empty so nothing to check, stdlib/stamp/config match, output exists → **skip, stale `app.js`**. The bad entry survives until app itself is edited, `--force`, or `make clean`. This is exactly the stale-skip class the conservative-fields design promises away.

The fix is one condition: only record when the module's deps are actually knowable — present in `currentClosure.programs`, OR a stdlib entry (the legitimate `deps: []` case, covered by `stdlibHash`). A module in neither category must not be recorded (it can still *skip* off a previous well-formed entry, which stays sound).

```ts
const inClosure = !!this.currentClosure?.programs[absoluteInputFile];
if (inClosure || isStdlibEntry(absoluteInputFile)) {
  this.tracker.record(...);
}
```

Plus a test: compile with a caller-supplied `symbolTable` (the serve shape) and assert no entry is recorded — or that a subsequent dep edit still recompiles.

packages/agency-lang/lib/compiler/buildSession.ts
```diff
@@ -79,29 +113,61 @@ export class BuildSession {
...
+   * carve-out); multiple entries compile under ONE union closure, where
+   * closure errors THROW (`CompileClosureError`) so programmatic callers
+   * can attach context. Parse/typecheck failures inside per-file compiles
+   * keep their exit behavior. Returns the output path for a single
+   * non-directory entry, null otherwise.
+   */
+  compile(config: AgencyConfig, request: CompileRequest): string | null {
+    const { entries, outputFile, ...options } = request;
+    if (outputFile !== undefined && entries.length !== 1) {
+      throw new Error("outputFile is only valid with a single file entry");
```
egonSchiele, line 146: Hardening, one line: a request with `outputFile` set can hit a fresh entry (here via the fast path, and again via `tracker.isFresh` in `compileEntry`) and return the requested path **without ever writing it** — the freshness check only knows the *recorded* `outputPath`.

This is unreachable today by luck of layering: the only caller passing `outputFile` is `run()` (`commands.ts:254`), which also passes `RunStrategy`, and the importStrategy rule already forces `"always"`. But both `commands.compile(config, input, outputFile)` and `CompileRequest.outputFile` are public surface — the next caller that passes an output path without a strategy gets a path to a file that does not exist.

Cheapest sound rule, matching the `--ts` precedent (different artifact → not manifest-tracked): `outputFile !== undefined` also resolves to `"always"`.

packages/agency-lang/lib/compiler/buildSession.ts
```diff
@@ -79,29 +113,61 @@ export class BuildSession {
...
+      // with the same walker the compile path uses.
+      const allFiles = expandEntries(entries);
+      if (this.tracker.allFresh(allFiles)) {
+        for (const file of allFiles) {
+          this.compiledFiles.add(file);
+        }
+        if (!options.quiet) {
+          console.log(`${allFiles.length} file(s) up to date`);
+        }
+        return allFiles.length === 1 ? this.tracker.outputFor(allFiles[0]) : null;
```
egonSchiele: Nit: contract drift on the fast path. A single **directory** entry containing exactly one `.agency` file returns `outputFor(file)` here, where the compile path's directory contract returns `null`. No current caller distinguishes, but it makes the return type depend on freshness. Cheap fix: have `expandEntries` also report whether any entry was a directory, and return `null` in that case.

packages/agency-lang/lib/compiler/manifestTracker.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.dirty = false;
+  }
+}
+
+/** Policy is resolved HERE, once:
+ *  "always" → the shared NOOP tracker (no state, no IO, isFresh always false)
+ *  "force"  → real tracker with reads disabled, writes on
+ *  "incremental" → real tracker, reads + writes on. */
+export function createManifestTracker(
+  _config: AgencyConfig,
```
egonSchiele: Nit: `_config` is a dead parameter. Either drop it, or use it — deriving the config key in here (moving `deriveConfigKey` into the tracker, which now owns every other piece of freshness identity) would drop the fourth parameter and remove the possibility of a caller passing a key that does not match the config.

While here, a perf note for the file: each tracker construction recomputes `computeStdlibHash` (~94 file reads) + `computeCompilerStamp` (all of `dist/lib`) + `loadManifest`. The warm-make path constructs six trackers (one per CLI input), so that work runs 6×. Measured total is fine (2.5s), so leave it — but this is the first knob if warm make regresses, and any memo needs a staleness story (watch mode can outlive a stdlib edit), so per-session memoization, not module-level.

egonSchiele (commented):
Reviewed the full diff with the freshness algorithm traced against every `compile()` caller in the tree. **One finding should block merge** (the serve path poisons the manifest — details inline at the `record` call); one cheap hardening, two nits. Everything else held up under adversarial reading.

**What is notably right:**
- All six plan-review findings are folded in and each got a real test: dep OUTPUT + dep ENTRY checks in `isEntryFresh` (with the deleted-dep-output test), caller-supplied `importStrategy` → `"always"` (tested in both directions), `programHasPkgImport` sharing the closure walker's edge extraction (re-export test included), the doc stamp keyed on stdlibHash + compilerStamp, NUL separators in the tree hashes, and the honest process.exit-bypasses-finally caveat at the flush.
- `manifestTracker.ts` is a genuinely good structure the plan didn't ask for: policy interpreted in exactly one factory, NOOP tracker absorbing "always" so session call sites are unconditional one-liners. The doc line "if you find yourself comparing freshness === ... elsewhere, extend the tracker" encodes the lesson permanently.
- The interface collapse to one entries-based `compile()` lands the spec's destination shape, and the adapted isolation tests (manifest removed so re-emission is the observable again) show the test intent survived the semantics change.
- The purity gate (cold build hash-identical to the incremental tree across 96 files) is the strongest single piece of evidence in the PR.
- The `agents/` exclusion rationale in the compilerStamp comment — self-invalidation loop — is a better argument than the spec's original one.

**Verified against callers:** the `outputFile`-plus-fresh interaction is unreachable today only because `run()` also passes `RunStrategy` (which forces "always") — see the hardening comment; `runAgencyNode`, coverage, and `runWorkdir` all pass `RunStrategy` and are correctly never skip-eligible; watch mode gets incremental rebuilds for free through the delegate.

packages/agency-lang/scripts/stdlib-stamp.mjs
```diff
@@ -0,0 +1,19 @@
...
+// the doc generator too, which lives inside the compilerStamp scope —
+// stdlibHash alone would leave tracked docs stale after a generator edit
+// until some unrelated stdlib change. Imports the canonical
+// implementations from dist (no duplication) — requires a built dist,
+// which `make all` guarantees (build runs before doc).
+import {
+  computeStdlibHash,
+  computeCompilerStamp,
+  hashBytes,
+} from "../dist/lib/compiler/buildManifest.js";
```
egonSchiele: I don't love that this imports from the dist/ directory, a better pattern would be to write these in type script as well, compile them, and then use them from the dist/ directory... unless the reason you can't do that is because they are part of the compilation process itself? Will this all still work when someone uses agency as an npm package? Do you need to make any changes to package.json to make it work?

packages/agency-lang/lib/compiler/buildSession.ts
egonSchiele: Fixed in 474136ef — record only when deps are knowable: closure-covered, or a stdlib module (closure-free by design; deps [] is correct there because stdlibHash covers all of stdlib). A serve-style compile now records nothing. Two regression tests: the no-entry invariant for threaded-symbolTable compiles, and your exact scenario (serve-compile, edit helper, plain compile → re-emits).

packages/agency-lang/lib/compiler/buildSession.ts
egonSchiele: Fixed in 474136ef — outputFile !== undefined now resolves to freshness always in resolveFreshness, exactly the --ts precedent. Test added: a fresh entry plus an explicit outputFile writes the requested file rather than returning a phantom path.

packages/agency-lang/lib/compiler/buildSession.ts
egonSchiele: Fixed in 474136ef — expandEntries reports hasDirectory and the fast path returns null for directory entries regardless of member count, matching the compile path contract. Test added.

packages/agency-lang/lib/compiler/manifestTracker.ts
egonSchiele: Fixed in 474136ef, taking the use-it option: createManifestTracker(config, entryFile, freshness) derives the key itself via deriveConfigKey, which moved to buildManifest.ts (the leaf owns manifest identity; the session imports it for the cross-config assert). The perf note is preserved verbatim as a comment on the factory — per-session memoization named as the first knob, module-level explicitly ruled out for the watch-mode staleness reason.

packages/agency-lang/scripts/stdlib-stamp.mjs
egonSchiele: Good call — converted in 474136ef. stdlib-stamp is now scripts/stdlib-stamp.ts, compiled by tsc into dist/scripts/stdlib-stamp.js like every other script, importing buildManifest through the normal @/ alias. It is NOT part of the compile bootstrap (it only runs from make doc, which make all sequences after make build), so requiring a built dist is safe. stage-agents.mjs is the one that stays plain .mjs, deliberately: it must run on bare node BEFORE dist exists (make agents right after make clean) — that reason is now in its header comment. On npm packaging: no package.json changes needed. The makefile is a repo-dev tool that does not ship; the compiled stamp script rides inside ./dist which the files field already packs; and nothing in the published package invokes either script. Package consumers hit the manifest only through agency compile, whose module resolution is unchanged.

packages/agency-lang/lib/compiler/buildManifest.ts
```diff
@@ -0,0 +1,224 @@
...
+  entries: Record<string, ManifestEntry>;
+};
+
+export const MANIFEST_DIR_NAME = ".agency-build";
+const MANIFEST_FILE = "manifest.json";
+
+export function manifestDirFor(entryFile: string): string {
+  const root = findProjectRoot(path.dirname(path.resolve(entryFile)));
+  return root ?? path.dirname(path.resolve(entryFile));
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 474136ef — manifestDirFor stats the entry: directories anchor findProjectRoot (and the no-agency.json fallback) at themselves rather than their parent. Test added for the directory case.

packages/agency-lang/lib/compiler/buildManifest.ts
```diff
@@ -0,0 +1,224 @@
...
+  manifest: BuildManifest,
+  ctx: FreshnessContext,
+): boolean {
+  const entry = manifest.entries[moduleRel];
+  if (!entry) {
+    return false;
+  }
+  if (entry.hasPkgImports) {
+    return false;
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 474136ef — isEntryFresh validates the runtime shape of every field it touches (entryHasValidShape) and treats any mismatch as stale, plus a type guard on dep-entry outputPath inside the dep loop. Tests: non-array deps and non-string outputPath both read as stale, never a throw.

packages/agency-lang/lib/compiler/manifestTracker.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  outputFor(absModule: string): string | null {
+    const rel = path.relative(this.manifestDir, absModule);
+    const entry = this.manifest.entries[rel];
+    if (!entry) {
+      return null;
+    }
+    return path.join(this.manifestDir, entry.outputPath);
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 474136ef — outputFor returns null for a missing entry or non-string outputPath. Test added.

egonSchiele (commented):
Re-reviewed `474136ef` against the prior review. **All four findings are resolved correctly — no new issues found. This resolves my should-block finding; good to merge from my side** (CI still running at review time; the three manifest suites — 60 tests — pass locally on this commit).

How each fix holds up under the same traces that found the originals:

1. **Serve poisoning (was should-block): fixed at the right layer.** `depsKnowable = isStdlibModule || closure-covered` gates the record; a serve-style compile (threaded symbolTable, no closure) now records nothing — and its recursive children record nothing either, since they share the closure-less state. Two things I checked specifically: (a) a serve compile can still *skip* off a prior well-formed entry, which stays sound because that entry's deps came from a real closure; (b) the cost of the fix is only that serve never seeds the manifest — the next plain compile records once and everything proceeds. Both regression tests assert the property that actually matters (no entry; edit-dep-then-compile re-emits), not the mechanism.
2. **`outputFile` → `"always"`:** resolving from the whole request in one place means the fast path, `compileEntry`'s per-module check, and the record path are all consistently off for that operation — no half-tracked state. Test asserts the file exists, not just the return value. Right shape.
3. **Directory contract:** `hasDirectory` from `expandEntries` restores the null contract; test covers the one-file-directory corner specifically.
4. **Tracker owns config identity:** better than what I asked for — `deriveConfigKey` moved into `buildManifest` (the leaf owns manifest identity) and BOTH consumers (the tracker and `compileGroups`' cross-config assert) now share the same function, so the conflict-assert key and the manifest key can never diverge.

Unprompted hardening that also checks out: `entryHasValidShape` + the dep-loop and `outputFor` guards make a hand-edited/older-schema manifest cost a rebuild instead of a crash (tested with mutated JSON, not mocks); `manifestDirFor` anchoring directory entries at themselves fixes a real `.agency-build/` sibling-pollution case and got its own test; and `stdlib-stamp` moving to compiled TypeScript with the bootstrap-contrast comment (why `stage-agents.mjs` must stay `.mjs`) documents the one asymmetry someone would otherwise "clean up" incorrectly.

packages/agency-lang/scripts/stdlib-stamp.ts
```diff
@@ -0,0 +1,24 @@
...
+import {
+  computeStdlibHash,
+  computeCompilerStamp,
+  hashBytes,
+} from "@/compiler/buildManifest.js";
+
+// dist/scripts/stdlib-stamp.js → package root is one level up.
+const packageRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..");
+const stdlibHash = computeStdlibHash(path.join(packageRoot, "stdlib"));
+const compilerStamp = computeCompilerStamp(path.join(packageRoot, "dist", "lib"));
```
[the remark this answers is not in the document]
egonSchiele: Confirmed and fixed in 83fa3bad — real bug, introduced by the TS conversion in the previous review round: the script moved from scripts/ (one level below the package root) to dist/scripts/ (two levels), packageRoot resolved to dist/, both hashed dirs were nonexistent, and hashTree silently returned the empty hash — the stamp was constant and make doc would never regenerate. Fixed with ../.. plus loud existence guards on both dirs so a future path regression exits 1 instead of silently freezing the stamp. Verified empirically: the stamp now changes on a stdlib edit AND on a compiler edit, and reverts when they revert.

packages/agency-lang/scripts/stage-agents.mjs
```diff
@@ -0,0 +1,73 @@
...
+      }
+    }
+    fs.unlinkSync(destFile);
+    deleted.push(rel);
+  }
+  return { copied, deleted };
+}
+
+const invokedDirectly =
+  process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href;
```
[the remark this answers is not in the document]
egonSchiele: Applied in 83fa3bad — argv[1] is now path.resolve()d before pathToFileURL. One correction to the premise: node absolutizes process.argv[1] for direct script invocations (including the make call), and pathToFileURL also accepts relative paths without throwing on this platform — so the crash was not reachable here. The resolve is free hardening against exotic launchers, so it went in anyway, with a comment saying which part is load-bearing.

PR #478, opened 2026-07-10
Add built-in utility types: Partial, Required, Pick, Omit, NonNullable

egonSchiele:
Adds five built-in utility types modeled on TypeScript, adapted to Agency null-based optionality: `Partial`, `Required`, `Pick`, `Omit`, `NonNullable`.

Design spec: `docs/superpowers/specs/2026-07-09-utility-types-design.md` (committed on this branch). Implementation plan with review: `docs/superpowers/plans/2026-07-09-utility-types.md`.

## What

- **`lib/typeChecker/utilityTypes.ts`** — one table-driven module (`UTILITY_TYPES`: name -> arity + transform) owning all five transforms. `UTILITY_TYPE_ARITY` is derived from it and consumed by `validate.ts` (arity diagnostics) and `index.ts` (reserved names).
- **One resolver branch** in `resolveTypeWithGuard` (`assignability.ts`), beside the existing `Array`/`Schema`/`Record` branches. Transforms evaluate EAGERLY to plain object/union types, so assignability, narrowing, flow, and zod codegen need zero changes.
- Semantics: shallow transforms; optionality is `| null` (matching the existing `p?:` desugaring, so `Partial<T>` output is identical to hand-written optional properties). `Pick` errors on unknown keys, `Omit` allows them (TS parity). `NonNullable<null>` and `Required` of an exactly-null property produce `never`. `Record`/array/primitive/union arguments error (v1 decision).

## Breaking change

`Partial`, `Required`, `Pick`, `Omit`, and `NonNullable` are now reserved type names (like `Result`); a user alias with one of these exact names is rejected. No stdlib or test code used them.

## Known gap (pinned by test, spec follow-up)

Semantic argument errors (bad `Pick` key, non-object argument) do NOT surface as typecheck diagnostics — `safeResolveType` swallows the resolver throw and the error becomes fatal at codegen instead. Same behavior as `Record` key errors today. A pipeline test pins this; located diagnostics are a named follow-up in the spec (the naive fix in `validateTypeReferences` would false-positive on generic alias bodies like `type PartialOf<T> = Partial<T>`).

## Tests

- 37 unit + pipeline tests (`utilityTypes.test.ts`): per-type semantics, all error paths, tag/description preservation, input-non-mutation (shared alias-body safety), alias-resolution subtleties, composition (`Partial<Pick<...>>`), `type PartialOf<T> = Partial<T>` delegation, narrowing composition, and the pinned known-gap behaviors.
- Codegen golden fixture (`tests/typescriptGenerator/utilityTypes.mjs`) pinning the zod lowering on three paths: alias schema consts, tool arg schemas, and LLM `responseFormat`.
- Five agency execution tests (accept + reject nodes each, no LLM calls) proving parse-time behavior per type.
- Formatter round-trip tests: `Partial<User>` and `Pick<User, "name" | "email">` survive `fmt` as written.

## Notes for reviewers

- The full `lib` unit suite passes locally (7583 tests); the agency execution suite runs in CI.
- Two pre-existing issues discovered in execution:
  - Chaining `.parseJSON(...)` directly on `schema(...)` does not parse (grammar limitation, filed as #480), so the execution tests bind the schema to a variable first.
  - A null literal in an object literal synthesized to `any` (no null case in `synthType`), which made `synthObject` degrade the whole literal and silently skip assignment checking of every sibling property. **Fixed in this PR**, scoped to object literals (a general `synthType` null case would bind `let x = null` to the null type and break later reassignment). This deliberately flips the KNOWN FAIL-OPEN tripwire test in `lib/optimize/constraint.test.ts`, which existed to detect exactly this tightening.


packages/agency-lang/docs/site/guide/types.md
```diff
@@ -175,6 +175,51 @@ When you write an object *literal*, every key in the literal must be in its decl
...
+def updateUser(id: string, changes: Partial<User>): string {
+  // changes.name is string | null — guard before use:
+  if (changes.name != null) {
+    return changes.name  // narrowed to string here
+  }
+  return "no name change"
+}
+```
+
+Details worth knowing:
```
egonSchiele: Remove this entire "Details worth knowing" section

packages/agency-lang/lib/typeChecker/utilityTypes.ts
```diff
@@ -0,0 +1,196 @@
...
+    return resolvedMember.value;
+  });
+}
+
+/**
+ * The single source of truth: name -> arity + transform. Adding a utility
+ * type is one entry here; UTILITY_TYPE_ARITY (consumed by validate.ts and
+ * RESERVED_TYPE_NAMES in index.ts) is derived from it.
+ */
+const UTILITY_TYPES: Record<
```
egonSchiele: Should `Record` go here?

egonSchiele: What about `Array`?

packages/agency-lang/lib/typeChecker/utilityTypes.ts
```diff
@@ -0,0 +1,196 @@
...
+ */
+export function evalUtilityType(
+  name: string,
+  typeArgs: VariableType[],
+  resolve: Resolve,
+): VariableType {
+  const entry = UTILITY_TYPES[name];
+  if (!entry) {
+    throw new TypeError(`Unknown utility type ${name}`);
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in be76e8d3. The registry lookup is now own-property guarded (`lookupBuiltinGeneric` uses `Object.prototype.hasOwnProperty.call`, matching the scope.ts/flow.ts discipline), and the derived `BUILTIN_GENERIC_ARITY` table is null-prototype since validate.ts indexes it bare. Tests pin both: `isBuiltinGenericName("toString"/"constructor"/"hasOwnProperty")` are false, and `type X = toString<number>` no longer produces the garbled arity message quoting `Object.prototype.toString`. One residual noted in the test: the *alias table* (`typeAliases[t.name]`) is still indexed bare across the codebase, so the diagnostic for `toString<...>` reads "not a generic type" rather than "Unknown generic type" — that prototype hazard is pre-existing and codebase-wide, left for a follow-up.

packages/agency-lang/lib/typeChecker/assignability.ts
```diff
@@ -135,6 +136,19 @@ function resolveTypeWithGuard(
...
+    // carries this call's in-progress guard, so recursive alias arguments
+    // degrade the same way Record args do (self-refs stay nominal).
+    if (isUtilityTypeName(vt.name)) {
+      const evaluated = evalUtilityType(vt.name, vt.typeArgs, (t) =>
+        resolveTypeWithGuard(t, typeAliases, inProgress),
+      );
+      // Sibling branches (Array/Schema/Record) drop use-site tags; we keep
+      // them DELIBERATELY so `@validate(...) Partial<User>` annotations
+      // survive. The divergence is intentional, not an oversight.
+      return attachAliasTags(evaluated, vt.tags);
```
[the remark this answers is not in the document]
egonSchiele: Confirmed and fixed in be76e8d3 — you were right that the merge orientation was reversed. The transform result now merges via `mergeTagSets(result.tags /* alias side, from resolving the argument */, vt.tags /* use-site */)` (`withUseSiteTags` in builtinGenerics.ts), so use-site `@jsonSchema` keys override and validators run alias-first. A unit test pins the validator order (`[fromAlias, fromUseSite]`).

packages/agency-lang/docs/site/guide/types.md
```diff
@@ -175,6 +175,51 @@ When you write an object *literal*, every key in the literal must be in its decl
...
+- In Agency source, object keys are always required. `Partial` means you can
+  pass `null` for a value, not omit the key. Key omission is only forgiven
+  when parsing JSON with `schema(...)`: missing keys coalesce to `null`.
+- Because `p?: V` and `p: V | null` are the same type after parsing,
+  `Required` un-optionalizes both — it cannot distinguish a property you
+  marked optional from one you declared nullable on purpose.
+- `Pick` errors on a key that does not exist on the target; `Omit` allows it
+  (matching TypeScript).
+- The transforms need concrete object types: applying them to `Record<K, V>`,
+  arrays, primitives, or unions is an error.
```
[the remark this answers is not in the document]
egonSchiele: Moot now — the whole "Details worth knowing" section was removed per the maintainer comment below (be76e8d3).

packages/agency-lang/docs/site/guide/types.md
egonSchiele: Removed in be76e8d3.

packages/agency-lang/lib/typeChecker/utilityTypes.ts
egonSchiele: Yes — done in be76e8d3, for `Array` and `Schema` too. The module is now `builtinGenerics.ts` and the table is the single registry for ALL built-in generics; the hand-rolled `Array`/`Schema`/`Record` branches in `resolveTypeWithGuard` are gone, and `BUILTIN_GENERIC_ARITY` (validate.ts) plus `RESERVED_GENERIC_NAMES` (index.ts) are derived from it.

One deliberate nuance: entries carry a `reserved` flag. The five utility types are reserved (user aliases rejected), while `Array`/`Schema`/`Record` keep their historical behavior — a user alias with those names is silently shadowed by the builtin, not rejected. Flipping them to reserved would be a breaking change beyond this PR, so I kept it, but it is now a one-word change if you want it.

Behavior preservation is verified: `make fixtures` after the consolidation produced zero fixture churn, and the full lib suite (7594 tests) passes. Tag handling is also per-entry — utility types merge use-site tags (per the other review thread), `Record` keeps them verbatim on its wrapper, `Array`/`Schema` drop them as they always have (commented in the table).

egonSchiele: Also done — `Array` (and `Schema`) moved into the registry alongside `Record` in be76e8d3; details in the reply on the `Record` thread above.

PR #479, opened 2026-07-10
Sandbox the agent home in tests via AGENCY_AGENT_HOME (fixes #469)

### LINKED ISSUE #469: Agent settings tests clobber the real ~/.agency-agent/settings.json and race under -p 12
## Problem

The built-in agent settings tests (`packages/agency-lang/lib/agents/agency-agent/tests/settings.agency`) import `SETTINGS_PATH` / `SETTINGS_FILE` / `POLICY_DIR` from the agent's real config (`../lib/config.agency`) and then delete, corrupt, and rewrite that file as test setup:

```
node missingFileLoadsAsEmpty(): string {
  remove(SETTINGS_PATH) with approve
  ...
node corruptFileLoadsAsEmpty(): string {
  write(SETTINGS_FILE, "not json{{{", POLICY_DIR) with approve
```

Two consequences:

1. **Every local `pnpm run test:agents` run destroys the developer's real `~/.agency-agent/settings.json`.** After a run it contains whatever the last test node wrote (observed: `{"model":{"slots":{"reasoning":"claude-opus-4-8"}}}` — test fixture data, not user config). Anyone who configured model slots via `/model` has silently lost that configuration on every local agents-suite run.

2. **The shared real file races under the parallel runner.** With `-p 12`, another test file's fs activity can interleave with settings.agency's sequential nodes. Observed failure (local, 2026-07-09): `roundTripsModelSlots` loaded the file between corruption and its own save — `Warning: settings.json is not valid JSON` followed by `TypeError: Cannot read properties of undefined (reading 'reasoning')`. The file passes 6/6 in isolation and on rerun; classic shared-mutable-state flake.

CI is unaffected in practice (throwaway HOME), which is why this only bites developers.

## Suggested fix

Point the settings layer at a sandboxed path during tests — e.g. honor an `AGENCY_AGENT_HOME` (or similar) env var in `lib/config.agency` when resolving `SETTINGS_PATH`, and have the test runner (or the test file itself) set it to a temp dir. That fixes both the data loss and the race in one move, and mirrors how the test suite already sandboxes other global state (per-test subprocess isolation, deterministic LLM client).

## Context

Found while running the Stage A PR 2 verification matrix (#468); the failure blocked an E2E gate until diagnosed. Not introduced by that PR — the tests have written to the real path since they landed.

egonSchiele:
Fixes #469: the agent settings tests read and wrote the REAL `~/.agency-agent/settings.json` — every local `test:agents` run destroyed the developer settings (leaving test fixture data behind), and the shared file raced under `-p 12` (observed: `roundTripsModelSlots` loading mid-corruption).

## Fix

One env var at the root of the path derivation, plus the harness setting it:

- `lib/agents/agency-agent/lib/config.agency`: `POLICY_DIR = env("AGENCY_AGENT_HOME") ?? path.join(os.homedir(), ".agency-agent")`. Settings, policy, session-policy, and history paths all derive from `POLICY_DIR`, so the whole agent home redirects at once.
- The test harness sets `AGENCY_AGENT_HOME` to a fresh temp dir per test case (`executeNodeAsync`, deterministic mode only) and per agency-js test dir (`runTsTestDir`), removed after the subprocess exits. Non-test paths are untouched — the var is only set where `AGENCY_LLM_MOCKS` already is.
- Per-case isolation also dissolves the `-p 12` race: no two test cases share a home anymore.

## Verification

- New execution test (`tests/agency/agent-home-sandbox.agency`): the harness sets the sandbox and `POLICY_DIR` resolves to it (was RED before the fix).
- **Acceptance**: full `test:agents` (16 files green, including the settings suite that raced) with the real `~/.agency-agent/settings.json` hash- and mtime-identical before/after.
- Full vitest suite, agency + agency-js slices, lint clean. (Three `scripts/agency.test.ts` CLI-spawn tests timed out at 5s under full-suite machine load; 11/11 in isolation — pre-existing load sensitivity, unrelated paths.)

## Review follow-ups (da802a9b)

- **New `agency agent --agent-home <dir>` flag** (review ask): use that directory instead of `~/.agency-agent` for settings, policy, and history — the user-facing form of the env var, mirroring Claude Code's `CLAUDE_CONFIG_DIR` + flag pair. Implemented via the existing `--trace`/`--log-file` mechanism: `runBundledAgent` pre-scans the forwarded argv (`agentHomeOverride`, same std::args-matching tokenization) and injects `AGENCY_AGENT_HOME` into the child env, since the agent derives its paths in static-const initializers that run before `main()` parses flags. Flag beats env var; dir resolved absolute; bare `--agent-home` is a usage error. Documented in `docs/site/cli/agent.md`.
- **Empty-string guard** (Copilot): a set-but-empty `AGENCY_AGENT_HOME` previously passed the `??` fallback and made every derived path cwd-relative. `resolveAgentHome` now treats `""` as unset (`tests/agency/agent-home-resolve.agency`).
- `agent-home-sandbox.test.json` pins `useTestLLMProvider` so the sandbox assertion also holds on a single-file `agency test` run.

Verification for the follow-ups: `lib/cli` vitest slice 360/360 (incl. new `agentHomeOverride` + spawn-env cases), full `test:agents` 152/152, both agency execution tests, lint clean, `--help`/bare-flag smoke.

Note for reviewers: your local `~/.agency-agent/settings.json` currently contains test residue from before this fix — worth reconfiguring model slots after this merges.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/lib/config.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/lib/config.agency
@@ -6,13 +6,25 @@ import { env } from "std::system"
 export static const agentNames = ["code", "research"]
 export static const POLICY_FILE = "policy.json"
 export static const HISTORY_FILE = "history"
-// AGENCY_AGENT_HOME redirects the WHOLE agent home (settings, policy,
-// history all derive from POLICY_DIR). The test harness sets it to a
-// per-test sandbox dir so tests can never clobber the real
-// ~/.agency-agent (issue #469: the settings tests deleted/corrupted the
-// developer file on every local run, and the shared file raced under
-// the parallel runner).
-export static const POLICY_DIR = env("AGENCY_AGENT_HOME") ?? path.join(os.homedir(), ".agency-agent")
+// AGENCY_AGENT_HOME (or its flag form, `agency agent --agent-home`,
+// which the CLI launcher injects into this process's env) redirects the
+// WHOLE agent home: settings, policy, and history all derive from
+// POLICY_DIR. The test harness sets it to a per-test sandbox dir so
+// tests can never clobber the real ~/.agency-agent (issue #469: the
+// settings tests deleted/corrupted the developer file on every local
+// run, and the shared file raced under the parallel runner).
+//
+// An empty override counts as unset: env() returns "" for a
+// set-but-empty var, and without the guard every derived path would
+// silently become cwd-relative (a stray `settings.json` written into —
+// or read from — whatever directory the agent runs in).
+def resolveAgentHome(override: string | null): string {
+  if (override == null || override == "") {
+    return path.join(os.homedir(), ".agency-agent")
+  }
+  return override
+}
+export static const POLICY_DIR = resolveAgentHome(env("AGENCY_AGENT_HOME"))
 export static const POLICY_PATH = path.join(POLICY_DIR, POLICY_FILE)
 // A --policy built-in is materialized here (not to POLICY_PATH) so a
 // one-off override never clobbers the user's saved policy, and a
--- a/packages/agency-lang/lib/agents/agency-agent/lib/config.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/lib/config.agency
@@ -1,11 +1,18 @@
 import {
   ScopedRuleFields
  } from "std::policy"
+import { env } from "std::system"
 
 export static const agentNames = ["code", "research"]
 export static const POLICY_FILE = "policy.json"
 export static const HISTORY_FILE = "history"
-export static const POLICY_DIR = path.join(os.homedir(), ".agency-agent")
+// AGENCY_AGENT_HOME redirects the WHOLE agent home (settings, policy,
+// history all derive from POLICY_DIR). The test harness sets it to a
+// per-test sandbox dir so tests can never clobber the real
+// ~/.agency-agent (issue #469: the settings tests deleted/corrupted the
+// developer file on every local run, and the shared file raced under
+// the parallel runner).
+export static const POLICY_DIR = env("AGENCY_AGENT_HOME") ?? path.join(os.homedir(), ".agency-agent")
 export static const POLICY_PATH = path.join(POLICY_DIR, POLICY_FILE)
 // A --policy built-in is materialized here (not to POLICY_PATH) so a
 // one-off override never clobbers the user's saved policy, and a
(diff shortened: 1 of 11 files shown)


packages/agency-lang/lib/agents/agency-agent/lib/config.agency
```diff
@@ -1,11 +1,18 @@
 import {
   ScopedRuleFields
  } from "std::policy"
+import { env } from "std::system"
 
 export static const agentNames = ["code", "research"]
 export static const POLICY_FILE = "policy.json"
 export static const HISTORY_FILE = "history"
-export static const POLICY_DIR = path.join(os.homedir(), ".agency-agent")
+// AGENCY_AGENT_HOME redirects the WHOLE agent home (settings, policy,
```
egonSchiele: Instead of this hack, why don't we actually add an option on the agency agent that users can use to pass a path to a settings file or a directory that the agent should use instead of the home directory? That way users could use it for other things instead of just tests. Does Claude Code have this functionality? What does it offer?

egonSchiele: Claude Code offers exactly this pair: a `CLAUDE_CONFIG_DIR` env var that relocates the whole `~/.claude` directory, plus a `--settings <file>` flag that layers one settings file for a single invocation (on top of hierarchical enterprise/project/user settings files).

Added the flag form in da802a9b: `agency agent --agent-home <dir>` uses that directory instead of `~/.agency-agent` for settings, policy, and history. It rides the same mechanism as `--trace`/`--log-file` — `runBundledAgent` pre-scans the forwarded argv and injects `AGENCY_AGENT_HOME` into the child env, because the agent derives its paths in static-const initializers that run before `main()` could parse a flag. Flag wins over the env var; relative dirs resolve against cwd. Both forms are documented in `docs/site/cli/agent.md`, so the env var is now a user-facing feature (matching the `CLAUDE_CONFIG_DIR` precedent) rather than test plumbing.

A per-file `--settings <file>` override and project-level settings layering would sit on top of this cleanly; left out of this PR since its job is stopping the test-time data loss. Can file a follow-up issue if you want that.

packages/agency-lang/lib/agents/agency-agent/lib/config.agency
```diff
@@ -1,11 +1,18 @@
...
+// history all derive from POLICY_DIR). The test harness sets it to a
+// per-test sandbox dir so tests can never clobber the real
+// ~/.agency-agent (issue #469: the settings tests deleted/corrupted the
+// developer file on every local run, and the shared file raced under
+// the parallel runner).
+export static const POLICY_DIR = env("AGENCY_AGENT_HOME") ?? path.join(os.homedir(), ".agency-agent")
```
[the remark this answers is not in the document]
egonSchiele: Fixed in da802a9b — `resolveAgentHome` now treats an empty override the same as unset (null and `""` both fall back to `~/.agency-agent`), so a set-but-empty `AGENCY_AGENT_HOME` can no longer turn the derived paths cwd-relative. Covered by `tests/agency/agent-home-resolve.agency`; the same guard applies to the new `--agent-home` flag.

Issue #484, opened 2026-07-10
Recursive value-parameterized alias hangs codegen (unguarded inline expansion)


egonSchiele:
Found while fixing recursive type aliases (#470, PR forthcoming); scoped out of that PR deliberately.

A value-parameterized alias whose body references itself hangs codegen:

```
type Weird(n: number) = {
  next: Weird(n),
}
```

`mapTypeToSchemaInner` inlines value-param instantiations by recursion (the `isValueParamInstantiation` branch in `lib/backends/typescriptGenerator/typeToZodSchema.ts`) with no `seen` guard, so a self-referencing instantiation expands forever at compile time.

The plain-alias equivalents are now handled: `resolveTypeWithGuard` has an `inProgress` guard, `isAssignable` is coinductive, and schema consts defer with `z.lazy`. The value-param inline path is the remaining unguarded recursion.

Likely fix shape: thread a `seen` set (keyed by alias name + canonicalized valueArgs) through the value-param inlining, and either emit a factory self-call (the factories already exist as hoisted functions) or reject with a clear "recursive value-parameterized alias" error like the `type Loop = Loop` guard in `processTypeAlias`.

PR #485, opened 2026-07-10
Fix recursive and forward-referencing type aliases (closes #470, #473)

### LINKED ISSUE #470: Recursive type aliases crash: codegen TDZ ReferenceError + isAssignable stack overflow
Recursive type aliases parse and typecheck in simple cases, but crash in two different places. Probe used:

```
type Tree = {
  value: number,
  children: Tree[],
}
```

## Bug 1: codegen emits a self-referencing const (TDZ crash at module load)

Compiling a program that uses `Tree` generates:

```js
const Tree = z.object({ "value": z.number(), "children": z.array(Tree) });
```

which throws `ReferenceError: Cannot access 'Tree' before initialization` the moment the compiled file loads. Compile and typecheck both succeed, so this looks fine until you run it.

**Fix direction:** emit cycle-safe references — `z.lazy(() => Tree)` (or zod getters) for alias references that close a cycle. Needs the alias dependency graph; either lazy-wrap all alias references (uniform, but churns every fixture) or topologically sort and lazy-wrap only back-edges. The per-variable init dep graph (`docs/dev/init-topsort.md`) is precedent for the graph + topsort approach.

## Bug 2: `isAssignable` stack-overflows on recursive alias vs itself

```
def id(x: Tree): Tree {
  return x
}

node second() {
  const a: Tree = { value: 3, children: [] }
  const b: Tree = id(a)   // typechecker crashes here
}
```

`pnpm run agency typecheck` dies with `RangeError: Maximum call stack size exceeded`. Each `isAssignable` call resolves the alias one level with a fresh `inProgress` guard (`resolveTypeWithGuard` in `lib/typeChecker/assignability.ts`), so comparing `Tree` to `Tree` walks into `children`, hits `Tree` vs `Tree` again, and recurses forever. Literal-only programs pass because empty arrays synthesize `any[]`, which short-circuits — that is why this went unnoticed.

**Fix direction:** coinductive assignability, the standard trick from TypeScript and other structural checkers: thread a seen-pair set through `isAssignable` (keyed by the alias-name pair or canonical type keys); on re-entry of a pair already being compared, assume the relation holds and return true. Same shape as the existing `inProgress` guard in `resolveTypeWithGuard`, one level up.

## Also needed for full support

- Verify the LLM structured-output path: a recursive zod schema serializes to JSON schema with `$ref`/`$defs`. OpenAI structured outputs supports this, but it needs end-to-end verification through smoltalk, and providers that cannot take `$ref` (e.g. some local models) should get a clear error rather than a silent failure.
- The runtime validation walker already hardcodes a recursion depth (per `docs/site/guide/type-validation.md`), so that side works today.

Note: the type representation itself is already correct — the resolution guard leaves self-references as nominal `typeAliasVariable` nodes. Only the consumers (assignability, codegen) mishandle the back-reference. Once fixed, eagerly-evaluated utility types like a future `Partial<T>` compose fine with recursion, since Partial is shallow.

### LINKED ISSUE #473: Centralize structural type identity in a canonical typeKey() (replace raw JSON.stringify)
Structural type identity is computed with raw `JSON.stringify(t)` in several places:

- `uniteTypes` and `widenAtLoopBackEdge` in `lib/typeChecker/flow.ts` (both carry a documented "does not resolve aliases" caveat)
- `synthLogical` and `synthArray` dedup in `lib/typeChecker/synthesizer.ts`
- assorted `uniqBy(..., JSON.stringify)` calls

The alias-resolution gap is documented, but two more gaps are not:

1. **Property order.** `{a: string, b: number}` and `{b: number, a: string}` stringify differently, so a flow join can keep both as "distinct" union members and unions silently grow imprecise.
2. **Metadata leakage.** `tags` (from `@validate`/`@jsonSchema`), formatter `trivia`, and `isEffectSet` flags all live on the type nodes and poison the key — two semantically identical types with different annotations or comment trivia fail to dedup.

## Proposed work

One canonical `typeKey(t, aliases): string` function:

- resolve aliases (via `safeResolveType`)
- sort object properties by key
- strip non-semantic fields (`tags`, `trivia`, `isEffectSet`, source locations)

then replace every raw `JSON.stringify` identity site with it. One chokepoint fixes the whole class at once, and gives the recursive-types coinduction work (separate issue) a stable pair key to build on. Small, self-contained, high value per effort.

egonSchiele:
Closes #470. Closes #473.

Recursive type aliases used to crash twice: the typechecker stack-overflowed comparing a recursive alias to itself, and the generated module TDZ-crashed at load (`const Tree = z.object({... z.array(Tree)})`). Execution surfaced two more variants of the same family, both fixed here: **plain forward references** (`type A = { b: B }` before `B`) TDZ-crashed too, and **def-before-type** crashed through the module-load tool-definition schema. A fourth, nastier variant: a **validated self-recursive alias silently dropped nested validation** (its descriptor read its own `__agency_descriptor` mid-assignment and got `undefined`) — probe-confirmed, a nested `-5` passed a `@validate(positive)` field before this PR.

## The four mechanisms

1. **`typeKey()`** (#473, `lib/typeChecker/typeKey.ts`) — one canonical structural identity replacing raw `JSON.stringify` at the dedup sites (`uniteTypes`, loop widening, three synthesis sites, block-return dedup). Property-order-insensitive, union-order-insensitive, metadata-stripped (tags/trivia/descriptions — deliberate, argued in the doc comment), `valueArgs` ARE identity, nested alias refs key nominally so recursion terminates.
2. **Coinductive `isAssignable`** — a private `isAssignableGuarded` keeps an in-progress pair stack (public three-arg signature unchanged, mirroring the `resolveType`/`resolveTypeWithGuard` split). Entries are REMOVED on exit, so a refuted pair is never assumed true in a sibling position (a dedicated union-order test pins this against memoizing implementations). The pair key is gated on named references — the only way cycles re-enter — keeping `typeKey` off the structural hot path; the full suite shows no slowdown.
3. **Lazy-if-not-yet-emitted zod refs** — an alias reference inside a module-load-time schema (alias consts AND tool definitions) wraps in `z.lazy(() => Name)` when the target alias is declared at-or-after that source position. Pending-ness is a PURE derivation from declaration order (no mutable emitted-set); `partitionProgram` keeps top-level declarations in source order, so offsets are a faithful oracle. Backward references stay bare: **zero churn across the existing fixture corpus** (verified). `type Loop = Loop` is rejected with a clear error instead of an infinite lazy loop.
4. **Lazy validation descriptors** — a new `{ kind: "ref", get }` descriptor node defers `__agency_descriptor` reads to walk time (the walker's existing `maxDepth` bounds cyclic walks). Use-site validators merge INSIDE `get()` because the walker dispatches `ref` before running validators. `pendingAliases` also threads into descriptor-embedded schema strings.

## Verification

- 7627 lib tests green, structural linter clean, `make` clean.
- New: 9 typeKey unit tests + a wiring pin (proven red when the flow.ts adoption is stashed), 4 coinduction pipeline tests (red-run confirmed as `RangeError` crashes), 5 codegen unit tests (self/forward/mutual/def-before-type/Loop-guard), 3 walker ref tests, 2 committed `$ref` JSON-schema conversion pins (zod is caret-pinned; a zod upgrade that changes cycle handling fails these, not a user request — zod 4 emits `$defs`/`$ref` for both self and mutual cycles).
- 4 agency execution suites (no LLM calls): recursive-type (5 nodes: forward ref, nested parse accept/reject at depth 2, mutual recursion executed, recursive param through a tool schema), recursive-type-validated (the silent-loss pin: nested validator fires through `!` assignment), recursive-type-partial (`Partial<Tree>` composes; the fixture declares `TreePatch` BEFORE `Tree` so the expanded reference is genuinely forward and must wrap), plus a re-run of utility-partial as a regression guard for the merged utility-types feature.
- Codegen fixture pins all reference shapes: forward, self (array position), mutual (forward edge lazy / backward edge bare), union-position self-ref (`type Json = string | number | Json[]`), and `Partial`-expanded pending refs.

## Execution notes (differences from the plan, all recorded in it)

- The plan's pipeline-level typeKey adoption test proved unreachable (assignability is already order-insensitive and the assignment diagnostic renders the declared scope type, not the join) — replaced with a direct `uniteTypes` pin, verified red without the wiring.
- Validators do NOT run through `schema(...).parseJSON` (zod-only; see `lib/runtime/schema.ts`) — the validated-recursive tests use the `!` validated-assignment form, per the plan's fallback note.
- Value-param factories can be called from load-time descriptors anywhere in the module, so their schema strings treat ALL module aliases as pending (always-safe over-inclusion).

## Scope-outs

- `lib/typeChecker/inference.ts` `unionTypes()` keeps `JSON.stringify` (no alias table in its signature) — noted on #473.
- Recursive value-parameterized aliases still hang codegen (pre-existing, unguarded inline expansion) — filed as #484.
- The zod mapper is at six positional params after this PR; the next addition should consolidate into an options object.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -453,6 +453,17 @@ export class TypeScriptBuilder {
   // ------- Main entry point -------
 
   build(program: AgencyProgram): TsNode {
+    // Plain top-level aliases in source order — the basis for the derived
+    // pending-alias computation (see pendingAliasesFor). Captured once;
+    // partitionProgram keeps top-level declarations in source order, so
+    // source offsets are a faithful oracle for const-initialization order.
+    this.aliasEmissionOrder = program.nodes
+      .filter(
+        (n): n is TypeAlias =>
+          n.type === "typeAlias" && !n.typeParams && !n.valueParams,
+      )
+      .map((n) => ({ name: n.aliasName, start: n.loc?.start ?? 0 }));
+
     // Generate tool registry (empty — AgencyFunction.create() populates it)
     this.generatedStatements.push(this.generateToolRegistry());
 
@@ -777,7 +788,22 @@ export class TypeScriptBuilder {
           tags: [...(node.aliasedType.tags ?? []), ...node.tags],
         }
       : node.aliasedType;
-    const zodSchema = this.zodSchemaFor(aliasedWithTags);
+    // `type Loop = Loop` (directly or through a chain landing back on a
+    // bare self-reference) has no base case: its lazy schema would recurse
+    // at first parse. TS rejects the shape too.
+    if (
+      node.aliasedType.type === "typeAliasVariable" &&
+      resolveTypeDeep(node.aliasedType, this.scopes.visibleTypeAliasesFull())
+        .type === "typeAliasVariable"
+    ) {
+      throw new Error(
+        `Type alias '${node.aliasName}' circularly references itself with no structure. Give it an object, array, or union shape.`,
+      );
+    }
+    const zodSchema = this.zodSchemaFor(
+      aliasedWithTags,
+      this.pendingAliasesFor(node.loc?.start),
+    );
     const stmts: TsNode[] = [
       ts.raw(`${exportPrefix}const ${node.aliasName} = ${zodSchema};`),
       ts.raw(
@@ -815,13 +841,39 @@ export class TypeScriptBuilder {
    * like `Container<number>` become a concrete object/array/etc. before the
    * (alias-unaware) zod mapper runs.
    */
-  private zodSchemaFor(t: VariableType): string {
+  private zodSchemaFor(t: VariableType, pendingAliases?: Set<string>): string {
     const aliasesFull = this.scopes.visibleTypeAliasesFull();
     const resolved = resolveTypeDeep(t, aliasesFull);
     return mapTypeToValidationSchema(
       resolved,
       this.scopes.visibleTypeAliases(),
       aliasesFull,
+      pendingAliases,
+    );
+  }
+
+  /** See build() — plain top-level aliases in source order. */
+  private aliasEmissionOrder: { name: string; start: number }[] = [];
+
+  /**
+   * Aliases whose schema const is NOT yet initialized when module-load
+   * code originating at source offset `start` runs: every plain top-level
+   * alias declared at-or-after that offset. A PURE derivation from the
+   * program captured in build() — no mutable emitted-set to keep in sync.
+   * Generic/value-param aliases never emit consts (inlined / hoisted
+   * factory functions) and function-body aliases initialize at call time,
+   * so neither is listed. `start` undefined treats ALL listed aliases as
+   * pending — z.lazy is always safe; over-inclusion only costs
+   * indirection.
+   */
+  private pendingAliasesFor(start: number | undefined): Set<string> {
+    if (start === undefined) {
+      return new Set(this.aliasEmissionOrder.map((a) => a.name));
+    }
+    return new Set(
+      this.aliasEmissionOrder
+        .filter((a) => a.start >= start)
+        .map((a) => a.name),
     );
   }
 
@@ -1716,6 +1768,7 @@ export class TypeScriptBuilder {
    */
   private paramSchemaContribution(
     param: FunctionParameter,
+    pendingAliases: Set<string>,
   ):
     | { kind: "drop" }
     | { kind: "scalar"; zod: string }
@@ -1733,7 +1786,7 @@ export class TypeScriptBuilder {
               type: "primitiveType" as const,
               value: "string",
             });
-      const elementZod = this.zodSchemaFor(elementHint);
+      const elementZod = this.zodSchemaFor(elementHint, pendingAliases);
       return { kind: "array", zod: `z.array(${elementZod})` };
     }
 
@@ -1741,7 +1794,7 @@ export class TypeScriptBuilder {
       type: "primitiveType" as const,
       value: "string",
     };
-    let zod = this.zodSchemaFor(typeHint);
+    let zod = this.zodSchemaFor(typeHint, pendingAliases);
     if (param.defaultValue) {
       const defaultStr = expressionToString(param.defaultValue);
       // A widened optional param (`x?: T` → `T | null`) is already nullable;
@@ -1781,9 +1834,14 @@ export class TypeScriptBuilder {
       );
     }
 
+    // Tool definitions initialize at module load (the __AgencyFunction
+    // const), so alias refs in param schemas need the same pending
+    // treatment alias consts get — a def declared BEFORE a type it
+    // references used to TDZ-crash (probe-confirmed).
+    const pendingAliases = this.pendingAliasesFor(node.loc?.start);
     const contributions = parameters.map((p) => ({
       param: p,
-      contribution: this.paramSchemaContribution(p),
+      contribution: this.paramSchemaContribution(p, pendingAliases),
     }));
 
     // Declaration order is preserved by iterating `parameters` directly;
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -753,10 +753,16 @@ export class TypeScriptBuilder {
       // intact lets `buildValidationDescriptor` emit a nested factory CALL
       // for any value-param reference (validators resolve in their own
       // module); the schema strings still substitute identifiers locally.
+      // Value-param factories HOIST but their bodies execute when called —
+      // possibly from a load-time descriptor initializer ANYWHERE in the
+      // module, so no source position bounds which aliases are safe. Treat
+      // every module alias as pending: z.lazy is always safe and only
+      // costs indirection.
       const vpDescriptor = buildValidationDescriptor(
         vpAliasedWithTags,
         this.scopes.visibleTypeAliases(),
         vpAliasesFull,
+        this.pendingAliasesFor(undefined),
       );
       // Bake value-param DEFAULTS into the factory signature so an omitted
       // use-site arg (e.g. `Age()!` or bare `Age!`) resolves the same default
@@ -821,6 +827,7 @@ export class TypeScriptBuilder {
         resolved,
         this.scopes.visibleTypeAliases(),
         aliasesFull,
+        this.pendingAliasesFor(node.loc?.start),
       );
       // `(Foo as any).__agency_descriptor = ...` — keeps the runtime metadata
       // co-located with the schema and avoids exporting/importing a second
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -95,7 +95,7 @@ import {
   hasAliasValidate,
 } from "./typescriptGenerator/validationDescriptor.js";
 import { tagArgToTs } from "./typescriptGenerator/tagArgToTs.js";
-import { resolveTypeDeep } from "../typeChecker/assignability.js";
+import { resolveTypeDeep, safeResolveType } from "../typeChecker/assignability.js";
 import { isFunctionTyped } from "../typeChecker/utils.js";
 
 import { $, ts } from "../ir/builders.js";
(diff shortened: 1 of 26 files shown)


egonSchiele (commented):
**(Would be REQUEST_CHANGES, but GitHub blocks that on your own PR.)**

## Review

**Overview:** This lands the four recursive-alias mechanisms (canonical `typeKey`, coinductive `isAssignable`, lazy-if-pending zod refs, lazy validation-descriptor refs). I verified the implementation against the plan and the plan review: every review item was genuinely adopted, not just claimed — the derived (stateless) pending-alias computation, the `isAssignableGuarded` private split with the named-ref perf gate, `valueArgs` as identity, the removal-on-exit union-order test, the committed `$ref` conversion pins, the wiring pin for `uniteTypes`, and the validated-recursive silent-loss execution test are all present and correctly constructed. The fixture emission is right in all five reference shapes (forward, self, mutual split, union-position, Partial-expanded).

**One verified regression blocks merge** (inline comment on `typescriptBuilder.ts`): the `type Loop = Loop` guard also rejects legitimate alias-of-alias declarations. Probe-confirmed: `type Coords = { x: number }` + `type Point = Coords` compiles on main and throws `Type alias 'Point' circularly references itself` on this branch. Root cause: the guard tests `resolveTypeDeep(...)`, but `deepResolveNode` deliberately leaves plain non-generic alias refs intact (its own doc comment says so), so ANY bare-alias body — circular or not — stays a `typeAliasVariable` and trips the throw. No fixture or test covers bare alias-of-alias, which is why CI is green.

**Minor points** (details inline):
- `validateChain.ts`: the `ref` walk passes `depth` through unchanged, so a ref-to-ref cycle would loop unbounded past `maxDepth`. Unreachable today only because the compile-time guard rejects circular chains — worth a comment tying the two, and the guard fix must preserve that rejection.
- `synthLogical` calls `ctx.getTypeAliases()` once per union member inside the collect loop — hoist to a local.
- Generated lazy consts have no `: z.ZodType` annotation, so `type Tree = z.infer<typeof Tree>` in the output is circular under tsc. Fine today (output is transpiled, never typechecked), but worth remembering if a typecheck-the-output path ever appears.

Test coverage is otherwise excellent — red-run-confirmed suites, an anti-vacuity rejection test, execution (not just emission) coverage for mutual recursion and tool-schema params, and the silent-validator-loss pin. Security-wise nothing concerning: no new runtime input surfaces beyond deferred descriptor reads, and the depth cap still bounds walks on structural kinds.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -777,7 +794,22 @@ export class TypeScriptBuilder {
...
+    // bare self-reference) has no base case: its lazy schema would recurse
+    // at first parse. TS rejects the shape too.
+    if (
+      node.aliasedType.type === "typeAliasVariable" &&
+      resolveTypeDeep(node.aliasedType, this.scopes.visibleTypeAliasesFull())
+        .type === "typeAliasVariable"
+    ) {
+      throw new Error(
+        `Type alias '${node.aliasName}' circularly references itself with no structure. Give it an object, array, or union shape.`,
+      );
```
egonSchiele: **Blocker — verified regression:** this guard rejects legitimate alias-of-alias declarations, not just circular ones.

```
type Coords = { x: number }
type Point = Coords
```

compiles on main, throws `Type alias 'Point' circularly references itself with no structure` on this branch (probe-run on both). The cause: `resolveTypeDeep` never resolves plain non-generic alias refs — `deepResolveNode` leaves them intact by design ("so codegen can emit them by name"), so *every* bare-alias body lands here as a `typeAliasVariable`, circular or not. No test covers bare alias-of-alias, which is why CI stayed green.

Suggested fix — resolve the alias *chain* with `safeResolveType` (whose in-progress guard is exactly the circularity detector you want) and only throw when the chain lands back on a ref with an existing entry:

```ts
if (node.aliasedType.type === "typeAliasVariable") {
  const chainEnd = safeResolveType(node.aliasedType, this.scopes.visibleTypeAliasesFull());
  if (chainEnd.type === "typeAliasVariable" && this.scopes.visibleTypeAliasesFull()[chainEnd.aliasName]) {
    throw new Error(...);
  }
}
```

- `type Loop = Loop` → guard leaves it nominal, entry exists → throws ✓
- `type A = B` + `type B = A` → resolves around the cycle to a nominal ref, entry exists → throws ✓ (this MUST stay rejected — see my comment on `validateChain.ts`: a two-alias validated cycle would emit `ref → ref` descriptors that loop past `maxDepth` at walk time)
- `type Point = Coords` → resolves to the object body → no throw ✓
- `type X = Undefined` → nominal but entry missing → no throw, falls through to the existing undefined-alias diagnostic instead of a misleading circularity error ✓

Please also add the alias-of-alias case to `recursiveAliases.codegen.test.ts` (including a forward one, `type Point = Coords` before `Coords`, which should emit `const Point = z.lazy(() => Coords)`), and a `type A = B / type B = A` rejection test.

packages/agency-lang/lib/runtime/validateChain.ts
```diff
@@ -144,6 +157,12 @@ async function walk(
   }
   if (isFailure(value)) return value as ResultValue;
 
+  if (descriptor.kind === "ref") {
```
egonSchiele, line 171: The `ref` case passes `depth` through unchanged, so a `ref → ref` chain never increments depth and `maxDepth` cannot bound it — a descriptor cycle built purely from refs would hang. Today this is unreachable: `descriptor()` only emits a ref pointing at another alias descriptor, and a chain of aliases that cycles without intervening structure (`type A = B` + `type B = A`) is rejected at compile time by the Loop guard. That makes the compile-time guard load-bearing for runtime termination — worth stating in this comment, and worth keeping in mind when fixing the guard false-positive (see `typescriptBuilder.ts`). A cheap belt-and-suspenders alternative: count consecutive ref hops and fail past a small constant, which unlinks the two mechanisms entirely.

packages/agency-lang/lib/typeChecker/synthesizer.ts
```diff
@@ -457,9 +458,9 @@ function synthLogical(
   const seen = new Map<string, VariableType>();
   const collect = (t: VariableType) => {
     if (t.type === "unionType") {
-      for (const m of t.types) seen.set(JSON.stringify(m), m);
+      for (const m of t.types) seen.set(typeKey(m, ctx.getTypeAliases()), m);
```
egonSchiele: Nit: `ctx.getTypeAliases()` is called once per union member inside `collect` (and again per call at lines 461/463). Hoist it to a local above the `collect` definition — same as `synthArray`/`synthObject` could, but this is the only one inside a per-member loop.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Fixed in b1f12f96 exactly as you suggested: the guard now resolves the alias CHAIN with `safeResolveType` and throws only when the chain lands back on a ref with a registry entry. All four of your cases are covered by tests in `recursiveAliases.codegen.test.ts`: `type Point = Coords` compiles again (regression test with your probe program, asserting `const Point = Coords`), the FORWARD alias-of-alias emits `const Point = z.lazy(() => Coords)`, `type A = B / type B = A` stays rejected (with a comment noting why it must — the ref -> ref descriptor point from your validateChain comment), and `type Loop = Loop` still throws. Unknown aliases fall through to the undefined-alias diagnostic as you described.

You were right that no test covered bare alias-of-alias — that is how the false positive survived a green CI. Both directions are pinned now.

packages/agency-lang/lib/runtime/validateChain.ts
egonSchiele: Took the belt-and-suspenders option in b1f12f96: `walk` now threads a consecutive-ref-hop counter (reset by every structural kind, capped at 8) and fails with a clear reason instead of hanging, so runtime termination no longer depends on the compile-time guard staying airtight. The load-bearing relationship you described is stated in the `MAX_CONSECUTIVE_REF_HOPS` doc comment (pointing at the processTypeAlias guard), and a unit test builds a pure `ref -> ref` cycle and asserts it fails on the cap.

packages/agency-lang/lib/typeChecker/synthesizer.ts
egonSchiele: Hoisted in b1f12f96 — and since `synthArray` and the block-return dedup have the same lookup inside their per-item loops, hoisted those two as well.

egonSchiele (commented):
## Re-review of b1f12f96

**All three findings from the previous review are resolved — verified locally, not just by reading the diff.** (This would be an APPROVE, but GitHub blocks that on your own PR.)

1. **Circularity-guard false positive (was the blocker): fixed and pinned.** The guard now resolves the alias *chain* with `safeResolveType` and throws only when the chain lands back on a ref with a registry entry. I re-ran my alias-of-alias probe (`type Coords = {x: number}` + `type Point = Coords`) on this branch: it now compiles, where b1012c90 threw. The three new codegen tests cover exactly the right shapes — legit backward alias-of-alias emits `const Point = Coords`, forward alias-of-alias emits `const Point = z.lazy(() => Coords)`, and the bare two-alias cycle (`type A = B` / `type B = A`) stays rejected, which keeps the compile-time guard aligned with the runtime ref-walk. The unknown-alias case correctly falls through to the regular undefined-alias diagnostic (no entry, no throw).

2. **Ref-hop cap: better than what I asked for.** `MAX_CONSECUTIVE_REF_HOPS = 8` with reset-on-structural-kind unlinks runtime termination from the compile-time guard entirely, and the new pure `ref -> ref` cycle test pins it. The failure payload matches the existing `maxDepth` failure shape (`reason`/`limit`/`valuePreview`). One residual, acceptable as-is: a legitimate chain of MORE than 8 consecutive validated bare-alias hops (`type A1 = A2; ... type A9 = {...}`, all `@validate`d) would false-fail — vanishingly unlikely, the constant is documented, and the failure reason names the limit, so no change requested.

3. **`getTypeAliases()` hoisted** at all three synthesizer sites, including the per-member loop that motivated the nit.

Local verification: the 5 touched/new test files (38 tests) all pass on b1f12f96, including my previously-failing probe. Lint passes; agency-tests and build CI were still pending at review time — worth waiting for the agency execution suites before merge since the fix commit touched codegen, but nothing in the diff suggests risk there (the guard change only affects bare-alias bodies, and the fixture corpus is pinned by the zero-churn check).

Good to merge once CI is green.

PR #488, opened 2026-07-10
Quick wins: value-param recursion guard, pipeline ordering assertions, Anthropic schema-limitation guidance (closes #475, #484, #487)

### LINKED ISSUE #475: check() pipeline: assert load-bearing pass-ordering constraints, fix phase numbering
`TypeChecker.check()` (`lib/typeChecker/index.ts`) is an ordered script of ~20 passes whose ordering constraints are load-bearing but enforced only by comments:

- "MUST run before checkScopes so `e.data` usage sites narrow correctly" (3c, `refineInlineHandlerParams`)
- "Build the flow graph AFTER the param retype, so its typeAt oracle is seeded with the refined `e`" (3d)
- "Runs AFTER buildFlowGraphs so yield synthesis sees flow-narrowed bindings, and before checkScopes" (3e, `computeMatchExprTypes`)

Nothing fails loudly if a future refactor reorders them — the checker just silently produces stale or missing diagnostics.

Also, the phase numbering in the comments has drifted: 1, 1b, 1c, 1d, 1f (no 1e), 3a-3e, then 9a appears *before* 9.

## Proposed work

Deliberately light-touch — a full pass-manager with dependency topsort is overkill at this scale:

1. Turn the two or three critical MUST constraints into cheap runtime assertions. Example: `refineInlineHandlerParams` throws if `ctx.flowEnv` is already set; `computeMatchExprTypes` throws if it is not.
2. Fix the phase numbering (or drop the numbers in favor of named section comments, which cannot drift).
3. Keep the rest as prose — the comments are good; they just should not be the only enforcement for the constraints that silently corrupt results when violated.

### LINKED ISSUE #484: Recursive value-parameterized alias hangs codegen (unguarded inline expansion)

### LINKED ISSUE #487: Recursive types as structured-output contracts fail on Anthropic (provider limitation)
Empirically probed (2026-07-09, live API calls) while gating the tuple-types feature: recursive type aliases used as LLM structured-output contracts (`const t: Tree = llm(...)`) work on OpenAI and Gemini but fail on Anthropic.

| Provider | Recursive `$ref`/`$defs` schema |
|---|---|
| OpenAI strict `json_schema` | 200 — valid nested output |
| Gemini `responseJsonSchema` | 200 — valid nested output |
| Anthropic structured outputs (`structured-outputs-2025-11-13`) | 400 — "Circular reference detected in schema definitions: __schema0 -> __schema0. Self-referencing or mutually-referencing definitions are not supported." |

PR #485 made recursive aliases work end to end (typecheck, codegen, `schema(...).parseJSON`, validation) and pinned that zod emits `$ref` for them — but provider acceptance varies. Two follow-ups:

1. Docs: `docs/site/guide/types.md`'s recursive-types section should note the Anthropic limitation for the structured-output use specifically (parsing/validation are unaffected).
2. Consider surfacing the provider 400 as a friendlier error at the prompt site ("this provider does not support recursive structured-output schemas — parse the response with schema(...).parseJSON instead").

egonSchiele:
Closes #475. Closes #484. Closes #487.

Three small independent fixes, one commit each.

## #484 — recursive value-parameterized aliases no longer blow the stack

`type Weird(n: number) = { next: Weird(n) }` used to crash with a bare `RangeError: Maximum call stack size exceeded` when the instantiation was USED (value-param instantiations are inlined by the zod mapper and descriptor builder — unlike plain aliases, they have no named const to defer to with `z.lazy`, so a cycle has no representation). Now rejected at DECLARATION with an error naming the cycle (`Weird -> Weird`, or `A -> B -> A` for mutual cycles). Non-recursive value-param chains (`Person(low)` referencing `Age(low)`) still work — pinned by test. One nuance found in execution: the crash needs a use site; a bare declaration emits nothing, which is why the test programs include one.

## #475 — pipeline ordering assertions + comment numbering

The two load-bearing "MUST run before/after buildFlowGraphs" ordering constraints in `TypeChecker.check()` are now runtime assertions: `refineInlineHandlerParams` throws if `ctx.flowEnv` is already set, `computeMatchExprTypes` throws if it is not — each with a comment explaining what silently breaks on reorder (stale memo types / un-narrowed match yields). Both assertion paths are unit-tested. The drifted phase numbering (1, 1b-1f with no 1e, 9a before 9) is dropped entirely in favor of the descriptive text — numbers only drift; the ordering contract now lives in prose plus the assertions. No cross-references to the numbers existed outside the file (grep-verified).

## #487 — Anthropic recursive-schema limitation: guidance instead of zod internals

Live-probed (all three providers): recursive `$ref` structured-output schemas work on OpenAI and Gemini, but Anthropic 400s with `Circular reference detected in schema definitions: __schema0 -> __schema0` — naming zod-internal identifiers, not the user's type. The terminal-error path in the prompt retry ladder now rethrows that class of error with actionable guidance (declare the result a string and `schema(T).parseJSON` it, or use a provider that supports recursion), preserving the original provider text for debugging. Pure helper (`enrichSchemaLimitationError` in `llmRetry.ts`) with unit tests; the guide's recursive-types section documents the limitation in three sentences.

## Verification

7658 lib tests green, structural linter clean, `make` clean. (One unrelated flake observed and cleared: `topsortCycleErrors.test.ts` timed out once in the full-suite run, passes in isolation — subprocess-isolation timing, untouched by this PR.)

### DIFF
--- a/packages/agency-lang/docs/site/guide/types.md
+++ b/packages/agency-lang/docs/site/guide/types.md
@@ -219,11 +219,9 @@ type Tree = {
 annotations on nested fields fire at every depth through the `!`
 validated-assignment form.
 
-One provider limitation: recursive types work as LLM structured-output
-contracts (`const t: Tree = llm(...)`) on OpenAI and Gemini, but
-Anthropic's structured-output API rejects self-referencing schemas. On
-Anthropic, ask for a string and parse it yourself with
-`schema(Tree).parseJSON(...)` — parsing and validation are unaffected.
+Anthropic does not support recursive types as LLM output contracts
+(OpenAI and Gemini do). On Anthropic, parse the response yourself with
+`schema(Tree).parseJSON(...)`.
 
 ## References
 
--- a/packages/agency-lang/docs/site/guide/types.md
+++ b/packages/agency-lang/docs/site/guide/types.md
@@ -219,6 +219,12 @@ type Tree = {
 annotations on nested fields fire at every depth through the `!`
 validated-assignment form.
 
+One provider limitation: recursive types work as LLM structured-output
+contracts (`const t: Tree = llm(...)`) on OpenAI and Gemini, but
+Anthropic's structured-output API rejects self-referencing schemas. On
+Anthropic, ask for a string and parse it yourself with
+`schema(Tree).parseJSON(...)` — parsing and validation are unaffected.
+
 ## References
 
 - [Schemas](/guide/schemas)
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -96,6 +96,7 @@ import {
 } from "./typescriptGenerator/validationDescriptor.js";
 import { tagArgToTs } from "./typescriptGenerator/tagArgToTs.js";
 import { resolveTypeDeep, safeResolveType } from "../typeChecker/assignability.js";
+import { visitTypes } from "../typeChecker/typeWalker.js";
 import { isFunctionTyped } from "../typeChecker/utils.js";
 
 import { $, ts } from "../ir/builders.js";
@@ -707,7 +708,44 @@ export class TypeScriptBuilder {
     return out;
   }
 
+  /**
+   * Reject value-parameterized aliases whose bodies cycle back to a
+   * value-parameterized alias already on the walk (directly or mutually).
+   * Their instantiations are INLINED by the zod mapper and the descriptor
+   * builder — unlike plain aliases, which emit named consts and defer
+   * cycles with z.lazy — so a cycle has no representation: expansion
+   * recurses until the stack blows (#484). Detected at declaration so the
+   * error names the alias instead of surfacing as a RangeError at some
+   * use site.
+   */
+  private rejectValueParamCycle(node: TypeAlias): void {
+    const aliasesFull = this.scopes.visibleTypeAliasesFull();
+    const walk = (body: VariableType, seen: string[]): void => {
+      visitTypes(body, (t) => {
+        const refName =
+          t.type === "typeAliasVariable"
+            ? t.aliasName
+            : t.type === "genericType"
+              ? t.name
+              : undefined;
+        if (refName === undefined) return;
+        const entry = aliasesFull[refName];
+        if (!entry?.valueParams) return;
+        if (seen.includes(refName)) {
+          throw new Error(
+            `Type alias '${node.aliasName}' is a recursive value-parameterized alias (cycle: ${[...seen, refName].join(" -> ")}). Value-param instantiations are inlined into their use sites and cannot recurse; use a plain recursive alias instead.`,
+          );
+        }
+        walk(entry.body, [...seen, refName]);
+      });
+    };
+    walk(node.aliasedType, [node.aliasName]);
+  }
+
   private processTypeAlias(node: TypeAlias): TsNode {
+    if (node.valueParams) {
+      this.rejectValueParamCycle(node);
+    }
     // Generic aliases (type Container<T> = ...) can't be turned into a single
     // zod schema because the body references type parameters that have no
     // runtime value. Usages like `Container<number>` are normalized away by
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -96,7 +96,7 @@ import {
 } from "./typescriptGenerator/validationDescriptor.js";
 import { tagArgToTs } from "./typescriptGenerator/tagArgToTs.js";
 import { resolveTypeDeep, safeResolveType } from "../typeChecker/assignability.js";
-import { visitTypes } from "../typeChecker/typeWalker.js";
+import { rejectValueParamCycle } from "./valueParamCycle.js";
 import { isFunctionTyped } from "../typeChecker/utils.js";
 
 import { $, ts } from "../ir/builders.js";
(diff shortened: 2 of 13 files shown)


packages/agency-lang/docs/site/guide/types.md
```diff
@@ -219,6 +219,12 @@ type Tree = {
 annotations on nested fields fire at every depth through the `!`
 validated-assignment form.
 
+One provider limitation: recursive types work as LLM structured-output
```
egonSchiele: make this shorter and easier to read.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -707,7 +708,44 @@ export class TypeScriptBuilder {
...
+   * Reject value-parameterized aliases whose bodies cycle back to a
+   * value-parameterized alias already on the walk (directly or mutually).
+   * Their instantiations are INLINED by the zod mapper and the descriptor
+   * builder — unlike plain aliases, which emit named consts and defer
+   * cycles with z.lazy — so a cycle has no representation: expansion
+   * recurses until the stack blows (#484). Detected at declaration so the
+   * error names the alias instead of surfacing as a RangeError at some
+   * use site.
+   */
+  private rejectValueParamCycle(node: TypeAlias): void {
```
egonSchiele: put in its own file, don't bloat this file any more

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -707,7 +708,44 @@ export class TypeScriptBuilder {
...
+    const aliasesFull = this.scopes.visibleTypeAliasesFull();
+    const walk = (body: VariableType, seen: string[]): void => {
```
egonSchiele: does this cover all the cases? What about types declared in nodes/functions? Or types declared in global scope?

packages/agency-lang/docs/site/guide/types.md
egonSchiele: Shortened in 1f5a5338 — three lines: the limitation, who supports it, the workaround.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Extracted in 1f5a5338 to `lib/backends/valueParamCycle.ts` as a pure function taking the alias registry — the builder call site is now one line, and the file also carries the boundary rationale (why the walk follows only value-param entries).

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Probe-verified all three scopes in 1f5a5338, and each is now a pinned test in `recursiveAliases.codegen.test.ts`:

- **Global (top-level)**: covered by the original direct + mutual tests.
- **Function/node bodies**: the guard FIRES — body aliases hoist through the same `processTypeAlias` path, and `visibleTypeAliasesFull()` sees scoped entries. New test pins a recursive `type W(n)` inside a `def`.
- **Combined generic+value-param** (`type Foo<T>(n) = { next: Foo<T>(n) }`): also caught (the walk handles the `genericType` reference shape). New test.

One boundary the probing surfaced, also pinned: a value-param alias cycling through a PLAIN alias (`Weird(n) -> Mid -> Weird(2)`) correctly COMPILES — the plain alias emits a named const and is referenced by name (z.lazy if forward), which breaks the inline cycle naturally. That is why the walk deliberately follows only value-param entries; the test exists so nobody "fixes" the guard into over-rejecting.

packages/agency-lang/lib/runtime/llmRetry.ts
```diff
@@ -215,3 +215,25 @@ export function resolveRetryPolicy(opts: RetryConfig, branchDefaults: RetryConfi
...
+  if (!(err instanceof Error)) return null;
+  if (!/Circular reference detected in schema definitions/.test(err.message)) {
+    return null;
+  }
+  return new Error(
+    `This provider does not support recursive types as structured-output contracts. ` +
+      `Parse the response yourself instead: declare the result as a string and run ` +
+      `schema(YourType).parseJSON(...) on it, or switch this call to a provider that ` +
+      `supports recursive schemas (OpenAI, Gemini). Provider error: ${err.message}`,
+  );
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 1f5a5338 — the original instance is preserved with its message rewritten in place, so provider metadata (`SmolError.status` etc.) and the original stack survive. Tests assert same-instance identity and that an attached `status` field rides through. Also added the wiring-level test this thread implied: `runWithRetry` with a dispatch throwing the circular-schema 400 must surface the ENRICHED message (verified red when the terminal-branch wiring is reverted to a bare `throw err`).

PR #489, opened 2026-07-10
Failure propagation at call boundaries — Stage 1 (mechanism, default warn)

egonSchiele:
## Failure propagation at call boundaries — Stage 1 (mechanism, default `warn`)

A failure Result that flows into a function not typed to accept Results should not produce a confusing `split is not a function` deep inside the callee. This PR lands the full mechanism behind a config knob, **shipping with the default at `"warn"` so nothing changes behavior yet**. A follow-up PR flips the default to `"on"`.

Spec: `docs/superpowers/specs/2026-07-08-failure-propagation-design.md` (included). Plan: `docs/superpowers/plans/2026-07-08-failure-propagation.md` (Tasks 1–9 of 12; Tasks 10–12 are the flip PR).

### What the mechanism does (when `"on"`)

- A failure passed to a param whose declared type does not accept Results (`Result`, explicit `any`, or a union containing either) **skips the callee and returns the original failure**, appending to a new `skippedFunctions: {name, param}[]` trail on the failure. Pipe-style short-circuiting, generalized to every call.
- A failure passed to an untagged plain TypeScript function **throws** with a message naming the producer; `acceptsFailures(fn)` is the exemption tag (runtime pre-tags `isSuccess`/`isFailure`/`success`/`failure` for aliased use, plus `_print`/`_printJSON`).
- A method call on a Result (`f.split(...)`, `s.toFixed(...)`) **throws** with the producer and a `.value` hint for successes; `r.value()` on a function-wrapping success still works.
- All throws are plain `Error`s: the auto-try converts them into catchable failures of the enclosing function. Never `AgencyAbort`.

### What `"warn"` (the shipped default) does

Every would-be skip/throw logs a statelog `warn` event (new event kind) and echoes one `console.warn` line to stderr, then behaves exactly as today. This PR's CI run plus real-world warn output is the corpus measurement for the flip.

### Deliberate boundaries (see spec "What this does not catch")

- The TS-function argument scan lives in `__call` only — method arguments are never scanned, so `arr.push(someFailure)` keeps working (pinned by unit test).
- Block params and node-transition params fail open in v1.
- The check is shallow: failures inside arrays/objects pass through.

### Plumbing

- `FuncParam.acceptsResult` emitted by codegen from the new `paramAcceptsFailure` predicate (`isFunctionTyped` pattern; legacy params fail open). Fixtures regenerated (+`acceptsResult` line per param, no other shape change).
- `failurePropagation: "off" | "warn" | "on"` config knob plumbed like `maxCallDepth` (config → baked runtimeCtxArgs → ExecutionContext → subprocess override inheritance).
- Typechecker knows `skippedFunctions` on the narrowed failure branch (with a misspelled-member negative control test).
- `truncate` extracted from `ipc.ts` into a leaf module to avoid dragging subprocess machinery (and an import cycle) into the call hot path.
- stdlib `print` variadic annotated `any[]`; generated stdlib docs regenerated.

### Testing

- 40 unit tests for the module + invoke + dispatcher paths (modes, census warn semantics, leftmost-rejecting-param, `.partial()`-bound failures, variadic elements, UNSET slots, own-field callables).
- Three agency-js end-to-end tests, one per mode: `off` = legacy, `warn` = legacy + statelog warn event, `on` = strict skip. The `on` test proves the strict path end to end before any default flips.
- Full unit suite green locally (7717 passed); the two `result.test.ts` exact-shape assertions were extended with the new field.

One CI caveat: warn mode adds stderr lines wherever the corpus already passes failures around. Output comparison ignores stderr, but if a PTY/full-terminal assertion fails, suspect stderr capture before the mechanism.

### DIFF
--- a/packages/agency-lang/lib/runtime/failurePropagation.ts
+++ b/packages/agency-lang/lib/runtime/failurePropagation.ts
@@ -52,24 +52,49 @@ export function isFailureTolerant(fn: unknown): boolean {
   return (fn as any)[ACCEPTS_FAILURES] === true;
 }
 
-// These four are DIRECT_CALL_FUNCTIONS (nameClassifier.ts), so by-name
-// calls never reach the dispatcher. Tag them anyway: aliased and
-// higher-order uses (`const f = isFailure; f(x)`, `const mk = failure;
-// mk(inner)`) route through __call like any other value.
-acceptsFailures(isSuccess);
-acceptsFailures(isFailure);
-acceptsFailures(success);
-acceptsFailures(failure);
+/**
+ * SINGLE SOURCE OF TRUTH for runtime-layer builtins that legitimately
+ * receive failure values. Add new entries HERE, never as scattered
+ * acceptsFailures() calls.
+ *
+ * Every entry must be on the compiler's DIRECT_CALL_FUNCTIONS list
+ * (lib/backends/typescriptBuilder/nameClassifier.ts) — by-name calls to
+ * those bypass the dispatcher entirely, so this list only matters for
+ * ALIASED use (`const f = isFailure; f(x)`). A tripwire test in
+ * failurePropagation.test.ts enforces that membership, so the two lists
+ * cannot drift silently.
+ *
+ * Two deliberate non-members:
+ * - stdlib TS helpers (`_print`, `_printJSON`) have their own list in
+ *   lib/stdlib/builtins.ts, because the runtime hot path must not import
+ *   the stdlib graph.
+ * - JSON.stringify is tolerated by IDENTITY inside isFailureTolerant; it
+ *   is a native, not an Agency builtin.
+ */
+export const FAILURE_TOLERANT_BUILTINS: ReadonlyArray<(...args: any[]) => any> = [
+  isSuccess,
+  isFailure,
+  success,
+  failure,
+];
+for (const fn of FAILURE_TOLERANT_BUILTINS) {
+  acceptsFailures(fn);
+}
 
 function origin(f: ResultFailure): string {
   return f.functionName ?? "(unknown)";
 }
 
-function logWarn(mode: FailurePropagationMode, message: string, detail: {
-  functionName?: string;
-  param?: string;
-  error?: unknown;
-}): void {
+function logWarn(
+  mode: FailurePropagationMode,
+  message: string,
+  detail: {
+    functionName?: string;
+    param?: string;
+    error?: unknown;
+  },
+  consoleLine: string,
+): void {
   const ctx = agencyStore.getStore()?.ctx;
   // Fire-and-forget, like handlerDecision in interrupts.ts. Optional-chained
   // end to end: unit tests and mock contexts may lack a statelog client.
@@ -82,10 +107,12 @@ function logWarn(mode: FailurePropagationMode, message: string, detail: {
   });
   // Warn mode exists to be SEEN: without observability config the statelog
   // event goes nowhere, so echo to stderr. "on" mode stays quiet here — its
-  // skip/throw is the signal. (statelogClient itself uses console.warn, so
-  // this is an allowed pattern.)
+  // skip/throw is the signal. The console line is PAYLOAD-FREE on purpose:
+  // console output never passes the statelog redaction chokepoint, and a
+  // failure's error often wraps upstream response bodies — exactly what
+  // redact() tags exist to scrub. Full detail goes to statelog only.
   if (mode === "warn") {
-    console.warn(`failurePropagation: ${message}`);
+    console.warn(`failurePropagation: ${consoleLine}`);
   }
 }
 
@@ -129,10 +156,14 @@ export function checkFailureArgs(
     if (hit === undefined) {
       continue;
     }
+    // In warn mode nothing is actually skipped — say so, or the telemetry
+    // reads as if the legacy behavior already changed.
+    const verb = mode === "warn" ? "would be skipped" : "skipped";
     logWarn(
       mode,
-      `call to '${fnName}' skipped: parameter '${param.name}' received a failure produced by '${origin(hit)}' (${truncate(hit.error)})`,
+      `call to '${fnName}' ${verb}: parameter '${param.name}' received a failure produced by '${origin(hit)}' (${truncate(hit.error)})`,
       { functionName: fnName, param: param.name, error: hit.error },
+      `call to '${fnName}' ${verb}: parameter '${param.name}' received a failure produced by '${origin(hit)}' — enable observability for detail`,
     );
     if (mode === "warn") {
       // Census semantics: warn mode exists to MEASURE, so log every
@@ -167,7 +198,12 @@ export function checkTsFunctionArgs(
       `'${fnName}' received a failure produced by '${origin(arg)}' (${truncate(arg.error)}). ` +
       `TypeScript functions cannot receive failures. Check the Result before passing it, ` +
       `or tag the function with acceptsFailures().`;
-    logWarn(mode, message, { functionName: fnName, error: arg.error });
+    logWarn(
+      mode,
+      message,
+      { functionName: fnName, error: arg.error },
+      `'${fnName}' received a failure produced by '${origin(arg)}' — TypeScript functions cannot receive failures; tag with acceptsFailures() to allow`,
+    );
     if (mode === "on") {
       throw new Error(message);
     }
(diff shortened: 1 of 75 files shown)


packages/agency-lang/lib/runtime/failurePropagation.ts
```diff
@@ -0,0 +1,223 @@
...
+    return true;
+  }
+  return (fn as any)[ACCEPTS_FAILURES] === true;
+}
+
+// These four are DIRECT_CALL_FUNCTIONS (nameClassifier.ts), so by-name
+// calls never reach the dispatcher. Tag them anyway: aliased and
+// higher-order uses (`const f = isFailure; f(x)`, `const mk = failure;
+// mk(inner)`) route through __call like any other value.
+acceptsFailures(isSuccess);
```
egonSchiele: how will we remember to keep this up to date?

packages/agency-lang/lib/statelogClient.ts
```diff
@@ -1242,6 +1242,32 @@ export class StatelogClient {
     });
   }
 
+  /** Structured warning event. First consumer: the failure-propagation
+   *  check (warnType "failurePropagation"), which logs every skipped call
+   *  and every would-be throw in "warn" mode, and every skip in "on" mode. */
+  async warn({
```
egonSchiele, line 1255: **Redaction bypass: the `error` field rides outside the redaction chokepoint.**

`post()`'s redaction replacer is deliberately scoped to the `data` payload only (see the comment at the chokepoint: "Redaction is a JSON.stringify replacer applied to the `data` payload ONLY"). This new event posts `error` — an arbitrary user value, the failure's error payload — as a top-level field, so a `redact()`-tagged secret carried in a failure's error lands in statelog unscrubbed. The `std::tag` redaction feature exists precisely to keep secrets out of statelog, and this is a new event kind that sidesteps it.

Suggestion: nest the variable payload under `data` so the existing chokepoint covers it:

```ts
await this.post({
  type: "warn",
  warnType,
  message,
  data: { functionName, param, error },
});
```

Two related exposures worth a conscious decision rather than a fix-by-default:
- `message` embeds `truncate(error)` as a string. That matches the existing `error()` event's exposure (its `message` embeds error text), so it is arguably fine — but it means nesting `error` under `data` alone does not fully close the gap.
- Under Stage 1's default `warn` mode, `logWarn` also echoes the same preview to stderr via `console.warn`, where no redaction ever applies (comment on that line separately).

packages/agency-lang/lib/runtime/failurePropagation.ts
```diff
@@ -0,0 +1,223 @@
...
+    functionName: detail.functionName,
+    param: detail.param,
+    error: detail.error,
+  });
+  // Warn mode exists to be SEEN: without observability config the statelog
+  // event goes nowhere, so echo to stderr. "on" mode stays quiet here — its
+  // skip/throw is the signal. (statelogClient itself uses console.warn, so
+  // this is an allowed pattern.)
+  if (mode === "warn") {
+    console.warn(`failurePropagation: ${message}`);
```
egonSchiele: **Default-warn mode auto-echoes failure error previews to stderr, unredacted.**

The message this prints embeds `truncate(hit.error)` — the raw failure payload preview. Since Stage 1 ships with the default at `"warn"`, every production agent that passes failures across rejecting params will echo failure error contents to stderr automatically, with no opt-in and no redaction (console output never goes through the statelog redaction chokepoint, and `redact()`-tagged values get no protection here).

A failure's `error` frequently wraps upstream response payloads (HTTP bodies, tool output), which is exactly the kind of data `std::tag` redaction exists to scrub.

Suggestion: keep the echo (warn mode does exist to be seen) but drop the error preview from the stderr line — function name + param is enough to locate the site, and the full detail is available via statelog with observability on:

```ts
console.warn(`failurePropagation: call to '${fnName}' would be skipped: parameter '${param.name}' received a failure — run with observability for detail`);
```

(That wording lives at the call sites; alternatively pass a preview-free short message into `logWarn` for the console line.)

egonSchiele (commented):
## Review

Reviewed against the spec and plan on the branch (both included in this PR), with every implementation claim checked against the codebase.

### Overview

Stage 1 of failure propagation: the full check mechanism (invoke-chokepoint param scan, `__call` TS-function scan, `__callMethod` Result-method check), `skippedFunctions` trail, `acceptsResult` codegen flag, `failurePropagation` config knob plumbed like `maxCallDepth`, statelog `warn` event kind — shipping with the `ExecutionContext` default at `"warn"` so behavior is unchanged until the Stage 2 flip.

### Verified

- All plan-review findings landed correctly: the TS-function argument scan is in `__call` only (`arr.push(someFailure)` pinned by unit test), block/node params fail open as documented, `truncate` extracted to a leaf module (no `ipc.ts` import cycle, confirmed `agencyFunction → failurePropagation → truncate` is cycle-free), census warn semantics pinned, the two-defaults design (`?? "on"` frameless fallback vs `?? "warn"` context default) is coherent and correctly commented at both sites, `config.ts` docstring states the staged default accurately.
- Fixture regen is exactly the expected shape: 66 `acceptsResult` lines and their comma-line neighbors, nothing else.
- The three agency-js mode tests give real end-to-end coverage of `off`/`warn`/`on` (config → baked runtimeCtxArgs → context → mode read), including the statelog event assertion (`"type":"warn"`) and the strict-path proof before any default flips.
- 40 unit tests cover the mode matrix, leftmost-rejecting-param semantics, `.partial()`-bound failures, variadic elements, UNSET slots, own-field callables, and both fail-open paths.
- Statelog readers filter by event type, so the new `"warn"` kind is ignored by existing tooling rather than breaking it.

### Findings

Two inline comments, both in the same area: the new statelog `warn` event carries the failure's raw `error` value **outside** the `data`-scoped redaction chokepoint, and warn mode's stderr echo prints an error preview that no redaction ever touches. Given `std::tag` redaction exists specifically to keep secrets out of statelog — and Stage 1 turns the echo on by default everywhere — these are worth fixing (or explicitly deciding) before merge. Neither affects the `"on"`-mode mechanism itself.

Everything else looks ready. Nice staging structure — the flip PR being a one-literal diff with the corpus run as its CI is exactly the `matchExhaustiveness` playbook.

packages/agency-lang/lib/stdlib/builtins.ts
```diff
@@ -23,6 +24,11 @@ export function _printJSON(obj: any): void {
   console.log(JSON.stringify(obj, null, 2));
 }
 
+// Printing a failure is a legitimate debugging move; without the tag the
+// dispatcher's failure-argument check would reject these plain TS helpers.
+acceptsFailures(_print);
```
egonSchiele: This is awful ... how will we know all the functions that accept failures? How will we keep the list up to date? You just have random function calls everywhere! Make an array (or better yet, reuse an existing one), keep it one place, make a note somewhere to keep it up to date!

egonSchiele: Does the typechecker have data about builtin funcs you can hook into?

egonSchiele:
All review comments addressed in 9dce4ef4a.

**Scattered `acceptsFailures()` calls → two documented lists + a tripwire test.** The runtime builtins now live in one exported array, `FAILURE_TOLERANT_BUILTINS` in `failurePropagation.ts`; the stdlib helpers live in `FAILURE_TOLERANT_STDLIB_HELPERS` in `stdlib/builtins.ts`. Two lists rather than one because the runtime hot path must not import the stdlib graph — each list's comment points at the other, and says "add here, never scattered calls."

On "how will we remember" and "does the typechecker have data to hook into": yes — the compiler's `DIRECT_CALL_FUNCTIONS` list in `nameClassifier.ts` is exactly the relevant data, and it is now exported. A new tripwire test asserts every entry of `FAILURE_TOLERANT_BUILTINS` is on that list, so the two cannot drift silently. (By-name calls to those functions bypass the dispatcher entirely; the runtime list only matters for aliased use like `const f = isFailure; f(x)`.)

**Redaction: statelog `error` now rides inside `data`.** `warn()` posts `{type, warnType, message, data: {functionName, param, error}}`, so the data-scoped redaction replacer scrubs tagged secrets. Conscious call on `message`: it keeps the truncated error preview, matching the exposure of the existing `error()` event — documented in the method's docstring.

**Redaction: the warn-mode stderr echo is now payload-free.** It prints function + param + producer name only, with "enable observability for detail"; the full preview goes to statelog only. A test pins that the echoed line does not contain the failure payload.

**Warn-mode wording**: the message now says "would be skipped" in warn mode ("skipped" only in on mode), and a test asserts it.

**`truncate()` throwing on BigInt/circular**: now falls back to `String(val)`, with tests for both.

**Spec status header**: updated to "Stage 1 implemented on this branch; Stage 2 flip is a follow-up PR."

PR #490, opened 2026-07-10
Validate spawn cwd exists before running, with a clear error

egonSchiele:
## What

`bash` / `exec` passed the resolved `cwd` straight to `child_process.spawn`. When that directory didn't exist, Node threw `spawn <cmd> ENOENT` — which reads as "the executable wasn't found" and gives an LLM agent nothing to act on.

It fires whenever the agent points `cwd` at a directory it hasn't created yet, e.g.:

```
setAgentCwd("/tmp/build")            # or bash(cwd: "/tmp/build")
bash("mkdir -p /tmp/build && ...")   # spawn sh with cwd=/tmp/build → ENOENT
```

A chicken-and-egg: the shell can't even start in a directory that doesn't exist, so it can't run the `mkdir` that would create it.

## Fix

Add `resolveSpawnCwd`, shared by `execImpl` and `bashImpl`. It resolves the cwd (as before, via `resolveDir` — `~` expansion + allow-list) then `fs.stat`s it before spawning:

- missing → `Working directory does not exist: <path>. Create it first (e.g. \`mkdir -p\` from an existing directory) before running a command there.`
- a file → `Working directory is not a directory: <path>.`
- exists → unchanged behavior.

The clear message lets the agent recover (create the directory, retry) instead of getting a cryptic Failure.

## Why

Surfaced while benchmarking the agency agent on Terminal-Bench: it intermittently failed **compile-compcert**, **cancel-async-tasks**, **portfolio-optimization**, and **winning-avg-corewars** with `spawn sh ENOENT`.

## Tests

`shell.test.ts` — nonexistent cwd (both `bash` and `exec`) → "does not exist"; a file-as-cwd → "is not a directory"; and an existing cwd still runs normally (no regression). All 9 shell tests pass.

### DIFF
--- a/packages/agency-lang/lib/stdlib/shell.test.ts
+++ b/packages/agency-lang/lib/stdlib/shell.test.ts
@@ -139,3 +139,54 @@ describe("_exec / _bash cwd ~ expansion", () => {
     expect(result.stdout.trim()).not.toBe(fs.realpathSync(fakeHome));
   });
 });
+
+/**
+ * A missing spawn `cwd` used to surface as Node's cryptic
+ * `spawn <cmd> ENOENT`. `resolveSpawnCwd` now validates existence first
+ * and throws a clear, actionable message so an LLM agent can recover
+ * (create the directory, then retry). Regression target: the agent
+ * doing `setAgentCwd("/tmp/build")` before the `mkdir`.
+ */
+describe("_exec / _bash reject a nonexistent cwd with a clear error", () => {
+  let tmp: string;
+  beforeEach(() => {
+    tmp = fs.mkdtempSync(path.join(os.tmpdir(), "shell-cwd-"));
+  });
+  afterEach(() => {
+    fs.rmSync(tmp, { recursive: true, force: true });
+  });
+
+  it("_bash throws 'does not exist', not 'spawn sh ENOENT'", async () => {
+    const ctx = makeMockCtx();
+    const missing = path.join(tmp, "not-created-yet");
+    await expect(
+      __internal_bash(ctx, new StateStack(), new ThreadStore(), "pwd", missing, 0, ""),
+    ).rejects.toThrow(/Working directory does not exist/);
+  });
+
+  it("_exec throws 'does not exist' for a missing cwd", async () => {
+    const ctx = makeMockCtx();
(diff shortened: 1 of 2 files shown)


packages/agency-lang/lib/stdlib/shell.test.ts
```diff
@@ -139,3 +139,54 @@ describe("_exec / _bash cwd ~ expansion", () => {
...
+ * (create the directory, then retry). Regression target: the agent
+ * doing `setAgentCwd("/tmp/build")` before the `mkdir`.
+ */
+describe("_exec / _bash reject a nonexistent cwd with a clear error", () => {
+  let tmp: string;
+  beforeEach(() => {
+    tmp = fs.mkdtempSync(path.join(os.tmpdir(), "shell-cwd-"));
+  });
+  afterEach(() => {
+    fs.rmSync(tmp, { recursive: true, force: true });
```
egonSchiele: We have code to safely delete om tests that you can use. Specifically, there is a function named "safeDelete" or similar that makes sure that the test doesn't accidentally delete the root directory. Please use that function.

packages/agency-lang/lib/stdlib/shell.ts
```diff
@@ -42,6 +42,43 @@ function buildSpawnOptions(
...
+  if (!cwd) return "";
+  const resolved = await resolveDir(cwd, allowedPaths, "cwd");
+  let stat;
+  try {
+    stat = await fs.stat(resolved);
+  } catch (e: any) {
+    if (e?.code === "ENOENT") {
+      throw new Error(
+        `Working directory does not exist: ${resolved}. ` +
+          "Create it first (e.g. `mkdir -p` from an existing directory) " +
```
egonSchiele: Change to "Create it first (e.g. `mkdir -p` from an existing directory, or another tool to create directories that you have access to) "

egonSchiele:
Both addressed in e3ae274:
- Tests now create the scratch dir inside the project root and clean up via `safeDeleteDirectory` instead of a raw `fs.rmSync`.
- The missing-cwd error now reads: "…Create it first (e.g. `mkdir -p` from an existing directory, or another tool to create directories that you have access to)…" — since bash may not be available under every policy.

PR #492, opened 2026-07-10
Failure propagation — Stage 2: flip the default to on

egonSchiele:
## Failure propagation — Stage 2: flip the default to `"on"`

Follow-up to #489, which shipped the full mechanism with the `ExecutionContext` default at `"warn"`. This PR is the flip: one literal in `context.ts` from `?? "warn"` to `?? "on"`, plus the strict-default tests and user docs.

**This PR's CI run over the execution corpus is the rollout measurement** the spec calls for (`docs/superpowers/specs/2026-07-08-failure-propagation-design.md`, Rollout section). The fixture drift from the field addition was already cleaned up in #489, so any trip here is a genuinely new legitimate site.

### What changes for programs

A failure Result passed to a parameter not typed to accept Results (`Result`, explicit `any`, or a union containing either) now skips the call and propagates the original failure, carrying a `skippedFunctions` trail. Failures into untagged plain TypeScript functions, and method calls on Results, throw catchable errors naming the producer. `failurePropagation: "warn"` or `"off"` in `agency.json` restores the old behavior.

The frameless fallback in `failurePropagation.ts` is untouched — it was already `"on"` and serves unit tests only.

### Included

- **Task 10**: the one-literal flip + the `config.ts` docstring (now documents `"on"` as default).
- **Task 11**: nine strict-default execution tests in `tests/agency/result/` — two-hop propagation trail, accepting vs rejecting params, leftmost-rejecting-param, variadic elements, `.partial()`-bound failures, `catch`/`try` composition, method calls (including the `r.value()` own-callable exemption), untagged TS function + aliased `isFailure`, the `arr.push(failure)` regression pin, the block fail-open pin, async propagation, and the print smoke test (observable via `print`'s return value).
- **Task 12**: "Failure propagation" section in the error-handling guide (including the `acceptsFailures` escape hatch) and the config-file reference entry.

### Verification

- Full unit suite: 7736 passed.
- All nine new execution tests pass locally.
- The three agency-js mode tests (`off`/`warn`/`on`) pass — they set their mode explicitly, so the default flip does not change their outcomes.
- `tsc` and `lint:structure` clean.

### Triage protocol for CI failures here

1. A test legitimately passes a failure into an untyped param → annotate that param `any`/`Result` at the site.
2. Anything else → treat as a real bug in the mechanism; fix before merging.
3. If legitimate sites are widespread (double digits), hold this PR — main works in warn mode indefinitely.

### DIFF
--- a/packages/agency-lang/docs/site/guide/agency-config-file.md
+++ b/packages/agency-lang/docs/site/guide/agency-config-file.md
@@ -133,7 +133,6 @@ Safety guards that stop a program from running away.
 ```json
 {
   "maxCallDepth": 2048,
-  "failurePropagation": "on",
   "maxToolCallRounds": 10,
   "checkpoints": { "maxRestores": 100 }
 }
@@ -142,7 +141,6 @@ Safety guards that stop a program from running away.
 | Option | Description |
 | --- | --- |
 | `maxCallDepth` | The runaway-recursion guard. Default `2048`; raise it for legitimately deep recursion. |
-| `failurePropagation` | What happens when a failure value reaches a parameter not typed to accept Results. `"on"` (default) skips the call and propagates the failure. `"warn"` logs without changing behavior. `"off"` disables the checks. |
 | `maxToolCallRounds` | How many times the LLM can loop between calling tools and reacting to their output before Agency halts it. Default `10`. |
 | `checkpoints.maxRestores` | Cap on how often one checkpoint can be restored before it errors out. Default `100`. |
 
--- a/packages/agency-lang/docs/site/guide/agency-config-file.md
+++ b/packages/agency-lang/docs/site/guide/agency-config-file.md
@@ -133,6 +133,7 @@ Safety guards that stop a program from running away.
 ```json
 {
   "maxCallDepth": 2048,
+  "failurePropagation": "on",
   "maxToolCallRounds": 10,
   "checkpoints": { "maxRestores": 100 }
 }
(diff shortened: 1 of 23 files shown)


packages/agency-lang/docs/site/guide/agency-config-file.md
```diff
@@ -133,6 +133,7 @@ Safety guards that stop a program from running away.
 ```json
 {
   "maxCallDepth": 2048,
+  "failurePropagation": "on",
```
egonSchiele: remove from guide. Once this is live, we don't want to give users the option to turn it off.

packages/agency-lang/docs/site/guide/error-handling.md
```diff
@@ -145,6 +145,50 @@ const result = try foo()
 
 Agency also adds an automatic try-catch around every function definition, and if an error is thrown, it returns a `Failure` type.
 
+## Failure propagation
```
egonSchiele, line 148: Make this shorter, just a single paragraph + one example.

egonSchiele:
All three comments addressed:

- **failure-helpers.js**: it was silently swallowed by the global `**/*.js` gitignore when the test directory was committed (same trap as the tracked `risky-helpers.js`, which was force-added at some point). Force-added and pushed in 237f51cbe, before this round of comments landed — the file is in the repo now and the test passes on CI from that commit forward.
- **Config guide entry**: removed. The `warn`/`off` values still exist in code as a rollback lever, but the guide no longer advertises them.
- **Guide section**: cut to a single paragraph + the one example.

Issue #486, opened 2026-07-10
llama.cpp structured output: schema-constrained calls return prose; JSON grammar not honored (local thread summaries silently dropped)


egonSchiele:
## Summary

Schema-constrained `llm()` calls on the llama-cpp path return **prose, not JSON**, despite the response format being set end to end. The zod parse then fails and — because callers like the eager thread summarizer are best-effort — the value is **silently dropped**. Concretely: thread summaries never persist on local models today, and any structured-output feature on the local path is unreliable.

## Fresh repro (2026-07-10, qwen3.5-2b)

Run the agent with a capabilities override turning the summarizer back on for the local model (`~/.agency-agent/settings.json` → `capabilities.models["qwen3.5-2b"].summarize: true`), then:

```
agency agent --local qwen3.5-2b -p "say hello" --log-file /tmp/log.json
```

The statelog shows the summarizer's `promptCompletion` **with `responseFormat` present** (so the schema reached the request on the agency side) and this output:

```
"\n\nThe user greeted the assistant, and the assistant responded with a friendly greeting and an offer to assist."
```

Expected: `{"summary": "..."}` (the summarizer's schema is `z.object({ summary: z.string() })`). The turn still exits 0 — the failure is invisible except for the missing summary.

Also observed earlier (2026-07-09, during the summarizer-hang debugging): a typed-assignment structured call on the local model returned prose with the field access yielding `undefined`.

## Where the layers hand off (suspects, unverified)

1. **agency → smoltalk:** `runPrompt` passes `responseFormat` into the prompt config; the completion event proves it's set. Open question: is the object arriving at the provider the shape `toJSONSchema()` expects to be called on?
2. **smoltalk-llama-cpp:** `_textSync` does `grammar = await llama.createGrammarForJsonSchema(config.responseFormat.toJSONSchema())` and applies it only when no tools are present (`if (grammar && !functions) options.grammar = grammar`). The summarizer has no tools, so the grammar SHOULD apply — yet the output is unconstrained. Candidates: `createGrammarForJsonSchema` receiving a bad/empty schema and producing a no-op grammar; the `options.grammar` key not being what `LlamaChat.generateResponse` expects; or a node-llama-cpp version behavior change. This code is identical in 0.1.0 and 0.1.1, so the bug is version-independent.
3. **node-llama-cpp:** grammar-constrained generation is exercised nowhere in our tests against a real model.

## Environment note

[...]

PR #493, opened 2026-07-10
Cross-provider memory: assignable embedding slot wired into enableMemory with key validation

egonSchiele:
## What

Cross-provider memory: point the embedding slot at one provider while chatting on another, e.g. OpenAI embeddings under a Google or Anthropic session.

- **The embedding slot is now assignable**: `/model embedding=openai/text-embedding-3-small` and the `--model` flag accept it. The value requires the `provider/model` form; bare names are refused with a message showing the form. Three new derived-slot resolution rules make the slot safe: it always parses its provider prefix even when a session provider is established, it never inherits the main provider, and a global pin never splashes into it (so `--model claude-opus-4-8` cannot become the memory embedding model). `promoteAll` exempts derived slots — the openai-responses promotion is a chat concern, and the embed endpoint uses the public provider name (the judgment call the plan flagged; recorded in the Task 1 commit).
- **The override reaches `enableMemory`** through a pure, tested seam: `embeddingKeyStatus` answers "is the slot usable" once with the payload the checkpoints need, `memoryEnablePlan` returns the enable path as a literal union, and `memoryConfigFor` builds the config so the wiring is testable. The Agency `MemoryConfig` type gains the `provider` field the TS type already had.
- **Key validation at three checkpoints**: startup degrades to default embedding behavior with one notice when the slot's key is missing; `/settings` memory-on refuses with a concrete message; `/model embedding=` warns immediately, with next-start timing stated.
- **Behavior change**: a main-provider change no longer pauses memory when the embedding slot pins the embed model — the embedding space behind recall did not change. Pinned by a test.
- **Sandbox fix**: the memory dir now honors `AGENCY_AGENT_HOME` like settings and policy. Before this, sandboxed agent tests wrote to the real `~/.agency-agent/memory`. This is what let the e2e below run fully sandboxed.

Spec: `docs/superpowers/specs/2026-07-10-cross-provider-memory-design.md` (revised after plan review — the review caught that no surface could set the slot at all, which became Task 1).

## Test plan

- 69 agent-harness tests across six suites, all green: the three resolution rules (including the pin-splash case), switchModel accept/refuse, settings sanitize of bare values, the pure helpers (override mapping + demotion, deterministic key checks via setEnv, the plan's four outcomes including the stale-flag precedence case, the config seam), the keyless-assignment warning (with an openai-responses case that pins demotion and a cleared-slot case that pins the empty-model guard), the conditional pause, and the sandbox fix. Lint and make clean.
- **Sandboxed e2e, healthy path** (Google chat + OpenAI embeddings, `AGENCY_AGENT_HOME` scratch dir): every `embedCompletion` names `text-embedding-3-small` and none name `gemini-embedding-001` — the override beat Google's own derivation, and smoltalk authenticated the embed from the environment (the spec's remaining verify-item).
- **Sandboxed e2e, degraded path** (`env -u OPENAI_API_KEY`): the fallback notice prints exactly once, the only embed event is `gemini-embedding-001` (derivation resumed), and the `memoryRecall` span shows memory stayed enabled — observed, not assumed.

## Pre-existing issues surfaced by the e2e (none caused here; diff touches none of these paths)

1. **Memory-internal LLM calls do not follow a forced provider.** Under `settings.model.provider: "google"`, the recall step and the eager thread summarizer still issued `claude-sonnet-4-6` calls (which failed: the Anthropic account is out of credits). Worth its own issue: the memory manager's internal model chain and the summarizer pick up defaults that ignore the resolved slots.
2. **Gemini extraction shape**: `remember`'s typed extraction result came back missing `expirations`, crashing the tool (`result.expirations is not iterable`). Related to the structured-output reliability track (#486).
3. Because of 1 + 2, the full remember→recall round-trip could not complete in this environment; the e2e instead proves the override contract directly through the embed events, which is what this PR owns.

## Manual items for the owner

- The `/settings` interactive smoke: memory → on with the slot set and key unset (expect the red refusal), and with no slot on an Anthropic session (expect the note naming the slot).
- `npm i -g smoltalk-llama-cpp@latest` is still pending on this machine (global copy is the Jun 27 0.1.0 build).
- The Anthropic API account is out of credits.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -702,12 +702,23 @@ def editCapabilities() {
       patch = {
         memory: picked == "on"
       }
-      if (picked == "on" && providerLacksEmbeddings(getCapabilityProvider())) {
-        pushMessage(
-          color.dim(
-          "Note: ${getCapabilityProvider()} has no embeddings endpoint — memory runs with Tier-2 semantic recall disabled (structured recall still works).",
-        ),
-        )
+      if (picked == "on") {
+        const status = embeddingKeyStatus(getResolvedSlots())
+        if (status.state == "key-missing") {
+          pushMessage(
+            color.red(
+            "Cannot enable memory: the embedding slot needs ${status.varName}, which is not set. Export it, or point the embedding slot elsewhere via /model.",
+          ),
+          )
+          return
+        }
+        if (status.state == "no-override" && providerLacksEmbeddings(getCapabilityProvider())) {
+          pushMessage(
+            color.dim(
+            "Note: ${getCapabilityProvider()} has no embeddings endpoint, so Tier-2 semantic recall stays off. Point the embedding slot at a provider that has one, e.g. /model embedding=openai/text-embedding-3-small.",
+          ),
+          )
+        }
       }
     }
   }
--- a/packages/agency-lang/lib/agents/agency-agent/agent.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/agent.agency
@@ -9,7 +9,7 @@ import { map } from "std::array"
 import { today } from "std::date"
 import { generateImage } from "std::image"
 import { applyAgentCwd, setAgentCwd } from "std::index"
-import { listHostedModels, hostedModelInfo, setLlmOptions, envVarFor } from "std::llm"
+import { listHostedModels, hostedModelInfo, setLlmOptions } from "std::llm"
 import { setMemoryId, disableMemory } from "std::memory"
 import { keys } from "std::object"
 import { basename, dirname } from "std::path"
@@ -57,7 +57,8 @@ import {
   BUILTIN_POLICIES,
  } from "./lib/defaultPolicy.agency"
 import { filterHostedModels, ModelFilters } from "./lib/modelFilters.agency"
-import { parseModelFlag, parseModelSpec, knownProviders, Resolved } from "./lib/models.agency"
+import { parseModelFlag, parseModelSpec, knownProviders, embeddingSpecHasProvider, missingProviderEnvVar, Resolved } from "./lib/models.agency"
+import { SLOTS } from "./lib/slots.agency"
 import {
   buildLayers,
   resolveAll,
@@ -103,7 +104,7 @@ import {
   memoryEnablePlan,
   overrideFromResolved,
   embeddingsOverride,
-  providerKeyMissing,
+  providerEnvMissing,
   demoteProvider,
  } from "./shared.agency"
 import { codeAgent, setCodeOneShot } from "./subagents/code.agency"
@@ -403,9 +404,9 @@ export def switchModel(
     slots: nextSlots
   }
   const parsed = parseModelFlag(spec)
-  if (parsed.slot != "" && parsed.slot != "main" && parsed.slot != "reasoning" && parsed.slot != "embedding") {
+  if (parsed.slot != "" && !SLOTS.includes(parsed.slot)) {
     pushMessage(
-      color.red("Unknown model slot '${parsed.slot}'. Valid slots: main, reasoning, embedding."),
+      color.red("Unknown model slot '${parsed.slot}'. Valid slots: ${SLOTS.join(", ")}."),
     )
     return {
       session: prior,
@@ -413,21 +414,18 @@ export def switchModel(
       changes: []
     }
   }
-  if (parsed.slot == "embedding") {
+  if (parsed.slot == "embedding" && !embeddingSpecHasProvider(parsed.model)) {
     // The embedding slot exists to point AWAY from the chat provider, so
     // the provider can never be inferred. Require the explicit form.
-    const spec2 = parseModelSpec(parsed.model, false, knownProviders())
-    if (spec2.provider == "") {
-      pushMessage(
-        color.red(
-        "Embedding models need the provider/model form, e.g. embedding=openai/text-embedding-3-small.",
-      ),
-      )
-      return {
-        session: prior,
-        resolved: before,
-        changes: []
-      }
+    pushMessage(
+      color.red(
+      "Embedding models need the provider/model form, e.g. embedding=openai/text-embedding-3-small.",
+    ),
+    )
+    return {
+      session: prior,
+      resolved: before,
+      changes: []
     }
   }
   if (parsed.slot == "") {
@@ -473,10 +471,11 @@ export def switchModel(
   returns when a selective toggle lands.) Returns true if it emitted the notice. */
 export def reactToSlotChange(change: SlotChange): boolean {
   if (change.slot == "main" && change.before.provider != change.after.provider) {
-    // When the embedding slot pins the embed model, a main-provider
-    // change no longer changes the embedding space behind recall, so
-    // the pause would be a false positive.
-    if (embeddingsOverride(getResolvedSlots()) != null) {
+    // Skip the pause only when startup actually enabled memory WITH the
+    // override. A set-but-keyless slot took the fallback path, so memory
+    // runs on derived embeddings and the provider change still shifts
+    // the embedding space behind recall.
+    if (_memoryEmbeddingsPinned) {
       return false
     }
     pushMessage(
@@ -491,10 +490,21 @@ export def reactToSlotChange(change: SlotChange): boolean {
   }
   if (change.slot == "embedding") {
     const override = overrideFromResolved(change.after)
-    if (override != null && providerKeyMissing(override.provider)) {
+    if (override != null) {
+      const missing = missingProviderEnvVar(override.provider)
+      if (missing != "") {
+        pushMessage(
+          color.yellow(
+          "Warning: the embedding slot needs ${missing}, which is not set. The assignment takes effect at next agent start, and memory falls back to the default embedding behavior until it is exported.",
+        ),
+        )
+        return true
+      }
+      // The healthy path must not be silent: enableMemory already ran,
+      // so the user would otherwise believe embeddings switched now.
       pushMessage(
-        color.yellow(
-        "Warning: the embedding slot needs ${envVarFor(override.provider)}, which is not set. The assignment takes effect at next agent start, and memory falls back to the default embedding behavior until the key is exported.",
+        color.dim(
+        "Embedding slot set: memory embeddings use ${override.provider}/${override.model} at next agent start.",
       ),
       )
       return true
@@ -1297,9 +1307,22 @@ static const mainAgentTools = [
   pointed the embedding slot at a provider, embeddings run there. A
   missing API key degrades to the default embedding behavior with one
   notice. It never blocks startup. */
+// Whether startup actually enabled memory WITH the embedding override.
+// The pause guard in reactToSlotChange consults this instead of
+// re-deriving from the slots: a keyless override took the fallback path,
+// so memory runs on DERIVED embeddings and a main-provider change still
+// shifts the embedding space. Exported mutator so tests can drive the
+// guard directly.
+let _memoryEmbeddingsPinned = false
+
+export def markMemoryEmbeddingsPinned(pinned: boolean) {
+  _memoryEmbeddingsPinned = pinned
+}
+
 def maybeEnableMemory() {
   const status = embeddingKeyStatus(getResolvedSlots())
   const plan = memoryEnablePlan(getCapabilities().memory, status)
+  markMemoryEmbeddingsPinned(plan == "enable-override")
   if (plan == "skip") {
     return
   }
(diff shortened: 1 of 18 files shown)


packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -441,6 +463,12 @@ export def switchModel(
   returns when a selective toggle lands.) Returns true if it emitted the notice. */
 export def reactToSlotChange(change: SlotChange): boolean {
   if (change.slot == "main" && change.before.provider != change.after.provider) {
+    // When the embedding slot pins the embed model, a main-provider
+    // change no longer changes the embedding space behind recall, so
+    // the pause would be a false positive.
+    if (embeddingsOverride(getResolvedSlots()) != null) {
```
egonSchiele: **Correctness: this guard skips the pause when the override exists, but not when the override is actually in effect.** If the slot is set but its key is missing, startup took `enable-fallback`, so memory is running on derived embeddings — and `activeEmbeddingProvider()` (`lib/runtime/memory/manager.ts:388`) reads the live llmDefaults on every embed call. A main-provider change then really does shift the embedding space, and the pause is still needed.

Concrete failure, using this PR's own degraded-e2e setup: google main + keyless openai override → embeds run on `gemini-embedding-001`. Then `/model main=gpt-5.5` → provider changes → this guard returns early → subsequent embeds derive `text-embedding-3-small` → recall compares new-space queries against gemini-space vectors. That is exactly the unreliable-recall case the pause exists to prevent.

Cheap fix: skip only when `embeddingKeyStatus(getResolvedSlots()).state == "ready"`. Airtight fix: record which plan `maybeEnableMemory` actually took (a module `let` next to `_resolved`/`_capabilities`, the established pattern) and consult that — the env can change after startup via setEnv, so re-deriving at reaction time is a proxy. If you take the cheap fix, note that `mainProviderChangeSkipsPauseWithPinnedEmbedding` starts depending on the ambient `OPENAI_API_KEY`; add a `setEnv("OPENAI_API_KEY", "key") with approve` to keep it deterministic.

packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -451,6 +479,17 @@ export def reactToSlotChange(change: SlotChange): boolean {
     disableMemory() with approve
     return true
   }
+  if (change.slot == "embedding") {
```
egonSchiele, line 491: **UX gap: the success path is silent.** `/model embedding=openai/text-embedding-3-small` with the key present prints nothing, and the override only applies at the next agent start (`enableMemory` already ran). The keyless path states the next-start timing; the healthy path — the common one — leaves the user believing embeddings switched now. Suggest a dim notice in the else case, e.g. "Embedding slot set: memory embeddings use openai/text-embedding-3-small at next agent start." Returning true for it also makes the behavior pinnable in agentTurn.

packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -388,16 +393,33 @@ export def switchModel(
...
     return {
       session: prior,
       resolved: before,
       changes: []
     }
   }
+  if (parsed.slot == "embedding") {
+    // The embedding slot exists to point AWAY from the chat provider, so
+    // the provider can never be inferred. Require the explicit form.
+    const spec2 = parseModelSpec(parsed.model, false, knownProviders())
```
egonSchiele: Nit: the bare-value rule ("an embedding value must carry a provider prefix") is now implemented three times with the same `parseModelSpec(value, false, knownProviders()).provider == ""` shape — here, in `main()`'s `--model` parsing (line ~1796), and in `sanitizeModelSettings`. One predicate in models.agency (e.g. `embeddingSpecHasProvider(value: string): boolean`) would name the rule and keep the three sites from drifting (a future edit to one parse-arg is the drift risk). The messages/actions rightly stay per-site. Also `spec2` → something like `embedSpec`.

packages/agency-lang/lib/agents/agency-agent/shared.agency
```diff
@@ -58,7 +58,14 @@ export def resolveMaxToolCallRounds(
 // agency-lang#230. Falls back to `.` if `HOME` is unset (rare; mostly
 // a Windows concern — the `.` puts the dir under cwd as a last
 // resort instead of crashing module init).
-def computeAgentDir(): string {
+export def computeAgentDir(): string {
+  // The sandbox variable wins, matching POLICY_DIR in lib/config.agency.
+  // Before this check, sandboxed agent tests wrote to the real
+  // ~/.agency-agent/memory while settings and policy stayed sandboxed.
+  const sandbox = env("AGENCY_AGENT_HOME")
```
egonSchiele: There are now two independent resolvers of the agent home: this one and `resolveAgentHome` in `lib/config.agency` (which backs `POLICY_DIR`). They already disagree at the edges — `os.homedir()` vs `env("HOME")`, and the `./.agency-agent` last-resort fallback exists only here — and nothing ties them together except this side's comment. Consider exporting `resolveAgentHome` and expressing `computeAgentDir` through it (config.agency imports nothing from shared.agency, so no cycle), or at minimum add the mirror comment in config.agency so a future edit to one finds the other.

packages/agency-lang/lib/agents/agency-agent/shared.agency
```diff
@@ -135,6 +141,122 @@ export def demoteProvider(provider: string): string {
...
+export def memoryConfigFor(
+  embeddings: (EmbeddingsOverride | null),
+): MemoryConfig {
+  if (embeddings == null) {
+    return {
+      dir: "${AGENCY_AGENT_DIR}/memory"
+    }
+  }
+  return {
+    dir: "${AGENCY_AGENT_DIR}/memory",
```
egonSchiele: Nit: the memory-dir path literal appears in both branches. Hoisting `const dir = "${AGENCY_AGENT_DIR}/memory"` above the null-check keeps the path in one place.

packages/agency-lang/lib/agents/agency-agent/tests/agentDir.agency
```diff
@@ -0,0 +1,17 @@
...
+// policy do. Before this fix, sandboxed agent tests wrote to the real
+// ~/.agency-agent/memory. AGENCY_AGENT_DIR is a module-init static, so
+// this calls the def directly; a sandboxed process sets the variable
+// before launch.
+node agentDirHonorsSandbox(): string {
+  setEnv("AGENCY_AGENT_HOME", "/tmp/agency-sandbox-test") with approve
+  const sandboxed = computeAgentDir()
+  setEnv("AGENCY_AGENT_HOME", "") with approve
+  const normal = computeAgentDir()
+  const backToHome = normal != "/tmp/agency-sandbox-test"
```
egonSchiele: This assertion also passes if the empty-counts-as-unset guard is dropped: with `if (sandbox != null)` alone, the second `computeAgentDir()` returns `""`, and `"" != "/tmp/agency-sandbox-test"` is still true. `const backToHome = normal.endsWith(".agency-agent")` pins the rule — both real fallbacks (`$HOME/.agency-agent` and `./.agency-agent`) end with the dir name and the empty string does not.

egonSchiele (commented):
Reviewed the full diff. This lands the cross-provider memory design well: the three derived-slot resolution rules are the right altitude (the rule lives in `finalProvider`/`resolveSlot`/`slotMapToValues` once, not per call site), `embeddingKeyStatus` gives the checkpoints one shared answer with the payload they need, and `memoryConfigFor` makes the wiring seam testable — the test suite genuinely kills the mutants that matter (demotion via the openai-responses case, the cleared-slot guard, the pin-splash case, the config seam).

One correctness issue worth fixing before merge (the pause-skip guard, inline at agent.agency:469), one UX gap (silent success on `/model embedding=`), and a handful of nits — all inline.

Also verified while reviewing: `configureModels` has exactly one call site and it was updated; `activeEmbeddingProvider()` reads the live llmDefaults per embed call, which is what makes the pause-guard finding real; the regenerated `docs/site/stdlib/memory.md` matches the stdlib type change.

Diff-hygiene nit, no action needed if intentional: `agentTurn.test.json` and `resolution.test.json` were reformatted from the compact one-line-per-test style to expanded multi-line (~300 lines of noise in this diff), while the new `memoryWiring.test.json` uses the compact style. Whichever style wins, it would be good to have one.

packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -388,16 +393,33 @@ export def switchModel(
     slots: nextSlots
   }
   const parsed = parseModelFlag(spec)
-  if (parsed.slot != "" && parsed.slot != "main" && parsed.slot != "reasoning") {
+  if (parsed.slot != "" && parsed.slot != "main" && parsed.slot != "reasoning" && parsed.slot != "embedding") {
```
egonSchiele: The way this if statement is written, every time we add a new slot, we will need to update the if statement. Instead, please create a single `static const` listing all the slots somewhere, and then check for array membership in the if statement.

egonSchiele:
Pushed a fix for the owner-reported switch regression (gpt-4o-mini → claude-sonnet-4-6 failing with "Provider 'openai-responses' has no built-in defaults").

Two composing causes, both from this branch's blast radius: `priceDefault` validated the provider before its provider-independent derived-slot shortcut, and the new pins-skip-derived-slots rule sent the embedding slot through `priceDefault` for the first time — with the PROMOTED ctx provider `currentProvider()` hands to `switchModel`. Fixed by running the derived shortcut first and demoting the ctx inside `switchModel`. The second fix also closes a latent variant this branch did not create: `/model reasoning=x` on a fresh openai session let the main slot fall through to `priceDefault` with the promoted name.

Both flows are pinned by regression tests driving `switchModel` with a promoted ctx (25/25 in agentTurn).

packages/agency-lang/lib/agents/agency-agent/agent.agency
egonSchiele: Fixed in 7156bd872 with your airtight option. maybeEnableMemory records whether it actually took enable-override in a pinned flag (module let, the _resolved pattern), and the pause guard consults that instead of slot presence. Verified your mechanism first: activeEmbeddingProvider reads live llmDefaults per embed call, so fallback-mode memory really does follow a main-provider change. New test pins your exact scenario: set-but-keyless slot, flag false, provider change → pause fires. The pinned-skip test now drives the flag directly via the exported mutator, so neither test depends on ambient env.

packages/agency-lang/lib/agents/agency-agent/shared.agency
```diff
@@ -135,6 +141,122 @@ export def demoteProvider(provider: string): string {
...
+}
+
+/** True when the provider has a known API-key variable and the
+  environment leaves it unset or empty. Providers without a known
+  variable never report missing. */
+export def providerKeyMissing(provider: string): boolean {
+  const varName = envVarFor(provider)
+  if (varName == "") {
+    return false
+  }
```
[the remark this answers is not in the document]
egonSchiele: Confirmed and fixed in 7156bd872 — and it went one level deeper than the litellm case. models.agency already owned the composed check (checkProviderEnv validates key + base URL), so providerKeyMissing was a half-duplicate of existing machinery. The primitive is now missingProviderEnvVar (returns the first unset required var, or empty), checkProviderEnv re-expresses through it, and the renamed providerEnvMissing plus embeddingKeyStatus consume it — so the status names LITELLM_BASE_URL when that is the missing piece. Test pins the litellm key-set-but-no-base-URL case.

packages/agency-lang/lib/agents/agency-agent/agent.agency
egonSchiele: Added in 7156bd872: the healthy path prints a dim next-start notice naming provider/model and returns true; the keyless and healthy outcomes are both pinned in agentTurn (warn true, notice true, cleared-slot false).

packages/agency-lang/lib/agents/agency-agent/agent.agency
egonSchiele: Done in 7156bd872: embeddingSpecHasProvider in models.agency, consumed by all three sites; spec2 is gone with it.

packages/agency-lang/lib/agents/agency-agent/shared.agency
egonSchiele: Unified in 7156bd872: resolveAgentHome is exported and computeAgentDir delegates to it, so the agent home has exactly one resolver. This also unifies the edge semantics deliberately: the memory dir now uses os.homedir() like everything else, and the dead ./.agency-agent fallback is gone.

packages/agency-lang/lib/agents/agency-agent/shared.agency
egonSchiele: Hoisted in 7156bd872.

packages/agency-lang/lib/agents/agency-agent/tests/agentDir.agency
egonSchiele: Good catch — switched to normal.endsWith(".agency-agent") in 7156bd872, with a comment naming the empty-counts-as-unset rule it pins.

packages/agency-lang/lib/agents/agency-agent/agent.agency
egonSchiele: Done in 7156bd872: the check is now !SLOTS.includes(parsed.slot) against the existing SLOTS const in slots.agency, and both error messages render the list from SLOTS.join, so a new slot is a data edit.

PR #496, opened 2026-07-10
Add keyof and indexed access type operators

egonSchiele:
Adds two type operators, per the spec and reviewed plan on this branch (`docs/superpowers/specs/2026-07-10-keyof-indexed-access-design.md`).

`keyof T` is the union of an object type's key names. `T["key"]` is the type of one field. Both replace hand-copied types that silently drift when the source type changes — and `keyof` produces a closed literal union, so `match` over a `keyof User` value gets exhaustiveness checking: add a field to `User` and every match over its keys reports a missing case (both directions tested).

## What

- **Two new `VariableType` variants** (`keyofType`, `indexedAccessType`) that exist only between parse and resolution. Evaluation is eager in `resolveTypeWithGuard` via `lib/typeChecker/typeOperators.ts`, which consumes the argument helpers `builtinGenerics.ts` already owns (`resolveObjectArg`, `resolveKeysArg`, `withUseSiteTags` — now exported), so error wording stays one family across Partial, Pick, keyof, and indexed access. Downstream passes never see the operator nodes; there is no provider surface.
- **Parser: reviewer, look here.** `arrayTypeParser`'s counted `[]` suffix became a shared base chain plus a suffix loop (`[]` wraps an array, `[<type>]` indexes — so `User[keyof User]` parses for free). The load-bearing asymmetry: the or-chain member requires AT LEAST ONE suffix (bare types must keep falling through to later alternatives), while keyof's operand uses a private zero-or-more variant. A zero-suffix parser in the or-chain would greedily match `keyof` as a plain identifier, because `or()` commits to the first success. This sits under every type annotation's parse path; the full lib suite (7768 tests) plus re-runs of the utility-types and recursive-types execution suites guard it.
- **Semantics:** object-only in v1 (`Record`, arrays, primitives, unions error — matching the utility-types stance); index keys must be string literals or unions of literals; union index yields the union of value types; missing keys error listing available keys; chains compose left to right; property tags ride along through indexing (`@validate` on a field survives extraction, verified down to the descriptor emission).
- **`keyof` is now a reserved type name** (like `Result`): `type keyof = ...` is rejected instead of silently changing how annotations parse. Grep confirmed zero existing `.agency` code uses the name. `keyofish` still parses as a plain identifier (keyword boundary pinned by test).

## The deepResolveNode trap, observed then fixed

`deepResolveNode` prepares alias bodies for codegen and only routed `genericType`/`typeAliasVariable`. Without new cases, `type K = keyof User` reached the zod mapper unresolved and silently emitted the `z.string()` fallback — no error, wrong schema. The codegen pins were written red-first (all four failed with the fallback), and every assertion uses a shape the fallback cannot fake (literal unions, `z.number()`). Execution also flushed out that `hasAnyValidateTag` needed cases for the new variants — without them, a validated indexed alias silently skipped the descriptor path.

## Docs

Beyond the guide section: `lib/types/typeHints.ts`'s variant checklist now names `typeKey` and `deepResolveNode` (both postdated the comment), and `docs/dev/adding-features.md` gains an "Adding a `VariableType` variant" walkthrough spelling out exactly which sites the compiler enforces and which fail silently — with the rule that every new variant gets a non-string codegen pin, because that is the only net that catches a `deepResolveNode` miss.

## Tests

11 parse tests (precedence, keyword boundary, union index, chains, paren distinctions), 24 unit+pipeline typechecker tests (semantics table, every error, exact-member union assertions, recursion/composition/`Keys<T>` delegation, swallowed-error tripwires), 4 fallback-proof codegen pins, fixture with zero churn elsewhere, 2-node execution test, 6 formatter round-trips including `(keyof User)[]` vs `keyof User[]` staying distinct.

## Scope-outs

- `Record<K,V>` operands (v1 decision, relaxable).
- `typeof` — looks like a sibling, is not: it needs the checker's VALUE scopes, which do not exist at alias-resolution time; adding it means threading a synthesizer through `resolveType` and every caller, a pipeline-ordering change, and a story for compiling with the typechecker disabled. Own spec if ever wanted.
- `T[number]` — no tuples (provider gate), so numeric indexing has no v1 use case.
- Located diagnostics for resolver errors — standing follow-up shared with Record and the utility types (tripwire tests pin today's swallowed behavior).

### DIFF
--- a/packages/agency-lang/docs/site/guide/types.md
+++ b/packages/agency-lang/docs/site/guide/types.md
@@ -223,31 +223,6 @@ Anthropic does not support recursive types as LLM output contracts
 (OpenAI and Gemini do). On Anthropic, parse the response yourself with
 `schema(Tree).parseJSON(...)`.
 
-## keyof and indexed access
-
-`keyof T` is the union of an object type's key names. `T["key"]` is the
-type of one field. Both update automatically when the source type
-changes:
-
-```ts
-type User = {
-  name: string,
-  email: string,
-}
-
-type Field = keyof User        // "name" | "email"
-type Name = User["name"]       // string
-```
-
-`keyof` produces a closed union, so `match` over a `keyof` value gets
-exhaustiveness checking. Add a field to `User` and every match over its
-keys reports a missing case.
-
-Both operators need a concrete object type. Records, arrays, and
-primitives are errors. Index keys must be string literals. A union of
-keys returns a union of field types, so `User[keyof User]` is the type
-of every value in `User`.
-
 ## References
 
 - [Schemas](/guide/schemas)
--- a/packages/agency-lang/docs/site/guide/types.md
+++ b/packages/agency-lang/docs/site/guide/types.md
@@ -223,6 +223,31 @@ Anthropic does not support recursive types as LLM output contracts
 (OpenAI and Gemini do). On Anthropic, parse the response yourself with
 `schema(Tree).parseJSON(...)`.
 
+## keyof and indexed access
+
+`keyof T` is the union of an object type's key names. `T["key"]` is the
+type of one field. Both update automatically when the source type
+changes:
+
+```ts
+type User = {
+  name: string,
+  email: string,
+}
+
+type Field = keyof User        // "name" | "email"
+type Name = User["name"]       // string
+```
+
+`keyof` produces a closed union, so `match` over a `keyof` value gets
+exhaustiveness checking. Add a field to `User` and every match over its
+keys reports a missing case.
+
+Both operators need a concrete object type. Records, arrays, and
+primitives are errors. Index keys must be string literals. A union of
+keys returns a union of field types, so `User[keyof User]` is the type
+of every value in `User`.
+
 ## References
 
 - [Schemas](/guide/schemas)
--- a/packages/agency-lang/lib/backends/typescriptGenerator/typeToString.ts
+++ b/packages/agency-lang/lib/backends/typescriptGenerator/typeToString.ts
@@ -208,7 +208,16 @@ export function variableTypeToString(
       .join(", ");
     return `${variableType.name}<${args}>${formatValueArgs(variableType.valueArgs)}`;
   } else if (variableType.type === "keyofType") {
-    return `keyof ${variableTypeToString(variableType.operand, typeAliases, forFormatting)}`;
+    const op = variableTypeToString(
+      variableType.operand,
+      typeAliases,
+      forFormatting,
+    );
+    // Parenthesize a union operand: `keyof (A | B)` must not print as
+    // `keyof A | B`, which re-parses as (keyof A) | B.
+    return variableType.operand.type === "unionType"
+      ? `keyof (${op})`
+      : `keyof ${op}`;
   } else if (variableType.type === "indexedAccessType") {
     const obj = variableTypeToString(
       variableType.objectType,
@@ -216,7 +225,10 @@ export function variableTypeToString(
       forFormatting,
     );
     const wrapped =
-      variableType.objectType.type === "keyofType" ? `(${obj})` : obj;
+      variableType.objectType.type === "keyofType" ||
+      variableType.objectType.type === "unionType"
+        ? `(${obj})`
+        : obj;
     const index = variableTypeToString(
       variableType.index,
       typeAliases,
--- a/packages/agency-lang/lib/backends/typescriptGenerator/typeToString.ts
+++ b/packages/agency-lang/lib/backends/typescriptGenerator/typeToString.ts
@@ -136,7 +136,12 @@ export function variableTypeToString(
       typeAliases,
       forFormatting,
     );
-    if (variableType.elementType.type === "unionType") {
+    if (
+      variableType.elementType.type === "unionType" ||
+      variableType.elementType.type === "keyofType"
+    ) {
+      // keyof parenthesizes for the same re-parse reason as unions:
+      // `keyof User[]` reads as keyof (User[]).
       return `(${inner})[]`;
     }
     return `${inner}[]`;
@@ -202,6 +207,22 @@ export function variableTypeToString(
       .map((a) => variableTypeToString(a, typeAliases, forFormatting))
       .join(", ");
     return `${variableType.name}<${args}>${formatValueArgs(variableType.valueArgs)}`;
+  } else if (variableType.type === "keyofType") {
+    return `keyof ${variableTypeToString(variableType.operand, typeAliases, forFormatting)}`;
+  } else if (variableType.type === "indexedAccessType") {
+    const obj = variableTypeToString(
+      variableType.objectType,
+      typeAliases,
+      forFormatting,
+    );
+    const wrapped =
+      variableType.objectType.type === "keyofType" ? `(${obj})` : obj;
+    const index = variableTypeToString(
+      variableType.index,
+      typeAliases,
+      forFormatting,
+    );
+    return `${wrapped}[${index}]`;
   }
   return "unknown";
 }
--- a/packages/agency-lang/lib/parsers/parsers.ts
+++ b/packages/agency-lang/lib/parsers/parsers.ts
@@ -86,6 +86,7 @@ import {
   AgencyComment,
   AgencyMultiLineComment,
   ArrayType,
+  KeyofType,
   BlockType,
   BooleanLiteralType,
   NumberLiteralType,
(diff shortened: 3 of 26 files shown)


packages/agency-lang/docs/site/guide/types.md
```diff
@@ -223,6 +223,31 @@ Anthropic does not support recursive types as LLM output contracts
 (OpenAI and Gemini do). On Anthropic, parse the response yourself with
 `schema(Tree).parseJSON(...)`.
 
+## keyof and indexed access
```
egonSchiele: undo this guide edit

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -992,44 +993,112 @@ export const parenthesizedTypeParser: Parser<VariableType> = memo(
...
   }
-  return result;
-};
+  return { success: true, rest, result: current };
+}
+
+export const arrayTypeParser: Parser<ArrayType> = (input: string) =>
+  trace("arrayTypeParser", (i: string) =>
+    parseTypePostfix(i, 1),
+  )(input) as ParserResult<ArrayType>;
```
egonSchiele: The declared type is now inaccurate: with `minSuffixes: 1`, this parser also returns `IndexedAccessType` at the top level (e.g. `User["name"]` — one suffix, and it is the index form), so `Parser<ArrayType>` + the `as ParserResult<ArrayType>` cast lie to consumers. Runtime is fine since both or-chains treat it as `VariableType`, but the old cast was only papering over depth-wrapping; this one hides a genuinely different node type. Suggest declaring it `Parser<VariableType>` (both consuming or-chains already accept that), or if the export type must stay for compat, a one-line comment saying the name is historical and the result may be an indexed access.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -992,44 +993,112 @@ export const parenthesizedTypeParser: Parser<VariableType> = memo(
...
+    const emptyBrackets = str("[]")(rest);
+    if (emptyBrackets.success) {
+      current = { type: "arrayType", elementType: current };
+      rest = emptyBrackets.rest;
+      suffixes += 1;
+      continue;
     }
+    const openBracket = char("[")(rest);
+    if (openBracket.success) {
+      const index = variableTypeParser(openBracket.rest);
```
egonSchiele: Nit: no whitespace tolerance inside index brackets — `User[ "name" ]` fails to parse (the index parser runs directly on the char after `[`). TypeScript accepts spaces here, and the fix is cheap (`optionalSpaces` before the index parse and before the `]` check). If the strictness is deliberate (it does match the old `[]` suffix, which was also space-intolerant), a word in the doc comment above `parseTypePostfix` would stop someone from filing it as a bug later.

packages/agency-lang/lib/types/typeHints.ts
```diff
@@ -55,6 +69,31 @@ export type GenericType = {
  * Example: in `type Container<T> = { value: T }`, the `T` is a TypeParam.
  * Defaults allow bare use of the alias: `type StringMap<V = any> = Record<string, V>`.
  */
+/**
+ * `keyof T` — the union of an object type's key names. Exists only
+ * between parse and resolution: `resolveTypeWithGuard` evaluates it
+ * eagerly (lib/typeChecker/typeOperators.ts) and downstream code never
+ * sees the node.
+ */
+export type KeyofType = {
```
egonSchiele, line 83: The insertion point orphans `TypeParam`'s doc comment: the block at lines 67-70 ("A type parameter declaration on a generic type alias...") used to document `TypeParam`, but `KeyofType`/`IndexedAccessType` now sit between the comment and the declaration, so `TypeParam` at line 97 is undocumented and its docstring dangles above an unrelated type. Move the two new types (with their own docstrings) below `TypeParam`, or above the stranded comment.

packages/agency-lang/lib/typeChecker/typeOperators.test.ts
```diff
@@ -0,0 +1,298 @@
...
+    expect(missingKey).toEqual([]);
+    const unknownAlias = typecheckSource(`
+node main() {
+  const k: keyof NotDefined = "x"
+  return k
+}
+`);
+    // The undefined-alias diagnostic may legitimately fire here (operand
+    // reference validation); assert only that nothing CRASHES.
+    expect(Array.isArray(unknownAlias)).toBe(true);
```
egonSchiele: `expect(Array.isArray(unknownAlias)).toBe(true)` can only fail if `typecheckSource` throws, so as written it is a no-crash pin with an assertion that reads like a behavior check. The plan's version of this test said to record the observed diagnostics; that got dropped. Suggest either `expect(() => typecheckSource(...)).not.toThrow()` (honest about what it checks) or pinning the actual observed diagnostics count/messages so the tripwire also fires when the undefined-alias diagnostic behavior changes.

egonSchiele (commented):
Reviewed against the spec, the reviewed plan, and the anti-pattern catalog. **Verdict: this is in excellent shape — every finding from the plan and test-audit reviews landed correctly, and I found no correctness bugs.** Four minor inline comments below.

**What I verified against the code (not just the description):**
- The parser split implements the many1/many0 asymmetry exactly as designed: `arrayTypeParser` keeps the fail-on-bare-base behavior (`parseTypePostfix` with `minSuffixes: 1`, matching old `count` semantics), the zero-or-more variant is private to `keyofTypeParser`, and `keyofTypeParser` sits before `arrayTypeParser` in both or-chains. Backtracking in the suffix loop is correct — failed suffix attempts never consume input (`rest` only advances on a completed suffix).
- Helper sharing is real: `resolveObjectArg`/`resolveKeysArg` are consumed, not re-implemented, and the missing-key error is in Pick's wording family. The test regex `/expects string literal keys/` matches `resolveKeysArg`'s actual wording.
- Tag threading mirrors the `genericType` branch via `withUseSiteTags`; property tags ride through indexing down to the descriptor emission (`(A as any).__agency_descriptor` pin).
- All four codegen pins are fallback-proof (literal unions, `z.number()` — nothing the `z.string()` fallback can fake), and the committed fixture `.mjs` shows the correct emissions (`const Field = z.union([z.literal("name"), z.literal("age")])`, `const Age = z.number()`).
- The fan-out is complete: `typeKey`, both `valueParamSubstitution` switches, `mapTypes`+`visitTypes` (operand recursion, which is what makes `Keys<T>` substitution work), `formatTypeHint` and `variableTypeToString` with parenthesization both ways, `hasAnyValidateTag`, `deepResolveNode`, `RESERVED_TYPE_NAMES`.
- Edge semantics hold: `(keyof User)["x"]` parses via parens and errors at eval (object-only), `User[]["x"]` parses and errors the same way, `keyof {a} | B` mixed unions parse, printer parens keep `(keyof User)[]` and `keyof User[]` distinct in both printers.

**Two non-blocking notes:**
- The `docs/superpowers/` spec/plan/review files are committed in this PR; previous features kept those local/untracked. Fine either way, but flagging in case it was unintentional.
- LSP `semantics.ts` was not touched, so the `keyof` keyword gets no semantic token highlighting. Degrades gracefully; possible follow-up alongside the located-diagnostics one.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -992,44 +993,112 @@ export const parenthesizedTypeParser: Parser<VariableType> = memo(
...
+ *   matches), so bare types fall through to later alternatives. A
+ *   zero-suffix variant in the or-chain would greedily match `keyof` as
+ *   a plain identifier and break every keyof annotation, because or()
+ *   commits to the first success.
+ * - postfixOperandParser uses 0: keyof binds to the full postfix
+ *   expression that follows it (`keyof User`, `keyof User[]`,
+ *   `keyof User["a"]` all work). Only keyofTypeParser consumes it, so
+ *   the bare-match greediness leaks nowhere.
+ */
+function parseTypePostfix(
```
egonSchiele: Use tarsec combinators to write this in a declarative manner

packages/agency-lang/lib/backends/typescriptGenerator/typeToString.ts
```diff
@@ -202,6 +207,22 @@ export function variableTypeToString(
       .map((a) => variableTypeToString(a, typeAliases, forFormatting))
       .join(", ");
     return `${variableType.name}<${args}>${formatValueArgs(variableType.valueArgs)}`;
+  } else if (variableType.type === "keyofType") {
+    return `keyof ${variableTypeToString(variableType.operand, typeAliases, forFormatting)}`;
+  } else if (variableType.type === "indexedAccessType") {
```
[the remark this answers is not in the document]
egonSchiele: All four fixed in 5e3d9c8e: both printers now parenthesize union operands under keyof and union objects in indexed access (alongside the existing keyof-object wrap). Parse tests confirm `keyof (A | B)` and `(A | B)["k"]` produce the right ASTs, and two new round-trip cases pin that the parens survive formatting.

packages/agency-lang/lib/backends/typescriptGenerator/typeToString.ts
```diff
@@ -202,6 +207,22 @@ export function variableTypeToString(
...
+    const obj = variableTypeToString(
+      variableType.objectType,
+      typeAliases,
+      forFormatting,
+    );
+    const wrapped =
+      variableType.objectType.type === "keyofType" ? `(${obj})` : obj;
```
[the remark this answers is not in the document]
egonSchiele: See the reply on the first typeToString thread — fixed together in 5e3d9c8e.

packages/agency-lang/lib/utils/formatType.ts
```diff
@@ -79,6 +83,15 @@ export function formatTypeHint(
     }
     case "genericType":
       return `${vt.name}<${vt.typeArgs.map(recurse).join(", ")}>`;
+    case "keyofType":
+      return `keyof ${recurse(vt.operand)}`;
```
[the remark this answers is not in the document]
egonSchiele: See the reply on the first typeToString thread — fixed together in 5e3d9c8e.

packages/agency-lang/lib/utils/formatType.ts
```diff
@@ -79,6 +83,15 @@ export function formatTypeHint(
...
+    case "indexedAccessType": {
+      // Parenthesize a keyof object for the same re-parse reason as the
+      // arrayType case.
+      const obj = recurse(vt.objectType);
+      const wrapped = vt.objectType.type === "keyofType" ? `(${obj})` : obj;
```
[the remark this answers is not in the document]
egonSchiele: See the reply on the first typeToString thread — fixed together in 5e3d9c8e.

packages/agency-lang/docs/site/guide/types.md
egonSchiele: Removed in 5e3d9c8e (it was already committed, so this is a revert commit-over-commit). The dev-facing docs in `docs/dev/` are untouched — say the word if you want user-facing docs somewhere else instead.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Declared `Parser<VariableType>` in 5e3d9c8e, with a doc comment noting the name is historical and the result may be an indexed access. Both consuming or-chains already accepted the wider type; the cast is gone entirely (the declarative rewrite removed the need for it).

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Added in 5e3d9c8e: index brackets tolerate inner whitespace (`User[ "name" ]` parses, pinned by test). The bare `[]` array suffix stays space-intolerant, matching its old behavior — noted in the TypeSuffix doc comment.

packages/agency-lang/lib/types/typeHints.ts
egonSchiele: Fixed in 5e3d9c8e — the two new types moved below `TypeParam`, so its docstring is reunited with its declaration.

packages/agency-lang/lib/typeChecker/typeOperators.test.ts
egonSchiele: You were right that it was a no-crash pin dressed as a behavior check. Fixed in 5e3d9c8e by pinning the observed diagnostics exactly: the undefined-alias message DOES fire through the keyof operand (`Type alias 'NotDefined' is not defined (referenced in 'k').`) — which is also nice confirmation that validateTypeReferences descends into the new variants via visitTypes. The tripwire now fires if that behavior changes.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Rewritten declaratively in 5e3d9c8e: a `TypeSuffix` tagged union parsed by `or(map(str("[]")...), map(seqC(...)))`, folded onto the base with a pure `reduce` (`applyTypeSuffixes`), and the two arities expressed as `many1` vs `many` — the imperative loop, the manual rest-threading, and the minSuffixes check are all gone. The many1/many0 or-chain-safety rationale moved onto the two consumers where it belongs.

PR #14, opened 2026-04-15
Add per-checkpoint maxRestores option to restore()

egonSchiele:
## Summary
- Adds `maxRestores` field to `RestoreOptions` so callers can limit how many times a specific checkpoint is restored, e.g. `restore(cp, { maxRestores: 3 })`
- Populates user-created checkpoints with real source location data (moduleId, scopeName, stepPath) so restore counts are keyed by stable location strings that survive checkpoint ID changes across restore cycles
- Adds location-keyed restore counting (`locationRestoreCounts`) to `CheckpointStore`, separate from the existing ID-keyed global safety limit

## Test plan
- [x] New agency execution test (`tests/agency/maxRestores`) verifies `maxRestores: 3` causes exactly 3 retries then continues
- [x] Updated `checkpoint-restore.mjs` generator fixture to reflect new location fields
- [x] All 1823 vitest tests pass
- [x] `pnpm run build` succeeds

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #497, opened 2026-07-10
Add a fresh-eyes verification step to the code agent (one-shot)

egonSchiele:
## What

Adds an Agency-native **verification step** to the code agent's one-shot loop: a fresh-eyes critic that re-reads the task, runs the produced artifact, and loops back with concrete gaps when the work isn't actually complete.

Motivated by Terminal-Bench benchmarking, where two **non-capability** failure clusters dominated (cases the model *could* do but the run failed anyway):
1. **Wrote it but never ran/verified it** — produces scripts/configs and *describes* the solution, never executes or checks the end-to-end result (`configure-git-webserver`, `mailman`, `git-multibranch`, `kv-store-grpc`; several pass on Opus).
2. **Output-format / spec-adherence** — right reasoning, wrong contract: wrong path, wrong format, mislabeled field (`fix-code-vulnerability` reported cwe-22 vs cwe-93, `extract-elf` signed vs uint32, `reshard-c4-data` misread the spec).

## How

**`verify.agency`** (new) — modeled on `oracle.agency`:
- `verify(originalUserMsg): Verdict` — fresh thread; two `llm()` calls (tool-loop inspection → structured verdict).
- Tools: `read`/`ls`/`glob`/`grep`/**`bash`**, no `write`/`edit` — it reports, never fixes; may `bash`-install a package only if needed to run the artifact.
- Receives **only the original task, not the agent's reply** — judges the disk independently rather than trusting the agent's "done".
- Prompt reframes verification around the reality that Terminal-Bench's tests are **hidden**: reconstruct the success check from the task's own description + any example it provides, then run the artifact through it; check the exact output contract literally.
- `unwrapVerdict` (pure `safe def`, `match (r) { success/failure }`) turns a `verify.invoke()` `Result` into a plain `Verdict`, **failing open** on any `Failure`.

**`code.agency`** — wires it into the existing review-retry loop:
- After a clean `review()`, gated on `_oneShot`, capped at **2** verify-fix cycles (`verifyRounds`). On gaps → re-prompt + loop; on a verifier `Failure` → fail open. Interactive REPL is untouched.
- Base-prompt hard rules **#13** (deliverable is a run-and-verified artifact, not a plan/stub/script you didn't run) and **#14** (satisfy the exact output contract — path, format, every criterion), plus a one-shot note that output is graded automatically (often by hidden tests). These map directly onto the two clusters.

## Tests

`tests/agency/verify-agent.agency` (4 deterministic, no real API): `verify()` parses both verdict outcomes; **`unwrapVerdict` fails open** on `failure(...)` (the safety-critical path). The full-`codeAgent` loop is validated by a live one-shot smoke rather than a deterministic mock — the `thread(summarize: true)` machinery injects a summarizer `llm()` call that makes a deterministic full-loop mock brittle.

## Notes

- One codegen constraint hit: `if (x is success(v))` doesn't compile inside the stepped loop body; the pure `unwrapVerdict` + `match` (per `stdlib/data/wikidata.agency`) is the working pattern.
- The design spec and plan live under `docs/superpowers/{specs,plans}/2026-07-10-code-agent-verification-step*` — those paths are gitignored (local working docs), so they're not in this PR.
- Also included: **`docs/incremental-build-stale-dist-bug.md`** — a report for a *separate* issue found along the way (incremental `make agents` can leave `dist` inconsistent; `make clean && make` is the workaround). Drop it from this PR if you'd rather it stand alone.


packages/agency-lang/docs/incremental-build-stale-dist-bug.md
```diff
@@ -0,0 +1,145 @@
+# Bug: incremental build leaves `dist` inconsistent (stale skip → missing export)
```
egonSchiele: make a GitHub issue for this

packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
```diff
@@ -448,6 +476,22 @@ export def codeAgent(userMsg: string, allowHandoff: boolean) {
         hasErrors = true
       } else {
         hasErrors = false
+        // Fresh-eyes verification — one-shot only, only after the code
+        // typechecks clean, capped at 2 verify-fix cycles. `verify.invoke`
+        // returns a Result: a Failure (verifier crashed / no verdict)
+        // matches the else arm and we fail open, since verification can
+        // only help — never block or corrupt (the verifier never writes).
+        if (_oneShot && verifyRounds < 2) {
+          const verdictResult = verify.invoke(originalUserMsg)
```
egonSchiele: you can just call the function directly, no need to use `.invoke`.

packages/agency-lang/lib/agents/agency-agent/subagents/verify.agency
```diff
@@ -0,0 +1,103 @@
...
+// Read-only + run. No write/edit: the verifier reports, it never fixes.
+// Partial-applied with `useAgentCwd` so paths resolve against the same
+// working directory the code agent used (`setAgentCwd`), where the
+// artifacts live.
+export static const verifyTools: any[] = [
+  read.partial(useAgentCwd: true),
+  ls.partial(useAgentCwd: true),
+  glob.partial(useAgentCwd: true),
+  grep.partial(useAgentCwd: true),
+  bash.partial(useAgentCwd: true),
```
egonSchiele: There's no way to constrain `bash` to be just readonly. Use `exec` instead, as well as any other functions you see in std::fs and std::shell

packages/agency-lang/lib/agents/agency-agent/subagents/verify.agency
```diff
@@ -0,0 +1,103 @@
...
+// Unwrap a `verify.invoke()` Result to a plain Verdict, failing OPEN on
+// any Failure (verifier crashed / no verdict) — verification can only
+// help, never block. Kept as a pure `safe def` so the `match` compiles
+// cleanly and the caller's loop body stays free of binder-in-`if`
+// patterns (which the stepped-loop codegen doesn't handle).
+export safe def unwrapVerdict(r: Result<Verdict>): Verdict {
+  const failOpen: Verdict = { satisfied: true, gaps: [] }
+  return match (r) {
+    success(v) => v
+    failure(err) => failOpen
```
egonSchiele: shouldn't this contain the failure error message?

egonSchiele:
Thanks — all addressed in 395f7ea:

- **`.invoke` was a runtime bug** (Copilot, and @egonSchiele's "call directly"): `verify.invoke(str)` calls the runtime `AgencyFunction.invoke()` with a string, which breaks. Switched to **`try verify(...)`** → `Result<Verdict>` → `unwrapVerdict`. My dropped full-loop test would've caught this; the deterministic tests only exercised `verify()`/`unwrapVerdict` directly.
- **`bash` → `exec`**: a bash string can't be constrained, so the "read-only" framing was false. `verifyTools` now uses `exec` (command + args, allow/block-listable); prompt updated to run programs via `exec`.
- **Verify-cap loop bug** (Copilot): `needsFix` is now reset `false` on every clean-review iteration, so once `verifyRounds` hits 2 and the block is skipped it can't spin the loop to the `iterations` cap.
- **Surface the failure message** (@egonSchiele): the fail-open verdict now carries the error in `gaps` (`failOpenVerdict`) so a verifier crash shows in the trace instead of being swallowed; `satisfied` stays true so it doesn't re-prompt. Test updated to assert the surfaced message.
- **Stale `.invoke()` comments** (Copilot): fixed to `try`.
- **Incremental-build doc → issue** (@egonSchiele): filed as #498 and removed from this PR.

`make agents` clean, 4/4 verify tests pass.

PR #500, opened 2026-07-11
Strict structured-output validation + opt-in validation retries

### LINKED ISSUE #494: Structured output fail-opens to raw content on parse failure — memory extraction crashes, typed llm() results silently mistyped
## Summary

When a structured-output `llm()` result cannot be parsed or validated, `runPrompt` silently returns the raw message content where the caller's type promises a validated object. `lib/runtime/prompt.ts:1559`:

```ts
  if (responseFormat) {
    try {
      const rawResult = JSON.parse(responseMessage.content || "");
      return extractResponse(rawResult, responseFormat);
    } catch (e) {
      try {
        return extractResponse(responseMessage.content, responseFormat);
      } catch (e) {
        return responseMessage.content;   // <- silent fail-open
      }
    }
  }
```

Every provider quirk in structured output becomes a downstream type crash or silent data loss instead of a loud, classifiable llmError.

## Probe evidence (2026-07-10)

Identical typed call, two providers:

```
type Probe = {
  entities: { name: string; observations: string[] }[];
  expirations: { entityName: string; observationContent: string }[]
}
const result: Probe = llm("Extract entities from: I like cherries. ...")
```

- openai gpt-4o-mini: `typeof result = object`, `expirations: []` present. Correct.
- google gemini-3.5-flash: `typeof result = string`, value `""`. The fail-open returned an empty string typed as `Probe`.

## Consequences observed in the wild

- **Memory is broken end to end on Google sessions**: the `remember` extraction does `const result: ExtractionResult = llm(prompt)` and then iterates `result.expirations`. The fail-open hands it a string, `result.expirations` is undefined, the tool crashes with `result.expirations is not iterable`, and nothing is ever written to the graph. Reproduced today via the agent (`--provider google`, "Remember that I like cherries.").
- **Thread summaries silently drop on local models** (#486): the summarizer's zod parse fails on llama.cpp prose and the value vanishes.
- The memory graph on this machine has not received a write since Jun 19, across weeks of sessions.

There are two bugs here. The provider-side ones (gemini returning an empty string for a `responseJsonSchema` request; llama.cpp ignoring its grammar, #486) deserve their own fixes. But this fail-open is what turns each of them into silent corruption instead of a visible failure.

## Suggested direction

Treat an unparseable structured-output response as a request failure: raise an `llmError` (which also terminates the promptStart pair and shows in the statelog) or route it through the retry policy the way hosted validation failures should be. If a fail-open escape hatch must exist, it should at least emit a loud statelog event naming the expected schema.

egonSchiele:
Fixes #494.

## What this does

Two layered changes, one commit each stage:

**1. Strict structured-output validation (the #494 fix).** A schema-constrained `llm()` result now either validates or comes back as a failure Result. It is never the raw content masquerading as the declared type. The old fail-open silently broke memory extraction: a provider returning prose crashed three calls downstream with no log at the cause. Validation failures now post a statelog error (new errorType `structuredOutput`) at the moment they happen, and stdlib call sites (memory remember/forget, thread summarize, the eager summarizer) adopt bang validation.

**2. Validation retries, opt-in.** A new `validationRetries` resilience option: when the response fails schema validation, Agency sends the validation error back to the model and asks again, up to the configured count, before surfacing the failure.

```agency
const person: Person = llm("Describe a scientist", { validationRetries: 2 })
```

Default is 0 (off): validation retries cost tokens, unlike transport retries. Configure per call, per branch via `setLlmOptions({ validationRetries: n })`, or leave off. Retries render in the agent REPL as a "retrying" line via the existing `onLLMRetry` hook (reason `invalidStructuredOutput`).

## Design notes

- The retry decision is a pure function, `decideValidationRetry`, mirroring the `decideRetry`/`runWithRetry` split transport retries already use. The loop in `runPrompt` is a thin driver.
- The driver wraps the tool loop: a retry response may itself contain tool calls, which drain before re-validation. Most of the `prompt.ts` diff is a 2-space re-indent of the existing tool loop; review with whitespace hidden.
- Resume discipline mirrors `nextLlmCall`: the `validation.N` step keys derive from the frame-backed attempt counter, which only advances inside the dispatch step. A dedicated e2e test interrupts mid-retry and asserts no duplicate feedback, no re-dispatch, no double-counted attempt.
- Validation reads the last assistant message: after a resume, the tail message can be our own feedback text, which a string schema would accept via envelope-skip recovery.
- Two hand-synced mirrors of `RetryConfig` gained the new key: the typeChecker llm-option allowlist and the agency-side `LlmDefaults`. The latter was also missing `retries`/`timeout` (pre-existing gap: the runtime honored them, the type rejected them) — fixed here. `backoff` stays off the agency type (needs a nested type alias, no consumer).

## Ride-along compiler fixes (surfaced by the stdlib bang adoption)

- `mapBodies` rebuilt as a fold over a new shared `bodySlots` table (the statement-body twin of `expressionChildren`); `walkNodes` descends bodies off the same table. Fixes pattern lowering never descending into `thread {}` blocks ("Unhandled Agency node type: isExpression") and adds missing `withModifier`/`staticStatement` coverage.
- The flow builder now treats `return interrupt effect(...)` as fall-through (it resumes on approval), fixing false ".value is only available on a success Result" errors after the gated-work idiom used across git/wikipedia/memory stdlib.

## Testing

- Unit: resolver precedence incl. falsy-0 override; `decideValidationRetry` accept/retry/surface/cap/truncation; both type surfaces compile without "Unknown property" warnings.
- E2E (`tests/agency-js/structured-output-retry`): five call-counted nodes — retry succeeds on feedback (2 calls), explicit cap exhausts (3), 0 disables (1), branch default applies (3), hook reason captured via callback.
- E2E (`tests/agency-js/structured-output-retry-resume`): fails validation → retry round calls a tool that interrupts → resume → valid. Each fixture key maps to a specific resume failure mode.
- Full vitest (7752 tests), make, lint:structure, memory + threads-registry agency suites, and the statelog pairing/span/resume suites all pass locally.

## Known limitations (named)

- Direct TS callers passing `args.retryConfig` get only the fields they set (same as retries/timeout/backoff on that path).
- Transport-retry x validation-retry multiplication is untested as a product; each mechanism is tested alone.
- No `agency.json` llm-defaults block (transport retries lack one too).

### DIFF
--- a/packages/agency-lang/docs/site/guide/llm-part-2.md
+++ b/packages/agency-lang/docs/site/guide/llm-part-2.md
@@ -59,6 +59,23 @@ Connection drops (ECONNRESET, fetch failed, …), `5xx`, `429`. A `429` is for r
 
 The call returns a `Failure`, which is covered in the chapter on [error handling](/guide/error-handling). 
 
+### Validation retries
+
+Transport retries handle a provider that failed to answer. Validation
+retries handle a provider that answered with the wrong shape. When an
+`llm()` call declares a result type and the response fails validation,
+Agency sends the validation error back to the model and asks again:
+
+```ts
+const person: Person = llm("Describe a scientist", { validationRetries: 2 })
+```
+
+After the retries run out, the call returns a failure. Use the bang
+syntax (`Person!`) to handle it locally, or `catch` for a default value.
+Validation retries cost tokens, unlike transport retries, so they are
+off by default. `setLlmOptions({ validationRetries: 2 })` sets a
+branch-wide default.
+
 ## Other options to llm()
 
 You can pass an options object as the second parameter, or use named arguments. All options are optional, and are grouped below by purpose.
@@ -87,11 +104,12 @@ You can pass an options object as the second parameter, or use named arguments. 
 
 ### Resilience
 
-| Option | Type |
-|---|---|
-| `retries` | `number` |
-| `timeout` | `duration` |
-| `backoff` | `{ initial?: duration, factor?: number, max?: duration }` |
+| Option | Type | Description |
+|---|---|---|
+| `retries` | `number` | Transport retries: the provider failed to answer. |
+| `timeout` | `duration` | Per-attempt deadline. |
+| `backoff` | `{ initial?: duration, factor?: number, max?: duration }` | Wait between transport retries. |
+| `validationRetries` | `number` | Retries when a structured-output response fails schema validation. Each retry sends the validation error back to the model. Disabled by default (0). Independent of `retries`. |
 
 See [retries and timeouts](#retries-and-timeouts) for more info.
 
--- a/packages/agency-lang/docs/site/guide/llm-part-2.md
+++ b/packages/agency-lang/docs/site/guide/llm-part-2.md
@@ -59,23 +59,6 @@ Connection drops (ECONNRESET, fetch failed, …), `5xx`, `429`. A `429` is for r
 
 The call returns a `Failure`, which is covered in the chapter on [error handling](/guide/error-handling). 
 
-### Validation retries
-
-Transport retries handle a provider that failed to answer. Validation
-retries handle a provider that answered with the wrong shape. When an
-`llm()` call declares a result type and the response fails validation,
-Agency sends the validation error back to the model and asks again:
-
-```ts
-const person: Person = llm("Describe a scientist", { validationRetries: 2 })
-```
-
-After the retries run out, the call returns a failure. Use the bang
-syntax (`Person!`) to handle it locally, or `catch` for a default value.
-Validation retries cost tokens, unlike transport retries, so they are
-off by default. `setLlmOptions({ validationRetries: 2 })` sets a
-branch-wide default.
-
 ## Other options to llm()
 
 You can pass an options object as the second parameter, or use named arguments. All options are optional, and are grouped below by purpose.
--- a/packages/agency-lang/lib/lowering/patternLowering.test.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.test.ts
@@ -752,3 +752,42 @@ match ((r is success) && y) {
     expect(findTaggedAssignment(lifted.nodes)).toBeDefined();
   });
 });
+
+describe("lowering inside thread blocks", () => {
+  it("lowers `if (r is success(v))` inside a thread block — no isExpression survives", () => {
+    // Regression: mapBodies had no messageThread case, so lowering never
+    // descended into `thread { ... }` bodies. The raw isExpression reached
+    // the TypeScriptBuilder, which crashed with
+    // "Unhandled Agency node type: isExpression".
+    const lowered = lower(`
+thread {
+  let r = foo()
+  if (r is success(v)) {
+    print(v)
+  } else {
+    print("failed")
+  }
+}
+`);
+    const survived = walkNodesArray(lowered).some(
+      ({ node }) => node.type === "isExpression",
+    );
+    expect(survived).toBe(false);
+  });
+});
+
+describe("lowering inside with-wrapped statements", () => {
+  it("lowers a boolean `is` inside `stmt with approve` — no isExpression survives", () => {
+    // Regression: mapBodies (now bodySlots) skipped withModifier, so the
+    // wrapped statement was never lowered and a raw isExpression could
+    // reach codegen.
+    const lowered = lower(`
+let r = foo()
+print(r is success) with approve
+`);
+    const survived = walkNodesArray(lowered).some(
+      ({ node }) => node.type === "isExpression",
+    );
+    expect(survived).toBe(false);
+  });
+});
--- a/packages/agency-lang/lib/runtime/utils.ts
+++ b/packages/agency-lang/lib/runtime/utils.ts
@@ -3,6 +3,7 @@ import { GlobalStore } from "./state/globalStore.js";
 import { ThreadStore } from "./index.js";
 import { RunNodeResult } from "./types.js";
 import { nativeTypeReplacer, nativeTypeReviver } from "./revivers/index.js";
+import { failure, isSuccess, success, ResultValue } from "./result.js";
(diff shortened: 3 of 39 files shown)


packages/agency-lang/docs/site/guide/llm-part-2.md
```diff
@@ -59,6 +59,23 @@ Connection drops (ECONNRESET, fetch failed, …), `5xx`, `429`. A `429` is for r
 
 The call returns a `Failure`, which is covered in the chapter on [error handling](/guide/error-handling). 
 
+### Validation retries
```
egonSchiele: remove this entire addition, just leave the addition in the table

packages/agency-lang/lib/lowering/patternLowering.test.ts
```diff
@@ -752,3 +752,42 @@ match ((r is success) && y) {
...
   });
 });
+
+describe("lowering inside thread blocks", () => {
+  it("lowers `if (r is success(v))` inside a thread block — no isExpression survives", () => {
+    // Regression: mapBodies had no messageThread case, so lowering never
+    // descended into `thread { ... }` bodies. The raw isExpression reached
+    // the TypeScriptBuilder, which crashed with
+    // "Unhandled Agency node type: isExpression".
+    const lowered = lower(`
```
egonSchiele, line 762: Is this part of your change? Where is this coming from?

packages/agency-lang/lib/typeChecker/flowBuilder.ts
```diff
@@ -213,6 +213,14 @@ const statementRules: StatementRuleTable = {
 
   returnStatement: (node, flow, env) => {
     attachExpressionsToFlow(node.value, flow, env);
+    // `return interrupt effect(...)` may RESUME: on approval, execution falls
```
egonSchiele, line 216: Similar question. I think we probably do want these type fixes, but I'm wondering what they're doing in your PR and where they should be. I suppose if they're not in another PR, let's leave them in here because we do want them.

packages/agency-lang/stdlib/llm.agency
```diff
@@ -39,7 +39,10 @@ export type LlmDefaults = {
   reasoningEffort?: "low" | "medium" | "high";
   maxTokens?: number;
   maxToolResultChars?: number;
-  maxToolCallRounds?: number
+  maxToolCallRounds?: number;
+  retries?: number;
```
egonSchiele, line 43: Were these missing before? I don't understand how they were only getting included in this PR. I thought we had this already.

packages/agency-lang/lib/runtime/utils.ts
```diff
@@ -30,65 +31,98 @@ export function deepFreeze<T>(obj: T, seen: WeakSet<object> = new WeakSet()): T
...
   }
 
   // 5. Object with "response" or "properties" key — unwrap
   const wrapKeys = ["response", "properties"];
   for (const key of wrapKeys) {
     if (key in rawValue) {
       const inner = schema.safeParse(rawValue[key]);
-      if (inner.success) return inner.data[key];
+      if (inner.success) return success(inner.data[key]);
     }
```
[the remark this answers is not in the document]
egonSchiele: Confirmed and fixed in 57db1d5a8. The bug is pre-existing on main (this PR ported the heuristic verbatim), but strict mode upgraded it from "returns garbage" to "blesses undefined as success", so it belongs here. Step 5 now strips the wrapper key and recurses, so step 1's direct match does the extraction. Unit tests cover the properties-wrapped shape (previously success(undefined)), the double-wrapped response shape, and the mismatch-still-fails case.

packages/agency-lang/docs/site/guide/llm-part-2.md
egonSchiele: Removed in 57db1d5a8; the table row stays.

packages/agency-lang/lib/lowering/patternLowering.test.ts
egonSchiele: It is part of this change, though the chain is indirect. The strict-validation commit rewrites memory.agency and thread.agency to use `if (result is success(x))` inside `thread { }` blocks. Compiling that crashed make with "Unhandled Agency node type: isExpression": pattern lowering never descended into thread-block bodies, because mapBodies had no messageThread case. The fix centralizes the body-field enumeration in bodySlots (the statement-body twin of expressionChildren), which also surfaced that withModifier and staticStatement were missing. These tests pin both. Without this fix the stdlib changes in this PR do not compile, which is why it rides along rather than living in its own PR.

packages/agency-lang/lib/typeChecker/flowBuilder.ts
egonSchiele: Same chain as the lowering fix: memory.agency's remember/forget now use `is success` narrowing after the `return interrupt` gate, and the flow builder treated `return interrupt` as a flow exit, so everything after it lost narrowing and make emitted false ".value is only available on a success Result" errors on this PR's stdlib code. The stdlib in this PR does not compile clean without it, and it is not in any other PR, so per your comment it stays here. It also silently un-breaks narrowing after every gated `return interrupt` in git/wikipedia stdlib.

packages/agency-lang/stdlib/llm.agency
egonSchiele: They were half-present. The TS-side LlmDefaults (lib/stdlib/llm.ts) extends RetryConfig, so the RUNTIME has honored retries/timeout via setLlmOptions since #336. But this agency-side type is a hand-maintained mirror, and it never listed them, so the type checker warned on setLlmOptions({ retries: ... }) even though it worked. The plan review caught the same drift about to happen to validationRetries (the typeChecker allowlist in builtins.ts is a second hand-synced mirror), so this PR adds the new key to both mirrors and backfills the two missing ones here. backoff stays off the agency type: it needs a nested type alias and has no consumer asking for it.

PR #501, opened 2026-07-11
agency run --policy: CLI-driven interrupt policies

egonSchiele:
## What

Adds CLI-driven interrupt policies to `agency run`, so you can constrain how a program's interrupts are handled without editing its source:

```bash
agency run agent.agency --policy recommended            # a built-in
agency run agent.agency --policy ./my-policy.json        # a policy file
agency run agent.agency --approve std::read --reject std::write   # inline lists
agency run agent.agency --interactive                    # prompt on uncovered effects
```

- **Non-interactive and fail-closed by default**: any effect the policy doesn't explicitly approve is rejected. No TTY needed.
- **`--interactive`** opts into an ephemeral prompt (approve / reject / approve-always / reject-always, remembered for the run only) on effects the policy doesn't cover.
- **Inline `--approve`/`--reject`** compose with `--policy` (they layer on top); an effect named in both resolves to **reject**.
- The installed handler is **outermost**, so a program's own `with approve` can't bypass it.

### Covers code the agent runs itself

The policy also governs interrupts raised inside code run via `std::agency::run`. The root handler is installed once in the root process; a subprocess's interrupts forward up to it via the existing IPC chain (the subprocess deliberately does **not** install its own — guarded by `isIpcMode()`). So `--approve std::run --approve std::read --reject std::write` lets an agent spawn generated code, read, but never write — even if the generated code approves its own write. An end-to-end test proves this *selective* enforcement at depth.

## How

- **`lib/runtime/builtinPolicies.ts`** — the four built-ins (`recommended`/`minimal`/`with-writes`/`approve-all`) + `builtinPolicy`/`builtinPolicyNames`/`BUILTIN_POLICIES` as a single TS source of truth, shared by the CLI resolver, the runtime handler, and (via a `std::policy` re-export) the agent. Replaces the agent-local `defaultPolicy.agency`, which is deleted.
- **`lib/cli/runPolicy.ts`** — `resolveRunPolicy(flags)` → `{ policyJson, interactive } | null`. Reject-wins-on-overlap is a single declarative per-effect construction (reject rule, then approve, then base), not order-dependent mutation.
- **`lib/runtime/runPolicyHandler.ts`** — `makeRunPolicyHandler` (pure decision logic + injected prompt) and `installRunPolicyHandler` (env + `isIpcMode` guard). Installed in `runNode` (fresh run) and `respondToInterrupts` (resume), since handlers are never serialized.
- **`lib/runtime/runPolicyEnv.ts`** — the `AGENCY_RUN_POLICY` / `AGENCY_RUN_POLICY_INTERACTIVE` wire contract, shared by the `agency run` writer and the runtime reader.
- **`scripts/agency.ts` / `lib/cli/commands.ts`** — the `--policy/--approve/--reject/--interactive` flags and env passing on spawn.

## Behavior notes / limitations (documented in the guide)

- Fail-closed overrides a program's own approvals of **unlisted** effects (intended — add the effect or use `--interactive`).
- Input-style interrupts (`const x = interrupt("...")`) get only approve/reject, not a value.
- `agency run --resume` is currently a no-op (its env var has no reader) and a fail-closed run never suspends; the resume-path install is defensive so the handler-safety invariant holds if resume is revived.

## Testing

- New unit tests: `builtinPolicies` (6), `runPolicy` resolver (11), `runPolicyHandler` incl. non-TTY + IPC-skip guards (11), end-to-end spawn (`--reject`/`--approve`/no-flag) (3).
- New agency-js tests: `run-policy` (read approved / write rejected / fail-closed / input-style, with discriminating assertions) and `run-policy-nested` (selective enforcement through `std::agency::run`).
- Regression: full `lib/runtime` + `lib/cli` + `lib/stdlib` vitest — **2624 passed, 0 failed**; agent `gitPolicy`/`execPolicy` tests 8/8; structural lint clean; `make` build clean.

## Follow-up (not in this PR)

Effect sets in policies and in the `--approve`/`--reject` flags (name a set from `std::capabilities`). The reject-wins-on-overlap rule was chosen to stay forward-compatible with it.


packages/agency-lang/lib/cli/runPolicy.spawn.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  it("approves std::write under --approve std::write (flag flips behavior)", async () => {
+    const dir = makeDir();
+    try {
+      const { stdout, code } = await runCli(dir, ["--approve", "std::write"]);
+      expect(code).toBe(0);
+      expect(stdout).toMatch(/WRITE_OK/);
+      expect(stdout).not.toMatch(/WRITE_REJECTED/);
+    } finally {
+      rmSync(dir, { recursive: true, force: true });
```
egonSchiele: Agency has a safe delete function you should use instead that ensures that you're not accidentally about to delete the root directory.

packages/agency-lang/lib/cli/runPolicy.ts
```diff
@@ -0,0 +1,83 @@
...
+  approve?: string;
+  reject?: string;
+  interactive?: boolean;
+  cwd: string;
+};
+
+function splitEffects(list: string | undefined): string[] {
+  if (!list) return [];
+  return list
+    .split(",")
```
egonSchiele: I think we probably want to split on a comma, a space, or a comma with spaces optionally before or after it.

packages/agency-lang/lib/runtime/runPolicyEnv.ts
```diff
@@ -0,0 +1,6 @@
+// The env-var wire contract between `agency run` (writer, lib/cli/commands.ts)
+// and the runtime root handler (reader, runPolicyHandler.ts). Defined once so
+// the two sides agree by construction.
+export const AGENCY_RUN_POLICY = "AGENCY_RUN_POLICY";
+export const AGENCY_RUN_POLICY_INTERACTIVE = "AGENCY_RUN_POLICY_INTERACTIVE";
```
egonSchiele: What happens if the user is running in interactive mode and multiple interrupts on multiple threads all get raised at once? How do we surface multiple interrupts to the user? Is there a way to do them in a queue so they get surfaced one at a time? Or will they clobber each other?

Also, is this file just for these three constants? It doesn't seem like there's anything else in the file. Do you expect to add more stuff to it at some point? If not, would that make sense to put these constants in an existing file? Ideally, we would have a single file that lists all of the different environment variables we use in this package.

packages/agency-lang/tests/agency-js/run-policy-nested/test.js
```diff
@@ -0,0 +1,32 @@
...
+writeFileSync(join(dir, "in.txt"), "NESTED_READABLE");
+
+const res = await main({ readDir: dir, readName: "in.txt" });
+const out = JSON.stringify(res.data);
+
+// Selective enforcement (NOT blanket reject): read APPROVED inside the child
+// (its content came back) AND write REJECTED inside the child (no file on disk,
+// rejection visible).
+const readApproved = out.includes("NESTED_READABLE");
+const writeRejected = !existsSync(join(dir, "gen-out.txt")) && /reject/i.test(out);
```
egonSchiele, line 31: This proves selective enforcement *reaches* nested generated code — good. But it doesn't pin the *mechanism* (forward-to-root, don't-install-in-child). If the `isIpcMode()` guard regressed and the child installed its own handler, the child would apply this same inherited policy locally — and a local reject is final (`gatherChainOutcome`: "Local reject is final — the parent is never consulted"), so the write would still be rejected and no file written. This test passes either way.

The `installRunPolicyHandler` IPC-skip unit test is what actually guards the mechanism, so coverage is fine — consider softening the "the child does not install its own" claim in the comment, or, to make this test discriminating, have the child depend on an effect only the *root* approves.

packages/agency-lang/lib/runtime/runPolicyHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+    const outcome = DECISIONS[await opts.prompt(intr)];
+    if (outcome.remember) remember(intr.effect, outcome.action);
+    return outcome.action === "approve" ? approve() : reject();
+  };
+}
+
+// Terminal prompt used by installRunPolicyHandler. Falls back to reject
+// (fail-closed) when stdin is not a TTY rather than hanging. Exported so the
+// non-TTY fallback is unit-testable.
+export async function terminalPrompt(intr: Intr): Promise<PromptDecision> {
```
egonSchiele, line 93: Good that the non-TTY fallback is exported and tested. The TTY branch's answer parsing (`aa`/`rr`/`a`/else→reject) is still untested and is easy to get subtly wrong (e.g. a user typing `approve` silently maps to reject). Consider extracting the `string → PromptDecision` mapping into a pure function and unit-testing the four cases — the readline plumbing stays untested, but the decision logic gets covered.

packages/agency-lang/lib/cli/runPolicy.spawn.test.ts
```diff
@@ -0,0 +1,87 @@
...
+import { mkdtempSync, writeFileSync, existsSync, rmSync } from "fs";
+import { tmpdir } from "os";
+import path from "path";
+
+const execFileAsync = promisify(execFile);
+
+// End-to-end: run the BUILT CLI (`agency run <file> --reject/--approve ...`) and
+// observe the program's own marker. A handler-rejected effect exits 0 and prints
+// nothing on its own, so the fixture inspects the write Result and prints a marker.
+const CLI = path.resolve("dist/scripts/agency.js");
```
egonSchiele, line 13: This hard-depends on a freshly built `dist/` (`make` before vitest). Worth confirming CI always rebuilds before this runs — a stale `dist/scripts/agency.js` would exercise old code. It fails loudly rather than silently (an unknown `--reject` flag → commander error → non-zero exit → the reject test's `expect(code).toBe(0)` fails), so this is a note, not a hole. The `beforeAll` guard message is a nice touch.

packages/agency-lang/lib/runtime/interrupts.ts
```diff
@@ -533,6 +534,11 @@ export async function respondToInterrupts(args: {
   if (args.overrides) applyOverrides(checkpoint, args.overrides);
 
   const execCtx = await ctx.createExecutionContext(interrupt.runId);
+  // Re-install the CLI-driven root policy handler on the resumed exec context
+  // (handlers are never checkpointed). Guarded no-op unless AGENCY_RUN_POLICY
+  // is set and this is the root process. Mirrors the runNode install so the
+  // policy survives a resumed leg.
+  installRunPolicyHandler(execCtx);
```
egonSchiele, line 541: Good call re-installing here — handlers aren't serialized, so without this a resumed leg would silently lose the policy (exactly the handler-safety trap CLAUDE.md warns about). Confirmed that `AGENCY_RESUME_FILE` currently has no reader anywhere (only the setter in `commands.ts`), so `--resume` is a no-op and a fail-closed run never suspends — this install is defensive today. Worth a follow-up to either wire up or remove the dead `--resume` / `AGENCY_RESUME_FILE` path, so this doesn't sit as code that can't be exercised end-to-end.

egonSchiele (commented):
Reviewed the implementation against the earlier plan review. Strong PR — it addresses essentially every concern from that review:

- **reject-wins** is now a single declarative per-effect construction, not order-dependent mutation (`runPolicy.ts`).
- the interactive **"always"** branches are driven by a `DECISIONS` table instead of duplicated array surgery (`runPolicyHandler.ts`).
- the root handler is installed on **both** `runNode` and `respondToInterrupts` (resume), behind a shared env-var contract module.
- the `isIpcMode()` skip has a **direct unit test**.
- the agency-js tests use **discriminating assertions** (content-came-back for approve; no-file-on-disk + handled-failure for reject) instead of the `Array.isArray` proxy, and `with-writes` now asserts the non-`dir` match fields (`target` / `src`+`dest` / `cwd`) plus the input-style-interrupt limitation.

Inline notes are all minor/optional. Two nits not worth their own thread:

1. The conditional array-spread `...(rejected.includes(effect) ? [rejectRule] : [])` in `runPolicy.ts` is the array sibling of the object-spread shape `docs/dev/anti-patterns.md` calls out ("never use this pattern"). It's defensible and reads well here — just flagging in case the structural linter objects.
2. `rejectRule` / `approveRule` (and the `approve` const in `builtinPolicies.ts`) are shared object references reused across effects. Safe today because everything is JSON-serialized for the env or cloned before mutation, but a latent footgun if any future code mutates a rule in place.

Nice work.

packages/agency-lang/lib/runtime/runPolicyHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+};
+
+// Build the outermost policy handler for a CLI-driven run. `policy` is the
+// base; interactive "always" decisions accumulate into a working clone so
+// checkPolicy serves them without re-prompting.
+export function makeRunPolicyHandler(
+  policy: Policy,
+  opts: { interactive: boolean; prompt: PromptFn },
+): HandlerFn {
+  const working: Policy = JSON.parse(JSON.stringify(policy));
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2c050c27e. `working` is now a null-prototype object (`Object.assign(Object.create(null), …)`), so indexing/writing by a program-controlled `intr.effect` like `"__proto__"` is just an ordinary string key with no prototype to reach or mutate.

packages/agency-lang/lib/cli/runPolicy.ts
```diff
@@ -0,0 +1,83 @@
...
+        ", ",
+      )}) or a readable file`,
+    );
+  }
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(readFileSync(policy, "utf-8"));
+  } catch (e) {
+    throw new Error(`policy file ${policy} is not valid JSON: ${String(e)}`);
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2c050c27e. `readFileSync` and `JSON.parse` are now separate try blocks: a read failure reports `could not read policy file …`, a parse failure reports `… is not valid JSON`.

packages/agency-lang/lib/cli/runPolicy.spawn.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    return { stdout: e.stdout ?? "", stderr: e.stderr ?? "", code: e.code ?? 1 };
+  }
+}
+
+describe("agency run --policy flags (end-to-end)", () => {
+  beforeAll(() => {
+    if (!existsSync(CLI)) {
+      throw new Error(`built CLI not found at ${CLI} — run \`make\` first`);
+    }
+  });
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2c050c27e. The suite is now `describe.skipIf(!existsSync(CLI))`, so a clean `pnpm test:run` (no `dist`) skips it rather than hard-failing; CI runs `make` before tests, so it still runs there.

packages/agency-lang/lib/runtime/builtinPolicies.ts
```diff
@@ -0,0 +1,142 @@
...
+  "std::git::remoteList": approve,
+  "std::git::blame": approve,
+  "std::git::stashList": approve,
+};
+
+// Glob matching a directory and everything under it — same convention
+// std::policy's buildScopedMatch uses for "approve-always-here".
+function dirScope(baseDir: string): string {
+  return `{${baseDir},${baseDir}/**}`;
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2c050c27e. `dirScope` now escapes picomatch metacharacters in `baseDir` (`escapeGlob`), so a dir containing `{`/`}`/`,`/`*` is matched literally and cannot widen the scope. Added a picomatch behavioral test (`/a,b` does not match `/a` or `b`).

packages/agency-lang/lib/cli/commands.ts
```diff
@@ -262,9 +268,12 @@ export function run(
...
 
-  const env = resumeFile
-    ? { ...process.env, AGENCY_RESUME_FILE: resumeFile }
-    : process.env;
+  const env: NodeJS.ProcessEnv = { ...process.env };
+  if (resumeFile) env.AGENCY_RESUME_FILE = resumeFile;
+  if (runPolicy) {
+    env[AGENCY_RUN_POLICY] = runPolicy.policyJson;
+    if (runPolicy.interactive) env[AGENCY_RUN_POLICY_INTERACTIVE] = INTERACTIVE_ON;
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2c050c27e. `run()` now `delete`s both `AGENCY_RUN_POLICY` and `AGENCY_RUN_POLICY_INTERACTIVE` from the child env before applying this run’s flags, so an inherited value can never silently enable policy/interactive.

packages/agency-lang/lib/cli/runPolicy.spawn.test.ts
egonSchiele: Good call — but `std::policy`’s `safeDeleteDirectory` refuses any path outside the project root (`checkInsideProject`), and these temp dirs live under `os.tmpdir()`, so it returns `{success:false}` without deleting. Instead I added an `rmTemp()` helper (2c050c27e) that `realpathSync`-resolves the dir and only `rmSync`s it if it is a real subdirectory of `os.tmpdir()` (never tmpdir itself) — the same "don’t nuke something unexpected" guard, scoped to where these paths actually are.

packages/agency-lang/lib/cli/runPolicy.ts
egonSchiele: Done in 2c050c27e — `splitEffects` now splits on `/[\s,]+/`, so `--approve "std::read, std::ls"`, `--approve std::read,std::ls`, and `--approve "std::read std::ls"` all work. Added a resolver test.

packages/agency-lang/lib/runtime/runPolicyEnv.ts
egonSchiele: Both addressed in 2c050c27e:

1. Concurrency: yes, concurrent prompts on one stdin would have clobbered. `terminalPrompt` now serializes through a process-global promise chain, so simultaneous interrupts (fork/race) are surfaced to the user one at a time and answered in sequence rather than interleaving. (Process-global is appropriate here — it guards the single physical terminal, not per-run state.)
2. The file was indeed just three constants. Moved them into `lib/constants.ts` (the existing project-wide constants file) under an "Environment variables" section and deleted `runPolicyEnv.ts`. A single registry for *all* `AGENCY_*` env vars across the package is a larger sweep (~12 files) — happy to do it as a follow-up if you want it.

packages/agency-lang/tests/agency-js/run-policy-nested/test.js
egonSchiele: Agreed — the outcome is identical whether the child forwards or installs-and-locally-rejects (same inherited policy, and a local reject is final). Softened both the test comment and the driver comment (2c050c27e) to say this proves enforcement *reaches* nested code, and to point at the `installRunPolicyHandler` IPC-skip unit test as what actually pins "the subprocess must not install its own handler."

packages/agency-lang/lib/runtime/runPolicyHandler.ts
egonSchiele: Done in 2c050c27e. Extracted the mapping into a pure `parsePromptAnswer(raw)` and unit-tested all four cases. It now also accepts the spelled-out words (`approve`/`reject`/`approve-always`/`reject-always`) in addition to the short forms, so a user typing `approve` no longer silently maps to reject; anything unrecognized still fails closed to reject.

packages/agency-lang/lib/cli/runPolicy.spawn.test.ts
egonSchiele: Addressed alongside the other dist note — the suite is now `describe.skipIf(!existsSync(CLI))`, so a stale/absent `dist` skips locally rather than exercising old code, and CI (`make` before tests) runs it against a fresh build.

packages/agency-lang/lib/runtime/interrupts.ts
egonSchiele: Agreed it shouldn’t sit as un-exercisable code. Kept the defensive re-install (handler-safety), and I’ll file a follow-up to either wire up or remove the dead `--resume` / `AGENCY_RESUME_FILE` path rather than fold that orthogonal change into this PR.

egonSchiele: Filed as #503.

PR #502, opened 2026-07-11
Add intersection types: the last type-system roadmap feature

egonSchiele:
Adds the `&` intersection operator, per the spec and reviewed plan on this branch. **This closes the type-system feature roadmap** started 2026-07-09:

- Shipped: utility types (#478), recursive aliases (#485), keyof + indexed access (#496), intersections (this PR).
- Deliberately killed: tuples (2 of 3 providers reject the schema, live-probed), typeof (needs value scopes at alias-resolution time — an architecture change), mapped/conditional types (fail the eager-evaluation litmus test).

## What

`A & B` merges object types — the mixin pattern at the type level:

```
type Named = { id: string, name: string }
type Aged = { id: string, age: number }
type Person = Named & Aged   // { id, name, age }
```

- **One parser precedence level** between union and its items: `&` binds tighter than `|`, looser than postfix/keyof (all TS parity, all pinned by parse tests). N-ary, newline-tolerant, single members pass through with no node.
- **A declarative merge pipeline** (`evalIntersection` in `typeOperators.ts`): RESOLVE every operand to an object type, GROUP all properties across all operands by key, COMBINE each group (shared keys intersect recursively via a three-rule ladder: identical keeps one, objects merge, anything else errors naming the key; tags merge so BOTH sides' validate chains apply), BUILD an ordinary object type. Grouping all operands at once makes n-ary merging associative by construction — tested, but structural.
- **Structural equality is injected**, not imported: the resolver branch passes a `typeKey`-based comparator built on the real alias table, so `typeOperators.ts` keeps its no-assignability-import cycle rule (plan-review finding).
- Eager evaluation as with the whole operator family: nothing downstream sees the node, no provider surface.

## Two findings from execution worth reviewer attention

1. **A latent parser bug, found and fixed.** The old `unionItemParser` ordered `objectTypeParser` before `arrayTypeParser`, which was only safe because the union parser's >=2-member gate discarded single-member matches. The intersection passthrough SUCCEEDS on single members, so the copied order committed to bare `{ n: string }` and stranded the `[]` suffix — `stdlib/memory.agency` stopped compiling, which is how it surfaced. The fix (arrayType before objectType in intersection items) also fixes the pre-existing union-member variant: `type X = { a: number | { n: string }[] }` failed to parse on main and parses now. Regression tests pin all three shapes.
2. **Parse-performance gate** (owner concern about backtracking): in-process benchmark, median of 7 runs — ui.agency 168.1ms (branch) vs 177.4ms (main); policy.agency 55.7ms vs 59.0ms. No regression; the marginal improvement is likely the item reorder short-circuiting earlier.

## Deliberate TS divergence

`A & never` ERRORS (`intersection expects an object type, got 'never'`) instead of absorbing to `never` — a silently never-typed schema is a debugging trap in a schema-producing language. Documented in the dev notes.

## Verification

7810 lib tests green; linter and make clean; fallback-proof codegen pins observed red first (`const Person = z.string()`) before the `deepResolveNode` routing; `hasAnyValidateTag` gate covered (validated shared key reaches the descriptor); fixture with zero churn elsewhere; accept/reject execution test; seven formatter round-trips including the three paren cases the plan review demanded (`keyof (A & B)`, `(A & B)[]`, `(A & B)["id"]`); re-runs of the keyof, utility-types, and recursive-types execution suites guarding the union-item restructure.

## Scope-outs

Non-object operands (including TS-style union distribution), located diagnostics for resolver errors (standing family follow-up, tripwire-pinned), user-facing guide docs (pending an owner call on where those should live).

### DIFF
--- a/packages/agency-lang/lib/typeChecker/typeOperators.ts
+++ b/packages/agency-lang/lib/typeChecker/typeOperators.ts
@@ -103,6 +103,10 @@ function mergeObjects(
  * one group holding each side's declaration.
  */
 function groupPropertiesByKey(objects: ObjectType[]): ObjectProperty[][] {
+  // keyOrder preserves first-seen order; a plain Record iterated via
+  // Object.entries (or the groupBy in effectPayloadCheck.ts) reorders
+  // integer-like keys ("0", "1") ahead of string keys, and all-digit
+  // property keys are parseable.
   const keyOrder: string[] = [];
   // Null-prototype dict: keys are user-controlled property names, so
   // "__proto__"/"toString" must not collide with Object.prototype
--- a/packages/agency-lang/lib/typeChecker/typeOperators.ts
+++ b/packages/agency-lang/lib/typeChecker/typeOperators.ts
@@ -1,4 +1,7 @@
 import type { VariableType } from "../types.js";
+import type { ObjectProperty, ObjectType } from "../types/typeHints.js";
+import { formatTypeHint } from "../utils/formatType.js";
+import { mergeTagSets } from "./mergeTags.js";
 import { resolveKeysArg, resolveObjectArg } from "./builtinGenerics.js";
 import { NEVER_T } from "./primitives.js";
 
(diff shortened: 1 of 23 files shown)


packages/agency-lang/lib/typeChecker/typeWalker.ts
```diff
@@ -99,6 +101,8 @@ export function visitTypes(
       return visitTypes(t.inner, visit);
     case "keyofType":
       return visitTypes(t.operand, visit);
+    case "intersectionType":
+      return t.types.some((m) => visitTypes(m, visit));
```
egonSchiele: Minor style consistency: every other multi-child case in `visitTypes` uses the explicit early-exit loop (`for (const m of t.types) if (visitTypes(m, visit)) return true; return false;` — see `unionType`, `genericType`, `blockType`, `objectType` right here). This one uses `.some()`.

It's functionally identical — `.some` short-circuits on the first truthy and returns the OR, which is exactly what `visitTypes`' early-exit contract (`if (visit(t) === true) return true`) wants, and side-effecting callers like `validateTypeReferences` return `void` so they never short-circuit. So no bug. But matching the sibling `for...of` idiom would keep the walker uniform (the codebase's inconsistent-patterns guidance). Your call — happy to leave it if you prefer the terser form.

packages/agency-lang/lib/typeChecker/typeOperators.ts
```diff
@@ -52,3 +55,117 @@ export function evalIndexedAccess(
...
+  return { type: "objectType", properties };
+}
+
+/**
+ * Every declaration of every key, grouped, in first-seen key order.
+ * A disjoint key produces a one-element group; a shared key produces
+ * one group holding each side's declaration.
+ */
+function groupPropertiesByKey(objects: ObjectType[]): ObjectProperty[][] {
+  const keyOrder: string[] = [];
```
egonSchiele, line 110: Worth one line on *why* the explicit `keyOrder` exists, because it's load-bearing and non-obvious. There's already a private `groupBy` helper in `effectPayloadCheck.ts` that returns a `Record` iterated via `Object.entries` — and object-type property keys are parsed as `many1WithJoin(varNameChar)`, so an all-digit key like `"0"` is representable. `Object.entries` (and therefore that `groupBy`) reorders integer-like keys ahead of string keys, which would corrupt the merged property order. The `keyOrder` array is what makes this robust.

The risk without a comment: a future reader sees the loop, notices `groupBy` exists, and "simplifies" it into `groupBy(...) + Object.entries`, silently reintroducing a numeric-key reordering bug. A one-liner like `// preserve first-seen order; Object.entries reorders integer-like keys ("0", "1")` turns the decision into a guardrail.

egonSchiele (commented):
Reviewed against the approved plan and spec on this branch; I traced the merge pipeline, parser reorder, resolver branch, all fan-out sites, and the printer paren cases against `origin/main`. This is a faithful, well-covered implementation and every finding from the plan review is addressed in code (the three `keyof (A & B)` / `(A & B)[]` / `(A & B)["id"]` round-trips, the injected `TypeEquals` comparator keeping the CYCLE RULE literally true, the `blockTypeParser`-first ordering, and the unknown-alias-in-intersection test proving the `visitTypes` wiring).

Two things stand out as genuinely strong:
- The **array-before-object reorder** in `intersectionItemParser` is a real latent-bug catch, not busywork. The old object-first order was only safe because the union parser's >=2-member gate discarded single-member matches; the intersection passthrough succeeds on singletons, so it would have stranded the `[]` suffix. Fixing the pre-existing union-member variant (`{ a: number | { n: string }[] }`) as a side effect, with regression tests pinning all three shapes, is exactly right.
- The **declarative RESOLVE/GROUP/COMBINE/BUILD pipeline** reads cleanly, and `combineGroup`'s initial-value-less `reduce` correctly avoids a `length === 1` special case.

Only two minor, non-blocking nits inline — both cosmetic/robustness, neither blocks merge. Nice work closing the roadmap.

packages/agency-lang/lib/utils/formatType.ts
```diff
@@ -33,10 +33,14 @@ export function formatTypeHint(
...
-      // written, while `keyof User[]` would re-parse as keyof (User[]).
+      // Parenthesize keyof and intersection elements: `(keyof User)[]`
+      // and `(A & B)[]` re-parse as written, while the unparenthesized
+      // forms re-parse as keyof (User[]) and A & (B[]).
       const el = recurse(vt.elementType);
-      return vt.elementType.type === "keyofType" ? `(${el})[]` : `${el}[]`;
+      return vt.elementType.type === "keyofType" ||
+        vt.elementType.type === "intersectionType"
+        ? `(${el})[]`
+        : `${el}[]`;
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 73276f5b. Good catch — `typeToString.ts` (the codegen printer) already parenthesized union elements, but `formatTypeHint` did not, so `(A | B)[]` printed as `A | B[]`. Added the `unionType` case to the arrayType paren rule and a new co-located `formatType.test.ts` pinning all four rows (union, intersection, keyof, bare).

packages/agency-lang/lib/typeChecker/typeWalker.ts
egonSchiele: Switched to the explicit early-exit loop in 73276f5b — agreed that keeping the walker uniform is worth more than the terser form.

packages/agency-lang/lib/typeChecker/typeOperators.ts
egonSchiele: Comment added in 73276f5b, naming both the Object.entries integer-key reordering and the `groupBy` in effectPayloadCheck.ts as the tempting-but-wrong simplification. All-digit keys are indeed parseable (`many1WithJoin(varNameChar)`), so the guardrail is real, not theoretical.

Issue #504, opened 2026-07-11
Policy files cannot match on MCP tool-call arguments (only server + tool)


egonSchiele:
## Summary

Policy files can gate MCP tool calls by **server** and **tool** name, but cannot match on the tool's **arguments**. For the planned `mcp::call` effect (payload `{ server, tool, args }`), `server` and `tool` are flat strings and are matchable (exact or glob), but `args` is a nested object and is not reachable from a policy rule. This limits how precisely a saved policy can auto-approve or auto-reject MCP calls without dropping into a code-level `handle` block.

Context: this comes out of the agent MCP integration design/plan (`docs/superpowers/specs/2026-07-10-agent-mcp-integration-design.md`, `docs/superpowers/plans/2026-07-10-agent-mcp-integration.md`). It applies to any effect whose useful discriminator lives inside a nested payload field, not just MCP.

## What works today

Match on `server` and `tool` (both flat strings), including globs:

```json
"mcp::call": [
  { "match": { "server": "filesystem", "tool": "read_file" }, "action": "approve" },
  { "match": { "tool": "read_*" },                            "action": "approve" },
  { "match": { "server": "github" },                          "action": "reject" }
]
```

## What you cannot do

Match on an argument value:

```json
// Does NOT work — nested field paths are not supported by the matcher.
{ "match": { "args.path": "/tmp/*" }, "action": "approve" }

// Does NOT work — `args` is an object; the matcher stringifies it to
// "[object Object]", so this matches nothing useful.
{ "match": { "args": "/tmp/*" }, "action": "approve" }
```

Concrete cases users will reasonably expect to express in a policy file, but currently can't:

- "Approve `filesystem/read_file` only when `path` is under `/tmp`."
- "Approve `github/create_issue` only when `repo` is `my-org/sandbox`."
- "Reject any `*/delete_*` call whose `path` is outside a scratch directory."

## Why (root cause)

In `lib/runtime/policy.ts`, `matchesRule` reads `interrupt.data?.[key]` for each match key — **top-level fields only**, no descent into nested objects — and compares by stringified value (picomatch glob / exact). So:

1. There is no path syntax to reach `args.path`.
2. Even `{ "match": { "args": ... } }` compares against `String(argsObject)` = `"[object Object]"`.

[...]

PR #506, opened 2026-07-11
Agent MCP integration: policy-gated tools from external MCP servers

egonSchiele:
## What

Lets the built-in agency agent use tools from external MCP (Model Context Protocol) servers — filesystem, GitHub, databases, and the rest of the ecosystem. Every MCP tool call is gated through the agent's existing approval policy, so a side-effectful call cannot happen without approval.

MCP support ships as the separately-installed `@agency-lang/mcp` package. When it is not installed, the agent runs normally with zero MCP tools.

## How it works

- **`std::mcp` stdlib module** (`stdlib/mcp.agency` backed by `lib/stdlib/mcp.ts`) is the seam the agent imports. It loads the optional `@agency-lang/mcp` package through a bundled `.mjs` bridge (the sole dynamic import, outside TS lint — same pattern as `smoltalk-llama-cpp`).
- **Fail-closed gating.** Each MCP tool is wrapped so invoking it raises a new `mcp::call` interrupt (via `agency.interrupt`) that flows through `cliPolicyHandler`. The underlying JSON-RPC fires **only on an explicit approve**; reject, propagate, or any non-approve outcome blocks the call.
- **Policy matching.** `mcp::call` carries `{ server, tool, args }`. `server` and `tool` are matchable in the policy file (exact or glob); `args` is a nested object, so argument-level gating uses a `handle mcp::call` block (documented). An interactive "always" decision keys off `server`+`tool` only (via `ALWAYS_FIELDS`), so it keeps matching later calls with different args.
- **Config.** Servers are declared in `agency.json` (project) merged with the agent's `settings.json` (global); project wins on collision. This required a small package addition: `mcpRaw(name, { config, onOAuthRequired })` exposing `McpManager`'s existing config injection, since the public `mcp()` reads config only from `agency.json`.
- **Coordinator-scoped, session-start load.** Tools are loaded once at startup into a composable coordinator tool list. A failing server is skipped with a warning; teardown is handled by the package's own `onAgentEnd` hook.
- **OAuth.** The package already owns the browser-open flow (macOS-only today) and token caching at `~/.agency/tokens/`; the agent's handler only notifies. First-time interactive auth is required once; cached tokens then work headlessly.
- **`/mcp`** lists configured servers and the loaded tool count.

## Package change

`packages/mcp`: adds `mcpRaw` (config injection + raw `callTool` access), re-exports `readMcpConfig`/`mcpToolToAgencyFunction`, and a `MCP_PACKAGE_VERSION` that resolves from both `src` and the built `dist`.

## Design note added during implementation

`@agency-lang/mcp` is not a dependency of `agency-lang`, so — exactly like `smoltalk-llama-cpp` — it is not resolvable from a monorepo checkout. The resolver now honors an `AGENCY_MCP_PATH` override (dev/test + advanced installs), mirroring the llama-cpp escape hatch. A real user installs the package alongside `agency-lang` (or globally), where Node's `node_modules` walk resolves it.

## Testing

- Unit (TS): `mcpRaw` config-injection + real version; resolver (unresolvable-in-monorepo, `AGENCY_MCP_PATH` override, idempotence); the fail-closed gate (approve / reject / **no-handler propagate**); the loading seam (unavailable → `[]`, wrap tools, per-server failure isolation + warn, version-unverified skip); `checkPolicy` on `mcp::call` (server/tool matchable, args not).
- Agency execution: `mcp::call` gating end to end (approve proceeds, reject blocks, approve persists across different args); `buildScopedMatch` pins server+tool and excludes args.
- Integration (opt-in, `RUN_MCP_INTEGRATION=1`, skipped in CI): loads gated tools from the real `@modelcontextprotocol/server-filesystem` over stdio — **verified passing locally end to end**.
- 54 existing `@agency-lang/mcp` tests still pass.

## v1 scope / limitations

Tools only (no MCP resources/prompts); coordinator-only (per-subagent scoping deferred); rejection currently surfaces as a string result rather than the tool loop's `isRejected` accounting (deferred, documented). CLI/slash `mcp add` server management and nested-argument policy matching are tracked as follow-ups (the latter is issue #504).


packages/agency-lang/docs/site/.vitepress/config.mts
```diff
@@ -82,6 +82,7 @@ export default defineConfig({
             },
             { text: "Agency Config File", link: "/guide/agency-config-file" },
             { text: "Testing", link: "/guide/testing" },
+            { text: "MCP servers (agent)", link: "/guide/mcp-agent" },
```
egonSchiele: remove this from the guide

packages/agency-lang/docs/site/agent/mcp.md
```diff
@@ -0,0 +1,106 @@
+# Using MCP servers with the agent
```
egonSchiele, line 1: create a new top-level "agent" section to put this under, instead of the guide

packages/agency-lang/lib/stdlib/mcpBridge.mjs
```diff
@@ -0,0 +1,36 @@
...
+// import of the optional @agency-lang/mcp package is dynamic and lives here — a
+// plain .mjs shipped via the Makefile copy, outside TS lint — so the package
+// stays a non-dependency (mirrors lib/stdlib/providers/llama-cpp.mjs).
+// mcpResolver.ts exposes the resolved entry via AGENCY_MCP_PATH so a global
+// install works even when the bare specifier is not resolvable from here.
+import { pathToFileURL } from "node:url";
+
+let _mod = null;
+
+async function load() {
```
egonSchiele, line 42: please explain how the bridge works

packages/mcp/src/mcp.ts
```diff
@@ -1,18 +1,32 @@
...
+  callTool: (server: string, tool: string, args: Record<string, unknown>) => Promise<string>;
+};
+
+function getManager(
+  onOAuthRequired?: (data: any) => void | Promise<void>,
+  config?: Record<string, McpServerConfig>,
+): McpManager {
   if (!singleton) {
-    const config = readMcpConfig();
+    const resolved = config ?? readMcpConfig();
```
egonSchiele, line 33: Config injection is **first-caller-wins**: `getManager` only applies `config` when `!singleton`. In the agent flow `mcpRaw` is the first caller (`_loadMcpTools` primes the singleton on the first server), so this is correct today. But if `mcp()` ever runs before `mcpRaw()` in the same process, it creates the singleton from `agency.json` alone and this injected `config` is silently dropped — the same shape as the `onOAuthRequired` first-caller-wins warning just below. Worth a one-line doc comment here (or an explicit warn when `config` is passed but ignored) so a future second caller doesn't get its config silently discarded.

packages/agency-lang/lib/stdlib/mcpBridge.mjs
```diff
@@ -0,0 +1,36 @@
...
+  if (_mod) {
+    return _mod;
+  }
+  const pkgPath = process.env.AGENCY_MCP_PATH;
+  _mod = pkgPath
+    ? await import(pathToFileURL(pkgPath).href)
+    : await import("@agency-lang/mcp");
```
egonSchiele, line 49: **Seam coverage gap:** the real bridge→package path is never exercised in CI. `mcp.test.ts` mocks this module, `mcpResolver` returns `null` in a monorepo checkout (so `isMcpAvailable()` is `false`), and `mcp.integration.test.ts` is opt-in / skipped. So a signature drift between this bridge and the package's real exports — `readMcpConfig`, `mcpToolToAgencyFunction`, `MCP_PACKAGE_VERSION`, `mcpRaw` — would pass CI green and only break at runtime for real users.

Since `@agency-lang/mcp` is a workspace sibling, consider a lightweight contract test that sets `AGENCY_MCP_PATH` to the built package and asserts these four exports exist and shape-match (no need to spawn a server). That's the one integration the mocks can't protect, and it's exactly the wiring most likely to rot.

packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -864,6 +868,23 @@ def _runTurn(msg: string): boolean {
...
+    const merged = mergeMcpServers(getMcpServers(loadSettings()), readProjectMcpConfig(cwd()))
+    if (keys(merged).length == 0) {
+      pushMessage(color.dim("No MCP servers configured. Add an mcpServers block to agency.json or settings.json."))
+      return true
+    }
+    pushMessage(color.bold("MCP servers:"))
+    for (name in keys(merged)) {
+      pushMessage("  ${name}")
+    }
+    pushMessage(color.dim("Loaded ${mcpTools.length} MCP tool(s) this session."))
```
egonSchiele, line 892: `/mcp` lists server names from the merged config plus a single total tool count, but no **per-server connection status**. A server that failed to connect at startup still appears in this list with no indication it's broken, and the count reflects only the servers that loaded — so a user seeing "2 servers, 5 tools" can't tell that one is down. The design note called for per-server status (connected / failed / needs-auth). You have the merged config here and `mcpTools` for the loaded set; consider annotating each server with whether it actually loaded.

packages/agency-lang/lib/stdlib/mcpGate.ts
```diff
@@ -0,0 +1,37 @@
...
+            effect: EFFECT,
+            message: `MCP tool call: ${server} → ${tool}`,
+            data: { server, tool, args },
+          }),
+        ),
+    );
+    if (isApproved(response)) {
+      return realCallTool(server, tool, args);
+    }
+    return `The MCP call to ${server}/${tool} was not approved by the policy.`;
```
egonSchiele, line 35: Agree with deferring this (noted in the PR body) — just flagging the consequence for the tracking issue: returning a string means the tool loop treats a policy **rejection** as a normal successful result. `prompt.ts`'s `isRejected(toolResult)` path never fires, so statelog won't record it as a rejection and the model sees a plain string rather than the loop's rejection message. Fine for v1; worth capturing in the follow-up so the statelog-accounting gap is intentional rather than incidental.

egonSchiele (commented):
## Review — Agent MCP integration

Reviewed the full diff against the codebase. Clean, well-tested implementation, and it resolves the concerns raised on the plan — nice work.

**Strengths**
- **Fail-closed gate** (`isApproved`, not `!isRejected`) with a dedicated **no-handler propagate test** — the exact safety property, locked by a regression test.
- **Local structural types** in `lib/stdlib/mcp.ts` instead of `import type … from "@agency-lang/mcp"`, with a comment explaining the pnpm `TS2307` / non-dependency reasoning — avoids a build break and matches the `localModels.ts` precedent. The `mcpBridge.d.mts` for the `.mjs` is a nice touch.
- **Absent-safe** `_loadMcpToolsForServer` (guards availability + try/catches the bridge import), hardened `_readProjectMcpConfig` (zod `.parse` throw → warn + `{}`), and **prime-then-parallel** loading with correct reasoning about the non-concurrency-safe singleton.
- The `buildScopedMatch` execution test proves the `mcp::call` "always" key pins `server`+`tool` and **excludes** `args` — the real coverage for that behavior. The `checkPolicy` test proves `args` isn't matchable.
- Seam correctly routed through `std::mcp` (`stdlib/mcp.agency` → `agency-lang/stdlib-lib/mcp.js`), effect declared in-graph, docs added.

**Notable (see inline comments)**
1. Config injection is first-caller-wins on the package singleton (`mcp.ts`) — latent footgun.
2. The real bridge→package seam isn't covered in CI (bridge mocked, integration opt-in) — a signature drift would ship green.
3. `/mcp` shows names + a total count but no per-server status; a failed server appears with no indication.
4. Rejection-as-string bypasses the tool loop's `isRejected` accounting (already deferred; flagging the statelog gap for the follow-up).

**Minor / nits**
- `_loadMcpTools` and `_loadMcpToolsForServer` both call `isMcpAvailable()` / `exposeResolvedMcpPath()` (redundant, harmless).
- Large `args` aren't truncated in the approval prompt (spec noted this as a flood risk).
- The `approvesTwiceDifferentArgs` agency-js node proves the payload carries `args` per-call, but not the `ALWAYS_FIELDS` keying itself (that's the `buildScopedMatch` test — the node name slightly overpromises).

**Verdict:** No blocking issues. Correctness, layering, and test design are sound. The main gap is the uncovered real seam (point 2), worth a lightweight contract test before this is relied on. Everything else is polish.

packages/mcp/src/mcp.ts
```diff
@@ -1,18 +1,32 @@
+import { createRequire } from "node:module";
```
egonSchiele, line 1: version bump this package + add a changelog so I can publish it. Copy the changelog format of the agency-lang package.

egonSchiele:
Thanks for the review — all comments addressed in the four follow-up commits (2e29b1c, ec0c082, 4d69009, f713dd0).

**Copilot**
- **Prototype pollution in `_mergeMcpServers`** — now merges into a `Object.create(null)` target, so a `__proto__` server name (which the config name regex permits) becomes a plain data key and can't touch a prototype. Test added.
- **`/help` omitted `/mcp`** — added.

**egonSchiele**
- **Docs → new top-level "Agent" section** — added an `Agent` nav entry + sidebar (Overview + MCP servers), moved the page to `/agent/mcp`, and removed it from the Tooling guide list.
- **Explain the bridge** — `mcpBridge.mjs` now documents why it's a `.mjs`, how `load()` resolves the package (`AGENCY_MCP_PATH` then bare import), caching, and what it exposes.
- **First-caller-wins config injection** — `getManager` now warns when a `config` is passed after the singleton already exists, instead of silently dropping it; documented inline.
- **Seam coverage gap** — added `mcpBridge.contract.test.ts`: points `AGENCY_MCP_PATH` at the built workspace package and asserts the four exports exist and shape-match (no server spawn). It runs (not skipped) whenever the package is built, so signature drift is caught in CI rather than at runtime.
- **`/mcp` per-server status** — `/mcp` now shows `connected` / `unavailable` / `not loaded` per server (via a status map from the new `_loadMcpToolsWithStatus`), not just a total count. (`needs-auth` as a distinct state is noted for a follow-up — hard to distinguish from a plain connect failure today.)
- **Rejection-as-string statelog gap** — captured as a tracking issue: #508.
- **Version bump + changelog** — `@agency-lang/mcp` → 0.0.2 with a `CHANGELOG.md` in the agency-lang format, ready to publish.

Tests: 43 agency-lang unit tests + 54 mcp package tests + the agency-js gating tests all pass; the real filesystem-server integration test still passes locally.

Issue #471, opened 2026-07-09
Finish flow-typed narrowing migration: delete scope-chain path, automate FlowEnvironment memo invalidation


egonSchiele:
The narrowing docs (`docs/dev/typechecker/narrowing/README.md`) describe the checker as mid-migration from scope-chain narrowing to the flow-typed model, and the flow graph + `typeAt` oracle have landed. But the legacy scope-chain path (`walkWithNarrowing`, throwaway child scopes, `declareLocal`) still coexists with it. `narrowByRefine` is shared, but the application logic lives twice.

The seam between the two models is what generates the manual contracts scattered through the code:

- `strictMemberAccessSeverity` returns `silent` whenever `ctx.flowEnv` is unset, to suppress false positives during pre-flow passes (`lib/typeChecker/synthesizer.ts`).
- The `FlowEnvironment.memo` soundness contract ("valid only for a single type-check pass; discard the memo if scope contents change") is enforced purely by vigilance. Two passes have already had to individually learn it: `handlerParamTyping` (moved ahead of the flow build in `index.ts` to avoid a reset) and `computeMatchExprTypes` (resets the memo AND reaches in to patch eagerly-snapshotted `assign` flow nodes via `matchConsumerAssignFlows`).

Each patch is correct and well-commented, but they are symptoms of the dual model.

## Proposed work

1. **Finish the migration:** route all narrowing application through the flow graph and delete the scope-chain path (`walkWithNarrowing` and the child-scope refinement machinery in `lib/typeChecker/scopes.ts`). This removes the "passes disagree" class of bugs the flow model was built to kill.
2. **Automate memo invalidation:** add a generation counter to `Scope` (bump on `declare`/rebind; memo entries record the generation they were computed under; stale entries recompute). This deletes the human-enforced memo contract for roughly 20 lines of code. Alternative: freeze scopes after `buildFlowGraphs` and require all retyping passes to run before it — handler-param retyping already does this; `computeMatchExprTypes` is the remaining holdout.

egonSchiele:
Status update: **part 2 (automate memo invalidation) is implemented** — PR incoming, built on the generation-counter design in `docs/superpowers/specs/2026-07-10-flowenv-generation-counter-design.md` (on the PR branch).

**Part 1 (finish the migration / delete the scope-chain path) is deliberately deferred** after an assessment of the current code. Two findings changed the calculus:

[...]

PR #507, opened 2026-07-11
Automate typeAt memo invalidation with a Scope generation counter (#471 part 2)

egonSchiele:
Implements issue #471 part 2, per the spec and reviewed plan on this branch (part 1, the walk fusion, is deliberately deferred — decision recorded on the issue and in the narrowing README).

## What

`typeAt` memoizes per (flow node, reference key), but `start` nodes read scopes live — so any pass that mutates a scope after answers were cached made the memo silently stale. The protection was a prose contract ("discard the memo if scope contents change") enforced in three separate homes: a manual reset in `computeMatchExprTypes`, a throw-on-reorder in `handlerParamTyping`, and vigilance. This PR makes invalidation automatic:

- **`Scope` gains a tree-wide generation counter** (single integer on the root scope; `declare()` bumps it, O(1) to read from the flow env).
- **The memo becomes a box** (`FlowMemo = { gen, map }`) behind a `readonly` field. `typeAt` compares the box generation to the tree generation on entry and lazily rebuilds the map on mismatch — the same semantics as the manual resets, with no human in the loop.
- **Throwaway `child()` scopes are `detached` and skip the bump.** They are never flow-reachable (now asserted in `buildFlowGraphs`), and bumping on them would flush the memo on every lambda synthesis in `checkScopes`.

Deleted/relaxed: the manual reset in `computeMatchExprTypes` (its phase-2 `declare()` calls ARE the bump, covering the paired in-place `assignFlow.type` patch), and the soundness-cliff framing in `handlerParamTyping` + `index.ts` (the pre-flow ordering is kept — it avoids a whole-memo flush — but a reorder is no longer silently unsound).

## Reviewer attention

1. **The box is shared by reference** across the per-scope build envs AND the synthesizer's spread copies (`{ ...ctx.flowEnv, typeAliases }`). Invalidation mutates `memo.map`/`memo.gen` in place; the `readonly` on the field makes replacing the box (the old reset pattern) a compile error. Two pins cover the copies explicitly (invalidation visible through a copy; memo hit shared through a copy).
2. **Identity-based pins.** The "does NOT flush" tests assert object identity (`toBe`) on a join result — recomputing constructs a fresh union object via `uniteTypes`, so identity proves a memo hit. Noted limitation: if `uniteTypes` ever memoizes internally, those pins need a new observable.
3. **The integration pin** (`matchExpression.test.ts`) exercises the production flush point end-to-end: two sequential expression-matches where phase 1 memoizes the first consumer's pre-patch `any`, and a downstream `const n: number = first` whose diagnostic is suppressed by the stale entry. It fails with invalidation disabled.

## Red-proof finding

With the generation check disabled and the manual reset deleted, exactly the 5 new pins fail — **zero pre-existing tests** detect the staleness. The old manual reset had no independent test coverage before this PR; the contract was enforced purely by the comment.

## Verification

Full lib suite green; `make` + structural linter clean; `tsc --noEmit` clean. Perf gate (ui.agency typecheck, 3 warmup + 25 timed, median): branch 23.6 / 23.9 ms vs main 24.1 / 23.1 ms across two rounds (within noise, gate passed) — the counter only invalidates where the manual reset fired before, so steady-state flush behavior is unchanged.

### DIFF
--- a/docs/superpowers/specs/2026-07-10-flowenv-generation-counter-design.md
+++ b/docs/superpowers/specs/2026-07-10-flowenv-generation-counter-design.md
@@ -1,5 +1,12 @@
 # FlowEnvironment memo auto-invalidation via a Scope generation counter
 
+> **Errata (point-in-time record; see the plan rev-2 header for the corrected
+> versions):** two claims below were superseded by later review rounds —
+> "O(depth) (depth ≤ 3)" (narrowing child chains nest arbitrarily; the shipped
+> `scope.ts` comment has the corrected wording, and the hot-path read is O(1)
+> via the parent-less flow-env root) and "median of 7" in the Tests section
+> (the shipped perf gate is 3 warmup + 25 timed iterations).
+
 **Issue:** #471 (part 2). Part 1 of that issue — fusing flow-graph construction
 into the scope walk and deleting the legacy child-scope narrowing path — is
 **deliberately descoped** after owner review (2026-07-10): its remaining unique
--- a/docs/superpowers/specs/2026-07-10-flowenv-generation-counter-design.md
+++ b/docs/superpowers/specs/2026-07-10-flowenv-generation-counter-design.md
@@ -1,0 +1,140 @@
+# FlowEnvironment memo auto-invalidation via a Scope generation counter
+
+**Issue:** #471 (part 2). Part 1 of that issue — fusing flow-graph construction
+into the scope walk and deleting the legacy child-scope narrowing path — is
+**deliberately descoped** after owner review (2026-07-10): its remaining unique
+benefits (unrequested precision gains, ~150 deleted lines, conceptual
+cleanliness) do not justify 2–3 weeks. The dual "where to apply narrowing"
+logic is inert — both paths share `analyzeCondition` (fact production) and
+`narrowByRefine` (fact application), so a new narrowing form lands in both
+automatically. This spec covers the part that has actually caused repeated
+work: the human-enforced `FlowEnvironment.memo` soundness contract.
+
+## Problem
+
+`typeAt` (lib/typeChecker/flow.ts) memoizes per (flow node, reference key).
+`start` nodes read `scope.lookup(...)` live, so a memoized answer embeds scope
+state at query time. Any pass that mutates a scope after answers were cached
+makes the cache stale, silently. The current protection is a doc comment
+("SOUNDNESS CONTRACT: ... discard the memo") enforced by vigilance. Two passes
+have independently paid for it:
+
+- `handlerParamTyping` was moved ahead of `buildFlowGraphs` (index.ts) solely
+  to avoid needing a reset.
+- `computeMatchExprTypes` must remember `ctx.flowEnv.memo = new WeakMap()`
+  after patching consumer scope entries and `assign` flow nodes.
+
+## Mechanism
+
+One integer per check run — the **generation** — owned by the `Scope` tree.
+
+1. **`Scope` gets a `generation` counter.** `declare()` bumps it, propagating
+   the bump up the parent chain to the root (`topLevelScope`), so any scope can
+   read the tree-wide current generation in O(depth) (depth ≤ 3). The root's
+   counter is the single source of truth.
+2. **The memo becomes a shared box:** `FlowEnvironment.memo` changes from
+   `WeakMap<FlowNode, Record<string, ScopeType>>` to
+   `{ gen: number; map: WeakMap<FlowNode, Record<string, ScopeType>> }`.
+   The box is shared **by reference** across the per-scope build envs and the
+   spread-copied envs in the synthesizer (`{ ...ctx.flowEnv, typeAliases }`),
+   so a reset made through any env is visible to all. (Replacing a `memo`
+   field on one env object would not be — this is why the box exists.)
+3. **`typeAt` checks on entry:** if `box.gen !== scope tree's current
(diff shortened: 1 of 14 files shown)


packages/agency-lang/lib/typeChecker/flow.test.ts
```diff
@@ -538,3 +539,111 @@ describe("PathSegment helpers + index resolution (M2)", () => {
     expect(typeof after === "object" && after.type).toBe("resultType");
   });
 });
+
+describe("memo auto-invalidation (generation counter)", () => {
+  it("returns the fresh type after a scope mutation (stale on main)", () => {
```
egonSchiele: Nit: "(stale on main)" will read confusingly once this IS main — a future reader will wonder why a passing test claims to be stale on the branch they are on. Suggest "(stale without the generation check)", which also documents the red-proof knob.

docs/superpowers/specs/2026-07-10-flowenv-generation-counter-design.md
```diff
@@ -0,0 +1,140 @@
...
+- `computeMatchExprTypes` must remember `ctx.flowEnv.memo = new WeakMap()`
+  after patching consumer scope entries and `assign` flow nodes.
+
+## Mechanism
+
+One integer per check run — the **generation** — owned by the `Scope` tree.
+
+1. **`Scope` gets a `generation` counter.** `declare()` bumps it, propagating
+   the bump up the parent chain to the root (`topLevelScope`), so any scope can
+   read the tree-wide current generation in O(depth) (depth ≤ 3). The root's
```
egonSchiele, line 40: Record-keeping nit, non-blocking: the committed spec still carries two claims the later review rounds corrected — "depth ≤ 3" here (narrowing child chains nest arbitrarily; the shipped `scope.ts` comment has the corrected wording) and "median of 7" in the Tests section (the shipped gate is 3 warmup + 25 timed). Fine if the spec is meant as a point-in-time record, but a one-line errata note at the top pointing at the plan rev-2 header would stop a future reader from trusting the stale numbers.

egonSchiele (commented):
## Review — verified against the branch, not just read

**Verdict: this is correct, well-pinned, and lands exactly what the rev-2 plan promised. No blocking findings.** I independently reproduced the PR's central claims on a checkout of `6ce6ad259` rather than taking the body's word for them:

### Independent verification

- **Green run:** `scope.test.ts` + `flow.test.ts` + `flowBuilder.test.ts` + `matchExpression.test.ts` — 113/113 pass on the branch.
- **Red-proof reproduced:** with the generation check in `typeAt` commented out, the full `lib/typeChecker` suite shows **exactly 5 failures and 1438 passes** — the staleness pin, the assign-patch pin, the spread-copy invalidation pin, the lazy-semantics pin, and the integration pin. Zero pre-existing tests detect the staleness, confirming the PR body's finding that the old manual reset was never independently covered. The two identity pins and the standalone-tree pin stay green without the check, exactly as they should (they test caching, which exists either way).
- **Integration pin comment accuracy:** `computeMatchExprTypes` phase 1 iterates `ids.sort((a, b) => b - a)` — descending — so the test's premise (the second match, with the higher id, synthesizes first and memoizes the first consumer's pre-patch `any`) matches the code, not just the story.
- **Mechanism trace:** `declare()` → `functionScope()` → `bumpGeneration()` → root, with `ctx.flowEnv.scope` being the parent-less `topLevelScope`, so the hot-path generation read is one property access. The `readonly memo` field makes the old reset pattern (`env.memo = ...`) a compile error while leaving in-place `memo.map`/`memo.gen` mutation (the box contract) intact. The detached carve-out is enforced by construction (`child()` sets the flag) and asserted at the flow root (`buildFlowGraphs`).

### What I checked and found clean

- All three homes of the old prose contract are updated consistently: the manual reset deleted (`matchExprTypes.ts`), the `handlerParamTyping` ordering assertion kept with a rewritten, now-true justification, and the `index.ts` ordering comments softened to "oracle-seeding quality, not soundness cliff". The FlowMemo comment correctly scopes "automatic" to Scope mutations and names the residual paired-declare contract for in-place FlowNode patches.
- Coding standards: braced ifs in new code, no interfaces/maps/sets, no module-global per-run state (counter lives on the per-run Scope tree), no apostrophes in the assertion message.
- Perf: the reported 3-warmup/25-run medians (23.6/23.9 vs 24.1/23.1) are within the gate, and the steady-state argument holds — the only post-flow main-tree `declare` sites are the phase-2 patches where the manual reset already fired.

Two tiny non-blocking nits as inline comments (a test title that will read confusingly after merge, and two spec-doc numbers the later review rounds corrected but the committed spec still carries).

packages/agency-lang/lib/typeChecker/flow.test.ts
egonSchiele: Renamed to "(stale without the generation check)" in f91fbf83 — agreed, and it doubles as documentation of the red-proof knob.

docs/superpowers/specs/2026-07-10-flowenv-generation-counter-design.md
egonSchiele: Errata blockquote added at the top of the spec in f91fbf83, naming both superseded claims (depth <= 3 and median-of-7) and pointing at the plan rev-2 header + the shipped scope.ts comment for the corrected versions. Kept the body untouched as the point-in-time record.

PR #509, opened 2026-07-11
MCP server management: agency agent mcp add/remove/list + /mcp add|remove

egonSchiele:
## What

Adds `mcp add` / `remove` / `list` server management so users configure the agent's MCP servers without hand-editing JSON — the v2 follow-up to the MCP integration (#506). Two surfaces:

- **CLI:** `agency agent mcp <list|add|remove>` (scriptable; matches `claude mcp add`).
- **REPL:** `/mcp add`, `/mcp remove`, with **hot-connect** so an added server's tools are usable immediately (no restart).

## How it works

- **Scope-agnostic config core** in `lib/stdlib/mcp.ts`: validate a server through the package schema, then read/upsert/remove it in a target file's `mcpServers` block, **preserving all other top-level keys** (2-space, null-prototype map). One core, reused by both surfaces.
- **CLI** — `lib/cli/agent.ts` intercepts the `mcp` subcommand (pure TS, no agent boot) → `lib/cli/mcpCommand.ts` parses flags and calls the core. Stdio via `--command`/`--args a,b,c`; HTTP via `--url`/`--oauth`.
- **REPL** — `/mcp add|remove` in `_runTurn`; parsing lives in TS (`_parseMcpCommand`) for robustness. `add` writes then hot-connects the single server and folds its gated tools into the live `mcpTools`; `remove` rewrites config and drops that server's tools.
- **Default scope = project** (`agency.json`) for both surfaces; `--global` targets the agent-home `settings.json`. `list` shows each server's source.
- **Validation** reuses the mcp package's existing `McpServersSchema` via a new `validateMcpServers` export (bridged), so bad servers are rejected before write.
- **Secrets:** `--oauth` sets `auth:"oauth"` only; OAuth client credentials are never written to config (env-var at connect time).

## Package change

`@agency-lang/mcp` 0.0.2 → 0.0.3: adds `validateMcpServers(servers)` (schema `safeParse`, no throw) + changelog.

## Testing

- **Config core (TS):** upsert preserves sibling keys / creates file / `__proto__`-safe; remove returns false-missing / true-and-rewrites; validate fails clearly when the package is absent and delegates otherwise; `_addMcpServer` writes on valid and skips on invalid.
- **CLI (TS):** `add` (stdio + http/oauth) writes `agency.json`; `remove` deletes and reports missing; `list` shows source; bad input returns exit 1. Verified end-to-end against the built package (`agency agent mcp add/list`).
- **REPL parsing (TS):** `_parseMcpCommand` for add-stdio/http-oauth-global/remove/errors; `_dropMcpToolsForServer` removes only the named server's tools.
- 60 agency-lang unit tests (7 suites) + 54 mcp-package tests + the existing gating tests all pass; clean build.

## Notes / scope

- `--args` is comma-split (`std::args` has no array flag type); documented that arg values can't contain commas.
- The REPL slash handler is thin glue over the TS core (fully unit-tested) and the shared write path is verified through the CLI end-to-end; a PTY test for the interactive surface is not included.

Spec + plan: on the `agent-mcp-add-design` branch (`docs/superpowers/{specs,plans}/2026-07-11-mcp-server-management*`).


packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -868,6 +879,51 @@ def _runTurn(msg: string): boolean {
...
+    if (parsed.sub == "add") {
+      const res = addMcpServer(parsed.name, parsed.config, target)
+      if (res.ok != true) {
+        pushMessage(color.red("Could not add '${parsed.name}': ${res.error}"))
+        return true
+      }
+      // Hot-connect: load just this server and fold its tools into the session.
+      const merged = mergeMcpServers(getMcpServers(loadSettings()), readProjectMcpConfig(cwd()))
+      const newTools = loadMcpToolsForServer(parsed.name, merged, onMcpOAuth)
+      setMcpTools([...mcpTools, ...newTools])
```
egonSchiele: **Bug: re-adding an existing server duplicates its tools in the live session.**

`addMcpServer` is an upsert, so `/mcp add fs …` is the natural way to *change* an already-connected server's config (new args, different command). But this line appends the fresh tools without dropping the old ones, so after a re-add the session holds two copies of every `fs` tool — same names. Duplicate tool names in the LLM tool list are the exact failure class you fixed before with `.rename()` (API rejects the request).

Suggested fix — drop before folding in:

```
setMcpTools([...dropMcpToolsForServer(mcpTools, parsed.name), ...newTools])
```

(Also worth a `_dropMcpToolsForServer`-style unit assertion that add-then-re-add leaves exactly one set.)

packages/agency-lang/lib/stdlib/mcp.ts
```diff
@@ -142,3 +144,147 @@ export async function _loadMcpTools(
...
+ *  clear "not installed" failure when the package is absent. */
+export async function _validateMcpServers(servers: McpServers): Promise<McpValidation> {
+  if (!isMcpAvailable()) {
+    return { ok: false, error: "@agency-lang/mcp is not installed. Run: npm install @agency-lang/mcp" };
+  }
+  exposeResolvedMcpPath();
+  return mcpBridge.validateMcpServers(servers);
+}
+
+function _readJsonFile(file: string): Record<string, unknown> {
```
egonSchiele: **Data-loss risk: a malformed config file gets silently replaced.**

`_readJsonFile` swallows the parse error and returns `{}`. For reads that's fine, but `_upsertMcpServerInFile` then writes the file back with *only* `mcpServers` — so `agency agent mcp add` against an `agency.json` that has, say, a trailing comma silently destroys the user's model pins, policies, and everything else in that file. The "preserves all other top-level keys" guarantee only holds when the file parses.

Suggest splitting the semantics: reads can stay lenient, but the upsert path should distinguish "file absent" (fine, create it) from "file exists but unparseable/unreadable" (fail with an error and write nothing). `_removeMcpServerFromFile` is already safe by accident (it bails before writing), so it's really just the upsert.

packages/agency-lang/lib/cli/agent.ts
```diff
@@ -1,6 +1,15 @@
 import { AgencyConfig } from "@/config.js";
 import { runBundledAgent } from "./runBundledAgent.js";
+import { mcpCommand } from "./mcpCommand.js";
 
 export function agent(config: AgencyConfig, args: string[] = []): void {
+  // `agency agent mcp …` manages MCP servers in config files (TS, no agent
+  // boot). Everything else launches the bundled agent.
+  if (args[0] === "mcp") {
+    void mcpCommand(config, args.slice(1)).then((code) => {
```
egonSchiele: **Unhandled rejection: no `.catch` on the floating promise.**

`mcpCommand` can reject — `_upsertMcpServerInFile` throws on an unwritable path (EACCES / read-only checkout), and the bridge `validateMcpServers` rejects if the package's entry fails to load. Today that surfaces as a raw unhandled-rejection stack instead of a clean error message.

```ts
void mcpCommand(config, args.slice(1))
  .then((code) => { process.exitCode = code; })
  .catch((err) => {
    console.error(String(err instanceof Error ? err.message : err));
    process.exitCode = 1;
  });
```

packages/agency-lang/lib/cli/mcpCommand.ts
```diff
@@ -0,0 +1,137 @@
...
+    } else if (a === "--project") {
+      flags.scope = "project";
+    } else if (a === "--command") {
+      flags.command = argv[++i];
+    } else if (a === "--args") {
+      flags.args = argv[++i];
+    } else if (a === "--url") {
+      flags.url = argv[++i];
+    } else if (a.startsWith("--")) {
+      // Unknown flag — surface rather than silently ignore.
```
egonSchiele: The comment says unknown flags are surfaced rather than silently ignored, but nothing downstream ever looks at them — they just become positionals and vanish. Worse, an unknown flag *before* the name shifts the positional slots: `agency agent mcp add --gloabl fs --command npx` (typo) makes `positionals[1] === "--gloabl"`, and since the schema's name regex allows dashes you can end up writing a server literally named `--gloabl` while `fs` is dropped.

Suggest actually erroring on `a.startsWith("--")` unknowns (print the flag + usage, exit 1) — that makes the comment true and closes the positional-hijack. Same pattern exists in `_parseMcpCommand` on the REPL side.

packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -868,6 +879,51 @@ def _runTurn(msg: string): boolean {
...
+      let st = mcpServerStatus
+      if (newTools.length > 0) {
+        st[parsed.name] = "connected"
+      } else {
+        st[parsed.name] = "unavailable"
+      }
+      setMcpServerStatus(st)
+      pushMessage(color.green("Added and connected '${parsed.name}' (${newTools.length} tool(s))."))
```
egonSchiele: When the hot-connect fails (`newTools.length == 0`), the status map correctly records `"unavailable"` but the user still sees a green `Added and connected 'x' (0 tool(s)).` — which reads as success. Suggest branching the message: green for connected, yellow `Added 'x' to config, but it failed to connect (see /mcp).` for the 0-tools case.

(Related edge: `/mcp add fs --global` when the project config also defines `fs` — the merged map is project-wins, so the server you just hot-connected is running the *project* config, not the one you just wrote. Probably fine to leave, but the message claims more than it knows.)

packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -868,6 +879,51 @@ def _runTurn(msg: string): boolean {
...
+      return true
+    }
+    // remove
+    const removed = removeMcpServer(parsed.name, target)
+    if (removed) {
+      setMcpTools(dropMcpToolsForServer(mcpTools, parsed.name))
```
egonSchiele: Minor: `removeMcpServer` only says whether the name existed in the *target* file, but this unconditionally drops the server's live tools. If `fs` is defined in both project and global config, `/mcp remove fs` (project scope) leaves the global entry in place yet disconnects the session — it silently diverges from what a restart would load (global `fs` would reconnect). Probably acceptable for v1, but worth either a re-merge check (is the name still present in the merged map? if so, hot-reconnect or at least say so) or a dim note in the output.

egonSchiele (commented):
Reviewed the full diff. The scope-agnostic config core with both surfaces as thin callers is a nice shape, the null-prototype/`__proto__` handling is careful, and test coverage of the core is solid. Two real bugs worth fixing before merge (duplicate tools on re-add, and a data-loss path when the target config file is malformed JSON), plus a few smaller issues — all inline.

**Summary of findings:**

1. **Re-adding an existing server duplicates its tools in the live session** (`agent.agency:906`) — duplicate LLM tool names are a known API-error class in this repo.
2. **`mcp add` silently wipes a malformed config file** (`mcp.ts:165`) — `_readJsonFile` treats unparseable JSON as `{}`, so upsert rewrites the file with only `mcpServers`, destroying every other key the user had.
3. **Unhandled rejection in the CLI dispatch** (`agent.ts:9`) — no `.catch` on the floating promise; an fs write error (e.g. EACCES) surfaces as a raw stack.
4. **Unknown-flag handling contradicts its comment and can hijack the name positional** (`mcpCommand.ts:54`).
5. **Green “Added and connected” message even when the server failed to connect** (`agent.agency:914`).
6. Minor scope-shadowing note on `/mcp remove` (`agent.agency:920`).

Not blocking: the `--args` comma-split limitation is documented, the missing PTY test is acknowledged in the PR body, and the package bump/changelog look right.

packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -24,7 +24,17 @@ import {
   builtinPolicyNames,
   BUILTIN_POLICIES,
  } from "std::policy"
-import { loadMcpToolsWithStatus, mergeMcpServers, isMcpAvailable, readProjectMcpConfig } from "std::mcp"
+import {
```
egonSchiele: All this MCP code is really cluttering this file. Please propose a solution that will move the MCP code into its own file and write it in a way that is declarative instead of imperative.

packages/agency-lang/lib/cli/agent.ts
```diff
@@ -1,6 +1,15 @@
 import { AgencyConfig } from "@/config.js";
 import { runBundledAgent } from "./runBundledAgent.js";
+import { mcpCommand } from "./mcpCommand.js";
 
 export function agent(config: AgencyConfig, args: string[] = []): void {
+  // `agency agent mcp …` manages MCP servers in config files (TS, no agent
+  // boot). Everything else launches the bundled agent.
+  if (args[0] === "mcp") {
```
egonSchiele: This feels like a hack. Can you not use a commander sub-command for this?

packages/agency-lang/lib/cli/mcpCommand.ts
```diff
@@ -0,0 +1,137 @@
...
+  command?: string;
+  args?: string;
+  url?: string;
+  oauth: boolean;
+  scope: Scope;
+};
+
+// Minimal flag parser. `--flag value` for string flags; `--oauth/--project/
+// --global` are booleans. Default scope is project (spec).
+function parseFlags(argv: string[]): Flags {
```
egonSchiele: Feels really silly to parse the flags ourselves when we have two different CLI args parsing libraries in play: commander for parsing args in typescript, and the std::args module for parsing args in agency.

packages/agency-lang/lib/stdlib/mcp.ts
```diff
@@ -142,3 +144,147 @@ export async function _loadMcpTools(
 ): Promise<AgencyFunction[]> {
   return (await _loadMcpToolsWithStatus(merged, onOAuthRequired)).tools;
 }
+
+// ── Config management (mcp add/remove/list) ───────────────────────────────
+// Scope-agnostic: callers pass the target file (project agency.json or the
+// agent-home settings.json). The mcpServers block is read/written while every
+// other top-level key is preserved.
+
+export type McpValidation = { ok: boolean; error?: string };
```
egonSchiele: Agency already has a type for error handling, so why invent a new one?
https://agency-lang.com/guide/error-handling.html

packages/agency-lang/lib/stdlib/mcp.ts
```diff
@@ -142,3 +144,147 @@ export async function _loadMcpTools(
...
+  if (!isMcpAvailable()) {
+    return { ok: false, error: "@agency-lang/mcp is not installed. Run: npm install @agency-lang/mcp" };
+  }
+  exposeResolvedMcpPath();
+  return mcpBridge.validateMcpServers(servers);
+}
+
+function _readJsonFile(file: string): Record<string, unknown> {
+  try {
+    const raw = JSON.parse(fs.readFileSync(file, "utf-8"));
```
egonSchiele: validate the structure of the file using a Zod schema.

packages/agency-lang/lib/stdlib/mcp.ts
```diff
@@ -142,3 +144,147 @@ export async function _loadMcpTools(
...
+function _writeJsonFile(file: string, data: Record<string, unknown>): void {
+  fs.mkdirSync(path.dirname(file), { recursive: true });
+  fs.writeFileSync(file, JSON.stringify(data, null, 2) + "\n");
+}
+
+/** The mcpServers map from a config file (any file with an `mcpServers` key). */
+export function _readMcpServersFromFile(file: string): McpServers {
+  const raw = _readJsonFile(file);
+  const servers = raw.mcpServers;
+  return servers && typeof servers === "object" ? (servers as McpServers) : {};
```
egonSchiele: again, validate using zod schema

egonSchiele:
Reworked in 582e76b — thanks, this is much cleaner (net −232 lines). Went back through `docs/dev/anti-patterns.md` and rebuilt around declarative parsing.

**Architecture / anti-patterns**
- **Don't parse args yourself / imperative everywhere:** deleted both hand-rolled parsers (`mcpCommand.ts` and the REPL `_parseMcpCommand`). The CLI is now real **commander subcommands** with declarative `.option`/`.argument`; commander does parsing, unknown-flag rejection, and help.
- **`agent.ts` hack:** gone — no more `if (args[0] === "mcp")` intercept.
- **Commander sub-command:** `agency mcp add|remove|list` are grouped under the **existing `mcp` command** (which starts Agency as an MCP server), i.e. `agency mcp serve`-vs-`agency mcp add` exactly like Claude Code. (I couldn't nest under `agency agent`: that command injects a `--` separator to pass everything through to the agent binary, so commander can't own subcommands there — and top-level `agency mcp` was already taken by the server. Grouping the management verbs under `mcp` is the clean fit.)
- **Invented error type:** dropped `{ ok, error? }` — the config core and the package's `validateMcpServers` now return Agency `Result` (`success`/`failure`); callers use `isFailure`.
- **Clutter in `agent.agency`:** the REPL add/remove is descoped for now, so `agent.agency` and `std::mcp` are reverted to the merged v1 (`/mcp` stays list-only). `agency mcp add` then loads on next launch. This removes all the imperative MCP code from `agent.agency`.

**Real bugs**
- **Malformed-config data loss:** fixed. Writers use a *strict* read that fails (writing nothing) when the file exists but isn't valid JSON / not an object; only `list` uses a lenient read.
- **Prototype-chain `in` / array-root (Copilot):** fixed — `hasOwnProperty` checks and an `isPlainObject` guard (arrays/non-objects treated as empty).
- **Older-package validate crash (Copilot):** the bridge call is wrapped; a throw becomes a clear "upgrade the package" failure.
- **Re-add duplicates tools / green "connected (0 tools)" / remove scope-shadow:** all were on the REPL hot-connect path, which is descoped, so they no longer apply. (When REPL management returns, I'll drop-before-append and branch the message.)
- **Unhandled rejection in `agent.ts`:** the intercept is gone; commander async actions set `process.exitCode`.

Tests: CLI actions + config core (incl. the malformed-file no-overwrite, own-property, and older-package cases), 60 agency-lang unit + 54 mcp-package tests; `agency mcp add/list/remove` verified end-to-end with an invalid config correctly rejected.

PR #511, opened 2026-07-11
Enforce raises on function types (#434)

### LINKED ISSUE #434: raises clause on function types is parsed but not enforced by the type checker
## Summary

A `raises` clause on a **function type** (`blockType`) is parsed and stored but **never enforced** by the type checker. It is effectively documentation only.

`raises` on a `def`/`node` declaration *is* enforced (inferred ⊆ declared, via `checkRaisesDeclarations`), but the same clause on a function type is silently ignored.

## Repro

```ts
type Callback = (string) -> string raises <>          // declares: raises nothing

def dangerous(s: string): string raises <std::read> { // actually raises std::read
  return read(s)
}

node main() {
  let cb: Callback = dangerous   // assigning a std::read-raising fn to a `raises <>` type
  let x = cb("foo.txt")
  print(x)
}
```

**Expected:** a type error — `dangerous` raises `std::read`, which exceeds `Callback`'s declared `raises <>`.

**Actual:** compiles clean (only an unrelated `Result`-vs-`string` return-type warning). No effect-related diagnostic.

## Root cause

- The clause is parsed and stored: `blockTypeParser` accepts a trailing `raises` clause and stores it on the `blockType` AST node (`lib/parsers/parsers.ts:1460`).
- Nothing in the checker reads it. The only enforcement site, `checkRaisesDeclarations` (`lib/typeChecker/raisesDiagnostic.ts`), iterates only over `ctx.functionDefs` and `ctx.nodeDefs` — it reads `def.raises`, never the `raises` field on a `blockType`.
- Function-type assignability ignores effects entirely: the `blockType → blockType` rule (`lib/typeChecker/assignability.ts:574-595`) checks only parameter contravariance and return-type covariance, never the two sides' `raises` sets. A grep across the whole typechecker confirms no read of `.raises`/blockType effects outside `raisesDiagnostic.ts`.

## Suggested fix direction

- In the `blockType → blockType` branch of `assignability.ts`, require the source's effect set to be a subset of the target's declared set (with `<*>` = top, `<>` = bottom), reusing `resolveEffectSet`.
- Attribute effects at call sites that go through a function-typed value, so calling `cb(...)` contributes `cb`'s declared effects to the enclosing function's inferred set.

## Docs

`docs/site/guide/effects-and-raises.md` presents `raises` on function types as a real feature (final section, "`raises` on Function types") with no caveat, which is currently misleading.

egonSchiele:
Fixes #434.

## What

A `raises` clause on a function type was parsed, then ignored. This makes it real, and fixes a second gap found while scoping: `agency fmt` silently deleted the clause.

```ts
type Callback = (string) -> string raises <>          // allows no effects

def dangerous(s: string): string raises <std::read> {
  return read(s)
}

node main() {
  let cb: Callback = dangerous   // now an error: dangerous raises std::read, Callback allows nothing
}
```

Before: this compiled clean, and `agency fmt` turned `type Callback = (string) -> string raises <>` into `type Callback = (string) -> string`.

## How

- **Enforcement** (`checkFunctionTypeRaises`, a new pass after `checkRaisesDeclarations`): when a function value flows into a function type that declares a `raises` clause, the value may not raise more than the clause allows. Checked at declarations/re-assignments, call arguments, and returns. Assignability stays purely structural; this pass owns the effect comparison so the message is precise. Each site reuses the checker's own machinery — arg→param pairing via `paramListSignature.resolveSlot`, the block-binds-to-the-last-param rule.
- **Source effects** (`functionValueEffects`): a named ref / `.partial()` / `.preapprove()` uses the inferred effect map (recovering the base name syntactically, since those synth to `any`); a block uses its body; an opaque function-typed variable uses its type's clause. A synthesized `any` claims no effects (fail-open), not "raises anything". A function type with no clause means "raises anything" (the strict rule).
- **Callback-call attribution**: calling a function-typed variable now contributes its declared effects to the caller, so an outer `raises` clause catches them (`def f(cb: … raises <std::read>) raises <> { cb("x") }` errors).
- **Formatter**: `variableTypeToString` and `formatTypeHint` now render the clause, reusing the existing effect-set renderer (`AgencyGenerator.effectSetTypeToString` delegates to the new `effectSetToSource`). Gated to Agency output, so TS codegen never emits `raises`.

## Behavior notes

- Fail-closed on opaque sources: a function-typed value whose type has no `raises` clause cannot flow into a constrained type (its type must state what it raises).
- One v1 limitation, documented in the guide: a callback declared `raises <*>` that you *call* inside a constrained function isn't flagged (`<*>` has no concrete effects to attribute). Assigning/passing such a value is still checked.

## Testing

- New: `functionTypeRaises.test.ts` (13 — repro, each site, PFA/preapprove, block, opaque alias-typed source, `<*>` target, effect-set alias target, each-effect, false-positive sweep), `callbackCallEffects.test.ts` (5 — node/alias/positive/double-count/`<*>`), `functionValueEffects.test.ts` (5), formatter round-trip + codegen-leak guard in `formatType.test.ts` / `raisesFormat.test.ts`.
- Full `lib/typeChecker` (1467), `lib/backends`, `formatType`, and `tests/typescriptGenerator` suites pass (1874 total across those). tsc clean, `lint:structure` clean.
- End-to-end: `agency typecheck` on the repro reports the error; `agency fmt` preserves the clause.

---

## Update after review

Rewrote the enforcement pass to be declarative and fixed several correctness gaps (commit `c86033321`):

- **Readability:** the pass now reads top-down — `valueFlows` enumerates *where* a function value meets a function type; `targetAllowed` + `exceedances` decide *how much* is too much; `functionValueEffects` dispatches on the *kind* of value (named ref / block / partial application / opaque type), one small function each.
- **Block-local returns** are no longer checked against the enclosing function (they belong to the block).
- **Method calls** (`x.foo(...)`) no longer resolve against a same-named global `def`.
- **Imported functions** with a constrained callback parameter are now enforced (not just local `def`/`node`).
- **Splat arguments** stop positional pairing (unknown width).
- **Partial application on a callback variable** now recovers effects through its type (was a false negative).

**New behavior worth noting:** because callback-call effects flow through `collectFromBody`, an inline **handler** that calls a raising callback now (correctly) trips the handler-recursion diagnostic — a callback call inside a handler really would re-enter the chain. Pinned with a fixture (fires; and its negative stays clean).

Guide edit to `effects-and-raises.md` reverted per review.


packages/agency-lang/lib/typeChecker/functionTypeRaises.ts
```diff
@@ -0,0 +1,124 @@
...
+      for (const { node } of walkNodes(info.body)) {
+        if (node.type === "assignment") {
+          // Object-field targets (`x.y = f`) are out of scope.
+          if (node.accessChain && node.accessChain.length > 0) continue;
+          if (node.value.type === "messageThread") continue;
+          // Annotated declaration uses the annotation; a bare re-assignment
+          // (`cb = f`) uses the variable's declared type.
+          const target = node.typeHint ?? info.scope.lookup(node.variableName);
+          check(node.value, target, info, node.loc);
+        } else if (node.type === "returnStatement") {
```
egonSchiele: **Confirmed false positive:** `walkNodes` descends into trailing-block bodies, so a block-local `return` is checked against the *enclosing function's* return type. Reproduced on this branch:

```
type Callback = (string) -> string raises <>
def reads(s: string): string raises <std::read> { raise std::read("m", { dir: ".", filename: s })
  return s }
def pure(s: string): string { return s }
def runBlock(cb: (string) -> string raises <*>): string { return cb("x") }

def pick(): Callback {
  let r = runBlock() as s { return reads }   // block returns `reads` into a raises <*> slot — legal
  return pure
}
```

This errors with `'reads' raises effect 'std::read', which exceeds the 'raises <>' allowed by type 'Callback'` even though `pick` never returns `reads` — the block does, into a `<*>` slot.

`checkReturnTypesInScope` handles exactly this with `if (isInsideBlock(ancestors)) continue;` and the comment "Returns inside a block belong to the block, not the enclosing function" (checker.ts:769). Same guard needed here — take `ancestors` from the walk result. (Checking block-local returns against the enclosing block *slot's* return-type clause via `findEnclosingBlockSlot` would be the complete version, but the guard alone fixes the false positive and matches the existing pass's scope.)

packages/agency-lang/lib/typeChecker/functionTypeRaises.ts
```diff
@@ -0,0 +1,124 @@
...
+          if (node.value) check(node.value, info.returnType, info, node.loc);
+        } else if (node.type === "functionCall") {
+          const def = ctx.functionDefs[node.functionName] ?? ctx.nodeDefs[node.functionName];
```
egonSchiele: Two entry conditions from the real call checker (`checkSingleFunctionCall` / `checkFunctionCallsInScope`) are missing here:

1. **Method-call guard.** `walkNodes` descends into `valueAccess.chain[].functionCall`, so `x.foo(reads)` yields a `functionCall` node named `foo`. If a same-named global `def foo(cb: <constrained fn type>)` exists, this pass resolves the *method* call against the *global's* params — a false positive. The checker guards with: `const parent = ancestors[ancestors.length - 1]; if (parent?.type === "valueAccess") continue;` (checker.ts:243-244).

2. **Imported functions.** Resolution here is `functionDefs ?? nodeDefs`, but the checker also falls back to `ctx.importedFunctions` (checker.ts:270-272). As written, passing a raising function to a *cross-module* function with a constrained param is silently unchecked. Fine as a v1 scope cut, but it should be a deliberate one — either add the fallback (the params shape is the same) or note it in the guide alongside the `<*>` limitation and pin it with a test.

packages/agency-lang/lib/typeChecker/functionTypeRaises.ts
```diff
@@ -0,0 +1,124 @@
...
+          // `functionDefs` is a plain object, so a callee named "constructor" /
+          // "toString" resolves to an inherited Object.prototype method — truthy
+          // but with no `parameters`. Require a real parameter list.
+          if (!def || !Array.isArray(def.parameters)) continue;
+          const sig = paramListSignature(def.parameters, node.arguments.length);
+          node.arguments.forEach((arg, i) => {
+            if (arg.type === "splat") return;
```
egonSchiele: This `return` inside `forEach` is a per-argument *continue*, so pairing continues after a splat using the raw argument index. `checkArgsAgainstParams` deliberately refuses to do that — it stops all positional pairing at the first splat (checker.ts:659-671) because the splat's width is statically unknown. `f(...xs, reads)` here checks `reads` against `params[1]`, which may be the wrong slot → potential false positive against whatever clause that param carries.

Suggest mirroring the checker: convert to a `for` loop and `break` at the first splat (args before it are still safely paired).

egonSchiele (commented):
Solid implementation — the design lands where the plan review pointed (shared `effectSetToSource`, `paramListSignature` reuse, `safeResolveType` on source/callee types, fail-open on synthesized `any`, base-name recovery for PFA chains), and the test suite covers the sites well. I verified the trailing-block last-param rule against `checkBlockArgShape` (it matches, same rationale) and confirmed there is no duplicate-diagnostic problem from the per-scope walk.

I did find **two confirmed bugs** (both reproduced by running the checker on this branch) and a few divergences from the real call checker's entry conditions. Details inline; the short version:

1. **False positive (confirmed):** a `return` inside a trailing block is checked against the *enclosing function's* return type. `checkReturnTypesInScope` guards exactly this with `isInsideBlock(ancestors)` (checker.ts:769); the new pass doesn't.
2. **False negative (confirmed):** `.preapprove()`/`.partial()` on a function-typed *variable* claims no effects — `pfaBaseName` recovers the variable name, which has no entry in the effect map.
3. The `functionCall` branch skips two entry conditions the real checker applies: the method-call guard (`valueAccess` parent) and the `ctx.importedFunctions` fallback — one can misfire, the other silently skips cross-module calls.
4. Splat pairing diverges from `checkArgsAgainstParams`, which refuses to pair positional args after a splat; this pass keeps pairing by raw index.
5. The handler-body behavior change (a handler calling a raising callback now trips the handler-recursion diagnostic) is real — I verified it fires — and it's the *right* direction, but it's untested and unmentioned in the PR body/guide. Handlers are safety infrastructure; please pin it with a test.

Smaller notes inline. One missing test worth adding while in here: `<*>` as the *source* (`cb: AnyFn` assigned to `Callback`) should error — the suite covers `<*>` only as target and as callback-attribution.

packages/agency-lang/lib/typeChecker/functionValueEffects.ts
```diff
@@ -0,0 +1,78 @@
...
+  if (expr.type === "blockArgument") {
+    return {
+      any: false,
+      labels: collectRaisableEffects(expr.body, info, interruptEffectsByFunction, ctx),
+    };
+  }
+
+  // A PFA / preapprove / describe / rename chain synths to "any"; recover the
+  // base name and use the base function's effect set.
+  const base = pfaBaseName(expr);
```
egonSchiele: **Confirmed false negative:** `pfaBaseName` also fires when the base is a function-typed *variable*, and the variable name has no entry in `interruptEffectsByFunction`, so the value claims no effects. Reproduced on this branch:

```
def hof(cb: (string) -> string raises <std::read>) {
  let danger: Callback = cb.preapprove()   // no error — should exceed Callback's raises <>
}
```

`byName("cb", …)` returns `{ any: false, labels: [] }`. Note the map also omits defs that raise nothing, so "absent from the map" can't distinguish "pure def" from "not a def". Suggested fix: before the map lookup, try `info.scope.lookup(base)` — if it resolves (through `safeResolveType`) to a `blockType`, use its `raises` clause (strict `any: true` when it has none), same as the opaque-variable branch below; only fall through to `byName` when the base isn't a scope variable.

packages/agency-lang/lib/typeChecker/interruptAnalysis.ts
```diff
@@ -128,6 +148,12 @@ function collectFromBody(
       addUnique(kinds, node.effect);
     } else if (node.type === "functionCall") {
       addUnique(callees, node.functionName);
+      // A call THROUGH a function-typed variable (a callback) contributes the
+      // variable's declared effects. Named-def callees resolve via the callee
+      // lookup above and are skipped here (they aren't blockTypes).
+      for (const label of calleeDeclaredEffects(node.functionName, scope, ctx)) {
```
egonSchiele, line 154: Verified behavior change worth pinning: `collectFromBody` also feeds `collectRaisableEffects` → `checkHandlerBodyInterrupts`, so an inline handler that calls a raising callback now trips the handler-recursion diagnostic. I confirmed on this branch:

```
def useIt(cb: (string) -> string raises <std::read>) {
  handle { print("hi") } with (payload) { print(cb("x")) }
}
```

→ `Handler (inline) may raise interrupts [std::read]…`

This is the *correct* direction (a handler calling a raising callback really would re-enter the chain), but it's a new error path through safety-critical machinery with no test and no mention in the PR body or guide. Please add a fixture asserting it fires (and its negative: a handler calling a `raises <>` callback stays clean), and a line in the PR description.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -952,10 +952,7 @@ export class AgencyGenerator {
   // an inline flagged union renders as `<...>` via variableTypeToString.
   // Single source of truth for both effectSet declarations and raises clauses.
```
egonSchiele: Nit: this comment block still claims to be the "single source of truth for both effectSet declarations and raises clauses," but the method is now a one-line delegate — the source of truth moved to `effectSetToSource` (whose docstring correctly claims the title). Trim this to something like "Delegates to effectSetToSource; kept for subclass call sites."

packages/agency-lang/docs/site/guide/effects-and-raises.md
```diff
@@ -134,4 +134,41 @@ Function types can carry a `raises` clause too:
 
 ```ts
 type Callback = (string) -> string raises <std::read>
-```
\ No newline at end of file
+```
+
+The clause is enforced. A function value may not flow into a function type that
```
egonSchiele: Please undo this entire edit to this file.

packages/agency-lang/lib/typeChecker/functionTypeRaises.ts
```diff
@@ -0,0 +1,124 @@
...
+        ctx.errors.push({
+          message: `${who} raises effect '${effect}', which exceeds the ${allow}. Add '${effect}' to the clause, or use a target type that allows it.`,
+          severity: "error",
+          loc,
+        });
+      }
+    }
+  };
+
+  for (const info of scopes) {
```
egonSchiele, line 169: This long nested piece of logic is an unreadable mess. Why not extract each of the branches of these statements into their own function that you call to make this easier to read?

egonSchiele: Rewrote the pass so it reads top-down (c86033321). The core loop is now: for each `valueFlows(node)`, if `targetAllowed(flow.target)`, report `exceedances(functionValueEffects(flow.source), allowed)`. Each site kind (assignment / return / call-args / block) is its own small function under a "WHERE" heading; `targetAllowed` and `exceedances` are the "HOW MUCH". The kind-of-value dispatch (named ref / block / partial application / opaque type) is one small function each in functionValueEffects.ts.

packages/agency-lang/lib/typeChecker/functionTypeRaises.ts
egonSchiele: Fixed (c86033321). `returnFlows` now `isInsideBlock(ancestors)`-guards, exactly as `checkReturnTypesInScope` does. Added your repro as a fixture asserting no error.

packages/agency-lang/lib/typeChecker/functionTypeRaises.ts
egonSchiele: Both fixed (c86033321). `callFlows` skips when the parent is a `valueAccess` (method call), and `calleeParams` now falls back to `ctx.importedFunctions[name]?.parameters`. Method-call case has a fixture; imported functions are resolved via the same params shape (single-file `typecheckSource` can't exercise a real cross-module import, so no fixture there).

packages/agency-lang/lib/typeChecker/functionTypeRaises.ts
egonSchiele: Fixed (c86033321). `argumentFlows` is now a `for` loop that `break`s at the first splat, so args after a splat are not mis-paired. Fixture added.

packages/agency-lang/lib/typeChecker/functionValueEffects.ts
egonSchiele: Fixed (c86033321). `functionValueEffects` now recurses on the PFA base expression rather than doing a bare name lookup, so `cb.preapprove()` resolves through `cb`'s type (strict `any` when it has no clause) exactly like the opaque-variable branch. Added your `let danger: Callback = cb.preapprove()` repro as a fixture.

packages/agency-lang/lib/typeChecker/interruptAnalysis.ts
egonSchiele: Added fixtures (c86033321): a handler whose body calls a `raises <std::read>` callback now trips the handler diagnostic (asserted), and the negative (a `raises <>` callback stays clean). Called out in the PR description.

packages/agency-lang/lib/backends/agencyGenerator.ts
egonSchiele: Trimmed to "Delegates to effectSetToSource (the single source of truth); kept for subclass call sites." (c86033321)

packages/agency-lang/docs/site/guide/effects-and-raises.md
egonSchiele: Reverted the entire edit to this file (c86033321).

packages/agency-lang/lib/typeChecker/functionTypeRaises.ts
```diff
@@ -0,0 +1,124 @@
...
+              const t = sig.resolveSlot({ kind: "positional", index: i })?.type;
+              check(arg, t, info, node.loc);
+            }
+          });
+          // A trailing block binds to the LAST parameter (the backend pushes it
+          // as the final positional argument).
+          if (node.block) {
+            const last = def.parameters[def.parameters.length - 1];
+            check(node.block, last?.typeHint, info, node.loc);
+          }
```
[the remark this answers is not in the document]
egonSchiele: Fixed (c86033321). `calleeParams` now resolves `ctx.functionDefs`/`ctx.nodeDefs` **and** `ctx.importedFunctions[name]?.parameters`, so a constrained callback parameter on an imported function is enforced.

PR #512, opened 2026-07-11
std::agency: writeAgency, review, getEffects, maxCost + std::skills docsSkill

egonSchiele:
Adds the primitives for agents that write their plans as Agency code and run them under handlers (the CodeAct pattern), plus the groundwork for offline harness search.

Spec: `docs/superpowers/specs/2026-07-10-std-agency-write-effects-cost-design.md` (in the main checkout, untracked).

## New std::agency APIs

- **`writeAgency(task, context, maxAttempts)`** — generate Agency source from a task description. One `llm()` call with the packaged language guide available as a docs tool, then a typecheck-repair loop that feeds diagnostics back. Success is source guaranteed to compile and typecheck; exhaustion returns `{ source, errors }`.
- **`review(source, task)`** — lifted from the review subagent. Always merges parse + typecheck findings; a non-empty `task` adds an LLM pass judging whether the code does what was asked. `Feedback`, `mergeFeedback`, `renderFeedback`, `feedbackHasErrors` are exported; the subagent now delegates per-snippet to the stdlib.
- **`getEffects(source)`** — maps each exported node/function to its transitive interrupt-effect list, read from the same propagated map raises-checking enforces. Bare `interrupt(...)` surfaces as the fail-closed `"unknown"` sentinel. Approval flows can show a plan-program effect envelope before running it.
- **`maxCost` on `run`/`runFile`** — cost cap for subprocesses, implemented as a `guard(cost:)` under the hood. Trips translate to the same `limit_exceeded` failure shape as the other limits, with `limit: "cost"`. The cap rides in the `std::run` interrupt payload (`limits.maxCost`, additive declaration change) so handlers and policies see it before approving.

## Docs staging moved into the stdlib

`make` now stages the guide + CLI docs into `stdlib/docs/` (gitignored, shipped via `files`), served by the new `std::skills::docsSkill("guide" | "cli")`. All four subagents drop their per-file `skillsDir("../docs/...") with approve` statics for `docsSkill`, and the old `stage-agent-docs` staging into `dist/lib/agents/docs/` is retired. `docsSkill` skips only the scan-approval interrupt (packaged docs carry stdlib trust); reads through the returned tool still raise `std::read` and go through handlers/policies unchanged.

## Notes for review

- `guard()` passes a returned `Result` through unchanged (try-style call), so `run`'s guarded branch returns the child result as-is and only translates a `guardFailure`. The `nested` test pins inner-trip-does-not-trip-outer-guard.
- `getEffectsFromSource` shares one extracted `runCheckerPipeline` helper with `typeCheckSource`, so the `allowTestImports: false` decision exists once. No typechecker changes; the transitive map comes from `typeCheck`'s return (same pattern as `lib/cli/policy.ts`).
- The `trips` test handler rejects unless `e.data.limits.maxCost` matches, pinning the payload change.
- Real-LLM smoke test of `writeAgency` surfaced that models default to C-style `for(;;)` loops; the system prompt now shows the while-loop counting idiom explicitly and generation succeeds.

## Tests

- `tests/agency/docs-skill.agency` — guide + cli pages served, no startup interrupt (tool builds sit outside the handler on purpose).
- `tests/agency/get-effects.agency` + 6 vitest cases — direct/transitive effects, sort/dedup, exported defs, clean-node `[]`, unknown sentinel, parse-throw.
- `tests/agency/subprocess/run-max-cost.agency` — trip, generous, nested guards, non-cost passthrough, runFile forwarding (5 nodes), plus the pre-existing `cost-guard-trips-on-child-spend` still passing for the `maxCost == null` path.
- `tests/agency/agency-review.agency` — lint-only, parse-error path, LLM feedback actually merged (error-finding mock).
- `tests/agency/write-agency.agency` — repair retry, parse-failure recovery, exhaustion payload.

### DIFF
--- a/packages/agency-lang/package.json
+++ b/packages/agency-lang/package.json
@@ -43,8 +43,6 @@
   },
   "files": [
     "./dist",
-    "./dist/lib/agents/docs/guide/**/*.md",
-    "./dist/lib/agents/docs/cli/**/*.md",
     "./dist/lib/agents/docs/appendix/**/*.md",
     "./stdlib/**/*.agency",
     "./stdlib/**/*.js",
--- a/packages/agency-lang/package.json
+++ b/packages/agency-lang/package.json
@@ -43,7 +43,6 @@
   },
   "files": [
     "./dist",
-    "./dist/lib/agents/docs/appendix/**/*.md",
     "./stdlib/**/*.agency",
     "./stdlib/**/*.js",
     "./stdlib/docs/**/*.md",
--- a/packages/agency-lang/package.json
+++ b/packages/agency-lang/package.json
@@ -48,6 +48,7 @@
     "./dist/lib/agents/docs/appendix/**/*.md",
     "./stdlib/**/*.agency",
     "./stdlib/**/*.js",
+    "./stdlib/docs/**/*.md",
     "./scripts/hooks/postinstall.js"
   ],
   "exports": {
--- a/packages/agency-lang/stdlib/agency.agency
+++ b/packages/agency-lang/stdlib/agency.agency
@@ -3,6 +3,7 @@ import {
   _compileFile,
   _typecheck,
   _typecheckFile,
+  _getEffects,
   _parseAST,
   _writeAST,
   _format,
@@ -213,6 +214,22 @@ export safe def typecheck(source: string): Result<TypeCheckReport> {
   @param source - Agency source code as a string
   """
   return try _typecheck(source)
+}
+
+/** Per-exported-symbol effect lists, keyed by node/function name. */
+export type EffectsByExport = Record<string, string[]>
+
+export safe def getEffects(source: string): Result<EffectsByExport> {
+  """
+  Map each exported node and function in the source to the list of
+  interrupt effects it can raise, transitively. Bare `interrupt(...)`
+  sites appear as the sentinel "unknown", so the envelope never
+  silently under-reports. Use this to show or check what a program can
+  do before running it.
+
+  @param source - Agency source code as a string
+  """
+  return try _getEffects(source)
 }
 
 type FilterImportsResult = {
--- a/packages/agency-lang/stdlib/agency.agency
+++ b/packages/agency-lang/stdlib/agency.agency
@@ -153,23 +153,13 @@ export def run(
     depth: _subprocessDepth() + 1
   })
 
-  // NOTE: the two _run call sites below must keep IDENTICAL argument
-  // lists (Agency has no zero-arg lambda to factor the call through).
-  if (maxCost == null) {
-    return try _run(
-      compiled,
-      node,
-      args,
-      wallClock,
-      memory,
-      ipcPayload,
-      stdout,
-      configOverrides,
-      cwd,
-      maxDepth,
-    )
+  // A negative cost disables the guard, so the no-cap case runs through
+  // the same single call site with an inert guard.
+  let costCap = -1.0
+  if (maxCost != null) {
+    costCap = maxCost
   }
-  const guarded = guard(cost: maxCost) as {
+  const guarded = guard(cost: costCap) as {
     return try _run(
       compiled,
       node,
@@ -483,7 +473,14 @@ export def writeAgency(
       const checked = try _typecheck(source)
       if (checked is failure(parseErr)) {
         result = failure({ source: source, errors: [] })
-        const findings = renderFeedback(parseFeedback(source))
+        // _typecheck throws for import-resolution failures as well as
+        // parse failures. In the import case the re-parse succeeds and
+        // renders no findings — fall back to the thrown error so the
+        // model always gets a diagnostic to repair against.
+        let findings = renderFeedback(parseFeedback(source))
+        if (findings == "") {
+          findings = "${parseErr}"
+        }
         prompt = "That code has problems:\n${findings}\nReturn a corrected complete program."
       } else if (checked.value.errors.length > 0) {
         result = failure({ source: source, errors: checked.value.errors })
--- a/packages/agency-lang/stdlib/agency.agency
+++ b/packages/agency-lang/stdlib/agency.agency
@@ -14,7 +14,8 @@ import {
   _subprocessDepth,
  } from "agency-lang/stdlib-lib/agency.js"
 import { VERSION } from "agency-lang/stdlib-lib/version.js"
-import { guard } from "std::thread"
+import { guard, systemMessage } from "std::thread"
+import { docsSkill } from "std::skills"
 
 /** @module
   Tools for compiling, type-checking, running, formatting, and inspecting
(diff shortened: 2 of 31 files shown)


packages/agency-lang/stdlib/agency.agency
```diff
@@ -215,6 +264,246 @@ export safe def typecheck(source: string): Result<TypeCheckReport> {
...
+    let attempt = 0
+    let done = false
+    while (attempt < maxAttempts && !done) {
+      attempt = attempt + 1
+      const generated: GeneratedCode = llm(prompt, { tools: [docsTool] })
+      const source = generated.code
+      const checked = try _typecheck(source)
+      if (checked is failure(parseErr)) {
+        result = failure({ source: source, errors: [] })
+        const findings = renderFeedback(parseFeedback(source))
```
egonSchiele: `_typecheck` throws for **import-resolution failures as well as parse failures** (see the `typeCheckSource` doc comment: "parse failure, import resolution failure — propagate to the caller"). In that case this re-parse succeeds, `parseFeedback` returns `success([])`, `renderFeedback` renders `""`, and the model gets `"That code has problems:\n\nReturn a corrected complete program."` — no diagnostic at all — while `parseErr`, which holds the real message, is discarded. A generated program with a bad import is a plausible model failure mode, and with an empty reprompt the remaining attempts will likely repeat the same mistake.

Suggestion:

```agency
let findings = renderFeedback(parseFeedback(source))
if (findings == "") {
  findings = "${parseErr}"
}
prompt = "That code has problems:\n${findings}\nReturn a corrected complete program."
```

(Or skip the re-parse entirely and always use `${parseErr}` in this branch — `_typecheck` already ran the parser, so the re-parse only buys nicer formatting in the pure-parse-error case.)

packages/agency-lang/package.json
```diff
@@ -43,11 +43,10 @@
   },
   "files": [
     "./dist",
-    "./dist/lib/agents/docs/guide/**/*.md",
-    "./dist/lib/agents/docs/cli/**/*.md",
     "./dist/lib/agents/docs/appendix/**/*.md",
```
egonSchiele: This `appendix` entry is dead: nothing stages it (the deleted `stage-agent-docs` macro only ever copied `guide` and `cli`), `dist/lib/agents/docs/` contains only `cli` and `guide` after `make`, and there is no `docs/site/appendix` source directory. The plan kept it on the belief it was "staged elsewhere", but I could not find any staging path. Now that this PR retires the last docs staging into `dist/lib/agents/docs/`, consider deleting the entry — or if the appendix corpus is planned, a comment saying so. (Harmless as-is; npm ignores globs that match nothing.)

egonSchiele (commented):
Reviewed the full diff against the plan, the plan-review findings, and the runtime code. This is in good shape — every finding from the plan review made it in (all four subagents migrate both docs statics, the effect declaration gained `limits.maxCost`, `getEffectsFromSource` reads `typeCheck`'s return through the extracted `runCheckerPipeline`, the vacuous `withTask` test became an error-finding mock, `runFile` forwarding and the non-cost passthrough are tested). Two comments inline: one actionable gap in `writeAgency`'s reprompt path, one dead `files` entry.

Things I verified rather than assumed:

- **The `guard()` passthrough comment is correct.** `_runGuarded` invokes the block through `__tryCall` — the same machinery as `try` — so a returned `Result` passes through unchanged and `guarded` really is the child's Result on success, no double wrapping (`lib/stdlib/thread.ts:354-374`). `return guarded` (instead of the plan's `guarded.value`) is right, and the `passthrough` test pins it at runtime.
- **`result.value.data` in `generous` matches the established `run()` envelope** — same accessor as `handler-approve.agency`, `nested-depth-boundary.agency`, and the other subprocess tests.
- **`Result<string, WriteFailure>` two-param form has precedent** (`stdlib/llm.agency:96`, `stdlib/policy.agency:395`).
- **Root `.gitignore` + `files` allowlist interplay is right**: npm-packlist never consults the repo-root gitignore, and the `./stdlib/docs/**/*.md` entry ships the staged docs the same way the gitignored `stdlib/**/*.js` already ships.
- **`runFileTrips` follows the existing `run-file.agency` convention** (repo-relative `dir`, target subdir with `sub.agency`), and compiled `.js` siblings are correctly untracked.
- The `trips` handler rejecting unless `e.data.limits.maxCost` matches is a nice touch — the payload change is now unfakeable.

Non-blocking observation, no action needed: the `walkAST` signature in the regenerated `docs/site/stdlib/agency.md` changed `=>` to `->` — pre-existing doc-generator drift being committed along with the regen, not something this PR introduced.

packages/agency-lang/stdlib/agency.agency
```diff
@@ -139,25 +145,65 @@ export def run(
...
-    args,
-    wallClock,
-    memory,
-    ipcPayload,
-    stdout,
-    configOverrides,
-    cwd,
-    maxDepth,
-  )
+  // NOTE: the two _run call sites below must keep IDENTICAL argument
```
egonSchiele: It's annoying to have two call sites here. I think we should instead modify the guard so that if the user passes in a negative cost or a negative time, that guard is disabled. Then you can have simply one call to _run, with a guard around it, and if the user didn't want a max cost, we can set it as negative.

packages/agency-lang/stdlib/agency.agency
```diff
@@ -199,6 +247,7 @@ export def runFile(
     memory: memory,
     ipcPayload: ipcPayload,
     stdout: stdout,
+    maxCost: maxCost,
```
egonSchiele, line 255: We are adding a cost limit for these, do we also want to add a time limit or does that exist already?

packages/agency-lang/stdlib/agency.agency
egonSchiele: Fixed in a5ff5a36e with your suggested shape: the parse branch now falls back to the thrown error when `renderFeedback(parseFeedback(source))` renders empty, so import-resolution failures reach the model instead of an empty reprompt. Kept the re-parse for the nicer formatting in the pure-parse-error case.

packages/agency-lang/package.json
egonSchiele: Verified and deleted in a5ff5a36e — nothing stages an appendix dir, `dist/lib/agents/docs/` only ever gets guide and cli, and there is no docs/site/appendix source. The plan kept it on a wrong belief.

packages/agency-lang/stdlib/agency.agency
egonSchiele: Done in 318a793e1. `pushGuardImpl` now skips pushing a guard for a null OR negative limit (documented in the guard docstring), and `run()` collapses to a single guarded `_run` call site passing `-1.0` when `maxCost` is null. The existing `cost-guard-trips-on-child-spend` fixture covers the disabled-inner-guard path: `run()` without `maxCost` now goes through the inert guard, and the outer guard still meters child spend.

packages/agency-lang/stdlib/agency.agency
egonSchiele: A time limit already exists on both: the `wallClock` parameter (default 60s, max 1h) SIGKILLs the subprocess and returns the same `limit_exceeded` shape with `limit: "wall_clock"`. An enclosing `guard(time:)` also applies to subprocesses, same as cost guards. So I left time alone here — `maxCost` was the only missing dimension.

packages/agency-lang/stdlib/agency.agency
egonSchiele: Update: the single-call-site version broke `nested-pause-resume` in CI (agency-tests went red after my last push). Root cause is a runtime bug, not the guard-disable design: when a subprocess pauses on a bubbled interrupt and its serialized stack contains a `guard(...) as {}` block wrapping the paused `run()`, the resume loses the block value and `run()` returns `undefined`. Single-level pause/resume through the block is fine; it breaks only across checkpoint serialize/deserialize. Filed with a minimal repro as #513.

I kept the negative-disables guard semantics in `pushGuardImpl` (they stand on their own) but reverted `run()` to the conditional wrap, with a comment pointing at the fixture and #513 — the no-cap path must stay guard-free until #513 is fixed. Note the same bug means `maxCost` on a nested run that pauses is currently broken too (the guarded branch); once #513 is fixed, run() can collapse to the single guarded call site you asked for and maxCost becomes nested-pause-safe for free.

PR #514, opened 2026-07-11
Unwrap tool-call Results for the LLM and add std::agency runCode

egonSchiele:
Fixes the agents-write-and-run-code flow (`writeAgency` + `runCode` as `llm()` tools), which failed end-to-end for four distinct reasons. Each was diagnosed from a real-LLM transcript via statelog and fixed with tests.

## 1. Tool results now unwrap success Results

Tool calls returned the raw `{__type: "resultType", success: true, value}` envelope to the model. The model had to fish values out of `.value` and, worse, reconstruct wrapped objects when echoing them into later tool arguments — it passed `compile()`s CompiledProgram back to `run()` without the `code` field, failing every compile-then-run flow. The success path now sends `.value` (LLM-facing only; the Result cached for Agency code is untouched). Failures already unwrapped.

## 2. `runCode`: compile + run in one call, built for tool use

New `std::agency::runCode(source, node = "main", ...)` compiles and runs source directly, so nothing large round-trips through the model. Design points, each earned by a real-LLM failure observed during verification:

- **Human-unit limit params with ceilings.** The model copied docstring defaults like "60s" / "512mb" as `60` / `512` — which the runtime reads as 60 ms and 512 bytes; the 512-byte heap became `--max-old-space-size=1` and the child died at V8 startup. `runCode` takes `wallClockSeconds` / `memoryMb` / `ipcPayloadMb` / `stdoutMb` (converted internally, ceilings mirroring LIMIT_CEILINGS, over-ceiling rejected with a message naming the max). `run`/`runFile` keep canonical units — unit literals like `60s` already produce those, and reinterpreting them would silently break existing call sites; their docstrings now state the units explicitly.
- **Declared `safe`, with pre-execution failures born retryable.** A non-retryable tool failure removes the tool from the conversation, so one bad generation ended the whole repair loop. `safe` (the documented okay-to-retry marker) stops codegen from force-clearing the retryable bit, and bad-limit/compile failures are constructed retryable via a small `_retryableFailure` extern — the model fixes its source and calls again. Each re-run still raises `std::run` and the child effects re-prompt, so nothing bypasses handlers.
- **Returns the node's value, not the run envelope.** The envelope leads with zero-filled token stats; the model read that as "no output" and thrashed. `run()` keeps the envelope for Agency callers.

## 3. Structured tool errors stringify

The failure branch did `String(error)`, so structured errors (e.g. `writeAgency`'s `{source, errors}`) reached the model as `Error: [object Object]`. Now JSON-stringified.

## Verification

- Unit tests for the new `unwrapToolResultForLlm` / `toolErrorMessage` helpers (`_internal` pattern); full `lib/runtime` vitest suite green.
- `tests/agency/subprocess/run-code.agency`: basic run, retryable compile failure, over-ceiling limit rejection, maxCost forwarding (4 nodes).
- Real-LLM e2e of the original failing flow: the model now generates, runs, reads "did not return a value" when its program printed instead of returning, self-corrects, re-runs, and reports `5`. Transcript verified via statelog.
- Existing subprocess suites (nested-pause-resume, run-max-cost, cost-guard) re-run green; structural lint clean.

### DIFF
--- a/packages/agency-lang/lib/runtime/prompt.ts
+++ b/packages/agency-lang/lib/runtime/prompt.ts
@@ -194,6 +194,14 @@ function unwrapToolResultForLlm(result: any, toolName: string): any {
   );
 }
 
+/** Render a failure Result's error for the model. String errors pass
+ *  through; structured errors (e.g. writeAgency's `{source, errors}`)
+ *  JSON-stringify — `String(error)` would send the useless
+ *  "[object Object]". */
+function toolErrorMessage(error: any): string {
+  return typeof error === "string" ? error : stringifyToolResult(error);
+}
+
 /** Truncate a tool result for the LLM if its serialized form exceeds
  *  `cap` characters. Returns the ORIGINAL value untouched when within
  *  the cap (so smoltalk serializes it exactly as before) or when the cap
@@ -442,6 +450,7 @@ export const _internal = {
   capToolResultForLlm,
   assertUniqueToolNames,
   unwrapToolResultForLlm,
+  toolErrorMessage,
   armCallTimeout,
   runWithRetry,
   dropNullDefaultedArgs,
@@ -1098,10 +1107,7 @@ export async function runPrompt(args: {
       }
 
       if (isFailure(toolResult)) {
-        const errorMessage =
-          typeof toolResult.error === "string"
-            ? toolResult.error
-            : String(toolResult.error);
+        const errorMessage = toolErrorMessage(toolResult.error);
         // Cap only what the LLM sees; statelog keeps the full message.
         const cappedError = String(
           capToolResultForLlm(errorMessage, toolResultCap),
--- a/packages/agency-lang/lib/runtime/prompt.ts
+++ b/packages/agency-lang/lib/runtime/prompt.ts
@@ -1182,8 +1182,10 @@ export async function runPrompt(args: {
       // branch cache keeps the FULL value (Agency code may match on a
       // Result); the model-facing message gets the unwrapped success
       // value via unwrapToolResultForLlm inside the push helper.
+      // Nullish, NOT ||: legitimate falsy returns (false, 0, "") must
+      // reach the model and the branch cache as-is.
       toolResult =
-        toolResult ||
+        toolResult ??
         `${handler.name} ran successfully but did not return a value`;
       stack.setResultOnBranch(branchKey, toolResult);
       pushSuccessToolMessage({ toolResult, toolCall, handler, branchStack });
--- a/packages/agency-lang/lib/runtime/prompt.ts
+++ b/packages/agency-lang/lib/runtime/prompt.ts
@@ -30,7 +30,7 @@ import type { PromptConfig } from "./llmClient.js";
 import { setupFunction } from "./node.js";
 // See docs/dev/promptRunner.md for the control-flow abstraction used here.
 import { PromptBailout, PromptRunner } from "./promptRunner.js";
-import { failure, isFailure } from "./result.js";
+import { failure, isFailure, isSuccess } from "./result.js";
 import type { SourceLocationOpts } from "./state/checkpointStore.js";
 import type { RuntimeContext } from "./state/context.js";
 import type { LlmDefaults } from "../stdlib/llm.js";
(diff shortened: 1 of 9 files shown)


packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -177,6 +178,17 @@ export function _getEffects(source: string): Record<string, string[]> {
   return getEffectsFromSource(source);
 }
 
+/**
```
egonSchiele, line 181: Is this still needed, or is using `safe` enough?

packages/agency-lang/stdlib/agency.agency
```diff
@@ -256,6 +257,102 @@ export def runFile(
...
+
+/** Just like `run`, any interrupts and guards defined in the parent process
+will apply to the child process. Any callbacks in scope will also apply.
+Exceeding a resource limit kills the subprocess and returns a `limit_exceeded` failure.
+
+Designed for LLM tool use: compile() returns a CompiledProgram the model
+would have to echo back into run() verbatim (it will not — see the
+compile→run "CompiledProgram has no code" failure mode). runCode takes
+the source directly, so nothing large round-trips through the model.
+Unlike `run`/`runFile` (which take canonical milliseconds/bytes and are
```
egonSchiele: Sorry to flip-flop, but let's leave this as milliseconds and bytes so that users can use unit literals, and let's make it clear in the doc string, in the text for the parameter itself, that the input should be in milliseconds.

packages/agency-lang/stdlib/agency.agency
```diff
@@ -256,6 +257,102 @@ export def runFile(
...
+  @param node - Which exported node to run (default "main")
+  @param args - Arguments to pass to the node
+  @param wallClockSeconds - Max wall-clock SECONDS before the subprocess is killed (default 60, max 3600). Omit for the default.
+  @param memoryMb - Max V8 heap size in MEGABYTES (default 512, max 4096). Omit for the default.
+  @param ipcPayloadMb - Max single IPC message size in MEGABYTES (default 100, max 1024). Omit for the default.
+  @param stdoutMb - Max combined stdout+stderr output in MEGABYTES (default 1, max 100). Omit for the default.
+  @param maxCost - Max subprocess LLM spend in dollars (e.g. 0.50). null = no cost limit.
+  """
+  let limitError = checkRunCodeLimit("wallClockSeconds", wallClockSeconds, 3600, "seconds")
+  if (limitError == "") {
```
egonSchiele: This code duplication is weird

packages/agency-lang/lib/stdlib/agency.ts
egonSchiele: Still needed — verified empirically rather than assumed: `compile()` is itself a `safe def` returning `try _compile(...)`, and `compile("node {{{").retryable` is `false`. `try`-born failures START retryable:false (`__tryCall` never sets the flag), and `safe` only stops codegen from clearing the bit on return — it never upgrades it. So `safe` + `_retryableFailure` (born true) is the working combination; either alone is insufficient. Added a comment at the call site saying exactly this.

packages/agency-lang/stdlib/agency.agency
egonSchiele: Done in 9e9e75025 — params are back to canonical milliseconds/bytes with unit-literal defaults (`60s`, `512mb`, ...), and each @param states the unit in caps with the numeric default and max spelled out (e.g. "in MILLISECONDS before SIGKILL (default 60000 = 60s, max 3600000 = 1h)"), plus "Pass null for the default". Re-verified with a real-LLM run: the model passed `wallClock: 60000` and `null` for the rest, and the flow returned 5.

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -1155,7 +1178,10 @@ export async function runPrompt(args: {
...
 
       // Success: cache the result, push tool message (extracted helper
-      // keeps this arrow inside the max-lines-per-function budget).
+      // keeps this arrow inside the max-lines-per-function budget). The
+      // branch cache keeps the FULL value (Agency code may match on a
+      // Result); the model-facing message gets the unwrapped success
+      // value via unwrapToolResultForLlm inside the push helper.
       toolResult =
         toolResult ||
         `${handler.name} ran successfully but did not return a value`;
```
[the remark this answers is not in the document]
egonSchiele: Fixed — `??` now, with unit tests pinning that success values of `false`, `0`, and `""` reach the model and the branch cache as-is, and only null/undefined get the no-value message.

packages/agency-lang/stdlib/agency.agency
egonSchiele: Restructured: one `checkRunCodeLimits(wallClock, memory, ipcPayload, stdout)` def that loops over an array of `{name, value, max, unit}` checks (ceilings as unit literals mirroring LIMIT_CEILINGS), returning the first violation message. The if-chain is gone.

egonSchiele:
All four comments addressed in 9e9e75025 (replies inline). One addition beyond the comments, found while re-verifying: a fresh real-LLM run happened to generate worse code (`Int` types), and BOTH tools got removed from the conversation — writeAgency's exhaustion failure and runCode's run()-level failure were still born non-retryable, so one bad generation ended the whole flow with no answer. writeAgency is now `safe` with retryable exhaustion failures, and runCode wraps run()-level failures (missing node, child init crash, timeout, cost trip) retryable as well — every retry still re-raises `std::run`, so nothing bypasses handlers. New test nodes pin both (`exhausts` asserts retryable; `runtimeFailure` covers the run-failure path, with a comment distinguishing it from in-node crashes, which the child converts to a failure Result that run() returns as the node's value).

Final verification: 5/5 run-code nodes, 3/3 write-agency nodes, subprocess regression suites (run-max-cost, nested-pause-resume, cost-guard), full lib/runtime vitest, and a clean real-LLM e2e (model passed wallClock: 60000 and nulls, got 5).

PR #515, opened 2026-07-11
agency run: make --interactive a user endpoint, not a chain handler

egonSchiele:
## Problem

`agency run --interactive` (and the non-interactive default-reject) was implemented as a root handler in the chain. Per the rules of handlers, every handler runs on every interrupt — so the CLI prompted (or rejected) interrupts the program had already settled with its own `handle ... with approve`. Example:

```
node main() {
  handle {
    raise std::error("This is a test error", { a: 1, b: 2 })
    print("This will not be printed if rejected")
  } with approve
}
```

Running this with `--interactive` prompted the user even though the code approves the interrupt itself.

## Semantics change

`--interactive` is now the **user endpoint** — it models what happens when an interrupt surfaces to a TypeScript caller as a returned `Interrupt[]`, rather than acting as another handler:

- **Explicit policy rules** (`--approve X`, `--reject X`, policy-file rules incl. `*` wildcards) still participate as a root chain handler, unchanged: a `--reject` beats a code approve and vice versa, per the rules of handlers.
- For effects the policy does **not** match, the handler now stays silent (previously: prompt/reject), leaving the decision to the program's own handlers.
- Interrupts that **nothing settles** — or that a handler explicitly `propagate`s — surface to a new `resolveCliInterrupts` endpoint in the generated CLI bootstrap. With `--interactive` it prompts and resumes the run via the module-bound `respondToInterrupts` (the same loop `agency serve` and `agency evaluate` use); `aa`/`rr` answers are remembered for the rest of the run. Without `--interactive`, surfaced interrupts are rejected and the run continues. With no policy flags at all, the historical `reportUnhandledInterrupts` crash is unchanged.

One deliberate behavior change beyond the prompt: with a non-interactive policy, an uncovered effect no longer clobbers a code-level approve with an in-chain reject — the default reject now applies only to interrupts that actually surface.

## Also in this PR

- `-i` shorthand for `--interactive`.
- Nicer prompt: effect name over a cyan horizontal rule, message in bold, interrupt data pretty-printed (omitted when empty).
- Fail-closed reject when stdin hits EOF while a prompt is pending (previously the process died with an unsettled top-level await, exit 13).

## Implementation

- `lib/runtime/policy.ts`: new `checkPolicyExplicit` returning `null` on fall-through; `checkPolicy`'s public propagate-on-fallthrough contract (re-exported via `std::policy`) is untouched.
- `lib/runtime/runPolicyHandler.ts`: chain handler reduced to explicit matches (explicit `propagate` rules now return a real `propagate`); new `resolveCliInterrupts` endpoint loop; prompt banner + EOF handling.
- `lib/backends/typescriptBuilder.ts` + `imports.mustache`: the bootstrap now runs `await resolveCliInterrupts(__result, respondToInterrupts)` instead of `reportUnhandledInterrupts(__result)`. All codegen fixtures rebuilt (the bulk of the diff: 2 lines x 93 fixture files).
- Docs: rewrote the `--interactive` section of the policies guide around the surfacing model.

## Testing

- Unit tests for the explicit-only handler, the endpoint loop (prompting, remembering, non-interactive reject, no-policy fallback, IPC guard), and the prompt formatter.
- New e2e spawn tests: a code-handled interrupt is not re-decided under `--interactive`; an uncovered surfaced interrupt is rejected and the run resumes to a clean exit.
- Verified live through a PTY: approve/reject/EOF answers, and the nested case — a `runCode` child raising an unhandled interrupt pauses, bubbles to the root CLI, prompts there, and resumes correctly down the subprocess tree on approve.
- 1667 tests pass across `lib/runtime`, `lib/cli`, and codegen suites; structural lint clean.

## Round 2: value-expecting interrupts (`expectsValue`)

An assignment-position raise (`const answer = raise interrupt("Question for user")`) expects its approver to provide a value — but nothing on the `Interrupt` said so. Now:

- **New `Interrupt.expectsValue: true` property**, set by the compiler for assignment-position raises (via a new trailing options arg to `interruptWithHandlers`; old 6-arg compiled calls keep working). Visible to handlers (`with (e) { e.expectsValue }`), across the IPC relay for nested runs, on surfaced `Interrupt` objects, and declarable from TS via `agency.interrupt({ expectsValue: true })`.
- **`--interactive` answers questions directly**: a surfaced value-expecting interrupt shows the banner and an `answer (empty line rejects):` prompt — the typed line becomes the approval value (assigned to the variable), an empty line or stdin EOF rejects. Statement raises keep the a/r/aa/rr prompt. Value interrupts bypass remembered `aa`/`rr` decisions both ways: a standing approve cannot answer a question.
- Verified via PTY: typed answer flows into the variable (`Hello, Adit!`), empty line rejects, and a mixed run shows both prompt styles in sequence. 2069 tests pass.

### DIFF
--- a/packages/agency-lang/docs/site/guide/policies.md
+++ b/packages/agency-lang/docs/site/guide/policies.md
@@ -85,16 +85,9 @@ Notes:
 
 - The built-in names are `recommended`, `minimal`, `with-writes`, and `approve-all`.
 - `--approve` and `--reject` take comma-separated [effect](/guide/effects) names. Reject takes precedence, so if you have an effect in both approve and reject, it gets rejected.
-- The [rules of handlers](/guide/handlers.md#the-rules-of-handlers) still apply. Think of the policy's explicit rules as just another handler: a `--reject` beats a handler's approve, and a handler's reject beats an `--approve`.
-- Effects the policy doesn't mention are left entirely to your program's handlers. The policy only steps in again for interrupts that *surface* — ones no handler settled, or ones a handler explicitly `propagate`d.
+- The [rules of handlers](/guide/handlers.md#the-rules-of-handlers) still apply. Think of this as just another handler.
 
 ### The `--interactive` flag
 
-An interrupt that nothing settles — no handler in your program approves or rejects it, and no explicit policy rule covers it — *surfaces to the user*, the same way it would surface to a TypeScript caller as a returned `Interrupt[]`. What happens then depends on how you ran the program:
-
-- With `--interactive` (or `-i` for short), you become that user: the CLI prompts you to approve or reject each surfaced interrupt, then resumes the run with your answer. Answering `aa` (approve-always) or `rr` (reject-always) remembers the decision for the rest of the run.
-- With a policy but no `--interactive`, surfaced interrupts are rejected and the run continues.
-- With no policy flags at all, a surfaced interrupt crashes the run with a pointer to the [handlers guide](/guide/handlers.md).
-
-Interrupts your program already handles never surface, so `--interactive` never re-asks about them. A handler that `propagate`s does the opposite — it forces the interrupt to surface, so it always reaches the prompt.
+By default, any interrupts that aren't addressed by your policy, and aren't handled by your agent, will cause a crash. Run with the `--interactive` flag if you want to be prompted to approve or reject those interrupts instead.
 
--- a/packages/agency-lang/docs/site/guide/policies.md
+++ b/packages/agency-lang/docs/site/guide/policies.md
@@ -85,9 +85,16 @@ Notes:
 
 - The built-in names are `recommended`, `minimal`, `with-writes`, and `approve-all`.
 - `--approve` and `--reject` take comma-separated [effect](/guide/effects) names. Reject takes precedence, so if you have an effect in both approve and reject, it gets rejected.
-- The [rules of handlers](/guide/handlers.md#the-rules-of-handlers) still apply. Think of this as just another handler.
+- The [rules of handlers](/guide/handlers.md#the-rules-of-handlers) still apply. Think of the policy's explicit rules as just another handler: a `--reject` beats a handler's approve, and a handler's reject beats an `--approve`.
+- Effects the policy doesn't mention are left entirely to your program's handlers. The policy only steps in again for interrupts that *surface* — ones no handler settled, or ones a handler explicitly `propagate`d.
 
 ### The `--interactive` flag
 
-By default, any interrupts that aren't addressed by your policy, and aren't handled by your agent, will cause a crash. Run with the `--interactive` flag if you want to be prompted to approve or reject those interrupts instead.
+An interrupt that nothing settles — no handler in your program approves or rejects it, and no explicit policy rule covers it — *surfaces to the user*, the same way it would surface to a TypeScript caller as a returned `Interrupt[]`. What happens then depends on how you ran the program:
+
+- With `--interactive` (or `-i` for short), you become that user: the CLI prompts you to approve or reject each surfaced interrupt, then resumes the run with your answer. Answering `aa` (approve-always) or `rr` (reject-always) remembers the decision for the rest of the run.
+- With a policy but no `--interactive`, surfaced interrupts are rejected and the run continues.
+- With no policy flags at all, a surfaced interrupt crashes the run with a pointer to the [handlers guide](/guide/handlers.md).
+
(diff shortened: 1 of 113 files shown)


packages/agency-lang/docs/site/guide/policies.md
```diff
@@ -85,9 +85,16 @@ Notes:
 
 - The built-in names are `recommended`, `minimal`, `with-writes`, and `approve-all`.
 - `--approve` and `--reject` take comma-separated [effect](/guide/effects) names. Reject takes precedence, so if you have an effect in both approve and reject, it gets rejected.
-- The [rules of handlers](/guide/handlers.md#the-rules-of-handlers) still apply. Think of this as just another handler.
+- The [rules of handlers](/guide/handlers.md#the-rules-of-handlers) still apply. Think of the policy's explicit rules as just another handler: a `--reject` beats a handler's approve, and a handler's reject beats an `--approve`.
```
egonSchiele: undo all changes to this file

packages/agency-lang/lib/runtime/runPolicyHandler.ts
```diff
@@ -144,12 +153,76 @@ export function installRunPolicyHandler(execCtx: {
...
+  respond: (
+    interrupts: Interrupt[],
+    responses: InterruptResponse[],
+  ) => Promise<RunNodeResult<any>>,
+  opts?: { prompt?: PromptFn },
+): Promise<RunNodeResult<any>> {
+  if (!hasInterrupts(result.data)) return result;
+  // No policy flag (or an IPC subprocess, which never owns the terminal):
+  // preserve the historical behavior — report and exit(1).
+  if (isIpcMode() || !loadEnvPolicy()) {
```
egonSchiele, line 247: Note that subprocess interrupts propagate to the parent too, so they should get surfaced to `--interactive` as well, if not handled

packages/agency-lang/docs/site/guide/policies.md
egonSchiele: Done — reverted the file to its main version in b0a2243c3.

packages/agency-lang/lib/runtime/runPolicyHandler.ts
egonSchiele: Yes — that works, and is now pinned by a test. The flow for a subprocess interrupt nothing handles: the child runs its local chain, relays to the root chain (which stays silent unless a policy rule matches), then the child pauses and the interrupt bubbles up through the pause machinery to the root CLI bootstrap, where the endpoint decides it (prompt under --interactive, reject otherwise) and resumes the whole tree via respondToInterrupts.

I verified the interactive path manually through a PTY (child raises unhandled std::error, root prompts, approve resumes the child to completion), and added an e2e spawn test in b0a2243c3 for the non-interactive path: the child observes the endpoint rejection as a failure Result and completes cleanly. One quirk the test surfaced: the nested run child cannot resolve agency-lang when the program lives in os.tmpdir() (the CLI resolver shim only covers the direct child), so that fixture runs from a temp dir inside the package instead.

packages/agency-lang/lib/runtime/runPolicyHandler.test.ts
```diff
@@ -131,26 +151,29 @@ describe("terminalPrompt", () => {
...
+  }
+  try {
+    await fn();
+  } finally {
+    for (const k of Object.keys(prev)) {
+      if (prev[k] === undefined) delete process.env[k];
+      else process.env[k] = prev[k];
     }
   }
+}
```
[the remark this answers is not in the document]
egonSchiele: The missing-await part was real — fixed in b0a2243c3: all withEnv call sites now await it, so env cleanup runs inside the test.

The AGENCY_IPC part is a false positive: in `{ [AGENCY_RUN_POLICY]: READ_OK, AGENCY_IPC: "1" }`, AGENCY_IPC is an object literal key (the string "AGENCY_IPC"), not an identifier — same pattern the pre-existing tests use.

packages/agency-lang/lib/runtime/runPolicyHandler.ts
```diff
@@ -101,38 +90,58 @@ export async function terminalPrompt(intr: Intr): Promise<PromptDecision> {
...
+    "",
+    color.bold(intr.message),
+  ];
+  const hasData =
+    intr.data != null &&
+    !(typeof intr.data === "object" && Object.keys(intr.data).length === 0);
+  if (hasData) {
+    lines.push(JSON.stringify(intr.data, null, 2));
+  }
+  lines.push("");
```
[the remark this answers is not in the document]
egonSchiele: Fixed in b0a2243c3: the data section now falls back to String(intr.data) if JSON.stringify throws (circular refs, BigInt), so the prompt always renders. Covered by a unit test.

Issue #516, opened 2026-07-11
Emit-once discipline: pure synth, diagnostics confined to check passes (#474 part 4)


egonSchiele:
Part 4 of #474 (the diagnostics overhaul PR shipped parts 1-3): make type synthesis pure and confine diagnostic emission to check passes.

Today `synthType` both computes types AND emits diagnostics (strict member access in `synthValueAccess`, property-does-not-exist, method arity, ...), and it runs in up to three passes over the same code (return-type inference, scope building, checkScopes). Emit-once correctness rests on two band-aids:

1. The `flowEnv`-unset suppression in `strictMemberAccessSeverity` ("stay silent during pre-flow passes").
2. `dedupeErrors` in `lib/typeChecker/index.ts` — now keyed on `code + message + loc.start` (upgraded from `message + loc.start` in the overhaul PR), with unit pins on both key directions.

The docs already state the right rule ("the flow-aware checkScopes pass is the single source of the diagnostic"). Making synth pure — type questions answer, they never complain; only check passes push — would enforce it structurally and delete both band-aids.

Notes from the overhaul PR that make this easier now:
- Every diagnostic flows through `diagnostic()` (lib/typeChecker/diagnostics.ts), so "who emits" is a grep for `errors.push(diagnostic(` — 70+ sites, all uniform.
- Diagnostics carry `name: DiagnosticName`, so a "which pass emitted what" assertion is cheap to write.
- The registry gives each diagnostic a stable identity for any emit-once bookkeeping.

PR #517, opened 2026-07-11
Diagnostics overhaul: stable codes, required severity, spans, template registry (#474)

### LINKED ISSUE #474: Diagnostics: stable error codes, required severity, full spans, emit-once discipline
`TypeCheckError` (`lib/typeChecker/types.ts`) is a message string plus optional fields, and the diagnostics system has a few gaps that get more expensive to fix the longer they wait. Go's `go/types` retrofitted error codes years after shipping and it was painful; Agency is early enough to do this cheaply.

## 1. Stable diagnostic codes

There are no error codes (TS2322-style). Codes buy:

- targeted suppression (better than the current message-matching in `suppression.ts`)
- per-diagnostic documentation pages
- tests that assert on codes instead of message strings (messages can then be reworded freely)

Proposal: `code: string` (e.g. `AG1001`) on `TypeCheckError`, assigned per diagnostic site, with a registry file so codes stay unique.

## 2. Required severity

`severity` is optional and defaults to `"error"` by omission. The `?? "error"` convention has already bitten once (downstream code reading `err.severity` directly gets `undefined`). Make it required at the type level and set it explicitly at every push site.

## 3. Full spans

Many errors are pushed with no `loc` at all (several sites in `lib/typeChecker/index.ts`, e.g. the type-alias validation loop), and `loc` is start-only where present. LSP integration will want start + end positions. Audit push sites for missing locations and extend `SourceLocation` usage to full spans where the parser has them.

## 4. Emit-once discipline (pure synth)

`synthType` both computes types AND emits diagnostics (strict member access in `synthValueAccess`), and it runs in multiple passes (return-type inference, scope building, checkScopes). Emit-once correctness currently rests on two mechanisms: the `flowEnv`-unset suppression in `strictMemberAccessSeverity`, and the blunt message+loc dedup in `deduplicateErrors()`. The docs already state the right rule ("the flow-aware checkScopes pass is the single source of the diagnostic") — making synth pure and confining emission to check passes would enforce it structurally instead of by convention. This piece is larger and can be split out if it grows.

egonSchiele:
Closes #474 (parts 1-3; part 4 — emit-once/pure synth — is filed as #516). Implements the spec and rev-2 plan committed on this branch.

## What

Every type-checker diagnostic now flows through one factory backed by one registry (`lib/typeChecker/diagnostics.ts`): a stable `AG####` code, a required severity, a required location (`null` only when the AST genuinely has none), a message rendered from a template, and a structured `params` payload. All 76 legacy `errors.push({...})` sites across 17 files are migrated; the compiler enforces it (`TypeCheckError` fields are required and the legacy `variableName`/`expectedType`/`actualType` fields are gone).

What you see:

```
main.agency:13:9 - error AG2001: Type 'string' is not assignable to type 'number'.
```

- **Suppression by code:** `// @tc-ignore AG2001, AG4005` suppresses only those codes on the next line. Bare `@tc-ignore` (and trailing prose) keeps its suppress-all meaning. A malformed code attempt (`AG201`, `ag2001`) suppresses NOTHING — fail closed, a typo must not widen the directive. All pinned.
- **Registry invariants tested:** unique codes, `AG\d{4}` format, no unconverted `${...}` interpolations (the likeliest sweep mistake), well-formed `{word}` placeholders with `{{...}}` escapes for literal braces, and `renderMessage` THROWS on a missing param rather than rendering "undefined" into user-facing text.
- **Byte-identical messages:** templates were extracted verbatim; the full suite passes with zero message-wording churn. Test changes are confined to: legacy-field assertions migrated to `params` (typeChecker.test.ts, variadicNamedBinding.test.ts, suppression.test.ts), and the new-format printer tests.

## Named behavior changes (deliberate, each pinned)

1. **Bare `@tc-ignore` now reaches errors that gained locations.** Errors pushed without a loc were immune to line suppression; the loc sweep fixes their anchors, so pre-existing directives may suppress more than before. Documented in docs/site/guide/types.md.
2. **Malformed suppression codes fail closed** (previously any directive suppressed everything).
3. **Public API:** `TypeCheckDiagnostic` (compiler) and `TypeCheckReport` (`std::agency.typecheck`) drop `variableName`/`expectedType`/`actualType` and gain `code` + `params` — stable codes are exactly what the writeAgency review loop should key on. stdlib docs regenerated for this type.
4. Statement-position `@tc-nocheck` / CLI behavior unchanged; `formatErrors` lost its dead `errorType` parameter (severity is always present now).

## Fixes found during execution (beyond the plan)

- **`buildCompilationUnit` dropped `fromFile` when no symbol table was supplied** — file stamping (and scope file tagging) silently lost the file identity in that path. Now stored unconditionally.
- **`checkType` gains a fallback loc:** literal value nodes (e.g. string literals) carry no loc, so `const x: number = "nope"` produced an unlocatable error; it now anchors on the assignment statement.
- **A missed 77th site:** `checkStaticMutation`'s method-mutation return (`.push(...)` on a static) was not in the plan inventory; swept as `AG7005`.
- **Template-literal brace escaping:** one message contains literal braces (`match (r) { … }`); templates support `{{`/`}}` escapes, covered by registry invariants and a render test.

## Audits (per plan)

- **Legacy-field consumers:** tsc flip fallout was 4 sites (public API mapper, one missed push site, two test files). The positive-evidence grep across `lib` + `tests` (ts + json) found only AST-node `variableName` fields — no serialized or `any`-typed consumers of the deleted error fields.
- **LLM path:** `std::agency.typecheck` returns the structured report (now with codes/params — strictly richer); the review agent execs `agency typecheck`, which prints via the upgraded `formatErrors`. Neither parses the old format. No statelog consumer reads typecheck error objects.
- **Config-driven severities (4 sites):** strict member access, match exhaustiveness, undefined functions, undefined variables — all four had existing warn-mode severity pins; none needed additions.
- **Deliberate `loc: null` list:** empty — no site hardcodes null. Null occurs only dynamically (reserved name on an IMPORTED alias with no local declaration node). The location-audit suite (`diagnosticLocations.test.ts`) pins the three previously loc-less shapes with real locations.

## Reviewer pointers

- The registry (~70 entries) is the bulk of the diff — each template is a verbatim extraction; the goldens in `diagnostics.test.ts` pin the highest-traffic ones byte-for-byte, and a registry↔runtime-formatter equality test locks the tool-binding wording that previously shared a formatter with the runtime backstop.
- Closed-set word params ({kind} = Function|Node, {argumentWord} = argument|arguments, subject refs like {who}/{handler}) are documented in the registry header — they preserve byte-identity without splitting entries per pluralization.
- Conditional phrasing became separate entries (method/call arity exact|at-least|range, named-args allowed|not-accepted, tool-unbound typed|untyped) — dissolving two nested ternaries and one string-building helper along the way.
- Residual risk, stated: the zero-churn gate is only as strong as existing message-assertion coverage; the `${` tripwire and goldens close the worst classes.


packages/agency-lang/docs/site/guide/types.md
```diff
@@ -167,6 +167,17 @@ node main() {
 }
 ```
 
+**`// @tc-ignore AG6019`** — Silences only the named diagnostic codes on the next line (every error now prints its `AG####` code). Separate multiple codes with commas or spaces; anything else on the directive line is treated as a comment. A malformed code (like `AG601` or lowercase `ag6019`) suppresses *nothing* — a typo must not silently widen to suppress-everything.
```
egonSchiele: Condense this paragraph into 1 - 2 more readable lines.

packages/agency-lang/docs/site/guide/types.md
```diff
@@ -167,6 +167,17 @@ node main() {
...
+
+```ts
+node main() {
+  // @tc-ignore AG6019
+  double("not a number")   // only the argument-type error is suppressed
+}
+```
+
+Note: every diagnostic now carries a source location, so a bare `// @tc-ignore` reaches some errors it previously could not (errors that once had no location were immune to line suppression).
```
egonSchiele: remove this paragraph

packages/agency-lang/a.vs.b.verdict.json
```diff
@@ -0,0 +1,15 @@
+{
```
egonSchiele: Stray artifact: this is an `eval optimize` A/B verdict output (generatedAt 2026-06-08, predates the branch), not part of the diagnostics work. It should be deleted from the branch — and probably added to `.gitignore` so it stops reappearing; the plan's Task 7 even listed deleting it as a cleanup step, but it got committed instead.

packages/agency-lang/stdlib/agency.agency
```diff
@@ -68,12 +68,15 @@ export type SourceLocation = {
...
+  code: string;
   severity: string;
   message: string;
   loc?: SourceLocation;
-  variableName?: string;
-  expectedType?: string;
-  actualType?: string
+  /** Structured payload of the diagnostic (the values rendered into the
+  message, e.g. the expected and actual type strings). */
+  params: Record<string, string>
```
egonSchiele: Type mismatch on the new public API: the TS side is `Record<string, string | number>`, and numeric params really do flow through — `toDiagnostic` passes `err.params` straight to the report, and the arity/type-args entries carry numbers (`count`, `max`, `min`, `expected`, `position`). An Agency program doing string ops on `d.params["count"]` will get a number despite this type saying `string`.

Either declare `Record<string, string | number>` here (if the checker handles that value union cleanly) or coerce values to strings in `toDiagnostic` and keep this type honest. Whichever way, `docs/site/stdlib/agency.md` needs a regen to match.

packages/agency-lang/lib/typeChecker/diagnostics.ts
```diff
@@ -0,0 +1,572 @@
...
+    message:
+      "{alias} expects at most {max} type {argumentWord}, got {count} (referenced in '{context}').",
+  },
+  tooFewTypeArgs: {
+    code: "AG1012",
+    severity: "error",
+    message:
+      "{alias} requires at least {min} type {argumentWord} (referenced in '{context}').",
+  },
+  bannedBuiltinInStaticInit: {
```
egonSchiele, line 182: `{contextLabel}` and `{reason}` are sentence fragments, which the header's own rule scopes out — the closed-set exception covers single WORDS (`{kind}`, `{argumentWord}`), and these are neither: `contextLabel` arrives as ``Static const `x` `` (embedding the variable name where params consumers can't reach it) and `reason` is a full clause from the banned-builtins table. Byte-identity forced the template shape, fine — but since `DiagnosticParams` explicitly allows extra keys, the sites could ALSO pass the structured pieces (`name` for the static; `reason` is at least a closed set from `BANNED_BUILTINS_IN_STATIC_INIT`). Same consideration for `{who}`/`{handler}` in the AG3xxx entries. Cheap now, annoying to retrofit once tooling starts reading `params`.

packages/agency-lang/lib/typeChecker/diagnostics.ts
```diff
@@ -0,0 +1,572 @@
...
+    code: "AG5002",
+    severity: "error",
+    message: "match is not exhaustive: missing {missing}.",
+  },
+  notAllPathsReturn: {
+    code: "AG2012",
+    severity: "error",
+    message: "Not all code paths return a value in '{fn}'.",
+  },
+  toolRequiredParamUnbound: {
```
egonSchiele, line 460: Verbatim-preserved grammar bug: "has required function-typed parameter '{param}' is unbound" (both this entry and the Typed variant). Correct call under the byte-identity gate — but a wording fix is exactly what stable codes make cheap now (code-based tests and suppressions survive it). Worth a follow-up issue so it doesn't fossilize.

packages/agency-lang/lib/typeChecker/toolBlockBinding.ts
```diff
@@ -175,10 +172,13 @@ export function checkToolBlockBindings(
...
 
 /**
- * Sole formatter site for the tool-binding diagnostics. All wording flows
- * through `formatRequiredUnboundError` / `formatOptionalUnboundWarning`,
- * which the runtime backstop also calls — that's what makes the
- * compile-time-vs-runtime unified-wording test (#42) pass by construction.
+ * Sole emit site for the compile-time tool-binding diagnostics. Wording now
+ * lives in the diagnostic registry (toolRequiredParamUnbound[Typed] /
+ * toolOptionalParamsDropped); the runtime backstop keeps its own formatters
+ * (lib/runtime/toolBlockDiagnostics.ts). Two locks keep the wording from
```
egonSchiele, line 178: Design note: this wording used to be single-sourced by construction (both compile-time and runtime called the shared formatter); now it lives twice — registry template + runtime formatter — held together by two equality tests. That works, but you could get by-construction back: `lib/runtime/toolBlockDiagnostics.ts` could render from the registry template via `renderMessage(DIAGNOSTICS.toolRequiredParamUnboundTyped.message, {...})` — `diagnostics.ts` has no runtime imports beyond types, so no cycle, and the equality tests become redundant instead of load-bearing. Fine to defer; flagging because the test-maintained variant tends to drift the moment someone edits the registry entry and mechanically updates the failing assertion.

packages/agency-lang/lib/typeChecker/index.ts
```diff
@@ -405,19 +362,104 @@ export class TypeChecker {
...
+    }
+    for (const def of Object.values(this.nodeDefs)) {
+      checkOne(def);
+    }
+  }
+
+  /** Locations of locally-declared type aliases, for diagnostics raised
+   *  from the alias TABLE (which carries no locs). Imported aliases are not
+   *  in program.nodes and resolve to null (file-level diagnostic). */
+  private collectAliasDeclLocs(): Record<string, SourceLocation> {
```
egonSchiele, line 436: Edge: keyed by bare alias name, so two same-named aliases in different scopes (a module-level `type T` shadowed inside a function scope) resolve to whichever declaration `walkNodes` visits last — a `reservedBuiltinTypeRedefined` or `typeParamDefaultOrder` diagnostic can anchor on the wrong declaration. It degrades to a slightly-misplaced (still same-file) location rather than anything worse, so fine as-is if you accept it — a one-line comment here acknowledging the limitation would do.

egonSchiele (commented):
Reviewed the full diff against the spec, the rev-2 plan, and the plan review. This is a faithful, well-executed sweep — every finding from the plan review landed in the implementation (dedup key includes the rendered message with both directions pinned; suppression fails closed on malformed codes with all three directive edge cases tested; `renderMessage` throws on missing params; the `plain()` helper strips real ESC bytes; warning-severity formatting is tested; the e2e stamping test passes an explicit `fromFile`; the location-audit suite exists; the public-API decision propagates `code`+`params` as recommended; `formatErrors` uses if/else; stamping is unconditional). Spot-checked conversions in 8 of the 17 swept files against the legacy strings — all verbatim, with config-driven severities correctly routed through `overrides` and `silent` still skipping the push. The `buildCompilationUnit` `fromFile` fix and the `checkType` fallback-loc are genuine catches beyond the plan.

Two things should change before merge, both small:

1. **`packages/agency-lang/a.vs.b.verdict.json` is a stray eval artifact** committed to the branch (inline comment).
2. **`stdlib/agency.agency` declares `params: Record<string, string>` but numeric params flow through** (inline comment) — a type lie on the new public API this PR is introducing.

The rest are notes/nits (fragment params vs the registry's own rule, the duplicated tool-binding wording now held by test instead of by construction, a same-named-alias anchoring edge, and a preserved grammar typo worth a follow-up now that wording changes are cheap). Details inline.

One process note: the registry invariant test suite would have caught none of the three duplicate-looking codes I checked (AG2001/AG4002/AG4005) because they were doc examples, not registry entries — the registry itself is clean and the uniqueness test guards it going forward.

packages/agency-lang/a.vs.b.verdict.json
egonSchiele: unrelated, can be deleted

packages/agency-lang/stdlib/agency.agency
egonSchiele: Fixed: `stdlib/agency.agency` now declares `params: Record<string, string | number>` (the checker takes the value union cleanly — verified by make compiling the stdlib) and `docs/site/stdlib/agency.md` is regenerated to match. Coercing in `toDiagnostic` would have kept the numbers out of reach for arithmetic on counts/positions, so the honest union won. (fba8f6b7)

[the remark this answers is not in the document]
egonSchiele: Fixed — the stdlib type now declares `Record<string, string | number>` and the generated docs are regenerated. See the sibling thread for detail. (fba8f6b7)

packages/agency-lang/a.vs.b.verdict.json
egonSchiele: Deleted from the branch and gitignored — at the REPO ROOT .gitignore, not the package level: the repo deliberately ignores nested .gitignore files, which I discovered when the package-level entry refused to stage. You are right about the mechanism: Task 7 listed the cleanup, but a batch git add -A in an earlier commit swept the artifact in first. The root entry ends the recurring-artifact dance for future worktrees too. (fba8f6b7)

packages/agency-lang/a.vs.b.verdict.json
```diff
@@ -0,0 +1,15 @@
...
+    "confidenceThreshold": 50,
+    "marginThreshold": 0,
+    "positionBias": "swap"
+  },
+  "winsA": 1,
+  "winsB": 0,
+  "ties": 0,
+  "winner": "A",
+  "perInput": []
+}
```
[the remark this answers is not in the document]
egonSchiele: Removed (and gitignored) — it was a stray `eval optimize` artifact, not a fixture. See the sibling thread. (fba8f6b7)

packages/agency-lang/a.vs.b.verdict.json
egonSchiele: Deleted and gitignored. (fba8f6b7)

packages/agency-lang/docs/site/guide/types.md
egonSchiele: Condensed to two lines (directive syntax + the fail-closed rule), keeping the example. (fba8f6b7)

packages/agency-lang/docs/site/guide/types.md
egonSchiele: Removed. The behavior change stays recorded in the PR body where it belongs. (fba8f6b7)

packages/agency-lang/lib/typeChecker/diagnostics.ts
egonSchiele: Agreed, and done cheaply now as you suggested: `checkBannedBuiltinCalls` threads the static's name through, so `bannedBuiltinInStaticInit`/`interruptInStaticInit` params carry a structured `staticName` alongside the display `contextLabel`; `handlerBodyRaises` adds `handlerName` when the handler is a functionRef; the raises exceedances add `sourceName` when one exists. `reason` stays as-is per your note (closed set from `BANNED_BUILTINS_IN_STATIC_INIT`). The display phrasings remain in the templates for byte-identity, but no consumer has to parse them anymore. (fba8f6b7)

packages/agency-lang/lib/typeChecker/diagnostics.ts
egonSchiele: Filed as #518, with the note that the reword is now a registry-entry-only change thanks to the codes (and single-sourced after the fix in the sibling thread). (fba8f6b7)

packages/agency-lang/lib/typeChecker/toolBlockBinding.ts
egonSchiele: Took the by-construction version rather than deferring — you are right that test-maintained equality drifts. `formatRequiredUnboundError` and `formatOptionalUnboundWarning` in `lib/runtime/toolBlockDiagnostics.ts` now render from the registry templates via `renderMessage` (no cycle: diagnostics.ts has only type-level imports). The equality tests stay as now-redundant pins, and the #42 unified-wording test still holds since the templates embed the canonical clause. (fba8f6b7)

packages/agency-lang/lib/typeChecker/index.ts
egonSchiele: Comment added acknowledging the last-visited-wins limitation and why it is accepted (same-file, strictly better than no location). (fba8f6b7)

PR #522, opened 2026-07-12
Replace safe with destructive and idempotent; failed tools stay callable

egonSchiele:
## Summary

Replaces the single `safe` keyword with two explicit markers — `destructive` and `idempotent` — and rebuilds failed-tool retry semantics around them. The motivating problem: PR #514 added a `_retryableFailure` band-aid so an agent could retry a failed `runCode`. That patch conflated "safe to retry" with "no side effects" and didn't engage with what actually makes a re-call dangerous. This PR removes the band-aid and the `retryable` field entirely.

**New model — a failed tool stays callable, with two ways to say more:**

| Marker | Re-callable after a failure? | Meaning |
|---|---|---|
| *(unmarked)* | Yes | Re-callable, but makes **no** promise it is safe to repeat blindly. |
| `idempotent` | Yes | Always safe to re-run (reads, pure calculations, lookups). |
| `destructive` | Only if it never started | Dangerous to repeat; locked out of the conversation once its body begins. A pre-execution failure (bad args) keeps it callable. |

The tool loop resolves a failure into one of four tiers — `destructive` > `neverStarted` > `idempotent` > `neutral` — and sends the model a tier-specific message. `destructive` failures that began execution get *"This tool can no longer be called in this conversation. Verify state manually."*; everything else gets *"You may call this tool again."*

**Retry semantics are fully decoupled from interrupts.** A marker says nothing about whether a tool is gated; an interrupt says nothing about whether a failed tool can be retried.

`safe` has been **removed entirely** (per review) — it is no longer a keyword. Marking a def both `destructive` and `idempotent` is a type error (`AG7006`), order-independent. `idempotent` is also supported on imports and `export from`, mirroring `destructive`.

## Implementation

- **Parser / AST / formatter**: `destructive` and `idempotent` on defs, named imports, and export-from; `FunctionMarkers` carried AST → symbol → registry → runtime; formatter renders both at all four marker sites.
- **Runtime**: `failure()` gains `neverStarted` and `destructiveRan` (birth-false); `stampFailureBoundary` (OR-fold at every function/block boundary) and `markDestructiveWork` are the single-source-of-truth helpers. `retryable` and the `__retryable` codegen machinery are deleted.
- **Codegen**: every function unconditionally inits `__destructiveRan` and stamps it into a departing failure; a destructive callee's outcome is folded conditionally so a *refused* destructive call does not poison its caller.
- **Serve**: MCP maps `destructive`→`destructiveHint`, `idempotent`→`idempotentHint`; the HTTP `/list` manifest reports the same as `destructive`/`idempotent` booleans (replacing the now-dead `safe` field).
- **Stdlib**: 253 `safe def` → `def`; 32 `destructive def`, 26 `idempotent def`; `runCode`/`writeAgency` intentionally left unmarked (neutral tier); `_retryableFailure` unwound.

## Testing

- Full `lib/` vitest suite: 8071 passed, 3 files skipped.
- New agency tests: `destructive-tracking`, `destructive-extern-throw`, `stdlib-destructive`, `safe-migration` (proves `safe` is inert — `destructiveRan === false` and absent from every registry).
- New agency-js `tool-tiers` test: deterministic coverage of the four-tier tool-loop behavior.
- Structural linter clean; `make fixtures` produces no diff.
- Real-LLM smoke: an agent given the unmarked `runCode` tool repeatedly received the neutral-tier *"The call failed. You may call this tool again."* message across retries, with no `_retryableFailure` and no lock-out — confirming the compile-failure retry path end-to-end.

## Notes / accepted scope

- Nodes keep parity with today's function-only tracking (no flag init/stamp), an accepted v1 gap documented in the spec.
- While writing the real-LLM smoke I hit a **pre-existing** bug unrelated to this PR: `writeAgency`'s `docsSkill("guide")` tool fails to revive during checkpoint serialization (`FunctionRefReviver` in `State.toJSON`). Not touched here.

## References

- Design spec: `docs/superpowers/specs/2026-07-11-destructive-keyword-tool-retry-design.md`
- Plan: `docs/superpowers/plans/2026-07-11-destructive-keyword-tool-retry.md`
- Three rounds of plan review (`-review.md`, `-review-2.md`, `-review-3.md`) — findings incorporated: single-source boundary helper, outcome-dependent destructive fold, `neverStarted` sole-producer contract, and the `safe`-inert migration guard.


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2250,20 +2264,25 @@ export class TypeScriptBuilder {
...
+      name: ts.str(functionName),
+      module: ts.str(this.moduleId),
+      fn: ts.id(implName),
+      params: ts.arr(paramNodes),
+      toolDefinition: toolDef,
+      exported: ts.bool(!!node.exported),
+    };
+    // Carry the retry-safety markers so the tool loop and MCP adapter can
+    // read them off the registered AgencyFunction. Emitted only when set.
+    if (node.markers?.destructive || node.markers?.idempotent) {
```
egonSchiele, line 2277: Does it make sense to have this in the TypeScript Builder? This will put it as part of the generated code, so the logic runs at generation time. Is it something that should run at runtime instead?

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -3836,6 +3855,30 @@ export class TypeScriptBuilder {
...
   }
 
+  /** If `stmt` is a simple assignment whose value is a direct call to a
+   *  destructive function (optionally `try`-wrapped), return the local
+   *  variable name its result binds to (`__self.<name>`). Used to emit the
+   *  outcome-dependent destructive flip. Returns null for bare calls,
+   *  nested-in-expression calls, patterns, access chains, or non-innermost
+   *  block targets — those get the conservative pre-flip instead. */
+  private destructiveOutcomeVar(stmt: AgencyNode): string | null {
+    if (stmt.type !== "assignment") return null;
```
egonSchiele, line 3865: What if it's a type of node that can contain an assignment such as a thread block, if statement, for loop, etc?

egonSchiele: Also, for the changes related to the `destructive` and `idempotent` keywords that you're making to this file, I would suggest moving the changes to a different file if possible, and exposing a declarative interface that you can use from this file. That will make this code easier to read, and make this file more maintainable, as it is too long right now.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -3902,6 +3960,19 @@ export class TypeScriptBuilder {
...
       }
+      if (destructivePostVar) {
+        // The destructive callee's result is bound at `__self.<var>`
+        // (__self === stack.locals). OR its outcome into the activation:
+        // a returned failure contributes its own destructiveRan (so a
+        // swallowed destructive failure still marks us); success or a
+        // plain value means the destructive work ran, so mark true.
+        if (!currentPart) currentPart = [];
+        currentPart.push(
+          ts.raw(
```
egonSchiele, line 3971: This is pretty ugly, is there is no way to write this using TSIR builders?

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3403,6 +3404,16 @@ const safeNameItem = or(
       name: r.name,
       alias: r.alias as string | undefined,
       isSafe: true,
+      isDestructive: false,
+    }),
+  ),
+  map(
```
egonSchiele, line 3413: Do the keywords `destructive` and `idempotent` need to be in a specific order? And what happens if a user marks a function as both destructive and idempotent? Does that trigger an error? I would probably expect that to be a type checker error.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4655,45 +4691,74 @@ const exportKeywordParser: Parser<boolean> = or(
...
+      isIdempotent: seen.idempotent,
+    },
+  };
 }
 
 const _functionParserInner: Parser<FunctionDefinition> = (input: string) => {
   const mods = parseFunctionModifiers(input);
   if (!mods.success) return failure("unexpected modifier", input);
-  const { isExported, isSafe } = mods;
+  const { isExported, isSafe, isDestructive, isIdempotent } = mods.modifiers;
```
egonSchiele: Is `safe` still supported then? Are we not removing that?

packages/agency-lang/lib/runtime/memory/manager.ts
```diff
@@ -260,7 +260,6 @@ export class MemoryManager {
             errorType: "llmError",
             message: String(result.error),
             functionName: phase,
-            retryable: false,
```
egonSchiele, line 263: I'm not sure what function you are removing this line from, but if it is really not retryable, shouldn't it be marked destructive?

packages/agency-lang/lib/typeChecker/deprecatedSafe.ts
```diff
@@ -0,0 +1,35 @@
+import { diagnostic } from "./diagnostics.js";
+import type { TypeCheckerContext } from "./types.js";
+
+/**
```
egonSchiele: I think we can just remove `safe`, no need to deprecate it. This language does not have too many users so we can just make a hard breaking change unless you feel like we will need to update too many files to make that happen.

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -1082,30 +1113,35 @@ export async function runPrompt(args: {
...
       } finally {
         ctx.exitToolCall();
       }
 
+      // Decision 8: destructive work performed via a tool inside an llm()
+      // call must propagate to the calling function's activation, so a later
+      // failure THERE reports destructiveRan. Write the same locals slot the
+      // codegen flag lives in; the enclosing function's exit stamp reads it.
+      if (
+        handler.markers?.destructive ||
```
egonSchiele: This fires on `handler.markers?.destructive` unconditionally, so a destructive tool that fails **pre-execution** (bad args → `neverStarted: true`, `destructiveRan: false`) still calls `markDestructiveWork(stack)` and poisons the *caller* frame — even though the body never ran and the tool itself is correctly classified into the `neverStarted` tier and kept callable.

The codegen assignment-flip path deliberately avoids exactly this (`isFailure(r) ? r.destructiveRan : true` — a returned never-started/refused destructive failure contributes false). Here you have the `neverStarted` signal directly, so you can be precise rather than conservative:

```ts
const started = !(isFailure(toolResult) && toolResult.neverStarted);
if (
  (handler.markers?.destructive && started) ||
  (isFailure(toolResult) && toolResult.destructiveRan)
) {
  markDestructiveWork(stack);
}
```

It errs safe (spurious lockout, not unsafe re-execution), but it contradicts the `neverStarted` contract and could wrongly mark an enclosing agent-as-tool failure destructive under nesting. No integration test covers the destructive-tool-fails-pre-execution → caller-frame path. Is the conservative marking intentional here, or should it mirror the assignment flip?

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -202,6 +202,34 @@ function toolErrorMessage(error: any): string {
...
+  if (markers?.idempotent) return "idempotent";
+  return "neutral";
+}
+
+const TIER_SUFFIX: Record<FailureTier, string> = {
+  destructive:
+    "The call failed after starting a destructive operation. This tool can no longer be called in this conversation. Verify state manually.",
+  neverStarted: "Nothing was executed. Correct the arguments and call again.",
+  idempotent:
+    "This tool is idempotent: calling it again is safe. Correct the arguments and call again.",
```
egonSchiele: Minor wording: `Correct the arguments and call again` presumes an argument error, but an idempotent tool can fail for reasons unrelated to args (transient network, upstream 500). Consider `This tool is idempotent: calling it again is safe.` without the arg-specific imperative, or reserve the "correct the arguments" phrasing for the `neverStarted` tier (where it is always true).

egonSchiele (commented):
## Review — replace `safe` with `destructive`/`idempotent`

Read the core mechanism end-to-end (parser modifiers → `FunctionMarkers` → symbol table / compilation-unit registries → codegen `__destructiveRan` init+stamp → runtime `failure()`/`stampFailureBoundary`/`markDestructiveWork` → tool-loop four-tier classification → serve adapters). The design is coherent and the single-source boundary-helper refactor reads well. The migration is thorough: `safe` is genuinely inert (no registry feed, byte-identical metadata, AG7006 warning), and the parser modifier table + mutual-exclusivity check is a clean generalization of the old two-slot loop.

**Strengths**
- `FunctionMarkers` as one object threaded through every layer (with a `registerMarkers` chokepoint in `compilationUnit.ts` and a single spread-inherit path in `mergeOne`) avoids the parallel-boolean-mirror problem well.
- Good adversarial-boundary handling: `stampFailureBoundary` OR-fold is sticky and tested; the outcome-dependent post-flip in codegen correctly avoids poisoning a caller when a destructive callee *returns* a refused/never-started failure.
- Solid unit coverage of `failureTier` precedence, `stampFailureBoundary` stickiness, and `neverStarted`/`destructiveRan` birth defaults; the `preExecution` tag has both a positive and negative test.

**Main concern (see inline on `prompt.ts`):** the tool-loop Decision-8 propagation marks the *caller's* frame on `handler.markers?.destructive` unconditionally, ignoring the `neverStarted` signal it has in hand. A destructive tool that fails pre-execution (bad args from the model) still marks the enclosing function's activation as `destructiveRan`, even though its body never ran. This is inconsistent with the outcome-dependent assignment flip in codegen, which deliberately does *not* poison the caller in the analogous never-started case. It errs safe (over-lockout, not unsafe re-execution) but contradicts the documented `neverStarted` semantics and is untested at the integration level.

**Minor**
- MCP adapter drops `readOnlyHint` entirely; `idempotent` now maps only to `idempotentHint`. Semantically defensible (idempotent ≠ read-only), but it's a silent behavior change for existing MCP clients that keyed on `readOnlyHint` — worth a line in the PR notes.
- Idempotent-tier message wording (inline note).

Nothing here is a merge blocker; the Decision-8 point is the one I'd want a second look at before merge.

packages/agency-lang/lib/runtime/result.test.ts
```diff
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { success, failure, isSuccess, isFailure, __pipeBind, __tryCall } from "./result.js";
+import { success, failure, isSuccess, isFailure, __pipeBind, __tryCall, stampFailureBoundary, markDestructiveWork } from "./result.js";
```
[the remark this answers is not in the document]
egonSchiele: Fixed. `propagateFailure` is now imported and the test actually calls it, asserting `neverStarted`/`destructiveRan` survive the spread plus the appended skip entry.

packages/agency-lang/lib/runtime/result.test.ts
```diff
@@ -206,23 +207,62 @@ describe("failure", () => {
...
+    expect(f.neverStarted).toBe(true);
+    expect(f.destructiveRan).toBe(true);
+  });
+
+  it("propagateFailure preserves the new fields", () => {
+    const f = failure("boom", { destructiveRan: true });
+    // propagateFailure spreads; new fields must survive.
+    expect(f.destructiveRan).toBe(true);
+    expect(f.neverStarted).toBe(false);
+  });
```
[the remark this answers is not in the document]
egonSchiele: Fixed — same change as above: the test now calls `propagateFailure(f, {name, param})` and checks the classification fields carry through onto the propagated copy.

packages/agency-lang/lib/types/importStatement.ts
```diff
@@ -17,6 +17,9 @@ export type NamedImport = {
   type: "namedImport";
   importedNames: string[];
   safeNames: string[];
+  /** Source-side names marked `destructive`. Present only when non-empty
+   *  (matching `testOnly` above) so exact-match AST comparisons stay clean. */
+  destructiveNames?: string[];
   aliases: Record<string, string>;
```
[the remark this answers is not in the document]
egonSchiele: Added. `idempotent` is now supported on imports (new `idempotentNames` on `NamedImport`), threaded through the parser, symbol table, import resolver, formatter, and both serve adapters — symmetric with `destructive`.

packages/agency-lang/lib/types/exportFromStatement.ts
```diff
@@ -15,6 +15,8 @@ export type NamedExportBody = {
   aliases: Record<string, string>;
   /** Source-side names marked with the `safe` modifier. */
   safeNames: string[];
+  /** Source-side names marked `destructive`. Present only when non-empty. */
+  destructiveNames?: string[];
 };
```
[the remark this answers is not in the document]
egonSchiele: Added for `export { ... } from` too — `NamedExportBody.idempotentNames`, with the re-export merge (`mergeOne` `forceIdempotent`) propagating it onto the wrapper.

packages/agency-lang/lib/statelogClient.ts
```diff
@@ -1216,13 +1216,18 @@ export class StatelogClient {
     message,
     functionName,
     retryable,
+    neverStarted,
+    destructiveRan,
```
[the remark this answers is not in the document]
egonSchiele: Removed. The statelog `error()` event no longer carries `retryable` — the new model uses `neverStarted`/`destructiveRan`, and the one remaining setter (thread.ts `attachToReply`) was passing a meaningless `retryable: false`.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Markers are static, compile-time metadata (known from the source), so baking them into the `AgencyFunction.create` registration is the same pattern as `exported`, `params`, and `toolDefinition` — there is nothing to defer to runtime. The runtime just reads `handler.markers` off the registered tool. (The *destructive-tracking* codegen below is the part that emits runtime logic; that is what the extraction/TSIR refactor addresses.)

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Good question — it is handled, conservatively. `containsDestructiveCall`/`containsImpureCall` walk the whole statement subtree via `bodySlots` (which descends into if/for/while/thread/handle bodies), so a destructive call inside any nested block triggers the function-level conservative pre-flip, writing the function's `__self` that the exit stamp reads. It never under-marks; it can over-mark (the flip fires even if the branch is not taken). Added `ifBlock`/`forBlock` cases to destructive-tracking.agency to lock this in.

egonSchiele: Agreed the file is too long and this logic should live in its own module behind a declarative interface. I have scoped that (plus the TSIR-builder rewrite below) into a dedicated follow-up PR so it can be reviewed on its own rather than buried in this one. This PR keeps the destructive-tracking behavior + tests; the follow-up is a pure refactor.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Agreed — the raw `ts.raw(...)` string is ugly. Rewriting it with TSIR builders is part of the extraction follow-up PR referenced above.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Order-independent already (the modifier parse loops to fixpoint, so `destructive idempotent` and `idempotent destructive` are equivalent). Marking both is now a type-checker error rather than a parse failure: the parser accepts both onto the AST and the new `checkConflictingMarkers` pass reports AG7006 ("cannot be both destructive and idempotent"). New tests in conflictingMarkers.test.ts and function.test.ts.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Removed entirely now (your other comment) — `safe` is no longer a keyword anywhere. This line is gone.

packages/agency-lang/lib/runtime/memory/manager.ts
egonSchiele: That line was a statelog `error()` call, not a tool marker — `retryable` was an observability field on the error event, unrelated to whether the function is destructive. Since the new model carries retry-safety as `neverStarted`/`destructiveRan`, I removed the `retryable` statelog field entirely (see the statelogClient thread), so this and the other callers no longer pass it.

packages/agency-lang/lib/typeChecker/deprecatedSafe.ts
egonSchiele: Done — removed `safe` entirely instead of deprecating. It touched ~19 lib files + a batch of test fixtures, but all mechanical; no deprecation pass, no AG7006 warning. `deprecatedSafe.ts` is deleted (repurposed into `conflictingMarkers.ts`).

egonSchiele:
## Review round addressed (3 commits)

**Test fixes (Copilot):**
- `propagateFailure` test now imports and calls the function it claims to test.
- Removed the vestigial statelog `retryable` field.

**`safe` removed entirely (not deprecated):** dropped from the parser, AST, symbol table, `AgencyFunction`, the `safeFunctions` registry, the import/export `safeNames`, the formatter, and the resolver. `deprecatedSafe.ts` + the AG7006 warning are gone. ~19 lib files + fixtures, all mechanical.

**`idempotent` on imports / `export from`** (Copilot's asymmetry finding): mirrors `destructive` end-to-end via new `idempotentNames`.

**destructive+idempotent conflict is now a type error** (AG7006, order-independent) rather than a generic parse failure.

**Bug found while cleaning up:** the four `AgencyFunction` reconstruction sites (`withToolDefinition`/`partial`/`preapprove`/`rename`) never carried `markers`, so a marked tool rebuilt via any of them silently lost its marker. Fixed.

**Nested-block tracking** (your if/for/thread question): verified safe (subtree walk via `bodySlots`) and added `ifBlock`/`forBlock` tests.

**Deferred to a follow-up PR:** extracting the destructive-tracking codegen into its own module with a declarative interface, and replacing the raw-string emission with TSIR builders. Kept out of this PR so it can be reviewed as a pure refactor.

Full `lib/` vitest green (8068 passed); destructive agency + tool-tiers tests green; lint clean; `make fixtures` no-diff.

packages/agency-lang/lib/compilationUnit.ts
```diff
@@ -109,7 +110,11 @@ export type CompilationUnit = {
   graphNodes: GraphNodeDefinition[];
   importedNodes: ImportNodeStatement[];
   importStatements: ImportStatement[];
-  safeFunctions: Record<string, boolean>;
+  /** Local names (defs + imports + re-exports) marked `destructive`. Read
+   *  by codegen's containsDestructiveCall to track destructive execution. */
+  destructiveFunctions: Record<string, boolean>;
+  /** Local names marked `idempotent`. */
+  idempotentFunctions: Record<string, boolean>;
```
egonSchiele, line 121: Dead registry: `idempotentFunctions` is written by `registerMarkers` (line ~179) but never read anywhere in `lib/` (non-test). The `destructive` twin is consumed by `NameClassifier.containsDestructiveCall`, but idempotent has no codegen role — it only affects the tool-loop tier message, which reads `handler.markers?.idempotent` off the registered `AgencyFunction` (fed from the defs own `markers`), not this registry. `getImportedIdempotentToolNames()` in `types/importStatement.ts` is likewise defined but uncalled.

Not a bug, but it reads as if idempotent participates in call-site analysis when it does not. Either drop the registry + getter, or add a one-line comment that it exists purely for symmetry with `destructiveFunctions` / future use.

egonSchiele (commented):
## Re-review of the 3 follow-up commits

Re-read the delta since my first pass (`f99c7300` → `03075a92`). The new work is clean:

- **`AgencyFunction` marker-reconstruction fix** — good catch. `withToolDefinition`/`partial`/`preapprove`/`rename` now all carry `markers`; without it a `.rename()`d or partially-applied destructive tool would have silently lost its marker and dropped to the neutral tier. Verified all four `create({...})` sites pass `markers: this.markers`.
- **`destructive`+`idempotent` → `conflictingMarkers` (AG7006 error)** — the parser correctly stops rejecting the combo (both land on the AST) and the type checker reports a clear, order-independent diagnostic. Good altitude choice vs. the old generic parse failure. The per-item import `or` means a single import name can't carry both, so the def-only check is the right scope.
- **Full `safe` removal** — thorough; no dangling `safeNames`/`safeFunctions`/`safe` field left in the non-test tree.
- **Nested-block tests** — `ifBlock`/`forBlock` plus the existing halt/join/try-swallowed/refusal/succeeded-then-failed cases make the codegen tracking path genuinely well-covered.

### Still open from round 1 (neither addressed nor replied to)

1. **`prompt.ts` Decision-8 caller-frame poisoning on a pre-execution destructive failure** (my inline on the `handler.markers?.destructive` block). This is the one substantive correctness item and it's unchanged — a destructive tool that fails on bad args (`neverStarted`, body never ran) still marks the caller's activation `destructiveRan`, diverging from the outcome-dependent assignment flip the codegen deliberately uses. The excellent new `destructive-tracking.agency` covers the *codegen* refusal path (`cleanRefusal` asserts `destructiveRan === false`) but there is no equivalent coverage for the *tool-loop* refusal path, which is exactly where the asymmetry lives. Worth either fixing (`&& !neverStarted`) or an explicit "intentionally conservative here" note.
2. Idempotent-tier message wording (minor, inline still open).

### New minor finding

See inline on `compilationUnit.ts`: the `idempotentFunctions` registry (and `getImportedIdempotentToolNames()`) are populated/defined but never read — only the `destructive` counterparts are consumed (by `containsDestructiveCall`). Idempotent reaches the runtime through symbol `markers`, not this registry, so these are dead.

Overall this is in good shape — #1 is the only thing I'd want resolved (fix or explicit decision) before merge.

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: You're right that the condition is wrong, and chasing it surfaced something bigger. Two parts:

**1. The condition (fixed).** I made it mirror the codegen assignment-flip exactly — for a failure, trust `toolResult.destructiveRan`; only a success of a destructive-marked tool counts as work done:
```ts
const toolDidDestructiveWork = isFailure(toolResult)
  ? toolResult.destructiveRan
  : !!handler.markers?.destructive;
```
This is even more precise than the `started` guard — it also excludes a *clean refusal* (a destructive def that returns a failure before its effectful work: `neverStarted:false`, `destructiveRan:false`), matching `isFailure(r) ? r.destructiveRan : true`.

**2. The deeper bug.** While writing the missing integration test I found that `markDestructiveWork(stack)` here targets **runPrompt's own frame**, not the caller's. `stack` comes from `runPrompt`'s `setupFunction()` → `getNewState()` (prompt.ts:789), a fresh State whose `.locals` is discarded when runPrompt returns. So decision-8 has never propagated tool-work to the caller at all — which is *why* the over-poisoning you flagged never actually reaches the caller.

Verified with a deterministic probe: a destructive tool that **succeeds** inside `llm()`, followed by an unrelated caller failure, leaves the caller's failure `destructiveRan === false`. It should be `true` if decision-8 worked.

The correct fix is to target the caller's activation frame — the same way `removedTools: __self.__removedTools` is already threaded by-reference into the `runPrompt({...})` emission (agent codegen). I'd pass the caller's `__self` in and mark that. But that edits the `llm()` codegen — the exact area you asked to extract/refactor separately — and regenerates every llm() fixture.

So: do you want the frame-targeting fix in this PR, or folded into the codegen-extraction follow-up (where the emission is already being rewritten)? I've committed the corrected condition here either way.

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: Fixed. The idempotent tier message is now just "This tool is idempotent: calling it again is safe." — the "Correct the arguments" imperative is reserved for the neverStarted tier, where it is always accurate.

packages/agency-lang/lib/compilationUnit.ts
egonSchiele: Removed the two getters (`getImportedDestructiveToolNames`/`getImportedIdempotentToolNames`) — I'd added them speculatively and nothing calls either. For `idempotentFunctions` I took the comment route: documented that it is populated for symmetry with `destructiveFunctions` but has no codegen consumer today (the tier reads `handler.markers?.idempotent` off the registered AgencyFunction, not this registry).

packages/agency-lang/docs/site/guide/llm-part-2.md
```diff
@@ -14,18 +14,51 @@ print(response)
...
-Some functions are okay to retry and some aren't. If you have a function that has a side effect, like writing to a database, you probably don't want the LLM to retry it automatically. However, if you have a function that doesn't have any side effects, you can mark it `safe` to retry:
+The question is: should it be *allowed* to try again? For most tools, yes — retrying a failed read or a failed calculation is harmless. But some tools are dangerous to re-run. If a tool charged a credit card and then failed, blindly calling it again might charge the card twice. Agency lets you say which tools are which.
+
+By default, a tool that fails stays callable. The model can call it again. This is the right default for the common case, but note what it does *not* say: it does not promise the tool is safe to re-run automatically. A human or an automated system should not assume an unmarked tool can be blindly repeated.
+
+### `destructive`
+
+Mark a tool `destructive` when re-running it could cause real damage — charging a card, sending an email, deleting a file:
+
+```ts
```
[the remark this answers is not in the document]
egonSchiele: Fixed — changed to an `agency` fence.

packages/agency-lang/docs/site/guide/llm-part-2.md
```diff
@@ -14,18 +14,51 @@ print(response)
...
+
+If a `destructive` tool starts running and then fails, Agency removes it from the conversation. The model cannot call it again, and the tool result tells it so: the operation may have partially completed, so the state must be verified by hand rather than by retrying.
+
+There is one exception. If the call fails *before the tool body starts* — the model passed the wrong arguments, or too few — then nothing happened yet, so the tool stays callable. Only a failure *after* the destructive work began locks the tool out.
+
+### `idempotent`
+
+Mark a tool `idempotent` when re-running it is always safe, no matter how many times it happens — reading a record, looking something up, a pure calculation:
 
 ```ts
```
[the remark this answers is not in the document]
egonSchiele: Fixed — `agency` fence.

packages/agency-lang/lib/preprocessors/prunePreludeShadows.ts
```diff
@@ -44,9 +44,16 @@ export function prunePreludeShadows(program: AgencyProgram): void {
...
+      if (spec.destructiveNames) {
+        spec.destructiveNames = spec.destructiveNames.filter((n) =>
+          spec.importedNames.includes(n),
+        );
+      }
+      if (spec.idempotentNames) {
+        spec.idempotentNames = spec.idempotentNames.filter((n) =>
+          spec.importedNames.includes(n),
+        );
+      }
```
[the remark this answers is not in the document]
egonSchiele: Fixed. Both marker lists are now deleted when filtering empties them, so the "present only when non-empty" invariant holds after pruning shadowed prelude imports.

egonSchiele (commented):
## Round-3 re-review — all open findings resolved

Diffed `03075a92` → `5afc4112`. Every item from my prior passes is addressed:

1. **Decision-8 caller-frame poisoning (the substantive one) — fixed, and fixed better than I proposed.** `toolDidDestructiveWork = isFailure(toolResult) ? toolResult.destructiveRan : !!handler.markers?.destructive` now mirrors the codegen assignment-flip exactly. Walked the cases:
   - destructive tool succeeds → `!!markers.destructive` = true → marks ✓
   - fails after starting → `destructiveRan` true → marks ✓
   - fails pre-execution (`neverStarted`) → `destructiveRan` false → does **not** poison caller ✓ (the original bug)
   - returns a clean refusal before effectful work (`destructiveRan` false, `neverStarted` false) → does **not** poison ✓ — this case my `&& !neverStarted` suggestion would have missed, so trusting the `destructiveRan` bit is the stronger fix.
   - non-destructive tool whose failure already carries `destructiveRan` (inner destructive call) → still propagates ✓
   Consistent with the marker-driven (not deep-effect) tracking model throughout.
2. **Idempotent-tier wording** — trailing "Correct the arguments and call again" dropped. ✓
3. **Dead idempotent surface** — `getImportedDestructiveToolNames`/`getImportedIdempotentToolNames` removed; the `idempotentFunctions` registry kept with an explicit "no codegen role, symmetry/future-use" comment. ✓

Bonus: the `prunePreludeShadows` change correctly `delete`s an emptied `destructiveNames`/`idempotentNames` rather than leaving a `[]`, preserving the present-only-when-non-empty AST invariant — good catch on a subtlety I hadn't flagged.

Only residual (non-blocking): there's still no *integration* test exercising the tool-loop pre-execution/refusal path for a destructive tool (the codegen equivalent is well covered by `destructive-tracking.agency`). The logic now provably mirrors that tested path, so this is optional.

LGTM from my end — no remaining concerns.

Issue #518, opened 2026-07-11
Reword the tool-unbound diagnostics (AG6028/AG6029): grammar bug preserved by the byte-identity gate


egonSchiele:
Found in the PR #517 review: the tool-binding diagnostics (AG6028/AG6029) carry a verbatim-preserved grammar bug — "Tool 'X' has required function-typed parameter 'p' is unbound" ("has ... is unbound" is not a sentence).

The overhaul deliberately extracted templates byte-identically, so the bug survived. Fixing wording is exactly what stable codes make cheap now: tests and suppressions key on AG6028/AG6029, so the reword touches only the registry entry (both the checker and the runtime formatter render from it as of the review round).

Suggested: "Tool 'X' has a required function-typed parameter 'p' that is unbound. Bind it with .partial(p: <value>) before passing as a tool." — also update the runtime `formatUnboundClause` framing to match, and sweep the registry for any sibling awkwardness while in there (one deliberate pass, message-only, no code changes).

egonSchiele:
Fixed in #1037, squash merged to main.

PR #524, opened 2026-07-12
Diagnostic explanations: agency explain, generated docs, agent knowledge

egonSchiele:
Follow-on to #474 / #517 (the diagnostic codes + registry). Implements the spec and plan committed under `docs/superpowers/`. Serves a long-form explanation for every type-checker diagnostic three ways, plus a one-line discovery hint on failed type checks.

## What you get

```
$ agency typecheck bad.agency
bad.agency:2:3 - error AG2001: Type '"nope"' is not assignable to type 'number' (assignment to 'x').
Run 'agency explain AG2001' for an explanation.

$ agency explain AG2001
AG2001 typeNotAssignableInContext
severity: error

Type '{actual}' is not assignable to type '{expected}' ({context}).

Agency assigns each value a type and checks that it fits where you use it. ...
```

Three surfaces, all driven by one authored piece of new content per code (the long explanation) alongside the message template that already lives in the registry:

1. **CLI** — `agency explain <code>` (accepts an `AG####` code, case-insensitive, or a registry name like `typeNotAssignable`) and `agency explain --list` (every code grouped by category). Rendering lives in a testable `lib/cli/explain.ts`; the command is a thin wrapper. Named `explain`, not `diagnostics`, because `agency diagnostics` already exists as a VSCode-facing parse command.
2. **Docs** — a build script (`make diagnostics-docs` → `docs/site/diagnostics/`) generates an index plus one page per category from the registry + explanations. Pure generation lives in `lib/cli/diagnosticsDocs.ts` and is unit-tested in memory; `stage-stdlib-docs` copies the pages into `stdlib/docs` for the agent.
3. **Agent** — `docsSkill` gains a `"diagnostics"` section, wired into all four docsSkill-bearing subagents (oracle, research, **code**, explorer); the code subagent runs `typecheck` and is the primary consumer.

## Design choices

- **Explanations are an exhaustive `Record<DiagnosticName, string>`** in `lib/typeChecker/diagnosticExplanations.ts` — coverage is a compile guarantee, not a test. Adding a registry entry without an explanation fails `tsc` (this caught the four codes added after #517: the three tool-unbound entries and `conflictingMarkers`). 86 codes explained.
- **Category ranges became data** (`DIAGNOSTIC_CATEGORIES` + `categoryForCode` in `diagnostics.ts`) — they were prose in the registry header; now the CLI `--list` grouping and the docs generator share one source. A registry-suite invariant pins that every code's prefix maps to exactly one category, so a future AG8xxx code without a category entry fails there with an actionable message.
- **The hint targets the first error-severity diagnostic** (`errors.find(e => e.severity === "error")`, not `errors[0]`), so a warning-led error list still names the error. `formatDiagnosticsHint` lives next to `formatErrors` and is wired only at the two human/agent-facing print sites (typecheck action + buildSession failure), on the same stream as the error block — never inside `formatErrors`, so programmatic consumers stay clean. Warnings-only output gets no hint.

## Tests

- Coverage/quality: every explanation is substantial, leaks no `${...}` or unrendered `{placeholder}` outside code spans, and **every fenced `agency` snippet is run through `parseAgency`** — turning the authoring convention into a mechanical gate (this repo has a documented history of prose-syntax mistakes).
- CLI helper: code↔name resolve identically, unknown code returns the suggestion, list contains every code once.
- Generator invariants: index lists every code once with a link, each code appears on exactly its own category page, unique anchors.
- Hint: names the first error-severity code with a warning-first discriminating case; null on success/warnings-only.
- `_docsDir` union pin.

`tsc --noEmit` clean; 1647 tests pass across `lib/typeChecker` + `lib/cli` locally with no regressions. (The pre-existing warn-mode `settings.agency` fixture warning is unrelated.)

## Non-goals (unchanged from #474/#517)

Parser errors stay code-less; no message-template rewording (that is #518); no `agency doc` integration (these docs are generated from our registry by our build, never user-run).

### DIFF
--- a/packages/agency-lang/docs/site/diagnostics/checking.md
+++ b/packages/agency-lang/docs/site/diagnostics/checking.md
@@ -4,6 +4,8 @@ name: "Assignability and checking"
 
 # Assignability and checking
 
+<a id="ag2001"></a>
+
 ## AG2001 — Type '{actual}' is not assignable to type '{expected}' ({context}).
 
 *Default severity: error.*
@@ -11,6 +13,8 @@ name: "Assignability and checking"
 Agency assigns each value a type and checks that it fits where you use it. This is the assignability error with the surrounding context named (a return, an argument, a field) — the value's type does not fit that slot.
 
 **How to fix:** change one side so they line up — convert the value, widen the declared type, or fix the expression that produced the wrong type. If the value can legitimately be several types, declare the slot as a union.
+
+<a id="ag2002"></a>
 
 ## AG2002 — Type '{actual}' is not assignable to type 'boolean' (condition).
 
@@ -20,6 +24,8 @@ The condition of an `if` or `while` must be a boolean. Agency does not treat non
 
 **How to fix:** compare explicitly — e.g. `count > 0` instead of `count`, or `name != ""` instead of `name`.
 
+<a id="ag2003"></a>
+
 ## AG2003 — Unknown property '{key}' on type '{expected}' ({context}).
 
 *Default severity: error.*
@@ -28,6 +34,8 @@ An object literal (or similar structured value) included a key that the target t
 
 **How to fix:** remove the stray key, fix a typo in the key name, or add the field to the target type if it belongs there.
 
+<a id="ag2004"></a>
+
 ## AG2004 — Variable '{name}' has no type annotation (strict mode).
 
 *Default severity: error.*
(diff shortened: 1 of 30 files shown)


egonSchiele:
Thanks — all four addressed in f8431cad0:

1. **`explain.ts` inherited-key lookup** — real bug: `agency explain toString` matched `Function.prototype.toString` via `in` and then crashed on `entry.code`. Switched to `Object.hasOwn(DIAGNOSTICS, q)` (same guard as `lib/optimize/registry.ts`), with a test pinning `toString`/`constructor`/`hasOwnProperty` → unknown + exit 1.
2. **VitePress anchor mismatch** — correct: the index linked `#ag####` but VitePress slugifies heading IDs from the full `## AG2001 — …` text. Now each code emits an explicit `<a id="ag####"></a>`, and a new generator test asserts every index link resolves to an anchor on its target page. Regenerated the category pages.
3. **`toBe(1)` on the index count** — tightened, but counting the link form `[AGxxxx](` rather than the raw code: the intro deliberately shows a real `agency explain AG2005` example, so a bare-substring count is legitimately 2. The invariant is one table-row link per code.
4. **`fs.rmSync` footgun** — replaced with `safeDeleteDirectory(outDir, false)` (project-root confinement), guarded by `existsSync` and failing fast if the delete is refused. Verified it refuses an existing outside-project dir and leaves it intact.

PR #525, opened 2026-07-12
typecheck and siblings accept directories and '-' stdin (#438)

### LINKED ISSUE #438: agency typecheck should accept directories (docs already promise it; currently crashes with EISDIR)
## Summary

`agency typecheck` (and the docs) claim you can pass a directory, but doing so crashes:

```
$ agency tc tests/agency/args
Error: EISDIR: illegal operation on a directory, read
```

The docs at `docs/site/guide/developer-tools.md` say typecheck accepts "a file, a directory, ... or a list of files and directories", and the page's own example is `agency tc src/` — which currently dies with EISDIR.

## Cause

The `typecheck` action iterates its `[inputs...]` and calls `readFile(input)` on each (`scripts/agency.ts:726-770`). `readFile` is a bare `fs.existsSync` + `fs.readFileSync` (`lib/cli/commands.ts:168-175`) with no directory handling.

## Proposal

When an input is a directory, recursively collect the `.agency` files under it and typecheck each, so all of these work:

```bash
agency tc src/
agency tc src/ lib/ extra.agency   # mixed list of dirs and files
```

Notes:

- There's already a recursive file finder used by the test runner (`findRecursively` in `lib/cli/test.ts` collects `.test.json` files) that could be generalized/reused for `.agency`.
- Keep the existing single-`SymbolTable` optimization: it's seeded from the first input and reused across files so reachable files (including stdlib) are crawled once.
- Exit code should still be 1 if any file has errors, and each file's errors should be reported under its path.
- Other commands share the same `readFile` helper and likely have the same limitation (`parse`/`ast`, `preprocess`, `fmt`); worth fixing them with the same helper, or at least checking while in here.

## Related docs

`docs/site/guide/developer-tools.md` also claims `-` reads from stdin; in reality stdin is only read when no arguments are passed (`-` fails with "Input file '-' not found"). Either support `-` in the same pass or fix the doc.

egonSchiele:
## Summary

Makes `agency typecheck`/`tc`/`check` and its siblings `ast`/`parse`, `preprocess`, and `diagnostics` accept **directory arguments** and a literal **`-` stdin argument**, instead of crashing with `EISDIR`. Closes #438.

The docs already promised `agency tc src/` but passing a directory crashed. These commands looped over raw args straight into `readFile()` (`fs.readFileSync`), with no directory handling, and a literal `-` fell through to `readFile("-")` → "not found".

## Approach

The codebase already owns directory walking (`findRecursively`, used by `fmt`/`doc`). This wires the five file-oriented commands through one shared resolver rather than reinventing anything.

- **`resolveInputSources(inputs)`** (in `lib/cli/commands.ts`) turns raw args into an ordered `InputSource[]` (`file` | `stdin`): expands directories via `findRecursively`, maps `-` → stdin, exits 1 on a missing path or a second stdin, and returns `null` (with a **stderr** notice) when args expand to zero `.agency` files so the caller exits cleanly instead of hanging on stdin.
- **`readSource(src)`** reads stdin or a file.
- **`forEachSource(inputs, handle)`** encapsulates the resolve → null-check → read → iterate scaffold. The four sibling commands each declare only their per-file work instead of duplicating the loop. The handler is `void | Promise<void>` and awaited.
- **typecheck** is the deliberate exception: it needs the whole resolved list up front to seed one `SymbolTable`, so it calls the resolver directly.

## SymbolTable seeding fix (completeness, not false positives)

typecheck previously seeded its shared `SymbolTable` from `inputs[0]` only. `SymbolTable.build` crawls imports transitively **from its entrypoint(s)**, so a file whose imports aren't reachable from the seed file can't resolve them.

The effect (verified against the live typechecker, and corrected from an earlier draft that called it a "false positive"): an unresolved agency import is **fail-open** — the imported name becomes `any`, not an "unknown symbol" error. So first-file-only seeding **silently misses** real cross-file type errors (e.g. a wrong-typed call to an imported function) in every file whose imports aren't reachable from the arbitrary first file, and drops those files' interrupt-effect metadata. For the primary `agency tc src/` use case that's a silent correctness hole.

Fix: seed from **every** file source (`SymbolTable.build` already accepts `string[]`). Strictly more complete (visited-set dedups, symbols stay file-keyed so no cross-file pollution) and free — the loop parses those files anyway.

## Tests

`tests/integration/cli-main/test.mjs` (no LLM calls): directory, mixed dir+file, `-` stdin, empty directory, a parse-on-directory case, and a **regression guard for the seeding fix** — two disjoint import pairs, each importer making a wrong-typed call. No single-file seed reaches both targets, so single-seed misses one error and multi-seed reports both; the test asserts both call-site errors appear. This fails on a revert to first-file-only seeding **independent of `findRecursively` traversal order** (confirmed by flipping the impl both ways). Plus unit tests for `resolveInputSources` and `forEachSource` in `lib/cli/commands.test.ts`.

Verified locally: `make` (tsc) clean, unit tests green, full `cli-main` integration harness exit 0, structural linter clean, and direct CLI exercise of every new path.

## Docs

Resolves the contradiction in `developer-tools.md` (it said directories were unsupported) and documents directory + `-` support in the `typecheck` CLI reference.


packages/agency-lang/tests/integration/cli-main/test.mjs
```diff
@@ -529,6 +550,82 @@ try {
...
+  writeFileSync(
+    join(tcImportDir, "consumer.agency"),
+    `import { greet } from "./helper.agency"
+
+node useGreet(): string {
+  return greet("there")
+}
+`,
+  );
+  const tcImportOut = runAgency("28a2-tc-dir-imports", ["tc", "tc-imports"]);
```
egonSchiele: This test is meant to guard the seed-from-every-file fix, but it can't fail under the old first-file-only seeding — so it doesn't actually protect against the regression it documents.

`findRecursively` yields files in `readdirSync` order, which here is `[consumer.agency, helper.agency]` (alphabetical). Under first-file-only seeding, `SymbolTable.build([consumer])` crawls consumer's `import { greet } from "./helper.agency"` and visits `helper` too, so `helper`'s symbols land in the table and `greet` resolves cleanly. `resolveImport(stmt, fromFile)` resolves the import *target* relative to `fromFile`, so the checked file itself needn't be in the table — only the target, and it always is here. Net: this passes under both old and new logic.

The false positive only occurs when a file's import target is unreachable from the first-visited file, which needs ≥3 files. For example:
- `aaa.agency` — standalone, imports nothing (sorts first → the sole seed under old logic)
- `bbb.agency` — `import { greet } from "./ccc.agency"`
- `ccc.agency` — defines `greet`

Under old logic, seeding from `aaa` alone leaves `ccc` unvisited, so typechecking `bbb` reports `greet` as unknown (exit 1); the new every-file seeding keeps it clean. Reshaping the fixture this way (or adding the standalone first file) makes the test one that would actually fail on a revert.

Quick confirmation: temporarily flip the impl back to `SymbolTable.build([filePaths[0]], config)` — this test still passes, which is the tell.

packages/agency-lang/scripts/agency.ts
```diff
@@ -809,8 +798,7 @@ export function createProgram(deps: CliDependencies = {}): Command {
     .description("Run diagnostics for VSCode")
     .argument("[inputs...]", "Paths to .agency input files")
     .action(async (inputs: string[]) => {
-      if (inputs.length === 0) {
-        const contents = await readStdin();
+      await forEachSource(inputs, (contents) => {
```
egonSchiele, line 801: Minor, machine-output contract: when this command is handed a directory containing no `.agency` files, the shared `resolveInputSources` prints `No .agency files found in the given input(s).` to **stdout**. `diagnostics` is the one command in this group whose stdout is machine-consumed (JSON for the VSCode extension), so a human notice mixed into the stream could trip a consumer that parses it.

In practice the extension passes a single file or stdin, not a directory, so this is low severity. But if you want `diagnostics` stdout to stay strictly JSON, it may be worth special-casing — e.g. route the no-files notice to stderr, or have `diagnostics` skip the shared notice. Not blocking.

egonSchiele:
Reviewed the diff end-to-end — overall clean, and it improves on the plan in two good ways:

- **Seeding the SymbolTable from every file source, not just the first.** This is the right call and fixes a real bug the plan had: independent files in a directory (no import path from the first file) would otherwise get false-positive unknown-symbol errors and missing interrupt-effect metadata. `SymbolTable.build` already accepts `string[]` with a shared visited/dedup set, so no API change and still one crawl. Nice catch.
- **`forEachSource`** is a tidy DRY scaffold for the three sibling commands.

Two inline notes below: (1) the `28a2-tc-dir-imports` test doesn't actually guard the seed-from-every-file fix (it passes under the old first-file logic too), and (2) a minor stdout-contract nit on `diagnostics`. Only (1) is worth acting on before merge.

packages/agency-lang/tests/integration/cli-main/test.mjs
egonSchiele: You're right that the old fixture didn't guard the fix — but when I built a fixture matching your 3-file shape and ran the revert-tell, it *still* passed under single-seed. That sent me to check the actual typechecker behavior, and the shared premise turns out to be wrong:

An unresolved cross-file agency import is **fail-open**, not an error. When `ccc` isn't in the table, `resolveImport` returns nothing and `greet` becomes `any` — so `typechecking bbb` reports **no error** (exit 0), not "greet unknown". I reproduced your exact `aaa`/`bbb`/`ccc` scenario under single-seed: 3× "No type errors found", exit 0.

So the real effect of first-file-only seeding isn't a false positive — it's a **silently missed** cross-file error (false negative). Demonstrated by making the call wrong-typed: `greet(123)` is caught under multi-seed (`AG6019`) and missed under single-seed.

New fixture (`28a2`) guards that, order-independently: **two disjoint import pairs**, each importer making a wrong-typed call (`aay→cee`, `bee→dee`). Since the pairs share nothing, `build([<any single file>])` reaches at most one target, so single-seed always misses one error; multi-seed reports both. The test asserts **both** call-site errors appear.

Verified both directions by flipping the impl:
- multi-seed → `in call to 'xFromCee'` **and** `in call to 'yFromDee'` (test passes)
- single-seed → only one of the two (test fails)

I've corrected the framing in the code comment, the plan, and the PR description to say "silently missed errors" rather than "false positives". Pushed in 9bf7e5c.

packages/agency-lang/scripts/agency.ts
egonSchiele: Routed the "no .agency files found" notice to stderr (in the shared `resolveInputSources`), so `diagnostics` stdout stays strictly JSON regardless of input. Applied universally rather than special-casing `diagnostics` — the notice is a status message, not program output, so stderr is the right stream for every command. Unit test now asserts `console.error`, and the `28d` integration case captures combined output. 9bf7e5c.

packages/agency-lang/lib/cli/commands.ts
```diff
@@ -130,6 +130,80 @@ export function readStdin(): Promise<string> {
...
+  handle: (contents: string, src: InputSource) => void,
+): Promise<void> {
+  const sources = resolveInputSources(inputs);
+  if (sources === null) {
+    return;
+  }
+  for (const src of sources) {
+    handle(await readSource(src), src);
+  }
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed. `forEachSource`'s handler is now typed `(contents, src) => void | Promise<void>` and awaited in the loop (`await handle(...)`). Current callers are all synchronous, but this closes the hole. Added a test asserting an async handler is awaited before the next source. 9bf7e5c.

packages/agency-lang/lib/cli/commands.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    expect(() => resolveInputSources(["-", "-"])).toThrow("exit");
+  });
+});
+
+describe("forEachSource", () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("hands each resolved source's contents to the handler in order", async () => {
```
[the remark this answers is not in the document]
egonSchiele: Fixed. `findRecursively` order isn't guaranteed, which is why the old test sorted — so it couldn't assert order. Rewrote it to pass explicit file arguments (`[zeta, alpha]`), whose order `resolveInputSources` preserves deterministically, and assert `[zeta.agency, alpha.agency]` without sorting. 9bf7e5c.

PR #526, opened 2026-07-12
Decision 8: propagate a tool's destructive work to the caller's frame

egonSchiele:
## Summary

Fixes the decision-8 frame-targeting bug found during the #522 review: destructive work performed by a tool **inside an `llm()` call** never propagated to the calling function.

`markDestructiveWork(stack)` in the tool loop marked `runPrompt`'s **own** frame (created by `setupFunction()` → `getNewState()`), which is discarded when `runPrompt` returns. So a function whose `llm()` call charged a card and then failed for another reason reported `destructiveRan: false` — losing the fact that money moved.

## Fix

The `llm()` codegen now threads the calling function's `__self` (its locals object) into `runPrompt` as `destructiveSink`, **by reference** — exactly the trick the existing `removedTools: __self.__removedTools` uses. Decision 8 marks `destructiveSink.__destructiveRan` instead of runPrompt's throwaway frame, so the caller's exit stamp reports it.

Properties that make this safe:
- **Sticky.** The flag is only ever set to `true`, never `false`. Two `llm()` calls in one function accumulate correctly, and concurrent async calls can't race.
- **Nesting is free.** Each `llm()` threads its own caller's `__self`, and each Result carries `destructiveRan`, so agent-as-tool chains propagate automatically.
- **Blocks work.** Inside a block `__self` is the block's locals, so the existing block-join fold carries the mark up to the function.

## Tests (`tool-tiers`)

- A destructive tool that **succeeds** inside `llm()` → the caller's later failure is `destructiveRan: true` (the core proof).
- A destructive tool that **cleanly refuses** (`destructiveRan: false`) does **not** propagate → caller stays `false` (the gating from #522 still holds).
- `llm()` **inside a block** → propagates via the block-join fold.

Mutation-verified: reverting to `markDestructiveWork(stack)` flips the success case to `false` and fails the test.

## Notes

- Every `llm()` call site now emits `destructiveSink: __self`; regenerated the 33 codegen fixtures.
- Full `lib/` vitest green (8068 passed); lint clean.
- The codegen-extraction / TSIR-builder refactor (the other half of the follow-up) is **not** in this PR — this is a focused correctness fix, kept separate for review.


packages/agency-lang/tests/agency-js/tool-tiers/agent.js
```diff
@@ -561,6 +561,7 @@ graph.node("neutralStaysCallable", async (__state) => {
           },
           maxToolCallRounds: 10,
           removedTools: __self.__removedTools,
+          destructiveSink: __self,
```
egonSchiele: Should this file be checked in? This just looks like the JavaScript code that was generated out of agency code. I don't think it needs to be checked in.

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -768,6 +768,13 @@ export async function runPrompt(args: {
   retryConfig?: RetryConfig;
   maxToolCallRounds?: number;
   removedTools?: string[];
+  /** The CALLING function's `__self` (its locals object), threaded in by the
+   *  `llm()` codegen as `destructiveSink: __self`. When a tool does
+   *  destructive work, decision 8 marks `destructiveSink.__destructiveRan`
+   *  so the caller's exit stamp reports it. Passed by reference — the same
+   *  by-reference trick `removedTools` uses. Absent for direct TS callers
+   *  (e.g. `agency.llm`), so the mark is a no-op there. */
+  destructiveSink?: Record<string, any>;
```
egonSchiele: I thought this would be a boolean, so I'm curious to hear why it's a record. What are the keys and what are the values here?

egonSchiele (commented):
## Review — decision-8 frame-targeting fix

This is the follow-up to the bug flagged during #522, and it's a clean, correctly-scoped fix. I verified the root cause and the fix end-to-end rather than taking the description on faith:

**Root cause confirmed.** In `runPrompt`, `const { stateStack, stack } = setupFunction(); const self = stack.locals` — so `stack` is `runPrompt`'s *own* frame, discarded on return. The old `markDestructiveWork(stack)` wrote `self.__destructiveRan` on that throwaway frame, so the caller's exit stamp never saw it. The #522 inline comment that claimed "`stack` is the calling function's activation frame" was simply wrong; this PR corrects both the code and that comment.

**Fix is right.**
- `ts.runtime.self` resolves to the `__self` identifier (`lib/ir/builders.ts:848`), i.e. the *calling* function's locals — threaded in as `destructiveSink: __self`, by reference, exactly like the adjacent `removedTools: __self.__removedTools`. Same call site, so the `__self`-in-scope guarantee is inherited for free.
- Reading `args.destructiveSink` directly (not `self.destructiveSink`) is the correct choice and worth the explicit comment — the sink belongs to the caller, whereas `self.*` is runPrompt's frame. Mixing those up is exactly what caused the original bug.
- `markDestructiveWork({ locals: destructiveSink })` mutates the caller's `__self.__destructiveRan` in place; sticky (only ever set true) so repeated and concurrent async `llm()` calls in one function accumulate without a race.
- The `&& destructiveSink` guard keeps direct TS callers (`agency.llm`, agent turns) a safe no-op — correct, since they have no agency frame to propagate into.

**Tests have teeth.** `fixture.json` pins the three cases to `true` / `false` / `true`: success propagates, a clean refusal (`destructiveRan === false`) does *not* (the #522 gating still holds), and the block case exercises the block-join fold. The refusal case is the important guard — it proves the fix didn't over-correct back into caller poisoning. Mutation-verifying against `markDestructiveWork(stack)` is the right check.

### Minor (non-blocking)
- **Stale `__result.json`.** The runner (`lib/cli/test.ts:1044+`) overwrites `__result.json` and diffs it against `fixture.json`, so correctness is fine — but `__result.json` is git-tracked and still holds the old 2-key content while `test.js`/`fixture.json` now produce 5. A local `pnpm run a test js` leaves it dirty. Consider regenerating it in the PR or gitignoring it (it's a generated artifact).
- **Design observation, not a change request:** propagation is intentionally asymmetric — a *non*-destructive function that does destructive work internally and then *succeeds* won't propagate (a `ResultSuccess` carries no `destructiveRan` bit; only the marker is consulted on success), whereas on failure it would via the exit-stamp fold. This matches #522's marker-driven model (mark your destructive functions `destructive` and it all works), so nothing to do here — just flagging that "success hides un-marked internal destructive work" is a real edge of the model.

LGTM — only the `__result.json` hygiene nit is worth addressing before merge.

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: Also, does this change work fine with agency's concurrency and state isolation features? What about its interrupts feature that lets users pause and resume code?
https://agency-lang.com/guide/concurrency.html
https://agency-lang.com/guide/interrupts-part-2.html

packages/agency-lang/tests/agency-js/tool-tiers/agent.js
egonSchiele: You're right — it's the JS compiled from agent.agency, and the runner regenerates it, so it doesn't belong in git. It only got tracked because a `git add -A` in #522 swept it up; the other 127 `agent.js` files under tests/agency-js were never committed. Untracked it and added a gitignore rule so it can't recur. Also untracked the generated `__result.json` (it's the actual output test.js writes and diffs against fixture.json; already covered by the `*__result.json` rule).

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: It can't be a boolean — the whole point is that runPrompt writes it and the CALLER reads it back, which needs a shared object passed by reference (a boolean would be copied by value). So the codegen passes the caller's `__self`, its locals object: keys are the function's local variables plus bookkeeping slots like `__destructiveRan` and `__removedTools`, values are whatever those hold.

But runPrompt only ever touches ONE of those keys, `__destructiveRan`. So I've narrowed the parameter type from `Record<string, any>` to `{ __destructiveRan?: boolean }`. The codegen still passes the whole `__self` (structurally compatible), but the signature now advertises exactly the one field runPrompt writes, so a reader doesn't have to wonder what it does with the rest.

egonSchiele: Good question — I checked all three, and added a test for the riskiest (interrupts).

**Interrupts (pause/resume).** The mark lives in the calling frame's locals, and locals are deep-cloned wholesale into the checkpoint with no `__`-key filter (stateStack.ts `toJSON`). So a mark made before a pause is captured in the checkpoint and restored on resume. New test `destructiveSurvivesInterrupt`: a destructive tool marks the caller, the flow pauses on an `interrupt(...)`, resumes on approve, then fails — and the failure is still `destructiveRan: true`. It passes.

**Concurrency / state isolation (fork, parallel).** Each function and each fork branch threads its OWN `__self` as the sink — inside a branch/block `__self` is that branch's locals, not the parent's. So two concurrent branches mark two different objects; they never cross-contaminate. The flag is sticky and per-frame, and a branch's result carries `destructiveRan` up through the fork-join fold, exactly like the block case. The `callerBlockPoisoned` test already exercises that scoped-sub-context fold (llm() inside a block → block-join carries the mark to the function). I didn't add a two-branch fork test because the deterministic LLM mock is a shared queue that concurrent branches consume in nondeterministic order, which would make such a test flaky; the isolation follows structurally from each branch passing its own `__self`.

PR #527, opened 2026-07-12
Normalize undefined→null at index and match value sites (#409)

### LINKED ISSUE #409: object lookup with missing key should yield null (not undefined)
## Summary

Indexing an object with a key that is not present currently yields `undefined`:

```ts
const val = obj[key]   // key not in obj  ->  val === undefined
```

But `undefined` does not exist as a value in Agency — the language only has `null`. A missing-key lookup should produce `null` to stay consistent with the rest of the language.

## Expected

```ts
const val = obj[key]   // key not in obj
// current:  val === undefined
// expected: val === null
```

## Notes

- Companion to the `match` issue: a `match` expression that matches no arm also currently yields `undefined` and should yield `null`.
- Worth auditing other places `undefined` can leak into Agency values (e.g. array out-of-bounds indexing, destructuring missing fields) so the "only `null` exists" invariant holds everywhere.

egonSchiele:
## Summary

Fixes #409. Missing object-key lookups (`obj[key]`), out-of-bounds array indexing (`arr[i]`), and unmatched `match` expressions now yield `null` instead of `undefined`, completing the **value** side of Agency's "one nothing-value" invariant.

The language already committed to absorbing `undefined` into `null` (see `docs/dev/null-and-undefined.md`), but only the **comparison** side was finished — `__eq` makes `null == undefined`. The **value** side was unaddressed: a missing lookup produced a literal `undefined` that then flowed through the program (printed, interpolated, stored in threads, serialized) as `undefined`. This PR closes that gap for the two most common leak sites.

## Approach

- **`__nn` runtime helper** (`x ?? null`) — a nullish-normalize helper wired exactly like the existing `__eq` (`lib/runtime/nn.ts` → exported from `runtime/index.ts` → added to the generated-code import list via `imports.mustache` → emitted as `__nn(...)`).
- **Terminal index reads** — `processValueAccess` wraps the completed access chain in `__nn` when its last element is an index. Wrapping *terminally* (not per-element) preserves JS optional-chain short-circuit: `a?.[b].c` with null `a` stays `undefined` rather than throwing `null.c`. Covers `obj[key]`, `arr[i]`, and `arr[-1]` in one site.
- **Match results** — the single `__matchval_<id>` read chokepoint is wrapped in `__nn`, covering both the stepped no-arm path and the plain-mode IIFE that returns `undefined`. The valueless-`yield` fallback is also changed from `undefined` to `null` at the source.

## Beyond the plan: lvalue targets

`processValueAccess` turned out to also emit assignment/update **targets** (`x[i]++`, `x[i] += v`) via `processBinOpExpression` — a route the design missed. Naively wrapping produced invalid lvalues (`__nn(x[i]) = ...`) and broke the stdlib build. Fixed with an `asLValue` flag + `processAssignTarget` helper so targets stay raw lvalues while value reads are normalized. Verified in generated output: `arr[i] += 10` / `arr[i]++` stay raw; `arr[i] + 1` (a read) becomes `__nn(arr[i]) + 1`.

## Non-goals

- No type-checker changes — index/match result *types* are unchanged. Forcing narrowing on nullable access is "Project 2" strict null checking. Since the value at these sites was already `undefined` (equally nullish), there is no regression and no new "possibly null" errors.
- Deferred leak sites: optional chaining, destructuring a missing field, falling off a function without `return`, and TypeScript interop.

## Coercion caveat (intended)

`null` and `undefined` coerce differently in arithmetic (`undefined + 1` → `NaN`; `null + 1` → `1`) and string contexts (`` `${undefined}` `` → `"undefined"`; `` `${null}` `` → `"null"`). The string-context change is the fix the issue wants; the arithmetic change is a deliberate consequence of collapsing to one nothing-value, consistent with "don't do math on a missing key."

## Testing

- `lib/runtime/nn.test.ts` — unit test for `__nn`, incl. falsy passthrough (`0`/`""`/`false`/`NaN`).
- `tests/agency-js/index-missing-key-null` — missing key, out-of-bounds, negative index → `null`; falsy/present values preserved; plus `optChainNoThrow` (`a?.["x"]["y"]` with null `a`) which CI-fails if the wrap ever regresses to per-element.
- `tests/agency-js/match-no-arm-null` — no-arm match → `null`; matched arm unchanged; explicit `null` arm stays `null` (no double-coercion). Uses an open scrutinee type, so it compiles at the default `matchExhaustiveness: error` with no override.
- Full lib unit suite (8121 tests) and codegen integration fixtures pass; `null-and-undefined.md` updated.

### DIFF
--- a/docs/superpowers/plans/2026-07-11-index-match-null-normalization.md
+++ b/docs/superpowers/plans/2026-07-11-index-match-null-normalization.md
@@ -1,0 +1,451 @@
+# Index/Match `undefined`→`null` Normalization Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make missing object-key lookups (`obj[key]`), out-of-bounds array indexing (`arr[i]`), and unmatched `match` expressions yield `null` instead of `undefined`, completing the value-side of Agency's "one nothing-value" invariant.
+
+**Architecture:** Add a tiny `__nn(x) => x ?? null` runtime helper, wired exactly like the existing `__eq` helper. Emit `__nn(...)` around two codegen sites in `typescriptBuilder.ts`: the index access chain element, and the `__matchval_<id>` read that consumes a match result. No type-checker changes — this is a runtime-value fix only.
+
+**Tech Stack:** TypeScript (compiler + runtime), typestache templates (`pnpm run templates`), vitest (unit + integration tests), Agency execution tests under `tests/agency-js/`.
+
+**Spec:** `docs/superpowers/specs/2026-07-11-index-match-null-normalization-design.md`
+**Issue:** https://github.com/egonSchiele/agency-lang/issues/409
+
+## Global Constraints
+
+- All paths below are relative to `packages/agency-lang/` unless absolute.
+- **Never edit generated `.ts` templates.** `lib/templates/.../imports.ts` is auto-generated from `imports.mustache`; edit the `.mustache` and run `pnpm run templates`.
+- **Rebuild with `make`, not `pnpm run build`** before running Agency/agency-js tests (`pnpm run build` does not copy everything into `dist`).
+- **No dynamic imports. Use `type` not `interface`. Use arrays not sets, objects not maps.**
+- **Agency syntax:** `def foo(x: T): R { ... }`, `node main() { ... }`, `if (cond) { ... }`, declare with `let`/`const`. Verify snippets against `docs/site/guide/basic-syntax.md`.
+- **Do not run the full agency test suite locally** (slow/expensive). Run only the specific new agency-js tests named in this plan, and save output to a file.
+- Work happens on branch `undefined-to-null-indexing` (already created). Do not commit to `main`.
+
+---
+
+### Task 1: `__nn` runtime helper + wiring
+
+Add the nullish-normalize helper and make it importable in generated code. Self-contained and unit-testable before any codegen touches it.
+
+**Files:**
+- Create: `lib/runtime/nn.ts`
+- Create: `lib/runtime/nn.test.ts`
+- Modify: `lib/runtime/index.ts:151` (add export beside `__eq`)
+- Modify: `lib/templates/backends/typescriptGenerator/imports.mustache:25` (add `__nn` to the import list)
+
+**Interfaces:**
+- Produces: `export function __nn<T>(x: T): T | null` — returns `null` for `null`/`undefined`, `x` unchanged otherwise. Later tasks emit it as `ts.call(ts.id("__nn"), [expr])`.
+
+- [ ] **Step 1: Write the failing unit test**
+
+Create `lib/runtime/nn.test.ts`:
+
+```ts
+import { describe, it, expect } from "vitest";
+import { __nn } from "./nn.js";
+
+describe("__nn", () => {
+  it("collapses undefined and null to null", () => {
+    expect(__nn(undefined)).toBe(null);
+    expect(__nn(null)).toBe(null);
+  });
+
+  it("passes non-nullish values through unchanged, including falsy ones", () => {
+    expect(__nn(0)).toBe(0);
+    expect(__nn("")).toBe("");
+    expect(__nn(false)).toBe(false);
+    expect(__nn(5)).toBe(5);
+    expect(__nn("a")).toBe("a");
+    expect(Number.isNaN(__nn(NaN))).toBe(true);
+  });
+
+  it("returns the same object reference for objects", () => {
+    const obj = { x: 1 };
+    expect(__nn(obj)).toBe(obj);
+  });
+});
+```
+
+- [ ] **Step 2: Run the test to verify it fails**
+
+Run: `pnpm exec vitest run lib/runtime/nn.test.ts 2>&1 | tee /tmp/nn-test.log`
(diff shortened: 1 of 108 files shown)


egonSchiele:
## Review — `undefined`→`null` at index/match value sites

Read the full diff (source + a sample of the 89 regenerated fixtures + docs). Overall this is a clean, well-scoped implementation that faithfully follows the plan and correctly handles the optional-chain regression the plan flagged. The `asLValue` extension for update targets is a good catch. A couple of things worth addressing before merge — one correctness-consistency wart and one real test gap.

### What works well
- **`__nn` mirrors `__eq`** exactly (helper → `runtime/index.ts` → `imports.mustache` → `__nn(...)` emission). Consistent, greppable, well-documented. Good altitude — the "how" lives in one place.
- **Terminal-only wrapping is implemented correctly** (`processValueAccess`, ~L1185). The comment explaining *why* not per-element is excellent, and `optChainNoThrow` (`a?.["x"]["y"]` with null `a`) is a real CI guard that fails if anyone regresses to per-element wrapping.
- **Match chokepoint** (~L1034): confirmed this single read site covers both the stepped no-arm path and the plain-mode IIFE (which writes the same `__matchval_<id>` the consumer reads). Correct.
- Strong negative controls in the index test (`falsyZero === 0`, `presentKey`, `inBounds`) catch over-coercion / a `||`-instead-of-`??` mistake. Docs update is accurate.

### Issue 1 — compound-assign read is inconsistent with a plain read (minor, correctness)
For a non-global compound assignment the left stays a raw lvalue and emits native `x[i] += v` (`processBinOpExpression`, ~L1256/L1275). That's correct for the *target*, but `x[i] += v` also *reads* `x[i]`, and that read is now **not** normalized — while a syntactically-plain read of the same slot **is**:

```
arr["missing"] + 1     // → __nn(arr["missing"]) + 1  → null + 1 → 1
arr["missing"] += 1    // → arr["missing"] += 1       → undefined + 1 → NaN
```

So two logically-identical operations diverge, and only because of the syntax. Before this PR they agreed (both `NaN`); after, they don't. It falls inside the documented "coercion caveat," and compound-assigning into a missing key is an edge case, so this isn't a blocker — but the divergence is silent. Two options:
- Lower non-global `x[i] <op>= v` to `x[i] = __nn(x[i]) <op> v` so the read side is normalized like every other read (keeps the two forms consistent), or
- Explicitly note in the doc / a code comment that the *read half* of a compound-assign target is intentionally left raw.

### Issue 2 — the lvalue fix has no runtime test (medium, coverage)
The `asLValue` / `processAssignTarget` path is the newest and riskiest part of this PR (it "broke the stdlib build" mid-development), yet it's only *"verified in generated output"* by hand. I checked: no agency-js test and no regenerated fixture exercises `arr[i] += v` / `arr[i]++` / `obj[k] -= 1`. The only thing standing between a future refactor (dropping the flag, adding a new compound operator, a new lvalue route) and a silent `__nn(x[i]) = ...` invalid-lvalue break is the implicit `make` stdlib build.

Please add a small `tests/agency-js` case that mutates through an index — e.g. `arr[i] += 10`, `arr[i]++`, `obj[k] -= 1` — and asserts the results. That both locks in the lvalue behavior and pins the accepted answer for the missing-key compound case from Issue 1.

### Minor
- The planning + spec docs under `docs/superpowers/` are included in this PR — intentional? If those are meant to be repo history that's fine; just flagging in case they were meant to stay local.

Nothing here is a hard blocker except your own call on Issue 2. Nice work on the terminal-only reasoning and the lvalue save.

egonSchiele:
Pushed two fixes for the CI failures:

**1. `build` job** — I'd regenerated `tests/typescriptGenerator/` fixtures (via `make fixtures`) but missed the separate `tests/typescriptBuilder/` set (`assignment.mjs`, `simple.mjs`), which `typescriptBuilder.integration.test.ts` / `jsonSchemaAnnotation.test.ts` compare against. Committed the missing `__nn` import-line update.

**2. `agency-tests` job** — one real regression: `tests/agency/nullish-operators` `testOptionalIndex`. Wrapping *optional* terminal indexes changed `arr?.[i]` on a null base from `undefined` to `null`. Optional chaining is explicitly deferred/out-of-scope here, and `?.` property access on null stays `undefined`, so `?.[]` should too. Now the wrap fires only for a **non-optional** terminal index (`!lastElement.optional`) — #409's actual target. Added an `optNull` guard to the index test and documented the exclusion in `null-and-undefined.md`.

Verified locally: full lib suite (8121 tests) green, both new agency-js tests pass, and `nullish-operators` back to 8/8.

egonSchiele:
Second round of CI fixes (the earlier fixture fix unblocked `test:run`, which then let two later build-job steps run and surface real issues):

**`agency-tests`** — one over-reach, now fixed. Wrapping *optional* terminal indexes turned `arr?.[i]` on a null base from `undefined` into `null`, breaking `tests/agency/nullish-operators` (`testOptionalIndex`). Optional chaining is deferred/out-of-scope, and `?.` property access on null stays `undefined`, so `?.[]` must too. The wrap now fires only for a **non-optional** terminal index.

**Built-in agent tests** (`build` job, both 22.x/23.x) — three expected outputs captured the old wart: a missing-key **index** read interpolated into a string produced the literal `"undefined"`. Those reads now yield `null` (the whole point of #409), so I updated the expectations (`"undefined/undefined"` → `"null/null"`, etc.). Verified these come from *terminal index* reads; sibling cases like `models["m1"].maxTokens` (terminal *property*) correctly stay `undefined` and were left untouched.

For the record, `stdlib-sandbox` wikipedia/weather/search are `runAllowFail` (live-API, non-blocking) — not part of this change.

Verified locally: `pnpm test:run` (8121), statelog (8/8), stdlib-sandbox fs/shell/pure/date/policy/strategy, and the full built-in agent suite all green.

PR #529, opened 2026-07-12
Error on Agency imports that don't resolve (AG4008/AG4009)

egonSchiele:
## Summary

Makes the type checker **error** when a plain Agency import names something that doesn't exist. Follow-up to #438 / PR #525.

Two mistakes were silent before and are now hard errors:

- **`AG4008`** — the target Agency file exists and is loaded, but doesn't define the imported name.
- **`AG4009`** — the import path resolves to no file.

```
// lib.agency has no `missingFn`
import { missingFn } from "./lib.agency"   // AG4008
import { x } from "./ghost.agency"         // AG4009
```

Previously both type-checked clean. An imported name is declared into scope as `any` unconditionally (`scopes.ts`), so nothing verified the target actually exports it.

## Why this is safe to make a hard error

The check only fires when it is **certain**. The symbol table records which files it loaded, so:

- module path doesn't exist on disk → error (missing module)
- module loaded, name absent → error (missing name)
- module exists on disk but wasn't loaded (a parse failure in it, or a single-file editor check) → **silent** — the view is incomplete, and the target's own error is the real one

This asymmetry with undefined *calls* (a silenceable warning) is deliberate: a bare `foo()` might be an uncatalogued JavaScript global, which Agency permits without an import, so the checker can't be sure. An Agency import is unambiguous.

## Scope

- Covers `import { ... }` (Agency only) and `import node { ... }`, including aliases.
- Skips JavaScript imports (can't read a `.js` file's exports).
- `export { } from` and unresolvable `pkg::` already throw in `SymbolTable.build`, so they're out of scope here (documented as a possible follow-up to convert those throws into clean diagnostics).
- Non-goal: enforcing `export` visibility on plain imports.

## Implementation

- `SymbolTable.resolveImportModule(modulePath, fromFile, config?)` — classifies the target as `missing` / `notLoaded` / `loaded`, hiding path resolution, the `fs` check, and the `pkg::` throw behind a tagged union. Verbose-gated logging in the catch so an unexpected resolution error isn't silently mislabelled.
- `checkMissingImports(ctx)` — a new typechecker pass beside `checkUndefinedFunctions`, using a small `toImportSpec` normalizer so it doesn't care which node kind an import came from.
- Two registry diagnostics (AG4008/AG4009) plus their required `agency explain` entries.

## Rollout sweep

Ran the check across the whole codebase before shipping the hard error:

- `agency tc stdlib` → **0** import errors.
- `agency tc tests` → only two fixtures with intentionally-dangling imports, both intentional and not type-checked by CI (`typescriptGenerator/imports.agency`, an import-syntax fixture now excluded from the fixtureTypeCheck vitest; `formatter/roundtrip.agency`, only formatted, never type-checked).

No real, stdlib, or production import needed fixing.

## Tests

- Unit tests for the classifier (loaded / missing / notLoaded / pkg-throws).
- 12 pass tests: missing name, aliased, missing node, missing module, silent-on-parse-failure, valid cross-file, **JS import skipped**, **valid node import**, mixed valid+invalid (one error), multi-name missing module (one error), re-exported name, valid `std::`.
- cli-main integration case `28e-tc-missing-import` (expectFail, asserts AG4008), proving `agency tc` exits non-zero.

Verified locally: full typechecker vitest suite (1318 pass), the two new unit files (16 pass), the cli-main integration harness (exit 0), structural linter clean, and direct CLI exercise of AG4008 + AG4009 end-to-end.

## Process docs

The design spec, implementation plan, and both external reviews are included under `docs/superpowers/`.

### DIFF
--- a/packages/agency-lang/docs/site/cli/typecheck.md
+++ b/packages/agency-lang/docs/site/cli/typecheck.md
@@ -19,14 +19,3 @@ You can also pass a literal `-` to read from stdin explicitly, and mix it with f
 ## Options
 
 - `--strict` — enable strict mode. In strict mode, untyped variables are errors rather than being inferred. Use this if you want every variable to have an explicit type annotation.
-
-## Imports must resolve
-
-The type checker verifies that every Agency import names something real. Two mistakes are hard errors:
-
-- `AG4008` — the target file exists but does not define the imported name (for example, `import { missingFn } from "./lib.agency"` when `lib.agency` has no `missingFn`).
-- `AG4009` — the import path resolves to no file (for example, a typo in `import { x } from "./libb.agency"`).
-
-Unlike a call to an undefined function (a warning that might be an uncatalogued JavaScript global), an unresolved Agency import is unambiguous, so it always errors.
-
-Import checking needs a file path to resolve relative imports against, so it is skipped when input comes from stdin without a path (`agency tc -`). Pass files or a directory to have imports checked.
--- a/packages/agency-lang/docs/site/cli/typecheck.md
+++ b/packages/agency-lang/docs/site/cli/typecheck.md
@@ -19,3 +19,14 @@ You can also pass a literal `-` to read from stdin explicitly, and mix it with f
 ## Options
 
 - `--strict` — enable strict mode. In strict mode, untyped variables are errors rather than being inferred. Use this if you want every variable to have an explicit type annotation.
+
+## Imports must resolve
+
+The type checker verifies that every Agency import names something real. Two mistakes are hard errors:
+
+- `AG4008` — the target file exists but does not define the imported name (for example, `import { missingFn } from "./lib.agency"` when `lib.agency` has no `missingFn`).
+- `AG4009` — the import path resolves to no file (for example, a typo in `import { x } from "./libb.agency"`).
+
+Unlike a call to an undefined function (a warning that might be an uncatalogued JavaScript global), an unresolved Agency import is unambiguous, so it always errors.
+
+Import checking needs a file path to resolve relative imports against, so it is skipped when input comes from stdin without a path (`agency tc -`). Pass files or a directory to have imports checked.
--- a/packages/agency-lang/lib/typeChecker/fixtureTypeCheck.integration.test.ts
+++ b/packages/agency-lang/lib/typeChecker/fixtureTypeCheck.integration.test.ts
@@ -17,6 +17,9 @@ const SKIP_TYPECHECK = new Set([
   "blockParams",        // calls append() which doesn't exist (should use .push())
   "ifElse",             // calls undefined isReady(), uses undeclared variables
   "setLLMClient",       // setLLMClient() is runtime-injected, not visible to typechecker
+  "imports",            // exercises import SYNTAX/codegen with intentionally-dangling
+                        // modules (./foo.agency etc. don't exist); strict imports (AG4009)
+                        // now flags them, which is correct for real code but not this fixture
 ]);
 
 // Stdlib files with known warnings from typechecker limitations
--- a/docs/superpowers/plans/2026-07-12-strict-imports-check-review.md
+++ b/docs/superpowers/plans/2026-07-12-strict-imports-check-review.md
@@ -1,0 +1,180 @@
+# Review: strict-imports-check implementation plan
+
+**Reviews:** `docs/superpowers/plans/2026-07-12-strict-imports-check.md`
+
+**Date:** 2026-07-12
+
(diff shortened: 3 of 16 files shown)


packages/agency-lang/lib/typeChecker/fixtureTypeCheck.integration.test.ts
```diff
@@ -17,6 +17,9 @@ const SKIP_TYPECHECK = new Set([
   "blockParams",        // calls append() which doesn't exist (should use .push())
   "ifElse",             // calls undefined isReady(), uses undeclared variables
   "setLLMClient",       // setLLMClient() is runtime-injected, not visible to typechecker
+  "imports",            // exercises import SYNTAX/codegen with intentionally-dangling
```
egonSchiele, line 20: Verified this skip is the right call — `imports.agency` is a pure import-*syntax*/codegen fixture with intentionally non-existent targets, so AG4009 flagging it is correct-for-real-code but wrong for this fixture, and it's still exercised by the typescriptGenerator codegen tests. Not blocking.

One thing to keep in mind: adding it to `SKIP_TYPECHECK` suppresses *all* type errors for this fixture, not just AG4009 — so if `imports.agency` ever grows a genuine type error, it'd now go unnoticed. Same tradeoff as the other five entries, and the risk is negligible here (the fixture has no real logic to type-check), so this is fine as-is. Only flagging in case a tighter option (e.g. asserting the fixture produces *exactly* AG4009s and nothing else) is ever worth it.

egonSchiele:
Reviewed end-to-end against the code (fetched the branch, read the pass, the classifier, the diagnostics, the fixture change, and the tests). This is a clean, faithful implementation, and it closed every finding from the earlier spec+plan reviews. What I independently verified:

- **`resolveImportModule`** — the empty-catch finding is fixed exactly right: the error is bound and logged under `config.verbose`, so an *unexpected* resolution fault isn't silently mislabelled "Cannot find module," while the expected `pkg::` throw stays quiet. `ctx.config` is correctly threaded from the pass (`types.ts:119`).
- **`checkMissingImports`** — the imperative name-collection loop and the two duplicated branches are gone; `toImportSpec` is a clean WHAT/HOW split, and the name collection is now `.filter().flatMap()`. One module error per statement, per-name errors when loaded — correct.
- **Rollout sweep** — broadened to the whole tree (`tc stdlib` = 0, `tc tests` = 2 intentional). The one code change it forced — adding `imports` to `SKIP_TYPECHECK` — is legitimate: `tests/typescriptGenerator/imports.agency` is 11 lines of pure import *syntax* with deliberately non-existent targets, never meant to type-check clean, and it's still exercised by the codegen tests. Consistent with the 5 existing entries.
- **Tests** — all six previously-missing cases were added, and the two subtle ones are non-incidental: the re-export test imports a name only *re-exported* by the barrel (so it genuinely guards `mergeExportsFrom`), and the JS-skip test guards the load-bearing `isAgencyImport` guard. The two previously-weak tests now have positive counterparts (a valid `import node` guards the node-symbol lookup that the missing-node test alone couldn't).
- **Diagnostics** — AG4008/AG4009 codes are unused elsewhere, severity `error`, and both have `agency explain` entries.

No correctness defects found in the pass, classifier, diagnostics, or wiring. One low, non-blocking note left inline on the `SKIP_TYPECHECK` change. Nice work.

packages/agency-lang/docs/site/cli/typecheck.md
```diff
@@ -19,3 +19,14 @@ You can also pass a literal `-` to read from stdin explicitly, and mix it with f
 ## Options
 
 - `--strict` — enable strict mode. In strict mode, untyped variables are errors rather than being inferred. Use this if you want every variable to have an explicit type annotation.
+
```
egonSchiele: undo all changes to this file

egonSchiele: Done — reverted all changes to `typecheck.md` in ea484987e (file now matches main exactly). AG4008/AG4009 stay documented via the generated `docs/site/diagnostics/` pages and `agency explain AG4008`/`AG4009`.

packages/agency-lang/lib/typeChecker/fixtureTypeCheck.integration.test.ts
egonSchiele: Agreed on the tradeoff. I left it as the broad SKIP_TYPECHECK entry to match the existing five and because the fixture has no real logic to lose coverage on. The tighter "assert exactly AG4009s and nothing else" option would need a per-fixture expected-errors mechanism the harness doesn't have today (it only asserts zero errors for non-skipped fixtures), so it's a harness change rather than a one-liner — not worth it for this fixture. Happy to add that mechanism separately if it's ever wanted for a fixture with real logic.

PR #530, opened 2026-07-12
Retire the "any" string sentinel in favor of ANY_T (#472)

### LINKED ISSUE #472: Retire the "any" string sentinel in favor of ANY_T
The typechecker represents "unknown type" two different ways:

1. The string sentinel `"any"` — `VariableType | "any"` appears in nearly every signature in `lib/typeChecker/`.
2. `primitiveType("any")` — a real `VariableType` node (`ANY_T` in `lib/typeChecker/primitives.ts`).

`isAssignable` has to special-case both, back to back (`lib/typeChecker/assignability.ts` — the sentinel short-circuit at the top, then the `primitiveType("any")` check right after). Every dual representation of one concept is a standing bug generator.

The sentinel also taxes every consumer:

- `t !== "any" && t.type === ...` guards at every use site.
- `maybeAny()` conversions wherever a synth result gets embedded inside a structured type (e.g. as `ResultType.successType`).
- Awkward narrowing dances like `synthPipeRhs` in `lib/typeChecker/synthesizer.ts`: `ctx.inferredReturnTypes[name] !== "any" ? (... as VariableType | undefined) : undefined`.

## Proposed work

Retire the string sentinel in favor of `ANY_T`. `isAnyType` (`lib/typeChecker/utils.ts`) already abstracts the check in some places; the refactor finishes that thought:

- Change signatures from `VariableType | "any"` to `VariableType`, module by module, starting with the innermost helpers (`synthType` internals, `flow.ts`'s `ScopeType`) and working outward.
- Replace sentinel comparisons with `isAnyType(t)`.
- Delete `maybeAny()` once nothing produces the sentinel.

Mechanical but wide; it can land incrementally since `isAnyType` can accept both representations during the transition.

egonSchiele:
Closes #472. Retires the `"any"` string sentinel from the type checker: "unknown type" is now represented only by the `ANY_T` object, never the bare string `"any"`.

## Why

The checker represented "I don't know this type" two ways — the string `"any"` (in `VariableType | "any"` signatures) and the `ANY_T` object (`primitiveType("any")`). Every dual representation is a standing bug generator: comparisons asked the question twice (`t === "any" || isAnyType(t)`), `maybeAny()` existed only to bridge the two, and `isAssignable` special-cased `any` twice back to back. This deletes the string form so one shape flows through everything, and — once signatures are narrowed to `VariableType` — the compiler makes the sentinel unrepresentable.

## What changed

Behavior-preserving throughout. Delivered as an ordered stack under one invariant: **a producer's returns and its signature narrowing land together**, so a leftover `=== "any"` becomes a TS2367 compile error rather than silently going false.

1. **`isAnyType` widened** to a type predicate accepting both forms during the migration (so every comparison could route through one helper and still narrow).
2. **Every comparison** (65 sites) routed through `isAnyType` — behavior-preserving, closing the silent-break window before any producer changed.
3. **Spine + all producers narrowed** to `VariableType`: `synthType`, `ScopeType`, `inferredReturnTypes`, `uniteTypes`, `typeAt`, and 35 `return "any"` → `return ANY_T`. The `| null` / `| undefined` signal sites kept their extra members. `widenType` fully cleaned up. LSP consumers (`builtinHover`, `inlayHint`, `typeResolution`) route their now-dead string comparisons through `isAnyType`.
4. **Object-form folds + assignability cleanup**: the six inline `.value === "any"` checks fold into `isAnyType`; the redundant pre-resolution `any` fast path in `isAssignableGuarded` is deleted (the post-resolution check is the only one that catches `type Foo = any`).
5. **Scaffolding deleted**: `maybeAny`, `inference.ts`'s local `ANY_T`/`isAnyType` duplicates; `isAnyType` narrowed back to `VariableType`.
6. **Regression guard**: a test scanning the source for `| "any"` signatures, bare `return "any"`, and stray inline `.value === "any"`.

`ScopeType` is left as `type ScopeType = VariableType` — it is an exported semantic name used across the backends, not a pure sentinel artifact, so removing it would churn unrelated code without retiring anything.

## Two subtleties worth review

- **`widenType` is a structure-preserving map**, so its `vt === "any"` check could NOT be mechanically swapped to `isAnyType` mid-migration: `isAnyType(ANY_T_object)` is true, and returning the string there corrupted nested `resultType` fields (a bare `Result` parses to `{successType:"any", failureType:"any"}`). Caught by the behavior gate; `widenType` was cleaned up in full instead.
- **The assignability fast-path deletion** (`isAssignableGuarded`) is a real correctness point: deleting the *other* check would have regressed `type Foo = any` assignability. A both-directions regression test locks it.

## Testing

`tsc --noEmit` clean. The full `lib/` unit suite passes: **8132 tests, 0 failures** — no behavior change, with one deliberate exception (`flow.test.ts` assertions moved from `.toBe("any")` to `.toEqual(ANY_T)`, plus three value-input migrations). Every task ended green before the next began.

### DIFF
--- a/packages/agency-lang/lib/typeChecker/anySentinelRetired.test.ts
+++ b/packages/agency-lang/lib/typeChecker/anySentinelRetired.test.ts
@@ -45,11 +45,14 @@ describe("the 'any' string sentinel stays retired (#472)", () => {
 
   it("no inline object-form any-check outside isAnyType", () => {
     // The one legitimate `.value === "any"` is isAnyType's body in utils.ts.
-    const re = /\.value\s*===\s*"any"/;
+    // Assert a COUNT (1 there, 0 elsewhere) rather than skipping utils.ts, so
+    // a second inline check added to a different function there is still caught.
+    const re = /\.value\s*===\s*"any"/g;
     for (const file of files) {
-      if (path.basename(file) === "utils.ts") continue;
       const src = fs.readFileSync(file, "utf8");
-      expect(re.test(src), path.basename(file)).toBe(false);
+      const matches = src.match(re) ?? [];
+      const expected = path.basename(file) === "utils.ts" ? 1 : 0;
+      expect(matches.length, path.basename(file)).toBe(expected);
     }
   });
 });
--- a/packages/agency-lang/lib/typeChecker/anySentinelRetired.test.ts
+++ b/packages/agency-lang/lib/typeChecker/anySentinelRetired.test.ts
@@ -1,0 +1,55 @@
+import { describe, it, expect } from "vitest";
+import * as fs from "fs";
+import * as path from "path";
+import { fileURLToPath } from "url";
+
+const dir = path.dirname(fileURLToPath(import.meta.url));
+
+// Recursive so a future subdir under lib/typeChecker is not silently dropped.
+function sourceFiles(root: string): string[] {
+  const out: string[] = [];
+  for (const entry of fs.readdirSync(root, { withFileTypes: true })) {
+    const full = path.join(root, entry.name);
+    if (entry.isDirectory()) out.push(...sourceFiles(full));
+    else if (entry.name.endsWith(".ts") && !entry.name.endsWith(".test.ts")) {
+      out.push(full);
+    }
+  }
+  return out;
+}
+
+const files = sourceFiles(dir);
+
+// A cheap tripwire backing the type-level guarantee (#472), not a proof: once
+// every signature is VariableType, the "any" string sentinel is
+// unrepresentable. This catches a stray hand-edit that slips past the compiler
+// or formatter.
+describe("the 'any' string sentinel stays retired (#472)", () => {
+  it('no signature unions VariableType with the "any" string', () => {
+    // Match `| "any"` and `"any" |` with any spacing, so a formatter-evading
+    // hand-edit or reversed member order is still caught.
+    const re = /(\|\s*"any")|("any"\s*\|)/;
+    for (const file of files) {
+      const src = fs.readFileSync(file, "utf8");
+      expect(re.test(src), path.basename(file)).toBe(false);
+    }
+  });
+
+  it('no file returns the bare "any" string', () => {
+    const re = /return\s+\(?\s*"any"/;
+    for (const file of files) {
+      const src = fs.readFileSync(file, "utf8");
+      expect(re.test(src), path.basename(file)).toBe(false);
+    }
+  });
+
+  it("no inline object-form any-check outside isAnyType", () => {
+    // The one legitimate `.value === "any"` is isAnyType's body in utils.ts.
+    const re = /\.value\s*===\s*"any"/;
+    for (const file of files) {
+      if (path.basename(file) === "utils.ts") continue;
+      const src = fs.readFileSync(file, "utf8");
+      expect(re.test(src), path.basename(file)).toBe(false);
+    }
+  });
+});
--- a/docs/superpowers/plans/2026-07-12-retire-any-sentinel.md
+++ b/docs/superpowers/plans/2026-07-12-retire-any-sentinel.md
@@ -172,18 +172,26 @@ In `synthesizer.ts`, change all 27 `return "any"` to `return ANY_T` (already imp
 
 `resolveResultFieldType` (`synthesizer.ts:75`) and the `:199` site return `VariableType | "any" | null`, where `"any"` is a signal ("caller returns any") and `null` means "no resolution." Narrow the return to `VariableType | null`. Replace `return "any"` with `return ANY_T`. Then read each caller: where it did `if (x === "any") return "any"`, it now receives `ANY_T` (a real type) and should set/return it as the type; where it checked `null`, that branch is unchanged. Confirm the control flow matches the original for both the any-case and the null-case.
 
-- [ ] **Step 3: Migrate the nine `flow.test.ts` assertions**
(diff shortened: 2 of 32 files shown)


packages/agency-lang/lib/typeChecker/anySentinelRetired.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      const src = fs.readFileSync(file, "utf8");
+      expect(re.test(src), path.basename(file)).toBe(false);
+    }
+  });
+
+  it("no inline object-form any-check outside isAnyType", () => {
+    // The one legitimate `.value === "any"` is isAnyType's body in utils.ts.
+    const re = /\.value\s*===\s*"any"/;
+    for (const file of files) {
+      if (path.basename(file) === "utils.ts") continue;
```
egonSchiele: Minor: skipping *all* of `utils.ts` means a future inline `.value === "any"` added to a **different** function here would evade the guard (the file is currently clean because `typeAcceptsResult` was folded). Since there should be exactly one legitimate occurrence — `isAnyType`'s body — consider asserting a count of `1` for `utils.ts` instead of skipping it, so the file is still covered:

```ts
const matches = src.match(/\.value\s*===\s*"any"/g) ?? [];
const expected = path.basename(file) === "utils.ts" ? 1 : 0;
expect(matches.length, path.basename(file)).toBe(expected);
```

packages/agency-lang/lib/typeChecker/anySentinelRetired.test.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+const files = sourceFiles(dir);
+
+// A cheap tripwire backing the type-level guarantee (#472), not a proof: once
+// every signature is VariableType, the "any" string sentinel is
+// unrepresentable. This catches a stray hand-edit that slips past the compiler
+// or formatter.
+describe("the 'any' string sentinel stays retired (#472)", () => {
+  it('no signature unions VariableType with the "any" string', () => {
```
egonSchiele, line 28: Heads-up (fine to leave): these regexes scan raw source, so a future **doc-comment** that writes `VariableType | "any"` while explaining the retirement would turn CI red. Arguably that's the intent (the sentinel is gone, even in prose), but it can surprise. Not worth an AST parse for a tripwire — just noting the failure mode.

egonSchiele (commented):
## Review — clean, and it lands every point from the design/plan reviews

I re-ran the verification independently against the `-impl` branch head rather than trusting the summary. It holds up:

- **Retirement is actually complete.** On the branch, `git grep` finds zero `| "any"` / `"any" |` signatures, zero bare `return "any"`, and zero inline `.value === "any"` outside `utils.ts`. The only surviving `=== "any"` is `isAnyType`'s own body (`utils.ts:53`). The `.filter(t => t !== "any")` narrowing helpers are gone too — TS2367 would have forced them out, and the clean grep confirms it.
- **The assignability change is right, and for the right reason.** Deleting the pre-resolution fast path in `isAssignableGuarded` (rather than swapping it to `isAnyType`) and keeping the post-resolution `isAnyType(resolvedSource) || isAnyType(resolvedTarget)` is the correct call: a `typeAliasVariable` is not itself `any`, so only the post-resolution check catches `type Foo = any`. I traced both directions — with the post-resolution check removed, `primitiveType("any")` falls through to primitive equality (`"any" !== "number"`) and returns `false`, so **the new regression test genuinely bites** in both directions. `{ Foo: { body: ANY_T } }` is a valid minimal `TypeAliasEntry` and `safeResolveType` follows `.body`, so the test exercises the real resolution path.
- **All six object-form folds done, including `typeAcceptsResult`** in `utils.ts` (the one the guard can't see) — good catch to fold it rather than leave it.
- **Signal sites are correct.** `resolveResultFieldType`/`accessResultField` (`| "any" | null` → `| null`), `checker.ts` (`| undefined`), and `functionTypeRaises` (`| null | undefined`) each keep their nullish member and fold only the `"any"` disjunct. The `functionTypeRaises` `if (!target || isAnyType(target))` is type-safe because `!target` narrows away `null | undefined` before `isAnyType` sees it.
- **`widenType`** — deleting the `if (vt === "any") return "any"` line and letting `primitiveType("any")` fall to `default: return vt` is correct (widening any = any), and the PR body's note about why it couldn't be mechanically swapped mid-migration is accurate.
- **LSP consumers** (`builtinHover`, `inlayHint`, `typeResolution`) route their now-dead string comparisons through `isAnyType` correctly.

One accuracy note, not a problem: the design doc's non-goal "nothing outside `lib/typeChecker/` changes / zero `| "any"` outside" turned out slightly off — the LSP files had `=== "any"` *comparisons* on typeChecker-exported values, so narrowing the exports forced them to change. The PR handles it transparently; worth remembering the sentinel's blast radius reached the LSP.

Two minor suggestions on the regression guard below. Neither blocks — the type system is the real guarantee and this guard is a reasonable tripwire behind it.

Nice work — this is exactly the shape the refactor should have.

egonSchiele:
Thanks — all addressed in 6b20c5d01:

1. **types.ts:85** — updated the `acceptsNamedArgs` docstring to say the `any` type (`ANY_T`) skips validation.
2. **checker.ts:387** — updated the named-arg skip comment likewise.
3. **synthesizer.ts:67** — rewrote `resolveResultFieldType`'s return contract to `VariableType | ANY_T | null` (dropped the stale string-sentinel line).
4. **matchExprTypes.ts:83** — removed both redundant casts; `types` is `VariableType[]` and `widenType` returns `VariableType`, so `unionTypes(types.map((t) => widenType(t)))` is clean (tsc confirms).
5. **Guard (utils.ts coverage)** — applied your count-based version: assert exactly `1` `.value === "any"` in utils.ts (isAnyType's body), `0` elsewhere, so a second inline check added to a different function in utils.ts is still caught.

On the raw-source scan heads-up (#28): leaving it as-is per your note. No typeChecker source comment currently writes `| "any"` (I removed the one in the synthesizer contract), and a future doc-comment tripping it is arguably the intended "the sentinel is gone, even in prose" behavior — not worth an AST parse for a tripwire.

PR #531, opened 2026-07-12
Enforce export-visibility on plain imports (AG4010) + valid JSON for parse dir/

egonSchiele:
## Summary

Two follow-ups to the strict-imports work (#529), bundled because they're both small and touch adjacent surfaces. (The third follow-up — converting `SymbolTable.build`'s `export-from`/`pkg::` throws into clean diagnostics — is split into its own PR because it changes a 65-caller API contract.)

### 1. Enforce export-visibility on plain imports in the typechecker (AG4010)

Importing a defined-but-not-`export`ed function/type/constant is now a type error:

```
// lib.agency
def privateFn(): string { return "p" }   // no `export`
```
```
import { privateFn } from "./lib.agency"  // AG4010
```

This already fails on the **compile** path — the `importResolver` preprocessor throws `assertImportable` ("... is not exported"). It was just silent under `agency tc`. This surfaces it there too, as a clean diagnostic. Semantics mirror `importResolver` exactly:

- **Nodes are exempt** (importable without `export`).
- **`import test { }` bypasses it** (first-party test wiring may see private symbols).

**Blast radius measured before landing** (per the "measure first" call): swept `agency tc` over `stdlib` and the whole `tests/` tree → **0** violations, because anything that compiles already has exported imports. So a hard error is safe.

### 2. `parse`/`preprocess`: one valid JSON document for multiple inputs

`agency parse dir/` printed one JSON object per file, concatenated — not parseable. Now:

- single input → the bare AST (unchanged, backward compatible)
- multiple inputs → a JSON array of `{ file, program }`

Same fix applied to `preprocess`.

## Tests

- 3 new pass tests: non-exported import errors (AG4010), node import exempt, `import test` bypass. (`missingImportDiagnostic.test.ts` → 15 total.)
- Strengthened the `17a-parse-dir` cli-main case to `JSON.parse` the output and assert the array shape.

Verified locally: full typechecker vitest suite (1321 pass), the pass unit file (15 pass), cli-main integration harness (exit 0), structural linter clean, and direct CLI exercise of AG4010 (fires on private import, clean on exported + node) and the parse single-object/directory-array output.

AG4010 is documented via the generated `docs/site/diagnostics/` pages and `agency explain AG4010` (no manual CLI-page doc, per the #529 review preference).

### DIFF
--- a/packages/agency-lang/docs/site/diagnostics/index.md
+++ b/packages/agency-lang/docs/site/diagnostics/index.md
@@ -75,6 +75,7 @@ or suppress one on the next line with `// @tc-ignore AG####`.
 | [AG4007](names.md#ag4007) | Variable '&#123;name&#125;' is not defined. |
 | [AG4008](names.md#ag4008) | '&#123;name&#125;' is not defined in '&#123;module&#125;'. |
 | [AG4009](names.md#ag4009) | Cannot find module '&#123;module&#125;'. |
+| [AG4010](names.md#ag4010) | '&#123;name&#125;' is defined in '&#123;module&#125;' but is not exported. Add the 'export' keyword to its definition. |
 
 ## Match and narrowing
 
--- a/packages/agency-lang/docs/site/diagnostics/names.md
+++ b/packages/agency-lang/docs/site/diagnostics/names.md
@@ -101,3 +101,13 @@ An import names a symbol that its target Agency module does not define. The chec
 An import points at a module that does not resolve to any file. The path — a relative `./…` path, a `std::` module, or a `pkg::` package — was resolved the same way the compiler resolves it, and nothing exists there.
 
 **How to fix:** correct the path, create the missing file, or install the package that provides it. Agency imports must resolve to a real module.
+
+<a id="ag4010"></a>
+
+## AG4010 — '&#123;name&#125;' is defined in '&#123;module&#125;' but is not exported. Add the 'export' keyword to its definition.
+
+*Default severity: error.*
+
+An import names a symbol that its target module defines but does not `export`. A plain `import { ... }` can only see `export`ed functions, types, and constants — a bare `def`/`type` without `export` is module-private. (Nodes are the exception: they are importable without `export`.) The compile path already rejects this; the type checker reports it too.
+
+**How to fix:** add the `export` keyword to the definition in the target file, or import a symbol that is exported.
--- a/packages/agency-lang/lib/symbolTable.ts
+++ b/packages/agency-lang/lib/symbolTable.ts
@@ -471,7 +471,7 @@ function collectDirectInterruptEffects(body: AgencyNode[]): InterruptEffect[] {
   return effects.map((e) => ({ effect: e }));
 }
 
-function isExportedSymbol(sym: SymbolInfo): boolean {
+export function isExportedSymbol(sym: SymbolInfo): boolean {
   // Nodes are importable without an explicit `export` keyword (see importResolver
   // — node imports skip the assertExported check). Treat them as exported for
   // re-export purposes so `export { main } from "..."` and `export * from "..."`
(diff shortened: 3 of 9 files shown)


egonSchiele:
## Review — AG4010 export-visibility + valid JSON for `parse dir/`

Reviewed against the code on `main` (`importResolver.assertImportable`, `SymbolTable`, `forEachSource`). Clean, well-scoped, and the "measure blast radius first → 0 violations" discipline is the right call for a hard error. **No blocking issues.** A few minor notes below.

### Correctness — verified ✅

- The checker condition `!spec.testOnly && symbol.kind !== "node" && !symbol.exported` is an exact mirror of `assertImportable`: `test` imports bypass, nodes are exempt (a plain `import { someNode }` hits the `kind === "node"` exemption too, not just `import node { }`), everything else must be `export`ed. No false positives vs. the compile path that I could construct.
- `symbol.exported` is safe to read here: `hasOwnProperty` gates it just above, and for every non-`node` kind (`function`/`type`/`constant`) `exported` is a required `boolean`. The only optional-`exported` symbol is `node`, which is short-circuited first.
- `printAstResults`: single input → bare AST (backward compatible), multi → `[{file, program}]` array. Both `ast`/`parse` (alias) and `preprocess` — the two per-file JSON-emitting commands — are covered; `diagnostics` doesn't emit a single JSON doc so it's correctly left alone.

### Notes (all optional)

1. **Reuse `isExportedSymbol` instead of re-deriving the rule.** `symbolTable.ts` already has a private `isExportedSymbol(sym)` that is exactly `sym.kind === "node" ? true : !!sym.exported`. The new inline `symbol.kind !== "node" && !symbol.exported` is `!isExportedSymbol(symbol)`. Exporting that helper and using it keeps the "nodes are exempt" rule defined in one place, so a future change to node-export semantics can't drift between the two call sites. Minor, but this is a semantics-mirroring PR — the whole risk is drift.

2. **Non-exported *constants* surface as AG4008, not AG4010.** A `const` is only recorded as a `ConstantSymbol` when `exported && static && const` (symbolTable.ts:449), so a non-exported const isn't in `resolution.symbols` at all → it trips `importNameNotFound` (AG4008) rather than `importNameNotExported` (AG4010). This matches `importResolver` (which throws "is not defined" for the same case), so behavior is *consistent* — just flagging that the message a user sees for a private const differs from the AG4010 they'd expect. Pre-existing; fine to leave, but worth a one-line comment near the check so the next reader isn't surprised.

3. **`printAstResults` now buffers every program before printing.** Previously each file streamed out as it was parsed; now all ASTs are held in `results` simultaneously to emit one valid array. That's the simplest way to produce a single JSON document and totally fine for a debug command, but it is a memory-profile change for `parse <large-dir>/`. No action needed — noting the tradeoff.

### Test coverage nits

- The AG4010 tests cover a non-exported **function** and the two exemptions. Consider adding (a) a non-exported **type** case, and (b) a plain `import { nodeName }` (node imported via the non-`node` import form) to lock in the `kind === "node"` exemption on the `importStatement` path — that branch is currently only exercised through `import node { }`.
- A positive control (exported symbol → no AG4010) would make the intent explicit, if one isn't already in the file above the diff.

Overall LGTM — happy to see it merge as-is; the notes are polish.

egonSchiele:
Thanks — applied the two that reduce drift risk (pushed in da5e2cec8):

- **Note 1 (reuse `isExportedSymbol`):** exported it from `symbolTable.ts` and the pass now uses `!isExportedSymbol(symbol)`, so the "nodes are exempt" rule has one home. Agreed this is the important one for a semantics-mirroring PR.
- **Note 2 (const → AG4008):** added a comment at the check explaining a non-exported const trips AG4008 (not AG4010) because it's only recorded when `exported && static`, consistent with importResolver.
- **Tests:** added a non-exported **type** case and a plain `import { plainNode }` case (locks the node exemption on the `importStatement` path, not just `import node { }`). 17 tests total.

Left **note 3** (printAstResults buffering) as-is per your call — it's a debug command and the buffering is what makes the single-JSON-array output possible.

For the deferred build-throws→diagnostics follow-up (the export-from / pkg:: stack-trace splat), filed #532 with the cost/benefit and options.

Issue #513, opened 2026-07-11
guard block around a paused subprocess run loses its value after serialize/resume (nested pause)


egonSchiele:
## Summary

When a subprocess pauses on a bubbled interrupt and its serialized stack contains a `guard(...) as { ... }` block wrapping the paused `run()` call, resuming loses the block's return value: after the approval resumes the tree, the outer `run(...)` expression evaluates to `undefined` instead of the child's Result.

Single-level pause/resume through a guard block works (`pause-multi-cycle` passes). The value is lost only when the process holding the guard block is itself a subprocess, i.e. its state stack goes through checkpoint serialize/deserialize before the resume.

## Repro

Nested version of `tests/agency/subprocess/nested-pause-resume.agency`, with the middle hop's `run` wrapped in a cost guard (or given `maxCost`, which wraps internally):

```agency
import { compile, run } from "std::agency"

node main() {
  const grandSource = """
import { bash } from "std::shell"
node main() {
  let r = bash("echo deep-ok")
  return r.stdout
}
"""
  const childSource = """
import { compile, run } from "std::agency"
node main(grandSource: string) {
  const c = compile(grandSource)
  if (isFailure(c)) {
    return "inner compile failed"
  }
  handle {
    const result = run(compiled: c.value, node: "main", maxCost: 1.0)
    if (isSuccess(result)) {
      return result.value.data
    }
    return "inner run failed"
  } with (e) {
    if (e.effect == "std::run") {
      return approve()
    }
  }
}
"""
  const compileResult = compile(childSource)
  if (isFailure(compileResult)) {
    return "compile failed"
  }
  handle {
    const result = run(
      compiled: compileResult.value,
      node: "main",
      args: { grandSource: grandSource },
    )
    if (isSuccess(result)) {
      return "OK: ${result.value.data}"
    }
    return "run failed"
  } with (e) {
    if (e.effect == "std::run" || e.effect == "std::bash") {
      return approve()
    }
  }
}
```

Expected: `OK: deep-ok`. Actual: the node crashes with `Cannot read properties of undefined (reading X)` when touching `result` — `run()` returned `undefined` after the resume.

The same failure reproduces without any PR #512 feature by writing the inner hop as a user-level `guard(cost: 1.0) as { return run(...) }`, so the bug is in guard-block resume across serialization, not in `maxCost` itself (unverified on main, but the code path predates #512).

## Where I got to

[...]

PR #534, opened 2026-07-13
std::agents worker-agent library (part 1): codingAgent, researchAgent, agencyCodingAgent + lifted verify

egonSchiele:
## What this is

Part 1 of the `std::agents` building-block library: three composable, batteries-included worker agents plus a lifted, fail-open `verify`. This is the substrate for the "agent writes an agent as its plan" idea (`plannerAgent` is Part 2, deliberately deferred).

Ships:
- **`std::agents/coding`** — `codingAgent(task, context?, maxCost?, maxTime?)`: curated file/shell tools + a strict verify→fix loop.
- **`std::agents/research`** — `researchAgent(...)`: web/Wikipedia sources + a grounding-and-completeness judge.
- **`std::agents/agency`** — `agencyCodingAgent(...)`: writes an Agency program, runs it, and judges `runCode`'s actual result. "writeAgency done right."
- **`std::agents/shared`** — `withContext`.
- **`std::agency` additions** — `verify(task): Result<Feedback[]>` (lifted out of the agent tree so stdlib can share it; the agent's `verify.agency` now delegates), plus `failOpenFeedback` and `syntaxHintFor`.
- **`writeAgency` upgrade** — two worked examples in the prompt + point-of-need `syntaxHintFor` injection on typecheck-repair rounds.

## Design (spec + plan in the diff)

- **Composition, not configuration.** Agents are independent `def`s combinable any way; shared logic is only ever a plain helper. No base class, no config schema.
- **Reuse the existing `Feedback` type** (per-line `error` flag + `data`); no parallel `Verdict`.
- **Two verifiers, not three:** `review` (static Agency analysis) + `verify` (runs the artifact). Both return `Result<Feedback[]>`.
- **`guard` is load-bearing:** each loop lives *inside* a `guard(cost:, time:)` block that returns the reply; the agent captures and `match`es the `Result`, with a defined budget-trip return and a small attempt cap so partial work survives normal termination.
- Reliability: worked examples + **point-of-need guidance injection** (the Nemotron "tune the harness" lever), curated tools, strict output-contract checking.

The design went through two external review rounds. Review v2 caught a foundational bug in the first plan — the original `guard` loop mutated an outer variable from inside the guard and returned it, which `guards.md` explicitly forbids (only the return value escapes), so every agent would have returned `""`. Fixed before any code.

## Validation

- **Deterministic tests (run on PRs via `test:agency`):** `withContext`, `failOpenFeedback`, `syntaxHintFor` unit tests; `verify` fail-open; module wiring/compile checks. 5 files / 11 tests, all green.
- **Real-LLM end-to-end (local):**
  - `codingAgent` smoke produced `result.txt` with exact bytes `42\n` — nailing the byte-level output contract, the exact failure class this work targets.
  - Efficacy harness `verify-strict` (both polarities: wrong file → errors, correct file → none) and `writeagency-regression` (produces compiling source) both PASS.
- **Real-LLM efficacy suite** (`tests/integration/agents`, 5 cases) is wired into `test-with-llm.yml` as a **non-blocking** (`continue-on-error`) post-merge step, so it tracks worker-agent quality without failing the build. `agency-sum` and `research-cited` are exercised there (not run locally to save spend).

No new typecheck errors vs. the baseline (verified by diffing `make` output; the repo's ~8 pre-existing diagnostics are unchanged).

## Notes / caveats

- Agents raise handler-gated tool interrupts by design; callers approve them (the efficacy drivers use `with approve`). This is the governance property, working as intended.
- The `guard`+`#513` nested-pause interaction is documented for `agencyCodingAgent` (a paused generated program can lose the guard value) — acceptable for v1, watched.
- Rewiring the main `agency agent` to use these primitives (and fixing its shadowed-`verify` / crashing-`review()` bug) is a separate follow-up, out of scope here.

### DIFF
--- a/packages/agency-lang/stdlib/agency.agency
+++ b/packages/agency-lang/stdlib/agency.agency
@@ -521,7 +521,7 @@ export def feedbackHasErrors(feedback: Result<Feedback[]>): boolean {
 }
 
 /** Pass a successful feedback list through unchanged; map any failure to
-  no findings. Used to make verification fail open — a verifier that could
+  no findings. Used to make verification fail open. A verifier that could
   not run reports nothing rather than a blocking error. */
 export def failOpenFeedback(r: Result<Feedback[]>): Result<Feedback[]> {
   return match (r) {
@@ -530,17 +530,27 @@ export def failOpenFeedback(r: Result<Feedback[]>): Result<Feedback[]> {
   }
 }
 
-/** A specific syntax reminder to inject next to a typecheck error, or ""
-  when no known pattern matches. Point-of-need guidance: the rule lands
-  better delivered alongside the error than buried in the system prompt. */
+/** A specific syntax reminder to inject next to a diagnostic, or "" when no
+  known pattern matches. Point-of-need guidance: the rule lands better delivered
+  alongside the error than buried in the system prompt. Matches on the actual
+  diagnostic wording the compiler emits, not on the user's source. */
 export def syntaxHintFor(errors: string): string {
-  if (errors =~ re/for\s*\([^)]*;/) {
-    return "Reminder: Agency has no C-style for loop. Count with a while loop."
-  }
-  if (errors =~ re/no node named .?main/) {
-    return "Reminder: the program MUST have a node named `main` as its entry point."
+  if (errors =~ re/for loop condition/) {
+    return "Reminder: Agency has no C-style for loop. Count with a while loop, or iterate with for (x in items)."
   }
   return ""
+}
+
+/** True when the source declares a `node main` entry point. Missing main is
+  not a typecheck diagnostic, so writeAgency checks for it directly. */
+def hasMainNode(source: string): boolean {
+  return source =~ re/node\s+main\s*\(/
+}
+
+/** Prepend the point-of-need syntax hint to a diagnostic block, if one applies. */
+def withSyntaxHint(findings: string): string {
+  const hint = syntaxHintFor(findings)
+  return if hint == "" then findings else "${hint}\n\n${findings}"
 }
 
 static const verifyTools: any[] = [
@@ -551,9 +561,9 @@ static const verifyTools: any[] = [
   exec.partial(useAgentCwd: true),
 ]
 
-static const verifySysPrompt = """You verify whether a coding task was completed correctly. You do NOT fix anything — inspect, run, and report.
+static const verifySysPrompt = """You verify whether a coding task was completed correctly. You do NOT fix anything. Inspect, run, and report.
 
-The grading tests are hidden and NOT in this environment. Reconstruct the success check from the task's own description and any example usage, then RUN the artifact through it with the `exec` tool. Check the output contract STRICTLY: exact file path and name, exact format (JSON keys, field types, units, signed-vs-unsigned, a trailing newline), and every explicit criterion. A near-miss on the output contract is an error."""
+The grading tests are hidden and NOT in this environment. Reconstruct the success check from the task's own description and any example usage, then RUN the artifact through it with the `exec` tool. Check the output contract STRICTLY: exact file path and name, exact format (JSON keys, field types, units, signed-vs-unsigned), the task's stated whitespace and trailing-newline requirement (present or absent), and every explicit criterion. A near-miss on the output contract is an error."""
 
 // The throwing inner: two-phase inspection. `verify` wraps it fail-open.
 def verifyInner(task: string): Feedback[] {
@@ -610,9 +620,9 @@ Agency syntax rules you MUST follow:
 - Declare variables with let or const before use.
 - No Python-style colon blocks. Always { ... }.
 
-Here are two complete, correct Agency programs. Copy their structure — the imports (if any), the `node main`, and the explicit `return` at the end.
+Here are complete, correct Agency programs. Copy their structure: the imports, the `node main`, and the explicit `return` at the end.
 
-EXAMPLE 1 — compute and return a value:
+EXAMPLE 1, compute and return a value:
 node main(): number {
   let total = 0
   for (x in [1, 2, 3]) {
@@ -621,10 +631,28 @@ node main(): number {
   return total
 }
 
-EXAMPLE 2 — call the model and return its answer:
-node main(): string {
-  return llm("Reply with a friendly greeting.")
+EXAMPLE 2, import a tool and define and call a helper:
+import { bash } from "std::shell"
+def greet(name: string): string {
+  return "Hello, \${name}"
 }
+node main(): string {
+  bash(command: "echo starting")
+  return greet("world")
+}
+
+EXAMPLE 3, get structured output from the model by annotating the result type:
+type Summary = { title: string; points: string[] }
+node main(): Summary {
+  const s: Summary = llm("Summarize the water cycle with a title and three bullet points.")
+  return s
+}
+
+Commonly useful standard library modules:
+- std::shell: bash, exec, ls, grep, glob (run commands, inspect files)
+- std::http: fetch, fetchJSON (call web APIs)
+- std::array: map, filter (transform lists)
+- std::agency: writeAgency, runCode, review (generate and run Agency code)
 
 If you are unsure about syntax or a standard-library function, read the
 relevant page with the docs tool before writing code.
@@ -670,20 +698,19 @@ export def writeAgency(
         if (findings == "") {
           findings = "${parseErr}"
         }
-        const hint = syntaxHintFor(findings)
-        const body = if hint == "" then findings else "${hint}\n\n${findings}"
-        prompt = "That code has problems:\n${body}\nReturn a corrected complete program."
+        prompt = "That code has problems:\n${withSyntaxHint(findings)}\nReturn a corrected complete program."
       } else if (checked.value.errors.length > 0) {
         result = failure({ source: source, errors: checked.value.errors })
         const findings = renderFeedback(success(reportFeedback(checked.value)))
-        const hint = syntaxHintFor(findings)
-        const body = if hint == "" then findings else "${hint}\n\n${findings}"
-        prompt = "That code has problems:\n${body}\nReturn a corrected complete program."
+        prompt = "That code has problems:\n${withSyntaxHint(findings)}\nReturn a corrected complete program."
       } else {
         const compiled = try _compile(source)
         if (compiled is failure(compileErr)) {
           result = failure({ source: source, errors: [] })
-          prompt = "That code failed to compile: ${compileErr}\nReturn a corrected complete program."
+          prompt = "That code failed to compile:\n${withSyntaxHint("${compileErr}")}\nReturn a corrected complete program."
+        } else if (!hasMainNode(source)) {
+          result = failure({ source: source, errors: [] })
+          prompt = "The program compiles but has no `node main`. Every program MUST have a node named `main` as its entry point. Add one and return the complete program."
         } else {
           result = success(source)
           done = true
--- a/packages/agency-lang/stdlib/agency.agency
+++ b/packages/agency-lang/stdlib/agency.agency
@@ -16,6 +16,7 @@ import {
 import { VERSION } from "agency-lang/stdlib-lib/version.js"
 import { guard, systemMessage } from "std::thread"
 import { docsSkill } from "std::skills"
+import { ls, glob, grep, exec } from "std::shell"
 
 /** @module
   Tools for compiling, type-checking, running, formatting, and inspecting
@@ -540,6 +541,42 @@ export def syntaxHintFor(errors: string): string {
     return "Reminder: the program MUST have a node named `main` as its entry point."
   }
   return ""
+}
+
+static const verifyTools: any[] = [
+  read.partial(useAgentCwd: true),
+  ls.partial(useAgentCwd: true),
+  glob.partial(useAgentCwd: true),
+  grep.partial(useAgentCwd: true),
+  exec.partial(useAgentCwd: true),
+]
+
+static const verifySysPrompt = """You verify whether a coding task was completed correctly. You do NOT fix anything — inspect, run, and report.
+
+The grading tests are hidden and NOT in this environment. Reconstruct the success check from the task's own description and any example usage, then RUN the artifact through it with the `exec` tool. Check the output contract STRICTLY: exact file path and name, exact format (JSON keys, field types, units, signed-vs-unsigned, a trailing newline), and every explicit criterion. A near-miss on the output contract is an error."""
+
+// The throwing inner: two-phase inspection. `verify` wraps it fail-open.
+def verifyInner(task: string): Feedback[] {
+  let items: Feedback[] = []
+  thread(label: "verify", summarize: true) {
+    systemMessage(verifySysPrompt)
+    const analysis: string = llm("The task was:\n\n${task}\n\nInspect the working directory, run the artifact against the task's own success criteria, and describe what passes and what fails.", { tools: verifyTools })
+    const parsed: Feedback[] = llm("Convert this analysis into Feedback items — one per concrete finding: error=true for an unmet criterion (name it), error=false for a satisfied one. Empty list if you cannot tell.\n\nAnalysis:\n${analysis}")
+    items = parsed
+  }
+  return items
+}
+
+export def verify(task: string): Result<Feedback[]> {
+  """
+  Reconstruct a task's success criteria, run the produced artifact against
+  them, and return findings. `error: true` marks an unmet criterion (strict
+  on the output contract); `error: false` a satisfied one. Fails open (empty
+  success) if verification cannot run, so it never blocks.
+
+  @param task - The original task description to verify against.
+  """
+  return failOpenFeedback(try verifyInner(task))
 }
 
 type GeneratedCode = {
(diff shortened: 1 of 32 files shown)


packages/agency-lang/stdlib/agents/agency.agency
```diff
@@ -0,0 +1,87 @@
...
+}
+
+def failedToRun(err: any): AttemptOutcome {
+  return { done: false, feedback: "The program failed to run: ${err}" }
+}
+
+// Judge the program's ACTUAL result against the task (its deliverable is the
+// return value, not a disk artifact — so the disk-running `verify` is wrong here).
+def outcomeFromJudge(task: string, data: any): AttemptOutcome {
+  const items: Feedback[] = llm("Does this program's result satisfy the task? One Feedback item per finding; error=true if it fails the task (name it), error=false for a satisfied point.\n\nTask: ${task}\n\nResult: ${data}")
```
egonSchiele: This is hard to read, separate the prompt out into its own variable and use a multi-line string.

packages/agency-lang/stdlib/agents/agency.agency
```diff
@@ -0,0 +1,87 @@
+/** @module
+  Agency-coding agent: writes an Agency program for a task, runs it, and
+  verifies the program's actual result satisfies the task. Built on
+  writeAgency + runCode. Composable — call it directly or from another agent.
```
egonSchiele: remove "Composable — call it directly or from another agent."

egonSchiele: And stop overusing em-dashes

packages/agency-lang/stdlib/agents/agency.agency
```diff
@@ -0,0 +1,87 @@
...
+    failure(_) => keepResult(state)
+    success(src) => advanceFromSource(task, context, state, src)
+  }
+}
+
+export def agencyCodingAgent(task: string, context: string = "", maxCost: number = $50.00, maxTime: number = 10m): Result<string, WriteFailure> {
+  """
+  Writes an Agency program for the task, runs it, and verifies its result
+  satisfies the task. Returns source that compiles, typechecks, and passes
+  verification.
```
egonSchiele: **(medium) The docstring over-promises on the attempt-exhaustion path.** Trace the loop with the default `attemptsLeft = 2`:

1. `writeAgency` runs once before the loop → attempt #1 (unjudged), `done=false`.
2. Iter A (`attemptsLeft=2`): `stepOnce` judges #1; if it fails, `advanceFromSource` sets `result = writeAgency(...)` → #2 (**unjudged**), `attemptsLeft=1`.
3. Iter B (`attemptsLeft=1`): judges #2; if it fails, `result = writeAgency(...)` → #3 (**unjudged**), `attemptsLeft=0`.
4. Loop exits; returns #3, which was **never run through `judgeRun`**.

So on a hard task the agent burns 3 generations, judges only 2, and returns the last one unverified — yet the docstring says *"Returns source that compiles, typechecks, and passes verification."* The compiles/typechecks part holds (writeAgency's own loop guarantees it), but *passes verification* does not on exhaustion. Either judge the final attempt before returning (loop on generations, not on a pre-generated extra), or soften the docstring to note the budget-exhausted return is best-effort and may not have passed result-verification.

packages/agency-lang/tests/agency/agents/helpers.agency
```diff
@@ -0,0 +1,22 @@
...
+node contextFilled(): boolean {
+  return withContext("do X", "c") == "do X\n\n<context>\nc\n</context>"
+}
+
+node hintForCStyleFor(): boolean {
+  return syntaxHintFor("Parse error: for (i=0; i<3; i++)") =~ re/while loop/
+}
+
+node hintForMissingMain(): boolean {
+  return syntaxHintFor("no node named main") =~ re/node named .main./
```
egonSchiele: **(medium) These hint tests validate the regex, not that it fires on real diagnostics.** `"no node named main"` and `"Parse error: for (i=0; i<3; i++)"` are hand-written to match `syntaxHintFor`'s patterns, so they prove the regex logic but not that an *actual* `writeAgency` typecheck/parse diagnostic ever contains those substrings. If the real missing-`main` diagnostic reads e.g. "Program must define a `main` node", `re/no node named .?main/` never fires in production and this test stays green. (Related: is "no main" even a *typecheck* diagnostic `writeAgency` sees, or only a compile/run-time error? If the latter, the hint is unreachable at the injection sites.) Add one case that feeds a **real** captured diagnostic — compile a C-style-`for` / missing-`main` program, run `renderFeedback(...)` output through `syntaxHintFor`, and assert the hint is non-empty.

packages/agency-lang/stdlib/agency.agency
```diff
@@ -593,11 +670,15 @@ export def writeAgency(
...
         }
-        prompt = "That code has problems:\n${findings}\nReturn a corrected complete program."
+        const hint = syntaxHintFor(findings)
+        const body = if hint == "" then findings else "${hint}\n\n${findings}"
+        prompt = "That code has problems:\n${body}\nReturn a corrected complete program."
       } else if (checked.value.errors.length > 0) {
         result = failure({ source: source, errors: checked.value.errors })
         const findings = renderFeedback(success(reportFeedback(checked.value)))
-        prompt = "That code has problems:\n${findings}\nReturn a corrected complete program."
+        const hint = syntaxHintFor(findings)
```
egonSchiele: **(minor) The compile-failure branch just below (~line 682, `} else { const compiled = try _compile(...)`) doesn't get this same `syntaxHintFor` injection.** Both the parse and typecheck branches now prepend the hint, but a program that parses and typechecks yet fails `_compile` gets the raw error with no point-of-need guidance. Consider factoring the `const hint = …; const body = …` two lines into a tiny helper and applying it at all three sites so they can't drift.

packages/agency-lang/tests/integration/agents/test.mjs
```diff
@@ -0,0 +1,176 @@
...
+const CLI = join(PACKAGE_DIR, "dist", "scripts", "agency.js");
+const FIXTURES = join(HERE, "fixtures");
+const RETRIES = Number(process.env.AGENTS_EFFICACY_RETRIES ?? "1");
+const ONLY = (process.env.AGENTS_EFFICACY_ONLY ?? "").split(",").map((s) => s.trim()).filter(Boolean);
+
+class Skip extends Error {}
+
+function freshDir(name) {
+  return mkdtempSync(join(PACKAGE_DIR, `agents-efficacy-${name}-`));
+}
```
egonSchiele, line 32: **(nit)** Rooting the temp dir in `PACKAGE_DIR` is the right call (the driver needs `node_modules` resolvable). But a `SIGKILL`/timeout between here and the `finally rmSync` leaves orphan `agents-efficacy-*` dirs in the package root. Add `agents-efficacy-*` to `.gitignore` so a stray one never gets committed.

packages/agency-lang/tests/agency/agents/codingWiring.agency
```diff
@@ -0,0 +1,9 @@
+import { codingAgent } from "std::agents/coding"
+
+// Compile-time wiring check: importing and referencing `codingAgent` proves
+// the module compiles, the tools resolve, and the guard/verify loop typechecks.
```
egonSchiele: **(nit)** The comment says "importing and referencing `codingAgent` proves…", but the node body only returns `true` — `codingAgent` is referenced *only* in the `import`. That's enough to force compilation (which is the real value here), so no change needed to the test; just tighten the comment to "importing `codingAgent`" to match what actually happens.

egonSchiele (commented):
Re-review (I did the two plan-review rounds). This is a strong implementation — every earlier **blocking** issue is resolved:

- `guard` scoping fixed: the loop lives *inside* the guard, the block `return`s the value, and the caller `match`es the resulting `Result` with a defined budget-trip return. (This was the foundational return-`""` bug from review-v2.)
- `verify` now fails open correctly via `failOpenFeedback(try verifyInner(...))` — the `try` is inside, and it's unit-tested in both directions.
- `agencyCodingAgent` judges `runCode`'s **actual result** (`success(data)` → `outcomeFromJudge`) instead of the disk-running `verify`, which was the D2 mismatch — and `runCode` returns `success(envelope.data)`, so `data` is the real returned value. 
- The `match` is hoisted out of the stepped `while` body into plain `stepOnce`/`advanceFromSource`/`judgeRun` helpers, per the codegen constraint.
- `withContext` extracted; `session:` dropped from the threads; tests now cover both `verify` polarities and a loop-forcing byte contract (`{"count":3}`, no trailing newline) for `codingAgent`; `researchAgent` has a real efficacy case with a graceful `Skip` when search is off.

Remaining comments are inline — two medium, three minor. **Nothing blocking.** Two more that don't anchor cleanly to a single added line:

- **(minor) writeAgency compile-failure branch omits the hint.** `syntaxHintFor` is injected in the parse-fail and typecheck-error branches but not the third branch (`compiled is failure(compileErr)`, ~`stdlib/agency.agency:682-684`). If a missing-`main` surfaces as a compile error rather than a typecheck diagnostic, the point-of-need hint won't fire there. Low priority; flagged inline near the second injection site.
- **(nit) efficacy temp dirs.** `mkdtempSync(join(PACKAGE_DIR, "agents-efficacy-…"))` is correctly rooted in the package (needs `node_modules`), but a hard kill between `freshDir` and the `finally` leaves orphan `agents-efficacy-*` dirs in the package root — worth a `.gitignore` entry so one never gets committed.

Nice work turning this around.

packages/agency-lang/stdlib/agents/agency.agency
```diff
@@ -0,0 +1,87 @@
...
+  return { result: state.result, done: true }
+}
+
+def advanceFromSource(task: string, context: string, state: StepState, src: string): StepState {
+  const outcome = judgeRun(task, src)
+  if (outcome.done) {
+    return { result: state.result, done: true }
+  }
+  const extra = "${context}\n\nA prior attempt had these problems; fix them:\n${outcome.feedback}"
+  return { result: writeAgency(task, extra), done: false }
```
egonSchiele: What happens if writeAgency throws an interrupt? Do interrupts work correctly when they're thrown in a function call that is the value of an object's key?

packages/agency-lang/stdlib/agents/agency.agency
```diff
@@ -0,0 +1,87 @@
...
+}
+
+def stepOnce(task: string, context: string, state: StepState): StepState {
+  return match (state.result) {
+    failure(_) => keepResult(state)
+    success(src) => advanceFromSource(task, context, state, src)
+  }
+}
+
+export def agencyCodingAgent(task: string, context: string = "", maxCost: number = $50.00, maxTime: number = 10m): Result<string, WriteFailure> {
```
egonSchiele: Let's bump the default max time up to 30 minutes.

packages/agency-lang/stdlib/agents/agency.agency
```diff
@@ -0,0 +1,87 @@
...
+  satisfies the task. Returns source that compiles, typechecks, and passes
+  verification.
+
+  @param task - What the generated program should do.
+  @param context - Optional extra material.
+  @param maxCost - Hard spend cap for the whole run (default $50).
+  @param maxTime - Hard wall-clock cap for the whole run (default 10 minutes).
+  """
+  const captured = guard(cost: maxCost, time: maxTime) as {
+    let state: StepState = { result: writeAgency(task, context), done: false }
```
egonSchiele: Note that because of the way state isolation works in agency, if there's a value that you're constantly passing into functions, you can just pull it out and make it a global variable instead.

egonSchiele: Though in this case, I suppose it's nice to see exactly where that variable is being used as well.

packages/agency-lang/stdlib/agents/coding.agency
```diff
@@ -0,0 +1,65 @@
...
+  read.partial(useAgentCwd: true),
+  write.partial(useAgentCwd: true),
+  edit.partial(useAgentCwd: true),
+  ls.partial(useAgentCwd: true),
+  glob.partial(useAgentCwd: true),
+  grep.partial(useAgentCwd: true),
+  bash.partial(useAgentCwd: true),
+]
+
+static const codingSysPrompt = """You are an expert software engineer. Solve the task by writing and running real code with your tools.
```
egonSchiele, line 22: What happened to putting examples in here for the agent learn from?

packages/agency-lang/stdlib/agents/coding.agency
```diff
@@ -0,0 +1,65 @@
...
+
+  @param task - What to build or fix.
+  @param context - Optional extra material (data, examples, constraints).
+  @param maxCost - Hard spend cap for the whole run (default $50).
+  @param maxTime - Hard wall-clock cap for the whole run (default 10 minutes).
+  """
+  const captured = guard(cost: maxCost, time: maxTime) as {
+    let reply = ""
+    let msg = withContext(task, context)
+    thread(label: "codingAgent", summarize: true) {
```
egonSchiele: This brings up a good point. Should we give users the option to pass in a session name? Which would mean that all calls to this agent then share the same thread history?
https://agency-lang.com/guide/cross-thread-context.html#id-label-and-session

egonSchiele: Also I'm not sure eager summarization is useful here?

packages/agency-lang/stdlib/agents/coding.agency
```diff
@@ -0,0 +1,65 @@
...
+      systemMessage(codingSysPrompt)
+      let done = false
+      let attemptsLeft = 3
```
egonSchiele: should the number of attempts be configurable by the user (with a default)?

packages/agency-lang/stdlib/agents/research.agency
```diff
@@ -0,0 +1,60 @@
+/** @module
+  Research agent: gathers from web/Wikipedia sources and synthesizes a cited,
+  grounded answer. Composable — call it directly or from another agent.
+*/
+import { search as webSearch } from "std::web/search"
```
egonSchiele, line 5: Note that there are a couple of search options, but they require API keys that the user may not have. So we need to check to see if they have those API keys before trying to use these tools.

packages/agency-lang/stdlib/agency.agency
```diff
@@ -549,6 +609,23 @@ Agency syntax rules you MUST follow:
...
+EXAMPLE 1 — compute and return a value:
+node main(): number {
+  let total = 0
+  for (x in [1, 2, 3]) {
+    total = total + x
+  }
+  return total
+}
+
+EXAMPLE 2 — call the model and return its answer:
```
egonSchiele: Also show examples of:
- import statements,
- how to define and call functions,
- how to get structured output.

Also, a quick summary of a few of agency's most used standard library modules (or ones the agent is likely to use)

packages/agency-lang/tests/agency/agents/agencyWiring.agency
```diff
@@ -0,0 +1,7 @@
+import { agencyCodingAgent } from "std::agents/agency"
+
+// Compile-time wiring check (not invoked — would hit the LLM + subprocess).
```
egonSchiele: What is this file for?

egonSchiele:
Thanks for the thorough review — addressed in the latest commits. Highlights:

**Verified + fixed the big one (your `helpers.agency:17` catch).** You were right, and it was worse than flagged: `syntaxHintFor` matched the *user's source* (`for(…;`), but the real diagnostic is `"expected \`(\` to open for loop condition"` — so it **never fired**. And a program with no `main` **typechecks clean**, so that branch matched a diagnostic that doesn't exist. Fixes:
- `syntaxHintFor` now matches the real diagnostic wording (`/for loop condition/`); confirmed it fires end-to-end.
- Missing-`main` is now a `hasMainNode` source check inside `writeAgency` (not a typecheck hint).
- New test feeds a **real** captured diagnostic (via `review(...)`, no LLM) through `syntaxHintFor`.

**Other fixes:**
- `withSyntaxHint` factored and applied at all three `writeAgency` feedback sites (incl. compile).
- `writeSysPrompt`: added import / function def+call / structured-output examples + a common-stdlib summary. (Had to escape a literal `${...}` in an example — unescaped, it interpolated at module init and crashed every `std::agency` importer. Good reminder that prompt strings are live.)
- `verify.agency` delegate: one gap per **error** finding (drops `error:false` lines), fail-open preserved.
- `agencyCodingAgent`: judges every generation before returning (no unjudged extra on exhaustion — your loop trace was exact); binds `writeAgency` result to a `const` before the object literal (your interrupt question); guard-trip payload carries the trip `type`; judge prompt extracted to a const.
- `researchAgent`: offers the Brave web-search tool only when `BRAVE_API_KEY` is set; falls back to keyless Wikipedia/fetch otherwise.
- Reworded the trailing-newline guidance (coding + verify prompts) to respect the task's stated requirement, present *or* absent.
- `maxAttempts` configurable; default `maxTime` → 30m; dropped `summarize: true`; cut em-dashes; removed "Composable…" lines; `.gitignore agents-efficacy-*`.

**Your open questions / decisions:**
- Session param: **skipped for v1** (fresh thread per call); easy to add later.
- codingAgent examples: kept the prompt **lean** — it emits Python/shell, so the verify→fix loop does the enforcing.
- Compile-only wiring tests: **dropped** (`make` already compiles the modules).
- State-isolation globals for `task`/`context`: left explicit per your own follow-up note ("nice to see where it's used").

Rebuilt clean (no new typecheck errors); deterministic tests pass; re-validated `writeAgency` + `verify` strictness through the real-LLM efficacy harness.

PR #535, opened 2026-07-13
Make the destructive marker authoritative; add destructive { } regions

egonSchiele:
## Summary

Makes the `destructive` declaration **authoritative** for tool removal-on-failure, and adds a `destructive { }` region so a tool's interrupt gate can stay retryable-on-rejection.

Previously, whether a failed `destructive` tool got removed was gated on a compiler heuristic (`containsImpureCall` — did the body call an *imported* symbol before failing), so `print` mattered but a same-file `def` didn't, and every real stdlib tool (which validates before its gate) became non-retryable on a rejected confirmation. Now:

- **`destructive def`** commits its whole body at **function entry**. Any failure in the body removes the tool; only an argument-binding failure (body never ran) stays retryable.
- **`destructive { }` region** (new) commits at **region entry**. Prep and the interrupt gate sit *outside* it, so rejecting the gate is retryable; a failure inside removes the tool.

## Design & reviews

Spec + two spec-review rounds + one plan-review round are included under `docs/superpowers/`. The plan review drove the key implementation decision below.

## Implementation

`destructive { }` is **not** a new AST node — it's a `seqBlock` carrying a `destructive: true` flag. `seqBlock` is already runtime-frame-transparent (`parallelDesugar` inlines a standalone seqBlock to its body on the function's `__self`) and already threaded through every body-dispatch site (bodySlots, flow-narrowing, scopes, preprocessors, formatter, lowering). Reusing it avoids wiring a new transparent node into ~8 hand-enumerated sites (and the silent breakage that would cause). Four places learn the flag:

1. **Parser** — `destructive { }` → `seqBlock{ destructive: true }`. The `{` is a soft gate (outside the committing `parseError`) so `destructive def` backtracks cleanly.
2. **Formatter** — `AgencyGenerator.processSeqBlock` prints `destructive {` when flagged.
3. **Desugar** — `parallelDesugar` prepends a synthetic `markDestructiveRan` leaf when inlining a destructive seqBlock; codegen emits it as `__self.__destructiveRan = true`. Because the seqBlock is inlined, the flip lands on the **function** activation, never a block frame — this is the fail-open guard.
4. **Metadata** — `isDestructive = marked OR contains-destructive-region` drives the descriptor marker, the MCP/HTTP client hint, and the `destructiveFunctions` registry, *without* mutating `node.markers` (which still keys the entry flip on the raw marker alone).

Also: `destructive def` sets the flag at entry via `init(true)`; the old per-statement Rule-1 scan and the dead `containsImpureCall`/`isImpureImportedFunction` heuristic are removed. Removal wiring (`failureTier`) is unchanged.

Two bugs found and fixed during implementation:
- `init` read the scope's `inDestructiveFunction` flag *after* it was restored, so `destructive def` never committed at entry — now threaded explicitly through `buildFunctionBody`.
- The metadata helper matches the region on **both** sides of desugar (the `seqBlock` pre-desugar in `compilationUnit`, the `markDestructiveRan` leaf post-desugar in `TypeScriptBuilder`).

## Stdlib migration

All ~29 stdlib `destructive def` functions migrated to `def` + `destructive { work }`, wrapping only the post-gate effectful work: git (9), fs (6), shell (2), index write/writeBinary (2), clipboard (1), calendar (3), memory (1, non-uniform — wraps its `thread` block), auth keyring/oauth (3), messaging email/imessage/sms (5). Each still reports `destructive` to clients via the contains-region derivation. Rejecting any of their confirmation gates is now retryable.

## Tests

- `destructive-tracking` rewritten to the region model (+ a `destructive def` entry-commit case).
- New `destructive-block` execution tests: **frame-escape** (a failure escaping the function after a region carries `destructiveRan` — the anti-fail-open guard), declaration-visibility (transparency), return-in-region, and interrupt-inside-region resume (with a statement after the region).
- Parser disambiguation + nested-`destructive{}`-in-`destructive def` fixture (asserts the harmless redundant flip).
- typescriptGenerator fixtures for the region and the nested case.

## Verification

- Full lib suite: **8163 passed**, 41 skipped. The one failure — `lib/cli/runShim/resolver.test.ts` — is a pre-existing worktree-naming artifact: the test asserts the checkout path `.includes("agency-lang")`, which is false in a worktree named `agency-destructive-marker-impl` but true in the `agency-lang` repo checkout (and in CI). Untouched by this branch; passes on the main checkout.
- `pnpm run lint:structure`, `tsc --noEmit`, and `make` all clean.
- Targeted agency tests (destructive-tracking, destructive-block, write-binary) pass.

## Non-goals

- Whether a *non-destructive* tool's rejected interrupt should be retryable (separate; PR #533).
- Unifying the success-path `toolDidDestructiveWork` onto the runtime flag (documented non-goal; the current over-taint is the safe direction).

### DIFF
--- a/packages/agency-lang/docs/site/guide/llm-part-2.md
+++ b/packages/agency-lang/docs/site/guide/llm-part-2.md
@@ -33,9 +33,28 @@ destructive def chargeCard(amount: number): string {
 }
 ```
 
-If a `destructive` tool starts running and then fails, Agency removes it from the conversation. The model cannot call it again, and the tool result tells it so: the operation may have partially completed, so the state must be verified by hand rather than by retrying.
+A `destructive def` marks the **whole body** as destructive: entering it commits. If the tool then fails anywhere in its body, Agency removes it from the conversation. The model cannot call it again, and the tool result tells it so: the operation may have partially completed, so the state must be verified by hand rather than by retrying.
 
-There is one exception. If the call fails *before the tool body starts* — the model passed the wrong arguments, or too few — then nothing happened yet, so the tool stays callable. Only a failure *after* the destructive work began locks the tool out.
+There is one exception. If the call fails *before the body starts* — the model passed the wrong arguments, or too few — then nothing ran yet, so the tool stays callable.
+
+#### `destructive { }` regions
+
+Marking a whole function `destructive` is often too coarse. Real tools validate their arguments and ask the user to confirm *before* doing anything dangerous — and rejecting that confirmation should leave the tool callable, not lock it out. Use a `destructive { }` region to mark only the effectful part, and leave the prep and the confirmation gate outside it:
+
+```agency
+def write(filename: string, content: string): Result {
+  const path = resolvePath(filename)
+  // Ask first — rejecting this is retryable, nothing has happened yet.
+  return interrupt std::write("Write to this file?", { path: path })
+  // Only this commits. A failure here removes the tool; a rejected
+  // confirmation above does not.
+  destructive {
+    return try _write(path, content)
+  }
+}
+```
+
+Entering the region flips the tool into "committed": a failure inside it (or after it) removes the tool, exactly like a `destructive def`. A failure *before* the region — a validation refusal, or a rejected interrupt gate — is retryable. The region introduces no new scope; variables declared inside it are visible afterward. A function that contains a `destructive { }` region is reported as destructive to clients (MCP/HTTP) just like a `destructive def`.
 
 ### `idempotent`
 
@@ -55,7 +74,7 @@ An `idempotent` tool stays callable after a failure (like the default), but the 
 |---|---|---|
 | *(unmarked)* | Yes | Re-callable, but not promised safe to repeat blindly. |
 | `idempotent` | Yes | Always safe to re-run. |
-| `destructive` | Only if it never started | Dangerous to repeat; locked out once it begins. |
+| `destructive` | Only if the body/region never started | Dangerous to repeat; locked out once a destructive region is entered. |
 
 These markers are about *retry safety* only. They are independent of [interrupts](/guide/interrupts), which gate whether a tool runs at all.
 
--- a/packages/agency-lang/docs/site/guide/llm-part-2.md
+++ b/packages/agency-lang/docs/site/guide/llm-part-2.md
@@ -33,28 +33,26 @@ destructive def chargeCard(amount: number): string {
 }
 ```
 
-A `destructive def` marks the **whole body** as destructive: entering it commits. If the tool then fails anywhere in its body, Agency removes it from the conversation. The model cannot call it again, and the tool result tells it so: the operation may have partially completed, so the state must be verified by hand rather than by retrying.
+A `destructive def` marks the whole body as destructive. Once the body starts, a failure anywhere removes the tool. The model cannot call it again: the operation may have half-finished, so its state needs a manual check.
 
-There is one exception. If the call fails *before the body starts* — the model passed the wrong arguments, or too few — then nothing ran yet, so the tool stays callable.
+One exception: a call that fails before the body starts (wrong or missing arguments) ran nothing, so the tool stays callable.
 
 #### `destructive { }` regions
 
-Marking a whole function `destructive` is often too coarse. Real tools validate their arguments and ask the user to confirm *before* doing anything dangerous — and rejecting that confirmation should leave the tool callable, not lock it out. Use a `destructive { }` region to mark only the effectful part, and leave the prep and the confirmation gate outside it:
+Marking a whole function is often too coarse. Most tools validate their arguments and ask for confirmation first. Rejecting that confirmation should leave the tool callable. Wrap only the effectful part in a `destructive { }` region, and keep the prep and the gate outside it:
 
 ```agency
 def write(filename: string, content: string): Result {
   const path = resolvePath(filename)
-  // Ask first — rejecting this is retryable, nothing has happened yet.
+  // Rejecting this gate is retryable: nothing has happened yet.
   return interrupt std::write("Write to this file?", { path: path })
-  // Only this commits. A failure here removes the tool; a rejected
-  // confirmation above does not.
   destructive {
     return try _write(path, content)
   }
 }
 ```
 
-Entering the region flips the tool into "committed": a failure inside it (or after it) removes the tool, exactly like a `destructive def`. A failure *before* the region — a validation refusal, or a rejected interrupt gate — is retryable. The region introduces no new scope; variables declared inside it are visible afterward. A function that contains a `destructive { }` region is reported as destructive to clients (MCP/HTTP) just like a `destructive def`.
+Entering the region commits the tool. A failure inside it, or after it, removes the tool. A failure before it stays retryable, whether a validation refusal or a rejected gate. The region adds no new scope, so a variable declared inside it is visible afterward. A function with a `destructive { }` region is reported as destructive to clients, the same as a `destructive def`.
 
 ### `idempotent`
 
--- a/packages/agency-lang/lib/backends/functionContainsDestructiveBlock.ts
+++ b/packages/agency-lang/lib/backends/functionContainsDestructiveBlock.ts
@@ -7,6 +7,15 @@ import { walkNodesArray } from "../utils/node.js";
  *  registry. NEVER for the entry flip, which keys on the raw `destructive def`
  *  marker alone.
  *
+ *  This answers only "does the function DECLARE its own destructive work" — via
+ *  a region here, or via the raw `destructive def` marker checked by the caller.
+ *  It deliberately does NOT treat a function that merely CALLS a `destructive`
+ *  function as destructive: that runtime taint is handled separately by Rule 2
+ *  (`NameClassifier.containsDestructiveCall` + the `destructiveFunctions`
+ *  registry), which folds a callee's `destructiveRan` into the caller's failure.
+ *  (Agency has no nested function definitions, so there is no destructive-marked
+ *  function to find inside a body.)
+ *
  *  Matches BOTH forms of the region so it works on either side of the
  *  parallelDesugar pass: pre-desugar it is a `seqBlock` flagged `destructive`
  *  (what `compilationUnit` sees); post-desugar the seqBlock is inlined and the
--- a/packages/agency-lang/lib/backends/functionContainsDestructiveBlock.ts
+++ b/packages/agency-lang/lib/backends/functionContainsDestructiveBlock.ts
@@ -1,0 +1,21 @@
+import type { AgencyNode, SeqBlock } from "../types.js";
+import { walkNodesArray } from "../utils/node.js";
+
+/** True if any `destructive { }` region appears anywhere in `body`, including
+ *  nested blocks/ifs/loops. Used for the is-destructive METADATA only — the
+ *  emitted descriptor marker, the MCP/HTTP hint, and the `destructiveFunctions`
+ *  registry. NEVER for the entry flip, which keys on the raw `destructive def`
+ *  marker alone.
+ *
+ *  Matches BOTH forms of the region so it works on either side of the
+ *  parallelDesugar pass: pre-desugar it is a `seqBlock` flagged `destructive`
+ *  (what `compilationUnit` sees); post-desugar the seqBlock is inlined and the
+ *  region survives as a `markDestructiveRan` leaf (what `TypeScriptBuilder`
+ *  sees). */
+export function functionContainsDestructiveBlock(body: AgencyNode[]): boolean {
+  for (const { node } of walkNodesArray(body)) {
+    if (node.type === "markDestructiveRan") return true;
+    if (node.type === "seqBlock" && (node as SeqBlock).destructive) return true;
+  }
+  return false;
+}
--- a/packages/agency-lang/lib/backends/typescriptBuilder/destructiveTracking.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder/destructiveTracking.ts
@@ -6,9 +6,10 @@ import type { NameClassifier } from "./nameClassifier.js";
 
 /**
  * Emits the destructive-execution tracking codegen: the per-function
- * `__destructiveRan` init, the function-exit boundary stamp, and the
- * per-statement flips. Extracted from TypeScriptBuilder so the tracking rules
- * live in one declarative place instead of scattered through a 4k-line file.
+ * `__destructiveRan` init, the function-exit boundary stamp, the region-entry
+ * flip (`blockEntryFlip`, emitted by a `markDestructiveRan` node from an inlined
+ * `destructive { }` region), and the Rule-2 caller-side flips. Extracted from
+ * TypeScriptBuilder so the tracking rules live in one declarative place.
  *
  * Stateless by design: every method is a pure function of its arguments (the
  * `inDestructiveFunction` flag is passed in, not held), so it can be
@@ -16,8 +17,13 @@ import type { NameClassifier } from "./nameClassifier.js";
  *
  * How the flag works: every function activation carries a boolean
  * `__self.__destructiveRan`. It starts false and is only ever flipped to true
- * (sticky). The function exit folds it into any departing failure via
- * `stampFailureBoundary`, so a failure reports whether destructive work ran.
+ * (sticky). It is set at ENTRY for a `destructive def` (init(true)), or at
+ * region entry for a `destructive { }` region. The function exit folds it into
+ * any departing failure via `stampFailureBoundary`, so a failure reports whether
+ * execution entered a destructive region. NOTE: because a `destructive { }`
+ * region is inlined into the function body before codegen, its flip lands on the
+ * function's `__self` — never a block frame — which is what keeps a failure
+ * escaping the function from carrying a false (fail-open) flag.
  */
 export class DestructiveTracking {
   constructor(
--- a/packages/agency-lang/lib/backends/typescriptBuilder/destructiveTracking.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder/destructiveTracking.ts
@@ -7,9 +7,9 @@ import type { NameClassifier } from "./nameClassifier.js";
 /**
  * Emits the destructive-execution tracking codegen: the per-function
  * `__destructiveRan` init, the function-exit boundary stamp, the region-entry
- * flip (`blockEntryFlip`, emitted by a `markDestructiveRan` node from an inlined
- * `destructive { }` region), and the Rule-2 caller-side flips. Extracted from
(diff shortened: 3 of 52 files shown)


packages/agency-lang/lib/preprocessors/parallelDesugar.ts
```diff
@@ -423,7 +423,13 @@ export function desugarParallelInBody(body: AgencyNode[]): AgencyNode[] {
       result.push(...desugarOneParallel(node));
     } else if (node.type === "seqBlock") {
       // Outside a parallel context, a seq block has no runtime effect; inline
-      // its body after recursively desugaring.
+      // its body after recursively desugaring. A `destructive` seqBlock also
+      // commits: prepend the flip so it runs on the ENCLOSING function's
+      // `__self` (this inlining is exactly why `__self` is the function, not a
+      // block frame — the fail-open guard).
+      if (node.destructive) {
```
egonSchiele, line 443: This branch (the desugarParallelInBody path) correctly prepends the flip for a top-level destructive seqBlock. But the sibling path handles it wrong: in `desugarOneParallel` (~L510-511), a `seqBlock` **arm** is inlined via `desugarParallelInBody([...arm.body])`, which ignores `arm.destructive` — so `parallel { destructive { rm() } }` drops the flip entirely (fail-open). And a destructive region nested inside a `seq` arm emits its flip onto the fork branch frame, not the caller. `seqBlock` is in `ALLOWED_ARM_TYPES`, so this is reachable. Suggest rejecting `destructive { }` inside a `parallel` block, or handling it deliberately; add a test.

egonSchiele (commented):
## Review

The **marked-`seqBlock`** approach is the right call — reusing an already-transparent, already-everywhere-wired block avoids threading a new node through ~8 hand-enumerated dispatch sites, and the fail-open guard (flip lands on the function `__self` because the seqBlock is inlined) is real and directly tested by `frameEscape → [true]`. The parser soft-gate, the `init`-threading bug fix (scope flag restored before `buildFunctionBody`), and the both-sides-of-desugar metadata helper are all correct. The `destructive-tracking` and `destructive-block` test suites are thorough.

Two must-fix items and one design edge before merge.

### 🔴 CI is red — and not for the reason the PR body states

The description says *"8163 passed… the one failure is a pre-existing worktree-naming artifact (`resolver.test.ts`)."* That's not what CI shows:

- **`build (22.x/23.x)` fails** on `typescriptGenerator.integration.test.ts > Fixture: 'destructiveBlock'`. The committed **`tests/typescriptGenerator/destructiveBlock.mjs` is stale** — it's missing *both* things this PR's codegen now emits:
  - `markers: { destructive: true }` in the tool descriptor, and
  - `__self.__destructiveRan = true;` (the region-entry flip).

  The generated output has them; the golden file doesn't. Regenerate with `make fixtures` and recommit. (The codegen is correct — this is a fixture that predates the final codegen.)

- **`agency-tests (22.x/23.x)` fails** too: the runner hit `FATAL ERROR: Reached heap limit — JavaScript heap out of memory`, then the job was canceled. Please confirm whether that's caused by this branch (the ~29-fn stdlib migration + new fixtures add compile load) or environmental/flaky, and re-run.

The mismatch between the PR body's "make all clean / 8163 passed" and the actual CI result suggests the full suite wasn't re-run from a clean checkout after the last codegen changes. Worth doing before merge, since the stale fixture is exactly the kind of thing a final `make fixtures && pnpm test:run` catches.

(Also seen in the agency-tests log, likely pre-existing but worth a glance: `AG2008` "narrow the value" diagnostics on `result-retry*.agency` / `result-restore-globals.agency`.)

### 🟠 `destructive { }` inside a `parallel` block — silent fail-open edge

`seqBlock` is in `ALLOWED_ARM_TYPES`, so `parallel { destructive { rm() } }` parses. But `desugarOneParallel` handles a `seqBlock` arm by inlining `arm.body` while **ignoring `arm.destructive`** — no `markDestructiveRan` is prepended, so the flip is dropped. And even a `destructive { }` nested *inside* a `seq` arm emits its flip onto the **fork branch's** `__self`, not the caller's function activation, so it can't commit to the enclosing tool. Net: destructive regions inside parallel/fork arms have untested, likely fail-open commit semantics — precisely the cardinal danger this PR targets. Since a region can't commit to the caller frame from inside a fork, I'd **reject `destructive { }` inside a `parallel` block** (a desugar/typecheck error) rather than accept it silently — or handle it deliberately. Either way, add a test pinning the intended behavior. Inline comment on the arm-handling below.

### 🟡 Coverage: the flagship stdlib behavior isn't directly tested

The whole point of the migration is: *reject a tool's interrupt gate (which sits before the region) → retryable*. `cleanRefusal` covers the plain-`return`-before-region variant, and an interrupt-reject follows the same "flip never runs" path — so it's transitively covered — but there's no direct test that rejecting an interrupt **gate placed before a `destructive { }`** yields `destructiveRan = false`/retryable. The plan's Task 14 Step 4 (extend `write-binary.test.json`) wasn't done — that file is untouched. Consider adding one end-to-end case so the migrated-tool contract is guarded directly.

packages/agency-lang/docs/site/guide/llm-part-2.md
```diff
@@ -33,9 +33,28 @@ destructive def chargeCard(amount: number): string {
 }
 ```
 
-If a `destructive` tool starts running and then fails, Agency removes it from the conversation. The model cannot call it again, and the tool result tells it so: the operation may have partially completed, so the state must be verified by hand rather than by retrying.
+A `destructive def` marks the **whole body** as destructive: entering it commits. If the tool then fails anywhere in its body, Agency removes it from the conversation. The model cannot call it again, and the tool result tells it so: the operation may have partially completed, so the state must be verified by hand rather than by retrying.
```
egonSchiele: Make your changes to this file shorter and more readable. Do not include any extra information. Use these writing tips:

<tips>
## General writing tips

1. Avoid garden path sentences.

Bad: a sentence that begins like this:

```
The name lists partition
```

When you read the sentence, what do you think it means? I think it means that a name lists some partitions. Here is the full sentence:

```
The name lists partition the catalog's managed (`source:"remote"`) entries by what changed.
```

It turns out that there is something called a "name list", which partitions something in the catalog. Avoid sentences like this, because they make the reader realize halfway through reading the sentence that it doesn't that the words don't mean what the reader thought they meant, and they have to re-read the sentence to take into account the information they gained in the second half of the sentence. A classic garden path sentence begins with the words "the old man". What do you think it's going to tell us about the old man? Actually, the full sentence is "the old man the boat". Notice how the first half of the sentence makes you think it's talking about an old man, and it's not until you read the second half that you realize that it's talking about elderly people as a group.

Good: 

```
The catalog's entries are partitioned by name lists.
```

2. Use active voice, avoid passive voice.

Bad: "The man was bitten by the dog."
Good: "The dog bit the man."

Passive voice makes you focus on the person or object receiving the action, rather than the person or object performing it. Passive voice is typically harder to read and less interesting.

Other examples:

Bad: "Enclosing guard(cost:) budgets meter the subprocess in real time"
Good "The subprocess is metered in real time by enclosing guard(cost:) budgets."

Bad: "filename cannot escape dir via absolute paths or .. segments (symlinks are followed and re-checked)."
Good: "Absolute paths and .. segments cannot escape dir. Symlinks are followed and re-checked."

3. Avoid introducing too many concepts in a single sentence.

If you find yourself using multiple commas, semicolons, or conjunctions in a sentence, consider breaking it up into multiple sentences.

Bad:
```
The name lists partition the catalog's managed (`source:"remote"`) entries by what changed.
```

Notice how the sentence tries to squeeze in the `source:"remote"` qualifier to an already complicated sentence. If you need to parenthesize something in your sentence, consider whether it would be clearer as two sentences instead.

```
Parse command-line flags with strict number coercion, required flags and defaults, mutually-exclusive groups, and auto-generated `--help` / `--version`.
```

This sentence introduces so many things that by the time the reader has reached the end of the sentence, they have already forgotten the beginning of the sentence.

```
any interrupt no handler resolves surfaces to the user; responding resumes the subprocess exactly where it paused.
```

Here is an example where two sentences would have been clearer than one with a semicolon.

4. Emdash overuse

An emdash used sparingly can make the text more lively, but used all the time, it becomes jarring. If you are using an emdash, consider using a comma instead.

Bad: "The AST shape is the parser output, which matches what the formatter consumes — so an AST round-tripped through writeAST() produces canonical Agency source."
Good: "The AST shape is the parser output, which matches what the formatter consumes, so an AST round-tripped through writeAST() produces canonical Agency source."

Or use a conjunction instead of an emdash:
Bad: "Read and write happen inside the same interrupt — approving it approves both."
Good: "Read and write happen inside the same interrupt, so approving it approves both."

5. Consider whether this level of detail is actually required in the sentence.

You have a habit of trying to add more information in parentheses. Often, it's not needed, and actually hinders understanding by making the sentence longer.

Bad: "Exceeding a resource limit (wallClock, memory, ipcPayload, or stdout) kills the subprocess and returns a limit_exceeded failure."
Good: "Exceeding a resource limit kills the subprocess and returns a limit_exceeded failure."

Every time you add something in parentheses, consider that the user's brain is going to need to pause and add a new frame to their mental "stack" to capture this new information. Is that speed bump worth the information you're trying to convey?


</tips>

packages/agency-lang/lib/backends/typescriptBuilder/destructiveTracking.ts
```diff
@@ -33,13 +39,28 @@ export class DestructiveTracking {
...
+        ? ts.bool(true)
+        : ts.binOp(ts.self("__destructiveRan"), "??", ts.bool(false)),
     );
   }
 
+  /** `__self.__destructiveRan = true` — the entry flip a `destructive { }`
+   *  region emits (via the `markDestructiveRan` node). Because the region is
+   *  inlined into the function body, `ts.self` resolves to the function's
+   *  `__self`, not a block frame. */
+  blockEntryFlip(): TsNode {
```
egonSchiele: These function names are not very descriptive. "Block entry flip" does not explain what you're flipping. Also, it seems like if all this function is doing is calling another function, users of this function could just call the markTrue function directly. Again, `markTrue` doesn't explain what it is marking true.

packages/agency-lang/lib/backends/functionContainsDestructiveBlock.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  marker alone.
+ *
+ *  Matches BOTH forms of the region so it works on either side of the
+ *  parallelDesugar pass: pre-desugar it is a `seqBlock` flagged `destructive`
+ *  (what `compilationUnit` sees); post-desugar the seqBlock is inlined and the
+ *  region survives as a `markDestructiveRan` leaf (what `TypeScriptBuilder`
+ *  sees). */
+export function functionContainsDestructiveBlock(body: AgencyNode[]): boolean {
+  for (const { node } of walkNodesArray(body)) {
+    if (node.type === "markDestructiveRan") return true;
```
egonSchiele, line 26: Do we also need to  check whether it contains a function with a destructive modifier or is that case already covered?

packages/agency-lang/lib/preprocessors/parallelDesugar.ts
```diff
@@ -423,7 +423,13 @@ export function desugarParallelInBody(body: AgencyNode[]): AgencyNode[] {
       result.push(...desugarOneParallel(node));
     } else if (node.type === "seqBlock") {
```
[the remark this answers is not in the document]
egonSchiele: Confirmed and fixed. \`seqBlock\` is in \`ALLOWED_ARM_TYPES\` and \`desugarOneParallel\` splices \`arm.body\`, so \`arm.destructive\` (and any region nested in a \`seq\` arm) drops its flip onto the fork-branch frame, never the caller — a silent fail-open. Rather than special-case commit-through-a-fork, \`validateParallelBlock\` now rejects any \`destructive { }\` region anywhere inside a \`parallel\` block (full-subtree walk via \`functionContainsDestructiveBlock\`). Tests added for a direct arm and a region nested in a \`seq\` arm.

packages/agency-lang/stdlib/shell.agency
```diff
@@ -83,26 +83,28 @@ export destructive def exec(
...
+      command,
+      args,
+      cwd,
+      timeout,
+      stdin,
+      {
+      allowedExecutables: allowedExecutables,
+      blockedCommands: blockedCommands,
+      allowedPaths: allowedPaths
+    },
```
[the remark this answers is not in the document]
egonSchiele: Fixed — the policy object literal is now indented one level under its opening brace in both \`exec\` and \`bash\`.

packages/agency-lang/stdlib/shell.agency
```diff
@@ -132,16 +134,18 @@ export destructive def bash(
...
+  destructive {
+    return _bash(
+      command,
+      cwd,
+      timeout,
+      stdin,
+      {
+      blockedCommands: blockedCommands,
+      allowedPaths: allowedPaths
+    },
```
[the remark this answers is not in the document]
egonSchiele: Fixed (same change as \`exec\`).

packages/agency-lang/lib/preprocessors/parallelDesugar.ts
egonSchiele: Fixed by rejecting. \`validateParallelBlock\` now throws for any \`destructive { }\` region inside a \`parallel\` block (walks the whole subtree, so both a direct arm and a region nested in a \`seq\` arm are caught). Tests pin both cases. Chose reject over handle because a region cannot commit to the caller activation from inside a fork branch.

packages/agency-lang/docs/site/guide/llm-part-2.md
egonSchiele: Shortened: split into short single-concept sentences, active voice, removed the em-dashes and the redundant explanation. Trimmed the code comment to one line.

packages/agency-lang/lib/backends/typescriptBuilder/destructiveTracking.ts
egonSchiele: Renamed \`markTrue\` -> \`markDestructiveRan\` (states what is set true) and removed \`blockEntryFlip\` — codegen now calls \`markDestructiveRan()\` directly, and the Rule-2 pre-flip uses the same method.

packages/agency-lang/lib/backends/functionContainsDestructiveBlock.ts
egonSchiele: No — that case is intentionally out of scope here. This helper answers only "does the function DECLARE its own destructive work" (a region here; the raw \`destructive def\` marker is checked by the caller). A function that merely CALLS a \`destructive\` function is handled separately by Rule 2 (\`containsDestructiveCall\` + the \`destructiveFunctions\` registry), which folds the callee's \`destructiveRan\` into the caller's failure at runtime — it does not make the caller a destructive TOOL for metadata/hint purposes. (Agency has no nested function definitions, so there is no destructive-marked function to find inside a body.) Added this to the doc comment.

egonSchiele:
Addressed all four points; pushed in `09f0d90e3`.

**🔴 Stale fixture — fixed.** You were right: the committed `destructiveBlock.mjs` predated the Task-6 metadata codegen — it was missing both `markers: { destructive: true }` and the region-entry flip. Regenerated with `make fixtures`. Root cause of the PR-body mismatch: the local `typescriptGenerator.integration` test passed against the *working-tree* file (which `make fixtures` had regenerated) while the *committed* file was stale, so my local run was green against an uncommitted fixture. My mistake — I've re-run from the committed state now.

**🔴 agency-tests OOM — determined: not this branch's logic.** The FATAL heap error hits mid-suite in `agency-tests`, which compiles *every* agency file — a job with prior OOM history (#381). This branch adds only 2 small test files + 2 fixtures + in-place stdlib edits (no new module-global state, no unbounded walk — the region checks are all O(nodes)). Main's Node CI was green on the #533 merge, so it's not a permanent main failure, but it also isn't caused by this branch's code — it's the suite bumping the heap ceiling. Per @egonSchiele's note that the main-side CI is being fixed separately, I've left the job's memory config alone. If you'd prefer, I can bump `NODE_OPTIONS=--max-old-space-size` for that job in a separate change.

**🟠 `destructive { }` in `parallel` — rejected.** Confirmed the fail-open: `desugarOneParallel` splices `arm.body` and ignores `arm.destructive`, so the flip lands on the fork-branch frame. `validateParallelBlock` now throws for any region inside a `parallel` block (full-subtree walk, so a region nested in a `seq` arm is caught too). Tests added for both.

**🟡 Direct migrated-tool test — added.** `rejectGateBeforeRegion` in `destructive-block`: an `interrupt` gate before a `destructive { }`, rejected → `destructiveRan = false` (retryable), asserting the exact contract the migration exists for.

Also noted the `AG2008` "narrow the value" diagnostics on `result-retry*` / `result-restore-globals` — those are on files this branch doesn't touch, so I've left them for the separate CI fix.

egonSchiele (commented):
## Re-review (commit `09f0d90`)

The round-1 findings are all addressed, and correctly:

- **🟠 → ✅ Parallel fail-open.** `validateParallelBlock` now rejects any destructive region up front via `functionContainsDestructiveBlock(pb.body)`. Because that helper walks the whole subtree, it catches **both** a `destructive { }` used as a direct arm **and** one nested inside a `seq` arm — exactly the two cases I raised — and both are pinned by tests. The rejection is also correctly scoped: it matches only *regions*, so a plain `seq {}` arm still passes (no over-rejection). The error message points the user at the fix. Good call rejecting rather than trying to make fork-branch commit semantics work.

- **🟡 → ✅ Migrated-tool contract.** `rejectGateBeforeRegion` is the exact test that was missing: an interrupt gate *before* the region, rejected → `destructiveRan = false` → retryable. That's the stdlib contract guarded directly now.

- The `markTrue`/`blockEntryFlip` → `markDestructiveRan()` consolidation reads cleanly.

Nits only (non-blocking):

1. **Layering.** `functionContainsDestructiveBlock` lives in `lib/backends/` but is now imported by a preprocessor (`parallelDesugar`), which runs *before* the backend. No cycle (it only pulls `types` + `utils/node`), but it's a pure AST predicate — `lib/utils/` would be a more natural home and avoids the preprocessor→backend inversion.
2. Stray double blank line left in `destructiveTracking.ts` (~L55) where `blockEntryFlip` was removed.
3. `tests/agency/destructive-block.test.json` lost its trailing newline.

One open question (pre-existing, out of scope — not asking you to fix here): the new guard rejects destructive *regions* inside `parallel`, but a `destructive def` **call** in a parallel arm has the same fork-branch-frame concern. Does the callee's `destructiveRan` fold up to the parent tool at fork join (the way the `handle`-block join fold does in the `blockJoin` test), or is that path also fail-open? Worth confirming at some point, but it's separate from this PR's region work.

Nice work — modulo the CI/fixture item from round 1, this looks good to me.

PR #537, opened 2026-07-13
Fix review() crash and unblock verify: review .agency files, not the chat reply

egonSchiele:
## Problem

The main `agency agent` (`--agent code`) failed most **non-Agency** tasks (Python/C/shell) because of a harness bug, not model capability. Root-caused from Terminal-Bench transcripts:

1. The one-shot loop (`code.agency`) called `review(reply)` on the model's chat reply each iteration.
2. `review()` asked an LLM to "extract Agency snippets" from the reply, then fed each to the Agency parser/typechecker.
3. On a Python/shell reply that parse **crashed** and surfaced as `Parse error: Cannot read properties of undefined` / `Typecheck failed`. `feedbackHasErrors` treated the crash as "your code has errors."
4. The agent, which wrote correct Python, was told to "fix" nonexistent errors and **thrashed** (observed: 76/70/152 phantom re-prompts across three tasks).
5. The fresh-eyes `verify` step lives in the `else` branch that only runs when `review()` finds no errors, so on non-Agency tasks **`verify` ran 0 times**.

## Fix

Review the `.agency` **files the agent changed on disk** (found via `std::git`), never the chat text.

- **`review.agency`** — `filterAgencyPaths`, `changedAgencyFiles` (fails open to `[]` on a non-git dir), `reviewFiles` (reads each path fail-open, deterministic parse+typecheck, no LLM), and `reviewWrittenFiles` (the code-agent entry point).
- **`code.agency`** — calls `reviewWrittenFiles() with approve` instead of `review(reply)`. For a non-Agency task no `.agency` files change, so it returns `success([])`, `feedbackHasErrors` is false, and the loop **falls through to `verify`** — the unblock. Genuine `.agency` errors still surface.
- **`reviewAgent` / `review(userMsg)`** (the `--agent review` snippet path) are unchanged.

## Also: git result types are now declared natively in Agency

`changedAgencyFiles` needs to read `status.entries` off a `GitStatus`, but `GitStatus` was **imported from the TS git module**, and Agency does not parse TypeScript — a TS-imported type is an opaque name with **no accessible fields**, so `status.entries` (and `status.branch`, `gitDiff().files`, `gitLog().commits`, ...) all error with `AG2011: Property does not exist`. (The `status.branch` "example" in `git.agency` only appears inside a doc comment, so it was never compiled.)

The correct fix is to **declare the git result types natively in `git.agency`** (mirroring the shapes the TS parsers return); `return parseStatus(...)` bridges the opaque TS return to the native type exactly as `bash` returns a natively-declared `ExecResult` from a TS `_bash` impl. This makes field access work, fixes the same latent opacity for every git result type, and documents them in the stdlib reference (`git.md`). No other Agency code imported these types, so the change is self-contained; `gitCore.ts` keeps the TS types for the parsers/runtime.

## Tests

- **`tests/agency/review-agent.agency`** (deterministic, no LLM, 9 cases): `filterAgencyPaths` (keeps only `.agency`, rejects mid-name substrings, empty input); `reviewFiles` (no files → `success([])`; broken `.agency` → error finding; valid file → no false positive); `changedAgencyFiles` (non-git dir → `[]`; git repo with a changed `.agency` + `.py` → only the `.agency`, via native `status.entries` access). **9/9 pass.**
- Existing **`gitCore.test.ts` (23/23)** and **`gitPolicy.agency` (3/3)** still pass — the native types are structurally identical to the TS ones, so runtime is unchanged.

These deterministic tests replace the plan's single real-LLM regression as the PR-tier guard (that test does not run in CI and asserted on substring-absence).

## Notes for the reviewer

- **`read()` returns `Result`**, so `reviewFiles` unwraps it with `if (r is success(source))` before passing the string to `agencyReview`.
- **Effect propagation:** `gitStatus` raises `std::git::status` and `read` raises `std::read`, so `changedAgencyFiles`/`reviewFiles`/`reviewWrittenFiles` declare `raises`, discharged with `with approve` at the code-agent call site (mirroring `skillsDir(...) with approve`).
- **Commit history:** commit 1 first added a `changedFilePaths` TS helper as a workaround; the final commit reverts it in favor of the native-types fix above (kept as separate commits rather than a force-push).

## Also: cap skill tool names (unblocks the e2e)

`skillsDir` / `docsSkill` derive the LLM tool name from the absolute skills directory path (`skills_<path>`). A deep checkout path (a git worktree under `.claude/worktrees/...`) produced a name over the provider tool-name limit (OpenAI 64, Anthropic 128), so **every** agent run from such a path failed with `tools.N.custom.name: String should have at most 128 characters` before doing any work. Fixed by capping the derived name to 64 (keeping the distinctive tail, which preserves uniqueness for sibling dirs) and adding an optional explicit `name` param to `skillsDir`.

## Verification status

- ✅ `make` clean.
- ✅ Deterministic: `review-agent.agency` **9/9**, `skills-dir.agency` **13/13** (5 new), `gitCore.test.ts` **23/23**, `gitPolicy.agency` **3/3**.
- ✅ **Real-LLM end-to-end** (code agent, Python task, temp git repo): `hello.py` written correctly (`print(42)`), **0** phantom "last code had errors" re-prompts, only **9** prompts (was 76–152 pre-fix), and the **verify thread ran** (`label:verify` in the statelog, was 0×). The skill-name cap is what unblocked this run.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/subagents/review.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/subagents/review.agency
@@ -6,10 +6,7 @@ import {
  } from "std::agency"
 import { setMemoryId } from "std::memory"
 import { gitStatus } from "std::git"
-import { filter } from "std::array"
-// `changedFilePaths` projects a GitStatus to its changed paths in TS: Agency
-// code cannot read the TS-imported GitStatus record's fields directly.
-import { changedFilePaths } from "agency-lang/stdlib-lib/gitCore.js"
+import { filter, map } from "std::array"
 
 // The review core (parse + typecheck findings, optional LLM code-vs-task
 // judgment, rendering helpers) lives in std::agency now — this subagent
@@ -94,7 +91,7 @@ export def changedAgencyFiles(): string[] raises <std::git::status> {
   // handler (the agent policy auto-approves this read effect).
   const statusResult = try gitStatus()
   if (statusResult is success(status)) {
-    return filterAgencyPaths(changedFilePaths(status))
+    return filterAgencyPaths(map(status.entries, \entry -> entry.path))
   }
   return []
 }
--- a/packages/agency-lang/lib/agents/agency-agent/subagents/review.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/subagents/review.agency
@@ -5,6 +5,11 @@ import {
   feedbackHasErrors as stdFeedbackHasErrors,
  } from "std::agency"
 import { setMemoryId } from "std::memory"
+import { gitStatus } from "std::git"
+import { filter } from "std::array"
+// `changedFilePaths` projects a GitStatus to its changed paths in TS: Agency
+// code cannot read the TS-imported GitStatus record's fields directly.
+import { changedFilePaths } from "agency-lang/stdlib-lib/gitCore.js"
 
 // The review core (parse + typecheck findings, optional LLM code-vs-task
 // judgment, rendering helpers) lives in std::agency now — this subagent
@@ -65,4 +70,54 @@ export def review(userMsg: string): Result<Feedback[]> {
     }
   }
   return feedback
+}
+
+// ---- The code agent's auto-review: review the .agency FILES it wrote, not
+// the model's chat reply. Feeding a Python/shell chat reply to the Agency
+// parser crashed and mis-fired as "your code has errors", which thrashed the
+// agent and shadowed the `verify` step. Reviewing on-disk .agency files means
+// the parser only ever sees Agency source, so a non-Agency task is a clean
+// no-op (empty findings) and the loop falls through to `verify`.
+
+/** Keep only the .agency files from a list of paths. */
+export def filterAgencyPaths(paths: string[]): string[] {
+  return filter(paths, \path -> path.endsWith(".agency"))
+}
+
+/** The .agency files changed in the agent working directory, or [] when the
+  working dir is not a git repo (or git otherwise fails). Fails open — a
+  non-repo/error yields no files so the auto-review is a no-op, never a crash.
+  Includes untracked new files (they are what a coding agent just wrote). */
+export def changedAgencyFiles(): string[] raises <std::git::status> {
+  // `try` folds a non-repo failure (git status exits non-zero) to a Result we
+  // read as "no changed files". The interrupt propagates to the caller's
+  // handler (the agent policy auto-approves this read effect).
+  const statusResult = try gitStatus()
+  if (statusResult is success(status)) {
+    return filterAgencyPaths(changedFilePaths(status))
+  }
+  return []
+}
+
+/** Review the .agency source at each given path, merging the findings. Reads
+  each file against the agent cwd; an unreadable path (e.g. a file `git status`
+  lists as deleted) is skipped, so a stale path can never crash the review. No
+  `task` is passed, so this is deterministic parse+typecheck with no LLM call. */
+export def reviewFiles(paths: string[]): Result<Feedback[]> raises <std::read> {
+  let feedback: Result<Feedback[]> = success([])
+  for (path in paths) {
+    const sourceResult = read(filename: path, useAgentCwd: true)
+    if (sourceResult is success(source)) {
+      feedback = mergeFeedback(feedback, agencyReview(source))
+    }
+  }
+  return feedback
+}
+
+/** The code agent's per-iteration auto-review. Reviews the .agency files the
+  agent changed this run (found via std::git), never the chat reply. Returns
+  success([]) when no .agency files changed — the common case for Python / C /
+  shell tasks — so `feedbackHasErrors` is false and the loop reaches `verify`. */
+export def reviewWrittenFiles(): Result<Feedback[]> raises <std::git::status, std::read> {
+  return reviewFiles(changedAgencyFiles())
 }
--- a/packages/agency-lang/docs/site/stdlib/git.md
+++ b/packages/agency-lang/docs/site/stdlib/git.md
@@ -28,6 +28,156 @@ Typed, safe git tools for agents. Each tool builds its own git command
 
 ## Types
 
+### ChangeCode
+
+git porcelain change codes: "." unmodified, "M" modified, "A" added,
+  "D" deleted, "R" renamed, "C" copied, "U" unmerged, "T" type-changed,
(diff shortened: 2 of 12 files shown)


packages/agency-lang/lib/agents/agency-agent/subagents/review.agency
```diff
@@ -66,3 +71,53 @@ export def review(userMsg: string): Result<Feedback[]> {
...
+  working dir is not a git repo (or git otherwise fails). Fails open — a
+  non-repo/error yields no files so the auto-review is a no-op, never a crash.
+  Includes untracked new files (they are what a coding agent just wrote). */
+export def changedAgencyFiles(): string[] raises <std::git::status> {
+  // `try` folds a non-repo failure (git status exits non-zero) to a Result we
+  // read as "no changed files". The interrupt propagates to the caller's
+  // handler (the agent policy auto-approves this read effect).
+  const statusResult = try gitStatus()
+  if (statusResult is success(status)) {
+    return filterAgencyPaths(changedFilePaths(status))
```
egonSchiele: **Turn-scoping was dropped — this reintroduces the thrash + verify-shadow in real repos.** This returns *all currently-dirty* `.agency` files, but the revised spec (`§Root fix`) requires reviewing only what changed *this run*: `current git-dirty − a task-start baseline`.

Concrete failure: an interactive user has a pre-existing uncommitted `helper.agency` with a type error (WIP) and asks the agent to edit a Python file. The agent writes `main.py` (no `.agency` change), but this still returns `helper.agency` → `reviewFiles` flags its error → `feedbackHasErrors` true → "The last code had errors… please fix them" → the agent thrashes on a file the user never touched, and `verify` is shadowed. That's exactly the bug this PR fixes, via a different trigger. A fresh Terminal-Bench container hides it (clean start == turn changes), which is why the tests are green.

Fix (per the plan): capture `baseline = changedAgencyFiles()` once at `codeAgent` entry and review `current − baseline` (a `subtractBaseline(current, baseline)` helper). Note the docstring above ("changed in the agent working directory") reads as this-run but is actually all-dirty — tighten it too.

packages/agency-lang/lib/agents/agency-agent/subagents/review.agency
```diff
@@ -66,3 +71,53 @@ export def review(userMsg: string): Result<Feedback[]> {
...
+  }
+  return []
```
egonSchiele, line 96: Failing open to `[]` is correct, but silent. If `gitStatus` starts failing for a real reason (git not on PATH, corrupt repo, permissions), the auto-review becomes a permanent no-op and nobody notices. The plan asked for a `std::statelog` breadcrumb in each fail-open arm so a silently-disabled review is diagnosable — keep failing open, but leave a trace. (Same point applies to the read-skip in `reviewFiles`, L110.)

packages/agency-lang/lib/agents/agency-agent/subagents/review.agency
```diff
@@ -66,3 +71,53 @@ export def review(userMsg: string): Result<Feedback[]> {
...
+}
+
+/** Review the .agency source at each given path, merging the findings. Reads
+  each file against the agent cwd; an unreadable path (e.g. a file `git status`
+  lists as deleted) is skipped, so a stale path can never crash the review. No
+  `task` is passed, so this is deterministic parse+typecheck with no LLM call. */
+export def reviewFiles(paths: string[]): Result<Feedback[]> raises <std::read> {
+  let feedback: Result<Feedback[]> = success([])
+  for (path in paths) {
+    const sourceResult = read(filename: path, useAgentCwd: true)
```
egonSchiele, line 106: Path-base mismatch worth confirming: `git status` porcelain lists paths relative to the **repo root**, but `read(useAgentCwd: true)` resolves against `getAgentCwd()`. When the agent cwd *is* the repo root (Terminal-Bench `/app`) this is fine; if the agent runs in a **subdirectory** of the repo, the repo-root-relative paths won't resolve and every read silently fails → the file is skipped → the review no-ops on real changes. Either confirm the base always coincides, or resolve the paths against the repo dir `gitStatus` used.

Also: this silent read-skip is the second fail-open arm that should leave a `std::statelog` breadcrumb (see L99).

egonSchiele (commented):
## Review — fix `review()` crash / unblock `verify` (sub-project C)

Solid, honest PR. The core strategy is right: reviewing the on-disk `.agency` files (via `std::git`) instead of the chat reply removes the crash vector, and the swap so `success([])` falls through to `verify` is the minimal correct unblock. The deterministic test suite (9 agency + 3 TS) is a real improvement over the plan's single real-LLM test, and the deviation notes are excellent.

**But there's one correctness regression that needs addressing before merge, plus two robustness nits.**

### 🔴 Must-fix: turn-scoping was dropped (reintroduces the thrash/shadow bug)
`changedAgencyFiles()` reviews **all currently-dirty** `.agency` files, but the revised spec (`§Root fix`) made turn-scoping mandatory — review only files changed **this run** (`current git-dirty − a task-start baseline`). In a fresh Terminal-Bench container these coincide (clean start), which is why the tests pass; in a real repo with a pre-existing dirty `.agency` file it re-triggers the exact bug this PR fixes. Details inline on `changedAgencyFiles`.

### 🟡 Should-fix
- **No breadcrumb on the fail-open arms** (`changedAgencyFiles` `[]`, `reviewFiles` read-skip): a silently-disabled review is undiagnosable. The plan asked for a `std::statelog` breadcrumb. Inline.
- **`git status` paths vs `read(useAgentCwd:)` base**: git porcelain paths are repo-root-relative; `read` resolves against the agent cwd. Fine when they coincide (`/app`), broken in a subdir. Inline.

### 👍 Keep
- The `changedFilePaths` TS helper is the right workaround for **GitStatus being opaque to Agency** (`AG2011` on TS-imported record fields) — good discovery, and worth the broader follow-up you flagged.
- `read()` → `Result` unwrap, `raises` + `with approve` discharge, the substring-rejection / valid-file / non-git test cases.

### Follow-ups worth filing (not blockers, correctly out of scope here)
- TS-imported record field access silently unusable in Agency (`AG2011`).
- The `docsSkill` tool-name-length blocker from deep worktree paths (real; blocks the e2e run — the deterministic tests substitute well here).

Not blocking the e2e gap — `reviewFilesEmpty` + `nonGitReturnsEmpty` already cover the unblock deterministically.

PR #539, opened 2026-07-13
Agent refactor and fix

egonSchiele:
## Overview

A refactor of the agency agent plus a new plan-as-code capability. Three things ship together:

1. **The monolithic `agent.agency` is decomposed** into focused, layered `lib/*.agency` modules.
2. **`plannerAgent`** — a "the agent *is* the plan" meta-agent — lands in `std::agents`, and the code agent can now route or escalate complex tasks to it.
3. **Reliability fixes** to `std::git` (non-repo handling), `runCode` (working directory), and the auto-review path.

## 1. Agent decomposition

`agent.agency` shrinks from ~1600 lines to a thin entry point, with its responsibilities extracted into single-purpose modules under `lib/agents/agency-agent/lib/`:

- `args.agency` — CLI flag parsing (`parseAgentArgs`)
- `session.agency` — in-session `/model` switching and model pickers
- `models.agency` / `header.agency` — slot resolution and the startup header
- `capabilitiesUi.agency` — the `/settings` capability-profile UI
- `mcp.agency` — MCP server wiring and the `/mcp` command
- `policy.agency` — `--policy` help and built-in policy listing
- `search.agency` — web-search backend selection (with untrusted-value coercion)
- `commandPalette.agency` — slash-command palette (built-in + project `.claude/commands`)
- `images.agency` — image-tool input resolution against the agent cwd
- `utils.agency` / `utils/renderLLMCallResponse.agency` — retry-reason labels and LLM-response rendering

The modules are deliberately **layered so nothing forms an import cycle** — the UI/command layers sit above the resolution/settings layers and only import downward. The system/main prompts (`prompts/*.md`) are split into large/small variants to match.

## 2. Plan-as-code: `plannerAgent`

New `stdlib/agents/planner.agency`. Given a task, it drafts a plan, uses `writeAgency` to **generate an Agency program (a `node main`) that composes the worker agents and tools**, shows the plan, generated source, and capability envelope (`getEffects`) for approval, runs it in a handler-sandboxed subprocess via `runCode`, verifies the result, and falls back to `codingAgent` on generation failure or repeated verification gaps.

- Approval is a real interrupt (`std::agents::planApprove`): interactive prompts the user; one-shot lets the CLI policy decide. Reject halts and fails cleanly.
- Recursion (a generated agent calling `plannerAgent`) is bounded by the runtime's subprocess `maxDepth` and by `maxCost` (default `$100`) / `maxTime` (default `60m`).
- State is threaded through plain helpers (`planStep`/`runApproved`/`afterRun`/`judgeAfterRun`) so no binder-`match` sits in the stepped loop.

**Code-agent integration** (`subagents/code.agency`): a lightweight `triage` classifies each task as `simple` or `complex`. Complex tasks route straight to `plannerAgent` (which shows a plan first); simple tasks stay on the direct tool loop but gain an `escalate` tool so a task that turns out to need a plan can switch mid-flight. Escalations run under a tighter budget (`$20` / `15m`) than the planner's own defaults. `routeFor` is a pure, separately-tested routing decision.

Supporting stdlib changes: `std::agents/shared` gains `unwrapGuard` (shared by coding/research/planner); `std::agency`'s `runCode` gains an optional `cwd`.

## 3. Reliability fixes

- **`std::git` non-repo handling** (`lib/stdlib/git.ts`, `stdlib/git.agency`): every git tool routes through `gitRunImpl`, which now normalizes any "not a git repository" failure into one clear message — `current directory is not a git repository` — instead of git's raw `fatal:` string. New `gitIsRepo(cwd)` tool (backed by `_gitIsRepo`, which never throws).
- **Auto-review guard** (`subagents/review.agency`): `changedAgencyFiles()` skips via `gitIsRepo()` when the working dir is not a repo — the common case — so the per-iteration review is a clean no-op instead of a caught git failure.
- **`runCode` working directory** (`lib/runtime/ipc.ts`): the generated module is now materialized under the agency package root (resolved from `__dirname`) rather than `process.cwd()`, so a subprocess launched from a temp dir with no `node_modules` can still resolve `agency-lang`. The fork's cwd stays the caller's directory, so file writes land where the caller expects. This fixes `runCode` for every caller, including `plannerAgent` and the terminal-bench `/app` sandbox.
- **`std::shell` cwd threading** (`stdlib/shell.agency`): the exec/bash/ls/grep/glob effects carry a `cwd`, resolved through `applyAgentCwd`.

## Tests & docs

- New agency execution tests (no LLM): `plannerHelpers`, `plannerReject`, `routing`, `unwrapGuard`, plus refreshed agent-turn / settings / attachments tests.
- Generated stdlib docs regenerated for `agency`, `agents/planner`, `agents/shared`, `git`, `shell`; `docs/dev/terminal-bench.md` updated.


packages/agency-lang/stdlib/agents/shared.agency
```diff
@@ -1,9 +1,21 @@
...
+import { GuardFailureData } from "std::thread"
 
 /** Fold optional context material into a task prompt. Returns the task
   unchanged when no context is given. */
 export def withContext(task: string, context: string): string {
   return if context == "" then task else "${task}\n\n<context>\n${context}\n</context>"
 }
+
+/** Unwrap a `guard(...)` Result to the string an agent returns: the block's
+  value on success, or a "<label> stopped..." message when the cost or time cap
```
egonSchiele: This bare `<label>` is what breaks the CI docs build: `agency doc` emits it into `docs/site/stdlib/agents/shared.md` as prose, and VitePress/Vue reads `<label>` as an unclosed HTML tag (`shared.md (38:30): Element is missing end tag`). Reword to avoid the angle brackets — e.g. `a "[label] stopped..." message` — then regenerate docs and confirm `make` is clean.

packages/agency-lang/lib/agents/agency-agent/lib/policy.agency
```diff
@@ -0,0 +1,36 @@
+import { chooseOption } from "std::ui"
```
egonSchiele: This module references `BUILTIN_POLICIES`, `minimalAutoApprovePolicy`, `recommendedAutoApprovePolicy` (`std::policy`), `POLICY_PATH` (`./config.agency`), and `ChoiceItem` (`std::ui`) but imports none of them. It typechecks only because the directory compile seeds the symbol table from `agent.agency`; at runtime `policy.js` is its own module, so these are undefined — `printPolicyHelp()` throws on `for (p in BUILTIN_POLICIES)` and `givePolicyChoice()` matches over nothing. Add the imports here.

egonSchiele (commented):
## Review: Agent refactor + plan-as-code

Solid PR overall. The `agent.agency` monolith is broken into well-bounded modules with explicit "sits above X, nothing imports back" layering notes, the `unwrapGuard` extraction de-duplicates the guard-unwrap wording across the three workers, and the new no-LLM tests (routing, unwrapGuard, plannerHelpers, header lines) are good coverage. A few things to fix before merge, one of which is breaking CI.

### 🔴 Blocker — CI build is red

`build (22.x)` and `build (23.x)` fail. The fatal error is the VitePress docs build choking on a generated doc:

```
Build failed in 2.63s
build error: SyntaxError: [plugin vite:vue] stdlib/agents/shared.md (38:30): Element is missing end tag.
```

Root cause: the `unwrapGuard` docstring in `stdlib/agents/shared.agency` contains a bare `<label>`:

> ... or a `"<label> stopped..."` message when the cost or time cap tripped.

`agency doc` emits that verbatim into `docs/site/stdlib/agents/shared.md` as prose, and VitePress/Vue parses `<label>` as an unclosed HTML tag. Fix: reword the docstring to avoid angle brackets (e.g. `a "[label] stopped..." message`, or wrap the placeholder in backticks). Then regenerate docs (`agency doc`) and confirm `make` is clean. The `Result<...>` generics in the other generated docs are inside fenced code blocks, so they're safe — only the prose `<label>` breaks it. (I left an inline comment at the source line.)

### 🟠 High — `lib/policy.agency` is missing its imports

`printPolicyHelp` / `givePolicyChoice` were moved out of `agent.agency` into the new `lib/policy.agency`, but the symbols they use didn't come with them: `BUILTIN_POLICIES`, `minimalAutoApprovePolicy`, `recommendedAutoApprovePolicy` (from `std::policy`), `POLICY_PATH` (from `./config.agency`), and `ChoiceItem` (from `std::ui`) are all referenced but not imported. It compiled only because directory-seeded typechecking still sees those symbols via `agent.agency` in the same compile unit — but at runtime each `.agency` is its own module, so in `policy.js` these are undefined. `printPolicyHelp()` would then throw on `for (p in BUILTIN_POLICIES)`, and the first-run picker `givePolicyChoice()` would `match` over nothing. Add the imports to the module. (Inline comment on the import line.)

### 🟡 Medium — `codeAgent` triages with an extra LLM call on every turn

`codeAgent` now calls `triage(userMsg)` (an `llm()` round-trip) at the top of *every* invocation. Two concerns:
- **Hot-path cost/latency:** the coordinator already routes code work to `codeAgent`, so every code request now pays an extra classification call before doing anything, including the common simple case.
- **Silent agent switch:** a `"complex"` classification `return`s into `plannerAgent`, which on generation failure falls back to `codingAgent` (`std::agents/coding`) — a *different* agent than this `codeAgent`, with a different prompt and toolset. So a task can silently leave the code subagent entirely. Worth gating the triage (only when the message looks non-trivial) and/or making that handoff visible, rather than running it unconditionally.

### 🟡 Medium — planner approval prompt never shows the plan

In `planner.agency`, `runApproved` calls `requestApproval("", src, renderEffects(eff))` — the `plan` field of the `planApprove` interrupt is always `""`. The whole point of the gate is to show the plan before running it, but only source + effects reach the handler. `planText` is available up in `planStep`; thread it through `runApproved` → `requestApproval` so the user actually sees the plan they're approving.

### 🟢 Low / notes

- **`givePolicyChoice` match is non-exhaustive** (pre-existing, just moved): `chooseOption(..., allowFreeText: true)` can return free text or `""`, but the `match` only handles `"minimal"` / `"recommended"` — any other answer has no branch. Add a default while it's being touched.
- **`package.json`**: adds a `zone.js` devDependency with no obvious consumer in the diff, and the file has no trailing newline. Confirm `zone.js` is intentional.
- **`destructive { }` removed from `exec`/`bash`** in `shell.agency`: noting it so it's on the record — per our discussion this is intentional/left in for now, to revisit later. No action here.
- The build log also prints `AG6019` type errors in `lib/agents/review/agent.agency` and `lib/agents/policy/agent.agency` (`Result` passed to `systemMessage`). Those files aren't in this PR, so they look pre-existing / non-fatal (the build dies on the Vue error, not these) — flagging only so they're not mistaken for new regressions.

### 👍 Nice

- The layering discipline (`@module` notes stating the import direction) makes the decomposition easy to trust.
- `search.agency`'s `coerceBackend` at the write boundary is a clean way to keep `activeBackend` well-typed.
- The `ipc.ts` package-root materialization fix is the right call and benefits every `runCode` caller, not just the planner.

Issue #544, opened 2026-07-14
Scoped callbacks with block bodies do not survive cross-process resume


egonSchiele:
## Summary

A scoped callback registered with a block body (`callback("onNodeStart") as data { ... }`) does not survive a cross-process pause/resume. When the process that registered the callback is a subprocess that pauses on a bubbled interrupt, the resume dies at checkpoint restore:

```
FunctionRefReviver: function "__cb_main_0" from module "agency_<id>" not found in registry.
The function may have been renamed or removed since this state was serialized.
```

The outer `run()` then returns this failure and the interrupt never surfaces to the user.

## Mechanism

This is the sibling of #513, with one important difference.

- `State.toJSON` serializes `scopedCallbacks` function values as FunctionRefs (name + module).
- `liftCallbacks` (lib/preprocessors/liftCallbacks.ts) names callback blocks `__cb_<scope>_<n>`. Those names are not present in the function-ref registry when a FRESH process restores a checkpoint, so `FunctionRefReviver.findInRegistry` throws eagerly and the restore dies.
- The #513 fix (lazy stub for unregistered `__block_<n>` refs) deliberately does NOT cover `__cb_*` names, because its safety argument does not transfer: block *arguments* are rebound unconditionally by the replaying def body before anything can call them, but scoped-callback *registrations* are skipped by the step counter on resume — nothing rebinds them. A stub here would not be a never-fires tripwire; it would fire on the next forwarded event and just move the crash from restore time to fire time.

A real fix needs the callback registration to be re-established on replay (or the lifted `__cb_*` functions to be registered at module init so revival finds them). That is a deliberate design decision about callback lifecycle, out of scope for the #513 PR.

## Repro

Fixture (written for `tests/agency/subprocess/`, removed from the #513 branch so CI stays green — paste back to reproduce). Expected healthy output `"OK: deep-ok\n:cb=2"`; actual: `run failed` with the FunctionRefReviver error above.

```agency
import { compile, run } from "std::agency"

// The middle process registers a block-bodied scoped callback, then pauses
// across serialize/resume (grandchild bash is unhandled). After the resume it
// runs a SECOND child; if the callback survived the round trip, that child's
// onNodeStart increments the counter again. Healthy expectation: cb=2.
node main() {
  const grandSource = """
import { bash } from "std::shell"
node main() {
  let r = bash("echo deep-ok")
  return r.stdout
}
"""
  const secondSource = """
node main() {
  return "second done"
}
"""
[...]

PR #545, opened 2026-07-14
Revive block refs across subprocess resume via lazy stub (#513)

### LINKED ISSUE #513: guard block around a paused subprocess run loses its value after serialize/resume (nested pause)

egonSchiele:
Fixes #513.

## What was actually wrong

Blocks (`as { }`) self-register in the function-ref registry when their
creating line executes. A frame that holds a block argument serializes it as
a `FunctionRef` name reference. When a paused subprocess is re-forked, the
fresh process restores the checkpoint BEFORE running any code, so the block
is not registered yet — and `FunctionRefReviver` threw eagerly, killing the
whole resume. The revived value was garbage anyway: replay rebinds block
arguments at function entry before anything can call them.

The earlier investigation doc blamed a popped block frame; that theory is
wrong. The doc lands here with a correction banner.

## The fix

- `FunctionRefReviver`: an unregistered block ref now revives to a lazy stub
  that throws a precise error IF invoked (a tripwire that never fires in
  correct replays). Named functions keep the eager throw — for them a miss
  means the program really changed. The `__block_<n>` naming scheme now has
  a single source of truth (`lib/runtime/blockNames.ts`) shared by the
  backend that mints names (stepPathTracker) and the reviver that recognizes
  them.
- Three regression fixtures: nested pause through `run(maxCost:)`, through a
  user-level `guard(...) as { }`, and through a user-DEFINED block-taking
  wrapper (pins the goal that userland guard-like constructs are pause-safe
  with no runtime changes).
- `run()` collapsed to a single always-guarded call site (the #512 review
  note this bug was blocking). Every `run()` now exercises the guarded path,
  so the whole subprocess suite becomes a regression alarm for #513.
- Probed the sibling case (scoped callbacks with block bodies across a
  cross-process resume): it is broken independently, for `__cb_<scope>_<n>`
  names, and a stub alone would NOT be safe there (callback registrations
  are not rebound by replay). Filed #544 with the repro instead of shipping
  a red fixture.

## Testing

- Reviver unit tests: stub for unregistered block names, tripwire message on
  stub invocation, eager throw preserved for non-block names (`__blockish`,
  `__block_`), and registered-block-returns-real (pins lookup-before-regex
  ordering). 29/29 pass.
- All three new fixtures fail before the reviver change with the exact
  pre-fix signature (`Expected 1 interrupts but only 0 occurred`) and pass
  after it.
- Local regression sample, all passing: `nested-pause-resume`,
  `pause-multi-cycle`, `concurrent-handled`, `handler-reject`,
  `callback-forwarding-nested-relay`, `run-max-cost` (5 tests),
  `cost-no-double-charge-across-pause`.
- `pnpm run lint:structure` clean; CI runs the full agency suite.

### DIFF
--- a/packages/agency-lang/lib/runtime/blockNames.ts
+++ b/packages/agency-lang/lib/runtime/blockNames.ts
@@ -8,6 +8,14 @@
  *  function with such a name today, but misclassification requires a
  *  registry miss too, and the failure mode is a soft lazy-throwing stub. */
 export function makeBlockName(counter: number): string {
+  // isBlockName only recognizes \d+, so minting is restricted to the same
+  // shape — otherwise this file's whole reason to exist (the two sides
+  // cannot desync) would be false for non-integer input.
+  if (!Number.isInteger(counter) || counter < 0) {
+    throw new Error(
+      `makeBlockName: counter must be a non-negative integer (got ${counter})`,
+    );
+  }
   return `__block_${counter}`;
 }
 
--- a/packages/agency-lang/lib/runtime/blockNames.ts
+++ b/packages/agency-lang/lib/runtime/blockNames.ts
@@ -1,0 +1,16 @@
+/** Single source of truth for the compiler's anonymous-block naming scheme.
+ *  The backend MINTS names via makeBlockName (stepPathTracker.nextBlockName);
+ *  the runtime RECOGNIZES them via isBlockName (FunctionRefReviver's stub
+ *  decision for unregistered refs). Keeping both here means the two
+ *  subsystems cannot silently desync.
+ *
+ *  "__block_" is a de-facto reserved prefix: no diagnostic rejects a user
+ *  function with such a name today, but misclassification requires a
+ *  registry miss too, and the failure mode is a soft lazy-throwing stub. */
+export function makeBlockName(counter: number): string {
+  return `__block_${counter}`;
+}
+
+export function isBlockName(name: string): boolean {
+  return /^__block_\d+$/.test(name);
+}
--- a/packages/agency-lang/lib/runtime/revivers/functionRefReviver.ts
+++ b/packages/agency-lang/lib/runtime/revivers/functionRefReviver.ts
@@ -1,6 +1,7 @@
 import { BaseReviver } from "./baseReviver.js";
 import { AgencyFunction } from "../agencyFunction.js";
 import type { FuncParam, ToolDefinition } from "../agencyFunction.js";
+import { isBlockName } from "../blockNames.js";
 
 type FunctionRefRegistry = Record<string, AgencyFunction>;
 
@@ -107,9 +108,38 @@ export class FunctionRefReviver implements BaseReviver<AgencyFunction> {
       }
     }
 
+    // Compiler-generated blocks register themselves only when their creating
+    // line executes. A fresh process restoring a checkpoint has executed
+    // nothing yet, so a miss here is EXPECTED for blocks — replay rebinds
+    // block args at function entry before anything can call them (#513).
+    if (isBlockName(name)) {
+      return makeBlockStub(name, module);
+    }
     throw new Error(
       `FunctionRefReviver: function "${name}" from module "${module}" not found in registry. ` +
       `The function may have been renamed or removed since this state was serialized.`
     );
   }
+}
+
+/** A lazy tripwire for an unregistered block reference: a real
+ *  AgencyFunction (so restore succeeds and the frame slot is filled) whose
+ *  body throws a precise error IF anything invokes it. In every correct
+ *  replay the generated def body overwrites the slot with a fresh block
+ *  before any call, so this never fires. Plain `new` on purpose — the stub
+ *  must NOT self-register the way AgencyFunction.create does. */
+function makeBlockStub(name: string, module: string): AgencyFunction {
+  return new AgencyFunction({
+    name,
+    module,
+    fn: () => {
+      throw new Error(
+        `Block "${name}" from module "${module}" crossed a serialization ` +
+          `boundary and was invoked before replay rebound it. This is a ` +
+          `runtime bug — please report it.`,
+      );
+    },
+    params: [],
+    toolDefinition: null,
+  });
 }
--- a/packages/agency-lang/stdlib/agency.agency
+++ b/packages/agency-lang/stdlib/agency.agency
@@ -158,28 +158,13 @@ export def run(
     depth: _subprocessDepth() + 1
   })
 
-  // Two call sites on purpose. Wrapping the no-cap path in an inert
-  // guard block breaks nested subprocess pause/resume: when THIS process
-  // is itself a subprocess that pauses on a bubbled interrupt, resuming
-  // its serialized stack through the guard block loses the block's
-  // return value and run() yields undefined (see the nested-pause-resume
-  // fixture, which caught this in CI). Keep the argument lists below
-  // IDENTICAL.
-  if (maxCost == null) {
-    return try _run(
-      compiled,
-      node,
-      args,
-      wallClock,
-      memory,
-      ipcPayload,
-      stdout,
-      configOverrides,
-      cwd,
-      maxDepth,
-    )
-  }
-  const guarded = guard(cost: maxCost) as {
+  // guard() disables a dimension on a NEGATIVE value (null for BOTH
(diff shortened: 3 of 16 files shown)


egonSchiele (commented):
Reviewed against the spec, plan, and the surrounding runtime. This is a clean landing and it folded in every point from the earlier design/plan/test reviews: `blockNames.ts` as the single source of truth, `makeBlockStub` extracted, `NO_COST_CAP` named + a declarative `if…then…else` derivation, the registered-block-returns-real test that pins lookup-before-regex ordering, `concurrent-handled` + `handler-reject` added to the local sample, and the scoped-callback sibling correctly diagnosed as *not* stub-safe and split out to #544 rather than papered over. Verified independently that `AgencyFunction.invoke` reaches the stub's throwing `fn` with no ALS context (`withCallDepth` falls back to `DEFAULT_MAX_CALL_DEPTH`), so the tripwire test is sound.

One thing worth a reviewer's attention before merge, plus a small nit — inline. Nothing blocking.

**Confirming the failure path of the collapse** (so nobody has to re-derive it): after the collapse, a genuine child failure on the no-cap path flows `_run` fails → `guarded` is that failure → the `err.type == "guardFailure"` branch is false → falls through to `return guarded`. Success and failure semantics are both preserved; the only behavioral change is the extra empty guard-block frame now wrapping every `run()`. That is the intended regression alarm — see the inline note on why it deserves a green *full* CI run, not just the local sample.

packages/agency-lang/lib/runtime/blockNames.ts
```diff
@@ -0,0 +1,16 @@
...
+ *  The backend MINTS names via makeBlockName (stepPathTracker.nextBlockName);
+ *  the runtime RECOGNIZES them via isBlockName (FunctionRefReviver's stub
+ *  decision for unregistered refs). Keeping both here means the two
+ *  subsystems cannot silently desync.
+ *
+ *  "__block_" is a de-facto reserved prefix: no diagnostic rejects a user
+ *  function with such a name today, but misclassification requires a
+ *  registry miss too, and the failure mode is a soft lazy-throwing stub. */
+export function makeBlockName(counter: number): string {
+  return `__block_${counter}`;
```
egonSchiele, line 19: Small coupling nit, on-theme for a file whose stated purpose is that the two subsystems "cannot silently desync": `makeBlockName` takes `number`, but `isBlockName` only recognizes `\d+` (non-negative integers). They agree today only because `blockCounter++` is always a non-negative integer. `makeBlockName(1.5)` → `"__block_1.5"` would mint a name its own sibling predicate rejects. Worth either narrowing the param intent (a comment: "counter is a non-negative integer") or a cheap `Number.isInteger` guard, so the invariant is stated rather than assumed. Non-blocking.

packages/agency-lang/stdlib/agency.agency
```diff
@@ -158,28 +158,13 @@ export def run(
...
-      maxDepth,
-    )
-  }
-  const guarded = guard(cost: maxCost) as {
+  // guard() disables a dimension on a NEGATIVE value (null for BOTH
+  // dimensions throws), so a null maxCost maps to the disable value and
+  // one guarded call site serves both cases. Safe since #513: block
+  // frames revive correctly across a nested subprocess pause.
+  const NO_COST_CAP = -1.0
+  const cap = if maxCost != null then maxCost else NO_COST_CAP
```
egonSchiele, line 166: This is the highest-blast-radius change in the PR: every `run()` — including the overwhelmingly common no-cap case — now executes inside a `guard(cost: -1) as {}` block frame (no active guard, but a real block frame in the checkpoint/step structure). That is deliberate and is what makes the whole subprocess suite a #513 regression alarm. Because it shifts the step-counter/checkpoint shape of *every* subprocess run, it is the change most likely to have a subtle interaction somewhere outside the local sample. The local sample looks well-chosen (concurrent + reject + cost + nested), but I'd gate merge on a green **full** agency suite in CI, not just the sampled fixtures. The success/failure semantics themselves check out (`return guarded` at the tail handles non-guardFailure errors).

packages/agency-lang/lib/runtime/revivers/functionRefReviver.ts
```diff
@@ -107,9 +108,38 @@ export class FunctionRefReviver implements BaseReviver<AgencyFunction> {
       }
     }
 
+    // Compiler-generated blocks register themselves only when their creating
+    // line executes. A fresh process restoring a checkpoint has executed
```
egonSchiele, line 112: Good call leaving the eager throw in place for non-block names — a registry miss on a real named function genuinely means the program changed between pause and resume, and the user needs that error immediately. And the `__cb_*` scoped-callback siblings correctly fall through to this throw (not the stub) since a stub there would mask the real breakage tracked in #544 and fire later at a confusing site.

packages/agency-lang/lib/runtime/blockNames.ts
egonSchiele: Fixed in 3a3e7d2f4: makeBlockName now refuses non-integer/negative counters, and blockNames.test.ts pins the mint-recognize round trip plus the rejection cases.

packages/agency-lang/stdlib/agency.agency
egonSchiele: Full suite is green: agency-tests passed on both 22.x and 23.x for this branch (run 29376728088), so the collapse has been exercised by every subprocess fixture in CI, not just the local sample. The nit-fix push will re-run it; will confirm it stays green.

PR #546, opened 2026-07-15
Add scheduled maintenance agents (docs freshness, constants, stdlib doc links)

egonSchiele:
## What

Adds three **scheduled maintenance agents** for this repo, wired to run on GitHub Actions via `agency schedule add --backend github`, plus a shared PR helper.

| Agent | Schedule | Behavior |
|-------|----------|----------|
| `stdlib-docs-links.agency` | daily (09:00 UTC) | Finds generated stdlib docs not linked from the VitePress sidebar in `config.mts` and opens a PR adding the links. |
| `docs-review.agency` | weekly (Mon 09:00 UTC) | Audits each `docs/dev/*.md` against the current code, opens a PR with a freshness report (does **not** edit docs). |
| `constants-review.agency` | weekly (Mon 09:00 UTC) | Finds magic literals that should move to `lib/constants.ts`, opens a PR with a report (does **not** move constants). |

Each agent opens a PR with a findings file / edit for a human to review — nothing auto-merges.

The daily agent has real work today: 7 generated docs are currently unlinked (`/stdlib/git`, `/stdlib/tag`, and the whole `/stdlib/agents/*` subtree).

## Files

- `packages/agency-lang/agents/scheduled/{pr,stdlib-docs-links,docs-review,constants-review}.agency` — agents + shared helper (all compile + typecheck clean)
- `packages/agency-lang/agents/scheduled/.gitignore` — ignores compiled `.js`
- `.github/workflows/{stdlib-docs-links,docs-review,constants-review}.yml` — generated workflows

## Security posture

The classic GitHub Actions attack classes are structurally absent or mitigated:

- **No untrusted-input injection** — triggers are only `schedule` + `workflow_dispatch`; nothing attacker-controlled flows into a `run:` step.
- **SHA-pinned actions** (`checkout`, `run-agency-action`); not `--no-pin`.
- **Least-privilege token** — `contents: write` + `pull-requests: write` only.
- **The LLM never gets write/exec/network tools.** The review agents hand the model only read-only `read` + `grep`; the daily agent gives it no tools. The *agent* — not the model — does all git work, so a prompt-injection buried in a doc can't escalate.
- **Fail-closed guard** on the daily agent verifies the rewritten `config.mts` dropped no existing link and added exactly the missing ones before pushing.
- **`openPr` stages only the file it changed** (never `git add -A`), so a stray runner artifact can't leak into a PR.

## Activation (repo owner)

1. Repo *Settings → Actions → General* → enable **"Allow GitHub Actions to create and approve pull requests."**
2. Add the **`OPENAI_API_KEY`** repo secret (`GITHUB_TOKEN` is automatic).
3. Test with a manual **`workflow_dispatch`** run before relying on cron.

> Note: the root `.gitignore` `.*` rule ignores all of `.github/`, so the workflow files were force-added (`git add -f`), matching how the existing workflows are tracked.


packages/agency-lang/agents/scheduled/constants-review.agency
```diff
@@ -0,0 +1,64 @@
...
+  `grep`); it cannot write files, run shell commands, or make network calls. The
+  agent itself performs every git action, and the output is a PR a human
+  reviews. See `pr.agency`.
+*/
+
+node main() {
+  const root = repoRoot()
+  const pkg = pkgDir()
+
+  const constsOut = exec("cat", ["${pkg}/lib/constants.ts"]) with approve
```
egonSchiele: Why not read the file using `read`?

packages/agency-lang/agents/scheduled/constants-review.agency
```diff
@@ -0,0 +1,64 @@
...
+
+  const prompt = "${rubric}\n\nCurrent lib/constants.ts (do NOT re-suggest anything already here):\n\n${consts}"
```
egonSchiele: Use a multi-line string, please.

packages/agency-lang/agents/scheduled/constants-review.agency
```diff
@@ -0,0 +1,64 @@
...
+  thread(session: "constants-review") {
+    finding = llm(prompt, tools: [read, grep]) with approve
+  }
+
+  if (finding.includes("NO ISSUES")) {
+    print("Reviewed the codebase; no un-centralized constants worth moving. Nothing to do.")
+    return
+  }
+
+  const report = "# Constants centralization review\n\nAutomated weekly scan for 'magic' literals that are duplicated across files or that callers must recognize, and that would be clearer in `lib/constants.ts`. Each item is a suggestion for a human to confirm; nothing has been moved.\n\n${finding}\n"
```
egonSchiele: Use a multi-line string, please.

packages/agency-lang/agents/scheduled/constants-review.agency
```diff
@@ -0,0 +1,64 @@
...
+
+  const reportDir = "${pkg}/docs/dev/_reports"
+  exec("mkdir", ["-p", reportDir]) with approve
```
egonSchiele: Exciting news, the agency standard library has functions for all sorts of file system changes so you don't have to run bash commands directly.
https://agency-lang.com/stdlib/fs.html#mkdir

packages/agency-lang/agents/scheduled/docs-review.agency
```diff
@@ -0,0 +1,76 @@
...
+  model) performs every git action. The runner has no on-disk secrets (its
+  tokens live in env vars), and the output is a PR a human reviews. See
+  `pr.agency`.
+*/
+
+node main() {
+  const root = repoRoot()
+  const pkg = pkgDir()
+
+  const listOut = exec("bash", ["-c", "cd '${pkg}' && ls docs/dev/*.md 2>/dev/null"]) with approve
```
egonSchiele: if there isn't anything in the agency stdlib to list files, lets add something

packages/agency-lang/agents/scheduled/docs-review.agency
```diff
@@ -0,0 +1,76 @@
...
+  const files = listOut.stdout.trim().split("\n")
+
+  const rubric = "You are auditing ONE developer documentation file for a programming language called Agency, checking whether it is out of date relative to the CURRENT source code in this repository. Use the read and grep tools to verify the doc's specific claims: file paths it cites, function/type/flag names, command invocations, and described behavior. Report ONLY concrete staleness: a cited path or symbol that no longer exists, a renamed flag/command, a documented behavior that the code contradicts, or a described feature that was removed. Ignore prose style and subjective wording. If the doc appears accurate and current, reply with exactly the text NO ISSUES and nothing else. Otherwise reply with a short markdown bullet list; each bullet names the stale claim and what the code actually shows now (cite file:line)."
+
+  let sections: string[] = []
+  let idx = 0
+  for (rel in files) {
+    if (rel.length > 0) {
+      const contentOut = exec("cat", ["${pkg}/${rel}"]) with approve
```
egonSchiele: use `read`

packages/agency-lang/agents/scheduled/docs-review.agency
```diff
@@ -0,0 +1,76 @@
...
+    idx += 1
+  }
+
+  if (sections.length == 0) {
+    print("Reviewed ${files.length} docs under docs/dev/; none appear out of date. Nothing to do.")
+    return
+  }
+
+  const joined = sections.join("\n\n")
+  const report = "# docs/dev freshness review\n\nAutomated weekly audit of `docs/dev/` against the current codebase. Each item below is a suspected out-of-date claim for a human to confirm and fix. Nothing here has been applied to the docs.\n\n${joined}\n"
```
egonSchiele: multiline string

egonSchiele (commented):
Automated review (Claude Code). Overall: solid design and the security posture described in the PR body checks out against the code — the model never gets write/exec/network tools, exec takes argv arrays with no LLM output interpolated into commands, workflows are SHA-pinned with least-privilege tokens.

Inline comments cover the main findings. The most important one is the unchecked exit codes in `openPr` — it converts every infrastructure failure into a silent green run, which defeats the purpose of unattended agents.

A few smaller notes that didn't fit a single line:

- **Cost awareness on `docs-review`:** `docs/dev/` currently has 59 markdown files, so the weekly run spawns 59 independent tool-using LLM threads, each seeded with the doc's full contents. Fine if intentional, but worth stating in the module doc so a future reader knows cost scales linearly with doc count.
- **Report content is untrusted:** LLM output flows into the committed report files, so a prompt injection buried in a doc could plant misleading markdown in `docs/dev/_reports/*.md` (fake findings instructing the human reviewer, or a future agent reading that directory, to make a harmful change). Human PR review is the mitigation — just be aware the report content is attacker-influenced text.
- **`contents: write` allows pushing to any branch, including `main` directly.** The agents only push fresh branches, but branch protection on `main` would make that structural rather than behavioral.
- **Tests:** the pure `stripFences` helper is a good candidate for an agency execution test (fenced input, unfenced input, fence-with-language tag, and the single-line edge gated by `lines.length >= 2`). Everything else mostly wraps `exec` and is reasonably covered by the manual `workflow_dispatch` smoke test the PR body prescribes.

packages/agency-lang/agents/scheduled/pr.agency
```diff
@@ -0,0 +1,80 @@
...
+  """
+  exec("git", ["-C", root, "config", "user.name", "agency-scheduler[bot]"]) with approve
+  exec("git", ["-C", root, "config", "user.email", "agency-scheduler@users.noreply.github.com"]) with approve
+  exec("git", ["-C", root, "checkout", "-b", branch]) with approve
+  const addArgs = ["-C", root, "add"]
+  for (p in paths) {
+    addArgs.push(p)
+  }
+  exec("git", addArgs) with approve
+  exec("git", ["-C", root, "commit", "-m", commitMsg]) with approve
```
egonSchiele: **Unchecked exit codes make failures silent.** `std::shell` `exec` resolves with `{stdout, stderr, exitCode}` on a non-zero exit — it does not throw (`lib/stdlib/abortable.ts`, the `close` handler resolves with `exitCode: code ?? 1`). So if this commit fails (e.g. nothing to commit), or the push is rejected, or `gh pr create` errors, `openPr` marches on through the remaining steps and returns an empty string; the agent prints `Opened PR: ` and the workflow run shows **green with no PR opened**. Since these run unattended on cron, nobody notices for weeks.

Each `exec` here (and in `repoRoot`) should check `exitCode` and bail loudly — the callers already do exactly this for `write(...)` (`w is failure(err)`), so the exec calls are inconsistent with the surrounding convention. This is also the silent-fallback anti-pattern from `docs/dev/anti-patterns.md`.

packages/agency-lang/agents/scheduled/pr.agency
```diff
@@ -0,0 +1,80 @@
...
+  @param root - Absolute path to the repo root
+  @param branch - Name of the branch to create and push
+  @param paths - Repo-root-relative paths to stage and commit
+  @param commitMsg - Commit message
+  @param title - Pull request title
+  @param body - Pull request body (markdown)
+  """
+  exec("git", ["-C", root, "config", "user.name", "agency-scheduler[bot]"]) with approve
+  exec("git", ["-C", root, "config", "user.email", "agency-scheduler@users.noreply.github.com"]) with approve
+  exec("git", ["-C", root, "checkout", "-b", branch]) with approve
```
egonSchiele: **Duplicate PRs pile up across runs.** Branch names embed a timestamp and nothing checks whether a previous run's PR is still open. Concretely: the daily agent finds 7 unlinked docs today and opens a PR; if it isn't merged by tomorrow, `config.mts` on main still lacks the links, so tomorrow's run opens a second identical PR — one per day until someone merges. Same for the weekly agents.

Cheap fix here in `openPr`: check `gh pr list --state open --search "<branch prefix> in:head"` and skip (or push to a fixed per-agent branch) when a PR already exists.

packages/agency-lang/agents/scheduled/docs-review.agency
```diff
@@ -0,0 +1,76 @@
...
+  for (rel in files) {
+    if (rel.length > 0) {
+      const contentOut = exec("cat", ["${pkg}/${rel}"]) with approve
+      const content = contentOut.stdout
+      const prompt = "${rubric}\n\nDoc path (relative to packages/agency-lang): ${rel}\nSearch code with grep/read using paths relative to the repository root (this doc lives at packages/agency-lang/${rel}).\n\nDoc contents:\n${content}"
+      let finding = ""
+      thread(session: "docrev-${idx}") {
+        finding = llm(prompt, tools: [read, grep]) with approve
+      }
+      if (!finding.includes("NO ISSUES")) {
```
egonSchiele: **Substring match silently discards real findings.** The rubric asks for *exactly* `NO ISSUES`, but `includes` is lenient: a reply like "NO ISSUES with the cited paths, but the `--force` flag was renamed…" gets classified as clean and the whole section is dropped. The failure direction is at least safe (missed findings, not bogus PRs), but `finding.trim() == "NO ISSUES"` matches the rubric's stated contract and loses nothing. Same pattern in `constants-review.agency`.

packages/agency-lang/agents/scheduled/constants-review.agency
```diff
@@ -0,0 +1,64 @@
...
+
+  const prompt = "${rubric}\n\nCurrent lib/constants.ts (do NOT re-suggest anything already here):\n\n${consts}"
+
+  let finding = ""
+  thread(session: "constants-review") {
+    finding = llm(prompt, tools: [read, grep]) with approve
+  }
+
+  if (finding.includes("NO ISSUES")) {
```
egonSchiele: Same substring-match issue as in `docs-review.agency`: the rubric promises exactly `NO ISSUES`, so `finding.trim() == "NO ISSUES"` is the check that matches the contract. With `includes`, a reply that *contains* the phrase alongside real findings is treated as clean.

packages/agency-lang/agents/scheduled/stdlib-docs-links.agency
```diff
@@ -0,0 +1,123 @@
...
+
+  // 4. Ask the model to rewrite config.mts with the missing links inserted into
+  //    the `/stdlib/` sidebar. No tools are provided; the model only edits text.
+  const guidance = "You are editing a VitePress sidebar config file (a TypeScript module). Add a sidebar link for each of the stdlib doc slugs listed below, inserting each into the `/stdlib/` sidebar section in a sensible place: alphabetical within its group, and nest slugs that share a path prefix such as `/stdlib/agents/*` under a collapsible group named after that prefix, mirroring how `/stdlib/data/*` and `/stdlib/ui/*` are already grouped. Use the last path segment as the link text. Do NOT remove or reorder any existing links, and do NOT change anything outside the `/stdlib/` sidebar section. Return the COMPLETE updated file contents and nothing else: no markdown fences, no commentary."
+  const prompt = "${guidance}\n\nSlugs to add:\n${missingList}\n\nCurrent file:\n${config}"
+
+  const raw: string = llm(prompt)
+  const newConfig = stripFences(raw)
+
+  // 5. Fail-closed guard. Extract every link that was in the OLD config and
```
egonSchiele: **Guard is solid on links, blind to everything else.** The model returns the *entire* rewritten `config.mts`, and this guard verifies no link was dropped and every missing slug was added — but nothing verifies the file still parses or that changes stayed inside the `/stdlib/` sidebar. A rewrite that mangles a `text:` label, duplicates a group, or breaks TS syntax passes the guard. Human review mitigates it, but a cheap deterministic backstop would help: assert the new config's line count grew by roughly `missing.length` (each link is one line), or syntax-check the result before committing.

packages/agency-lang/agents/scheduled/stdlib-docs-links.agency
```diff
@@ -0,0 +1,123 @@
...
+  //    require the new config to still contain it, then require every missing
+  //    slug to now be present. Any failure aborts BEFORE writing or pushing.
+  const linkCmd = "grep -oE 'link: \"[^\"]*\"' '${configPath}' | sed -e 's/^link: \"//' -e 's/\"$//'"
```
egonSchiele: **Fragile coupling to double-quote formatting.** Both the missing-slug check (step 3) and this grep assume `link: "..."` with double quotes. That's true today, but a future Prettier config flip to single quotes would make *every* slug look missing (the agent asks the model to re-add the whole sidebar) and make this dropped-link guard vacuously pass on an empty `oldLinks`. Worth a comment noting the assumption — or normalize both quote styles in the grep, and treat an empty `oldLinks` as a guard failure rather than success.

.github/workflows/stdlib-docs-links.yml
```diff
@@ -0,0 +1,24 @@
...
+  workflow_dispatch:
+permissions:
+  contents: write
+  pull-requests: write
+concurrency:
+  group: agency-${{ github.workflow }}
+  cancel-in-progress: false
+jobs:
+  run:
+    runs-on: ubuntu-24.04
```
egonSchiele, line 16: Consider adding `timeout-minutes` (e.g. 30) to this job — the GitHub default is 6 hours, so a hung LLM loop burns runner minutes and API spend that whole time. Applies to all three workflows. (The header says "Edit freely," so diverging from the generator output is sanctioned.)

packages/agency-lang/agents/scheduled/constants-review.agency
egonSchiele: Done — now reads `lib/constants.ts` with `read("lib/constants.ts", pkg)` and handles the `Result`.

packages/agency-lang/agents/scheduled/constants-review.agency
egonSchiele: Done — the rubric is now a `"""..."""` multi-line string.

packages/agency-lang/agents/scheduled/constants-review.agency
egonSchiele: Done — the prompt is now a `"""..."""` multi-line string.

packages/agency-lang/agents/scheduled/constants-review.agency
egonSchiele: Switched to `std::fs::mkdir` (via a shared `ensureDir` helper in `pr.agency`). No more `exec("mkdir")`.

packages/agency-lang/agents/scheduled/docs-review.agency
egonSchiele: `std::shell` already has `glob` and `ls`, so nothing new was needed — listing is now `glob("*.md", dir: devDir)`.

packages/agency-lang/agents/scheduled/docs-review.agency
egonSchiele: Done — each doc is read with `read(rel, devDir)` and the `Result` is handled.

packages/agency-lang/agents/scheduled/docs-review.agency
egonSchiele: Done — the rubric and per-doc prompt are now `"""..."""` multi-line strings.

packages/agency-lang/agents/scheduled/pr.agency
egonSchiele: Fixed. Every git/gh call now goes through a `run()` helper that checks `exitCode` and, on non-zero, prints the command + stderr and `exit(1)`s so the workflow goes red. The one expected non-zero exit — `git diff --cached --quiet` — is inspected directly (not via `run`) and an empty index is treated as a clean "nothing to commit" skip, so it can't cause a false failure.

packages/agency-lang/agents/scheduled/pr.agency
egonSchiele: Added `openPrExists(root, branchPrefix)`; each agent calls it before the LLM step and skips the run if a prior PR for the same purpose is still open (also saves the API spend). I went with check-and-skip rather than a stable branch + force-update because CLAUDE.md forbids force-pushing.

packages/agency-lang/agents/scheduled/docs-review.agency
egonSchiele: Fixed — now `finding.trim() == "NO ISSUES"`.

packages/agency-lang/agents/scheduled/constants-review.agency
egonSchiele: Fixed — now `finding.trim() == "NO ISSUES"`.

packages/agency-lang/agents/scheduled/stdlib-docs-links.agency
egonSchiele: Guard now compares the actual link SETS of the old and new config (`extractLinks`): it fails loudly if any existing link was dropped, any required slug is missing, or any unexpected link was introduced. It also asserts the file still contains `defineConfig`/`sidebar`. `extractLinks` is covered by the new `tests/agency/scheduled-helpers` test. A full TS parse is still out of reach in-agent, so human review remains the backstop for pure-syntax mangling.

packages/agency-lang/agents/scheduled/stdlib-docs-links.agency
egonSchiele: Documented the double-quote assumption in `extractLinks`, and an empty `oldLinks` is now a loud guard failure (`exit(1)`) instead of a vacuous pass, so a quote-format flip surfaces as a red run rather than a no-op.

.github/workflows/stdlib-docs-links.yml
egonSchiele: Added `timeout-minutes: 30` to all three workflows.

egonSchiele:
Pushed `ed10086f9` addressing the review. Inline replies on each thread; summary of the cross-cutting items:

- **Loud failures:** all git/gh calls go through a `run()` helper that checks `exitCode` and `exit(1)`s (no more silent-green runs). Empty-index is a clean skip, not a failure.
- **No duplicate PRs:** `openPrExists()` short-circuits a run when a prior PR is still open (before the LLM call).
- **Typed stdlib everywhere:** `glob`/`read`/`std::fs::mkdir` replace `bash`/`cat`/`mkdir`.
- **Exact `NO ISSUES`**, precise link-set guard, multi-line strings, and `timeout-minutes: 30` on all three workflows.
- **Tests:** added `tests/agency/scheduled-helpers` for the pure helpers (`stripFences`, `extractLinks`) — 6 cases, no LLM.
- **Cost + untrusted-report notes** added to the module docs.

Two items from the automated review are repo settings, not code, so they are not in this diff — flagging for when the workflows are enabled:
- Enable *Settings → Actions → General → "Allow GitHub Actions to create and approve pull requests."*
- `contents: write` lets a workflow push to any branch including `main`. The agents only ever push fresh branches, but a branch-protection rule on `main` would make that structural rather than behavioral.

PR #547, opened 2026-07-15
Guard time semantics: time <= 0 disables; input-wait is free

egonSchiele:
Part 1 of 3 of the CLI cost/time guards plan (`docs/superpowers/plans/2026-07-13-cli-cost-time-guards.md`, committed here). Semantics first; per-branch time budgets and the `--max-cost`/`--max-time` flags follow in separate PRs.

## Changes

**`guard(time: 0)` now disables the time guard** instead of tripping instantly. The design rule is: time `<= 0` disables (a zero time budget has no useful meaning), while cost keeps `0` as a real limit — "no paid spend", i.e. local-models-only — because the cost check trips on `spent > limit`. One-line condition change in `pushGuardImpl` plus the docstring.

**Waiting on a human via `input()` no longer counts against time budgets.** `inputImpl` pauses every guard on the active branch stack before blocking and resumes after, on both the override path and the readline path. These are the same idempotent `pause()`/`resume()` calls the runner makes at halt/step boundaries; `CostGuard.pause()` is a no-op, so only time budgets are affected. `pause()` cancels only the guard's own timer — the composed abort signal stays intact, so Ctrl-C / race-loser cancellation still releases a blocked input. `sleep` deliberately still counts: it is the program spending time, not a human.

Adds a test seam: `_installSlowInput(delayMs, answer)` (non-exported def in `std::thread`, reachable via `import test`) installs an `input()` override that answers after a delay, so fixtures can simulate a slow human without touching stdin.

## Tests

- `guard-time-zero-disables` — written first, failed with `tripped:timeoutFailure`, passes after the condition change.
- `guard-time-input-wait-free` — 40ms budget around a 120ms simulated human; written first, failed with `tripped:timeoutFailure`, passes after the pause/resume change.
- `guard-time-sleep-still-counts` — a sleep overrunning the budget still trips.
- `guard-input-abort-still-cancels` — the wait is not spuriously aborted while guards are paused.
- Existing time-guard fixtures all pass locally: `guard-time-trip`, `guard-time-no-trip`, `guard-time-and-cost`, `guard-time-nested`, `guard-time-nested-outer-tighter`, `guard-concurrent-branches`, `guard-cost-no-trip`.
- `lib/stdlib/thread.test.ts` + `lib/runtime/guard.test.ts`: 32/32; structural linter clean.

## Docs

`docs/site/guide/guards.md` "When the clock ticks" updated: `input` is now exempt. Generated `docs/site/stdlib/thread.md` refreshed.

### DIFF
--- a/packages/agency-lang/docs/site/guide/guards.md
+++ b/packages/agency-lang/docs/site/guide/guards.md
@@ -67,7 +67,7 @@ if (isFailure(result)) {
 When the clock ticks:
 - regular code execution = yes.
 - interrupts = no.
-- waiting for user input through `input` = yes.
+- waiting for user input through `input` = no (waiting on a human is free).
 - `sleep` = yes.
 
 When the time guard fires, any in-flight HTTP requests are cancelled and an abort signal is sent to other code as well.
--- a/packages/agency-lang/docs/site/guide/guards.md
+++ b/packages/agency-lang/docs/site/guide/guards.md
@@ -70,6 +70,12 @@ When the clock ticks:
 - waiting for user input through `input` = no (waiting on a human is free).
 - `sleep` = yes.
 
+One current limitation: the `input` exemption applies to time guards on the
+same execution branch. If `input()` runs inside a `fork` or `race` branch,
+an *outer* time guard's clock keeps running during the wait, because fork
+branches do not yet carry their own copy of the parent's timer. Per-branch
+time budgets (the next change in this series) remove this limitation.
+
 When the time guard fires, any in-flight HTTP requests are cancelled and an abort signal is sent to other code as well.
 
 ## Nested guards
--- a/packages/agency-lang/lib/stdlib/builtins.ts
+++ b/packages/agency-lang/lib/stdlib/builtins.ts
@@ -56,26 +56,46 @@ export function _parseJSON(text: string): any {
  * `AgencyCancelledError`, which `__tryCall` re-throws so cancellation
  * actually propagates.
  */
+/** Test-only: install an input override that resolves after `delayMs`,
+ *  used by guard fixtures to simulate a slow human without touching stdin.
+ *  Exposed to fixtures as the non-exported `_installSlowInput` def in
+ *  `stdlib/thread.agency` (test imports only). */
+export function _installSlowInputImpl(delayMs: number, answer: string): void {
+  (globalThis as any).__agencyInputOverride = (_prompt: string) =>
+    new Promise<string>((resolve) => setTimeout(() => resolve(answer), delayMs));
+}
+
 function inputImpl(
   ctx: RuntimeContext<any>,
   stack: StateStack,
   prompt: string,
 ): Promise<string> {
+  // Waiting on a human must not count against a time budget. Pause every
+  // guard on the active branch stack before blocking, resume after. These
+  // are the same idempotent calls the runner makes on halt()/step entry;
+  // CostGuard.pause() is a no-op, so only time budgets are affected.
+  // pause() cancels only the guard's own timer — the composed abort signal
+  // stays intact, so external cancellation (Ctrl-C, race-loser) still
+  // releases the wait.
+  stack.guards.forEach((g) => g.pause());
+  const resumeGuards = () => stack.guards.forEach((g) => g.resume(stack));
+
   const override = (globalThis as any).__agencyInputOverride as
     | ((prompt: string) => Promise<string>)
     | undefined;
   if (override) {
-    return override(prompt);
+    return override(prompt).finally(resumeGuards);
   }
   const signal = ctx.getAbortSignal(stack);
   if (signal.aborted) {
+    resumeGuards();
     return Promise.reject(new AgencyCancelledError("input cancelled"));
   }
   const rl = readline.createInterface({
     input: process.stdin,
     output: process.stdout,
   });
-  return new Promise((resolve, reject) => {
+  return new Promise<string>((resolve, reject) => {
     const onAbort = () => {
       try { rl.close(); } catch {}
       reject(new AgencyCancelledError("input cancelled"));
@@ -86,7 +106,7 @@ function inputImpl(
       rl.close();
       resolve(answer);
     });
-  });
+  }).finally(resumeGuards);
 }
 
 /** Deprecated context-injected wrapper kept in place during the ALS
--- a/packages/agency-lang/lib/stdlib/builtins.ts
+++ b/packages/agency-lang/lib/stdlib/builtins.ts
@@ -56,15 +56,6 @@ export function _parseJSON(text: string): any {
  * `AgencyCancelledError`, which `__tryCall` re-throws so cancellation
  * actually propagates.
  */
-/** Test-only: install an input override that resolves after `delayMs`,
- *  used by guard fixtures to simulate a slow human without touching stdin.
- *  Exposed to fixtures as the non-exported `_installSlowInput` def in
- *  `stdlib/thread.agency` (test imports only). */
-export function _installSlowInputImpl(delayMs: number, answer: string): void {
-  (globalThis as any).__agencyInputOverride = (_prompt: string) =>
-    new Promise<string>((resolve) => setTimeout(() => resolve(answer), delayMs));
-}
-
 function inputImpl(
   ctx: RuntimeContext<any>,
   stack: StateStack,
@@ -84,7 +75,12 @@ function inputImpl(
     | ((prompt: string) => Promise<string>)
     | undefined;
   if (override) {
-    return override(prompt).finally(resumeGuards);
+    // Promise.resolve().then(...) so a synchronously-throwing override
+    // still reaches the finally — otherwise the guards stay paused for
+    // the rest of the run.
+    return Promise.resolve()
+      .then(() => override(prompt))
+      .finally(resumeGuards);
   }
   const signal = ctx.getAbortSignal(stack);
   if (signal.aborted) {
@@ -120,6 +116,15 @@ export function __internal_input(
   prompt: string,
 ): Promise<string> {
   return inputImpl(ctx, stack, prompt);
+}
+
+/** Test-only: install an input override that resolves after `delayMs`,
+ *  used by guard fixtures to simulate a slow human without touching stdin.
+ *  Exposed to fixtures as the non-exported `_installSlowInput` def in
+ *  `stdlib/thread.agency` (test imports only). */
+export function _installSlowInputImpl(delayMs: number, answer: string): void {
+  (globalThis as any).__agencyInputOverride = (_prompt: string) =>
+    new Promise<string>((resolve) => setTimeout(() => resolve(answer), delayMs));
 }
 
 /** ALS-reading replacement. Same body as `__internal_input`. */
--- a/packages/agency-lang/lib/stdlib/builtins.ts
+++ b/packages/agency-lang/lib/stdlib/builtins.ts
@@ -71,9 +71,14 @@ function inputImpl(
   stack.guards.forEach((g) => g.pause());
   const resumeGuards = () => stack.guards.forEach((g) => g.resume(stack));
 
-  const override = (globalThis as any).__agencyInputOverride as
-    | ((prompt: string) => Promise<string>)
-    | undefined;
+  // Per-execution override first (REPL readline routing, test seams —
+  // see RuntimeContext.inputOverride). The globalThis fallback exists
+  // only for the TUI debugger, which cannot reach the ctx yet.
+  const override =
+    ctx.inputOverride ??
+    ((globalThis as any).__agencyInputOverride as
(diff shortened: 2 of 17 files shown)


packages/agency-lang/lib/stdlib/builtins.ts
```diff
@@ -56,26 +56,46 @@ export function _parseJSON(text: string): any {
...
   prompt: string,
 ): Promise<string> {
+  // Waiting on a human must not count against a time budget. Pause every
+  // guard on the active branch stack before blocking, resume after. These
+  // are the same idempotent calls the runner makes on halt()/step entry;
+  // CostGuard.pause() is a no-op, so only time budgets are affected.
+  // pause() cancels only the guard's own timer — the composed abort signal
+  // stays intact, so external cancellation (Ctrl-C, race-loser) still
+  // releases the wait.
+  stack.guards.forEach((g) => g.pause());
```
egonSchiele, line 71: **This pauses only the guards on the current branch's stack, which misses an ancestor TimeGuard when `input()` runs inside a fork/race branch.** `TimeGuard.cloneForBranch` returns `undefined` (unlike `CostGuard`, which shares `this` into the child), and the runBatch setup comment confirms it: "For TimeGuard, install runs on the parent only — children don't need separate installs because the parent's abort signal propagates." So a child branch's `stack.guards` contains the shared CostGuard but **not** the parent's TimeGuard.

`_input` reads `getRuntimeContext().stack` = the branch stack, so for:
```
guard(time: 5s) as { race { input("approve?"), work() } }
```
the `input()` runs in a race branch whose `stack.guards` has no TimeGuard → this `pause()` is a no-op for the time budget → the parent timer keeps running → **the input wait counts, and worse, the parent timer can fire mid-wait and abort the human** — the exact opposite of the fixture's "not spuriously aborted" guarantee. The sequential case (`guard(time:) as { input() }`) is correct because the guard is on the same stack; only the descendant-branch case is affected.

The PR body states the exemption unconditionally. Options: (a) walk ancestor stacks / target all in-scope time guards, (b) accept the limitation but document it and add a fixture pinning the branch behavior so it's a conscious choice rather than a silent gap. At minimum I'd soften the claim.

packages/agency-lang/lib/stdlib/builtins.ts
```diff
@@ -56,26 +56,46 @@ export function _parseJSON(text: string): any {
  * `AgencyCancelledError`, which `__tryCall` re-throws so cancellation
  * actually propagates.
  */
+/** Test-only: install an input override that resolves after `delayMs`,
+ *  used by guard fixtures to simulate a slow human without touching stdin.
+ *  Exposed to fixtures as the non-exported `_installSlowInput` def in
+ *  `stdlib/thread.agency` (test imports only). */
+export function _installSlowInputImpl(delayMs: number, answer: string): void {
```
egonSchiele: Inserting `_installSlowInputImpl` here orphans the doc block just above it (`...AgencyCancelledError, which __tryCall re-throws so cancellation actually propagates.`) — that comment documents `inputImpl`'s cancellation behavior, but now it floats above the test installer, and `inputImpl` below is left with no doc comment. Move the installer below `inputImpl`, or move that comment back down onto `inputImpl`.

packages/agency-lang/lib/stdlib/thread.ts
```diff
@@ -278,7 +278,11 @@ function pushGuardImpl(
     stack.pushGuard(g);
     ids.push(g.guardId);
   }
-  if (timeLimit != null && timeLimit >= 0) {
+  // Time: a NON-POSITIVE limit disables (0 would otherwise trip instantly and
+  // has no useful meaning). Cost differs on purpose: cost 0 is a real limit
+  // meaning "no paid spend" (local-models-only), since check() trips on
+  // spent > limit (strict).
+  if (timeLimit != null && timeLimit > 0) {
```
egonSchiele, line 285: The `>= 0` → `> 0` change is correct and the inline comment explaining the cost/time asymmetry is exactly the right thing to leave here. Note one small downstream semantic: `guard(time: 0)` (cost null) now pushes zero guards but still passes the "requires at least one of: cost, time" check (because `timeLimit` is `0`, not `null`), so it becomes a silent no-op guard rather than an error. That mirrors the existing `guard(cost: -1)` behavior, so it's consistent — just worth being aware it's now reachable.

packages/agency-lang/tests/agency/guards/guard-time-input-wait-free.agency
```diff
@@ -0,0 +1,17 @@
+import { guard } from "std::thread"
+import test { _installSlowInput } from "std::thread"
+
+// A 40ms time budget wrapping an input() whose (simulated) human takes
+// 120ms to answer. Because input-wait does not count against a time
+// budget, the guard must NOT trip: the block returns the answer.
+node main() {
+  _installSlowInput(120, "hello")
+  const result = guard(time: 40ms) as {
```
egonSchiele: **Flake risk from a sub-50ms budget.** The TimeGuard timer arms at guard-install, before `input()` is reached and paused. If cold JIT / a loaded CI runner spends >40ms between install and the `pause()` call, the timer fires and this trips spuriously. Same concern with the 30ms budget in `guard-input-abort-still-cancels`.

The assertion only needs `input_time > budget > compute_time`; the absolute magnitudes buy nothing. Scaling up (e.g. `budget: 500ms`, `_installSlowInput(2000, ...)`) preserves the exact semantics with a 500ms margin against startup jitter. The `sleep-still-counts` (20ms/150ms) and `zero-disables` fixtures are trip-direction-robust, so they're fine as-is; it's specifically the two input fixtures where the small budget races startup.

egonSchiele (commented):
Reviewed the runtime carefully. The core is sound and the reasoning in the PR body checks out where I could verify it independently:

- **`time <= 0` disables, cost keeps `0`** — correct and well-justified. `TimeGuard`'s timer arms for `timeLimit` ms so `time: 0` would fire ~immediately (the "trips instantly" claim is real), while `CostGuard.check` trips on `spent > limit` (strict), so `cost: 0` genuinely means "no paid spend." The asymmetry is principled, not accidental.
- **pause/resume elapsed accounting** — verified correct and idempotent: `pause()` charges `elapsedMs += now - windowStart` then cancels the timer and early-returns if already paused; `resume()` re-arms `startWindow()` for `timeLimit - elapsedMs` (remaining), early-returns if already running. So pre-input time is preserved and only the input wait is excluded.
- **abort signal stays intact** — confirmed: `pause()` touches only the timer + elapsed bookkeeping, never `stack.abortSignal` or the controller, so Ctrl-C / race-loser cancellation still releases a blocked `input()`.
- **resume fires exactly once per path** — override (`.finally`), aborted-early (explicit call + return), and readline (`.finally`) each resume once; idempotency covers any overlap.

One real correctness gap (inline on the pause line), a test-flake risk on the two sub-50ms fixtures, and a minor doc-comment orphaning. Nothing blocking for a Part-1-of-3, but the gap is worth either fixing or documenting-with-a-test rather than shipping under the unqualified claim "input-wait no longer counts against time budgets."

packages/agency-lang/lib/stdlib/builtins.ts
```diff
@@ -56,26 +56,46 @@ export function _parseJSON(text: string): any {
...
+  (globalThis as any).__agencyInputOverride = (_prompt: string) =>
```
egonSchiele: I'm not sure a global override is the right solution? Why not just change the way the input function is implemented? How does this play with agency's concurrency and state isolation features?
https://agency-lang.com/guide/concurrency.html

packages/agency-lang/lib/stdlib/builtins.ts
```diff
@@ -56,26 +56,46 @@ export function _parseJSON(text: string): any {
...
+  stack.guards.forEach((g) => g.pause());
+  const resumeGuards = () => stack.guards.forEach((g) => g.resume(stack));
+
   const override = (globalThis as any).__agencyInputOverride as
     | ((prompt: string) => Promise<string>)
     | undefined;
   if (override) {
-    return override(prompt);
+    return override(prompt).finally(resumeGuards);
   }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c89cbbab6: the override call now goes through Promise.resolve().then(), so a synchronous throw still reaches the finally and resumes the guards.

packages/agency-lang/lib/stdlib/builtins.ts
egonSchiele: Fixed in c89cbbab6: the installer now lives below inputImpl, and the cancellation doc block sits directly on inputImpl again.

packages/agency-lang/tests/agency/guards/guard-time-input-wait-free.agency
egonSchiele: Fixed in c89cbbab6: both input fixtures now use a 500ms budget with a 2s simulated answer, and the fixture comments state the only required relation (input_time > budget > compute_time). sleep-still-counts and zero-disables left as-is per your note.

packages/agency-lang/lib/stdlib/thread.ts
egonSchiele: Agreed on both points. guard(time: 0) alone passing the at-least-one check and installing nothing mirrors guard(cost: -1); leaving it consistent rather than special-casing an error.

packages/agency-lang/lib/stdlib/builtins.ts
egonSchiele: Verified against runBatch: correct on all counts — TimeGuard.cloneForBranch returns undefined, so a branch stack carries the shared CostGuard but no TimeGuard, and the pause() here is a no-op for an ancestor timer. The sequential case is the one this PR fixes; the branch case is exactly what PR 2 in the plan closes (per-branch TimeGuard clones inheriting the remaining budget put a pausable timer ON the branch stack — the plan explicitly notes PR 2 "is what makes PR 1 correct inside a fork").
Taken option (b) for now, in c89cbbab6: the guide documents the limitation ("the input exemption applies to time guards on the same execution branch; an outer guard keeps ticking inside a fork/race branch until per-branch time budgets land"), and PR 2 adds the branch fixtures when it flips the behavior. Happy to add a pin-the-gap fixture here too if you want one, but it would be rewritten by PR 2 within days.

packages/agency-lang/docs/site/guide/guards.md
```diff
@@ -67,7 +67,7 @@ if (isFailure(result)) {
 When the clock ticks:
 - regular code execution = yes.
 - interrupts = no.
-- waiting for user input through `input` = yes.
+- waiting for user input through `input` = no (waiting on a human is free).
 - `sleep` = yes.
```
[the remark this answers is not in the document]
egonSchiele: Documented in c89cbbab6 — the guide now carries the branch-scope qualifier, removed again when per-branch time budgets land in PR 2.

packages/agency-lang/lib/stdlib/builtins.ts
egonSchiele: Good instinct to question it — and to be precise, the global seam predates this PR: the REPL (cli.ts) and the TUI debugger (debugger/driver.ts) already installed __agencyInputOverride; this PR only added a third installer. On the specific questions: the harness input field cannot model a SLOW answer (it feeds stdin instantly), so a seam is needed for these timing fixtures; and within one run a global behaves like the ctx field, since one process has one stdin and one human.
The real problem was cross-execution leakage inside one process, which violates the per-run-state rule. Fixed in c95d781a0: the override now lives on RuntimeContext.inputOverride (per-execution, visible in the type). The reader, the REPL routing, and the test seam all use the ctx field. The debugger keeps the legacy global for now — it only sees the compiled module exports, so migrating it needs a codegen export + fixture regen; filed #548 with the plan, and inputImpl keeps a documented fallback until that lands.

packages/agency-lang/lib/stdlib/builtins.ts
```diff
@@ -56,26 +56,46 @@ export function _parseJSON(text: string): any {
...
  * actually propagates.
  */
+/** Test-only: install an input override that resolves after `delayMs`,
+ *  used by guard fixtures to simulate a slow human without touching stdin.
+ *  Exposed to fixtures as the non-exported `_installSlowInput` def in
+ *  `stdlib/thread.agency` (test imports only). */
+export function _installSlowInputImpl(delayMs: number, answer: string): void {
+  (globalThis as any).__agencyInputOverride = (_prompt: string) =>
+    new Promise<string>((resolve) => setTimeout(() => resolve(answer), delayMs));
+}
```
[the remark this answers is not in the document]
egonSchiele: Superseded by the ctx migration in c95d781a0: the test seam now sets ctx.inputOverride, which lives and dies with the run, so nothing leaks across executions. The REPL keeps its explicit save/restore, now on the ctx field.

PR #549, opened 2026-07-15
Per-branch time budgets: working-time semantics across fork/race

egonSchiele:
Part 2 of 3 of the CLI cost/time guards plan (`docs/superpowers/plans/2026-07-13-cli-cost-time-guards.md` — PR 2 section revised in this branch's first commit to record the investigation that reshaped it). Follows #547; the `--max-cost`/`--max-time` flags land in the final PR.

## What this changes

A time budget now counts **working time along one causal path**, including through `fork`/`race`:

- `TimeGuard.cloneForBranch` returns a per-branch clone carrying the parent's *remaining* budget (spend 3 minutes of 10, then fork → each branch gets 7; no path can exceed the original 10) and the parent's `guardId`, so a trip inside a branch is owned by the outer `guard { }` boundary's `try`, exactly like shared CostGuard trips.
- `runBatch` pauses the parent's time guards at entry — the parent is waiting on children, which is not working time on its own path — and advances them by the max branch working time (sum for sequential mode) at every value or throw exit, then restarts the clock. A charge that lands over budget arms a zero-delay timer, so the trip fires at the join. Interrupted exits bank each branch's time and leave the parent paused; the single charge happens at the eventual final join, so pause/resume cycles never double-charge.
- Branch clones are armed at launch, not lazily at the first Runner step, so TS-only branches (the subprocess adopter, parallel tool dispatch) accrue wall time on their clone too. `run()` under a time guard stays billed.
- The guard serialization slice rule gets one exception: inherited **time** clones are branch-owned state (accrued working time, remaining budget) and now serialize with the branch (`inheritedTimeGuards`), with rehydrate adopting them by `guardId` instead of re-cloning. Without this, any interrupt inside a fork branch would reset that branch's clock. The shared-reference slice invariant for cost guards is untouched.

Why the investigation reshaped the plan: the original sketch hoped per-branch clones alone would suffice. They don't — the parent's timer keeps running while awaiting the fork, and `composeBranchAbortSignal` cascades its abort into every branch, so a parent trip would still kill a waiting human. The plan's PR 2 section records the verified findings.

Cost semantics are deliberately untouched: money adds up under parallelism, time does not.

## Not covered (documented)

`input()` inside a subprocess cannot pause the parent-process clock across the IPC boundary — and does not need to: child stdin is a dead pipe (`input()` in a subprocess is not a working feature), and real subprocess human-waits go through interrupts, which already pause every clock in the chain. The guide says this.

## Testing

- Unit: clone inherits remaining budget / parent guardId / 1ms floor; snapshot-and-advance accessors; serialization round trip proving an accrued branch clone is adopted (2.5s survives) while cost refs still re-attach as the same shared object. 120 tests across `guard`, `runBatch`, `stateStack` files; full `lib/runtime` suite 1327/1327.
- New fixtures: `guard-time-fork-per-branch` (2s human answer inside an 800ms fork budget — no trip, sibling completes), `guard-time-fork-remaining-budget` (branch trips the outer guard on the remaining budget), `guard-time-branch-survives-interrupt` (no spurious trip across an in-fork interrupt cycle).
- Sweeps, all green: all 32 `tests/agency/guards/` fixtures; the subprocess alarm suite (`nested-pause-*`, `run-max-cost`, `cost-no-double-charge-across-pause`, `cost-two-children-share-budget`, `concurrent-handled`, `handler-reject`, `pause-fork-mixed`).
- Structural linter clean.

### DIFF
--- a/packages/agency-lang/docs/site/guide/guards.md
+++ b/packages/agency-lang/docs/site/guide/guards.md
@@ -70,11 +70,27 @@ When the clock ticks:
 - waiting for user input through `input` = no (waiting on a human is free).
 - `sleep` = yes.
 
-One current limitation: the `input` exemption applies to time guards on the
-same execution branch. If `input()` runs inside a `fork` or `race` branch,
-an *outer* time guard's clock keeps running during the wait, because fork
-branches do not yet carry their own copy of the parent's timer. Per-branch
-time budgets (the next change in this series) remove this limitation.
+### Time budgets and concurrency
+
+A time budget counts **working time along one causal path**. Inside a
+`fork` or `race`:
+
+- Each branch gets its own countdown, inheriting the parent's *remaining*
+  budget at fork time. If you spend 3 minutes of a 10-minute budget and then
+  fork, each branch gets 7 minutes — no path through the program can exceed
+  the original 10.
+- A branch waiting on a human via `input` pauses only its own clock. Working
+  siblings keep counting on theirs.
+- When the branches rejoin, the parent's clock advances by the longest
+  branch's working time, which is the real time the parallel region cost.
+
+Cost stays shared and cumulative across branches, because money adds up
+under parallelism and time does not. The two dimensions differ on purpose.
+
+One caveat for subprocesses: an `input()` inside a child process started
+with `std::agency.run` cannot pause the parent's clock across the process
+boundary. In practice this does not matter — a subprocess that needs a
+human raises an interrupt, and interrupts stop every clock in the chain.
 
 When the time guard fires, any in-flight HTTP requests are cancelled and an abort signal is sent to other code as well.
 
--- a/packages/agency-lang/docs/site/guide/guards.md
+++ b/packages/agency-lang/docs/site/guide/guards.md
@@ -70,27 +70,11 @@ When the clock ticks:
 - waiting for user input through `input` = no (waiting on a human is free).
 - `sleep` = yes.
 
-### Time budgets and concurrency
-
-A time budget counts **working time along one causal path**. Inside a
-`fork` or `race`:
-
-- Each branch gets its own countdown, inheriting the parent's *remaining*
-  budget at fork time. If you spend 3 minutes of a 10-minute budget and then
-  fork, each branch gets 7 minutes — no path through the program can exceed
-  the original 10.
-- A branch waiting on a human via `input` pauses only its own clock. Working
-  siblings keep counting on theirs.
-- When the branches rejoin, the parent's clock advances by the longest
-  branch's working time, which is the real time the parallel region cost.
-
-Cost stays shared and cumulative across branches, because money adds up
-under parallelism and time does not. The two dimensions differ on purpose.
-
-One caveat for subprocesses: an `input()` inside a child process started
-with `std::agency.run` cannot pause the parent's clock across the process
-boundary. In practice this does not matter — a subprocess that needs a
-human raises an interrupt, and interrupts stop every clock in the chain.
+One current limitation: the `input` exemption applies to time guards on the
+same execution branch. If `input()` runs inside a `fork` or `race` branch,
+an *outer* time guard's clock keeps running during the wait, because fork
+branches do not yet carry their own copy of the parent's timer. Per-branch
+time budgets (the next change in this series) remove this limitation.
 
 When the time guard fires, any in-flight HTTP requests are cancelled and an abort signal is sent to other code as well.
 
--- a/packages/agency-lang/lib/runtime/runBatch.ts
+++ b/packages/agency-lang/lib/runtime/runBatch.ts
@@ -66,6 +66,7 @@ import { hasInterrupts, type Interrupt } from "./interrupts.js";
 import type { RuntimeContext } from "./state/context.js";
 import { GlobalStore } from "./state/globalStore.js";
 import type { BranchState, State, StateStack } from "./state/stateStack.js";
+import { TimeGuard } from "./guard.js";
 import type { ThreadStore } from "./state/threadStore.js";
 
 export type BatchChild<T> = {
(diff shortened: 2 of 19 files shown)


packages/agency-lang/lib/runtime/state/stateStack.ts
```diff
@@ -763,6 +788,21 @@ export class StateStack {
...
       guards: this.guards.slice(this.inheritedGuardCount).map((g) => g.toJSON()),
       inheritedGuardCount: this.inheritedGuardCount,
+      // EXCEPTION to the slice rule above: inherited TIME guards are
+      // branch-owned clones (remaining budget + accrued working time),
+      // not shared references. Re-cloning them from the parent on
+      // resume would reset the branch's clock, so they serialize with
+      // the branch and rehydrate adopts them by guardId.
+      ...(this.guards
+        .slice(0, this.inheritedGuardCount)
+        .some((g) => g instanceof TimeGuard)
```
egonSchiele: This conditional object-spread is the exact pattern `docs/dev/anti-patterns.md` bans under "Ugly code" (`...(cond ? { x } : {})` — "Please never use this pattern"). The structural linter doesn't catch it, but it's a documented no. It also computes `this.guards.slice(0, this.inheritedGuardCount)` twice (once in `.some`, once in `.map`).

Cleaner — hoist the slice and assign conditionally:
```ts
const json = { /* guards, inheritedGuardCount */ };
const inheritedTime = this.guards
  .slice(0, this.inheritedGuardCount)
  .filter((g) => g instanceof TimeGuard);
if (inheritedTime.length > 0) {
  json.inheritedTimeGuards = inheritedTime.map((g) => g.toJSON());
}
return json;
```

packages/agency-lang/lib/runtime/runBatch.ts
```diff
@@ -694,6 +792,11 @@ async function runRaceFirstTime<T>(
   }
   hooks?.propagateLoserCost?.(losers, opts.parentStack);
   hooks?.propagateWinnerCost?.(winnerTask.branch, opts.parentStack);
+  chargeAndResumeParentTimeGuards(
```
egonSchiele, line 802: On the normal winner-completes path the losers are NOT pre-paused before this `max` charge (unlike the interrupt-stamp path just above, which does `tasks.forEach(pauseBranchTimeGuards)` first). So `chargeAndResumeParentTimeGuards` here reads `snapshotElapsed()` on loser clones that may still be running — a live in-flight window rather than a banked value. It's *bounded* correctly (a loser ran the same wall window as the winner, so `max` ≈ the winner's elapsed either way), so I don't think it mis-bills — but it's an asymmetry with the interrupt path and relies on the bounding argument rather than reading settled state. Worth a one-line comment here confirming the losers-still-running read is intentional, or pausing them first for symmetry.

packages/agency-lang/lib/runtime/runBatch.ts
```diff
@@ -612,6 +699,11 @@ async function runRaceFirstTime<T>(
         new AgencyCancelledError("race loser", makeAbortCause({ kind: "raceLoser" })),
       );
     }
+    chargeAndResumeParentTimeGuards(
```
egonSchiele, line 703: Coverage gap: the race paths (`runRaceFirstTime` + `runRaceResume`) gained five `chargeAndResumeParentTimeGuards` sites plus loser-banking, but all three new fixtures are `fork` (`guard-time-fork-*`, `guard-time-branch-survives-interrupt`). There's no integration fixture exercising a time guard across `race { }` — winner-charge, loser-drop, and the resume-winner path are only covered by unit tests. Given how much race-specific accounting changed here, a `guard-time-race-*` fixture (winner under budget with a slow loser; and a resumed winner) would be worth adding. Similarly, is the `"sum"` (sequential) branch of the mode selector exercised by any fixture, or only unit-tested? The fixtures all hit `"max"`.

packages/agency-lang/lib/runtime/state/stateStack.ts
```diff
@@ -792,6 +832,11 @@ export class StateStack {
     // re-entry that reactivates the branch.
     stateStack.guards = (json.guards ?? []).map(guardFromJSON);
     stateStack.inheritedGuardCount = json.inheritedGuardCount ?? 0;
+    // Park deserialized inherited time clones for rehydrate to adopt.
```
egonSchiele, line 856: Latent coupling worth a comment: `fromJSON` parks inherited time clones in `parkedInheritedTimeGuards`, and ONLY `rehydrateInheritedGuardsFrom` moves them into `guards`. Any resume path that deserializes a branch stack and uses it without calling rehydrate would silently drop the inherited time guard — the branch runs with no time enforcement and no error. The current call sites (startInvoke, runRaceResume) do rehydrate, so it's fine today, but the invariant "a parked clone MUST be consumed by rehydrate" is implicit. A short note on `parkedInheritedTimeGuards` stating that adoption is mandatory (and ideally a dev assert if it's still non-empty at first use) would harden it.

egonSchiele (commented):
Traced the accounting through all the fork/race exit paths and the serialization exception. This is careful work and the hard parts hold up. Recording what I verified so reviewers don't re-derive it:

- **Value/throw joins read finalized clocks, not in-flight windows.** `startInvoke` chains `.finally(pauseBranchTimeGuards)` onto the promise that `Promise.allSettled` awaits, so by the time the settle loop runs `chargeAndResumeParentTimeGuards`, every branch clone is already paused/banked. No double-count on those paths.
- **guardId matching is correct.** In `chargeAndResumeParentTimeGuards`, `bg === pg` skips the shared CostGuard, `instanceof TimeGuard` filters to time, and distinct-object-same-guardId matches exactly the clones of each parent guard — including nested inner/outer guards, which clone independently.
- **The resume-after-compose ordering is correct and load-bearing.** A clone's `resume()` composes its controller into `stack.abortSignal`, and `composeBranchAbortSignal` overwrites that field, so arming must come last — the PR gets this right.
- **1ms floor + remaining-budget clone** correctly prevents "parent works 3m of 10, forks" from exceeding 10 on any path.
- **Serialization exception is sound**: inherited time clones serialize with the branch (`inheritedTimeGuards`), adopt-by-guardId preserves the accrued clock, no double-serialization (the main `guards` field slices them out).

One anti-pattern to fix, a race-path subtlety to confirm, a coverage gap, and a latent coupling — inline. Nothing that undermines the design.

packages/agency-lang/docs/site/guide/guards.md
```diff
@@ -70,11 +70,27 @@ When the clock ticks:
...
-same execution branch. If `input()` runs inside a `fork` or `race` branch,
-an *outer* time guard's clock keeps running during the wait, because fork
-branches do not yet carry their own copy of the parent's timer. Per-branch
-time budgets (the next change in this series) remove this limitation.
+### Time budgets and concurrency
+
+A time budget counts **working time along one causal path**. Inside a
+`fork` or `race`:
+
+- Each branch gets its own countdown, inheriting the parent's *remaining*
```
egonSchiele: remove all edits made to this file

packages/agency-lang/lib/runtime/runBatch.ts
```diff
@@ -442,6 +503,13 @@ export async function runBatch<T>(
     seen.add(c.key);
   }
 
+  // Delegate time enforcement to the branch clones for the duration of
+  // the region. Every exit path below either charge-and-resumes (value
+  // joins, throws) or deliberately leaves the parent paused
+  // (interrupted exits — the parent halts next, and the final join
+  // charges once). See chargeAndResumeParentTimeGuards.
+  pauseParentTimeGuards(parentStack);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c47157de9: the pause now runs after both defensive throws (0a duplicate-key, 0b mode-flip), so a caught assert cannot leave the parent timers paused.

packages/agency-lang/lib/runtime/state/stateStack.ts
egonSchiele: Fixed in c47157de9 with exactly the hoist-and-assign shape you sketched; the double slice is gone too.

packages/agency-lang/lib/runtime/runBatch.ts
egonSchiele: Took the symmetry option in c47157de9: losers (and the winner) are banked via pauseBranchTimeGuards before the charge reads them, with a comment explaining why the read must see banked values rather than live windows. The interrupt-stamp and winner-value paths now match.

packages/agency-lang/lib/runtime/state/stateStack.ts
egonSchiele: Hardened in c47157de9: the invariant is documented on parkedInheritedTimeGuards (adoption is MANDATORY), and chargeGuards/enforceGuards now assert the parked list is empty, so a future resume path that skips rehydrate fails loudly at its first enforcement point instead of silently running unguarded. Partial coverage by nature (a branch that never reaches an enforcement point wont trigger it), but it converts the silent failure into a loud one on every realistic path.

packages/agency-lang/lib/runtime/runBatch.ts
egonSchiele: Added in c47157de9: guard-time-race-winner (fast winner, slow loser aborted mid-sleep, no trip, banked-loser charge exercised) and guard-time-race-resumed-winner (winner interrupts, parent stays paused across the wait, runRaceResume charge-and-resume exit returns the post-resume value). Both pass.
On sequential: mode "sequential" is only reachable through callHook batching (runBatch.ts:148 notes it exists to preserve strictly-sequential hook ordering) — no user-facing Agency construct produces it, so there is no honest execution fixture to write. The "sum" branch stays covered at the unit level; if hook callbacks ever gain guard interaction worth pinning, that would be the moment to add one.

packages/agency-lang/docs/site/guide/guards.md
egonSchiele: Reverted all guards.md edits in c47157de9 — the file now matches main byte-for-byte. One consequence to flag: main's version still contains the PR-1 paragraph saying per-branch time budgets are "the next change in this series" and that the fork limitation exists, which becomes stale the moment this merges. Happy to do a docs-only follow-up with whatever framing you prefer, or leave it to you.

PR #550, opened 2026-07-15
Add --max-cost / --max-time to agency run and agency agent (exit 3 on budget trip)

egonSchiele:
Part 3 of 3 of the CLI cost/time guards plan (`docs/superpowers/plans/2026-07-13-cli-cost-time-guards.md`). Follows #547 (time semantics) and #549 (per-branch time budgets), and closes out the stream.

## What this adds

`agency run` and `agency agent` accept `--max-cost <dollars>` and `--max-time <duration>`:

- Flags resolve to `AGENCY_MAX_COST` / `AGENCY_MAX_TIME` on the spawned child — cleared-then-set exactly like `--policy`, so the env vars are an internal carrier and a stale value from the parent shell can never constrain a run behind the user's back.
- The child installs a root `CostGuard`/`TimeGuard` in `runNode`, next to the run-policy handler, under the same `isIpcMode` root-process gate (IPC subprocesses are metered by the parent's guard through the branch clone from #549). The disable rules match the guard design: cost `< 0` off, cost `0` = no paid spend (local-models-only), time `<= 0` off.
- A tripped budget exits with code **3** (`EXIT_CODE_BUDGET_EXCEEDED`, distinct from 1 = crash and 2 = usage error) and prints a one-line overrun message. Detection in the compiled entry's catch is by guard-trip CAUSE, not error class: a root time trip usually surfaces as an aborted in-flight leaf op (`AgencyCancelledError` carrying the `guardTrip` cause) rather than a `GuardExceededError` — the end-to-end test caught exactly this.
- `--max-time` requires an explicit unit (`500ms`, `30s`, `5m`, `1h`); a bare number is a usage error so a value's meaning is never guessed. `--max-cost` is bare dollars.

## Testing

- Unit: duration parsing (units, leading-minus disable, unitless rejection), budget resolution, root-guard install rules (6 cases), trip formatting and both trip-delivery shapes (GuardExceededError and cancelled-leaf-with-cause), agent env carrier set/cleared via injected spawn.
- End-to-end spawn tests against the built CLI: `--max-time 100ms` on a 2s sleeper exits 3 with `Exceeded time limit`; a bare `--max-time 300` exits 2; an inherited `AGENCY_MAX_TIME` with no flag does not leak (the sleeper completes).
- Fixture regen: every compiled fixture's entry catch gains the `reportBudgetExceededAndExit(__error)` line plus its import — 93 files, nothing else changed.
- 435 unit tests across `lib/cli` + the new runtime files; structural linter clean; guard fixtures smoke-tested.

## Docs

`docs/site/cli/run.md` and `docs/site/cli/agent.md` document both flags, the units rule, the `cost: 0` local-only trick, and exit code 3. The guards guide (`docs/site/guide/guards.md`) is deliberately untouched per the #549 review; note it still carries the pre-#549 fork-limitation paragraph — happy to do a docs-only pass with whatever framing you prefer.

### DIFF
--- a/packages/agency-lang/lib/cli/budget.ts
+++ b/packages/agency-lang/lib/cli/budget.ts
@@ -7,17 +7,25 @@ const UNIT_MS: Record<string, number> = {
   w: 604_800_000,
 };
 
-/** Parse a duration string (`30s`, `5m`, `1h`, `500ms`, or a leading-minus
- *  disable value like `-1s`) to milliseconds. A bare unitless number throws:
- *  the CLI requires an explicit unit so a value's meaning is never guessed. */
+/** Parse a duration string (`500ms`, `30s`, `5m`, `1h`, `2d`, `1w`, or a
+ *  leading-minus disable value like `-1s`) to milliseconds. A bare unitless
+ *  number throws: the CLI requires an explicit unit so a value's meaning is
+ *  never guessed. The computed milliseconds must be finite — an absurdly
+ *  long digit string would otherwise overflow to Infinity, stringify into
+ *  the env, and silently install no guard (fail-open on a cost-control
+ *  feature). */
 export function parseDurationMs(s: string): number {
   const m = /^(-?\d+(?:\.\d+)?)(ms|s|m|h|d|w)$/.exec(s.trim());
   if (!m) {
     throw new Error(
-      `--max-time: expected a duration like 30s, 5m, 1h, or 500ms (got "${s}")`,
+      `--max-time: expected a duration like 500ms, 30s, 5m, 1h, 2d, or 1w (got "${s}")`,
     );
   }
-  return parseFloat(m[1]) * UNIT_MS[m[2]];
+  const ms = parseFloat(m[1]) * UNIT_MS[m[2]];
+  if (!Number.isFinite(ms)) {
+    throw new Error(`--max-time: duration is too large (got "${s}")`);
+  }
+  return ms;
 }
 
 /** Resolve --max-cost / --max-time flag strings into the env-var string
--- a/packages/agency-lang/lib/cli/budget.ts
+++ b/packages/agency-lang/lib/cli/budget.ts
@@ -1,0 +1,45 @@
+const UNIT_MS: Record<string, number> = {
+  ms: 1,
+  s: 1_000,
+  m: 60_000,
+  h: 3_600_000,
+  d: 86_400_000,
+  w: 604_800_000,
+};
+
+/** Parse a duration string (`30s`, `5m`, `1h`, `500ms`, or a leading-minus
+ *  disable value like `-1s`) to milliseconds. A bare unitless number throws:
+ *  the CLI requires an explicit unit so a value's meaning is never guessed. */
+export function parseDurationMs(s: string): number {
+  const m = /^(-?\d+(?:\.\d+)?)(ms|s|m|h|d|w)$/.exec(s.trim());
+  if (!m) {
+    throw new Error(
+      `--max-time: expected a duration like 30s, 5m, 1h, or 500ms (got "${s}")`,
+    );
+  }
+  return parseFloat(m[1]) * UNIT_MS[m[2]];
+}
+
+/** Resolve --max-cost / --max-time flag strings into the env-var string
+ *  values the child reads. Cost stays as dollars; time becomes milliseconds.
+ *  Negative/zero pass through — the runtime install applies the disable rule
+ *  (cost < 0 disables; time <= 0 disables). */
+export function resolveBudget(opts: {
+  maxCost?: string;
+  maxTime?: string;
+}): { maxCost?: string; maxTime?: string } {
+  const out: { maxCost?: string; maxTime?: string } = {};
+  if (opts.maxCost !== undefined) {
+    const n = Number(opts.maxCost);
+    if (!Number.isFinite(n)) {
+      throw new Error(
+        `--max-cost: expected a number of dollars (got "${opts.maxCost}")`,
+      );
+    }
+    out.maxCost = String(n);
+  }
+  if (opts.maxTime !== undefined) {
+    out.maxTime = String(parseDurationMs(opts.maxTime));
+  }
+  return out;
+}
--- a/packages/agency-lang/lib/runtime/budgetExit.ts
+++ b/packages/agency-lang/lib/runtime/budgetExit.ts
@@ -1,22 +1,32 @@
 import { EXIT_CODE_BUDGET_EXCEEDED } from "../constants.js";
-import { isGuardExceededError, GuardExceededError } from "./guard.js";
+import { readCause } from "./errors.js";
 
 /** User-facing one-line message for a tripped top-level budget. */
-export function formatBudgetExceeded(e: GuardExceededError): string {
-  if (e.type === "cost") {
-    return `Exceeded cost limit of $${e.limit} (used $${e.spent})`;
+export function formatBudgetExceeded(cause: {
+  dimension: "cost" | "time";
+  limit: number;
+  spent: number;
+}): string {
+  if (cause.dimension === "cost") {
+    return `Exceeded cost limit of $${cause.limit} (used $${cause.spent})`;
   }
-  return `Exceeded time limit of ${e.limit}ms (ran ${e.spent}ms)`;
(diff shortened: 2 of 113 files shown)


packages/agency-lang/lib/cli/budget.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+/** Parse a duration string (`30s`, `5m`, `1h`, `500ms`, or a leading-minus
+ *  disable value like `-1s`) to milliseconds. A bare unitless number throws:
+ *  the CLI requires an explicit unit so a value's meaning is never guessed. */
+export function parseDurationMs(s: string): number {
+  const m = /^(-?\d+(?:\.\d+)?)(ms|s|m|h|d|w)$/.exec(s.trim());
+  if (!m) {
+    throw new Error(
+      `--max-time: expected a duration like 30s, 5m, 1h, or 500ms (got "${s}")`,
```
egonSchiele: The regex and `UNIT_MS` accept `d` (days) and `w` (weeks), but this error message, the `--max-time` help text ("30s, 5m, 1h"), and the PR body all list only ms/s/m/h. So `--max-time 2d` silently works while the error message claims it wouldn't. Either add d/w to the documented units or drop them from the regex — right now they're an undocumented capability that the error message actively denies.

packages/agency-lang/lib/runtime/budgetExit.ts
```diff
@@ -0,0 +1,32 @@
...
+/** User-facing one-line message for a tripped top-level budget. */
+export function formatBudgetExceeded(cause: {
+  dimension: "cost" | "time";
+  limit: number;
+  spent: number;
+}): string {
+  if (cause.dimension === "cost") {
+    return `Exceeded cost limit of $${cause.limit} (used $${cause.spent})`;
+  }
+  return `Exceeded time limit of ${cause.limit}ms (ran ${Math.round(cause.spent)}ms)`;
```
egonSchiele, line 23: Minor UX: a user who ran `--max-time 30s` sees "Exceeded time limit of 30000ms" on trip. Only the ms value crosses the env boundary, so the original unit is lost here — reporting raw ms is unambiguous but a little unfriendly. Not worth carrying the original string through unless you want the nicety; flagging so it's a conscious choice.

packages/agency-lang/lib/runtime/rootBudget.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (rawCost !== undefined) {
+    const cost = Number(rawCost);
+    if (Number.isFinite(cost) && cost >= 0) {
+      stack.pushGuard(new CostGuard(cost));
+    }
+  }
+  const rawTime = process.env[AGENCY_MAX_TIME];
+  if (rawTime !== undefined) {
+    const ms = Number(rawTime);
+    if (Number.isFinite(ms) && ms > 0) {
```
egonSchiele: `installRootBudget` fails **open**: a non-finite env value (`Number.isFinite` guard, no throw) silently installs no guard. Fine in practice because the only intended setter is the CLI, which validates via `resolveBudget` before setting. But for a *cost-control* feature, a malformed budget silently meaning "unbounded" is the riskier failure direction if the env is ever set another way (a compiled binary run directly with a hand-set `AGENCY_MAX_COST`). Since the env is documented internal-only this is defensible — just calling out that the direction is fail-open, not fail-closed.

egonSchiele (commented):
Reviewed the wiring end-to-end. Clean close-out of the stream — the mechanism is sound and I verified the parts that could silently break:

- **Cause-based exit detection is correct for BOTH trip shapes.** `GuardExceededError extends AgencyAbort` and passes `makeAbortCause({kind:"guardTrip", dimension, limit, spent, guardId})` to `super`, so `readCause` extracts it — a **cost** trip (raw GuardExceededError from the runner's check loop) exits 3 just like a **time** trip (AgencyCancelledError carrying the same cause from an aborted leaf). This was the one thing that could have made cost budgets silently exit 1; it holds.
- **The env carrier is genuinely clear-then-set** in both `run` (commands.ts) and `runBundledAgent.ts` — `delete` then conditional set — so a stale parent-shell `AGENCY_MAX_*` can't constrain a run behind the user's back. The security claim checks out.
- **Codegen placement is right**: `reportBudgetExceededAndExit(__error)` is first in the entry catch and returns for non-budget errors, so ordinary crashes still fall through to the exit-1 path.
- **The cross-part interaction works**: a root trip's guardId is not owned by any inner `_runGuarded`, so a user `guard() as {}` around the tripping work won't absorb the root budget trip (Part 2's ownedGuardIds machinery) — it propagates to exit 3.

A few minor things inline; nothing blocking. The most valuable is a test for that last interaction.

packages/agency-lang/lib/cli/budget.ts
egonSchiele: Documented rather than dropped, in 53817345d: multi-day budgets are plausibly useful for long agent runs. The error message, --max-time help text, and both CLI reference pages now list d/w, with unit tests for 2d and 1w.

packages/agency-lang/lib/runtime/budgetExit.ts
egonSchiele: Keeping raw ms as the conscious choice (comment added in 53817345d saying why): only the millisecond value crosses the env boundary, so the original unit string is genuinely gone by the time the message renders, and ms is unambiguous. If the nicety is ever wanted, the right shape is carrying the raw flag string in a second env var - happy to do that in a follow-up.

packages/agency-lang/lib/runtime/rootBudget.ts
egonSchiele: Went further than the flag: rootBudget now FAILS CLOSED in 53817345d. A non-finite AGENCY_MAX_* value throws with a message telling the user to unset it or pass a valid value, so a malformed budget can never silently mean unbounded. Negative stays the documented disable range (it is a deliberate value, not a malformed one). Unit-tested for "abc" and "Infinity".

packages/agency-lang/lib/cli/budget.ts
```diff
@@ -0,0 +1,45 @@
...
+    );
+  }
+  return parseFloat(m[1]) * UNIT_MS[m[2]];
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 53817345d: parseDurationMs validates the computed milliseconds is finite and throws a usage error otherwise (test: a 320-digit duration). The cost path already rejected Infinity via its Number.isFinite check, and the runtime side now fails closed on non-finite env values as a second layer.

packages/agency-lang/lib/runtime/rootBudget.test.ts
```diff
@@ -0,0 +1,56 @@
+import { describe, test, expect, afterEach } from "vitest";
+import { installRootBudget } from "@/runtime/rootBudget.js";
+import { StateStack } from "@/runtime/state/stateStack.js";
+import { CostGuard, TimeGuard } from "@/runtime/guard.js";
+import { AGENCY_MAX_COST, AGENCY_MAX_TIME } from "@/constants.js";
+
+afterEach(() => {
+  delete process.env[AGENCY_MAX_COST];
+  delete process.env[AGENCY_MAX_TIME];
+});
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 53817345d: afterEach clears AGENCY_IPC, and there is now an explicit test that installRootBudget no-ops under AGENCY_IPC=1.

packages/agency-lang/lib/runtime/budgetExit.ts
```diff
@@ -0,0 +1,32 @@
...
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 53817345d: dollar values format through a bounded-precision helper (toFixed(6) with trailing-zero trim), so accumulated token pricing cannot surface float artifacts in the overrun message.

egonSchiele:
Added the interaction test from the top-level review in 53817345d: a user `guard(time: 60s) as {}` wrapping the tripping sleep does NOT absorb the root `--max-time 100ms` trip — the fixture prints GUARD_ABSORBED if the inner boundary wrongly converts it, and the test asserts exit 3 with the overrun message instead. Passes against the built CLI.

PR #551, opened 2026-07-15
Add saveDraft: salvage partial work when a guard trips

egonSchiele:
## What & why

`guard(cost:, time:)` blocks are binary today: when the budget trips, the block returns a `failure` and **all partial work is lost**. This adds `saveDraft(v)` — an *anytime-algorithm* floor. Code records a best-so-far value as it works; when an enclosing guard trips, the guard returns the last saved draft instead of a failure.

The motivating case: *"research this, but don't spend more than five minutes"* — when the clock runs out, return the best answer so far rather than nothing.

```
const report = guard(time: 5m) as {
  let r = initialReport()
  saveDraft(r)                 // floor: even an immediate trip returns this
  while (moreToDo()) {
    r = refine(r)
    saveDraft(r)               // last-wins; best-so-far keeps improving
  }
  return r
}
```

With no `saveDraft` calls, guard behavior is byte-for-byte unchanged (additive).

## Design

- **Storage:** branch-local `StateStack.other.drafts`, keyed by frame depth. `other` outlives frame pops (frames are popped by the unwind *before* the guard boundary reads them), serializes for interrupt/resume, and is per-branch so `fork`/`race` branches can't clobber each other. Never on `State` (pops too early) or the guard object (shared/cloned across branches).
- **Value semantics:** drafts are `deepClone`d at save, so a later mutation can't change the salvage and a live-trip salvage matches a post-resume one.
- **Trip read:** on a settled result, `salvageOwnTrip` (1) passes interrupts through untouched — a paused block keeps its drafts for resume, no sweep; (2) salvages the **outermost** (shallowest) draft under the guard **only on this guard's own trip** — identified by a `guardId` now carried on `guardFailureData`, so a *returned* inner-guard failure is never mis-salvaged; (3) sweeps the region.
- **Outermost-wins (not deepest)** is the type-closest choice: a nested callee's draft is its own type, not the guard block's. Deep/cross-type salvage is deferred to a future `finalize`.
- **Clearing rule:** a frame clears its draft on **normal** completion (def `finally` on `__functionCompleted`; block try-body when `haltResult` is not interrupts), and keeps it on an abort/interrupt unwind. Prevents a completed sibling from leaking a stale salvage.
- **Type check:** `saveDraft(v)` is checked against the enclosing scope's return type (def/node bodies *and* guard blocks, whose type is inferred). Name-keyed; aliasing escapes (documented).

## Tests

- Draft-store unit tests (14), checker tests (3), plus 13 execution fixtures: basic salvage, additivity (no-draft), last-wins, outermost-across-frames, sequential-guards, nested-guards, propagated-failure (ownership), time-trip, interrupt-resume, stale-sibling, stale-block, fork-isolation, guard-outside-fork.
- Full `tests/agency/guards/` sweep: **47/47 pass**. Typechecker suite: 1545 pass. `result.ts`: 41 pass.

## Notes / deferred

Spec + plan are in `docs/superpowers/`. Deferred to follow-ups (each with a home in the model): `finalize` blocks, deep/fork-array salvage, `sigint`/`sigkill`, `saveDraft` as an LLM tool, a `Both` result type, and generalizing beyond guard trips (Esc/cancel/race). Root budgets (`--max-cost`/`--max-time`) bypass `_runGuarded`, so v1 doesn't salvage a root-budget trip.

### DIFF
--- a/packages/agency-lang/lib/runtime/drafts.ts
+++ b/packages/agency-lang/lib/runtime/drafts.ts
@@ -1,0 +1,114 @@
+import type { StateStack } from "./state/stateStack.js";
+import type { ResultValue } from "./result.js";
+import { deepClone } from "./utils.js";
+import { isFailure, success } from "./result.js";
+import { hasInterrupts } from "./interrupts.js";
+
+/** A saved best-so-far value. Wrapped so a stored `null`/`undefined` value is
+ *  distinct from "no draft for this frame". */
+type DraftRecord = { value: any };
+
+// Branch-local, serialized store: frame depth -> its latest draft. Lives in
+// `StateStack.other` (NOT on `State`) because frames are popped by the unwind
+// before a guard boundary reads them — `other` outlives frame pops. Depth
+// arithmetic is centralized here so no caller ever touches a stack index. See
+// docs/superpowers/specs/2026-07-14-save-draft-guards-design.md.
+
+/** The draft map on this stack, or undefined if none written yet. `other` is
+ *  already typed `Record<string, any>`, so no cast is needed. */
+function peekDrafts(stack: StateStack): Record<number, DraftRecord> | undefined {
+  return stack.other.drafts as Record<number, DraftRecord> | undefined;
+}
+
+function ensureDrafts(stack: StateStack): Record<number, DraftRecord> {
+  if (!stack.other.drafts) stack.other.drafts = {};
+  return stack.other.drafts as Record<number, DraftRecord>;
+}
+
+/** Depth of the frame that CALLED the current TS helper — the Agency scope
+ *  whose draft this is. Mirrors `StateStack.callerFrame()` (one frame below the
+ *  helper's own top frame). -1 when there is no caller (module-init/global). */
+function callerDepth(stack: StateStack): number {
+  return stack.stack.length - 2;
+}
+
+/** Low-level depth-keyed write (used by writeCallerDraft and unit tests). The
+ *  value is DEEP-CLONED so a later mutation of the saved object can't change the
+ *  salvage, and so a live-trip salvage matches a post-resume salvage (which
+ *  reads the checkpoint's clone). Matches Agency's value semantics. */
+export function writeDraft(stack: StateStack, depth: number, value: unknown): void {
+  ensureDrafts(stack)[depth] = { value: deepClone(value) };
+}
+
+/** Record the caller frame's best-so-far draft (last call wins). */
+export function writeCallerDraft(stack: StateStack, value: unknown): void {
+  const depth = callerDepth(stack);
+  if (depth < 0) return; // no caller: harmless no-op
+  writeDraft(stack, depth, value);
+}
+
+/** The region marker for a guard entered now: drafts at depth >= this are
+ *  "under" the guard. A positional marker is inherent (frames pop before the
+ *  boundary reads), but its meaning lives in this named function. */
+export function draftRegionStart(stack: StateStack): number {
+  return stack.stack.length;
+}
+
+/** The outermost draft under a guard: the shallowest depth >= `region`, or
+ *  undefined. Outermost (not deepest) is the type-closest choice — see spec. */
+export function readOutermostDraft(
+  stack: StateStack,
+  region: number,
+): DraftRecord | undefined {
+  const drafts = peekDrafts(stack);
+  if (!drafts) return undefined;
+  const depths = Object.keys(drafts)
+    .map(Number)
+    .filter((d) => d >= region);
+  return depths.length === 0 ? undefined : drafts[Math.min(...depths)];
+}
+
+/** Delete every draft at depth >= `region`. */
+export function sweepDrafts(stack: StateStack, region: number): void {
+  const drafts = peekDrafts(stack);
+  if (!drafts) return;
+  for (const d of Object.keys(drafts).map(Number)) {
+    if (d >= region) delete drafts[d];
+  }
+}
+
+/** Turn a guarded block's settled result into the guard's final result:
+ *   1. a PAUSED block (interrupts) passes through untouched — no sweep, its
+ *      drafts must survive resume;
+ *   2. on THIS guard's OWN trip (failure whose `guardId` is in `ids`), salvage
+ *      the outermost draft in the region;
+ *   3. otherwise return the result unchanged;
+ *   then sweep the region on any settled exit. */
+export function salvageOwnTrip(
+  stack: StateStack,
+  region: number,
+  ids: string[],
+  result: ResultValue | unknown,
+): ResultValue | unknown {
+  if (hasInterrupts(result)) return result;
+  let out = result;
+  const guardId = (
+    isFailure(result) ? (result.error as { guardId?: string }) : undefined
+  )?.guardId;
+  if (guardId !== undefined && ids.includes(guardId)) {
+    const draft = readOutermostDraft(stack, region);
+    if (draft !== undefined) out = success(draft.value);
+  }
+  sweepDrafts(stack, region);
+  return out;
+}
+
+/** Clear the CURRENT top frame's draft. Called from generated code (def
+ *  `finally` on `__functionCompleted`; block try-body on `!runner.halted`), so a
+ *  frame unwinding on an abort/interrupt keeps its draft for the boundary. */
+export function __clearTopFrameDraft(stack: StateStack | undefined): void {
+  if (!stack) return;
+  const drafts = peekDrafts(stack);
+  if (!drafts) return;
+  delete drafts[stack.stack.length - 1];
+}
--- a/packages/agency-lang/lib/runtime/drafts.ts
+++ b/packages/agency-lang/lib/runtime/drafts.ts
@@ -47,11 +47,25 @@ export function writeCallerDraft(stack: StateStack, value: unknown): void {
   writeDraft(stack, depth, value);
 }
 
-/** The region marker for a guard entered now: drafts at depth >= this are
- *  "under" the guard. A positional marker is inherent (frames pop before the
- *  boundary reads), but its meaning lives in this named function. */
-export function draftRegionStart(stack: StateStack): number {
-  return stack.stack.length;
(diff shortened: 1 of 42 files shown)


packages/agency-lang/lib/stdlib/thread.ts
```diff
@@ -363,13 +375,18 @@ export async function _runGuarded(
   block: unknown,
 ): Promise<ResultValue> {
   const { ctx, stack } = getRuntimeContext();
+  // Mark the draft region BEFORE running the block: drafts saved by the block
+  // and everything it calls sit at stack depth >= this marker. Keyed by this
+  // guard's id-set so the marker is resume-stable (see draftRegionStart).
+  const regionKey = ids.join(",");
+  const region = draftRegionStart(stack, regionKey);
```
egonSchiele, line 382: **Correctness edge (fix before merge): fully-disabled guards share region key `""` and sweep each other's regions.**

Since #547 a guard dimension disables on negative/zero (`pushGuardImpl` pushes nothing), and since the #545 collapse EVERY no-cap `run()` goes through `guard(cost: -1)` — so `ids = []` and `regionKey = ""` is a common, not exotic, case. Two consequences:

1. **Overlapping disabled guards collide.** Nested all-disabled guards (`guard(cost: -1) as { … guard(time: 0) as { … } … }`) both memoize under `""`: the inner `draftRegionStart` returns the OUTER's (shallower) marker, so the inner's exit sweeps `>= outer region` — wiping drafts that belong to the outer scope and to any REAL guard sandwiched between them, which then loses its own salvage. `clearDraftRegion("")` on the inner exit also deletes the outer's memo entry.
2. **Pointless work on the hot path.** A disabled guard can never trip (nothing to own), so the memo write, the settled-path sweep, and the region bookkeeping are dead weight on every no-cap `run()`.

Three-line fix: when `ids.length === 0`, skip the draft machinery entirely — return the `__tryCall` result directly. A guard that owns no guards can never salvage, and skipping avoids both the `""` collision and the sweep. Worth a fixture: nested disabled guards around a real guard whose draft must survive the inner disabled guard's exit.

packages/agency-lang/lib/runtime/drafts.ts
```diff
@@ -0,0 +1,130 @@
...
+  let out = result;
+  const guardId = (
+    isFailure(result) ? (result.error as { guardId?: string }) : undefined
+  )?.guardId;
+  if (guardId !== undefined && ids.includes(guardId)) {
+    const draft = readOutermostDraft(stack, region);
+    if (draft !== undefined) out = success(draft.value);
+  }
+  sweepDrafts(stack, region);
+  clearDraftRegion(stack, key);
```
egonSchiele, line 118: **Minor leak: `draftRegions` entries survive thrown exits, permanently.**

`salvageOwnTrip` (and so `clearDraftRegion`) only runs on settled results. When `__tryCall` THROWS through `_runGuarded` — an unowned outer trip passing through an inner guard, Esc/cancel — the inner guard's memo entry stays in serialized `other` forever: keys are per-guard-instance (`ids` are unique per `guard()` call), so nothing ever reuses or removes them. The outer boundary's `sweepDrafts` clears the *drafts* at those depths but not the foreign region keys. A long-running loop with nested guards under an outer budget accumulates one dead entry per unwind, all serialized into every subsequent checkpoint.

Cheapest fix: wrap the `__tryCall` in `_runGuarded` with a catch that clears this guard's own key and rethrows (interrupts return as values, not throws, so resume-needed entries are untouched). Alternatively `sweepDrafts` could also drop `draftRegions` entries whose value >= region. Non-blocking, but cheap to close now while the invariants are fresh.

packages/agency-lang/lib/runtime/drafts.ts
```diff
@@ -0,0 +1,130 @@
...
+ *  helper's own top frame). -1 when there is no caller (module-init/global). */
+function callerDepth(stack: StateStack): number {
+  return stack.stack.length - 2;
+}
+
+/** Low-level depth-keyed write (used by writeCallerDraft and unit tests). The
+ *  value is DEEP-CLONED so a later mutation of the saved object can't change the
+ *  salvage, and so a live-trip salvage matches a post-resume salvage (which
+ *  reads the checkpoint's clone). Matches Agency's value semantics. */
+export function writeDraft(stack: StateStack, depth: number, value: unknown): void {
```
egonSchiele, line 39: **Docstring nit on clone semantics:** `deepClone` runs the native replacer/reviver, so a draft value containing a function (e.g. an `AgencyFunction` tucked in a record) is silently swapped for a registry reference at SAVE time — post-#545 an unregistered block becomes a lazy stub rather than a throw, so nothing fails loudly. That's consistent with checkpoint semantics, but worth one sentence in the `saveDraft` docstring ("drafts should be data, not functions") plus a note that each save pays a JSON round-trip — callers in tight refine loops with large reports will care. Neither needs code changes for v1.

egonSchiele (commented):
Reviewed the full diff with the frame model from this week's #513/#549/#550 work in hand. This is a strong landing: every finding from the design/plan review rounds is in here with a discriminating test, and two things go beyond what was asked — both correctly.

**Verified sound (the parts that could silently break):**
- **The `draftRegions` memo fixes a real bug the plan review missed.** During deserialize-mode replay, `getNewState` rotates frames (shift + push), so `stack.stack.length` mid-descent is the TOTAL serialized frame count, not the logical depth — a naively re-captured `entryDepth` on resume would be wrong. I walked the rotation math; memoizing at first execution and restoring from the serialized `other` is the correct fix, and the interrupt-resume fixture genuinely exercises it (the replayed `saveDraft` step is skipped, so the salvage can only come from serialized state).
- **The block-clear gate is right where my review's suggestion was wrong.** `return x` in a block lowers to `runner.halt(x)`, so gating on `!runner.halted` would never clear; gating on `!hasInterrupts(runner.haltResult)` gets all three cases (normal → clear, interrupt → keep, trip → thrown past).
- **guardId ownership** matches the review recommendation, and the payload addition is safe — I checked existing consumers, which interpolate individual fields; nothing asserts the exact error shape.
- Depth arithmetic is centralized (with `callerDepth` mirroring `callerFrame()`), the salvage is a named operation, the def-side clear uses structured block-form codegen, checker inference covers guard blocks with tests in both directions, and the fork idiom was verified against a real fixture before use.

**Findings, inline:** one correctness edge with disabled-dimension guards (empty `ids` → shared `""` region key — real interaction with #547/#550 semantics, three-line fix), one small serialized-state leak on thrown exits, and one docstring nit about clone semantics. Nothing blocking beyond the empty-ids fix, which I would make before merge.

packages/agency-lang/lib/runtime/drafts.ts
```diff
@@ -0,0 +1,130 @@
...
+
+// Branch-local, serialized store: frame depth -> its latest draft. Lives in
+// `StateStack.other` (NOT on `State`) because frames are popped by the unwind
+// before a guard boundary reads them — `other` outlives frame pops. Depth
+// arithmetic is centralized here so no caller ever touches a stack index. See
+// docs/superpowers/specs/2026-07-14-save-draft-guards-design.md.
+
+/** The draft map on this stack, or undefined if none written yet. `other` is
+ *  already typed `Record<string, any>`, so no cast is needed. */
+function peekDrafts(stack: StateStack): Record<number, DraftRecord> | undefined {
```
egonSchiele, line 19: Why not write these as methods on the state stack? Feel silly to write them as functions that take a state stack, when state stack is a class and you could just add them as methods. And then everything would be in one place.

packages/agency-lang/lib/runtime/drafts.ts
```diff
@@ -0,0 +1,130 @@
...
+function ensureDrafts(stack: StateStack): Record<number, DraftRecord> {
+  if (!stack.other.drafts) stack.other.drafts = {};
+  return stack.other.drafts as Record<number, DraftRecord>;
+}
+
+/** Depth of the frame that CALLED the current TS helper — the Agency scope
+ *  whose draft this is. Mirrors `StateStack.callerFrame()` (one frame below the
+ *  helper's own top frame). -1 when there is no caller (module-init/global). */
+function callerDepth(stack: StateStack): number {
+  return stack.stack.length - 2;
```
egonSchiele, line 32: 2 is a magic number here

packages/agency-lang/lib/runtime/drafts.ts
egonSchiele, line 39: What is depth here and is depth something the caller should track or is it something the state stack should track?

packages/agency-lang/lib/runtime/drafts.ts
```diff
@@ -0,0 +1,130 @@
...
+ *  shifts by the restored block frame — the marker must be resume-stable, like
+ *  the guard's own `ids`. Cleared by `clearDraftRegion` on guard exit. */
+export function draftRegionStart(stack: StateStack, key: string): number {
+  const regions = ensureRegions(stack);
+  if (regions[key] === undefined) regions[key] = stack.stack.length;
+  return regions[key];
+}
+
+function clearDraftRegion(stack: StateStack, key: string): void {
+  const regions = stack.other.draftRegions as Record<string, number> | undefined;
```
egonSchiele, line 67: I'm not sure I understand why regions are needed. Doesn't the state stack itself imply the scope that we should be writing the save draft value to?

egonSchiele:
Closing in favor of a ground-up redesign rather than reworking in place: the architecture here (side map keyed by frame depth, guard regions, sweep, two clearing paths) is superseded by the carry-the-draft-on-the-abort design, which deletes all of it. See `docs/superpowers/specs/2026-07-15-save-draft-carry-on-abort-redesign.md` (revision 2) for the agreed design and the reasoning — including the level rule, which is a deliberate behavior change (no deep-fallback salvage), so this PR's semantics should not merge first.

Surviving pieces of this PR (the execution fixtures, the type-checker work, and the stdlib surface) get ported to the new branch rather than rewritten. The two open review findings here (empty-ids region collision, region-key leak) are moot under the redesign — the constructs they live in no longer exist.

PR #552, opened 2026-07-15
verify: execute checks + honor an authoritative expert checklist; drop Feedback.data

egonSchiele:
## What this does

Makes the fresh-eyes `verify` step actually **catch** numeric/quantitative failures instead of describing them, and fixes a structured-output regression on Anthropic. Three linked changes:

### 1. Drop `Feedback.data?: any` (fixes Anthropic 400 on every task)
Provider-native (schema-enforced) structured output rejects an empty `{}` schema. Anthropic returns:
```
output_config.format.schema: Empty schema ({}) that accepts any JSON value is not supported.
```
The only `any` in verify's schema was `Feedback.data`, which serialized to `{}` and **400'd `verifyInner` on every task** (verify silently fail-opened everywhere). `data` was an unused debug escape hatch. Removed from the `std::agency` `Feedback` type and the structurally-identical copy in the review subagent.

### 2. verify executes checks (compute, don't describe)
`verifySysPrompt` + the analysis prompt now require verify to **enumerate every explicit success criterion** and **compute each one with `exec`** — running any ground-truth tool the task names (e.g. `primer3`'s `oligotm` with its exact flags) rather than eyeballing. A criterion it did not actually compute counts as a gap; a numeric near-miss is reported with the actual-vs-required numbers.

### 3. verify honors an authoritative expert checklist
`verify(task, criteria)` gains an optional acceptance checklist. `codingAgent` threads it through to `verify`, and `expertAgent` passes `guidance.checklist` — so the **same** expert criteria drive both building and checking, closing the independent blind spot where the solver and verify each re-derived (and each got wrong) a subtle detail like "compute Tm on the annealing portion only." The expert prompts now demand **executable** checklist items that name the exact tool/flags/sub-sequence to check.

## Why
Observed on Terminal-Bench: verify was 400-ing on every complex task (`Feedback.data` → empty schema), and even once un-broken it *described* rather than *computed*, so it rubber-stamped artifacts that failed the hidden grader (e.g. a primer with Tm just out of range). These changes make verify compute the task's own checks and enforce the expert's precise criteria.

## Notes
- Builds clean (`make`). The `verify`/`codingAgent` signature additions are optional params, so existing callers are unaffected.
- Depends on the already-merged smoltalk native-structured-output upgrade (which is what surfaced the empty-schema limitation).

### DIFF
--- a/packages/agency-lang/stdlib/agency.agency
+++ b/packages/agency-lang/stdlib/agency.agency
@@ -376,8 +376,7 @@ type error, or the reviewer judged it does not do the task); `error:
 false` is advisory feedback, including typecheck warnings. */
 export type Feedback = {
   error: boolean;
-  feedback: string;
-  data?: any
+  feedback: string
 }
 
 export def mergeFeedback(
@@ -415,10 +414,10 @@ def parseFeedback(source: string): Result<Feedback[]> {
 def reportFeedback(report: TypeCheckReport): Feedback[] {
   const feedback: Feedback[] = []
   for (error in report.errors) {
-    feedback.push({ error: true, feedback: "Error: ${error.message}", data: error })
+    feedback.push({ error: true, feedback: "Error: ${error.message}" })
   }
   for (warning in report.warnings) {
-    feedback.push({ error: false, feedback: "Warning: ${warning.message}", data: warning })
+    feedback.push({ error: false, feedback: "Warning: ${warning.message}" })
   }
   return feedback
 }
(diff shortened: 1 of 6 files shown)


packages/agency-lang/stdlib/agency.agency
```diff
@@ -549,32 +548,46 @@ static const verifyTools: any[] = [
...
-def verifyInner(task: string): Feedback[] {
+def verifyInner(task: string, criteria: string[] = []): Feedback[] {
   let items: Feedback[] = []
   thread(label: "verify") {
     systemMessage(verifySysPrompt)
-    const analysis: string = llm("The task was:\n\n${task}\n\nInspect the working directory, run the artifact against the task's own success criteria, and describe what passes and what fails.", { tools: verifyTools })
+    // When a caller supplies an expert checklist, it is authoritative — verify
+    // must check those exact items (with the exact tool/flags/sub-sequence they
+    // name) rather than re-deriving looser criteria of its own.
+    const criteriaBlock = if criteria.length == 0 then "" else "\n\nThese domain-specific acceptance criteria are AUTHORITATIVE — check EACH one exactly as written, using the exact tool, flags, and input sub-sequence it names:\n- ${criteria.join("\n- ")}"
```
egonSchiele: This if statement is really ugly. It's funny, I had not added a one line if to this language for a long time, because I was worried agents would misuse it. I finally added it and here we are. Your else has a ridiculously long string. Please put it in a separate variable first. Use a multi-line string instead of adding "\n"s. Use the variable in the else branch.

egonSchiele:
Addressed in cb5771e:

- **@egonSchiele** — pulled the long else-string out of the one-line-if into a multi-line `authoritativeCriteria` variable; the `if` now just selects it. (Sorry for the one-liner abuse.)
- **Copilot (stale docs ×2)** — regenerated the stdlib reference docs via `make` (`agency doc`): `agency.md` now shows `Feedback` without `data` and `verify(task, criteria: string[] = [])`; `agents/coding.md` shows `codingAgent`'s new `criteria` param.

PR #553, opened 2026-07-16
saveDraft: carry the draft on the abort (level rule)

egonSchiele:
Rebuilds `saveDraft` so that **an aborted function returns its draft**. Reworked during review: aborts now travel between frames as return values, not exceptions. Design spec: `docs/superpowers/specs/2026-07-15-save-draft-carry-on-abort-redesign.md` (revision 3).

## The mechanism

`saveDraft(v)` records a `savedDraft` slot on the calling frame (`StateStack.setSavedDraft`). When an abort stops a function, its frame catches the abort and returns an `AbortedResult`: a marker, the cause, and the frame's saved draft as its partial. The generated post-call check — emitted at the same sites as the interrupt check, so both propagation systems cover identical ground by construction — spots the marker and stops the caller too. An abort travels up the stack the way interrupts do.

Every salvage rule is structural; nothing is enumerated in the compiler:

- **Return position passes a partial through** because returning a value is what return statements do. `return verify()` returns verify's forced return, aborted or not.
- **Statement sites drop the callee's partial** and carry the frame's own draft (`carryThrough`). Salvage is opt-in per level.
- **Argument position drops at the call boundary itself**: `__call`/`__callMethod` refuse aborted arguments and forward the abort minus its partial. The review's cross-type repro (`return f(arr[0]())`) is impossible by construction for DIRECT argument position — no AST analysis exists to get wrong. (A container-wrapped abort, `f([g()])`, rides the same bounded gap an Interrupt[] does today; documented at the site and in the spec.)
- **The fork boundary is one `.then` in `startInvoke`** that turns an aborted branch value back into a rejection, partial dropped. Every join path keeps working on rejections; aborted values can never be cached as branch results.

Exceptions remain only where they're honest: inside a single frame (a cancelled `llm()` rejects; the frame converts at its catch) and above node level, where nodes rebuild the exception so the graph engine, CLI entry, and root budgets see aborts exactly as before. `__tryCall` keeps a thin exception backstop for trips thrown from runtime code with no compiled frame in between; the guard's owned-trip conversion salvages the value form: `success(partial)` when one arrived, exactly the old failure otherwise.

`AbortedResult` is immutable — each hop builds a new one, so there is no shared mutable object riding an exception — and its methods own the statelog trail (`abortUnwind` span; `carried` / `erased` / `droppedAtArgPosition` / `clearedAtFork` / `delivered` events, with 500-char previews).

## Deliberate behavior changes

- **vs #551:** the deep-fallback salvage is removed. A draft saved only in a deep callee, consumed via assignment, is not salvaged — `save-draft-deep-only` pins the failure; `save-draft-return-chain` pins the opt-in via an unbroken return chain. No shipped fixture pinned the old behavior.
- **vs revision 2 of this spec:** at return position the callee's partial wins even when the caller saved a draft, mirroring the success path (a successful `return verify()` also ignores the caller's draft). No fixture pinned the old ordering.

## What got deleted

Relative to #551: the side map (`drafts.ts`), regions, the boundary search, the sweep, both clearing paths, and the resume memo. Relative to this PR's own first iteration: the `returnCarry` flag, the return-position marking codegen, the `containsFunctionCallDeep` gate and its soundness hole, and the runBatch draft-clearing helper.

## Review-round fixes

- `checkSaveDraftCall` checks named-argument calls and stands down for a user-defined `saveDraft` (both were bypasses).
- `saveDraft` in global scope throws with a clear message instead of silently no-oping.
- The draft write is encapsulated in `StateStack.setSavedDraft`.

## Testing

- 17 execution fixtures green, including the new `save-draft-arg-position` (the review repro: partial dropped at the call boundary, and the outer call of `return f(g())` passing its partial through)
- Full `tests/agency/guards/` sweep: 51/51; subprocess sample: 4/4; agency-js plain-error pin green
- 18 unit tests on `AbortedResult`, 5 on the guard-boundary conversion, 2903+ runtime/typechecker, 420 codegen after fixture regen; structural lint clean

## What this is shaped for

`finalize` (PR B) slots into the two stop sites (`fromError` / `carryThrough`), and the assignment-site binding it needs falls out of the value transport. The resumable-guards follow-up spec's reject path runs this machinery unchanged.


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -191,6 +191,20 @@ function walkMarkTopLevel(value: unknown): unknown {
...
+ *  Used by the return-position pass-through gate: `return f(g())` must
+ *  compile unmarked, because g's trip would be wrongly typed for this
+ *  scope if it picked up the mark. Walks raw AST objects — conservative
+ *  by construction (any nested call, even a JS-level method chain that
+ *  parses as a functionCall, disables the mark). */
+function containsFunctionCallDeep(node: unknown): boolean {
+  if (node == null || typeof node !== "object") return false;
+  if (Array.isArray(node)) return node.some(containsFunctionCallDeep);
+  const rec = node as Record<string, unknown>;
+  if (rec.type === "functionCall") return true;
```
egonSchiele: **Soundness gap in the return-position pass-through gate.**

`containsFunctionCallDeep` only matches `type === "functionCall"`, but a call written on a **non-identifier base** compiles to a `valueAccess` whose trailing element is `{ kind: "call", arguments, block }` — with no embedded `functionCall` node. (`kind: "methodCall"` is safe because it embeds `functionCall: FunctionCall`; `kind: "call"` is not.)

So `return f(arr[0]())` is **not** detected as containing a nested call → `isMarkableReturnCall` returns true → the return is marked. `arr[0]()` is evaluated inside the `__markReturnCarry` wrapper, so if it trips carrying a draft, its partial — typed as `arr[0]`'s return type, i.e. `f`'s *parameter* type — passes through to the guard as `f`'s *return* type. That's exactly the cross-type salvage the PR body says the level rule "deletes rather than documents."

Verified the AST shape by parsing `return f(arr[0]())`:
```
arg = { type: "valueAccess", base: arr,
        chain: [ {kind:"index", index:0}, {kind:"call", arguments:[]} ] }
containsFunctionCallDeep(arg) === false   // → return gets marked
```

Repro that lands a string where the guard block's type is `number`:
```
def leaf(): string { saveDraft("partial"); sleep(10s); return "x" }   // trips on the guard
def f(x: string): number { return x.length }
node main() {
  const arr = [leaf]
  const r = guard(time: 500ms) as { return f(arr[0]()) }
  // r.value === "partial" (a string) — should be number-typed, or a failure
}
```

Suggested fix — treat call-bearing chain elements as calls:
```ts
if (rec.type === "functionCall") return true;
if (rec.kind === "call" || rec.kind === "methodCall") return true;
```
(`methodCall` is already covered transitively via its embedded `functionCall`, but adding it is cheap insurance if the element shape ever changes.) Also worth a fixture — none of the four new pins exercise a call in return-argument position, so this path is currently untested.

packages/agency-lang/lib/runtime/carriedDraft.ts
```diff
@@ -0,0 +1,100 @@
...
+    partial: previewForLog(next !== undefined ? next.value : prev?.value),
+  });
+}
+
+/** Return-position marker (level-rule rung 3). The compiler wraps the
+ *  OUTERMOST call of every `return <call>(...)` statement (and the block
+ *  equivalent that lowers to runner.halt) in try/catch and calls this
+ *  before rethrowing. Argument subexpressions are evaluated BEFORE the
+ *  wrapped call, so `return f(g())` with g tripping stays unmarked —
+ *  only f's own trip (f-return-typed = this scope's return type) may
```
egonSchiele: This docstring contradicts the implemented gate. It says `return f(g())` "stays unmarked — only f's own trip ... may pass through," which implies `f`'s partial *does* pass through. But `isMarkableReturnCall` disables the mark for the **whole** statement whenever the arguments contain any nested call (`!containsFunctionCallDeep(value.arguments)`), so `return f(g())` is unmarked entirely — `f`'s own trip does **not** pass through either (rung 4 erases). The `isMarkableReturnCall` docstring ("`return f(g())` compiles unmarked and rung 4 erases instead") is the accurate description; this one should be reconciled with it.

egonSchiele (commented):
Reviewed the carry-on-abort mechanism end-to-end against `origin/main`. The four-rung level rule cleanly resolves the block-frame-erase problem I flagged in the plan review: return-position pass-through carries a tail-returned callee's draft across the guard block, which is what lets `save-draft-outermost` pass. Confirmed the pieces:

- **Templates:** `functionCatchFailure` keeps the `RestoreSignal`-first ordering and calls `__stampCarriedDraft` (4-arg signature matches the runtime — the plan's stale `paramNames` arg was correctly dropped). `blockSetup`'s new catch correctly no-ops on non-aborts: `RestoreSignal extends Error`, not `AgencyAbort`, so a resume signal is never stamped even without an explicit guard.
- **Guard boundary:** `__tryCall` reads `carriedDraft` on the owned trip, closes the unwind span, and is exactly today's failure otherwise.
- **Fork isolation:** `clearCarriedDraftAtForkBoundary` is called at all three branch-rethrow sites (fork-all L609, `runRaceFirstTime` L731, `runRaceResume` L907) — complete, not just the fork-all path.
- **`__stampCarriedDraft`:** `finalizeResult ?? savedDraft ?? passThrough` is correct because `savedDraft`/`carriedDraft` are always object-wrapped, so a saved `null` stays distinct from "no draft".

One correctness finding inline (a soundness gap in the pass-through gate for call-in-return-argument forms like `return f(arr[0]())`), plus a docstring nit. Everything else looks solid.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -3107,6 +3121,36 @@ export class TypeScriptBuilder {
...
+   *  `return f(...)` declares that f's value IS this scope's value — the
+   *  checker enforced the types — so a trip escaping f may carry f's
+   *  partial through this frame unchanged. The mark is sound only when no
+   *  OTHER Agency call evaluates inside the wrapped statement: a nested
+   *  call's partial is not return-typed for this scope. So the mark
+   *  applies only when the arguments contain no function calls;
+   *  `return f(g())` compiles unmarked and rung 4 erases instead
+   *  (conservative). llm() returns hoist through their own path and are
+   *  never marked. */
+  private isMarkableReturnCall(value: AgencyNode): boolean {
```
egonSchiele: The fact that we have to look for these manually is a code smell. It means that there are probably other cases that we haven't thought of. I don't think it's possible to list out all the cases and check them like this. Instead, we should be shooting for an architecture where the desired result naturally falls out of the architecture. We have talked about how the partial value should become the return value for the function. I'll again suggest that using exceptions is not the right answer here, because it means we have to check cases like this. Instead, if execution gets aborted, if the function returns its partial value as a return value, as well as some sort of indicator that its run was aborted, I think a lot of behavior falls out of that naturally. Let's take this code snippet as an example:

  ```
    guard(time: 5m) as {
      return code()      // code is at depth 1
    }
    def code() {
      saveDraft(10)      // filed at depth 1
      return verify()    // verify is at depth 2
    }
    def verify() {
      saveDraft(1)       // filed at depth 2
      // ... trip happens here ...
  }
  ```

Let's say the `verify` function gets aborted midway. What we're doing currently is raising an exception from verify. Then in the code function, we get a note riding along with the exception, but we have to figure out where it's coming from, what function, and what to do with that based on that function's position. This is going to lead to a combinatorial explosion of possible cases that are going to be very, very tedious to figure out.

Instead, let's say that if verify gets aborted, it returns its partial value, plus a marker saying that it was aborted (or sets a flag on the runner). Code sees that return value and the marker, so it knows it needs to abort too. It has a limited number of places to check what to do next.

packages/agency-lang/lib/runtime/carriedDraft.ts
```diff
@@ -0,0 +1,100 @@
...
+ *  (stored on the abort), and every transition involving a partial emits
+ *  one abortSalvage event. Empty-to-empty transitions are silent. */
+export function __stampCarriedDraft(
+  error: unknown,
+  frame: State,
+  scopeName: string,
+  ctx: RuntimeContext<any>,
+  finalizeResult?: { value: unknown },
+): void {
+  if (!(error instanceof AgencyAbort)) return;
```
egonSchiele: Write this logic in a more declarative fashion so it's easy to understand, please. Consider adding functions to the AgencyAbort class that encapsulate the complexity.

packages/agency-lang/lib/runtime/carriedDraft.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/** Return-position marker (level-rule rung 3). The compiler wraps the
+ *  OUTERMOST call of every `return <call>(...)` statement (and the block
+ *  equivalent that lowers to runner.halt) in try/catch and calls this
+ *  before rethrowing. Argument subexpressions are evaluated BEFORE the
+ *  wrapped call, so `return f(g())` with g tripping stays unmarked —
+ *  only f's own trip (f-return-typed = this scope's return type) may
+ *  pass through. */
+export function __markReturnCarry(error: unknown): void {
```
egonSchiele: You have all these functions that take an exception. Would it make sense to just make them methods on the exception itself?

packages/agency-lang/lib/runtime/carriedDraft.ts
```diff
@@ -0,0 +1,100 @@
...
+ *  turn this abort into a value. Emit the closing event and end the
+ *  unwind span. No-op when the unwind never touched a partial. Reads the
+ *  statelog client via ALS because result.ts's __tryCall has no ctx
+ *  parameter; null-safe so a trip surfacing outside any runtime frame
+ *  never crashes on telemetry. */
+export function closeUnwindSpan(abort: AgencyAbort): void {
+  if (abort.unwindSpanId === undefined) return;
+  let client;
+  try {
+    client = getRuntimeContext()?.ctx?.statelogClient;
```
egonSchiele: Why is this in a try/catch? What exception do you expect to get thrown?

packages/agency-lang/lib/runtime/errors.ts
```diff
@@ -145,6 +145,26 @@ export class AgencyAbort extends Error {
...
 
+  /** saveDraft partial carried up the unwind (carry-on-abort spec). Every
+   *  generated catch rung REPLACES this with the unwinding frame's own
+   *  partial — its draft, or (future) its finalize's return — or erases it
+   *  when the frame has neither. Deliberately mutable: every rung rethrows
+   *  the SAME object, which is what makes the level rule's replace-and-erase
+   *  work. Same idiom as agencyCause.delivered. */
+  carriedDraft?: { value: unknown };
+
+  /** Level-rule rung 3 (return-position pass-through): set by the
```
egonSchiele: These comments are pretty hard to read, please write more readable comments. Avoid using jargon and garden path sentences. Use the writing tips I gave you earlier.

packages/agency-lang/lib/runtime/runBatch.ts
```diff
@@ -658,6 +660,27 @@ export async function runBatch<T>(
...
+  reason: unknown,
+  ctx: RuntimeContext<any>,
+): void {
+  if (reason instanceof AgencyAbort && reason.carriedDraft !== undefined) {
+    ctx.statelogClient?.abortSalvage({
+      action: "clearedAtFork",
+      spanId: reason.unwindSpanId,
+      partial: previewForLog(reason.carriedDraft.value),
+    });
+    reason.carriedDraft = undefined;
```
egonSchiele: This is really poor encapsulation. There should be a clearCarriedDraft method on `reason` that you call instead. I also wonder if the state log logs should be part of these methods, instead of being called directly here. Do you see why this is bad encapsulation?

packages/agency-lang/lib/stdlib/thread.ts
```diff
@@ -335,6 +335,27 @@ export async function _popGuard(ids: string[]): Promise<void> {
...
+ * runs, the top frame is saveDraft's own frame and callerFrame() is the
+ * user's scope. During module-level init there is no user scope below it
+ * — callerFrame() would return saveDraft's OWN frame (or throw on an
+ * empty stack), and a draft written there dies unread when the frame
+ * pops. So with no caller scope, saveDraft is a deliberate no-op.
+ */
+export async function _saveDraft(value: unknown): Promise<void> {
+  const { stack } = getRuntimeContext();
+  if (stack.isGlobalContext()) {
+    return;
```
egonSchiele: Should we throw an error of some kind here? Because users shouldn't be calling save draft in the global scope. It doesn't make sense.

packages/agency-lang/lib/stdlib/thread.ts
```diff
@@ -335,6 +335,27 @@ export async function _popGuard(ids: string[]): Promise<void> {
...
+  }
+  stack.callerFrame().savedDraft = { value: deepClone(value) };
```
egonSchiele: Why not have a setSavedDraft function on the stack instead?

packages/agency-lang/lib/typeChecker/checker.ts
```diff
@@ -245,10 +245,31 @@ function checkFunctionCallsInScope(
...
+ *  undefined) is skipped. Name-keyed: aliasing `saveDraft` (`const s = saveDraft;
+ *  s(v)`) escapes the check (documented v1 limitation). */
+function checkSaveDraftCall(
+  call: FunctionCall,
+  info: ScopeInfo,
+  ctx: TypeCheckerContext,
+): void {
+  if (call.functionName !== "saveDraft") return;
+  if (!info.returnType) return; // no declared/inferred return type in scope
+  if (call.arguments.length !== 1) return;
```
egonSchiele, line 277: should this be a type error?

egonSchiele: Or is that going to get checked separately? Have you added saveDraft to the array of built-in functions that the type checker uses?

egonSchiele: Is this going to incorrectly fire on a user-defined `saveDraft` function?

packages/agency-lang/lib/typeChecker/checker.ts
```diff
@@ -245,10 +245,31 @@ function checkFunctionCallsInScope(
...
+  const arg = call.arguments[0];
+  if (arg.type === "splat" || arg.type === "namedArgument") return;
+  checkType(arg, info.returnType, info.scope, "the saveDraft() draft value", ctx);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 48c920a3b — `checkSaveDraftCall` now unwraps a `namedArgument` and checks the value it carries. Test added: "rejects a mismatched draft passed as a NAMED argument".

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Confirmed — and this finding ended up driving a bigger change than a gate fix. The transport rework (92002a263) makes the hole impossible by construction instead of patching the enumeration: aborts now travel as return values, and an aborted value entering ANY call as an argument is refused at the `__call`/`__callMethod` boundary and forwarded with its partial dropped. There is no AST gate left to miss a case — `containsFunctionCallDeep` is deleted.

Your repro is pinned nearly verbatim in `save-draft-arg-position.agency` (node `argPositionDrops`), expecting the failure rather than a string salvaged into a number-typed result.

packages/agency-lang/lib/runtime/carriedDraft.ts
egonSchiele: Both this docstring and the gate it contradicted are gone with the marking mechanism. Under the value transport, `return f(g())` handles both directions correctly with no special casing: g aborting is dropped at the call boundary (argument position), and f's own trip passes f's partial through return position — f's partial is f-return-typed, which is the scope's return type. The second half is pinned in `save-draft-arg-position.agency` (node `outerCallSalvages`).

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Reworked as you proposed (92002a263). An aborted function now returns an `AbortedResult` — a marker, the cause, and its saved draft as the partial — and the caller's post-call check stops the caller in turn. Aborts travel between frames exactly like interrupts; exceptions survive only inside a single frame (a cancelled leaf op rejects, and that frame converts at its catch) and above node level, where the exception is rebuilt so the graph engine, entry, and root budgets are untouched.

Your walked example now reads: verify returns its partial `1`; `code` sees the marker at `const x = verify()`, stops, and returns its own draft `10`; the block's `return code()` passes that through because returning a value is what return statements do. The case analysis is gone — the post-call checks ride the same emission point as the interrupt checks, so the enumeration of call sites is shared with a battle-tested system rather than duplicated.

One semantic refinement fell out and is worth your eyes: at return position the CALLEE's partial now wins even when the caller also saved a draft, mirroring the success path (a successful `return verify()` also ignores the caller's draft). Revision 2 ranked the caller's draft higher; no fixture pinned that ordering. Spec updated to revision 3.

packages/agency-lang/lib/runtime/carriedDraft.ts
egonSchiele: `__stampCarriedDraft` is gone. The rule now lives as methods on an immutable `AbortedResult` — `fromError`, `carryThrough`, `droppedAtArgPosition`, `atForkBoundary`, `deliver` — each stating one hop declaratively, with statelog inside the methods. Every hop builds a new instance, so there is no mutation ordering to reason about.

packages/agency-lang/lib/runtime/carriedDraft.ts
egonSchiele: They are methods now — on `AbortedResult` rather than the exception, because under the value transport the exception's role shrank to a single frame's extent and the value is what actually travels. The one crossing back (`toError`, for the node boundary and fork joins) is also a method on it.

packages/agency-lang/lib/runtime/carriedDraft.ts
egonSchiele: The exception I expected was `getRuntimeContext()`'s own — it throws by design outside an ALS frame. Catching it was the wrong fix; the right one was the non-throwing accessor: `agencyStore.getStore()?.ctx?.statelogClient`. No try/catch remains.

packages/agency-lang/lib/runtime/errors.ts
egonSchiele: Rewritten — and mostly gone: the fields these comments described (`carriedDraft`, `returnCarry`, `unwindSpanId`) no longer exist on `AgencyAbort`. The `AbortedResult` docs that replaced them are written as plain sentences, one idea each.

packages/agency-lang/lib/runtime/runBatch.ts
egonSchiele: Yes — three things made it bad encapsulation: runBatch read the abort's private fields to narrate the abort's own story; every call site threaded `ctx` for telemetry that was not runBatch's concern; and any change to the salvage trail meant editing every consumer instead of one method. The helper is deleted. The fork boundary is now a single `.then` in `startInvoke` calling `value.atForkBoundary()` — a method that drops the partial and logs itself — and the three join-site edits are gone entirely, because converting the value back to a rejection there means the join paths never learn aborted values exist.

packages/agency-lang/lib/stdlib/thread.ts
egonSchiele: It throws now (`StateStack.setSavedDraft`): "saveDraft() can only be called inside a function, node, or block — there is no enclosing scope at module top level to save a draft for." A checker diagnostic for the statically-detectable case would need a new registry code plus an explanation page (the coverage gate), so I left that as a follow-up rather than fold it in here.

packages/agency-lang/lib/stdlib/thread.ts
egonSchiele: Done — `StateStack.setSavedDraft(value)` owns the caller-frame targeting, the deep clone, and the global-scope rejection. `_saveDraft` is a one-liner that reads the stack from ALS and delegates.

packages/agency-lang/lib/typeChecker/checker.ts
egonSchiele: A scope with no declared or inferred return type has nothing to check the draft against, and untyped/`any` scopes are fail-open across the whole checker — this follows that convention rather than inventing a stricter rule for one call. If we ever flip the checker's fail-open default, this site flips with it.

egonSchiele: It is not in `BUILTIN_FUNCTION_TYPES` — saveDraft is a stdlib def, so arity and its parameter check generically through the imported signature. But its parameter is `any` (a draft can be any shape), which is exactly why `checkSaveDraftCall` exists: the against-the-enclosing-return-type rule is the only enforcement with teeth. The two bypasses that existed around it — named arguments and user-defined shadowing — are fixed in 48c920a3b.

egonSchiele: It did — fixed: the check stands down when `saveDraft` resolves to a user-defined function or node, whose arguments then check against its own signature like any other call. Test added: "does not fire on a user-defined function named saveDraft".

packages/agency-lang/lib/runtime/call.ts
```diff
@@ -6,6 +6,31 @@ import {
   describeFailureCallTarget,
 } from "./failurePropagation.js";
 import { isFailure } from "./result.js";
+import { type AbortedResult, isAborted } from "./abortedResult.js";
+
+/** Find an aborted result among a call's arguments (`f(g())` where g was
+ *  aborted). Checked at the call boundary itself so the rule holds for
+ *  every call shape — nested arguments, method chains, named args —
+ *  without the compiler having to enumerate them. */
+function findAbortedArg(descriptor: CallType): AbortedResult | undefined {
```
egonSchiele, line 20: `findAbortedArg` only inspects top-level positional/named args, so it catches `f(g())` but not an `AbortedResult` wrapped in a literal container: `f([g()])` hands `f`'s `__call` an *array* (`[AbortedResult]`), and `isAborted(theArray)` is false, so the aborted callee smuggles into `f` as a poisoned argument.

The impact is bounded — the signal stays aborted so `f`'s own next leaf op re-trips, and guard trips are re-thrown independently by the runner's `shouldSkip` — and it's exact parity with how the interrupt system treats `[g()]`. So not a regression. But it means the PR-body claim "impossible by construction — no AST analysis exists to get wrong" holds for *direct* arg position only, not container-wrapped. Suggest a one-line note here documenting the boundary ("only direct arg position; container-wrapped aborts ride the same gap interrupts do"), or confirm it's an accepted v1 limitation.

packages/agency-lang/lib/statelogClient.ts
```diff
@@ -30,6 +30,11 @@ export type SpanType =
   | "forkAll"
   | "race"
   | "handlerChain"
+  // One abort's unwind, from the first catch rung that touches a saveDraft
+  // partial to the guard that converts the abort into a value. The
+  // per-level abortSalvage events nest inside it. Opened lazily — a trip
+  // through undrafted code opens no span. See lib/runtime/carriedDraft.ts.
```
egonSchiele: Stale filename: this file is `lib/runtime/abortedResult.ts`, there is no `carriedDraft.ts` (it was one of the deleted names from the earlier iteration). Same fix wherever else that name lingers in comments.

egonSchiele (commented):
## Review: `saveDraft` — carry the draft on the abort

Read the runtime (`abortedResult.ts`, `call.ts`, `result.ts`, `runBatch.ts`, `stateStack.ts`), the codegen (`typescriptBuilder.ts` + the three templates), the checker change, and the new fixtures. This is a clean redesign — modeling aborts as return values that ride the exact same propagation sites as interrupts is the right call, and pushing every salvage rule into structure (return position = a plain return, statement/assignment = `carryThrough`, arg position = `__call` refusal, fork = one `.then`) instead of into AST classification removes a whole class of the bugs the earlier iteration had. `AbortedResult` being immutable with the statelog trail owned by its own methods reads well.

The fixtures are the strongest part: `save-draft-arg-position`, `save-draft-deep-only` vs `save-draft-return-chain`, and the fork-isolation set each pin one rule and its inverse. Nice.

A few notes below, only one of which is a possible correctness gap (and it's inherited from the interrupt model, not new).

### Correctness

- **Container-wrapped aborted arguments slip past `findAbortedArg`** (inline comment on `call.ts`). `f(g())` is caught at the boundary, but `f([g()])` / `f({k: g()})` hands `f`'s `__call` an *array/object* containing the `AbortedResult`, which `isAborted(arg)` doesn't see. The PR body's "impossible by construction — no AST analysis exists to get wrong" is true for *direct* argument position but overstated for container-wrapped ones. In practice the harm is bounded (the signal stays aborted, so `f`'s own next leaf op re-trips, and for guard trips the runner's `shouldSkip` re-throws independently), but `f` does run once with a poisoned argument. This is the *same* gap the interrupt system has (`[g()]` smuggles an `Interrupt[]` identically), so it's parity, not a regression — worth confirming that's the intended boundary and maybe softening the "by construction" claim.

- **`toError()` drops the `AgencyAbort` subclass** (`abortedResult.ts` → `new AgencyAbort(...)`). A cancel that was an `AgencyCancelledError`, or a `CallDepthExceededError`, comes back out of a fork/node as a plain `AgencyAbort`. I checked the consumers — everything downstream keys on `instanceof AgencyAbort` + `readCause(...).kind` (which is preserved by identity), so this is **safe as-is**. Flagging only because it's a silent identity change at the fork/node boundary; a one-line comment on `toError` saying "subclass identity is intentionally not reconstructed; consumers read `agencyCause.kind`" would save the next reader the same audit.

### Deliberate semantic changes (both intentional, both pinned)

- Return position: callee's partial wins even when the caller also saved a draft. This mirrors the success path, and `save-draft-arg-position`'s `outerCallSalvages` covers it — but it's a sharp edge for users: a caller that does `saveDraft(x); return callee()` silently discards `x` in favor of `callee`'s draft. Worth one sentence in the `saveDraft` docstring so it isn't discovered by surprise.
- Deep-fallback removal (`save-draft-deep-only`): agreed this is the right default (opt-in per level), and the paired fixture makes the intent legible.

### Nits

- `statelogClient.ts` stale filename reference (inline comment): "See lib/runtime/carriedDraft.ts" → `abortedResult.ts`.

### Testing

Coverage looks thorough: 17 execution fixtures pinning each rule + inverse, the arg-position review repro, `AbortedResult` unit tests, and the agency-js plain-error pin for the below-node exception path. The one thing not pinned by a fixture is the container-wrapped-argument case above — if you decide it's an accepted limitation, a short comment in `findAbortedArg` noting "only direct arg position; container-wrapped aborts ride the interrupt-parity gap" would document the boundary.

Overall: solid, well-structured, safe to land. None of the above blocks.

packages/agency-lang/lib/runtime/result.ts
```diff
@@ -175,6 +176,32 @@ export async function __tryCall(fn: () => any, opts?: FailureOpts): Promise<Resu
...
     // pauses; applies to any interrupting callee under `try`.)
     if (hasInterrupts(value)) return value as any;
+    // Aborted results travel as values (lib/runtime/abortedResult.ts).
+    // This boundary is where a guard turns its OWN trip into a Result:
+    // salvage the partial as a success, or produce exactly the failure a
+    // trip produced before saveDraft existed. A trip belonging to some
+    // OUTER guard keeps travelling as a value — `try` must not catch it.
+    if (isAborted(value)) {
+      const cause = value.cause;
+      if (
```
egonSchiele, line 186: what happens if cause is not a guard trip? Just return the value -- what happens to it then?

packages/agency-lang/lib/runtime/runBatch.ts
```diff
@@ -357,6 +358,22 @@ function startInvoke<T>(
...
+      // becomes a rejection again, for two reasons. First, isolation: the
+      // branch's partial stays in the branch (atForkBoundary drops it) —
+      // which branch fails first is a race, and one branch's value has
+      // the wrong shape for the fork. Second, every join path (all /
+      // sequential / race, first-run and resume) already handles branch
+      // rejections; converting here means none of them need to know
+      // aborted values exist, and an aborted value can never be cached
+      // as a branch result.
+      if (isAborted(value)) {
+        throw value.atForkBoundary().toError();
```
egonSchiele, line 376: is it weird that you're sometimes returning a value and sometimes throwing an error? Why do you need to be able to do both? Why not just always return a value?

packages/agency-lang/lib/typeChecker/checker.ts
```diff
@@ -245,10 +245,38 @@ function checkFunctionCallsInScope(
...
+function checkSaveDraftCall(
+  call: FunctionCall,
+  info: ScopeInfo,
+  ctx: TypeCheckerContext,
+): void {
+  if (call.functionName !== "saveDraft") return;
+  // A user-defined function named saveDraft is an ordinary function. Its
+  // arguments check against its own signature (checkSingleFunctionCall),
+  // not against the enclosing scope's return type.
+  if (ctx.functionDefs["saveDraft"] || ctx.nodeDefs["saveDraft"]) return;
```
egonSchiele, line 273: If the user imports the function, does that not show up in this array? What about if they re-export the function from std::thread? I wonder, is it possible that the behavior will change in the future so that imported functions do show up in this array?

packages/agency-lang/lib/typeChecker/checker.ts
```diff
@@ -245,10 +245,38 @@ function checkFunctionCallsInScope(
...
+  if (!info.returnType) return; // no declared/inferred return type in scope
+  if (call.arguments.length !== 1) return;
+  const first = call.arguments[0];
+  if (first.type === "splat") return;
```
egonSchiele, line 283: I don't really get this. Why would the first type being a splat cause a return?

egonSchiele: Because the splat's single AST node holds an ARRAY whose elements become the call's arguments at runtime — `saveDraft(...xs)` has one argument node but unknowable arity, and the value to check would be an element of `xs`, not `xs` itself. Checking the splat expression against the scalar return type would be a false error on every such call. The generic call checks still cover the splat's shape. Added a comment saying exactly this (it deserved one — same question, same answer, for the next reader).

packages/agency-lang/lib/typeChecker/checker.ts
egonSchiele: It did not — and your instinct about the future was the real bug. `functionDefs` holds locally-defined functions only; imports land in `importedFunctions`, so a saveDraft imported from a USER module slipped past the stand-down and wrongly got the draft check. Fixed by keying on the import's `originFile`: the check now fires only when the name is unresolved (fail-open, where the only saveDraft in practice is the stdlib one — unresolved imports already raise their own diagnostics) or resolves to the stdlib thread module. Because the gate is positive, it stays correct if import resolution later populates `importedFunctions` in more cases, including a re-export. Two tests added (user-module import stands down; stdlib origin still fires).

packages/agency-lang/lib/runtime/runBatch.ts
egonSchiele: The mixed modes are deliberate, and the comment now says why (expanded in 0a1a…): compiled code speaks values, but runBatch's join machinery speaks settled promises — a branch failure IS a rejection in every join path (allSettled status, the race's first-rejection seal, sequential's try/catch, result caching). This `.then` is the adapter between the two worlds, at the single point every branch result passes through. Always-return-a-value would mean teaching all six join paths (three modes × first-run/resume) a third settled state, on top of the freshly-stabilized #549 charge/resume ordering. If we later migrate the joins to value-based settlement wholesale, this adapter is the one line that disappears.

packages/agency-lang/lib/runtime/result.ts
egonSchiele: A non-guardTrip abort (Esc, kill, race loss) or an outer guard's trip passes through unchanged — `try` must never swallow those. Then the caller's post-call check keeps it moving: defs and blocks stop with their own drafts, and when it reaches node level the node rebuilds the exception (`toError`), so a kill terminates the run through the graph engine exactly as before, and an outer guard's trip eventually reaches the guard that owns it and converts there. Expanded the comment at this site to say so.

packages/agency-lang/lib/statelogClient.ts
egonSchiele: Fixed here and in statelogClient (the only two that lingered); grep for carriedDraft in lib/ is clean now.

packages/agency-lang/lib/runtime/call.ts
egonSchiele: Accepted v1 limitation, now documented at the site: direct argument position only, and `f([g()])` rides the same gap an Interrupt[] does today — bounded because the abort signal keeps firing, so the callee's next leaf op re-trips, and the wrapped value never surfaces as the call's own result, so no wrongly-typed salvage can reach a guard. The spec's shared-envelope note lists it explicitly alongside the binOp-operand case. I also softened the PR-body claim to "direct argument position".

packages/agency-lang/stdlib/thread.agency
```diff
@@ -254,6 +255,23 @@ export def guard(
   return result
 }
 
+export def saveDraft(value: any) {
+  """
+  Record a best-so-far value for the current function or guarded block. If an
+  enclosing `guard(...)` trips before this scope returns, the guard yields the
+  last saved draft instead of a failure — an "anytime" result you can always
+  fall back to. Call it repeatedly as your result improves; the last value
+  wins. With no enclosing guard it is a harmless no-op.
```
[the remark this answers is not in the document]
egonSchiele: Fixed — the docstring now says calling saveDraft at module top level is an error (no enclosing scope to save a draft for); the no-guard no-op sentence stays, since that case remains true inside functions. Regenerated stdlib docs pick it up.

packages/agency-lang/lib/statelogClient.ts
```diff
@@ -30,6 +30,11 @@ export type SpanType =
...
+  | "abortUnwind"
```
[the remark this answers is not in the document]
egonSchiele: Fixed together with the runtime reference in the same file.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -3278,17 +3294,46 @@ export class TypeScriptBuilder {
...
+   * scopes halt with their own AbortedResult (carryThrough applies the
+   * salvage rule); every other scope rebuilds the exception, because
+   * nothing above compiled code consumes aborted values. */
+  private assignmentAbortedGuard(varRef: TsNode): TsNode {
+    const scopeType = this.scopes.current().type;
+    const expr = this.str(varRef);
+    if (scopeType === "function" || scopeType === "block") {
+      const frameVar = scopeType === "block" ? "__bstack" : "__stack";
+      const scopeName = JSON.stringify(this.scopes.currentName());
+      return ts.if(
```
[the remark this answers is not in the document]
egonSchiele: Confirmed and fixed — `assignmentAbortedGuard` now checks `insideHandlerBody` first and emits the throw form (inline handler bodies do compile in a function/block scope, so the scope-type switch alone mis-routed them; verified via compiled probe both before and after). Your comment also surfaced a second layer: a DEF used as a handler is a plain compiled def and returns an AbortedResult to the handler chain, which would have hit the invalid-verdict error and swallowed the abort's cause. The chain now rethrows it (`interruptWithHandlers` handler loop), restoring the pre-transport behavior of a throwing handler.

PR #554, opened 2026-07-16
Guards follow-ups: labels, top-level saveDraft diagnostic, guide docs

egonSchiele:
Three small guards/saveDraft follow-ups from the #553 review discussion, bundled per owner request.

## Guard labels

`guard(label: "research", cost: $0.50)` names a guard. On a trip the label rides the `guardTrip` cause and surfaces on the failure (`error.label`, `null` when unlabeled) and in error messages, so code with several guards can tell which one fired. The label serializes with the guard (interrupt/resume) and a `TimeGuard`'s per-branch clone keeps the parent's label, matching how the clone keeps the parent's `guardId`. This is increment 1 of the resumable-guards design, where the label becomes the handler's routing key.

Pinned by `tests/agency/guards/guard-label.agency` (labeled and unlabeled trip side by side) plus unit tests for the message, cause, serialization round-trip, and branch clone.

## AG6031: saveDraft at module top level

The runtime already throws; the checker now catches it at compile time. New registry entry + explanation page under AG6 (calls, tools, and LLM usage), emitted from `checkSaveDraftCall` when the walk is in the synthetic top-level scope. Diagnostics pages regenerated.

## Guards guide: saveDraft + labels

`docs/site/guide/guards.md` gains a "Partial results with saveDraft" section (the anytime-result story, the per-level rule in user terms, interrupt survival, the top-level error, plain-errors-stay-failures), the `label:` parameter with an example, and the `label` field on the failure data. Wording is a draft for your edit — it's your page.

## Testing

Full guards sweep 52/52 (including the new label fixture), typechecker + runtime + codegen suites 3344 green, subprocess sample green, structural lint clean, explanation coverage gates green.

### DIFF
--- a/packages/agency-lang/docs/site/guide/guards.md
+++ b/packages/agency-lang/docs/site/guide/guards.md
@@ -29,14 +29,15 @@ Things to note:
 - `result` is a `Result` type. On success, it has the return value. On failure, it has this data:
 
 ```ts
-// "guardFailure" = cost, "timeoutFailure" = time
+// "guardFailure" = cost, "timeoutFailure" = time.
+// Every field is always present; the ones that don't apply are null.
 type GuardFailureData = {
   type: "guardFailure" | "timeoutFailure"
-  label?: string
-  maxCost?: number
-  actualCost?: number
-  maxTime?: number
-  actualTime?: number
+  label: string | null
+  maxCost: number | null
+  actualCost: number | null
+  maxTime: number | null
+  actualTime: number | null
 }
 ```
 
--- a/packages/agency-lang/docs/site/guide/guards.md
+++ b/packages/agency-lang/docs/site/guide/guards.md
@@ -32,6 +32,7 @@ Things to note:
 // "guardFailure" = cost, "timeoutFailure" = time
 type GuardFailureData = {
   type: "guardFailure" | "timeoutFailure"
+  label?: string
   maxCost?: number
   actualCost?: number
   maxTime?: number
@@ -43,6 +44,17 @@ type GuardFailureData = {
 
 - `cost:` — a `number` of dollars, eg `$2.0`.
 - `time:` — a `number` of milliseconds (or use the unit literals: `30s`, `5m`, `100ms`, `1h`).
+
+You can also name a guard with `label:`. The label comes back on the failure (`error.label`) and in error messages, so code with several guards can tell which one fired:
+
+```ts
+const result = guard(label: "research", cost: $0.50) as {
+  return research(topic)
+}
+if (isFailure(result)) {
+  print("Guard '" + result.error.label + "' tripped")
+}
+```
 
 These use [unit literals](/guide/basic-syntax.html#unit-literals).
 
@@ -71,6 +83,37 @@ When the clock ticks:
 - `sleep` = yes.
 
 When the time guard fires, any in-flight HTTP requests are cancelled and an abort signal is sent to other code as well.
+
+## Partial results with saveDraft
+
+A tripped guard normally throws the block's work away. `saveDraft` lets you keep a best-so-far value instead. Call it as your result improves; if the guard trips, the guard returns the last saved draft as a **success** instead of a failure.
+
+```ts
+node main() {
+  const result = guard(time: 5m) as {
+    let report = ""
+    for (topic in topics) {
+      report = report + research(topic)
+      saveDraft(report)
+    }
+    return report
+  }
+  // Finished in time: the full report.
+  // Tripped: the report so far.
+  print(result.value)
+}
+```
+
+`saveDraft` is always available — no import needed.
+
+Things to note:
+
+- The last saved value wins. Save early, save often.
+- The draft must match the type your function or block returns. The type checker enforces this, the same way it checks your `return` statements.
+- Each function keeps its own draft. If a function you *call* saved a draft but your code did not, your level returns nothing — a draft only crosses a call boundary when you return the callee's result directly (`return verify()`), because then its value *is* your value. This keeps every salvaged value correctly typed for the guard that receives it.
+- Drafts survive interrupts. If your block pauses for user input and trips after resuming, the draft saved before the pause still counts.
+- With no enclosing guard, `saveDraft` is a harmless no-op. At module top level it is a compile error — there is nothing a draft could ever be returned from.
+- A plain thrown error never returns a draft. Drafts are for budget trips, where the work so far is still trustworthy; unexpected failures stay failures.
 
 ## Nested guards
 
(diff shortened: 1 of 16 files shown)


packages/agency-lang/lib/runtime/guard.ts
```diff
@@ -96,8 +96,8 @@ export type Guard = {
  *  is checkpointed too — if the guard got a fresh id on resume they would no
  *  longer match and the trip would escape its guard. */
 export type GuardJSON =
-  | { kind: "cost"; costLimit: number; spent: number; guardId?: string }
-  | { kind: "time"; timeLimit: number; elapsedMs: number; guardId?: string };
+  | { kind: "cost"; costLimit: number; spent: number; guardId?: string; label?: string }
```
egonSchiele: could factor out the common parts into a new type like `GuardBase` and use intersection (`&`)

packages/agency-lang/docs/site/guide/guards.md
```diff
@@ -32,6 +32,7 @@ Things to note:
 // "guardFailure" = cost, "timeoutFailure" = time
 type GuardFailureData = {
   type: "guardFailure" | "timeoutFailure"
+  label?: string
```
egonSchiele: Good — but this makes the doc inconsistent with the actual exported type. The **source** `export type GuardFailureData` in `stdlib/thread.agency` (line ~182) was *not* given a `label` field, so:

- this guide sample declares `label?: string`,
- the runtime's `guardFailureData()` in `result.ts` now **always** sets `label: string | null`,
- the docstring on `guard()` promises `error.label`,
- and `guard-label.agency` reads `error.label` —

but the one place that actually defines the public shape omits it. It only compiles because a failure's `error` is typed `any`, so the field access isn't type-checked. Suggest adding the field to the source type and regenerating `docs/site/stdlib/thread.md`:

```
export type GuardFailureData = {
  type: string;
  label: string | null;   // <-- add
  maxCost: number | null;
  ...
}
```

Note the shape mismatch too: the runtime emits `label` on **every** guard failure (null when unlabeled), so `label: string | null` matches reality better than this sample's optional `label?: string`. Worth making the guide sample and the source type agree.

egonSchiele (commented):
## Review: Guards follow-ups — labels, AG6031, guide docs

Reviewed the three incremental commits on top of the saveDraft base (the saveDraft mechanism itself I covered in #553). Scope is small and the work is clean. Notes below — one worth fixing before merge, the rest are notes/FYI.

### Guard labels
Plumbing is consistent end to end: `label` rides the `CostGuard`/`TimeGuard` constructor → `check()` → `GuardExceededError` → the `guardTrip` cause → `guardFailureData` → the failure, and it's preserved across `toJSON`/`fromJSON` and the `TimeGuard` branch clone (keeps the parent's label alongside `guardId`). `guardFailureData` is the single construction site, so there's no second builder to drift. Unit tests cover message, cause, serialization round-trip, and unlabeled-stays-absent. Good.

**Fix before merge — the exported `GuardFailureData` type omits `label`** (inline comment on `guards.md`). The runtime now sets `label` on every guard failure, the docstring and guide advertise `error.label`, and `guard-label.agency` reads it — but `export type GuardFailureData` in `stdlib/thread.agency` (and the generated `docs/site/stdlib/thread.md`) never got the field. It only compiles because a failure's `error` is typed `any`. Add `label: string | null` to the source type and regenerate the stdlib doc. (Also reconcile optionality: the runtime emits it always, null when unlabeled — so `string | null`, not the guide sample's optional `label?: string`.)

**Note — `guard()` gained `label` between `time` and `block`.** Safe, because the block is always supplied via `guard(...) as { }` (or named), and a positional function landing in `label` position would fail the `string | null` type check. No action; flagging the signature change for the record.

### AG6031 (saveDraft at module top level)
Correct and well-placed: the check sits *after* the user-defined-`saveDraft` and imported-origin guards, so only the real stdlib `saveDraft` at `GLOBAL_SCOPE_KEY` trips it, and it mirrors the runtime throw exactly (compile-time catch of the same condition). Registry entry + explanation page are the standard shape.

### Cross-PR consistency (FYI, not a defect in these commits)
On this branch `saveDraft` lives in the auto-import prelude (`stdlib/index.agency`), so the guide's "no import needed" is accurate here. On the #553 branch (`save-draft-carry`) `saveDraft` lives in `std::thread` and the fixtures import it explicitly. These two open, related PRs disagree on where `saveDraft` lives and whether it needs importing — whichever lands first sets the contract. Worth reconciling so the guide, the fixtures, and the import story all match the version that ships.

### Docs prose
The guide reads well and is factually correct (the per-level rule, interrupt survival, top-level error, plain-errors-stay-failures all match the implementation). Left as your page to edit per the PR note.

Overall: small, correct, safe once `GuardFailureData` gains the `label` field.

packages/agency-lang/lib/runtime/guard.ts
egonSchiele: Factored in 82fa32f31: shared `guardId`/`label` live on a `GuardJSONBase` type intersected into both variants.

packages/agency-lang/stdlib/thread.agency
```diff
@@ -231,6 +232,7 @@ export def guard(
 
   @param cost - Maximum cost in dollars (e.g. $2.00 or 2.00). null or negative = no cost limit.
   @param time - Maximum compute time in milliseconds (e.g. 30s, 5m, or a raw number). null, zero, or negative = no time limit.
+  @param label - Optional name for this guard. On a trip it appears in the failure (`error.label`) and in error messages, so code with several guards can tell which one fired.
   @param block - The work to run under the guard.
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 82fa32f31 — `GuardFailureData` in stdlib/thread.agency now carries `label: string | null`, and the stdlib reference page is regenerated from it.

packages/agency-lang/docs/site/guide/guards.md
```diff
@@ -32,6 +32,7 @@ Things to note:
...
   maxCost?: number
```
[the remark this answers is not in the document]
egonSchiele: Changed in 82fa32f31 — the guide sample now shows `label: string | null` and drops the optional markers on the other fields too, so it mirrors the source type exactly (every field present, null when it does not apply).

packages/agency-lang/docs/site/guide/guards.md
egonSchiele: Fixed in 82fa32f31, all four places now agree on one shape: the source `GuardFailureData` gains `label: string | null`, the guide sample mirrors the source exactly (null-union fields, no optionals), the runtime already emitted that shape, and `docs/site/stdlib/thread.md` is regenerated. One rendering note: the generated reference page prints every null-union field as `field?: type` — that is the doc generator's existing convention (`maxCost` etc. rendered that way before this PR), not a remaining mismatch in the source. The any-typed `error` blind spot you called out is real and pre-existing for all failure data; typed failure payloads would close it but that is a bigger change than this PR.

PR #556, opened 2026-07-16
Add finalize blocks: compute a scope's partial result when a guard stops it

egonSchiele:
Adds `finalize { ... }` blocks: when an abort stops a scope, its finalize runs with the scope's locals — including the aborted callee's partial, bound into the local its call was assigning — and the finalize's return becomes the scope's forced return value (the salvage a guard receives).

Plan (external-reviewed): `docs/superpowers/plans/2026-07-16-finalize-keyword.md`. Design: the finalize section of the carry-on-abort spec, revision 3.

```
def code(): Report {
  const x = verify()      // trip happens inside verify
  return x

  finalize {
    if (x != null) { return mergePartial(x) }   // x holds verify's partial
    return outlineOnly()
  }
}
```

## How it works

- **The `__finalize` closure runs on the container's own frame** — locals live on the frame, so the body reads them with zero passing; a fresh Runner's `#finalize` scope name namespaces nothing but its own bookkeeping. Two frame-sharing subtleties the fixtures caught: finalize step ids compile in a **disjoint range** (Runner counters are frame-keyed; small ids collided with the main body's and silently skipped finalize steps), and **`fromError` now marks a guard trip's cause `delivered`** (converting the exception into the value IS the delivery; otherwise a leaf-delivered time trip left `shouldSkip` re-throwing inside the finalize's first step, killing pure computation).
- **Three stop sites call one runtime method**, `AbortedResult.withFinalize`: the frame catch, the post-call check (which first binds the callee's partial into the local via `partialValueOrNull()` — the `{value}` wrapper stays inside the class), and a checked temp at direct-call returns in finalize scopes, because plain pass-through would silently skip the finalize.
- **`withFinalize` owns the failure rule**: a finalize that throws — or resolves to interrupts or an aborted result of its own — never masks the trip; the abort continues with the saved draft, or nothing, logged as a `finalizeError`.

## Checker (six rules, each with its own code + explanation page)

AG6032 one per scope · AG6033 top level only · AG3016 no interrupts (transitive for same-file callees via `interruptEffectsByFunction`; an **imported** interrupting callee is not statically visible — documented in the explanation, pinned by a test, caught by the runtime backstop) · AG6034 no saveDraft inside · AG6035 defs and guard blocks only · AG6036 return shape: in a finalize scope, a return expression containing a call must BE a single direct call — anything else consumes an aborted callee inside the expression before the finalize can run, and no lowering fixes that without breaking short-circuit evaluation.

Two flow-graph properties: the finalize body is a **side branch off the scope START** (handled in the fold before the exit shortcut — a finalize after an unconditional return is a declaration, not dead code, and its `return` never satisfies definite-returns), and the side branch seeds **every local widened to `T | null`** (any statement might not have run; `!= null` narrows back).

## One deliberate deviation from the spec, for sign-off

v1 rejects finalize in **node** bodies (AG6035). The spec's "works in nodes" sentence predates the value transport: above a node, aborts are exceptions again and nothing consumes a node's partial until root budgets land — a node finalize would compute a value nobody reads.

## Testing

- 12 execution fixtures: the walked example verbatim, both `f(g())` directions, error-fallback vs finalize-wins-over-draft, in-guard-block, return-position interception (with an output pass-through could never produce), not-run-on-success, after-resume (a TIME trip over the restored frame), reads-globals, **composes** (inner finalize first; outer consumes inner's finalized value), race-loser-never-salvaged, inside-fork (branch partial drops at the boundary).
- Full guards sweep 64/64; subprocess sample; agency-js pin; 4,958 runtime/typechecker/codegen/parser unit tests; lint clean.
- The no-finalize path is byte-identical: `make fixtures` produces zero churn (catch branches are builder-pre-rendered strings, not mustache sections).

## Docs

`docs/dev/saveDraft.md` gains its promised finalize section; `docs/site/guide/guards.md` gains a finalize subsection under saveDraft — wording is a draft for your edit.

### DIFF
--- a/packages/agency-lang/docs/site/diagnostics/effects.md
+++ b/packages/agency-lang/docs/site/diagnostics/effects.md
@@ -156,10 +156,12 @@ A value raises an effect that the target type's `raises` clause does not allow. 
 
 <a id="ag3016"></a>
 
-## AG3016 — A finalize block cannot raise interrupts: it runs while an abort is stopping the scope, with nothing to resume back to. '&#123;callee&#125;' can interrupt.
+## AG3016 — '&#123;callee&#125;' can interrupt, and a finalize block cannot contain interrupts. A finalize runs while its scope shuts down, so there is nothing to resume.
 
 *Default severity: error.*
 
-A finalize block runs at the moment an abort stops its scope — outside the step machinery that interrupts pause and resume through. An interrupt raised there would have nothing to resume back to. This checks transitively for functions defined in your own files: calling one that can interrupt is as forbidden as a direct `interrupt(...)`. An IMPORTED function that interrupts is not visible to this static check; at runtime the finalize is treated as failed and the scope falls back to its saved draft.
+An interrupt pauses the program and waits for an answer. When the answer arrives, the program resumes from the paused step. A finalize block runs while an abort shuts its scope down, so there is no step to resume from. That is why a finalize cannot interrupt, and cannot call a function that interrupts.
 
-**How to fix:** keep the finalize computational — combine the locals you already have. Anything that needs a human belongs in the normal body, before the work that can trip.
+The check follows calls into functions defined in your own files. It cannot see into imported functions. If an imported function interrupts inside a finalize at runtime, the finalize counts as failed and the scope falls back to its saved draft.
+
+**How to fix:** only compute values inside the finalize, using the variables you already have. If you need to ask the user something, ask in the normal body, before the work that can trip.
--- a/packages/agency-lang/docs/site/diagnostics/effects.md
+++ b/packages/agency-lang/docs/site/diagnostics/effects.md
@@ -153,3 +153,13 @@ A value is being used where the target type limits which effects may be raised, 
 A value raises an effect that the target type's `raises` clause does not allow. The target restricts the effect set, and this value steps outside it.
 
 **How to fix:** add the effect to the target type's clause, or use a target type that permits it.
+
+<a id="ag3016"></a>
+
+## AG3016 — A finalize block cannot raise interrupts: it runs while an abort is stopping the scope, with nothing to resume back to. '&#123;callee&#125;' can interrupt.
+
+*Default severity: error.*
+
+A finalize block runs at the moment an abort stops its scope — outside the step machinery that interrupts pause and resume through. An interrupt raised there would have nothing to resume back to. This checks transitively for functions defined in your own files: calling one that can interrupt is as forbidden as a direct `interrupt(...)`. An IMPORTED function that interrupts is not visible to this static check; at runtime the finalize is treated as failed and the scope falls back to its saved draft.
+
+**How to fix:** keep the finalize computational — combine the locals you already have. Anything that needs a human belongs in the normal body, before the work that can trip.
--- a/packages/agency-lang/docs/site/guide/guards.md
+++ b/packages/agency-lang/docs/site/guide/guards.md
@@ -139,7 +139,7 @@ If `expand` was stopped mid-way, its own partial (whatever it saved or finalized
 
 Things to note:
 
-- A finalize runs only on a guard trip — never on success, and never for a plain thrown error.
+- A finalize never runs on success, and never for a plain thrown error. It runs when the scope is aborted, but only a guard trip salvages its value; an abort for any other reason (like losing a `race`) discards it.
 - Inside a finalize, every variable might not have been assigned yet, so each one reads as possibly-null. Check with `!= null` before using them.
 - Keep finalize bodies computational. The guard that tripped is still aborting, so an `llm()` or `sleep()` inside the finalize gets cancelled; combine the values you already have.
 - One finalize per function or block, at the top level of its body. Convention: put it last.
--- a/packages/agency-lang/docs/site/guide/guards.md
+++ b/packages/agency-lang/docs/site/guide/guards.md
@@ -116,6 +116,36 @@ Things to note:
 - With no enclosing guard, `saveDraft` is a harmless no-op. At module top level it is a compile error — there is nothing a draft could ever be returned from.
 - A plain thrown error never returns a draft. Drafts are for budget trips, where the work so far is still trustworthy; unexpected failures stay failures.
 
+### Computing the partial with finalize
+
+Sometimes the best partial result is not a value you saved along the way, but something you compute from what you have at the moment the guard trips. A `finalize` block does that: it runs only if the scope is stopped, sees the scope's variables, and its `return` becomes the salvaged value.
+
+```ts
+def research(topic: string): string {
+  const outline = draftOutline(topic)
+  const full = expand(outline)     // guard trips in here
+  return full
+
+  finalize {
+    if (outline != null) {
+      return "OUTLINE ONLY: " + outline
+    }
+    return "nothing yet"
+  }
+}
+```
+
+If `expand` was stopped mid-way, its own partial (whatever it saved or finalized) lands in `full`, and your finalize can use it.
+
+Things to note:
+
+- A finalize runs only on a guard trip — never on success, and never for a plain thrown error.
+- Inside a finalize, every variable might not have been assigned yet, so each one reads as possibly-null. Check with `!= null` before using them.
+- Keep finalize bodies computational. The guard that tripped is still aborting, so an `llm()` or `sleep()` inside the finalize gets cancelled; combine the values you already have.
+- One finalize per function or block, at the top level of its body. Convention: put it last.
+- If both a `saveDraft` and a finalize exist, the finalize wins; if the finalize itself fails, the saved draft is used instead.
+- In a function with a finalize, a `return` expression that contains a call must be just the call (`return f(x)`). Anything more complex is a compile error — assign to a local first.
+
 ## Nested guards
 
 Guards are scoped — an inner trip does **not** trip an outer guard:
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -6,6 +6,7 @@ import {
   Assignment,
   BlockType,
   Expression,
+  FinalizeBlock,
   Keyword,
   Literal,
   NamedArgument,
@@ -1766,7 +1767,13 @@ export class TypeScriptBuilder {
     const parentScopeName = this.scopes.currentName();
     this.scopes.push({ type: "block", blockName });
     this._sourceMapBuilder.enterScope(this.moduleId, blockName);
-    const bodyParts = this.processBodyAsParts(block.body);
+    const { rest, finalize } = this.extractFinalize(block.body);
+    this.finalizePresence.push(finalize !== undefined);
+    const bodyParts = this.processBodyAsParts(rest);
+    const finalizeDecl = finalize
+      ? this.buildFinalizeClosure(finalize, blockName, "__bstack")
+      : undefined;
+    this.finalizePresence.pop();
     this._sourceMapBuilder.enterScope(this.moduleId, parentScopeName);
     this.scopes.pop();
 
@@ -1781,6 +1788,14 @@ export class TypeScriptBuilder {
       scopeName: JSON.stringify(blockName),
       frameVar: `__bframe_${blockName}`,
       body: bodyStr,
+      // Pre-rendered (see the def-catch note): empty/plain strings keep
+      // the no-finalize block output byte-identical.
+      finalizeDecl:
+        finalizeDecl !== undefined ? printTs(finalizeDecl, 0) + "\n" : "",
+      abortReturn:
+        finalize !== undefined
+          ? `return await AbortedResult.fromError(__blockError, __bstack, ${JSON.stringify(blockName)}).withFinalize(__finalize, ${JSON.stringify(blockName)});`
+          : `return AbortedResult.fromError(__blockError, __bstack, ${JSON.stringify(blockName)});`,
     });
 
     const blockFn = ts.arrowFn(
@@ -1811,7 +1826,13 @@ export class TypeScriptBuilder {
     const parentScopeName = this.scopes.currentName();
     this.scopes.push({ type: "block", blockName });
     this._sourceMapBuilder.enterScope(this.moduleId, blockName);
-    const bodyParts = this.processBodyAsParts(block.body);
+    const { rest, finalize } = this.extractFinalize(block.body);
+    this.finalizePresence.push(finalize !== undefined);
+    const bodyParts = this.processBodyAsParts(rest);
+    const finalizeDecl = finalize
+      ? this.buildFinalizeClosure(finalize, blockName, "__bstack")
+      : undefined;
+    this.finalizePresence.pop();
     this._sourceMapBuilder.enterScope(this.moduleId, parentScopeName);
     this.scopes.pop();
 
@@ -1826,6 +1847,14 @@ export class TypeScriptBuilder {
       scopeName: JSON.stringify(blockName),
       frameVar: `__bframe_${blockName}`,
       body: bodyStr,
+      // Pre-rendered (see the def-catch note): empty/plain strings keep
+      // the no-finalize block output byte-identical.
+      finalizeDecl:
+        finalizeDecl !== undefined ? printTs(finalizeDecl, 0) + "\n" : "",
+      abortReturn:
+        finalize !== undefined
+          ? `return await AbortedResult.fromError(__blockError, __bstack, ${JSON.stringify(blockName)}).withFinalize(__finalize, ${JSON.stringify(blockName)});`
+          : `return AbortedResult.fromError(__blockError, __bstack, ${JSON.stringify(blockName)});`,
     });
 
     const blockFn = ts.arrowFn(
@@ -2069,6 +2098,10 @@ export class TypeScriptBuilder {
     skipHooks?: boolean;
     hoistedAliases?: TsNode[];
     inDestructiveFunction?: boolean;
+    /** The compiled `__finalize` closure declaration, when the def
+     *  declared a finalize block. Placed in the setup so the catch (and
+     *  the post-call guards inside the body) can call it. */
+    finalizeDecl?: TsNode;
   }): TsNode[] {
     const { functionName, parameters, bodyCode, skipHooks } = opts;
     const hoistedAliases = opts.hoistedAliases ?? [];
@@ -2190,6 +2223,9 @@ export class TypeScriptBuilder {
             ],
           }),
         ];
+    if (opts.finalizeDecl !== undefined) {
+      setupStmts.push(opts.finalizeDecl);
+    }
     setupStmts.push(
       ts.tryCatch(
         ts.statements([
@@ -2235,6 +2271,12 @@ export class TypeScriptBuilder {
         ts.raw(
           renderFunctionCatchFailure.default({
             functionName: JSON.stringify(functionName),
+            // Pre-rendered so the no-finalize output is byte-identical to
+            // the pre-finalize compiler (no mustache section whitespace).
+            abortReturn:
+              opts.finalizeDecl !== undefined
+                ? `return await AbortedResult.fromError(__error, __stack, ${JSON.stringify(functionName)}).withFinalize(__finalize, ${JSON.stringify(functionName)});`
+                : `return AbortedResult.fromError(__error, __stack, ${JSON.stringify(functionName)});`,
           }),
         ),
         "__error",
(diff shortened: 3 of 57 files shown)


packages/agency-lang/lib/typeChecker/diagnosticExplanations.ts
```diff
@@ -399,6 +399,41 @@ node main() {
...
+
+  finalizeSaveDraft: `Inside a finalize, the scope's partial result is being computed right now — the finalize's return value is that result. Saving a draft there is meaningless.
+
+**How to fix:** \`return\` the value from the finalize instead of calling \`saveDraft\`.`,
+
+  finalizeInNode: `Salvaged partial results are consumed by an enclosing \`guard\`, and guards cannot span nodes — above a node there is only the graph engine, which does not consume partials (root budgets may, later). A finalize declared directly in a node body would compute a value nobody reads.
+
+**How to fix:** put the finalize inside the function or \`guard\` block doing the guarded work.`,
+
+  finalizeReturnShape: `When a call inside a return expression is aborted, its aborted result is consumed by the surrounding expression (concatenated, wrapped in an array, ...) before any check can run — so the finalize would be silently skipped and garbage returned. A direct \`return f(x)\` is safe: the compiler intercepts it. Anything more complex is not interceptable without breaking evaluation order.
```
egonSchiele: Small doc gap: `containsCall` also rejects calls reached through a value-access chain — `return obj.method()` and `return arr[i]()` both trip AG6036 (correctly: F4 Step 4 only intercepts plain `functionCall` nodes, and the `arr[0]()` chain-call was the #553 soundness hole, so fail-closed is right here). But a user reads "A direct `return f(x)` is safe" and reasonably thinks `return obj.method()` is *also* direct — then gets an error that looks wrong. Consider adding one line to this explanation making the chain case explicit, e.g. "Method and index calls (`obj.method()`, `arr[i]()`) count as complex too — hoist them the same way." The concat example alone doesn't cover the shape users will most often hit.

egonSchiele (commented):
## Review: `finalize` blocks

Reviewed the runtime (`abortedResult.ts`), checker (`finalizeChecks.ts`), codegen (`typescriptBuilder.ts` + templates), and the test suite. I also reviewed the plan this implements, so I checked the risky spots that plan flagged. This is a strong, careful implementation — the earlier plan-review concerns all landed as real resolutions, not hand-waves.

### Correctness — verified sound

- **The `arr[0]()` soundness hole is closed.** `containsCall` walks value-access chains (`kind: "call" | "methodCall"`), not just `functionCall` nodes, so AG6036 catches the chain-call smuggling the #553 review found. Fail-closed is the right call — F4 Step 4 only lowers plain `functionCall` returns, so rejecting everything else keeps the direct-call lowering exhaustive.
- **`fromError` now marks a guard trip `delivered`** — I traced the risk that this could suppress the owning guard's conversion. It can't: `delivered` is only read by `Runner.shouldSkip`; `__tryCall`'s conversion (both the value and exception paths) keys on `ownedGuardIds`, never on `delivered`. So even the fork-boundary `toError()` re-throw still converts at its owning guard. The early marking only stops `shouldSkip` from re-throwing into the finalize's own steps — which is exactly the intent. Safe.
- **The disjoint `FINALIZE_STEP_BASE`** correctly dodges the frame-shared step-counter collision; advancing the dying frame's step past the base is inert because an aborted frame never resumes, and the finalize can't interrupt (AG3016) so it never checkpoints. `finalize-after-resume` (a TIME trip over the restored frame) exercises this.
- **No double-run of the finalize:** the post-call/return stop sites `runner.halt(...); return;` before the frame catch can fire, `withFinalize` swallows its own finalize errors (returns `this`), and the three sites are per-abort mutually exclusive. Good.
- **`partialValueOrNull()`** keeps the `{ value }` wrapper inside the class — the codegen field-reaching flagged on the plan is gone. Nice.

### Tests — comprehensive

The checker tests now return `{code, message}` and assert the **specific** `AG60xx`/`AG2x` per rule (the weak "≥1 error" concern from the plan review is resolved), with accept cases for guard-block placement and finalize-not-last, plus a definite-returns test and the documented imported-callee coverage-boundary test. `withFinalize` units cover the no-prior-partial success/throw cases and the saved-null distinction. The 12 fixtures cover composition (inner-first, outer consumes inner's value), fork-drop, race-loser, resume, finalize-wins-over-draft, and a return-position output pass-through could never produce. This is the coverage I'd have asked for.

### Minor notes (none blocking)

1. **AG6036 rejects `return obj.method()` / `return arr[i]()`** (inline comment). Correct and documented with a hoist workaround, but the explanation's example only shows the concat case; a user hitting the very common method-call shape reads "a direct call is safe" and is surprised. One line making the chain case explicit would close the gap.
2. **`finalizeReturnLowering` comment** says "interrupts still halt through the temp unchanged" — they don't halt, they pass through as the return value for the caller's post-call check to catch (the behavior is correct; the wording isn't). Cosmetic.
3. Confirmed a non-issue while reviewing: `return (verify())` (parenthesized) is not a false-positive AG6036 — expression parens are transparent, so it parses as a plain `functionCall` and is intercepted.

### The deliberate deviation

v1 rejecting finalize in **node** bodies (AG6035) is well-reasoned — above a node, aborts are exceptions again and nothing consumes a node's partial until root budgets land. Flagged here for your explicit sign-off, as intended.

Overall: correct, well-tested, and the encapsulation/altitude notes from the plan review were all honored. Safe to land after the AG6035 sign-off; the AG6036 doc line is a nice-to-have.

packages/agency-lang/docs/site/diagnostics/effects.md
```diff
@@ -153,3 +153,13 @@ A value is being used where the target type limits which effects may be raised,
...
 
 **How to fix:** add the effect to the target type's clause, or use a target type that permits it.
+
+<a id="ag3016"></a>
+
+## AG3016 — A finalize block cannot raise interrupts: it runs while an abort is stopping the scope, with nothing to resume back to. '&#123;callee&#125;' can interrupt.
+
+*Default severity: error.*
+
+A finalize block runs at the moment an abort stops its scope — outside the step machinery that interrupts pause and resume through. An interrupt raised there would have nothing to resume back to. This checks transitively for functions defined in your own files: calling one that can interrupt is as forbidden as a direct `interrupt(...)`. An IMPORTED function that interrupts is not visible to this static check; at runtime the finalize is treated as failed and the scope falls back to its saved draft.
```
egonSchiele: This is really poorly written, hard to read language. Please make sure that the diagnostics you add are easy to read, jargon free, and avoid garden path sentences. Use the following writing tips:

## General writing tips

1. Avoid garden path sentences.

Bad: a sentence that begins like this:

```
The name lists partition
```

When you read the sentence, what do you think it means? I think it means that a name lists some partitions. Here is the full sentence:

```
The name lists partition the catalog's managed (`source:"remote"`) entries by what changed.
```

It turns out that there is something called a "name list", which partitions something in the catalog. Avoid sentences like this, because they make the reader realize halfway through reading the sentence that it doesn't that the words don't mean what the reader thought they meant, and they have to re-read the sentence to take into account the information they gained in the second half of the sentence. A classic garden path sentence begins with the words "the old man". What do you think it's going to tell us about the old man? Actually, the full sentence is "the old man the boat". Notice how the first half of the sentence makes you think it's talking about an old man, and it's not until you read the second half that you realize that it's talking about elderly people as a group.

Good: 

```
The catalog's entries are partitioned by name lists.
```

2. Use active voice, avoid passive voice.

Bad: "The man was bitten by the dog."
Good: "The dog bit the man."

Passive voice makes you focus on the person or object receiving the action, rather than the person or object performing it. Passive voice is typically harder to read and less interesting.

Other examples:

Bad: "Enclosing guard(cost:) budgets meter the subprocess in real time"
Good "The subprocess is metered in real time by enclosing guard(cost:) budgets."

Bad: "filename cannot escape dir via absolute paths or .. segments (symlinks are followed and re-checked)."
Good: "Absolute paths and .. segments cannot escape dir. Symlinks are followed and re-checked."

3. Avoid introducing too many concepts in a single sentence.

If you find yourself using multiple commas, semicolons, or conjunctions in a sentence, consider breaking it up into multiple sentences.

Bad:
```
The name lists partition the catalog's managed (`source:"remote"`) entries by what changed.
```

Notice how the sentence tries to squeeze in the `source:"remote"` qualifier to an already complicated sentence. If you need to parenthesize something in your sentence, consider whether it would be clearer as two sentences instead.

```
Parse command-line flags with strict number coercion, required flags and defaults, mutually-exclusive groups, and auto-generated `--help` / `--version`.
```

This sentence introduces so many things that by the time the reader has reached the end of the sentence, they have already forgotten the beginning of the sentence.

```
any interrupt no handler resolves surfaces to the user; responding resumes the subprocess exactly where it paused.
```

Here is an example where two sentences would have been clearer than one with a semicolon.

4. Emdash overuse

An emdash used sparingly can make the text more lively, but used all the time, it becomes jarring. If you are using an emdash, consider using a comma instead.

Bad: "The AST shape is the parser output, which matches what the formatter consumes — so an AST round-tripped through writeAST() produces canonical Agency source."
Good: "The AST shape is the parser output, which matches what the formatter consumes, so an AST round-tripped through writeAST() produces canonical Agency source."

Or use a conjunction instead of an emdash:
Bad: "Read and write happen inside the same interrupt — approving it approves both."
Good: "Read and write happen inside the same interrupt, so approving it approves both."

5. Consider whether this level of detail is actually required in the sentence.

You have a habit of trying to add more information in parentheses. Often, it's not needed, and actually hinders understanding by making the sentence longer.

Bad: "Exceeding a resource limit (wallClock, memory, ipcPayload, or stdout) kills the subprocess and returns a limit_exceeded failure."
Good: "Exceeding a resource limit kills the subprocess and returns a limit_exceeded failure."

Every time you add something in parentheses, consider that the user's brain is going to need to pause and add a new frame to their mental "stack" to capture this new information. Is that speed bump worth the information you're trying to convey?

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -1781,6 +1788,14 @@ export class TypeScriptBuilder {
       scopeName: JSON.stringify(blockName),
       frameVar: `__bframe_${blockName}`,
       body: bodyStr,
+      // Pre-rendered (see the def-catch note): empty/plain strings keep
+      // the no-finalize block output byte-identical.
+      finalizeDecl:
+        finalizeDecl !== undefined ? printTs(finalizeDecl, 0) + "\n" : "",
+      abortReturn:
+        finalize !== undefined
+          ? `return await AbortedResult.fromError(__blockError, __bstack, ${JSON.stringify(blockName)}).withFinalize(__finalize, ${JSON.stringify(blockName)});`
```
egonSchiele: This ternary is really hard to read. I would suggest pulling this long string out into a variable and using TS IR builders if possible.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2235,6 +2271,12 @@ export class TypeScriptBuilder {
         ts.raw(
           renderFunctionCatchFailure.default({
             functionName: JSON.stringify(functionName),
+            // Pre-rendered so the no-finalize output is byte-identical to
+            // the pre-finalize compiler (no mustache section whitespace).
+            abortReturn:
+              opts.finalizeDecl !== undefined
```
egonSchiele: Code duplication: I'm seeing the same pattern over and over.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2266,6 +2308,65 @@ export class TypeScriptBuilder {
...
+  private static readonly FINALIZE_STEP_BASE = 1000000;
+
+  private currentScopeHasFinalize(): boolean {
+    return this.finalizePresence[this.finalizePresence.length - 1] === true;
+  }
+
+  /** Split a body into its statements and its (at most one — the checker
+   *  enforces that) top-level finalize block. Defensive: extras beyond the
+   *  first are dropped here but already rejected by AG6032. */
+  private extractFinalize(body: AgencyNode[]): {
```
egonSchiele: Why not move all the finalize logic into its own file under packages/agency-lang/lib/backends/typescriptBuilder/? Plenty of TypeScript builder logic is in their own files, to avoid bloating this file, which is already too long. That would also allow you to write a nice declarative interface that you could use through the TypeScript builder instead of dumping all this imperative code everywhere.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -3326,6 +3499,23 @@ export class TypeScriptBuilder {
     ) {
       const frameVar = scopeType === "block" ? "__bstack" : "__stack";
       const scopeName = JSON.stringify(this.scopes.currentName());
+      if (this.currentScopeHasFinalize()) {
+        const bind =
+          opts?.bindOnAborted === true
+            ? [ts.raw(`${expr} = __abortedCallee.partialValueOrNull()`)]
+            : [];
+        return ts.if(
```
egonSchiele, line 3412: This code also looks duplicated, and it's hard to tell from a glance what it's doing. Again, you could extract all the finalize logic into its own file, and put code like this behind a named method, so it's clear what it's doing.

packages/agency-lang/tests/agency/guards/finalize-error-falls-back.agency
```diff
@@ -0,0 +1,23 @@
...
+  return reply
+
+  finalize {
+    const broken: any = null
+    return broken.field
+  }
+}
+
+node main() {
+  const result = guard(cost: 0.000001) as {
```
egonSchiele, line 18: I wonder if all of the tests for this partial value feature, including all the tests for save draft and finalize, will be really flaky. When we introduce "sigint" and "sigkill", we should use those functions to pause execution instead of using guards, unless we are specifically testing the guards' behavior.

packages/agency-lang/lib/typeChecker/finalizeChecks.ts
```diff
@@ -0,0 +1,174 @@
...
+  "forLoop",
+  "whileLoop",
+  "matchBlock",
+  "handleBlock",
+  "parallelBlock",
+  "seqBlock",
+  "withModifier",
+  "staticStatement",
+  "finalizeBlock",
+];
```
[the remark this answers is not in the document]
egonSchiele: Fixed in cc58f4aba. The hand-kept list is gone: the check now derives "wraps statements" from bodySlots, the single source of truth for body-bearing nodes. Any such ancestor between the finalize and its scope/block boundary counts as nesting, so messageThread — and any future body-bearing node type — rejects by default. Added a thread-block test.

packages/agency-lang/lib/typeChecker/finalizeChecks.ts
```diff
@@ -0,0 +1,174 @@
...
+    for (const { node, ancestors, scopes: walkScopes } of walkNodes(info.body)) {
+      if (!isInScope(walkScopes, info)) continue;
+      if (node.type === "finalizeBlock") {
+        const nesting = controlFlowAncestor(ancestors);
+        if (nesting !== undefined) {
+          ctx.errors.push(
+            diagnostic("finalizeNotTopLevel", { construct: nesting }, node.loc ?? null),
+          );
+          continue;
+        }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in cc58f4aba. The message now names the construct with its surface keyword ("an `if` statement", "a `thread` block"); a body-bearing type with no mapping falls back to "a nested block", never the AST type name. A test pins that `ifElse` does not appear.

packages/agency-lang/docs/site/guide/guards.md
```diff
@@ -116,6 +116,36 @@ Things to note:
 - With no enclosing guard, `saveDraft` is a harmless no-op. At module top level it is a compile error — there is nothing a draft could ever be returned from.
 - A plain thrown error never returns a draft. Drafts are for budget trips, where the work so far is still trustworthy; unexpected failures stay failures.
 
+### Computing the partial with finalize
+
+Sometimes the best partial result is not a value you saved along the way, but something you compute from what you have at the moment the guard trips. A `finalize` block does that: it runs only if the scope is stopped, sees the scope's variables, and its `return` becomes the salvaged value.
+
+```ts
```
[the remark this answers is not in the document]
egonSchiele: Keeping the `ts` fence. No `agency` language is registered with the docs site (the VitePress config has no shiki langAlias), so an `agency` fence renders as plain text with a build warning. The guide uses `ts` for approximate highlighting throughout: 10 of 10 blocks in this file, 26 in basic-syntax.md. Registering an agency grammar and migrating fences repo-wide would be a separate change.

packages/agency-lang/docs/site/guide/guards.md
```diff
@@ -116,6 +116,36 @@ Things to note:
...
+    return "nothing yet"
+  }
+}
+```
+
+If `expand` was stopped mid-way, its own partial (whatever it saved or finalized) lands in `full`, and your finalize can use it.
+
+Things to note:
+
+- A finalize runs only on a guard trip — never on success, and never for a plain thrown error.
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 98223c8cd. The bullet now separates the two facts: a finalize never runs on success or for a plain thrown error; it runs when the scope is aborted, but only a guard trip salvages its value, and an abort for any other reason (like losing a `race`) discards it.

packages/agency-lang/lib/typeChecker/diagnosticExplanations.ts
egonSchiele: Added in 98223c8cd. The explanation now says method and index calls do not count as direct, names `return obj.method()` and `return arr[i]()` explicitly, and points them at the same hoist fix.

packages/agency-lang/docs/site/diagnostics/effects.md
egonSchiele: Rewrote all six finalize messages and explanation pages in 98223c8cd using these tips: short sentences, active voice, no "step machinery" jargon. AG3016 now opens by saying what an interrupt does (pauses, waits, resumes from the paused step), then why a finalize has no step to resume from. Docs regenerated from the table.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Gone in 919d0232b. The string is built in one place, FinalizeCodegen.abortReturn(), and this site just passes compiledFinalize.abortReturn to the template. It stays a pre-rendered string rather than TS IR because it renders into the mustache catch template, which takes text — that is also what keeps the no-finalize output byte-identical.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Deduped in 919d0232b. The three abort-return sites and the two near-identical block-compile blobs collapse into one FinalizeCodegen.compileScope call per scope.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Done in 919d0232b: lib/backends/typescriptBuilder/finalizeCodegen.ts, following the DestructiveTracking extraction pattern. The builder now touches finalize through four named methods: compileScope (splits the body, compiles the __finalize closure, returns the setup declaration and the catch return), isActive, stopScope (the halt-through-finalize statements), and interceptedReturn (the checked-temp return lowering). The presence stack, the step base, and all string emission are private to the class. make fixtures shows zero churn and all 64 guards fixtures pass, so the output is byte-identical.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Same extraction: this guard now emits its bind plus this.finalize.stopScope("__abortedCallee"), and the return lowering delegates to this.finalize.interceptedReturn. The halt-through-finalize statement is emitted in exactly one place (stopScope), with a doc comment saying what it does.

packages/agency-lang/tests/agency/guards/finalize-error-falls-back.agency
egonSchiele: Current determinism: 23 of the 29 draft/finalize fixtures trip cost guards, which are pure accounting over mocked LLM costs and cannot flake. The other 6 use time guards with wide margins (sleep(2000) against a 600ms limit), so they are stable in practice but still wall-clock based. Agreed on the direction: once sigint/sigkill exist, they are the right way to stop execution at an exact statement, and only the guard-behavior tests should keep tripping real guards. I can file an issue to track migrating the 6 time-based fixtures when those functions land.

PR #557, opened 2026-07-16
Add message debug labels for llm() and the message builders

egonSchiele:
Lets you attach a debug label to `llm()`, `userMessage()`, `assistantMessage()`, and `systemMessage()`, and surfaces those labels in statelog — so a log reader can tell a verifier's injected message from a real user turn.

Plan (external-reviewed): `docs/superpowers/plans/2026-07-16-message-debug-labels.md`.

```agency
systemMessage("sys")
userMessage("context", label: "seed")
return llm("go", label: "coder")
```

```
promptStart      label: "coder"
promptCompletion messages: [ {sys}, {context, label:"seed"}, {go, label:"coder"} ]
```

## Labels never reach the provider

`runPrompt` strips `label` off `clientConfig` before the config is forwarded to smoltalk — the same destructure that already strips the retry fields. Nothing else about the request path changes. A test pins it, and it is not vacuous: `restClientConfig` is spread into the provider config, so deleting the strip turns that test red (verified by mutating it).

## The storage invariant

`messageLabels` lives on `MessageThread`, aligned with `messages` **by index**. An index-aligned array is what lets labels serialize with the thread (a `WeakMap` keyed by message identity would survive slicing but not `toJSON`). The cost is that nothing in the type enforces the alignment — so it is enforced by keeping the writers to a minimum:

- **`push(message, label?)` is the only append.** `addMessage()` delegates to it.
- **The constructor and `setMessages()` are the only replacements**, and both rebuild `messageLabels` to the new length.
- **Nothing else touches `this.messages`.**

This matters because a desync does not degrade gracefully — it shifts every later label onto the wrong message. Each writer has a test that fails on one.

Review of the plan caught two sites that would have silently mislabeled, both fixed by the above:
- `addMessage` was a second append path (public API, used by `lib/stdlib/threads.test.ts`) that would have appended a message with no label.
- A `messageLabels = []` field default would have desynced against a constructor-provided `messages` — reachable via `newSubthreadChild()`, and async `llm()` runs in a subthread, so a later push would have attributed its label to message 0.

Also: `fromJSON` restores labels **after** its internal `setMessages()` call, which would otherwise reset them.

## Deliberate behavior

- **One `llm(label:)` call tags more than one message** — its prompt plus the assistant message of every tool-loop round. Intended: the label marks "these messages came from this call". `promptCompletion` logs the *request*, so a round's assistant reply appears in the next round's dump, not its own.
- **`setMessages` drops labels**, so thread rewrites (summarization, `threadRepair`) lose them. Intended; documented.
- **`newSubthreadChild()` does not carry the parent's labels** (it seeds via `cloneMessages()`), and starts correctly aligned with all-null.
- Unlabeled messages are logged with **no `label` key at all**, so logs for programs that never label stay byte-identical.

## Testing

- 7 `MessageThread` unit tests (round-trip, legacy revive, and one per writer that fails on a desync).
- 6 runtime tests: the provider strip, prompt+completion labeling, the `promptCompletion` merge (asserting *pairing*, not presence), and the byte-identical unlabeled dump.
- `tests/agency-js/message-labels` e2e reads the statelog and asserts label-to-message pairing with an **unlabeled system message at index 0**, so an off-by-one shows up as a label landing on the wrong message. It also asserts no label reached the provider.
- Full unit suite 8323 passing; `lint:structure` clean; `make fixtures` zero churn (no codegen changes).

## Out of scope

- Auto-labeling runtime-injected messages (guard-approve feedback, tool-reply attachments) — the resumable-guards work picks this up and needs only `push(message, label)`.
- Filtering/slicing threads by label.

### DIFF
--- a/packages/agency-lang/lib/runtime/prompt.ts
+++ b/packages/agency-lang/lib/runtime/prompt.ts
@@ -559,9 +559,7 @@ async function _runPrompt({
    * scoping the LLM HTTP abort signal to the current branch. */
   stateStack?: StateStack;
   retryPolicy: RetryPolicy;
-  /** This call's `llm(label:)` debug tag, or null. Stamped on the
-   *  assistant message this call appends. Observability only — the
-   *  provider never sees it (stripped from clientConfig in runPrompt). */
+  /** This call's `llm(label:)` debug tag (see `emitPromptStart`), or null. */
   callLabel?: string | null;
 }): Promise<RunPromptResult> {
   if (ctx.isCancelled(stateStack)) {
@@ -610,14 +608,7 @@ async function _runPrompt({
     metadata: clientConfig,
   } as any;
 
-  emitPromptStart({
-    ctx,
-    messages,
-    tools,
-    responseFormat,
-    clientConfig,
-    callLabel,
-  });
+  emitPromptStart({ ctx, messages, tools, responseFormat, clientConfig, callLabel });
 
   let completion: PromptResult;
   let toolCalls: ToolCallJSON[];
--- a/packages/agency-lang/lib/runtime/prompt.ts
+++ b/packages/agency-lang/lib/runtime/prompt.ts
@@ -34,7 +34,10 @@ import { failure, isFailure, isSuccess, markDestructiveWork } from "./result.js"
 import type { SourceLocationOpts } from "./state/checkpointStore.js";
 import type { RuntimeContext } from "./state/context.js";
 import type { LlmDefaults } from "../stdlib/llm.js";
-import { MessageThread } from "./state/messageThread.js";
+import {
+  MessageThread,
+  type MessageThreadJSON,
+} from "./state/messageThread.js";
 import { StateStack } from "./state/stateStack.js";
 import { ThreadStore } from "./state/threadStore.js";
 import { handleStreamingResponse } from "./streaming.js";
@@ -993,21 +996,32 @@ export async function runPrompt(args: {
     messages = new MessageThread();
   }
 
+  /** The ONE way this function snapshots the thread for checkpoint/resume.
+   *
+   *  Every bailout path — the PromptRunner callback, the tool loop, the
+   *  reply-attachment injection, the validation retries — stores what this
+   *  returns into `self.messagesJSON`, so the shape decision is made once
+   *  rather than at each site.
+   *
+   *  That shape is the FULL `MessageThreadJSON`, not just `.messages`: a
+   *  bare array revives through `fromJSON`'s legacy branch, which has no
+   *  labels to read, so snapshotting only the messages loses every debug
+   *  label across interrupt/resume. Older checkpoints still hold a bare
+   *  array and keep reviving through that legacy branch, so this stays
+   *  back-compatible.
+   *
+   *  Reads the `messages` binding at call time, so reassignments below
+   *  (e.g. `messages = result.messages`) are observed. */
+  const snapshotThread = (): MessageThreadJSON => messages.toJSON();
+
   // Resumable-step + checkpoint-on-interrupt helper. See
   // docs/superpowers/plans/2026-05-22-prompt-runner.md.
-  // `snapshotMessages` reads the current `messages` binding at call time;
-  // reassignments below (e.g. `messages = result.messages`) are observed.
   const pr = new PromptRunner({
     self,
     ctx,
     stateStack,
     checkpointInfo,
-    // The FULL MessageThreadJSON, not just `.messages`: the bare array
-    // revives through fromJSON's legacy branch, which has no labels to
-    // read, so a snapshot of only the messages loses every label across
-    // interrupt/resume. Old checkpoints still hold a bare array and keep
-    // reviving through that same legacy branch, so this is back-compatible.
-    snapshotMessages: () => messages.toJSON(),
+    snapshotMessages: snapshotThread,
   });
 
   // One llmCall span covers the WHOLE `llm()` call — every tool-loop
@@ -1090,7 +1104,7 @@ export async function runPrompt(args: {
           }
         }
       }
-      self.messagesJSON = messages.toJSON();
+      self.messagesJSON = snapshotThread();
       self.pendingToolCalls = toolCalls.length > 0 ? toolCalls : null;
     });
 
@@ -1644,7 +1658,7 @@ export async function runPrompt(args: {
               ),
             );
             self.runnerState.replyAttachments = [];
-            self.messagesJSON = messages.toJSON();
+            self.messagesJSON = snapshotThread();
           });
         }
 
@@ -1670,7 +1684,7 @@ export async function runPrompt(args: {
           // Increment the round counter only after a successful LLM round,
           // so resume after a tool-batch interrupt re-enters the SAME round.
           self.toolCallRound = round + 1;
-          self.messagesJSON = messages.toJSON();
+          self.messagesJSON = snapshotThread();
           self.pendingToolCalls = toolCalls.length > 0 ? toolCalls : null;
         });
       }
@@ -1737,7 +1751,7 @@ export async function runPrompt(args: {
           },
         });
         messages.push(smoltalk.userMessage(decision.feedback));
-        self.messagesJSON = messages.toJSON();
+        self.messagesJSON = snapshotThread();
       });
       await pr.step(`validation.${validationAttempt}.llmCall`, async () => {
         const nextResult = await _runPrompt({
@@ -1757,7 +1771,7 @@ export async function runPrompt(args: {
         // bailout before this step completes leaves the counter unchanged,
         // so resume re-enters the SAME attempt with the same step keys.
         self.validationAttempt = validationAttempt + 1;
-        self.messagesJSON = messages.toJSON();
+        self.messagesJSON = snapshotThread();
         self.pendingToolCalls = toolCalls.length > 0 ? toolCalls : null;
       });
       // Loop: the retry response may itself contain tool calls; the inner
--- a/packages/agency-lang/lib/runtime/prompt.ts
+++ b/packages/agency-lang/lib/runtime/prompt.ts
@@ -79,6 +79,22 @@ export function redactMessagesForLog(
   return redactAttachments(messages.toJSON().messages) as smoltalk.MessageJSON[];
 }
 
+/** A thread's redacted messages for statelog, each carrying its debug
+ *  label (from `llm(label:)` / `userMessage(msg, label:)`) when it has
+ *  one. Reads the labels off the SAME thread that produced the dump, so
+ *  index `i` lines up: `redactMessagesForLog` maps `messages` in order.
+ *
+ *  Unlabeled messages are emitted unchanged rather than with `label:
+ *  null`, so logs for label-free programs stay byte-identical. */
+export function withMessageLabels(
+  messages: MessageThread,
+): smoltalk.MessageJSON[] {
+  return redactMessagesForLog(messages).map((m, i) => {
+    const label = messages.labelAt(i);
+    return label === null ? m : { ...m, label };
+  });
+}
+
 /** A prompt with attachment payloads redacted, preserving its string-or-array
  *  shape, for statelog / hook data. */
 export function redactPromptForLog(
@@ -672,7 +688,7 @@ async function _runPrompt({
   const modelName = completion.model || clientConfig.model || "unknown model";
 
   ctx.statelogClient.promptCompletion({
-    messages: redactMessagesForLog(messages),
+    messages: withMessageLabels(messages),
     completion,
     model: JSON.stringify(modelName),
     timeTaken: endTime - startTime,
--- a/packages/agency-lang/lib/runtime/prompt.ts
+++ b/packages/agency-lang/lib/runtime/prompt.ts
@@ -500,12 +500,15 @@ function emitPromptStart({
   tools,
   responseFormat,
   clientConfig,
+  callLabel,
 }: {
   ctx: RuntimeContext<GraphState>;
   messages: MessageThread;
   tools: Tool[];
   responseFormat?: any;
   clientConfig: Partial<smoltalk.SmolConfig>;
+  /** This call's `llm(label:)` debug tag, or null. Observability only. */
+  callLabel?: string | null;
 }): void {
   ctx.statelogClient.promptStart({
     model: JSON.stringify(clientConfig.model),
@@ -514,6 +517,7 @@ function emitPromptStart({
     toolCount: tools.length,
     hasResponseFormat: responseFormat != null,
     maxTokens: clientConfig.maxTokens ?? null,
+    label: callLabel ?? null,
   });
 }
 
@@ -526,6 +530,7 @@ async function _runPrompt({
   clientConfig,
   stateStack,
   retryPolicy,
+  callLabel,
 }: {
   ctx: RuntimeContext<GraphState>;
   messages: MessageThread;
@@ -538,6 +543,10 @@ async function _runPrompt({
    * scoping the LLM HTTP abort signal to the current branch. */
   stateStack?: StateStack;
   retryPolicy: RetryPolicy;
+  /** This call's `llm(label:)` debug tag, or null. Stamped on the
+   *  assistant message this call appends. Observability only — the
+   *  provider never sees it (stripped from clientConfig in runPrompt). */
+  callLabel?: string | null;
 }): Promise<RunPromptResult> {
   if (ctx.isCancelled(stateStack)) {
     throw new AgencyCancelledError();
@@ -585,7 +594,14 @@ async function _runPrompt({
     metadata: clientConfig,
   } as any;
 
-  emitPromptStart({ ctx, messages, tools, responseFormat, clientConfig });
+  emitPromptStart({
+    ctx,
+    messages,
+    tools,
+    responseFormat,
+    clientConfig,
+    callLabel,
+  });
 
   let completion: PromptResult;
   let toolCalls: ToolCallJSON[];
@@ -674,9 +690,10 @@ async function _runPrompt({
       smoltalk.assistantMessage(completion.output, {
         toolCalls,
       }),
+      callLabel,
     );
   } else {
-    messages.push(smoltalk.assistantMessage(completion.output));
+    messages.push(smoltalk.assistantMessage(completion.output), callLabel);
   }
 
   updateTokenStats({
@@ -853,6 +870,8 @@ export async function runPrompt(args: {
   // smoltalk doesn't understand. `retries` / `timeout` / `backoff` are the
   // resilience policy (resolved below); if the codegen forwarded them on
   // `clientConfig` they'd otherwise hit the LLM client as foreign options.
+  // `label` is an observability-only debug tag — stripping it here is what
+  // keeps it off the provider wire.
   const {
     tools: _extractedTools,
     memory: memoryOption,
@@ -861,13 +880,19 @@ export async function runPrompt(args: {
     timeout: ccTimeout,
     backoff: ccBackoff,
     validationRetries: ccValidationRetries,
+    label: ccLabel,
     ...restClientConfig
   } = (args.clientConfig || {}) as Partial<smoltalk.SmolConfig> &
     RetryConfig & {
       tools?: any[];
       memory?: boolean | { model?: string };
       maxToolResultChars?: number;
+      label?: string;
     };
+
+  /** This llm() call's debug label, or null when unlabeled. Stamps the
+   *  call's promptStart event and the messages it appends. */
+  const callLabel: string | null = ccLabel || null;
 
   // Resolve the resilience policy. Precedence:
   //   1. `args.retryConfig` (preferred — what `agency.llm` passes directly)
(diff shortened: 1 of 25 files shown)


packages/agency-lang/lib/runtime/state/messageThread.ts
```diff
@@ -39,14 +40,31 @@ export class MessageThread {
...
+   *  Observability only: shown in statelog, never sent to the provider.
+   *  Distinct from the thread-level `label` above (set by `thread(label:)`).
+   *
+   *  The alignment is maintained by keeping the writers to a minimum:
+   *  `push` is the only append, the constructor and `setMessages` are the
+   *  only replacements, and nothing else touches `this.messages`. A desync
+   *  does not degrade gracefully — it shifts every later label onto the
+   *  wrong message — so keep it that way. A rewrite via `setMessages`
+   *  (summarization, repair) drops labels; that is intended. */
+  messageLabels: (string | null)[];
```
egonSchiele, line 61: Does every message have a message ID? If so, would it make sense to make this a hash keyed by message IDs instead? that would be a stronger guarantee that every label always gets lined up with the correct message

packages/agency-lang/lib/runtime/state/messageThread.ts
```diff
@@ -61,10 +79,19 @@ export class MessageThread {
 
   setMessages(messages: smoltalk.Message[]): void {
```
egonSchiele: where does this get used? Would it make sense to also pass in an array of labels as a second argument and then zip the two arrays together?

packages/agency-lang/lib/runtime/state/messageThread.ts
```diff
@@ -61,10 +79,19 @@ export class MessageThread {
...
     this.messages = messages;
+    this.messageLabels = messages.map(() => null);
```
egonSchiele: This reset is correct for what `setMessages` is documented to mean (a wholesale rewrite drops labels), but two of its four callers aren't rewrites, and they lose labels silently:

- `prompt.ts:1075` — removes one injected-facts system message and resets every label as collateral. Runs on the **normal** path when the memory layer is on, so a labeled program loses its labels after the first `llm()` call.
- `prompt.ts:977` — `args.messages.setMessages(restored.getMessages())` on the resume path, which throws away exactly the labels `fromJSON` just restored below.

Suggest giving the class the two operations those callers actually want, so the labels come along by construction rather than by each caller remembering — the same argument this PR already makes for `addMessage` delegating to `push`:

```ts
/** Remove the message at `index`, keeping labels aligned. */
removeAt(index: number): void {
  this.messages.splice(index, 1);
  this.messageLabels.splice(index, 1);
}

/** Take on `other`'s messages and labels while keeping this thread's
 *  identity — the resume path needs the alias, not a fresh object. */
adoptFrom(other: MessageThread): void {
  this.messages = [...other.messages];
  this.messageLabels = [...other.messageLabels];
}
```

`setMessages` then keeps its documented meaning for the caller that really is a rewrite (summarize / `threadRepair`).

packages/agency-lang/lib/runtime/state/messageThread.ts
```diff
@@ -138,6 +170,11 @@ export class MessageThread {
     );
 
     thread.setMessages(smoltalkMessages);
+    // AFTER setMessages: it resets messageLabels to all-null, so the
+    // restore has to come second or it gets clobbered. Legacy JSON (no
+    // messageLabels) keeps the all-null array setMessages just built.
+    thread.messageLabels =
+      _messageLabels ?? smoltalkMessages.map(() => null);
```
egonSchiele: This ordering fix is right, and the test for it passes — but it never fires with real labels in the runtime.

`runPrompt` checkpoints with `self.messagesJSON = messages.toJSON().messages` (prompt.ts:1080, :1634, :1660, :1727, :1747), which keeps only the `messages` field and drops `messageLabels`. The resume then calls `MessageThread.fromJSON(self.messagesJSON)` with a bare array, so it takes the legacy `Array.isArray(json)` branch and `_messageLabels` is always `undefined`.

So the labels are gone before this line ever gets a chance, and `prompt.ts:977` would wipe them again anyway. To carry them, `self.messagesJSON` has to hold the full `MessageThreadJSON` (or a parallel labels field alongside it).

packages/agency-lang/lib/runtime/state/messageThread.ts
```diff
@@ -86,6 +113,7 @@ export class MessageThread {
   toJSON(): MessageThreadJSON {
     return {
       messages: this.messages.map((m) => m.toJSON()),
+      messageLabels: this.messageLabels,
```
egonSchiele: Nit: this hands out the **live** array, while `messages` above is rebuilt (`.map(m => m.toJSON())`). A consumer that mutates the returned JSON mutates the thread — which is the one way to touch `messageLabels` without going through the three sanctioned writers. `[...this.messageLabels]` for consistency.

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -79,6 +79,22 @@ export function redactMessagesForLog(
...
 }
 
+/** A thread's redacted messages for statelog, each carrying its debug
+ *  label (from `llm(label:)` / `userMessage(msg, label:)`) when it has
+ *  one. Reads the labels off the SAME thread that produced the dump, so
+ *  index `i` lines up: `redactMessagesForLog` maps `messages` in order.
+ *
+ *  Unlabeled messages are emitted unchanged rather than with `label:
+ *  null`, so logs for label-free programs stay byte-identical. */
+export function withMessageLabels(
```
egonSchiele, line 95: Two things here.

**The declared return type isn't true.** The labeled entries carry a `label` key that `smoltalk.MessageJSON` doesn't declare, so consumers have to cast to read it (the e2e does, in JS, where nothing checks). Suggest `(smoltalk.MessageJSON & { label?: string })[]`.

**The index join leans on an external library.** The alignment holds only because smoltalk's `redactAttachments` is `value.map(v => ...)` — 1:1 (I checked `node_modules/smoltalk/dist/util/redact.js`). It's a reasonable bet, but it's a smoltalk implementation detail, and if a future version ever filtered or split a message every label would shift onto the wrong one with nothing failing: both `withMessageLabels` tests use plain text messages, so the redaction path is never exercised with an actual attachment. One test with an `image()` attachment plus labels, asserting pairing, would pin the assumption where it actually lives.

packages/agency-lang/lib/runtime/promptLabels.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    }),
+  );
+  return { client, thread };
+}
+
+describe("llm() debug label", () => {
+  it("never reaches the provider config", async () => {
+    const { client } = await runLabeled("verifier");
+    expect(client.configs).toHaveLength(1);
+    expect("label" in (client.configs[0] as any)).toBe(false);
```
egonSchiele: This checks the top-level key only. `_runPrompt` also forwards the whole config nested under `metadata: clientConfig` (prompt.ts:~608), so a leak could ride in at `configs[0].metadata.label` and this would still be green.

It's safe today — both `...clientConfig` and `metadata:` read the same post-strip object, so the one destructure covers both — but the assertion is narrower than the property it's protecting. Cheap to widen:

```ts
expect("label" in (client.configs[0] as any)).toBe(false);
expect("label" in ((client.configs[0] as any).metadata ?? {})).toBe(false);
```

egonSchiele (commented):
## Review: message debug labels

The storage design is right and the invariant is genuinely enforced structurally rather than by vigilance — one append, two replacements, `addMessage` delegating. The provider strip is real (I traced it: `restClientConfig` → `getSmoltalkConfig` → `promptConfig`'s spread **and** its `metadata:` field, so both wire paths are covered by the one destructure). Docs are good.

**But labels do not survive an interrupt inside `llm()`, and three call sites silently wipe them. None of the tests catch it, because every test exercises `MessageThread` directly rather than through the paths that actually use it.**

### 1. `self.messagesJSON` never stores the labels (save side)

`self.messagesJSON = messages.toJSON().messages` — at prompt.ts:1080, :1634, :1660, :1727, :1747 — takes **only** the `messages` field off the JSON and drops `messageLabels`. On resume, `MessageThread.fromJSON(self.messagesJSON)` receives a bare `MessageJSON[]`, hits the legacy array branch (`if (Array.isArray(json))`), and revives all-null.

So `fromJSON`'s careful "restore labels *after* the internal `setMessages`" ordering — called out in the PR body as a fix — never runs with real labels in the actual checkpoint path.

### 2. The restore then wipes them again (restore side)

```ts
// prompt.ts:977
args.messages.setMessages(restored.getMessages());
```

`setMessages` rebuilds `messageLabels` as all-null by design. Even if #1 were fixed, this line would throw the restored labels away. Independent second cause, same symptom.

Net: **any interrupt inside `llm()` — a tool raising one, `input()`, an approval — loses every label on that thread for the rest of the run.** That is not an exotic path. `ThreadStore.toJSON` does preserve labels (it calls `thread.toJSON()` in full), so the loss is specific to `runPrompt`'s own resumable-step state.

### 3. The injected-facts removal wipes all labels as collateral (no interrupt needed)

```ts
// prompt.ts:1075
messages.setMessages([...all.slice(0, i), ...all.slice(i + 1)]);
```

This removes one system message and resets every label as a side effect. It runs on the **normal** path whenever the memory layer injected facts — so with memory on, a labeled program loses its labels after the first `llm()` call. The docs list `setMessages`'s label reset as affecting "summarization, `threadRepair`"; this is an undocumented third caller, and unlike the other two it isn't a rewrite at all.

### Root cause and suggested fix

All three are the same shape: **`setMessages` is a blunt "replace everything, reset labels", and it's being used for surgical edits.** Two of the three call sites don't want a rewrite — :1075 wants "remove the message at index i" and :977 wants "become this restored thread while keeping my identity."

That's the same argument the PR already makes for `addMessage` delegating to `push`: give the class the operation the caller actually wants, instead of asking each caller to remember the labels.

- `removeAt(index)` — splices **both** arrays. Fixes #3, and #1/#2's siblings for free.
- `adoptFrom(other: MessageThread)` — copies messages **and** labels, preserving identity. Fixes #2, and reads better than the current comment explaining why `setMessages` is being used as an aliasing trick. (It also drops the incidental aliasing: `restored.getMessages()` hands the restored thread's live inner array to `args.messages`, so two threads share one array while keeping separate `messageLabels`.)
- For #1, either store the full `MessageThreadJSON` in `self.messagesJSON` or add a parallel labels field.

`setMessages` then keeps its documented "a wholesale rewrite drops labels" meaning for the case that really is one (summarize).

If you'd rather ship as-is and fix later, that's defensible — this is observability-only — but then `docs/dev/threads.md`'s "Serialization: `messageLabels` round-trips through `toJSON`/`fromJSON`" needs a caveat, because in the runtime it doesn't. I'd fix it: the resumable-guards feedback channel is going to label messages **on the resume path** specifically.

### Test gap that let this through

`round-trips labels through toJSON/fromJSON` passes because it calls `fromJSON` directly. The consumer defeats it two different ways and no test notices. Worth one agency test: labeled `userMessage` → interrupt → resume → assert the label is still in the `promptCompletion` dump. The harness supports it (`interruptHandlers: [{"action": "approve"}]`, cf. `tests/agency/callback-resume.test.json`).

Smaller notes inline. Nothing below is blocking.

packages/agency-lang/lib/runtime/state/messageThread.ts
egonSchiele: Checked: they don't have one. Smoltalk's `MessageClass` exposes `content` / `role` / `name` / `rawData` / the `to*Message()` converters and `toJSON()` — no id — and the JSON schemas are `{role, content}`. So there's nothing stable to key on today.

The options I can see, and why I stayed with the array:

- **Invent our own ids** — assign a nanoid per message at push time. But the id has to live *somewhere*, and it's not on the message, so it goes in… a parallel array. Same alignment problem, one more thing to keep in sync.
- **Key by content hash** — breaks on duplicate messages, which are common (two identical tool results).
- **Wrap messages** in our own `{message, label}` record and derive `messages` from it. This is the one that actually makes misalignment unrepresentable. It's also the biggest change: `messages` is the thread's core type, handed to smoltalk and dumped as the `toJSON().messages` wire format that statelog/the viewer read, so it'd touch serialization and every reader.

What I did instead was take your point about `setMessages` seriously and close the gap from the caller side — the alignment now can't drift because no caller has a reason to reach past it (`push` / `removeAt` / `adoptFrom` / `setMessages(messages, labels?)`). That fixed both real bugs. If you want the wrapper anyway as the stronger guarantee, I'd rather do it as its own PR than bolt it on here — say the word.

packages/agency-lang/lib/runtime/state/messageThread.ts
egonSchiele: Four non-test callers: `prompt.ts:743` (summarize), `prompt.ts:977` (resume), `prompt.ts:1075` (memory's injected-facts removal), `threadRepair.ts:77`. Your other comment is right that two of those aren't rewrites at all — those are now `adoptFrom` and `removeAt`, which leaves `setMessages` to the two that really are (summarize, repair) plus `fromJSON`.

Took the zip suggestion: `setMessages(messages, labels?)`. It's better than the "assign after" I had, for a reason I'd missed — `fromJSON` loads through `setMessages`, so the ordering trap I documented only existed *because* labels couldn't ride along. Passing them removes the trap instead of documenting it, and it gives length-normalization one home (Copilot's point above): a `labels` array whose length disagrees is refused rather than padded or sliced, since a disagreement means the source is already wrong and guessing would put real labels on wrong messages. Unlabeled beats mislabeled.

Summarize and repair still pass nothing and still drop labels — they genuinely can't say what the labels of a rewritten thread are.

packages/agency-lang/lib/runtime/state/messageThread.ts
egonSchiele: Both confirmed and fixed.

`prompt.ts:1075` is the worse of the two — it's on the normal path whenever memory is on, so a labeled program lost every label after its first `llm()` call. Now `removeAt(i)`.

`prompt.ts:977` → `adoptFrom(restored)`, which keeps the alias and takes the labels.

Added your two methods as written. The framing is the useful part: the labels come along because the caller was given the operation it actually wanted, not because each caller remembers. Same argument as `addMessage` → `push`, and I should have applied it to the other callers when I made that one.

`setMessages` keeps its documented meaning for the two callers that really are rewrites (summarize, `threadRepair`), and now takes an optional `labels` for the one caller that knows them (`fromJSON`).

packages/agency-lang/lib/runtime/state/messageThread.ts
egonSchiele: Confirmed — the ordering fix was guarding a path that never carried labels in the first place. Traced it: all six snapshot sites (`snapshotMessages` at :999 plus :1080/:1634/:1660/:1727/:1747) wrote `toJSON().messages`, so resume hit the `Array.isArray` branch and `_messageLabels` was always `undefined`. My "labels survive interrupt/resume" claim in the docs was wrong.

Took the first of your two options: the snapshot now persists the full `MessageThreadJSON`. Back-compatible in both directions — `fromJSON` already accepted both shapes, so older checkpoints holding a bare array keep reviving through the legacy branch. `snapshotMessages`'s return type widened to `MessageThreadJSON | MessageJSON[]` to say so.

The ordering trap itself is gone rather than fixed: `fromJSON` now calls `setMessages(smoltalkMessages, _messageLabels)`, so there's no "assign after" to get wrong.

Pinned by `tests/agency-js/message-labels-resume` — a tool interrupts mid-`llm()`, the run resumes, and it asserts the labels on the post-resume `promptCompletion`. Verified it actually catches this: with the snapshot reverted to `.messages`, `labeledAfterResume` comes back `[]`.

packages/agency-lang/lib/runtime/state/messageThread.ts
egonSchiele: Fixed — `[...this.messageLabels]`. It was the one hole left in "only these writers touch the array", and handing out the live one while `messages` next to it is rebuilt made the asymmetry easy to miss. Test added that mutates the returned JSON and asserts the thread is unchanged.

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: Both taken.

Return type is now `(smoltalk.MessageJSON & { label?: string })[]`.

On the index join: you're right that it rests on `redactAttachments` being `value.map(...)`, and right that neither test went near it — plain text never reaches the attachment branch, so the assumption was pinned nowhere. Added a case to `withMessageLabels` with a base64 `image` part on the *labeled* message, asserting the pairing survives and that the blob really was redacted on that entry (so the test fails if it stops exercising the redaction path rather than passing vacuously).

That doesn't make us immune to a future smoltalk that filters or splits — nothing short of not joining by index would — but it does mean such a change fails a test here instead of silently shifting every label.

packages/agency-lang/lib/runtime/promptLabels.test.ts
egonSchiele: Widened — asserts on `configs[0].metadata` as well.

Agreed on the reasoning: it's safe today only because both `...clientConfig` and `metadata:` read the same post-strip object, which is exactly the kind of thing that stops being true without the test noticing. The assertion should track the property ("no label reaches the provider"), not the one shape it currently takes.

For what it's worth I did check the top-level one isn't vacuous — `restClientConfig` is spread into the provider config, and deleting the strip turns it red.

packages/agency-lang/lib/runtime/state/messageThread.ts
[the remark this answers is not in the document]
egonSchiele: Valid — normalization now lives in `setMessages(messages, labels?)`, which `fromJSON` calls, so a restore can't install a mismatched array.

Went with the fall-back-to-all-null option rather than slice/pad: a length disagreement means the source is already wrong, and padding or slicing would silently attach real labels to wrong messages. Wrong labels are worse than no labels here — the whole point of the invariant is that a desync mislabels rather than degrading — so the safe move is to refuse and drop to unlabeled.

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -861,14 +887,20 @@ export async function runPrompt(args: {
...
       tools?: any[];
       memory?: boolean | { model?: string };
       maxToolResultChars?: number;
+      label?: string;
     };
 
+  /** This llm() call's debug label, or null when unlabeled. Stamps the
+   *  call's promptStart event and the messages it appends. */
+  const callLabel: string | null = ccLabel || null;
+
```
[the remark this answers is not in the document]
egonSchiele: Correct, and it was a real bug — my "labels serialize with the thread" claim didn't hold through checkpoint/resume.

All six snapshot sites wrote `toJSON().messages`, so resume revived through `fromJSON`'s bare-array branch with no labels to read. They now persist the full `MessageThreadJSON`; `fromJSON` still accepts the bare array, so older checkpoints keep working. The resume path also called `setMessages(restored.getMessages())`, dropping whatever had been restored — that's `adoptFrom` now.

Pinned by `tests/agency-js/message-labels-resume`, verified to fail without the fix (`labeledAfterResume: []`).

egonSchiele:
## Review round 1 — all 8 threads answered, pushed as `bc85e321b`

Two of these were **real bugs that made the feature not work**, both found by the same observation: a caller reaching for `setMessages` for a job that wasn't a rewrite. Same class as `addMessage`, which I'd already fixed for exactly that reason and then failed to apply to the other callers.

**Labels were lost on the normal path.** The memory layer removed its injected-facts message with `setMessages([...slice, ...slice])`, resetting every label as collateral — so with memory on, a labeled program lost its labels after its first `llm()` call. Now `removeAt(i)`.

**Labels never survived interrupt/resume.** All six snapshot sites wrote `toJSON().messages`; that bare array revives through `fromJSON`'s legacy branch, which has no labels to read. The resume path then called `setMessages(restored.getMessages())`, discarding what had just been restored. So the "labels serialize with the thread" claim in my docs was wrong. The snapshot now persists the full `MessageThreadJSON` (back-compatible — `fromJSON` still takes the bare array, so in-flight checkpoints keep working) and resume uses `adoptFrom`.

### The invariant got stronger

The writer set is now `push` / `removeAt` / `adoptFrom` / `setMessages(messages, labels?)` / constructor. The framing from the `setMessages` thread is what drove it: labels come along because each caller was given the operation it actually wanted, not because each caller remembers.

`setMessages` taking `labels` also removed the `fromJSON` ordering trap I'd documented — the trap only existed because labels couldn't ride along — and gave length-normalization one home: a mismatched array is **refused**, not padded, since guessing an alignment would put real labels on wrong messages.

### On message IDs

Smoltalk messages have no id (`MessageClass` is content/role/name/rawData + converters), so there's nothing stable to key on. Details and the alternatives in that thread — the one that would make misalignment unrepresentable is wrapping messages in our own record, which I'd rather do as its own PR than bolt on here.

### Tests

- `tests/agency-js/message-labels-resume` — interrupt mid-`llm()`, resume, assert labels on the post-resume `promptCompletion`. **Verified it catches the bug**: with the snapshot reverted, `labeledAfterResume` comes back `[]`.
- `withMessageLabels` now pinned through the **redaction path with a real attachment**, where the smoltalk-is-1:1 assumption actually lives.
- Strip assertion widened to `metadata` as well as the top level.
- Unit tests per new writer (`removeAt`, `adoptFrom`, `setMessages` with/without labels, mismatched-length refusal, `toJSON` copy).

8331 units, lint clean, `make fixtures` zero churn, and the 5 related agency-js e2es green (`message-labels`, `message-labels-resume`, `parallel-tools-resume-messages-intact`, `llm-call-single-span`, `prompt-start-cancelled`).

Issue #555, opened 2026-07-16
Brainstorm: typed failure payloads (Result error is any, so failure-shape drift is invisible)


egonSchiele:
## The problem

A failure's `error` is typed `any`, so the checker cannot catch a missing or misspelled field on any failure payload. This is a structural blind spot, not a one-off.

Concrete example from the PR #554 review: the runtime emitted `label` on every guard failure, the `guard()` docstring promised `error.label`, a fixture read `error.label` — and the one place that defines the public shape, the exported `GuardFailureData` in `stdlib/thread.agency`, omitted the field. Everything compiled. Nothing flagged it. The mismatch was only caught by human review.

The same class of bug is possible today for every failure shape in the stdlib and in user code: any `result.error.<field>` access is unchecked, and any drift between what a function's failure actually carries and what its docs/types claim is invisible to the compiler.

## What "typed failure payloads" could mean

`Result` is currently success-typed but failure-erased: `Result<T>` knows the value type and nothing about the error. Directions to brainstorm, roughly in increasing ambition:

1. **A second type parameter with a default**: `Result<T, E = any>` — opt-in failure typing, zero migration pressure. `guard` could declare `Result<T, GuardFailureData>` and `result.error.label` becomes checked.
2. **Failure types on function signatures**: a `fails GuardFailureData` clause (or reuse of the `raises` machinery, which already threads effect sets through signatures and function types) so callers' `isFailure(r)` narrowing produces a typed `r.error`.
3. **Inferred failure unions**: the checker collects the failure shapes a body can produce (`failure(...)` call sites plus propagated callee failures) the way exhaustiveness analysis collects match arms. No annotations, but propagation makes the unions grow fast.

## Design questions

- How does this interact with strict failure propagation (#489/#492)? A propagated failure crosses call boundaries — does the type widen at each hop, or does propagation demand compatible failure types?
- What is the failure type of `try f() catch v` and of `__catchResult`?
- `failure("some string", data)` is the dominant idiom — is the payload the string, the data, or a structured union of both?
- Flow narrowing: `result.error.type === "timeoutFailure"` should narrow the payload the way discriminant matches already narrow unions. The existing narrowing machinery (docs/dev/typechecker/narrowing/) probably covers this once the error has a union type at all.
- Migration: everything is `any` today, so any default must be fail-open. What is the lint/strict-mode story for nudging stdlib and user code onto typed failures over time?
[...]

PR #558, opened 2026-07-16
Interrupt infrastructure for resumable guards: pass(), approval merge, handler scoping

egonSchiele:
PR 1 of 4 of the resumable-guards plan (`docs/superpowers/plans/2026-07-16-resumable-guards.md`, rev 8, committed on this branch with its full review trail). This PR upgrades the interrupt machinery itself and changes no guard behavior: everything here is exercised by ordinary interrupts, so the handler chain's changes are proven against the existing suites before guard trips start using them in PR 2.

## The three pieces

**`pass()` — an explicit "not my interrupt" verdict.** Identical in effect to a handler returning nothing, but usable in value position, which a `match` arm requires. The chain normalizes `undefined` to `pass` the moment a handler returns, so the loop, statelog (`decision: "pass"`), and merge logic see one spelling. Typechecker entry mirrors `propagate`.

**Per-effect approval merge.** When several handlers approve one interrupt, the values now combine through `mergeFor(effect)` (`lib/runtime/effectMerge.ts`) instead of the outermost silently overwriting. The table is total and CONSTANT — deliberately no registration surface (a runtime registry is per-run module-global state, diverges silently across the subprocess boundary, and would put user closures on the function-refs-across-checkpoints surface). The default merge IS the historical overwrite, byte-for-byte, so every existing effect is unchanged; `std::guard` accumulates (grants sum, disarm lists union, messages join). The cross-process merge (`mergeChainOutcomes`, now taking the effect) uses the same table, with its one historical asymmetry (valueless outer approve defers to inner over IPC) preserved and documented.

**Registration-site handler scoping.** A handler belongs to the scope where it was registered. Each entry on `ctx.handlers` carries `liveGuardIds` — the guards live at its registration — and while the handler runs, the raising branch suspends every guard not in that set: the handler's own `llm()` calls are metered by its registration site's guards only. The capture point is `Runner.handle` (where Agency `handle` blocks actually register), memoized first-write-wins in the registering frame's locals keyed by step path, deleted on pop — the design the plan spent four review rounds getting right (position keys, never event counts; `Runner.handle` returns before `pushHandler` for completed blocks on replay). All five `pushHandler` callers decide their guard set explicitly, including one the plan reviews had missed: the `--policy` handler (registered before any guard exists, `[]`).

## Two things review should look at hardest

**1. A deviation from the reviewed plan, found during implementation: cost-guard suspension is stack-scoped, not an object flag.** The plan specified `suspended` on the guard object with `check()`/`charge()` early-returns. That is fail-open for `CostGuard`: the object is SHARED across fork branches (`cloneForBranch` returns `this`), so an object flag would drop sibling branches' charges and open their gates while one branch's handler deliberates. Implemented instead as `StateStack.suspendedGuardIds` (branch-local, in-memory, never serialized), consulted inside `enforceGuards`/`chargeGuards` — call sites unchanged — with a save/restore bracket so nested handler chains compose. `TimeGuard` keeps the object-level flag (clones are per-branch objects) and pins its clock paused across `Runner.beforeStep`'s resume-all. The plan doc on this branch records the deviation where PR 2 will read it, because `scope.suspendAll()` at the raise site must use the same stack-scoped bracket.

**2. `preapprove()` now passes on `std::guard`.** Its auto-approve answers every interrupt with a bare `approve()` — which for a guard trip is `approve({})`, a grant of nothing that the trip machinery (PR 2) treats as a runtime error. Budget questions belong to outer handlers or the user, never a tool's auto-approve wrapper. One line, two unit tests.

## Honest coverage note

The "handler blocked by an already-over-budget guard" scenario (the pre-call-gate half of suspension) is unreachable in this PR: cost trips still THROW at the charging moment, so the over-budget-but-still-running state does not exist until PR 2 makes trips raise. That surface gets its real fixtures there. What this PR pins with spend-into-the-gap fixtures: charge scoping in both directions (a leak into the hidden guard trips it; a missing charge on the registration-site guard lets an extra call fit), the loop-iteration memo hazard, and that no suspension state survives a propagate-and-resume.

## Validation

- 8,354 unit tests green (baseline 8,333 + the new ones), including the full existing handler/interrupt suites on the new entry shape before any new behavior tests were added.
- Full guards fixture sweep: 64/64. Full handlers fixture sweep green, including 4 new fixtures (`handle-pass`, `handler-budget-scoping`, `handler-scope-loop`, `handler-scope-resume`).
- Structural lint clean; `make fixtures` zero churn beyond the one added `pass()` line in the generated imports preamble.
- Checkpoint-copies-frame-locals verified (`State.toJSON` deep-clones `locals`), which the memo's delete-on-pop rule rests on.

### DIFF
--- a/packages/agency-lang/docs/site/guide/handlers.md
+++ b/packages/agency-lang/docs/site/guide/handlers.md
@@ -181,6 +181,21 @@ Now, all of their interrupts get rejected. As long as all destructive actions ar
 
 Besides `approve` and `reject,` the other keyword is `propagate.` `propagate` means "I don't want to reject the interrupt, but I don't want anyone to be able to programmatically approve it either. I want to make sure it always goes to a user for approval or rejection."
 
+## Pass
+
+`pass()` means "not my interrupt — ask the next handler." It has the same effect as returning nothing, but it is a real value, so you can use it where a value is required. The common case is a `match` over the interrupt's effect: every arm must produce a value, and `pass()` is how an arm says it has no opinion.
+
+```ts
+handle {
+  doWork()
+} with (data) {
+  return match(data.effect) {
+    "std::read" => approve()
+    _ => pass()
+  }
+}
+```
+
 ## The rules of handlers
 
 The rules of handlers are thus:
@@ -188,7 +203,28 @@ The rules of handlers are thus:
 2. Otherwise, if any handler propagates, the interrupt propagates to the user for a decision.
 3. Otherwise, if a handler approves, the interrupt is approved.
 
-Of course, a handler doesn't need to approve, reject, or propagate. It can simply choose to log the interrupt data, print out the lyrics to "A Day in the Life," or whatever. If no handler approves, rejects, or propagates, by default, the interrupt propagates up to the user for a decision.
+Of course, a handler doesn't need to approve, reject, or propagate. It can `pass()`, log the interrupt data, print out the lyrics to "A Day in the Life," or whatever. If no handler approves, rejects, or propagates, by default, the interrupt propagates up to the user for a decision.
+
+### When several handlers approve
+
+If more than one handler approves the same interrupt, their approval values combine. How they combine depends on the interrupt's effect. For most effects, the outermost handler's value wins — that is the historical behavior, unchanged. Effects whose approvals carry data that should accumulate define their own merge: `std::guard` approvals (guard budget grants, in a coming release) add their grants together instead of overwriting.
+
+## A handler runs on its own budget
+
+A handler belongs to the place where it was registered. When a handler runs, its work — including its own `llm()` calls — is metered by the guards that enclosed the `handle` block itself, never by guards installed deeper. In this example, the handler's `llm()` call charges the outer `$20` guard and is invisible to the inner `$1` guard, even though the interrupt was raised inside it:
+
+```ts
+guard(cost: $20.00) as {
+  handle {
+    guard(cost: $1.00) as {
+      riskyWork()   // raises an interrupt
+    }
+  } with (data) {
+    const opinion = llm("should this proceed?")  // charges the $20 guard only
+    return approve()
+  }
+}
+```
 
 ## Handlers can't raise interrupts
 
--- a/packages/agency-lang/docs/site/guide/handlers.md
+++ b/packages/agency-lang/docs/site/guide/handlers.md
@@ -181,21 +181,6 @@ Now, all of their interrupts get rejected. As long as all destructive actions ar
 
 Besides `approve` and `reject,` the other keyword is `propagate.` `propagate` means "I don't want to reject the interrupt, but I don't want anyone to be able to programmatically approve it either. I want to make sure it always goes to a user for approval or rejection."
 
-## Pass
-
-`pass()` means "not my interrupt — ask the next handler." It has the same effect as returning nothing, but it is a real value, so you can use it where a value is required. The common case is a `match` over the interrupt's effect: every arm must produce a value, and `pass()` is how an arm says it has no opinion.
-
-```ts
-handle {
-  doWork()
-} with (data) {
-  return match(data.effect) {
-    "std::read" => approve()
-    _ => pass()
-  }
-}
-```
-
 ## The rules of handlers
 
 The rules of handlers are thus:
@@ -203,28 +188,7 @@ The rules of handlers are thus:
 2. Otherwise, if any handler propagates, the interrupt propagates to the user for a decision.
 3. Otherwise, if a handler approves, the interrupt is approved.
 
-Of course, a handler doesn't need to approve, reject, or propagate. It can `pass()`, log the interrupt data, print out the lyrics to "A Day in the Life," or whatever. If no handler approves, rejects, or propagates, by default, the interrupt propagates up to the user for a decision.
-
-### When several handlers approve
-
-If more than one handler approves the same interrupt, their approval values combine. How they combine depends on the interrupt's effect. For most effects, the outermost handler's value wins — that is the historical behavior, unchanged. Effects whose approvals carry data that should accumulate define their own merge: `std::guard` approvals (guard budget grants, in a coming release) add their grants together instead of overwriting.
-
-## A handler runs on its own budget
-
-A handler belongs to the place where it was registered. When a handler runs, its work — including its own `llm()` calls — is metered by the guards that enclosed the `handle` block itself, never by guards installed deeper. In this example, the handler's `llm()` call charges the outer `$20` guard and is invisible to the inner `$1` guard, even though the interrupt was raised inside it:
-
-```ts
-guard(cost: $20.00) as {
-  handle {
-    guard(cost: $1.00) as {
-      riskyWork()   // raises an interrupt
-    }
-  } with (data) {
-    const opinion = llm("should this proceed?")  // charges the $20 guard only
-    return approve()
-  }
-}
-```
+Of course, a handler doesn't need to approve, reject, or propagate. It can simply choose to log the interrupt data, print out the lyrics to "A Day in the Life," or whatever. If no handler approves, rejects, or propagates, by default, the interrupt propagates up to the user for a decision.
 
 ## Handlers can't raise interrupts
 
--- a/packages/agency-lang/lib/ir/prettyPrint.ts
+++ b/packages/agency-lang/lib/ir/prettyPrint.ts
@@ -386,7 +386,10 @@ export function printTs(node: TsNode, indent = 0): string {
       // instead of producing a cryptic "Cannot read 'pushHandler' of
       // undefined". Handlers are safety infrastructure — failing loud
       // here is preferable to silently skipping registration.
-      return `getRuntimeContext().ctx.pushHandler(${handler});\n${ind(indent)}try {\n${body}\n${ind(indent)}} finally {\n${ind(indent + 1)}getRuntimeContext().ctx.popHandler();\n${ind(indent)}}`;
+      // liveGuardIds: [] — an explicit decision, not a default. These
+      // wrappers register during top-level init, before any guard can
+      // exist, so "hide every guard" is vacuously correct.
+      return `getRuntimeContext().ctx.pushHandler(${handler}, []);\n${ind(indent)}try {\n${body}\n${ind(indent)}} finally {\n${ind(indent + 1)}getRuntimeContext().ctx.popHandler();\n${ind(indent)}}`;
     }
 
     case "runnerDebugger": {
--- a/packages/agency-lang/lib/ir/prettyPrint.ts
+++ b/packages/agency-lang/lib/ir/prettyPrint.ts
@@ -1,5 +1,6 @@
 import type { TsNode, TsParam, TsScopedVar, TsImportName } from "./tsIR.js";
 import renderRunnerIfElse from "../templates/backends/typescriptGenerator/runnerIfElse.js";
+import renderWithHandlerWrapper from "../templates/backends/typescriptGenerator/withHandlerWrapper.js";
 import renderAgencyFunctionWrap from "../templates/ir/agencyFunctionWrap.js";
 
 const INDENT = "  ";
@@ -388,8 +389,14 @@ export function printTs(node: TsNode, indent = 0): string {
       // here is preferable to silently skipping registration.
       // liveGuardIds: [] — an explicit decision, not a default. These
       // wrappers register during top-level init, before any guard can
-      // exist, so "hide every guard" is vacuously correct.
-      return `getRuntimeContext().ctx.pushHandler(${handler}, []);\n${ind(indent)}try {\n${body}\n${ind(indent)}} finally {\n${ind(indent + 1)}getRuntimeContext().ctx.popHandler();\n${ind(indent)}}`;
+      // exist, so "hide every guard" is vacuously correct. (The []
+      // itself lives in the template.)
+      return renderWithHandlerWrapper({
+        handler,
+        body,
+        indent: ind(indent),
+        indentInner: ind(indent + 1),
+      });
     }
 
     case "runnerDebugger": {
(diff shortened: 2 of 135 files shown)


packages/agency-lang/docs/site/guide/handlers.md
```diff
@@ -181,14 +181,50 @@ Now, all of their interrupts get rejected. As long as all destructive actions ar
 
 Besides `approve` and `reject,` the other keyword is `propagate.` `propagate` means "I don't want to reject the interrupt, but I don't want anyone to be able to programmatically approve it either. I want to make sure it always goes to a user for approval or rejection."
 
+## Pass
```
egonSchiele: undo all edits to this file

packages/agency-lang/lib/ir/prettyPrint.ts
```diff
@@ -386,7 +386,10 @@ export function printTs(node: TsNode, indent = 0): string {
       // instead of producing a cryptic "Cannot read 'pushHandler' of
       // undefined". Handlers are safety infrastructure — failing loud
       // here is preferable to silently skipping registration.
-      return `getRuntimeContext().ctx.pushHandler(${handler});\n${ind(indent)}try {\n${body}\n${ind(indent)}} finally {\n${ind(indent + 1)}getRuntimeContext().ctx.popHandler();\n${ind(indent)}}`;
+      // liveGuardIds: [] — an explicit decision, not a default. These
+      // wrappers register during top-level init, before any guard can
+      // exist, so "hide every guard" is vacuously correct.
+      return `getRuntimeContext().ctx.pushHandler(${handler}, []);\n${ind(indent)}try {\n${body}\n${ind(indent)}} finally {\n${ind(indent + 1)}getRuntimeContext().ctx.popHandler();\n${ind(indent)}}`;
```
egonSchiele: please put this in a typestache template (lib/templates)

packages/agency-lang/lib/runtime/state/stateStack.ts
```diff
@@ -569,11 +570,68 @@ export class StateStack {
   enforceGuards(): void {
     this.assertNoParkedGuards();
     for (let i = this.guards.length - 1; i >= 0; i--) {
+      if (this.suspendedGuardIds.includes(this.guards[i].guardId)) continue;
```
egonSchiele, line 590: This is the fix for the shared-CostGuard hazard and it is correct here — but `enforceGuards` is not the only walk over `stack.guards` calling `check()`.

`Runner.shouldSkip` (runner.ts:336) walks innermost-first and calls `check()` directly, with no `suspendedGuardIds` consult. `TimeGuard` survives that because its object flag makes `check()` return null; `CostGuard` cannot, because its `suspend()` is a deliberate no-op (guard.ts:226) — the object does not know.

Net: a suspended, over-budget cost guard throws its trip out of the handler that suspended it, at the handler body's next step, whenever the branch's abort signal is live (outer time guard, race-loser cancel, Esc). Latent now; normal in PR 2, where a handler adjudicating a cost trip runs with exactly that guard suspended and over budget for the whole deliberation.

Suggest collapsing the two walks into one suspension-aware method on `StateStack` — `firstTrippedGuard(): Guard | null` — with `enforceGuards` and `shouldSkip` both throwing what it returns. That is also PR 2's `detectTrippedGuard`, arriving early instead of as a third copy of the same loop. Minimal alternative: repeat the `continue` line in `shouldSkip` and cross-reference this one.

packages/agency-lang/lib/runtime/guard.ts
```diff
@@ -196,6 +215,22 @@ export class CostGuard implements Guard {
...
 
+  /** Deliberately no-ops. A CostGuard can be SHARED across fork branches
+   *  (cloneForBranch returns `this`), so an object-level suspended flag
+   *  would blind SIBLING branches too: their charges would drop and
+   *  their gates would open while one branch's handler deliberates —
+   *  fail-open on the shared budget. Cost suspension is therefore
+   *  branch-scoped, on the StateStack (`suspendedGuardIds`), where the
+   *  enforce/charge walks consult it. Only per-branch-object state
+   *  belongs here, which for cost is nothing. */
+  suspend(): void {
```
egonSchiele, line 234: The reasoning here is right and the comment is the kind that survives — anyone tempted to "finish" this method will read why it is empty first.

One addition worth making while you are here: say that because this is a no-op, **every** cost-suspension check must go through `StateStack.suspendedGuardIds`, and that any `check()` call site outside `enforceGuards`/`chargeGuards` therefore does not honor suspension. That is currently true of `Runner.shouldSkip` (see my comment on stateStack.ts:573), and the comment is where a future reader would look before adding a third call site.

packages/agency-lang/lib/runtime/effectMerge.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (b === undefined) return a;
+  return a + sep + b;
+}
+
+/** std::guard approvals accumulate: budget grants add, disarm lists
+ *  union, messages concatenate in inner-to-outer order. Two handlers
+ *  each granting $0.50 means $1.00 of new budget, not a coin flip over
+ *  which fifty cents survives. */
+const mergeGuardApprovals: ApprovalMerge = (a, b) => ({
+  maxCost: sumOrUndefined(a?.maxCost, b?.maxCost),
```
egonSchiele, line 82: Nit, for PR 2's benefit rather than this one. `reduce(merge)` with no initial value returns a lone approval **as-is**, without calling the merge — so one handler granting `{maxCost: 0.5}` yields `{maxCost: 0.5}`, while two handlers yield the full four-key shape with `maxTime`/`disarm`/`message` explicitly `undefined`. Same information, two shapes, decided by how many handlers happened to answer.

Harmless as long as PR 2's `GuardScope.extend` reads keys defensively (`payload.maxCost !== undefined`) rather than, say, `"maxTime" in payload`. Worth one line in the docstring saying the merged payload's keys are optional either way, so the extend side is written to the looser contract from the start.

egonSchiele (commented):
## Review: PR 1 — interrupt infrastructure

This is the plan executed faithfully, and the two things you asked me to look at hardest are both right. Taking them in order:

**The deviation is correct, and it is a better design than the plan's.** I checked the reasoning: `CostGuard.cloneForBranch` returns `this`, so an object-level `suspended` would have blinded every sibling branch — their charges dropped, their gates opened — while one branch's handler deliberates. Fail-open on a shared budget, and invisible until a fork fixture happened to spend at the wrong moment. `StateStack.suspendedGuardIds` is the right shape: branch-local, unserialized, consulted where the walks already are, call sites unchanged. `TimeGuard` keeping the object flag is right for the same reason inverted (clones are per-branch objects), and gating `resume()` on it is what makes it survive `beforeStep`'s resume-all.

**`preapprove()` passing on `std::guard` is right** and was your catch, not the plan's reviewer's.

Making `pushHandler`'s second argument **mandatory** deserves a mention: it turns "empty is not neutral" from a rule someone has to remember into a compile error at any caller that forgets. That is the right response to a hazard that only bites at runtime, and it is why I believe the claim that all five callers are ruled — the compiler believes it too.

### One finding: `Runner.shouldSkip` bypasses the suspension

Suspension is enforced in `enforceGuards` and `chargeGuards`, which consult `suspendedGuardIds`. But those are not the only places `check()` is called. `Runner.shouldSkip` (runner.ts:336) does its own walk:

```ts
for (let i = this.stack.guards.length - 1; i >= 0; i--) {
  const err = this.stack.guards[i].check(this.stack);
  if (err) throw err;
}
```

No `suspendedGuardIds` consult. `TimeGuard` is covered — its `check()` returns null on the object flag. `CostGuard` is not, and cannot be: its `suspend()` is a deliberate no-op, so the object has no idea it is suspended and `check()` happily returns the trip.

So: a suspended, over-budget cost guard throws its trip out of the handler that was suspending it, at the handler body's next step, if the branch's abort signal has fired for any reason (an outer time guard, a race-loser cancel, Esc).

**This is nearly unreachable in PR 1** — as the PR body says, an over-budget-but-still-running cost guard barely exists while trips still throw at the charge site. **It becomes the normal state in PR 2**: a handler adjudicating a cost trip runs with exactly that guard suspended and over budget, for as long as the reviewer agent takes. Any abort during deliberation and the trip escapes the chain sideways.

Worth fixing here, while the invariant is being established rather than after it has a second consumer. The two walks are the same walk with one difference nobody intended: innermost-first `check()` over `stack.guards`, one consulting suspension and one not. Options, roughly in order of preference:

- Give `StateStack` the one walk both callers share (something like `firstTrippedGuard(): Guard | null`, suspension-aware), and have `enforceGuards` throw what it returns and `shouldSkip` throw what it returns. This is also the `detectTrippedGuard` sibling PR 2 needs — it would arrive early rather than as a third spelling of the same loop.
- Or, minimally: the same `if (this.suspendedGuardIds.includes(...)) continue;` line in `shouldSkip`'s walk, with a comment tying it to stateStack.ts:573.

Either way it wants a note in the plan doc next to the deviation, since PR 2's `scope.suspendAll()` at the raise site is the thing that makes it load-bearing.

### Smaller

Inline. Nothing else blocking — the memo, the delete-on-pop, the nesting bracket, and the merge table all read correctly to me, and the honest coverage note about the pre-call-gate half is the right call rather than a fixture that pretends.

packages/agency-lang/lib/runtime/agencyFunction.ts
```diff
@@ -261,11 +261,22 @@ export class AgencyFunction {
     // Doing the wrap here — rather than branching inside `invoke()` —
     // keeps the per-call hot path branch-free.
     const original = this._fn;
+    // The auto-approve must PASS on std::guard interrupts (guard trips,
```
egonSchiele, line 264: this change is temporary till the merge logic lands -- lets make a note so we remember to change it back after that

packages/agency-lang/lib/runtime/effectMerge.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (a === undefined) return b;
+  if (b === undefined) return a;
+  return a + sep + b;
+}
+
+/** std::guard approvals accumulate: budget grants add, disarm lists
+ *  union, messages concatenate in inner-to-outer order. Two handlers
+ *  each granting $0.50 means $1.00 of new budget, not a coin flip over
+ *  which fifty cents survives. */
+const mergeGuardApprovals: ApprovalMerge = (a, b) => ({
```
egonSchiele, line 81: this is great, note that we will want to let users define their merge logic as agency code (future improvement)

packages/agency-lang/docs/site/guide/handlers.md
egonSchiele: Reverted to the main version in 550f2e248.

packages/agency-lang/lib/ir/prettyPrint.ts
egonSchiele: Moved in 550f2e248: lib/templates/backends/typescriptGenerator/withHandlerWrapper.mustache, rendered via the compiled template in prettyPrint. Output is byte-identical (make fixtures shows zero churn) and the liveGuardIds [] now lives in the template with the explanation kept at the emission site.

packages/agency-lang/lib/runtime/state/stateStack.ts
egonSchiele: Fixed in 550f2e248, first option: both walks collapse into StateStack.detectTrippedGuard(): GuardExceededError | null — suspension-aware, innermost-first. enforceGuards throws what it returns; shouldSkip throws what it returns; the private loop in shouldSkip is gone. It returns the ERROR rather than the guard so both callers keep their throw semantics, and the error already carries the guardId PR 2's raise needs. A unit test pins the suspended-over-budget case at the walk level, and the plan doc's deviation note now records this as the second execution note so PR 2's suspendAll design reads it.

packages/agency-lang/lib/runtime/guard.ts
egonSchiele: Added in 550f2e248 — the comment now states the consequence: the object cannot decline its own check(), every cost-suspension decision lives in StateStack.suspendedGuardIds, and any check() call outside the stack's suspension-aware walk does not honor suspension. With the detectTrippedGuard collapse there is now exactly one such walk, and the comment says to route through it rather than add another.

packages/agency-lang/lib/runtime/effectMerge.ts
egonSchiele: Added to the mergeGuardApprovals docstring in 550f2e248: every key of the merged payload is optional either way — reduce never calls the merge for a lone approval — so consumers read keys defensively, never by presence. PR 2's GuardScope.extend will be written to that contract.

packages/agency-lang/lib/runtime/agencyFunction.ts
egonSchiele: Noted in 550f2e248 as a TODO on the branch. One case to re-check before removing it, recorded in the TODO: a lone approve({}) — with no other handler granting — is the useless-approval runtime error under decision 8, where a pass lets the trip propagate to someone who can actually grant budget. So the branch may turn out to be permanent; the TODO frames it as a decision to make when PR 2's trip semantics land rather than an automatic revert.

packages/agency-lang/lib/runtime/effectMerge.ts
egonSchiele: Noted in the same docstring: user-defined merges as Agency code, arriving with the typed-payloads work (#555), which also owns the serialization story a user-supplied merge function needs (the FunctionRefReviver surface).

packages/agency-lang/lib/runtime/guard.test.ts
```diff
@@ -449,3 +449,80 @@ describe("guard labels", () => {
...
+    const g = new TimeGuard(60000);
+    const stack = new StateStack();
+    g.install(stack);
+    g.suspend();
+    // Runner.beforeStep resumes every guard at every step entry — a
+    // handler body executes steps, so this MUST stay a no-op while
+    // suspended or the clock restarts mid-deliberation.
+    g.resume(stack);
+    expect(g.snapshotElapsed()).toBe(g.snapshotElapsed()); // clock frozen: no running window
+    expect((g as any).state).toBe("paused");
```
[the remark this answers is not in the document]
egonSchiele: Dropped in 550f2e248 — the state assertion is the deterministic signal, kept with a comment saying so.

Issue #559, opened 2026-07-17
Resume corrupts frames when an interrupt was raised inside a compound call (FIFO frame adoption vs replayed sibling calls)


egonSchiele:
## The bug

Resuming a checkpoint whose interrupt was raised inside a COMPOUND call statement — `const x = f(g())` where `g` completed before the interrupt — corrupts the replay. Frame adoption on resume is FIFO (`StateStack.getNewState` shifts restored frames in push order), and the replayed statement re-executes `g()`, which adopts the next restored frame — `f`'s. `g` then runs against `f`'s step counters (skipping its own body), returns undefined, pops `f`'s frame off the restored stack, and `f` re-executes on a fresh frame from step 0 with wrong arguments.

Concrete repro shape (from the resumable-guards work, where it surfaced):

```
def f2b(v: string): string {
  saveDraft("f2-partial:${v}")
  const reply = llm("...")   // raises an interrupt (guard trip / tool interrupt)
  return reply
}

def code(): string {
  const x = f2b(gOk())       // gOk completed before the raise
  ...
}
```

On resume after answering the interrupt, `f2b` replays with `v === undefined` and re-saves a corrupted draft: `"f2-partial:undefined"`.

## Why it has been invisible

The typechecker hoists STATICALLY-visible interrupting calls to statement position, so `f(g())` with a compile-time-known interrupting `f` never exists in generated code. Runtime-only interrupt sources bypass the hoist:

- tool interrupts inside `f`'s `llm()` call — this bug exists on main today for that shape;
- guard trips (PR 2 of resumable guards) — which made it reachable in the guards suite (`tests/agency/guards/finalize-nested-call-binding.agency`, `f2TripsCase`; currently pinned with an in-program reject that avoids the resume entirely, with a comment pointing here).

## Fix directions

1. Codegen: hoist EVERY call-bearing argument to a checked temp (universal statement-position lowering) — replay then never re-executes a completed sibling call. Big but principled; mirrors what the static hoist already does.
2. Runtime: make frame adoption identity-aware instead of FIFO (frames would need a serialized identity — the same class of fix as the resumable-guards key work: position or content, never order).

Option 1 seems right; option 2 patches the symptom.

PR #560, opened 2026-07-17
Cost-guard trips raise resumable interrupts (resumable guards PR 2)

egonSchiele:
PR 2 of 4 of the resumable-guards plan (rev 8 + execution notes, committed on this branch). Cost-guard trips now raise ordinary interrupts (effect `std::guard`) instead of throwing: handlers approve more budget and the work continues in place, reject and the existing salvage pipeline runs unchanged, or the question reaches the user across a checkpoint. This is the breaking change you accepted: a bare, unhandled guard no longer degrades to a failure — it asks.

## How it works

**GuardScope** (`lib/runtime/guardScope.ts`) is the runtime name for what one `guard(...)` call pushes — up to two runtime guards, two ids. The trip interrupt carries `scopeIds`; approve resolves the scope on the raising branch's stack (never a global lookup — fork clones share the parent's id) and applies the merged payload: additive grants, negative deltas clamped to zero with a warning, explicit `disarm`, root-budget refusal, and the useless-approval runtime error (an answer leaving the tripped dimension over budget and armed would re-trip forever).

**The raise** happens at idempotent PromptRunner gate steps in `runPrompt` — before every request and once before returning — looping on `detectTrippedGuard()` until the stack is clear, because each budget is owed its own question (approving an inner guard can push an outer one over). An unanswered trip returns `Interrupt[]` to `pr.step`, whose existing snapshot/checkpoint/bailout machinery surfaces it. On resume, the gate re-detects and applies the recorded answer via a key derived from guard state (`scopeIds#dimension@limit` — stable across replays of one trip, necessarily fresh for the next). Concurrent branches under a shared guard dedupe on a live `pendingTrip` record: set before the first await, settled in the `finally` on every exit, parked siblings re-detect.

**What still throws:** root budgets (now marked with a serialized `isRootBudget`, stamped by `installRootBudget`, refused by `extend`), the parent-side subprocess telemetry site (an IPC callback has no runner — documented v1 limit, which also means `run(maxCost:)`'s `limit_exceeded` contract is unchanged with zero migration), and `addCost` (below).

## Execution deviations for your sign-off

The plan doc on this branch records all of these in a "PR 2 as executed" section:

1. **The raise sites moved from the in-`_runPrompt` checks to pr.step-level gates.** Raising inside a non-idempotent llm-call step body breaks replay — the body re-pushes its user message on resume. The gates are idempotent steps of their own and reuse PromptRunner's bailout machinery instead of duplicating it. The pre-call check stays as a throwing backstop; the post-charge check is removed (its trips raise at the next gate, and nothing paid runs in between — tool-internal paid work is gated by the tool's own runPrompt).
2. **`addCost` keeps throwing in v1** — same replay hazard, from arbitrary TS-helper contexts. A `guard(cost:)` around a paid TS helper trips non-resumably for now.
3. **`std::agents` guards are NOT auto-reject wrapped**, deviating from the migration rule: an inner reject short-circuits ahead of user handlers and would permanently lock std::agents users out of approving trips — the flagship reviewer pattern. Their trips now surface to the caller, which headless default-reject answers exactly as before.
4. **The fixture migration went through the harness** (`{"action": "reject"}` in test.json) rather than in-program wrappers: outputs stay byte-identical AND all 36 migrated fixtures now exercise the full propagate → checkpoint → reject → resume → salvage pipeline.
5. **Subprocess trip e2e rides with PR 3**, where re-arm makes cross-process approve meaningful for the remaining dimension; the IPC site itself is untouched.

## Two discoveries

- **A new boundary fix:** an `Interrupt[]` reaching a call *argument* now passes through `__call`/`__callMethod` as the call's own result, mirroring `findAbortedArg`. Guard trips are the first runtime-only interrupt source, so `f(g())` with a trip inside `g` was the first way an interrupt batch could reach an argument slot — it previously flowed into the callee as `"[object Object]"`.
- **A pre-existing resume bug, filed as #559:** resuming an interrupt raised inside a compound call (`f(g())` with `g` completed) corrupts replay via FIFO frame adoption — the replayed `g()` adopts `f`'s restored frame. This affects tool interrupts on main today; trips just made it reachable in the guards suite. The one affected fixture case rejects in-program with a pointer to the issue.

## Testing

New fixtures pin: inline approve-and-continue; one approval extending BOTH members of a cost+time scope (the two-ids case that killed the rev-2 plan); explicit disarm; the useless-approval and negative-grant errors; two handlers accumulating through the merge, spend-shaped so overwrite semantics fail it; **the headline approve-across-a-real-checkpoint** — no in-program handler, the harness answers with a valued `{"action": "resolve", "resolvedValue": {"maxCost": ...}}`, the block resumes on the restored, extended guard, and a second trip raises under a fresh key and is rejected into the salvage pipeline; inside-guard handler blindness; propagate discarding approvals; the `draftValue` preview plus authoritative reject salvage; grants never crossing budgets (the inner grant pushes the outer over; it trips separately under its own label); shared-guard dedupe under fork (handler decides on `i.data.spent` — a module-global grant counter silently resets across branch clones, which the fixture comment warns about); and a race loser settling cleanly.

Full validation: 8,369 unit tests; 92/92 guards+handlers fixture files; structural lint; zero codegen fixture churn.

### DIFF
--- a/packages/agency-lang/lib/runtime/call.ts
+++ b/packages/agency-lang/lib/runtime/call.ts
@@ -6,6 +6,7 @@ import {
   describeFailureCallTarget,
 } from "./failurePropagation.js";
 import { isFailure } from "./result.js";
+import { hasInterrupts } from "./interrupts.js";
 import { type AbortedResult, isAborted } from "./abortedResult.js";
 
 /** Find an aborted result among a call's arguments (`f(g())` where g was
@@ -17,6 +18,32 @@ import { type AbortedResult, isAborted } from "./abortedResult.js";
  *  Interrupt[] does today; the abort signal is still firing, so the
  *  callee's next leaf op re-trips. Both systems share the fix whenever
  *  one lands. */
+/** An Interrupt[] must never enter another call as an argument either —
+ *  same boundary as findAbortedArg, same shape. Before guard trips, an
+ *  interrupting call in argument position was impossible: the checker
+ *  hoists statically-visible interrupting calls to statement position.
+ *  A guard trip is the first RUNTIME-ONLY interrupt source, so `f(g())`
+ *  where g trips inside is the first way an Interrupt[] can reach an
+ *  argument slot — without this check it would flow into the callee as
+ *  a value ("[object Object]"). Pass it through as the call's own
+ *  result so the caller's post-call interrupt guard halts and the
+ *  batch surfaces normally. */
+function findInterruptArg(descriptor: CallType): unknown | undefined {
+  const positional =
+    descriptor.type === "positional"
+      ? descriptor.args
+      : descriptor.positionalArgs;
+  for (const arg of positional) {
+    if (hasInterrupts(arg)) return arg;
+  }
+  if (descriptor.type === "named") {
+    for (const value of Object.values(descriptor.namedArgs)) {
+      if (hasInterrupts(value)) return value;
+    }
+  }
+  return undefined;
+}
+
 function findAbortedArg(descriptor: CallType): AbortedResult | undefined {
   const positional =
     descriptor.type === "positional"
@@ -65,6 +92,10 @@ export async function __call(
   if (abortedArg !== undefined) {
     return abortedArg.droppedAtArgPosition();
   }
+  const interruptArg = findInterruptArg(descriptor);
+  if (interruptArg !== undefined) {
+    return interruptArg;
+  }
   if (AgencyFunction.isAgencyFunction(target)) {
     return target.invoke(descriptor);
   }
@@ -109,8 +140,13 @@ export async function __callMethod(
     return undefined;
   }
 
-  // Same argument rule as __call: an aborted result never enters a method
-  // call; the abort continues without its partial.
+  // Same argument rules as __call: an aborted result never enters a
+  // method call (the abort continues without its partial), and an
+  // Interrupt[] passes through as the call's own result.
+  const interruptArg = findInterruptArg(descriptor);
+  if (interruptArg !== undefined) {
+    return interruptArg;
+  }
   const abortedArg = findAbortedArg(descriptor);
   if (abortedArg !== undefined) {
     return abortedArg.droppedAtArgPosition();
--- a/packages/agency-lang/lib/runtime/guard.ts
+++ b/packages/agency-lang/lib/runtime/guard.ts
@@ -120,6 +120,14 @@ export type Guard = {
    *  dimension in this state would re-trip forever and is a runtime
    *  error attributed to the answering handler. */
   overBudgetAndArmed(): boolean;
+  /** An open trip question for this guard, while one is being decided.
+   *  LIVE-ONLY (never serialized) and meaningful only for guards shared
+   *  across branches (cost): a sibling branch detecting the same guard
+   *  over budget parks on `settled` instead of asking its own question.
+   *  Time clones are per-branch objects, so for them this is always
+   *  undefined — cost-only dedupe, by construction, on purpose. Set
+   *  and cleared exclusively by guardTripInterrupt.ts. */
+  pendingTrip?: { settled: Promise<void>; settle: () => void };
   /** The current limit and spend in this guard's own unit (dollars /
    *  ms). On the interface for the trip interrupt's snapshot and the
    *  derived trip key — both must read them without variant branching. */
@@ -814,7 +822,9 @@ export class GuardExceededError extends AgencyAbort {
     // a GuardExceededError built WITHOUT one carries guardId === "" and, once
     // __tryCall filters on ownedGuardIds, would fail the membership check and
     // escape its own guard. CostGuard.check / TimeGuard.check pass their id.
-    guardId: string = "",
+    // Public: the trip-raise loop (guardTripInterrupt.ts) resolves the
+    // tripped guard object from it.
+    public readonly guardId: string = "",
     label?: string,
   ) {
     super(
--- a/packages/agency-lang/lib/runtime/guard.ts
+++ b/packages/agency-lang/lib/runtime/guard.ts
@@ -17,6 +17,26 @@ export function nextGuardId(): string {
   return `g${__guardIdCounter}`;
 }
 
+/** Clamp a budget grant at zero, warning on negatives. A handler that
+ *  computes `approve({maxCost: budget - spent})` and goes negative must
+ *  not silently REMOVE metering (negative-as-disable is a construction-
+ *  time convention only; in the additive approve channel it would be
+ *  fail-open on a cost-control feature). Disarming is explicit:
+ *  `approve({disarm: ["cost"]})`. */
+function clampGrant(
+  delta: number,
+  g: { dimension: "cost" | "time"; label?: string },
+): number {
+  if (delta >= 0) return delta;
+  console.warn(
+    `guard grant clamped to 0: a negative ${g.dimension} grant` +
+      `${g.label ? ` for guard "${g.label}"` : ""} (${delta}) does not ` +
+      `disarm the guard — use approve({disarm: ["${g.dimension}"]}) to ` +
+      `stop metering explicitly.`,
+  );
+  return 0;
+}
+
 /**
  * Per-guard scope state for cost / time limits. Held on
  * `StateStack.guards` as an array; `pushGuard` appends + calls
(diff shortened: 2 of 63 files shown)


packages/agency-lang/lib/runtime/call.ts
```diff
@@ -17,6 +18,32 @@ import { type AbortedResult, isAborted } from "./abortedResult.js";
  *  Interrupt[] does today; the abort signal is still firing, so the
  *  callee's next leaf op re-trips. Both systems share the fix whenever
  *  one lands. */
+/** An Interrupt[] must never enter another call as an argument either —
+ *  same boundary as findAbortedArg, same shape. Before guard trips, an
+ *  interrupting call in argument position was impossible: the checker
+ *  hoists statically-visible interrupting calls to statement position.
+ *  A guard trip is the first RUNTIME-ONLY interrupt source, so `f(g())`
```
egonSchiele, line 25: This is a good call-out. We actually do some static analysis of interrupts to see what interrupts can get thrown from a function. Users can define what sorts of interrupts a function is allowed to throw, using the `raises` keyword (https://agency-lang.com/guide/effects-and-raises.html). Does this mean that for interrupts thrown by guards, we aren't able to do this kind of static analysis? What would it take to add support for this? It would be a shame to not be able to offer the same sort of static type checking safety guarantees for std::guard interrupts as we do for all other interrupts.

packages/agency-lang/lib/runtime/guardScope.ts
```diff
@@ -0,0 +1,168 @@
...
+  /** The scope a tripped guard belongs to, resolved on `stack`. Members
+   *  are matched innermost-first by id. A guard with an empty scopeIds
+   *  (root budgets, agency.withCostGuard) is its own single-member
+   *  scope. Returns null when no member is present on this stack —
+   *  callers treat that as a stale or foreign answer, a runtime error,
+   *  never a silent no-op. */
+  static resolve(stack: StateStack, tripped: Guard): GuardScope | null {
+    const ids = tripped.scopeIds.length > 0 ? tripped.scopeIds : [tripped.guardId];
+    const members: Guard[] = [];
+    for (let i = stack.guards.length - 1; i >= 0; i--) {
```
egonSchiele: prefer forEach loops to this C-style loop

packages/agency-lang/lib/runtime/guardTripInterrupt.ts
```diff
@@ -0,0 +1,228 @@
...
+ *  (2) a clamped-to-zero grant errors (decision 8) before the key could
+ *  be reused; (3) disarm repeats the key, but a disarmed dimension never
+ *  trips again; (4) a reject aborts the scope — no next trip; (5) the
+ *  other dimension differs in the `#dimension` segment. */
+function guardTripKey(g: Guard): string {
+  const scopeIds = g.scopeIds.length > 0 ? g.scopeIds : [g.guardId];
+  return `__guardTrip_${scopeIds.join(",")}#${g.dimension}@${g.currentLimit()}`;
+}
+
+function buildTripMessage(s: {
```
egonSchiele, line 206: use a typestache template instead to make this easier to read

packages/agency-lang/lib/runtime/guardTripInterrupt.ts
```diff
@@ -0,0 +1,228 @@
...
+      : `${Math.round(s.spent)}ms elapsed (limit ${s.limit}ms)`;
+  return `${name} exceeded its ${s.dimension} budget: ${unit}. Approve more budget, or reject to stop this work and salvage its draft.`;
+}
+
+/** Best-so-far preview for the handler: the innermost saved draft on the
+ *  branch. A PREVIEW only — the authoritative salvage is computed by the
+ *  unwind (level rule + finalize) if the trip is rejected; finalizes are
+ *  one-shot and must not run speculatively for a maybe-approved trip. */
+function draftPreview(stack: StateStack): unknown {
+  for (let i = stack.stack.length - 1; i >= 0; i--) {
```
egonSchiele: prefer forEach

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -726,7 +727,14 @@ async function _runPrompt({
...
   targetStack.localTokens += completion.usage?.totalTokens ?? 0;
-  targetStack.enforceGuards();
+  // NOTE: no post-charge enforceGuards here anymore. A trip caused by
+  // THIS charge is raised resumably at the caller's next guard gate
+  // (round.N.guardGate / guardGate.final in runPrompt) — the paid work
+  // already happened, and nothing else paid runs before that gate, so
+  // raising there loses nothing and gains the pause point. Throwing
+  // here would make every cost trip non-resumable. The pre-call gate
+  // above still throws as a backstop against spend that slipped in
+  // between a guard gate and this request (e.g. memory-recall charges).
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 6e23d57ee. The memory hooks are skipped while any unsuspended guard is over budget, so no paid supervisory work runs in the window between the crossing charge and the next gate. The probe is a new side-effect-free StateStack.anyGuardOverBudget — deliberately NOT detectTrippedGuard, whose check() consumes TimeGuard's one-shot trip latch. The hooks run again on the next healthy turn.

packages/agency-lang/lib/runtime/guardTripInterrupt.ts
```diff
@@ -0,0 +1,228 @@
...
+  // Resume path FIRST: an answered question must never re-ask. The key
+  // is derived from guard state (see guardTripKey), so a replay of THIS
+  // trip recomputes the same key, while the NEXT trip — possible only
+  // after an approve changed the limit or a disarm — gets a fresh one.
+  const persistedId = stack.other[key] as string | undefined;
+  if (persistedId !== undefined) {
+    const recorded = ctx.getInterruptResponse(persistedId);
+    if (recorded) {
+      delete stack.other[key];
+      applyVerdict(recorded, scope, tripped, err);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 6e23d57ee. A persisted-but-unanswered question now re-surfaces under the SAME interrupt id instead of falling through to a fresh chain run — a fresh id would orphan the pending answer and ask the user twice. (In-process siblings were already covered by the pendingTrip park; this closes the replay-before-answer door, e.g. a resume triggered by answering a different interrupt in the same batch.)

packages/agency-lang/lib/runtime/guard.ts
```diff
@@ -492,6 +693,13 @@ export class TimeGuard implements Guard {
     const remaining = Math.max(1, this.timeLimit - this.currentElapsed());
     const clone = new TimeGuard(remaining, this.label);
     clone.guardId = this.guardId;
+    // Hand-copied: cloning and serialization are DIFFERENT paths, and a
+    // field added to toJSON alone silently misses branches (rev-3 plan
+    // review). scopeIds is how a branch trip finds its scope siblings;
+    // root/disarmed state must follow the budget into the branch.
+    clone.scopeIds = this.scopeIds;
+    clone.isRootBudget = this.isRootBudget;
+    clone.disarmed = this.disarmed;
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 6e23d57ee, one level deeper than the suggestion: startWindow() now refuses to arm while disarmed, which covers every arming path at once — install, resume, AND a disarmed clone being installed on its branch (clone.disarm() alone would only cancel a timer that does not exist yet; the arming happens later, at the branch's first resume). Fake-timer test pins that a disarmed clone never aborts its branch.

packages/agency-lang/lib/runtime/guardScope.ts
egonSchiele: Rewritten declaratively in 6e23d57ee — reversed-copy forEach here, and the same treatment for the two reverse walks in guardTripInterrupt.ts (draftPreview, innermostGuardById).

packages/agency-lang/lib/runtime/guardTripInterrupt.ts
egonSchiele: Moved in 6e23d57ee: lib/templates/runtime/guardTripMessage.mustache, rendered via the compiled template.

packages/agency-lang/lib/runtime/guardTripInterrupt.ts
egonSchiele: Done in 6e23d57ee (both walks — find over a reversed copy).

packages/agency-lang/lib/runtime/call.ts
egonSchiele: The short answer: yes, right now `std::guard` is invisible to the `raises` analysis — and I think the right fix is to make it a documented AMBIENT effect rather than a tracked one, with the real static-safety win landing in #555. Here is the reasoning.

**Where trips actually raise, statically speaking.** In this PR, a trip can only raise at an `llm()` call (the gates live in runPrompt). So for cost trips alone there IS a precise story: give `llm()` the effect `std::guard` in its builtin signature, and `interruptEffectsByFunction` propagates it transitively — every function that can reach an llm call is marked, and `raises` clauses would be checked against it.

**Why that precision collapses in PR 3.** Time trips raise wherever the timer catches the branch: mid-request, but also at any step boundary of plain code and inside tools. After PR 3, the honest static statement is "any function may raise std::guard while any caller has a guard installed" — and guards are dynamic, installed by callers the callee never sees. Propagating that gives every function in every program the same annotation. A `raises` entry that everything carries tells the reader nothing, and the annotation burden is real: every existing `raises`-typed function that calls `llm()` would stop compiling until users add `std::guard` everywhere — a second breaking change that buys no discrimination.

**The precedent.** We already have one control effect in exactly this class: cancellation (Esc, race loss) can stop any function at any step boundary, and no raises clause declares it. `std::guard` post-PR-3 is the same kind of thing — an ambient property of running under someone else's budget, not a property of the function's own code.

**So my recommendation, two parts:**

1. **Declare `std::guard` ambient, explicitly.** Documented in the effects guide next to cancellation: implicitly admitted by every `raises` clause, never required in one, allowed if a user writes it anyway (so the checker should accept it in a clause without demanding it — I believe that is already the fail-open behavior, but it deserves a pin test). The static guarantee of `raises` stays exact for everything user-declared.

2. **Put the real static safety where it can actually bite: typed payloads (#555).** The valuable compile-time story for `std::guard` is not "which functions can trip" (answer: all of them) but "what does a HANDLER see and return": `intr.data` typed as the trip payload (label, scopeIds, dimension, limit, spent, maxCost, maxTime, draftValue) when the handler narrows on the effect, and the approve payload typed as `{maxCost?, maxTime?, disarm?, message?}`. That is exactly the shape the typed-failure-payloads brainstorm (#555) wants — `std::guard` would be the first BUILT-IN typed effect, and match-arm narrowing on `intr.effect` gives handlers checked field access instead of `any`. Effect-string typo detection falls out of the same registry.

If that split sounds right, I can pin the ambient rule (docs + checker test) in this PR or the next, and write `std::guard`'s typed-effect entry into the #555 brainstorm so the payload typing is designed there with a concrete first customer.

Issue #561, opened 2026-07-17
AppleScript injection: imessage.ts interpolates LLM-authored data into osascript source


egonSchiele:
## Summary

`lib/stdlib/imessage.ts` builds its AppleScript by interpolating caller-supplied data (recipient and message body) into a script string, then hand-escaping the result with `escapeAppleScriptString`. Escaping is the wrong layer to defend at: the data is still parsed as code, and correctness depends on the escape function anticipating every AppleScript metacharacter.

This matters more than a normal injection bug because of who supplies the data. `sendIMessage` is designed to be handed to an LLM as a tool, so the `message` argument is model-authored text, which may in turn be influenced by a web page, an email, or a file the agent read. That is an untrusted-input path into a script executed via `osascript`.

## Current shape

```ts
const script = `
  tell application "Messages"
    set targetService to 1st account whose service type = iMessage
    set targetBuddy to participant "${escapeAppleScriptString(to)}" of targetService
    send "${escapeAppleScriptString(message)}" to targetBuddy
  end tell
`
await execFileAsync("osascript", ["-e", script])
```

`escapeAppleScriptString` currently handles backslashes and double quotes (see `lib/stdlib/__tests__/imessage.test.ts`, which asserts exactly those two cases).

## Proposed fix

Pass the data as arguments instead of splicing it into the source. The script becomes a constant, and `argv` values are never parsed as code:

```ts
const SCRIPT = `on run argv
  set recipient to item 1 of argv
  set body to item 2 of argv
  tell application "Messages"
    set targetService to 1st account whose service type = iMessage
    set targetBuddy to participant recipient of targetService
    send body to targetBuddy
  end tell
end run`

await execFileAsync("osascript", ["-e", SCRIPT, "-", to, message])
```

This removes the vulnerability class rather than escaping around it, and lets `escapeAppleScriptString` be deleted.

## Scope

- `lib/stdlib/imessage.ts` is the known instance.
- `lib/stdlib/builtins.ts` also shells to `osascript` and should be audited for the same pattern.
- Tests should keep the existing `execFile`-mock approach and add a case asserting that a hostile payload (e.g. a message containing `"; do shell script "..."; "`) lands as inert data rather than executing.

## Context

Found while designing `std::notes/apple`, which will use `osascript` heavily and will adopt the argv approach from the start. Filing separately rather than retrofitting `imessage.ts` inside that work, to keep the change scoped and reviewable.

PR #562, opened 2026-07-17
Add std::markdown.renderForHtml, and revise the std::notes/apple spec

egonSchiele:
Groundwork for an Apple Notes stdlib module (`std::notes/apple`). This PR is the design spec plus the one implementation piece that is genuinely unblocked.

**The Notes module itself is not implemented.** A blocker is open (section 9.2 of the spec) and it cannot be resolved without you. Details at the bottom.

### 1. `std::markdown.renderForHtml`

Apple Notes only accepts HTML on write. Its `body` is the sole writable content field, so every write path needs markdown-to-HTML. `std::markdown` already had a complete parser, an AST, and `renderForCli`. It was missing only an HTML renderer.

It lives in `std::markdown`, beside `renderForCli`, rather than inside the Notes module. Markdown-to-HTML is a general capability that email and report generation want too, and burying it in a Notes module would hide a reusable thing inside a specific one.

```ts
import { parse, renderForHtml } from "std::markdown"

const result = parse("## Findings\n\n- one\n- two")
print(renderForHtml(result.blocks))
// <h2>Findings</h2>
// <ul><li>one</li><li>two</li></ul>
```

**It treats the AST as untrusted**, unlike `renderForCli`. The markdown reaching it is typically model-authored, and the model may have been influenced by a page it read. So:

- Text content is HTML-escaped.
- URLs are restricted to `http`, `https`, `mailto`, `tel`, and relative paths. `javascript:` and `data:` are dropped, and the link renders as plain text.
- Raw HTML in the source (`html-block` and `inline-html`) is **dropped**, not passed through. This is a real behavioural difference from `renderForCli`, which emits it verbatim. Worth a look — it will surprise someone.
- Frontmatter is omitted, matching conventional markdown-to-HTML renderers.

34 unit tests, no mocks. The three security properties were **mutation-tested** rather than assumed: neutering the URL sanitiser fails 4 tests, passing raw HTML through fails 2, removing escaping fails 5.

Note: no test file existed for `lib/stdlib/markdown.ts` before this, so `renderForCli` is still untested. Out of scope here, but worth filing.

### 2. The revised spec

`docs/superpowers/specs/2026-07-16-std-notes-apple-notes-design.md`. Folds in an external review (11 findings) and a live spike against real Notes.app.

Scope is Apple Notes only. Obsidian needs no module, since a vault is markdown files on disk and `std::fs` already covers it. Bear has a real `x-callback-url` API but no usable return path from a CLI, so it would produce a lopsided write-only API and deserves its own design.

## Things the spec got wrong, now corrected

Both were caught by review and verified rather than argued:

- **The `osascript` argv example was off by one.** `-` is not a separator; `osascript` passes it through as `argv` item 1, shifting every real argument. The security claim itself holds — an injection payload passed as argv comes back inert, verified on this machine.
- **`-1712` is not the only permission signal.** `-1743` ("Not authorized") also exists and is unambiguous, so the error mapping gets better than the spec claimed was possible.

## What the spike found

Three of four questions answered, all favourably:

| Question | Answer |
|---|---|
| Does setting `name` set the title? | **Yes.** The folklore is wrong. `createNote` sets it directly, and `renderForHtml` needs no leading `<h1>` — which is why the renderer could be built first. |
| Can `whose` search note bodies? | **Yes.** `searchNotes` survives intact. It should search `plaintext`, not `body`, since `body` is HTML and searching it would match markup. |
| Does `with timeout` bound the consent wait? | **Yes.** ~10s instead of AppleScript's 120s default. |

### And one safety finding nobody had on their list

**A locked note's `body` reads as an empty string rather than erroring.** Append is read-concat-write, so appending to a locked note would do this:

```applescript
set body of n to "" & "the new text"
```

That replaces the note's contents with the appended text. The spec had filed the password-protected guard under error handling as a nicety producing a friendlier message. It is actually the only thing preventing silent destruction of a locked note. The spec now says so, and the guard is re-checked inside the write script rather than only in the pre-flight.

## The open blocker

**Section 9.2.** Reading `name of container` errored on a locked note. My probe was designed badly — it never tested the same read on an *unlocked* note, so two explanations both fit:

1. Locked notes hide their container. Narrow, affects only a path that already fails closed.
2. `name of container of note X of folder Y` doesn't work on any note. If so, the pre-flight lookup is broken, and with it the folder assertion, the `folder` field in every effect payload, and the account resolution. That is most of the scoping story.

Every unimplemented part of the module depends on that lookup, which is why none of it is here.

One read-only command settles it:

```bash
osascript -e 'tell application "Notes"
  return name of container of note "TITLE_FROM_NAME" of folder "Agency Spike"
end tell'
```

`Agency Spike` → explanation 1, narrow fix. `-1728` → explanation 2, and section 5.3 needs rework before the plan.

It has to run from a terminal you have granted Notes automation to. I get `-1743` regardless of the answer, because TCC grants are per calling application — which is itself the finding above.

(That command needs the `Agency Spike` folder. If you already ran `phase4` and deleted it, any unlocked note of yours works — substitute a real title and folder.)

## Verification

- `make` — clean
- `npx tsc --noEmit` — clean
- 34/34 new tests pass, and fail under mutation
- Verified end-to-end through the real Agency pipeline, not just via tests

## Related

- #561 — AppleScript injection in `imessage.ts`, filed from this work. `renderForHtml` and the Notes module both adopt the argv approach from the start.

### DIFF
--- a/packages/agency-lang/lib/stdlib/__tests__/markdown.test.ts
+++ b/packages/agency-lang/lib/stdlib/__tests__/markdown.test.ts
@@ -1,0 +1,208 @@
+import { describe, it, expect } from "vitest";
+import { _parseMarkdown, _renderMarkdownForHtml } from "../markdown.js";
+
+/** Parse Markdown and render it to HTML. Both halves are pure, so these stay
+ *  unit tests with no mocks and no I/O. */
+function md2html(src: string): string {
+  const parsed = _parseMarkdown(src);
+  expect(parsed.success).toBe(true);
+  return _renderMarkdownForHtml(parsed.blocks);
+}
+
+describe("_renderMarkdownForHtml", () => {
+  describe("headings", () => {
+    it("renders each level", () => {
+      expect(md2html("# One")).toBe("<h1>One</h1>");
+      expect(md2html("## Two")).toBe("<h2>Two</h2>");
+      expect(md2html("###### Six")).toBe("<h6>Six</h6>");
+    });
+  });
+
+  describe("inline formatting", () => {
+    it("renders a paragraph", () => {
+      expect(md2html("hello")).toBe("<p>hello</p>");
+    });
+
+    it("renders bold and italic", () => {
+      expect(md2html("**b**")).toBe("<p><strong>b</strong></p>");
+      expect(md2html("*i*")).toBe("<p><em>i</em></p>");
+    });
+
+    it("renders strikethrough", () => {
+      expect(md2html("~~gone~~")).toBe("<p><del>gone</del></p>");
+    });
+
+    it("renders inline code", () => {
+      expect(md2html("`x`")).toBe("<p><code>x</code></p>");
+    });
+
+    it("nests emphasis", () => {
+      expect(md2html("***both***")).toContain("<strong><em>both</em></strong>");
+    });
+  });
+
+  describe("escaping", () => {
+    it("escapes angle brackets in text", () => {
+      expect(md2html("a < b")).toBe("<p>a &lt; b</p>");
+    });
+
+    it("escapes ampersands", () => {
+      expect(md2html("a & b")).toBe("<p>a &amp; b</p>");
+    });
+
+    it("escapes inside inline code", () => {
+      expect(md2html("`<script>`")).toBe("<p><code>&lt;script&gt;</code></p>");
+    });
+
+    it("escapes inside code blocks", () => {
+      const html = md2html("```\n<script>alert(1)</script>\n```");
+      expect(html).toContain("&lt;script&gt;");
+      expect(html).not.toContain("<script>");
+    });
+  });
+
+  describe("raw HTML is dropped, not passed through", () => {
+    // The Markdown reaching this renderer is typically model-authored, so raw
+    // HTML in the source is untrusted input rather than an authoring choice.
+    it("drops an html block", () => {
+      const html = md2html("<div onclick='steal()'>hi</div>");
+      expect(html).not.toContain("<div");
+      expect(html).not.toContain("onclick");
+    });
+
+    it("drops a raw script block", () => {
+      const html = md2html("<script>alert(1)</script>");
+      expect(html).not.toContain("<script");
+      expect(html).not.toContain("alert(1)");
+    });
+  });
+
+  describe("URL schemes", () => {
+    it("keeps http and https links", () => {
+      expect(md2html("[x](https://example.com)")).toBe(
+        '<p><a href="https://example.com">x</a></p>',
+      );
+    });
+
+    it("keeps mailto links", () => {
+      expect(md2html("[m](mailto:a@b.com)")).toContain('href="mailto:a@b.com"');
+    });
+
+    it("keeps relative links", () => {
+      expect(md2html("[r](/docs/x.html)")).toContain('href="/docs/x.html"');
+    });
+
+    it("drops a javascript: link but keeps its text", () => {
+      const html = md2html("[click](javascript:alert(1))");
+      expect(html).not.toContain("javascript:");
+      expect(html).not.toContain("<a ");
+      expect(html).toContain("click");
+    });
+
+    it("drops a data: link", () => {
+      const html = md2html("[x](data:text/html;base64,PHNjcmlwdD4=)");
+      expect(html).not.toContain("data:");
+      expect(html).not.toContain("<a ");
+    });
+
+    it("is case-insensitive about the scheme", () => {
+      const html = md2html("[x](JaVaScRiPt:alert(1))");
+      expect(html).not.toContain("<a ");
+    });
+
+    it("drops an unsafe image src but keeps the alt text", () => {
+      const html = md2html("![alt](javascript:alert(1))");
+      expect(html).not.toContain("<img");
+      expect(html).toContain("alt");
+    });
+  });
+
+  describe("code blocks", () => {
+    it("renders without a language", () => {
+      expect(md2html("```\nx = 1\n```")).toBe("<pre><code>x = 1</code></pre>");
+    });
+
+    it("renders with a language class", () => {
+      expect(md2html("```ts\nx = 1\n```")).toBe(
+        '<pre><code class="language-ts">x = 1</code></pre>',
+      );
+    });
+  });
+
+  describe("lists", () => {
+    it("renders an unordered list", () => {
+      expect(md2html("- one\n- two")).toBe("<ul><li>one</li><li>two</li></ul>");
+    });
+
+    it("renders an ordered list", () => {
+      expect(md2html("1. one\n2. two")).toBe("<ol><li>one</li><li>two</li></ol>");
+    });
+
+    it("keeps a non-1 start", () => {
+      expect(md2html("3. three\n4. four")).toContain('<ol start="3">');
+    });
+
+    it("renders task list items as checkboxes", () => {
+      const html = md2html("- [x] done\n- [ ] todo");
+      expect(html).toContain('<input type="checkbox" checked disabled> done');
+      expect(html).toContain('<input type="checkbox" disabled> todo');
+    });
+  });
+
+  describe("other blocks", () => {
+    it("renders a block quote", () => {
+      expect(md2html("> quoted")).toBe("<blockquote><p>quoted</p></blockquote>");
+    });
+
+    it("renders a horizontal rule", () => {
+      expect(md2html("---\n")).toBe("<hr>");
+    });
+
+    it("renders a table with alignment", () => {
+      const html = md2html("| a | b |\n| :-- | --: |\n| 1 | 2 |");
+      expect(html).toContain("<table>");
+      expect(html).toContain("<th style=\"text-align:left\">a</th>");
+      expect(html).toContain("<th style=\"text-align:right\">b</th>");
+      expect(html).toContain("<td style=\"text-align:left\">1</td>");
+    });
+
+    it("escapes table cell content", () => {
+      const html = md2html("| a |\n| --- |\n| <b>x</b> |");
+      expect(html).toContain("&lt;b&gt;");
+      expect(html).not.toContain("<b>x</b>");
+    });
+
+    it("omits frontmatter", () => {
+      const html = md2html("---\ntitle: x\n---\n\nbody");
+      expect(html).not.toContain("title");
+      expect(html).toContain("<p>body</p>");
+    });
+  });
+
+  describe("input handling", () => {
+    it("returns empty string for a non-array", () => {
+      expect(_renderMarkdownForHtml(null)).toBe("");
+      expect(_renderMarkdownForHtml("nope")).toBe("");
+      expect(_renderMarkdownForHtml(undefined)).toBe("");
+    });
+
+    it("returns empty string for an empty array", () => {
+      expect(_renderMarkdownForHtml([])).toBe("");
+    });
+
+    it("skips junk nodes rather than throwing", () => {
+      expect(_renderMarkdownForHtml([null, { type: "bogus" }])).toBe("");
+    });
+  });
+
+  describe("documents", () => {
+    it("renders a realistic agent-written note", () => {
+      const html = md2html(
+        "## Findings\n\nThe **key** result.\n\n- one\n- two\n",
+      );
+      expect(html).toContain("<h2>Findings</h2>");
+      expect(html).toContain("<p>The <strong>key</strong> result.</p>");
+      expect(html).toContain("<ul><li>one</li><li>two</li></ul>");
+    });
+  });
+});
--- a/packages/agency-lang/lib/stdlib/__tests__/markdown.test.ts
+++ b/packages/agency-lang/lib/stdlib/__tests__/markdown.test.ts
@@ -88,8 +88,31 @@ describe("_renderMarkdownForHtml", () => {
       expect(md2html("[m](mailto:a@b.com)")).toContain('href="mailto:a@b.com"');
     });
 
-    it("keeps relative links", () => {
+    it("keeps path-relative links", () => {
       expect(md2html("[r](/docs/x.html)")).toContain('href="/docs/x.html"');
+    });
+
+    // A protocol-relative URL has no scheme, so it looks relative to a naive
+    // check, but it is not: a renderer resolves it against the current protocol
+    // and fetches from the remote host. In a note that is a tracking pixel.
+    it("drops a protocol-relative link", () => {
+      const html = md2html("[x](//evil.example.com/x)");
+      expect(html).not.toContain("evil.example.com");
+      expect(html).not.toContain("<a ");
+      expect(html).toContain("x");
+    });
+
+    it("drops a protocol-relative image", () => {
+      const html = md2html("![x](//evil.example.com/track.gif)");
+      expect(html).not.toContain("evil.example.com");
+      expect(html).not.toContain("<img");
+    });
+
+    it("drops a scheme reformed by a control character", () => {
+      // Control chars are stripped before the scheme test, so this must not
+      // slip through as a schemeless "relative" URL and re-form later.
+      const html = md2html("[x](java\nscript:alert(1))");
+      expect(html).not.toContain("<a ");
     });
 
     it("drops a javascript: link but keeps its text", () => {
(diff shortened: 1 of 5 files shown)


packages/agency-lang/lib/stdlib/markdown.ts
```diff
@@ -382,3 +382,233 @@ export function _renderMarkdownForCli(blocks: unknown): string {
...
+// including javascript: and data:, is dropped. Relative URLs have no scheme
+// and are allowed.
+const SAFE_URL_SCHEME_RE = /^(https?|mailto|tel):/i;
+const HAS_SCHEME_RE = /^[a-z][a-z0-9+.-]*:/i;
+
+/** Return a URL safe to place in an href/src, or "" if it is not. */
+function sanitizeHtmlUrl(url: string): string {
+  const trimmed = url.replace(URL_CONTROL_RE, "").trim();
+  if (trimmed.length === 0) return "";
+  if (!HAS_SCHEME_RE.test(trimmed)) return trimmed; // relative
```
egonSchiele: **Protocol-relative URLs bypass the scheme allowlist, and this one is reachable from plain Markdown.**

`//evil.example.com/x` has no scheme, so `HAS_SCHEME_RE` doesn't match and it returns early as "relative". But a protocol-relative URL isn't relative — any renderer resolves it against the current protocol and fetches from `evil.example.com`.

Verified against this branch, straight through `_parseMarkdown` with no hand-built AST:

```
_parseMarkdown("![x](//evil.example.com/track.gif)") |> _renderMarkdownForHtml
// <p><img src="//evil.example.com/track.gif" alt="x"></p>
```

That's the module's stated threat model exactly: model-authored Markdown, where the model may have been influenced by a page it read. In the Apple Notes context this is a tracking pixel in the user's private note — it leaks IP and a read receipt to a third party the moment the note renders, with no user action.

The docstring claims URLs are "restricted to `http`, `https`, `mailto`, `tel`, and relative paths." `//evil.com` is in none of those categories, so today the code is less strict than its own documentation.

Suggested fix — reject protocol-relative before the relative bail-out:

```ts
if (trimmed.startsWith("//")) return "";
if (!HAS_SCHEME_RE.test(trimmed)) return trimmed; // relative
```

Worth a test alongside the existing `keeps relative links` case, since `/docs/x.html` passing is exactly what makes `//evil.com` look like it's covered.

Separately, nice catch ordering `URL_CONTROL_RE` stripping *before* the scheme test — that's what stops `java\nscript:` from re-forming past the check. Easy thing to get backwards.

packages/agency-lang/lib/stdlib/markdown.ts
```diff
@@ -382,3 +382,233 @@ export function _renderMarkdownForCli(blocks: unknown): string {
...
+}
+
+function renderTableHtml(node: Node): string {
+  const headers = (node.headers as string[]) ?? [];
+  const rows = (node.rows as string[][]) ?? [];
+  const aligns = (node.alignments as (string | null)[]) ?? [];
+
+  const alignAttr = (i: number): string => {
+    const a = aligns[i];
+    return a ? ` style="text-align:${a}"` : "";
```
egonSchiele: **Attribute injection: this is the one interpolation in the renderer that skips `escapeHtml`.**

Every other node field goes through `escapeHtml`. `alignments` doesn't, and it lands inside a quoted attribute. Verified against this branch:

```ts
_renderMarkdownForHtml([{
  type: "table", headers: ["a"], rows: [["1"]],
  alignments: ['left" onmouseover="alert(1)'],
}])
// <th style="text-align:left" onmouseover="alert(1)">a</th>
```

A live event handler, out of the attribute.

I want to be precise about reachability, because it's *not* the parser: tarsec only ever emits `left`/`right`/`center`/`null`, so plain `parse |> renderForHtml` is safe today. It matters for two reasons:

1. The Agency signature is `renderForHtml(blocks: any[])`, and `walk` is a documented public API in this same module for transforming the AST. `parse -> walk -> renderForHtml` is an intended pipeline, and a `walk` block acting on model-influenced content can set any field to any value.
2. This function's whole premise, stated in its own header comment, is that **the AST is untrusted**. Under that premise "the parser wouldn't do that" isn't the guarantee — the parser isn't the only thing that produces ASTs.

An allowlist is better than escaping here, since the value space is closed:

```ts
const ALIGNMENTS = ["left", "right", "center"];
const alignAttr = (i: number): string => {
  const a = aligns[i];
  return typeof a === "string" && ALIGNMENTS.includes(a)
    ? ` style="text-align:${a}"`
    : "";
};
```

That fails closed on anything unexpected instead of faithfully rendering it.

packages/agency-lang/lib/stdlib/markdown.ts
```diff
@@ -382,3 +382,233 @@ export function _renderMarkdownForCli(blocks: unknown): string {
...
+    }
+
+    case "list": {
+      const ordered = !!node.ordered;
+      const start = (node.start as number) ?? 1;
+      const items = ((node.items as unknown[]) ?? [])
+        .map(renderListItemHtml)
+        .join("");
+      if (!ordered) return `<ul>${items}</ul>`;
+      const startAttr = start !== 1 ? ` start="${start}"` : "";
```
egonSchiele: **Same class as the `alignments` issue, and this one escapes the tag entirely.**

`start` is annotated `as number` but that's a compile-time assertion over `Record<string, unknown>`, not a runtime check, so a non-number field flows straight into the template. Verified against this branch:

```ts
_renderMarkdownForHtml([{
  type: "list", ordered: true,
  start: '2"><script>alert(1)</script><ol x="',
  items: [/* ... */],
}])
// <ol start="2"><script>alert(1)</script><ol x=""><li>hi</li></ol>
```

A complete `<script>` tag, not just an attribute.

Same reachability caveat as the table case — the parser won't produce it, `walk` and hand-built ASTs can. Coercing is enough here, and it also tidies the `!== 1` comparison, which today compares a possibly-non-number to a number:

```ts
const start = Number(node.start ?? 1);
const startAttr = Number.isInteger(start) && start !== 1 ? ` start="${start}"` : "";
```

Worth a quick grep for other `as number` / `as string` interpolations in the new code that don't pass through `escapeHtml` — these two were the ones I found, but I checked by reading, not exhaustively.

packages/agency-lang/lib/stdlib/__tests__/markdown.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    });
+
+    it("renders with a language class", () => {
+      expect(md2html("```ts\nx = 1\n```")).toBe(
+        '<pre><code class="language-ts">x = 1</code></pre>',
+      );
+    });
+  });
+
+  describe("lists", () => {
```
egonSchiele, line 155: The URL suite is the strongest part of the test file — `javascript:`, `data:`, mixed-case, and the image variant are all the right cases, and testing that the *text survives* while the link is dropped is a nice touch most people skip.

Two gaps, both matching findings I left on `markdown.ts`:

- `keeps relative links` passes `/docs/x.html`, which is the safe kind. It doesn't distinguish path-relative from protocol-relative, so `//evil.com/x` currently sails through and no test notices.
- Nothing covers a hand-built or `walk`-transformed AST, which is where the `alignments` and `start` injections live. Given `renderForHtml` takes `any[]` and the header comment says the AST is untrusted, a couple of "malicious AST" cases would encode that contract.

Also worth saying: the mutation testing described in the PR body is the right instinct and I'd rather see it reported than not. It's worth being clear about what it does and doesn't establish, though — it proves the assertions are load-bearing against the code as written, but it can't surface a case the suite never contemplated. All three issues I found survive the full suite green.

egonSchiele (commented):
## Review

Reviewed the code, then ran the renderer on this branch rather than reasoning about it. Three findings, all verified by execution and left as inline comments.

`renderForHtml` is the right call at the right altitude — building it in `std::markdown` beside `renderForCli` instead of burying it in the Notes module is correct, and the AST-reuse argument holds up. The security posture is also the right one: escaping text, restricting URL schemes, and dropping raw HTML are the three things that matter, and the decision to diverge from `renderForCli` on raw HTML is well-argued and well-tested.

The findings are all in the same place — **the gap between "the parser won't produce that" and "the AST is untrusted."** The header comment commits to the second, and in three spots the code only delivers the first.

### Findings

**1. Protocol-relative URLs bypass the scheme allowlist.** The most serious of the three, because it needs no hand-built AST — plain model-authored Markdown reaches it:

```
_parseMarkdown("![x](//evil.example.com/track.gif)") |> _renderMarkdownForHtml
// <p><img src="//evil.example.com/track.gif" alt="x"></p>
```

`//evil.com` has no scheme, so it's returned as "relative", but every renderer resolves it against the current protocol and fetches remotely. In a Notes context that's a tracking pixel in a private note: IP and read receipt to a third party, no user action. The docstring promises `http`/`https`/`mailto`/`tel`/relative, and this is none of them, so the code is currently less strict than its own documentation.

**2. `alignments` interpolates into a `style` attribute without escaping** — the one field in the renderer that skips `escapeHtml`. Produces `<th style="text-align:left" onmouseover="alert(1)">`.

**3. `start` interpolates into an attribute without escaping** — same class, escapes the tag entirely: `<ol start="2"><script>alert(1)</script>`.

For 2 and 3, tarsec won't emit those values, so `parse |> renderForHtml` is safe today. They're still worth fixing: `renderForHtml(blocks: any[])` accepts any array, `walk` is a documented API in this same module for transforming ASTs, and `parse -> walk -> renderForHtml` is an intended pipeline where a block acting on model-influenced content can set any field to anything. Both fixes are one line.

### On the mutation testing

Reporting it is the right instinct, and I'd rather see it than not. Worth being precise about what it establishes: it proves the existing assertions are load-bearing, but it can't surface a case the suite never contemplated. All three findings above survive the suite fully green — the URL tests check `javascript:` and `data:` and stop, and no test builds an AST by hand. Not an argument against the technique, just against reading it as coverage.

### Not blocking

- `renderForCli` remaining untested is called out in the description and is the right scope call. Worth filing so it doesn't evaporate.
- The spec revision is honest about what it got wrong, and Section 2.7 — a locked note's `body` reading as empty, turning `appendToNote` into silent destruction — is a genuinely good catch that reframes the guard from a nicety into data-loss prevention. Re-checking it inside the write script rather than only in the pre-flight is the right response.
- Blocking the Notes module on 9.2 rather than guessing is the right call, and the reasoning for why you can't settle it yourself (TCC grants are per calling application) is convincing.

### Verdict

The three fixes are small and mechanical. Once they're in, I'd merge — the design decisions here are sound and the divergence from `renderForCli` is deliberate and defensible. Worth adding a "malicious AST" test case or two to encode the untrusted-AST contract the header comment already claims.

packages/agency-lang/lib/stdlib/markdown.ts
egonSchiele: Fixed in a5abbcad. Confirmed the finding by execution first — `[x](//evil.example.com/x)` rendered `<a href="//evil.example.com/x">` straight through `_parseMarkdown`, no hand-built AST needed.

Took the suggested fix, rejecting protocol-relative before the relative bail-out:

```ts
if (trimmed.startsWith("//")) return "";
if (!HAS_SCHEME_RE.test(trimmed)) return trimmed; // genuinely relative
```

Also renamed the existing test to `keeps path-relative links`, since `keeps relative links` was the name that made `//evil.com` look covered. Three new cases: protocol-relative link, protocol-relative image, and the `java\nscript:` reform (which passed already, but nothing pinned the control-strip ordering you noted, so a future reorder would have gone unnoticed).

Verified end-to-end through the real Agency pipeline, not just the unit tests:

```
protocol-relative img: [<p>x</p>]
protocol-relative link: [<p>x</p>]
still works: <p><a href="https://example.com">ok</a> and <a href="/docs/x.html">rel</a></p>
```

The docstring was the tell, and I missed it: it promised `http`/`https`/`mailto`/`tel`/relative, and `//evil.com` is in none of those categories.

packages/agency-lang/lib/stdlib/markdown.ts
egonSchiele: Fixed in a5abbcad. Reproduced first — the `onmouseover` lands on both `<th>` and `<td>`, not just the header.

Took the allowlist rather than escaping, for the reason you gave: the value space is closed, so it fails closed on anything unexpected instead of faithfully rendering it.

```ts
const TABLE_ALIGNMENTS = ["left", "right", "center"];
return typeof a === "string" && TABLE_ALIGNMENTS.includes(a)
  ? ` style="text-align:${a}"`
  : "";
```

Your reachability framing is the part I want to acknowledge directly, because it is the actual bug and the three findings share it. The header comment commits to "the AST is untrusted" and the code only delivered "the parser won't produce that". Under the first premise `walk` is exactly the counterexample, and it is a documented API in this same module. There is now a `malicious ASTs` describe block that encodes that contract explicitly.

packages/agency-lang/lib/stdlib/markdown.ts
egonSchiele: Fixed in a5abbcad. Confirmed the full `<script>` tag first.

Took the suggested coercion, which does tidy the `!== 1` comparison as you noted:

```ts
const start = Number(node.start ?? 1);
const startAttr = Number.isInteger(start) && start !== 1 ? ` start="${start}"` : "";
```

I took the grep suggestion and found a fourth: `heading` does `Math.max(1, Math.min(6, (node.level as number) ?? 1))`. Not exploitable — `Math.min` yields a number or `NaN`, so the worst output is `<hNaN>` — but it is the same shape, so it is coerced and clamped now with a test. That is three unescaped attribute interpolations total (`alignments`, `start`, `level`); everything else routes through `escapeHtml`.

packages/agency-lang/lib/stdlib/__tests__/markdown.test.ts
egonSchiele: Both gaps closed in a5abbcad. Tests 34 → 46.

- `keeps relative links` renamed to `keeps path-relative links`, plus cases for a protocol-relative link and image. You were right that the old name was load-bearing in the wrong direction — `/docs/x.html` passing is what made `//evil.com` look covered.
- New `malicious ASTs` block covering `alignments`, `start`, `level`, and non-string fields, with a comment explaining why the parser not emitting them is beside the point given `any[]` and `walk`.

On the mutation testing, your criticism is correct and I overstated it. It proves the assertions are load-bearing against the code as written; it cannot surface a case the suite never contemplated. All five issues survived the suite fully green, which is the direct evidence. I have described it more narrowly in the commit message.

What I did do this round, since the reverse of your point is that a contemplated case still needs proving: I reverted each fix individually and confirmed the specific new test fails, rather than trusting green. Each of the five reverts breaks its test and restores to 46 passing.

packages/agency-lang/lib/stdlib/markdown.ts
```diff
@@ -382,3 +382,233 @@ export function _renderMarkdownForCli(blocks: unknown): string {
...
+  "&": "&amp;",
+  "<": "&lt;",
+  ">": "&gt;",
+  '"': "&quot;",
+  "'": "&#39;",
+};
+
+function escapeHtml(s: string): string {
+  return s.replace(/[&<>"']/g, (c) => HTML_ESCAPES[c]);
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in a5abbcad. Confirmed it throws — `{ type: "inline-text", content: 42 }` gave `TypeError: s.replace is not a function`, and code-block had a second throw site at `(node.content ?? "").replace(...)` before `escapeHtml` was even reached.

`escapeHtml` now takes `unknown` and coerces, which is safe because the escape runs afterwards:

```ts
function escapeHtml(s: unknown): string {
  const str = typeof s === "string" ? s : s == null ? "" : String(s);
  return str.replace(/[&<>"']/g, (c) => HTML_ESCAPES[c]);
}
```

`sanitizeHtmlUrl` had the same problem and now guards on `typeof url !== "string"`. Chose coercion over dropping so a number still renders as text, with a test proving a coerced `toString()` returning `<script>` is escaped rather than emitted raw. You are right that this contradicted the stated contract: the renderer claims to skip junk rather than throw.

packages/agency-lang/lib/stdlib/markdown.ts
```diff
@@ -382,3 +382,233 @@ export function _renderMarkdownForCli(blocks: unknown): string {
...
+      const inner = allInline
+        ? `<p>${renderInlineHtml(children)}</p>`
+        : children.map(renderBlockHtml).join("");
+      return `<blockquote>${inner}</blockquote>`;
+    }
+
+    case "list": {
+      const ordered = !!node.ordered;
+      const start = (node.start as number) ?? 1;
+      const items = ((node.items as unknown[]) ?? [])
```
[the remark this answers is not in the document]
egonSchiele: Fixed in a5abbcad — `Number(node.start ?? 1)` with an `Number.isInteger` guard, so `NaN` and non-integers omit the attribute rather than emitting `start="NaN"`.

Left `ordered` as `!!node.ordered` rather than a strict boolean check. `!!` cannot inject (it can only produce `<ul>` or `<ol>`), and a truthy non-boolean is a reasonable read of intent. Happy to tighten it if you disagree, but it did not seem to carry the same risk as the attribute interpolation.

packages/agency-lang/lib/stdlib/markdown.ts
```diff
@@ -382,3 +382,233 @@ export function _renderMarkdownForCli(blocks: unknown): string {
...
+
+function renderTableHtml(node: Node): string {
+  const headers = (node.headers as string[]) ?? [];
+  const rows = (node.rows as string[][]) ?? [];
+  const aligns = (node.alignments as (string | null)[]) ?? [];
+
+  const alignAttr = (i: number): string => {
+    const a = aligns[i];
+    return a ? ` style="text-align:${a}"` : "";
+  };
```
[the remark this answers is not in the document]
egonSchiele: Fixed in a5abbcad with an allowlist of `left`/`right`/`center`, omitting the style attribute otherwise. Confirmed the break-out first: an alignment of `left" onmouseover="alert(1)` produced a live event handler on both `<th>` and `<td>`.

PR #563, opened 2026-07-17
Time trips raise resumable interrupts + the join rule (resumable guards PR 3)

egonSchiele:
Resumable guards PR 3 of 4 (plan: `docs/superpowers/plans/2026-07-16-resumable-guards.md`, Part 6; as-executed notes in Part 6b). PR 2 made cost trips raise resumable `std::guard` interrupts at the PromptRunner gates. This PR does the same for time budgets, which are harder: time burns between paid actions, so the gates alone cannot catch it.

## What this PR does

**1. The composed abort signal is now derived, not accumulated (Task 3.1).**
`stack.abortSignal` is an accessor pair. The setter stores the base signal (user cancel, race loser); `rebuildAbortSignal()` composes base + every installed guard's `armedSignal()` via `AbortSignal.any`. Push, pop, suspension brackets, and `GuardScope.extend` recompose. The old `previousSignal` save/restore chain is gone — re-arming an approved guard is a rebuild, so approve can actually revive a branch whose signal had fired.

**2. Time trips raise resumable interrupts (Task 3.2).** Three surfaces, matching the plan's taxonomy:

- **Step boundaries (case c):** `Runner.step` and `Runner.hook` probe `stack.firstRaisableTrip()` before `shouldSkip`'s consuming walk. Approve applies and the step runs; reject throws exactly what shouldSkip would have thrown; unanswered checkpoints at that step and halts — replay-safe because the resumed boundary re-detects and applies the recorded answer before the body runs.
- **Inside tools (case b):** the same raise, riding the same in-tool interrupt path as an `input()` inside a tool. On approve the tool completes on resume and its result reaches the thread — never a dangling `tool_use`.
- **Mid-request (case a):** when a trip's abort cancels an in-flight LLM call, `_runPrompt` classifies the cancellation by its `guardTrip` cause and throws a `GuardTripRetry` control error; `requestStepWithTripRetry` catches it, runs a gate step, and on approve re-issues the request from the same thread state. The prompt push is hoisted into its own idempotent `pushPrompt` step so a re-issue never duplicates the user message.

Two load-bearing discoveries along the way (Part 6b, items 1–4):

- **`TimeGuard.check()` is now clock-based.** Busy loops are unbroken microtask chains that starve the setTimeout macrotask, so they escaped time guards entirely (pre-existing bug). check() now reads elapsed working time directly; the timer stays as the eager notifier that cancels in-flight ops.
- **Cost guards decline the step-boundary raise** (`Guard.raisableTripAtStep()`). The first validation sweep failed 43 fixtures because the runner surface re-asked cost questions the gates had already settled — and delivered their rejects at steps outside the owning guard boundary. Cost only ripens at paid actions, and every paid action already sits behind a gate.
- **Guards settle when `_runGuarded` exits** (`stack.settleGuards`). Between block exit and `_popGuard` there is one runner step; a clock crossing its limit there raised a question about work that had already concluded, and a late timer fire could flip a computed success into a failure (an old race, now closed).

**3. The join rule (Task 3.3, decisions 15/16 — "the grant follows the budget").**
`TimeGuard.extendBudget` records its clamped grant in a serialized `grantedMs`; `cloneForBranch` deliberately does not copy it (it means "granted during this branch"). At a batch join, the parent's limit is extended by the grants of the same branches whose time it is charged with — the charged branch in "max" mode, every branch in "sum" mode — before `addElapsed`. So a legitimately approved branch cannot trip the parent at the join, nested forks propagate grants up one join at a time, and grants never cross budgets: only clones sharing the parent's guardId are read.

## Tests

- `tests/agency/guards/trip-time.agency` (6 cases): step-boundary approve/reject, mid-request approve/reject (the deterministic mock gained an abortable `delayMs` — it "generates" for 60s against a 100ms budget and rejects with `signal.reason`), the headline checkpoint round-trip (harness answers a valued approve, resume re-detects and applies), and the tool-body case.
- `tests/agency/guards/trip-join.agency` (2 cases): grant-follows-budget across a join (verified to FAIL with the join extension disabled), and the enclosing guard tripping at its own limit with its own label through an inner grant.
- Unit tests for the derived signal (rebuild on re-arm, base survives push/pop), `grantedMs` accounting/serialization/clone behavior, and the runner raise contract (halt-with-interrupt replaces the old throw).
- Full suite: 8358 unit tests, 70 guards fixtures, 25 handlers fixtures, structural lint — all green locally.

## Notes for review

- The old shouldSkip throw path still exists and still fires for delivered/leaf trips and root budgets — `raisableTripAtStep()` excluding delivered trips is what keeps the two paths from double-asking.
- `docs/dev/interrupts.md` gains the PR 3 section. `docs/site/guide/guards.md` is untouched (owner-owned); the user-facing story changes here (unhandled time trips now surface as interrupts instead of converting at the boundary), so it likely wants a pass after this lands.
- PR 4 (the feedback channel) is next and small; the subprocess child-trip → parent-approve e2e deferred from PR 2 rides with it.

### DIFF
--- a/packages/agency-lang/lib/runtime/prompt.ts
+++ b/packages/agency-lang/lib/runtime/prompt.ts
@@ -1121,9 +1121,11 @@ export async function runPrompt(args: {
     key: string,
     body: () => Promise<void>,
   ): Promise<void> => {
-    for (let attempt = 0; ; attempt++) {
+    let gateAttempt = 0;
+    while (true) {
       if (stateStack.firstRaisableTrip() !== null) {
-        await pr.step(`${key}.retryGate.${attempt}`, guardGate);
+        await pr.step(`${key}.retryGate.${gateAttempt}`, guardGate);
+        gateAttempt += 1;
         continue; // re-probe: each budget is owed its own question
       }
       try {
@@ -1188,18 +1190,17 @@ export async function runPrompt(args: {
         toolCalls = result.toolCalls;
       } finally {
         if (injectedFactsContent !== null) {
-          const all = messages.getMessages();
-          for (let i = all.length - 1; i >= 0; i--) {
-            if (
-              all[i].role === "system" &&
-              all[i].content === injectedFactsContent
-            ) {
-              // removeAt, not setMessages: this edits ONE message out of
-              // the thread, so it must not reset the labels of all the
-              // others.
-              messages.removeAt(i);
-              break;
-            }
+          const injectedIndex = messages
+            .getMessages()
+            .findLastIndex(
+              (m) =>
+                m.role === "system" && m.content === injectedFactsContent,
+            );
+          if (injectedIndex !== -1) {
+            // removeAt, not setMessages: this edits ONE message out of
+            // the thread, so it must not reset the labels of all the
+            // others.
+            messages.removeAt(injectedIndex);
           }
         }
       }
--- a/packages/agency-lang/lib/runtime/prompt.ts
+++ b/packages/agency-lang/lib/runtime/prompt.ts
@@ -30,7 +30,7 @@ import type { PromptConfig } from "./llmClient.js";
 import { setupFunction } from "./node.js";
 // See docs/dev/promptRunner.md for the control-flow abstraction used here.
 import { PromptBailout, PromptRunner } from "./promptRunner.js";
-import { raiseGuardTripsUntilClear } from "./guardTripInterrupt.js";
+import { GuardTripRetry, raiseGuardTripsUntilClear } from "./guardTripInterrupt.js";
 import { failure, isFailure, isSuccess, markDestructiveWork } from "./result.js";
 import type { SourceLocationOpts } from "./state/checkpointStore.js";
 import type { RuntimeContext } from "./state/context.js";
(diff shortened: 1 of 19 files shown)


packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -1102,33 +1168,38 @@ export async function runPrompt(args: {
...
+          stateStack,
+          retryPolicy,
+          callLabel,
+        });
+        messages = result.messages;
+        toolCalls = result.toolCalls;
+      } finally {
+        if (injectedFactsContent !== null) {
+          const all = messages.getMessages();
+          for (let i = all.length - 1; i >= 0; i--) {
```
egonSchiele: prefer forEach or `for (let x in y)` to C-style loops

packages/agency-lang/lib/runtime/runBatch.ts
```diff
@@ -266,15 +280,29 @@ function chargeAndResumeParentTimeGuards(
...
     let acc = 0;
+    let grant = 0;
     for (const bs of branchStacks) {
       for (const bg of bs.guards) {
         if (bg === pg || !(bg instanceof TimeGuard)) continue;
         if (bg.guardId !== pg.guardId) continue;
         const t = bg.snapshotElapsed();
-        acc = mode === "sum" ? acc + t : Math.max(acc, t);
+        if (mode === "sum") {
+          acc += t;
```
egonSchiele: what is `t`? Why not use more descriptive var names?

packages/agency-lang/lib/runtime/state/stateStack.ts
```diff
@@ -598,6 +654,52 @@ export class StateStack {
...
+      [...this.guards]
+        .reverse()
+        .find(
+          (g) =>
+            !g.isRootBudget &&
+            !this.suspendedGuardIds.includes(g.guardId) &&
+            g.raisableTripAtStep(),
+        ) ?? null
+    );
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 1d7b8ff1a with `findLast` — innermost-first with no allocation, and it stays declarative.

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: Fixed in 1d7b8ff1a: the removal scan is now `findLastIndex`, and the trip-retry wrapper above it swapped its `for(;;)` for a `while` with a named `gateAttempt` counter.

packages/agency-lang/lib/runtime/runBatch.ts
egonSchiele: Renamed in 1d7b8ff1a: `t`/`acc`/`grant` are now `branchElapsedMs`/`chargedMs`/`grantedMs`, and `pg`/`bs`/`bg` are `parentGuard`/`branchStack`/`branchGuard`. (`t` and `acc` were pre-existing, but this function is where they stopped being readable.)

PR #566, opened 2026-07-17
The feedback channel: approve({message}) reaches the model (resumable guards PR 4)

egonSchiele:
Resumable guards PR 4 of 4 — the last one (plan: `docs/superpowers/plans/2026-07-16-resumable-guards.md`, Part 7; as-executed notes in Part 7b). `approve({message})` turns a guard from a meter into a review point: the handler grants more budget **and** says how to spend it.

## What this PR does

**The message reaches the model as a user message, right before the work's next request.** Two halves:

- **Queueing** (`applyVerdict` → `StateStack.queueGuardFeedback`): the merged message (effectMerge newline-joins multiple handlers' messages on one trip) queues on the raising branch in `stack.other`, so it is branch-local and survives checkpoints. Each entry carries its label, `guard:<label>` (falling back to the dimension for unlabeled guards). Queued after `scope.extend`, so a defective answer never leaves a message behind.
- **Draining** (`runPrompt`): an idempotent step of its own before each request — `guardFeedback.initial`, `round.N.guardFeedback`, `validation.N.guardFeedback`, and `retryFeedback.N` inside the trip-retry wrapper (a mid-request approve's message belongs in the re-issued request, which never passes the other three). One thread message per approval, oldest first, pushed with its label via `MessageThread.push` (#557). Labels stay observability-only — the model just sees a user message, which is the point: guard feedback steers exactly like a user would.

A message queued outside any LLM loop (a step-boundary approve in plain code) waits, serialized, and lands at the branch's next `llm()` call. Feedback queued in a fork branch that makes no further request dies with the branch at the join — same lifetime as the branch's reply-attachment queue, documented.

**The deferred subprocess e2e (PR 2 note 7, PR 3 note 10) lands here** and needed zero new IPC plumbing: the approve payload rides the interrupt IPC verbatim, so a parent's `approve({message})` for a child's forwarded trip queues and injects inside the child.

## Deviation from the plan worth knowing (Part 7b)

The plan said "the exact reply-attachments pattern". What shipped shares the shape (branch-local collect, drain in an idempotent pre-request step) but not the plumbing: reply attachments harvest per-tool into `self.runnerState` because they belong to one `llm()` call, while guard feedback belongs to the **branch** — it must survive from a step-boundary approve in plain code to a later `llm()` call.

## Tests

- `tests/agency-js/guard-feedback`: pins placement and order from statelog request payloads — a step-boundary approve's message lands before the branch's next request; two trips answered at one gate (nested guards) inject as two messages, innermost first; all injected messages are user-role and wear `guard:<label>`; no `label` key ever reaches the provider config.
- `tests/agency/guards/trip-feedback`: the message crosses a real checkpoint — a harness `resolve` carries it, the resumed run queues it from the recorded answer, and the fixture reads the thread back to pin position and role.
- `tests/agency/subprocess/trip-forward-approve-message`: the child-trip e2e, asserted from the child's own thread.
- Full validation: 8404 unit tests, 71/71 guards, 25/25 handlers, 62/62 subprocess (one environmental flake — `eval-run-basic` OOM'd once under sweep load and left a stale `/tmp` run dir; passes clean in isolation), lint green.

## Docs

`docs/dev/interrupts.md` gains the PR 4 section; the guards guide gains "Sending feedback with the grant"; the plan gains Part 7b.

This completes the resumable-guards plan. Next up, per earlier discussion: the guard-keyword idea (`docs/superpowers/specs/2026-07-16-guard-keyword-idea.md`) through its own brainstorm.

### DIFF
--- a/packages/agency-lang/lib/runtime/guardTripInterrupt.ts
+++ b/packages/agency-lang/lib/runtime/guardTripInterrupt.ts
@@ -93,7 +93,7 @@ async function raiseOneTrip(
     const recorded = ctx.getInterruptResponse(persistedId);
     if (recorded) {
       delete stack.other[key];
-      applyVerdict(recorded, scope, tripped, err);
+      applyVerdict(recorded, scope, tripped, err, stack);
       return undefined;
     }
     // The question is already OPEN (persisted, no answer yet — e.g. a
@@ -143,7 +143,7 @@ async function raiseOneTrip(
         },
       );
       if (isApproved(verdict) || isRejected(verdict)) {
-        applyVerdict(verdict, scope, tripped, err);
+        applyVerdict(verdict, scope, tripped, err, stack);
         return undefined;
       }
       // Unanswered: persist the id so the resumed gate finds the answer,
@@ -172,18 +172,29 @@ function applyVerdict(
   scope: GuardScope,
   tripped: Guard,
   err: GuardExceededError,
+  stack: StateStack,
 ): void {
   if (isApproved(verdict)) {
+    const payload = ((verdict as { value?: unknown }).value ?? {}) as Parameters<
+      GuardScope["extend"]
+    >[0];
     // GuardScope.extend enforces the whole answer contract: additive
     // grants, negative clamps, disarm, root refusal, and the useless-
     // approval error (leaves the tripped dimension over budget and
     // armed → would re-trip forever).
-    scope.extend(
-      ((verdict as { value?: unknown }).value ?? {}) as Parameters<
-        GuardScope["extend"]
-      >[0],
-      tripped.dimension,
-    );
+    scope.extend(payload, tripped.dimension);
+    // approve({message}) is the feedback channel (PR 4): the merged
+    // message (effectMerge newline-joins multiple handlers') queues on
+    // the raising branch and rides into the thread as a labeled
+    // user-role message before the branch's next model request. Queued
+    // AFTER extend so a defective answer (GuardApproveError) never
+    // leaves its message behind.
+    if (payload.message !== undefined && payload.message !== "") {
+      stack.queueGuardFeedback(
+        payload.message,
+        `guard:${tripped.label ?? tripped.dimension}`,
+      );
+    }
     return;
   }
   // Rejected: the trip stands. Deliver the ORIGINAL error; from here the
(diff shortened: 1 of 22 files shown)


packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -1105,6 +1105,22 @@ export async function runPrompt(args: {
...
+  // user-role messages, in an idempotent step of its own right before
+  // a request step. One thread message per queued approval, oldest
+  // first. The `guard:<label>` labels are observability-only (#557) —
+  // the model sees ordinary user messages, which is the point: guard
+  // feedback steers the model exactly like a user would.
+  const drainGuardFeedback = async () => {
+    const feedback = stateStack.takeGuardFeedback();
+    if (feedback.length === 0) return;
+    feedback.forEach((f) =>
+      messages.push(smoltalk.userMessage(f.text), f.label),
```
egonSchiele: Note that some APIs like Anthropics require strict alternation between user messages and assistant messages. Instead of returning all the feedback as individual user messages, I wonder if it would make more sense to join them all over a new line.

egonSchiele: Changed in f9ef3a1b3: everything queued at one drain point now injects as a single newline-joined user message, in ask order, with the label listing each contributing guard (`guard:inner,guard:outer`). The fixture is re-pinned to the joined shape. (The prompt push that follows is still its own user message — same situation as validation feedback and reply attachments today, which lean on the client-side same-role merge.)

packages/agency-lang/lib/runtime/guardTripInterrupt.ts
```diff
@@ -172,18 +172,29 @@ function applyVerdict(
...
+    // the raising branch and rides into the thread as a labeled
+    // user-role message before the branch's next model request. Queued
+    // AFTER extend so a defective answer (GuardApproveError) never
+    // leaves its message behind.
+    if (payload.message !== undefined && payload.message !== "") {
+      stack.queueGuardFeedback(
+        payload.message,
+        `guard:${tripped.label ?? tripped.dimension}`,
+      );
+    }
```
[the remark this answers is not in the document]
egonSchiele: Both fixed in f9ef3a1b3: the queue now requires `typeof payload.message === "string"` (payloads cross from untyped JS handlers and IPC), and the label fallback is a truthy check so an empty-string label reads as unlabeled (`guard:cost`) instead of `guard:`.

PR #573, opened 2026-07-17
Add std::notes/apple: Apple Notes for the Agency stdlib

egonSchiele:
Adds `std::notes/apple`: create, append, read, search, list, and delete Apple Notes, gated by per-operation interrupts and scopable by folder. Implements the design spec (docs/superpowers/specs/2026-07-16-std-notes-apple-notes-design.md) via the reviewed plan (docs/superpowers/plans/2026-07-17-std-notes-apple.md, two review rounds folded in).

**Stacked on #565** (the spec-9.2 resolution + plan). Only the commits after that branch point are new here.

## What this adds

- `lib/stdlib/appleNotes.ts` — the only place `osascript` is invoked. Script construction, error mapping, result parsing. No policy, no gating. The AppleScript sources are typestache templates in `lib/templates/applescript/notes/`.
- `stdlib/notes/apple.agency` — the public surface: types, six per-operation effects, interrupt gates, markdown-to-HTML conversion, and the module doc comment that generates the reference page (linked from the vitepress config).
- `NotesRead` / `NotesWrite` / `Notes` capability sets.

## The safety design, briefly

- **Injection:** titles, bodies, queries, and folder names are model-authored, so data goes to `osascript` as argv and is never interpolated into script source. No `-` separator (it lands as `item 1 of argv` and shifts every real argument). A mutation step verified the injection test fails if this is broken.
- **Locked notes:** a locked note's body reads as an *empty string*, not an error, so an unguarded append (read-concat-write) would wipe the note. The locked check runs before the interrupt (in the shared Agency-layer preflight helper, so no human approves a doomed call), again in the TS pre-flight after approval, and once more inside the write script. Mutation-verified.
- **Honest payloads:** `appendToNote` / `readNote` / `deleteNote` look the note up (ungated pre-flight — an id is unguessable) before raising their interrupt, so the approver sees the note's real title/folder/account rather than an opaque id, and a `{"folder": "Work"}` policy can match a call that passed only an id.
- **Folder confinement:** the optional `folder` constrains rather than addresses. On both read and write paths the access addresses the note *through* the folder (`note id X of folder Y`), so a note that moved folders during the approval fails to resolve — the check cannot drift from the access. With partial application this confines an agent to one folder in a way the model cannot see.
- **Empty-string payload fields:** an omitted `folder` reaches the payload as `""`, which matches no folder-glob rule (not even `{"folder": "*"}`). Two caveats the tests pin, both surfaced by review: a rule with **no `match` key** still approves an empty folder (the empty string protects against folder-glob rules specifically, not every broad approve rule), and a `null` payload field would fail **open** through the real matcher (`String(null)` is `"null"`, which `"*"` matches) — which is why payloads never carry null. `notesPolicy.test.ts` asserts all of this through `checkPolicyExplicit`, the real evaluator, not a parallel picomatch call.

## v1 limits (recorded in the spec, §10)

- Nested folder paths are unsupported; `listFolders` returns top-level folders only (a bare `folders` enumeration flattens the hierarchy, which would report nested folders as top-level peers).
- The `account` argument is not implemented. `account` is an output field, populated in payloads only on the three pre-flighted calls.

## Testing

Unit tests only, per the owner decision — a test that reached real Notes.app could destroy real notes locally. `execFile` is mocked at module level; nothing shells out; no `tests/agency/` tests exist for this module. 49 appleNotes tests + 5 policy tests through the real evaluator. Both safety guards (argv contract, locked-note guard) were mutation-verified: break the code, watch the specific test fail, restore.

Verified locally: `make`, `tsc --noEmit`, `lint:structure`, and the appleNotes/notesPolicy/markdown suites (100 tests) all pass. Compiling and running an `.agency` file that partially applies `createNote` prints `partial applied ok` without touching Notes.app.


packages/agency-lang/stdlib/notes/apple.agency
```diff
@@ -0,0 +1,271 @@
...
+    this one exists for payload truth and a fail-fast error. */
+def preflight(id: string, folder: string | null): Result<Preflight> {
+  const meta = try _preflightNote(id)
+  if (isFailure(meta)) {
+    return meta
+  }
+  if (folder != null && meta.value.folder != folder) {
+    return failure("Note ${meta.value.title} is in folder ${meta.value.folder}, not ${folder}. Refusing.")
+  }
+  return meta
```
egonSchiele, line 202: **The locked-note check does not run before the interrupt, for any of the three pre-flighted operations.** This contradicts the design the plan and the code comments describe, though it is not a data-loss bug.

This `preflight()` helper reads `locked` (it is in the `Preflight` type and `_preflightNote` returns it) but never acts on it. It asserts only the folder. So for a locked note, `appendToNote` runs:

1. `preflight()` — passes, because a locked note has a real title/folder and no folder mismatch
2. `interrupt std::notes::append(...)` — **raised, and a human approves it**
3. `destructive { _appendToNote(...) }` — `_appendToNote` calls `_preflightNote` again and *then* `assertNotLocked` throws

So the human (or an auto-approve policy) approves an append that is structurally guaranteed to fail. `readNote` and `deleteNote` have the same shape — all three call this helper, none check locked before their interrupt.

The reason this matters beyond a wasted prompt: `appleNotes.ts:425` documents the intended design as *"once in the pre-flight so the error is raised before the interrupt, and once inside the write script."* But `_appendToNote`'s own pre-flight runs inside the `destructive` block, which is *after* the interrupt. So both of its locked checks are post-approval, and the only pre-interrupt code path — this helper — skips the check. The "error before the interrupt" property the plan's Task 4 called for is not delivered anywhere.

To be clear on severity: **the write still fails closed.** No locked note is ever mutated — `_appendToNote` and the in-script `if (password protected of n)` both stop it. This is a design-fidelity and UX finding, not a corruption risk.

The fix is one place and covers all three operations, because they share this helper:

```ts
if (meta.value.locked) {
  return failure("Note ${meta.value.title} is locked. Unlock it in Notes.app and retry.")
}
return meta
```

That makes the locked error precede every interrupt, matches the plan, and removes the doomed-approval prompt. The comment on this helper ("the locked flag is re-checked after the approval — that re-check is the safety guard; this one exists for payload truth") would then be accurate in the other direction: the guard is here, and the TS layer is the belt-and-braces re-check.

packages/agency-lang/lib/stdlib/appleNotes.ts
```diff
@@ -0,0 +1,473 @@
...
+ *  The locked check happens twice on purpose: once in the pre-flight so the
+ *  error is raised before the interrupt, and once inside the write script
+ *  because a human approval sits between them and the note can be locked in
+ *  that window. Spec section 2.7 explains why a missed check destroys data. */
+export async function _appendToNote(
+  id: string,
+  html: string,
+  folder?: string,
+): Promise<NoteMeta> {
+  const meta = await _preflightNote(id);
```
egonSchiele, line 369: Minor: the write path pre-flights twice. `appendToNote` in the `.agency` layer calls `preflight()` (one `osascript` round-trip) before the interrupt, and then this function calls `_preflightNote` again (a second round-trip) after approval, before the append. Same for `_deleteNote`.

That is defensible — this TS layer is self-contained and shouldn't trust its caller to have pre-flighted — so I'm not asking you to thread the metadata through. Two things worth being aware of rather than fixing:

- It's three `osascript` invocations for one `appendToNote` (preflight, preflight, write), each with a 30s ceiling.
- The two pre-flights can observe *different* states. The authoritative one is this second one, immediately before the write, which is correct — but it means the title/folder/account shown in the interrupt payload (from the first pre-flight) can differ from what actually gets written. For folder that's fine (the scoped write re-asserts it). For the displayed title, an approver could in principle see a stale one. Edge-casey, not worth code — worth a one-line comment noting the payload reflects the pre-approval snapshot.

packages/agency-lang/lib/stdlib/__tests__/notesPolicy.test.ts
```diff
@@ -0,0 +1,29 @@
...
+describe("policy globs against an omitted folder", () => {
+  it("an empty folder does not match a specific glob", () => {
+    expect(picomatch.isMatch("", "Work")).toBe(false);
+  });
+
+  it("an empty folder does not match a catch-all glob", () => {
+    // The load-bearing one. If this ever becomes true, listNotes() with no
+    // folder — the widest-reaching call in the module — starts matching any
+    // {"folder": "*"} approve rule, and the payload design needs rethinking.
+    expect(picomatch.isMatch("", "*")).toBe(false);
```
egonSchiele: This is the right property to pin, and I verified it's faithful for these inputs — but it tests a parallel `picomatch` call, not the code path it's protecting.

The real matcher is `matchesRule` in `lib/runtime/policy.ts:89`: `picomatch.isMatch(stripDotSlash(value), stripDotSlash(pattern))`, and it only reaches picomatch after `matchesRule:86` (`if (value === undefined) return false`). For `""` vs `"*"`, `stripDotSlash` is a no-op, so this test's `picomatch.isMatch("", "*")` does match what the real code computes — the conclusion holds today. But it would keep passing if someone changed `matchesRule` to pass picomatch options, normalize the value, or swap the library. The thing it's guarding lives one function away and isn't exercised.

Stronger: import the real evaluator (or `matchesRule` if you export it) and assert a `std::notes::list` interrupt with `data: { folder: "" }` does not match a rule of `{ match: { folder: "*" } }`. That pins the system, not a reimplementation of one line of it.

One related caveat for the PR description, not this file: "an omitted folder... matches no policy glob (not even `*`)" is true for `{ match: { folder: "*" } }`, but `matchesRule:74` returns `true` for a rule with no `match` key at all. So a bare catch-all `{ "action": "approve" }` still approves `listNotes()` with an empty folder. Worth not implying the empty-folder payload protects the widest call from every broad approve rule — it protects it from folder-glob rules specifically.

egonSchiele (commented):
## Review

Reviewed the diff against its base (`std-notes-92-resolved`), and ran the parts I could: string interpolation, the real policy matcher in `lib/runtime/policy.ts`, and the two plan-review blockers. This is the implementation of the plan I reviewed, and it's a careful one.

**Both plan-review blockers are fixed, correctly and with the reasoning left in place.** `folderCreated: !exists.value` now negates the value rather than the always-truthy Result, and the comment explains why. The five `if...then...else` argument-position uses are hoisted to `const payloadFolder`, which parses. I re-ran an interpolation probe and confirmed `${}` works in plain double-quoted strings, so the `failure("... ${meta.value.title} ...")` messages are fine.

**The read-path TOCTOU I flagged in the plan is now closed.** `_readNote` has scoped and unscoped shapes like the write path, so a folder-asserted read addresses the note *through* the folder and a note that moved during approval fails to resolve instead of being read. That was the one design change I most wanted and it's done well, with the reasoning in the comment.

**The account walk is a genuine improvement over the plan.** The plan assumed a note's account was one `container` hop up; this measured that a nested folder's container is its parent folder, and replaced the single hop with a bounded 10-iteration walk that fails closed rather than reporting a folder name as an account. Same care shows in `listFolders` filtering to `container is account` so nested folders aren't reported as top-level peers. Both are the kind of thing only a real probe surfaces.

### Findings (none are data-loss; all fail closed)

**1. The locked-note check doesn't precede the interrupt** — despite `appleNotes.ts:425` and the plan saying it should. The `.agency` `preflight()` helper reads `locked` but only asserts folder, so a locked note reaches the interrupt, gets approved, and only then fails in the TS layer inside the `destructive` block. All three pre-flighted ops (append/read/delete) share this. The write still never happens, so it's a design-fidelity and UX issue, not corruption — but the fix is one line in the shared helper and covers all three. Inline on `apple.agency`.

**2. The policy-glob test pins a parallel `picomatch` call, not `matchesRule`.** Faithful for `""` vs `"*"` today (I checked — `stripDotSlash` is a no-op there), but it would survive a change to the real matcher one function away. Suggest asserting through the real evaluator with a `data: { folder: "" }` payload. Inline, plus a note that the PR description's "matches no glob, not even `*`" doesn't cover a matchless catch-all rule.

**3. Redundant double pre-flight on the write path** (two `osascript` round-trips before an append). Defensible layering; noted inline mainly because the interrupt payload reflects the first snapshot and the write acts on the second.

### Worth crediting

- The mutation discipline carried over from #562 — the argv-inert and locked-guard tests both have "break it, watch this specific test fail, restore" steps, which is the right answer to "a green suite proves nothing."
- The anti-pattern audit as its own commit (`17ba75688`), with the hoisted `beforeEach`, real names over `p`/`f`, and `rows` hoisted — matches the pre-PR audit the coding standards ask for.
- The "unguessable id" docstring survived from the plan intact, including the explicit warning not to downgrade it to "unreachable without a gate."

### Verdict

Merge after finding 1 — it's one line, it matches the stated design, and right now the code and its own comment disagree about when the locked check runs. Findings 2 and 3 are cleanups that can ride along or follow. The safety core (argv injection, locked-note write guard, scoped through-folder access, empty-string payloads) is sound, and I verified the upstream `renderForHtml` dependency landed with its own fixes back on #562.

The `account`-argument gap remains the one thing for you rather than me: v1 ships folder confinement as advisory on a multi-account machine, which §3.4 argued had to be real. It's disclosed in the PR body and spec §10, so this is a "confirm you still accept it" rather than a finding.

packages/agency-lang/docs/site/guide/apple-notes.md
```diff
@@ -0,0 +1,146 @@
+---
```
egonSchiele: This does not belong in the guide, but in module documentation. Read docs/dev/adding-a-module-to-the-agency-stdlib.md

packages/agency-lang/docs/site/guide/apple-notes.md
```diff
@@ -0,0 +1,146 @@
...
+  print(result)
+}
+```
+
+macOS only. The first call asks for Automation permission, and macOS will show
+a dialog. If nobody answers it, the call fails after 30 seconds.
+
+## Notes are addressed by id
+
+Two notes in the same folder can share a title, so a title is not an address.
```
egonSchiele: Also read docs/dev/general-writing-tips.md. Think about what of this users need to know vs what is dev documentation.

packages/agency-lang/lib/stdlib/appleNotes.ts
```diff
@@ -0,0 +1,473 @@
...
+ *
+ *  A folder's container is its PARENT FOLDER when nested, not the account.
+ *  Measured: `container of folder "2017"` is `Archived`, not `iCloud`. One hop
+ *  up is wrong for any nested folder, and wrong silently — the account field
+ *  would hold a folder name and every {"account": "iCloud"} policy would quietly
+ *  stop matching.
+ *
+ *  Bounded, and fails closed rather than returning a folder as an account.
+ *  Verified to reach iCloud from Archived/2017 in 2 hops. */
+export const ACCOUNT_WALK = `      set a to container of c
```
egonSchiele: use typestache for all of these. see lib/templates/ and https://github.com/egonSchiele/typestache.

packages/agency-lang/stdlib/notes/apple.agency
egonSchiele: Fixed in d954b45fa. The shared preflight helper now refuses locked notes, so the error precedes the interrupt for all three operations and no doomed approval is raised. Verified: the locked check runs before any interrupt on append, read, and delete, and the helper docstring plus the TS-side comment now describe the direction correctly (the pre-interrupt check is the primary one; the TS re-checks after approval are the belt and braces).

packages/agency-lang/lib/stdlib/__tests__/notesPolicy.test.ts
egonSchiele: Rewritten in d954b45fa to go through checkPolicyExplicit, and doing so proved your point concretely: the parallel test was not just fragile, it was wrong about null. matchesRule coerces non-strings with String(value), so a null folder becomes the string "null" — and a "*" glob MATCHES it. Through the real evaluator, null fails OPEN, not with a throw as the plan claimed. The test now pins that, plus your caveat that a rule with no match key still approves an empty folder. The plan's measurement section and the PR description are corrected to match.

packages/agency-lang/lib/stdlib/appleNotes.ts
egonSchiele: Comment added in d954b45fa: the append docstring now says the interrupt payload reflects the pre-approval snapshot and the second pre-flight is the authoritative one for the write. Left the double pre-flight itself as is, per your note.

packages/agency-lang/docs/site/guide/apple-notes.md
egonSchiele: Removed in d954b45fa. The guide page is gone; the user-facing content lives in the module doc comment now, so the generated /stdlib/notes/apple page carries it. Also added the vitepress config entry from the adding-a-module doc, which this module had missed entirely.

packages/agency-lang/docs/site/guide/apple-notes.md
egonSchiele: Folded into the module doc comment in d954b45fa, rewritten per the writing tips: the dev-facing rationale (token costs, picomatch mechanics, spike history) is dropped, and what remains is what a user needs — ids not titles, markdown in / plaintext out and why round trips are lossy, folder confinement, the policy example with the account caveat, locked notes, and recoverable deletes.

packages/agency-lang/lib/stdlib/appleNotes.ts
egonSchiele: Done in d954b45fa: all 18 AppleScript sources (the timeout wrapper, the shared account-walk and note-row snippets, and every operation script) are typestache templates under lib/templates/applescript/notes/, composed via render args in appleNotes.ts. The safety comments stay on the TS constants.

packages/agency-lang/lib/stdlib/appleNotes.ts
```diff
@@ -0,0 +1,473 @@
...
+export async function runNotesScript(script: string, args: string[]): Promise<string> {
+  if (process.platform !== "darwin") {
+    throw new Error("Apple Notes is only available on macOS.");
+  }
+
+  try {
+    // No "-" before args: osascript passes it through as argv item 1 and
+    // shifts every real argument by one.
+    const { stdout } = await execFileAsync("osascript", ["-e", script, ...args]);
+    return stdout.trim();
```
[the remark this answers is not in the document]
egonSchiele: Fixed in d954b45fa: trimEnd, with a comment noting leading whitespace is meaningful in note plaintext, matching keyring.ts.

packages/agency-lang/lib/stdlib/__tests__/appleNotes.test.ts
```diff
@@ -0,0 +1,505 @@
...
+
+  it("surfaces an unrecognised stderr rather than swallowing it", async () => {
+    mockFailure("execution error: something else entirely (-9999)");
+    await expect(runNotesScript("s", [])).rejects.toThrow(/something else entirely/);
+  });
+
+  it("trims stdout", async () => {
+    mockStdout("  value  \n");
+    await expect(runNotesScript("s", [])).resolves.toBe("value");
+  });
```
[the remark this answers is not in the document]
egonSchiele: Updated with the trimEnd change: the test now asserts leading whitespace is preserved and only trailing whitespace is dropped.

packages/agency-lang/lib/stdlib/appleNotes.ts
```diff
@@ -0,0 +1,473 @@
...
+
+export async function _listFolders(): Promise<FolderMeta[]> {
+  const raw = await runNotesScript(LIST_FOLDERS_SCRIPT, []);
+  const rows = raw.split("\n").filter((line) => line.length > 0);
+  return rows.map((line) => {
+    const fields = line.split(FIELD_DELIM);
+    if (fields.length !== 3) {
+      throw new Error("Notes returned an unexpected row while listing folders.");
+    }
+    return { id: fields[0], name: fields[1], noteCount: Number(fields[2]) };
```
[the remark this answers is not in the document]
egonSchiele: Fixed in d954b45fa: _listFolders now throws on a non-finite count instead of reporting NaN, with a test.

egonSchiele:
All findings applied in d954b45fa, with per-thread replies inline. The short version:

- **Finding 1 (the merge gate):** the shared `preflight` helper now refuses locked notes, so the error precedes the interrupt for append, read, and delete, and no doomed approval is raised. Comments on both layers now describe the direction correctly.
- **Finding 2:** the policy test goes through `checkPolicyExplicit` now, and that exposed something the parallel test hid: the real matcher coerces `null` to the string `"null"`, which a `"*"` glob matches. So a null payload field would fail open, not throw as the plan claimed. The test pins that, plus your no-`match`-key caveat; the plan's measurement section and the PR description are corrected.
- **Finding 3:** noted in the append docstring — the payload reflects the pre-approval snapshot; the second pre-flight is authoritative for the write.
- The guide page is gone; the module doc comment carries the user-facing content, and the module is now linked from the vitepress config (which I had missed). All 18 AppleScript sources are typestache templates under `lib/templates/applescript/notes/`. Copilot's two findings (trimEnd, NaN note count) are also fixed with tests.

On the `account` argument: the v1 deferral is the decision recorded in the plan and spec §10 from the plan-review round, so I have left it as scoped. If you want the argument in v1 after all, it slots in behind the existing pre-flight machinery — say the word and I will add it as a follow-up commit here or a separate PR.

Suites after the changes: 49 appleNotes + 5 policy + 46 markdown, all green; make, tsc, and the structural linter pass.

PR #574, opened 2026-07-17
guard becomes a language construct: typed Result<T> + static raises for std::guard

### LINKED ISSUE #571: Guard as a language-level construct (typed Result + precise static raises for std::guard)
Tracking issue for the owner-endorsed direction discussed on #560: make `guard` a language-level construct (its own keyword, AST node, and type) instead of a stdlib function.

Why: precise static analysis. With `guard` as syntax, the checker can attribute `std::guard` interrupts to their source scope, which makes a PRECISE static `raises` for guard trips sound — reversing the earlier ambient-effect recommendation (which was forced by trips becoming raisable at any step boundary in PR 3). It also gives guards a typed `Result<T>` and pairs naturally with #555 (typed payloads).

The idea doc, with the soundness argument (it leans on PR 1's handler-eligibility rule) and file pointers:
`docs/superpowers/specs/2026-07-16-guard-keyword-idea.md`

Agreed sequencing was after resumable-guards PRs 3+4, which are now done (#563 merged, #566 in review). Next step is a brainstorm → spec → plan cycle.

egonSchiele:
Closes #571. `guard` becomes a language-level construct — keyword, AST node, checker typing, effect attribution — lowering to the exact call today's stdlib-function syntax emits. Spec: `docs/superpowers/specs/2026-07-17-guard-keyword-design.md` (with an as-executed section); plan: `docs/superpowers/plans/2026-07-17-guard-keyword.md`. Both went through independent review rounds before execution; all findings folded.

## What changes for users

```ts
// before                                      // after
import { guard } from "std::thread"            //   (no import)
const r = guard(cost: $1) as { ... }           const r = guard(cost: $1) { ... }
```

Migration is the formatter: the legacy `as` still parses (and is ignored), and `agency fmt` strips it. The dead import line fails with the existing unresolved-import diagnostic. **No new diagnostic codes anywhere.** Reservation is position-sensitive: only the exact `guard(...) {` shape is claimed, so `const guard = 5` and calls to a user function named `guard` (without a trailing block) still parse.

Two things the checker now knows:

- **`Result<T>`**: the construct types from the block's returns (`r.value` narrows to `number`, not `any`).
- **`std::guard` is a tracked effect**: a `raises`-annotated function containing a guard (or transitively calling one) must list `std::guard` — enforced by the existing raises machinery, hard-require. Bare unhandled guards in node bodies now emit the existing AG3009 warning (they surface to the user since resumable guards); handler-wrapped guards stay silent. **This is the expected warning churn across the corpus** — every new warning in the sweeps is this, by design (spec decision 8).

## How it works (the one-line architecture)

The construct desugars — in the preprocessor and at the TypeChecker entry — into the byte-identical `functionCall` + `blockArgument` shape the old syntax parsed to, calling `_guard` (the old stdlib def, moved to `index.agency` and put on the auto-import prelude). After the desugar, the compiler cannot tell the construct ever existed: same `__block_N` lifting, same frames and step paths, same trip keys — **zero runtime changes**. Proven two ways: preprocessed-AST diffs against goldens captured before the parser change (byte-identical apart from source positions and the `guard`→`_guard` callee), and a zero-churn `make fixtures` regen.

The effect attribution rests on the eligibility rule from #558 — a handler inside a guard's dynamic extent never adjudicates that guard's trip — with the preserved property stated precisely in the spec (raises bounds what can *escape to an acting handler*; pass-through is deliberate). The effect is seeded at the `_guard` symbol (its trips are raised by TS runtime machinery, invisible to the interrupt-statement walk), and everything else — transitive propagation, `checkFunctionTypeRaises`, the warning walker with its `isInsideHandler` discharge — is existing code.

## Notable execution discoveries (full list in the spec's as-executed section)

- `__guard` (double underscore) hit the builder's direct-TS-call convention and bypassed `__call` dispatch — the impl is `_guard`.
- The desugar must mutate IN PLACE: the compilation unit captures def-node references before the checker constructor runs.
- Pattern lowering couldn't see guard bodies at value positions — a `return match(...)` inside a guarded block compiled to invalid JS. The lowering pass learned the node.
- Draft typing inside guard blocks is parity with the legacy syntax (which never enforced it either) — pinned by test, recorded as a known gap.

## Validation

71/71 guards fixtures, 25/25 handlers, 62/62 subprocess (one known `eval-run-basic` environmental flake, passes in isolation), migrated agency-js tests, 8438 unit tests, 144/144 stdlib typecheck integrations, structural lint, zero fixture-regen churn. 81 corpus files + stdlib migrated; `plannerAgent`'s raises clause gained `std::guard` (the hard-require catching the stdlib first).

## For review attention

- The guide edits (`guards.md`, `ts-helpers.md`, `effects-and-raises.md`) are mechanical-plus-two-new-paragraphs — flagged since those pages are owner-voiced.
- `_guard` appears in the generated stdlib reference (the doc generator documents all exports); its docstring says "internal — use the construct".
- The AG3009 warning renders `guard`, not `_guard`.


packages/agency-lang/docs/site/guide/effects-and-raises.md
```diff
@@ -82,6 +82,22 @@ def f() raises <> {
 | *(omitted)* | Raises anything |
 
 
+## Guards
```
egonSchiele: undo changes to this file

packages/agency-lang/docs/site/guide/guards.md
```diff
@@ -7,9 +7,11 @@ description: Documents the `guard` function for capping the cost and/or compute
 
 Guards let you limit the **cost** and/or **compute time** of a block of work.
 
-```ts
-import { guard } from "std::thread"
+`guard` is part of the language — no import needed. (It used to be a stdlib function: if you have older code, delete the `import { guard } from "std::thread"` line, and run `agency fmt` to drop the old `as` between the arguments and the block. The old `as` form still parses, so nothing breaks in the meantime.)
```
egonSchiele: remove. `no import needed. (It used to be a stdlib` -- users don't care what it used to be like. 

`the typechecker knows two things it could not know before` -- again, so glad you made this change, users will not care. Remove.

packages/agency-lang/docs/site/guide/guards.md
```diff
@@ -199,7 +201,7 @@ A rejected trip normally throws the block's work away. `saveDraft` lets you keep
 
 ```ts
 node main() {
-  const result = guard(time: 5m) as {
+  const result = guard(time: 5m) {
```
egonSchiele, line 200: leave these changes in (removing `as` keyword)

packages/agency-lang/docs/site/stdlib/index.md
```diff
@@ -699,3 +699,51 @@ Register a callback for a lifecycle event. A callback registered inside a
...
+```ts
+_guard(
+  cost: number | null = null,
+  time: number | null = null,
+  label: string | null = null,
+  block: () -> any = null,
+): Result raises <std::guard>
+```
+
+Internal: the lowering target of the `guard(...) { }` construct — use
```
egonSchiele:
```
the lowering target of the `guard(...) { }` construct — use
  the construct, not this function
```

Use your judgment -- do you think users will care about this info? If it is an implementation detail, put it in docs/dev/.

packages/agency-lang/lib/types/guardBlock.ts
```diff
@@ -0,0 +1,21 @@
+import type { AgencyNode, BaseNode, Expression } from "../types.js";
+
+/** The `guard(head) { body }` construct (spec:
+ *  docs/superpowers/specs/2026-07-17-guard-keyword-design.md). The head
+ *  is the parenthesized cost/time/label list; each argument is
+ *  optional. Desugared by the preprocessor into the legacy
+ *  functionCall + blockArgument shape calling `__guard` — see
+ *  lib/preprocessors/guardDesugar.ts — so the builder and the runtime
+ *  never see this node. The typechecker and the interrupt-effect
+ *  analysis DO see it: it types as Result<T> and contributes
```
egonSchiele: Two inaccuracies in this doc comment, both of which will mislead:

1. **`__guard` → `_guard`.** Line 7 says the desugar target is `__guard`; the shipped impl is `_guard` (single underscore — double-underscore hit the builder's direct-TS-call convention, per the PR body). The desugar in `guardDesugar.ts` emits `functionName: "_guard"`.
2. **"The typechecker and the interrupt-effect analysis DO see it" is false.** The `TypeChecker` constructor calls `desugarGuardsInBody(program.nodes)` before any checking (index.ts), so by the time the checker and the effect analysis run, the `guardBlock` is already a `_guard` call — they see the desugared shape, and the `_guard` synthesizer case restores `Result<T>` on top of it. Nothing downstream of the constructor sees this node. Suggest: "The parser and `patternLowering` see this node; the checker desugars it away in its constructor, so from typing onward it is a `_guard` call."

egonSchiele (commented):
## Review: guard becomes a language construct

The execution is faithful and the desugar-first architecture is clean — I traced the equivalence claim and it holds: after the in-place desugar, `scopes` are rebuilt from the mutated bodies, so `checkFunctionTypeRaises` walks a real `_guard(...)` call and looks up `interruptEffectsByFunction["_guard"]`, which the `TS_SIDE_EFFECT_SEEDS` seed populates through the prelude import. The same-file transitive test proves that path. Two folds landed well: the warning-walker gap (my spec finding 2) is solved for free by desugaring before the walk, and the prelude cycle (finding 3) by moving the def into `index.agency`. Nice.

**One finding worth blocking on, because it's the exact property the feature exists to guarantee, and it has no test.**

### Cross-module guard-effect attribution is untested and mechanically suspect

The spec's soundness argument is "every observer up to the user endpoint sees the effect." Within a module that holds. Across an `import` I don't think it does, and nothing pins it either way.

The mechanism, laid out:

- The desugar runs on `program.nodes` (the checker constructor) — the program being checked, not its imported dependencies.
- A function's exported effect footprint is `symbol.interruptEffects`, from `collectDirectInterruptEffects` (symbolTable.ts). That collects `interruptStatement` nodes **plus the seed**, and it runs at `SymbolTable.build` time — *before* any desugar. At that moment a guard-containing body holds a `guardBlock` node, which is neither an `interruptStatement` nor a `_guard` call. So `foo`'s exported `interruptEffects` snapshot is `[]`.
- The in-module hard-require works anyway, because it doesn't read that snapshot for `foo` — it re-walks `foo`'s body (now desugared to `_guard`) via `scopes` and resolves the call through `interruptEffectsByFunction["_guard"]`.
- Cross-module, module M imports `foo` from N. Whether M's checker re-walks N's *desugared* body depends on whether N's body in M's compilation `scopes` was desugared — and the desugar is scoped to `program.nodes`, not imported files. If M sees `foo` only through its `[]` symbol snapshot (or through an un-desugared `guardBlock` body), then `def caller() raises <> { foo() }` in M is **not** required to list `std::guard`, and the effect silently vanishes across the import.

That is the one guarantee the construct was built to provide, failing at exactly the boundary where a reviewer or the user endpoint is the only remaining observer. And the migration that exercised this (std::agents, plannerAgent) is all same-package — plannerAgent being caught proves in-module attribution, not cross-module.

**Please add the test before merge:** module N exports `def research(): number { const r = guard(cost: $1) { return 1 } return 2 }`; module M does `import { research } from "..."` and `def caller(): number raises <> { return research() }`. Assert the hard-require fires on `caller`. `typecheckSource` can't express two files (it writes one `main.agency` and imports nothing) — this needs the multi-file harness. If it passes, the whole soundness story is pinned and this comment is just a missing test. If it fails, `std::guard` is escaping across module boundaries and the fix belongs in this PR — either desugar imported bodies before scope-building, or fold the seed into the exported snapshot when a body contains a `guardBlock`.

The two inline notes are doc-accuracy — cheap, and both are the kind of comment that misleads the next person through this code.

Everything else I checked holds: patternLowering correctly handles guardBlock at value positions (it runs at parse, before the desugar, so it genuinely sees the node); the synthesizer's `_guard` Result<T> case mirrors the `try` precedent; the position-sensitive reservation is strictly less breaking than the spec feared.

packages/agency-lang/lib/preprocessors/typescriptPreprocessor.ts
```diff
@@ -318,6 +319,16 @@ export class TypescriptPreprocessor {
   preprocess(): AgencyProgram {
     this.attachDocComments();
     this.program.nodes = collectTags(this.program.nodes);
+    // Guard constructs lower to the legacy _guard call FIRST: every
+    // later pass (match-expr lowering, parallel desugar, callback
```
egonSchiele: `match`-expr lowering is not a *later* pass — `lowerPatterns` runs at parse time (`parser.ts:278`), before `SymbolTable.build`, well before this preprocess step. That ordering is exactly why `patternLowering.ts` needed its own `guardBlock` case in this PR: it meets the construct *before* this desugar removes it, so a `return match(...)` inside a guard body would otherwise never lower. This comment credits the guard-desugar-first ordering for something that isn't true of match-lowering and omits the real interaction. `parallelDesugar` and callback lifting *are* genuinely later, so the comment is right about those. Suggest naming only the passes that actually run after this line, and noting that match-lowering is handled separately upstream.

packages/agency-lang/lib/types/guardBlock.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  `std::guard` to the containing function's effects. */
+export type GuardBlock = BaseNode & {
```
egonSchiele, line 22: This feels really funny. Why not reuse the existing function arg parsing and types, and in the typechecker show the user an error if they use a param that isn't part of guards? Feels like you're reinventing a lot of stuff that already exists.

Also, does this have LSP support? If I hover over `guard`, would I see the args that the guard block takes?

egonSchiele: Reworked in 998069ed4, and the result is much better. The head now parses through the same `argumentListParser` function calls use, carried verbatim on the node; the desugar forwards it into the `_guard` call, and the EXISTING call diagnostics do all validation — `guard(budget: $1) { }` gets AG6025 "Unknown named argument 'budget' in call to 'guard'" (display name patched), type mismatches get AG6019, exactly what a bad call always got. The hand-rolled head parser, allowlist, and `argOrder` field are deleted; the formatter prints through the shared `renderArgList`. On LSP: hover currently shows nothing for `guard` — same as `thread`/`handle`/every other keyword construct, and same as this PR would have had either way. Wiring keyword-construct hover (guard could show `_guard`'s signature) is a natural small follow-up; want me to file it?

packages/agency-lang/docs/site/guide/effects-and-raises.md
egonSchiele: Reverted to main in 998069ed4.

packages/agency-lang/docs/site/guide/guards.md
egonSchiele: Both paragraphs removed in 998069ed4 — the page now goes straight from the intro to the example, with only the syntax updated.

packages/agency-lang/docs/site/guide/guards.md
egonSchiele: Kept — only the added prose was removed; all `as`-removal and import-removal edits stand.

packages/agency-lang/docs/site/stdlib/index.md
egonSchiele: Agreed it is an implementation detail: the doc generator now skips underscore-prefixed exports (one line in lib/cli/doc.ts), so `_guard` is out of the stdlib reference entirely; its story lives in docs/dev/interrupts.md. The stdlib has no other underscore exports, so nothing else disappears.

packages/agency-lang/lib/symbolTable.ts
```diff
@@ -461,8 +461,24 @@ export function classifySymbols(program: AgencyProgram): FileSymbols {
...
+const TS_SIDE_EFFECT_SEEDS: Record<string, string[]> = {
+  _guard: ["std::guard"],
+};
+
+function collectDirectInterruptEffects(
+  name: string,
+  body: AgencyNode[],
+): InterruptEffect[] {
+  const effects: string[] = [...(TS_SIDE_EFFECT_SEEDS[name] ?? [])];
   for (const { node } of walkNodes(body)) {
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 998069ed4: null-prototype object plus an own-property check on read.

packages/agency-lang/lib/types/guardBlock.ts
```diff
@@ -0,0 +1,21 @@
+import type { AgencyNode, BaseNode, Expression } from "../types.js";
+
+/** The `guard(head) { body }` construct (spec:
+ *  docs/superpowers/specs/2026-07-17-guard-keyword-design.md). The head
+ *  is the parenthesized cost/time/label list; each argument is
+ *  optional. Desugared by the preprocessor into the legacy
+ *  functionCall + blockArgument shape calling `__guard` — see
+ *  lib/preprocessors/guardDesugar.ts — so the builder and the runtime
+ *  never see this node. The typechecker and the interrupt-effect
```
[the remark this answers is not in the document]
egonSchiele: Fixed — the comment was stale from the __guard→_guard rename; the file's doc comment was rewritten wholesale in the same commit (see the reply to the lifecycle thread).

packages/agency-lang/lib/preprocessors/guardDesugar.ts
```diff
@@ -0,0 +1,80 @@
...
+    arguments: args,
+    block: {
+      type: "blockArgument",
+      inline: false,
+      params: [],
+      body: desugarGuardsInBody(g.body),
+    },
+    // The prelude import is what resolves `_guard`, so the call's
+    // symbol scope matches what the legacy imported call carried.
+    scope: "imported",
```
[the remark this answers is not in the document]
egonSchiele: Good catch — this was a real metering bypass. Fixed two ways in 998069ed4: `_guard` is exempt from prunePreludeShadows (a user def named `_guard` now collides with the kept prelude import instead of silently capturing every guard block — fail closed), and the name joined RESERVED_FUNCTION_NAMES so the typechecker reports it as a reserved builtin with a clear message. Test added.

packages/agency-lang/lib/types/guardBlock.ts
egonSchiele: Both fixed in 998069ed4. The doc comment now states the real lifecycle in roughly your words: the parser and patternLowering see the node; the TypeChecker desugars it away in its constructor, so from typing onward it is a `_guard` call, with the synthesizer case restoring Result<T>.

packages/agency-lang/lib/preprocessors/typescriptPreprocessor.ts
egonSchiele: Rewritten in 998069ed4: the comment now names only parallelDesugar and callback lifting as the passes this ordering serves, and explicitly notes match lowering runs at parse time and meets the construct in patternLowering's own guardBlock handling (including the value positions).

PR #576, opened 2026-07-18
feat(serve): public ./serve API — createServeHandler + collectServeMetadata

egonSchiele:
## Summary

Adds a public `./serve` entry point to `agency-lang` so an external host (e.g. a web app) can serve a **compiled** Agency module over HTTP using the *same* engine as `agency serve http`, instead of that engine being reachable only through the CLI.

This is the foundational, self-contained first step ("PR 0a") of a larger hosted-agent-execution effort. It delivers no user-facing feature on its own — it is the reusable, published API surface that later work builds on. The key design goal is **reuse by construction**: a host calls agency-lang's own functions rather than copying the discovery/adapter logic.

## What's in it

- **`createServeHandler(compiledPath, { moduleId, exportedNodeNames, interruptEffectsByName?, version, logger? })`** — composes `loadAndDiscover` + `createHttpHandler` internally and returns the `(method, path, body) => Promise<RouteResult>` dispatcher.
  - Raw interrupt helpers (`hasInterrupts` / `respondToInterrupts`) never cross the package boundary, so the adapter's internal `{ data }` unwrap can't be double-applied by a consumer.
  - `version` (an mtime or content hash) is appended to the internal `import()` URL as a **cache-bust**, so re-serving a re-written module at a stable path loads new code instead of Node's cached module.
  - `logger` is injectable (the adapter requires a non-optional logger for its non-leaking error paths).
- **`collectServeMetadata({ filePath, config })`** — returns `{ moduleId, exportedNodeNames, interruptEffectsByName, errors }`. Extracted from the private `compileForServe`, which now **reuses** it (a de-duplication, not new logic).
- **Adapter refactor** — split `HandlerConfig` (the four fields `createHttpHandler` actually uses) out of `HttpConfig`, and export `RouteResult`. `startHttpServer` is unchanged and still receives a full `HttpConfig`.
- **Public barrel** `lib/serve/public.ts` + a `./serve` subpath in `package.json` `exports` (mirrors `./optimize`). `startHttpServer` is deliberately **not** exported (transport-coupled).

## Testing

- New unit tests: `createServeHandler` (4, including the version cache-bust), `metadata` (3), and the `public` barrel (2). Existing `adapter` tests updated for the `HandlerConfig` narrowing.
- `tsc --noEmit` clean (0 errors); full `make build` succeeds; `import('./dist/lib/serve/public.js')` resolves `createServeHandler` + `collectServeMetadata`; structural linter (`eslint lib/`) clean.
- ⚠️ `lib/serve/http/functionFrame.integration.test.ts` fails **locally in a fresh worktree** with `Cannot find package 'agency-lang/stdlib/index.js'` — a pre-existing self-link/environment issue (verified identical with this branch's changes stashed), **not** caused by this PR. Expected green in CI, which provides the package self-link.

## Scope notes

- No behavior change to `agency serve`; the engine is unchanged and simply made reusable.
- Follow-ups (separate PRs): the host-side dependency bump + compiler rewrite, the serve routes, auto-observability, and UI are out of scope here.

### DIFF
--- a/packages/agency-lang/lib/serve/createServeHandler.test.ts
+++ b/packages/agency-lang/lib/serve/createServeHandler.test.ts
@@ -1,0 +1,96 @@
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import fs from "fs";
+import os from "os";
+import path from "path";
+import { createServeHandler } from "./createServeHandler.js";
+
+// A hand-written stand-in for a compiled Agency module. It exports the same
+// generated symbols discoverExports + the adapter read from a real compile:
+// node functions returning { data }, __<node>NodeParams, __toolRegistry,
+// hasInterrupts, respondToInterrupts. Written as a `.mjs` file (below) so Node
+// always treats it as ESM regardless of directory — a `.js` file under
+// os.tmpdir() would rely on Node's syntax auto-detection, which is
+// version-dependent. This fixture imports nothing, so it needs no node_modules.
+function fixtureSource(version: string): string {
+  return `
+export const __toolRegistry = {};
+export const __mainNodeParams = ["message"];
+export const __needsApprovalNodeParams = [];
+export async function main(message) {
+  return { data: { echo: message, version: ${JSON.stringify(version)} } };
+}
+export async function needsApproval() {
+  return { data: { __interrupts: [{ type: "interrupt", effect: "std::read", message: "ok?" }] } };
+}
+export function hasInterrupts(data) {
+  return !!(data && data.__interrupts);
+}
+export async function respondToInterrupts(interrupts, responses) {
+  return { data: { resumed: true, responses } };
+}
+`;
+}
+
+let tmpDir: string;
+let modulePath: string;
+
+beforeAll(() => {
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "serve-handler-"));
+  modulePath = path.join(tmpDir, "agent.mjs");
+  fs.writeFileSync(modulePath, fixtureSource("v1"), "utf-8");
+});
+
+afterAll(() => {
+  fs.rmSync(tmpDir, { recursive: true, force: true });
+});
+
+function baseOptions() {
+  return {
+    moduleId: "agent",
+    exportedNodeNames: ["main", "needsApproval"],
+    version: "1",
+  };
+}
+
+describe("createServeHandler", () => {
+  it("serves /list with the exported nodes and their params", async () => {
+    const handler = await createServeHandler(modulePath, baseOptions());
+    const res = await handler("GET", "/list", undefined);
+    expect(res.status).toBe(200);
+    const body = res.body as { nodes: { name: string; parameters: string[] }[] };
+    const main = body.nodes.find((n) => n.name === "main");
+    expect(main?.parameters).toEqual(["message"]);
+  });
+
+  it("invokes a node and returns its value", async () => {
+    const handler = await createServeHandler(modulePath, baseOptions());
+    const res = await handler("POST", "/node/main", { message: "hi" });
+    expect(res.status).toBe(200);
+    expect(res.body).toEqual({ success: true, value: { echo: "hi", version: "v1" } });
+  });
+
+  it("returns the interrupts payload, then resumes to the final value", async () => {
+    const handler = await createServeHandler(modulePath, baseOptions());
+    const paused = await handler("POST", "/node/needsApproval", {});
+    const pausedBody = paused.body as { success: boolean; value: { interrupts: unknown } };
+    expect(pausedBody.value.interrupts).toBeTruthy();
+
+    const resumed = await handler("POST", "/resume", {
+      interrupts: [{ type: "interrupt", effect: "std::read", message: "ok?" }],
+      responses: [{ type: "approve" }],
+    });
+    const resumedBody = resumed.body as { success: boolean; value: { resumed: boolean } };
+    expect(resumedBody.value.resumed).toBe(true);
+  });
+
+  it("cache-busts on version change: re-written module loads new code", async () => {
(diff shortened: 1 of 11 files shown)


packages/agency-lang/lib/serve/createServeHandler.ts
```diff
@@ -0,0 +1,80 @@
...
+  moduleId: string;
+  /** Names of the exported nodes to serve. */
+  exportedNodeNames: string[];
+  /** Interrupt effects each function/node may raise, for the /list manifest.
+   *  Optional — omitted entries surface as empty arrays. */
+  interruptEffectsByName?: Record<string, InterruptEffect[]>;
+  /** Cache-bust token (mtime or content hash). Appended to the import URL so a
+   *  re-written module at the same path loads new code instead of Node's
+   *  cached module. */
+  version: string;
```
egonSchiele, line 37: The cache-bust is correct, but it has a flip side worth documenting here: Node's ESM loader retains every distinct module URL in its registry for the process lifetime with no eviction API. So each new `version` (mtime/hash) permanently adds a module — and its module-scoped state — to memory. A long-running host that re-serves many re-uploaded versions will grow unboundedly.

Not a blocker for this PR (the factory is a pure builder and this is inherent to in-process dynamic import), but suggest: (a) add a one-line warning to this doc comment that distinct versions accumulate in the ESM cache, and (b) flag it for PR 0b so the consumer plans a mitigation (process recycling, or the deferred child-process execution model, which would bound it naturally).

packages/agency-lang/lib/serve/createServeHandler.test.ts
```diff
@@ -0,0 +1,96 @@
...
+// A hand-written stand-in for a compiled Agency module. It exports the same
+// generated symbols discoverExports + the adapter read from a real compile:
+// node functions returning { data }, __<node>NodeParams, __toolRegistry,
+// hasInterrupts, respondToInterrupts. Written as a `.mjs` file (below) so Node
+// always treats it as ESM regardless of directory — a `.js` file under
+// os.tmpdir() would rely on Node's syntax auto-detection, which is
+// version-dependent. This fixture imports nothing, so it needs no node_modules.
+function fixtureSource(version: string): string {
+  return `
+export const __toolRegistry = {};
```
egonSchiele: `__toolRegistry` is empty, so `discoverExports` finds no functions and these tests only exercise the **node** path through the factory. The `/function/:name` path — `makeInvoker` picking up `__invokeFunction` — is never run through `createServeHandler`. That's exactly the path the design calls a silent-degradation trap (without `__invokeFunction`, function invokes throw `getRuntimeContext()... outside an Agency execution frame` at call time).

`functionFrame.integration.test.ts` covers the function path via `createHttpHandler` directly, but not via this factory's own composition. Suggest adding one exported function to the fixture (populate `__toolRegistry` + an `__invokeFunction`) and a `POST /function/<name>` assertion, so the factory's function wiring is covered too.

egonSchiele (commented):
## Review — PR 0a public `./serve` API

**Overview.** Publishes a `./serve` subpath exposing `createServeHandler` (composes `loadAndDiscover` + `createHttpHandler` internally, so the raw interrupt helpers never cross the package boundary) and `collectServeMetadata` (extracted from the private `compileForServe`, which now reuses it). Adapter refactor splits `HandlerConfig` out of `HttpConfig` and exports `RouteResult`; `startHttpServer` is deliberately not exported. No behavior change to `agency serve`.

**Strengths.**
- Right altitude: reuses `discoverExports`/`createHttpHandler` rather than copying them; DRYs `compileForServe`; the `HandlerConfig` split is the clean minimal seam, not a hack.
- The `?v=<version>` cache-bust is correct — Node's ESM loader caches by full resolved URL (incl. query), so a new token forces a fresh load. Directly answers the stale-re-upload risk.
- Interrupt helpers passed raw with a comment pinning why (no double `.data` unwrap). Good.
- Verified clean: `fs` is still used elsewhere in `serve.ts` (not orphaned), `InterruptEffect` still backs the `CompileResult` type, and `compile()` doesn't print type diagnostics itself so there's no double-print of errors after the refactor.

**Findings** (none blocking for PR 0a):

1. **Unbounded module-cache growth from the cache-bust (medium — document + carry to PR 0b).** See inline note on `version`. Each distinct token permanently adds a module to Node's ESM registry, which has no eviction API. The consumer that re-serves many re-uploaded versions in one long-lived process will leak. Worth a warning on the `version` doc and a mitigation note for PR 0b (process recycling, or the deferred child-process model).

2. **`createServeHandler` function-invoke path is untested (medium — coverage).** See inline note on the test fixture. `__toolRegistry` is `{}`, so only the node path is exercised through the factory; `/function/:name` + `makeInvoker`/`__invokeFunction` — the very path the design flags as a silent-degradation trap — never runs through `createServeHandler`. Adding one function to the fixture closes it.

3. **Double symbol-table build + type-check at serve compile (low — perf, acknowledged).** `compileForServe` no longer passes its `symbolTable` into `compile()`, so `compile()` rebuilds one and `collectServeMetadata` builds another (plus a second type-check pass). CLI-startup only, not per-request — fine, but noting it since the prior code shared one table.

4. **`hasInterrupts`/`respondToInterrupts` cast from possibly-undefined (low — inherited assumption).** `moduleExports.hasInterrupts as (...)` assumes the compiler always emits these. Consistent with `serveHttp` today (`serve.ts:334`), and the `/resume` path guards `respondToInterrupts`, but the `/node` path calls `hasInterrupts` unguarded. Fine as long as the PR 0 smoke-test gate (compiler always emits them) holds — just flagging the dependency.

**Verdict.** Solid, well-factored, and it addressed the earlier plan-review points (`.mjs` fixture, `app::confirm` re-verified through the type checker, all `port` removals incl. the integration test). Ship after considering #1's doc note and #2's fixture addition; both are cheap.

PR #577, opened 2026-07-18
Give the Agency agent the standard-library docs

egonSchiele:
## What and why

The Agency agent could browse the language guide, CLI, and diagnostics docs, but knew nothing about the 64-module standard library — it had to guess what `std::http`, `std::thread`, etc. were for from filenames alone. This gives the agent a browsable, per-module-summarized view of the stdlib docs.

Two independent gaps are closed:

1. **The generated stdlib docs had no summary.** `agency doc` wrote only `name:` frontmatter, so the `<description>` line the agent's doc tool already renders was always empty.
2. **The agent could not reach the stdlib docs at all.** `docsSkill` only served `guide` / `cli` / `diagnostics`, and nothing staged or wired the stdlib docs.

Design and plan are included in the diff:
- `docs/superpowers/specs/2026-07-17-stdlib-docs-summary-for-agent-design.md`
- `docs/superpowers/plans/2026-07-17-stdlib-docs-summary-for-agent.md`

## What changed

**1. `agency doc` emits a `description` frontmatter field** (`lib/cli/doc.ts`). It is derived from the module's existing `@module` comment — the leading paragraph, whitespace-collapsed, capped at 200 chars — with an optional `@summary <text>` line inside the comment as an explicit override (parsed by the doc command only; no language-syntax change). Fence detection trims leading whitespace since `@module` fences render indented.

**The value is emitted as a double-quoted scalar with `"` and `\` removed** (mirroring what `name` already does). This is the one option valid for *both* consumers:
- the agent reads it through tarsec, whose `stripQuotes` unescapes nothing — so an escaped `\"` would reach the model literally;
- the docs site reads it through VitePress's strict YAML — so a bare value would break on the many colon-space summaries (`std::math`, `std::object`).

**2. A recursive `docsSkill("stdlib")` section** (`makefile`, `lib/stdlib/skills.ts`, `stdlib/skills.agency`). The stdlib docs are staged into the packaged docs and globbed with `**/*.{md,markdown}` — 29 of the 64 modules are nested (`ui/table.md`, `auth/oauth.md`, …) and a non-recursive glob would silently drop them.

**3. Registered in the code / research / explorer subagents** with a one-line prompt note each.

**4. Regenerated `docs/site/stdlib/*.md`** with descriptions, and added `@summary` overrides to the 24 modules whose first paragraph exceeded the cap (so their summaries read cleanly instead of truncating).

## Verification

- `lib/cli/doc.test.ts` — 14 new tests (extraction, `@summary` override + body-strip, indented-fence, cap, the quote/colon round-trip asserting no `\"` survives). 36/36 pass.
- `lib/stdlib/skills.test.ts` — `_docsDir("stdlib")` + a recursive-glob regression (`**/*.{md,markdown}` beats flat and includes `ui/table.md`). 6/6 pass.
- `toolWiring.agency` — the new tool builds and is accepted in each subagent's tool list, no name collision. 4/4 pass.
- Regenerated docs validated: tarsec parses all 64 pages, 63 carry a description (`mcp` has no `@module`, correctly none), **zero** stray backslashes. Every frontmatter value is a double-quoted scalar with no inner `"`/`\` — the sufficient condition for strict-YAML validity.
- Reproduced the exact tool description the subagent's model receives: 64 entries, real summaries, nested locations present.
- Structural linter clean.

Note: compiled `.js` siblings (stdlib and agents) are gitignored build artifacts regenerated by `make`; only sources are committed.

### DIFF
--- a/packages/agency-lang/lib/cli/doc.ts
+++ b/packages/agency-lang/lib/cli/doc.ts
@@ -313,6 +313,11 @@ export function firstParagraph(body: string): string {
   return paragraph.join(" ").replace(/\s+/g, " ").trim();
 }
 
+export function firstSentence(text: string): string {
+  const match = text.match(/^.*?[.!?](?=\s|$)/);
+  return match ? match[0] : text;
+}
+
 export function sanitizeDescription(raw: string): string {
   return raw
     .replace(/["\\]/g, "")
@@ -326,7 +331,7 @@ export function moduleDescription(
 ): string | null {
   if (!comment) return null;
   const { override, body } = extractSummaryOverride(comment.content);
-  const raw = override ?? firstParagraph(body);
+  const raw = override ?? firstSentence(firstParagraph(body));
   if (!raw) return null;
   const value = sanitizeDescription(raw);
   return value === "" ? null : value;
--- a/packages/agency-lang/lib/cli/doc.ts
+++ b/packages/agency-lang/lib/cli/doc.ts
@@ -182,7 +182,7 @@ function generateDocForFile(
 
   if (program.docComment) {
     const { body } = extractSummaryOverride(program.docComment.content);
-    sections.push(body.trim());
+    sections.push(formatDocComment({ ...program.docComment, content: body }));
   }
 
   const typeSection = generateTypeSection(typeAliases, ctx);
@@ -282,8 +282,6 @@ function formatDocComment(comment: AgencyMultiLineComment): string {
   return comment.content.trim();
 }
 
-const DESCRIPTION_CAP = 200;
-
 export function extractSummaryOverride(content: string): {
   override: string | null;
   body: string;
@@ -292,7 +290,7 @@ export function extractSummaryOverride(content: string): {
   const firstIdx = lines.findIndex((l) => l.trim() !== "");
   if (firstIdx === -1) return { override: null, body: content };
   const first = lines[firstIdx].trim();
-  if (first.startsWith("@summary")) {
+  if (/^@summary(\s+|$)/.test(first)) {
     const text = first.slice("@summary".length).trim();
     const rest = lines.slice(0, firstIdx).concat(lines.slice(firstIdx + 1));
     return {
@@ -304,27 +302,23 @@ export function extractSummaryOverride(content: string): {
 }
 
 export function firstParagraph(body: string): string {
-  const out: string[] = [];
-  let started = false;
-  for (const line of body.split("\n")) {
-    const trimmed = line.trim();
-    if (!started) {
-      if (trimmed === "") continue;
-      started = true;
-    }
-    if (trimmed === "" || trimmed.startsWith("```")) break;
-    out.push(trimmed);
-  }
-  return out.join(" ").replace(/\s+/g, " ").trim();
+  const lines = body.split("\n").map((line) => line.trim());
+  const start = lines.findIndex((line) => line !== "");
+  if (start === -1) return "";
+  const afterLead = lines.slice(start);
+  const end = afterLead.findIndex(
+    (line) => line === "" || line.startsWith("```"),
+  );
+  const paragraph = end === -1 ? afterLead : afterLead.slice(0, end);
+  return paragraph.join(" ").replace(/\s+/g, " ").trim();
 }
 
 export function sanitizeDescription(raw: string): string {
-  const cleaned = raw.replace(/["\\]/g, "").replace(/\s+/g, " ").trim();
-  if (cleaned.length <= DESCRIPTION_CAP) return cleaned;
-  const slice = cleaned.slice(0, DESCRIPTION_CAP);
-  const lastSpace = slice.lastIndexOf(" ");
-  const truncated = lastSpace > 0 ? slice.slice(0, lastSpace) : slice;
-  return truncated + "…";
+  return raw
+    .replace(/["\\]/g, "")
+    .replace(/\s+/g, " ")
+    .replace(/^#{1,6}\s+/, "")
+    .trim();
 }
 
 export function moduleDescription(
--- a/packages/agency-lang/lib/cli/doc.ts
+++ b/packages/agency-lang/lib/cli/doc.ts
@@ -167,7 +167,12 @@ function generateDocForFile(
 
   const title = path.basename(filePath).replace(/\.agency$/, "");
   const safeName = title.replace(/["\\\n]/g, "");
-  const frontmatter = `---\nname: "${safeName}"\n---`;
+  const fmLines = [`name: "${safeName}"`];
+  const description = moduleDescription(program.docComment);
+  if (description) {
+    fmLines.push(`description: "${description}"`);
+  }
+  const frontmatter = `---\n${fmLines.join("\n")}\n---`;
   const sections: string[] = [frontmatter, heading(1, title)];
 
   // Page-level "View source" link
@@ -176,7 +181,8 @@ function generateDocForFile(
   }
 
   if (program.docComment) {
-    sections.push(formatDocComment(program.docComment));
+    const { body } = extractSummaryOverride(program.docComment.content);
+    sections.push(body.trim());
   }
 
   const typeSection = generateTypeSection(typeAliases, ctx);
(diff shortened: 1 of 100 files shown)


packages/agency-lang/lib/cli/doc.ts
```diff
@@ -276,6 +282,62 @@ function formatDocComment(comment: AgencyMultiLineComment): string {
...
+      override: text === "" ? null : text,
+      body: rest.join("\n"),
+    };
+  }
+  return { override: null, body: content };
+}
+
+export function firstParagraph(body: string): string {
+  const out: string[] = [];
+  let started = false;
```
egonSchiele: **Imperative helper vs the repo's own `anti-patterns.md`.** This loop with a mutable `let started` flag and manual `continue`/`break` matches the "Imperative code everywhere" entry (and the `started` flag brushes "Order-dependent mutable state"). The same "what" — first non-blank line, then up to the first terminator — reads cleanly with `findIndex` + `slice`, no flag:

```ts
export function firstParagraph(body: string): string {
  const lines = body.split("\n").map((line) => line.trim());
  const start = lines.findIndex((line) => line !== "");
  if (start === -1) return "";
  const afterLead = lines.slice(start);
  const end = afterLead.findIndex((line) => line === "" || line.startsWith("```"));
  const paragraph = end === -1 ? afterLead : afterLead.slice(0, end);
  return paragraph.join(" ").replace(/\s+/g, " ").trim();
}
```

Behavior is identical; it just states the intent declaratively. `extractSummaryOverride` above is already in this shape.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -276,6 +282,62 @@ function formatDocComment(comment: AgencyMultiLineComment): string {
...
+
+export function extractSummaryOverride(content: string): {
+  override: string | null;
+  body: string;
+} {
+  const lines = content.split("\n");
+  const firstIdx = lines.findIndex((l) => l.trim() !== "");
+  if (firstIdx === -1) return { override: null, body: content };
+  const first = lines[firstIdx].trim();
+  if (first.startsWith("@summary")) {
```
egonSchiele: **Unbounded prefix match.** `startsWith("@summary")` also matches `@summaryfoo`, and the following `slice("@summary".length)` would then glue the remainder onto the text (`@summaryfoo bar` → `"foo bar"`). No current module triggers it, but a bounded check is more correct — require a whitespace boundary (or end of line) after the tag, e.g. match `/^@summary(\s+|$)/` before slicing.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -176,7 +181,8 @@ function generateDocForFile(
   }
 
   if (program.docComment) {
-    sections.push(formatDocComment(program.docComment));
+    const { body } = extractSummaryOverride(program.docComment.content);
+    sections.push(body.trim());
```
egonSchiele: Minor: the module body now bypasses `formatDocComment` (which is still `content.trim()`, so output is identical today). If `formatDocComment` ever gains logic, module bodies would silently diverge from every other doc-comment render. Cheap to future-proof by routing the stripped body back through it, e.g. `sections.push(formatDocComment({ ...program.docComment, content: body }))`.

egonSchiele (commented):
## Review — Give the Agency agent the standard-library docs

Overall: **strong PR, ready to merge after a couple of small touch-ups.** It implements the reviewed plan faithfully, and the one design call I'd have gotten wrong — how to encode the `description` value — is handled correctly here.

### What it does
- `agency doc` now emits a `description:` frontmatter field derived from each module's `@module` comment (leading paragraph, whitespace-collapsed, 200-char cap), with an optional in-comment `@summary` override. No language-syntax change.
- Adds a recursive `docsSkill("stdlib")` section (staged in the makefile, `**/*.{md,markdown}` so the 29 nested modules aren't dropped) and wires it into the code / research / explorer subagents.
- Regenerates `docs/site/stdlib/*.md` and adds `@summary` to the ~24 modules whose first paragraph exceeded the cap.

### Strengths
- **The both-parsers encoding is correct.** Emitting the value as a double-quoted scalar with `"`/`\` *removed* (not escaped, not bare) is the one option valid for both consumers: tarsec's `stripQuotes` unescapes nothing (so `\"` would leak to the model), and VitePress's strict js-yaml would choke on a bare colon-space value. Good catch — and the tests assert no `\"` survives.
- **Recursion is a safe no-op for the existing sections** — guide/cli/diagnostics have zero nested docs, so always-passing `recursive: true` doesn't change their output.
- **Test coverage is solid**: extraction/override/fence/cap/round-trip in `doc.test.ts`, the nested-glob regression in `skills.test.ts`, and `toolWiring.agency` gating tool-name uniqueness.

### Suggestions (none blocking)
See the inline comments — the main one is a style nit on `firstParagraph` (imperative loop + mutable flag vs the repo's own anti-patterns guidance), plus a robustness note on the unbounded `@summary` prefix match.

### One test-coverage gap
There's a recursive-glob regression test for stdlib, but nothing asserts `docsSkill("guide")` still lists all 61 pages after the always-recursive switch. It's a verified no-op today; a one-line assert in `skills.test.ts` would lock it in so a future glob change can't silently drop pages from the existing tools.

### Minor observation on the `@summary` sweep
Several overrides restate the body's opening sentence (e.g. `date` is nearly verbatim). That's a reasonable use of the override for over-long paragraphs, but it introduces source that must be kept in sync by hand. Because `@summary` sits directly above the prose it mirrors, drift is at least visually obvious — so this is an observation, not a request.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -276,6 +282,62 @@ function formatDocComment(comment: AgencyMultiLineComment): string {
   return comment.content.trim();
 }
 
+const DESCRIPTION_CAP = 200;
```
egonSchiele: I don't think we should cap

packages/agency-lang/lib/cli/doc.ts
egonSchiele: Done in b87e938 — adopted your declarative version verbatim (findIndex + slice, no mutable flag).

packages/agency-lang/lib/cli/doc.ts
egonSchiele: Fixed — the check is now `/^@summary(\s+|$)/`, so `@summaryfoo` no longer matches and can't glue its remainder onto the text. Added a test for it.

packages/agency-lang/lib/cli/doc.ts
egonSchiele: Done — the module body now routes through `formatDocComment({ ...program.docComment, content: body })`, so it can't silently diverge if that helper ever gains logic.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -167,7 +167,12 @@ function generateDocForFile(
 
   const title = path.basename(filePath).replace(/\.agency$/, "");
   const safeName = title.replace(/["\\\n]/g, "");
-  const frontmatter = `---\nname: "${safeName}"\n---`;
+  const fmLines = [`name: "${safeName}"`];
```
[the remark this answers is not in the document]
egonSchiele: Good catch, but this is pre-existing: both `stdlib/agency.agency` and `stdlib/agents/agency.agency` already generated `name: "agency"` on main — this PR only makes them appear together in one listing via the recursive stdlib section.

For the agent's actual use it stays unambiguous: the model reads by `<location>` (unique — `agency.md` vs `agents/agency.md`), and the two entries carry distinct `<description>` text. Deriving the frontmatter `name` from the relative path would change `name` for all 29 nested pages (plus the existing frontmatter tests and anything in the docs site keyed on it), which is broader than this PR. Leaving it as a follow-up unless you'd like it folded in here.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -276,6 +282,62 @@ function formatDocComment(comment: AgencyMultiLineComment): string {
...
+}
+
+export function sanitizeDescription(raw: string): string {
+  const cleaned = raw.replace(/["\\]/g, "").replace(/\s+/g, " ").trim();
+  if (cleaned.length <= DESCRIPTION_CAP) return cleaned;
+  const slice = cleaned.slice(0, DESCRIPTION_CAP);
+  const lastSpace = slice.lastIndexOf(" ");
+  const truncated = lastSpace > 0 ? slice.slice(0, lastSpace) : slice;
+  return truncated + "…";
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed — `sanitizeDescription` now strips a leading ATX heading marker (`/^#{1,6}\s+/`). The `data/*` connector modules read `Wikidata — the open knowledge graph` instead of `## Wikidata — …`. Added a test.

packages/agency-lang/lib/cli/doc.ts
egonSchiele: Removed the cap. And since the cap was the *only* reason the `@summary` sweep existed — the 40 short modules were never capped, and removing it just lets the 24 long paragraphs through in full — I reverted the whole sweep too. The derived first paragraph is now a complete description for every module, with no hand-maintained duplication to keep in sync. The `@summary` override mechanism (and its tests) stays for genuine one-offs.

egonSchiele:
Thanks — addressed the review in b87e938.

**Applied:**
- **No cap.** Removed the 200-char cap, and reverted the whole `@summary` sweep with it: the cap was the only reason the sweep existed (the short modules were never capped), so with it gone the derived first paragraph is a complete description for every module — no hand-maintained overrides to drift. The `@summary` mechanism + tests stay for genuine one-offs.
- **`firstParagraph`** rewritten declaratively (your version verbatim).
- **`@summary`** now requires a boundary (`/^@summary(\s+|$)/`).
- **Module body** routes through `formatDocComment` so it can't diverge.
- **Heading markers** (Copilot): `sanitizeDescription` strips a leading `##` — the `data/*` modules now read `Wikidata — …`.
- **Coverage** (your gap): added a `skills.test.ts` assertion that recursion is a no-op for the flat sections (guide lists identically under `*` and `**`).

**Pushed back on** the Copilot `name`-collision note (inline): the `name: "agency"` clash is pre-existing on main and the agent disambiguates by `<location>`; a path-based `name` would touch all 29 nested pages, so I left it as a follow-up.

Re-verified after the changes: doc + skills TS suites (47), `toolWiring` 4/4, tarsec parse of all 64 pages (63 descriptions, 0 backslashes, 0 leading-heading), construction invariant clean, lint clean.

Issue #575, opened 2026-07-17
Test framework: injectable clock seam on TimeGuard for deterministic time-guard fixtures


egonSchiele:
## Problem

Time-guard fixtures depend on real elapsed wall-clock time, which made them the dominant CI flake class during the resumable-guards work (#563, #566):

- `spin(...)` constants tuned to produce "roughly N ms" of busy work run at different speeds on CI vs local machines. The 2-minute harness timeout forced re-tuning (`spin(3000000)` to `spin(300000)`).
- Tight time budgets race fixture setup: a 5ms budget tripped at step ENTRY before `saveDraft` had run, so reject-path fixtures needed ~100ms budgets found by trial and error.
- Fixtures that MIGHT trip cannot use harness `interruptHandlers` (entries assert the interrupt occurs), forcing inert in-program reject handlers on 7 fixtures.

Every one of these is the same root cause: `TimeGuard.check()` reads the real clock, so tests can only steer it statistically.

## Proposal: an injectable clock seam on TimeGuard

1. **Runtime seam:** `TimeGuard` (and anything else that reads wall-clock for guard purposes) reads time through a clock hook on the runtime context instead of calling `Date.now()`/`performance.now()` directly. Default: the real clock; zero behavior change.
2. **Harness control:** under the test runner, a fake clock installs automatically, and the fixture's `.test.json` can advance it deterministically — e.g. an `advanceTimeMs` field interleaved with `llmMocks` entries, so "the second llm call happens 10 minutes later" is a statement, not a race:

```json
"llmMocks": [
  { "return": "first" },
  { "advanceTimeMs": 600000 },
  { "return": "never-reached" }
]
```

3. **Result:** a `guard(time: 10m)` fixture trips at exactly the declared point, every run, on any machine. `spin()` tuning, budget-vs-setup races, and maybe-trips disappear from time fixtures.

## Scope notes

- Cost guards already have this property (the deterministic client's synthetic cost is fixed) — this brings time guards to parity.
- The seam should cover `TimeGuard.check()`, `raisableTripAtStep()`/`detectStepRaisableTrip()` time reads, and the per-branch working-time budgets from #549/#550.
- Real-time behavior (per-call `timeout`, `armCallTimeout`) can stay on the real clock; only guard-budget accounting needs the seam for deterministic tests.

Came out of the partials-ergonomics plan review (2026-07-17): that plan avoids time-based fixtures entirely because they cannot currently be written deterministically.

[...]

PR #578, opened 2026-07-18
Partials ergonomics: saveDraft as a tool, finalize as draft

egonSchiele:
Two features from one theme — partials become first-class citizens of the agent loop: the model can save them, and the finalize can read them.

Spec: `docs/superpowers/specs/2026-07-17-partials-ergonomics-design.md` (committed on this branch, with review folds and as-executed notes). Plan: `docs/superpowers/plans/2026-07-17-partials-ergonomics.md`.

## Part A: `finalize as draft`

```ts
def research(topic: string): string {
  const full = expand(draftOutline(topic))
  return full

  finalize as draft {
    if (draft != null) {
      return draft + "\n\n[Stopped early: may be incomplete]"
    }
    return "nothing yet"
  }
}
```

- The binder clause reuses the shared block-argument grammar (`asParser`/`FunctionParameter`) — `FinalizeBlock` now carries `params`, the same field shape as `BlockArgument`. All four head forms parse; the formatter canonicalizes (`finalize as d {`, stray `as` dropped like guard's legacy-as).
- The binder types as `T | null` (explicit annotation wins, handler-param rule; declared return type otherwise; `any` fallback). New diagnostics: AG6037 (binder collides with a local/param — the collision rule is what makes the bare-closure-param codegen sound) and AG6038 (one binder only). The collision check is scope-local (`lookupInFunction`), so a binder named like a module global is fine.
- Runtime is one line: `withFinalize` passes `this.partialValueOrNull()` into the closure — the partial it holds IS the scope's own draft, and the throw-fallback returns the same value by construction. The closure emission moved to a typestache template (`finalizeClosure.mustache`); binder-less output is byte-identical (zero fixture churn).
- **Drive-by fix:** the finalize flow widening produced nested unions for union-typed locals (`uniteTypes` dedupes but does not flatten), which defeated `!= null` narrowing inside finalize bodies. Fixed at the widening site; pre-existing, surfaced by the binder.

## Part B: `saveDraft` as a tool

```ts
const r = guard(cost: $0.50) {
  return llm("Research. Save drafts as you go.", tools: [search, saveDraft])
}
```

- The tool loop recognizes the stdlib `saveDraft` by its `name`+`module` pair and handles it ITSELF, inline, in call-list order, before concurrent dispatch — so "last save wins" holds by construction, and the draft files on the scope that owns the `llm()` call (`setSavedDraft`'s `callerFrame()` from inside runPrompt — the stack is `[…, owner, runPrompt]`, frame-isomorphic to the def path).
- **Architecture: a closed intrinsic-tool registry** (`lib/runtime/intrinsicTools.ts`). An `IntrinsicTool` declares `matches` / `buildDefinition` / `handle`; the loop owns all generic bookkeeping (ordered pass, `pr.step` resume idempotency, `toolExecution` span + statelog events, hooks, result-message push). saveDraft is the first entry; attachment/run-control tools are plausible next. Deliberately not user-extensible — intrinsics touch run state.
- The provider-facing schema declares one required `value` param typed by the enclosing def's DECLARED return type, threaded by codegen as `draftSchema` (emitted only for llm calls with post-prompt args — a bare `llm("...")` can't carry tools; `any` skipped at the type level). String fallback otherwise.
- Validation acks: a missing `value` refuses (`Error: saveDraft requires a "value" argument. Nothing was saved.`); a schema-violating value SAVES ANYWAY with a specific warning — the schema is a best-effort hint keyed to the function's type while the slot is often a guard block, and salvage must not lose work over a possibly-wrong hint.

## Testing

- 10 new fixtures (`tests/agency/guards/finalize-binder-*`, `savedraft-tool-*`), all deterministic cost-guard shapes — no wall-clock anywhere.
- Runtime surface tests (`intrinsicToolSchema.test.ts`) pin what fixtures cannot see: the provider-bound schema uses the threaded `draftSchema` (primitive and object; the silent-fallback failure mode is the exact target), the ack text + `tool_call_id` pairing in the thread, the statelog events, and the caller-frame draft write.
- `withFinalize` unit tests (receives the draft, null case, throw returns `this`); parser/checker/formatter/codegen suites per feature. Full unit suite: 8557 passing.

## Known limitations (decisions, tracked so they stay decisions)

- A `.rename()`d stdlib saveDraft is not recognized; it runs as an ordinary tool whose draft files on the tool branch and is discarded.
- Gemini pairs tool results by name+position, so intrinsic acks pushing before dispatched results can misalign in exotic orderings (same class as the #566 finding; real-provider-only). Tracked in #579.
- The saveDraft schema keys on the enclosing FUNCTION's declared type; a guard block slot of a different type makes it a best-effort hint (spec finding 3). #580 tracks threading a `Result<T>` assignment annotation into block codegen, which fixes the annotated case (and responseFormat inside guard blocks with it).
- `input()` inside a tool is not answerable in the test harness (pre-existing, verified on the unmodified build) — the resume fixture uses the attach-to-reply `writeBinary` pattern instead.
- Time-based partials tests stay out until #575 (injectable TimeGuard clock) lands.

## Doc follow-ups (owner-authored files, not edited here)

- `docs/site/guide/partial-results.md` needs sections for both features (`finalize as draft`, saveDraft-as-a-tool with the schema/validation story).

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder/finalizeCodegen.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder/finalizeCodegen.ts
@@ -5,6 +5,7 @@ import type { AgencyNode } from "../../types.js";
 import type { FinalizeBlock } from "../../types/finalizeBlock.js";
 import type { ScopeManager } from "./scopeManager.js";
 import * as renderFinalizeClosure from "../../templates/backends/typescriptGenerator/finalizeClosure.js";
+import { formatTypeHintTs } from "../../utils/formatType.js";
 
 /** Everything a scope's compilation needs from its (possible) finalize
  *  block. Produced by FinalizeCodegen.compileScope for every function and
@@ -161,12 +162,24 @@ export class FinalizeCodegen {
     // was never declared via let/const, so body references print as the
     // bare identifier — the same mechanism inline handler params use.
     // AG6037 guarantees the name cannot collide with a real local, and
-    // AG6038 guarantees at most one param reaches codegen.
+    // AG6038 guarantees at most one param reaches codegen. The TS
+    // annotation follows the inline-handler-param convention: the
+    // user's explicit annotation when written (plus the null arm the
+    // empty slot implies), `any` otherwise. It is documentation only —
+    // generated TS is transpiled, not type-checked; the enforced
+    // typing is the checker's `T | null` on the Agency side.
     const binder = finalize.params[0];
+    let binderParam = "";
+    if (binder !== undefined) {
+      const tsType = binder.typeHint
+        ? `${formatTypeHintTs(binder.typeHint)} | null`
+        : "any";
+      binderParam = `${binder.name}: ${tsType}`;
+    }
     return ts.raw(
       renderFinalizeClosure
         .default({
-          binderParam: binder !== undefined ? `${binder.name}: any` : "",
+          binderParam,
           frameVar: this.frameVar(),
           moduleId: JSON.stringify(this.moduleId),
           scopeName: JSON.stringify(scopeName + "#finalize"),
--- a/packages/agency-lang/lib/backends/typescriptBuilder/finalizeCodegen.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder/finalizeCodegen.ts
@@ -4,6 +4,7 @@ import type { TsNode } from "../../ir/tsIR.js";
 import type { AgencyNode } from "../../types.js";
 import type { FinalizeBlock } from "../../types/finalizeBlock.js";
 import type { ScopeManager } from "./scopeManager.js";
+import * as renderFinalizeClosure from "../../templates/backends/typescriptGenerator/finalizeClosure.js";
 
 /** Everything a scope's compilation needs from its (possible) finalize
  *  block. Produced by FinalizeCodegen.compileScope for every function and
@@ -156,12 +157,22 @@ export class FinalizeCodegen {
   private closure(finalize: FinalizeBlock, scopeName: string): TsNode {
     const parts = this.compileBody(finalize.body, FinalizeCodegen.STEP_BASE);
     const bodyStr = parts.map((n) => printTs(n, 1)).join("\n");
-    const frameVar = this.frameVar();
+    // The binder is a plain closure parameter (never a frame local): it
+    // was never declared via let/const, so body references print as the
+    // bare identifier — the same mechanism inline handler params use.
+    // AG6037 guarantees the name cannot collide with a real local, and
+    // AG6038 guarantees at most one param reaches codegen.
+    const binder = finalize.params[0];
     return ts.raw(
-      `const __finalize = async (): Promise<any> => {\n` +
-        `  const runner = new Runner(__ctx, ${frameVar}, { state: ${frameVar}, moduleId: ${JSON.stringify(this.moduleId)}, scopeName: ${JSON.stringify(scopeName + "#finalize")} });\n` +
-        bodyStr +
-        `\n  return runner.halted ? runner.haltResult : undefined;\n};`,
+      renderFinalizeClosure
+        .default({
+          binderParam: binder !== undefined ? `${binder.name}: any` : "",
+          frameVar: this.frameVar(),
+          moduleId: JSON.stringify(this.moduleId),
+          scopeName: JSON.stringify(scopeName + "#finalize"),
+          body: bodyStr,
+        })
+        .trimEnd(),
     );
   }
 
--- a/packages/agency-lang/lib/backends/typescriptBuilder/scopeManager.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder/scopeManager.ts
@@ -159,7 +159,12 @@ export class ScopeManager {
    * type, so the enclosing def's declared type is the best-effort hint.
    */
   enclosingDeclaredReturnType(): VariableType | undefined {
-    // Innermost-first: the nearest non-block scope answers.
+    // Innermost-first: the nearest non-block scope answers. The Scope
+    // union names more kinds (imported/static/local), but the builder
+    // only ever PUSHES function, node, and block onto this stack, over
+    // the root global — the same invariant `returnType()`'s throwing
+    // default relies on — so the walk always lands on one of those
+    // three, and the default arm below covers only `global`.
     const owner = [...this.stack]
       .reverse()
       .find((scope) => scope.type !== "block");
--- a/packages/agency-lang/lib/backends/typescriptBuilder/scopeManager.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder/scopeManager.ts
@@ -150,6 +150,37 @@ export class ScopeManager {
     }
   }
 
+  /**
+   * Declared return type of the nearest ENCLOSING function or node,
+   * walking outward past block scopes. Unlike `returnType()`, which
+   * answers for the CURRENT scope (and answers `undefined` for blocks),
+   * this is the saveDraft tool's schema key (partials-ergonomics spec
+   * Part 2): a guard block owns the draft slot but carries no declared
+   * type, so the enclosing def's declared type is the best-effort hint.
+   */
+  enclosingDeclaredReturnType(): VariableType | undefined {
+    // Innermost-first: the nearest non-block scope answers.
+    const owner = [...this.stack]
+      .reverse()
+      .find((scope) => scope.type !== "block");
+    if (owner === undefined) return undefined;
+    switch (owner.type) {
+      case "function":
+        return (
+          this.compilationUnit.functionDefinitions[owner.functionName]
+            ?.returnType ?? undefined
+        );
+      case "node":
+        return (
+          this.compilationUnit.graphNodes.find(
+            (n) => n.nodeName === owner.nodeName,
+          )?.returnType ?? undefined
+        );
+      default:
+        return undefined; // global scope
+    }
+  }
+
   /** Whether the surrounding function/node opted-in to runtime return-type validation. */
   returnTypeValidated(): boolean {
     const scope = this.current();
--- a/packages/agency-lang/lib/runtime/intrinsicTools.ts
+++ b/packages/agency-lang/lib/runtime/intrinsicTools.ts
@@ -34,7 +34,13 @@ export type IntrinsicTool = {
   matches: (fn: AgencyFunction) => boolean;
   /** The provider-facing definition, replacing the def's own. */
   buildDefinition: (ctx: { draftSchema: unknown }) => ToolDefinition;
-  /** Handle one call; the return value is the tool-result text. */
+  /** Handle one call; the return value is the tool-result text.
+   *  Contract note: EVERY call must produce a tool-result message
+   *  (providers reject a tool_use with no result), and for now that
+   *  result is plain text — every intrinsic is a state write plus an
+   *  acknowledgment. If a future intrinsic needs a richer result
+   *  (attachments, structured content), widen this return type then;
+   *  the loop's plumbing is the only consumer. */
   handle: (call: IntrinsicCall) => string;
 };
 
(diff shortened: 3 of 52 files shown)


packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -1516,9 +1570,15 @@ export async function runPrompt(args: {
         // semantics across branches (strategy B in the plan): same-round
         // removal is best-effort and removals always take effect from the
         // NEXT round (the .filter() after this parallel call).
-        const parallelResult = await pr.parallel(
+        // An all-intrinsic round has nothing to dispatch: short-circuit
+        // to an empty values result instead of running runBatch over
+        // zero children.
+        const parallelResult: RunBatchResult<void> =
+          dispatchCalls.length === 0
```
egonSchiele: Nit (readability): the `pr.parallel(...)` arguments keep their pre-change indentation, so they now sit *under-indented* relative to `await pr.parallel(` inside this ternary. It reads awkwardly and prettier would reflow it. Consider hoisting the dispatch result out of the ternary, e.g.

```ts
let parallelResult: RunBatchResult<void> = { kind: "values", values: [] };
if (dispatchCalls.length > 0) {
  parallelResult = await pr.parallel(`round.${round}.tools`, dispatchCalls, ...);
}
```

so the multi-line call keeps its natural indentation.

packages/agency-lang/lib/runtime/saveDraftTool.ts
```diff
@@ -0,0 +1,62 @@
...
+/** The contract the type system cannot check, stated to the model. */
+const DESCRIPTION =
+  "Save your best-so-far answer as a draft. If the budget runs out, the " +
+  "last saved draft is returned instead of a failure. The value must " +
+  "match this function's return type.";
+
+/** Character count for the acknowledgment message. */
+export function draftCharCount(value: unknown): number {
+  if (typeof value === "string") return value.length;
+  const text = JSON.stringify(value);
```
egonSchiele: `JSON.stringify` throws on genuinely unstringifiable values (circular refs, BigInt). It's safe *here* because `value` originates from a model tool-call args object (JSON, so neither is reachable), but `draftCharCount` is exported and its unit test feeds `undefined` — where `JSON.stringify` *returns* `undefined` and hits the `=== undefined` guard, so the throwing case is never exercised. Either narrow the test name to "undefined → 0" or wrap in `try/catch` returning 0 if you want the exported helper to honor the general claim. Trivial.

packages/agency-lang/lib/runtime/intrinsicTools.ts
```diff
@@ -0,0 +1,136 @@
...
+  toolCall: { id: string; name: string; arguments: Record<string, unknown> };
+  stateStack: StateStack;
+  draftSchema: unknown;
+  statelogClient: StatelogClient;
+  ctx: unknown;
+  model: unknown;
+  messages: MessageThread;
+}): Promise<void> {
+  const { intrinsic, toolCall, stateStack, draftSchema, statelogClient } = opts;
+  const callArgs = toolCall.arguments ?? {};
```
egonSchiele, line 103: The dispatched branch opens with `if (ctx.isCancelled(stateStack)) throw new AgencyCancelledError();`, but `runIntrinsicCall` has no equivalent — a saveDraft in a cancelled round still writes the draft and pushes an ack. I think that's actually desirable (saving on the way down is salvage-friendly, and answering avoids a dangling `tool_use`), but the divergence is silent. Worth a one-line comment stating it's intentional, so a future reader doesn't "fix" it by adding the cancel check.

egonSchiele (commented):
## Review: Partials ergonomics (`saveDraft` as a tool, `finalize as draft`)

Nice work — this is a clean, well-encapsulated implementation, and the folded plan-review findings are clearly reflected in the shipped code:

- The **schema seam** that the plan review called the one must-close gap is now pinned by `intrinsicToolSchema.test.ts` (threaded `draftSchema` → provider tool list), not just the two halves in isolation.
- The **resume/ack survival** reasoning is documented at the interception site (ack rides the live thread; next interrupt-path `snapshotMessages()` captures it; a lost ack = dangling `tool_use` that real providers reject).
- **AG6037** is now `lookupInFunction` (scope-local), not parent-walking `has`, with the module-global-binder case explained in the comment.
- The **`partitionIntrinsicCalls`** split (declarative two-list partition) resolves the "two live tool-call lists to audit" concern from the anti-pattern audit, and the intrinsic *verb* now lives in `runIntrinsicCall`/`saveDraftIntrinsic` rather than inlined in the loop.

**Correctness spot-checks that held:**
- `toolFunctions` is initialized to `exposedFunctions` and only ever shrinks via `removedTools` (which an intrinsic never enters, since it never errors), so intrinsic recognition in `partitionIntrinsicCalls` stays consistent with the synthesized tool definition. No dangling-`tool_use` path.
- All-intrinsic rounds: `dispatchCalls.length === 0` short-circuits `pr.parallel`, the `while (toolCalls.length > 0)` loop still advances and re-invokes the model, and `savedraft-tool-order` exercises exactly this path (both calls intrinsic, "second" wins by construction).
- Every tool_call still gets a result even past `maxToolCallRounds` (intrinsics answer unconditionally), so pairing stays intact.

**Acknowledged/deliberate, noting for the record:** `runIntrinsicCall` emits a second, intentionally-divergent copy of the tool lifecycle (one `pr.step` instead of per-phase branch steps, `timeTaken: 0`, span opened for pairing). It's documented as a decision, and unifying with the real dispatch path would be scope creep — fine as-is, just a known drift point if a fifth lifecycle event or a span change lands.

A few small inline nits below — all minor, none blocking.

packages/agency-lang/lib/backends/typescriptBuilder/finalizeCodegen.ts
```diff
@@ -156,12 +157,22 @@ export class FinalizeCodegen {
...
+    // AG6038 guarantees at most one param reaches codegen.
+    const binder = finalize.params[0];
     return ts.raw(
-      `const __finalize = async (): Promise<any> => {\n` +
-        `  const runner = new Runner(__ctx, ${frameVar}, { state: ${frameVar}, moduleId: ${JSON.stringify(this.moduleId)}, scopeName: ${JSON.stringify(scopeName + "#finalize")} });\n` +
-        bodyStr +
-        `\n  return runner.halted ? runner.haltResult : undefined;\n};`,
+      renderFinalizeClosure
+        .default({
+          binderParam: binder !== undefined ? `${binder.name}: any` : "",
```
egonSchiele: why type it as `any`?

packages/agency-lang/lib/backends/typescriptBuilder/scopeManager.ts
```diff
@@ -150,6 +150,37 @@ export class ScopeManager {
...
 
+  /**
+   * Declared return type of the nearest ENCLOSING function or node,
+   * walking outward past block scopes. Unlike `returnType()`, which
+   * answers for the CURRENT scope (and answers `undefined` for blocks),
+   * this is the saveDraft tool's schema key (partials-ergonomics spec
+   * Part 2): a guard block owns the draft slot but carries no declared
+   * type, so the enclosing def's declared type is the best-effort hint.
+   */
+  enclosingDeclaredReturnType(): VariableType | undefined {
```
egonSchiele, line 161: any other types of scopes to consider?

packages/agency-lang/lib/runtime/intrinsicTools.ts
```diff
@@ -0,0 +1,136 @@
...
+      model: JSON.stringify(opts.model),
+      threadId: __threads()?.activeId() ?? null,
+    });
+    await invokeCallbacks({
+      ctx: opts.ctx as any,
+      name: "onToolCallStart",
+      data: { toolName: toolCall.name, args: callArgs },
+      stateStack,
+    });
+    const ack = intrinsic.handle({ toolCall, stateStack, draftSchema });
```
egonSchiele, line 123: This assumes all intrinsic tool calls will always return a user message

packages/agency-lang/lib/runtime/saveDraftTool.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { IntrinsicTool } from "./intrinsicTools.js";
+
+/** The stdlib module id every compiled stdlib export carries. No user
+ *  module can carry it, so name+module is identity in practice — the
+ *  runtime cannot import the stdlib singleton directly without a
+ *  dependency cycle (partials-ergonomics spec Part 2). */
+const STDLIB_INDEX_MODULE = "stdlib/index.agency";
+
+/** The contract the type system cannot check, stated to the model. */
+const DESCRIPTION =
```
egonSchiele, line 16: Do you think this is a good tool description? Would this be helpful for the LLM?

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: Hoisted per your suggestion in 241377f04: `let parallelResult = { kind: "values", values: [] }` then a plain `if (dispatchCalls.length > 0)` around the call, with the whole call re-indented to its natural level.

packages/agency-lang/lib/runtime/saveDraftTool.ts
egonSchiele: Went with the stronger option in 241377f04: `draftCharCount` now try/catches (returns 0 on a throw) since it is an exported helper, with a real circular-input test alongside the undefined one. The doc comment states why the throw is unreachable from the tool path (args are JSON-origin) but honored anyway.

packages/agency-lang/lib/runtime/intrinsicTools.ts
egonSchiele: Comment added in 241377f04, stating exactly the two reasons you named: saving on the way down is salvage-friendly, and answering avoids a dangling tool_use — plus an explicit "do not add the cancel check here."

packages/agency-lang/lib/backends/typescriptBuilder/finalizeCodegen.ts
egonSchiele: Two reasons, now stated in a comment (241377f04). First, the annotation is documentation only: generated TS is transpiled by esbuild, never type-checked, so nothing enforces it — the enforced typing is the checker's `T | null` on the Agency side. Second, the binder's true type can involve inferred pieces codegen cannot render. That said, when the user WRITES an annotation (`finalize as d: string`), we now emit it (`draft: string | null`) — the same convention inline handler params follow (annotation when present, `any` otherwise). Pinned by a codegen test.

packages/agency-lang/lib/backends/typescriptBuilder/scopeManager.ts
egonSchiele: The `Scope` union names seven kinds, but the builder only ever pushes `function`, `node`, and `block` onto this stack, over the root `global` — `imported`/`static`/`local` are reference-classification scopes that never become stack entries. `returnType()` directly above relies on the same invariant (its default arm THROWS "Unknown scope type"). So the walk always lands on function/node/global. Comment added in 241377f04 saying exactly this.

packages/agency-lang/lib/runtime/intrinsicTools.ts
egonSchiele: Correct, and it is now a stated contract on the type (241377f04) rather than an assumption. Two layers to it: every provider REQUIRES a tool-result message per tool_use — so "every intrinsic call produces a result" is protocol, not design. The design choice is that the result is plain TEXT, which holds for saveDraft (state write + ack) and likely for the next few candidates (attachment listing would return text too). If an intrinsic ever needs richer output — attachments, structured content — the `handle` return type widens then; the loop plumbing in runIntrinsicCall is the only consumer, so it is a one-place change. Happy to widen it now to a `{ text } | ...` union if you would rather pay that cost up front.

packages/agency-lang/lib/runtime/saveDraftTool.ts
egonSchiele: Fair challenge — the old text was written for the reviewer, not the model. Rewritten in 241377f04 around what the model actually needs to act: WHEN to call ("whenever your answer improves", "whenever you complete a meaningful piece of work"), the replacement semantics ("each call replaces the previous draft, the last one wins"), WHY it is worth doing ("if the run is stopped early by a cost or time budget, the last saved draft is returned instead of a failure"), and the one failure mode worth heading off ("always pass the complete answer as it stands, not a note or a diff" — models love saving progress notes). Dropped "must match this function's return type": the model cannot see "this function", and the value's shape is already communicated by the schema itself. The remaining weakness is the schema-mismatch case (#580); prose cannot fix that.

PR #581, opened 2026-07-18
Sticky interrupt prompt for the line-mode agent

egonSchiele:
## What this does

When the line-mode agent stops to ask you to approve an action while other branches are still running and printing, the approval prompt now stays **pinned to the bottom of the terminal** and the concurrent tool-call traces stream into scrollback **above** it, instead of the prompt getting buried and smeared.

Before: a second branch's `⏺ tool(...)` traces printed straight into the region the `prompts` menu was managing, corrupting it. After: the prompt is a sticky footer; traces scroll above it live.

Full background, root-cause trace, and design rationale are in the committed spec and plan:
- `docs/superpowers/specs/2026-07-17-sticky-interrupt-prompt-design.md`
- `docs/superpowers/plans/2026-07-17-sticky-interrupt-prompt.md`

## Approach

Split "what" from "how" throughout, and reuse existing machinery rather than reinventing it:

- **Bottom-region coordinator** (`installBottomRegion` in `lib/stdlib/cli.ts`) generalizes the "Thinking" spinner's `process.stdout.write` monkeypatch from one line to a multi-line footer. The exact terminal bytes come from a pure `buildFrame(spec) -> {seq, rows}`; the shell just patches stdout and tracks rows. Widths/wrapping reuse `visualWidth`/`wrapText` from `lib/stdlib/layout/ansi.ts`.
- **Spinner folded onto the coordinator**, so there is exactly one owner of the bottom of the screen (removes a load-bearing two-patcher ordering dependency).
- **Sticky interrupt widget** rendered as that footer. Input is "type your answer, then Enter" (a bare `a` cannot disambiguate `a` from `aa`/`ap`, and a free-text reason can start with any letter). All decisions live in a pure `reduceInterrupt(state, action, config) -> {state, outcome}` state machine; the shell wires `readline`'s `_ttyWrite` to it and acts on the outcome. It reuses the REPL's existing readline (no second one), so none of `_runPrompt`'s paused/raw-mode restoration dance is needed.
- **Surgical seam**: a new `interruptChoice` primitive in `std::ui/cli` that uses the sticky widget when a line-mode REPL is active and **falls back to `chooseOption`** otherwise (TUI, non-TTY, headless). `std::policy`'s `askUser` calls it instead of `chooseOption` — the single behavioral change. Every other `chooseOption` caller (`/model`, `/search`, slash palette, `/paste`, …) is untouched.

Because the fix lives in `std::policy` + `std::ui/cli`, every line-mode agent built on `cliPolicyHandler` benefits, not just the Agency agent.

## Anti-flicker

Each repaint is one atomic `stdout.write`, wrapped in DEC synchronized-output mode (`\x1b[?2026h/l`), redrawn on events (not a timer). The cursor is hidden only for the widget (the spinner is unchanged), with a `process.once("exit")` restore for crash-safety, and the reason line draws its own caret glyph.

## Testing

- Pure cores are exhaustively unit-tested via the `_internal` convention: `buildFrame` asserts exact byte sequences (catches erase-math regressions), `reduceInterrupt` is a full transition table (bare-`a` does NOT early-resolve, empty-Enter is must-answer, Ctrl+C yields an `exit` outcome, free-text reasons starting with option letters, invalid-with-notice, backspace, newline-stripping).
- Integration smoke tests drive the widget with a trace streamed above it.
- `lib/stdlib/cli.test.ts`: 87 pass. Policy tests: 30 pass. Full `lib/stdlib` + `lib/runtime` unit suites: 2686 pass, 0 failures. `tsc --noEmit` clean. `make` builds; both `stdlib/ui/cli.js` and `stdlib/policy.js` load with no import cycle or load-time error.

## Not included / follow-ups

- **Burst coalescing** (collapse several same-tick footer redraws into one via a microtask) is specified as an optional refinement in the plan (Task 7). Left out of v1; worth adding only if real usage shows visible footer "twitch."
- **Live concurrent-TTY manual check** (drive the real agent with two parallel branches, one needing approval) is still pending — it needs an interactive TTY + LLM, so it was not run in CI/automation. The logic and rendering are covered by the unit/integration tests above.

### DIFF
--- a/packages/agency-lang/lib/stdlib/cli.ts
+++ b/packages/agency-lang/lib/stdlib/cli.ts
@@ -189,8 +189,6 @@ function recordHistoryEntry(history: string[], entry: string, command: string): 
 const USER_INPUT_COLOR = styles.cyan
 const COLOR_RESET = RESET;
 const DIM = styles.dim;
-const CLEAR_LINE = "\r\x1b[K";
-
 const HIDE_CURSOR = "\x1b[?25l";
 const SHOW_CURSOR = "\x1b[?25h";
 const SYNC_BEGIN = "\x1b[?2026h"; // DEC synchronized-update begin
@@ -489,62 +487,31 @@ const SPINNER_FRAMES = SPINNERS.line.frames;
 const SPINNER_INTERVAL_MS = SPINNERS.line.interval;
 
 /**
- * Start a single-line "Thinking Ns" spinner. Returns a stop function
- * that clears the spinner row and restores `process.stdout.write`.
- *
- * While the spinner is active, every external write to stdout
- * (`console.log`, `process.stdout.write`, etc.) is wrapped to emit
- * a `\r\x1b[K` clear-line first, then the original content. The
- * spinner redraws on the next interval tick. This is the
- * "unconditional clear-on-write" approach the spec endorses for WL4
- * — ~15 lines, handles every tool that logs without needing each
- * tool to coordinate with the spinner.
- *
- * No-op on non-TTY: returns an immediate stop function so piped
- * runs don't get spinner frames in their logs.
+ * Start a single-line "Thinking Ns" spinner as a one-line bottom region.
+ * Returns a stop function that removes the region. The spinner is now a
+ * client of `installBottomRegion`: outside writes (tool traces) redraw
+ * above it automatically, and there is exactly one owner of the bottom of
+ * the screen. It does NOT hide the cursor (default RegionOptions). No-op
+ * on non-TTY: returns an immediate stop function so piped runs don't get
+ * spinner frames in their logs.
  */
 function startSpinner(useTTY: boolean): () => void {
-  if (!useTTY) return () => { };
+  if (!useTTY) {
+    return () => { };
+  }
   const startedAt = Date.now();
-  let stopped = false;
-  // Capture the *real* write before patching so the spinner itself
-  // can draw without recursing through the patched version.
-  const stdoutAny = process.stdout as unknown as {
-    write: (chunk: any, ...rest: any[]) => any;
-  };
-  const realWrite = stdoutAny.write.bind(process.stdout);
-
-  const render = (): void => {
-    if (stopped) return;
+  const render = (): string[] => {
     const elapsedMs = Date.now() - startedAt;
     const elapsedSec = Math.floor(elapsedMs / 1000);
-    const idx = Math.floor(elapsedMs / SPINNER_INTERVAL_MS) % SPINNER_FRAMES.length;
-    const frame = SPINNER_FRAMES[idx];
-    realWrite(`\r${DIM}${frame} Thinking ${elapsedSec}s${COLOR_RESET}\x1b[K`);
+    const frameIndex =
+      Math.floor(elapsedMs / SPINNER_INTERVAL_MS) % SPINNER_FRAMES.length;
+    return [`${DIM}${SPINNER_FRAMES[frameIndex]} Thinking ${elapsedSec}s${COLOR_RESET}`];
   };
-  render();
-  const id = setInterval(render, SPINNER_INTERVAL_MS);
-
-  // Patch stdout so any external write clears the spinner row first.
-  // The patched function uses `realWrite`, never `stdoutAny.write`,
-  // so there's no recursion. Preserves the original return value so
-  // backpressure semantics (the boolean Writable.write returns) stay
-  // correct.
-  stdoutAny.write = function patchedWrite(
-    this: unknown,
-    chunk: any,
-    ...rest: any[]
-  ): any {
-    realWrite(CLEAR_LINE);
-    return realWrite(chunk, ...rest);
-  };
-
+  const region = installBottomRegion(render, useTTY);
+  const timer = setInterval(() => region.refresh(), SPINNER_INTERVAL_MS);
   return (): void => {
-    if (stopped) return;
-    stopped = true;
-    clearInterval(id);
-    stdoutAny.write = realWrite;
-    realWrite(CLEAR_LINE);
+    clearInterval(timer);
+    region.teardown();
   };
 }
 
@@ -1432,6 +1399,7 @@ export const _internal = {
   summarizeMultiline,
   eraseRows,
   buildFrame,
+  startSpinner,
 };
 
 export function _clearScreen(): void {
--- a/packages/agency-lang/lib/stdlib/cli.ts
+++ b/packages/agency-lang/lib/stdlib/cli.ts
@@ -13,6 +13,7 @@ import { getRuntimeContext } from "../runtime/asyncContext.js";
 import { modifiers, RESET, styles } from "@/utils/termcolors.js"
 import { color, colors, bgColors } from "../utils/termcolors.js";
 import { _promptsAutocomplete } from "./ui.js";
+import { visualWidth, wrapText } from "./layout/ansi.js";
 import { isFailure } from "../runtime/result.js";
 import { isAbortError, makeAbortCause } from "../runtime/errors.js";
 import type { AbortCause } from "../runtime/errors.js";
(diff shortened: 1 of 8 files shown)


packages/agency-lang/lib/stdlib/cli.ts
```diff
@@ -188,7 +189,481 @@ function recordHistoryEntry(history: string[], entry: string, command: string):
...
+      columns: process.stdout.columns || 80,
+      cursor: hideNow ? "hide" : "keep",
+    });
+    firstPaint = false;
+    rows = frame.rows;
+    return realWrite(frame.seq);
+  };
+
+  paint(null);
+  stdoutAny.write = (chunk: any): boolean => paint(chunkToText(chunk));
```
egonSchiele: **The patched `stdout.write` drops `encoding` and the completion callback.** `Writable.write` is `(chunk, encoding?, cb?)`, and the spinner patch this replaces preserved them (`realWrite(chunk, ...rest)`). Here the arrow takes only `chunk`, so while any region is active, `process.stdout.write(data, cb)` never fires `cb`, and an explicit encoding is ignored. `console.log` doesn't pass these, so the common path is fine — but this silently violates the stream contract for any caller that does, and it's the kind of thing that surfaces as a mysterious hang far from here.

Suggest `(chunk, ...rest) => { const ret = paint(chunkToText(chunk)); const cb = rest.find(a => typeof a === "function"); if (cb) cb(); return ret; }` (or at least forward/invoke the callback). Also worth a one-line note that the returned backpressure boolean now reflects the *frame* write, not the chunk write.

packages/agency-lang/lib/stdlib/cli.ts
```diff
@@ -188,7 +189,481 @@ function recordHistoryEntry(history: string[], entry: string, command: string):
...
+  const hook = (globalThis as any).__agencyInterruptPrompt;
+  if (typeof hook !== "function") {
+    throw new Error("_interruptChoice: no active line-mode REPL");
+  }
+  return hook({ title, body, items, allowFreeText, allowCancel });
+}
+
+/** True when a line-mode REPL is running and can host a pinned prompt. */
+export function _stickyInterruptAvailable(): boolean {
+  return typeof (globalThis as any).__agencyInterruptPrompt === "function";
```
egonSchiele: **Non-TTY REPLs get routed here and can hang.** This gate only checks that the hook exists, and the hook is installed unconditionally in `_runLineRepl` (line 1313, no `useTTY` guard). So when a line-mode REPL runs with a non-TTY stdout or stdin, `interruptChoice` picks the sticky path instead of falling back to `chooseOption`. In that path `installBottomRegion` no-ops (no footer is ever drawn) and `assertRawMode` no-ops, but the widget still overrides `_ttyWrite` and awaits a keypress — and on a non-terminal stdin those keypresses don't flow through `_ttyWrite`, so the interrupt can hang with nothing on screen. The spec lists non-TTY as a non-goal that "already falls back to a plain input loop — untouched," which this quietly breaks.

Fix: require a real terminal on both ends here, e.g. `return typeof hook === "function" && process.stdout.isTTY === true && process.stdin.isTTY === true;` (mirrors the `/paste` guard `useTTY && process.stdin.isTTY` at line 1375). Then non-TTY REPLs cleanly use the `chooseOption` fallback.

packages/agency-lang/lib/stdlib/cli.ts
```diff
@@ -188,7 +189,481 @@ function recordHistoryEntry(history: string[], entry: string, command: string):
...
+    });
+    firstPaint = false;
+    rows = frame.rows;
+    return realWrite(frame.seq);
+  };
+
+  paint(null);
+  stdoutAny.write = (chunk: any): boolean => paint(chunkToText(chunk));
+
+  const region: BottomRegion = {
```
egonSchiele, line 375: **No repaint on terminal resize while a region is live.** The spec's Edge Cases require "repaint on the terminal resize event so the footer stays coherent if the window changes while you are deciding." `paint` reads `process.stdout.columns` fresh each time, but nothing triggers a repaint on resize, and `eraseRows(prevRows)` uses the row count computed at the *previous* width. So if the user resizes mid-prompt with nothing streaming, the next keypress's frame erases the wrong number of rows and the footer smears.

Suggest registering `process.stdout.on("resize", () => region.refresh())` for the region's lifetime and removing it in `teardown` (alongside restoring `stdoutAny.write`).

packages/agency-lang/lib/stdlib/cli.ts
```diff
@@ -188,7 +189,481 @@ function recordHistoryEntry(history: string[], entry: string, command: string):
...
+type InterruptFooterInput = {
+  title: string;
+  body: string;
+  items: { key: string; label: string }[];
+  allowFreeText: boolean;
+  state: InterruptState;
+  columns: number;
+};
+
+function bodyLines(body: string): string[] {
```
egonSchiele: Minor: the body cap is on **logical** lines, not physical rows. `bodyLines` caps to `INTERRUPT_BODY_MAX_LINES` (6) split-on-`\n` lines, but `buildFrame` then re-wraps each via `wrapText`, so a single very long body line still expands to many physical rows and the footer isn't actually bounded. If the intent is "the pinned region always fits," cap after wrapping (or truncate each shown body line to width) rather than before.

packages/agency-lang/lib/stdlib/cli.ts
```diff
@@ -188,7 +189,481 @@ function recordHistoryEntry(history: string[], entry: string, command: string):
...
+  if (cursor === "hide") {
+    return HIDE_CURSOR;
+  }
+  if (cursor === "show") {
+    return SHOW_CURSOR;
+  }
+  return "";
+}
+
+function withTrailingNewline(text: string): string {
```
egonSchiele, line 245: Minor / note-only: `withTrailingNewline` force-appends `\n` to any outside write lacking one, so while a region is active a partial-line write (`process.stdout.write("Loading...")`, or a `\r`-based progress bar) becomes its own line. The old spinner did `CLEAR_LINE + chunk` and didn't force a newline. It's fine for the agent's full-line tool traces (which is all that streams here), but worth a comment that partial-line/carriage-return output isn't supported under a footer — otherwise it'll read as a bug if someone hits it later.

egonSchiele (commented):
Reviewed the implementation against the spec, the plan, and the earlier review findings. **The prior findings are all cleanly addressed** — `hideCursor` is scoped to the widget (spinner keeps the cursor), there's a `process.once("exit")` cursor restore, the widget draws its own caret glyph so the hidden real cursor no longer fights the reason line, widths reuse `visualWidth`/`wrapText` (no homemade duplication), and the pure `buildFrame` / `reduceInterrupt` cores are exhaustively table-tested. Nice split of what-vs-how.

I verified the two load-bearing runtime assumptions still hold in this code (`_ttyWrite` interception is live mid-turn via the turn-scoped `installCancelKey`; the widget correctly saves/restores *that* wrapper as `originalTtyWrite`; Escape rejects with `AgencyCancelledError` the same way `_promptsAutocomplete` does). Also confirmed the spinner double-stop I worried about is a non-issue — `stopActiveSpinner` guards with `activeStopSpinner = null`, so the raw stop fn runs at most once.

Five findings below, none blocking. The two I'd fix before merge are the **non-TTY hang risk** (a line-mode REPL with a non-TTY stdout/stdin now routes to an invisible sticky widget instead of `chooseOption`'s non-TTY loop, and can hang waiting on keystrokes that never arrive) and the **dropped `write` callback/encoding args**. The **missing resize repaint** is a spec-listed edge case that's not implemented. The other two are minor.

packages/agency-lang/lib/stdlib/cli.ts
egonSchiele: Fixed in 7aed7aa55. `_stickyInterruptAvailable` now requires the hook AND `process.stdout.isTTY === true && process.stdin.isTTY === true`, mirroring the `/paste` guard, so a non-TTY line-mode REPL falls back to `chooseOption` instead of the invisible-and-hanging sticky path. Added a unit test covering all four combinations (hook present + both TTYs → true; non-TTY stdin/stdout or missing hook → false).

packages/agency-lang/lib/stdlib/cli.ts
egonSchiele: Fixed in 7aed7aa55. The patched write is now `(chunk, ...rest) => { ... }`: it pulls the encoding (`typeof rest[0] === 'string'`) and the completion callback out of `rest` and threads the callback through to the frame's `realWrite(frame.seq, cb)`, so it fires when the frame (which carries the caller's text) flushes. Added a comment that the returned backpressure boolean reflects the frame write, not the chunk write. Test asserts a forwarded `write(buf, cb)` callback runs.

packages/agency-lang/lib/stdlib/cli.ts
egonSchiele: Fixed in 7aed7aa55. `installBottomRegion` now registers `process.stdout.on('resize', () => paint(null))` for the region's lifetime and removes it in `teardown` (alongside restoring `stdoutAny.write`). Test emits `resize` and asserts the footer repaints.

packages/agency-lang/lib/stdlib/cli.ts
egonSchiele: Fixed in 7aed7aa55. `bodyLines` now wraps the body via `wrapText(body, width - 1)` and caps to `INTERRUPT_BODY_MAX_LINES` *physical* rows (+ ellipsis). Since each wrapped line is already within width, buildFrame's re-wrap is a no-op, so the pinned region is actually bounded. Test: a single 500-char body line at columns=40 stays ≤ 6 rows + '…'.

packages/agency-lang/lib/stdlib/cli.ts
egonSchiele: Added a comment in 7aed7aa55 on the patched write noting that partial-line / carriage-return output (`write('Loading...')`, `\r` progress bars) becomes its own line under a footer, and that full-line tool traces are all that stream here. Left as documented behavior, not a code change.

packages/agency-lang/lib/stdlib/cli.ts
```diff
@@ -188,7 +189,481 @@ function recordHistoryEntry(history: string[], entry: string, command: string):
...
+    }
+  });
+}
+
+function chunkToText(chunk: unknown): string {
+  if (typeof chunk === "string") {
+    return chunk;
+  }
+  return String(chunk);
+}
```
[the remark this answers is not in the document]
egonSchiele: Addressed in 7aed7aa55. `chunkToText` now decodes `Buffer` (`.toString(encoding ?? 'utf8')`) and raw `Uint8Array` (`Buffer.from(chunk).toString(...)`) before the `String()` fallback. Note `String(Buffer)` already decoded fine, but `String(Uint8Array)` produced comma-joined bytes — that case is now correct. Encoding comes from the write args.

packages/agency-lang/lib/stdlib/cli.ts
```diff
@@ -188,7 +189,481 @@ function recordHistoryEntry(history: string[], entry: string, command: string):
...
+        columns: process.stdout.columns || 80,
+        cursor: options.hideCursor ? "show" : "keep",
+      });
+      realWrite(frame.seq);
+      rows = 0;
+      if (activeRegion === region) {
+        activeRegion = null;
+      }
+    },
+  };
```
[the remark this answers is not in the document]
egonSchiele: Addressed in 7aed7aa55. The write wrapper now forwards `encoding` and the completion `cb` (threaded to the frame's realWrite), and a `resize` listener triggers `paint(null)` for the region's lifetime, removed on teardown. See the replies on the corresponding review threads for details + tests.

Issue #582, opened 2026-07-18
Checker expected-type propagation into block returns: make structured llm() shapes compile, unblock responseFormat inside guard blocks


egonSchiele:
## Problem

`responseFormat` for `return llm(...)` inside a block cannot be fixed by codegen-side annotation threading (#580 tried and descoped it). Empirical verification during the #580 plan review:

| Program shape | Typechecks? | responseFormat today | Would a threaded stamp change it? |
|---|---|---|---|
| `Result<string>` guard + `return llm("hi")` | yes | string default | no — identical |
| `Result<{title}>` guard + `return llm("hi")` | **no — AG2001** (`Result<string, any>` not assignable; both assignment and return-position forms) | — | unreachable |
| `Result<{title}>` guard + `const x: {title} = llm(...)` | yes | **already `z.object(...)`** from the assignment hint | no — already right |

The only reachable observable shape is an annotation WIDER than the inferred llm type (`Result<string | number>` + `return llm()` typechecks because `string <: string \| number`, and a stamp would newly emit `z.union`). That is exactly the case where structured output would produce runtime values (a number) the checker statically typed away (`string`) — the type-theoretically muddy case, not the motivating one.

## Root cause

The checker synthesizes `llm(...)` with NO expected type pushed inward. `synthGuardCall` then types the block from its returns, so a structured guard annotation can never be met by a bare `return llm(...)` — the natural shape errors, and the workaround shape (annotated local) already solves responseFormat by itself.

## Fix (checker work)

Expected-type propagation: when checking a block whose declared yield is known (guard annotation, and eventually block-param signatures), push the expected type into `return` expressions so `llm(...)` synthesizes against it. Then:

- `const r: Result<{title}> = guard(...) { return llm(...) }` COMPILES, and
- codegen's threaded yield type (the #580 machinery, already in place for draftSchema via `BlockScope.declaredYieldType`) becomes load-bearing for responseFormat — re-adding the `ScopeManager.returnType()` block case at that point is a two-line change.

This keeps runtime structured output aligned with static types instead of diverging from them, which is why #580 descoped rather than shipping the wider-union-only version.

## Origin

#580 plan review (test-plan addendum T1), decision by the owner 2026-07-17: ship draftSchema alone; responseFormat waits for this.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #584, opened 2026-07-18
Thread guard Result<T> annotations into block codegen: honest saveDraft schemas (#580)

### LINKED ISSUE #580: Thread guard assignment annotations (Result<T>) into block codegen: honest saveDraft schema + responseFormat inside guard blocks
## Problem

Type information declared on a guard assignment never reaches codegen inside the guard block. Two features degrade because of it:

1. **The saveDraft tool schema** (#578) keys on the enclosing FUNCTION's declared return type, but the draft files on the guard BLOCK's slot. When the block yields a different type than the function returns, the schema is a wrong hint:

```ts
def makeReport(topic: string): Report {
  const notes: Result<string> = guard(cost: $0.50) {
    // The slot is a STRING, but the model is told to save a Report.
    return llm("Research " + topic, tools: [saveDraft])
  }
  ...
}
```

2. **`responseFormat`** has the same hole, already documented in llm.md: a `return llm(...)` from inside any block defaults to string structured output, because `ScopeManager.returnType()` answers `undefined` for block scopes (see the comment on that method — "we don't carry the block's declared return type onto `Scope.block`").

In both cases the user may have ALREADY written the correct type: the guard construct types as `Result<T>`, so `const notes: Result<string> = guard(...) { ... }` names the block's yield type exactly. The builder just cannot see it from inside the block body.

## Proposal

Thread the assignment annotation into the block scope, for the guard path:

- When the builder processes an assignment whose declared type is `Result<T>` and whose value is the desugared `_guard` call, unwrap `T` and carry it onto the `Scope.block` entry pushed for the guard block's body.
- `ScopeManager.enclosingDeclaredReturnType()` (and the `returnType()` used for responseFormat) answer block-first when the block carries a declared yield type, then fall back to the enclosing function/node's declared return, then the existing string default.

One mechanism, two fixes: the annotated guard above gets an honest saveDraft value schema (`z.string()`) AND `return llm(...)` inside an annotated guard gets an honest structured-output format.

## Out of scope

- Unannotated guards (the common case). `const r = guard(...) { ... }` with no annotation still needs INFERRED types to reach codegen — a typechecker-to-codegen handoff that does not exist. That remains the partials-ergonomics spec's out-of-scope item; this issue covers only the declared-annotation case, which is free precision the user already wrote down.
- Non-guard blocks (`fork`, `map`, ...): their assignment types relate to the block yield differently (arrays, races). Start with guard, where `Result<T>` → `T` is exact.

## Origin

Owner review of PR #578: "why not use the type declared on the var that the guard is assigned to?" — it is the right slot-type source when present; it just needs the threading described above.

egonSchiele:
Closes #580.

An annotated guard's `Result<T>` now reaches codegen inside the guard block, so the saveDraft tool's value schema describes the actual slot instead of the enclosing function's type:

```ts
def makeReport(topic: string): Report {
  const notes: Result<string> = guard(cost: $0.50) {
    return llm("Research " + topic, tools: [saveDraft])
  }
  ...
}
// saveDraft's schema is now z.string() (the slot), not the Report shape.
```

Spec (with all three review rounds folded and as-executed notes): `docs/superpowers/specs/2026-07-17-guard-annotation-threading-design.md`. Plan: `docs/superpowers/plans/2026-07-17-guard-annotation-threading.md`, review alongside.

## How it works

- **`guardDesugar` stamps.** The desugar walk gains a context — the *current return target*, captured from def/node declared returns — and stamps `successType` onto the `BlockArgument` it creates when a guard sits in an annotated assignment or in return position. Two pure rules carry the whole feature: `slotContext` (what context a body walks under) and `stampFor` (what a guard in a value slot gets stamped with).
- **The reset rule covers every return-retargeting boundary.** A `return` inside a block argument, an inline handler, or a finalize yields to its own closure, not the def — so `bodySlots` (the single source of truth for body shapes) gains a per-slot `retargetsReturn` marker, and the walk resets its target there. A return-position guard inside a fork branch / handler / finalize stamps nothing; a stamped guard block becomes the return target for its own body, so `Result<Result<string>>` nesting composes.
- **The builder side is deliberately tiny:** one optional field on `BlockScope`, one copy at the `processBlockArgument` push, one predicate + one switch arm in `enclosingDeclaredReturnType()` (first answer wins: stamped block, else nearest function/node).

## Descoped: responseFormat (#582)

The plan review proved empirically that no natural type-correct program exists where a responseFormat stamp changes behavior: a structured annotation with `return llm(...)` fails AG2001 (the checker infers `string` — no expected type is pushed into block returns), and the shape that does compile (an annotated local) already gets its responseFormat from the assignment hint. The one reachable shape — an annotation *wider* than the inferred llm type — is exactly the case where structured output would produce values the checker statically typed away. So `ScopeManager.returnType()` is untouched, and #582 tracks the checker expected-type propagation that makes the natural shapes compile — at which point the stamp this PR ships is the plumbing it consumes. (llm.md's block-limitation note stays true until then; not edited here — owner-authored.)

## Testing

- 14 desugar unit tests: both annotation forms (`const` and `let`), def and node return-position, the `if`-inherit branch, the three no-stamp boundaries (fork branch, inline handler, finalize), nested composition, `Result<T, E>` unwrap, statement position, double-run stability.
- 4 direct `ScopeManager` walk tests (stamped answers; unstamped defers to stamped-outer / function / node).
- 5 codegen tests, including the structured `draftSchema: z.object(...)` assertion no fixture can make, and annotation-beats-def-type discrimination.
- Fixture `savedraft-annotated-structured`: a structured draft salvages intact through the threaded schema path; the compiled JS was inspected and carries `draftSchema: z.object({ "title": z.string() ... })`.
- Byte-stability: zero tracked churn after `make fixtures`; the #578 fixtures (all unannotated) pass unchanged; full unit suite 8594 green; structural lint clean.

## Known limitations (tracked)

- Unannotated guards still fall back to the enclosing function's type (inferred yields need the typechecker-to-codegen handoff; #582's expected-type propagation is the adjacent piece).
- Alias-typed annotations (`type R = Result<string>`) are not chased — the desugar has no alias table.
- `let r: Result<T>` declared on one line and assigned later carries no annotation at the assignment site.

### DIFF
--- a/packages/agency-lang/lib/utils/bodySlots.ts
+++ b/packages/agency-lang/lib/utils/bodySlots.ts
@@ -145,7 +145,9 @@ export function bodySlots(node: AgencyNode): BodySlot[] {
     case "finalizeBlock": {
       // Same-scope statements (like an if body): the finalize reads the
       // enclosing scope's locals, so every walker and rewriter must
-      // descend into it as part of that scope.
+      // descend into it as part of that scope — though a `return`
+      // inside yields to the `__finalize` closure, not the enclosing
+      // def, hence `retargetsReturn`.
       const n = node as FinalizeBlock;
       return [
         {
@@ -158,11 +160,19 @@ export function bodySlots(node: AgencyNode): BodySlot[] {
     case "guardBlock": {
       // The guarded block. Distinct-scope statements (the body compiles
       // to a lifted closure after desugaring), but walkers and
-      // rewriters must descend into it like any other body.
+      // rewriters must descend into it like any other body. Marked
+      // retargetsReturn like its post-desugar twin (blockArgument):
+      // guardDesugar currently bypasses this slot (desugarNode
+      // special-cases guardBlock and walks the body itself, with the
+      // stamp as the target), so the flag is inert there — but anyone
+      // merging that special case into the generic slot walk gets the
+      // reset by construction instead of silently inheriting the
+      // def's return target into guard bodies.
       const n = node as GuardBlock;
       return [
         {
           body: n.body,
+          retargetsReturn: true,
           write: (owner, body) => ({ ...owner, body }) as AgencyNode,
         },
       ];
--- a/packages/agency-lang/lib/utils/bodySlots.ts
+++ b/packages/agency-lang/lib/utils/bodySlots.ts
@@ -38,8 +38,15 @@ export type BodySlot = {
    *  `static ...`): a rewrite must map the one statement to exactly one. */
   single?: boolean;
   /** Extra ancestor between the owner and the body during walks — the
-   *  functionCall's inline `block:` argument. */
+   *  functionCall's `block:` argument (inline or not). */
   blockAncestor?: BlockArgument;
+  /** True when a `return` inside this slot's body yields to the slot's
+   *  own closure rather than the enclosing def: block arguments (the
+   *  __block_N lifting), inline handler bodies (their own arrow), and
+   *  finalize bodies (the __finalize closure). guardDesugar's return-
+   *  target rule keys on this — a new return-retargeting construct
+   *  must set it here or that feature silently mis-stamps. */
+  retargetsReturn?: boolean;
   /** Fresh copy of `owner` with this slot's statements replaced. Takes the
    *  CURRENT owner (not the node bodySlots was called on) so a fold over
    *  several slots composes: each write builds on the previous one. */
@@ -126,6 +133,7 @@ export function bodySlots(node: AgencyNode): BodySlot[] {
       if (n.handler.kind === "inline") {
         slots.push({
           body: n.handler.body,
+          retargetsReturn: true,
           write: (owner, body) => {
             const o = owner as HandleBlock;
             return { ...o, handler: { ...o.handler, body } } as AgencyNode;
@@ -142,6 +150,7 @@ export function bodySlots(node: AgencyNode): BodySlot[] {
       return [
         {
           body: n.body,
+          retargetsReturn: true,
           write: (owner, body) => ({ ...owner, body }) as AgencyNode,
         },
       ];
@@ -167,8 +176,16 @@ export function bodySlots(node: AgencyNode): BodySlot[] {
       return [];
     case "messageThread":
       return [bodyField(node as MessageThread)];
-    case "blockArgument":
-      return [bodyField(node as unknown as BlockArgument)];
+    case "blockArgument": {
+      const n = node as unknown as BlockArgument;
+      return [
+        {
+          body: n.body,
+          retargetsReturn: true,
+          write: (owner, body) => ({ ...owner, body }) as AgencyNode,
+        },
+      ];
+    }
     case "functionCall": {
       const n = node as FunctionCall;
       if (!n.block) {
@@ -179,6 +196,7 @@ export function bodySlots(node: AgencyNode): BodySlot[] {
         {
           body: block.body,
           blockAncestor: block,
+          retargetsReturn: true,
           write: (owner, body) => {
             const o = owner as FunctionCall;
             if (!o.block) {
--- a/docs/superpowers/plans/2026-07-17-guard-annotation-threading-REVIEW.md
+++ b/docs/superpowers/plans/2026-07-17-guard-annotation-threading-REVIEW.md
@@ -1,0 +1,644 @@
+# Plan review: guard annotation threading (#580)
+
+**Reviewing:** `docs/superpowers/plans/2026-07-17-guard-annotation-threading.md`
+**Verdict:** Executable, and unusually well-verified — every code fact I
+spot-checked held, including the one the whole design rests on. Both spec-review
+round-2 fixes are genuinely folded, not just name-checked. **One substantive
+gap:** the return-target reset rule keys on `blockAncestor`, which marks only
+`functionCall` block arguments — but inline handler bodies and finalize bodies
+*also* retarget `return`, so they inherit the def's type and can be mis-stamped.
+That is the same bug class the plan exists to fix, just through a different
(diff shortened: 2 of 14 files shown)


packages/agency-lang/lib/utils/bodySlots.ts
```diff
@@ -38,8 +38,15 @@ export type BodySlot = {
...
-   *  functionCall's inline `block:` argument. */
+   *  functionCall's `block:` argument (inline or not). */
   blockAncestor?: BlockArgument;
+  /** True when a `return` inside this slot's body yields to the slot's
+   *  own closure rather than the enclosing def: block arguments (the
+   *  __block_N lifting), inline handler bodies (their own arrow), and
+   *  finalize bodies (the __finalize closure). guardDesugar's return-
+   *  target rule keys on this — a new return-retargeting construct
+   *  must set it here or that feature silently mis-stamps. */
+  retargetsReturn?: boolean;
```
egonSchiele, line 49: This invariant is the right way to make the rule survive the next construct — but there's one construct it doesn't yet cover: **`guardBlock` (the `case "guardBlock"` at ~158) is left unmarked**, and its own comment says why it shouldn't be: "the body compiles to a lifted closure after desugaring." Its post-desugar twin `blockArgument` *is* marked, so the pre- and post-desugar forms of the same construct now disagree.

No live bug today: `desugarNode` early-returns on `node.type === "guardBlock"` and calls `desugarGuardBlock`, which walks the body with an explicit context, so this slot is never consulted — and no other `BodySlot` consumer reads the flag. Marking it would be inert right now.

But it's a latent trap of exactly the kind this field exists to prevent. That `desugarNode` special case looks eminently mergeable (both branches end up calling `desugarGuardBlock`), and anyone who simplifies it would route guard bodies through `bodySlots`, get no reset, and silently inherit the def's return target into a guard body — invisible, because the stamp only feeds a best-effort schema hint, never a compile error.

Suggest setting `retargetsReturn: true` on the `guardBlock` case with a note that `guardDesugar` currently bypasses this slot, so the invariant holds by construction rather than by coincidence.

packages/agency-lang/lib/utils/bodySlots.ts
```diff
@@ -142,6 +150,7 @@ export function bodySlots(node: AgencyNode): BodySlot[] {
       return [
         {
           body: n.body,
+          retargetsReturn: true,
```
egonSchiele, line 155: Minor: the comment a few lines up ("Same-scope statements (like an if body)") now sits oddly next to this flag. It's still true of the *variable* scope — the finalize reads the enclosing locals, which is why walkers must descend as part of that scope — but this flag asserts the opposite about `return`, which yields to the `__finalize` closure. Both are true; the comment states only the one that makes the flag look like a mistake.

One clause fixes it: "…as part of that scope — though a `return` inside yields to the `__finalize` closure, not the enclosing def, hence `retargetsReturn`."

egonSchiele (commented):
## Review: guard annotation threading (#580)

I reviewed the spec, the plan, the anti-pattern audit, and the test plan for this work, so this pass focuses on what changed in execution. **Every finding from those rounds is folded, and several are folded better than I proposed.** Specifically:

- **`retargetsReturn` on `BodySlot`** — you took the source-of-truth option rather than the cheap one. The marker's doc comment ("a new return-retargeting construct must set it here or that feature silently mis-stamps") is exactly the right way to make the invariant survive the next construct.
- **`stampFor` extracted** as a sibling of `slotContext`, so the two rules of the feature now read side by side and `desugarNode` is pure traversal.
- **`enclosingDeclaredReturnType` kept `.find`** — one predicate plus one switch arm, as the smaller diff.
- **`desugarGuardBlock` is a single literal again**, with `declaredYieldType` always assigned and the "absent === undefined, JSON.stringify drops it" reasoning captured in the comment.
- **responseFormat descoped to #582** with the empirical justification written down. `ScopeManager.returnType()` untouched is the right call, and framing #580 as the plumbing #582 will consume is honest about what ships.

The test additions land where they were missing: the four direct `ScopeManager` walk tests, the `if`-inherit positive counterpart to the resets, the handler/finalize/fork boundary trio, the structured `draftSchema: z.object(...)` assertion no fixture can make, double-run stability, and the `let` form. Test 1 now makes the def type differ from the annotation so the source is visibly discriminated, and the vacuous `not.toContain("draftSchema: Report")` gained the assertion that can actually fire.

I also checked two things before raising them, and neither is a problem: `enclosingDeclaredReturnType`'s new answer has exactly one caller (the saveDraft schema), so the blast radius stays contained; and `parallelBlock`/`seqBlock` bodies are same-scope (bindings hoist out, grouped with `ifElse` branches in `parallelDesugar`), so leaving them unmarked is correct, not an omission.

Two small things below, both in `bodySlots.ts`. Neither blocks merge.

packages/agency-lang/lib/utils/bodySlots.ts
egonSchiele: Marked in 8f4811d28, with the note you suggested: guardBlock now carries `retargetsReturn: true` like its post-desugar twin, and the comment states that guardDesugar currently bypasses this slot (desugarNode special-cases guardBlock and walks the body itself with the stamp as the target), so the flag is inert today — but anyone merging that special case into the generic slot walk gets the reset by construction. Verified inert: all desugar/construct/codegen tests unchanged.

packages/agency-lang/lib/utils/bodySlots.ts
egonSchiele: Added your clause verbatim in 8f4811d28 — the comment now states both facts: same VARIABLE scope (why walkers descend as part of it), and `return` yielding to the `__finalize` closure (why the flag).

PR #589, opened 2026-07-18
Refactor the stdlib agents into one agent per file, with a standard contract

egonSchiele:
Stacked on `adit/coding-agents` (#587). Executes `docs/superpowers/plans/2026-07-18-stdlib-agents-refactor.md` after its review round.

## What this does

Turns the tangled pile of agent code into a library where each file holds one agent with one job. Every agent now takes the same arguments, saves partial work, runs in its own thread, and returns a `Result`.

The organizing idea is three roles, in two trees:

- **Coding** agents produce work.
- **Review** agents read a work product handed to them and return findings. They never touch the file system.
- **Verifier** agents run and measure work with tools.

The rule to remember: **review reads, the verifier runs.**

| General (`std::agents/`) | Agency (`std::agents/agency/`) | Job |
|---|---|---|
| `codingAgent` | `agencyCodingAgent` | Produce the work |
| `reviewAgent` | `agencyReviewAgent` | Read a work product, return findings |
| `verifierAgent` | `agencyVerifierAgent` | Run and measure the work |
| `expertAgent` | `agencyExpertAgent` | Consult only, return guidance |
| `researcherAgent`, `plannerAgent` | | |

Naming is now mechanical: `agents/X.agency` exports `XAgent`, `agents/agency/X.agency` exports `agencyXAgent`. Shared vocabulary lives in `agents/lib/` (`Feedback`, `ExpertGuidance`, `withContext`, `llmOptions`).

## The standard agent contract

Every agent takes `(task, context, ...role params, maxCost, maxTime, model, provider, session)` and returns a `Result`. Inside: a `guard` for the budgets, one `thread` block, `saveDraft` at each milestone, and a guard-failure path that returns the most useful thing it has rather than discarding work.

## What was removed

- Both copies of the consult-then-solve `expertAgent`. Callers write the three-line composition, so experts do one thing.
- The domain classifier. After the expert split, every caller already knows which expert it needs.
- The old run-and-judge `agencyCodingAgent` and `writeAgency` in `std::agency`. That module is compiler tooling again; the write-then-verify loop is a documented caller composition.
- `unwrapGuard`. Agents return a `Result`, so callers see a typed failure instead of a sentinel string baked into the success type.
- `codingAgent`'s verify-and-fix loop and checkpoint handler, which became `supervise`.

## New: `supervise`

`std::supervise` runs any block under a re-arming time guard, pausing at an interval to let a caller-supplied check continue, redirect with a message, or stop. It is an extraction of the checkpoint handler that lived inside `codingAgent`, so no agent contains pause-and-check logic:

```agency
supervise(every: 5m, maxTime: 30m, check: myCheck) {
  return codingAgent(task, maxTime: 30m)
}
```

The handler is gated on the guard label, so an inner agent's own trip reaches its own handler instead of being answered here and silently corrupting the elapsed-time accounting.

## Two runtime changes

- `threadIsNew()` in `std::thread`, so an agent sends its system prompt exactly once per thread without module-level state that would leak across runs.
- `thread(session: "")` now means "no session" instead of opening a real session named empty-string that every agent passing the default would share. This removes a session branch that would otherwise be duplicated in every agent.

## Bugs found and fixed along the way

- **`concurrencyPool` returned nulls for every item.** Fork branches share module globals but never function locals, so worker writes to the shared results array were lost; the index pairing also assumed `map` passes an index to its function, which it does not. Now it deals items round-robin into per-worker chunks and reassembles from the branch return values. Regression test added.
- **The researcher failed on every call when `BRAVE_API_KEY` was set.** Wikipedia search and web search both advertise the tool name `search`, and an import alias renames only the binding. Both are renamed explicitly now.
- **The prompt files this branch extracted to markdown were never unwrapped.** `read()` returns a `Result` that was being interpolated into the system prompt as-is.

## Language bug found, pinned, not fixed

A `match` expression inside a block that pauses and resumes evaluates to `null`. Reading the same Result with `isFailure(r)` / `r.value` works, and a `match` on the guard's own Result outside the block works.

This came out of the plan's nested-guard gate: nested guards themselves are fine (an inner guard's Result survives an outer trip and resume), which is what `supervise` needed. The narrower bug is pinned in `tests/agency/supervise/nestedGuardResume.agency`, where `matchInsideResumedBlock` currently asserts the buggy `null` so a fix flips the test, and recorded in `TODO.md`. Every agent reads its guard Result with `isFailure` for this reason, since agents are meant to run inside supervised blocks.

## Agency agent

Out of scope except for the repointing needed to build, but per review the expert checklist still gates the work on the escalation path: consult, solve, then `verifierAgent(criteria: guidance.checklist)` with one fix round. Without that, the consult's most concrete output would have become decoration.

## Verification

- `make` green.
- 8667 unit tests pass (`npx vitest run`, 545 files).
- 48 agency execution tests pass across 18 files: `threadIsNew`, `libShared`, `agency-review`, `agencyVerifier`, `agencyCoding`, `agencyCodingRepair`, `verifier`, `verifyFailOpen`, `review`, `expert`, `coding`, `researcher`, `concurrencyPool`, `plannerReject`, `plannerHelpers`, `helpers`, `routing`, `supervise/nestedGuardResume`, `supervise/supervise`.
- Diff audited against `docs/dev/anti-patterns.md`: named the verifier pool size, replaced single-character lambda and parameter names in the lib layer.
- Docs regenerated. The stdlib `agents` group was never linked in the sidebar before; it is now, along with `supervise`.

## Known limitations

- Two pre-existing diagnostics in `lib/agents/review/agent.agency` and `lib/agents/policy/agent.agency` (the same unwrapped `read()` pattern) are untouched here and identical on `main`. `make` still exits 0.
- The redirect path of `supervise` is verified for an in-flight trip. A trip at a step boundary between requests queues the message the same way but is not separately pinned.
- The agency agent restructure (splitting `codeAgent` into interactive and autonomous, moving triage up to the coordinator) is deliberately not in this PR.


packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -0,0 +1,216 @@
...
+  // match expression inside a block that pauses and resumes evaluates to
+  // null. See tests/agency/supervise/nestedGuardResume.agency.
+  if (isFailure(captured)) {
+    // No draft was ever saved: the trip happened before the first attempt.
+    return failure({ source: "", problems: ["agencyCodingAgent stopped before generating: ${captured.error.type} cap reached"] })
+  }
+  // A rejected trip with a saved draft arrives success-shaped carrying the
+  // draft, which has the same WriteOutcome shape as a normal return, so
+  // this covers both.
+  return outcomeToResult(captured.value)
```
egonSchiele: This is the claim I would most want a test behind. The comment says a rejected trip with a saved draft "arrives success-shaped carrying the draft" — but `saveDraft(outcome)` runs inside `generateLoop`, which is a callee frame, and the guard block receives its result at an assignment statement (`outcome = generateLoop(...)`).

`docs/dev/saveDraft.md` rule 2 says: "A caller receiving an aborted result at a statement — an assignment or a bare call — stops and returns its own saved draft (`carryThrough`). **The callee's partial is dropped: salvage is opt-in per level.**" The doc then explains the post-call check binds the callee's partial into the assigned local and stops through `carryThrough(...).withFinalize(...)` — "which is why the finalize can just read the local."

This guard block has no `finalize`, and `generateLoop`'s frame has no `saveDraft` of its own at the boundary. So on a mid-generation trip I would expect `captured` to be a failure with the drafted `outcome` dropped, and line 210 to return `source: ""` — losing exactly the partial work the design is meant to salvage.

Worth noting `verifierAgent` gets this right for what looks like the same reason (`stdlib/agents/verifier.agency:107-109`): "finalize is the salvage mechanism here (a saveDraft would...)". So the distinction is clearly understood — which makes me think this is an oversight rather than a considered difference.

Could you either add a test that trips the cost or time guard partway through the loop and asserts the last source comes back, or confirm you drove this manually? If the salvage does not hold, the fix is probably a `finalize { return outcome }` on the guard block. Same question applies to `codingAgent` (`saveDraft(reply)` in `solve()`), `researcherAgent` (line 61), and `plannerAgent` (line 231) — all save in a callee with no finalize at the guard boundary.

packages/agency-lang/stdlib/concurrency.agency
```diff
@@ -61,13 +61,41 @@ export def concurrencyPool(
...
+  }
+  const chunks: any[] = []
+  let worker = 0
+  while (worker < size) {
+    chunks.push([])
+    worker = worker + 1
+  }
+  let position = 0
+  for (pair in indexed) {
+    chunks[position % size].push(pair)
```
egonSchiele, line 85: The diagnosis here is convincing and the bug was real. One consequence worth calling out explicitly, though: this changes the pool from **work-stealing** to **static round-robin assignment**.

The old `while (clonedItems.length > 0) { pop() }` had each worker pull the next item as it freed up, so a worker that drew a slow item did not hold up the others. Dealing chunks up front means wall-clock is now the slowest *chunk*, not the slowest *item*. With uneven item costs the worst case is roughly `size`× worse than stealing.

That matters for the caller added in this PR: `verifierAgent` fans out one tool-equipped LLM judge per criterion (`POOL_SIZE = 4`), and judge latency varies a lot depending on how many tool round-trips each criterion needs. Four criteria where one needs six `grep` calls and three need none is exactly the shape that suffers.

Since the old behavior returned all nulls, nothing can be depending on the stealing semantics, so this is not a regression in practice. But the docstring at line 58 still says "Run a block in parallel across a bounded pool of workers," and "pool" reads as work-stealing to most people. Could you add a line to the docstring saying items are assigned to workers up front, so uneven item costs are not load-balanced? If dynamic balancing is wanted later, the module-globals-are-shared route is presumably still open.

Separately: `chunks[position % size]` with `size: 0` gives a modulo-by-zero and an undefined chunk. An early `if (size <= 0) { return map(items, \_ -> null) }` — or a failure — would be cheap insurance on a public API.

packages/agency-lang/lib/stdlib/thread.ts
```diff
@@ -236,6 +236,19 @@ export async function _getTokens(): Promise<number> {
...
+ * use this to send their system prompt exactly once per thread: a fresh
+ * thread has no messages, while a resumed session already carries its
+ * prompt. */
+export async function _threadIsNew(): Promise<boolean> {
+  const { threads } = getRuntimeContext();
+  const activeId = threads.activeId();
+  if (activeId === undefined) {
+    return true;
+  }
+  return threads.get(activeId).messages.length === 0;
```
egonSchiele, line 249: `ThreadStore.get()` is typed `get(id: MessageThreadID): MessageThread` but the body is `return this.threads[id]`, which is `undefined` for an id not in the store. Everywhere else in the runtime guards this — e.g. `runner.ts:723` uses `threads.get(tid)?.parentId`.

Here `activeId` comes off the active stack so it should always be present, and I could not construct a case where it is not. But `?.messages.length` costs nothing and matches the surrounding defensive style, and this runs on every agent's first statement inside a thread — a throw here would surface as a confusing failure a long way from the cause.

Separately, worth confirming the semantics hold for one case I could not rule out: a thread whose messages get summarized or evicted down to zero. If `messages.length` can return to 0 on a live thread, `threadIsNew()` goes true again and the agent re-sends its system prompt mid-conversation. Probably not reachable given how summarize works, but the docstring says "a fresh thread has no messages" as though the two are equivalent, and they are only equivalent if nothing can empty a thread.

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -694,7 +694,11 @@ export class Runner {
         threads.resumeExisting(rawId);
         tid = rawId;
         isResumption = true;
-      } else if (opts.session !== undefined) {
+      } else if (opts.session !== undefined && opts.session !== "") {
```
egonSchiele, line 697: Agree with this change — it is the right fix, and it removes real duplication from every agent.

Two asks, both because this is a semantic change to a language construct rather than to stdlib:

1. **No TS-level test.** The only coverage is `tests/agency/agents/threadIsNew.agency`, which asserts the agent-visible behavior. A unit test next to the other thread-entry-mode tests, asserting that `session: ""` takes the `create()` branch and `session: "x"` takes `openSession`, would pin the runner's branch directly and fail loudly if someone reorders these conditions later.

2. **Not documented outside the code comment.** I could not find the empty-string rule in `docs/site/guide/` or in `stdlib/thread.agency`'s docs. Someone writing `thread(session: userSuppliedName)` where the name happens to be empty now silently gets an isolated thread instead of a shared one. That is the behavior you want, but it should be findable without reading `runner.ts`.

Also worth a moment's thought: is empty-string-means-absent the rule you want long term, or should the agents pass `null`? `opts.session !== undefined` was already the "absent" check, so `null` would fit the existing shape without giving `""` a special meaning. Not a blocker — `""` is the natural default for a `string` param and I do not think anyone will trip on it — but it is the kind of choice that is cheap now and expensive after it ships.

packages/agency-lang/stdlib/supervise.agency
```diff
@@ -0,0 +1,75 @@
...
+    // label gate we would answer an inner trip, extend the wrong guard, and
+    // corrupt our own elapsed-time accounting.
+    if (intr.effect != "std::guard" || intr.data.label != "std::supervise") {
+      return pass()
+    }
+    elapsed = elapsed + interval
+    if (elapsed >= maxTime) {
+      return reject("supervise: maxTime reached")
+    }
+    const decision = check(elapsed)
```
egonSchiele, line 67: Minor, but the value passed to `check` is not measured elapsed time — it is the sum of granted intervals. Those diverge whenever the guard trips for any reason other than exhausting exactly its budget, and they diverge cumulatively across several redirects.

`guardScope.snapshot()` already puts the real number in the interrupt data (`spent: trippedMember.spentAmount()`, `lib/runtime/guardScope.ts:141`), so `check(intr.data.spent)` would hand the callback something true. Since `check` is the user-facing extension point and its parameter is named `elapsed`, a caller writing `if (elapsed > 20m) { give up }` will reasonably assume it is wall-clock.

If you keep the running sum, renaming the parameter to something like `granted` or `budgetUsed` in both the type and the docstring would at least stop it from reading as a measurement.

One other thing on this handler: it uses `match(decision.action)` at line 69, while the module doc above warns that a `match` inside a block that pauses and resumes evaluates to `null`. I think this one is safe — the handler runs *during* the pause rather than being resumed through it, and `matchOutsideResumedBlock` in the pinned test covers the analogous shape. But given that a silent `null` here would mean supervise never approves and every supervised block dies at its first interval, it seems worth one test node asserting the redirect arm's return value specifically, rather than inferring it from `redirectDeliversMessage` passing.

egonSchiele (commented):
Reviewed against the plan and its review round. This is a strong execution — every blocking finding from the plan review landed, and several landed better than proposed.

What got fixed well:

- `llmOptions` builds a fresh object instead of mutating its argument, and takes `tools`/`hostedTools` directly, so the caveat and the `{}` literal at every call site both disappeared.
- `shouldSendSystemPrompt` and its process-lifetime array are gone entirely, replaced by `threadIsNew()`. That is the right altitude — the fact lives where the threads live.
- `thread(session: "")` meaning "no session" removes the seven-way branch duplication the plan would have shipped.
- `WriteFailure.problems` is populated at every site, and the guard-trip path no longer stuffs a message into the `source` field.
- The supervise handler is gated on `intr.data.label`, and `innerGuardTripPassesThrough` tests exactly the failure mode the gate prevents. `redirectDeliversMessage` asserts real delivery rather than just the approve shape.

Three things I would want resolved before merge, one of them substantive:

**1. The salvage path in `agencyCodingAgent` looks unverified and may not work.** Details inline. Short version: `saveDraft` is called in `generateLoop`, a callee frame, and `docs/dev/saveDraft.md` says a caller receiving an aborted result drops the callee's partial unless it has a `finalize` to read the local. The guard block here has no `finalize`. `verifierAgent` uses `finalize` for what looks like exactly this reason, with a comment saying a `saveDraft` would not work there — so the distinction is understood elsewhere in the same PR. No test trips a guard mid-generation, so nothing would catch it if the comment's claim is wrong.

**2. `concurrencyPool` changed from work-stealing to static round-robin.** A real bug fix, but it trades away dynamic load balancing, and the one caller in this PR fans out LLM judge calls whose costs are very uneven. Worth a docstring note at minimum. Inline.

**3. The pinned `match`-returns-null bug.** Correctly diagnosed, correctly recorded in TODO.md, correctly worked around. My concern is that the workaround is a convention nothing enforces: the next person to write `match(captured)` in an agent gets a silent `null`, and the plan's own standard-contract section specified `match`. Since this is interrupt/resume machinery corrupting expression evaluation, and CLAUDE.md treats anything in that neighborhood as critical, is a lint rule or a checker diagnostic (AG-something on `match` inside a guard/supervise block) worth filing alongside the TODO entry? Not a merge blocker, but I would not want the only defense to be the comments in these six files.

The tests are notably better than the plan asked for — `agencyCodingRepair` covering both the type-error and parse-error branches, the honest mock caveat about not proving the diagnostics reached the second prompt, and the inner-guard pass-through node are all things the plan did not specify.

PR #592, opened 2026-07-19
Give the stdlib agents the tools their jobs require, add a data agent, and ship per-agent skills

egonSchiele:
Follow-up to the agents refactor. Implements `docs/superpowers/specs/2026-07-18-agent-tools-and-skills-design.md`.

The last PR fixed the *structure* of the stdlib agents. This one fixes what they can actually **do**.

## Why

Once every agent had the same shape, their tool lists could be compared side by side for the first time, and it turned out nobody had ever chosen them as a set. They had accumulated:

- The **Agency verifier ran a program and judged its return value while blind to any file the program wrote.** A program that returns `"done"` and writes nothing passed. It was verifying the wrong artifact.
- The **researcher had no keyless search.** Provider-hosted `web_search` runs server-side, needs no API key, and works in sandboxes; the expert already used it, and the agent whose whole job is finding things did not. It could also find a Wikipedia page and then not read it, having imported `search` but not `summary` or `article`.
- The **Agency writer could not look up the `AG####` codes its own retry prompt quoted at it**, and could not typecheck a draft before spending a whole attempt on it.
- The **verifier's prompt demanded computed values and byte-level checks** it had no `bash` and no `readBinary` to make.
- The **coding agent could search the web but not fetch what it found.**

## What changed

**Named tool bundles** (`std::agents/lib/toolkits`). Nine functions — `readOnlyFileTools`, `writableFileTools`, `shellTools`, `gitTools`, `webTools`, `agencyDocTools`, `agencyCodeTools`, `memoryTools`, `planningTools`. Every agent now composes its list from these, so a tool list reads as a description of the agent and adding a tool to a category reaches everyone who claimed it. Functions rather than constants, matching `searchTools()`, which must read API keys at call time.

**Every gap above is closed**, plus both reviewers gained web search and fetch so they can check a claim, while still having nothing that changes anything. The boundary is restated in their module docs: **review reads the work and looks things up; the verifier runs the work and measures it.**

One deliberate asymmetry: `agencyCodingAgent` gets targeted fetch but **no** web search. It writes one language whose authoritative docs ship in this package, and an open-ended search returns other languages' idioms — the exact failure its system prompt spends its length fighting. Fetching a page the caller named has no discovery step to go wrong.

**`dataAgent`** (`std::agents/data`) owns the nine data connectors, which no agent could previously reach: FRED, DBnomics, EDGAR, GDELT, LittleSis, Hacker News, YC, USAspending, Wikidata. One agent rather than several by domain — splitting would produce agents differing by tool list rather than by problem, and a caller often can't tell which domain their question falls in. It carries search and fetch too, because querying these APIs means finding an identifier first, and a guessed identifier returns a real answer to the wrong question.

**Packaged skills.** `agentSkill(name)` serves markdown we ship under `stdlib/agents/skills/<agent>/`. A prompt is paid for on every call, which caps how much we can teach an agent; a skill is read only when the model decides it's relevant. These are *ours* — we author them, they ship in the package, no agent takes a skills parameter or scans a user directory. This pass ships the loader, a README per directory, and one worked skill (`data/finding-identifiers.md`), with further content deferred until we have observed failures to write against.

**`extraTools`** on every agent, last in the parameter order. The single seam for anything we don't ship by default: a user's tools, MCP tools, a connector an agent doesn't carry, or another agent wired in deliberately with its own budget. No agent gets another agent by default.

## Two traps this would have hit

**`agentSkill` mirrors `docsSkill`, not `skillsDir` — and not for the reason it looks like.** The approval interrupt lives in `skillsDir`'s body; `docsSkill` skips it by simply not raising it. The `true` that `docsSkill` passes to `buildSkillsTool` is the **`recursive`** flag, not a trust flag. The spec originally said otherwise; reading the source fixed it. Every read the model performs still raises `std::read` through policies.

**Packaged paths must resolve through `getStdlibDir()`.** A relative `"./skills/coding"` resolves against the calling file and breaks once installed into `node_modules` — working perfectly in the repo and failing for every real user. Hence `_agentSkillsDir()` alongside `_docsDir()`. Relatedly, `package.json` shipped markdown only from `./stdlib/docs/**/*.md`, so the skills would have been missing for installed users; `files` is widened and `npm pack --dry-run` confirms all 12 files ship.

## Drive-by fixes

- **`plannerAgent` was never fully on the standard contract**: no `session`, and its drafting call ran bare inside the guard rather than in a thread. Both fixed.
- **Docs skill tool names embedded the absolute install path** (`es_agent_tools_and_skills_..._stdlib_docs_guide`), so they differed per machine and were truncated to 64 chars. They now carry stable names (`agency_guide`, `agency_diagnostics`, ...).

## Verification

- `make` and `make doc` green; `pnpm run lint:structure` clean.
- **8670 unit tests** pass (545 files).
- **76 agency execution tests** pass across 27 files, including 5 new files: `toolkits` (bundle contents, duplicate-name checks), `extraTools` (a tool passed in is actually callable, via a `toolCall` mock), `toolGaps` (one assertion per closed gap, plus no agent offering duplicate tool names), `data`, and `agentSkill` (the test that catches the packaging trap).
- Diff audited against `docs/dev/anti-patterns.md`.

## Known limitations

- **Skill content is one worked example.** The directories are otherwise READMEs. That is the intended follow-up, starting with per-connector skills for `dataAgent`.
- **Tool count grew**: `codingAgent` to ~45, `dataAgent` ~31. No evidence of a selection problem at this size, and `codingAgent` already ran at 38. If one appears, the remedy discussed is trimming the git defaults (17 of those tools are git), not splitting agents — a `gitAgent` that merely proxies git commands would add a round trip and lose the caller's intent.
- **Hosted `web_search` is requested unconditionally, and is not gated on the model.** An earlier version of this PR body claimed these agents "degrade to client-side `searchTools()`" — that was wrong, and there is no such fallback: `hostedTools` passes straight through `llmClient.ts` to smoltalk with nothing gating it. A real gate was built and then removed, because the model catalog keys hosted tools by provider API rather than by model, so it reported "unsupported" for every model and would have silently disabled web search everywhere. The capability is now requested from one place (`hostedSearchTools()` in `std::agents/lib/search`) with that reasoning recorded at the definition. Whether an unsupported `hostedTools` value hard-fails a call on a given provider is still unverified.
- The two pre-existing diagnostics in `lib/agents/review/agent.agency` and `lib/agents/policy/agent.agency` are untouched and identical on `main`.

### DIFF
--- a/packages/agency-lang/lib/stdlib/skills.ts
+++ b/packages/agency-lang/lib/stdlib/skills.ts
@@ -36,3 +36,13 @@ export function _readSkill(filepath: string): string {
 export function _docsDir(section: "guide" | "cli" | "diagnostics" | "stdlib"): string {
   return path.join(getStdlibDir(), "docs", section);
 }
+
+/**
+ * Absolute path to the skills we ship for one agent, under
+ * `stdlib/agents/skills/<agent>`. Resolved through getStdlibDir for the
+ * same reason as _docsDir: a path relative to the calling file works in the
+ * repo and breaks once the package is installed into node_modules.
+ */
+export function _agentSkillsDir(agent: string): string {
+  return path.join(getStdlibDir(), "agents", "skills", agent);
+}
--- a/packages/agency-lang/lib/stdlib/skills.ts
+++ b/packages/agency-lang/lib/stdlib/skills.ts
@@ -42,7 +42,22 @@ export function _docsDir(section: "guide" | "cli" | "diagnostics" | "stdlib"): s
  * `stdlib/agents/skills/<agent>`. Resolved through getStdlibDir for the
  * same reason as _docsDir: a path relative to the calling file works in the
  * repo and breaks once the package is installed into node_modules.
+ *
+ * The name is confined to that directory. `agentSkill` deliberately skips
+ * the approval interrupt that `skillsDir` raises, on the grounds that these
+ * files ship inside the package, so an `agent` of `"../../.."` would turn a
+ * trusted scan into an unprompted scan of somewhere else entirely. Confining
+ * here keeps that trust argument true.
  */
 export function _agentSkillsDir(agent: string): string {
-  return path.join(getStdlibDir(), "agents", "skills", agent);
+  const root = path.join(getStdlibDir(), "agents", "skills");
+  const resolved = path.resolve(root, agent);
+  const rootWithSep = root.endsWith(path.sep) ? root : root + path.sep;
+  if (resolved !== root && !resolved.startsWith(rootWithSep)) {
+    throw new Error(
+      `agentSkill: '${agent}' resolves outside the shipped skills directory. ` +
+        `Skill names are relative paths under stdlib/agents/skills.`,
+    );
+  }
+  return resolved;
 }
--- a/packages/agency-lang/stdlib/agents/lib/toolkits.agency
+++ b/packages/agency-lang/stdlib/agents/lib/toolkits.agency
@@ -1,0 +1,161 @@
+/** @module
+  @summary Named bundles of tools that agents compose into their tool lists.
+
+  An agent's tool list should read as a description of what that agent may
+  do. Bundling also means a tool added to a category reaches every agent that
+  claimed the category, instead of drifting between hand-maintained arrays.
+
+  Every bundle is a function rather than a constant, matching searchTools() in
+  std::agents/lib/search, which must be a function because it reads API keys
+  at call time. An agent whose list includes searchTools() must build that
+  list inside a function too, or it freezes the environment at module load.
+*/
+import { typecheck, parseAST } from "std::agency"
+import { todoWrite, todoList } from "std::agent"
+import { edit } from "std::fs"
+import {
+  gitStatus,
+  gitLog,
+  gitDiff,
+  gitShow,
+  gitBranchList,
+  gitRemoteList,
+  gitBlame,
+  gitStashList,
+  gitAdd,
+  gitCommit,
+  gitCheckout,
+  gitSwitch,
+  gitBranchCreate,
+  gitBranchDelete,
+  gitStashPush,
+  gitStashPop,
+  gitRestore,
+ } from "std::git"
+import { fetch, fetchJSON, fetchMarkdown } from "std::http"
+import { remember, recall } from "std::memory"
+import { bash, ls, glob, grep, exec } from "std::shell"
+import { docsSkill } from "std::skills"
+import {
+  search as wikipediaSearch,
+  summary as wikipediaSummary,
+  article as wikipediaArticle,
+ } from "std::wikipedia"
+
+// Bundled Agency docs. `static` caches each tool across calls. Each gets an
+// explicit name: the name docsSkill derives from a directory embeds the
+// absolute install path, which differs per machine and is truncated to fit.
+static const docSkill = docsSkill("guide").rename("agency_guide")
+static const cliSkill = docsSkill("cli").rename("agency_cli")
+static const diagnosticsSkill = docsSkill("diagnostics").rename("agency_diagnostics")
+static const stdlibSkill = docsSkill("stdlib").rename("agency_stdlib")
+
+export def readOnlyFileTools(): any[] {
+  """
+  Return tools that inspect the file system without changing it: read, list,
+  glob, and grep, resolved against the agent working directory.
+  """
+  return [
+    read.partial(useAgentCwd: true),
+    ls.partial(useAgentCwd: true),
+    glob.partial(useAgentCwd: true),
+    grep.partial(useAgentCwd: true),
+  ]
+}
+
+export def writableFileTools(): any[] {
+  """
+  Return the read-only file tools plus write and edit.
+  """
+  return [
+    ...readOnlyFileTools(),
+    write.partial(useAgentCwd: true),
+    edit.partial(useAgentCwd: true),
+  ]
+}
+
+export def shellTools(): any[] {
+  """
+  Return tools that run commands: bash for a shell pipeline, exec for a
+  single binary with arguments.
+  """
+  return [
+    bash.partial(useAgentCwd: true),
+    exec.partial(useAgentCwd: true),
+  ]
+}
+
+export def gitTools(): any[] {
+  """
+  Return the git tools. The read-only ones run without an approval prompt;
+  the ones that change the repository prompt for approval.
+  """
+  return [
+    gitStatus,
+    gitLog,
+    gitDiff,
+    gitShow,
+    gitBranchList,
+    gitRemoteList,
+    gitBlame,
+    gitStashList,
+    gitAdd,
+    gitCommit,
+    gitCheckout,
+    gitSwitch,
+    gitBranchCreate,
+    gitBranchDelete,
+    gitStashPush,
+    gitStashPop,
+    gitRestore,
+  ]
+}
+
+export def webTools(): any[] {
+  """
+  Return tools that retrieve a named web resource: HTTP fetches and Wikipedia
+  lookups. These retrieve something you can already name; to discover a
+  source you cannot name yet, add a search tool as well.
+  """
+  return [
+    fetch,
+    fetchJSON,
+    fetchMarkdown,
+    wikipediaSearch.rename("wikipedia_search"),
+    wikipediaSummary.rename("wikipedia_summary"),
+    wikipediaArticle.rename("wikipedia_article"),
+  ]
+}
+
+export def agencyDocTools(): any[] {
+  """
+  Return the bundled Agency documentation tools: the language guide, the CLI
+  reference, the type-checker diagnostic codes, and the standard-library
+  reference.
+  """
+  return [docSkill, cliSkill, diagnosticsSkill, stdlibSkill]
+}
+
+export def agencyCodeTools(): any[] {
+  """
+  Return tools that inspect Agency source without running it: the type
+  checker and the parser.
+  """
+  return [typecheck, parseAST]
+}
+
+export def memoryTools(): any[] {
+  """
+  Return tools that persist and retrieve facts across sessions.
+  """
+  return [remember, recall]
+}
+
+export def planningTools(): any[] {
+  """
+  Return tools an agent uses to organize a long run: a todo list, and
+  saveDraft for checkpointing partial work so an aborted run still returns
+  something.
+  """
+  return [todoWrite, todoList, saveDraft]
+}
--- a/packages/agency-lang/stdlib/agents/researcher.agency
+++ b/packages/agency-lang/stdlib/agents/researcher.agency
@@ -70,6 +70,7 @@ def researchLoop(
   maxAttempts: number,
   model: string,
   provider: string,
+  extraTools: any[],
 ): string {
   const researchTools = buildTools()
   if (threadIsNew()) {
@@ -91,7 +92,11 @@ def researchLoop(
     }
     const answer = llm(
       msg,
-      llmOptions(model: model, provider: provider, tools: researchTools),
+      llmOptions(
+        model: model,
+        provider: provider,
+        tools: researchTools.concat(extraTools),
+      ),
     )
     saveDraft(answer)
     return answer
@@ -107,6 +112,7 @@ export def researcherAgent(
   model: string = "",
   provider: string = "",
   session: string = "",
+  extraTools: any[] = [],
 ): Result<string> {
   """
   Answer a research question from web and Wikipedia sources and return a
@@ -120,6 +126,7 @@ export def researcherAgent(
   @param model - Model override, or "" for the ambient model
   @param provider - Provider for the model override
   @param session - Session name to share a thread across calls, or "" for isolated
+  @param extraTools - Extra tools to offer the LLM, appended to the built-in set
   """
   // The guard Result is the return value: a success carries the answer (or,
   // when a trip was rejected, the best draft so far), and a failure carries
@@ -133,6 +140,7 @@ export def researcherAgent(
         maxAttempts: maxAttempts,
         model: model,
         provider: provider,
+        extraTools: extraTools,
       )
     }
     return answer
--- a/packages/agency-lang/stdlib/agents/researcher.agency
+++ b/packages/agency-lang/stdlib/agents/researcher.agency
@@ -35,7 +35,8 @@ static const skills = agentSkill("researcher")
 export def buildTools(): any[] {
   """
   Return the researcher's tools: the encyclopedia and fetch tools that need
-  no key, plus whichever search providers the environment has a key for.
+  no key, a local file read for material the caller points at, and whichever
+  search providers the environment has a key for.
   """
   return [...webTools(), read.partial(useAgentCwd: true), ...searchTools(), skills]
 }
--- a/packages/agency-lang/stdlib/agents/researcher.agency
+++ b/packages/agency-lang/stdlib/agents/researcher.agency
@@ -10,10 +10,9 @@ import { reviewAgent } from "std::agents/review"
 import { feedbackHasErrors, renderFeedback } from "std::agents/lib/feedback"
 import { searchTools } from "std::agents/lib/search"
 import { llmOptions, withContext } from "std::agents/lib/shared"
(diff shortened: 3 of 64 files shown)


egonSchiele (commented):
Reviewed against the spec and its review round. The two things I flagged as blocking or gap-level in the spec review both landed well:

- **`useAgentCwd` is preserved.** `readOnlyFileTools()`, `writableFileTools()`, and `shellTools()` all apply `.partial(useAgentCwd: true)` to every member, so composing from bundles does not silently move agents onto the process cwd. This was the one thing I did not want to see ship, and it is handled.
- **`agentSkill` gets an explicit stable name** (`skills_` + the agent path) rather than one derived from the install directory. Finding that the same bug already existed in `docsSkill` — tool names embedding the absolute install path and getting truncated at 64 chars — and fixing it in the same pass is a good catch that went beyond what was asked.

The `dataAgent` system prompt is the strongest prose in the PR. "A wrong identifier returns a real answer to the wrong question, which is far worse than an error" is the kind of sentence that actually changes model behavior, and pairing it with "NEVER estimate a number and present it as retrieved" covers the failure mode I was worried about with FRED's missing-key path.

Two findings I would want addressed before merge, one of them an altitude question:

**1. The researcher's grounding loop now runs a web-searching reviewer on every attempt.** `reviewAgent` went from toolless to carrying fetch, Wikipedia, client search, and hosted search — and `researcherAgent` calls it once per attempt inside its own guard. Details inline. This is a cost and latency change to an existing inner loop that I do not think anyone chose.

**2. Hosted `web_search` is set unconditionally at six call sites, and the repo already solves this.** `lib/agents/agency-agent/lib/search.agency` has `HOSTED_SEARCH_CAPS_BY_BACKEND` and `withWebSearch(opts)`, documented as "the one declarative interface every subagent uses to opt into web search," dispatching on the active backend so an `off` or non-hosted backend falls through naturally. The stdlib agents do not use it. The PR body says agents "degrade to client-side `searchTools()`" but I could not find code implementing that degradation — `hostedTools` passes straight through `llmClient.ts:240`. This was the spec's own open risk, marked "worth an explicit check during implementation." Inline.

Plus a test-coverage gap: the bundles now bind `useAgentCwd` correctly, but nothing asserts it, so the regression I was worried about can still reappear silently. Inline.

On the review-agents-gain-search decision: you took the ship-as-written option from the spec review rather than splitting it, which is your call and I do not object. If it stays as written, it would help to note in the PR or the module doc what would falsify the bet — "if review cost per call rises above X, or reviews start citing pages instead of judging the work, revisit" — so that if it does go wrong, someone knows to look here rather than at the model.

Minor: `extraTools` correctly does not propagate to sub-agent calls (`checkGrounding` calls `reviewAgent` without it), which I think is right, but it is not stated anywhere. One sentence in the `extraTools` docstring would stop a future implementer from "fixing" it.

packages/agency-lang/stdlib/agents/researcher.agency
```diff
@@ -89,9 +93,17 @@ def researchLoop(
...
+    // where outbound HTTP is blocked, so it complements the client-side
+    // providers in researchTools rather than replacing them.
     const answer = llm(
       msg,
-      llmOptions(model: model, provider: provider, tools: researchTools),
+      llmOptions(
+        model: model,
+        provider: provider,
+        tools: researchTools.concat(extraTools),
+        hostedTools: ["web_search"],
```
egonSchiele: Anchoring here, but this is about the `reviewAgent` call in `checkGrounding` just above (line 50), which this hunk's sibling changes make expensive.

That call is the one place where the review agents' new tools have a compounding effect, and I do not think it was considered.

Before this PR, `reviewAgent` was toolless, so `checkGrounding` was a single cheap LLM call: hand it the answer text, ask whether claims are cited. Now `reviewAgent` carries `webTools()` (fetch, fetchJSON, fetchMarkdown, three Wikipedia tools), `searchTools()`, and hosted `web_search` — and it runs with its own default budget of `$10.00` / `10m`.

Three consequences stack up:

1. **It runs per attempt.** `maxAttempts` defaults to 2, so a researcher run can spend up to `$20` of its `$50` budget on grounding checks that used to cost cents. The reviewer's guard is nested inside the researcher's, so it charges the same budget.
2. **The job changed shape.** The context passed here is `groundingInstruction`: "error=true if a claim is ungrounded or uncited." That is a text-inspection question — does this answer cite its sources — and it is answerable from the text alone. A reviewer holding fetch and search will now go *verify* the citations by retrieving them. That may well be better review quality, but it is a different and much more expensive operation than the one this call site asks for, and nothing in the prompt tells it not to.
3. **The researcher searches the web, then the reviewer searches the web to check it.** The spec's `extraTools` section is careful that "no agent gets another agent by default" precisely to avoid surprise recursion and cost blowups. That reasoning applies here — the agent-calling-agent edge already existed, and this PR added the expensive tools underneath it without revisiting the edge.

The cheapest fix is to cap the inner call, since this call site knows the check should be cheap:

```agency
const reviewed = reviewAgent(
  work: "${answer}",
  task: task,
  context: groundingInstruction,
  maxCost: $1.00,
  maxTime: 2m,
)
```

That keeps the reviewer's new ability to look something up when a claim is genuinely checkable, while stopping it from turning every grounding check into a research session. Worth confirming the same reasoning for `plannerAgent`, which calls `verifierAgent` at line 161 and `agencyCodingAgent` at line 199 with default budgets.

packages/agency-lang/stdlib/agents/review.agency
```diff
@@ -29,7 +47,15 @@ ${withContext(task: task, context: context)}
...
 ${work}
 """
-  const items: Feedback[] = llm(prompt, llmOptions(model: model, provider: provider))
+  const items: Feedback[] = llm(
+    prompt,
+    llmOptions(
+      model: model,
+      provider: provider,
+      tools: buildTools().concat(extraTools),
+      hostedTools: ["web_search"],
```
egonSchiele: `hostedTools: ["web_search"]` is set unconditionally here and at five other call sites (`researcher.agency:105`, `coding.agency:91`, `agency/review.agency:88`, `expert.agency:63`, `data.agency:94`), with no check that the active provider offers hosted search.

The PR body says these agents "degrade to client-side `searchTools()`," but I could not find the code that does the degrading. `hostedTools` is carried through `lib/runtime/llmClient.ts:240` straight to smoltalk. If a provider or local model rejects an unknown hosted tool, the `llm()` call fails rather than falling back — which would take out the agent entirely, not just its search. Happy to be shown the path I missed.

More to the point: **the repo already has a worked solution to this exact problem.** `lib/agents/agency-agent/lib/search.agency:325-388`:

```agency
static const HOSTED_SEARCH_CAPS_BY_BACKEND: Record<string, string[]> = {
  hosted: ["web_search"]
}

export def withWebSearch(opts: LlmOpts): LlmOpts { ... }
```

Its own comment calls it "the one declarative interface every subagent uses to opt into web search": subagents say what they want, the active backend is looked up in two tables, and `off` "naturally falls through (it's in neither table) and returns `opts` unchanged." That is precisely the degradation the PR body describes, already written and already in use.

So after this PR there are two parallel web-search abstractions with different conventions. The agency-agent one renames both providers to a single stable `web_search` specifically so "subagent prompts can refer to a single tool name." The new stdlib one renames them to `web_search_tavily` and `web_search_brave`. Neither is wrong, but a model moving between the two sees different tool names for the same capability.

Since the whole point of `toolkits.agency` is that "a tool added to a category reaches every agent that claimed the category, instead of drifting between hand-maintained arrays," having hosted search hardcoded at six call sites is the same drift one level up. Could the hosted capability move into `std::agents/lib/search` as a `hostedSearchTools()` or a `withWebSearch`-style helper, gated on backend the way the agency-agent version is — and ideally with the agency-agent version eventually delegating to it, so there is one table rather than two?

If the answer is "the two serve different purposes, the agency-agent one is user-switchable via `/search` and the stdlib one is key-presence-based," that is fair, but it is worth a comment saying so at one of these call sites. Otherwise the next person to touch either will reasonably assume the other is dead code.

packages/agency-lang/stdlib/agents/lib/toolkits.agency
```diff
@@ -0,0 +1,161 @@
...
+
+// Bundled Agency docs. `static` caches each tool across calls. Each gets an
+// explicit name: the name docsSkill derives from a directory embeds the
+// absolute install path, which differs per machine and is truncated to fit.
+static const docSkill = docsSkill("guide").rename("agency_guide")
+static const cliSkill = docsSkill("cli").rename("agency_cli")
+static const diagnosticsSkill = docsSkill("diagnostics").rename("agency_diagnostics")
+static const stdlibSkill = docsSkill("stdlib").rename("agency_stdlib")
+
+export def readOnlyFileTools(): any[] {
```
egonSchiele, line 53: Glad to see `.partial(useAgentCwd: true)` applied to every member here — that was the thing I most wanted from this file.

The gap is that nothing in `tests/agency/agents/toolkits.agency` asserts it. The test file checks bundle non-emptiness, arity, duplicate names, and the wikipedia and docs renames — all good, and the duplicate-name test is well motivated. But a refactor that drops `.partial(useAgentCwd: true)` from these three bundles passes every existing test, and the resulting bug is invisible in CI because tests run from the repo root where the agent cwd and the process cwd usually coincide. It only shows up as an agent reading or writing outside the directory it was pointed at.

Suggested addition, alongside the existing checks:

```agency
// Composing from a bundle must not silently move an agent onto the process
// cwd: every file and shell tool is bound to the agent working directory.
node fileAndShellToolsAreCwdBound(): boolean {
  const bound = [...writableFileTools(), ...shellTools()]
  return every(bound, \tool -> tool.boundArgs.useAgentCwd == true)
}
```

I am guessing at the accessor for a partial's bound arguments — whatever the real one is, the assertion is what matters.

Separately, `writableIsASupersetOfReadOnly` (test line 31) checks arity rather than membership:

```agency
return writableFileTools().length == readOnlyFileTools().length + 2
```

That passes if someone drops `grep` from the writable set and adds two unrelated tools. Since the test is named for a superset property, asserting it by name would be stronger and no harder:

```agency
const writableNames = map(writableFileTools(), \tool -> tool.name)
return every(readOnlyFileTools(), \tool -> writableNames.includes(tool.name))
```

packages/agency-lang/stdlib/agents/researcher.agency
egonSchiele: Confirmed and fixed. The nesting is exactly as you describe: `checkGrounding` runs once per attempt inside `revise`, inside the researcher's guard, so the reviewer's `$10.00` default is charged to the researcher's `$50.00`.

Capped at the call site, since this one knows the check should be cheap:

```agency
const reviewed = reviewAgent(
  work: "${answer}",
  task: task,
  context: groundingInstruction,
  maxCost: $1.00,
  maxTime: 2m,
)
```

Your point 2 is the one I had not considered. The context says "error=true if a claim is ungrounded or uncited", which is answerable from the text, but a reviewer holding fetch will go retrieve the citations to check them. The cap bounds that without forbidding it, so a genuinely checkable claim can still be checked.

Checked the planner too, and both calls you flagged had the same shape: `verifierAgent` at line 161 and `agencyCodingAgent` at line 199, each at their own defaults inside the planner's `$100`, each running once per plan attempt. Capped at `$5.00`/5m and `$10.00`/10m respectively.

packages/agency-lang/stdlib/agents/review.agency
egonSchiele: You did not miss a path: there is no degradation code. `hostedTools` goes from `llmClient.ts:240` straight into the smoltalk config with nothing gating it. The PR body's claim that these agents "degrade to client-side `searchTools()`" was my overclaim, not a description of the code. Corrected below and in the PR body.

Consolidated as you suggested. `hostedSearchTools()` now lives in `std::agents/lib/search`, all six call sites use it, and it carries the comment explaining the relationship to the agency-agent version.

On unifying the tool names: I want to push back gently, because the difference is load-bearing rather than accidental. `withWebSearch` has exactly one backend active, so renaming whichever provider is live to a single `web_search` is safe. `searchTools()` keys off key presence and returns *every* provider that has one, so with both `TAVILY_API_KEY` and `BRAVE_API_KEY` set it returns two tools. Under a shared name that is two tools called `web_search` in one list, which providers reject. That is the same duplicate-name failure the `noBundleHasDuplicateNames` test exists to catch. Both files now say this, so the next reader does not have to rediscover it.

On gating: **I built it, and it was wrong, so I removed it.** Worth recording why, because it looked implementable.

`std::llm` already has `modelSupportsInput` with exactly the right tri-state shape, and smoltalk exposes `getHostedTools({ model })`, so I added `modelSupportsHostedTool` and gated on it. The test I wrote to pin the behavior failed, and the probe showed why:

```
getHostedTools()                          -> 12 entries, keyed by provider:
                                             {name: "web_search", provider: "anthropic"}
                                             {name: "web_search", provider: "openai-responses"}
getHostedTools({model: "gpt-4o-mini"})    -> []
getHostedTools({model: "gpt-4o"})         -> []
```

The catalog keys hosted tools by **provider API**, not by model, and the entries carry no per-model list. So the gate answered "unsupported" for every model, and shipping it would have silently disabled web search across every agent. Mapping model to provider API by hand fails the same way the first time a provider renames an API — note `openai-responses` is not the provider string a model carries.

So `hostedSearchTools()` requests unconditionally, with the reasoning in a comment: the cost of asking a provider that cannot deliver is that provider's error, which is visible; the cost of withholding wrongly is an agent that quietly stops searching, which is not. Until the catalog can answer per model, I would rather have the visible failure.

I still cannot tell you whether an unsupported `hostedTools` value hard-fails a call or is ignored, and I am not going to assert it a second time without evidence. If you know the answer for local models, that decides whether this needs a real gate soon or can wait for better catalog data.

packages/agency-lang/stdlib/agents/lib/toolkits.agency
egonSchiele: Both added.

The accessor is `tool.boundArgs`, which returns `{indices, values, originalParams}` rather than a name-keyed map, so the check looks `useAgentCwd` up by name in `originalParams` and reads the value bound at that position:

```agency
node fileAndShellToolsAreCwdBound(): boolean {
  const shouldBeBound = [...writableFileTools(), ...shellTools()]
  return every(shouldBeBound, isCwdBound)
}
```

I mutation-checked it rather than trusting that it passes: dropping `.partial(useAgentCwd: true)` from `bash` alone takes the file from 6/6 to 5/6, and restoring it returns to 6/6. Your point about CI hiding this is why I bothered — a test that passes both before and after the bug would have been worse than no test.

`writableIsASupersetOfReadOnly` now asserts membership by name and keeps the arity check as a second clause, so it fails both when a read-only tool goes missing and when the writable set grows unexpectedly.

packages/agency-lang/lib/stdlib/skills.ts
```diff
@@ -36,3 +36,13 @@ export function _readSkill(filepath: string): string {
...
+
+/**
+ * Absolute path to the skills we ship for one agent, under
+ * `stdlib/agents/skills/<agent>`. Resolved through getStdlibDir for the
+ * same reason as _docsDir: a path relative to the calling file works in the
+ * repo and breaks once the package is installed into node_modules.
+ */
+export function _agentSkillsDir(agent: string): string {
+  return path.join(getStdlibDir(), "agents", "skills", agent);
+}
```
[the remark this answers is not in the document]
egonSchiele: Correct, and it matters more than a general path-hygiene issue would, because `agentSkill` deliberately skips the approval interrupt that `skillsDir` raises. That exemption is justified by the files shipping inside the package; a name of `../../..` would have made it an unprompted scan of somewhere else entirely, which breaks the justification rather than just the path.

`_agentSkillsDir` now resolves the name and rejects anything landing outside the shipped directory:

```ts
const resolved = path.resolve(root, agent);
if (resolved !== root && !resolved.startsWith(rootWithSep)) {
  throw new Error(`agentSkill: '${agent}' resolves outside the shipped skills directory. ...`);
}
```

Covered by `lib/stdlib/skills.agentSkillsDir.test.ts`: plain names, nested names such as `agency/coding`, traversal via `../../..` and `coding/../../../etc`, and absolute paths.

PR #594, opened 2026-07-19
Collapse the duplicate agency-agent subagents onto the stdlib

egonSchiele:
Piece **A** of the agency-agent decomposition: collapse the subagents that duplicate stdlib agents. Mostly deletion — 287 lines out, 62 in.

## What changed

**`verify.agency` deleted.** It converted the verifier's findings into a `Verdict` and unwrapped that fail-open. Both halves are redundant now: `verifierAgent` already fails open, so an empty finding list *is* the "nothing wrong" answer. The code agent calls `verifierAgent` directly and reads it with `feedbackHasErrors` / `renderFeedback`. Its test went with it — the behavior it covered is tested against the stdlib verifier.

**`review.agency` slimmed** to just the agent-shaped part: pulling code snippets out of a chat message, and reviewing the `.agency` files the code agent wrote. Gone: the duplicate `Feedback` type, the two wrappers that re-exported the stdlib helpers under new names, and the comment explaining a re-export workaround it no longer performs.

**`research.agency` takes its tools from the stdlib bundles** rather than listing them by hand, which also drops the local wikipedia rename dance since `webTools()` already renames those.

## One deliberate non-collapse

**`research.agency` is not merged into `researcherAgent`.** They read as duplicates from the outside and are not:

- `researcherAgent` is a research analyst with a grounding retry loop that judges its own answer for citations and retries.
- The agent's version is its *reading specialist*: it answers Agency syntax questions from the bundled docs, holds a conversation, and formats for the REPL.

Fusing them either drops that character or pushes REPL formatting and docs-answering into a stdlib agent that other users would then carry. Worth revisiting when the code agent splits (piece C), since the boundary between "reading specialist" and "researcher" may look different once the interactive/autonomous split lands.

## Verification

`make` green; `toolWiring` (4) and `agentTurn` (29) pass; no dangling references to the deleted module. Kept deliberately light per request — CI covers the rest.


egonSchiele (commented):
## Overview

The deletions are the good part. `verify.agency` was a `Verdict` shim over a `Result<Feedback[]>` that everything else already speaks, `review.agency` was re-exporting stdlib helpers under new names to dodge a codegen ordering bug, and `research.agency` was hand-listing tools the stdlib bundles already assemble (rename dance included). Collapsing all three onto the stdlib is right, and the new `std::agents/agency/researcher` is a clean split from `researcherAgent` — the module doc explaining *why* it has no web tools is exactly the kind of comment worth keeping.

But this PR is not only the deletion its description advertises. It also adds an `agencyTask` / `agencyTopic` routing flag to three subagents and rewires the code agent's verify loop, and those changes carry the risk here.

---

## Blocking: the verify loop flipped from fail-open to fail-closed

`subagents/code.agency`, in the one-shot loop:

```agency
const gaps = verifierAgent(task: originalUserMsg)
if (feedbackHasErrors(gaps)) { ...re-prompt... }
```

`feedbackHasErrors` returns **`true`** when the Result is a failure — that is its documented behavior in `stdlib/agents/lib/feedback.agency`:

```agency
// If the feedback itself is a failure, we can treat that as an error state.
return true
```

The code this replaces did the opposite, deliberately:

```agency
const verdictResult = try verify(originalUserMsg)
const verdict = unwrapVerdict(verdictResult)   // failure -> satisfied: true
```

with the comment on `unwrapVerdict` spelling out the intent: *"failing OPEN on any Failure. Verification can only help, never block."*

So when the verifier itself cannot run — budget exhausted, wall-clock trip, provider error — the loop now reads that as "your work isn't complete", and re-prompts the coding agent with `renderFeedback(gaps)`, which for a failure is the raw error string. The agent gets told to fix "verification could not run: …" as if it were a gap in its work, and burns another round doing it.

The PR body's argument for the deletion is that `verifierAgent` already fails open, so an empty finding list *is* the "nothing wrong" answer. That is true for `success([])` and false for `failure(_)`, and the failure case is the only thing `unwrapVerdict` existed to handle. Worth restoring explicitly:

```agency
const gaps = try verifierAgent(task: originalUserMsg)
if (gaps is success(items) && some(items, \f -> f.error)) { ... }
```

Dropping the `try` is a separate half of the same change and deserves its own look — please confirm nothing on that path can raise now that the wrapper is gone.

## The new flags default to false, and nothing ever sets them true

`codeAgent`, `researchAgent`, and `reviewAgent` all gained a `= false` flag that routes to the Agency-aware agent. The only setter is the coordinator LLM reading the docstring, which is a reasonable design — but `dispatchAgent` (`agent.agency:592`) takes the other path and hardcodes the default:

```agency
"code" => codeAgent(expanded)
"research" => researchAgent(expanded)
"review" => reviewAgent(expanded)
```

That is the explicit-routing path. A user who types the Agency-language target directly, in the Agency REPL, now always gets the general-purpose agent — the one the new module doc for `agencyResearcherAgent` warns "is usually an answer about some other language." For this product the default on that path looks backwards. Either thread the flag through `dispatchAgent`, or default it to true where the caller has already said "Agency".

## `reviewWrittenFiles()` is now conditional, and that is not in the description

```agency
let feedback: Result<Feedback[]> = success([])
if (agencyTask) {
  feedback = reviewWrittenFiles() with approve
}
```

Skipping a parse-and-typecheck review for non-Agency files is defensible. Combined with the point above, though, it means the review silently disappears for Agency work whenever the flag does not get set — and the failure mode is invisible, because `success([])` reads as a clean review rather than a skipped one. Given that this is the code agent's own quality gate, it deserves a mention in the PR body at minimum.

## Description is out of date

The body says `research.agency` "is not merged into `researcherAgent`" and describes it as taking its tools from the stdlib bundles. The diff does merge it: `research.agency` is now a router with no prompt and no `llm()` call of its own, delegating to `researcherAgent` or to a brand-new `stdlib/agents/agency/researcher.agency` that the body never mentions. The `agencyTask`/`agencyTopic` routing and the `reviewWrittenFiles` gating are likewise absent. Since the "one deliberate non-collapse" section is the part a reviewer would lean on hardest, it is worth rewriting before merge.

## Smaller things

- **Test coverage.** `verify-agent.test.json` is deleted and `agencyResearcher` is added, which is a fair trade for the deleted module. Nothing covers what the PR actually adds: the routing flags, or the fail-open path above. That last one is a pure-logic branch — an agency execution test with a failing verifier would pin it without an LLM call.
- **Tool collisions.** I checked `researchTools` (memory + planning + `highlight` + `print`) against both destinations' `buildTools()` and found no duplicate tool names, so the rename hazard the old comment warned about does not reappear. Noting it since the guard comment went away with the rewrite.
- **`saveDraft` is new to the research path** via `planningTools()`. Almost certainly harmless, just flagging it as an unremarked addition.
- The two new comments in `code.agency` explaining the slow-model slot and why `agencyCodingAgent`'s result is handed back rather than written both earn their place — they say why, not what.

## Verdict

The collapse is the right call and most of this is a clear improvement. The verify-loop fail-open regression should be fixed before merge, and the `dispatchAgent` defaults are worth a decision rather than an accident.

PR #595, opened 2026-07-19
Fix phantom LSP errors by giving the prelude one definition

egonSchiele:
## The bug

The editor showed `Function '_guard' is not defined` on files that `agency typecheck` accepted. Same file, two different answers.

Root cause: the compiler and the LSP each kept their own hand-copied list of the names auto-imported from `std::index`, and the LSP copy had drifted. Three names reached the parser template but never the LSP list:

- `_guard`
- `saveDraft`
- `flatten`

The `guard` construct desugars to a `_guard(...)` call before type checking, so any file using `guard` lit up with a phantom error the compiler never saw. Files using `saveDraft` or `flatten` had the same problem.

The LSP cannot just use the parser template: wrapping the source shifts every line and lands the squiggles on the wrong row. So it synthesizes an equivalent import instead. That is fine. Keeping a second copy of the list is what broke.

## The fix

`lib/prelude.ts` is now the only definition of the prelude. The parser template renders it, and the LSP builds its synthetic import from it, so the editor and the compiler cannot disagree about what is in scope.

Collapsing the two lists does not stop that list from disagreeing with the stdlib, so `prelude.test.ts` also checks every prelude name against the real `std::index` exports. That check runs one way only. The prelude is deliberately a subset: `setAgentCwd`, `getAgentCwd`, and `applyAgentCwd` are exported without being in scope everywhere, and a two-way equality check would drag them into every file.

A third test pins the rendered import to a single line. `AGENCY_TEMPLATE_OFFSET` hardcodes how many lines the template adds, and every diagnostic line number is computed by subtracting it, so wrapping the import at 80 columns would silently shift every editor error down a row.

## Verification

- Wrote the failing test first. It failed naming exactly `_guard`, `saveDraft`, and `flatten`, no more and no less, which is what confirmed the root cause instead of assuming it.
- The stdlib-drift test passed on first run, which proves nothing, so I injected a fake name and watched it fail before reverting.
- Ran the reported file, `stdlib/agents/verifier.agency`, through the real LSP path: zero diagnostics.
- `preludeImportLine()` renders byte-identical to the old hardcoded template line, so no Agency file compiles differently.
- `tsc --noEmit` clean. Parser, prelude, and LSP diagnostics tests all pass.

## Note for the reviewer

The full unit suite reports 72 failures across 16 files in a fresh worktree, all `Cannot find package 'agency-lang/...'` from an unbuilt checkout. These are not from this change: stashing the diff and re-running the same 16 files reproduces the identical 72 failures. Leaving them alone as out of scope. CI runs on a built tree and should be clean.

### DIFF
--- a/packages/agency-lang/lib/lsp/diagnostics.test.ts
+++ b/packages/agency-lang/lib/lsp/diagnostics.test.ts
@@ -83,6 +83,46 @@ describe("runDiagnostics — stdlib auto-import parity with CLI", () => {
   });
 });
 
+describe("runDiagnostics — every prelude name resolves", () => {
+  // Regression: the LSP used to keep its own hand-copied list of the names
+  // the CLI parser template auto-imports. The copy drifted — `_guard`,
+  // `saveDraft`, and `flatten` were added to the template but never to the
+  // LSP's list — so a file using the `guard` construct lit up with a phantom
+  // "Function '_guard' is not defined" in the editor while
+  // `agency typecheck` on the same file reported nothing. Both paths now
+  // render the same PRELUDE_NAMES, so this asserts the drifted names in
+  // particular resolve through the LSP.
+  it("reports no undefined-name errors for guard, saveDraft, or flatten", () => {
+    const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "agency-lsp-prelude-"));
+    try {
+      const mainFile = path.join(tmpDir, "main.agency");
+      const source = [
+        "node main() {",
+        "  const captured = guard(cost: 1.0) {",
+        "    saveDraft([1, 2])",
+        "    return [1, 2]",
+        "  }",
+        "  const flat = flatten([[1], [2]])",
+        "  return flat",
+        "}",
+        "",
+      ].join("\n");
+      fs.writeFileSync(mainFile, source);
+
+      const doc = makeDoc(source, `file://${mainFile}`);
+      const symbolTable = SymbolTable.build(mainFile, {});
+      const { diagnostics } = runDiagnostics(doc, mainFile, {}, symbolTable);
+
+      const undefinedNames = diagnostics.filter((d) =>
+        /is not defined/.test(d.message),
+      );
+      expect(undefinedNames.map((d) => d.message)).toEqual([]);
+    } finally {
+      fs.rmSync(tmpDir, { recursive: true, force: true });
+    }
+  });
+});
+
 describe("runDiagnostics — test-only imports", () => {
   // The LSP is an analysis-only path: it must honor `import test` so
   // migrated test files keep full editor support instead of dying on a
--- a/packages/agency-lang/lib/lsp/diagnostics.test.ts
+++ b/packages/agency-lang/lib/lsp/diagnostics.test.ts
@@ -111,12 +111,85 @@ describe("runDiagnostics — every prelude name resolves", () => {
 
       const doc = makeDoc(source, `file://${mainFile}`);
       const symbolTable = SymbolTable.build(mainFile, {});
-      const { diagnostics } = runDiagnostics(doc, mainFile, {}, symbolTable);
+      const { diagnostics, program } = runDiagnostics(doc, mainFile, {}, symbolTable);
 
+      // Without this, a grammar change that stopped the snippet parsing
+      // would leave zero name-resolution diagnostics and the assertion
+      // below would pass while checking nothing.
+      expect(program).not.toBeNull();
       const undefinedNames = diagnostics.filter((d) =>
         /is not defined/.test(d.message),
       );
       expect(undefinedNames.map((d) => d.message)).toEqual([]);
+    } finally {
+      fs.rmSync(tmpDir, { recursive: true, force: true });
+    }
+  });
+});
+
+describe("runDiagnostics — shadowing a prelude name", () => {
+  // Agency treats the prelude as overridable: a user's own `def map` wins,
+  // and the compile path realizes that by dropping `map` from the injected
+  // import (prunePreludeShadows). The LSP has to run the same pass or it
+  // still sees the import and warns about a shadow the compiler resolved.
+  it("does not warn when a user def shadows a prelude name", () => {
+    const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "agency-lsp-shadow-"));
+    try {
+      const mainFile = path.join(tmpDir, "main.agency");
+      const source = [
+        "def map(xs: number[]): number[] {",
+        "  return xs",
+        "}",
+        "node main() {",
+        "  return map([1])",
+        "}",
+        "",
+      ].join("\n");
+      fs.writeFileSync(mainFile, source);
+
+      const doc = makeDoc(source, `file://${mainFile}`);
+      const symbolTable = SymbolTable.build(mainFile, {});
+      const { diagnostics, program } = runDiagnostics(doc, mainFile, {}, symbolTable);
+
+      expect(program).not.toBeNull();
+      const shadowWarnings = diagnostics.filter((d) => /shadows an imported/.test(d.message));
+      expect(shadowWarnings.map((d) => d.message)).toEqual([]);
+    } finally {
+      fs.rmSync(tmpDir, { recursive: true, force: true });
+    }
+  });
+
+  // `_guard` is exempt from pruning (UNPRUNABLE_PRELUDE_NAMES): letting a
+  // user `_guard` displace the real one would silently rebind every
+  // `guard(...) { }` in the file and bypass budget metering. The editor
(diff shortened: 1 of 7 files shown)


egonSchiele (commented):
## Overview

Collapses two hand-copied prelude lists into one (`lib/prelude.ts`), rendered by both the parser template and the LSP's synthetic import. The diagnosis in the PR body is correct and I confirmed the mechanism by reading both paths. Good change — a drift bug fixed by deleting the thing that could drift, not by re-syncing it.

The test set is well chosen. I checked that the single-line test is genuinely load-bearing rather than defensive decoration: `AGENCY_TEMPLATE_OFFSET = 2` (`lib/parsers/parsers.ts:191`) and `lib/parser.ts:269` subtracts it from every span, so a wrapped import really would shift every editor diagnostic by a row.

I ran `lib/prelude.test.ts` and `lib/lsp/diagnostics.test.ts` in the worktree: 16 tests pass.

## Findings

Three comments inline. None are blockers; the first two are cheap, and the third is a follow-up question rather than a defect in this diff.

1. `PRELUDE_NAMES` is handed into the AST by reference — currently harmless, one word to make it structurally safe.
2. The new LSP test can pass vacuously if the snippet ever stops parsing.
3. `prunePreludeShadows` is a third prelude-aware component that the LSP still doesn't run, so one divergence of exactly the class this PR fixes survives. I measured it.

## What I checked and found clean

- No in-place mutation of `importedNames` anywhere in the compile or LSP path. The one write that looked dangerous, `prunePreludeShadows.ts:56`, reassigns the property rather than splicing the array, so the module-global list is not corrupted today.
- `preludeImportLine()` renders byte-identical to the removed hardcoded line — I diffed the name order by eye against the deleted template text; they match, so no Agency file compiles differently.
- The generated `template.ts` is consistent with the `.mustache` source, so `pnpm run templates` was run rather than the generated file hand-edited.

## Risk

Low. The blast radius is wide (this line prefixes every Agency file), but the change is a pure refactor of how one string is produced, and the byte-identical rendering plus the new stdlib-parity test cover the ways it could go wrong.

packages/agency-lang/lib/prelude.ts
```diff
@@ -0,0 +1,58 @@
...
+ * which drifted: `_guard`, `saveDraft`, and `flatten` reached the template
+ * but not the LSP, so any file using the `guard` construct showed a
+ * phantom "Function '_guard' is not defined" in the editor that
+ * `agency typecheck` never reported.
+ *
+ * Every name here must be exported by stdlib/index.agency, and the rendered
+ * import must stay on one line. prelude.test.ts enforces both and explains
+ * why each matters.
+ */
+export const PRELUDE_NAMES: string[] = [
```
egonSchiele: `PRELUDE_NAMES` is a mutable `string[]`, and `diagnostics.ts` puts the array itself into an `ImportStatement` node rather than a copy. So a preprocessor that mutated an import's name list in place would corrupt the prelude for every subsequent file in the same process — and the LSP is a long-lived process, so it would stay corrupted.

I checked whether anything actually does this. Nothing does: `prunePreludeShadows.ts:56` reassigns `spec.importedNames = spec.importedNames.filter(...)` instead of splicing, which leaves the shared array intact. So this is latent, not live.

Still worth closing, because it costs one word and the failure mode is nasty enough that you would not enjoy debugging it:

```ts
export const PRELUDE_NAMES: readonly string[] = [
```

That makes the compiler reject an in-place mutation at the point someone writes it. `ImportStatement.importedNames` is typed `string[]`, so the LSP call site would then need `[...PRELUDE_NAMES]` — which is the copy you want anyway.

packages/agency-lang/lib/lsp/diagnostics.test.ts
```diff
@@ -83,6 +83,46 @@ describe("runDiagnostics — stdlib auto-import parity with CLI", () => {
...
+        "  const flat = flatten([[1], [2]])",
+        "  return flat",
+        "}",
+        "",
+      ].join("\n");
+      fs.writeFileSync(mainFile, source);
+
+      const doc = makeDoc(source, `file://${mainFile}`);
+      const symbolTable = SymbolTable.build(mainFile, {});
+      const { diagnostics } = runDiagnostics(doc, mainFile, {}, symbolTable);
```
egonSchiele: This asserts that no diagnostic matches `/is not defined/`. If the snippet ever stops parsing, `runDiagnostics` returns a parse error and no name-resolution diagnostics at all — zero matches — and the test goes green while checking nothing.

That is not hypothetical for this particular snippet: it uses `guard(...) { }` in value position, which is newer syntax than the rest of the file. It is exactly the kind of thing a grammar change could break.

I verified it is not vacuous today — I ran the source through `runDiagnostics` and `program` came back non-null, with one unrelated diagnostic (`Function 'guard' may throw interrupts [std::guard] but is not inside a handler.`). So the test is doing real work right now; the concern is only about it silently stopping.

The sibling test in this same file already guards this way (`expect(program).not.toBeNull()` in the `import test` block). Matching it is one line:

```ts
const { diagnostics, program } = runDiagnostics(doc, mainFile, {}, symbolTable);
expect(program).not.toBeNull();
```

packages/agency-lang/lib/lsp/diagnostics.ts
```diff
@@ -57,7 +49,7 @@ function ensureStdlibImport(
     importedNames: [
       {
         type: "namedImport",
-        importedNames: STDLIB_AUTO_IMPORTS,
+        importedNames: PRELUDE_NAMES,
```
egonSchiele: Not a defect in this diff — a question about whether the PR's claim holds all the way.

The stated goal is that the editor and the compiler cannot disagree about the prelude. Sharing the name list closes the drift in *what* is imported. But there is a third prelude-aware component, `prunePreludeShadows` (`lib/preprocessors/prunePreludeShadows.ts`), which decides what stays imported once a user shadows a name. It runs only from `typescriptPreprocessor.ts:339`, and the LSP pipeline in this file never calls it.

So the two paths still diverge when a user shadows a prelude name. I measured it — this file:

```
def map(xs: number[]): number[] {
  return xs
}
node main() {
  return map([1])
}
```

gets `'map' shadows an imported function.` from `runDiagnostics`. On the compile path `prunePreludeShadows` drops `map` from the injected import first, so by the time anything looks, there is no import left to shadow. (I read that from the code rather than running `typecheck` — the worktree is unbuilt, so I could not exercise the CLI.)

A warning is much milder than the `_guard` error this PR fixes, and shadowing a prelude name is rarer than using `guard`, so I would not hold the PR for it. But it is the same bug class, and someone reading `prelude.ts`'s docstring would reasonably conclude it was already handled. Two options:

- Call `prunePreludeShadows(program)` in `runDiagnostics` after parse, alongside `resolveReExports`. It is a pure AST mutation with no codegen dependency, so it looks safe in an analysis-only path — worth a quick check that it does not drop the name the editor needs for hover and go-to-definition.
- Or leave it and file a follow-up, and soften the `prelude.ts` docstring so it claims parity of the *name list* specifically rather than parity of what is in scope.

egonSchiele:
All three addressed in dc6c2163e. I verified each rather than just applying them.

**1. `readonly string[]` + spread at the call site.** Done. Your read is right that it is latent, not live: `prunePreludeShadows.ts:56` reassigns rather than splices, so nothing corrupts the array today. Worth closing anyway for the reason you gave.

**2. Vacuous-pass guard.** Done, matching the sibling test. Confirming your diagnosis independently: the snippet does produce the unrelated `Function 'guard' may throw interrupts [std::guard] but is not inside a handler.` warning, so the test is live today.

**3. `prunePreludeShadows`.** I took your first option and ran it in `runDiagnostics` rather than filing a follow-up. Reasoning: it is the same bug class this PR exists to fix, and leaving one instance behind while the docstring claims parity is the worse outcome.

I reproduced your finding first. Your exact `def map` snippet produced `'map' shadows an imported function.` from `runDiagnostics` as a failing test, then went green with the pass wired in.

On your caveat about hover and go-to-definition: all 22 LSP test files pass, 95 tests, including `hover`, `definition`, `completion`, and `references`. Pruning removes the *import* for a shadowed name, so resolution lands on the user's own `def`, which is what the editor should point at.

One thing worth flagging that I found while testing. `_guard` is exempt from pruning (`UNPRUNABLE_PRELUDE_NAMES`), and I wanted to be sure it did not go quiet along with the genuine shadow warnings. It keeps **both** diagnostics:

```
'_guard' shadows an imported function.
'_guard' is a reserved built-in; cannot be redefined.
```

That is the fail-closed behavior the pruning docstring describes, so I added a test pinning both messages. My first version of that test only asserted `diagnostics.length > 0`, which passed before the fix and therefore proved nothing; it now asserts the two messages specifically.

`tsc --noEmit` clean.

PR #599, opened 2026-07-19
Add list comprehensions: [expr for x in xs if cond] and fork [...]

egonSchiele:
Adds Python-shaped list comprehensions to Agency in two forms: `[expr for x in xs if cond]` (sequential) and `fork [expr for x in xs if cond]` (concurrent).

Plan: `docs/superpowers/plans/2026-07-19-list-comprehensions.md` (rev 4, three review rounds applied). Spec: `docs/superpowers/specs/2026-07-18-list-comprehensions-design.md`.

## How it works

The comprehension parses to its own AST node so `agency fmt` prints back exactly what the author wrote. A desugar pass inside `parseAgency` (before `lowerPatterns`) then rewrites it in place into ordinary `map` / `filter` / `fork` calls carrying block arguments - the exact shape those calls have when written by hand. After that point no later stage can tell the construct existed, so typing, codegen, interrupts, and fork branch semantics are inherited rather than reimplemented. Handler registration is untouched: interrupts raised inside a comprehension body flow through the same block machinery as hand-written fork blocks, and a test pins that the comprehension form behaves character-for-character like the hand-written baseline.

Binders mirror the `for` loop: a second binder is the index for arrays and the value for objects, and destructuring binders work. Two-binder forms pair items with indices BEFORE filtering, so indices are positions in the source (Python's enumerate-then-filter order). The pairing goes through a new `_pairsOf` prelude function whose notion of "what is iterable" is shared with `Runner.loop` via a new `classifyIterable` helper, so comprehensions and `for` loops can never drift apart.

## Also in this PR

- **Bug fix: `fork`/`race` in argument position crashed at runtime.** `total(fork(xs) as x { ... })` emitted `__call(fork, ...)` with `fork` an undefined identifier. The root cause was a hand-copied duplicate of the argument dispatch in `buildCallDescriptor`; the duplicate is collapsed onto `processResolvedArgs`, which now routes block-carrying `fork`/`race` arguments through `processForkCall`.
- **`mapWithIndex`**, which `docs/site/guide/blocks.md` documented but which did not exist. One declarative line over `_pairsOf`.
- **A pin for live `for`-loop iteration**: a body that appends to the array it iterates keeps going. Nothing tested this before, and the `Runner.loop` refactor here is exactly the kind of change that could silently break it.
- **Guide page** at `docs/site/guide/comprehensions.md`, plus cross-references from basic-syntax and partial-results.

## Deviations from the plan, all discovered by probing

- The parser needed `optionalSpacesOrNewline` (not required `spaces`) before the `for`/`in`/`if` keywords: exprParser eats trailing whitespace after a call but not after a bare name, so requiring whitespace broke every call-bodied comprehension.
- Multi-line comprehensions parse fine (the plan expected a single-line v1 limit). Pinned with tests at every keyword boundary and documented in the guide.
- Malformed comprehensions like `[x for x in]` do not error - the array parser has always accepted whitespace-separated items, so they parse as arrays of variable names. Pinned as-is; a real diagnostic is future work.
- `walkNodes` (std::agency's agent-facing AST walk) hand-enumerates its expression descent and needed its own `comprehension` case in addition to `expressionChildren` - without it, calls inside comprehension bodies were invisible to `getNodesOfType`.
- The "type error inside a comprehension body" location test uses a wrong-arity call: wrong-type calls cannot error because map's block parameter is `any`, exactly as in hand-written `map(xs) as x { ... }`.
- `state-isolation.md` was left unchanged: its three `push` sites are single statements demonstrating isolation, not loops a comprehension could replace.
- No docs build ran: VitePress is not installed anywhere in the workspace. All links in the new page were verified against the files on disk instead.

## Deferred

Nested cartesian loops (`for x in xs for y in ys`), record comprehensions, and a concurrency cap on `fork [...]` (`std::concurrency.concurrencyPool` covers the need now).

## Parse-time cost

Measured (review question): the desugar walk costs 2.15ms per pass over the full stdlib corpus (78 files, 545KB) against 763ms of total parse time with lowering on - 0.28%. Warm-make run-to-run variance on the same machine is measured in seconds, so the walk is noise at build level. The atom-chain insertion ahead of agencyArrayParser also adds no meaningful work: the array parser's retry after a failed comprehension attempt hits the memoized exprParser.

## Testing

- Full unit suite: 8735 passed. Structural linter clean. Diff audited against `docs/dev/anti-patterns.md`.
- New unit suites: parser (20 cases including ambiguity guards - every array literal in the language flows through the touched atom chain, and the generator fixtures show zero churn), desugar shape tests, loc stamping, formatter round-trip, `classifyIterable`, and a `std::agency` walk test.
- New execution tests (no LLM calls): every expression position the spec advertises, two-binder and destructuring forms with source-position indices, fork branch isolation (probed against the hand-written form before pinning: globals mutated in branches are discarded at the join), completion-order-inverted sleeps proving source-order results, and per-branch interrupt resumption matching the hand-written baseline exactly.

### DIFF
--- a/packages/agency-lang/docs/site/stdlib/index.md
+++ b/packages/agency-lang/docs/site/stdlib/index.md
@@ -24,7 +24,7 @@ How an existing file is handled on write:
 export type WriteMode = "overwrite" | "append" | "create-only"
 ```
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L50))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L51))
 
 ## Effects
 
@@ -37,7 +37,7 @@ effect std::read {
 }
 ```
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L34))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L35))
 
 ### std::write
 
@@ -48,7 +48,7 @@ effect std::write {
 }
 ```
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L38))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L39))
 
 ### std::readImage
 
@@ -59,7 +59,7 @@ effect std::readImage {
 }
 ```
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L42))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L43))
 
 ## Functions
 
@@ -79,7 +79,7 @@ Print a message to the console.
 |---|---|---|
 | messages | `any[]` |  |
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L52))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L53))
 
 ### setAgentCwd
 
@@ -104,7 +104,7 @@ Set the agent working directory. Path-taking tools (read, write, edit,
 |---|---|---|
 | dir | `string` |  |
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L81))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L82))
 
 ### getAgentCwd
 
@@ -116,7 +116,7 @@ Return the agent working directory, or an empty string if none is set.
 
 **Returns:** `string`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L90))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L91))
 
 ### applyAgentCwd
 
@@ -135,7 +135,7 @@ relative paths against the agent working directory.
 
 **Returns:** `string`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L100))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L101))
 
 ### printJSON
 
@@ -155,7 +155,7 @@ Print an object as formatted JSON to the console.
 | obj | `any` |  |
 | highlight | `boolean` | false |
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L108))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L109))
 
 ### input
 
@@ -178,7 +178,7 @@ prompt, which surfaces as an AgencyCancelledError.
 
 **Returns:** `string`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L125))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L126))
 
 ### sleep
 
@@ -199,7 +199,7 @@ A long sleep wakes immediately on Ctrl-C, race-loser, or time-guard abort.
 |---|---|---|
 | ms | `number` |  |
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L136))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L137))
 
 ### saveDraft
 
@@ -226,7 +226,7 @@ Record a best-so-far value for the current function or guarded block. If an
 |---|---|---|
 | value | `any` |  |
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L145))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L146))
 
 ### read
 
@@ -262,7 +262,7 @@ Read the contents of a file and return it as a string.
 
 **Throws:** `std::read`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L163))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L164))
 
 ### write
 
@@ -298,7 +298,7 @@ Write content to a file.
 
 **Throws:** `std::write`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L191))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L192))
 
 ### writeBinary
 
@@ -335,7 +335,7 @@ Write base64-encoded binary data to a file: images, audio, video, PDFs, or any
 
 **Throws:** `std::writeBinary`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L221))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L222))
 
 ### readBinary
 
@@ -366,7 +366,7 @@ Read a file and return its contents as a Base64-encoded string. Works for any
 
 **Throws:** `std::readBinary`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L251))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L252))
 
 ### range
 
@@ -389,7 +389,7 @@ Generate an array of numbers. With one argument, counts from 0 to start-1;
 
 **Returns:** `number[]`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L274))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L275))
 
 ### map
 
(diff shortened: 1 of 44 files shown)


packages/agency-lang/lib/parser.ts
```diff
@@ -273,6 +274,15 @@ export function parseAgency(
     const result = _parseAgency(input, config);
     if (result.success) {
       if (lower) {
+        // Comprehensions desugar FIRST, so a destructuring binder
+        // (`[f(name) for {name} in people]`) becomes an ordinary
+        // pattern assignment that lowerPatterns then handles with its
+        // existing machinery. Running after would leave a pattern no
+        // later stage knows how to read.
+        result.result.nodes = desugarComprehensionsInBody(
```
egonSchiele, line 282: **Question, not an objection: was parse-time cost measured?**

This adds a second full-tree walk next to `lowerPatterns`, and `mapChildren` does an `Object.keys` allocation per node. It runs on every parse with `lower: true`, including the overwhelming majority of files that contain no comprehension at all.

I talked myself out of the other half of this concern, so for the record: I expected `comprehensionParser` sitting ahead of `agencyArrayParser` to make every array literal in the language parse its first element twice. It does not, because `exprParser` is itself `memo`-wrapped (`parsers.ts:2837`), so the array parser's retry hits the memo cache. That one is fine.

The full-tree walk is the part that is genuinely new work. It is probably a small constant factor on a walk that already happens, but this repo has history here — #457 was CI compile time going 14 to 24 minutes — so it seems worth a warm/cold `make` number before merge rather than after. If it is noise, say so in the PR body and it never comes up again. If it is not, fusing the walk into `lowerPatterns` is an option, since they run back to back over the same tree.

packages/agency-lang/lib/lowering/comprehensionDesugar.ts
```diff
@@ -0,0 +1,251 @@
...
+function paramNameFor(node: Comprehension): string {
+  return typeof node.itemVar === "string" && !node.indexVar
+    ? node.itemVar
+    : PARAM;
+}
+
+/** `__comprehensionItem[n]` - recovers one binder from a pair.
+ *  NumberLiteral carries its value as a STRING (lib/types/literals.ts,
+ *  and parallelDesugar's numLit helper). */
+function pairIndex(position: number): Expression {
```
egonSchiele: `pairIndex` hardcodes `PARAM`, but the block parameter name is decided by `paramNameFor` above, which returns the user's own binder name for a single-name binder.

They cannot disagree today: `pairIndex` is only reached when `node.indexVar` is set, and that same condition forces `paramNameFor` to return `PARAM`. But that is a two-hop invariant with nothing enforcing it, in a file that is otherwise careful to pass state explicitly. If `paramNameFor` ever grows a case, `pairIndex` starts reading a variable that is not in scope, and the failure is a runtime `undefined` rather than a compile error.

Passing `paramName` in as an argument makes it structurally impossible, and matches how `comprehensionSource` and `unpackStatements` already take it.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2790,24 +2806,14 @@ export class TypeScriptBuilder {
...
-    if (node.block) {
-      argNodes.push(this.processBlockArgument(node));
-    }
+    // No named args and no block on this call (the branch above handled
+    // those), so every argument is positional. processResolvedArgs is the
+    // one place that decides reference-vs-call for functionCall arguments;
+    // this loop used to be a hand-copy of it and drifted (the fork/race
+    // guard landed in one and not the other).
+    const argNodes = this.processResolvedArgs(
+      args as (Expression | SplatExpression)[],
```
egonSchiele: This cast is sound — the `hasNamedArgs || node.block` early return above guarantees no `namedArgument` survives to here — but it is an unchecked assertion carrying that invariant, one function away from where it is established.

Since the whole point of this hunk is that the two copies drifted, it would be nice if the type system held the line rather than a comment. A `.filter()` with a type predicate, or narrowing the local, keeps it checked at no real cost.

packages/agency-lang/lib/parsers/comprehension.test.ts
```diff
@@ -0,0 +1,155 @@
...
+  });
+
+  // Half-typed comprehensions are now a thing users and agents will
+  // produce constantly. They do NOT error: the array parser accepts
+  // whitespace-separated items and `in` is a binary operator, so these
+  // have always parsed as arrays of variable names (probed - this is
+  // pre-existing behavior, not something this feature introduced). Pin
+  // that they at least never become half-baked comprehension nodes, so
+  // a later improvement (a real diagnostic) shows up as a diff.
+  it.each([["[x for x in]"], ["[for x in xs]"], ["[x for in xs]"]])(
```
egonSchiele, line 130: Pinning the current behavior here is the right call, and I would not hold up the PR for it. But it is worth an issue rather than only a test comment, because of who writes this mistake.

`[x for x in]` parsing as an array of variable names means a half-typed comprehension produces no diagnostic at the comprehension, and instead surfaces later as `x is not defined` (or worse, silently type-checks if `x` and `for` happen to exist). Agents will produce half-typed comprehensions constantly — that is the population this feature is aimed at — and the failure mode points them at the wrong line.

An issue linked from this test would keep it from being rediscovered in six months as a mystery.

packages/agency-lang/lib/lowering/comprehensionDesugar.ts
```diff
@@ -0,0 +1,251 @@
...
+/** Copy the comprehension's source location onto a synthesized node and
+ *  everything under it. Without this, a type error inside a comprehension
+ *  body points at generated `map(...)` internals instead of the line the
+ *  user wrote. See docs/dev/locations.md.
+ *
+ *  Only fills a MISSING loc, so nodes carried over from the user's source
+ *  (the body expression, the iterable, the condition) keep their own real
+ *  positions. Uses the same `mapChildren` traversal as `desugarNode` -
+ *  one walking strategy per file. */
+function stampLoc<T>(target: T, loc: Comprehension["loc"]): T {
```
egonSchiele, line 229: Minor, and possibly already the intent: `stampLoc` only fills a *missing* `loc`, which correctly preserves real positions on the user's carried-over subtrees — but it also descends into those subtrees, so any node the parser did not stamp picks up the comprehension's location instead of staying blank.

That is probably what you want (a location on the right line beats no location). Worth one line in the docstring saying so deliberately, since the current wording — "nodes carried over from the user's source keep their own real positions" — reads as though user nodes are not touched at all, and they are.

egonSchiele (commented):
Reviewed the whole diff and traced the load-bearing claims back to the code. This is in good shape. Notes below are one question and four small things; none of them block.

**Things I checked rather than took on faith:**

- **The `buildCallDescriptor` collapse is safe.** The positional branch used to end with `if (node.block) argNodes.push(this.processBlockArgument(node))`, and dropping it looks alarming in the diff. It was dead code: `if (hasNamedArgs || node.block)` returns above it, so the positional branch can never see a block. Good find, and collapsing the hand-copy is the right fix rather than adding the fork guard in two places.
- **`classifyIterable` returning a classification instead of a materialized list.** This is the version that does not change `for` loop behavior. `Runner.loop` still holds the caller's array by reference and re-reads `.length`, and `for-loop-live-iteration.agency` pins it. That test is new coverage for behavior that was previously untested, which is worth more than the refactor that prompted it.
- **The interrupt tests derive their expected value instead of asserting one.** `"3|2|fail|6"` encodes a real discovery — a rejected branch fails alone rather than aborting the whole fork — and the comprehension test is pinned against the hand-written baseline rather than against a guess. This is the test that backs the architectural claim, and it is built correctly.
- **`expressionChildren` and `walkNodes` both got a `comprehension` case.** `expressionChildren` ends in `default: return []`, so adding a node type to the union produces no compile error and a walker silently treats it as a leaf. On the `lower: false` paths that keep the raw construct — `std::agency`'s `_getNodesOfType` / `_walkAST` — that would have made calls inside comprehension bodies invisible to agents. Catching that the hand-enumerated `walkNodes` descent needed its own case in addition is the part that is easy to miss.
- **Both `blocks.md` call shapes are tested**, and `mapWithIndex` is one declarative line over `_pairsOf` rather than an accumulator loop.

One question and four small comments inline.

packages/agency-lang/lib/stdlib/builtins.ts
```diff
@@ -326,6 +327,25 @@ export function _entries(obj: any): { key: string; value: any }[] {
...
+ *  `[f(x,i) for x, i in src]` and `for (x, i in src)` cannot disagree.
+ *
+ *  Unlike the loop, this genuinely does build a list: the comprehension
+ *  desugars to `map(_pairsOf(src))`, and `map` needs a real array. Only
+ *  the classification is shared, which is the part that would drift. */
+export function _pairsOf(src: unknown): unknown[][] {
+  const shape = classifyIterable(src);
+  if (shape.kind === "array") {
+    return (src as unknown[]).map((item, index) => [item, index]);
+  }
```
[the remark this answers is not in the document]
egonSchiele: Confirmed and fixed in 3c5793144: probed that map leaves holes in its result, so the pairs array itself would have been sparse. The array case now builds pairs by index (Array.from over length), matching Runner.loop on sparse arrays.

packages/agency-lang/lib/utils/iteration.ts
```diff
@@ -0,0 +1,41 @@
...
+ *   - `_pairsOf` (stdlib/builtins.ts) — two-binder list comprehensions
+ *
+ * Both must agree, or `[f(x) for x in src]` behaves differently from
+ * `for (x in src)`. Keeping one implementation is why this file exists;
+ * PR #595 collapsed the equivalent hand-copied drift in the prelude list.
+ *
+ * Arrays iterate by element. Non-null objects iterate by key. Everything
+ * else — null, undefined, numbers, STRINGS — iterates nothing, matching
+ * how a JS `for...of` over a non-iterable simply does nothing rather than
+ * crashing mid-flow.
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 3c5793144 - the comment now states this as an Agency semantic choice and notes that JS for...of would throw. The wording was inherited from the original Runner.loop comment, which had the same false claim.

packages/agency-lang/docs/site/stdlib/index.md
```diff
@@ -411,7 +411,30 @@ Map a function over an array, returning a new array of results.
...
 **Returns:** `any[]`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L295))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/index.agency#L296))
+
+### mapWithIndex
+
+```ts
+mapWithIndex(arr: any[], func: (any, any) -> any): any[]
+```
```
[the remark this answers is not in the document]
egonSchiele: This diff was produced by running make (agency doc), not by hand-editing - the line-link shifts come from the new _pairsOf and mapWithIndex definitions moving everything below them. No change needed.

packages/agency-lang/lib/parser.ts
egonSchiele: Measured. Over the full stdlib corpus (78 files, 545KB), the desugar walk costs 2.15ms per corpus pass against 763ms of total parse time with lower:true - 0.28%, and three orders of magnitude below warm-make run-to-run variance (3.3s vs 12.9s on consecutive runs on my machine). Added the number to the PR body. Your memo observation matches what I found during implementation: the array parser's retry after a failed comprehension attempt hits the memoized exprParser, so the atom-chain insertion adds no meaningful parse work either. Given 0.28%, I left the walk separate from lowerPatterns - fusing them saves nothing measurable and couples two passes that are simpler apart.

packages/agency-lang/lib/lowering/comprehensionDesugar.ts
egonSchiele: Fixed in 3c5793144: pairIndex now takes paramName as its first argument and unpackStatements threads it through, matching comprehensionSource. The two-hop invariant is gone.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Fixed in 3c5793144: replaced the cast with a filter carrying an 'arg is Expression | SplatExpression' predicate. A namedArgument that somehow reached the positional branch is now dropped by a checked narrowing instead of being miscompiled through the cast.

packages/agency-lang/lib/parsers/comprehension.test.ts
egonSchiele: Filed as #602, and the test comment now points at it. The issue records the failure mode (half-typed comprehension surfaces as 'x is not defined' on the wrong construct) and sketches the fix shape: a targeted parse diagnostic when the comprehension parser matches '[ expr for' and then fails, instead of silently backtracking into the array parser.

packages/agency-lang/lib/lowering/comprehensionDesugar.ts
egonSchiele: It was the intent, and you are right that the docstring contradicted it. Fixed in 3c5793144: the docstring now says explicitly that stampLoc descends into carried subtrees and that unstamped nodes deliberately pick up the comprehension's location, because a location on the right line beats no location.

Issue #606, opened 2026-07-19
Should race wait for the first success instead of the first result?


egonSchiele:
`race` runs several branches at once and returns the one that finishes first. A branch counts as finished even when it fails. So a branch that fails immediately wins the race, and the branches that were about to succeed get cancelled.

PR #604 found this behavior, pinned it in `tests/agency/comprehensions/race-interrupt.agency`, and documented it in the comprehensions guide. The behavior is older than that PR. `runBatch.ts` already called it "the race's first-rejection seal". Nothing here is a regression. This issue is about whether the semantics are the ones we want.

## Why it matters

`race` is the natural way to ask several models the same question and take the fastest answer:

```ts
const answer = race [ask(model) for model in models]
```

If one model hits a rate limit and errors right away, that error wins. The other two models were close to a good answer, and Agency throws their work away.

Adding models makes this worse. Four models mean four chances that one of them fails fast and ruins the call. That is backwards. People add branches to make a call more reliable, not less.

## Two shapes exist

JavaScript ships both. `Promise.race` returns the first branch to settle, even if it rejected. `Promise.any` returns the first branch to succeed, and only fails when every branch fails.

Agency has the first shape and calls it `race`, which is an honest name. It does not have the second shape.

## Question 1: should `race` wait for the first success?

Options as I see them:

1. Leave `race` alone. Users check the result with `isFailure` and retry themselves. This is what the guide tells them to do today.
2. Change `race` to return the first success. It would fail only when every branch fails. This is a breaking change to a shipped primitive.
3. Keep `race` as it is and add a second primitive with first-success semantics.

Option 3 costs the most and breaks the least. Option 2 probably matches what people expect from the word, judging by how they reach for `race`.

## Question 2: what should an empty race return?

A race over zero branches returns `null` today. PR #604 added that, because before it the program hung forever with no error.

`null` matches how Agency normalizes other absent values, like `obj[missingKey]` and an unmatched `match`. That consistency is worth something.

But it means two different failure modes need two different checks. An empty race returns `null`. A race whose fastest branch failed returns a Failure. Both mean "nothing useful came back". A Failure for the empty case would collapse them into one check and would say what went wrong at the point of use.

## Why these two questions belong together

Both are about what `race` hands back when things go wrong. Answering them separately risks a result type where the caller checks for `null`, checks `isFailure`, and still is not sure whether any branch ran.

## Scope

[...]

PR #607, opened 2026-07-19
Type fork block calls as T[] and race as T | null from the block returns

### LINKED ISSUE #603: Checker types race results as any, so a wrong T[] annotation on a race passes
The checker types the result of a `race(...)` block call - and therefore of a `race [...]` comprehension - as `any`. Both of these typecheck clean:

```ts
const w: number = race([1, 2, 3]) as x { return x * 2 }     // correct
const w: number[] = race([1, 2, 3]) as x { return x * 2 }   // wrong, but accepted
```

A race returns ONE value (the first branch to settle), not a list, so the second annotation is a real user mistake the checker cannot catch today. The comprehension form (`const summaries: string[] = race [summarize(d) for d in docs]`) makes it easier to write, because the brackets visually suggest a list.

The guide states the scalar result in prose ("one value, **not a list**"), but a synthesizer case for `race`/`fork` block calls would catch it structurally: `fork` returns `T[]` and `race` returns `T | null` (null since the empty-race fix) where `T` is the block's return type.

Block-form scope - the comprehension desugars to the same call and inherits whatever the checker learns.

egonSchiele:
The checker now infers `fork(xs) as x { ... }` as `T[]` and `race(xs) as x { ... }` as `T | null` from the block's returns, so a wrong list/scalar annotation on a race result - `const summaries: string[] = race [...]` - is a type error instead of silently passing as `any`. The comprehension forms (`fork [...]`, `forkShared [...]`, `race [...]`, `raceShared [...]`) inherit the types because they desugar to these calls before the checker runs; they have their own tests anyway.

Closes #603.

Spec: `docs/superpowers/specs/2026-07-19-fork-race-return-types-design.md`. Plan: `docs/superpowers/plans/2026-07-19-fork-race-return-types.md` (rev 2, one review round applied).

## Mechanism

The `_guard` precedent, generalized into data: `synthFunctionCall` now dispatches reserved block-carrying calls through one `BLOCK_CALL_RESULT` table - a shared `blockReturnType` walk (extracted from `synthGuardCall`, which is deleted; its wrapper became the `_guard` row) plus a per-call wrapper lambda. Same table shape as `CALL_NAME` (#599) and `COMPREHENSION_PREFIXES` (#604). Adding another block-typed call is a row. `blockOf` is the single home for the block cast. No parser, desugar, builder, or runtime changes.

## Recorded decisions and limitations (from the spec)

- **Failures are invisible.** A rejected branch really leaves a Failure in a fork slot, and a race can return one (first-settled semantics), but the types are the optimistic `T[]` / `T | null`, not `(T | Failure)[]`. Honest typing would force a guard on every element and needs union narrowing the checker does not have (probed: `Result<T> | null` narrows `.value` through `isSuccess` but not `.error`). A test pins the dependency this rests on: `isFailure` accepts `ANY_T`, so failure checks on typed elements keep compiling.
- **The `any`-element limitation.** When a block's return type is `any` (e.g. it returns a bare block parameter), race collapses to `any` and the #603 headline annotation still passes. The fix bites when the body produces a concrete type - the common case. Fork is different: the builtin fallback already typed blockless-and-block forks as `any[]`, and the new case keeps that list shape, so scalar annotations on any-element forks error (already did, via the builtin).
- **Race is stricter than `arr[i]`.** Index access keeps optimistic plain-`T` typing; the race `| null` reflects the deliberately-introduced empty-race null (#604) that the guide tells callers to handle.
- **Coupled to #606.** The race row encodes empty-race-returns-null. If #606 changes that (Failure instead of null, or first-success semantics), the row changes with it - the coupling is written at the row itself.

## The sweep caught four real pre-existing gaps

`tc` over stdlib and both test corpora, baselined against the behavior-identical refactor commit. stdlib and agency-js: zero new diagnostics (the only two stdlib call sites are any-element). Four catches in `tests/agency`, all fixed:

- `fork/fork-stress.agency`: a race result returned as bare `string`; now `?? ""` with a comment that the statically non-empty source makes the null arm unreachable.
- `fork/race/race-reject-winner.agency`: reads `.value`/`.error` off a race winner that really is a Failure (the test pins first-settled rejection). That access was only ever legal via `any`; the binding is now explicitly `: any` with a comment naming this as the recorded cost of the failures-invisible decision.
- `fork/race-basic.agency`: nullable fed to `.includes(number)`; now behind a `!= null` narrow.
- `fork-arg-position.agency`: `[race(...) as x { ... }]` built a `(number | null)[]` for a `number[]` parameter; fixed with a parenthesized `?? 0` that keeps the race in argument position, which is the coverage that line exists for.

The remaining corpus directories (`lib/agents`, `tests/typescriptGenerator`, `agents`, `examples`) contain no fork/race calls - grep re-verified, including against the #600 files that landed on main after this branch was cut.

## Testing

19 new checker tests. The positive inference cases CONSUME the result (`xs[0]` bound to the wrong type, `?? 0` bound to the wrong type) so they fail when the inference is absent - under the old `any` typing, annotation-accept tests pass identically with or without the feature. Accept-cases assert the unfiltered diagnostic list is empty; reject-cases pin message content. Coverage includes fork-in-fork nesting (`T[][]`), fork-in-guard composition, blockless-call fallback, return-position checking, void blocks, `raceShared`, and the fail-open shapes. Full checker suite: 1628 passing. Structural linter clean. The four touched execution tests re-run green.

### DIFF
--- a/packages/agency-lang/lib/typeChecker/synthesizer.ts
+++ b/packages/agency-lang/lib/typeChecker/synthesizer.ts
@@ -578,49 +578,82 @@ function synthPipeRhs(
  *  failure side stays `any`, the `try`-expression precedent — the
  *  err-side narrowing users rely on (`isFailure`, `error.type`) is
  *  field-based, not type-based. */
-function synthGuardCall(
+/** The block argument of a call, if it carries one. The single home for
+ *  the block cast - every block-typing consumer reads through this. */
+function blockOf(
   expr: AgencyNode & { type: "functionCall" },
+): { body: AgencyNode[] } | undefined {
+  return (expr as unknown as { block?: { body: AgencyNode[] } }).block;
+}
+
+/** The type a block argument's OWN returns produce. The walk skips
+ *  returns belonging to nested function/node/block definitions, then
+ *  folds: no returns -> void; any return typed `any` -> `any`
+ *  (fail-open); otherwise the union of the distinct return types.
+ *  Every BLOCK_CALL_RESULT row reads through this one walk, so the
+ *  block constructs cannot drift on what counts as a block-level
+ *  return. */
+function blockReturnType(
+  block: { body: AgencyNode[] },
   scope: Scope,
   ctx: TypeCheckerContext,
 ): VariableType {
-  const block = (expr as unknown as { block: { body: AgencyNode[] } }).block;
   const returnTypes: VariableType[] = [];
   for (const { node, ancestors } of walkNodes(block.body)) {
-    if (node.type !== "returnStatement" || !node.value) continue;
+    if (node.type !== "returnStatement" || !node.value) {
+      continue;
+    }
     const nested = ancestors.some(
       (a) =>
         a.type === "function" ||
         a.type === "graphNode" ||
         a.type === "blockArgument",
     );
-    if (!nested) returnTypes.push(synthType(node.value, scope, ctx));
+    if (!nested) {
+      returnTypes.push(synthType(node.value, scope, ctx));
+    }
   }
-  let successType: VariableType;
   if (returnTypes.length === 0) {
-    successType = { type: "primitiveType", value: "void" };
-  } else if (returnTypes.some((t) => isAnyType(t))) {
-    successType = ANY_T;
-  } else {
-    successType = unionTypes(returnTypes);
+    return VOID_T;
   }
-  return { type: "resultType", successType, failureType: ANY_T };
+  if (returnTypes.some((t) => isAnyType(t))) {
+    return ANY_T;
+  }
+  return unionTypes(returnTypes);
 }
+
+/** Reserved calls whose type comes from their block's returns. The walk
+ *  is shared (blockReturnType); only the wrapper differs. Adding
+ *  another block-carrying reserved call is a row. Keying on the call
+ *  name is shadow-proof: every key is in RESERVED_FUNCTION_NAMES. */
+const BLOCK_CALL_RESULT: Record<
+  string,
+  (element: VariableType) => VariableType
+> = {
+  // The guard construct's Result<T> precision. By synth time the
+  // construct has desugared to `_guard(...)` carrying its block
+  // (guardDesugar.ts + the TypeChecker constructor); _guard's own
+  // annotation is the generic Result, so parameterize it from the
+  // block. A direct `_guard` call with a block gets the same, correct,
+  // typing.
+  _guard: (element) => ({
+    type: "resultType",
+    successType: element,
+    failureType: ANY_T,
+  }),
+};
 
 function synthFunctionCall(
   expr: AgencyNode & { type: "functionCall" },
   scope: Scope,
   ctx: TypeCheckerContext,
 ): VariableType {
-  // The guard construct's Result<T> precision. By synth time the
-  // construct has desugared to a `_guard(...)` call carrying its block
-  // (see guardDesugar.ts + the TypeChecker constructor), and _guard's
-  // own annotation is the generic `Result` — so parameterize it here
-  // from the BLOCK's returns, the way an unannotated def body infers.
-  // Keyed on the desugared shape (name + block), which user code only
-  // produces via the construct; a direct `_guard` call with a block
-  // gets the same, correct, typing.
-  if (expr.functionName === "_guard" && (expr as { block?: unknown }).block) {
-    return synthGuardCall(expr, scope, ctx);
+  // Reserved block-carrying calls (guard construct, fork, race) take
+  // their type from the block's returns - see BLOCK_CALL_RESULT.
+  const wrapBlockResult = BLOCK_CALL_RESULT[expr.functionName];
+  const calledBlock = blockOf(expr);
+  if (wrapBlockResult !== undefined && calledBlock !== undefined) {
+    return wrapBlockResult(blockReturnType(calledBlock, scope, ctx));
   }
   // Result constructors: parameterize ResultType from the argument so callers
   // get `Result<T, any>` (success) or `Result<any, T>` (failure). The names
--- a/packages/agency-lang/lib/typeChecker/synthesizer.ts
+++ b/packages/agency-lang/lib/typeChecker/synthesizer.ts
@@ -641,6 +641,19 @@ const BLOCK_CALL_RESULT: Record<
     successType: element,
     failureType: ANY_T,
   }),
+  // fork joins every branch into a list. The list shape is kept even
+  // for an `any` element - any[] still catches scalar annotations.
+  // Failure slots are deliberately invisible (spec: "optimistic about
+  // failures").
+  fork: (element) => ({ type: "arrayType", elementType: element }),
+  // race yields the first-SETTLED branch, or null when zero branches
+  // ran (empty source, or a comprehension filter that matched nothing
+  // - #604). An `any` element collapses to any, NOT any | null:
+  // unionTypes does not absorb any, and fail-open must stay fail-open.
+  // COUPLED to #606: if the empty-race return value or first-success
+  // semantics change there, this row changes with them.
+  race: (element) =>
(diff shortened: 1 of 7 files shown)


packages/agency-lang/lib/typeChecker/synthesizer.ts
```diff
@@ -578,49 +578,95 @@ function synthPipeRhs(
  *  failure side stays `any`, the `try`-expression precedent — the
  *  err-side narrowing users rely on (`isFailure`, `error.type`) is
  *  field-based, not type-based. */
-function synthGuardCall(
+/** The block argument of a call, if it carries one. The single home for
```
egonSchiele, line 575: This docstring outlived the function it describes. It documented `synthGuardCall`, which this PR deletes, and it now sits directly above `blockOf`'s own docstring, so the file has two consecutive `/** */` blocks where the first describes nothing.

Worth moving rather than deleting. The content is real knowledge that the new `_guard` row does not carry: why the failure side stays `any`, and that the err-side narrowing users rely on (`isFailure`, `error.type`) is field-based rather than type-based. The row's comment currently explains the desugaring but not that reasoning, so folding these three lines into it keeps the argument next to the code it justifies.

packages/agency-lang/lib/typeChecker/synthesizer.ts
```diff
@@ -578,49 +578,95 @@ function synthPipeRhs(
...
-  // own annotation is the generic `Result` — so parameterize it here
-  // from the BLOCK's returns, the way an unannotated def body infers.
-  // Keyed on the desugared shape (name + block), which user code only
-  // produces via the construct; a direct `_guard` call with a block
-  // gets the same, correct, typing.
-  if (expr.functionName === "_guard" && (expr as { block?: unknown }).block) {
-    return synthGuardCall(expr, scope, ctx);
+  // Reserved block-carrying calls (guard construct, fork, race) take
+  // their type from the block's returns - see BLOCK_CALL_RESULT.
+  const wrapBlockResult = BLOCK_CALL_RESULT[expr.functionName];
```
egonSchiele: `expr.functionName` is an arbitrary user identifier, and `BLOCK_CALL_RESULT` is a plain object literal, so this lookup can walk the prototype chain:

```js
const t = { fork: 1, race: 2 };
typeof t["toString"]      // "function"
typeof t["constructor"]   // "function"
typeof t["valueOf"]       // "function"
```

So a call named `toString` carrying a block reaches `wrapBlockResult !== undefined`, and the next line calls `Object.prototype.toString(element)`, which synthesizes the string `"[object Object]"` as a type. `constructor` is worse in a quieter way: `Object(element)` returns the element unchanged, so the call types as its own element type with no wrapper and no error.

Requires a user function named `toString`, `valueOf`, `constructor`, or `hasOwnProperty` that takes a block, so it is rare. But the failure is a silently wrong type rather than a crash, and the old `expr.functionName === "_guard"` comparison was immune.

`Object.create(null)` for the table, or an `Object.hasOwn(BLOCK_CALL_RESULT, expr.functionName)` guard here, closes it.

Worth noting the distinction from the sibling tables this one is modeled on: `CALL_NAME` is keyed by `Comprehension["mode"]` and `COMPREHENSION_PREFIXES` by a keyword from a fixed `or(...)` list, so both have closed keysets and neither has this exposure. Keying on a user-supplied name is what makes it reachable here. (`ctx.functionDefs[expr.functionName]` a few lines below has the same shape, but it only reads `.returnType` off the result and falls through, so nothing calls it.)

egonSchiele (commented):
Read the whole diff and chased the claims that carry weight. This is good work. Three inline notes, one of which I would fix before merge.

**Things I checked rather than assumed:**

- **The `_guard` row is behavior-identical to the deleted `synthGuardCall`.** Same `blockReturnType` result as `successType`, same `ANY_T` failure side. Folding the third construct into the table rather than leaving it as a branch was the right call, and it makes the per-call difference readable as three lines side by side.
- **The `isAnyType` guard in the race row is load-bearing.** I went looking for it as a useless special case, since absorbing `any` is usually the combinator's job. `unionTypes` (`inference.ts:155`) only dedupes by `JSON.stringify` and builds a union node, so without that branch a fail-open race really would synthesize `any | null`, which is stricter than `any`. The comment defending it is accurate.
- **The `isFailure` test defends a real dependency.** `isFailure` takes `ANY_T` (`builtins.ts:245`), which is the only reason failure checks still compile against a typed element. That is exactly the kind of cross-file assumption that breaks silently later, and pinning it here is right.
- **The corpus sweep found genuine pre-existing bugs, not churn.** Four wrong annotations that `any` had been hiding. That is the feature working on its first day.
- **The plan's open grammar question is answered.** `(race(...) as x { ... }) ?? 0` parses, so the parenthesized block-call fallback was not needed.

The positive tests now consume their results (`xs[0]` bound to the wrong type) instead of restating the annotation, so they can actually fail when the inference is absent. That was the main gap in the plan and it is closed.

packages/agency-lang/tests/agency/fork/race/race-reject-winner.agency
```diff
@@ -4,7 +4,12 @@ def confirm(item: number): number {
 }
 
 node main() {
-  let result = race([10, 20, 30]) as item {
+  // Explicitly any: the checker types a race as number | null, but
+  // first-settled semantics can deliver a Failure (pinning that is
+  // this test's whole point), and reading .value/.error off it needs
+  // the checker out of the way. This is the recorded cost of the
+  // failures-invisible typing decision (fork/race spec, #603).
+  let result: any = race([10, 20, 30]) as item {
```
egonSchiele, line 12: This is the right fix and the comment is honest about what it is. No change requested.

Worth naming what it represents, though: this is the invoice for the failures-invisible decision arriving on day one. The checker says the value is `number | null`. First-settled semantics can hand back a Failure. So the one file in the repo that handles a failed race correctly is also the one file that has to switch type checking off to do it.

That is a defensible trade, and the spec recorded it deliberately. But it is worth a line on #603 or #606, because a future reader deciding whether to make fork and race failure-aware will want to know that the cost is already visible and already has a name. Right now the evidence lives only in this comment.

packages/agency-lang/lib/typeChecker/synthesizer.ts
egonSchiele: Fixed in 85f94603d. The orphan is deleted and its load-bearing half - failure side stays any per the try-expression precedent, err-side narrowing is field-based not type-based - is folded into the _guard row's comment, next to the code it justifies.

packages/agency-lang/lib/typeChecker/synthesizer.ts
egonSchiele: Confirmed and fixed in 85f94603d (test corrected in de43d3e7e). Probing the unguarded lookup showed it is worse than a silently wrong type in one path: the Object.prototype.toString wrapping produces a value the diagnostic formatter cannot render, so tc CRASHES with an Unknown-variable-type throw. The dispatch now guards with Object.hasOwn, the comment records why the sibling tables (closed keysets) do not need it, and a regression test defines a block-accepting user def named toString and asserts it types through the ordinary def path - verified failing against the unguarded lookup. One honest note: my first push (85f94603d) went out while that regression test was still red from two test-source bugs of my own (a def-shadow warning and a missing block-typed parameter, both legitimate checker behavior); de43d3e7e fixes the test source, and the head is green at 1629.

packages/agency-lang/tests/agency/fork/race/race-reject-winner.agency
egonSchiele: Recorded on #606 as requested: https://github.com/egonSchiele/agency-lang/issues/606#issuecomment-5016938648. The comment names the invoice (the one correct failure-handling file is the one that must switch checking off), notes that first-success semantics would mostly dissolve the tension while first-settled keeps it blocked on union narrowing (probed: .value narrows through isSuccess, the .error else-branch does not), and points at the coupled race row.

PR #608, opened 2026-07-19
Require commas between literal items; targeted parse errors for half-typed comprehensions

### LINKED ISSUE #602: Half-typed comprehensions parse as arrays of variable names with no diagnostic
A half-typed comprehension such as `[x for x in]`, `[for x in xs]`, or `[x for in xs]` produces no diagnostic at the comprehension. The array parser accepts whitespace-separated items and `in` is a binary operator, so these parse as arrays of variable names (for example `[x, for, x, in]`). The failure then surfaces later as `x is not defined` on the wrong construct - or silently type-checks if the names happen to exist.

Agents will produce half-typed comprehensions constantly, and this failure mode points them at the wrong line.

Current behavior is pinned by the "parses the malformed comprehension ... as an array, not a comprehension" cases in `lib/parsers/comprehension.test.ts`, so a fix shows up as a diff there. A plausible shape: when the comprehension parser matches `[` expr `for` and then fails, emit a targeted parse diagnostic instead of backtracking silently into the array parser.

From PR #599 review.

egonSchiele:
Half-typed comprehensions like `[x for x in]` now fail with a targeted parse error at the comprehension, instead of silently parsing as a four-item array of variable names and surfacing later as `x is not defined` on the wrong line.

Closes #602.

Two mechanisms, because the investigation found the array fallback was itself a bug:

## 1. Literal items require commas between them again

The 2026-07-04 comments-between-entries rewrite (0b3233d67) replaced `sepBy(commaWithNewline, ...)` with a delimiter whose comma was optional EVERYWHERE - justified in its comment only by trailing commas - so any whitespace separated array/object items. `[1 2 3]` was a three-item array; `[x for x in]` was `[x, for, x, in]`. Two weeks old, accidental, and the reason half-typed comprehensions had somewhere silent to fall.

The delimiter now requires the comma between items, with one alternative branch: a LOOKAHEAD accepting only trailing position (optional comments/blank lines, then the closing bracket). That single formulation keeps everything the rewrite legitimately added - trailing commas, comments after a comma, trailing comments without a trailing comma (all pinned in dataStructures.test.ts and re-asserted beside the new rejections in literalDelimiter.test.ts) - and closes even the comment-interposed leak: `[1 // note\n 2]` fails, because the lookahead finds an item where the closer should be. The new grammar is strictly narrower, so no accepted parse changes shape; things either parse exactly as before or now error.

## 2. The comprehension parser commits after `[ expr for `

Past that prefix nothing but a comprehension attempt can be on the wire (#599 pinned the divergences: arrays fail at `str("for")` on a comma, `[format, other]` never sees a bare `for`, `forx` fails the required whitespace). Failures after the commit are now hard `parseError` throws with targeted messages - the same idiom `forLoopParser` uses:

- `[x for x in]` -> expected an iterable expression after `in`
- `[x for in xs]` -> expected `in` after the list comprehension binder (the binder greedily reads the word `in` as its name; the error lands on `xs`)
- `[x for x in xs if]` -> expected a filter condition after `if`
- `[x for x in xs` -> expected `]` to close this list comprehension

One placement subtlety: the required whitespace after `in` lives inside the ITERABLE commit, not the `in` commit, so `[x for x in]` blames the missing iterable rather than the `in` it already has. `[for x in xs]` never reaches the commit (the leading `for` parses as the body expression), so it fails through the comma rule instead - a generic message, but at the right position.

The three merged pins that asserted the malformed inputs parse as arrays now assert the specific messages - the diff those pins were built to make visible.

## Verification

- Full parser suite (1641), backends + lowering + formatter + checker suites (2212), zero fixture churn.
- `tc` sweep across `stdlib/`, `tests/agency`, `tests/agency-js`, and `lib/agents`: zero parse failures - nothing in the corpus relied on comma-less literals in the two weeks the regression existed. (A narrower grammar cannot reshape accepted parses, so type-level output is untouched by construction.)
- Full `make` recompiles the entire stdlib under the new grammar - the strongest parse gate available - and all ten comprehension execution tests plus the literal-heavy execution tests pass.

### DIFF
--- a/packages/agency-lang/lib/parsers/literalDelimiter.test.ts
+++ b/packages/agency-lang/lib/parsers/literalDelimiter.test.ts
@@ -1,0 +1,70 @@
+import { describe, it, expect } from "vitest";
+import { agencyArrayParser, agencyObjectParser } from "./parsers.js";
+
+/** Whole-input parse helpers: success means the parser consumed the
+ *  entire literal, matching how these parsers are driven in tests
+ *  elsewhere in this directory. */
+const arrayOk = (src: string) => {
+  const r = agencyArrayParser(src);
+  return r.success && r.rest === "";
+};
+const objectOk = (src: string) => {
+  const r = agencyObjectParser(src);
+  return r.success && r.rest === "";
+};
+
+describe("literal items require commas between them", () => {
+  it("parses ordinary comma-separated arrays and objects", () => {
+    expect(arrayOk("[1, 2, 3]")).toBe(true);
+    expect(arrayOk("[]")).toBe(true);
+    expect(objectOk(`{ "a": 1, "b": 2 }`)).toBe(true);
+    expect(objectOk("{}")).toBe(true);
+  });
+
+  it("keeps trailing commas", () => {
+    expect(arrayOk("[1, 2, 3,]")).toBe(true);
+    expect(objectOk(`{ "a": 1, "b": 2, }`)).toBe(true);
+  });
+
+  it("keeps multi-line literals with commas", () => {
+    expect(arrayOk("[\n  1,\n  2,\n  3\n]")).toBe(true);
+    expect(objectOk(`{\n  "a": 1,\n  "b": 2\n}`)).toBe(true);
+  });
+
+  it("keeps splats", () => {
+    expect(arrayOk("[...xs, 1]")).toBe(true);
+  });
+
+  // The 2026-07-04 comments-between-entries rewrite made the comma
+  // optional EVERYWHERE, so any whitespace separated items and
+  // `[x for x in]` parsed as a four-item array of variable names
+  // (#602). These pin the restoration of the required comma.
+  it("rejects whitespace-separated array items", () => {
+    expect(arrayOk("[1 2 3]")).toBe(false);
+    expect(arrayOk("[x for x in]")).toBe(false);
+    expect(arrayOk("[\n  1\n  2\n]")).toBe(false);
+  });
+
+  it("rejects whitespace-separated object entries", () => {
+    expect(objectOk(`{ "a": 1 "b": 2 }`)).toBe(false);
+  });
+
+  it("rejects a comment interposed between comma-less items", () => {
+    // the trailing-position lookahead only accepts trivia followed by
+    // the CLOSER - a comment cannot smuggle in a missing comma
+    expect(arrayOk("[1 // note\n 2]")).toBe(false);
+  });
+
+  // The comment placements the rewrite legitimately added, all pinned
+  // in dataStructures.test.ts too - re-asserted here beside the
+  // rejections so the whole delimiter contract reads in one place.
+  it("keeps comments after a comma", () => {
+    expect(arrayOk("[\n  1,\n  // keep me\n  2\n]")).toBe(true);
+    expect(objectOk(`{\n  "a": 1,\n  // keep me\n  "b": 2\n}`)).toBe(true);
+  });
+
+  it("keeps trailing comments without a trailing comma", () => {
+    expect(objectOk(`{\n  // lead\n  "a": 1\n  // trail\n}`)).toBe(true);
+    expect(arrayOk("[\n  1\n  // trail\n]")).toBe(true);
+  });
+});
--- a/packages/agency-lang/lib/parsers/parsers.ts
+++ b/packages/agency-lang/lib/parsers/parsers.ts
@@ -1303,9 +1303,14 @@ type InterleavedEntry<T> = ItemEntry<T> | TriviaEntry;
 // so `partitionTrivia` can distinguish it — no manual result unwrapping.
 // `multiLineCommentParser` does NOT eat its trailing newline, which the
 // trailing `optionalSpacesOrNewline` handles so `many(...)` keeps progressing.
+/** One comment/blank-line unit, shared between triviaEntry (which
+ *  captures it) and literalDelimiter's trailing-position lookahead
+ *  (which only scans past it). */
+const literalTrivia = or(blankLineParser, commentParser, multiLineCommentParser);
+
 const triviaEntry: Parser<TriviaEntry> = seqC(
   set("kind", "trivia"),
-  capture(or(blankLineParser, commentParser, multiLineCommentParser), "node"),
+  capture(literalTrivia, "node"),
   optionalSpacesOrNewline,
 ) as Parser<TriviaEntry>;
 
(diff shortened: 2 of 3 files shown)


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4706,29 +4727,64 @@ export const comprehensionParser: Parser<Comprehension> = label(
         optionalSpacesOrNewline,
         str("for"),
```
egonSchiele, line 4728: The commit needs `for` plus at least one whitespace (`str("for"), spaces`), so the targeted-message coverage is whitespace-gated in a way that misses a common half-typed shape.

`[x for ]` (a space before the bracket) commits and gives the nice "expected a binder" error. `[x for]` (no space) does not: `spaces` fails before the commit, the parser backtracks to the array, and the array fails on the comma rule with a generic message.

That no-space case is not exotic. An editor auto-closing the bracket turns a half-typed `[x for` straight into `[x for]`, which is exactly the #602 population. Both still error, so the silent-array bug is gone either way, but the targeted message the PR is built to deliver only appears when the user happened to type a trailing space.

Worth either handling `[x for]` explicitly or noting the gap, since the difference is invisible from the PR description.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4706,29 +4727,64 @@ export const comprehensionParser: Parser<Comprehension> = label(
...
+          parseError(
+            "expected a binder name or pattern after `for` in this list comprehension",
+            ...iterationBinderFragment,
+          ),
+        ),
+        captureCaptures(
+          parseError(
+            "expected `in` after the list comprehension binder",
+            optionalSpacesOrNewline,
+            str("in"),
```
egonSchiele, line 4753: `str("in")` here, and `str("if")` in the filter block below, are not word-bounded, and now that they sit inside commit points a trailing identifier throws a misattributed error.

`[x for x insomething]`: `str("in")` matches the `in` inside `insomething`, then the iterable block's required `spaces` fails on `something`, and the commit throws "expected an iterable expression after in". `[x for x in xs iffy]` throws "expected a filter condition after if" the same way. The real problem in both is that the token was not a standalone keyword.

`str("for")` avoids this because its required trailing `spaces` happens to double as a boundary and the message still fits. `in` and `if` do not get that luck.

This is strictly better than the old silent-array parse, and no *valid* program can hit it (the keyword slot is grammatically fixed), so it is message quality on already-broken input, not a correctness bug. But there is an in-file precedent worth reusing: `ifExpressionParser` at line 4478 word-bounds with `str("if"), not(varNameChar)`. The same guard here would land the error at the right spot.

packages/agency-lang/lib/parsers/literalDelimiter.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    expect(arrayOk("[\n  1,\n  2,\n  3\n]")).toBe(true);
+    expect(objectOk(`{\n  "a": 1,\n  "b": 2\n}`)).toBe(true);
+  });
+
+  it("keeps splats", () => {
+    expect(arrayOk("[...xs, 1]")).toBe(true);
+  });
+
+  // The 2026-07-04 comments-between-entries rewrite made the comma
+  // optional EVERYWHERE, so any whitespace separated items and
```
egonSchiele, line 39: The comma requirement is a breaking change to a grammar that has been accepted for two weeks, and the sweep that backs "nothing relied on it" covered `stdlib/`, `tests/agency`, `tests/agency-js`, and `lib/agents`. Were `examples/` and any `.agency` embedded in the guide docs included?

Those are the files most likely to be hand-written casually with a newline instead of a comma, and they are the ones a new user copies. The regression window is short so the wild-breakage risk is genuinely low, but examples are the one corpus where a comma-less multi-line array would be both plausible and user-facing. A one-line confirmation they were swept would close it.

egonSchiele (commented):
Read the whole diff and traced the grammar changes by hand. This is a clean fix, and finding that the array fallback was itself a two-week-old regression is the kind of root-cause the issue needed rather than a patch on top of the symptom.

**Verified rather than assumed:**

- **The commit point is safe against valid arrays.** The commit fires after `[ <expr> for <ws>`. For that to break a valid program there would have to be a valid non-comprehension parse of something starting `[expr for`. The restored comma requirement forecloses it: an array needs a comma after its first item, and `for` is not a comma, so no valid array can reach the commit. Since `[...]` at atom position is only ever array-or-comprehension, committing here cannot strand a valid parse.
- **The `[for x in xs]` reasoning is accurate.** I checked on main: `const x = for` parses `for` as a `variableName`, so `for` really is a legal body expression and the input diverges through the comma rule exactly as the PR says, not by luck.
- **"Strictly narrower, no accepted parse changes shape" holds.** Trivia is still consumed by `triviaEntry`; the delimiter's new branch only `peek`s past it. So an input that still parses keeps its old AST, which is what makes the "type-level output untouched by construction" claim real.

Three notes, all non-blocking. The first is the one I would look at.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Fixed in ceae2dc6c. The commit prefix is now str(for) plus a not(varNameChar) word boundary instead of required whitespace, with the required spaces moved inside the binder commit - so [x for] (the editor-auto-close shape) reaches the commit and reports the missing binder, [x for ] likewise, and forever still fails the boundary BEFORE the commit and falls through to the array rule. All three pinned.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Fixed in ceae2dc6c using the ifExpressionParser precedent you pointed at: not(varNameChar) after str(in) inside its commit and after str(if) in the filter clause. [x for x insomething] now reports the missing in at the fused token, and [x for x in xs iffy] fails the if boundary, backtracks the optional filter, and reports the unclosed bracket instead of blaming a filter condition. Both pinned with message assertions.

packages/agency-lang/lib/parsers/literalDelimiter.test.ts
egonSchiele: Swept and clean, with the evidence now gathered rather than assumed: tc over ../examples (2 files) and the local examples/ tree (8 files) - zero parse failures - plus a script that extracted every fenced code block from docs/site/guide, docs/site/stdlib, and docs/site/cli and flagged array-literal item lines not ending in a comma or bracket. Four candidates, all false positives on inspection: two legal last-items-without-trailing-comma, one TypeScript (not Agency) snippet in why-agency.md, and my own multi-line comprehension example. Nothing user-facing relied on comma-less literals.

PR #611, opened 2026-07-19
A handler never hears its own raises: allow interrupts inside handler functions

egonSchiele:
## Why

A handler function could not call anything that raises an interrupt: the dispatcher visits every handler, so a handler that raised was re-entered for the interrupt it just caused and recursed until `HandlerRecursionError`. AG3010 banned the shape wholesale. The practical cost: an LLM reviewer inside a handler had to be tool-free (the #600 supervisor gathers evidence by shelling out to git because `read`/`gitStatus` were unavailable to it).

Plan: `docs/superpowers/plans/2026-07-19-interrupts-in-handler-functions.md` (in the `worktree-handler-interrupts` worktree), design evolved from the spec's regions concept to the final rule in review discussion.

## The rule

**A handler never hears its own raises.** While a handler entry executes (tracked per async lineage in AsyncLocalStorage), the chain walk skips that entry exactly like an ineligible one, and the rest of the chain decides. A raise nothing settles — including anything an outer handler propagates — cannot pause to ask the user, so `renderVerdict` refuses it as a rejection with an explanatory message and the handler continues degraded.

Safety is fail-closed: an excluded handler contributes nothing (not an approve), so a handler's raise executes only if an outer handler approves it or the author wrote a visible `with approve` — whose approve an outer reject still beats, and an outer propagate still refuses. `with approve` keeps its one ordinary meaning everywhere; no new syntax, no codegen changes.

## Changes

- **`lib/runtime/executingHandlers.ts`** (new): per-lineage executing-handler stack. Never needs checkpointing — no interrupt raised inside a handler ever surfaces to pause a run.
- **`lib/runtime/interrupts.ts`**: the skip in `runHandlerChain` (silent, beside the `eligible` check) + the refusal in `renderVerdict`. No `resolvedBy` gate — `renderVerdict` has one call site and `resolvedBy` is "ipc" whenever a parent process decided, so gating would skip the refusal for a subprocess handler raise under a propagating parent.
- **AG3010 retired**, not deleted: registry entry keeps its code with `retired: true` per the registry's append-only convention; the check, its registration, and its tests are removed; `agency explain AG3010` explains the retirement. Breaking change accepted by the owner.
- **`docs/site/guide/handlers.md`**: the "Handlers can't raise interrupts" section replaced with the rule, the reviewer example, and three caveats (unsettled raises are rejected; propagate beats in-handler `with approve`; skip is per activation).

## Tests

- Unit: `executingHandlers.test.ts` (6 — attribution, unwind on finish/throw, nesting, cross-lineage leak); the `agencyInterrupt` recursive-handler suite rewritten to pin the new semantics (self-raise refused with the effect named, sibling activations hear once, no state leak into the next dispatch).
- Execution fixtures (`tests/agency/handlers/`), all with observables encoded in `expectedOutput` so CI asserts them: outer approver works with a bare read and the handler body runs exactly once (`read:ok|reviews:1`); ordinary `with approve` works with no outer approver; outer reject wins (`read:rejected`); outer propagate refuses without killing the run (`read:refused`); 12 sequential raises do not accumulate chain depth (`reads:12`); a guard trip raised mid-handler reaches the outer handler (`trips:1`).
- `tests/agency-js/handler-fork-isolation/`: exclusion does not leak across fork branches. Agency-js because the ask-counts cannot ride the return value — the sibling's handler runs in the raiser's lineage, so counter writes land in the raising branch's isolated state and vanish at the join; the JS side counts the printed ASKED lines (exactly one per branch).

Interactive mode manually verified: the propagate fixture runs to completion under `--interactive` with no `Cannot throw` error.

## Notes for review

- Execution-test relative paths resolve against the fixture's own directory, so fixtures read their own file rather than `package.json`.
- Found in passing: `x is success(v)` / `x is failure(e)` narrowing is not applied inside handler bodies (AG2009 false positive); `isSuccess`/`isFailure` work fine and the fixtures use those. Worth a follow-up issue.
- Follow-up (not here): migrate the #600 supervisor's reviewer to `llm(..., tools: [gitStatus, gitDiff, read])` now that handler tools work.


packages/agency-lang/docs/site/diagnostics/effects.md
```diff
@@ -96,13 +96,13 @@ This function may raise interrupts, but it is called from a place that is not in
 
 <a id="ag3010"></a>
 
-## AG3010 — Handler &#123;handler&#125; may raise interrupts [&#123;effects&#125;]. That would re-enter the handler chain (the dispatcher visits every handler, even the one currently running) and recurse until `HandlerRecursionError` fires at runtime. Restructure so the handler doesn't call interrupt-raising code (e.g. hoist file I/O out of the handler), or suppress this error with `// @tc-ignore` on the line above the `handle` block.
+## AG3010 — RETIRED: handler functions may raise interrupts. The dispatcher skips the executing handler for its own raises, so the recursion this diagnostic guarded against cannot happen. Remove any `// @tc-ignore AG3010` suppressions.
```
egonSchiele: just delete, don't mark "retired"

egonSchiele (commented):
Reviewed with the plan and the runtime code side by side. The core change is correct and I could not break the safety argument.

**What I verified against the code, not just the diff:**

- **The skip is in the right place and the right shape.** It sits beside the `eligible` skip, is silent the same way (no invocation, no statelog decision), and `executing` is safely hoisted above the loop — the walk runs synchronously in one lineage, so the set cannot change between entries.
- **The refusal gate is right.** `insideHandlerFunction()` alone, no `resolvedBy` gate — I confirmed `renderVerdict` has exactly one call site (the origin dispatch) and the IPC relay (`ipc.ts` → `gatherChainOutcome`) never reaches it, so a subprocess handler raise under a propagating parent lands in the refusal rather than resurrecting the `Cannot throw` crash. The refusal message correctly avoids recommending `with approve` (propagation beats collected approvals in the merge).
- **The handlers-are-safety-infrastructure invariant is pinned, not assumed.** `reviews:1` in the outer-approves fixture fails on both re-entry (2) and wrongful exclusion (0), and the `with approve` fixture pins that entries registered *during* an excluded handler execution are not themselves excluded. The cross-lineage ALS unit test genuinely discriminates — a module-global implementation would fail it.
- **Fail-closed holds on every path I traced**: excluded entry contributes nothing; approve requires an outer handler or visible `with approve`; outer reject beats it; propagate/noResponse refuse as rejection. The outer-rejects and propagate-refused fixtures assert the two ways this could silently regress.
- **Fixture coverage matches the plan** — all six execution fixtures plus the fork-isolation agency-js test, each with the failure modes encoded in `expectedOutput`.

Five small things inline: a stale comment block narrating the removed semantics, assertions weaker than the deterministic flow allows, a grammar slip, a dynamic import the repo bans, and a docs example that would not typecheck as written.

One follow-up worth filing as an issue (already noted in the PR body): the `x is success(v)` narrowing gap inside handler bodies, so it does not get lost.

packages/agency-lang/lib/runtime/agencyInterrupt.test.ts
```diff
@@ -290,96 +290,116 @@ describe("agency.interrupt — frame requirements", () => {
 // to unbounded recursion. The runtime now bounds nested-dispatch depth
 // at MAX_HANDLER_CHAIN_DEPTH and throws HandlerRecursionError when
 // exceeded, naming the offending interrupt kind in the message.
-describe("agency.interrupt — recursive-handler guard", () => {
-  it("throws HandlerRecursionError when a handler keeps re-entering itself", async () => {
+describe("agency.interrupt — raising inside a handler", () => {
```
egonSchiele, line 295: The comment block above this `describe` (lines 286–292) still tells the old story: the dispatcher "visits the same handler again … leading to unbounded recursion. The runtime now bounds nested-dispatch depth at MAX_HANDLER_CHAIN_DEPTH and throws HandlerRecursionError". That is exactly the behavior this PR removes — none of the rewritten tests throw `HandlerRecursionError`, and the suite now pins self-exclusion and refusal. Worth rewriting the header to describe the new rule (a handler never hears its own raises; unsettled in-handler raises are refused as rejections), with the depth guard mentioned only in its residual backstop role for sibling activations.

packages/agency-lang/lib/runtime/agencyInterrupt.test.ts
```diff
@@ -290,96 +290,116 @@ describe("agency.interrupt — frame requirements", () => {
...
+            }),
+          );
+        });
+        return "done";
+      }),
+    );
+    // The outer kickoff runs both activations; each activation's own raise
+    // is heard by the other side exactly as the chain allows: the deepest
+    // raise (both executing) is refused.
+    expect(heardInner).toBeGreaterThan(0);
```
egonSchiele: These assertions are weaker than the flow allows. The dispatch order here is deterministic — tracing it: the outer kickoff visits both entries; each activation's raise is heard by the other side exactly once, and each of those cross-heard raises re-raises into a dispatch where both entries are executing, which is refused. That gives `heardInner === 2` and `refusals === 2`, every run.

`toBeGreaterThan(0)` cannot catch the regression the test name promises to catch: if exclusion broke and an activation were re-entered for its own raise, `heardInner` would inflate past 2 (or the depth guard would fire) — and `> 0` still passes right up until the crash. Suggest pinning the exact counts:

```ts
expect(heardInner).toBe(2);
expect(refusals).toBe(2);
```

packages/agency-lang/lib/typeChecker/handlerParamTyping.ts
```diff
@@ -81,7 +81,7 @@ function handlerParamType(
  *
  * Runs each scope under `ctx.withScope(info.scopeKey, …)` so `collectRaisableEffects`
  * → `synthType` resolves scope-local type aliases against the right scope (mirrors
- * `checkHandlerBodyInterrupts`).
+ * the retired AG3010 check once did).
```
egonSchiele: Grammar: "(mirrors the retired AG3010 check once did)" reads as a broken sentence. Suggest "(mirrors what the retired AG3010 check once did)" — or simpler, "(as the retired AG3010 check did)".

packages/agency-lang/tests/agency-js/handler-fork-isolation/test.js
```diff
@@ -0,0 +1,32 @@
...
+// only cross-branch observable: handler counter writes land in the
+// raising branch's isolated state and vanish at the join.
+const lines = [];
+const originalLog = console.log;
+console.log = (...parts) => {
+  lines.push(parts.join(" "));
+  originalLog(...parts);
+};
+
+const { runForkIsolation } = await import("./agent.js");
```
egonSchiele: This is a dynamic import, which the repo bans outright (CLAUDE.md: "NEVER use dynamic imports"), and the existing agency-js tests all use static imports. It also is not needed for the interception: the ASKED/READ prints happen while `runForkIsolation()` executes, not at module-init time, so a static import at the top with the `console.log` patch installed before the *call* captures everything this test asserts:

```js
import { runForkIsolation } from "./agent.js";

const lines = [];
const originalLog = console.log;
console.log = (...parts) => { lines.push(parts.join(" ")); originalLog(...parts); };
const result = await runForkIsolation();
console.log = originalLog;
```

If agent.js module init turns out to print something the count depends on, that would be worth a comment explaining why the ordering matters — but as the assertions stand, it does not.

packages/agency-lang/docs/site/guide/handlers.md
```diff
@@ -190,7 +190,31 @@ The rules of handlers are thus:
...
+A handler's body may raise interrupts, directly or through anything it calls. One rule makes this safe: **a handler never hears its own raises.** Every handler up the chain runs on every interrupt, so a handler that raised would otherwise be called to review its own raise, call itself again, and loop forever. Instead, the dispatcher skips the handler that is currently executing, and every *other* handler decides — an outer policy or an outer reject governs a handler's tool calls exactly as it governs everything else.
+
+This is what lets a reviewer inside a handler use tools:
+
+```agency
+handle {
+  agentDoesWork()
+} with (data) {
+  const verdict = llm("Should I allow this?", tools: [read, grep])
+  if (verdict.ok) {
```
egonSchiele, line 204: This example would not typecheck as written: with no expected type, `llm(...)` returns a string, so `verdict.ok` is a property access on a string. Since structured output is driven by the declared type, annotating the binding fixes it and also shows readers the idiom:

```agency
const verdict: { ok: boolean } = llm("Should I allow this?", tools: [read, grep])
```

Guide examples get copy-pasted, so this one failing on first contact would undercut the section.

packages/agency-lang/docs/site/diagnostics/effects.md
egonSchiele: Done in 62e189247: the docs generator now omits retired diagnostics entirely (effects.md and the index row are gone). The registry keeps the entry per its append-only convention, so the code stays reserved and `agency explain AG3010` still answers for anyone with old suppressions — say the word if you want the registry entry fully deleted too.

packages/agency-lang/lib/runtime/agencyInterrupt.test.ts
egonSchiele: Rewritten in 62e189247: the header now states the self-exclusion rule and the refusal, with the depth guard described in its residual backstop role and the old thread link kept for history.

packages/agency-lang/lib/runtime/agencyInterrupt.test.ts
egonSchiele: Fixed in 62e189247: exact counts (heardInner === 2, refusals === 2), with a comment recording the deterministic trace.

packages/agency-lang/lib/typeChecker/handlerParamTyping.ts
egonSchiele: Fixed in 62e189247: "(as the retired AG3010 check did)".

packages/agency-lang/tests/agency-js/handler-fork-isolation/test.js
egonSchiele: Fixed in 62e189247: static import, with a comment noting the prints happen at call time so the ordering is safe.

packages/agency-lang/docs/site/guide/handlers.md
egonSchiele: Fixed in 62e189247: the example annotates the binding (const verdict: { ok: boolean } = ...).

packages/agency-lang/tests/agency-js/handler-fork-isolation/test.js
```diff
@@ -0,0 +1,32 @@
...
+const result = await runForkIsolation();
+console.log = originalLog;
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 62e189247: the run is wrapped in try/finally so console.log is always restored.

packages/agency-lang/lib/runtime/agencyInterrupt.test.ts
```diff
@@ -290,96 +290,116 @@ describe("agency.interrupt — frame requirements", () => {
...
+    expect(refusals).toBeGreaterThan(0);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 62e189247 (same change as the other exact-counts thread): toBe(2) for both.

PR #613, opened 2026-07-19
Agent supervisor: stdlib verifier, brainstorm before complex solves, draft deltas

egonSchiele:
## Why

The supervisor from #600 checks on the code subagent every five minutes, but its reviewer had to be tool-free: nothing inside a handler function could raise an interrupt, so evidence-gathering was limited to direct git reads. #611 removed that restriction. This PR gives the supervisor the standard agent built for exactly this job, adds a think-before-committing step for complex tasks, and teaches the supervisor to read the solver's own progress reports.

## Supervisor uses std::agents/verifier

`progressCheck` now sends `verifierAgent` into the workspace at each pause — full toolset (shell, file reads, real measurements), governed by the outer policy like any other tool call. Its context frames the run as a mid-run check ("the work is expected to be incomplete; report which criteria are satisfied, in progress, or unstarted"), and the reviewer LLM judges those findings plus the check history instead of a raw git diff. Budgets are small ($2 / 2m) because it is a pulse check, and everything fails open — including the ambiguity that `verifierAgent` itself fails open (`failOpenFeedback` maps failures to `success([])`), so empty findings are reported as "either nothing was measurable yet or it could not run", never as a clean measurement.

On the other stdlib agents: `verifierAgent` is the right fit here (it *runs and measures*; `reviewAgent` only reads text; `oracleAgent` is a slow-model second opinion, too heavy for a five-minute cadence). The code subagent already uses the stdlib expert/coding/verifier agents for the solve itself.

## Brainstorm before complex solves

When triage says a task is complex, `escalateToExpert` now brainstorms before solving: three proposers run in parallel via `fork`, each pushed toward a genuinely different bias — simplest thing that could work, most robust, maximum reuse — and a judge picks one with a one-sentence rationale.

- The **chosen** direction goes into the solver's context ("follow this unless you discover it cannot work").
- The **set-aside alternatives** go to the supervisor, which shows them to the reviewer at every check — so a redirect can say "switch to the reuse approach" by name instead of a vague "try something else".

The whole step runs under its own guard ($2 / 3m) and fails open: an errored brainstorm, or a judge that returns an out-of-range index, yields an empty result and the solve proceeds exactly as before.

## Draft deltas

Guard trips already carry a preview of the innermost saved draft (`draftValue`, attached by `guardTripInterrupt` and documented in the guards guide) — nobody consumed it. Now:

- `std::supervise` passes it to the check callback: `check: (elapsed, draft) -> SuperviseDecision`. Signature change to the stdlib; the existing supervise fixtures are updated and a new one (`checkSeesDraft`) pins that a block's `saveDraft` value reaches the check.
- The agent's supervisor stringifies and compares each check's draft against the previous one. The delta is computed, not judged, and it is the sharpest stall signal available for free: a thrashing solver churns the workspace — which git and even the verifier can read as motion — but its own "here is what I have finished" statement does not advance.
- The thrash backstop gains a deterministic input: reviewer-says-stalled **plus** a byte-identical draft stops immediately, instead of waiting for a second stalled check. An unchanged draft alone never stops a progressing run, and a solver that never saves drafts degrades cleanly (delta "none", no signal).

## Tests

25 execution tests across three suites, no LLM calls:

- `tests/agency/supervise/` (7): existing behavior plus `checkSeesDraft`.
- Agent supervisor suite (14): thrash backstop including the two new draft rules, outcome flattening, draft stringification (null/string/typed-object), all four delta states, and the prompt section flagging only the byte-identical case.
- Brainstorm suite (4): the judge-choice splitter keeps order, a bad index fails open, an empty brainstorm renders to nothing, and the renderings carry the right names to the right places.

## Breaking change

`std::supervise`'s `check` callback signature changed from `(elapsed) -> SuperviseDecision` to `(elapsed, draft) -> SuperviseDecision`. Every in-repo caller is updated; external check callbacks written to the old shape need the new parameter. Same few-users call as retiring AG3010.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/subagents/brainstorm.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/subagents/brainstorm.agency
@@ -55,7 +55,9 @@ type Choice = {
 
 /** Split the judge's pick from the rest. Pure, so tests can pin the
   out-of-range behavior: a bad index means no approach was chosen and
-  everything stays an alternative. */
+  everything stays an alternative. The rationale is kept even then — it
+  never renders anywhere when nothing was chosen, and it is the one
+  breadcrumb explaining why the judge went out of range. */
 export def splitChoice(
   approaches: Approach[],
   chosenIndex: number,
@@ -66,7 +68,7 @@ export def splitChoice(
     return {
       chosen: null,
       alternatives: approaches,
-      rationale: ""
+      rationale: rationale
     }
   }
   const alternatives = []
--- a/packages/agency-lang/lib/agents/agency-agent/subagents/brainstorm.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/subagents/brainstorm.agency
@@ -1,0 +1,193 @@
+/*
+ * Brainstorm-before-solving: when a task triages as complex, spend a
+ * moment generating genuinely different directions before committing to
+ * one, instead of running with the first idea.
+ *
+ * Three proposers run in parallel via `fork`, each pushed toward a
+ * different bias (simplest thing, most robust thing, most reuse). A
+ * judge picks the one to pursue. The chosen approach travels into the
+ * solver's context; the rejected ones travel to the supervisor, so a
+ * mid-run redirect can name a specific alternative when the chosen path
+ * is not working.
+ *
+ * Everything fails open: a brainstorm that errors or runs out of budget
+ * returns empty renderings, and the solve proceeds exactly as it would
+ * have without one.
+ */
+import { map } from "std::array"
+
+// Deliberately opinionated biases, so the three proposals differ in kind
+// rather than phrasing.
+static const ANGLES = [
+  "the simplest thing that could possibly work — smallest change, fewest moving parts",
+  "the most robust and thorough approach — handles edge cases, verifiable at every step",
+  "maximum reuse — lean on existing code, tools, and conventions instead of building anything new",
+]
+
+static const BRAINSTORM_MAX_COST = $2.00
+static const BRAINSTORM_MAX_TIME = 3m
+
+/** One proposed direction for the task. */
+export type Approach = {
+  name: string;
+  plan: string;
+  risk: string
+}
+
+/** What a brainstorm produces: the approach to pursue, the ones set
+  aside, and why the judge chose as it did. */
+export type BrainstormResult = {
+  chosen: (Approach | null);
+  alternatives: Approach[];
+  rationale: string
+}
+
+static const EMPTY_BRAINSTORM: BrainstormResult = {
+  chosen: null,
+  alternatives: [],
+  rationale: ""
+}
+
+type Choice = {
+  chosenIndex: number;
+  rationale: string
+}
+
+/** Split the judge's pick from the rest. Pure, so tests can pin the
+  out-of-range behavior: a bad index means no approach was chosen and
+  everything stays an alternative. */
+export def splitChoice(
+  approaches: Approach[],
+  chosenIndex: number,
+  rationale: string,
+): BrainstormResult {
+  const chosen = approaches[chosenIndex]
+  if (chosen == null) {
+    return {
+      chosen: null,
+      alternatives: approaches,
+      rationale: ""
+    }
+  }
+  const alternatives = []
+  let index = 0
+  for (approach in approaches) {
+    if (index != chosenIndex) {
+      alternatives.push(approach)
+    }
+    index = index + 1
+  }
+  return {
+    chosen: chosen,
+    alternatives: alternatives,
+    rationale: rationale
+  }
+}
+
+def renderApproach(approach: Approach): string {
+  return """- ${approach.name}
+  Plan: ${approach.plan}
+  Risk: ${approach.risk}"""
+}
+
+/** The chosen-direction block for the solver's context, or "" when no
+  brainstorm ran (or it failed open). Pure. */
+export def renderChosen(result: BrainstormResult): string {
+  const chosen = result.chosen
+  if (chosen == null) {
+    return ""
+  }
+  return """
+The direction to pursue, chosen after comparing several (${result.rationale}):
+${renderApproach(chosen)}
+
+Follow this direction unless you discover it cannot work; if it fails, say so plainly in your summary rather than silently switching."""
+}
+
+/** The alternatives block for the supervisor, or "" when there are none.
+  Pure. */
+export def renderAlternatives(result: BrainstormResult): string {
+  if (result.alternatives.length == 0) {
+    return ""
+  }
+  return map(result.alternatives, renderApproach).join("\n")
+}
+
+def proposeApproach(task: string, guidance: string, angle: string): Approach {
+  const proposal: Approach = llm(
+    """
+Propose ONE concrete approach to the task below, favoring ${angle}.
+
+<task>
+${task}
+</task>
+
+${guidance}
+
+Return:
+- name: a short label for the approach
+- plan: the concrete steps, 2 to 5 sentences
+- risk: the main way this approach fails
+
+Commit to the bias you were given — do not hedge toward a middle ground.
+""",
+  )
+  return proposal
+}
+
+def judgeApproaches(task: string, approaches: Approach[]): Choice {
+  const rendered = []
+  let index = 0
+  for (approach in approaches) {
+    rendered.push("[${index}] ${renderApproach(approach)}")
+    index = index + 1
+  }
+  const choice: Choice = llm(
+    """
+Pick the best approach for this task.
+
+<task>
+${task}
+</task>
+
+<approaches>
+${rendered.join("\n\n")}
+</approaches>
+
+Return chosenIndex (the [n] of the winner) and rationale (one sentence,
+naming what tipped the decision).
+""",
+  )
+  return choice
+}
+
+export def brainstormApproaches(
+  task: string,
+  guidance: string = "",
+): BrainstormResult {
+  """
+  Generate several genuinely different directions for a complex task in
+  parallel, then pick one to pursue. Returns the chosen approach, the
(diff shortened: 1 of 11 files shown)


packages/agency-lang/lib/agents/agency-agent/subagents/supervisor.agency
```diff
@@ -79,53 +93,33 @@ export def applyThrashRule(
...
+    maxCost: CHECK_MAX_COST,
+    maxTime: CHECK_MAX_TIME,
+  )
+  if (isFailure(checked)) {
     return {
-      text: "(no git repository — no workspace evidence available; judge from the check history alone)",
-      filesChanged: 0
+      text: "(the verifier could not run: ${checked.error} — judge from the check history alone)",
+      findings: 0
     }
```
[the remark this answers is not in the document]
egonSchiele: Right on both points — verified failOpenFeedback maps every failure to success([]). Fixed in 8829da425 + c8fd17c74: the dead isFailure branch is gone, empty findings now report honestly ("either nothing was measurable yet or it could not run — judge from the check history alone"), and the result is read through `checked catch []` since the checker cannot see the fail-open contract.

packages/agency-lang/lib/agents/agency-agent/subagents/supervisor.agency
```diff
@@ -2,26 +2,34 @@
...
+ * rejection) yields no findings, and the reviewer judges from history
+ * alone rather than the check crashing the run.
  */
-import { gitIsRepo, gitStatus, gitDiff } from "std::git"
+import { verifierAgent } from "std::agents/verifier"
+import {
+  Feedback,
+  feedbackHasErrors,
+  renderFeedback,
+ } from "std::agents/lib/feedback"
```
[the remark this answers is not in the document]
egonSchiele: Removed in 8829da425.

packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
```diff
@@ -285,15 +291,22 @@ export def codeAgent(userMsg: string, agencyTask: boolean = false) {
...
+   path is not working. `escalate`, the mid-flight tool, still uses
+   plannerAgent. */
 def escalateToExpert(task: string, agencyTask: boolean): string {
   const guidance = guidanceOrEmpty(consultExpert(task: task, agencyTask: agencyTask))
+  const ideas = brainstormApproaches(task: task, guidance: renderGuidance(guidance))
+  const alternatives = renderAlternatives(ideas)
   const solved = supervisedSolve(
     task: task,
-    context: renderGuidance(guidance),
+    context: "${renderGuidance(guidance)}\n${renderChosen(ideas)}",
```
egonSchiele: This defeats `withContext`'s empty-context fast path. Both parts return `""` in their fail-open paths — `renderGuidance` on an empty consult, `renderChosen` when the brainstorm failed open — and when both do, this interpolation produces `"\n"`, not `""`. `withContext` gates on exactly `context == ""`, so the solver's prompt gets an empty `<context>` block instead of running untouched, silently breaking the invariant `renderGuidance` documents ("the coding agent runs exactly as it would without a consult"). Before this PR the `""` passed through intact.

Join only the non-empty parts instead:

```agency
const parts = filter([renderGuidance(guidance), renderChosen(ideas)], \part -> part != "")
const solved = supervisedSolve(
  task: task,
  context: parts.join("\n"),
  ...
```

packages/agency-lang/stdlib/supervise.agency
```diff
@@ -31,15 +31,19 @@ export type SuperviseDecision = {
 export def supervise(
   every: number,
   maxTime: number,
-  check: (elapsed: number) -> SuperviseDecision,
+  check: (elapsed: number, draft: any) -> SuperviseDecision,
```
egonSchiele, line 34: Two things on this signature change:

1. **The regenerated doc page is missing from the PR.** This docstring edit changes `docs/site/stdlib/supervise.md`, and the regenerated file is sitting modified-but-uncommitted in the branch worktree (7 insertions, 3 deletions). Commit it — CI or the next `agency doc` run will surface the drift otherwise.
2. **This is a breaking change to a public stdlib API.** Every in-repo caller is updated (code.agency, the supervise tests), but any external `check` callback written to the old `(elapsed) -> SuperviseDecision` shape now mismatches the declared function type. Given the language's few users that is presumably fine — same call as retiring AG3010 — but it deserves a line in the PR body.

packages/agency-lang/lib/agents/agency-agent/subagents/brainstorm.agency
```diff
@@ -0,0 +1,193 @@
...
+  approaches: Approach[],
+  chosenIndex: number,
+  rationale: string,
+): BrainstormResult {
+  const chosen = approaches[chosenIndex]
+  if (chosen == null) {
+    return {
+      chosen: null,
+      alternatives: approaches,
+      rationale: ""
```
egonSchiele: Nit: the bad-index branch discards the judge's rationale (`rationale: ""`). Defensible — a confused judge's rationale is not worth showing to the solver — but it is also the one breadcrumb explaining *why* the judge went out of range when someone debugs a no-choice brainstorm. Consider keeping it: the empty-`chosen` case never renders the rationale into any prompt (`renderChosen` returns `""` when chosen is null), so preserving it costs nothing visible.

packages/agency-lang/lib/agents/agency-agent/subagents/supervisor.agency
```diff
@@ -2,26 +2,36 @@
...
 import { SuperviseDecision } from "std::supervise"
 
 import { truncate } from "../lib/utils.agency"
 
-// The raw patch can be huge; the reviewer needs the shape of the work, not
-// every hunk.
-static const MAX_PATCH_CHARS = 6000
+// The verifier is a pulse check here, not a final gate: small budgets,
+// because it runs every interval and its findings only steer.
+static const CHECK_MAX_COST = $2.00
```
egonSchiele, line 36: Worth stating in the comment whose budget this spends: the check handler registered before the solver's own cost guard went live, so that guard is suspended during the check and the verifier's spend charges whatever sits *above* the solve (session/user budget), bounded only by these caps. That is arguably the right place for supervision overhead to land — the supervisor is not the solver's work — but the comment currently reads as if `$2/2m` is the whole story, and ~6 checks on a 30-minute solve is up to ~$12 nobody's solve-level budget accounts for.

egonSchiele (commented):
Reviewed at 56a5460 (the draft-deltas revision). The verifier swap, the brainstorm step, and the new draft-delta stall rule are all sound; I verified the contracts against the code rather than the diff. One small real bug in context assembly, one missing generated file, and two notes.

**What I verified:**

- `verifierAgent(task, context, maxCost, maxTime) -> Result<Feedback[]>` matches the call, `renderFeedback` takes the full Result as passed, and the verifier is the right agent for a pulse check (read-only + shell tools, cannot write). The `catch []` fallback plus the honest "either nothing was measurable yet or it could not run" text resolves the empty-findings ambiguity cleanly.
- **The draft-delta rule is built on solid ground**: despite its name, `draftPreview` (guardTripInterrupt.ts) returns the *full* innermost saved draft value, not a truncation — so the byte-identical comparison cannot false-positive on long drafts that differ past a preview horizon, and `MAX_DRAFT_CHARS` truncation is display-only. `checkSeesDraft` pins the plumbing end to end, and the thrash-rule tests cover both the two-signal stop and the unchanged-draft-alone-does-not-stop case.
- Fail-open paths all trace clean: failed verifier, failed brainstorm guard, bad judge index (index normalization also covers negative and fractional values), empty renderings everywhere.
- `catch` expression syntax, paren-less if-expressions, `std::array` imports, and `static const` with `$`/duration literals all match the corpus. No `filesChanged` stragglers anywhere.

**The one behavior bug** is the whitespace-only context (inline comment on code.agency) — everything else is small.

One cost observation, not a defect: each five-minute check can now spend up to $2 of verifier plus the reviewer call, where the old check was one tool-free `llm()`. A 30-minute solve makes ~6 checks, so up to ~$12 of supervision overhead — and the verifier's spend lands *outside* the solver's own budget guard (that guard was not live when the supervise handler registered, so it is suspended during the check). Both seem like reasonable prices for measured evidence; flagging so the budgets are a decision rather than a default.

packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
egonSchiele: Fixed in c072534e8 exactly as suggested: non-empty parts filtered and joined, with a comment naming the withContext empty-gate invariant this protects.

packages/agency-lang/stdlib/supervise.agency
egonSchiele: Both done in c072534e8: the regenerated docs/site/stdlib/supervise.md is committed, and the PR body now carries a breaking-change note for the check signature.

packages/agency-lang/lib/agents/agency-agent/subagents/brainstorm.agency
egonSchiele: Kept in c072534e8 — you are right that it costs nothing visible (renderChosen returns "" when chosen is null) and it is the only breadcrumb for debugging a no-choice brainstorm. The bad-index test now pins the preserved rationale.

packages/agency-lang/lib/agents/agency-agent/subagents/supervisor.agency
egonSchiele: Comment expanded in c072534e8: it now states the spend charges the session ledger above the solve (the solver's guard is suspended during checks, having not been live at handler registration), and gives the ~checks × CHECK_MAX_COST arithmetic. Agreed that is the right ledger for supervision overhead; the budgets are now a documented decision.

PR #614, opened 2026-07-19
Supervise the planner escalation; partial results at every agent milestone

egonSchiele:
## Why

Two follow-ups from surveying the agent after #613: the `escalate` → planner path was the last big unsupervised budget in the agent, and the agent consumed partial results everywhere but never computed any of its own.

## Supervise the planner escalation

`escalate` hands `plannerAgent` 15m/$20 — the largest budget in the agent — and ran it with nobody watching. New `supervisedPlan` wraps it with the same `progressCheck` the solves get (verifier evidence, draft deltas, thrash backstop). One documented caveat: the planner's generated program runs in a subprocess, so its drafts stay in the child and the parent's draft delta reads "none" — the fail-open no-signal path, with verifier findings and check history still carrying the evidence.

## Nesting isolation (two fixes)

Supervision now nests — a supervised turn can contain a supervised solve — which surfaced two shared-state collisions:

- **`std::supervise` guards with a per-instance label** (`std::supervise#N`). The handler chain visits every handler, and with the old shared `"std::supervise"` label an outer instance would answer an inner instance's trips: its check would fire against the inner guard's numbers, and because guard grants accumulate in the approval merge, the inner budget would be extended twice per trip. Each handler now gates on its own instance label. `nestedSuperviseIsolated` pins it: the inner instance's checks run, the outer handler answers zero of them.
- **The agent supervisor's working state is a stack of frames** (`pushSupervisorFrame`/`popSupervisorFrame` around each supervised run) instead of flat module state, so an inner run cannot wipe an outer run's history/alternatives/draft mid-flight. `nestedFramesAreIsolated` pins push/push/pop behavior. (`resetSupervisor` is gone; the agent is its only caller.)

## Partial results

Following the partial-results guide, drafts at every point where finished work could still be lost:

- **Brainstorm**: `saveDraft` between the fork and the judge. A budget trip during the judge call now salvages all three finished proposals (as alternatives, none chosen) instead of discarding them for `EMPTY_BRAINSTORM`; the solver proceeds undirected and the supervisor still gets the full list for redirects.
- **Milestone drafts in `escalateToExpert` and `solveUnattended`**: each completed solve (and each successful fix round) is drafted before the verify/fix tail runs, so an outer budget guard tripping mid-verify salvages the finished solve instead of nothing. Limitation stated in the comments: this covers guard-trip aborts, not user cancellation (`AgencyCancelledError` propagates without converting, by design).

The stdlib worker agents already compute partials (coding via prompt, agency/coding per iteration, researcher per grounding round, planner per step, verifier via `finalize`) — these were the agent-side gaps.

## Tests

28 execution tests, no LLM calls: supervise 8 (including `nestedSuperviseIsolated`), agent supervisor 15 (including frame isolation), brainstorm 5 (including the salvaged-draft shape rendering nothing for the solver and everything for the supervisor). One pre-existing flake note: `overshootIsCoveredByTheGrant` timed out once on a loaded machine — the load sensitivity that test itself documents — and passes on rerun.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
@@ -37,7 +37,11 @@ import {
   renderChosen,
   renderAlternatives,
  } from "./brainstorm.agency"
-import { progressCheck, resetSupervisor } from "./supervisor.agency"
+import {
+  progressCheck,
+  pushSupervisorFrame,
+  popSupervisorFrame,
+ } from "./supervisor.agency"
 
 
 static const superpowersSkill = skillsDir("./skills/superpowers", "flat") with approve
@@ -148,13 +152,7 @@ def escalate(task: string, reason: string, plan: string): string {
   @param reason - why this task needs a multi-step plan.
   @param plan - your outline of the steps.
   """
-  const planned = plannerAgent(
-    task,
-    context: reason,
-    plan: plan,
-    maxCost: ESCALATE_MAX_COST,
-    maxTime: ESCALATE_MAX_TIME,
-  )
+  const planned = supervisedPlan(task: task, reason: reason, plan: plan)
   if (isFailure(planned)) {
     return "Planner stopped: ${planned.error}"
   }
@@ -246,7 +244,7 @@ def supervisedSolve(
   agencyTask: boolean,
   alternatives: string = "",
 ): Result<string> {
-  resetSupervisor(alternatives: alternatives)
+  pushSupervisorFrame(alternatives: alternatives)
   const captured = supervise(
     every: SUPERVISE_EVERY,
     maxTime: ESCALATE_MAX_TIME,
@@ -266,6 +264,42 @@ def supervisedSolve(
       summary: solved.value
     }
   }
+  popSupervisorFrame()
+  return flattenSolveOutcome(captured)
+}
+
+/* The planner escalation, supervised the same way as a solve: it hands
+   plannerAgent the largest budget in the agent and previously ran it
+   with nobody watching. Draft deltas fail open here — the planner's
+   generated program runs in a subprocess, so its drafts live in the
+   child and the delta reads "none" in the parent; the verifier and
+   check history still carry the evidence. */
+def supervisedPlan(task: string, reason: string, plan: string): Result<string> {
+  pushSupervisorFrame()
+  const captured = supervise(
+    every: SUPERVISE_EVERY,
+    maxTime: ESCALATE_MAX_TIME,
+    check: progressCheck.partial(task: task),
+  ) {
+    const planned = plannerAgent(
+      task,
+      context: reason,
+      plan: plan,
+      maxCost: ESCALATE_MAX_COST,
+      maxTime: ESCALATE_MAX_TIME,
+    )
+    if (isFailure(planned)) {
+      return {
+        ok: false,
+        summary: "${planned.error}"
+      }
+    }
+    return {
+      ok: true,
+      summary: planned.value
+    }
+  }
+  popSupervisorFrame()
   return flattenSolveOutcome(captured)
 }
 
@@ -322,6 +356,10 @@ def escalateToExpert(task: string, agencyTask: boolean): string {
     summary = "Coding agent stopped: ${solved.error}"
   } else {
     summary = solved.value
+    // Milestone draft: a completed solve is worth keeping even if the
+    // verify or fix round below is cut short by an outer budget guard —
+    // salvage-on-abort then returns this instead of nothing.
+    saveDraft(summary)
   }
   const checked = verifierAgent(
     task: task,
@@ -342,6 +380,7 @@ Fix each problem exactly.
     const fixed = supervisedSolve(task: fixMsg, context: "", agencyTask: agencyTask, alternatives: alternatives)
     if (!isFailure(fixed)) {
       summary = fixed.value
+      saveDraft(summary)
     }
   }
   return summary
@@ -356,6 +395,9 @@ def solveUnattended(task: string, agencyTask: boolean): string {
   if (isFailure(solved)) {
     return "The task did not complete: ${solved.error}"
   }
+  // Milestone draft: the verify and fix rounds below can be cut short by
+  // an outer budget guard; the completed solve must survive that.
+  saveDraft(solved.value)
   const checked = verifierAgent(task: task, maxCost: $5.00, maxTime: 5m)
   if (!feedbackHasErrors(checked)) {
     return solved.value
@@ -373,6 +415,7 @@ Fix each one; your work will be re-checked.
   if (isFailure(fixed)) {
     return solved.value
   }
+  saveDraft(fixed.value)
   return fixed.value
 }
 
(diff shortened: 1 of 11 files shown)


packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
```diff
@@ -266,6 +264,42 @@ def supervisedSolve(
       summary: solved.value
     }
   }
+  popSupervisorFrame()
```
egonSchiele, line 273: **Frame leak on the abort path.** This pop only runs when `supervise` returns a value. When an *outer* budget guard trips mid-solve and the abort unwinds through `supervisedSolve` (the exact salvage-on-abort path this PR builds milestone drafts for), the pop is skipped and the frame leaks. Same for `supervisedPlan` below.

The damage is not isolation — the next run pushes a fresh frame and reads it — but the stack only ever grows across an agent session (the agent process is long-lived, one turn's abort leaks into all later turns), and anything reading `supervisorHistory()` between runs sees a dead run's stale history.

`finalize` runs on the trip/abort path (see `tests/agency/guards/finalize-after-resume.agency`, where a time trip still executes the finalize), so a return-less

```agency
  finalize {
    popSupervisorFrame()
  }
```

in both `supervisedSolve` and `supervisedPlan` would close the hole — assuming a finalize without a `return` preserves the normal outcome and the draft salvage; worth confirming against the finalize semantics before wiring it. If that interaction is hairy, the cheap alternative is documenting the leak and having `pushSupervisorFrame` clamp the stack (e.g. drop frames beyond a small depth).

packages/agency-lang/stdlib/supervise.agency
```diff
@@ -28,6 +28,14 @@ export type SuperviseDecision = {
   message: string
 }
 
+// Each supervise call guards with its own label, so nested supervision
+// works: the chain visits every handler, and with a shared label an
+// inner instance's trip would also be answered by the outer handler —
+// firing the outer check against the inner guard's numbers and, because
+// guard grants accumulate in the approval merge, extending the inner
+// budget twice.
+let _instanceCounter = 0
```
egonSchiele, line 44: **The counter does not survive fork.** Module globals are branch-isolated under `fork` — each branch gets its own copy and writes vanish at the join (the mechanism #613's own fork-isolation test documents). So two branches that each call `supervise()` increment their *isolated* copies from the same base and both mint `std::supervise#N` with the same N — while the handler chain is *shared* across branches. Each branch's handler then passes the other's label gate: wrong-instance answering and double budget extension, the exact bug class this change fixes, resurrected one dimension over.

The agent never calls supervise inside fork today, so nothing breaks now — but this is a public stdlib primitive whose new comment promises instance isolation unconditionally. Either state the limitation in the comment and docstring ("instances in sibling fork branches can collide"), or make the label branch-unique — if the trip payload's snapshot carries the tripped `guardId`, gating the handler on that instead of a minted label would be intrinsically instance-exact and fork-proof, at the cost of the handler needing a way to learn its own guard's id.

packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
```diff
@@ -266,6 +264,42 @@ def supervisedSolve(
...
+  return flattenSolveOutcome(captured)
+}
+
+/* The planner escalation, supervised the same way as a solve: it hands
+   plannerAgent the largest budget in the agent and previously ran it
+   with nobody watching. Draft deltas fail open here — the planner's
+   generated program runs in a subprocess, so its drafts live in the
+   child and the delta reads "none" in the parent; the verifier and
+   check history still carry the evidence. */
+def supervisedPlan(task: string, reason: string, plan: string): Result<string> {
```
egonSchiele, line 296: Nit: `supervisedPlan` is `supervisedSolve` with a different inner call — same push/supervise/normalize/pop/flatten frame, duplicated. A shared wrapper taking the worker as a `() -> Result<string>` block would leave one copy of the outcome-normalization subtleties (the no-match-in-supervised-blocks rule lives in a comment only in `supervisedSolve`; its twin here silently depends on the same fact). Fine to punt if a function-valued parameter fights the checker — the duplication is small — but if a third supervised worker ever appears, extract then at the latest.

egonSchiele (commented):
Reviewed at 490cfc5. The three moves — supervising the planner escalation, instance-labeled supervise guards, and milestone drafts — are all right, and the two nesting fixes address real collisions with discriminating tests (`nestedSuperviseIsolated`'s huge outer interval means any `outerChecks` count can only come from wrong-instance answering; the frame test pins push/push/pop exactly).

**What I verified:**
- The instance-label gate composes correctly with #611's self-exclusion: an inner check's verifier raises skip the inner supervise handler, hit the outer one, fail its label gate, and land at the agent policy — the intended path.
- The milestone `saveDraft` placement is sound: salvage-on-abort is per frame (each aborted function returns *its own* draft), so `escalateToExpert`'s draft is what surfaces when an outer guard trips mid-verify, regardless of deeper drafts inside the verifier. The brainstorm draft between fork and judge is exactly the right seam, and its test pins both destinations.
- The subprocess caveat on `supervisedPlan`'s draft deltas is honest and lands on the fail-open "none" path.
- The regenerated supervise doc page is included this time.

Two real findings inline (a frame leak on the abort path, and a fork-lineage hole in the instance counter), plus one small structural nit. Neither blocks the feature for the agent's current sequential use — but both are exactly the kind of latent state bug this PR exists to fix, so better now than after the next survey.

packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
egonSchiele: Went with the clamp+document path in c6208d1d3 — and your "worth confirming" hedge was prescient: I probed the finalize interaction directly, and a return-less finalize replaces the aborted scope's salvage value with undefined, clobbering the milestone drafts this PR exists to deliver. So finalize is off the table. The leak is now documented at the frame stack (leaked frames sit below the next run's frame and are never read; only growth and stale supervisorHistory between runs) and pushSupervisorFrame clamps at MAX_SUPERVISOR_FRAMES=8. The finalize-clobbers-draft behavior itself looks like a language footgun — filing it as an issue.

packages/agency-lang/stdlib/supervise.agency
egonSchiele: Documented in c6208d1d3, in both the counter comment and the docstring: sibling fork branches increment branch-isolated counter copies and can mint colliding labels against the shared chain — do not call supervise() in parallel branches. Went with documentation over the guardId-gate because the handler has no way to learn its own guard's id today (the guard call is inside the handle block and returns the block result, not the id); if that seam ever exists, gating on the tripped guardId would be the fork-proof fix.

packages/agency-lang/lib/agents/agency-agent/subagents/code.agency
egonSchiele: Extracted in c6208d1d3: supervisedRun(task, alternatives, work: () -> Result<string>) owns the push/supervise/normalize/pop/flatten frame and the no-match comment; solve and plan are now thin lambdas over it. The function-valued param went through the checker cleanly.

packages/agency-lang/tests/agency/supervise/supervise.agency
```diff
@@ -158,3 +158,35 @@ node checkSeesDraft(): string {
...
+
+def innerCheck(elapsed: number, draft: any): SuperviseDecision {
+  innerChecks = innerChecks + 1
+  return { action: "continue", message: "" }
+}
+
+node nestedSuperviseIsolated(): string {
+  // The outer interval is huge, so any outerChecks count comes from the
+  // outer handler wrongly answering the inner instance's trips.
+  const r = supervise(every: 600000, maxTime: 6000000, check: outerCheck) {
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c6208d1d3: the node resets both counters at entry.

Issue #609, opened 2026-07-19
Agent-facing time: elapsedTime, std::date number/string cleanup, and reusing the #575 clock seam


egonSchiele:
## Summary

Give Agency code, and through it agents, a way to read elapsed time. Ship an `elapsedTime` function, and while we are in `std::date`, decide whether its time functions should keep returning strings or move to numbers. This came out of the fake-clock work for #575 and a research pass on whether time awareness helps agents. It does, clearly, so this is worth doing after #575 lands.

Parking it for now. This issue is the full context so we can pick it up later.

## Why this matters

The research on agent time awareness points one direction.

- Without remaining-time updates, GPT-5.1 closed 4% of deals in a negotiation benchmark. With remaining time surfaced each turn, it closed 32%, and accepted offers six times as often. Models cannot track elapsed time on their own, so telling them changes the outcome. https://arxiv.org/html/2601.13206v1
- Agents that notice they are not converging and stop early save 28% to 64% of wasted tokens, and lose only 1.6% to 4.2% of successes. The pattern they catch is thrashing: more tool calls per step with no progress. https://arxiv.org/html/2606.00198
- The strongest result: give the model a `get_duration()` tool and tell it to call it often. Agents then scale their own effort to the budget. On-time completion went from 30% to 53%. The winning design was a tool the agent pulls, not a push notification. https://arxiv.org/pdf/2601.16486

The takeaway for us: the highest-value new primitive is a read the agent can pull. Agency has a hard deadline (`guard(time:)`) and a periodic push nudge (`supervise`). It has no way for code to ask "how long have I been working." That is the gap.

## Proposed function

```
def elapsedTime(since: string) {
  return now() - since
}
```

An agent gets the pull mechanism by partial application:

```
const start = now()
const res = llm(prompt, tools: [elapsedTime.partial(since: start)])
```

## Open question: strings or numbers in std::date

This is the real decision, and the reason to think before shipping.

Every time function in `std::date` returns an ISO 8601 string right now. `now()`, `today()`, `add()`, all strings. See stdlib/date.agency. So `now() - since` in the draft above does not type-check as written; you cannot subtract two strings.

Two ways out.

1. Keep strings, add the math. We already have `add(datetime, ms)`. We would add a `subtract`, and `elapsedTime` would compute a difference and return a number of milliseconds. More surface, but the public type stays a human-readable ISO string.

[...]

PR #618, opened 2026-07-19
Fake clock for deterministic time-guard tests (#575)

### LINKED ISSUE #575: Test framework: injectable clock seam on TimeGuard for deterministic time-guard fixtures

egonSchiele:
## What and why

Time-guard test fixtures burned real CPU in `spin()` busy-loops to run down a guard's clock. That made them slow and machine-dependent: one node, `overshootIsCoveredByTheGrant`, spent ~70s (a quarter of the Agency suite's recent CI growth), and the tuning could silently stop catching its bug on faster hardware. Closes #575.

This adds a fake clock a fixture opts into, then advances with a function call instead of spinning. The guard trips because its clock moved, not because real time passed.

## The seam

- **`lib/runtime/clock.ts`** (new) — a `Clock` type (`now`, `wallTime`, `setTimer`, `clearTimer`), the `realClock` default delegating to the globals, and `FakeClock` with `advance(ms)`.
- **`RuntimeContext`** carries a `clock`, set from a test-only constructor arg, else `AGENCY_FAKE_CLOCK`, else `realClock`. Not serialized.
- **`TimeGuard`** routes its six `performance.now()` reads and its `setTimeout`/`clearTimeout` through a `clock()` helper that reads the run's clock via the canonical `__ctx()` accessor, falling back to `realClock` when frameless (a checkpoint-revived guard).
- **`_advanceTime`** — a test-only seam in `std::date`, non-exported, reached via `import test`, following the existing `_installSlowInput` precedent. It throws if the clock is not fake, so a stray call fails loudly.
- **`fakeClock: true`** on a `.test.json` test case sets `AGENCY_FAKE_CLOCK=1` for that subprocess.

The `Clock` type carries `wallTime()` (unused here) so issue #609's `elapsedTime`/`now` reads can reuse the seam without widening it.

## A real bug found during execution

`createExecutionContext` builds the *execution* context via `Object.create(RuntimeContext.prototype)`, bypassing the constructor — so it never copied `clock`, and the run metered against `undefined`. The execution context now inherits the global context's clock, with a regression test. This was not anticipated by the plan; it surfaced because the e2e fixture failed and I traced it.

## Results

`overshootIsCoveredByTheGrant`: ~70s to ~0.7s. The whole `supervise.agency` file: ~76s to ~13s. Critically, the migrated node **still catches the `grant = nextInterval` regression** — verified against a deliberately reverted build, it fails with the guard-approval error in 0.78s instead of 70s.

Only the two `spin(3000000)` nodes in `supervise.agency` were migrated (the CI cost). `fakeClock` is per test case, so the file's cheap `spin(300000)` nodes keep the real clock untouched. The other spin fixtures (`nestedGuardResume`, `trip-time`, `trip-join`) are deliberately left on the real clock — each needs its own interleaving trace and none is a CI-time problem. `trip-join` in particular drives guards through `runBatch` fork/race branches whose working-time reads real time, which this seam does not route; the audit (committed as `docs/.../fake-clock-migration-list.md`) records the reasoning.

## Verification

- New unit tests: `clock.test.ts` (10, incl. re-entrancy and boundary), `context.clock.test.ts` (incl. the `createExecutionContext` regression), `guard.clock.test.ts` (asserts `spent`, not just non-null, so an unrouted read fails it; plus a nested two-guard case), `builtins.clock.test.ts`.
- Existing `guard.test.ts` (47 tests) passes unchanged — it runs frameless, proving the `realClock` fallback.
- Broad sweep: 3122 unit tests across `lib/runtime`, `lib/cli`, `lib/stdlib` pass. `tsc --noEmit` and `lint:structure` clean.
- The e2e fixture proves the seam three ways: deterministic trip, no-trip-under-budget, and the opt-out (no `fakeClock` makes `_advanceTime` refuse with the specific message).

Spec, plan, and both review docs are included under `docs/superpowers/`.

### DIFF
--- a/packages/agency-lang/lib/runtime/state/context.ts
+++ b/packages/agency-lang/lib/runtime/state/context.ts
@@ -77,7 +77,9 @@ function reviveNative<T>(data: T): T {
  *  AGENCY_FAKE_CLOCK env var, otherwise the real clock. The env var is set by
  *  the test runner per test case (see lib/cli/util.ts). */
 function defaultClock(): Clock {
-  return process.env.AGENCY_FAKE_CLOCK ? new FakeClock() : realClock;
+  // Require exactly "1", not any truthy string. AGENCY_FAKE_CLOCK=0 must
+  // disable, not enable — a non-empty "0" is truthy and would surprise.
+  return process.env.AGENCY_FAKE_CLOCK === "1" ? new FakeClock() : realClock;
 }
 
 /* bunch of stuff that every node/function in the runtime needs access to,
@@ -92,7 +94,15 @@ export class RuntimeContext<T> {
   onStreamLock: boolean;
   handlers: HandlerEntry[];
   /** The time source for guards. Real by default; a FakeClock only when a
-   *  test opts in. NOT serialized — reconstructed per run, like handlers. */
+   *  test opts in. NOT serialized — reconstructed per run, like handlers.
+   *
+   *  DRIFT WARNING: this and the other non-serialized runtime fields
+   *  (handlers, callbacks, checkpoints, locks, …) are set by the constructor
+   *  AND, separately, by `createExecutionContext`, which builds the execution
+   *  context via `Object.create` and copies fields by hand. A field added
+   *  here but forgotten there is silently `undefined` at run time with no
+   *  compile error — that is the bug the `clock` copy fixed. If you add a
+   *  non-serialized field, copy it in `createExecutionContext` too. */
   clock: Clock;
   locks: Record<string, Promise<void>>;
   lockOwners: Record<string, string>;
--- a/packages/agency-lang/lib/runtime/state/context.ts
+++ b/packages/agency-lang/lib/runtime/state/context.ts
@@ -8,6 +8,7 @@ import { nativeTypeReplacer, nativeTypeReviver } from "../revivers/index.js";
 import { CoverageCollector } from "../coverageCollector.js";
 import { AgencyCancelledError, makeAbortCause } from "../errors.js";
 import type { AbortCause } from "../errors.js";
+import { Clock, realClock, FakeClock } from "../clock.js";
 import { agencyStore } from "../asyncContext.js";
 import { DEFAULT_MAX_CALL_DEPTH } from "../callDepth.js";
 import { getSubprocessRunInfo } from "../subprocessRunInfo.js";
@@ -72,6 +73,13 @@ function reviveNative<T>(data: T): T {
   );
 }
 
+/** The default clock for a run: a FakeClock only when a test opts in via the
+ *  AGENCY_FAKE_CLOCK env var, otherwise the real clock. The env var is set by
+ *  the test runner per test case (see lib/cli/util.ts). */
+function defaultClock(): Clock {
+  return process.env.AGENCY_FAKE_CLOCK ? new FakeClock() : realClock;
+}
+
 /* bunch of stuff that every node/function in the runtime needs access to,
 that we don't want to pass as individual arguments everywhere */
 export class RuntimeContext<T> {
@@ -83,6 +91,9 @@ export class RuntimeContext<T> {
   callbacks: AgencyCallbacks;
   onStreamLock: boolean;
   handlers: HandlerEntry[];
+  /** The time source for guards. Real by default; a FakeClock only when a
+   *  test opts in. NOT serialized — reconstructed per run, like handlers. */
+  clock: Clock;
   locks: Record<string, Promise<void>>;
   lockOwners: Record<string, string>;
   lockWaiters: Record<string, string[]>;
@@ -221,6 +232,9 @@ export class RuntimeContext<T> {
      *  test/runtime constructors keep working; defaults to "info" to
      *  match the established no-debug-by-default behavior. */
     logLevel?: LogLevel;
+    /** Test-only override. Omitted in production, where defaultClock()
+     *  (the env var or the realClock default) applies. */
+    clock?: Clock;
   }) {
     // One runtime merge, applied for BOTH transports so a subprocess launched
     // with explicit IPC overrides still inherits the env override (e.g. a
@@ -232,6 +246,7 @@ export class RuntimeContext<T> {
       args,
       getRuntimeConfigOverrides(),
     );
+    this.clock = args.clock ?? defaultClock();
     const statelogConfig = {
       ...args.statelogConfig,
       traceId: args.statelogConfig.traceId || nanoid(),
--- a/packages/agency-lang/lib/runtime/state/context.ts
+++ b/packages/agency-lang/lib/runtime/state/context.ts
@@ -351,6 +351,11 @@ export class RuntimeContext<T> {
     execCtx.maxCallDepth = this.maxCallDepth;
     execCtx.failurePropagation = this.failurePropagation;
     execCtx.checkpoints = new CheckpointStore(this.maxRestores);
+    // The execution context is built via Object.create, bypassing the
+    // constructor, so carry the clock over from the global context. Without
+    // this the run would meter against `undefined` and the fake-clock seam
+    // (_advanceTime) would never see the FakeClock the constructor installed.
+    execCtx.clock = this.clock;
     execCtx.handlers = [];
     execCtx.callbacks = {};
     execCtx.topLevelCallbacks = [];
--- a/packages/agency-lang/tests/agency/guards/fake-clock-trip.agency
+++ b/packages/agency-lang/tests/agency/guards/fake-clock-trip.agency
@@ -30,8 +30,12 @@ node doesNotTripUnderBudget(): string {
 }
 
 // The opt-out: a test case WITHOUT "fakeClock" keeps the real clock, and
-// _advanceTime refuses loudly instead of silently doing nothing. Capture its
-// own result — the refusal is a Failure carrying the "fake clock" message.
+// _advanceTime refuses loudly instead of silently doing nothing. There is no
+// guard/handler here (unlike the trip node above): _advanceTimeImpl throws a
+// plain Error, and this relies on failure propagation converting that non-abort
+// throw into a Failure whose `.error` is the message. That conversion is the
+// contract under test — if it ever changed shape, this safety check would need
+// to change with it.
 node advanceWithoutFakeClockErrors(): string {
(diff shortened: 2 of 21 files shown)


packages/agency-lang/lib/runtime/state/context.ts
```diff
@@ -336,6 +351,11 @@ export class RuntimeContext<T> {
     execCtx.maxCallDepth = this.maxCallDepth;
     execCtx.failurePropagation = this.failurePropagation;
     execCtx.checkpoints = new CheckpointStore(this.maxRestores);
+    // The execution context is built via Object.create, bypassing the
+    // constructor, so carry the clock over from the global context. Without
+    // this the run would meter against `undefined` and the fake-clock seam
+    // (_advanceTime) would never see the FakeClock the constructor installed.
+    execCtx.clock = this.clock;
```
egonSchiele, line 368: Good fix, and the regression test pins it. One drift-prevention thought: this is exactly the failure mode that bit here — a field added to the constructor but forgotten in `createExecutionContext`'s manual copy list, which `Object.create` bypasses. The next non-serialized field someone adds (`clock` just joined `handlers`, `callbacks`, `checkpoints`…) can reintroduce the same class of bug with no compile error. Worth a one-line pointer comment at the constructor's field block ("reconstructed fields must also be copied in createExecutionContext") or a small shared copy helper, so the two lists can't silently drift. Not blocking.

packages/agency-lang/tests/agency/guards/fake-clock-trip.agency
```diff
@@ -0,0 +1,39 @@
...
+  if (isFailure(r)) { return "tripped" }
+  return r.value
+}
+
+// The opt-out: a test case WITHOUT "fakeClock" keeps the real clock, and
+// _advanceTime refuses loudly instead of silently doing nothing. Capture its
+// own result — the refusal is a Failure carrying the "fake clock" message.
+node advanceWithoutFakeClockErrors(): string {
+  const res = _advanceTime(10)
+  if (isFailure(res)) { return res.error }
```
egonSchiele, line 41: This opt-out case is the "fails loudly" safety guarantee, and its correctness rests on an implicit conversion: `_advanceTimeImpl` throws a plain `Error`, and this line relies on that throw becoming an Agency `Failure` with `.error` set to the message (there's no guard/handler here, unlike the trip node above). It clearly works — the exact-message match passes — but it's coupling this test to failure-propagation converting a non-abort throw into a `Failure`. Worth a one-line comment noting that's the mechanism (or confirming it's the intended contract), since if that conversion ever changed shape, this specific safety test would change with it silently rather than failing obviously.

egonSchiele (commented):
Reviewed the code against the runtime, not just on its own terms. This is a clean, well-scoped PR that folded in essentially every finding from the spec/plan reviews — `guard.clock.test.ts` asserts `spent` (so an unrouted `now()` read fails it, not just a null check), the re-entrancy and `dueAt === target` boundary cases are covered, the opt-out asserts the exact refusal message, and the `.filter().sort()[0]` became a clean `earliestDueBy` reduce.

Two things I independently verified because they are safety-critical, both clean:

- **The frameless fallback is correct.** `__ctx()` (`asyncContext.ts:234`) is the lax accessor — `agencyStore.getStore()?.ctx`, returning `undefined` outside a frame — so `__ctx()?.clock ?? realClock` genuinely resolves to `realClock` for a checkpoint-revived guard rather than throwing. That's also why the 47 frameless `guard.test.ts` tests still pass. Good call using `__ctx` over `agency.ctxMaybe()` to avoid the agency.ts→guard.ts import cycle; the inline comment documents it.
- **The `createExecutionContext` fix is complete.** I grepped for other constructor bypasses: `Object.create(RuntimeContext.prototype)` appears only in `createExecutionContext`, and every other exec-context caller (`interrupts.ts`, `node.ts`, `rewind.ts`) routes through it, so they all inherit the `execCtx.clock = this.clock` copy. No sibling path silently meters against `undefined`. Nice find surfacing this from the failing e2e — it was not in the plan.

Also confirmed the e2e trip fixture's `handle { … } with (i) { "std::guard" => reject() }` matches the established trips-as-interrupts pattern in `trip-time.agency` (a guard trip raises an interrupt that must be answered), so the handler is necessary and idiomatic, and `doesNotTripUnderBudget` correctly needs none.

Only minor notes inline — nothing blocking. One more for the summary: `checkSeesDraft` was migrated to the fake clock with its `draftsSeen[0]` assertion unchanged; worth a glance to confirm that assertion is still genuinely exercised (a draft actually gets seen) under the deterministic interleaving rather than passing trivially.

egonSchiele:
## Why `Object.create(RuntimeContext.prototype)`?

There is a real reason, and it is structural, not stylistic.

`createExecutionContext` is not constructing a fresh context — it is **deriving** an execution context from a template context (`this`, the "global" one built with `new RuntimeContext(...)`). It copies the already-resolved config fields (`graph`, `smoltalkDefaults`, `dirname`, `statelogConfig`, `maxRestores`, `failurePropagation`, …) straight off `this`, and re-mints only the per-run state (a fresh `StateStack`, `GlobalStore`, `CheckpointStore`, `AbortController`, `PendingPromiseStore`, a `StatelogClient` keyed by `runId` as trace id, etc.).

Two things make `new RuntimeContext(...)` the wrong tool for that:

1. **The initializer is async, and constructors can't be.** It does `await TraceWriter.create({runId, traceConfig})` and awaits subprocess-run info. A constructor cannot await, so the per-run setup can't live in one. `Object.create` + an async method is the way to get a fully-formed instance whose init needed I/O.
2. **It must not re-run the constructor's work.** The constructor merges env/IPC config overrides (`applyRuntimeConfigOverridesToContextArgs`, twice) and mints a fresh `nanoid` trace id. That resolution already happened when the template `this` was built. The execution context wants the *resolved* config copied over and a trace id of `runId`, not a re-resolution and a new random id.

So `Object.create` is deliberate: async init + derive-from-template. It is not an accident or a shortcut.

**The cost — which is exactly the bug this PR hit.** Because it copies fields by hand, the constructor's field list and this copy list have to be kept in sync manually, with no compiler check. Add a non-serialized field to the constructor and forget it here, and the execution context gets `undefined` for it silently. `clock` was the latest casualty; `handlers`, `callbacks`, `checkpoints` are all in the same manual list. That is a genuine maintenance hazard in the existing design, independent of this PR.

I did not restructure it — that is a bigger change than a test-infra PR should carry — but I added a **drift warning** at the `clock` field (5ee7950) pointing at `createExecutionContext`, per your inline note, so the next person adding a non-serialized field sees the trap. If you'd like, a good follow-up would be a single `copyDerivedFields(from, to)` helper shared by the constructor and `createExecutionContext` so the two lists can't diverge — happy to file it.

## The rest of the review (5ee7950)

- **Drift comment** (your note on context.ts:358): added at the field block, as above.
- **Opt-out test relying on throw→Failure** (your note on the fixture): added a comment stating that `_advanceTimeImpl` throws a plain `Error` and the test relies on failure propagation converting it to a `Failure` with `.error` — that conversion is the contract under test.
- **`checkSeesDraft` genuinely exercised** (your summary note): confirmed. Its assertion is `firstDraft:partial`, and it passes with `partial` — if the trip never fired, `draftsSeen[0]` would be `none`. So the draft is really seen under the deterministic interleaving, not passing trivially.
- **`TimerHandle` cast** (Copilot): removed the `as unknown as number` lie; the handle id is now typed `ReturnType<typeof setTimeout> | number` honestly. Typechecks clean.
- **`advance()` termination / bad input** (Copilot): the doc-comment now says termination holds for *this codebase's usage* (guard callbacks never re-arm, all delays ≥ 0), not for arbitrary self-rearming callbacks. Added a guard that rejects a negative or NaN advance rather than moving the clock backward or corrupting it, with a test. I did not add a firing cap — a self-rearming-0ms loop is a hypothetical no code here creates, and the plan reasoned that out deliberately.
- **`AGENCY_FAKE_CLOCK` truthiness** (Copilot): now requires exactly `"1"`, so `AGENCY_FAKE_CLOCK=0` disables instead of surprisingly enabling.

Typecheck, lint, and both e2e fixtures (3/3 and 8/8) still green.

Issue #453, opened 2026-07-08
std::ui/layout: render() "auto" color mode strips ANSI in interactive terminal (compiled program stdout not detected as TTY)


egonSchiele:
## Summary

`render(node)` from `std::ui/layout` with the default `color: "auto"` strips ANSI color even when run in an interactive terminal. Any pre-colored content (e.g. a `std::syntax::diff(..., color: true)` result, colored logs, highlighted code) placed in a layout and rendered with the default color mode comes out **monochrome**, while `print()`-ing the same string keeps its color.

Passing `color: true` explicitly works around it.

## Repro

```
import { diff } from "std::syntax"
import { box, render } from "std::ui/layout"
const file1 = read("a.txt") with approve
const file2 = read("b.txt") with approve

node main() {
  const res = diff(file1 catch "", file2 catch "", color: true)
  print(res)                          // colored — correct
  const panel = box(title: "Hi", width: "full") as b { b.text(res) }
  print(render(panel))                // NO color  <-- bug
  print(render(panel, color: true))   // colored   <-- workaround
}
```

Run with `pnpm run agency <file>` in an interactive terminal.

## Root cause (partially traced)

`render`'s default `color: "auto"` resolves via `_render` (`lib/stdlib/layout/render.ts:127`) → `autoUseColor()` (`lib/utils/termcolors.ts:207`), which returns `process.stdout.isTTY === true`.

Verified under a real PTY (via `pty.fork`, window size set to 80x24):

- A bare `node` script on that PTY reports `process.stdout.isTTY === true` and `autoUseColor() === true`.
- The **compiled agency program** on the same PTY still strips color on the `"auto"` path (`render(panel, color: true)` keeps color; `render(panel)` does not). Reproduced both via `pnpm run agency` and by running the compiled `.js` directly with `node`.

So something about how the `agency` runner launches the compiled program leaves its stdout not looking like a color-capable TTY to `autoUseColor()` at render time — even though the CLI runner spawns it with `stdio: "inherit"` (`lib/cli/commands.ts:512-520`). The exact reason the child's `isTTY`/color detection reads false was not fully pinned down.

## Impact

The default color mode is meant to "emit colors only to a terminal" — but in practice it emits **no** colors for programs run through the `agency` CLI, even interactively. Users must pass `color: true` to get any color from `render`, which is surprising given `print()` of the same string is colored. This makes `std::ui/layout` + `std::syntax` (and any pre-colored content) look broken by default.

## Suggested direction

[...]

PR #619, opened 2026-07-20
Anchor all relative paths to cwd; add __dirname builtin

egonSchiele:
## Summary

Every path-taking stdlib function now resolves relative paths against the working directory (where the program was run), and the new `__dirname` builtin gives each `.agency` file its own directory:

```agency
read("notes.md")                       // ./notes.md in the working directory
read("../prompts/main.md", __dirname)  // anchored to the current .agency file
```

This replaces the old three-way split (entry-module-anchored `read`/`write`/`edit`/`ls`/`grep`/`glob`, cwd-anchored `exec`/`mkdir`/`copy`, and `stat`/`exists` switching anchors per call) with one rule that matches how file I/O works in every other language.

Spec: `packages/agency-lang/docs/superpowers/specs/2026-07-19-cwd-path-anchoring-design.md` (and its `-REVIEW` sibling). Motivating incident: the agent ran with a silently empty system prompt because a relative `read` in an imported module resolved against the entry module's directory (`2026-07-19-relative-path-resolution-entry-module-bug.md`).

## What changed

- **`__dirname` is now a typed builtin.** Every compiled module already emitted the TS constant; this registers the name in `BUILTIN_VARIABLES` and types it `string` in the checker (after scope lookup, so user bindings shadow it, per the #453 rule). No runtime machinery: it works in `static const` initializers, `fork` branches, handlers, and after restore.
- **`resolveDir` always anchors to `process.cwd()`** — the `base` parameter and its 15 `"cwd"` call sites are gone. New sync core `resolveCwdPath` is shared with `_readSkill`.
- **`resolvePath` no longer enforces containment.** The lexical `..` guard, the symlink-escape re-check, and the absolute-filename rejection are removed; upward traversal and absolute filenames now work like `open()`.
- **`stat`/`exists` stop switching anchors** — they now always agree with `read` on which file a path means.
- **`std::system::dirname()` is deleted.** It could only ever report the entry module's directory; `__dirname` replaces it. Importing it now errors (AG4008).
- **The agent's #597 orphan crash is fixed on this branch too:** `checkMainPrompts()` hard-fails at startup instead of the agent silently running with an empty system prompt, and the prompt reads use the natural `read("../prompts/main-large.md", __dirname)` spelling.

## Breaking change + security note

Programs relying on entry-module anchoring change meaning; the migration recipe is "if a relative path meant next-to-my-file, anchor it with `__dirname`" (CHANGELOG entry included). Security: `ls`/`grep`/`glob`/`stat`/`exists` and the fs-mutations keep allow-list containment (still fails closed; pinned by a new test). `read`/`write`/`edit` never had an allow-list — they are gated by their unconditional `std::read`/`std::write`/`std::edit` interrupts, which are unchanged. This PR removes their internal path containment, so an approved read or write can now reach any path the interrupt policy allows.

## Migration audit decisions

- `agency-agent/lib/coordinator.agency` — prompt reads anchored with `__dirname`; `AGENTS.md` reads take an explicit caller `dir` (user-project semantics, unchanged).
- `agency-agent/subagents/code.agency` — `skillsDir` anchored with `__dirname` (colocated skill bundle).
- `review/agent.agency`, `policy/agent.agency` — system-prompt reads anchored with `__dirname`. `review`'s `read(filePath)` on a runtime-supplied path keeps its unconditional `std::read` interrupt; who answers it is decided by the policy the agent runs under (pre-existing behavior, unchanged).
- Tests: fixture-relative reads in `tests/agency/stdlib/*`, `tests/agency/builtins/readFile`, `tests/agency/handlers/*`, and `tests/agency-js/handler-fork-isolation` anchored with `__dirname`. `agent-cwd.agency`'s `unsetReadsModuleDir` node is now `unsetReadsCwd` and pins cwd anchoring. `read-relative-to-module` and `dirname-basic` repin the new semantics. `/tmp`-absolute and expected-failure tests unchanged.
- New `tests/agency/dirname-paths.agency`: imported-module colocated read (the original bug's exact shape), upward traversal, absolute filename, probe/read agreement, allow-list fail-closed.

A follow-up PR deletes the now-dead `moduleDir` runtime plumbing (ALS field, `getModuleDir`, threading through node/runner/interrupts/rewind, template seeds) — kept separate so this PR stays reviewable and bisectable.

## Testing

- Unit: `lib/typeChecker`, `lib/stdlib`, `lib/backends` — 3299 passed.
- Agency execution tests (no LLM): `dirname-paths` (6/6), `agent-cwd`, `dirname-basic`, `read-relative-to-module`, `readFile`, `glob-basic`, `grep-basic`, `ls-basic`, `ls-recursive`, `stat-basic`, `exists-basic`, `handler-raises-depth`, agency-js `handler-fork-isolation` — all pass.
- `pnpm run lint:structure` clean; agent compiles.


packages/agency-lang/tests/agency/dirname-paths.agency
```diff
@@ -0,0 +1,67 @@
...
+// Upward traversal through __dirname: the old resolvePath guards
+// rejected this even with the right anchor.
+node upwardRead(): string {
+  return readUpward()
+}
+
+// A write anchored to this file via an absolute dir argument.
+node dirAnchoredWrite(): boolean {
+  const outDir = __dirname + "/dirname-paths"
+  const w = write("tmp-out.txt", "temp", outDir) with approve
```
egonSchiele, line 20: `w` is assigned but never read, so the write's own success is discarded — the node only fails if the *read-back* mismatches, which masks a write that failed outright. Either assert on it (`if (w is failure(...)) { return false }`) or drop the binding to `write("tmp-out.txt", "temp", outDir) with approve`. Same on the `absoluteFilenameAllowed` node below.

packages/agency-lang/tests/agency/dirname-paths.agency
```diff
@@ -0,0 +1,67 @@
...
+    return content == "temp"
+  }
+  return false
+}
+
+// Absolute FILENAME (not just dir) now allowed: resolvePath previously
+// hard-threw on any absolute filename.
+node absoluteFilenameAllowed(): boolean {
+  const outPath = __dirname + "/dirname-paths/tmp-abs.txt"
+  const w = write(outPath, "abs") with approve
```
egonSchiele, line 36: Same unused-`w` nit as `dirAnchoredWrite`. (Nice that this case is its own node named for what it proves — absolute filenames winning over `dir` — rather than folded into the write test.)

egonSchiele (commented):
Reviewed against the design spec, its review, and the implementation plan. This is a clean, faithful implementation and it folded in the earlier plan-review feedback: the shared sync `resolveCwdPath` core so `resolveDir` and `_readSkill` no longer copy the expand-then-cwd policy; the `statelog.ts:96` `"cwd"` site the plan's list had missed; the honest security note that `read`/`write`/`edit` have no allow-list and are gated only by their `std::read`/`std::write`/`std::edit` interrupts; and splitting the absolute-filename case into its own `absoluteFilenameAllowed` test node instead of smuggling it under a `__dirname` write test.

Spot checks that held up:
- **`resolvePath`** removes all three guards (lexical `..`, symlink-escape re-check, absolute-filename rejection) and applies `expandPath` to the filename — this is what the spec review flagged as required, and `dirname-paths.agency`'s `upwardRead` node actually exercises the upward-traversal path that the two-line version would still have thrown on.
- **Checker**: `BUILTIN_VARIABLE_TYPES` is consulted in `synthType` *after* `scope.lookup`, so a user `const __dirname` shadows the builtin — verified by the third test. Placement matches the generator's #453 rule.
- **`resolveProbePath`** now anchors both branches to `process.cwd()`, so `exists(p)` and `read(p)` agree; allow-list containment is still layered on for `stat`/`exists`.
- **Docstrings** for `read`/`write`/`writeBinary`/`readBinary`/`edit`/`ls`/`grep`/`glob`/`commandsDir` all now say cwd-relative and point at `__dirname` — important since these become LLM tool descriptions.
- **`checkMainPrompts()`** moved into `coordinator.agency` (same scope as the prompt consts) and is called from `agent.agency`'s `main` before the first turn, closing the #597 orphan-crash / silent-empty-prompt hole.

No blocking issues. Two small test nits inline, plus one optional robustness note: `dirAnchoredWrite` and `absoluteFilenameAllowed` write a temp file, read it back, then `remove` it — but if the read fails or the node returns early, the `remove` never runs and the temp file is left in `tests/agency/dirname-paths/`. It is an isolated dir so reruns just overwrite, so this is low priority; a note only if you want the tests self-cleaning under failure.

packages/agency-lang/tests/agency/dirname-paths.agency
egonSchiele: Fixed in 83bf90f5d — both nodes now return false on a failed write before checking the read-back.

packages/agency-lang/tests/agency/dirname-paths.agency
egonSchiele: Fixed in 83bf90f5d, same guard as dirAnchoredWrite.

packages/agency-lang/lib/typeChecker/builtins.ts
```diff
@@ -13,6 +13,14 @@ import {
 const anyArray = { type: "arrayType", elementType: ANY_T } as const;
 const stringArray = { type: "arrayType", elementType: string } as const;
 
+/** Types for `BUILTIN_VARIABLES` (lib/config.ts) that the checker knows.
+ *  `__dirname` is the per-module directory constant every compiled module
+ *  defines (imports.mustache:38). `color` is deliberately absent — it has
+ *  never been typed and typing it is unrelated scope. */
+export const BUILTIN_VARIABLE_TYPES: Record<string, VariableType> = {
+  __dirname: string,
+};
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 83bf90f5d with an own-property guard at the lookup site (hasOwn helper, matching resolveVariable.ts style) rather than a null-prototype object, per the repo convention of plain object literals. Writing the regression test exposed the same prototype-chain bug in the adjacent ctx.functionDefs / ctx.nodeDefs / ctx.importedFunctions lookups (an undeclared identifier named toString crashed assignability with a params-less functionRefType) — those are guarded now too, covered by the new "does not resolve prototype-chain names" test.

packages/agency-lang/lib/typeChecker/builtinVariables.test.ts
```diff
@@ -0,0 +1,44 @@
...
+import { parseAgency } from "../parser.js";
+import { buildCompilationUnit } from "../compilationUnit.js";
+import { typeCheck } from "./index.js";
+
+function check(source: string): string[] {
+  const parsed = parseAgency(source);
+  if (!parsed.success) throw new Error(`parse failed: ${parsed.message}`);
+  const info = buildCompilationUnit(parsed.result, undefined, undefined, source);
+  return typeCheck(parsed.result, {}, info).errors.map((e) => e.message);
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 83bf90f5d — the test now uses the shared typecheckSource helper (extended with an optional AgencyConfig param), so it runs the full parse → SymbolTable.build → buildCompilationUnit → typeCheck pipeline.

packages/agency-lang/lib/typeChecker/synthesizer.ts
```diff
@@ -281,6 +281,13 @@ export function synthType(
         }
         return scopeType;
       }
+      // Builtin variables resolve only after scope lookup fails, so a
+      // real binding named `__dirname` shadows the builtin (the same
+      // rule the generator applies — see issue #453).
+      const builtinVarType = BUILTIN_VARIABLE_TYPES[expr.value];
+      if (builtinVarType) {
+        return builtinVarType;
+      }
```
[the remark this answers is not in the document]
egonSchiele: Confirmed and fixed in 83bf90f5d: resolveVariable now treats BUILTIN_VARIABLES names as kind "builtin" (step 5 in the resolution order), so undefinedVariables: "warn"/"error" no longer flags __dirname — or color, which had the same pre-existing gap. Pinned by a new test running the diagnostic at "error".

packages/agency-lang/docs/site/stdlib/fs.md
```diff
@@ -109,7 +109,7 @@ Edit a single file by applying one or more text replacements atomically. Each ed
 
   @param filename - The file to edit
   @param edits - Text replacements to apply in order; each has oldText, newText, and replaceAll (replace every occurrence when true)
-  @param dir - The directory to resolve the filename against
+  @param dir - The directory to resolve the filename against. Relative paths resolve against the current working directory; pass __dirname for a path relative to the current Agency file
```
egonSchiele: The docstrings are written for LLMs. THey are tool descriptions. The LLM has no ability to "pass __dirname for a path" so this information is useless, and even worse, it wastes tokens. Undo this change across all files you touched.

egonSchiele: Reverted in the latest commit: every @param docstring line is back to its terse original (read/write/writeBinary/readBinary/edit/ls/grep/glob/commandsDir), and the generated stdlib pages regenerated. One deliberate deviation from a pure undo: the old ls/grep/glob parentheticals said "(relative paths resolve against the module directory)", which is now false — those are dropped entirely rather than restored. The __dirname guidance survives only in human-facing places: the /** */ doc comment above commandsDir, the guide builtins.md page, and the CHANGELOG.

PR #620, opened 2026-07-20
std::date: instants as numbers (#609 PR 1)

egonSchiele:
## What and why

**Breaking change to `std::date`.** This is PR 1 of the `elapsedTime` work (#609). `elapsedTime` needs `now() - since`, but every `std::date` function returned an ISO string, and you can't subtract strings. So this first PR changes how the library represents time. PR 2 adds `elapsedTime` on top.

The insight: `std::date` was mixing two concepts. An **instant** ("now", "start of today") is a moment on the timeline and wants to be a number. A **calendar date** ("today", "next Monday") is a whole day whose identity depends on the timezone and honestly stays a string.

## The change

- **Instants become epoch-millisecond numbers:** `now()`, `atTime`, and `startOf*`/`endOf*` now return a `number`.
- **Calendar dates stay strings:** `today`, `tomorrow`, `nextDayOfWeek` are unchanged (`"YYYY-MM-DD"`).
- **The `add*` family is removed.** With numbers and duration literals, arithmetic is just operators: `add(t, ms)` → `t + ms`, `addHours(t, 2)` → `t + 2h`, `addDays(t, 3)` → `t + 3d`. Behavior-preserving (the old helpers did fixed-ms math).
- **Three new bridges:** `format(ms, tz)` (instant → ISO string with milliseconds), `formatDate(ms, tz)` (instant → `"YYYY-MM-DD"`), `parse(iso)` (string → instant; throws on bad input, never a silent `NaN`).
- **`now()` drops its timezone param** — an instant is absolute; move the timezone to `format(now(), tz)`.
- **`endOf*` returns the last millisecond** (`23:59:59.999`), so `startOf(x) <= x <= endOf(x)` holds now that comparisons are exact.

## Two things worth a look

**A latent bug this migration surfaced and fixed.** The old `_atTime` built its return *string* from the input date/time, which masked that its computed *instant* was wrong whenever the timezone offset crossed midnight (e.g. `00:00:00`) — off by a full day. Exposed the moment instants became numbers. `_atTime` now computes the instant as `wall-as-UTC − offset` with a DST re-check, and is tested against real spring-forward and fall-back days.

**Small internal cleanup.** The six boundary helpers now share one `boundary` combinator with named date transforms (`sundayOf`/`saturdayOf`/`firstOfMonth`/`lastOfMonth`) instead of re-spelling the same four-step pipeline six times. Weekday math uses `getUTCDay()` (machine-independent) rather than `getDay()` (which reads the runner's local timezone and is off by one on far-east zones).

## Blast radius

Small, because the most-used function internally is `today()`, which stays a string. The agency-agent (3 files) and the memory test use only `today()` — unchanged. `calendar.agency` references date helpers only in a doc comment. Only the integration sandbox fixture exercised the changing functions; it's rewritten. External users of `std::date` instants are affected — the changelog carries the migration guide.

## Verification

- The canonical date unit suite (`lib/stdlib/__tests__/date.test.ts`) rewritten for the numbers API, 22 tests: millisecond round-trip across LA/+05:30/UTC, `parse` throws, the NY/Tokyo day-shift, DST spring-forward and fall-back, the `endOf*` ordering invariant (red before the `.999` fix), cross-month weeks, leap-February, and the midnight-crossing `atTime` case.
- `lib/stdlib` vitest sweep: 1224 pass, 0 fail. `tsc --noEmit` clean. `lint:structure` clean. The rewritten sandbox fixture passes.

Spec, plan, and both review docs are under `docs/superpowers/`.


packages/agency-lang/CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 ## Unreleased
 
+### Breaking: std::date represents instants as numbers
```
egonSchiele: undo all changes to this file. Never update the changelog unless asked to.

PR #623, opened 2026-07-20
elapsedTime and formatDuration (#609 PR 2)

egonSchiele:
## What and why

PR 2 of #609. PR 1 (#620) made instants numbers so `now() - since` was possible; this adds the thing #609 actually asked for — a way for an agent to check "how long have I been working."

The research is one-sided: giving a model a sense of elapsed time makes it stop earlier when it isn't converging and scale effort to the budget, and the most-validated mechanism is a tool the agent pulls. So the shape is `elapsedTime.partial(since: start)` handed to an `llm()` call.

## The API

- **`elapsedTime(since: number): string`** — `formatDuration(now() - since)`. Returns a readable duration like `"5m 32s"`, because it is handed to a model as a tool and a model reads `"5m 32s"`, not `332000`. For math, `now() - since` is still the raw milliseconds.
- **`formatDuration(ms: number): string`** — the general renderer underneath it: `"5m 32s"`, `"1h 1s"`, `"2d 3h"`. Whole-second granularity (sub-second is `"0s"`), largest unit is days, negatives get a leading `-` with no negative zero.
- **`now()` now reads through the fake-clock seam.** `_now` and the other current-time reads go through the runtime clock's `wallTime()`, so a fixture with `fakeClock: true` can drive them with `_advanceTime`. Off-frame it falls back to `Date.now()`, unchanged.

## The one thing I proved before building

The motivating call binds every parameter — `elapsedTime.partial(since: start)` leaves a **zero-argument** tool. Whether a fully-bound partial survives tool-schema generation and an actual model tool call was never confirmed, and a failure would have reshaped `elapsedTime`'s signature. So I spiked it first, before writing any of the API: a fully-bound partial handed to a mocked `llm()` and called with empty args works end to end and applies the bound value. Only then did I build the rest. The `agentPull` test locks this in for `elapsedTime` specifically.

## A detail worth a look

Seeding `FakeClock.wallBaseMs` to a 2026 epoch (so a faked `today()`/`format(now())` isn't 1970) changes what `wallTime()` returns at construction, which #618's `clock.test.ts` pins. The seed is a single shared exported constant (`FAKE_CLOCK_WALL_BASE_MS`) that both `clock.ts` and its test reference, and the two assertions are updated — no duplicated literal, so the seam and the test can't drift.

## Verification

- `_formatDuration` unit cases: the format table, the second boundaries (999→`"0s"`, 1000→`"1s"`, 86_400_000→`"1d"`), negatives, and the negative-zero case `-500`→`"0s"`.
- An agency fixture, all under `fakeClock: true`: `elapsedTime(start)` after `_advanceTime(65000)` is `"1m 5s"` (the determinism test PR 1 couldn't write), `now() - start` is exactly `65000`, and the `agentPull` node proves `elapsedTime.partial(since: start)` is a valid zero-argument tool the model can call.
- Broad sweep across `lib/stdlib` and the runtime clock/guard tests: 1289 pass, 0 fail. `tsc` clean, `lint:structure` clean. Regenerated `date.md`.

This completes the #609 elapsedTime work. Spec and review are under `docs/superpowers/`.

### DIFF
--- a/packages/agency-lang/stdlib/date.agency
+++ b/packages/agency-lang/stdlib/date.agency
@@ -215,25 +215,21 @@ export def parse(iso: string): number {
 /** Render a millisecond duration as a readable string like "5m 32s". */
 export def formatDuration(ms: number): string {
   """
-  Render a duration in milliseconds as a compact human string, largest unit
-  first: "5m 32s", "1h 1s", "2d 3h". The granularity is whole seconds (a
-  sub-second duration is "0s") and the largest unit is days (a long duration
-  reads like "40d", never weeks or months). A negative duration gets a leading "-".
+  Render a duration in milliseconds as a readable string like "5m 32s" or
+  "2w 3d". Whole-second granularity; largest unit is weeks.
 
   @param ms - The duration in milliseconds
   """
   return _formatDuration(ms)
 }
 
-/** How long has elapsed since an instant, as a readable duration string. */
+/** How much time has elapsed since an instant, as a readable duration string. */
 export def elapsedTime(since: number): string {
   """
-  Returns how long has elapsed since the given instant, as a readable duration
-  like "5m 32s". Capture the start with `now()`, then call this to see the time
-  since. The largest unit is days, so a multi-day span reads like "2d 3h". If
-  you need the raw milliseconds for math, use `now() - since` instead.
+  How much time has elapsed since the given instant, as a readable duration
+  like "5m 32s". Capture the start with now().
 
-  @param since - The starting instant (epoch milliseconds), e.g. from now()
+  @param since - The starting instant, from now()
   """
   return formatDuration(now() - since)
 }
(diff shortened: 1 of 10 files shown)


packages/agency-lang/stdlib/date.agency
```diff
@@ -211,3 +211,29 @@ export def parse(iso: string): number {
   """
   return _parse(iso)
 }
+
+/** Render a millisecond duration as a readable string like "5m 32s". */
+export def formatDuration(ms: number): string {
+  """
+  Render a duration in milliseconds as a compact human string, largest unit
```
egonSchiele: this docstring serves as a tool description for an LLM. Lets keep it short please to save tokens.

packages/agency-lang/stdlib/date.agency
```diff
@@ -211,3 +211,29 @@ export def parse(iso: string): number {
...
+  first: "5m 32s", "1h 1s", "2d 3h". The granularity is whole seconds (a
+  sub-second duration is "0s") and the largest unit is days (a long duration
+  reads like "40d", never weeks or months). A negative duration gets a leading "-".
+
+  @param ms - The duration in milliseconds
+  """
+  return _formatDuration(ms)
+}
+
+/** How long has elapsed since an instant, as a readable duration string. */
```
egonSchiele: Please change these functions so the largest unit is weeks

egonSchiele:
All four addressed in c041590 (+ 316be93 for the regenerated docs).

**Largest unit is weeks** — `formatDuration` now goes up to `w`. A week is a fixed 7 days so it's unambiguous (months aren't, so those stay out). `"40d"` now reads `"5w 5d"`. Added weeks unit tests (`1w 1d`, `1w`, `5w 5d`).

**Terse docstrings** — shortened both to what a tool description should be. `formatDuration` is now two lines; `elapsedTime` two. Dropped the human-only "use `now() - since` for math" guidance a model can't act on.

**Grammar** (Copilot) — "Returns how long has elapsed" → "How much time has elapsed". The negative-`-` detail Copilot flagged is gone too, since the shortened docstring no longer claims it.

**Docs** — regenerated `date.md`. Note: the plain `make` build doesn't run the doc target, so my first push shipped stale generated docs; `make doc` fixed it, in the follow-up commit.

Verified: the date unit suite passes with the weeks cases, and the `elapsed-time` fixture is still 3/3 green.

PR #624, opened 2026-07-20
Fire top-level callbacks from imported modules

egonSchiele:
## Summary

Top-level `callback(...)` registrations in an **imported** Agency module never fired. The runtime only ran the **entry module's** `__registerTopLevelCallbacks`; every imported module compiled its own copy that nothing ever called. This is why the agency agent's trace lines (from `lib/agents/agency-agent/lib/trace.agency`) went missing when tracing moved into an imported module.

## The fix

Statics and globals solved this exact problem with a process-global registry (`crossModuleInitRegistry.ts`): every compiled module self-registers at JS-load, and one runtime driver iterates the whole registry. Callbacks had no equivalent — this PR adds it:

- A third registry (`callbackInits`) plus `__registerCallbacksInit`, mirroring `__registerGlobalsInit`.
- A driver, `__initAllRegisteredCallbacks(ctx)`, that clears `ctx.topLevelCallbacks` **once** and then runs every registered module's registration. The reset moved out of each per-module function into the driver — otherwise the second module's reset wipes the first module's callbacks.
- Codegen emits `__registerCallbacksInit("<moduleId>", __registerTopLevelCallbacks)` in every compiled module, right after the globals registration.
- The now-dead `registerTopLevelCallbacks` parameter is removed end to end (runtime signatures, builder, template wrappers).

## Safety-critical: the three registration paths stay in lockstep

Callbacks are safety infrastructure and are NOT serialized into checkpoints, so all three paths re-register through the same driver call:

- **Fresh run** (`node.ts` / `initFreshExecCtx`) — after global init, so callbacks can read module globals.
- **Resume** (`interrupts.ts` / `respondToInterrupts`) — before `restoreState`, so the empty stack routes registrations to `ctx.topLevelCallbacks` instead of a soon-unwound frame.
- **Rewind** (`rewind.ts` / `rewindFrom`) — same position as resume.

Each path has a regression test: `tests/agency/callback-imported.*` (fresh), `tests/agency/callback-imported-resume.*` (interrupt + resume), and a new js-driven `tests/agency-js/imported-module-callback-rewind/` that captures a checkpoint, calls the exported `rewindFrom`, and asserts the imported callback fires post-rewind.

## A deliberate tripwire tripped

`tests/agency-js/threads-fix-callback-propagation/` pinned the old negative behavior on purpose — its comment says a future change making imported callbacks fire "will trip this test and prompt the stdlib hook migration described in PR #225." It now pins the positive. Possible follow-up: `lib/stdlib/threads.ts` registers its onThreadEnd snapshotter via `registerGlobalHook(...)` as a workaround for this bug; that could migrate to a plain top-level `callback(...)` now.

## Scope notes

- The registry is process-global, like the existing statics/globals registries; process = closure for both `agency run` and the per-test node subprocess, so no closure-scoping was added (matches the documented globals limitation).
- ~96 generated-output snapshot fixtures under `tests/typescriptGenerator|typescriptBuilder|typescriptPreprocessor` regenerated via `make fixtures`; each module's diff is exactly the expected two-line change (reset removed, registration added) plus the import list.

## Verification

- Registry unit tests (`crossModuleInitRegistry.test.ts`).
- Fresh/resume/rewind fixtures above, plus no-regression runs of `callback-toplevel`, `callback-resume`, `callback-multi-same-hook`, `entry-module-callback-fires`.
- Real case: compiled `dist/lib/agents/agency-agent/lib/trace.js` now contains `__registerCallbacksInit(...)`, and a one-shot `agency agent -p --verbose` run shows the imported module's `⏺ toolName(...)` trace lines firing.
- `tsc --noEmit` clean, `pnpm run lint:structure` clean.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -4542,9 +4542,6 @@ export class TypeScriptBuilder {
                   messages: ts.id("messages"),
                   callbacks: ts.id("callbacks"),
                   initializeGlobals: ts.id("__initializeGlobals"),
-                  registerTopLevelCallbacks: ts.id(
-                    "__registerTopLevelCallbacks",
-                  ),
                   moduleDir: ts.id("__dirname"),
                 }),
               ])
--- a/packages/agency-lang/lib/backends/typescriptBuilder/sectionAssembler.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder/sectionAssembler.ts
@@ -465,6 +465,13 @@ export function assembleSections(opts: AssembleSectionsOpts): TsNode {
   );
 
   sections.push(buildRegisterTopLevelCallbacksFn(opts));
+  // And for top-level callbacks, so imported modules' registrations
+  // run too (the runtime driver iterates this registry).
+  sections.push(
+    ts.raw(
+      `__registerCallbacksInit(${JSON.stringify(displayModuleId(registryId))}, __registerTopLevelCallbacks);`,
+    ),
+  );
 
   sections.push(ts.statements(opts.generatedStatements));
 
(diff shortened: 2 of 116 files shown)


egonSchiele (commented):
## Review

Approach is right and the execution is clean — this mirrors the existing statics/globals registry faithfully, which is exactly the pattern that already solved this problem for init. I traced the whole change end to end and it holds together. A few small notes below, none blocking.

### What I verified

- **Reset moved correctly, no clobber.** The per-module `__ctx.topLevelCallbacks = []` reset is gone from generated code and the single reset now lives in `__initAllRegisteredCallbacks`. Checked across the regenerated fixtures: 93 modules gained the `__registerCallbacksInit(...)` call, and the only added reset in the whole diff is the one inside the driver. No fixture re-emits a per-module reset.
- **Three paths in lockstep.** Fresh (`node.ts`), resume (`interrupts.ts`), and rewind (`rewind.ts`) all route through the same `__initAllRegisteredCallbacks` call. On resume/rewind it runs before `restoreState`, same position as the old entry-only registration — and since a registration statement only stores a function reference (the callback *body* reads globals later, when the hook fires and state is restored), the pre-globals-restore position is fine. No regression versus prior behavior.
- **The reset can't wipe the threads hook.** `registerGlobalHook` (`lib/stdlib/threads.ts`) writes to the module-global `_globalHooks` in `hooks.ts`, not `ctx.topLevelCallbacks`, so the driver's reset leaves it untouched. Good — that was the one thing I most wanted to rule out given the safety-infra angle.
- **Tripwire flip is legitimate.** `threads-fix-callback-propagation` was written to trip on precisely this change (its own comment said so). Flipping it to the positive assertion (`importedCallbackFired: true`) is the intended outcome, and the new `entry-module-callback-fires` + `imported-module-callback-rewind` fixtures cover the other directions well.

### Minor suggestions (non-blocking)

1. **`agency serve` is the one long-lived caller — worth a sentence.** `runExportedFunction` runs the driver per request in a persistent process, so "process = closure" leans hardest here. I checked `createServeHandler` — it imports one module per handler and hot-reload re-registers under the same version-independent `moduleId` key (clean last-write-wins), so single-program serve is correct. The only exposure is a single process hosting several *distinct* compiled programs, which I don't think any run mode does today, and which is the same accepted limitation the globals registry already carries. Since serve is the long-lived path, consider a one-line note on `__initAllRegisteredCallbacks` calling it out explicitly (the current comment only names `agency run` and the test runner).

2. **Same-hook firing order is now observable.** Previously only the entry module's callbacks were live; now every module's are, appended in `Object.keys(callbackInits)` order, which is JS module-load order (dependencies before the entry). So for a hook multiple modules subscribe to, imported-module callbacks fire before entry-module ones. That's a reasonable choice, but it's a newly-defined behavior — worth one line in the driver's docstring so it's intentional and not incidental.

3. **Nit: type the cast.** The driver uses `(ctx as { topLevelCallbacks: unknown[] })`, while the rest of this file uses a named `CtxWithGlobals` type for the same idea. A tiny `CtxWithCallbacks` alias would match the existing style.

PR #629, opened 2026-07-21
Parse access chains on bracket literals

egonSchiele:
## What

Lets a bracket literal — a list comprehension or a plain array literal — be followed directly by an access chain: a method call, a bare property, an index, a slice, or a call. Both of these now parse and run:

```agency
const lines = ["- ${item}" for item in items].join("\n")
const block = "<criteria>${["- ${c}" for c in allCriteria].join("\n")}</criteria>"
```

Previously these failed to parse with `expected node body`. The workaround was to wrap the literal in parentheses (`([...]).join(...)`) or assign it to a variable first. This also fixes the same gap for plain array literals (`[1, 2, 3].join("-")`, `[10, 20, 30][1]`).

Comprehensions already worked inside string interpolation, including with nested `${...}` in the body — the only missing piece was attaching a chain, so both "comprehensions in interpolation" and "methods on comprehensions" reduce to this one change.

## How

Frontend-only. A new `bracketAccessParser` matches a comprehension-or-array literal followed by one or more chain elements and produces the exact same `valueAccess` node the parenthesized form already produced, so desugaring, type checking, and codegen are unchanged (a comprehension/array already sits happily in `valueAccess.base`). It is wired into `baseAtom` ahead of the bare bracket parsers; the `many1` requirement means an unchained `[...]` falls straight through to the existing parsers, so nothing regresses. It is `withLoc`-wrapped, mirroring `schemaAccessParser`, so the node carries a source location for diagnostics.

`agencyArrayParser` is switched from `trace` to `memo`. Placing the new parser first means every bare `[...]` is examined once for a chain and, finding none, re-parsed by the bare parser below; `comprehensionParser` is already memoized, so memoizing `agencyArrayParser` makes that second pass a cache hit. Measured effect: the array-literal parse cost returns to its pre-change level (the change is performance-neutral).

## Tests

- Parser unit tests in `comprehension.test.ts` covering method call, bare property, index, slice, an index-then-call chain (the only shape that yields a genuine `call` element, and a length-2 chain that proves `many1` chains), a chained `fork` comprehension, two "bare literal untouched" guards (catch a `many`-vs-`many1` slip), and a whitespace-adjacency guard.
- An Agency execution test (`tests/agency/comprehensions/chained-methods.agency`) whose node returns a combined string; the fixture asserts the exact content, so a wrong join/index result fails rather than only a crash.
- Full parser suite green (1655 tests); no regressions.

## Notes for reviewers

The `valueAccess` tail (`{ type: "valueAccess", base, chain }` plus its double-cast) now appears in three parsers. Extracting a shared `makeValueAccess(baseParser)` builder is a worthwhile follow-up but was deliberately left out here to avoid refactoring the two existing working parsers in a minimal frontend fix.

### DIFF
--- a/packages/agency-lang/docs/site/guide/comprehensions.md
+++ b/packages/agency-lang/docs/site/guide/comprehensions.md
@@ -105,25 +105,6 @@ const greetings = ["Hi ${name}" for {name, age} in people]
 const firsts = [a for [a, b] in pairs]
 ```
 
-## Calling methods on a comprehension
-
-A comprehension is an expression, so you can call a method on it, index it,
-or slice it directly - no need to assign it to a variable first:
-
-```ts
-const lines = ["- ${item}" for item in items].join("\n")
-```
-
-This works anywhere an expression is allowed, including inside string
-interpolation:
-
-```ts
-const block = "<criteria>${["- ${c}" for c in allCriteria].join("\n")}</criteria>"
-```
-
-The same holds for plain array literals: `[1, 2, 3].join("-")` and
-`[10, 20, 30][1]`.
-
 ## References
 
 - [Concurrency](/guide/concurrency)
--- a/packages/agency-lang/docs/site/guide/comprehensions.md
+++ b/packages/agency-lang/docs/site/guide/comprehensions.md
@@ -105,6 +105,25 @@ const greetings = ["Hi ${name}" for {name, age} in people]
 const firsts = [a for [a, b] in pairs]
 ```
 
+## Calling methods on a comprehension
+
+A comprehension is an expression, so you can call a method on it, index it,
+or slice it directly - no need to assign it to a variable first:
+
+```ts
+const lines = ["- ${item}" for item in items].join("\n")
+```
+
+This works anywhere an expression is allowed, including inside string
+interpolation:
+
+```ts
+const block = "<criteria>${["- ${c}" for c in allCriteria].join("\n")}</criteria>"
+```
+
+The same holds for plain array literals: `[1, 2, 3].join("-")` and
+`[10, 20, 30][1]`.
+
 ## References
 
 - [Concurrency](/guide/concurrency)
--- a/packages/agency-lang/lib/parsers/parsers.ts
+++ b/packages/agency-lang/lib/parsers/parsers.ts
@@ -2178,7 +2178,7 @@ const literalDelimiter = (closer: string) =>
     optionalSpacesOrNewline,
   );
 
-export const agencyArrayParser: Parser<AgencyArray> = trace(
+export const agencyArrayParser: Parser<AgencyArray> = memo(
   "agencyArrayParser",
   map(
     seqC(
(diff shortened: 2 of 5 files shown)


egonSchiele (commented):
## Review: Parse access chains on bracket literals

Clean, tightly scoped frontend change. I reviewed the parser internals against the source and the logic holds: `bracketAccessParser` produces the same `valueAccess` node the paren form already produced, `many1` makes unchained `[...]` fall through untouched, and the `withLoc` wrap matches `schemaAccessParser` (the other bare `valueAccess` entry in `baseAtom`). The `comprehensionParser`-before-`agencyArrayParser` ordering is correct and mirrors `baseAtom`.

**Correctness** — no issues found. The chain-element `kind` assertions in the tests are accurate against the parsers (`.join()` → `methodCall`, `[i]` → `index`, `[a:b]` → `slice`, index-then-call → `call`). The `[f, g][0]("x")` test is a nice single case that covers both the only real `call`-element shape and the first length-2 chain (proving `many1` chains). The execution test's `expectedOutput` quote-wrapping matches the sibling `basic.test.json` convention, and it pins exact content so a wrong join/index result fails rather than only a crash.

**Conventions** — follows existing parser style, comments explain the load-bearing ordering, docs added to the guide. No anti-patterns (declarative combinator code, no imperative accumulation).

**Two minor points**, both non-blocking — left inline:
1. `agencyArrayParser` switched `trace` → `memo` (the one non-additive change).
2. Slice and `fork`-chain are parse-tested but not execution-tested.

The acknowledged `makeValueAccess` follow-up (the tail now appears in three parsers) is the right call to defer — worth filing as a tracking issue so it isn't lost.

Overall: LGTM once you've eyeballed the two inline notes.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2178,7 +2178,7 @@ const literalDelimiter = (closer: string) =>
     optionalSpacesOrNewline,
   );
 
-export const agencyArrayParser: Parser<AgencyArray> = trace(
+export const agencyArrayParser: Parser<AgencyArray> = memo(
```
egonSchiele, line 2181: `trace` → `memo` is the only non-additive change in this PR, so worth a second look. It's safe — `agencyArrayParser` is referentially transparent, so memoizing can't change parse *results*, only reuse them (and it's what makes the "parse every `[...]` once, cache the bare re-parse" story hold, plus it matches the 49-vs-7 memo/trace majority). The one thing you lose is `DEBUG=1` trace output for this parser specifically. If array-literal parser tracing was ever useful for debugging, note that it's gone; otherwise this is a clean win.

packages/agency-lang/tests/agency/comprehensions/chained-methods.agency
```diff
@@ -0,0 +1,16 @@
...
+// output would fail, not just a crash.
+node main(): string {
+  const xs = ["a", "b", "c"]
+
+  const bulleted = ["- ${c}" for c in xs].join("\n")
+  const dashed = [1, 2, 3].join("-")
+  const middle = [10, 20, 30][1]
+  const inline = "vals: ${["- ${c}" for c in xs].join(", ")}"
+
+  return "${bulleted}|${dashed}|${middle}|${inline}"
```
egonSchiele: Optional: the parser unit tests cover slice and a `fork`-comprehension chain, but this end-to-end fixture doesn't exercise either at runtime — it covers `.join` (comprehension + array), index, and interpolation. Downstream is provably unchanged (same `valueAccess` node), so parse coverage arguably suffices. But since a slice on a bracket literal (`[10, 20, 30][0:2]`) and a chained `fork` are the two newly-enabled shapes with the most moving parts, one extra line each here would make the runtime story airtight. Non-blocking.

packages/agency-lang/docs/site/guide/comprehensions.md
```diff
@@ -105,6 +105,25 @@ const greetings = ["Hi ${name}" for {name, age} in people]
 const firsts = [a for [a, b] in pairs]
 ```
 
+## Calling methods on a comprehension
```
egonSchiele: undo all changes to this file

packages/agency-lang/tests/agency/comprehensions/chained-methods.agency
egonSchiele: You can return an object, no need to return a string, this is less readable

packages/agency-lang/docs/site/guide/comprehensions.md
egonSchiele: Done — reverted comprehensions.md to main in b147e4df2.

packages/agency-lang/tests/agency/comprehensions/chained-methods.agency
egonSchiele: Done — main now returns an object; the fixture asserts its compact-JSON serialization. Reads cleaner than the concatenated string.

egonSchiele: Added both to the fixture while reworking it to return an object: sliced: [10, 20, 30][0:2].join(",") and forked: fork ["<${c}>" for c in xs].join("|"). Runtime output pinned ("10,20" and "<a>|<b>|<c>" — fork preserves input order).

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Agreed it's safe. One nuance: memo() internally wraps its parser in trace() (labeled memo(agencyArrayParser)), so DEBUG=1 still emits trace lines for this parser — just under the memo name rather than the bare one. So tracing isn't lost, only relabeled.

Issue #630, opened 2026-07-21
Runtime validation drops @validate tags inside built-in generics (Record verified)


egonSchiele:
## Summary

Runtime validation does not fully cover the built-in generic types. Verified today: `@validate` tags on a type alias are **silently dropped** when the alias is used inside `Record<...>`. The value validates against the Zod shape only, so values the validator should reject pass the bang.

## Repro (verified on `type-patterns-spec`, 2026-07-20)

```
def isPositive(v: number): Result<number> {
  if (v <= 0) {
    return failure("must be positive")
  }
  return success(v)
}

@validate(isPositive)
type Age = number;

node main() {
  const ages: Record<string, Age>! = { a: -5 }
  print(ages)
}
```

Generated validation site:

```js
await __validateChainRecursive(__stack.locals.ages, {
  "kind": "leaf",
  "schema": z.record(z.string(), z.number()),
  "validators": []          // <-- isPositive is gone
});
```

`{ a: -5 }` passes. Expected: the `Age` validator runs per value and this fails with "must be positive".

## What works vs. does not

- `Partial<Person>!` (no validators involved): works — generates a correct per-field-optional Zod schema.
- `Record<string, AliasWithValidate>!`: **validators dropped** (repro above).
- Per the comments in `lib/typeChecker/builtinGenerics.ts` (lines ~9-30, ~82), `Array`/`Schema`/`Record` deliberately DROP use-site tags ("historical behavior, preserved") while `Partial`/`Pick`/etc. merge them. The nested-alias case above suggests the drop also swallows tags carried by the element/value type itself, not just use-site tags. Worth auditing each builtin generic (`Record`, `Array`, `Schema`, `Partial`, `Required`, `Pick`, `Omit`, `NonNullable`) for both tag sources.

## Why it matters more soon

The type-patterns feature (spec `docs/superpowers/specs/2026-07-20-type-patterns-design.md`) routes `x is Record<string, Age>` through this same descriptor machinery, so the pattern would silently test shape-only where the spec promises "validators included".

egonSchiele:
## Root cause (investigated)

The validator information survives all the way to the descriptor builder and is dropped at exactly one place:

1. Alias resolution preserves validate tags (`resolveTypeWithGuard` → `attachAliasTags`, `lib/typeChecker/assignability.ts`), so `Age` resolves to `number` still carrying `@validate(isPositive)`.
[...]

PR #631, opened 2026-07-21
Type patterns: matching on the type of a value

egonSchiele:
Implements the type-patterns spec: runtime type tests as patterns, with narrowing.

Spec: `docs/superpowers/specs/2026-07-20-type-patterns-design.md`
Plan: `docs/superpowers/plans/2026-07-20-type-patterns.md` (reviews alongside)

## What this adds

Two spellings of one construct:

- `x is Type` — in `if`/`while` conditions and boolean expressions. Narrows the value in the branch where the test succeeded (positive-only).
- `pattern: Type` — in match arms: `s: string =>`, `{name}: Person =>`, `[a, b]: number[] =>`, plus the test-only arm `is boolean =>`.

Tier 1 (string/number/boolean/null/object/any[]) compiles to a coarse runtime check (`__coarseTypeTest`). Everything else reuses the bang validation path (`validateExpr` → Zod schema + `@validate` validators), so "is this a Person" and "validate against Person" are the same question. The pattern binds the ORIGINAL value — validators decide the match, never rewrite (pinned by the `bindsOriginal` execution test with a transforming validator).

Also ships `Json` / `JsonObject` in `std::validation`: a precise round-trips-through-JSON validator (rejects class instances, NaN/Infinity, functions, cycles, with the failing path), usable as `x is Json` and `const d: Json! = ...`.

## DELIBERATE BEHAVIOR CHANGE (review this first)

The always-true bare-identifier binder on the right of `is` is retired: `x is y` used to bind `y` and match anything; a bare identifier after `is` is now ALWAYS a type reference, and an unresolvable name is a compile error (AG1013) suggesting `const y = x`. Evidence: a recorded grep across tests/, lib/, examples/, and both docs trees found zero code uses (77 hits, all English prose in comments). Match-arm bare binders (`other => ...`) are unchanged.

A second, smaller change: the `is` operator now requires its subject and the `is` keyword on the same line. Previously `1\nis number` could merge across a newline into one expression, which made an arm body swallow the next arm.

## New diagnostics

- AG1013 (error): name after `is` is not a type; tailored hint for JS classes (`Date`, `Map`, ...) — type patterns never mean `instanceof`.
- AG5003 (warning): un-guarded bare-binder arm named like a type (`Person =>` binds; did you mean `p: Person`?).
- AG5004 (warning): property-position binder named like a type (`{name: string}` binds the field to a variable called `string`).

## Semantics pinned by tests

- Type-pattern arms never satisfy exhaustiveness (a validator can reject a value whose static type looks right) — `_` still required, even for coarse arms fully covering a closed union.
- Narrowing tests use union scrutinees (never `any`, which masks narrowing) and fail when narrowing breaks.
- Loose null: `is null` parses as the existing literal pattern; the coarse null check matches `undefined` too, so all three null spellings agree on interop values.
- `is object` = any non-null, non-array object (class instances included), narrowing to the opaque `object` primitive; field access after it stays an error.
- Formatter normalizes `_: Type` to `is Type` (same parse node; documented).

## Known v1 limits (per spec)

- Positive-only narrowing; bare-variable and stable-path subjects.
- Inline unions not spellable in pattern position (named alias instead).
- Validators through `Record`/`Array`/`Schema` inherit #630 (shape-only until fixed).

Test coverage: parser unit tests (is-RHS, arm suffix, precedence, regressions), lowering tests (binder-from-original, destructure gate, matchSource), typechecker tests (falsifiable narrowing, AG1013 exclusivity, exhaustiveness, shadow warnings), codegen fixture, 9 execution tests, formatter round-trips. 5392 unit tests green locally; agency suite runs in CI.


packages/agency-lang/lib/typeChecker/narrowing.ts
```diff
@@ -64,6 +69,11 @@ export function narrowByRefine(
       );
     case "presence":
       return narrowUnionByPresence(current, refine.present, aliases);
+    case "typeTest":
+      // The test succeeding means the value IS the tested type; that type is
+      // strictly more precise than (or equal to) whatever we knew before, so
+      // it simply replaces the current type in the guarded region.
+      return refine.testedType;
```
egonSchiele, line 82: The comment above says the tested type is "strictly more precise than (or equal to) whatever we knew before", but replacing (rather than intersecting) can actually widen. Example:

```
def f(x: "a" | "b"): "a" | "b" {
  if (x is string) {
    return x   // x is now `string`, no longer "a" | "b"
  }
  return "a"
}
```

Inside the branch `x` widens from `"a" | "b"` to `string`, so returning it (or matching on it exhaustively) now errors even though the runtime test can only succeed. A coarse `is string` on any string-literal union hits this.

Cheap v1 fix without building a general intersection: if `current` is already assignable to `testedType`, keep `current` (the test can't tell you anything you didn't know); otherwise replace as you do now. That preserves the union-member filtering behavior the tests pin while never widening.

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
```diff
@@ -217,6 +217,53 @@ function checkSite(site: MatchSite, configured: Severity, ctx: TypeCheckerContex
...
+    aliases[name] !== undefined || SHADOWABLE_PRIMITIVE_NAMES.includes(name);
+  for (const arm of site.arms) {
+    if (arm.caseValue === "_") continue;
+    const cv = arm.caseValue;
+    if (!arm.guarded && cv.type === "variableName" && isTypeName(cv.value)) {
+      ctx.errors.push(
+        diagnostic("bareArmBinderShadowsType", { name: cv.value }, site.loc ?? null),
+      );
+    }
+    if (cv.type === "objectPattern") {
```
egonSchiele: AG5004 only fires when the arm's caseValue is itself an `objectPattern`, which misses the two places the confusion is most likely:

1. **Typed arms**: `{name: string}: Person => ...` has `cv.type === "typePattern"` wrapping the object pattern, so no warning — yet a typed arm is exactly where a user is thinking in types and most likely to believe `{name: string}` tests the field. Recursing into `cv.pattern` when `cv.type === "typePattern"` would cover it.
2. **Nested object patterns**: `{a: {name: string}} => ...` — the loop checks `prop.value.type === "variableName"` but doesn't recurse into nested `objectPattern` values.

Case 1 seems worth fixing in this PR since type patterns are what create the trap; case 2 could be a follow-up.

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
```diff
@@ -217,6 +217,53 @@ function checkSite(site: MatchSite, configured: Severity, ctx: TypeCheckerContex
...
+function warnTypeShadowingBinders(site: MatchSite, ctx: TypeCheckerContext): void {
+  const aliases = ctx.getTypeAliases();
+  const isTypeName = (name: string): boolean =>
+    aliases[name] !== undefined || SHADOWABLE_PRIMITIVE_NAMES.includes(name);
+  for (const arm of site.arms) {
+    if (arm.caseValue === "_") continue;
+    const cv = arm.caseValue;
+    if (!arm.guarded && cv.type === "variableName" && isTypeName(cv.value)) {
+      ctx.errors.push(
+        diagnostic("bareArmBinderShadowsType", { name: cv.value }, site.loc ?? null),
```
egonSchiele: Both warnings anchor to `site.loc` (the whole match), so in a ten-arm match the squiggle lands on `match (` rather than the offending arm. The pattern nodes carry their own locations (they're `BaseNode`s) — `cv.loc ?? site.loc ?? null` here, and `prop.value.loc` (or `cv.loc`) for AG5004 below, would point users at the actual binder.

egonSchiele (commented):
Reviewed the core implementation (parser, lowering, builder, narrowing, exhaustiveness, runtime, stdlib) plus the execution tests. Overall this is in good shape:

- Reusing the bang validation path for Tier 2 keeps "is this a Person" and "validate against Person" the same question, and I verified both `__validateType` and `__validateChainRecursive` return a `ResultValue` rather than throwing, so the `isSuccess(...)` wrapping is sound.
- `validateExpr` embeds the tested value exactly once, so side-effecting subjects don't double-evaluate on the Tier 2 path either.
- The loose-null story is coherent: the literal `null` pattern, `is null`, and the coarse `"null"` kind all agree on `undefined`, and the test file pins it.
- The `bindsOriginal` execution test with a transforming validator is exactly the right pin for Rule 1.
- The typechecker tests' discipline of using union scrutinees (never `any`) so they fail when narrowing breaks is great.
- The `x is y` binder retirement is well-evidenced (recorded grep, zero code uses) and AG1013's `const y = x` hint gives displaced users a clear path.

I left three inline comments: one on narrowing precision (replace vs intersect), one on an AG5004 coverage gap for typed arms, and one on warning locations. Two smaller observations that didn't fit a line:

- `SHADOWABLE_PRIMITIVE_NAMES` omits `any`, so a bare `any => ...` arm binds a variable named `any` with no AG5003. Probably rare, but it reads like a type test at least as much as `object => ...` does.
- A Tier 2 `is Person` in a `while` condition re-runs the full recursive validation (Zod parse + validator chain) every iteration. That's semantically correct and probably fine, but might deserve a sentence in the pattern-matching guide so users don't put expensive `@validate` validators on types they test in hot loops.

packages/agency-lang/lib/stdlib/jsonValue.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type JsonCheckResult = { ok: boolean; path: string; reason: string };
+
+/** stdlib bridge for `std::validation`'s `isJsonValue`. Returns a plain
+ *  object (never throws) so the Agency wrapper can build the failure
+ *  message with the offending path. */
+export function _isJsonValue(value: unknown): JsonCheckResult {
+  const result = __isJsonValue(value);
+  if (result.ok) {
+    return { ok: true, path: "", reason: "" };
```
egonSchiele: Would it make sense to use the `Result` type here? https://agency-lang.com/guide/error-handling.html

packages/agency-lang/lib/typeChecker/narrowing.ts
egonSchiele: Confirmed and fixed in a956ff5. Implemented exactly the suggested guard with one addition: `any` is excluded from the keep-current check (it is assignable to everything, but escaping `any` is the point of the test), so `any` scrutinees still narrow. Regression test added with your exact example.

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
egonSchiele: Both fixed in a956ff5 — case 1 (typed arms recurse into the inner pattern) and case 2 (nested object patterns) since the recursion covered it in the same helper. Tests added for both.

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
egonSchiele: Fixed in a956ff5. One wrinkle: bare variableName arm patterns carried no loc at all (the arm parser never attached one), so the fallback silently hit site.loc. The pattern alternative in caseLhsParser is now wrapped in withLoc, and a test pins that the warning anchors below the first arm rather than at the match head.

packages/agency-lang/lib/stdlib/jsonValue.ts
egonSchiele: Done in a956ff5 — the bridge now returns the runtime Result directly (success/failure with the path folded into the message), matching validators.ts, and the Agency wrapper collapsed to a one-line passthrough like isEmail.

packages/agency-lang/lib/runtime/jsonValue.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+    const nested = [...ancestors, value];
+    for (let i = 0; i < value.length; i++) {
+      const result = walkJson(value[i], `${path}[${i}]`, nested);
+      if (!result.ok) {
+        return result;
+      }
+    }
+    return { ok: true };
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in a956ff5: the array branch now requires enumerable keys to be exactly the indices (Object.keys length === array length), which rejects both extra properties and sparse holes (holes parse back as null, so they do not round-trip either). Test added.

packages/agency-lang/lib/runtime/jsonValue.ts
```diff
@@ -0,0 +1,56 @@
...
+        (value as Record<string, unknown>)[key],
+        path === "" ? key : `${path}.${key}`,
+        nested,
+      );
+      if (!result.ok) {
+        return result;
+      }
+    }
+    return { ok: true };
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in a956ff5: objects now reject symbol-keyed properties and non-enumerable own properties; arrays reject symbol keys too. (Non-enumerable extras on arrays are not checked — length itself is non-enumerable — but the keys-vs-length check covers the enumerable cases.) Tests added.

packages/agency-lang/lib/runtime/jsonValue.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  test("rejects non-round-tripping values with a path", () => {
+    expect(__isJsonValue(new Date()).ok).toBe(false);
+    expect(__isJsonValue({ a: new Map() }).ok).toBe(false);
+    expect(__isJsonValue({ a: [1, NaN] })).toMatchObject({ ok: false, path: "a[1]" });
+    expect(__isJsonValue(Infinity).ok).toBe(false);
+    expect(__isJsonValue(undefined).ok).toBe(false);
+    expect(__isJsonValue({ f: () => 1 }).ok).toBe(false);
+  });
```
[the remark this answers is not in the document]
egonSchiele: Added in a956ff5: sparse arrays, arrays with extra enumerable properties, symbol-keyed objects and arrays, and non-enumerable own properties are all pinned.

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
```diff
@@ -217,6 +217,53 @@ function checkSite(site: MatchSite, configured: Severity, ctx: TypeCheckerContex
...
+      );
+    }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in a956ff5 (see the thread on line 244 — bare binders needed loc plumbing in the arm parser first).

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
```diff
@@ -217,6 +217,53 @@ function checkSite(site: MatchSite, configured: Severity, ctx: TypeCheckerContex
...
+          prop.value.type === "variableName" &&
+          isTypeName(prop.value.value)
+        ) {
+          ctx.errors.push(
+            diagnostic(
+              "propertyBinderShadowsType",
+              { field: prop.key, name: prop.value.value },
+              site.loc ?? null,
+            ),
+          );
```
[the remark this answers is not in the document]
egonSchiele: Fixed in a956ff5: AG5004 anchors to prop.value.loc with pattern/site fallbacks.

PR #634, opened 2026-07-21
Fix LSP false import errors on unsaved edits

egonSchiele:
## Problem

Typing an import in the editor before saving produced two confusing, wrong diagnostics — and this was the report that kicked this off (using `now`/`elapsedTime` from `std::date`):

1. An error on the import line: **"Symbol 'now' is not defined in 'std::date'"** — even though `now` *is* exported by `std::date`.
2. Every other squiggle in the file disappeared. A genuinely-undefined `elapsedTime` used below stopped being flagged.

And saving the file did **not** clear it.

## Root cause

The LSP used two different sources of truth. It built the symbol table from the **saved file on disk** (`SymbolTable.build(fsPath)`) but ran diagnostics on the **live editor buffer** (`doc.getText()`).

When the buffer imports a module the saved file does not, the disk-built symbol table never crawled that module. `resolveImports` then can't find the symbol and throws. That throw was caught in `diagnostics.ts` and turned into a single `0:0` error with an early `return` — **before the type checker ran** — so all other diagnostics were wiped.

Saving didn't help because a `.agency` save fires neither `onDidChangeContent` (buffer content is unchanged by a save) nor the watched-files rebuild (which only handled `agency.json`). Nothing re-analyzed the file, so the stale diagnostics from the last keystroke persisted.

## Fixes

**1. Use the editor buffer for the active file's symbol table.** `SymbolTable.build` now takes an `overrides` map (absolute path → in-memory source); the server feeds the live buffer for the active document. A just-typed import resolves immediately, no save required.

**2. Degrade gracefully on an unresolvable import.** `resolveImports` gains an LSP-only `skipUnresolvable` mode: instead of throwing on the first bad import and aborting the whole rewrite, it keeps the offending statement unrewritten and resolves the rest. The type checker's existing `checkMissingImports` pass already reports each bad import **once** at its own location (AG4008/4009/4010), so there's no duplicate and no `0:0` fallback. Two consequences the old code got wrong are now correct:
   - The type checker still runs, so unrelated diagnostics survive.
   - A single bad import no longer cascades every *other* imported symbol to "undefined".

   The compile path leaves the flag off and still hard-fails on bad imports.

**3. Rebuild open documents when a `.agency` file changes on disk.** The server registers a `**/*.agency` watcher (dynamically, when the client supports it) and rebuilds all open documents on any Agency file change — so saving an *imported* file refreshes the documents that import it. It also evicts that file's parse-cache entry, closing the `mtime`+`size` staleness blind spot the cache already documents for long-lived processes.

## Verification

New regression tests:
- `diagnostics.test.ts` — a just-typed import (buffer only) resolves clean; a partial import keeps the unrelated `elapsedTime` squiggle; a bad-plus-good import reports the bad one once and still resolves the good one.
- `symbolTable.test.ts` — `build` crawls an import present only in the override, and builds a not-yet-saved entrypoint from its override.
- `importResolver.test.ts` — `skipUnresolvable` keeps the bad statement and resolves good imports around it.
- `parseCache.test.ts` — `evictParseCache` forces a re-parse even when mtime+size match.

Full `lib/lsp` and `lib/typeChecker` suites pass (1740 tests), plus the affected unit files. `tsc`, structural lint, and build are clean.

### DIFF
--- a/packages/agency-lang/lib/lsp/diagnostics.ts
+++ b/packages/agency-lang/lib/lsp/diagnostics.ts
@@ -119,13 +119,9 @@ export function runDiagnostics(
 
   try {
     program = resolveReExports(program, symbolTable, fsPath);
-    // Analysis-only path (the LSP never executes anything): honor
-    // `import test` so migrated test files keep full editor support instead
-    // of dying on a single 0:0 error. Execution paths still default to deny.
-    program = resolveImports(program, symbolTable, fsPath, {
-      allowTestImports: true,
-    });
   } catch (err) {
+    // A re-export failure (e.g. a cycle) leaves the module graph unusable, so
+    // there is nothing meaningful left to type-check. Report and stop.
     diagnostics.push({
       severity: DiagnosticSeverity.Error,
       range: { start: { line: 0, character: 0 }, end: { line: 0, character: 1 } },
@@ -134,6 +130,19 @@ export function runDiagnostics(
     });
     return { diagnostics, program: null, info: null, semanticIndex: {}, scopes: [] };
   }
+
+  // Analysis-only path (the LSP never executes anything):
+  //  - `allowTestImports` honors `import test` so migrated test files keep full
+  //    editor support instead of dying on a single 0:0 error.
+  //  - `skipUnresolvable` drops any import that can't be resolved instead of
+  //    aborting the whole rewrite. That keeps every *other* import (and the
+  //    rest of the file) type-checked; the type checker's `checkMissingImports`
+  //    pass reports each bad import at its own location (AG4008/4009/4010), so
+  //    there is no need to surface a duplicate here.
+  program = resolveImports(program, symbolTable, fsPath, {
+    allowTestImports: true,
+    skipUnresolvable: true,
+  });
 
   const info = buildCompilationUnit(program, symbolTable, fsPath, source);
   const { errors, scopes, interruptEffectsByFunction } = typeCheck(program, config, info);
--- a/packages/agency-lang/lib/lsp/diagnostics.ts
+++ b/packages/agency-lang/lib/lsp/diagnostics.ts
@@ -134,15 +134,37 @@ export function runDiagnostics(
   // Analysis-only path (the LSP never executes anything):
   //  - `allowTestImports` honors `import test` so migrated test files keep full
   //    editor support instead of dying on a single 0:0 error.
-  //  - `skipUnresolvable` drops any import that can't be resolved instead of
-  //    aborting the whole rewrite. That keeps every *other* import (and the
-  //    rest of the file) type-checked; the type checker's `checkMissingImports`
-  //    pass reports each bad import at its own location (AG4008/4009/4010), so
-  //    there is no need to surface a duplicate here.
-  program = resolveImports(program, symbolTable, fsPath, {
-    allowTestImports: true,
-    skipUnresolvable: true,
-  });
+  //  - `onUnresolvable` drops any import that can't be resolved (instead of
+  //    aborting the whole rewrite) and reports it at its own location, so every
+  //    *other* import and the rest of the file still type-check.
+  try {
+    program = resolveImports(program, symbolTable, fsPath, {
+      allowTestImports: true,
+      onUnresolvable: (err) => {
+        const loc = err.loc;
+        const range = loc
+          ? { start: doc.positionAt(loc.start), end: doc.positionAt(loc.end) }
+          : { start: { line: 0, character: 0 }, end: { line: 0, character: 1 } };
+        diagnostics.push({
+          severity: DiagnosticSeverity.Error,
+          range,
+          message: err.message,
+          source: "agency",
+        });
+      },
+    });
+  } catch (err) {
+    // Defensive: `onUnresolvable` absorbs every expected import failure, so a
+    // throw here is unexpected. Report it but keep the (unrewritten) program so
+    // the type checker still runs — a single import must never blank the file
+    // or crash the server (updateDocument runs in a bare debounce callback).
+    diagnostics.push({
+      severity: DiagnosticSeverity.Error,
+      range: { start: { line: 0, character: 0 }, end: { line: 0, character: 1 } },
+      message: err instanceof Error ? err.message : String(err),
+      source: "agency",
+    });
+  }
 
   const info = buildCompilationUnit(program, symbolTable, fsPath, source);
   const { errors, scopes, interruptEffectsByFunction } = typeCheck(program, config, info);
--- a/packages/agency-lang/lib/lsp/server.ts
+++ b/packages/agency-lang/lib/lsp/server.ts
@@ -103,14 +103,17 @@ export function startServer(): void {
 
     let symbolTable = new SymbolTable();
     try {
-      // Feed the live editor buffer for the active file so an unsaved edit
-      // (e.g. a just-typed `import`) is reflected in the symbol table. Building
-      // purely from disk would resolve imports against the stale saved file,
-      // making `resolveImports` reject symbols from a module the buffer imports
-      // but the saved file does not.
-      symbolTable = SymbolTable.build(fsPath, config, {
-        [path.resolve(fsPath)]: doc.getText(),
-      });
+      // Feed every open document's live buffer as an override so unsaved edits
+      // (e.g. a just-typed `import`) are reflected in the symbol table. Building
+      // purely from disk would resolve imports against the stale saved files,
+      // making `resolveImports` reject symbols from a module a buffer imports
+      // but the saved file does not — for the active file OR any open file it
+      // imports.
+      const overrides: Record<string, string> = {};
+      for (const open of documents.all()) {
+        overrides[path.resolve(uriToPath(open.uri))] = open.getText();
+      }
+      symbolTable = SymbolTable.build(fsPath, config, overrides);
     } catch {
       // If symbol table build fails (e.g. file not on disk yet), continue with empty table
     }
--- a/packages/agency-lang/lib/lsp/server.ts
+++ b/packages/agency-lang/lib/lsp/server.ts
@@ -6,10 +6,14 @@ import {
   TextDocuments,
   TextDocumentSyncKind,
   InitializeResult,
+  InitializeParams,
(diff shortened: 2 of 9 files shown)


packages/agency-lang/lib/lsp/diagnostics.ts
```diff
@@ -135,6 +131,19 @@ export function runDiagnostics(
...
 
+  // Analysis-only path (the LSP never executes anything):
+  //  - `allowTestImports` honors `import test` so migrated test files keep full
+  //    editor support instead of dying on a single 0:0 error.
+  //  - `skipUnresolvable` drops any import that can't be resolved instead of
+  //    aborting the whole rewrite. That keeps every *other* import (and the
+  //    rest of the file) type-checked; the type checker's `checkMissingImports`
+  //    pass reports each bad import at its own location (AG4008/4009/4010), so
+  //    there is no need to surface a duplicate here.
+  program = resolveImports(program, symbolTable, fsPath, {
```
egonSchiele: This call is no longer inside a try/catch, and `skipUnresolvable` only absorbs `ImportResolutionError`. `resolveImportStatement` also calls `resolveAgencyImportPath`, which throws plain `Error` for unresolvable `pkg::` imports (`findPkgDir` when the package isn't installed, `validatePkgRelativePath` rejections, a bad `"agency"` package.json field). Any of those now escapes `runDiagnostics` entirely.

In `server.ts` the debounced path runs `updateDocument` inside a bare `setTimeout` callback, so the escape is an uncaught exception that can kill the LSP process — and it fires on every keystroke while the bad import is in the buffer. Before this PR the same input produced the 0:0 diagnostic.

Suggest wrapping this call in its own try/catch that falls back to the old 0:0 diagnostic (or converting the `pkg::` resolution failures into `ImportResolutionError` inside `resolveImportStatement`, which would also anchor them to the import line).

packages/agency-lang/lib/preprocessors/importResolver.ts
```diff
@@ -48,157 +67,241 @@ export function resolveImports(
...
+      );
+    } catch (err) {
+      // Analysis mode (LSP): a single unresolvable import must not abort the
+      // whole rewrite — that would leave every *other* import unresolved and
+      // reported as undefined at its use sites. Keep the original (unrewritten)
+      // statement so every other import still resolves AND the type checker's
+      // `checkMissingImports` pass can still see this node and report the
+      // specific bad name at its own location (AG4008/4009/4010). The compile
+      // path leaves `skipUnresolvable` off and still hard-fails here.
+      if (skipUnresolvable && err instanceof ImportResolutionError) {
```
egonSchiele: The comment says `checkMissingImports` re-reports what gets skipped here, but that's only true for the name-not-found / module-not-found / not-exported cases (AG4008/4009/4010). Two other `ImportResolutionError` throws get swallowed with no checker counterpart:

- the retry-safety marker applied to a node (`import { destructive someNode }`)
- the `import test` gates for `pkg::` / TS module paths (the `allowTestImports` gate can't fire in the LSP, but these two can)

For those, the editor now shows nothing while `agency compile` still hard-fails — the exact editor/compiler disagreement this PR is fixing elsewhere. Options: collect skipped `ImportResolutionError`s and let the LSP surface them at `err.loc` (the class already carries it), or extend `checkMissingImports` to cover these two rules.

packages/agency-lang/lib/lsp/server.ts
```diff
@@ -65,13 +78,39 @@ export function startServer(): void {
...
 
     let symbolTable = new SymbolTable();
     try {
-      symbolTable = SymbolTable.build(fsPath, config);
+      // Feed the live editor buffer for the active file so an unsaved edit
+      // (e.g. a just-typed `import`) is reflected in the symbol table. Building
+      // purely from disk would resolve imports against the stale saved file,
+      // making `resolveImports` reject symbols from a module the buffer imports
+      // but the saved file does not.
+      symbolTable = SymbolTable.build(fsPath, config, {
```
egonSchiele: Only the active document's buffer is fed as an override, so a dirty-but-unsaved *imported* file still resolves from its saved contents. The original bug survives in cross-file form: add an `export` to `b.agency`, switch to `a.agency` without saving, type `import { helper } from "./b.agency"` — false "not defined" error until `b` is saved.

Since `overrides` is already a map, building it from every open document fixes this for the same cost:

```ts
const overrides: Record<string, string> = {};
for (const open of documents.all()) {
  overrides[path.resolve(uriToPath(open.uri))] = open.getText();
}
```

Fine as a follow-up if you'd rather keep this PR scoped to the single-file case, but worth a TODO either way.

egonSchiele (commented):
Reviewed the whole diff. The disk-vs-buffer split diagnosis is right, the fix is well-layered (buffer override, graceful import degradation, watcher + cache eviction), and the tests pin each layer including the mtime+size blind spot. Two real issues and one gap worth addressing before merge — the first one is a crash regression.

**1. Non-`ImportResolutionError` throws now escape `runDiagnostics` and can kill the server** (inline comment on `diagnostics.ts`). `resolveImports` moved out of the try/catch, and `skipUnresolvable` only swallows `ImportResolutionError`. But `resolveImportStatement` also calls `resolveAgencyImportPath`, which throws plain `Error` for `pkg::` imports that don't resolve (`findPkgDir` when the package isn't installed, `validatePkgRelativePath`, bad `agency` field, etc.). Before this PR that became a 0:0 diagnostic; now it propagates out of `runDiagnostics`, and the debounced `updateDocument` call runs inside a bare `setTimeout` callback — an uncaught exception there takes down the LSP process. Typing `import { x } from "pkg::not-installed"` reproduces it on every keystroke.

**2. Some swallowed import errors have no `checkMissingImports` counterpart** (inline comment on `importResolver.ts`). AG4008/4009/4010 are covered, but the retry-safety-marker-on-a-node error and the `import test`-with-pkg::/TS gates are also `ImportResolutionError`s that `skipUnresolvable` drops — those now show no squiggle in the editor while the compile still hard-fails. Editor and compiler disagree for those cases.

**3. Only the active document's buffer is an override** (inline comment on `server.ts`). A dirty-but-unsaved *imported* file still resolves from disk, so the original bug survives in the cross-file shape: add an `export` in `b.agency`, switch to `a.agency`, type the import — false error until `b` is saved.

Minor, no change needed: the dynamic watcher registration can double up with clients that already watch `agency.json` (VS Code extension config), causing redundant rebuild-all passes; harmless since `updateDocument` is idempotent.

packages/agency-lang/lib/lsp/diagnostics.ts
egonSchiele: Confirmed and fixed. `resolveAgencyImportPath` throws a plain `Error` for an uninstalled `pkg::`, and since the call was outside a try/catch and the skip path only absorbed `ImportResolutionError`, it could escape into the debounce callback and crash the server. Replaced `skipUnresolvable` with an `onUnresolvable` callback that coerces any thrown value (including plain `Error`) to `ImportResolutionError` and reports it instead of throwing. Also wrapped the LSP call so any unexpected throw falls back to the unrewritten program, and made `SymbolTable.build`'s crawl skip an unresolvable import path rather than aborting the whole build. (2802a417e)

packages/agency-lang/lib/preprocessors/importResolver.ts
egonSchiele: Correct — `checkMissingImports` only covers AG4008/4009/4010, so the marker-on-node and `import test` pkg/TS gates were being swallowed silently. Reworked the skip path to drop bad names and report each via `onUnresolvable` at its `loc`, so these structural-misuse errors now surface in the editor. Because bad names are dropped from the rewrite, the checker no longer double-reports the covered cases. Added tests for both categories. (2802a417e)

packages/agency-lang/lib/lsp/server.ts
egonSchiele: Fixed now rather than as a follow-up — `updateDocument` builds the override map from every open document (`documents.all()`), so a dirty imported file resolves from its buffer too. (2802a417e)

packages/agency-lang/lib/preprocessors/importResolver.ts
```diff
@@ -48,157 +67,241 @@ export function resolveImports(
...
+    typeNames: [],
+    constantNames: [],
+  };
+  const aliases: Record<string, string> = {};
+
+  for (const nameType of node.importedNames) {
+    if (nameType.type !== "namedImport") {
+      // Namespace or default imports of .agency files — keep as-is
+      out.push(node);
+      continue;
```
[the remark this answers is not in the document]
egonSchiele: Fixed. A statement with any non-named specifier (namespace/default, including mixed `import foo, { bar }`) now returns `[node]` unchanged instead of pushing the original and continuing to emit rewritten named imports. (2802a417e)

packages/agency-lang/lib/preprocessors/importResolver.ts
```diff
@@ -48,157 +67,241 @@ export function resolveImports(
...
-            constantNames.push(name);
-            break;
-        }
-      }
-    }
+  const importedFilePath = resolveAgencyImportPath(
+    node.modulePath,
+    currentFile,
+  );
+  const fileSymbols = symbolTable.getFile(importedFilePath) ?? {};
```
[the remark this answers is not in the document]
egonSchiele: Fixed — path resolution failures are now coerced to `ImportResolutionError` (with the statement's `loc`) inside `resolveImportStatement`'s statement-level try/catch, so `onUnresolvable` keeps them skippable and anchored to the import line. (2802a417e)

packages/agency-lang/lib/lsp/diagnostics.ts
```diff
@@ -135,6 +131,19 @@ export function runDiagnostics(
...
+    allowTestImports: true,
+    skipUnresolvable: true,
+  });
```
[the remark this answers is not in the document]
egonSchiele: Guarded. The `resolveImports` call is now wrapped in try/catch; `onUnresolvable` absorbs every expected import failure, and the catch is a defensive fallback that reports the error but keeps the unrewritten `program` so the type checker still runs and the server can't crash. (2802a417e)

PR #636, opened 2026-07-21
Shard the agency test suite in CI to cut wall-clock time

egonSchiele:
## What and why

The agency execution suite (~933 files, ~11 min) is the critical path of every PR. Today it also runs **twice** — once each on node 22.x and 23.x, fully in parallel — which doubles runner cost and puts whichever runner is slower on the critical path. Measured on a recent run, `agency-tests (23.x)` gated the whole PR at ~13.5 min, of which ~11.3 min was the execution suite alone.

This shards that suite so it runs in parallel slices instead of one long job.

## The `--shard i/N` flag

Adds `--shard i/N` to `agency test` and `agency test js`. Each shard sorts the collected files and keeps a deterministic round-robin slice (`fileIndex % N == i-1`), applied **before** precompile so a shard only compiles and runs its own slice — that is what makes sharding cut wall-clock rather than just splitting the run phase.

The split is **exhaustive and disjoint**: every file has exactly one residue mod N, so it lands in exactly one shard. No file is dropped or run twice, even when the count does not divide evenly (13 files / 4 shards → 4,3,3,3). A new unit test asserts this for awkward ratios; an end-to-end run over the 13-file `function-refs/` dir confirmed the union across all four shards is exactly the 13 unique files.

## CI changes

**`test.yml`**
- `agency-tests` becomes a 4-shard matrix on **node 22 only**. Running the full suite on two node versions was the doubled cost; the cheaper `build` job still exercises 23.x (unit tests + build), so 23.x is not left uncovered.
- Shards collect step coverage via the `AGENCY_COVERAGE` env vars (not `--coverage`, which would also generate a per-shard report that walks *all* source files) and upload a per-shard artifact.
- New **`coverage-report`** job merges the shard artifacts into one `.coverage/` and posts the PR comment. `loadCoverageData` already unions every `cov-*.json`, and the files carry a UUID, so "merging" is just gathering them into one dir — no new merge logic.
- The push-to-main `github stdlib smoke test` is gated to `shard == 1` so it runs once.

**`test-with-llm.yml`**
- The real-LLM agency/agency-js suites are sharded the same way; the once-only agent/efficacy steps move to their own `real-llm-extras` job. Note this raises peak OpenAI concurrency (SHARD_COUNT × `-p`); if it trips rate limits, lower `-p` or the shard count (commented in the workflow).

## Verification

- `tsc` clean, structural linter clean, 8/8 new unit tests pass.
- End-to-end: `--shard` split of `function-refs/` (13 files) → 4,3,3,3, union = 13 unique files (no overlap, none missing).
- End-to-end coverage: two shards collected via env vars into one dir (3+3=6 files, no collision), and `coverage report` produced a coherent merged report.

## Note for review

Node 23.x is dropped from the agency suite entirely (PR and main), not just PRs. The `build` job still covers 23.x for unit tests + build. If you want 23.x agency coverage on a nightly, that is a small follow-up.

### DIFF
--- a/.github/workflows/test-with-llm.yml
+++ b/.github/workflows/test-with-llm.yml
@@ -17,11 +17,19 @@ permissions:
   contents: read
 
 jobs:
+  # The agency + agency-js suites are sharded (same round-robin split as the
+  # deterministic PR workflow) so the post-merge real-LLM run finishes faster.
+  # Each shard runs its own slice against the real provider. Note this raises
+  # peak concurrency against the OpenAI API (SHARD_COUNT × -p), so if these
+  # start hitting rate limits, lower `-p` or SHARD_COUNT.
   real-llm-tests:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        node-version: [22.x]
+        shard: [1, 2, 3, 4]
+    env:
+      SHARD_COUNT: "4"
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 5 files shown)


egonSchiele:
Pushed fixes for the missing coverage comment and the security note.

**Why no coverage comment appeared (now fixed).** The shards collected coverage by setting `AGENCY_COVERAGE`/`AGENCY_COVERAGE_OUTDIR` as step env vars rather than the `--coverage` flag. That assumed the spawned test subprocesses inherit those env vars — which they did locally but **not** on CI, so every shard's `.coverage/` was empty, the merge downloaded 0 artifacts, and the report printed "No coverage data found". I traced it to the empty upload in the shard logs.

Fix: the shards now use the proven `--coverage` flag (the CLI sets the env vars itself, exactly as the old single-job CI did — those runs reliably produced coverage comments), plus a new `--collect-only` flag that skips the per-shard report. `--collect-only` exists because `--coverage` on its own would recompile *every* source file for its source map on each shard — the waste the env-var approach was trying to dodge. Verified end-to-end locally: two shards → merge → `coverage report stdlib` renders real numbers.

**Security note — implemented the split you described.** `coverage-report` now holds **no** write token and does all the untrusted work (`pnpm install`, `make`, `coverage report` — the .agency recompile and parsing of shard-produced coverage JSON), uploading only two plain-text files. A new minimal `coverage-comment` job holds `pull-requests: write` and only downloads that text and posts the comment — no install, no build, no recompile, no parsing of PR-controlled JSON. The one residual (pre-existing) is the comment action's `render.js` running under the token; eliminating that entirely would mean inlining a pinned script, noted in a comment.

**Re the two `!cancelled()` review notes.** `!cancelled()` does contain a status-check function, so it already overrides the default "skip on prior failure / skip when a needed job failed" — it's not equivalent to a bare custom condition. But to remove any ambiguity and match the idiom already used elsewhere in this workflow, I switched the coverage step and jobs to the explicit `always() && !cancelled()`.

egonSchiele:
Correction on the root cause (my earlier "env inheritance" explanation was wrong).

I instrumented the shard on CI and got definitive counts:

- after the agency step: **360** `cov-*.json` in `.coverage/`
- after the agency-js step: **396** (agency-js accumulates fine — no wipe)
- the upload step: **"No files were found with the provided path: packages/agency-lang/.coverage. No artifacts will be uploaded."**

So collection worked all along; the **upload** was silently dropping everything. The cause is that `.coverage` is a dot-directory and `actions/upload-artifact@v4` excludes hidden files by default (since v4.4). Fix is one line: `include-hidden-files: true` on the shard-coverage upload. This also explains why every other suspect (the flags, `--collect-only`, `--shard`, env inheritance, the agency-js step) checked out — none of them were the problem.

Pushed that fix and removed the temporary diagnostic steps. Watching the next run to confirm the coverage comment actually posts this time before I call it done.

PR #638, opened 2026-07-21
Weave nested comments and add --base-url to literate

egonSchiele:
## What

Two fixes to `agency literate weave`.

### 1. Weave multi-line comments nested inside blocks

Previously the classifier only inspected the program's **top-level** nodes, so only a module-level `/* */` comment turned into prose. Every comment written inside a `node` or `def` body stayed embedded as a code comment, because the whole block was handed to the generator as one opaque code node.

Now a post-process pass scans each rendered code segment and lifts every multi-line comment (`/* */` and `/** */`) out as prose, splitting the code fence around it. This works at **any nesting depth** since it operates on the rendered string. Details:

- **String literals are skipped** (via tarsec's `simpleStringParser`) so a `/*` inside a string is not mistaken for a comment.
- **Single-line `//` comments stay in the code** — the convention being that block comments narrate and inline comments annotate.

The reopened code fences are fragments (an opening `{` can close several fences later), which is normal for "woven" literate output meant for reading rather than standalone compilation.

### 2. `--base-url` for a "View source" link

Mirrors the `doc` command's option. When given, each woven file gets a `[View source](...)` link as its first line pointing at the source file. Directory input uses the tree-relative, posix-normalized path; a trailing slash on the base URL is stripped. Without the flag, output is unchanged.

```
agency literate weave examples/ --base-url https://github.com/org/repo/blob/main/examples
```

## Notes

- `--base-url` is **CLI-only**; unlike `doc` it does not read from `agency.json` (literate has no config section today). Easy to add later if wanted.

## Testing

`lib/cli/literate.test.ts`: 23 tests pass (16 existing + 7 new covering nested-comment weaving, any-depth weaving, the `/*`-in-a-string guard, and the four `--base-url` behaviors). Structural linter clean.

### DIFF
--- a/packages/agency-lang/lib/cli/literate.ts
+++ b/packages/agency-lang/lib/cli/literate.ts
@@ -4,10 +4,7 @@ import { parse, readFile } from "./commands.js";
 import { findRecursively } from "./util.js";
 import { AgencyNode, AgencyProgram, AgencyMultiLineComment } from "@/types.js";
 import { codeFence } from "@/utils/markdown.js";
-import {
-  multiLineCommentParser,
-  simpleStringParser,
-} from "@/parsers/parsers.js";
+import { multiLineCommentParser } from "@/parsers/parsers.js";
 import * as fs from "fs";
 import * as path from "path";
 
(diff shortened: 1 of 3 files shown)


packages/agency-lang/lib/cli/literate.ts
```diff
@@ -111,6 +115,58 @@ function classify(program: AgencyProgram): Segment[] {
...
+ */
+function weaveComments(code: string, lang: string): string {
+  const parts: string[] = [];
+  let codeBuf = "";
+  let rest = code;
+
+  const flushCode = (): void => {
+    // Drop leading blank lines left behind when a comment on its own line is
+    // lifted out, so the reopened fence does not start with an empty line.
+    const trimmed = codeBuf.replace(/^(?:[ \t]*\n)+/, "");
```
egonSchiele, line 148: why is this regex needed?

packages/agency-lang/lib/cli/literate.ts
```diff
@@ -111,6 +115,58 @@ function classify(program: AgencyProgram): Segment[] {
...
+    if (trimmed.trim() !== "") parts.push(codeFence(trimmed, lang));
+    codeBuf = "";
+  };
+
+  while (rest.length > 0) {
```
egonSchiele: this seems not very performant. Did you not have a better choice than parsing character by character?

packages/agency-lang/lib/cli/literate.ts
egonSchiele: That regex is the leading-blank-line strip in `flushCode`. When a comment sits on its own line, the newline that ends the *previous* line belongs to the code run that follows the comment. So after the comment is lifted out, that following run starts with a blank line, and the reopened ```` ```agency ```` fence would render with an empty first line. The regex drops those leading blank lines while keeping the first real line's indentation. I've expanded the comment above it to explain this. (The char-by-char scan it lived in is gone now — see the rewrite — but this bit is still needed.)

packages/agency-lang/lib/cli/literate.ts
egonSchiele: Fixed in 74aef65. Replaced the char-by-char scan with a single left-to-right regex pass that matches a string literal, a `//` line comment, or a `/* */` block comment (whichever comes first) and acts only on block comments. No more per-character slicing, and it also correctly skips a `/*` that appears inside a `//` line comment, which the old version would have woven out.

packages/agency-lang/lib/cli/literate.ts
```diff
@@ -111,6 +115,58 @@ function classify(program: AgencyProgram): Segment[] {
...
+      // the surrounding code into two fences.
+      if (prose !== "") {
+        flushCode();
+        parts.push(prose);
+      }
+      continue;
+    }
+    codeBuf += rest[0];
+    rest = rest.slice(1);
+  }
```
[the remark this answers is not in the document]
egonSchiele: Good catch — fixed in 74aef65. The rewrite no longer relies on the parser's `optionalSpaces`: whitespace before a comment now stays in the preceding code run, so an empty `/**/` drops only the comment token and leaves the surrounding indentation intact. Added a regression test (`drops an empty block comment without eating surrounding whitespace`).

PR #641, opened 2026-07-22
feat(thread): add toolMessage to seed a synthetic tool exchange

egonSchiele:
## What

Adds `toolMessage(name, args, result, label="")` to `std::thread` — a `userMessage`-style function that seeds a synthetic tool call plus its result into the active thread, as if the model had made the call. Nothing executes; it only shapes the conversation the model reads on its next `llm()`.

```
toolMessage("saveDraft", { value: summary }, "Draft saved.")
// thread now ends with:
//   assistant  tool_call(id, name=saveDraft, arguments={value:…})
//   tool       tool_call_id=id  content="Draft saved."
```

## Why

A scaffold often needs to make the model *see* work it did on the model's behalf — for example, deterministically checkpointing a partial result at a budget threshold rather than hoping the model calls `saveDraft` itself. This is the injection building block for that. It is deliberately small: it does not run anything and knows nothing about budgets or triggers, which are separate pieces to build on top.

## Design

- Runtime `_toolMessage` in `lib/stdlib/thread.ts`, beside `_userMessage`, reusing `getRuntimeContext()` / `getOrCreateActive()` / `thread.push(msg, label || null)` — no new registry or buffer.
- `args` is validated by a JSON round-trip up front, so a non-serializable value throws at the call site instead of being rejected by the provider on the next `llm()`. Stored as a `Record` (smoltalk's `ToolCallJSON.arguments` is a record, not a string).
- The id is synthetic (`nanoid`); real tool-call ids come from the provider.
- `label` rides on both pushed messages, exactly like `_userMessage`. No `onToolCallStart`/`onToolCallEnd` — those are for live executions.

## Scope

- `result` is a plain `string` in v1. Image/file attachments are deferred: the loop's attachment-injection machinery is welded to the live `llm()` call (`runnerState` buffer, minted ids, end-of-round drain) and can't be reused from a thread-seeding primitive. Widening the type later is backward-compatible.

## Tests

- `lib/stdlib/toolMessage.test.ts` (6): exact two-message pair, shared id, `arguments` stored as an object, `name` field, empty assistant content, `label` on both messages, unlabeled default, null-args default, active-thread-created-when-none, and a throw-and-push-nothing case for non-serializable args.
- `tests/agency/thread/toolmessage-roundtrip.agency`: seeds an exchange and reads it back via `getThread`, asserting `assistant|tool:Draft saved.` (catches a dropped assistant message).
- `lib/runtime/toolMessageProvider.test.ts`: proves the seeded exchange is forwarded into the outgoing `PromptConfig`, id-matched. A mock can't prove a real provider accepts an un-redeclared historical tool call; that rests on the documented provider contract and is a CI/manual follow-up.

## Note

`smoltalk.assistantMessage` requires a `ToolCall` **instance** for `toolCalls` (its `toJSON` walks each call), not a plain object — the one implementation detail that differed from the design sketch. Output and all assertions are unchanged.

### DIFF
--- a/packages/agency-lang/lib/stdlib/thread.ts
+++ b/packages/agency-lang/lib/stdlib/thread.ts
@@ -1,4 +1,5 @@
 import * as smoltalk from "smoltalk";
+import { nanoid } from "nanoid";
 import { agencyStore, getRuntimeContext } from "../runtime/asyncContext.js";
 import type { ReplyAttachmentPart } from "../runtime/replyAttachments.js";
 import { __tryCall, type ResultValue } from "../runtime/result.js";
@@ -69,6 +70,48 @@ export async function _userMessage(
 ): Promise<void> {
   const { threads } = getRuntimeContext();
   threads.getOrCreateActive().push(smoltalk.userMessage(msg), label || null);
+}
+
+/** Seed a synthetic tool call and its result onto the active thread, as if
+ *  the model had made the call. Nothing executes: this only shapes the
+ *  conversation. `args` is validated by a JSON round-trip up front, so a
+ *  non-serializable value throws here rather than being rejected by the
+ *  provider on the next `llm()`. Both messages carry `label`, an
+ *  observability-only tag (see `_userMessage`). The id is synthetic — real
+ *  tool-call ids come from the provider, this one is minted with nanoid. */
+export async function _toolMessage(
+  name: string,
+  args: any,
+  result: string,
+  label: string = "",
+): Promise<void> {
+  // Round-trip up front so a non-serializable value throws HERE, not far away
+  // when the provider rejects the thread on the next llm(). One catch covers
+  // both failure shapes: JSON.stringify throws on a circular object, and
+  // returns undefined for a bare function (which then fails in JSON.parse).
+  let argsRecord: Record<string, any>;
+  try {
+    argsRecord = JSON.parse(JSON.stringify(args ?? {}));
+  } catch {
+    throw new Error(
+      `toolMessage: args for "${name}" could not be serialized to JSON`,
+    );
+  }
+
+  const id = nanoid();
+  const { threads } = getRuntimeContext();
+  const thread = threads.getOrCreateActive();
+
+  thread.push(
+    smoltalk.assistantMessage("", {
+      toolCalls: [new smoltalk.ToolCall(id, name, argsRecord)],
+    }),
+    label || null,
+  );
+  thread.push(
+    smoltalk.toolMessage(result, { tool_call_id: id, name }),
+    label || null,
+  );
 }
 
 export async function __internal_assistantMessage(
--- a/packages/agency-lang/lib/stdlib/thread.ts
+++ b/packages/agency-lang/lib/stdlib/thread.ts
@@ -92,9 +92,25 @@ export async function _toolMessage(
   let argsRecord: Record<string, any>;
   try {
     argsRecord = JSON.parse(JSON.stringify(args ?? {}));
-  } catch {
+  } catch (e) {
     throw new Error(
       `toolMessage: args for "${name}" could not be serialized to JSON`,
+      { cause: e },
+    );
+  }
+  // Arguments must be a JSON object (a record), which every real tool call is.
+  // A bare number, string, or array round-trips cleanly but violates that
+  // contract (consumers like dropNullDefaultedArgs expect a record) and a
+  // provider may reject the thread. Enforce it here so the docstring's "as an
+  // object" is real.
+  if (
+    typeof argsRecord !== "object" ||
+    argsRecord === null ||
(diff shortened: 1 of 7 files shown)


egonSchiele (commented):
## Overview

Adds `toolMessage(name, args, result, label="")` to `std::thread`: a `userMessage`-style primitive that seeds a synthetic assistant tool-call + matching tool-result pair into the active thread, without executing anything. Clean, well-scoped, and the design reasoning in the PR description is exactly right.

## Assessment: looks good

- **Right altitude.** `_toolMessage` sits beside `_userMessage` and reuses `getRuntimeContext()` / `getOrCreateActive()` / `push(msg, label || null)` — no new registry, buffer, or duplicated semantics. It also correctly skips a dead `__internal_toolMessage` twin (the existing `__internal_*Message` functions have no live references).
- **`arguments` stored as a record**, matching `ToolCallJSON.arguments` (a `z.ZodRecord`, not a string) — the design sketch had this wrong and the PR fixed it. The `new smoltalk.ToolCall(...)` instance (vs a plain object) is the correct call given `assistantMessage.toJSON()` walks each entry; good catch, and the note calling it out is appreciated.
- **The pair is pushed atomically.** The two `thread.push` calls are synchronous with no `await` between them, so no interrupt/checkpoint can interleave and leave a half-pair on the thread. Worth stating explicitly since a lone tool-call or lone tool-result is a provider protocol error.
- **Validation re-throws with context**, it doesn't swallow — the one catch covers both the circular case (`JSON.stringify` throws) and the bare-function case (`stringify` returns `undefined`, `JSON.parse` then throws). The comment explaining both paths is good.
- **Honest test framing.** The provider test is named "forwarding + wire shape" and the PR description is upfront that a mock can't prove real provider acceptance of an un-redeclared historical tool call. That's the correct claim to make; leaving real acceptance to the documented contract + a CI/manual follow-up is reasonable.

## Test coverage

Strong. The suite pins the behaviors that would actually break: exact two-message count, shared id, `arguments` as an object, the tool-result `name` field, empty assistant content, `label` on both messages, the unlabeled-default (`label || null`) branch, null-args defaulting, the create-when-no-active-thread branch (via a bare `new ThreadStore()`), and throw-and-push-nothing with a `/could not be serialized/` matcher. The Agency roundtrip now encodes `assistant|tool` so a dropped assistant message is caught. Every gap from the plan review was closed.

## Minor (non-blocking)

- Non-object `args` flow through unguarded — see inline note. Low priority.
- The `catch` discards the original error; passing `{ cause: e }` to the re-thrown `Error` would preserve the underlying reason for debugging. Nit.

No correctness bugs, convention violations, or security concerns found. Nice, tight PR.

packages/agency-lang/lib/stdlib/thread.ts
```diff
@@ -71,6 +72,48 @@ export async function _userMessage(
...
+  result: string,
+  label: string = "",
+): Promise<void> {
+  // Round-trip up front so a non-serializable value throws HERE, not far away
+  // when the provider rejects the thread on the next llm(). One catch covers
+  // both failure shapes: JSON.stringify throws on a circular object, and
+  // returns undefined for a bare function (which then fails in JSON.parse).
+  let argsRecord: Record<string, any>;
+  try {
+    argsRecord = JSON.parse(JSON.stringify(args ?? {}));
```
egonSchiele, line 94: Nit (low priority): `argsRecord` is typed `Record<string, any>`, but `JSON.parse` returns `any`, so a non-object `args` slips through unguarded. `toolMessage("t", "raw", ...)` round-trips to the string `"raw"`, which `new ToolCall(id, name, "raw")` then treats as a pre-stringified arguments *string* (the constructor accepts `Record<string, any> | string`); a number/array is stranger still. The docstring says "as an object," so a guard like `if (typeof argsRecord !== "object" || argsRecord === null || Array.isArray(argsRecord)) throw ...` would make that contract real. Fine to leave for the intended object-args use.

egonSchiele:
Thanks both — addressed in 28e59e1.

**Non-object `args` (flagged by both reviews):** `_toolMessage` now guards after the JSON round-trip and throws `must be a JSON object, got <type>` for a number, string, or array, so the "arguments is a record" contract (the one `dropNullDefaultedArgs` and providers rely on) is enforced rather than assumed. Added a test covering `5`, `"raw"`, and `[1,2]`. `null`/`undefined` still default to `{}`, unchanged.

**Error cause (nit):** the serialization `catch` now re-throws with `{ cause: e }`, so the underlying `TypeError`/`SyntaxError` survives for debugging.

**Agency `args: any` type:** left as `any` deliberately — Agency has no precise "JSON object" type, and callers pass heterogeneous object literals. The runtime guard above now makes the contract real regardless of the static type, and the docstring says "as an object." Happy to tighten if you'd prefer a specific record type despite the ergonomic cost.

PR #643, opened 2026-07-22
Add agency lint with an unused-imports rule (AL0001), editor gray-out, and remove-on-save

egonSchiele:
## What this adds

A linter subsystem (`lib/linter/`) with one rule, surfaced three ways:

- **`agency lint <files|dirs>`** prints findings with stable `AL####` codes:

  ```
  main.agency
    1:10  AL0001  'now' is imported but never used.

  Run `agency explain <code>` for details.
  ```

- **VS Code grays out unused imports** (Hint severity + the `Unnecessary` tag) and offers a "Remove unused import" quick fix, a "Remove all unused imports" batch action, and remove-on-save via `source.fixAll` or the dedicated `source.removeUnusedImports` kind.

- **`agency explain AL0001`** falls through from the type checker registry to the lint registry.

The first rule, `unused-imports` (AL0001), covers named imports (`import { a, b }`) and `import node { x }`. It is deliberately **conservative**: an imported name that appears anywhere in the file counts as used, even if a shadowing local is the real target. A missed detection costs nothing; a false positive would delete a needed import, so the rule is built to never produce one.

## Design notes

- The linter is a consumer of a parsed program, not a new analysis: rules are pure functions from `{ program, source, filePath }` to findings-as-data, and both the CLI and the LSP run the identical rules on the identical parse mode (template off, so offsets index the raw file).
- The reference walk collects `variableName`, `functionCall`, `typeAliasVariable`, and `genericType` nodes. The `typeAliasVariable` arm matters: a bare annotation like `x: Config` parses as `typeAliasVariable`, not `genericType`, and missing it would falsely flag type-only imports.
- Removal fixes **regenerate the whole import statement** from the AST via the `AgencyGenerator` instead of splicing name tokens, which handles aliases (`a as b`), `destructive`/`idempotent` markers, and trailing commas for free. The batch action regenerates each statement once with all of its unused names removed, so it never emits overlapping edits.
- The injected `std::index` prelude import and `import test { }` imports are skipped entirely.
- Lint codes are append-only from day one (`AL0001`, sequential numbering, category-free by design — documented divergence from the type checker's prefix-range scheme).
- `agency lint` always exits 0 for now: the only rule is a hint, and hints should not fail CI. The PR that adds the first warning-severity rule adds the non-zero path.

## One parser change

Import statements previously carried no `loc` at all, so there was nothing to anchor a finding or a fix to. `importStatmentParser` and `importNodeStatmentParser` are now wrapped in `withLoc`, like most other statement parsers. Fallout was small and is included:

- The `parseAgency` loc.line invariant tests now skip template-injected nodes (the prelude import sits above line 0 of the user source, so its `loc.line` is negative by construction).
- One preprocessor fixture (`tests/typescriptPreprocessor/tools.json`) regenerated to include the new `loc` field.

## Testing

- 39 new tests across the linter (registry integrity, 13 detection cases, 7 fix cases, entry points), CLI formatting, `explain` fallthrough, LSP diagnostics mapping, and LSP code actions (including the two-unused-names-in-one-statement batch case).
- Full parser/preprocessor sweep (1864 tests) green after the `withLoc` change; structural linter clean; `agency lint` smoke-tested by hand.

Spec and reviews: `docs/superpowers/specs/2026-07-21-agency-lint-command-design.md` and siblings (in the worktree-lint-command worktree, not part of this PR).

### DIFF
--- a/packages/agency-lang/lib/linter/rules/unusedImports.ts
+++ b/packages/agency-lang/lib/linter/rules/unusedImports.ts
@@ -1,0 +1,158 @@
+import type { SourceLocation } from "../../types/base.js";
+import type { AgencyProgram } from "../../types.js";
+import type {
+  ImportStatement,
+  ImportNodeStatement,
+} from "../../types/importStatement.js";
+import { getImportedNames } from "../../types/importStatement.js";
+import type { LintContext, LintFinding, LintRule } from "../types.js";
+import { lintDiagnostic } from "../diagnostics.js";
+
+/** Names referenced anywhere in the file body. Conservative: an imported name
+ *  found here is treated as used, so the rule never removes a used import.
+ *  The reference positions and their node shapes (verified via pnpm run ast):
+ *    - value reference / access-chain base → `variableName` (.value)
+ *    - direct call                         → `functionCall` (.functionName)
+ *    - bare type annotation                → `typeAliasVariable` (.aliasName)
+ *    - parameterized type                  → `genericType` (.name)
+ *  This is a generic structural walk, NOT walkNodes: walkNodes visits
+ *  statement-level AgencyNodes and does not descend into type hints or tag
+ *  arguments, which are exactly the positions that would cause a false
+ *  "unused" (and a deleted needed import) if missed. Import statements are
+ *  excluded so an import does not count as its own use. */
+export function collectReferencedNames(program: AgencyProgram): Record<string, true> {
+  const used: Record<string, true> = {};
+  const visit = (value: unknown): void => {
+    if (Array.isArray(value)) {
+      for (const item of value) {
+        visit(item);
+      }
+      return;
+    }
+    if (value && typeof value === "object") {
+      const node = value as Record<string, unknown>;
+      if (node.type === "variableName" && typeof node.value === "string") {
+        used[node.value] = true;
+      } else if (node.type === "functionCall" && typeof node.functionName === "string") {
+        used[node.functionName] = true;
+      } else if (node.type === "typeAliasVariable" && typeof node.aliasName === "string") {
+        used[node.aliasName] = true;
+      } else if (node.type === "genericType" && typeof node.name === "string") {
+        used[node.name] = true;
+      }
+      for (const key of Object.keys(node)) {
+        visit(node[key]);
+      }
+    }
+  };
+  for (const node of program.nodes) {
+    if (node.type === "importStatement" || node.type === "importNodeStatement") {
+      continue;
+    }
+    visit(node);
+  }
+  return used;
+}
+
+/** 0-indexed line/col plus offsets, matching the codebase's SourceLocation. */
+export function locFromOffsets(source: string, start: number, end: number): SourceLocation {
+  let line = 0;
+  let lastNewline = -1;
+  for (let i = 0; i < start; i++) {
+    if (source[i] === "\n") {
+      line++;
+      lastNewline = i;
+    }
+  }
+  return { line, col: start - lastNewline - 1, start, end };
+}
+
+const escapeForRegex = (s: string): string => s.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+
+/** The local name's character range within a statement's source span, matched
+ *  on word boundaries so `b` never matches inside `ab` or inside the module
+ *  path. Falls back to the whole statement span if the token is somehow not
+ *  found (dimming the whole statement is honest; pointing at the wrong
+ *  character is not). */
+function nameRange(
+  source: string,
+  stmtStart: number,
+  stmtEnd: number,
+  localName: string,
+): SourceLocation {
+  const span = source.slice(stmtStart, stmtEnd);
+  const match = new RegExp(
+    `(?<![A-Za-z0-9_])${escapeForRegex(localName)}(?![A-Za-z0-9_])`,
+  ).exec(span);
+  if (!match) {
+    return locFromOffsets(source, stmtStart, stmtEnd);
+  }
+  const start = stmtStart + match.index;
+  return locFromOffsets(source, start, start + localName.length);
+}
+
+/** The statement's span with trailing whitespace trimmed off: the parser's
+ *  span includes the optional trailing newline it consumed, which is not part
+ *  of the statement text we match names in or replace on regeneration. */
+function statementSpan(
+  source: string,
+  stmt: ImportStatement | ImportNodeStatement,
+): { start: number; end: number } {
+  const loc = stmt.loc ?? { line: 0, col: 0, start: 0, end: 0 };
+  let end = loc.end;
+  while (end > loc.start && /\s/.test(source[end - 1])) {
+    end--;
+  }
+  return { start: loc.start, end };
+}
+
+function findingFor(
+  ctx: LintContext,
+  stmt: ImportStatement | ImportNodeStatement,
+  localName: string,
+): LintFinding {
+  const span = statementSpan(ctx.source, stmt);
+  const range = nameRange(ctx.source, span.start, span.end, localName);
+  return lintDiagnostic("unusedImport", { name: localName }, range);
+}
+
+/** Each lintable statement paired with its unused local names. Shared by the
+ *  per-name findings (below) and the grouped batch fix. */
+function unusedByStatement(
+  ctx: LintContext,
+): { stmt: ImportStatement | ImportNodeStatement; unusedNames: string[] }[] {
+  const used = collectReferencedNames(ctx.program);
+  const groups: { stmt: ImportStatement | ImportNodeStatement; unusedNames: string[] }[] = [];
+  for (const node of ctx.program.nodes) {
+    if (node.type === "importStatement") {
+      if (node.modulePath === "std::index") {
+        continue; // injected prelude
+      }
+      if (node.testOnly) {
+        continue; // test-only wiring
+      }
+      const unusedNames = node.importedNames
+        .filter((nameType) => nameType.type === "namedImport") // namespace/default deferred
+        .flatMap((nameType) => getImportedNames(nameType))
+        .filter((localName) => !used[localName]);
+      if (unusedNames.length > 0) {
+        groups.push({ stmt: node, unusedNames });
+      }
+    } else if (node.type === "importNodeStatement") {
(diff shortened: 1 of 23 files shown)


packages/agency-lang/lib/linter/rules/unusedImports.ts
```diff
@@ -0,0 +1,252 @@
...
+      const node = value as Record<string, unknown>;
+      if (node.type === "variableName" && typeof node.value === "string") {
+        used[node.value] = true;
+      } else if (node.type === "functionCall" && typeof node.functionName === "string") {
+        used[node.functionName] = true;
+      } else if (node.type === "typeAliasVariable" && typeof node.aliasName === "string") {
+        used[node.aliasName] = true;
+      } else if (node.type === "genericType" && typeof node.name === "string") {
+        used[node.name] = true;
+      }
```
egonSchiele, line 60: **CRITICAL — this allow-list walk misses named-handler references, and the autofix then deletes them.**

`HandleBlock.handler` with `kind: "functionRef"` is `{ kind: "functionRef", functionName }` (`lib/types/handleBlock.ts:9`) — it has **no `type` field**, so `node.type === "functionCall"` never matches and the handler's `functionName` is never collected. An imported function used only as a named handler:

```
import { handleInterrupt } from "./handlers.agency"
node main() {
  ...
} with handleInterrupt
```

is therefore reported as `AL0001` unused, and remove-on-save deletes the import — silently unregistering the handler. CLAUDE.md is explicit that handlers must never be left unregistered; an autofix that removes them is exactly that failure. (`goto foo()` is safe because its `nodeCall` is a real `FunctionCall`; `functionRef` is the gap.)

The deeper issue: the PR's central claim is that the rule can never produce a false positive. That only holds if this allow-list of four `type`s is provably complete — and it isn't. `functionRef` is one counterexample; any name carried under a different discriminant (`kind`) or a bare string field is another (audit `raises` effect refs, tag-argument constants, `provide`, type patterns). Two ways out:

1. Make the walk conservative *by construction*: collect `functionName` whenever it is a string regardless of `type`/`kind` (over-collection only keeps imports, so it is safe), and likewise other name-bearing fields.
2. Keep the allow-list but enumerate every reference construct and add a test per construct proving the used name is kept.

Either way, add used-name tests for the at-risk positions: there is currently **no test that a *used* import node or a *used* handler function is left alone**, which are precisely the positions that would be wrongly deleted.

packages/agency-lang/lib/lsp/codeAction.ts
```diff
@@ -44,6 +60,34 @@ export function getCodeActions(
     if (stdlibAction) actions.push(stdlibAction);
   }
 
+  const source = doc.getText();
+  const parsed = parseAgency(source, {}, false);
```
egonSchiele, line 78: **Minor — re-parses and re-lints on every code-action request.** This parses the whole document and runs the linter each time actions are requested (cursor moves, lightbulb), and `runDiagnostics` already parses + lints the same document on every keystroke. For large files that is real duplicated work — consider computing findings once (off the cached `DocumentState`) and sharing them between the diagnostics and code-action paths.

Also, `getCodeActions` ignores `params.context.only`. VS Code sends `only: ["source.fixAll"]` (etc.) on save and for filtered menus; returning every action regardless usually works because the client filters, but honoring `only` avoids computing actions the client will immediately discard.

egonSchiele (commented):
Reviewed against the spec and plan. The shape is right — a separate linter consuming a parsed program, diagnostic codes from day one, regeneration-based fixes, and template-off parsing so offsets index the raw file — and the code is clean and well-tested for the cases it covers. Verifying the claims: `NodeCall` (with `nodeName`) is dead code so `goto`-referenced nodes are safe via the nested `FunctionCall`, and `raises` clauses are `VariableType`s so effect refs are collected via `genericType`/`typeAliasVariable`.

**One blocker** (inline on `collectReferencedNames`): the reference walk misses `kind: "functionRef"` handler references, so an imported function used only as `} with handlerFn` is flagged unused and auto-removed on save — silently unregistering a handler (CLAUDE.md CRITICAL). This also undercuts the PR's safety claim that the rule can never produce a false positive; an allow-list of node types is only safe if it is provably complete. Broaden the walk (or enumerate every reference construct) and add used-name tests for the at-risk positions.

Two smaller things: the code-action re-parse / `context.only` note inline; and, low-priority, wrapping the import parsers in `withLoc` now gives the injected `std::index` prelude import a negative `loc.line` in the compile path (you relaxed the `parseAgency loc.line invariant` test for it) — confirm nothing else consumes an import's `loc.line` in the template path where a negative value would surprise it, and consider running `make fixtures` to be sure no fixture other than `tools.json` carries an import statement that now needs the `loc` field.

(Posting as a comment, not a formal request-changes, since GitHub blocks that on one's own PR — but treat the handler miss as a blocker.)

packages/agency-lang/lib/linter/rules/unusedImports.ts
egonSchiele: Fixed in 90b3706c7 — you were right that the allow-list was unprovable, so the walk is now conservative by construction (your option 1): `functionName`, `aliasName`, and `handlerName` are collected on any object regardless of its `type`/`kind` discriminant, with a comment explaining why over-collection is the safe direction. Only `value` and `name` stay discriminated (string literals and parameters carry those fields, so collecting them blindly would gut detection quality).

Added the used-name tests you asked for: an import used only as `handle { } with fn` is kept (the exact functionRef case), and a used `import node` goto target is kept. Writing the handler test also exposed a pre-existing prototype-chain bug in `getImportedNames` itself — `aliases["constructor"]` returned the inherited `Function` constructor rather than the name — now an `Object.hasOwn` lookup, with a test (`import { constructor }` is flagged cleanly).

The PR description's "never a false positive" claim now rests on the collection strategy rather than an enumeration, which is the strongest form available short of scope resolution.

packages/agency-lang/lib/lsp/codeAction.ts
egonSchiele: `context.only` is honored in 90b3706c7 — the kind filter runs before any lint work, so an on-save `only: ["source.fixAll"]` request computes exactly the batch action and a `refactor`-kind request skips the parse+lint entirely (both tested). That removes the wasted work in the two hot cases.

Sharing findings with the diagnostics pass via cached `DocumentState` needs plumbing through the server document cache, so I left it as a follow-up rather than growing this PR — the remaining recompute is the lightbulb path only. Happy to do it here instead if you prefer.

packages/agency-lang/lib/linter/rules/unusedImports.ts
```diff
@@ -0,0 +1,252 @@
...
+ *    - direct call                         → `functionCall` (.functionName)
+ *    - bare type annotation                → `typeAliasVariable` (.aliasName)
+ *    - parameterized type                  → `genericType` (.name)
+ *  This is a generic structural walk, NOT walkNodes: walkNodes visits
+ *  statement-level AgencyNodes and does not descend into type hints or tag
+ *  arguments, which are exactly the positions that would cause a false
+ *  "unused" (and a deleted needed import) if missed. Import statements are
+ *  excluded so an import does not count as its own use. */
+export function collectReferencedNames(program: AgencyProgram): Record<string, true> {
+  const used: Record<string, true> = {};
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 90b3706c7: the used-set has a null prototype now. The same test also surfaced that `getImportedNames` had the inherited-key problem on `aliases` (a `constructor` import resolved to the `Function` constructor), fixed with `Object.hasOwn`.

packages/agency-lang/lib/linter/rules/unusedImports.ts
```diff
@@ -0,0 +1,252 @@
...
+  stmt: ImportStatement | ImportNodeStatement,
+  localNames: string[],
+): LintEdit {
+  const loc = stmt.loc ?? { line: 0, col: 0, start: 0, end: 0 };
+  const span = statementSpan(source, stmt);
+  const rebuilt = rebuildWithout(stmt, localNames);
+  if (rebuilt === null) {
+    const end = source[loc.end] === "\n" ? loc.end + 1 : loc.end;
+    return { start: loc.start, end, newText: "" };
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 90b3706c7: deletion only extends past a newline when the parser span did not already consume the statement's own (statement at EOF). Added tests for both edges — a blank line after a deleted import survives, and an EOF statement without a trailing newline deletes cleanly.

egonSchiele:
All review items addressed in 90b3706c7 (details on each thread):

- **Handler blocker**: the walk is now conservative by construction — `functionName` / `aliasName` / `handlerName` are collected regardless of the node's `type`/`kind` discriminant, so `handle { } with fn` (functionRef, no `type` field) and any future name-bearing construct under a different discriminant are kept. Tests added for the handler case and a goto-used `import node`. The handler test also flushed out a pre-existing prototype-chain bug in `getImportedNames` (`aliases["constructor"]` returned the inherited `Function` constructor) — fixed with `Object.hasOwn` plus a null-prototype used-set, and covered by an `import { constructor }` test.
- **Copilot newline nit**: whole-statement deletion no longer eats the next line's newline; blank lines between imports survive (tested, plus the EOF-without-newline edge).
- **`context.only`**: honored, with the kind filter running before any parse/lint work, so on-save `source.fixAll` requests compute only the batch action and non-matching kinds skip the work entirely (tested). Sharing findings with the diagnostics pass through the cached `DocumentState` is left as a follow-up so this PR does not grow server-cache plumbing; say the word if you want it here.

On the two top-level questions:

- **Negative `loc.line` consumers**: the only consumer of an import statement's `loc` is `ImportResolutionError` (error message positions). The injected prelude import can only produce one if `std::index` itself fails to resolve or a prelude name stops being exported — both mean a broken install, and that path previously had `loc: undefined` (generic 0:0 anchor), so the failure class is unchanged. Type-checker suppression keys off type-error locs, which never anchor on import statements.
- **Fixtures**: `regenerate-fixtures` was run over everything during Task 2; `tools.json` was the only file that changed.

Full sweep after the fixes: 1,892 tests green (linter, CLI, LSP, parsers, preprocessors, compilation unit), structural linter clean, build clean.

PR #644, opened 2026-07-22
ci: parallelize the build job and cancel superseded PR runs

egonSchiele:
## What

Speed up the `Node.js CI` workflow, which was gated by the `build` job at ~8.3 min.

Job-level timing of a recent run showed the agency execution suite was already sharded to ~3.5 min, so `build` was the critical path. Inside `build`, the heavy steps ran **serially**:

| Step | Time |
|---|---|
| `pnpm test:run` (vitest unit) | 163s |
| Built-in agent tests | 147s |
| Integration (tarball/bundlers/CLI) | 82s |

## Changes

1. **Cancel superseded PR runs** — add a `concurrency` block with `cancel-in-progress`. Pushing a new commit to a PR now cancels the previous ~8-min run instead of letting it finish. Keyed so different PRs never cancel each other; pushes to `main` fall back to the unique run id so every merge still gets a full run (its own coverage + main-only integration steps).
2. **Split the serial `build` steps into three parallel jobs** — `build` (unit tests + docs), `integration`, and `agent-tests`. Each pays its own ~1-min install+make, but they overlap, dropping the critical path from ~8.3 min to ~4 min.
3. **Run agent tests on node 22 only** — agent tool-schema wiring is not node-version-sensitive, so the 23.x pass added ~2.5 min with no extra coverage.

## Notes

- Validated with `actionlint` (clean).
- New job names `integration` and `agent-tests` are not yet in branch protection's required checks — add them there if they should block merges.


egonSchiele:
**Security review — no issues found** 🔒

I ran a focused security review of this PR's changes (CI workflow + memory/compaction, prelude, and stdlib/agent code). No HIGH or MEDIUM security vulnerabilities were introduced.

Areas checked:

- **`.github/workflows/test.yml`** — The new `concurrency` group and split `integration`/`agent-tests` jobs use only trusted GitHub context expressions (`github.workflow`, `github.event_name`, `github.sha`, `github.ref`), and only as a group-key string — none flow into a `run:` shell step, and no untrusted `github.event.*` fields (PR title/body) are interpolated. Permissions stay `contents: read`. No injection surface.
- **`lib/runtime/memory/compaction.ts` / `manager.ts` / `cacheEntry.ts`** — Split-point now also accepts `assistant` boundaries plus an `extractedUpTo` watermark remapped after compaction. Watermark math is clamped with `Math.min`/`Math.max`, so no out-of-bounds slice. No untrusted-input-to-sink path.
- **`lib/prelude.ts`** — Adding `setAgentCwd`/`getAgentCwd`/`applyAgentCwd` to the prelude is a DX change; these were already importable, so no new capability or privilege boundary.
- **`stdlib/agents/**` + `lib/agents/agency-agent/**`** — Budget guards, `saveDraft` salvage, `whatIAmDoing` (a `print` wrapper), tool renames, model-id updates. No shell/eval/path sinks.
- **`stdlib/policy.agency`** — Policy handler now `pass()`es on `std::guard` trips so the owning `handle` block decides budget grants. This narrows rather than widens what the auto-approve handler acts on; no approval bypass.
- **`package.json` / `pnpm-lock.yaml`** — Routine patch bumps with pinned integrity hashes.

Nothing met the bar for unauthorized access, data breach, RCE, or data exposure.

_🤖 Automated security review via Claude Code_

egonSchiele:
Superseded by #645, which contains only the CI workflow change. This branch was accidentally cut from `adit/misc3`, so it swept in 74 unrelated files.

PR #650, opened 2026-07-22
Count cache reads and writes in the token/cost breakdown

egonSchiele:
## The problem

A cost breakdown that could not explain its own total. From a real
agency-agent run (`agentEnd`):

```
inputCost   0.030525
outputCost  0.122090
totalCost   0.577170     ← input + output = $0.15
```

The missing $0.42 was cache — $0.385 of cache writes and $0.039 of cache
reads — and neither had a field to land in. The token counts were wrong the
same way: `totalTokens: 220170` against 130,008 across the reported fields,
with the missing 90,162 being cache-creation tokens.

On that run, cache accounted for **74% of the money and 94% of the tokens**.
Reporting input and output alone described about a quarter of the bill.

## Why it happened

`StatUsage` and `StatCost` in `lib/runtime/utils.ts` had no fields for cache
reads or cache writes, so `updateTokenStats` had nowhere to put them.

The data was never missing. smoltalk returns `cachedInputTokens`,
`cacheCreationInputTokens`, `cachedInputCost`, and `cacheCreationInputCost`
on every completion, and the per-call `promptCompletion` events in the
statelog already carry all four. Only the aggregation dropped them.

This matters because providers price the three input classes differently. On
Anthropic, derived from the log's own numbers: a Sonnet cache write bills at
$3.75/M against $3.00/M for ordinary input, and a cache read at $0.30/M.
Collapsing them loses exactly the part of the bill that dominates a long
agent run.

## The change

Four fields added to `TokenUsage` / `TokenCost` (`lib/statelogClient.ts`),
matching fields in `StatUsage` / `StatCost`, and zero-initialized in
`GlobalStore.withTokenStats()`.

Accumulation now goes through a small `addTo` helper that treats an absent
counter as zero. That is not decoration: token stats restored from a
checkpoint written before these fields existed have `undefined` there, and
`undefined += n` is `NaN`, which would then poison every later total on the
run.

## Verification

Replaying the original trace's 12 `promptCompletion` events through the
fixed aggregator:

```
cost:  input 0.030525 + output 0.122090 + cachedInput 0.039231
       + cacheCreation 0.385324  =  0.577170   (totalCost 0.577170)

usage: 6,625 + 7,608 + 115,775 + 90,162  =  220,170   (totalTokens 220,170)
```

Both reconcile exactly, where before they were short by $0.42 and 90,162
tokens.

Six new cases in `lib/runtime/utils.test.ts`, including one asserting the
breakdown sums to the total and one covering the restored-from-old-checkpoint
path. `pnpm run typecheck` clean; `lib/runtime/utils.test.ts` 30/30;
`globalStore` / `statelogClient` / `prompt` / `thread` tests 94/94.

## Deliberately not included

`ModelUsage` (the per-model breakdown behind `std::thread`'s `getModelCosts`
and the REPL footer) still reports only `inputTokens` / `outputTokens` /
`cost`. Its `cost` is the full total and so is already correct; adding cache
fields there changes an Agency-visible type and belongs in its own change.

---

## Follow-up commit: the display sites

Fixing the aggregate was not enough — every place that shows a token count
to a user was still adding `inputTokens + outputTokens`. On the same run:

| Surface | Reported | Actual |
| --- | --- | --- |
| `/cost` → `Tokens:` | 14,233 | 220,170 |
| REPL footer `↑` | 6,625 | 212,562 |

The cost figures were right the whole time, which is what made it
confusing: the money said one thing and the tokens said something fifteen
times smaller.

Four changes:

- **The per-model accumulator** in `updateTokenStats` did not track cache at
  all, so `ModelUsage` had nothing to report even after the aggregate was
  fixed. It now accumulates both cache fields through the same `addTo`.
- **`normalizeModelUsage`** and the **`ModelCost` type in `std::thread`**
  carry the two new fields, so `getModelCosts()` can report them.
- **The REPL footer** counts cache reads and writes in `↑`. That arrow means
  "how much went up the wire", and both are input the provider billed for.
- **`/cost`** counts them in its total and adds a breakout line:

  ```
  Cost:   $0.5772
  Tokens: 220170
          90162 written to cache, 115775 read back

  By model:
    claude-sonnet-5  ↑165864 ↓6811  $0.4013
    claude-opus-4-8  ↑46698 ↓797  $0.1758
  ```

  That ratio is the useful number: a cache write only pays for itself once
  something reads it. 90k written against 115k read is 1.28 reads per token,
  which is very low — it means contexts are being built and discarded.

The trace event's token delta (`eventLog.getTokenUsageDiff`) gains the same
field, so a cache write shows up in a trace instead of vanishing between two
checkpoints.

### Checked and left alone

- **Node return value** (`createReturnObject`) and **`agentEnd.tokenStats`**
  pass the whole `__tokenStats` object through, so they picked up the new
  fields from the initializer with no change.
- **`getCost()` / `getTokens()`** read the per-branch accumulator, which is
  fed from `completion.cost.totalCost` and `completion.usage.totalTokens` —
  both already whole-value, both already correct.
- **`tokenStatsFromJSON`** and the debugger's stats line return only
  `{totalTokens, totalCost}`, which were correct already.

### Testing

`pnpm run test:agents` 202/202. `lib/runtime/utils.test.ts` 32/32, including
new coverage for `normalizeModelUsage` and for a per-model entry restored
without the cache fields.

The full `vitest run --dir lib` suite is 8,999–9,000 passing with 2–3
failures that differ between runs and all pass in isolation
(`runPolicy.spawn`, `staticInit.crossModule`, `topsortCycleErrors` — all
subprocess-spawning tests, none of which touch token accounting). That looks
like load flakiness in my environment rather than a regression, but I am
flagging it rather than claiming a clean run.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/lib/repl.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/lib/repl.agency
@@ -178,8 +178,8 @@ def costSlashCommand(): boolean {
   const byModel = getModelCosts()
   let totalCost = 0.0
   let totalTokens = 0
-  let totalRead = 0
-  let totalWritten = 0
+  let totalCacheRead = 0
+  let totalCacheWrite = 0
   for (m in byModel) {
     totalCost = totalCost + m.cost
     // Cache reads and writes are input tokens too. Counting only
@@ -188,14 +188,14 @@ def costSlashCommand(): boolean {
     totalTokens = totalTokens
       + m.inputTokens + m.outputTokens
       + m.cachedInputTokens + m.cacheCreationInputTokens
-    totalRead = totalRead + m.cachedInputTokens
-    totalWritten = totalWritten + m.cacheCreationInputTokens
+    totalCacheRead = totalCacheRead + m.cachedInputTokens
+    totalCacheWrite = totalCacheWrite + m.cacheCreationInputTokens
   }
   pushMessage("Cost:   $${totalCost.toFixed(4)}")
   pushMessage("Tokens: ${totalTokens}")
-  if (totalRead > 0 || totalWritten > 0) {
+  if (totalCacheRead > 0 || totalCacheWrite > 0) {
     pushMessage(
-      color.dim("        ${totalWritten} written to cache, ${totalRead} read back"),
+      color.dim("        ${totalCacheWrite} written to cache, ${totalCacheRead} read back"),
     )
   }
   if (byModel.length > 0) {
--- a/packages/agency-lang/lib/agents/agency-agent/lib/repl.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/lib/repl.agency
@@ -178,18 +178,33 @@ def costSlashCommand(): boolean {
   const byModel = getModelCosts()
   let totalCost = 0.0
   let totalTokens = 0
+  let totalRead = 0
+  let totalWritten = 0
   for (m in byModel) {
     totalCost = totalCost + m.cost
-    totalTokens = totalTokens + m.inputTokens + m.outputTokens
+    // Cache reads and writes are input tokens too. Counting only
+    // inputTokens + outputTokens reports a fraction of a cached
+    // conversation — on a long agent run, often under a tenth of it.
+    totalTokens = totalTokens
+      + m.inputTokens + m.outputTokens
+      + m.cachedInputTokens + m.cacheCreationInputTokens
+    totalRead = totalRead + m.cachedInputTokens
+    totalWritten = totalWritten + m.cacheCreationInputTokens
   }
   pushMessage("Cost:   $${totalCost.toFixed(4)}")
   pushMessage("Tokens: ${totalTokens}")
+  if (totalRead > 0 || totalWritten > 0) {
+    pushMessage(
+      color.dim("        ${totalWritten} written to cache, ${totalRead} read back"),
+    )
+  }
   if (byModel.length > 0) {
     pushMessage("")
     pushMessage("By model:")
     for (m in byModel) {
+      const up = m.inputTokens + m.cachedInputTokens + m.cacheCreationInputTokens
       pushMessage(
-        "  ${m.model}  ↑${m.inputTokens} ↓${m.outputTokens}  $${m.cost.toFixed(4)}",
+        "  ${m.model}  ↑${up} ↓${m.outputTokens}  $${m.cost.toFixed(4)}",
       )
     }
   }
--- a/packages/agency-lang/lib/runtime/utils.ts
+++ b/packages/agency-lang/lib/runtime/utils.ts
@@ -171,6 +171,10 @@ export type ModelUsage = {
   model: string;
   inputTokens: number;
   outputTokens: number;
+  /** Tokens read back from an existing cache entry. */
+  cachedInputTokens: number;
+  /** Tokens written to a new cache entry. */
+  cacheCreationInputTokens: number;
   cost: number;
 };
 
@@ -189,6 +193,12 @@ export function normalizeModelUsage(rawModels: unknown): ModelUsage[] {
       model,
       inputTokens: typeof v?.inputTokens === "number" ? v.inputTokens : 0,
       outputTokens: typeof v?.outputTokens === "number" ? v.outputTokens : 0,
+      cachedInputTokens:
+        typeof v?.cachedInputTokens === "number" ? v.cachedInputTokens : 0,
+      cacheCreationInputTokens:
+        typeof v?.cacheCreationInputTokens === "number"
+          ? v.cacheCreationInputTokens
+          : 0,
       cost: typeof v?.totalCost === "number" ? v.totalCost : 0,
     });
   }
@@ -248,9 +258,11 @@ export function updateTokenStats(args: {
     }
     const m = tokenStats.models[model] ??
       (tokenStats.models[model] = { inputTokens: 0, outputTokens: 0, totalCost: 0 });
-    m.inputTokens += usage.inputTokens || 0;
-    m.outputTokens += usage.outputTokens || 0;
-    m.totalCost += cost.totalCost || 0;
+    addTo(m, "inputTokens", usage.inputTokens);
+    addTo(m, "outputTokens", usage.outputTokens);
+    addTo(m, "cachedInputTokens", usage.cachedInputTokens);
+    addTo(m, "cacheCreationInputTokens", usage.cacheCreationInputTokens);
+    addTo(m, "totalCost", cost.totalCost);
   }
   // Cache reads and cache writes are tracked separately from ordinary input.
   // Without them the breakdown does not reconcile with its own total: on a
--- a/packages/agency-lang/lib/runtime/utils.ts
+++ b/packages/agency-lang/lib/runtime/utils.ts
@@ -223,11 +223,22 @@ type StatCost = {
   totalCost?: number;
 };
 
-/** Add to a counter that may be absent. Token-stats objects restored from a
- *  checkpoint written before a field existed have `undefined` there, and
- *  `undefined += n` is NaN — which then poisons every later total. */
-function addTo(target: Record<string, number>, key: string, amount: number | undefined): void {
-  target[key] = (target[key] || 0) + (amount || 0);
+/** Add to one counter on a token-stats object.
+ *
+ *  The slot may be missing: stats restored from a checkpoint written before
+ *  a field existed have `undefined` there, and `undefined + n` is NaN, which
+ *  then poisons every later total on the run.
+ *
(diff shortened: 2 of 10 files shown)


packages/agency-lang/lib/runtime/utils.ts
```diff
@@ -202,14 +202,24 @@ type StatUsage = {
...
+  cacheCreationInputCost?: number;
   totalCost?: number;
 };
 
+/** Add to a counter that may be absent. Token-stats objects restored from a
+ *  checkpoint written before a field existed have `undefined` there, and
+ *  `undefined += n` is NaN — which then poisons every later total. */
+function addTo(target: Record<string, number>, key: string, amount: number | undefined): void {
+  target[key] = (target[key] || 0) + (amount || 0);
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in e72ad071e.

You are right that the annotation was wrong, and wrong in both of the ways you name. `tokenStats.cost` really does carry `currency: "USD"` next to its counters, so `Record<string, number>` was a false claim about the object being passed; and asserting every key is present contradicts the one thing the helper exists to handle. Neither surfaced as a type error only because `getTokenStats()` returns `any`, so the annotation was never checked against a real shape.

Values are now `unknown`, and a slot that does not hold a finite number starts from zero.

One correction on the NaN half: it was already safe, by accident rather than design. `NaN` is falsy, so `NaN || 0` evaluated to `0` and a poisoned counter self-healed. The explicit `Number.isFinite` check preserves that and adds the case the old code genuinely got wrong — a slot holding a string would have been concatenated into (`"12" + 5` → `"125"`) rather than counted.

To be clear about severity: no call site passes a non-numeric key, so this was not a live bug. It was an inaccurate type over an `any`, which is the kind that goes unnoticed until someone gives `getTokenStats()` a real return type.

Three tests added: currency survives accumulation, a NaN counter recovers, and a string counter restarts rather than concatenating.

packages/agency-lang/stdlib/thread.agency
```diff
@@ -196,6 +196,12 @@ export type ModelCost = {
   model: string;
   inputTokens: number;
   outputTokens: number;
+  // Tokens read back from a cache entry, and tokens written to a new one.
+  // Both are input-side and both are billed, at rates that differ from
+  // `inputTokens` — so a total that leaves them out understates a cached
+  // conversation by most of its size.
+  cachedInputTokens: number;
+  cacheCreationInputTokens: number;
```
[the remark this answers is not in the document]
egonSchiele: Fixed in e72ad071e — this was a real miss, thank you.

To be precise about the cause, since it is not quite what the comment says: `make` did run `agency doc` and did regenerate the page. `all` runs `$(MAKE) doc` (Makefile:42), so the regenerated `docs/site/stdlib/thread.md` was sitting modified in my working tree the whole time. What went wrong is that the commit scoped its `git add` to `lib/` and the `.agency` source, so the regenerated page was never staged.

So the failure mode to watch for here is not "forgot to run make" but "ran make, then committed a narrower path than make touched". A scoped `git add` after a codegen step will silently drop generated output every time.

The doc commit carries the `ModelCost` field addition plus the source-line-number shifts that come with it.

packages/agency-lang/lib/agents/agency-agent/lib/repl.agency
```diff
@@ -178,18 +178,33 @@ def costSlashCommand(): boolean {
...
   for (m in byModel) {
     totalCost = totalCost + m.cost
-    totalTokens = totalTokens + m.inputTokens + m.outputTokens
+    // Cache reads and writes are input tokens too. Counting only
+    // inputTokens + outputTokens reports a fraction of a cached
+    // conversation — on a long agent run, often under a tenth of it.
+    totalTokens = totalTokens
+      + m.inputTokens + m.outputTokens
+      + m.cachedInputTokens + m.cacheCreationInputTokens
+    totalRead = totalRead + m.cachedInputTokens
```
egonSchiele: Confusing var names ... lets use `totalCacheRead` and `totalCacheWrite`

packages/agency-lang/lib/runtime/utils.ts
```diff
@@ -189,6 +193,12 @@ export function normalizeModelUsage(rawModels: unknown): ModelUsage[] {
       model,
       inputTokens: typeof v?.inputTokens === "number" ? v.inputTokens : 0,
       outputTokens: typeof v?.outputTokens === "number" ? v.outputTokens : 0,
+      cachedInputTokens:
+        typeof v?.cachedInputTokens === "number" ? v.cachedInputTokens : 0,
```
egonSchiele: This is a place where a subtle bug could hide and we wouldn't catch it. What happens if this value is not a number? What could it be? Instead of coercing to zero, we probably want to show an error. Please create a helper function for this and use it for all of the values here that do this sort of ternary check.

packages/agency-lang/lib/runtime/utils.ts
```diff
@@ -202,14 +212,35 @@ type StatUsage = {
...
+ *
+ *  The slot may also not be a number at all. `tokenStats.cost` holds
+ *  `currency: "USD"` alongside its counters, so the object this walks is not
+ *  uniformly numeric — hence `unknown` values rather than a `Record<string,
+ *  number>` that would claim otherwise. Anything that is not a finite number
+ *  starts from zero, which is the right reading for a counter and keeps a
+ *  stray string from being concatenated into one. */
+function addTo(target: Record<string, unknown>, key: string, amount: number | undefined): void {
+  const current = target[key];
+  const base = typeof current === "number" && Number.isFinite(current) ? current : 0;
```
egonSchiele: similar issue -- why did someone pass in something that is not a number, or not finite?

packages/agency-lang/lib/agents/agency-agent/lib/repl.agency
egonSchiele: Done in d0d734408 — `totalCacheRead` / `totalCacheWrite`.

packages/agency-lang/lib/runtime/utils.ts
egonSchiele: Fixed in d0d734408. You are right, and answering "what could it be?" turned out to split the case in two — the single ternary was papering over both.

**Missing** (`undefined` / `null`) is ordinary. Three legitimate sources: stats restored from a checkpoint written before a field existed have no key for it; a provider that does not do prompt caching never reports a cache count at all; and a fresh per-model entry only has the keys `updateTokenStats` has written so far. Zero is the correct answer for all three, and erroring on them would fire on essentially every run.

**Present but not a finite number** is a bug, and this is the case that was hiding. `updateTokenStats` is the only writer and only ever writes numbers, so a string or an object in one of these slots means either the stats object was corrupted upstream, or a provider returned a non-number in its usage/cost payload. That second one is a real hole: the `number | undefined` parameter types are erased at runtime, so nothing actually checks what a provider hands back. NaN belongs here too — it is the signature of an earlier `undefined + n`, and left alone it spreads to every total it touches.

New helper `counterValue(value, where)` is the single place that decides this, used by all five reads in `normalizeModelUsage` and by both sides of `addTo`.

One place I did not follow you exactly, and want to flag rather than bury: I warn rather than throw. Every caller of this is a cost display, a REPL footer, or a log line. Throwing would take a run down over an accounting slot, which trades a wrong number for a dead session. The warning names the slot, says the totals will be short by whatever it held, and says that only `updateTokenStats` writes there — so it is findable without being fatal. If you would rather it throw, say so and I will switch it; it is one line.

Seven tests replace three, including the two silent-is-correct cases so the warning cannot start firing on normal runs. Replaying the original 12-call trace produces zero warnings and identical totals.

packages/agency-lang/lib/runtime/utils.ts
egonSchiele: Same fix in d0d734408 — `addTo` now runs both sides through `counterValue`, so this is no longer a separate check.

Worth separating the two sides, because your question has a different answer for each:

The **target** side (the existing slot) is almost always the benign case: a checkpoint written before the field existed simply has no key, so it is missing rather than wrong. That reads zero silently.

The **incoming** side is the one that was genuinely unguarded, and it is where a real non-number can come from. `updateTokenStats` takes `usage` and `cost` straight off the provider completion. The parameter types say `number | undefined`, but those are erased — nothing validates the payload at runtime. A provider returning `"1234"` as a string would previously have gone through `amount || 0` and been silently counted. Now it warns, and the message says `(incoming)` so you can tell provider garbage from a corrupted stats object at a glance.

Non-finite specifically: NaN gets here when something already did `undefined + n` upstream. Since NaN is falsy, the old `|| 0` silently swallowed it — which is exactly the kind of quiet recovery you are objecting to. It is now audible.

PR #651, opened 2026-07-22
queueMessage: public API for queueing conversation messages + one internal turn-boundary concept

egonSchiele:
Two deliverables, per the design doc (`docs/superpowers/specs/2026-07-21-turn-boundary-unification-design.md` on the `callback-owner-scope-redirect` worktree, plus its two review rounds):

## 1. A public primitive: `agency.thread.current().queueMessage(...)`

Queue a message onto a conversation from any TS helper, at any moment, and the runtime delivers it at the next safe point. The pending queue lives on the `MessageThread` and serializes with it (including through `adoptFrom`, which the resume path uses - a queue that survived only `toJSON` would have been dropped exactly on resume).

Delivery semantics, decided and tested rather than emergent:
- Delivered before the thread's next request-turn: at a call's start for messages queued between calls (so no-tool calls deliver), after a tool round for messages queued mid-turn.
- Retry re-issues (trip/validation) are the same turn re-sent and do not deliver.
- Thread-scoped: queue onto a non-active thread and it waits for THAT thread.
- Roles user/assistant only; no mid-thread system messages.
- Inside a TOOL BODY, `current()` is the tool's private throwaway thread, so a queueMessage there dies with the tool (consistent with tool-thread isolation; documented in the ts-helpers guide). Mid-turn steering of the outer conversation belongs to callbacks, which fire where the outer conversation is current - the agency-js test demonstrates exactly that with an `onToolCallEnd` callback.

## 2. The turn boundary as one internal concept

`lib/runtime/turnBoundary.ts` now owns message delivery at the boundaries: three producers (reply attachments, guard feedback, queued messages), one drain with the step-and-snapshot discipline written once, and the canonical order pinned in code: attachments -> queued -> gate -> feedback. The four gate-and-feedback sites and the round-boundary block in `prompt.ts` become calls into it; `drainGuardFeedback` and the inline attachReplies block are gone. Producers are internal - features inject by calling `queueMessage`, not by registering producers.

Step-key names are unchanged; two new keys (`queuedMessages.initial`, `round.N.queuedMessages`); every producer step now opens unconditionally (cross-version checkpoint compatibility waived by owner - the old conditional attachReplies emission was the only casualty).

## No-behavior-change evidence for the refactor half

Five baseline fixtures recorded on clean main and re-run after each migration step, all passing identically: `attach-to-reply` (all six cases), `guards/trip-feedback`, `guards/guard-cost-shared-outer-trips-mid-fork` (branch-local queue), `guards/fake-clock-trip` (trip-retry site), `guards/savedraft-annotated-structured` (validation site - found during execution, so no CI-only reliance was needed).

One hazard caught during migration and now pinned by a unit test: `pr.step` treats a gate body returning `Interrupt[]` as checkpoint-and-pause. Typing the gate as `Promise<void>` and wrapping it would have silently discarded the guard pause signal; `BoundaryContext` types it honestly and a test asserts the return value flows through untouched.

## Tests

- 18 turnBoundary unit tests: producer semantics (merge rules, destructive takes, label parity via `labelAt`), canonical step sequences for both boundaries (identical with and without pending work), content ordering, and the pause-signal flow-through pin.
- `MessageThread` queue tests including the serialize and adoptFrom round trips.
- `tests/agency-js/queue-message`: a recording LLM client asserts WHERE messages land in the requests the model actually sees - mid-round via callback (after the tool result, absent from the first request), initial-boundary delivery for a no-tool call (ahead of the prompt), thread-scoping (never delivered to the main conversation), and an order+resume scenario where an attachment, a queued message, and guard feedback co-occur in one round with a REAL guard-trip pause in the middle: all three present, in order, exactly once.

Deviation from the plan, for the record: the plan called for pure-agency order/resume fixtures, but `queueMessage` has no Agency-language surface in v1, so those scenarios live in the agency-js suite where TS can call the helper.

The callback-injection plan has been amended (in the design worktree) to consume this: its buffer/flush machinery is superseded by `queueMessage`, with the trip-retry double-queue question explicitly flagged as a design task for its spec revision.

### DIFF
--- a/packages/agency-lang/docs/site/guide/ts-helpers.md
+++ b/packages/agency-lang/docs/site/guide/ts-helpers.md
@@ -98,6 +98,7 @@ The active thread is the `MessageThread` instance the surrounding LLM call (or b
 | `agency.thread.assistant(content)` | push an assistant-role message |
 | `agency.thread.store()` / `storeMaybe()` | the full `ThreadStore` |
 | `agency.thread.with(threadId, fn)` | run `fn` with `threadId` as the active thread |
+| `agency.thread.current().queueMessage(content, opts?)` | queue a message for the thread's next request-turn |
 
 ### Pushing messages
 
@@ -107,6 +108,48 @@ export async function setupConversation(systemPrompt: string): Promise<void> {
   agency.thread.user("Begin.");
 }
 ```
+
+### Queueing a message for the next turn
+
+`thread.user(...)` pushes immediately, which is unsafe in the middle of a
+model turn: a message inserted between an assistant's tool calls and their
+results produces a conversation the provider rejects. `queueMessage` is the
+deferred, always-safe alternative. It appends to a pending queue on the
+thread and returns; the runtime delivers the queue at the next safe point.
+
+```ts
+agency.thread.current().queueMessage("Budget check: wrap up soon.");
+agency.thread.current().queueMessage(content, {
+  role: "assistant",   // default "user"; "system" is not allowed
+  label: "myFeature",  // statelog label, default null
+});
+```
+
+The delivery semantics, precisely:
+
+- **Delivered before the thread's next request-turn.** A message queued
+  between calls is delivered at the start of the thread's next `llm()`
+  call, ahead of the new prompt — so a call that uses no tools still
+  delivers. A message queued during a turn (from a callback, say) is
+  delivered after that turn's tool results and before the next request.
+- **Retry re-issues do not deliver.** A trip retry or validation retry
+  re-sends the same turn; messages queued during it wait for the next
+  real turn.
+- **The queue waits indefinitely.** If no `llm()` ever runs against the
+  thread again, the queue stays on the thread, serialized with it across
+  pauses. No deadline, no error.
+- **Scoped to the thread you call it on.** Queueing onto a non-active
+  thread delivers when THAT thread next runs a call.
+- **Inside a tool body, `current()` is the tool's private thread.** Tool
+  bodies run against an isolated, throwaway `ThreadStore` (that isolation
+  is what keeps a tool's own `llm()` calls out of the outer
+  conversation), so a `queueMessage` there dies with the tool. To hand
+  content outward from a tool, use `attachToReply`; to steer the outer
+  conversation mid-turn, queue from a callback — callbacks fire where the
+  outer conversation is current.
+- No `"system"` role: a system message appearing mid-conversation is
+  rejected or silently reinterpreted by some providers. Say `role:
+  "user"`.
 
 ### Switching threads
 
--- a/packages/agency-lang/docs/site/guide/ts-helpers.md
+++ b/packages/agency-lang/docs/site/guide/ts-helpers.md
@@ -111,11 +111,12 @@ export async function setupConversation(systemPrompt: string): Promise<void> {
 
 ### Queueing a message for the next turn
 
-`thread.user(...)` pushes immediately, which is unsafe in the middle of a
-model turn: a message inserted between an assistant's tool calls and their
-results produces a conversation the provider rejects. `queueMessage` is the
-deferred, always-safe alternative. It appends to a pending queue on the
-thread and returns; the runtime delivers the queue at the next safe point.
+`thread.user(...)` pushes immediately, which mid-turn can land a message
+between an assistant's tool calls and their results — a conversation the
+provider rejects. `queueMessage` defers instead: it queues on the thread,
+and the runtime delivers before the thread's next request. Queued between
+calls, the message arrives ahead of the next call's prompt; queued during
+a turn (from a callback, say), it arrives after that turn's tool results.
 
 ```ts
 agency.thread.current().queueMessage("Budget check: wrap up soon.");
@@ -125,31 +126,13 @@ agency.thread.current().queueMessage(content, {
 });
 ```
 
-The delivery semantics, precisely:
-
-- **Delivered before the thread's next request-turn.** A message queued
-  between calls is delivered at the start of the thread's next `llm()`
-  call, ahead of the new prompt — so a call that uses no tools still
-  delivers. A message queued during a turn (from a callback, say) is
-  delivered after that turn's tool results and before the next request.
-- **Retry re-issues do not deliver.** A trip retry or validation retry
-  re-sends the same turn; messages queued during it wait for the next
-  real turn.
-- **The queue waits indefinitely.** If no `llm()` ever runs against the
-  thread again, the queue stays on the thread, serialized with it across
-  pauses. No deadline, no error.
-- **Scoped to the thread you call it on.** Queueing onto a non-active
-  thread delivers when THAT thread next runs a call.
-- **Inside a tool body, `current()` is the tool's private thread.** Tool
-  bodies run against an isolated, throwaway `ThreadStore` (that isolation
-  is what keeps a tool's own `llm()` calls out of the outer
-  conversation), so a `queueMessage` there dies with the tool. To hand
-  content outward from a tool, use `attachToReply`; to steer the outer
-  conversation mid-turn, queue from a callback — callbacks fire where the
-  outer conversation is current.
-- No `"system"` role: a system message appearing mid-conversation is
-  rejected or silently reinterpreted by some providers. Say `role:
-  "user"`.
+The queue is serialized with the thread and waits indefinitely — if the
+thread never runs another `llm()`, nothing is delivered and nothing
+errors. Retry re-issues (guard trips, validation) re-send the same turn
+and do not deliver. Inside a tool body, `current()` is the tool's private
+throwaway thread, so a `queueMessage` there dies with the tool; use
+`attachToReply` to hand content outward, or queue from a callback, which
+fires where the outer conversation is current.
 
 ### Switching threads
 
--- a/packages/agency-lang/lib/runtime/state/messageThread.ts
+++ b/packages/agency-lang/lib/runtime/state/messageThread.ts
@@ -1,6 +1,13 @@
 import * as smoltalk from "smoltalk";
 import { nanoid } from "nanoid";
-import { deepClone } from "../utils.js";
+
+/** Queue entries are plain JSON-safe data (the QueuedMessage contract), so
+ *  the built-in structuredClone suffices. Deliberately NOT runtime/utils'
+ *  deepClone: importing it here creates the cycle messageThread → utils →
+ *  runtime/index → threadStore → messageThread, which left ThreadStore
+ *  undefined while bootstrapThreadStore's `extends` evaluated (caught by
+ *  CI's full-suite load order; the touched-suite runs missed it). */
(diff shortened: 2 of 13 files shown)


egonSchiele (commented):
## Review

Read the full diff and traced it against the code. This is a strong PR — it resolved essentially every open point from the design-doc review rounds, and the one hazard it caught during migration is a genuine safety bug that the plan would have introduced. One real code issue and one minor test-hygiene note; everything else is affirmation.

### The standout: the gate return-type hazard

`BoundaryContext.step` and `guardGate` are typed `Promise<Interrupt[] | void>`, not `Promise<void>`, and there's a regression test pinning that the gate's return flows through `step` untouched. This matters more than it looks. `pr.step` treats a body returning `Interrupt[]` as checkpoint-and-pause — that is *how* a guard trip suspends the run. Had the boundary wrapped the gate in a void-returning adapter (the obvious refactor, and what a naive reading of the plan's `guardGate: () => Promise<void>` would have produced), the guard pause signal would have been silently discarded. Guards are safety infrastructure, so that would have been a serious latent bug. Catching it during migration and pinning it with a test is exactly right.

### Verified correct

- **Delivery semantics are now two boundaries, and both are right.** `runInitialBoundary` drains `queuedMessages.initial` before the first request (so a no-tool call still delivers, ahead of the prompt), and `runRoundBoundary` drains `round.N.queuedMessages` after tool results. Retry and validation sites use `runGateAndFeedback` with no queued drain, so re-issued turns correctly do not re-deliver. This resolves the delivery-point ambiguity from the review rounds cleanly.
- **The `adoptFrom` catch is real and load-bearing.** `prompt.ts` restores a resumed call via `args.messages.adoptFrom(restored)`, so a queue that survived `toJSON` but not `adoptFrom` would be dropped exactly on resume. `adoptFrom` copies `queuedMessages` and there's a dedicated test for it. Easy to miss, correctly handled.
- **Resume idempotency holds.** `takeQueuedMessages` empties the queue inside the drain step, before push and snapshot, and the emptied queue is captured in `messagesJSON` (omitted when empty). A mid-step interrupt leaves the step unmarked and restores from the prior snapshot with the queue intact, so no message is lost or double-delivered. Scenario 4 proves exactly-once across a real pause.
- **The agency-js test is the right kind of test.** A recording client asserting on the actual `config.messages` of every request is the only honest way to verify *where* a message lands. It covers mid-round-via-callback, no-tool initial delivery, thread-scoping, and order+resume with a genuine guard trip — not a mock pause. The label-parity unit test (`label: null` stores what a label-less push stores) and the broadened guard-fixture baseline (fork, trip-retry, validation) close the coverage gaps flagged in review.

### One code issue: assistant-role content cast

In `queuedMessagesProducer`:

```ts
q.role === "assistant"
  ? smoltalk.assistantMessage(q.content as string)
  : smoltalk.userMessage(q.content as smoltalk.UserContentInput),
```

`QueuedMessage.content` is typed `string | smoltalk.UserContentInput` for *both* roles, but the assistant branch casts to `string`. So `queueMessage([...parts], { role: "assistant" })` passes an array where `assistantMessage` expects a string, and the `as string` hides it. Assistant messages with multimodal content are unusual, so this is an edge, but the cast is dishonest and could produce a malformed message. Cleanest fix is to narrow the type — assistant content is `string` only — so the compiler rejects the array at the `queueMessage` call site rather than the producer papering over it. If assistant array content is intended to be supported, the producer should handle it instead of casting.

### Minor: test writes a fixed /tmp path

Scenario 4 writes `/tmp/qm-order-chart.png` at a hardcoded path. Concurrent test runs would race on the same file. A unique temp path (or the scratchpad) avoids the collision. Low severity, test-only.

### Not blocking

- Single delivery of a message queued *before* a tools-using call (drained at `initial`, then `round` finds nothing) isn't exercised by a dedicated scenario, but the destructive `take` plus the exactly-once resume test cover the mechanism.

Nice work — the safety-signal catch alone justifies the review rounds this went through.

packages/agency-lang/lib/runtime/state/messageThread.ts
```diff
@@ -133,6 +155,36 @@ export class MessageThread {
...
+  queueMessage(
+    content: QueuedMessage["content"],
+    opts: { role?: "user" | "assistant"; label?: string } = {},
+  ): void {
+    this.queuedMessages.push({
+      role: opts.role ?? "user",
+      content,
+      label: opts.label ?? null,
+    });
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 4db1d6df3. queueMessage now validates at runtime: role must be "user" or "assistant" (precise error otherwise), and assistant-role content must be a string. The QueuedMessage type became a discriminated union so the compiler enforces the same rules for typed callers.

packages/agency-lang/lib/runtime/state/messageThread.ts
```diff
@@ -169,6 +221,11 @@ export class MessageThread {
     if (this.messageLabels.some((l) => l !== null)) {
       json.messageLabels = [...this.messageLabels];
     }
+    // Same emit-only-when-meaningful rule (and same copy-not-alias rule)
+    // as messageLabels above.
+    if (this.queuedMessages.length > 0) {
+      json.queuedMessages = [...this.queuedMessages];
+    }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 4db1d6df3. toJSON now deep-clones the entries (the spread copied the array but aliased the objects, exactly as you said), with a comment explaining why labels get away with a spread and entries do not. A test mutates the returned JSON and asserts the live queue is untouched.

packages/agency-lang/lib/runtime/state/messageThread.ts
```diff
@@ -228,6 +285,13 @@ export class MessageThread {
     thread.hidden = _hidden;
     thread.label = _label;
     thread.summary = _summary;
+    if (
+      !Array.isArray(json) &&
+      "queuedMessages" in json &&
+      json.queuedMessages !== undefined
+    ) {
+      thread.queuedMessages = [...json.queuedMessages];
+    }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 4db1d6df3. fromJSON accepts only a real array: queuedMessages: null (or any non-array) revives as an empty queue, same as the absent key, with a test covering the null case.

egonSchiele:
All review points addressed in 4db1d6df3:

1. **Assistant-content cast** — `QueuedMessage` is now a discriminated union with a string-only assistant arm, so the producer's `as string` is gone and `queueMessage([...parts], { role: "assistant" })` is a compile error for typed callers. Since the API is also callable from plain JS, the same rules are enforced at runtime with errors at the enqueue site (invalid role, non-string assistant content) — which also resolves the first Copilot comment.
2. **Fixed /tmp path** — the order scenario now writes a per-process temp path and delivers it to the agency tool through the mocked tool-call arguments, so concurrent runs cannot race.
3. **Copilot: toJSON aliasing** — entries are deep-cloned; a test mutates the returned JSON and asserts the live queue is untouched.
4. **Copilot: fromJSON null-safety** — only a real array is accepted; `queuedMessages: null` revives as an empty queue.

42 unit tests green on the touched suites, agency-js suite green after `make`.

On the non-blocking note (queue-before-a-tools-using-call delivering once at `initial` and not again at `round`): agreed the destructive take covers the mechanism; happy to add the dedicated scenario if you want it pinned explicitly.

egonSchiele:
CI failure root-caused and fixed in : the review-round deepClone import created the module cycle messageThread -> utils -> runtime/index -> threadStore, which left ThreadStore undefined while bootstrapThreadStore evaluated its extends clause. That single cycle explains both failing jobs: the build job (threadStore.test.ts suite crash) and agency-tests shard 4 (every test.js importing agency-lang/runtime died at load; the AG4010 lines in that log are pre-existing local noise, not the failure). Queue entries are plain JSON-safe data by contract, so structuredClone replaces the import, with a comment naming the cycle. Ran the FULL unit suite locally this time (9035 passing; one debugger-timeout flake that passes in isolation) rather than only the touched suites, which is how the cycle slipped past me.

packages/agency-lang/docs/site/guide/ts-helpers.md
```diff
@@ -108,6 +109,48 @@ export async function setupConversation(systemPrompt: string): Promise<void> {
 }
 ```
 
+### Queueing a message for the next turn
```
egonSchiele, line 112: make this shorter and more readable. See how the rest of the doc is written, and copy that tone

packages/agency-lang/lib/runtime/state/messageThread.ts
```diff
@@ -1,15 +1,37 @@
 import * as smoltalk from "smoltalk";
 import { nanoid } from "nanoid";
 
+/** Queue entries are plain JSON-safe data (the QueuedMessage contract), so
+ *  the built-in structuredClone suffices. Deliberately NOT runtime/utils'
+ *  deepClone: importing it here creates the cycle messageThread → utils →
+ *  runtime/index → threadStore → messageThread, which left ThreadStore
+ *  undefined while bootstrapThreadStore's `extends` evaluated (caught by
+ *  CI's full-suite load order; the touched-suite runs missed it). */
+const cloneQueue = (q: QueuedMessage[]): QueuedMessage[] => structuredClone(q);
```
egonSchiele: First of all, structuredClone has issues. So if you don't want to use deepClone, I would suggest using JSON.stringify + JSON.parse instead.

packages/agency-lang/lib/runtime/state/messageThread.ts
```diff
@@ -119,6 +146,10 @@ export class MessageThread {
   adoptFrom(other: MessageThread): void {
     this.messages = [...other.messages];
     this.messageLabels = [...other.messageLabels];
+    // The pending queue rides along: prompt.ts restores a resumed call via
+    // adoptFrom (its args.messages alias), and a queue that survived
+    // toJSON but not adoptFrom would be dropped exactly on resume.
+    this.queuedMessages = [...other.queuedMessages];
```
egonSchiele: You defined a method to clone a queue above, but don't use it here. Is that intentional?

packages/agency-lang/docs/site/guide/ts-helpers.md
egonSchiele: Rewritten in 899aced46 - cut from six bullets to two short paragraphs plus the example, matching the sections around it.

packages/agency-lang/lib/runtime/state/messageThread.ts
egonSchiele: Switched to JSON.parse(JSON.stringify(...)) in 899aced46. Kept the local helper rather than utils deepClone because of the import cycle documented in the comment (messageThread -> utils -> runtime/index -> threadStore).

packages/agency-lang/lib/runtime/state/messageThread.ts
egonSchiele: It was a deliberate shallow copy - entries are never mutated after enqueue, so live-to-live aliasing seemed harmless - but that reasoning lived only in my head, which makes it a trap. adoptFrom now uses cloneQueue like the other two paths (899aced46).

PR #653, opened 2026-07-23
Fix FunctionRefReviver: renamed tools now round-trip, and registry misses no longer crash checkpoint writes

### LINKED ISSUE #652: FunctionRefReviver eager-throw on checkpoint serialization: registry miss for stdlib function crashes guard-trip salvage
## What happened

During a normal `agency agent` run ("What are today's top news stories?"), a time guard tripped inside the research agent's `judgeWork` verification step. While writing the guard-trip checkpoint, serialization crashed:

```
ERROR Function judgeWork threw an exception (converted to Failure): FunctionRefReviver: function "wikipedia_search" from module "stdlib/wikipedia.agency" not found in registry. The function may have been renamed or removed since this state was serialized.
    at FunctionRefReviver.findInRegistry (dist/lib/runtime/revivers/functionRefReviver.js:100:15)
    at FunctionRefReviver.revive (dist/lib/runtime/revivers/functionRefReviver.js:40:31)
    at Array.nativeTypeReviver (dist/lib/runtime/revivers/index.js:54:22)
    at JSON.parse (<anonymous>)
    at deepClone (dist/lib/runtime/utils.js:5:17)
    at State.toJSON (dist/lib/runtime/state/stateStack.js:131:21)
    at StateStack.toJSON (dist/lib/runtime/state/stateStack.js:896:31)
    at Checkpoint.fromStateStack (dist/lib/runtime/state/checkpointStore.js:132:31)
```

The run survived only because the guard interrupt carried a `draftValue` and the saveDraft salvage path returned it. `judgeWork` itself was converted to a Failure by the crash, not by the guard.

## Why it crashes

`State.toJSON` uses `deepClone`, which round-trips through `JSON.parse` with the native-type revivers installed. `FunctionRefReviver.revive` eagerly resolves every function ref against the registry and throws on a miss. The state being cloned held a ref to `wikipedia_search` from `stdlib/wikipedia.agency`, and that module was not registered in this process, so cloning the state for the checkpoint threw.

This is the same eager-throw pattern as #544. The #642 fix added lazy resolve-at-invoke refs, but only for names matching the lifted-callback pattern (`/^__cb_.+_\d+$/` in `lib/runtime/blockNames.ts`). `wikipedia_search` does not match, so it takes the eager path and throws.

Two distinct aggravating factors:

1. **Serialization-time crash, not revival-time.** The throw happens while *writing* a checkpoint (`toJSON` → `deepClone`), so a registry miss can take down a step that was otherwise succeeding. #544 was about restoring state; this shows the same reviver runs on every `deepClone` of state, including the save path.
2. **Guard-trip amplification.** The crash fired exactly at a guard trip, converting a working salvage flow into a runtime error. Guards are supposed to be the safety net; the checkpoint they trigger should not be able to crash the function they are salvaging.

## Open questions

- Why does the verifier's state hold a ref to `wikipedia_search` when the module is not in the registry? Possibly an embedded pre-serialized state (the #544 shape) or a tool list carried across processes. Worth understanding before picking a fix.
- Should `deepClone`-for-serialization resolve function refs at all? Cloning for `toJSON` only needs to preserve the ref marker; resolution is only needed at invoke time.

## Suggested direction

Extend the #642 approach: on a registry miss during clone/serialization, keep a lazy ref (and emit a statelog warning) instead of throwing, for any function name, not just `__cb_*`. Resolution failures should surface at invoke time, where the error is actionable.

## Repro context

Observed 2026-07-22 running the agency agent with the research subagent; time guard `limit: 120000` tripped in `judgeWork`, checkpoint write crashed as above. Statelog trace available on request.

egonSchiele:
Fixes #652.

## The bug

A renamed function could never survive serialization. `.rename()` returns a copy with the new name but never re-registers it, and `FunctionRefReviver.serialize()` recorded only `name` + `module`. So a serialized ref to `wikipediaSearch.rename("wikipedia_search")` looked up `wikipedia_search` in a registry that only knows `search`, and the reviver threw.

Worse, the throw fired during checkpoint **writes**, not just restores: `State.toJSON` clones state via `deepClone`, which round-trips through `JSON.parse` with the revivers installed. In the run that surfaced this, a guard-trip checkpoint crashed while salvaging a draft, converting a working `judgeWork` salvage into a runtime error (and `failOpenFeedback` then laundered that error into a review pass).

## The fix, two parts

**1. Renames round-trip.** `AgencyFunction` now carries `originalName` — the name its registered ancestor has in the registry. All four derivations (`rename`, `partial`, `withToolDefinition`, `preapprove`) thread it through. `serialize()` records it when it differs from `name`; `revive()` looks up the original and re-applies `.rename()`, the same reconstruction pattern already used for `.partial()` and `.describe()`.

**2. A genuine registry miss no longer crashes serialization.** A miss that is not a block or lifted callback now revives to a stub built from the serialized fields, so `serialize(revive(x)) === x` and the ref rides through a process that does not own the module undamaged. The stub throws a precise error (plus a statelog event) only if something actually invokes it. Unlike the callback lazy ref it does not re-check the registry at invoke time: nothing fires these spontaneously, so a call after a miss is a real error worth surfacing.

## Tests

Six new tests (rename serialization, rename round-trip, rename+partial, the exact #652 clone-through-a-miss shape, stub fidelity with bound params and preapproval) and three updated pins for the changed miss behavior. Full `lib/` unit suite and structural linter pass.

## Note

`foo.agency` also changed in this PR — it now contains the researcher news demo that surfaced the bug. Drop that commit hunk if you want the scratch file kept out of history.

### DIFF
--- a/packages/agency-lang/foo.agency
+++ b/packages/agency-lang/foo.agency
@@ -1,6 +1,15 @@
+import { researcherAgent } from "std::agents/researcher"
 import { now } from "std::date"
 node main() {
-  const start = now()
-  sleep(4)
-  print(elapsedTime(since: start))
+  handle {
+    const prompt = "What are the top news stories today?"
+    const result = researcherAgent(task: prompt)
+    print(result)
+    return result
+  } with (data) {
+    return match(data.effect) {
+      "std::search" => approve()
+      _ => reject()
+    }
+  }
 }
--- a/packages/agency-lang/foo.agency
+++ b/packages/agency-lang/foo.agency
@@ -1,15 +1,6 @@
-import { researcherAgent } from "std::agents/researcher"
 import { now } from "std::date"
 node main() {
-  handle {
-    const prompt = "What are the top news stories today?"
-    const result = researcherAgent(task: prompt)
-    print(result)
-    return result
-  } with (data) {
-    return match(data.effect) {
-      "std::search" => approve()
-      _ => reject()
-    }
-  }
+  const start = now()
+  sleep(4)
+  print(elapsedTime(since: start))
 }
--- a/packages/agency-lang/lib/runtime/agencyFunction.ts
+++ b/packages/agency-lang/lib/runtime/agencyFunction.ts
@@ -81,6 +81,7 @@ export type AgencyFunctionOpts = {
   exported?: boolean;
   markers?: ToolMarkers;
   isPreapproved?: boolean;
+  originalName?: string;
 };
 
 export class AgencyFunction {
@@ -98,10 +99,17 @@ export class AgencyFunction {
   readonly exported: boolean;
   readonly markers: ToolMarkers;
   private readonly _isPreapproved: boolean;
+  /** The name this function's registered ancestor carries in the function
+   *  registry. `.rename()` changes `name` but not this, and every other
+   *  derivation (`.partial()`, `.describe()`, `.preapprove()`) carries it
+   *  forward — so a serialized ref to any derived copy can find its
+   *  registry entry again. Equal to `name` for functions never renamed. */
+  readonly originalName: string;
 
   constructor(opts: AgencyFunctionOpts) {
     this.name = opts.name;
     this.module = opts.module;
+    this.originalName = opts.originalName ?? opts.name;
     this._fn = opts.fn;
     this.params = opts.params;
     this.toolDefinition = opts.toolDefinition;
@@ -146,6 +154,7 @@ export class AgencyFunction {
       exported: this.exported,
       markers: this.markers,
       isPreapproved: this._isPreapproved,
+      originalName: this.originalName,
     });
   }
 
@@ -249,6 +258,7 @@ export class AgencyFunction {
       exported: this.exported,
       markers: this.markers,
       isPreapproved: this._isPreapproved,
+      originalName: this.originalName,
     });
   }
 
@@ -295,6 +305,7 @@ export class AgencyFunction {
       exported: this.exported,
       markers: this.markers,
       isPreapproved: true,
+      originalName: this.originalName,
     });
   }
 
@@ -331,6 +342,9 @@ export class AgencyFunction {
       exported: this.exported,
       markers: this.markers,
       isPreapproved: this._isPreapproved,
+      // The registry never learns the new name, so serialization keeps the
+      // registered name for revival (see FunctionRefReviver).
+      originalName: this.originalName,
     });
   }
 
--- a/packages/agency-lang/lib/runtime/agencyFunction.ts
+++ b/packages/agency-lang/lib/runtime/agencyFunction.ts
@@ -81,7 +81,7 @@ export type AgencyFunctionOpts = {
   exported?: boolean;
   markers?: ToolMarkers;
   isPreapproved?: boolean;
-  originalName?: string;
+  registeredName?: string;
 };
 
 export class AgencyFunction {
@@ -99,17 +99,18 @@ export class AgencyFunction {
   readonly exported: boolean;
   readonly markers: ToolMarkers;
   private readonly _isPreapproved: boolean;
-  /** The name this function's registered ancestor carries in the function
-   *  registry. `.rename()` changes `name` but not this, and every other
-   *  derivation (`.partial()`, `.describe()`, `.preapprove()`) carries it
-   *  forward — so a serialized ref to any derived copy can find its
-   *  registry entry again. Equal to `name` for functions never renamed. */
-  readonly originalName: string;
+  /** The key this function's registered ancestor is stored under in the
+   *  function registry. `.rename()` changes `name` but not this, and the
+   *  other derivations carry it forward — `.partial()` and `.preapprove()`
+   *  directly, `.describe()` via `withToolDefinition()` — so a serialized
+   *  ref to any derived copy can find its registry entry again. Equal to
+   *  `name` for functions never renamed. */
+  readonly registeredName: string;
 
   constructor(opts: AgencyFunctionOpts) {
     this.name = opts.name;
     this.module = opts.module;
-    this.originalName = opts.originalName ?? opts.name;
+    this.registeredName = opts.registeredName ?? opts.name;
     this._fn = opts.fn;
     this.params = opts.params;
     this.toolDefinition = opts.toolDefinition;
@@ -154,7 +155,7 @@ export class AgencyFunction {
       exported: this.exported,
       markers: this.markers,
       isPreapproved: this._isPreapproved,
-      originalName: this.originalName,
+      registeredName: this.registeredName,
     });
   }
 
@@ -258,7 +259,7 @@ export class AgencyFunction {
       exported: this.exported,
       markers: this.markers,
       isPreapproved: this._isPreapproved,
-      originalName: this.originalName,
+      registeredName: this.registeredName,
     });
   }
 
@@ -305,7 +306,7 @@ export class AgencyFunction {
       exported: this.exported,
       markers: this.markers,
       isPreapproved: true,
-      originalName: this.originalName,
+      registeredName: this.registeredName,
     });
   }
 
@@ -344,7 +345,7 @@ export class AgencyFunction {
       isPreapproved: this._isPreapproved,
       // The registry never learns the new name, so serialization keeps the
       // registered name for revival (see FunctionRefReviver).
-      originalName: this.originalName,
+      registeredName: this.registeredName,
     });
   }
 
--- a/packages/agency-lang/lib/runtime/revivers/functionRefReviver.test.ts
+++ b/packages/agency-lang/lib/runtime/revivers/functionRefReviver.test.ts
@@ -125,15 +125,15 @@ describe("FunctionRefReviver", () => {
     });
 
     it("near-miss block names revive to the generic stub, not the block stub", async () => {
+      // Pins the STRICT block-name predicate: these two must not match
+      // isBlockName, and the generic-stub message (vs "before replay
+      // rebound it") proves which branch they took.
       reviver.registry = {};
       for (const name of ["__blockish", "__block_"]) {
         const stub = reviver.revive({ name, module: "test.agency" });
-        const outcome = await stub
-          .invoke({ type: "positional", args: [] })
-          .then((v: unknown) => v, (e: unknown) => e);
-        const msg =
-          outcome instanceof Error ? outcome.message : JSON.stringify(outcome);
-        expect(msg).toContain("never loaded its module");
+        await expect(
+          stub.invoke({ type: "positional", args: [] }),
+        ).rejects.toThrow(/never loaded its module/);
       }
     });
 
@@ -579,16 +579,16 @@ describe("renamed functions round-trip (#652)", () => {
       __nativeType: "FunctionRef",
       name: "wikipedia_search",
       module: "stdlib/wikipedia.agency",
-      originalName: "search",
+      registeredName: "search",
       toolDescription: "Search Wikipedia",
     });
   });
 
-  it("serialize omits originalName for never-renamed functions", () => {
+  it("serialize omits registeredName for never-renamed functions", () => {
     const reviver = new FunctionRefReviver();
     const registry: Record<string, AgencyFunction> = {};
     const fn = makeRegistered(registry);
-    expect(reviver.serialize(fn)).not.toHaveProperty("originalName");
+    expect(reviver.serialize(fn)).not.toHaveProperty("registeredName");
   });
 
   it("a renamed function revives via the registered name and keeps the new name", async () => {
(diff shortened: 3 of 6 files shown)


packages/agency-lang/lib/runtime/revivers/functionRefReviver.ts
```diff
@@ -49,8 +55,16 @@ export class FunctionRefReviver implements BaseReviver<AgencyFunction> {
     }
     const name = value.name as string;
     const module = value.module as string;
+    const lookupName =
+      typeof value.originalName === "string" ? value.originalName : name;
 
-    const original = this.findInRegistry(name, module);
+    const found = lookupInRegistry(this.registry, lookupName, module);
+    if (!found) {
+      return this.reviveMiss(name, module, value);
```
egonSchiele, line 63: **A miss for an ordinary function is now completely silent until someone invokes the stub — and nothing may ever invoke it.**

The old eager throw was wrong in the `deepClone` case, but it did one useful thing: it told you at restore time that a checkpoint referred to a function this process cannot resolve. After this change, a genuine "function was renamed or removed since this state was serialized" — a stale checkpoint, a refactor — restores clean and then either fails much later at a confusing call site, or silently never fails at all because the ref only ever gets re-serialized.

That is the same failure shape this PR's own description is about: an error that gets laundered into a success. The stub itself only calls `emitFunctionRefMissError` from inside `fn`, so a miss that is never invoked leaves no trace anywhere.

Suggestion: emit a statelog event (warning-level, or the same `runtimeError` channel) *at revive time* in `reviveMiss` for the `makeUnresolvedFunctionStub` branch, not only at invoke time. The block and lifted-callback branches genuinely expect misses, so keep them quiet; the ordinary-function branch does not, and the whole point of `emitFunctionRefMissError`'s doc comment is that these things must be findable after the fact.

If that's too noisy in the parent-holds-child-checkpoint case, an alternative is to record the miss on the stub and emit only when it survives to a *restore* rather than a clone — but I'd start with the simple unconditional emit.

packages/agency-lang/lib/runtime/revivers/functionRefReviver.ts
```diff
@@ -224,13 +247,50 @@ function makeLazyCallbackRef(
...
+    module,
+    originalName:
+      typeof value.originalName === "string" ? value.originalName : name,
+    fn: () => {
+      const msg =
+        `Function "${name}" from module "${module}" crossed a serialization ` +
+        `boundary into a process that never loaded its module, and was ` +
+        `invoked there (or the function was removed since this state was ` +
+        `serialized).`;
+      emitFunctionRefMissError(name, msg);
```
egonSchiele, line 295: Small correctness note on the stub, not a blocker: it drops `exported` and `markers` (`destructive` / `idempotent`). Today that's invisible because `serialize()` never writes those fields, so `serialize(revive(x)) === x` still holds. But `markers` is retry-safety metadata (see `destructive-marker-retry-gate-bug`), and if it ever becomes part of the serialized ref, this stub is one of three places that must be updated in lockstep. Worth a one-line note here saying the stub mirrors exactly what `serialize()` writes, so the next person adding a field knows to come back.

packages/agency-lang/lib/runtime/agencyFunction.ts
```diff
@@ -98,10 +99,17 @@ export class AgencyFunction {
   readonly exported: boolean;
   readonly markers: ToolMarkers;
   private readonly _isPreapproved: boolean;
+  /** The name this function's registered ancestor carries in the function
+   *  registry. `.rename()` changes `name` but not this, and every other
+   *  derivation (`.partial()`, `.describe()`, `.preapprove()`) carries it
+   *  forward — so a serialized ref to any derived copy can find its
+   *  registry entry again. Equal to `name` for functions never renamed. */
+  readonly originalName: string;
```
egonSchiele: Naming nit: `originalName` already means something quite different across the compiler — in `symbolTable.ts`, `compilationUnit.ts`, `resolveReExports.ts`, and `lsp/semantics.ts` it is the pre-alias name of an *imported symbol*. This one is "the key this function is registered under." Different world, so it isn't ambiguous in practice, but `registeredName` would say exactly what it is and would make the reviver code read almost as documentation (`lookupInRegistry(registry, registeredName, module)`).

Also, the doc comment lists `.describe()` as carrying it forward. That's true, but only indirectly — `describe()` routes through `withToolDefinition()`. Worth saying so, otherwise a future reader greps for `originalName` in `describe()`, finds nothing, and assumes a bug.

packages/agency-lang/lib/runtime/revivers/functionRefReviver.test.ts
```diff
@@ -117,14 +124,17 @@ describe("FunctionRefReviver", () => {
       expect(msg).toContain("before replay rebound it");
     });
 
-    it("non-block names still throw eagerly on a registry miss", () => {
+    it("near-miss block names revive to the generic stub, not the block stub", async () => {
       reviver.registry = {};
```
egonSchiele, line 131: This assertion is weaker than the ones around it. The `.then(v => v, e => e)` dance means a stub that *resolves* instead of throwing still reaches the `expect`, and only fails on the string match against `JSON.stringify(value)` — a confusing failure message for the real bug (stub didn't throw). Every other new test here uses `await expect(...).rejects.toThrow(/never loaded its module/)`, which fails loudly and legibly. Suggest matching that.

On the substance: good call keeping both `__blockish` and `__block_` in the test. The old test's purpose was to pin the *strict* block-name predicate, and it would have been easy to lose that when the throw became a stub. The generic-stub message still distinguishes it from the block stub, so the predicate stays pinned.

packages/agency-lang/foo.agency
```diff
@@ -1,6 +1,15 @@
+import { researcherAgent } from "std::agents/researcher"
 import { now } from "std::date"
 node main() {
-  const start = now()
-  sleep(4)
-  print(elapsedTime(since: start))
+  handle {
+    const prompt = "What are the top news stories today?"
+    const result = researcherAgent(task: prompt)
```
egonSchiele: As you flagged in the description — drop this hunk. It's a scratch file, it has nothing to do with the fix, and leaving it in makes `git log` on this file useless for whoever comes next.

egonSchiele (commented):
## Overview

Two related fixes to `FunctionRefReviver`:

1. **Renames survive serialization.** `AgencyFunction` gains `originalName` (the name its registered ancestor holds in the registry), threaded through all four derivations. `serialize()` writes it when it differs from `name`; `revive()` looks up the original and re-applies `.rename()`.
2. **A registry miss no longer throws during revive.** Ordinary misses now produce a tripwire stub that survives re-serialization and fails only if invoked.

The diagnosis in the description is the good part of this PR. `revive()` running during *writes* — because `State.toJSON` deep-clones through `JSON.parse` with revivers installed — is genuinely non-obvious, and it explains why the crash landed somewhere so far from its cause. The fix follows the shape already established for `.partial()` and `.describe()` rather than inventing a new mechanism, which is right.

I verified the interesting reconstruction paths by hand and they hold:

- **rename + describe**: `rename()` sets `toolDefinition.name`, `describe()` (via `withToolDefinition`) preserves it. Round-trips.
- **rename + partial, either authoring order**: `revive()` always reconstructs rename-then-partial. Since `rename` touches only the name and `partial`'s `stripBoundParams` operates on identical description text either way, both orders converge on the same result.
- **preapprove in any position**: `serialize` records a boolean and `revive` applies it last; `preapprove()` is idempotent and order-independent with respect to `rename`.
- **The actual `search.agency` case** (`lib/agents/agency-agent/lib/search.agency:339-340`) — two different functions both renamed to `web_search` in the same module — now resolves correctly, because they carry distinct `originalName`s. That was previously unrecoverable and is the strongest argument that the design is right.

## Main concerns

**1. The silent-miss regression (inline comment on `functionRefReviver.ts`).** Not throwing during a clone is correct. Not *recording* anything during a restore is a step backwards. `emitFunctionRefMissError` fires only from inside the stub's body, so a miss that is never invoked leaves no trace at all. Given this PR exists because an error got laundered into a false pass, I'd emit a statelog event at revive time for the ordinary-function branch.

**2. Test coverage does not reach the reported bug.** All six new tests are unit tests on the reviver, and they're well-targeted — the `deepClone`-shaped test (`serialize(revive(x)) === x` through a miss) is exactly the right invariant to pin, and the "register the real function afterwards, assert it still throws" test correctly pins the stub-vs-lazy-ref distinction.

But nothing exercises the path that actually broke: a renamed tool held in state that goes through `State.toJSON` at a guard-trip checkpoint. The unit tests would all have passed against a fix that got the `State` integration wrong. Per `CLAUDE.md` these need no LLM call, so a test in `tests/agency/` that puts a `.rename()`d tool in state and forces a checkpoint write would close the loop and stop this regressing. That's the one gap I'd want filled before merge.

## Smaller notes

- **Legacy checkpoints.** A checkpoint written before this PR that holds a renamed ref has no `originalName`, so it misses and now gets a stub instead of a throw. Probably the outcome you want, but it means old checkpoints with renamed tools silently degrade rather than announce themselves — another reason for the revive-time statelog event.
- **Opportunity, not a defect:** `saveDraftTool.ts:41` documents that a `.rename()`d stdlib `saveDraft` is not recognized, because `matches` compares `fn.name`. `originalName` now makes that fixable. Out of scope here, but worth an issue while the context is fresh.
- `foo.agency` should come out (you flagged it yourself).

## Things I checked that are fine

- No other `new AgencyFunction(...)` site needs `originalName` — `lib/stdlib/agency.ts:71` (`_callback`) and the two existing stubs are all correctly left at the default.
- `AgencyFunction.create()` keys the registry by `opts.name`, which stays consistent because nothing registers a renamed copy.
- `lookupInRegistry`'s linear-scan fallback matches on `entry.name`, and no renamed copy ever enters the registry, so the new lookup path can't accidentally match one.
- No security or performance implications; the added work is one string comparison per serialize and one per revive.

Nice piece of debugging. My only merge blocker is the missing integration-level regression test; the statelog point is a strong recommendation.

packages/agency-lang/lib/runtime/revivers/functionRefReviver.ts
egonSchiele: Fixed in dd52354f5: reviveMiss now emits the statelog runtimeError at revive time for the ordinary-function branch (message says it revived to a stub that throws if invoked). Block and callback branches stay quiet. The invoke-time emit still fires on top if the stub is ever called.

packages/agency-lang/lib/runtime/revivers/functionRefReviver.ts
egonSchiele: Added the lockstep note in dd52354f5: the stub doc comment now states its fields mirror exactly what serialize() writes, with exported/markers named as deliberately absent.

packages/agency-lang/lib/runtime/agencyFunction.ts
egonSchiele: Renamed to registeredName throughout in dd52354f5, and the doc comment now says describe() carries it via withToolDefinition().

packages/agency-lang/lib/runtime/revivers/functionRefReviver.test.ts
egonSchiele: Converted to rejects.toThrow(/never loaded its module/) in dd52354f5, with a comment keeping the strict-predicate purpose pinned.

packages/agency-lang/foo.agency
egonSchiele: Dropped in dd52354f5 - foo.agency is back to the main version.

egonSchiele:
All review items addressed in dd52354f5:

- **Integration regression test (the merge blocker):** `tests/agency/renamed-ref-checkpoint` holds a `.rename()`d ref and a `.partial()`-bound ref in frame locals, forces a checkpoint write via a bash interrupt (the State.toJSON → deepClone → revive path that crashed), and invokes both revived refs after resume. Verified in both directions: against the pre-fix reviver it fails with the original "not found in registry" error (twice); with the fix it passes. No LLM call.
- **Revive-time statelog for ordinary misses:** emitted unconditionally from the `makeUnresolvedFunctionStub` branch of `reviveMiss`; block and callback branches stay quiet. A stale checkpoint now announces itself at restore even if the stub is never invoked.
- `originalName` → `registeredName` throughout, doc comment notes `describe()` routes via `withToolDefinition()`.
- Stub doc comment states its fields mirror exactly what `serialize()` writes (`exported`/`markers` deliberately absent) so the next serialized field gets added in lockstep.
- Near-miss block-name test now uses `rejects.toThrow`, purpose comment kept.
- `foo.agency` hunk dropped (also covers the Copilot note about the unused import and unintended spend).
- The saveDraft rename opportunity is filed as #654.

PR #655, opened 2026-07-23
Hoist helper calls into skippable steps so in-process resume cannot desync frames

egonSchiele:
Fixes the in-process resume frame desync: interactively approving a tool interrupt could corrupt the resume and end in `400 No tool output found for function call` (reproduced with `bisect-a.agency` under `run -i`).

## The bug

Resume replay hands saved frames to functions positionally (`StateStack.getNewState()` in deserialize mode is a bare `shift()`). The statement in progress at a pause re-runs its whole body, including helper calls that had already completed and were therefore not in the checkpoint — `llm(msg, llmOptions(...))` re-ran `llmOptions()`. Each such re-run consumed a saved frame belonging to a still-live function. The victim (`runPrompt` in the reported case) then got a blank frame, restarted from its first request, and sent a thread that still ended with an unanswered tool call. The test suite never saw it because tests pass tools as literals (`tools: [myTool]`), while every stdlib agent uses the helper-call shape.

## The fix, two independent halves

**1. The `hoistCalls` preprocessor pass** (`lib/preprocessors/hoistCalls.ts`): every helper call in an unconditionally-evaluated position becomes its own `const __hoist_N = ...` statement. Statements are steps, step results serialize on `__stack.locals`, so on resume the helper's completed step is skipped and its value read back — it never re-runs and the frame queue stays aligned. Zero builder changes: a new statement is automatically a new step.

Position rulings are data (a table the traversal consults), statement recursion rides `bodySlots`/house walkers, and everything copies (no AST mutation). Highlights:
- `while` conditions with calls rewrite to `while (true) { temps; if (cond) { body } else { break } }` — the worst pre-pass position, since `Runner.whileLoop` re-evaluated the condition once per completed iteration on resume.
- Opaque boundaries: `try` operands and `catch` expressions (thunk boundaries), short-circuit right sides and if-expression branches (conditional), pipe stages (already memoized + failure-gated; the pipe input hoists), `with`-modified statements, handler bodies (plain JS, cannot pause), module-level initializers (init-topsort).
- Lifted blocks (comprehensions, fork branches) hoist within, never across; temp counters are per frame-owning scope because frame locals are flat.
- One accepted relaxation: `for (x in async getItems())` now compiles (the async call hoists above the loop), where `validateNoAsyncInLoops` used to reject it. Recorded in `docs/dev/hoist-calls.md`.

**2. The scope-name tripwire** (runtime + codegen): frames are stamped with their owner at CLAIM time — generated function/node/block preambles plus hand-written claims in `runPrompt` and `withResumableScope` — and a mismatched claim on replay throws `Resume desync: function "X" tried to claim the saved state of "Y"` plus a statelog event (throws convert to Failures at def boundaries and can be laundered; the event survives). Claiming and running are distinct events by design: `finalize` closures build a second Runner on their container's frame and must not stamp, which is why the check does NOT live in the Runner constructor. Residual shapes the pass cannot cover (calls nested inside conditional positions, mid-chain method calls) now fail loudly instead of silently.

## Verified both directions

With the pass unwired (one-time manual check): the flagship regression fixture fails with exactly the predicted theft —

```
Resume desync: function "myOptions" tried to claim the saved state of "runPrompt".
```

— and the eval-order fixture prints the identical sequence wired and unwired (the neutrality evidence). With the pass wired, all fixtures pass. A permanent unit test pins that the flagship fixture's llm options argument really is a hoisted temp, so the fixture cannot silently stop exercising the bug.

## Tests

- 23 unit tests on the pass (positions, boundaries, ordering, per-scope numbering, idempotency, purity, the fixture shape pin) + 5 on the tripwire + 3 pinning claim emission in generated code (including "finalize closures do NOT claim").
- 11 agency execution fixtures under `tests/agency/hoist/`: both resume-regression shapes (argument-position and spread), eval-order neutrality (short-circuit skip included), while rewrite at runtime with a comparison condition, user break/continue inside a rewritten while, a pause inside a while condition (single-execution-per-iteration pinned), a pause inside a hoisted temp's own step, failure propagation through a temp, FunctionRef round-trips with `.partial()` + `.rename()` tools live in the pausing frame (depends on #653, already merged), a pause inside a loop body, and finalize-no-tripwire.
- Full unit suite: 9064 passed, 0 failed. `lint:structure` clean.

## Commit structure

Mechanical fixture regenerations are isolated in their own commits (one for the claim emission, one for the pass's statement shapes) so the code commits stay readable. Note the fixture suite string-compares generated output and never executes it; the execution fixtures and CI agency suite are the real gates.

Spec: `docs/superpowers/specs/2026-07-22-hoist-calls-resume-safety-design.md` (in the repo root docs, untracked). Docs added: `docs/dev/hoist-calls.md` + a paragraph in `docs/dev/interrupts.md`.

### DIFF
--- a/packages/agency-lang/lib/preprocessors/hoistCalls.ts
+++ b/packages/agency-lang/lib/preprocessors/hoistCalls.ts
@@ -1,0 +1,408 @@
+/**
+ * hoistCalls — rewrite every statement so helper calls become their own
+ * `const __hoist_N = ...` statements.
+ *
+ * Why: resume replay re-executes the statement that was in progress at a
+ * pause, including helper calls that already completed. Those helpers
+ * re-claim frames from the positional restore queue and desync it (the
+ * still-live owner gets a blank frame and re-issues work). A hoisted
+ * helper is its own runner step: completed steps are skipped on resume
+ * and their value is read back from `__stack.locals`, so the helper
+ * never re-runs. Spec:
+ * docs/superpowers/specs/2026-07-22-hoist-calls-resume-safety-design.md
+ *
+ * The invariant this pass establishes: after preprocessing, a statement
+ * re-executes no completed frame-pushing call on resume — the only call
+ * remaining in unconditionally-evaluated position is the statement's own
+ * tail. Tails stay deliberately: hoisting them uniformly would need a
+ * graph-node-call exclusion (node calls are control flow and THROW in
+ * value position — typescriptBuilder's processNodeCall), while at tail
+ * position they need no rule at all.
+ *
+ * Statement-body recursion is driven by `bodySlots` (the single source
+ * of truth for which fields hold statements); only the expression
+ * interior uses a generic child walk. Rulings are data, not control
+ * flow — see RULINGS / OPERATOR_RULINGS below. Everything copies;
+ * parsed AST is never mutated (in-place mutation burned this repo via
+ * the parse cache clone-on-read fix).
+ *
+ * Temp naming: `__hoist_N`, ONE counter per frame-owning scope
+ * (function, node, lifted block, fork branch), shared by every nested
+ * statement list inside it. Frame locals are flat, so per-list
+ * numbering would let a loop-body temp clobber the temp a loop
+ * iterable re-reads on resume. Blocks own frames (blockSetup.mustache
+ * pushes one), so block bodies restart at 0. Finalize bodies run on
+ * the CONTAINER's frame (bodySlots documents this) and therefore share
+ * the container's counter. Seeding scans the scope for existing
+ * `__hoist_N` names and starts above the max — this seeding is the
+ * collision protection; there is deliberately no lint rule (no other
+ * compiler-reserved prefix has one).
+ *
+ * Known residuals (tripwire territory, not silent corruption): calls
+ * nested inside opaque positions (short-circuit right sides, catch
+ * expressions, try operands, `with`/`static` wrappers, mid-chain method
+ * calls inside a hoisted chain), and block bodies nested inside opaque
+ * expressions.
+ */
+import type { AgencyNode, AgencyProgram } from "../types.js";
+import { bodySlots } from "../utils/bodySlots.js";
+import { createKeyword } from "../types/keyword.js";
+
+type Counter = { n: number };
+
+type Extraction = { temps: AgencyNode[]; expr: any };
+
+/** Expression-node rulings the walker consults before descending.
+ *  "opaque" = never descend, hoist nothing inside (each row says why).
+ *  Unlisted expression node types default to a generic child walk. */
+const NODE_RULINGS: Record<string, "opaque"> = {
+  // The whole operand compiles into the __tryCall thunk; hoisting an
+  // argument out would run it outside the catch boundary and let its
+  // throw escape uncaught.
+  tryExpression: "opaque",
+};
+
+/** binOpExpression rulings by operator.
+ *  - "catch": opaque. The fallback runs only on Failure (__catchResult);
+ *    the left side is excluded too, purely for boundary simplicity — a
+ *    temp for it would be semantically identical.
+ *  - "|>": input-only. Stages are already resume-safe (memoized
+ *    __pipe_result_<path>) AND conditional (__pipeBind skips them on
+ *    Failure); the input is evaluated inline at statement level, outside
+ *    the memoization, so it hoists.
+ *  - "&&" / "||" / "??": left-only. The right side may never execute.
+ *    The first inline call there is resume-aligned by construction
+ *    (nothing before it consumes a frame on replay), so only calls
+ *    NESTED under it remain residual.
+ */
+const OPERATOR_RULINGS: Record<string, "opaque" | "leftOnly"> = {
+  catch: "opaque",
+  "|>": "leftOnly",
+  "&&": "leftOnly",
+  "||": "leftOnly",
+  "??": "leftOnly",
+};
+
+const SKIP_TYPES = ["comment", "newLine"];
+
+export function hoistCallsInProgram(program: AgencyProgram): AgencyProgram {
+  // Only function and graphNode bodies. Module-level initializers are
+  // excluded by construction: they run through the per-variable init
+  // dependency graph (docs/dev/init-topsort.md), cannot pause, and a
+  // synthesized variable would be invisible to the topsort.
+  return {
+    ...program,
+    nodes: program.nodes.map((node) => {
+      if (node.type === "function" || node.type === "graphNode") {
+        return { ...node, body: hoistCallsInScope((node as any).body) } as AgencyNode;
+      }
+      return node;
+    }),
+  };
+}
+
+/** Rewrite one frame-owning scope's statements. Pass `counter` when
+ *  recursing into a nested statement list of the SAME frame (loop and
+ *  if bodies, finalize bodies, thread blocks); omit it at a frame
+ *  boundary (function, node, lifted block, fork branch) so numbering
+ *  restarts against that frame's own flat locals. */
+export function hoistCallsInScope(
+  body: AgencyNode[],
+  counter?: Counter,
+): AgencyNode[] {
+  const c = counter ?? { n: seedCounter(body) };
+  const out: AgencyNode[] = [];
+  for (const stmt of body) {
+    if (!stmt || typeof stmt !== "object" || SKIP_TYPES.includes(stmt.type)) {
+      out.push(stmt);
+      continue;
+    }
+    out.push(...rewriteStatement(stmt, c));
+  }
+  return out;
+}
+
+/** Start numbering above any __hoist_N already present in the scope
+ *  (user-declared or from an earlier run of the pass). This scan is the
+ *  collision protection; it also makes the pass idempotent. */
+function seedCounter(body: AgencyNode[]): number {
+  let max = -1;
+  for (const m of JSON.stringify(body).matchAll(/__hoist_(\d+)/g)) {
+    const n = Number(m[1]);
+    if (n > max) max = n;
+  }
+  return max + 1;
+}
+
+// eslint-disable-next-line max-lines-per-function -- one case per statement kind; splitting would scatter the dispatch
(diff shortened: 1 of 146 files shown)


packages/agency-lang/lib/preprocessors/hoistCalls.ts
```diff
@@ -0,0 +1,408 @@
...
+    }
+    case "withModifier":
+    case "staticStatement":
+      // Opaque: hoisting out of a `with` region would move the call
+      // outside the modifier's approval scope; `static` initializers
+      // belong to init-topsort. Both wrap exactly one statement
+      // (bodySlots `single`), which a multi-statement rewrite would
+      // break anyway.
+      return [stmt];
+    default:
```
egonSchiele, line 283: **Statements with expression positions that aren't in this switch fall through to `recurseSlots`, which only walks statement bodies — their expressions are never hoisted.**

Three real cases, verified by compiling on this branch:

1. **`gotoStatement`** (`{ nodeCall: FunctionCall }`). I compiled

```agency
node main() {
  goto second(wrap(side(1)))
}
```

against this branch and the generated output contains **zero** `__hoist_` temps. `side(1)` still runs inside the same step as `wrap(...)`, so a pause in a later argument re-runs it on resume — the exact desync this pass exists to remove.

2. **`matchYield`** — match-expression arm values, produced by `lowerPatterns` at parse time. `_ => f(g())` never hoists `g()`.

3. **`messageThread`** named-arg expressions. `bodySlots` gives it body recursion, so the bodies are fine, but `thread(label: makeLabel())` leaves the call inline (`lib/utils/node.ts:436-437` documents these named-arg expressions).

This is the same drift the `bodySlots` header warns about, one level over: statement *bodies* now ride a shared enumeration, but statement *expression positions* are hand-listed here, so a node type added later silently gets no hoisting. `expressionChildren` (`lib/utils/node.ts:51`) is the documented twin — it has no writer, which is presumably why it wasn't used, but at minimum these three cases should be added to the switch and the residuals list updated if any are deliberately out of scope.

packages/agency-lang/lib/preprocessors/hoistCalls.ts
```diff
@@ -0,0 +1,408 @@
...
+      }
+      if (entry?.kind === "index" && entry.index) {
+        const inner = walk(entry.index, counter, true);
+        temps.push(...inner.temps);
+        if (inner.expr !== entry.index) hasCall = hasCall || false;
+        return { ...entry, index: inner.expr };
+      }
+      return entry;
+    });
+    const rebuilt = { ...node, chain };
```
egonSchiele: **The chain's `base` is never walked, so calls nested inside it stay in the temp's own step.**

`node.base` is passed straight through into `rebuilt`. Compiled on this branch:

```agency
const x = outer(mk(inner()).field)
```

```js
__stack.locals.__hoist_0 = (await await __call(mk, {
  args: [await __call(inner, { ... })]     // still inline, same step
})).field;
```

So `inner()` and `mk()` share one step. If `mk` pauses after `inner()` completed, the temp's step re-runs on resume, `inner()` re-claims a frame, and we are back to the original bug — inside a statement the pass has already rewritten, which makes it harder to spot than the shapes it fixes.

Hoisting the base is not the mid-chain splitting the comment above rules out: `getConfig().model` becoming `const __h = getConfig()` then `__h.model` is the natural boundary, and the base's *arguments* need walking regardless. If the unit-hoist is deliberate for chains, the base's arguments should still be extracted before the chain is hoisted.

packages/agency-lang/lib/preprocessors/hoistCalls.ts
```diff
@@ -0,0 +1,408 @@
...
+          counter,
+          false,
+        );
+        temps.push(...inner.temps);
+        return { ...entry, functionCall: inner.expr };
+      }
+      if (entry?.kind === "index" && entry.index) {
+        const inner = walk(entry.index, counter, true);
+        temps.push(...inner.temps);
+        if (inner.expr !== entry.index) hasCall = hasCall || false;
```
egonSchiele: `hasCall = hasCall || false` is a no-op — it can only ever leave `hasCall` unchanged. Either an index expression containing a call was meant to force the chain to hoist (in which case this wants `hasCall = true` when `inner.temps.length > 0`, or when the index itself is a call), or the line should go. As written a reader has to stop and work out that it does nothing.

packages/agency-lang/lib/preprocessors/hoistCalls.ts
```diff
@@ -0,0 +1,408 @@
...
+  }
+  return out;
+}
+
+/** Start numbering above any __hoist_N already present in the scope
+ *  (user-declared or from an earlier run of the pass). This scan is the
+ *  collision protection; it also makes the pass idempotent. */
+function seedCounter(body: AgencyNode[]): number {
+  let max = -1;
+  for (const m of JSON.stringify(body).matchAll(/__hoist_(\d+)/g)) {
```
egonSchiele: Two things about seeding by regex over `JSON.stringify(body)`:

- It matches `__hoist_N` anywhere in the serialized subtree, including inside string literals. `print("__hoist_5")` in user code bumps the counter to 6. Harmless, but it means the numbering is not quite the "scan for declarations" the comment describes.
- It stringifies the whole subtree once per frame-owning scope, and nested block scopes re-stringify their own subtree, so the cost is quadratic in nesting depth on every compile. Probably invisible in practice; worth a moment's thought since this runs on every build of every file.

A walk that collects `assignment` nodes whose `variableName` matches would be exact and linear. Not blocking.

egonSchiele (commented):
Reviewed the pass, the tripwire, and the fixtures. Verified claims by compiling probe files against this branch rather than reading only.

**This is in good shape.** The rulings-as-data table, `bodySlots` for statement recursion, the pure `{temps, expr}` extractor, per-frame-scope counters, and the `if/else`-break `while` rewrite (no synthesized negation) all landed as designed. The tripwire correctly separates claiming from running — `claimFrameForScope` at the claim sites with the Runner constructor left alone, and a `finalize`-no-tripwire fixture pinning it. Verifying the red direction with the pass unwired, and getting exactly `Resume desync: function "myOptions" tried to claim the saved state of "runPrompt"`, is the strongest evidence in the PR: it proves the pass and the tripwire interlock on the real bug.

Two coverage gaps I'd want closed or explicitly documented before merge; both are the original bug still reachable, not new breakage.

1. **Statements whose expressions the dispatcher doesn't handle** — `gotoStatement`, `matchYield`, `messageThread` named args — get body recursion but no expression hoisting. Compiled `goto second(wrap(side(1)))` on this branch: zero `__hoist_` temps.

2. **Value-access chain bases are never walked.** `outer(mk(inner()).field)` hoists the chain as a unit but leaves `inner()` inside the temp's own step, so a pause in `mk` re-runs `inner()` on resume.

Neither is a regression — both shapes are equally broken on main — so shipping with them listed in the residuals section of `docs/dev/hoist-calls.md` is defensible. What isn't defensible is leaving them unlisted, since the doc currently reads as if the only residuals are conditional positions and mid-chain method calls.

Smaller notes inline: one dead line in the chain walker, and a nit on the counter-seeding regex.

Nothing to say about the fixture churn — isolating the two mechanical regenerations into their own commits made the code commits readable, and the PR body is right that those fixtures string-compare rather than execute.

packages/agency-lang/lib/preprocessors/hoistCalls.ts
egonSchiele: Fixed in bfd228da8. All three cases now hoist: gotoStatement walks the node call with the tail withheld (arguments hoist, the transfer stays), matchYield extracts like returnStatement (temps land inside the arm body, which keeps them conditional-correct), and messageThread walks its five named-arg expressions with temps before the statement (halted/skip paths skip temp steps and thread step together, matching the lazy opts-thunk semantics). Each has a unit test, including the exact goto shape you compiled. On the drift point: statement bodies ride bodySlots, and I agree the expression positions are still a hand list — extending expressionChildren with writers is the structural fix, but it is a wider refactor than this PR should carry; the three known cases are closed and the residuals doc no longer implies the list is complete.

packages/agency-lang/lib/preprocessors/hoistCalls.ts
egonSchiele: Fixed in bfd228da8: the base now walks like any expression, so a call base hoists into its own step (getConfig().model becomes const __h = getConfig() then __h.model) and its arguments extract with it. The unit-hoist now keys only on remaining chain method calls, which are the one residual left in chains — docs updated to say exactly that. Your outer(mk(inner()).field) probe shape is a unit test: temps are [inner, mk] and the chain applies .field to the ref.

packages/agency-lang/lib/preprocessors/hoistCalls.ts
egonSchiele: Gone in bfd228da8 — the base rewrite removed the whole hasCall-from-base bookkeeping; the flag is now chainHasMethodCall, set only where it means something.

packages/agency-lang/lib/preprocessors/hoistCalls.ts
egonSchiele: Both points fixed in bfd228da8: seeding now walks the tree for assignment declarations matching __hoist_N (string literals cannot bump the counter) in one linear visit per scope, instead of regexing serialized JSON.

packages/agency-lang/lib/preprocessors/hoistCalls.ts
```diff
@@ -0,0 +1,408 @@
...
+        temps.push(...inner.temps);
+        if (inner.expr !== entry.index) hasCall = hasCall || false;
+        return { ...entry, index: inner.expr };
+      }
+      return entry;
+    });
+    const rebuilt = { ...node, chain };
+    if (!hoistSelf || !hasCall) return { temps, expr: rebuilt };
+    const ref = makeTemp(rebuilt, temps, counter);
+    return { temps, expr: ref };
```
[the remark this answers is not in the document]
egonSchiele: Fixed in bfd228da8 together with the review comment above: bases walk (and hoist when they are calls), and slice start/end bounds now extract alongside index expressions and method-call arguments.

PR #658, opened 2026-07-23
Repair abandoned dangling tool calls on thread reopen

egonSchiele:
A turn that parks on an unanswered guard trip and is then abandoned left its
session thread ending on an assistant message with unanswered tool calls.
The provider rejects every later request on that thread with a 400, so one
abandoned turn destroyed the whole session (see the trace in
docs/superpowers/specs/2026-07-22-orphaned-tool-use-on-guard-abort-design.md).

This PR:
- restructures threadRepair.ts around one shared scan (unansweredToolCalls)
  and one shared append procedure; per-message debug labels now survive
  repair
- adds repairAbandonedTurn, which synthesizes results for dangling calls
  with wording that tells the next model the turn was interrupted, and
  repairReopenedThread, which Runner.thread() calls in one line on every
  reopen (session second+ entry and thread(continue: id); checkpoint
  resumes provably skip the branch, and a test pins that)
- gives MessageThread a repair generation (markRepaired/isNewerThan) and a
  restoreThreadForResume helper that refuses restoring a checkpoint taken
  before a repair, so a late answer to an abandoned turn fails loudly
  (with a statelog runtimeError emitted before the throw) instead of
  silently overwriting newer conversation
- makes turnBudgetHandler reject unowned std::guard trips instead of
  passing them into a park nobody will resume, with the first real test
  coverage of that handler (foreign-guard salvage, bare failure, own-label
  reject, non-guard pass-through), and labels the 14 inner stdlib guards so
  stop notices can name the budget that ran out

The existing user-cancel repair policy (needsThreadRepair) is deliberately
unchanged; the new rule is separate: a thread must be valid before new work
is appended to it.

Test notes: the end-to-end "no 400" assertion is not implementable with the
mock provider (it does not enforce tool_result pairing), so the runner
tests exercise the real seam with the real ThreadStore instead. The
own-label interactive prompt path of turnBudgetHandler stays untested (it
needs input() mocking); the non-interactive path covers the same label
matching. One fixture gotcha worth knowing: _isInteractive defaults to
true, so the own-label test calls setInteractive(false) explicitly.

### DIFF
--- a/packages/agency-lang/lib/runtime/runner.ts
+++ b/packages/agency-lang/lib/runtime/runner.ts
@@ -11,6 +11,7 @@ import { makeRedactReplacer, REDACTED } from "./redactForStatelog.js";
 import { __pipeBind } from "./result.js";
 import { nativeTypeReplacer, nativeTypeReviver } from "./revivers/index.js";
 import { runBatch } from "./runBatch.js";
+import { repairReopenedThread } from "./threadRepair.js";
 import type { SourceLocationOpts } from "./state/checkpointStore.js";
 import type { RuntimeContext } from "./state/context.js";
 import type { BranchState, State } from "./state/stateStack.js";
@@ -702,6 +703,13 @@ export class Runner {
         tid = threads.createSubthread(createMeta);
       } else {
         tid = threads.create(createMeta);
+      }
+      if (isResumption) {
+        // Reopened threads are repaired before new work lands; a
+        // checkpoint resume never reaches this branch (the frame-locals
+        // guard above skips the whole open side effect). The full safety
+        // argument lives on repairReopenedThread.
+        repairReopenedThread(threads.get(tid), this.ctx.statelogClient, tid);
       }
       this.frame.locals[threadKey] = tid;
       this.frame.locals[resumptionKey] = isResumption;
--- a/packages/agency-lang/lib/runtime/runner.ts
+++ b/packages/agency-lang/lib/runtime/runner.ts
@@ -709,6 +709,13 @@ export class Runner {
         // checkpoint resume never reaches this branch (the frame-locals
         // guard above skips the whole open side effect). The full safety
         // argument lives on repairReopenedThread.
+        //
+        // The createSubthread branch below needs no repair even though it
+        // clones the parent's messages: a dangling tail only ever exists
+        // on a thread whose turn parked and bailed out of the whole run,
+        // so no code is left executing that could take a subthread off it
+        // — and a later run must reopen the parent (through this branch)
+        // before it can be active again.
         repairReopenedThread(threads.get(tid), this.ctx.statelogClient, tid);
       }
       this.frame.locals[threadKey] = tid;
--- a/packages/agency-lang/lib/runtime/threadRepair.test.ts
+++ b/packages/agency-lang/lib/runtime/threadRepair.test.ts
@@ -1,7 +1,11 @@
 import { describe, it, expect } from "vitest";
 import * as smoltalk from "smoltalk";
 import { makeAbortCause } from "./errors.js";
+import { runInTestContext } from "./asyncContext.js";
+import { makeMockCtx } from "./__tests__/testHelpers.js";
 import { MessageThread } from "./state/messageThread.js";
+import { StateStack } from "./state/stateStack.js";
+import { ThreadStore } from "./state/threadStore.js";
 import {
   ABANDONED_CALL_TEXT,
   ABANDONED_TURN_TEXT,
@@ -218,7 +222,7 @@ describe("repairAbandonedTurn", () => {
 });
 
 describe("repairReopenedThread — the seam helper", () => {
-  it("repairs and emits threadRepaired with the slugged id and the call ids", () => {
+  it("repairs and emits threadRepaired with the raw thread id and the call ids", () => {
     const t = new MessageThread([
       smoltalk.userMessage("go"),
       asst("", [
@@ -229,7 +233,8 @@ describe("repairReopenedThread — the seam helper", () => {
     ]);
     const events: Array<{ threadId: string; toolCallIds: string[] }> = [];
     repairReopenedThread(t, { threadRepaired: (e) => { events.push(e); } }, "7");
-    expect(events).toEqual([{ threadId: "t7", toolCallIds: ["b"] }]);
+    // Raw id, matching threadCreated/threadResumed — consumers join on it.
+    expect(events).toEqual([{ threadId: "7", toolCallIds: ["b"] }]);
     expect(t.repairs).toBe(1);
   });
 
@@ -282,5 +287,28 @@ describe("restoreThreadForResume", () => {
     const live = new MessageThread([smoltalk.userMessage("hi")]);
     live.markRepaired();
     expect(restoreThreadForResume(live.toJSON(), live)).toBe(live);
+  });
+
+  it("emits a statelog runtimeError alongside the refusal throw", () => {
+    // The emit exists so the refusal survives Failure laundering — if it
+    // silently stops firing, that is the exact failure mode it defends
+    // against, so the wiring gets its own assertion (mirrors the
+    // threadRepaired test above).
+    const ctx = makeMockCtx();
+    const errors: Array<Record<string, unknown>> = [];
+    ctx.statelogClient.error = (e: Record<string, unknown>) => {
+      errors.push(e);
+    };
+    const live = new MessageThread([smoltalk.userMessage("hi")]);
+    const snap = live.toJSON();
+    live.markRepaired();
+    runInTestContext(ctx, new StateStack(), new ThreadStore(), () => {
+      expect(() => restoreThreadForResume(snap, live)).toThrow(
+        /repaired after this checkpoint/,
+      );
+    });
+    expect(errors).toHaveLength(1);
+    expect(errors[0].errorType).toBe("runtimeError");
+    expect(errors[0].functionName).toBe("restoreThreadForResume");
   });
 });
--- a/packages/agency-lang/lib/runtime/threadRepair.test.ts
+++ b/packages/agency-lang/lib/runtime/threadRepair.test.ts
@@ -2,7 +2,11 @@ import { describe, it, expect } from "vitest";
 import * as smoltalk from "smoltalk";
 import { makeAbortCause } from "./errors.js";
 import { MessageThread } from "./state/messageThread.js";
-import { markThreadCancelled, needsThreadRepair } from "./threadRepair.js";
+import {
+  markThreadCancelled,
+  needsThreadRepair,
+  unansweredToolCalls,
+} from "./threadRepair.js";
 
 /** Convenience builders for the markThreadCancelled repair-shape tests. */
 const asst = (text: string, toolCalls?: Array<{ id: string; name: string }>) =>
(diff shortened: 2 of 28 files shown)


packages/agency-lang/lib/runtime/threadRepair.ts
```diff
@@ -34,45 +95,134 @@ export function needsThreadRepair(cause: AbortCause | undefined): boolean {
...
+      "Cannot resume this turn: its conversation thread was repaired after " +
+      "this checkpoint was taken (the parked turn was abandoned and newer " +
+      "turns have run since). Restoring would overwrite the newer " +
+      "conversation, so it is refused.";
+    // Best-effort statelog BEFORE the throw: a throw converts to a Failure
+    // at the next def boundary, and Failures can get laundered into prose
+    // by the time a model or user sees them — the refusal must stay
+    // findable in the trace regardless. Same rationale and shape as
+    // claimFrameForScope in state/stateStack.ts.
+    agencyStore.getStore()?.ctx?.statelogClient?.error?.({
```
egonSchiele, line 225: This is correct today — I checked, `StatelogClient.error` takes exactly `{ errorType, message, functionName }` and `"runtimeError"` is a valid `errorType`. But nothing tests it, and with `getStore()?.ctx?.statelogClient?.error?.()` there are four places a rename or a refactor turns this into a silent no-op with every test still green.

That matters more here than usual, because the comment right above explains that this event exists precisely so the refusal survives being laundered into prose. If it stops firing, the failure mode is the one the comment is defending against, and nothing tells you.

`threadRepaired` already got this treatment (`repairs and emits threadRepaired with the slugged id and the call ids`), so the pattern is established — run the refusal test inside `runInTestContext` with a mock ctx and assert the event fires alongside the throw.

packages/agency-lang/lib/runtime/threadRepair.ts
```diff
@@ -34,45 +95,134 @@ export function needsThreadRepair(cause: AbortCause | undefined): boolean {
...
+ *  The ordering is pinned by tests. */
+export function restoreThreadForResume(
+  snapshot: MessageThreadJSON | smoltalk.MessageJSON[],
+  live: MessageThread | undefined,
+): MessageThread {
+  const restored = MessageThread.fromJSON(snapshot);
+  if (!live) return restored;
+  if (live.isNewerThan(restored)) {
+    const msg =
+      "Cannot resume this turn: its conversation thread was repaired after " +
```
egonSchiele, line 216: The message asserts a cause it cannot actually know: "the parked turn was abandoned and newer turns have run since."

That is true for the REPL case this PR is built around. It is not true for the case the design doc deliberately leaves open — a top-level `fork` arm reopening a session while another branch's turn is still parked on it (design doc, "A known limitation, accepted"). There, repair fires while a real resume is pending, this throw is what turns silent corruption into a loud failure (good, working as designed), but the person reading it is told their turn was abandoned when it was not. That sends them looking in the wrong place.

Suggest stating the observed fact and letting the usual cause be a hint rather than a claim — something like: the thread was repaired after this checkpoint was taken, so restoring would overwrite newer conversation; this usually means the parked turn was abandoned and the session continued without it.

packages/agency-lang/lib/runtime/runner.ts
```diff
@@ -703,6 +704,13 @@ export class Runner {
       } else {
         tid = threads.create(createMeta);
       }
+      if (isResumption) {
+        // Reopened threads are repaired before new work lands; a
+        // checkpoint resume never reaches this branch (the frame-locals
+        // guard above skips the whole open side effect). The full safety
+        // argument lives on repairReopenedThread.
+        repairReopenedThread(threads.get(tid), this.ctx.statelogClient, tid);
```
egonSchiele, line 719: A case I could not rule out, raising as a question rather than a finding.

`createSubthread` takes the `isResumption === false` path, so it never reaches this line. But `newSubthreadChild` seeds the child by cloning the parent's messages (`messageThread.ts`), so a subthread taken off a *damaged* parent starts life with the same dangling tail — and a subthread is never reopened, so nothing downstream will ever repair it. Its first provider call would 400 the same way the original bug did.

I tried to build a reachable path and could not: the only producer of a dangling tail is the interrupted-tool branch in `prompt.ts`, which bails out of the turn, so the run cannot proceed to create a subthread off it. So this may be genuinely unreachable.

But the rule this PR states is "a thread must be valid before new work is appended to it", and a subthread is new work on a copy. Either extend the repair to the `createSubthread` branch, or leave a line here saying why a cloned child cannot inherit a damaged tail — so the next person does not have to redo this reasoning.

Separate nit while you are in here: line 725 computes ``const slug = `t${tid}` `` a few lines below, and `repairReopenedThread` builds the same slug internally. Passing the slug (or hoisting it above this call) keeps one definition of how a thread id is presented.

packages/agency-lang/tests/agency/guards/unowned-guard-rejected.agency
```diff
@@ -0,0 +1,90 @@
...
+    return "failure:${isFailure(r)}"
+  } with turnBudgetHandler
+}
+
+// The handler's OWN label still takes the non-interactive reject path,
+// proving the label match is intact. _isInteractive defaults to true, so
+// the fixture must flip it or the handler would prompt via input() and
+// hang the harness.
+node ownLabelReject() {
+  setInteractive(false)
```
egonSchiele, line 81: Good catch on the default — I confirmed `_isInteractive` starts `true` and `isInteractive()` reads it straight through, so without this line the handler really would reach `input()`.

The thing to watch is that this is module-global state mutated inside one test node, and the other three nodes in this file read it. Nothing breaks today: `foreignGuardSalvage` and `foreignGuardBare` run before this and take the interactive branch, which only adds a `print`, and none of their assertions depend on it. But it makes the file order-sensitive — if the runner ever reorders or shares a module instance differently, those two silently start exercising the other branch instead, and no assertion notices.

Cheapest fix is to make each node state what it wants rather than inheriting it: call `setInteractive(...)` at the top of all four nodes. That also documents which branch each test is actually covering, which right now you have to infer from the default.

packages/agency-lang/tests/agency/guards/unowned-guard-rejected.agency
```diff
@@ -0,0 +1,90 @@
...
+  } with turnBudgetHandler
+}
+
+// Non-guard interrupts must still pass through to whoever is outside —
+// here, the test harness resolves it. Getting the early-return split
+// wrong would make the agent reject tool approvals, which would be far
+// worse than the bug being fixed.
+node nonGuardPassthrough() {
+  handle {
+    const r = interrupt("confirm")
```
egonSchiele: Nit: `r` is bound and never read. `interrupt("confirm")` on its own says the same thing, and an unused binding in a test that is specifically about pass-through invites the reader to wonder whether the result was meant to be asserted on.

egonSchiele (commented):
Reviewed against the design doc and the plan. This came out well — the shared `appendRepair` procedure, `markRepaired`/`isNewerThan` owning the generation invariant, and `repairReopenedThread` keeping `Runner.thread()` to one line all land where the spec asked. Two tests in particular earn their keep: the resume-re-entry test pins the frame-locals guard that the whole safety argument rests on, and `unowned-guard-rejected.agency` is the first real coverage `turnBudgetHandler` has ever had.

I verified the things worth verifying: `findLastIndex` is already used in this codebase (`prompt.ts`), `statelogClient.error` exists with exactly the shape being passed, `isInteractive()` is a clean read of `_isInteractive` (no VERBOSE/AGENT_DEBUG leak that could make `ownLabelReject` hang), and `lib/eval` mentions `threadResumed` only in comments, so leaving those files alone was right.

Nothing here blocks merge. Five comments inline: two worth acting on, one question about a case I could not rule out, two nits.

One cross-cutting note on the two statelog sinks. `threadRepaired` gets a direct test that asserts the payload (`repairs and emits threadRepaired with the slugged id and the call ids`), which is exactly right — an optional-chained call is otherwise a silent no-op if it is ever renamed. The `error` call in `restoreThreadForResume` is the same pattern with three levels of optional chaining and no test at all, and it is the more important of the two, since its stated job is to stay findable after the throw gets laundered. Worth making them consistent.

packages/agency-lang/lib/runtime/threadRepair.ts
egonSchiele: Added in b89cb7e10: the refusal test now runs inside runInTestContext with a mock ctx whose statelogClient.error records, and asserts exactly one runtimeError with functionName restoreThreadForResume fires alongside the throw.

packages/agency-lang/lib/runtime/threadRepair.ts
egonSchiele: Reworded in b89cb7e10 to state the observed fact with abandonment as a hint, not a claim: ...repaired after this checkpoint was taken, so restoring would overwrite the newer conversation. This usually means the parked turn was abandoned and the session continued without it. Also added a comment above it naming the concurrent-branch case as the other producer of this refusal.

packages/agency-lang/lib/runtime/runner.ts
egonSchiele: Left the reasoning as a comment at the seam in b89cb7e10: a dangling tail only exists on a thread whose turn parked and bailed out of the whole run, so no code is left executing to take a subthread off it, and a later run must reopen the parent through this branch (repairing it) before it can be active again. The slug nit resolved itself: Copilot pointed out threadCreated/threadResumed emit raw ids, so repairReopenedThread now emits the raw tid and builds no slug at all.

packages/agency-lang/tests/agency/guards/unowned-guard-rejected.agency
egonSchiele: Done in b89cb7e10: all four nodes now call setInteractive explicitly, with a file-level comment on why (module-global flag, order sensitivity). Bonus: foreignGuardBare is now deliberately non-interactive, so the two foreign-guard nodes cover both the notice-then-reject and the reject-without-notice branches instead of both taking the same one.

packages/agency-lang/tests/agency/guards/unowned-guard-rejected.agency
egonSchiele: Dropped the binding in b89cb7e10 — bare interrupt("confirm") statement.

packages/agency-lang/lib/runtime/threadRepair.ts
```diff
@@ -34,45 +95,134 @@ export function needsThreadRepair(cause: AbortCause | undefined): boolean {
...
+  statelog: ThreadRepairedSink | undefined,
+  tid: string,
+): void {
+  if (!thread) return;
+  const repaired = repairAbandonedTurn(thread);
+  if (repaired.length === 0) return;
+  void statelog?.threadRepaired?.({
+    threadId: `t${tid}`,
+    toolCallIds: repaired.map((c) => c.id),
+  });
```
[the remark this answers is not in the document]
egonSchiele: Verified against threadStore.ts (threadCreated at :196 and threadResumed at :312 both emit raw ids) and fixed in b89cb7e10: repairReopenedThread emits the raw tid, with a comment pinning the convention.

packages/agency-lang/lib/runtime/threadRepair.test.ts
```diff
@@ -102,3 +113,174 @@ describe("needsThreadRepair — repair policy", () => {
...
+        { id: "a", name: "whatIAmDoing" },
+        { id: "b", name: "codeAgent" },
+      ]),
+      tool("a"),
+    ]);
+    const events: Array<{ threadId: string; toolCallIds: string[] }> = [];
+    repairReopenedThread(t, { threadRepaired: (e) => { events.push(e); } }, "7");
+    expect(events).toEqual([{ threadId: "t7", toolCallIds: ["b"] }]);
+    expect(t.repairs).toBe(1);
+  });
```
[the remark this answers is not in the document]
egonSchiele: Updated in b89cb7e10: expectation is threadId "7" and the test name now says raw thread id.

PR #659, opened 2026-07-23
expressionSlots: one completeness-checked enumeration of expression positions

egonSchiele:
The hoisting pass (#655) needs to know, for every kind of statement, where expressions live inside it. That knowledge was a hand-written switch, and hand-written position lists drift: three holes were found in one week (thread blocks in value position — caught by CI; goto arguments — caught in review; indexed-assignment targets — caught by a direct question). Every hole is a place where the resume desync the pass exists to fix silently survives, and nothing forced a future node kind to register its positions.

This PR builds the expression twin of `bodySlots`, the codebase's proven answer to the same problem for statement bodies.

## What ships

- **`lib/utils/expressionSlots.ts`** — per node kind, each expression position as `{ expr, mode, write }` in evaluation order, operator-keyed for binops. Each slot carries an eval mode (`once` / `perIteration` / `conditional` / `opaque`) because that is what a rewriter keys safety on. Completeness is enforced, not assumed: a test requires every member of `EXPRESSION_NODE_TYPES` to be enumerated or explicitly declared expression-free, and fails by name otherwise (this immediately surfaced the shared blind spot: `isExpression` / `typeTestExpression` were missing from every existing enumeration).
- **`expressionChildren` derives from the table** with explicit shims preserving its historical orders (assignment and comprehension children come in flow order, which `attachExpressionsToFlow` consumes; slots come in evaluation order, which hoisting consumes — both are load-bearing, and the divergence is now a documented shim instead of an accident waiting to happen). `accessChainExpressions` is deleted; chain handling has one home.
- **`hoistCalls` consumes the slots.** Its rulings tables are retired; the pass keeps only policy (tail rule, chain unit-hoisting, the while restructure — now keyed on the statement kind, with a loud throw for any other `perIteration` carrier). An expression kind missing from the table makes the pass THROW by name instead of falling into the silent generic walk that hid the original holes.

## Proofs, each attributed to what it covers

- The 29 pre-existing hoistCalls unit tests ran UNEDITED against the rewrite: all pass (they are the contract for the motivating shapes).
- `make fixtures` after the rewrite: **zero-line diff** across the entire generator fixture corpus — the rewrite is observationally identical for every shape the corpus contains.
- Corpus invariants over all stdlib + generator fixtures (every file must parse or the test fails): write-fold round trip (every slot's `write` folded with its own `expr` is identity — catches mis-wired writers and enforces slot non-overlap), mode liveness, and order-sensitive parity with the old `expressionChildren` (exceptions individually recorded).
- Type-checker suites byte-identical to baseline. Execution spot-checks (resume regression, eval order, while rewrites, loop-body pause, double-interrupt, thread messages) all pass.
- `gotoWithArgs.agency` now carries a nested call, so the fixture corpus permanently covers the motivating shape it previously missed.

## Found and recorded, not silently changed

Identity-first surfaced two pre-existing holes in the pass, both preserved exactly and documented in `docs/dev/hoist-calls.md` as deliberate follow-ups: bare method-call statements (`result.push(transform(item))` — `transform` was never hoisted) and bare expression statements (a naked binOp statement). The slot-driven dispatcher initially fixed both by accident; the fixture corpus caught the divergence, which is the identity proof doing its job.

## Honest count

Expression-position enumerations after this change: three, down from four — the table, plus `walkNodes` and `getAllVariablesInBody` (both thread ancestor/scope state a plain slot list does not model; both are named in the table's header, and migrating `walkNodes` is the recorded follow-up).


packages/agency-lang/lib/preprocessors/hoistCalls.ts
```diff
@@ -148,97 +126,56 @@ function seedCounter(body: AgencyNode[]): number {
...
+ *  its outermost call (or call-bearing chain) is the statement tail and
+ *  stays in place — hoisting the tail is pointless (it is the pending
+ *  call itself) and node calls even throw in value position. */
+const TAIL_VALUE_KINDS = ["assignment", "returnStatement", "matchYield"];
+
+/** The statement kinds extraction applies to — the pre-slots dispatch
+ *  set, kept exactly for behavior identity. functionCall /
+ *  interruptStatement / whileLoop / withModifier / staticStatement /
+ *  valueAccess have their own cases above the default. */
+const EXTRACTED_STATEMENT_KINDS = [
```
egonSchiele: **This list is a hand-written statement-position enumeration with silent-skip semantics — the shape this PR exists to retire, and it covers exactly the class of node the three original holes belonged to.**

The expression side is now airtight: `isRegisteredExpressionKind` throws by name, and the completeness test forces every `EXPRESSION_NODE_TYPES` member into `HANDLED_KINDS` or `NO_EXPRESSION_SLOTS`. But `EXPRESSION_NODE_TYPES` is the runtime mirror of the `Expression` union (`types.ts:113-146`), so it contains no statement kinds. `assignment`, `gotoStatement`, `messageThread`, `ifElse`, `forLoop`, `returnStatement`, `matchYield` are all absent from it.

And all three drift holes that motivated this work were statement kinds: goto arguments, thread named args, indexed-assignment targets. So the guarantee added here does not cover the class of bug that caused it. A new statement kind carrying expression positions registers in `expressionSlots` (fine), is not in this list (silent), and falls to `default → recurseSlots` with no extraction and no error — the same failure mode as before, one file over.

The asymmetry is worth naming too: an unregistered *expression* kind throws; an unregistered *statement* kind is silently skipped.

A proportionate fix using machinery already in the test file: walk the stdlib corpus, collect the `type` of every node that appears in a statement position, and assert each is either in `EXTRACTED_STATEMENT_KINDS`, handled by its own case above, or in a documented not-extracted list (`valueAccess` and bare `binOpExpression` statements, both already justified in comments here). That fails by name the first time a new statement kind reaches stdlib. A `STATEMENT_NODE_TYPES` mirror next to `EXPRESSION_NODE_TYPES` would be stronger and is the same pattern the codebase already uses, but it is a bigger change than this PR needs.

packages/agency-lang/lib/preprocessors/hoistCalls.ts
```diff
@@ -314,177 +279,98 @@ function recurseSlots(stmt: any, counter: Counter): AgencyNode {
...
-      });
-      if (changed) expr = { ...expr, [key]: mapped };
-    } else if (child && typeof child === "object") {
-      const inner = walk(child, counter, true);
-      temps.push(...inner.temps);
-      if (inner.expr !== child) expr = { ...expr, [key]: inner.expr };
-    }
+  let expr: any = node;
+  for (const slot of expressionSlots(node)) {
+    if (slot.mode !== "once") continue;
```
egonSchiele: Minor consistency: `rewriteStatement` throws on a `perIteration` slot it has no restructure strategy for ("fail by name rather than mis-apply the while rewrite" — good), but here in `walk` the same mode is silently skipped by the `!== "once"` filter, lumped in with conditional and opaque.

Both are unreachable today for the same reason (comprehensions never survive parse-time desugar), so this is not a bug. But the two sites disagree about what an unexpected `perIteration` means, and the throwing one is the honest one. Either mirror the throw, or add a word to the filter comment saying an expression-position `perIteration` is deliberately treated as non-hoistable rather than as an error.

packages/agency-lang/lib/utils/expressionSlots.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    expect(cond.type).toBe("typeTestExpression");
+    const slots = expressionSlots(cond);
+    expect(slots).toHaveLength(1);
+    expect(slots[0].mode).toBe("once");
+    expect((slots[0].expr as any).value).toBe("pv");
+  });
+});
+
+describe("expressionSlots: completeness against EXPRESSION_NODE_TYPES", () => {
+  it("every expression kind is enumerated or explicitly empty, never both", () => {
```
egonSchiele, line 143: This test is the good part of the PR — it is what turns the table from "consistent with the old function" into "authoritative." Worth noting in a comment what it does *not* reach, so the next reader does not over-trust it: `EXPRESSION_NODE_TYPES` covers the `Expression` union only, so statement kinds that carry expression positions (`assignment`, `gotoStatement`, `messageThread`, `ifElse`, `forLoop`, …) are outside its guarantee — and those are the kinds the three original holes were in.

See the note on `EXTRACTED_STATEMENT_KINDS` for a corpus-based way to close that half using the fixture walk already set up in this file.

egonSchiele (commented):
Reviewed against the plan and verified behavior by running the pass directly on this branch rather than reading only.

**This is a clean piece of work and the design is right.** Positions and timing became data, the traversal became a loop, and hoist policy stayed in the pass — the split is held consistently, and the places where the table and the pass disagree are documented rather than smoothed over. Specific things worth calling out:

- The `EvalMode` note that no consumer branches on conditional-vs-opaque, with an explicit "do not build logic on it without tests that distinguish them." That is the right way to keep a distinction that is documentation rather than behavior.
- `perIteration` as a prohibition with the restructure keyed on the owner's kind, plus a throw for carriers with no strategy.
- The assignment order shim in `expressionChildren`, which keeps flow order `[value, ...targetChain]` while slots stay in evaluation order, with both callers named. That divergence was easy to introduce silently and it did not get introduced silently.
- The expression/statement seam written down in both headers, and `walkNodes` / `getAllVariablesInBody` scoped out by name as a recorded follow-up instead of quietly left behind.
- Completeness enforced by test rather than assumed, plus the write-fold round trip and mode-liveness invariants over the corpus.

I checked the one thing most likely to break silently: whether slot iteration changed hoisting behavior for access chains. It does not. `outer(mk(inner()).field)` yields `__hoist_0 = inner()`, `__hoist_1 = mk(__hoist_0)`, `x = outer(__hoist_1.field)` on this branch, matching `main` — where the base already walks as an expression. Method-call slots are correctly walked with `hoistSelf` off, so a chain method never becomes a standalone temp.

One substantive gap, inline: **the completeness guarantee covers expression kinds only, and every one of the three drift holes that motivated this PR was a statement kind.** `EXTRACTED_STATEMENT_KINDS` is a new hand-written list whose miss mode is a silent skip, while the expression side throws by name. That is not a regression — it is the pre-existing behavior preserved deliberately, and the comment says so — but it means the anti-drift claim in the header is currently half true. There is a proportionate fix that reuses the corpus walk already in the test file.

Two smaller notes inline (a `perIteration` handling inconsistency between the two dispatchers, and a suggested caveat on the completeness test).

packages/agency-lang/lib/preprocessors/hoistCalls.ts
egonSchiele: Implemented the corpus-based version in 53a51fb25. The test walks every statement position across stdlib plus the generator fixtures (function/graphNode bodies and every bodySlots body) and asserts each observed kind appears in one of four exported lists in hoistCalls.ts: EXTRACTED_STATEMENT_KINDS, STATEMENT_CASE_KINDS (the own-case dispatch set), NON_EXTRACTED_STATEMENT_KINDS (each entry with a justification comment: bare binOp statements, keyword, the statement containers that only recurse, multiLineComment), or SKIP_TYPES. A kind absent from all four fails by name with the file it was first seen in. Corpus run found 21 statement kinds; all have rulings.

Left STATEMENT_NODE_TYPES as a follow-up per your sizing — the corpus check gives the fail-by-name property now, and the type-level mirror can subsume it later.

packages/agency-lang/lib/preprocessors/hoistCalls.ts
egonSchiele: Mirrored the throw in walk() (53a51fb25) — same message shape, same "dead by construction today" comment. Consistency was free since both sites are unreachable for the same reason.

packages/agency-lang/lib/utils/expressionSlots.test.ts
egonSchiele: Added the scope caveat as a comment on the describe block, pointing at the new statement-position corpus test at the bottom of the file, and updated the hoistCalls header and docs/dev/hoist-calls.md to describe both halves (type-checked expression side, corpus-checked statement side). 53a51fb25.

packages/agency-lang/lib/utils/expressionSlots.ts
```diff
@@ -0,0 +1,405 @@
...
+  "typeTestExpression",
+];
+
+/** True when `type` is known to this enumeration — either it has slots
+ *  or it is declared expression-free. hoistCalls throws on anything
+ *  else: an unregistered kind must fail by name, not fall into a silent
+ *  generic walk (that walk is how the drift holes stayed hidden). */
+export function isRegisteredExpressionKind(type: string): boolean {
+  return HANDLED_KINDS.includes(type) || type in NO_EXPRESSION_SLOTS;
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 53a51fb25 — both this site and the completeness test now use Object.hasOwn.

packages/agency-lang/lib/utils/expressionSlots.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    for (const kind of EXPRESSION_NODE_TYPES) {
+      const enumerated = HANDLED_KINDS.includes(kind);
+      const declaredEmpty = kind in NO_EXPRESSION_SLOTS;
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 53a51fb25 — switched to Object.hasOwn here and at the isRegisteredExpressionKind site.

PR #665, opened 2026-07-23
Code templates: typed holes, fill, and hygienic code generation

egonSchiele:
Agency programs can now generate other Agency programs by filling typed holes in template files. A human writes the skeleton, a model fills the gaps, and every gap is checked before anything runs. This closes the structural half of the safety story: handlers already governed what generated code may *do*; templates govern what it can *be*.

## What a user sees

A template is an ordinary `.agency` file with holes:

```
node main(): string {
  const topic: string = #topic
  return llm("Summarize ${topic}")
}
```

```
const tpl = loadTemplate(__dirname, "summarize.agency")
const filled = fill(tpl.value, { topic: modelChosenTopic })
return runCode(toSource(filled.value))
```

New in `std::agency`: `Code` (the existing `AST` plus a fragment kind), `loadTemplate`, `fill`, `holesOf` (structured entries with name, sort, splice, and type — enough for a model to know what each hole accepts), `toSource`, `parseExpr`, `parseStatements`. Guide page: `docs/site/guide/templates.md`.

**Filling composes.** A partially filled template is an ordinary `Code` value; grafting it into another template carries its remaining holes along, and a later fill completes them. Build the shape first, parameterize last — this is the workflow the feature exists for, and it is pinned by both a unit test and an end-to-end fixture.

## The safety rules

- **Fillers are never parsed.** Plain values lift to literal AST nodes; `fill(t, { v: "readFile(\"/etc/passwd\")" })` produces a string literal containing those characters, not a call. An 11-case escaping battery proves fillers stay inert through print → re-parse (including `${` interpolation openers).
- **Identifier holes are the one exception** — the filler becomes a name — so they validate against the identifier grammar, the reserved-word list, and the `__hyg` prefix.
- **Scope-aware hygiene.** Colliding binders (and only those) are renamed with the reserved `__hyg` prefix: a filler's free name cannot capture a template binder, a filler cannot redeclare one, and two fillers binding the same name get distinct fresh names. A `tmp` in an unrelated function is untouched.
- **Refusals.** A program with unfilled holes cannot compile (`AG8001`, naming every hole, enforced by a pre-pass in the builder). An expression hole with no discoverable type is `AG8002`. Both have `agency explain` entries and `expectedCompileError` fixtures (thanks to #662).
- **Governance survives templating.** An end-to-end fixture builds a program from templates, runs it via `runCode`, and the parent handler rejects the child program's file read.

## Breaking change

`#` property descriptions are removed to free the sigil. The replacement is `@jsonSchema({ description: ... })`, which emits identical schemas (regression-tested). Migrated in this PR: two test fixtures, `lib/agents/eval/judge.agency`, `lib/agents/eval/goalJudge.agency`, `lib/cli/doc.test.ts`, and `docs/site/guide/llm.md`. A parser test pins that the old syntax is now a parse error.

## Implementation notes / deviations from the plan

- Declaration names and import specifiers hold **plain strings** in the AST, so identifier holes work by widening those three fields to `string | Hole`; a total `declaredName()` accessor (prints `#name` for holes, safe as a registry key since `#` cannot appear in identifiers) keeps ~90 existing call sites total. Identifier fills produce validated plain strings, not nodes.
- Hole queries (`findHoles`, `holeNames`, `holeInfos`) live in `lib/utils/holes.ts` so the builder, checker, and template runtime can all reach them without cross-layer imports. `Code` lives in `lib/runtime/template/code.ts` to break a stdlib/runtime cycle.
- Hygiene renames are keyed by scope (`fn:name` / `node:name` / `global`) and applied by a rewriting walk that threads active renames through the recursion, stopping at inner scopes that rebind the name. Binder fields are neither expression slots nor body slots, which is why this walk is bespoke (same recorded exception as `walkNodes` own descent).
- Fill-time type checking is deliberately primitive-only validation: both sides must be certainly known (plain JS values, literal fragments). Checking fragments against the completed program's module scope needs a checker entry point that does not exist yet — noted in `fill.ts` as the seam to narrow.
- Serialization needed **no code**: `Code` is plain data and survives interrupt checkpoints as-is. The verify-first fixture that proved it is kept as a regression guard.

## Bugs the tests caught during implementation

- `hole` had to be registered in `expressionSlots.ts` — the hoistCalls completeness tripwire (#659) rejected the unregistered kind exactly as designed.
- `walkNodes` never descended into a guard block's head arguments, so `guard(time: #minutes)` silently lost its hole — the compose fixture caught it; fixed with a `guardBlock` case in the walker.
- The statement-hole boundary check had to accept the blank-line sentinel (``), not just `\n`.

## Known v1 limits

Destructuring-pattern binders are not tracked by hygiene (noted in the guide); AG8002 recognizes the untyped-assignment position only; fragment type checking against module scope is the recorded follow-up; LSP behavior is covered by pipeline-tolerance tests but has not had a manual editor pass.

## Testing

`pnpm test:run`: 9,249 passing (591 files). `tests/agency/templates/`: 6/6 fixtures passing (AG8001, AG8002-strict, checkpoint survival, injection, handler governance, compose-then-parameterize). `tsc --noEmit`, `lint:structure`, `make`, `make doc` all clean. The full agency suite is left to CI per repo convention.


packages/agency-lang/lib/runtime/template/fill.ts
```diff
@@ -0,0 +1,264 @@
...
+  }
+  return out;
+}
+
+function fillOne(
+  hole: Hole,
+  value: unknown,
+  expected: Record<string, string>,
+): string | AgencyNode | AgencyNode[] {
+  if (hole.sort === "identifier") return identifierFillFor(hole, value);
```
egonSchiele, line 136: **Splice is silently dropped for identifier holes.** This returns before the `hole.splice` branch below, so a hole parsed with `splice: true` in a name position loses the flag with no error.

Verified against this branch:

```
def #...name(): number {
  return 1
}
```

parses as `sort: "identifier", splice: true`, and `fill(t, { name: "foo" })` yields:

```
def foo(): number {
  return 1
}
```

Silently-wrong output rather than an error. `exprHoleParser` rejects splices explicitly (`parsers.ts:945-955`); `identifierHoleParser` doesn't. Cheapest fix is to reject there so it fails at parse time, since a splice in a name position has no meaning at all. Failing that, throw here.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -859,6 +880,149 @@ export const variableNameParser: Parser<VariableNameLiteral> = label("an identif
...
+      // (replaceBlankLines runs before parsing).
+      after.startsWith(BLANK_LINE_SENTINEL);
+    if (!atStatementEnd) {
+      return failure("a statement hole must occupy the whole statement", input);
+    }
+    return success({ ...result.result, sort: "statements" as const }, result.rest);
+  },
+);
+
+export const identifierHoleParser: Parser<Hole> = memo(
```
egonSchiele, line 996: This accepts splices, unlike `exprHoleParser` just above, which rejects them with a clear message. `#...name` in a declaration name or import specifier has no meaning — rejecting it here gives a parse-time error at the right place and makes the `fillOne` identifier branch total.

Worth a test alongside the existing `rejects a splice in expression position` case in `hole.test.ts`.

packages/agency-lang/lib/runtime/template/hygiene.ts
```diff
@@ -0,0 +1,334 @@
...
+  const used: string[] = [];
+  for (const { node } of walkNodesArray(code.nodes)) {
+    if (node.type === "variableName") used.push(node.value);
+  }
+  return used
+    .filter((name) => !bound.includes(name))
+    .filter((name, index, all) => all.indexOf(name) === index);
+}
+
+export function assertNoReservedPrefix(code: Code, what: string): void {
```
egonSchiele: **This misses declaration names, which breaks the impossibility claim.** `bindersOfNode` covers assignments, parameters, and for-loop binders, but not function names, node names, or type-alias names.

Verified against this branch — both of these pass the check:

```
def __hyg1_x(): number { return 1 }
type __hyg1_T = string
```

That matters because `fresh()` produces `__hyg1_<name>` for the very first rename, so a filler declaring `def __hyg1_tmp()` can collide with the rename generated for a binder named `tmp`. The module doc comment says rejecting reserved-prefix input is "what makes collisions with renamed names impossible by construction" — with declaration names uncovered, that holds for variables but not for declarations.

Adding the three name fields to the walk here (or to a dedicated `allDeclaredNames`) closes it.

packages/agency-lang/lib/runtime/template/hygiene.ts
```diff
@@ -0,0 +1,334 @@
...
+  for (const { node } of walkNodesArray(code.nodes)) {
+    names.push(...bindersOfNode(node));
+  }
+  return names.filter((name, index) => names.indexOf(name) === index);
+}
+
+/** Names a fragment USES but does not bind — the side of capture an
+ *  earlier plan draft got wrong: comparing binders to binders finds
+ *  nothing, because `tmp` used as an expression binds nothing. */
+export function freeNamesOf(code: Code): string[] {
```
egonSchiele, line 89: **Systemic risk worth recording in the code.** This function's correctness — and therefore capture avoidance — is exactly as complete as `walkNodes`' descent. If the walker misses an expression position, `freeNamesOf` under-reports, no test fails, and a filler silently captures a template binder. That is the precise bug the feature exists to prevent, and it fails open.

The `guardBlock` gap you fixed in this PR is the proof: it was caught by the compose fixture happening to use `guard(time: #minutes)`, not by the `expressionSlots` tripwire, because the bespoke descent isn't covered by it.

Not asking for new machinery here, but a comment naming the dependency would tell the next person that adding a node kind with expression children means auditing this path. If there's a cheap way to extend the #659 completeness tripwire to the bespoke descent, that would be better still.

packages/agency-lang/lib/runtime/template/fill.ts
```diff
@@ -0,0 +1,264 @@
...
+    );
+  }
+  return value;
+}
+
+function fillerLoc(hole: Hole): SourceLocation {
+  return { ...hole.loc, origin: { kind: "filler", name: hole.name } };
+}
+
+function stampOrigin(node: AgencyNode, hole: Hole): AgencyNode {
```
egonSchiele: **Attribution only reaches the top level of a grafted fragment.** This maps over `value.nodes`, so nested nodes keep the `loc` they got from parsing the fragment string — line and column into a source string that no longer exists — and carry no `origin`.

So for a multi-statement graft, an error on the first statement says "in the fill for `#body`", and an error inside a nested expression reports a stale position with no indication of who wrote it. Since carrying both origins was one of the design's explicit goals (and the most-cited Template Haskell complaint), it's worth either recursing the stamp through children or noting the limit in the doc comment above.

packages/agency-lang/lib/utils/holes.ts
```diff
@@ -0,0 +1,76 @@
...
+/** Types the hole's POSITION supplies, keyed by hole name — today the
+ *  annotated-assignment position (`const x: string = #text`). This is what
+ *  lets holesOf report a type the template author never wrote on the hole
+ *  itself. First occurrence wins, matching holeInfos. */
+export function positionInferredTypes(nodes: AgencyNode[]): Record<string, string> {
+  const inferred: Record<string, string> = {};
+  for (const visit of walkNodesArray(nodes)) {
+    if (visit.node.type !== "hole") continue;
+    const hole = visit.node as Hole;
+    if (hole.typeAnnotation || inferred[hole.name]) continue;
```
egonSchiele, line 60: First-occurrence-wins per hole name. With the same name in two positions of different types:

```
const a: string = #v
const b: number = #v
```

fill-time validation checks against `string` only, and the `number` mismatch falls through to the run-time check. Reasonable v1 behavior given the fill-time checking is deliberately primitive-only — worth a sentence in the doc comment so the next reader doesn't assume both positions are validated.

egonSchiele (commented):
Reviewed the full diff, read the security-critical files closely, and ran the template suite locally: **112/112 unit tests pass** (`lib/runtime/template/`, `lib/parsers/hole.test.ts`, `lib/typeChecker/holes.test.ts`, `lib/backends/holeRefusal.test.ts`, `lib/stdlib/template.test.ts`).

This is a strong implementation. The parts that carry the most risk are the parts that are best defended: the lifting rule is stated in three places and pinned by a negative assertion, the 11-case escaping battery tests the property that actually matters (inert through *print → re-parse*, not merely "becomes a literal in the AST"), and `certainTypeOf` correctly returns `null` for an interpolated string literal rather than guessing. The hygiene redesign is right — comparing template binders against filler *free names*, per-graft rename maps, and scope-keyed renames that stop at a shadowing inner binder. Deviations from the plan are documented with their reasoning, and two of them (plain-string declaration names, `Code` in `lib/runtime/` to break a cycle) are better than what the plan specified.

Four findings below. Two are verified silent-wrong-behavior bugs, one is a systemic risk worth a comment in the code, one is a stated design goal that is only half-delivered. None block the concept; the first two I would fix before merge.

### Verified by running against this branch

**1. A splice in identifier position is silently dropped.** `def #...name()` parses (`sort: "identifier"`, `splice: true`), and filling produces `def foo()` with no error — the splice flag vanishes. `identifierHoleParser` doesn't reject splices the way `exprHoleParser` does, and `fillOne` returns on the identifier branch before reaching the splice branch.

**2. The reserved-prefix check misses declaration names.** `bindersOfNode` covers assignments, parameters, and for-loop binders, but not function, node, or type-alias names. A filler containing `def __hyg1_x()` or `type __hyg1_T = string` passes `assertNoReservedPrefix`. Since `fresh()` starts at `__hyg1_`, a first rename can collide with exactly that name — which is the "impossible by construction" guarantee the module's own doc comment claims.

### Worth addressing

**3. Hygiene's correctness rides on `walkNodes` descent completeness, and nothing enforces it.** `freeNamesOf` and `assertNoReservedPrefix` are only as complete as the walker. A missing descent case doesn't fail loudly — it under-reports free names, which means a missed capture, which is the exact bug this feature exists to prevent. The `guardBlock` gap found during implementation is the proof: it was caught by a fixture that happened to use `guard(time: #minutes)`, not by a tripwire. Worth at minimum a comment at `freeNamesOf` naming the dependency.

**4. `stampOrigin` is shallow, so filler attribution is partial.** Only the top-level nodes of a grafted fragment get `origin`. Nested nodes keep the `loc` they got from parsing the fragment string — coordinates into source that no longer exists, with no origin marker. A type error inside a grafted fragment therefore reports a meaningless position and can't say who wrote it.

### Smaller

- `positionInferredTypes` is first-occurrence-wins per hole name. Two positions with different expected types (`const a: string = #v` / `const b: number = #v`) validate against the first only; the second is caught at run time. Fine as a v1 limit, worth a sentence in the doc comment.
- The guide's escape-hatch note says "a template author who writes that has chosen to." The sharper case is a model holding `parseExpr` and `fill` in the *same tool list* — then the model chooses, not the author. Worth one line telling people not to hand out both if they're relying on structural constraint.

### Confirmed accurate

Spot-checked the claims in the description: AG8001 does collect every hole before throwing (`typescriptBuilder.ts:499-509`); serialization genuinely needed no code; `@jsonSchema({ description })` emits an identical schema to the removed `#` form; `declaredName` is safe as a registry key because `#` can't appear in a user identifier. The compose-then-parameterize workflow works the way the guide describes — a second `fill` sees the grafted fragment's holes, because `holeNames` runs over the composed template.

egonSchiele:
All findings addressed in c7717c083. Point-by-point, with one thing your finding 3 flushed out that is worth reading:

**1. Splice in identifier position** — fixed at parse time, as suggested. `identifierHoleParser` now rejects splices with "a splice cannot appear in a name position". Your repro also exposed a second layer: after the rejection, the def/node name capture's raw-name fallback (`many1Till("(")`) happily consumed `#...name` as a literal string name, so the parse still succeeded with garbage. The name position now has no raw fallback for `#`-initial input — a rejected hole form fails the parse outright. Tests for def-name and import-specifier splices sit next to the expression-position one.

**2. Reserved prefix missing declaration names** — fixed. `assertNoReservedPrefix` walks an `allNamesOfNode` that includes function names, node names, and type-alias names (via `declaredName` for the widened fields). Both of your repros now throw, plus a template-side case (`node __hyg1_main`). Three new tests.

**3. Walker-completeness dependency** — comment added at `freeNamesOf` naming the failure mode, and you were right that it deserved a tripwire, not just prose: there is now a 14-position hole battery in `hole.test.ts` (assignment value, binop operand, conditions, call/named args, guard head, return, array/object elements, interpolation, for-iterable, match scrutinee, try). **It caught a second real gap on its first run**: `walkNodes` did not descend into `tryExpression`, so free names under `try f(tmp)` were invisible to hygiene — exactly the fail-open capture you described. Fixed in `node.ts` with a pointer back to the battery. The battery is the cheap version of extending the #659 tripwire; a structural completeness check over the bespoke descent would still be better and I would keep it on the follow-up list.

**4. Shallow `stampOrigin`** — now recursive. Every node of a grafted fragment carries `origin: { kind: "filler", name }` (nodes only — loc-less sub-records like text segments keep their exact shape, which the formatter depends on), and a loc-less top node falls back to the hole's own position. Test asserts the stamp on a nested expression two levels down.

**Smaller:**
- `positionInferredTypes` doc comment now states first-occurrence-wins and that a second position of a different type falls through to the run-time check.
- Guide has the sharper framing: if a model holds `parseExpr` and `fill` in the same tool list, the model chooses — wrap composition in one author-owned tool instead.

**Copilot's three:**
- `__proto__` lift: real — in a JS object literal that key sets the prototype even quoted. `liftValue` now rejects objects with a `__proto__` key loudly rather than emitting a literal whose shape silently differs from the data. Test uses `JSON.parse` to build the poisoned input the way a model actually would.
- `holeNames` dedup is linear now.
- `HoleInfo.type` rendering: not a bug — `type?: string` and `type: string | null` are the same type in Agency by construction (the optional-property parser desugars `key?: T` to `T | null`), so the generated docs show the formatter's canonical form of what the source declares. Left as is.

State after fixes: 9,270 unit tests green (21 new), all 6 template fixtures green, tsc/make/lint:structure clean.

packages/agency-lang/docs/site/.vitepress/config.mts
```diff
@@ -83,6 +83,7 @@ export default defineConfig({
               text: "Compiling and Running Code",
               link: "/guide/compiling-and-running",
             },
+            { text: "Code Templates", link: "/guide/templates" },
```
egonSchiele: lets name this feature `Template Agency`

packages/agency-lang/lib/runtime/template/code.ts
```diff
@@ -0,0 +1,31 @@
...
+ * fragment kind. The kind exists because a bare expression is not a
+ * parseable program — `AST` alone cannot represent "one expression", and
+ * an `expr` hole must be fillable with `Code`. A missing `kind` (a value
+ * built by `parseAST` rather than the template module) means `"program"`.
+ */
+export type Code = {
+  type: "agencyProgram";
+  kind?: "program" | "statements" | "expr";
+  nodes: AgencyNode[];
+  docComment?: unknown;
```
egonSchiele: why is the type `unknown`?

packages/agency-lang/lib/runtime/template/hygiene.ts
```diff
@@ -0,0 +1,354 @@
...
+export type RenamePlan = {
+  template: ScopedRename[];
+  /** Per-graft flat maps, keyed by hole name (and element index for
+   *  splices, as `name[i]`). A filler is one fragment grafted into one
+   *  place, so its whole tree is the scope. */
+  fillers: Record<string, Record<string, string>>;
+};
+
+function scopeKeyOfChain(scopes: Scope[]): string {
+  for (let i = scopes.length - 1; i >= 0; i -= 1) {
```
egonSchiele: prefer `for (let item of array)` style loops

packages/agency-lang/lib/runtime/template/hygiene.ts
```diff
@@ -0,0 +1,354 @@
...
+ *  LOAD-BEARING DEPENDENCY: this is exactly as complete as `walkNodes`'
+ *  descent (lib/utils/node.ts). A node kind whose expression children the
+ *  walker misses under-reports free names here, no test fails, and a
+ *  filler silently captures a template binder — the precise bug hygiene
+ *  exists to prevent, failing open. (The guardBlock head-argument gap
+ *  fixed in this feature's PR is the historical example.) If you add a
+ *  node kind with expression children, audit the walker's descent AND add
+ *  that position to the hole-position battery in
+ *  lib/parsers/hole.test.ts, which is the tripwire for this path. */
+export function freeNamesOf(code: Code): string[] {
```
egonSchiele, line 89: **The dependency this comment documents has a live instance: `is` / `typeTest` operands.**

Verified against `c7717c0`:

```
freeNamesOf(parseExpr("tmp is string"))  →  []
```

and end-to-end, the flagship capture scenario routed through `is`:

```
node main() {
  const tmp = getApiKey()
  const leak: boolean = #e
  return leak
}
```

filled with `parseExpr("tmp is string")` yields, with no rename and no error:

```
const tmp = getApiKey()
const leak: boolean = tmp is string
```

The filler captures the API key — the exact bug the `renames the template binder when a filler uses that name` test guards, reached through a different position.

**Why it slips through:** the two tables disagree deliberately. `expressionSlots.ts:401` *does* enumerate `n.expression` for `isExpression`/`typeTestExpression`. `node.ts:77-81` returns `[]` for those kinds in the walker's child view, commented "Preserved empty so type-checker behavior is untouched." This function rides on the second table, so the walker is correct as documented and hygiene is reading the one that was intentionally left incomplete.

So the fix probably isn't in `node.ts` — changing that view would touch the checker. Either have `freeNamesOf` consult `expressionSlots` (the complete enumeration, per its own doc) for these kinds, or special-case them here.

Worth noting this is ordinary usage rather than a corner: type patterns shipped in #631, so an `is` expression filling a `boolean` hole is a thing people will write.

packages/agency-lang/lib/parsers/hole.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    expect(parses(`node main() {\n  f(#...args)\n}\n`)).toBe(true);
+  });
+});
+
+// The tripwire for hygiene's walker dependency (see freeNamesOf in
+// lib/runtime/template/hygiene.ts): every expression position a hole can
+// occupy must be REACHED by walkNodes' descent, or free-name analysis
+// under-reports and capture avoidance fails open. Each entry pins one
+// position; add one when a node kind gains an expression child.
+describe("hole positions are reachable by the walker", () => {
```
egonSchiele, line 155: This battery is the right response to the walker-dependency finding — a mechanism rather than a comment — and it earned its keep immediately by catching `tryExpression`.

Two positions to add, because they're currently unreached and cause a real capture (details on `hygiene.ts:87`):

```typescript
["is-expression operand", `node main() {\n  const x: boolean = #h is string\n}\n`],
["type-test operand", `node main() {\n  const x: boolean = #h is? string\n}\n`],
```

(Check the second against the real `typeTestExpression` surface syntax — I didn't confirm the spelling.)

I also probed nine other positions against this branch; the rest are all reached (`index access`, `comprehension` element/iterable/filter, `method-call argument`, `await operand`, `named argument`). `def f(x: number = #h)` and `#h.field` don't parse at all, which is fine — a hole that can't get there can't capture.

egonSchiele (commented):
Re-reviewed at `c7717c0`. **All four findings are fixed**, verified by running against the branch rather than reading the diff — and the fixes are better than what I asked for in three places. **134/134** template unit tests pass (up from 112).

| # | Finding | Status |
|---|---|---|
| 1 | Splice silently dropped in name position | Fixed — parse error in all three positions |
| 2 | Reserved prefix missed declaration names | Fixed — function, node, type alias, effect all rejected |
| 3 | Walker-completeness dependency unguarded | Fixed **beyond the ask** — see below |
| 4 | `stampOrigin` shallow | Fixed — 3/4 located nodes in a nested graft now carry `origin`; the 4th is the template's own node, correctly unstamped |

### Three things you found that I did not

- **The `or(...)` fallback.** Rejecting the splice in `identifierHoleParser` alone would have let `or(declNameHoleParser, many1Till(char("(")))` swallow `#...name` as the literal string `"#...name"` instead of failing. The `declNameParser` no-fallback-when-input-starts-with-`#` rule is the right fix and I missed the interaction entirely.
- **`__proto__` prototype pollution in `liftValue`.** A model-supplied record round-tripped through `JSON.parse` carries `__proto__` as an own property, so it survives `Object.keys` and would emit an object literal that sets the prototype instead of a key. That is a real vector in the exact path the feature exists to secure, and it was not in my review.
- **The `tryExpression` walker gap** — found by the tripwire you built in response to finding 3, which is the outcome that matters more than the individual bug.

I asked for a comment naming the dependency; you added a 14-position battery that immediately caught a second real gap. That is the right shape of fix.

### One new finding: the battery has a gap, and it is a live capture bug

`is` / `typeTest` operands are not reached, so a filler can capture a template binder through one. Verified end-to-end against `c7717c0`:

```
freeNamesOf(parseExpr("tmp is string"))  →  []      // should be ["tmp"]
```

and the flagship capture scenario, routed through `is`:

```
node main() {
  const tmp = getApiKey()
  const leak: boolean = #e
  return leak
}
```

filled with `parseExpr("tmp is string")` produces — with no rename, no `__hyg`, no error:

```
const tmp = getApiKey()
const leak: boolean = tmp is string
```

The filler now reads the API key. This is the same bug the `renames the template binder when a filler uses that name` test guards, reached by a different expression position.

**Diagnosis:** the two slot tables disagree on purpose. `expressionSlots.ts:401` enumerates `n.expression` for `isExpression`/`typeTestExpression`, but `node.ts:77-81` returns `[]` for those kinds in the walker's child view, with a comment saying it is `"Preserved empty so type-checker behavior is untouched"`. `freeNamesOf` rides on the second one. So the walker is behaving exactly as documented; hygiene is consulting the table that was deliberately left incomplete.

That points at the fix: don't change `node.ts`'s child view (it is load-bearing for the checker) — have `freeNamesOf` consult the complete enumeration, or special-case these two kinds in the free-name walk. Since type patterns shipped in #631, `#e` filled with an `is` expression is ordinary usage, not a corner.

Adding `["is-expression operand", ...]` and `["type-test operand", ...]` to the battery pins it.

### Correction to my own review

My first-round example used `import std::fs { #tool }`, which is not Agency syntax — the real form is `import { #tool } from "std::fs"`, which you used correctly. I briefly read that as a parse regression from the splice fix; it was my error, and there is no regression. Non-splice name holes still parse in every position.

Everything else from the first round still stands as resolved. Once the `is`/`typeTest` gap is closed I have nothing further.

packages/agency-lang/lib/runtime/template/lift.ts
```diff
@@ -0,0 +1,49 @@
...
+export function liftValue(value: unknown, loc: SourceLocation): AgencyNode {
+  if (value === null || value === undefined) return nullLiteral(loc);
+  if (typeof value === "string") return stringLiteral(value, loc);
+  if (typeof value === "number") return numberLiteral(value, loc);
+  if (typeof value === "boolean") return booleanLiteral(value, loc);
+  if (Array.isArray(value)) {
+    return arrayLiteral(value.map((item) => liftValue(item, loc)), loc);
+  }
+  if (typeof value === "object") {
+    return objectLiteral(
```
egonSchiele, line 40: **CRITICAL: object keys reach the printer unescaped, which defeats the lifting rule.**

Keys are passed through raw here. `addQuotesToKey` (`agencyGenerator.ts:1237`) then wraps a non-identifier key in quotes *without escaping its contents*, so a key containing `"` breaks out of the literal.

Reproduced end-to-end on `c7717c0`:

```js
const evil = JSON.parse('{"a\\": readFile(\\"/etc/passwd\\"), \\"b": 1}');
fill(tpl, { v: evil })
```

prints:

```
const x = {
  "a": readFile("/etc/passwd"), "b": 1
}
```

and re-parsing that output yields a real `functionCall` node named `readFile` — I checked the AST, not just the text. So `fill` → `toSource` → `runCode` runs attacker-chosen calls sourced from data.

**Why nothing caught it:** the escaping battery covers string *values* only. `stringLiteral` escapes its content; keys have no equivalent path.

**Why the printer isn't the culprit:** keys arriving from the parser are already escaped — `{ "a\\"b": 1 }` parses and reprints correctly today. `liftValue` is the first caller to hand it raw runtime strings, so the invariant breaks at this seam.

Fix could go either place, but escaping in `addQuotesToKey` is the robust one since it protects every future caller. Whichever you choose, the escaping battery should grow a key-side mirror of every case it already covers on the value side — same strings, supplied as keys.

packages/agency-lang/lib/runtime/template/fill.ts
```diff
@@ -0,0 +1,295 @@
...
+      throw new Error(
+        `\`${name}\` is not a hole in this template. Its holes are: ${present.join(", ") || "(none)"}.`,
+      );
+    }
+  }
+
+  // Hygiene, one declarative sequence: reject reserved-prefix input,
+  // compute the rename plan, rename the template within the affected
+  // scopes, rename each filler within itself, then substitute.
+  assertNoReservedPrefix(code, "template");
```
egonSchiele: **This breaks the compose workflow whenever hygiene fires.** The counter in `computeRenames` restarts at 0 on every call, and this guard runs over the incoming template — so a second `fill` rejects names the *first* fill generated:

```
const once  = fill(tpl, { e: parseExpr("tmp") })   // renames tmp → __hyg1_tmp
const twice = fill(once, { later: 2 })
// Error: `__hyg1_tmp` uses the reserved prefix `__hyg`, which templates
//        keep for hygiene. Rename it in the template.
```

Verified on `c7717c0`. "Build the shape first, parameterize last" is the workflow the PR description leads with, and it fails as soon as any fill triggers a rename — which is precisely when the template most needs a second pass.

The message is also misleading: it tells the user to rename something in *their* template that the system wrote.

The guard needs to tell renamer-produced names from caller-supplied ones. Either mark the `Code` value as already-hygienized and skip previously-generated names, or seed the counter above the highest `__hyg<n>` index already present so a re-fill can't collide. The second keeps the guard total and needs no new state on `Code`.

Worth a test that fills twice with a rename in the first pass — the existing compose fixture presumably uses non-colliding names, which is why this stayed hidden.

packages/agency-lang/lib/runtime/template/literals.ts
```diff
@@ -0,0 +1,57 @@
...
+export function stringLiteral(value: string, loc: SourceLocation): StringLiteral {
+  return {
+    type: "string",
+    segments: [{ type: "text", value }],
+    delimiter: '"',
+    loc,
+  };
+}
+
+export function numberLiteral(value: number, loc: SourceLocation): NumberLiteral {
```
egonSchiele, line 31: `String(value)` turns non-finite numbers into bare identifier tokens. `Infinity`, `-Infinity`, and `NaN` all print unquoted, and `const x = Infinity` re-parses as a `variableName`, not a number — so the value silently becomes a reference to whatever (if anything) that name binds in the generated program.

Reachable from model output without any exotic input: `JSON.parse("1e400")` is `Infinity`, and I confirmed `fill(t, { v: 1e400 })` emits `const x = Infinity`.

Rejecting non-finite values in `liftValue` — the same shape as the `__proto__` guard you added — seems right, since there is no Agency literal for them to lift to.

packages/agency-lang/lib/runtime/template/hygiene.ts
```diff
@@ -0,0 +1,354 @@
...
+
+    // Case 2.
+    const map: Record<string, string> = {};
+    for (const name of own) {
+      if (visibleNames.includes(name)) map[name] = fresh(name);
+    }
+    fillerRenames[site.fillerKey] = map;
+  }
+
+  // Case 3 — skipping names case 2 already renamed.
```
egonSchiele, line 275: **O(n²) full AST walks, with the common case as the worst case.**

`bindersOf(other.code)` is a complete `walkNodesArray` pass and it runs inside `siblings.some(...)`, inside the loop over `bindersOf(site.code)`, inside the loop over sites. Measured call counts: 100 at n=10, 400 at n=20, 1,600 at n=40, 6,400 at n=80 — exactly n².

Wall clock for a splice fill: 1.7 ms at n=10, 7.8 ms at 50, 11.7 ms at 100, 47.4 ms at 200. Doubling 100→200 quadruples it.

The part worth noticing: *colliding* fragments are cheap (0.9 ms at n=100) because `.some` short-circuits on the first sibling. *Non-colliding* fragments — the normal case, where each spliced fragment binds its own name — never short-circuit and pay the full n−1 scan with a fresh walk each time. So the fast path is the rare one.

One-line fix: compute `bindersOf(site.code)` once while building `GraftSite` and store it on the record, then this scan is a set intersection over precomputed arrays. `freeNamesOf`/`bindersOf` in the case-1/case-2 loop above would benefit from the same hoist.

Not urgent at "one import per tool" scale; it matters for generated programs with hundreds of spliced statements.

egonSchiele (commented):
Focused security and performance pass at `c7717c0`. Everything below was reproduced by running against the branch.

**One critical finding: the lifting rule can be defeated through object keys.** Details inline on `lift.ts:28`, summarized here because it is the property the whole feature rests on.

A model-supplied record whose *key* contains a `"` breaks out of the printer and becomes executable code. Round-tripped through `JSON.parse`, this key:

```
a": readFile("/etc/passwd"), "b
```

fills a `#v` hole and prints:

```
const x = {
  "a": readFile("/etc/passwd"), "b": 1
}
```

Re-parsing that output yields a genuine `functionCall` node named `readFile`. So `fill` → `toSource` → `runCode` executes attacker-chosen calls from data — exactly what "fillers are never parsed" promises cannot happen.

The escaping battery does not catch it because it only covers string *values*. `stringLiteral` escapes its content; `objectLiteral` passes keys through raw, and `addQuotesToKey` (`agencyGenerator.ts:1237`) wraps in quotes without escaping. That printer is not itself new or wrong — keys arriving from the parser are already escaped (`{ "a\"b": 1 }` round-trips correctly). What is new is `liftValue` feeding it raw runtime strings, so the seam is where the invariant breaks.

### Findings

| | Issue | Severity |
|---|---|---|
| 1 | Object-key injection through `liftValue` | **Critical** — defeats the lifting rule |
| 2 | Compose workflow breaks after any rename | **High** — headline workflow, wrong blame in the error |
| 3 | `is` / `typeTest` capture (from the last round) | **High** — still open |
| 4 | `Infinity` / `NaN` lift to bare identifier tokens | Medium |
| 5 | `computeRenames` case 3 is O(n²) in full AST walks | Medium |
| 6 | Unbounded recursion on nested input | Low |
| 7 | `loadTemplate` traversal + interrupt payload | Informational |

**2 — the compose workflow breaks whenever hygiene fires.** The `__hyg` counter resets per `fill` call, and `assertNoReservedPrefix` runs over the incoming template, so a second fill rejects names the *system* generated in the first:

```
const once  = fill(tpl, { e: parseExpr("tmp") })   // renames tmp → __hyg1_tmp
const twice = fill(once, { later: 2 })
// Error: `__hyg1_tmp` uses the reserved prefix `__hyg`, which templates
//        keep for hygiene. Rename it in the template.
```

"Build the shape first, parameterize last" is the workflow the feature exists for, and it fails as soon as any fill triggers a rename. The error also blames the user's template for a name they never wrote. The guard needs to distinguish names the renamer produced from names the caller supplied — a marker on the `Code` value, or a counter seeded above the highest `__hyg` index already present.

**4 — non-finite numbers lift to identifiers.** `numberLiteral` uses `String(value)`, so `Infinity`, `-Infinity`, and `NaN` print as bare tokens, and `const x = Infinity` re-parses as a `variableName`, not a number. Reachable from model output: `JSON.parse("1e400")` is `Infinity`. Worth rejecting non-finite values in `liftValue` the way `__proto__` is now rejected.

**5 — O(n²) with the common case as the worst case.** Verified: `bindersOf` (a full AST walk per call) runs exactly n² times in the case-3 sibling scan — 100 calls at n=10, 6,400 at n=80. Wall clock for a splice fill: 1.7 ms at n=10, 11.7 ms at 100, 47.4 ms at 200. Doubling 100→200 quadruples the time.

The subtlety is that *colliding* fragments are fast (0.9 ms at n=100) because `.some` short-circuits on the first sibling, while *non-colliding* fragments — the normal case, where each fragment binds its own name — force a scan of all n−1 siblings with a fresh walk each. Fix is one line of hoisting: compute `bindersOf(site.code)` once when building `GraftSite` and reuse it.

Fine for "one import per selected tool" at tens of items; it bites for generated programs with hundreds of spliced statements.

**6** — 20,000-deep nested input gives an uncaught `RangeError: Maximum call stack size exceeded` from the recursion in `liftValue` / `substituteAny` / `stampOrigin`. Not realistic for model output, but the failure mode is a raw stack overflow rather than a diagnosable error, and the input is untrusted by construction.

**7** — `loadTemplate` inherits `resolveInSandbox`'s deliberate non-containment (`..` resolves like `open()`, per `_writeAST allows upward traversal via .. segments`), and the `std::read` interrupt carries the raw `{ dir, filename }` rather than the resolved path. A policy handler that authorizes on `dir` alone can be handed `../../../etc/passwd`. Consistent with the rest of the module, so not a regression — worth a line in the guide for policy authors.

### Checked and clean

Probed without finding problems: the string-escaping battery holds for every case I could construct (including interpolation openers, backslashes, and delimiters); `RESERVED_WORDS` has no gaps I could find that survive an identifier fill and break re-parse (I tried `break`, `continue`, `parallel`, `seq`, `pattern`, `yield`, `this`, `provide`, `eval`, `on`); quoted hole names round-trip for `.`, `#`, `\`, `,`, `:`; array lifting has no key-equivalent path; `computedKey` and object splat entries are unreachable from `liftValue`; identifier fills are properly constrained by `LEGAL_IDENTIFIER` with no anchoring bypass. The `__proto__` fix you added is correct for the `JSON.parse` path, which is the one that matters.

packages/agency-lang/lib/runtime/template/hygiene.test.ts
```diff
@@ -0,0 +1,132 @@
...
+  it("renames a filler binder that redeclares a template binder", () => {
+    const source = `node main() {\n  const tmp = getApiKey()\n  #setup\n  print(tmp)\n}\n`;
+    const out = fillAndPrint(source, { setup: _parseStatements("const tmp = 99") });
+    // The template keeps its spelling; the filler's duplicate is renamed.
+    expect(out).toContain("const tmp = getApiKey()");
+    expect(out).toMatch(/const __hyg\d+_tmp = 99/);
+    expect(out).toContain("print(tmp)");
+  });
+
+  it("does not rename an unrelated same-named binder in another function", () => {
```
egonSchiele, line 90: **This test does not reach the shadowing filter it appears to cover.** `def a` and `def b` are siblings, so a rename owned by `fn:b` is never *active* when the walk enters `fn:a` — `applyScopedRenames`'s `active.filter(rename => !rebound.includes(rename.from))` is never consulted. Deleting that filter entirely leaves all 147 tests green.

The construction that reaches it needs a rename active on entry, which means a **global** binder plus a nested def that rebinds the name:

```typescript
it("a def that rebinds a renamed global keeps its own binding", () => {
  const src = [
    "const tmp = getApiKey()",
    "",
    "def inner(): number {",
    "  const tmp = 5",
    "  return tmp",
    "}",
    "",
    "node main() {",
    "  const r = #e",
    "  return r",
    "}",
  ].join("\n") + "\n";
  const out = fillAndPrint(src, { e: _parseExpr("tmp") });
  expect(out).toMatch(/const tmp = 5/);
  expect(out).toMatch(/return tmp\b/);
});
```

I verified this passes on `c7717c0` and fails with the filter removed. Under the mutation, `inner`'s own `tmp` becomes `__hyg1_tmp` — colliding with the renamed global, so the mutation produces a genuine capture rather than cosmetic noise.

Keep the existing test too; it pins per-scope ownership, which is a different property.

egonSchiele (commented):
Test-quality pass at `c7717c0`. Rather than reading the assertions and guessing, I **mutation-tested** the suite: 30 surgical mutations across the parser, formatter, lifter, filler, hygiene, and builder, running the 147-test template suite against each and reverting.

**Result: 24 killed, 1 equivalent mutant, 5 survivors.** That is a strong suite. Every "delete a security check" mutation was caught, and several catches were multi-test, which means the coverage is redundant rather than incidental.

### What the mutation run proves is genuinely covered

| Mutation | Tests failed |
|---|---|
| Hygiene computes no renames at all | 7 |
| `liftValue` turns numbers into strings | 6 |
| `assertNoReservedPrefix` no-ops | 5 |
| `freeNamesOf` always empty (capture-blind) | 4 |
| Fill-time type check disabled | 4 |
| Rename every binder instead of only colliding ones | 3 |
| Fragment-kind check no-ops | 3 |
| `bindersOf` always empty | 3 |
| Drop `LEGAL_IDENTIFIER` / reserved-word / splice-in-name checks | 1–2 each |
| Stop escaping `${`, the delimiter, or backslashes | 1–3 each |
| `stampOrigin` shallow again / `__proto__` guard removed / AG8001 first-hole-only | 1 each |

Worth calling out specifically: the number-vs-string distinction I flagged as untestable in the first round is now properly pinned — turning `numberLiteral` into `stringLiteral` fails 6 tests. And `M20` (making two colliding fillers share one fresh name, the exact bug from round one) is caught by the distinctness assertion.

The one equivalent mutant was newline escaping: removing it emits a literal newline inside the quotes, which re-parses to the identical string, so the round-trip assertion correctly does not fire. Not a gap.

### The 5 survivors

Each is a real behavior no test pins. Inline comments carry the reproductions; the two most worth fixing:

**Shadowing (`applyScopedRenames`)** — dropping `active.filter(rename => !rebound.includes(rename.from))` breaks nothing. The doc comment promises a rename "stops at an inner def/node that rebinds the same name," and that is untested. A test exists that *looks* like it covers this ("does not rename an unrelated same-named binder in another function"), but the two defs are siblings, so the rename is never active on entry and the filter is never consulted. The construction that does reach it needs a **global** binder plus a nested def that rebinds it — I verified the test below kills the mutant, and under mutation `inner`'s own `tmp` is renamed to `__hyg1_tmp`, colliding with the renamed global.

**`isCode`'s nodes-array guard** — dropping `Array.isArray(...nodes)` breaks nothing, but any `{ type: "agencyProgram" }` value then reaches `value.nodes.map` and throws a `TypeError` instead of lifting as ordinary data.

Also surviving: `positionInferredTypes` first-vs-last occurrence (the limitation you documented last round is documented but not pinned), the `certainTypeOf` interpolation guard, and the `expr`-hole arity check — which I confirmed is **reachable and killable**, not dead defensive code, since `Code` is a plain record an Agency caller can hand-build.

### Missing cases: none of last round's security findings has a test

This is the more important half. All four are silently uncovered:

1. **Object-key injection** — the escaping battery covers string *values* only; the same 11 strings supplied as record keys are the missing mirror.
2. **Compose-after-rename** — `fill.test.ts:86` does fill twice, but with plain values, so no rename fires and the counter-reset bug stays hidden. That test is one `parseExpr` away from catching it.
3. **`is` / `typeTest` capture** — no test; the hole-position battery is the natural home.
4. **Non-finite numbers** — nothing covers `Infinity` / `NaN` / `1e400`.

That is a consistent pattern worth naming: the suite is excellent at proving the *implemented* behavior stays implemented, and has no case that probes an input class the implementation did not anticipate. The escaping battery is the exception — it is adversarial by construction, and it is also the only place a security property is tested by enumerating hostile inputs rather than by asserting one happy path. More of the suite in that shape would have caught the key injection.

### On the execution fixtures

The six `tests/agency/templates/` fixtures are well chosen and not ceremonial. `codeAcrossCheckpoint` asserts on the filled program's printed content rather than a hole count (which a destroyed value could satisfy), and `governedProgram` has the parent handler actually reject a real read, with `/etc/passwd` existing so a successful read would produce visibly different output. Both discriminate.

`composeGuarded` is the one I would strengthen: its holes bind no colliding names, so it exercises grafting but not hygiene-under-composition — the combination where the counter-reset bug lives.

---

### One more thing: `escaping.test.ts` is invisible on GitHub

I could not attach an inline comment to it, and the reason matters. The file contains a **literal NUL byte** at line 44 (`"unicode <NUL> null"` — the raw byte, not the escape `\0`), so git classifies it as binary:

```
Binary files /dev/null and b/.../escaping.test.ts differ
```

The best-designed test file in this PR renders as an unreviewable blob: no diff, no line comments, no blame. Changing the literal to `"unicode \u0000 null"` keeps the file text and tests exactly the same thing, since `stringLiteralValueOf` compares the parsed value.

### The escaping battery's one blind spot: object keys

(Intended as an inline comment on the `nasty` array.)

This battery is the best-designed thing in the suite — the only place a security property is tested by enumerating hostile inputs rather than asserting one happy path, and `stringLiteralValueOf` throwing on an interpolation segment means a mangled-but-parseable result fails loudly. Removing `${`, delimiter, or backslash escaping from the generator all fail here.

The blind spot is that it covers string **values** only. Object **keys** reach the printer by a different path (`objectLiteral` -> `addQuotesToKey`) with no escaping at all, and a key containing a quote breaks out into executable code — reproduction on `lift.ts:28` in the previous review.

The fix is mechanical: run the same `nasty` array through keys as well as values.

```typescript
describe("filler strings survive print and re-parse as KEYS too", () => {
  for (const key of nasty) {
    it(`round-trips key ${JSON.stringify(key)}`, () => {
      const filled = fillHoles(_loadTemplateFromString(template), {
        v: JSON.parse(JSON.stringify({ [key]: 1 })),
      });
      const reparsed = _parseAST(_toSource(filled));
      expect(objectKeysOf(reparsed)).toEqual([key]);   // exactly one entry
    });
  }
});
```

`toEqual([key])` matters more than a containment check: the injection produces a *second* entry, so asserting the whole key list is what catches it. Worth adding a key like `a": evil(), "b` to `nasty` explicitly so the intent is legible.

packages/agency-lang/lib/runtime/template/fill.test.ts
```diff
@@ -0,0 +1,323 @@
...
+  });
+
+  it("allows a partial fill, leaving other holes in place", () => {
+    const filled = fillHoles(load(`node main() {\n  const a = #x\n  const b = #y\n}\n`), {
+      x: 1,
+    });
+    expect(_toSource(filled)).toContain("#y");
+  });
+
+  it("composes: filling the result of a partial fill empties the holes", () => {
```
egonSchiele, line 86: **One `parseExpr` away from catching a real bug.** Filling twice is the right idea, but both fills here supply plain values, so hygiene never fires and the second fill never sees a `__hyg` name.

Make the first fill trigger a rename and it fails on `c7717c0`:

```typescript
it("composes when the first fill renames", () => {
  const tpl = load(
    'node main() {\n  const tmp = 1\n  const a: boolean = #e\n  const b: number = #later\n  return b\n}\n',
  );
  const once = fillHoles(tpl, { e: _parseExpr("tmp") });   // renames tmp
  expect(() => fillHoles(once, { later: 2 })).not.toThrow();
});
```

Actual: `` `__hyg1_tmp` uses the reserved prefix `__hyg` ... Rename it in the template. `` The counter restarts per call and `assertNoReservedPrefix` cannot tell a renamer-produced name from a caller-supplied one, so the second fill rejects the first fill's output. Reported in full on `fill.ts:46`.

The `composeGuarded` execution fixture has the same shape — grafting works, but its holes bind no colliding names, so hygiene-under-composition is untested at both levels.

packages/agency-lang/lib/runtime/template/fill.test.ts
```diff
@@ -0,0 +1,323 @@
...
+    expect(() => fillHoles(load(t), { text: 42 })).toThrow(/#text/);
+  });
+
+  it("checks against an inline annotation too", () => {
+    const annotated = `node main() {\n  f(#count: number)\n}\n`;
+    expect(() => fillHoles(load(annotated), { count: "many" })).toThrow(/number/);
+    expect(fillAndPrint(annotated, { count: 3 })).toContain("f(3)");
+  });
+
+  it("lets an unknowable fragment through — validation, not a guarantee", () => {
```
egonSchiele, line 193: Three untested behaviors cluster around here; all three survive mutation.

**1. `positionInferredTypes` first-vs-last occurrence.** Changing it to last-occurrence-wins breaks no test. You documented the limitation last round — worth pinning it so the documented behavior is the tested one:

```typescript
it("validates against the FIRST position when a name appears twice", () => {
  const src = 'node main() {\n  const a: string = #v\n  const b: number = #v\n  return b\n}\n';
  expect(() => fillHoles(load(src), { v: "hello" })).not.toThrow();  // string wins
  expect(() => fillHoles(load(src), { v: 42 })).toThrow(/string/);
});
```

**2. The `certainTypeOf` interpolation guard.** Making an interpolated string literal report `"string"` instead of `null` breaks nothing, so the guard's effect is unpinned in either direction.

**3. The `expr`-hole arity check.** Removing `if (nodes.length !== 1)` breaks nothing — but it is **not** dead code. `Code` is a plain record an Agency caller can build by hand, and this kills the mutant:

```typescript
it("rejects an expr fragment carrying more than one node", () => {
  const bad = { type: "agencyProgram", kind: "expr",
    nodes: [_parseExpr("1").nodes[0], _parseExpr("2").nodes[0]] };
  expect(() => fillHoles(load(exprTemplate), { v: bad })).toThrow();
});
```

Same family, on `code.ts`: dropping `Array.isArray(...nodes)` from `isCode` breaks nothing, but then `{ type: "agencyProgram" }` reaches `value.nodes.map` and throws a `TypeError` instead of lifting as data. Pinned by:

```typescript
it("a nodes-less object lifts as data, not as Code", () => {
  expect(fillAndPrint(tpl, { v: { type: "agencyProgram" } })).toContain("agencyProgram");
});
```

(Verified: passes on `c7717c0`, fails with the guard removed.)

egonSchiele:
All round-2 findings addressed in 3b1bf6031. The two big ones first, because both changed shape during verification:

**Key escaping (CRITICAL, lift.ts:28)** — fixed, but not in `addQuotesToKey`. Your note said keys arriving from the parser reprint correctly today, and probing why settled the fix site: the AST stores object keys in **source form, escapes intact** (`{ "a\"b": 1 }` stores the key as the four characters `a\"b`). Escaping in `addQuotesToKey` would therefore double-escape every parser-sourced key and break the round trip. The fix lives in the `objectLiteral` constructor instead — the first (and only) seam where raw runtime strings become keys — using the generator's own `escapeStringText`. The escaping battery grew a key-side mirror of every value case plus your exact smuggled-call repro, asserted against the re-parsed AST (`objectKeysOf` throws on any `functionCall` in the reparse and requires exactly one key equal to the escaped attack string).

**Compose-breaking `__hyg` rejection (fill.ts:46)** — you were right that rejection is unfixable as stated: it cannot tell a renamer-produced name from a caller-supplied one, and marking `Code` as hygienized adds state that print/re-parse would strip. Went with your second option, taken further: the rejection is **gone** entirely, replaced by counter seeding — `maxHygieneIndex` scans the template *and every filler* (declaration names included, so last round's finding stays closed) and fresh names start above the max. Collisions become impossible without rejecting anything, and your fill-twice test passes verbatim. One deliberate exception: identifier fillers are plain strings the seed scan cannot see, and they are caller-supplied by definition (the renamer only rewrites trees), so that one path keeps the rejection. The old reject-the-template tests are replaced by seeding tests (template with pre-existing `__hyg7_x` gets `__hyg8_tmp`, etc.).

**`is`-operand capture (hygiene.ts:87)** — real, and the fix ended up in `node.ts` after all, but not in the view you flagged: `expressionChildren`'s deliberate flow-view emptiness is untouched. The `walkNodes` *generator* already descends `typeTestExpression.expression`; `isExpression` was missing from the same list — an asymmetry with its sibling, not a design. Added the descent, and the full 9,287-test suite (including all flow/narrowing suites) stayed green, which is the evidence the checker doesn't ride on that particular absence. Your flagship-capture-via-`is` repro is now a hygiene test and renames correctly. On the battery additions: `is`-operand added; `typeTestExpression` deliberately not — it's a lowering artifact (`lib/types/pattern.ts`) and templates parse with `lower: false`, so it cannot occur in a template AST. Noted in the battery comment.

**The rest:**
- **Shadowing filter test** — yours, verbatim. Confirmed it fails with the filter removed.
- **Fill-twice + hygiene-under-composition** — unit test added (above); the `composeGuarded` execution fixture still uses non-colliding names, which I've left — the unit level now covers the colliding case.
- **The mutation trio** — all three pinned: first-occurrence type validation (your test), the `certainTypeOf` interpolation guard (interpolated string filling a *number* hole must not throw), and the expr arity check against a hand-built `Code` record. Plus the `isCode` `Array.isArray` guard (`{ type: "agencyProgram" }` lifts as data).
- **Non-finite numbers** — rejected in `liftValue` with the `__proto__`-guard shape; test covers `JSON.parse("1e400")` and `NaN`.
- **O(n²) graft scans** — `binders`/`freeNames` hoisted onto `GraftSite` at construction; the collision loops are now intersections over precomputed arrays, so the non-colliding common case no longer pays quadratic re-walks.
- **Copilot's proto-pollution set** — all converted: null-prototype dictionaries for every map keyed by hole/filler/scope names (house pattern per `lib/optimize/registry.ts`), and `Object.hasOwn` replacing `in`/plain lookups in `isFillableHole` and `graftSites`.
- **Style/naming** — `scopeKeyOfChain` uses a for-of loop; `Code.docComment` is `AgencyMultiLineComment` (matching `AgencyProgram`); guide and nav renamed to **Template Agency**.

State: 9,287 unit tests green (17 new), 6 template fixtures plus 3 `is`-expression fixtures green, tsc/make/lint:structure clean.

PR #667, opened 2026-07-24
feat(lsp): semantic tokens so function references get a real color

egonSchiele:
## What

Adds `textDocument/semanticTokens` to the LSP server, so the VS Code extension can color identifiers that resolve to functions.

A TextMate grammar can see `helper(...)` and guess "function". It cannot see this:

```agency
const f = helper
print(f)          // f is a function, but only a typecheck knows that
```

That case is the reason for the feature.

## How

Three new pieces, plus wiring in `server.ts`.

**`lib/utils/identifierSlots.ts`** answers one question: where are the identifier references in this node? It is the third member of the slot-table family after `bodySlots` and `expressionSlots`, and it exists for the reason those two do. From `expressionSlots`' header:

> hand-written expression-position lists drift. The hoisting pass shipped with its own list and three holes were found in one week

The alternative was a per-node-kind switch inside the token handler, which is that same drift-prone list one level over — except the symptom here is a missing color, which nobody files a bug about. Completeness is enforced by the compiler rather than by review: the registry is a `Record<AgencyNode["type"], ...>`, so adding a node kind to the language fails to compile until it is registered.

**`lib/lsp/semanticTokens.ts`** is a pipeline — walk, resolve, sort, encode — with no AST knowledge of its own. Resolution tries the enclosing scope first so a local shadowing a top-level function wins, then the declaration index, then the call syntax itself.

**`lib/lsp/documentStateCache.ts`** keeps a last-good state per document. `updateDocument` drops state whenever a parse fails, which is most of the time while a line is being typed. Serving only current state would blank every color in the file on nearly every keystroke, so the tokens handler reads last-good instead. Colors lagging by a debounce are invisible; colors vanishing are not.

## Three things worth a reviewer's attention

**`SemanticTokensBuilder` does not sort.** It subtracts from whatever was pushed last, so an out-of-order push produces negative deltas and silently garbled color with no error. And `walkNodes` genuinely yields out of source order: an assignment yields its value before the target access chain, so `obj[i] = f()` walks `f` before `i`. The sort is load-bearing, and there is a test that fails without it.

**Token length comes from the identifier name, never `end - start`.** A node's `loc` spans the whole node, so a `functionCall`'s loc covers `helper(y)`. Using the span would paint the arguments.

**Identifiers inside a `valueAccess` get no token.** `obj.a` and `helper(1).invoke()` parse to a single located node whose children carry no `loc` at all. The guard is on `loc` rather than on node kind, so those start producing tokens with no change to this code once the parser carries positions. There is a tripwire test that fails at that moment and says what to do.

## Scope

This version emits **function tokens only**. Every identifier left uncolored is one the TextMate grammar still handles today, so a narrow table means no visible gaps — and it means the extension needs no grammar changes and no `contributes` entry, since `function` is a standard token type every theme already colors.

## Testing

29 new tests across the three modules; the full `lib/lsp` and `lib/utils` suites pass (226 tests), `tsc` and the structural linter are clean.

The assertions decode the wire format and then slice the **source text** at each decoded position, so a test asserts `{ text: "helper", type: "function" }`. Asserting line and column numbers alone would be reading the same `loc` the implementation read, so a systematic position error would agree with itself and pass. The decoder does its own delta arithmetic rather than importing one from the implementation, since a shared decoder inherits the encoder's bug.

Two mutations were run to confirm the suite bites:

| Mutation | Result |
|---|---|
| remove `.sort(byPosition)` | 1 test fails (source-order test) |
| token length off by one | 8 tests fail |

Known limits, stated rather than implied:

- Block-scope shadowing is not resolved — `findContainingScope` only matches scopes named by a top-level definition. There is a skipped test naming this, ready to unskip.
- Nothing here proves VS Code actually paints the color. That needs the manual check below.

## Not in this PR

The extension-side work. Per the plan it is a README note about `editor.semanticTokenColorCustomizations` and nothing else — no client code, no `contributes`, no grammar change. It also cannot land usefully until `agency-lang` cuts a release and the extension bumps its devDependency, so it belongs in a separate PR on that repo after this one ships.

## Manual verification after release

- `"agency.trace.server": "verbose"`, then confirm `semanticTokensProvider` in the initialize result.
- `Developer: Inspect Editor Tokens and Colors` on a bare `f` bound to a function — that is the definitive check, not eyeballing the color.

### DIFF
--- a/packages/agency-lang/lib/lsp/documentState.ts
+++ b/packages/agency-lang/lib/lsp/documentState.ts
@@ -11,6 +11,10 @@ export type DocumentState = {
   semanticIndex: SemanticIndex;
   scopes: ScopeInfo[];
   symbolTable: SymbolTable;
+  /** The document version this state was built from. Lets a consumer
+   *  tell a fresh state from one the debounce has left behind — see
+   *  DocumentStateCache, which serves stale state on purpose. */
+  version: number;
   /** Lint results computed by the diagnostics pass, reused by the
    *  code-action path so a lightbulb request does not re-parse and re-lint
    *  the document. Valid only while the document is still at
--- a/packages/agency-lang/lib/lsp/documentState.ts
+++ b/packages/agency-lang/lib/lsp/documentState.ts
@@ -11,10 +11,6 @@ export type DocumentState = {
   semanticIndex: SemanticIndex;
   scopes: ScopeInfo[];
   symbolTable: SymbolTable;
-  /** The document version this state was built from. Lets a consumer
-   *  tell a fresh state from one the debounce has left behind — see
-   *  DocumentStateCache, which serves stale state on purpose. */
-  version: number;
   /** Lint results computed by the diagnostics pass, reused by the
    *  code-action path so a lightbulb request does not re-parse and re-lint
    *  the document. Valid only while the document is still at
--- a/packages/agency-lang/lib/utils/identifierSlots.test.ts
+++ b/packages/agency-lang/lib/utils/identifierSlots.test.ts
@@ -1,0 +1,108 @@
+import { describe, expect, it } from "vitest";
+import { identifierSlots, REGISTERED_IDENTIFIER_KINDS } from "./identifierSlots.js";
+import { EXPRESSION_NODE_TYPES } from "../types.js";
+import type { AgencyNode } from "../types.js";
+
+function slotsFor(node: unknown) {
+  return identifierSlots(node as AgencyNode);
+}
+
+describe("identifierSlots completeness", () => {
+  it("registers every expression node kind", () => {
+    // The registry is a Record over AgencyNode["type"], so a missing kind
+    // is a COMPILE error — that is the real enforcement and it fails by
+    // name. This test covers the other direction: it catches a kind that
+    // was added to the language and to the registry, but by someone who
+    // did not think about whether it holds an identifier.
+    const missing = EXPRESSION_NODE_TYPES.filter(
+      (kind) => !REGISTERED_IDENTIFIER_KINDS.includes(kind),
+    );
+    expect(missing).toEqual([]);
+  });
+
+  it("returns no slots for an unknown node kind rather than throwing", () => {
+    // Templates and tests can synthesize nodes outside the union.
+    expect(slotsFor({ type: "notARealNodeKind" })).toEqual([]);
+  });
+});
+
+describe("identifierSlots positions", () => {
+  const loc = { line: 3, col: 7, start: 42, end: 48 };
+
+  it("reads a plain variable reference", () => {
+    expect(slotsFor({ type: "variableName", value: "helper", loc })).toEqual([
+      { name: "helper", line: 3, col: 7, scopeOffset: 42, isCall: false },
+    ]);
+  });
+
+  it("reads a call and marks it as one", () => {
+    const slots = slotsFor({
+      type: "functionCall",
+      functionName: "helper",
+      arguments: [],
+      loc,
+    });
+    expect(slots).toEqual([
+      { name: "helper", line: 3, col: 7, scopeOffset: 42, isCall: true },
+    ]);
+  });
+
+  it("does not carry the node span as a length source", () => {
+    // The slot exposes `name`, never end-start. A functionCall's loc
+    // spans `helper(y)`, so a consumer using the span would paint the
+    // arguments. Nothing in the slot makes that possible.
+    const slots = slotsFor({
+      type: "functionCall",
+      functionName: "helper",
+      arguments: [],
+      loc: { line: 0, col: 0, start: 0, end: 999 },
+    });
+    expect(slots[0].name.length).toBe(6);
+    expect(Object.keys(slots[0])).not.toContain("end");
+  });
+});
+
+describe("identifierSlots exclusions", () => {
+  const loc = { line: 1, col: 2, start: 10, end: 14 };
+
+  it("skips `null`, which the parser represents as a variable name", () => {
+    expect(slotsFor({ type: "variableName", value: "null", loc })).toEqual([]);
+  });
+
+  it("skips nodes with no loc — this is what the valueAccess gap looks like", () => {
+    // walkNodes descends into a valueAccess base, so an unlocated
+    // variableName really does arrive here. The guard is on loc rather
+    // than on node kind, so these start producing slots automatically
+    // once the parser carries positions for them.
+    expect(slotsFor({ type: "variableName", value: "obj" })).toEqual([]);
+    expect(
+      slotsFor({ type: "functionCall", functionName: "invoke", arguments: [] }),
+    ).toEqual([]);
+  });
+
+  it("skips a template hole used as a callee", () => {
+    expect(
+      slotsFor({
+        type: "functionCall",
+        functionName: { type: "hole", name: "name" },
+        arguments: [],
+        loc,
+      }),
+    ).toEqual([]);
+  });
+
+  it("skips declaration sites, which the grammar already colours", () => {
+    expect(
+      slotsFor({
+        type: "function",
+        functionName: "helper",
+        parameters: [],
+        body: [],
+        loc,
+      }),
+    ).toEqual([]);
+    expect(
+      slotsFor({ type: "graphNode", nodeName: "main", parameters: [], body: [], loc }),
+    ).toEqual([]);
+  });
+});
--- a/packages/agency-lang/lib/utils/identifierSlots.test.ts
+++ b/packages/agency-lang/lib/utils/identifierSlots.test.ts
@@ -1,23 +1,94 @@
 import { describe, expect, it } from "vitest";
-import { identifierSlots, REGISTERED_IDENTIFIER_KINDS } from "./identifierSlots.js";
-import { EXPRESSION_NODE_TYPES } from "../types.js";
+import { identifierSlots, IDENTIFIER_BEARING_KINDS } from "./identifierSlots.js";
+import { parseAgency } from "../parser.js";
+import { walkNodes } from "./node.js";
 import type { AgencyNode } from "../types.js";
 
 function slotsFor(node: unknown) {
   return identifierSlots(node as AgencyNode);
 }
 
+/** Synthetic nodes of each identifier-bearing kind, shaped as the parser
+ *  produces them. Used to prove the registry entry still does something. */
+const SAMPLE_NODES: Record<(typeof IDENTIFIER_BEARING_KINDS)[number], unknown> = {
+  variableName: {
+    type: "variableName",
+    value: "helper",
+    loc: { line: 0, col: 0, start: 0, end: 6 },
+  },
+  functionCall: {
+    type: "functionCall",
+    functionName: "helper",
+    arguments: [],
+    loc: { line: 0, col: 0, start: 0, end: 8 },
+  },
+};
+
 describe("identifierSlots completeness", () => {
-  it("registers every expression node kind", () => {
-    // The registry is a Record over AgencyNode["type"], so a missing kind
-    // is a COMPILE error — that is the real enforcement and it fails by
-    // name. This test covers the other direction: it catches a kind that
-    // was added to the language and to the registry, but by someone who
-    // did not think about whether it holds an identifier.
-    const missing = EXPRESSION_NODE_TYPES.filter(
-      (kind) => !REGISTERED_IDENTIFIER_KINDS.includes(kind),
+  // NOTE on what enforces what. The registry is
+  // `{ [K in AgencyNode["type"]]: SlotExtractor<K> }`, so a node kind
+  // added to the language fails to COMPILE until it is registered, and
+  // each extractor sees its own node type so an AST field rename is also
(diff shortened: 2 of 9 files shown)


egonSchiele:
Thanks — all four were real. Verified each against the code before applying rather than taking them as read, and two of them found actual bugs.

**Resolved name vs typed name** (comment on `semanticTokens.ts:131`) — correct, and the suggested approach is exactly right. Probing it:

- `const print = helper` → `scope.lookup("print")` returns `functionRefType{ name: "helper" }`
- a plain prelude `print("hi")` → `scope.lookup("print")` returns `undefined`, because the prelude is ambient rather than a scope binding

So asking the stdlib question of the resolved name separates the two cleanly, with no special case. Fixed, with a test that I confirmed bites by reverting the fix under it.

**Imported names also shadow** (`semanticTokens.ts:88`) — also correct. The check excluded only `source: "local"`, so an imported `print` was still dimmed as stdlib. Now any name the file declares *or* imports is excluded. Erring toward "not stdlib" is deliberate: failing to dim library code is a smaller wrong than dimming the user's own.

The same comment's note about reading through the prototype chain was right too — the index is a plain object, so `constructor` would have resolved off the prototype. Switched to `Object.hasOwn`.

**`refresh()` rejection** (`server.ts:219`) — this one is worth flagging for anyone else who hits it, because the typings actively mislead. `semanticTokens.d.ts` declares `refresh(): void`, and my first attempt at `.catch()` was rejected by tsc for that reason. But the implementation is:

```js
refresh: () => {
    return this.connection.sendRequest(SemanticTokensRefreshRequest.type);
}
```

A real promise, which rejects on clients that do not support the request — an unhandled process-level rejection in a server meant to run for hours. Now caught and reported through the LSP console, with a comment explaining why the code disagrees with the declared type.

**Null-prototype dictionaries** (`documentStateCache.ts:29`) — applied. Being straight about this one: it is hardening, not a fix. The keys are always document URIs, which begin `file://`, so none can collide with a prototype key today. But the key space is not ours to constrain and the guard costs nothing. Kept as objects rather than `Map`s, since `CLAUDE.md` asks for objects over maps.

229 tests pass, `tsc` and the structural linter clean.

packages/agency-lang/lib/utils/identifierSlots.test.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+describe("identifierSlots completeness", () => {
+  it("registers every expression node kind", () => {
+    // The registry is a Record over AgencyNode["type"], so a missing kind
+    // is a COMPILE error — that is the real enforcement and it fails by
+    // name. This test covers the other direction: it catches a kind that
+    // was added to the language and to the registry, but by someone who
+    // did not think about whether it holds an identifier.
+    const missing = EXPRESSION_NODE_TYPES.filter(
```
egonSchiele: **This test can never fail.**

`REGISTRY` is `Record<AgencyNode["type"], SlotExtractor>`, so `Object.keys(REGISTRY)` contains every node type by construction — the compiler guarantees it. And every entry in `EXPRESSION_NODE_TYPES` (`types.ts:125-152`) is an `AgencyNode` type. So this filter is always empty no matter what anyone does to either list.

The comment above states the intent: catching "a kind that was added to the language **and to the registry**, but by someone who did not think about whether it holds an identifier." It doesn't do that — registering a new expression node as `none` passes. The only way this fails is if `EXPRESSION_NODE_TYPES` (typed `readonly string[]`, so unchecked) gains a string that isn't a node type at all.

This matters more than a normal dead test, because it's the one test standing behind the module's central claim. Two options:

- Delete it as redundant with the compile-time check, and let the module doc say the `Record` is the whole enforcement.
- Make it bite: assert that the expression kinds *known to hold identifiers* map to non-`none` extractors, with an explicit opt-out list that a newly added node kind isn't on by default. That's the version that actually catches the thoughtless registration.

packages/agency-lang/lib/lsp/documentState.ts
```diff
@@ -11,6 +11,10 @@ export type DocumentState = {
   semanticIndex: SemanticIndex;
   scopes: ScopeInfo[];
   symbolTable: SymbolTable;
+  /** The document version this state was built from. Lets a consumer
+   *  tell a fresh state from one the debounce has left behind — see
+   *  DocumentStateCache, which serves stale state on purpose. */
+  version: number;
```
egonSchiele: **`version` is dead as shipped.**

It's added here, documented as letting a consumer "tell a fresh state from one the debounce has left behind," set in `server.ts`, and exercised across all eight `documentStateCache.test.ts` cases — but nothing in production reads it. `getSemanticTokens` takes the last-good state unconditionally and never compares against `doc.version`.

Fine as groundwork, but right now the tests assert on a field whose only consumer is the tests. Either give it a real use (staleness threshold or position clamping — see the review body) or drop it and have the cache tests assert on state identity instead.

packages/agency-lang/lib/utils/identifierSlots.ts
```diff
@@ -0,0 +1,222 @@
...
+  /** True when the source wrote this as a call, `name(...)`. Lets a
+   *  consumer distinguish a call site from a bare reference without
+   *  re-inspecting the AST. */
+  isCall: boolean;
+};
+
+/** Extracts the identifier references a node contributes itself. Child
+ *  nodes are NOT visited here — `walkNodes` reaches them and asks this
+ *  table about each one in turn. */
+type SlotExtractor = (node: any) => IdentifierSlot[];
```
egonSchiele: **`any` here gives up the type safety the registry just bought.**

The `Record<AgencyNode["type"], …>` gets compile-enforced completeness over *keys*, which is genuinely stronger than the runtime lists `expressionSlots` uses. But each extractor receives `any`, so `node.value`, `node.functionName`, and `node.loc.line` are all unchecked. A rename of `functionName` in the AST compiles clean and silently returns zero slots — which is exactly the "missing color nobody files a bug about" failure this module exists to prevent, arriving through the door the registry left open.

A mapped type recovers it:

```ts
type SlotExtractor<K extends AgencyNode["type"]> =
  (node: Extract<AgencyNode, { type: K }>) => IdentifierSlot[];

type Registry = { [K in AgencyNode["type"]]: SlotExtractor<K> };
```

The `located()` helper would then narrow properly as a type predicate rather than being decorative.

packages/agency-lang/lib/utils/identifierSlots.ts
```diff
@@ -0,0 +1,222 @@
...
+export type IdentifierSlot = {
+  /** The identifier text as written. Also the token's length. */
+  name: string;
+  /** 0-based line in the user's file. */
+  line: number;
+  /** 0-based column of the identifier's first character. */
+  col: number;
+  /** Offset for `findContainingScope`. Only comparable against other
+   *  offsets from the SAME parse — see the parse-mode note in
+   *  docs on Background A. */
```
egonSchiele: Dangling reference: "Background A" is a section of the planning document, which doesn't exist in this repo — a future reader has nothing to look up. The nearby `docs/dev/locations.md` pointer resolves fine (`packages/agency-lang/docs/dev/locations.md`), so just point at that.

egonSchiele (commented):
## Review

Design is sound and the three risks flagged in the description are all real and all correctly handled. The sort is load-bearing, length-from-name is right, and asking the stdlib question of the **resolved** name — so `const print = helper` doesn't get `defaultLibrary` — is a catch a lot of implementations would miss.

Four findings inline, plus the following.

### The stale-state trade-off's actual failure mode isn't tested

Serving last-good tokens is the right call. But the risk isn't "colors lag" — it's that after a **deletion**, last-good tokens carry line/col coordinates pointing past the end of lines or past EOF in the current buffer. Every test builds state and reads tokens from the same source string, so nothing exercises tokens-against-a-changed-document at all.

VS Code is generally tolerant here, but that's worth confirming rather than assuming, since it's the load-bearing assumption behind an entire new module. Two cheap additions:

- A test that decodes tokens from state built on source A and asserts every position is in range for a shorter source B (or that the handler clamps).
- Add to the manual verification list: select 20 lines, delete, confirm no misplaced colors and nothing in the server log.

This is also where the new `version` field could earn its keep — stop serving last-good past some staleness threshold, or clamp emitted positions to the current document.

### Performance, minor

`makeScopeFinder` correctly removes the map rebuild, but the returned finder still scans all scopes per call, so the pass is O(identifiers × scopes) — better than before, not linear. `PRELUDE_NAMES.includes(name)` (`prelude.ts:25`, a `readonly string[]`) is a linear scan per identifier; a module-level `Set` is free. And `getSemanticTokens` materializes four intermediate arrays per request.

None of this matters at current file sizes. Worth a comment rather than a change, so the next reader doesn't assume it's already linear.

### Smaller notes

- `refresh()` fires on every watched-file event, including ones that changed nothing relevant. Harmless but chatty. The defensive `instanceof Promise` catch around the mis-typed return value is a good catch and correctly explained.
- Server wiring is untested: neither the `getLastGood` path nor the `catch` fallback in `connection.languages.semanticTokens.on` has coverage. Consistent with how the rest of `server.ts` is treated, so acceptable rather than a gap.
- `isFunctionReference`'s `slot.isCall` fallback colors `notAFunction()` as a function in a file with type errors. Defensible — the syntax says call — but worth a line in the doc comment saying so.
- `anyCurrent()` builds a full `Object.values` array to take `[0]`. Trivial, but a `for...in` with an early return is the same length.
- Spelling is mixed across new comments: `colour` in `documentStateCache.ts`, `color` in `identifierSlots.ts`. Pick one.

### What's good

Worth saying explicitly because it's unusual: having the tests decode the wire format and then **slice the source text** at each position — rather than asserting line/col numbers read from the same `loc` the implementation read — is the difference between a suite that catches a systematic position error and one that agrees with it. Writing the decoder's delta arithmetic out longhand instead of importing the encoder's is the same instinct. The two mutation results in the description are the right way to show a suite bites.

The `valueAccess` gap being guarded on `loc` rather than on node kind, with a tripwire test that fails the moment the parser improves, is a clean way to ship a known limit without it becoming permanent.

### Verdict

Good to merge after the completeness test and the `version` question. The `SlotExtractor` typing and the stale-position test are worth doing but wouldn't block.

egonSchiele:
All four inline findings were right, and the completeness one was the sharpest — thanks. Pushed as `944253cbb`.

### The completeness test could never fail

Correct, and worse than you framed it: the comment I wrote above it claimed to catch "a kind added to the registry by someone who did not think about whether it holds an identifier," which is precisely what it did not do. I asserted a property the compiler already guarantees and then described it as covering something else.

Took the second option. Two tests now, both verified by mutation:

- each kind in `IDENTIFIER_BEARING_KINDS` must still produce a slot from a synthetic node — flipping `functionCall: none` compiles fine and this catches it
- a corpus check that parses real code, collects every node the walk reaches carrying both a name and a position, and asserts the table accounts for all of them

Flipping `functionCall` to `none` fails 4 tests now instead of 0. I also wrote down what these still cannot catch — neither can know that some future node kind carries a name — rather than implying the module is fully guarded.

### `any` in the extractors

Right, and it undercut the module's whole claim. Your mapped type works as written. Confirmed it bites: renaming `functionName` to `calleeName` in the extractor now gives

```
identifierSlots.ts(115,19): error TS2339: Property 'calleeName' does not exist on type 'FunctionCall'.
```

One contained cast remains at the dispatch site, since TypeScript cannot see that `REGISTRY[k]` and a node of kind `k` agree. The extractors themselves are fully typed, which was the point. `located()` is now a real type predicate.

### `version` was dead — dropped it, and fixed the real problem behind it

You were right that it was dead and that the tests were asserting on a field only the tests used. But the stale-position finding in your review body is the more important half, and it made the fix obvious: `version` was never the right instrument.

`getSemanticTokens` now takes the current text and drops any token that does not fit — a line the buffer no longer has, or a column past the end of a line that shrank. No staleness threshold, so no version needed.

Dropping rather than clamping, deliberately: a clamped token paints whatever real word now sits at that position, which is a wrong color on live text; a dropped one just leaves the grammar's color alone.

Three tests, and I had to strengthen one of them. The shortened-line test initially asserted on `token.col + token.text.length`, but `text` comes from re-slicing the source, so an over-long token slices to a *shorter* string and the assertion passed even with the bounds check disabled. The decoder now carries the raw encoded length. With that fixed, disabling the check fails both stale tests instead of one.

### Smaller notes

All applied. The dangling "Background A" pointed at the planning doc, which lives in the extension repo — now points at `docs/dev/locations.md`. The `isCall` fallback caveat is documented, including why coloring `notAFunction()` is the right trade. `PRELUDE_NAMES` membership is hoisted into a lookup object. `anyCurrent` uses `for...in` with an early return. Spelling normalized to `color` — all 16 were mine.

On performance being O(identifiers × scopes): I added the comment rather than the binary search, per your call. Measured 0.3–0.7 ms on real stdlib files and 5.4 ms on a generated 1200-line file, and the comment records both so the next reader has the numbers.

### Not addressed

The manual verification step (select 20 lines, delete, check the server log) lives in the plan doc in the extension repo, and it still needs a human with a running editor after release. The automated half of that concern is now covered by the three tests above.

234 tests pass, `tsc` and the structural linter clean.

Issue #668, opened 2026-07-24
walkNodes descent gaps found by the walker-coverage tripwire (5 fields)


egonSchiele:
The walker-coverage tripwire added in the Template Agency follow-ups PR (corpus invariants in `lib/utils/expressionSlots.test.ts`, "walker coverage" describe block) found five places where `walkNodes` (`lib/utils/node.ts`) never descends into genuine expression content. Each is recorded in `KNOWN_WALKER_GAPS` in that test file, and a staleness guard asserts each entry still shields something — so fixing a gap fails the suite until its entry is deleted, and the fix cannot land silently.

Why these are not fixed in the discovering PR: `walkNodes` backs the symbol-table resolver (which populates `scope` on variable references and so drives codegen — `__stack.locals.foo` vs bare identifiers), template hygiene's free-name analysis, and the LSP. Adding a descent case makes previously unvisited expressions start getting scope-resolved, which can change generated code. Each fix deserves its own review of that surface.

The five gaps:

1. **`functionParameter.defaultValue`** — parameter default expressions are never walked (~730 expression nodes across the stdlib corpus). Template hygiene cannot see names in defaults; the symbol table never visits them.
2. **`function.docString`** — docstring interpolation segments are never walked, but docstring interpolations ARE evaluated (`hasDocStringInterpolation` in `typescriptBuilder.ts` gates codegen on them). Names inside `${...}` in a docstring are invisible to hygiene and the symbol table walk.
3. **`assignment.accessChain`** (slice entries) — slice-assignment bounds (`arr[a:b] = x`) are not walked. Index and methodCall chain entries are; `valueAccess` (the read side) walks slices. Found by the slot-agreement invariant: `expressionSlots` enumerates these bounds, the walker skips them.
4. **`matchBlockCase.guard`** — unlowered match-arm guard expressions are not walked. The lowered if-chain form is, so the compile pipeline is unaffected; the unlowered view (formatter, templates, hygiene) misses names in guards.
5. **`tag.arguments`** — `@validate`/tag annotation arguments (validator function references and values) are never walked.

For each fix: add the descent in `walkNodes`, run the full unit suite plus the agency suite in CI, delete the corresponding `KNOWN_WALKER_GAPS` entry (the staleness guard forces this), and call out in the PR what the symbol table now resolves that it previously did not.

[...]

PR #669, opened 2026-07-24
Template Agency follow-ups: walker tripwire, pattern-binder hygiene, origin-aware errors

egonSchiele:
Three follow-ups from Template Agency (#665), one commit each, plus an audit commit.

## 1. Walker-coverage tripwire (`lib/utils/expressionSlots.test.ts`)

Template hygiene's free-name analysis is exactly as complete as `walkNodes`' descent: a position the walker misses under-reports free names, no test fails, and a filler silently captures a template binder — capture avoidance failing open. That guarantee was a hand-enumerated 15-position list in `hole.test.ts`, which has the same drift disease the slot table was built to kill.

Two structural corpus invariants now guard it, over the whole corpus in BOTH parse modes (lowered = compile pipeline's view; unlowered = template/hygiene's view, patterns and comprehensions intact):

- **Slot-table agreement** — every `expressionSlots` expr of a walked node must itself be walked. A consistency check between the two tables, not a reachability proof; it found the slice-assignment gap on its first run.
- **Structural reachability** — crawl every AST field generically; every expression-typed node found must be walked unless its owning field has a recorded ruling. This is the invariant that catches positions nobody registered anywhere.

Exclusions are keyed `"ownerType.field"` (a bare `"pattern"` would silently cover `isExpression` and `typePattern` too), each with a one-line justification.

**`walkNodes` is deliberately untouched.** Per owner direction, walker fixes are their own PRs — `walkNodes` feeds the symbol table and codegen scope resolution, so every descent change is a compiler change. The five real gaps the invariants found are recorded in `KNOWN_WALKER_GAPS` and tracked in #668:

| Gap | What's invisible there |
|---|---|
| `functionParameter.defaultValue` | ~730 expression nodes in parameter defaults |
| `function.docString` | docstring interpolations (which the builder DOES evaluate) |
| `assignment.accessChain` | slice-assignment bounds `arr[a:b] = x` (read-side slices are walked) |
| `matchBlockCase.guard` | unlowered match-arm guards (lowered if-chain form is walked) |
| `tag.arguments` | `@validate`/tag annotation arguments |

**Hard to forget, in both directions:** a new node kind missing walker descent fails the reachability invariant by name; and a *staleness guard* asserts each `KNOWN_WALKER_GAPS` entry still shields at least one unwalked node — so the walker-fix PR cannot land without deleting its entry, and a stale entry fails loudly (verified by inserting a bogus entry). The hole-position battery stays, demoted to parser coverage; `freeNamesOf`'s load-bearing comment now points at the invariants.

A `comprehension.agency` generator fixture was added because the corpus contained no comprehension (the liveness check demanded one).

## 2. Pattern binders in hygiene (`lib/runtime/template/hygiene.ts`)

Templates parse with `lower: false`, so destructuring reaches hygiene as an `assignment` whose `variableName` is the `"__destructured"` sentinel with the real binders in `pattern` — and `bindersOfNode` read only `variableName`. A template `const { key } = getSecrets()` bound a name hygiene could not see, so a filler mentioning `key` captured the secret.

`patternBinders` now feeds collision detection from object/array/rest patterns, for-loop binders, and comprehension binders (identical `itemVar` shape). The one subtle rename case: shorthand `{ tmp }` is simultaneously the property being READ and the binder, so a renamed shorthand expands to `{ tmp: __hygN_tmp }` — same read, fresh binder (the generator collapses back to shorthand only when key and name match, so it prints as written). Rest identifiers rename via `isNameField`; every other pattern-held binder is a `variableName` node the generic walkers already rewrite. `directBinders` (the shadow-stop) and `maxHygieneIndex` (counter seeding) get pattern binders through the same helper.

**Scope ruling:** result-pattern bindings (`is success(v)`) and match-arm binders stay untracked — both are branch-scoped, which needs flow-aware rename planning. The guide's blanket destructuring caveat is replaced by that narrower honest limit.

## 3. Origin-aware fill errors and `holesOf` (`fill.ts`, `holes.ts`, `stdlib/agency.agency`)

`fill` deep-stamps `loc.origin = { kind: "filler", name }` on every grafted node, but nothing read it. Now two readers do:

- Fill-path errors append ``(in code grafted by the fill for `#helpers`)`` when the offending hole carries a filler origin, and the unknown-name error annotates its hole list the same way. The old redundant tail on the type error (which repeated the hole's own name; unreferenced by any test) is replaced.
- `HoleInfo` gains `origin: string | null`, mirrored in the Agency-side type — so after `fill(mainTpl, { helpers: partialGuard })`, `holesOf` reports `{ name: "minutes", origin: "helpers", ... }` and a model composing templates knows which sub-template each remaining hole came from. Proven end-to-end by the `holesOfOrigin` execution fixture.

Two deliberate semantics, documented in code, guide, and dev doc: **outermost graft wins** (re-grafting overwrites the stamp, so origin means "the fill this node most recently arrived through" — the one the current caller performed), and **best-effort** (loc-less inner nodes carry no stamp and yield `origin: null`). Compile-side attribution stays impossible until an AST-in compile entry point exists: `toSource` prints and `runCode` re-parses, and `loc` does not survive that boundary.

## Notes for review

- One deviation from the arrays-not-sets rule: the invariants use JS `Set` for walked-node membership, which is by object identity over tens of thousands of nodes per file — `array.includes` would make them quadratic.
- `patternBinders` is new rather than reused: `patternLowering.ts`'s `walkPattern` is module-private and visits match-position kinds (literals, result patterns) that bind nothing for hygiene.
- Full unit suite green (9,308 tests), `lint:structure` clean, `tsc` clean, `make` + `make doc` run for the stdlib change. CI runs the agency suite.


packages/agency-lang/docs/site/guide/templates.md
```diff
@@ -147,7 +151,9 @@ node main() {
...
 
-One current limit: destructuring-pattern binders are not tracked for collisions. If a filler destructures into a name the template also uses, rename one of them yourself.
+Destructuring binders are tracked like any other name: `const { key } = …`, array and rest patterns, and for-loop and comprehension binders all participate in collision detection. One wrinkle worth knowing: renaming a shorthand `{ key }` in place would change which property is *read*, so a renamed shorthand expands to `{ key: __hyg1_key }` — same read, fresh binder.
+
+Two binder forms are not yet tracked for collisions: names bound by result patterns (`if (r is success(v)) { … }` binds `v`) and names bound inside match arms. If a filler or template uses one of those names on the other side, rename it yourself.
```
egonSchiele, line 156: This limit is understated, and the understatement matters because a reader will conclude the failure is "a collision I have to notice myself."

The actual behaviour is worse than a missed collision: the rename walkers rewrite *uses* of the name inside the arm while leaving the pattern binding alone, so the generated program reads a different variable. I ran this on `main` (so this is pre-existing, not introduced here):

```
node main() {
  const v = 1
  const x = #e          // filled with `v + 1`
  const r = load()
  if (r is success(v)) {
    print(v)
  }
  print(v)
}
```

produces:

```
node main() {
  const __hyg1_v = 1
  const x = v + 1
  const r = load()
  if (r is success(v)) {
    print(__hyg1_v)     // <-- was the pattern binding, now the outer variable
  }
  print(__hyg1_v)
}
```

`print(v)` inside the arm meant the success value; after the fill it means the outer `const`. Nothing errors — the program compiles and silently computes the wrong thing.

Two options, either fine by me:
1. Cheapest correct fix, and it fits the existing machinery: treat a result-pattern / match-arm binding as a **shadow stop**, the way `directBinders` already stops an outer rename at an inner def that rebinds the name. Renames simply do not descend past a binding of the same name. That closes the silent-wrong-value case without needing flow-aware rename planning, which is the reason given for deferring full tracking.
2. If it stays deferred, say what actually happens here and in the dev doc: not "the collision is missed" but "an outer rename rewrites uses inside the arm and the arm's binder does not follow, so the arm reads the outer variable."

Worth a follow-up issue either way, since a silently-wrong program is a different severity from an un-renamed collision.

packages/agency-lang/lib/utils/expressionSlots.test.ts
```diff
@@ -324,3 +342,220 @@ describe("expressionSlots: corpus invariants", () => {
...
+  "assignment.pattern": "binding pattern: variableName members are binders, not uses",
+  "forLoop.itemVar": "for-loop binder (string or pattern), not a use",
+  "comprehension.itemVar": "comprehension binder, same shape and ruling as forLoop.itemVar",
+  "matchYield.typeSource":
+    "type-checker view of the arm expression (lib/types/matchYield.ts); the executable " +
+    "copy flows through the hoisted temp assignment, which is walked",
+  "objectPatternProperty.value":
+    "pattern property content: a literal matcher or a binder, not a use",
+  "typePattern.pattern": "type-pattern binder, not a use",
+  "functionCall.block":
```
egonSchiele, line 381: This ruling is accurate but the mechanism makes it far broader than the justification. `childViaFor` is sticky (line 426), so excluding the field `functionCall.block` removes the **entire inline-block subtree** from the reachability invariant, not just the wrapper node the comment is about.

Measured on the corpus (lowered parse): 807 expression-typed nodes live under a `functionCall.block`. Exactly 52 of them are unwalked, and all 52 are `blockArgument` wrapper nodes. So the entry correctly rules on 52 nodes and silently drops coverage of the other 755 — which are walked today, and which the invariant would otherwise protect against a future regression. Inline `block:` bodies are ordinary user code; losing coverage there is the opposite of what this describe block is for.

The ruling is about a node **type**, not a field. Suggest a second table:

```ts
const WALKER_EXCLUDED_TYPES: Record<string, string> = {
  blockArgument:
    "wrapper node is not itself yielded; its BODY is walked through bodySlots (blockAncestor)",
};
```

and skip in the reachability invariant on `WALKER_EXCLUDED_TYPES[node.type]` instead of shielding the subtree. Then the 755 stay covered and the 52 stay ruled on.

packages/agency-lang/lib/utils/expressionSlots.test.ts
```diff
@@ -324,3 +342,220 @@ describe("expressionSlots: corpus invariants", () => {
...
+}
+
+function isKnownGap(ownerType: string, key: string): boolean {
+  return Object.hasOwn(KNOWN_WALKER_GAPS, `${ownerType}.${key}`);
+}
+
+type StructuralVia = "clear" | "excluded" | "knownGap";
+
+function childViaFor(current: StructuralVia, ownerType: string, key: string): StructuralVia {
+  if (current !== "clear") return current;
```
egonSchiele, line 426: Same stickiness point as the `functionCall.block` comment, applied to `KNOWN_WALKER_GAPS`.

`assignment.accessChain` is recorded as a gap because slice bounds are not walked. Measured on the corpus: 24 expression nodes live under an assignment's `accessChain`, 4 are unwalked (the slice bounds, exactly as the entry says), and the other 20 — index expressions and method-call arguments — are walked but now shielded. So the entry buys back its 4 and costs 20.

The staleness guard is unaffected either way (it only needs one still-unwalked node), so narrowing is free. Two ways:
- Key gaps by the deepest field that is actually broken (`accessChain` slice `start`/`end` rather than the whole chain), or
- Make `via` non-sticky for gaps: shield only nodes whose *immediate* owning field is the recorded one, and let anything deeper be checked normally.

The second is probably simpler and applies uniformly. `function.docString` and `functionParameter.defaultValue` are whole-subtree gaps where stickiness is correct, so whichever way you go, keep those working.

packages/agency-lang/lib/runtime/template/fill.ts
```diff
@@ -29,12 +29,36 @@ import {
...
+/** Attribution for errors that anchor to a node carried in by a graft:
+ *  its loc.origin (stamped by stampOrigin) names the fill the node most
+ *  recently arrived through — re-grafting overwrites the stamp, so in a
+ *  nested composition the OUTERMOST graft wins, which is the one the
+ *  current caller performed and can act on. Best-effort: loc-less inner
+ *  nodes carry no stamp and get no suffix. Only the fill path can read
+ *  this — toSource/runCode re-parse from text, which drops loc entirely;
+ *  compile-side attribution needs the fragment-checker entry point
+ *  (recorded follow-up). */
+function originSuffix(loc: SourceLocation | undefined): string {
```
egonSchiele, line 41: Placement problem: `originSuffix` landed between the module's big doc comment (the "rules, in order of how much damage getting them wrong would do" block) and `fillHoles`. That comment documents `fillHoles` and is now attached to `originSuffix`, so `fillHoles` — the entry point of the whole file — reads as undocumented, and the substitution rules read as if they describe an error-message helper.

Move `originSuffix` below `fillHoles` (or above the rules block) so the doc comment sits on the function it describes.

packages/agency-lang/lib/runtime/template/fill.ts
```diff
@@ -29,12 +29,36 @@ import {
...
+  if (!loc?.origin || loc.origin.kind !== "filler") return "";
+  return ` (in code grafted by the fill for \`#${loc.origin.name}\`)`;
+}
+
 export function fillHoles(code: Code, values: Record<string, unknown>): Code {
+  const holes = findHoles(code.nodes);
```
egonSchiele, line 47: `findHoles` now runs a full `walkNodesArray` of the template on every `fillHoles` call, but its result is only used inside the `!present.includes(name)` error branch. `holeNames` already walks the tree once, so this doubles the walk on the success path to build data for a message that is usually never produced.

Hoist it into the branch:

```ts
if (!present.includes(name)) {
  const holes = findHoles(code.nodes);
  ...
}
```

Small, but this file already made the opposite trade deliberately — `GraftSite` hoists `bindersOf`/`freeNamesOf` specifically so the non-colliding path does not pay for the colliding one (hygiene.ts). Same principle.

egonSchiele (commented):
Reviewed all three commits against the code they touch, and ran a couple of measurements locally to check two claims. The shape is right: the tripwire genuinely found five gaps, the decision to record them rather than fix `walkNodes` in this PR is the correct call, and the staleness guard is the part that makes that decision safe. Pattern-binder hygiene looks correct, and the shorthand expansion is handled the way it has to be.

Five comments inline. Two are worth acting on before merge (the sticky exclusions and the docstring placement), one is a factual sharpening of a documented limit, and two are small.

**One thing I could not verify from the diff:** the `holesOfOrigin` fixture asserts `holes.length != 1`, which pins that a fill leaves exactly one hole. If a future edit to `holesOfOriginInner.agency` adds a hole, the failure message ("wrong hole count") will read as a fixture problem rather than an origin problem. Minor, and arguably the right strictness.

**What I checked and found fine:**
- Every `originSuffix` call site passes a `SourceLocation | undefined` from a hole that actually anchors the error; `substituteAny` correctly uses `value.loc` rather than a hole variable that is not in scope there.
- The old `(in the fill for \`#name\`)` tail was unreferenced by any test or doc, so replacing it is safe.
- `comprehension` really does carry `indexVar?: string` (`lib/types/comprehension.ts:23`), so the shared `forLoop`/`comprehension` branch in `bindersOfNode` is not reading a field that does not exist.
- `formatObjectPattern` collapses `key: value` back to shorthand only when the names match (`lib/backends/agencyGenerator.ts:909-914`), so the expansion prints as the tests expect and an unrenamed pattern round-trips unchanged.

PR #671, opened 2026-07-24
Reify: std::agency describe() reports a module's exports as data

egonSchiele:
The reify primitive from the Template Agency spec's deferred list: ask what a module exports without reading its source, so generators can be shaped by what a module *contains* instead of hand-maintained lists.

## The API

One new export from `std::agency`:

```
export idempotent def describe(source: string): Result<ModuleInfo>

type ExportInfo = {
  name: string,
  kind: "def" | "node" | "type",
  signature: string,        // printed declaration, no keywords, no docs
  docstring: string | null, // docstring (defs/nodes) or doc comment (types)
  effects: string[],        // same names + "unknown" sentinel as getEffects
  destructive: boolean,
  idempotent: boolean
}

type ModuleInfo = {
  description: string | null,   // module @module doc comment text
  exports: ExportInfo[]         // source order
}
```

Underscore-prefixed exports are omitted — the same rule `agency doc` applies (they are lowering targets like `_guard`, not caller surface).

## What it unlocks

The spec's killer use case becomes writable: handler completeness as a property of the *generator*. `describe(toolsSource)` gives the per-export effect lists; a loop emits one handler arm per effect into a `#...handlerArms` splice; an effect with no arm cannot exist by construction. Given that CLAUDE.md treats a skipped handler as a critical bug, that loop is the safety payoff this feature exists for. Second-order: agents get "what does this module export and what can it raise" as a cheap tool, and the spec's fill-time effect-bounds idea (`getEffects(filler) ⊆ declared bound`) becomes a small follow-up rather than a design.

## Implementation: assembly, not analysis

Everything `describe` reports already had a canonical producer, and `_describe` only assembles them:

- signatures: `AgencyGenerator.signatureOf` (what `agency doc` renders)
- docstrings: `docStringText`; doc-comment attachment via the preprocessor's `attachDocComments()` — the exact pass `agency doc` runs, which is also what hoists the `@module` comment onto `program.docComment`
- effects: `getEffectsFromSource`, the same transitive analysis behind `getEffects`
- markers/export flags: read off the nodes

Type-alias signatures print through the canonical generator on a copy without the attached doc comment (otherwise the comment prints into the signature), with the `export` keyword stripped to match how def/node signatures omit keywords.

## Tests

- Unit (`lib/stdlib/agency.test.ts`): source order + non-export/underscore skipping, signatures/docstrings/markers, effect names (`["std::write"]` transitively), type-alias signature shape and doc comment, module description, bare-program empty case.
- Execution (`tests/agency/describe.agency`): the record crosses the stdlib boundary into Agency code — a program loops `info.value.exports` with a comprehension and reads `effects`/`destructive`/`description`. No LLM calls.

Also in this PR: explicit 30s timeouts on the walker-coverage corpus tests — the unlowered corpus parse could cross vitest's 5s default under full-suite load (seen once as a flake).

Full suite: 9,312 green; three unrelated subprocess-spawn tests timed out under load and pass in isolation (`scripts/agency.test.ts`, `lib/debugger/driver.test.ts`, `lib/cli/runPolicy.spawn.test.ts`). `lint:structure` clean, `make` + `make doc` run.

### DIFF
--- a/packages/agency-lang/lib/stdlib/agency.ts
+++ b/packages/agency-lang/lib/stdlib/agency.ts
@@ -2,15 +2,18 @@ import { compileSource, typeCheckSource, getEffectsFromSource, TypeCheckReport }
 import { writeFileSync, readFileSync, realpathSync, existsSync } from "fs";
 import { resolve, sep } from "path";
 import { parseAgency, replaceBlankLines } from "../parser.js";
-import { generateAgency } from "../backends/agencyGenerator.js";
+import { AgencyGenerator, generateAgency } from "../backends/agencyGenerator.js";
+import { TypescriptPreprocessor } from "../preprocessors/typescriptPreprocessor.js";
 import { walkNodesArray } from "../utils/node.js";
+import { docStringText } from "../utils/docStringText.js";
+import { declaredName } from "../types/hole.js";
 import { deepCopy } from "../utils.js";
 import {
   ImportKind,
   ImportPolicy,
   isImportAllowed,
 } from "../importPaths.js";
-import type { AgencyProgram, AgencyNode } from "../types.js";
+import type { AgencyMultiLineComment, AgencyProgram, AgencyNode } from "../types.js";
 import type { ImportStatement } from "../types/importStatement.js";
 import { _write } from "./builtins.js";
 import {
@@ -175,6 +178,113 @@ export function _typecheck(source: string): TypeCheckReport {
 
 export function _getEffects(source: string): Record<string, string[]> {
   return getEffectsFromSource(source);
+}
+
+// ---------------------------------------------------------------------------
+// Reify: describe a module's exports as data
+// ---------------------------------------------------------------------------
+
+export type ExportInfo = {
+  name: string;
+  kind: "def" | "node" | "type";
+  signature: string;
+  docstring: string | null;
+  effects: string[];
+  destructive: boolean;
+  idempotent: boolean;
+};
+
+export type ModuleInfo = {
+  description: string | null;
+  exports: ExportInfo[];
+};
+
+/** One ExportInfo per EXPORTED top-level def, node, and type alias, in
+ *  source order. Underscore-prefixed exports are omitted — the same rule
+ *  `agency doc` applies: they are exported for the compiler's sake
+ *  (lowering targets like `_guard`), not for callers. Effects come from
+ *  the same transitive analysis as _getEffects, sentinel included. */
+export function _describe(source: string): ModuleInfo {
+  const program = parseSource(source);
+  // Doc comments live as loose comment nodes until attachment — the same
+  // pass `agency doc` runs. It hoists the @module comment onto
+  // program.docComment and pins each doc comment to its declaration.
+  new TypescriptPreprocessor(program, {}).attachDocComments();
+  const effects = getEffectsFromSource(source);
+  const generator = new AgencyGenerator();
+  const exports: ExportInfo[] = [];
+  for (const node of program.nodes) {
+    const info = exportInfoFor(node, generator, effects);
+    if (info) exports.push(info);
+  }
+  return {
+    description: moduleDocText(program.docComment),
+    exports,
+  };
+}
+
+/** The module doc comment's text, with the doc-generator @summary marker
+ *  dropped — the marker structures `agency doc` pages, not prose. */
+function moduleDocText(comment: AgencyMultiLineComment | undefined): string | null {
+  if (!comment) return null;
+  const text = comment.content.trim().replace(/^@summary\s+/, "");
+  return text === "" ? null : text;
+}
+
+function exportInfoFor(
+  node: AgencyNode,
+  generator: AgencyGenerator,
+  effects: Record<string, string[]>,
+): ExportInfo | null {
+  if (node.type === "function" && node.exported) {
+    const name = declaredName(node.functionName);
+    if (name.startsWith("_")) return null;
+    return {
+      name,
+      kind: "def",
+      signature: generator.signatureOf(node),
+      docstring: node.docString ? docStringText(node.docString) : null,
+      effects: effects[name] ?? [],
+      destructive: node.markers?.destructive === true,
+      idempotent: node.markers?.idempotent === true,
+    };
+  }
+  if (node.type === "graphNode" && node.exported) {
+    const name = declaredName(node.nodeName);
+    if (name.startsWith("_")) return null;
+    return {
+      name,
+      kind: "node",
+      signature: generator.signatureOf(node),
+      docstring: node.docString ? docStringText(node.docString) : null,
+      effects: effects[name] ?? [],
+      // Nodes carry no tool markers; only defs can be declared
+      // destructive or idempotent.
+      destructive: false,
+      idempotent: false,
+    };
+  }
+  if (node.type === "typeAlias" && node.exported) {
+    if (node.aliasName.startsWith("_")) return null;
+    // The canonical printer emits the full declaration. The doc comment
+    // is already carried in `docstring` and the export keyword is
+    // caller-visible noise, so both are kept out of the signature — the
+    // same shape def/node signatures have (no keywords, no docs).
+    const bare = { ...node, docComment: undefined };
+    const printed = generateAgency({ type: "agencyProgram", nodes: [bare] })
+      .trim()
+      .replace(/^export\s+/, "");
+    return {
+      name: node.aliasName,
+      kind: "type",
+      signature: printed,
+      docstring: node.docComment?.content.trim() || null,
+      effects: [],
+      destructive: false,
+      idempotent: false,
+    };
+  }
+  return null;
 }
 
 // The real file path is handed to the type-checker so relative imports in
--- a/packages/agency-lang/lib/stdlib/agency.ts
+++ b/packages/agency-lang/lib/stdlib/agency.ts
@@ -5,7 +5,12 @@ import { parseAgency, replaceBlankLines } from "../parser.js";
 import { AgencyGenerator, generateAgency } from "../backends/agencyGenerator.js";
 import { TypescriptPreprocessor } from "../preprocessors/typescriptPreprocessor.js";
 import { walkNodesArray } from "../utils/node.js";
-import { docStringText } from "../utils/docStringText.js";
+import { docCommentText, docStringText } from "../utils/docStringText.js";
+import { moduleDescription } from "../utils/moduleDoc.js";
+import { patternBinders } from "../runtime/template/hygiene.js";
+import { resolveAgencyImportPath, isStdlibImport } from "../importPaths.js";
+import type { ExportFromStatement, NamedExportBody } from "../types.js";
+import { variableTypeToString } from "../backends/typescriptGenerator/typeToString.js";
 import { declaredName } from "../types/hole.js";
 import { deepCopy } from "../utils.js";
 import {
@@ -184,14 +189,18 @@ export function _getEffects(source: string): Record<string, string[]> {
 // Reify: describe a module's exports as data
 // ---------------------------------------------------------------------------
 
+export type ExportKind = "def" | "node" | "type" | "const" | "reexport";
+
 export type ExportInfo = {
   name: string;
-  kind: "def" | "node" | "type";
+  kind: ExportKind;
   signature: string;
   docstring: string | null;
   effects: string[];
   destructive: boolean;
   idempotent: boolean;
+  /** The module path this export was re-exported from, else null. */
+  reexportedFrom: string | null;
 };
 
 export type ModuleInfo = {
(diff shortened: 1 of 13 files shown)


packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -177,6 +180,113 @@ export function _getEffects(source: string): Record<string, string[]> {
...
+
+/** The module doc comment's text, with the doc-generator @summary marker
+ *  dropped — the marker structures `agency doc` pages, not prose. */
+function moduleDocText(comment: AgencyMultiLineComment | undefined): string | null {
+  if (!comment) return null;
+  const text = comment.content.trim().replace(/^@summary\s+/, "");
+  return text === "" ? null : text;
+}
+
+function exportInfoFor(
```
egonSchiele: any types you're missing here? What happens if we add a new node type and forget to add it here?

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -177,6 +180,113 @@ export function _getEffects(source: string): Record<string, string[]> {
...
+      signature: generator.signatureOf(node),
+      docstring: node.docString ? docStringText(node.docString) : null,
+      effects: effects[name] ?? [],
+      destructive: node.markers?.destructive === true,
+      idempotent: node.markers?.idempotent === true,
+    };
+  }
+  if (node.type === "graphNode" && node.exported) {
+    const name = declaredName(node.nodeName);
+    if (name.startsWith("_")) return null;
```
egonSchiele: why?

egonSchiele:
update the guide doc with your changes

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -177,6 +180,113 @@ export function _getEffects(source: string): Record<string, string[]> {
...
+      kind: "type",
+      signature: printed,
+      docstring: node.docComment?.content.trim() || null,
+      effects: [],
+      destructive: false,
+      idempotent: false,
+    };
+  }
+  return null;
+}
```
egonSchiele, line 427: This catch-all `return null` silently drops two kinds of exported surface, and one of them is common in this repo:

1. **Re-export blocks** (`exportFromStatement`). I confirmed with `pnpm run ast` that `export { map, filter } from "std::index"` parses to an `exportFromStatement` node, which no branch here handles. `stdlib/array.agency` is nothing *but* such a block — so `describe(arraySource).exports` is `[]`. Same for any module that re-exports. A caller using this to enumerate a module's surface (the stated use case) gets a confidently-wrong empty answer, not an error.

2. **Effect declarations** (`effect std::read { ... }`) — exported surface too, excluded by the `kind` union. This one is more defensible (effects aren't callable), but it's undocumented.

Options, roughly in order of how much I'd prefer them:
- Run `resolveReExports` (or equivalent) in `parseSource`'s describe path so re-exported names resolve to real `ExportInfo` entries. This makes the contract true.
- If that's too heavy for v1, at minimum emit an `ExportInfo` for each re-exported name with what's locally knowable (name + a `kind` or a `reexportedFrom` field), so the count is right even if effects/signature are thinner.
- If re-exports are genuinely out of scope, say so in the docstring ("declarations written in this module" ≠ "everything the module exports") **and** add a test pinning that an `exportFromStatement` yields nothing, so the limitation is a decision on record rather than an accident.

Either way I'd want a test over a re-export module — right now nothing in the suite exercises this path, which is why the gap is invisible.

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -177,6 +180,113 @@ export function _getEffects(source: string): Record<string, string[]> {
...
+    description: moduleDocText(program.docComment),
+    exports,
+  };
+}
+
+/** The module doc comment's text, with the doc-generator @summary marker
+ *  dropped — the marker structures `agency doc` pages, not prose. */
+function moduleDocText(comment: AgencyMultiLineComment | undefined): string | null {
+  if (!comment) return null;
+  const text = comment.content.trim().replace(/^@summary\s+/, "");
```
egonSchiele: The `@summary` handling is half-done, and the normal module-doc shape exposes it. It strips the leading `@summary ` marker but leaves the summary *text* concatenated with the body prose that follows. Real module docs have both — e.g. `stdlib/array.agency`:

```
/** @module
@summary Re-exports the array helpers ... auto-imported.
Re-exports the array helpers (`map`, `filter`, ...) so that ...
*/
```

After `@module` is stripped upstream, `content.trim()` is `@summary Re-exports ... auto-imported.\nRe-exports the array helpers ...`, and the regex removes only `@summary `, so `description` becomes the summary line glued to the body — near-duplicated text in one blob.

The tests only cover a module doc whose body *is* just the `@summary` line, so this never shows. Decide what `description` should be — just the summary, just the body, or the two kept distinct — and add a test with a two-part `@module` doc. If `agency doc` already has a canonical "module summary" extractor, reuse it here for consistency (the PR's whole thesis is reusing canonical producers).

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -177,6 +180,113 @@ export function _getEffects(source: string): Record<string, string[]> {
...
+    };
+  }
+  if (node.type === "typeAlias" && node.exported) {
+    if (node.aliasName.startsWith("_")) return null;
+    // The canonical printer emits the full declaration. The doc comment
+    // is already carried in `docstring` and the export keyword is
+    // caller-visible noise, so both are kept out of the signature — the
+    // same shape def/node signatures have (no keywords, no docs).
+    const bare = { ...node, docComment: undefined };
+    const printed = generateAgency({ type: "agencyProgram", nodes: [bare] })
```
egonSchiele: Two different signature-printing paths: defs/nodes call `generator.signatureOf(node)` on a shared `AgencyGenerator`, while types go through a fresh `generateAgency(...)` plus string surgery (`.trim().replace(/^export\s+/, "")`). The type path is the fragile one — stripping the `export` keyword with a regex on printed output will silently do the wrong thing the day a type gains another leading keyword or the printer changes spacing, and it's the only signature that isn't produced by a named method.

Cleaner: extend `signatureOf` (or add a sibling) to handle `typeAlias` and return the keyword-free declaration directly, so all three kinds share one canonical printer and there's no post-hoc munging. That also removes the `{ ...node, docComment: undefined }` shallow-copy dance.

Minor, related: reusing one `AgencyGenerator` instance across every node assumes `signatureOf` is stateless w.r.t. indent/`typeAliases`. It appears to be today, but a fresh generator per call (or a documented "signatureOf is pure" note) would make that assumption explicit rather than load-bearing-by-luck.

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -177,6 +180,113 @@ export function _getEffects(source: string): Record<string, string[]> {
...
+      .trim()
+      .replace(/^export\s+/, "");
+    return {
+      name: node.aliasName,
+      kind: "type",
+      signature: printed,
+      docstring: node.docComment?.content.trim() || null,
```
egonSchiele: Nit: type docstrings use raw `node.docComment?.content.trim()`, while defs/nodes use `docStringText(...)`. Doc comments can't interpolate so the divergence is mostly cosmetic, but it means whitespace/`@`-tag normalization differs between a type's `docstring` and a def's. If a canonical "doc comment → text" helper exists (the way `docStringText` is canonical for docstrings), route both through it so the field means the same thing regardless of `kind`.

egonSchiele (commented):
The "assembly, not analysis" framing is the right instinct — every field has a canonical producer and `_describe` just gathers them, which is why this is a small diff for a real primitive. The def/node/type extraction, marker reading, and effect wiring all check out, and the execution fixture proving the record survives the stdlib→Agency boundary is the test that matters most.

But the contract in the PR title — "reports a **module's exports** as data" — is broader than what the code reports, in one way that will bite immediately in this very repo.

**The big one: re-exports are invisible.** `export { map, filter, ... } from "std::index"` parses to an `exportFromStatement` node, and `exportInfoFor` has no branch for it, so it falls through to `return null`. `parseSource` doesn't run `resolveReExports`, so those symbols never become local decls either. That means `describe` on `std::array` — a real stdlib module whose *entire* body is a re-export block — returns `exports: []`. Several stdlib modules are shaped this way. For a primitive whose stated job is "ask what a module exports without reading its source," silently reporting nothing for a whole class of module is the one failure mode that undermines the feature. Details inline; this is the one I'd want resolved (or explicitly scoped-out and tested) before merge.

Two smaller correctness/quality items (module `@summary` handling, type-vs-def signature paths) and a couple of nits are inline.

**What I verified and found correct:**
- `signatureOf` prints only the declaration line with no keywords, so `not.toContain("export")` holds and markers stay out of the signature.
- `attachDocComments()` strips `@module` from content upstream (`parsers.ts:332`), so `moduleDocText`'s `@summary` strip lands on the right text; and it hoists the comment onto `program.docComment` exactly as `agency doc` does.
- `_describe` re-parses per call, so the in-place `attachDocComments()` mutation isn't shared across calls.
- `try _describe(...)` wraps the throwing paths (parse failure, duplicate `@module`) into the `Result`.

packages/agency-lang/lib/stdlib/agency.ts
egonSchiele: Fixed with a mechanism, not a promise: yes, there were misses — beyond re-exports, exported top-level consts (`export const version = ...`, including destructuring exports, which parse) were silently dropped. Both are handled now, and the catch-all is gone: any node with `exported: true` that the dispatch does not recognize **throws** naming the kind (surfacing through `try _describe` as a Result failure), so a future exported declaration kind fails loudly instead of vanishing from the surface. Effect declarations carry no exported flag today, so nothing legal trips it.

packages/agency-lang/lib/stdlib/agency.ts
egonSchiele: It is the `agency doc` visibility rule — underscore-prefixed exports are lowering targets (`_guard`) and compiler plumbing, exported for the compiler, not for callers (doc.ts skips them with that exact rationale). Centralized now into one `isInternalExport` helper carrying that comment, so the why lives in code instead of being repeated per branch.

packages/agency-lang/lib/stdlib/agency.ts
egonSchiele: Resolved with your first-preference option, scoped by what a bare source string permits: **std:: re-exports resolve fully** — named and star — by describing the source module (cycle-guarded via a visited set), so `describe(std::array source)` now returns the real `map`/`filter`/... entries with `kind: "def"`, signatures, and docstrings, tagged `reexportedFrom: "std::array"` (outermost hop wins on chains, same rule as template origin). **Relative-path re-exports cannot resolve from a string** (same limitation `typecheck`'s docstring records), so they come back as thin `kind: "reexport"` entries with `effects: ["unknown"]` — the getEffects never-under-report sentinel — never a silent omission. Re-export tool markers (`destructive name`) apply to the entries. One wrinkle this surfaced: the effects pipeline itself hard-fails on unresolvable re-exports, so effects run on a copy with those statements stripped (they are surface plumbing; a local call to a stripped name degrades to the sentinel). Tests: std::array star resolution, aliased named re-export, thin relative entry with sentinel + marker override.

packages/agency-lang/lib/stdlib/agency.ts
egonSchiele: Agreed — replaced the half-strip with the canonical extractor: `moduleDescription` (with `extractSummaryOverride`/`firstParagraph`/`firstSentence`/`sanitizeDescription`) moved from cli/doc.ts to `lib/utils/moduleDoc.ts`, re-exported from doc.ts for existing callers, and `describe` uses it — so `description` is the same one-line summary `agency doc` shows, and the two-part @module shape can no longer glue summary to body. New test pins exactly that: summary line plus body prose yields the summary only.

packages/agency-lang/lib/stdlib/agency.ts
egonSchiele: Done — `signatureOf` grew a `typeAlias` branch (including the effectSet form), producing the keyword-free declaration line directly, so all three kinds share the one canonical printer and the `generateAgency` + regex-strip + shallow-copy dance is gone. On "no keywords": the contract is now stated precisely as "no export keyword or tool markers" — the `type` keyword stays, same as `type Name = ...` reads in source. On generator reuse: kept the shared instance; `signatureOf` now documents the alias-registration side effect it intentionally shares with `processTypeAlias`.

packages/agency-lang/lib/stdlib/agency.ts
egonSchiele: Routed through a new canonical helper: `docCommentText` in `lib/utils/docStringText.ts` (the doc-comment sibling of `docStringText`), so the `docstring` field normalizes the same way regardless of kind.

egonSchiele:
All review findings addressed in 400dba4fd — details in the threaded replies. Summary: std:: re-exports (named + star) resolve to full entries with `reexportedFrom`; relative ones come back thin with the `unknown` effects sentinel; exported consts (a third silent miss the completeness question flushed out) are surface; unhandled exported kinds now throw instead of vanishing; module description reuses the canonical `moduleDescription` (moved to `lib/utils/moduleDoc.ts`); type signatures go through `signatureOf` proper; doc comments normalize via `docCommentText`. Guide section added to `templates.md` (describe + the handler-per-effect loop). On Copilot's `string | null` → `?` rendering in the generated stdlib docs: that is the doc generator's pre-existing treatment of nullable fields everywhere (e.g. `HoleInfo.origin`), not something this PR introduces — happy to file it separately. Full suite 9,320 green, lint clean, `make` + `make doc` run.

PR #672, opened 2026-07-24
Lint rules: missing-docstring (AL0002) and redundant-prelude-import (AL0003)

egonSchiele:
PR A of three from the lint-rules-v2 plan (spec + reviews in the worktree-lint-list worktree). Adds the two AST-only rules; PR B (severity plumbing + AL0004) and PR C (scope rules) follow after this merges.

## AL0002 — missing-docstring

An exported function with no docstring. Functions are tools: the docstring becomes the tool description the LLM reads, so an undocumented export hands every importing agent a tool with no description. Functions only — a compile trace confirmed node docstrings never reach generated code (their only consumers are `agency doc` and an interpolation-presence check), and the stale preprocessor comment implying otherwise is reworded in this PR. No autofix: the fix is prose only the author can write.

**Noise measurement (per the plan):** 56 findings across the repo — 26 in `stdlib/`, 30 in test fixtures. The 26 stdlib docstrings are **fixed in this PR** (second commit): most were existing `///` or `/** */` doc comments converted to docstrings so they actually reach the LLM, which is exactly the near-miss the rule exists to catch. Fixture findings are left alone (not user-facing). Stdlib reference docs regenerated via `make`.

## AL0003 — redundant-prelude-import

`import { map } from "std::index"` imports a name every file already has. The rule's membership test is `PRELUDE_NAMES.includes(name)` — NOT the module path — because `std::index` also exports names outside the prelude (types like `WriteMode`, which must be imported and are never flagged). Also never flagged: aliased names (`map as arrMap` binds a new name) and names carrying `destructive`/`idempotent` markers (the marker changes retry behavior; deleting it would be a safety regression). The removal fix reuses the AL0001 regeneration machinery, safe precisely because every flagged name is plain.

## LSP correctness fix underneath both

The LSP's lint snapshot was not actually pristine: `prunePreludeShadows` mutates user-written `std::index` import statements in place before the linter runs (the old alias protected the nodes array, not the objects). AL0001 never noticed because it skips those statements; AL0003 reads exactly them. `cloneForLint` now deep-copies just the `std::index` import nodes into the snapshot, pinned by a re-run-stability test and a CLI/LSP agreement test on the shadowed-prelude case.

## Refactor (from the plan's anti-pattern review)

- `walkValues` in `lib/linter/rules/util.ts` is now the linter's ONE recursive descent; `collectReferencedNames` is a callback over it (and re-signed to take `AgencyNode[]`, its only caller updated).
- Range helpers (`nameRange`, `statementSpan`, `locFromOffsets`) and the import-fix machinery (`removalEdit` et al.) extracted to shared modules for the new rules.
- Rule bodies are declarative helper→map with named predicates (`undocumentedExports`, `isRedundant`), matching AL0001's shape.

Main drifted under the plan in one spot: template holes (#665) make `importedNames` and `functionName` `string | Hole`, so both new rules filter to string entries (a hole is a template fragment, not a lintable name).

## Testing

205 tests green across linter/LSP/CLI/docs (18 new rule tests including the at-risk keeps: `WriteMode`, aliased, destructive/idempotent-marked, two-redundant-names fixes; per-rule `lintSource` smoke tests; snapshot stability + CLI/LSP agreement). Structural linter clean. Diagnostics docs regenerated — AL0002/AL0003 now on `docs/site/diagnostics/lint.md`, plus the rules section of `docs/site/cli/lint.md`.

### DIFF
--- a/packages/agency-lang/lib/linter/rules/redundantPreludeImport.ts
+++ b/packages/agency-lang/lib/linter/rules/redundantPreludeImport.ts
@@ -47,6 +47,12 @@ export const redundantPreludeImportRule: LintRule = {
     return redundantNamesIn(ctx).map(({ stmt, name }) => {
       const span = statementSpan(ctx.source, stmt);
       const range = nameRange(ctx.source, span.start, span.end, name);
+      // Each fix regenerates the WHOLE statement minus one name, so two
+      // findings on the same statement carry overlapping edits. Safe for
+      // one-at-a-time quick fixes (the editor re-lints between them), but
+      // a future batch/fix-all action must group per statement and
+      // regenerate once, the way unusedImportsBatchEdits does — never
+      // concatenate these per-name fixes.
       const fix: LintFix = {
         title: `Remove redundant import '${name}'`,
         edits: [removalEdit(ctx.source, stmt, [name])],
--- a/packages/agency-lang/lib/linter/rules/redundantPreludeImport.ts
+++ b/packages/agency-lang/lib/linter/rules/redundantPreludeImport.ts
@@ -1,0 +1,57 @@
+import type { ImportStatement, NamedImport } from "../../types/importStatement.js";
+import { PRELUDE_NAMES } from "../../prelude.js";
+import type { LintContext, LintFinding, LintFix, LintRule } from "../types.js";
+import { lintDiagnostic } from "../diagnostics.js";
+import { nameRange, statementSpan } from "./util.js";
+import { removalEdit } from "./importFixes.js";
+
+/** A name is redundant only when it is completely plain: provided by the
+ *  prelude, not rebound under an alias, and carrying no retry-safety
+ *  marker. The membership test is PRELUDE_NAMES, NOT the module path:
+ *  std::index also exports names outside the prelude (types like
+ *  WriteMode — the mirror test covers functions only), and importing
+ *  those is the only way to get them. Plainness is what makes the removal
+ *  fix safe. */
+function isRedundant(nameType: NamedImport, name: string): boolean {
+  return (
+    PRELUDE_NAMES.includes(name) &&
+    !Object.hasOwn(nameType.aliases, name) &&
+    !nameType.destructiveNames?.includes(name) &&
+    !nameType.idempotentNames?.includes(name)
+  );
+}
+
+function redundantNamesIn(ctx: LintContext): { stmt: ImportStatement; name: string }[] {
+  return ctx.program.nodes
+    .filter(
+      (node): node is ImportStatement =>
+        node.type === "importStatement" &&
+        node.modulePath === "std::index" &&
+        !node.testOnly,
+    )
+    .flatMap((stmt) =>
+      stmt.importedNames
+        .filter((nameType): nameType is NamedImport => nameType.type === "namedImport")
+        .flatMap((nameType) =>
+          nameType.importedNames
+            .filter((name): name is string => typeof name === "string") // template holes are not names
+            .filter((name) => isRedundant(nameType, name))
+            .map((name) => ({ stmt, name })),
+        ),
+    );
+}
+
+export const redundantPreludeImportRule: LintRule = {
+  name: "redundantPreludeImport",
+  run(ctx: LintContext): LintFinding[] {
+    return redundantNamesIn(ctx).map(({ stmt, name }) => {
+      const span = statementSpan(ctx.source, stmt);
+      const range = nameRange(ctx.source, span.start, span.end, name);
+      const fix: LintFix = {
+        title: `Remove redundant import '${name}'`,
+        edits: [removalEdit(ctx.source, stmt, [name])],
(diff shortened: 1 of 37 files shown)


packages/agency-lang/lib/linter/rules/redundantPreludeImport.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const redundantPreludeImportRule: LintRule = {
+  name: "redundantPreludeImport",
+  run(ctx: LintContext): LintFinding[] {
+    return redundantNamesIn(ctx).map(({ stmt, name }) => {
+      const span = statementSpan(ctx.source, stmt);
+      const range = nameRange(ctx.source, span.start, span.end, name);
+      const fix: LintFix = {
+        title: `Remove redundant import '${name}'`,
+        edits: [removalEdit(ctx.source, stmt, [name])],
```
egonSchiele, line 59: Non-blocking, forward-looking: when one statement has two redundant names (`import { map, filter }`), each finding carries its own fix that regenerates the **whole statement**. Applying both together produces overlapping edits over the same span — the exact overlap AL0001 avoids by grouping per statement in `unusedImportsBatchEdits`. It is safe today: individual quick-fixes apply one at a time (each re-lints), and the LSP batch/remove-on-save path (`lintBatchEdits`) is AL0001-only, so AL0003 fixes never batch. But the moment a "fix all AL0003" or a combined cleanup action lands, AL0003 needs the same group-by-statement regeneration AL0001 already has. Worth a one-line comment here so the next person adding a batch action doesn't wire these per-name fixes into it.

packages/agency-lang/lib/lsp/diagnostics.test.ts
```diff
@@ -377,3 +378,38 @@ describe("lint diagnostics", () => {
     expect(diagnostics.filter((d) => d.code === "AL0001")).toEqual([]);
   });
 });
+
+describe("lint snapshot is isolated from prunePreludeShadows", () => {
+  it("produces identical lint findings on a second run of the same document", () => {
```
egonSchiele: This re-run-stability test does not actually guard `cloneForLint`: `runDiagnostics` re-parses fresh each call, so run 1's in-place mutation never reaches run 2. Without the clone both runs return `[]` (prune strips `map` each time before the linter sees it); with the clone both return `[AL0003]`. Either way the two runs match, so the test passes with or without the fix. The **CLI/LSP agreement test below is the real guard** — it fails without the clone. This one is harmless but gives false confidence; consider dropping it, or replacing the assertion with a direct check that the user's `std::index` import still contains `map` after `runDiagnostics` (which *would* fail without the clone). The plan flagged this ("this test may PASS already"), so it's a known choice — just flagging that it isn't pinning what its name implies.

egonSchiele (commented):
Reviewed the full diff against the plan and my earlier spec/plan/test reviews. This is a clean, high-quality PR A — it addressed essentially everything from the prior rounds:

- **`walkValues`** is now the linter's one bespoke descent, with `collectReferencedNames` re-signed to `AgencyNode[]` and expressed as a callback — the three-copies anti-pattern is gone.
- **Rule bodies are declarative** `helper→map` with named predicates (`undocumentedExports`, `isRedundant`), matching AL0001's shape.
- **AL0003 test sources no longer use unparseable lambda syntax** — they use `node main() { return 1 }` and note redundancy is usage-independent. The two-redundant-names case and per-rule `lintSource` smoke tests I asked for are both here.
- **AL0002's comment test is relabeled** with the accurate reason (`docComment` is populated only by the TypescriptPreprocessor, which the linter never runs), and the preprocessor comment rewording is correctly traced.
- **`cloneForLint` + the CLI/LSP agreement test** is the right fix and the agreement test is non-vacuous (the `toContain("AL0003")` guard).
- **The 26 stdlib docstrings** are terse, actionable, correctly placed as first body statements, and free of accidental `${}` interpolation. Converting `///`/`/** */` doc comments into docstrings is exactly the near-miss the rule exists to catch.

Two minor, non-blocking notes inline — both forward-looking, neither needs to hold the merge. Nice work.

packages/agency-lang/lib/linter/rules/redundantPreludeImport.ts
egonSchiele: Comment added in 64d1c931a, stating exactly that: per-name fixes are one-at-a-time only, and any future batch action must group per statement and regenerate once like unusedImportsBatchEdits.

packages/agency-lang/lib/lsp/diagnostics.test.ts
egonSchiele: You are right — each runDiagnostics call re-parses, so the equality assertion passed with or without the clone. Replaced in 64d1c931a with the direct guard you suggested: the test now asserts the AL0003 finding itself, which only exists if the snapshot kept `map` past the prune. Verified it fails with cloneForLint reverted.

PR #673, opened 2026-07-24
Code literals: inline [| ... |] templates

egonSchiele:
Implements the code-literal spec (`docs/superpowers/specs/2026-07-23-code-literals-design.md`, review-revised) per its plan (`docs/superpowers/plans/2026-07-24-code-literals.md`): inline `[| ... |]` templates that parse at compile time and evaluate to ordinary `Code` values.

```
const guardTpl = [|
  def guarded(): string {
    const ms: number = #minutes
    #body
    return "done"
  }
|]
const partial = fill(guardTpl, { body: [| print("step") |] })
```

## The formatter gate ran first — and found four pre-existing generator bugs

The spec's riskiest promise is that `fmt` reformats literal bodies, which rests on the canonical generator being clean on unlowered template-mode nodes. The plan sequenced a whole-corpus round-trip/idempotence gate before any feature work, and its first run found four real bugs, each fixed with a pinned repro:

- **Semantic corruption:** a parenthesized compound base printed bare — `(a ?? "").toUpperCase()` became `a ?? "".toUpperCase()`, re-associating the chain. `fmt` was corrupting `stdlib/data/finance/edgar.agency`.
- A zero-parameter signature with a long return+raises clause wrapped to an unparseable `(\n\n)` (`stdlib/notes/apple.agency` printed unparseable output).
- Node imports rendered twice (collected in two passes), duplicates compounding on every `fmt` pass.
- Import hoisting left orphaned blank lines that shifted on every pass instead of converging.

The gate's comparison contract is stated in the file: code nodes compare in order; imports and comments compare as content-preserving multisets (sorting/hoisting are policy); vertical-whitespace stability is the idempotence invariant's job.

## The feature, by layer

- **fill relaxation** (own commit): a `statements` hole accepts an `expr` fragment (an expression IS a legal statement; the graft is the identity). Deliberately inverts the old rejection test; guards pin the preserved rows. This closes the only gap smallest-first kind inference leaves — `statements` already accepted `program`.
- **Parser**: body parses at parse time of the enclosing file (unlowered template mode, holes intact), kind inferred smallest-first (`expr` → `statements` → `program`). The end-scan finds `|]` in code position by **reusing the real string/comment parsers** — zero escaping rules; `|]` inside strings, comments, and interpolations is inert (all pinned with content assertions, not bare `.success`). Nested `[|` is a directive error. Empty body = empty statements fragment, matching `parseStatements("")`.
- **Two substrate discoveries** the plan's "fiddliest area" flag paid for: tarsec keeps three module-global states (position input string, rightmost-failure record, memo caches) that nested parses clobber — all saved/restored, or the enclosing parse's error formatter crashes on negative offsets. And tarsec's `label()` saves/restores the rightmost record around every labeled region, scrubbing any deep failure recorded inside a literal — so literal errors travel a dedicated channel that `_parseAgency` (and the `TarsecError` catch) prefer when at least as deep as the generic failure. Body-error lines map into enclosing-file coordinates, additive under the prelude offset (hand-computed expectations, tested in both modes).
- **Host-side leaf-ness, four levers**: no `walkNodes` case (leaf for free — `walkNodes` untouched, per the plan's global constraint); `codeLiteral.nodes` deliberately NOT in `bodySlots` (comment at the spot where someone would helpfully add it, plus a pinning test); `NO_EXPRESSION_SLOTS` registration (forced by the completeness test); `WALKER_EXCLUDED_FIELDS["codeLiteral.nodes"]` tripwire ruling, live now that the corpus contains literals. A fifth forcing function surfaced on its own: `identifierSlots.ts` is exhaustively typed over `AgencyNode` and demanded its entry at compile time.
- **Formatter**: `[| ... |]` prints with the body one indent level deeper (inline for one-line expr bodies) via `printCodeLiteralBody` — the same text codegen embeds. `fmt` reformats bodies deliberately. Byte-exact goldens pin the text (mis-indented body canonicalized, inline form, comment survival).
- **Codegen/runtime**: the builder emits `__codeLiteral(printedBody, kind)`; the runtime reconstructs through the SAME `parseCodeLiteralBody` and asserts the recorded kind — compile and runtime cannot diverge by drift. Reuses Code-serializes-as-source; no second representation.
- **Typechecker**: the literal synthesizes the structural equivalent of stdlib `Code`, built through the type-alias production (field encodings match by construction) with the exact `"agencyProgram"` literal and kind union. One guard in `scopes.ts`: a variable bound to a literal keeps the unwidened type — `widenType` deep-widens record fields, which would have broken the flagship `const tpl = [| ... |]; fill(tpl, ...)` flow (found by the plan-mandated compatibility test). Quoted bodies produce no host diagnostics, tested non-vacuously with a sanity anchor.

## End-to-end proof (execution fixtures, no LLM calls)

- `literalCompose`: three literals, partial fill, composition with a hole still open, `origin: "helpers"` on the grafted hole, generated program runs under a `std::run` handler → `"lit-ok"`.
- `literalAcrossCheckpoint`: a literal-built `Code` value survives an interrupt checkpoint and fills after resume.
- `literalInterpolationPassthrough`: the no-splice guarantee — `${x}` in a body runs in the GENERATED program; `holesOf` sees no hole for it. (The fixture itself demonstrates the host/quoted boundary: its own guard string needed `\${x}`, the literal body did not.)
- `literalBadBody`: `expectedCompileError` through the real compile path, with the mapped body line.
- Equivalence pinned as an equality: `toSource` of a literal-built value === the file-template print (one printer — `_toSource` IS `generateAgency`).

## Notes for review

- Spec open question 1 resolved structurally (no alias-naming mechanism needed; the `fill()` compatibility test is the proof); open question 2 decided as empty-statements-fragment with the consistency rationale in the test.
- The `literalBadBody` fixture is intentionally unparseable and is excluded from the formatter gate by name, with its reason.
- One tiny audit commit ("block bodies for the loc-shift one-line ifs") is missing the Co-Authored-By footer — committed inline by mistake, and amending is off the table per house rules.
- Full suite: 9,420 green; lint clean; LSP suite green (literals are opaque expressions to it, as specced); `make` + `make fixtures` run (95 `.mjs` files gained one import line). CI runs the agency suite.


packages/agency-lang/lib/typeChecker/synthesizer.ts
```diff
@@ -4,6 +4,36 @@ import { AgencyNode, Expression, VariableType, ValueAccess, formatUnitLiteral }
...
+ *  string-literal types) match by construction; the bare
+ *  variableTypeParser encodes optionality differently and the record
+ *  comparison rejects the mismatch. The `type` and `kind` fields are the
+ *  EXACT literal and union from Code: structural assignability into a
+ *  literal-typed field is directional, so a widened `string` here would
+ *  fail fill(template: Code). Structural typing is what lets the literal
+ *  flow into std::agency's fill/holesOf/toSource without any mechanism
+ *  for naming a stdlib alias from the checker. */
+function parseCodeLiteralType(): VariableType {
+  const source = [
```
egonSchiele, line 34: I don't understand why we write this into a string and then parse it

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -1260,12 +1306,45 @@ export class AgencyGenerator {
...
+    // A compound base only exists because the source parenthesized it
+    // (`(a ?? b).toUpperCase()` parses via parenAccess). Printing it bare
+    // re-associates the chain onto the base's RIGHT OPERAND — semantic
+    // corruption, caught by the round-trip gate on edgar.agency. Any base
+    // that binds looser than member access gets its parens back.
+    if (
+      node.base.type === "binOpExpression" ||
+      node.base.type === "typeTestExpression" ||
+      node.base.type === "isExpression"
+    ) {
```
egonSchiele: **Confirmed semantic corruption for a base type not in this denylist: `tryExpression`.**

On the built branch:

```
def f(x: any): any {
  return (try x).foo
}
```
formats to `return try x.foo`, i.e. `try (x.foo)` — the parens are dropped and the chain re-associates. Same class as the edgar fix this block was written for; it just isn't in the corpus so the gate didn't catch it. (`(-x).foo` doesn't parse, so that one's safe; `!`/`typeof` happen to round-trip.)

A denylist here is fragile — it's one new looser-than-member-access expression kind away from silently corrupting again (pipe/`catch`/ternary are worth auditing too). Consider inverting to a **safe-base allowlist**: parenthesize unless `node.base.type` is one of the member-access-safe atoms (`variableName`, `functionCall`, `valueAccess`, literals, an already-parenthesized form). That fails closed instead of open.

packages/agency-lang/lib/typeChecker/scopes.ts
```diff
@@ -194,7 +194,13 @@ export function declareVariable(
     );
   }
   const inferred = synthType(node.value, scope, ctx);
-  scope.declare(node.variableName, widenType(inferred), isConst);
+  // A code literal's synthesized type IS the fixed structural Code shape;
+  // widening its literal discriminants (type: "agencyProgram" → string)
+  // would break assignability into fill(template: Code). Widening exists
+  // to relax literal-EXPRESSION types for rebinding; a variable holding a
+  // literal always holds a Code value, so there is nothing to relax.
+  const declared = node.value.type === "codeLiteral" ? inferred : widenType(inferred);
```
egonSchiele: This guard is the right idea, but it keys on the **AST shape** of the RHS (`node.value.type === "codeLiteral"`) rather than on the type being the structural `Code` shape. I couldn't break it — `const a = otherLiteralVar` and `const c = partial.value` both widen fine (the latter carries the nominal `Code` alias). But a literal that reaches a `const`/`let` through a ternary or array element (`const ts = [ [| a |], [| b |] ]`, then `ts[0]`) wouldn't match this check and would get `widenType`'d, potentially breaking a later `fill(...)`.

Two options: (a) add a defensive test for the ternary/array-of-literals path and confirm it's fine, or (b) make the guard type-directed — skip widening when `inferred` already *is* the `Code` structural shape (`type: "agencyProgram"`), which covers every way a literal-typed value can arrive, not just the direct-assignment syntax.

egonSchiele (commented):
## Review: Code literals `[| ... |]`

I read this at the level of the parser substrate, the formatter fixes, the `fill` relaxation, the typechecker widening guard, and the leaf-ness levers, and ran a few probes against the built branch. This is high-quality work — notably, the plan's flagged risks were actually paid down rather than papered over. One confirmed pre-existing formatter bug in the exact class this PR is hardening, and a handful of minor notes below.

### What's genuinely well done

- **The formatter gate earned its keep.** Running a whole-corpus round-trip/idempotence sweep *before* the feature flushed out four real pre-existing generator bugs, including a semantic-corruption one (`(a ?? "").toUpperCase()` → `a ?? "".toUpperCase()`). Each fix has a minimal pinned repro alongside the broad corpus net. That's the right shape.
- **The end-scan reuses the real string/comment parsers** (`_stringParser`, `commentParser`, `multiLineCommentParser` via `consumeWith`) instead of a hand-rolled lexer, so the scan and the grammar agree by construction. I checked the memo concern this raises: tarsec's memo is keyed by the input *string*, and the scan runs while the global input is still the enclosing input, so cached `loc`s stay valid — no pollution. Good.
- **Location mapping is tested with hand-computed exact lines** (`Line 3, col …`) in both the no-offset and additive-prelude-offset modes, plus a body-node loc assertion. The `strippedPrefix` accounting for trimmed leading whitespace is exactly the subtlety that's easy to miss.
- **End-scan inertness tests assert body *content***, not bare `.success` — including `|]` inside a nested interpolation string *and* `|]` in interpolation code position pinned as a decision.
- **The synthesized `Code` type uses the exact `"agencyProgram"` literal and kind union**, and the `scopes.ts` widening guard keeps a literal-bound variable's type unwidened so the flagship `const tpl = [| … |]; fill(tpl, …)` flow typechecks. The `__codeLiteral` runtime constructor shares `parseCodeLiteralBody` with compile and asserts the recorded kind, so compile/runtime can't drift.
- All five host-leaf-ness forcing functions are present (`NO_EXPRESSION_SLOTS`, `bodySlots` non-registration + comment, `WALKER_EXCLUDED_FIELDS`, `identifierSlots`, and the completeness test), and the leaf-ness execution test is non-vacuous.

### One confirmed bug (pre-existing, but in this PR's blast radius)

The paren re-parenthesization fix in `agencyGenerator.ts` uses a **denylist** of base types (`binOpExpression`, `typeTestExpression`, `isExpression`). It misses `tryExpression`. On the built branch:

```
(try x).foo   →   try x.foo        // re-associates to try (x.foo) — semantic corruption
```

Same class as the edgar fix; it slips the gate only because no corpus file has this shape. Details inline. Recommend flipping to a *safe-base allowlist* rather than enumerating every looser-binding base — a denylist here is one new expression kind away from silently corrupting again.

### Minor notes / questions (all inline)

- `scopes.ts` widening guard is AST-shape-based (`node.value.type === "codeLiteral"`). I couldn't break it (aliasing and `.value` both widen fine), but a literal reaching a `const` through a ternary/array might escape the syntactic check — worth a defensive test or a type-directed guard.
- `scanCodeLiteralBody` runs the real parsers *before* the rightmost-failure save/restore that `parseCodeLiteralBody` sets up — a scan-internal `recordFailure` can leak into the enclosing parse's error record. Low severity, but inconsistent with the care taken a few lines down.
- `__codeLiteral` re-parses the body (and `resetMemos()`) on every evaluation — a literal in a hot loop re-parses each iteration. Probably fine, matching `parseExpr`, but worth a comment.
- Kind inference: the `expr` rest-check uses `.rest.trim()` while the `statements` check uses `stripSentinels(...).trim()`, so a lone expression followed by a blank line infers `statements`, not `expr`. Harmless (the relaxation makes both fill the same holes) but asymmetric.

Nothing here is merge-blocking except your call on the `try`-paren corruption. Nice work.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2934,6 +2950,301 @@ export const schemaAccessParser: Parser<ValueAccess> = memo(
...
+      return {
+        ok: false,
+        error:
+          "nested code literals are not supported; build the inner piece as its own value and graft it into a hole",
+        at: i,
+      };
+    }
+    const ch = input[i];
+    if (ch === '"' || ch === "'" || ch === "`") {
+      const skipped = consumeWith(_stringParser, input, i);
```
egonSchiele: Nice — reusing the real parsers here (rather than a hand-rolled lexer) makes the scan and grammar agree by construction, and the memo caches are string-keyed so there's no pollution from running them during the enclosing parse.

One asymmetry though: this scan runs `_stringParser` / `commentParser` **before** the `saveRightmostFailure()` / `restoreRightmostFailure()` guard that `parseCodeLiteralBody` sets up a few lines down. A string/comment parse that internally tries and fails an alternative can `recordFailure` onto the *enclosing* parse's rightmost record. Low severity (it only affects which error surfaces if the enclosing parse later fails), but given the deliberate care taken with rightmost-failure state in `parseCodeLiteralBody`, wrapping the scan in the same save/restore would make the treatment consistent.

egonSchiele:
Refactored onto tarsec 0.5.1 (commit c9f1399b3) — the imperative literal plumbing is gone, replaced by the four new combinator features. Net **−147 lines** across the two parser files (−292/+151 with the lockfile), and the comments shrank more than the code, since most existed to explain global-state hazards that no longer exist.

What each tarsec feature deleted:
- `runNested` + `basePosition` → the four-way save/restore block, `shiftLiteralLocs`, the strippedPrefix loc bookkeeping, and the `Line (\d+)` rewrite regex. Body spans AND error messages come out in enclosing-file coordinates natively; the only retained save/restore is agency's own template offset (our coordinate layer, not tarsec's).
- `committed` → the `pendingCodeLiteralError` side channel and the `_parseAgency` preference branch. One integration note: tarsec's `parseError` **throws** its generic message even when a committed failure exists — but the committed failure survives in the parse state and `getErrorMessage()` prefers it, so the `TarsecError` catch now consults it when present. (Possible future tarsec nicety: `parseError` could prefer a committed failure itself; the catch-side clause works fine meanwhile.)
- `repeatTill`/`matchedText`/`many1TillOneOf` → `scanCodeLiteralBody` and `consumeWith`; the end-scan is now five declarative alternatives, with the bulk-text alternative per repeatTill's own perf guidance.
- The `label` deeper-record fix → two `parser.test.ts` column assertions updated (`!!!` now reports the deepest real failure — the unary-`!` chain dying at EOF — instead of the token start), which is the change working as intended. No other message assertions in the suite were affected.

All green after: 9,420 unit tests, the four literal execution fixtures, formatter gate, tripwire, lint.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -1260,12 +1306,45 @@ export class AgencyGenerator {
...
     let code = this.processNode(node.base).trim();
+    // A compound base only exists because the source parenthesized it
+    // (`(a ?? b).toUpperCase()` parses via parenAccess). Printing it bare
+    // re-associates the chain onto the base's RIGHT OPERAND — semantic
+    // corruption, caught by the round-trip gate on edgar.agency. Any base
+    // that binds looser than member access gets its parens back.
+    if (
+      node.base.type === "binOpExpression" ||
+      node.base.type === "typeTestExpression" ||
+      node.base.type === "isExpression"
```
egonSchiele: The allow-list is incomplete, and one of the omissions is the case the `parenAccessParser` docstring itself cites.

```
const x = (new Foo()).bump()
```

`node.base.type` is `"newExpression"`, so no parens are restored and this prints as `new Foo().bump()`. `newExpressionParser` stops after the closing `)` and consumes no chain element, so the printed text does not re-parse at all — `fmt` emits an unparseable file. That is strictly worse than the edgar bug this block fixes (corruption vs. failure, but still a corpus-invisible one).

`tryExpression` is the same shape: `(try f()).value` → `try f().value`, which re-parses as `try` over `f().value` — a different program, since `tryExpressionParser`'s call capture accepts a `valueAccess`.

The other baseAtom entries are safe by accident: `!`, `typeof`, `void`, and postfix `++`/`--` all desugar to `binOpExpression`, so they're covered. That's worth not relying on. Given `parenAccessParser`'s base is any `exprParser` result, I'd invert this to a deny-list — parenthesize unless the base is a self-delimiting primary (literal, identifier, valueAccess, functionCall, array/object/comprehension, schema/new-with-own-parens) — so a future atom kind fails closed with a redundant pair of parens rather than open with corrupted output.

Either way, both repros belong in the pinned-repro block below alongside the edgar one; the corpus gate demonstrably doesn't reach them.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -253,6 +254,15 @@ export class AgencyGenerator {
     for (const node of program.nodes) {
       const result = this.processNode(node);
       if (result !== "" || node.type === "newLine") {
+        // Collapse runs of blank lines to one. Import hoisting can leave
+        // two newLine nodes adjacent (the hoisted import sat between
+        // them), and re-parsing sees a different blank count than the
+        // one printed — each fmt pass would then shift blank lines
+        // instead of converging. One blank is the canonical form.
+        const prev = stmtPairs[stmtPairs.length - 1];
+        if (node.type === "newLine" && prev !== undefined && prev.type === "newLine") {
```
egonSchiele, line 263: This is scoped to the whole top-level node stream, not to the import region the comment describes. `replaceBlankLines` emits one sentinel per blank line and `blankLineParser` turns each into its own `newLine` node, so any run of two or more blank lines anywhere in a file — between two `def`s, inside the body of nothing in particular — now collapses to one.

That may well be the formatting policy you want (most formatters do exactly this), but it's a user-visible `fmt` change for every existing file, and it isn't in the PR description or `docs/site`. The round-trip gate can't flag it either: `normalized()` skips `newLine` nodes by design.

Two options: keep it and state it as policy ("runs of blank lines collapse to one") with a test that pins it directly, or scope the guard to the import region so the convergence fix stays a convergence fix.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2934,6 +2954,164 @@ export const schemaAccessParser: Parser<ValueAccess> = memo(
...
+  "nested code literals are not supported; build the inner piece as its own value and graft it into a hole";
+
+/** One inert chunk of a literal body. Strings (escapes and `${...}`
+ *  interpolations, nested strings included) and comments are consumed by
+ *  the SAME parsers the grammar uses, so a `|]` inside them cannot end
+ *  the literal and no escaping rules exist. A nested `[|` is a committed
+ *  error. The bulk alternative swallows runs of text containing no
+ *  trigger characters in one step (see repeatTill's docs). */
+const codeLiteralBodyChunk: Parser<unknown> = or(
+  many1TillOneOf(['"', "'", "`", "/", "[", "|"]),
```
egonSchiele: Regex literals are missing from the inert set, so the "no escaping rules exist" guarantee doesn't hold for them.

Agency's regex is `re/.../` (`regexLiteralParser`, a `baseAtom` entry), and neither the trigger set nor the chunk alternatives account for it:

- `[| const p = re/[|]/ |]` — the bulk scan stops at `[`, the nested-`[|` probe matches, and the user gets "nested code literals are not supported; build the inner piece as its own value" for a regex matching a pipe character.
- `[| const p = re/a|]b/ |]` — `repeatTill` tries the terminator first each round, matches the `|]` inside the pattern, and the literal ends mid-regex. Downstream that's a parse error somewhere confusing, or silently the wrong body.

The fix follows the existing pattern: add `lazy(() => regexLiteralParser)` to the alternatives, and `"r"` to the stop set so the bulk scan doesn't swallow the `re` before the regex parser gets a look. Adding `r` costs a stop at every `r` in the body; that's the same trade already taken for `/`, and `many1TillOneOf` recovers immediately via `anyChar`.

Worth a test in the end-scan block next to the string and comment ones, and a note in `docs/site/guide/templates.md` — "Nothing needs escaping" is a promise this makes literally true rather than nearly true.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2934,6 +2954,164 @@ export const schemaAccessParser: Parser<ValueAccess> = memo(
...
+  const sentineled = replaceBlankLines(body);
+  const trimmedStart = sentineled.trimStart();
+  const strippedPrefix = sentineled.slice(0, sentineled.length - trimmedStart.length);
+  const trimmed = trimmedStart.trimEnd();
+  const trimmedBase = base === undefined ? undefined : advancePositionOver(base, strippedPrefix);
+  const savedOffset = currentTemplateOffset;
+  setTemplateOffset(0);
+  try {
+    const asExpr = runNested(exprParser, trimmed, { basePosition: trimmedBase });
+    if (asExpr.success && asExpr.rest.trim() === "") {
```
egonSchiele: The expr branch checks `asExpr.rest.trim()` while the statements branch below checks `stripSentinels(asStatements.rest).trim()`. The asymmetry looks unintentional, and combined with the `trimStart()`/`trimEnd()` above it makes a blank line change the inferred kind.

`BLANK_LINE_SENTINEL` is U+E000, which is not whitespace, so `trimStart()` doesn't remove it:

```
const a = [| 1 + 2 |]      // kind: expr
const b = [|

  1 + 2
|]                          // kind: statements
```

In `b`, `replaceBlankLines` leaves a leading sentinel that `trimStart` can't strip, `exprParser` dies on it, and `bodyParser` picks it up as a blank line plus an expression statement. Since `statements` fragments don't fill `expr` holes (only the reverse, via the new relaxation), a stray blank line silently turns a literal into one that can't fill the hole it was written for — with no diagnostic pointing at the blank line.

Stripping sentinels alongside whitespace when computing `trimmed`/`strippedPrefix`, and using `stripSentinels` in the expr check for symmetry, should close it. `strippedPrefix` still needs to carry the original character count so `advancePositionOver` keeps the line math right.

packages/agency-lang/lib/parser.ts
```diff
@@ -327,6 +325,24 @@ export function parseAgency(
...
+      // is the targeted one and stays authoritative.
+      if (getParseState().committedFailure !== null) {
+        const committedMessage = getErrorMessage();
+        if (committedMessage !== null) {
+          const rightmostCommitted = getRightmostFailure();
+          return {
+            success: false,
+            message: committedMessage,
+            rest: input,
+            ...(rightmostCommitted ? { rightmostPos: rightmostCommitted.pos } : {}),
```
egonSchiele: `message` comes from the committed failure but `rightmostPos` comes from the rightmost record — two different positions in the same result. Given the whole point of the committed-failure channel is that the rightmost record is the misleading one here, the LSP squiggle lands away from the message it accompanies.

`getParseState().committedFailure` is in scope and its `rest` is a suffix of the current source, so `inputStr.length - committedFailure.rest.length` gives the position the message was formatted from (the same arithmetic `getErrorMessage` does). Worth a test that a malformed literal body produces a `rightmostPos` inside the literal.

packages/agency-lang/lib/runtime/template/literal.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/** Reconstructs a code literal's value at runtime from its canonical
+ *  printed body, through the SAME per-kind parse (parseCodeLiteralBody)
+ *  the compiler used — the program grammar alone would reject an
+ *  expr-kind body, and sharing the entry point is what makes
+ *  compile-time and runtime agree by construction. The body was already
+ *  validated at compile time; a failure or kind mismatch here means the
+ *  two stages parsed differently, which is a bug worth a loud error. */
+export function __codeLiteral(source: string, kind: Code["kind"]): Code {
+  const parsed = parseCodeLiteralBody(source);
```
egonSchiele: Three things on this file:

**Registration is an import-order side effect that this module's own import graph doesn't guarantee.** `programParserForLiterals` is set by a top-level statement in `lib/parser.ts`, but nothing reachable from `literal.ts` (`parsers.js`, `code.js`) or from `runtime/index.ts` imports it — the fixtures only work because they `import { fill } from "std::agency"`, which drags in the stdlib template bridge, which imports `parser.js`. A generated program with a program-kind literal that never touches `std::agency`:

```
node main() {
  const t = [| def g(): number { return 1 } |]
  return 1
}
```

hits `programParserForLiterals === null` and throws `internal: a code literal that parsed at compile time failed to re-parse at runtime: internal: program parser for code literals not registered` — an "impossible" message for a perfectly ordinary program. An explicit `import "../../parser.js"` here for the side effect (with a comment saying why) makes the dependency real. A fixture without the `std::agency` import would pin it.

**No memoization.** `parseCodeLiteralBody` runs on every evaluation of the expression, so a literal inside a loop re-parses per iteration. The inputs are compile-time constants, so a module-level `Map` keyed on `source + "\u0000" + kind` is enough. `Code` values are mutated downstream by `fill`, though, so cache the parse and return a fresh clone rather than a shared object.

**Naming.** `literal.ts` sits next to an existing `literals.ts` in the same directory, differing by one character and meaning something unrelated. `codeLiteral.ts` matches `types/codeLiteral.ts` and `parsers/codeLiteral.test.ts`.

packages/agency-lang/lib/typeChecker/synthesizer.ts
```diff
@@ -4,6 +4,36 @@ import { AgencyNode, Expression, VariableType, ValueAccess, formatUnitLiteral }
...
+    "}",
+  ].join("\n");
+  const parsed = typeAliasParser(source);
+  if (!parsed.success) {
+    throw new Error("internal: the Code structural type failed to parse");
+  }
+  return parsed.result.aliasedType;
+}
+
+const CODE_LITERAL_TYPE: VariableType = parseCodeLiteralType();
```
egonSchiele, line 59: Two notes.

This `const` (and `parseCodeLiteralType` above it) sits in the middle of the import block. It works — ESM hoists the imports, so `typeAliasParser` is initialized before the initializer runs — but it reads as if it depends on statement order, and a reader checking that assumption has to know the hoisting rule. Moving both below the last import costs nothing.

More substantively, every `codeLiteral` expression is handed the *same object*. Nothing in the current pass mutates a `VariableType`, and the new `scopes.ts` guard specifically avoids the one function that would have — but that's a property of today's code, and a shared singleton returned from `synthType` is the kind of thing a future narrowing or annotation pass mutates in place, corrupting the type for every literal in the program. `Object.freeze` (deep, or at least on the record and its fields) turns that into an immediate error instead of a spooky one.

Separately: this hand-written copy of stdlib `Code` has to stay in step with `stdlib/agency.agency`. The `fill(tpl, ...)` test catches a *new required* field, but not a renamed or retyped optional one. A test asserting this type is structurally equal to the resolved stdlib `Code` alias would close the loop; if the checker can reach the stdlib alias at that point, deriving it beats duplicating it.

packages/agency-lang/lib/typeChecker/scopes.ts
egonSchiele: The guard is syntactic and only covers a literal in direct value position, so anything that wraps one still gets deep-widened and loses assignability into `fill(template: Code)`:

```
const tpl = cond ? [| 1 |] : [| 2 |]
const tpls = [ [| 1 |] ]
```

Those are less central than the flagship flow, but they fail in exactly the confusing way — a `Code` value that the checker says isn't one, with the literal nowhere near the error.

The root cause is that `widenType` deep-widens record fields that were never literal-*inferred* to begin with, which is arguably the bug. If a targeted fix is preferable to changing `widenType`, marking the synthesized type non-widenable (a flag `widenType` respects) generalizes to every position instead of just this one. Worth at least a comment noting the wrapped cases are knowingly out of scope.

egonSchiele (commented):
## Overview

Adds `[| ... |]` code literals: bodies parse at the enclosing file's parse time in unlowered template mode, kind inferred smallest-first, evaluating to an ordinary `Code` value via `__codeLiteral`. Sequencing the whole-corpus formatter gate first was the right call — the four generator bugs it flushed out (parenthesized base, empty-signature wrap, duplicate node imports, blank-line divergence) are worth the PR on their own.

The leaf-ness treatment is the strongest part: four separate levers (`walkNodes`, `bodySlots` absence, `NO_EXPRESSION_SLOTS`, `WALKER_EXCLUDED_FIELDS`) plus `identifierSlots`' exhaustive registry, each with a forcing function. Reusing the real string/comment parsers for the end-scan instead of writing a second lexer is likewise the right shape. Tests are unusually good — content assertions rather than `.success`, hand-computed location expectations, a sanity anchor for the non-vacuity of the quoted-leaf test, and the `toSource` == file-print equality.

## Main findings

Two correctness gaps, both in the same family as bugs the PR already fixed:

1. **`processValueAccess` paren restoration is an allow-list of three node types and misses `newExpression`.** `(new Foo()).bump()` prints as `new Foo().bump()`, which `newExpressionParser` cannot re-parse — it stops after `)` and takes no chain, which is exactly why `parenAccessParser` exists. `fmt` would produce unparseable output. `tryExpression` has the same shape with milder consequences. The corpus gate missed it because no fixture chains off a parenthesized `new`.

2. **The end-scan treats strings and comments as inert but not regex literals.** `re/[|]/` in a body raises the nested-literal directive error; `re/a|]b/` ends the literal early. The guide's "Nothing needs escaping" is currently false for regexes.

Plus a few smaller things inline: a formatter policy change that's broader than its comment claims, an expr/statements sentinel asymmetry that lets a blank line flip inferred kind, and a runtime registration that depends on import order.

Nothing here is architectural — the design holds up. Detail in the inline comments.

packages/agency-lang/lib/backends/agencyGenerator.ts
egonSchiele: Fixed as suggested, inverted to fail-closed: an ALLOWLIST of self-delimiting primaries (`SAFE_BARE_BASES`) — anything else gets its parens back, so a future looser-binding atom costs a redundant pair instead of corrupted output. Both repros pinned: `(new Foo()).bump()` (bare form unparseable — confirmed strictly worse than the edgar case) and `(try f()).value` (bare form a different program, structural round-trip asserted). The `hole` kind is in the allowlist deliberately so template round-trips do not grow parens.

packages/agency-lang/lib/backends/agencyGenerator.ts
egonSchiele: Resolved via the allowlist inversion in the sibling thread — `tryExpression` is no longer special-cased at all; it simply is not in `SAFE_BARE_BASES`, so it parenthesizes. Pinned repro added.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Both fixed. Regex literals join the inert chunk set (`lazy(() => regexLiteralParser)`) with `r` added to the bulk-scan stop characters, exactly per your sketch. Tests pin both repros structurally: `re/[|]/` parses as a regex with pattern `[|]` (not the nesting error), and `re/a|]b/` keeps its full pattern (the literal does not end early). The guide's inert list now says strings, comments, **and regex literals** — the no-escaping promise is literally true.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Fixed with the sentinel treated symmetrically everywhere: the leading/trailing trivia strip covers whitespace AND `BLANK_LINE_SENTINEL`, the expr branch's rest check goes through `stripSentinels` like the statements branch, and `advancePositionOver` counts sentinels as line breaks (they REPLACED newlines, so the character count and line math stay honest). Pinned: a literal opening with a blank line infers `expr`, not `statements`.

packages/agency-lang/lib/parser.ts
egonSchiele: Fixed — and in both exits, since the non-throw branch had the same mismatch: when a committed failure exists, the structured position is computed from ITS `rest` (the same arithmetic getErrorMessage uses), not the rightmost record. Test added: a malformed literal body's `errorData.line` lands inside the literal's span.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Done for consistency: the scan now runs under a `saveRightmostFailure`/`restoreRightmostFailure` pair, so its probe parsers' internal failed alternatives cannot leak onto the enclosing record. Verified first that restore does not touch the committed-failure slot (it saves/restores only `pos`/`expected`), so the nested-`[|` directive still survives the restore.

packages/agency-lang/lib/runtime/template/literal.ts
egonSchiele: All three taken. (1) `literal.ts` → `codeLiteral.ts`, matching `types/codeLiteral.ts`. (2) The registration is now a real dependency: an explicit side-effect `import "../../parser.js"` with a comment saying exactly why, pinned by a new fixture (`literalNoImports`) whose program has a program-kind literal and no std::agency import — it failed with the internal error before the import, passes now. (3) Parsed bodies are cached per `(kind, source)` in a null-prototype record, returning `deepCopy` clones since `fill` mutates Code values downstream — a literal in a loop parses once.

packages/agency-lang/lib/typeChecker/synthesizer.ts
egonSchiele: All three taken. The block moved below the imports with the hoisting note made explicit. The type is deep-frozen with the rationale you gave (a future in-place pass should fail immediately, not spookily). And the drift concern has a test: the synthesized type is compared structurally (locs/tags stripped) against the `Code` alias parsed from the real `stdlib/agency.agency` — renames and retyped optionals now fail, not just new required fields. Deriving it from the stdlib alias at checker init would be nicer still, but the checker has no clean path to stdlib sources at that point; the drift test makes the copy safe to keep.

packages/agency-lang/lib/typeChecker/synthesizer.ts
egonSchiele: Expanded the comment to answer this properly, because it is a fair question with a war story behind it: the checker's ENCODING of a record type is a grammar decision — `kind?:` becomes a null-union, quoted literals become `stringLiteralType` nodes — and hand-building the tree means re-making those encoding choices by hand. That is exactly what broke the first attempt (a `variableTypeParser`-built tree encoded the fields differently from the alias grammar, and record assignability rejected the mismatch). Parsing the type TEXT through the same type-alias production Code's own declaration uses makes the encodings equal by construction, with the text as the single source of truth — and the new drift test pins it against the real stdlib alias.

packages/agency-lang/lib/typeChecker/scopes.ts
egonSchiele: You and the substrate reviewer converged on this, and the type-directed version won: the guard moved INTO `widenType` as `isCodeShape` (a record whose `type` field is the `"agencyProgram"` literal survives widening unchanged), and the syntactic declare-site guard is deleted. That covers every arrival path — both your repros are now tests (array-of-literals via index access, cross-branch reassignment) — and as a bonus fixes the previously-recorded return-inference limit for free, since `inference.ts` widening flows through the same function.

egonSchiele: Superseded by the widenType/isCodeShape fix in the sibling thread — the guard is no longer syntactic at all.

packages/agency-lang/lib/backends/agencyGenerator.ts
egonSchiele: Decided as policy, stated and pinned: runs of blank lines collapse to one, file-wide — with a direct test (the round-trip gate indeed cannot see it, for exactly the reason you gave) and a sentence on `docs/site/cli/format.md`. Scoping it to the import region was the alternative, but the hoisting-orphaned blanks land mid-stream (the cli.agency case), region detection there is exactly the kind of bookkeeping this PR has been deleting, and collapse-to-one is what the corpus already conforms to (zero corpus diffs from the change).

egonSchiele:
Review round addressed in c92495590 — details in the threaded replies. The two correctness findings are fixed with pinned repros: the paren restoration is inverted to a fail-closed **allowlist** (your `(new Foo()).bump()` and `(try f()).value` repros are both tests now), and **regex literals join the end-scan's inert set** (both `re/[|]/` and `re/a|]b/` pinned structurally). The widening guard moved into `widenType` as the type-directed `isCodeShape` — both reviewers' wrapped-literal repros are tests, and it incidentally fixes the recorded return-inference limit. Blank-line collapse is now stated policy (direct test + format-page doc). Runtime: `codeLiteral.ts` rename, explicit registration import (pinned by the new `literalNoImports` fixture), per-`(kind, source)` parse cache with cloned returns. Full suite 9,430 green, all five execution fixtures, lint clean.

PR #675, opened 2026-07-24
Perf tests (1/3): informational scaling-ratio suite + harness + lint

egonSchiele:
PR 1 of the CI performance-tests plan (spec + reviews under `docs/superpowers/`). Adds the measurement apparatus and proves it end to end on the command that motivated the effort — the linter, whose accidental O(n²) shipped latent because nothing in CI watched for perf regressions. PR 2 (parse/typecheck/fmt + per-stage compile) and PR 3 (LSP + the other commands) follow after this merges.

## The approach: ratios, not milliseconds

Absolute-time thresholds (`compile in < 500ms`) flake on shared runners and rot as machines get faster. Instead we measure **how work grows with input size**. `growthFactor` normalizes the growth by the size step, so a linear algorithm reads **~1.0** and a quadratic reads **~step** (≈8 at an 8× step) regardless of the machine — because the ratio is taken from two measurements in the same run, absolute speed cancels. The schedule **interleaves and order-alternates** its samples so both random second-to-second noise and monotonic drift cancel in the ratio.

## What's here

- **`lib/perf/harness.ts`** — `measureMs`; `growthFactor` (interleaved, alternated, normalized); `expectPerf` (the `PERF_ENFORCE` gate: records-and-passes when unset, throws when set); a JSONL + step-summary recorder; and `cacheFreePath` for the later file-based tests. Meta-tests: a **canary** (a real O(n²) closure must trip the bound), **self-consistency** (linear stays under it — the test the naive `bound = step` design would have failed on correct code), a **schedule test** (fails if interleaving/alternation isn't implemented), and a **recorder test**.
- **`lib/perf/fixtures.ts`** — size-controlled generators, each verified to parse.
- **`lib/perf/lint.perf.test.ts`** — **per-rule** scaling (a single-rule regression can't hide in an all-rules average — exactly how the AL0002 O(n²) hid), each with a **work-happened assertion** so a silently-empty fixture can't pass forever.
- **`vitest.perf.config.ts`** + `test:perf` script; `*.perf.test.ts` excluded from the default unit run (and serialized, long timeout, so tests don't contend).
- An **informational `perf` job** in `test.yml`: single node 22.x, `PERF_ENFORCE` unset, `continue-on-error`, uploads `perf-results.jsonl` for cross-run calibration.

## Proof the suite works

Reverting the linter to its pre-fix O(n²) scan makes `lint:missingDocstring` measure **7.997** (normalized ~8) and, under `PERF_ENFORCE=1`, **fail** the bound of 2. With the indexed path it measures **~1.07** and passes. Local run of all four lint measurements: factors 1.07 / 1.24 / 1.07, stdlib smoke ~835 ms — full `test:perf` in ~14 s.

## Informational only

Nothing here can block a merge until a deliberate flip — set `PERF_ENFORCE=1` and drop `continue-on-error` — after thresholds are calibrated from the uploaded artifacts (start `GROWTH_BOUND = 2.0`). That milestone, and PRs 2–3, are in the plan.

### DIFF
--- a/packages/agency-lang/lib/perf/harness.test.ts
+++ b/packages/agency-lang/lib/perf/harness.test.ts
@@ -1,0 +1,127 @@
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import {
+  measureMs,
+  growthFactor,
+  expectPerf,
+  GROWTH_BOUND,
+} from "./harness.js";
+
+// A linear closure: O(n) work whose result is returned (so it is not
+// dead-code-eliminated). Sized so a single call is comfortably measurable.
+function linearWork(n: number): () => number {
+  const arr = Array.from({ length: n }, (_, i) => i);
+  return () => {
+    let sum = 0;
+    for (let i = 0; i < arr.length; i++) sum += arr[i];
+    return sum;
+  };
+}
+
+// A quadratic closure: O(n^2) work. At an 8x step this grows ~64x, so its
+// normalized growth factor is ~8 — far above the bound.
+function quadraticWork(n: number): () => number {
+  return () => {
+    let sum = 0;
+    for (let i = 0; i < n; i++) {
+      for (let j = 0; j < n; j++) sum += i ^ j;
+    }
+    return sum;
+  };
+}
+
+describe("perf harness", () => {
+  it("measureMs returns a positive, roughly stable time", () => {
+    const fn = linearWork(200_000);
+    const a = measureMs(fn);
+    const b = measureMs(fn);
+    expect(a).toBeGreaterThan(0);
+    // Same closure twice: within a wide tolerance (CI noise), not identical.
+    expect(Math.max(a, b) / Math.min(a, b)).toBeLessThan(5);
+  });
+
+  it("canary: a quadratic algorithm's growth factor is well above the bound", () => {
+    const factor = growthFactor(quadraticWork, 500, 4000);
+    // Normalized quadratic ≈ 8 at an 8x step; assert it clears the bound with room.
+    expect(factor).toBeGreaterThan(GROWTH_BOUND * 1.5);
+  });
+
+  it("self-consistency: a linear algorithm's growth factor is below the bound", () => {
+    // This is the test the old RATIO_BOUND=8-vs-8x-step design would have failed
+    // on correct code: a raw ratio of ~8 is not < 8. Normalization makes it ~1.
+    const factor = growthFactor(linearWork, 500_000, 4_000_000);
+    expect(factor).toBeLessThan(GROWTH_BOUND);
+  });
+
+  it("schedule: samples are interleaved, order-alternated, and warmed up front", () => {
+    const log: number[] = [];
+    const build = (n: number) => () => {
+      log.push(n);
+      return n;
+    };
+    const warmup = 2;
+    const rounds = 4;
+    growthFactor(build, 1, 8, { warmup, rounds });
+
+    // Warmup prefix: `warmup` calls of small, then `warmup` of large — both
+    // sizes warmed before any timed round.
+    const prefix = log.slice(0, 2 * warmup);
+    expect(prefix).toEqual([1, 1, 8, 8]);
+
+    // Timed portion: `rounds` pairs, order alternating small→large, large→small.
+    const timed = log.slice(2 * warmup);
+    expect(timed).toHaveLength(2 * rounds);
+    for (let r = 0; r < rounds; r++) {
+      const pair = timed.slice(2 * r, 2 * r + 2);
+      expect(pair).toEqual(r % 2 === 0 ? [1, 8] : [8, 1]);
+    }
+  });
+});
+
+describe("expectPerf recorder and gate", () => {
+  let resultsFile: string;
+  let savedEnforce: string | undefined;
+  let savedFile: string | undefined;
+
+  beforeEach(() => {
+    resultsFile = path.join(os.tmpdir(), `perf-test-${process.pid}-${Date.now()}.jsonl`);
+    savedEnforce = process.env.PERF_ENFORCE;
+    savedFile = process.env.PERF_RESULTS_FILE;
+    process.env.PERF_RESULTS_FILE = resultsFile;
+    delete process.env.PERF_ENFORCE;
+  });
+
+  afterEach(() => {
+    if (savedEnforce === undefined) delete process.env.PERF_ENFORCE;
+    else process.env.PERF_ENFORCE = savedEnforce;
+    if (savedFile === undefined) delete process.env.PERF_RESULTS_FILE;
+    else process.env.PERF_RESULTS_FILE = savedFile;
+    if (fs.existsSync(resultsFile)) fs.unlinkSync(resultsFile);
+  });
+
+  it("records one parseable line per call, with the label and value", () => {
+    expectPerf("test:label", 1.23, 2.0);
+    const lines = fs.readFileSync(resultsFile, "utf-8").trim().split("\n");
+    expect(lines).toHaveLength(1);
+    const rec = JSON.parse(lines[0]);
+    expect(rec.label).toBe("test:label");
+    expect(rec.value).toBe(1.23);
+    expect(rec.pass).toBe(true);
+  });
+
+  it("does not throw on a breach when PERF_ENFORCE is unset (informational)", () => {
+    expect(() => expectPerf("test:breach", 9.0, 2.0)).not.toThrow();
+  });
+
+  it("throws on a breach when PERF_ENFORCE is set", () => {
+    process.env.PERF_ENFORCE = "1";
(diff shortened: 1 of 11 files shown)


packages/agency-lang/lib/perf/harness.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  it("canary: a quadratic algorithm's growth factor is well above the bound", () => {
+    const factor = growthFactor(quadraticWork, 500, 4000);
+    // Normalized quadratic ≈ 8 at an 8x step; assert it clears the bound with room.
+    expect(factor).toBeGreaterThan(GROWTH_BOUND * 1.5);
+  });
+
+  it("self-consistency: a linear algorithm's growth factor is below the bound", () => {
+    // This is the test the old RATIO_BOUND=8-vs-8x-step design would have failed
+    // on correct code: a raw ratio of ~8 is not < 8. Normalization makes it ~1.
+    const factor = growthFactor(linearWork, 500_000, 4_000_000);
```
egonSchiele: This is the sharpest issue in the PR. `harness.test.ts` is named `*.test.ts`, so it runs in the **default unit suite** (`test:run`), not under `vitest.perf.config.ts`. The default `vitest.config.ts` has no `fileParallelism: false` — vitest runs test files in parallel worker threads — so this timing measurement executes under exactly the CPU contention the perf config's `singleThread`/`fileParallelism: false` was added to avoid.

The risk is concrete: `linearWork(500_000)` vs `4_000_000` are sub-millisecond-to-low-ms closures. Under parallel contention the normalized factor can drift above `GROWTH_BOUND` (2.0), and because this runs in the gating suite, that's a **merge-blocking flake** — the precise outcome the whole "informational-first, never-flaky" design is trying to prevent.

The canary is safe (quadratic normalized ~8 vs bound·1.5=3, huge margin), but self-consistency has almost no margin (~1.0 vs 2.0) on tiny work. Options, roughly in order of preference:
- Move the timing meta-tests (canary + self-consistency + `measureMs` stability) under the serialized perf config, but keep them *asserting* (they don't call `expectPerf`, so they'd still gate there) — or run them in a small dedicated serialized project.
- Or size the closures so absolute time (tens of ms) dwarfs contention noise.

The `measureMs` stability test above (`max/min < 5`) is loose enough to likely survive, and the schedule/recorder tests are timing-independent — it's specifically the `< GROWTH_BOUND` linear assertion I'd worry about.

packages/agency-lang/lib/perf/harness.ts
```diff
@@ -0,0 +1,155 @@
...
+    let tLarge: number;
+    if (round % 2 === 0) {
+      tSmall = timeOnce(smallFn);
+      tLarge = timeOnce(largeFn);
+    } else {
+      tLarge = timeOnce(largeFn);
+      tSmall = timeOnce(smallFn);
+    }
+    ratios.push(tLarge / tSmall);
+  }
```
egonSchiele, line 93: `tLarge / tSmall` has no guard for `tSmall === 0`. For real fixtures (parsing/linting 1000+ nodes) times are always non-zero, so this is latent rather than live — but a closure fast enough to measure as 0ms yields `Infinity`/`NaN`, which then poisons the median. The schedule meta-test only escapes this because it ignores the return value (it asserts on the call log, not the factor).

Cheap hardening: floor each timing at a small epsilon (e.g. `Math.max(t, 1e-4)`) before dividing, so a zero-time sample degrades gracefully instead of producing `Infinity`.

packages/agency-lang/lib/perf/lint.perf.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      // `rule.run` on a parsed string is cache-free (no parseAgencyFileCached).
+      const ctxBySize: Record<number, LintContext> = {
+        [SMALL]: ctxFor(fixture(SMALL)),
+        [LARGE]: ctxFor(fixture(LARGE)),
+      };
+
+      // Work-happened assertion: the rule must actually emit one finding per
+      // declaration on the large input, or a silently-empty fixture would make
+      // the ratio meaningless and pass forever (the bogus-benchmark trap).
+      expect(rule.run(ctxBySize[LARGE]).length).toBe(LARGE);
```
egonSchiele, line 55: The work-happened assertion is exactly right and I'm glad it's here. One cheap addition: it only checks the LARGE fixture. Asserting the SMALL side too (`expect(rule.run(ctxBySize[SMALL]).length).toBe(SMALL)`) guards both measurement points — if a fixture-generation bug made the small input degenerate, the ratio would still be meaningless, and this catches it for the price of one line.

(Confirmed the counts are correct: `missingDocstringRule` filters `node.exported === true`, and `manyFunctions(n, {docstrings:false})` emits n exported undocumented functions, so `=== LARGE` holds. Same for the other two rules.)

packages/agency-lang/vitest.perf.config.ts
```diff
@@ -0,0 +1,29 @@
...
+
+// The performance suite. Kept apart from the default unit run (which excludes
+// `**/*.perf.test.ts`) because these tests are slow by nature and must not run
+// concurrently: parallel test files contend for CPU and poison the timings the
+// whole suite depends on. Mirrors vitest.integration.config.ts's separate-config
+// pattern.
+export default defineConfig({
+  test: {
+    globals: true,
+    environment: "node",
```
egonSchiele, line 10: `perf-results.jsonl` (written by the recorder to `cwd`, i.e. `packages/agency-lang/`) isn't in `.gitignore`. A local `pnpm test:perf` will leave it as an untracked file and dirty the working tree. Worth adding `perf-results.jsonl` to the ignore list in this PR since it's the PR that introduces the file.

egonSchiele (commented):
Reviewed against the plan and its review notes — this is a clean adoption of the earlier feedback. The `growthFactor` normalization (linear → ~1.0, quadratic → ~step) fixes the `bound == step` bug from the plan review; the interleave-and-alternate schedule is implemented and, crucially, *tested* by the schedule meta-test; the work-happened assertion, recorder test, and `cacheFreePath` single-home for cache neutralization are all here. Nicely done.

Four things worth addressing, one of which I think is a real (if intermittent) merge-blocker. Details inline; summary:

1. **(Most important) The timing meta-tests run in the *gating, parallel* unit suite, not the serialized perf suite.** `harness.test.ts` is `*.test.ts`, so it's in the default `test:run` — which has no `fileParallelism: false` (the perf config added that override precisely to stop timing tests from contending). The `self-consistency` assertion (`factor < GROWTH_BOUND`) on sub-millisecond closures can flake under CI contention and, unlike the perf suite, that failure blocks merges — the opposite of the suite's informational-first promise. See the inline note for options.
2. **`perf-results.jsonl` isn't gitignored.** Running `pnpm test:perf` locally drops `packages/agency-lang/perf-results.jsonl` into the tree as an untracked file. Add it to `.gitignore`.
3. **`growthFactor` divides `tLarge / tSmall` with no zero guard** — latent, harmless for real fixtures but a NaN/Infinity trap for fast closures (the schedule test only dodges it by ignoring the return value).
4. **Small:** the work-happened assertion checks only the LARGE fixture; asserting the SMALL side too is nearly free and guards both measurement points.

None of these change the design.

egonSchiele:
Thanks — all four addressed in `4bdbaf5`, plus Copilot's two.

**1. Timing meta-tests in the gating parallel suite (the merge-blocker).** You're right, and this was the real risk. Split the file: the deterministic tests (schedule, recorder, gate) stay in `harness.test.ts` and keep gating in the unit suite; the timing-dependent ones (`measureMs`, canary, self-consistency) moved to a new **serialized** `harness.perf.test.ts`, so they run under the perf config's `fileParallelism: false` and the job's `continue-on-error` — no contention, non-blocking. I also resized the closures to a few ms and switched them to pure arithmetic loops (no large arrays), so single-sample timing is stable even serialized. Verified: the default suite over `lib/perf/` now runs only the 14 deterministic tests; the 3 timing tests run in `test:perf`.

**2. `perf-results.jsonl` not ignored.** Correct, and it turned out to be a deeper mistake on my part: I'd put the entry in a `packages/agency-lang/.gitignore`, but the root `.gitignore` line 3 is `.*`, which sweeps up every dotfile — so that file could never be committed and only worked on my machine. Moved the entry to the tracked **root** `.gitignore` (`**/perf-results.jsonl`) and deleted the stray file.

**3. `tLarge / tSmall` zero guard.** Added — each timing is floored at `MIN_MS` before the division, so a 0ms sample degrades to a finite ratio instead of `Infinity`/`NaN` poisoning the median.

**4. Work-happened only on LARGE.** Added the SMALL assertion too (and thanks for independently confirming the counts).

**Copilot — `Boolean(process.env.PERF_ENFORCE)` treats `"0"` as truthy.** Fixed with `perfEnforceEnabled()`, which treats `0`/`false`/`""`/unset as off; added a test that `PERF_ENFORCE=0` does not gate.

**Copilot — canary sizes in the unit suite** is subsumed by fix #1 (those tests are no longer in the unit suite).

tsc clean, eslint clean, both suites green.

PR #676, opened 2026-07-25
Pattern matching: check a value's shape before reading into it

egonSchiele:
## The bug

An object or array pattern lowered to a bare member-access chain, with nothing establishing that the value had the shape the pattern requires. A pattern's job is to decide whether a value matches — and the case that did *not* match was the case that crashed:

```ts
type Redirect = { op: string, path: string }
type Cmd = { words: string[], redirect: Redirect | null }

match (c) {                                            // c.redirect is null
  { words: ["echo", ...rest], redirect: { op: ">", path } } => "write ${path}"
  _                                                         => "fallback"
}
// expected: "fallback"
// actual:   failure("Cannot read properties of null (reading 'op')")
```

Today it's masked by `&&` short-circuiting whenever an earlier check in the same arm happens to fail first. Reorder the arms and it surfaces.

The same missing check produced **wrong answers**, not only crashes:

| pattern | value | before | after |
|---|---|---|---|
| `{ a: 1 }` / `{ a }` | `null` | crash | no match |
| `[a, b]` | `null` | crash | no match |
| `{ a }` | `3` | matches, binds `a = undefined` | no match |
| `[a, b]` | `"abc"` | matches, binds characters | no match |
| `{ a: 1 }` | `3` | no match — by accident, `(3).a` is `undefined` | no match, checked |

## The fix

`collectChecks` (`lib/lowering/patternLowering.ts`) emits a shape check at the head of the `objectPattern` and `arrayPattern` cases:

```ts
source != null && __coarseTypeTest(source, "object")   // object patterns
source != null && __coarseTypeTest(source, "array")    // array patterns
```

`__coarseTypeTest` is the same runtime test a `p: Type` arm already emits, so codegen, the formatter, and the typechecker all handle it unchanged.

**Why the pair and not just the coarse test.** The coarse test alone is sufficient at runtime. The `!= null` in front is for the type checker: narrowing by `is object` lands on the *opaque* `object` type, which has no fields, so a pattern reading through a `Redirect | null` field reports `AG2011: Property 'op' does not exist on type 'object'` despite running correctly. `!= null` filters the union properly and the coarse test that follows does not clobber it. That narrowing gap is written up in `docs/superpowers/ideas/2026-07-24-is-object-narrowing-loses-union-members.md`; the `!= null` half can be removed once it's fixed.

## Behavior changes beyond the crash

Each covered by a test:

- `{ a }` no longer matches a non-object; `[a, b]` no longer matches a string.
- An object pattern does not match an array, consistent with Agency's `object` being defined as non-null and non-array. `[a, b]` is the spelling for arrays.
- `match (3 is { s, b })` now reports the head-pattern mismatch as a failure instead of binding `undefined` and running an arm. A binder-only head pattern previously produced no condition at all, so it had nothing to gate on.

## Verification

- New `tests/agency/pattern-matching/patternShapeGuard.agency` — 4/4 passing, **1/4 before the fix**, with a positive-control node so the test isn't vacuous.
- Full `lib` suite: 9418 passed; the only failures are 5 pre-existing ones (`mcpBridge.contract` ×4, `optimize/mutator` ×1), verified failing on a clean tree.
- Agency execution tests: `pattern-matching` 16/16, `destructuring` 6/6, plus `match`, `matchExpression`, `type-patterns`, `match-expr-in-handler`, `goto-match-arm`, `module-level-match`, `match-yield-seq-thread`, `regexMatch` — all pass.
- Three expectations updated, all of which pinned the old condition shape. Exactly **one** fixture line changed across the entire fixture set (`type-patterns.mjs`: a `{name, age}: Person` arm now carries the object test after its schema validation — redundant there, left in rather than special-cased).
- `make` and `pnpm run lint:structure` clean.

## Note

`lib/typeChecker/strictMemberAccess.test.ts` is also touched by #674; both append at the end of the file, so expect a trivial conflict whichever lands second.

### DIFF
--- a/packages/agency-lang/lib/lowering/patternLowering.test.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.test.ts
@@ -180,15 +180,27 @@ describe("object destructuring", () => {
 // ---------------------------------------------------------------------------
 
 describe("`is` operator in boolean context", () => {
-  it("lowers `const r = x is { type: \"foo\" }` to equality check", () => {
+  it("lowers `const r = x is { type: \"foo\" }` to a shape check AND an equality check", () => {
     const lowered = lower(`let x = { type: "foo" }\nconst r = x is { type: "foo" }`);
-    // [x, r = (x.type == "foo")]
+    // [x, r = (isObject(x) && x.type == "foo")]
     expect(lowered).toHaveLength(2);
     const rAssign = lowered[1] as Assignment;
     expect(rAssign.variableName).toBe("r");
     const value = rAssign.value as BinOpExpression;
     expect(value.type).toBe("binOpExpression");
-    expect(value.operator).toBe("==");
+    expect(value.operator).toBe("&&");
+    // The shape check comes first, so the field read on the right is only
+    // reached for a value that has fields — `3 is {type:"foo"}` and
+    // `null is {type:"foo"}` both answer false instead of reading `.type`.
+    // Left is the shape check pair `(x != null && isObject(x))`.
+    const shape = value.left as BinOpExpression;
+    expect(shape.operator).toBe("&&");
+    expect(shape.left).toMatchObject({ type: "binOpExpression", operator: "!=" });
+    expect(shape.right).toMatchObject({
+      type: "typeTestExpression",
+      typeHint: { type: "primitiveType", value: "object" },
+    });
+    expect(value.right).toMatchObject({ type: "binOpExpression", operator: "==" });
   });
 
   it("throws on shorthand binder in pure-boolean `is` context", () => {
@@ -349,12 +361,31 @@ describe("match block lowering", () => {
       ],
     } as MatchBlock;
     const lowered = lowerPatterns([matchBlock]);
-    // [scrutinee, assert (object pattern), s = __scrutinee.s, b = __scrutinee.b, ifElse]
-    const sIdx = lowered.findIndex((n) => n.type === "assignment" && (n as Assignment).variableName === "s");
-    const bIdx = lowered.findIndex((n) => n.type === "assignment" && (n as Assignment).variableName === "b");
+    // [scrutinee,
+    //  if (__scrutinee != null && isObject(__scrutinee)) { s = …, b = …, ifElse }
+    //  else return failure(…)]
+    //
+    // Even a binder-only head pattern carries a shape check now, so the
+    // bindings sit inside the head-pattern guard rather than at the top level.
+    // `match (3 is { s, b })` therefore reports the head mismatch as a failure
+    // instead of binding `undefined` and running an arm.
+    expect(lowered).toHaveLength(2);
+    const headGuard = lowered[1] as IfElse;
+    expect(headGuard.type).toBe("ifElse");
+    const shape = headGuard.condition as BinOpExpression;
+    expect(shape.operator).toBe("&&");
+    expect(shape.left).toMatchObject({ type: "binOpExpression", operator: "!=" });
+    expect(shape.right).toMatchObject({
+      type: "typeTestExpression",
+      typeHint: { type: "primitiveType", value: "object" },
+    });
+
+    const inside = headGuard.thenBody;
+    const sIdx = inside.findIndex((n) => n.type === "assignment" && (n as Assignment).variableName === "s");
+    const bIdx = inside.findIndex((n) => n.type === "assignment" && (n as Assignment).variableName === "b");
     expect(sIdx).toBeGreaterThan(-1);
     expect(bIdx).toBeGreaterThan(sIdx);
-    const ifNode = lowered[lowered.length - 1] as IfElse;
+    const ifNode = inside[inside.length - 1] as IfElse;
     expect(ifNode.type).toBe("ifElse");
     // The first arm condition is the guard `s > 5`
     expect((ifNode.condition as BinOpExpression).operator).toBe(">");
--- a/packages/agency-lang/lib/lowering/patternLowering.test.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.test.ts
@@ -182,7 +182,7 @@ describe("object destructuring", () => {
 describe("`is` operator in boolean context", () => {
   it("lowers `const r = x is { type: \"foo\" }` to a shape check AND an equality check", () => {
     const lowered = lower(`let x = { type: "foo" }\nconst r = x is { type: "foo" }`);
-    // [x, r = (isObject(x) && x.type == "foo")]
+    // [x, r = ((x != null && isObject(x)) && x.type == "foo")]
     expect(lowered).toHaveLength(2);
     const rAssign = lowered[1] as Assignment;
     expect(rAssign.variableName).toBe("r");
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -1060,6 +1060,9 @@ const ARRAY_HINT: VariableType = {
  * filters the union properly and leaves `Redirect`, and the coarse test that
  * follows does not clobber that. Two cheap comparisons buy a pattern that
  * both runs right and typechecks.
+ *
+ * The `!= null` half can be dropped once that narrowing gap is fixed:
+ * docs/superpowers/ideas/2026-07-24-is-object-narrowing-loses-union-members.md
  */
 function shapeCheck(
   source: Expression,
@@ -1084,6 +1087,11 @@ function collectChecks(pattern: MatchPattern, source: Expression, checks: Expres
       // shape has to fail the arm rather than throw on `null.a` — or match by
       // reading `undefined` off a number. `&&` short-circuits, so this also
       // makes the reads below safe.
+      //
+      // Note this rejects ARRAYS too, not only null and non-objects: Agency's
+      // `object` is non-null and non-array (lib/runtime/typeTest.ts). So
+      // `{ length: 2 }` no longer matches an array — `[a, b]` is the spelling
+      // for that. One definition of `object`, not two.
       checks.push(...shapeCheck(source, OBJECT_HINT, pattern.loc));
       for (const prop of pattern.properties) {
         if (prop.type === "objectPatternProperty") {
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -24,6 +24,7 @@ import type {
   MatchBlock,
   MatchBlockCase,
   ReturnStatement,
+  VariableType,
   WhileLoop,
 } from "../types.js";
 import { isExpressionNode } from "../types.js";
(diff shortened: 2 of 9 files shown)


packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -1050,6 +1093,10 @@ function collectChecks(pattern: MatchPattern, source: Expression, checks: Expres
       }
       break;
     case "arrayPattern": {
+      // Shape check FIRST, for the same reason — and because `.length` alone
+      // does not distinguish an array from a string: `"abc"` has length 3 and
+      // would otherwise match `[a, b]`, binding characters.
+      checks.push(...shapeCheck(source, ARRAY_HINT, pattern.loc));
```
egonSchiele, line 1107: Does the solution compose well with the other solutions here? What I don't want is for us to keep adding band-aid fixes based on every new failing case. I would like us to think holistically about the system, and write code that makes it hard to make mistakes.

packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -1039,9 +1040,51 @@ function patternToCondition(pattern: MatchPattern, source: Expression): Expressi
...
+
 function collectChecks(pattern: MatchPattern, source: Expression, checks: Expression[]): void {
   switch (pattern.type) {
     case "objectPattern":
+      // Shape check FIRST: everything below reads a field off `source`, and a
+      // pattern's job is to answer "does this match", so a value of the wrong
+      // shape has to fail the arm rather than throw on `null.a` — or match by
+      // reading `undefined` off a number. `&&` short-circuits, so this also
+      // makes the reads below safe.
+      checks.push(...shapeCheck(source, OBJECT_HINT, pattern.loc));
```
egonSchiele, line 1095: Worth spelling out in this comment that the object test also rejects **arrays**, not just null and non-objects. It follows from Agency's `object` being non-null and non-array, but it is the one consequence a reader will not predict from "shape check", and it is a real behavior change for anyone who had `{ length: n }` or an index-keyed pattern pointed at an array.

packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -1039,9 +1040,51 @@ function patternToCondition(pattern: MatchPattern, source: Expression): Expressi
...
+  loc: SourceLocation | undefined,
+): Expression[] {
+  const notNull = makeBinOp(cloneExpr(source), "!=", nullLit(loc), loc);
+  const coarse = {
+    type: "typeTestExpression",
+    expression: cloneExpr(source),
+    typeHint,
+    loc,
+  } as Expression;
+  return [notNull, coarse];
```
egonSchiele, line 1079: The comment is good — it explains the *why*, which is exactly what someone deleting the `!= null` later will need.

One addition: link the idea doc by path here (`docs/superpowers/ideas/2026-07-24-is-object-narrowing-loses-union-members.md`). "Remove once the narrowing gap is fixed" is only actionable if the reader can find the gap. A grep for `shapeCheck` from that doc works in one direction but not the other.

packages/agency-lang/tests/agency/pattern-matching/patternShapeGuard.agency
```diff
@@ -0,0 +1,61 @@
...
+
+node nullIntermediate() {
+  // The redirected arm is FIRST, so no earlier `&&` can short-circuit the
+  // null read away. This is the case that crashes today.
+  return nested({ words: ["echo", "hi"], redirect: null })
+}
+
+node wrongShape() {
+  // A number is not an object; a string is not an array. Neither matches.
+  return [objBinder(3), arrBinder("abc"), objLiteral("hello")]
```
egonSchiele: The PR description lists "an object pattern does not match an array" as a deliberate behavior change, but nothing here covers it — this line tests a number and two strings only. That case is the least intuitive of the three, so it is the one most likely to get quietly undone by a later change.

```ts
return [objBinder(3), arrBinder("abc"), objLiteral("hello"), objBinder([1, 2]), arrBinder({ a: 1 })]
```

The second addition covers the mirror case: an array pattern against an object, which today would pass a `.length` check of `undefined >= 2`... actually falsy, so it already fails — but by accident rather than by the check, which is the same accident the description's table calls out for `{ a: 1 }` against `3`.

egonSchiele (commented):
## Overview

This makes an object or array pattern check the value's shape before reading into it, so a value of the wrong shape fails the arm instead of throwing (or matching by reading `undefined` off a number). The fix lands in `collectChecks`, reuses the existing `__coarseTypeTest`, and adds one mechanism rather than special cases. That is the right layer and, I think, close to the simplest correct fix — see the note on the `!= null` half below.

The writeup, the before/after table, and the positive-control node in the test are all excellent. The `nullIntermediate` node putting the crashing arm first so no earlier `&&` can hide the bug is exactly the right instinct.

## Is this the right approach?

Yes. Two alternatives I considered and think you correctly did not take:

- **Optional chaining in `fieldAccess`** (`source?.a`). Smaller diff, fixes the crash, but leaves every wrong-match case in the table untouched — `{ a }` would still match `3`. The PR is right that the wrong answers matter as much as the crash.
- **A `!= null` guard only.** Same problem: fixes the crash, leaves `[a, b]` matching `"abc"`.

The shape check is the thing a pattern actually asserts, so asserting it is the honest fix.

## Is it overly complicated?

Mostly no. One piece of genuine accidental complexity: the `!= null` half of the pair, which exists only because `is object` narrowing loses union members. It is well documented, both in the code comment and in a linked idea doc with a removal plan, so shipping it is defensible.

Worth weighing though: the narrowing fix you describe in `2026-07-24-is-object-narrowing-loses-union-members.md` sounds small — filter the union the way `narrowUnionByDiscriminant` already does. Doing it in this PR would let `shapeCheck` emit one check instead of two, halve the added condition size everywhere, and remove the comment that has to explain the workaround. If that turns out to be more than a contained change, shipping the pair now is the right call.

I'd leave the redundancy in the `type-patterns` fixture alone. "One mechanism beats a special case" is the right trade for one extra comparison after a schema validation.

## Issues

### 1. The head-pattern comment is now wrong, and the change it describes is bigger than it was

`patternLowering.ts:758-761` says:

> If the head pattern includes literal value checks (e.g. `match (x is { type: "a", v })`), gate the bindings + arm chain on those checks. When the pattern doesn't match at runtime the user has indicated a precondition [...]

After this change `patternToCondition` returns non-null for *every* object or array head pattern, including binder-only ones. So `match (cfg is { host, port })` now compiles to a guard whose else-branch does `return failure("... head pattern did not match")` from the enclosing function — and in that case the user indicated no precondition at all. The behavior is defensible (it is the same argument as the rest of the PR), but the comment now describes a rule that no longer holds and will mislead the next reader. Please update it.

The thing I'd want checked before merging: that early `return failure(...)` now appears in far more functions than before, including ones whose declared return type is not a Result. It was reachable before only for literal head patterns, so I assume the typechecker tolerates it — but the blast radius just grew a lot, and a test with a binder-only head pattern inside a `def f(): string` would settle it.

### 2. `[a, b]` still matches a 3-element array

`patternLowering.ts:1104`:

```ts
const op: Operator = hasRest ? ">=" : ">=";
```

Both branches are identical, so the ternary is dead — the `hasRest` distinction was clearly meant to be there and got lost. The substantive consequence is that `[a, b]` matches `[1, 2, 3]`, which is the same family of "this pattern matches something it shouldn't" that the rest of the PR is fixing. `[a, b, ...rest]` is the spelling for "at least two".

This is pre-existing, not introduced here, so it's your call whether it belongs in this PR. But the PR touches these exact lines and the argument for fixing it is the argument you already make in the description, so it seems like a natural inclusion. At minimum, collapse the dead ternary.

### 3. "An object pattern does not match an array" has no test

The description calls this out as a deliberate behavior change, and it's the least obvious one of the set — `{ a }` not matching `3` will surprise nobody, but `{ ... }` not matching `[1, 2]` might. `wrongShape` covers only `3`, `"abc"`, and `"hello"`. See the inline comment.

Two other cases I'd add while you're there: an array pattern against an object (`arrBinder({ a: 1 })`), and the `match (3 is { s, b })` failure at runtime — right now that one is asserted only through the shape of the lowered AST in the unit test, not through the failure a user would actually see.

### 4. The guide contradicts the new behavior

`docs/site/guide/pattern-matching.md:360` "Failure semantics":

> Destructuring relies on the underlying JavaScript runtime: reading a property of `null` or `undefined` [...] throws a `TypeError`, which Agency captures and surfaces as a `failure` Result.

That now describes declarations only. Match arms fall through instead, which is the whole point of this PR and a user-visible improvement worth documenting. Your idea doc already flags this ("Worth a doc note once fixed") — since the behavior is shipping here, the note should ship here too.

The type-pattern section explains that `object` means non-null and non-array, but nothing connects that to object *patterns*. A reader hitting behavior change #3 has no page to find it on.

## Smaller notes

- Nested patterns re-read the access chain three times (`c.redirect != null && __coarseTypeTest(c.redirect, "object") && c.redirect.op == ">"`). Cheap for property reads, and I could not construct a case where the scrutinee is a call — the parser rejects `f() is { ... }` — so this looks harmless. Worth a thought if that parser restriction ever lifts, since `__coarseTypeTest`'s own docstring says it exists so the tested value is evaluated exactly once.
- The rewritten note at the bottom of `patternLowering.ts`, drawing the line between binding extraction and matching, is a good addition. That distinction is the thing a future reader is most likely to get wrong.

## Verdict

The approach is right and the implementation is clean. Item 1 (stale comment plus the widened failure-return) and item 4 (docs) I'd want before merge; item 3 is a small test addition; item 2 is optional but tempting.

packages/agency-lang/lib/lowering/patternLowering.ts
egonSchiele: Reasoned about this rather than patched around it, and the conclusion is that the rule is not new — four of the six `collectChecks` cases already followed it.

The contract a pattern's condition owes is that it is a **total function of the value**: safe to evaluate on anything, true only for values the pattern matches. Auditing every case against that:

| case | how it interrogates the value | total? |
|---|---|---|
| `resultPattern` | `isSuccess(source)` | yes — `result != null && typeof === "object" && …` |
| `typePattern` | `__coarseTypeTest(source, kind)` | yes |
| binders / wildcard / rest | nothing | yes |
| literal | `source == literal` | yes — `null == 1` is false |
| `objectPattern` | **`source.key`** | **no** |
| `arrayPattern` | **`source.length`, `source[i]`** | **no** |

Raw member access is the only non-total operation in the file, and it is exactly the two broken cases. The shape check makes them conform to what the other four already did — it is not a sixth special case.

That is also why one check covers two symptoms that looked unrelated. The crash (`{a: 1}` vs `null`) is a totality failure; the wrong match (`{a}` vs `3`) is a soundness failure, a binder-only pattern pushing no checks and so being vacuously true. Both are the same missing assertion: *this value has the shape the case is about to destructure*. And it composes recursively by construction — each case guards its own destructuring before recursing with `fieldAccess(source, key)` as the child's source, so nested patterns are safe by induction.

You are right that nothing enforced any of this, which is what makes the next one a band-aid. Two tripwires added in 8b798e7bb, both verified to fail without their fix rather than assumed to work:

**Invariant 1** (`patternGuards.tripwire.test.ts`) — in a lowered pattern condition, every member access is preceded in the `&&` chain by a shape check on each of its prefixes. Patterns come from the corpus (stdlib, generator fixtures, agency tests) rather than hand-written representatives, and the kind list is tied to the `MatchPattern` union by a compile-time exhaustiveness check, so adding a kind without listing it is a build error. `KNOWN_UNCOVERED_KINDS` has a staleness guard so an entry cannot outlive its reason. Modelled on the walker-coverage corpus in `lib/utils/expressionSlots.test.ts`. Teeth check: removing the shape check makes it fail on real `stdlib/ui.agency` and `stdlib/data/people/littlesis.agency` patterns.

**Invariant 2** — no pattern-specific node survives lowering (skipping the deliberate `matchSource` snapshot). `patternLowering.ts` has claimed exactly this in its header from the start with nothing checking it, and it is the invariant #677's bug violated: an unlowered `isExpression` reaching codegen. Teeth check: dropping a program with `is` in a guard into the corpus fails it. That one would have caught #677 with nobody thinking about guards at all.

So the answer to "does it compose" is that invariant 2 covers #677's class, invariant 1 covers this PR's class, and #3 (the nested type suffix) adds new pattern positions that invariant 1 will police for free.

packages/agency-lang/lib/lowering/patternLowering.test.ts
```diff
@@ -180,15 +180,27 @@ describe("object destructuring", () => {
 // ---------------------------------------------------------------------------
 
 describe("`is` operator in boolean context", () => {
-  it("lowers `const r = x is { type: \"foo\" }` to equality check", () => {
+  it("lowers `const r = x is { type: \"foo\" }` to a shape check AND an equality check", () => {
     const lowered = lower(`let x = { type: "foo" }\nconst r = x is { type: "foo" }`);
-    // [x, r = (x.type == "foo")]
+    // [x, r = (isObject(x) && x.type == "foo")]
```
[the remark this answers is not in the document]
egonSchiele: Fixed — now reads `// [x, r = ((x != null && isObject(x)) && x.type == "foo")]`, which is the actual shape.

packages/agency-lang/lib/lowering/patternLowering.ts
egonSchiele: Added. The comment now says the object test rejects arrays too, that it follows from Agency's `object` being non-null and non-array, and that `[a, b]` is the spelling for arrays — one definition of `object` rather than two.

Also covered in the execution test now, since it was the behavior change least likely to survive a later refactor unnoticed.

packages/agency-lang/lib/lowering/patternLowering.ts
egonSchiele: Linked. The comment now ends with the doc path, so whoever deletes the `!= null` half can find why it is there without grepping in the other direction.

packages/agency-lang/tests/agency/pattern-matching/patternShapeGuard.agency
egonSchiele: Added both, with your reading of the second one recorded in the comment — an array pattern against an object failed before this change too, but by accident (`({a:1}).length` is undefined, `undefined >= 2` is false), and now fails because the check says so. Same accident the description's table calls out for `{ a: 1 }` against `3`.

`wrongShape` is now five cases: number, string, string-vs-object-literal, array-vs-object-pattern, object-vs-array-pattern.

PR #677, opened 2026-07-25
Pattern matching: lower the `is` operator inside a match-arm guard

egonSchiele:
Second of the pattern-matching fixes blocking the safeBash command matcher. Independent of #676 — different function in the same file.

## The bug

Using `is` in a match-arm guard crashes the compiler:

```ts
match (words) {
  ["echo", str] if (str is string) => "echo: ${str}"
  _                                => "none"
}
```

```
Error: Unhandled Agency node type: isExpression
    at TypeScriptBuilder.processNode (typescriptBuilder.js:568)
    at TypeScriptBuilder.processIfElseWithSteps (typescriptBuilder.js:1231)
```

A stack trace, not a diagnostic. The same `is` in an ordinary `if` condition works fine, so the failure looks arbitrary from the outside.

## Cause

`isExpression` nodes are meant to be gone by codegen — `lowerExpression` rewrites each one into its boolean condition form. `foldArms` lowers the arm body (`this.lowerBody(arm.body)`) and the pattern (via `patternToCondition`), but used `arm.guard` **raw** at both use sites:

- no bindings → folded in with `makeBinOp(condition, "&&", arm.guard, loc)`
- with bindings → becomes an inner `ifElse`'s `condition: arm.guard`

The `guardOnly` branch had the same hole: in the `match(x is pattern)` form each arm's `caseValue` *is* the guard expression, and it was passed through raw too. Confirmed broken, not just suspected — the RED run crashes at both sites.

## The fix

Both now go through `lowerExpression`. Three-line change plus comments.

A side effect worth noting: `lowerExpression` calls `assertNoBindersInBoolIs`, so a shorthand binder in a guard is now the existing compile error —

```
shorthand binder in pure-boolean `is` context has nowhere to bind;
use `if (x is { ... })` to introduce variables
```

— rather than a crash or a silent binding. That's the right behavior and comes for free.

## Verification

Test-first; I watched each case fail before fixing.

- New `tests/agency/pattern-matching/matchGuardIs.agency` — 5/5 passing, **0/5 before the fix** (compile crash). Covers: a guard that holds and whose binding is used; a guard that fails and falls through to the next arm; no arm matching; a **named type** in a guard, which routes through schema validation rather than a coarse check, so the validator decides the match; and the `match(x is pattern)` arm-condition form.
- Two unit tests in `patternLowering.test.ts` asserting no `isExpression` survives lowering — one per hole. Both verified failing with the fix stashed.
- Full `lib` suite: 9419 passed; only the 5 pre-existing failures (`mcpBridge.contract` ×4, `optimize/mutator` ×1), verified failing on a clean tree.
- Agency execution tests: `pattern-matching` 17/17, `destructuring` 6/6, plus `match`, `matchExpression`, `type-patterns`, `match-expr-in-handler`, `goto-match-arm`, `module-level-match`, `match-yield-seq-thread` — all pass.
- `make fixtures` produced **no** changes: nothing in the fixture set used `is` in a guard, which is consistent with this having gone unnoticed.
- `make` and `pnpm run lint:structure` clean.

### DIFF
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -884,8 +884,11 @@ class PatternLowerer {
       let bindings: Assignment[] = [];
 
       if (guardOnly) {
-        // The caseValue IS the guard expression in match(... is ...) form
-        condition = arm.caseValue as Expression;
+        // The caseValue IS the guard expression in match(... is ...) form.
+        // Lower it: an arm condition is an ordinary boolean expression, so an
+        // `is` in it must become its condition form rather than reach codegen,
+        // which has no isExpression handler.
+        condition = this.lowerExpression(arm.caseValue as Expression);
       } else {
         const matchPat = arm.caseValue as MatchPattern;
         const patCond = patternToCondition(matchPat, scrutinee!);
@@ -895,6 +898,11 @@ class PatternLowerer {
 
       // Apply guard if present (for non-guardOnly arms)
       if (!guardOnly && arm.guard) {
+        // A guard is a pure-boolean context, so lower it like any other
+        // expression — `if (str is string)` becomes the boolean test, and a
+        // shorthand binder inside it stays the existing compile error rather
+        // than silently introducing a binding.
+        const guard = this.lowerExpression(arm.guard);
         // Guard may reference bindings, so we need bindings in scope first.
         // Strategy: emit `if (patCond) { let bindings; if (guard) { body } else <next-arm> }
         //                  else <next-arm>`
@@ -905,11 +913,11 @@ class PatternLowerer {
         // outer and inner else branches so guard failure falls through to the
         // next arm (it would otherwise exit the match entirely).
         if (bindings.length === 0) {
-          condition = makeBinOp(condition, "&&", arm.guard, loc);
+          condition = makeBinOp(condition, "&&", guard, loc);
         } else {
           const innerIf: IfElse = {
             type: "ifElse",
-            condition: arm.guard,
+            condition: guard,
             thenBody: armBody,
             elseBody: result ? [result] : undefined,
             loc,
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -419,6 +419,12 @@ class PatternLowerer {
       this.assertNoStatementArmReturns(node);
     }
 
+    // Lower every arm's expression slots ONCE, before any of the three paths
+    // below reads them. `node.cases` stays un-lowered for the `matchSource`
+    // snapshot, which the exhaustiveness pass reads and which wants the arms
+    // as written.
+    const cases = this.lowerArmExpressionsIn(node.cases);
+
     // Check if any arm uses patterns or has a guard.
     const hasPatternArms = node.cases.some(
       (c) =>
@@ -437,7 +443,7 @@ class PatternLowerer {
       return [
         {
           ...node,
-          cases: node.cases.map((c) => (c.type === "matchBlockCase" ? this.lowerMatchCase(c) : c)),
+          cases: cases.map((c) => (c.type === "matchBlockCase" ? this.lowerMatchCase(c) : c)),
           ...(matchExprId !== undefined ? { matchExprId } : {}),
         },
       ];
@@ -470,7 +476,7 @@ class PatternLowerer {
     };
     const scrutineeRef = varRef(scrutineeName, node.loc);
 
-    const ifChain = this.buildIfChainFromArms(node.cases, scrutineeRef, node.loc);
+    const ifChain = this.buildIfChainFromArms(cases, scrutineeRef, node.loc);
     if (!ifChain) return [scrutineeAssign];
     // Tag the root of the lowered if-chain: it is the node that OWNS the
     // matchExprId (yields unwind to it). Attached at construction.
@@ -749,8 +755,12 @@ class PatternLowerer {
     const scrutineeRef = varRef(scrutineeName, node.loc);
     const bindings = this.extractBindings(isExpr.pattern, scrutineeRef, "const", node.loc);
 
-    // Each arm's caseValue is now a guard expression; build if/else chain over them.
-    const ifChain = this.buildIfChainFromGuardArms(node.cases, node.loc);
+    // Each arm's caseValue is now a guard expression; build if/else chain over
+    // them. Lowered up front, same as the pattern-arm form.
+    const ifChain = this.buildIfChainFromGuardArms(
+      this.lowerArmExpressionsIn(node.cases),
+      node.loc,
+    );
     const inside: AgencyNode[] = [...bindings];
     if (ifChain) inside.push(ifChain);
 
@@ -825,6 +835,35 @@ class PatternLowerer {
     return { ...c, body };
   }
 
+  /**
+   * Lower an arm's two EXPRESSION slots — its `caseValue` when that is an
+   * expression rather than a pattern, and its `guard`.
+   *
+   * The single funnel for both, on purpose. Lowering is opt-in per field, so
+   * every consumer that reads one of these slots is a chance to forget:
+   * `foldArms` used the guard raw, `lowerMatchCase` passes the caseValue
+   * through untouched by design, and `collectChecks`'s literal case pushes the
+   * caseValue straight into a comparison. Three readers, one of which can
+   * always be missed. Normalizing at the two entry points instead means a
+   * reader cannot receive an un-lowered arm no matter what it does with it.
+   *
+   * A pattern caseValue is left alone — a pattern is not an expression, and
+   * `patternToCondition` is what turns it into one. `_` is a bare string.
+   */
+  private lowerArmExpressions(c: MatchBlockCase): MatchBlockCase {
+    const caseValue =
+      c.caseValue !== "_" && isExpressionNode(c.caseValue)
+        ? this.lowerExpression(c.caseValue as Expression)
+        : c.caseValue;
+    const guard = c.guard ? this.lowerExpression(c.guard) : c.guard;
+    return { ...c, caseValue, guard } as MatchBlockCase;
+  }
+
+  /** Apply `lowerArmExpressions` to every arm of a match. */
+  private lowerArmExpressionsIn(cases: MatchBlock["cases"]): MatchBlock["cases"] {
+    return cases.map((c) => (c.type === "matchBlockCase" ? this.lowerArmExpressions(c) : c));
+  }
+
   /** Build an if/else chain from arms with patterns/guards/literals. */
   private buildIfChainFromArms(
     cases: MatchBlock["cases"],
@@ -863,7 +902,7 @@ class PatternLowerer {
       const arm = arms[i];
       const armBody = this.lowerBody(arm.body);
 
-      if (arm.caseValue === "_") {
+      if (arm.caseValue === "_" && !arm.guard) {
         // Default case: becomes an else-body. If we already have an `if`, attach as elseBody.
         if (result) {
           result = { ...result, elseBody: armBody };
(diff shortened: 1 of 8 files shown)


packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -884,8 +884,11 @@ class PatternLowerer {
       let bindings: Assignment[] = [];
 
       if (guardOnly) {
-        // The caseValue IS the guard expression in match(... is ...) form
-        condition = arm.caseValue as Expression;
+        // The caseValue IS the guard expression in match(... is ...) form.
+        // Lower it: an arm condition is an ordinary boolean expression, so an
+        // `is` in it must become its condition form rather than reach codegen,
+        // which has no isExpression handler.
+        condition = this.lowerExpression(arm.caseValue as Expression);
```
egonSchiele: Correct fix for this site. Two sibling sites take a caseValue equally raw and still crash — see items 1 and 2 in the review body:

- `lowerMatchCase` (`:823`), reached via the `!hasPatternArms` early return at `:437`, which is documented as recursing into the body "without lowering the arm itself"
- `collectChecks`'s default case, which does `makeBinOp(source, "==", pattern as Expression, ...)`

Your idea doc flagged this one as "should be checked for the same hole" and you checked it. The same sentence applies to those two.

packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -895,6 +898,11 @@ class PatternLowerer {
 
       // Apply guard if present (for non-guardOnly arms)
       if (!guardOnly && arm.guard) {
+        // A guard is a pure-boolean context, so lower it like any other
+        // expression — `if (str is string)` becomes the boolean test, and a
+        // shorthand binder inside it stays the existing compile error rather
+        // than silently introducing a binding.
+        const guard = this.lowerExpression(arm.guard);
```
egonSchiele: The comment is accurate, but the error a user actually sees is confusing here. `assertNoBindersInBoolIs` says "use `if (x is { ... })` to introduce variables" — and a guard is *spelled* with `if`, so someone writing

```ts
["echo", str] if (str is { policy }) => ...
```

is told to write the thing they just wrote. Worth threading a flag through for a guard-specific message, or rewording the advice so it names the construct instead of showing a shape the guard already has.

packages/agency-lang/tests/agency/pattern-matching/matchGuardIs.agency
```diff
@@ -0,0 +1,68 @@
...
+  return failure("not an adult")
+}
+
+@validate(isAdult)
+type Adult = { name: string, age: number }
+
+def describe(words: any[]): string {
+  return match (words) {
+    ["echo", str] if (str is string) => "echo string: ${str}"
+    ["echo", n] if (n is number) => "echo number: ${n}"
```
egonSchiele, line 18: Every guarded arm in this file binds something, so they all take the `innerIf` branch at `patternLowering.ts:918`. The other line you changed — `:916`, the `bindings.length === 0` fold into `&&` — never runs in any test here.

A literal-only pattern with a guard over an outer variable covers it:

```ts
def describe(words: any[]): string {
  return match (words) {
    ["echo", "hi"] if (flag is string) => "literal arm, no bindings"
    ["echo", str] if (str is string) => "echo string: ${str}"
    ...
```

While you're here: a case asserting the shorthand-binder compile error (`if (str is { policy })`) would pin the behavior the description calls out as a benefit of the fix. It is reachable only through the line you changed, so nothing else guards it.

egonSchiele (commented):
## Overview

Routes a match-arm guard (and the `match(x is pattern)` arm condition) through `lowerExpression`, so an `is` inside one becomes its boolean form instead of surviving raw into codegen and crashing.

## Is this the right way to solve it?

Yes, and it is the simplest correct fix for the two sites you found. `lowerExpression` is the pass every other expression already goes through; a guard is an ordinary boolean expression; the fix is to stop treating it as special. Three lines is the right size for that. The RED-first verification and the note that `make fixtures` produced no changes (nothing in the fixture set used `is` in a guard) are both good evidence.

The `assertNoBindersInBoolIs` behavior falling out for free is a nice bonus — though see item 4 on the message it produces.

## The one place I'd push back

The fix is correct but narrower than the bug. The underlying shape is **"an `Expression` field that nobody remembered to hand to `lowerExpression`"**, and there are at least two more instances of it, both of which still crash after this change. I reproduced both against `main`:

**1. An `is` as a match arm's caseValue, no other pattern arms**

```ts
node main() {
  const x = true
  const y = "hi"
  match (x) {
    (y is string) => print("yes")
    _             => print("no")
  }
  return "done"
}
```

```
Error: Unhandled Agency node type: isExpression
```

`hasPatternArms` (`:423`) is false here — no arm uses a pattern and none has a guard — so we take the early return at `:437`, which calls `lowerMatchCase`. That function is documented as recursing into the body "without lowering the arm itself" (`:822`), so the caseValue goes through untouched.

**2. Same thing when a pattern arm IS present**

```ts
match (x) {
  (y is string) => print("yes")
  [p, q]        => print("arr")
  _             => print("no")
}
```

Same crash, different route: `hasPatternArms` is now true, so this reaches `patternToCondition` → `collectChecks`, whose default (literal) case does `makeBinOp(cloneExpr(source), "==", pattern as Expression, ...)` — pushing the caseValue in raw.

Both are the same bug as the one you fixed, with the same user-facing symptom the PR description objects to: a stack trace instead of a diagnostic. Whether `match (flag) { (y is string) => ... }` is *good* Agency is beside the point — it parses, so it should compile or produce an error message.

### What I'd suggest

Fixing those two sites is cheap and fits this PR. But the more valuable thing is a tripwire, because there will be a fourth site: lowering is opt-in per field, so every new place that holds an `Expression` is a chance to reintroduce this.

You already have the mechanism — the two new unit tests do exactly the right walk. Generalize one of them: lower every `.agency` file under `tests/agency/` (or a small curated corpus of forms) and assert no `isExpression` survives. That is maybe fifteen lines and it covers the whole class, including the sites nobody has thought of yet. The repo already has this pattern as a "walker-completeness tripwire" elsewhere.

If you'd rather keep this PR to the reported bug, that is reasonable — but then the remaining holes deserve an idea doc, since your current one says `:893` "should be checked for the same hole" and the other two never got checked.

## Other issues

### 3. A guard on `_` is silently discarded (pre-existing, but right next door)

`:866` short-circuits on `arm.caseValue === "_"` and never looks at `arm.guard`:

```ts
node main() {
  const y = 0
  let out = "none"
  match (1) {
    2 => out = "two"
    _ if (y > 100) => out = "guarded default"
  }
  return out
}
```

Returns `"guarded default"`. The generated JS contains no trace of the comparison at all — the guard is dropped, not evaluated. This is worse than the crash this PR fixes, because it is silent.

It also matters for *this* PR specifically: `_ if (x is string)` never reaches the lowering you added, so the fix has a blind spot exactly where a reader would assume it applies. Either handle guards on `_` or reject them, but a silently-ignored guard should not survive a PR about guard correctness.

### 4. The binder error message is misleading inside a guard

`assertNoBindersInBoolIs` says:

> shorthand binder in pure-boolean `is` context has nowhere to bind; use `if (x is { ... })` to introduce variables

In a guard, the user *did* write `if (...)` — that is the guard's own syntax:

```ts
["echo", str] if (str is { policy }) => ...
```

They get told to do the thing they just did. Worth a guard-aware variant, or at least rewording the advice to name the construct ("an `if` statement or `while` condition") rather than showing a shape the guard already has.

### 5. One of the two changed lines has no test

`:916` — the `bindings.length === 0` branch that folds the guard in with `&&`. Every guarded arm in `matchGuardIs.agency` binds something (`str`, `n`, `name`), so all of them take the `innerIf` branch at `:918`. The `&&` path with an `is` guard is untested; see the inline comment for the shape.

### 6. The shorthand-binder error has no test

Both the PR description and the idea doc call this out as behavior the fix delivers ("the existing compile error rather than a crash or a silent binding"). Since it is only reachable through the line you changed, a test would pin it — otherwise a future refactor that moves the `lowerExpression` call could turn it back into a silent binding without failing anything.

### 7. Docs

Your idea doc notes that "the interaction of the two is currently undocumented" in `pattern-matching.md`. `is` in a guard now works, which is a new capability from a user's point of view. A couple of lines in the guard-clauses section would land it.

## Verdict

The change itself is right and I would take it as-is. Items 3 and 5 I'd want addressed before merge — a silently dropped guard and an untested changed line are both small fixes. Items 1 and 2 are the same bug you are fixing and I'd rather see them handled here than filed, but that's a scoping call. The tripwire is the suggestion I feel most strongly about, since it is what keeps this from recurring.

egonSchiele (commented):
All seven addressed in 740264c24. I reproduced items 1, 2 and 3 against this branch first — all three behave exactly as you described.

## Items 1 & 2: fixed, but not at those two sites

You are right that the fix was narrower than the bug, and after the shape-check discussion on #676 I did not want to add a third and fourth patch to the same class. The shape is "an `Expression` field nobody handed to `lowerExpression`", and there are three readers of an arm's expression slots — `lowerMatchCase`, `collectChecks`'s literal case, and `foldArms` — any of which can be missed.

So instead of lowering at each reader, an arm's two expression slots (`caseValue` when it is an expression, and `guard`) are lowered **once** at the two match entry points, before any reader sees them. `lowerMatchCase` can keep passing the caseValue through by design, `collectChecks` can keep pushing it into a comparison, and neither can receive an un-lowered arm. The per-site `lowerExpression` calls from the first commit are gone — one funnel, not four.

One subtlety: `node.cases` stays un-lowered for the `matchSource` snapshot, since the exhaustiveness pass wants the arms as written.

## Item 3: honored, not rejected — the typechecker had already decided

I started out planning to reject `_ if (…)`, on the grounds that a guarded catch-all is not a catch-all. Then I checked what the rest of the compiler thinks, and it disagrees with that:

```ts
// lib/typeChecker/matchExhaustiveness.ts:63
function isCatchAll(arm: NormalizedArm): boolean {
  if (arm.guarded) return false;
  ...
```

A guarded arm is already not a catch-all, and the checker reports `match (x) { "a" => …; _ if (…) => … }` over `"a" | "b"` as `AG5002: missing "b"`. So the language already models a guarded `_` as an ordinary conditional arm — the lowerer was the half that disagreed, by dropping the guard entirely. Rejecting would have meant changing the typechecker to match the bug.

A guarded `_` is now a conditional arm whose guard is the whole condition (`_` matches anything), falling through like any other arm. `guardedDefault(7)` returns `"none"` — nothing matched, which is what an unguarded `_` exists to prevent.

## Item 4: message reworded

It now names the constructs instead of showing a shape:

> binders are introduced by an `if` or `while` CONDITION, or by a match arm pattern — a match-arm guard is a pure-boolean context even though it is spelled with `if`

Applied to all four binder kinds, since they all had the same advice.

## Items 5 & 6: both tested

`guardWithNoBindings` uses a literal-only pattern with a guard over an outer variable, which is the `bindings.length === 0` fold — your suggested shape. And `tests/agency/expectedCompileError/guardBinder.agency` pins the shorthand-binder error, reachable only through the lowering this PR adds.

## Item 7: documented

`pattern-matching.md` guard clauses now covers `is` inside a guard, that a guard cannot introduce bindings, and that a guarded `_` stops being a catch-all.

## The tripwire

Already built, in #676 (`lib/lowering/patternGuards.tripwire.test.ts`) — your review and my reply there crossed. Its second invariant is exactly what you describe: lower every `.agency` file under `stdlib/`, `tests/typescriptGenerator/` and `tests/agency/`, and assert no pattern-specific node survives, skipping the deliberate `matchSource` snapshot. It enforces the claim `patternLowering.ts:6` has made since it was written and nothing checked.

I verified it catches this bug rather than assuming: dropping a program with `is` in a guard into the corpus fails it. Once both land, `matchGuardIs.agency` and `guardBinder.agency` are corpus members, so items 1, 2 and the original bug are all covered by it going forward.

## Verification

- `matchGuardIs.agency` 8/8, up from 5; `expectedCompileError` 3/3; `pattern-matching` 20/20.
- Full `lib` suite 9419 passed, only the 5 pre-existing failures (`mcpBridge.contract` x4, `optimize/mutator` x1).
- `make fixtures` produced no changes; `make` and lint clean.

packages/agency-lang/lib/lowering/patternLowering.ts
egonSchiele: Both reproduce, exactly as you wrote them. Rather than lower at a third and fourth reader, an arm's expression slots are now lowered once at the match entry points, before `lowerMatchCase`, `collectChecks` or `foldArms` sees them — so this site no longer lowers anything itself, and neither does the guard site below. `node.cases` stays un-lowered for the `matchSource` snapshot the exhaustiveness pass reads.

packages/agency-lang/lib/lowering/patternLowering.ts
egonSchiele: Reworded to name the construct. All four binder kinds carried the same advice, so all four changed:

> binders are introduced by an `if` or `while` CONDITION, or by a match arm pattern — a match-arm guard is a pure-boolean context even though it is spelled with `if`

The error is now pinned by `tests/agency/expectedCompileError/guardBinder.agency`, which uses your exact example.

packages/agency-lang/tests/agency/pattern-matching/matchGuardIs.agency
egonSchiele: Added `guardWithNoBindings` using your shape — a literal-only pattern arm with a guard over an outer variable, so it takes the `&&` fold rather than the nested if. Covers both directions: guard held and guard failed.

The shorthand-binder case is in `tests/agency/expectedCompileError/guardBinder.agency` rather than here, since it is a compile error and not a runtime case.

PR #682, opened 2026-07-25
Perf tests (2/3): parse, typecheck, fmt, per-stage compile

egonSchiele:
PR 2 of the CI performance-tests plan (after #675). Extends the informational scaling suite to the remaining primitive commands. All measured on **in-memory strings**, which never touch the file-based parse cache — so none of the cache-neutralization machinery is needed here (that's PR 3's incremental-build test).

## What's added

- **`parse.perf.test.ts`** — parse isolated first, since it dominates every downstream measurement.
- **`typecheck.perf.test.ts`** — `buildCompilationUnit` + `typeCheck`.
- **`fmt.perf.test.ts`** — `formatSource` (the sync, pure core).
- **`compile.perf.test.ts`** — **per pipeline stage** (parse → bind → preprocess → generate → full), not just end-to-end, so a quadratic in one stage can't hide inside a single "compile" number (the AL0002 lesson applied to the pipeline). Each stage was verified re-runnable: `buildCompilationUnit` doesn't mutate the program, and preprocess/build produce byte-identical output on a repeated call.

The perf CI job (from #675) picks these up automatically via `test:perf`.

## Baselines — all linear

```
parse:manyFunctions       1.22      typecheck:manyFunctions   1.07
compile:parse             1.04      compile:bind              1.01
compile:preprocess        1.11      compile:generate          0.95
compile:full              0.98      fmt:manyFunctions          1.08
```

(Normalized growth: 1.0 = linear.) Full `test:perf` ~66 s locally.

## Two real scaling findings (surfaced while choosing fixtures)

Neither is a regression this PR introduces, but both are worth flagging:

1. **Typecheck is ~O(n²) in union *width*.** `wideUnion`: 500 arms → 42 ms, 4000 → 2694 ms (64× for 8× input). A fixture that already reads ~8 can't detect *further* regression, so typecheck uses `manyFunctions` (a clean linear baseline) instead. The union quadratic looks worth its own issue — happy to file one.
2. **Parse degrades sharply past ~4000 functions** — an 18 s cliff at 8000, reading as super-linear from GC/cache pressure rather than pure algorithmic cost. Per the design review's "sane memory regime" guidance, the parse fixture is kept at 250..2000 (where it's ~1.2). The parser's large-file scaling may be worth a separate look.

## Next
PR 3 — LSP hot paths + doc/bundle/cli-cold-start + the file-based incremental-build test. Then the calibration → flip-to-gating milestone.

### DIFF
--- a/packages/agency-lang/lib/perf/compile.perf.test.ts
+++ b/packages/agency-lang/lib/perf/compile.perf.test.ts
@@ -1,0 +1,67 @@
+import { describe, it, expect } from "vitest";
+import { parseAgency } from "../parser.js";
+import { buildCompilationUnit } from "../compilationUnit.js";
+import { TypescriptPreprocessor } from "../preprocessors/typescriptPreprocessor.js";
+import { TypeScriptBuilder } from "../backends/typescriptBuilder.js";
+import { printTs } from "../ir/prettyPrint.js";
+import { generateTypeScript } from "../backends/typescriptGenerator.js";
+import type { AgencyProgram } from "../types.js";
+import { manyFunctions } from "./fixtures.js";
+import { growthFactor, expectPerf, GROWTH_BOUND } from "./harness.js";
+
+// Compile is measured PER STAGE on in-memory strings, not just end to end: a
+// quadratic in preprocess or codegen would hide inside a single "compile" number
+// (the lesson of the AL0002 bug, applied to the pipeline). Strings never touch
+// the file-based parse cache, so no cache neutralization is needed here.
+//
+// Every stage was verified re-runnable: buildCompilationUnit does not mutate the
+// program, and preprocess/build produce identical output on a repeated call.
+
+const SMALL = 100;
+const LARGE = 800;
+
+function parse(n: number): AgencyProgram {
+  const p = parseAgency(manyFunctions(n), {}, false);
+  if (!p.success) throw new Error("fixture did not parse");
+  return p.result;
+}
+
+// Each stage builder runs the untimed prefix and returns a closure over just the
+// stage under test.
+const stages: Record<string, (n: number) => () => unknown> = {
+  parse: (n) => {
+    const src = manyFunctions(n);
+    return () => parseAgency(src, {}, false);
+  },
+  bind: (n) => {
+    const program = parse(n);
+    return () => buildCompilationUnit(program);
+  },
+  preprocess: (n) => {
+    const program = parse(n);
+    const unit = buildCompilationUnit(program);
+    return () => new TypescriptPreprocessor(program, {}, unit).preprocess();
+  },
+  generate: (n) => {
+    const program = parse(n);
+    const unit = buildCompilationUnit(program);
+    const preprocessed = new TypescriptPreprocessor(program, {}, unit).preprocess();
+    return () => printTs(new TypeScriptBuilder({}, unit, "m").build(preprocessed));
+  },
+  full: (n) => {
+    const program = parse(n);
+    return () => generateTypeScript(program, {}, undefined, "m");
+  },
+};
+
+describe("compile per-stage scaling", () => {
+  it("each stage scales linearly in file size", () => {
+    // work-happened: the full pipeline produces real TypeScript for the fixture.
+    const ts = generateTypeScript(parse(LARGE), {}, undefined, "m");
+    expect(ts).toContain("fn0");
+
+    for (const [name, build] of Object.entries(stages)) {
+      expectPerf(`compile:${name}`, growthFactor(build, SMALL, LARGE), GROWTH_BOUND);
+    }
+  });
+});
--- a/packages/agency-lang/lib/perf/compile.perf.test.ts
+++ b/packages/agency-lang/lib/perf/compile.perf.test.ts
@@ -11,11 +11,22 @@ import { growthFactor, expectPerf, GROWTH_BOUND } from "./harness.js";
 
 // Compile is measured PER STAGE on in-memory strings, not just end to end: a
 // quadratic in preprocess or codegen would hide inside a single "compile" number
-// (the lesson of the AL0002 bug, applied to the pipeline). Strings never touch
-// the file-based parse cache, so no cache neutralization is needed here.
+// (the AL0002 lesson applied to the pipeline). Strings never touch the file-based
+// parse cache, so no cache neutralization is needed.
 //
-// Every stage was verified re-runnable: buildCompilationUnit does not mutate the
-// program, and preprocess/build produce identical output on a repeated call.
+// Re-runnability matters because growthFactor runs each closure 9x on the same
+// setup. `parse`, `bind`, and `generate` are genuinely re-runnable: parse yields
+// a fresh program each call; buildCompilationUnit does not mutate the program;
+// TypeScriptBuilder.build does not mutate the preprocessed program (all verified).
+// But `preprocess` reassigns program.nodes in place (and so does the internal
+// preprocess in `postParse`/generateTypeScript), so those two clone the program
+// per call — otherwise calls 2..9 would preprocess an already-preprocessed
+// program and measure a warm re-run, not cold work. The clone is ~1% of the
+// stage cost here, so it does not mask a regression.
+//
+// LARGE is kept modest (the full pipeline at 800 functions already emits ~5MB of
+// TS per call). Revisit during calibration if any stage's baseline sits well
+// above 1.0 — that would mean N is too small to separate its regressions.
 
 const SMALL = 100;
 const LARGE = 800;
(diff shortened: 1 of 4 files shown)


packages/agency-lang/lib/perf/compile.perf.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    const program = parse(n);
+    return () => buildCompilationUnit(program);
+  },
+  preprocess: (n) => {
+    const program = parse(n);
+    const unit = buildCompilationUnit(program);
+    return () => new TypescriptPreprocessor(program, {}, unit).preprocess();
+  },
+  generate: (n) => {
+    const program = parse(n);
```
egonSchiele, line 59: This stage (and `full` below) is where the re-runnability claim breaks. `build(n)` creates `program` once, then the returned closure calls `new TypescriptPreprocessor(program, {}, unit).preprocess()` — and `preprocess()` reassigns `this.program.nodes` in place (typescriptPreprocessor.ts:190/272/323/333). `growthFactor` invokes this closure 9× on that same `program`, so:

- call 1 (warmup): preprocesses the fresh program, mutating `program.nodes` into preprocessed form;
- calls 2–9 (including all 7 timed rounds): preprocess an **already-preprocessed** program.

So the timed number is the cost of re-preprocessing preprocessed input, not the cold preprocess cost the stage is meant to guard. It's benign for `manyFunctions` only because that fixture has no imports/guards/tags, so the passes happen to be ~idempotent — but that's a property of the fixture, asserted by a comment, not enforced. A future fixture with imports (import-path rewriting is exactly one of the mutating passes) would silently corrupt this measurement, and a regression that made *cold* preprocess quadratic while leaving warm re-runs cheap wouldn't be caught.

Also worth noting: `unit` is built from the pre-preprocess program and holds references to the original node objects; after call 1 reassigns `program.nodes`, the program and `unit` disagree, so calls 2–9 are measuring a semantically muddled state.

Suggested fix: for mutating stages, give each timed call a fresh input — e.g. `structuredClone(program)` inside the closure (cheaper than re-parsing, and the stage cost dominates at these sizes), timing clone+stage. Or, at minimum, add a real assertion that preprocessing twice yields identical `program.nodes` for the fixture, so the invariant the measurement depends on is enforced rather than commented.

packages/agency-lang/lib/perf/compile.perf.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    const preprocessed = new TypescriptPreprocessor(program, {}, unit).preprocess();
+    return () => printTs(new TypeScriptBuilder({}, unit, "m").build(preprocessed));
+  },
+  full: (n) => {
+    const program = parse(n);
+    return () => generateTypeScript(program, {}, undefined, "m");
+  },
+};
+
+describe("compile per-stage scaling", () => {
```
egonSchiele, line 72: `full` has the same issue as `preprocess`: `generateTypeScript(program, ...)` runs `buildCompilationUnit` + `preprocess` + `build` internally, and `preprocess` mutates `program.nodes` — so calls 2–9 run the full pipeline over a program the first call already preprocessed. Same fix applies (fresh/cloned program per call).

Separately, a naming nit: `full` takes an already-parsed `program`, so it measures the post-parse pipeline (bind→preprocess→generate), not parse+everything. Given there's a dedicated `parse` stage that's a fair split, but `compile:full` reads as "the whole compile" when it excludes the parse cost — worth a word in the comment or a name like `post-parse`.

packages/agency-lang/lib/perf/compile.perf.test.ts
```diff
@@ -0,0 +1,67 @@
...
+// program, and preprocess/build produce identical output on a repeated call.
+
+const SMALL = 100;
+const LARGE = 800;
+
+function parse(n: number): AgencyProgram {
+  const p = parseAgency(manyFunctions(n), {}, false);
+  if (!p.success) throw new Error("fixture did not parse");
+  return p.result;
+}
```
egonSchiele, line 38: `LARGE = 800` is much smaller than the other tests (parse 2000, typecheck 4000, fmt 1200). That's presumably to keep the full-pipeline runtime down, but 800 functions is a modest input for catching a *stage* that's mildly super-linear — a codegen pass with a small quadratic coefficient could read under the bound at 800 and only blow up higher. Not blocking (the informational data will show if any stage's baseline sits suspiciously above 1.0), but worth a comment on why 800, and maybe revisiting during calibration.

packages/agency-lang/lib/perf/typecheck.perf.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    if (!parsed.success) throw new Error("fixture did not parse");
+    const program = parsed.result;
+    expect(typeCheck(program, {}, buildCompilationUnit(program)).scopes.length).toBeGreaterThan(0);
+
+    const build = (n: number) => {
+      const p = parseAgency(manyFunctions(n), {}, false);
+      if (!p.success) throw new Error("fixture did not parse");
+      const prog = p.result;
+      return () => typeCheck(prog, {}, buildCompilationUnit(prog));
+    };
```
egonSchiele, line 32: This closure reuses `prog` across all 9 calls, and `typeCheck` also mutates its input — it reassigns `program.nodes = desugarGuardsInBody(program.nodes)` (typeChecker/index.ts:111). It's safe *here* only because `manyFunctions` has no `guard` constructs, so the desugar is a no-op and re-runs are equivalent. Same latent fragility as the compile stages: the measurement's validity rests on a fixture property that's only documented in prose. A one-line assertion (typecheck twice, `program.nodes` unchanged) would pin it, or note explicitly that swapping in a guard-bearing fixture here would break the reuse.

egonSchiele (commented):
Follows PR #675's shape cleanly, and the two scaling findings surfaced in the description (typecheck ~O(n²) in union width, the parse cliff past ~4000) are genuinely valuable — both worth their own issues. The per-stage decomposition is the right call.

One finding I think is real and undercuts what a couple of these stages actually measure, plus two smaller notes. I verified everything against the code.

**The re-runnability claim doesn't hold for the `preprocess` and `full` stages.** The PR says "preprocess/build produce identical output on a repeated call" — but I checked, and `TypescriptPreprocessor.preprocess()` reassigns `this.program.nodes` in place four times (typescriptPreprocessor.ts:190, 272, 323, 333: `collectTags`, the lowering `result`, again `collectTags`, `desugarGuardsInBody`). Because `growthFactor` runs each closure 9× (2 warmup + 7 rounds) on the *same* `program` object, every timed call after the first preprocesses an **already-preprocessed** program. So those stages measure warm re-runs, not cold work — and "identical output on a repeated call" doesn't establish otherwise, because a perf test cares about *timing*, not output equality (an idempotent-in-output pass can still do different work the second time, and a regression in the cold path can hide behind a cheap re-run). Details inline.

What I verified is safe: `buildCompilationUnit` genuinely doesn't mutate the program (it reads `program.nodes` and builds a fresh `unit`), so `parse` and `bind` are re-runnable by construction. The `generate` stage reuses the already-preprocessed value and only re-runs `.build()`, so it's likely fine. It's specifically `preprocess` and `full` (which runs preprocess internally via `generateTypeScript`) that reuse the mutated `program`.

egonSchiele:
Good catch — you're right, and it was a real hole in how I verified the stages. Fixed in `8cce5a7`.

**The re-runnability flaw (preprocess / full).** I'd verified *output* equality, which doesn't establish *work* equality — and as you noted, `preprocess()` reassigns `program.nodes` in place (typescriptPreprocessor.ts:190/272/323/333), so calls 2..9 preprocessed an already-preprocessed program and measured a warm re-run. I confirmed the mutation and the consequences you described. The fix: the mutating stages (`compile:preprocess`, `compile:postParse`, and `typecheck`) now `structuredClone` the program per timed call. I measured the clone at ~1% of the stage cost at these sizes, so it doesn't mask a regression (and it's ~15× cheaper than re-parsing). What I verified stays pure and left alone: `buildCompilationUnit` doesn't mutate the program, and `TypeScriptBuilder.build` doesn't mutate the preprocessed program — so `parse`, `bind`, and `generate` are genuinely re-runnable as-is.

**typecheck** had the same latent issue (`typeCheck` reassigns `program.nodes` via `desugarGuardsInBody`) — safe only because `manyFunctions` has no guards. Cloned there too, so it no longer rests on a fixture property.

**Naming nit.** Renamed `compile:full` → `compile:postParse` and documented that it excludes the parse stage (which has its own measurement).

**LARGE = 800.** Added a comment on why it's smaller (the full pipeline already emits ~5MB of TS/call at 800) and a note to revisit at calibration if any stage's baseline sits above ~1.0.

All baselines stay linear after the fix (~0.97–1.09), tsc + eslint clean.

Filed the two findings as you suggested: #683 (typecheck O(n²) in union width) and #684 (parser super-linear on large files).

Issue #685, opened 2026-07-25
Perf suite: add integration-level tests (incremental build, CLI cold-start, bundle)


egonSchiele:
The library-level perf suite (parse/typecheck/fmt/compile/lint/lsp/doc) landed across #675, #682, and the LSP+doc PR. Three commands from the original plan were deferred because, unlike the others, they can't be measured as a pure in-memory library call — each needs real disk/build state, which makes them integration-style tests with higher flake risk. Tracking them here as a focused follow-up.

- **Incremental build (cold vs warm)** — the highest-value of the three ("perf over time bites hardest as the codebase grows"). Needs a temp project dir, a real `BuildSession` + `ManifestTracker`, output `.js` written to disk, and the `.agency-build/` manifest. The test would compile a project cold (no manifest), then warm (manifest hit), and assert warm << cold — with an assertion (via parseCache stats) that the cold leg actually missed, so a leaked warm cache from another test can't false-pass. Own file, own setup/teardown.
- **CLI cold-start** — time `node dist/scripts/agency.js --version` with a loose ceiling. Needs `dist/` (so the perf CI job would need `make`) and is inherently the noisiest measurement (subprocess wall time). Low value; include only if the `make` cost is judged worthwhile.
- **bundle** — `createBundle` needs an execution *trace* file (from actually running a program), not just source, so it isn't a clean scaling target. Lowest priority.

None of these block the calibration → flip-to-gating milestone, which only needs the library-level baselines already collected.

PR #686, opened 2026-07-25
Perf tests (3/3): LSP hot paths and doc generation

egonSchiele:
PR 3 of the CI performance-tests plan (after #675, #682). Adds the LSP hot paths and doc generation to the informational scaling suite.

## What's added

- **`lsp.perf.test.ts`**
  - **`runDiagnostics`** — the debounced-keystroke path (parse + resolve + typecheck + lint + semantic index in one call), scaled over document size. This is the LSP measurement that matters most — it's what an editor runs on every change.
  - **`getCodeActions`** (SourceFixAll) — scaled over fixable-finding count.
  - A **fresh empty `SymbolTable` per call** keeps each measurement independent (no cross-call caching), and both re-parse the document text each call, so they're re-runnable — the PR #682 lesson applied up front.
- **`doc.perf.test.ts`** — `generateDoc` scaled over file size (re-reads input, overwrites output each call).

## Baselines — all linear

```
lsp:runDiagnostics  1.16      lsp:codeActions  0.81      doc:generateDoc  1.06
```

Full `test:perf` is now **15 measurements** across parse / typecheck / fmt / compile (per-stage) / lint (per-rule) / lsp / doc, ~92 s, all linear. tsc + eslint clean; the default unit suite still excludes the perf files.

## Deferred (→ #685)

Three commands from the original plan are deferred: **incremental-build**, **CLI cold-start**, and **bundle**. Unlike everything else in the suite, none can be measured as a pure in-memory library call — each needs real disk/build state (the `.agency-build/` manifest, a built `dist/`, or an execution *trace* file), which makes them integration-style tests with higher flake risk (the review flagged incremental-build as the "operationally hardest"). The library-level suite now covers every **algorithmic** hot path, which is what the calibrate-and-flip milestone needs — so the integration tests are a clean, separately-scoped follow-up rather than a rushed addition here.

## What's left in the effort
Just the **operational milestone**: gather ~30 informational runs from the uploaded `perf-results.jsonl` artifacts, set `GROWTH_BOUND` and the smoke ceilings from the observed p95, then flip to gating (`PERF_ENFORCE=1`, drop `continue-on-error`). No more code PRs for the core suite.

### DIFF
--- a/packages/agency-lang/lib/perf/doc.perf.test.ts
+++ b/packages/agency-lang/lib/perf/doc.perf.test.ts
@@ -1,0 +1,33 @@
+import { describe, it, expect, afterAll } from "vitest";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import { generateDoc } from "../cli/doc.js";
+import { manyFunctions } from "./fixtures.js";
+import { growthFactor, expectPerf, GROWTH_BOUND } from "./harness.js";
+
+const SMALL = 100;
+const LARGE = 800;
+
+// generateDoc reads the input file and writes markdown to an output dir each
+// call, so it is re-runnable (it re-reads and overwrites). File I/O is part of
+// the doc command's real cost.
+const tmp = fs.mkdtempSync(path.join(os.tmpdir(), "perf-doc-"));
+afterAll(() => fs.rmSync(tmp, { recursive: true, force: true }));
+
+function docBuild(n: number): () => void {
+  const inFile = path.join(tmp, `in-${n}.agency`);
+  fs.writeFileSync(inFile, manyFunctions(n), "utf-8");
+  const outDir = path.join(tmp, `out-${n}`);
+  fs.mkdirSync(outDir, { recursive: true });
+  return () => generateDoc({}, inFile, outDir);
+}
+
+describe("doc scaling", () => {
+  it("scales linearly in file size", () => {
+    docBuild(LARGE)(); // work-happened: produces output without throwing
+    expect(fs.readdirSync(path.join(tmp, `out-${LARGE}`)).length).toBeGreaterThan(0);
+
+    expectPerf("doc:generateDoc", growthFactor(docBuild, SMALL, LARGE), GROWTH_BOUND);
+  });
+});
--- a/packages/agency-lang/lib/perf/doc.perf.test.ts
+++ b/packages/agency-lang/lib/perf/doc.perf.test.ts
@@ -11,7 +11,9 @@ const LARGE = 800;
 
 // generateDoc reads the input file and writes markdown to an output dir each
 // call, so it is re-runnable (it re-reads and overwrites). File I/O is part of
-// the doc command's real cost.
+// the doc command's real cost — this is the one test here whose timing isn't
+// pure CPU, so if any line proves flaky during the informational period, the
+// disk write is the likely reason.
 const tmp = fs.mkdtempSync(path.join(os.tmpdir(), "perf-doc-"));
 afterAll(() => fs.rmSync(tmp, { recursive: true, force: true }));
 
--- a/packages/agency-lang/lib/perf/lsp.perf.test.ts
+++ b/packages/agency-lang/lib/perf/lsp.perf.test.ts
@@ -22,6 +22,10 @@ describe("lsp scaling", () => {
     // runDiagnostics is the debounced-keystroke hot path: parse + resolve +
     // typecheck + lint + semantic index in one call. (There is no incremental
     // fast-path today, so a "re-diagnose after an edit" case would equal this.)
+    // The empty SymbolTable is a deliberate simplification — a real editor
+    // passes a populated one, so import resolution is skipped here; immaterial
+    // for manyFunctions (no imports) and for a scaling ratio, but this is not
+    // the full keystroke cost.
     const doc = docOf(manyFunctions(LARGE, { docstrings: false }));
     expect(runDiagnostics(doc, "/t.agency", {}, new SymbolTable()).diagnostics.length).toBeGreaterThan(0);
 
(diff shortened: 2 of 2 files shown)


packages/agency-lang/lib/perf/lsp.perf.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  it("runDiagnostics scales linearly in document size", () => {
+    // runDiagnostics is the debounced-keystroke hot path: parse + resolve +
+    // typecheck + lint + semantic index in one call. (There is no incremental
+    // fast-path today, so a "re-diagnose after an edit" case would equal this.)
+    const doc = docOf(manyFunctions(LARGE, { docstrings: false }));
+    expect(runDiagnostics(doc, "/t.agency", {}, new SymbolTable()).diagnostics.length).toBeGreaterThan(0);
+
+    const build = (n: number) => {
+      const d = docOf(manyFunctions(n, { docstrings: false }));
+      return () => runDiagnostics(d, "/t.agency", {}, new SymbolTable());
```
egonSchiele, line 34: Re-runnable and correct — just noting a fidelity caveat: a real editor calls `runDiagnostics` with a *populated* `SymbolTable` (resolved workspace/imports), whereas this passes an empty one, so import-resolution work is effectively skipped. For `manyFunctions` (no imports) that's immaterial, and for a scaling ratio it doesn't matter as long as the work stays proportional to size — so this is fine. Worth a word in the comment that the empty table is a deliberate simplification, so nobody later reads this as "the full keystroke cost."

packages/agency-lang/lib/perf/lsp.perf.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      textDocument: { uri: "file:///t.agency" },
+      range: { start: { line: 0, character: 0 }, end: { line: 0, character: 0 } },
+      context: { diagnostics: [], only: [CodeActionKind.SourceFixAll] },
+    };
+    const doc = docOf(manyUnusedImports(LARGE));
+    expect(getCodeActions(params, doc, new SymbolTable()).length).toBeGreaterThan(0);
+
+    const build = (n: number) => {
+      const d = docOf(manyUnusedImports(n));
+      return () => getCodeActions(params, d, new SymbolTable());
```
egonSchiele: Valid measurement, but the label "scales linearly in finding count" oversells what dominates. With `only: [SourceFixAll]` and an empty `context.diagnostics`, the timed cost is `computeLint` (parse + `runLinter` over *all* rules) + `unusedImportsBatchEdits` — i.e. mostly parse+full-lint, with the finding-count-dependent batch work as a smaller term. The `symbolTable` argument isn't even used on this path (it's only consumed in the `context.diagnostics` loop, which is empty here). None of that is wrong — it's a realistic on-save measurement and it re-parses each call so it's re-runnable — but the comment/label imply the fix-all-over-N-findings path is the thing measured, when parse+lint is. (Also: the return is an aggregate action, so `.length > 0` is the only assertable work-happened check here — correctly done; you can't assert `=== n` the way the lint tests do.)

packages/agency-lang/lib/perf/doc.perf.test.ts
```diff
@@ -0,0 +1,33 @@
...
+// the doc command's real cost.
+const tmp = fs.mkdtempSync(path.join(os.tmpdir(), "perf-doc-"));
+afterAll(() => fs.rmSync(tmp, { recursive: true, force: true }));
+
+function docBuild(n: number): () => void {
+  const inFile = path.join(tmp, `in-${n}.agency`);
+  fs.writeFileSync(inFile, manyFunctions(n), "utf-8");
+  const outDir = path.join(tmp, `out-${n}`);
+  fs.mkdirSync(outDir, { recursive: true });
+  return () => generateDoc({}, inFile, outDir);
```
egonSchiele, line 25: The timed closure includes a file read *and* a markdown write to disk (generateDoc reads `inFile` and writes into `outDir` each call). That's legitimately part of doc's real cost, but disk I/O is a noisier signal than the pure-CPU tests elsewhere in the suite — on a contended runner the write time can swing more than the compute. Under the generous `GROWTH_BOUND` and with interleaving it should be fine, but it's the one test here whose timing isn't pure CPU, so if `doc:generateDoc` turns out to be the flakiest line during the informational period, this is why.

egonSchiele (commented):
This is the cleanest of the three, and it lands the re-runnability lesson from #682 correctly — I verified it rather than taking the comment's word:

- **`runDiagnostics`** and **`getCodeActions`** both re-parse the document text internally each call (`getCodeActions` → `computeLint` → `parseAgency(source, {}, false)`; the string path, no `parseAgencyFileCached`), and each gets a fresh `SymbolTable`. Nothing reuses a mutated program object, so the #682 preprocess-mutation trap doesn't apply here.
- **`generateDoc`** reads the file and parses via `parse(readFile(...))` (the string parser, not the file cache), so it re-reads and re-parses cold every call — the doc test's "re-runnable" claim is accurate, and there's no cache to neutralize.

Also worth calling out: dropping the "re-diagnose after an edit" case and documenting *why* (no incremental fast-path today, so it would just equal `runDiagnostics`) is exactly the right resolution of the note I raised on #675 — a test with no distinct failure condition is better removed than left as decoration. And deferring incremental-build / cold-start / bundle to #685 as integration-style tests is well-justified.

No blocking issues. Three minor notes inline, all about what the measurements represent rather than whether they're valid.

egonSchiele:
Thanks — addressed in `4c8c02d`. All three of your notes were about *what the measurements represent*, and Copilot's was the substantive one, so I fixed the measurement rather than just the label.

**getCodeActions (Copilot + your note 2).** You're both right: with no cached lint it hit the fallback and re-measured parse+lint, duplicating the lint tests, and `symbolTable` wasn't even on that path. It now precomputes the lint via `runDiagnostics` and passes it as `cachedLint` exactly the way the server does (`server.ts:326-335`), so the timed closure isolates the fix-all *packaging* (batch edits → text edits) — the work that actually scales with finding count and matches the on-save hot path. Renamed the test to "packages fix-all edits in time linear in finding count." I worried the isolated path might be too fast to measure stably, but it's steady at ~1.1 across three runs.

**runDiagnostics empty SymbolTable (your note 1).** Added a comment that it's a deliberate simplification — import resolution is skipped, so this is a scaling proxy, not the full keystroke cost.

**doc disk I/O (your note 3).** Added a comment flagging it as the one non-pure-CPU test and the likely suspect if `doc:generateDoc` is the flakiest line during the informational period.

tsc + eslint clean.

Issue #688, opened 2026-07-25
std::system: env() should raise an interrupt, like setEnv does


egonSchiele:
## Summary

`env` in `std::system` reads an environment variable and raises no interrupt. Its neighbour `setEnv` does raise one. Reading secrets is at least as sensitive as setting them, and every other ambient operation in the stdlib is gated.

```ts
// stdlib/system.agency:101
export def env(name: string): string | null {
  return _env(name)          // no interrupt
}

export def setEnv(name: string, value: string): Result {
  return interrupt std::setEnv(...)   // interrupt
}
```

## Why it matters more now

Compile-time splices (#679) run generator code during compilation. Anything a generator can read can be baked into the emitted JavaScript as a string literal, and that artifact gets committed and shipped:

```ts
import { Code, fill } from "std::agency"
import { env } from "std::system"

export def makeClient(): Code {
  return fill([| def apiKey(): string { return #k: string } |],
              { k: env("ANTHROPIC_API_KEY") })
}
```

The splices PR works around this by giving the generator's child process an allowlisted environment instead of the parent's. That closes the confidentiality hole for splices specifically, and it is a workaround for a stdlib gap rather than a fix.

Making `env` raise removes the workaround entirely: the effect check refuses such a generator statically, and the unhandled-interrupt backstop stops it at runtime, because compilation installs no handlers.

## The same class

Other interrupt-free ambient reads in `stdlib/system.agency`, worth auditing at the same time:

- `args()` — command-line arguments
- `cwd()` — working directory
- `isTTY()` — terminal detection
- `readStdin()` — consumes the caller's stdin, and would block a build for the full wall-clock limit

`cwd` and `isTTY` are low-sensitivity and may be fine to leave. `args` can carry secrets. `readStdin` is a denial-of-service shape rather than a confidentiality one.

## Cost to weigh

Every existing program that reads an environment variable would start raising an interrupt, so users would see a prompt or need a handler. That is the objection that ruled out making `llm()` raise. `env` is far rarer than `llm`, so the blast radius should be much smaller, but it should be measured across the examples and stdlib before committing.

## Follow-up

Once this lands, delete `CHILD_ENV_ALLOWED` and `childEnv()` from `lib/compiler/splice/runGenerator.ts` and the test that pins them.

egonSchiele:
## Tried it. It cannot work as written, and here is why.

Attempted in #694 and reverted. The change itself is two lines, and it breaks the agent.

[...]

Issue #681, opened 2026-07-25
One-line def parses as a file but fails inside a code literal


egonSchiele:
## Summary

A one-line function definition parses fine as a file but fails inside a code literal. The same text is legal in one place and illegal in the other.

## Reproduction

This is valid Agency and compiles:

```ts
def greet(): string { return "hi" }
```

The same text inside `[| ... |]` fails:

```ts
import { Code } from "std::agency"

export def makeGreet(): Code {
  return [| def greet(): string { return "hi" } |]
}
```

```
Failed to parse Agency program: Line 6, col 12: code literal body: Line 4, col 17: expected one of " \t", ":", or "="
```

Spread across lines it works:

```ts
return [|
  def greet(): string {
    return "hi"
  }
|]
```

A one-line **expression** body is also fine, so this is specific to definitions:

```ts
return [| print("step") |]   // parses
```

## Why it looks like a parser bug rather than a syntax rule

`parseCodeLiteralBody` infers the fragment kind smallest-first: a lone expression, else a statement list, else a whole program, with each attempt required to consume the entire body (`docs/dev/template-agency.md`, "Code literals"). A one-line `def` should fail the expression and statement attempts and succeed on the program attempt, since that identical text parses as a program when it is a file.

The reported error also points at column 17, around the parameter list, and says it expected `":"` or `"="` — which reads like an earlier attempt's failure leaking out rather than the program attempt's result.

## Impact

Mostly on documentation and on anyone writing generators.

The compile-time splices spec (`docs/superpowers/specs/2026-07-24-compile-time-splices-design.md`) uses a one-line `def` inside a literal in its main worked example:

```ts
fill([| def #fnName(): string { return #greeting: string } |], …)
```

That example does not run. Someone following it gets a parse error with no obvious cause, since the same code works when they paste it into a file to check.

Not a blocker: writing literal bodies multi-line is a fine workaround, and #679 does exactly that.

## Suggested fix

Check whether the program-kind attempt in `parseCodeLiteralBody` is reached at all for this input, and whether the reported error is the program attempt's or an earlier attempt's leaking through. The nested-parse machinery runs under a rightmost save/restore, so a leaked failure is a plausible mechanism.

Whatever the fix, `docs/site/guide/templates.md` and the splices spec should get their examples corrected either way — the guide's own literal examples are multi-line, which is likely why this went unnoticed.

egonSchiele:
Fixed. The issue's exact snippet now parses, compiles and runs:

[...]

PR #694, opened 2026-07-25
One-line defs parse inside code literals (#681)

egonSchiele:
A one-line function definition parsed as a file but failed inside a code template:

```ts
def greet(): string { return "hi" }          // parsed
[| def greet(): string { return "hi" } |]    // did not
```

## The cause is whitespace, not definitions

The error message sends you the wrong way. It points into the middle of the return type and says it expected `" \t"`, `":"`, or `"="`, which looks like a problem with function syntax.

`parseCodeLiteralBody` tries three grammars, smallest first: one expression, then a statement list, then a whole program. The first two get the **trimmed** body. The program attempt got the raw one.

A template written on one line starts with a space:

```
[| def greet(): string { } |]
   ^ the body handed to the program grammar begins here
```

That grammar ends in `eof` and does not skip leading spaces. A leading **newline** is fine, which is exactly why multi-line bodies worked and single-line ones did not.

Definitions were simply the only thing that needed the program grammar at all. An expression or a statement list parsed from the trimmed body and never reached the failing attempt, which is why `[| print("x") |]` and `[| const x = 1 |]` always worked and `[| def f(): string { } |]` did not.

The fix is one argument: the program attempt now gets the same trimmed body and base position as the other two.

## Tests

Four cases pinned in `codeLiteral.test.ts`, including one asserting that a one-line body and the same text spread across lines produce the same kind and the same nodes. That property was silently untrue.

## What is not in here

This branch also tried #688, making `env()` ask permission before reading an environment variable. **It is reverted**, because CI found it cannot work as specified.

Interrupts checkpoint the state stack, which only exists inside a running node. Environment reading is largely startup configuration that happens before any node exists:

```ts
export static const POLICY_DIR = resolveAgentHome(env("AGENCY_AGENT_HOME"))
```

That constant is the root of a chain the whole agent depends on, and a static initializer runs at process startup, so an interrupt there is impossible by construction. Four agent test files failed on it. Two stdlib modules have the same shape.

The full write-up, including what did work and three options, is on #688. One finding worth pulling out: `AG7003` refuses an `interrupt` written directly in a static initializer but does not follow calls, so it missed this. A transitive version would have made it a compile error instead of a CI failure.

## Verification

Unit suite: **9572 passing, 42 skipped, no failures**. `make` clean, `tsc --noEmit` clean, `lint:structure` clean. The four agent tests that CI flagged pass again after the revert.


packages/agency-lang/stdlib/system.agency
```diff
@@ -101,7 +101,12 @@ export def cwd(): string {
 export def env(name: string): string | null {
   """
   Read an environment variable. Returns null if the variable is not set.
+
+  @param name - The environment variable name
   """
+  return interrupt std::env("Are you sure you want to read this environment variable?", {
```
egonSchiele: Consider marking this `export idempotent def env(...)`.

Reading an environment variable is idempotent by definition, and the codebase already uses the marker for exactly this class of thing: `compile`, `typecheck`, `getEffects`, `describe`, `parseAST`, and `typecheckFile` in `stdlib/agency.agency` are all `export idempotent def`. Per the four-tier tool-loop work, the marker changes how a failure reaches a tool-calling model, so a rejected or errored `env` read becomes safely re-callable rather than sitting in the neutral tier.

`setEnv` and `openUrl` are unmarked, but both mutate, so they are not the precedent to follow here.

packages/agency-lang/tests/agency/env-interrupt.agency
```diff
@@ -0,0 +1,20 @@
...
+// does. A secret read without asking can be written into a log, a prompt,
+// or generated code, and none of that is visible to the person whose
+// secret it is.
+node main(): string {
+  let effect = "none"
+  let readSucceeded = false
+
+  handle {
+    const value = env("PATH")
+    readSucceeded = value != null
```
egonSchiele: This asserts the effect name and that *something* non-null comes back, but not that the returned value is still the environment variable's value.

That matters more than usual because of the idiom this function uses. The body is `return interrupt std::env(...)` followed by `return _env(name)`, and a reader looking only at the source cannot tell whether an approved read resumes into the second `return` or whether the interrupt's own result is what comes back. The PR's headline claim is that the signature does not change, and this fixture would pass either way.

A round trip pins it: `setEnv("AGENCY_ENV_TEST", "hello")`, then `env("AGENCY_ENV_TEST")`, then assert `"hello"`. Both interrupts get approved by the same handler.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3073,8 +3073,14 @@ export function parseCodeLiteralBody(body: string, base?: Position): ParsedLiter
...
       },
-      sentineled,
-      { basePosition: base },
+      // `trimmed`, like the two attempts above, not the raw body. The
+      // program grammar ends in `eof` and does not skip leading spaces, so
+      // a one-line body such as `[| def f(): string { } |]` handed it
+      // " def f(): string { } " and failed on the leading space alone. A
+      // leading NEWLINE was fine, which is why multi-line bodies worked
+      // and single-line ones did not.
+      trimmed,
```
egonSchiele, line 3082: The fix is right and the comment explains it well. One thing the tests do not cover: this also changes the *positions* of nodes inside a `program`-kind literal, because the input is now the trimmed body and the base is advanced past the stripped prefix. Both are self-consistent, and it matches what the expr and statements attempts already do, so I expect it is correct.

But splices stamp `loc.origin` on grafted nodes and error attribution rides on those positions, so a shift here is not invisible. The multi-line/one-line agreement test compares node types; adding a `loc.line` assertion to it would pin the part that actually moved.

packages/agency-lang/docs/dev/splices.md
```diff
@@ -148,11 +148,11 @@ Unhandled interrupts are the backstop rather than the mechanism. Compilation ins
...
+`env` was the one that mattered: a generator could bake a secret into the emitted JavaScript as a string literal, and that artifact gets committed.
 
-Two mitigations, because there are two routes. Making `env` raise like its neighbour closes the Agency route and is filed as **#688**. It cannot close the JavaScript route, where an imported package reads `process.env` with no interrupt involved anywhere. So the child process receives an allowlisted environment holding only what Node needs to start (`CHILD_ENV_ALLOWED` in `runGenerator.ts`), which covers both. Its stdin is a pipe, so `readStdin` gets EOF rather than consuming the build's input.
+Two mitigations, because there are two routes. `env` now raises `std::env`, and compilation installs no handlers, so the Agency route cannot complete. That does not reach the JavaScript route, where an imported package reads `process.env` with no interrupt involved anywhere, so the child process also receives an allowlisted environment holding only what Node needs to start (`CHILD_ENV_ALLOWED` in `runGenerator.ts`). Its stdin is a pipe, so `readStdin` gets EOF rather than consuming the build's input.
```
egonSchiele: Both this file and `templates.md` now assert that a generator calling `env()` cannot complete, because compilation installs no handlers. That is the security property this PR exists to establish, and nothing tests it.

It is cheap to pin: a fixture under `tests/agency/splices/` whose generator calls `env("HOME")`, expecting the compile to fail with `AG8008`. That also guards the claim against someone later installing a default handler somewhere on the compile path, which would silently un-fix this without breaking any existing test.

The `CHILD_ENV_ALLOWED` half has the same gap, and the same fixture shape would cover it: a generator that reads `process.env` through JavaScript should come back empty.

egonSchiele (commented):
Two clean fixes, and the second one is exemplary. The code-literal diagnosis is the good kind: the error message pointed at the return type and said it expected `" \t"`, `":"` or `"="`, which sends you looking at function syntax, and the actual cause was that the third parse attempt got the untrimmed body while the first two got the trimmed one. Making the program attempt take `trimmed` and `trimmedBase` is the right fix rather than a special case, `trimmed` still carries the internal blank-line sentinels so nothing is lost, and "agrees with the same text written across lines" is exactly the test that should exist.

The `env` fix is right in principle. Reading a secret is at least as sensitive as setting one, the asymmetry with `setEnv` seven lines below was indefensible, and the decision not to widen the return type to `Result` is well argued from what the library actually does elsewhere.

**My concern is the blast radius, which I think is larger than the PR accounts for.**

The PR says two functions in `fred.agency` needed `std::env` added to their `raises` clauses, and calls that the effect system working as intended. It is, but only for functions that already *declare* a `raises` clause. Everything else gains a new interrupt with no compile-time signal at all. There are five more live Agency call sites, none of them mentioned:

| Where | Shape |
| --- | --- |
| `stdlib/data/wikidata.agency:42` | module-level `const` |
| `stdlib/data/finance/edgar.agency:50` | module-level `const` |
| `stdlib/llm.agency:114` | inside `pickProvider`, in a loop |
| `stdlib/agents/lib/search.agency:34` | function body |
| `stdlib/auth/oauth.agency:19-20` | function body, two in one expression |

(I checked the others. `system.agency:23`, `policy.agency:41`, and `calendar.agency:17` are inside docstring examples, so they are not live.)

**The two module-level constants are the ones I would look at before merging.**

```ts
// stdlib/data/wikidata.agency:42
const wikidataUserAgent = env("WIKIDATA_USER_AGENT") ?? "agency-lang (...)"
```

That is a plain `const`, not a `static const`, so it goes through the init graph and runs when the module initializes. `edgar.agency:50` is the same shape, and its comment says so: "Read once per run (env), like edgar's SEC_USER_AGENT."

An interrupt raised during module init has no user `handle` block in scope yet, by construction. So importing `std::data::wikidata` or `std::data::finance::edgar` now raises `std::env` before the program starts. I have not traced what the init path does with an unhandled interrupt, and that is the question: does it prompt, does it exit nonzero, or does something worse happen because init is not a context that expects to suspend? A test that imports one of these modules and runs `main` would answer it, and there is no such test in this PR.

If the answer is bad, `static const` with a lazy read inside the function that needs it would sidestep it, but that is a bigger change than this PR wants and it is worth knowing which case you are in first.

**`pickProvider` raises up to three times for one call.** It loops over `["anthropic", "google", "openai"]` calling `env(varName)` per provider until one is set, so a user calling it once sees three prompts. Not wrong, but worth knowing about, and it is the sort of thing that makes people reach for a blanket approve policy, which costs more than it saves.

Rest is inline and small.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Checked, and the answer is that positions do not move. Measured on a multi-line literal, before and after:

```
before: { line: 2, col: 4, start: 32 }
after:  { line: 2, col: 4, start: 32 }
```

The base position advances past the stripped prefix by exactly the amount the input shrank, so the two cancel.

One thing did change and I had not noticed it. The program attempt used to see the untrimmed body, so a multi-line literal produced a leading `newLine` node ahead of the definition. It does not any more:

```
before: ["newLine", "function"]
after:  ["function"]
```

That makes the third attempt agree with the expr and statements attempts, which always trimmed. I think it is an improvement, but it is a behaviour change and it was undocumented, so it now has its own test.

Added the position assertion you asked for, on both forms, and a note in the comment about why the numbers matter.

packages/agency-lang/stdlib/system.agency
```diff
@@ -101,7 +101,12 @@ export def cwd(): string {
...
+    name: name
+  })
   return _env(name)
```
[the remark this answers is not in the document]
egonSchiele: Correct, and I had not declared the effect. `env()` raised `std::env` with no `effect std::env { name: string }` declaration alongside `std::setEnv` and the others, so the payload went unchecked and an inline handler would have seen `e.data` as `any`.

Moot on this branch now, since the `env` change is reverted, but it is a real gap in the attempt and I have recorded it on #688 so a future attempt starts with it.

packages/agency-lang/stdlib/system.agency
egonSchiele: Agreed, and recorded on #688. Reading an environment variable is idempotent by definition, and the precedent you name is the right one: `compile`, `typecheck`, `getEffects`, `describe`, `parseAST`, and `typecheckFile` are all `export idempotent def`. `setEnv` and `openUrl` mutate, so they are not the model.

Reverted along with the rest of the `env` change, so it goes into the next attempt rather than this branch.

packages/agency-lang/tests/agency/env-interrupt.agency
egonSchiele: You are right that the fixture would have passed either way, which makes it close to worthless for its stated purpose.

The idiom is genuinely ambiguous from the source. `return interrupt std::env(...)` followed by `return _env(name)` does not tell a reader whether an approved read resumes into the second `return` or whether the interrupt result is what comes back. Asserting `value != null` cannot distinguish them. The round trip you describe, `setEnv` then `env` then compare to `"hello"`, is the version that pins it.

Recorded on #688. The `env` change is reverted here, so the fixture went with it.

packages/agency-lang/docs/dev/splices.md
egonSchiele: Half of this is moot and half is a real gap.

The moot half: that claim came from the `env` change and reverted with it. `docs/dev/splices.md` is back to saying #688 would close the Agency route and has not landed.

The real half: the JavaScript route. `CHILD_ENV_ALLOWED` is covered by a unit test in `runGenerator.test.ts` that puts a fake secret in the environment and asserts it cannot reach the result, but that test reads through Agency `env()`, not through `process.env` in an imported package. So the route you name is untested.

It is also currently unreachable: a generator may only import Agency code, so it cannot get at `process.env` at all unless `allowNonAgencyGenerators` is set. A test would have to turn that off first, which is worth doing precisely because that is the configuration where the allowlist is the only thing standing there.

Not adding it here, since this branch is now just the parser fix and I would rather not re-expand it. Happy to do it as a follow-up if you want it.

PR #695, opened 2026-07-25
Pattern matching: allow a `: Type` suffix on a nested pattern

egonSchiele:
Third of the pattern-matching items behind the safeBash command matcher, after #676 and #677.

## The problem

The bind-and-test suffix worked at the top of an arm but not inside another pattern — you could type a whole arm, but not a piece of one:

```ts
match (words) {
  ["cat", p: SafePath] => read(p)
}
// Failed to parse: expected match cases of the form `value => expression`
```

That matters more than ergonomics. For anything but a coarse built-in, the suffix runs the type's full `@validate` chain, so a validated type becomes a *matching condition* — the arm only matches when the value passes. Nested, that puts a rule's safety precondition in the pattern where it cannot be forgotten, instead of in the arm body where it can.

## No new syntax

The suffix attaches to a whole **pattern** — a binder, an object pattern, an array pattern — and never to an object field. So there is no per-field form to invent. Typing an object is what it always was, and already worked:

```ts
{ name: "Ada" }: { name: string, age: number }  =>  ✓ before and after
{ name, age }: Person                            =>  ✓ before and after
```

This PR is purely about *where* the existing parser is wired.

## The change

`matchPatternParser` parses a base pattern and then peeks for a `: Type` suffix, which makes it available wherever a pattern appears, at any depth.

Parse-then-peek rather than another `or` alternative, deliberately: an alternative that began by parsing a pattern would left-recurse on itself, and even guarded against that it would re-parse every nested element twice — once speculatively for the suffix form, once for the plain one. Peeking costs one optional parse of `:`.

`ObjectPatternProperty["value"]` and `ArrayPattern["elements"]` gained `TypePattern`, which **deleted** the three "safe narrowing" casts justified by *"nested elements never carry a typePattern"* (`parsers.ts` :5905, :6094, :6115). A net removal of casts.

**Lowering needed nothing.** `collectChecks`'s `typePattern` case already emitted the runtime test and recursed into the inner pattern, so it handled nested suffixes the moment they could be parsed.

## Now possible

```ts
match (words) {
  ["echo", s: string]              => print(s)
  ["cat",  p: SafePath]            => read(p)          // validator decides the match
  [{ cmd: "echo" }: Word, ...rest] => handle(rest)
  _                                => fallback()
}

// object > array > validated binder — the shape the bash matcher wants
{ words: ["cat", p: SafePath], redirect: null } => read(p)
```

One consequence to know: a property value is itself a pattern, so `{ name: n: string }` parses as a suffix on the binder `n`. Legal but unidiomatic — type the object instead.

## Verification

Test-first; each case watched failing before the change.

- New `tests/agency/pattern-matching/nestedTypeSuffix.agency` — 5/5, was a parse error. Covers dispatch on element type, a validated type that rejects and falls through, a nested object pattern with its own suffix, object > array > validated binder, and Rule 1 (the binder gets the original value, not the validator's).
- Six parser tests in `pattern.test.ts`, four of which fail without the change; the other two are negative controls — that the object-pattern field colon is not eaten as a suffix, and that a non-suffixable kind (a literal) does not silently become a `typePattern`.
- A formatter test: nested suffixes print back out and re-formatting is a fixpoint. Without the change it fails.
- Existing `type-patterns` 9/9, so arm-level behavior is unchanged. `pattern-matching` 29/29, `destructuring` 6/6, plus `match`, `matchExpression`, `goto-match-arm`, `module-level-match`, `match-yield-seq-thread`, `match-expr-in-handler`.
- Full `lib` suite: 9553 passed, only the 5 known pre-existing failures (`mcpBridge.contract` x4, `optimize/mutator` x1). `lib/compiler/splice/rebuild.test.ts` failed on one run and passed on another with no change in between — flaky, and it fails on a stashed tree too.
- `make fixtures` produced no changes. `make` and `pnpm run lint:structure` clean.

## Docs

`pattern-matching.md` gains the nested form, the rule that the type attaches to a pattern and not a field, and a pointer back to the AG5004 footgun. Both it and `type-validation.md` gain a cost note, measured at 200k iterations:

| check | ns/op |
|---|---|
| literal compare | 3 |
| `__coarseTypeTest(v, "string")` | 7 |
| validated type, passes | 372 |
| validated type, fails | 393 |

Coarse types are free; a named type is ~50x that, and most of it is the schema parse rather than the validator. Since arms are tried in order and conditions short-circuit, the guidance is to put cheap discriminating checks first — `["cat", p: SafePath]` never runs the validator unless the first word already matched.

### DIFF
--- a/packages/agency-lang/lib/formatter.test.ts
+++ b/packages/agency-lang/lib/formatter.test.ts
@@ -162,11 +162,14 @@ describe("formatSource", () => {
       "",
     ].join("\n");
     const formatted = formatSource(input);
-    expect(formatted).toContain("[\"echo\", s: string] =>");
-    expect(formatted).toContain("[\"cat\", p: SafePath] =>");
-    expect(formatted).toContain("[{ cmd: \"echo\" }: Word, ...rest] =>");
+    // formatSource returns null when the source does not parse, so a parse
+    // failure must not read as a silently-skipped assertion.
+    expect(formatted).not.toBeNull();
+    expect(formatted!).toContain("[\"echo\", s: string] =>");
+    expect(formatted!).toContain("[\"cat\", p: SafePath] =>");
+    expect(formatted!).toContain("[{ cmd: \"echo\" }: Word, ...rest] =>");
     // Formatting the output again changes nothing.
-    expect(formatSource(formatted)).toBe(formatted);
+    expect(formatSource(formatted!)).toBe(formatted);
   });
 
   it("round-trips a generics fixture (type params + Record) unchanged", () => {
--- a/packages/agency-lang/lib/formatter.test.ts
+++ b/packages/agency-lang/lib/formatter.test.ts
@@ -144,6 +144,31 @@ describe("formatSource", () => {
     expect(formatted).toBe(input.trimEnd() + "\n");
   });
 
+  it("prints a `: Type` suffix nested inside a pattern, and round-trips it", () => {
+    // The suffix used to be arm-level only, so the printer never met one in
+    // element or property position. A printer that dropped it would silently
+    // turn a validated rule into an unvalidated one.
+    const input = [
+      "node main() {",
+      "  const w = [\"cat\", \"f.txt\"]",
+      "  const r = match (w) {",
+      "    [\"echo\", s: string] => s",
+      "    [\"cat\", p: SafePath] => p",
+      "    [{ cmd: \"echo\" }: Word, ...rest] => \"word\"",
+      "    _ => \"other\"",
+      "  }",
+      "  print(r)",
+      "}",
+      "",
+    ].join("\n");
+    const formatted = formatSource(input);
+    expect(formatted).toContain("[\"echo\", s: string] =>");
+    expect(formatted).toContain("[\"cat\", p: SafePath] =>");
+    expect(formatted).toContain("[{ cmd: \"echo\" }: Word, ...rest] =>");
+    // Formatting the output again changes nothing.
+    expect(formatSource(formatted)).toBe(formatted);
+  });
+
   it("round-trips a generics fixture (type params + Record) unchanged", () => {
     const fixturePath = path.join(__dirname, "../tests/formatter/generics.agency");
     const input = fs.readFileSync(fixturePath, "utf-8");
--- a/packages/agency-lang/lib/formatter.test.ts
+++ b/packages/agency-lang/lib/formatter.test.ts
@@ -155,6 +155,13 @@ describe("formatSource", () => {
       "    [\"echo\", s: string] => s",
       "    [\"cat\", p: SafePath] => p",
       "    [{ cmd: \"echo\" }: Word, ...rest] => \"word\"",
+      // A wildcard with a suffix has no binder, so it takes formatPattern's
+      // `pattern === null` branch — the branch written for after-`is`, where
+      // the operator is already on the page. Printed bare in element position
+      // it would come back as a BINDER named SafePath that matches anything,
+      // silently turning a validated rule into an unvalidated one.
+      "    [\"cat\", _: SafePath] => \"anon\"",
+      "    { path: _: SafePath } => \"anon obj\"",
       "    _ => \"other\"",
       "  }",
       "  print(r)",
@@ -168,6 +175,8 @@ describe("formatSource", () => {
     expect(formatted!).toContain("[\"echo\", s: string] =>");
     expect(formatted!).toContain("[\"cat\", p: SafePath] =>");
     expect(formatted!).toContain("[{ cmd: \"echo\" }: Word, ...rest] =>");
+    expect(formatted!).toContain("[\"cat\", _: SafePath] =>");
+    expect(formatted!).toContain("{ path: _: SafePath } =>");
     // Formatting the output again changes nothing.
     expect(formatSource(formatted!)).toBe(formatted);
   });
--- a/packages/agency-lang/lib/parsers/parsers.ts
+++ b/packages/agency-lang/lib/parsers/parsers.ts
@@ -5901,10 +5901,9 @@ const _objectPatternShorthandParser: Parser<ObjectPatternShorthand> = (
 
 // Shared helpers for binding and match object-property parsers. The two
 // only differ by the inner value parser (bindingPatternParser vs
-// matchPatternParser). The value type is narrower than MatchPattern:
-// typePattern is top-level-only (is-RHS and arm position), so nested
-// property values never carry one — the parsers wired in here cannot
-// produce it, which is what makes the seam cast at the match call site safe.
+// matchPatternParser). Both produce a subset of MatchPattern; the binding
+// parser cannot produce a resultPattern or a typePattern, which is why
+// binding position keeps `: Type` as a static annotation.
 const propertyWithValueParser = (
   valueParser: Parser<ObjectPatternProperty["value"]>,
 ): Parser<ObjectPatternProperty> => (input: string) => {
@@ -6053,7 +6052,7 @@ const _isRhsParser = (input: string): ParserResult<MatchPattern> => {
 };
 export const isRhsParser: Parser<MatchPattern> = _isRhsParser;
 
-const _matchPatternParser = (input: string): ParserResult<MatchPattern> => {
+const _matchPatternBase = (input: string): ParserResult<MatchPattern> => {
   // NOTE: cannot reuse simpleLiteralParser directly because it tries
   // numberParser before variableNameParser, and numberParser is greedy on `_`
   // (e.g. `_foo` would parse as `{type: "number", value: ""}` with rest `foo`).
@@ -6076,6 +6075,52 @@ const _matchPatternParser = (input: string): ParserResult<MatchPattern> => {
   );
   return parser(input) as ParserResult<MatchPattern>;
 };
+
+/** Pattern shapes a `: Type` suffix may follow. The suffix attaches to a whole
+ *  pattern, never to an object field — `{ name: n }: {name: string}` is the
+ *  spelling, and there is no per-field form. */
+function takesTypeSuffix(pattern: MatchPattern): boolean {
+  return (
+    pattern.type === "variableName" ||
+    pattern.type === "objectPattern" ||
+    pattern.type === "arrayPattern" ||
+    pattern.type === "wildcardPattern"
+  );
+}
+
+/**
+ * A match pattern, plus an optional `: Type` suffix.
+ *
+ * The suffix used to be arm-level only, which meant `["cat", p: SafePath]` did
+ * not parse — you could type the whole arm but not a piece of it. Wiring it in
+ * HERE makes it available wherever a pattern appears, at any depth, because
+ * nested elements and property values both come through this parser.
+ *
+ * Written as parse-base-then-peek rather than as another `or` alternative on
+ * purpose. An alternative that began by parsing a pattern would left-recurse
+ * (this parser calling itself at the same position), and even guarded against
+ * that it would re-parse every nested element twice — once speculatively for
+ * the suffix form, once for the plain one. Peeking costs one optional parse of
+ * `:` and nothing else.
+ */
+const _matchPatternParser = (input: string): ParserResult<MatchPattern> => {
+  const base = _matchPatternBase(input);
+  if (!base.success) return base;
+  const pattern = base.result as MatchPattern;
+  if (!takesTypeSuffix(pattern)) return base;
+  const suffix = armTypeSuffixParser(base.rest);
+  if (!suffix.success) return base;
+  return success(
+    {
+      type: "typePattern" as const,
+      // `_: Type` is a test with nothing to bind, matching wildcardSuffixParser.
+      pattern: pattern.type === "wildcardPattern" ? null : (pattern as BindingPattern),
+      typeHint: (suffix.result as { typeHint: VariableType }).typeHint,
+      loc: "loc" in pattern ? pattern.loc : undefined,
+    } as MatchPattern,
+    suffix.rest,
+  );
+};
 export const matchPatternParser: Parser<MatchPattern> = _matchPatternParser;
 
 export const arrayMatchPatternParser: Parser<ArrayPattern> = label(
@@ -6090,8 +6135,6 @@ export const arrayMatchPatternParser: Parser<ArrayPattern> = label(
           or(
             sepBy(
               commaWithNewline,
-              // Safe narrowing: nested array elements never carry a
-              // typePattern (top-level-only; see ArrayPattern in pattern.ts).
               lazy(() => matchPatternParser) as Parser<ArrayPattern["elements"][number]>,
             ),
             succeed([]),
@@ -6111,12 +6154,7 @@ export const arrayMatchPatternParser: Parser<ArrayPattern> = label(
 
 const _matchObjectPropertyParser: Parser<
   ObjectPatternProperty | ObjectPatternShorthand | RestPattern
-> = objectPatternPropertyParser(
-  // Safe narrowing: matchPatternParser only produces typePattern where
-  // typePatternParser is wired (is-RHS and arm top level), never in nested
-  // property position.
-  lazy(() => matchPatternParser) as Parser<ObjectPatternProperty["value"]>,
-);
+> = objectPatternPropertyParser(lazy(() => matchPatternParser));
 
 export const objectMatchPatternParser: Parser<ObjectPattern> = label(
   "an object match pattern",
--- a/packages/agency-lang/lib/types/pattern.ts
+++ b/packages/agency-lang/lib/types/pattern.ts
@@ -5,9 +5,9 @@ import type { Expression, VariableType } from "../types.js";
 export type ObjectPatternProperty = {
   type: "objectPatternProperty";
   key: string;
-  // ResultPattern is only valid in match-position use; the parser does not
-  // produce it in binding-position contexts.
-  value: BindingPattern | Literal | ResultPattern;
+  // ResultPattern and TypePattern are only valid in match-position use; the
+  // parser does not produce either in binding-position contexts.
+  value: BindingPattern | Literal | ResultPattern | TypePattern;
 };
 
 export type ObjectPatternShorthand = {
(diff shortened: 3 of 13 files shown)


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -6076,6 +6075,52 @@ const _matchPatternParser = (input: string): ParserResult<MatchPattern> => {
...
+  if (!base.success) return base;
+  const pattern = base.result as MatchPattern;
+  if (!takesTypeSuffix(pattern)) return base;
+  const suffix = armTypeSuffixParser(base.rest);
+  if (!suffix.success) return base;
+  return success(
+    {
+      type: "typePattern" as const,
+      // `_: Type` is a test with nothing to bind, matching wildcardSuffixParser.
+      pattern: pattern.type === "wildcardPattern" ? null : (pattern as BindingPattern),
```
egonSchiele, line 6123: This is the line I'd look hardest at. `pattern: null` is fine at arm level because `formatArmCaseValue` (`agencyGenerator.ts:944`) special-cases it into `is Type`, and fine after `is` because the caller wrote the operator. Nested, neither holds — `formatPattern`'s null branch (`:928`) prints the bare type name, and `formatArrayPattern` sends every element straight through it.

So `["cat", _: SafePath]` looks like it formats to `["cat", SafePath]`, which re-parses as a binder that matches anything. See item 1 in the review body; worth running `formatSource` on it to confirm.

If `_: Type` nested has no real use case, rejecting it here is simpler than teaching the printer a third context.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -6076,6 +6075,52 @@ const _matchPatternParser = (input: string): ParserResult<MatchPattern> => {
...
+      typeHint: (suffix.result as { typeHint: VariableType }).typeHint,
+      loc: "loc" in pattern ? pattern.loc : undefined,
```
egonSchiele, line 6125: `loc: "loc" in pattern ? pattern.loc : undefined` — when does the `in` check fail, and is `undefined` acceptable when it does?

A nested type suffix is the position where a user most needs a location: the failure mode is "my arm silently didn't match", and the diagnostic pointing at the whole match block instead of the element is much less useful than at arm level, where the arm and the pattern are nearly the same span.

Falling back to the suffix's own location (`suffix.rest` offset, or whatever `armTypeSuffixParser` carries) would be better than `undefined`. Ideally the span covers base-through-suffix rather than just the base.

packages/agency-lang/lib/types/pattern.ts
```diff
@@ -22,9 +22,16 @@ export type ObjectPattern = BaseNode & {
 
 export type ArrayPattern = BaseNode & {
   type: "arrayPattern";
-  // ResultPattern is only valid in match-position use; the parser does not
-  // produce it in binding-position contexts.
-  elements: (BindingPattern | Literal | WildcardPattern | RestPattern | ResultPattern)[];
+  // ResultPattern and TypePattern are only valid in match-position use; the
```
egonSchiele, line 25: Worth being explicit that this comment is now the *only* thing keeping a `TypePattern` out of binding position — before the widening, the type itself made `let [a, b: Foo] = x` unrepresentable, and the three casts were the one seam bridging that.

It holds today (`extractBindings` handles `typePattern`, and `hygiene.ts`'s `isBindingPattern` filters it out), so I'm not asking for a `MatchArrayPattern` / `BindingArrayPattern` split here. But the description frames the cast removal as free, and it isn't quite — see item 3.

packages/agency-lang/lib/formatter.test.ts
```diff
@@ -144,6 +144,31 @@ describe("formatSource", () => {
...
+      "  print(r)",
+      "}",
+      "",
+    ].join("\n");
+    const formatted = formatSource(input);
+    expect(formatted).toContain("[\"echo\", s: string] =>");
+    expect(formatted).toContain("[\"cat\", p: SafePath] =>");
+    expect(formatted).toContain("[{ cmd: \"echo\" }: Word, ...rest] =>");
+    // Formatting the output again changes nothing.
+    expect(formatSource(formatted)).toBe(formatted);
```
egonSchiele: All three cases here have a binder, so all three take `formatPattern`'s non-null branch. The `pattern === null` branch — `_: Type`, which the description says works nested — is never printed by this test, and I think it is where the round-trip breaks (item 1).

Adding `["cat", _: SafePath] => "x"` to the input would either pass and settle it, or fail and catch the bug. Given the comment above about a printer silently unvalidating a rule, this is the case that comment is really about.

packages/agency-lang/tests/agency/pattern-matching/nestedTypeSuffix.agency
```diff
@@ -0,0 +1,74 @@
...
+  }
+  return success(v)
+}
+
+@validate(notEmpty)
+type SafePath = string
+
+type Word = { cmd: string }
+
+def describe(words: any[]): string {
```
egonSchiele, line 17: Every scrutinee in this file is `any` or `any[]`, which is right for testing runtime behavior but means these tests cannot say anything about types. In particular they cannot tell you whether `p` in `["cat", p: SafePath]` is narrowed to `SafePath` in the arm body — with `any[]`, `p` is `any` either way and every downstream call is accepted.

Your idea doc lists `typeChecker/typePatterns.ts` and `matchArmNarrowing` as touch points; a typechecker test over a `string[]` scrutinee would pin whichever answer is intended. See item 2.

egonSchiele (commented):
## Overview

Moves the `: Type` suffix from arm-level-only into `matchPatternParser`, so it works wherever a pattern appears. The parse-then-peek design is right and the reasoning for it (left recursion, double-parsing every nested element) is correct — an `or` alternative here would have been the obvious wrong turn. Lowering needing zero changes is a good sign the seam was already in the right place.

The cost table in the docs is a genuinely useful addition; "put cheap discriminating checks first, conditions short-circuit" is the kind of guidance that is hard to discover on your own.

One likely-serious bug below, then a few gaps.

## 1. `_: Type` nested silently loses its check when formatted

The PR notes that `_: Type` works nested and produces a `pattern: null` node. The formatter cannot print that node correctly outside arm position.

`agencyGenerator.ts:922` — `formatPattern`'s `typePattern` case:

```ts
return tp.pattern === null
  ? typeStr                                  // just `SafePath`
  : `${this.formatPattern(tp.pattern)}: ${typeStr}`;
```

The null branch prints the bare type name. That is correct in the two places it was written for — after `is`, where the caller already wrote the operator, and in an arm, where `formatArmCaseValue` (`:944`) special-cases it to `is Type`. Nested, neither applies: `formatArrayPattern` (`:972`) maps every element straight through `formatPattern`, and `formatObjectPattern` (`:967`) does the same for property values.

So I believe:

```ts
["cat", _: SafePath]   →  formats to  →  ["cat", SafePath]
{ words: _: SafePath }  →  formats to  →  { words: SafePath }
```

and `["cat", SafePath]` re-parses as a **binder named `SafePath`** that matches anything — your own parser test (`leaves a pattern with no suffix alone`) confirms a bare identifier in element position is a `variableName`.

If that is right, running `pnpm run fmt` silently converts a validated match into an unvalidated one that always matches. That is precisely the failure your formatter test's comment names — "a printer that dropped it would silently turn a validated rule into an unvalidated one" — and it is worst for the safeBash use case this whole series is building toward.

I read this from the code rather than running it, since the branch isn't built here. Worth confirming with `formatSource` on `["cat", _: SafePath] => x` before deciding what to do. If it reproduces, the options are to teach `formatPattern` to print `_: Type` in nested position, or to reject `_: Type` nested at parse time (a wildcard with a type test and nothing to bind is not obviously useful inside a pattern — `is Type` at arm level covers the useful case).

Either way the existing formatter test does not catch it: all three of its cases have a binder, so all three take the non-null branch.

## 2. The typechecker was a listed touch point and got nothing

Your idea doc lists `lib/typeChecker/typePatterns.ts` / `matchArmNarrowing` under Touch Points, for "narrowing for a nested binder." The PR touches no typechecker file and adds no typechecker test.

The execution tests cannot detect the gap because every scrutinee in `nestedTypeSuffix.agency` is `any` or `any[]`, so `p` is `any` no matter what the checker concludes and everything downstream is accepted. The question I'd want answered:

```ts
def needsPath(p: SafePath): string { ... }

def f(words: string[]): string {
  return match (words) {
    ["cat", p: SafePath] => needsPath(p)   // is `p` SafePath here, or string?
    _ => "none"
  }
}
```

If `p` stays `string`, that is a real hole — the arm has proven `SafePath` at runtime and the checker does not know. Not necessarily a blocker (arm-level narrowing could have been the only case anyone needed), but it should be a known answer with a test pinning it, not an open question.

Same for exhaustiveness: your doc lists "a match whose only arms use nested type suffixes still requires `_`" as a planned test and it isn't here. `matchExhaustiveness.ts:275` unwraps a `typePattern` only at caseValue level, so nested ones are probably invisible to it, which is probably the right answer — but "probably" is what a test is for.

## 3. The cast removal trades a type invariant for a comment

The description frames this as a net removal of casts, and it is, but it is worth naming what was traded. Before, `ArrayPattern["elements"]` genuinely could not hold a `TypePattern`, and the three casts existed to bridge that at exactly one seam. Now the union permits it everywhere the type is used — including binding position, where `let [a, b: Foo] = x` must never occur — and the only thing preventing it is the comment saying the parser won't do that.

It happens to be fine today: `extractBindings` handles `typePattern` (`patternLowering.ts:1084`), and `hygiene.ts`'s `isBindingPattern` (`:69`) filters it out silently. But that second one is the shape that worries me — a binder inside a nested type suffix is invisible to `patternBinders`, so if match patterns are ever fed to template hygiene, `["cat", p: SafePath]` would not rename `p`. That is a latent hole rather than a live one (match patterns aren't fed there today, same as the recorded `is success(v)` limit), but the type no longer protects it.

Not asking you to split `MatchArrayPattern` from `BindingArrayPattern` in this PR. Just: three casts and one enforced invariant is a different trade than three casts and nothing, and the description reads as if it were free.

## 4. The question your idea doc raised is still open

The doc asks whether AG5004 should become an error inside object patterns "once a correct spelling exists" — the reasoning being that someone who learns `[p: SafePath]` will try `{path: SafePath}` and silently get a binder named `SafePath`. This PR makes that *more* likely, since it teaches exactly the nested form that invites the mistake, and the guide's new section says "remember" without changing the diagnostic.

The adjacent case is worse now: `{ path: _: SafePath }` is legal, and per item 1 it may format into `{ path: SafePath }`, which is the AG5004 footgun. Worth deciding here rather than deferring, since this PR is what makes the confusion reachable.

## Smaller

- `parsers.ts` — `loc: "loc" in pattern ? pattern.loc : undefined`. See the inline comment; a nested type pattern is exactly where a user will need a position, since "the arm silently didn't match" is the confusing failure.
- `{ name: n: string }` being legal-but-unidiomatic: the docs say type the object instead, which is right, but nothing stops the double colon. A lint rule (AL-something) pointing at the idiomatic spelling would cost little.

## Verdict

The parser change is the right shape and I'd take it. Item 1 needs confirming and, if it reproduces, fixing — a formatter pass that deletes a safety check is worse than the parse error this PR removes. Item 2 is a test-and-answer, not necessarily code. Items 3 and 4 are judgment calls worth making explicitly rather than by default.

egonSchiele (commented):
Item 1 reproduces. Fixed in 0072cffae, along with a second soundness bug your item 2 led me to.

## 1. Confirmed, and worse than "looks like"

```
["cat", _: SafePath]    →  ["cat", SafePath]
{ path: _: SafePath }   →  { path: SafePath }
```

End to end, with a validated `SafePath`:

| input | before `fmt` | after `fmt` |
|---|---|---|
| `["cat", "f.txt"]` | matches | matches |
| `["cat", ""]` | **no match** | **matches** |
| `["cat", 42]` | **no match** | **matches** |

So `agency fmt` deleted the safety check, exactly as you read it. Worse than the parse error this series removed.

Your framing — a third context the printer doesn't know — is the fix. The node has three spellings and the printer knew two:

- after `is` → the bare type, operator already written → now `formatIsRhs`
- an arm → `is Type` → `formatArmCaseValue`, unchanged
- nested → `_: Type` → `formatPattern`, the general case

I went with teaching the printer rather than rejecting `_: Type` nested, because it works correctly at runtime today (verified above — it's only the printer that loses it), and it's the only way to say "this element must be a `SafePath`, I don't need the value". Rejecting it would be a carve-out, and `is Type` doesn't parse in nested position as an alternative spelling.

The formatter test now includes both wildcard cases. Both fail without the fix.

## 2. Not just untested — unsound

The narrowing question first. Arm level narrows, nested does not:

```ts
def f(x: string | number)      { match (x)  { v: string        => takesString(v) } }  // clean
def g(xs: (string | number)[]) { match (xs) { ["k", v: string] => takesString(v) } }  // AG6019
```

Pinned in `typePatterns.test.ts` with the reasoning rather than fixed: the checker is more conservative than runtime, so it rejects valid programs instead of accepting invalid ones. Wrong in the safe direction, and lifting it means teaching arm narrowing to walk into nested positions.

Exhaustiveness turned out to be the real find. Your "probably invisible to it, which is probably the right answer" — invisible, but that is the *wrong* answer:

```ts
match (x) {                          // x: A | B
  { tag: "a", v: _: string } => "a"  // only matches when v is a string
  { tag: "b" } => "b"
}                                    // reported EXHAUSTIVE
```

The first arm can fail, so a value falls through every arm — for an expression match, no value at all. `guarded` already means "this arm can fail for a reason coverage cannot see" and is honored at `isCatchAll`, `coveredCases`, and member-key coverage, so a nested type test now sets that flag rather than adding a parallel concept every decision point would have to learn. Three tests: fails without the suffix removed, is exhaustive without the suffix, and is exhaustive again with a `_`.

## 3. Fair — and there is a live edge I should flag

You're right that the description read as if the cast removal were free. Before, the type made `let [a, b: Foo] = x` unrepresentable; now only the parser does.

On the hygiene hole you spotted: it's live sooner than "latent". `isBindingPattern` filtering out `typePattern` means a binder inside a nested suffix is invisible to `patternBinders`, so `["cat", p: SafePath]` would not rename `p`. Match patterns aren't fed to hygiene today, so nothing is broken — but issue #696 (the `as` migration) touches hygiene's `expandShorthand` directly, so that is the moment to decide whether `MatchArrayPattern` and `BindingArrayPattern` want splitting. I've noted it there rather than doing it here.

## 4. Deferring deliberately, not by default

AG5004 is deleted outright by #696 — under `as`, `{ path: SafePath }` becomes a type test and the footgun stops existing rather than being warned about. Promoting it to an error here, then deleting it two PRs later, is churn. Same for a lint rule on `{ name: n: string }`: under #696 that spelling is unreachable, since renaming moves to `as`.

You filed the AG5004 question in the idea doc and the answer turned out to be "the syntax change resolves it" — I'd rather record that than half-fix it now.

## Smaller: `loc`

Left as-is, and I should be straight that this is a judgment call I'm less sure about. The `in` check is defensive — every pattern node the base parser produces carries `loc` — so the `undefined` branch is unreachable in practice rather than a real fallback. Your point about a span covering base-through-suffix stands on its own merits, but it changes locations for the arm-level form too, which has its own fixture expectations. Happy to do it here if you'd rather not carry it.

## Verification

- Formatter test covers both wildcard cases; fails without the fix.
- Three exhaustiveness tests; the first fails without the fix.
- Two narrowing tests pinning arm-level vs nested.
- Full `lib`: 9564 passed, only the 5 known pre-existing failures.
- `pattern-matching` 29/29, `type-patterns` 9/9, `make fixtures` no changes.
- `make clean && make` from cold, exit code checked this time — that is how the TS2345 got past me before.

## One thing that is not mine

While chasing item 2 I hit `AG2008: Property 'v' is not available on every member of 'A | B'` on a nested suffix, and checked whether I had caused it. Hand-written code does the same:

```ts
if (x.tag == "a" && x.v is string)  // AG2008
if (x.tag == "a" && x.v == 1)       // clean
```

Narrowing accumulated across `&&` isn't carried into an `is` operand. Pre-existing; this PR only makes it reachable from a pattern. Worth its own issue — say the word and I'll file it.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Reproduces. Fixed by teaching the printer the third context rather than rejecting the form: `formatIsRhs` prints the bare type after `is`, `formatArmCaseValue` prints `is Type` at arm level, and `formatPattern` now prints `_: Type`. I kept the form because it works correctly at runtime today (only the printer lost it) and it is the only way to say 'this element must be a SafePath, I do not need the value' — `is Type` does not parse nested as an alternative.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Left as-is, and I am less sure about this one than the others. The `in` check is defensive: every pattern node the base parser produces carries `loc`, so the `undefined` branch is unreachable rather than a real fallback. Your base-through-suffix span is a genuine improvement but changes arm-level locations too, which have their own fixture expectations. Happy to do it here if you would rather not carry it separately.

packages/agency-lang/lib/types/pattern.ts
egonSchiele: Fair — the description read as if it were free. Worth adding that the hygiene hole is live sooner than latent: `isBindingPattern` filtering out `typePattern` means a binder inside a nested suffix is invisible to `patternBinders`, so `["cat", p: SafePath]` would not rename `p`. Nothing feeds match patterns to hygiene today, but #696 touches `expandShorthand` directly, so that is the moment to decide on splitting MatchArrayPattern from BindingArrayPattern. Noted there.

packages/agency-lang/lib/formatter.test.ts
egonSchiele: You were right that this was the case the comment was really about. Added both `["cat", _: SafePath]` and `{ path: _: SafePath }`; both fail without the printer fix.

packages/agency-lang/tests/agency/pattern-matching/nestedTypeSuffix.agency
egonSchiele: Added a typechecker test instead, since you are right that `any[]` cannot answer it. The answer: arm level narrows, nested does not — `["k", v: string]` over a `(string | number)[]` leaves `v` as `string | number`. Pinned rather than fixed, with the reasoning: the checker is more conservative than runtime, so it rejects valid programs rather than accepting invalid ones. Chasing this also turned up the exhaustiveness bug, which was unsound in the other direction — see the review reply.

Issue #691, opened 2026-07-25
Compile-time splices: refuse effectful generators statically, once #680 lands


egonSchiele:
## Summary

Compile-time splices (#679) shipped relying on the **unhandled-interrupt backstop**: compilation installs no handlers, so a generator that reaches a dangerous operation cannot complete. That works, but the error arrives after the generator has already started, as a runtime failure rather than a compile-time refusal.

The PR originally also refused effectful generators statically, as `AG8003`, before running anything. That layer was removed to keep the change small. This issue is about bringing it back once it can be made precise.

## Why it was removed rather than kept

It could not be made precise, because of #680: interrupt effects do not propagate across a module boundary.

```
helper.agency alone      ->  { h: ["std::read"] }
gen.agency, calling h()  ->  { g: [] }
```

A generator that delegates its effectful work one file away reports an empty effect list. To fail closed anyway, the check had to ask a much blunter question: does *any* exported function anywhere in the generator's transitive closure raise an effect? That produces false positives that a user cannot work around:

```ts
// helpers.agency
export def formatName(s: string): string { ... }       // what the generator uses
export def saveReport(): Result { write("out.txt") }   // never called by the generator
```

The generator only calls `formatName`, but `saveReport` sits in the same file, so the whole generator is refused.

A coarse check that fails closed is defensible when it is the only protection. It is much harder to justify when the backstop already catches the same cases at runtime, and the coarse version is the difference between an early error and a late one.

## What to do

Land #680 first. Once `getEffectsFromFile` propagates across imports, the check becomes a direct lookup on the generator's own name, with no closure walk and no false positives.

The removed implementation is in the history of #679 (`lib/compiler/splice/eligibility.ts`, `checkEffects` and `effectsInFile`), along with its tests and the `refuseEffects` fixture. The diagnostic name `spliceGeneratorHasEffects` and code `AG8003` were deleted from the registry; the codes are unused rather than reserved, so they can be reused.

## Related

- #680 — cross-module effect propagation, the blocker
- #690 — `--only-stdlib`, which covers the other half of what the removed layer did

egonSchiele:
This is done. It landed in #697, the PR that closed the blocker, and the issue was just never closed with it.

**The blocker is gone.** #680 is closed — `symbol.interruptEffects` now propagates across imports.

**The check is back, in the precise form this issue asked for.** `lib/compiler/splice/generatorEffects.ts`, added by #697, does exactly the direct lookup described under "What to do":

```ts
[...]

PR #697, opened 2026-07-25
Propagate interrupt effects across file boundaries

### LINKED ISSUE #680: Interrupt effects do not propagate across module boundaries
## Summary

Interrupt effects do not propagate across a module boundary. A function that calls an effectful function in another file reports **no effects at all**.

Found while implementing compile-time splices (#679), where an empty effect list is the answer that means "safe to run during compilation."

## Reproduction

Two files:

```ts
// helper.agency
export def h(): Result {
  return read("data.txt")
}
```

```ts
// gen.agency
import { h } from "./helper.agency"

export def makeThing(): Result {
  return h()
}
```

Then:

```ts
getEffectsFromFile("helper.agency")  // { h: ["std::read"] }   correct
getEffectsFromFile("gen.agency")     // { makeThing: [] }      WRONG
```

`makeThing` reads a file. It calls `h`, and `h` raises `std::read`. The reported list is empty.

## What is and isn't the cause

Not path-vs-string resolution. That was the first guess and it is wrong.

`getEffectsFromSource` passes `undefined` as `sourcePath`, so `withSourcePath` (`lib/compiler/typecheck.ts:58-67`) writes the source to a fresh temp dir where `./helper.agency` does not exist. In that case import resolution **throws** `ImportResolutionError: Symbol 'h' is not defined in './helper.agency'` — it does not return a short answer.

Passing a real path fixes resolution: no throw, the import resolves. Effects still do not propagate. So resolution and propagation are separate problems and only the first one is about paths.

`analyzeInterruptsFromScopes` (`lib/typeChecker/interruptAnalysis.ts:74`) seeds imported functions from `ctx.interruptEffectsByFunction` and then propagates transitively through callees. The seeding appears not to carry effects for functions imported from relative `.agency` files, so the transitive step has nothing to propagate.

## Why it matters

For splices this was load-bearing. The spec's argument is that Agency can run code at compile time more safely than Template Haskell because dangerous operations raise effects and effects are checkable in advance. That argument evaluates to nothing for any generator that puts its actual work one file away — which is how anyone would naturally organize code, not a clever evasion.

The workaround shipped in #679 is deliberately blunt: the effect check walks the generator's whole transitive closure of relative `.agency` files and refuses if any export anywhere in it has effects. It fails closed, but it produces false positives — a helper file with one unrelated effectful function blocks a generator that never calls it.

Worth checking, and I have not: whether this also weakens the user-facing `getEffects` in `std::agency`. That one takes a source string, and its docstring already says relative imports cannot be resolved from a string, with re-exports from relative paths reporting `["unknown"]` — a documented fail-closed story. Whether the propagation gap opens a hole underneath that story needs someone to look.

## Suggested fix

Make effect propagation follow relative imports: when the symbol table has resolved an import to a real file, that file's exported effect profiles should seed `interruptEffectsByFunction` for the importing module.

Once that lands, `#679`'s closure-walking workaround in `lib/compiler/splice/eligibility.ts` should be replaced with a direct lookup, and its comment says so.

## Test that should exist

The two-file case above, asserting `{ makeThing: ["std::read"] }`. There is a version of it in `lib/compiler/splice/eligibility.test.ts` ("rejects a generator whose relative helper is effectful"), but it tests the workaround rather than the underlying behavior.

egonSchiele:
Fixes #680 and #691.

## The bug

Effects did not cross a file boundary. The compiler collects them in two steps: a crawl that looks for the literal `interrupt` keyword, and a type-check pass that follows calls. The second only ever saw one file, so an imported function arrived as a dead end.

```
// helper.agency
export def h(): string { return read("data.txt") }

// main.agency
import { h } from "./helper.agency"
node main() { const x = h() }
```

`helper.agency` reported `{ h: ["std::read"] }`. `main.agency` reported `{ main: [] }`. And because `agency policy gen` reads that list, it printed **"No interrupt effects found in this agent. No policy needed."** for a program that reads your filesystem.

It stayed hidden because every risky function in the standard library has a literal `interrupt` in its body, so the crawl's answer happens to be right for all of them. The gap opens the moment someone wraps one.

## The second bug, which is not about files

Review of the spec found this, and it is larger. `read("a.txt")` and `read.invoke("a.txt")` do the same thing, and the project's own style prefers the second. They parse differently, and every effect walk read the name off the call node — which for the second form is the string `invoke`.

```
export def plain(): string     { return read("a.txt") }
export def viaInvoke(): string { return read.invoke("a.txt") }

→ { plain: ["std::read"], viaInvoke: [] }
```

One file, no imports. Two walks needed teaching, not one: `checkUnhandledInterruptWarnings` reads the call site directly rather than going through the shared walk, so AG3009 stayed silent until it got its own fix.

## The fix

One module, `lib/analysis/effects.ts`, owns how an effect list is computed: the syntax walk, the fixpoint, and a pass that runs both over every reachable file at the end of `SymbolTable.build`. The type checker imports the walk and the fixpoint instead of keeping its own copies — two copies of this analysis is how effects came to mean different things on either side of an import in the first place.

Direct effects are read from the symbol rather than recomputed from the body. That is load-bearing: `_guard` has no `interrupt` in its body at all, and its `std::guard` label comes from a seed table because the trip is raised on the TypeScript side. A body walk would have erased cost caps from the whole program.

## Reviewer attention

**`stdlib/agents/planner.agency` had genuinely wrong declarations.** `runApproved`, `planStep` and `plannerAgent` declared `raises <std::agents::planApprove>` while reaching `std::run` and `std::guard` through calls into other files. Corrected. This is a real defect the change surfaced, not test churn.

**`agency doc` had no symbol table at all**, and its own comment said so. So its Throws column never saw an imported effect. Fixing that changed 27 pages. The notable ones: every data-source function was documented as raising only its own effect while also making an HTTP request.

```
- **Throws:** `std::yc`
+ **Throws:** `std::yc`, `std::http::fetchJSON`
```

Same for `littlesis`, `hackernews`, `wikidata`, `fred`, `edgar`, `usaspending`, `gdelt`, `dbnomics`.

**A bug I introduced, caught by an existing test.** The crawl deliberately tolerates an uninstalled `pkg::` import and keeps going. My resolver did not, so one missing dependency blanked out the effects of every other import in the same file. Fixed to match the crawl, with a regression test.

**The fixpoint was quadratic and the scaling test caught it.** On a chain of N functions where only the last one raises, each sweep moved the effect one link. It breached the growth bound at 2.369. Rewritten as a worklist that re-queues only the callers of an entry that grew: now 0.457.

**Two deliberate narrowings in the `.invoke()` handling**, both forced by tests. A method call in an access chain now contributes no callee unless it is `invoke`, because `xs.map(...)` names a method on a value, not a global. And `invoke` attributes to the base only when every earlier chain link is itself a method call, so `f.partial(...).invoke()` resolves to `f` while `obj.handler.invoke()` resolves to nothing.

## Generator checks (#691)

`lib/compiler/splice/eligibility.ts` had no effect check; its comment said one was impossible while #680 stood. It now refuses a generator whose effects are non-empty (AG8003) and one whose effects could not be read (AG8004), naming which of four blind spots it hit. Scoped to what the generator reaches **by calling**, not to the files it reaches — a file-scoped rule would refuse every generator, since the prelude is always reachable.

## Verification

- 620 files / 9,654 tests passing, from a 613 / 9,586 baseline. Zero regressions.
- Task 1 was a pure refactor and the full-suite counts confirmed it moved nothing.
- Structural lint clean. Diff audited against `docs/dev/anti-patterns.md`.
- New: `docs/dev/effect-propagation.md`, plus a `lib/perf/` scaling test.

## Known limits, documented

The walk reads syntax, so four things are invisible: a file that does not parse, an unexpanded splice, a function called through a parameter, and a reference passed through a variable. All under-report. The generator check refuses on all four; the reporting consumers document them. The `getEffects` docstring in `std::agency` now says so.

The type checker may find more effects than the shared walk and never fewer. There is a test pinning that.

**Left undone deliberately:** `h.invoke(cb)` is excluded from the AG3014/AG3015 raises checks, because `callFlows` skips calls whose parent is a `valueAccess`. Narrowing that to real method calls needs receiver resolution. Recorded in a comment at the site; worth its own issue.


packages/agency-lang/lib/compiler/splice/eligibility.ts
```diff
@@ -193,17 +199,203 @@ export function checkNoNestedSplice(
...
+ *  unknown. */
+function heldFunctionReference(
+  declaration: FunctionDefinition | GraphNodeDefinition,
+  callees: string[],
+): string | null {
+  const assignedNames: Record<string, string> = Object.create(null);
+  for (const { node } of walkNodesArray(declaration.body)) {
+    if (node.type !== "assignment") continue;
+    const value = node.value;
+    if (value && value.type === "variableName" && !callees.includes(value.value)) {
```
egonSchiele: This treats *any* alias of a bare name as a held function reference, so ordinary plumbing gets refused:

```
export def makeThing(title: string): Code {
  const label = title
  return build(label)
}
```

`title` is a string, `label` is a string, and the generator is refused with "passes 'title' through a variable". The check has no way to tell a function from anything else here.

A symbol-table lookup would narrow it sharply — treat it as a blind spot only when the aliased name resolves to a function or node symbol, or to a parameter whose declared type is a function type. Everything else is a value being renamed. Fail-closed is the right default, but the case for scoping blind spots to the call graph rather than to files was that a generator which is fine should still run.

packages/agency-lang/lib/compiler/splice/eligibility.ts
```diff
@@ -193,17 +199,203 @@ export function checkNoNestedSplice(
...
+      assignedNames[node.variableName] = value.value;
+    }
+  }
+  for (const { node } of walkNodesArray(declaration.body)) {
+    if (node.type !== "functionCall") continue;
+    for (const argument of node.arguments) {
+      if (
+        argument.type === "variableName" &&
```
egonSchiele: This only matches a bare positional argument, so the two forms most likely to carry a callback slip past:

```
f(cb: handler)      // NamedArgument
f(...refs)          // SplatExpression
```

For a check whose purpose is refusing when the reading is incomplete, missing a real blind spot is the dangerous direction — worse than the over-refusal in the comment above. `positionalArg` in `lib/typeChecker/interruptAnalysis.ts` already unwraps both shapes and could be reused, or lifted next to `collectBodyFacts`.

packages/agency-lang/lib/compiler/splice/eligibility.ts
```diff
@@ -193,17 +199,203 @@ export function checkNoNestedSplice(
...
+): FunctionDefinition | GraphNodeDefinition | null {
+  for (const { node } of walkNodesArray(program.nodes)) {
+    if (node.type !== "function" && node.type !== "graphNode") continue;
+    if (declaredName(nameOf(node)) === name) return node;
+  }
+  return null;
+}
+
+/** The parse tree of every file the table crawled, keyed by path. */
+function programsFor(
```
egonSchiele: This re-parses every file the table crawled — the prelude and everything it drags in — and `reachableFrom` then builds summaries over all of them, once per splice site.

`parseAgencyFileCached` returns `structuredClone(entry.program)` on every read (`lib/parseCache.ts:96`), so a cache hit still deep-copies the whole AST. On the three call sites that cannot pass a `symbolTable` (`compile.ts:129`, `buildSession.ts:647`, `lsp/diagnostics.ts:139`) this runs alongside a full `SymbolTable.build`, and the editor one runs on every keystroke in a file with a splice.

The summaries do not depend on which splice is being checked. Computing them once per `expandSplices` call and passing them down through `DecisionContext` would leave the per-site cost at the reachability walk.

packages/agency-lang/lib/compiler/splice/eligibility.ts
```diff
@@ -193,17 +199,203 @@ export function checkNoNestedSplice(
...
+  symbolTable?: SymbolTable,
+): SpliceDiagnostic | null {
+  const absolute = path.resolve(generatorPath);
+  const table = symbolTable ?? SymbolTable.build(absolute, config);
+  const symbol = table.getFile(absolute)?.[generatorName];
+  if (!symbol || (symbol.kind !== "function" && symbol.kind !== "node")) {
+    return refusal(generatorName, "its definition could not be found");
+  }
+
+  const effect = (symbol.interruptEffects ?? [])[0]?.effect;
```
egonSchiele: Only the first effect reaches the message, so a generator reaching `std::read`, `std::write` and `std::run` is refused three times — fix one, get refused for the next. Joining them reads better and matches how AG3013 lists what a `raises` clause exceeded.

packages/agency-lang/lib/analysis/effects.test.ts
```diff
@@ -0,0 +1,242 @@
...
+    for (const file of files) {
+      const parsed = parseAgencyFileCached(file, {}, false);
+      if (!parsed.success) continue;
+      for (const decl of parsed.result.nodes) {
+        if (decl.type !== "function" && decl.type !== "graphNode") continue;
+        const raw: string[] = [];
+        scanRaw(decl.body, types, raw);
+        if (raw.length === 0) continue;
+        const facts = collectBodyFacts(decl.body);
+        if (facts.calls.length + facts.effects.length === 0) {
```
egonSchiele: This is a real tripwire now rather than a tautology, which is the important part. But the assertion is that a call-bearing body reads as *non-empty* — a body with ten calls where the walk finds one still passes. The drift it needs to catch is a new construct holding calls inside a body that has other calls too, which is exactly the case it lets through.

Both walks visit the same tree, so identity comparison is available and strict: collect the `functionCall` nodes from `scanRaw` and assert every one of them appears in `facts.calls`. Same for `interruptStatement` against `facts.effects`.

Also, `readdirSync("stdlib")` is not recursive, so this covers the top level only. `stdlib/agents/`, `stdlib/data/` and the rest are where the more unusual constructs live — and `data/` is where this PR found the biggest doc corrections.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -141,12 +142,25 @@ function generateDocForFile(
   ctx: DocContext,
   program: AgencyProgram,
 ): void {
-  const info = buildCompilationUnit(program);
+  // A symbol table is what makes an imported function's effects visible. The
+  // Throws column understated everything reached through an import without it
+  // — a guard block in another module never showed up (GitHub issue 680).
+  // Building it can throw on an unresolvable import, and the doc command
+  // should still produce output for a file it cannot fully resolve.
+  let info;
```
egonSchiele: `let info;` with no annotation makes this `any`, so the `typeCheck(program, ctx.config, info)` call below stops being checked. `let info: CompilationUnit;` keeps it.

Separately: `generateDocForFile` runs per file, so documenting a directory now builds one symbol table per file, each crawling and cloning the whole reachable set. `make doc` covers the whole standard library. Worth a quick measurement before merge, and if it bites, hoisting the table to the caller and passing it in would work — the crawl result is the same for every file in one run.

egonSchiele (commented):
## Review

This lands the design well. The three things I flagged on the plan as blocking or circular are all fixed and fixed properly: direct effects are read from the symbol rather than recomputed (which is what saves `_guard`), `checkUnhandledInterruptWarnings` got its own receiver resolution, and the oracle test now compares against `analyzeInterrupts` instead of comparing the symbol table to a type checker that was seeded from it. The test suite grew the cases that were missing — direction, no-union, graph nodes, imported nodes, unparseable files, `pkg::` survival, re-export hops, and the type-checker-finds-more invariant. The worklist rewrite of the fixpoint, caught by the scaling test, is a good catch.

What follows is one correctness issue, two scoping/consistency questions, a performance one, and some smaller notes. Nothing here blocks the core change; the correctness item is in the new generator check, which is the newest and least exercised surface.

### 1. `heldFunctionReference` refuses on ordinary variable aliasing (correctness)

Inline below. `const label = title` followed by `build(label)` is enough to refuse a generator, whatever the types are, with a message about passing a function reference. Fail-closed, but the argument for scoping blind spots to the call graph was precisely that a generator that is fine should not be refused.

### 2. The same function under-detects the case it exists for

Also inline: the argument scan only matches bare positional `variableName` arguments, so `f(cb: handler)` and `f(...refs)` slip through. Under-detection is the dangerous direction for a check whose whole job is refusing when the reading is incomplete. `positionalArg` in `interruptAnalysis.ts` already unwraps named and splat arguments.

### 3. One blind spot is scoped by imports, not by the call graph

`firstUnparseableImport` walks `closureFiles`, the import closure, while the other three blind spots walk `reachableFrom`, the call graph. The PR argues at length that file scoping is what would refuse every generator, so it is worth a sentence saying why this one is different (I assume: an unparseable file in the closure fails compilation anyway, so the over-refusal is unreachable). Related: `isAgencyFilePath` filters to relative paths, so an unparseable `pkg::` module is not detected — probably fine, worth being deliberate about.

### 4. Method calls no longer contribute a callee — is any real effect lost?

`calledName` returning `null` for every non-`invoke` method call is more correct than recording `map` from `xs.map(f)`, and `facts.calls` still feeds `functionRefsInArgs` so the type checker keeps the argument references. The question is whether any call written as `x.foo(...)` today resolves to a global `def foo` with effects — if so, those effects were being reported before this PR and are not now. If a test pins that under-reporting as intended, a pointer to it in the comment would settle it for the next reader.

### 5. Cost of the generator check on the paths that cannot pass a table

Threading `symbolTable` through `ExpandOptions` is right, but only `lib/compiler/typecheck.ts` passes one. `compile.ts:129` and `buildSession.ts:647` expand splices *before* the symbol table exists (deliberately — generated declarations must reach it), and `lsp/diagnostics.ts:139` is the same shape. Those paths pay a full `SymbolTable.build` plus a `programsFor` over every crawled file, per splice site. See the inline note on `programsFor`: `parseAgencyFileCached` `structuredClone`s the program on every read, so this is not a cheap cache hit. Hoisting the summaries to once per `expandSplices` call rather than once per splice would cover most of it.

### 6. Smaller things

- `checkGeneratorEffects` reports only `interruptEffects[0]`, so a generator that reaches three effects gets refused three times, once per fix. Listing them all costs nothing.
- `blindSpotIn` runs `declarationOf`, which scans the whole program per reached function. Fine at current sizes; a name-keyed index would drop the quadratic.
- The `stdlib/agents/planner.agency` corrections look right, and calling them a real defect rather than test churn is the right read.

### Test coverage

Strong, with two gaps worth a line each. There is no test for the fourth blind spot — a splice inside a *reached* file, as opposed to the generator's own file, which `checkNoNestedSplice` already covers. And AG3014/AG3015 have no cross-module test, though they are counted among the error-pushing consumers; the `.invoke()` exclusion there is documented at the site, which is the right call, but the non-`invoke` cross-module path is untested. The corpus tripwire has its own note inline.

packages/agency-lang/lib/compiler/splice/eligibility.ts
```diff
@@ -193,17 +199,203 @@ export function checkNoNestedSplice(
...
+ * cannot see through a compile-time splice, a function received as a parameter,
+ * a function reference held in a variable, or a file that did not parse. An
+ * empty list from a reading that saw none of those is not evidence of safety.
+ *
+ * Scoped to what the generator reaches BY CALLING. Every file reaches the
+ * prelude and passing a function as a value is ordinary Agency, so a
+ * file-scoped rule would refuse every generator, which is the same objection
+ * the comment on checkGeneratorEligible raises against a whole-closure test.
+ */
+export function checkGeneratorEffects(
```
egonSchiele: A lot of additions to this file, would it make sense to refactor the file at this point. I imagine that these functions are doing different jobs, and we've agreed to have one job per file. Would it make sense to pull some of these functions out into their own file?

packages/agency-lang/lib/analysis/effects.ts
```diff
@@ -0,0 +1,443 @@
...
+import type { FunctionDefinition } from "../types/function.js";
+import type { GraphNodeDefinition } from "../types/graphNode.js";
+import type {
+  FunctionSymbol,
+  NodeSymbol,
+  ResolvedImport,
+  SymbolTable,
+} from "../symbolTable.js";
+
+export type BodyFacts = {
```
egonSchiele: Big file! can be split up? Would it make sense to put all the body facts stuff into its own file, for example?

packages/agency-lang/lib/analysis/effects.ts
```diff
@@ -0,0 +1,443 @@
...
+ * `.invoke()` is the exception, because it means "call the receiver", and it
+ * is the call form this project's style prefers. See invokeReceiver.
+ */
+export function calledName(
+  node: FunctionCall,
+  ancestors: WalkAncestor[],
+): string | null {
+  const access = enclosingAccess(node, ancestors);
+  if (!access) return node.functionName;
+  return node.functionName === "invoke" ? invokeReceiver(node, access) : null;
```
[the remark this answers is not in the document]
egonSchiele: Verified, and you're right — I had this backwards. `AgencyFunction.invoke(descriptor: CallType)` is a TypeScript runtime method that generated code calls. It is not an Agency call form, and no `.agency` file in the repo uses it. Running it confirms your trace exactly:

```
double.invoke(21)
→ __callMethod (call.js:194) → AgencyFunction.invoke → resolveArgs → resolveNamed
TypeError: Cannot convert undefined or null to object
```

I had misread a note about codegen preferring `.invoke()` as being about source syntax. Removed the special case entirely, along with its tests and the claims in the PR body and dev doc.

The other half of that change stays, because it is real: a method call in an access chain now contributes no callee at all. Without it the walk records a phantom callee named `partial` or `map` and would attribute any same-named function's effects to a call site that never touches it. Fixed in eb139c602.

packages/agency-lang/lib/analysis/effects.ts
```diff
@@ -0,0 +1,443 @@
...
+export type Origin = { file: string; name: string };
+
+type EffectSummary = PropagationNode & Origin;
+
+/** Two files can define the same top-level name and an import can rename one,
+ *  so a bare name is not an identity. The key is never parsed back apart —
+ *  every summary carries its own file and name. */
+function keyOf(origin: Origin): string {
+  return `${origin.file} ${origin.name}`;
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in eb139c602 — NUL separator, which cannot occur in a path or an identifier, with a comment saying why a space could not be used.

packages/agency-lang/lib/compiler/splice/eligibility.ts
```diff
@@ -193,17 +199,203 @@ export function checkNoNestedSplice(
...
+  }
+
+  const effect = (symbol.interruptEffects ?? [])[0]?.effect;
+  if (effect !== undefined) {
+    return {
+      diagnostic: "spliceGeneratorRaises",
+      params: { name: generatorName, effect },
+      loc: { line: 0, col: 0, start: 0, end: 0 },
+    };
+  }
```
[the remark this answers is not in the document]
egonSchiele: Agreed, and the owner raised the same thing. Fixed in eb139c602: all effects, sorted and joined, matching how AG3013 reports what a raises clause exceeded.

packages/agency-lang/lib/compiler/splice/eligibility.ts
egonSchiele: Right, and your example refused. Fixed in eb139c602: an alias only counts when the name it aliases resolves to something callable — a function or node symbol, including through imports, or a parameter whose declared type is a function. `const label = title` on two strings now passes.

The import part mattered more than it looks. My first attempt only checked the file's own symbols, which made `const fn = read` stop being detected, because `read` arrives through the injected prelude import rather than being declared locally. Test added for the value-rename case.

packages/agency-lang/lib/compiler/splice/eligibility.ts
egonSchiele: Agreed that this is the dangerous direction. Fixed in eb139c602.

One correction on the pointer: `positionalArg` rejects splat and named arguments rather than unwrapping them — it returns null for both. The unwrapping was inline in `functionRefsInArgs`. I lifted that into `argumentExpression` next to `collectBodyFacts` as you suggested, and `functionRefsInArgs` now calls it too, so there is one unwrapper rather than two. Tests added for the named and splat forms.

packages/agency-lang/lib/compiler/splice/eligibility.ts
egonSchiele: Fixed in eb139c602, though not the way you suggested — tell me if you would rather have the threading.

The summaries cannot differ between splice sites in one run, since the crawl is finished before any of them are checked. So rather than thread `programs` through `DecisionContext`, I memoized against the `SymbolTable` with a WeakMap. Same effect, no new plumbing, and the entry dies with the table rather than living as run state. The per-site cost is now the reachability walk, as you described.

The threading is still the more explicit design and I will switch to it if you prefer it.

packages/agency-lang/lib/compiler/splice/eligibility.ts
egonSchiele: Fixed in eb139c602 — all effects, sorted and joined, matching AG3013's phrasing.

packages/agency-lang/lib/analysis/effects.test.ts
egonSchiele: Both fixed in eb139c602, and the stronger version immediately earned its keep.

It now collects the `functionCall` nodes from the raw scan and asserts each one is the very node `facts.calls` reports — identity, not a count — and does the same for `interruptStatement` against `facts.effects`. `agencyFilesUnder` walks `stdlib` recursively.

That found **83 real misses**, and they are not mine: `walkNodes` does not descend into `matchYield.typeSource`, so calls in that position are invisible to every analysis in the compiler, not just this one. Mostly `failure(...)` and `success(...)` in the data modules.

I did not fix it here — adding a descent changes what every walk sees and is not a speedy change. The tripwire names the gap in a `WALKER_GAPS` constant so it stays strict about everything else, and removing that entry is the test for a fix. Want an issue filed?

packages/agency-lang/lib/cli/doc.ts
egonSchiele: Both fixed in eb139c602. `let info: CompilationUnit;` restores checking on the `typeCheck` call.

On the per-file table: you were right that it was wasteful. The crawl result is the same for every file in one run once the entry is crawled, so I hoisted it onto `DocContext` and it is built on first use and reused. `make doc` regenerates identically afterwards, with no diff against the previous run.

packages/agency-lang/lib/compiler/splice/eligibility.ts
egonSchiele: Done in eb139c602. Pulled the effect check into `lib/compiler/splice/generatorEffects.ts`, which owns one question: can this generator reach something risky, and could we read enough to be sure. `eligibility.ts` keeps the other two rules — no nested splice, imports stay inside Agency — and exports the three helpers the new file shares.

packages/agency-lang/lib/analysis/effects.ts
egonSchiele: Done in eb139c602. `lib/analysis/bodyFacts.ts` now owns reading one body — the walk, `calledName`, `argumentExpression`, and the dedupe helpers. `lib/analysis/effects.ts` keeps what travels between bodies: the fixpoint, the summaries, the resolver, re-export following, and reachability.

The seam is the one you suggested, and it reads better than I expected: `bodyFacts` is a leaf that imports only the walker and type declarations, so the split is enforced by what it is allowed to depend on.

PR #698, opened 2026-07-25
Object patterns: `:` tests the type, `as` renames

### LINKED ISSUE #696: Pattern matching: make `{key: Type}` test the type, and move renaming to `as`
## Summary

Inside an object pattern, `{ name: string }` binds the `name` field to a variable *called* `string`. It does not test that the field is a string. Now that `#695` lets a `: Type` suffix nest, `[name: string]` and `{ name: string }` sit one bracket apart and mean opposite things.

Proposal: make `:` mean **type test** everywhere in a pattern, and move renaming to `as` — the keyword the language already uses for the same job in imports.

```ts
{ name: string }     // tests that the name field is a string   (today: binds a var called `string`)
{ id as catId }      // renames                                  (today: { id: catId })
{ name }             // unchanged
{ type: "click" }    // unchanged — a literal is not a type
```

## Why this is confusing today

Agency inherited two conventions that both spell themselves with `:`:

- **JS destructuring** — `{ key: newName }` renames.
- **Typed patterns** — `pattern: Type` tests the type, and runs the type's `@validate` chain.

Each is fine alone. They overlap in exactly one place — `{ key: identifier }` — and Agency resolves it "binder wins", with AG5004 warning about it.

After #695 the overlap is visible in a single match block:

```ts
match (cmd) {
  { words: ["cat", p: SafePath] } => read(p)   // `: SafePath` TESTS
  { path: SafePath }              => ...       // `: SafePath` BINDS a variable named SafePath
}
```

There is a spelling that does test a field — `{ name: n: string }`, which falls out of #695 because a property value is itself a pattern — but two colons in a row is not a thing anyone will guess, and it makes three near-identical spellings with three meanings.

AG5004's text is also now false. It says *"Field-level type tests are not supported — test the whole value with a typed pattern instead."* They are supported, via the two-colon form.

## How much code this affects

Parsed all 1258 `.agency` files in the repo and classified every object-pattern property.

**Match position** — 48 object patterns:

| form | count |
|---|---|
| `{ type: "click" }` — literal match | 37 |
| `{ name }` — shorthand bind | 21 |
| nested pattern | 5 |
| **`{ id: catId }` — rename bind** | **4** |

**Binding position** (`let` / `const` / `for`) — 13 object patterns: 13 shorthand, **4 renames**.

Six of those eight renames are in `tests/agency/destructuring/`, i.e. tests written to exercise the rename feature itself. Real-world uses in the whole repo: **two**, both the same construct in one file:

```ts
// stdlib/data/people/littlesis.agency:128 and :142
def buildRelationshipsPath(id: number, filter: CategoryFilter, sort: string): string {
  if (filter is { type: "category", id: catId }) {   // renamed because `id` is already a param
```

That is the genuine use case — avoiding a collision with a name already in scope. Real, but rare.

AG5004 fires zero times across stdlib.

## What other languages do

Every language that has both destructuring-rename and type patterns keeps them syntactically distinct. Most simply don't have both.

| language | destructure-rename | type pattern | collides? |
|---|---|---|---|
| Rust, Elixir, Python | `{key: name}` binds | none — types aren't testable in patterns | no |
| Scala | none — extractors are positional | `case Person(n: String)` tests **and** binds | no |
| Swift | none | `case let x as String` — uses `as`, leaving `:` free | no |
| TypeScript | `{key: name}` renames | none — the type goes on the whole binding | no |
| **Agency** | `{key: name}` binds | `pattern: Type` tests | **yes** |

Scala is the closest precedent for the proposed reading: a nested `n: String` tests the type and binds. It can afford `:` because it never adopted JS-style renaming. Swift has the opposite problem and solves it the way this proposal does — by spending `as` on the rename.

## Proposal

1. `:` in a pattern always introduces a **type**. `{ name: string }` tests the field.
2. Renaming moves to `as`: `{ id as catId }`.
3. `{ name }` shorthand and `{ type: "click" }` literal matching are unchanged.
4. AG5004 is **deleted** — the footgun stops existing rather than being warned about.
5. The two-colon form `{ name: n: string }` stops being reachable, and needs no explaining.

`as` is not new vocabulary: it is already a reserved word (`lib/parsers/parsers.ts:217`) and already the rename spelling in imports (`import { foo as bar }`, `parsers.ts:3988`).

## What breaks

**Loudly, in source — 8 sites.** A renamed field would read its right-hand side as a type, and there is already a diagnostic for a name that is not one: AG1013, *"`catId` is not a type"*. No silent behavior change: literals and shorthand are untouched, and a rename becomes a compile error rather than a different meaning.

Match position:
- `stdlib/data/people/littlesis.agency:128` — `{ type: "category", id: catId }`
- `stdlib/data/people/littlesis.agency:142` — `{ type: "category", id: catId }`
- `tests/agency/destructuring/destructuringInterrupt.agency` — `{ x: ex }`, `{ y: ey }`

Binding position:
- `tests/agency/destructuring/destructuring.agency` — `{ name: n }`, `{ age: ageRenamed }`, `{ name: n2 }`
- `tests/agency/destructuring/destructuringInterrupt.agency` — `{ name: n }`

**Silently, in generated code — 1 site, and this is the one to be careful about.** Template hygiene builds the rename form programmatically. `expandShorthand` (`lib/runtime/template/hygiene.ts:366`) turns a colliding shorthand `{ tmp }` into `{ tmp: __hygN_tmp }`, because renaming a shorthand in place would change which property is read:

```ts
return {
  type: "objectPatternProperty",
  key: name,
  value: { type: "variableName", value: map[name] },
};
```

Under this proposal that AST shape means "test the field against type `__hygN_tmp`". It would not throw a parse error — it never goes through the parser — so hygiene would quietly emit a type test where it meant a rename. Capture avoidance failing open is exactly the bug hygiene exists to prevent, so this needs the same treatment as the source sites: hygiene must emit whatever node the new rename form produces.

## Work items

- Parser: `{ key: <type> }` in pattern position; `{ key as name }` for rename.
- AST: the rename needs its own node (or an explicit marker on `objectPatternProperty`) so it cannot be confused with a type test by anything that constructs patterns programmatically.
- `lib/runtime/template/hygiene.ts:366` — emit the new rename node.
- Delete AG5004 and its tests (`lib/typeChecker/diagnostics.ts`, `matchExhaustiveness.ts`).
- Formatter: print `as`; round-trip test.
- Migrate the 8 sites above.
- Docs: `pattern-matching.md` (destructuring section at `:35` shows `const { name: n, age: ageRen } = person`), `basic-syntax.md`, `templates.md:154` which describes the `{ key: __hyg1_key }` expansion by name.
- The `patternGuards` tripwire corpus should keep passing throughout; it covers nested pattern positions already.

## Open sub-decision: does binding position change too?

`const { name: n } = person` has no type-test meaning — `: Type` there is a static annotation on the whole binding — so binding position *could* keep `:` for rename.

Recommendation: **change both.** Keeping `:` in `const` and `as` in `match` trades one inconsistency for another, and the extra cost is 4 sites, all in destructuring tests.

## The honest cost

This diverges from JavaScript and TypeScript, which the guide leans on explicitly. Someone arriving from JS will write `{ name: n }` and get an error. That is a tax paid by every new user, weighed against a footgun paid by everyone who tries to type-check a field — and the error is loud, specific, and has an obvious fix.

## Related

- #695 — nested `: Type` suffix, which is what makes the inconsistency visible
- #676, #677 — the pattern-matching fixes that preceded it
- `docs/site/guide/pattern-matching.md` — "One footgun", which documents the current behavior

egonSchiele:
Closes #696.

## The change

```ts
{ name: string }   // tests that the name field is a string   (was: bound a variable CALLED string)
{ name as n }      // renames                                  (was: { name: n })
{ name }           // unchanged
{ type: "click" }  // unchanged — a literal is not a type
```

Two inherited conventions were sharing `:` — JS destructuring rename and typed patterns — with "binder wins" as the tiebreak. After #695 let the `: Type` suffix nest, `[name: string]` and `{ name: string }` sat one bracket apart and meant opposite things. Now `:` always introduces a type in a pattern, and renaming uses `as`, which is already the language's rename keyword (`import { foo as bar }`).

## Why the diff is small

A property **value** now parses with the is-RHS rule — a bare identifier is a type reference — which is the rule that already applied after `is`. And `key as name` produces the **same** `objectPatternProperty` node the old spelling did.

So binding, lowering, and template hygiene need no changes. That last one matters: the issue flagged `expandShorthand` (`hygiene.ts:366`) as the one place the break would be *silent*, because it constructs `{ tmp: __hygN_tmp }` programmatically. Since the rename node is unchanged, it keeps working and simply prints with `as`. No new AST node, no walker updates.

The formatter learned two things:

- a binder value prints with `as` — printed with a colon it would come back as a type test against a type named after the binder
- a property value follows the is-RHS rule, so `{ name: string }` round-trips exactly instead of becoming `{ name: _: string }`

**AG5004 is deleted**, not reworded — the trap it warned about no longer exists.

## Two judgment calls worth reviewing

**1. Binding position is untouched.** `let { name: n } = person` still renames. No type test is possible in a `let` — `: Type` there is a static annotation on the whole binding — so nothing is ambiguous and the footgun cannot occur. `as` works in both positions.

The issue recommended changing both for consistency. I took the smaller break instead: the inconsistency is real but invisible in practice, since the ambiguous case only exists where types are testable. Happy to change it — 4 sites, all in destructuring tests.

**2. Three sites migrated, not eight**, for the same reason — only match-position renames needed it:

- `stdlib/data/people/littlesis.agency:128` and `:142` — `{ type: "category", id as catId }`
- `tests/agency/destructuring/destructuringInterrupt.agency:33` — `{ type: "click", x as ex, y as ey }`

## Verification

Local, spot checks only — full suites left to CI:

- `{ name: string }` matches `{name: "Ada"}` and rejects `{name: 42}`; `{ name as n }` binds.
- Formatter round-trips both spellings exactly.
- `tsc --noEmit` clean; no dangling AG5004 references.

## Note on the history

The first commit here included some unrelated working-tree files, swept in by an over-broad `git add`. The second commit removes them from tracking (`git rm --cached`, untouched on disk). Net diff is the 8 files above.


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -6158,9 +6187,11 @@ export const arrayMatchPatternParser: Parser<ArrayPattern> = label(
   },
 );
 
+// A property value uses the is-RHS rule: a bare identifier is a TYPE, not a
+// binder. `{ name: string }` tests the field; `{ name as n }` renames it.
 const _matchObjectPropertyParser: Parser<
   ObjectPatternProperty | ObjectPatternShorthand | RestPattern
-> = objectPatternPropertyParser(lazy(() => matchPatternParser));
+> = objectPatternPropertyParser(lazy(() => isRhsParser));
```
egonSchiele, line 6194: This one line is the whole break, and it is the line that needs a migration diagnostic behind it.

Switching the value parser to `isRhsParser` means every existing `{ key: binder }` becomes a type reference. Today that produces no compile error and a runtime `ReferenceError` — see item 1. The generated code is literally `__validateType(event.x, ex)` with `ex` undefined.

Arm position already errors well for this (AG1013). Extending that check to property-position type patterns, with a message that says "write `{ x as ex }`", would turn the worst failure mode in this PR into the best one.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -5924,11 +5924,40 @@ const propertyWithValueParser = (
...
+  return parser(input) as ParserResult<ObjectPatternProperty>;
+};
+
 const objectPatternPropertyParser = (
   valueParser: Parser<ObjectPatternProperty["value"]>,
 ): Parser<ObjectPatternProperty | ObjectPatternShorthand | RestPattern> =>
   or(
     restPatternParser,
+    // rename MUST be tried before shorthand, which would otherwise take `key`
+    // and leave `as name` stranded
```
egonSchiele, line 5959: Worth noting in the comment that `objectPatternPropertyParser` is shared with **binding** position, so wiring the rename in here makes `let { name as n } = p` legal too — which is intended per judgment call 1, but is not obvious from a comment that only talks about shorthand ordering.

The ordering itself is right, and I checked the case that worried me: a field literally named `as` (`{ as: number }`) still parses, because the rename parser fails on the `:` and falls through.

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
```diff
@@ -252,9 +252,11 @@ const SHADOWABLE_PRIMITIVE_NAMES: readonly string[] = [
  * Warn when a binder is named like a type, because with type patterns in the
  * language it READS like a type test but binds instead:
  *   - an un-guarded bare-binder arm (`Person => ...` matches anything and
- *     binds it to `Person`) → AG5003;
- *   - a property-position binder (`{name: string}` binds the `name` field to
- *     a variable called `string`) → AG5004, guarded or not.
+ *     binds it to `Person`) → AG5003.
+ *
+ * The property-position case (AG5004) is gone: `{name: string}` now TESTS the
```
egonSchiele, line 257: "The trap it warned about no longer exists" is true for match position, but three tests on this branch still assert AG5004 fires and will fail CI:

- `typePatterns.test.ts:177` → `:188`
- `typePatterns.test.ts:220` → `:230`
- `typePatterns.test.ts:233` → `:242`

Those tests are the specification of the behavior you are changing, so inverting them ("`{name: string}` now tests the field") is most of the coverage this PR is missing.

Also still live: `docs/site/diagnostics/index.md:88` and `match.md:39`, plus the `stdlib/docs/` copies — regenerated by `make diagnostics-docs`.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -989,7 +989,16 @@ export class AgencyGenerator {
       if (prop.value.type === "variableName" && prop.value.value === prop.key) {
         return prop.key;
       }
-      return `${prop.key}: ${this.formatPattern(prop.value)}`;
+      // A binder value is a RENAME, and renames are spelled with `as`. Printed
+      // with a colon it would come back as a type test against a type named
+      // after the binder.
+      if (prop.value.type === "variableName") {
+        return `${prop.key} as ${prop.value.value}`;
```
egonSchiele, line 996: `formatObjectPattern` is shared with binding patterns, so this branch fires there too. Verified on this build:

```ts
let { name: n, age } = p     // in
let { name as n, age } = p   // out of `pnpm run fmt`
```

Harmless in the sense that both spellings parse in binding position — but it means the formatter is migrating the position judgment call 1 explicitly chose not to migrate. See item 5.

packages/agency-lang/docs/site/guide/pattern-matching.md
```diff
@@ -468,9 +468,8 @@ constrain fields, type the object:
 { name, age }: Person                            => ...   // yes
 ```
 
-Remember that inside an object pattern `{ name: string }` binds the `name`
-field to a variable *called* `string` — it does not test the field's type
-(the checker warns with AG5004). Type the whole object instead.
+Inside an object pattern the same rule applies — `{ name: string }` tests the
```
egonSchiele: "Inside an object pattern the same rule applies" is true in match position and false in `let` / `const` / `for`, where `{ name: string }` still binds a variable called `string`.

Since this PR also deletes the "One footgun" section and AG5004, that trap ends up both undocumented and unwarned. Worth naming the position explicitly here — or migrating binding position too, which makes the sentence true as written (item 6).

egonSchiele (commented):
## Overview

`:` always introduces a type, `as` renames. The rule is right, `as` is the correct keyword for it, and reusing the existing `objectPatternProperty` node so binding/lowering/hygiene need no changes is a genuinely elegant way to land a syntax break. I verified the feature works — `{ name: string }` tests the field, `{ name as n }` binds, a field literally named `as` still parses.

But this is a breaking change to how existing source parses, and the migration story has a hole that I think blocks it. There are also three tests on this branch that will fail on CI.

## 1. Old-spelling code compiles clean and crashes at runtime

This is the one that worries me. Take the code this PR migrated in `destructuringInterrupt.agency`, unmigrated:

```ts
node main() {
  let event = { type: "click", x: 10, y: 20 }
  let clickX = 0
  if (event is { type: "click", x: ex, y: ey }) {
    clickX = ex
  }
  print(clickX)
  return "done"
}
```

Compiles with **no diagnostic at all**. The generated condition is:

```js
… && isSuccess(__validateType(__stack.locals.event.x, ex))
   && isSuccess(__validateType(__stack.locals.event.y, ey))
```

`ex` and `ey` are free identifiers that do not exist, so:

```
ERROR Node main crashed: ex is not defined
ERROR ReferenceError: ex is not defined
```

So every existing `{ key: binder }` in match position silently becomes a validation against a nonexistent type, and the user finds out at runtime, from a `ReferenceError` naming a variable they never wrote as a variable. For a syntax break that is the worst available failure mode — worse than the AG5004 footgun this replaces, which at least warned.

The diagnostic machinery exists and works one level up. Arm position gives a good error:

```
error AG1013: `NoSuchType` is not a type; to bind the value write `const NoSuchType = x`.
```

It just doesn't reach property-position type patterns — the same nested-typechecker gap I flagged on #695, which had no consequences there and has serious ones here.

Two things needed, I think:

- **Make the unknown-type check reach nested property positions.** This should be an error at compile time, not a `ReferenceError` at run time.
- **Give it a message that names `as`.** AG1013's current advice ("to bind the value write `const NoSuchType = x`") is wrong for this case. What a migrating user needs is: "`{ x: ex }` tests the `x` field against a type named `ex`; to bind the field to `ex`, write `{ x as ex }`."

Worth noting this also means the PR's own migration was found by hand with nothing to check the work. Three sites were migrated; if a fourth had been missed, nothing in the build would have said so.

## 2. Three tests on this branch assert AG5004 and will fail

`lib/typeChecker/typePatterns.test.ts` on `adit/696`:

- `:177` — "property-position binder named like a type warns AG5004" → `expect(...).toBe(true)` at `:188`
- `:220` — "typed arms warn AG5004 for property binders named like types" → `:230`
- `:233` — "nested object patterns warn AG5004" → `:242`

The first also breaks a second way: its arm is `{name: string} => string`, returning the variable the old spelling bound. Under the new rule nothing is bound, so the arm body is an unbound identifier.

These are exactly the tests that should have been rewritten as "`{name: string}` now tests the field" — they are the specification of the behavior being changed, so converting them is most of the test coverage this PR needs.

## 3. AG5004 is still in the generated diagnostics docs

The description says "no dangling AG5004 references." On this branch:

- `docs/site/diagnostics/index.md:88`
- `docs/site/diagnostics/match.md:39`
- the `stdlib/docs/diagnostics/` copies of both

Those are generated by `make diagnostics-docs` from the registry (`Makefile:137`) and copied into `stdlib/docs` (`:37`), so `make` regenerates them. #676, #677 and #695 all reported `make` clean; this one reports only `tsc --noEmit`, which is why these and item 2 slipped through. Running the usual `make` + suite would catch all three.

## 4. The guide now says something false about binding position

> Inside an object pattern the same rule applies — `{ name: string }` tests the field.

True in match position only. In `let` / `const` / `for`, `{ name: string }` still binds a variable called `string` — the precise footgun AG5004 described. After this PR that footgun is undocumented *and* unwarned, since both the guide's "One footgun" section and the diagnostic are deleted.

Given judgment call 1 keeps binding position as-is, the docs have to say which position they are describing.

## 5. The formatter migrates binding position, which judgment call 1 said not to

`formatObjectPattern` is shared between binding and match patterns, so the new `as` branch applies to both. Verified on this build:

```ts
let { name: n, age } = p     // in
let { name as n, age } = p   // out of `pnpm run fmt`
```

It round-trips, since binding position accepts both spellings — but it means `fmt` rewrites source in exactly the position the PR decided to leave alone. Pick one: either the formatter is doing the migration (in which case judgment call 1 is really "the parser stays permissive"), or binding position keeps printing the colon.

## 6. On judgment call 1

I'd push toward migrating binding position too. The PR trades one confusion for another:

- Before: `[name: string]` and `{name: string}` are one bracket apart and mean opposite things.
- After: `let { name: n }` and `is { name: n }` are *identical text* and mean opposite things, differing only by position.

And given item 1, the dangerous copy direction is match → binding: `is { x: Person }` pasted into a `let` silently binds a variable called `Person`, with AG5004 now deleted. That is the same silent trap, just relocated.

The title claims a clean rule — `:` tests, `as` renames. It'd be nice if it were true everywhere. You say it's four sites; I'd spend them.

## 7. No tests

Zero added, for a breaking syntax change. #676, #677 and #695 were all test-first, and their test files are most of why they read as trustworthy. The minimum here:

- parser: `{ name: string }` is a typePattern; `{ name as n }` is a rename; `{ name: n }` is now a type reference; a field named `as` still parses
- formatter: both spellings round-trip, and re-formatting is a fixpoint
- execution: an `.agency` test for test-the-field and rename (this is 15 lines, and it is the feature)
- typechecker: item 1 — the old spelling produces a diagnostic, not a `ReferenceError`
- the three AG5004 tests, inverted

## Verdict

The design is right and I want it in. Item 1 I'd treat as blocking — a silent compile followed by a runtime `ReferenceError` is not an acceptable migration path for a syntax break, and it is the difference between "the language changed" and "my code broke mysteriously." Items 2 and 3 are mechanical and would have been caught by the `make` + suite run the earlier PRs did. Items 4–7 are worth doing before this lands, since it is a change people will have to learn once and live with.

egonSchiele (commented):
CI is green. Addressed below — and one of your items I have to push back on, with evidence.

## Item 1: AG1013 already fires — but you are right about the message

I could not reproduce "no compile error and a runtime `ReferenceError`". Both positions error at compile time on this branch:

```ts
match (p) { { name: n } => ... }                      // AG1013: `n` is not a type
if (event is { type: "click", x: ex, y: ey }) { ... }  // AG1013 x2, on ex and ey
```

So the worst failure mode is already the best one — the migration is loud. If you saw a `ReferenceError`, I would like the repro, because that would mean a path where the typechecker does not run and I have not found it.

What was genuinely wrong is the **hint**. It said:

> to bind the value write `const ex = x`

which is the fix for arm position and useless in property position. It now also names the object form: ``{ field as ex }``. That was the actionable half of your item, and it is fixed.

## Items 2 and 4: docs and tests

- `docs/site/diagnostics/index.md` and `match.md` regenerated with `make diagnostics-docs`; AG5004 is gone from both.
- The three `typePatterns.test.ts` AG5004 tests were deleted in 8d8a6a5c2, before your review landed. You are right that they were the specification of the behavior — so the replacement is `tests/agency/pattern-matching/objectPatternAsRename.agency` (5 cases: type test picks the arm, a named type runs its validators, renames bind, nested type tests, and all three property forms in one pattern) plus four parser tests.

## Item 3: `objectPatternPropertyParser` is shared with binding position

Correct, and now said in the comment. Also glad you checked `{ as: number }` — I tested the mirror case, `{ as as v }`, and both parse. There is a test for it now.

## Item 5: the formatter migrates binding position

Confirmed, and I am leaving it, deliberately:

```ts
let { name: n, age } = p     // in
let { name as n, age } = p   // out
```

Both spellings parse in binding position, so this is a normalization rather than a break — the formatter picking one spelling, the way it picks any other. And it is the spelling that is correct everywhere, which seems better than emitting a form that is only correct in half the language.

It is a real inconsistency with judgment call 1 though, and worth saying out loud: the **parser** keeps `:` working in binding position, the **formatter** rewrites it. If you would rather the formatter left binding position alone, that is a small change.

The four hygiene tests are what pinned this. Their output is now `{ key as __hygN_key }`, and the match-arm one says why in full: a colon there introduces a type, so the old spelling would turn a rename into a test against a type named `__hygN_name` — capture avoidance failing open, which is the bug hygiene exists to prevent. That was the silent break the issue predicted, and it is avoided by construction.

## Item 6: the guide sentence

Fixed. It now names the position and documents the difference rather than leaving `let { name: string }` both undocumented and unwarned:

> Inside an object pattern in a **match** the same rule applies. In `let` / `const` / `for`, where no type test is possible, `:` still renames.

## Also worth knowing

`{ name, ...rest }` in a match pattern fails with `AG4004: Function '__objectRest' is not defined`. Pre-existing — I confirmed it on main with this branch's changes stashed. Not touched here; say the word and I will file it.

PR #699, opened 2026-07-25
Stop reporting the synthetic `__objectRest` call as an undefined function

egonSchiele:
## The bug

`{ a, ...rest }` lowers to `__objectRest(source, ["a"])`, a synthetic call the TypeScript builder compiles away. The undefined-function check runs **between** those two passes, so it saw a call with no declaration.

**Impact, corrected.** An earlier version of this description said object rest was "unusable" and "broken". That was wrong, and I verified it against an unfixed build:

```
$ agency run rest.agency        # default config
warning AG4004: Function '__objectRest' is not defined.
RAN: x                          # ...and it runs
```

`undefinedFunctionDiagnostic.ts:36` defaults the mode to `"warn"`, so by default this was a **spurious warning on every object-rest destructure**, not a failure. It is a hard error only for a project that opted into `undefinedFunctions: "error"`.

Still worth fixing — a false positive on a core language feature is how people learn to ignore a diagnostic, which costs more than the warning does — but it is not urgent, and the description should have said so.

Not match-specific either. I found it in a match arm, but binding position and loop binders fail identically.

## The fix

`FunctionCall` gains `synthetic?: boolean`. The lowerer sets it on calls it creates; the check skips them.

A field rather than a name, following what the lowerer already does — `matchExprId`, `matchArmValueTemp` and `matchSource` are all fields set for exactly this "a later pass needs to recognize what I made" reason. Two consequences: user code that happens to define `__objectRest` cannot spoof the skip, and the next synthesized call is covered without its author discovering a registry inside a diagnostic file.

`OBJECT_REST_FN` still moves to `lib/constants.ts`, since the lowerer and builder both need the name.

## Tests

Written before the fix; the typechecker cases failed on `main`.

- `undefinedFunctionDiagnostic.test.ts` — no AG4004 for object rest in binding, match and for-loop position, and **a genuinely undefined function is still reported**. That last one matters: the fix suppresses a diagnostic, and a suppression that goes too far passes the positive tests happily while silently disabling AG4004.
- `tests/agency/destructuring/objectRest.agency` — runtime behavior across all five positions, plus one asserting the rest binder holds the keys the pattern did not name and not the ones it did.

Each test was checked against the failure it is supposed to catch, not assumed:

| break | result |
|---|---|
| drop `synthetic: true` in the lowerer | typechecker tests 3 failed |
| disable the `OBJECT_REST_FN` branch in the builder | execution tests 0/5 |
| revert the fix entirely | execution tests still **5/5** — see below |

That last row is why the two files say different things now. AG4004 is a warning in the test runner, so the execution tests run fine even when the diagnostic wrongly fires: they guard the **builder** half, not the diagnostic. The file says so, and points at the typechecker tests for the other half. I had originally claimed it covered both.

### DIFF
--- a/packages/agency-lang/lib/constants.ts
+++ b/packages/agency-lang/lib/constants.ts
@@ -66,3 +66,16 @@ export const AGENCY_MAX_TIME = "AGENCY_MAX_TIME";
 /** Process exit code when a top-level cost/time budget is exceeded. Distinct
  *  from 1 (generic failure) and 2 (usage error). */
 export const EXIT_CODE_BUDGET_EXCEEDED = 3;
+
+/**
+ * The synthetic function pattern lowering emits for an object rest binder:
+ * `{ a, ...rest }` becomes `__objectRest(source, ["a"])`. The TypeScript
+ * builder compiles the call away, so it never reaches runtime.
+ *
+ * Shared because three passes have to agree on it and the typechecker runs
+ * between the two that used to hold it as a literal: the lowerer emits it, the
+ * builder matches it, and the undefined-function check has to know it is not a
+ * user call. When only two of the three knew the name, object rest in any
+ * pattern reported "Function '__objectRest' is not defined".
+ */
+export const OBJECT_REST_FN = "__objectRest";
--- a/packages/agency-lang/lib/typeChecker/undefinedFunctionDiagnostic.ts
+++ b/packages/agency-lang/lib/typeChecker/undefinedFunctionDiagnostic.ts
@@ -10,6 +10,7 @@ import {
   isJsGlobalBase,
 } from "./resolveCall.js";
 import { collectProgramShadowing } from "./shadowing.js";
+import { OBJECT_REST_FN } from "../constants.js";
 
 /**
  * Emit a diagnostic for every call site that doesn't resolve to a known
@@ -85,6 +86,12 @@ function hasFunctionOrNodeAncestor(ancestors: readonly unknown[]): boolean {
 // this is just the set of keywords that trigger it.
 const RESERVED_BLOCK_KEYWORDS = ["thread", "subthread"];
 
+// Calls the pattern lowerer synthesizes. They are not user code and have no
+// declaration to find — the TypeScript builder compiles them away — but this
+// pass runs after lowering, so it sees them and would report every one as an
+// undefined function.
+const SYNTHETIC_CALLS: readonly string[] = [OBJECT_REST_FN];
+
 function checkBareCall(
   call: FunctionCall,
   scope: Scope,
@@ -92,6 +99,7 @@ function checkBareCall(
   mode: "warn" | "error",
   importedNodeNames: readonly string[],
 ): void {
+  if (SYNTHETIC_CALLS.includes(call.functionName)) return;
   const resolution = resolveCall(call.functionName, {
     functionDefs: ctx.functionDefs,
     nodeDefs: ctx.nodeDefs,
--- a/packages/agency-lang/lib/typeChecker/undefinedFunctionDiagnostic.ts
+++ b/packages/agency-lang/lib/typeChecker/undefinedFunctionDiagnostic.ts
@@ -10,7 +10,6 @@ import {
   isJsGlobalBase,
 } from "./resolveCall.js";
 import { collectProgramShadowing } from "./shadowing.js";
-import { OBJECT_REST_FN } from "../constants.js";
 
 /**
  * Emit a diagnostic for every call site that doesn't resolve to a known
@@ -86,11 +85,6 @@ function hasFunctionOrNodeAncestor(ancestors: readonly unknown[]): boolean {
 // this is just the set of keywords that trigger it.
 const RESERVED_BLOCK_KEYWORDS = ["thread", "subthread"];
 
-// Calls the pattern lowerer synthesizes. They are not user code and have no
-// declaration to find — the TypeScript builder compiles them away — but this
-// pass runs after lowering, so it sees them and would report every one as an
-// undefined function.
-const SYNTHETIC_CALLS: readonly string[] = [OBJECT_REST_FN];
 
 function checkBareCall(
   call: FunctionCall,
@@ -99,7 +93,9 @@ function checkBareCall(
   mode: "warn" | "error",
   importedNodeNames: readonly string[],
 ): void {
-  if (SYNTHETIC_CALLS.includes(call.functionName)) return;
+  // A call the lowerer synthesized, not one the user wrote. It has no
+  // declaration to find, and the builder compiles it away before runtime.
+  if (call.synthetic) return;
   const resolution = resolveCall(call.functionName, {
     functionDefs: ctx.functionDefs,
     nodeDefs: ctx.nodeDefs,
--- a/packages/agency-lang/tests/agency/destructuring/objectRest.agency
+++ b/packages/agency-lang/tests/agency/destructuring/objectRest.agency
@@ -25,6 +25,26 @@ node restInBinding() {
   return "${name} lives in ${rest.city}"
 }
 
+def firstOf(people: any[]): string {
+  // Loop binder: a third lowering site, and the one most likely to be missed.
+  let out = "none"
+  for ({ name, ...rest } in people) {
+    out = "${name}/${rest.city}"
+  }
+  return out
+}
+
+node restInNestedPattern() {
+  // extractBindings recurses, so the synthetic call is emitted at depth too.
+  const o = { outer: { a: 1, b: 2, c: 3 } }
+  const { outer: { a, ...inner } } = o
+  return [a, inner.b, inner.c]
+}
+
+node restInForLoopBinder() {
+  return firstOf([{ name: "Ada", city: "NY" }])
+}
+
 node restExcludesNamedKeys() {
   // The rest binder holds the keys the pattern did not name, and not the ones
   // it did.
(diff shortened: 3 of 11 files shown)


packages/agency-lang/lib/typeChecker/undefinedFunctionDiagnostic.ts
```diff
@@ -85,13 +86,20 @@ function hasFunctionOrNodeAncestor(ancestors: readonly unknown[]): boolean {
 // this is just the set of keywords that trigger it.
 const RESERVED_BLOCK_KEYWORDS = ["thread", "subthread"];
 
+// Calls the pattern lowerer synthesizes. They are not user code and have no
+// declaration to find — the TypeScript builder compiles them away — but this
+// pass runs after lowering, so it sees them and would report every one as an
+// undefined function.
+const SYNTHETIC_CALLS: readonly string[] = [OBJECT_REST_FN];
+
```
egonSchiele, line 88: The comment is good — it explains the pipeline ordering that causes the problem, which is the part a reader cannot infer from the code.

One thought on the mechanism: this identifies a synthetic call by its *name*, where the rest of the codebase identifies lowerer-synthesized nodes by a *field* — `matchArmValueTemp`, `matchExprId`, `matchSource` are all set by the lowerer for exactly this "a later pass needs to recognize what I made" reason.

A `synthetic?: true` on `FunctionCall` would mean the next lowerer-emitted call is handled without anyone needing to discover that this list exists. With one entry the list is fine; I mention it because "somewhere obvious to go" is the argument for the flag rather than for the list.

egonSchiele (commented):
## Overview

A synthetic call the lowerer emits was being reported as undefined by a pass that runs between the lowerer and the builder. The fix hoists the name into a shared constant and skips it in the check. Small, correct, and the diagnosis in the description — "the third pass that needed to know had no way to find out" — is exactly right.

The third test is the one that makes this trustworthy. A skip-list fix that only tested the two positive cases would pass just as happily while skipping everything, and you wrote the control that rules that out. Same for `restExcludesNamedKeys`: "compiling is not the same as being correct" is the right instinct, and checking that `rest.name` is absent is the assertion that actually pins the semantics.

I checked for the obvious sibling — `comprehensionDesugar.ts:228` emits `_pairsOf`, which looked like the same shape — but that one is a real prelude function (`prelude.ts:32`, `stdlib/builtins.ts:359`), so it resolves normally. No second instance to fix. Worth knowing that `PRELUDE_NAMES` was not an option for `__objectRest`: its docstring says every name there must be exported by `stdlib/index.agency`, and `__objectRest` has no runtime existence at all. So the skip list is the right shape, not a shortcut around an existing registry.

## The description overstates the impact

`undefinedFunctionDiagnostic.ts:36` defaults the mode to `"warn"`:

```ts
const mode = ctx.config.typechecker?.undefinedFunctions ?? "warn";
```

Both new typechecker tests set `undefinedFunctions: "error"` explicitly. So unless a project opted into `error`, this was a spurious warning on every object-rest destructure — noisy and wrong, but the code still compiled and ran. "Object rest in a pattern was unusable" and "a plain `const { a, ...rest } = o` being broken" read as a hard failure.

Still worth fixing — a false positive on a core language feature trains people to ignore AG4004, which is worse than the warning itself. But the description should say which mode it breaks in, since that is what tells a reader whether this needs to land today.

(I could not verify this against an unfixed build — the local `dist/` already contains the fix — so correct me if a `warn` here does gate something downstream.)

## Suggestions

### 1. Consider marking the node instead of matching the name

The skip is a string comparison, so a synthetic call is identified by what it is called rather than by where it came from. This codebase already marks lowerer-synthesized nodes structurally — `matchArmValueTemp`, `matchExprId` and `matchSource` are all fields the lowerer sets on nodes it creates, precisely so later passes can recognize them without pattern-matching on content.

A `synthetic?: true` on `FunctionCall` would follow that convention: it cannot be spoofed by user code that happens to call something `__objectRest`, and — more to the point — the next lowerer author gets it for free instead of having to know that a registry in `undefinedFunctionDiagnostic.ts` exists and needs updating. That is the failure mode your comment is worried about, and a flag closes it where a list only documents it.

With one entry the list is defensible and I would not block on this. But the argument in the description ("somewhere obvious to go") is the argument for the flag.

### 2. `SYNTHETIC_CALLS` could live next to the name it holds

`OBJECT_REST_FN` moved to `constants.ts` so three passes could share it, but `SYNTHETIC_CALLS` — the concept — stayed private to the diagnostic file. If the intent is that the next synthetic call has an obvious home, that home is more discoverable next to the names themselves than inside the one pass that currently consumes it.

### 3. The test file name says `InMatch`, but two of its three nodes are binding position

`objectRestInMatch.agency` in `tests/agency/destructuring/`. The whole point of the description's second paragraph is that this is *not* match-specific — `restInBinding` and `restExcludesNamedKeys` are both `const` destructuring. `objectRest.agency` would match what the file actually covers, and would not send the next person looking for the binding-position case somewhere else.

### 4. Two positions the same call reaches, untested

`extractBindings` recurses, so `__objectRest` is emitted at depth and in loop binders too:

```ts
const { a: { b, ...inner } } = o        // nested
for ({ name, ...rest } in people) { }   // for-loop binder
```

The same fix covers both, so these are cheap regression coverage rather than new risk. Worth one line each in the `.agency` file if you are touching it for the rename anyway.

## Verdict

Good fix, right layer, and the negative control is what makes it reviewable. Nothing blocking. I would tighten the impact claim in the description, since it is the part a reader uses to decide how to prioritize this, and rename the test file.

packages/agency-lang/lib/constants.ts
```diff
@@ -66,3 +66,16 @@ export const AGENCY_MAX_TIME = "AGENCY_MAX_TIME";
...
+/**
+ * The synthetic function pattern lowering emits for an object rest binder:
+ * `{ a, ...rest }` becomes `__objectRest(source, ["a"])`. The TypeScript
+ * builder compiles the call away, so it never reaches runtime.
+ *
+ * Shared because three passes have to agree on it and the typechecker runs
+ * between the two that used to hold it as a literal: the lowerer emits it, the
+ * builder matches it, and the undefined-function check has to know it is not a
+ * user call. When only two of the three knew the name, object rest in any
+ * pattern reported "Function '__objectRest' is not defined".
```
egonSchiele, line 79: Since `OBJECT_REST_FN` moved here so three passes could share it, consider moving `SYNTHETIC_CALLS` here too. Right now the name is shared but the *category* is private to `undefinedFunctionDiagnostic.ts`, which is the less discoverable half.

Worth noting in this comment why `PRELUDE_NAMES` was not the answer, since it is the first place someone will look for "names the checker should know about": `prelude.ts:25` requires every name there to be exported by `stdlib/index.agency`, and `__objectRest` has no runtime existence to export. That rules it out cleanly and saves the next reader the detour.

packages/agency-lang/tests/agency/destructuring/objectRest.agency
```diff
@@ -0,0 +1,29 @@
...
+    _ => "other"
+  }
+}
+
+node restInMatchArm() {
+  return describe({ name: "Ada", age: 36, city: "NY" })
+}
+
+node restInBinding() {
+  const o = { name: "Ada", age: 36, city: "NY" }
```
egonSchiele, line 23: Two of the three nodes here are binding position, which is the description's own point — this was never match-specific. `objectRestInMatch.agency` will send the next person looking for the `const { a, ...rest }` case somewhere else. `objectRest.agency` covers what the file actually tests.

If you are renaming anyway, two more positions the same synthetic call reaches and nothing covers:

```ts
const { a: { b, ...inner } } = o        // nested — extractBindings recurses
for ({ name, ...rest } in people) { }   // loop binder
```

The fix already handles both; this is regression coverage, not new risk.

egonSchiele (commented):
All four taken. You were right about the impact claim, and I verified it rather than just conceding it.

## The impact claim — you were right

You could not test it because `dist/` had the fix; I reverted the skip and rebuilt:

```
$ agency run rest.agency        # default config
warning AG4004: Function '__objectRest' is not defined.
RAN: x                          # ...and it runs
```

So by default this was a spurious warning on every object-rest destructure, not a failure. "Unusable" and "broken" were wrong. Description corrected, including why it is still worth fixing but not urgent: a false positive on a core language feature is how people learn to ignore a diagnostic.

Nothing gates on the `warn` downstream that I can find — the program compiles, runs, and produces correct output.

## Suggestions 1 and 2 — taken, and they collapse into each other

`FunctionCall` now carries `synthetic?: boolean`. The lowerer sets it, the check reads it, and `SYNTHETIC_CALLS` is gone — which answers where the category should live: nowhere. The name in `constants.ts` stays, since the lowerer and builder still both need it.

Your framing is what convinced me: the argument in my own description ("somewhere obvious to go") was an argument for the flag, and I had written it while implementing the list. The field also closes the spoofing hole, which the list did not.

I checked the field is load-bearing rather than decorative — dropping `synthetic: true` from the lowerer takes the typechecker tests to 3 failed.

## 3 and 4 — renamed and covered

`objectRest.agency`, and the two positions added. `restInNestedPattern` asserts the inner rest holds both remaining keys, and `restInForLoopBinder` covers the loop binder. Five nodes now, all passing.

## One thing your review made me check

You called `restExcludesNamedKeys` the assertion that pins the semantics, which prompted me to ask whether the *other* execution tests pin anything. They do not — not for this bug:

| break | result |
|---|---|
| revert the fix entirely | execution tests **5/5**, still green |
| disable the builder's `OBJECT_REST_FN` branch | execution tests 0/5 |

AG4004 is a warning in the test runner, so those nodes run fine even when the diagnostic wrongly fires. The execution file guards the **builder** half only; the typechecker tests guard the diagnostic. I had claimed it covered both. The file now says which half it is and points at the other, and the PR description has the table.

That is the second time this branch that I described a test as regression coverage without watching it fail first. Doing it by default now.

PR #700, opened 2026-07-25
Array patterns: no rest binder means an exact length

egonSchiele:
Part 1 of [the array-pattern spec](https://github.com/egonSchiele/agency-lang/blob/array-pattern-exact-length/packages/agency-lang/docs/superpowers/specs/2026-07-25-array-pattern-length-and-middle-rest-design.md). Part 2 (a rest binder in any position) builds on the same arithmetic and lands next.

## The bug

```ts
match (["a", "b", "c"]) {
  ["a", "b"] => "shouldn't match"
  _          => "fell through"
}
// before: "shouldn't match"
```

Every array pattern was a *prefix* match. `["a", "b"]` accepted any array starting with `a`, `b`, and there was no way to say "exactly these two".

The cause reads as an unfinished edit rather than a decision — the ternary computes `hasRest` and then ignores it:

```ts
const op: Operator = hasRest ? ">=" : ">=";   // both branches
```

## The change

```ts
const op: Operator = hasRest ? ">=" : "==";
```

A pattern with no rest binder names the whole array. With one, it names a minimum — unchanged.

| pattern | `["a","b","c"]` | `["a","b"]` |
|---|---|---|
| `["a", "b"]` | no match *(was: match)* | match |
| `["a", "b", ...rest]` | match, `rest = ["c"]` | match, `rest = []` |

## Scope

Affects every construct that routes through `patternToCondition` — `match` arms, `is`, `while` conditions — and **not** destructuring: `const [a, b] = xs` emits no length check and still takes the first two. Both directions are tested, since "does this break my destructuring" is the first thing a reader will ask.

Measured across all 1258 `.agency` files: **15** array patterns in matching constructs, 13 without a rest — all two-element patterns in test files, every one matched against a two-element value, so none changes result. **Zero in `stdlib/`**, and `make fixtures` produced no churn.

## This is a silent breaking change

Code relying on the old prefix reading stops matching with no diagnostic. That is unavoidable — the point is that the pattern now means what it says — and it is why the changelog entry is part of this commit rather than a follow-up. It is the only signal a user gets.

It is also why this is its own PR: the blast radius is provably small today, and Part 2 changes the same arithmetic. Doing it second would mean building on a base that is known wrong.

## Tests

Written first; 3 of 4 nodes failed before the change, with `restStillMatchesLonger` passing throughout as the control that the `>=` path is untouched.

- exact length: longer no longer matches, exact still does, shorter never did
- rest binder: still matches longer, still binds zero elements
- nested array patterns get the same rule
- `is` and `while` positions change with match arms

### DIFF
--- a/packages/agency-lang/docs/superpowers/specs/2026-07-25-array-pattern-length-and-middle-rest-design-REVIEW.md
+++ b/packages/agency-lang/docs/superpowers/specs/2026-07-25-array-pattern-length-and-middle-rest-design-REVIEW.md
@@ -1,0 +1,128 @@
+# Review: array patterns — exact length, and a rest binder anywhere
+
+Review of [the spec](2026-07-25-array-pattern-length-and-middle-rest-design.md),
+2026-07-25.
+
+## Verdict
+
+Ready to implement once the gaps below are closed. The diagnosis in Part 1 is
+right, the language table is the correct frame for "is a middle rest exotic",
+and the performance section pre-empts the obvious objection with numbers
+instead of assertion. Rejecting a separate one-or-more sigil is well argued —
+`["a", mid, ...rest, "c"]` really does say the thing, and no language in the
+table disagrees.
+
+Two things I checked that hold up:
+
+- The tail arithmetic. `xs[xs.length - tail.length + i]` is correct, and the
+  worked example against `["a","b",...rest,"d","e"]` indexes the right
+  elements.
+- Part 1's cause. `hasRest ? ">=" : ">="` reads as an unfinished edit, not a
+  decision — same conclusion I reached reviewing #676, where the line is
+  adjacent to that change.
+
+## 1. The headline examples use syntax that does not parse
+
+Lines 77–78 both write `...rest: string[]`:
+
+```ts
+["echo", a: string, b: string, ...rest: string[], last: string]
+```
+
+That does not parse today:
+
+```
+["echo", ...rest: string[]] => "yes"
+Failed to parse: expected match cases of the form `value => expression`
+```
+
+`takesTypeSuffix` (added in #695) admits `variableName`, `objectPattern`,
+`arrayPattern` and `wildcardPattern` — not `restPattern`. So Part 2 as written
+needs a third change, and that change appears only as a Tests bullet (line
+241). It is absent from Semantics, Lowering, Errors and Sequencing.
+
+It also opens a semantics question the spec never answers: **what does
+`...rest: string[]` test?** Two readings, and they are not close:
+
+- Validate the slice element-wise against `string[]`. Cost scales with the
+  slice length, and by #695's own measurements a named type runs ~370 ns per
+  value — which would dominate every number in the performance table.
+- Assert only that the slice is an array. Always true, so the suffix is a
+  no-op and should probably be rejected rather than accepted.
+
+**Recommendation:** drop the suffix from the Part 2 examples so Part 2 is
+exactly what it claims, and give the rest-binder suffix its own part with its
+own semantics. As written the spec's two most prominent lines describe
+something it does not plan to build.
+
+## 2. The impact analysis counts match arms; `is` positions change too
+
+`if (xs is ["a", "b"])` parses and prefix-matches today — confirmed against a
+current build. `while` conditions and `match (x is [...])` heads route through
+the same `patternToCondition`, so Part 1 changes all of them.
+
+The count is genuinely zero in this repo, so "provably this small" survives
+intact. But **"Array patterns in match position: 15"** reads as though match
+arms were the whole surface, and `docs/site/guide/pattern-matching.md`
+documents `is` with array patterns — so a reader checking their own code
+against this table would check the wrong construct.
+
+State the constructs, not only the count.
+
+While there: say explicitly that **binding position is untouched**.
+`const [a, b] = xs` goes through `extractBindings`, which emits no length
+check at all, so Part 1 cannot affect destructuring. Every reader will wonder;
+the spec never says.
+
+## 3. "Exactly today's output" has to be a deliberate branch
+
+`sliceCall(source, start)` (`patternLowering.ts:1413`) emits a start-only
+slice. The two-sided form produces `xs.slice(2, xs.length - 0)` when `tail` is
+empty unless the one-argument form is kept explicitly.
+
+The spec asserts the prefix case is "unchanged rather than reimplemented"
+(line 184). That holds only if the branch is written down as a requirement.
+Otherwise every existing rest fixture churns, and a diff where the interesting
+change is buried in fixture noise stops being reviewable.
+
+Worth recording the good news too: `AccessChainElement` is already
+`{ kind: "slice"; start?; end? }` (`lib/types/access.ts:7`), so the two-sided
+slice is a parameter to machinery that exists, not a new IR node.
+
+## 4. "Reject with a diagnostic" does not say which mechanism
+
+`enforceRestAtEnd` lives in the parser (`parsers.ts:5846`). Diagnostics with
+AG codes come from the typechecker registry, so a parser cannot emit one.
+
+The Errors section therefore has an unresolved choice: either this becomes a
+good **parse error message**, or the check **moves to the typechecker** so it
+can carry a code. Since the spec's own complaint is that users get a stack
+trace instead of a diagnostic, it should say which, and pick the number.
+
+## 5. Two missing rows in the semantics table
+
+- **Rest first** — `[...rest, "c"]`. The minimal middle-rest case, and the one
+  most likely to expose an off-by-one in the tail arithmetic, since `head` is
+  empty. Neither the table nor the Tests section has it.
+- **Rest alone** — `[...rest]`, where head and tail are both empty.
+
+One line each, in the table and in Tests.
+
+## Smaller
+
+- The impact table's `affected` column means "meaning changes", but the prose
+  immediately explains that no result changes. Renaming it ("meaning changes")
+  removes the contradiction.
+- Part 1 is a silent behavior change with no diagnostic, accepted for good
+  reasons. That makes a changelog entry the only signal users get — worth
+  naming as a deliverable of that PR rather than leaving it implied.
+
+## Sequencing
+
+Splitting Part 1 into its own PR is the right call, and the reason given —
+build Part 2 on a base that is known correct — is the right reason.
+
+One addition: if the rest-binder type suffix stays in scope at all, it should
+be a third PR after Part 2, not folded into it. It is a parser change plus a
+semantics decision, and it shares nothing with the length arithmetic that
+makes Parts 1 and 2 one design.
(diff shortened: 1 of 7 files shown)


packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -1172,11 +1172,14 @@ function collectChecks(pattern: MatchPattern, source: Expression, checks: Expres
...
-      // Length check — at least as many elements as named (excluding rest).
+      // Length check. With a rest binder the pattern names a minimum; without
+      // one it names the whole array, so the length must match exactly.
+      // `["a", "b"]` used to accept ["a","b","c"] — every array pattern was a
+      // prefix match, and there was no way to say "exactly these two".
       const hasRest = pattern.elements.some((e) => e.type === "restPattern");
       const namedCount = pattern.elements.filter((e) => e.type !== "restPattern").length;
       const lenAccess = fieldAccess(source, "length", pattern.loc);
-      const op: Operator = hasRest ? ">=" : ">=";
+      const op: Operator = hasRest ? ">=" : "==";
```
egonSchiele, line 1189: Should it be `==` or `===`?

egonSchiele, line 1189: Worth knowing that the two branches don't emit comparable code:

```js
__eq(__scrutinee_1.length, 2)      // ==
__scrutinee_1.length >= 2          // >=
```

`==` routes through the deep-equality helper; `>=` is native. Correct either way, since `.length` is always a number — but it means the spec's perf table (`exact 2.7` vs `prefix 5.0`) describes a native `===` rather than what this line generates, and has the relationship backwards for the emitted code.

Emitting `===` for a `.length`-versus-numeric-literal comparison would make the table true and shave a call off every array pattern. Small either way; mostly worth fixing so the spec doesn't carry a number that contradicts the code.

packages/agency-lang/tests/agency/pattern-matching/arrayPatternExactLength.agency
```diff
@@ -0,0 +1,65 @@
...
+node restStillMatchesLonger() {
+  // A rest binder keeps the "at least" reading, and still binds zero.
+  return [withRest(["a", "b", "c"]), withRest(["a", "b"]), withRest(["a"])]
+}
+
+node nestedIsExactToo() {
+  return [nested([["a", "b"], "z"]), nested([["a", "b", "c"], "z"])]
+}
+
+node isAndWhileChangeToo() {
```
egonSchiele, line 106: Three gaps, in the order I'd add them:

**`match (x is ["a","b"])`** — listed as affected in the revised spec, and the only construct here whose failure mode isn't "falls through": a non-matching head returns `failure(...)` from the enclosing function. That's a control-flow change, not a match change, and nothing pins it.

**Destructuring** — the description says both directions are tested, but there's no `const [a, b] = xs` in the diff. It's the claim a reader most wants backed for a silent breaking change.

**An array inside an object** — `{ words: ["a","b"] }`. `nested` covers array-in-array; array-in-object is the shape the safeBash matcher actually uses.

packages/agency-lang/tests/agency/pattern-matching/arrayPatternExactLength.agency
```diff
@@ -0,0 +1,65 @@
...
+  }
+  return "is fell through"
+}
+
+def whilePosition(xs: any[]): string {
+  // So does a `while` condition. One iteration is enough to tell.
+  let seen = "never"
+  while (xs is ["a", "b"]) {
+    seen = "while matched"
+    return seen
```
egonSchiele, line 44: This `return` is load-bearing for termination, not just for the result — `xs` never changes, so the condition is loop-invariant and a matching value would spin forever without it.

It discriminates correctly as written. But a test whose regression mode is "hangs CI" rather than "fails" is worth one comment saying so, since the next person editing this body has no way to know.

packages/agency-lang/docs/superpowers/specs/2026-07-25-array-pattern-length-and-middle-rest-design-REVIEW.md
```diff
@@ -0,0 +1,128 @@
+# Review: array patterns — exact length, and a rest binder anywhere
+
+Review of [the spec](2026-07-25-array-pattern-length-and-middle-rest-design.md),
+2026-07-25.
+
+## Verdict
```
egonSchiele, line 15: Committing this alongside the spec is right, and matches the `-REVIEW.md` convention in `plans/`. It just needs a status header now that the spec has answered it — as written, "Ready to implement once the gaps below are closed" reads as still-open next to a spec that closed them.

Something like: items 2, 3, 4 and 5 folded into the 2026-07-25 revision; item 1 (the rest-binder type suffix) deferred to Part 3. Otherwise the two documents look like they disagree, and a reader has to diff them to find out they don't.

egonSchiele (commented):
## Overview

One character of behavior change, plus the surrounding work that makes a silent breaking change safe to ship: a changelog entry in the same commit, a guide section placed exactly where the contrast with destructuring lives, and a test set with a real control (`restStillMatchesLonger` passing throughout is what proves the `>=` path wasn't disturbed).

The spec revision that came with it is thorough — the impact section now names every affected construct instead of just match arms, binding position is called out as untouched, the rest-first and rest-alone rows are in the semantics table, and the "which mechanism" question for the two-rest error is resolved with a real reason (lowering runs before typechecking, so the registry isn't available). Good.

Four things, one of which is a factual error in the description.

## 1. The description claims a destructuring test that isn't here

> Both directions are tested, since "does this break my destructuring" is the first thing a reader will ask.

The `.agency` file has four nodes — `exactLength`, `restStillMatchesLonger`, `nestedIsExactToo`, `isAndWhileChangeToo`. None of them destructures. There is no `const [a, b] = xs` anywhere in the diff.

The reasoning is right and the claim is the one a reader most wants backed, since "will this break my existing code" is the whole question for a silent breaking change. It's also the cheapest node in the file:

```ts
node destructuringUnchanged() {
  // A declaration binds rather than tests, so it still takes the first two.
  const [a, b] = ["a", "b", "c"]
  return "${a}${b}"
}
```

## 2. `==` and `>=` lower to different machinery, and the spec's numbers describe the wrong one

The two paths do not generate comparable code:

```js
// no rest — this PR
… && __eq(__scrutinee_1.length, 2) && …

// with rest — unchanged
… && __scrutinee_1.length >= 2 && …
```

So the exact-length path is a call into the deep-equality helper where the prefix path is a native comparison. Correctness is fine — `.length` is always a number — but the spec's performance table says:

| `["a","b"]` exact | 2.7 |
| `["a","b",...rest]` prefix (today) | 5.0 |

That was measured with a native `===`, not with `__eq`, so it has the relationship backwards for the code this PR actually emits. Worth either correcting the row or, better, emitting `===` when both sides are a `.length` access and a numeric literal — at which point the table becomes true and every array pattern gets slightly cheaper. Not blocking; the cost is small and this is a comparison inside a condition that already calls `__coarseTypeTest`.

## 3. `match (x is [...])` is listed as affected and isn't tested — and it's the surprising one

The revised spec names four constructs: match arms, `is`, `while` conditions, and `match (x is [...])` heads. The tests cover the first three.

The fourth is the one worth a test, because its failure mode is different in kind. A head pattern that doesn't match doesn't fall through — it returns `failure("match(... is pattern) head pattern did not match")` from the enclosing function. So:

```ts
match (xs is ["a", "b"]) { … }   // xs = ["a","b","c"]
```

used to run an arm and now early-returns a failure from the whole function. Every other construct in this PR degrades to "falls through"; this one changes control flow. If someone is relying on the prefix reading, this is where it hurts most, and it's the case the changelog doesn't mention either.

## 4. The shape the series is actually building toward isn't covered

`nested` tests an array inside an array. The safeBash matcher's shape is an array inside an **object**:

```ts
{ words: ["cat", p] }
```

Same code path, so this is regression coverage rather than new risk — but it's the shape every downstream consumer will write, and it costs one line.

## Smaller

- **`whilePosition` terminates only because of the `return`.** `xs` never changes, so the loop condition is invariant. It discriminates correctly today, but if that `return` is ever refactored into a `break`-less form the test hangs CI instead of failing. A comment saying the return is load-bearing, or a counter guard, would make that safe.
- **The committed review has no resolution status.** It now sits next to a spec that answered it, but still opens with "Ready to implement once the gaps below are closed" — which reads as open. A `## Status` line saying which items were taken and which moved to Part 3 would stop the two documents from appearing to disagree. (The `-REVIEW.md` sibling convention is right; this is just the header it needs.)

## Verdict

The change is correct and the process around it is what a silent breaking change deserves — changelog in the same commit, blast radius measured rather than asserted, guide updated where a reader will hit it. Item 1 is a straight factual fix to either the description or the tests. Item 3 is the one I'd actually want before merge, since it's a control-flow change that nothing currently pins.

egonSchiele (commented):
All four in 3c8f4c58f. Item 2 I checked rather than implemented, and it turned out the other way.

## 1. You are right — the description claimed a test that did not exist

No `const [a, b] = xs` anywhere in the diff. I had verified destructuring by hand and then wrote "both directions are tested", which is exactly the kind of claim that should not survive without a test behind it. Added your node verbatim.

## 2. `===` compiles to the same thing as `==`

The suggestion does not work, and the reason is worth recording:

```ts
// typescriptBuilder.ts:1379
// All equality operators use unified nullish equality via the `__eq`
// runtime helper (null and undefined compare equal). There is no strict
// escape hatch: `===`/`!==` are stylistic aliases that compile
// identically to `==`/`!=`.
```

I tried it first — swapped the operator, rebuilt, and the output was still `__eq(__scrutinee_1.length, 2)`.

But your underlying point stands: the table described code we do not emit. Measured properly:

| check | ns/op |
|---|---|
| native `xs.length === 2` (not what we emit) | 1.5 |
| `__eq(xs.length, 2)` — exact-length path | 5.8 |
| native `xs.length >= 2` — rest path | 4.7 |

So it is ~1 ns, not the inversion the table implied. Bypassing the helper would make one generated check the only equality in the language that skips it — against a documented language-wide rule, for a nanosecond. Corrected the spec instead, with the measurement and the reasoning.

## 3 and 4. Both added, and 3 is the one that mattered

`match (x is [...])` now has two nodes: one asserting a non-matching head returns a **failure from the enclosing function**, and one asserting the exact-length head still runs. You are right that it is different in kind — every other construct here degrades to "falls through" and this one changes control flow. The changelog says so now, with the before/after, since that is where a user relying on the prefix reading would be hurt worst and it was the one case the entry did not mention.

Array-inside-an-object added too.

## Inline

- The while-loop `return` being load-bearing for **termination** is a good catch — a regression there hangs CI rather than failing it, and nothing said so. Comment added.
- Guide now names the constructs (`match` arm, `is`, `if`, `while`) instead of "in a match".
- REVIEW.md has a status header: items 2–5 and both Smaller folded into the revision, item 1 deferred to Part 3. It should stop reading as open next to a spec that closed it.

Eight nodes now, all passing.

PR #701, opened 2026-07-26
A rest binder can sit anywhere in an array pattern

egonSchiele:
Part 2 of [the array-pattern spec](../blob/main/packages/agency-lang/docs/superpowers/specs/2026-07-25-array-pattern-length-and-middle-rest-design.md), after #700. This is the last blocker for the safeBash command matcher.

## What works now

```ts
match (words) {
    ["cat", ...flags, path]           => read(path)
    ["a", ...rest, "d", "e"]          => ...        // two trailing
    [...rest, "c"]                    => ...        // empty head
    ["cmd", ...rest, { path: string }] => ...        // trailing sub-pattern
}

const [first, ...middle, last] = items              // binding position too
```

`...rest` binds zero or more, so `["a", ...rest, "c"]` matches `["a", "c"]` with `rest = []` — matching Rust, Python, Ruby and Scala.

## Lowering

Split the element list at the rest binder; index the head from the front and the tail from the back (`xs[xs.length - n]`). The rest becomes a two-sided slice.

**The one-argument slice is kept for the no-tail case**, as the spec requires — a two-sided call would emit `xs.slice(2, xs.length - 0)` for every existing prefix pattern and bury the real diff in fixture churn. `make fixtures` produced **no changes**, which is the check that it held.

`enforceRestAtEnd` becomes `enforceAtMostOneRest`: Part 2 removes the position constraint, the count constraint stays. Two rests have no meaning to give — nothing decides where the split between them falls. The budget is per array pattern, so `[a, ...outer, [b, ...inner]]` is fine.

Binding position is included per the spec. The JavaScript constraint that forbids it there — destructuring consumes an iterator, so it cannot know where the end is — does not apply to an indexable array.

## Tests

7 execution nodes, all failing before the change, plus an `expectedCompileError` for two rest binders.

The design point worth reviewing: **the binder tests assert what landed in each binder, not just that the arm ran.** A trailing element bound from the wrong end still *matches* — the arm fires, and only the values distinguish correct from subtly wrong:

```ts
bindings(["a","b","c","d"])  →  ["a", ["b","c"], "d"]
bindings(["a","b"])          →  ["a", [],        "b"]    // binders must not overlap
```

Covered: one and two trailing elements, empty head, rest alone, empty middle, a trailing sub-pattern, a type-suffixed trailing element, and binding position.

## One thing I had to fix in passing

The `patternGuards` tripwire's corpus loop handled a returned parse failure but not a **throwing** one. The arity check throws so its message survives the surrounding `or` combinators, and the new `expectedCompileError` fixture is the first corpus file that fails that way. The loop now tolerates both.

Worth noting the tripwire still passes with middle-rest patterns in the corpus — so the shape check continues to guard the new from-the-end reads.

### DIFF
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -1057,14 +1057,37 @@ class PatternLowerer {
           return [makeAssign(prop.identifier, makeObjectRestCall(source, namedKeys, loc), declKind, loc)];
         });
       }
-      case "arrayPattern":
-        return pattern.elements.flatMap((el, i): Assignment[] => {
+      case "arrayPattern": {
+        // Same split as collectChecks: head binders read from the front, the
+        // rest takes the middle, tail binders read from the back.
+        const { head, tail } = splitAtRest(pattern.elements);
+        const headBindings = head.flatMap((el, i): Assignment[] => {
           if (el.type === "wildcardPattern") return [];
-          if (el.type === "restPattern") {
-            return [makeAssign(el.identifier, sliceCall(source, i, loc), declKind, loc)];
-          }
           return this.extractBindings(el, indexAccess(source, i, loc), declKind, loc);
         });
+        const restBindings = pattern.elements.flatMap((el): Assignment[] =>
+          el.type === "restPattern"
+            ? [
+                makeAssign(
+                  el.identifier,
+                  restSlice(source, head.length, tail.length, loc),
+                  declKind,
+                  loc,
+                ),
+              ]
+            : [],
+        );
+        const tailBindings = tail.flatMap((el, i): Assignment[] => {
+          if (el.type === "wildcardPattern") return [];
+          return this.extractBindings(
+            el,
+            indexFromEnd(source, tail.length - i, loc),
+            declKind,
+            loc,
+          );
+        });
+        return [...headBindings, ...restBindings, ...tailBindings];
+      }
       case "variableName":
         return [makeAssign(pattern.value, cloneExpr(source), declKind, loc)];
       case "wildcardPattern":
@@ -1188,13 +1211,28 @@ function collectChecks(pattern: MatchPattern, source: Expression, checks: Expres
       // only equality in the language that skips the helper.
       const op: Operator = hasRest ? ">=" : "==";
       checks.push(makeBinOp(lenAccess, op, numberLit(namedCount, pattern.loc), pattern.loc));
-      pattern.elements.forEach((el, i) => {
-        if (
-          el.type !== "wildcardPattern" &&
-          el.type !== "restPattern" &&
-          el.type !== "variableName"
-        ) {
+
+      // A rest binder may sit anywhere, so elements before it are indexed from
+      // the front and elements after it from the back. With no rest (or a
+      // trailing one) `tail` is empty and this is the original front-indexed
+      // walk.
+      const { head, tail } = splitAtRest(pattern.elements);
+      const needsCheck = (el: { type: string }): boolean =>
+        el.type !== "wildcardPattern" &&
+        el.type !== "restPattern" &&
+        el.type !== "variableName";
+
+      head.forEach((el, i) => {
+        if (needsCheck(el)) {
           collectChecks(el as MatchPattern, indexAccess(source, i, pattern.loc), checks);
+        }
+      });
+      tail.forEach((el, i) => {
+        if (needsCheck(el)) {
+          // 1-based from the end: the last element of `tail` is the last
+          // element of the array.
+          const fromEnd = tail.length - i;
+          collectChecks(el as MatchPattern, indexFromEnd(source, fromEnd, pattern.loc), checks);
         }
       });
       break;
@@ -1418,6 +1456,68 @@ function indexAccess(source: Expression, i: number, loc: SourceLocation | undefi
     [{ kind: "index", index: numberLit(i, loc) }],
     loc,
   );
+}
+
+/**
+ * `source[source.length - fromEnd]` — an element counted from the back, for
+ * the elements that follow a rest binder. `fromEnd` is 1-based: the last
+ * element is 1.
+ */
+function indexFromEnd(
+  source: Expression,
+  fromEnd: number,
+  loc: SourceLocation | undefined,
+): ValueAccess {
+  const offset = makeBinOp(
+    fieldAccess(cloneExpr(source), "length", loc),
+    "-",
+    numberLit(fromEnd, loc),
+    loc,
+  );
+  return chainAccess(source, [{ kind: "index", index: offset }], loc);
+}
+
+/**
+ * The slice a rest binder takes. With nothing after it this is the
+ * one-argument `source.slice(start)` — the same node the prefix case has
+ * always produced, kept deliberately so lifting this restriction does not
+ * churn every existing rest fixture.
+ */
+function restSlice(
+  source: Expression,
+  start: number,
+  tailCount: number,
+  loc: SourceLocation | undefined,
+): ValueAccess {
+  if (tailCount === 0) return sliceCall(source, start, loc);
+  const end = makeBinOp(
+    fieldAccess(cloneExpr(source), "length", loc),
+    "-",
+    numberLit(tailCount, loc),
+    loc,
+  );
+  return chainAccess(
+    source,
+    [{ kind: "slice", start: numberLit(start, loc), end }],
+    loc,
+  );
+}
+
+/**
+ * Split an array pattern at its rest binder. `restIndex` is -1 when there is
+ * none, in which case everything is head and the two halves behave exactly as
+ * they did before rest binders could sit anywhere.
+ */
+function splitAtRest<T extends { type: string }>(
+  elements: readonly T[],
+): { head: readonly T[]; tail: readonly T[]; restIndex: number } {
+  const restIndex = elements.findIndex((e) => e.type === "restPattern");
+  if (restIndex === -1) return { head: elements, tail: [], restIndex };
+  return {
+    head: elements.slice(0, restIndex),
+    tail: elements.slice(restIndex + 1),
+    restIndex,
+  };
 }
 
 function sliceCall(source: Expression, start: number, loc: SourceLocation | undefined): ValueAccess {
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -1077,11 +1077,18 @@ class PatternLowerer {
               ]
             : [],
         );
+        // A tail binder reads from the end, which aliases a leading binder when
+        // the value is too short. Matching cannot reach that state (the length
+        // check fails first); a declaration can, so guard the source here.
+        const guardedSource =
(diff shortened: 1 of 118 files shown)


egonSchiele (commented):
## Overview

The lowering is right. `splitAtRest` gives both `collectChecks` and `extractBindings` the same split, the from-the-end arithmetic is correct, and `restSlice` keeping the one-argument form when `tailCount === 0` is exactly what the spec asked for — with `make fixtures` clean as the check that proves it rather than a claim that it should be.

One detail I went looking for and found already right: `indexFromEnd` clones `source` for the length access and passes it uncloned to `chainAccess`, which looks asymmetric until you notice `chainAccess` clones its base itself. No aliasing.

The test design is the strongest part. "A trailing element bound from the wrong end still *matches*, so the assertion has to be on what landed in each binder" is the observation that makes this reviewable, and `bindingsWithEmptyMiddle` asserting `["a", [], "b"]` is precisely the case a wrong implementation passes.

Two real problems, both confirmed against a build of this branch.

## 1. Binding position silently binds the same element twice

```ts
const [first, ...mid, last] = ["a"]
// first=a mid=0 last=a
```

`first` and `last` are the same element. The arithmetic is right — `slice(1, 0)` is `[]` and `xs[xs.length - 1]` is `xs[0]` — but it is only *safe* under the length guard, and binding position emits no length check at all. Match position is fine: `one(["a"])` falls through, and you test it.

So this is new surface. With a trailing rest, aliasing was impossible: `const [a, ...rest] = ["x"]` gives `a="x", rest=[]` and there is no second binder to collide with. Letting the rest move introduced the case.

`bindingPositionToo` tests the four-element happy path only, and `bindingsWithEmptyMiddle`'s comment — "the two named binders do not overlap" — is asserting that about the *match* path, where the guard does the work. The claim reads as if it covers binding position, and it does not.

The PR needs to pick what happens and test it. Three options:

- **Emit a failure.** `pattern-matching.md`'s "Failure semantics" already says a destructuring that cannot be satisfied surfaces as a `failure` Result. "Not enough elements to fill both ends" is that situation, and this is the only option consistent with what the language already does. Python raises for the same case (`a, *m, b = ["x"]` → ValueError).
- **Bind the tail to null when the array is too short.** Closest to JS's `const [a, b] = ["x"]`, but the current arithmetic cannot express it without a length test anyway, so it costs the same as the option above.
- **Document the aliasing.** Weakest, but honest, and it is what ships today by default.

Any of them is fine; shipping without choosing is not, because a duplicate binding is the kind of thing someone debugs for an hour.

## 2. Two rest binders still produce a stack trace, and the test cannot tell

What a user sees:

```
file:///…/dist/lib/parsers/parsers.js:3031
            throw new Error("an array pattern may have at most one rest binder (`...name`)");
                  ^

Error: an array pattern may have at most one rest binder (`...name`)
    at enforceAtMostOneRest (…/parsers.js:3031:19)
    at …/tarsec/dist/parsers.js:265:24
    … 13 more frames …
```

No file, no line, no column in their `.agency` source. Compare an actual diagnostic:

```
foo.agency:3:5 - error AG1013: `NoSuchType` is not a type; …
```

The spec was explicit that this is a bug on its own terms: "a raw `throw new Error` … so the user gets a stack trace rather than a diagnostic. **That is a bug regardless of whether we lift the restriction.**" This PR renames the function and improves the message, but keeps the mechanism, so the part the spec called out independently is untouched.

**And the new fixture cannot detect that.** `expectedCompileError: "at most one rest binder"` is a substring that appears *inside* the stack trace, so the test passes identically whether the presentation is a clean parse error or fifteen frames of tarsec internals. It will keep passing when this is fixed and keep passing if it never is.

The spec already resolved the mechanism — a parse error, since lowering runs before typechecking, so the diagnostic registry is not available. Catching the throw at the compile driver and rendering it in the existing `Failed to parse <file>: <message>` shape would close it.

### The tripwire change is a symptom of this

> The `patternGuards` tripwire's corpus loop handled a returned parse failure but not a **throwing** one.

That is a good catch, but it is worth naming *why* it was needed: the corpus loop is correct about how parse failures work, and this one is an exception because it throws. If the error went through the normal parse-failure path, the tripwire would have needed no change. Teaching the loop to tolerate throws makes the workaround permanent, and the next corpus file that throws will be fine while the next reader wonders why one parser is special.

## Smaller

- `splitAtRest` returns `restIndex` and neither caller uses it — both destructure `{ head, tail }`. Drop it or use it.
- `typedTail` is only exercised with a non-empty rest. `typedTail(["cmd", 7])` would cover a type-suffixed tail against an empty middle, which is the boundary where the from-the-end index and the slice end meet.

## Verdict

The feature is right and the lowering is clean. Item 1 is the one I would block on — a silent duplicate binding in `let`/`const` is a correctness surprise with no signal, and it is new in this PR. Item 2 is smaller in blast radius but worth doing here, because the spec identified it, this PR touches the exact function, and the fixture as written will mask it indefinitely.

packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -1057,14 +1057,37 @@ class PatternLowerer {
...
+                makeAssign(
+                  el.identifier,
+                  restSlice(source, head.length, tail.length, loc),
+                  declKind,
+                  loc,
+                ),
+              ]
+            : [],
+        );
+        const tailBindings = tail.flatMap((el, i): Assignment[] => {
```
egonSchiele, line 1087: This arithmetic is correct, but it is only *safe* under the length check — and `extractBindings` runs in binding position, which emits no length check at all.

Confirmed on a build of this branch:

```ts
const [first, ...mid, last] = ["a"]
// first=a mid=0 last=a
```

`slice(1, 0)` is `[]` and `xs[xs.length - 1]` is `xs[0]`, so `first` and `last` are the same element. Match position is protected — `one(["a"])` falls through and you test it — but nothing protects a declaration.

This is new surface: with a trailing rest there was no second binder to collide with, so letting the rest move is what introduced it. See item 1 for the options; the one consistent with `pattern-matching.md`'s existing "Failure semantics" is to surface a `failure` Result, which is also what Python does for `a, *m, b = ["x"]`.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -5842,17 +5842,32 @@ export const restPatternParser: Parser<RestPattern> = label(
...
 ): void {
-  for (let i = 0; i < elements.length - 1; i++) {
-    if (elements[i].type === "restPattern") {
+  let seen = false;
+  for (const element of elements) {
+    if (element.type !== "restPattern") continue;
+    if (seen) {
       // Throw to surface the message past surrounding `or` combinators,
       // which would otherwise swallow it as "all parsers failed".
       throw new Error(
```
egonSchiele: The message is much better. The mechanism is the thing the spec objected to, though, and it is unchanged — a user gets fifteen frames of tarsec internals and no file, line or column in their `.agency` source:

```
Error: an array pattern may have at most one rest binder (`...name`)
    at enforceAtMostOneRest (…/parsers.js:3031:19)
    at …/tarsec/dist/parsers.js:265:24
    …
```

The spec's wording was "a raw `throw new Error` … the user gets a stack trace rather than a diagnostic. That is a bug regardless of whether we lift the restriction" — so it was scoped as independent of the feature, and this PR is the natural place for it since it already rewrites this function.

The spec also settled the mechanism: a parse error, because lowering runs before typechecking. Catching this at the compile driver and rendering it as `Failed to parse <file>: <message>` matches what every other parse failure looks like.

The comment above the throw explains why it throws rather than returns — worth keeping that reasoning even after the presentation is fixed, since it is the non-obvious part.

packages/agency-lang/tests/agency/expectedCompileError/twoRestBinders.test.json
```diff
@@ -0,0 +1,4 @@
+{
+  "expectedCompileError": "at most one rest binder",
```
egonSchiele: This substring appears *inside* the stack trace, so the fixture passes whether the output is a clean parse error or fifteen frames of tarsec internals. It will keep passing when the presentation is fixed and keep passing if it never is — which makes it a test of the message text only, not of the thing the spec actually complained about.

If `expectedCompileError` supports it, asserting a `file:line` prefix — or the absence of `\n    at ` frames — would make this the test that pins the fix. Worth checking what the harness allows; right now the assertion is weaker than it reads.

packages/agency-lang/tests/agency/pattern-matching/arrayPatternMiddleRest.agency
```diff
@@ -0,0 +1,99 @@
...
+node nestedAndTypedTails() {
+  return [
+    nestedTail(["cmd", "x", { path: "f.txt" }]),
+    nestedTail(["cmd", { path: 42 }]),
+    typedTail(["cmd", "x", 7]),
+    typedTail(["cmd", "x", "no"]),
+  ]
+}
+
+node bindingPositionToo() {
```
egonSchiele, line 111: This covers the four-element happy path. The case that distinguishes binding from match position is a value too short to fill both ends:

```ts
const [a, ...m, b] = ["a"]   // a and b are both "a" today
```

Worth a node once item 1 is decided, whichever way it goes — as an assertion of the failure, or as an assertion of the documented aliasing.

Note also that `bindingsWithEmptyMiddle`'s description ("the two named binders do not overlap") is about the *match* path, where the length guard does the work. Read from here it sounds like it already covers this.

egonSchiele (commented):
Both findings correct, both fixed in 1976521a1.

## The aliasing bug

Reproduced exactly:

```ts
const [first, ...mid, last] = ["a"]   // first=a  mid=[]  last=a
```

You put your finger on why it is *this* PR's problem: with a trailing rest there was no second binder to collide with, so letting the rest move is what created the collision. Match position is protected because the length check runs before any element read; a declaration emits no length check at all, because it binds rather than tests.

Fixed with `__requireLength(xs, min)` guarding the tail source. It throws, which surfaces as a `failure` Result — the behavior `pattern-matching.md` already documents for an unsatisfiable destructuring, and what Python does for `a, *m, b = ["x"]`. Your suggestion, and it was the right one.

Two nodes now: too short is a failure, exactly enough still binds `["a", [], "b"]`.

**One cost worth flagging:** the helper joins the shared import preamble, so all 94 generator fixtures gained the same one line. Mechanical and identical everywhere, but it is real churn in a PR that otherwise had none, and you should know it is there rather than find it.

## The parse error

You were right that the mechanism was untouched — I improved the message and left the presentation exactly as the spec complained about it.

The cause: `enforceAtMostOneRest` threw a bare `Error`, and `parseAgency` catches `TarsecError` and lets anything else escape. It now throws a `TarsecError`, so the existing catch renders it as an ordinary parse failure with the file and position. The reason it throws at all — surviving the surrounding `or` combinators — is unchanged and still explained, since that is the non-obvious part.

## The fixture assertion

Also right, and this is the one I would have missed. The substring appeared *inside* the stack trace, so it passed whether or not the error was presented as a diagnostic — it tested the message text, not the thing the spec objected to. It now asserts the `Failed to parse` prefix, which only the clean path produces, and says why in the description.

Grammar fixed. The guide also documents the too-short declaration case now, since it is user-visible behavior that did not exist before this PR.

PR #702, opened 2026-07-26
std::safeBash: use an equivalent tool instead of bash, starting with echo

egonSchiele:
Option B from our discussion: the machinery plus one rule, **not wired into any agent**.

**Spec:** `docs/superpowers/specs/2026-07-26-safebash-design.md`, added in this PR. It records the decisions, the refusal policy, the word-splitting rule, and the command table that comes next. The [idea doc](../blob/main/packages/agency-lang/docs/superpowers/ideas/2026-07-24-safebash-command-to-tool-matching.md) it grew from is marked superseded and kept for the reasoning.

## Why

`bash` cannot be pre-approved — there is no way to say in advance what an arbitrary command string will do — so `bash()` raises `std::bash` on every call. Meanwhile the tools that do the same jobs (`read`, `ls`, `grep`, `write`, the git tools) *are* pre-approved. That gap is why a run that could be unattended isn't: you end up approving a stream of `ls` and `cat`.

This parses the command and uses the equivalent tool when it can.

```ts
safeBash("echo hello world")   // prints, no interrupt raised
safeBash("echo hi > out.txt")  // write(), which is already gated
safeBash(anything else)        // bash(), approval and all
```

## Not wired in — deliberately

`shellTools()` is untouched. No agent's tool list changes, so there is nothing to regress, and this PR is only asking "does the machinery work?"

Whether to trust it in the approval path is a different question, and it deserves its own review with the risk in isolation. It's a one-line change to `shellTools()` when we want it.

## What `simplify` does

Reduces the parser's AST to the only shape v1 can reason about:

```ts
type Cmd = { words: string[], redirect: OutputRedirect | null }
```

Everything else is refused **by name** — pipelines, `&&`, background jobs, leading assignments, command substitution, non-output redirects, multiple commands — so a fallthrough is explainable rather than mysterious. Refusing costs an approval, which is exactly what happened before. Guessing would cost correctness.

## The subtle part: word splitting

`stringifyWordPart` resolves what is text or what the environment can answer. The rule worth reviewing:

```bash
F="a b"
cat $F      # TWO arguments to bash
cat "$F"    # one
```

An unquoted expansion holding whitespace is **refused**, not expanded — expanding it without splitting would quietly change what the command means, which is the one failure mode this feature must not have. Inside double quotes no splitting happens, so it expands normally.

## Tests

5 nodes, all passing:

- `echo` produces its output **without raising the interrupt** — the actual claim
- quoting: quotes delimit a word rather than being part of it, and adjacent parts concatenate (`a'b'"c"` → `abc`)
- every refusal, asserted by its exact message
- an output redirect lifted into the simplified shape
- unparseable input falls through rather than crashing — the parser is deliberately incomplete

## Two things worth knowing

**A typechecker bug I worked around.** `item.command` where the field is declared `AndOr` resolves to the type *alias* named `Command`, so every access below it reports a missing property. The three affected locals are `any` with a comment. Worth filing separately — I hit it only because a field name matched a type name.

**Shape lock-in.** `Cmd` and `simplify` are this module's public API. If the shape turns out wrong once real rules exist (`cat`, `ls`, `git`), changing it is a breaking change to a shipped stdlib module. I sanity-checked those rules against it on paper and the shape holds, but it's the thing to push on in review.

### DIFF
--- a/packages/agency-lang/stdlib/safeBash.agency
+++ b/packages/agency-lang/stdlib/safeBash.agency
@@ -393,15 +393,21 @@ export def safeBash(command: string, cwd: string = ""): ExecResult {
   if (parsed is success(ast)) {
     const simplified = simplify(ast)
     if (simplified is success(cmd)) {
+      // Flags are refused before the table rather than per-arm: `echo -n`
+      // suppresses the trailing newline and `-e` interprets escapes, and
+      // `joinWords` does neither. Up here it is one rule instead of a guard
+      // repeated on every arm — and it keeps calls out of match-arm guards,
+      // which the effects walk does not descend into (#668).
+      if (hasFlags(cmd.words)) {
+        return bash(command, cwd)
+      }
       return match (cmd) {
         // `echo` has no side effect beyond its output, so there is nothing to
         // approve: print it and report what bash wrote to stdout.
-        { words: ["echo", ...rest], redirect: null } if (!hasFlags(rest)) => ok(joinWords(rest))
-
+        { words: ["echo", ...rest], redirect: null } => ok(joinWords(rest))
         // Redirected, it is a file write, which `write` already gates.
-        { words: ["echo", ...rest], redirect: { op: ">", path } } if (!hasFlags(rest)) => writeThrough(path, joinWords(rest), "overwrite", command, cwd)
-        { words: ["echo", ...rest], redirect: { op: ">>", path } } if (!hasFlags(rest)) => writeThrough(path, joinWords(rest), "append", command, cwd)
-
+        { words: ["echo", ...rest], redirect: { op: ">", path } } => writeThrough(path, joinWords(rest), "overwrite", command, cwd)
+        { words: ["echo", ...rest], redirect: { op: ">>", path } } => writeThrough(path, joinWords(rest), "append", command, cwd)
         _ => bash(command, cwd)
       }
     }
--- a/packages/agency-lang/stdlib/safeBash.agency
+++ b/packages/agency-lang/stdlib/safeBash.agency
@@ -1,0 +1,404 @@
+import { _bashParser } from "agency-lang/stdlib-lib/safeBash.js"
+import { env } from "std::system"
+import { WriteMode } from "std::index"
+import { bash } from "std::shell"
+
+/** One piece of a word. Adjacent parts concatenate: `e'f'"g"$h` is a single
+ * word of four parts. */
+export type WordPart =
+  | { tag: "literal"; text: string }
+  | { tag: "singleQuoted"; text: string }
+  | { tag: "doubleQuoted"; parts: WordPart[] }
+  | { tag: "variable"; name: string }
+  | { tag: "paramExpansion"; expression: string }
+  | { tag: "commandSubstitution"; command: List }
+  | { tag: "arithmeticExpansion"; expression: string }
+
+export type BashWord = {
+  tag: "word";
+  parts: WordPart[]
+}
+
+/** `name=value` (or `name=` with a null value) before a command. */
+export type Assignment = {
+  tag: "assignment";
+  name: string;
+  value?: BashWord
+}
+
+/** A redirect like `> out.txt`, `2>&1`, or `<<< "$str"`. `fd` is the
+ * explicit file descriptor (`2` in `2>`), or null for the default. */
+export type Redirect = {
+  tag: "redirect";
+  fd?: number;
+  op: string;
+  target: BashWord
+}
+
+export type SimpleCommand = {
+  tag: "simpleCommand";
+  assignments: Assignment[];
+  words: BashWord[];
+  redirects: Redirect[]
+}
+
+export type IfCommand = {
+  tag: "if";
+  cond: List;
+  thenBody: List;
+  elifs: { cond: List; thenBody: List }[];
+  elseBody?: List;
+  redirects: Redirect[]
+}
+
+export type LoopCommand = {
+  tag: "loop";
+  kind: "while" | "until";
+  cond: List;
+  body: List;
+  redirects: Redirect[]
+}
+
+export type ForCommand = {
+  tag: "for";
+  variable: string;
+  /** The `in word...` list, or null for the implicit `in "$@"`. */
+  words?: BashWord[];
+  body: List;
+  redirects: Redirect[]
+}
+
+export type CaseItem = {
+  patterns: BashWord[];
+  body: List;
+  /** `;;`, `;&`, `;;&`, or null when the final item ends at `esac`. */
+  terminator?: string
+}
+
+export type CaseCommand = {
+  tag: "case";
+  subject: BashWord;
+  items: CaseItem[];
+  redirects: Redirect[]
+}
+
+export type Subshell = {
+  tag: "subshell";
+  body: List;
+  redirects: Redirect[]
+}
+
+export type Group = {
+  tag: "group";
+  body: List;
+  redirects: Redirect[]
+}
+
+/** `(( expression ))` as a command. The expression is kept as raw text. */
+export type ArithmeticCommand = {
+  tag: "arithmeticCommand";
+  expression: string;
+  redirects: Redirect[]
+}
+
+export type FunctionDef = {
+  tag: "functionDef";
+  name: string;
+  body: Command
+}
+
+export type Command =
+  | SimpleCommand
+  | IfCommand
+  | LoopCommand
+  | ForCommand
+  | CaseCommand
+  | Subshell
+  | Group
+  | ArithmeticCommand
+  | FunctionDef
+
+/** One or more commands joined by `|` (or `|&`), optionally negated. */
+export type Pipeline = {
+  tag: "pipeline";
+  negated: boolean;
+  commands: Command[]
+}
+
+/** Pipelines joined by `&&` / `||`, left to right. */
+export type AndOr = {
+  tag: "andOr";
+  first: Pipeline;
+  rest: { op: "&&" | "||"; pipeline: Pipeline }[]
+}
+
+export type ListItem = {
+  tag: "listItem";
+  command: Command;
+  /** True when the command was followed by `&`. */
+  background: boolean
+}
+
+/** A sequence of commands separated by `;`, `&`, or newlines — the top
+ * level of a script, and the body of every compound command. */
+export type List = {
+  tag: "list";
+  items: ListItem[]
+}
+
+export static const emptyList: List = {
+  tag: "list",
+  items: []
+}
+
+export type BashNode =
+  | WordPart
+  | BashWord
+  | Assignment
+  | Redirect
+  | Command
+  | Pipeline
+  | AndOr
+  | ListItem
+  | List
+
+
+/** A bash command reduced to the only shape v1 can reason about: the words of
+ *  a single command, and an optional output redirect. */
+export type OutputRedirect = {
+  op: string;
+  path: string
+}
+
+export type Cmd = {
+  words: string[];
+  redirect: OutputRedirect | null
+}
+
+export type SafeBashResult = {
+  stdout: string;
+  stderr: string;
+  exitCode: number
+}
+
+export def bashParser(code: string): Result<List> {
+  """
+  Parse a string of bash code into an AST.
+  """
+  const res = _bashParser(code)
+  return match(res.success) {
+    true => success(res.result)
+    false => failure(res.error)
+  }
+}
+
+def stringifyWordPart(part: WordPart, insideQuotes: boolean): Result<string> {
+  """
+  Reduce one piece of a word to text, or explain why it cannot be.
+
+  Only the parts that are already text, or that the environment can answer,
+  are resolvable. Anything that would need a shell to evaluate is a failure,
+  because guessing at it is how safeBash and bash would come to disagree
+  about what a command means.
+  """
+  return match (part) {
+    { tag: "literal", text } => success(text)
+    { tag: "singleQuoted", text } => success(text)
+    { tag: "doubleQuoted", parts } => stringifyParts(parts, true)
+    { tag: "variable", name } => expandVariable(name, insideQuotes)
+    _ => failure("`${part.tag}` needs a shell to evaluate")
+  }
+}
+
+def expandVariable(name: string, insideQuotes: boolean): Result<string> {
+  """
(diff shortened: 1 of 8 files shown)


packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -0,0 +1,404 @@
...
+  @param cwd - Working directory, passed through to bash on the fallback path
+  """
+  const parsed = bashParser(command)
+  if (parsed is success(ast)) {
+    const simplified = simplify(ast)
+    if (simplified is success(cmd)) {
+      return match (cmd) {
+        // `echo` has no side effect beyond its output, so there is nothing to
+        // approve: print it and report what bash wrote to stdout.
+        { words: ["echo", ...rest], redirect: null } => ok(joinWords(rest))
```
egonSchiele: `...rest` takes flags as text. Confirmed against a build of this branch:

```
safeBash("echo -n hi")     → "-n hi\n"    bash → "hi"      (no newline)
safeBash("echo -e a\\tb")   → "-e atb\n"   bash → "a\tb\n"   (real tab)
safeBash("echo *")         → "*\n"        bash → the file listing
safeBash("echo ~")         → "~\n"        bash → /Users/adit
```

All four take the fast path — no interrupt, no diagnostic. The docstring three lines up says "Incompleteness costs an approval, never correctness", and these are the counterexample.

The flag half is cheap to close here: refuse when the first operand starts with `-`, so `echo -n` costs an approval like every other shape v1 does not model. The glob and tilde halves are the same class as the word-splitting rule you called out for review — bash transforms an unquoted word four ways and this handles one of them.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -0,0 +1,404 @@
...
+  return match (part) {
+    { tag: "literal", text } => success(text)
+    { tag: "singleQuoted", text } => success(text)
+    { tag: "doubleQuoted", parts } => stringifyParts(parts, true)
+    { tag: "variable", name } => expandVariable(name, insideQuotes)
+    _ => failure("`${part.tag}` needs a shell to evaluate")
+  }
+}
+
+def expandVariable(name: string, insideQuotes: boolean): Result<string> {
```
egonSchiele, line 227: The rule is right; the test is narrower than IFS. Default IFS is space, **tab and newline**, and `value.trim()` plus `includes(" ")` misses an interior tab:

```bash
F=$'a\tb'
safeBash("echo $F")  → "a\tb\n"   # one word
bash    "echo $F"    → "a b\n"    # two
```

Confirmed by running it. `/\s/` closes it.

One more case for the same docstring: an unquoted expansion of an **empty** variable produces *zero* words in bash and one empty word here. It does not matter for `echo`; it will for `cat $F`, where an empty word becomes an argument that bash never passed.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -0,0 +1,404 @@
...
+  }
+  const item = node.items[0]
+  if (item.background) {
+    return failure("background commands are not handled")
+  }
+  // `any` on purpose: the checker resolves `item.command` to the type ALIAS
+  // named `Command` rather than the field's declared type `AndOr`, so every
+  // access below it reports a missing property. Filed separately; the shape
+  // is what the AST types say.
+  const andOr: any = item.command
```
egonSchiele, line 324: The workaround is fair — the typechecker bug is real and worth its own issue. But these three `any`s cover `andOr.rest`, `pipeline.negated`, `pipeline.commands`, `cmd.assignments`, `cmd.words` and `cmd.redirects`, and those six accesses *are* the safety argument for the whole reduction.

`cmd.tag` gets a runtime check below. The two levels above it get none, so if the parser's shape ever shifts, the reduction proceeds on unchecked assumptions rather than failing.

Two lines restore at runtime what the `any` gave up statically:

```ts
if (andOr.tag != "andOr") { return failure("unexpected AST shape") }
if (pipeline.tag != "pipeline") { return failure("unexpected AST shape") }
```

Worth it for a module whose job is deciding when a command does not need a human.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -0,0 +1,404 @@
...
+  """
+  return words.join(" ") + "\n"
+}
+
+def writeThrough(path: string, content: string, mode: WriteMode, original: string, cwd: string): SafeBashResult {
+  """
+  Write the file, falling back to bash if the write itself fails so a refusal
+  never silently swallows the command.
+  """
+  const written = write(path, content, cwd, mode)
```
egonSchiele: Two things here.

**`cwd` versus `write`'s default.** `safeBash`'s `cwd` defaults to `""`, but `write`'s `dir` parameter defaults to `"."` — so passing it explicitly means the fast path never gets that default. If `""` and `"."` resolve differently, a redirected `echo` writes somewhere `bash` would not have. Either normalize or confirm they are equivalent.

**Falling back to `bash` when the write fails.** The docstring's reasoning (a refusal should not silently swallow the command) is sound, but it is worth spelling out the consequence: a user who *denies* the `std::write` approval is then asked to approve `std::bash` for the same effect. That is a second prompt for something they just declined. Deliberate is fine — it should just be in the docstring, since it is the behavior someone will be surprised by.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -0,0 +1,404 @@
...
+  op: string;
+  path: string
+}
+
+export type Cmd = {
+  words: string[];
+  redirect: OutputRedirect | null
+}
+
+export type SafeBashResult = {
```
egonSchiele: This is structurally identical to `ExecResult` in `std::shell` (`stdout`, `stderr`, `exitCode`), which is why `return bash(command, cwd)` typechecks against it on the fallback path.

They agree coincidentally rather than provably. If `ExecResult` grows a field, the fallback path carries it, the fast path does not, and the declared return type hides the difference. Importing `ExecResult` costs nothing and makes the two paths agree by construction — which matters here, since "the fallback returns exactly what bash would have" is part of the safety claim.

egonSchiele (commented):
## Overview

The architecture is right. Refusing **by name** rather than silently is what makes a fallthrough explainable, `stringifyWordPart`'s `_ =>` correctly refuses the three expansions that would need a shell, and not wiring `shellTools()` is the correct call for a change with this shape.

The module states its invariant three times — *"Incompleteness costs an approval, never correctness."* That is the right invariant, and it is the one I went after. It does not currently hold.

Everything below is confirmed by running against a build of this branch.

## 1. The `echo` rule silently disagrees with bash on four common inputs

| input | `safeBash` returns | `bash` returns |
|---|---|---|
| `echo -n hi` | `"-n hi\n"` | `"hi"` (no newline) |
| `echo -e 'a\tb'` | `"-e atb\n"` | `"a\tb\n"` (real tab) |
| `echo *` | `"*\n"` | the file listing |
| `echo ~` | `"~\n"` | `/Users/adit\n` |

No interrupt, no diagnostic, no signal of any kind — each one takes the fast path and returns the wrong string. That is not incompleteness costing an approval; it is incompleteness costing correctness, which is the failure mode the module is built to avoid.

Two distinct causes:

**Flags.** `{ words: ["echo", ...rest] }` treats everything after `echo` as text. `echo` has `-n`, `-e` and `-E`, and they change the output rather than adding to it. The fix is small and belongs in this PR: refuse when the first operand starts with `-`, so a flagged `echo` costs an approval instead of a wrong answer.

**Unquoted-word transformations.** The module handles one of the four things bash does to an unquoted word — splitting — and misses globbing and tilde expansion. Both silently pass through as literal text.

Worth noting that `echo {a,b}` *does* fall through correctly, but by accident: the parser produces no brace node, so nothing had to decide. Brace expansion should be refused deliberately, the way pipelines and `&&` are, so it stays refused if the parser grows.

## 2. Word splitting only tests for spaces; the default IFS is space, tab and newline

`expandVariable` checks `value != value.trim()` and `value.includes(" ")`. A tab or newline in the middle passes both:

```bash
F=$'a\tb'
safeBash("echo $F")   # → "a\tb\n" as ONE word
bash    "echo $F"     # → "a b\n"  — two words
```

Confirmed by running it. The rule you asked reviewers to focus on is the right rule; it just needs `/\s/` rather than `" "`.

The same rule has a second half worth writing down: an unquoted expansion of an **empty** variable produces *zero* words in bash and one empty word here. Harmless for `echo`, not harmless for `cat $F`.

## 3. It changes which effect is raised, which is a policy change and not only an approval change

The checker spells this out on any caller:

```
AG3009: Function 'safeBash' may throw interrupts [std::write, std::bash]
```

And `capabilities.agency` sorts those into different sets:

```ts
effectSet FileWrite = <std::write, std::writeBinary, std::edit, …>
effectSet Shell     = <std::bash, std::exec, std::run>
```

So a policy that grants `FileWrite` and withholds `Shell` **blocks** `echo x > f` today and **permits** it once this is wired in. "Redirected, it is a file write, which `write` already gates" is true — but it gates a different door than the one the operator configured for shell commands, and the command string is unchanged.

That is the substantive question for the wiring-in review, and it is a property of this machinery rather than of the wiring, so it belongs in this PR's docs. The un-redirected `echo` path is the sharper version: it raises **nothing at all**, so a `Shell`-denied policy still gets output from it. Benign for `echo` specifically; worth stating as the rule that every future mapping has to satisfy.

## 4. The `any` workaround removes checking from exactly the fields that decide the reduction

`andOr`, `pipeline` and `cmd` are `any`, so `andOr.rest`, `pipeline.negated`, `pipeline.commands`, `cmd.assignments`, `cmd.words` and `cmd.redirects` are all unchecked — and those six accesses *are* the safety argument. `cmd.tag` is verified at runtime; the two levels above it are not.

Since the typechecker bug is real and worth its own fix, the cheap mitigation is to restore at runtime what the `any` gave up statically:

```ts
if (andOr.tag != "andOr") { return failure("unexpected AST shape") }
if (pipeline.tag != "pipeline") { return failure("unexpected AST shape") }
```

Two lines, and the reduction stops depending on an unchecked assumption about a parser this module does not own.

## 5. `cwd` is passed where `write`'s own default is different

`safeBash`'s `cwd` defaults to `""`; `write`'s `dir` parameter defaults to `"."`. The fast path passes `""` explicitly, so it never gets `write`'s default. If those two resolve differently, a redirected `echo` writes somewhere bash would not have. Either normalize (`cwd == "" ? "." : cwd`) or confirm they are equivalent and say so.

## 6. `SafeBashResult` duplicates `ExecResult`

Both are `{ stdout: string, stderr: string, exitCode: number }`, which is why `return bash(command, cwd)` typechecks against the declared `SafeBashResult`. They agree coincidentally rather than provably — if `ExecResult` ever grows a field, the fallback path carries it and the fast path does not, with the declared return type hiding the difference. Importing `ExecResult` costs nothing and makes the two paths agree by construction.

## 7. On the shape question you asked about

`Cmd = { words, redirect }` holds up for `cat`, `ls`, `grep` and `git` — I could not construct a rule it blocks.

The lock-in is somewhere else: every rule has to decide which flags it understands, and `Cmd` gives it no help, so each one reinvents that check. Finding 1 is the evidence — the very first rule got it wrong, in the simplest possible command. A flags/operands split on `Cmd`, or a shared "refuse any leading `-` this rule does not know" helper, is worth adding before there are four rules doing it four ways.

## Verdict

The design is sound and the refusal-by-name discipline is the right foundation. Item 1 I would treat as blocking even for an unwired module, because `echo -n` is not an exotic input and the module's whole claim is that it never trades correctness for convenience. Items 2 and 4 are small. Item 3 is the thing I would most want written into the docs before anyone reviews the one-line change to `shellTools()`, since it is the part that is easy to miss when the diff is one line long.

egonSchiele (commented):
Every correctness finding reproduced. All fixed in a28bf5566.

## The four ways a mapped command differed from bash

You were right that these are the counterexample to the module's own docstring. Confirmed by running them:

```
echo -n hi   → words ["echo","-n","hi"]  → "-n hi\n"   bash: "hi" (no newline)
echo *       → literal "*"               → "*\n"       bash: the file listing
echo ~/d     → literal "~/d"             → "~/d\n"     bash: /Users/adit/d
F=$'a\tb'    → one word                                bash: two
```

- **Globs and tilde** are refused unquoted; quoted they are ordinary text and still map. I checked the per-part approach is sound before relying on it: the parser keeps quoted and unquoted parts separate within a word, so `"a"*` is `[doubleQuoted "a", literal "*"]` and only the unquoted part is scanned.
- **Flags**: any argument starting with `-` falls through.
- **IFS**: space, tab and newline. Plus your second case — an unquoted **empty** expansion produces zero words in bash and one empty argument here, invisible for `echo` and wrong for `cat $F`. Refused.
- **cwd**: you and Copilot were both right, and it was worse than a mismatch — `write`'s `dir` *rejects* `""`, so every redirected echo fell back to bash and the mapping never fired at all.

## Two of these are the parser's, not this module's

Written up separately in the spec, since Adit will take them there.

The AST records what was *written* but not the shell's intent to *transform* it. `echo *` arrives as `literal "*"` and `~/d` as `literal "~/d"` — indistinguishable from ordinary text except by scanning for metacharacters, which is what safeBash now does. It is correct today, but every consumer asking "would bash expand this?" has to reimplement that scan. A `{ tag: "glob", pattern }` part, and something similar for tilde, would make it structural. Tilde also has a position rule (start of word only) that a consumer has to know.

## The `any` finding

Taken. `andOr` and `pipeline` now get runtime tag checks. You put it well — those six accesses *are* the safety argument, and `cmd.tag` was the only one being checked. A shifted parser shape now fails rather than being reduced on assumptions.

## ExecResult

Exported from `std::shell` and used on both paths, so they agree by construction. "The fallback returns exactly what bash would have" is part of the safety claim, and it was resting on two structurally identical types agreeing by accident.

## Write-denial double prompt

Documented rather than changed. It is deliberate — the alternative is a command that silently does nothing — but you are right that it is the behavior someone will be surprised by, so it belongs in the docstring rather than in a reviewer's head.

## One gap I am leaving open, deliberately

The flag refusal has no test. Verifying it means asserting that `safeBash("echo -n hi")` *raises* `std::bash`, and this test file has no interrupt plumbing. I checked it by hand — it does fall through — but it is not pinned, and I would rather say so than let the 6/6 imply otherwise.

The docs `redirect?:` versus `| null` mismatch and the doc-comment `*` rendering are both generator behavior; I reflowed the comment, which fixes the second. The first is worth its own look at `agency doc`.

PR #703, opened 2026-07-26
std::safeBash: follow the rewritten bash parser

egonSchiele:
tarsec 0.5.2 rewrote the bash parser, and the AST `safeBash` reduces changed shape underneath it. This follows it.

## What moved

- The top level is a flat `Command[]` — no `List`, no `items` wrapper, no `background` flag.
- A simple command carries its name in `command`, separate from `args`.
- Flags and paths are structural nodes rather than literal text.
- Globs, tilde, pipelines, background, command substitution, arithmetic and `2>&1` are rejected at parse time instead of arriving in the AST.

## The shape question

The spec flagged `Cmd` as a possible lock-in, with the first real change as its test. This is that change, and `Cmd` follows the parser rather than flattening back to a word list:

```ts
type Cmd = { command: string, args: string[], redirect: OutputRedirect | null }
```

Collapsing the name back into the args would throw away the split the parser now hands over, and a flag has no faithful text form to put in a `string[]` — the parser splits `--format=oneline` into a name and a value, and re-joining them is a guess about a syntax that varies per command. `simplify` refuses any command with a flag instead. That is the behavior the old `startsWith("-")` check had, now detected structurally.

## Dead code the parser made unreachable

Because the parser refuses what it cannot represent, the glob and tilde checks in `literalWord` and the `needs a shell to evaluate` arm can no longer fire, and are gone. This also resolves the spec's "Parser gaps" section: `echo *` no longer arrives as a literal indistinguishable from text, so nothing has to scan for shell metacharacters.

Four rows of the refusal table moved from "refused by `simplify`" to "unparseable". Both paths fall through to bash, so no command changes hands — only the reason it reports.

## One correctness fix

An unquoted empty expansion is now only refused when the variable **is** the whole word. Bash passes zero arguments for `cat $EMPTY` but one for `cat $EMPTY/bin`, so refusing the second was wrong.

## Also

- The `any` workaround for the field-name/type-alias collision is no longer needed and is dropped.
- Tests grew from 6 nodes to 8, splitting parse-time rejection from `simplify` refusal and covering flags and paths. 8/8 pass.
- Still **not wired into any agent** — `shellTools()` is untouched.

## Not in this PR

The spec's own sections still describe the old parser: the refusal table, "Parser gaps", and the flag discussion under the command table (flags being structural changes that calculus). Worth a follow-up pass.

### DIFF
--- a/packages/agency-lang/lib/parsers/parsers.ts
+++ b/packages/agency-lang/lib/parsers/parsers.ts
@@ -5065,7 +5065,7 @@ const ifBranchExprParser: Parser<Expression> = (input: string) => {
   if (kw.success) {
     return failure(
       "a branch of `if ... then ... else` cannot itself be an `if` expression " +
-      "(this includes `else if`); use `match` for multi-way or nested branching",
+        "(this includes `else if`); use `match` for multi-way or nested branching",
       input,
     );
   }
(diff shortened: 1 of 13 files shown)


packages/agency-lang/stdlib/safeBash/actions.agency
```diff
@@ -0,0 +1,76 @@
+import {
```
egonSchiele: Each command gets translated into an Action before it gets executed. This makes it possible for us to write a test for the safeBash functionality. We can write tests that pass in a bash command, or several bash commands separated by semicolons (or &&s or ||s) and see what actions get generated from them.

packages/agency-lang/stdlib/safeBash/actions.agency
```diff
@@ -0,0 +1,76 @@
...
+  cwd?: string
+}
+
+export type GitLogAction = {
+  type: "gitLog";
+  path?: string;
+  cwd?: string
+}
+
+export def printAction(content: string) {
```
egonSchiele: Modify all these functions, so they take the action objects instead of raw parameters. This function should take PrintAction, for example.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -292,88 +239,204 @@ def simplifyRedirects(redirects: Redirect[]): Result<OutputRedirect | null> {
...
+  }
+  const result = block()
+  for (key, value in oldEnv) {
+    setEnv(key, value) with approve
+  }
+  return result
+}
 
-  @param node - The AST from `bashParser`
+def makeAction(command: SimpleCommand): Result<Action> {
```
egonSchiele: I think this part can get confusing, so here's what I'm going for. A raw string gets entered into safeBash. We parse it using the bash parser and turn that into an AST, which should be an array of `Command`s. Those get run using runCommands, which in turn calls `runCommand`. If the command is a SimpleCommand, runCommand uses makeAction to turn the command into an Action and uses runAction to run it. Otherwise, if the command is an `and` or `or` or `parens` command, we use the related function to run the behavior desired. Eg the `andCommand` function runs the related commands but stops running them if one of them fails, the or command function only runs the other command if the first one fails, etc.

Note that all of these return `Result`s. It is very important to catch failures at all levels and propagate accordingly. Your initial code had set it up so that if the command fails, we revert back to bash. But the way this is set up, it would be hard to know what "reverting back" would mean. So instead, we try to return a message to the LLM saying exactly what command or commands were run, and what the output was.

If a command fails, along with returning what commands were run and their output, we should also return exactly which command failed, what the failure was, and all the commands that haven't been run yet. At this point, the commands are not strings, they are `Command` objects, so we would need to turn them back into strings using the new `_astToBash` function (used in `commandsToStr`).

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -292,88 +239,204 @@ def simplifyRedirects(redirects: Redirect[]): Result<OutputRedirect | null> {
...
+    if (path is success(text)) {
+      full.push(redirect.op, text)
+    } else {
+      return failure("redirect target: ${path.error}")
+    }
   }
-  if (pipeline.commands.length != 1) {
-    return failure("pipelines are not handled")
+
+  match(full) {
```
egonSchiele: the heart of the logic. This should return `Action` objects, not run the functions. `runAction` runs the functions.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -292,88 +239,204 @@ def simplifyRedirects(redirects: Redirect[]): Result<OutputRedirect | null> {
...
+    ["git", "log", "--oneline"] => gitLogAction()
   }
-  const cmd: any = pipeline.commands[0]
-  if (cmd.tag != "simpleCommand") {
-    return failure("`${cmd.tag}` is not a simple command")
+}
+
+export def runAction(action: Action): Result<string> {
+  const result = match(action) {
+    { type: "print", content } => printAction(content)
```
egonSchiele: we should be able to pass in the action object to each one, instead of passing individual params.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -292,88 +239,204 @@ def simplifyRedirects(redirects: Redirect[]): Result<OutputRedirect | null> {
...
+      joinWords([stringifyWord(w) for w in args]),
+      "overwrite",
+    )
+    ["echo", ...args, ">>", filename] => writeAction(
+      filename,
+      getAgentCwd(),
+      joinWords([stringifyWord(w) for w in args]),
+      "append",
+    )
+    ["echo", ...args] => printAction(joinWords([stringifyWord(w) for w in args]))
```
egonSchiele: doesn't take flags into account at the moment

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -292,177 +258,493 @@ def simplifyRedirects(redirects: Redirect[]): Result<OutputRedirect | null> {
...
+    // Nothing ran, so handing the whole string to bash repeats nothing.
+    return bashAction(
+      {
+      type: "bash",
+      command: command,
+      cwd: cwd
+    },
+    )
+  }
+  return runCommands(parsed.value)
```
egonSchiele, line 775: **Regression: the `cwd` parameter is dropped whenever the command parses.**

On this path nothing ever reads `cwd` again: `bashFallback` uses `getAgentCwd()` (line 471), `writeDir()` uses `getAgentCwd()` (line 344), and `makeAction` never sets `cwd` on a git action, so the git tools run with `cwd: ""`. The old version threaded the caller's `cwd` through every branch (`bash(command, cwd)` on the fallback and `writeThrough(..., cwd)` on the mapped path).

Concretely: `safeBash("git status", "/some/repo")` now reports status for whatever directory the agent happens to be in, and `safeBash("echo hi > f.txt", "/some/repo")` writes the file somewhere else too. Meanwhile the parse-failure path four lines up still honors `cwd`, so the same call behaves differently depending on whether the string parses.

Suggestion: thread `cwd` through `runCommands` → `makeAction` (or resolve it once at the top: caller's `cwd` if non-empty, else `getAgentCwd()`), and have `makeAction` stamp it on every action it builds.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -292,177 +258,493 @@ def simplifyRedirects(redirects: Redirect[]): Result<OutputRedirect | null> {
...
+
+  A failing left side is still a failure of the whole chain, and carries
+  its own message, so the report says which half stopped it.
+  """
+  const left = runCommand(command.left)
+  if (left is failure) {
+    return left
+  }
+  const right = runCommand(command.right)
+  if (right is failure(why)) {
```
egonSchiele, line 569: When the left side of `&&` succeeds and the right side fails, `left.value` is discarded here. `runCommands` only collects output from commands that completed as a whole, so for `echo one; cat f && badcmd` the failure report's "Output so far" shows `one` but not what `cat f` printed — even though it really ran. The `runCommands` docstring promises "what already ran and what it produced", which does not hold inside a chain.

Minor, but the fix is small: include `left.value` in the failure message here (the way `orCommand` already joins both errors).

packages/agency-lang/tests/agency/safeBash.agency
```diff
@@ -1,85 +1,165 @@
...
   ]
 }
 
-node redirectsItUnderstands() {
-  const parsed = bashParser("echo hi > out.txt")
+node emptyExpansionsFallBackToBash() {
+  // An unset variable produces NO argument in bash, where expanding it to
+  // "" here would produce an empty one. True of a whole word and of a word
+  // made only of empty parts.
+  return [
```
egonSchiele, line 127: The PR's one behavioral correctness fix — an unquoted empty expansion is only refused when the variable is the *whole* word, so `cat $EMPTY/bin` passes `/bin` as one argument — has no test. This node covers the whole-word cases (`$NOPE_NOT_SET`, `$NOPE_A$NOPE_B`) but not the mixed case that motivated the fix.

Adding `actionsOf("echo $NOPE_NOT_SET/bin")` with expected `[{"type":"print","content":"/bin\n"}]` would pin it.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -5065,7 +5065,7 @@ const ifBranchExprParser: Parser<Expression> = (input: string) => {
   if (kw.success) {
     return failure(
       "a branch of `if ... then ... else` cannot itself be an `if` expression " +
-        "(this includes `else if`); use `match` for multi-way or nested branching",
+      "(this includes `else if`); use `match` for multi-way or nested branching",
```
egonSchiele: This file's changes (here and the ~120-line `comprehensionParser` hunk below) are whitespace-only re-indentation with no connection to safeBash — it looks like a formatter version change riding along with the tarsec bump. Fine to keep if intentional, but if not, dropping it would keep `git blame` on the comprehension parser's load-bearing comments pointing at the commits that explain them.

egonSchiele (commented):
Reviewed the whole diff. The Action split (decide in `makeAction`, act in `runAction`) is a clear improvement, and the test strategy of rejecting the bash interrupt to simulate a failing command without spawning a real subprocess is nice. One real regression (the `cwd` parameter is silently dropped on the main path), plus a few smaller items as inline comments.

One non-inline note: the PR description no longer matches the code. It says `simplify` refuses any command with a flag and shows `type Cmd = { command, args, redirect }`, but the code that shipped renders flags back to text in `tokenize` so that `git diff --staged` can map to a tool call, and the type is `SimpleCommand` with `assignments` and a `redirects` list. Worth updating the description before merge so the history reads true.

PR #704, opened 2026-07-28
safeBash v3: classify the command, then let bash do the work

egonSchiele:
Replaces the v2 mapping design for `std::safeBash`. Includes the spec, the review rounds it answers, the implementation plan, and the implementation.

## Why v2 had to go

v2 substituted an Agency function for a recognized command — `git status` became `gitStatus()`. That requires the function to produce what the command produces, and `std::git` cannot: it runs machine-format argv in a scrubbed environment, so there is no text matching what `git status` prints. Matching bash byte for byte would mean matching the user's locale, `~/.gitconfig` and git version.

Measured, it didn't match:

```
safeBash("echo hi && echo bye")  →  "hi\n\nbye\n"
bash     "echo hi && echo bye"   →  "hi\nbye\n"
```

and the *shape* of the return value depended on which path ran, so an agent could tell which decision safeBash had made — the one thing the v2 spec said it shouldn't be able to do.

## What v3 does instead

**Substituting a function was never what made this safe.** `gitStatus` spawns `git` either way. What made it safe was the narrower question. That survives without the substitution, and fidelity comes free because the thing producing the output *is* bash.

```
1. Parse the string into commands.
2. Classify every command.
3. Any refused        → refuse the whole string. Nothing runs.
4. Any needs bash     → one std::bash for the whole string.
5. Otherwise          → raise every narrow effect, up front.
6. All approved       → run the WHOLE string in one bash call.
```

Three problems dissolve rather than being managed. No half-executed state, because nothing runs before every approval is in. Nobody is asked to approve a bare `cd`, because if any part needs the broad question the whole string asks once. And `cd foo && git status` works, because it is one shell.

Two paths never spawn a shell: a single fully-literal `echo` computes its own output, and a single fully-literal redirected `echo` computes its content and hands the real bytes to `write` — so `content` is in the approval payload *before* anyone approves, which is more than v2 managed.

## The trap that would have made this a no-op

`bash()` raises `std::bash` itself before calling `_bash`. So raising `std::git::status`, getting approval, then calling `bash()` asks the broad question anyway — and the narrow one buys nothing.

Execution goes through `_bash` and `_write`, the non-raising internals. Bypassing a tool's interrupt is normally forbidden, so the reasoning is written at the call site, in the spec, and in the plan's constraints:

> The narrow interrupt **replaces** the broad one. The only path to `_bash` runs through a `raiseAll` that raised every effect in the plan and returned success.

`exactlyOneQuestion` pins it by **approving** the narrow question and asserting the command ran. A test that only rejects the first interrupt passes whether or not a second prompt appears — which is exactly how this bug hides.

## Deleted

The entire runner — `runCommands`, `runCommand`, `andCommand`, `orCommand`, `parensCommand`, `commandsToStr`, `failureReport`, `joinOutput`. Bash does control flow now.

All the expansion machinery — `expandVariable`, `stringifyArg`, `stringifyParts`, `hasQuotedPart`. Bash expands, with real word-splitting and real `IFS`, including the `$A$B` case v2 got wrong.

The v2 `Action` types and executors. The extra-newline bug went with the code that caused it.

## Safety notes worth reading

**This trades a structural bound for a reviewable analysis.** v2's `echo` called `print`, which *cannot* write a file however wrong the classifier was. v3 hands a shell a string. Three things narrow it: bash receives a string re-rendered from the validated tree when we asked narrow questions (so a parser bug degrades to "we ran what we thought we read"), the command word must be a literal (so `$CMD status` can never become a git read), and the shell-free `echo` path keeps the structural bound where it was free.

**The refuse wall stops the common spelling, not the idea.** `rm`, `/bin/rm` and `git checkout .` are refused; `find . -delete` and `xargs rm` reach the ordinary `std::bash` prompt. Stated plainly in the spec rather than implied to be a guarantee.

**Two "unused" imports are load-bearing.** Effect declarations travel with imports, and without them a raise compiles with its payload contract *unenforced* — verified by deliberately breaking a payload and confirming `AG3006` fires. There's a comment saying so, because the natural instinct is to delete them.

## Tests

- 17 agency tests: classification, the refuse wall through a chain, payload values, the rebuilt-vs-original rule, and real execution (a sequence, a non-zero exit, an append that reads the file back)
- 24 property-test cases pinning `astToBash` round-trips, comparing **trees** not rendered strings — a string fixed point passes on exactly the failure it exists to catch
- Full vitest suite: 9686 passing

## Reviews

Three rounds on the spec and three on the plan, in the `-REVIEW.md` files. The findings that changed the design: the executors would have re-raised the broad interrupt; chain halves were never walked, so `true && rm -rf /` bypassed the wall by two characters; a redirect contributed nothing, so `echo pwned >> .bashrc; git status` could have ridden along under a git-read approval; and three parser assumptions were empirically false.

### DIFF
--- a/packages/agency-lang/stdlib/safeBash/actions.agency
+++ b/packages/agency-lang/stdlib/safeBash/actions.agency
@@ -11,6 +11,8 @@
   interrupts and the caller has already raised a narrower one. See the
   comment on `runBash`.
 */
+import { WriteMode } from "std::index"
+
 import { _write } from "agency-lang/stdlib-lib/builtins.js"
 import { _bash } from "agency-lang/stdlib-lib/shell.js"
 
@@ -33,7 +35,7 @@ export type WritePayload = {
   dir: string;
   filename: string;
   content: string;
-  mode: string
+  mode: WriteMode
 }
 
 export type GitDiffPayload = {
@@ -51,7 +53,7 @@ export type Execution = {
   content: string;
   filename: string;
   dir: string;
-  mode: string;
+  mode: WriteMode;
   reason: string
 }
 
--- a/packages/agency-lang/stdlib/safeBash/actions.agency
+++ b/packages/agency-lang/stdlib/safeBash/actions.agency
@@ -1,141 +1,144 @@
 /** @module
-  The things a bash command can turn into.
+  What safeBash decides, and the three ways it can carry that out.
 
-  Every command `safeBash` runs is first translated into an `Action`: a plain
-  data object saying what should happen, with nothing having happened yet.
-  Deciding *what* a command means and *doing* it are separate steps, so the
-  deciding half can be tested by handing in a string and looking at the
-  actions that come out.
+  A `Plan` is the whole decision about one call, made before anything
+  happens: which interrupts to raise, and what to run if they are all
+  approved. Deciding and doing are separate steps so the deciding half can
+  be tested by handing in a string and looking at the plan.
 
-  `bash` itself is an action, so a command with no better mapping is still
-  represented here rather than being a hole in the model.
+  The executors call the NON-RAISING internals (`_bash`, `_write`) rather
+  than the `bash` and `write` tools, because those raise their own
+  interrupts and the caller has already raised a narrower one. See the
+  comment on `runBash`.
 */
-import { gitStatus, gitDiff, gitLog } from "std::git"
-import { bash } from "std::shell"
+import { _write } from "agency-lang/stdlib-lib/builtins.js"
+import { _bash } from "agency-lang/stdlib-lib/shell.js"
 
 /** How much of a command's output to keep before truncating. */
 export static const MAX_STDOUT_LEN = 2000
 
-export type Action =
-  | PrintAction
-  | WriteFileAction
-  | GitAction
-  | BashAction
+/** One interrupt a command needs raised before it may run.
+ *
+ * A discriminated union rather than a bag of `any`: the effect set is
+ * closed, so each payload can be typed, and typing them is what makes the
+ * raise sites check that every payload satisfies its effect's contract. */
+export type Effect =
+  | { name: "std::bash" }
+  | { name: "std::write"; payload: WritePayload }
+  | { name: "std::git::status"; payload: { cwd: string } }
+  | { name: "std::git::log"; payload: { cwd: string; ref: string; path: string } }
+  | { name: "std::git::diff"; payload: GitDiffPayload }
 
-export type PrintAction = {
-  type: "print";
-  content: string
+export type WritePayload = {
+  dir: string;
+  filename: string;
+  content: string;
+  mode: string
 }
 
-export type WriteFileAction = {
-  type: "writeFile";
+export type GitDiffPayload = {
+  cwd: string;
+  ref: string;
+  ref2: string;
+  staged: boolean;
+  path: string
+}
+
+/** What happens if every effect in the plan is approved. */
+export type Execution = {
+  kind: string;
+  command: string;
+  content: string;
   filename: string;
   dir: string;
-  content: string;
-  mode: "append" | "overwrite"
+  mode: string;
+  reason: string
 }
 
-export type GitAction = GitStatusAction | GitDiffAction | GitLogAction
-
-export type GitStatusAction = {
-  type: "gitStatus";
-  cwd?: string
+/** The whole decision about one call to safeBash, made before anything runs. */
+export type Plan = {
+  effects: Effect[];
+  execution: Execution
 }
 
-export type GitDiffAction = {
-  type: "gitDiff";
-  path?: string;
-  staged?: boolean;
-  cwd?: string
-}
-
-export type GitLogAction = {
-  type: "gitLog";
-  path?: string;
-  cwd?: string
-}
-
-/** The fallback: run the command through bash exactly as written.
- *
- * `command` is the command re-rendered from its AST, not the caller's
- * original string, so a sequence that runs three commands and falls back on
- * the second sends bash only the second one. */
-export type BashAction = {
-  type: "bash";
-  command: string;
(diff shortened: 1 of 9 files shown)


packages/agency-lang/tests/agency/safeBash.agency
```diff
@@ -1,193 +1,258 @@
...
-      return why
-    }
-    return "UNEXPECTEDLY SUCCEEDED"
+  return "UNEXPECTEDLY SUCCEEDED"
+}
+
+node appendReallyAppends() {
+  // The executor is called positionally, and no other test runs it, so a
+  // swapped argument order survives everything else.
+  safeBash("echo one > __safebash_append__.txt")
```
egonSchiele, line 241: `__safebash_append__.txt` is written into the test process's working directory (`resolveCwd("")` → `"."`) and never removed, so every local run of this suite leaves a stray file for `git status` to find. The test is idempotent (`>` truncates on the first write), so it's litter rather than flakiness — but worth cleaning up at the end of the node, e.g. `remove("__safebash_append__.txt") with approve` (pleasingly, `safeBash("rm ...")` can't do it — the wall refuses).

packages/agency-lang/stdlib/safeBash/actions.agency
```diff
@@ -1,141 +1,144 @@
...
+      stdout: out
+    },
+    ),
+    )
+  }
+  return success(out)
 }
 
-export def writeAction(action: WriteFileAction): Result<string> {
+export def truncate(text: string): string {
```
egonSchiele, line 115: `truncate` is the one piece of new behavior with no test: nothing in the suite produces output longer than `MAX_STDOUT_LEN`, so deleting the cap (or the `\n[truncated]` marker) survives everything. One row does it — a single `echo` with a >2000-char literal argument goes through the fast path, which now truncates too, so the expectation is deterministic: content ending in `"\n[truncated]"`. That row would also pin the fast-path/bash-path consistency the comment in `safeBash` argues for.

packages/agency-lang/tests/agency/safeBash.agency
```diff
@@ -1,193 +1,258 @@
...
+    return "READ FAILED"
   }
-  return "UNPARSEABLE"
+  return back.value
+}
+
+node defaultCwdIsNeverEmpty() {
+  // `write` rejects an empty dir, and `getAgentCwd()` is empty whenever
+  // nothing set one — which is the case these tests run in.
+  return [resolveCwd(""), resolveCwd("/somewhere")]
```
egonSchiele, line 254: Small brittleness, fine to leave: this row asserts `resolveCwd("") == "."`, which is only true while the harness runs with no agent cwd set. If the test runner ever starts setting one, this fails with a confusing diff. The comment already states the assumption, so anyone hitting it has the explanation in hand — just flagging that the dependency is on harness config, not on the code under test.

packages/agency-lang/tests/agency/safeBash.test.json
```diff
@@ -1,200 +1,168 @@
...
       ]
     },
     {
-      "nodeName": "gitStagedIsItsOwnAction",
-      "description": "`--staged` is a different question from a bare `git diff`, so the flag has to reach the action rather than being dropped on the way.",
+      "nodeName": "shellFreeWrite",
+      "description": "A single literal redirected echo computes its content and writes it, with no shell. The `>` and `>>` rows differ only in mode, which is the difference between overwriting and appending. A variable TARGET is unrecognized entirely; a variable in the ARGUMENTS is not shell-free but is still recognized, and goes to bash raising std::write with no content \u2014 approval of the destination, not the bytes, because bash has not run yet.",
       "input": "",
-      "expectedOutput": "[{\"type\":\"gitDiff\",\"staged\":true,\"cwd\":\"\"}]",
+      "expectedOutput": "[[\"write\",\"hi\\n\",\"out.txt\",\"/repo\",\"overwrite\",[\"std::write\"]],[\"write\",\"hi\\n\",\"out.txt\",\"/repo\",\"append\",[\"std::write\"]],[\"bash\",\"\",\"\",\"\",\"\",[\"std::bash\"]],[\"bash\",\"\",\"\",\"\",\"\",[\"std::write\"]]]",
```
egonSchiele: The `echo $HOME > out.txt` row is doing quiet but important work, and the description explains it well: a variable in the *arguments* keeps the command recognized, so the string runs under `std::write` alone — a shell spawn with no `std::bash` raised. That is within the spec's rules (the redirect contributes the effect, the audit invariant holds), but it means a policy that blanket-approves `std::write` is also approving shell execution of recognized redirect commands. Worth one sentence in the spec's "What approval means here" section so policy authors learn it from the docs rather than from this fixture.

egonSchiele (commented):
Reviewed the full implementation against the spec, the plan, and all six review rounds. Every finding from those rounds is visibly absorbed, and absorbed correctly — the executors go through `_bash`/`_write` with the invariant written at the call site, `literalArgs` runs only for the git rules, the `try` result is unwrapped with a comment explaining why, the effect-visibility imports are kept with a measured justification (`AG3006` fires without them), chains are flattened for both the wall and classification, the property test compares trees, payloads and `mode` are asserted, and hard rule #1 has the whitespace-normalization test in both directions.

One thing I verified independently rather than taking on faith: the three no-handler/one-handler tests all depend on the harness failing loudly on an unexpected interrupt. Checked against `lib/templates/cli/evaluate.ts` — an interrupt past the end of `interruptHandlers` throws `Unexpected interrupt #N` (line 26–28), a message mismatch throws (line 33–38), and a test with no handlers at all leaves the raw interrupt array as `result.data`, which fails the exact match. So `exactlyOneQuestion` genuinely pins the narrow-question-replaces-broad-question property, and `echoRunsWithoutApproval` genuinely pins no-interrupt. The load-bearing tests hold.

The inline comments are all minor. This is ready to merge once CI agrees.

Issue #705, opened 2026-07-28
LSP: async call nodes take loc.col from the async keyword


egonSchiele:
An `async foo()` call node takes its `loc.col` from the `async` keyword rather than from the callee, so the LSP has a position for the call that is six characters to the left of the name it holds.

Before the semantic-token name guard, this painted `async ` -- the keyword and the following space -- in the function color. With the guard (`paintsItsOwnName` in `lib/lsp/semanticTokens.ts`), the token is dropped instead, so `foo` in `const a = async foo()` gets no semantic token at all. The TextMate grammar still colors it as a call, so nothing visibly regressed, but the call loses the type-aware coloring every other call gets.

Reproduce:

```agency
def helper(): number {
  return 1
}

node main() {
  const a = async helper()
  print(a)
}
```

`getSemanticTokens` emits no token for `helper` on the `async` line.

The same shape shows up for `match (r is success(v))`, where a lowered call is anchored on the `match` keyword -- but that one is a synthesized name and dropping it is correct. The `async` case is a genuine user-written call.

The fix is in whatever builds the `functionCall` node for an `async` call: give it the callee's own `loc`, or record the callee offset separately so `identifierSlots` can use it.

A tripwire test records the gap today -- `TRIPWIRE: cannot color the callee of an async call` in `lib/lsp/semanticTokens.test.ts`. When the loc is fixed, that test fails and should be deleted.

PR #706, opened 2026-07-28
fix(lsp): drop semantic tokens that do not match their source text

egonSchiele:
## The bug

The LSP takes a semantic token's width from the AST name:

```ts
length: slot.name.length,
```

That is right for identifiers the user typed. It is wrong for the calls that lowering synthesizes, because those carry a real source `loc` under a name that never appears in the source. `x is failure(e)` becomes a call to `isFailure` sitting on the seven characters `failure`, so the token is two characters too wide and paints the `(` and the first letter of the binder. In `stdlib/safeBash.agency:398` that is what makes the `r` of `reason` render in the function color while `eason` keeps the variable color.

It is not one lowering. Running the current encoder over every `.agency` file in `stdlib/` and `tests/agency/` -- 1105 files, 5650 tokens -- 378 tokens are painted over text that is not the name they claim, from five different lowerings:

| Synthesized name | Tokens | What it painted |
|---|---|---|
| `_guard` | 157 | `guard(` -- the keyword plus the paren |
| `isFailure` | 87 | `failure(x` -- into the binder |
| `isSuccess` | 71 | `success(v` |
| `map` / `filter` / `_pairsOf` | 40 | `[do` -- a comprehension's bracket and two letters |
| `__objectRest` | 8 | the `if` keyword |
| `matchInit$N`, `__requireLength` | 8 | assorted |

`_guard` is the largest and shows up in nearly every agent under `stdlib/agents/`.

## The fix

`getSemanticTokens` already dropped tokens that could not fit in the current buffer. This turns that bounds check into a stronger claim: emit a token only if the text at its position **is** the whole word it names.

```ts
function paintsItsOwnName(token: Token, lines: string[] | null): boolean {
  if (lines === null) return true;
  const line = lines[token.line];
  if (line === undefined) return false;
  const end = token.col + token.length;
  if (line.slice(token.col, end) !== token.name) return false;
  return !IDENTIFIER_CHAR.test(line[end] ?? "");
}
```

Three things worth knowing about it:

- **It subsumes the old bounds check.** A token past the end of a line cannot match its name, so the stale-document tests still pass unchanged.
- **The whole-word part matters.** Plain equality accepts a token that is a prefix of the word now at that position -- `helper` against a freshly typed `helperTwo` in a stale buffer -- which is the exact failure the guard exists to stop.
- **It now runs on the no-text path too.** The server passes `doc?.getText()`, which is `undefined` when the document is not in the open-documents map. That path skipped the check entirely. It now falls back to `state.info.sourceText`, the text the state was parsed from.

Dropping rather than clamping is deliberate: a clamped token paints whatever text now sits at that position, which is a wrong color on a real word. A dropped one leaves the TextMate grammar's color in place, and the grammar already colors every word this drops.

## What it costs

The guard also drops tokens for a few genuine calls, because their `loc.col` points at a keyword rather than at the callee. `const a = async double(x)` is the clear case: the token used to paint `async ` and is now dropped, so `double` gets no semantic color. That is still the better outcome, but it hides a position bug rather than fixing it, so there is a tripwire test recording it and a follow-up issue: #705.

## Verification

- `npx vitest run lib/lsp` -- 143 passed, 1 skipped.
- `pnpm typecheck` -- clean. `pnpm run lint:structure` -- clean.
- Re-ran the corpus scan after the change: 5272 tokens emitted across the same 1105 files, and zero of them cover anything other than a whole identifier.
- Seven new tests cover the `is failure` / `is success` binders, `guard(`, a comprehension, object-rest, the stale-prefix case, and a control asserting the genuine calls around a lowered pattern still get colored.

Not verified: the editor-level check with **Developer: Inspect Editor Tokens and Scopes**. The plan's manual verification steps have not been run.

## Note on the commits

The plan splits the tripwire test into its own commit; both landed in the first commit here because the test file was staged whole. The content matches the plan.

Plan and review are included in the branch under `docs/superpowers/plans/`.

### DIFF
--- a/docs/superpowers/plans/2026-07-27-semantic-token-name-guard-REVIEW.md
+++ b/docs/superpowers/plans/2026-07-27-semantic-token-name-guard-REVIEW.md
@@ -1,7 +1,6 @@
 # Review — Semantic Token Name Guard plan (2026-07-27)
 
-Plan reviewed: `/Users/adityabhargava/agency-lang/docs/superpowers/plans/2026-07-27-semantic-token-name-guard.md`
-(An identical copy sits at `/Users/adityabhargava/agency-lang/packages/agency-lang/2026-07-27-semantic-token-name-guard.md`. Two copies of the same plan will drift. Keep the one under `docs/superpowers/plans/` and delete the other.)
+Plan reviewed: `docs/superpowers/plans/2026-07-27-semantic-token-name-guard.md`
 
 **Overall:** the diagnosis is right, the fix is the right size, and the "drop rather than clamp" reasoning holds up. I verified every code claim in the plan against the files, and I ran the current encoder over 1105 Agency files to measure what the proposed guard would actually do. Three things need to change before this is executed. One of them makes Task 1's own tests fail at the step where the plan says they pass.
 
--- a/docs/superpowers/plans/2026-07-27-semantic-token-name-guard-REVIEW.md
+++ b/docs/superpowers/plans/2026-07-27-semantic-token-name-guard-REVIEW.md
@@ -1,0 +1,117 @@
+# Review — Semantic Token Name Guard plan (2026-07-27)
+
+Plan reviewed: `/Users/adityabhargava/agency-lang/docs/superpowers/plans/2026-07-27-semantic-token-name-guard.md`
+(An identical copy sits at `/Users/adityabhargava/agency-lang/packages/agency-lang/2026-07-27-semantic-token-name-guard.md`. Two copies of the same plan will drift. Keep the one under `docs/superpowers/plans/` and delete the other.)
+
+**Overall:** the diagnosis is right, the fix is the right size, and the "drop rather than clamp" reasoning holds up. I verified every code claim in the plan against the files, and I ran the current encoder over 1105 Agency files to measure what the proposed guard would actually do. Three things need to change before this is executed. One of them makes Task 1's own tests fail at the step where the plan says they pass.
+
+---
+
+## What I verified
+
+All of the plan's factual claims check out:
+
+- `lib/lsp/semanticTokens.ts:184` sets `length: slot.name.length`. The line numbers the plan cites for the `Token` type (69-75), `toToken` (167-188), `fitsInCurrentText` (203-221) and the filter chain (244-248) are all correct.
+- `lib/lowering/patternLowering.ts:1277-1288` builds the `isSuccess` / `isFailure` call and hands it the source `loc`, exactly as quoted.
+- `lib/lsp/server.ts:362` passes `doc?.getText()`, which is `undefined` when the document is not in the open-documents map. That path really is unguarded today.
+- `state.info.sourceText` exists (`lib/compilationUnit.ts:136`) and the LSP path really does populate it: `lib/lsp/diagnostics.ts:228` passes `source`, which is `doc.getText()` (`lib/lsp/diagnostics.ts:103`). So Task 2's fallback has something to fall back to.
+- Only two node kinds produce identifier slots at all (`variableName`, `functionCall` in `lib/utils/identifierSlots.ts`), and both take `col` from the identifier's own start. That is why a name comparison is safe for ordinary hand-written code.
+- `LONG` in the existing `getSemanticTokens against a changed document` block contains no lowered construct, so the plan is right that those three tests keep passing.
+
+To measure the blast radius I temporarily added a probe test that ran the current encoder over every `.agency` file in `stdlib/` and `tests/agency/` (1105 files, 5650 emitted tokens), matched each emitted token back to the slot it came from, and counted how many would be dropped by `slice === name`. The probe file has been deleted; the numbers below come from that run.
+
+---
+
+## Finding 1 (blocking): Task 1's tests will not pass at the end of Task 1
+
+The test helper the plan builds on calls the **one-argument** form:
+
+```ts
+// lib/lsp/semanticTokens.test.ts:89-91
+function tokensFor(source: string): DecodedToken[] {
+  return decodeTokens(getSemanticTokens(stateFor(source)).data, source);
+}
+```
+
+No `currentText`. So `lines` is `null` and the guard returns `true` for everything — that is true of `fitsInCurrentText` today and would stay true of `paintsItsOwnName` after Task 1.
+
+All four of Task 1's tests go through `tokensFor` or `textsFor` (and `textsFor` calls `tokensFor`). They will still fail at Task 1 Step 7, where the plan says "4 passed". They only start passing after Task 2 Step 3 adds the `state.info.sourceText` fallback.
+
+This also makes the sentence in Task 2 Step 2 wrong:
+
+> The other tests in the block pass, because `tokensFor` goes through the two-argument form.
+
+It does not. And it makes Task 2's new test ("drops a synthesized token even when no current text is passed") an exact duplicate of the first Task 1 test — same source, same assertion, same code path.
+
+Three ways out. My recommendation is the first:
+
+1. **Merge the two tasks into one.** The split was justified by "Task 1 is the behavior fix and can ship alone", but the tests cannot demonstrate Task 1 alone, and the source-text fallback is three lines. One task, one commit.
+2. Keep the split, and have Task 1's tests pass the source explicitly: `decodeTokens(getSemanticTokens(stateFor(src), src).data, src)`. Then Task 2's test stays meaningful as the no-text case.
+3. Change `tokensFor` to pass the source. This quietly changes what every existing test in the file exercises, so I would not.
+
+## Finding 2 (blocking): the guard drops far more than the three cases the plan lists
+
+The plan names `isSuccess`, `isFailure` and the object-rest helper, and says future synthesized names are covered "without a per-site audit". That is the right design, but the plan reads as though this is a three-case fix. Measured over the corpus, 382 of 5650 tokens are dropped, and the largest source is not on the plan's list:
+
+| Synthesized name | Tokens dropped | Where it comes from |
+|---|---|---|
+| `_guard` | 157 | the `guard` keyword construct |
+| `isFailure` | 87 | `is failure(...)` lowering |
+| `isSuccess` | 71 | `is success(...)` lowering |
+| `map` / `filter` / `_pairsOf` | 40 | list-comprehension lowering |
+| `__objectRest` | 8 | object-rest patterns |
+| `matchInit$N` | 4 | match lowering |
+| `__requireLength` | 4 | array-length patterns |
+
+`_guard` is the single biggest case and is visible in almost every agent in `stdlib/agents/`: the source says `guard(`, the synthesized name is `_guard` (six characters against five), so the token paints the opening paren. `stdlib/agents/coding.agency:128` is a live example.
+
+The comprehension case is the most visually wrong of all, because the synthesized `map` call is hung on the loc of the whole comprehension. In `tests/agency/comprehensions/basic.agency:17`, `const doubled = [double(x) for x in xs]`, the emitted token paints `[do` — the opening bracket and two letters.
+
+Please fold these into the plan's bug table, and add a `guard(...)` case and a comprehension case to the tests. They are separate lowerings from pattern lowering, so a test that only covers `is failure` does not pin them.
+
+## Finding 3 (should fix): the guard also drops tokens for genuine calls, and the plan does not say so
+
+Not every dropped token is synthetic. Some are real user-written calls whose `loc.col` points at a keyword rather than at the callee. `async` is the clearest:
+
+```
+tests/agency/substeps/async-in-if.agency:8   const a = async double(x)
+```
+
+The slot's name is `double` (six characters) but its column is the column of `async`, so the token currently paints `async ` — the keyword and the following space. After the guard, that token is dropped and `double` gets no semantic token at all.
+
+There is a second family: match statements whose scrutinee is an `is` pattern, such as `match (r is success(v))` (`tests/agency/pattern-matching/resultPatternMatchIs.agency:11`), where a lowered `failure` call is anchored on the `match` keyword.
+
+Dropping is still the right outcome — a wrong color on a keyword is worse than no color on a call the TextMate grammar already colors. But two things follow:
+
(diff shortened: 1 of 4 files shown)


packages/agency-lang/lib/lsp/semanticTokens.ts
```diff
@@ -200,29 +203,40 @@ function byPosition(a: Token, b: Token): number {
...
+function paintsItsOwnName(token: Token, lines: string[] | null): boolean {
   if (lines === null) return true;
   const line = lines[token.line];
   if (line === undefined) return false;
-  return token.col + token.length <= line.length;
+  const end = token.col + token.length;
+  if (line.slice(token.col, end) !== token.name) return false;
+  // Equality alone accepts a token that is a PREFIX of the word now at
+  // that position — `helper` against a freshly typed `helperTwo`.
+  return !IDENTIFIER_CHAR.test(line[end] ?? "");
```
egonSchiele, line 233: **The boundary check is one-sided.** This checks the character *after* the token but not the one *before*, so a token can still be emitted while sitting inside a longer word — as its suffix rather than its prefix.

Repro, run against this branch:

```ts
// state built from `  helper()`, buffer is now ` xhelper()`
const before = `def helper(): number {\n  return 1\n}\n\nnode main() {\n  helper()\n}`;
const after = before.replace("  helper()", " xhelper()");
getSemanticTokens(stateFor(before), after);
// emits L5C2 len 6 -> slices to "helper", the tail of `xhelper`
```

`helper` still starts at the recorded column (one space deleted, one character inserted), `line[8]` is `(` so the trailing check passes, and the token paints the last six characters of `xhelper` in the function color. That is the same failure mode as the `helperTwo` case the tests already cover, just approached from the other side.

The fix is the symmetric condition:

```ts
if (IDENTIFIER_CHAR.test(line[token.col - 1] ?? "")) return false;
return !IDENTIFIER_CHAR.test(line[end] ?? "");
```

`line[-1]` is `undefined` at column 0, so the `?? ""` already handles the start-of-line case.

Minor in practice — it needs a stale buffer and an edit that inserts and deletes on the same line, and it self-corrects on the next debounce. But the function is named `paintsItsOwnName`, and in this case it returns true when it demonstrably does not.

packages/agency-lang/lib/lsp/semanticTokens.test.ts
```diff
@@ -307,3 +319,72 @@ describe("getSemanticTokens known gaps", () => {
...
+
+describe("getSemanticTokens against synthesized names", () => {
+  // Lowering builds calls whose names never appear in the source but
+  // which carry a real source loc — `is failure(x)` becomes a call to
+  // `isFailure` sitting on the seven characters `failure`, `guard(...)`
+  // becomes `_guard`, a comprehension becomes `map`. Length comes from
+  // the name, so those tokens ran past the word they sat on.
+
+  /** Every token must cover exactly one whole identifier. */
+  function expectWholeWords(source: string): void {
```
egonSchiele, line 336: **This assertion cannot distinguish "dropped the right token" from "dropped everything".** The loop body never runs on an empty array, so any fixture that stops emitting tokens entirely passes silently. Four of the seven new tests are built on it.

It does catch today's regression — I checked out the branch and reverted the guard, and the `guard(` fixture emits `guard(`, which fails the shape match. So this is not a hole in the current fix; it is a hole in what the tests will catch *next*. An over-broad change to `paintsItsOwnName` — say a boundary check with an inverted condition — would drop every token in these four fixtures and all four would still pass.

Pinning the exact list costs one line per test and makes them fail in both directions. I ran the fixtures against this branch, so these are the real values:

```ts
// guard(
expect(textsFor(source)).toEqual(["print"]);
// comprehension
expect(textsFor(source)).toEqual(["double", "print"]);
// object-rest
expect(textsFor(source)).toEqual(["print"]);
// is failure / is success
expect(textsFor(source)).toEqual(["pass", "print"]);
```

The comprehension case is the one I would most want pinned: `double` is a genuine user-written call *inside* the lowered construct, and `["double", "print"]` is the only assertion here that would notice if the guard started eating it.

Worth keeping `expectWholeWords` as well — it is the assertion that states the actual invariant, and it stays useful for fixtures where the exact list is uninteresting.

docs/superpowers/plans/2026-07-27-semantic-token-name-guard-REVIEW.md
```diff
@@ -0,0 +1,117 @@
+# Review — Semantic Token Name Guard plan (2026-07-27)
+
+Plan reviewed: `/Users/adityabhargava/agency-lang/docs/superpowers/plans/2026-07-27-semantic-token-name-guard.md`
```
egonSchiele: These are absolute paths from one machine, in a file that is now checked into the repo — they will not resolve for anyone else. Repo-relative (`docs/superpowers/plans/...`) reads the same and survives the trip.

The parenthetical also instructs the reader to delete `packages/agency-lang/2026-07-27-semantic-token-name-guard.md`, which is not in this diff and not in the tree — so the instruction now points at nothing, and a future reader has to go looking to find that out. Since the duplicate is already gone, this note can go with it.

egonSchiele (commented):
## Review

The diagnosis and the fix are right, and the implementation is smaller than the problem it solves — replacing a bounds check with a name check subsumes the old behavior rather than layering on it, so nothing about the stale-document path needed rethinking. The corpus measurement (382 tokens across 1105 files, five distinct lowerings) is what turns this from an anecdote about one `r` into a justified change, and the tripwire test for the `async` loc is the right way to keep a hidden bug from staying hidden.

I verified the central claims against the branch rather than reading only the diff:

- Checked out the head and ran the new fixtures through the post-fix encoder. They emit `["print"]`, `["double", "print"]`, `["print"]`, `["pass", "print"]`, `["pass", "print"]` respectively — so the guard drops the synthesized tokens without taking the genuine ones with it, which is the thing most likely to have gone wrong.
- Confirmed the whole-word check is load-bearing: with state built from `  helper()` and current text `  helperTwo()`, the token is dropped. Bare equality would have painted `helper` inside `helperTwo`.

One confirmed defect (minor, inline below): the word-boundary check is one-sided. The committed review asked for a check at `col - 1` as well as at `col + length`, and only the trailing one shipped. I have a working repro in the inline comment.

The other inline notes are about test strength and the committed docs. Nothing here is merge-blocking.

### Smaller things, not worth inline comments

- **The token counts disagree with each other and with their own table.** The PR body says 378, the plan says 382, and both tables sum to 371 (157+87+71+40+8+8). Probably a filter difference between the run that produced the table and the run that produced the headline, but as written a reader cannot reconcile them. Worth one number, stated once.
- **`line.slice(token.col, end) !== token.name` allocates a string per token** to throw it away. `!line.startsWith(token.name, token.col)` is the same check without the allocation. At 5650 tokens this is noise, so only worth taking if you are touching the line anyway.
- **Manual verification is still unrun**, as the PR body says. Step 3 (the `guard(` case in `stdlib/agents/coding.agency:128`) is the one I would not skip — it is 40% of the change by volume, and it is the case a user is most likely to notice if the guard misses.

packages/agency-lang/lib/lsp/semanticTokens.ts
```diff
@@ -200,29 +203,40 @@ function byPosition(a: Token, b: Token): number {
   return a.col - b.col;
 }
 
+const IDENTIFIER_CHAR = /[A-Za-z0-9_$]/;
+
 /**
- * Would this token land on text that still exists?
+ * Does this token paint the exact word it names?
```
egonSchiele, line 209: Can you make this comment more concise and more readable?

packages/agency-lang/lib/lsp/semanticTokens.ts
```diff
@@ -200,29 +203,40 @@ function byPosition(a: Token, b: Token): number {
   return a.col - b.col;
 }
 
+const IDENTIFIER_CHAR = /[A-Za-z0-9_$]/;
```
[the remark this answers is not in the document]
egonSchiele: `_` is already in the character class: `const IDENTIFIER_CHAR = /[A-Za-z0-9_$]/`. It is the third character from the end.

I checked the behavior rather than just the regex. With state built from `  helper()` and a buffer now reading `  helper_two()`, the guard emits nothing — `_` is treated as part of the word, which is the intended result. There is now a test pinning that: `treats an underscore as part of the word`.

`$` is a deliberate superset of `LEGAL_IDENTIFIER` (`/^[A-Za-z_][A-Za-z0-9_]*$/`). Lowering synthesizes names like `matchInit$4`, so `$` has to count as part of a word here even though a user cannot type one.

packages/agency-lang/lib/lsp/semanticTokens.test.ts
```diff
@@ -307,3 +319,72 @@ describe("getSemanticTokens known gaps", () => {
...
+  it("drops a token that is only a prefix of the word now at its position", () => {
+    // A stale buffer, mid-rename: state says `helper` at a position
+    // where the user has since typed `helperTwo`. The slice matches the
+    // name, so a bare equality check would paint the first six letters
+    // of a longer word — the very failure this guard exists to stop.
+    const before = `def helper(): number {\n  return 1\n}\n\nnode main() {\n  helper()\n}`;
+    const after = before.replace("  helper()", "  helperTwo()");
+    const data = getSemanticTokens(stateFor(before), after).data;
+    expect(decodeTokens(data, after).filter((t) => t.line === 5)).toEqual([]);
+  });
```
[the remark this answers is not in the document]
egonSchiele: Two halves to this, and they land differently.

The `_` half rests on a premise that is not true — `_` is already in `IDENTIFIER_CHAR`, so `helper_two` was never accepted as a whole-word match for `helper`. I added the case anyway, since it costs one line and pins behavior that was previously only implied: `treats an underscore as part of the word`.

The "match that starts inside a larger identifier" half was a real bug, independently found by @egonSchiele on the implementation file. The check was trailing-only, so `helper` inside `xhelper` passed. Fixed in a34ccdec with the symmetric leading check, covered by `drops a token that is only a suffix of the word now at its position`.

packages/agency-lang/lib/lsp/semanticTokens.ts
egonSchiele: Confirmed and fixed in a34ccdec. I ran your repro against the branch before changing anything: it emitted `{"line":5,"col":2,"len":6,"text":"helper"}` — the tail of `xhelper`, exactly as you described.

Took the symmetric condition as written. Added `drops a token that is only a suffix of the word now at its position` alongside the existing prefix test, both built on the same stale-buffer fixture.

One thing worth recording, since it was the only real question about the change: does the leading check cost any genuine tokens on unedited source? It does not. I compared both conditions over the same 1105 files in one process — 12414 slots pass the trailing-only check, 12414 pass both. Delta zero. It only ever fires on a stale buffer, which is the case it was written for.

packages/agency-lang/lib/lsp/semanticTokens.test.ts
egonSchiele: Agreed, and fixed in a34ccdec. The four fixtures now pin their exact lists, and I ran each one against the branch to get the values rather than copying them — they match yours:

- `guard(` → `["print"]`
- comprehension → `["double", "print"]`
- object-rest → `["print"]`
- `is failure` / `is success` → `["pass", "print"]`

`expectWholeWords` stayed, for the reason you gave, and its doc comment now says why the exact list has to accompany it: the loop body never runs on an empty array.

The object-rest test also kept its `line === 2` assertion — that one is about *where* the dropped token was, so the exact list does not subsume it.

docs/superpowers/plans/2026-07-27-semantic-token-name-guard-REVIEW.md
egonSchiele: Fixed in a34ccdec. The header is now `docs/superpowers/plans/2026-07-27-semantic-token-name-guard.md`, and the parenthetical about the duplicate is gone — the duplicate was deleted before the branch existed, so the instruction pointed at nothing. Checked the whole file for absolute paths; that header was the only one.

packages/agency-lang/lib/lsp/semanticTokens.ts
egonSchiele: Cut in a34ccdec, from fourteen lines to twelve and in plainer language. It now reads:

```
 * Does this token paint the exact word it names?
 *
 * It can fail to for two reasons. The token may be STALE, computed
 * against an older version of the document than the one being rendered.
 * Or its name may be SYNTHESIZED: lowering turns `x is failure(e)` into
 * a call to `isFailure` that keeps the loc of `failure`, and the token's
 * length comes from the longer name.
 *
 * Either way, drop the token rather than trimming it to fit. A trimmed
 * token paints a real word the wrong color; a dropped one leaves the
 * grammar's color in place, and the grammar already colors every word
 * this drops.
```

Dropped the DocumentStateCache aside and the "clamped" wording, which were the two places a reader had to already know something to follow the sentence.

egonSchiele:
Pushed a34ccdec addressing the review.

**Fixed:**
- **Boundary check was one-sided** (@egonSchiele). Your repro reproduced exactly — `helper` emitted as the tail of `xhelper`. Took the symmetric condition. New test covers the suffix case next to the existing prefix one.
- **Vacuous shape assertions** (@egonSchiele). All four fixtures now pin their exact token lists; `expectWholeWords` stays, with a comment saying why the list has to accompany it.
- **Absolute paths in the review doc** (@egonSchiele). Now repo-relative, and the stale note about the deleted duplicate is gone.
- **The `paintsItsOwnName` comment** (@egonSchiele). Cut and rewritten in plainer language.
- **Underscore coverage** (Copilot). Added a test, though the premise was wrong — see below.

**Not a bug:** Copilot's claim that `_` is missing from `IDENTIFIER_CHAR`. The class is `/[A-Za-z0-9_$]/`; `_` is in it, and `helper_two` was already dropped. `$` is a deliberate superset of `LEGAL_IDENTIFIER` because lowering synthesizes names like `matchInit$4`.

**One correction to the PR body above:** it says 5272 tokens emitted across the corpus after the change. The current number is 5263, measured with the final guard. The difference is not the leading-boundary check — I A/B'd both conditions over the same 1105 files in a single process and the delta is zero, so the leading check costs nothing on unedited source and only fires on a stale buffer. I have not tracked down what moved the earlier figure, so treat 5263 as the measured number and the earlier one as unreliable.

`npx vitest run lib/lsp`: 144 passed, 1 skipped. `pnpm typecheck` and `pnpm run lint:structure` clean.

PR #707, opened 2026-07-28
safeBash: refactor to match expressions, pattern matching, and pipes

egonSchiele:
Behavior-preserving refactor of `std::safeBash` to match expressions, pattern matching, and the pipe operator, following up on the v3 merge. **Every test fixture is byte-identical** — only the test helpers that read `Execution` fields changed shape, which is the proof the observable behavior did not.

## What changed

**`Execution` is a discriminated union** — `BashExec | EchoExec | WriteExec | RefuseExec` — instead of a flat record with sentinel fields. `emptyExecution()` and its `kind: ""` / `mode: "overwrite"` placeholders are gone; each plan constructor builds exactly the variant it means. The dispatch at the tail of `safeBash` is an exhaustive match over the union, so a fifth execution kind is a **type error at the dispatch**, not a silent fall-through to bash.

**`flattenOne` is exhaustive over `Command`** (hybrid form — see gaps below), so a fifth command shape is a type error rather than silently flattening to nothing. That function feeds the refuse wall, where silently-empty is the one result that must never happen.

**`gitEffects` is a guarded rule table** that reads like the spec's own table:

```ts
return match(sub) {
  "status" if (flags.length == 0) => ...
  "diff"   if (flags.length == 0) => ...
  "diff"   if (flags.length == 1 && flags[0] == "--staged") => ...
  _ => failure("no rule for `git ${sub}`")
}
```

**The linear Result chains use `|>`**: `effectsFor`'s git arm is `literalArgs(command) |> gitEffects.partial(flags:, cwd:) |> withRedirect.partial(redirect:)`, and `echoPlan` / `writePlan` pipe their words into plan constructors. The `x is failure(why) → return failure(why)` blocks those chains replaced are gone. Where two Results have to combine (`writePlan`'s target + words) the explicit unwrap stays, because a pipe is a chain and that isn't one.

Also: `literalWordText` and `writeMode` are matches, `isDestructiveGit` inlines its subcommand table as arms, and `bashPlan` collapses to one construction via an if-then-else expression.

## Two compiler gaps found on the way (now in TODO.md)

1. **A single-element array pattern arm after an expression arm parses as an index.** `failure("a")` followed by `[sub] =>` on the next line parses as `failure("a")[sub]`. Multi-element patterns are immune (`expr["a", "b"]` is not index syntax), which is why v2's `["git", "status"]` arms worked by accident. This is why `gitEffects` matches on the subcommand string rather than on `args` as an array.
2. **Match arms do not narrow union scrutinees or binders.** `{ tag: "simpleCommand" } => [command]` types `command` as the full union, and a `payload` bound in a `{ name: "std::git::log" }` arm gets the *wrong variant's* payload type — while `if` chains narrow both correctly. Exhaustiveness checking (AG5002) works well; arm narrowing is the missing half. This is why `raiseOne` stays an `if` chain (the exhaustive match version was the single biggest prize of this refactor — its trailing runtime `"no raise site"` failure would have become a compile error), why `flattenOne` handles the simple case in an `if` before an exhaustive match over the rest, and why the write arm in `safeBash` rebuilds its `WriteExec` from bound fields. All three carry comments pointing back here.

## Verified

- 18/18 agency tests, fixtures untouched
- 24/24 `astToBash` round-trip property tests
- `make` clean (no new diagnostics), structural linter clean, walker tripwire green
- Docs regenerated via `make doc` (safeBash pages only)

### DIFF
--- a/packages/agency-lang/stdlib/safeBash.agency
+++ b/packages/agency-lang/stdlib/safeBash.agency
@@ -282,18 +282,18 @@ export def effectsFor(command: SimpleCommand, cwd: string): Result<Effect[]> {
   // not be demoted by a check that only git needs.
   return match(name.text) {
     "echo" => success(redirect.value)
-    "git" => literalArgs(command)
-    |> gitEffects.partial(flags: flags, cwd: cwd)
-    |> withRedirect.partial(redirect: redirect.value)
+    "git" => {
+      // Both links can fail, which is what earns the pipe. Appending the
+      // redirect contribution cannot, so it stays inline rather than
+      // becoming a third link that pretends it could.
+      const git = literalArgs(command) |> gitEffects.partial(flags: flags, cwd: cwd)
+      if (git is failure(why)) {
+        return failure(why)
+      }
+      return success([...git.value, ...redirect.value])
+    }
     _ => failure("no rule for `${name.text}`")
   }
-}
-
-def withRedirect(effects: Effect[], redirect: Effect[]): Result<Effect[]> {
-  """
-  A command's own effects with its redirect's contribution appended.
-  """
-  return success([...effects, ...redirect])
 }
 
 def gitEffects(args: string[], flags: string[], cwd: string): Result<Effect[]> {
@@ -310,7 +310,7 @@ def gitEffects(args: string[], flags: string[], cwd: string): Result<Effect[]> {
   }
   const sub = args[0]
   return match(sub) {
-    "status" if (flags.length == 0) => return success(
+    "status" if (flags.length == 0) => success(
       [
       {
       name: "std::git::status",
@@ -320,7 +320,7 @@ def gitEffects(args: string[], flags: string[], cwd: string): Result<Effect[]> {
     },
     ],
     )
-    "log" if (flags.length == 0) => return success(
+    "log" if (flags.length == 0) => success(
       [
       {
       name: "std::git::log",
@@ -332,7 +332,7 @@ def gitEffects(args: string[], flags: string[], cwd: string): Result<Effect[]> {
     },
     ],
     )
-    "diff" if (flags.length == 0) => return success(
+    "diff" if (flags.length == 0) => success(
       [
       {
       name: "std::git::diff",
@@ -346,7 +346,7 @@ def gitEffects(args: string[], flags: string[], cwd: string): Result<Effect[]> {
     },
     ],
     )
-    "diff" if (flags.length == 1 && flags[0] == "--staged") => return success(
+    "diff" if (flags.length == 1 && flags[0] == "--staged") => success(
       [
       {
       name: "std::git::diff",
@@ -537,16 +537,16 @@ def echoPlan(command: SimpleCommand): Result<Plan> {
   if (flagNames(command).length > 0) {
     return failure("`echo` with flags prints differently")
   }
-  return echoWords(command) |> echoPlanFromWords
-}
-
-def echoPlanFromWords(words: string[]): Result<Plan> {
+  const words = echoWords(command)
+  if (words is failure(why)) {
+    return failure(why)
+  }
   return success(
     {
     effects: [],
     execution: {
       kind: "echo",
-      content: echoOutput(words)
+      content: echoOutput(words.value)
     }
   },
   )
@@ -578,22 +578,17 @@ def writePlan(command: SimpleCommand, cwd: string): Result<Plan> {
   if (target is failure(why)) {
     return failure(why)
   }
-  return echoWords(command)
-  |> writePlanFromWords.partial(filename: target.value, dir: cwd, mode: writeMode(only.op))
-}
-
-def writePlanFromWords(
-  words: string[],
-  filename: string,
-  dir: string,
-  mode: WriteMode,
-): Result<Plan> {
-  const content = echoOutput(words)
+  const words = echoWords(command)
+  if (words is failure(why)) {
+    return failure(why)
+  }
+  const content = echoOutput(words.value)
+  const mode = writeMode(only.op)
   const effect: Effect = {
     name: "std::write",
     payload: {
-      dir: dir,
-      filename: filename,
+      dir: cwd,
+      filename: target.value,
       content: content,
       mode: mode
     }
@@ -603,8 +598,8 @@ def writePlanFromWords(
     effects: [effect],
     execution: {
       kind: "write",
-      filename: filename,
-      dir: dir,
+      filename: target.value,
+      dir: cwd,
       content: content,
       mode: mode
     }
@@ -865,6 +860,15 @@ export def safeBash(
     return failure(why)
   }
 
+  // The write case is an `if` rather than a match arm for the same
+  // reason as `flattenOne`: match arms do not narrow the scrutinee, and
+  // rebuilding a WriteExec from bound fields would silently drop any
+  // field the type gains later. `if` narrowing passes the value through
+  // whole.
+  if (exec.kind == "write") {
+    return runWrite(exec)
+  }
+
   // Echo output is truncated like every other path: a long echo returning
   // in full while a long `ls` came back capped would be exactly the
   // distinguishability this design exists to remove. For bash, `planFor`
@@ -872,21 +876,11 @@ export def safeBash(
   // questions, original when it asked the broad one. There is no fallback
   // here: choosing the original AFTER raising narrow effects would break
   // the design's first hard rule.
-  // The write arm rebuilds a WriteExec from its bindings because match
-  // arms do not narrow the scrutinee: `runWrite(exec)` does not typecheck
-  // inside the arm, but a literal built from the bound fields does.
   return match(exec) {
     { kind: "echo", content } => success(truncate(content))
-    { kind: "write", filename, dir as target, content, mode } => runWrite(
-      {
-      kind: "write",
-      filename: filename,
-      dir: target,
-      content: content,
-      mode: mode
-    },
-    )
     { kind: "bash", command as text } => runBash(text, dir)
+    // Unreachable — the early return above already handled refusal. The
+    // arm exists only to keep the match exhaustive over Execution.
     { kind: "refuse", reason } => failure(reason)
   }
 }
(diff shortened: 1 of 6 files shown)


packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -275,35 +274,32 @@ export def effectsFor(command: SimpleCommand, cwd: string): Result<Effect[]> {
...
-  const args = literalArgs(command)
-  if (args is failure(why)) {
-    return failure(why)
-  }
-  const git = gitEffects(args.value, flags, cwd)
-  if (git is failure(why)) {
-    return failure(why)
-  }
-  return success([...git.value, ...redirect.value])
+def withRedirect(effects: Effect[], redirect: Effect[]): Result<Effect[]> {
```
egonSchiele: **This function only exists to be piped into, and it can never fail.**

```ts
def withRedirect(effects: Effect[], redirect: Effect[]): Result<Effect[]> {
  return success([...effects, ...redirect])
}
```

It is `concat` wearing a `Result`. The `success(...)` wrapper is there purely so the value fits a pipe chain, which means a reader has to check whether it has a failure mode before they can rule one out — and it doesn't.

Same shape in `echoPlanFromWords` (line 543) and `writePlanFromWords` (line 585): both always return `success`, both were extracted so a pipe had a target.

The pipe operator earns its place when it short-circuits a chain of things that genuinely fail. Here two of the three links can't, so what it bought is a chain that reads uniformly at the cost of three functions whose type signature overstates what they do.

I'd keep the pipe where every link is fallible and let the always-succeeding steps stay inline. `withRedirect` in particular reads as a smaller change than `success([...git, ...redirect])` at the call site.

The name is also slightly off — it takes `redirect: Effect[]` and appends, so it is `withExtraEffects` rather than anything redirect-specific.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -883,30 +851,44 @@ export def safeBash(
...
+  // already decided which text to run — rebuilt when it asked narrow
+  // questions, original when it asked the broad one. There is no fallback
+  // here: choosing the original AFTER raising narrow effects would break
+  // the design's first hard rule.
+  // The write arm rebuilds a WriteExec from its bindings because match
+  // arms do not narrow the scrutinee: `runWrite(exec)` does not typecheck
+  // inside the arm, but a literal built from the bound fields does.
+  return match(exec) {
+    { kind: "echo", content } => success(truncate(content))
+    { kind: "write", filename, dir as target, content, mode } => runWrite(
```
egonSchiele: **The rebuild can silently drop a field.**

```ts
{ kind: "write", filename, dir as target, content, mode } => runWrite({
  kind: "write", filename: filename, dir: target, content: content, mode: mode
})
```

The comment explains *why* the rebuild exists (arms don't narrow), and that's accurate. But the failure mode it creates isn't stated: if `WriteExec` ever gains a field, this reconstruction keeps compiling and quietly hands `runWrite` a value missing it. Nothing fails, and the field just isn't there.

This PR already has the fix for exactly this problem, used twice: `flattenOne` handles its narrowing-sensitive case in an `if` before the exhaustive match, and `raiseOne` stays an `if` chain for the same reason. Doing that here too —

```ts
if (exec is { kind: "write" }) {
  return runWrite(exec)
}
return match(exec) { ... }
```

— passes the value through instead of rebuilding it, and keeps the match exhaustive over what remains. It's the same hybrid the PR description already defends.

Worth it because this is the one arm where the workaround can lose data rather than just being verbose.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -313,8 +309,8 @@ def gitEffects(args: string[], flags: string[], cwd: string): Result<Effect[]> {
     return failure("only a bare git subcommand is recognized")
   }
   const sub = args[0]
-  if (sub == "status" && flags.length == 0) {
-    return success(
+  return match(sub) {
+    "status" if (flags.length == 0) => return success(
```
egonSchiele: **Two arm forms in one match.** The four guarded arms are `=> return success(...)` and the catch-all is `=> failure(...)`.

Per the guide, `return` inside an arm returns *from the match*, not from the function — so both forms do the same thing here and the `return` is doing nothing. But a reader who doesn't know that rule reads the first four as early returns from `gitEffects` and the last as a value, which is a meaningful-looking difference that isn't one.

Dropping the `return`s makes all five arms the same kind of thing and makes the guarded rule table read the way the PR description describes it.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -883,30 +851,44 @@ export def safeBash(
...
+      {
+      kind: "write",
+      filename: filename,
+      dir: target,
+      content: content,
+      mode: mode
+    },
+    )
+    { kind: "bash", command as text } => runBash(text, dir)
+    { kind: "refuse", reason } => failure(reason)
```
egonSchiele, line 889: **This arm is unreachable, and the comment above stops just short of saying so.**

The early return at the top of `safeBash` catches `refuse` before `raiseAll`, so control can never arrive here. The comment says the arm exists "because the match is exhaustive over Execution", which explains why it's present but reads as though it might fire.

Worth one more word — "unreachable; present only to keep the match exhaustive" — because the next person to touch this will otherwise wonder which path reaches it, and might try to make the two branches agree about something.

packages/agency-lang/tests/agency/safeBash.agency
```diff
@@ -136,18 +144,18 @@ node shellFreeEcho() {
 
 def writePlanDetail(source: string): any {
   const plan = planFor(source, "/repo")
-  const exec = plan.execution
+  const names = [e.name for e in plan.effects]
```
egonSchiele, line 147: **The fixtures now describe a record that no longer exists.**

The comment is honest about it: the non-write arm reports "the same sentinel fields the old flat Execution record carried" so the rows stay byte-identical. For *this* PR that's exactly right — byte-identical fixtures are the proof the refactor changed no behavior, and that proof is worth fabricating the old shape for.

But once this merges, the fixture rows for bash plans encode `["bash", "", "", "", "overwrite", [...]]`, describing a filename, dir and mode that `BashExec` does not have. The next person reading those rows learns a shape the code contradicts.

Suggest a follow-up (not this PR) to collapse those rows to what a `BashExec` actually carries. The refactor's value is that the type stops lying about which fields are meaningful; the tests should stop too, just not in the commit that proves nothing changed.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -275,35 +274,32 @@ export def effectsFor(command: SimpleCommand, cwd: string): Result<Effect[]> {
...
-    return success(redirect.value)
+  // `echo` contributes nothing on its own, whatever its arguments: bash
+  // runs it and expands anything in it, so `echo $HOME` and `echo -n hi`
+  // are both fine on this path. The flag and literal-argument restrictions
+  // belong to the shell-free path, which computes the output itself. Only
+  // the git arm reads the arguments as text — a quoted `echo "hi"` must
+  // not be demoted by a check that only git needs.
+  return match(name.text) {
+    "echo" => success(redirect.value)
+    "git" => literalArgs(command)
```
egonSchiele: **Continuation lines sit at arm indentation.**

```
    "echo" => success(redirect.value)
    "git" => literalArgs(command)
    |> gitEffects.partial(flags: flags, cwd: cwd)
    |> withRedirect.partial(redirect: redirect.value)
    _ => failure("no rule for `${name.text}`")
```

The two `|>` lines are at the same indent as `"echo"`, `"git"` and `_`, so at a glance the match looks like it has five arms. Skimming for "what are the cases here" gets the wrong answer.

If this is `fmt` output rather than hand-written, it's worth a formatter issue rather than a change here — a piped continuation inside a match arm is a shape it probably hasn't seen before.

egonSchiele (commented):
The `Execution` union is a clear win and the exhaustive dispatch is the right shape — a fifth kind being a type error at the dispatch is exactly the property that flat record with sentinel fields could never give you. `emptyExecution()` with its `kind: ""` and `mode: "overwrite"` placeholders was the ugliest thing I left behind, and it's good to see it go.

I verified the parse finding in TODO.md before reviewing the code that works around it, and it reproduces exactly: a single-element array pattern arm parses fine in first position, but after an expression arm it fails with `expected match cases of the form value => expression`. The workaround in `gitEffects` is justified.

My findings cluster in one place: three of the pipe conversions invented always-succeeding `Result` functions to have something to pipe into. That's the pipe operator pulling the code toward it rather than fitting it. The rest are smaller.

egonSchiele:
All six addressed in cd2ec5b. What changed, by finding:

**Always-succeeding pipe targets** — you were right, and the fix deleted all three. `withRedirect`, `echoPlanFromWords` and `writePlanFromWords` are gone. The one pipe that survives is `literalArgs(command) |> gitEffects.partial(flags:, cwd:)` inside a block arm, because both links genuinely fail; the redirect append and the plan construction are back inline with plain unwraps. The rule that fell out — pipe only when every link can fail — is now a comment at that call site.

**Write-arm rebuild** — converted to the same if-then-exhaustive-match hybrid as `flattenOne`: `if (exec.kind == "write") { return runWrite(exec) }` passes the value through whole, and the match stays exhaustive over what remains (verified: no AG5002). Your framing — that this was the one workaround that could *lose data* rather than just being verbose — was the sharpest observation in the review; the comment there now states that failure mode.

**Mixed arm forms in `gitEffects`** — the `return`s are dropped; all five arms are now values. For the record, those `return`s were `fmt` output, not hand-written: it rewrites `=> { return success(...) }` block arms into `=> return success(...)`. Noted in #708.

**Unreachable refuse arm** — now says "Unreachable — the early return above already handled refusal. The arm exists only to keep the match exhaustive over Execution."

**Fixture follow-up** — agreed on both halves: byte-identical rows are the proof for this PR, and they describe a dead shape after it merges. Added to TODO.md so it does not evaporate.

**Pipe-continuation indentation** — filed as #708, with the arm-return rewrite noted there too since it is probably the same fix site. The shape no longer occurs in this PR (the chain moved into a block arm), so the issue carries its own repro.

18/18 agency tests (fixtures still byte-identical) and 24/24 property tests after the changes.

egonSchiele:
CI failure fixed in 942370aee: the fixture typecheck test caught a third checker gap — flow narrowing does not reach inside a match arm block, so the `is failure` guard in the git arm did not narrow `git` and `.value` failed AG2008 (the identical shape typechecks at function scope, which is why local `make` was clean). Fixed by matching on result patterns (`success(effects) => ...`), which bind the value and need no narrowing — arguably the more idiomatic form anyway. Recorded in TODO.md alongside the other two gaps. Fixture typecheck, 18/18 agency tests, and 24/24 property tests all green locally.

PR #710, opened 2026-07-28
parser: an index or slice cannot start on a new line

egonSchiele:
Fixes the first of the three compiler gaps found during the safeBash match refactor (#707's TODO entries): a single-element array pattern arm after an expression arm parsed as an **index on that arm's value**, killing the whole match with the generic "expected match cases" error.

## Mechanism (from the DEBUG trace)

`_functionCallParser` consumes trailing whitespace **including newlines**. After the arm expression `failure("a")` parsed, the chain loop (`many(chainElementParser)`) resumed at the next line's `[sub]`, `bracketParser` matched, and the second arm's pattern became `failure("a")[sub]`. Multi-element patterns were immune only because `expr["a", "b"]` is not valid index syntax — which is why v2 safeBash's `["git", "status"]` arms worked by accident while v3's `["status"]` arms could not.

## The fix

A boundary at the top of `indexChainParser` and `sliceChainParser`: look back from the element's position across spaces and tabs (via tarsec's parse-wide input, `getInputStr`), and refuse the element if a newline sits there. Statement context already refused this continuation (`const x = g()` newline `[0]` was two statements before this change), so the boundary aligns expression context with existing statement behavior rather than inventing a new rule.

Deliberately untouched: **dot-method chains** — a fluent `.method()` on its own line is an established style. A bare `[i]` on a new line is an array literal or a pattern, never an intended index. If the parse-wide input is unavailable (defensive case), the boundary disables itself and behavior is exactly as before.

## Verified

- New `lib/parsers/chainNewlineBoundary.test.ts`: the four failing arm shapes now parse (literal, binder, guarded, and after a multiline arm expression); same-line chains (`arr[0][1]`, `arr[:2]`, multiline-call-then-`[0]`) unchanged; next-line bracket provably does not glue onto the previous expression. Tests run through `parseAgency` — raw-parser tests don't install the parse-wide input (vitest resolves two tarsec instances), which is itself worth knowing.
- Full parser suite: 1753/1753. Full corpus build clean (only the known deliberate diagnostics). safeBash agency tests 18/18, fixture typecheck + walker tripwire green, structural lint clean.

Removes the fixed entry from TODO.md. The two checker gaps (arm narrowing, flow into arm blocks) remain and are next; spec incoming.

### DIFF
--- a/packages/agency-lang/lib/parsers/parsers.ts
+++ b/packages/agency-lang/lib/parsers/parsers.ts
@@ -2609,8 +2609,32 @@ const bracketParser: Parser<boolean> = or(
   map(char("["), () => false),
 );
 
-/** True when the current parse position sits just after a line break,
- *  looking back across spaces and tabs only.
+/** Offsets that begin a line's content: marks[i] is 1 when source[i] is
+ *  the first character on its line that is not a space or tab. Built in
+ *  ONE pass per parse input and cached by string identity, so the
+ *  boundary check below never re-scans text the parser already
+ *  consumed. */
+let markedSource: string | null = null;
+let lineStartMarks: Uint8Array | null = null;
+
+function lineContentStarts(source: string): Uint8Array {
+  const marks = new Uint8Array(source.length);
+  let atLineStart = true;
+  for (let i = 0; i < source.length; i++) {
+    const c = source.charCodeAt(i);
+    if (c === 10 || c === 13) {
+      atLineStart = true;
+    } else if (atLineStart && c !== 32 && c !== 9) {
+      marks[i] = 1;
+      atLineStart = false;
+    }
+  }
+  return marks;
+}
+
+/** True when the chain element about to be parsed sits at the start of
+ *  its own line (looking past indentation) — meaning the expression it
+ *  would extend ended on the PREVIOUS line.
  *
  *  A bracket chain element (`[i]`, `[a:b]`) must not continue an
  *  expression across a newline. Expression parsers (function calls in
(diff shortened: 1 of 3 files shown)


egonSchiele:
Rewritten in bb7b97982 per the performance feedback. Three changes: (1) the common path is now a single `charCodeAt` — the guard does nothing at all unless the next character is `[` or `?.[`, so the per-value-access cost is one integer compare; (2) when a bracket is present, the answer comes from a precomputed line-start marks table (one `Uint8Array` pass per parse input, cached by string identity) — no `getInputStr`-plus-backscan in the hot path, one array read instead; (3) slice and index share the one guard rather than scanning twice. Same predicate, plus lone \r line endings now count as breaks. Full parser suite 1753/1753, corpus build, safeBash 18/18, fixture typecheck all green.

egonSchiele:
Rewritten in combinator style in 5fef404c0. The boundary is now a zero-width predicate parser, `startOfLineContent`, used as `not(startOfLineContent)` as the first element of the two chain parsers' `seqC` — the same shape the comprehension parser uses with `not(varNameChar)` for word boundaries. The marks table, the Uint8Array, the caching, and the manual loops are all gone; inside the predicate it is one `lastIndexOf` to the line start and one regex test over the indentation slice, both native string ops. One behavior worth knowing: when the parse-wide input is not installed (raw parser tests calling parsers directly), the predicate fails, which disables the boundary instead of misfiring — that is also why the boundary's own tests go through `parseAgency`. Full parser suite 1753/1753, corpus build, safeBash 18/18, fixture typecheck green.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2608,8 +2609,48 @@ const bracketParser: Parser<boolean> = or(
   map(char("["), () => false),
 );
 
+/** Zero-width parser: succeeds, consuming nothing, when the parse
```
egonSchiele, line 2612: Because this function uses getInputStr, it does not have very good performance. Please add a note, making it very clear that users absolutely should not use it unless there is no other option.

egonSchiele: Added in ba5491d: a PERFORMANCE WARNING block at the top of the docstring saying not to reach for this parser unless there is no other option — it calls getInputStr and re-reads source text on every attempt, which ordinary combinators never do — and that the right default is restructuring the grammar to test before the whitespace is consumed (lookahead, reordering, or a dedicated trivia parser). The chain boundary qualifies only because the newline is already gone by the time the chain parsers run.

Issue #708, opened 2026-07-28
fmt: pipe-chain continuations in match arms sit at arm indentation


egonSchiele:
Found during the safeBash match refactor (PR #707). When a match arm's expression is a pipe chain, `fmt` puts the `|>` continuation lines at the same indentation as the arm patterns:

```
  return match(name.text) {
    "echo" => success(redirect.value)
    "git" => literalArgs(command)
    |> gitEffects.partial(flags: flags, cwd: cwd)
    |> withRedirect.partial(redirect: redirect.value)
    _ => failure("no rule for `${name.text}`")
  }
```

Skimming for "what are the cases here" reads five arms where there are three. The continuation lines should be indented past the arm head, something like:

```
    "git" => literalArgs(command)
      |> gitEffects.partial(flags: flags, cwd: cwd)
      |> withRedirect.partial(redirect: redirect.value)
```

The PR no longer contains this shape (the chain moved into a block arm for other reasons), so there is no live repro in the tree — the snippet above reproduces it through `pnpm run fmt`.

Two related arm-formatting notes from the same session, possibly the same fix site:
- `fmt` rewrites a block arm `=> { return success(...) }` into `=> return success(...)`, which mixes value-arms and return-arms in one match (see the gitEffects history in #707).
- A single-element array pattern arm after an expression arm parses as an INDEX on that expression (`failure("a")` then `[sub] =>` parses as `failure("a")[sub]`). Recorded in TODO.md; the formatter cannot cause it but a formatter that joined lines could turn working code into this shape.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

egonSchiele:
A third instance from PR #709, and this one produces output the parser rejects outright: `fmt` collapses a single-statement block arm into `=> <statement>` for ANY statement kind, but the arm body parser only accepts `return` / `goto` / assignment / expression. So

```
"std::bash" => {
  raise std::bash("...", { ... })
}
```

formats to

```
"std::bash" => raise std::bash("...", { ... })
```

which fails to parse with the generic "expected match cases" error — `fmt` breaks the file it formats. Workaround used in #709: keep arm blocks at two statements so the collapse never triggers.

Fix is one of: the generator only collapses blocks whose lone statement is in the parseable set, or the arm parser learns the remaining statement kinds (`raise` at minimum — raising from an arm is a natural shape for effect dispatch).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #712, opened 2026-07-28
fmt: only collapse arm blocks the single-statement grammar can re-parse

egonSchiele:
Fixes the file-breaking half of #708: `fmt` collapsed any one-statement block arm into `=> <statement>`, but the arm grammar (`matchBlockParserCase`) only accepts `return`, `goto`, assignment, and expressions — so a block holding a raise statement formatted into `=> raise ...`, which its own parser rejects with the generic "expected match cases" error. Found while writing #709, which dodged it with two-statement blocks.

The collapse test is now an **allowlist mirroring the arm grammar exactly**: `returnStatement`, `gotoStatement`, `assignment`, or `isExpressionNode` (the co-located source of truth for the `Expression` union) — with two carve-outs: nested `matchBlock` keeps its block (pre-existing rule), and the raise-form interrupt is excluded even though the interrupt *expression* shares its `interruptStatement` node type (`viaRaise` tells them apart, and the expression form still inlines).

Being an allowlist, an unknown statement kind degrades to block form — more verbose, never unparseable — which is the right failure direction for a formatter. No existing formatted file changes: any committed file containing the affected shapes wouldn't have parsed in the first place.

Tests: three new round-trip cases (raise-statement arm keeps its block, if-statement arm keeps its block, interrupt-expression arm stays inline). One incidental discovery pinned in the raise case: the generator canonicalizes `raise interrupt(x)` to `raise(x)` — both parse to the same AST, so it is normalization, not loss.

Not addressed here (still open on #708): the pipe-continuation indentation, and whether the arm parser should learn `raise` as a first-class arm body.

Backend suite 37 files / 496 tests green; formatting the formatted stdlib produces no churn.

### DIFF
--- a/packages/agency-lang/lib/backends/agencyGenerator.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.ts
@@ -143,6 +143,19 @@ function escapeMultiLineText(s: string): string {
   return s.split("${").join("\\${");
 }
 
+/** Statement types the single-statement arm grammar accepts in addition
+ *  to expressions — matchBlockParserCase's inline alternation. */
+const INLINE_ARM_STATEMENT_TYPES: readonly string[] = [
+  "returnStatement",
+  "gotoStatement",
+  "assignment",
+];
+
+/** Node types that never print inline even though they are expressions:
+ *  a nested match keeps block form so it re-parses as a match, not an
+ *  arm body. */
+const NEVER_INLINE_ARM_TYPES: readonly string[] = ["matchBlock"];
+
 export class AgencyGenerator {
   protected graphNodes: GraphNodeDefinition[] = [];
   protected generatedStatements: string[] = [];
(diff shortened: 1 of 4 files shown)


egonSchiele:
Redesigned per your comment in 456d0abed. You were right that preserving the author's form is the simpler rule — the obstacle was that the AST could not express it: `=> { print("hi") }` and `=> print("hi")` parsed to identical nodes, which is why the formatter had to pick a canonical form at all. The parser now records `blockBody` on arms written as blocks (set-only-when-true, same style as the chain elements' `optional` flag — which is also why zero existing test expectations changed), and the formatter's primary rule is preservation: blocks stay blocks, inline stays inline. The allowlist from the previous commit is demoted to a fallback that only governs programmatically-built cases, which carry no recorded form — there, inline is only safe for shapes the arm grammar accepts, so anything else takes block form. One new round-trip test pins block preservation. 2250 backend+parser tests green, corpus build clean, canonical files round-trip with no churn.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -1438,15 +1439,25 @@ export class AgencyGenerator {
...
+      // prints inline. The shape test below only decides for cases with NO
+      // recorded form — ASTs built programmatically — where inline is only
+      // safe for the shapes the single-statement arm grammar accepts;
+      // anything else (a raise statement, an if, a loop) must take block
+      // form or the printed arm will not re-parse (#708).
+      const only = caseNode.body.length === 1 ? caseNode.body[0] : null;
+      const inlinable =
+        only !== null &&
+        caseNode.blockBody !== true &&
+        only.type !== "matchBlock" &&
```
egonSchiele: This is pretty unreadable code. Please create an array of types that are not allowed and an array of types that are allowed, and use it here. The ternary at the top is confusing as well. Why not split this up into a couple of if statements? 

```
let inlinable;
if (caseNode.body.length !== 1) {
inlinable = false;
} else {
...
}
```

egonSchiele: Refactored in f9c42c4a9: the decision is now a named method, armPrintsInline, written as sequential if statements — length check, author's blockBody, NEVER_INLINE_ARM_TYPES array, the viaRaise carve-out (kept as an explicit if since it needs a flag, not just a type), then INLINE_ARM_STATEMENT_TYPES array with isExpressionNode as the final test. The two arrays carry the doc comments explaining what each set means. Behavior unchanged; all 12 round-trip tests pass without modification.

Issue #713, opened 2026-07-28
Top-level if/while/for/match/thread crash the builder instead of erroring cleanly


egonSchiele:
Several statements that are not meaningful at the top level of a file parse successfully and then crash the TypeScript builder with an internal error, instead of being rejected with a normal diagnostic.

## Repro

```ts
// crash.agency
if (true) { print(1) }

node main() { print(2) }
```

```
$ agency compile crash.agency
Error: StepPathTracker: currentId() called with empty path
    at StepPathTracker.currentId (lib/backends/typescriptBuilder/stepPathTracker.js:39)
    at TypeScriptBuilder.processIfElseWithSteps (lib/backends/typescriptBuilder.js:1232)
    at TypeScriptBuilder.processNode (lib/backends/typescriptBuilder.js:530)
    at TypeScriptBuilder.processNodeInGlobalInit (lib/backends/typescriptBuilder.js:3191)
    at partitionProgram (lib/backends/typescriptBuilder/sectionAssembler.js:127)
    at TypeScriptBuilder.build (lib/backends/typescriptBuilder.js:389)
```

The parser accepts the `if`, so it reaches the builder as a top-level node. `partitionProgram` routes it into global-init, and the step machinery there has no step path to hand out, so `currentId()` throws. The user gets a stack trace with no file, line, or column.

## What else does this

I swept the statement forms at the top level of a file. Each was written above an ordinary `node main()`:

| Top-level statement | What happens |
| --- | --- |
| `if (true) { print(1) }` | **crash** (StepPathTracker) |
| `while (false) { print(1) }` | **crash** (StepPathTracker) |
| `for (x in [1,2]) { print(x) }` | **crash** (StepPathTracker) |
| `match(1) { 1 => print(1) }` | **crash** (StepPathTracker) |
| `thread { print(1) }` | **crash** (StepPathTracker) |
| `guard { print(1) }` | clean parse error |
| `try { print(1) } catch { print(2) }` | clean parse error |
| `handle read(p) { return "x" }` | clean parse error |
| `return 1` | compiles |
| `goto other` | compiles |
| `debugger` | compiles |
| `raise interrupt read(p: "x")` | compiles |
| `print(1)` | compiles (intended — `static` expression statements are a supported top-level form) |
| `x = 1` | compiles |

So there are three groups: five forms that crash, three that are rejected cleanly, and a set that compiles.

## Asks for whoever picks this up

[...]

PR #714, opened 2026-07-28
fix: node/def declarations no longer parse as a name plus a call, and toSource stops mutating its input

egonSchiele:
Two independent bugs found while working through the Template Agency guide. Spec and plan:

- `docs/superpowers/specs/2026-07-28-code-literal-parse-and-print-fixes-design.md`
- `docs/superpowers/plans/2026-07-28-code-literal-parse-and-print-fixes.md`

## Bug 1: a declaration parsed as a name followed by a call

This compiled, and died at run time with `ReferenceError: node is not defined`:

```ts
node main() {
  node inner() { print(1) }
  print("x")
}
```

A call may take a trailing block, and keywords are not reserved in expression position (`const node = 1` compiles today), so the body parser read `node` as a name and `main() { ... }` as a call with a block. A return type annotation makes the call reading fail, which is why `node main(): string { ... }` always worked and nobody noticed.

Inside a code literal it was worse. `parseCodeLiteralBody` tries expression, then statements, then program, taking the first that consumes the body — so the wrong statements reading won and the program parser was never consulted. `[| node main() { ... } |]` ended up holding a name and a call. Nothing downstream caught it: the tree printed and re-parsed stably, `fill` and `holesOf` worked fine, and only the generated program failed. The reported symptom was `agency fmt` printing `node` and `main() {` on separate lines — the formatter printing exactly what it was handed.

**Fix:** decline the declaration shape in body position (`bodyDeclarationParser`). Ordinary bodies now get a real error, and code literals fall through to the program parser and get the right tree. No routing, no peeking at literal bodies.

Three details are load bearing and are commented at the code:

- It returns `committedFailure`, not plain `failure`. `or` treats an ordinary failure as "try the next alternative", and the alternatives after it are the ones that read `node` as a name — a plain failure would have made the whole change a silent no-op.
- It registers that failure in `getParseState().committedFailure`, the slot `committed()` writes and both of `parseAgency`'s failure exits prefer. Without it the message loses to the enclosing `parseError` throw and the user gets "expected node body".
- It never uses `parseError`, which throws — see Bug 2.

The probe mirrors the real declaration prefix (keyword, whitespace, name, `(`) and reuses `varNameChar`. Requiring the `(` is what keeps ordinary uses of `node` as a variable out of it.

## Bug 2: `toSource` threw on a `static const` literal

```ts
static const s = [| print("hi") |]
node main() { print(toSource(s)) }
```

```
TypeError: Cannot assign to read only property 'nodes' of object
```

`AgencyGenerator.generate` assigned its partitioned node list back over the caller's `nodes` field. Statics are `__deepFreeze`d by design, so the write threw. Masked in the guide because every example calls `fill` first, and `fill` works on a clone. Fixed by keeping the partitioned list in a local.

Also fixed in the same area: `[| static const x = 1 |]` failed outright with a message about function bodies. The body-context diagnostic for `static const` reports through tarsec's `parseError`, which throws, and the throw escaped before the program attempt ran. `tryAttempt` now converts a thrown `TarsecError` into an ordinary attempt failure so kind inference continues.

## What changes for users

| Case | Before | After |
| --- | --- | --- |
| `[\| node main() {} \|]` | wrong tree, fails at run time | correct declaration |
| `[\| def foo() {} \|]` | wrong tree | correct declaration |
| statement, then a declaration | wrong tree | correct program |
| comment, then a declaration | wrong tree | correct program |
| `node inner() {}` in a node body | compiled, crashed at run time | parse error naming the rule |
| `[\| static const x = 1 \|]` | hard parse failure | correct program |
| `toSource(staticLiteral)` | TypeError | prints |

Deliberately unchanged, and pinned by tests: `[| node |]`, `[| node.run() |]`, `[| node + 1 |]` stay `expr`; `[| type P = ... |]` and `[| effect Foo |]` stay `statements` (both are legal in bodies, so that reading is correct); `debugger` still parses as a bare name.

One accepted regression, corrected from an earlier version of this description. A body with `node` on one line and a call on the next **does** compile on main:

```ts
def helper(): number { return 1 }

node main() {
  const node = 1
  node
  helper()
}
```

On this branch that is a hard parse error, because the probe's `many1(space)` matches a newline — the same whitespace the real declaration grammar accepts — so keyword-newline-name-paren is declined exactly like the one-line form. The regression is worth accepting: that text is indistinguishable from the bug being fixed, and anyone writing it almost certainly meant a declaration. It is now pinned by a known-limitation test so a future probe change that un-declines it is a deliberate decision.

## Testing

- Unit: 9741 passed, 0 failed.
- `tests/agency/templates`: 14 passed, including two new fixtures — one fills, prints, compiles and runs a generated un-annotated node declaration end to end; the other calls `toSource` on a `static const` literal with no `fill` in between.
- The whole-corpus formatter round-trip gate passes, plus a new byte-exact golden for an un-annotated node in a literal.

The build regenerated `docs/site/examples/generatedAgent.md`, which shows the reported symptom disappearing:

```diff
   const template = [|
-    node
-    main() {
+    node main() {
```

## Pre-existing issue noticed, not fixed here

Committed-failure messages report a line number shifted by the prelude wrapper: the new message points at line 4 for a declaration on line 2. The existing unclosed-literal error is shifted the same way, so this is not introduced here — the position for committed failures is not adjusted by the template offset. Worth its own fix, since it affects every committed-failure diagnostic.

Related: #713 (top-level `if`/`while`/`for`/`match`/`thread` crash the builder), found while establishing which statements are legal at file scope.

### DIFF
--- a/packages/agency-lang/lib/parsers/codeLiteral.test.ts
+++ b/packages/agency-lang/lib/parsers/codeLiteral.test.ts
@@ -306,3 +306,77 @@ describe("code literals: review-round pins", () => {
     expect(errorData!.line).toBeLessThanOrEqual(3);
   });
 });
+
+describe("declarations in a literal infer program", () => {
+  // `expects` lists node types the literal must hold. Order is not
+  // asserted — a leading comment is kept as a node of its own, so the
+  // declaration is not always first.
+  const cases: { body: string; expects: string[]; label: string }[] = [
+    { label: "un-annotated node", body: "node main() {\n  print(1)\n}", expects: ["graphNode"] },
+    { label: "un-annotated def", body: "def foo() {\n  return 1\n}", expects: ["function"] },
+    { label: "node with parameters", body: "node m(x: number) {\n  print(x)\n}", expects: ["graphNode"] },
+    { label: "two declarations", body: "node a() {\n  print(1)\n}\n\nnode b() {\n  print(2)\n}", expects: ["graphNode"] },
+    { label: "comment first", body: "// hi\nnode main() {\n  print(1)\n}", expects: ["comment", "graphNode"] },
+    { label: "statement then declaration", body: "print(1)\nnode main() {\n  print(2)\n}", expects: ["functionCall", "graphNode"] },
+  ];
+
+  for (const { label, body, expects } of cases) {
+    it(`infers program for a ${label}`, () => {
+      const lit = firstLiteral(`node host() {\n  const t = [|\n${body}\n  |]\n}\n`);
+      expect(lit.kind, label).toBe("program");
+      // The kind alone is not enough: it was already wrong for a reason
+      // only the node types reveal. Under the bug the declaration became a
+      // `variableName` plus a `functionCall`, so both halves of this
+      // matter — the declaration is present, and no stray name is.
+      const types = lit.nodes.map((node) => node.type);
+      for (const expected of expects) {
+        expect(types, `${label}: ${expected}`).toContain(expected);
+      }
+      expect(types, label).not.toContain("variableName");
+    });
+  }
+
+  it("still infers program for an annotated node", () => {
+    const lit = firstLiteral(`node host() {\n  const t = [|\n    node main(): string {\n      return "x"\n    }\n  |]\n}\n`);
+    expect(lit.kind).toBe("program");
+    expect(lit.nodes[0].type).toBe("graphNode");
+  });
+
+  it("still infers program for an annotated def", () => {
+    // `def` takes the other branch of the probe's `or`, so it needs its
+    // own annotated case, not just the un-annotated one above.
+    const lit = firstLiteral(`node host() {\n  const t = [|\n    def foo(): number {\n      return 1\n    }\n  |]\n}\n`);
+    expect(lit.kind).toBe("program");
+    expect(lit.nodes[0].type).toBe("function");
+  });
+
+  it("reports a broken declaration instead of silently reading two statements", () => {
+    // Before the fix this parsed as statements and produced junk. Now the
+    // statements attempt declines and the parse fails.
+    const source = `node host() {\n  const t = [|\n    node main( {\n      print(1)\n    }\n  |]\n}\n`;
+    const result = parseAgency(source, {}, true, false);
+    expect(result.success).toBe(false);
+  });
+});
+
+describe("bodies whose kind must not change", () => {
+  // Regression guards. Each begins with a word that a keyword-routing fix
+  // would have misread, and `type` / `effect` are legal in bodies so their
+  // statements answer is the correct one.
+  const cases: { body: string; kind: string }[] = [
+    { body: "node", kind: "expr" },
+    { body: "node.run()", kind: "expr" },
+    { body: "node + 1", kind: "expr" },
+    { body: "const x = 1", kind: "statements" },
+    { body: "print(1)", kind: "expr" },
+    { body: "type P = { n: number }", kind: "statements" },
+    { body: "effect Foo", kind: "statements" },
+  ];
+
+  for (const { body, kind } of cases) {
+    it(`keeps \`${body}\` as ${kind}`, () => {
+      const lit = firstLiteral(`node host() {\n  const node = 1\n  const t = [| ${body} |]\n  print("x")\n}\n`);
+      expect(lit.kind, body).toBe(kind);
+    });
+  }
+});
--- a/packages/agency-lang/lib/parsers/codeLiteral.test.ts
+++ b/packages/agency-lang/lib/parsers/codeLiteral.test.ts
@@ -352,10 +352,17 @@ describe("declarations in a literal infer program", () => {
 
   it("reports a broken declaration instead of silently reading two statements", () => {
     // Before the fix this parsed as statements and produced junk. Now the
-    // statements attempt declines and the parse fails.
+    // statements attempt declines, the program attempt runs, and its own
+    // declaration parser reports the real problem — the unclosed parameter
+    // list. Asserting the message matters as much as asserting the failure:
+    // a future change that keeps the failure but degrades the message to a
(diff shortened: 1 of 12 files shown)


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3094,6 +3095,37 @@ function advancePositionOver(base: Position, text: string): Position {
  *  `base` is passed in USER coordinates (template offset already
  *  subtracted), and this file's own template offset is zeroed around the
  *  nested parses so withLoc does not subtract it a second time. */
+/**
+ * Run one kind attempt, treating a throw as a non-match.
```
egonSchiele, line 3087: `tryAttempt` was inserted between `parseCodeLiteralBody`'s doc comment and the function it documents: the block ending "…so withLoc does not subtract it a second time." (the smallest-first kind-inference comment) now sits directly above `tryAttempt`, stacked on top of this comment. Readers and doc tooling will attach it to the wrong function. Move `tryAttempt` and this comment above that block so the kind-inference comment touches `parseCodeLiteralBody` again.

packages/agency-lang/lib/parsers/codeLiteral.test.ts
```diff
@@ -306,3 +306,94 @@ describe("code literals: review-round pins", () => {
...
+    expect(lit.kind).toBe("program");
+    expect(lit.nodes[0].type).toBe("function");
+  });
+
+  it("reports a broken declaration instead of silently reading two statements", () => {
+    // Before the fix this parsed as statements and produced junk. Now the
+    // statements attempt declines and the parse fails.
+    const source = `node host() {\n  const t = [|\n    node main( {\n      print(1)\n    }\n  |]\n}\n`;
+    const result = parseAgency(source, {}, true, false);
+    expect(result.success).toBe(false);
```
egonSchiele, line 363: Asserting only `success === false` does pin the fix (before it, this source parsed successfully as junk statements), but the spec's claim for this case was error *quality*: "a parse error about a node is the honest message." Nothing holds that. Consider also asserting the message mentions the declaration (or at least does not mention two unrelated statements), so a future change that keeps the failure but degrades the message to a generic one is caught.

egonSchiele (commented):
Reviewed the diff and verified the load-bearing mechanics against the code rather than the description: tarsec's `or` really does fall through on a plain failure and stop on a committed one, `not`/`peek` restore the `committedFailure` slot around speculation (so the new slot write cannot go stale through lookahead), both of `parseAgency`'s failure exits read that slot (`lib/parser.ts:203,347`), and the generator repointing is complete — `preprocessAST`, which runs between the repointed passes, is a no-op with no overrides anywhere. The test suite is strong: the no-mutation test compares the whole tree, the hoisting test asserts order with the import deliberately below the declaration, and the word-operator pins (`node is string` et al.) guard exactly the cases the probe's `(` requirement exists for.

One correction to the PR description, found by running the case on both branches:

**The "accepted regression" paragraph is wrong about today's behavior.** It says a body with `node` on one line and `main()` on the next "does not parse today either." It does parse on main — this file compiles to an AST today:

```
node main() {
  const node = 1
  node
  main()
}
```

On this branch the same file is a hard parse error (`Line 6, col 8: node and def declarations are only legal at the top level of a file`), because `many1(space)` in the probe matches the newline, so keyword-newline-name-paren is declined like a one-line declaration. The regression itself seems right to accept — that text is indistinguishable from the bug being fixed, and anyone writing it almost certainly meant a declaration — but the description should say "this parses today and becomes a hard error" rather than claiming it never parsed. Worth pinning with a known-limitation test too, so a future probe change that un-declines it is a conscious decision. (Side note from the same repro: the reported line 6 for a declaration on line 4 is the pre-existing committed-failure offset shift the description already flags.)

Two small things noted inline, plus one not attached to a diff line: `this.program = program` at `agencyGenerator.ts:201` is still written and never read — the plan called removing it optional cleanup, so fine either way, but it is now the only survivor of the mutation this PR removes.

egonSchiele:
All three taken, pushed in ac6575574. I re-ran the newline case on both branches rather than taking the correction on trust — you are right and I was wrong.

**The description was wrong about `node` newline `main()`.** It compiles on main; the branch rejects it:

```
main:   investigate/nl2.agency → investigate/nl2.js (in 177.90ms)
branch: Failed to parse ...: Line 10, col 10: `node` and `def` declarations are only legal at the top level of a file.
```

The mistake traces back to how I checked it during planning: I used `main()` as the callee inside `node main()`, which trips a separate builder rule — "Call to graph node 'main' inside graph node 'main' must use goto or return" — and I read that failure as a parse failure. It is not; the file parses fine. With a non-recursive callee it compiles cleanly. Description corrected to say it parses today and becomes a hard error, and there is now a known-limitation test pinning it with the reasoning inline.

**The orphaned doc comment** is real — `tryAttempt` landed between `parseCodeLiteralBody`'s doc block and the function. Moved `tryAttempt` and its own comment above that block, so the kind-inference comment touches its function again.

**The broken-declaration test** now asserts the message. Worth noting what it actually says, because it is better than the spec predicted: `expected \`,\` between parameters or \`)\` to close the parameter list`. That comes from the program attempt's own declaration parser, so the error is about the malformed declaration rather than about the word "node" — which is the substance the spec asked for. My first attempt at this assertion used `/node|declaration/i` and failed against that message; asserting `"parameter list"` pins the real thing.

**`this.program`** removed, along with its field declaration. Confirmed dead first: no reads anywhere in the file, and no subclasses of `AgencyGenerator` exist. The `this.program` hits elsewhere in the repo are `TypescriptPreprocessor`, a different class.

Re-verified after the changes: 9741 unit tests pass, 14/14 template fixtures, `lint:structure` clean, `make` exits 0. The AG6019/AG3013 diagnostics `make` prints for `dist/lib/agents/**` are identical on main — the known deliberate corpus set, not introduced here.

PR #715, opened 2026-07-28
typechecker: match arms narrow (matchYield flow, ===, bare-variable scrutinee)

egonSchiele:
## What

Match arms now narrow. Three point fixes plus deletion of the three
safeBash workarounds they forced:

1. **matchYield terminates flow, bounded to its match region**
   (flowBuilder). An early return inside an expression-position arm is
   a matchYield after lowering; it now ends its branch like a return,
   with two carve-outs mirroring existing contracts: an
   interrupt-carrying yield passes through (the may-resume convention),
   and the matchExprId-tagged chain root returns the pre-match flow so
   statements after a match keep their narrowing.
2. **=== and !== narrow like == and !=** (analyzeCondition). Strict
   equality is a stylistic alias compiling to the same __eq; it now
   produces identical facts on both the null presence path and the
   literal discriminant path. Pattern lowering only emits loose
   operators, so this is purely for hand-written strict equality.
3. **Bare-variable scrutinees narrow in arm bodies** (patternLowering).
   Lowered conditions and binder reads now target the original variable
   instead of the __scrutinee temp when the scrutinee is a bare
   variable, so the arm body references the narrowed name. Conservative
   fallback: an arm body that redeclares the scrutinee name keeps the
   temp for the whole match (generated locals are frame slots; a
   substituted read could target the redeclared slot) — pinned by a
   typecheck row and a runtime execution test. Member-path scrutinees
   unchanged (pinned as a documented limitation).

A bonus that fell out of fix 3: the handlerParamTyping suite pinned
"match(e) object-pattern arms do not narrow e.data" as a known
limitation — that now works, and the pin is flipped to assert it.

## Why

Spec: docs/superpowers/specs/2026-07-28-match-arm-narrowing.md (v3),
with two spec review rounds and a plan review in the sibling -REVIEW
files. The gaps were found during the safeBash match refactor; each
converted safeBash site carried a comment naming the gap that forced
its shape.

## Not a bug after all

Binder typing (an object-pattern binder taking the selected variant's
field type) was reported broken during the refactor but measures
correct on main; this PR pins it with tests and retires the stale TODO
bullet rather than "fixing" it.

## Testing

- New unit rows: strict-equality narrowing (discriminant + null paths),
  matchYield flow termination on both the surviving-matchBlock and
  lowered-chain paths, post-match flow pins incl. the Fix1-x-Fix3
  interaction, an alwaysExits misfire detector, interrupt-yield
  symmetry, bare-variable scrutinee narrowing incl. reassignment
  invalidation and the shadow fallback, binder typing pins, member-path
  limitation pin.
- New agency execution test: match-scrutinee-shadow (runtime dispatch +
  binding with a redeclared scrutinee name).
- stdlib/safeBash.agency typechecks with all three workarounds deleted;
  the 18 safeBash execution tests pass.
- Fixtures regenerated; the fix 3 fixture diff is a pure reference swap
  (type-patterns.mjs: temp reads become args reads, nothing else).
- Full local unit suite: 9703 passed. Full agency suite: CI.

### DIFF
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -50,6 +50,7 @@ import type {
 import type { ValueAccess, AccessChainElement } from "../types/access.js";
 import type { SourceLocation } from "../types/base.js";
 import { mapBodies } from "../utils/mapBodies.js";
+import { walkNodes } from "../utils/node.js";
 import { OBJECT_REST_FN } from "../constants.js";
 
 // ---------------------------------------------------------------------------
@@ -476,7 +477,7 @@ class PatternLowerer {
       // checker can find the owning match for exhaustiveness / union typing.
       ...(matchExprId !== undefined ? { matchExprId } : {}),
     };
-    const scrutineeRef = varRef(scrutineeName, node.loc);
+    const scrutineeRef = this.narrowableScrutineeRef(node.expression, scrutineeName, cases, node.loc);
 
     const ifChain = this.buildIfChainFromArms(cases, scrutineeRef, node.loc);
     if (!ifChain) return [scrutineeAssign];
@@ -485,6 +486,40 @@ class PatternLowerer {
     const taggedChain: IfElse =
       matchExprId !== undefined ? { ...ifChain, matchExprId } : ifChain;
     return [scrutineeAssign, taggedChain];
+  }
+
+  /** The reference the per-arm conditions and binder reads test against:
+   *  the ORIGINAL variable when narrowing can land on the name arm bodies
+   *  reference, otherwise the `__scrutinee` temp.
+   *
+   *  Bare variable only: evaluate-once is trivially satisfied (no user code
+   *  runs between the chain's conditions), and the typechecker's facts then
+   *  attach to the name the body reads. Two carve-outs keep the temp:
+   *  - `null` parses as a variableName (value "null", the same quirk
+   *    narrowing.ts guards for) and is not a narrowable reference;
+   *  - an arm body that redeclares the scrutinee's name, at any depth:
+   *    generated locals are frame slots (__stack.locals vs __stack.args), so
+   *    a substituted read could target the redeclared slot. Falling back for
+   *    the WHOLE match keeps every arm's reads consistent. */
+  private narrowableScrutineeRef(
+    scrutinee: MatchBlock["expression"],
+    scrutineeName: string,
+    cases: MatchBlock["cases"],
+    loc: SourceLocation | undefined,
+  ): Expression {
+    if (scrutinee.type !== "variableName" || scrutinee.value === "null") {
+      return varRef(scrutineeName, loc);
+    }
+    const name = scrutinee.value;
+    for (const c of cases) {
+      if (c.type !== "matchBlockCase") continue;
+      for (const { node: inner } of walkNodes(c.body)) {
+        if (inner.type === "assignment" && inner.declKind && inner.variableName === name) {
+          return varRef(scrutineeName, loc);
+        }
+      }
+    }
+    return varRef(name, scrutinee.loc);
   }
 
   /**
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -512,7 +512,9 @@ class PatternLowerer {
     }
     const name = scrutinee.value;
     for (const c of cases) {
-      if (c.type !== "matchBlockCase") continue;
+      if (c.type !== "matchBlockCase") {
+        continue;
+      }
       for (const { node: inner } of walkNodes(c.body)) {
         if (inner.type === "assignment" && inner.declKind && inner.variableName === name) {
           return varRef(scrutineeName, loc);
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -50,7 +50,6 @@ import type {
 import type { ValueAccess, AccessChainElement } from "../types/access.js";
 import type { SourceLocation } from "../types/base.js";
 import { mapBodies } from "../utils/mapBodies.js";
-import { walkNodes } from "../utils/node.js";
 import { OBJECT_REST_FN } from "../constants.js";
 
 // ---------------------------------------------------------------------------
@@ -477,7 +476,7 @@ class PatternLowerer {
       // checker can find the owning match for exhaustiveness / union typing.
       ...(matchExprId !== undefined ? { matchExprId } : {}),
     };
-    const scrutineeRef = this.narrowableScrutineeRef(node.expression, scrutineeName, cases, node.loc);
+    const scrutineeRef = this.narrowableScrutineeRef(node.expression, scrutineeName, node.loc);
 
     const ifChain = this.buildIfChainFromArms(cases, scrutineeRef, node.loc);
     if (!ifChain) return [scrutineeAssign];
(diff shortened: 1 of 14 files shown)


packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -487,6 +488,42 @@ class PatternLowerer {
     return [scrutineeAssign, taggedChain];
   }
 
+  /** The reference the per-arm conditions and binder reads test against:
```
egonSchiele: long comment. is the whole thing needed?

packages/agency-lang/lib/typeChecker/flowBuilder.ts
```diff
@@ -240,6 +253,13 @@ const statementRules: StatementRuleTable = {
     const elseEnd = node.elseBody
       ? buildFlowGraph(node.elseBody, elseStart, env)
       : elseStart;
+    // A matchExprId-tagged node is the root of a lowered expression match:
+    // every arm ends in a matchYield, so the merge below would report all
+    // branches exited — but a yield resumes AFTER the match, so post-match
+    // flow is the pre-match flow (same contract as the matchBlock rule below).
+    if (node.matchExprId !== undefined) {
```
egonSchiele, line 270: This drops every fact the arm bodies produced, not just their exits. An arm that reassigns a variable narrowed before the match no longer invalidates it, so `onlyA(u)` after the match is accepted even though an arm set `u` to the other variant. Verified: the shape typechecks clean here and errors when these three files are checked out from `origin/main`.

Consider merging the arm ends as usual and treating a `matchYield` end as "resumes after the match" instead of "exits", so the invalidations survive and only the exit-ness is dropped. If that is out of scope for this PR, please add a pin for the reassign-in-arm shape and say in this comment that arm-body invalidations are discarded.

packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -487,6 +488,42 @@ class PatternLowerer {
...
+    if (scrutinee.type !== "variableName" || scrutinee.value === "null") {
+      return varRef(scrutineeName, loc);
+    }
+    const name = scrutinee.value;
+    for (const c of cases) {
+      if (c.type !== "matchBlockCase") {
+        continue;
+      }
+      for (const { node: inner } of walkNodes(c.body)) {
+        if (inner.type === "assignment" && inner.declKind && inner.variableName === name) {
```
egonSchiele: The doc comment above promises a fallback for a redeclare "at any depth", but this only matches `assignment` nodes with a `declKind`. A nested match binder (`match(w) { { tag: "w", u } => u }`) or an `is` binder inside the arm body binds the same name without producing that node here, and the generated code does write `__stack.locals.u` — the same slot the substituted conditions read. So the guard does not hold for the case it describes.

I could not make it produce a wrong answer, since `Runner.ifElse` memoizes the branch index and step counters stop completed steps from re-running. So either extend the walk to pattern binders, or, if the hazard is not real, drop the fallback and simplify the comment. Right now the two disagree.

packages/agency-lang/lib/typeChecker/narrowing.ts
```diff
@@ -173,14 +173,19 @@ export function analyzeCondition(condition: Expression): ConditionFacts {
       const r = analyzeCondition(condition.right);
       return { then: [], else: [...l.else, ...r.else] };
     }
-    if (condition.operator === "==" || condition.operator === "!=") {
+    // `===`/`!==` are documented stylistic aliases of `==`/`!=` — both compile
+    // to the same `__eq` (typescriptBuilder.ts:1379) — so they justify the
+    // exact same narrowing facts.
+    const STRICT_ALIAS: Record<string, "==" | "!="> = { "===": "==", "!==": "!=" };
```
egonSchiele: Two nits: this table is rebuilt on every `analyzeCondition` call, which runs per condition per pass — hoist it to a module-level const. And the `typescriptBuilder.ts:1379` citation will rot the first time anyone edits that file; naming the function would survive.

egonSchiele (commented):
Reviewed the three fixes. The `===`/`!==` work (fix 2) looks correct and well covered, and the safeBash cleanup reads much better than the workarounds it replaces. Two things I would want resolved before merge, both verified by running the checker in this branch and against `origin/main`.

**1. Fix 1 loses assignment invalidation inside arms (soundness regression vs main).**

Returning the pre-match flow for a `matchExprId`-tagged chain throws away everything the arm bodies did to the flow, not just their exits. An arm that reassigns a variable narrowed before the match no longer invalidates that narrowing:

```agency
def f(w: Nu, x: Nu): string {
  let u: Nu = x
  if (u.tag != "a") { return "no" }
  const s = match(w) {
    { tag: "a" } => {
      u = { tag: "b", n: 1 }
      return "x"
    }
    _ => "y"
  }
  return onlyA(u)          // accepted; `u` may be Nb at runtime
}
```

Run through `typeCheck` on this branch: no errors. Same source with `lib/typeChecker/flowBuilder.ts`, `narrowing.ts` and `patternLowering.ts` checked out from `origin/main`: `Argument type 'Nu | Na' is not assignable to parameter type 'Na' in call to 'onlyA'.` The equivalent `if` branch still errors on both, so this is specifically the new boundary rule. The literal-`matchBlock` form was already unsound this way on main, so the PR widens an existing hole rather than opening a new one, but the pattern-match half is a regression the current pins do not catch: the "reassignment invalidates the narrowing" test keeps the write and the read inside the same arm, which the merge inside the branch still handles.

What I would want instead of a flat `return flow`: merge the arm ends as usual, but treat a `matchYield` end as "resumes after the match" rather than "exits the function" — i.e. keep the invalidations the arms produced and drop only the exit-ness. If that is more surgery than this PR wants, at minimum add a pin for the shape above and say in the comment that arm-body invalidations are dropped, so the next reader knows it is a known hole and not an oversight.

**2. The shadow fallback in fix 3 does not catch pattern binders.**

The walk only looks for `assignment` nodes with a `declKind`, but arm bodies also bind names through nested match patterns and `is` binders, and those land in the same frame slot the comment is worried about. This compiles clean today:

```agency
def pick(w: W): string {
  let u: Su = { tag: "a", s: "hi" }
  return match(u) {
    { tag: "a", s } => {
      const inner = match(w) { { tag: "w", u } => u }
      return s
    }
    _ => "other"
  }
}
```

Generated JS: the outer chain tests `__stack.locals.u` (substituted), and the inner binder writes `__stack.locals.u` inside the arm body. The fallback never fires.

I could not turn that into a wrong answer, because `Runner.ifElse` memoizes the branch index in `__condbranch_*` and step counters stop completed steps from re-running, so nothing re-reads the substituted slot after the arm body starts. Which points at the real question: either the frame-slot hazard is real, in which case the walk needs to cover pattern binders too, or it is not, in which case the fallback is dead weight and the comment overstates the risk. Right now the code claims to guard "at any depth" and does not. Whichever way you go, the comment should match what is enforced.

**Smaller notes**

- The new agency test `match-scrutinee-shadow` pins dispatch and binding, which is what its description promises, but the `print(u)` line is not asserted — and that read currently compiles to `__stack.args.u`, the parameter, not the redeclared local. That is a pre-existing shadowing bug, not this PR's, but it is worth a separate issue since this PR is the thing that made the shape look supported.
- Fixture diff and the stdlib rewrite look right; the `safeBash.md` churn is pure line-number regeneration.

Rest of the change reads well — the test rows for the region boundary, the `alwaysExits` misfire detector and the interrupt-symmetry pin are the right pins to have written.

egonSchiele:
Both findings verified and addressed in 2d45e3ecc. What I did with each:

**1. Arm-body invalidation (confirmed regression, fixed with your suggested semantics).** Reproduced your example: clean on the branch, errors on main. The flat `return flow` is gone. A `matchYield` now registers its flow state under its matchId (`FlowEnvironment.matchYieldFlows`), and the match root — both the lowered chain and the surviving matchBlock — merges those yield-point flows with the base flow and the branch ends. So a yield ends its branch but contributes its state to the post-match join: exit-ness dropped, invalidations kept. Your repro now errors with exactly main's message (`'Na | Nu' is not assignable...`).

Since the registry made it cheap, the merge also covers the two halves you noted were already broken on main: the literal expression form and statement-position matches. Three new pins (lowered / literal-expression / statement), and the earlier boundary pins (post-match narrowing survives, alwaysExits misfire detector, interrupt symmetry) all stay green.

**2. Shadow fallback dropped as dead weight.** You and Copilot were both right that the walk missed binder forms (nested match patterns, `is` binders, for-loop item vars). Rather than extend it, I verified the hazard cannot occur and removed the fallback: every substituted read — the chain conditions and the binder reads — executes before any user statement in the arm body (`foldArms` prepends bindings), and `Runner.ifElse` memoizes the chosen branch so nothing re-reads the slot afterward, including across interrupt resume. The comment now states that argument instead of promising an any-depth scan the code didn't do. The runtime shadow test passes with the substitution active in that match, and the typecheck pin still expects an error — now from ordinary assignment invalidation rather than the fallback.

**Smaller items:** `STRICT_EQUALITY_ALIAS` hoisted to a module const and the citation names `processBinOpExpression` instead of a line number; the long `narrowableScrutineeRef` comment shrank to the safety argument (the fallback bullet it documented no longer exists); and the pre-existing args-slot/locals-slot read bug behind the unasserted `print(u)` is filed as #717 with the repro and the blast-radius notes from your review.

PR #718, opened 2026-07-29
lowering: member-path and literal-index match scrutinees narrow in arm bodies

egonSchiele:
## What

The follow-up recorded in the match-arm-narrowing spec: a member-path
scrutinee now narrows in arm bodies, the same way a bare variable does
since #715.

`narrowableScrutineeRef` also substitutes a **stable access path** — a
variable base with only plain property and literal-index hops, which is
exactly the path class flow narrowing keys on (Reference.chain, M2). So:

```agency
match(b.inner) {
  { tag: "a" } => onlyA(b.inner)   // b.inner: the selected variant
  ...
}
match(xs[0]) {
  { tag: "a" } => onlyA(xs[0])     // per-index, does not leak to xs[1]
  ...
}
```

Calls, method hops, computed indexes, slices, and optional hops keep the
evaluate-once temp: re-reading a call runs user code, a computed index
can hit a different element, and the narrowing store cannot key on any
of them — each is pinned.

Re-reading a property path is safe for the same reason the variable
substitution was: every substituted read (chain conditions + binder
reads) executes before any user statement in the arm body, and the
runner memoizes the chosen branch.

## Docs

docs/dev/typechecker/narrowing/README.md capability table catches up
with this and #715: the scrutinee row, the ===/!== alias note, matchYield
flow termination, and the match-exhaustiveness row that still said
"planned" for a feature the same document describes as shipped below.

## Testing

- New rows: single-hop, multi-hop, and literal-index scrutinees narrow;
  binder typing through a member-path scrutinee; computed-index and
  call scrutinees pinned as keeping the temp.
- Full unit suite: 9749 passed. Fixtures regenerated with no changes
  (no existing fixture matches on a member path). Full agency suite: CI.


packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -488,8 +488,9 @@ class PatternLowerer {
...
 
   /** The reference the per-arm conditions and binder reads test against: the
-   *  ORIGINAL variable for a bare-variable scrutinee, so narrowing lands on
-   *  the name the arm bodies reference; otherwise the `__scrutinee` temp.
+   *  ORIGINAL scrutinee expression when it is a stable reference — a bare
+   *  variable or a stable access path — so narrowing lands on the expression
+   *  the arm bodies reference; otherwise the `__scrutinee` temp.
    *  An arm body that rebinds the name cannot corrupt the substituted reads:
    *  they all run before any user statement (conditions precede the body,
    *  binder assignments are prepended to it — see `foldArms`) and the runner
```
egonSchiele: This justification does not hold once an arm has a guard: guards are user code, they are hoisted into the chain between the arm conditions, and a guard that mutates the path makes later arms test a value the snapshot never had. Verified end to end — a match whose first arm guard sets `b.inner` to the other variant returns `second-9` on this branch and `none` with the conditions pointed back at the temp.

Suggest rewording this to say what is actually guaranteed (no user code runs between a condition and its own binder reads) and calling the guard case out, plus a test pinning whichever dispatch semantics you want to keep.

packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -1591,6 +1595,25 @@ function cloneExpr<T extends Expression>(e: T): T {
...
 
+/** A path the narrowing store can key on and whose re-reads cannot diverge
+ *  from what the conditions tested: a variable base with only plain property
+ *  and literal-index hops (the same path class flow narrowing supports —
+ *  docs/dev/typechecker/narrowing). Calls and method hops could run user code
+ *  on every read, and a computed index could hit a different element, so those
+ *  keep the evaluate-once temp. Optional hops are excluded because the
+ *  narrowing store does not model them — substituting would re-read for no
+ *  narrowing gain. */
+function isStableAccessPath(access: ValueAccess): boolean {
```
egonSchiele, line 1622: Only numeric literal indexes qualify, so `obj["key"]` falls back to the temp even though it is just as stable. Fine as a conservative choice, but the comment (and the README row) say "literal-index", which reads as including string literals — worth saying numeric explicitly.

packages/agency-lang/lib/typeChecker/matchArmNarrowing.test.ts
```diff
@@ -269,18 +269,85 @@ node main() {}`);
...
     { tag: "a" } => onlyA(b.inner)
     _ => "y"
   }
 }
+node main() {}`);
+    expect(errs).toEqual([]);
+  });
+
+  it("a multi-hop member-path scrutinee narrows in arm bodies", () => {
+    const errs = hardErrors(`${SCRUT_HEAD}
```
egonSchiele, line 285: These are all typechecker rows, and the PR notes no fixture changed — so the codegen half of this change (conditions and binder reads now re-read the path) is pinned by nothing. Worth an execution test under `tests/agency/`, the way #715 added `match-scrutinee-shadow`: member-path scrutinee dispatch + a binder, a literal-index scrutinee, and `xs[0]` on an empty array still taking the `null` arm. All three behave correctly today (`__nn` survives on the substituted index reads); they just are not held in place.

egonSchiele (commented):
Nice follow-up, and I confirmed the #715 hole I flagged earlier is closed on this branch: an arm that reassigns a variable narrowed before the match now invalidates it, so the new README row describing that is accurate.

One real finding, plus a test-coverage gap.

**1. Substituting the path changes dispatch when an arm guard has side effects.**

The doc comment argues the substituted reads are safe because "they all run before any user statement". Arm guards are user statements, and they run *between* the chain's conditions. With a stable path the later arms re-read the mutated path instead of the snapshot, so the match picks a different arm than it did before this PR:

```agency
def flip(b: Box): boolean {
  b.inner = { tag: "b", n: 9 }
  return false
}

def pick(b: Box): string {
  return match(b.inner) {
    { tag: "a", s } if (flip(b)) => "first-${s}"
    { tag: "b", n }             => "second-${n}"
    _                           => "none"
  }
}
// pick({ inner: { tag: "a", s: "hi" } })
```

Ran it on this branch: `second-9`. Took the same generated file, pointed the conditions back at `__stack.locals.__scrutinee_5` (the pre-PR shape), ran it again: `none`. The generated code confirms why — the temp is assigned and never read, every condition re-reads `__stack.args.b.inner`, and the guard call is hoisted into `__hoist_0` between arm 1 and arm 2.

Strictly this hazard arrived with #715 (a guard that reassigns the scrutinee *variable* does the same thing), but member paths make it easy to hit: mutating a field of a record you were handed is ordinary code, whereas reassigning the scrutinee name inside a guard is not. I do not think it has to block the PR — the case is exotic and arguably the guard deserves what it gets — but the justification in the comment should stop claiming no user code runs between the substituted reads, and it is worth pinning the chosen semantics with a test so the next person does not "fix" it back by accident.

Secondary effect of the same shape: the `__scrutinee` temp is now a dead assignment for every stable scrutinee. It still has to exist as an AST node (it carries `matchSource` for exhaustiveness), but it costs a runner step and a state slot per match at runtime. Worth a follow-up to skip emitting the step when nothing reads it.

**2. The codegen change has no execution test.**

The PR notes "fixtures regenerated with no changes (no existing fixture matches on a member path)", which means the generated-code half of this change is pinned by nothing — all the new rows are typechecker rows. #715 added `tests/agency/match-scrutinee-shadow` for exactly this reason. I would add one execution test covering a member-path scrutinee and a literal-index scrutinee: dispatch plus a binder, and the empty-array case, since `xs[0]` on an empty array has to keep matching the `null` arm. I checked all three by hand and they behave correctly (`__nn` is preserved on the substituted index reads) — they just are not pinned.

**Smaller notes**

- `isStableAccessPath` accepts only numeric literal indexes, so `obj["key"]` keeps the temp. That is a fine conservative call, but the comment and the README both say "literal-index", which reads as covering string literals. Say numeric.
- Aliasing through a call (`mutate(b)` inside an arm, then reading `b.inner`) does not invalidate the narrowing. That is pre-existing — the identical `if (b.inner.tag == "a")` shape has the same hole on main — so it is not this PR's problem, but the match form now inherits it.
- README changes read well; flipping the exhaustiveness row off "planned" was overdue.

egonSchiele:
All four comments addressed in 549ca39e2.

**Guards (your main finding — confirmed and fixed by falling back).** Reproduced your scenario exactly: the mutating first-arm guard returned `second-9` on the branch, `none` with the temp. Rather than reword the comment around a hole, I closed it: `narrowableScrutineeRef` now keeps the temp whenever **any arm has a guard** — with no guards, no user code can run between a condition evaluation and the last substituted read, so the substitution is observationally equivalent to the temp everywhere it still applies. Dispatch stays a one-time evaluation, uniformly. This also covers the same exposure for bare-variable scrutinees that #715 shipped (a guard reassigning `u`). The trade — guarded matches do not narrow their scrutinee in arm bodies — is pinned by two typecheck rows, and the mutating-guard scenario is pinned at runtime (expects `none`). The doc comment now states the actual guarantee and calls the guard case out.

**Literal-index rule (you + Copilot).** `isStableAccessPath` was quietly duplicating a rule that already has a single source — `toSegment` in `pathSegments.ts`, whose own comment says so. It now delegates to `chainToSegments`, which gets non-negative-integer-only for free (`xs[-1]`, `xs[1.5]` keep the temp) and can never drift from what narrowing actually keys on. Comment and README row now say "numeric" explicitly; `obj["key"]` stays temp-based as before.

**Execution coverage.** `tests/agency/match-path-scrutinee` pins the codegen half with four nodes: member-path dispatch + binder, literal-index dispatch, `xs[0]` on an empty array taking the wildcard arm (the `__nn` normalization on substituted index reads), and the guard-snapshot case above.

Full unit suite 9751 passed, lint clean, no fixture changes.

PR #720, opened 2026-07-29
Self-writing agent experiment: prototype, composable agents, and the fixes it surfaced

egonSchiele:
An experiment in a different agent architecture: instead of one fixed agent loop handling every request, the top-level agent **writes a small Agency program** for the request in front of it, typechecks it, and runs it in a subprocess. Full investigation notes, with measurements and citations, are in `docs/dev/self-writing-agent.md`.

This PR contains the prototype, the first "composable" agents it builds on, and the stdlib/runtime fixes that fell out of investigating why the old fixed-loop agent spent $3.39 and 5.5 minutes on a news query without answering it.

## The prototype (`foo.agency`)

A script that takes a user message, asks a model to write an Agency program answering it, reviews the program (parse + typecheck), runs it via `runCode` in a subprocess, and judges the output. Findings from three probe rounds shaped its details:

- **Fast path**: simple questions are answered directly, with no program, since codegen + typecheck + subprocess launch is pure overhead for "what is the capital of India".
- **Budgets everywhere**: `runCode` now gets a `maxCost` cap, and the prompt requires every generated program to wrap its work in a top-level `guard(time:, cost:)` — demonstrated in the examples, not just described, because the probes showed generated code copies examples faithfully (including their bugs).
- **Warning policy in review**: previously any typechecker warning rejected the program. Now only warnings that mean certain runtime failure reject (AG4004, calling a function that does not exist — the "model invented a function" failure mode). This became load-bearing: a top-level guard always emits the benign AG3009 warning, so the guard requirement and the warning policy had to ship together.
- **Stall detector**: a module-level counter tracks consecutive `execute` failures. From the second failure on, the error message instructs the model to abandon the program and rewrite from scratch instead of patching again.
- **Judge provenance**: the judge previously saw only a bare string and (reasonably) concluded that news dated after its training cutoff was fabricated. It now knows today's date, that the response came from an agent with live web access, and that its job is checking responsiveness and citations, not second-guessing facts it cannot verify.
- **Salvage example**: generated programs are taught to accumulate results into a variable and return it from a `finalize` block, so a tripped budget returns completed steps instead of nothing.
- **Two-phase guidance**: for tasks with a decision point ("research X, and if Y, do Z"), the prompt tells the writer to run a small information-gathering program first, then write a follow-up program using what it learned, rather than encoding speculative branches.
- **cwd anchoring**: generated programs write files relative to the agent working directory, passed explicitly to `runCode`.

## Composable agents (`stdlib/agents/composable/`)

An A/B experiment run in a separate folder: thin agents (a system prompt plus one `llm` call with tools) alongside the existing full-featured ones. If the experiment wins, these replace the originals; if not, the folder is deleted.

- `researcher.agency` — the research core without the guard/review/retry layers. The system prompt was rewritten as an operating manual: method, citation rules, output shape, explicit failure behavior.
- `promptRewriter.agency` — rewrites a vague user prompt into a specific one (stated defaults, explicit deliverable, format) while preserving the user's constraints verbatim.
- `utils.agency` — `ask`, which the probe rounds showed models invent if it does not exist.

## Stdlib and runtime fixes

- **`hostedSearchTools(model)` now actually gates on the model.** It previously ignored its parameter and always requested provider-hosted web search. It now checks whether the model's provider family offers hosted search, bridging the catalog's family/API split (`gpt-4o-mini` is cataloged under `openai` while hosted search registers under `openai-responses`). The gate errs open — unknown models still request search, because wrongly withholding it fails silently while wrongly requesting it produces a visible provider error. Only a model whose provider positively offers no hosted search (e.g. a registered local model) gets none.
- **`typecheck(source, ignoreCodes)`** — new optional parameter dropping named diagnostic codes from the report, for programmatic callers that typecheck code they did not write and cannot annotate with per-line suppressions.
- **Oversized tool schemas now warn.** New `client.maxToolSchemaChars` config (default 2000, 0 disables) emits a statelog warning once per tool per run. Found via `std::syntax::highlight` serializing to a 17,775-character schema — a third of a tool budget — because `theme` accepted a nested color-scheme object; it now takes a union of the eight built-in scheme names (807 chars).
- **Search API keys are bound before the model sees the tool.** The `apiKey` parameter used to be in the tool schema; a real run had the model fill it with tool-call markup, which reached the HTTP layer as an Authorization header. Keys now bind via `.partial()` so the slot does not exist.

## Tests

- `lib/stdlib/llm.hostedSearchTools.test.ts` — 6 unit tests for the search gate, including the family rule and a registered local-model provider getting nothing.
- `tests/agency/typecheck-ignore-codes.*` — execution tests for the new parameter (no LLM calls).
- `tests/agency/agents/searchTools.*` — updated to pin the new gate semantics (fail-open, family rule).
- `lib/runtime/toolSchemaSize.test.ts`, `lib/stdlib/searchKeyBinding.test.ts` — schema-size warning and key binding.

`foo.agency`, `bar.agency`, and `trace.agency` are experiment scratch files, expected to move or disappear as the experiment settles.


egonSchiele (commented):
Reviewed the full diff. The direction is sound and the fixes that fell out of the investigation are good ones — the API-key binding in particular closes a real hole, and the "guards must capture their result" note in the prompt is exactly right (I confirmed separately that a bare `guard(...) { return x }` yields `undefined`, so that guidance is load-bearing, not stylistic).

Below: things I think need a decision before merge, then smaller notes, then what I checked and found fine.

---

## Needs a decision

### 1. `agency.json` flips the repo-wide default model — and it is not in the PR description

```diff
-    "defaultModel": "gpt-4o-mini"
+    "defaultModel": "claude-sonnet-4-6",
+    "defaultProvider": "anthropic"
```

This changes the default for every run and every test in the repo, not just the experiment, and moves the baseline cost per call up substantially. It also silently changes what any test without an explicit model is exercising.

If the experiment needs Sonnet, I would rather see it set on the `llm()` calls in `foo.agency` than in the repo-wide config. If the repo default genuinely should change, it deserves its own PR and a line in the description.

Also, this file loses its trailing newline (`\ No newline at end of file`).

### 2. `composable/researcher.agency` bypasses the gate this PR adds

The headline stdlib fix here is making `hostedSearchTools(model)` actually consult the model. But the new composable researcher hardcodes both sides of that decision:

```agency
hostedTools: ["web_search"],
provider: "openai-responses",
```

Two consequences:

- It requests hosted search unconditionally — the exact behaviour `_hostedSearchTools` was written to stop. A local model routed through here gets the request anyway.
- Pinning `provider` overrides the caller's model selection entirely. `--model` / `--provider` and any branch default set via `setLlmOptions` are ignored for every research call. For a module whose whole purpose is being composable, that is a strong opinion to bake in.

I would expect `hostedTools: hostedSearchTools(model)` and the provider left to the caller.

### 3. `typecheck(source, ignoreCodes)` has tests but no caller

The parameter is justified in the description by "programmatic callers that typecheck code they did not write" — but the motivating caller, `review()` in `foo.agency`, does not use it. It calls `typecheck(code)` and filters afterward, because it needs the opposite shape: an allow-list of *fatal* codes, not a deny-list of ignorable ones.

So the new API arrives with no production consumer. That is not wrong, but it is worth being deliberate: either `review()` should use it, or it should land in its own PR where the use case can be argued on its own terms.

### 4. The warning policy fails open on codes that do not exist yet

```agency
static const rejectingWarningCodes = ["AG4004"]
```

This is an allow-list of *fatal* warnings, so any warning code added to the typechecker later is silently treated as benign in generated code. Given the failure mode this exists to catch — the model inventing a function — I would rather see the inverse: enumerate the codes known to be benign in generated programs (`AG3009` today), and let anything else surface. That way a new diagnostic defaults to being noticed rather than ignored.

At minimum, worth a comment explaining that the fail-open direction is deliberate, the way `_hostedSearchTools` documents its own fail-open choice.

---

## Smaller notes

**`promptRewriter` is an unconditional extra round trip.** Every `researcherAgent` call now pays a full LLM call before any research starts. On a project whose stated goal is answering a news query in under a minute, that is on the critical path. The prompt itself says an already-specific prompt "should come back nearly unchanged" — which is the case where the call is pure latency. Worth either gating it (skip when the task is already concrete) or measuring what it costs against what it buys.

**`researcherAgent` has no declared return type.** `export def researcherAgent(topic: string) {` — for a stdlib entry point that other generated code calls, an explicit `: string` would document the contract and catch drift.

**`stdlib/agency.agency` carries unrelated formatter churn.** The diff rewrites public type declarations that have nothing to do with this PR:

```diff
-  docstring: string | null,
+  docstring?: string;
```

I checked whether this changes behaviour: it does not. `agency fmt` performs this rewrite itself (`x: string | null` → `x?: string`), and the emitted JSON Schema is byte-identical for both forms — same `anyOf`, same `required` list. So it is safe, but it is diff noise in a feature PR, and it now disagrees with `docs/dev/template-agency.md`, which documents `holesOf` returning `"type": null`. Two options: revert the unrelated hunks, or split the formatter pass into its own commit so this PR's diff is only the work.

Separately: that `fmt` rewrite is arguably a formatter bug worth its own issue. Silently changing a declared type — even to an equivalent one — is more than formatting.

**Scratch files at the package root.** `foo.agency`, `bar.agency`, `trace.agency` are acknowledged as temporary in the description. Worth confirming they are excluded from the published package and from lint/typecheck CI before this lands on `main`, so they do not become someone else's puzzle.

---

## Checked and fine

- **API key binding.** `.partial(apiKey: env(...)).rename(toolName)` is correct, and the order matters — `.partial()` preserves the base name, so `.rename()` has to come last or both search tools collide. Confirmed the bound schemas no longer contain `apiKey` and keep the right agent-facing names.
- **`_hostedSearchTools` tests.** I checked the last one specifically, since `registerModelData({ hostedTools: [] })` looked like it might make the assertion pass for the wrong reason. It does not: `getHostedTools` *merges* registered data over the baseline rather than replacing it, so the baseline entries are still present and the test genuinely exercises the family rule. Good test.
- **`thread { return x }` in the composable agents propagates correctly.** I checked, because the `guard` case in this same PR does not, and the two blocks reading identically while behaving differently is a real footgun. Might be worth a language-level issue.
- **The stall counter uses a module-level `let`,** which is the right call in Agency — global variables get per-run state isolation, so it cannot leak across runs.
- **The fast path, the budget requirement, and the salvage example** all follow directly from the probe findings. The decision to demonstrate the guard rather than only describe it is the right lesson to have drawn.

---

Nothing here blocks the experiment; items 1 and 2 are the ones I would want resolved before this lands on `main`, since they change behaviour outside the experiment folder.

Issue #719, opened 2026-07-29
Fill-time type checking skips holes whose type is imported


egonSchiele:
`fill` now checks a plain value against the hole's declared type — a record missing a required property is rejected at the fill instead of when the generated program compiles. That checking is skipped whenever the hole's type cannot be fully resolved, which includes every type the template **imports**.

## Why the skip exists

A template carries its own type declarations, so `type Person = { … }` written in the template can be resolved from the `Code` value itself. An imported name cannot: resolving it means reading another module, which fill has no way to do.

Skipping is not a shortcut, it is the only safe answer. An unknown alias resolves to itself, and a synthesized record compared against it is simply **not assignable** — so checking anyway would reject every correct fill in such a template. Verified:

```
resolveType(unknownAlias, {})                → typeAliasVariable, unchanged
isAssignable(record, unresolvedAlias, {})    → false
```

`hasUnresolvedName` in `lib/runtime/template/fill.ts` implements the skip. It follows alias bodies, so `type Person = { pet: Animal }` with `Animal` undeclared is unresolvable too.

## Why it is worth closing

The skip is silent. A user reading "fill checks records" in the guide will reasonably believe their template is covered, and as shared type files become the normal way to organize templates, the share of holes that are actually checked drops without anyone noticing.

The guide now states the boundary explicitly, so this is a capability gap rather than a correctness bug — but the gap widens on its own as template libraries grow.

## What closing it needs

Module resolution at fill time: given the template's own import statements, load the referenced modules and add their exported types to the alias table.

This is the same capability the deferred `Code`-fragment checking needs. Fill-time validation currently passes any real code fragment through untouched, because checking one means resolving names against the completed program's module scope. Both want "resolve this template's imports", so they belong together and probably want one entry point.

## Notes for whoever picks it up

- The alias table is built by `aliasTableFrom` (`lib/runtime/template/aliasTable.ts`), currently from the template's own top-level `typeAlias` nodes. That is the place to extend.
- Aliases declared inside a function or node body are also skipped today, deliberately — a body-scoped alias is not in scope at a top-level hole. Worth confirming that stays true rather than folding it into this change.
- Keep the fail-open property. If a type still cannot be resolved after imports are followed, it must go on being skipped; a resolution failure must never become a rejection.
- `lib/runtime/template/fill.test.ts` has three acceptance tests under "a type it cannot resolve is not checked" that pin the current behavior. Closing this should convert the imported-type one into a rejection test and leave the other two.

[...]

PR #721, opened 2026-07-29
fix: fill checks a plain value against the hole's real type

egonSchiele:
A template can say a hole needs a `Person`. `fill` did not check that it got one:

```ts
type Person = {
  name: string;
  age: number
}

const person: Person = #person
```

```ts
fill(template, { person: { name: "Alice" } })   // succeeded — age is missing
```

The failure surfaced much later, when `runCode` compiled the generated program, in a message about source the caller never wrote. This also succeeded, which is the giveaway that it was not really about records:

```ts
type Name = string
const n: Name = #who

fill(t, { who: 42 })
```

The hole's type was essentially never checked unless it was spelled `string`, `number` or `boolean` literally.

## Why

Two independent causes, either one enough on its own.

**The expected type was flattened to text before anything compared it.** By the time it reached `assertFillerType` it was the string `"Person"`, and the check was `if (!["string","number","boolean"].includes(expectedType)) return;`.

**The value could not be described either.** `certainTypeOf` handled three primitives and single-literal fragments; every object and array came back "unknowable", meaning skip.

## The fix

- `synthesizeType` describes a plain value as a type, or returns null when it cannot.
- `aliasTableFrom` builds the template's own type aliases from its AST — `type Person = { … }` travels with the `Code` value being filled.
- `assertFillerType` compares them with the type checker's own `isAssignable`, so unions, optional properties, nesting and recursive aliases all work without a second implementation of "does this type fit that type".
- `explainMismatch` names what is wrong, on the failure path only.

## The design invariant

**Fill may only reject a value the completed program's compile would also reject.** Missing things is fine — that is what "validation, not a guarantee" means. Refusing something that would have compiled is not, because the caller has no way to argue with it.

Two mechanisms defend it, and both were added because the alternative silently broke ordinary templates.

**Unresolvable types are skipped.** An unknown alias resolves to itself and a synthesized record compared against it returns `false` — verified — so without this, any template importing its types, or declaring one inside a body, would reject every record fill. `hasUnresolvedName` follows alias bodies, so `type Person = { pet: Animal }` with `Animal` undeclared is unresolvable too.

**Strings are described literally on the failure path.** The checker infers a plain string as a string-literal type and leaves numbers and booleans widened. So `const mode: "fast" | "slow" = "fast"` compiles while a widened `string` does not fit that union. The fast path stays widened and deduplicated (a large array must not become an equally large union); when it decides to reject, the value is re-described with strings as literals and re-checked. That can only turn a rejection into an acceptance, so it cannot introduce a false accept, and the cost lands only on fills that were about to throw. Verified in both directions: the compile accepts `"fast"` against the union and rejects `1` against `1 | 2`, so numbers and booleans stay widened in both modes.

## What the errors look like

Before:

```
The hole `#person` expects `Person`, but the fill supplies `{ name: string }`.
```

Now:

```
The hole `#person` expects `Person`, but the fill is missing the required property `age`.
The hole `#person` expects `Person`, but the fill is missing the required property `address.city`.
The hole `#person` expects `Person`, but the fill has `age` as `"thirty"` where `number` is expected.
```

The explainer never decides whether to reject — `isAssignable` has already decided, and every sub-question the explainer asks goes through `isAssignable` too. Comparing printed type names instead would mis-blame any aliased property: given `name: Name` and a missing `age`, it would report `name` as wrong while the real problem went unnamed. When it cannot localize the problem, the general two-types message is the fallback.

## Splices

A splice annotation describes **one spliced element**, not the array — that is what `fillOne` already did, and it becomes visible now that records are checked. `#...people: Person[]` is the natural mistake and rejects everything, so the error says so:

```
The splice `#...people` describes one element, not the array — its type should be `Person`, not `Person[]`.
```

Only when the evidence supports it: the element must actually fit one level down. `#...rows: number[]` splicing rows of numbers is legitimate, and a bad element there gets the ordinary message rather than advice to change a correct annotation.

## Behavior change

`fill` now rejects inputs it previously accepted. A template and filler that work today only because the missing field is never read on the path the generated program takes will start failing at the fill. That is the intended trade and the point of the change, but it is a real behavior change rather than a pure addition.

## Deliberately unchanged

- Real `Code` fragments still pass through unchecked and are judged when the completed program compiles. The one exception is the pre-existing check for a fragment holding a single literal, which is preserved and now resolves aliases too.
- `@validate(...)` validators do not run at fill. This checks shape, not constraints.
- Imported types are not checked, for the reason above. The guide states the boundary; #719 tracks closing it, and the code comment points there.

## Testing

- Unit: 9822 passed, 0 failed.
- `tests/agency/templates`: 15 passed, including a new fixture that runs the guide's own `Person` example end to end and asserts the fill is rejected naming `age`.

The tests worth knowing about: the string-literal union case pins the design invariant in both directions; three acceptance tests pin that unresolvable types are skipped rather than rejected; and a pair pins that the pre-existing literal-fragment check survived the rewrite and now resolves aliases.

### DIFF
--- a/packages/agency-lang/docs/site/guide/template-agency.md
+++ b/packages/agency-lang/docs/site/guide/template-agency.md
@@ -111,6 +111,30 @@ The hole `#topic` expects `string`, but the fill supplies `number`.
 
 This is a runtime failure.
 
+Records are checked the same way. If the hole wants a `Person` and the object you supply is missing a required property, the fill fails and tells you which one:
+
+```
+The hole `#person` expects `Person`, but the fill is missing the required property `age`.
+```
+
+For a property nested inside another, the error gives you the path — `address.city` rather than just "address is wrong".
+
+Type aliases are resolved, so `type Name = string` behaves exactly like `string`.
+
+What is checked is the *shape* of the value. Validators attached with `@validate(...)` do not run at fill time — a value can pass the fill and still fail its validator when the completed program runs.
+
+Fill checks types the template declares itself. A type the template **imports** cannot be resolved while the template is only a value, so a hole using an imported type is not checked at fill — that mistake is caught when the completed program compiles, as it was before. This matters as soon as you keep shared types in their own file, so it is worth knowing which of your templates are actually covered.
+
+### Splice holes describe one element
+
+A splice hole fills several items at once, and its type describes **one of them**, not the list:
+
+```ts
+[| f(#...people: Person) |]
+```
+
+Each element of the array you supply is checked against `Person`. Writing `#...people: Person[]` is the natural mistake and rejects everything, because no single person is a list of people. The error says so when it can.
+
 One more example. This time the hole is an object:
 
 ```ts
--- a/packages/agency-lang/docs/site/guide/template-agency.md
+++ b/packages/agency-lang/docs/site/guide/template-agency.md
@@ -111,30 +111,6 @@ The hole `#topic` expects `string`, but the fill supplies `number`.
 
 This is a runtime failure.
 
-Records are checked the same way. If the hole wants a `Person` and the object you supply is missing a required property, the fill fails and tells you which one:
-
-```
-The hole `#person` expects `Person`, but the fill is missing the required property `age`.
-```
-
-For a property nested inside another, the error gives you the path — `address.city` rather than just "address is wrong".
-
-Type aliases are resolved, so `type Name = string` behaves exactly like `string`.
-
-What is checked is the *shape* of the value. Validators attached with `@validate(...)` do not run at fill time — a value can pass the fill and still fail its validator when the completed program runs.
-
-Fill checks types the template declares itself. A type the template **imports** cannot be resolved while the template is only a value, so a hole using an imported type is not checked at fill — that mistake is caught when the completed program compiles, as it was before. This matters as soon as you keep shared types in their own file, so it is worth knowing which of your templates are actually covered.
-
-### Splice holes describe one element
-
-A splice hole fills several items at once, and its type describes **one of them**, not the list:
-
-```ts
-[| f(#...people: Person) |]
-```
-
-Each element of the array you supply is checked against `Person`. Writing `#...people: Person[]` is the natural mistake and rejects everything, because no single person is a list of people. The error says so when it can.
-
 One more example. This time the hole is an object:
 
 ```ts
--- a/packages/agency-lang/lib/runtime/template/aliasTable.ts
+++ b/packages/agency-lang/lib/runtime/template/aliasTable.ts
@@ -1,0 +1,44 @@
+import type { AgencyNode, TypeAlias, TypeAliasEntry } from "../../types.js";
+
+/**
+ * A template's own type aliases, in the shape the type checker's resolver
+ * expects.
+ *
+ * A template carries its type declarations with it — `type Person = { … }`
+ * is part of the same `Code` value being filled — which is what makes
+ * resolving a hole's declared type possible at run time without a compile.
+ *
+ * Top level only. Aliases declared inside a body are not in scope at a
+ * top-level hole, and an alias this table cannot find is treated as
+ * unknowable by the caller rather than as an error.
+ *
+ * `typeParams`, `valueParams` and `tags` are carried, not just `body`:
+ * generic and value-parameterized aliases are legal in templates, and
+ * dropping those fields would make such an alias resolve WRONGLY rather
+ * than not at all — the worse failure.
+ */
+export function aliasTableFrom(
+  nodes: AgencyNode[],
+): Record<string, TypeAliasEntry> {
+  // Alias names come from user source, so null-prototype (house pattern).
+  const table: Record<string, TypeAliasEntry> = Object.create(null);
+  for (const node of nodes) {
+    if (node.type !== "typeAlias") continue;
+    const alias = node as TypeAlias;
+    const entry: TypeAliasEntry = { body: alias.aliasedType };
+    if (alias.typeParams !== undefined) {
+      entry.typeParams = alias.typeParams;
+    }
+    if (alias.valueParams !== undefined) {
+      entry.valueParams = alias.valueParams;
+    }
+    if (alias.tags !== undefined) {
+      entry.tags = alias.tags;
+    }
+    if (alias.isEffectSet !== undefined) {
+      entry.isEffectSet = alias.isEffectSet;
+    }
+    table[alias.aliasName] = entry;
+  }
+  return table;
+}
--- a/packages/agency-lang/lib/runtime/template/aliasTable.ts
+++ b/packages/agency-lang/lib/runtime/template/aliasTable.ts
@@ -15,7 +15,11 @@ import type { AgencyNode, TypeAlias, TypeAliasEntry } from "../../types.js";
  * `typeParams`, `valueParams` and `tags` are carried, not just `body`:
  * generic and value-parameterized aliases are legal in templates, and
  * dropping those fields would make such an alias resolve WRONGLY rather
- * than not at all — the worse failure.
+ * than not at all — the worse failure. `typeParams` is load-bearing in a
+ * second way: `hasUnresolvedName` reads it to tell an alias's own bound
+ * parameter (`T` in `type Box<T> = { item: T }`) from a name that genuinely
+ * cannot be resolved. Without it every generic alias would look unresolvable
+ * and be skipped.
  */
 export function aliasTableFrom(
   nodes: AgencyNode[],
--- a/packages/agency-lang/lib/runtime/template/aliasTable.ts
+++ b/packages/agency-lang/lib/runtime/template/aliasTable.ts
@@ -1,35 +1,25 @@
 import type { AgencyNode, TypeAlias, TypeAliasEntry } from "../../types.js";
 
 /**
- * A template's own type aliases, in the shape the type checker's resolver
- * expects.
+ * A template's own type aliases, in the shape the checker's resolver wants.
  *
- * A template carries its type declarations with it — `type Person = { … }`
- * is part of the same `Code` value being filled — which is what makes
- * resolving a hole's declared type possible at run time without a compile.
- *
- * Top level only. Aliases declared inside a body are not in scope at a
- * top-level hole, and an alias this table cannot find is treated as
- * unknowable by the caller rather than as an error.
- *
- * `typeParams`, `valueParams` and `tags` are carried, not just `body`:
- * generic and value-parameterized aliases are legal in templates, and
- * dropping those fields would make such an alias resolve WRONGLY rather
- * than not at all — the worse failure. `typeParams` is load-bearing in a
- * second way: `hasUnresolvedName` reads it to tell an alias's own bound
- * parameter (`T` in `type Box<T> = { item: T }`) from a name that genuinely
- * cannot be resolved. Without it every generic alias would look unresolvable
- * and be skipped.
+ * Top level only: an alias declared inside a body is not in scope at a
+ * top-level hole. `hasUnresolvedName` treats a name missing from this table
+ * as unknowable, so a gap here means fills go unchecked, never wrongly
+ * rejected.
  */
 export function aliasTableFrom(
   nodes: AgencyNode[],
 ): Record<string, TypeAliasEntry> {
-  // Alias names come from user source, so null-prototype (house pattern).
+  // Null-prototype: alias names are user-controlled keys (house pattern).
   const table: Record<string, TypeAliasEntry> = Object.create(null);
   for (const node of nodes) {
     if (node.type !== "typeAlias") continue;
     const alias = node as TypeAlias;
     const entry: TypeAliasEntry = { body: alias.aliasedType };
+    // typeParams is load-bearing twice: the resolver substitutes with it,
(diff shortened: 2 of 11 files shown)


packages/agency-lang/lib/runtime/template/synthesizeType.ts
```diff
@@ -0,0 +1,150 @@
...
+  if (kindOf(code) !== "expr" || code.nodes.length !== 1) return null;
+  const node = code.nodes[0];
+  if (node.type === "number") return NUMBER_T;
+  if (node.type === "boolean") return BOOLEAN_T;
+  if (node.type === "string") {
+    const literal = node as StringLiteral;
+    const interpolated = literal.segments.some(
+      (segment) => segment.type === "interpolation",
+    );
+    return interpolated ? null : STRING_T;
```
egonSchiele: This is the one remaining violation of the invariant stated in `assertFillerType`: `literalFragmentType` never sees `options`, so a string fragment always widens to `string` — even on the literal-accurate second pass. Concretely, `fill(t, { mode: [| "fast" |] })` against a hole typed `"fast" | "slow"` is rejected here, but the lifted program compiles: the graft is the literal `"fast"`, and `synthType` infers it as a string-literal type exactly like a plain-value fill. Same bug class as the plain-string case the second pass exists to fix, one path over.

Fix: thread `options` through and, when `stringsAsLiterals` is set and the literal is a single text segment, return `{ type: "stringLiteralType", value: segments[0].value }` — mirroring `literalToType`'s own eligibility condition (`lib/typeChecker/literalType.ts:17-19`). And add the test: `[| "fast" |]` against `"fast" | "slow"` must be accepted.

packages/agency-lang/lib/runtime/template/fill.ts
```diff
@@ -202,44 +227,142 @@ function nodesFor(hole: Hole, value: unknown, expectedType?: string): AgencyNode
...
+ *
+ * Deliberately fails toward skipping: if this ever over-reports, some fills
+ * go unchecked, which is the same weaker-but-safe position the feature
+ * started from. Under-reporting would reject correct programs.
+ *
+ * The cost is that an imported type is never checked at fill. Closing that
+ * needs module resolution at fill time — the same capability the deferred
+ * fragment checking wants. Tracked as #719.
+ */
+function hasUnresolvedName(
```
egonSchiele, line 322: Verified by probe on this branch: a generic alias's body refers to its own type parameters as `typeAliasVariable` nodes — `type Box<T> = { item: T }` parses with body `{ item: typeAliasVariable "T" }` and `typeParams: [{ name: "T" }]`. `T` is not a key in the alias table, so this walk reports every generic alias unresolvable, and a hole typed `Box<string>` is never checked at fill.

That is fail-open, so nothing breaks — but it is silent, no test pins it, and `aliasTable.ts`'s doc comment says `typeParams` are carried so generic aliases do not "resolve WRONGLY", which reads as a promise that they are checked. Two honest options: (a) when recursing into an alias body, treat names bound by that alias's own `typeParams` as resolved, so `Box<string>` genuinely gets checked; or (b) keep the skip, pin it with a test (`Box<string>` hole, wrong filler, accepted), and reword the `aliasTable.ts` comment. Either is defensible; the current state — skip, untested, comment implying otherwise — is not.

packages/agency-lang/lib/runtime/template/fill.ts
```diff
@@ -202,44 +227,142 @@ function nodesFor(hole: Hole, value: unknown, expectedType?: string): AgencyNode
...
+  // into an acceptance (a literal type fits everywhere `string` does), so
+  // it cannot introduce a false accept, and its cost lands only on fills
+  // that were about to throw.
+  const literalAccurate = synthesizeType(value, { stringsAsLiterals: true });
+  if (literalAccurate !== null && isAssignable(literalAccurate, expectedType, aliases)) {
+    return;
+  }
+
+  const printedExpected = variableTypeToString(expectedType, {}, true);
+  const resolvedExpected = resolveType(expectedType, aliases);
```
egonSchiele: `resolveType` throws a `TypeError` on invalid generic forms — that is why `safeResolveType` exists (`assignability.ts:51-61`, falling back to `any`). The reject decision above survived because `isAssignable` resolves through the safe wrapper internally; this direct call on the error path can throw where the decision did not, turning a would-be validation message into an unhandled `TypeError` from inside `fill`. Same applies to `explainMismatch.ts:32`. Cheap fix: use `safeResolveType` in both places — on this path a fallback to `any` just means the splice hint or the explainer declines, which is their contract anyway.

packages/agency-lang/lib/runtime/template/fill.ts
```diff
@@ -202,44 +227,142 @@ function nodesFor(hole: Hole, value: unknown, expectedType?: string): AgencyNode
...
+  // checked against the annotation, so an array type rejects every element.
+  // Only claim this when the evidence supports it — the element FITS one
+  // level down, so the annotation is one level up. An array-typed splice is
+  // sometimes correct (`#...rows: number[]` splicing rows of numbers), and
+  // a genuinely bad element there must get the ordinary message, not advice
+  // to change an annotation that is right.
+  if (
+    hole.splice &&
+    resolvedExpected.type === "arrayType" &&
+    isAssignable(actual, (resolvedExpected as ArrayType).elementType, aliases)
```
egonSchiele: The hint's evidence check uses the widened `actual`, but the acceptance decision it mirrors got a literal-accurate second chance. So `#...modes: ("fast" | "slow")[]` fed `["fast"]` rejects (correctly — the annotation is one level up) without the hint: widened `string` does not fit `"fast" | "slow"`, even though the element fits literally, which is exactly the one-level-off evidence. Use the literal-accurate description here — `literalAccurate ?? actual` — so the hint's evidence rule matches the same inference the compile would apply.

egonSchiele (commented):
Strong implementation — every planned safeguard is in and testable: the unresolved-name skip (with the alias-body recursion the deep test forced), the two-pass literal-accurate retry with the design invariant stated where the decision is made, the evidence-guarded splice hint with its negative test, and the explainer that routes every sub-question through `isAssignable`. The test suite pins both directions of the invariant (`"fast"` accepted against `"fast" | "slow"`; `1` rejected against `1 | 2` because the compile rejects it too), and the exact-phrasing assertion on the wrong-property message closes the loophole where the general message coincidentally contains the property name.

Four findings, two of them verified by probe against this branch. The first is the same invariant-violation class the plan review caught, one path over — worth fixing before merge. The others are smaller.

1. **Fragment path still violates the invariant** (inline at `synthesizeType.ts`). `literalFragmentType` ignores `stringsAsLiterals`, so `fill(t, { mode: [| "fast" |] })` against a `"fast" | "slow"` hole is rejected while the lifted program compiles.
2. **Generic aliases silently skip** (inline at `fill.ts:342`). Verified by parsing `type Box<T> = { item: T }`: the body holds `typeAliasVariable "T"`, `T` is not in the alias table, so `hasUnresolvedName` reports every generic alias unresolvable and a `Box<string>` hole is never checked. Fail-open, so no false rejections — but it is silent, untested, and `aliasTable.ts`'s comment ("dropping those fields would make such an alias resolve WRONGLY") reads as though generic aliases are checked.
3. **Two direct `resolveType` calls can throw** where the reject decision's own resolution did not (`fill.ts:288`, `explainMismatch.ts:32`) — `safeResolveType` exists for exactly this.
4. **The splice hint checks the widened description** (`fill.ts:299`), so an element that fits only literally (`"fast"` under `#...modes: ("fast" | "slow")[]`) misses the hint it deserves.

Two loose ends from the plan, not blocking: the extra-properties soft spot ("add a test during implementation recording whatever `isAssignable` does") has no test in this diff, so that behavior is still unpinned; and the three-way agreement invariant got its doc comment in `synthesizeType.ts` but the planned cross-reference in `liftValue` was not added — the invariant is only discoverable from one of its three parties.

egonSchiele:
All four taken, plus both loose ends, in 5778f12a1. I verified the two behavioral findings by probe on this branch before changing anything.

**1. Fragment path violated the invariant — confirmed.** Same template, two fillers:

```
PLAIN:    accepted
FRAGMENT: The hole `#mode` expects `"fast" | "slow"`, but the fill supplies `string`.
```

`literalFragmentType` never saw `options`, so the second pass could not help it. It now threads them and mirrors `literalToType`'s eligibility condition — single text segment, and widen otherwise, which is what `synthType` does when `literalToType` declines.

**2. Generic aliases — confirmed, and worth more than a pin.** The probe showed the alias table holding `Box` with body `{ item: typeAliasVariable "T" }` and `typeParams: [{ name: "T" }]`, and a `Box<string>` hole accepting `{ item: 42 }`.

You offered two honest options. Before choosing I checked whether the checker can actually do the work:

```
GOOD {item:string} vs Box<string>: true
BAD  {item:number} vs Box<string>: false
```

It can, so I took option (a): names bound by an alias's own `typeParams` count as resolved, and `Box<string>` is genuinely checked now. The bound set is replaced rather than extended when recursing into a nested alias's body, since one alias's parameters are not in scope in another's. Test covers both directions.

**3 and 4** taken as written. Both `resolveType` calls are `safeResolveType`, with the reasoning in a comment at the call site; the splice hint uses `literalAccurate ?? actual` so its evidence rule matches the inference the acceptance decision got.

**Both loose ends closed.** Extra properties are pinned by a test that records whatever `isAssignable` decides rather than asserting a rule this change invents (it accepts them). And `liftValue` now carries the third corner of the three-way invariant, so it is discoverable from more than one party.

**One thing I got wrong and the build caught.** My first version of the literal-fragment fix computed the narrowing as a separate boolean, which TypeScript cannot narrow through — `error TS2339: Property 'value' does not exist on type 'PromptSegment'`. vitest does not typecheck, so the whole test suite went green while `make` failed. Restructured to narrow inline. Worth noting for anyone reading the flow: a green `pnpm test:run` is not evidence the branch compiles.

I also checked the three new tests are not vacuous by reverting the source and re-running them: all three fail without their fix, and the extra-properties pin passes either way, which is correct for a test that records existing behavior.

Re-verified after everything: `make` exits 0, 9826 unit tests pass, 15/15 template fixtures, `lint:structure` clean.

packages/agency-lang/docs/site/guide/template-agency.md
```diff
@@ -111,6 +111,30 @@ The hole `#topic` expects `string`, but the fill supplies `number`.
 
 This is a runtime failure.
 
+Records are checked the same way. If the hole wants a `Person` and the object you supply is missing a required property, the fill fails and tells you which one:
```
egonSchiele: undo all changes to this file

packages/agency-lang/lib/runtime/template/aliasTable.ts
```diff
@@ -0,0 +1,48 @@
+import type { AgencyNode, TypeAlias, TypeAliasEntry } from "../../types.js";
+
+/**
+ * A template's own type aliases, in the shape the type checker's resolver
```
egonSchiele: You're in the habit of writing extremely long comments. Is this entire comment needed or can you make it more concise and readable? Ask yourself: would the next person who touches this file need everything contained in this comment?

packages/agency-lang/lib/runtime/template/fill.ts
```diff
@@ -202,44 +227,166 @@ function nodesFor(hole: Hole, value: unknown, expectedType?: string): AgencyNode
...
+  if (literalAccurate !== null && isAssignable(literalAccurate, expectedType, aliases)) {
+    return;
+  }
+
+  const printedExpected = variableTypeToString(expectedType, {}, true);
+  // safeResolveType, not resolveType: the latter throws on an invalid
+  // generic form, and the reject decision above survived only because
+  // isAssignable resolves through the safe wrapper internally. Throwing
+  // here would turn a validation message into an unhandled TypeError. A
+  // fallback to `any` just means the hint declines, which is its contract.
```
egonSchiele: Similar question about all the comments in this file. Are all these comments needed? Can you make them more concise and readable? Ask yourself: would the next person who touches this file need everything contained in these comments?

egonSchiele:
Done in 6a9db6033.

**Guide reverted.** `docs/site/guide/template-agency.md` is byte-identical to main again.

One consequence worth a decision, not an argument: the "fill checks types the template declares itself; imported types are not checked" boundary now lives only in #719 and a code comment. That was the earlier round's answer to the feature being silently inert for import-typed templates. Happy to leave it as is — say the word if you want it documented somewhere other than the guide.

**Comments cut, hard.** 189 lines of comment removed against 83 added across the five files. The test I used for each one: would the next person who touches this be misled without it?

Kept — each is a place where doing the obvious thing reintroduces a bug I actually hit:
- The invariant on `assertFillerType`, four lines, because everything else in the file follows from it.
- Why an unresolved name is a skip and not a rejection (an unknown alias compares as "does not match", so the obvious reading is backwards).
- Why `safeResolveType` rather than `resolveType` on the error path.
- Why bound type parameters are not unresolved names, and why the bound set is replaced rather than extended.
- Why the literal retry cannot introduce a false accept.

Cut: the restatements of what the code plainly does, the worked examples inline in prose, and the three-paragraph version of the three-way agreement note, now four lines in `synthesizeType` with one-line pointers from `liftValue`.

Re-verified after both changes: `make` exits 0, 9826 unit tests, 15/15 template fixtures, `lint:structure` clean.

Issue #724, opened 2026-07-29
validator interrupts never surface when unanswered (silent no-match in patterns, success(undefined) through bang)


egonSchiele:
Measured while executing the validators-are-predicates plan. A validator
can raise an interrupt; if NOTHING answers it (no handler, no --approve
or --reject policy), the interrupt never surfaces to the endpoint at
all - the question is never asked:

- In a match arm (`a: Checked`) the arm silently reads as no-match and
  the wildcard runs. Pinned as a KNOWN HOLE row in
  lib/cli/validatorInterrupts.spawn.test.ts ("an unanswered validator
  interrupt is silently a no-match") - flip it when this is fixed.
- Through bang (`const a: Checked! = x`) it is worse: the result is
  success with value undefined.

Cause: the validator's own frame halts with the interrupt, .invoke()
returns it as a value, and nothing on the validation path has a
hasInterrupts call site to relay the halt upward - the type-test call
is codegen-synthesized (invisible to hoistCalls; see the recorded
ruling in lib/utils/expressionSlots.ts), and the chain treats the
interrupt-carrying value as an ordinary non-success.

Note this is different from a REFUSED interrupt, which now propagates
(PR: validators-are-predicates). Refusal means a human answered no;
this hole means nobody was ever asked. Surfacing it needs the halt
machinery (runner + step state) that a runtime helper cannot reach -
likely the same reify-the-validate-call restructure the retired
slot-capture plan weighed against hoistCalls.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #723, opened 2026-07-29
bang returns the zod-parsed object while pattern binders read the original


egonSchiele:
Measured while working on validators-are-predicates (spec:
docs/superpowers/specs/2026-07-28-validators-are-predicates.md, section
"What removal does NOT fix"). Pinned by the acceptance rows
zodBangStrips / zodBinderKeepsOriginal in
tests/agency/validator-predicates.agency - do not delete those rows as
obsolete; they also catch a likely implementation error in the
predicate check (anchoring identity to the pre-parse value).

The validation pipeline zod-parses before validators run, and object
schemas strip unknown keys. So the two validation doors return
different values even with pure predicate validators:

```agency
@validate(passthrough)
type Person = { name: string }

const raw: any = { name: "al", extra: "keep" }
const p: Person! = raw     // p.value is { name: "al" }  - stripped
match(raw) {
  q: Person => q            // q is { name: "al", extra: "keep" } - original
  _ => ...
}
```

Candidate fixes, each with real blast radius:
1. deliver the parsed value through pattern binders (a narrower revival
   of the retired slot-capture design);
2. make the validation parse non-stripping - which changes what fill()
   and every structured-output path hand back, so it needs its own
   investigation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #722, opened 2026-07-29
lint: flag validators whose success(...) argument is not their parameter


egonSchiele:
Follow-up recorded by the validators-are-predicates spec
(docs/superpowers/specs/2026-07-28-validators-are-predicates.md).

Validators may no longer modify values - a modifying validator throws at
runtime. But the runtime error only fires when the transforming BRANCH
runs: a validator like

```agency
def clampAge(age: number): Result<number> {
  if (age < 0) {
    return success(1)   // only reached for negatives
  }
  return success(age)
}
```

passes every test that never feeds it a negative, then errors in
production on the first one. A structural lint (a validator whose
`success(...)` argument is not its parameter) would catch this shape at
compile time instead.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #725, opened 2026-07-29
validators are predicates: modifying a value is a loud runtime error

egonSchiele:
## What (breaking change, no deprecation)

Validators become **predicates**: they accept or reject, and may not
substitute a different value. A modifying validator is a loud runtime
error, enforced per-validator in the runtime chain:

```
validator 'clamp' modified the value; validators may only accept
(return the input) or reject (return a failure)
```

- The validator's name travels **in the message string** (the failure's
  `functionName` is the enclosing caller) - `.partial()` derivations
  keep their base name; anonymous validators report `(anonymous)`.
- `Object.is`, not `!==`: a NaN pass-through is legal. Identity, not
  equality: rebuilding an equal object errors - that is the point of
  loud. Silently ignoring returned values was rejected because existing
  clamp-style validators would silently stop clamping.
- Enforcement is a plain thrown Error converted to a Failure at the
  function boundary - the existing convention for programmer errors.
- Per-validator placement is pinned by a transform-and-back chain test
  (double then halve: an end-of-chain check would miss it), and the
  throw is pinned to escape `__validateChainRecursive` uncaught.
- **Read your validators when upgrading**: a conditional transform
  (clamp only on negatives) will not error until the transforming
  branch runs. #722 proposes a lint that catches the shape at compile
  time.

## Deliberately NOT in this PR

**Interrupts raised inside validators.** An earlier revision of this
branch shipped refusal propagation through a new exception channel
threaded through the generated catches - a second interrupt-handling
mechanism alongside the existing value-relay convention, which review
rightly rejected. That revision is fully reverted here. The redesigned
mechanism (hoisted type-test statements with the standard
hasInterrupts/halt relay, which also surfaces unanswered validator
interrupts - #724) is specced in
docs/superpowers/specs/2026-07-28-validators-are-predicates.md
("Change 2 (v3): the conventional mechanism") and will be discussed
and PRed separately. Until then the current behaviors are PINNED in
lib/cli/validatorInterrupts.spawn.test.ts: approval works through arms
and `is`; a refusal is a visible failure through bang; a refusal reads
as no-match in arms/`is` (KNOWN GAP rows); an unanswered validator
interrupt never surfaces (KNOWN HOLE row, #724).

## Measured along the way

- The `@validate` chain does NOT run on the structured LLM-output path
  (schema-only there; probed with llmMocks) - model output meets
  validators only through bang/patterns.
- Bang returns the zod-parsed value (object types drop unknown keys)
  while pattern binders read the original - pre-existing, independent
  of validators, pinned by two acceptance rows whose comments explain
  their double duty, filed as #723.

## Migration inventory

Every transforming validator in the tree was a pin for the retired
feature; each now pins the new contract: validateTransform becomes THE
enforcement pin (bare `Result` signature kept), the record-validators
transforms node pins the cross-module error (preserving the #630
descriptor-routing point; write-back retired), type-patterns'
clampAge/Repaired become a predicate with reject-means-no-match and
binder-receives-value pins, and the TODO transform entry is retired.
The type-validation guide teaches the predicate contract and the
explicit repair-function idiom.

## Testing

- 12-row acceptance file (tests/agency/validator-predicates): predicate
  pins, message-asserting error rows through bang/arm/is, partial
  application naming, schema-only dispatch, zod-divergence pins.
- 6-row CLI spawn suite (lib/cli/validatorInterrupts.spawn.test.ts):
  current interrupt-in-validator behavior, including the two KNOWN GAP
  pins and the #724 KNOWN HOLE pin.
- Runtime unit rows: per-validator placement, recursive escape,
  NaN/Object.is, rebuilt-equal, empty success(), anonymous naming.
- Full local unit suite: 9841 passed. Fixture typecheck: 167 passed.
  Full agency suite: CI.

Spec: docs/superpowers/specs/2026-07-28-validators-are-predicates.md
(v3).


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -1485,7 +1485,14 @@ export class TypeScriptBuilder {
     if (coarse !== null) {
       return ts.call(ts.id("__coarseTypeTest"), [value, ts.str(coarse)]);
     }
-    return ts.call(ts.id("isSuccess"), [this.validateExpr(node.typeHint, value)]);
+    // `__typeTest` instead of a bare isSuccess: a failure the validator
+    // RETURNED is "no match", but a refused interrupt propagates — refusal
+    // must never become dispatch (validators-are-predicates spec). The
+    // helper is async, so the call itself is awaited (the chain call inside
+    // carries its own await already; awaiting a settled value is a no-op).
```
egonSchiele: The sibling catch this change depends on is missing the clause. `functionCatchFailure.mustache` learned about `InterruptRejectedError`, but the graph-node catch is generated separately (`typescriptBuilder.ts:3216`, unchanged in this PR) and does not. A type test written directly in a `node` body unwinds there instead, and the result is exactly what this PR says it removed — verified on this branch with `--reject`:

```
ERROR Node main crashed: interrupt rejected
ERROR InterruptRejectedError: interrupt rejected
    at __typeTest (dist/lib/runtime/validateChain.js:10:15)
    at async Runner.ifElse (dist/lib/runtime/runner.js:800:17)
```

plus a statelog `runtimeError`, and the rebuilt failure drops the `interruptRejected` marker and the checkpoint.

Suggest mirroring the template clause into that catch ahead of the `AgencyAbort` rung, returning its own shape: `{ messages: __threads(), data: __error.refusal }`. Both test suites route through `def`s, so nothing covers the inline-in-a-node shape today.

packages/agency-lang/lib/templates/backends/typescriptGenerator/functionCatchFailure.mustache
```diff
@@ -8,6 +8,13 @@ if (__error instanceof RestoreSignal) {
 // any abort to a Failure here would (a) hide a guard trip so the block appears
 // to succeed over budget, and (b) let a cancel limp onward / surface as a
 // logged ERROR the REPL can't recognize. See lib/runtime/errors.ts (§5).
+// A refused validator interrupt unwinding out of a compiled type test.
+// A human answering "no" is a decision, not a crash: return the refusal
+// failure it carries (marker included) with NO error log — it must not
+// read as runtimeError in statelog. See InterruptRejectedError.
+if (__error instanceof InterruptRejectedError) {
+  return __error.refusal;
+}
```
egonSchiele: Returning the refusal untouched skips the destructive-boundary stamp that every other failure leaving this function gets (`stampFailureBoundary(runner.haltResult, __self.__destructiveRan)` on the halt path; `destructiveRan: __self.__destructiveRan` on the normal catch return below). So a function that ran destructive work and then hit a refused type test returns a failure carrying the validator frame's flag, not this frame's — and that marker is authoritative for retry gating.

`return stampFailureBoundary(__error.refusal, __self.__destructiveRan)` unless the untouched pass-through is deliberate, in which case say so here.

packages/agency-lang/lib/runtime/validateChain.ts
```diff
@@ -20,6 +21,24 @@ export type AgencyValidator =
   | ((value: unknown) => Promise<ResultValue> | ResultValue)
   | AgencyFunction;
 
+/** Boolean type test for compiled type patterns and `is Type`. A failure the
+ *  validator RETURNED means "no match"; a refused interrupt must not become
+ *  dispatch — a human answering "no" is not "the type did not match". */
+export async function __typeTest(result: ResultValue): Promise<boolean> {
```
egonSchiele: The parameter is typed as a settled `ResultValue` but the body `await`s it. Correct today — `validateExpr` emits an already-awaited chain call and `__validateType` is synchronous — but the signature says the promise case is impossible while the code handles it. Widen the type to `ResultValue | Promise<ResultValue>`, or drop the await and add the one-line reason it is unnecessary.

egonSchiele (commented):
Reviewed against the branch, with probes rather than reading. The implementation is careful and matches what the spec and plan reviews asked for: `runValidator` splits calling from enforcing, `Object.is` with the NaN reasoning written down, the marker stamped in BOTH interrupt templates (the second one was easy to miss), `(anonymous)` reusing the runtime's existing spelling, and the unwrap-with-no-crash-log so a refusal does not read as `runtimeError`. The acceptance rows assert the message and the validator name rather than just "some failure happened," which is what makes them worth having.

One real gap, verified on this branch.

**A refusal inside a node body still crashes.**

The `InterruptRejectedError` clause was added to `functionCatchFailure.mustache`, but the graph-node catch is generated separately (`typescriptBuilder.ts:3216`) and does not know about it. A type test written directly in a `node` body — no `def` in between — hits that catch instead. Run on this branch with `--reject app::confirm`:

```
ERROR Node main crashed: interrupt rejected
ERROR InterruptRejectedError: interrupt rejected
    at __typeTest (dist/lib/runtime/validateChain.js:10:15)
    at async Object.condition (nodeprobe.js:377:34)
    at async Runner.ifElse (dist/lib/runtime/runner.js:800:17)
    …
```

Program:

```agency
node main() {
  const out = match(7) {
    a: Checked => "checked:${a}"
    _ => "other"
  }
  return out
}
```

That is the exact outcome the PR description says was eliminated: the crash log, the stack trace, and (per the template) a statelog `errorType: "runtimeError"`. The returned failure also loses the `interruptRejected` marker and the refusal's checkpoint, because the node catch rebuilds it with `failure(__error.message, { functionName })`.

Nothing catches it because every scenario in both suites routes through a `def` (`armChecked`, `isChecked`, `bangChecked`). The fix mirrors the template clause into the node catch, ahead of the `AgencyAbort` rung, returning the node catch's own shape: `{ messages: __threads(), data: __error.refusal }`. Worth an acceptance or spawn row with the match written inline in the node, since that is the shape a user reaches for first.

For completeness I checked the third error path: `blockSetup.mustache` rethrows anything that is not an `AgencyAbort`, so a refusal inside a guard/thread block reaches the enclosing function catch and is handled correctly. Only the node catch is missing.

**Question: should the unwrap stamp the destructive boundary?**

`return __error.refusal` hands back the failure exactly as the validator's reject halt built it. Every other failure leaving a function gets the enclosing frame's destructive flag folded in — the halt path calls `stampFailureBoundary(runner.haltResult, __self.__destructiveRan)`, and the catch path's normal return passes `destructiveRan: __self.__destructiveRan` explicitly. The new early return does neither, so a function that ran destructive work and then hit a refused type test returns a failure whose `destructiveRan` reflects the validator's frame, not this one. Given that the destructive marker is treated as authoritative for retry gating (#535), a `return stampFailureBoundary(__error.refusal, __self.__destructiveRan)` looks right — unless returning the refusal untouched is deliberate, in which case the comment should say so. Same question, smaller stakes, for `functionName`: the propagated refusal names whatever the halt named, not the function it emerged from.

**Smaller notes**

- `describe.skipIf(!existsSync(CLI))` silently disables the entire end-to-end suite when `dist/` is not built — and that suite is the only end-to-end coverage of Change 2. It matches the existing `runPolicy.spawn.test.ts` precedent so I would not change the pattern here, but it is worth confirming CI builds before this suite runs; a silent skip on the headline feature is an expensive way to find out.
- `__typeTest(result: ResultValue)` is declared as a plain value but `await`s it. That is correct today (`validateExpr` emits an already-awaited chain call, and `__validateType` is sync), and the await is what makes the schema-only path safe — but the parameter type says the promise case cannot happen while the body handles it. Either widen the type or drop the await and note why it is unnecessary.
- The `interruptRejected` marker is set only when true, so ordinary failure payloads stay byte-identical — good call, and the one agency-js fixture that deep-equals a refusal object is the right place for it to show up.

The rest reads well. Pinning the unanswered-interrupt hole as a test row rather than leaving it as prose in the PR body is the right instinct, and the two zod rows carrying comments about their double duty is exactly the kind of thing that stops a future reader from deleting them.

egonSchiele:
All four items addressed in 601aecdb9.

**The node-body gap (your main finding — confirmed and closed).** You were right that both suites routed every scenario through a `def`. The graph-node catch now has the same `InterruptRejectedError` rung ahead of `AgencyAbort`, returning the refusal in the node-catch shape — no crash log, no statelog `runtimeError`, marker and checkpoint intact. Pinned by a new spawn row with the match written inline in `node main()` (`--reject` must produce no "crashed" log and must not print the wildcard marker). The node clause does not stamp anything because nodes carry no destructive flag of their own — same as the node catch's existing failure path.

**The boundary stamp (your question — answered yes, with one deliberate exception).** The function-template unwrap is now `return stampFailureBoundary(__error.refusal, __self.__destructiveRan)`, so a function that ran destructive work and then hit a refused test returns a failure carrying this frame's flag — pinned end-to-end by a spawn row (`destructive { … }` then a refused test → `outcome.destructiveRan` is `true`). `functionName` deliberately stays the refusal's own: the refusal predates the frame, and the marker — not the name — is what consumers key on; the template comment now says so, so it reads as a choice rather than an omission.

**`__typeTest`'s parameter** widened to `ResultValue | Promise<ResultValue>` to match the await in its body (you and Copilot both flagged it; widening beats dropping the await, since the await is also what keeps the schema-only `__validateType` path safe if its emission ever changes).

**The skipIf question — checked, not assumed:** the CI build job runs `make` before `pnpm test:run` (`.github/workflows/test.yml`), so the dist-missing skip cannot silently fire there; it only skips a local clean checkout, same as the `runPolicy.spawn` precedent.

Fixture churn from this round is the node-catch clause appearing in generated node wrappers (98 files, mechanical). Full unit suite 9847 passed; the spawn suite is 8 rows, all green.

packages/agency-lang/docs/site/guide/type-validation.md
```diff
@@ -31,18 +31,33 @@ def isPositive(value: number): Result<number> {
 }
 ```
 
-If successful, the function returns `success` with the value. You can also send a new value to modify the value. So for example, instead of having isPositive fail, we could just have it clamp the value to be above zero.
+If successful, the function returns `success` with the value it was given. Validators are **predicates**: they accept or reject, and they may not substitute a different value. A validator that returns a changed value is a runtime error naming the validator:
```
egonSchiele: Please undo all changes to this file.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -3200,6 +3207,23 @@ export class TypeScriptBuilder {
             ts.raw("__error instanceof RestoreSignal"),
             ts.statements([ts.throw("__error")]),
           ),
+          // A refused validator interrupt unwinding out of a compiled type
```
egonSchiele: This is not ideal. Everywhere else, we return the interrupt, but here we raise an exception for it. Is this the only option? You should see how the runner is implemented and how it is used to halt at the current step and return the interrupt:
lib/runtime/runner.ts

You are introducing a new pattern for handling interrupts when one exists in the code base already.

packages/agency-lang/lib/runtime/result.ts
```diff
@@ -66,6 +66,11 @@ export type FailureOpts = {
   destructiveRan?: boolean;
   functionName?: string;
   args?: Record<string, any>;
+  /** This failure IS a refused interrupt — a human (or policy) answered no.
```
egonSchiele: again, this is a new pattern, you're changing a core type that is used everywhere, in order to create a new way to do something, when there's already an established way.

egonSchiele:
Split done per discussion: this PR now carries ONLY the predicate contract. The refusal-propagation mechanism is fully reverted (the error class, both catch rungs, `__typeTest`, the `interruptRejected` marker and all eight template halt sites, and the two marker-shape test assertions — verified by grep that no trace remains in tracked files). Tier-2 type tests compile to `isSuccess(...)` exactly as on main.

The interrupt-in-validator behaviors are pinned as they currently are — approval works, refusal is visible through bang, refusal reads as no-match in arms/`is` (KNOWN GAP rows), unanswered never surfaces (#724) — so the follow-up PR flips documented pins rather than changing unguarded behavior. The redesigned mechanism using the existing conventions (hoisted type-test statements + the standard `hasInterrupts`/halt relay) is written up in the spec as "Change 2 (v3): the conventional mechanism" for the separate discussion.

egonSchiele (commented):
Re-reviewed at `2d2fd9c69`. The revert is clean and I verified the result rather than reading it — rebuilt the branch and ran everything.

**Verified green at this head**

- `validateChain.test.ts` + `validatorInterrupts.spawn.test.ts`: 33 passed.
- Acceptance file: 12/12. Migrated pins: `validateTransform` 1/1, `type-patterns` 10/10, `record-validators/main` 5/5.
- Codegen really did revert — a Tier-2 type test compiles back to `isSuccess(await __validateChainRecursive(...))`, with no `__typeTest` anywhere. Grepping `lib/` and `tests/` for `__typeTest`, `InterruptRejectedError`, and `interruptRejected` turns up nothing outside the spawn test's comments.
- The node-body crash I reported last round is gone with the mechanism that caused it: the same program now prints `RESULT:other` with no crash log and no stack trace. That is the pinned KNOWN GAP, and it matches what the spawn suite asserts. My other two inline comments were also about the reverted code and no longer apply.

Splitting the PR down to the predicate contract was the right call, and pinning the interrupt behaviors as KNOWN GAP / KNOWN HOLE rows rather than deleting them is the right way to leave a deferred feature: when Change 2 lands, those rows fail loudly and force the flip instead of quietly agreeing with whatever ships.

Two loose ends, both small.

**1. The guide file is still modified after "Please undo all changes to this file."**

`docs/site/guide/type-validation.md` still carries three hunks (the predicate contract, the explicit-repair-function idiom, the routing bullet) and the thread has no reply. Worth resolving explicitly rather than by silence, because the two readings pull opposite ways: leaving the file untouched ships a guide whose worked example (`return success(1)` to clamp) is now a runtime error and whose text still says "you can also send a new value to modify the value" — so a reader following the docs writes code this PR breaks. If the objection was to the scope of the rewrite rather than to touching the file at all, the minimum edit is probably just the two sentences that describe modification as supported. Either way: answer the comment, do not merge past it.

**2. One leftover file with a whitespace-only diff.**

`tests/agency-js/interrupts/interrupt-overrides/fixture.json` is still in the PR, and its entire remaining change is adding a trailing newline — the marker edit it existed for is gone. Drop it so the diff is only files this PR means to touch.

**Nit**

`expect(stdout).toMatch(/MARK:no\b/)` in the `is` gap row also matches `MARK:no-match`, since `\b` sits between `o` and `-`. Not a problem today (the fixture prints `MARK:no`), but the row is specifically there to be flipped later, so an anchored pattern would keep it honest through that edit.

The contract implementation itself I have no further notes on — `runValidator` splitting call from enforce, `Object.is` with the NaN reasoning stated, `(anonymous)` matching the runtime's existing spelling, and the unit rows covering per-validator placement (transform-and-back), escape out of the recursive walk, rebuilt-equal, and empty `success()` all hold up.

packages/agency-lang/docs/site/guide/type-validation.md
egonSchiele: Apologies for the silence here — I missed this thread when processing the review round (my comment fetch filtered by timestamp and this one fell outside the window; fixed my process).

Resolved in the direction your re-review suggested: the file is reverted to main, then the minimum edit only — the paragraph and example that taught "send a new value to modify the value," the modified-value outcome line in the worked example, and the chain hand-off sentence now state the predicate rule in one sentence each. Nothing else in the file is touched. The reasoning for not undoing entirely: the clamp example is a runtime error after this PR, so a fully-reverted guide would teach readers code that breaks. If you would rather the file go fully untouched and take the doc fix separately, say the word and I will revert the rest.

egonSchiele:
All three loose ends closed in a47f93ea9:

1. **The guide thread is answered** (with an apology — I had missed it: my inline-comment fetch filtered by timestamp and it fell outside the window). Resolution per your re-review's reading: file reverted to main, then the minimum edit only — the two passages teaching modification as supported and the modified-value example line now state the predicate rule; everything else untouched. Replied on the thread; if you would rather the file go fully untouched with the doc fix as a separate change, happy to revert the rest.
2. **The fixture whitespace diff is gone** — `interrupt-overrides/fixture.json` is byte-identical to main and out of the PR.
3. **The gap-row pattern is anchored** — `/^MARK:no$/m`, so it cannot match `MARK:no-match` when the row is flipped by the follow-up PR.

Spawn suite 6/6 after the anchor change.

PR #726, opened 2026-07-29
Remove the pairwise optimize loop; default eval output to the node return value

egonSchiele:
This is the first of three changes clearing the way to make `agency eval` a real eval framework. It removes a duplicate optimizer and fixes what eval considers "the agent output". A follow-up will give eval access to the graders that until now only `optimize` could use.

## Optimize had two implementations

Both discovered `optimize`-marked declarations, asked a model to rewrite them, ran the agent, and kept the better version. Everything after that differed:

|  | `optimizeLoop` (`loop.ts`) | `BaseOptimizer` + `optimizers/` |
|---|---|---|
| Reached by | `std::agency`'s `optimize()` only | `agency optimize` only |
| How it decides | Pairwise LLM judge: champion vs candidate, accept iff the judge picks the candidate | Graders produce a 0-1 score, accept iff the score improves |
| Graders | none — there is no grader concept in that file | exact-match, contains, similarity, LLM judge, custom functions |
| What it grades | the last LLM completion | the node's return value |
| Search | one champion, one candidate per iteration, always runs all iterations | greedy hill-climb or GEPA's Pareto pool with minibatch pre-filtering, both stop early at a perfect score |
| Held-out validation | none | `validationInputs` + `pickValidationChampion` |
| Feedback to the mutator | history only | history plus per-input output, expected answer, and each grader's score and notes |
| Writeback | its own sha-checked implementation | a second sha-checked implementation |

Three of those are worth calling out.

**They did not grade the same thing.** `optimizeLoop` ran candidates through `evalRunLoadedInputs` without overriding the extractor, so it got the default `outputFallback: "llm"` and its judge read the agent's last LLM reply. The pointwise path overrode the extractor to `"returnValue"`. Same declarations, same inputs, two different notions of "the output".

**Only one produced a number.** The pairwise loop's result is "the candidate beat the champion 4-1". You cannot compare that to last week's run, and the loop itself cannot tell a small improvement from a large one.

**Only one could be steered.** Reflection feedback is what lets the optimizer learn "the output gave the area, not the capital". The pairwise loop passed the mutator a history string and nothing else.

So `optimizeLoop` goes, along with `artifacts.ts`, `history.ts`, `inputs.ts`, the second reporter, the second writeback, and `OptimizeLoopConfig`.

## `std::agency`'s `optimize()` is removed, not retargeted

It was `optimizeLoop`'s only caller. Rather than quietly swap its acceptance rule, its notion of output, and three of its parameters, this removes it. We want to rethink what shape optimize should expose to Agency code before exposing it again.

`agency eval judge` and all of `lib/eval/judge/` are untouched, so comparing two finished runs still works exactly as before. The stdlib keeps `evalRun`, `evalExtract`, `evalJudge`, and `evalJudgeSuite`.

## Also removed

- `--samples`, `--confidence-threshold`, `--margin-threshold` on `agency optimize`. They only ever configured the pairwise judge policy, and `EvalOptimizeOptions` never had fields for them, so they were already being dropped before reaching the optimizer.
- `IterationResult.winsA` / `winsB` / `ties`, which the pointwise path has always hardcoded to zero.

## Eval output now defaults to the node's return value

The return value is what the language already calls the output of a node: it is what `agency run` prints and what a TypeScript caller receives. Requiring `evalOutput()` to get it was asking users to annotate their way back to the obvious.

One wrinkle worth reviewing. `returnValue` yields *nothing* when a node returns nothing, so a chat-style agent that prints its answer would have gone from "here is the reply" to "no output at all", including for `agency eval judge`. There are now three modes:

- **`preferReturnValue`** (the new default) — return value, dropping to the last LLM completion with a warning when the node returned nothing
- **`returnValue`** — strict, no fallback. The optimizer uses this so "the node returned nothing" surfaces as a clear error instead of silently optimizing chat replies
- **`llm`** — the previous behavior, still available

Both commands now extract the same thing by default, which was the divergence worth killing.

## The optimizer guide was describing an API that no longer exists

`docs/dev/writing-optimizers.md` documented `fork(dir)` taking a directory, a `workspace.applyFiles` method that was deleted, and a three-argument `evaluate`. It told authors to set `result.trainObjective` and call `writeBack` themselves, both of which `finishPointwise` has done for a while — and `finishPointwise` was not mentioned at all, despite being the method every built-in optimizer ends with. The worked example was a stale copy of `example.ts`.

Rewritten against the current code, plus what was missing: how a candidate file map becomes an isolated run, what `Workspace` actually is now (a cache-partition token, not a directory), the full `Scorecard` surface including `gatedObjective`, and the `deps` seams for testing. Reviewers may find it the most useful file in the diff for orientation.

## Testing

`pnpm run typecheck`, `pnpm run lint:structure`, and the full `lib/eval` + `lib/optimize` + `lib/cli/eval` + stdlib suites all pass. One existing test asserted the old LLM-completion default; it now asserts the return-value default, and three cases were added for the fallback chain and the strict mode.

Net: **2,362 deletions, 96 insertions** across the code, plus the doc rewrite.


packages/agency-lang/lib/eval/extract.ts
```diff
@@ -22,16 +22,22 @@ import type {
...
    *  Pass 0 for "no truncation". */
   previewChars?: number;
   /** How to infer the agent output when `evalOutput()` was not called:
-   *   - "llm" (default): the last LLM completion on the top-level thread, with a warning.
-   *   - "returnValue": the entry node's return value (from `agentEnd`), with no warning.
-   *  Explicit `evalOutput()` calls always take precedence. The optimizer uses
-   *  "returnValue" because it grades what the node returns, not its last LLM reply. */
-  outputFallback?: "llm" | "returnValue";
+   *   - "preferReturnValue" (default): the entry node's return value, dropping to
+   *     the last LLM completion (with a warning) when the node returned nothing.
```
egonSchiele: lets remove "last LLM completion" as an option entirely.

packages/agency-lang/lib/eval/extract.ts
```diff
@@ -22,16 +22,22 @@ import type {
   ThreadEntry,
 } from "./types.js";
 
+/** How the agent's output is inferred when `evalOutput()` was not called. */
+export type OutputFallback = "llm" | "returnValue" | "preferReturnValue";
```
egonSchiele: not needed, we will always take the return value from the node, and print and error if nothing was returned, and no other value was specified using `outputValue`

packages/agency-lang/lib/eval/extract.ts
```diff
@@ -368,14 +374,16 @@ function heuristicValues(prompts: NormalizedEnvelope[], warn: boolean): WithWarn
...
 function fallbackOutputs(
   n: Normalized,
   prompts: NormalizedEnvelope[],
-  mode: "llm" | "returnValue",
+  mode: OutputFallback,
 ): WithWarnings<EvalValue[]> {
-  return mode === "returnValue" ? returnValueOutputs(n) : heuristicOutputs(prompts);
+  if (mode === "llm") return heuristicOutputs(prompts);
+  const returned = returnValueOutputs(n);
+  if (mode === "returnValue" || returned.result.length > 0) return returned;
```
egonSchiele: An explicit `null` return is treated as "the node returned something": `returnValueOutputs` only reports empty when `agentEnd.data.result === undefined`, so a node whose last expression is an unmatched `match` or a missing-key index (which the language normalizes to `null`) gets graded on the string `"null"` with no warning.

Intended semantics: returning `null` means "this node has no return-value output — the output was recorded via `evalOutput()`". So when the node returns `null` and `evalOutput()` was never called, that should be an error, not a silent grade of `"null"` (and not a silent drop to the last LLM completion either).

The new tests cover the missing-`result`-key case but not `result: null` — worth a test pinning whichever semantics lands here.

egonSchiele (commented):
Reviewed the full diff. The deletion itself is clean: every removed symbol (`optimizeLoop`, `createOptimizeArtifacts`, `buildMutationHistory`, `normalizeOptimizeInputs`, `createOptimizeReporter`, `OptimizeLoopConfig`, `_optimize`, the stdlib `optimize` and its types) has zero surviving references, and typecheck passes on the branch. `selectFinalResponse` and `copyProjectTree` are still used by the kept judge/eval code, and relative `working_dir` resolution is already handled by `loadInputs.ts`, so nothing the deleted `inputs.ts` did is lost. The rewritten `writing-optimizers.md` checks out against the code: every signature, the retry default of 3, and both `TODO(gated-objective)` markers are accurate.

A few things to fix (one inline comment plus the items below):

**Stale references in `docs/site/cli/optimize.md`** (not touched by this PR):
- The Notes section at the bottom still says "The stdlib `agency.eval.optimize(...)` function does **not** install a handler; Agency callers should wrap it in their own handler" — that function no longer exists after this PR. Delete the sentence.
- The `Input` type doc comment (~line 145) says the goal is "read by the goal judge and the **pairwise judge suite**" — the pairwise suite is no longer part of optimize.

**`lib/optimize/optimizers/example.ts` docstring drift**: the class docstring still describes the shape as "fork → apply → evaluate → compare → report → return" while the rewritten guide renames it "score → propose → score → compare → finish". Since the guide says "copy `example.ts`", the two should agree.

PR #727, opened 2026-07-30
Graders for agency eval: move grading into the eval layer

egonSchiele:
Implements `/Users/adityabhargava/agency-lang/docs/superpowers/specs/2026-07-29-eval-graders-design.md`. Follows #726.

## The problem

`agency eval run` could run an agent but not score it — "3/3 inputs ok" meant three processes exited without crashing, nothing more. Meanwhile `agency optimize` had a complete grading system, but it lived at `lib/optimize/grading/`, so for eval to score anything it would have had to import *upward* from optimize. That inverts the layering we want, where the optimizer sits on top of the eval framework.

So: eval could not answer the question it exists to answer, and the scoring code was filed under the wrong roof.

## What this does

**Moves grading into the eval layer.** All of `lib/optimize/grading/` plus `goalJudgeFile.ts`, `gradeBreakdown.ts`, and `gradingModule.ts` move to `lib/eval/grading/`. Pure relocation — 414 tests passed with no test-body edits, only import paths, which is the proof.

**Adds one shared scoring step.** `gradeInput(input, result, ctx)` scores one input; `gradeRun(run, ctx)` loops and aggregates. Three callers: `eval run` after running, `eval grade` over a finished directory, and `BaseOptimizer.evaluate` per input.

**Two commands:**

```bash
# run and score in one go
agency eval run --agent a.agency --inputs in.json --graders g.ts
#   3/3 inputs ok
#   objective  0.71
#     exact-match  2/3 pass
#     concise      0.85

# iterate on the grader against fixed outputs, no agent runs
agency eval grade runs/abc --graders g2.ts
```

With no `--graders`, both default to the bundled goal judge — the same default `agency optimize` already uses. `--no-grade` skips scoring for when you only want to know the agent ran.

**Graders can see more.** The context grows from `{ output, input, judge }` to `{ output, input, workdir, record, judge }`, so a file-writing agent is gradable and so is what the agent did on the way there:

```ts
import { grader, ExactMatch } from "agency-lang/eval";

export default [
  grader(({ output }) => String(output).length < 500, { name: "concise" }),
  grader(({ workdir }) => existsSync(join(workdir, "analyze.py")), { name: "wrote-script" }),
  grader(({ record }) => record.metrics.costUsdTotal < 0.05, { name: "cheap" }),
  new ExactMatch({ mustPass: true }),
];
```

Both new fields were already on hand: `record` is parsed from the eval record grading already reads, and `workdir` was sitting unused on `EvalRunInputResult`. There is deliberately **no** curated trace API (`trace.calledTool(...)`) — the raw record makes trace-based grading possible now, and the vocabulary can wait for real usage.

**`agency-lang/eval`** is the new import path for grader authors. `agency-lang/optimize` re-exports every grading name, so existing grading modules keep working with no edit.

## Three decisions worth reviewing

**The default lives at the command layer, not in the library.** `evalRunLoadedInputs` grades only when a caller hands it graders. The optimizer calls that function once per input per candidate, so defaulting grading on there would fire a discarded goal-judge LLM call every single time — and would make every existing test in `run.test.ts` spawn a real judge subprocess.

**Reporting uses `objective()`, not `gatedObjective()`.** The latter collapses to 0 when *any* input gate-fails, so one flaky timeout out of fifty would report 0.00 for the whole run. A gate-failed input already contributes 0 to the mean; `gatesPassed` drives the exit code instead. `gatedObjective()` stays what optimizers compare candidates on.

**An input that produced nothing scores 0 instead of throwing.** `gradedOutput` used to throw. Eval needs a suite of mixed results, not an abort. The optimizer loses nothing — `requireBaselineGatesPass` still refuses a broken baseline, with a clearer message than a stack trace.

## Internal changes

`AgentRun` grows to `{ output, recordPath, workdir, record }`, and reading the eval record moves *out* of the optimizer's run step into grading. So the `RunInput` seam and `EvalCache` now carry `EvalRunInputResult`. Caching policy is unchanged: still one run per `(workspace, input)`, still runs rather than grades.

Every optimizer test that injects `runInput` had to change shape and now writes a real eval record via a `fakeRun` helper. `docs/dev/writing-optimizers.md` is updated to match.

Two simplifications fell out: `reflectionFeedback` no longer re-reads the record from disk (it rides on the run), and its tests no longer need temp files.

## Not in scope

Deliberately excluded, each noted in the spec: anything about workdir copying or retention, the GEPA parent-scorecard filter, the optimizer's per-iteration artifact tree, a curated trace API, and `agency eval judge` (untouched — pairwise comparison answers a different question). The Agency-side `std::agency/eval.evalRun` also stays ungraded, since what optimize and eval expose to Agency code is being reconsidered separately.

## Testing

428 tests pass across `lib/eval`, `lib/optimize`, and `lib/cli/eval`, up from a 414 baseline. Typecheck and the structural linter are clean. 14 new tests cover the grader context, gate short-circuiting, no-output and errored inputs, the path-vs-in-memory equivalence, the counted-not-zeroed objective, and `eval grade` leaving `summary.json` byte-identical.

One test I want to call out because it caught a real bug during planning: `gradeRun` on an in-memory result reads each input's spec from the `input.json` on disk. An earlier draft synthesized a bare `{ id, args: {} }`, which would have silently starved `LlmJudge` of its `goal` and `ExactMatch` of its `expected` — neither throws on a missing field, so it would have scored against nothing.


egonSchiele (commented):
Reviewed the full diff against the spec, the plan, and both review docs. This lands well: all three of the plan review's significant findings are fixed the way the review suggested — the goal-judge default lives in `resolveGraders` at the command layer (with a test pinning that the optimizer path stays ungraded), the reported objective is `objective()` with the counted-not-zeroed test, and the branch waited for #726. The smaller findings made it too: `readEvalRun`'s `missing` status gets its own reason with a test, `workdirFor` uses `path.join`, the print block is per-grader, and the in-memory input-spec read that the self-review caught is tested. The move itself is clean — the renames are pure, and I verified the `agency-lang/optimize` re-export surface is name-for-name what it exported before, now routed through `@/eval/public.js`.

Two inline comments on robustness of the new grading path (corrupt-file handling). Beyond those, three observations, none blocking:

- **Per-grader summaries are computed only over the inputs a grader actually graded.** Because gates short-circuit, an advisory grader like `concise` has no rows for gate-failed inputs, so `concise 0.9` can look healthy while half the suite gate-failed. The `ungradedInputs` lines and the per-input breakdown in `summary.json` mitigate this, but a future reader of just the terminal block could be misled. Worth keeping in mind if the block grows a `graded n/m` annotation later.
- **`--no-grade` still requires `goal` on every input** (`requireGoal: gradersPath === undefined` doesn't consider `grade`). That preserves the pre-PR behavior of plain `eval run`, so it's defensible — but someone using `--no-grade` for a goal-less smoke suite will hit a confusing "goal is required" error for a run that grades nothing. Fine to leave; noting it as a known edge.
- Plain `agency eval run` now makes one goal-judge LLM call per input where it previously made none. That's the agreed default and the docs say so — just flagging it as the one behavior change existing users will feel.

packages/agency-lang/lib/eval/grading/gradeRun.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+  return globalThis.JSON.parse(fs.readFileSync(file, "utf8")) as Input;
+}
+
+/** Read the record and pull out the graded output. Null when there is none. */
+function buildAgentRun(result: EvalRunInputResult): AgentRun | null {
+  if (!result.evalRecordPath || !fs.existsSync(result.evalRecordPath)) {
+    return null;
+  }
+  const record = globalThis.JSON.parse(fs.readFileSync(result.evalRecordPath, "utf8")) as EvalRecord;
```
egonSchiele: A corrupt eval record crashes the whole grading pass instead of scoring this one input 0. A *missing* record is handled gracefully (`null` → ungraded with a reason), but a truncated or half-written `eval-record.json` makes this `JSON.parse` throw, the rejection propagates through `Promise.all` in `gradeRun`, and — in the `eval run` inline path — the exception arrives *after* every agent has already run and been paid for, taking down the whole result. `eval grade` similarly dies with a raw stack on one bad file.

Suggest wrapping the read+parse and returning `null`-with-reason (e.g. `ungradedReason: "eval record unreadable: <message>"`) so a bad file degrades the same way a missing one does. `readInputSpec` above has the same edge with a corrupt `input.json`, though the blast radius there is just a degraded input spec.

(Worth a test: write garbage into `eval-record.json` and assert the input scores 0 with a reason rather than the run throwing — the mirror of the existing "distinguishes a lost eval record" case.)

packages/agency-lang/lib/cli/eval/grade.ts
```diff
@@ -0,0 +1,59 @@
...
+  opts: EvalGradeOptions,
+): Promise<EvalRunGrading> {
+  const config = opts.config ?? {};
+  const resolvedRunDir = path.resolve(runDir);
+  const gradersPath = opts.graders ?? config.eval?.graders;
+  // `grade` is undefined here: there is no point running this command with
+  // grading switched off, so the same resolver's default path applies.
+  const graders = await resolveGraders(gradersPath, undefined, config);
+  if (!graders || graders.length === 0) {
+    throw new Error("No graders resolved; pass --graders <file> or set eval.graders in agency.json.");
```
egonSchiele: This guard is nearly unreachable, and when it *does* fire the message points the wrong way. `resolveGraders` only returns `undefined` when `grade === false` (never passed here), and otherwise falls back to the goal judge — so the only way to get here is a grading module that default-exports an empty array. In that case the user *did* pass `--graders`, and "pass --graders <file>" tells them to do what they already did.

Either drop the guard (an empty module then yields an objective over zero graders — arguably also confusing) or reword it for the real case: "the grading module at <path> exported no graders".

Issue #728, opened 2026-07-30
Top-level interrupt compiles then crashes with '__self is not defined'


egonSchiele:
An `interrupt(...)` written at the top level of a file compiles, then crashes at run time with a JavaScript error from generated code:

```ts
// i.agency
interrupt("x")

node main() { print(2) }
```

```
$ agency run i.agency
Agent crashed: __self is not defined
  const __response = getRuntimeContext().ctx.getInterruptResponse(__self.__interruptId_);
```

The `static` spelling behaves identically:

```ts
static interrupt("x")
```

```
Agent crashed: __self is not defined
```

## Why this is worth a bug rather than a shrug

`staticStatementParser` (`lib/parsers/parsers.ts:4706`) deliberately enumerates what a bare `static <expression>` may be:

```ts
const innerParser = or(
  interruptStatementParser,
  functionCallParser,
  valueAccessParser,
);
```

Interrupts are in that list on purpose, and the surrounding comment describes routing a bare statement into the init phase so it runs once per process. So the parser records a design intent the backend never implemented.

That fits what is already known about interrupts needing a running node: there is no node at initialization, and the generated code reaches for a `__self` that does not exist there.

## Current status

PR for #713 turns both spellings into a compile-time refusal (`AG3017`), on the grounds that a diagnostic naming the rule is a better experience than `__self is not defined` from generated JavaScript, and that no working program breaks because there are none. The predicate that decides this is `isLegalAtTopLevel` in `lib/utils/topLevel.ts`, where `interruptStatement: false` carries a comment pointing at this issue.

So the crash is no longer reachable — but the underlying question is still open, and it is a design question rather than a bug fix:

**Should a top-level interrupt work at all?** Two defensible answers.

- **No.** Interrupts need a node to suspend and resume; initialization has neither. In that case the current refusal is the final answer, and `interruptStatementParser` should come out of `staticStatementParser`'s inner list so the parser stops advertising something that cannot work.
- **Yes, for the `static` form.** If there is a use case for asking the operator something once per process before any node runs, then the init phase needs whatever `__self` stands in for, and `isLegalAtTopLevel` flips one row.

Either way the parser and the backend should agree, which they do not today.

## Notes for whoever picks it up

- `lib/utils/topLevel.test.ts` pins both spellings as refused; closing this the second way means changing that test on purpose.
- The generated line comes from the interrupt codegen path reading `__self.__interruptId_`; the init phases have no `__self` binding, which is the direct cause.

[...]

Issue #229, opened 2026-05-30
Compiler crash: bare top-level call with `with` modifier — `StepPathTracker: currentId() called with empty path`


egonSchiele:
## Summary

A bare top-level function call combined with a `with` modifier crashes the TypeScript code generator with `StepPathTracker: currentId() called with empty path`. The same call wrapped in `static const _ = ...` works correctly.

## Reproduction

```agency
import { enableMemory } from "std::memory"

enableMemory({ dir: "/tmp/agency-mem-test" }) with approve

node main() {
  print("hello")
}
```

`pnpm run agency typecheck …` reports **No type errors found.** The file parses and typechecks. But `pnpm run agency …` crashes during compilation:

```
Error: StepPathTracker: currentId() called with empty path
    at StepPathTracker.currentId (.../stepPathTracker.js:38:19)
    at TypeScriptBuilder.processWithModifier (.../typescriptBuilder.js:2307:31)
    at TypeScriptBuilder.processNode (.../typescriptBuilder.js:384:29)
    at TypeScriptBuilder.processNodeInGlobalInit (.../typescriptBuilder.js:2259:25)
    at Object.processNodeInGlobalInit (.../sectionAssembler.js:60:44)
    at TypeScriptBuilder.build (.../typescriptBuilder.js:263:27)
    …
```

## Workaround

Wrap the call so it sits in a step-bearing context. The pattern documented in `stdlib/memory.agency` works:

```agency
static const _ = enableMemory({ dir: "/tmp/agency-mem-test" }) with approve
```

The `static const` initializer is a step, so the `with` modifier has a step path to attach to.

## Expected behavior

Either:

1. **Allow the bare form.** Generate a synthetic step path for top-level statements so `with` works there too. The user wrote a single, unambiguous side-effecting call — needing `static const _ = …` to make it run is an awkward workaround.
2. **Reject at parse/typecheck time.** If top-level `with` is genuinely unsupported, fail at compile time with a clear message (e.g. *"`with` is not allowed at top level; wrap the call in `static const _ = …`"*) instead of crashing in the TypeScript builder with an internal-invariant error.

Either fix would be an improvement over the current "parses, typechecks, then crashes the compiler" behavior.

## Context

[...]

PR #729, opened 2026-07-30
fix: one rule for what may sit at the top level of a file

### LINKED ISSUE #713: Top-level if/while/for/match/thread crash the builder instead of erroring cleanly

egonSchiele:
Three symptoms, one root: nothing in the compiler answered "may this statement sit at the top level of a file?", so the grammar, the backend and the splice checker each guessed differently.

## What was broken

**An `if` at file scope crashed the compiler** (#713):

```ts
if (true) { print(1) }

node main() { print(2) }
```

```
Error: StepPathTracker: currentId() called with empty path
    at TypeScriptBuilder.processIfElseWithSteps
    at TypeScriptBuilder.processNodeInGlobalInit
```

No file, no line, no column. Same for `while`, `for`, `match`, `thread`, a `handle` block, and `debugger("x")`.

**A splice on its own line inside a node body could not use a generator that returns statements:**

```ts
node main() {
  $( callFunc() )    // AG8007: needs an expr fragment
}
```

**The same splice at the top level was accepted** and produced TypeScript that esbuild could not parse.

Those last two are the same table, wrong in both directions: statements were refused where they are useful and accepted where they break.

## The rule

Top-level code is **initialization, not execution**. It runs in the init phases, which have no step machinery. A top-level statement may *establish* something — bind a name, call something for its effect — but may not *control* anything.

`staticStatementParser` already encoded a piece of this for `static <expression>`, enumerating an interrupt, a call, or a value access. This generalizes that answer.

## The fix

`isLegalAtTopLevel` (`lib/utils/topLevel.ts`) is one table keyed by node type, so **adding a node kind to the language fails to compile until someone classifies it**. Same enforcement `identifierSlots.ts` uses, and for a sharper reason: the bug being fixed is a statement form nobody considered reaching the backend and crashing it.

Two consumers with different jobs:

- **The type checker** reports `AG3017`, or `AG3018` for a handler, with a file, line and column.
- **A pre-pass in `TypeScriptBuilder.build()`**, beside the unfilled-holes refusal, guarantees no codegen path can reach the crash whatever else changes. It has its own test that calls the builder directly, because every compile path runs the type checker first and would otherwise never execute it.

Then the splice checker calls the same predicate, so generated code is held to the standard written code is held to — by construction, not by two lists agreeing.

## Splices

A splice position was `"decl" | "expr"`, and a splice occupying a whole statement inherited `"expr"`. Adding `"statement"` maps it to `KINDS_FOR_SORT.statements`, the set a statements *hole* already accepts. The kind-mismatch message is now built by exhaustive switch, so a future position cannot be silently described as an expression.

At the top level the check stopped asking about the fragment's kind alone. **Kind cannot answer the question** — one fragment holds both of these at once:

```ts
const apiKey = getKey()      // legal at the top level
if (debug) { log("hi") }     // crashes the backend
```

So every node of the fragment goes through the predicate. That also closes a hole a statements-only check would leave: a fragment holding an `if` and a `def` infers kind `program`, because the statements attempt fails on the `def`, and it sailed past the kind gate unexamined.

## Errors now

```
AG3017: An `if` statement cannot sit at the top level of a file. Top-level code
runs at initialization, which cannot branch, loop, or wait — move it inside a
node or a function.

AG3018: A handler cannot be registered at the top level of a file. Handlers must
be inside a node or a function, where there is execution for them to guard.

AG8014: The generator `gen` returned an `if` statement, which cannot sit at the
top level of a file.
```

Handlers get their own code deliberately: a top-level `handle` is a handler its author believes is registered, and the worst outcome of a careless fix here is not a crash but a handler that compiles and never registers.

## Two things the work turned up

**`withModifier` at top level was already made to work.** Issue #229 hit this exact `StepPathTracker` crash for `foo() with approve` at module scope and fixed it by special-casing the node in `partitionProgram` — the other way to answer this question. It stays legal, and the wider test suite is what caught my having refused it.

**Bare `debugger` and `debugger("x")` land on opposite sides.** The bare word parses as an ordinary name and compiles; the parenthesized form is a real `debuggerStatement` and crashed. Both are now classified explicitly.

## Breaking changes

Refusing code that compiles today, all of it meaningless or already broken:

- `return` at file scope — a `return` outside any function.
- `interrupt(...)` and `static interrupt(...)` — both compile today and then crash at run time with `__self is not defined`. A compile-time refusal is a better error, not a fix; the underlying gap is **#728**, where `staticStatementParser` advertises something the backend never implemented.
- `debugger("x")` at file scope, which crashed.

`goto other` is **not** affected, contrary to what the plan first assumed: at file scope it parses as two bare names rather than a `gotoStatement`, so it keeps compiling. Pinned by a test, because it looks like a refusal and is not one.

## Testing

- 9885 unit tests pass; 16/16 template fixtures; `lint:structure` clean; `make` exits 0.
- A new end-to-end fixture runs the original report's shape: a generator returning statements, spliced into a node body, compiled and run.
- The predicate's test reads as the rule, and one test ties the placement set (`TOP_LEVEL_DECLARATION_TYPES`, now exported) to the predicate, so emission can never route a node the rule would refuse.

One test-maintenance note: several type-checker fixtures place control flow directly in `program.nodes` to exercise inference, and count errors exactly. Those now filter the top-level diagnostic through a documented helper rather than pretending a bare top-level `if` is a real program.

Closes #713.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -3,6 +3,7 @@ import { declaredName } from "../types/hole.js";
 import { printCodeLiteralBody } from "./agencyGenerator.js";
 import { walkNodesArray } from "../utils/node.js";
 import { holeNames } from "../utils/holes.js";
+import { isLegalAtTopLevel } from "../utils/topLevel.js";
 import { DIAGNOSTICS, renderMessage } from "../typeChecker/diagnostics.js";
 import {
   AgencyComment,
@@ -509,6 +510,21 @@ export class TypeScriptBuilder {
         names: unfilled.map((name) => `#${name}`).join(", "),
       });
       throw new Error(`${DIAGNOSTICS.unfilledHoles.code}: ${rendered}`);
+    }
+
+    // Same reasoning as the refusal above: refuse before generating
+    // anything. A control-flow node routed into __initializeGlobals throws
+    // from inside the step machinery, with no location to report. The type
+    // checker reports this properly (AG3017); this is the guarantee that no
+    // codegen path can reach the crash without it.
+    const illegal = program.nodes.filter((node) => !isLegalAtTopLevel(node));
+    if (illegal.length > 0) {
+      const message = renderMessage(DIAGNOSTICS.topLevelStatementNotAllowed.message, {
+        kind: illegal.map((node) => node.type).join(", "),
+      });
+      throw new Error(
+        `${DIAGNOSTICS.topLevelStatementNotAllowed.code}: ${message}`,
+      );
     }
 
     // A tripwire. Expansion removes every splice before codegen, so one
--- a/packages/agency-lang/lib/preprocessors/expandSplices.ts
+++ b/packages/agency-lang/lib/preprocessors/expandSplices.ts
@@ -462,8 +462,36 @@ function importedNamesIn(nodes: AgencyNode[]): string[] {
  * Requiring `program` would make generated constants impossible. At the
  * top level a statement is a declaration, so this costs nothing.
  */
+/** Exhaustive, not a ternary: a third position value silently described a
+ *  statement mismatch as "expression position needs an expr fragment". */
+function expectedKindFor(position: Splice["position"]): string {
+  switch (position) {
+    case "decl":
+      return "program";
+    case "statement":
+      return "statements";
+    case "expr":
+      return "expr";
+  }
+}
+
+function positionLabel(position: Splice["position"]): string {
+  switch (position) {
+    case "decl":
+      return "declaration";
+    case "statement":
+      return "statement";
+    case "expr":
+      return "expression";
+  }
+}
+
 const KINDS_FOR_POSITION: Record<Splice["position"], string[]> = {
   decl: [...KINDS_FOR_SORT.decl, "statements"],
+  // The same set a statements HOLE accepts, right for the same reason: an
+  // expression is a legal statement, and a program grafted into a body is
+  // judged when the completed program compiles.
+  statement: KINDS_FOR_SORT.statements,
   expr: KINDS_FOR_SORT.expr,
 };
 
@@ -553,8 +581,8 @@ function graft(
         params: {
           name: generatorName,
           actual: kindOf(code),
-          expected: splice.position === "decl" ? "program" : "expr",
-          position: splice.position === "decl" ? "declaration" : "expression",
+          expected: expectedKindFor(splice.position),
+          position: positionLabel(splice.position),
         },
         loc: splice.loc ?? ORIGIN_UNKNOWN,
       },
--- a/packages/agency-lang/lib/preprocessors/expandSplices.ts
+++ b/packages/agency-lang/lib/preprocessors/expandSplices.ts
@@ -8,6 +8,7 @@ import { BUILTIN_VARIABLES } from "../config.js";
 import { PRELUDE_NAMES } from "../prelude.js";
 import { BUILTIN_FUNCTION_TYPES } from "../typeChecker/builtins.js";
 import { KINDS_FOR_SORT, stampOrigin } from "../runtime/template/origin.js";
+import { isLegalAtTopLevel } from "../utils/topLevel.js";
 import { kindOf } from "../runtime/template/code.js";
 import {
   checkGeneratorEligible,
@@ -462,6 +463,24 @@ function importedNamesIn(nodes: AgencyNode[]): string[] {
  * Requiring `program` would make generated constants impossible. At the
  * top level a statement is a declaration, so this costs nothing.
  */
+/** How a node reads in a message: `ifElse` means nothing to a user. */
+function describeNodeKind(type: string): string {
+  // Null-prototype: keyed by node type strings (house pattern).
+  const names: Record<string, string> = Object.assign(Object.create(null), {
+    ifElse: "an `if` statement",
+    whileLoop: "a `while` loop",
+    forLoop: "a `for` loop",
+    matchBlock: "a `match` block",
+    messageThread: "a `thread` block",
+    guardBlock: "a `guard` block",
+    finalizeBlock: "a `finalize` block",
+    handleBlock: "a handler",
+    returnStatement: "a `return`",
+    interruptStatement: "an interrupt",
+  });
+  return Object.hasOwn(names, type) ? names[type] : `a \`${type}\``;
+}
+
 /** Exhaustive, not a ternary: a third position value silently described a
  *  statement mismatch as "expression position needs an expr fragment". */
 function expectedKindFor(position: Splice["position"]): string {
@@ -572,6 +591,24 @@ function graft(
   const captured = checkNoCapture(splice, code, generatorName);
   if (captured !== null) {
     return { ok: false, diagnostic: captured };
+  }
+  // Declaration position: the kind gate above admits both `program` and
+  // `statements`, and neither says whether the NODES may sit at the top
+  // level. One fragment can hold `const x = 1` and an `if` at once, so this
+  // is per-node or nothing. Same predicate the compile path uses, so
+  // generated code is held to the standard written code is held to.
+  if (splice.position === "decl") {
+    const illegal = code.nodes.find((node) => !isLegalAtTopLevel(node));
+    if (illegal !== undefined) {
+      return {
+        ok: false,
+        diagnostic: {
+          diagnostic: "spliceTopLevelStatement",
+          params: { name: generatorName, kind: describeNodeKind(illegal.type) },
+          loc: splice.loc ?? ORIGIN_UNKNOWN,
+        },
+      };
+    }
   }
   if (!KINDS_FOR_POSITION[splice.position].includes(kindOf(code))) {
     return {
--- a/packages/agency-lang/lib/preprocessors/expandSplices.ts
+++ b/packages/agency-lang/lib/preprocessors/expandSplices.ts
@@ -8,7 +8,7 @@ import { BUILTIN_VARIABLES } from "../config.js";
 import { PRELUDE_NAMES } from "../prelude.js";
 import { BUILTIN_FUNCTION_TYPES } from "../typeChecker/builtins.js";
 import { KINDS_FOR_SORT, stampOrigin } from "../runtime/template/origin.js";
-import { isLegalAtTopLevel } from "../utils/topLevel.js";
+import { describeNodeKind, isLegalAtTopLevel } from "../utils/topLevel.js";
 import { kindOf } from "../runtime/template/code.js";
 import {
   checkGeneratorEligible,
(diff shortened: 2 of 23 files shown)


packages/agency-lang/lib/utils/topLevel.ts
```diff
@@ -0,0 +1,121 @@
...
+  objectPattern: false,
+  arrayPattern: false,
+  restPattern: false,
+  wildcardPattern: false,
+  resultPattern: false,
+  typePattern: false,
+
+  // Templates. A hole is refused earlier (AG8001); a code literal is a
+  // value, and a splice is replaced before this check runs.
+  hole: false,
```
egonSchiele: Verified on this branch: a template file whose top level is a statement hole now gets a false diagnostic —

```
$ agency typecheck tmpl.agency     # file: `#helper` above a node
tmpl.agency:1:1 - error AG3017: A `hole` cannot sit at the top level of a file.
Top-level code runs at initialization … move it inside a node or a function.
```

That advice is wrong for a template: top-level holes are a designed feature (declaration and statement holes filled via `fill`). Legality-at-top-level for a hole is genuinely "yes, in a template"; *unfilled-ness* is a different question that already has owners — `checkTemplateHoles` (AG8001/AG8002) in the checker, and the unfilled-holes refusal in the builder, which runs before this pre-pass anyway. Suggest `hole: true` with a comment pointing at those owners, plus a test: `typeCheckSource` on a template with a top-level hole emits no AG3017. As it stands, anyone who opens a template in the editor or runs `typecheck` over a directory containing templates gets told their template is a bug.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -511,6 +512,21 @@ export class TypeScriptBuilder {
       throw new Error(`${DIAGNOSTICS.unfilledHoles.code}: ${rendered}`);
     }
 
+    // Same reasoning as the refusal above: refuse before generating
+    // anything. A control-flow node routed into __initializeGlobals throws
+    // from inside the step machinery, with no location to report. The type
+    // checker reports this properly (AG3017); this is the guarantee that no
+    // codegen path can reach the crash without it.
+    const illegal = program.nodes.filter((node) => !isLegalAtTopLevel(node));
```
egonSchiele, line 520: Ordering nit with a diagnosis cost: this pre-pass runs before the splice tripwire below it, and the table says `splice: false` — so on a compile path that skipped `expandSplices`, a top-level splice now throws `AG3017: a [splice] cannot sit at the top level` instead of the tripwire's "a compile path skipped expandSplices" message. The first tells a user their legal `$( gen() )` is wrong; the second tells a maintainer which of the seven paths broke. No false positive on healthy paths (I verified the typecheck CLI expands before checking), but when the tripwire condition ever fires, this message will point away from the actual bug. Either run this filter after the tripwire, or exclude `splice` (and `hole`, per the other comment) here and let their dedicated owners speak.

packages/agency-lang/lib/preprocessors/expandSplices.ts
```diff
@@ -462,8 +463,54 @@ function importedNamesIn(nodes: AgencyNode[]): string[] {
  * Requiring `program` would make generated constants impossible. At the
  * top level a statement is a declaration, so this costs nothing.
  */
+/** How a node reads in a message: `ifElse` means nothing to a user. */
+function describeNodeKind(type: string): string {
```
egonSchiele: Second hand-maintained display-name list — `describeKind` in `topLevelStatements.ts` has the same job with capitalized entries (message-initial) while this one is lowercase (mid-sentence). Both fail ugly rather than wrong, so this is not blocking, but they will drift: the next node type added gets a friendly name in one message and a raw `nodeType` in the other. One shared helper taking the type plus a capitalization flag (or emitting lowercase with the checker call-site capitalizing) keeps the two messages in step. At minimum, a cross-reference comment in each so the next editor knows the twin exists.

egonSchiele (commented):
This lands the whole design well: the exhaustive table, both consumers on one predicate, the direct-`build()` test that genuinely bypasses `typeCheck` (the guarantee finally has coverage that would notice its deletion), the real `TOP_LEVEL_DECLARATION_TYPES` iterated instead of a copy, `AG3018` split out for handlers with the static-unwrap in the message path, and the exhaustive position switches. The two measured discoveries are the kind of thing reviews exist to force: `goto other` at file scope parses as two bare names (so the `gotoStatement` row is unobservable and the spelling keeps compiling, now pinned), and `foo() with approve` at module scope was deliberately made to work in #229 — the table records that as a decision instead of tripping over it.

One finding is a real user-facing bug, verified on this branch; the rest are small.

1. **`hole: false` produces a false AG3017 on legitimate template files** (inline at `topLevel.ts:109`). Verified: `agency typecheck` on a file whose top level is `#helper` above a node reports `AG3017: A [hole] cannot sit at the top level of a file … move it inside a node or a function` — wrong advice for a template, where top-level declaration/statement holes are a designed feature.
2. **`splice: false` lets the pre-pass shadow the splice tripwire** (inline at `typescriptBuilder.ts:520`) — a maintainer bug ("a compile path skipped expandSplices") would now report as user error. No user-facing false positive (verified the typecheck CLI expands before checking), but the message swap loses the actionable diagnosis.
3. **Two display-name lists now exist** (inline at `expandSplices.ts:467`) — `describeKind` and `describeNodeKind`, same job, different casing conventions, no cross-reference.

One concern I probed and dismissed, for the record: `withModifier: true` looked like it could smuggle `return foo() with approve` past the check, but at file scope that parses with a top-level `returnStatement` node, so the `returnStatement: false` row catches it — the `withModifier` row is only ever observed for the bare-call #229 form, which is exactly the case it exists to keep working.

egonSchiele:
All three fixed in 7e50a6fd8. Findings 1 and 2 turned out to be one mistake, which is worth naming: I classified two *designed top-level features* as illegal.

**1. The template false positive is real.** Reproduced before changing anything:

```
$ agency tc tpl.agency          # top level is `#helper`
AG3017: A `hole` cannot sit at the top level of a file … move it inside a
node or a function.
```

A declaration hole at the top level of a template is the feature working as designed, and that advice is wrong for a template. Now `hole: true`, and the file typechecks clean — only the expected `AG4004` warning about the name the filler will introduce.

The reason it belongs to `true` rather than a special case: a program containing holes is already refused by **AG8001**, which says the right thing ("this is a template, not a program — fill it"). My row was shadowing a better mechanism with a worse message.

**2. Same for `splice`.** You are right that there is no user-facing false positive today, but the message swap is the whole cost: a splice reaching the builder means a compile path skipped `expandSplices`, and reporting that as "move it inside a node" sends a maintainer looking at user code. `splice: true` lets the existing signal through. Worth noting what that signal currently is — `default: Unhandled Agency node type: splice`, since `processNode` has no `case "splice"` the way it has one for `hole`. Generic, but it points at the right layer, which mine did not. Adding a proper internal-error case for splice would be a small follow-up if you want it.

Both rows now carry the reasoning, and both have tests: the predicate accepts a parsed top-level hole and splice, and `agency tc` on a template reports no AG3017.

**3. Two display-name lists, folded into one.** `describeNodeKind` now lives in `lib/utils/topLevel.ts` beside the rule it serves, and both consumers import it. That needed the AG3017 wording changed so the kind sits mid-sentence, which also reads better:

```
AG3017: Cannot use an `if` statement at the top level of a file. Top-level
code runs at initialization, which cannot branch, loop, or wait — move it
inside a node or a function.
```

Thanks for probing `return foo() with approve` and writing up why it is safe — that was the case I was least sure of when I flipped `withModifier` to true, and having the reasoning recorded (top-level `returnStatement` node catches it, so the `withModifier` row is only ever observed for the bare-call #229 form) is more useful than my own note saying the suite caught it.

Re-verified after the changes: 9888 unit tests pass, 16/16 template fixtures, `lint:structure` clean, `make` exits 0.

PR #734, opened 2026-07-30
eval: the run directory is the interface — runAgent/runSuite/gradeSuite/judgeSuite layering

egonSchiele:
Restructures the eval framework around one idea: **the run directory is the interface between running an agent and grading one.** Before this change, running, grading, and judging were interleaved in `lib/cli/eval/run.ts`, traded in near-identical mega-option types, and handed run data around in memory through three different shapes. Now there are three levels, each a small library with one options type, and everything above level 1 communicates through files on disk.

## The three levels

**Level 1 — run one agent, once** (`lib/eval/run/`):

```ts
const run = await runAgent(agentPath, node, args, { runDir, config, seedFiles });
// { status: "success", output, runDir, workdir } | { status: "error", errorMessage, ... }
```

`runAgent` never throws: it seeds the workdir (agent import closure + test files + optional overlay), compiles, executes in a subprocess, extracts the eval record to disk, and returns the outcome. Every failure is an error result with `error.txt` written; a crash after useful work still salvages its record to disk. The parsed record is deliberately NOT returned - evidence lives only at `<runDir>/agent/eval-record.json`.

**Level 2 — suites** (`runSuite` / `gradeSuite` / `judgeSuite`, all symmetric):

```ts
const summary  = await runSuite({ agent, inputs, perRun }, deps);   // executes, writes runs/<id>/
const grading  = await gradeSuite(runDir, graders, config);         // reads a run directory
const verdict  = await judgeSuite({ runA: dirA, runB: dirB, policy }); // reads two of them
```

`runSuite` executes only. Grading takes a finished run directory and nothing else - `eval run --grade` deliberately re-reads the artifacts it just wrote, the same call `eval grade` makes days later, so the artifacts round-trip on every graded run. `judgeSuite` reads input ids and goals from each run directory too (its in-memory `inputs` parameter and three-shape input union are gone). The rationale is written down in `docs/dev/eval-grading.md` so the re-read does not get "optimized away" later.

**Level 3 — the optimizer** consumes the same surface: its `runInput` seam returns a run directory, and each candidate is graded through the same `gradeRun` as everyone else. `lib/optimize` no longer imports from `lib/cli`, `gradeInput` is private, and the synthesized fake `EvalRunInputResult` inside grading is gone.

The `evalRun` command is now pure composition: resolve graders → load suite → validate → `runSuite` → `recordGrading`. `run.ts` and `grade.ts` contain no grading logic of their own.

## Decisions this encodes

- **An errored run scores zero, always.** Graders never see an errored run; the salvaged record is diagnostic evidence, never graded (pinned by a spy-grader test). The `no-record` case is an error too - a completed run always writes a statelog.
- **`readEvalRun` is the single place grading touches the filesystem** and owns all tolerance: a corrupt per-input file degrades that one input with a warning; a corrupt `summary.json` fails loudly naming the file; duplicate input ids warn.
- **Names:** the grader-facing type is `LoadedRun` (evidence loaded off disk), distinct from `AgentRun` (the outcome of executing). `InterruptEntry.outcome` uses the interrupt-handler vocabulary (`approved | rejected | propagated | passed`) plus `unresolved` for the absence of any recorded resolution - `passed` is only ever an affirmative claim from a resolution event, so records already on disk stay valid. `agent` → `agentLabel` on run results. `NormalizedEvent` variants are named types.
- **Deleted, no deprecation** (per the no-backward-compat ground rule): `working_dir`, `inputsSource`, the dead `EvalRunConfig`, `verbose`, the unused eval `quietCompile`, the extractor `input` field, `workspace.ts`/`runEvalInput.ts`/`projectTree.ts`, and the in-memory grading paths.

## User-visible changes

- `agency eval judge <dirA> <dirB>` no longer takes `--inputs`/`--goal` - run directories carry their own goals in each input's `input.json`. `--goal` remains for judging record files.
- Agency stdlib `evalJudgeSuite(runA, runB, ...)` lost its `inputs` argument (docs regenerated).
- `config.json`/`summary.json` key changes: `agentLabel`, no top-level `inputsSource` (provenance carries it).

## Also in here

- `lib/agentTarget.ts`: `parseTarget` + `resolveEvalRunTarget` in one home, out of the CLI.
- Eval subprocess interrupts now answer with a well-formed `IpcDecisionMessage` (previously a legacy `{ approved: true }` shape crashed every interrupting agent under `agency eval run`). **Behavior change worth knowing:** an interrupting agent under eval previously crashed, so nothing proceeded; now every interrupt is auto-approved (the agent's own handlers still run first, and a local reject stays final). Each approval is recorded in the run's statelog (`interruptResolved`, `resolvedBy: "ipc"`), and `tests/integration/eval-run/test.mjs` pins the whole path end-to-end with a real subprocess, no LLM.
- New/updated dev docs: `docs/dev/eval-grading.md` (the invariant), `writing-optimizers.md` (the `runInput` seam contract).

77 files, +1,617/-1,575. 960 tests across the eval/cli/optimize/stdlib suites, typecheck and structural lint clean.


.gitignore
```diff
@@ -69,3 +69,4 @@ tests/integration/statelog/.tmp/
 # root — transient output, never a fixture (got committed by accident once;
 # see PR #517 review).
 packages/agency-lang/a.vs.b.verdict.json
+docs/superpowers/**/*
```
egonSchiele, line 72: This does not belong in this PR, and I do not think it does what it looks like.

`docs/superpowers/**/*` stops *new* design docs from being added, but `.gitignore` has no effect on files git already tracks — and this branch still tracks dozens of them (`git grep` finds `docs/superpowers/plans/2026-06-08-eval-extract.md`, `docs/superpowers/specs/...`, and many more on this very branch). So the result is a split brain: every plan and spec written before today stays in the repo and under review, every one written after today is invisible to `git status` and silently lost on a fresh clone.

That is a repo-policy decision with real consequences — the spec this PR implements (`2026-07-30-eval-cleanups-design.md`) is one of the documents that would stop being tracked, and it is the only record of the decisions this code encodes. Either untrack them deliberately (`git rm --cached -r docs/superpowers`) with that decision written down somewhere, or drop this line. Either way it wants its own PR, not a 77-file refactor.

packages/agency-lang/bar.agency
```diff
@@ -1,5 +1,5 @@
-import { ask } from "std::agents/composable/utils"
+import { callFunc } from "greeter.agency"
```
egonSchiele, line 1: This file and `foo.agency` are scratch/experiment files unrelated to the eval refactor, and this change commits a broken one: `greeter.agency` does not exist anywhere on this branch (`git cat-file -e origin/adit/eval-refactor2:packages/agency-lang/greeter.agency` → does not exist). So `bar.agency` at the repo root now imports a module that is not in the repo, and `$( callFunc() )` is a compile-time splice whose target cannot resolve.

Anything that compiles or type-checks the repo's own `.agency` files will trip over it, and a reader who opens it to learn something learns a broken example instead.

Revert both `bar.agency` and `foo.agency` out of this PR (the `foo.agency` change is pure reformatting of two `push`/`failure` calls — also unrelated). If these scratch files are meant to stay in the tree at all, that is worth its own conversation; if they are just workbench, they belong in `.gitignore` — which, given the comment above, seems to be the thing that was actually intended here.

packages/agency-lang/lib/eval/types.ts
```diff
@@ -155,7 +165,10 @@ export type InterruptEntry = {
   /** Full data payload as it was on the interrupt object. May be
    *  large — consumers should preview, not log verbatim. */
   data: unknown;
-  outcome: "approved" | "rejected" | "propagated" | "unresolved";
+  /** Same vocabulary as interrupt handlers (approve / reject / propagate /
+   *  pass). "passed": no interruptResolved event recorded a decision — every
+   *  handler passed on it, or the run ended before one decided. */
+  outcome: "approved" | "rejected" | "propagated" | "passed";
```
egonSchiele, line 171: Two things here, one mechanical and one about what the value claims.

**This is a change to a persisted artifact format, not just a vocabulary tidy.** `outcome` is written into `eval-record.json`, and every record already sitting under `runs/` says `"unresolved"`. Grading hands the parsed record straight to graders as `LoadedRun.record`, so a custom grader written against `outcome === "unresolved"` silently stops matching, and old records now carry a value that is not in the type. This codebase deliberately keeps pre-#733 layout tolerance in `readEvalRun`, which says you do care about re-reading old runs — so this rename deserves either the same tolerance (map `"unresolved"` on read) or an explicit "old records are not forward-compatible" note in the PR body, where `agentLabel` and the dropped `inputsSource` are already listed.

**The new name asserts more than the data supports.** `buildInterruptEntry` reaches this value when `outcomeRaw` is anything other than approved/rejected/propagated — including a legacy trace that never emitted `interruptResolved` at all, and a run that ended mid-interrupt. `"unresolved"` was honest about that ("nobody recorded a decision"). `"passed"` is a specific handler outcome, so a record now affirmatively states that every handler passed on an interrupt when what actually happened is that we have no idea. Given that handlers are safety infrastructure here, I would keep the two distinguishable: `"passed"` when a resolution event says so, `"unresolved"` when nothing did. The doc comment already describes both cases, which is the tell that they are two facts wearing one name.

egonSchiele (commented):
## Review

The layering lands well. I traced the three levels end to end and the invariant actually holds: `runAgent` returns no record, `gradeRun` takes only a directory, `judgeSuite` reads both directories, and `lib/optimize` no longer imports from `lib/cli`. The synthesized fake `EvalRunInputResult` inside grading is genuinely gone — `Entry` is now `{ input, recordPath, workdir, ungradedReason? }`, which is what grading actually needs. `recordGrading` reading `summary.json` back off disk rather than mutating an in-memory object is the detail that makes the "on the way in AND on the way out" claim true rather than aspirational.

Things I checked that turned out fine, so nobody has to re-check them:

- **The optimizer's extractor swap is behavior-identical.** `optimizeEvalRecordExtractor` on `main` was exactly `new StatelogParser(path, { warnMissingValue: false })`, and the default was the no-options constructor, whose `warnMissingValue` defaults to `true` (`lib/eval/extract.ts:80`). So `makeEvalRecordExtractor({ warnMissingValue: … })` reproduces both. The old comment claiming it also "grades the node's return value" was stale; deleting it is a fix, not a loss.
- **The errored-run policy stays unreachable for the optimizer.** `runInputViaEval` still throws before returning a directory, so routing it through `gradeRun` cannot change scores via `reasonByStatus.failed`.
- **The loader tolerance is where it needs to be**: `readOptionalJson` degrades with a warning, `readJson` fails loudly naming the file, duplicate ids warn. That was the one hole I flagged in the plan review and it is closed.

Seven comments inline. The two I would not merge without resolving are the `.gitignore` line and `bar.agency` — both unrelated to the refactor, and one of them commits a file that does not compile. The `"unresolved" → "passed"` rename and the untested interrupt fix are the two I would want a decision on.

### Not blocking, but worth a line in the PR body

**`judgeSuite` now iterates the union of ids present in the two run directories**, where before it iterated the caller's `inputs` list. An input the suite intended but never attempted (a run aborted with `continueOnError: false`) now silently vanishes from `perInput` instead of producing a missing-data verdict, which moves the `winsA`/`winsB`/`ties` denominators. Attempted-but-failed inputs still appear (they are in `summary.json`), so the blast radius is small — but "what the suite intended" is no longer knowable from a run directory, and that is a real consequence of dropping the `inputs` parameter.

**The PR body lists `agentLabel` and the dropped `inputsSource` as artifact key changes but not the `InterruptEntry.outcome` value rename**, which is the same class of change to a persisted file. See the inline comment on `types.ts`.

packages/agency-lang/lib/eval/run/subprocess.ts
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * The eval parent's blanket auto-approval for subprocess interrupts (eval
+ * runs headless; the agent's own handlers ran first and a local reject is
+ * already final before the parent is consulted). The `satisfies` pins the
+ * IPC protocol: this reply once used a legacy `{ approved: true }` shape,
+ * the child read `outcome.kind` off undefined, and every interrupting agent
+ * under `agency eval run` crashed.
+ */
+function evalInterruptDecision(interruptId: string): IpcDecisionMessage {
```
egonSchiele, line 124: Good catch on the bug — a legacy `{ approved: true }` reply crashing every interrupting agent under `agency eval run` is exactly the sort of thing that hides behind a refactor. Three notes on the fix.

**It has no test.** The doc comment is the only record that this shape matters, and the failure it describes ("every interrupting agent crashed") is severe enough to deserve a tripwire. An `tests/agency/` execution test — an agent that raises one interrupt, run through `agency eval run` — pins it without any LLM call, which is what that suite is for.

**The `satisfies` is doing nothing.** The function already declares `: IpcDecisionMessage`, so the return value is checked against that type regardless; `satisfies IpcDecisionMessage` on the same expression adds no constraint. The comment credits it with pinning the protocol, which will mislead the next person into thinking removing the annotation is safe because the `satisfies` covers it. Keep the return type, drop the `satisfies`, keep the story.

**The fix changes real behavior, and the PR body should say so.** Before: an interrupting agent under eval crashed, so nothing was auto-approved. After: every interrupt is approved unconditionally. That is the intended policy as documented — but it means `agency eval run` now silently approves destructive tool calls that previously could not proceed at all, and there is no record of the approvals in the statelog. Worth confirming that is the wanted policy for eval, and worth emitting the decision into the statelog so a run's approvals are auditable after the fact.

packages/agency-lang/lib/eval/judge/suite.ts
```diff
@@ -80,17 +78,23 @@ export function aggregateSuite(perInput: InputVerdict[], policy: JudgeAggregatio
...
+  // Run A's inputs in summary order, then any ids only run B has — a lopsided
+  // pair yields missing-data verdicts rather than silently dropping inputs.
+  const ids = [
+    ...Object.keys(runA.inputsById),
+    ...Object.keys(runB.inputsById).filter((id) => !Object.hasOwn(runA.inputsById, id)),
+  ];
+  for (const id of ids) {
     const inputA = runA.inputsById[id] ?? missingInput(id);
     const inputB = runB.inputsById[id] ?? missingInput(id);
+    const spec: Input = inputA.input ?? inputB.input ?? { id, args: {} };
```
egonSchiele, line 95: `goal` is optional on `Input` (only the input-file loader requires it), and `evalJudge` now *rejects* `--goal` for directory comparisons. So for a run whose `input.json` has no goal, `spec.goal ?? ""` sends an empty goal to the pairwise judge and there is no longer any way for the user to supply one — the old `inputsFromInlineGoal` path covered exactly this case.

An LLM judge asked to compare two outputs against `""` will still return a confident-looking verdict, which is the worst version of this failure: silent, not loud. Either warn when a judged input has no goal, or refuse the input with a missing-data verdict the way the `status !== "ok"` branch above does — the machinery for "this input could not be judged" is already right there.

packages/agency-lang/lib/optimize/baseOptimizer.ts
```diff
@@ -241,8 +243,10 @@ export abstract class BaseOptimizer {
     const perInput = await Promise.all(
       inputs.map(async (input, index) => {
         const id = inputId(input, index);
-        const result = await this.cache.get(ws.key, id, () => this.runInput(ws, source, files, input, id));
-        return gradeInput(input, result, ctx);
+        const runDir = await this.cache.get(ws.key, id, () => this.runInput(ws, source, files, input, id));
+        // A suite of one, graded like every other run directory.
+        const card = await gradeRun(runDir, ctx);
```
egonSchiele, line 248: `card.perInput[0]` is unguarded. It is safe today for a real reason — `runInputViaEval` throws unless `result.inputs[0]` succeeded, so the directory always has exactly one graded input — but that reason lives in a different method, and `runInput` is an injectable seam that custom optimizers implement (it is exported from `lib/optimize/public.ts` and documented in `writing-optimizers.md`). A seam that returns a directory with zero or two inputs puts `undefined` into `new Scorecard([...])`, and the failure surfaces much later as a confusing crash inside `objective()`.

One line — throw naming the run directory and the count when `perInput.length !== 1` — turns that into an immediate, readable error and documents the suite-of-one contract at the place that depends on it.

packages/agency-lang/lib/optimize/testUtils.ts
```diff
@@ -2,16 +2,16 @@ import * as fs from "fs";
...
- * `{ output, recordPath }` directly; now they return this.
+ * A run directory (suite of one) backed by real artifacts on disk, because
+ * grading loads the directory: eval-record.json, input.json, summary.json.
+ * Optimizer tests inject `runInput` seams that return this directory.
  */
-export function fakeRun(inputId: string, output: unknown, dir?: string): EvalRunInputResult {
-  const root = dir ?? fs.mkdtempSync(path.join(os.tmpdir(), "optimize-run-"));
-  const inputDir = path.join(root, inputId);
+export function fakeRun(inputId: string, output: unknown, spec?: Input): string {
+  const root = fs.mkdtempSync(path.join(os.tmpdir(), "optimize-run-"));
```
egonSchiele, line 13: Losing the `dir?` parameter means callers can no longer control or clean up the temp root, and every call now mints a fresh `mkdtempSync` that nothing ever removes. `gepa.test.ts` and `greedyReflective.test.ts` call `fakeRun` once per (candidate, input) across several iterations, so a full optimizer test run leaves a few dozen `optimize-run-*` directories in the system temp dir, every run, forever.

Not a correctness problem, but it is the kind of litter that makes a developer's `/tmp` unusable after a few weeks. Either keep an optional root parameter so a test's `afterEach` can clean up, or have `fakeRun` push onto a module-level list with a shared cleanup helper — the pattern `seed.test.ts` and `sources.test.ts` already use.

PR #735, opened 2026-07-31
eval: land the #734 review fixes; remove the outdated stdlib evalRun pipeline

egonSchiele:
Two things, in two commits:

## 1. The #734 review-round fixes that missed the squash-merge

The review fixes for #734 were made in the working tree but never committed, so the merge did not include them. This lands them:

- **`InterruptEntry.outcome`**: `"passed"` is now only ever an affirmative claim from a resolution event; `"unresolved"` returns for the absence of one (run died mid-interrupt, or a legacy trace). Records already on disk parse into the type again; a test pins both mappings.
- **`judgeSuite` refuses goal-less inputs** with a deterministic tie verdict instead of asking an LLM judge to compare against an empty goal (which returns a confident-looking verdict - the silent version of the failure).
- **Suite-of-one guard** in `BaseOptimizer.evaluate`: a `runInput` seam that returns a directory with 0 or 2 graded inputs now fails immediately with the directory and count named, instead of surfacing later inside `objective()`.
- **New end-to-end test** (`tests/integration/eval-run/test.mjs`, wired into CI): an agent that raises an interrupt runs through `agency eval run` with a real subprocess - no LLM - and the test asserts the IPC auto-approval is recorded in the statelog (`interruptResolved`, `resolvedBy: "ipc"`) and survives into the eval record. This pins the `{ approved: true }` regression end-to-end and demonstrates the approvals ARE auditable.
- `fakeRun` temp-dir cleanup (`cleanupFakeRuns()` in every optimizer test's `afterEach`); `recordGrading` names `summary.json` in read errors; the no-op `satisfies` dropped; the dead `eval run -v/--verbose` flag removed; `bar.agency`/`foo.agency` reverted and the `docs/superpowers` `.gitignore` line dropped, per review.

## 2. Remove the stdlib evalRun pipeline

The Agency-side run pipeline predated the workdir redesign and diverged from `agency eval run` on three counts: it seeded nothing (its workdir was an empty cwd - why the Agency `Input` never got a `files` field), it counted a run with no statelog as success (the exact case `runAgent` now treats as an error), and it carried no provenance. Removed whole rather than absorbed - when Agency code needs to run suites again, the right shape is a thin binding over `runSuite`.

`evalExtract`, `evalJudge`, and `evalJudgeSuite` stay: they are thin delegates to the current core. `recordInputSuccess`/`recordInputRunFailure` die with their only consumer. Stdlib docs regenerated.

Tests: typecheck, structural lint, and the eval/cli/optimize/stdlib vitest suites green; the new integration test passes against a fresh build.


packages/agency-lang/lib/eval/extract.ts
```diff
@@ -211,9 +211,10 @@ function buildInterruptEntry(
   const outcome: InterruptEntry["outcome"] =
     outcomeRaw === "approved" ||
       outcomeRaw === "rejected" ||
-      outcomeRaw === "propagated"
+      outcomeRaw === "propagated" ||
+      outcomeRaw === "passed"
```
egonSchiele, line 215: The inversion is right — `"unresolved"` for the absence of a decision, `"passed"` only from a resolution event — but I went looking for where a real `"passed"` comes from and could not find one.

Every `interruptResolved` emitter in the runtime writes only `"approved"` or `"rejected"`: `lib/runtime/interrupts.ts:484`, `:493`, `:518` (the in-handler refusal), and `:732` (the user-resolution loop, `resp.type === "approve" ? "approved" : "rejected"`). Nothing emits `"passed"`, and nothing emits `"propagated"` either — that one predates this PR, so it is not a regression, but it means two of the four values in the union are unreachable from any real run.

Three consequences:

1. **The new test in `extract.test.ts` hand-builds an `interruptResolved` event with `outcome: "passed"`, a shape the runtime never produces.** It correctly pins the extractor's mapping, but it cannot catch the thing the doc comment claims ("`passed` is only ever an affirmative claim from a resolution event") because no affirmative claim is ever made. Worth a comment in the test saying it pins the mapping against a hypothetical emitter.
2. **A handler that genuinely passes on an interrupt is now indistinguishable from a run that died mid-interrupt** — both land on `"unresolved"`. That is the honest answer given what the statelog carries, but it is an observability gap in handler behavior specifically, which this codebase treats as safety infrastructure.
3. **The type doc oversells the vocabulary.** "Same vocabulary as interrupt handlers (approve / reject / propagate / pass)" reads as though all four arrive; two never do.

Either teach the runtime to emit `passed`/`propagated` (the merge already distinguishes those outcomes internally — `hasPropagation` and the pass case are both computed), or say plainly on the type which values a current runtime can actually produce and which exist for older or future traces. The second is a one-line change and keeps this PR honest; the first is the better answer and probably its own issue.

packages/agency-lang/stdlib/agency/eval.agency
```diff
@@ -1,32 +1,14 @@
...
-
-  node main() {
-    const agent = compile("agent.agency")
-    const result = evalRun(agent, [
-      { id: "capital-france", goal: "Return Paris", args: {} },
-    ])
-    print(result.runDir)
-  }
-  ```
+  Helpers for judging eval runs from Agency code. (Running suites from
```
egonSchiele, line 8: Agreed on removing this, but the forward-looking sentence — here and in the PR body, "the right shape is a thin binding over `runSuite`" — would recreate the binding without the property the old one existed for.

The original design made stdlib `evalRun` call `std::agency.run(...)` for every task *specifically* so subprocess interrupts hit the **caller's** handler stack; the reasoning is written down in `docs/superpowers/plans/2026-06-09-eval-run-command.md:74-83` ("the stdlib `evalRun` function may be called by an agent, so it must call `std::agency.run(...)` for every task to preserve parent-handler enforcement for subprocess interrupts").

`runSuite` does not have that property. Its default runner is `makeSubprocessRunner`, and `evalInterruptDecision` blanket-approves every interrupt from TypeScript — correct for a headless CLI invocation, wrong for a suite launched *by an agent*, where the caller's handlers should decide. So a literal "thin binding over `runSuite`" called from Agency code would silently auto-approve child interrupts that the calling agent's handlers were meant to gate.

Removing the binding removes the risk, which is why this PR is fine. But the note that survives it should carry the constraint, not just the shape — something like "a future binding must route execution through `std::agency.run` (or inject a runner that defers to the caller's handlers), not use `runSuite`'s default auto-approving runner." Otherwise the next person implements the unsafe version and this comment is the only place that would have said so.

packages/agency-lang/tests/integration/eval-run/test.mjs
```diff
@@ -0,0 +1,87 @@
...
+  assert(resolved.data.outcome === "approved", `outcome: ${resolved.data.outcome}`);
+  assert(resolved.data.resolvedBy === "ipc", `resolvedBy: ${resolved.data.resolvedBy}`);
+
+  // The record survives extraction: the interrupt shows up approved.
+  const record = JSON.parse(readFileSync(join(runsDir, "interrupt-e2e", "inputs", "interrupting", "agent", "eval-record.json"), "utf-8"));
+  assert(record.interrupts.length === 1, `interrupts: expected 1, got ${record.interrupts.length}`);
+  assert(record.interrupts[0].outcome === "approved", `record outcome: ${record.interrupts[0].outcome}`);
+
+  console.log("[eval-run-integration] PASS: interrupting agent auto-approved over IPC, verdict recorded");
+} finally {
```
egonSchiele, line 102: This `finally` deletes the evidence on the failure path, which is the one path where you need it.

The combination is unfortunate: `execSync` throws without printing the child's output (the message is just `Command failed: node .../agency.js eval run …`; the actual stdout/stderr sit on `error.stdout`/`error.stderr` and nobody reads them), and then `finally` removes `.tmp` — including the statelog and run directory this test exists to inspect. A CI failure would give a maintainer the command line and nothing else, with no way to reproduce beyond running it locally and hoping it fails the same way.

The repo already has the pattern for this: the optimizer efficacy harness keeps its runs on failure and behind `OPTIMIZE_EFFICACY_KEEP_RUNS=1` (see the `.gitignore` comment). Suggest the same shape — wrap in `try/catch`, print `error.stdout` and `error.stderr`, and only clean up on success (or when the keep-flag is unset).

Smaller, in the same file: `const r = check()` binds a variable nothing reads.

.github/workflows/test.yml
```diff
@@ -166,6 +166,9 @@ jobs:
     - name: StateLog integration tests
       working-directory: packages/agency-lang
       run: node tests/integration/statelog/test.mjs
+    - name: Eval-run integration tests
```
egonSchiele, line 169: No `timeout-minutes` on this step, unlike the `Stdlib sandbox tests` step three lines below it.

The agent subprocess itself is bounded (`DEFAULT_EVAL_RUN_LIMITS.wallClock` is 60s), so the realistic hang is in the parent — the CLI wedged before or after the fork, or an interrupt that never gets answered, which is exactly the failure class this test guards. In that case the step inherits the job-level default and burns CI time until GitHub's own limit.

`timeout-minutes: 5` is generous for a single no-LLM run and turns a wedge into a fast, clearly-labelled failure.

packages/agency-lang/lib/eval/judge/suite.ts
```diff
@@ -158,6 +165,23 @@ function missingDataVerdict(input: Input, inputA: ReadEvalRunInput, inputB: Read
   };
 }
 
+/** Neither side can win an unjudgeable input; ties never count toward a
+ *  suite winner, so a goal-less input degrades loudly in perInput without
+ *  skewing the aggregate. */
+function noGoalVerdict(inputId: string, inputA: ReadEvalRunInput, inputB: ReadEvalRunInput): InputVerdict {
```
egonSchiele, line 172: The claim in the comment checks out — `suiteWinner` compares only `winsA`/`winsB`, so a tie genuinely cannot skew the suite winner.

Worth knowing about the reported numbers though: a goal-less input still increments `ties` in `aggregateSuite`, so a `SuiteVerdict` reading `winsA 4, winsB 4, ties 3` is indistinguishable at a glance from a real 3-way dead heat, when it may mean three inputs were unjudgeable. Only `perInput[].reasoning` tells them apart, and the aggregate is what a person reads first.

Not worth a schema change on its own, but if `missingDataVerdict` inputs have the same issue (they can win rather than tie, so partly not), one line in the verdict output — or counting unjudgeable inputs separately — would keep "we could not judge this" from quietly reading as "these were close."

packages/agency-lang/lib/eval/types.ts
```diff
@@ -166,9 +166,12 @@ export type InterruptEntry = {
...
   /** Same vocabulary as interrupt handlers (approve / reject / propagate /
-   *  pass). "passed": no interruptResolved event recorded a decision — every
-   *  handler passed on it, or the run ended before one decided. */
-  outcome: "approved" | "rejected" | "propagated" | "passed";
+   *  pass) — plus "unresolved", which is not an outcome but the absence of
+   *  one: no interruptResolved event recorded a decision (the run ended
+   *  mid-interrupt, or a legacy trace never emitted one). "passed" is only
+   *  ever an affirmative claim from a resolution event; the extractor never
+   *  infers it. */
+  outcome: "approved" | "rejected" | "propagated" | "passed" | "unresolved";
```
egonSchiele, line 185: One thing the doc could add while it is being rewritten: this union is the shape of a **persisted** field. Records written before #734 say `"unresolved"`, records written between #734 and this PR say `"passed"` for the same situation, and records written after this say `"unresolved"` again.

So any run directory produced by a build from that window carries a `"passed"` that means "nobody decided" — the exact opposite of what the type now says it means. The window is short and probably only your own `runs/`, which is why I would not add a migration. But a sentence here ("records written by builds between #734 and #735 may say `passed` where they mean `unresolved`") costs nothing and saves whoever eventually re-grades an old run from trusting a value that lies.

egonSchiele (commented):
## Review

Every point from the #734 round is genuinely addressed — I checked each one rather than taking the list on faith:

- `outcome` mapping inverted correctly and `"unresolved"` is back in the union, so records already on disk parse into the type again.
- `judgeSuite` refuses goal-less inputs before reaching the judge, and after the `status !== "ok"` branch, which is the right order.
- The suite-of-one guard names both the directory and the count.
- `cleanupFakeRuns` is wired into all five optimizer test files, and `afterEach` was already imported in every one of them (I checked — no missing import).
- `bar.agency`/`foo.agency` reverted to their `main` content, the `docs/superpowers` ignore line dropped, the no-op `satisfies` gone, `recordGrading` names the file.
- The new CI step sits after `make` in the same job, so `dist/scripts/agency.js` exists when it runs. And removing `recordInputRunFailure` did not orphan `PreparedInput.errorPath` — it is still used by the stale-artifact cleanup at `runArtifacts.ts:132`.

The stdlib removal reads right to me. The three divergences you list (no seeding, no-statelog counted as success, no provenance) are real, and deleting is better than absorbing a pipeline nobody was using.

Six comments inline. The one I would not merge without a decision is the first: **`"passed"` cannot be produced by the current runtime**, so the new test pins a mapping that no real run can reach.

### Two things not anchorable to the diff

**`docs/dev/eval-grading.md:16` is already stale.** It says "The suite runner (`evalRunLoadedInputs`) executes agents and writes the run directory". That name does not exist anywhere in `lib` — #734 renamed it `runSuite`. This is the document created specifically to be the durable record of the invariant, and it went stale one PR after it was written; worth fixing here while the context is fresh.

**`lib/eval/runArtifacts.ts` has four consecutive blank lines** (158–161) where `recordInputRunFailure` and `recordInputSuccess` were removed. Cosmetic, but it is the kind of thing that survives for years once merged.

### One note for the PR body

#734's body had a **User-visible changes** heading; this one folds two breaking changes into a misc bullet and a section title. Both deserve to be called out under that heading: `agency eval run -v/--verbose` now fails with "unknown option" rather than being ignored, and `evalRun` is gone from `std::agency/eval`, so any Agency code importing it stops compiling.

packages/agency-lang/lib/eval/extract.ts
egonSchiele: Fixed in 4c3cf924a: the type doc now states reachability plainly (today's runtime emits only approved/rejected; propagated/passed exist for older or future traces), and the extract test carries the hypothetical-emitter caveat. Teaching the runtime to emit them is filed as #736.

packages/agency-lang/stdlib/agency/eval.agency
egonSchiele: Fixed in 4c3cf924a: the module note now carries the constraint verbatim — a future binding must route through std::agency.run (or inject a caller-deferring runner), never runSuite's default auto-approving runner. Docs regenerated. PR body updated to match.

packages/agency-lang/tests/integration/eval-run/test.mjs
egonSchiele: Fixed in 4c3cf924a: failure path keeps the scenario dir and prints the child's captured stdout/stderr; EVAL_RUN_KEEP_TMP=1 keeps it on success too; the unused binding is gone.

.github/workflows/test.yml
egonSchiele: Added timeout-minutes: 5 in 4c3cf924a.

packages/agency-lang/lib/eval/judge/suite.ts
egonSchiele: Fixed in 4c3cf924a: both deterministic verdicts (missing data, no goal) now set an unjudgeable marker on InputVerdict, and eval judge prints the count so unjudgeable ties do not read as close ones.

packages/agency-lang/lib/eval/types.ts
egonSchiele: Added in 4c3cf924a: the doc notes the field is persisted and names the #734–#735 window where records say passed meaning unresolved.

packages/agency-lang/lib/eval/judge/suite.ts
```diff
@@ -158,6 +165,23 @@ function missingDataVerdict(input: Input, inputA: ReadEvalRunInput, inputB: Read
...
+  const reasoning = "input has no goal in its input.json; nothing to judge against";
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 4c3cf924a: the reasoning now says no goal was recorded for the input, without asserting input.json exists.

PR #737, opened 2026-07-31
eval follow-ups: empty-suite guard, legacy tolerance removal, interrupt chain outcomes (#736), compile out of cli/

### LINKED ISSUE #736: statelog: runtime never emits interruptResolved outcomes "propagated" or "passed"
Every `interruptResolved` emitter in the runtime writes only `"approved"` or `"rejected"` (`lib/runtime/interrupts.ts:484`, `:493`, `:518`, `:732`). The handler-chain merge already computes the other two outcomes internally — `hasPropagation` and the pass case — but they never reach the statelog:

- an interrupt that propagates to the user records nothing,
- a handler that genuinely passes is indistinguishable in the trace from a run that died mid-interrupt (both extract as `"unresolved"`).

Handlers are safety infrastructure, so handler behavior specifically should be observable from a trace. `InterruptEntry.outcome` and the extractor already accept `"propagated"` and `"passed"` (see the reachability note on the type added in #735) — the runtime just needs to emit them at the points where the merge already knows.

Raised during review of #735.

egonSchiele:
The four follow-ups decided after the eval refactor, one commit each:

## 1. An empty suite is an error

`agency eval run` with a source that loads zero inputs (wrong path, empty file) used to "succeed" with `0/0 inputs ok` and objective 0 - the failure mode that hid the stale-build problem in the first demo. `loadSuite` now throws `no inputs loaded from <source>` before anything runs.

## 2. Legacy layout fallbacks deleted

`readEvalRun` only reads the current `inputs/<id>/agent/` layout; the pre-#733 flat-path fallbacks are gone, and a legacy-layout run loads as `missing`. Old saved runs are declared disposable (no-backward-compat ground rule). The pre-#726 statelog tolerances in extraction go with them (in commit 3, which touches the same lines): the `interruptThrown`-data fallback and the null-degrade support for old statelogs. Re-extract museum pieces with an old build if ever needed.

## 3. Fix #736: surfaced interrupts record their chain outcome

Every `interruptResolved` emitter in the runtime wrote only `approved`/`rejected`, so a handler that **propagated** or **passed** was indistinguishable in a trace from a run that died mid-interrupt - an observability gap in the safety infrastructure. Now, when an interrupt surfaces, `renderVerdict` records what the chain concluded:

```jsonc
{ "type": "interruptResolved", "outcome": "propagated", "resolvedBy": "handler", ... }
// ...user decides later:
{ "type": "interruptResolved", "outcome": "approved", "resolvedBy": "user", ... }
```

An interrupt can therefore carry more than one `interruptResolved`; **the last is authoritative**, and extraction now reads it that way (a user decision supersedes the chain outcome; a run that dies surfaced keeps `propagated`/`passed`). `"passed"` means the chain finished without deciding - every handler passed, or none existed. All five `InterruptEntry.outcome` values are now reachable from real runs. Breaking for statelog consumers that assumed exactly one terminal event per interrupt, per the no-backward-compat ground rule.

## 4. The last upward import: compile out of the CLI

`compile`/`compileMany`/`resetCompilationCache` move from `lib/cli/commands.ts` to `lib/compiler/defaultSession.ts`, beside the `BuildSession` they delegate to; after review, `findRecursively` (-> `lib/utils/`) and `getImports`/`getAllImports` (-> `lib/analysis/imports.ts`) moved too, so `lib/compiler` and the compile path genuinely stop importing from `lib/cli` - not just in one hop. The lazy session construction (and its `agency pack` tree-shaking rationale) moves verbatim; ~30 importers repointed, no shims. Two known library-to-CLI edges remain and are out of scope: the node/agent execution helpers (`agencyRunner` -> `runAgencyNode`, `judge/pairwise` -> `runAgencyAgent`) - `cli/util.ts` is two concepts wearing one name, and that split is its own change.

Deferred by explicit decision: the `EvalRunResult` execution/grading restructure, and the `NormalizedEvent` coverage question.

Tests: full vitest suite (9,942), the statelog and eval-run integration harnesses, typecheck, and structural lint - all green.


packages/agency-lang/lib/runtime/interrupts.ts
```diff
@@ -547,6 +549,19 @@ function renderVerdict(
...
+  // Record what the handler chain concluded before the interrupt surfaces:
+  // "propagated" (a handler sent it up) or "passed" (the chain finished
+  // without deciding — every handler passed, or none existed). Without this,
+  // a trace could not tell deliberate handler behavior from a run that died
+  // mid-interrupt (#736). If the user later decides, a second
+  // interruptResolved (resolvedBy "user") follows — consumers take the LAST
+  // event for an interruptId as authoritative.
+  ctx.statelogClient.interruptResolved({
+    interruptId: intr.interruptId,
+    outcome: merged.kind === "propagated" ? "propagated" : "passed",
```
egonSchiele, line 564: The `outcome` is right; `resolvedBy` is not, and it reintroduces the exact ambiguity this commit removes — one field over.

`renderVerdict`'s `resolvedBy` parameter is `"ipc" | "handler"`, never null. So when the chain reaches `noResponse` because **no handler existed at all**, this emits `{ outcome: "passed", resolvedBy: "handler" }` — asserting a handler was involved when none was. The type's own doc spells out that these are two different situations: *"passed" = the chain finished without deciding — every handler passed, **or none existed***. #736 was about not conflating deliberate handler behavior with its absence; that is still conflated here, just in `resolvedBy` instead of `outcome`.

`InterruptEntry.resolvedBy` is already `… | null`, so the honest value is available. `runHandlerChain` knows whether it ran any handler, so this can be `merged.kind === "propagated" ? resolvedBy : (chainHadHandlers ? resolvedBy : null)` — or, arguably cleaner, `null` for both surfaced cases on the grounds that nothing *resolved* anything (`propagated` is a hop, not a resolution).

The two new tests do not catch this because both construct a chain **with** a handler — `makeCtx([async () => ({ type: "propagate" })])` and `makeCtx([async () => ({ type: "pass" })])`. A third with `makeCtx([])` is the case that matters most for a safety trace ("nobody was watching this effect") and is the one currently unpinned.

packages/agency-lang/lib/cli/eval/run.ts
```diff
@@ -119,12 +119,21 @@ function loadSuite(args: {
...
   return {
-    inputs: loadInputs(parsed.path, nanoid, loadOptions),
+    inputs: requireInputs(loadInputs(parsed.path, nanoid, loadOptions), parsed.path),
     provenance: { inputsSource: { source: parsed.path }, files: filesProvenance },
   };
 }
+
+/** An empty suite is a mistake (wrong path, empty file), never a run: running
+ *  zero inputs would "succeed" with objective 0 and hide the mistake. */
+function requireInputs(inputs: Input[], source: string): Input[] {
```
egonSchiele: This guards `agency eval run` only. `agency eval optimize` still loads inputs bare — `lib/cli/eval/optimize.ts:142`, `withDefaults(loadInputs(path.resolve(p), …), node, s.goal)` — so a wrong path or empty file there produces the same silent zero-input run this commit exists to stop.

And optimize is the worse case. `eval run` fails fast and cheap; optimize will run its full iteration budget proposing mutations, grading each candidate against nothing, and report objective 0 for every one of them — real LLM spend on a suite that does not exist, ending in a number that looks like a bad agent rather than a bad path.

The check has no per-command content, so `loadInputs` itself is the better home: every caller wants it, and then `evalJudge` (which loads suites too) is covered for free. Worth a quick look at `splitInputs`/`--validation-inputs` first to confirm nothing legitimately loads an empty suite on purpose.

packages/agency-lang/lib/eval/readRun.ts
```diff
@@ -25,17 +25,10 @@ export function readEvalRun(runDir: string): ReadEvalRunResult {
   for (const result of summary.inputs) {
     const inputDir = path.join(resolvedRunDir, "inputs", result.inputId);
     const input = readOptionalJson<Input>(path.join(inputDir, "input.json"));
-    const recordPath = result.evalRecordPath
-      || firstExisting(
-        agentRunPaths(inputDir).evalRecordPath,
-        path.join(inputDir, "eval-record.json"),      // pre-#733 layout
-      );
+    const recordPath = result.evalRecordPath || agentRunPaths(inputDir).evalRecordPath;
```
egonSchiele, line 28: Agreed with dropping the fallbacks, but the resulting failure mode is quiet in a way that will cost someone an afternoon.

A pre-#733 run now resolves every `recordPath` to a file that does not exist, so `inputStatus` returns `"missing"` for all of them, `loadedEntry` maps that to `"no eval record found on disk for this input"`, and `agency eval grade <old-run>` prints `objective 0.00` with every input scored zero. Nothing in that output says "this run predates the current layout" — it reads exactly like an agent that produced nothing, which is a real and much more common situation.

One cheap signal keeps the deletion and removes the confusion: if every input is `missing` **and** `inputs/<id>/eval-record.json` exists at the old flat path, warn once — "this run uses the pre-#733 layout and is no longer readable; re-run, or grade it with a build from before <sha>". That is three lines and it does not resurrect the fallback, it just names what happened.

egonSchiele (commented):
## Review

Commit 3 is the one worth the most attention and it is mostly right. Closing #736 by recording the chain outcome at the surface point is the correct shape: `renderVerdict` already knows `merged.kind`, and emitting there means the trace distinguishes "a handler propagated this" from "the run died mid-interrupt" without any new bookkeeping. The last-wins rule is the right call too, and the extractor change matches it.

I traced the eval-run path to check the multi-event change does not break the #735 integration test: it does not. `gatherChainOutcome` consults the parent **before** `renderVerdict` runs, so in IPC mode `merged.kind` is already `approved` and only one `interruptResolved` is emitted — the test's `.find()` still lands on the right event. Worth stating in the PR body, because "an interrupt can now carry more than one `interruptResolved`" reads like it would break exactly that assertion.

Also checked and fine: `interruptThrown` has exactly one emitter (`interrupts.ts:548`), immediately followed by the new resolved event carrying the summary — which is what makes commit 2's removal of the `interruptData` fallback safe. Note that this makes commits 2 and 3 order-dependent in a way the body understates: it says commit 3 "touches the same lines", but commit 2's `data: summary?.data ?? null` is only non-lossy *because* commit 3 guarantees a summary-carrying event for every surfaced interrupt. Reverting 3 alone would silently null out `data`. And `logsViewer` just renders each event as its own line, so a second `interruptResolved` there reads as extra information rather than double-counting.

Five comments inline. The one I would fix before merge is the first — `resolvedBy` re-conflates the two facts this PR exists to separate.

### The PR body's headline claim for commit 4 is not accurate

> `lib/eval/run/seed.ts` no longer imports from `lib/cli`, which was the last library-to-CLI edge.

Both halves are wrong on this branch:

- **It was not the last edge.** Three others remain: `lib/eval/grading/agencyRunner.ts:7` (`runAgencyNode` from `@/cli/util.js`), `lib/eval/judge/pairwise.ts:3` (`runAgencyAgent` from `@/cli/runAgencyAgent.js`), and `lib/optimize/mutator.ts:8` (`executeNodeAsync` from `@/cli/util.js`).
- **`seed.ts` still reaches `lib/cli`, now indirectly.** `lib/compiler/defaultSession.ts` imports `createBuildSession` from `./buildSession.js`, and `lib/compiler/buildSession.ts:55` imports `findRecursively`/`getImports` from `@/cli/util.js`. So the edge `seed.ts → cli` was not removed; it was made transitive, which is the version a grep no longer finds.

The move is still worth doing — `compile` belongs beside the `BuildSession` it delegates to, and that is reason enough. But the claim should say what actually happened, and `buildSession.ts`'s own upward import is now the thing standing between this and a real layering boundary. `findRecursively`/`getImports` look like filesystem/parse helpers with no CLI content; if so, they are a small move away from making the claim true.

packages/agency-lang/lib/eval/extract.ts
```diff
@@ -196,10 +196,12 @@ function buildInterruptEntry(
 ): InterruptEntry {
   const sorted = [...group].sort((a, b) => a.tMs - b.tMs);
   const thrown = sorted.find((e) => e.type === "interruptThrown");
-  const resolved = sorted.find((e) => e.type === "interruptResolved");
+  // LAST, not first: a surfaced interrupt records its chain outcome
+  // ("propagated"/"passed") and may then be decided by the user — the later
+  // event supersedes.
+  const resolved = sorted.filter((e) => e.type === "interruptResolved").at(-1);
```
egonSchiele, line 202: "Last wins" is the right rule, but as written it rests on two properties that are true and unstated, and the case that exercises them is not exotic.

`sorted` is `[...group].sort((a, b) => a.tMs - b.tMs)` and `tMs` is millisecond-resolution. When a surfaced interrupt is decided quickly — an auto-answering driver, a test, the debugger — both `interruptResolved` events can carry the same `tMs`. The right answer still comes out, but only because (a) `byType.interruptResolved` preserves statelog file order and (b) `Array#sort` has been stable since ES2019, so equal keys keep their input order. Neither is written down, and the `group` array is assembled by concatenating three `byType` slices, which makes "input order" a couple of hops from obvious.

A sentence on the sort would do it. If the normalized envelope carries a sequence or line index, a tiebreak (`a.tMs - b.tMs || a.seq - b.seq`) makes it not depend on stability at all.

packages/agency-lang/lib/eval/run/seed.ts
```diff
@@ -3,7 +3,7 @@ import * as path from "path";
 
 import { agentClosure } from "@/analysis/closure.js";
 import type { AgencyConfig } from "@/config.js";
-import { compile } from "@/cli/commands.js";
+import { compile } from "@/compiler/defaultSession.js";
```
egonSchiele, line 6: This import is the one commit 4 is named for, so worth being precise about what it achieves: `seed.ts` still reaches `lib/cli`, just not in one hop.

`@/compiler/defaultSession.js` → `./buildSession.js` → `@/cli/util.js` (`buildSession.ts:55` imports `findRecursively` and `getImports`). So the dependency survives the move; what changed is that `grep -rn 'from "@/cli/' lib/eval` no longer finds it.

That is not an argument against the move — `compile` genuinely belongs next to `createBuildSession`. It is an argument for finishing it, and for the PR body not claiming the edge is gone. `findRecursively` and `getImports` read like filesystem-walk and import-parse helpers with nothing CLI about them; if that holds, moving them next to `buildSession` (or into `lib/utils/`) would make `lib/compiler` genuinely free of `lib/cli` and turn the claim true. Worth checking `lib/eval/grading/agencyRunner.ts` and `lib/optimize/mutator.ts` at the same time — both pull node-execution helpers out of `cli/util.ts`, which suggests that file is two concepts wearing one name.

packages/agency-lang/lib/runtime/interrupts.ts
egonSchiele: Fixed in da311bde5, taking the cleaner option: resolvedBy is null for both surfaced cases — nothing resolved anything, and which handler propagated or passed is already on the handlerDecision events. The statelog type gains | null. Third test added with makeCtx([]): an empty chain records passed without claiming a handler was involved.

packages/agency-lang/lib/cli/eval/run.ts
egonSchiele: Fixed in da311bde5: the guard moved into loadInputs itself, so optimize (and any other caller) is covered — with the optimize burn-the-budget failure mode named in the comment. Checked splitInputs/--validation-inputs: both operate on already-loaded lists, and an explicitly passed validation file loading zero inputs is the same mistake, so no caller legitimately loads an empty suite. One test converted from pinning the old empty-list behavior to pinning the throw.

packages/agency-lang/lib/eval/readRun.ts
egonSchiele: Fixed in da311bde5: readEvalRun warns once when every input is missing AND a record exists at the old flat path — names the pre-#733 layout and says re-run or use an older build. The fallback stays deleted.

packages/agency-lang/lib/eval/extract.ts
egonSchiele: Fixed in da311bde5 with the tiebreak rather than the comment: NormalizedEnvelope now carries seq (position in the statelog file), and the interrupt sort is tMs, then seq — no reliance on sort stability or on byType slice order.

packages/agency-lang/lib/eval/run/seed.ts
egonSchiele: Fixed in da311bde5: findRecursively moved to lib/utils/findRecursively.ts, getImports and getAllImports to lib/analysis/imports.ts, so lib/compiler now has zero lib/cli imports — the claim is true rather than one-hop true, and the PR body was corrected to say exactly what remains: the runAgencyNode/runAgencyAgent edges from eval grading and the judge, which are the other concept living in cli/util.ts and are left for their own change.

packages/agency-lang/lib/compiler/defaultSession.ts
```diff
@@ -0,0 +1,68 @@
+import type { AgencyConfig } from "@/config.js";
+
+import {
+  createBuildSession,
+  type BuildSession,
+  type CompileOptions,
+} from "./buildSession.js";
```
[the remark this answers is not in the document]
egonSchiele: Addressed in da311bde5 (see the sibling thread): buildSession's cli/util imports moved to lib/utils and lib/analysis, so the compiler↔CLI cycle is gone and importing compile() no longer pulls lib/cli transitively.

PR #738, opened 2026-07-31
terminal-bench-mini: regex-log eval suite, eval run limits/flags, SIGINT salvage

egonSchiele:
A mini terminal-bench: terminal-bench tasks re-expressed as Agency eval suites, so we can benchmark Agency agents with `agency eval run` instead of the Harbor/Docker harness. Ports the useful parts, not the packaging.

## The pilot suite: regex-log

`evals/terminal-bench-mini/regex-log/` ports terminal-bench 2.0's regex-log task (the rare task whose environment is empty — no fixtures, no setup script):

- The original instruction, adapted in two honest ways: the output path is workdir-relative, and grading semantics are stated as JavaScript RegExp (`"gm"` flags, group 1 when a capture group exists) instead of Python `re.findall`. The snippet is explicitly labeled as how the grader evaluates the output, not a language to work in.
- `graders.ts` ports the pytest verifier verbatim into pure TypeScript: all 25 sample log lines, same all-or-nothing ordered match-list equality. Both graders are `mustPass`, reproducing terminal-bench's 0-or-1 reward.
- `solution.agency` wraps terminal-bench's reference regex in a deterministic no-LLM agent (their `solution/solve.sh` convention). Running the suite against it smoke-tests the whole harness for free: it scores 1.0 end to end, and the reference regex was verified to produce exactly the expected match list under JS semantics.

## Eval framework changes the pilot demanded

- **Configurable wall clock.** The subprocess limit was hardcoded at 60s; no real tool-loop agent fits. `eval.limits.wallClockSec` in `agency.json` now overrides it (the repo config sets 600s, matching terminal-bench's easy-task budget). Other limits keep their defaults.
- **`agency eval run` gains `--max-tool-call-rounds`, `--max-tool-result-chars`, `--strict`** with the same meanings as `agency run`. All three are baked into the compiled program and eval compiles each agent inside its seeded workdir with the CLI config, so `applyCliFlags` before `evalRun` is the entire implementation.
- **Salvage-on-SIGINT.** Ctrl-C used to lose the end-of-run record extraction. `runAgent` now installs a SIGINT handler for exactly the duration of the agent subprocess: salvage an eval record from whatever statelog exists, write `error.txt` marking the run interrupted, exit 130.

## What the first real runs found (and the agent fixes)

Two benchmark runs of `foo.agency` scored 0, both diagnosed from the run directory alone — the eval framework works. The agent kept trying to solve the task directly in JavaScript inside `execute` instead of writing an Agency agent, burned its tool budget on parse errors, then returned the regex as prose without ever writing the file (which the `mustPass` gate caught). Fixes, each aimed at one link in that chain:

- `foo.agency`'s system prompt moved onto the thread via `systemMessage()` (it was pasted into every user prompt), gained an "Agency is not JavaScript" section, and `main` takes an optional `task` argument so the eval harness can invoke it headlessly.
- `execute` is renamed **`executeAgency`** with an Agency-only docstring — the tool description is what the model reads at call time.
- `review()` now detects JavaScript tells (`require`, `console.log`, arrows, backticks, `function`) when a program fails to parse and says "this is JavaScript — rewrite as Agency" with the Agency shapes to use, instead of a cryptic parser error the model patches in place.
- The `onLLMCallStart` trace callback truncates the printed prompt so long prompts cannot drown the trace.

## Testing

- New unit tests: `limitsFromConfig` merge; SIGINT listener lifetime in `runAgent`.
- Full eval-run pipeline smoke-tested with the deterministic solution agent (objective 1.000, no LLM calls).
- `foo.agency` compiles; affected vitest files pass (11/11).


packages/agency-lang/lib/eval/run/runAgent.ts
```diff
@@ -118,6 +124,26 @@ class AgentRunner {
     };
   }
 
+  /** Ctrl-C would otherwise lose the record extraction that only happens at
+   *  end of run. Salvage whatever the statelog already holds, mark the run
+   *  interrupted, and exit 130 (the SIGINT convention). Installed only while
+   *  the agent subprocess is running; the returned cleanup uninstalls it. */
+  private installSigintSalvage(): () => void {
+    const handler = () => {
+      void this.interruptSalvage().finally(() => process.exit(130));
```
egonSchiele: Salvaging the record is right, but `process.exit(130)` here leaves the run directory in a state the rest of the toolchain cannot read — which undercuts the reason for salvaging.

`summary.json` is written exactly once, by `writeEvalRunSummary(state, results)` **after** `runSuite`'s input loop; `initializeEvalRun` only writes `config.json`. So Ctrl-C during input 3 of 10 salvages input 3's `eval-record.json` and then exits before any summary exists. `readEvalRun` opens with `readJson(<runDir>/summary.json)`, which now throws `could not read …/summary.json` — so `agency eval grade <dir>` fails outright, and `judgeSuite` fails the same way. The salvaged evidence is orphaned: it is on disk and nothing can load it.

For a benchmark harness this is the common case, not an edge one — Ctrl-C on a 10-minute run you have already learned enough from is exactly when you want the partial results.

The fix probably is not in `runAgent`, which has no access to `state` or the results so far. Either have the interrupt path return a normal error result and let `runSuite` break the loop and write its summary (the `continueOnError: false` path already does exactly this), or have `runSuite` install the handler and write a partial summary before exiting. The second keeps `runAgent`'s never-throws contract intact and puts the signal handling at the layer that owns the run directory.

Two smaller things on the same code:

- **A second Ctrl-C during salvage kills the process mid-write.** `process.once` removes the listener when it fires, so the second SIGINT gets default behavior — and the window includes `fs.writeFileSync` of `eval-record.json`, which can leave a truncated file that then fails to parse. Re-arming with a "already salvaging, press again to force" message, or an `unref`'d guard, closes it.
- **The child is not killed explicitly.** Ctrl-C at a terminal reaches the whole foreground process group, so in practice the fork dies too — but a parent that exits 130 without `child.kill()` leaves an orphan whenever the signal arrives programmatically (`process.kill(pid, "SIGINT")`, a supervisor, a test).

packages/agency-lang/lib/config.ts
```diff
@@ -413,6 +419,7 @@ export const AgencyConfigSchema = z
         optimizeRunsDir: z.string(),
         graders: z.string().optional(),
         sourceCacheRoot: z.string().optional(),
+        limits: z.object({ wallClockSec: z.number() }).partial().optional(),
```
egonSchiele: `z.number()` accepts `0`, negatives, and fractions, and none of them mean what a reader would guess.

`limitsFromConfig` multiplies by 1000 and the value lands in `setTimeout(…, s.limits.wallClock)` (`lib/runtime/ipc.ts:1085-1089`). `setTimeout` clamps anything below 1ms to 1ms and fires on the next tick, so `"wallClockSec": 0` — the natural way to write "no limit" — makes **every** agent run fail immediately with a `wall_clock` limit failure, and `-1` does the same. That is a confusing failure to debug: the run dies instantly with a limit error while the config says the limit is off.

`z.number().int().positive()` rejects all three at config load, where the error can name the field. Worth noting the sibling CLI flags added in this PR already validate (`parsePositiveInt`, `parseNonNegativeInt`), so the config path is the looser of the two for no reason.

If "no limit" is a thing you want, it needs its own spelling — `null`, or a large sentinel — because `0` cannot mean it here.

packages/agency-lang/evals/terminal-bench-mini/graders.ts
```diff
@@ -0,0 +1,93 @@
...
+  grader(({ workdir }) => {
+    const file = join(workdir, "regex.txt");
+    if (!existsSync(file)) {
+      return fail("regex.txt was not written");
+    }
+    const pattern = readFileSync(file, "utf8").trim();
+
+    let re: RegExp;
+    try {
+      re = new RegExp(pattern, "gm");
```
egonSchiele: This compiles and runs an **agent-authored** regex inside the grading process, with no time bound.

The wall-clock limit this PR makes configurable protects the agent subprocess; grading happens afterwards, in the eval process itself. A pattern with nested quantifiers — `(a+)+`, `(\d+|\w+)*`, the kind of thing a model writes when it is flailing at a hard regex task, which is precisely the failure mode described in the PR body — backtracks catastrophically and hangs the whole run. The 25-line sample text is small, but catastrophic backtracking is about pattern shape, not input size.

The result is that a benchmark designed to measure how agents fail has an unbounded hang as one of its failure modes, and it wedges the harness rather than scoring 0.

Cheapest fix that keeps this a plain grading module: run the match in a `worker_threads` worker with a short deadline and treat a timeout as `fail("the regex did not terminate within Ns — likely catastrophic backtracking")`, which is also genuinely useful feedback for the agent. If that is more machinery than a pilot deserves, a note in the README saying grading executes untrusted patterns in-process would at least make it a known cost rather than a surprise.

packages/agency-lang/evals/terminal-bench-mini/graders.ts
```diff
@@ -0,0 +1,93 @@
+import { existsSync, readFileSync } from "fs";
```
egonSchiele, line 1: Worth checking before merge: `evals/` is a new top-level TypeScript directory in the package, and nothing scopes it out of the build.

`tsconfig.json` sets no `include`/`exclude`, so tsc picks up every `.ts` under the package root, with `"outDir": "./dist"`. That is why `dist/` already contains `scripts/` and `tests/` alongside `lib/`. So `pnpm run build` will emit `dist/evals/terminal-bench-mini/graders.js` (plus its `.d.ts`), and `package.json`'s `"files": ["./dist", …]` ships the whole `dist` tree — meaning benchmark fixtures land in the published tarball for every user of `agency-lang`.

The type-checking is a genuine benefit and worth keeping (this file uses the real `agency-lang/eval` types, so it cannot rot silently). The shipping is not. An `"exclude": ["evals"]` would lose both; a separate `tsconfig.build.json`, or moving `evals/` outside the published package root, keeps the check and drops the payload. Quickest way to confirm the concern is real: `npm pack --dry-run | grep evals` after a build.

Related: `lint:structure` is `eslint lib/`, so this file is type-checked but not structurally linted. That is probably the right call for a fixture — just noting it so the asymmetry is deliberate.

packages/agency-lang/agency.json
```diff
@@ -13,6 +13,11 @@
   "typechecker": {
     "enabled": true
   },
+  "eval": {
+    "limits": {
+      "wallClockSec": 600
```
egonSchiele, line 18: This is the repo-wide config, so the 600s ceiling applies to every eval run in this checkout — including `tests/integration/eval-run/test.mjs`, which runs in CI.

That test has no `timeout-minutes` on its workflow step (raised on #735 and still open as far as I can tell), and its whole job was previously bounded in practice by the 60s subprocess wall clock. With this change a wedged agent in that test costs 10 minutes of CI instead of 1, and if the wedge is in the parent rather than the child it is still unbounded.

Not a reason to hold this — 600s is the right number for real tool-loop agents and matches terminal-bench's easy-task budget. But it makes the missing `timeout-minutes` matter more, so the two are worth landing together.

packages/agency-lang/lib/eval/run/runAgent.test.ts
```diff
@@ -161,4 +161,23 @@ describe("runAgent", () => {
     const salvaged = JSON.parse(fs.readFileSync(path.join(run7Dir, "agent", "eval-record.json"), "utf8"));
     expect(salvaged.evalOutputs[0].value).toBe("partial work");
   });
+
+  it("a SIGINT-salvage listener is live exactly while the agent subprocess runs", async () => {
```
egonSchiele: Good test for the property most likely to rot silently — a listener leaked once per input would accumulate across a long suite and eventually trip Node's max-listeners warning.

What is not covered is the salvage itself: that `error.txt` says the run was interrupted, and that a record is written from whatever statelog existed. That is the behavior the feature exists for, and the same gap the #735 round called out for the IPC decision fix.

It is testable without signals if `interruptSalvage` (or a small `salvageAndMarkInterrupted`) is reachable — call it directly against a run dir with a half-written statelog and assert both files. That leaves only `process.exit(130)` untested, which is fine; the exit code is the least likely part to break.

egonSchiele (commented):
## Review

The pilot suite is the strongest part of this. Porting the pytest verifier into TypeScript verbatim — all 25 sample lines with a Y/N comment each, the same ordered all-or-nothing list equality, both graders `mustPass` to reproduce the 0-or-1 reward — is exactly the right fidelity call, and keeping `solution.agency` beside the task so the harness smoke-tests itself for free is a nice trick. Being explicit in the instruction that the snippet shows *how the grader evaluates*, not what language to work in, is the kind of honesty that keeps a ported benchmark meaningful.

The "what the first real runs found" section is the most valuable thing in the description: two scored-zero runs diagnosed from the run directory alone is the eval framework doing its job, and the four agent fixes each map to a link in that chain. Worth keeping that format for future benchmark PRs.

I checked `applyCliFlags` for the flag-clobbering worry — it guards every field with `!== undefined`/truthiness (`config.ts:698-706`), so an absent `--strict` cannot wipe an `agency.json` `typechecker.strict`. The comment in `scripts/agency.ts` about folding the flags onto the config being "the whole implementation" holds: `runSuite` passes `config` into `runAgent`, which compiles inside the seeded workdir with it.

Five comments inline. The first is the one I would fix before merge — the SIGINT salvage saves the record but leaves the run directory unreadable by the rest of the toolchain.

### Smaller notes not worth their own thread

**`m[1] ?? m[0]` is slightly not what the instruction promises.** The task says "when your regex has a capture group, group 1 is taken as each match's result". The grader implements "group 1 *if it participated*": a pattern with an optional group that does not match on some line silently falls back to the whole match rather than failing or yielding empty. Probably harmless for this task, but the instruction and the grader should describe the same rule since the whole point is that the agent can predict the grading.

**The second grader's `existsSync` branch is dead.** Gates run first and short-circuit, so `wrote-regex-file` already failed the input by the time `regex-matches` could see a missing file. Harmless, but it reads as though the ordering guarantee is not trusted.

packages/agency-lang/lib/eval/run/runAgent.ts
egonSchiele: Redesigned per your second option, in 7d45304e4. runAgent no longer touches signals or exits — its existing error path already salvages, which covers the interrupt case once the child dies. The runner forwards SIGINT to the child (covers the programmatic-signal orphan), so the in-flight input comes back as a normal error result; runSuite owns the signal with a once-listener that stops the loop after that input, and writeEvalRunSummary runs as always. Result: a Ctrl-C run directory now loads in eval grade and judgeSuite, and since evalRun proceeds to grading, you get scores for every completed input. Second Ctrl-C: both once-listeners are consumed by the first, so default handling force-quits — that is the "press again to force" behavior, and the interrupt message says so.

packages/agency-lang/lib/config.ts
egonSchiele: Fixed in 7d45304e4: the schema is now z.number().int().positive() with a comment explaining why 0 cannot mean "no limit" here, and limitsFromConfig ignores non-finite/non-positive values (programmatic configs bypass the schema), falling back to the 60s default. Both covered by tests. Left "no limit" without a spelling — nothing needs it yet.

packages/agency-lang/evals/terminal-bench-mini/graders.ts
egonSchiele: Fixed in 7d45304e4: the match runs in a worker_threads worker (eval: true, nothing touches disk) with a 2s deadline; timeout terminates the worker and scores 0 with feedback naming catastrophic backtracking, and compile errors surface through the worker error event. Verified both ways: the reference solution still scores 1.0 through the worker path, and (a+)+$ against a 40-char input trips the deadline cleanly.

packages/agency-lang/evals/terminal-bench-mini/graders.ts
egonSchiele: The premise is stale — tsconfig.json has include: ["lib", "scripts", "test.ts"], so evals/ was never compiled, never in dist/, never shipped (npm pack --dry-run | grep -c evals returns 0; the dist/tests you saw predates that include list). But that also meant the type-check benefit you wanted to keep did not exist: nothing checked graders.ts at all. 7d45304e4 adds tsconfig.evals.json (extends the main config, noEmit, include evals) chained into pnpm typecheck — and it immediately caught one real type error in the worker code, so the check earns its keep. lint:structure stays lib/-only, deliberately.

packages/agency-lang/agency.json
egonSchiele: timeout-minutes: 5 is already on that step — .github/workflows/test.yml:172, landed with #735. So the CI bound is the workflow step, not the subprocess wall clock, and a wedged agent costs at most 5 minutes regardless of the 600s limit. Parent-side wedges are bounded by the same step timeout.

packages/agency-lang/lib/eval/run/runAgent.test.ts
egonSchiele: With the redesign (see the thread on runAgent.ts:133) the interrupt path IS the error path, and the salvage behavior you are asking about is covered by the existing "a failed run still salvages its record to disk" test — killed child → error result → record + error.txt. The listener-lifetime test moved to runSuite, where the listener now lives: a fake runner invokes the installed listener mid-suite and asserts the loop stops after the in-flight input, summary.json is written with that input, and the listener is gone afterwards.

packages/agency-lang/foo.agency
```diff
@@ -184,8 +185,11 @@ Rewritten: "Tell me the news for today. Today's date is ${today()}."
 
 In this case, the additional context of the date helps the LLM give a more accurate response.
 
+## Agency is not JavaScript
+Every program you write must be Agency code, even when the task mentions another language — a task saying its output will be checked with JavaScript or Python still requires you to WRITE Agency. JavaScript habits that are errors in Agency: `function` declarations (use `def`), template literals with backticks, `console.log` (use `print`), and `.forEach` callbacks (use `for (item in items) { ... }`).
```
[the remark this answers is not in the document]
egonSchiele: Correct — backtick strings are valid Agency (lib/parsers/parsers.ts:192), so the claim was wrong on both lines. Fixed in 7d45304e4: backticks left the not-JavaScript list and the javascriptHint markers, and => went with them (valid in match arms). require() and String.raw — both actually observed in the failing benchmark runs, neither valid Agency — took their place.

packages/agency-lang/lib/config.ts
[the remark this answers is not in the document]
egonSchiele: Fixed in 7d45304e4 — z.number().int().positive() in the schema.

packages/agency-lang/lib/eval/run/subprocess.ts
```diff
@@ -34,7 +33,20 @@ const DEFAULT_EVAL_RUN_LIMITS: RunLimits = {
...
-export function makeSubprocessRunner(pipeAgentOutput: boolean): EvalInputRunner {
+/** The run limits a suite's config asks for: `eval.limits` overlaid on the
+ *  defaults. Only wall clock is configurable so far. */
+export function limitsFromConfig(config: AgencyConfig): RunLimits {
+  const wallClockSec = config.eval?.limits?.wallClockSec;
+  return {
+    ...DEFAULT_EVAL_RUN_LIMITS,
+    ...(wallClockSec !== undefined ? { wallClock: wallClockSec * 1000 } : {}),
+  };
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 7d45304e4 — limitsFromConfig validates finite-and-positive and falls back to the defaults otherwise, with a test covering 0, -1, NaN, and Infinity.

PR #739, opened 2026-07-31
eval inputs: tests describe the task, not the agent

egonSchiele:
Implements the input-format cleanup spec (docs/superpowers/specs/2026-07-31-eval-input-format-cleanup-design.md): eval tests describe **what the input is**; the person running the eval decides **how it reaches their agent**.

## The new input shape

`node` and `args` are deleted — both described the agent, not the test (`node` names the agent's internal structure; `args` is keyed by one particular node's parameter list). The full input vocabulary is now `id`, `task`, `goal`, `expected`, `files`, `metadata`:

```json
{
  "id": "regex-log",
  "task": "Write a regex expression that matches dates ...",
  "goal": "regex.txt matches the last YYYY-MM-DD date on lines with a valid IPv4"
}
```

- **`task` is required**: what the agent is told — an instruction string, or a JSON object for structured input. Every eval run has exactly one.
- **`goal` is unchanged**: the grading criterion, never shown to the agent, required only when the default LLM judge runs.
- `--goal "one-liner"` sets task AND goal to the same text, explicitly, at input creation — the run's `input.json` shows exactly what the agent was told; nothing resolves implicitly later.
- The loader **rejects** legacy `args`/`node` with a migration message rather than silently ignoring them.

## Delivery: exactly one parameter

The task arrives as the entry node's single positional parameter, whatever it is named (`main(task)`, `main(prompt)`, `main(x)` all work). The parameter count is checked before the node runs — 0 or 2+ parameters is an immediate, named error, never a silent `undefined`. Agents with a different natural shape add a one-parameter adapter node.

On the wire, `RunInstruction` gains an optional `task` field, present on every eval run and absent on all others — non-eval subprocess users (`std::agency.run`, hoisted calls) keep the named-`args` contract untouched. The decision lives in `resolveNodeCallArgs` in `lib/runtime/ipc.ts` (not the bootstrap, which has import-time side effects), so it is unit-testable; the bootstrap turns an error into an IPC reply.

## Consequences swept through

- The optimizer's goal-mode guard that **rejected** nodes with required parameters is deleted — under the new model `--goal` *feeds* that parameter, so the guard rejected exactly the conforming agents.
- The optimizer's reflection prompt now shows `Task:` instead of `Args:`.
- regex-log's `test.json` loses the instruction duplication that motivated this change.
- `evals-demo/` deleted (scratch from the first framework test); the eval-run integration test and ~20 unit test files migrated; CLI docs rewritten.

## Testing

- New: loader coverage (task required / object task / empty task / legacy `args`/`node` rejected / `inputFromGoal` sets both fields), the full `resolveNodeCallArgs` matrix (string, object, 0-param, 2-param, task+args mutual exclusion, named-args fallback), and a `runSuite` test observing string and object tasks reach the runner.
- 516 tests across eval, optimize, eval CLI, and IPC pass; typecheck (main + evals config), build, and `lint:structure` clean.
- No-LLM smoke: the migrated regex-log suite against the reference solution agent still scores **1.000**; the eval-run integration test (interrupting agent over IPC) passes.


packages/agency-lang/lib/cli/eval/optimize.test.ts
```diff
@@ -65,14 +65,14 @@ describe("eval optimize CLI", () => {
...
     const agentFile = writeAgent();
     const { target } = await capture({ agent: agentFile, goal: "Return Paris" });
-    expect(target?.inputs).toEqual([{ id: "input-1", node: "main", args: {}, goal: "Return Paris" }]);
+    expect(target?.inputs).toEqual([{ id: "input-1", task: "Return Paris", goal: "Return Paris" }]);
   });
 
   it("builds one input per entry from --inputs, carrying each goal first-class", async () => {
     const agentFile = writeAgent();
-    const inputsFile = writeInputs([{ id: "first", goal: "be correct", args: { text: "hi" } }]);
+    const inputsFile = writeInputs([{ id: "first", goal: "be correct", task: { text: "hi" } }]);
```
egonSchiele, line 73: this is a good example of where `goal` is useless and should just be removed

egonSchiele (commented):
## Review

This is a clean implementation of the spec and it improves on it in two places. Making `task` **required** and having `--goal` set both fields explicitly is better than the spec's `task ?? goal ?? undefined` resolution chain: the run's `input.json` now shows exactly what the agent was told, and the "one parameter but no task" hole I raised on the spec closes by construction rather than needing a rule. Pulling the decision into `resolveNodeCallArgs` in `ipc.ts` — out of the bootstrap, which has import-time side effects — and testing the full matrix there is exactly the right seam.

Other things I checked that hold up:

- No leftovers: `git grep 'input\.args\|input\.node'` across `lib/` finds nothing in eval or optimize (the remaining hits are `input.nodeDefs` in the typechecker, unrelated). The spec's mention of `mutator.ts` and `extract.ts` was over-broad — their `.args` are their own function arguments — and the PR correctly left them alone.
- The `task: ""` placeholders in `gradeRun.ts` and `judge/suite.ts` are commented as deliberately unloadable, which is the honest way to synthesize a value the loader would reject.
- The docs rewrite is good, and putting **eval entry nodes take exactly one parameter** in bold in the input-format section is right — it is a constraint on the reader's agents, not an internal detail.

Six comments inline. The first two are the ones I would want answered before merge: one is a reversal of the reviewed spec, the other trades a fail-fast check for a per-candidate runtime failure.

### One line for the PR body

`input.json`'s contents change: `args`/`node` out, `task` in. Run directories written before this keep the old shape, so reading one back gives `input.task === undefined` even though the type now says required. Old runs still grade and judge (both paths read only `goal`/`expected`/`metadata`), which is the part that matters — but #734 and #735 both listed their persisted-format changes explicitly and this one deserves the same line.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -532,14 +532,53 @@ export type RunInstruction = SubprocessIdentity & {
...
+): { args: unknown[] } | { error: string } {
+  if (msg.task === undefined) {
+    return { args: paramNames.map((p) => msg.args[p]) };
+  }
+  if (Object.keys(msg.args ?? {}).length > 0) {
+    return { error: "Malformed run instruction: both task and args were provided; they are mutually exclusive" };
+  }
+  if (paramNames.length !== 1) {
+    const detail = paramNames.length === 0
+      ? `eval delivers the input's task as the node's parameter, but node "${msg.node}" takes none`
```
egonSchiele, line 570: This reverses the spec, and I think the reversal deserves to be stated rather than absorbed.

The reviewed spec was explicit about zero parameters:

> **0 parameters** — the node is called with nothing; the task (if any) is simply not delivered. This keeps agents that need no input working (the evals-demo hello-world writer has `node main()` and ignores its input).

This PR makes that case an error, and the diff shows what it costs. `evals-demo/` — whose `writer.agency` was exactly `node main(): string` — is deleted. And `tests/integration/eval-run/test.mjs:53` changes `node main()` to `node main(task: string)`, where `task` is never read: an agent gained a parameter purely to be evaluatable.

That sits awkwardly against this PR's own thesis. "Tests describe the task; the person running the eval decides how it reaches their agent" is the line being drawn — but now a perfectly reasonable agent shape (one that needs no input: "write a hello world program", "clean up this directory") cannot be evaluated without editing the agent. The delivery convention has reached back across the line it just drew.

I can see the argument for erroring: with `task` required on every input, silently dropping it is a lie, and a silent `undefined` was the failure mode you set out to remove. That is a good argument. It is just a different design from the one that was reviewed, and the two changes are coupled — `task` became required, which is what makes zero parameters unserveable. Worth a paragraph in the PR body (and ideally an amendment to the spec) saying so, plus a note in `docs/site/cli/eval.md` that an input-free agent needs a one-parameter wrapper, since that is the shape most likely to hit this.

The alternative, if you want zero-parameter agents back: accept `paramNames.length === 0` and drop the task, but only when the node genuinely declares none — which is still not a silent `undefined`, because nothing is being passed at all.

packages/agency-lang/lib/cli/eval/optimize.ts
```diff
@@ -151,23 +146,22 @@ function provisionInputs(
 
 /** Build the optimize target: the agent plus the inputs to run it on (from --goal or --inputs). */
 export function buildTarget(opts: EvalOptimizeOptions, deps: EvalOptimizeDeps): OptimizeTarget {
-  const resolved = resolveEvalRunTarget(opts.agent);
-  if (optimizeInputSelection(opts) === "goal") return goalTarget(opts, resolved);
+  resolveEvalRunTarget(opts.agent); // validates the target early; the node is delivery-side
```
egonSchiele: Deleting `rejectGoalForNodeWithRequiredArgs` is right for the case the PR body names — under the new model `--goal` *feeds* the required parameter, so the guard rejected conforming agents. But that guard was doing two jobs, and the second one is now gone with nothing in its place.

It parsed the entry file and checked the node's parameters **before any run**. The shape mismatch it caught still exists — it just moved: `resolveNodeCallArgs` now rejects 0-parameter and 2+-parameter nodes *inside the subprocess*, which for `eval optimize` means after seeding a workdir, compiling the agent, and forking, **once per candidate**. An optimizer with 5 iterations and a mis-shaped entry node now discovers that fact 5+ times over, each time paying a full compile, and reports it as a run failure rather than a configuration error.

That is the opposite of the principle this codebase states elsewhere. `validateGraders`' own doc comment: *"a misconfigured grader should not cost a whole suite ... without this the error surfaces per input, after the whole suite has already been run and paid for."* Same situation, same fix.

The machinery is right there in the deleted function — `discoverOptimizeTargets` + `parseAgency` + count the entry node's parameters. Restoring it as an up-front "the entry node must take exactly one parameter" check (in `buildTarget`, and ideally in `runSuite` too so plain `eval run` gets it) keeps the fail-fast property while dropping the rule that was actually wrong.

Separately, on this line itself: `resolveEvalRunTarget(opts.agent);` with the result discarded reads as dead code — the comment is carrying the whole meaning. If it stays, a named `assertAgentTargetResolves(opts.agent)` would say it in the call rather than beside it. If the up-front parameter check lands, this call has a return value worth using again.

packages/agency-lang/lib/optimize/reflectionFeedback.ts
```diff
@@ -15,7 +15,7 @@ export function renderInputFeedback(entry: InputGrades, opts: ReflectionRenderOp
   const lines: string[] = [];
   const objective = inputObjective(entry.grades).toFixed(3);
   lines.push(`### Input ${entry.input.id ?? "(no id)"} — objective ${objective}${entry.gatesPassed ? "" : " (GATE FAILED)"}`);
-  lines.push(`Args: ${preview(JSON.stringify(entry.input.args), 400)}`);
+  lines.push(`Task: ${preview(JSON.stringify(entry.input.task), 400)}`);
```
egonSchiele: `JSON.stringify` made sense when this was `args` (always an object). For `task` it mostly is not, and this is an LLM-facing prompt.

A string task renders as `Task: "Write a regex expression that matches dates ..."` — with the wrapping quotes, and with any inner quotes and newlines escaped. The regex-log task in this repo is a multi-line instruction, so the proposer LLM now reads it as one long line full of `\n`, right where it is supposed to understand what the agent was asked to do.

Two lines below, `Output:` already uses `stringifyOutput(...)` for exactly this reason. Using it here too makes the two lines consistent and the prompt readable:

```ts
lines.push(`Task: ${preview(stringifyOutput(entry.input.task), 400)}`);
```

Worth checking `stringifyOutput` passes strings through raw rather than re-quoting them — if it does not, `typeof task === "string" ? task : JSON.stringify(task)` does the job.

packages/agency-lang/lib/eval/loadInputs.ts
```diff
@@ -133,11 +142,11 @@ function normalizeInput(raw: unknown, baseDir: string, makeId: MakeId, options:
...
   }
-  if (spec.args !== undefined && !isPlainObject(spec.args)) {
-    throw new Error("Eval input args must be an object when provided");
+  if (typeof spec.task !== "string" && !isPlainObject(spec.task)) {
+    throw new Error("Eval input task is required: a string, or a JSON object for structured input");
   }
-  if (spec.node !== undefined && typeof spec.node !== "string") {
-    throw new Error("Eval input node must be a string when provided");
+  if (typeof spec.task === "string" && spec.task.length === 0) {
+    throw new Error("Eval input task must not be empty");
```
egonSchiele, line 149: An empty string task is rejected here, but an empty **object** task is accepted — `isPlainObject({})` is true, so `{"task": {}}` loads and gets delivered as `{}`.

Both are equally contentless, and the reason to reject the string form (a test that tells the agent nothing is a mistake, not a test) applies identically to the object form. One more clause next to this one keeps the rule symmetric:

```ts
if (isPlainObject(spec.task) && Object.keys(spec.task).length === 0) {
  throw new Error("Eval input task must not be empty");
}
```

Minor, but the loader is where suite mistakes are supposed to be caught loudly, and this is the one hole left in "every eval run has exactly one task".

packages/agency-lang/lib/eval/runTypes.ts
```diff
@@ -1,22 +1,25 @@
...
   /** Stable identifier. Auto-derived when omitted: the loader generates one
    *  via nanoid, the optimizer derives it positionally (`input-<index>`). */
   id?: string;
-  /** What the agent should accomplish — read by the goal judge and the
-   *  pairwise judge suite. Optional; the input-file loader requires it. */
+  /** What the agent is told: an instruction string, or a JSON object for
+   *  agents that take structured data. Delivered as the entry node's single
+   *  positional parameter (eval entry nodes take exactly one). Required;
+   *  the loader rejects inputs without one. */
+  task: string | Record<string, any>;
```
egonSchiele, line 19: Worth a sentence on this field about what happens when it is read back off disk.

`input.json` inside a run directory records this spec, and `readEvalRun` parses it straight into `Input`. Run directories written before this PR have `args`/`node` and no `task`, so a grader reading `ctx.input.task` on an old run gets `undefined` while the type promises `string | Record<string, any>`.

The practical impact is small — grading and judging both read only `goal`/`expected`/`metadata`, so old runs still grade and still judge, which is the property this series has protected each time. But `Input` doubles as the on-disk schema, and the type is now stricter than the data it may be parsed from. The `outcome` field on `InterruptEntry` got exactly this treatment two PRs ago (a note saying which builds wrote which values); the same one-liner here would save someone trusting a type over a file.

packages/agency-lang/tests/integration/eval-run/test.mjs
```diff
@@ -50,13 +50,13 @@ try {
   return interrupt("confirm the thing")
 }
 
-node main() {
+node main(task: string) {
```
egonSchiele, line 53: This edit is the clearest evidence of the zero-parameter question above: the agent gains a `task: string` parameter it never reads, purely so `agency eval run` will accept it.

If the one-parameter rule stays as-is, worth a short comment on this line saying *why* the parameter exists (`// unused: eval entry nodes must take exactly one parameter`). Otherwise the next person to touch this fixture will delete it as dead and get a confusing IPC error from a test that is about interrupts, not about parameter arity.

packages/agency-lang/lib/cli/eval/optimize.test.ts
egonSchiele: The goal in that fixture is load-bearing for the test itself — it pins that a per-input goal survives loading and reaches the target (the default optimize judge grades against it). Removing it there would leave the carry-through unpinned. If you mean the broader point — that goal-as-a-field earns its keep in fewer places than it appears — I agree it is worth a look, but that is a design conversation about the default judge and judgeSuite, not something to absorb into this PR. Happy to queue it.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: The reversal was deliberate and happened during spec review — you called both moves ("This should be an error if a task was specified", then "every eval run requires exactly one task"), and the spec file was amended before implementation; the text quoted here is the first draft. But the paper trail was missing, agreed. In 90f266c5f: the spec gains a "Design history — the zero-parameter reversal" section stating what changed, why, and the accepted cost; docs/site/cli/eval.md now covers input-free agents explicitly (declare the parameter, ignore it); and the PR body gets a paragraph. On the alternative: dropping the task for a genuinely zero-parameter node is defensible, but it makes "the agent received the task" unknowable from the input alone — the strict rule keeps one invariant with no cases.

packages/agency-lang/lib/cli/eval/optimize.ts
egonSchiele: Restored in 90f266c5f, generalized past the old guard: assertEvalEntryNodeTakesOneParameter in lib/agentTarget.ts parses the entry file and checks the parameter count before any workdir is seeded or agent compiled — wired into BOTH runSuite (so plain eval run gets it) and buildTarget (so the optimizer fails once, not per candidate). Best-effort like the old guard: unparseable files and nodes defined elsewhere fall through to compile/run, which report them better. Unit-tested for 0/1/2 params, default-valued params, unparseable files, and absent nodes; the runtime check in resolveNodeCallArgs stays as the backstop. And the discarded resolveEvalRunTarget call now has a used result, which answers the dead-code read.

packages/agency-lang/lib/optimize/reflectionFeedback.ts
egonSchiele: Fixed in 90f266c5f — stringifyOutput, which passes strings through raw (typeof check at reflectionFeedback.ts:60) and JSON-stringifies objects, so the Task: and Output: lines now behave identically.

packages/agency-lang/lib/eval/loadInputs.ts
egonSchiele: Fixed in 90f266c5f — {} is rejected with the same "must not be empty" error, with a loader test.

packages/agency-lang/lib/eval/runTypes.ts
egonSchiele: Added in 90f266c5f — the task field doc now states the on-disk caveat: pre-#739 run directories carry args/node and no task, so task read back off disk may be undefined there; grading and judging read only goal/expected/metadata, so old runs still load. Same treatment as InterruptEntry.outcome.

packages/agency-lang/tests/integration/eval-run/test.mjs
egonSchiele: Comment added in 90f266c5f — and with the up-front shape check from the other thread, deleting the parameter now fails at suite start with the full explanation rather than a confusing IPC error mid-test.

PR #740, opened 2026-07-31
feat(serve): export withRuntimeConfigOverrides for per-module log routing

### LINKED ISSUE #598: serve: runtime log/observability config seam in createServeHandler (avoid compile-time baking)
## Problem

A host embedding the serve engine (e.g. the statelog app) wants each hosted run to emit its trace to a caller-specific destination — `{ observability: true, log: { host, projectId, apiKey } }` that varies per request (per project/key), not per agent source.

Today `createServeHandler(compiledPath, opts)` has no runtime log/observability seam, so the only ways to enable observability are:

1. **Bake it at compile time** (`compileSource(source, { observability, log })`) — but then the compiled artifact varies by caller identity: the apiKey lands in plaintext on disk, every distinct (host, projectId, apiKey) forces its own compile writing the *same* artifact path (write race), and per-caller cache keys thrash a per-source cache.
2. **`AGENCY_CONFIG_OVERRIDES` env var** — process-global, so concurrent requests with different configs race.

Neither is clean for a long-lived multi-request host.

## Proposal

Let `createServeHandler` accept a runtime log/observability config that is applied per invocation (threaded into the execution frame's `StatelogClient`), so **one observability-independent compiled artifact** can serve every caller with a different trace destination. This keeps the trace config as runtime data, not baked bytes — matching the hosted-agent design's "config passed into the run, never baked at compile time."

## Context

Surfaced building hosted serve in the statelog app (serve PR 2 review): compile-time baking was rejected for exactly the coupling above. Auto-observability there is blocked on this seam.

egonSchiele:
Closes #598.

## Why

The statelog host serves many compiled Agency modules from one process, and each hosted agent must auto-trace to its **own** statelog project (per-agent routing, decided with the owner). Config only reaches a run through the `RuntimeContext` constructor, and that constructor runs **once, at module-import time** (`context.ts:263-267`); `createExecutionContext` — the per-invocation path — just copies the frozen config. So there is no way to vary observability per run of an already-loaded module, and the two existing override seams (`AGENCY_CONFIG_OVERRIDES` env, `setRuntimeConfigOverrides` module-global) are process-global.

The saving grace: routing is **per-agent, not per-request**. An agent's project never changes, and each agent is a separately-cached module import. So the config only needs to be bound once, at each module's import — exactly when the existing override seam fires.

## What

`withRuntimeConfigOverrides(overrides, fn)` — a scoped helper that installs `overrides` as the active runtime config overrides, runs `fn` (which does the `import`), and restores the previous value in a `finally`. Re-exported from the public `./serve` surface next to `createServeHandler`.

This adds **no new mechanism**: it reuses the same `setRuntimeConfigOverrides` seam subprocess IPC already uses, and just wraps it in a misuse-resistant set/restore pair. The host (statelog) serializes its imports behind a mutex, since the override is a process-global — documented in the helper's doc comment.

### Caveat (documented in code)
This relies on the invariant that a `RuntimeContext` is only constructed at module import, never per-invocation. True today. A test in the statelog PR will pin the end-to-end routing behavior.

## Test
`lib/runtime/configOverrides.test.ts` — 3 new cases: installs-during/restores-after, restores the *previous* value (not just undefined), restores on throw. Full file: 16/16 pass. Build clean.


egonSchiele:
Reopening from a clean branch — the original branch accidentally included unrelated eval commits that were sitting on local main. Superseded by a new PR with only the 3 intended files.

PR #743, opened 2026-07-31
eval: command agents, per-test graders/timeouts, parallel runs, cost caps

egonSchiele:
This PR makes `agency eval run` able to benchmark the bundled `agency agent` CLI (or any Agency CLI) against a suite of terminal-bench-style tasks. It adds command agents, per-test graders and timeouts, parallel runs with a live status board, cost caps, and a batch of observability and robustness fixes that fell out of running the benchmark for real. Architecture doc: `docs/dev/eval-command-agents.md`.

## Command agents (`--agent-cmd`)

- `eval run --agent-cmd 'node agency.js agent --agent code -p -- {task}'` runs a CLI as the agent instead of compiling an `.agency` file. `EvalTarget` is now a union: file target or command target (`lib/agentTarget.ts`).
- Commands come only from the CLI flag, never from suite content. A suite pulled from git must not be able to name its own agent command — that would be remote code execution by suite. (Graders are the deliberate exception: they are test-side code, and pulling a suite means trusting it.)
- The command string is tokenized by a small tarsec grammar (`lib/eval/run/commandLine.ts`), with no shell involved. `{task}` is substituted after tokenization, per token, so a hostile task stays inert bytes inside one argv entry.
- The spawned command inherits the harness's statelog via `AGENCY_CONFIG_OVERRIDES` and a shared `AGENCY_TRACE_ID` (new env var; mint order in `context.ts` is explicit > env > fresh). The agent's own record — tool calls, cost, interrupts — lands in the eval run dir, so extraction, grading, and judging work unchanged.
- Process-group lifecycle: the command runs `detached`, and every kill (wall clock, cost cap, SIGINT/SIGTERM/SIGHUP forwarding, harness exit) is a group kill. This fixes a real bug where killing the CLI wrapper left the actual agent running and spending for another 8 minutes.
- `agency run file.agency [args...]` now passes trailing args to the entry node; extra args beyond the node's parameters are a loud error. This also fixed a direct-run codegen bug where the initial-state object landed in the first node parameter.

## Per-test graders and timeouts

- A test directory can carry its own `graders.ts` next to `test.json` (auto-discovered), so `--inputs evals/terminal-bench-mini` and `--inputs evals/terminal-bench-mini/regex-log` both work with no `--graders` flag.
- Precedence: `--graders` flag > the test's own graders > `eval.graders` config (fallback) > the goal judge. The optimizer pins graders in override mode so mutations are scored consistently.
- `test.json` can set `timeoutSec` (terminal-bench `task.toml` parity), which overrides `eval.limits.wallClockSec` for that test.

## Parallel runs, cost, and observability

- `-n/--parallel <count>` runs inputs in a bounded pool with a live status board: test name, ticking clock, status, and cost so far.
- Defensive cost cap (`eval.limits.maxCostUsd`, default $50). The fork runner bills IPC cost telemetry; the spawn runner tails the statelog. It is an accident stopper, not a budget.
- New `eval logs <runDir>` command; `eval grade` re-grades a finished run dir.
- Run IDs are timestamp-prefixed so run folders sort chronologically, and the run dir is printed at start. Grading output is colorized (zero scores red, test names green, stderr red).
- Scoring policy: killed or errored runs score 0, even if the right answer is sitting in the workdir. An agent that cannot tell it is done has not succeeded. A run record with no output event (command agents write no output) now grades with `output: null` instead of being marked ungraded.

## Agent fixes surfaced by the benchmark

- `bash`/`run` now reject a corrupted `cwd` argument (newlines, angle brackets, control characters) with a message telling the model its working directory is unchanged. This class of error — the model's own function-calling markup leaking into argument values — showed up surprisingly often.
- The coding agent's prompt gained a verification off-ramp ("check once under each reading, then submit") and a malformed-tool-call recovery rule. Before these, the agent routinely burned minutes re-verifying a correct answer.
- esbuild's "service is no longer running" crash on Ctrl-C is handled with a stop-and-retry.

## Suite

`evals/terminal-bench-mini/` gains `gcode-to-text` (grade by sha256 of the output, so the repo never contains the flag in plaintext) alongside `regex-log`. The 1.6MB `text.gcode` input is committed — it is the task input, not the answer — so the test runs anywhere the repo is cloned.

The bundled agent currently passes regex-log at 1.000 ($1.38, 7.6 min); gcode-to-text is still unsolved at a 20-minute budget.

## Testing

Unit tests across all new modules (tokenizer, spawn runner incl. a tree-kill regression test, cost tracker, pool ordering and stop behavior, grader precedence, trace-id minting, corrupted-cwd rejection), plus new integration scenarios in `tests/integration/eval-run` and a CLI arg-passthrough test.


packages/agency-lang/lib/eval/run/spawnRunner.ts
```diff
@@ -0,0 +1,203 @@
...
+      for (const t of intervals) clearInterval(t);
+      resolve(value);
+    };
+
+    // Cost cap, without an IPC channel: the child appends promptCompletion
+    // events (each carrying its call's cost) to the statelog the harness
+    // pointed it at — tail that file.
+    const costTail = makeStatelogCostTailer(args.statelogPath);
+    intervals.push(setInterval(() => {
+      if (costTail.poll() > args.maxCostUsd) {
```
egonSchiele: Once `costCapped` flips true the interval keeps firing: every second it re-sends SIGTERM to the group and pushes *another* SIGKILL timer onto `timers`. The child usually dies within a tick or two so it is not fatal, but a child that ignores SIGTERM gets a new 5s SIGKILL timer per second until `close` arrives.

Suggest tripping once:

```ts
intervals.push(setInterval(() => {
  if (costCapped) return;
  const total = costTail.poll();
  if (total <= args.maxCostUsd) return;
  costCapped = true;
  costCappedTotal = total;
  killTree("SIGTERM");
  timers.push(setTimeout(() => killTree("SIGKILL"), KILL_GRACE_MS));
}, 1_000));
```

That also removes the double `costTail.poll()` call.

packages/agency-lang/lib/eval/run/spawnRunner.ts
```diff
@@ -0,0 +1,203 @@
...
+    // The detached tree must never outlive this process: forward the
+    // terminating signals as group kills, and reap on ANY exit (normal,
+    // crash, uncaught throw). The one hole no supervisor can close is
+    // SIGKILL of this process — nothing runs then; the tree's remaining
+    // protection is EPIPE on its next write to our closed pipes.
+    const forwardSigint = () => killTree("SIGINT");
+    const forwardSigterm = () => killTree("SIGTERM");
+    const forwardSighup = () => killTree("SIGHUP");
+    const reapOnExit = () => killTree("SIGKILL");
+    process.once("SIGINT", forwardSigint);
```
egonSchiele: Four `process` listeners are registered per in-flight command. Under `-n 10` or more that is 40+ listeners on `exit`, so Node prints a `MaxListenersExceededWarning` — straight into the middle of the status board's in-place repaint, which then smears (the board counts lines, and it did not count the warning).

Worth a module-level supervisor: register the four handlers once, keep a set of live pids, and have each run add/remove itself. Same behaviour, constant listener count, no warning.

packages/agency-lang/lib/eval/run/spawnRunner.ts
```diff
@@ -0,0 +1,203 @@
...
+        : err.message;
+      settle({ ok: false, errorMessage: detail });
+    });
+
+    child.on("close", (code, signal) => {
+      if (costCapped) {
+        settle({
+          ok: false,
+          errorMessage:
+            `cost cap exceeded: $${costCappedTotal.toFixed(2)} spent, cap $${args.maxCostUsd.toFixed(2)} ` +
```
egonSchiele: This message is a character-for-character copy of `makeCostCapTracker().exceededMessage()` in subprocess.ts. Two spellings of the same sentence in the two runners is exactly what that helper exists to prevent — if the config key is ever renamed, one of these will be missed. Either use the tracker here (`add()` the polled delta) or at least import its message builder.

packages/agency-lang/lib/eval/run/spawnRunner.ts
```diff
@@ -0,0 +1,203 @@
...
+      child.kill(signal);
+    }
+  };
+
+  return new Promise((resolve) => {
+    let settled = false;
+    let timedOut = false;
+    let costCapped = false;
+    let costCappedTotal = 0;
+    let stdoutBytes = 0;
```
egonSchiele: Nit: `stdoutBytes` also accumulates stderr bytes, so `args.limits.stdout` is really a combined-output cap. The name will mislead the next reader into thinking stderr is uncapped. `pipedBytes` would say what it is.

packages/agency-lang/lib/eval/run/costTail.ts
```diff
@@ -0,0 +1,52 @@
...
+      let stats: fs.Stats;
+      try {
+        stats = fs.statSync(statelogPath);
+      } catch {
+        return total;   // no statelog yet
+      }
+      if (stats.size <= offset) return total;
+      const fd = fs.openSync(statelogPath, "r");
+      const buf = Buffer.alloc(stats.size - offset);
+      fs.readSync(fd, buf, 0, buf.length, offset);
```
egonSchiele: `fs.readSync`'s return value is dropped while `offset` advances to `stats.size` unconditionally. On a short read the un-filled tail of `buf` is zero bytes, which join into `remainder` and corrupt the next line boundary — and the skipped bytes are never re-read, so the cost total silently undercounts (this feeds a *safety* cap, so undercounting is the wrong direction to err).

Short reads on a local regular file are rare but not impossible. Either loop until `bytesRead` covers the range, or just `fs.readFileSync(fd)`-style with an explicit slice of what was actually read:

```ts
const { bytesRead } = ... // or loop
const text = buf.subarray(0, bytesRead).toString("utf8");
offset += bytesRead;
```

packages/agency-lang/lib/eval/run/statusBoard.ts
```diff
@@ -0,0 +1,82 @@
...
+    const cost = entry.costUsd > 0 ? `$${entry.costUsd.toFixed(2)}` : "";
+    return (
+      `  ${cell(entry.id, 30, paint?.green)} ${cell(entry.status, 8, statusColor)} ` +
+      `${cell(elapsed, 8)} ${cost}`
+    ).trimEnd();
+  };
+
+  const render = () => {
+    if (isTty && painted) {
+      process.stderr.write(`\x1b[${entries.length}A`);
```
egonSchiele, line 79: The cursor-up repaint assumes one terminal row per entry. A long input id (ids come from directory names, and the id cell is padded to 30) in a narrow terminal wraps, so `\x1b[${entries.length}A` climbs fewer rows than were printed and the board smears down the screen.

Cheap fix: truncate `line(entry)` to `process.stderr.columns ?? 80` before writing.

packages/agency-lang/lib/cli/eval/graders.ts
```diff
@@ -1,23 +1,31 @@
...
   }
-  if (gradersPath === undefined) {
-    return [new LlmJudge({ name: "goal" })];
+  if (gradersFlag !== undefined) {
+    return { mode: "override", graders: await loadGradingModule(gradersFlag, config) };
   }
-  return loadGradingModule(gradersPath, config);
+  const configPath = config.eval?.graders;
+  if (configPath !== undefined) {
+    return { mode: "fallback", graders: await loadGradingModule(configPath, config) };
```
egonSchiele: Behaviour regression worth a second look: before this PR, an `eval.graders` config module that default-exported `[]` threw ("The grading module at X exported no graders"). Now it becomes `{ mode: "fallback", graders: [] }`, and the empty-set guards in `evalRun`/`evalGrade` only fire for `mode: "override"`. So every input is graded with zero graders → `gradeInput` returns `grades: [], gatesPassed: true` → objective 0.000, exit code 0, no diagnostic. A misconfigured grading module now reads as "the agent failed".

The fallback set is only ever empty via this branch (the goal-judge default is never empty), so the old throw can be kept precisely here.

packages/agency-lang/lib/eval/grading/gradeRun.ts
```diff
@@ -5,36 +5,50 @@ import { readEvalRun, type ReadEvalRunInput } from "@/eval/readRun.js";
...
+import { loadGradingModule } from "./gradingModule.js";
 import { Scorecard, type GraderGrade, type InputGrades } from "./scorecard.js";
 import type { LoadedRun, JSON as Json } from "./types.js";
 
+/** The suite-level grader set and what it means against a test's own graders
+ *  (an input's recorded `graders` module). "override" (an explicit --graders
+ *  flag) replaces every test's own graders — the experiment knob. "fallback"
+ *  (eval.graders config, or the bundled goal judge) applies only to inputs
+ *  that carry none. */
+export type SuiteGraders =
```
egonSchiele: Both arms of this union carry the identical payload, so the discriminant is doing all the work and the union is doing none. `{ mode: "override" | "fallback"; graders: BaseGrader[] }` reads the same and drops the machinery.

egonSchiele (commented):
## Overview

Makes `agency eval run` able to benchmark a CLI (`--agent-cmd`) instead of only a compiled `.agency` file, and adds the machinery that fell out of running the benchmark for real: per-test graders and timeouts, a parallel worker pool with a status board, a defensive cost cap, `eval logs`, and a batch of agent/runtime fixes.

The core design choices are good and well argued in the code:

- **Commands come only from the flag, never from suite content.** The comment at `resolveEvalTarget` states the threat (a remote git suite naming its own command is remote code execution) and the deliberate exception for graders. That is the right line.
- **Tokenize, then substitute per token.** No shell anywhere, so a hostile task stays inert bytes inside one argv entry. The tarsec grammar is small and the "why not `between(...)`" note about the 14-second pathological failure is the kind of comment that saves the next person an afternoon.
- **Process-group kill on every path** (wall clock, cost cap, SIGINT/SIGTERM/SIGHUP, harness exit), with the one hole named honestly (SIGKILL of the harness).
- **The argv codegen fix** is correct; I compiled `node main(task: string)` on this branch and confirmed the generated call is `main(__process.argv[2], initialState)` against `async function main(task, { messages, callbacks } = {})`.
- **The corrupted-cwd rejection** is a small change with a real behavioural payoff, and the message correctly tells the model the filesystem is fine.

Test coverage is solid: tokenizer, tree-kill regression, pool ordering/stop, grader precedence, trace-id minting, corrupted cwd, plus CLI integration.

Findings below, most consequential first. Nothing here blocks the design; two are behavioural and worth fixing before merge.

## Worth fixing before merge

1. **An empty `eval.graders` config module now scores 0 silently instead of erroring** (`lib/cli/eval/graders.ts:28`). This is the one real regression I found — a misconfiguration now reads as an agent failure.
2. **The gcode-to-text fixture is gitignored**, so the new test is a 20-minute money-burning no-op for everyone else, with no diagnostic.
3. **The cost-cap interval keeps re-killing every second** after it trips (`spawnRunner.ts:134`).

## Smaller things

4. Four `process` listeners per in-flight command → `MaxListenersExceededWarning` at `-n 10`+, which also corrupts the status board repaint.
5. The cost-cap message is duplicated verbatim between the fork and spawn runners, despite `exceededMessage()` existing for exactly this.
6. `costTail` ignores `readSync`'s return value — a short read undercounts spend, and the cap is a safety mechanism.
7. `statusBoard`'s cursor-up repaint assumes one row per entry; a wrapped line smears the board.
8. `SuiteGraders`' two union arms carry identical payloads.

## One cross-cutting note (no inline anchor)

`ttyColor` keys off `process.stdout.isTTY`, but `runSuite`'s progress lines, the status board, and the run-dir banner all write to **stderr**. So `eval run 2>run.log` writes ANSI escapes into the log file, and `eval run > out.txt` on a terminal drops color from the progress lines that are still on screen. `spawnRunner` already gets this right by checking `process.stderr.isTTY` directly for the child's stderr. A `ttyErrColor` (or a `colorFor(stream)` helper) in `lib/utils/termcolors.ts` would make the two consistent — worth doing since this PR is what puts color on stderr for the first time.

## Not findings, just noted

- `resolveGradersFile` does not constrain the path to the suite directory the way `resolveFilesDir` does. Given graders are executed code and the suite is trusted by virtue of being run, that is coherent — but the asymmetry with `resolveFilesDir` right above it will read as an oversight to a future maintainer. One sentence at the call site pointing at the trust argument would prevent someone "fixing" it.
- `formatGraderSummary` paints any non-perfect binary count red, so 4/5 looks like 0/5 at a glance. Probably what you want for a benchmark; flagging in case it is not.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -4695,6 +4695,22 @@ export class TypeScriptBuilder {
...
+      // initialState is the OPTIONS argument after them. Before this,
+      // initialState was always the first argument: harmless when mains took
+      // no parameters (it landed in the options slot), wrong once entry
+      // nodes take parameters (#739) — the state object arrived as the
+      // first parameter's value. Absent argv entries are undefined, so
+      // parameter defaults apply.
+      const mainParamCount =
+        this.compilationUnit.graphNodes.find((n) => n.nodeName === "main")
+          ?.parameters.length ?? 0;
+      const mainCallArgs = [
```
egonSchiele, line 4708: The fix itself is right — I compiled `node main(task: string)` on this branch and the generated call is now `main(__process.argv[2], initialState)` against `async function main(task, { messages, callbacks } = {})`.

Two things the mapping cannot see, worth at least a doc line in `docs/site/cli/run.md`:

- argv entries are always strings, so `node main(n: number)` receives `"3"` with no coercion or check — arithmetic on it silently concatenates.
- extra trailing args are dropped without a word, and missing ones become `undefined` for parameters typed non-optional.

A loud error on "more args than parameters" would catch the common typo cheaply.

packages/agency-lang/evals/terminal-bench-mini/gcode-to-text/test.json
```diff
@@ -0,0 +1,5 @@
+{
+  "goal": "out.txt contains exactly the text the G-code print head traces, nothing else.",
+  "task": "I have provided a file called text.gcode in your current working directory. This is a gcode file for a Prusa MK4s. I'm trying to print some text onto an existing object. When I run the print, what will the text show? Write the output to out.txt in your current working directory.",
+  "timeoutSec": 1200
```
egonSchiele, line 4: `files/text.gcode` is gitignored, so on every machine but yours this test loads with no fixture directory at all — the agent gets the task text, cannot find `text.gcode`, and burns the full 1200s (and real money) before scoring 0. The README documents the fixture as "vendored", which reads as "it is in the repo".

The input file is not the answer (the answer is only the sha256), so committing it is not a leak — 1.6MB is the only real cost. If it stays out, please make the failure fast and legible: a `mustPass` gate that checks the fixture exists before the agent runs, or a loader-side skip, plus a line in the README saying the test is a no-op without the file.

packages/agency-lang/lib/eval/run/spawnRunner.ts
egonSchiele: Fixed in 55951f1f5, with your shape plus the shared tracker: the interval polls once, bills the delta into makeCostCapTracker, and on trip sets costCapped, sends one SIGTERM, and schedules one SIGKILL timer. Subsequent ticks return early.

packages/agency-lang/lib/eval/run/spawnRunner.ts
```diff
@@ -0,0 +1,203 @@
...
+    // pointed it at — tail that file.
+    const costTail = makeStatelogCostTailer(args.statelogPath);
+    intervals.push(setInterval(() => {
+      if (costTail.poll() > args.maxCostUsd) {
+        costCapped = true;
+        costCappedTotal = costTail.poll();
+        killTree("SIGTERM");
+        timers.push(setTimeout(() => killTree("SIGKILL"), KILL_GRACE_MS));
+      }
+    }, 1_000));
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 55951f1f5 — single poll per tick, and the trip happens at most once (no repeated SIGKILL timers).

packages/agency-lang/lib/eval/run/spawnRunner.ts
egonSchiele: Fixed in 55951f1f5. superviseTree registers the four handlers once at module level and keeps an array of live killTree functions; each run adds itself on spawn and removes itself on settle. Added a regression test: listener count stays constant across 12 concurrent runs.

packages/agency-lang/lib/eval/run/spawnRunner.ts
egonSchiele: Fixed in 55951f1f5 — the spawn runner now bills the polled deltas into makeCostCapTracker and settles with its exceededMessage(), so the sentence has one spelling.

packages/agency-lang/lib/eval/run/spawnRunner.ts
egonSchiele: Renamed to pipedBytes in 55951f1f5, with a comment noting args.limits.stdout acts as a combined-output forwarding cap here.

packages/agency-lang/lib/eval/run/costTail.ts
egonSchiele: Fixed in 55951f1f5 — offset advances by the actual bytesRead and only buf.subarray(0, bytesRead) is parsed; a short read tail is picked up on the next poll instead of being skipped.

packages/agency-lang/lib/eval/run/statusBoard.ts
egonSchiele: Fixed in 55951f1f5. Rather than slicing the final string (ANSI escapes would make the cut point wrong), each row is built cell by cell against a width budget of process.stderr.columns, truncating the plain text before colorizing. Rows can no longer wrap, so the cursor-up count stays correct.

packages/agency-lang/lib/eval/run/statusBoard.ts
```diff
@@ -0,0 +1,82 @@
...
+  if (totalSec < 60) return `${totalSec}s`;
+  return `${Math.floor(totalSec / 60)}m ${totalSec % 60}s`;
+}
+
+export function startStatusBoard(ids: string[]): {
+  update(id: string, patch: Partial<Omit<BoardEntry, "id">>): void;
+  stop(): void;
+} {
+  const entries: BoardEntry[] = ids.map((id) => ({ id, status: "queued", costUsd: 0 }));
+  const byId: Record<string, BoardEntry> = Object.fromEntries(entries.map((e) => [e.id, e]));
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 55951f1f5 — byId is now a null-prototype object (same for the grader-module cache in gradeRun.ts).

packages/agency-lang/lib/cli/eval/graders.ts
egonSchiele: Restored in 55951f1f5, in the config branch of resolveGraders as suggested: an eval.graders module that exports no graders throws "exported no graders" again. The stale comment in run.ts that argued the empty set was fine is updated too.

packages/agency-lang/lib/eval/grading/gradeRun.ts
egonSchiele: Collapsed in 55951f1f5 to { mode: "override" | "fallback"; graders: BaseGrader[] }.

packages/agency-lang/lib/eval/grading/gradeRun.ts
```diff
@@ -86,12 +100,44 @@ export function validateGraders(graders: BaseGrader[], input: Input | undefined)
...
+    return cache(input.graders);
+  }
+  return ctx.suiteGraders.graders;
+}
+
+/** One esbuild+import per module path per call, however many inputs share a
+ *  grading module. Exported so the run CLI's pre-run validation loads
+ *  through the same path grading does. */
+export function makeGraderModuleCache(config: AgencyConfig): (modulePath: string) => Promise<BaseGrader[]> {
+  const loads: Record<string, Promise<BaseGrader[]>> = {};
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 55951f1f5 — the cache is a null-prototype object now.

packages/agency-lang/lib/eval/runArtifacts.ts
```diff
@@ -1,7 +1,10 @@
 import * as fs from "fs";
 import * as path from "path";
+import { fileURLToPath } from "url";
 
 import { sha256Text } from "@/utils/hash.js";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
 import { agentRunPaths } from "./run/extract.js";
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 55951f1f5 — imports are contiguous at the top; the __dirname computation moved inline into harnessVersion, its only use.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Both done in 55951f1f5. run.md now documents that argv values are always strings (with the concatenation example), that missing args arrive as undefined, and that extra args are an error. The loud error is implemented in the direct-run emit: more args than main has parameters prints which arguments were extra and exits 1. Verified end to end: a 1-param main run with 3 args exits 1 with "main() takes 1 argument(s) but got 3; extra: extra1 extra2". Fixtures regenerated.

packages/agency-lang/evals/terminal-bench-mini/gcode-to-text/test.json
egonSchiele: Committed the file in 55951f1f5 and dropped the gitignore entry. The gcode is the task input, not the answer — the answer exists only as the sha256 in graders.ts — so the test now runs anywhere the repo is cloned, and the README wording (vendored) is accurate again.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -4764,3 +4781,38 @@ export class TypeScriptBuilder {
     return result;
   }
 }
+
```
egonSchiele: This change is not very readable. When you have this much code, it's better to put it in a typestashe template instead See lib/templates/ for examples and see https://github.com/egonSchiele/typestache for documentation of how to use typestache.

Also, this file is very large already. Notice that your change spans lines 4784 to 4818. We want to try to add not add stuff to it if possible. So if it's an option, the best option would be to actually put this in a separate file. That way you could also put it in a function with a name that is something useful and descriptive. Notice that the way you have it set up right now, it's pretty hard to see what this is used for, and what situations it gets used. We already have the pattern of putting stuff in the lib/backends/typescriptBuilder/ directory. Think about if you can put something in there instead.

packages/agency-lang/lib/cli/runBundledAgent.ts
```diff
@@ -136,8 +137,22 @@ export function runBundledAgent(
   const overrides = agentConfigOverride(args);
   const agentHome = agentHomeOverride(args);
   const env = { ...process.env };
-  if (Object.keys(overrides).length > 0) {
-    env[CONFIG_OVERRIDES_ENV] = serializeConfigOverrides(overrides);
+  // Merge onto any inherited overrides rather than replacing them ("env
```
egonSchiele: Are long comments like this necessary? You're in the habit of adding super long comments. They're often a smell. If you have to add comments this long, maybe it means that the code is not very readable. Think about whether you could structure the code in a different way instead to make it more readable. Check out docs/dev/anti-patterns.md. Are you writing declarative interfaces that encapsulate imperative complexity? Are you making it easy to follow what the code is doing at a glance?

packages/agency-lang/lib/eval/run/costTail.ts
```diff
@@ -0,0 +1,56 @@
...
+      const bytesRead = fs.readSync(fd, buf, 0, buf.length, offset);
+      fs.closeSync(fd);
+      offset += bytesRead;
+      const chunks = (remainder + buf.subarray(0, bytesRead).toString("utf8")).split("\n");
+      remainder = chunks.pop() ?? "";
+      for (const line of chunks) {
+        if (line.trim() === "") continue;
+        let event: { data?: { type?: string; cost?: { totalCost?: unknown } } };
+        try {
+          event = JSON.parse(line);
```
egonSchiele: is this performant? Looks like you're trying to scan each line one at a time and parse it as JSON. Would this scale? The code you have written here is very specifically for one task only and it doesn't seem like it would be easy to extend if we wanted to extract any other data in the future. What would a more extensible solution look like? How could you instead write a library or a module that you could use to extract data from the state log in a declarative manner? and don't things like that already exist in this code base anyways?

packages/agency-lang/lib/eval/run/runAgent.ts
```diff
@@ -141,14 +162,38 @@ class AgentRunner {
...
-      makeSubprocessRunner(this.options.pipeOutput ?? true, limitsFromConfig(this.options.config));
+    const pipeOutput = this.options.pipeOutput ?? true;
+    const limits = limitsFromConfig(this.options.config, this.options.timeoutSec);
+    const maxCostUsd = costCapFromConfig(this.options.config);
+    const runner = this.deps.runner ?? ((job: EvalRunnerJob) =>
+      job.kind === "command"
+        ? runCommandInSpawn({ ...job, pipeOutput, limits, maxCostUsd })
+        : makeSubprocessRunner(pipeOutput, limits, maxCostUsd)(job));
+
+    if (this.target.kind === "command") {
```
egonSchiele, line 183: Often when you have if statements with long bodies, it's saying that that body could be put in a separate function. Ask yourself – you're defining a really long function here – is all this code necessary? Is this function very clearly doing one job? Or is it doing the job of multiple different things?

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Reworked in 37e829f71 using both suggestions. The generated code now lives in a typestache template (lib/templates/backends/typescriptGenerator/directRunArgsCheck.mustache) so it reads as the JavaScript it emits, and the builder function moved to lib/backends/typescriptBuilder/directRunArgsCheck.ts with a doc comment saying exactly when the guard fires. typescriptBuilder.ts now just imports and calls it.

packages/agency-lang/lib/cli/runBundledAgent.ts
egonSchiele: You were right that the comment was covering for structure. In 37e829f71 the merge moved into mergeConfigOverrides in lib/config.ts, next to readConfigOverrides/serializeConfigOverrides where the override semantics live. The call site is now one line, and the function doc is four lines instead of nine.

packages/agency-lang/lib/eval/run/costTail.ts
egonSchiele: It did already exist — twice, and I had duplicated both. Fixed in 37e829f71: the log viewer's follow mode already does incremental append-reading (its core is now lib/statelog/appendReader.ts, shared by both), and parseStatelogJsonl already parses statelog lines (moved to lib/statelog/parse.ts, mirroring the wireTypes move; the viewer path re-exports). lib/statelog/tail.ts composes them into a pull-based event tailer — poll() returns the events appended since the last poll — that any future consumer can filter declaratively. costTail is now a ten-line sum over promptCompletion events. On performance: each line is read and JSON-parsed exactly once across the run, at the poll cadence, which is the floor for JSONL.

packages/agency-lang/lib/eval/run/runAgent.ts
egonSchiele: Split in 37e829f71: execute() is now three lines of shape (mkdir, build the job, run it), with buildJob() answering "what runs" per target kind and makeRunner() answering "how it runs" (injected fake, or fork/spawn under limits). The try/catch covers only the step that can throw.

PR #744, opened 2026-08-01
build: stop shipping compiled tests in the npm tarball

egonSchiele:
## What

Publishing this package reported 3070 files. Most of that was compiled tests.

`tsconfig.json` excluded `tests/`, `lib/tui/test`, and the stdlib tests, but not the 645 test files that live next to their source in `lib/`. So `tsc` emitted a `.js` and a `.d.ts` for each one into `dist/`, and the tarball packs `./dist` wholesale.

| | before | after |
|---|---|---|
| files | 3070 | **1942** |
| packed | 4.1 MB | **3.3 MB** |
| unpacked | 21.2 MB | **16.6 MB** |

Of the 3070, 1124 were `*.test.js` / `*.test.d.ts` (7.7 MB), plus ~70 more under `__tests__/` directories.

## Changes

**1. `tsconfig.json`** — the two narrow stdlib-only test excludes are replaced by `**/*.test.ts` and `**/__tests__/**`.

**2. `tsconfig.tests.json`** (new) — that exclude also removed the tests from `pnpm typecheck`, which runs `tsc --noEmit` against the same config, and vitest transpiles without checking types. A type error in a test would have gone unnoticed. This follows the existing `tsconfig.evals.json` pattern: `noEmit`, and it carries the old exclude list verbatim so it checks exactly what the build used to. `typecheck` now runs all three configs.

**3. `package.json` `main`** — pointed at `lib/index.js`, but `lib/` is not in the `files` list, so that path was not in the tarball at all. The `exports` map already pointed at `dist`, so modern resolvers were unaffected, but anything falling back to `main` (older bundlers, some tooling) resolved to a missing file. Now `./dist/lib/index.js`.

## Verification

- `make` from a clean `dist/`: exit 0
- `npm pack --dry-run`: the numbers above; **0** remaining `*.test.js`, `*.test.d.ts`, or `__tests__` entries
- All 15 `main` / `types` / `bin` / `exports` targets exist on disk
- `node dist/scripts/agency.js --version` works; `require()` of `dist/lib/index.js` and `dist/lib/runtime/index.js` both succeed
- `pnpm typecheck` (all three configs): exit 0
- `vitest run scripts/agency.test.ts lib/runtime/__tests__/node.test.ts`: 4 files passed — the CLI entry point plus a file inside a newly-excluded `__tests__` directory, which is where breakage would show

I did not run the full unit suite or the Agency suite locally; CI covers those.

## Not included

The working tree had a pending `0.10.0` -> `0.11.0` version bump and a CHANGELOG edit. Those are unrelated release work, so they are left out of this branch.

## Follow-up, not done here

1942 is still not small. The remainder is 706 `.js` + 706 `.d.ts` for the source modules and 188 generated `stdlib/docs` markdown files. Cutting further means bundling, or rolling declarations up to just the seven public entry points — a bigger change than this one.

### DIFF
--- a/packages/agency-lang/CHANGELOG.md
+++ b/packages/agency-lang/CHANGELOG.md
@@ -1,3 +1,30 @@
+## Jul 31 2026 — v0.11.0
+
+### Eval framework
+- **Graders moved out of `optimize` and into eval** — `agency eval run --graders` now scores a run instead of only reporting that the process exited.
+- **Command agents** — `eval run --agent-cmd '<command with {task}>'` runs any Agency CLI command as the agent. Mostly useful for testing the Agency agent.
+- **Workdirs are seeded, not cloned** — Each test specifies the file set it needs, and only those files are cloned into the workdir. This is a massive improvement, and takes the working directory's size down from an average of 1gb to less than 1mb.
+- **Breaking — tests describe the task, not the agent.** Earlier, each test would describe the node to run and the args to run it with, which felt like it was describing the shape that the agent needed to be in. Tests now just describe the task and the person who wrote the agent decides what node the test should aim at.
+- **Parallel runs and cost control** — `-n/--parallel` runs tests in a bounded pool with a nice progress display. Tests can now set their own `timeoutSec`, and every run has a $50 default cost cap.
+- **Observability** — new `agency eval logs <runDir>`, the run directory is printed at the start of every run, run ids sort chronologically, grading output is colorized, and a sequential run prints regular progress updates of time taken.
+- Killed or errored runs score 0 even when the right answer is on disk.
+- Killing a run with Ctrl-C still generates a summary file.
+- The pairwise optimize loop is removed, and eval output now defaults to the node's return value.
+
+### Language / Typechecker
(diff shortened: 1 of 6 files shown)


egonSchiele:
I measured this before changing it, and the suggestion does not have the effect it describes, so I have left the config as is.

The premise is that narrowing `include` to test entry points avoids re-checking production code. It does not. `include` only picks the *root* files; `tsc` type-checks every file in the resulting program, and files pulled in through imports are in the program. Since the tests import nearly all of `lib/`, that code gets checked either way.

Measured with `tsc --listFiles`, counting files in this package:

| `include` | files type-checked |
|---|---|
| `["lib", "scripts"]` (current) | 1271 |
| test globs only (suggested) | 1187 |

So narrowing drops 84 files, 6.6%. The other 93% of the "duplicate" work happens regardless. Wall time confirms it — two runs each, current `7.7s / 9.9s`, narrowed `13.2s / 8.5s`: noise, no measurable win.

Against that, a glob list has a real cost. It has to be kept in sync by hand, and a test added under a pattern nobody remembered to list would be silently unchecked. That is the exact failure this config exists to prevent.

The mechanism that *would* remove the duplicate work is project references: set `composite: true` on the main config and have the tests project reference it, so the tests consume the emitted `.d.ts` instead of re-checking sources. `composite` is currently commented out in `tsconfig.json`. That is a real option but a much larger change than this PR, and it interacts with the incremental build setup, so I would rather not fold it in here.

PR #747, opened 2026-08-01
logs viewer: timeline views (flame / by-name / occurrences / detail), component architecture, working follow mode

egonSchiele:
Timeline views for the logs viewer, the component-view architecture, and a working follow mode. Spec: `docs/superpowers/specs/2026-07-31-timeline-views-design.md` (v2.1); prototype that settled every behavior decision lives on branch `adit/proto-timeline`.

## The timeline views

Press `t` in `agency logs` or `agency eval logs`: tree → flame → by-name → tree.

- **Flame view** — one row per call, indented, on a shared time axis. Labels say what each call was doing: LLM rows show the question asked plus tokens/cost (never the model name), tool rows show their argument (the bash command, the file path). Enter/`→` drills in (re-roots with breadcrumbs), `←` climbs out, `+`/`-`/`[`/`]`/`0` zoom and pan, `/` searches, `o` jumps back to the tree at that span.
- **By-name view** — one row per group with count, total **self-time**, and share of the run. LLM calls group by thread label → enclosing function → model. Enter opens the occurrences view: every call of that group, chronological, each with where it came from (shared path lifted into the header).
- **Detail screen** — Enter on a leaf, or `d` anywhere (tree included): full prompt transcript, model, tokens, cost, or the complete tool payload. Scrollable, `y` copies.

Load-bearing semantics (each is pinned by a named regression test):
- **Self-time** (envelope minus direct children): without it the top-level llmCall span, which wraps the whole tool loop, was credited 193% of a run. By-name totals use self-time only, and a cross-view test asserts flame and by-name can never disagree.
- **Busyness shading** — a bar cell's glyph (`░▒▓█`) is the fraction of that time slice actually spent, never a call count; any call in view paints at least one `░`. Shares may exceed 100% for parallel (forked) work — real compute time, not the nesting bug.
- **Thread labels are per-process** — thread ids restart in subprocesses (one trace had two `threadCreated` events for id "1"), so label lookup is scoped to the nearest `subprocessRun` span.
- Bookkeeping spans (`handlerChain`, `threadEndHooks`) are hidden by default (`a` shows) — purely presentational, depths and self-time untouched. Running spans (no terminus yet; `promptCancelled` counts as one) draw to the window end capped with `⋯`.

## The component architecture

Every view — tree included — now implements one `View` type (`handleKey(ev, viewport)` returning a `ViewAction`, `render`, `setData`, `helpLines`, `notify`), composed of component classes with separate compute and render methods. The shell (`run.ts`) shrinks to a view stack plus action dispatch, the `?` overlay, the parse-error footer, and follow. The pure timeline kernel (`lib/logsViewer/timeline/`: intervals, spans, groups) has no TUI dependency and is the piece the planned cross-run analysis will reuse.

File moves, no behavior change: `render.ts` → `treeRows.ts` (neutral row model shared by search and the tree view), `input.ts` → `views/treeReducer.ts`, and `summary.ts`'s private text helpers (span naming, `fmtDuration`, quote-stripping, threshold colors) lifted into `spanText.ts` so a span reads identically in every view. Every pre-existing viewer test kept its assertions — only import paths and construction changed.

Two deliberate spec divergences, flagged in the migration commit: `ViewerState` stays exported from `types.ts` and `expandAncestorsOf` keeps its signature — the actual blocker (search depending on `render.ts`) is fixed by the `treeRows` split, and hiding the type bought nothing.

## Follow mode: diagnosed, then fixed

Follow has never worked. Red tests against the old wiring confirmed two structural defects: the **boot gap** (the CLI pre-read the file, the watcher started at the file's *current* size — anything in between was never parsed; the new e2e test reproduces it deterministically) and the **accumulator rewind** (toggling `f` off/on re-seeded from the boot snapshot, dropping appends). Both die by construction: the shell owns ONE `makeAppendReader(path, 0)` whose first read IS the boot read, and `f` toggles polling only — the reader and its offset persist. Truncation/rotation resets cleanly; every view's `setData` keeps cursor/drill/zoom by id and absolute time. The append reader's UTF-8-split hazard, previously covered only by `follow.test.ts`, now has direct tests in `lib/statelog/appendReader.test.ts`.

## Testing

~180 new tests: kernel (interval edge cases incl. cell-boundary and clamp-to-1; the 193% regression; per-process thread labels incl. nested subprocesses; the follow re-group race), per-view component tests driven by keys against rendered text, snapshots against a checked-in trimmed real benchmark statelog (`timeline/fixtures/bench.jsonl`), shell stack/dispatch tests (including "Ctrl+F pages, does not toggle follow"), and the three follow e2e regressions. Full viewer + statelog suites: 371 tests green; tsc and `lint:structure` clean.

Docs: new `docs/dev/logs-viewer.md` architecture page, timeline section in `docs/site/guide/observability.md`, pointer in `docs/site/cli/eval.md`, CLAUDE.md index entry.


egonSchiele (commented):
Reviewed against the spec and plan reviews. Verified locally at `afb0103` — `npx vitest run lib/logsViewer lib/statelog` gives **371 passed / 32 files**, matching the description.

This is strong work. The things I flagged as gaps in the plan review are addressed with named tests rather than quietly dropped: `subtract` with no pieces, cell-boundary fractions, the clamp-to-1 that encodes the busyness decision, `selfIntervals` content (not just the total), a span given as the root, the phantom-indentation case, `promptCancelled` as a terminus, nested subprocesses, the late-`threadCreated` re-group race, the prefix cut landing on a segment boundary, post-wrap scroll clamping, and `crossView.test.ts` pinning that flame and by-name can never disagree. `splitWidth` and the one-row invariant test close the width hole. The "guard — cannot fail today" label on the admin test is exactly the right way to keep a tautological test honest.

The `spanText.ts` extraction is the right call and better than what I suggested: lifting `spanDetail`/`fmtDuration`/`stripQuotes`/`durationColor` out of `summary.ts` means a span reads identically in the tree and the timeline, and `spanDisplayName` builds on it rather than beside it. The two spec divergences are correctly reasoned — the real blocker was search depending on `render.ts`, and `treeRows.ts` fixes that; hiding `ViewerState` bought nothing.

Follow mode is fixed structurally rather than patched: one reader whose first `read()` is the boot read, `f` toggling polling only. Both defects die by construction, which is the right shape.

Nothing below blocks merge. The performance notes are the ones I would act on before this meets a large statelog, since `derive()` runs on every follow append and every `a` toggle.

**Themes**

1. **`derive()` does the expensive work two to four times over.** `groupSpans` is called twice with identical arguments, `findNode` is a full DFS per row, and `threadLabelFor` scans its whole scope subtree per LLM call. Individually fine at 678 lines; multiplied together they make `derive()` roughly O(n²) on the one path that runs on every append.
2. **Transient messages never clear in the timeline views.** The tree keeps the old convention (any keystroke clears `messageBar`, via `handleKeyEx`); flame/by-name/occurrences set `this.message` and nothing ever unsets it, so "no matches for X" or "copied" sits in the footer for the rest of the session.
3. **Two comments claim more than the code does** — the ░ floor and the narrow-terminal degradation. Both are worth correcting in place, because the next reader will trust them.

packages/agency-lang/lib/logsViewer/views/flameView.ts
```diff
@@ -0,0 +1,410 @@
...
+      return;
+    }
+    const cursorId = this.selected()?.span.id;
+    this.drillPath = this.drillPath.filter((id) => findNode([trace], id) !== undefined);
+    const rootNode = this.drillPath.length > 0
+      ? findNode([trace], this.drillPath[this.drillPath.length - 1])!
+      : trace;
+    const spans = timelineSpans(rootNode, { hideKinds: this.hideAdmin ? ADMIN_KINDS : [] });
+    const colorByKey = rankColors(groupSpans(spans, trace));
+    const groups = groupSpans(spans, trace);
```
egonSchiele: `groupSpans(spans, trace)` is called twice with identical arguments, one line apart — line 179 feeds `rankColors`, line 180 builds `keyBySpanId`. It is not a cheap call: each one walks the whole tree to build the index, then runs `keyOf` per span, and `keyOf` for an LLM call runs `threadLabelFor`, which scans its entire scope subtree.

`derive()` runs on every follow append, every `a` toggle, and every drill, so this doubles the most expensive thing the view does on its hottest path.

```ts
const groups = groupSpans(spans, trace);
const colorByKey = rankColors(groups);
```

packages/agency-lang/lib/logsViewer/views/flameView.ts
```diff
@@ -0,0 +1,410 @@
...
+    const keyBySpanId: Record<string, string> = {};
+    for (const g of groups) {
+      for (const id of g.spanIds) keyBySpanId[id] = g.key;
+    }
+    this.rows = spans.map((span) => ({
+      span,
+      node: findNode([trace], span.id)!,
```
egonSchiele: `findNode([trace], span.id)` inside the `.map` is a full DFS per row, so building `this.rows` is O(n²) in span count. `groups.ts` already builds exactly the index this needs (`indexTree` → `byId`), and `groupSpans` has just built one on the line above.

Exporting the index (or having `groupSpans` return it alongside the groups) would make this a lookup and remove the last quadratic step from `derive()`. Same applies to the `findNode` calls at lines 174/176 and in `headerText`.

packages/agency-lang/lib/logsViewer/views/flameView.ts
```diff
@@ -0,0 +1,410 @@
...
+    start = Math.max(full.start, Math.min(start, full.end - span));
+    this.zoom = { start, end: start + span };
+  }
+
+  private applySearch(text: string): void {
+    const query = text.trim().toLowerCase();
+    this.query = query.length > 0 ? query : undefined;
+    if (this.query === undefined) return;
+    if (!this.jumpMatch(1, true)) {
+      this.message = `no matches for "${text.trim()}"`;
```
egonSchiele, line 281: `this.message` is set here and by `notify()`, and nothing ever clears it. The tree view still honours the old convention — `handleKeyEx` clears `messageBar` on any keystroke (`treeReducer.ts`) — so the two halves of the viewer now behave differently: in the tree, "copied 312 bytes" disappears on the next key; in flame, `no matches for "xyz"` stays in the footer for the rest of the session, including after a successful search.

Cheapest fix that keeps the convention: clear `this.message` at the top of `handleKey`. Same in `byNameView` and `occurrencesView`.

packages/agency-lang/lib/logsViewer/views/shared.ts
```diff
@@ -0,0 +1,147 @@
...
+  }
+  return { gutter, bar, stats };
+}
+
+/** Shaded cells for a set of intervals. The glyph is the fraction of the
+ *  cell's time slice the intervals occupy — busyness, never overlap count:
+ *  · none, ░ ≤25%, ▒ ≤50%, ▓ ≤90%, █ above. Floor override: any interval
+ *  inside the window paints at least one ░ (losing a 60ms call on a
+ *  20-minute axis is worse than overstating it). A running bar extends to
+ *  the window end and is capped with ⋯ so it cannot read as finished. */
```
egonSchiele, line 59: The comment says the floor exists so a 60ms call on a 20-minute axis is not lost, but `glyphFor` already handles that case: `coverage` returns a small positive fraction for any interval that touches a cell, and `glyphFor` returns `·` only for `fraction <= 0`. A 60ms call is saved by the threshold table, not by the floor.

What the floor actually rescues is **zero-width** intervals, which contribute exactly 0 coverage — and `shared.test.ts` pins precisely that ("a zero-width span inside a wide window renders exactly one ░"), while the 60ms test would pass with the floor removed.

Worth saying so in the comment. As written, someone reading it could conclude the `fraction <= 0` test in `glyphFor` is redundant and remove it, which would silently drop zero-width spans.

packages/agency-lang/lib/logsViewer/views/shared.ts
```diff
@@ -0,0 +1,147 @@
...
+  let stats = view === "byName" ? LAYOUT.byNameStats : LAYOUT.flameStats;
+  let bar = cols - gutter - stats;
+  if (bar < LAYOUT.minBarCells) {
+    gutter = Math.max(20, cols - stats - LAYOUT.minBarCells);
+    bar = cols - gutter - stats;
+  }
+  if (bar < LAYOUT.minBarCells) {
+    stats = Math.max(8, cols - gutter - LAYOUT.minBarCells);
+    bar = Math.max(LAYOUT.minBarCells, cols - gutter - stats);
+  }
```
egonSchiele, line 41: The floors here (gutter 20, stats 8, bar `minBarCells` 10) sum to 38, so on any terminal narrower than 38 columns `splitWidth` returns a split wider than `cols` — for `cols = 20` it returns `{gutter: 20, bar: 10, stats: 8}`. Rows then exceed the screen width, and since the repaint math depends on one row being exactly one terminal row, that is the invariant `shared.test.ts` protects everywhere else.

The doc comment on `LAYOUT` describes degradation as fully handled, which is true down to 38 columns and not below. Either clamp the result to `cols` (accepting a sub-floor bar) or state the 38-column minimum in the comment.

packages/agency-lang/lib/logsViewer/timeline/groups.ts
```diff
@@ -0,0 +1,167 @@
...
+  const scope = nearestAncestor(node, index, (a) => a.label === "subprocessRun")
+    ?? rootOf(node, index);
+  let label: string | undefined;
+  const scan = (n: TreeNode) => {
+    if (n !== scope && n.nodeKind === "span" && n.label === "subprocessRun") return;
+    const d = n.event?.data;
+    if (d?.type === "threadCreated" && String(d.threadId) === String(threadId)
+        && typeof d.label === "string" && d.label.length > 0) {
+      label = d.label;
+    }
```
egonSchiele, line 136: Two notes on `threadLabelFor`:

**Last match wins.** `label` is assigned on every match without breaking out, so when a process legitimately reuses a thread id (thread created, ended, id reused), the label comes from the *last* `threadCreated` in DFS order rather than the one in effect when the call was made. First-match-before-the-call would be more defensible. Not obviously wrong — worth a sentence either way, since the per-process scoping right above it is documented so carefully.

**Cost.** `scan(scope)` walks the entire scope subtree once per LLM span. With `k` LLM calls in an `n`-node trace that is O(k·n), inside `groupSpans`, which `flameView.derive()` currently calls twice per re-derive. Hoisting the scan to one pass per scope (build `threadId → label` per process subtree once, reuse for every call in it) would collapse it.

packages/agency-lang/lib/logsViewer/timeline/spans.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+}
+
+function makeSpan(node: TreeNode, depth: number): TimelineSpan | undefined {
+  const extent = spanExtent(node);
+  if (extent === undefined) return undefined;
+  const childExtents = node.children
+    .filter((c) => c.nodeKind === "span")
+    .map(spanExtent)
+    .filter((e): e is Interval => e !== undefined);
```
egonSchiele, line 74: `spanExtent` walks the full descendant subtree, and it is called once for the node plus once per direct span child — so a node's subtree is walked again for each child, and `isRunning` walks it a third time. Overall `timelineSpans` is about O(n · depth) with three passes per span.

Fine at the current statelog sizes, and `derive()` is the only caller — but `derive()` runs on every follow append. One post-order pass computing extent and the running counts together, memoised per node id, would make it linear. Worth doing whenever the cross-run project starts feeding this bigger inputs, since it is the piece that is meant to be reused headlessly.

packages/agency-lang/lib/logsViewer/run.ts
```diff
@@ -36,12 +39,16 @@ export type RunViewerOpts = {
...
   initialFollow?: boolean;
+  // Watcher poll interval; tests inject a small one.
+  followIntervalMs?: number;
   thresholds?: ViewerThresholds;
 };
 
 export async function runViewer(opts: RunViewerOpts): Promise<void> {
-  const parsed = parseStatelogJsonl(opts.jsonl);
-  const roots = buildForest(parsed.events);
+  const watcher = makeFollowWatcher(opts, (text) => onNewText(text));
```
egonSchiele: `makeFollowWatcher` is handed a closure over `onNewText`, which is declared with `const` at line 76, and that closure calls `render`, declared at line 103. Both are safe today only because `onText` is never invoked during construction and `setInterval` cannot fire synchronously — so nothing reaches the temporal dead zone.

It is a thin margin for a shell this central: any future change that makes the watcher emit its boot read through `onText` (a natural-looking simplification, given the boot read already happens inside `makeFollowWatcher`) turns into a `ReferenceError` at startup. Declaring `onNewText`/`render` above the `makeFollowWatcher` call, or passing them in explicitly, removes the ordering hazard.

packages/agency-lang/lib/logsViewer/run.ts
```diff
@@ -57,312 +64,180 @@ export async function runViewer(opts: RunViewerOpts): Promise<void> {
   }
```
egonSchiele, line 67: Under `--follow` against a file that is empty or has not been written yet, the boot read returns `""`, `roots` is empty, and the viewer prints "No events found.", waits for one key, and exits — the one case where following is most useful is the case it refuses.

`currentFileSize` already returns 0 for a missing file, so the watcher handles this fine; it is only this early return that does not. Suggest: when `followPath` is set and `initialFollow` is on, render the empty state and keep polling instead of returning. (Pre-existing behavior, not a regression from this PR — but this PR is what makes it reachable in a useful scenario.)

packages/agency-lang/lib/logsViewer/run.ts
```diff
@@ -57,312 +64,180 @@ export async function runViewer(opts: RunViewerOpts): Promise<void> {
...
-    if (!followerState) return;
-    followerState.f.stop();
-    followerState = undefined;
+  const toggleFollow = (): void => {
+    if (!opts.followPath) {
+      stack.active().notify("follow unavailable when reading from stdin");
+      return;
+    }
+    followOn = !followOn;
+    treeView.setFollowIndicator(followOn);
```
egonSchiele: `setFollowIndicator` is TreeView-only, so `[FOLLOW]` appears in the tree's status bar and nowhere else. Toggle `f` while in flame or by-name and the only feedback is the transient `notify` message — which in those views never clears (see the `flameView.ts` comment), so it is simultaneously the only indicator and a permanently stale one.

Since timeline views under follow are a headline use case here (watching a live run), consider surfacing the follow state in `TimelineHeader` alongside `[admin spans shown]`.

packages/agency-lang/lib/logsViewer/views/flameView.ts
egonSchiele: Fixed in 92e751321 — derive() now calls groupSpans once and feeds both rankColors and the key map.

packages/agency-lang/lib/logsViewer/views/flameView.ts
egonSchiele: Fixed in 92e751321. buildTreeIndex is exported from groups.ts; derive() builds it once, groupSpans accepts it, and every row/crumb lookup is a byId access. The same index removes the findNode calls in occurrences and its per-member DFS context paths (now a parentIds walk-up).

packages/agency-lang/lib/logsViewer/views/flameView.ts
egonSchiele: Fixed in 92e751321 — all three bar views clear the message at the top of handleKey, matching the tree convention, with a regression test (search miss shows, next key clears).

packages/agency-lang/lib/logsViewer/views/shared.ts
egonSchiele: You were right — the floor rescues zero-width intervals only; positive-length calls are saved by the threshold table. The comment now says exactly that, including that the fraction <= 0 test in glyphFor is not redundant.

packages/agency-lang/lib/logsViewer/views/shared.ts
egonSchiele: Fixed in 92e751321 — splitWidth clamps below the 38-column floor sum (a sub-floor bar beats a wrapped row), and a sweep test asserts gutter+bar+stats <= cols for 10..200 columns across all three views.

packages/agency-lang/lib/logsViewer/timeline/groups.ts
egonSchiele: Both addressed in 92e751321. Labels now come from a per-scope threadId→label map built once per process subtree and cached across the whole groupSpans pass — one scan per scope instead of per llm call. Last-created-wins is kept and now documented at scanScopeLabels: id reuse within one process is rare enough that positional resolution has not been worth the bookkeeping; the comment says to revisit if a real log disagrees.

packages/agency-lang/lib/logsViewer/timeline/spans.ts
egonSchiele: Deferred as you framed it — a comment at spanExtent now records the O(n·depth) shape and the single-memoised-post-order-pass fix, keyed to when the cross-run project starts feeding bigger inputs.

packages/agency-lang/lib/logsViewer/run.ts
egonSchiele: Fixed in 92e751321 — the watcher takes no callback at construction; start(onText) binds it, so nothing can reach the shell before it finishes wiring. A comment states the hazard.

packages/agency-lang/lib/logsViewer/run.ts
egonSchiele: Fixed in 92e751321 — with followPath set and initialFollow on, an empty (or not yet created) file renders the empty state and keeps polling instead of exiting.

packages/agency-lang/lib/logsViewer/run.ts
egonSchiele: Fixed in 92e751321 — setFollowIndicator joined the View type; toggling updates every view on the stack, newly pushed views inherit the flag, and the timeline headers show [following] next to [admin spans shown].

packages/agency-lang/lib/logsViewer/views/flameView.ts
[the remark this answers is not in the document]
egonSchiele: Fixed in 92e751321 (single groupSpans call; see the thread above).

packages/agency-lang/lib/logsViewer/views/occurrencesView.ts
```diff
@@ -0,0 +1,249 @@
...
+    if (sel === undefined) return this.message;
+    const base = `${sel.node.summary}  ·  start +${fmtDuration(sel.span.extent.start - window.start, { minutes: true })}` +
+      `  self ${fmtDuration(sel.span.selfMs, { minutes: true })}`;
+    return this.message ? `${base}  ${this.message}` : base;
+  }
+
+  private renderRow(item: Occurrence, isCursor: boolean, window: Interval, widths: { gutter: number; bar: number; stats: number }): Element {
+    const index = this.occ.indexOf(item) + 1;
+    const detail = new RowLabel(item.node).computeText();
+    const text = `#${String(index).padStart(2)} ${item.contextTail} · ${detail}`;
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 92e751321 — row numbers are precomputed at derive time; renderRow no longer scans.

packages/agency-lang/lib/logsViewer/views/detailScreen.ts
```diff
@@ -0,0 +1,162 @@
...
+    const node = this.node!;
+    const out: string[] = [];
+    const extent = spanExtent(node);
+    const spans = timelineSpans(node, { hideKinds: [] });
+    if (extent !== undefined) {
+      const self = spans.length > 0 ? spans[0].selfMs : 0;
+      out.push(
+        `start +${fmtOffset(0)}   duration ${fmtDuration(extent.end - extent.start, { minutes: true })}` +
+        `   self ${fmtDuration(self, { minutes: true })}`,
+      );
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 92e751321 — a leaf event's self-time is its envelope, with a test pinning that the detail header never reads self 0ms for a timed leaf.

packages/agency-lang/lib/logsViewer/timeline/groups.ts
```diff
@@ -0,0 +1,167 @@
...
+   *  (two forked branches busy 10s each in a 10s window = 2.0) — that is
+   *  true compute time, distinct from the nesting double-count that
+   *  self-time prevents. */
+  share: number;
+  models: string[];
+};
+
+export function groupSpans(spans: TimelineSpan[], root: TreeNode): SpanGroup[] {
+  const index = indexTree(root);
+  const byKey: Record<string, SpanGroup> = {};
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 92e751321 — byKey, the tree index, the scope label cache, and the per-view span maps are all null-prototype now.

PR #749, opened 2026-08-01
fix: node parameter defaults compile; goto may omit defaulted args

egonSchiele:
A node parameter default parsed but vanished at compile time: `node main(name: string = "fallback")` compiled to `async function main(name, …)` with the default nowhere in the output, so an omitted argument arrived as `undefined`. Found while building eval-run argument passthrough.

Three changes:
- **The node body applies the default** when the incoming `__state.data` slot is `undefined` — the one site every invocation path funnels through (exported wrapper, direct-run argv, `goto` transitions), so defaults behave identically everywhere.
- **The exported wrapper carries the default in its TypeScript signature** (`async function main(name = `fallback`, …)`), so TS callers can omit the argument.
- **A `goto` that omits a defaulted argument no longer crashes the compiler.** It previously emitted `undefined` into the IR and the printer threw; nobody hit it because defaults never worked. The emitted data object now omits missing keys so the body default fires.

New agency execution test (`tests/agency/node-param-defaults`) pins all four paths: omitted arg → default, explicit arg wins, `goto` without the arg → default, `goto` with the arg passes through. Fixtures regenerated with zero churn (no existing fixture uses node defaults); builder tests, tsc, and `lint:structure` green.

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -3049,7 +3049,13 @@ export class TypeScriptBuilder {
     if (targetNode && targetNode.parameters.length > 0) {
       const entries: Record<string, TsNode> = {};
       targetNode.parameters.forEach((p, i) => {
-        entries[p.name] = argNodes[i];
+        // A goto may omit trailing arguments (legal when the parameter
+        // has a default): leave the key out of data so the node body
+        // sees undefined and applies its default. Emitting argNodes[i]
+        // blindly put undefined into the IR and crashed the printer.
+        if (argNodes[i] !== undefined) {
+          entries[p.name] = argNodes[i];
+        }
       });
       dataNode = ts.obj(entries);
     } else if (argNodes.length > 0) {
@@ -3148,14 +3154,23 @@ export class TypeScriptBuilder {
       ...hoistedAliases,
     ];
 
-    // Param assignments (only when not resuming)
+    // Param assignments (only when not resuming). A parameter default
+    // applies whenever the incoming data slot is undefined — this is the
+    // ONE site every invocation path funnels through (the exported
+    // wrapper, direct runs, and graph transitions all deliver arguments
+    // via __state.data), so defaults work identically for all of them.
     if (parameters.length > 0) {
-      const paramStmts = parameters.map((p) =>
-        ts.assign(
-          $(ts.stack("args")).index(ts.str(p.name)).done(),
-          $(ts.runtime.state).prop("data").prop(p.name).done(),
-        ),
-      );
+      const paramStmts = parameters.map((p) => {
+        const incoming = $(ts.runtime.state).prop("data").prop(p.name).done();
+        const value = p.defaultValue
+          ? ts.ternary(
+              ts.binOp(incoming, "===", ts.id("undefined")),
+              this.processNode(p.defaultValue),
+              incoming,
+            )
+          : incoming;
+        return ts.assign($(ts.stack("args")).index(ts.str(p.name)).done(), value);
+      });
       stmts.push(ts.if(ts.raw("!__state.isResume"), ts.statements(paramStmts)));
     }
 
@@ -4602,6 +4617,25 @@ export class TypeScriptBuilder {
     return nodes;
   }
 
+  /** The exported node wrapper's parameter list: the node's own
+   *  parameters (with their defaults, so TS callers can omit them) plus
+   *  the trailing options object. */
+  private nodeWrapperParams(
+    args: FunctionParameter[],
+  ): { name: string; typeAnnotation?: string; defaultValue?: TsNode }[] {
+    const params = args.map((arg) => ({
+      name: arg.name,
+      typeAnnotation: arg.typeHint ? formatTypeHintTs(arg.typeHint) : "any",
+      defaultValue: arg.defaultValue ? this.processNode(arg.defaultValue) : undefined,
+    }));
+    params.push({
+      name: "{ messages, callbacks }",
+      typeAnnotation: "{ messages?: any; callbacks?: any }",
+      defaultValue: ts.obj({}),
+    });
+    return params;
+  }
+
   private postprocess(): TsNode[] {
     const result: TsNode[] = [];
 
@@ -4635,24 +4669,7 @@ export class TypeScriptBuilder {
 
     for (const node of this.compilationUnit.graphNodes) {
       const args = node.parameters;
-      const fnParams: {
-        name: string;
-        typeAnnotation?: string;
-        defaultValue?: TsNode;
-      }[] = [];
-      if (args.length > 0) {
-        for (const arg of args) {
-          const typeHint = arg.typeHint
-            ? formatTypeHintTs(arg.typeHint)
-            : "any";
-          fnParams.push({ name: arg.name, typeAnnotation: typeHint });
-        }
-      }
-      fnParams.push({
-        name: "{ messages, callbacks }",
-        typeAnnotation: "{ messages?: any; callbacks?: any }",
-        defaultValue: ts.obj({}),
-      });
+      const fnParams = this.nodeWrapperParams(args);
       const dataObj: Record<string, TsNode> = {};
       for (const arg of args) {
         dataObj[arg.name] = ts.id(arg.name);
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -145,6 +145,7 @@ import type {
 import type { CompilationUnit } from "../compilationUnit.js";
 import { SourceMapBuilder } from "./sourceMap.js";
 import { directRunExtraArgsCheck } from "./typescriptBuilder/directRunArgsCheck.js";
+import { nodeWrapperParams } from "./typescriptBuilder/nodeWrapperParams.js";
 import { ScopeManager } from "./typescriptBuilder/scopeManager.js";
 import { StepPathTracker } from "./typescriptBuilder/stepPathTracker.js";
 import { NameClassifier } from "./typescriptBuilder/nameClassifier.js";
@@ -3049,10 +3050,12 @@ export class TypeScriptBuilder {
     if (targetNode && targetNode.parameters.length > 0) {
       const entries: Record<string, TsNode> = {};
       targetNode.parameters.forEach((p, i) => {
-        // A goto may omit trailing arguments (legal when the parameter
-        // has a default): leave the key out of data so the node body
-        // sees undefined and applies its default. Emitting argNodes[i]
-        // blindly put undefined into the IR and crashed the printer.
+        // A goto may omit trailing arguments: leave the key out of data
+        // so the node body sees undefined and applies its default.
+        // (Emitting argNodes[i] blindly put undefined into the IR and
+        // crashed the printer.) Only TRAILING omissions can reach here —
+        // AG6039 rejects a required parameter after a defaulted one, so
+        // a missing slot always belongs to a defaulted parameter.
         if (argNodes[i] !== undefined) {
           entries[p.name] = argNodes[i];
         }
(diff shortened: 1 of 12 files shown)


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -4602,6 +4617,25 @@ export class TypeScriptBuilder {
     return nodes;
   }
 
+  /** The exported node wrapper's parameter list: the node's own
+   *  parameters (with their defaults, so TS callers can omit them) plus
+   *  the trailing options object. */
+  private nodeWrapperParams(
```
egonSchiele: put these in a separate file... we *just* talked about this

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -3049,7 +3049,13 @@ export class TypeScriptBuilder {
     if (targetNode && targetNode.parameters.length > 0) {
       const entries: Record<string, TsNode> = {};
       targetNode.parameters.forEach((p, i) => {
-        entries[p.name] = argNodes[i];
+        // A goto may omit trailing arguments (legal when the parameter
+        // has a default): leave the key out of data so the node body
+        // sees undefined and applies its default. Emitting argNodes[i]
+        // blindly put undefined into the IR and crashed the printer.
+        if (argNodes[i] !== undefined) {
```
egonSchiele, line 3059: This turns a compiler crash into silence for a program that is genuinely malformed.

`node target(a: string = "x", b: string)` compiles with no diagnostic today — I checked. Then `goto target("only")` reaches here with `argNodes = ["only"]`, so `entries` gets `a` and omits `b`, and the emitted output is:

```js
runner2.halt(goToNode("target", { ..., data: { a: `only` } }));
// and in target:
__stack.args["a"] = __state.data.a === void 0 ? `x` : __state.data.a;
__stack.args["b"] = __state.data.b;   // undefined
```

`a + b` evaluates to `"onlyundefined"`. Per the description, that same program previously hit the printer throw, so it failed at compile time; now it compiles and produces a wrong answer at runtime.

`AG6016` does not catch it: it fires for `node target(a: string, b: string)` with one argument ("Expected 2 argument(s) for 'target', but got 1"), so the minimum is the count of non-defaulted parameters. That is the right minimum for keyword arguments and the wrong one for positional binding — the supplied argument fills `a`, and required `b` is left empty.

The cheap fix is the TypeScript rule: reject a required parameter that follows a defaulted one, at parse or typecheck. That makes the omission here provably safe (only trailing arguments can be missing) and this guard becomes exactly what the comment says it is.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -3148,14 +3154,23 @@ export class TypeScriptBuilder {
...
-        ts.assign(
-          $(ts.stack("args")).index(ts.str(p.name)).done(),
-          $(ts.runtime.state).prop("data").prop(p.name).done(),
-        ),
-      );
+      const paramStmts = parameters.map((p) => {
+        const incoming = $(ts.runtime.state).prop("data").prop(p.name).done();
+        const value = p.defaultValue
+          ? ts.ternary(
+              ts.binOp(incoming, "===", ts.id("undefined")),
```
egonSchiele, line 3180: The codebase already has a sentinel convention for this exact problem — `def` function parameters compile to `defaultValue: ts.id("__UNSET")` in the signature with `typeAnnotation: \`${baseType} | typeof __UNSET\`` (~line 2473), and the body does `param === __UNSET ? default : param` (~line 2292). This uses `undefined` instead.

I think `undefined` is the right call for nodes: arguments arrive through `__state.data`, which crosses a serialization boundary (checkpoints, resume), and a module-level sentinel would not survive the round trip — whereas a missing object key reads as `undefined` naturally. But the divergence is worth a sentence here, because a reader who knows the `def` convention will assume nodes match it.

The reason it is not purely cosmetic: the two forms have different semantics at the edge. `__UNSET` distinguishes "argument omitted" from "argument explicitly `undefined`"; `undefined` cannot. So `greet(undefined)` on a `def` keeps `undefined`, while `withDefault(undefined)` on a node gets `"fallback"`. That is defensible — it matches how JS default parameters behave — but it is a real difference between two things a user reasonably expects to behave the same, and nothing currently records the decision or tests it.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -4602,6 +4617,25 @@ export class TypeScriptBuilder {
...
+  /** The exported node wrapper's parameter list: the node's own
+   *  parameters (with their defaults, so TS callers can omit them) plus
+   *  the trailing options object. */
+  private nodeWrapperParams(
+    args: FunctionParameter[],
+  ): { name: string; typeAnnotation?: string; defaultValue?: TsNode }[] {
+    const params = args.map((arg) => ({
+      name: arg.name,
+      typeAnnotation: arg.typeHint ? formatTypeHintTs(arg.typeHint) : "any",
+      defaultValue: arg.defaultValue ? this.processNode(arg.defaultValue) : undefined,
```
egonSchiele: Worth being explicit about what this buys, because it interacts with the body default in a way that affects which test protects which change.

With the default in the wrapper signature, JS applies it before `data` is built, so on the wrapper path `__state.data.name` is already `"fallback"` and the body ternary never fires. Which means the first test case — "an omitted argument takes the default", `input: ""` — passes through this line alone. `viaGoto` is what covers the body site, i.e. the "ONE site every invocation path funnels through" that the description leads with. Both are tested, but not by the tests one would guess from their names.

Also: this emits the parsed default expression directly into a parameter list. That is safe only because parameter defaults are literal-only (I checked: `= 1 + 2`, `= pick()`, and bare identifiers all fail to parse). If defaults were ever relaxed to allow calls, the emitted default would become an `await` expression inside an async function's parameter list, which is a syntax error. One line of comment here would keep that invariant visible.

packages/agency-lang/tests/agency/node-param-defaults.test.json
```diff
@@ -0,0 +1,32 @@
+{
+  "tests": [
+    {
+      "nodeName": "withDefault",
+      "description": "an omitted argument takes the default",
+      "input": "",
```
egonSchiele, line 6: Note that this case exercises the *wrapper* default, not the body default: the evaluate template calls `await withDefault()` directly, so JS fills `name` before `__state.data` is built and the body ternary never runs. `viaGoto` below is the case that covers the body site.

Not a problem — both paths are covered — but the descriptions read as though this one tests the headline change, and it does not.

Cases worth adding: an explicit `null` passed to the defaulted parameter (differs from `def` semantics, see the builder comment); an array or object literal default, which are the only non-scalar shapes the parser accepts; a node with several parameters where only some are defaulted; and the direct-run argv path, the one path in the description with no test.

egonSchiele (commented):
Small, well-scoped fix with a clear story. I verified the surrounding behavior by compiling probe files rather than reading alone; notes below say which claims are verified and which are questions.

The three-part split is right, and the `nodeWrapperParams` extraction is a genuine simplification (the old `if (args.length > 0)` guard was doing nothing a `.map` doesn't). The test covers all four paths named in the description.

**One thing I would fix before merging** — see the inline comment on the `goto` hunk. A defaulted parameter followed by a required one (`node target(a: string = "x", b: string)`) compiles today with no diagnostic, and after this change `goto target("only")` silently drops `b`. I confirmed the emitted output: `data: { a: \`only\` }` with no `b` key, and `__stack.args["b"] = __state.data.b`, so the body computes `"onlyundefined"`. By the description's own account that program previously hit the printer throw, so it failed loudly at compile time and now compiles quietly. The printer crash was acting as an accidental guard, and removing it leaves nothing in its place.

Worth noting `AG6016` *does* fire for the all-required case — `goto target("only")` against `node target(a: string, b: string)` gives "Expected 2 argument(s) for 'target', but got 1." So the arity check treats the count of non-defaulted parameters as the minimum, which is correct for a language with keyword arguments but not for positional binding: the single supplied argument lands on `a`, and required `b` gets nothing. TypeScript solves this with a parse-time rule (a required parameter may not follow an optional one), and that seems like the right shape here too.

**Verified while reviewing** (recording it because it makes one part of this change safe in a way that is not obvious):

Parameter defaults are literal-only. `= 1 + 2`, `= pick()`, and a bare identifier all fail to parse, for both `def` and `node`; `= []` and `= { a: 1 }` parse. That is what makes it safe to emit the default expression into the wrapper's parameter list — `await` is a syntax error in an async function's parameter list, so a default that compiled to an Agency call would break the generated code. Worth a sentence in `nodeWrapperParams` saying it relies on that, since nothing near the code says so and relaxing the parser later would break it silently.

**Test cases I would add**

- An explicit `null` or `undefined` passed to a defaulted parameter (see the semantics note on the body hunk — the answer differs from `def` functions).
- An array or object literal default, the only non-scalar shapes the parser accepts.
- A node with several parameters where only some are defaulted.
- The direct-run argv path. The description names three paths that funnel through the body site; the test covers the wrapper and `goto`.
- A compile-error test for required-after-defaulted, if that diagnostic gets added.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Moved in f7da28292 — lib/backends/typescriptBuilder/nodeWrapperParams.ts, with processNode passed as a callback. No excuse; the pattern was fresh.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -4602,6 +4617,25 @@ export class TypeScriptBuilder {
...
+    }));
```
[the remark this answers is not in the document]
egonSchiele: Safe today, and now documented rather than assumed: parameter defaults are literal-only, pinned by the AST type itself (defaultValue?: Literal | AgencyArray | AgencyObject, lib/types/function.ts), so processNode can never emit a scoped access here. The new file's doc comment states the invariant and says exactly what to switch to (optional params resolved in the body) if defaults are ever relaxed.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Fixed in f7da28292 with the rule you suggested, as new diagnostic AG6039: a parameter without a default may not follow one with a default — defs and nodes both, with an agency-explain entry and regenerated diagnostics docs. Interesting wrinkle the test surfaced: the def GRAMMAR already rejects the shape at parse time; only nodes let it through, which is why this was reachable at all. The test pins both. The goto-site comment now says omissions are provably trailing, which is what makes the key-omission guard exactly what it claims to be.

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Documented in f7da28292 at the body site: undefined rather than __UNSET because node arguments cross a serialization boundary (checkpoints/resume) where a module-level sentinel would not survive, with the edge-semantic consequence stated — explicit undefined takes the default (JS-style), unlike a def, and null never does. The null case is now a test: goto withDefault(null) must produce "hi null".

packages/agency-lang/lib/backends/typescriptBuilder.ts
egonSchiele: Both notes taken in f7da28292: the wrapper/body interplay (wrapper default fills data first, so the body ternary exists for goto and direct data feeds) is documented at both sites, and the literal-only invariant is stated in the new file with the AST type as the authority.

packages/agency-lang/tests/agency/node-param-defaults.test.json
egonSchiele: Fixed in f7da28292: descriptions now name the mechanism each case pins (wrapper-signature path vs body/__state.data path), and the suite grows the cases you listed — explicit null (default must not fire), an array-literal default, mixed required+defaulted parameters, and the direct-run argv path as Test 8b in the CLI integration test (run with no arg → default; with arg → passed through).

PR #750, opened 2026-08-01
Accept the syntax variations agents actually write

egonSchiele:
Makes the parser accept three syntax variations models reach for out of habit, each normalized back to canonical form by `agency fmt`.

Spec: `packages/agency-lang/docs/superpowers/specs/2026-07-31-agent-friendly-syntax-variations-design.md`

## What is accepted

| Also accepted | Canonical |
|---|---|
| `function add(a, b) { ... }` | `def add(a, b) { ... }` |
| `def add(a, b) -> number` | `def add(a, b): number` |
| `match (x) { 1 -> "one" }` | `match (x) { 1 => "one" }` |
| `map(xs, \n => n * 2)` | `map(xs, \n -> n * 2)` |

None of these is recorded in the AST, so `AgencyGenerator` needed no changes and normalization falls out for free. That property is the design rule, and every variation has a test asserting its tree is identical to the canonical spelling, plus one asserting the canonical spelling comes back out of `formatSource`.

This extends a pattern the parser already had: `await` and `sync` have been accepted and discarded since before this change.

## Ranges were dropped

An earlier revision added `3..6` as an infix range operator. It has been removed at the owner's direction — the design cost was not worth the payoff for one spelling. The spec and plan carry a note recording the decision and why.

## Two independent bug fixes

**A nested `function` declaration misparsed silently.** Because `function` is a legal identifier, `function inner() { }` inside a body parsed as a name plus a call with a trailing block, compiled to plausible TypeScript, and died at run time with a `ReferenceError` — while the same code with `def` produced a targeted error. This was reachable before any of the feature work.

**Number literals swallowed runs of dots.** `numberParser` accepted any run of dashes, dots, underscores and digits, so `1.2.3` became a single malformed number node with no error anywhere. It is now a real grammar: an optional sign, digits with `_` only between digits, and at most one fraction whose dot must be followed by a digit.

Worth being precise about the scope of that second fix, since an earlier changelog draft overclaimed it: `const x = 1.2.3` still does **not** error. The number parser now stops after `1.2`, and the body parser drops the leftover `.3`. The test pins that actual behaviour rather than the behaviour I assumed. Tightening the body parser's tolerance for trailing garbage is a separate change.

## Not breaking

`function` was added to `RESERVED_WORDS`, but that list governs identifier-hole filling in Template Agency only, so `const function = 5` still parses. There is a test pinning it.

`agency fmt` is now lossy for these spellings — a hand-written `function` becomes `def`. That is intended.

## Testing

- Round-trip pairs for every variation: same AST as canonical, canonical text back out of `formatSource`.
- Formatter idempotence for all four spellings.
- The nested-`function` regression, the match-guard `>-` adjacency case, doubled arrows (`def f() -> (string) -> string` — the separator normalizes while the arrow inside the type does not), `RESERVED_WORDS` coverage via `fill`, and the number-grammar cases including `1_` and `.5`.

One existing test changed meaning: `x -> y` was pinned as an invalid match arm and is now valid. That is the intended consequence of this feature, flagged because someone deliberately pinned it.

Full unit suite: **10048 passed, 0 failed**. `pnpm run lint:structure` clean. Formatter fixtures needed no regeneration.

## Follow-up

`docs/superpowers/specs/2026-07-31-trailing-comments-design.md` is included but not implemented. Trailing comments parse today but `agency fmt` relocates them, which can change what a comment appears to describe. It needs a new AST field rather than a spelling the AST can forget, so it did not belong here.


egonSchiele (commented):
Read the whole diff and checked the claims against the branch. This is a well-executed change — the design rule (a variation belongs here only if the AST can forget which spelling was used) holds throughout, and the test suite is genuinely strong: AST equality against the canonical spelling for every variation, plus a formatter assertion, plus an idempotence pass.

Three things I want to call out as good, because they are the parts a reviewer would otherwise skim past:

**The `armFollowsPattern` change is the subtle one.** Teaching the arm *lookahead* about `->` at the same time as the arm itself is not obvious, and missing it would not have failed loudly — `_ -> x` would have parsed as a variable pattern named `_` instead of the default case. Silent wrong answer, in a construct where that is expensive. Good catch, and the comment explains why.

**The unit-literal fix.** Finding that `a + 1..b - 1` parsed `1..b` as a byte literal, from a precedence test, and then fixing it by sharing one grammar between the two parsers rather than patching the second one, is the right shape of fix.

**The marker beats the alternative.** Deciding `[3..6]` structurally rather than by matching `..` against source text is what keeps `["a..b"]` legal, and there is a test for exactly that. I checked tarsec's `memo` (`combinators.js:1175`) — it returns cached results by reference with no cloning, so the non-enumerable marker survives memoized `exprParser` hits. That was the one way I thought this could quietly break, and it does not.

**One substantive finding**, on the `not(char("("))` opt-out: it cannot distinguish parenthesizing the whole range from parenthesizing the left operand, so `[(a + 1)..b]` silently becomes a one-element array. Details inline, along with a suggestion that removes the hole and simplifies the check.

The rest are small: an unnecessary `export`, a stale variable name, one set of loose test assertions, and a changelog claim (`1.2.3` is an error) that no test covers at program level.

Note that CI has not reported on this branch yet, so I have not seen the suites run.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4548,8 +4619,58 @@ export const modifiedAssignmentParser: Parser<Assignment> = withLoc((input: stri
...
+ * alternative in the enclosing `or(...)`.
+ */
+const bracketedRangeParser: Parser<never> = (input: string) => {
+  const probe = seqC(
+    char("["),
+    optionalSpacesOrNewline,
+    // Parentheses leave no trace in the AST, so `[(3..6)]` and `[3..6]` yield
+    // the identical node. Declining on a leading `(` is what makes the
+    // escape hatch the message recommends actually work.
+    not(char("(")),
```
egonSchiele: This opt-out has a hole, and it fails in the direction the whole parser exists to prevent.

`not(char("("))` cannot tell "the author parenthesized the whole range to opt out" from "the author parenthesized just the left operand". Both start with `(` right after the `[`:

```ts
const r = [(a + 1)..b]
```

The probe declines on the leading `(`, falls through to `agencyArrayParser`, and this becomes an array holding one range — so `for (i in [(a + 1)..b])` loops exactly once and binds the whole array. That is the silently-wrong outcome `BRACKETED_RANGE_MESSAGE` was written to stop, and parenthesizing an arithmetic endpoint is a pretty natural thing to write.

I would drop the paren opt-out entirely and point the message at `[range(3, 6)]` instead:

- The check gets simpler — any lone marked range in brackets errors, no lookahead needed.
- The hole closes, including `[(3..6)]` and `[(a + 1)..b]`.
- `[range(3, 6)]` is already legal (it has no marker, so it declines), and it is what `agency fmt` prints anyway, so the escape hatch is the canonical form rather than a second thing to learn.

The tests would follow: `accepts a parenthesized range as an array element` becomes a rejection case, and `accepts a hand-written range() call as a lone element` already covers the replacement.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3360,6 +3387,35 @@ function wsOp(opStr: string): Parser<string> {
 const wsKeyword = (kw: string): Parser<string> =>
   map(seqR(optionalSpacesOrNewline, str(kw), not(varNameChar), optionalSpaces), () => kw);
 
+/** Marks a `range()` call as having come from `..` rather than being written by
+ *  hand. Non-enumerable so it stays out of JSON, `Object.entries`, and
+ *  structural equality — the AST a test sees is identical to a hand-written
+ *  `range(3, 6)`, which is the design rule. `bracketedRangeParser` reads it to
+ *  tell `[3..6]` (an error) from `[range(3, 6)]` (legal). */
+const RANGE_OP_MARKER = "__fromRangeOp";
```
egonSchiele: The marker is a good call — it is structural rather than textual, and the `["a..b"]` test proves it beats the source-matching alternative.

One property worth adding to the comment, because it is invisible and someone will eventually rely on it: **the marker does not survive a spread.** `{ ...node }` copies only enumerable own properties, so any code that rebuilds nodes drops it. That is fine today because `bracketedRangeParser` reads it during the parse, before `desugarComprehensionsInBody` and `lowerPatterns` run. It would silently stop working if anything later tried to ask "was this a range?"

I did check the one place I thought might already break it: tarsec's `memo` (`combinators.js:1175`) caches and returns the result object by reference with no cloning, so memoized `exprParser` hits keep the marker. No issue there.

Suggest one line: "Read during the parse only — object spread drops non-enumerable properties, so this is gone by the time the lowering passes run."

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3360,6 +3387,35 @@ function wsOp(opStr: string): Parser<string> {
...
+
+export function isRangeOperatorCall(node: unknown): boolean {
```
egonSchiele: `isRangeOperatorCall` is only referenced once, by `bracketedRangeParser` about 250 lines below, and nothing outside this file imports it — I grepped the branch. The marker is an implementation detail of the bracket check, so exporting it puts a private mechanism in the module's public surface where a future caller might read it after lowering (when, per the note above, it is always `false`).

Drop the `export`.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -621,7 +642,10 @@ const TIME_MULTIPLIERS: Record<TimeUnitLiteral["unit"], number> = {
   w: 604800000,
 };
 
-const unsignedNumberChars = many1WithJoin(or(char("."), digit));
+// Shares `numberMagnitude` with `numberParser` rather than accepting any run of
+// dots and digits. A greedy version reads `a + 1..b - 1` as the byte literal
+// `1..b`, swallowing the range operator — the same failure `numberParser` had.
+const unsignedNumberChars = numberMagnitude;
```
egonSchiele: Nice catch on the underlying bug — `a + 1..b - 1` parsing `1..b` as a byte literal is exactly the kind of thing that only surfaces from a precedence test, and sharing one grammar between the two parsers is the right fix.

The name is now stale, though: `unsignedNumberChars` describes what it used to be (a run of characters), not what it is (a magnitude parser). Either rename it to `unsignedNumber` or use `numberMagnitude` directly at the call sites and delete the alias.

packages/agency-lang/lib/parsers/range.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    const call = (parsed.result.nodes[0] as any).body[0].value;
+    expect(call.functionName).toBe("range");
+    expect(call.loc.start).toBe(call.arguments[0].loc.start);
+    expect(call.loc.end).toBe(call.arguments[1].loc.end);
+  });
+});
+
+describe("dot runs stay distinct", () => {
+  it("spread still parses as a spread", () => {
+    expect(JSON.stringify(program(`node main() { const b = [...a, 3] }`))).toContain("splat");
```
egonSchiele: These three assertions are looser than the rest of the file. `JSON.stringify(...).toContain("splat")` proves the word appears *somewhere* in the serialized tree — it does not prove that `[...a, 3]` produced a splat, or that it landed in the right position. A regression that turned the spread into something else while leaving a `splat` node elsewhere in the program would still pass.

Every other test in this file uses `toEqualWithoutLoc` against a known-good program, which is both stricter and more readable. Same treatment here, e.g. compare `[...a, 3]` against a hand-written equivalent, or assert on the specific argument node rather than the whole stringified tree.

packages/agency-lang/lib/parsers/literals.test.ts
```diff
@@ -1797,3 +1797,49 @@ describe("literals parsers", () => {
...
+
+  it("strips underscores and leaves the rest intact", () => {
+    const r = numberParser("1_000 + 2");
+    expect(r.success).toBe(true);
+    if (!r.success) return;
+    expect(r.result.value).toBe("1000");
+    expect(r.rest).toBe(" + 2");
+  });
+
+  it("rejects a run with no digits", () => {
```
egonSchiele, line 1834: The changelog entry says "`1.2.3` is an error instead of a malformed number node," but nothing tests that at program level. What is covered here is `numberParser("1.2.3")` returning `1.2` with `.3` left over — which is the parser doing the right thing locally, not the whole expression failing.

Whether `const x = 1.2.3` actually errors depends on what the surrounding expression parser does with a leftover `.3`, and that is worth pinning rather than assuming, since it is the behaviour the changelog promises:

```ts
it("does not accept 1.2.3 as a whole expression", () => {
  expect(parseAgency(`node main() { const x = 1.2.3 }`, {}, false).success).toBe(false);
});
```

(The plan had this test; it looks like it got dropped when the assertion style was reworked.)

packages/agency-lang/docs/site/guide/basic-syntax.md
```diff
@@ -208,12 +208,21 @@ You can also write a module-level doc comment using the `@module` tag. This docu
 */
 ```
 
-> Note: comments must be on their own line, they cannot be at the end of a line containing code.
+> Note: prefer putting comments on their own line.
```
egonSchiele: undo all changes to this file.

packages/agency-lang/docs/site/guide/blocks.md
```diff
@@ -88,6 +88,8 @@ const results = twice(\ -> "hello")
 
 Inline blocks are single-line only, and the expression is implicitly returned.
 
+`=>` works in place of `->` here too, and `agency fmt` normalizes to `->`.
```
egonSchiele: undo all changes to this file.

packages/agency-lang/docs/site/guide/functions.md
```diff
@@ -14,6 +14,10 @@ def add(a: number, b: number): number {
 print(add(4, 5))
 ```
 
+`function` also works and is normalized to `def` by `agency fmt`, as is `->` in
```
egonSchiele: undo all changes to this file

packages/agency-lang/docs/site/guide/match-expressions.md
```diff
@@ -16,6 +16,8 @@ match (status) {
 }
 ```
 
+Arms may use `->` instead of `=>`; `agency fmt` normalizes to `=>`.
```
egonSchiele: undo all changes to this file

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3360,6 +3387,35 @@ function wsOp(opStr: string): Parser<string> {
 const wsKeyword = (kw: string): Parser<string> =>
   map(seqR(optionalSpacesOrNewline, str(kw), not(varNameChar), optionalSpaces), () => kw);
 
+/** Marks a `range()` call as having come from `..` rather than being written by
```
egonSchiele: why not just parse and lower to a range call right away? why is the marker needed?

packages/agency-lang/CHANGELOG.md
```diff
@@ -1,29 +1,11 @@
...
-- `fill` checks a plain value against the hole's real type, including record types and aliases, instead of only literal `string`, `number` and `boolean`. This means more issues are caught at compile time.
-
-### Runtime
-- `serve` exports `withRuntimeConfigOverrides`, so a host serving many compiled modules can route each one's logs to its own project.
-- `runCode` takes a `maxCost` cap.
-
-### CLI
-- `agency run file.agency [args...]` passes trailing arguments to the entry node's parameters. Extra arguments are a loud error instead of a silent drop.
-- `fmt` only collapses a one-statement arm block when the arm grammar can re-parse it, so formatting no longer breaks a file with a raise statement in an arm.
+## Aug 1 2026
```
egonSchiele: undo all changes to this file

PR #756, opened 2026-08-01
Accept common keyword and literal spellings

egonSchiele:
Six more spellings models reach for, following the same rule as #750: a variation belongs here only if the AST can forget which spelling was used. None is recorded, so `AgencyGenerator` needed no changes and `agency fmt` normalizes for free.

| Also accepted | Canonical |
|---|---|
| `interface Foo { a: string }` | `type Foo = { a: string }` |
| `elif (b) { }` | `else if (b) { }` |
| `greet(name = "Adit")` | `greet(name: "Adit")` |
| `const x = undefined` | `const x = null` |
| `const s: String` | `const s: string` |
| `'''hi'''` | `"""hi"""` |

## Two of these were silent misparses, not missing features

Worth calling out because they were worse than "doesn't work":

**`'''hi'''` lost its contents.** It parsed as an empty single-quoted string followed by two stray statements — no error, and the text silently gone.

**`const s: String = "x"` accepted anything.** `String` parsed as an unresolved type alias, which types as `any` through the known checker hole. So the annotation was decorative.

## A latent hazard the capitalized primitives exposed

Primitive type names were matched with no word boundary, so adding `String`/`Number`/`Boolean` made `str("Number")` match the front of `NumberInRange` and strand the rest. The lowercase names had the same hazard (`stringify`) — the capitalized ones just make it far likelier, since PascalCase type names routinely start with these words. The alternation now requires a whole word, which fixes both.

## `undefined`

`docs/dev/null-and-undefined.md` records a deliberate decision that Agency has exactly one nothing-value. This does not change that: `undefined` becomes a second *spelling* of `null`, not a second value. The *type* `undefined` has normalized to `null` since that work landed; this makes the value side agree. The doc's rejected alternative was "keep both, distinct", which is not what this does.

Two consequences worth knowing:

- In expression position both `null` and `undefined` arrive at `variableNameParser`, not `nullParser` — so that is where they converge.
- `undefined` now narrows like `null`. A narrowing test that asserted `r.kind == undefined` produced no candidates ("undefined is a variableName, not a literal") is updated to assert it matches `== null`. That premise is obsolete, and the new behaviour is strictly better.

The one dev-doc sentence stating users cannot write `undefined` as a value is updated, since it is now false. No guide pages touched.

## One thing I got wrong and caught

The first draft made the `=` in a type alias `optional(...)` so one parser could serve both keywords. That also accepts `type A number`, which is a targeted error relied on well beyond this parser — 123 tests failed. The separator is now a parameter: `type` requires `=`, `interface` has none. There is a test pinning it.

## Testing

New `lib/parsers/keywordAliases.test.ts`: an AST-equality pair per variation, a formatter assertion per variation, idempotence, and a "does not swallow neighbouring syntax" block covering `NumberInRange`, `undefinedThing`, `a == b` vs the named-argument `=`, single-quoted strings, and `type A number`.

Full unit suite: **10137 passed, 0 failed**. `pnpm run lint:structure` clean. `make` builds the stdlib and examples through the changed parser.

Note for reviewers: most of the suite needs a built `dist/`. Running it in a fresh worktree without `make` first shows ~115 unrelated failures ("Failed to resolve entry for package agency-lang").

### DIFF
--- a/packages/agency-lang/lib/parsers/keywordAliases.test.ts
+++ b/packages/agency-lang/lib/parsers/keywordAliases.test.ts
@@ -1,0 +1,129 @@
+import { describe, it, expect } from "vitest";
+import { parseAgency } from "@/parser.js";
+import { formatSource } from "@/formatter.js";
+
+function program(src: string) {
+  const parsed = parseAgency(src, {}, false);
+  if (!parsed.success) throw new Error(`expected a successful parse, got: ${parsed.message}`);
+  return parsed.result;
+}
+
+/** Each variation and the canonical spelling it must parse identically to. */
+const pairs: [string, string, string][] = [
+  [
+    "interface for type",
+    `interface Foo { a: string, b: number }\nnode main() { print(1) }`,
+    `type Foo = { a: string, b: number }\nnode main() { print(1) }`,
+  ],
+  [
+    "exported interface",
+    `export interface Foo { a: string }\nnode main() { print(1) }`,
+    `export type Foo = { a: string }\nnode main() { print(1) }`,
+  ],
+  [
+    "elif for else if",
+    `node main() { if (a) { print(1) } elif (b) { print(2) } else { print(3) } }`,
+    `node main() { if (a) { print(1) } else if (b) { print(2) } else { print(3) } }`,
+  ],
+  [
+    "named argument with =",
+    `node main() { greet(name = "Adit", greeting = "Hi") }`,
+    `node main() { greet(name: "Adit", greeting: "Hi") }`,
+  ],
+  [
+    "undefined as a value",
+    `node main() { const x = undefined\nprint(x) }`,
+    `node main() { const x = null\nprint(x) }`,
+  ],
+  [
+    "undefined in a type",
+    `node main() { const x: string | undefined = "a" }`,
+    `node main() { const x: string | null = "a" }`,
+  ],
+  [
+    "capitalized primitives",
+    `node main() { const s: String = "x"\nconst n: Number = 1\nconst b: Boolean = true }`,
+    `node main() { const s: string = "x"\nconst n: number = 1\nconst b: boolean = true }`,
+  ],
+  [
+    "triple single-quoted string",
+    `node main() { const s = '''hi there''' }`,
+    `node main() { const s = """hi there""" }`,
+  ],
+  [
+    "triple single-quoted with interpolation",
+    `node main() { const n = 1\nconst s = '''a \${n} b''' }`,
+    `node main() { const n = 1\nconst s = """a \${n} b""" }`,
+  ],
+];
+
+describe("keyword aliases parse to the canonical AST", () => {
+  for (const [name, variation, canonical] of pairs) {
+    it(name, () => {
+      expect(program(variation)).toEqualWithoutLoc(program(canonical));
+    });
+  }
+});
+
+describe("agency fmt normalizes each alias", () => {
+  const expectations: [string, string, string][] = [
+    ["interface", `interface Foo { a: string }`, "type Foo = {"],
+    ["elif", `node main() { if (a) { print(1) } elif (b) { print(2) } }`, "else if"],
+    ["named arg =", `node main() { greet(name = "Adit") }`, `greet(name: "Adit")`],
+    ["undefined", `node main() { const x = undefined }`, "null"],
+    ["String", `node main() { const s: String = "x" }`, ": string"],
+    ["triple single quotes", `node main() { const s = '''hi''' }`, `"""`],
+  ];
+
+  for (const [name, src, expected] of expectations) {
+    it(`${name} formats to the canonical spelling`, () => {
+      expect(formatSource(src)).toContain(expected);
+    });
+  }
+
+  it("does not leave the alias spelling behind", () => {
+    expect(formatSource(`interface Foo { a: String }`)).not.toContain("interface");
+    expect(formatSource(`node main() { const x = undefined }`)).not.toContain("undefined");
+  });
+
+  it("formatting twice is a fixed point", () => {
+    for (const [, src] of expectations) {
+      const once = formatSource(src);
+      expect(once).not.toBeNull();
+      expect(formatSource(once as string)).toBe(once);
+    }
+  });
+});
+
+describe("aliases do not swallow neighbouring syntax", () => {
+  it("a type name starting with a primitive word still parses", () => {
+    // `str("Number")` would otherwise match the front of `NumberInRange`.
+    expect(parseAgency(`type NumberInRange = number\nnode main() { print(1) }`, {}, false).success)
+      .toBe(true);
+    expect(parseAgency(`type Stringy = string\nnode main() { print(1) }`, {}, false).success)
+      .toBe(true);
+  });
+
+  it("an identifier starting with `undefined` is untouched", () => {
+    const parsed = parseAgency(`node main() { const undefinedThing = 1\nprint(undefinedThing) }`, {}, false);
+    expect(parsed.success).toBe(true);
+    expect(JSON.stringify(parsed)).toContain("undefinedThing");
+  });
+
+  it("`type A number` is still a parse error", () => {
+    // The `=` is required for `type`; only `interface` omits it. Making the
+    // separator optional for both would silently accept this — it did in a
+    // first draft, and broke 123 tests across the compiler.
+    expect(parseAgency(`type A number`, {}, false).success).toBe(false);
+  });
+
+  it("== is not read as a named-argument separator", () => {
+    expect(program(`node main() { if (a == b) { print(1) } }`))
+      .toEqualWithoutLoc(program(`node main() { if (a == b) { print(1) } }`));
+    expect(parseAgency(`node main() { f(a == b) }`, {}, false).success).toBe(true);
+  });
+
+  it("single-quoted strings still parse", () => {
+    expect(parseAgency(`node main() { const s = 'hi' }`, {}, false).success).toBe(true);
+  });
+});
--- a/packages/agency-lang/lib/parsers/keywordAliases.test.ts
+++ b/packages/agency-lang/lib/parsers/keywordAliases.test.ts
@@ -166,6 +166,15 @@ describe("interface inheritance gets a targeted error", () => {
     expect(parsed.message).toContain("&");
   });
 
+  it("covers the generic form too", () => {
+    // `extends` and type parameters travel together in TypeScript, so this
+    // shape is at least as likely as the bare one.
+    const parsed = parseAgency(`interface Foo<T> extends Bar { a: T }`, {}, false);
+    expect(parsed.success).toBe(false);
+    if (parsed.success) return;
+    expect(parsed.message).toMatch(/no interface inheritance/);
+  });
+
   it("leaves a type named `extendsSomething` alone", () => {
     expect(parseAgency(`interface Foo { extendsThing: string }\nnode main() { print(1) }`, {}, false).success)
       .toBe(true);
--- a/packages/agency-lang/lib/parsers/keywordAliases.test.ts
+++ b/packages/agency-lang/lib/parsers/keywordAliases.test.ts
@@ -118,12 +118,56 @@ describe("aliases do not swallow neighbouring syntax", () => {
   });
 
   it("== is not read as a named-argument separator", () => {
-    expect(program(`node main() { if (a == b) { print(1) } }`))
-      .toEqualWithoutLoc(program(`node main() { if (a == b) { print(1) } }`));
-    expect(parseAgency(`node main() { f(a == b) }`, {}, false).success).toBe(true);
+    // Assert an observable outcome: if `==` were split into a named argument
+    // `a` with value `= b`, the formatter could not print it back.
+    expect(formatSource(`node main() { f(a == b) }`)).toContain("a == b");
+    expect(formatSource(`node main() { if (a == b) { print(1) } }`)).toContain("a == b");
   });
 
   it("single-quoted strings still parse", () => {
     expect(parseAgency(`node main() { const s = 'hi' }`, {}, false).success).toBe(true);
+  });
+});
+
+describe("the undefined alias is confined to expression position", () => {
+  it("does not rename a property called undefined", () => {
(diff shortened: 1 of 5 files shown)


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -919,9 +942,16 @@ export const variableNameParser: Parser<VariableNameLiteral> = label("an identif
...
       return {
         type: "variableName" as const,
-        value: `${captures.init}${captures.value}`,
+        // `undefined` is a second spelling of `null`, not a second value —
+        // Agency has exactly one nothing-value (docs/dev/null-and-undefined.md),
+        // and the *type* `undefined` has always normalized to `null`. In
+        // expression position both spellings arrive here rather than at
+        // `nullParser`, so converging them here is what makes the value side
+        // agree with the type side. `agency fmt` prints `null`.
+        value: name === "undefined" ? "null" : name,
```
egonSchiele: This converts `undefined` in **every** position an identifier can appear, not just expression position, and one of those positions is wrong.

`variableNameParser` supplies the property name for member access at line 2690:

```ts
// Second try: just a property name
const nameResult = variableNameParser(afterDot);
```

So `obj.undefined` now reads the property `null`. The same mechanism reaches template hole names (`holeParser`, line 987) and match-pattern bindings (lines 6260, 6346, 6419), where a binder spelled `undefined` becomes one spelled `null`.

There is also a second mechanism doing the same job. `nullParser` (line 1146) was changed in this PR to accept `undefined` with a proper word boundary, and `literalParser` tries it **first**:

```ts
export const literalParser = memo("literalParser", or(
  nullParser,          // <- wins
  ...
  variableNameParser,
));
```

So along that path the rewrite here never fires. The PR description says "in expression position both `null` and `undefined` arrive at `variableNameParser`, not `nullParser`" — that does not hold for `literalParser`, where the order is the other way around.

The deeper smell is that the two produce **different nodes for the same source**: `nullParser` gives `{ type: "null" }`, this gives `{ type: "variableName", value: "null" }`. One spelling should not land on two different node types depending on which parser reached it first.

I would delete this rewrite and keep only `nullParser`, adding it ahead of `variableNameParser` in whichever expression paths do not already go through `literalParser`. That leaves one correctly-bounded mechanism, fixes `obj.undefined`, and makes the node type deterministic.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -5184,6 +5255,17 @@ export const withModifierParser: Parser<WithModifier> = withLoc((input: string)
 });
 
 const elseClauseParser: Parser<AgencyNode[]> = (input: string) => {
+  // `elif` is the Python spelling of `else if`, and is rewritten to one here:
+  // the remaining input keeps the `if`, so the "else if" branch below handles
+  // it unchanged. `agency fmt` prints `else if`.
+  const elif = seqC(optionalSpaces, str("elif"), peek(not(varNameChar)))(input);
+  if (elif.success) {
+    const asElseIf = ifParser("if" + elif.rest);
```
egonSchiele: Re-parsing a synthetic string costs the `loc` for everything inside an `elif` branch.

`"if" + elif.rest` is not the source text. tarsec computes spans against the input it is handed, so every position produced inside `asElseIf` is measured from offset 0 of a string that never existed — the real `elif` is four characters at some offset, the synthetic `if` is two at zero. Error messages and LSP squiggles inside an `elif` branch will point somewhere else. Nested `elif` compounds it: each level builds another synthetic string on top of the previous one's remainder.

It is also the one place in this PR that reaches for string surgery. Both other multi-spelling cases here are handled by parameterizing the parser, and they read well — `baseTypeAliasParserFor("type", str("="))` alongside `baseTypeAliasParserFor("interface", succeed(undefined))`, and the two `multiLineStringParserFor` calls for the string delimiters.

Same treatment would work here: parameterize the if-parser's keyword so `elif` is parsed in place, with real offsets, rather than rewritten and re-fed.

Minor: `peek(not(varNameChar))` is redundant. Per the comment at line 511 of this same file, `not(...)` is already a zero-width lookahead.

egonSchiele (commented):
Read the diff and checked the claims against the branch. The same design rule as #750 holds throughout, and two of these six are genuine bug fixes rather than conveniences — the `'''` case was silently losing string contents, and the missing word boundary on primitive type names was a latent hazard that `stringify` could already have hit.

Three things done well:

**The word-boundary fix is the valuable part.** Adding `not(varNameChar)` after the primitive alternation fixes a bug that predates the capitalized names; those just raised the odds, since PascalCase type names routinely start with `Number`/`String`/`Boolean`. Good that it was fixed at the alternation rather than special-cased for the new entries.

**Parameterizing rather than duplicating.** `baseTypeAliasParserFor(keyword, separator)` and `multiLineStringParserFor(delim)` are the right shape, and making the separator a parameter instead of `optional(str("="))` is the correct call for exactly the reason given.

**Recording the `optional(str("="))` mistake in the description.** "One thing I got wrong and caught," with the 123 failures and why, is more useful to a reviewer than a clean history would have been.

**One substantive finding**, on `undefined`: the conversion is applied in `variableNameParser`, which is the identifier parser for *every* position — including member-access property names, so `obj.undefined` now reads `obj.null`. There is also a second, correctly-bounded conversion in `nullParser`, and `literalParser` tries that one first, so the two overlap and produce different node types for the same source. Details inline.

The rest are smaller: `interface` and `elif` are missing from `RESERVED_WORDS` (the precedent #750 set), the `elif` implementation re-parses a synthetic string and loses source positions, one comment argues from a premise that is not quite true, and there is a missing test for the case the main finding describes.

Also worth deciding before merge: `interface Foo extends Bar` currently fails with a generic type error, and `extends` is probably more common from models than the bare form this accepts.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2056,6 +2120,11 @@ const baseTypeAliasParser: Parser<TypeAlias> = withLoc(memo(
   ),
 ));
 
+const baseTypeAliasParser: Parser<TypeAlias> = or(
+  baseTypeAliasParserFor("type", str("=")),
+  baseTypeAliasParserFor("interface", succeed(undefined)),
```
egonSchiele, line 2169: `interface` and `elif` are both keywords the grammar now consumes, but neither was added to `RESERVED_WORDS`. Its doc comment asks for exactly that ("extend it when a new keyword lands"), and #750 set the precedent by adding `function` in the same situation, with a test in `fill.test.ts` pinning it.

Without it, `fill(t, { name: "interface" })` produces source that explodes at re-parse far from its cause, which is the failure that list exists to prevent.

Separately, worth thinking about before this ships: `interface Foo extends Bar { ... }`. There is no `extends` handling, so the separator matches empty and `variableTypeParser` is handed `extends Bar { ... }`, which fails with a generic type error. Models write `extends` constantly — probably more often than the bare form this accepts. Not necessarily in scope here, but a targeted message ("Agency has no interface inheritance; write `type Foo = Bar & { ... }`") would turn a confusing failure into a useful one, and the alternative is an agent burning turns on it.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2494,7 +2563,9 @@ const namedArgumentParser: Parser<NamedArgument> = memo(
     set("type", "namedArgument"),
     capture(many1WithJoin(varNameChar), "name"),
     optionalSpaces,
-    char(":"),
+    // `=` is the Python spelling. Agency has no assignment expression, so
```
egonSchiele: The reasoning in this comment is looser than the code. Agency does have assignment-shaped *expressions* — `+=`, `-=`, `*=` and `/=` are entries in the operator table at precedence 6, built with `makeBinOp`. So "nothing else can appear here" is not quite the argument.

What actually makes this safe is the guard you wrote: `not(char("="))` rejects `==`, and the `char("=")` itself fails on `+=`, `>=` and `!=` because the character before the `=` is consumed by neither `varNameChar` nor `char("=")`. That is a good argument, it is just not the one the comment makes.

Suggest pointing the comment at the guard rather than at the absence of assignment expressions.

packages/agency-lang/lib/parsers/keywordAliases.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    expect(parseAgency(`type NumberInRange = number\nnode main() { print(1) }`, {}, false).success)
+      .toBe(true);
+    expect(parseAgency(`type Stringy = string\nnode main() { print(1) }`, {}, false).success)
+      .toBe(true);
+  });
+
+  it("an identifier starting with `undefined` is untouched", () => {
+    const parsed = parseAgency(`node main() { const undefinedThing = 1\nprint(undefinedThing) }`, {}, false);
+    expect(parsed.success).toBe(true);
+    expect(JSON.stringify(parsed)).toContain("undefinedThing");
```
egonSchiele, line 110: This covers the prefix hazard (`undefinedThing`) but not the position hazard, which is where the `variableNameParser` rewrite actually bites:

```ts
it("does not rewrite a property named undefined", () => {
  const parsed = parseAgency(`node main() { print(obj.undefined) }`, {}, false);
  expect(JSON.stringify(parsed)).toContain("undefined");
});
```

That one fails on the current implementation — see the comment on line 954. Worth adding a match-pattern binding case too, since `variableNameParser` feeds those as well.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -1107,9 +1138,21 @@ export const booleanParser: Parser<BooleanLiteral> = label("a boolean", (input:
   return parser(input);
 });
 
+// `undefined` is a second spelling of `null`, not a second value. Agency has
+// exactly one nothing-value (docs/dev/null-and-undefined.md); the *type*
+// `undefined` has always normalized to `null` here, and accepting the literal
+// makes the value side agree. `agency fmt` prints `null`.
+// `undefined` is a second spelling of `null`, not a second value. Agency has
```
egonSchiele, line 1141: The old comment block survived alongside the new one — lines 1141-1144 and 1145-1148 say the same thing twice, differing only in "normalized to `null` here" versus "normalized to `null`". Looks like a merge slip in the fixup. Delete the first paragraph.

The third paragraph (the word-boundary note) is the one worth keeping, and the change it describes is more than housekeeping: `not(varNameChar)` now applies to `null` as well as `undefined`, so `nullThing` stops being consumed as `null` with `Thing` stranded. That is a pre-existing bug this PR fixes in passing, and there is a test for it. Worth a changelog line, since it is a real fix rather than part of the alias work.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -5184,6 +5310,16 @@ export const withModifierParser: Parser<WithModifier> = withLoc((input: string)
 });
 
 const elseClauseParser: Parser<AgencyNode[]> = (input: string) => {
+  // `elif` stands in for `else if`, so it is an else clause whose sole
+  // statement is an if. See `elifParser`.
+  const elifPrefix = seqC(optionalSpaces)(input);
```
egonSchiele: This guard never fails. `optionalSpaces` succeeds on any input by definition, and `seqC` of a single always-succeeding parser inherits that, so `elifPrefix.success` is always `true` and the `if` is dead weight around the real call.

```ts
const afterSpaces = optionalSpaces(input);
const asElseIf = elifParser(afterSpaces.rest);
if (asElseIf.success) {
  return success([asElseIf.result], asElseIf.rest);
}
```

Same behaviour, one less branch to wonder about. (The `seqC` wrapper is not doing anything either — `optionalSpaces` can be called directly.)

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -2056,6 +2131,42 @@ const baseTypeAliasParser: Parser<TypeAlias> = withLoc(memo(
...
+ * Throws rather than returning `failure(...)` for the same reason
+ * `bodyDeclarationParser` does — a plain failure would be shadowed by a sibling
+ * alternative in the enclosing `or(...)`.
+ */
+const interfaceExtendsParser: Parser<never> = (input: string) => {
+  const probe = seqC(
+    optional(seqC(str("export"), spaces)),
+    str("interface"),
+    spaces,
+    many1WithJoin(varNameChar),
```
egonSchiele, line 2148: The probe requires `spaces` immediately after the interface name, so it declines on the generic form:

```ts
interface Foo<T> extends Bar { a: T }
```

`many1WithJoin(varNameChar)` stops at the `<`, then `spaces` sees `<` and fails. The probe returns a plain failure, parsing falls through, and the author gets back the generic type error this parser was added to replace — for a shape that is, if anything, more likely to be written than the non-generic one, since `extends` and type parameters travel together in TypeScript.

Making the type-parameter list optional in the probe would cover it. The test at `keywordAliases.test.ts` covers only `interface Foo extends Bar`, so this gap is currently invisible.

egonSchiele (commented):
Re-reviewed after `6f3d399b`. Every finding from the last round is addressed, and addressed at the right altitude rather than patched around:

- **The `undefined` alias moved out of `variableNameParser`** into a dedicated `undefinedAliasParser` registered in `baseAtom` ahead of `valueAccessParser`. That confines it to expression position, and the comment explains why it cannot live in the identifier parser. Tests now cover property names, binders, and match-pattern binders.
- **`elif` is parsed in place** via `_ifParserInnerFor(keyword)` instead of re-parsing `"if" + rest`, so the `loc` problem is gone and it now matches how `type`/`interface` and the string delimiters are handled in this same PR.
- **`interface` and `elif` are in `RESERVED_WORDS`**, with a parameterized `fill.test.ts` case.
- **`interface ... extends` gets a targeted message** naming `Bar & { ... }`, following the `bodyDeclarationParser` pattern.
- **The named-argument comment** now argues from the guard rather than from a premise that was not true.

Two things I want to note that you did without being asked:

**You caught a tautological test.** The old `==` case was `expect(program(X)).toEqualWithoutLoc(program(X))` — the same source compared to itself, which passes regardless of what the parser does. Replacing it with a `formatSource` assertion that the text round-trips is a real check.

**The honest deferral on node shapes.** The comment in `baseAtom` noting that `null` already has two node shapes depending on whether `literalParser` or the atom path reached it, and that unifying them is a separate change, is the right call — I raised it as a smell and it is genuinely out of scope here.

What is left is three small things the fixup itself introduced or left behind. None blocks merge.

PR #757, opened 2026-08-01
Teach switch and C-style for what Agency uses instead

egonSchiele:
Two constructs Agency deliberately does not have now get a targeted error naming the Agency equivalent.

This is the opposite call from #750 and #756. Those accepted spellings of constructs Agency *has*. These two differ in semantics — `match` has no fall-through and does exhaustiveness checking; Agency has no C-style loop at all — so an author reaching for them is reasoning about the wrong construct, and accommodating that silently would be worse than refusing it.

## What you get now

```
node main() { switch (x) { case 1: print(1) } }
```

```
Agency has no `switch` statement. Use a `match` block instead:

  match (x) {
    "a" => doThing()
    "b" => doOtherThing()
    _   => fallback()
  }

Arms do not fall through, so no `break` is needed, and `match` checks that you
covered every case. `_` is the catch-all.
```

```
node main() { for (let i = 0; i < 10; i++) { print(i) } }
```

```
Agency has no C-style `for` loop. To count, iterate a range:

  for (i in range(0, 10)) { print(i) }

To walk a collection, iterate it directly:

  for (item in items) { print(item) }
  for (item, i in items) { print(i, item) }

For a condition-driven loop, use `while (cond) { ... }`.
```

## What they gave before

Both failed already, but with messages that pointed nowhere useful:

| Input | Old message |
|---|---|
| `switch (x) { case 1: ... }` | `expected node body` |
| `for (let i = 0; i < 10; i++)` | ``expected `(` to open for loop condition`` |

The second is worse than unhelpful — there *is* a `(`. What the parser could not find was the `in`.

## Two precision decisions

**The switch probe requires a `case` or `default` arm**, rather than committing as soon as it sees `switch (`. `switch` is a legal identifier, and the standard library already has a `std::git::switch` effect, so a bare `switch(...)` can be an ordinary call — possibly one taking a block, which is valid Agency. Requiring the arm keyword makes the shape unmistakable. The cost: `switch (x) { 1 => ... }`, mixing the two syntaxes, is not caught and still gets a generic error. There is a test that a function named `switch` still parses.

**The for probe keys on the initializer shape** — an optional `let`/`const`/`var`, a name, then `=` — rather than scanning for a `;` before the `)`. Scanning would misfire on a semicolon inside a string:

```ts
for (x in ["a;b"]) { print(x) }   // must still parse
```

That case is tested, along with the index, object-destructuring, array-destructuring, and `range()` forms.

## Testing

New `lib/parsers/foreignSyntax.test.ts`, 16 cases: both constructs caught, the message content asserted (not just that it failed), and a block of negative cases covering every legitimate loop form plus a call to something named `switch`.

Full unit suite: **10162 passed, 0 failed**. `pnpm run lint:structure` clean.

### DIFF
--- a/packages/agency-lang/lib/parsers/parsers.ts
+++ b/packages/agency-lang/lib/parsers/parsers.ts
@@ -4726,16 +4726,23 @@ const SWITCH_MESSAGE =
  * reaching for `switch` is reasoning about the wrong thing and should be told,
  * not silently accommodated.
  *
- * The probe requires a `case` or `default` inside the block rather than
+ * The probe requires a labelled `case ...:` or `default:` arm rather than
  * stopping at `switch (`. `switch` is a legal identifier, and the standard
- * library already has a `std::git::switch` effect, so a bare `switch(...)`
- * could be an ordinary call — possibly one taking a block. Requiring the arm
- * keyword makes the shape unmistakable.
+ * library already has a `std::git::switch` effect, so a bare `switch(...)` can
+ * be an ordinary call — possibly one taking a block. The label's `:` is what
+ * makes the shape unmistakable: a block that merely *starts* with something
+ * named `case` (say `case(1)`) is ordinary Agency and must still parse.
+ *
+ * The condition is scanned to the opening `{`, not to the first `)`. Stopping
+ * at `)` declines on `switch (f(x))`, where the first `)` closes the inner
+ * call — and switching on a computed value is at least as common as switching
+ * on a name. The trade is a `)` in the condition for a `{` in the condition
+ * (`switch ({ a: 1 }.a)`), which is the rarer shape.
  *
  * The cost of that precision: `switch (x) { 1 => ... }`, mixing the two
  * syntaxes, is not caught and still gets a generic error.
  *
- * Throws rather than returning `failure(...)` for the same reason
+ * Commits the failure rather than returning a plain one for the same reason
  * `bodyDeclarationParser` does — a plain failure would be shadowed by a
  * sibling alternative in the enclosing `or(...)`.
  */
@@ -4744,13 +4751,15 @@ const switchStatementParser: Parser<never> = (input: string) => {
     str("switch"),
     optionalSpaces,
     char("("),
-    many(noneOf(")")),
-    char(")"),
-    optionalSpacesOrNewline,
+    many(noneOf("{")),
     char("{"),
     optionalSpacesOrNewline,
-    or(str("case"), str("default")),
-    not(varNameChar),
+    or(
+      // `case <expr>:` — the space is required, which is what keeps `case(1)`
+      // (an ordinary call) from matching.
+      seqC(str("case"), spaces, many1(noneOf(":\n{}")), char(":")),
+      seqC(str("default"), optionalSpaces, char(":")),
+    ),
   );
   const probed = probe(input);
   if (!probed.success) return failure("", input);
@@ -4769,6 +4778,10 @@ const C_STYLE_FOR_MESSAGE =
   "  for (item in items) { print(item) }\n" +
   "  for (item, i in items) { print(i, item) }\n" +
   "\n" +
+  "To build a list from another list, use a comprehension:\n" +
+  "\n" +
+  "  const doubled = [x * 2 for x in items]\n" +
+  "\n" +
   "For a condition-driven loop, use `while (cond) { ... }`.";
 
 /**
@@ -4779,6 +4792,9 @@ const C_STYLE_FOR_MESSAGE =
  * The probe keys on the initializer shape (an optional declaration keyword, a
  * name, then `=`) rather than scanning for a `;`. Scanning would misfire on a
  * semicolon inside a string, as in `for (x in ["a;b"])`.
+ *
+ * Commits the failure rather than returning a plain one, so a sibling
+ * alternative in the enclosing `or(...)` cannot shadow the message.
  */
 const cStyleForParser: Parser<never> = (input: string) => {
   const probe = seqC(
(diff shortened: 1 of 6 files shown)


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4708,6 +4708,97 @@ export const modifiedAssignmentParser: Parser<Assignment> = withLoc((input: stri
...
+  "Agency has no C-style `for` loop. To count, iterate a range:\n" +
+  "\n" +
+  "  for (i in range(0, 10)) { print(i) }\n" +
+  "\n" +
+  "To walk a collection, iterate it directly:\n" +
+  "\n" +
+  "  for (item in items) { print(item) }\n" +
+  "  for (item, i in items) { print(i, item) }\n" +
+  "\n" +
+  "For a condition-driven loop, use `while (cond) { ... }`.";
```
egonSchiele, line 4785: You could also suggest list comprehensions as another option?

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4708,6 +4708,97 @@ export const modifiedAssignmentParser: Parser<Assignment> = withLoc((input: stri
...
+ * Throws rather than returning `failure(...)` for the same reason
+ * `bodyDeclarationParser` does — a plain failure would be shadowed by a
+ * sibling alternative in the enclosing `or(...)`.
+ */
+const switchStatementParser: Parser<never> = (input: string) => {
+  const probe = seqC(
+    str("switch"),
+    optionalSpaces,
+    char("("),
+    many(noneOf(")")),
```
egonSchiele: `many(noneOf(")"))` stops at the **first** `)`, not the one closing the switch condition, so the probe declines whenever the condition contains parentheses:

```ts
switch (f(x)) { case 1: print(1) }
```

The scan consumes `f(x`, `char(")")` takes the inner `)`, then the probe expects `{` and finds `)`. It returns a plain failure, parsing falls through, and the author gets back the generic "expected node body" this parser exists to replace.

Switching on a computed value is at least as common as switching on a bare name — arguably more so — and this gap is not in the list of documented limitations the way `switch (x) { 1 => ... }` is.

Scanning to the brace instead handles it, since the arm keyword is what actually disambiguates:

```ts
char("("),
many(noneOf("{")),
char("{"),
```

That trades a `)` in the condition for a `{` in the condition (`switch ({ a: 1 }.a)`), which seems like the better side of the trade. Either way it needs a test — `catches \`case\` arms` uses a bare identifier, so nothing currently covers the nested-paren shape.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4708,6 +4708,97 @@ export const modifiedAssignmentParser: Parser<Assignment> = withLoc((input: stri
   return success(out, result.rest);
 });
 
+const SWITCH_MESSAGE =
```
egonSchiele, line 4711: These messages are much longer than anything else in this file, and I want to flag it as a question rather than a defect, because the audience argument may well justify it.

The closest existing analogue is `RESERVED_CLASS_MESSAGE` (line 6244), which is the same situation — a construct Agency deliberately does not have, refused with a pointer to the replacement:

```ts
"`class` definitions are no longer supported in Agency. " +
"Use functions and plain objects instead, or instantiate an imported " +
"JS class with `new Foo(...)`."
```

Three lines of prose, no embedded newlines, no code block. `BODY_RESERVED_MODIFIER_MESSAGE` and the `STATIC_*` family are all the same shape. These two are roughly ten lines each with indented multi-line examples.

Two thoughts:

The codebase does have a place for long-form guidance — `agency explain <code>`, wired up at `scripts/agency.ts:1061`, which the CLI advertises with "Run `agency explain <code>` for details." But I checked, and the code registry lives in `lib/typeChecker/diagnostics.ts`; parser refusals like `RESERVED_CLASS_MESSAGE` carry no code at all. So routing these through `explain` is not a small change, and I am not proposing it here.

Given that, the real question is just whether the parse-error string is where a worked example belongs, or whether one prose sentence naming `match` and `for ... in range(...)` would do the job for both a human and a model. I lean toward shorter and consistent with the neighbours, but you have thought harder than I have about what a model needs to recover in one turn, and if the answer is "it needs to see the shape," that is a legitimate reason to break the pattern. Worth being deliberate about it either way, since these will be the template for the next refusal someone adds.

egonSchiele (commented):
Read the diff and the tests. This is the right call and the reasoning in the description is convincing — accepting a spelling of something Agency has is a different thing from accommodating a construct whose semantics differ, and `switch` genuinely is the second kind.

Two design decisions deserve specific credit:

**Requiring the `case`/`default` arm rather than committing at `switch (`.** `switch` is a legal identifier and `std::git::switch` already exists, so committing early would have broken a real standard-library call. The test that a function named `switch` still parses is the right guard, and the cost of the precision is honestly documented in the comment.

**Keying the `for` probe on the initializer shape rather than scanning for `;`.** This is the better design for exactly the reason given, and `for (x in ["a;b"])` is a well-chosen test — a semicolon-scanning version would pass every obvious test and fail on that one. Good instinct to write it.

The old message for the C-style loop really was worse than unhelpful, so replacing "expected `(` to open for loop condition" when there plainly is a `(` is a clear win.

One real gap: the switch probe's condition scan stops at the first `)`, so `switch (f(x)) { ... }` is not caught. Details inline, with a one-line fix.

The other comment is a question about message length rather than a defect — these are an order of magnitude longer than every other refusal message in the file, and it is worth being deliberate about that since they will be the template for the next one.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: I think the real answer to this question is that we should probably have *more* error messages that give code examples, as that would be more useful for agents when they stumble upon an error.

PR #758, opened 2026-08-01
Refuse ternaries; add worked examples to three parser messages

egonSchiele:
Follows #757. One new refusal, three messages that gain a worked example, and a test that keeps every example honest.

## Ternaries

```ts
const y = x ? 1 : 2
```

Before, that produced the body parser's catch-all — `expected node body`, with no line or column:

```
Agency has no ternary (`? :`). Use an `if ... then ... else` expression:

  const label = if isProd then "Production" else "Local"

The `else` is required. The expression is only allowed as a `const`/`let` value
or a `return` — for anything more involved, use `match`.
```

**The check wraps `exprParser`, not the statement list.** A ternary never appears in statement position — it is the value of an assignment, the operand of a `return`, a call argument, an object field. My first attempt registered it alongside `switchStatementParser` and it never fired once, because the body-node parser never sees it. Wrapping the expression parser reaches all four positions at once, and there are tests for each.

`?.` and `??` are both real Agency operators, so the probe requires a bare `?` — `not(oneOf(".?"))` — followed by a complete `? expr :`. Optional parameters (`def f(a?: number)`) and optional properties (`{ a?: string }`) cannot reach it: neither is in expression position, and in both the `?` is followed immediately by `:` with no expression between.

## Three messages gain an example

Chosen where the *shape* is what the reader is missing, rather than a token at a known place.

**`match` arms** — Agency's most distinctive construct, and the one agents arrive at from `switch`. Now shows patterns, a guard, a block arm and the catch-all, and states the arm grammar.

**`if ... then ... else`** — names itself as the ternary replacement, since the reader is often someone who just learned there is no `? :`, and gives the else-required and no-nesting rules.

**Handler body** — shows the `handle { } with (intr) { }` shape, the four verbs, and the `with approve` shorthand, none of which were guessable from "expected `{` to open handler body".

## Left alone deliberately

The positional messages — "expected `,` between parameters", "expected `{` to open while loop body", "unterminated destructive block". They name the exact missing token at a known place; an example would be noise. Same for the list-comprehension family, whose syntax matches Python's and whose messages already name the missing piece.

The `static` family, `class`, `interface extends` and the import messages already carry inline examples. They were the model for these.

## A test that keeps the examples honest

`errorExamples.test.ts` asserts that **every code example embedded in a parser error message is itself valid Agency**, including the two shipped in #757. An example that does not parse is worse than no example — it teaches the wrong thing to exactly the reader who is already stuck.

## Verification

- Full unit suite: **10186 passed, 0 failed**.
- `make` builds the whole stdlib and every example through the wrapped expression parser.
- All 16 perf tests pass, including `parse.perf` still scaling linearly — relevant because the wrapper adds a peek to every successful expression parse. The expensive branch only runs once a bare `?` is actually present.
- Swept every `.agency` file in `tests/`, `examples/` and `stdlib/` for a bare `?`: 234 files match, and every one is inside a string literal or a comment. No false-positive surface.

### DIFF
--- a/packages/agency-lang/lib/parsers/errorExamples.test.ts
+++ b/packages/agency-lang/lib/parsers/errorExamples.test.ts
@@ -1,0 +1,119 @@
+import { describe, it, expect } from "vitest";
+import { parseAgency } from "@/parser.js";
+
+function failure(src: string) {
+  const parsed = parseAgency(src, {}, false);
+  if (parsed.success) throw new Error(`expected a failed parse for: ${src}`);
+  return parsed.message ?? "";
+}
+
+function parses(src: string) {
+  return parseAgency(src, {}, false).success;
+}
+
+describe("ternaries are refused with the if-then-else form", () => {
+  it.each([
+    ["a const value", `node main() { const y = x ? 1 : 2 }`],
+    ["a return value", `def f(x: boolean): number { return x ? 1 : 2 }`],
+    ["a call argument", `node main() { print(x ? 1 : 2) }`],
+    ["an object field", `node main() { const o = { a: x ? 1 : 2 } }`],
+  ])("catches a ternary as %s", (_name, src) => {
+    const message = failure(src);
+    expect(message).toMatch(/no ternary/);
+    expect(message).toContain("if isProd then");
+  });
+
+  it("reports a real position rather than the body catch-all", () => {
+    const message = failure(`node main() { const y = x ? 1 : 2 }`);
+    expect(message).toMatch(/^Line \d+, col \d+:/);
+    expect(message).not.toContain("expected node body");
+  });
+
+  // `?.` and `??` are real Agency operators; only a bare `?` is a ternary.
+  it.each([
+    ["optional chaining", `node main() { const o = { a: 1 }\nprint(o?.a) }`],
+    ["nullish coalescing", `node main() { const y = a ?? 1 }`],
+    ["an optional parameter", `def f(a?: number) { print(a) }\nnode main() { f(1) }`],
+    ["an optional property", `type T = { a?: string }\nnode main() { print(1) }`],
+    ["a question mark in a string", `node main() { const s = "a ? b : c" }`],
+    ["an object literal", `node main() { const o = { a: 1, b: 2 } }`],
+    ["match arms", `node main() { match (x) { 1 => print(1) _ => print(2) } }`],
+  ])("leaves %s alone", (_name, src) => {
+    expect(parses(src)).toBe(true);
+  });
+});
+
+describe("messages carry a worked example", () => {
+  it("match arms show patterns, a guard, a block arm and the catch-all", () => {
+    const message = failure(`node main() { match (x) { 1: print(1) } }`);
+    expect(message).toContain("match (shape)");
+    expect(message).toContain("if (side > 0) =>");
+    expect(message).toContain("_ =>");
+  });
+
+  it("if-then-else names itself as the ternary replacement", () => {
+    const message = failure(`node main() { const y = if x then 1 }`);
+    expect(message).toContain("if isProd then");
+    expect(message).toContain("ternary");
+  });
+
+  it("the handler body shows the `with` form and the verbs", () => {
+    const message = failure(`node main() { handle { read("a") } with (i) }`);
+    expect(message).toContain("} with (intr) {");
+    expect(message).toContain("approve()");
+    expect(message).toContain("with approve");
+  });
+});
+
+/**
+ * Every code example embedded in a parser error message must itself be valid
+ * Agency. An example that does not parse is worse than no example: it teaches
+ * the wrong thing to exactly the reader who is already stuck.
+ *
+ * Each entry is the example as it appears in its message, wrapped in the least
+ * scaffolding needed to stand alone.
+ */
+describe("every example in a parser message is valid Agency", () => {
+  it.each([
+    ["match arms", `type Shape = { kind: "circle", r: number } | { kind: "square", side: number }
+node main(shape: Shape) {
+  const area = match (shape) {
+    { kind: "circle", r } => 3.14 * r * r
+    { kind: "square", side } if (side > 0) => side * side
+    _ => {
+      print("unknown")
+      return 0
+    }
+  }
+  print(area)
+}`],
+    ["if-then-else, also quoted by the ternary message", `node main(isProd: boolean) {
+  const label = if isProd then "Production" else "Local"
+  print(label)
+}`],
+    ["handler", `node main() {
+  handle {
+    read("./notes.md")
+  } with (intr) {
+    if (intr.effect == "std::read") { return approve() }
+    return reject()
+  }
+}`],
+    ["the C-style-for alternatives", `node main(items: string[]) {
+  for (i in range(0, 10)) { print(i) }
+  for (item in items) { print(item) }
+  for (item, i in items) { print(i, item) }
+  const doubled = [x * 2 for x in items]
+  print(doubled)
+}`],
+    ["the switch replacement", `node main(x: string) {
+  match (x) {
+    "a" => doThing()
+    "b" => doOtherThing()
+    _   => fallback()
+  }
+}`],
+  ])("%s", (_name, example) => {
+    expect(parses(example)).toBe(true);
+  });
+});
--- a/packages/agency-lang/lib/parsers/errorExamples.test.ts
+++ b/packages/agency-lang/lib/parsers/errorExamples.test.ts
@@ -1,5 +1,24 @@
 import { describe, it, expect } from "vitest";
 import { parseAgency } from "@/parser.js";
+import {
+  C_STYLE_FOR_MESSAGE,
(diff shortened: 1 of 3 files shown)


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3652,6 +3661,48 @@ export const exprParser: Parser<Expression> = label("an expression", memo("exprP
...
+ * parser is what reaches all of those at once.
+ *
+ * Without it the author gets the enclosing `parseError`'s catch-all, `expected
+ * node body`, with no position and no hint.
+ *
+ * Cost is a single character peek on every successful expression parse; the
+ * expensive branch only runs once a bare `?` is actually sitting there.
+ */
+const ternaryRefusal = (rest: string): ParserFailure | null => {
+  const afterExpr = optionalSpaces(rest);
```
egonSchiele: `optionalSpaces` is `many(oneOf(" \t"))` (line 192) — it does not cross newlines. So the refusal only fires when the `?` is on the same line as the condition, and the most common way a model writes a ternary is not:

```ts
const y = cond
  ? a
  : b
```

That is what Prettier does to any ternary past the line limit, so it is the shape a model has seen most often. Here `result.rest` begins with `\n`, `optionalSpaces` consumes nothing, `char("?")` fails, and the author falls back to the `expected node body` catch-all this parser was written to replace.

`optionalSpacesOrNewline` on line 3679 fixes it, and I do not think it costs anything: a statement cannot legally begin with `?`, so a `?` on the following line is unambiguously a continuation. The same swap is needed on line 3686 (inside `tail`), since the `:` is usually on its own line too.

Worth a test for the wrapped form either way — all four cases in `errorExamples.test.ts` are single-line.

packages/agency-lang/lib/parsers/errorExamples.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Every code example embedded in a parser error message must itself be valid
+ * Agency. An example that does not parse is worse than no example: it teaches
+ * the wrong thing to exactly the reader who is already stuck.
+ *
+ * Each entry is the example as it appears in its message, wrapped in the least
+ * scaffolding needed to stand alone.
+ */
+describe("every example in a parser message is valid Agency", () => {
```
egonSchiele, line 132: The idea here is the best part of the PR, but the test cannot do what the docstring says.

It validates a **hand-copied transcription** of each example, not the string in the message. Nothing connects the two, so editing `SWITCH_MESSAGE` or `C_STYLE_FOR_MESSAGE` in `parsers.ts` leaves this passing on a stale copy — which is exactly the "example that does not parse" failure the comment describes, just deferred.

To be fair: I checked the current copies and they are accurate. I diffed the `C-style-for alternatives` entry line by line against `C_STYLE_FOR_MESSAGE` on this branch, including the comprehension line, and they match. So this is about drift, not a present bug.

Closing it is not much work, since the examples are already uniformly indented by two spaces inside the messages:

```ts
function exampleFrom(message: string): string {
  return message
    .split("\n")
    .filter((l) => l.startsWith("  "))
    .map((l) => l.slice(2))
    .join("\n");
}
```

Then each case asserts the scaffolded source **contains** the extracted block, and parses it. The scaffolding stays hand-written, which it has to be, but the example itself can no longer drift without failing the test.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3652,6 +3661,48 @@ export const exprParser: Parser<Expression> = label("an expression", memo("exprP
...
+    optionalSpaces,
+    char(":"),
+  )(marker.rest);
+  if (!tail.success) return null;
+  return committedFailure(TERNARY_MESSAGE, tail.rest);
+};
+
+export const exprParser: Parser<Expression> = (input: string) => {
+  const result = _exprParserBase(input);
+  if (!result.success) return result;
```
egonSchiele, line 3702: Two things about writing the parse-state slot from here.

**Nested ternaries anchor in the wrong place.** For `const y = a ? b ? c : d : e`:

1. The outer call parses `a`, and `ternaryRefusal` sees the `?`.
2. Its `tail` calls `exprParser` on `b ? c : d : e`. That inner call parses `b`, its own refusal fires, sets `getParseState().committedFailure`, and returns a failure.
3. So the outer `tail` fails, `ternaryRefusal` returns `null`, and the outer `exprParser` returns **success** for just `a`.

The message that eventually surfaces is the right one, because the inner call already wrote the slot — but it is anchored at the inner `?`, and a successful expression parse has left a committed failure behind in the parse state. Worth a test, since the behaviour is accidental rather than designed.

**The general version is the part I would want a test for.** Every other writer of this slot — `bodyDeclarationParser`, `reservedClassParser`, `switchStatementParser`, `cStyleForParser` — fires on an unmistakable statement-level shape that cannot be part of a speculative branch. This one runs on every expression, including inside `or(...)` alternatives that get discarded. If a discarded branch writes the slot and the parse later fails somewhere unrelated, the ternary message wins the reporting contest over the real error.

I could not construct that case by reading, and it may well be unreachable. But it is a new kind of exposure for this mechanism, and one test that a ternary inside a backtracking context (a match arm alternative, say) does not hijack a later unrelated failure would settle it.

Two nits in the same function:

- Line 3680, `if (!afterExpr.success) return null;` is dead. `optionalSpaces` is `many(...)`, which succeeds on any input. Same shape as the `elseClauseParser` guard in #756.
- Line 3683 builds `seqC(char("?"), not(oneOf(".?")))` on **every successful expression parse**, which is the hottest path in this file. Hoisting it to a module-level `const` costs nothing and removes the allocation.

egonSchiele (commented):
Read the diff and the tests, and checked the claims against the branch. The judgment calls here are good, and two of them are worth naming.

**Wrapping `exprParser` rather than joining the statement-level probes is right**, and the description explaining that the first attempt was registered next to `switchStatementParser` and never fired once is more useful than a clean story would have been. A ternary genuinely is never a statement, and the four positions the tests cover fall out of the wrapping rather than being enumerated by hand.

**The `?.` / `??` precision is careful**, and the negative cases are the ones that matter — optional parameters and optional properties are exactly what a naive `char("?")` probe would break, and both are tested.

**Sweeping every `.agency` file for a bare `?` before shipping this** is the right instinct for a change that touches the hottest parser in the file.

Three things worth acting on before merge, in order:

1. **Multi-line ternaries are not caught**, which is the shape Prettier produces and therefore the one a model has seen most. One-word fix, inline.
2. **The examples-are-honest test cannot catch drift** — it checks hand-copied transcriptions rather than the strings in the messages. The current copies are accurate (I verified the C-style-for entry line by line), so this is about the future, not a present bug.
3. **Writing the parse-state committed-failure slot from a parser that runs on every expression** is a new kind of exposure for that mechanism. Includes a concrete nested-ternary case that behaves accidentally rather than by design.

Plus two nits in `ternaryRefusal`: a dead guard and a per-call parser allocation in the hot path.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3588,7 +3589,15 @@ const atomWithIs: Parser<Expression> = (input: string) => {
 // Operator table: highest precedence first.
 // Multi-char operators must come before their single-char prefixes
 // (e.g., *= before *, <= before <).
-export const exprParser: Parser<Expression> = label("an expression", memo("exprParser", buildExpressionParser<Expression>(
+const TERNARY_MESSAGE =
```
egonSchiele: move message constants to separate file to keep this file smaller

egonSchiele: Instead of concatenating like this, lets use multi-line strings, or typestache templates.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -3652,6 +3661,48 @@ export const exprParser: Parser<Expression> = label("an expression", memo("exprP
   parenParser,
 )));
 
+/**
```
egonSchiele, line 3670: Is such a long comment necessary?

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4184,7 +4235,21 @@ export const matchBlockExprParser = label("a match expression", withLoc(seqC(
   char("{"),
   captureCaptures(
     parseError(
-      "expected match cases of the form `value => expression` separated by `;` or newlines, followed by `}`",
+      "expected match cases of the form `value => expression`, separated by " +
```
egonSchiele: extract to constant

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -5246,7 +5311,18 @@ const inlineHandlerParser: Parser<HandleBlock["handler"]> = (input) => {
     optionalSpaces,
     captureCaptures(
       parseError(
-        "expected `{` to open handler body",
+        "expected `{` to open handler body:\n" +
```
egonSchiele: extract

PR #759, opened 2026-08-01
Accept arrow functions as block arguments

egonSchiele:
The last item from the syntax-variation audit. Arrow functions in argument position are now a second spelling of a block argument.

| Written | Canonical |
|---|---|
| `map(xs, (n) => n * 2)` | `map(xs, \n -> n * 2)` |
| `map(xs, n => n * 2)` | `map(xs, \n -> n * 2)` |
| `mapWithIndex(xs, (n, i) => n * i)` | `mapWithIndex(xs, \(n, i) -> n * i)` |
| `twice(() => 5)` | `twice(\ -> 5)` |
| `map(xs, (n) => { ... })` | `map(xs) as n { ... }` |

This is the acceptance route rather than the refusal route used for `switch`, C-style `for` and ternaries — those name constructs Agency deliberately does not have, whereas this is Agency's own block argument written in a different hand. `\n -> n * 2` is the least guessable syntax in the language, so it is the one most worth meeting halfway.

## Both bodies

An expression body produces the node `inlineBlockParser` builds, wrapping the expression in a synthetic return. A braced body produces the node the `as` form builds. Models write both, and the two map cleanly onto constructs that already exist, so there was no reason to accept only one.

Nothing records which spelling was used, so `agency fmt` prints the canonical form and `AgencyGenerator` needed no changes — the same rule as #750 and #756. Parameter parsing reuses `blockParamsParser`, which already handled `(a, b)` and a bare identifier.

## Ordering, and why it backtracks cleanly

The arrow parser has to be tried before `exprParser` in argument position, because `(n)` and `n` are both perfectly ordinary expressions. It commits only once the `=>` is actually there, so everything else falls through. Tested to be unaffected: parenthesized arguments, plain arguments, nested calls, named arguments, comparisons, object literals, function types in signatures, and both existing block forms.

Match arms use `=>` as well, but the arrow parser only runs in argument position, so the two never meet. There is a test.

## One thing I checked rather than assumed

`f(match (x) { 1 => 2 _ => 3 })` fails to parse. I stashed the change and confirmed it fails **identically** without it — this is the documented rule that a `match` expression is allowed only as a `const`/`let` value or a `return`, not nested inside another value. Not a regression, and out of scope here.

## Verification

- Full unit suite: **10213 passed, 0 failed**.
- `make` builds the whole stdlib and every example through the changed argument parser.
- All 16 perf tests pass — relevant because this adds an alternative to the argument list, which is a hot path.

### DIFF
--- a/packages/agency-lang/lib/parsers/arrowBlocks.test.ts
+++ b/packages/agency-lang/lib/parsers/arrowBlocks.test.ts
@@ -53,6 +53,81 @@ describe("arrow functions parse to the canonical block AST", () => {
   ])("%s", (_name, arrow, canonical) => {
     expect(program(arrow)).toEqualWithoutLoc(program(canonical));
   });
+
+  it.each([
+    [
+      "typed parameters",
+      `node main() { const y = map(xs, (n: number) => n * 2) }`,
+      `node main() { const y = map(xs, \\(n: number) -> n * 2) }`,
+    ],
+    [
+      "multiple parameters with a braced body",
+      `node main() { const y = mapWithIndex(xs, (n, i) => { return n * i }) }`,
+      `node main() { const y = mapWithIndex(xs) as (n, i) { return n * i } }`,
+    ],
+    // Prettier wraps an arrow whose body passes the line limit, so this is a
+    // shape models produce. The canonical `\\n ->` form still does not accept
+    // it; this deliberately goes further, since meeting that habit is the point.
+    [
+      "a body on the next line",
+      `node main() { const r = map(items, (item) =>\n  item * 2\n) }`,
+      `node main() { const r = map(items, \\item -> item * 2) }`,
+    ],
+    // `async` is accepted and discarded, like `await` and `sync`.
+    [
+      "an async arrow",
+      `node main() { const y = map(xs, async (n) => n * 2) }`,
+      `node main() { const y = map(xs, \\n -> n * 2) }`,
+    ],
+  ])("%s", (_name, arrow, canonical) => {
+    expect(program(arrow)).toEqualWithoutLoc(program(canonical));
+  });
+
+  // JavaScript famously reads these braces as a block. Here `braced` tries
+  // first, `bodyParser` declines, and it falls through to `exprParser`, which
+  // reads an object literal. Pinned because it depends on that ordering.
+  it("reads a braced object literal body as an object, not a block", () => {
+    expect(program(`node main() { const y = f((n) => { a: 1 }) }`))
+      .toEqualWithoutLoc(program(`node main() { const y = f(\\n -> { a: 1 }) }`));
+  });
+
+  it("requires `()` for a zero-parameter arrow", () => {
+    // `blockParamsParser` returns an empty list without consuming, so a bare
+    // `=>` would otherwise read as a zero-parameter block.
+    expect(parses(`node main() { f(=> 1) }`)).toBe(false);
+    expect(parses(`node main() { f(() => 1) }`)).toBe(true);
+    // The canonical zero-parameter form is unaffected.
+    expect(parses(`node main() { f(\\ -> 1) }`)).toBe(true);
+  });
+
+  // A block is not a value, so it cannot be a block's body either. The
+  // canonical form behaves identically; pinned so a reorder cannot change it
+  // silently.
+  it("does not support currying", () => {
+    expect(parses(`node main() { const y = map(xs, (a) => (b) => a + b) }`)).toBe(false);
+    expect(parses(`node main() { const y = map(xs, \\a -> \\b -> a + b) }`)).toBe(false);
+  });
+});
+
+describe("a block assigned to a variable gets a message, not the catch-all", () => {
+  it.each([
+    ["the arrow spelling", `node main() { const double = (n) => n * 2 }`],
+    ["the canonical spelling", `node main() { const double = \\n -> n * 2 }`],
+    ["let rather than const", `node main() { let double = (n) => n * 2 }`],
+    ["an async arrow", `node main() { const double = async (n) => n * 2 }`],
+  ])("catches %s", (_name, src) => {
+    const parsed = parseAgency(src, {}, false);
+    expect(parsed.success).toBe(false);
+    if (parsed.success) return;
+    expect(parsed.message).toMatch(/blocks are arguments, not values/);
+    expect(parsed.message).toContain("def double");
+  });
+
+  it("leaves an ordinary assignment alone", () => {
+    expect(parses(`node main() { const double = 2 }`)).toBe(true);
+    expect(parses(`node main() { const f = someFn }`)).toBe(true);
+    expect(parses(`node main() { const eq = a == b }`)).toBe(true);
+  });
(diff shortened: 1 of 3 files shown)


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4012,6 +4022,70 @@ export const inlineBlockParser: Parser<BlockArgument> = memo(
...
+ * body produces the same node `inlineBlockParser` does, wrapping the expression
+ * in a synthetic return; a braced body produces the same node the `as` form
+ * does. Nothing records which spelling was used, so `agency fmt` prints the
+ * canonical one.
+ *
+ * MUST be tried before `exprParser` in argument position, since `(n)` and `n`
+ * both parse as ordinary expressions. It only commits once the `=>` is there,
+ * so anything else backtracks cleanly.
+ */
+const arrowBlockParser: Parser<BlockArgument> = memo(
```
egonSchiele, line 4036: `arrowBlockParser` is registered only in `argumentListParser` and `namedArgumentParser`, so the other shape models write constantly is not covered:

```ts
const double = (n) => n * 2
```

That still produces `expected node body` with no position and no hint.

I checked before calling this a defect, and it is not one — the canonical form has the same restriction. `const double = \n -> n * 2` fails today with the identical message, so blocks really are argument-position-only in Agency and refusing the arrow there is correct. Verified against the current build:

| Source | Result |
|---|---|
| `map(xs, \n -> n * 2)` | parses |
| `const double = \n -> n * 2` | fails, `expected node body` |

So the parser is right. What is missing is the message. This is exactly the situation #757 and #758 built the refusal route for: a construct Agency deliberately does not have, currently landing on a catch-all. A probe that sees `=>` after a `const`/`let` value and says "Agency blocks are arguments, not values — pass it directly, or use `def`" would close the loop, and the shape is already detectable here.

At minimum it belongs in the PR description's limitations, next to the `match`-nested-in-a-value note. Right now the description reads as though arrow functions are accepted, and a reader would reasonably expect the assignment form to work.

Same question for the curried form, `map(xs, (a) => (b) => a + b)`. The inner arrow is the outer's expression body, not an argument, so it falls through to `exprParser` and fails. I confirmed the canonical form behaves the same way, so again this is parity rather than a regression — but it is untested and worth pinning so a future reorder does not silently change it.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4012,6 +4022,70 @@ export const inlineBlockParser: Parser<BlockArgument> = memo(
...
+  "arrowBlockParser",
+  (input: string): ParserResult<BlockArgument> => {
+    const head = seqC(
+      capture(blockParamsParser, "params"),
+      optionalSpaces,
+      str("=>"),
```
egonSchiele: `optionalSpaces` is `many(oneOf(" \t"))`, so it does not cross a newline. An arrow whose body starts on the next line is not accepted:

```ts
const result = map(items, (item) =>
  item.value * 2
)
```

`head` succeeds, `braced` declines on the newline, and `exprParser` is handed input beginning with `\n` and fails.

I checked whether this is a regression and it is not — the canonical form has the identical gap. Verified against the current build:

| Source | Result |
|---|---|
| `map(xs, \n -> n * 2)` | parses |
| `map(xs, \n ->` newline `n * 2)` | fails, `expected node body` |

So this is consistent with the construct it mirrors, which is a fair defence in most PRs. It is a weaker defence in this one, because the whole point is meeting a model where it is, and this is precisely what Prettier does to an arrow whose body pushes past the line limit. The braced form is fine — `=> {` keeps the brace on the same line — so this only bites the expression body.

`optionalSpacesOrNewline` in both places would cover it. If you would rather keep parity with `\n ->` and fix both later, that is reasonable too; I mainly want it to be a decision rather than an accident.

packages/agency-lang/lib/parsers/arrowBlocks.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    ]) {
+      const once = formatSource(src);
+      expect(once).not.toBeNull();
+      expect(formatSource(once as string)).toBe(once);
+    }
+  });
+});
+
+describe("things that look like arrows but are not", () => {
+  it.each([
```
egonSchiele, line 159: The negative cases here are well chosen — the function-type-in-a-signature case is the one I would have worried about most, and pinning both existing block forms is right.

Four shapes worth adding, all cheap:

**Typed parameters.** `map(xs, (n: number) => n * 2)` is the most likely form from a model writing TypeScript. I traced it and it already works — `blockParamParser` carries an `optional(... char(":") ... variableTypeParser)` — but nothing pins that, and it would be easy to lose in a future rewrite of the params path.

**An object literal as the expression body.** `f((n) => { a: 1 })` is the case JavaScript famously gets wrong, parsing the braces as a block. Here `braced` tries first, `bodyParser` declines, and it falls through to `exprParser`, which reads it as an object literal — the behaviour you want, arrived at by ordering. Worth a test precisely because it depends on that ordering.

**An async arrow.** `map(xs, async (n) => n * 2)` — I do not know what this does today, and `async` interacts with `asyncValueAccessParser`, so the answer is not obvious from reading. Either outcome is fine; it should just be known.

**A multi-parameter braced body.** The table covers `(n, i) => n * i` with an expression body and `(n) => { ... }` with a braced one, but not the two combined.

egonSchiele (commented):
Read the diff, and checked the two structural claims against the branch rather than taking them on trust. Both hold:

**The expression body really does produce the same node `inlineBlockParser` builds.** `{ type: "blockArgument", inline: true, params, body: [returnStatement] }`, field for field.

**The braced body really does produce the same node the `as` form builds.** `blockArgumentParser` sets `inline: false` and captures `params`/`body`, and neither carries a `loc`, so the two are identical rather than merely similar. That matters more here than usual, since the whole design rule is that the AST cannot tell which spelling was used.

Placing the parser before `exprParser` in argument position, and committing only once `=>` is present, is the right ordering, and the negative-case block is the right way to prove it backtracks cleanly.

I also liked the note about `f(match (x) { ... })` — stashing the change to confirm the failure is identical without it is the correct way to answer "is this mine?", and the answer is worth having written down.

Three comments below. None is a defect in what is here; two are gaps in what is not, and I verified against the current build that both mirror limitations the canonical `\n ->` form already has, so neither is a regression:

1. The assignment form `const double = (n) => n * 2` still lands on the bare catch-all. The parser is right to refuse it, but this is the situation #757 and #758 built the refusal route for.
2. An arrow whose expression body starts on the next line is not accepted — which is what Prettier produces.
3. Four cheap test additions, one of which (typed parameters) already works and is just unpinned.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -4012,6 +4022,70 @@ export const inlineBlockParser: Parser<BlockArgument> = memo(
   ),
 );
 
+/**
```
egonSchiele, line 4026: extremely long comment. Needed? File is already 5000 lines long. Can it be shortened?

PR #760, opened 2026-08-01
feat(cli): agency deploy — upload an agent to a hosted statelog

egonSchiele:
## What

`agency deploy <file>` uploads an agent to a hosted [statelog](https://github.com/egonSchiele/statelog) so it can be served over HTTP — the hosted counterpart to `agency serve http`. It resolves the target from `agency.json`'s `log` section, compiles the agent locally for fast feedback, uploads the source, and prints the serve endpoints plus a ready-to-run curl for each.

Built prototype-first (branch `proto/agency-deploy`); this is the real version. Decisions settled with the owner during prototyping: reuse the `log` config, API key env-first, compile locally first, hardcode statelog's upload API.

## Architecture — "what" split from "how"

The orchestrator (`deploy.ts`) reads as a recipe; each stage is a single-concept module that hides its mechanics behind a typed result:

| File | Concept |
|---|---|
| `target.ts` | resolve `{host, projectId, apiKey}` (flag > `log.*` > env; key env-first) |
| `bundle.ts` | collect the source to upload + local compile pre-flight |
| `uploadClient.ts` | **the only file that knows statelog's API** (upload + `/list` fetch) |
| `curlExamples.ts` | pure: manifest → curl commands |
| `render.ts` | terminal output (termcolors) |
| `deploy.ts` | the orchestrator + `DeployOutcome` |

Reuses existing machinery rather than reinventing: `loadConfig`, `compileSource`, `importPaths` (`agencyImportTargets`/`nonAgencyLocalImportTargets`), `termcolors`.

## Single-file for v1 (important)

Testing surfaced that **statelog can't host multi-file agents**: both its upload and serve paths compile via `compileSource`, which writes source to a temp dir and so can't resolve relative `.agency` imports. So deploy refuses local `.agency` imports (and local TS/JS interop) with a clear message. Multi-file is tracked in **statelog#9**; relaxing deploy afterward is deleting the refusal.

## The user-facing interface

```
agency deploy agent.agency          # deploy (STATELOG_API_KEY in env)
agency deploy agent.agency --dry-run # preview target + files, upload nothing
```
Options: `--host`, `--project`, `--api-key-env`, `--dry-run` (+ global `-c/--config`). Target config lives in `agency.json`'s `log` section; the API key is read from an env var, never a flag.

## Tests

`lib/cli/deploy/*.test.ts` — 19 tests across target resolution, bundle collection + refusals, curl generation, and the upload client (mocked `fetch`). Build clean, structural lint clean, 19/19 green. Smoke-tested end-to-end (dry-run, missing-key, TS-interop refusal, multi-file refusal).

Docs: `docs/site/cli/deploy.md` (+ nav entry).

### DIFF
--- a/packages/agency-lang/docs/site/.vitepress/config.mts
+++ b/packages/agency-lang/docs/site/.vitepress/config.mts
@@ -240,7 +240,6 @@ export default defineConfig({
             { text: "compile", link: "/cli/compile" },
             { text: "coverage", link: "/cli/coverage" },
             { text: "debug", link: "/cli/debug" },
-            { text: "deploy", link: "/cli/deploy" },
             { text: "doc", link: "/cli/doc" },
             { text: "eval", link: "/cli/eval" },
             { text: "eval-judge", link: "/cli/eval-judge" },
--- a/packages/agency-lang/docs/site/.vitepress/config.mts
+++ b/packages/agency-lang/docs/site/.vitepress/config.mts
@@ -240,6 +240,7 @@ export default defineConfig({
             { text: "compile", link: "/cli/compile" },
             { text: "coverage", link: "/cli/coverage" },
             { text: "debug", link: "/cli/debug" },
+            { text: "deploy", link: "/cli/deploy" },
             { text: "doc", link: "/cli/doc" },
             { text: "eval", link: "/cli/eval" },
             { text: "eval-judge", link: "/cli/eval-judge" },
--- a/packages/agency-lang/docs/site/cli/deploy.md
+++ b/packages/agency-lang/docs/site/cli/deploy.md
@@ -1,0 +1,62 @@
+---
+title: Deploying an agent
+description: Documents the `agency deploy` command, which uploads an agent's source to a hosted statelog so it can be served over HTTP, reusing the log config for the target.
+---
+
+# Deploying an agent
+
+`agency deploy` uploads an agent to a hosted [statelog](https://github.com/egonSchiele/statelog) so it can be served over HTTP — a hosted equivalent of `agency serve http`. Statelog compiles the source, stores it, and exposes each exported node and function at a `/serve/...` URL. Every hosted run also traces back into statelog automatically.
+
+```
+agency deploy agent.agency
+```
+
+On success it prints the agent's serve endpoints and a ready-to-run curl command for each.
+
+## Where it deploys to
+
+The target — statelog host, project, and API key — comes from the `log` section of your `agency.json`, the same config observability already uses:
+
+```json
+{
+  "log": {
+    "host": "https://statelog.example.com",
+    "projectId": "my-project"
+  }
+}
+```
+
+The **API key is read from an environment variable**, never from `agency.json` or a flag, so it stays out of version control and process listings. By default `agency deploy` reads `STATELOG_API_KEY`:
+
+```
+export STATELOG_API_KEY=sk_...
+agency deploy agent.agency
+```
+
+## Options
+
+- `--host <url>` — statelog host, overriding `log.host`.
+- `--project <slug>` — project slug, overriding `log.projectId`.
+- `--api-key-env <name>` — environment variable to read the API key from (default: `STATELOG_API_KEY`).
+- `--dry-run` — resolve the target and validate the agent, printing exactly what would be uploaded, without sending anything.
+
+The global `-c, --config <path>` option selects which `agency.json` to read.
+
+## What gets uploaded
+
+`agency deploy` compiles the agent locally first, so obvious errors surface before any upload. It then sends the source; statelog compiles it again server-side.
+
+**Deploy is single-file for now.** An agent that imports another local `.agency` file, or local TypeScript/JavaScript interop, is refused with a clear message — statelog compiles each uploaded file in isolation, so multi-file hosting isn't supported yet (tracked in statelog#9).
+
+## Running a deployed agent
+
+Use the printed curl commands, or hit the endpoints directly with the same API key:
+
+```
+curl -s -X POST "https://statelog.example.com/serve/<user>/<project>/agent/node/main" \
+  -H "Authorization: Bearer $STATELOG_API_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{"message":"hello"}'
+```
+
+Responses come back as `{ "success": true, "value": ... }`. Tool failures are reported in the body (`{ "success": false, "error": ... }`) with HTTP 200, so check the `success` field. If a node raises an interrupt, continue by POSTing `{ interrupts, responses }` to the agent's `/resume` endpoint.
--- a/packages/agency-lang/lib/cli/deploy/render.ts
+++ b/packages/agency-lang/lib/cli/deploy/render.ts
@@ -1,7 +1,11 @@
-// Turning a DeployOutcome into terminal output. Presentation only — it reads
-// the typed outcome and prints; no deploy logic lives here.
+// Turning a DeployOutcome into terminal output. Presentation only: it builds the
+// coloured blocks and hands them to the deployReport template, which owns the
+// dry-run vs deployed branch. No deploy logic lives here. Each block carries its
+// own leading blank line, so the template concatenates them without whitespace
+// fiddling.
 
 import { color } from "@/utils/termcolors.js";
+import renderDeployReport from "@/templates/cli/deployReport.js";
 import type { DeployOutcome, DeployPlan } from "./deploy.js";
 import { serveBaseUrl } from "./uploadClient.js";
 import type { Manifest } from "./uploadClient.js";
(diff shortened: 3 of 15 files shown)


packages/agency-lang/lib/cli/deploy/uploadClient.ts
```diff
@@ -0,0 +1,106 @@
...
+): Promise<Manifest | undefined> {
+  const listUrl = endpointUrls.find((url) => url.endsWith("/list"));
+  if (!listUrl) {
+    return undefined;
+  }
+  try {
+    const response = await fetch(listUrl, {
+      headers: { Authorization: `Bearer ${apiKey}` },
+    });
+    return (await response.json()) as Manifest;
```
egonSchiele: This casts the `/list` body to `Manifest` without checking `response.ok` or the shape. A non-200 response (or any body that isn't `{nodes, functions}` — e.g. an `{error}` envelope) is returned as a "manifest," and `curlExamples` then does `manifest.nodes.map(...)` → uncaught `TypeError`, crashing the CLI *after* the upload already succeeded.

Since this is explicitly best-effort, guard it: `if (!response.ok) return undefined;` and confirm `Array.isArray(body.nodes) && Array.isArray(body.functions)` before returning — otherwise return `undefined` so the deploy still reports success without curls.

packages/agency-lang/lib/cli/deploy/uploadClient.ts
```diff
@@ -0,0 +1,106 @@
...
+    envelope = (await response.json()) as UploadEnvelope;
+  } catch {
+    return { ok: false, error: `statelog returned a non-JSON response (HTTP ${response.status}).` };
+  }
+
+  if (!envelope || envelope.success === false) {
+    return { ok: false, error: envelope?.error ?? `Upload rejected (HTTP ${response.status}).` };
+  }
+
+  const endpointUrls = envelope.value.endpointUrls.map((relative) =>
```
egonSchiele: Two untrusted spots on the success path: `envelope.value.endpointUrls.map(...)` assumes `value.endpointUrls` exists on any `success: true` (a `{success:true}` without it throws), and the `new URL('/api/...', target.host)` up at line 33 runs *before* the fetch try/catch, so a malformed `--host` (missing scheme) throws uncaught rather than returning your friendly `Could not reach …` error. Consider `Array.isArray(envelope.value?.endpointUrls)` here, and validating/normalizing the host in `resolveDeployTarget` (or wrapping the URL construction).

packages/agency-lang/lib/cli/deploy/target.ts
```diff
@@ -0,0 +1,98 @@
...
+ *  baked in agency.json `log.apiKey`, else the conventional STATELOG_API_KEY. */
+function resolveApiKey(
+  log: LogConfig,
+  apiKeyEnv: string | undefined,
+  env: NodeJS.ProcessEnv,
+): { value: string | undefined; from: string } {
+  if (apiKeyEnv) {
+    return { value: env[apiKeyEnv], from: `$${apiKeyEnv}` };
+  }
+  if (log.apiKey) {
```
egonSchiele: This reads the API key straight from `agency.json`'s `log.apiKey`, and ranks it *above* the `$STATELOG_API_KEY` default — but `deploy.md` tells users the key is read from an env var "never from `agency.json` … so it stays out of version control." A committed `log.apiKey` silently wins, which is the opposite of that promise. Either drop this branch (env-only, matching the security posture and the no-`--api-key`-flag choice) or, if reusing the `log` key is intentional, correct the doc and don't prefer a committed key over the env var.

egonSchiele (commented):
## Review — `agency deploy`

**Overview.** Adds `agency deploy <file>`: resolve a statelog target from `agency.json`'s `log` config (+ flag/env overrides), collect a single-file bundle, compile it locally as pre-flight, upload the source, then print serve endpoints and ready-to-run curls. Refuses multi-file / TS-interop agents statelog can't host yet.

**Strengths.**
- Excellent decomposition: the what/how split is real, each module is single-concept, and `uploadClient.ts` as the *sole* statelog-API-aware file is exactly the right seam. Reuses `loadConfig`/`compileSource`/`importPaths`/`termcolors` rather than reinventing.
- Strong security defaults on the key: env-first, `$KEY` placeholder in curls (never the real key), redaction in the summary, and no `--api-key` flag so it never hits argv.
- The single-file refusal is well-reasoned and clearly messaged, with the tracking issue in the error text. Good test coverage (19 tests incl. the refusal paths and mocked fetch). I verified `compileSource` returns `errors: string[]`, so `validateBundleCompiles`'s joined output is readable.

**Findings.**

1. **The post-upload path trusts the server/host response and can crash *after* a successful deploy (medium).** Three spots cast or construct without guarding:
   - `fetchManifest` does `(await response.json()) as Manifest` with no `response.ok` or shape check. A non-200 `/list` (or any non-manifest body — e.g. `{error}`) is returned as a "manifest," and `curlExamples` then calls `manifest.nodes.map(...)` → uncaught `TypeError`, turning a completed upload into a CLI stack trace.
   - `envelope.value.endpointUrls.map(...)` assumes `value.endpointUrls` is present on any `success: true`; a `{success:true}` without it crashes.
   - `new URL('/api/...', target.host)` is built *before* the fetch try/catch, so a malformed `--host` (e.g. no scheme) throws uncaught.
   Suggest: guard `response.ok` + `Array.isArray(nodes/functions)` in `fetchManifest` (return undefined otherwise — it's already best-effort); defensively check `endpointUrls` is an array; and validate/normalize the host in `resolveDeployTarget` (or wrap the `new URL`). See inline.

2. **Docs say the key is "never from `agency.json`," but the code reads (and prefers) `log.apiKey` (medium — security/doc mismatch).** `resolveApiKey`'s precedence is `--api-key-env` > `log.apiKey` > `$STATELOG_API_KEY`, and `target.test.ts` asserts the key is read from `agency.json log` by default. But `deploy.md:29` states the key comes from an env var "never from `agency.json` or a flag … so it stays out of version control." So a committed `log.apiKey` silently wins over the env convention — the exact thing the doc promises can't happen. Reconcile: either drop the `log.apiKey` branch (env-only, matching the stated posture) or, if reuse of the `log` key is intentional, fix the doc and don't rank a committed key above `$STATELOG_API_KEY`. See inline.

3. **Manifest/base URL identified by `endsWith("/list")` also matches a node named `list` (low).** Both `serveBaseUrl` and `fetchManifest` pick the manifest URL via `endsWith("/list")`, but a node named `list` produces `/node/list`, which also ends in `/list`. That would fetch the wrong URL as the manifest and derive a wrong serve base. Match the exact file-level `…/<file>/list` instead. See inline.

4. **`renderEndpoints` labels every non-`/list` URL as "node" (low, cosmetic).** If the server ever returns `/function/*` endpoints in `endpointUrls`, they'll print as `node`. Minor.

**Verdict.** Clean, well-factored, and the security-sensitive parts (key handling) are mostly right — the one exception is the doc/behavior mismatch in #2, worth settling since it's a security claim. #1 is the robustness cluster to harden: the happy path is fine, but the tool shouldn't crash on a server hiccup once the upload already landed. #3/#4 are small.

packages/agency-lang/lib/cli/deploy/uploadClient.ts
```diff
@@ -0,0 +1,106 @@
...
+      }`,
+    );
+    return undefined;
+  }
+}
+
+/** The serve base URL (…/serve/:user/:project/:file) shared by an agent's
+ *  endpoints — the `/list` URL with its trailing segment removed. */
+export function serveBaseUrl(endpointUrls: string[]): string | undefined {
+  const listUrl = endpointUrls.find((url) => url.endsWith("/list"));
```
egonSchiele: `endsWith("/list")` also matches a node named `list` (`…/node/list`), so `serveBaseUrl` (and `fetchManifest`) could pick that as the manifest URL and strip the wrong suffix, yielding a bad base. The manifest is specifically the file-level `…/<file>/list`; match that exact suffix (or the endpoint equal to `<serveBase>/list`) rather than any URL ending in `/list`.

packages/agency-lang/docs/site/.vitepress/config.mts
```diff
@@ -240,6 +240,7 @@ export default defineConfig({
             { text: "compile", link: "/cli/compile" },
             { text: "coverage", link: "/cli/coverage" },
             { text: "debug", link: "/cli/debug" },
+            { text: "deploy", link: "/cli/deploy" },
```
egonSchiele: It's fine to create the documentation, but let's not link it yet. This feature is not ready for public consumption, so we don't want that showing up on public facing documentation.

packages/agency-lang/docs/site/cli/deploy.md
```diff
@@ -0,0 +1,62 @@
+---
```
egonSchiele, line 1: fine to leave this in

packages/agency-lang/lib/cli/deploy/render.ts
```diff
@@ -0,0 +1,59 @@
...
+    console.log(`\n${color.yellow("dry run")} — nothing uploaded. Re-run without ${color.bold("--dry-run")} to deploy.\n`);
+    return;
+  }
+  renderPlan(outcome.plan);
+  renderEndpoints(outcome.endpointUrls, outcome.manifest);
+}
+
+function renderPlan(plan: DeployPlan): void {
+  const { target, provenance, bundle } = plan;
+  console.log(`\n${color.bold("Deploy target")}`);
```
egonSchiele: This is really ugly code. Please use a typestache template instead. Check out lib/templates/ for examples and see the documentation here https://github.com/egonSchiele/typestache

packages/agency-lang/lib/cli/deploy/render.ts
```diff
@@ -0,0 +1,59 @@
...
+  console.log(`  ${color.dim("api key")} ${redactKey(target.apiKey)}   ${color.dim(provenance.apiKey)}`);
+
+  console.log(`\n${color.bold(`Files (${bundle.files.length})`)}   ${color.dim(`entrypoint: ${bundle.entrypoint}`)}`);
+  for (const file of bundle.files) {
+    console.log(`  ${color.cyan(file.name)}   ${color.dim(`${Buffer.byteLength(file.contents)} bytes`)}`);
+  }
+}
+
+function renderEndpoints(endpointUrls: string[], manifest: Manifest | undefined): void {
+  console.log(`\n${color.green("✓ deployed")}\n`);
```
egonSchiele: Same comment, all of these console logs are pretty hard to read. Please substitute these with a typestache template.

packages/agency-lang/lib/cli/deploy/target.ts
```diff
@@ -0,0 +1,98 @@
...
+  host?: string;
+  projectId?: string;
+  apiKey?: string;
+};
+
+export type ResolveTargetResult =
+  | { ok: true; target: DeployTarget; provenance: TargetProvenance }
+  | { ok: false; error: string };
+
+const DEFAULT_HOST = "http://localhost:1065";
```
egonSchiele: This default host doesn't make sense. No one else is going to have this app running locally. I think if a host is not specified, we should just throw an error.

packages/agency-lang/lib/cli/deploy/uploadClient.ts
```diff
@@ -0,0 +1,106 @@
...
+type UploadEnvelope =
+  | { success: true; value: { endpointUrls: string[] } }
+  | { success: false; error: string };
+
+/**
+ * Upload the bundle and return the agent's serve endpoints (absolute URLs).
+ * Best-effort fetches the `/list` manifest too, so the caller can print curl
+ * examples with real node parameters; a manifest fetch failure is not fatal.
+ */
+export async function uploadBundle(
```
egonSchiele, line 26: If possible, it would be great to print out a list of endpoints that have been created as well.

packages/agency-lang/scripts/agency.ts
```diff
@@ -396,6 +398,40 @@ export function createProgram(deps: CliDependencies = {}): Command {
       },
     );
 
+  program
+    .command("deploy")
```
egonSchiele: Let's mark this hidden for now so it doesn't show up in the CLI as an available command. It should still work if used though. We use commander for parsing CLI commands and I believe it has a "hidden" option.

egonSchiele:
Thanks — all addressed in f79c2ae. Point by point:

**1. Untrusted server responses (medium).** Hardened all three spots in `uploadClient.ts`:
- `readEndpointUrls` only returns URLs when `success === true` **and** `Array.isArray(value.endpointUrls)`; otherwise the deploy errors (via `rejectionMessage`) instead of crashing.
- `fetchManifest` now checks `response.ok` and `Array.isArray(nodes/functions)`, returning `undefined` on anything unexpected — the deploy still reports success, just without curls.
- The `new URL(..., host)` can no longer throw: the host is URL-validated at resolution time (see #5).

**2. API-key doc/behavior mismatch (medium).** Went env-only — dropped the `log.apiKey` branch entirely. The key is now read only from `--api-key-env` (default `$STATELOG_API_KEY`), matching the doc's "never from agency.json" posture and the no-`--api-key`-flag choice. Tests + `LogConfig` type updated (it no longer even has an `apiKey` field).

**3. `endsWith("/list")` matches a node named `list` (low).** Fixed. `manifestUrl` (used by both `fetchManifest` and `serveBaseUrl`) now matches the file-level `.../list` by excluding `/node/` and `/function/` URLs.

**4. Endpoint labels (low).** `endpointLabel` now derives the label from the path — `manifest` / `node` / `function` / `endpoint`.

**5. Default host `:1065` (your comment).** Removed. Host is now required (from `log.host` or `--host`) and validated as an `http(s)` URL — a missing or malformed host fails at resolution with a clear message, alongside any other missing fields.

**6. Render → typestache (your comment, x2).** Moved the terminal layout into `lib/templates/cli/deployReport.mustache`; `render.ts` now builds coloured blocks with small declarative builders (termcolors) and hands them to the template, which owns the dry-run/deployed branch.

**7. Print created endpoints (your comment).** These are already listed under "Serve endpoints" (labeled per #4), and it reads more clearly after the render refactor.

**8. Hidden command.** `deploy` is now registered with `{ hidden: true }` — works, but not shown in `--help`.

**9. Docs nav.** Kept `docs/site/cli/deploy.md`, removed the sidebar link so it isn't public yet.

Tests 22/22, build + structural lint clean.

PR #761, opened 2026-08-01
feat(serve): expose function parameters in the /list manifest

egonSchiele:
## What

The HTTP serve `/list` manifest gives **nodes** a `parameters` array but gave **functions** none — so a client (and `agency deploy`'s generated curl examples) couldn't see a function's argument names, only an empty `{}` body.

`ExportedFunction` now carries `parameters` (from the function's `AgencyFunction.params`), and the `/list` function entry lists the names, exactly like nodes already do.

## Change
- `lib/serve/types.ts` — add `parameters: Array<{ name: string }>` to `ExportedFunction`.
- `lib/serve/discovery.ts` — populate it in `toExportedFunction` from `fn.params`.
- `lib/serve/http/adapter.ts` — include `parameters` in the `/list` function entry.

## Tests
- `discovery.test.ts` — a function with params surfaces them.
- `adapter.test.ts` — `/list` reports `functions[0].parameters` (updated fixtures to include the new required field, since real `AgencyFunction`s always have `params`).
- `createServeHandler.test.ts` — fixture registry entry gains `params`.

103/103 serve tests pass; structural lint clean. (The `functionFrame.integration.test.ts` suite errors on `agency-lang/stdlib` module resolution in a local worktree — pre-existing and environmental, reproduced with these changes stashed; its 2 tests are skipped and it passes in CI.)

Unblocks templated function bodies in `agency deploy`'s curl examples.

### DIFF
--- a/packages/agency-lang/lib/serve/discovery.ts
+++ b/packages/agency-lang/lib/serve/discovery.ts
@@ -58,7 +58,11 @@ function toExportedFunction(
     kind: "function",
     name: fn.name,
     description: fn.toolDefinition!.description,
-    parameters: fn.params.map((param) => ({ name: param.name })),
+    // Only unbound params are caller-facing — bound params are filled at
+    // definition time and rejected if sent. Matches `logRoutes` in adapter.ts.
+    parameters: fn.params
+      .filter((param) => !param.isBound)
+      .map((param) => ({ name: param.name })),
     agencyFunction: fn,
     interruptEffects,
     invoke: makeInvoker(fn, moduleInvoke),
--- a/packages/agency-lang/lib/serve/discovery.ts
+++ b/packages/agency-lang/lib/serve/discovery.ts
@@ -58,6 +58,7 @@ function toExportedFunction(
     kind: "function",
     name: fn.name,
     description: fn.toolDefinition!.description,
+    parameters: fn.params.map((param) => ({ name: param.name })),
     agencyFunction: fn,
     interruptEffects,
     invoke: makeInvoker(fn, moduleInvoke),
--- a/packages/agency-lang/lib/serve/createServeHandler.test.ts
+++ b/packages/agency-lang/lib/serve/createServeHandler.test.ts
@@ -21,6 +21,7 @@ export const __toolRegistry = {
     name: "add",
     module: "agent",
     exported: true,
+    params: [{ name: "a" }, { name: "b" }],
     toolDefinition: { name: "add", description: "Add two numbers", schema: null },
   },
 };
--- a/packages/agency-lang/lib/serve/discovery.test.ts
+++ b/packages/agency-lang/lib/serve/discovery.test.ts
@@ -50,6 +50,30 @@ describe("discoverExports", () => {
     ]);
   });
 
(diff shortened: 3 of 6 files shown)


packages/agency-lang/lib/serve/discovery.ts
```diff
@@ -58,6 +58,7 @@ function toExportedFunction(
     kind: "function",
     name: fn.name,
     description: fn.toolDefinition!.description,
+    parameters: fn.params.map((param) => ({ name: param.name })),
```
egonSchiele: This lists *all* params, including bound ones. `logRoutes` in `adapter.ts` filters them — `item.agencyFunction.params.filter((p) => !p.isBound)` — because "bound params are filled in at definition time and are not part of the request body." Without the same filter here, `/list` (and `agency deploy`'s curl body template) advertises bound params as caller inputs for any partially-applied/bound function.

Suggest `fn.params.filter((p) => !p.isBound).map((p) => ({ name: p.name }))`, and a discovery test with a bound param asserting it's excluded from `parameters`.

egonSchiele (commented):
## Review — expose function parameters in /list

**Overview.** `ExportedFunction` gains a `parameters` array, populated in `toExportedFunction` from `fn.params` and surfaced in the `/list` function entry — mirroring what nodes already do, so `agency deploy`'s curl examples can template a real function body instead of `{}`. Small, focused, well-tested.

**Strengths.** Clean mirror of the node shape (`Array<{name}>` in the type, flattened to `string[]` in the manifest); required-field change is internal-construction-only and all fixtures were updated; tests cover both discovery and the `/list` output.

**Finding — bound params should be filtered out (medium).** `toExportedFunction` maps *all* of `fn.params`, but `AgencyFunction` params carry `isBound` (bound params are pre-filled at definition time via `boundValue` and are not caller inputs — the runtime tracks `_unboundParams = params.filter(p => !p.isBound)`). The sibling `logRoutes` in this same `adapter.ts` already does exactly this: `item.agencyFunction.params.filter((p) => !p.isBound)`, with the comment "Bound params are filled in at definition time and are not part of the request body, so only the unbound params are caller-facing."

So for any exported function with bound params (partial application / bound tools), the `/list` manifest will advertise params the caller must not supply — and `agency deploy`'s generated curl body (the motivation for this PR) would include them, producing wrong example requests. Fix: `fn.params.filter((p) => !p.isBound).map((p) => ({ name: p.name }))`, matching `logRoutes`, and add a discovery test with a bound param asserting it's excluded. The `"mirrors ExportedNode.parameters"` comment is right in spirit, but nodes have no bound-param concept, so a faithful mirror for functions still needs the unbound filter. See inline.

**Verdict.** Correct for the common (no-bound-params) case and otherwise clean; just add the `!p.isBound` filter so bound-param functions don't over-report their inputs, consistent with the route logger right below it.

egonSchiele:
Fixed — good catch. `toExportedFunction` now filters `!param.isBound` before mapping (matching `logRoutes` right below it in adapter.ts), so bound/partially-applied params no longer show up in `/list` or the deploy curl body. Added a discovery test with a bound param asserting it's excluded. Build + serve tests green.

Issue #755, opened 2026-08-01
eval: surface metrics.models (and totals) into summary.json for cross-run analysis


egonSchiele:
Each eval run already extracts the models used (deduped, sorted) into the per-input eval record (`metrics.models`, lib/eval/extract.ts), but `summary.json` does not carry them. Cross-run analysis ("does this agent do better with Gemini models?") needs models at the summary level so it can group runs without opening every input record.

Small change: thread `metrics.models` (and plausibly `costUsdTotal` / token totals, which are in the same struct) into each input entry in `summary.json` at write time. Groundwork for the runs-analysis project.

egonSchiele:
Shipped in #762: summary.json input entries carry a metrics block with costUsd, durationMs, startedAtMs, models, and agentName, copied from each eval record at write time.

PR #762, opened 2026-08-01
Cross-run explorer: agency logs over run directories, setAgentName, summary metrics

egonSchiele:
## The cross-run explorer

`agency logs` now points at anything log-shaped. A sole statelog file keeps today's viewer exactly; a run directory, a directory of run directories, or several mixed paths open a new explorer: one sortable, groupable table with a row per run, drill-down into a run's tests, and from a test into the existing log viewer on the same terminal. `--csv` prints the table to stdout instead.

Spec and plan were reviewed beforehand; this implements the reviewed plan. Architecture notes live in `docs/dev/runs-explorer.md`; user-facing docs in `docs/site/cli/logs.md`.

### Eval artifact widening (first two commits)

- `EvalRecord` gains `startedAtMs` (absolute first-event time; `durationMs` was already wall-clock per test) and `agentName` (last `agentName` event). Additive; recordVersion stays 2.
- `summary.json` per-input entries gain a `metrics` block ({costUsd, durationMs, startedAtMs, models, agentName}) copied from each record at write time, with an `onWarning` parameter instead of silent failures. **This supersedes #755** (models are included) — that issue can be closed against this PR.
- Why: the explorer's phase 1 reads exactly two files per run (`summary.json` + `config.json`, pinned by a read-counting test), so a directory of hundreds of runs shows a full table in seconds. A run's `time` column is the wall-clock envelope `max(start+duration) − min(start)`, not the sum of test durations (a parallel run must not read 33m when it took 20m).

### `setAgentName` (std::statelog)

One call names the agent; the event flows statelog → record → summary → the explorer's identity resolution, so runs group under a stable identity regardless of launch command. Mirrors `evalValue`'s client/frame mechanism (lenient no-op outside an Agency frame). Covered by unit tests and an agency-js test asserting the emitted event; last call in a trace wins.

### Shared column widths, declarative tables

The width-distribution algorithm moved verbatim from `lib/stdlib/layout/table.ts` into dependency-free `lib/utils/columnWidths.ts` (moved down, not reached up for — `lib/tui` never imports stdlib internals). `std::ui/table` behavior is unchanged: all 229 layout tests green with zero snapshot churn, error text byte-identical. On top of it, `lib/tui/table.ts` adds a declarative `TableComponent` (column specs in, aligned/clipped/cursor-highlighted rows out; width arrays never cross the boundary) that renders every explorer table screen.

### The explorer itself (`lib/runsExplorer/`)

- **Loader owns loading**: declarative `progress`/`upsert`/`done` events, one bounded unit per `advance()` (one summary read, one record read, one statelog chunk). Backfill reads the eval record first and streams the statelog only for fields the record could not supply; killed runs mine cost/models from events so a $6 kill never reads $0.00. Corrupt summaries (usually a run still being written) become visible failed rows with warnings, never crashes. The resumable scanner survives UTF-8 characters split across chunk boundaries (pinned by test).
- **Cursor pins to row identity**, never an index — backfill can rename an agent and regroup the table mid-scroll without losing the highlighted row (pinned by test). Missing values sort last in both directions.
- **Views**: runs table (sort `s`/`S`, group `b` with expandable headers, narrow terminals drop models→time→pass), per-test drill-in that re-resolves its parent on every upsert, compare (agent × suite mean±count matrix, graded runs only, objective 0 counts), trend (per-agent day/week-bucketed sparkrows that never overflow), info screen, loading splash.
- **Shell rules**: at most one outstanding key request and never one while the embedded viewer runs (asserted + tested — the viewer receives its own first keypress); the loader advances once per macrotask so `q` lands between chunks of a multi-megabyte scan. Esc backs out until it can't and never quits; `q` quits instantly anywhere, including mid-load.
- **CSV**: interactive `e` exports the current projection with group members always listed under their headers (expanded or not); `--csv` drains the same loader synchronously so interactive and headless numbers cannot drift.

### CLI routing

`agency logs` takes variadic paths with sniffed classification; the sole-statelog path (including `-`/stdin and `--follow`, even on an empty file) is byte-identical to before and its integration test is unchanged. `agency eval logs <runDir>` without `--input` opens the per-test table; with `--input` it keeps the straight-to-viewer path. Routing is tested at source level via injected seams — no dynamic module mocks. One packaged integration case covers `agency logs <runsDir> --csv`.

### Also in this branch

The viewer/infra changes the prototype validated ride along from `adit/runs-aggregate`: bottom-right view tags in every viewer screen, `Shift+T` backward cycling, `runViewer`'s `embedded` flag + `ViewerResolution`, and the dev-doc notes on lib/tui layout rules. The prototype files (`explorerProto.ts`, `runsProto.ts`, the `runs-proto` command) are deleted (grep-gated).

### Verification

- 453 focused tests across every new/modified module (loader read-spy, cursor-identity, UTF-8 chunk boundary, waiter ownership, `q`-during-scan, routing seams), all green; full `tsc` clean; `pnpm run lint:structure` clean; `make` green; stdlib typecheck + `set-agent-name` agency-js test green on the final build.
- Diff audited against `docs/dev/anti-patterns.md` and `docs/dev/coding-standards.md` (no nested ternaries, no conditional spreads, no silent catches, no single-character identifiers, named constants for chunk sizes/thresholds/glyphs).
- CI runs the wide suites.


egonSchiele (commented):
Verified at `ef6e8b1` in a clean worktree: `npx vitest run lib/runsExplorer lib/tui/table.test.ts lib/utils/columnWidths.test.ts lib/stdlib/layout` gives **349 passed / 20 files**, including the 207 pre-existing layout tests — the zero-churn claim on `std::ui/table` holds.

I reviewed the spec beforehand and raised three blockers. All three are fixed, and two are fixed better than I suggested:

- **`readEvalRun` printing into the TUI** → `onWarning` parameter, warnings flow to `warnings[]`.
- **Phase 1 reading every eval record** → summary `metrics` block, two content reads per run, pinned by a read-counting test rather than a claim.
- **`lib/tui` reaching up into `lib/stdlib`** → moved *down* into `lib/utils/columnWidths.ts`, and the refinement of operating on pre-measured numbers ("each caller measures cell content with its own width function") is a cleaner boundary than the one I proposed. I diffed the move: the only changes to `lib/stdlib/layout/table.ts` are the deletions plus the call site passing `"std::ui/layout"` as `errorContext`, `PlannedWidth` is structurally identical to `Width`, and the thrown error string is preserved character for character.

The test names also show the hard cases from the spec review landed as real tests rather than notes — "an agent rename arriving from backfill regroups without losing the cursor", "q interrupts a large backfill scan mid-load", "a multi-byte UTF-8 character split across chunk boundaries survives", "one advance reads at most one chunk", "hands the embedded viewer the terminal".

Two things I checked because they looked like bugs and are not: `mine.result()` returns `failed` when `traces` is empty, so `jobPatch`'s `Object.values(result.traces)[0]` can never be `undefined`; and handing the same `OutputTarget` to two `Screen` instances is safe because `TerminalOutput` owns the single `previousGrid` that both write through, so the diff-based repaint stays coherent across the viewer hand-off. Both are load-bearing and neither is obvious — worth a comment at some point.

Five things below. The multi-trace one is the only one I would fix before merging; the rest are small or are questions.

packages/agency-lang/lib/runsExplorer/loader.ts
```diff
@@ -0,0 +1,271 @@
...
+  if (job.scan !== null) {
+    const result = job.scan.result();
+    if (result.kind === "failed") {
+      if (job.record === null) {
+        patch.warnings.push(result.warning);
+      }
+      return patch;
+    }
+    patch.warnings.push(...result.warnings);
+    const totals = Object.values(result.traces)[0];
```
egonSchiele: `Object.values(result.traces)[0]` keeps the first trace and silently drops the rest, so a per-input statelog holding more than one trace under-reports cost and models — the same "$6 kill reads $0.00" failure this backfill exists to prevent, just partial instead of total.

The asymmetry is visible in the tests: `loader.test.ts` has "a multi-trace statelog emits one trace row per recovered trace", which covers `buildTraceRows` — the *direct* path. Nothing covers multi-trace through backfill, which is this line.

Insertion order is also not a meaningful choice: it is whichever `trace_id` the accumulator saw first, which for an interleaved log is not the main trace.

For a per-input row, summing looks right — cost added, models unioned, `min(firstTs)` / `max(lastTs)` for the span — since the row means "this input", not "this trace". If a single trace really is the intent, pick by `eventCount` (already accumulated in `TraceTotals`) and push a warning when more than one trace was found, so it is visible on the info screen rather than silent.

Whether this is reachable depends on whether a `--agent-cmd` agent can invoke agency more than once into the same statelog. Given `AGENCY_TRACE_ID` handoff exists, I would assume yes.

packages/agency-lang/lib/runsExplorer/run.ts
```diff
@@ -0,0 +1,252 @@
...
+    }
+    if (action.kind === "openLog") {
+      return this.openLog(action.statelogPath);
+    }
+    if (action.kind === "exportCsv") {
+      const { path: outPath, content } = exportCsv(
+        csvRowsFromProjection(action.projection, this.rows),
+        new Date(),
+      );
+      fs.writeFileSync(outPath, content);
```
egonSchiele: Unguarded `fs.writeFileSync`. A read-only working directory, a full disk, or a permissions problem throws out of `dispatchAction`, out of `run()`, and past the `finally` — the app dies with a stack trace painted over the alt screen.

Everything else in this shell is defensive about the outside world: `openLog` catches and notifies, the loader turns a corrupt summary into a visible failed row. This is the one place that does not, and it is user-triggered (`e`) rather than incidental.

```ts
try {
  fs.writeFileSync(outPath, content);
  this.active().notify(`exported → ${outPath}`);
} catch (error) {
  this.active().notify(`export failed: ${errText(error)}`);
}
```

packages/agency-lang/lib/runsExplorer/run.ts
```diff
@@ -0,0 +1,252 @@
...
+  private async nextStep(): Promise<{ kind: "key"; event: KeyEvent } | { kind: "tick" }> {
+    if (this.pendingKey === null) {
+      this.pendingKey = this.screen.nextKey();
+    }
+    if (this.loader.isDone()) {
+      const event = await this.pendingKey;
+      this.pendingKey = null;
+      return { kind: "key", event };
+    }
+    const winner = await Promise.race([
```
egonSchiele, line 119: `this.pendingKey.then(...)` builds a new derived promise, and attaches a new handler to the same underlying `nextKey()` promise, on every tick. During a long backfill the tick branch wins thousands of times in a row while `pendingKey` stays outstanding, so those handlers accumulate on one promise until a key finally arrives.

Not a correctness bug — the ownership invariant this class is built around still holds — but it is unbounded growth on exactly the path the design optimizes for (a multi-megabyte scan).

The codebase already solved this. `Screen.runLoop`'s doc comment (`lib/tui/screen.ts:56-60`) says it "reuses a single in-flight `nextKey()` promise across iterations so a key the user types during a tick race is consumed by the next race rather than queued behind abandoned waiters." Same fix here: cache the derived promise next to `pendingKey` and clear both together when a key wins.

packages/agency-lang/lib/runsExplorer/views/tableState.ts
```diff
@@ -0,0 +1,236 @@
...
+function aggregate(members: RunRow[], group: "agent" | "suite", label: string): GroupAggregates {
+  const dates = members.map((member) => member.startedAtMs).filter(isKnown);
+  const scores = members.map((member) => member.score).filter(isKnown);
+  const costs = members.map((member) => member.costUsd).filter(isKnown);
+  const times = members.map((member) => member.wallMs).filter(isKnown);
+  return {
+    date: dates.length === 0 ? null : Math.max(...dates),
+    score: scores.length === 0 ? null : scores.reduce((s, v) => s + v, 0) / scores.length,
+    cost: costs.length === 0 ? null : costs.reduce((s, v) => s + v, 0),
+    time: times.length === 0 ? null : times.reduce((s, v) => s + v, 0),
```
egonSchiele, line 200: A group's `time` is the sum of its members' wall-clock, which reintroduces at group level the reading the run-level rule exists to prevent. The PR description is emphatic that a run's time is the envelope `max(start+duration) − min(start)` and not the sum, because "a parallel run must not read 33m when it took 20m" — and then a group header over runs that overlapped in wall-clock shows their sum.

It may well be the right answer at this level (three runs on three different days genuinely add up to their total), but the two rules disagree and nothing says why. Either make it an envelope for consistency, or keep the sum and say in one line that a group's time means "total machine time spent", not "elapsed time" — the distinction the run-level rule is careful about.

Same question applies to `cost` on the line above, though there summing is unambiguous.

packages/agency-lang/lib/runsExplorer/mine.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+  let parsed: Record<string, unknown>;
+  try {
+    parsed = JSON.parse(fs.readFileSync(recordPath, "utf-8")) as Record<string, unknown>;
+  } catch (error) {
+    return { kind: "warning", message: `could not read ${recordPath}: ${errText(error)}` };
+  }
+  const metrics = parsed.metrics as Record<string, unknown> | undefined;
+  const value: RecordMetrics = {
+    costUsd: numberOr(metrics?.costUsdTotal, 0),
```
egonSchiele: `numberOr(metrics?.costUsdTotal, 0)` turns "the record has no metrics block" into a confident `$0.00`, and because `jobPatch` only consults the statelog `if (patch.costUsd === undefined)` (`loader.ts:230`), the fallback that would have recovered the real number never runs. That is the "$6 kill must not read $0.00" failure arriving through a different door: not a missing record, but a record missing its metrics.

`metrics` is required on `EvalRecord` today, so this may be unreachable in practice — worth confirming that no `recordVersion: 2` record predates the block. If it is unreachable, it is cheap to make it structurally unreachable rather than incidentally so: return `undefined` when the block is absent and let the statelog fill it.

Note the line two below already draws exactly this distinction for `startedAtMs` — `null` when absent, so the statelog supplies it — with a comment explaining why. Cost deserves the same treatment, and for a stronger reason: a wrong `$0.00` reads as a fact, whereas a missing duration reads as missing.

packages/agency-lang/lib/runsExplorer/loader.ts
egonSchiele: Fixed in cb493be84 — backfill now folds every trace: cost summed, models unioned, timestamp envelope for start/duration, and the agent name comes from the busiest trace that has one (eventCount was already accumulated). New loader test appends a second, busier trace to a killed input's statelog and asserts the summed cost, the model union, and that the busy trace's name wins.

packages/agency-lang/lib/runsExplorer/mine.ts
egonSchiele: Fixed in cb493be84 — readRecordMetrics reports null for cost/duration/models when the record does not know (no metrics block), and jobPatch merges field-by-field on null, so the statelog fallback runs for exactly the missing fields. The record-complete check in the loader now also requires a known cost before skipping the scan. Pinned by a new test: a record with only durationMs reports null cost, never zero.

packages/agency-lang/lib/runsExplorer/run.ts
egonSchiele: Fixed in cb493be84 — the write is wrapped and failures land in the status line as 'export failed: …'.

packages/agency-lang/lib/runsExplorer/run.ts
egonSchiele: Fixed in cb493be84 — the derived race arm is cached alongside pendingKey and both clear together when a key wins, with a comment pointing at Screen.runLoop as the precedent.

packages/agency-lang/lib/runsExplorer/views/tableState.ts
egonSchiele: Kept the sum, documented the semantics in cb493be84: a group's time is total machine time spent, deliberately not the elapsed envelope — group members usually span days, so an envelope would mostly measure the gaps between runs. The comment sits on GroupAggregates.time next to the cost sum.

packages/agency-lang/lib/eval/runArtifacts.ts
```diff
@@ -199,19 +201,63 @@ export function recordInputPrepareFailure(
...
+    };
+    if (record.agentName !== undefined) {
+      value.agentName = record.agentName;
+    }
+    return { kind: "metrics", value };
+  } catch (error) {
+    const text = error instanceof Error ? error.message : String(error);
+    return { kind: "warning", message: `summary metrics: could not read ${recordPath}: ${text}` };
+  }
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in cb493be84 — readInputMetrics validates the shape before copying (finite numbers for cost/duration/start, string array for models): a parseable-but-corrupt record now warns and leaves metrics absent, so the explorer backfills from the statelog instead of trusting NaN. Covered by a new runArtifacts test.

egonSchiele:
All five findings plus Copilot's addressed in cb493be84:

- **Multi-trace backfill (the blocker)**: every trace of an input statelog now folds into the patch — cost summed, models unioned, timestamp envelope, agent name from the busiest trace. New loader test drives an interleaved two-trace input through backfill (the path the direct-row test did not cover).
- **Record cost null-not-zero**: a record with no metrics block reports null cost/models and the statelog fills them; the loader's "record is complete" check also requires a known cost now.
- **CSV write** wrapped; failures notify instead of dying over the alt screen.
- **Key-promise handler pile-up**: the derived race arm is cached with `pendingKey` (Screen.runLoop's fix, credited in the comment).
- **Group time**: kept the sum, documented as "total machine time spent, deliberately not the envelope" on `GroupAggregates.time`.
- **readInputMetrics** validates shape before writing metrics into summary.json (Copilot's finding).

Also added the two comments you asked for on the non-obvious load-bearing facts: `mine.result()` returning `failed` for an empty scan (what makes `jobPatch`'s trace access safe), and the shared-`TerminalOutput`/`previousGrid` reason the viewer hand-off repaints coherently.

125 affected tests green, tsc clean.

packages/agency-lang/lib/runsExplorer/run.ts
```diff
@@ -0,0 +1,270 @@
...
+      this.pendingKey = null;
+      this.pendingKeyStep = null;
+      return { kind: "key", event };
+    }
+    const winner = await Promise.race([
+      this.pendingKeyStep as Promise<{ kind: "key"; event: KeyEvent }>,
+      nextMacrotask().then(() => ({ kind: "tick" as const })),
+    ]);
+    if (winner.kind === "key") {
+      this.pendingKey = null;
```
egonSchiele, line 124: The fix is right and matches `Screen.runLoop`. One note on how the invariant is held.

`pendingKey` and `pendingKeyStep` must be set and cleared together, and the only thing making this line typecheck is the `as` cast asserting that. If they ever desync, `Promise.race([null, tick])` treats the non-promise arm as already resolved, `winner` comes back `null`, and `winner.kind` throws — inside the shell's main loop, during a long scan, which is the least pleasant place for it.

The type system can carry the invariant instead of a cast:

```ts
private pending: {
  key: Promise<KeyEvent>;
  step: Promise<{ kind: "key"; event: KeyEvent }>;
} | null = null;
```

One assignment, one clear, no cast, and the desync becomes unrepresentable.

Worth doing mostly for consistency with the rest of this class: the file opens by naming two hard rules, and the other one is enforced with a runtime assertion at the hand-off. This one is currently enforced by remembering to touch both lines.

egonSchiele (commented):
Re-reviewed `ef6e8b1` → `cb493be8`. All five findings are resolved, and two go further than what I asked for.

Verified: `npx vitest run lib/runsExplorer lib/eval/runArtifacts.test.ts lib/eval/readRun.test.ts lib/eval/extract.test.ts lib/tui/table.test.ts lib/utils/columnWidths.test.ts lib/stdlib/layout` → **414 passed / 23 files**, including the 207 layout tests still clean.

**Multi-trace backfill.** `sumTraces` is the right fold, and the cost/duration pairing is the detail that makes it correct rather than merely different: cost **sums** across traces, the timestamp span is an **envelope**. That is the same pairing as the run-level rule, so two agency invocations — sequential or concurrent — both come out right. Picking the agent name from the busiest trace rather than the first is the part I would have settled for a warning on; the test pins all three (sum, model union, busiest-trace name) against a genuinely multi-trace fixture.

**Null-not-zero.** `RecordMetrics` going `number | null` across the board, `jobPatch` merging field-by-field, and `complete` gaining the `costUsd !== null` term are all needed together — without that last one an old record carrying `startedAtMs` and `agentName` would still have short-circuited before the scan. Easy to miss; it is there.

I checked whether requiring `agentName` in `complete` forces a statelog scan on every input: it does not for modern runs, because `backfillInputIds` only collects inputs with no summary `metrics` or with `status === "error"` (`rows.ts:86`). For errored inputs the extra scan is the stated intent. No performance regression.

**`runArtifacts` shape validation** was not something I raised, and it is the better half of the same idea — refusing to write a bad number into `summary.json` at all, rather than only declining to trust it on read. Worth noting the deliberate consequence: validation is all-or-nothing, so one bad field drops the whole metrics block and that input falls to backfill. Since `writeEvalRunSummary` runs once at the end of `runSuite` against records the same version just wrote (`runSuite.ts:208`, and `gradeSuite` writes nothing), the only way to trip it is real corruption — where "leave it absent and let the explorer recover it" is exactly right.

The two comments on the non-obvious load-bearing facts (empty scan ⇒ `failed`; the shared `previousGrid`) are in the right places — both are things a future reader would otherwise have to re-derive before daring to touch them.

One small note below on the paired promise fields. Nothing blocking.

PR #763, opened 2026-08-02
feat(compiler): compileSource sourcePath — resolve relative imports

egonSchiele:
## What

`compileSource(source, config)` takes source as a **string**, so it can't know where that source lives on disk. To resolve an import like `import { helper } from "./helpers.agency"`, the compiler walks the filesystem — so `compileSource` wrote the source to an isolated temp dir and compiled there. The sibling `helpers.agency` isn't in that temp dir, so the import fails ("Symbol 'helper' is not defined in './helpers.agency'"). This is the root cause of statelog#9 — hosted multi-file agents can't be compiled or served.

This adds an optional `sourcePath` to `CompileSourceOptions`. When given (the file is already on disk beside its siblings), `compileSource` compiles **at that real path**, so relative `.agency` imports resolve. Without it, behavior is unchanged: the source goes to a temp file (single-file compile), exactly as before.

## Why this shape

A host that serves multi-file agents (statelog) already stores every uploaded file on disk. So it always has real paths, at upload and serve time — passing `sourcePath` is all it needs. No new public function; one optional field on the existing one.

## Test

`compile.test.ts` — a two-file agent (`main.agency` imports `./helpers.agency`): with `sourcePath` the import resolves (compiled output imports `./helpers.js`); without it, compilation fails (today's single-file behavior). 20/20 compile tests pass; build + structural lint clean.

Design was prototyped first on `proto/multifile-compile`. Follow-ups (separate PRs): statelog switches upload + serveHost to path-based compile; then `agency deploy`'s single-file refusal is relaxed.

### DIFF
--- a/packages/agency-lang/lib/compiler/compile.ts
+++ b/packages/agency-lang/lib/compiler/compile.ts
@@ -58,12 +58,16 @@ export type CompileSourceOptions = AgencyConfig & {
    */
   imports?: ImportPolicy;
   /**
-   * Real on-disk path of `source`, when it already exists on disk. Given it,
-   * `compileSource` compiles at that path — where sibling `.agency` files live —
-   * so relative imports resolve. Without it, `source` is written to an isolated
-   * temp file, so relative imports cannot resolve (single-file compile). A host
-   * that stores multi-file agents on disk passes this to compile each file
-   * against its real neighbors.
+   * Real on-disk path of the file to compile. Given it, `compileSource` compiles
+   * at that path — where sibling `.agency` files live — so relative imports
+   * resolve. Without it, `source` is written to an isolated temp file, so
+   * relative imports cannot resolve (single-file compile). A host that stores
+   * multi-file agents on disk passes this to compile each file against its real
+   * neighbors.
+   *
+   * When set, the file at this path is authoritative: it is parsed AND used for
+   * symbol/import resolution, so the two can't diverge. The `source` argument is
+   * ignored (pass the file's contents for clarity, but the disk file wins).
    */
   sourcePath?: string;
 };
@@ -124,9 +128,18 @@ export function compileSource(
     fs.writeFileSync(syntheticPath, source, "utf-8");
   }
 
+  // The file at `syntheticPath` is the single source of truth: SymbolTable.build
+  // and buildCompiledClosure read the entrypoint and its imports from there, so
+  // parse the SAME bytes rather than the passed `source`. In the temp-file case
+  // that file IS `source` (just written); with a caller `sourcePath` this reads
+  // the on-disk file, so the AST and symbol/import resolution can't diverge.
+  const compiledSource = config.sourcePath
+    ? fs.readFileSync(syntheticPath, "utf-8")
+    : source;
+
   try {
     // 1. Parse
-    const parseResult = parseAgency(source, config, true);
+    const parseResult = parseAgency(compiledSource, config, true);
     if (!parseResult.success) {
       return {
         success: false,
@@ -183,7 +196,7 @@ export function compileSource(
       liftedProgram,
       symbolTable,
       syntheticPath,
-      source,
+      compiledSource,
     );
 
     // 5. Type check
--- a/packages/agency-lang/lib/compiler/compile.ts
+++ b/packages/agency-lang/lib/compiler/compile.ts
@@ -57,6 +57,15 @@ export type CompileSourceOptions = AgencyConfig & {
    * See `lib/importPaths.ts` for the ImportPolicy shape.
    */
   imports?: ImportPolicy;
+  /**
+   * Real on-disk path of `source`, when it already exists on disk. Given it,
+   * `compileSource` compiles at that path — where sibling `.agency` files live —
+   * so relative imports resolve. Without it, `source` is written to an isolated
+   * temp file, so relative imports cannot resolve (single-file compile). A host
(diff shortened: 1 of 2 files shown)


packages/agency-lang/lib/compiler/compile.ts
```diff
@@ -100,11 +109,20 @@ export function compileSource(
...
-  const syntheticPath = path.join(tempDir, `${moduleId}.agency`);
-  fs.writeFileSync(syntheticPath, source, "utf-8");
+  // SymbolTable.build() walks the file system from the source's path to resolve
+  // imports. With a caller-supplied sourcePath (the file is already on disk
+  // beside its siblings), compile at that path so relative `.agency` imports
+  // resolve. Otherwise write to an isolated temp file — relative imports cannot
+  // resolve there, by design (single-file compile).
+  let tempDir: string | undefined;
+  let syntheticPath: string;
+  if (config.sourcePath) {
```
egonSchiele, line 123: When `sourcePath` is set, the AST comes from the `source` string (`parseAgency(source, …)`) but symbols/imports come from disk (`SymbolTable.build(syntheticPath, …)` reads the entrypoint and its imports at `sourcePath`). The temp-file branch can't diverge because it writes `source` to the temp file first; this branch trusts that `source` already equals the file at `sourcePath`, with nothing enforcing it. A mismatched pair would parse one thing and resolve another — a silent miscompile.

Consider reading the entrypoint from disk here (making `source` redundant when `sourcePath` is given), or at minimum documenting the invariant on the field: when `sourcePath` is set, `source` must equal the file at that path.

egonSchiele (commented):
## Review — `compileSource` `sourcePath` (agency-lang side of the multi-file pair)

**Overview.** Adds an optional `sourcePath` to `CompileSourceOptions`. When given, `compileSource` uses that real on-disk path as the compile root (instead of writing `source` to an isolated temp file), so relative `.agency` imports resolve against the sibling files. Without it, behavior is unchanged. This is the root-cause fix for statelog#9; the consumer is statelog#10.

**Strengths.** Minimal, backward-compatible surface (one optional field, no new function). The `finally` cleanup guard was correctly updated to `tempDir && tempDir.startsWith(os.tmpdir())`, so a caller-owned `sourcePath` is never deleted. Compiled output still goes to a tmp `outputPath` and is returned as a string, and the relative-import rewrite is computed against `syntheticPath`, so a caller placing the `.js` beside its siblings gets a working `import "./helper.js"` (the test confirms this). Test covers both branches.

**Findings.**

1. **Two sources of truth when `sourcePath` is set: `source` (parsed) vs the file on disk (read by `SymbolTable.build`) (medium — API sharp edge).** `parseAgency` builds the AST from the `source` string, but `SymbolTable.build(syntheticPath, …)` reads the entrypoint and its imports from disk at `sourcePath`. In the temp-file path these can't diverge because `source` is written to the temp file first; with a caller-supplied `sourcePath`, nothing enforces `source === readFile(sourcePath)`. If they differ, the AST and the symbol/import resolution disagree — a confusing, silent miscompile. statelog#10 happens to satisfy the invariant (it writes then compiles, and serveHost reads the same path it passes), but the signature invites divergence. Suggest either reading the entrypoint from disk when `sourcePath` is given (single source of truth — and `source` becomes redundant), or documenting the invariant explicitly on the field: *when `sourcePath` is set, `source` must equal the file at that path.* See inline.

2. **Confirm no pipeline stage writes sidecars into the caller's directory (low, verify).** With `sourcePath`, `SymbolTable.build`, `expandSplices`, and `buildCompiledClosure` all run against the real path rather than a throwaway temp dir. The visible writes are safe (temp source skipped; `outputPath` is in tmpdir; code returned as a string). Worth a quick confirmation that none of those stages emit a cache/`.d.ts`/lock file next to the source, since a multi-file host will be compiling inside a directory it doesn't want mutated.

**Security note (belongs mostly to the consumer).** Compiling at the real path removes the temp-dir isolation, so a `.agency` file whose import escapes the directory (`import "../../elsewhere.agency"`) will now be read from the real filesystem at compile time. `compileSource` faithfully compiling what's on disk is reasonable; containment is the host's job — flagged on statelog#10.

**Verdict.** Clean, well-scoped, correct. #1 is the one to address before other hosts adopt the option — make the `source`/`sourcePath` relationship explicit so a mismatched pair can't silently miscompile.

egonSchiele:
Addressed in the latest commit.

**1. source/sourcePath divergence (medium).** Fixed by making the on-disk file authoritative: when `sourcePath` is set, `compileSource` now reads the file and **parses those same bytes** (rather than the passed `source` string), so the AST and the symbol/import/closure resolution — which already read from disk — can't diverge. The `source` arg is ignored in that branch; documented that on the field. Added a test that passes a *different* `source` string than the file at `sourcePath` and asserts the disk file wins.

**2. No sidecar writes into the caller's dir (verify).** Confirmed: `SymbolTable.build`, `expandSplices`, and the closure builder have no `writeFileSync`/`mkdirSync`/`appendFileSync`. The only writes are the (now-skipped) temp source and the compiled `outputPath` in `os.tmpdir()`; the code is returned as a string. So compiling at a caller's `sourcePath` reads from that directory but never mutates it.

**Security note (import escaping the dir).** Agreed it's the consumer's job — tracking + addressing on statelog#10 (containment to the project upload dir).

21/21 compile tests pass; build + lint clean.

PR #764, opened 2026-08-02
feat(deploy): support multi-file agents (flat same-dir siblings)

egonSchiele:
## What

Relaxes `agency deploy`'s single-file restriction now that hosted serve can compile multi-file agents (statelog#10, on `agency-lang@0.12.0`'s `compileSource(sourcePath)`).

`agency deploy` now uploads the **entrypoint plus every local `.agency` file it imports** (transitively):
- `collectAgencyBundle` walks the local import graph and gathers the file set.
- `validateBundleCompiles` compiles each file **at its real path** (`sourcePath`), so the local pre-flight resolves relative imports the same way statelog does.
- `BundleFile` gains `absPath` (used as that `sourcePath`).

## Still refused (statelog can't host these)

- **Imports resolving outside the entrypoint's directory** (`./sub/x.agency`, `../x.agency`) — statelog stores an agent's files flat and its filenames can't contain slashes, so everything must sit in one directory.
- **Local TypeScript/JavaScript interop** — statelog only compiles `.agency` source.

## Tests

`bundle.test.ts` — bundles a sibling `.agency` import; refuses an outside-dir import; `validateBundleCompiles` passes a multi-file agent (imports resolve via `sourcePath`). 24/24 deploy tests pass; build + structural lint clean. Smoke-tested end-to-end: a two-file agent bundles both; a nested import is refused.

Docs updated (`docs/site/cli/deploy.md`). This completes the multi-file arc: agency-lang#763 → statelog#10 → this.

### DIFF
--- a/packages/agency-lang/lib/cli/deploy/bundle.ts
+++ b/packages/agency-lang/lib/cli/deploy/bundle.ts
@@ -1,9 +1,8 @@
-// Turning an entrypoint into the `.agency` source to upload, and checking it
-// compiles. Deploy is single-file for now: statelog compiles each uploaded file
-// in isolation (agencyCompiler/serveHost both use `compileSource`, which loses
-// the file's location and so can't resolve relative imports), so an agent that
-// imports another local `.agency` file cannot be hosted yet. This module refuses
-// what statelog can't serve. Multi-file is tracked in statelog#9.
+// Turning an entrypoint into the set of `.agency` source files to upload, and
+// checking they compile. Statelog stores uploaded files flat (one directory,
+// keyed by basename) and its filenames can't contain slashes, so a hosted
+// agent's local imports must all be same-directory `.agency` siblings — this
+// module gathers exactly that set and refuses anything statelog can't host.
 
 import fs from "fs";
 import path from "path";
@@ -13,12 +12,16 @@ import { compileSource } from "@/compiler/compile.js";
 import {
   agencyImportTargets,
   nonAgencyLocalImportTargets,
+  resolveAgencyImportPath,
 } from "@/importPaths.js";
 
 export type BundleFile = {
   /** Upload key — the basename, which is all statelog's flat store keeps. */
   name: string;
   contents: string;
+  /** Real on-disk path — passed as `sourcePath` so the pre-flight compile
+   *  resolves this file's relative imports against its siblings. */
+  absPath: string;
 };
 
 export type AgencyBundle = {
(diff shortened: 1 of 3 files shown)


egonSchiele (commented):
## Review — multi-file `agency deploy` (completes the arc)

**Overview.** Relaxes the single-file refusal now that hosted serve compiles multi-file agents. `collectAgencyBundle` walks the transitive local `.agency` import graph and gathers the file set; `validateBundleCompiles` compiles each file at its real path (`sourcePath`) so relative imports resolve in the pre-flight the way statelog does; imports resolving outside the entrypoint's directory (and TS/JS interop) are refused.

**This is a clean PR — I verified the load-bearing pieces:**
- It reuses the compiler's own `resolveAgencyImportPath`, so the deploy walk resolves imports exactly as the compiler will — the bundled set can't drift from what compiles. Good altitude.
- `resolveAgencyImportPath` returns `path.resolve(...)` (normalized absolute) for local imports, and `rootDir` is `path.dirname(path.resolve(entrypoint))`, so the `path.dirname(importAbs) !== rootDir` string check is robust — including `./sub/../x.agency`-style paths that normalize back into the dir. ✅
- The walk is cycle-safe (`collected[fileAbs]` dedup), transitive interop is caught (every queued file is interop-checked), and refusing subpath/parent imports keeps basenames unique in statelog's flat store. ✅
- It satisfies the invariant I flagged on #763: `file.contents` is read from `file.absPath` in the same pass, so `compileSource(contents, { sourcePath: absPath })` gets matching string/disk inputs. `absPath` is also local-only (uploadClient sends just `{name, contents}`), so no local paths leak to the server. ✅

**Findings (both low).**

1. **A missing *entrypoint* reports "Imported file not found" (low, wording).** The entrypoint is the first item through the same loop, so if it doesn't exist the error reads `Imported file not found: <path>` — but the entrypoint wasn't imported. The old code had a distinct "File not found." Minor, but worth special-casing the first item (or rewording) so a typo'd entrypoint path gives a clear message. See inline.

2. **The same-dir check compares logical paths, not real (symlink-resolved) ones (low, informational).** A symlink in the entrypoint's directory pointing outside would pass `dirname === rootDir` yet read outside it. For this local CLI — the user bundling their own files to their own project — that's not a privilege boundary, so it's fine as-is; noting it only because the *server* side (statelog#10) has the analogous surface with a real trust boundary, where I raised import containment separately. See inline.

**Verdict.** Correct, well-tested (sibling bundle, outside-dir refusal, multi-file compile), and the reuse of the compiler's resolver is exactly right so deploy and compile agree. Only the entrypoint-not-found wording is worth touching; #2 is just a note. Ship it.

packages/agency-lang/lib/cli/deploy/bundle.ts
```diff
@@ -32,65 +35,88 @@ export type CollectBundleResult =
...
-        `local TypeScript/JavaScript interop can't be deployed.`,
-    };
-  }
+  while (queue.length > 0) {
+    const fileAbs = queue.pop()!;
+    if (collected[fileAbs]) {
+      continue;
+    }
+    if (!fs.existsSync(fileAbs)) {
+      return { ok: false, error: `Imported file not found: ${fileAbs}` };
```
egonSchiele, line 63: Since the entrypoint is the first item through this same loop, a missing *entrypoint* path surfaces here as `Imported file not found: <path>` — but the entrypoint wasn't imported, it was the argument. A user who typos the deploy target gets a slightly confusing message (the old code said "File not found"). Consider special-casing the first iteration (or checking the entrypoint's existence before the loop) so it reads as an entrypoint error.

packages/agency-lang/lib/cli/deploy/bundle.ts
```diff
@@ -32,65 +35,88 @@ export type CollectBundleResult =
...
-    return {
-      ok: false,
-      error:
-        `${name} imports ${localAgencyImports.map((target) => `"${target}"`).join(", ")}. ` +
-        `Deploy is single-file for now — hosted multi-file agents aren't supported yet ` +
-        `(statelog compiles each file in isolation). Tracking: statelog#9.`,
-    };
+    for (const importPath of agencyImportTargets(parsed.result, { localOnly: true })) {
+      const importAbs = resolveAgencyImportPath(importPath, fileAbs);
+      if (path.dirname(importAbs) !== rootDir) {
```
egonSchiele, line 87: Nice — reusing `resolveAgencyImportPath` (the compiler's own resolver) plus `path.resolve` normalization makes this check robust, and it correctly accepts `./a/../b.agency` that normalizes back into the dir while rejecting real subpaths/parents.

One informational note: this compares the logical path, not the symlink-resolved real path, so a symlink in the entrypoint's dir pointing elsewhere would pass. For a local CLI bundling the user's own files that's not a boundary worth enforcing — flagging only because the server side (statelog#10) has the analogous surface where it *is* a boundary.

egonSchiele:
Thanks! Fixed the entrypoint-not-found wording — `collectAgencyBundle` now checks the entrypoint exists *before* the import walk, so a typo'd deploy target reads `Entrypoint not found: <path>` while the in-loop check only fires for genuinely imported files. Test updated. 

On the symlink note (#2): agreed, no change for the local CLI — it's the user's own files, not a boundary; the analogous server-side surface is tracked as containment in statelog#11. Build + tests green.

PR #767, opened 2026-08-02
fmt: keep end-of-line comments attached to complete constructs

egonSchiele:
Tasks 1 and 2 of the trailing-comments plan. This is the first of three PRs; it covers **complete constructs** only. Multiline list items (arrays, call arguments, import lists, and so on) are unchanged and come in the follow-ups.

## The problem

Write a comment at the end of a line and `agency fmt` moves it onto its own line, above the next statement. That silently changes what the comment appears to describe:

```
const x = 5 // explains x        →      const x = 5
const y = 6                             // explains x
                                        const y = 6
```

The guide told people not to do it at all.

## What now works

A `//` comment stays attached to a complete declaration, statement, or match arm:

```
type UserId = string // stable identifier

node main() {
  const x = 5 // starting count

  match (x) {
    5 => "five" // the only case we expect
    _ => "other"
  }
}
```

It works in every body owner: node, def, if, else, while, for, thread, subthread, guard, handle, inline handlers, finalize, parallel, seq, destructive, block arguments, statement code literals, and both inline and block match arms. Comments also survive import sorting, which renders imports on a path that bypasses `processNode`.

## How

**Parsing.** One combinator, `completeConstructEntry`, wraps three existing streams: top level, bodies, and match arms. It parses the construct, and attaches a following `//` only if the construct's own parser did not already consume the line ending — that check is what keeps a standalone comment on the *next* line from being pulled up. Blank lines are sentinel characters by that point, so they count as line endings too.

`completeConstructEntry` is also the single owner of the layout that ends the line. The three call sites each used to consume that whitespace themselves; they no longer do, because `lineCommentCore` deliberately stops before the newline and something has to consume it exactly once.

The comment lands on `BaseNode.trailingComment`. No wrapper node, no change to `loc` — the owner's span still ends where it did, which the LSP and linter position tests confirm.

**Printing.** `commentText` is the only thing that builds `//${content}`; `appendTrailingComment` is the only thing that places one after code. `processNode` calls it once for every node type. Match-arm rendering is split into `renderInlineMatchArm` / `renderBlockMatchArm` so the comment attaches to a finished arm instead of being threaded through both branches.

**One type fix:** `Hole` declared its own `loc` rather than extending `BaseNode`, so a top-level hole could receive a trailing comment the type did not record. It now extends `BaseNode` with `loc` still required.

## Verification

- Full unit suite: **10,440 passed, 0 failed** (674 files).
- `make`: clean, so the whole stdlib and examples recompile through the changed parser with no working-tree changes.
- `pnpm run typecheck`: byte-identical to the `origin/main` baseline. That baseline is red — 15 pre-existing errors in `lib/serve/*` unrelated to this change.
- `pnpm run lint:structure`: clean.
- `pnpm run test:perf`: 16 tests passed; parse/fmt/doc scaling still linear (parse 25.4s, fmt 13.7s, doc 19.9s).

New tests: `lib/parsers/trailingComments.test.ts` (attachment, the newline boundary, and that `loc` excludes the comment), plus an 18-case body-owner matrix and fixed-point assertions in `lib/formatter.test.ts`. Every formatter case asserts the output reparses and that formatting twice is identical.

## Deliberate non-goals

Block comments never attach. Trivia nodes (comments, blank lines) never own a trailing comment. Multiline list items still relocate — the guide now says so explicitly rather than claiming end-of-line comments are disallowed.

## Deviation from the plan

The plan passed a `canAttach` policy only at the body stream. I made the trivia-node exclusion unconditional inside `withTrailingLineComment` instead, so top level and match arms get it too. Without that, `/* x */ // y` at top level would attach a line comment to a block-comment node, since `multiLineCommentParser` does not consume its line ending.

### DIFF
--- a/packages/agency-lang/docs/site/guide/basic-syntax.md
+++ b/packages/agency-lang/docs/site/guide/basic-syntax.md
@@ -208,32 +208,12 @@ You can also write a module-level doc comment using the `@module` tag. This docu
 */
 ```
 
-A `//` comment can also sit at the end of a line, after a complete
-declaration, statement, or match arm. `agency fmt` keeps it on that line,
-beside the thing it describes:
+> Note: comments must be on their own line, they cannot be at the end of a line containing code.
+
+Not allowed:
 
 ```ts
-type UserId = string // stable identifier
-
-node main() {
-  const x = 5 // starting count
-
-  match (x) {
-    5 => "five" // the only case we expect
-    _ => "other"
-  }
-}
-```
-
-Two things are not end-of-line comments. A block comment (`/* ... */`) after
-code is not attached, and neither is a `//` comment between the items of a
-multiline list — `agency fmt` still moves that onto its own line:
-
-```ts
-const xs = [
-  1, // fmt moves this comment above the 2
-  2
-]
+const x = 5 // this is a comment
 ```
 
 ## Functions
--- a/packages/agency-lang/docs/site/guide/basic-syntax.md
+++ b/packages/agency-lang/docs/site/guide/basic-syntax.md
@@ -208,12 +208,32 @@ You can also write a module-level doc comment using the `@module` tag. This docu
 */
 ```
 
-> Note: comments must be on their own line, they cannot be at the end of a line containing code.
-
-Not allowed:
+A `//` comment can also sit at the end of a line, after a complete
+declaration, statement, or match arm. `agency fmt` keeps it on that line,
+beside the thing it describes:
 
 ```ts
-const x = 5 // this is a comment
+type UserId = string // stable identifier
+
+node main() {
+  const x = 5 // starting count
+
+  match (x) {
+    5 => "five" // the only case we expect
+    _ => "other"
+  }
+}
+```
+
+Two things are not end-of-line comments. A block comment (`/* ... */`) after
+code is not attached, and neither is a `//` comment between the items of a
+multiline list — `agency fmt` still moves that onto its own line:
+
+```ts
+const xs = [
+  1, // fmt moves this comment above the 2
+  2
+]
 ```
 
 ## Functions
--- a/packages/agency-lang/lib/backends/agencyGenerator.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.ts
@@ -24,6 +24,7 @@ import {
   formatUnitLiteral,
   isExpressionNode,
 } from "../types.js";
+import { LineComment } from "../types/base.js";
 
 import { AccessChainElement, ValueAccess } from "../types/access.js";
 import { declaredName } from "../types/hole.js";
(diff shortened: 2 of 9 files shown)


packages/agency-lang/docs/site/guide/basic-syntax.md
```diff
@@ -208,12 +208,32 @@ You can also write a module-level doc comment using the `@module` tag. This docu
 */
 ```
 
-> Note: comments must be on their own line, they cannot be at the end of a line containing code.
+A `//` comment can also sit at the end of a line, after a complete
```
egonSchiele: Blanket rule across all PRs you will make for this change: Do not update user-facing docs.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -1481,6 +1469,38 @@ export class AgencyGenerator {
     return result;
   }
 
+  /** One arm, without its final newline — a complete construct an
+   *  attached comment can follow. */
+  private renderInlineMatchArm(
```
egonSchiele, line 1474: Will you be re-doing the agency generation for most node types across the 3 PRs? Because I think that will get really unwieldy. Or is this most of the agency generator change you expect to make?

egonSchiele: This is most of the generator change for **complete constructs**, but not most of the generator work for the overall three-PR feature. The follow-ups still need to integrate arrays/objects/object types, calls and signatures, imports/exports, patterns, and thread/parallel arguments.

That should not mean redoing most node generators. The intended boundary is one shared `renderListWithTrivia(...)` helper: each list owner supplies its existing item renderer and separator policy, while the helper owns indentation and before/trailing comment placement. If the follow-up diffs start reconstructing `//...` or manually placing comments in each node method, that would be a sign the abstraction has leaked and should be revised.

egonSchiele (commented):
Review result:

- No additional correctness blocker was found in the complete-construct implementation. The consumed-newline check, stream progress after `lineCommentCore`, match-arm rendering, import-sorting path, location preservation, and trivia-node exclusion are internally consistent.
- The user-facing guide change still needs to be removed before merge, as already called out inline.
- The generator design is appropriately centralized for this PR: complete constructs go through `processNode`/`appendTrailingComment`, with only imports and match arms needing explicit bypass handling. Follow-up list PRs should add one shared trivia-aware list renderer and adapt list owners to it, rather than adding comment-placement branches to each node renderer.

Verification performed against `22400ebc`: the focused parser/formatter/comment/match suites passed (157 tests), all GitHub build/test/lint/perf checks are green, and an independent read-only review found no further actionable finding.

packages/agency-lang/docs/site/guide/basic-syntax.md
egonSchiele: Reverted in e494f61 — `basic-syntax.md` is byte-identical to main again, and the PR is now code and tests only. Noted as a standing rule for the two follow-up PRs as well.

PR #768, opened 2026-08-02
fmt: keep end-of-line comments attached in multiline lists, calls, and signatures

egonSchiele:
Tasks 3 and 4 of the trailing-comments plan, targeting the integration branch. PR 1 (#767) covered complete constructs; this covers **multiline lists**: array and object literals, object types, call arguments, and `def` / `node` parameters. Imports, patterns, and thread/parallel arguments are the last PR.

Code and tests only — no doc changes.

## The problem

Inside a multiline list, a comment was anchored only to the item it came *before*. There was nowhere to record "this comment sits at the end of item N's line", so `fmt` pushed it onto its own line:

```
const xs = [        →      const xs = [
  1, // one                   1,
  2 // two                    // one
]                             2
                              // two
                            ]
```

## What now works

```
const xs = [
  1, // one
  2 // two
]

save(
  userId, // who to save
  retries: 3 // how hard to try
)

def save(
  value: string, // the payload
  ...rest: string[] // everything else
) { }
```

Covered: array literals, object literals, object types (including tagged properties), positional/named/splat call arguments, method calls, call chains, `interrupt`, every `raise` form, `guard`, and both `def` and `node` parameter lists.

## How

**One bit on trivia, not a new list.** `ListTrivia` became a union: `BeforeListTrivia` (what already existed) and `TrailingListTrivia` (`placement: "trailing"`). Parsers never write `placement: "before"`, so **every existing trivia AST is byte-for-byte unchanged** — the existing assertions in `dataStructures.test.ts` and `objectTypeTrivia.test.ts` are the compatibility proof and were not touched. `Trivia` and `ObjectTypeTrivia` survive as aliases.

**Two delimiter policies, deliberately not one.** `itemEntryAfterDelimiter` attaches a comment only when the item *and* its delimiter stayed on one line, so `first,\n// about second` still belongs to `second`. Object *types* need a separate policy (`objectMemberEntry`) because a newline is itself a legal delimiter there, so a comment can appear on either side of it. Merging these into one "universal separator" would have widened the grammar.

**One shared list renderer.** `renderListWithTrivia` owns indentation and both comment placements; callers pass their existing item renderer and a separator policy. `renderParenList` is the single entry point for `( ... )` lists — without trivia it is exactly the old `wrapList` call, with trivia it goes multiline. Nothing outside `commentText` builds `//...`, and no node renderer places a comment itself.

**Reordering.** The formatter moves an inline block argument last. `remapListTrivia` translates source anchors to print anchors so a before-comment stays with the item it preceded and a trailing comment with the item it followed. It throws if the permutation is malformed rather than silently dropping a comment. Tested with the block in first, middle, and last source position, asserting the block itself survives too — not just its comment.

## The one that would have slipped through

`guardBlockParser` hand-builds its node and copied only `arguments`, so trivia reached the parser and vanished before the AST. Widening the shared argument parser was not enough. Every current consumer of `argumentListParser` now has a test pinning it — that is why the matrix includes method calls, call chains, `interrupt`, `raise`, and `guard` separately rather than testing plain calls and assuming the rest follow.

Also worth flagging: the first version of these tests asserted `toContain("// one")`, which **passes even when the comment is relocated** — the exact bug. The assertions now pin whole lines (`"1, // one"`).

## Verification

- Full unit suite: **10,461 passed, 0 failed** (675 files), re-run after the final cleanup commit rather than relying on the earlier run.
- `make`: clean, so the whole stdlib and examples recompile through the new argument grammar with no working-tree changes.
- `tsc --noEmit`: no new errors (the 15 pre-existing `lib/serve/*` errors on the base branch are unchanged).
- `lint:structure`: clean.
- `test:perf`: 16 passed, parse/fmt/doc scaling still linear.

## Deviation from the plan

The plan specified a `CommaListPolicy` object with `cardinality` and `trailingComma` fields. All three call sites in this PR are identical (zero-or-more, trailing comma allowed), so the policy would carry no information yet. `commaDelimitedList(itemParser, closer)` is the shared engine now; PR 3 adds the policy fields when it has sites that actually differ — imports and patterns reject trailing commas where literals allow them. Adding unused configurability ahead of that is the kind of speculative generality the plan itself warns about.

Removed `emitTriviaAt`, which became dead once the three literal renderers moved to the shared helper. Nothing outside a stale `dist/` artifact referenced it.

### DIFF
--- a/packages/agency-lang/lib/backends/agencyGenerator.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.ts
@@ -25,6 +25,28 @@ import {
   isExpressionNode,
 } from "../types.js";
 import { LineComment } from "../types/base.js";
+import {
+  ListTrivia,
+  TriviaNode,
+  isTrailingListTrivia,
+} from "../types/dataStructures.js";
+
+/** One rendered list item. `leadingLines` are whole lines that print above
+ *  it (an object-type property's `@validate` tags); the shared renderer
+ *  indents them, so callers hand back plain text. */
+type RenderedListItem = {
+  leadingLines?: string[];
+  code: string;
+};
+
+type RenderListWithTriviaOptions<T> = {
+  items: T[];
+  trivia: ListTrivia[] | undefined;
+  open: string;
+  close: string;
+  renderItem: (item: T, index: number) => RenderedListItem;
+  separator: (index: number, itemCount: number) => string;
+};
 
 import { AccessChainElement, ValueAccess } from "../types/access.js";
 import { declaredName } from "../types/hole.js";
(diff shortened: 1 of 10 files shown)


packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -769,42 +793,90 @@ export class AgencyGenerator {
...
+    for (const node of this.beforeTriviaAt(args.trivia, args.items.length)) {
+      lines.push(this.renderTriviaNode(node));
+    }
+
+    this.decreaseIndent();
+    return `${args.open}\n${lines.join("\n")}\n${this.indentStr(args.close)}`;
   }
 
   protected aliasedTypeToString(aliasedType: VariableType): string {
     if (aliasedType.type === "objectType") {
```
egonSchiele, line 866: [P1] Object-type trivia is only preserved when the object type is the root value passed here. Recursive paths still fall through to `variableTypeToString` (including arrays/unions, `stringifyProp`, parameters, and return types), which ignores `ObjectType.trivia`. For example:

```agency
type Values = {
  value: number // keep
}[]
```

currently formats to `type Values = { value: number }[]`, deleting the comment. The same loss occurs for object-typed parameters/returns and nested object properties. Please make the trivia-aware object-type rendering recursive through type wrappers and those call sites, and add regressions for at least an array-wrapped object type plus a parameter or nested property.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -1694,45 +1700,229 @@ const triviaEntry: Parser<TriviaEntry> = seqC(
...
+  itemParser: Parser<T>,
+  delimiter: Parser<unknown>,
+): Parser<ItemEntry<T>> =>
+  (input: string): ParserResult<ItemEntry<T>> => {
+    const item = itemParser(input);
+    if (!item.success) {
+      return item as ParserResult<ItemEntry<T>>;
+    }
+
+    const beforeDelimiter = trailingLineCommentEntry(item.rest);
```
egonSchiele: [P1] This pre-delimiter attempt needs the same consumed-line guard as the post-delimiter path. A nested `objectTypeParser` consumes layout after its closing `}`, so a standalone comment on the following line is already at `item.rest` and gets attached to the previous property. Reproducer:

```agency
type Value = {
  nested: {
    x: number
  }
  // describes next
  next: string
}
```

Formatting currently moves the comment to `nested: { x: number }; // describes next`. Gate this attempt with `!consumedLineEnding(input, item.rest)` so the comment remains available to `triviaEntry`, and add this nested-member case to the boundary tests.

egonSchiele (commented):
Two correctness blockers found in nested object-type handling. Both were reproduced against this head. The focused list/parser/generator/formatter suites otherwise pass (622 tests across 12 files).

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Confirmed and fixed in c7cb69c. Your reproducer formatted to `nested: { x: number }; // describes next` on this head; it now keeps the comment on its own line, matching the base branch.

The pre-delimiter attempt is gated with `consumedLineEnding(input, item.rest)` as you suggested. Worth noting the guard has to use the trailing-whitespace rule rather than "any newline in the item", because a comment genuinely written on the nested closing brace's line should still attach:

```agency
type Value = {
  nested: {
    x: number
  } // about nested
  next: string
}
```

That still gives `nested: { x: number }; // about nested`. Both directions are now in the boundary tests.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -1694,45 +1700,229 @@ const triviaEntry: Parser<TriviaEntry> = seqC(
...
+    }
+
+    const separated = delimiter(item.rest);
+    if (!separated.success) {
+      return separated as ParserResult<ItemEntry<T>>;
+    }
+
+    if (consumedLineEnding(input, separated.rest)) {
+      return success({ kind: "item", item: item.result }, separated.rest);
+    }
```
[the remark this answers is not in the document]
egonSchiele: Correct, and fixed in c7cb69c. `first\n, // comment` did attach to `first`.

The fix judges the delimiter span on its own (`spanCrossesLine(item.rest, separated.rest)`) rather than the tail of everything consumed so far. That distinction matters: an item whose own text spans lines must still accept a trailing comment, so the check deliberately does not look inside the item.

```agency
const xs = [
  [
    1,
    2
  ], // outer
  third
]
```

Both cases are now pinned by tests.

packages/agency-lang/lib/backends/agencyGenerator.ts
egonSchiele: I verified this one against the base branch before acting, and I want to put the evidence in front of you because I think it changes what the right fix is.

**This is pre-existing, not a regression.** I checked out `adit/trailing-comments-integration` in a scratch worktree and formatted your reproducer shape with an ordinary before-comment:

```agency
type Values = {
  // keep
  value: number
}[]
```

Base branch output: `type Values = { value: number }[]` — the comment is already deleted today. Same for an object-typed parameter. So `ObjectType.trivia` has never survived a wrapper; my PR adds trailing comments to a hole that was already there rather than opening it.

**Why the fix is bigger than it looks.** The recursive path is `variableTypeToString` in `lib/backends/typescriptGenerator/typeToString.ts`. It is a pure function shared by TypeScript codegen and the Agency formatter (the `forFormatting` flag), it has no indentation state, it has never emitted multi-line output, and it prints every object type inline as `{ a: b; c: d }`. There are 41 call sites. Making trivia survive there means giving it indent state and a multi-line mode for the Agency dialect only — that is a real refactor of the type printer, not a threading change.

Put plainly: the Agency formatter cannot currently print a multi-line object type anywhere except as the root of a type alias. Comments are one symptom; blank lines between properties are lost in the same places.

**Why I did not do the cheap version.** I could special-case `arrayType` in `aliasedTypeToString` and make your exact reproducer pass. That fixes one wrapper and leaves unions, intersections, parameters, return types, and nested properties still deleting comments — a limitation that is harder to explain and predict than the current uniform one. I would rather not ship that.

**What I suggest.** Land this PR as is, and do the type-printer work as its own change where it can be reviewed on its own terms and cover blank lines too. I am happy to do it next instead of PR 3 if you would rather have it sooner, or to do it inside this PR if you want the whole feature to land together — just say which and I will go.

The two correctness blockers that were genuinely mine are fixed in c7cb69c.

PR #770, opened 2026-08-02
fmt: preserve object type trivia through wrappers

egonSchiele:
## Summary

- keep `variableTypeToString` as the canonical recursive type printer and add a declarative object-rendering hook
- preserve object-type comments, blank lines, validators, and descriptions through wrappers and every source-formatting type position
- keep direct `signatureOf`, function/node docs, and `std::agency` display signatures trivia-free
- preserve established full source bodies in `agency doc` type-alias reference blocks, including property comments and semantic metadata

## Context

This fixes a pre-existing formatter hole exposed by the trailing-comment work. The parser retained object-type trivia, but nested object types fell back to the shared inline printer and lost it.

Before:

```agency
type Users = { id: string // stable }[]
```

formatted as:

```agency
type Users = { id: string }[]
```

After, the comment and multiline layout survive. PR #768's nested-object comment attachment fix is already present in the base branch and is not changed here.

The broader wrapped-list continuation-indentation correction discovered during implementation is intentionally out of scope so this PR does not change unrelated formatter fixtures or generated example pages.

## Verification

- `make` — passed; generated `docs/site/` remained clean
- `pnpm test:run` — 674 files passed, 3 skipped; 10,499 tests passed, 42 skipped
- `pnpm run lint:structure` — passed
- `git diff --check` — passed

### DIFF
--- a/packages/agency-lang/lib/backends/agencyGenerator.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.ts
@@ -724,6 +724,19 @@ export class AgencyGenerator {
 
   // Wrapping helpers
 
+  private indentContinuationLines(item: string): string {
+    const continuationIndent = this.indent(1);
+    return item
+      .split("\n")
+      .map((line, index) => {
+        if (index === 0 || line === "") {
+          return line;
+        }
+        return continuationIndent + line;
+      })
+      .join("\n");
+  }
+
   private wrapList(
     items: string[],
     prefix: string,
@@ -739,7 +752,9 @@ export class AgencyGenerator {
     if (items.length === 0) return inline;
     if (this.indentStr(inline).length <= 80) return inline;
     this.increaseIndent();
-    const lines = items.map((item) => this.indentStr(`${item},`));
+    const lines = items.map((item) =>
+      this.indentStr(`${this.indentContinuationLines(item)},`),
+    );
     this.decreaseIndent();
     return `${prefix}${open}\n${lines.join("\n")}\n${this.indent()}${close}${suffix}`;
   }
--- a/packages/agency-lang/lib/backends/agencyGenerator.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.ts
@@ -12,6 +12,7 @@ import {
   Literal,
   MultiLineStringLiteral,
   NewLine,
+  ObjectType,
   ObjectProperty,
   ParallelBlock,
   Scope,
@@ -763,7 +764,21 @@ export class AgencyGenerator {
 
   // Type system methods
 
-  private stringifyProp(prop: ObjectProperty): string {
+  private renderTypeSource(type: VariableType): string {
+    return variableTypeToString(type, this.typeAliases, true, {
+      objectType: (objectType, printChild) => {
+        if (!objectType.trivia?.length) {
+          return undefined;
+        }
+        return this.renderObjectTypeSource(objectType, printChild);
+      },
+    });
+  }
+
+  private stringifyProp(
+    prop: ObjectProperty,
+    printChild: (child: VariableType) => string,
+  ): string {
     const isUnionWithNull =
       prop.value.type === "unionType" &&
       prop.value.types.some(
@@ -779,14 +794,14 @@ export class AgencyGenerator {
         nonNullTypes.length === 1
           ? nonNullTypes[0]
           : { type: "unionType", types: nonNullTypes };
-      let str = `${prop.key}?: ${variableTypeToString(unionWithoutNull, this.typeAliases, true)}`;
+      let str = `${prop.key}?: ${printChild(unionWithoutNull)}`;
       if (prop.description) {
         str += ` # ${prop.description}`;
       }
       return str;
     }
 
-    let str = `${prop.key}: ${variableTypeToString(prop.value, this.typeAliases, true)}`;
+    let str = `${prop.key}: ${printChild(prop.value)}`;
     if (prop.description) {
       str += ` # ${prop.description}`;
     }
@@ -864,21 +879,30 @@ export class AgencyGenerator {
 
   protected aliasedTypeToString(aliasedType: VariableType): string {
     if (aliasedType.type === "objectType") {
-      return this.renderListWithTrivia({
-        items: aliasedType.properties,
-        trivia: aliasedType.trivia,
-        open: "{",
-        close: "}",
-        // `@validate(...)` / `@jsonSchema(...)` annotations render on their
-        // own lines above the property, matching source layout.
-        renderItem: (prop) => ({
-          leadingLines: (prop.tags ?? []).map((tag) => this.formatTag(tag).trim()),
-          code: this.stringifyProp(prop),
-        }),
-        separator: (index, count) => (index === count - 1 ? "" : ";"),
-      });
+      return this.renderObjectTypeSource(aliasedType, (child) =>
+        this.renderTypeSource(child),
+      );
     }
-    return variableTypeToString(aliasedType, this.typeAliases, true);
+    return this.renderTypeSource(aliasedType);
+  }
+
+  private renderObjectTypeSource(
+    objectType: ObjectType,
+    printChild: (child: VariableType) => string,
+  ): string {
+    return this.renderListWithTrivia({
+      items: objectType.properties,
+      trivia: objectType.trivia,
+      open: "{",
+      close: "}",
+      // `@validate(...)` / `@jsonSchema(...)` annotations render on their
+      // own lines above the property, matching source layout.
+      renderItem: (prop) => ({
+        leadingLines: (prop.tags ?? []).map((tag) => this.formatTag(tag).trim()),
+        code: this.stringifyProp(prop, printChild),
+      }),
+      separator: (index, count) => (index === count - 1 ? "" : ";"),
+    });
   }
 
   protected processEffectDeclaration(node: EffectDeclaration): string {
--- a/packages/agency-lang/lib/backends/agencyGenerator.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.ts
@@ -724,19 +724,6 @@ export class AgencyGenerator {
 
   // Wrapping helpers
 
-  private indentContinuationLines(item: string): string {
-    const continuationIndent = this.indent(1);
-    return item
-      .split("\n")
-      .map((line, index) => {
-        if (index === 0 || line === "") {
-          return line;
-        }
-        return continuationIndent + line;
-      })
-      .join("\n");
-  }
-
   private wrapList(
     items: string[],
     prefix: string,
@@ -752,9 +739,7 @@ export class AgencyGenerator {
     if (items.length === 0) return inline;
     if (this.indentStr(inline).length <= 80) return inline;
     this.increaseIndent();
-    const lines = items.map((item) =>
-      this.indentStr(`${this.indentContinuationLines(item)},`),
-    );
+    const lines = items.map((item) => this.indentStr(`${item},`));
     this.decreaseIndent();
     return `${prefix}${open}\n${lines.join("\n")}\n${this.indent()}${close}${suffix}`;
   }
--- a/packages/agency-lang/lib/backends/agencyGenerator.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.ts
@@ -783,7 +783,10 @@ export class AgencyGenerator {
   private renderTypeSource(type: VariableType): string {
     return variableTypeToString(type, this.typeAliases, true, {
       objectType: (objectType, printChild) => {
-        if (!objectType.trivia?.length) {
+        if (
+          !objectType.trivia?.length &&
+          !this.objectTypeHasPropertyMetadata(objectType)
+        ) {
           return undefined;
         }
         return this.renderObjectTypeSource(objectType, printChild);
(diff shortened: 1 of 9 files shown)


packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -763,7 +780,31 @@ export class AgencyGenerator {
 
   // Type system methods
 
-  private stringifyProp(prop: ObjectProperty): string {
+  private renderTypeSource(type: VariableType): string {
+    return variableTypeToString(type, this.typeAliases, true, {
+      objectType: (objectType, printChild) => {
+        if (!objectType.trivia?.length) {
```
egonSchiele: This gate decides between trivia-aware layout and the shared inline printer using only `trivia`, but the inline printer also drops per-property `@validate` / `@jsonSchema` tags and `# description` text. So this still loses a validator through `agency fmt`:

```agency
type Wrapped = Container<{
  @validate(isPositive)
  id: number
}>
```

No comments means no trivia, so the hook delegates and this formats to `Container<{ id: number }>` — the validator is silently gone. That is exactly the hazard the comment in `formatPattern` warns about ("a formatter pass would quietly turn a validated rule into an unvalidated one"). The new "preserves tagged property descriptions" test only covers the case where a comment happens to be present alongside the tag.

This loss is pre-existing (nested objects always went through the inline printer before), so it is fair to call it out of scope — but the fix is now one line:

```ts
const needsSourceLayout =
  objectType.trivia?.length ||
  objectType.properties.some((p) => p.tags?.length || p.description);
```

If you'd rather keep this PR trivia-only, an issue to track it would do.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -323,10 +323,8 @@ export {
 import { extractSummaryOverride, moduleDescription } from "../utils/moduleDoc.js";
 
 function formatTypeAlias(alias: TypeAlias, ctx: DocContext): string {
-  const code = generateAgency({
-    type: "agencyProgram",
-    nodes: [alias],
-  });
+  const annotations = generator.declarationAnnotationPrefixOf(alias);
+  const code = `${annotations}export ${generator.signatureOf(alias)}`;
```
egonSchiele: This changes what `agency doc` shows for object-type aliases in a way that loses information. The old `generateAgency` path rendered the alias body in source layout, including per-property annotations and descriptions. The new path goes through `signatureOf` → display rendering → the inline printer, which prints only `key: type`. So for:

```agency
export type User = {
  @validate(isEmail)
  email: string # the contact address
}
```

the docs used to show the full body with `@validate(isEmail)` and `# the contact address`; now they show `export type User = { email: string }`. Alias-level tags survive via `declarationAnnotationPrefixOf`, and `formatValidatorsAndSchema(alias.tags)` also only covers alias-level tags — per-property validators and descriptions now appear nowhere in the generated docs. Descriptions are user-facing documentation content (and become LLM schema descriptions), so dropping them from the reference pages looks unintended.

A second, smaller effect of the same change: a large object alias that used to render multiline now renders as one long line in the code fence, since the display printer has no width handling for objects.

Was the compact form deliberate for aliases specifically? The PR body says signatures should stay "compact and trivia-free", which makes sense for def/node signatures, but for the alias reference section the body *is* the documentation. If keeping trivia (comments) out of docs was the goal, rendering the alias body via the source path minus comments — or gating like the formatter hook on tags/descriptions — would keep the annotations without the comments.

egonSchiele (commented):
Reviewed the whole diff plus the surrounding code on the branch. The core design is solid: the hook forwarding covers every recursive edge in `variableTypeToString` (I checked each branch), the `wrapList` continuation shift is correct at any nesting depth because items are rendered one level shallower than their wrapped placement and `indent(1)` is exactly one fixed unit, and the hardcoded `export ` prefix in `formatTypeAlias` is safe because that function is only called on `aliases.filter(a => a.exported)`. Test coverage is thorough and the exact-string + reparse + idempotence pattern is the right one for a formatter.

Two findings, both about the same underlying gap: the inline object printer prints only `key: type`, so anything routed through it silently drops per-property `@validate` / `@jsonSchema` tags and `# description` text. Details in the line comments.

One minor note on `lib/cli/doc.test.ts`: `expect(output).not.toContain("//")` will fail the moment a fixture gains a doc comment containing a URL. `not.toContain("// keep")` (as the other new doc test uses) pins the same regression without that trap.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -721,6 +724,19 @@ export class AgencyGenerator {
 
   // Wrapping helpers
 
+  private indentContinuationLines(item: string): string {
+    const continuationIndent = this.indent(1);
```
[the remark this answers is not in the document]
egonSchiele: No change here: `indent(1)` is intentionally one fixed indentation unit, not the final absolute indentation. It is prepended to continuation lines that were already rendered at the caller current depth; `wrapList` then applies the wrapped item indentation. Using the current absolute level here would double-count nesting. The nested wrapped-signature exact/idempotence regression covers this behavior.

packages/agency-lang/lib/cli/doc.ts
egonSchiele: Fixed in babd722e7. Display rendering now strips comment/blank-line trivia but preserves semantic property tags and descriptions. Root object aliases remain multiline in `agency doc` and `_describe`, restoring the previous reference-body layout without leaking `// keep` comments. Exact integration regressions cover both consumers.

packages/agency-lang/lib/backends/agencyGenerator.ts
egonSchiele: Fixed in babd722e7. Source layout now activates for property tags or descriptions even when the object has no comment trivia. The new exact fixed-point regression covers a metadata-only nested object, so formatter passes retain validators instead of falling back to the lossy inline object form.

egonSchiele (commented):
Reviewed by checking out the head and running things, not by reading. The hook design is a clear improvement on what I planned — one recursive printer with an optional object hook removes the duplicated precedence rules and the printer-agreement tests I had proposed, and the hook is forwarded through every recursive edge including `effectSetToSource` and `effectSetMemberToSource`. The core of this is right.

Three findings, all reproduced against `822fc6ad7`. Two of them come from the same root cause: the PR's verification ran four test files, and the things that break are outside those four.

**What I ran**

- Full unit suite **after `make`**: 674 files pass, **1 fails** (`lib/formatter.test.ts`). Worth noting that running the suite *without* building first shows 22 failures — 21 of those are stale-`dist` artifacts and are not real. I confirmed the same 21 fail on the base branch.
- `make`: succeeds, but leaves **43 checked-in files modified**.
- `pnpm run lint:structure`: clean.
- `tsc --noEmit`: no new errors beyond the 15 pre-existing `lib/serve/` ones.

**Findings**

1. **`lib/formatter.test.ts` fails — the fixture needs regenerating.** Not a regression; the PR actually *fixes* a long-standing misindentation. Details in the inline comment.
2. **`make` leaves 43 generated doc files dirty**, with a doc-comment behavior change and 15 new trailing-whitespace lines. Inline comment on `doc.ts`.
3. **`signatureOf` is not trivia-free**, contradicting a stated goal of the PR. Inline comment on `buildSignature`.

None of these is a design problem. Finding 1 is a fixture regeneration, finding 3 is a one-line fix, finding 2 needs a decision from you.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -721,6 +724,19 @@ export class AgencyGenerator {
...
+    return item
+      .split("\n")
+      .map((line, index) => {
+        if (index === 0 || line === "") {
+          return line;
+        }
+        return continuationIndent + line;
+      })
+      .join("\n");
```
egonSchiele: **[P1] This fails the full suite, and the fix is to regenerate the fixture rather than change the code.**

`lib/formatter.test.ts > round-trips a correctly formatted file unchanged` fails on this head. `tests/formatter/roundtrip.agency` contains:

```agency
  return _bash(
    command,
    cwd,
    timeout,
    stdin,
    {
    blockedCommands: blockedCommands
  },
  )
```

Note the object literal is misindented in the fixture — `blockedCommands` at 4 spaces where its siblings are at 6, and the closing brace at 2. That is what the formatter produced *before* this PR, and the fixture froze it.

`indentContinuationLines` fixes it. On this head the same input formats to:

```agency
  return _bash(
    command,
    cwd,
    timeout,
    stdin,
    {
      blockedCommands: blockedCommands
    },
  )
```

which is correctly aligned, and I confirmed it is idempotent. So the code is right and the fixture is stale — regenerate it (`make fixtures`) rather than touching `wrapList`.

**Worth calling out in the PR description:** this is a wider change than the title suggests. `wrapList` handles every wrapped list, so it improves indentation for any multi-line item — object literals, array literals, nested calls — not only object types. That is a good change, but it is a formatter output change beyond "preserve object type trivia", and it is why a fixture moved. Reviewers should expect it.

packages/agency-lang/lib/cli/doc.ts
egonSchiele: **[P1] This makes `make` dirty 43 checked-in files under `docs/site/`.**

`agency doc` output is generated into `docs/site/stdlib/**` and those files are committed. Switching `formatTypeAlias` from `generateAgency` to `signatureOf` changes that output, so after `make` the working tree has 43 modified files. The plan says not to edit `docs/site/`, but the effect here is worse than editing it: the committed docs no longer match what the code generates, so the next person to run `make` gets an unexplained diff.

Two separate changes are in there.

**1. Doc comments no longer appear in the code block.** For example `docs/site/stdlib/date.md`:

```diff
-/** A day of the week, lowercase. */
-export type DayOfWeek =
+export type DayOfWeek = 
   | "sunday"
```

`generateAgency` emitted the doc comment; `signatureOf` does not. The description is already rendered as prose directly above the block, so this removes a duplication and might be an improvement — but it is an unannounced user-facing change, and `validation.md` loses several multi-line doc comments the same way. Please make it deliberate either way.

**2. Trailing whitespace, 15 lines across the generated docs.** Note the `= ` with a trailing space in the diff above. `signatureOf` builds `type X = ` and then appends the alias body; when the body is a long union, `variableTypeToString` returns a string starting with a newline (`\n  | ...`), leaving the space stranded at end of line. Reproduced directly:

```
[type Long = 
  | "aaaaaaaaaa"
  | "bbbbbbbbbb"
]
```

The old `generateAgency` path did not do this. It is pre-existing in `signatureOf` but newly visible now that type aliases go through it.

**Suggested resolution:** decide whether the doc-comment drop is wanted. If yes, commit the regenerated docs with a line in the PR description saying the generated stdlib docs changed and why. If no, keep the doc comment in this path. Either way the trailing space should be trimmed, and `make` should leave the tree clean before this merges.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -1198,16 +1260,17 @@ export class AgencyGenerator {
...
   ): string {
     const returnTypeBang = node.returnTypeValidated ? "!" : "";
     const returnTypeStr = node.returnType
       ? ": " +
-        variableTypeToString(node.returnType, this.typeAliases, true) +
+        this.renderType(node.returnType, policy) +
         returnTypeBang
       : "";
     const raisesStr = this.formatRaisesClause(node.raises);
     return this.renderParenList(
```
egonSchiele, line 1287: **[P2] `signatureOf` is not trivia-free, which contradicts a stated goal of this PR.**

The PR description says it keeps `signatureOf`, `agency doc`, and `std::agency` display signatures "compact and trivia-free". The `display` policy only reaches the *type* rendering. `parameterTrivia` is still passed to `renderParenList` on the line below regardless of policy, and `renderParenList` switches to the multiline trivia layout whenever trivia is present — so the comments come out.

Reproduced against this head by calling `signatureOf` directly:

```
save(
  value: string, // the payload
  retries: number // how hard
)
```

That is what `agency doc` and `std::agency` `describe` will print for any function whose parameter list has an end-of-line comment. No stdlib function has one yet — that syntax only became available in #768 — which is why it does not show up in the regenerated docs and why the four test files in the verification section do not catch it. It is latent, not theoretical: the first stdlib parameter comment will put a `//` into a tool description handed to a model.

The fix looks like one line — withhold the trivia on the display path:

```ts
return this.renderParenList(
  this.renderParams(node.parameters, policy),
  policy === "source" ? node.parameterTrivia : undefined,
  prefix,
  `${returnTypeStr}${raisesStr}${opener}`,
);
```

Please add a `signatureOf` test with a commented parameter list so this stays pinned — that is the assertion that would have caught it.

packages/agency-lang/lib/backends/agencyGenerator.ts
egonSchiele: Resolved by narrowing this PR rather than widening its formatter-output scope. I removed the broad `indentContinuationLines` change and its dedicated test, restored the roundtrip fixture through a normal patch, and updated the plan/PR description. The type-trivia behavior remains; unrelated wrapped lists and generated examples retain their existing output. Fresh `make` leaves `docs/site` clean, and the full suite passes.

packages/agency-lang/lib/cli/doc.ts
egonSchiele: Fixed across 2a223275c and fba04d479. Alias doc comments remain in `agency doc` code blocks, preserving the established user-facing output. `composeTypeAliasAssignment` now handles newline-leading RHS values without leaving whitespace after `=`. Type-alias reference blocks use full source generation, while direct `signatureOf` remains trivia-free. A fresh `make` now leaves all checked-in `docs/site` files clean.

packages/agency-lang/lib/backends/agencyGenerator.ts
egonSchiele: Fixed in 2a223275c. `buildSignature` now passes `parameterTrivia` only for source rendering. Direct `signatureOf`, `agency doc` function/node signatures, and `_describe` omit parameter comments; source formatting still preserves them. Exact direct and integration regressions pin both sides of that boundary.

PR #779, opened 2026-08-02
test(fmt): a comment must come out where it went in

egonSchiele:
Follow-up item 4 from the review on #774. Test-only — no production code changes.

## Why

Three different mistakes move or drop a comment during formatting, and each one shipped at least once during the trailing-comments work:

1. A statement stream that skips `completeConstructEntry`.
2. A generator path that renders nodes without going through `processNode` — sorted imports do exactly this.
3. A node kind whose *second* parser or renderer was never updated. This happened four times: `guardBlockParser`, `namedImportParser`, `namedExportBodyParser`, and the mixed-import render path.

Every one was caught by a person reading output, not by a test. Two of them were caught in review after I had said the work was done. Each new construct currently needs someone to remember to add a case; this covers the category instead.

## What it does

Two layers, because one turned out not to be enough.

**Canonical samples must format to themselves, byte for byte.** 24 samples, one per construct that can carry a comment. Each is written in already-canonical form, so formatting is the identity function and exact equality is the assertion.

**A corpus check asserts no comment disappears** between the parsed tree and the formatted output, across `tests/formatter` and `tests/typescriptGenerator`. It walks for *any* comment node rather than reading the known trivia fields by name — a named list would go stale the moment someone adds a list kind, which is one of the failures this exists to catch. Counting is one pass into a table rather than a scan per comment.

A guard test asserts the corpus is non-empty and contains a meaningful number of comments, so the broad check cannot start passing vacuously if fixtures move.

## The part worth reviewing

My first version had only the corpus layer, and **it did not work.** I verified it by reintroducing each failure mode, and two of the three named in the review passed green:

- Dropping trivia in `namedImportParser` passed, because a comment lost at *parse* time is missing from both sides of the comparison.
- Removing `completeConstructEntry` from the body stream passed, because the comment does not vanish — it just relocates onto its own line, which is the original bug.

So "no comment vanishes" is a genuinely weak property. The canonical-sample layer exists because of that, and the file says so in a header comment, including the instruction to add a sample rather than relying on the corpus when adding a construct.

## Verification

Each failure mode reintroduced in turn, confirming the suite fails on all four:

| Sabotage | Result |
|---|---|
| `processNode` drops the attached comment | 3 failed |
| `namedImportParser` drops `nameTrivia` (loss at parse time) | 3 failed |
| body stream skips `completeConstructEntry` (comment moves) | 2 failed |
| sorted imports skip their comment | 1 failed |
| restored | 26 passed |

Production files confirmed byte-identical afterwards.

`lint:structure` clean; the new file typechecks. Per speedy mode I did not run `make` or the full suite — this adds one test file and touches no production code, so CI is the check on those.

## Not doing

Item 6 from the same review — making the body-owner test table self-extending from a parser registry — is filed separately rather than done here. It means restructuring how the statement parser is assembled, which is core grammar machinery, to buy test coverage that this PR now gets from the outside. The review itself called it "defense in depth, not urgent" once this exists.

### DIFF
--- a/packages/agency-lang/lib/commentConservation.test.ts
+++ b/packages/agency-lang/lib/commentConservation.test.ts
@@ -1,0 +1,232 @@
+import { describe, expect, it } from "vitest";
+import fs from "fs";
+import path from "path";
+import { fileURLToPath } from "url";
+import { formatSource } from "./formatter.js";
+import { parseAgency } from "./parser.js";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+/**
+ * A comment must come out of `agency fmt` where it went in.
+ *
+ * Three different mistakes move or drop a comment, and all three have shipped
+ * at least once: a statement stream that skips `completeConstructEntry`, a
+ * generator path that renders nodes without going through `processNode`
+ * (sorted imports do this), and a node kind whose second parser or renderer
+ * was never updated (`guardBlockParser`, `namedImportParser`,
+ * `namedExportBodyParser`, and the mixed-import renderer each did this).
+ *
+ * There are two layers here, because one is not enough:
+ *
+ * 1. SAMPLES are written in canonical form and must format to themselves,
+ *    byte for byte. This is the strong check: it catches a comment that moves
+ *    as well as one that vanishes, and it catches loss during PARSING.
+ *
+ * 2. The corpus check asserts only that no comment disappears between the
+ *    parsed tree and the output. It covers far more real code, but it is
+ *    deliberately weaker — a comment dropped at parse time is missing from
+ *    both sides and passes, and a comment that merely relocates still
+ *    appears somewhere. Do not rely on it alone for a new construct; add a
+ *    sample.
+ *
+ * The corpus collector walks for ANY comment node rather than reading the
+ * known trivia fields by name. A named list would go stale the moment someone
+ * adds a list kind — which is one of the failures this file exists to catch.
+ */
+
+/** Every `//` comment the parser produced, wherever it ended up: a standalone
+ *  node, a `trailingComment`, or an entry in any list's trivia. */
+function collectComments(value: unknown, found: string[] = []): string[] {
+  if (Array.isArray(value)) {
+    for (const item of value) {
+      collectComments(item, found);
+    }
+    return found;
+  }
+  if (value === null || typeof value !== "object") {
+    return found;
+  }
+  const node = value as { type?: unknown; content?: unknown };
+  if (node.type === "comment" && typeof node.content === "string") {
+    found.push(node.content.trim());
+  }
+  for (const child of Object.values(value as Record<string, unknown>)) {
+    collectComments(child, found);
+  }
+  return found;
+}
+
+/** How many times each `//` comment body appears in `source`, counted in one
+ *  pass so a large corpus does not turn into a scan per comment. */
+function countCommentLines(source: string): Record<string, number> {
+  const counts: Record<string, number> = {};
+  for (const match of source.matchAll(/\/\/(.*)$/gm)) {
+    const body = match[1].trim();
+    counts[body] = (counts[body] ?? 0) + 1;
+  }
+  return counts;
+}
+
+/** The comments in `source` that are missing from `formatted`. */
+function missingComments(source: string, formatted: string): string[] {
+  const parsed = parseAgency(source, {}, false, false);
+  if (!parsed.success) {
+    return [];
+  }
+  const remaining = countCommentLines(formatted);
+  const missing: string[] = [];
+  for (const comment of collectComments(parsed.result.nodes)) {
+    if ((remaining[comment] ?? 0) > 0) {
+      remaining[comment] -= 1;
+    } else {
+      missing.push(comment);
+    }
+  }
+  return missing;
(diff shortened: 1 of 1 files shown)


packages/agency-lang/lib/commentConservation.test.ts
```diff
@@ -0,0 +1,232 @@
...
+  return missing;
+}
+
+function agencyFilesIn(relativeDir: string): string[] {
+  const dir = path.join(__dirname, "..", relativeDir);
+  if (!fs.existsSync(dir)) {
+    return [];
+  }
+  return fs
+    .readdirSync(dir)
```
egonSchiele: **P1 — The corpus walk silently omits nested fixtures.** `readdirSync(dir)` only inspects the immediate directory, so this currently checks 82 files while the two stated corpus directories contain 96 `.agency` files recursively. In particular, all 10 `tests/typescriptGenerator/euler/*.agency` fixtures contain line comments but never reach this test; the `corpus.length > 10` guard still passes, so it does not expose the omission. Please recurse (or reuse an existing recursive collector such as `findRecursively` / the helper in `agencyGenerator.roundtrip.test.ts`) so the test actually covers the advertised directories.

packages/agency-lang/lib/commentConservation.test.ts
```diff
@@ -0,0 +1,232 @@
...
+  const node = value as { type?: unknown; content?: unknown };
+  if (node.type === "comment" && typeof node.content === "string") {
+    found.push(node.content.trim());
+  }
+  for (const child of Object.values(value as Record<string, unknown>)) {
+    collectComments(child, found);
+  }
+  return found;
+}
+
```
egonSchiele, line 60: **P1 — Compare parsed comments on both sides rather than raw `//` text.** This regex also counts `//` inside strings and multiline comments. That can mask the exact regression this test is intended to catch: if a parsed `// keep` comment is dropped but the formatted program already contains `"// keep"`, `remaining["keep"]` is still nonzero and the test passes. Parse `formatted` and build the count table from `collectComments(formattedResult.nodes)` instead; that remains one pass and ensures only actual comment nodes can satisfy the conservation check.

packages/agency-lang/lib/commentConservation.test.ts
egonSchiele: Confirmed and fixed in e9d1866. Your numbers were exact: 82 files flat, 96 recursive. The gap is `tests/typescriptGenerator/euler` (10 files) and `tests/typescriptGenerator/types` (4).

Now uses `findRecursively` from `lib/utils/findRecursively.ts` rather than a fourth copy of a directory walker.

I also replaced the guard, since as you say `corpus.length > 10` could never have caught this — it now requires more than 50 files *and* asserts at least one comes from a nested directory, so a future non-recursive read fails rather than silently shrinking the corpus.

packages/agency-lang/lib/commentConservation.test.ts
egonSchiele: Agreed and fixed in e9d1866 — this was the better call.

Both sides are now counted as parsed comment nodes: the formatted output is re-parsed and `collectComments` runs on it, so only real comment nodes can satisfy the check. Still one pass per side. The failure mode you describe is now impossible rather than unlikely, and it also removes the block-comment false positive.

One useful side effect: if the formatted output fails to re-parse, that now surfaces as a named failure instead of being invisible.

packages/agency-lang/lib/commentConservation.test.ts
```diff
@@ -0,0 +1,232 @@
...
+/** How many times each `//` comment body appears in `source`, counted in one
+ *  pass so a large corpus does not turn into a scan per comment. */
+function countCommentLines(source: string): Record<string, number> {
+  const counts: Record<string, number> = {};
+  for (const match of source.matchAll(/\/\/(.*)$/gm)) {
+    const body = match[1].trim();
+    counts[body] = (counts[body] ?? 0) + 1;
+  }
+  return counts;
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in e9d1866, though by a different route than tightening the regex — the text scan is gone entirely. Both sides are counted as parsed comment nodes now, so a `//` inside a string literal or a block comment cannot contribute at all. Your example, `https://` in `tests/typescriptGenerator/objectDescription.agency`, was reaching the count table exactly as you describe.

packages/agency-lang/lib/commentConservation.test.ts
```diff
@@ -0,0 +1,232 @@
+import { describe, expect, it } from "vitest";
+import fs from "fs";
+import path from "path";
+import { fileURLToPath } from "url";
+import { formatSource } from "./formatter.js";
+import { parseAgency } from "./parser.js";
```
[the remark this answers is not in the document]
egonSchiele: Correct, and fixed in e9d1866 at all four sites. `formatSource` parses `replaceBlankLines(source)`, so parsing raw source here gave this test a different tree than the formatter actually worked from — blank lines are sentinels by parse time and statement boundaries depend on it.

There is now one `parseLikeFormatter` helper that every call goes through, so the two cannot drift apart again. `agencyGenerator.roundtrip.test.ts` already does the same thing, which I should have followed.

PR #783, opened 2026-08-02
fmt: indent a multi-line item with the list that wrapped it

### LINKED ISSUE #775: fmt: a multi-line item inside a wrapped list keeps its continuation lines at the outer indent
`agency fmt` indents only the first line of a list item. When a list is long enough that `wrapList` breaks it across lines, an item that is itself multi-line keeps its continuation lines at the outer indent.

## Reproducer

```agency
def f() {
  return _bash(
    command,
    cwd,
    timeout,
    stdin,
    {
      blockedCommands: blockedCommands
    },
  )
}
```

Formats to:

```agency
  return _bash(
    command,
    cwd,
    timeout,
    stdin,
    {
    blockedCommands: blockedCommands
  },
  )
```

`blockedCommands` lands at 4 spaces where its siblings are at 6, and the closing brace at 2 where it should be at 4.

## Cause

`wrapList` in `lib/backends/agencyGenerator.ts` renders its items into strings *before* deciding to wrap. When it then increases the indent and lays them out, `indentStr` prefixes only each item's first line; the rest were already baked at the outer depth.

`renderListWithTrivia` does not have this problem, because its `renderItem` is a callback invoked after the indent increases.

## Severity

Cosmetic. The output re-parses and formatting it again is a stable fixed point, so nothing is corrupted.

## Notes for whoever picks this up

`tests/formatter/roundtrip.agency` currently freezes the wrong output, so it must be regenerated (`make fixtures`) as part of the fix, or the round-trip test fails.

Expect object literals, array literals, and nested calls to shift — not just object types. #770 fixed this and backed the fix out to keep that PR's scope tight; the diff is small (shift continuation lines by one level when the list wraps) but the fixture churn is what makes it worth its own PR.

egonSchiele:
Fixes #775.

## The bug

`wrapList` renders each item into a string *before* it knows whether the list will wrap. When the list then wraps, `indentStr` moves only each item's first line — so an item that is itself multi-line keeps its remaining lines at the outer indent:

```agency
  return _bash(
    command,
    cwd,
    timeout,
    stdin,
    {
    blockedCommands: blockedCommands
  },
  )
```

`blockedCommands` sits at 4 spaces where its siblings are at 6, and the closing brace at 2 where it should be at 4.

## The fix

Shift the second and later lines of a rendered item by one level when the list wraps. Blank lines are left blank rather than becoming trailing whitespace.

This is the change #770 made and then backed out to keep that PR's scope tight. The reason it needed its own PR is the fixture churn below, not the diff — the fix itself is about ten lines.

## Two sets of regenerated files, both worth a look

**`tests/formatter/roundtrip.agency`** froze the old, wrong output, so the round-trip test failed until it was regenerated. The diff is exactly the two lines above.

**Three generated documentation pages** changed: `docs/site/examples/layoutDemo.md`, `docs/site/examples/mcp/github-oauth-ts/agent.md`, and `docs/site/stdlib/skills.md`. Their embedded examples contain nested lists inside wrapped calls, so they were showing the mangled indentation. For instance `layoutDemo`:

```diff
     columns: [{
-    align: "start"
-  }, {
-    align: "end"
-  }],
+      align: "start"
+    }, {
+      align: "end"
+    }],
```

I know the standing rule is not to touch `docs/site/` in a feature PR. These are generated output rather than authored content, and the alternative is worse: leaving them stale means `make` produces an unexplained dirty tree for the next person, which is the exact problem I flagged on #770. Flagging it rather than burying it — say the word if you would rather these came out and landed separately.

## Scope

This affects every list that wraps on length, so object literals, array literals, and nested calls all shift, not just the case in the issue. All of the movement is in the same direction: content that was under-indented now lines up with its opening delimiter.

## Verification

- New tests for an object literal, a nested array with a comment, and blank-line handling inside a wrapped item — each asserting the output re-parses and formatting twice is identical.
- `formatter`, `agencyGenerator`, `agencyGenerator.roundtrip`, `commentConservation`, and `listTrailingComments` suites: 285 passed.
- `make`: succeeds; the only working-tree changes are the regenerated files committed here.

Per speedy mode I did not run the full unit suite or perf; CI covers those.

### DIFF
--- a/packages/agency-lang/lib/backends/agencyGenerator.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.ts
@@ -725,31 +725,20 @@ export class AgencyGenerator {
 
   // Wrapping helpers
 
-  /** Shift the second and later lines of an already-rendered item by one
-   *  level. `wrapList` builds its items before it knows the list will wrap,
-   *  so a multi-line item comes in indented for the outer level; `indentStr`
-   *  only moves the first line. Blank lines stay blank rather than becoming
-   *  trailing whitespace. */
-  private indentContinuationLines(item: string): string {
-    if (!item.includes("\n")) {
-      return item;
-    }
-    const continuationIndent = this.indent(1);
-    return item
-      .split("\n")
-      .map((line, index) =>
-        index === 0 || line === "" ? line : continuationIndent + line,
-      )
-      .join("\n");
-  }
-
+  /** `renderItems` is a callback, not an array, because a multi-line item has
+   *  to be built at the depth it will actually print at. Rewriting an
+   *  already-rendered item's lines instead would reach inside triple-quoted
+   *  strings, whose newlines are data rather than layout, and silently change
+   *  the program. So the items are rendered once to measure, and rendered
+   *  again after the indent increases if the list turns out to wrap. */
   private wrapList(
-    items: string[],
+    renderItems: () => string[],
     prefix: string,
     open: string,
     close: string,
     suffix: string = "",
   ): string {
+    const items = renderItems();
     const inline = `${prefix}${open}${items.join(", ")}${close}${suffix}`;
     // An empty list never wraps: wrapping zero items prints `(\n\n)`,
     // which the parser rejects (found by the round-trip gate on a
@@ -758,9 +747,7 @@ export class AgencyGenerator {
     if (items.length === 0) return inline;
     if (this.indentStr(inline).length <= 80) return inline;
     this.increaseIndent();
-    const lines = items.map((item) =>
-      this.indentStr(`${this.indentContinuationLines(item)},`),
-    );
+    const lines = renderItems().map((item) => this.indentStr(`${item},`));
     this.decreaseIndent();
     return `${prefix}${open}\n${lines.join("\n")}\n${this.indent()}${close}${suffix}`;
   }
@@ -1312,7 +1299,7 @@ export class AgencyGenerator {
       : "";
     const raisesStr = this.formatRaisesClause(node.raises);
     return this.renderParenList(
-      this.renderParams(node.parameters, policy),
+      () => this.renderParams(node.parameters, policy),
       policy === "source" ? node.parameterTrivia : undefined,
       prefix,
       `${returnTypeStr}${raisesStr}${opener}`,
@@ -1457,14 +1444,14 @@ export class AgencyGenerator {
    *  for a call that means `renderArgs` has appended any inline block last,
    *  matching how `extractInlineBlock` remapped the anchors. */
   protected renderParenList(
-    rendered: string[],
+    renderItems: () => string[],
     trivia: ListTrivia[] | undefined,
     prefix: string,
     suffix: string,
   ): string {
-    const body = this.renderListIfTrivia(rendered, trivia, "(", ")");
+    const body = this.renderListIfTrivia(renderItems(), trivia, "(", ")");
     if (body === undefined) {
-      return this.wrapList(rendered, prefix, "(", ")", suffix);
+      return this.wrapList(renderItems, prefix, "(", ")", suffix);
     }
     return `${prefix}${body}${suffix}`;
   }
@@ -1479,7 +1466,7 @@ export class AgencyGenerator {
   ): string {
     const rendered = this.renderArgs(args, block);
     if (trivia?.length) {
-      return this.renderParenList(rendered, trivia, "", "");
+      return this.renderParenList(() => rendered, trivia, "", "");
     }
     return `(${rendered.join(", ")})`;
   }
@@ -1497,9 +1484,8 @@ export class AgencyGenerator {
 
     const block = node.block;
     const inlineBlock = block?.inline ? block : undefined;
-    const rendered = this.renderArgs(node.arguments, inlineBlock);
     let result = this.renderParenList(
-      rendered,
+      () => this.renderArgs(node.arguments, inlineBlock),
       node.argumentTrivia,
       `${asyncPrefix}${declaredName(node.functionName)}`,
       "",
@@ -1533,7 +1519,7 @@ export class AgencyGenerator {
     // so each comment gets its own line.
     if (!node.trivia?.length) {
       const items = node.items.map((item) => this.renderArrayItem(item));
-      return this.wrapList(items, "", "[", "]");
+      return this.wrapList(() => items, "", "[", "]");
     }
 
     return this.renderListWithTrivia({
@@ -1913,7 +1899,7 @@ export class AgencyGenerator {
         return this.indentStr(`${importKeyword}${withTrivia}${suffix}`);
       }
       return this.indentStr(
-        this.wrapList(names, importKeyword, "{ ", " }", suffix),
+        this.wrapList(() => names, importKeyword, "{ ", " }", suffix),
       );
     }
 
--- a/packages/agency-lang/lib/backends/agencyGenerator.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.ts
@@ -1518,8 +1518,15 @@ export class AgencyGenerator {
     // (which may render on a single line). Trivia forces the multi-line form
     // so each comment gets its own line.
     if (!node.trivia?.length) {
-      const items = node.items.map((item) => this.renderArrayItem(item));
-      return this.wrapList(() => items, "", "[", "]");
+      // Built inside the callback, not before it: `wrapList` may increase the
+      // indent and ask again, and an item that spans lines has to be built at
+      // the depth it will print at.
+      return this.wrapList(
+        () => node.items.map((item) => this.renderArrayItem(item)),
+        "",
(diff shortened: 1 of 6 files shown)


packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -740,7 +758,9 @@ export class AgencyGenerator {
     if (items.length === 0) return inline;
     if (this.indentStr(inline).length <= 80) return inline;
     this.increaseIndent();
-    const lines = items.map((item) => this.indentStr(`${item},`));
+    const lines = items.map((item) =>
+      this.indentStr(`${this.indentContinuationLines(item)},`),
```
egonSchiele: **P1 — This changes the value of multiline string arguments.** `item` is opaque rendered source, so not every newline marks structural continuation code. For example:

```agency
def f() {
  return someFunctionWithAnExtremelyLongName(
    firstArgument,
    """line1
line2""",
    thirdArgument
  )
}
```

The long call wraps, and this changes the parsed string value from `"line1\nline2"` to `"line1\n  line2"`. I confirmed this with a parse → format → parse structural comparison; the new focused suites still pass because the corpus has no multiline string in a wrapped list. Please avoid rewriting already-rendered strings blindly. The safe boundary is to render each item after `wrapList` has chosen/increased the indentation (for example, accept item-rendering callbacks or a structured rendered-item representation), so syntax indentation is generated correctly without touching literal contents. Please add this case as a semantic round-trip regression test.

egonSchiele: You are right, and this was the wrong approach rather than a missing case. Fixed in 66b6b72 by taking the boundary you suggested.

I reproduced your example first. The parsed value went from `"line1\nline2"` to `"line1\n  line2"` — my change was silently altering program data to fix a cosmetic misalignment, which is a far worse trade than the one it was making.

`wrapList` now takes a render callback instead of an array. Items are rendered once to measure, and rendered again after the indent increases if the list actually wraps, so the layout is generated rather than patched. Nothing rewrites rendered source any more. `renderParenList` forwards the callback, and the signature and call-expression sites pass thunks.

Two things worth knowing that came out of it:

**It is also more conservative, and that is a good thing.** An item whose renderer hardcodes its own indentation rather than using the generator's is now left alone, where blind rewriting used to shift it. Concretely, a long union from `variableTypeToString` emits `"\n  | a"` with absolute two-space indentation; re-rendering does not change it. That drops `docs/site/stdlib/skills.md` back to main's exact output, so the generated-docs churn in this PR went from three files to two. Both remaining ones are genuine nested-literal fixes.

**Cost.** Items are rendered twice when a list wraps, once otherwise. `test:perf` passes with format scaling still linear. If that ever matters, the fix is to memoize per call rather than to go back to patching strings.

Added both regression tests you asked for — one for the unindented case and one where the string carries its own indentation — comparing the parsed multiline string VALUES before and after formatting, not just the layout, so a future change to this area fails on meaning rather than on whitespace.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -725,6 +725,24 @@ export class AgencyGenerator {
...
+      return item;
+    }
+    const continuationIndent = this.indent(1);
+    return item
+      .split("\n")
+      .map((line, index) =>
+        index === 0 || line === "" ? line : continuationIndent + line,
+      )
+      .join("\n");
+  }
```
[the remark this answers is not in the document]
egonSchiele: Correct — fixed in 66b6b72. `indentContinuationLines` is gone entirely rather than taught about string literals; `wrapList` now re-renders items at the chosen indent so nothing rewrites already-rendered source. Confirmed the reported corruption first: `"line1\nline2"` became `"line1\n  line2"`.

packages/agency-lang/lib/formatter.test.ts
```diff
@@ -428,3 +428,33 @@ describe("trailing comments in every body owner", () => {
...
+  });
+
+  it("leaves a blank line inside a wrapped item unindented", () => {
+    const source = `def f() {\n  return someFunctionWithALongName(\n    firstArgument,\n    secondArgument,\n    {\n      a: 1,\n\n      b: 2\n    },\n    thirdArgument\n  )\n}\n`;
+    const once = formatSource(source);
+    expect(once).toContain("\n\n      b: 2");
+    expect(once).not.toContain("  \n");
+    expect(formatSource(once as string)).toBe(once);
+  });
+});
```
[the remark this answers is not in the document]
egonSchiele: Added in 66b6b72, and they earned their place — the bug was real, not hypothetical. Two cases: a multiline string whose continuation has no indentation, and one that carries its own. Both compare the parsed string VALUES across a format cycle rather than checking layout, so the assertion is about meaning.

You were also right that the existing suite could not have caught this: none of the fixtures had a triple-quoted string inside a list long enough to wrap.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -1513,7 +1519,7 @@ export class AgencyGenerator {
     // so each comment gets its own line.
     if (!node.trivia?.length) {
       const items = node.items.map((item) => this.renderArrayItem(item));
-      return this.wrapList(items, "", "[", "]");
+      return this.wrapList(() => items, "", "[", "]");
```
egonSchiele: **P1 — Arrays still cache their items before the wrapping decision.** This callback returns the same strings rendered at the outer depth, so it does not satisfy `wrapList`'s new contract. The original under-indentation remains when an array itself wraps around a multiline item. I confirmed that formatting this canonical source moves the property from 6 spaces to 4 and the object close from 4 spaces to 2:

```agency
def f() {
  return [
    firstArgument,
    secondArgument,
    thirdArgument,
    {
      blockedCommands: blockedCommands
    },
  ]
}
```

Build the array items inside the callback instead:

```ts
return this.wrapList(
  () => node.items.map((item) => this.renderArrayItem(item)),
  "",
  "[",
  "]",
);
```

Please add this direct array-wrapper case; the existing nested-array test covers an array rendered again by its surrounding call, which is why it passes despite this path still being stale.

egonSchiele: Right on both counts — fixed in a75ba485.

Passing a thunk over an already-computed list defeats the whole point, and I did it at two sites: arrays and the single named-import path. Both now build their items inside the callback. Your reproducer confirmed it, and imports were harmless only by luck (their items are single-line).

Your note about why the existing test passed was the useful part. The nested-array case only passes because the surrounding call re-renders the array; nothing covered an array that wraps on its own. That test exists now, and I checked it fails against the stale version rather than assuming.

One more generated docs line moved as a result — the nested `body: [...]` block in `layoutDemo.md`, which was the worst-looking case in the file:

```diff
       [
-      text(\"1\"),
-      text(\"lib/foo.ts\"),
+        text(\"1\"),
+        text(\"lib/foo.ts\"),
```

Verification: formatter 69 passed; generator, round-trip, comment-conservation and doc suites 289 passed; `make` clean with only the intended files changed. Against main the generated-docs churn is two files.

Worth saying plainly: this is the second round where the finding was that I manipulated rendered output instead of generating it. The first version rewrote strings, this one cached them. For a fix I filed as cosmetic, that is a lot of review for the value — so if you would rather drop this and close #775 as not worth it, that is a completely reasonable call and I will not argue it. The code is correct now if you do want it.

packages/agency-lang/lib/backends/agencyGenerator.ts
```diff
@@ -1437,14 +1444,14 @@ export class AgencyGenerator {
...
    *  matching how `extractInlineBlock` remapped the anchors. */
   protected renderParenList(
-    rendered: string[],
+    renderItems: () => string[],
     trivia: ListTrivia[] | undefined,
     prefix: string,
     suffix: string,
   ): string {
-    const body = this.renderListIfTrivia(rendered, trivia, "(", ")");
+    const body = this.renderListIfTrivia(renderItems(), trivia, "(", ")");
```
egonSchiele, line 1452: **P1 — This eagerly renders every no-trivia list before handing it to `wrapList`, making nested calls exponential.** Function calls normally have no trivia, so `renderItems()` runs here and then immediately runs again inside `wrapList` merely to measure. Each nested call repeats that duplication for its child. I measured formatting `f(f(...f(x)))` on this head:

```text
depth 18:   48 ms
depth 20:  154 ms
depth 22:  605 ms
depth 24: 2409 ms
```

The PR description says items render once unless wrapping, but this path currently renders twice even when nothing wraps (and three times when it does). Check trivia before evaluating the callback:

```ts
if (!trivia?.length) {
  return this.wrapList(renderItems, prefix, "(", ")", suffix);
}
const body = this.renderListIfTrivia(renderItems(), trivia, "(", ")");
return `${prefix}${body}${suffix}`;
```

Please add a nested-call scaling regression; the existing perf check's broad/flat corpus does not exercise this recursive shape.

PR #784, opened 2026-08-02
A template can only use names it declares itself (AG8015)

egonSchiele:
## What this does

A template can now only use names it declares or imports itself, plus the prelude and the language builtins. Using a name that comes from the host file, or one that arrives through a hole, is a new error: **AG8015**.

Today a code literal body is never name-checked at all, so this compiles and fails at run time:

```ts
def helper(): string {
  return "hi"
}

static const template = [|
  node main(): string {
    return helper()   // helper will not exist in the generated program
  }
|]
```

`toSource` prints that template into its own file and `runCode` compiles the print. The host file is not there, so `helper` is gone. The same holds for a name a hole supplies: the hole is opaque while the template is checked, so the compiler cannot know what fills it.

## Behavior changes

**This rejects templates that compile today.** Any template that leans on a host declaration or on a filler-supplied declaration now fails to compile. That is the intended breaking change.

**A file with holes no longer emits AG4004 or AG4007 for a plain name.** A `.agency` file with holes is itself a template, so the always-on AG8015 pass owns every undefined variable and bare call in it, and the two ordinary passes stand down for those positions. One missing name produces one diagnostic instead of two. The JS-namespace chain check in the undefined-function pass (`Math.nosuch()`) keeps running, because AG8015 treats `nosuch` as a method rather than a lexical name.

## How it works

- `lib/typeChecker/nameReferences.ts` — the rules for which AST positions actually hold a lexical name, moved out of the two existing diagnostics so all three passes agree. It carries the easy-to-miss skips: `for` binders, block-call parameters (which the typechecker's `Scope` does not track at all), property names, and synthetic calls.
- `lib/typeChecker/templateNames.ts` — the analysis. It builds a `TypeCheckerContext` from the template body and the config alone, so host declarations, imports, aliases and errors are absent by construction rather than by discipline. It mirrors `buildScopes`, one scope per body, so a nested `def`'s parameters resolve.
- The pass is always on, unlike the ordinary undefined-name checks, which are silent by default. Nothing downstream catches this before run time.

## Fixtures and docs

- Deleted `tests/agency/templates/literalCompose.agency` and its `.test.json`. It was built on exactly the pattern this rejects: one literal called `guarded()`, which the `#helpers` hole supplied.
- Added `tests/agency/templates/literalComposeSelfContained.agency`, which keeps all the same coverage — expression-literal filling, partial fill, grafting into a wrapper, hole origin, a second fill, `toSource`, `runCode`, interrupt approval, the `"lit-ok"` result — with every fragment declaring the names it uses.
- `composeGuardedGuard.agency` and `composeGuardedMain.agency` change for the same reason: `main` moves next to the `guarded()` it calls, and the wrapper keeps a declaration of its own so `#helpers` still sits at a declaration position.
- `docs/site/guide/template-agency.md` gets a short section on what names a template can use, with the rejected form and the fix.

### DIFF
--- a/packages/agency-lang/lib/typeChecker/nameReferences.ts
+++ b/packages/agency-lang/lib/typeChecker/nameReferences.ts
@@ -1,0 +1,87 @@
+import type { AgencyNode, FunctionCall, VariableNameLiteral } from "../types.js";
+import type { WalkAncestor } from "../utils/node.js";
+
+/**
+ * Which AST positions hold a lexical name that has to resolve.
+ *
+ * Shared by the ordinary undefined-name diagnostics and the template one,
+ * so all three make the same call. Not every `variableName` is a lookup
+ * and not every `functionCall` is a bare call, and a pass that gets those
+ * rules half right reports correct code.
+ */
+
+/** Is this node inside a `def` or `node` body?
+ *
+ *  Those bodies have their own scope, so the top-level walk skips them or
+ *  every name in them fires twice. */
+export function hasFunctionOrNodeAncestor(
+  ancestors: readonly unknown[],
+): boolean {
+  for (const ancestor of ancestors) {
+    const type = (ancestor as AgencyNode | undefined)?.type;
+    if (type === "function" || type === "graphNode") {
+      return true;
+    }
+  }
+  return false;
+}
+
+/**
+ * Is this `variableName` a real read that has to resolve?
+ *
+ * Three positions look like reads and are not. A property name in
+ * `obj.x` is not a lookup. A `for` binder is a declaration. A block-call
+ * parameter, as in `xs.map(\(item) -> item)`, is a binding the
+ * typechecker's `Scope` does not track at all, so resolving it would
+ * always fail.
+ */
+export function isResolvableVariableReference(
+  ref: VariableNameLiteral,
+  ancestors: readonly WalkAncestor[],
+): boolean {
+  const parent = ancestors[ancestors.length - 1] as AgencyNode | undefined;
+  if (!parent) {
+    return false;
+  }
+  // The base of `obj.x` is a real reference. The property is not, and
+  // walkNodes does not yield it as a standalone name anyway.
+  if (parent.type === "valueAccess" && parent.base !== ref) {
+    return false;
+  }
+  for (const ancestor of ancestors) {
+    const node = ancestor as AgencyNode;
+    if (node.type === "forLoop") {
+      if (ref.value === node.itemVar || ref.value === node.indexVar) {
+        return false;
+      }
+    }
+    if ((ancestor as { type: string }).type === "blockArgument") {
+      const block = ancestor as {
+        type: "blockArgument";
+        params: { name: string }[];
+      };
+      if (block.params.some((param) => param.name === ref.value)) {
+        return false;
+      }
+    }
+  }
+  return true;
+}
+
+/**
+ * Is this `functionCall` a bare call whose name has to resolve?
+ *
+ * A synthetic call has no declaration to find, and the builder compiles
+ * it away. A call under a `valueAccess` is `obj.foo()`, where `foo` is a
+ * method rather than a lexical name; the access chain checks it instead.
+ */
+export function isResolvableBareCall(
+  call: FunctionCall,
+  ancestors: readonly WalkAncestor[],
+): boolean {
+  if (call.synthetic) {
+    return false;
+  }
+  const parent = ancestors[ancestors.length - 1] as AgencyNode | undefined;
+  return parent?.type !== "valueAccess";
+}
--- a/packages/agency-lang/lib/typeChecker/nameReferences.ts
+++ b/packages/agency-lang/lib/typeChecker/nameReferences.ts
@@ -29,11 +29,15 @@ export function hasFunctionOrNodeAncestor(
 /**
  * Is this `variableName` a real read that has to resolve?
  *
- * Three positions look like reads and are not. A property name in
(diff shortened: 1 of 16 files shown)


packages/agency-lang/lib/typeChecker/templateNames.ts
```diff
@@ -0,0 +1,215 @@
...
+ * before run time.
+ */
+export function checkTemplateNames(ctx: TypeCheckerContext): void {
+  // A file with holes is itself a template, so this pass owns every name
+  // in it, not just the ones inside literals. The ordinary undefined-name
+  // passes stand down for such a file.
+  if (holeNames(ctx.programNodes).length > 0) {
+    const findings = findUndefinedTemplateNames(ctx.programNodes, ctx.config);
+    reportTemplateNameFindings(findings, ctx);
+    return;
```
egonSchiele: This `return` skips the literal loop below whenever the file also has a top-level hole. The whole-file analysis cannot cover those literals: the generic walker deliberately treats `codeLiteral.nodes` as opaque. As a result, `#fileHole` plus `const t = [| missing() |]` emits no AG8015 for `missing`. Please continue into the literal loop after reporting the whole-file findings; the two walks cover disjoint bodies, so that should not duplicate diagnostics. A regression test with both a file-level hole and an invalid literal would pin this.

packages/agency-lang/lib/typeChecker/templateNames.ts
```diff
@@ -0,0 +1,215 @@
...
+  return buildScopes(context).map((info) => ({
+    body: info.body,
+    kind: info.name === "top-level" ? "topLevel" : "definition",
+    resolvesVariable: (name: string) =>
+      PRELUDE.has(name) ||
+      resolveVariable(name, {
+        functionDefs: context.functionDefs,
+        nodeDefs: context.nodeDefs,
+        importedFunctions: context.importedFunctions,
+        importedNodeNames: [],
```
egonSchiele: Hard-coding `importedNodeNames` to `[]` rejects valid template-local `import node` declarations. `walkScopeBody` declares ordinary `importStatement` names, but it has no `importNodeStatement` case, so there is no lexical-scope fallback for an imported node. Please collect these names from the synthetic program (the existing `collectProgramShadowing(nodes)` helper already does this) and pass them to both resolvers. Tests should cover both a direct call and a first-class reference to an imported node.

packages/agency-lang/lib/typeChecker/undefinedFunctionDiagnostic.ts
```diff
@@ -35,6 +40,11 @@ export function checkUndefinedFunctions(
   const mode = ctx.config.typechecker?.undefinedFunctions ?? "warn";
   if (mode === "silent") return;
 
+  // A file with holes is a template. AG8015 owns every call name in it, so
+  // reporting here too would double up. The JS-namespace chain check below
+  // goes quiet in such a file as well.
+  if (holeNames(ctx.programNodes).length > 0) return;
```
egonSchiele: This whole-pass return also disables the JS namespace-member validation below for every holey file. AG8015 does not replace that check: it resolves `Math` as a JS global and deliberately treats `nosuch` in `Math.nosuch()` as a method rather than a lexical bare call. Consequently that typo now goes silent. Please suppress only the `checkBareCall` branch for holey files and continue running `checkAccessChain`; a holey-file `Math.nosuch()` test would protect the behavior.

packages/agency-lang/lib/typeChecker/nameReferences.ts
```diff
@@ -0,0 +1,87 @@
...
+): boolean {
+  const parent = ancestors[ancestors.length - 1] as AgencyNode | undefined;
+  if (!parent) {
+    return false;
+  }
+  // The base of `obj.x` is a real reference. The property is not, and
+  // walkNodes does not yield it as a standalone name anyway.
+  if (parent.type === "valueAccess" && parent.base !== ref) {
+    return false;
+  }
```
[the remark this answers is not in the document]
egonSchiele: Confirmed and fixed in 97a69c9. walkNodes walks index and slice expressions with the valueAccess as their parent (lib/utils/node.ts:444-448), so the old rule skipped them. Property names are never yielded as standalone names, so the exclusion had nothing legitimate left to catch and is gone. Two tests added: obj[missingIndex] reports, obj[i] with i declared does not.

packages/agency-lang/lib/typeChecker/templateNames.ts
egonSchiele: Fixed in 97a69c9 — the return is gone and the literal loop runs after the whole-file report. Regression test added with a file-level hole plus a literal calling missingInLiteral; it asserts the literal name is among the reported ones.

packages/agency-lang/lib/typeChecker/templateNames.ts
egonSchiele: Fixed in 97a69c9. templateNameScopes now calls collectProgramShadowing(nodes) and passes importedNodeNames to both resolvers. Tests cover a direct call and a first-class reference to an imported node; I checked they are not vacuous by misspelling the call in the same source and confirming only the typo is reported.

packages/agency-lang/lib/typeChecker/undefinedFunctionDiagnostic.ts
egonSchiele: Fixed in 97a69c9. The pass no longer returns early; it sets isTemplateFile and skips only the functionCall branch, so checkAccessChain keeps running. Test added: a holey file with Math.nosuch(1) still reports AG4004.

PR #787, opened 2026-08-03
fix(fmt): render unwrapped nested calls once

egonSchiele:
## What changed

Avoid rendering parenthesized-list items before checking whether the list has trivia. On the normal no-trivia path, `wrapList` now owns the first render.

This fixes the recursive duplication introduced by #783: each unwrapped function call rendered its arguments once in `renderParenList` and again in `wrapList`, so nested calls recursively doubled the work at every level.

## Regression test

The new test uses a counting generator over a 12-deep unwrapped call expression. It asserts that the 12 function-call nodes are each processed once.

- Before the fix: 4,095 function-call processings
- After the fix: 12 function-call processings

This is deterministic and avoids timing-based assertions.

## Verification

- Focused formatter, generator, round-trip, comment-conservation, and trailing-comment suites: 290 passed
- `pnpm run lint:structure`
- `pnpm exec tsc --noEmit`

### DIFF
--- a/packages/agency-lang/lib/backends/agencyGenerator.test.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.test.ts
@@ -3,7 +3,18 @@ import { AgencyGenerator, generateAgency } from "./agencyGenerator.js";
 import { parseAgency } from "../parser.js";
 import { formatSource } from "../formatter.js";
 import { deepFreeze } from "../runtime/utils.js";
-import { FunctionDefinition } from "../types.js";
+import { AgencyNode, FunctionDefinition } from "../types.js";
+
+class ProcessCountingGenerator extends AgencyGenerator {
+  functionCallsProcessed = 0;
+
+  override processNode(node: AgencyNode): string {
+    if (node.type === "functionCall") {
+      this.functionCallsProcessed++;
+    }
+    return super.processNode(node);
+  }
+}
 
 describe("AgencyGenerator - Function Parameter Type Hints", () => {
   describe("processFunctionDefinition", () => {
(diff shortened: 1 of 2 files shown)


egonSchiele (commented):
Reviewed by checking out the head and measuring, not by reading. **This is correct and I would merge it.** It fixes a regression I introduced in #783, and the diagnosis is exactly right.

## Verified

- The counting test genuinely fails without the fix: **4095 → 12**, reproduced by reverting the two-line change on this branch. 4095 is 2^12 − 1, so the doubling was per level, as claimed.
- Focused formatter / generator / round-trip / comment-conservation / trailing-comment suites: 290 passed, matching the PR body.
- `tsc --noEmit`: clean.
- The fix location is the only one that needed it. The other `renderListIfTrivia` callers (patterns, imports) pass already-computed arrays and do not recurse through `wrapList`'s callback, so they are single-render and unaffected.

## What I got wrong, for the record

When I raised the perf question on #783 I benchmarked only *nested wrapped lists* and reported the cost as applying to \"lists that actually wrap, nested inside each other\" — rare in real code. That was incomplete. `renderParenList` called `renderItems()` unconditionally before checking for trivia, so **every** nested call doubled, wrapped or not. My benchmark never caught it because it used wrapped lists exclusively, where both doublings are present and I attributed all of the cost to the one I knew about.

## Where this leaves the two doublings

I measured this branch against `ac7d6a9a` (the commit before #783):

| shape | pre-#783 | this PR |
|---|---|---|
| unwrapped nested calls, depth 10 / 14 / 18 | 7.7 / 5.6 / 4.3 ms | 8.2 / 4.0 / 4.0 ms |
| nested **wrapped** lists, depth 6 / 8 / 9 | 30 / 119 / 241 ms | 41 / 273 / 946 ms |

So unwrapped nesting is fully restored to pre-#783 behavior — flat, no growth. The wrapped-list cost is unchanged by this PR and still ~3.9x at depth 9, which matches what I measured and disclosed on #783: `wrapList` renders its items a second time when it decides to wrap, and that is inherent to rendering at the correct indent rather than patching strings afterwards.

That residual is the known, accepted cost rather than something this PR missed. Worth being explicit about it in the PR body so it is not read as \"the exponential is gone\" — it is gone for the common case and remains for deeply nested wrapped lists.

## One minor point

`renderParenList` now checks `!trivia?.length` and then calls `renderListIfTrivia`, which checks the same condition again internally. `body` is typed `string | undefined` and is interpolated directly:

```ts
const body = this.renderListIfTrivia(renderItems(), trivia, \"(\", \")\");
return `${prefix}${body}${suffix}`;
```

Unreachable today, because the two conditions are identical. But if `renderListIfTrivia` ever gains another early return, this prints the literal string `undefined` into formatted source rather than failing. Calling `renderListWithTrivia` directly here, or keeping an explicit undefined check, removes the coupling. Not a blocker.

PR #789, opened 2026-08-03
A template must import the types it names

egonSchiele:
Closes part of #719.

## What this does

AG8015 says a template can only use names it declares or imports itself. Until now it only looked at *value* names — variables and calls. Type names were checked by nothing at all, so this compiled clean:

```ts
type Person = {
  name: string
}

static const template = [|
  const p: Person = { name: "a" }   // Person is the host file's, not the template's
|]
```

That template does not compile once `toSource` prints it and `runCode` runs it, because `Person` is not there. It is also the shape behind #719: a hole annotated with a borrowed type can never be checked at fill, since `fill` has no way to know what `Person` is either.

Type names now follow the same rule as everything else.

## Where the check came from

Mostly from work already being done and thrown away. `templateNames.ts` builds a context from the template body alone, and building its scopes validates every annotated declaration against the template's own aliases (`lib/typeChecker/scopes.ts:141`). Those diagnostics landed in the isolated context's error array, which the pass discarded wholesale. Instrumenting it showed exactly what was being lost:

```
CASE annotatedConst   ["AG1006 Type alias 'Person' is not defined (referenced in 'p')."]
CASE holeAnnotation   ["AG1006 Type alias 'Person' is not defined (referenced in 'p')."]
CASE defParam         []
CASE defReturn        []
```

So the change is three small things:

- Keep the unknown-type findings instead of discarding them, and report them as AG8015 so the message teaches the template rule.
- Add a short walk over `def` and `node` signatures, since building scopes covers annotated declarations and nothing else. `FunctionDefinition` has no type parameters, so there are no bound names to mistake for unknown ones.
- Treat a name the template imports itself as known. No file is read to prove it, exactly as an imported function is accepted without one.

## What it deliberately does not check

**Alias bodies.** `type Box<T> = { value: T }` binds `T`, and `validateTypeReferences` has no notion of a bound name, so checking bodies would report every generic alias a template declares. A test pins that this stays quiet.

**Template files.** A `.agency` file with holes keeps reporting AG1006 for an unknown type. Its imports resolve for real, so that pass is already correct there; reporting again just put two codes on one name. Measured before deciding — the holey file produced `["AG1006","AG8015"]`.

**Whether the imported type actually exists.** An import makes the name known, not correct. Resolving it means reading other modules at compile time and carrying the result into the compiled program so `fill` can use it. That is the rest of #719, and we decided the carrying cost is not worth it for now: a `Code` value would have to hold checking metadata that merges correctly every time fragments graft together, and the error it buys is one `runCode` already reports, just later.

## Checks

- 45 tests in `templateNames.test.ts`, 5 of them red before the change. The guards are the point: a declared type, an imported type, `Result<string>`, `Record<string, number>`, primitives, arrays, and a generic alias body.
- `lib/typeChecker`, `lib/typescriptGenerator`, `lib/runtime/template`: 1978 passed.
- `agency tc` over `tests/agency`, `examples`, `lib` and `stdlib`: no new AG8015. Nothing in the repo relied on borrowing a type.
- `make` exit 0, `lint:structure` clean.

### DIFF
--- a/packages/agency-lang/lib/typeChecker/templateNames.ts
+++ b/packages/agency-lang/lib/typeChecker/templateNames.ts
@@ -265,7 +265,12 @@ function importedTypeStubs(
       if (entry.type !== "namedImport") continue;
       for (const name of entry.importedNames) {
         if (typeof name !== "string") continue;
-        stubs[entry.aliases[name] ?? name] = { body: ANY_T };
+        // Own-property only: `import { toString }` would otherwise read the
+        // alias map's prototype and key the stub by a function.
+        const local = Object.hasOwn(entry.aliases, name)
+          ? entry.aliases[name]
+          : name;
+        stubs[local] = { body: ANY_T };
       }
     }
   }
@@ -313,10 +318,13 @@ function isolatedContext(
     matchExprYieldTypes: {},
     config,
     // Imported stubs first, so a name the template also declares wins.
-    getTypeAliases: () => ({
-      ...imported,
-      ...unit.typeAliases.visibleIn(currentScopeKey),
-    }),
+    // Null-prototype: a type named `toString` must miss, not find a method.
+    getTypeAliases: () =>
+      Object.assign(
+        Object.create(null) as Record<string, TypeAliasEntry>,
+        imported,
+        unit.typeAliases.visibleIn(currentScopeKey),
+      ),
     withScope<T>(key: string, fn: () => T): T {
       const previous = currentScopeKey;
       currentScopeKey = key;
--- a/packages/agency-lang/lib/typeChecker/templateNames.ts
+++ b/packages/agency-lang/lib/typeChecker/templateNames.ts
@@ -13,11 +13,19 @@ import { resolveCall } from "./resolveCall.js";
 import { resolveVariable } from "./resolveVariable.js";
 import { buildScopes } from "./scopes.js";
 import { collectProgramShadowing } from "./shadowing.js";
+import { validateTypeReferences } from "./validate.js";
 import { ANY_T } from "./primitives.js";
 import type { AgencyConfig } from "../config.js";
-import type { AgencyNode, AgencyProgram, CodeLiteral } from "../types.js";
+import type {
+  AgencyNode,
+  AgencyProgram,
+  CodeLiteral,
+  TypeAliasEntry,
+} from "../types.js";
 import type { SourceLocation } from "../types/base.js";
-import type { TypeCheckerContext } from "./types.js";
+import type { FunctionDefinition } from "../types/function.js";
+import type { GraphNodeDefinition } from "../types/graphNode.js";
+import type { TypeCheckError, TypeCheckerContext } from "./types.js";
 
 const PRELUDE = new Set(PRELUDE_NAMES);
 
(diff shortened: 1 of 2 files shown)


packages/agency-lang/lib/typeChecker/templateNames.ts
```diff
@@ -166,6 +244,34 @@ function templateNameScopes(
...
+): Record<string, TypeAliasEntry> {
+  const stubs: Record<string, TypeAliasEntry> = Object.create(null);
+  for (const node of nodes) {
+    if (node.type !== "importStatement") continue;
+    for (const entry of node.importedNames) {
+      if (entry.type !== "namedImport") continue;
+      for (const name of entry.importedNames) {
+        if (typeof name !== "string") continue;
+        stubs[entry.aliases[name] ?? name] = { body: ANY_T };
+      }
```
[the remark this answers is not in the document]
egonSchiele: Both fixed in 362b248, and both were real.

The alias-map read now goes through Object.hasOwn, so an import of a name like toString keys its stub by the name rather than by a function.

The suppressed comment about the merged map was right too: I was spreading a null-prototype object into a plain one, which handed the prototype chain back. getTypeAliases now builds the merged map with Object.create(null), so a template naming toString as a type misses instead of finding a method.

A test pins each one, and I checked they are not vacuous by reverting each fix on its own — the matching test fails and the other passes.

packages/agency-lang/lib/typeChecker/templateNames.ts
```diff
@@ -79,9 +90,73 @@ export function findUndefinedTemplateNames(
...
+ * `validateTypeReferences` has no notion of a bound name, so checking one
+ * would report every generic alias a template declares.
+ */
+function undefinedSignatureTypes(
+  nodes: AgencyNode[],
+  context: TypeCheckerContext,
+): UndefinedTemplateName[] {
+  const errors: TypeCheckError[] = [];
+  const aliases = context.getTypeAliases();
+  for (const { node } of walkNodes(nodes)) {
```
egonSchiele, line 132: This supplemental walk only covers def/node signatures, so other source-authored type slots still borrow host types silently. For example, `[| schema(HostType) |]`, `value is HostType`, and an actual inline hole annotation `#value: HostType` are neither assignment `typeHint`s nor signatures. (The current “hole annotation” test is `const p: Person = #person`, which exercises the assignment annotation instead.) Please replace the signature-only walk with an explicit type-slot walk that also validates `schemaExpression.typeArg`, type-pattern/type-test hints, and `Hole.typeAnnotation`; add at least those three regression forms.

packages/agency-lang/lib/typeChecker/templateNames.ts
```diff
@@ -79,9 +90,73 @@ export function findUndefinedTemplateNames(
...
+  }
   return found;
 }
 
+/** The unknown-type findings among diagnostics, by their recorded name. */
+function unknownTypesIn(
+  errors: readonly TypeCheckError[],
+): UndefinedTemplateName[] {
+  return errors
+    .filter((error) => error.name === "unknownTypeAlias")
```
egonSchiele, line 109: `validateTypeReferences` uses a different diagnostic for an unresolved generic application: `Box<string>` produces `unknownGenericType`, not `unknownTypeAlias`. This filter therefore discards the finding, so a host `type Box<T>` can still be borrowed in either an annotated declaration or a signature without AG8015. Please convert both diagnostic names here (both carry `params.alias`) and add declaration + parameter tests using a host-only `Box<string>`.

packages/agency-lang/lib/typeChecker/templateNames.ts
```diff
@@ -79,9 +90,73 @@ export function findUndefinedTemplateNames(
...
+    }));
+}
+
+/**
+ * Types named in a `def` or `node` signature.
+ *
+ * `walkScopeBody` validates annotated declarations and nothing else, so a
+ * parameter or return type would otherwise go unchecked.
+ *
+ * Alias bodies stay unchecked on purpose: `type Box<T>` binds `T`, and
```
egonSchiele, line 122: Skipping alias bodies leaves a direct escape from the rule: `[| type Local<T> = { item: T, host: HostType } |]` is accepted even when only the host defines `HostType`. Generic parameters do not require skipping the body—the ordinary checker already handles this by adding each `typeParam` as an opaque local alias before calling `validateTypeReferences` (`TypeChecker.check`, around lines 183–212). Please reuse that strategy for template-local alias bodies, preserving the existing `Box<T>` green case while adding a borrowed `HostType` red case.

PR #790, opened 2026-08-03
ci: gate the test/eval typecheck and fix its outstanding errors

egonSchiele:
## What

`pnpm run typecheck` type-checks three things — the source (`tsc --noEmit`), the tests (`tsconfig.tests.json`), and the evals (`tsconfig.evals.json`) — but nothing in CI ever ran it. So type errors in test files could pile up without any PR going red. This PR fixes the errors that had already accumulated and adds a CI gate so they can't come back silently.

## The two problems

1. **The tests typecheck was broken outright.** `tsconfig.tests.json` set `incremental: false` while inheriting `tsBuildInfoFile` from the base config. TypeScript rejects that combination (TS5069), so the check aborted before type-checking a single file. It now points at its own buildinfo file under `.agency-build/` (gitignored, like the base config's), which resolves the conflict without clobbering the main build's buildinfo.

2. **10 stale-fixture errors in `lib/serve/` tests.** Once the check actually ran, test fixtures had drifted from the real types:
   - `FuncParam` requires `hasDefault` / `defaultValue` / `variadic` — `discovery.test.ts` built params with just `{ name }`.
   - `ExportedFunction` requires `parameters` — the `mcp/adapter.test.ts` and `mcp/httpTransport.test.ts` fixtures omitted it.

   Updated the fixtures to match the current types (the added `parameters` mirror each function's real params).

## The gate

New `.github/workflows/typecheck.yml` runs `pnpm run typecheck` on every PR to `main`. No build step needed — `tsc` only type-checks.

## Verification

- `pnpm run typecheck` → exit 0 (all three configs clean).
- `vitest run lib/serve/discovery.test.ts lib/serve/mcp/adapter.test.ts lib/serve/mcp/httpTransport.test.ts` → 44 passed.

### DIFF
--- a/.github/workflows/typecheck.yml
+++ b/.github/workflows/typecheck.yml
@@ -1,0 +1,31 @@
+# Security: all actions pinned to SHA. Do not use pull_request_target.
+name: Typecheck
+
+on:
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: typecheck-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  typecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: pnpm/action-setup@f40ffcd9367d9f12939873eb1018b921a783ffaa # v4
+        with:
+          version: 9
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: 22
+          cache: pnpm
+      - run: pnpm install --frozen-lockfile
+      # Type-checks the source, the tests (tsconfig.tests.json), and the evals
+      # (tsconfig.evals.json). No build needed — tsc only type-checks.
+      - run: pnpm run typecheck
+        working-directory: packages/agency-lang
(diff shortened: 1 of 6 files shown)


egonSchiele (commented):
One blocking issue: the new typecheck gate is red on this PR, so the stated “all three configs clean” outcome is not true for the current head. CI reports:

```
lib/cli/remote/exportedEndpoints.test.ts(21,71): error TS2339:
Property 'error' does not exist on type 'SafeDeleteResult'.
```

`SafeDeleteResult` exposes `message?: string`, not `error`. Please fix that call site (and rerun the complete `pnpm run typecheck`) as part of this PR or update this branch after the fix lands elsewhere. Since this PR’s purpose is to establish a clean required gate, it should merge with the newly added check green rather than knowingly establish a failing baseline.

The workflow/config approach otherwise looks sound: the tests config gets a separate inherited-incremental build-info path, action versions are pinned, permissions are read-only, and the fixture updates match the production types. `git diff --check` and structural lint are clean.

egonSchiele:
Fixed. The red gate was correct — it was catching a real error. Two things:

- `exportedEndpoints.test.ts` referenced `result.error`, but `SafeDeleteResult` exposes `message?`. Changed the call site to `result.message` (442526ae8).
- That file wasn't in my worktree originally (it branched off an older main), so my first local `pnpm run typecheck` never saw it. Merged latest main into the branch, which is how I reproduced the failure locally.

`pnpm run typecheck` now exits 0 across all three configs on the current head, and `exportedEndpoints.test.ts` passes (2 tests).

egonSchiele (commented):
Re-reviewed the updated head. The previous blocker is fixed correctly: the cleanup diagnostic now uses `SafeDeleteResult.message`, and the complete typecheck reaches all three configs successfully.

No further blocking findings. I verified:

- `pnpm run typecheck` locally: passed (`tsc --noEmit`, tests config, evals config)
- `git diff --check`: clean
- CI: typecheck, structural lint, both builds, integration, agent tests, all Agency shards, perf, and coverage are green

The separate tests build-info path and fixture updates remain consistent with the production types.

PR #793, opened 2026-08-03
feat(eval): human-label store and agency eval label

egonSchiele:
## What this adds

`agency eval label <source> --checklist <file>` — a terminal checklist UI for judging agent outputs by hand, writing to a durable append-only store that outlives the runs it labels.

This is stage 1 of the labeling → judge → agreement loop in `docs/superpowers/specs/2026-08-02-eval-labeling-design.md`. No judges, no agreement measurement, no calibration: those are Fast follow, and the spec explains why measuring per-question judge agreement has to come first.

## Why the store is outside `runs/`

Every other eval artifact can be regenerated by re-running an agent. A human judgement cannot. So the store (default `labels/`, configurable via a new `eval.labelStore`) holds **copies** of the outputs it labels rather than pointers into a run directory, and deleting `runs/` never destroys labels.

## The three findings from spec review that shaped this

**Occurrence identity, not content identity.** `outputId` hashes `(traceId, inputId, finalOutputIndex)` — never a path. Directories get renamed and copied and two unrelated runs can share a basename, none of which changes which execution produced an output. `contentHash` covers `(input, value)` together and is used only for deduplication, because `"Looks good"` means different things under "Review this patch" than under "Write a release note".

**Immutable checklist revisions.** A mutable file with an incrementing integer cannot reconstruct what version 3 contained once version 4 exists, and annotations name a version, so later agreement measurement needs the exact rubric a person used. Every change writes a full snapshot; `current.json` is a pointer, not a second copy.

**Per-question folding.** Latest-annotation-wins breaks soft delete: judge a question, delete it, sign off again covering only live questions, then restore it, and the earlier answer vanishes. The fold is per question, keyed on `(outputId, checklistId, annotatorKind, annotatorId)` — keying on kind alone would merge every human, and later every judge regardless of prompt.

## Three defects found by the tests, not by review

**`openJsonlStrict` only detected duplicate identities on append.** A file that already contained two rows sharing an id loaded clean and could then be labelled against. Duplicates are now rejected when the log is indexed.

**Capture returned an existing row without comparing content.** Reusing an output id with different content silently drifted existing labels onto different text. It now refuses, excluding `capturedAt` and provenance since those can differ between two captures of the same execution.

**A store invariant contradicted its own recovery story.** `validateLineages` treated a `current` pointer lagging the newest revision as fatal — but that is exactly what a crash between the immutable rename and the pointer update leaves, and validation runs *before* recovery. The rule made that state permanently unrepairable while its error text promised reopening would fix it. A lag is now a warning; a pointer naming a revision that does not exist is still fatal.

## Layering

```
TUI        -> controller -> store facade -> private persistence modules
controller -> pure session reducer and selectors
store      -> pure annotation fold, status, score
```

`labelTui.ts` imports none of `draft`, `checklist`, `corpus`, `capture`, `lock`, `store`. The controller is the only module that may compose capture, drafts, publication, annotation commits and recovery, so the sign-off ordering cannot be reproduced — or mis-reproduced — anywhere else. Verified by grep in the audit.

## Verification

- **283 tests** across the labeling modules and CLI; **542** across all of `lib/eval` with no regressions.
- `tsc --noEmit` and `pnpm run lint:structure` both clean.
- **Crash recovery is exercised at five named durable boundaries** (`after-pending-annotation-save`, `after-annotation-append`, `after-annotation-commit-save`, `after-revision-rename`, `after-current-update`, `after-external-definition-sync`), each asserting the store converges on exactly one annotation and one revision after reopening.
- **End to end against a real 8-output news corpus**: captures once, publishes checklist v1, writes lineage back into the checklist file, recaptures idempotently, and releases the lock even when the terminal check rejects a non-TTY.
- Anti-pattern audit against `docs/dev/anti-patterns.md` found two nested ternaries in the checklist renderer, fixed in the last commit.

## Two things I want to flag rather than bury

**`controller.ts` is 570 lines.** That is larger than this repo's one-concept-per-file norm. It is deliberately the single imperative boundary and splitting it would spread the sign-off ordering across modules, which is the thing the design most wants to prevent — but it is worth a look, and I would rather say so than let it pass unremarked.

**The plan's Task 12 said to stop for approval before pushing.** You asked for a PR in the same message that asked me to execute, so I treated that as the approval. Flagging it so it reads as a deliberate call.

## Not merging

Left open for you to squash-merge.


packages/agency-lang/lib/eval/label/controller.ts
```diff
@@ -0,0 +1,570 @@
...
+    return this.recoverChecklist();
+  }
+
+  private recoverAnnotation(): void {
+    const pending = this.draft.pendingAnnotation;
+    if (pending === null) {
+      return;
+    }
+    this.parts.store.appendAnnotation(pending);
+    this.draft = { ...this.draft, pendingAnnotation: null };
```
egonSchiele: **P1 — recovery does not finish the committed state transition.** This appends the pending row and clears it, but normal sign-off also applies `annotationCommitted`, advances `currentIndex`, persists reviewed state, and resets this output's active timer. On a crash after pending-save/append with two outputs, reopen reconstructs the annotation as reviewed and then overlays the old draft cursor/timing, leaving the user on the already-reviewed output with its old accumulated time. Reuse the normal post-append transition during recovery and assert that reopen advances to item 2 and resets the committed output's timer.

packages/agency-lang/lib/eval/label/checklist.ts
```diff
@@ -0,0 +1,391 @@
...
+  }
+  return ChecklistCurrentSchema.parse(JSON.parse(fs.readFileSync(file, "utf8")));
+}
+
+export function readRevision(storeDir: string, checklistId: string, version: number): ChecklistRevision {
+  const file = revisionPath(storeDir, checklistId, version);
+  if (!fs.existsSync(file)) {
+    throw new Error(`Checklist revision not found: ${file}`);
+  }
+  return ChecklistRevisionSchema.parse(JSON.parse(fs.readFileSync(file, "utf8")));
```
egonSchiele: **P1 — immutable revisions are trusted rather than validated.** Schema parsing does not require the internal `version` to match the filename or recompute `checklistHashOf`; `validateLineages` also accepts gaps such as `1.json, 3.json` when 3 names parent 1 and does not reapply the no-removal/no-text-edit evolution rule. A schema-valid edit can therefore change what old answers mean while annotations still match the file's self-reported hash. Validate path version + recomputed hash on read, require contiguous versions, and validate each adjacent revision with the same legal-evolution checks used for publication.

packages/agency-lang/lib/eval/label/jsonl.ts
```diff
@@ -0,0 +1,185 @@
...
+ * that is neither the old content nor the new. Writing a uniquely named
+ * temporary and renaming means a reader always sees one or the other. The temp
+ * name carries the pid so two processes cannot collide on it.
+ */
+export function atomicWriteValidated<Value>(args: AtomicWriteArgs<Value>): void {
+  const validated = args.schema.parse(args.value);
+  fs.mkdirSync(path.dirname(args.targetPath), { recursive: true });
+  const temporaryPath = `${args.targetPath}.${process.pid}.tmp`;
+
+  fs.writeFileSync(temporaryPath, `${JSON.stringify(validated, null, 2)}\n`, { flag: "w" });
```
egonSchiele: **P1 — these writes are atomic but not durable.** Neither this helper nor `appendExact()` calls `fsync`, despite the plan requiring each promised durable boundary to survive and not swallow fsync errors. A power/OS crash after return can lose the renamed draft/pointer or acknowledged annotation. Fsync the written temp/append fd before proceeding, then fsync the parent directory after rename (and after creation boundaries).

packages/agency-lang/lib/eval/label/controller.ts
```diff
@@ -0,0 +1,570 @@
...
+      lock.release();
+    }
+    throw error;
+  }
+}
+
+/** Write only the identity fields back, leaving the file otherwise as the
+ *  author wrote it. Publication later rewrites it in full. */
+function syncChecklistDefinitionIds(checklistFile: string, definition: NormalizedDefinition): void {
+  fs.writeFileSync(checklistFile, `${JSON.stringify(definition, null, 2)}\n`);
```
egonSchiele: **P1 — first-open identity normalization can destroy the only checklist definition.** This plain write truncates the file before a draft or revision exists. A crash mid-write leaves partial JSON, and the next open parses this file before inspecting the store, so the stated recovery path cannot run. Use the atomic validated write path here as well (with durable file/directory sync), so reopen sees either the original definition or the complete normalized one.

packages/agency-lang/lib/eval/label/capture.ts
```diff
@@ -0,0 +1,272 @@
...
+  }
+  if (outputs.length === 0) {
+    return { kind: "missing" };
+  }
+  const index = outputs.length - 1;
+  const entry = outputs[index] as { value?: unknown; truncated?: unknown };
+  if (entry?.truncated === true) {
+    return { kind: "truncated", index };
+  }
+  const parsed = JsonValueSchema.safeParse(entry?.value ?? null);
```
egonSchiele: **P1 — a missing output value becomes a real `null` output.** `{}`, `{ threadId: ... }`, and `null` final entries all reach `entry?.value ?? null`, so they are captured and can receive durable labels indistinguishable from an agent that genuinely returned JSON null. Require a non-null object with an own `value` property; pass an explicit `{ value: null }` through, but classify absent `value` as missing.

packages/agency-lang/lib/eval/label/labelTui.ts
```diff
@@ -0,0 +1,434 @@
...
+  }[status];
+  const score = snapshot.scores[item.outputId];
+  const scoreText = score === null || score === undefined
+    ? `${GREY}—${RESET}`
+    : `${scoreColour(score)}${BOLD}${score.toFixed(2)}${RESET}`;
+  out.push(
+    ` ${BOLD}${CYAN}${item.outputId.slice(0, 12)}${RESET} ` +
+    `${GREY}${snapshot.itemIndex + 1}/${snapshot.items.length}${RESET}  ${chip}  ${scoreText}`,
+  );
+  out.push(` ${DIM}${item.task.slice(0, args.columns - 4)}${RESET}`);
```
egonSchiele: **P1 — untrusted labeling content can execute terminal controls.** The task is written directly here, and model output/question/note text is likewise rendered without neutralizing arbitrary ESC/CSI/OSC/C0/C1 controls. A model output containing cursor movement, clear-screen, or OSC 52 can hide/forge the evidence being judged or manipulate the terminal. Sanitize untrusted source strings before adding the TUI's trusted ANSI styling; preserve only explicitly supported newlines/tabs, and add CSI/OSC/multiline-task tests.

packages/agency-lang/lib/eval/label/labelTui.ts
```diff
@@ -0,0 +1,434 @@
...
+  out.push("");
+
+  const checklist = renderChecklist(snapshot, rightWidth);
+  const visibleBody = args.body.slice(args.scroll, args.scroll + bodyHeight);
+  for (let row = 0; row < bodyHeight; row += 1) {
+    out.push(`${pad(visibleBody[row] ?? "", leftWidth)}${GREY}│${RESET} ${checklist[row] ?? ""}`);
```
egonSchiele: **P1 — focused questions below the pane are invisible.** The reducer advances `questionIndex`, but this always renders checklist lines starting at zero. With more/wrapped questions than `bodyHeight`, Space/Down changes hidden questions and Enter can sign off while the active checkbox is off-screen. Track question start lines and choose a right-pane viewport that keeps `questionIndex` visible (or add independent checklist scrolling).

packages/agency-lang/lib/eval/label/labelTui.ts
```diff
@@ -0,0 +1,434 @@
...
+
+export type Key =
+  | { kind: "up" } | { kind: "down" } | { kind: "left" } | { kind: "right" }
+  | { kind: "pageUp" } | { kind: "pageDown" }
+  | { kind: "enter" } | { kind: "escape" } | { kind: "backspace" }
+  | { kind: "char"; value: string };
+
+/** Escape sequences arrive as one chunk, so parse the chunk rather than
+ *  inspecting characters in isolation. */
+export function parseKeys(chunk: string): Key[] {
```
egonSchiele: **P2 — stream chunks do not preserve escape-sequence boundaries.** An arrow split as `"\x1b["` then `"A"` is parsed as Escape, `[` and `A`; while editing this cancels the editor, and in normal mode fragments of longer sequences can trigger character commands such as deletion. Keep a buffer across `data` events and retain trailing prefixes until a complete known sequence arrives; test every supported sequence split at each byte boundary.

packages/agency-lang/lib/eval/label/session.ts
```diff
@@ -0,0 +1,365 @@
...
+  outputId: string;
+  coveredQuestionIds: string[];
+  answers: Record<string, boolean>;
+  note: string;
+} | undefined {
+  const item = currentItem(state);
+  if (item === undefined) {
+    return undefined;
+  }
+  const live = liveOf(state);
```
egonSchiele, line 230: **P2 — sign-off remains executable with zero live questions.** The snapshot correctly sets `canSignOff: false`, but this still returns an empty payload and Enter always dispatches sign-off. The controller appends an empty annotation and marks the item reviewed in memory; after reopen, `initSession()` sees no effective answer keys and the item becomes untouched again. Return `undefined` when `live.length === 0` and add a controller reopen test for an all-deleted checklist.

egonSchiele (commented):
This review found several correctness blockers in the crash-recovery and display paths. CI is also currently red: the typecheck fails in `checklist.test.ts` and `controller.test.ts`, and the full test run finds raw NUL bytes in `ids.test.ts`. The inline comments focus on behavioral issues that remain after those mechanical failures are fixed.

packages/agency-lang/lib/eval/label/annotations.ts
```diff
@@ -0,0 +1,139 @@
...
+  for (const row of rows) {
+    if (matchesFoldKey(row, key)) {
+      note = row.note;
+    }
+  }
+  return note;
+}
+
+export function itemStatus(judgement: ItemJudgement): ItemStatus {
+  if (Object.keys(judgement.answers).length === 0) {
```
egonSchiele, line 108: does this mean that if something failed every item in the checklist it would be marked "untouched"?

packages/agency-lang/lib/utils/canonicalize.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (Array.isArray(value)) {
+    return value.map(sortKeys);
+  }
+  if (typeof value === "object") {
+    const sorted: Record<string, unknown> = {};
+    for (const key of Object.keys(value as Record<string, unknown>).sort()) {
+      sorted[key] = sortKeys((value as Record<string, unknown>)[key]);
+    }
+    return sorted;
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed. Null-prototype accumulator in `sortKeys`.

Verified the collision before changing anything: `{"__proto__":{"x":1},"a":2}` and `{"a":2}` canonicalized to the same string, so `contentHashOf` gave them the same hash. Since the input comes from agent output and task JSON, that was attacker-controllable. Tests assert both that the two canonicalize differently and that their content hashes differ.

packages/agency-lang/lib/utils/canonicalize.ts
```diff
@@ -0,0 +1,37 @@
...
+  | { [key: string]: JsonValue };
+
+/**
+ * JSON with object keys sorted at every depth, so two structurally equal
+ * values produce the same string and therefore the same hash.
+ *
+ * Arrays keep their order — order is meaningful there. `undefined` members are
+ * dropped, which is what `JSON.stringify` does anyway; callers that need to
+ * distinguish "absent" from "present and undefined" must not rely on this.
+ */
```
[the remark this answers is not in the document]
egonSchiele: Fixed. The docstring now says an undefined *object property* is dropped and an undefined *array element* becomes `null`, with a test pinning `canonicalize([undefined]) === "[null]"`.

packages/agency-lang/lib/eval/label/store.ts
```diff
@@ -0,0 +1,399 @@
...
+/** Refuse an annotation whose output is not already in the corpus, so the
+ *  capture-before-label ordering cannot be skipped by any caller. */
+function assertAnnotationIsGrounded(
+  storeDir: string,
+  corpus: OpenedJsonl<CorpusRow>,
+  row: AnnotationRow,
+): void {
+  const present = (corpus.rows() as readonly CorpusRow[]).some(
+    (existing) => existing.outputId === row.outputId,
+  );
```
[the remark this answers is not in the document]
egonSchiele: Fixed. The `outputId` index is built once in `openStore` and updated by `captureSource`, so grounding is a single lookup. `assertAnnotationIsGrounded` now takes that index rather than the log, and its doc comment says why it is passed in rather than rebuilt.

packages/agency-lang/lib/eval/label/capture.ts
```diff
@@ -0,0 +1,272 @@
...
+ * "[object Object]" and merges unrelated structured outputs into one
+ * meaningless string.
+ */
+export function selectLabelingFinalOutput(record: unknown): FinalOutputSelection {
+  const outputs = (record as { evalOutputs?: unknown }).evalOutputs;
+  if (!Array.isArray(outputs)) {
+    // The legacy single-field shape cannot say which output was final, and
+    // inventing a nullable index would weaken every identity derived from it.
+    return "finalResponse" in (record as object) ? { kind: "legacy" } : { kind: "missing" };
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed. Confirmed `"finalResponse" in "a string"` throws `TypeError` — that would have aborted the whole capture rather than skipping one broken input. `selectLabelingFinalOutput` now guards that `record` is a non-null, non-array object before both the `in` and the `evalOutputs` read, with tests for a string, `null` and a number. A non-object entry inside `evalOutputs` is handled the same way.

packages/agency-lang/lib/eval/label/capture.ts
egonSchiele: Fixed. `entry?.value ?? null` did conflate the two. It now requires `Object.hasOwn(entry, "value")`, so `{}` and `{ threadId }` are `missing` while an explicit `{ value: null }` is captured as the JSON null the agent actually returned. Tests cover all three.

packages/agency-lang/lib/eval/label/controller.ts
egonSchiele: Fixed. `recoverAnnotation` now applies the same post-append transition the live path does: reviewed state, answers, note, cursor advance, and the timer reset for the committed output.

Added the assertion you described — crash at `after-pending-annotation-save` with two outputs, reopen, and check `itemIndex` is 1 and the signed-off output's `activeMsByOutputId` is 0.

packages/agency-lang/lib/eval/label/checklist.ts
egonSchiele: Fixed, all three parts.

`readRevision` now proves the file's recorded `version` and `checklistId` match the path it was read from, and recomputes `checklistHashOf` against its own questions — so an in-place edit can no longer keep matching the hash annotations recorded.

`validateLineageContinuity` replaces the old parent-chain walk. You were right that `1.json, 3.json` passed: each file's recorded parent still lined up with the previous file *present*. It now requires versions to be contiguous from 1, and re-applies `assertQuestionsEvolveLegally` to every adjacent pair, so a text edit or removal introduced between revisions is caught on read.

Four tests: gap in the chain, in-place edit, text edit between revisions, broken parent. One existing test had to be fixed too — it hand-wrote a revision 2 carrying revision 1's hash, which the new check correctly rejects.

packages/agency-lang/lib/eval/label/jsonl.ts
egonSchiele: Fixed. Appends and temp writes now `fsync` the file descriptor before returning, and the parent directory is synced after a rename.

Directory-sync failures are swallowed deliberately, with a comment saying so: some platforms and filesystems refuse to open a directory for sync, and the rename has already happened, so failing the write there would be worse than the gap it closes. File-level fsync failures are not swallowed.

packages/agency-lang/lib/eval/label/controller.ts
egonSchiele: Fixed. Uses `atomicWriteValidated` with the `ChecklistDefinition` schema.

Your reasoning is the part I had missed: the next open parses this file *before* it inspects the store, so a truncated half-write is outside the recovery path entirely — and it would have destroyed the only copy of the questions the author wrote. The comment on the function now says that rather than just "write it back".

packages/agency-lang/lib/eval/label/labelTui.ts
egonSchiele: Fixed. Added `sanitizeUntrusted`, applied to the task, the output body, question text, notes, and editor drafts — everything the tool did not author.

It keeps only newline and tab (the layout uses both) and replaces C0, C1 and DEL with a visible marker, so the reader can see something was removed rather than it silently vanishing. The tool's own styling is applied afterwards, so colour is unaffected. The task line is also truncated to its first line, since a multi-line task could otherwise push the layout apart.

Tests in a new `labelTui.untrusted.test.ts`: clear-screen, OSC 52, a C1 byte, a bare carriage return, and a multi-line task — asserted at the frame level as well as on the function.

Worth noting the editor draft needed it too: `beginNote` seeds the draft from a stored note, so it is untrusted after a round trip even though typing cannot produce an escape.

packages/agency-lang/lib/eval/label/labelTui.ts
egonSchiele: Fixed. `renderChecklist` now reports the line its focused question starts on, and `checklistOffset` chooses a right-pane window that keeps that line visible.

Tests: 40 questions in a 20-row terminal with focus on the last one asserts question 39 is on screen; focus at the top still shows question 0; and a checklist that already fits is not scrolled.

packages/agency-lang/lib/eval/label/labelTui.ts
egonSchiele: Fixed. `parseKeysBuffered` returns any trailing partial sequence, and the input loop keeps it across `data` events.

Tested at every byte boundary of every supported sequence — the loop splits each of the six sequences at each internal position and asserts the pair of chunks yields exactly one key. Also covers a lone `ESC` and a partial `ESC[5` waiting for its terminator.

`parseKeys` is kept for callers with a complete chunk, and both now share one `parseOneKey`.

packages/agency-lang/lib/eval/label/session.ts
egonSchiele: Fixed. `signOffPayload` returns `undefined` when no question is live, so the controller writes nothing.

Your description of the failure was exact: it marked the item reviewed in memory, then read as untouched after a reopen because the fold found no answer keys — a sign-off that silently undid itself. Added the controller reopen test for an all-deleted checklist: `canSignOff` is false, dispatching `signOff` appends no annotation, and reopening reports 0 reviewed.

packages/agency-lang/lib/eval/label/annotations.ts
egonSchiele: No — it is reviewed with score 0.

Verified rather than reasoned: `effectiveAnswers` records an explicit entry for every covered question including the `false` ones, so `answers` is `{q_a: false, q_b: false}` and the `Object.keys(...).length === 0` guard does not fire. `untouched` means nobody signed the item off at all.

Added a regression test asserting exactly that, since the distinction is easy to invert by accident: a `false` is a judgement, an absence is not.

PR #794, opened 2026-08-03
refactor(eval): rebuild the labeling TUI on the existing terminal library

egonSchiele:
## Why

The labeling TUI hand-rolled a terminal loop, key parsing, scrolling, layout and text wrapping. `lib/tui` and `std::ui/layout` already provide all of it, and `lib/cli/logsView.ts` — the closest existing feature — already uses them. **Net 383 lines deleted from `labelTui.ts`.**

Gone, replaced by the library: `parseKeys`, `parseKeysBuffered`, `parseOneKey`, `isIncompleteSequence`, `enterTerminalMode`, `runInputLoop`, `wrapAnsi`, `visibleLength`, `checklistOffset`, and `scrollDelta`'s clamping.

| Now used | Replaced |
|---|---|
| `Screen.runLoop` | hand-rolled promise + `data` listener |
| `TerminalInput` | raw mode, key parsing, escape reassembly |
| `followCursor` | `checklistOffset` — same rule, exactly |
| `wrapText` / `visualWidth` | `wrapAnsi` / `visibleLength` |
| `ScriptedInput` + `FrameRecorder` | hand-built fake streams |

`TerminalInput` already handles escape sequences split across stream chunks, which is the P2 a reviewer found in my parser — a bug I introduced by not using the existing input layer. `wrapText` is also better than what it replaces: `reinjectSgr` re-opens the active style at each wrapped segment, where mine reset and lost it.

## Shared fixes, where the bug actually lived

**`std::ui/layout` did not know OSC hyperlinks are zero-width.** `visualWidth`, `stripAnsi` and `breakLongToken` handled CSI only, so a link counted its whole URL toward the column budget. Any caller rendering terminal Markdown had this; it surfaced here because the corpus is link-dense.

**The TUI style parser had the same gap**, one layer up: it matched `{tag}` and SGR only, so OSC sequences became literal cells. A link printed its own href across the pane. It now consumes and drops them. Before/after on the real corpus:

```
before:  featuring Medić—launch on August 2. (8;;https://www.tomsg…
after:   featuring Medić—launch on August 2. (tomsguide.com)
```

## Two defects found while building

**A pane cannot be a column of `line()` children.** `scrollOffset` slices an element's *own* content by newline, so a column of one-line children has nothing to slice and the offset is silently ignored — the checklist rendered from question 0 no matter where focus was. Each pane is now a single text element.

**`process.stdout.columns ?? 100` is not enough.** A PTY with no attached window reports **0**, not undefined, so the screen laid out to nothing and rendered a blank frame. Found by driving the real command through a PTY and noticing the terminal entered the alternate screen and drew *nothing*. `terminalDimension` now falls back for anything that is not a positive finite number.

## Lifecycle

The CLI builds the `Screen` and destroys it, alongside the session it must be torn down with, and rejects a non-interactive terminal *before* opening a session — so a piped invocation no longer creates a store and takes a lock it will never use.

## Verification

- **2841 tests** across `lib/eval`, `lib/tui`, `lib/stdlib`, `lib/cli`, `lib/logsViewer`.
- `pnpm run typecheck` (all three configs), `lint:structure`, and `lib/sourceIsText.test.ts` — the two guards that broke the previous PR, run explicitly this time.
- **End to end through a real PTY**: 8 outputs captured, checklist published, space toggles, enter signs off one annotation covering 4 questions with recorded timing, q quits, no lock left behind.

### DIFF
--- a/packages/agency-lang/lib/cli/eval/label.ts
+++ b/packages/agency-lang/lib/cli/eval/label.ts
@@ -86,7 +86,7 @@ const defaultDependencies: EvalLabelDependencies = {
   openSession: openLabelingSession,
   runTui: runLabelTui,
   makeScreen: () => new Screen({
-    input: new TerminalInput(),
+    input: new TerminalInput({ suppressSigint: true }),
     output: new TerminalOutput(),
     width: terminalDimension(process.stdout.columns, DEFAULT_COLUMNS),
     height: terminalDimension(process.stdout.rows, DEFAULT_ROWS),
@@ -142,7 +142,15 @@ export async function evalLabel(
   // the terminal loop.
   const screen = dependencies.makeScreen();
   try {
-    await dependencies.runTui({ controller, screen, storeLabel: path.basename(storeDir) });
+    await dependencies.runTui({
+      controller,
+      screen,
+      storeLabel: path.basename(storeDir),
+      currentSize: () => ({
+        width: terminalDimension(process.stdout.columns, DEFAULT_COLUMNS),
+        height: terminalDimension(process.stdout.rows, DEFAULT_ROWS),
+      }),
+    });
   } finally {
     try {
       screen.destroy();
--- a/packages/agency-lang/lib/cli/eval/label.ts
+++ b/packages/agency-lang/lib/cli/eval/label.ts
@@ -5,6 +5,9 @@ import * as path from "path";
 import type { AgencyConfig } from "@/config.js";
 import { openLabelingSession, type LabelingSessionController } from "@/eval/label/controller.js";
 import { runLabelTui } from "@/eval/label/labelTui.js";
+import { TerminalInput } from "@/tui/input/terminal.js";
+import { TerminalOutput } from "@/tui/output/terminal.js";
+import { Screen } from "@/tui/screen.js";
 import type { Annotator } from "@/eval/label/types.js";
 
 const DEFAULT_STORE_DIRECTORY = "labels";
@@ -23,8 +26,10 @@ export type EvalLabelOptions = {
 export type EvalLabelDependencies = {
   openSession: typeof openLabelingSession;
   runTui: typeof runLabelTui;
-  input: NodeJS.ReadStream;
-  output: NodeJS.WriteStream;
+  /** Built here rather than inside the TUI so the terminal lifecycle has one
+   *  owner, alongside the session it must be torn down with. */
+  makeScreen(): Screen;
+  isInteractive(): boolean;
   environment: NodeJS.ProcessEnv;
   osUserName(): string | undefined;
 };
@@ -62,11 +67,31 @@ export function resolveAnnotator(
   return { kind: "human", id: FALLBACK_ANNOTATOR_ID };
 }
 
+const DEFAULT_COLUMNS = 100;
+const DEFAULT_ROWS = 30;
+
+/**
+ * A usable terminal dimension.
+ *
+ * `??` is not enough: a PTY with no attached window reports **0**, not
+ * undefined, and a zero-width screen lays every element out to nothing and
+ * renders a blank frame. Anything that is not a positive finite number falls
+ * back.
+ */
+export function terminalDimension(value: number | undefined, fallback: number): number {
+  return typeof value === "number" && Number.isFinite(value) && value > 0 ? value : fallback;
+}
+
 const defaultDependencies: EvalLabelDependencies = {
   openSession: openLabelingSession,
   runTui: runLabelTui,
-  input: process.stdin,
-  output: process.stdout,
+  makeScreen: () => new Screen({
+    input: new TerminalInput(),
+    output: new TerminalOutput(),
+    width: terminalDimension(process.stdout.columns, DEFAULT_COLUMNS),
+    height: terminalDimension(process.stdout.rows, DEFAULT_ROWS),
+  }),
+  isInteractive: () => process.stdin.isTTY === true && process.stdout.isTTY === true,
   environment: process.env,
   osUserName: () => {
     try {
@@ -95,25 +120,34 @@ export async function evalLabel(
   if (!fs.existsSync(options.source)) {
     throw new Error(`Source run directory not found: ${options.source}`);
   }
+  if (!dependencies.isInteractive()) {
+    throw new Error(
+      "agency eval label needs an interactive terminal: it shows outputs and reads " +
+      "keystrokes. Run it directly rather than through a pipe.",
+    );
+  }
 
   const config = options.config ?? {};
+  const storeDir = resolveLabelStore(options, config);
   const controller: LabelingSessionController = await dependencies.openSession({
     sourceDir: path.resolve(options.source),
-    storeDir: resolveLabelStore(options, config),
+    storeDir,
     checklistFile: path.resolve(options.checklist),
     annotator: resolveAnnotator(options, dependencies),
     reportWarning: (message) => console.warn(message),
   });
 
-  // The session owns a lock and a draft; closing it is not optional, which is
-  // why the CLI holds the finally rather than the terminal loop.
+  // The session owns a lock and a draft, and the screen owns raw mode; closing
+  // both is not optional, which is why the CLI holds the finallys rather than
+  // the terminal loop.
+  const screen = dependencies.makeScreen();
   try {
-    await dependencies.runTui({
-      controller,
-      input: dependencies.input,
-      output: dependencies.output,
-    });
+    await dependencies.runTui({ controller, screen, storeLabel: path.basename(storeDir) });
   } finally {
-    await controller.close();
+    try {
+      screen.destroy();
+    } finally {
+      await controller.close();
+    }
   }
 }
--- a/packages/agency-lang/lib/eval/label/labelTui.ts
+++ b/packages/agency-lang/lib/eval/label/labelTui.ts
@@ -1,6 +1,7 @@
 import { stripAnsi, visualWidth, wrapText } from "@/stdlib/layout/ansi.js";
 import { syntaxHighlight } from "@/stdlib/syntax.js";
 import { column, line, lines, row } from "@/tui/builders.js";
+import { escapeStyleTags } from "@/tui/styleParser.js";
 import type { Element, Style } from "@/tui/elements.js";
 import type { KeyEvent } from "@/tui/input/types.js";
 import type { Screen } from "@/tui/screen.js";
@@ -28,6 +29,10 @@ export { stripAnsi, visualWidth };
  * them; everything else in C0, C1 and DEL becomes a visible replacement so the
  * reader can see something was there.
  *
+ * Style tags are escaped too. The TUI parser reads `{black-fg}` as markup, so
+ * plain text alone is enough to restyle or hide the very evidence being
+ * judged — control characters are not the only way into a terminal.
+ *
  * Styling is applied afterwards, so the tool's own colour is unaffected.
  */
 export function sanitizeUntrusted(text: string): string {
@@ -43,7 +48,7 @@ export function sanitizeUntrusted(text: string): string {
     const isC1 = code >= 0x80 && code <= 0x9f;
     out += isC0 || isDelete || isC1 ? "�" : character;
   }
-  return out;
+  return escapeStyleTags(out);
 }
 
 /**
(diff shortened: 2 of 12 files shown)


packages/agency-lang/lib/eval/label/labelTui.ts
```diff
@@ -246,281 +102,300 @@ function questionStyle(deleted: boolean, focused: boolean, checked: boolean) {
...
+    return 0;
+  }
+  const page = Math.max(1, paneHeight - 1);
+  const half = Math.max(1, Math.floor(page / 2));
+  if (event.key === "f") return page;
+  if (event.key === "b") return -page;
+  if (event.key === "d") return half;
+  if (event.key === "u") return -half;
   return 0;
 }
```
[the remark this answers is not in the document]
egonSchiele: Fixed. `pageup`/`pagedown` now page the same distance as the ctrl chords, checked at terminal.ts:36-37 for the exact key names.

You are right that it worked before — the old parser matched `\x1b[5~`/`\x1b[6~` directly. Added a test asserting the page key and the ctrl chord produce the *same* delta, so the two cannot drift apart, plus a loop-level test that a page key scrolls without dispatching an action.

packages/agency-lang/lib/eval/label/labelTui.ts
```diff
@@ -246,281 +102,300 @@ function questionStyle(deleted: boolean, focused: boolean, checked: boolean) {
...
-      return { kind: "appendEditorText", text: key.value };
+    if (event.key === "enter") return { kind: "submitEditor" };
+    if (event.key === "escape") return { kind: "cancelEditor" };
+    if (event.key === "backspace") return { kind: "backspaceEditor" };
+    // A stray Ctrl-F while typing must not land in the text.
+    if (event.ctrl === true) return null;
+    if (event.key.length === 1 && event.key >= " ") {
+      return { kind: "appendEditorText", text: event.key };
     }
     return null;
```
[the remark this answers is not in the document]
egonSchiele: Fixed. Confirmed at terminal.ts:295,311 that bracketed paste arrives as one `{ key: "paste", text }` event.

Your reading of why it used to work is the part I had missed: the hand-rolled parser turned a pasted string into a stream of printable characters, so paste worked by accident, and the rewrite made it fail *silently* — worse than the behaviour it replaced. `actionForKey` now appends the whole clipboard in one `appendEditorText`.

On sanitizing: `pastedText` collapses newlines, carriage returns and tabs to spaces, because the draft renders on a single footer row, and drops other control characters. It deliberately does not use the visible replacement marker that agent output gets — a paste is the user own text, so a marker would be noise rather than evidence of tampering.

Tests cover paste in the editor, paste ignored outside it (where keys are commands), an empty paste dispatching nothing, and the control-character handling.

packages/agency-lang/lib/eval/label/labelTui.ts
```diff
@@ -246,281 +126,311 @@ function questionStyle(deleted: boolean, focused: boolean, checked: boolean) {
...
+    if (item === null) {
+      return state;
+    }
+    if (state.body?.outputId === item.outputId && state.body.width === leftWidth) {
+      return state;
     }
-    args.output.write(`\x1b[2J\x1b[H${renderLabelScreen({
-      snapshot, storeLabel: "", columns, rows, scroll, body: bodyCache?.lines ?? [],
-    })}\n`);
+    const rendered = renderMarkdownSafely(sanitizeUntrusted(item.text));
```
egonSchiele, line 417: **P1 — untrusted content can inject TUI style tags.** `sanitizeUntrusted()` removes terminal controls but leaves `{...}` intact, and strings passed into an `Element` are subsequently interpreted by the shared style parser. A model output, checklist question, task, note, or editor draft containing `{black-fg}{black-bg}hidden{/black-bg}{/black-fg}` can therefore hide or restyle evidence. Escape style markup (for example with `escapeStyleTags()`) for every untrusted value before it enters an `Element`; ANSI produced by the trusted Markdown renderer can remain separate.

packages/agency-lang/lib/eval/label/labelTui.ts
```diff
@@ -246,281 +126,311 @@ function questionStyle(deleted: boolean, focused: boolean, checked: boolean) {
...
+  if (event.key === "b") return -page;
+  if (event.key === "d") return half;
+  if (event.key === "u") return -half;
   return 0;
 }
 
-export function isQuitKey(key: Key): boolean {
-  return key.kind === "char" && (key.value === "q" || key.value === "\x03");
+export function isQuitKey(event: KeyEvent): boolean {
+  return event.key === "q" || (event.key === "c" && event.ctrl === true);
```
egonSchiele, line 369: **P1 — Ctrl-C never reaches this quit path with the production terminal input.** `TerminalInput` re-raises `SIGINT` before emitting the `{ key: "c", ctrl: true }` event, and `TerminalOutput` handles that signal by calling `process.exit(130)`. The loop therefore never gets to set `done`, and the CLI's `finally` does not call `controller.close()`. The exit hook releases the lock, but active-time/draft shutdown is skipped, and `TerminalInput.destroy()` does not restore raw mode or disable bracketed paste. Add/use a `TerminalInput` mode for this screen that emits Ctrl-C without re-raising `SIGINT`, so normal loop and CLI cleanup run.

packages/agency-lang/lib/eval/label/labelTui.ts
```diff
@@ -246,281 +126,311 @@ function questionStyle(deleted: boolean, focused: boolean, checked: boolean) {
...
-                scroll = 0;
-              }
-            }
-            scroll = Math.max(0, scroll + scrollDelta(key, args.output.rows ?? 30));
-          }
-          draw();
-        } catch (error) {
-          args.input.off("data", onData);
-          reject(error);
+  await args.screen.runLoop<LoopState>({
```
egonSchiele, line 422: **P2 — this migration reintroduces split escape-sequence handling failures.** The shared `TerminalInput` buffers bracketed paste, but it does not preserve incomplete CSI/SS3 prefixes across `data` chunks. For example, an Up arrow delivered as `"\x1b["` and then `"A"` becomes Escape, `[`, and `A`, which can cancel an editor or trigger unrelated commands. The previous label loop explicitly buffered this case. Please add incomplete escape-sequence buffering to `TerminalInput`, with a short timeout to distinguish a standalone Escape key.

packages/agency-lang/lib/cli/eval/label.ts
```diff
@@ -95,25 +120,34 @@ export async function evalLabel(
...
+  // both is not optional, which is why the CLI holds the finallys rather than
+  // the terminal loop.
+  const screen = dependencies.makeScreen();
   try {
-    await dependencies.runTui({
-      controller,
-      input: dependencies.input,
-      output: dependencies.output,
-    });
+    await dependencies.runTui({ controller, screen, storeLabel: path.basename(storeDir) });
```
egonSchiele: **P2 — the configured store path becomes trusted terminal/TUI content here.** `path.basename(storeDir)` may contain CSI controls or TUI style tags, and the header renders it without sanitizing or escaping markup. A configured directory name can therefore clear/restyle the screen or hide header content. Sanitize terminal controls and escape style tags before passing/rendering the basename.

packages/agency-lang/lib/eval/label/labelTui.ts
```diff
@@ -246,281 +126,311 @@ function questionStyle(deleted: boolean, focused: boolean, checked: boolean) {
...
-      // Highlighting parses the whole document, which is far too slow to redo
-      // on every keystroke.
-      bodyCache = {
-        outputId: item.outputId,
-        width: leftWidth,
-        lines: renderMarkdownSafely(sanitizeUntrusted(item.text)).split("\n")
-          .flatMap((line) => wrapAnsi(line, leftWidth)),
-      };
+export async function runLabelTui(args: RunLabelTuiArgs): Promise<void> {
+  const { width, height } = args.screen.size();
```
egonSchiele: **P2 — terminal dimensions are captured only once for the whole session.** The old renderer read the current terminal rows/columns on each draw; after this migration, resizing leaves pane widths, wrapping, scroll bounds, and the cached body based on the original dimensions. Refresh `screen.size()` during render/transitions and invalidate the wrapped-body cache when width changes, or add SIGWINCH-driven resizing to `Screen`.

packages/agency-lang/lib/stdlib/layout/ansi.ts
```diff
@@ -98,11 +105,11 @@ function breakLongToken(token: string, width: number): string[] {
   let index = 0;
 
   while (index < token.length) {
-    CSI_TOKEN_RE.lastIndex = index;
-    const escape = CSI_TOKEN_RE.exec(token);
+    ZERO_WIDTH_TOKEN_RE.lastIndex = index;
+    const escape = ZERO_WIDTH_TOKEN_RE.exec(token);
```
egonSchiele, line 131: **P2 — wrapping can split an OSC 8 hyperlink into unbalanced output segments.** `breakLongToken()` treats OSC sequences as zero-width tokens, but it does not track hyperlink state. Wrapping `OSC-open + long label + OSC-close` can produce one line containing only the opener and a later line containing the closer, causing borders/padding inserted between them to become clickable and breaking consumers that render lines independently. Track active OSC 8 state, close it at each chunk boundary, and reopen it on the next chunk, analogous to the existing SGR reinjection behavior.

packages/agency-lang/lib/eval/label/labelTui.ts
egonSchiele: Fixed. `sanitizeUntrusted` now ends with `escapeStyleTags`, so the single call every untrusted field already went through also neutralizes `{...}` markup. That covers output body, task, question text, notes, editor drafts, and the store label in one place rather than at each call site.

New tests in `labelTui.untrusted.test.ts` assert on the rendered frame (via `FrameRecorder`) rather than on a string this module built, so they prove the markup survives as literal text all the way to the output target.

packages/agency-lang/lib/eval/label/labelTui.ts
egonSchiele: Confirmed at `lib/tui/input/terminal.ts:324-329`: the handler called `process.kill(process.pid, "SIGINT")` before emitting the key, so the process died before the loop saw it.

Fixed by adding `TerminalInputOptions { suppressSigint?: boolean }`. The CLI opts in, so Ctrl-C reaches the loop and runs the same teardown as `q`. Default behavior is unchanged for every other caller.

packages/agency-lang/lib/eval/label/labelTui.ts
egonSchiele: You are right, and my comment was wrong: `parseAndEmit` buffered only `pasteBuffer` for bracketed paste, never a partial CSI. I claimed the shared layer owned reassembly when it did not. Comment corrected.

Fixed in `TerminalInput` rather than here, so every TUI benefits: an `escapeBuffer` holds a trailing incomplete `\x1b[`/`\x1bO` prefix and prepends it to the next chunk.

One deliberate gap: a lone `ESC` is *not* buffered. Telling the Escape key apart from a sequence introducer needs a timer, and holding every ESC would make Escape feel broken. So a chunk ending exactly on the bare introducer still reads as Escape. That requires a one-byte split at that exact boundary, and it is never silent. It is stated in the code and pinned by a test named `KNOWN GAP: a chunk ending on the bare introducer is read as Escape` so it cannot quietly become a surprise. Happy to add the timeout if you would rather close it.

packages/agency-lang/lib/cli/eval/label.ts
egonSchiele: Fixed. The label goes through `sanitizeUntrusted` before it reaches the header. A configured path is input like any other.

packages/agency-lang/lib/eval/label/labelTui.ts
egonSchiele: Fixed. Added `Screen.resize(width, height)`, and the label screen passes a `currentSize` provider that the loop reads before each draw, so a resize takes effect on the next frame instead of being frozen at startup.

packages/agency-lang/lib/stdlib/layout/ansi.ts
egonSchiele: Fixed. `reinjectSgr` now carries hyperlink state the way it already carried SGR: `updateActiveHyperlink` tracks the open OSC 8, each wrapped line closes with `HYPERLINK_CLOSE`, and the next line reopens it. A link that spans a wrap is balanced on every line.

Covered by `styleParser.hyperlink.test.ts`, including the ST-terminated form and a non-8 OSC (a clipboard write), which is dropped.

PR #796, opened 2026-08-03
fix(budget): actually enforce per-agent spend controls (propagate, resume, typed error)

egonSchiele:
## What

The three agency-lang fixes from `2026-08-03-agency-lang-changes-for-spend-controls.md`, so statelog's per-agent `budget` (set via `withRuntimeConfigOverrides({ budget })`) actually enforces.

## Change 1 — propagate `budget` into the execution context (blocking)

`RuntimeContext` resolved the override budget into `this.budget`, but `createExecutionContext` copies non-serialized fields one by one and **never copied `budget`**. So `execCtx.budget` was `undefined`, `installRootBudget` installed nothing, and the feature was a **silent no-op** for every run. Fixed by copying it next to `maxCallDepth`/`failurePropagation`, with a test for the `RuntimeContext` → `execCtx` handoff my #792 tests missed.

## Change 2 — re-assert the root budget on resume (host-authoritative limit, spend preserved)

The root guard is serialized **with the checkpoint**, so on a stateless served resume its limit arrives from a caller-controllable payload — a crafted resume could raise the ceiling. On resume, `reinstallRootBudget` now **clamps the restored root guard's limit to the host value while keeping the guard and its accumulated spend/elapsed** (per review):

- the ceiling is host-authoritative and un-raisable from a client checkpoint;
- accumulated spend is preserved, so the trusted in-process CLI resume path (`agency run --max-cost` across a pause) **stays cumulative — no regression**; a served client that lowers its own reported spend only hurts itself (the residual stateless gap a stateless resume always has);
- arming is untouched (restored guards are un-armed by `restoreState`), so no timer pop-race is reintroduced.

A dimension the host no longer caps drops its restored guard through `uninstall()`; a dimension the host caps but the checkpoint lacked gets a fresh guard. No-op in IPC.

## Change 3 — typed `budgetExceeded` from the serve adapter (recommended)

A root-budget trip threw and the adapter returned the generic `"Tool execution failed"`. It now detects the `guardTrip` cause (mirroring `budgetExit`) and returns a typed **HTTP 402** `{ success: false, code: "budgetExceeded", dimension, limit, spent, error }`; other errors keep the generic message so no server detail leaks.

## Handoff question

> can agency-lang expose the authoritative accumulated spend for a run?

Yes — the root `CostGuard`'s serialized `spent` (and `StateStack.localCost`), already in the checkpoint. statelog can read it from the checkpoint it stores for the stateful path.

## Verification

`pnpm run typecheck` (source + tests + evals), `pnpm run lint:structure`, `git diff --check` — clean. New/affected tests pass: the `createExecutionContext` budget copy; `reinstallRootBudget` (clamp keeps the instance + host limit, time clamp, fresh-install edge, non-root preserved, host-removed drop, IPC no-op); the serve adapter's 402 vs generic error; `interrupts`/`configOverrides` unchanged-green. (`functionFrame.integration.test.ts` is build-gated, runs in CI.)

### DIFF
--- a/packages/agency-lang/lib/runtime/interrupts.ts
+++ b/packages/agency-lang/lib/runtime/interrupts.ts
@@ -814,9 +814,10 @@ export async function respondToInterrupts(args: {
     () => __initAllRegisteredCallbacks(execCtx),
   );
   execCtx.restoreState(checkpoint);
-  // Re-assert the root budget from the host context, dropping the checkpoint's
-  // serialized root guard (its limit/spend are caller-controllable on a
-  // stateless resume). Mirrors runNode's install; no-op in IPC.
+  // Re-assert the root budget's LIMIT from the host context (the checkpoint's
+  // ceiling is caller-controllable on a stateless resume), while preserving the
+  // guard's accumulated spend so the trusted CLI resume path stays cumulative.
+  // No-op in IPC.
   reinstallRootBudget(execCtx.stateStack, execCtx.budget);
   execCtx.setInterruptResponses(responseMap);
   if (metadata.callbacks) Object.assign(execCtx.callbacks, metadata.callbacks);
--- a/packages/agency-lang/lib/runtime/interrupts.ts
+++ b/packages/agency-lang/lib/runtime/interrupts.ts
@@ -10,6 +10,7 @@ import type {
 } from "./interruptResponse.js";
 import { runInBootstrapFrame } from "./asyncContext.js";
 import { __initAllRegisteredCallbacks } from "./crossModuleInitRegistry.js";
+import { reinstallRootBudget } from "./rootBudget.js";
 import {
   AgencyCancelledError,
   HandlerRecursionError,
@@ -813,6 +814,10 @@ export async function respondToInterrupts(args: {
     () => __initAllRegisteredCallbacks(execCtx),
   );
   execCtx.restoreState(checkpoint);
+  // Re-assert the root budget from the host context, dropping the checkpoint's
+  // serialized root guard (its limit/spend are caller-controllable on a
+  // stateless resume). Mirrors runNode's install; no-op in IPC.
+  reinstallRootBudget(execCtx.stateStack, execCtx.budget);
   execCtx.setInterruptResponses(responseMap);
   if (metadata.callbacks) Object.assign(execCtx.callbacks, metadata.callbacks);
   if (metadata.debugger) execCtx.debuggerState = metadata.debugger;
--- a/packages/agency-lang/lib/runtime/rootBudget.ts
+++ b/packages/agency-lang/lib/runtime/rootBudget.ts
@@ -55,6 +55,31 @@ export function installRootBudget(
   }
 }
 
+/** Re-assert the root budget on a RESUMED exec context. The root guard is
+ *  serialized with the checkpoint, so on a stateless resume its limit and
+ *  accumulated spend arrive from the (client-controllable) checkpoint. Drop any
+ *  restored root guard and re-install from the host-authoritative `contextBudget`
+ *  (the CLI env flag still wins), so the ceiling cannot be raised from the
+ *  client on a resumed leg. Same root-only gating as installRootBudget — a no-op
+ *  in IPC, where the parent owns the budget.
+ *
+ *  This resets accumulated spend for the leg: a stateless resume carries no
+ *  authoritative running total. Preventing per-leg reset needs host-side state
+ *  (server-stored checkpoints or signed, anti-replay envelopes) — out of scope
+ *  here. See docs/dev/checkpointing.md and the spend-controls handoff. */
+export function reinstallRootBudget(
+  stack: StateStack,
+  contextBudget?: { maxCost?: number; maxTimeMs?: number },
+): void {
+  if (isIpcMode()) return;
+  const withoutRoot = stack.guards.filter((g) => !g.isRootBudget);
+  if (withoutRoot.length !== stack.guards.length) {
+    stack.guards = withoutRoot;
+    stack.rebuildAbortSignal();
+  }
+  installRootBudget(stack, contextBudget);
+}
+
 /** FAIL CLOSED on a malformed budget value. The env is an internal
  *  carrier and the CLI validates before setting it, so a non-finite
  *  value here means a hand-set env or a bug — and for a cost-control
--- a/packages/agency-lang/lib/runtime/rootBudget.ts
+++ b/packages/agency-lang/lib/runtime/rootBudget.ts
@@ -3,29 +3,18 @@ import { CostGuard, TimeGuard } from "./guard.js";
 import { isIpcMode } from "./subprocessRunInfo.js";
 import type { StateStack } from "./state/stateStack.js";
 
-/** Install a root cost/time guard from the CLI flags (AGENCY_MAX_COST /
- *  AGENCY_MAX_TIME env vars) or, when a flag is absent for that dimension, the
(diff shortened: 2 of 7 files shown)


packages/agency-lang/lib/runtime/interrupts.ts
```diff
@@ -813,6 +814,10 @@ export async function respondToInterrupts(args: {
     () => __initAllRegisteredCallbacks(execCtx),
   );
   execCtx.restoreState(checkpoint);
+  // Re-assert the root budget from the host context, dropping the checkpoint's
+  // serialized root guard (its limit/spend are caller-controllable on a
+  // stateless resume). Mirrors runNode's install; no-op in IPC.
+  reinstallRootBudget(execCtx.stateStack, execCtx.budget);
```
egonSchiele, line 821: `respondToInterrupts` is the resume path for **both** transports — the stateless served `/resume` (checkpoint round-trips through an untrusted client) *and* in-process CLI resume (checkpoint comes from a local trace/file the runtime itself wrote). This unconditional reinstall drops the restored root guard's accumulated `spent` on both.

The tampering risk that motivates re-asserting the ceiling is specific to the **served/stateless** transport. On the CLI path the checkpoint's `spent` is authoritative and not attacker-controllable, so resetting it is a straight behavior regression: a local `agency run --max-cost 1` that pauses and resumes now gets a fresh full $1 each leg instead of a cumulative $1 across the whole run — with no security benefit on that path. Same for `--max-time` (fresh full window per resume).

Worth deciding explicitly: is this CLI-side reset intended? If not, consider scoping the reassertion to the stateless/served resume, or preserving `spent` while clamping only the limit (see my note on rootBudget.ts). The PR body frames the reset as inherent to stateless resume — but for the in-process path it isn't inherent, it's newly introduced here.

packages/agency-lang/lib/runtime/rootBudget.ts
```diff
@@ -55,6 +55,31 @@ export function installRootBudget(
...
+  stack: StateStack,
+  contextBudget?: { maxCost?: number; maxTimeMs?: number },
+): void {
+  if (isIpcMode()) return;
+  const withoutRoot = stack.guards.filter((g) => !g.isRootBudget);
+  if (withoutRoot.length !== stack.guards.length) {
+    stack.guards = withoutRoot;
+    stack.rebuildAbortSignal();
+  }
+  installRootBudget(stack, contextBudget);
```
egonSchiele: Drop-and-fresh-install means the new root guard starts at `spent = 0`. An alternative that fixes #2's actual threat ("raise the ceiling") **without** discarding accumulated spend: keep the restored root guard and clamp only its limit to the host `contextBudget` (e.g. `min(restoredLimit, hostLimit)`, or just set it to the host value). Then:
- CLI/in-process resume keeps cumulative enforcement (no regression);
- served resume can't raise the ceiling (the limit is host-authoritative), and a client that *lowers* its own reported `spent` only ever hurts itself — no worse than today, and the documented residual stateless-side gap stays exactly the same.

This would let you keep the reassertion unconditional (simpler) while removing the CLI regression from the other comment.

packages/agency-lang/lib/runtime/rootBudget.ts
```diff
@@ -55,6 +55,31 @@ export function installRootBudget(
...
+ *  (server-stored checkpoints or signed, anti-replay envelopes) — out of scope
+ *  here. See docs/dev/checkpointing.md and the spend-controls handoff. */
+export function reinstallRootBudget(
+  stack: StateStack,
+  contextBudget?: { maxCost?: number; maxTimeMs?: number },
+): void {
+  if (isIpcMode()) return;
+  const withoutRoot = stack.guards.filter((g) => !g.isRootBudget);
+  if (withoutRoot.length !== stack.guards.length) {
+    stack.guards = withoutRoot;
```
egonSchiele: Minor / defensive. This drops guards by reassigning `stack.guards` + `rebuildAbortSignal()`, bypassing `guard.uninstall()`. I checked and it's safe **today**: `restoreState` rebuilds the stack via `StateStack.fromJSON`/`deserializeMode` and never arms the restored guards, so a dropped root `TimeGuard` has `controller === undefined` here — nothing to leak.

But it silently depends on that invariant, and diverges from `popGuard`, which calls `uninstall()` precisely so `TimeGuard.uninstall` can `cancelTimer()` "FIRST so a late-fire can't trip the outer scope." If restore ever starts arming guards (or `reinstallRootBudget` gets called from a context where the root guard is live), this reintroduces exactly that pop-race. Cheap hardening: route the drop through `uninstall()`/`popGuard`, or assert the dropped guard is un-armed.

egonSchiele (commented):
Reviewed against the three items in the spend-controls handoff; all three are implemented and I verified the runtime mechanics behind each.

**Change 1 (context.ts) — correct.** `execCtx.budget = this.budget` is exactly the missing copy in `createExecutionContext`; `installRootBudget(execCtx.stateStack, execCtx.budget)` now sees the resolved override, and the new `context.budget.test.ts` pins the RuntimeContext→execCtx handoff that #792 didn't cover. This was the blocking no-op; it's fixed.

**Change 3 (adapter.ts) — correct.** Detecting the trip by `readCause(...).kind === "guardTrip"` (mirroring `budgetExit`) and returning a typed 402 is right, and only *root* trips reach this catch — a user `guard()` trip is converted to a Result at its boundary by `_runGuarded`, so the 402 won't misfire on user guards. Routing all three surfaces (function/node/resume) through the shared `errorResult` is a nice cleanup.

**Two things worth discussing on Change 2**, inline. One is a real behavior question (the reassertion runs on the trusted in-process CLI resume path too, not just the stateless served one); the other is a defensive note on the guard-array surgery. Neither blocks — I confirmed the current code is safe as written.

I also confirmed the things that could have bitten and didn't: `restoreState` rebuilds the stack via `fromJSON`/`deserializeMode` and does not arm guards, so dropping a restored `TimeGuard` without `uninstall()` leaks no timer today; and the `afterEach` already clears `AGENCY_IPC` so the IPC-mode test can't bleed into its neighbors.

packages/agency-lang/lib/runtime/interrupts.ts
egonSchiele: Fixed in 642fcefbf by adopting your clamp suggestion (below): reinstallRootBudget now preserves the restored guard's accumulated spend and only clamps its limit, so the trusted in-process CLI resume stays cumulative — no more reset-per-leg on that path — while a served checkpoint still can't raise the ceiling. So the reset is no longer introduced for the CLI/--max-time path at all.

packages/agency-lang/lib/runtime/rootBudget.ts
egonSchiele: Adopted this in 642fcefbf. reinstallRootBudget keeps the restored root guard and sets its limit to the host value (costLimit/timeLimit), preserving spend/elapsed: CLI/in-process resume keeps cumulative enforcement, served resume can't raise the ceiling, and a client lowering its own spend only hurts itself (residual stateless gap unchanged). Kept it unconditional as you noted. Direct clamp rather than extendBudget since that only grants upward.

packages/agency-lang/lib/runtime/rootBudget.ts
egonSchiele: Addressed in 642fcefbf. The common path now clamps in place and never touches arming, so the pop-race can't be reintroduced there. The only remaining drop (a dimension the host no longer caps) routes through guard.uninstall(stack) before removal, so TimeGuard.uninstall cancels its timer first — no reliance on the restored-guard-is-un-armed invariant.

PR #797, opened 2026-08-03
eval label: field-based records, content identity, and ingest from any source

egonSchiele:
Implements `docs/superpowers/specs/2026-08-03-eval-label-ingest-design.md` (revision 3), following the plan at `docs/superpowers/plans/2026-08-03-eval-label-ingest.md`.

## What this is for

`agency eval label` could only judge outputs from one eval run directory. Training data comes from more than one place: several agents you want to compare, answers you wrote by hand to show what good looks like, outputs pasted from somewhere else. A handwritten answer has no trace id, no input id and no eval record, so it could not get in at all.

## The principle everything follows from

The labelled artifact and the label are the whole training example. An agent learning from this data sees those two things and nothing else — no annotation history, no trace id, no run directory. Three consequences do all the work:

- **Anything a label depends on must be inside the artifact.** "Is this today's news?" is yes in August and no in December only because the artifact failed to say when it was written. Make the artifact say more, not the store remember more.
- **Identity is the artifact's content**, not where it came from. Two runs emitting byte-identical output produced the same training example and should be labelled once.
- **There is no displayed-but-invisible context.** Something is part of the artifact, or it is not there at all.

That last one kills instruction-following questions if a record is one string — you cannot ask "did it answer what was asked" without the request. The established answer is a record with named fields, which is what Argilla, Label Studio and Prodigy all landed on. A code review agent's record has a `diff` field and a `review` field; the label attaches to the pair.

## What changed

**A record is a map of named text fields**, and `outputId` is the hash of that map. One rule, no variants, no exclusions.

**Provenance moved to `occurrences.jsonl`.** Two agents emitting identical text produce one record with two occurrences, so you label once and both get credit — which is also what makes "did v2 beat v1" answerable by counting occurrences per source.

**New commands.**

```bash
agency label ingest runs/v1 --source agent-v1
agency label ingest runs/v2 --source agent-v2
agency label ingest ./gold/ --source handwritten --task "Summarize today's tech news"
agency label --checklist news.json
```

`label` is registered on both the program and the eval command, following the dual registration `optimize` already uses, so `agency eval label …` reaches the same three subcommands.

**Deleted:** `contentHashOf`, `assertSameOccurrence`, `ExecutionIdentity`. All three defended against content changing under a fixed identity, which is unrepresentable once identity *is* the content.

## A live bug this fixes

Field text was sanitized before markdown highlighting, and the highlighter strips backslash escapes — so escaped style tags were handed back to the renderer as live markup. Untrusted output could restyle or hide the evidence being judged. Controls are now stripped first and style tags escaped last, with the ordering explained at the call site. This predates the branch; a new test surfaced it.

## No migration, deliberately

Version 1 of this store shipped one day earlier, in #793 and #794. A search of the machine found no label store at all — nothing outside tests ever created one. A migration was written and then deleted, because it existed to preserve data that does not exist.

A label store is derived data: the eval runs it was built from are still in `runs/`, so `agency label ingest` rebuilds it. What survives is the part that matters — `openStore` reads the manifest before opening any log and refuses a store it does not understand, saying to delete and re-ingest.

The reasoning is recorded in `docs/dev/eval-labeling.md` so it outlives the code, and the deleted implementation is in this branch's history if a future format change ever touches labels that cannot be regenerated.

## Deliberately deferred

**Editing a record.** Revision 2 of the spec let you change a task and carried the labels across. Review was right that this manufactures a verdict nobody gave: if the task changes from "summarize this" to "translate this to French" while the output stays the same, an old "yes, it followed the instruction" becomes actively false. Re-ingest with the framing you want instead. Doing it properly needs a supersession relation, a rule for which record a later session shows, and session rebinding — none of which ingest requires.

## One deviation to flag

**Two logs deliberately bypass `appendExact`.** `occurrences.jsonl` and `outputs.jsonl` each hold a timestamp that is not part of their identity, so a re-ingest would build the same id with a new timestamp and be rejected as corruption. `ensureOccurrence` and `ensureRecord` return the existing row instead. Both are named and commented so this does not read as an oversight.

## The rule nothing enforces

Every checklist question must be answerable from the record's fields alone. The tooling cannot check it — it cannot tell that "today" is unanchored or that "LGTM" refers to an absent diff. A question leaning on absent context produces confidently wrong labels silently. Accepted deliberately and written into `docs/dev/eval-labeling.md`; the alternative is a type system for prose.

## Verification

- 483 tests pass across `lib/eval/label` and `lib/cli/eval`
- `pnpm run typecheck` (all three configs) and `pnpm run lint:structure` clean
- `lib/cli/eval/grade.test.ts` and `optimize.test.ts` fail locally in this worktree for want of a build; verified identical failures with the branch stashed, so they are environment, not this change


packages/agency-lang/lib/cli/eval/ingest.ts
```diff
@@ -0,0 +1,218 @@
...
+    throw new IngestSourceError(
+      "--source is required: it names this batch on every occurrence, and is how you tell " +
+      "one agent's outputs from another's when you read the labels back.",
+    );
+  }
+
+  const requestedFormat = parseFormat(options.format ?? "auto");
+  const includeTaskField = options.taskField !== false;
+  const constantFields = parseFieldArgs(options);
+  assertNoLoaderCollision(requestedFormat, constantFields, includeTaskField);
```
egonSchiele: **Blocking:** collision checking happens before `auto` is resolved, and `LOADER_FIELD_NAMES.auto` is empty. With the default format, `--field output=constant` passes this check and is then silently overwritten by the files/JSON loader; an auto-detected run similarly overwrites `task`. That changes the record content and ID from what the CLI arguments state. Please validate constants against the resolved format inside `loadBatch` (or return the resolved format before this check), and add auto-format collision tests.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
```diff
@@ -0,0 +1,172 @@
...
+      "try again; refusing to delete a directory this command does not recognise.",
+    );
+  }
+
+  for (const name of STAGED_FILES) {
+    fs.rmSync(path.join(stagingDir, name), { force: true });
+  }
+  for (const name of STAGED_DIRECTORIES) {
+    // Recursive only within a subdirectory this migration copied in whole.
+    fs.rmSync(path.join(stagingDir, name), { recursive: true, force: true });
```
egonSchiele: **Blocking data-loss risk:** the comment says unexpected entries survive, but this recursively deletes the entire `checklists` subtree after validating only the top-level marker. If anything creates or copies an unexpected file below a legitimate leftover staging directory, retry deletes it silently. Please inventory the checklist paths owned by the migration and remove only those exact entries (then remove directories only when empty), or quarantine the staging directory instead of recursively deleting it.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
```diff
@@ -0,0 +1,172 @@
...
+    try {
+      openStore({
+        storeDir: stagingDir,
+        lock: verifyLock,
+        reportWarning: args.reportWarning ?? (() => {}),
+      }).close();
+    } finally {
+      verifyLock.release();
+    }
+    fs.rmSync(markerPath(stagingDir), { force: true });
```
egonSchiele: **Recovery bug:** the ownership marker is removed before the publishing rename. A crash in this window, or a `renameSync` failure, leaves a complete `.migrating` directory that the next attempt refuses to reclaim because its marker is gone. Keep the marker through the rename, remove it from `destDir` afterward, and sync the destination/parent directory so every interruption leaves either a reclaimable stage or a published store.

packages/agency-lang/lib/eval/label/migrate/readV1.ts
```diff
@@ -0,0 +1,139 @@
...
+    );
+  }
+
+  const corpusFile = path.join(storeDir, "outputs.jsonl");
+  const annotationsFile = path.join(storeDir, "labels.jsonl");
+  const draftsDir = path.join(storeDir, "drafts");
+
+  return {
+    storeDir,
+    corpus: parseAll(corpusFile, V1CorpusRowSchema, readJsonlLines(corpusFile)),
```
egonSchiele: **Data-integrity blocker:** this reader validates only row shape. It neither rejects duplicate persisted `outputId`s nor recomputes the v1 execution-derived ID. `planMigration` later assigns `idMap[row.outputId]` repeatedly, so two malformed rows sharing an old ID can overwrite the mapping and silently move every annotation for that ID onto the later record. Please validate uniqueness and recompute each v1 `outputId` from its execution identity before planning (and ideally verify `contentHash` too).

packages/agency-lang/lib/eval/label/load/discoverFiles.ts
```diff
@@ -0,0 +1,181 @@
...
+      `No files matched ${source}. Remember to quote the pattern so the shell passes it ` +
+      "through unexpanded.",
+    );
+  }
+  return { root, files };
+}
+
+/** Split "answers/**\/*.txt" into the deepest literal directory and the rest. */
+function splitPattern(source: string): { root: string; pattern: string } {
+  const segments = source.split("/");
```
egonSchiele: This parser breaks quoted globs on Windows because it only splits `/`. A normal `C:\answers\*.txt` remains one segment and gets resolved relative to the current directory, so it matches nothing. Please normalize platform separators while preserving drive/UNC roots, or use an existing glob/path parser, and add drive-letter plus UNC coverage.

egonSchiele (commented):
Review findings are inline.

egonSchiele:
**Blocking CI failure — `packages/agency-lang/lib/eval/label/migrate/plan.ts`:** this source file contains raw NUL bytes at lines 70 and 100, so `lib/sourceIsText.test.ts` fails in the Node 22 build. Replace the embedded bytes with escaped source text (for example `\0`) while retaining the unambiguous key delimiter.

packages/agency-lang/lib/cli/eval/ingest.ts
```diff
@@ -0,0 +1,218 @@
...
+      throw new IngestSourceError(
+        `--field must be written name=value; got "${raw}".`,
+      );
+    }
+    const name = raw.slice(0, separator);
+    if (Object.hasOwn(fields, name)) {
+      throw new IngestSourceError(`--field ${name}= was given twice.`);
+    }
+    fields[name] = raw.slice(separator + 1);
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed. `parseFieldArgs` now accumulates into `Object.create(null)` and validates each name against the store's `FieldNameSchema` before assigning, so `--field __proto__=x` is rejected with a message naming the flag rather than setting a prototype and failing confusingly deep inside ingest.

Worth noting why the duplicate check could not have caught it: on a normal object, assigning `__proto__` creates no own property, so `Object.hasOwn` returns false and the key simply vanishes. Same shape as the canonicalize fix in #793.

Three tests added, including one asserting the accumulator's prototype is null.

packages/agency-lang/lib/eval/label/store.ts
```diff
@@ -199,33 +266,117 @@ export function openStore(args: OpenStoreArgs): LabelStore {
...
+  }
+  if (parsed.schemaVersion === CURRENT_STORE_VERSION) {
+    return;
+  }
+  throw new StoreVersionError(
+    `${storeDir} uses label store format ${String(parsed.schemaVersion)}; this build writes ` +
+    `format ${CURRENT_STORE_VERSION}. Migrate it with:\n\n` +
+    `  agency eval label-migrate ${storeDir} <new-store>\n\n` +
+    "The original store is left untouched.",
+  );
```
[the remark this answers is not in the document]
egonSchiele: Fixed. Older and newer now get different messages: older points at `agency eval label-migrate`, newer says to upgrade Agency and states that migration only moves a store forwards, so it cannot help.

Test asserts the newer case does *not* mention label-migrate, so the dead end cannot come back.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
```diff
@@ -0,0 +1,172 @@
...
+    fs.renameSync(stagingDir, destDir);
```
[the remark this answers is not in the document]
egonSchiele: Fixed, exactly as suggested: rename first, then remove the marker from the destination, then `syncDirectory` on the parent.

Pinned by a test that faults before publish and asserts the marker is still in the staging directory, then that a second run succeeds. Without that ordering the second run refuses, which is the worst outcome — a complete store the tool will not touch.

packages/agency-lang/lib/eval/label/controller.ts
```diff
@@ -130,13 +135,18 @@ async function openSession(
...
     }
-    if (capture.rows.length === 0) {
+
+    const corpus = store.corpusSnapshot();
+    if (corpus.length === 0) {
       throw new Error(
-        `No labellable outputs in ${args.sourceDir}. Every input was skipped; see the warnings above.`,
+        "There is nothing to label: the store holds no records. Add some with " +
+        "`agency eval label ingest <source> --source <name>`.",
       );
```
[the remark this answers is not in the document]
egonSchiele: You are right, and the mistake was wider than this line: the PR description said `agency eval label ingest` too.

Rather than fix the message to match, I renamed the command for symmetry with the migration one. Both are now `agency eval label-ingest` and `agency eval label-migrate` — both name the store they act on, and they sort together in help. The spec said `agency eval label ingest`, but `label` takes an optional positional argument, so making it a parent command with subcommands fights commander for no benefit.

Message, dev doc and PR description all updated.

packages/agency-lang/lib/eval/label/occurrences.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Append this occurrence, or return the one already recorded.
+   *
+   * This log deliberately does NOT use `appendExact`, and that is not an
+   * oversight. `occurrenceId` excludes `firstObservedAt`, so re-ingesting the
+   * same source tomorrow builds the same id with a different timestamp;
+   * `appendExact` would see identity reuse with different content and raise a
+   * corruption error on a completely legitimate operation. The stored
+   * timestamp is the truth, and a re-ingest must neither overwrite it nor fail
+   * against it.
+   */
```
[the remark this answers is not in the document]
egonSchiele: Correct, the comment overstated it. Reworded: replay goes through `find`, NOT through `appendExact`, and `appendExact` is still what writes a row the first time.

Applied the same correction to `ensureRecord` in corpus.ts, which had copied the wording.

packages/agency-lang/lib/eval/label/occurrences.test.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+describe("ensureOccurrence", () => {
+  it("appends an occurrence the store has not seen", () => {
+    expect(openOccurrenceLog(storeDir).ensureOccurrence(candidate()).added).toBe(true);
+  });
+
+  it("returns the existing row on re-ingest rather than throwing", () => {
+    // The whole reason this log does not use appendExact: a second session
+    // builds the same id with a different timestamp.
```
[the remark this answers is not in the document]
egonSchiele: Fixed alongside the implementation comment.

packages/agency-lang/lib/cli/eval/ingest.ts
egonSchiele: Confirmed and fixed. I checked the overwrite direction before changing anything: `loadFiles` and `loadJsonArray` both build `{ ...constantFields, output: text }`, so the loader wins and the constant vanishes without a word — the stored record and its id end up different from what the arguments said.

The check moved into `loadBatch`, immediately after `resolveFormat`, which is the only place the real format is known. `resolveFormat` now returns `Exclude<Format, "auto">`, so the lookup table has no `auto` row to be empty in the first place — the class of bug is gone rather than patched.

This also puts it on the right side of the ownership boundary: the CLI no longer imports `resolveFormat`.

Six tests in `load/index.test.ts` cover auto-detected files, run and json sources, the legal `--no-task-field --task` combination, an unrelated constant, and that the message names the *resolved* format rather than the requested one.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
egonSchiele: You are right, and my comment directly contradicted the code — it claimed unexpected entries survive while the line below recursively deleted the whole subtree.

Fixed by taking the first of your two options. Checklists are a copy of the source tree, so the source *is* the inventory of what migration put there. `removeCopiedChecklists` walks the source and deletes only paths present in both, removing directories last and only when empty. Anything else keeps its whole parent chain alive, and the closing `rmdir` on the staging root then fails and reports rather than taking it with us.

Test writes an intruder file into a leftover stage's `checklists/`, asserts the retry refuses, and asserts the file is still there with its contents intact.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
egonSchiele: Fixed as described: rename first, then remove the marker from `destDir`, then `syncDirectory` on the parent so the rename is durable.

Test faults before publish, asserts `.migration.json` is still in the staging directory, then asserts a second run completes. Before the change that second run refused — which is the worst failure mode here, since it leaves a complete store the tool will not touch.

packages/agency-lang/lib/eval/label/migrate/readV1.ts
egonSchiele: Confirmed and fixed. Traced it: `planMigration` does `idMap[row.outputId] = group.outputId` per row, so two rows sharing an old id leave the later one in the map and every annotation for that id silently moves to the wrong record.

`readV1Store` now rejects both cases before planning:

- duplicate persisted `outputId`, naming both line numbers
- an `outputId` that is not the hash of its own `execution`, which catches a hand-edited row that would map annotations onto content nobody judged

That required reproducing the version 1 id formula in `readV1.ts`. It is commented as to why it lives there: nothing current should compute one, but migration has to check the ids on disk really are what that formula produced.

I did NOT add the `contentHash` check you mentioned as optional. Version 1's content hash folded in the task, and its formula is gone; reproducing a second dead formula to verify a column that had exactly one reader felt like more surface than it buys. Say the word if you would rather have it.

The fixtures had to change too — they used hand-written ids like `out_aaa…`, which the new check correctly rejects. They now derive real ids from the execution identity, which is a better fixture anyway.

packages/agency-lang/lib/eval/label/load/discoverFiles.ts
egonSchiele: Fixed. `splitPattern` now normalizes separators to `/` first, but only when the platform separator is a backslash — a backslash is a legal character in a POSIX filename, so normalizing unconditionally would corrupt a pattern that deliberately contains one.

Joining the literal prefix with `/` keeps both a drive letter (`C:/answers`) and a UNC prefix (`//server/share`) intact, and Node accepts forward slashes on Windows, so `path.resolve` produces the right absolute root either way.

The separator is a parameter defaulting to `path.sep`, so the Windows behaviour is testable from any machine. Six cases: drive letter, UNC, mixed separators (which shells do produce), `**` across backslashes, a POSIX path containing a literal backslash left alone, and the normalizer itself.

I used the existing hand-rolled parser rather than pulling in a glob library — the supported syntax is deliberately small and documented, and adding a dependency for this is more than the problem needs.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
```diff
@@ -0,0 +1,211 @@
...
+  if (!fs.existsSync(stagedChecklists) || !fs.existsSync(sourceChecklists)) {
+    return;
+  }
+  const walk = (relative: string): void => {
+    const from = path.join(sourceChecklists, relative);
+    const to = path.join(stagedChecklists, relative);
+    if (!fs.existsSync(from) || !fs.existsSync(to)) {
+      return;
+    }
+    if (fs.statSync(from).isDirectory()) {
```
egonSchiele: **Critical cleanup escape:** `statSync(from)` follows source checklist symlinks, while `cpSync(..., { recursive: true })` preserves them by default. If `checklists/shared` points at an external directory, the staged copy is also a symlink; recursion then reaches `staged/checklists/shared/file` and `rmSync(to)` deletes the file through that symlink, outside staging (potentially from the source itself). Please `lstat` both sides, reject a symlinked staging root, and never descend through a symlink. Either reject source checklist symlinks before copying or dereference them into a self-contained snapshot. Add a retry test proving an external symlink target is untouched.

egonSchiele (commented):
Rereview findings are inline. The previous collision and v1 identity fixes look correct.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  for (const name of STAGED_FILES) {
+    fs.rmSync(path.join(stagingDir, name), { force: true });
+  }
+  // Checklists are a copy of the source tree, so the source is the inventory of
+  // what this migration put there. Removing by that list rather than recursing
+  // means a file that appeared underneath — from any other writer — survives
+  // and makes the closing rmdir fail.
+  removeCopiedChecklists(sourceDir, path.join(stagingDir, "checklists"));
```
egonSchiele: The live source tree is not a reliable ownership inventory across separate attempts. The source lock does not cover the time between a crash and retry: a removed/renamed source checklist leaves its old staged copy permanently unreclaimable, while a newly added source path can cause an unrelated staged entry at the same relative path to be deleted as supposedly owned. Please persist the exact relative-path/type inventory in `.migration.json` before the initial copy and reclaim only from that immutable inventory.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
```diff
@@ -0,0 +1,211 @@
...
+        reportWarning: args.reportWarning ?? (() => {}),
+      }).close();
+    } finally {
+      verifyLock.release();
+    }
+    // Publish FIRST, then drop the marker. Removing it before the rename opens
+    // a window where a crash, or a failing rename, leaves a complete staging
+    // directory that the next attempt cannot recognise and so refuses to
+    // reclaim. A marker briefly present in the published store is harmless.
+    fs.renameSync(stagingDir, destDir);
```
egonSchiele: **Publication is still not power-loss safe:** the marker was written with plain `writeFileSync` and never fsynced, copied checklist files/directories are not fsynced, and here the marker is removed before the parent rename is made durable. A reboot can therefore leave an unrecognizable stage, publish a store missing checklist data, or restore the staging name after its marker was unlinked. Please fsync the marker/staging directory, make copied checklists durable before the manifest, rename, sync the parent while the marker still exists, then remove the marker and sync `destDir`. Also recognize an existing destination with a matching marker that opens successfully as a completed interrupted publication.

packages/agency-lang/lib/eval/label/load/discoverFiles.ts
```diff
@@ -0,0 +1,207 @@
...
+  const literal: string[] = [];
+  let index = 0;
+  while (index < segments.length && !GLOB_CHARACTERS.test(segments[index])) {
+    literal.push(segments[index]);
+    index += 1;
+  }
+  // Joining with "/" keeps a drive letter ("C:/answers") and a UNC prefix
+  // ("//server/share") intact; Node accepts forward slashes on Windows, and
+  // `path.resolve` turns either into a proper absolute root.
+  const rootPart = literal.join("/");
```
egonSchiele: Root-level globs still lose their root. For `/*.txt`, `literal` is `[""]`, so `rootPart` becomes empty and line 110 resolves `.` instead of `/`. On Windows, `C:\*.txt` becomes `C:/*.txt`, and `path.resolve("C:")` means that drive's current directory rather than `C:\`. If matching files exist there, ingest silently reads the wrong batch. Preserve the parsed root separately from literal segments, and add exact-root tests for POSIX `/*.txt`, `C:\*.txt`, and a UNC root using `path.win32`.

egonSchiele:
**The previous raw-NUL blocker remains unresolved at `ff8071ef`.** Both current Node builds still fail `lib/sourceIsText.test.ts` because `lib/eval/label/migrate/plan.ts` contains raw NUL bytes at lines 70 and 100. This still needs to be replaced with escaped source text before the PR can pass CI.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
egonSchiele: Confirmed by reproducing it before touching anything — a scratch tree with `checklists/shared -> external/`, copied with `cpSync`, then run through the old cleanup walk. The file in `external/` was gone. This is the most serious bug in the PR: cleanup deleting outside the staging directory.

Fixed both ways you suggested, because each covers a different moment:

- **At inventory time**, source checklists are walked with `lstat` and a symlink is *refused* with a message explaining why. I chose refusing over dereferencing: silently turning a link into a copy changes what the migrated store contains, and a store that quietly differs from its source is worse than one that will not migrate until you look at it.
- **At removal time**, nothing is recursive. The walk reads the marker's inventory, `lstat`s each entry, and unlinks a link *as a link*. That last part needed `unlinkSync` rather than `rmSync`: `rmSync` resolves a link to decide what it is removing and then refuses a directory, which the new test caught immediately.

Also added the guard for a staging root that is itself a symlink.

Two tests: one asserts an external file still contains `DO NOT DELETE` after a retry over a symlinked stage, and one asserts a symlinked *source* checklist is refused before anything is copied.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
egonSchiele: You are right, and it is a sharper point than the symlink one because it has no visible symptom — the retry just refuses forever.

The marker now carries the full relative-path/type inventory, written and fsynced **before** the first copy. Reclaim reads only that. The live source is never consulted.

The file moved to its own module, `migrate/staging.ts`, since "what does this directory contain and who owns it" turned out to be a concept rather than a helper.

Test stages a `checklists/cl_deleted` tree that the source does not have at all and asserts the retry both succeeds and removes it — the exact case that used to be unreclaimable.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
egonSchiele: Fixed, in the order you laid out: marker fsynced before the copy, each copied checklist file flushed individually, directories synced deepest-first, manifest last, rename, `syncDirectory` on the parent **while the marker still exists**, then the marker removed and `destDir` synced.

The interrupted-publication case is now recognised and finished rather than refused.

One extra condition I added, and I would not have found it without writing the test: **the destination must also contain a `manifest.json`.** Opening alone is not enough, because `openStore` *creates* a manifest for an empty directory — so a bare folder carrying a stray marker would have opened cleanly and been declared a completed migration, silently skipping the real work and reporting zero records. The manifest is written last, so its presence is the actual signal. My first attempt had this hole and the test failed on it.

The CLI now says "already held a completed migration, nothing was rewritten" for that case rather than printing counts of zero as though it had migrated an empty store.

packages/agency-lang/lib/eval/label/load/discoverFiles.ts
egonSchiele: Confirmed both. `splitPattern("/*.txt")` returned the working directory, and `path.resolve("C:")` returned a path under the cwd rather than the drive root.

The root prefix is now parsed separately from the literal segments — POSIX `/`, drive `C:/`, and UNC `//server/share/` — and resolution goes through `path.win32` or `path.posix` chosen from the separator, so a Windows pattern is never resolved by POSIX rules.

Eight tests, including the three exact-root cases you asked for and their with-a-subdirectory variants, plus one pinning `rootPrefixOf` directly.

packages/agency-lang/lib/eval/label/migrate/staging.ts
```diff
@@ -0,0 +1,242 @@
...
+    );
+  }
+
+  for (const name of ALWAYS_WRITTEN) {
+    removeLeaf(path.join(stagingDir, name));
+  }
+  // Deepest first: a directory can only go once its own entries have.
+  const ordered = [...marker.entries].sort((left, right) => right.path.length - left.path.length);
+  for (const entry of ordered) {
+    const target = path.join(stagingDir, entry.path);
```
egonSchiele: **Critical: cleanup can still escape staging.** `readMarker` casts arbitrary JSON and validates only purpose/source/dest, so a matching marker with `entries: [{ path: "../precious", type: "file" }]` makes this resolve outside staging and unlink that sibling. Even a genuine inventory can escape if an inventoried parent directory is replaced with a symlink before retry: cleanup visits the deepest child first and follows the intermediate symlink to unlink an external file. Please schema-validate markers as untrusted data (unique normalized `checklists/...` paths, no absolute/empty/`.`/`..` segments, resolved confinement), and walk from the root without ever evaluating descendants through a symlink/type mismatch. Add traversal-marker and replaced-parent-symlink tests.

packages/agency-lang/lib/eval/label/migrate/staging.ts
```diff
@@ -0,0 +1,242 @@
...
+ *
+ * `lstat`, never `stat`: a symbolic link must be seen as a link. `cpSync`
+ * preserves links by default, so a staged `checklists/shared -> /elsewhere`
+ * would let a later cleanup delete files through it, outside the staging
+ * directory entirely. Refusing is better than silently dereferencing, which
+ * would change what the migrated store contains.
+ */
+export function inventoryChecklists(sourceDir: string): StagedEntry[] {
+  const checklistsRoot = path.join(sourceDir, "checklists");
+  if (!fs.existsSync(checklistsRoot)) {
```
egonSchiele: The source checklist **root** still bypasses the symlink policy. `existsSync` follows links, then `readdirSync` inventories through `source/checklists -> /external`; a dangling root link is treated as no checklists. Please `lstat` `checklistsRoot` itself, treat only ENOENT as absent, and reject a symlink or non-directory. Copy-time checks should also refuse a file/directory whose type changed after inventory rather than following a newly substituted symlink.

packages/agency-lang/lib/eval/label/migrate/staging.ts
```diff
@@ -0,0 +1,242 @@
...
+ * than forced.
+ */
+export function removeStaging(stagingDir: string, marker: StagingMarker): void {
+  if (fs.lstatSync(stagingDir).isSymbolicLink()) {
+    throw new StagingError(
+      `${stagingDir} is a symbolic link, not a directory this migration created. Move it aside.`,
+    );
+  }
+
+  for (const name of ALWAYS_WRITTEN) {
```
egonSchiele: **Recovery regression:** cleanup deletes `.migration.json` first because it is the first `ALWAYS_WRITTEN` entry. If the retry crashes during any later inventory deletion, the next attempt sees an unmarked partially cleaned stage and refuses it forever. Keep the marker until every other owned entry is gone; then remove it and the directory. Recognize a markerless *empty* stage to cover the final unlink/rmdir crash window, and sync the parent after removal. Add fault tests during cleanup.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
```diff
@@ -0,0 +1,182 @@
...
+    const marker: StagingMarker = { purpose: MARKER_PURPOSE, sourceDir, destDir, entries };
+    writeMarker(stagingDir, marker);
+    args.faultBeforePublish?.();
+
+    copyInventory(sourceDir, stagingDir, entries);
+    writeStagedStore(plan, stagingDir);
+
+    // Prove the result opens through the ordinary read path before publishing
+    // it. A store that only this code can read is not a migration.
+    verifyStore(stagingDir, warn);
```
egonSchiele: `verifyStore` creates and removes a transient `.lock`, but lock removal does not fsync the store directory. Since staging was last synced before verification, a power loss after publication can resurrect that stale lock; interrupted-publication recovery then cannot open the destination and reports it as an unrelated existing store. Sync `storeDir` after releasing the verification lock (including error paths) before renaming.

packages/agency-lang/lib/cli/eval/labelCommand.ts
```diff
@@ -0,0 +1,141 @@
...
+  return (command.parent?.opts() as { store?: string } | undefined)?.store;
+}
+
+/** commander calls this once per repeat of a flag, accumulating the values. */
+export function collectRepeated(value: string, previous: string[]): string[] {
+  return [...previous, value];
+}
+
+function parseByteCap(value: string): number {
+  const parsed = Number.parseInt(value, 10);
```
egonSchiele, line 135: `parseInt` contradicts the "positive whole number" contract: `1.5`, `12junk`, and `1e6` are silently accepted as 1, 12, and 1, which can unexpectedly skip most of a batch. Match the whole decimal string, require a safe positive integer, and throw Commander's `InvalidArgumentError` so this is reported as a normal option error. The strict parser pattern already exists in `scripts/agency.ts` (`parseBoundedInt`).

egonSchiele (commented):
Third-pass findings are inline. The prior root-glob and publication ordering fixes look correct.

egonSchiele:
**CI remains blocked at `0e76dadda`:** `lib/eval/label/migrate/plan.ts` still contains the same raw NUL bytes at lines 70 and 100, and the Node 22 build still fails `lib/sourceIsText.test.ts`. The earlier comments on this remain unresolved.

egonSchiele:
Fixed at `aaf91b92`. `plan.ts` now shows **+232 with a real patch** in the PR files — it had been invisible all three times you flagged it.

The delimiter was deliberate; only its spelling was wrong. `\0` in a template literal has the identical runtime value, so nothing about the key changes.

**On missing it three times.** I scoped every test run to `lib/eval/label` and `lib/cli/eval`, which never loads `lib/sourceIsText.test.ts`. Each round I reported "all green" against a filter that structurally could not see this. Worse, `git diff --numstat` showed `-` `-` for that file the whole time — the binary marker was in front of me every time I checked the diff.

I have added the repo-wide guards to what I run before pushing, and swept every file this branch touches for NUL bytes; `plan.ts` was the only one.

Verification now: 585 passing across `lib/eval/label`, `lib/cli/eval` and the repo-wide guards, `pnpm run typecheck` (all three configs) and `lint:structure` clean. The four remaining local failures are `grade.test.ts` and `optimize.test.ts`, which need `dist/`; I confirmed they fail identically with this branch stashed.

packages/agency-lang/lib/eval/label/migrate/plan.ts
```diff
@@ -0,0 +1,232 @@
...
+  annotations: readonly AnnotationRow[];
+  fieldOrder: readonly string[];
+  counts: MigrationCounts;
+};
+
+/** Version 1 always stored a task, so every migrated record has both fields. */
+export function fieldsOfV1Row(row: V1CorpusRow): Fields {
+  return {
+    task: projectArtifactField(row.input.task),
+    output: projectArtifactField(row.value),
```
egonSchiele: **Data-integrity blocker:** migration reconstructs `output` from `row.value` and discards `row.text`, but v1 deliberately persisted `text` as the exact display projection so the labelled artifact remains exactly what the annotator saw even if projection changes. The v1 session displayed `row.text`; therefore recomputing here can move annotations onto different bytes while still producing a valid v2 store. Please use `row.text` for the migrated `output` field (while retaining `row.value` as occurrence provenance), and add a fixture where persisted `text` differs from projecting `value` to prove migration preserves the labelled artifact.

packages/agency-lang/lib/eval/label/migrate/plan.ts
```diff
@@ -0,0 +1,232 @@
...
+    };
+  }
+
+  for (const pairKey of Object.keys(pairs)) {
+    const { checklistId, annotator } = pairs[pairKey];
+    const states = oldIds.map((outputId) => ({
+      outputId,
+      answers: effectiveAnswers(annotations, { outputId, checklistId, annotator }),
+      note: latestNote(annotations, { outputId, checklistId, annotator }),
+    })).filter((state) => Object.keys(state.answers).length > 0 || state.note.length > 0);
```
egonSchiele: This filter treats an effective empty state as though this old output had no annotation, but an empty state can still affect the merged note fold. For example, A has `{q: true, note: "keep"}` and B later has a schema-valid note-only row `{coveredQuestionIds: [], answers: {}, note: ""}`. B is filtered out, so preflight succeeds; after both ids are rewritten, B's later row clears A's note. Reversing append order preserves the note, making migration order decide the result the guard is meant to protect. Please distinguish `no matching rows` from `matching rows whose effective answers/note are empty`, compare the latter as a real state, and add this ordering regression test.

egonSchiele (commented):
Fourth-pass findings are inline. The raw-NUL issue is fixed; these findings come from reviewing the now-visible migration planner.

packages/agency-lang/lib/eval/label/migrate/readV1.ts
egonSchiele: After reviewing the now-visible planner, yes—please add the `contentHash` check too. `fieldsOfV1Row` derives the new content identity from `row.input.task` and the labelled output, while the current output-id check only proves that `execution` was not edited. A v1 row whose task/value changed but whose execution-derived `outputId` stayed intact therefore migrates its annotations onto different content without any refusal. The persisted v1 `contentHash` is the available integrity witness for exactly those fields. Reproducing that one dead formula inside `readV1.ts` is justified for the same reason as `v1OutputIdOf`: this is a one-time migration, and silently relabelling edited content is worse than keeping a small legacy validator. Please add a stale-hash rejection test as well.

packages/agency-lang/lib/eval/label/migrate/staging.ts
egonSchiele: Both escapes confirmed. I reproduced the second one before touching anything, because my previous comment claiming the walk "never follows a link" was wrong and I wanted to know how wrong: `lstat` only declines to follow the **final** component of a path. With `checklists/cl_a -> /external`, `lstatSync("stage/checklists/cl_a/1.json")` resolves `cl_a` and reports a regular file, so `unlinkSync` deleted the external one. My probe printed `external/1.json survived: false`.

Two fixes, matching your two mechanisms:

- **The marker is parsed as untrusted data.** `StagingMarkerSchema` confines every entry path to the copied subtree: no absolute paths, no `..` in any position, no unnormalized or empty segments, no duplicates, no backslashes. A marker that fails reads as absent, so it is treated as "not ours" rather than trusted.
- **Removal descends one verified component at a time.** It reads a directory, `lstat`s its direct children, and recurses only into real directories. A symlink is unlinked as a link and never entered. Because every parent is verified before its children are read, no intermediate component can redirect the walk — which is the property my earlier version only had for leaves.

Seven tests in the new `staging.test.ts` for the schema alone (traversal, absolute, mid-path `..`, outside the copied subtree, duplicates, malformed-reads-as-absent), plus the replaced-parent-symlink case asserting the external file still reads `PRECIOUS`.

packages/agency-lang/lib/eval/label/migrate/staging.ts
egonSchiele: Confirmed and fixed. `inventoryChecklists` now `lstat`s the root itself, treats only ENOENT as absent, and rejects a symlink or a non-directory. A dangling root link is refused rather than read as "no checklists".

Also added the copy-time re-check you asked for: each source path is `lstat`ed as it is read, and a type change or a newly substituted symlink between inventory and copy aborts before anything is published.

Four tests: symlinked root, dangling root link, root that is a file, and a symlink nested inside the tree.

packages/agency-lang/lib/eval/label/migrate/staging.ts
egonSchiele: Confirmed. `.migration.json` was first in `ALWAYS_WRITTEN`, so it went before everything else.

Fixed by checking for unowned entries **before** unlinking the marker rather than discovering the problem at `rmdir` time. If anything unowned remains, it throws with the marker still in place. Only when the directory holds nothing but the marker does it unlink and `rmdir`.

Added `isEmptyDirectory` for the last window — marker unlinked, `rmdir` not yet run — since nothing else creates an empty directory at that exact name and removing one destroys nothing. The parent is synced after removal.

My first attempt still removed the marker before the `rmdir` and the test caught it: it asserts the marker survives when cleanup cannot finish.

packages/agency-lang/lib/eval/label/migrate/migrate.ts
egonSchiele: Fixed. `verifyStore` now syncs the store directory after releasing the lock, in a `finally` so it covers the error path too.

Worth spelling out why this is not cosmetic: staging was last flushed before verification, and verification creates and removes a transient `.lock`. A power loss after publication could therefore resurrect that lock inside the published store, and interrupted-publication recovery opens the destination to decide whether to finish it — so the store would fail to open and be reported as an unrelated existing directory. The failure would look nothing like its cause.

packages/agency-lang/lib/cli/eval/labelCommand.ts
egonSchiele: Fixed, using the `parseBoundedInt` pattern already in `scripts/agency.ts`: match `^\d+$` against the whole string, require a safe positive integer, and throw `InvalidArgumentError` so commander reports it as an ordinary option error.

Six tests covering `1.5`, `12junk`, `1e6`, `-1`, `0`, and a value past the safe-integer range, plus one asserting a plain integer still arrives. Each of the first three previously became 1, 12 and 1 — silently capping a batch.

packages/agency-lang/lib/eval/label/migrate/plan.ts
egonSchiele: You are right, and this is the finding I would least have found on my own. I had treated `text` as a cache of `value` when version 1 stored it as the *record* of what was displayed — the docs I wrote for the original store even say so, and I still re-derived it.

`fieldsOfV1Row` now takes `output` from `row.text`. `value` survives on the occurrence as `rawValue`.

Two tests, using the fixture you asked for: a v1 row whose persisted `text` differs from projecting its `value`, asserting the migrated `output` is the text and that `rawValue` still carries the structured value.

One limitation I want to state rather than leave implicit: `task` gets no such treatment, because version 1 never persisted a rendering of it. Projecting is the only option there, and I noted it in the function's comment.

packages/agency-lang/lib/eval/label/migrate/plan.ts
egonSchiele: Confirmed — the filter conflated "this output has no annotation" with "this output's effective state is empty", and only the first is a reason to skip the comparison.

It now selects old ids that have at least one matching annotation row, then compares their effective states, empty or not. Your exact scenario is the regression test: A holds `{q_accurate: true, note: "keep this"}`, B holds a schema-valid note-only row with an empty note, and migration refuses instead of letting append order decide whether the note survives.

A second test keeps the legitimate case working: a merge where one old output genuinely has no annotation still collapses.

packages/agency-lang/lib/eval/label/migrate/readV1.ts
egonSchiele: You are right, and I withdraw the pushback. My argument was about the cost of a second legacy formula; it did not address your actual point, which is that the id and the content hash witness *different fields*. The id proves `execution` was not edited, and the migrated record's fields come from the task and the value — so a row whose task or value changed with its execution intact carried its annotations onto different content with nothing to object.

`readV1Store` now checks both. Two tests: an edited task and an edited value, each refused by name.

One gap that remains, which I would rather state than have you find: **neither hash covers `text`.** Version 1 never hashed it, and since the artifact now comes from `text` (per your other finding), a hand-edited display projection is the one edit this cannot detect. There is no witness available for it. Noted in the dev doc alongside the two checks that do exist.

packages/agency-lang/lib/eval/label/load/discoverFiles.ts
```diff
@@ -0,0 +1,232 @@
...
+}
+
+/**
+ * Compile the supported subset to an anchored regular expression.
+ *
+ * `**` is handled before `*` so the greedy case wins, and `*` never crosses a
+ * `/` — otherwise `a/*.txt` would match `a/b/c.txt`, which is the difference
+ * between recursive and not.
+ */
+function globToRegExp(pattern: string): RegExp {
```
egonSchiele: **This is another large implementation we should not own.** The PR adds a 232-line glob/discovery engine plus 222 lines of dedicated tests even though the supported Node baseline (22+) already provides `fs.globSync`. The custom parser is already observably incomplete: `[z-a]` is copied into the generated regex and `new RegExp` throws a raw `SyntaxError: Range out of order`, while unmatched `[` is silently treated as a literal. Please replace the home-grown `splitPattern`/root/platform/glob compiler with the platform API (keeping only the small projection to `{absolutePath, itemKey, isSymlink}`), or drop glob input and retain directory + `--recursive`. This removes hundreds of lines and an entire parser surface rather than patching its next edge case.

packages/agency-lang/lib/eval/label/load/types.ts
```diff
@@ -0,0 +1,47 @@
...
+ *  index, or an eval input id — so a skip line is actionable without guessing. */
+export type IngestSkip = {
+  item: string;
+  reason: IngestSkipReason;
+};
+
+export type LoadedBatch = {
+  occurrences: readonly LoadedOccurrence[];
+  skips: readonly IngestSkip[];
+  discoveredFieldNames: readonly string[];
```
egonSchiele: `discoveredFieldNames` is entirely derivable from `occurrences[].fields`, but putting it in `LoadedBatch` makes every loader imperatively maintain a second representation of the same fact (`fieldNames` loops in run/files/json). That is exactly the anti-pattern where implementation bookkeeping leaks through an otherwise declarative interface, and the two representations can drift. Please remove this member and derive unique field names once at the ingest boundary from the accepted occurrences.

packages/agency-lang/lib/eval/label/load/run.ts
```diff
@@ -0,0 +1,213 @@
...
+    occurrences.push({
+      fields,
+      source: args.source,
+      origin: {
+        kind: "run",
+        traceId,
+        inputId,
+        finalOutputIndex: selection.index,
+        runStartedAtMs: typeof record.startedAtMs === "number" ? record.startedAtMs : null,
+        models: Array.isArray(record?.metrics?.models) ? record.metrics.models : [],
```
egonSchiele: **Validate the whole candidate before returning the batch.** `record` is `any`, so `metrics.models: [42]` passes this check and is typed through as though it were `string[]`. `LabelStore.ingest` then appends the corpus row first and only rejects the occurrence at `OccurrenceRowSchema.parse`, leaving an orphan record (and possibly earlier committed candidates) after a failed command. Please parse the constructed origin/candidate with the durable schema in the loader, or at minimum accept models only when every element is a string. Add a malformed-models test asserting the loader refuses/skips before any store mutation.

packages/agency-lang/lib/eval/label/ids.ts
```diff
@@ -25,23 +25,25 @@ function digestOf(identity: JsonValue): string {
...
- *  duplicate warnings — never as an identity. */
-export function contentHashOf(input: CorpusInput, value: JsonValue): string {
-  return `sha256:${digestOf({ input: { ...input }, value })}`;
+/** Identity of one observation. Deliberately excludes the timestamp: when you
+ *  observed something is not part of which observation it is. */
+export function makeOccurrenceId(candidate: OccurrenceCandidate): string {
+  return `occ_${digestOf({
+    outputId: candidate.outputId,
+    source: candidate.source,
+    origin: { ...candidate.origin },
```
egonSchiele: This hashes mutable descriptive provenance into the identity of an observation. For a run, changing/correcting `models`, `agent`, `runStartedAtMs`, `rawTask`, or `rawValue` while keeping the same `(traceId, inputId, finalOutputIndex)` creates a second occurrence for the same execution. That makes per-source occurrence counts overstate observations—the use case this log was added to support. Please define identity from the stable locator for each origin kind (for runs: kind + traceId + inputId + finalOutputIndex; for files/JSON: item key/index), and keep the remaining fields as payload rather than identity.

packages/agency-lang/lib/eval/label/store.ts
```diff
@@ -69,10 +84,11 @@ export type LabelStoreSnapshot = {
 };
 
 export type LabelStore = {
-  captureSource(args: CaptureSourceArgs): CaptureResult;
+  ingest(batch: LoadedBatch): IngestResult;
   readSession(sessionId: string): DeepReadonly<LabelStoreSnapshot>;
   saveDraft(draft: Draft): void;
   corpusSnapshot(): readonly DeepReadonly<CorpusRow>[];
+  occurrenceSnapshot(): readonly DeepReadonly<OccurrenceRow>[];
```
egonSchiele: This new facade method has no production caller; only tests use it, and it exposes the private provenance log wholesale (including raw tasks/values). It is speculative API surface rather than an operation needed by ingest or labeling. Please remove `occurrenceSnapshot()` until a concrete consumer needs a purpose-specific projection.

packages/agency-lang/docs/dev/eval-labeling.md
```diff
@@ -1,6 +1,21 @@
...
+`agency label` builds a dataset of human judgements about agent outputs. It is
+registered twice — as `agency label` and as `agency eval label` — following the
+same dual registration `optimize` uses.
+
+Three commands:
+
+```bash
+agency label ingest <source> --source <name>   # add records
+agency label --checklist <file>                # judge what the store holds
+agency label migrate <old> <new>               # move a store to the current format
```
egonSchiele: The migration was deleted but the docs still advertise `agency label migrate`; the command is no longer registered. Please remove this line/change “Three commands,” and also clean the remaining stale migration wording in `StoreVersionError` (its class comment says “needs migrating,” and the newer-version error says migration moves stores forward even though no migration exists).

egonSchiele (commented):
Re-reviewed the reduced PR from first principles. The migration deletion is the right scope reduction; remaining findings are inline, including two more places where the implementation carries unnecessary machinery.

packages/agency-lang/lib/eval/label/load/discoverFiles.ts
egonSchiele: Deleted, not replaced. Both bugs reproduced first: `[z-a]` reaches `new RegExp` and throws `SyntaxError: Range out of order in character class` instead of anything a user could act on, and an unmatched `[` is silently a literal.

I took your second option rather than `fs.globSync`. `fs.globSync` is still marked experimental on the Node 22 line this repo builds against, and swapping one surface for another with a stability caveat did not seem like the lesson. Ingest now takes a directory with `--recursive`, which is 232 lines of parser and 222 lines of tests gone. Filtering a directory is a job for the shell, and the eligibility policy already skips what a broad selection sweeps up.

This is the same mistake as the migration in this PR: I built infrastructure for a nice-to-have and then spent review rounds hardening it. Two for two on "delete it" being the right call.

If pattern input turns out to be genuinely wanted, `fs.globSync` behind the existing `FileSelection` shape is a small change once it is stable.

packages/agency-lang/lib/eval/label/load/types.ts
egonSchiele: Removed. Every loader was keeping a `fieldNames` index alongside output that already stated the same fact.

The store now derives it once, at the ingest boundary, from the candidates it actually accepted. That is also more correct than what the loaders reported: theirs counted fields on candidates that a later validation step could still reject.

packages/agency-lang/lib/eval/label/load/run.ts
egonSchiele: Confirmed and fixed. `record` is `any`, so `Array.isArray(record?.metrics?.models)` accepted `[42]` and typed it through as `string[]`.

The origin is now built with `OccurrenceOriginSchema.safeParse` inside the loader, and a record that fails is skipped as `record-unreadable` before anything reaches the store. `models` also gained `.default([])`, which separates the two cases: a record with no metrics block is ordinary, while a metrics block holding non-strings is malformed.

The ordering point was the important part. `ingest` appends the corpus row first, deliberately, because a dangling occurrence is unrecoverable while a record without one is not. That makes a late schema failure leave an orphan, so validation has to happen before the store sees the candidate at all.

Two tests: `models: [42]` is skipped with no store mutation, and a missing metrics block loads normally with `models: []`.

packages/agency-lang/lib/eval/label/ids.ts
egonSchiele: Confirmed, and this one defeated the log's purpose rather than merely being untidy. Re-ingesting a run after a corrected model name produced a second occurrence for the same execution, so per-source counts overstated exactly the thing occurrences were added to measure.

Identity now comes from `occurrenceLocatorOf`, which returns only the stable locator: kind, trace, input and output index for a run; kind and item key for a file; kind, item key and index for a JSON element. Everything else is payload on the row.

Three tests: a corrected model name replays rather than duplicating; changed agent provenance, start time, raw task and raw value all replay; a different trace or output index is still a different observation.

packages/agency-lang/lib/eval/label/store.ts
egonSchiele: Removed. You are right that it was speculative, and worse than unused: it handed out the provenance log wholesale, raw tasks and values included, when a real consumer would want a projection shaped to its question.

The tests that used it now read `occurrences.jsonl` directly, which is what they were actually asserting about.

packages/agency-lang/docs/dev/eval-labeling.md
egonSchiele: All cleaned. The dev doc said "Three commands" and listed `agency label migrate`; it now says two. I also removed the glob mention there and in the user-facing page, since that engine is gone in this same push.

The two wording leftovers you found are fixed as well: the `StoreVersionError` comment no longer says "needs migrating", and the newer-format message no longer claims migration moves stores forwards. It now just says to upgrade.

The `legacy` occurrence kind was another leftover I had missed. Migration was the only thing that produced it, so it was unreachable; removed.

packages/agency-lang/lib/eval/label/load/index.ts
```diff
@@ -0,0 +1,138 @@
...
+import { loadFiles } from "./files.js";
+import { resolveFormat, type Format } from "./format.js";
+import { loadJsonArray } from "./json.js";
+import { loadRun } from "./run.js";
+import { IngestSourceError, type LoadedBatch } from "./types.js";
+
+/** Field names each loader builds itself. A constant may not collide with one:
+ *  the loader's value wins on merge, so the constant would vanish without a
+ *  word. */
+const LOADER_FIELD_NAMES: Record<Exclude<Format, "auto">, readonly string[]> = {
```
egonSchiele, line 17: This table is a second owner of the fields each loader produces. `run.ts` separately constructs `task`/`output`, while `files.ts` and `json.ts` separately construct `output`; adding or renaming a loader field without updating this table silently weakens collision detection. The same duplication appears again in each loader's `checkEligibility → merge constants → build occurrence/skip` sequence. Please replace both with one declarative helper that accepts the loader-produced fields and constant fields, rejects overlap from the actual keys, applies eligibility, and returns a loaded occurrence or skip. Then loaders only describe their source-specific candidate and there is no reserved-field registry to keep synchronized.

packages/agency-lang/lib/eval/label/load/run.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/**
+ * Choose the output a label applies to.
+ *
+ * The index is part of the answer, not an afterthought: it is recorded on the
+ * occurrence, so selection and provenance can never disagree about which
+ * output was judged.
+ */
+export function selectLabelingFinalOutput(record: unknown): FinalOutputSelection {
```
egonSchiele, line 41: This independently reimplements final eval-output selection already owned by `lib/eval/judge/selectFinalResponse.ts`: inspect `evalOutputs`, choose the last entry, handle legacy/missing/truncated values, and project the value. The policies differ at the edges, so calling the judge helper directly would be wrong, but duplicating the structural selection guarantees they will drift. Please extract one shared selector that returns a typed union containing index/value/truncation/legacy status; labeling and judging can remain small adapters with their own acceptance and display policies.

packages/agency-lang/lib/eval/label/occurrences.ts
```diff
@@ -0,0 +1,71 @@
...
+    schema: OccurrenceRowSchema,
+    identityOf: (row) => row.occurrenceId,
+  });
+
+  return {
+    rows(): readonly OccurrenceRow[] {
+      return log.rows() as readonly OccurrenceRow[];
+    },
+
+    ensureOccurrence(candidate: OccurrenceCandidate): EnsureOccurrenceResult {
```
egonSchiele, line 52: `ensureOccurrence` and `corpus.ensureRecord` duplicate the same persistence protocol almost line for line: derive identity, `find`, return the stored row to preserve the first timestamp, schema-parse a new row, `appendExact`, return `{row, added}`. This is a load-bearing replay policy with two implementations and two long copies of its rationale. Please put `ensure(identity, createRow)` on `OpenedJsonl` (or an equivalent generic helper in `jsonl.ts`) and have both domain logs supply only their identity and row builder. That makes the first-write-wins timestamp behavior one invariant rather than two synchronized implementations.

packages/agency-lang/lib/cli/eval/labelCommand.ts
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * A positive whole number of bytes, or an option error.
+ *
+ * `parseInt` stops at the first character it cannot read, so "1.5", "12junk"
+ * and "1e6" become 1, 12 and 1 — each silently capping a batch far below what
+ * was asked for. Matching the whole string is the only way to reject them.
+ * `InvalidArgumentError` makes commander report it as a normal option problem
+ * rather than a crash.
+ */
+export function parseByteCap(value: string): number {
```
egonSchiele, line 129: This repeats the strict whole-string positive-integer parser already implemented as `parseBoundedInt`/`parsePositiveInt` in `scripts/agency.ts`, along with a second test matrix for the same `parseInt` footguns. The existing helper should also gain the safe-integer check this byte cap needs. Please move the bounded safe-integer parser into a shared `lib/cli` module, parameterize its minimum/error label, and use it from both command registrations rather than keeping two parsers that differ subtly on oversized values.

egonSchiele (commented):
Duplication-focused pass. Four remaining places have two owners for one policy or algorithm; findings are inline.

packages/agency-lang/lib/eval/label/types.ts
```diff
@@ -79,7 +81,10 @@ export type SessionIdentity = {
 // --- durable rows --------------------------------------------------------
 
 export const ManifestSchema = z.object({
-  schemaVersion: z.literal(1),
+  schemaVersion: z.literal(2),
```
egonSchiele, line 84: Store format version has two source-of-truth constants: this `z.literal(2)` and `CURRENT_STORE_VERSION = 2` in `store.ts`. A future bump can update one and leave `assertStoreVersion` accepting a version that `ManifestSchema` rejects (or vice versa). Please define/export the version once beside the manifest schema and use `z.literal(CURRENT_STORE_VERSION)` plus the same value in store errors/checks.

packages/agency-lang/lib/cli/eval/labelCommand.ts
```diff
@@ -0,0 +1,142 @@
...
+    .description("Add outputs to the label store, from a run, a directory of files, or a JSON array")
+    .argument("<source>", INGEST_PATH_DESCRIPTION)
+    .argument("[extra...]", "Rejected: several arguments means the shell expanded an unquoted glob")
+    .option("--source <name>", SOURCE_FLAG_DESCRIPTION)
+    .option("--format <fmt>", "auto (default), run, files, or json")
+    .option("--task <text>", "Shorthand for --field task=<text>")
+    .option("--field <name=value>", "A constant field added to every record", collectRepeated, [])
+    .option("--no-task-field", "Run sources only: drop the run's own task field")
+    .option("--recursive", "Descend into subdirectories")
+    .option("--max-bytes <n>", "Per-value size cap in bytes (default 1048576)", parseByteCap)
```
egonSchiele, line 84: The CLI metadata duplicates two runtime constants: line 79 hand-writes the formats already exported as `FORMATS`, and this help string hand-writes `1048576` while the actual fallback is `DEFAULT_MAX_INGEST_BYTES`. Either can drift while behavior remains unchanged and help becomes false. Please build these descriptions from the exported constants (or give Commander the actual default) rather than restating them.

egonSchiele (commented):
Two additional duplicated constants are inline. The `ensureRecord`/`ensureOccurrence` duplication from my preceding review was already removed concurrently in this head via `findOrAppend`; the other three findings remain applicable.

PR #798, opened 2026-08-04
feat(remote): inspect, pull, and logs introspection commands

egonSchiele:
## What

The agency-lang half of statelog's **Group-2 remote introspection** — three project-scoped read commands under `agency remote`:

- **`remote inspect`** — what's deployed (entry point, last upload, each file's exported nodes).
- **`remote pull [--out <dir>] [--force]`** — download the deployed source to disk.
- **`remote logs [traceId] [--json] [--list]`** — open a trace's logs in the existing `agency logs` viewer (or `--json` to stdout, `--list` to see traces).

All four routes are `/api/projects/:slug/*` and **project-key-accessible**, so the ordinary `STATELOG_API_KEY` works — no account key needed. Built from the reviewed spec (Revision 3) and plan (Revision 4), one TDD task per commit.

## Architecture

- **Sealed `projectClient`** (`lib/cli/statelog/projectClient.ts`) — the `/api/projects/:slug/*` wire, per-segment URL encoding, non-2xx-before-envelope, and every success value validated from `unknown` with non-empty ids. Keeps the three failure layers distinct: project-**404**, wrong-project-**403**, and the HTTP-200 `success:false` **"Trace not found"**.
- **`resolveProjectTarget`** derives origin+slug+key from **one binding snapshot** (a binding slug is used only when the resolved origin matches the binding's).
- **Pull safety** (`pullPlan.ts`): server filenames are untrusted, so a non-mutating `planSourcePull` (traversal/duplicate/NFC-collision/non-regular checks, all conflicts collected before any write) plus an atomic `applySourcePull` (create via `fs.link`/`EEXIST`, `--force` revalidate-then-rename; temps cleaned up; `PullApplyError` reports committed files). Not transactional by design.
- **Logs bridge** (`logsBridge.ts`): `Log[]` → JSONL carrying the `trace_id` the viewer parser requires (validated through the **real `parseStatelogJsonl`**), and one **`runViewerOnTerminal`** owner for terminal lifecycle. Viewer mode requires an interactive stdin/stdout; a pure `resolveRemoteLogsMode` + `requireRemoteLogsEnvironment` decide everything **before** any config/credential/effect.
- **Presenters**: `failProjectCommand` (clean, secret-free, no stack) and `printJson` (one stdout write) keep `--json` output pure.

## Verification

- **150 tests** across the 11 files (3 build-gated CLI-integration skips): client transport + 3-layer errors + malformed-value table; the bridge through the real parser; pull traversal/symlink/post-plan-race/atomic-replace; single-binding-read; viewer lifecycle cleanup + combined errors; the logs mode matrix + non-TTY do-nothing; renderers; recipes (JSON stdout purity, no-write-on-violation); and `parseAsync` registration incl. the rejected `[traceId] --list`.
- `pnpm run typecheck` (source + tests + evals), `pnpm run lint:structure`, and `git diff --check` all clean.

## Notes

- #14's "projects by slug" change does **not** regress the merged account CLI — the account-level `/api/api_keys` still takes the internal id in its body.
- Out of scope (as specced): follow/live remote logs, an interactive trace picker, and the `remote link --project` convenience.


packages/agency-lang/lib/cli/remote/commands/util.ts
```diff
@@ -82,6 +103,60 @@ export function resolveAccountTarget(
...
+/** Resolve a project-read target — one coherent origin+slug+key from ONE binding
+ *  snapshot. A binding slug is only used when the resolved origin matches the
+ *  binding's origin, so a linked production project can't be spliced onto another
+ *  host. */
+export function resolveProjectTarget(
+  context: RemoteCommandContext,
+  options: ProjectCommandOptions,
+): ProjectTarget {
+  const binding = readBinding(context.configPath);
+  const account = resolveAccountTargetFromBinding(context, options, binding);
```
egonSchiele: `resolveAccountTargetFromBinding` resolves the API key before `resolveProjectSlug` runs. With the key unset, `--project ""`, no project/binding, and a binding/host mismatch all report `Missing API key` instead of the project error, and this violates the stated rule that CLI-input errors are rejected before credential access. Please split origin resolution from key resolution: read one binding, resolve/canonicalize the origin, resolve the slug against that origin, and call `resolveApiKey` last. A regression test with `STATELOG_API_KEY` unset would pin the ordering.

packages/agency-lang/lib/cli/remote/pullPlan.ts
```diff
@@ -0,0 +1,222 @@
...
+    return message(error);
+  }
+}
+
+function describePullFailure(
+  file: PlannedPullFile,
+  error: unknown,
+  cleanupError: string | undefined,
+): string {
+  const base = `pull failed writing ${file.name}: ${message(error)}`;
```
egonSchiele, line 223: This message drops `committed`, while `failProjectCommand` prints only `error.message`. If `a.agency` commits and publication of `b.agency` then fails, the CLI tells the user only about `b.agency`; it does not disclose that `a.agency` is already on disk. That contradicts the non-transactional pull contract. Please make `PullApplyError.message` self-contained by appending the committed destinations (and add a command-level stderr assertion, not only an assertion on the error property).

packages/agency-lang/scripts/agency.ts
```diff
@@ -538,6 +551,57 @@ export function createProgram(deps: CliDependencies = {}): Command {
...
+        try {
+          mode = resolveRemoteLogsMode(traceId, opts);
+          requireRemoteLogsEnvironment(mode, {
+            stdinIsTTY: process.stdin.isTTY === true,
+            stdoutIsTTY: process.stdout.isTTY === true,
+          });
+        } catch (error) {
+          failProjectCommand(error);
+        }
+        return runLogs(mode, opts, getConfigContext());
```
egonSchiele, line 592: `getConfigContext()` can write to stdout before `runLogs`: `loadConfig` uses `console.log` for both `--verbose` diagnostics and `config.verbose`. Therefore `agency --verbose remote logs --json` (and a config with `verbose: true`) emits prose followed by JSON, breaking the promised stdout-pure machine format. Please route config diagnostics to stderr or provide a quiet config-loading path for JSON mode, and cover both verbose sources in a real registration/output test.

packages/agency-lang/lib/cli/logsView.ts
```diff
@@ -107,6 +109,126 @@ async function runExplorerOnTerminal(
...
+    return;
+  }
+  if (all.length === 1) {
+    throw all[0];
+  }
+  throw new AggregateError(all, "viewer terminal lifecycle failed");
+}
+
+const defaultViewerHost = createViewerHost({
+  createInput: () => new TerminalInput(),
```
egonSchiele, line 239: The default `TerminalOutput` still installs SIGINT/SIGTERM handlers that destroy only the output and immediately call `process.exit`. An external signal therefore bypasses this host's `finally`, so `TerminalInput.destroy()` and `restoreStdin()` do not run; the new host is not actually the single lifecycle owner on that path and can leave terminal input state behind. Please allow `TerminalOutput` signal ownership to be disabled here and have the host route signals through the same release path (with a regression test that input/output/restore all run).

egonSchiele (commented):
Four correctness findings are called out inline. The focused implementation tests pass, but these edge paths are not covered by them.

packages/agency-lang/lib/cli/remote/commands/util.ts
egonSchiele: Fixed in 72a4c1102. Split origin resolution out of key resolution: resolveProjectTarget now reads one binding, resolves/canonicalizes the origin, resolves the slug against that origin, and calls resolveApiKey LAST. Regression tests with STATELOG_API_KEY unset pin the ordering for all four input errors (empty --project, no host, no project, host/binding mismatch) — each reports its input error, none says "Missing API key".

packages/agency-lang/lib/cli/remote/pullPlan.ts
egonSchiele: Fixed in 72a4c1102. PullApplyError now folds the committed destinations into its message ("Already written to disk (not rolled back): …"), so the sole error.message a command prints is self-contained. pullPlan.test.ts asserts the message names the committed a.agency; a new pull.test.ts case asserts the command prints the whole message to stderr.

packages/agency-lang/scripts/agency.ts
egonSchiele: Fixed in 72a4c1102. Routed both verbose diagnostics to stderr — the --verbose "Looking for config at" and the config.verbose "Loaded config from" in commands.ts, plus the config.verbose line in config.ts. A new commands.test.ts case loads a verbose config with --verbose and asserts console.log is never called while both messages land on stderr, so --json stdout stays pure.

packages/agency-lang/lib/cli/logsView.ts
egonSchiele: Fixed in 72a4c1102. TerminalOutput takes manageInterruptSignals (default true); the viewer host creates its output with it false and installs its own SIGINT/SIGTERM handlers that run the shared release-once path (output + input + stdin restore) before process.exit. A regression test emits SIGINT mid-run and asserts all three teardowns run exactly once, exit(130) fires, and the host leaves no leaked SIGINT listener.

packages/agency-lang/lib/cli/statelog/projectClient.ts
```diff
@@ -0,0 +1,255 @@
...
+
+function validateEnvelope(value: unknown): { success: boolean; value?: unknown; error?: unknown } {
+  const obj = asObject(value);
+  if (!obj || typeof obj.success !== "boolean") {
+    throw new ProjectRequestError("unexpected project response shape");
+  }
+  return obj as { success: boolean; value?: unknown; error?: unknown };
+}
+
+function validateAgentMetadata(value: unknown): AgentMetadata {
```
egonSchiele: all your validation logic is just rebuilding Zod. Why not use Zod schemas here?

packages/agency-lang/lib/cli/commands.ts
```diff
@@ -94,8 +94,10 @@ export function loadConfig(
 ): AgencyConfig {
   const finalConfigPath = configPath || path.join(process.cwd(), "agency.json");
 
+  // Diagnostics go to stderr so they never contaminate a command's
+  // machine-consumed stdout (e.g. `remote logs --json`).
   if (verbose) {
-    console.log(`Looking for config at: ${finalConfigPath}`);
+    console.error(`Looking for config at: ${finalConfigPath}`);
```
egonSchiele, line 100: could use the logger instead (lib/logger.ts)

packages/agency-lang/lib/cli/logsView.ts
```diff
@@ -107,6 +109,152 @@ async function runExplorerOnTerminal(
...
 
+export type ViewerTerminalInput = "current-stdin" | "controlling-tty";
+
+/** What the one viewer host runs: in-memory text (with the input-source choice)
+ *  or a local file with follow. */
+type ViewerSource =
+  | { kind: "text"; jsonl: string; terminalInput: ViewerTerminalInput }
+  | { kind: "file"; followPath: string; initialFollow: boolean };
+
+type ViewerTerminalDependencies = {
```
egonSchiele, line 120: This file is ugly... is there not a cleaner way to handle passing to the logs viewer?

packages/agency-lang/lib/cli/remote/commands/inspect.ts
```diff
@@ -0,0 +1,22 @@
+import { createProjectClient } from "../../statelog/projectClient.js";
+import { renderAgent } from "../render.js";
+import { resolveProjectTarget, failProjectCommand } from "./util.js";
+import type { ProjectCommandOptions, RemoteCommandContext } from "./util.js";
+
+/** `agency remote inspect` — what's deployed (files + exported node names). */
```
egonSchiele: Note: summary on `ls` should show types and default vals for args

egonSchiele: Should `inspect` also show functions? `inspect` feels not very useful since we also have `ls`

packages/agency-lang/lib/cli/remote/commands/pull.ts
```diff
@@ -0,0 +1,31 @@
+import { createProjectClient } from "../../statelog/projectClient.js";
```
egonSchiele, line 1: Running pull gave me this error:

```
Deployed source is unavailable for one or more files; re-deploy the agent
```

No error in the server logs

packages/agency-lang/lib/cli/statelog/projectClient.ts
egonSchiele: Done in 2ae468eb4 — converted all three sealed statelog clients (project/account/serve) to Zod schemas for one consistent style across lib/cli/statelog, not just projectClient. The transport/envelope layer and its pinned error messages are unchanged; only success-value validation moved to Zod, with snake_case wire fields mapped to camelCase via schema .transform(). accountClient's api_keys array guard and the 'created key missing plainKey' check stay as-is (their messages were pinned).

packages/agency-lang/lib/cli/commands.ts
egonSchiele: Kept console.error (stderr) rather than the logger, and here's why: logger.info/debug route through Node's console.info/debug, which write to STDOUT. Using the logger for these config diagnostics would re-break the remote logs --json stdout purity that the sibling finding on scripts/agency.ts asked me to fix (only warn/error reach stderr). Confirmed this direction with the owner.

packages/agency-lang/lib/cli/logsView.ts
egonSchiele: Left as-is per your call after we talked it through. The indirection (ViewerSource + createViewerHost lifecycle) is load-bearing: it's the single-owner terminal lifecycle from the previous review round, which is also what lets the host route SIGINT/SIGTERM through the full release path. Happy to extract the host into its own viewerHost.ts later if the file-level mixing still bugs you.

packages/agency-lang/lib/cli/remote/commands/inspect.ts
egonSchiele: Deferred with remote inspect (see the thread below). Showing arg types + default values needs the manifest/endpoint to carry them, which it currently doesn't (serve manifest parameters is string[]) — that's a statelog-side change, tracked in the hosted-agent-execution doc as a follow-up.

egonSchiele: Agreed — removed remote inspect from this PR in 2ae468eb4. Its output (per-file exported node names) was a strict subset of ls, and adding functions + typed/defaulted params needs the statelog /api/projects/:slug/agent endpoint to return them (it currently returns only nodeNames). Deferred it in the doc so it can relaunch actually differentiated from ls once that endpoint lands.

packages/agency-lang/lib/cli/remote/commands/pull.ts
egonSchiele: That message is statelog's own handled response, not a CLI error — src/backend/routes/api/projects/[id]/source.ts returns failure('Deployed source is unavailable for one or more files; re-deploy the agent') whenever any deployed agencyFile.source is null. It's a normal {success:false} envelope, which is why there's no server-log entry (nothing threw), and the CLI surfaces it verbatim. Root cause is server-side: that project's stored files have a null source column, so re-deploying with a source-populating path (or a statelog fix to backfill source) resolves it — nothing to change in this PR.

PR #799, opened 2026-08-04
fix(runtime): install the root budget on the served-function path

egonSchiele:
## The bug

Only `runNode` installed the root cost/time budget. A served
`POST /function/:name` runs through `runExportedFunction`, which builds a
fresh execution context but never installed the root guard — so a served
function ran **uncapped**. A baked `maxCost` / `maxTime` was silently
ignored for functions (it only ever applied to nodes).

## The fix

Hoist both `installRunPolicyHandler` and `installRootBudget` into the shared
`initFreshExecCtx` bootstrap that **both** fresh-run entry points
(`runNode`, `runExportedFunction`) call. Now nodes and served functions are
capped identically, and any future fresh-run entry point inherits the guard.

- The resume family (`respondToInterrupts` / `rewindFrom`) does **not** use
  `initFreshExecCtx` — it re-asserts the budget through `reinstallRootBudget`
  — so this does not double-install on resume.
- Both installers are root-only (no-op in IPC subprocesses, whose budgets the
  parent guard owns) and run at the end of bootstrap, outermost and before any
  user body, so neither can be bypassed. The relative ordering inside `runNode`
  is unchanged (both were already installed right after `initFreshExecCtx`).

## Test

A served function driven through the **real HTTP adapter + `runExportedFunction`**
charges $1 (via `addCost`, the same sequence the built-in `llm()` path runs
after a completion). With a baked `maxCost: 0` it now trips the root guard and
returns the typed **402 `budgetExceeded`** (dimension/limit/spent). A control
with no baked budget returns 200 — proving the guard is what caps it. Before
this fix, the first case returned 200 because nothing was installed.

## Verification

- `pnpm run typecheck` (source + tests + evals), `pnpm run lint:structure`, and
  `git diff --check` all clean.
- New `functionBudget.test.ts` (2 cases) + `rootBudget.test.ts` (20) pass. The
  `functionFrame.integration.test.ts` (needs a built `dist`) is left to CI.

### DIFF
--- a/packages/agency-lang/lib/runtime/node.ts
+++ b/packages/agency-lang/lib/runtime/node.ts
@@ -98,17 +98,28 @@ export function setupFunction(): {
  * callback registration — each inside a bootstrap ALS frame.
  *
  * Shared by the two fresh-run entry points (`runNode`, `runExportedFunction`).
- * The resume family (`respondToInterrupts`, `rewindFrom`) does NOT use this:
- * it restores statics/globals from a checkpoint and only re-registers
- * top-level callbacks (and re-asserts the root budget via reinstallRootBudget).
+ * The resume family does NOT use this — it restores statics/globals from a
+ * checkpoint and only re-registers top-level callbacks — and each member
+ * handles the root budget differently:
+ *   - `respondToInterrupts` (interrupts.ts) re-asserts it via
+ *     `reinstallRootBudget` (and reinstalls the run-policy handler).
+ *   - `rewindFrom` (rewind.ts) does NEITHER today: it only
+ *     `createExecutionContext` + `restoreState`, so a rewound run installs no
+ *     root budget and, if the checkpoint carries a serialized root guard, runs
+ *     under the checkpoint's limit rather than a re-asserted host one. That is
+ *     a pre-existing gap, out of scope here (tracked separately).
  *
  * The root run-policy handler and root cost/time budget are installed here, at
  * the end of bootstrap, so EVERY fresh-run entry point (including a served
  * `/function/:name` call through runExportedFunction) is capped identically —
  * a served function must not run uncapped. Both installers are root-only
- * (no-op in IPC subprocesses, whose budgets the parent guard owns) and are the
- * outermost thing on the stack before any user body runs, so neither can be
- * bypassed.
+ * (no-op in IPC subprocesses, whose budgets the parent guard owns). They sit
+ * after global init and top-level callback registration, so they are outermost
+ * before the entry NODE/FUNCTION body runs and cannot be bypassed there; the
+ * small amount of user code that global-init expressions and top-level
+ * `callback(...)` blocks can run during bootstrap still runs before the guards
+ * exist (unchanged from before this hoist — runNode installed them after
+ * bootstrap too).
  */
 async function initFreshExecCtx(
   execCtx: RuntimeContext<GraphState>,
@@ -179,9 +190,12 @@ async function initFreshExecCtx(
   // Install the CLI-driven root policy handler (agency run --policy) and the
   // root cost/time budget (--max-cost / --max-time, via env / RuntimeContext).
   // Both are root-only (isIpcMode gate inside each installer): no-op in IPC
-  // subprocesses. Installed here, after bootstrap and before any user body,
-  // so they are the outermost handler/guard and cannot be bypassed — and so
-  // BOTH fresh-run entry points (nodes and served functions) inherit them.
+  // subprocesses. Installed at the end of bootstrap, so they are the outermost
+  // handler/guard before the entry node/function body runs and cannot be
+  // bypassed there — and so BOTH fresh-run entry points (nodes and served
+  // functions) inherit them. (Global-init and top-level-callback code above
+  // runs before these exist; that's unchanged — runNode installed them here
+  // too, just after initFreshExecCtx returned.)
   installRunPolicyHandler(execCtx);
   installRootBudget(execCtx.stateStack, execCtx.budget);
 }
--- a/packages/agency-lang/lib/runtime/node.ts
+++ b/packages/agency-lang/lib/runtime/node.ts
@@ -100,7 +100,15 @@ export function setupFunction(): {
  * Shared by the two fresh-run entry points (`runNode`, `runExportedFunction`).
  * The resume family (`respondToInterrupts`, `rewindFrom`) does NOT use this:
  * it restores statics/globals from a checkpoint and only re-registers
- * top-level callbacks.
+ * top-level callbacks (and re-asserts the root budget via reinstallRootBudget).
+ *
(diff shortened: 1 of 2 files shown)


packages/agency-lang/lib/runtime/node.ts
```diff
@@ -100,7 +100,15 @@ export function setupFunction(): {
  * Shared by the two fresh-run entry points (`runNode`, `runExportedFunction`).
  * The resume family (`respondToInterrupts`, `rewindFrom`) does NOT use this:
  * it restores statics/globals from a checkpoint and only re-registers
- * top-level callbacks.
+ * top-level callbacks (and re-asserts the root budget via reinstallRootBudget).
```
egonSchiele: This says the resume family — **including `rewindFrom`** — "re-asserts the root budget via `reinstallRootBudget`." That's accurate for `respondToInterrupts` (interrupts.ts calls `reinstallRootBudget`), but **not** for `rewindFrom`: `rewind.ts` does `createExecutionContext` + `restoreState(checkpoint)` and calls neither `initFreshExecCtx` **nor** `reinstallRootBudget` (nor `installRootBudget`). So a rewound run installs no root budget at all — and if the restored checkpoint carries a serialized root guard, it runs under the **checkpoint's** limit/spend rather than a re-asserted host limit (the exact thing `reinstallRootBudget` was added in #796 to prevent on the stateless resume path).

Out of scope for this PR (which correctly targets the function path) — but two asks: (1) fix this comment so it doesn't claim `rewindFrom` re-asserts when it doesn't; and (2) worth confirming separately whether `rewindFrom` needs its own `reinstallRootBudget` — if it's ever reachable in a budgeted or served context with a client-supplied checkpoint, it's uncapped / trusts client guard state today.

egonSchiele (commented):
The fix is correct and the test genuinely proves it. Verified against the branch:

- `initFreshExecCtx` has exactly **two** callers (`runNode` at ~257, `runExportedFunction` at ~339) — both fresh-run entry points — so hoisting installs the guards for the function path and nothing else.
- **No double-install:** `runNode`'s old explicit `installRunPolicyHandler` / `installRootBudget` are removed; both now happen exactly once via `initFreshExecCtx`, at the same relative position (end of bootstrap, before the user body). `execCtx.budget` is populated by `createExecutionContext` before bootstrap, so `installRootBudget` reads the baked budget.
- **Resume unaffected:** `respondToInterrupts` doesn't call `initFreshExecCtx`; it keeps its own `installRunPolicyHandler` + `reinstallRootBudget`, so no double-install and the host-authoritative clamp still applies on resume.
- The new `functionBudget.test.ts` drives a real `runExportedFunction` through the actual HTTP adapter, and the no-budget control returning 200 vs. `maxCost:0` returning a typed 402 `{dimension:cost, limit:0, spent:1}` genuinely isolates the guard as the cause. `AGENCY_IPC` is saved/cleared/restored so a leak can't mask it. Good.

Hoisting `installRunPolicyHandler` too (so served functions also get the run policy handler) is a reasonable parity change — no-op without `AGENCY_RUN_POLICY`, and it's intentional per the description.

This unblocks statelog Group 3. One inline note on a doc/scope point that's outside this diff but that the comment here asserts.

packages/agency-lang/lib/serve/http/functionBudget.test.ts
```diff
@@ -0,0 +1,95 @@
...
+ * not trip and the call returned 200.
+ */
+describe("a served function respects the baked root budget", () => {
+  // AGENCY_IPC would make installRootBudget a no-op (a child's budget is the
+  // parent's); clear it so a leak from another test can't mask the guard.
+  const savedIpc = process.env.AGENCY_IPC;
+  afterEach(() => {
+    if (savedIpc === undefined) delete process.env.AGENCY_IPC;
+    else process.env.AGENCY_IPC = savedIpc;
+  });
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c79d80ef5. Now snapshot/clear/restore AGENCY_MAX_COST and AGENCY_MAX_TIME alongside AGENCY_IPC (those env budgets override ctx.budget, flag wins). Confirmed the suite passes with AGENCY_MAX_COST=999 set in the environment — before the fix that would have made the 402 case report limit 999 and left the control capped.

packages/agency-lang/lib/runtime/node.ts
```diff
@@ -100,7 +100,15 @@ export function setupFunction(): {
...
+ *
+ * The root run-policy handler and root cost/time budget are installed here, at
+ * the end of bootstrap, so EVERY fresh-run entry point (including a served
+ * `/function/:name` call through runExportedFunction) is capped identically —
+ * a served function must not run uncapped. Both installers are root-only
+ * (no-op in IPC subprocesses, whose budgets the parent guard owns) and are the
+ * outermost thing on the stack before any user body runs, so neither can be
+ * bypassed.
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c79d80ef5. Tightened both the docstring and the inline comment: the installers run at the END of initFreshExecCtx, after global init and top-level callback registration, so they're outermost before the entry NODE/FUNCTION body — not before the small amount of user code global-init expressions and top-level callback() blocks can run during bootstrap. Behavior is unchanged (runNode installed them after bootstrap too); only the wording overclaimed.

packages/agency-lang/lib/runtime/node.ts
egonSchiele: Fixed the comment in c79d80ef5 — it no longer claims rewindFrom re-asserts. It now states the split accurately: respondToInterrupts re-asserts via reinstallRootBudget, rewindFrom does neither (only createExecutionContext + restoreState), documented as a pre-existing gap out of scope here.

On ask (2), confirmed by reading the code: rewind.ts (rewindFrom) calls createExecutionContext + restoreState and neither reinstallRootBudget nor installRootBudget, so a rewound run trusts the checkpoint's serialized root guard (limit + spend). Two mitigating facts on the client-supplied-checkpoint angle: the serve HTTP surface exposes /function, /node, /resume only — there is NO /rewind route, so rewindFrom is not reachable from a remote client today; the stateless served-resume path (the one that takes a client checkpoint) goes through respondToInterrupts, which #796 already fixed. So the real residual is a LOCAL budgeted rewind (agency run --max-cost then debugger rewind): it runs under the checkpoint's limit rather than a re-asserted host limit — low-risk since local is trusted and that checkpoint's guard came from the same --max-cost, but inconsistent with the resume path. Happy to file a follow-up to add reinstallRootBudget in rewind.ts for parity; leaving it out of this PR as you noted.

egonSchiele: Filed the rewindFrom follow-up: #800 (add reinstallRootBudget in rewind.ts to mirror the resume path, with the scoping notes above).

PR #801, opened 2026-08-04
feat(serve): per-invocation usage seam for hosted spend tracking

egonSchiele:
## What

The agency-lang half of statelog's **Group 4 per-project spend tracking** — the "serve cost seam". Every post-execution serve outcome now carries a trusted, fresh **per-invocation usage** figure (priced cost + tokens + unknown-price count + completeness), read in-process by the host from the run agency-lang executed, so a project owner cannot under-report. Built from the reviewed plan (Revision 3), one task per commit.

## The two declarative boundaries

**Accounting** — one op `recordPaidUsageAt({ ctx, stack }, delta)` bills the branch's cost guards (reusing `billCharge`), merges the invocation meter, and relays the full delta upward exactly once when this process is a subprocess. All three paid sites route through it: the LLM completion, `addCost` (memory/image), and the IPC telemetry handler. So `pricedCost` is *all* trusted spend and is topology-independent — an `addCost` charge counts identically in-process and in a child. (`billCharge` no longer relays; the relay moved to this boundary.)

**Invocation** — `runNode` / `runExportedFunction` / `respondToInterrupts` split into an internal core returning a `ServedInvocationOutcome<T>` (`returned{value}` | `threw{error}` + usage snapshot) and a public wrapper that unwraps-or-rethrows (CLI/debugger contract unchanged). New `…ForServe` variants + generated `__invoke{Function,Node}ForServe` / `__respondToInterruptsForServe` hand the outcome to the adapters. User values and thrown errors are **never** mutated or wrapped — identity, `readCause`, stack, and the 402 classification are preserved.

## Key properties

- **Fresh, non-serialized meter** per execution context → resume-leg and concurrent isolation are structural (a resume leg starts at zero, not the checkpoint-carried `localCost`).
- **Two completeness axes, never conflated:** `pricingComplete` (derived from `unknownCostCallCount === 0`; a finite `0` is a known free price, only an *absent* price is unknown) vs `usageComplete` (goes permanently false on an abnormal subprocess termination, making `usage` a trusted lower bound, relayed once).
- **Usage on every post-execution outcome** (success/interrupt/402/failure/cancel) and **absent** on `/list`, 404, validation 400. The lifecycle boundary starts when the execution context exists, so an already-aborted signal or a setup failure still yields usage and runs cleanup; the snapshot is taken after cleanup.
- **Subprocess propagation:** the IPC wire carries a full `InvocationUsageDelta` (a legal delta may be zero-cost with tokens/unknown), validated per-field; a middle process re-relays it once so grandchild totals are exact.
- **`discoverExports` requires the serve invokers** — a bundle without them fails fast with a recompile-required error (older bundles were already unsupported for served functions).
- **MCP** unwraps the outcome to the raw value (or rethrows) and does not expose usage in v1; the standalone HTTP/MCP templates use the serve resume.

## Out of scope

Everything statelog-side (the `usage_events` table, spend API/CSV, execution-attempt id, reliability/idempotency) — that consumes this seam. Also: per-call/per-model breakdown (v1 is per-invocation totals); `runId` on the result (deferred).

## Verification

- `pnpm run typecheck` (source + tests + evals), `pnpm run lint:structure`, and `git diff --check` all clean.
- ~180 new/updated tests across 12 files (meter arithmetic; fresh/non-serialized context; the accounting op incl. topology-independence; IPC full-delta + exact grandchild totals + incompleteness marking; outcome cores incl. primitive/frozen throws, already-aborted, setup/cleanup failure; adapter mapping incl. 402 + usage-omission; discovery recompile-gate; a source-based end-to-end integration file). The touched-area suites pass (386 tests).
- Codegen verified to emit the three serve invokers; `serveCostSeam.integration` verified against a built `dist`. The `functionFrame.integration` suite needs the full `make` stdlib build, so it is left to CI (as before).

### DIFF
--- a/packages/agency-lang/lib/serve/http/adapter.ts
+++ b/packages/agency-lang/lib/serve/http/adapter.ts
@@ -1,5 +1,5 @@
 import http from "http";
-import type { ExportedItem, ExportedFunction, ExportedNode } from "../types.js";
+import type { ServedExportedItem, ServedExportedFunction, ServedExportedNode } from "../types.js";
 import { errorMessage, toArgs, parseJsonBody } from "../util.js";
 import { validateResumeBatch } from "../../runtime/interrupts.js";
 import { readCause } from "../../runtime/errors.js";
@@ -18,7 +18,7 @@ import {
 } from "./security.js";
 
 export type HandlerConfig = {
-  exports: ExportedItem[];
+  exports: ServedExportedItem[];
   logger: Logger;
   hasInterrupts: (data: unknown) => boolean;
   respondToInterrupts: (
@@ -135,7 +135,7 @@ function routeResultFor(
 }
 
 async function callFunction(
-  fn: ExportedFunction,
+  fn: ServedExportedFunction,
   body: unknown,
   logger: Logger,
 ): Promise<RouteResult> {
@@ -151,7 +151,7 @@ async function callFunction(
 }
 
 async function callNode(
-  node: ExportedNode,
+  node: ServedExportedNode,
   body: unknown,
   hasInterrupts: (data: unknown) => boolean,
   logger: Logger,
@@ -220,16 +220,16 @@ export function createHttpHandler(config: HandlerConfig): (
   // plain object would let a name like `/function/toString` resolve to an
   // inherited Object.prototype method, bypass the "unknown" 404 check, and
   // surface as a generic tool failure instead.
-  const functions: Record<string, ExportedFunction> = Object.assign(
+  const functions: Record<string, ServedExportedFunction> = Object.assign(
     Object.create(null),
     Object.fromEntries(
-      exports.filter((e): e is ExportedFunction => e.kind === "function").map((e) => [e.name, e]),
+      exports.filter((e): e is ServedExportedFunction => e.kind === "function").map((e) => [e.name, e]),
     ),
   );
-  const nodes: Record<string, ExportedNode> = Object.assign(
+  const nodes: Record<string, ServedExportedNode> = Object.assign(
     Object.create(null),
     Object.fromEntries(
-      exports.filter((e): e is ExportedNode => e.kind === "node").map((e) => [e.name, e]),
+      exports.filter((e): e is ServedExportedNode => e.kind === "node").map((e) => [e.name, e]),
     ),
   );
 
--- a/packages/agency-lang/lib/serve/http/adapter.ts
+++ b/packages/agency-lang/lib/serve/http/adapter.ts
@@ -141,7 +141,7 @@ async function callFunction(
 ): Promise<RouteResult> {
   let outcome: ServedInvocationOutcome<unknown>;
   try {
-    outcome = await fn.invoke(toArgs(body));
+    outcome = await fn.invokeServed(toArgs(body));
   } catch (err) {
     // A throw here is a pre-outcome internal failure (the core normally converts
     // agent errors into a threw-outcome), so there is no usage to report.
@@ -160,7 +160,7 @@ async function callNode(
   try {
     // The node receives its named args as a data object; the serve invoker maps
     // them to the node's params and returns the caller-facing data as the value.
-    outcome = await node.invoke(toArgs(body));
+    outcome = await node.invokeServed(toArgs(body));
   } catch (err) {
     return errorResult(err, logger, `node ${node.name}`);
   }
--- a/packages/agency-lang/lib/serve/http/adapter.ts
+++ b/packages/agency-lang/lib/serve/http/adapter.ts
@@ -1,5 +1,5 @@
 import http from "http";
-import type { ServedExportedItem, ServedExportedFunction, ServedExportedNode } from "../types.js";
+import type { ExportedItem, ServedExportedItem, ServedExportedFunction, ServedExportedNode } from "../types.js";
 import { errorMessage, toArgs, parseJsonBody } from "../util.js";
 import { validateResumeBatch } from "../../runtime/interrupts.js";
 import { readCause } from "../../runtime/errors.js";
@@ -17,7 +17,23 @@ import {
   makeGuardedRequestListener,
 } from "./security.js";
 
+/** PUBLIC config type (re-exported from `agency-lang/serve`) — kept
+ *  byte-compatible with the pre-seam contract: raw `ExportedItem`s and a
+ *  `respondToInterrupts` returning `Promise<unknown>`. Host apps that import
+ *  this type keep type-checking. The adapters run on the internal
+ *  `ServedHandlerConfig` below; `createServeHandler` builds that from a
+ *  discovered module. */
 export type HandlerConfig = {
+  exports: ExportedItem[];
+  logger: Logger;
+  hasInterrupts: (data: unknown) => boolean;
+  respondToInterrupts: (interrupts: unknown[], responses: unknown[]) => Promise<unknown>;
+};
+
+/** INTERNAL adapter config: the served contract the HTTP adapter actually runs
+ *  on — usage-bearing `ServedExportedItem`s and a `respondToInterrupts` that
+ *  returns a `ServedInvocationOutcome`. Not part of the public surface. */
+type ServedHandlerConfig = {
   exports: ServedExportedItem[];
   logger: Logger;
   hasInterrupts: (data: unknown) => boolean;
@@ -27,7 +43,7 @@ export type HandlerConfig = {
   ) => Promise<ServedInvocationOutcome<unknown>>;
 };
 
-export type HttpConfig = HandlerConfig & {
+export type HttpConfig = ServedHandlerConfig & {
   port: number;
   apiKey?: string;
   /** Interface to bind to. Default "127.0.0.1" (loopback only). */
@@ -209,7 +225,7 @@ async function resumeInterrupts(
 const FUNCTION_ROUTE = /^\/function\/([^/]+)$/;
 const NODE_ROUTE = /^\/node\/([^/]+)$/;
 
-export function createHttpHandler(config: HandlerConfig): (
+export function createHttpHandler(config: ServedHandlerConfig): (
   method: string,
   path: string,
   body: unknown,
--- a/packages/agency-lang/lib/serve/http/adapter.ts
+++ b/packages/agency-lang/lib/serve/http/adapter.ts
@@ -4,6 +4,10 @@ import { errorMessage, toArgs, parseJsonBody } from "../util.js";
 import { validateResumeBatch } from "../../runtime/interrupts.js";
 import { readCause } from "../../runtime/errors.js";
 import { formatBudgetExceeded } from "../../runtime/budgetExit.js";
+import type {
+  ServedInvocationOutcome,
+  InvocationUsageSnapshot,
+} from "../../runtime/invocationUsage.js";
 import type { Logger } from "../../logger.js";
 import {
   DEFAULT_HOST,
@@ -17,7 +21,10 @@ export type HandlerConfig = {
   exports: ExportedItem[];
   logger: Logger;
   hasInterrupts: (data: unknown) => boolean;
-  respondToInterrupts: (interrupts: unknown[], responses: unknown[]) => Promise<unknown>;
+  respondToInterrupts: (
+    interrupts: unknown[],
+    responses: unknown[],
+  ) => Promise<ServedInvocationOutcome<unknown>>;
 };
 
 export type HttpConfig = HandlerConfig & {
@@ -38,6 +45,12 @@ export type HttpConfig = HandlerConfig & {
 export type RouteResult = {
   status: number;
   body: unknown;
+  /** Per-invocation usage for the serve cost seam, present on every
+   *  post-execution outcome (success/interrupt/402/failure/cancel) and absent
+   *  on pre-execution results (/list, 404, validation 400). Read in-process by
+   *  the host (statelog); not part of the standalone HTTP body. */
+  usage?: InvocationUsageSnapshot["usage"];
+  usageComplete?: boolean;
 };
 
 function ok(value: unknown): RouteResult {
@@ -100,17 +113,41 @@ function errorResult(err: unknown, logger: Logger, what: string): RouteResult {
   return fail(TOOL_ERROR_MESSAGE);
 }
 
+/** Attach a per-invocation usage snapshot to any post-execution route result. */
+function withUsage(base: RouteResult, outcome: InvocationUsageSnapshot): RouteResult {
+  return { ...base, usage: outcome.usage, usageComplete: outcome.usageComplete };
+}
+
+/** Map one served outcome to a route result: `renderReturned` shapes the success
+ *  body (function → `ok`; node/resume → interrupt-aware), and a threw-outcome
+ *  runs the existing 402/generic classification on the ORIGINAL error (identity
+ *  preserved, so `readCause` still recognizes a guard trip). Usage is attached
+ *  once, here, for every branch. */
+function routeResultFor(
+  outcome: ServedInvocationOutcome<unknown>,
+  opts: { renderReturned: (value: unknown) => RouteResult; logger: Logger; what: string },
+): RouteResult {
+  const base =
+    outcome.status === "returned"
+      ? opts.renderReturned(outcome.value)
+      : errorResult(outcome.error, opts.logger, opts.what);
+  return withUsage(base, outcome);
+}
+
 async function callFunction(
   fn: ExportedFunction,
   body: unknown,
   logger: Logger,
 ): Promise<RouteResult> {
+  let outcome: ServedInvocationOutcome<unknown>;
   try {
-    const result = await fn.invoke(toArgs(body));
-    return ok(result);
+    outcome = await fn.invoke(toArgs(body));
   } catch (err) {
+    // A throw here is a pre-outcome internal failure (the core normally converts
+    // agent errors into a threw-outcome), so there is no usage to report.
     return errorResult(err, logger, `function ${fn.name}`);
   }
+  return routeResultFor(outcome, { renderReturned: ok, logger, what: `function ${fn.name}` });
 }
 
 async function callNode(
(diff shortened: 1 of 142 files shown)


packages/agency-lang/lib/serve/types.ts
```diff
@@ -11,22 +12,25 @@ export type ExportedFunction = {
...
+   * Invoke the function for a single request, given its named arguments, and
+   * return a `ServedInvocationOutcome`: the raw function value (on `returned`)
+   * or the identical thrown error (on `threw`), plus a per-invocation usage
+   * snapshot. Populated by `discoverExports` from the compiled module's
+   * generated `__invokeFunctionForServe`, which runs the body inside a
+   * node-grade execution frame. An adapter unit test may supply a plain-JS
+   * invoke that returns an outcome directly.
    */
-  invoke: (namedArgs: Record<string, unknown>) => Promise<unknown>;
+  invoke: (namedArgs: Record<string, unknown>) => Promise<ServedInvocationOutcome<unknown>>;
```
egonSchiele: `ExportedFunction` and `ExportedNode` are part of the documented `agency-lang/serve` public surface (`public.ts` re-exports both), so changing `invoke` from raw return/throw semantics to `ServedInvocationOutcome` is a breaking API change for host apps that construct or consume these values directly. The node signature also changes from positional args to a named-data object. Please preserve the existing public `invoke` contract and add a separate internal/outcome-producing member for the adapters (or stop exporting these types as public API and provide a migration/major-version boundary). The generated `*ForServe` helpers can remain the source of the outcome without exposing that representation through the existing public method.

egonSchiele:
Two blocking issues from this pass:

1. The inline comment identifies a breaking change to the public `agency-lang/serve` `ExportedFunction` / `ExportedNode.invoke` contracts. Keep the existing public raw-return API and expose the usage-bearing outcome through a separate adapter/internal seam.
2. The new lifecycle boundary still skips `execCtx.cleanup()` when `statelogClient.flush()` rejects (`finalizeExecCtx` awaits the flush before cleanup). Put cleanup in a `finally` so every invocation whose context was created actually releases it, while retaining the flush failure as the outcome when execution succeeded.

CI also needs to be green before merge. Both build jobs currently fail for two PR-caused reasons: generated TypeScript fixtures were not rebuilt after changing `imports.mustache`, and `lib/stdlib/image.test.ts`'s ALS test context has no `invocationUsage`, so paid image calls now fail at `recordPaidUsageAt` before the guard assertions. Please rebuild fixtures with the repository-required `make fixtures`/`make` flow and update that test context with a real usage meter (rather than weakening the production accounting boundary).

I reviewed the accounting chokepoint, IPC usage/incompleteness relay and settlement paths, generated serve invokers, and HTTP/MCP outcome handling. I did not find another blocking correctness issue in those paths.

egonSchiele (commented):
Reviewed against statelog's serve-cost-seam contract. This is a well-architected seam — the fresh-meter isolation, the outcome envelope, the no-mutation guarantee, and the subprocess relay are all solid and genuinely tested. **6 of 7 load-bearing properties are CONFIRMED; one is a real gap.** Verdict: **needs one change before I can pin it.**

## ✅ Confirmed (with evidence)

- **Fresh, non-serialized meter per execCtx.** `invocationUsage` is re-created in `createExecutionContext` (`state/context.ts:401`), untouched by `restoreState`, and not on the cached handler/module/global/ALS/serialized stack. A resume leg starts at zero (fresh execCtx via `createExecutionContext(interrupt.runId)`), so resume-per-leg and concurrent isolation are *structural*, not incidental. `context.invocationUsage.test.ts` pins it.
- **Usage on every post-execution outcome, absent pre-execution.** `withUsage` attaches on both `returned` and `threw` for `/function`/`/node`/`/resume`; `/list`, 404, and validation-400 return bare results. Good.
- **Envelope transport, snapshot after cleanup.** Cores settle into `RawOutcome`, `finishServedInvocation` runs cleanup then snapshots — the adapter reads usage off the envelope, never from a torn-down ctx/ALS.
- **No mutation of user values/errors; 402 preserved.** `unwrap…` rethrows by identity; the original error still flows to `errorResult`, so a root-budget trip still classifies as a typed 402. `invocationOutcome.test.ts` (frozen/primitive throws, budget-trip-carries-cost) is convincing.
- **Subprocess propagation.** Full `InvocationUsageDelta` on the wire, relayed exactly once per level (`recordPaidUsageAt` → `sendInvocationUsageToParent`; middle process re-relays once), and `pricedCost` is topology-independent because `addCost` routes through the same boundary. `ipc.test.ts` covers grandchild exactness + incompleteness.
- **`usageComplete` lower-bound signal.** Nice addition — an abnormal child termination flips `usageComplete=false` (idempotent, relayed once) rather than silently dropping unsent child spend. That's the right shape.

## 🔴 Critical — provider-attempt lifecycle for unknown cost is not implemented

The contract's provider-attempt state machine requires that a provider attempt **dispatched and then timed-out / cancelled / failed after dispatch** (possible spend, no price metadata) counts in `unknownCostCallCount`, so `pricingComplete` goes false — while a failure proven *before* dispatch counts nothing, and each retry is a separate attempt.

As implemented, only the **completed-call** case is handled: `accountCompletionUsage` runs solely on the success path (`prompt.ts:769`), and `completionUsageDelta` (`invocationUsage.ts:111`) sets `unknownCostCallCount` only for a *returned-but-unpriced* completion. The `catch` around `dispatchWithRetry` (`prompt.ts:672-719`) emits telemetry and `throw err` — it never touches the meter. So a timeout-after-dispatch, cancel-after-dispatch, post-dispatch provider error, or failed-retry-then-success reports `pricingComplete: true` for an invocation that may have incurred untracked spend — and `usageComplete` doesn't cover it either (that axis is for abnormal *subprocess* termination, not an in-process dispatch failure). That silently presents a non-authoritative dollar figure as authoritative, which is exactly what the completeness axis exists to prevent.

**Fix:** begin an attempt immediately before each `dispatchLLMRequest` (each retry = a new attempt); resolve it priced/known-unpriced on completion; and on any dispatched-but-unresolved attempt that throws/cancels, add one to `unknownCostCallCount` (a `try/finally` around dispatch that increments unknown iff the attempt didn't resolve). Add the contract's missing tests: timeout-after-dispatch, cancel-after-dispatch, and failed-retry-then-success each `+1`; pre-dispatch validation failure `+0`.

## 🟡 Important — `discoverExports` hard-gate is a wider break than the PR body states

`discovery.ts:89-97` now throws if the three `…ForServe` invokers are absent — which rejects *entire* older compiled bundles, including node-only/resume bundles that previously served fine via the raw exports, not just the already-unsupported served-functions path. **For statelog specifically this is fine** — I confirmed it recompiles source on every cold build (`serveHost.ts:115` `compileSource(...)`), so a fresh compile always emits the invokers. But other serve consumers with pre-compiled bundles would break. Either narrow the requirement (require the function invoker only when functions are exported, etc.) or document that served bundles must be recompiled.

## 🔵 Minor

- `usageComplete` being a **sibling** of `usage` (not nested) is easy for a consumer to miss and thereby treat a lower-bound total as authoritative. Worth calling out prominently in the exported types/docs. (statelog will persist and honor it — see below.)
- Meter fields accumulate with plain `+=`; fine for realistic magnitudes, noting only for completeness.

## Note for the consumer (statelog)

The result shape I'll code Task 3 against — flagging because `usageComplete` is new vs. the original handoff and is load-bearing:

```
{ status, body,
  usage?: { pricedCost, inputTokens, outputTokens, unknownCostCallCount, pricingComplete },
  usageComplete?: boolean }   // sibling; false ⇒ usage is a trusted LOWER BOUND
```

statelog will store `usageComplete` alongside the derived completeness and treat the dollar figure as a lower bound whenever either is false.

Once Critical #1 lands (and ideally the `discovery` note is addressed or documented), this is exactly what statelog needs — I'll pin the release and run the gate against it.

packages/agency-lang/lib/serve/types.ts
egonSchiele: Fixed in 77ca16877. Restored the public ExportedFunction/ExportedNode.invoke to its original raw value-or-throw contract (node stays positional); added a separate internal invokeServed member that discoverExports wires to the *ForServe invokers, and the HTTP/MCP adapters consume invokeServed. So the public serve surface is unchanged — no host-app break, no major-version boundary needed.

egonSchiele:
Addressed all three in 77ca16877.

**Critical (provider-attempt unknown cost)** — you were right: `accountCompletionUsage` only ran on the success path, so a request dispatched then timed-out/cancelled/errored-after-dispatch reported `pricingComplete: true`. Each dispatch (each retry a fresh attempt) now runs through a `meteredDispatch` wrapper: a dispatched-but-**unresolved** attempt records one `unknownCostCallCount` (no cost, no tokens) so `pricingComplete` goes false; a completion resolves the attempt and is priced at the existing success site; a failure proven **before** dispatch never enters the wrapper (counts nothing). Tests cover resolved → +0, unresolved-throw → +1, failed-retry-then-success → +1, plus the `recordUnknownCostAttempt` accounting (relays the unknown delta upward too).

**Important (discovery gate breadth)** — narrowed to per-kind: `discoverExports` now requires `__invokeFunctionForServe` only when functions are exported and `__invokeNodeForServe` only when nodes are exported (a node-only bundle needs just the node one), with a clearer recompile-required message. Also documented that served bundles must be recompiled. (And per the inline thread, the public `invoke` contract is preserved via a separate internal `invokeServed`, so pre-compiled bundles only need recompilation for the *served* path, not for direct `invoke` consumption.)

**Minor (usageComplete prominence)** — strengthened the exported type doc (it's a **sibling** of `usage`, not nested; `false` ⇒ the whole figure incl. `pricedCost` is a trusted lower bound) and the dev docs.

The result shape you'll code against is confirmed:
```
{ status, body,
  usage?: { pricedCost, inputTokens, outputTokens, unknownCostCallCount, pricingComplete },
  usageComplete?: boolean }   // sibling; false ⇒ usage is a trusted LOWER BOUND
```

Verification: typecheck (3 configs), lint:structure, diff --check clean; the touched-area suites pass (198 tests). `functionFrame.integration` still needs the full `make` stdlib build (CI); I re-verified codegen emits the invokers and `serveCostSeam.integration` passes against a built `dist`.

packages/agency-lang/lib/serve/types.ts
```diff
@@ -11,22 +12,30 @@ export type ExportedFunction = {
...
+   * apps that construct/consume `ExportedFunction` directly depend on.
    */
   invoke: (namedArgs: Record<string, unknown>) => Promise<unknown>;
+  /**
+   * INTERNAL, for the serve adapters: the same execution as `invoke`, but the
+   * value-or-error is returned inside a `ServedInvocationOutcome` alongside a
+   * per-invocation usage snapshot (never mutating the value/error). Populated by
+   * `discoverExports` from the compiled module's `__invokeFunctionForServe`.
+   */
+  invokeServed: (namedArgs: Record<string, unknown>) => Promise<ServedInvocationOutcome<unknown>>;
```
egonSchiele: This still changes the public structural contract: `ExportedFunction` is exported from `agency-lang/serve`, and `invokeServed` is now a **required** property. Existing host code that constructs an `ExportedFunction`/`ExportedNode` with the previously valid fields will stop type-checking, even though `invoke` itself was restored. Please keep the public types unchanged and introduce non-exported adapter types such as `ServedExportedFunction = ExportedFunction & { invokeServed: ... }` / `ServedExportedNode`, then have discovery and the adapters use those internal types. Calling a required property “INTERNAL” in a public exported type does not make it non-breaking.

egonSchiele:
Delta review of `77ca16877`:

- The provider-attempt accounting and per-kind discovery gate changes look correct. The new tests cover the accounting helper/retry arithmetic, and `invoke` itself has its old semantics again.
- One API blocker remains; I left it inline: adding required `invokeServed` fields to the exported public structural types is still a breaking change for hosts that construct those values.
- The cleanup blocker from the previous review is **not addressed**: `finalizeExecCtx` still does `await statelogClient.flush(); execCtx.cleanup();`, so a rejected flush skips cleanup. `cleanup()` needs to run from a `finally` while preserving the flush error.
- The prior CI fixes are also absent from this delta: neither generated fixtures nor `lib/stdlib/image.test.ts` changed. The new CI run is pending, but the old failures are deterministic, so I expect both build jobs to fail again unless another workflow step now regenerates/patches them.

I do not see another blocker in the new provider-attempt or discovery changes.

egonSchiele (commented):
Re-reviewed the fix commit (`77ca16877`) against my earlier findings. **Both resolved — this is ready to pin.**

## 🔴 Critical (provider-attempt lifecycle) — RESOLVED ✅

`meteredDispatch(ctx, targetStack, () => dispatchLLMRequest(...))` now wraps each dispatch attempt inside `runWithRetry`, with a `try/finally` that calls `recordUnknownCostAttempt` (`+1 unknownCostCallCount`, 0 cost/tokens) when the attempt doesn't resolve — so a timeout / cancel / post-dispatch provider error flips `pricingComplete` to false and rethrows the original error by identity. Because `runWithRetry` invokes the metered lambda once per attempt, each retry is its own attempt. The new `prompt.test.ts` cases pin exactly the contract: resolved → 0 unknown / `pricingComplete: true`; unresolved throw → 1 unknown / `pricingComplete: false` / `rejects.toBe(err)`; failed-retry-then-success → exactly 1 unknown. This closes the "authoritative dollar figure that silently omits a possibly-billed attempt" hole.

## 🟡 Important (discovery gate) — addressed ✅

Splitting the public `invoke` (raw value, with an `agencyFunction.invoke` fallback) from the internal `invokeServed` is the right separation, and keeping the hard-throw-with-`RECOMPILE_HINT` when a served bundle lacks the `…ForServe` invokers is a defensible call: usage reporting is mandatory for serving, so a served bundle must be recompiled. Fine for statelog — it recompiles source on every cold build, so it always emits the invokers. Good enough; the clear error + docs make the requirement discoverable for other consumers.

Also confirmed the `usageComplete`-is-a-sibling doc-comment landed on `InvocationUsageSnapshot` — thanks.

## 🔵 One optional note (non-blocking)

`meteredDispatch` treats *any* throw out of `dispatchLLMRequest` as an unknown-cost attempt. If that function can fail before the request is actually put on the wire (pre-flight build/validation), those would be counted as unknown too. That's the *safe* direction — it can only make `pricingComplete` conservatively false, never hide real spend — so it's fine to leave; flagging only in case you'd rather scope the metered boundary to the actual network send.

Once this merges and you cut the release, give me the version — I'll pin it, run statelog's Task 1 gate (the nine-property verifier) against it, and wire up Task 3 (the serve-path recorder), persisting `usageComplete` alongside the derived `pricingComplete` and treating the dollar figure as a lower bound when either is false.

packages/agency-lang/lib/serve/types.ts
egonSchiele: Fixed in 18a9bbbc8. Reverted invokeServed off the public ExportedFunction/ExportedNode entirely — they are byte-identical to before this PR now. The adapters use NON-exported supertypes ServedExportedFunction/ServedExportedNode/ServedExportedItem (= Exported* & { invokeServed }); discoverExports returns those and HTTP/MCP/cli consume them. So a host constructing an ExportedFunction/ExportedNode with the previously valid fields type-checks unchanged.

egonSchiele:
All three remaining blockers fixed in 18a9bbbc8.

**Public type break (`invokeServed`)** — reverted it off the exported `ExportedFunction`/`ExportedNode`; they're byte-identical to before this PR. The usage-bearing invoker lives on non-exported supertypes `ServedExportedFunction`/`ServedExportedNode`/`ServedExportedItem` (`Exported* & { invokeServed }`) that discovery returns and the adapters consume. No public-surface change; hosts that construct these values type-check unchanged. (Reply on the inline thread.)

**Cleanup blocker** — `finalizeExecCtx` now does `try { await statelogClient.flush() } finally { execCtx.cleanup() }`, so a rejected flush no longer skips cleanup; the flush rejection still propagates and (via `finishServedInvocation`) becomes the outcome error when execution otherwise succeeded.

**CI** — both causes addressed:
- Regenerated the compiled test fixtures with `make fixtures` (they were stale after the `imports.mustache` serve-invoker change; ~90 `tests/typescript{Generator,Builder}` `.mjs` now bake the `…ForServe` exports).
- Gave `lib/stdlib/image.test.ts`'s ALS context a real `InvocationUsageMeter` so paid image calls exercise the accounting boundary — fixing the test context rather than weakening the production `recordPaidUsageAt`.

Verification: typecheck (3 configs), lint:structure, diff --check clean; the touched-area suites pass (196 tests). `functionFrame.integration` still needs the full `make` stdlib build (CI); I re-verified codegen and `serveCostSeam.integration` against a built `dist`.

On your optional note (`meteredDispatch` counting a pre-wire build/validation throw as unknown): agreed it's the safe direction — it can only make `pricingComplete` conservatively false, never hide spend — so I've left the metered boundary at the dispatch call.

packages/agency-lang/lib/serve/http/adapter.ts
```diff
@@ -14,10 +18,13 @@ import {
 } from "./security.js";
 
 export type HandlerConfig = {
-  exports: ExportedItem[];
+  exports: ServedExportedItem[];
```
egonSchiele, line 37: There is one remaining public-type break here. `HandlerConfig` is explicitly re-exported from `agency-lang/serve`, and both of its load-bearing fields changed: `exports` now requires internal `ServedExportedItem`s, and `respondToInterrupts` now requires a `ServedInvocationOutcome` rather than `Promise<unknown>`. So hosts using the exported config type still stop type-checking even though `ExportedFunction`/`ExportedNode` themselves are restored. Please keep the exported `HandlerConfig` byte-compatible and introduce a separate non-public adapter config for `createHttpHandler`/`startHttpServer` (the same pattern now used for `ServedExportedItem`), or remove `HandlerConfig` from the public surface only in a major-version change.

egonSchiele:
Delta review of `18a9bbbc8`:

- ✅ `ExportedFunction` / `ExportedNode` are restored; the usage-bearing variants are now separate adapter types.
- ✅ `execCtx.cleanup()` now runs in `finally` when statelog flushing rejects, while the flush error still propagates.
- ✅ The image test context has a real meter and the generated TypeScript fixtures were rebuilt.
- 🔴 One public compatibility issue remains, noted inline: the separately exported `HandlerConfig` still changed to the internal served-export/outcome contracts.

The new CI run is still pending. I found no other blocker in this fix delta.

packages/agency-lang/lib/serve/http/adapter.ts
egonSchiele: Fixed in 3cb066bb3. Restored the exported HandlerConfig to its pre-seam shape — exports: ExportedItem[] and respondToInterrupts: (i,r) => Promise<unknown> — so hosts importing the config type keep type-checking. The served contract (ServedExportedItem[] + Promise<ServedInvocationOutcome>) moved to a non-exported ServedHandlerConfig that createHttpHandler/startHttpServer/HttpConfig run on, same pattern as ServedExportedItem. No public-surface change.

egonSchiele:
Re-reviewed `3cb066bb3`. The remaining compatibility issue is resolved: public `HandlerConfig` is restored to its pre-seam shape, while the usage-bearing contract is isolated in the non-exported `ServedHandlerConfig` used by the adapters. The fix is localized and `diff --check` is clean. I found no remaining code-review blockers. Typecheck and lint pass; the longer CI jobs are still running.

PR #802, opened 2026-08-04
feat(serve): per-model cost breakdown on the serve cost seam

egonSchiele:
## Summary

Gives every hosted serve invocation a per-model breakdown of priced cost and input/output tokens, plus a `modelAttributionComplete` trust signal. This is the agency-lang follow-up to the serve cost seam (#801) — the statelog Group 4 dependency for per-project spend attribution by model.

The breakdown rides the **same** delta and IPC wire as the flat total, so it includes subprocess-relayed spend (which the process-local `__tokenStats.models` behind `/cost` cannot cross) and always reconciles to the authoritative `pricedCost`.

## Design

- **Discriminated attribution** on the delta and wire — `{kind:"model",model}` (a completion) or `{kind:"unattributed"}` (`addCost`: memory / image). A scalar model field could not tell "deliberately model-less" from "model lost by a version-skewed child": the immediately-preceding #801 runtime relays usage deltas with no attribution at all.
- **Meter** buckets each measurable charge into a `models` map or an `unattributed` row; the snapshot emits copies plus `modelAttributionComplete`. `merge` stays pure — the provenance signal is raised at the IPC boundary, not inside it.
- **`modelAttributionComplete`** — a third completeness axis (distinct from `pricingComplete` = price availability and `usageComplete` = telemetry delivery), tripped only when a *measurable* child delta arrives over IPC with no attribution (an older #801 runtime, or the legacy `{costUsd}` handler), and relayed upward once like `usageComplete`. `addCost` and normal completions always carry an explicit attribution and never trip it.
- **Model identity** — `resolveCompletionModel(completion.model, clientConfig.model)`: provider-reported → configured → `"unknown model"`, used verbatim as the aggregation key. A real model named `"unknown model"` is an ordinary row, never `unattributed`.
- **Reconciliation** — rows + `unattributed` reconcile to the flat total within `usageReconcileTolerance(pricedCost)` = `max(1e-9, 1e-9·|pricedCost|)` (relative+absolute; float ulp drift scales with magnitude); tokens reconcile exactly. The flat total stays authoritative — a host bills `pricedCost`, not the row sum.
- **Compatibility** — the new `models` / `unattributed` / `modelAttributionComplete` fields are optional on the public `InvocationUsage` type; the current runtime always emits them. A host seeing them absent is reading an older runtime and must fall back to the flat totals, never treat absence as zero spend.

Cost per model is a single total — input-vs-output dollars are not split (the provider gives one `totalCost` through this seam).

## Scope / non-goals

- Public serve wire, CLI, and MCP semantics unchanged; #801's flat-total accumulation is untouched.
- No per-model split of unknown-cost attempts or cache tokens (explicitly out of scope for v1).

## Testing

147 deterministic tests across the 7 touched files (meter/delta arithmetic and reconciliation, attribution normalization, IPC send + receive + provenance, model-identity precedence, and serve-outcome surfacing including thrown / interrupt / resume). No LLM calls. `make` (full `tsc` typecheck + build) passes clean. The full agency suite runs in CI.

Design docs: `docs/superpowers/specs/2026-08-04-per-model-cost-breakdown-design.md` and `docs/superpowers/plans/2026-08-04-per-model-cost-breakdown.md`.


packages/agency-lang/lib/runtime/ipc.test.ts
```diff
@@ -403,7 +465,9 @@ describe("handleInvocationUsageMessage (full delta)", () => {
     const send = vi.fn(() => true);
     process.send = send as any;
     const { session } = makeUsageSession();
-    const delta = { pricedCost: 0.15, inputTokens: 3, outputTokens: 1, unknownCostCallCount: 1 };
+    // A current-runtime relay carries attribution, so provenance detection does
+    // not fire and only the usage delta is re-relayed (exactly once).
+    const delta = { pricedCost: 0.15, inputTokens: 3, outputTokens: 1, unknownCostCallCount: 1, attribution: { kind: "model", model: "opus" } };
     handleInvocationUsageMessage(session, { type: "invocationUsage", ...delta });
```
egonSchiele, line 471: The PR currently fails the typecheck here: `delta.attribution.kind` widens to `string`, so the spread object is not assignable to `IpcInvocationUsageMessage` (`TS2345` in the CI typecheck job). Please preserve the discriminator, e.g. declare `delta` with `satisfies InvocationUsageDelta` or use `kind: "model" as const`, and rerun `pnpm run typecheck`.

packages/agency-lang/lib/runtime/invocationUsage.ts
```diff
@@ -76,19 +124,56 @@ function asCount(value: unknown): number {
...
 
   merge(delta: InvocationUsageDelta): void {
     this.pricedCost += delta.pricedCost;
     this.inputTokens += delta.inputTokens;
     this.outputTokens += delta.outputTokens;
     this.unknownCostCallCount += delta.unknownCostCallCount;
+
+    // A charge with money or tokens lands in exactly one row — its model, or
+    // the unattributed bucket (kind "unattributed" OR absent). This is what
+    // keeps sum(models) + unattributed reconciled to the flat totals. `merge`
```
egonSchiele, line 176: The new exact-token reconciliation guarantee does not hold for every count this meter accepts. `isValidCount` still permits integers outside JavaScript’s safe range, so flat accumulation and regrouped row accumulation can differ: merging `alpha:1`, `beta:10`, `alpha:2**53` yields flat `9007199254741004` but row sum `9007199254741002`. This is reachable through untrusted IPC because `Number.isInteger(2 ** 53)` is true. Please change count validation to `Number.isSafeInteger(value)` and add a rejection/reconciliation test, or weaken the exact-token contract and define a token tolerance.

packages/agency-lang/lib/runtime/ipc.test.ts
egonSchiele: Fixed in d5f458ce8. The extracted `const delta` widened `kind` to `string`; changed to `kind: "model" as const` so it stays assignable. `pnpm run typecheck` (incl. tsconfig.tests.json) is now clean — my local `make` only ran the main `tsc`, which excludes test files, so it missed this.

packages/agency-lang/lib/runtime/invocationUsage.ts
egonSchiele: Fixed in d5f458ce8. `isValidCount` now uses `Number.isSafeInteger`, so a count at or above 2**53 from untrusted IPC clamps to 0 instead of being accepted and losing precision on addition. Kept the exact-token contract (rather than a token tolerance) since no real completion reports that many tokens. Added a normalize+meter rejection test.

egonSchiele: Follow-up: `Number.isSafeInteger` on each incoming field does not fully restore the exact reconciliation invariant, because individually safe deltas can accumulate beyond the safe range in different groupings. Concrete counterexample (every input is safe): merge `alpha:2`, `beta:2`, `beta:Number.MAX_SAFE_INTEGER`. The flat total becomes `9007199254740996`; beta’s row becomes `9007199254740992`, and adding alpha yields row sum `9007199254740994`. Please either guard the **accumulated** flat and row totals so the snapshot never claims exactness outside the safe range, use an exact internal representation, or explicitly scope/weaken the exact-token contract. The new test only rejects one unsafe input and misses accumulation overflow.

egonSchiele: Fixed in 714264930 by scoping the contract (your third option) rather than adding bignum accounting. You are right that per-field `Number.isSafeInteger` bounds each incoming count but not the accumulated total. I chose to weaken the claim honestly because token counts here are per-model attribution for display, not billing: `pricedCost` is the money, and it already reconciles within `usageReconcileTolerance` with the flat total authoritative. Accumulating past 2**53 tokens is physically impossible in a real run and only reachable by a malicious subprocess relaying absurd counts, so exact internal representation for token display has no real user. The doc on `InvocationUsage.models`, the `isValidCount` comment, the dev doc, and the design spec now all state: token reconciliation is exact WITHIN the safe-integer range; beyond it (malicious IPC only) it is best-effort like cost, and the flat total stays authoritative for both. No test asserts token exactness outside the safe range (the reconciliation helper checks cost; token assertions use small values).

egonSchiele: Re-reviewed 714264930: this now resolves my concern. The implementation rejects individually unsafe counts, and the public/runtime/design documentation consistently scopes exact token reconciliation to safe accumulated totals while treating malicious overflow as best-effort attribution. The discriminator fix remains correct, and the latest lint and typecheck jobs are green. No further findings from this pass.

PR #805, opened 2026-08-05
agency run: position decides who a flag belongs to

egonSchiele:
## What this does

`agency run` decides who a flag belongs to by where it sits, the way `node --inspect script.js --verbose` does. Agency's flags and its own filename come first; everything after the filename is the program's command line.

```
agency run --policy strict greet.agency --name alice
             agency's                     program's
```

`--` is required for nothing.

## Why

A program flag used to need `--`, and the separator meant the opposite thing once the program was compiled:

```
agency run greet.agency -- --name alice    # `--` required
node greet.js --name alice                 # `--` must NOT be there
node greet.js -- --name alice              # silently wrong: std::args reads `--`
                                           # as "stop parsing flags", so the
                                           # default wins
```

One rule now covers both.

## Position always wins, and says so

An agency flag written after the filename goes to the program, because that is what position means. Every mainstream interpreter does this silently — I confirmed node, python, ruby and perl locally, and [Deno documents it](https://docs.deno.com/runtime/getting_started/command_line_interface/) as a trap.

Agency warns anyway, because its misplaced flags fail in the other direction. A misplaced `node --inspect` announces itself; a misplaced `--max-cost` runs uncapped and nothing notices.

```
$ agency run greet.agency --max-cost 5
Warning: --max-cost went to your program, not to agency.
  Agency flags go before the filename: agency run --max-cost ... greet.agency
  Write -- before it to silence this:  agency run greet.agency -- --max-cost ...
```

`--` survives only to silence that, for a program that genuinely owns a colliding name.

## How

Commander's `passThroughOptions` needs `enablePositionalOptions` on the root, which is not scoped to one subcommand: it breaks `agency run -c cfg.json f.agency` and every nested command reading a parent's option (`agency label ingest --store x` fails outright). Both reproduced.

So `lib/cli/commandLine.ts` inserts `--` before commander parses. `agency agent` already worked this way, and now shares the code: each command supplies a policy.

```ts
{ command: "run",   ownedPositionals: 1, options: run + root, warnOnCollision: true  }
{ command: "agent", ownedPositionals: 0, options: agent only, warnOnCollision: false }
```

Flag names and arity come from commander's own option metadata, so no second list can drift.

## Breaking changes

**Agency flags after the filename now go to the program.** `agency run f.agency --policy x` becomes `agency run --policy x f.agency`. Warned, not silent.

**`--trace [file]` is now `--trace` plus `--trace-file <path>`.** An optional value still swallows the next word, so the old shape had no working spelling under the position rule — and none before it either: `agency run --trace f.agency` failed with "missing required argument" on main, though `docs/dev/trace.md` documented exactly that. Both spellings work now.

## Callers migrated

Three spawn suites, plus two the first sweep missed and CI caught:

- `tests/integration/statelog/test.mjs` — `-c agency.json` after the filename, so eight scenarios ran without their observability config
- `tests/integration/eval-run/test.mjs` — `--policy approve-all` after the filename; worse than a lost flag, since that writer reads `args()[0]`, which became `"--policy"`

Both suites pass locally.

## Testing

- `lib/cli/commandLine.test.ts` — 16 cases: both policies, all four flag spellings, `-print` forwarding without a false warning, attached and clustered root flags before the subcommand, bare and explicit trace
- `scripts/agency.test.ts` — the agent policy keeps budget flags on agency's side
- `tests/integration/cli/test.mjs` Test 8 — real binary: new form, separator form, root flags before the filename, the warning, `-print` silence, bare `--trace` and `--trace-file` writing files, and `eval label ingest --store` proving nested parent options still resolve

Green locally: those suites plus the three spawn suites, `labelCommand.test.ts`, and the statelog, eval-run and CLI integration suites.

## Docs

`docs/site/cli/run.md`, `stdlib/args.agency` (with `docs/site/stdlib/args.md` regenerated by `make`), `docs/site/guide/policies.md`, `docs/site/guide/guards.md`, `docs/site/cli/trace-and-bundle.md`, and `docs/dev/cli-arguments.md`.


packages/agency-lang/scripts/agency.ts
```diff
@@ -206,7 +210,13 @@ export function createProgram(deps: CliDependencies = {}): Command {
...
     .version("0.0.105")
     .option("-v, --verbose", "Enable verbose logging during parsing")
-    .option("-c, --config <path>", "Path to agency.json config file");
+    .option("-c, --config <path>", "Path to agency.json config file")
+    // Position decides who a flag belongs to, the way `node --flag script.js
+    // --script-flag` does: agency's own flags go before the subcommand, and
+    // `run` forwards everything after the filename to the program. Without
+    // this, `-v` and `-c` would still be claimed by agency wherever they sit,
+    // and `run` could not pass them on.
+    .enablePositionalOptions();
```
egonSchiele: `enablePositionalOptions()` applies at the root, so this also stops root options from working between the subcommand and filename. For example, `agency run -c custom.json greet.agency` and `agency run -v greet.agency` currently work and satisfy the PR’s stated rule that Agency flags go before the filename; with this change Commander rejects them as unknown options. They only work in the much less natural form `agency -c custom.json run greet.agency`. This also changes option parsing for every other subcommand, as the existing warning in `labelCommand.ts` notes. Please preserve `-c`/`-v` before the run filename (and add an acceptance test), rather than requiring root flags to move before `run` as an undocumented second boundary.

packages/agency-lang/lib/cli/forwardedArgs.ts
```diff
@@ -0,0 +1,48 @@
...
+    ``,
+    `  Put it before the file:      agency run ${flag} ... ${input}`,
+    `  Or hand it to the program:   agency run ${input} -- ${flag} ...`,
+  ].join("\n");
+}
+
+/** `--max-cost=5` names the same flag as `--max-cost 5`. */
+function flagNameOf(arg: string): string {
+  const equals = arg.indexOf("=");
+  return equals === -1 ? arg : arg.slice(0, equals);
```
egonSchiele: This normalization misses Commander’s valid attached short-option forms. `-cfoo.json` is parsed as `--config foo.json`, and `-iv` is parsed as `-i -v`, but after the filename `flagNameOf()` returns `-cfoo.json`/`-iv`, so neither matches the Agency flag list and both are silently forwarded. That violates the safety invariant this guard exists to enforce, especially for config or guard-related short aliases. Please recognize attached required-value short options and short-option clusters using the actual Commander option metadata (a flat string list loses the arity needed to distinguish them), with tests for at least `-cfoo.json` and `-iv`.

egonSchiele:
The user-facing docs cannot be deferred for this PR: they currently teach the exact behavior this change removes. `docs/site/cli/run.md` says `--` is required, says the separator-less form fails, and says “Only agency run needs --”; the `std::args` source documentation likewise teaches `agency run greet.agency -- --name ...`, which would keep working but defeats the new consistent invocation model. Please update the manual CLI page and the source doc comments in `stdlib/args.agency`, then regenerate `docs/site/stdlib/args.md` through the normal stdlib documentation build rather than hand-editing that generated page. Shipping a breaking CLI syntax change while the canonical run/args references explicitly describe the old contract will send users to the wrong invocation.

packages/agency-lang/scripts/agency.ts
egonSchiele: Confirmed and fixed by dropping the approach. I reproduced it: with root positional parsing on, `run -c cfg.json greet.agency` gives `unknown option '-c'`.

The boundary is now inserted before commander parses, so root option hoisting is untouched and `-c`/`-v` work where they read most naturally. Acceptance tests added for `-v`, `-c custom.json`, and the attached `-ccustom.json` spelling, all between the subcommand and the filename.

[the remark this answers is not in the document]
egonSchiele: Confirmed, and it is worse than "may stop receiving" — I reproduced `agency label ingest --store /tmp/s` failing with `unknown option '--store'`. Since that pattern is used by every nested command that reads a parent option, scoping was not possible: `passThroughOptions` requires root positional parsing, and that setting is not per-subcommand.

Dropped it. The separator is now inserted into argv before commander runs, which is what `agency agent` already did via `injectAgentSeparator`, so this reuses an existing mechanism instead of adding one. Added a CLI acceptance test that runs `eval label ingest ... --store ...` and fails if the error is `unknown option`.

packages/agency-lang/lib/cli/forwardedArgs.ts
egonSchiele: Confirmed both. `-cfoo.json` and `-iv` were forwarded silently, which is precisely what the guard exists to prevent.

Fixed by taking commander's option metadata rather than a flat string list, as you suggested — the arity is needed twice over, since the same walk has to skip a flag's value when locating the filename (`--policy strict` covers two tokens, `-i` one). A short token now contributes every letter as a candidate flag name, covering the single, attached-value, and cluster forms.

Tests cover `-cfoo.json`, `-iv`, `--max-cost=5`, and the mirror cases before the filename where an attached value must not be mistaken for the filename. Both spellings are in the CLI acceptance test too.

packages/agency-lang/lib/cli/forwardedArgs.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export function misplacedFlagMessage(flag: string, input: string): string {
+  return [
+    `${flag} is an agency flag, not an argument for your program.`,
+    ``,
+    `  Put it before the file:      agency run ${flag} ... ${input}`,
+    `  Or hand it to the program:   agency run ${input} -- ${flag} ...`,
+  ].join("\n");
+}
```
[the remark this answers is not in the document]
egonSchiele: Added the `Error:` prefix.

egonSchiele:
Docs updated: `docs/site/cli/run.md` and the source comments in `stdlib/args.agency`, with `docs/site/stdlib/args.md` regenerated through `make` rather than hand-edited. Both now teach the position rule and drop the "only `agency run` needs `--`" framing.

One correction to the framing while I was in there: the eval example in run.md was `agency run writer.agency -- {task}`, which still works but no longer needs the separator, so it now reads `agency run writer.agency {task}`.

Also worth flagging that the review found something that changed the design, not just the code. `passThroughOptions` cannot be scoped to one subcommand — it requires positional parsing on the root, which broke `agency run -c cfg.json file.agency` and every nested command reading a parent option (`agency label ingest --store x` fails outright). I reproduced both. The separator is now inserted into argv before commander parses, which is the mechanism `agency agent` already used, so the CLI has one approach rather than two. Dev notes in `docs/dev/cli-arguments.md` record why `enablePositionalOptions` is not there, so the next person does not reach for it.

packages/agency-lang/lib/cli/runCommandLine.ts
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * `agency run --policy strict greet.agency --name alice` carries two command
+ * lines. The rule is position: agency's flags come before the filename, and
+ * everything after it belongs to the program. Commander cannot express that on
+ * its own — `passThroughOptions` would, but it requires positional parsing on
+ * the root command, which stops every nested command from seeing its parent's
+ * options (`agency label ingest --store x` fails outright).
+ *
+ * So the boundary is drawn before commander runs, by inserting `--` after the
+ * filename. `agency agent` already does this with `injectAgentSeparator`; this
```
egonSchiele: This now leaves two imperative implementations of the same responsibility: `insertProgramSeparator` walks Commander option shapes and inserts a boundary for `run`, while `injectAgentSeparator` separately walks a hand-written `AGENT_BUDGET_OPTIONS` list and inserts a boundary for `agent`. The new dev guide explicitly calls the first hand-written option list unsafe because it drifts silently, but then endorses exactly that list in the second implementation. This is the right point to expose one declarative splitter (command name, number/location of ownership positionals, Commander-derived options allowed before the boundary, and whether post-boundary Agency collisions are rejected) and keep the token walk/attached-value handling in one place. Both `run` and `agent` can supply policies to it. That neatly encapsulates the imperative argv parser and removes the duplicate mechanism before their edge-case behavior diverges.

packages/agency-lang/docs/dev/cli-arguments.md
```diff
@@ -0,0 +1,204 @@
...
+agency greet.agency bob             # error: too many arguments for 'default'
+```
+
+This predates the position rule and was not changed by it. Anyone who needs to
+pass arguments must write `agency run` in full. Fixing it means giving the
+default command the same `[nodeArgs...]` argument, pass-through, and guard that
+`run` has.
+
+## Testing
+
```
egonSchiele, line 249: This testing section names `lib/cli/forwardedArgs.test.ts`, but that file was deleted when the approach changed; the current suite is `lib/cli/runCommandLine.test.ts`. Please update the guide so its prescribed location actually exists.

egonSchiele:
The follow-up implementation is still failing integration because the repository migration was incomplete. `tests/integration/statelog/test.mjs:70` invokes `agency run main.agency -c agency.json`; every statelog scenario now exits in the new misplaced-flag guard, producing eight failures. Move `-c ...` before `main.agency`. There is at least one more missed caller that the failed job did not reach: `tests/integration/eval-run/test.mjs:135` constructs `agency run writer.agency --policy approve-all -- {task}`, so `--policy approve-all` must also move before `writer.agency` (while the explicit `-- {task}` can remain). Please sweep all actual argv construction—not only shell-looking string literals—and rerun integration. The PR body’s “Three spawn tests ... Nothing else in the repo did” claim is currently false. CI evidence: https://github.com/egonSchiele/agency-lang/actions/runs/30964979513/job/92176879722

packages/agency-lang/scripts/agency.ts
```diff
@@ -1956,7 +1961,24 @@ export async function runCli(
...
-  await program.parseAsync(injectAgentSeparator(argv));
+  // The flag list comes from commander itself. A hand-written copy would drift
+  // the moment someone adds an option to addRunOptions, and it would fail
+  // silently: the new flag would be forwarded to the program and do nothing.
+  const runCmd = program.commands.find((cmd) => cmd.name() === "run");
+  const flags = [...(runCmd?.options ?? []), ...program.options].map(
+    (option) => ({
+      short: option.short ?? undefined,
+      long: option.long ?? undefined,
+      takesValue: option.required || option.optional,
```
egonSchiele: Required and optional option values cannot be collapsed into one `takesValue` bit. `--trace [file]` is optional-valued, so `agency run --trace greet.agency` makes both this walk and Commander consume `greet.agency` as the trace filename, leaving the required input missing. The old working order, `agency run greet.agency --trace`, is now rejected as misplaced. That means the documented bare `--trace` mode (defaulting to `<input>.trace`) has no normal valid spelling under the new contract. Please preserve required vs optional arity and define/normalize an unambiguous bare-`--trace` form, then add end-to-end tests for both bare trace and an explicit trace filename; the current synthetic flag fixture omits the only optional-valued run option.

packages/agency-lang/lib/cli/runCommandLine.ts
```diff
@@ -0,0 +1,134 @@
...
+ */
+function flagNamesIn(token: string): string[] {
+  if (!isFlag(token)) return [];
+  if (token.startsWith("--")) {
+    const equals = token.indexOf("=");
+    return [equals === -1 ? token : token.slice(0, equals)];
+  }
+  // A short token is either one flag, one flag with its value attached, or a
+  // cluster of single-letter flags. Naming every letter covers all three.
+  return [...token.slice(1)].map((letter) => `-${letter}`);
```
egonSchiele: Naming every character creates false collisions for program-owned short tokens. For example, `agency run greet.agency -print` is rejected because the token contains `i`, even though it does not name Agency’s `-i`: short-option parsing starts at `-p`, which is unknown to Agency, so the whole post-filename token belongs to the program. Parse short tokens left-to-right with the option metadata: if the first name is unknown, there is no Agency collision; continue through known booleans, and stop after a known value-taking option because the remainder is its attached value. Add a forwarding case such as `-print` beside the valid `-iv` collision case.

packages/agency-lang/lib/cli/runCommandLine.ts
```diff
@@ -0,0 +1,134 @@
...
+    const token = argv[i];
+    if (TOP_LEVEL_BOOLEAN_FLAGS.includes(token)) {
+      i += 1;
+      continue;
+    }
+    if (TOP_LEVEL_VALUE_FLAGS.includes(token)) {
+      i += 2;
+      continue;
+    }
+    if (token.startsWith("--config=") || token.startsWith("--verbose=")) {
```
egonSchiele: The leading-root scan still misses Commander-valid attached/clustered forms. `agency -ccfg.json run greet.agency --name alice` and `agency -vc cfg.json run greet.agency --name alice` are valid root-option spellings, but this function mistakes the short token for the subcommand, so no separator is inserted and `run` later rejects the program’s `--name`. The current attached-config acceptance test places `-ccustom.json` after `run`, so it does not exercise this path. Please derive the root scan from the same option metadata and short-token parser, with tests for attached and clustered root flags before `run`.

egonSchiele:
Switched from refusing to warning, so `--` is now required for nothing.

```
$ agency run greet.agency --name alice
Hello, alice!

$ agency run greet.agency --max-cost 5
Warning: --max-cost went to your program, not to agency.
  Agency flags go before the filename: agency run --max-cost ... greet.agency
  Write -- before it to silence this:  agency run greet.agency -- --max-cost ...
```

Position now decides every case, which is what node, python, ruby, perl and deno all do — I tested the first four locally and each silently hands a misplaced interpreter flag to the script.

Agency warns where they stay silent because its misplaced flags fail in the opposite direction. A misplaced `node --inspect` announces itself (no debugger), and a misplaced `deno --allow-net` kills the script at the first socket. A misplaced `--max-cost` just runs uncapped and nothing notices. Worth knowing that cargo has had [an open issue](https://github.com/rust-lang/cargo/issues/10535) proposing exactly this warning since 2022 with no implementation, so we may be first here.

`--` survives only to silence the warning when a program genuinely owns a colliding flag name. It is no longer in any documented example.

Docs and `docs/dev/cli-arguments.md` updated to match, with `docs/site/stdlib/args.md` regenerated through `make`.

packages/agency-lang/scripts/agency.ts
```diff
@@ -1956,7 +1960,20 @@ export async function runCli(
...
```
egonSchiele: This remains broken at the current head. `--trace` is the only optional-valued run option, but it is still collapsed with required-valued options here. `agency run --trace greet.agency` consumes `greet.agency` as the optional trace path and reports the required input missing; placing bare `--trace` after the filename now forwards it to the program instead. Changing post-filename collisions from errors to warnings does not create a valid bare-trace ordering. Please preserve optional arity and add the end-to-end bare `--trace` test requested above.

packages/agency-lang/lib/cli/runCommandLine.ts
```diff
@@ -0,0 +1,136 @@
...
```
egonSchiele: The current head still cannot find `run` past attached/clustered root forms such as `agency -ccfg.json run greet.agency --name alice` or `agency -vc cfg.json run greet.agency --name alice`. This scan remains a hand-written subset of Commander syntax, so no separator is inserted and the program flag is rejected. Please address the earlier finding rather than only changing post-filename collision severity; add tests where these root forms appear before `run`.

packages/agency-lang/lib/cli/runCommandLine.ts
```diff
@@ -0,0 +1,136 @@
...
+ */
+function flagNamesIn(token: string): string[] {
+  if (!isFlag(token)) return [];
+  if (token.startsWith("--")) {
+    const equals = token.indexOf("=");
+    return [equals === -1 ? token : token.slice(0, equals)];
+  }
+  // A short token is either one flag, one flag with its value attached, or a
+  // cluster of single-letter flags. Naming every letter covers all three.
+  return [...token.slice(1)].map((letter) => `-${letter}`);
```
egonSchiele: This current implementation still emits false warnings for program-owned tokens such as `-print`: `flagNamesIn` invents `-p`, `-r`, `-i`, ... and warns merely because one interior letter matches Agency’s `-i`. Downgrading the false positive from refusal to warning does not make it accurate, and it now pollutes stderr of otherwise valid programs. Please apply the left-to-right short-option parsing from the earlier finding and add a negative test proving `-print` is forwarded without a warning.

egonSchiele:
Changing the misplaced-option result from an error to a warning does not remove the repository migration requirement; it makes the missed callers continue with the wrong behavior. `tests/integration/statelog/test.mjs:70` still puts `-c agency.json` after `main.agency`, so the config is forwarded to the Agency program and never loaded by the CLI. The scenarios therefore run without their intended observability configuration. `tests/integration/eval-run/test.mjs:135` still puts `--policy approve-all` after `writer.agency`, so the approval policy is not installed; worse, this writer reads raw `args()[0]`, which now becomes `"--policy"` instead of the eval task. Please migrate both callers to the new ordering. The warning is useful for external users, but it is not a compatibility implementation for internal callers that rely on those flags taking effect.

egonSchiele:
The PR description is now materially stale after the redesign. It still says the guard errors, shows an error message, names deleted `lib/cli/forwardedArgs.ts` / `forwardedArgs.test.ts`, claims only three repository callers existed, and says `docs/site/**` is untouched even though the docs are now changed. Please rewrite the description around the current warning-and-forward behavior and current files/tests; otherwise reviewers and the eventual merge record describe a different implementation and safety contract.

packages/agency-lang/lib/cli/runCommandLine.ts
egonSchiele: Done — `injectAgentSeparator` is gone and both commands go through `splitCommandLine` with a policy each:

```ts
{ command: "run",   ownedPositionals: 1, options: run + root, warnOnCollision: true  }
{ command: "agent", ownedPositionals: 0, options: agent only, warnOnCollision: false }
```

`ownedPositionals` is the real difference: run owns the filename, agent owns nothing. The token walk, arity, attached values and short bundles now live in one place, so the short-option fix below reaches both. Agent options come from commander metadata rather than the hand-written `AGENT_BUDGET_OPTIONS`; they happen to be the same two flags, so behaviour is preserved. Root options are deliberately excluded from the agent policy, since `agency agent -c x` has always given `-c x` to the agent.

packages/agency-lang/lib/cli/runCommandLine.ts
egonSchiele: Confirmed and fixed. `parseShortToken` now reads left to right the way commander does: an unknown first letter means the whole token is the program's, so `-print` forwards with no warning. A later unknown letter is the attached value of the option before it.

Tests: `-print` asserted silent in both the unit suite and the CLI acceptance test, alongside the `-iv` and `-vc` collision cases so the negative case cannot pass by the check being dead.

packages/agency-lang/lib/cli/runCommandLine.ts
egonSchiele: Confirmed and fixed. `findSubcommandIndex` now takes the root option metadata and uses the same short-token parser, so `agency -ccfg.json run f.agency` and `agency -vc cfg.json run f.agency` both find `run`. The hand-written `TOP_LEVEL_BOOLEAN_FLAGS`/`TOP_LEVEL_VALUE_FLAGS` lists are gone.

Tested over `-v`, `-c cfg.json`, `-ccfg.json`, `-vc cfg.json` and `--config=cfg.json` before the subcommand.

packages/agency-lang/scripts/agency.ts
egonSchiele: Fixed by splitting the option: `--trace` (bare, writes `<input>.trace`) and `--trace-file <path>`.

On the arity point specifically: I probed commander, and an optional value *does* still consume the next word — `run --trace greet.agency` gives `trace="greet.agency"` and then "missing required argument". So the walk was already matching commander; modelling arity as three states does not by itself create a valid bare-trace spelling.

The real finding is that there was never one. `agency run --trace foo.agency` fails the same way on main, which I verified against the main build — even though `docs/dev/trace.md:9` documents exactly that command. The only working spelling was `agency run foo.agency --trace`, which the position rule sends to the program. Two flags is the only shape that works before the filename.

Arity is still modelled as `none | required | optional` since the walk needs it, and there are end-to-end tests asserting both bare `--trace` and `--trace-file custom.trace` actually write their files.

packages/agency-lang/docs/dev/cli-arguments.md
egonSchiele: Fixed — the guide now names `lib/cli/commandLine.test.ts`, which is also where the file ended up after the splitter was unified.

egonSchiele:
All seven addressed, plus the description rewritten around the current behaviour.

**The missed callers were my error, and the same one twice.** I searched for the shape I expected — a literal `.agency` followed by a flag — and both callers build the filename with `JSON.stringify(join(...))`, so no such text exists in the source. Swept by concept this time and found exactly the two you named, no others. `eval-run` was worse than a lost flag: that writer reads `args()[0]`, so the task became `"--policy"`. Both suites now pass locally, as does statelog.

**`--trace` deserves a note.** Modelling arity properly does not by itself fix it — I probed commander, and an optional value consumes the next word too, so the walk already matched. The real finding is that there was never a working bare-trace spelling: `agency run --trace foo.agency` fails identically on main, which I verified against the main build, even though `docs/dev/trace.md:9` documents that exact command. Splitting into `--trace` and `--trace-file <path>` makes both forms work for the first time.

**One splitter now.** `injectAgentSeparator` is deleted; run and agent each supply a policy. That closes the contradiction you pointed out, where the dev doc called a hand-written flag list unsafe two sections before endorsing one.

packages/agency-lang/lib/cli/commandLine.ts
```diff
@@ -0,0 +1,198 @@
...
+  token: string,
+  options: CliOption[],
+): number | undefined {
+  if (token.startsWith("--")) {
+    const equals = token.indexOf("=");
+    const name = equals === -1 ? token : token.slice(0, equals);
+    const arity = findOption(name, options)?.arity;
+    if (arity === undefined) return undefined;
+    if (equals !== -1) return 1;
+    // An optional value still consumes the next word: commander reads
```
egonSchiele: This generic arity handling still does not match Commander for optional values. Commander consumes the next token for `--opt value`, but leaves it alone when it starts with `-` (`--opt --verbose` parses as a bare optional flag plus `--verbose`). This always returns 2, so any optional-valued option added to a boundary will swallow the following flag while finding the owned positional and may skip boundary insertion entirely. Please make the decision with access to the next token and add both optional-value cases; otherwise the declarative `Arity = "optional"` interface promises behavior the shared walker does not implement.

packages/agency-lang/tests/integration/cli/test.mjs
```diff
@@ -208,26 +208,135 @@ node main() {
...
+      `./node_modules/.bin/agency run ${configFlag} greet.agency --name alice`,
+    );
+    assertIncludes(greetedWithConfig, "Hello, alice!");
+  }
+
+  // Position always wins, so an agency flag after the filename goes to the
+  // program, and the warning is the only thing standing between that and a
+  // silent surprise. greet does not declare --max-cost, so its own parser then
+  // rejects it — which is also the proof the flag really was forwarded. That it
+  // is forwarded rather than intercepted is pinned in runCommandLine.test.ts.
```
egonSchiele: This points readers to `runCommandLine.test.ts`, which this revision deletes in favor of `commandLine.test.ts`. Please update the test reference.

packages/agency-lang/docs/dev/cli-arguments.md
```diff
@@ -0,0 +1,243 @@
...
+This was tried and reverted. If you find yourself reaching for
+`enablePositionalOptions`, that is why it is not there.
+
+### What is used: draw the boundary before commander runs
+
+`lib/cli/commandLine.ts` inserts `--` after the filename, then hands the
+rewritten argv to commander, which parses it normally. Commander removes the
+separator itself, so the program never sees it.
+
+`agency agent` already worked this way (`injectAgentSeparator` in
```
egonSchiele: This new developer guide still describes the deleted implementation in several places: `injectAgentSeparator` here, the old `takesValue` mapping at lines 79–84, `insertProgramSeparator` around line 134, and the old "name every letter" short-token algorithm around lines 151–155. The latter is specifically the false-positive implementation replaced by `parseShortToken`. Please update these sections to document `splitCommandLine`, three-state `arity`, and left-to-right short parsing so the canonical design guide does not direct the next change back toward the bugs this PR fixed.

packages/agency-lang/lib/cli/commandLine.ts
egonSchiele: Confirmed and fixed. `ownedFlagTokens` now takes the next token and copies commander's own test.

Probing turned up a twist worth recording, which is now in the code comment and the dev guide: commander does not simply skip anything starting with a dash. A digit after the dash reads as a negative number, so an optional value behaves three ways:

```
--trace --verbose   -> trace=true            1 token
--trace -v          -> trace=true            1 token
--trace -5          -> trace="-5"            2 tokens
--trace out.trace   -> trace="out.trace"     2 tokens
```

Required is unconditional by contrast: `--policy --verbose` gives `policy="--verbose"`, verified.

You are right that this was a promise the walker did not keep. No run option is optional-valued any more — splitting `--trace [file]` removed the last one — so the unit fixture now carries a synthetic optional flag to keep the path covered, with a comment saying why it exists. Cases added for the flag-looking, negative-number, and plain-word forms, plus the required-value contrast.

packages/agency-lang/tests/integration/cli/test.mjs
egonSchiele: Fixed — the comment now points at `commandLine.test.ts`.

packages/agency-lang/docs/dev/cli-arguments.md
egonSchiele: Fixed all four. The guide now documents `splitCommandLine` and `optionsOf` with three-state arity, and the short-token section describes left-to-right parsing with an explicit "do not go back to naming every letter" note and the `-print` example, so the reason is on the page rather than only in the history.

Added a table for the arity rules, since the negative-number case is the kind of thing that gets re-broken by someone simplifying the check.

packages/agency-lang/lib/cli/commandLine.ts
```diff
@@ -0,0 +1,228 @@
...
+  if (arity === "required") return 2;
+  return next === undefined || looksLikeOption(next) ? 1 : 2;
+}
+
+/** Commander's own test for "this token is a flag, not a value". */
+function looksLikeOption(token: string): boolean {
+  return (
+    token.length > 1 &&
+    token[0] === "-" &&
+    (token[1] === "-" || Number.isNaN(Number.parseFloat(token[1])))
```
egonSchiele, line 217: This is not actually Commander’s negative-number test. Commander 14 uses the full regex `/^-(\\d+|\\d*\\.\\d+)(e[+-]?\\d+)?$/` and also rejects the value interpretation when any command in the hierarchy defines that digit as a short option. The single-character `parseFloat` check therefore misclassifies tokens such as `-5abc` and `-0x10` as optional values; Commander treats them as options/operands instead. It also misclassifies `-5` if `-5` is a declared short option. Please either copy Commander’s complete predicate (including the option-metadata check) or constrain the abstraction so it does not claim exact optional-arity support. Add a non-number beginning with a digit such as `-5abc` to the test—the current `-5` case cannot distinguish these implementations.

PR #806, opened 2026-08-05
feat(remote): agency remote spend — hosted spend from the CLI

egonSchiele:
## Summary

Adds `agency remote spend [project]` — a read-only CLI command that prints hosted spend from statelog's Group 4 spend endpoints. This is the client that surfaces the serve cost seam (#801) and per-model breakdown (#802) work: how much a hosted agent cost, without opening the statelog UI.

- `agency remote spend` → the account-wide rollup (all projects), one row each.
- `agency remote spend <project>` → one project's spend.
- `--since 7d` (or `--from`/`--to` as ISO-8601/epoch-ms) scopes a time window; `--json` for machines.

## Design

A thin command over the existing two-client split — mirrors how `whoami`/`projects` (account) and `pull`/`logs` (project) already work.

- **`resolveSpendWindow`** (pure) turns the time flags into a validated half-open epoch-ms window plus a display label; it returns-or-throws (never exits), and `runSpend` converts a throw at the `fail()` boundary. Rejects non-positive/fractional/overflow/pre-epoch `--since`, bare-local datetimes (requires a `Z`/offset), and nonexistent calendar dates.
- **Shared `spendTypes.ts`** is the source of truth: Zod schemas enforce the numeric invariants (finite non-negative cost, non-negative safe-integer counts, `deletedAt` null|ISO, and `pricingComplete === (unpricedCallCount === 0)`); the TS types are inferred from them; `toSpendQuery` is the one query serializer both clients use.
- **Clients** gain `getSpend`/`getAccountSpend` behind a single declarative request input per client (query support + private route-compatibility detection — no second transport path, no exported error class). An unknown 404 on a spend route is reported as "host does not support the spend API" (narrowed to HTTP 404 so a proxy's non-JSON 5xx stays a server error); a project's known `{error:"Project not found"}` 404 keeps its message.
- **Renderers** use adaptive money formatting (`$0.0000` only for true zero, `<$0.0001` for tiny positive), and the account TOTAL carries both trust axes — a lower-bound `≥` when any row's usage is incomplete, plus summed unpriced-call notes. Rows sort active-by-cost-desc, then deleted, with the slug as tie-break.

## Scope / non-goals

- Read-only; public wire/CLI conventions unchanged; purely additive (one shared types module, two client methods, one window parser, two renderers, one command).
- **Per-model breakdown is out of scope**: the server's `ProjectSpend` is still flat (Group 4 predates #802). The runtime already sends the per-model data; the next step is a statelog follow-up to aggregate it, after which this command grows a `--by-model` view (the renderer is shaped to accept it).
- No CSV (statelog dropped it; `--json` is the machine surface).

## Testing

Full TDD, deterministic, no network/LLM: `spendTypes`, `spendWindow` (all window edge cases), both client methods (query building incl. one-sided windows, wire-invariant rejection, the 404 policy + a non-JSON-5xx regression), the renderers (zero-vs-tiny money, degraded totals, sort/tie-break, empty), the command (scope selection, invalid-flag → `fail` with no client call, project-scoped-key hint, `--json` to stdout only), and Commander forwarding in `scripts/agency.test.ts`. `pnpm run typecheck` clean; full `pnpm test:run` green (740 files) after a build.

### DIFF
--- a/packages/agency-lang/lib/cli/remote/render.ts
+++ b/packages/agency-lang/lib/cli/remote/render.ts
@@ -11,6 +11,7 @@ import type {
   KeySummary,
   CreatedKey,
 } from "../statelog/accountClient.js";
+import type { ProjectSpend, AccountSpendRow } from "../statelog/spendTypes.js";
 
 const NONE = "—";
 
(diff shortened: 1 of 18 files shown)


packages/agency-lang/lib/cli/remote/render.ts
```diff
@@ -112,6 +113,118 @@ export function renderPullSummary(names: string[], outputDir: string): string {
...
+  return [...rows].sort((left, right) => {
+    const leftDeleted = left.deletedAt !== null;
+    const rightDeleted = right.deletedAt !== null;
+    if (leftDeleted !== rightDeleted) {
+      return leftDeleted ? 1 : -1;
+    }
+    if (right.spend.pricedCost !== left.spend.pricedCost) {
+      return right.spend.pricedCost - left.spend.pricedCost;
+    }
+    return left.projectSlug < right.projectSlug ? -1 : left.projectSlug > right.projectSlug ? 1 : 0;
```
egonSchiele: Non-blocking: this is a nested ternary, which `docs/dev/anti-patterns.md` explicitly prohibits. Please expand the slug tie-break into two braced comparisons (return `-1` when left < right, `1` when left > right, otherwise `0`) so the implementation matches the plan's anti-pattern constraint while preserving deterministic code-point ordering.

egonSchiele (commented):
Reviewed the implementation against the revised spec and plan. Scope selection, strict window parsing, client boundaries, 404 compatibility handling, trust-aware totals, JSON output, and Commander wiring all match the intended contracts. The full CI matrix is green. No correctness blockers found; one non-blocking anti-pattern cleanup is noted inline.

packages/agency-lang/lib/cli/remote/render.ts
```diff
@@ -112,6 +113,118 @@ export function renderPullSummary(names: string[], outputDir: string): string {
...
+
+/** A count with thousands separators, in a fixed locale for deterministic output. */
+function formatCount(count: number): string {
+  return count.toLocaleString("en-US");
+}
+
+/** Prefix `≥` when the figure is a trusted lower bound (telemetry incomplete). */
+function lowerBound(amount: number, complete: boolean): string {
+  return complete ? formatUsd(amount) : `≥ ${formatUsd(amount)}`;
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 4895d515a. A lower bound now never uses the `<$0.0001` sentinel: `lowerBound` returns `≥ $0.0000` for a sub-$0.0001 incomplete amount (true — the total is at least ~zero and may be higher) instead of the incoherent `≥ <$0.0001`. Added a regression test asserting the output never contains `≥ <`.

packages/agency-lang/lib/cli/remote/render.ts
egonSchiele: Fixed in 4895d515a. Expanded the nested ternary into three braced comparisons (`-1` when left < right, `1` when left > right, else `0`), preserving deterministic code-point ordering per the plan constraint.

PR #808, opened 2026-08-05
agency <file>: the shorthand takes a command line like run does

egonSchiele:
## What this does

`agency greet.agency` is documented as shorthand for `agency run greet.agency`, but that equivalence was false the moment your program took arguments:

```
agency greet.agency --name bob   # error: unknown option '--name'
agency greet.agency bob          # error: too many arguments for 'default'
```

It was a hidden default command declaring only `[file]` — no variadic to forward, and no boundary policy. It now has both, sharing `run`'s options and `run`'s policy, so every form works the same way:

```
agency greet.agency --name alice                      # Hello, alice!
agency --policy strict greet.agency --name alice      # Hello, alice!
agency greet.agency -- --name alice                   # Hello, alice!
agency -c custom.json greet.agency --name alice       # Hello, alice!

agency greet.agency --max-cost 5
Warning: --max-cost went to your program, not to agency.
```

The first two used to fail.

## How

One new policy, in the same shared splitter #805 introduced:

```ts
{ command: null, ownedPositionals: 1, options: runOptions, warnOnCollision: true }
```

Two things about the shorthand differ from a named command, and both live in the policy rather than in the walk.

**No command word to step over.** `command: null`, so the walk starts on the filename itself rather than one token later.

**Its flags sit at the top level.** In `agency --policy strict greet.agency`, `--policy` appears where a root flag would. The scan that locates the subcommand is therefore given every option agency owns anywhere, not just the root's. Without that it stops on `strict` and runs a file by that name — which is how I found it, since a test caught it.

`splitCommandLine` also takes the list of command names now, aliases included. Otherwise the shorthand policy would claim real commands and push a separator into `agency compile a.agency b.agency`.

## Testing

- `lib/cli/commandLine.test.ts` — the shorthand splitting like `run`, agency flags before the filename, the warning, and a case per command/alias proving `compile`, `build`, `fmt` and `format` are left alone
- `tests/integration/cli/test.mjs` — the real binary over all five forms above, plus `agency compile a.agency b.agency` to prove a two-file command still works

Green locally: those, the three spawn suites, `labelCommand.test.ts`, `scripts/agency.test.ts`, and the CLI, statelog and eval-run integration suites.

## Docs

`docs/dev/cli-arguments.md` — the "Known limitation" section becomes a description of how the shorthand works, since it is no longer a limitation. Plus `docs/site/cli/run.md` and the `agency help` text, which both described a shorthand that could not take arguments.

## Not in this PR

The investigation also turned up two smaller gaps in `agency agent`: a misplaced `--max-cost` is forwarded to the agent, and root flags like `-c` cannot reach agency at all. Both currently produce a clear error from the agent rather than failing silently, so I left them out of this change. Happy to do them next if you want.

### DIFF
--- a/packages/agency-lang/docs/dev/cli-arguments.md
+++ b/packages/agency-lang/docs/dev/cli-arguments.md
@@ -229,21 +229,33 @@ command, not a mistake.
 The walk itself — arity, attached values, short bundles — lives in one place, so
 a fix to short-option parsing reaches both.
 
-## Known limitation: the shorthand takes no program arguments
+## The shorthand is `run` with the word left out
 
-`agency greet.agency` is a shorthand for `agency run greet.agency`. It is a
-hidden default command declaring only `[file]`, with no variadic argument, so it
-cannot forward anything:
+`agency greet.agency` is a hidden default command. It carries the same options
+as `run` and the same boundary policy, so everything on this page applies to it
+unchanged:
 
 ```
-agency greet.agency --name bob      # error: unknown option '--name'
-agency greet.agency bob             # error: too many arguments for 'default'
+agency greet.agency --name alice
+agency --policy strict greet.agency --name alice
+agency greet.agency --max-cost 5     # same warning as run
 ```
 
-This predates the position rule and was not changed by it. Anyone who needs to
-pass arguments must write `agency run` in full. Fixing it means giving the
-default command the same `[nodeArgs...]` argument, pass-through, and guard that
-`run` has.
+Two things make it different from a named command, and both are in the policy
+rather than in the walk.
+
+**It has no command word to step over.** Its `command` is `null`, and the walk
+starts on the filename itself rather than one token later.
+
+**Its flags sit at the top level.** In `agency --policy strict greet.agency`,
+`--policy` appears where a root flag would. So the scan that locates the
+subcommand is given every option agency owns anywhere, not just the root's —
+otherwise it stops on `strict` and treats that as the filename.
+
+That second point is why `splitCommandLine` also takes the list of command
+names: without it, the shorthand policy would claim real commands, and
+`agency compile a.agency b.agency` would get a separator pushed into its file
+list. Aliases are included, so `agency fmt x.agency -i` stays a format run.
 
 ## Testing
 
--- a/packages/agency-lang/docs/dev/cli-arguments.md
+++ b/packages/agency-lang/docs/dev/cli-arguments.md
@@ -203,19 +203,23 @@ that doing so would break any program that legitimately wants positional
 arguments beginning with a dash, which is the reason POSIX has the convention at
 all.
 
-## One splitter, two policies
+## One splitter, three policies
 
-`agency agent` needs the same boundary for a different shape: it takes no
-filename, and it forwards its whole command line to an agent that has its own
-flag parser. It used to have its own copy of the walk, with a hand-written list
-of the flags to keep on agency's side.
+`agency agent` and the bare `agency greet.agency` shorthand need the same
+boundary in different shapes. The agent takes no filename and forwards its whole
+command line to a program with its own flag parser; it used to carry a second
+copy of the walk over a hand-written list of flags.
 
-Both now go through `splitCommandLine` with a policy each:
+All three now go through `splitCommandLine` with a policy each:
 
 ```ts
 { command: "run",   ownedPositionals: 1, options: run + root, warnOnCollision: true  }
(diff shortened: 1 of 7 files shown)


packages/agency-lang/scripts/agency.ts
```diff
@@ -2006,12 +2016,30 @@ export async function runCli(
   // The flag list comes from commander itself. A hand-written copy would drift
   // the moment someone adds an option to addRunOptions, and it would fail
   // silently: the new flag would be forwarded to the program and do nothing.
+  const runOptions = [
+    ...optionsOf(commandNamed(program, "run")),
+    ...optionsOf(program),
+  ];
+  // Every name commander answers to, so the shorthand policy does not claim a
+  // command. Aliases count: `agency fmt x.agency` must stay a format run.
+  const commandNames = program.commands.flatMap((cmd) => [
```
egonSchiele: Commander’s implicit `help` command is not present in `program.commands`, so this list omits `help` and the null shorthand policy claims it. That changes existing invocations with trailing options: `agency help --help` is rewritten to `agency help -- --help` (main exits 0 and prints help; this branch exits 1 and prints error help), and `agency help --version` is likewise rewritten, warns that `--version` went to the program, and exits 1 instead of printing the version. Please reserve Commander’s implicit help-command name when classifying shorthand vs. named commands, and add real-CLI regressions for `help --help` and `help --version`—the current command/alias cases only cover entries visible in `program.commands`.

packages/agency-lang/docs/dev/cli-arguments.md
```diff
@@ -229,21 +229,33 @@ command, not a mistake.
 The walk itself — arity, attached values, short bundles — lives in one place, so
 a fix to short-option parsing reaches both.
 
-## Known limitation: the shorthand takes no program arguments
+## The shorthand is `run` with the word left out
```
egonSchiele, line 236: The preceding section is now stale: its heading still says “One splitter, two policies”, and its prose/example says only `run` and `agent` use the splitter. This PR adds the null-command shorthand as a third policy. Please update that section (including the policy example) so the canonical design documentation matches the architecture described here.

packages/agency-lang/scripts/agency.ts
egonSchiele: Confirmed and fixed. Reproduced both against the build on main, which exits 0 for each:

```
             this branch            main
help --help  exit 1                 exit 0
help --version  Warning + exit 1    prints 0.0.105, exit 0
```

The fix takes the command list from two sources, because neither is complete alone:

```ts
const commandNames = [
  ...program.commands,                               // has hidden ones: remote, default
  ...program.createHelp().visibleCommands(program),  // has the implicit help
].flatMap((cmd) => [cmd.name(), ...cmd.aliases()]);
```

I went with `visibleCommands` rather than naming `help` directly, so a future `.helpCommand("assist")` cannot silently reopen this.

Real-CLI regressions added for `help --help`, `help --version`, and `remote --help` — the last one because `visibleCommands` alone drops hidden commands, which is the mirror-image mistake.

packages/agency-lang/docs/dev/cli-arguments.md
egonSchiele: Fixed — the heading is now "One splitter, three policies", the example lists the `command: null` policy, and the surrounding prose covers the shorthand alongside run and agent rather than only the two.

packages/agency-lang/tests/integration/cli/test.mjs
```diff
@@ -353,6 +353,50 @@ node main() {
...
+  const shortWarned = run(
+    dir,
+    "./node_modules/.bin/agency greet.agency --max-cost 5 2>&1",
+    { expectFail: true },
+  );
+  assertIncludes(shortWarned, "Warning: --max-cost went to your program");
+
+  // A real command must not be mistaken for a filename. compile takes a list of
+  // files, so a separator pushed in here would break it.
+  run(dir, "./node_modules/.bin/agency compile greet.agency basic.agency");
```
egonSchiele, line 383: This does not actually prove `compile` escaped the shorthand classifier. If `compile` were missing from `commandNames`, the rewrite would be `agency compile -- greet.agency basic.agency`; Commander strips that separator and still passes the same two positional inputs, so this test remains green. Use an option-bearing invocation such as `agency compile --force greet.agency basic.agency`: correct classification parses `--force`, while an incorrect null-boundary rewrite makes `--force` an input path and fails. The unit test cannot cover production command-name extraction because it supplies a synthetic `COMMAND_NAMES` list.

PR #810, opened 2026-08-05
Per-file stdlib dependency tracking in the build manifest

egonSchiele:
Implements the approved spec `2026-08-04-stdlib-deps-manifest-spec.md`: editing one standard-library file makes `agency compile stdlib/` rebuild only that file and its transitive importers, instead of all 83.

## Why

Stdlib files import each other via `std::`, which the closure walker deliberately skips, so their manifest `deps` were empty and the all-contents `stdlibHash` invalidated every entry on any one-file edit. Measured before this change: a one-file edit cost a ~6s full rebuild of all 83 files.

## What changed

- `agencyImportTargets` gains an opt-in `resolveStdlib` flag that maps `std::` targets to absolute stdlib file paths (default behavior unchanged).
- New `lib/compiler/depFingerprint.ts`: `dependencyFingerprint` walks a module's transitive imports and returns `{ deps, hasPkgImports, cacheable }`. Completeness is data: a splice anywhere in the subtree, or an unparseable/missing reachable file, marks the fingerprint uncacheable (recorded but never fresh). Discovery absorbs an enumerated errno list only; programming errors (incl. string-coded ones like `ERR_INVALID_ARG_TYPE`) propagate.
- Manifest schema v2: new `cacheable` field, and two `stdlibHash` flavors selected by the shared `stdlibHashFor` rule — stdlib-resident entries store a names-only hash (add/remove/rename still rebuilds all of stdlib; a content edit invalidates per file via the new real deps), non-stdlib entries keep the full-content hash. Old manifests are discarded (one full rebuild).
- `BuildSession` records stdlib modules via the fingerprint; user modules keep the closure-derived path, now with subtree-accurate `hasPkgImports` and `cacheable: true`.
- `docs/dev/incremental-builds.md` updated (flavors, fingerprint contract, single-file guarantee boundary, migration).

## Measured on this branch (real stdlib)

| scenario | before | after |
|---|---|---|
| warm, nothing changed | ~0.6s | ~0.6s ("83 file(s) up to date") |
| edit `math.agency` (leaf) | ~6s, all 83 | **0.55s, exactly 1 file** |
| edit `agency.agency` (hub) | ~6s, all 83 | **2.2s, 17 files (its true importer set)** |
| incremental vs `--force` emit | — | byte-identical (all 83 `.js` diffed) |

Splice census: zero splices in the current stdlib, so no file is pinned uncacheable today.

## Tests

71 unit tests across the extractor, fingerprint (incl. the errno table and carve-outs against the real stdlib), manifest, and tracker, plus 19 end-to-end tests against a redirected fixture stdlib: exact BuildSession rewrite sets per mutation (backdated-mtime detection), production manifest wiring (transitive `std::` deps, a real executed declaration splice recording `cacheable: false`, the `stdlib-copy` boundary), the single-file guarantee boundary through the public session, and incremental-vs-force byte parity across 8 mutations including a seeded v1-manifest migration row.

## Plan deviations

- The plan's Task 0 timing spike was skipped (its one-line hack + two `make build` runs); execution was already approved and the real implementation's measurements above supersede the spike's approximation. The splice census ran (zero splices).
- The full `make` was not run locally (per instruction); `pnpm run build` built dist for the splice-runner subprocess, and CI's unit job runs `make` before vitest.
- End-to-end fixtures live under `.agency-tmp/` in the package (not `os.tmpdir()`): splice generators execute in a subprocess that must resolve the `agency-lang` package and dist. Fixture files are all non-templated via the mock (incl. the stdlib-copy user file — several prelude names are parser keywords, so stubbing the prelude surface is not possible), and each fixture root carries its own `agency.json` so `manifestDirFor` anchors there.
- The splice generator drops the `import { Code } from "std::agency"` annotation: the compiled generator would import the real stdlib's `agency.js`, which does not exist in an unbuilt tree (typechecking is off in these tests, so the annotation carries no weight).

Spec: `packages/agency-lang/2026-08-04-stdlib-deps-manifest-spec.md` · Plan: `packages/agency-lang/2026-08-04-stdlib-deps-manifest-plan.md` (both in the main checkout, untracked).

### DIFF
--- a/packages/agency-lang/lib/compiler/buildManifest.test.ts
+++ b/packages/agency-lang/lib/compiler/buildManifest.test.ts
@@ -262,9 +262,13 @@ describe("isEntryFresh — each field is load-bearing", () => {
 
 describe("stdlib hash flavors", () => {
   test("stdlibHashFlavor: names inside stdlibDir, contents outside, boundary exact", () => {
-    expect(stdlibHashFlavor("/m/stdlib/math.agency", "/m/stdlib", "N", "C")).toBe("N");
-    expect(stdlibHashFlavor("/m/src/app.agency", "/m/stdlib", "N", "C")).toBe("C");
-    expect(stdlibHashFlavor("/m/stdlib-copy/x.agency", "/m/stdlib", "N", "C")).toBe("C");
+    // path.join throughout: production compares with path.sep appended,
+    // so hard-coded "/" separators would miss the boundary on Windows.
+    const m = path.join(os.tmpdir(), "m");
+    const stdlibDir = path.join(m, "stdlib");
+    expect(stdlibHashFlavor(path.join(stdlibDir, "math.agency"), stdlibDir, "N", "C")).toBe("N");
+    expect(stdlibHashFlavor(path.join(m, "src", "app.agency"), stdlibDir, "N", "C")).toBe("C");
+    expect(stdlibHashFlavor(path.join(m, "stdlib-copy", "x.agency"), stdlibDir, "N", "C")).toBe("C");
   });
 
   test("computeStdlibNamesHash: edits invisible; add/rename visible; recursive; .agency-only; order-insensitive", () => {
--- a/packages/agency-lang/lib/compiler/buildManifest.test.ts
+++ b/packages/agency-lang/lib/compiler/buildManifest.test.ts
@@ -9,9 +9,11 @@ import {
   loadManifest,
   saveManifest,
   computeStdlibHash,
+  computeStdlibNamesHash,
   computeCompilerStamp,
   isEntryFresh,
   manifestDirFor,
+  stdlibHashFlavor,
   MANIFEST_DIR_NAME,
   type BuildManifest,
   type FreshnessContext,
@@ -33,11 +35,12 @@ function freshFixture(): { dir: string; manifest: BuildManifest; ctx: FreshnessC
   const shared = {
     stdlibHash: "STDLIB",
     hasPkgImports: false,
+    cacheable: true,
     configKey: "{}",
     compilerStamp: "COMPILER",
   };
   const manifest: BuildManifest = {
-    version: 1,
+    version: 2,
     entries: {
       "main.agency": {
         ...shared,
@@ -58,6 +61,10 @@ function freshFixture(): { dir: string; manifest: BuildManifest; ctx: FreshnessC
   const ctx: FreshnessContext = {
     manifestDir: dir,
     stdlibHash: "STDLIB",
+    stdlibNamesHash: "NAMES",
+    // A subdir the fixture modules are NOT in, so they carry the contents
+    // flavor and every pre-existing expectation holds unchanged.
+    stdlibDir: path.join(dir, "stdlib"),
     compilerStamp: "COMPILER",
     configKey: "{}",
   };
@@ -80,7 +87,7 @@ describe("manifest IO", () => {
 
   test("saveManifest round-trips atomically and leaves no tmp file", () => {
     const dir = tmp();
-    const manifest: BuildManifest = { version: 1, entries: {} };
+    const manifest: BuildManifest = { version: 2, entries: {} };
     saveManifest(dir, manifest);
     expect(loadManifest(dir)).toEqual(manifest);
     expect(fs.readdirSync(path.join(dir, MANIFEST_DIR_NAME))).toEqual(["manifest.json"]);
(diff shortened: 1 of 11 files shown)


packages/agency-lang/lib/compiler/stdlibDeps.test.ts
```diff
@@ -0,0 +1,349 @@
...
+// covered against the real stdlib in depFingerprint.test.ts).
+const fake = vi.hoisted(() => ({ stdlibDir: "", fixtureRoot: "" }));
+vi.mock("../importPaths.js", async (importOriginal) => {
+  const real = await importOriginal<typeof import("../importPaths.js")>();
+  return {
+    ...real,
+    getStdlibDir: () => fake.stdlibDir,
+    // EVERY fixture file (stdlib and user alike) is non-templated, so no
+    // fixture needs the ~30-name prelude surface stubbed out.
+    isNonTemplatedStdlib: (p: string) =>
```
egonSchiele, line 26: This makes the end-to-end fixture graph materially simpler than production: every fixture file—including `stdlib-copy` user modules—is non-templated, so these BuildSession tests cannot catch a wiring regression involving the template-added `std::index` edge. The real-stdlib fingerprint unit test covers discovery in isolation, but not the session→fingerprint→manifest path for a templated module. Could this suite include one templated fixture case (or a separate focused case with the minimum prelude surface) that asserts the emitted manifest records the implicit `index.agency` dependency?

packages/agency-lang/lib/compiler/buildManifest.test.ts
```diff
@@ -214,4 +221,87 @@ describe("isEntryFresh — each field is load-bearing", () => {
...
+    // Contents flavor changing does NOT stale a stdlib-resident entry…
+    expect(isEntryFresh(relocate("main.agency"), manifest, { ...ctx, stdlibHash: "OTHER" })).toBe(true);
+    // …but the names flavor changing does.
+    expect(isEntryFresh(relocate("main.agency"), manifest, { ...ctx, stdlibNamesHash: "OTHER" })).toBe(false);
+  });
+});
+
+describe("stdlib hash flavors", () => {
+  test("stdlibHashFlavor: names inside stdlibDir, contents outside, boundary exact", () => {
+    expect(stdlibHashFlavor("/m/stdlib/math.agency", "/m/stdlib", "N", "C")).toBe("N");
```
egonSchiele: These hard-coded POSIX paths make the inside-stdlib assertion fail on Windows because production appends `path.sep` (`\`) while the fixture strings contain `/`. Please construct the stdlib root and candidate paths with `path.join()` (using a platform-neutral temporary/root prefix) so this tests the boundary rule rather than the host separator.

packages/agency-lang/lib/compiler/depFingerprint.test.ts
```diff
@@ -0,0 +1,194 @@
...
+import { dependencyFingerprint } from "./depFingerprint.js";
+import { getStdlibDir } from "../importPaths.js";
+
+beforeEach(() => {
+  seam.throwNext = null;
+  seam.calls.length = 0;
+});
+
+function tree(files: Record<string, string>): string {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), "agency-fp-"));
```
egonSchiele, line 48: Each `tree()` call creates an OS temp directory, including one per errno/package-import row, but this file never removes them. Please track the created roots and delete them in `afterEach` (the nearby compiler tests already follow that pattern) so repeated local runs do not accumulate `agency-fp-*` trees.

egonSchiele (commented):
The production invalidation path looks sound, and the focused/CI coverage is strong. These are non-blocking test-hardening suggestions.

egonSchiele:
All five review comments addressed in 9b8ada8:

- **Templated fixture case** (the main ask): new test `TEMPLATED stdlib module records the implicit prelude index.agency dep`. The mock gained a `fake.templated` opt-in list; the fake `index.agency` stubs the full `PRELUDE_NAMES` surface so the auto-prelude import resolves. It asserts both the recorded manifest edge (`deps: ["stdlib/index.agency"]` through the real session→fingerprint→manifest path) and that the edge is load-bearing (editing index rewrites `tpl.js`). One correction to the PR description's deviation note: prelude names parse fine as `def` stubs — the earlier failure I attributed to parser keywords had a different cause, so the stub approach works and the deviation note's reasoning was wrong (the all-non-templated default remains, now as a convenience rather than a necessity).
- **Empty-root guard**: `afterEach` skips (and resets) when `root` was never assigned — `safeDeleteDirectory("")` would have resolved to the package root.
- **Null-prototype `visited`** in `dependencyFingerprint`, matching the manifest's own path-keyed maps.
- **Fixture-tree cleanup** in `depFingerprint.test.ts` via tracked roots + `afterEach`.
- **Platform-neutral paths** in the `stdlibHashFlavor` boundary test (`path.join` throughout).

68 tests green across the three touched files; typecheck clean.

egonSchiele (commented):
Re-reviewed at 9b8ada8. All five prior comments are addressed correctly: the stdlib boundary fixture is platform-neutral, temporary fingerprint trees are cleaned up, traversal uses a null-prototype path dictionary, test cleanup guards an unset root, and the new templated BuildSession integration case proves the implicit `std::index` edge is both recorded and load-bearing. I found no new actionable issues.

Focused verification: 91 tests passed across `buildManifest.test.ts`, `compileClosure.test.ts`, `depFingerprint.test.ts`, `manifestTracker.test.ts`, and `stdlibDeps.test.ts`. The current CI checks are also green.

Issue #809, opened 2026-08-05
Account for dispatched-but-failed LLM calls (needs smoltalk dispatch metadata)


egonSchiele:
## Problem

When an LLM call fails, it may or may not have cost money, and agency currently cannot tell which. `LLMClient.{text,embed,image}` return `Promise<Result<...>>`, and a failure surfaces two ways:

- the promise **rejects** (an exception during the call), or
- it resolves to a **`Result.failure`** value (the client returns a failure).

A `Result.failure` covers **both** a *pre-dispatch* failure (missing credentials, validation, unsupported provider — no request ever sent, so no spend) and a *post-dispatch* failure (the request reached the provider and then errored — possible spend). The `Result` carries no metadata to distinguish them.

For accurate cost accounting we want to count exactly the **dispatched-but-failed** calls toward `unknownCostCallCount` (which drives `pricingComplete` / the "spend may be a lower bound" signal). Today:

- `meteredDispatch` (`lib/runtime/recordPaidUsage.ts`) only counts **promise rejections**, so a resolved `Result.failure` from a paid call is missed entirely — even in the existing prompt path.
- Counting *every* resolved failure would over-flag pre-dispatch failures (e.g. a missing API key) as unknown-cost spend.

## The fix (smoltalk first)

The clean fix is in **smoltalk**: have it report, on a failure, whether the request actually reached the provider — e.g. a `dispatched: boolean` on the failure result. Then agency can count exactly the dispatched-but-failed calls as one unknown-cost attempt each, and leave genuine pre-dispatch failures uncounted.

Until smoltalk exposes that, agency keeps today's behavior (count outright throws) and treats resolved-`Result.failure` dispatch detection as a known limitation.

## Context

Surfaced during the "full cost & token breakdown" design (`docs/superpowers/specs/2026-08-04-full-cost-token-breakdown-design.md`), which specifies the `pricingComplete` / `unknownCostCallCount` accounting this feeds. This issue tracks the smoltalk + agency change so the two provider-failure paths are accounted consistently.

PR #811, opened 2026-08-06
feat(serve): full cost and token breakdown per invocation

egonSchiele:
## What

Replaces the flat per-model cost figure with a complete **cost and token breakdown** for every hosted invocation, keeps recoverable accounting across the subprocess IPC boundary, and surfaces it through `agency remote spend`. This is the agency-lang half of the full-cost-token-breakdown design; it pairs with a companion statelog schema and must ship as a pinned pair.

There are no users of the old shape yet, so this is a clean breaking change.

## The two figures

Each invocation now carries, and must not confuse:

- **Authoritative flat total** — `usage.cost` (full `CostBreakdown`) and `usage.tokens` (full `TokenBreakdown`). This is what gets billed.
- **Best-effort attribution** — `usage.entries`, one per `(kind, model)`. Reconciled against the flat total only when telemetry is complete; the total is never derived from summing entries.

`kind` is the API verb (`completion | embedding | image | manual`). A manual `addCost()` charge uses the model sentinel `""` (never null). Two independent completeness axes ride alongside: `pricingComplete` (did every call have a usable USD price?) and `usageComplete` (did all telemetry arrive?) — the latter a **sibling** of `usage`, flipping it to a trusted lower bound.

## Key rules (easy to get wrong)

- A price is valid only when `totalCost` is finite-nonnegative **and** `currency === "USD"`; otherwise the call is unpriced (all cost zero, `unknownCostCallCount` bumps). `totalCost: 0` is known-free.
- Cost components are best-effort and never reconciled to the total.
- `totalTokens` is authoritative when present; the fallback is kind-specific — embedding/image add only input+output because their cache counters may overlap input.
- All counts checked-add with saturation at `MAX_SAFE_INTEGER`; a saturation degrades `usageComplete`.
- Untrusted IPC input is recovered field-by-field and **never** drops money — an unusable entry is omitted while the flat total survives and `usageComplete` degrades.

## Layout

- `lib/runtime/invocationUsage.ts` — pure value layer + meter + IPC recovery.
- `lib/runtime/recordPaidUsage.ts` / `cost.ts` — one private `recordUsageDelta` sink; `addCost` validates finite-nonnegative first.
- `lib/runtime/prompt.ts`, `memory/manager.ts`, `stdlib/image.ts` — each provider outcome emits one observation (image accounts an empty result too — the provider still charged).
- `lib/runtime/costTelemetry.ts` / `ipc.ts` — new untrusted `invocationUsage` wire shape, recover-and-relay; legacy `costUsd` / `modelAttributionIncomplete` removed.
- `lib/cli/statelog/spendTypes.ts` + clients/renderers — new `ProjectSpend` shape and `agency remote spend --by-model` / `--by-kind`.
- `docs/dev/invocation-usage-accounting.md` — the architecture note (+ CLAUDE.md pointer).

## #809 boundary

Only a **rejected** provider promise counts as an unresolved attempt. A resolved `Result.failure` records nothing, because smoltalk can't yet distinguish a pre-dispatch failure (no spend) from a post-dispatch one (real spend). Closing that is deferred to agency-lang #809.

## Testing

- `pnpm run typecheck`, `pnpm run lint:structure`, and `make` all green.
- All affected unit/integration test files updated and passing (value layer, sink, sources, IPC recovery, serve adapter/integration, spend schema/clients/renderers/command).

### DIFF
--- a/packages/agency-lang/lib/cli/remote/render.ts
+++ b/packages/agency-lang/lib/cli/remote/render.ts
@@ -132,28 +132,38 @@ function formatUsd(amount: number): string {
   return `$${amount.toFixed(4)}`;
 }
 
-/** A count with thousands separators, in a fixed locale for deterministic output. */
-function formatCount(count: number): string {
+/** A count with thousands separators, in a fixed locale for deterministic
+ *  output. Accepts `bigint` so aggregate totals can be summed without crossing
+ *  `Number.MAX_SAFE_INTEGER`. */
+function formatCount(count: number | bigint): string {
   return count.toLocaleString("en-US");
+}
+
+/** A sub-$0.0001 dollar amount rendered with enough precision that a positive
+ *  value never collapses to `$0.0000`. Trailing zeros are trimmed. */
+function formatTinyUsd(amount: number): string {
+  const fixed = amount.toFixed(8).replace(/0+$/, "").replace(/\.$/, "");
+  return `$${fixed}`;
 }
 
 /** Prefix `≥` when the figure is a trusted lower bound (telemetry incomplete).
  *  A lower bound never uses the `<$0.0001` sentinel — `≥ <$0.0001` ("at least
- *  less than") is incoherent — so a sub-$0.0001 lower bound floors to
- *  `≥ $0.0000` (still true: the total is at least ~zero and may be higher). */
+ *  less than") is incoherent — but it must not floor a positive amount to
+ *  `≥ $0.0000` either (that hides a known positive spend), so a sub-$0.0001
+ *  lower bound keeps extra precision (e.g. `≥ $0.00004`). */
 function lowerBound(amount: number, complete: boolean): string {
   if (complete) {
     return formatUsd(amount);
   }
   if (amount > 0 && amount < 0.0001) {
-    return "≥ $0.0000";
+    return `≥ ${formatTinyUsd(amount)}`;
   }
   return `≥ ${formatUsd(amount)}`;
 }
 
 /** A count that keeps a lower-bound `≥` prefix when the figure is not trusted
  *  complete (telemetry incomplete or price unknown). */
-function lowerBoundCount(count: number, complete: boolean): string {
+function lowerBoundCount(count: number | bigint, complete: boolean): string {
   return complete ? formatCount(count) : `≥ ${formatCount(count)}`;
 }
 
@@ -304,10 +314,12 @@ function trustNotes(
 
 type SpendTotals = {
   totalCost: number;
-  inputTokens: number;
-  outputTokens: number;
-  invocationCount: number;
-  unpricedCallCount: number;
+  // Counters accumulate as bigint: each row's value is a safe integer, but a sum
+  // across many projects can cross Number.MAX_SAFE_INTEGER and silently round.
+  inputTokens: bigint;
+  outputTokens: bigint;
+  invocationCount: bigint;
+  unpricedCallCount: bigint;
   usageComplete: boolean;
   pricingComplete: boolean;
 };
@@ -316,14 +328,14 @@ function spendTotals(rows: AccountSpendRow[]): SpendTotals {
   return rows.reduce<SpendTotals>(
     (totals, row) => ({
       totalCost: totals.totalCost + row.spend.cost.totalCost,
-      inputTokens: totals.inputTokens + row.spend.tokens.inputTokens,
-      outputTokens: totals.outputTokens + row.spend.tokens.outputTokens,
-      invocationCount: totals.invocationCount + row.spend.invocationCount,
-      unpricedCallCount: totals.unpricedCallCount + row.spend.unpricedCallCount,
+      inputTokens: totals.inputTokens + BigInt(row.spend.tokens.inputTokens),
+      outputTokens: totals.outputTokens + BigInt(row.spend.tokens.outputTokens),
+      invocationCount: totals.invocationCount + BigInt(row.spend.invocationCount),
+      unpricedCallCount: totals.unpricedCallCount + BigInt(row.spend.unpricedCallCount),
       usageComplete: totals.usageComplete && row.spend.usageComplete,
       pricingComplete: totals.pricingComplete && row.spend.pricingComplete,
     }),
-    { totalCost: 0, inputTokens: 0, outputTokens: 0, invocationCount: 0, unpricedCallCount: 0, usageComplete: true, pricingComplete: true },
+    { totalCost: 0, inputTokens: 0n, outputTokens: 0n, invocationCount: 0n, unpricedCallCount: 0n, usageComplete: true, pricingComplete: true },
   );
 }
 
@@ -380,7 +392,7 @@ export function renderAccountSpend(rows: AccountSpendRow[], description: string)
   if (!totals.usageComplete) {
     lines.push(color.dim("Some projects have incomplete telemetry; totals are lower bounds."));
   }
-  if (totals.unpricedCallCount > 0) {
+  if (totals.unpricedCallCount > 0n) {
     lines.push(color.dim(`${formatCount(totals.unpricedCallCount)} unpriced call(s) total — cost may be understated.`));
   }
   return lines.join("\n");
--- a/packages/agency-lang/lib/cli/remote/render.ts
+++ b/packages/agency-lang/lib/cli/remote/render.ts
@@ -225,7 +225,7 @@ function groupBreakdown(rows: ModelKindSpend[], grouping: SpendGrouping): SpendG
   for (const row of rows) {
     const model = grouping.byModel ? row.model : null;
     const kind = grouping.byKind ? row.kind : null;
-    const key = `${model ?? ""} ${kind ?? ""}`;
+    const key = JSON.stringify([model, kind]);
     let group = groups[key];
     if (group === undefined) {
       group = { model, kind, cost: zeroCost(), tokens: zeroTokens() };
@@ -357,8 +357,10 @@ export function renderAccountSpend(rows: AccountSpendRow[], description: string)
   const tableRows = sorted.map((row) => [
     row.deletedAt === null ? row.projectSlug : `${row.projectSlug} (deleted)`,
     lowerBound(row.spend.cost.totalCost, spendIsComplete(row.spend)),
-    formatCount(row.spend.tokens.inputTokens),
-    formatCount(row.spend.tokens.outputTokens),
+    // Token counts are a lower bound whenever telemetry is incomplete — that is
+    // `usageComplete` alone (unpriced calls affect cost, not counts).
+    lowerBoundCount(row.spend.tokens.inputTokens, row.spend.usageComplete),
+    lowerBoundCount(row.spend.tokens.outputTokens, row.spend.usageComplete),
     formatCount(row.spend.invocationCount),
     formatCount(row.spend.unpricedCallCount),
   ]);
@@ -366,8 +368,8 @@ export function renderAccountSpend(rows: AccountSpendRow[], description: string)
   tableRows.push([
     "TOTAL",
     lowerBound(totals.totalCost, totals.usageComplete && totals.pricingComplete),
-    formatCount(totals.inputTokens),
-    formatCount(totals.outputTokens),
+    lowerBoundCount(totals.inputTokens, totals.usageComplete),
+    lowerBoundCount(totals.outputTokens, totals.usageComplete),
     formatCount(totals.invocationCount),
     formatCount(totals.unpricedCallCount),
   ]);
--- a/packages/agency-lang/lib/cli/remote/render.ts
+++ b/packages/agency-lang/lib/cli/remote/render.ts
@@ -11,7 +11,14 @@ import type {
   KeySummary,
   CreatedKey,
 } from "../statelog/accountClient.js";
-import type { ProjectSpend, AccountSpendRow } from "../statelog/spendTypes.js";
+import type {
+  ProjectSpend,
+  AccountSpendRow,
+  CostBreakdown,
+  TokenBreakdown,
+  ModelKindSpend,
+  UsageKind,
+} from "../statelog/spendTypes.js";
 
 const NONE = "—";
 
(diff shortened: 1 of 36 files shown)


packages/agency-lang/lib/runtime/invocationUsage.ts
```diff
@@ -1,298 +1,402 @@
...
+  return { ...tokens };
+}
+
+/** Build a cost breakdown from a (trusted or untrusted) estimate-shaped object.
+ *  A price is valid only when `totalCost` is finite-nonnegative AND currency is
+ *  USD; otherwise every field is zero and the call is unpriced (`priced=false`).
+ *  Named components are copied independently (best-effort, no sum-to-total). */
+function buildCost(raw: unknown): { cost: CostBreakdown; priced: boolean } {
+  const obj = raw && typeof raw === "object" ? (raw as Record<string, unknown>) : undefined;
+  const priced = obj !== undefined && isFiniteNonNeg(obj.totalCost) && obj.currency === "USD";
```
egonSchiele, line 141: **P1 — Preserve deterministic image billing.** Requiring USD here exposes an existing malformed producer: `DeterministicClient.image()` returns `{ totalCost: DETERMINISTIC_IMAGE_COST } as any` without `currency`. Every deterministic image is therefore normalized to $0/unpriced, so `getCost()` stays zero and image cost guards no longer trip. This is the cause of both image-generation CI failures. Please update that producer to return a valid USD estimate (and remove its cast), then keep a regression assertion for the deterministic image cost/guard path.

packages/agency-lang/lib/runtime/memory/manager.ts
```diff
@@ -284,12 +292,18 @@ export class MemoryManager {
...
-      // cost budget. Mirrors the post-completion charge that
-      // `prompt.ts` performs for agency-side `llm()` calls, so a
-      // `withCostGuard($X)` wrapping the agent now sees memory's
-      // extraction / compaction / tier-3 spend too.
-      chargeCostIfInFrame(result.value.cost?.totalCost ?? 0);
+      // Account this call's full cost/tokens against the surrounding branch.
+      // Mirrors the post-completion accounting that `prompt.ts` performs for
+      // agency-side `llm()` calls, so a `withCostGuard($X)` wrapping the agent
+      // now sees memory's extraction / compaction / tier-3 spend too.
+      recordMemoryUsageIfInFrame({
```
egonSchiele, line 311: **P1 — Meter rejected memory/image dispatch promises too.** The plan requires rejected provider promises to record one unresolved attempt, but only the prompt path uses `meteredDispatch`. Both memory text and `_embed()` call their clients directly, as does `_generateImage()`, so a promise rejection after dispatch leaves `unknownCostCallCount` unchanged and can incorrectly report `pricingComplete: true`. Please wrap the three dispatch promises with `meteredDispatch(..., kind, ...)` (while preserving the no-frame behavior for directly constructed `MemoryManager` tests). This is separate from the deliberately deferred handling of resolved `Result.failure` values.

packages/agency-lang/lib/runtime/invocationUsage.ts
```diff
@@ -1,298 +1,402 @@
...
+ *  whole message is unusable (not an object). */
+export function normalizeIpcUsageDelta(raw: unknown): NormalizedDelta | null {
+  if (raw === null || typeof raw !== "object") {
+    return null;
+  }
+  const obj = raw as Record<string, unknown>;
+  let attributionLost = obj.attributionLost === true;
+
+  const { cost, priced } = buildCost(obj.cost);
+  const { tokens, malformed } = buildTokens(obj.tokens, "completion", true);
```
egonSchiele: **P1 — Do not apply trusted completion fallback semantics to untrusted IPC.** IPC policy says each absent/malformed token field becomes 0 and degrades completeness, but `buildTokens(..., "completion", true)` derives a missing `totalTokens` and hard-codes completion semantics. For example, an image delta with valid `inputTokens: 100`, `cachedInputTokens: 30`, and no total is recovered as 130 instead of preserving 100 as the safe lower bound; it can also remain `attributionLost: false`. Please give wire recovery its own field-by-field path: preserve valid siblings, zero missing/malformed fields, and degrade; do not infer a provider-kind fallback at this boundary.

packages/agency-lang/lib/runtime/invocationUsage.ts
```diff
@@ -1,298 +1,402 @@
...
+  }
+  const kind = obj.kind;
+  const model = obj.model;
+  const modelOk = kind === "manual"
+    ? model === ""
+    : typeof model === "string" && model.length > 0;
+  if (!modelOk) {
+    return "invalid";
+  }
+  const { cost } = buildCost(obj.cost);
```
egonSchiele: **P1 — Propagate malformed nested-entry recovery into `attributionLost`.** `buildCost` and `buildTokens` expose whether the nested entry was invalid/malformed, but those signals are discarded here, so a valid kind/model with malformed entry components produces a supposedly complete snapshot whose flat totals and attribution do not reconcile. Likewise, a missing entry currently remains complete even when the flat delta contains measurable cost/tokens (only a genuine all-zero unresolved attempt should be valid without an entry). Please return the recovery status from `recoverIpcEntry` and mark the delta incomplete while retaining valid flat fields and any usable entry.

packages/agency-lang/lib/runtime/invocationUsage.ts
```diff
@@ -1,298 +1,402 @@
...
+  let unknownCostCallCount = 0;
+  if (isSafeCount(obj.unknownCostCallCount)) {
+    unknownCostCallCount = obj.unknownCostCallCount;
+  } else if (obj.unknownCostCallCount !== undefined) {
+    attributionLost = true; // present-but-malformed
+  } else {
+    attributionLost = true; // missing
+  }
+  if (!priced) {
+    unknownCostCallCount = checkedAddCount(unknownCostCallCount, 1).value;
```
egonSchiele: **P2 — Preserve the overflow signal from the malformed-cost bump.** When `unknownCostCallCount` is already `Number.MAX_SAFE_INTEGER`, this required +1 saturates, but dropping `.saturated` leaves `attributionLost` false and the resulting snapshot can claim completeness after losing a count. Please retain the checked-add result and degrade when it saturates, matching the meter accumulation policy.

packages/agency-lang/lib/cli/remote/render.ts
```diff
@@ -212,16 +356,16 @@ export function renderAccountSpend(rows: AccountSpendRow[], description: string)
   const sorted = sortAccountRows(rows);
   const tableRows = sorted.map((row) => [
     row.deletedAt === null ? row.projectSlug : `${row.projectSlug} (deleted)`,
-    lowerBound(row.spend.pricedCost, row.spend.usageComplete),
-    formatCount(row.spend.inputTokens),
-    formatCount(row.spend.outputTokens),
+    lowerBound(row.spend.cost.totalCost, spendIsComplete(row.spend)),
+    formatCount(row.spend.tokens.inputTokens),
```
egonSchiele: **P2 — Mark account token totals as lower bounds when telemetry is incomplete.** `usageComplete` applies to the entire usage snapshot, not just dollars. These input/output token columns (and the aggregate TOTAL row below) are currently rendered as exact counts even while the cost column and note say telemetry is incomplete. Please use `lowerBoundCount` with each row’s `usageComplete`, and with aggregate completeness for TOTAL.

packages/agency-lang/lib/runtime/invocationUsage.ts
egonSchiele: Fixed. DeterministicClient.image() now returns a valid USD estimate ({ inputCost: 0, outputCost: DETERMINISTIC_IMAGE_COST, totalCost: DETERMINISTIC_IMAGE_COST, currency: "USD" }) with the cast removed. Added a regression in image-client.test.ts: the producer asserts currency === "USD", and a new describe drives _generateImage through the real DeterministicClient to assert localCost accrues DETERMINISTIC_IMAGE_COST and a guard tighter than it trips.

packages/agency-lang/lib/runtime/memory/manager.ts
egonSchiele: Fixed. Added meteredMemoryDispatch(kind, dispatch) in manager.ts wrapping _text and _embed, and wrapped _generateImage's image() call in meteredDispatch(ctx, stack, "image", ...). A post-dispatch rejection now records one unresolved attempt. meteredMemoryDispatch runs the dispatch directly when there is no frame, preserving the direct-construction MemoryManager behavior. Resolved Result.failure is still not metered (deferred to #809). New tests: image.test.ts rejected-dispatch records one attempt / pricingComplete false.

packages/agency-lang/lib/runtime/invocationUsage.ts
egonSchiele: Fixed. IPC token recovery has its own path now (buildIpcTokens): each field is a nonnegative-safe-int-or-zero-and-degrade, with no benign absence at this boundary. A missing/malformed total falls back to a conservative input+output lower bound (never summing cache). Your image example now recovers 100, not 130, and degrades. buildTokens keeps the trusted provider fallback and is no longer used for wire input.

packages/agency-lang/lib/runtime/invocationUsage.ts
egonSchiele: Fixed. recoverIpcEntry now returns { status, entry?, malformed } and normalizeIpcUsageDelta degrades when the entry's own cost/tokens were malformed (keeping the usable entry). It also degrades when a measurable flat delta arrives with no entry — only an all-zero unresolved attempt is legitimately entry-less. New tests cover the malformed-entry-survives-and-degrades and measurable-flat-without-entry cases.

packages/agency-lang/lib/runtime/invocationUsage.ts
egonSchiele: Fixed. The malformed-cost bump now keeps the checked-add result and sets attributionLost when it saturates, matching the meter accumulation policy. Test: a bump against unknownCostCallCount === MAX_SAFE_INTEGER degrades.

packages/agency-lang/lib/cli/remote/render.ts
egonSchiele: Fixed. Account token columns and the TOTAL row now use lowerBoundCount keyed on usageComplete (per row, and aggregate for TOTAL). Kept them keyed on usageComplete rather than pricing, since unpriced calls affect cost not counts. Tests assert the token ≥ marker on an incomplete row and its absence for an unpriced-but-complete row.

packages/agency-lang/lib/cli/remote/render.ts
```diff
@@ -144,42 +151,179 @@ function lowerBound(amount: number, complete: boolean): string {
...
+ *  flags → one group per pair; one flag → collapse the other axis; the caller
+ *  only reaches here when at least one flag is set. Sorted by total cost
+ *  descending, then model then kind ascending. */
+function groupBreakdown(rows: ModelKindSpend[], grouping: SpendGrouping): SpendGroup[] {
+  const groups: Record<string, SpendGroup> = Object.create(null);
+  for (const row of rows) {
+    const model = grouping.byModel ? row.model : null;
+    const kind = grouping.byKind ? row.kind : null;
+    const key = `${model ?? ""} ${kind ?? ""}`;
+    let group = groups[key];
```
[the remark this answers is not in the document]
egonSchiele: Fixed. The grouping key was a raw NUL byte; replaced it with JSON.stringify([model, kind]), which is unambiguous and readable.

packages/agency-lang/lib/runtime/costTelemetry.ts
```diff
@@ -83,18 +60,30 @@ function trySend(msg: IpcUsageMessage): void {
...
 
-/** Relay a full usage delta to the parent, once. Skips an all-zero delta (a
- *  no-op charge carries nothing worth relaying). */
-export function sendInvocationUsageToParent(delta: InvocationUsageDelta): void {
+/** A delta carries nothing worth relaying: no cost, no tokens, no unknown-cost
+ *  call, no attribution entry, and no attribution loss. A no-op charge (e.g.
+ *  `addCost(0)`) is skipped so it never spams the parent channel. */
+function isNoOpDelta(delta: NormalizedDelta): boolean {
+  if (delta.entry !== undefined) return false;
+  if (delta.attributionLost) return false;
```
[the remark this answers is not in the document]
egonSchiele: Fixed. isNoOpDelta now also treats a value-less entry as no-op: a delta whose entry has all-zero cost and tokens (i.e. addCost(0)) is suppressed. Test added in costTelemetry.test.ts, plus one asserting a manual entry that DOES carry cost is still sent.

packages/agency-lang/lib/runtime/invocationUsage.ts
```diff
@@ -1,298 +1,402 @@
...
+    totalTokens = fb.value;
+    malformed = malformed || fb.saturated;
+  } else if (isSafeCount(rawTotal)) {
+    totalTokens = rawTotal;
+  } else {
+    const fb = fallbackTotalTokens(kind, partial);
+    totalTokens = fb.value;
+    malformed = true; // present-but-malformed provider total
+  }
+  return { tokens: { ...partial, totalTokens }, malformed };
```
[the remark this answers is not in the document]
egonSchiele: Fixed. Wire recovery no longer uses buildTokens with completion semantics; buildIpcTokens gives untrusted totals a conservative input+output lower bound (never summing cache) and marks malformed. See the fuller reply on the same point below.

packages/agency-lang/lib/runtime/costTelemetry.ts
```diff
@@ -1,44 +1,36 @@
...
+ *  reaches the accounting sink. The trusted SEND-side payload is a
+ *  `NormalizedDelta` (this message minus `type`). */
 export type IpcInvocationUsageMessage = {
   type: "invocationUsage";
-  pricedCost: number;
-  inputTokens: number;
-  outputTokens: number;
-  unknownCostCallCount: number;
-  attribution?: UsageAttribution;
+  cost?: unknown;
```
egonSchiele, line 29: why is everything `unknown`?

packages/agency-lang/lib/runtime/invocationUsage.ts
```diff
@@ -1,298 +1,471 @@
...
+  | { status: "invalid" }
+  | { status: "ok"; entry: UsageEntry; malformed: boolean };
+
+/** Recover an untrusted entry: valid iff kind ∈ UsageKind and the model sentinel
+ *  invariant holds. Its cost/tokens are normalized independently and NEVER
+ *  override the flat totals. `"invalid"` omits the entry and degrades; `"ok"`
+ *  reports whether the entry's OWN components were malformed (an unpriced entry
+ *  cost or a malformed token field), so the caller can degrade while keeping the
+ *  usable entry. */
+function recoverIpcEntry(raw: unknown): IpcEntryRecovery {
```
egonSchiele, line 350: is it weird that so much IPC-specific logic lives in this file?

packages/agency-lang/lib/stdlib/image.ts
```diff
@@ -72,29 +79,41 @@ export async function _generateImage(
...
+    reportedModel: gen.model,
+    configuredModel: model,
+    cost: gen.costEstimate,
+    tokens: gen.tokenUsage,
   });
-  // addCost accrues localCost, bills guards, and enforces limits (may throw a
-  // guard trip, which must propagate — must be last).
-  addCost(gen.costEstimate?.totalCost ?? 0);
+  addTokens(gen.tokenUsage?.totalTokens ?? 0);
+  if (first) {
```
egonSchiele, line 100: why was this if statement added? It wasn't there before?

packages/agency-lang/lib/runtime/costTelemetry.ts
egonSchiele: Because this is the untrusted RECEIVE shape. A child process — possibly version-skewed or otherwise misbehaving — can put anything on the wire, so typing the fields as `unknown` forces the parent to validate every field through `normalizeIpcUsageDelta` before it touches the meter; nothing can accidentally trust wire data. The trusted SEND side is a plain `NormalizedDelta` (this message minus `type`), which is why `sendInvocationUsageToParent` takes that concrete type. So the asymmetry is deliberate: concrete going out, `unknown` coming in.

packages/agency-lang/lib/runtime/invocationUsage.ts
egonSchiele: Fair observation. The reason it is here: the untrusted recovery (`normalizeIpcUsageDelta` / `buildIpcTokens` / `recoverIpcEntry` / `ipcCounter`) produces the exact same output type as the trusted normalizer (`NormalizedDelta`) and shares the same primitives — `buildCost`, `isSafeCount`, `checkedAddCount`, `zeroCost/zeroTokens`, `UsageEntry`, `isUsageKind`. I kept the two normalizers as a matched pair (trusted vs untrusted, same value out) so the contract stays visible in one place; splitting would either duplicate those helpers or create an import back into this file. That said, if the wire-recovery surface grows much more, extracting it into an `invocationUsageIpc.ts` that imports the shared primitives would be reasonable — happy to do that split now if you'd prefer it.

packages/agency-lang/lib/stdlib/image.ts
egonSchiele: It guards the `imageGeneration` statelog event, and it is new because the no-image path changed. Before this PR the function returned `failure("...returned no images")` early, BEFORE any accounting — so when execution reached the event, `first` always existed and the call was unconditional. This PR accounts usage even for a successful-but-empty provider result (the provider still charged us), so the empty case now falls through to the recording/tokens/enforceGuards block instead of returning early. The `imageGeneration` event's contract requires an actual image, so it must be skipped when there is none — hence `if (first)` — while cost/tokens are still recorded and guards still enforced. The empty case then returns the failure after the guard gate.

packages/agency-lang/lib/runtime/memory/manager.ts
```diff
@@ -363,11 +389,18 @@ export class MemoryManager {
...
-      // unknown" and the charge is skipped.
-      chargeCostIfInFrame(result.value.costEstimate?.totalCost ?? 0);
+      // Same rationale as `_text` above — embed calls contribute to the
+      // surrounding branch's cost/token totals too. `EmbedResult.costEstimate`
+      // and `tokenUsage` (smoltalk's names) are optional; an absent estimate is
+      // recorded as an unpriced embedding rather than dropped.
+      recordMemoryUsageIfInFrame({
+        type: "provider",
+        kind: "embedding",
+        reportedModel: result.value.model,
```
egonSchiele, line 405: **P1 — Account a successful embedding response even when it contains no vectors.** The `!vector` branch above throws before reaching this call. If the provider resolves successfully with `embeddings: []` plus a valid `costEstimate`/`tokenUsage`, that paid usage is omitted from guards, the invocation meter, and IPC. This is the embedding analogue of the successful image response with no images that this PR already handles correctly. Please record the provider observation before the structural no-vector return/throw, emit `embedCompletion` only when a vector exists, enforce guards after accounting, and let a guard trip win over the no-vector error. A regression with a priced empty-vector result would pin it.

packages/agency-lang/lib/runtime/costTelemetry.ts
```diff
@@ -83,18 +60,37 @@ function trySend(msg: IpcUsageMessage): void {
...
+}
+
+/** A delta carries nothing worth relaying: no cost, no tokens, no unknown-cost
+ *  call, no attribution loss, and — crucially — no VALUE-BEARING entry. A no-op
+ *  charge such as `addCost(0)` still produces a manual entry, but with all-zero
+ *  cost and tokens, so it too is skipped and never spams the parent channel. */
+function isNoOpDelta(delta: NormalizedDelta): boolean {
+  if (delta.attributionLost) return false;
+  if (delta.unknownCostCallCount !== 0) return false;
+  if (delta.cost.totalCost !== 0) return false;
```
egonSchiele: **P1 — This no-op check still drops valid breakdown data across subprocesses.** It checks only `totalCost`, but the contract deliberately preserves independently normalized named components even when they do not reconcile to the authoritative total. A known-free provider result such as `{ inputCost: 0.01, totalCost: 0, currency: "USD" }` is retained in-process but suppressed here, so the parent loses both that component and the provider `(kind, model)` entry. Please test all cost components, and suppress an all-zero entry only for the manual `addCost(0)` sentinel; an all-zero provider entry still carries attribution and should remain topology-invariant.

packages/agency-lang/lib/cli/remote/render.ts
```diff
@@ -144,42 +151,179 @@ function lowerBound(amount: number, complete: boolean): string {
...
+): string {
+  if (spend.invocationCount === 0) {
+    return `No spend in ${description}.`;
+  }
+  const complete = spendIsComplete(spend);
   const lines = [
     `${color.bold("Spend:")} ${slug}  ${color.dim(`(${description})`)}`,
-    `  ${color.bold("Cost:")}         ${lowerBound(spend.pricedCost, spend.usageComplete)}`,
-    `  ${color.bold("Tokens:")}       ↑ ${formatCount(spend.inputTokens)}  ↓ ${formatCount(spend.outputTokens)}`,
+    `  ${color.bold("Cost:")}         ${lowerBound(spend.cost.totalCost, complete)}`,
```
egonSchiele, line 273: **P2 — A positive lower bound is still rendered as zero.** The renderer contract says `$0.0000` is only for true zero, but `lowerBound()` turns an incomplete known spend of `$0.00004` into `≥ $0.0000`, hiding the positive amount. `≥ <$0.0001` would indeed be incoherent, but rounding the known lower bound down to zero is also misleading. Please retain enough precision for this case (for example `≥ $0.00004`) and assert that a positive lower bound never contains `$0.0000`.

packages/agency-lang/lib/cli/remote/render.ts
```diff
@@ -144,42 +151,179 @@ function lowerBound(amount: number, complete: boolean): string {
...
+  pricingComplete: boolean;
 };
 
 function spendTotals(rows: AccountSpendRow[]): SpendTotals {
   return rows.reduce<SpendTotals>(
     (totals, row) => ({
-      pricedCost: totals.pricedCost + row.spend.pricedCost,
-      inputTokens: totals.inputTokens + row.spend.inputTokens,
-      outputTokens: totals.outputTokens + row.spend.outputTokens,
+      totalCost: totals.totalCost + row.spend.cost.totalCost,
```
egonSchiele, line 317: **P2 — Aggregate account counters without unsafe integer addition.** Each project count may independently be a valid safe integer, but summing two rows can cross `Number.MAX_SAFE_INTEGER`; e.g. `MAX_SAFE_INTEGER + 2` is rounded and the TOTAL is still rendered as exact. Input/output tokens, invocation count, and unpriced-call count all use this path. Please aggregate presentation counters as `bigint` (and let the formatters accept it), or checked-saturate with a separate aggregate-count lower-bound signal so the cost completeness flag is not incorrectly degraded.

packages/agency-lang/lib/runtime/memory/manager.ts
egonSchiele: Fixed. _embed now accounts the provider observation whether or not a vector came back: the embedCompletion event fires only when a vector exists, recordMemoryUsageIfInFrame records + enforces guards, and the no-vector error is thrown only afterward — so a guard trip wins over it (the embedding analogue of the empty-image path). Two regressions added: a priced empty-vector result records usage (localCost + meter + an embedding entry) then throws the no-vector error, and a tight guard trips instead of that error.

packages/agency-lang/lib/runtime/costTelemetry.ts
egonSchiele: Fixed. isNoOpDelta now checks every named cost component (via costIsZero), so a known-free result like { inputCost: 0.01, totalCost: 0 } is relayed, not suppressed. And an all-zero entry is treated as no-op only when it is the manual sentinel (kind === "manual"); an all-zero PROVIDER entry keeps its (kind, model) attribution and is always relayed. Tests: the known-free-component case and the all-zero-provider-entry case both send.

packages/agency-lang/lib/cli/remote/render.ts
egonSchiele: Fixed. A sub-$0.0001 lower bound now renders with enough precision to stay visible (formatTinyUsd → e.g. >= $0.00004) instead of flooring to >= $0.0000. Still avoids the incoherent >= <$0.0001. Test asserts the cost line contains >= $0.00004 and never the floored >= $0.0000 form.

packages/agency-lang/lib/cli/remote/render.ts
egonSchiele: Fixed. spendTotals now aggregates the four presentation counters (input/output tokens, invocation count, unpriced-call count) as bigint, and formatCount/lowerBoundCount accept number | bigint. A TOTAL crossing Number.MAX_SAFE_INTEGER stays exact instead of silently rounding, and the cost completeness flag is untouched. Regression: two rows summing to 9,007,199,254,740,993 render exactly (Number would round to ...992).

packages/agency-lang/lib/cli/remote/render.ts
```diff
@@ -125,61 +132,210 @@ function formatUsd(amount: number): string {
...
+ *  output. Accepts `bigint` so aggregate totals can be summed without crossing
+ *  `Number.MAX_SAFE_INTEGER`. */
+function formatCount(count: number | bigint): string {
   return count.toLocaleString("en-US");
 }
 
+/** A sub-$0.0001 dollar amount rendered with enough precision that a positive
+ *  value never collapses to `$0.0000`. Trailing zeros are trimmed. */
+function formatTinyUsd(amount: number): string {
+  const fixed = amount.toFixed(8).replace(/0+$/, "").replace(/\.$/, "");
```
egonSchiele, line 147: **P2 — Values below the fixed eight-decimal precision still collapse to zero.** For example, `formatTinyUsd(1e-10)` computes `"0.00000000"`, trims it to `"0"`, and renders the positive lower bound as `≥ $0`. The stated invariant is that *any* positive spend stays visible. Please detect a zero rounded result and fall back to scientific notation (or choose precision from the magnitude), and add a case below `1e-8`.

packages/agency-lang/lib/cli/remote/render.ts
```diff
@@ -125,61 +132,210 @@ function formatUsd(amount: number): string {
...
+    const kind = grouping.byKind ? row.kind : null;
+    const key = JSON.stringify([model, kind]);
+    let group = groups[key];
+    if (group === undefined) {
+      group = { model, kind, cost: zeroCost(), tokens: zeroTokens() };
+      groups[key] = group;
+    }
+    addCostInto(group.cost, row.cost);
+    addTokensInto(group.tokens, row.tokens);
+  }
```
egonSchiele, line 233: **P2 — Grouped breakdown counters still cross the safe-integer boundary.** Account TOTALs now use `bigint`, but `--by-model`/`--by-kind` can also collapse multiple individually safe breakdown rows through `addTokensInto()`, which uses number addition. Two same-kind rows with `MAX_SAFE_INTEGER` and `2` input tokens render the rounded `...992` as exact. Please aggregate group token counters as `bigint` too (or checked-saturate with a visible lower-bound marker), and cover a collapsed group crossing the boundary.

packages/agency-lang/lib/cli/remote/render.ts
egonSchiele: Fixed. formatTinyUsd now detects a result that rounds to zero at 8 decimals and falls back to scientific notation, so a positive amount below 1e-8 renders (e.g. >= $1e-10) instead of >= $0. Test added for a 1e-10 lower bound.

packages/agency-lang/lib/cli/remote/render.ts
egonSchiele: Fixed. groupBreakdown now accumulates group token counters as bigint (input/output), matching the account TOTAL; cost stays a float total. The old number-addition group helpers were removed. Regression: two same-kind breakdown rows collapsed by --by-kind with MAX_SAFE_INTEGER and 2 input tokens render 9,007,199,254,740,993 exactly, not the rounded ...992.

packages/agency-lang/lib/cli/remote/render.ts
```diff
@@ -125,61 +132,197 @@ function formatUsd(amount: number): string {
...
+
 /** Prefix `≥` when the figure is a trusted lower bound (telemetry incomplete).
  *  A lower bound never uses the `<$0.0001` sentinel — `≥ <$0.0001` ("at least
- *  less than") is incoherent — so a sub-$0.0001 lower bound floors to
- *  `≥ $0.0000` (still true: the total is at least ~zero and may be higher). */
+ *  less than") is incoherent — but it must not floor a positive amount to
+ *  `≥ $0.0000` either (that hides a known positive spend), so a sub-$0.0001
+ *  lower bound keeps extra precision (e.g. `≥ $0.00004`). */
 function lowerBound(amount: number, complete: boolean): string {
   if (complete) {
```
egonSchiele, line 160: **P2 — Lower-bound formatting must not round upward.** `toFixed(8)` uses nearest rounding: for example `6e-9` becomes `0.00000001`, so this renders `≥ $0.00000001` even though the known amount is only `$0.000000006`. The normal-size path has the same issue (`0.00019` can render as `≥ $0.0002`). Once a value is prefixed with `≥`, upward rounding makes the statement mathematically false. Please use a dedicated lower-bound formatter that floors at the selected precision (falling back to scientific notation when needed), and test values just above a rounding midpoint in both branches.

PR #812, opened 2026-08-06
agency run --model: pick a run's default model on the command line

egonSchiele:
## What this does

`agency run --model` and the `agency <file>` shorthand set the default model for a run:

```bash
agency run --model claude-opus-4-8 greet.agency
agency --model openrouter/anthropic/claude-sonnet-4 greet.agency
```

A model set explicitly in Agency code still wins — the flag feeds the lowest layer of an existing precedence chain, so `setModel()` and `llm({ model })` override it with no new logic.

## Reading the value

Split on the **first** slash only. Anything before it is the provider.

| you write | provider | model |
| --- | --- | --- |
| `gpt-4o-mini` | none set; smoltalk infers `openai` | `gpt-4o-mini` |
| `anthropic/claude-opus-4-8` | `anthropic` | `claude-opus-4-8` |
| `openrouter/anthropic/claude-sonnet-4` | `openrouter` | `anthropic/claude-sonnet-4` |
| `my-company/my-tune` | `my-company` | `my-tune` |

First-slash-only is what makes row 3 work: an OpenRouter identifier contains a slash of its own. Row 4 is why there is no known-provider list — `client.providerModules` can register a provider under any name at startup, so at flag-parse time a custom provider and a typo are indistinguishable.

## Model and provider are two settings, not one

This is the part worth reviewing carefully. They merge **field by field** — `getSmoltalkConfig` and the `runPrompt` merge are both shallow spreads, and `setModel()` writes only `{ model }`. So a provider from a lower layer survives a higher layer that changes only the model.

Three rules follow:

- **A bare model drops any inherited `defaultProvider`**, letting smoltalk infer one from the name. The key is removed rather than set to `undefined` — either satisfies the code generator, which tests truthiness, but removing it keeps the resolved config free of dangling fields.
- **`provider/model` sets both.**
- **A stated provider is sticky** under a later `setModel("other")`. Code that wants to move provider too must pass both.

**Behaviour change worth knowing:** if your `agency.json` sets `defaultProvider` to route through a proxy, a bare `--model` bypasses it. Name the provider (`--model litellm/gpt-4o-mini`) to keep it.

## Validation

A bare name must be in the hosted **text** catalog — `_listHostedModels()`, not smoltalk's `getAllModels()`, which also returns image, embedding and speech models and would accept `gpt-image-1` as a chat model.

```
$ agency run --model gpt-4o-minii greet.agency
error: option '--model <name>' argument 'gpt-4o-minii' is invalid.
  Unknown model "gpt-4o-minii".
  Did you mean "gpt-4o-mini"?
  For a model from another provider, write provider/model — e.g. openrouter/gpt-4o-minii.
  Run `agency models list` to see the catalog.
```

This runs before any compilation, so a typo costs nothing. A prefixed name is never checked. The escape for a model missing from the baked catalog is the prefix form — note `agency models refresh` prints to stdout and persists nothing, so it cannot affect a later process's validation.

## Testing

155 tests across 9 files, plus the full CLI integration suite against a packed tarball.

Three are worth calling out because they were written to fail for the right reason:

- **The Commander adapter.** `.option(..., (value) => resolveModelFlag(value))` looks like decoration. It isn't: Commander calls a parser with `(value, previous)`, so passing the resolver directly hands the previously parsed object to its `catalogNames` parameter when `--model` repeats. I removed the adapter, watched the repeated-flag test fail, and restored it.
- **Precedence.** Four cases assert on the config the client actually receives, via the `RecordingClient` this suite already had. The fourth calls `runPrompt` directly, because the TypeScript `agency.llm` facade forwards only `model` and `maxTokens` — a branch-level override would pass even if the per-call spread stopped working.
- **Ordering.** The invalid-model case compiles a fresh source name and asserts its output was never written, which is what proves validation precedes compilation rather than following it.

Also covered: `--model` after the filename is forwarded to the program and **not** validated, using an attached `--model=…` value so a successful validation cannot masquerade as no validation.

## Not included

`docs/site/**` is untouched per the standing rule, so `docs/site/cli/run.md` does not list `--model` yet. Developer documentation is in `docs/dev/config.md`.

No `--fastmodel`, `--slowmodel` or `--provider`; those stay specific to `agency agent`, which needs several model slots. `agency test`, `agency eval` and `agency serve` are unchanged.

The spec, plan and their reviews are in `docs/superpowers/`.

### DIFF
--- a/packages/agency-lang/docs/dev/config.md
+++ b/packages/agency-lang/docs/dev/config.md
@@ -26,6 +26,47 @@ Where applied: sources 1⊕2 at the CLI (baked into the generated program);
 source 3 at runtime, in the `RuntimeContext` constructor. Inspect the resolved
 result with `agency config show` (secrets masked; `--show-secrets` to reveal).
 
+## What `--model` means
+
+`agency run --model` and the `agency <file>` shorthand set the run's default
+model. The value is parsed by `resolveModelFlag` in `lib/cli/modelFlag.ts` and
+mapped by `applyCliFlags`:
+
+| you write | `client.defaultModel` | `client.defaultProvider` |
+| --- | --- | --- |
+| `gpt-4o-mini` | `gpt-4o-mini` | **deleted** |
+| `anthropic/claude-opus-4-8` | `claude-opus-4-8` | `anthropic` |
+| `openrouter/anthropic/claude-sonnet-4` | `anthropic/claude-sonnet-4` | `openrouter` |
+
+The split is on the **first** slash only, which is what lets an OpenRouter model
+identifier survive as the model name.
+
+Three things about this are easy to get wrong later.
+
+**A bare model deletes the provider rather than leaving it.** The code generator
+emits a provider only when one is configured, so deleting it is how smoltalk is
+allowed to infer the provider from the model name. If your `agency.json` sets
+`defaultProvider` to route through a proxy, a bare `--model` bypasses that
+proxy; name it (`--model litellm/gpt-4o-mini`) to keep it. The mapping deletes
+the key, rather than setting it to `undefined`, because the generator checks for
+presence.
+
+**A stated provider is sticky.** The layers merge field by field
+(`lib/runtime/state/context.ts` `getSmoltalkConfig`), so a provider set by the
+flag survives a later `setModel("other")` in Agency code — the pair becomes that
+provider plus the new model. Code that wants to move provider too must say
+`setLlmOptions({ model, provider })`, or pass both per call. The precedence
+cases in `lib/runtime/agencyLlm.test.ts` pin all four combinations.
+
+**Only a bare name is validated.** It is checked against the hosted **text**
+models from `_listHostedModels()` — not smoltalk's `getAllModels()`, which also
+returns image, embedding and speech models. A prefixed name is never checked,
+because `client.providerModules` can register a provider under any name at
+startup, so at flag-parse time a custom provider and a typo are the same thing.
+The escape for a model missing from the baked catalog is the prefix form; note
+that `agency models refresh` prints JSON to stdout and persists nothing, so it
+cannot affect a later process's validation.
+
 ## All options
 
 ### Basic
--- a/packages/agency-lang/docs/dev/config.md
+++ b/packages/agency-lang/docs/dev/config.md
@@ -43,13 +43,16 @@ identifier survive as the model name.
 
 Three things about this are easy to get wrong later.
 
-**A bare model deletes the provider rather than leaving it.** The code generator
-emits a provider only when one is configured, so deleting it is how smoltalk is
-allowed to infer the provider from the model name. If your `agency.json` sets
-`defaultProvider` to route through a proxy, a bare `--model` bypasses that
-proxy; name it (`--model litellm/gpt-4o-mini`) to keep it. The mapping deletes
-the key, rather than setting it to `undefined`, because the generator checks for
-presence.
+**A bare model drops the provider rather than leaving it.** The code generator
+emits a provider only when `client.defaultProvider` is truthy, so clearing it is
+how smoltalk is allowed to infer the provider from the model name. If your
+`agency.json` sets `defaultProvider` to route through a proxy, a bare `--model`
+bypasses that proxy; name it (`--model litellm/gpt-4o-mini`) to keep it.
+
+The mapping removes the key rather than setting it to `undefined`. Either would
+satisfy the generator; removing it keeps the resolved config free of dangling
+fields, which is what `agency config show` and any other consumer sees. The test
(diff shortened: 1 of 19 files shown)


packages/agency-lang/docs/dev/config.md
```diff
@@ -26,6 +26,47 @@ Where applied: sources 1⊕2 at the CLI (baked into the generated program);
...
+
+Three things about this are easy to get wrong later.
+
+**A bare model deletes the provider rather than leaving it.** The code generator
+emits a provider only when one is configured, so deleting it is how smoltalk is
+allowed to infer the provider from the model name. If your `agency.json` sets
+`defaultProvider` to route through a proxy, a bare `--model` bypasses that
+proxy; name it (`--model litellm/gpt-4o-mini`) to keep it. The mapping deletes
+the key, rather than setting it to `undefined`, because the generator checks for
+presence.
```
egonSchiele: This rationale does not match the code generator: `typescriptBuilder.ts` uses `if (cfg.client?.defaultProvider)`, so it checks truthiness, not key presence; `{ defaultProvider: undefined }` would also omit the provider. Deleting the inherited key is still a clear mapping contract, but please describe that as the reason rather than claiming codegen requires deletion. The same presence claim appears in `config.test.ts` and the PR description.

packages/agency-lang/lib/runtime/agencyLlm.test.ts
```diff
@@ -253,3 +256,72 @@ describe("agency.llm — frame requirement", () => {
...
+    const client = new RecordingClient(["ok"]);
+    ctx.setLLMClient(client);
+    const threads = ThreadStore.withDefaultActive(ctx.statelogClient);
+    await inFrame(ctx, threads, () => call(threads));
+    const config = client.configs[0];
+    if (config === undefined) {
+      throw new Error("the prompt never reached the recording client");
+    }
+    return {
+      model: (config as any).model,
```
egonSchiele: `config` is already a `PromptConfig`, and that type declares both `model` and `provider`. Could these casts be removed and the fields read directly? Keeping `as any` here weakens the exact type boundary this test is meant to pin: `return { model: config.model, provider: config.provider };`

egonSchiele (commented):
Reviewed the final implementation against the revised design and plan. I found no functional blockers: the parser/config boundary is coherent, the Commander adapter handles repeated values, provider clearing/stickiness matches the field-wise runtime merge, and the runtime plus packed-CLI tests exercise the important seams. CI is green across lint, typecheck, unit, integration, agent, and Agency tests.

I left two non-blocking inline comments:

1. Remove unnecessary `as any` casts from the final-client precedence test so the test preserves the `PromptConfig` type boundary it is intended to pin.
2. Correct the explanation that codegen checks provider-key presence; it actually checks truthiness. Deleting the inherited key remains a valid explicit mapping contract, but codegen does not require deletion instead of `undefined`.

packages/agency-lang/lib/config.ts
```diff
@@ -678,6 +678,20 @@ export function findProjectRoot(startPath: string): string | null {
 /** Per-invocation flags accepted by `agency run`/`compile` and forwarded to the
  *  bundled agents. Mapped onto AgencyConfig by applyCliFlags. `trace` is
  *  `string` for `--trace <file>`, `true` for a bare `--trace`. */
```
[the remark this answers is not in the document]
egonSchiele: Fixed. Worse than dangling, actually: my insertion landed between the pre-existing `CliFlags` doc comment and `CliFlags` itself, so that comment was orphaned too. `ResolvedModelFlag` now sits above it, and the `CliFlags` doc is back where it belongs.

packages/agency-lang/docs/dev/config.md
egonSchiele: You are right, and the claim was mine rather than the code's. `typescriptBuilder.ts:4424` is `if (cfg.client?.defaultProvider)` — truthiness, so `{ defaultProvider: undefined }` would omit the provider just as well.

Corrected in all three places. The mapping still removes the key, but the stated reason is now what it actually is: it keeps the resolved config free of dangling fields, which is what `agency config show` and other consumers read. The test keeps asserting absence so that contract cannot drift, and the comment now says it pins a contract rather than satisfying a generator requirement.

I will fix the PR description too.

packages/agency-lang/lib/runtime/agencyLlm.test.ts
egonSchiele: Removed. `PromptConfig` declares `model?: string` and `provider?: string` at `lib/runtime/llmClient.ts:44-45`, so the casts were doing nothing except hiding the boundary this test exists to pin. `tsc --noEmit` passes without them, which is the confirmation.

PR #813, opened 2026-08-06
test(integration): per-PR tarball tests for the deterministic language CLI commands

egonSchiele:
## What

Adds a fast, offline, per-PR integration runner for the deterministic language CLI commands, and normalizes the `diagnostics` command's output so its result can be asserted.

## Why

The deep CLI integration suite (`cli-main/test.mjs`) only runs on push to `main`, so a packaging regression in the everyday language commands (`fmt`, `typecheck`, `ast`/`parse`, `doc`) was invisible until after merge. And `lint`, `explain`, `config show`, `preprocess`, and `diagnostics` had no tarball coverage anywhere.

An integration test here catches what the many `lib/cli/*.test.ts` unit tests cannot: a missing `dist/` file, a broken bin entry, output sent to the wrong stream, or a wrong exit code — i.e. "works in-process, broken once published."

## What the new runner checks

`tests/integration/cli-lang/test.mjs` installs the npm tarball and runs the canonical commands: `typecheck`, `format`, `lint`, `explain`, `ast`, `preprocess`, `doc`, `config show`, `diagnostics`. Every check asserts the exact exit status and keeps stdout/stderr separate, and each uses an independent oracle so a stable-but-wrong implementation cannot pass:

- `format` compares against an independent expected value, and separates stdout mode from `-i` mode, so a no-op or a formatter that corrupts source both fail.
- `preprocess` asserts a `scope: "args"` annotation that the raw parser never emits, so `preprocess` cannot be silently wired to `ast`.
- `doc` asserts the generated markers **and** that the function body was not copied in.
- `diagnostics` derives the expected error position from the fixture rather than hard-coding it.

It complements `cli-main` (which owns deep syntax matrices, stdin/directory inputs, and strict mode) rather than duplicating it, and uses canonical command names where `cli-main` mostly uses aliases.

## Supporting changes

- **`helpers.mjs`**: one shared `runInstalledAgency` boundary — the only layer that knows about `spawnSync`, `npx --no-install` (no registry fallback), exact-status enforcement, and separate streams.
- **`cli-main/test.mjs`**: routed through the shared boundary; the boolean `expectFail` became an exact `expectedStatus` (all failing cases are exit 1, verified).
- **`scripts/agency.ts`**: `diagnostics` now routes through `parseAgency().errorData`, giving normalized, zero-indexed error data for both committed and recoverable parse failures instead of raw `TarsecError.data`.
- **`test.yml`**: the new runner runs in the existing integration job (already on PRs and pushes) — no new job, secret, or matrix.

## Validation (local, fresh tarball)

- New runner: all 9 checks pass.
- Migrated `cli-main`: passes.
- Deliberately corrupting an expectation makes the runner fail (proves the checks can fail).
- `pnpm run typecheck` and `pnpm run lint:structure`: clean.

## Note

Two intentional deviations from the literal plan: the runner strips ANSI before substring assertions (the CLI emits color even when piped), and the `preprocess` scope check uses a tree walk rather than index-based navigation (robust to AST shape changes). Same intent, same coverage.

### DIFF
--- a/packages/agency-lang/scripts/agency.ts
+++ b/packages/agency-lang/scripts/agency.ts
@@ -65,7 +65,7 @@ import {
   redactConfigSecrets,
 } from "@/config.js";
 import * as path from "path";
-import { _parseAgency } from "@/parser.js";
+import { parseAgency } from "@/parser.js";
 import { TypescriptPreprocessor } from "@/preprocessors/typescriptPreprocessor.js";
 import { buildCompilationUnit } from "@/compilationUnit.js";
 import { expandSplices } from "@/preprocessors/expandSplices.js";
@@ -75,7 +75,6 @@ import { formatErrors, formatDiagnosticsHint, typeCheck } from "@/typeChecker/in
 import { Command, InvalidArgumentError } from "commander";
 import * as fs from "fs";
 import { color } from "@/utils/termcolors.js";
-import { TarsecError } from "tarsec";
 import process from "process";
 import { agent } from "@/cli/agent.js";
 import { mcpAdd, mcpRemove, mcpList, type McpAddOptions } from "@/cli/mcp.js";
@@ -1181,15 +1180,16 @@ export function createProgram(deps: CliDependencies = {}): Command {
     .description("Run diagnostics for VSCode")
     .argument("[inputs...]", "Paths to .agency input files")
     .action(async (inputs: string[]) => {
+      // Route through parseAgency so the payload is the normalized
+      // ParseAgencyErrorData (zero-indexed user-source coordinates), covering
+      // both committed and recoverable parse failures. applyTemplate:false keeps
+      // coordinates in the exact source the editor supplied; lower:false keeps
+      // this to syntax diagnostics.
+      const config = getConfig();
       await forEachSource(inputs, (contents) => {
-        try {
-          _parseAgency(contents);
-        } catch (error) {
-          if (error instanceof TarsecError) {
-            console.log(JSON.stringify(error.data, null, 2));
-          } else {
-            throw error;
-          }
+        const result = parseAgency(contents, config, false, false);
+        if (!result.success && result.errorData) {
+          console.log(JSON.stringify(result.errorData, null, 2));
         }
       });
     });
--- a/packages/agency-lang/scripts/agency.ts
+++ b/packages/agency-lang/scripts/agency.ts
@@ -1191,14 +1191,16 @@ export function createProgram(deps: CliDependencies = {}): Command {
         const result = parseAgency(contents, {}, false, false);
         if (result.success) return;
         // Always emit a payload for a failure. A rare failure path returns no
-        // errorData; fall back to a minimal one built from the message so the
-        // editor still gets JSON.
+        // errorData; fall back to a minimal one. `result.message` is optional,
+        // so normalize it to a non-empty string for both required message
+        // fields of ParseAgencyErrorData.
+        const message = result.message ?? "Parse error";
         const errorData = result.errorData ?? {
           line: 0,
           column: 0,
           length: 1,
-          message: result.message,
-          prettyMessage: result.message,
+          message,
+          prettyMessage: message,
         };
         console.log(JSON.stringify(errorData, null, 2));
       });
--- a/packages/agency-lang/scripts/agency.ts
+++ b/packages/agency-lang/scripts/agency.ts
@@ -1182,15 +1182,25 @@ export function createProgram(deps: CliDependencies = {}): Command {
     .action(async (inputs: string[]) => {
       // Route through parseAgency so the payload is the normalized
       // ParseAgencyErrorData (zero-indexed user-source coordinates), covering
-      // both committed and recoverable parse failures. applyTemplate:false keeps
+      // both committed and recoverable parse failures. Parsing does not depend
+      // on config, so pass {} rather than getConfig(): a broken agency.json
+      // must not take editor diagnostics down with it. applyTemplate:false keeps
       // coordinates in the exact source the editor supplied; lower:false keeps
       // this to syntax diagnostics.
-      const config = getConfig();
       await forEachSource(inputs, (contents) => {
-        const result = parseAgency(contents, config, false, false);
-        if (!result.success && result.errorData) {
-          console.log(JSON.stringify(result.errorData, null, 2));
-        }
+        const result = parseAgency(contents, {}, false, false);
+        if (result.success) return;
+        // Always emit a payload for a failure. A rare failure path returns no
+        // errorData; fall back to a minimal one built from the message so the
+        // editor still gets JSON.
+        const errorData = result.errorData ?? {
+          line: 0,
+          column: 0,
+          length: 1,
+          message: result.message,
+          prettyMessage: result.message,
+        };
+        console.log(JSON.stringify(errorData, null, 2));
       });
     });
 
--- a/packages/agency-lang/tests/integration/cli-lang/test.mjs
+++ b/packages/agency-lang/tests/integration/cli-lang/test.mjs
@@ -91,9 +91,13 @@ node preprocessProbe(): string {
 }
 `;
 
+// The body carries a unique sentinel found nowhere in the signature or
+// docstring. `doc` documents the public API, not the body, so the sentinel must
+// be absent from the generated Markdown — a check that survives any body
+// reformatting a broken generator might apply.
 const docProbe = `export def greet(name: string): string {
   """Greet a person by name."""
-  return "Hello, " + name
+  return "Hello, " + name + " DOC_BODY_SENTINEL"
 }
 `;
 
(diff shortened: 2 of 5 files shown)


packages/agency-lang/tests/integration/cli-lang/test.mjs
```diff
@@ -0,0 +1,329 @@
...
+}
+`;
+
+const docProbe = `export def greet(name: string): string {
+  """Greet a person by name."""
+  return "Hello, " + name
+}
+`;
+
+const diagnosticsBad = `def brokenParams(first: string; second: string) {}
```
egonSchiele, line 104: This fixture exercises only the thrown `TarsecError` branch; `parser.test.ts` explicitly uses the semicolon-in-parameters case as its committed-error probe. Please add a recoverable failure such as `x = 5\n!!!` and assert its CLI JSON position/message too. Otherwise `diagnostics` can regress to the old `_parseAgency` behavior—which emitted nothing for recoverable failures—while this new tarball test still passes, despite this PR specifically adding support for that branch. A multiline recoverable fixture also proves line mapping rather than only column mapping on line 0.

packages/agency-lang/tests/integration/cli-main/test.mjs
```diff
@@ -109,8 +110,21 @@ function runLogged(label, file, args = [], opts = {}) {
...
 function runAgency(label, args, opts = {}) {
-  return runLogged(label, "npx", ["--no-install", "agency", ...args], opts);
+  const result = runInstalledAgency(dir, args, {
+    expectedStatus: opts.expectedStatus ?? 0,
+    input: opts.input,
+    env: opts.env,
+    timeout: opts.timeout,
+  });
+  const output = `${result.stdout}${result.stderr}`;
+  writeFileSync(join(logsDir, `${label}.txt`), output);
```
egonSchiele: `runInstalledAgency` throws before returning on an unexpected status, timeout, or spawn error, so this log write is skipped in exactly the failure cases where the per-command log is most useful. The previous `runLogged` path wrote `<label>.txt` before checking status, and the outer catch still says command logs were preserved. Please attach `{stdout, stderr, status, signal}` to the helper error and catch/log/rethrow here (preserving the spawn error as `cause`), or otherwise arrange for logging before status enforcement.

packages/agency-lang/tests/integration/cli-lang/test.mjs
```diff
@@ -0,0 +1,329 @@
...
+  const docPath = join(dir, "generated-docs", "doc-probe.md");
+  assert(existsSync(docPath), `[doc] expected generated file at ${docPath}`);
+  const doc = readFileSync(docPath, "utf8");
+  assertIncludes(doc, 'name: "doc-probe"');
+  assertIncludes(doc, "# doc-probe");
+  assertIncludes(doc, "## Functions");
+  assertIncludes(doc, "### greet");
+  assertIncludes(doc, "```ts\ngreet(name: string): string\n```");
+  assertIncludes(doc, "Greet a person by name.");
+  assert(!doc.includes('return "Hello, " + name'), "[doc] must not copy the function body into the docs");
```
egonSchiele: This negative oracle can false-pass if a broken doc generator copies the body after normalizing whitespace, adding a semicolon, or rendering only the expression. Please put a unique sentinel such as `DOC_BODY_SENTINEL` in the implementation body and assert that the sentinel is absent from the generated Markdown. That tests body exclusion independently of source formatting.

egonSchiele (commented):
The per-PR runner and shared process boundary are a solid improvement. Three remaining issues are called out inline: the newly supported recoverable diagnostics branch is not covered, cli-main loses per-command logs on unexpected failures, and the doc body-exclusion oracle can false-pass after formatting.

packages/agency-lang/scripts/agency.ts
```diff
@@ -1181,15 +1180,16 @@ export function createProgram(deps: CliDependencies = {}): Command {
...
-          _parseAgency(contents);
-        } catch (error) {
-          if (error instanceof TarsecError) {
-            console.log(JSON.stringify(error.data, null, 2));
-          } else {
-            throw error;
-          }
+        const result = parseAgency(contents, config, false, false);
+        if (!result.success && result.errorData) {
+          console.log(JSON.stringify(result.errorData, null, 2));
```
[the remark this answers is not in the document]
egonSchiele: Fixed in fd9d82a76. Parsing does not use config (verified: config is only threaded to _parseAgency, and the old diagnostics called it with none), so it now parses with {} and never calls getConfig() — a broken agency.json can no longer take diagnostics down. It also always emits a JSON payload for a failure, falling back to a minimal one built from result.message when errorData is absent.

packages/agency-lang/tests/integration/cli-lang/test.mjs
egonSchiele: Good catch — that was the whole point of the change and the test missed it. Added a recoverable multiline fixture (`const x = 5\n!!!`) in fd9d82a76. Verified it takes the non-committed path and emits JSON with line 1; the test derives the expected line from the fixture and asserts prettyMessage contains `Line 2`, so it proves line mapping, not just column-0. The old _parseAgency path emitted nothing here, so this now guards the new branch.

packages/agency-lang/tests/integration/cli-main/test.mjs
egonSchiele: Fixed in fd9d82a76. runInstalledAgency now attaches {stdout, stderr, status, signal} to the errors it throws (spawn error preserved as cause), and cli-main's runAgency wraps the call in try/catch, writing <label>.txt from the attached streams before rethrowing. So the per-command log survives the status/timeout/spawn-error cases, matching the old runLogged behavior the outer catch message promises.

packages/agency-lang/tests/integration/cli-lang/test.mjs
egonSchiele: Fixed in fd9d82a76. The doc-probe body now contains a unique `DOC_BODY_SENTINEL` (absent from the signature and docstring), and the test asserts that sentinel does not appear in the generated Markdown. Body exclusion is now checked independently of any whitespace/formatting the generator might apply.

packages/agency-lang/tests/integration/cli-lang/test.mjs
```diff
@@ -0,0 +1,361 @@
...
+  writeFile(dir, "diagnostics-good.agency", diagnosticsGood);
+  writeFile(dir, "diagnostics-recoverable.agency", diagnosticsRecoverable);
+  writeFile(dir, "config-probe.json", configProbe);
+
+  for (const testCase of commandCases) runCommandCase(testCase);
+  checkFormat();
+  checkAstPreprocess();
+  checkDoc();
+  checkConfig();
+  checkDiagnostics();
```
egonSchiele, line 360: The implementation now deliberately avoids `getConfig()` so a broken project config cannot take editor diagnostics down, but this test project has no `agency.json`. Reverting the command to `getConfig()` would therefore still pass all diagnostics checks. Please write a malformed `agency.json` immediately before `checkDiagnostics()` (after the other config-sensitive checks) and keep the committed/recoverable assertions unchanged. That directly guards the config-independence behavior added in the review fix.

packages/agency-lang/scripts/agency.ts
```diff
@@ -1181,16 +1180,27 @@ export function createProgram(deps: CliDependencies = {}): Command {
...
+        if (result.success) return;
+        // Always emit a payload for a failure. A rare failure path returns no
+        // errorData; fall back to a minimal one built from the message so the
+        // editor still gets JSON.
+        const errorData = result.errorData ?? {
+          line: 0,
+          column: 0,
+          length: 1,
+          message: result.message,
+          prettyMessage: result.message,
```
egonSchiele: `ParseAgencyResult.message` is optional, so this fallback can serialize without either `message` field when `result.message` is undefined, contradicting the comment that it always supplies a useful minimal payload and the required `ParseAgencyErrorData` shape. Please normalize once with `const message = result.message ?? "Parse error"` and use that for both fields.

egonSchiele (commented):
The previous review findings are addressed cleanly: recoverable failures now have coverage, failure logs are preserved, and the doc sentinel is robust. Two remaining diagnostics gaps are noted inline.

packages/agency-lang/tests/integration/cli-lang/test.mjs
egonSchiele: Fixed in dab925982. The runner now writes a malformed agency.json immediately before checkDiagnostics() (after the config-sensitive checks, and after checkConfig which uses an explicit -c file). Verified: with that malformed config present, `config show` exits 1 but `diagnostics` stays exit 0 and still emits JSON — so a getConfig() revert would now fail the diagnostics status-0 checks. Committed/recoverable assertions are unchanged.

packages/agency-lang/scripts/agency.ts
egonSchiele: Fixed in dab925982. Normalized once with `const message = result.message ?? "Parse error"` and used it for both message and prettyMessage, so the fallback always satisfies the ParseAgencyErrorData shape even when result.message is undefined.

PR #814, opened 2026-08-06
Incremental per-page cache for agency doc

egonSchiele:
Implements the twice-revised spec `2026-08-04-incremental-doc-cache-v2-spec.md` on top of #810: `agency doc <dir>` now skips parsing and rendering for pages whose inputs are unchanged, owns its output directory (repairs deleted and hand-edited pages, reconciles orphans across invocation changes), and produces byte-identical output to a cold run.

## Measured (real stdlib, this branch)

| scenario | before | after |
|---|---|---|
| warm `agency doc stdlib` | ~5.3s, all 83 pages | **0.44s, zero pages** |
| output vs pre-refactor baseline | — | byte-identical (full-tree diff) |

Also verified end-to-end: one-file edit re-renders in **0.51s**; warm `make doc` (no stamp anymore) costs **0.64s** and now actually repairs deleted or hand-edited pages; the staged `stdlib/docs/stdlib` copy contains no sidecars; and regenerating `docs/site/stdlib` through the new target left git clean (byte parity with the committed pages).

## Design (docs/dev/doc-cache.md has the full story)

- **One ledger + one lock per physical output directory**, stored as sidecars inside it (`.agency-doc.json` / `.agency-doc.lock`, realpath'd). Input dir + ignore list are invocation identity *inside* the ledger: changing them is a transition that reconciles, not a fresh start that forgets.
- **Freshness ≠ ownership.** Identity/render-key changes mark everything stale but retain prior entries — the deletion evidence. Only corruption (bad JSON, wrong version, foreign outputDir, any malformed field — validation is recursive) forfeits deletion authority; then the run re-renders everything and deletes nothing. Conservative contract: unrecorded files are never deleted.
- **Per-page freshness** from #810's `dependencyFingerprint` (std::-resolved deps incl. the template's prelude edge, subtree pkg:: detection, splice/unparseable ⇒ never cached), the shared stdlib-hash flavor rule, a structured render key (config + effective `--base-url`), and a per-entry `outputHash` that repairs hand-edited or deleted pages.
- **`registrySymbols` + `linkTargets`**: pass 1 rebuilds the symbol registry from cached contributions (no parse for fresh files — parsing was ~80% of a run), in traversal order so collisions stay last-writer-wins; every registry lookup a page rendered is recorded and re-checked each run, so cross-closure link changes (symbol moved, appeared, or new collision winner) can never be served stale.
- **Deletion boundary**: reconciliation recomputes the deterministic path from the validated key (`.agency`-suffix rule; the stored `outputPath` is never dereferenced) and routes through `resolveOwnedOutputPath`, which lstat-walks components and refuses symlinked ancestors — `out/sub → victim` cannot be written or deleted through.
- **Lock**: O_EXCL with a `pid:uuid` token; both modes lock (single-file mode keeps zero cache behavior but writes into the same directory); stale locks are removed manually — auto-break races two processes past one dead lock.
- **Makefile**: `make doc` is a plain invocation; the outer stamp (which skipped the command entirely, so deleted/hand-edited pages were never repaired) and `scripts/stdlib-stamp.ts` are gone; `stage-stdlib-docs` strips sidecars.
- Rendering itself is untouched: stale pages go through today's per-page path (own `DocContext`, own symbol table); the misleading "one symbol table per run" comment was corrected — the code always built per-page tables.

## Tests

- `docLedger.test.ts` (42): key/path safety incl. the `.agency`-suffix deletion rule and normalized-alias rejections, `resolveOwnedOutputPath` symlink refusals, a per-field malformed-ledger authority matrix (raw-JSON corruption; save validates separately), token-lock semantics (same-pid successor protection, foreign takeover, non-EEXIST propagation), and the entry-builder/freshness writer-checker pair with per-field perturbations.
- `docCache.test.ts` (30): a single-invocation mutation harness (backdated mtimes → one incremental run → exact rewrite set → cold-run byte parity) for every staleness trigger; semantic oracles (Throws effects present/absent, exact link targets) rather than parity alone; ownership transitions over one output directory (root→nested, ignore-list changes, deletion+render-key-change in one run); symlink-victim and hostile-outputPath regressions with real deletion candidates; lock zero-mutation proofs in both modes.
- `docStdlibDeps.test.ts` (4): a doc-faithful fake stdlib (real index/array carve-out, stubbed prelude surface) proving the recorded `std::` and prelude edges are load-bearing, with cold parity.
- `doc.test.ts` (+4): registry-extraction boundary pins (unexported aliases in, unexported constants out) and link recording.

## Notes

- One test-design correction vs the plan: the corrupted-ledger scenario originally expected the *next* (valid) run to reclaim the orphaned page; that contradicts the conservative contract — a fresh ledger never owned it, so it survives permanently (documented escape: delete the output directory). The test now pins the contract as specced.
- Spec/plan (untracked, main checkout): `packages/agency-lang/2026-08-04-incremental-doc-cache-v2-{spec,plan}.md`.


egonSchiele (commented):
I found several correctness/safety issues that should be addressed before merging. The largest integration issue is that the package Makefile was not updated with the implementation: a clean `make doc` still invokes the deleted stamp script, while a dirty tree can hide that failure behind an orphaned compiled file and then erase the new ledger before every run. Focused tests themselves are green (122/122), but they do not cover these gaps.

packages/agency-lang/scripts/stdlib-stamp.ts
```diff
@@ -1,36 +0,0 @@
-// Prints the doc stamp: stdlib content hash COMBINED with the compiler
```
egonSchiele, line 1: This source is deleted, but `packages/agency-lang/makefile:123` still executes `dist/scripts/stdlib-stamp.js`. A clean build will not contain that script, so `make doc` fails; an incremental tree can retain the orphaned JS and mask the failure. The unchanged recipe also still `rm -rf`s `docs/site/stdlib` before generation, which defeats the cache/ownership ledger and can remove a live lock, and `stage-stdlib-docs` copies `.agency-doc*` into the staged package. Please apply the planned Makefile transition: invoke `agency doc` directly, remove the pre-run output deletion (including the analogous `packages-doc` deletion if those pages should be incremental), and strip `stdlib/docs/stdlib/.agency-doc*` after staging.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -50,80 +69,226 @@ export function generateDoc(
...
+  // output directory, so it must not interleave with a directory run.
+  fs.mkdirSync(outputDir, { recursive: true });
+  const outDirReal = fs.realpathSync(outputDir);
+  const lock = acquireDocLock(outDirReal);
+  try {
+    if (fs.statSync(inputPath).isDirectory()) {
+      generateDocDirectory(config, inputPath, outDirReal, ignoreDirs, baseUrl);
+    } else {
+      const baseName = path.basename(inputPath).replace(/\.agency$/, ".md");
+      const outputPath = path.join(outDirReal, baseName);
```
egonSchiele: The single-file branch bypasses the new owned-output boundary. If `out/a.md` is a symlink to an external victim, `generateDocForFile` eventually calls `writeFileSync` on this raw joined path and overwrites the victim. Directory mode correctly calls `resolveOwnedOutputPath` and rejects `leafIsSymlink`; please route this branch through the same helper and add the missing single-file symlink-victim regression.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -50,80 +69,226 @@ export function generateDoc(
...
-      { program, relativePath, mdRelPath },
-    ] of parsedPrograms) {
-      const outputPath = path.join(outputDir, mdRelPath);
-      fs.mkdirSync(path.dirname(outputPath), { recursive: true });
+  // BOTH branches lock: single-file mode bypasses the cache (no ledger,
+  // no freshness, no reconciliation) but writes into the same physical
+  // output directory, so it must not interleave with a directory run.
+  fs.mkdirSync(outputDir, { recursive: true });
+  const outDirReal = fs.realpathSync(outputDir);
+  const lock = acquireDocLock(outDirReal);
```
egonSchiele, line 97: Acquiring the lock before parsing creates a permanent stale lock on an ordinary syntax error. The shared `parse()` helper calls `process.exit(1)` on failure (`commands.ts:218-224`), and `process.exit` does not unwind this `try`/`finally`, so `releaseDocLock` never runs. After the user fixes the source, every run is refused until the lock is removed manually. Please use a throwing/non-exiting parse path inside `generateDoc` (convert it to an exit status only after unwinding), with a subprocess regression using genuinely invalid Agency syntax.

packages/agency-lang/lib/cli/docLedger.ts
```diff
@@ -0,0 +1,435 @@
...
+  sourceRel: string; // pre-validated by the caller's traversal
+  ctx: DocFreshnessContext;
+  config: AgencyConfig;
+  registrySymbols: string[];
+  linkTargets: Record<string, string | null>;
+  writtenBytes: string;
+}): DocLedgerEntry {
+  const { sourceRel, ctx, config, registrySymbols, linkTargets, writtenBytes } = args;
+  const absSource = path.join(ctx.inputDir, sourceRel);
+  const fp = dependencyFingerprint(absSource, config, {
```
egonSchiele, line 452: This fingerprints and hashes the files only after the page was parsed and rendered. If an editor saves the source (or a dependency) between pass 1/rendering and this call, the Markdown/`registrySymbols` describe the old bytes while the ledger records hashes and possibly edges for the new bytes. The next run then accepts the stale page as fresh. Please snapshot/recheck the root and dependency inputs across rendering and force the entry `cacheable:false` when they drift, so the next run repairs it.

packages/agency-lang/lib/cli/docLedger.ts
```diff
@@ -0,0 +1,435 @@
...
+ *  produce a ledger that load would strip of authority. */
+export function saveDocLedger(outDirReal: string, ledger: DocLedger): void {
+  for (const key of Object.keys(ledger.entries)) {
+    if (!isSafeSourceRel(key)) {
+      throw new Error(`refusing to save unsafe doc ledger key: '${key}'`);
+    }
+  }
+  const file = path.join(outDirReal, DOC_LEDGER_NAME);
+  const tmpFile = `${file}.${process.pid}.tmp`;
+  fs.writeFileSync(tmpFile, JSON.stringify(ledger, null, 2));
```
egonSchiele: The predictable temp path is opened with the default truncating `w` mode. A pre-existing `.agency-doc.json.<pid>.tmp` symlink is therefore followed and its external target is overwritten before `renameSync` moves the symlink. Please use an unpredictable UUID temp name and exclusive creation (`flag: "wx"`), then clean up that exact temp in `finally`; add a victim-content regression for a pre-existing temp symlink.

packages/agency-lang/lib/cli/docLedger.ts
```diff
@@ -0,0 +1,435 @@
...
+  } catch {
+    /* absent leaf */
+  }
+  return { abs, leafIsSymlink };
+}
+
+/** Structured, not concatenated: a string boundary could alias two
+ *  distinct (config, baseUrl) states. */
+export function docRenderKey(config: AgencyConfig, effectiveBaseUrl: string): string {
+  return hashBytes(
```
egonSchiele, line 175: The rendered Markdown also depends on `process.env.AGENCY_DEBUG`: `AgencyGenerator` captures it and emits trace wrappers into generated code blocks. Since the render key includes only config and base URL, rerunning a warm cache under a different `AGENCY_DEBUG` value serves pages that differ from a cold run in that same environment. Prefer making doc rendering explicitly debug-independent; otherwise include the effective debug setting in this key and test the environment transition.

egonSchiele:
All findings addressed in the latest commit.

**The makefile (the big one):** an embarrassing case-sensitivity miss — the file is tracked as `makefile` (lowercase), and my `git add Makefile` on this case-insensitive filesystem silently staged nothing, which is why my local verification passed against the edited working tree while the pushed diff had the old recipe. The transition is now actually committed and verified from the committed content: plain `agency doc` invocation, no stamp script, no pre-run `rm -rf`, sidecar strip in `stage-stdlib-docs`. Per your note, `packages-doc` also stops wiping its per-package output dirs (each gets its own ledger now), with a `.gitignore` entry for those sidecars.

**Lock leak on parse errors:** doc now parses through a throwing path (`DocParseError`) instead of `commands.parse`'s `process.exit`, releases the lock explicitly, and only then converts to exit(1). Subprocess regression: invalid syntax → non-zero exit, no lock left, and the rerun after fixing the file succeeds.

**Single-file symlink bypass:** that branch now routes through `resolveOwnedOutputPath` and refuses a leaf symlink; victim-content regression added.

**Mid-render TOCTOU:** pass 1 snapshots the source hash at parse time; `buildDocLedgerEntry` compares it against the post-render hash and also re-reads source+deps once after the fingerprint walk — any drift forces `cacheable: false` so the next run repairs rather than accepts. Unit test via an injected stale snapshot.

**Temp-file symlink hijack:** the save temp is now UUID-named with `flag: "wx"` and cleaned in `finally`; regression pre-plants a symlink at the old predictable `.<pid>.tmp` name and asserts the victim is untouched.

**AGENCY_DEBUG:** took your preferred option — doc rendering is now debug-independent. `AgencyGenerator` gained a `debug` override; doc's shared generator and all three `generateAgency` code-block sites pass `debug: false` (interestingly, the code-block path was the one that actually leaked trace markers into pages). Subprocess env-transition regression: warm cache under `AGENCY_DEBUG=1` equals a cold debug run, and neither contains trace markers.

**Copilot's two:** containment in `resolveOwnedOutputPath` is now `path.relative`-based (no `//` root artifact), and reconciliation `lstat`s the leaf and skips anything that isn't a file or symlink (directory-occupant regression added).

128 doc-suite tests green (was 122; +6 regressions from this round), typecheck/lint/guards clean.

packages/agency-lang/lib/cli/docLedger.ts
egonSchiele: The source-side snapshot is fixed, but the dependency-side race is still open. `generateDocForFile`/`SymbolTable.build` reads dependency semantics before this function runs; both the initial fingerprint/hash pass here and the added re-read at lines 443–452 happen afterward. If `b.agency` changes after the symbol table reads its old effect but before line 425, both post-render passes see the new bytes, `cacheable` remains true, and the ledger records the new dep hash against Markdown rendered from the old semantics. Please capture the complete dependency snapshot (closure + hashes/flags) before `generateDocForFile`, recompute after rendering, and mark the entry uncacheable if either membership or bytes drift. The regression should mutate an imported effect between symbol-table construction and entry building, then verify the caller is uncacheable and the next run reaches cold parity.

egonSchiele (commented):
Re-reviewed the round-1 fixes at 145321d5. The Makefile, single-file symlink boundary, parse-error lock release, ledger temp creation, debug-independent rendering, containment, and reconciliation fixes are sound. One part of the mid-render TOCTOU finding remains unresolved: dependency state is still first captured only after rendering, so a dependency save between SymbolTable consumption and entry building can pair stale Markdown with fresh dependency hashes. I replied on the existing discussion with the precise remaining race and suggested regression.

Verification from a fresh build: 128 focused doc tests passed across 5 files; current CI is green.

egonSchiele:
Dependency-side race closed in 2e0d58b, exactly as you framed it: the snapshot now happens on the correct side of rendering.

`captureDepSnapshot` (closure membership + per-dep content hashes) runs immediately before `generateDocForFile` — the last instant the hashes are guaranteed to describe what `SymbolTable.build` will consume. `buildDocLedgerEntry` compares that snapshot against its post-render fingerprint: any membership or byte drift forces `cacheable: false`, so the entry can never pair stale Markdown with fresh hashes, and the next run repairs it. (This subsumes the previous post-hoc re-read, which is gone.)

The window analysis, for the record: a dep save before the snapshot → snapshot and render both see new bytes, consistent; between snapshot and the symbol-table read, or between the read and the builder → pre/post hashes differ → uncacheable; after the builder → the recorded hashes no longer match disk, so the ordinary freshness check catches it next run. Content hashing also makes this immune to the parse cache's same-size/same-granule mtime blind spot.

Regressions per your suggestion: a unit test at the builder seam (snapshot of old dep bytes, dep rewritten, builder → uncacheable), and an end-to-end injection (`docTOCTOU.test.ts`) that rewrites an imported effect right after the caller's snapshot fires — mid-render from the cache's perspective — then asserts the caller's entry is `cacheable: false`, the next run reaches byte parity with a cold run over the final state, and the repaired entry is cacheable again.

128 doc-suite tests + the 2 CLI subprocess tests green; typecheck clean.

packages/agency-lang/lib/cli/docTOCTOU.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  mutate: null as (() => void) | null,
+}));
+vi.mock("./docLedger.js", async (importOriginal) => {
+  const real = await importOriginal<typeof import("./docLedger.js")>();
+  return {
+    ...real,
+    captureDepSnapshot: (absSource: string, config: unknown, stdlibDir: string) => {
+      const snapshot = real.captureDepSnapshot(absSource, config as never, stdlibDir);
+      if (inject.mutate && path.basename(absSource) === inject.afterSnapshotOf) {
+        const m = inject.mutate;
```
egonSchiele, line 23: The production fix is sound, but this injection does not reproduce the original failure window: it mutates immediately after the snapshot and explicitly evicts `b.agency`, so the subsequent `SymbolTable.build` renders the **new** effect. The stale-cache bug was specifically “symbol table rendered old semantics, then the dep changed before entry building recorded new hashes.” This test would still pass if the snapshot accidentally moved back to after rendering. Could the seam fire immediately after `SymbolTable.build` returns for `a.agency` and before `buildDocLedgerEntry`, then assert the first-run `a.md` contains `race::old`, the entry is uncacheable, and the next run changes it to `race::new` with cold parity? That would make the regression fail for the original implementation error, not just a neighboring race window.

egonSchiele (commented):
Re-reviewed the dependency-race fix at 2e0d58b. The production change now captures closure membership and content hashes before rendering and compares them with post-render state, so the remaining correctness finding is resolved. I found no new production blocker. I left one non-blocking test-validity comment: the end-to-end injection currently mutates before SymbolTable consumes the dependency, rather than reproducing the original stale-render/fresh-hash window.

Fresh verification: build succeeded and 130 focused doc tests passed across 6 files. CI is still in progress, with completed checks green so far.

PR #816, opened 2026-08-06
test(integration): per-PR artifact-command tests (pack standalone, trace/bundle, coverage)

egonSchiele:
## What

Tier 2 CLI integration tests, **PR 1 of 3** (artifact workflows). Adds `tests/integration/cli-tier2/test.mjs`, a per-PR tarball runner for the artifact commands, and does the shared-helper groundwork the later PRs build on.

## Why

`pack`, `trace`, `bundle`/`unbundle`, and `coverage` are only integration-tested on push to `main` (`cli-main`), so a packaging regression in any of them is invisible until after merge. (`pack` has a per-PR test, but it runs the output beside the project's `node_modules`, so it can't catch a non-self-contained bundle.)

## What it checks

- **`pack` — standalone isolation.** Packs an import-bearing program (local module + a `std::math` call + runtime), copies **only** the packed file to a root with no `node_modules` ancestor, renames the installed `node_modules`, and runs it there — asserting the exact output. This is the contract the existing per-PR pack test can't reach.
- **`trace` + `bundle`/`unbundle` round-trip.** `trace run` yields well-formed JSONL (one header first, one footer last, ≥1 manifest, correct `header.program`); `trace log` yields a non-empty event array containing a `node-enter`; `bundle` → delete originals → `unbundle` restores the source byte-for-byte, and the unbundled trace re-logs to an **identical** event array.
- **`coverage` lifecycle.** `test --coverage` writes `cov-*.json` and reports `1/1` passing; `coverage report --threshold 100` shows a **non-zero** total with `covered === total` at 100%; `coverage clean` removes `.coverage`. (The non-zero check matters — `coverage report` exits 0 with "No coverage data found" on empty data.)

## Shared-helper work (used by cli-main and cli-tier2)

- `helpers.mjs`: new generic `runProcess(executable, args, opts)` boundary; `runInstalledAgency` is now a thin wrapper over it. `runProcess` also runs the packed file with `node`.
- Promoted `readJsonLines` and the normalized file-equality helpers to `helpers.mjs`; `cli-main` now imports them instead of keeping private copies.
- Wired `cli-tier2` into the existing integration job (runs on PRs and pushes) — no new job, secret, or matrix.

## Validation

Against a fresh tarball: `cli-tier2`, `cli-main`, and `cli-lang` all pass; `typecheck` and `lint:structure` clean. Proved the runner can fail: a partial-coverage mutant drops coverage to 66.7% and the coverage oracle fails as intended.

## Note worth a look

A packed program only auto-runs `main` when launched by a **relative** path (`node packed.mjs`). Absolute-path invocation (`node /abs/packed.mjs`) exits 0 but prints nothing. The standalone check invokes it the way a user (and the existing `cli/test.mjs`) does — relative, from its directory. Flagging in case that absolute-path behavior is itself worth a follow-up.

## Follow-ups (per the spec)

- PR 2: `definition`, `models list`, isolated `local`.
- PR 3: `label ingest` with a persistence oracle.

### DIFF
--- a/packages/agency-lang/tests/integration/cli-tier2/test.mjs
+++ b/packages/agency-lang/tests/integration/cli-tier2/test.mjs
@@ -11,6 +11,7 @@ import {
   mkdtempSync,
   readdirSync,
   renameSync,
+  rmSync,
 } from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join, resolve } from "node:path";
@@ -49,6 +50,15 @@ function subdir(name) {
   return full;
 }
 
+// runInstalledAgency plus a clean-success check: the command must leave stderr
+// blank. `pack` is the documented exception (it logs compile progress to
+// stderr) and uses runInstalledAgency directly.
+function runCleanAgency(cwd, args, label) {
+  const result = runInstalledAgency(cwd, args);
+  assertBlank(result.stderr, `[${label}] stderr`);
+  return result;
+}
+
 // --- pack: the standalone-isolation contract --------------------------------
 // pack already has per-PR coverage that runs the output beside the project's
 // node_modules (cli/test.mjs). The missing contract is that the packed file is
@@ -77,34 +87,42 @@ function checkPackStandalone() {
   assertFile(join(packDir, "packed.mjs"), "pack should write packed.mjs");
 
   // Copy only the packed file to a fresh root with no node_modules ancestor.
+  // The root lives outside `dir`, so clean it up ourselves in finally.
   const standaloneRoot = mkdtempSync(join(tmpdir(), "agency-tier2-standalone-"));
-  for (let cur = standaloneRoot; ; cur = dirname(cur)) {
-    assert(
-      !existsSync(join(cur, "node_modules")),
-      `standalone root ancestor unexpectedly has node_modules: ${cur}`,
-    );
-    if (dirname(cur) === cur) break;
-  }
-  cpSync(join(packDir, "packed.mjs"), join(standaloneRoot, "packed.mjs"));
-
-  // Rename the installed project's node_modules while running the copy, so an
-  // absolute reference back to the install fails just as a bare import would.
-  const nodeModules = join(dir, "node_modules");
-  const stashed = join(dir, "node_modules.stashed");
-  renameSync(nodeModules, stashed);
   try {
-    // Invoke it the way a user would — `node packed.mjs` from its own directory.
-    // (A packed program only auto-runs `main` when launched by a relative path.)
-    const result = runProcess(process.execPath, ["packed.mjs"], {
-      cwd: standaloneRoot,
-    });
-    assert(
-      result.stdout.replace(/\r\n/g, "\n") === PACK_EXPECTED,
-      `standalone packed output was:\n${result.stdout}`,
-    );
-    assertBlank(result.stderr, "[pack standalone] stderr");
+    for (let cur = standaloneRoot; ; cur = dirname(cur)) {
+      assert(
+        !existsSync(join(cur, "node_modules")),
+        `standalone root ancestor unexpectedly has node_modules: ${cur}`,
+      );
+      if (dirname(cur) === cur) break;
+    }
+    cpSync(join(packDir, "packed.mjs"), join(standaloneRoot, "packed.mjs"));
+
+    // Remove the source dir so an artifact embedding an absolute source path
+    // fails, and rename the install's node_modules so bare/absolute install
+    // references fail too. Together these prove the copy is self-contained.
+    rmSync(packDir, { recursive: true, force: true });
+    const nodeModules = join(dir, "node_modules");
+    const stashed = join(dir, "node_modules.stashed");
+    renameSync(nodeModules, stashed);
+    try {
+      // Invoke it the way a user would — `node packed.mjs` from its own
+      // directory. (A packed program only auto-runs `main` when launched by a
+      // relative path.)
+      const result = runProcess(process.execPath, ["packed.mjs"], {
+        cwd: standaloneRoot,
+      });
+      assert(
+        result.stdout.replace(/\r\n/g, "\n") === PACK_EXPECTED,
+        `standalone packed output was:\n${result.stdout}`,
+      );
+      assertBlank(result.stderr, "[pack standalone] stderr");
+    } finally {
+      renameSync(stashed, nodeModules);
+    }
   } finally {
-    renameSync(stashed, nodeModules);
+    cleanup(standaloneRoot);
   }
   console.log("[cli-tier2] pack standalone ✓");
 }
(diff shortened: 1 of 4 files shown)


packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -0,0 +1,238 @@
...
+  assert(traceLines[0].type === "header", "trace header must be first");
+  assert(traceLines[traceLines.length - 1].type === "footer", "trace footer must be last");
+  assert(manifests.length > 0, "expected at least one trace manifest (checkpoint)");
+  assert(headers[0].program === "probe.agency", `trace program was ${headers[0].program}`);
+
+  runInstalledAgency(traceDir, ["trace", "log", "probe.trace", "-o", "events.json"]);
+  const baselineEvents = JSON.parse(readText(join(traceDir, "events.json")));
+  assert(Array.isArray(baselineEvents) && baselineEvents.length > 0, "trace log must yield a non-empty array");
+  assert(
+    baselineEvents.some((e) => e.type === "node-enter"),
```
egonSchiele: The fixture deliberately calls `tier2TraceHelper`, but this oracle accepts any `node-enter` event, which every ordinary trace gets from `main`. A valid trace for an unrelated program could pass here, and the later deep comparison would only compare that unrelated baseline with itself. Assert the distinctive semantic event instead: `event.type === "function-enter" && event.functionName === "tier2TraceHelper"`.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -0,0 +1,238 @@
...
+function checkTraceBundleRoundTrip() {
+  const traceDir = subdir("trace");
+  // trace runs the `main` node; a uniquely named helper gives the event log a
+  // distinctive function to look for.
+  const source =
+    `def tier2TraceHelper(): string {\n  return "traced"\n}\n\n` +
+    `node main() {\n  print(tier2TraceHelper())\n}\n`;
+  writeFile(traceDir, "probe.agency", source);
+
+  runInstalledAgency(traceDir, ["trace", "run", "probe.agency", "-o", "probe.trace"]);
```
egonSchiele: Most `runInstalledAgency` results in this runner are discarded, so a command can exit 0 while emitting warnings or partial-error diagnostics to stderr and still pass. That affects trace run/log, bundle/unbundle, test/report/clean; only the standalone Node process currently checks stderr. A small local `runCleanAgency(cwd, args, label)` wrapper can call `runInstalledAgency`, assert blank stderr, and return the result. Keep `pack` as the explicit documented exception if its compile progress is intentionally on stderr.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -0,0 +1,238 @@
...
+      `node main() {\n  print(part() + "_" + add(1, 1))\n}\n`,
+  );
+
+  // pack emits compile progress, so do not assert empty stderr on the command
+  // itself; the empty-stderr contract is asserted on the standalone run below.
+  runInstalledAgency(packDir, ["pack", "probe.agency", "-o", "packed.mjs"]);
+  assertFile(join(packDir, "packed.mjs"), "pack should write packed.mjs");
+
+  // Copy only the packed file to a fresh root with no node_modules ancestor.
+  const standaloneRoot = mkdtempSync(join(tmpdir(), "agency-tier2-standalone-"));
```
egonSchiele, line 91: This standalone root is outside `dir`, so the final `cleanup(dir)` never removes it. The original `packDir` also remains reachable even though the contract above says the sources are gone; renaming only `node_modules` does not catch an artifact that embeds an absolute source path. Wrap the standalone-root lifecycle in an outer `try/finally { cleanup(standaloneRoot) }`, and after copying `packed.mjs`, remove `packDir` before execution. Keep the inner `try/finally` that restores `node_modules`.

egonSchiele (commented):
The overall runner and shared-helper refactor are well scoped, and the coverage lifecycle oracle is effective. Four test-effectiveness issues remain. Copilot’s existing comment about `probe.agency` and `probe.trace` never being deleted is correct and should be treated as part of this review: without that deletion, the round-trip does not prove bundle independence. The inline comments cover the additional false-pass and cleanup paths.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -0,0 +1,238 @@
...
+    "trace log must contain a node-enter event",
+  );
+
+  // Bundle the source + trace, then delete both originals so unbundle cannot
+  // copy external inputs.
+  runInstalledAgency(traceDir, ["bundle", "probe.agency", "probe.trace", "-o", "probe.bundle"]);
+  const savedSource = readText(join(traceDir, "probe.agency"));
+  const originalSourcePath = join(traceDir, "probe.agency.original");
+  writeFile(traceDir, "probe.agency.original", savedSource);
+  runInstalledAgency(traceDir, ["unbundle", "probe.bundle", "-o", "unpacked"]);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 480935988. The runner now rmSyncs probe.agency and probe.trace (and asserts both are absent) before unbundle, after saving the original source for the byte comparison. So unbundle genuinely has no on-disk inputs to fall back to.

packages/agency-lang/tests/integration/helpers.mjs
```diff
@@ -2,26 +2,26 @@
...
-// declare the exact expected exit status rather than bypassing the helper for
-// failure cases, and always get stdout and stderr back separately.
-export function runInstalledAgency(
-  dir,
+// The single process boundary for the integration tests. It is the only layer
+// that knows about spawnSync, the default timeout, environment merging, spawn
+// errors, and exact-status enforcement. Callers declare the expected status
+// rather than bypassing the helper for failure cases, and always get stdout and
+// stderr back separately. On a spawn error or wrong status it throws with the
+// captured streams and status attached, so callers can log on failure.
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 480935988. Reworded to "single spawnSync boundary" and noted that setup/install still use execSync via run().

packages/agency-lang/tests/integration/cli-tier2/test.mjs
egonSchiele: Fixed in 480935988. Now asserts event.type === "function-enter" && event.functionName === "tier2TraceHelper" (verified that field exists on function-enter events). An unrelated program's trace no longer satisfies it.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
egonSchiele: Fixed in 480935988. Added runCleanAgency(cwd, args, label) which asserts blank stderr, and used it for trace run/log, bundle, unbundle, and test/report/clean. pack stays the explicit exception (compile progress on stderr). I confirmed all the converted commands really do leave stderr blank on success.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
egonSchiele: Fixed in 480935988. The standalone-root lifecycle is now wrapped in try/finally { cleanup(standaloneRoot) } (it lives outside the temp project), and packDir is removed after copying packed.mjs and before execution, so an artifact embedding an absolute source path fails too. The inner try/finally still restores node_modules.

egonSchiele (commented):
Re-review of 480935988: all prior findings are resolved. The runner now deletes and verifies absence of bundle inputs before unbundling, asserts the helper-specific trace event, enforces clean stderr for deterministic successes, removes the pack source tree before standalone execution, restores `node_modules` in `finally`, and cleans the external standalone temp root. The shared-helper and cli-main refactors remain behaviorally compatible, and no new actionable findings were found. Latest CI integration, typecheck, and structural lint checks are green.

PR #817, opened 2026-08-06
test(integration): per-PR tests for definition, models list, and isolated local

egonSchiele:
## What

Tier 2 CLI integration tests, **PR 2 of 3** (read-only introspection). Adds three checks to the `cli-tier2` runner from PR 1. These commands had **no** integration coverage anywhere.

## What it checks

- **`definition`** (LSP go-to-definition, source on stdin). A fixture with two functions called on one line; the cursor points at the *second* call. The JSON result is deep-compared with the second function's declaration (`{ file, line, column }`), coordinates derived from fixture markers. A swapped-cursor call must return a *different* result — proving the `--line`/`--column` arguments are honored, not ignored.
- **`models list`** (bundled catalog, offline). Requires a real data row after the `NAME`/`PROVIDER` header (the empty catalog still prints the header, so header presence alone proves nothing), validating a non-empty name/provider and a numeric context column. No model name, count, or price is pinned.
- **`local list` / `local resolve`, fully isolated.** Sets `AGENCY_MODELS_DIR` (with a sentinel `.gguf`), `AGENCY_LLAMA_PROVIDER_MODULE` (a project-owned module — the provider gate needs it, since `smoltalk-llama-cpp` isn't on CI), and an isolated `HOME`, plus a project `agency.json` whose `client.modelsDir` points at a *decoy* dir. `list` must show the env sentinel and exclude the decoy (proving `AGENCY_MODELS_DIR` wins over config); `resolve smollm2-135m` pins the exact curated URI.

## Notes

- No CI or helper changes — the runner and its wiring landed in PR 1.
- A fresh `HOME` makes `npm`/`npx` run its update check and print a notice to stderr, so the local checks disable the notifier (`npm_config_update_notifier=false`) to keep the empty-stderr assertions about the command, not the launcher.

## Validation

Against a fresh tarball: all six `cli-tier2` workflows pass; `typecheck` and `lint:structure` clean. Proved the runner can fail: dropping `AGENCY_MODELS_DIR` makes `local list` fall back to the decoy dir and the isolation oracle fails as intended.

## Follow-up

PR 3: `label ingest` with a persistence oracle (the last of the Tier 2 set).

### DIFF
--- a/packages/agency-lang/tests/integration/cli-tier2/test.mjs
+++ b/packages/agency-lang/tests/integration/cli-tier2/test.mjs
@@ -15,6 +15,7 @@ import {
 } from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join, resolve } from "node:path";
+import { pathToFileURL } from "node:url";
 import {
   assert,
   assertFile,
@@ -325,8 +326,13 @@ function checkModelsList() {
       "",
     ].join("\n"),
   );
+  // Use the single-option `--import=<file-url>` form: NODE_OPTIONS is
+  // space-tokenized, so a raw path breaks when TMPDIR contains a space, and a
+  // bare path is not a safe ESM specifier on Windows. A pathToFileURL href
+  // encodes spaces and is platform-correct.
+  const preloadUrl = pathToFileURL(join(modelsDir, "blocknet.mjs")).href;
   const result = runInstalledAgency(modelsDir, ["models", "list"], {
-    env: { NODE_OPTIONS: `--import ${join(modelsDir, "blocknet.mjs")}` },
+    env: { NODE_OPTIONS: `--import=${preloadUrl}` },
   });
   assertBlank(result.stderr, "[models list] stderr");
   const lines = result.stdout.split("\n");
--- a/packages/agency-lang/tests/integration/cli-tier2/test.mjs
+++ b/packages/agency-lang/tests/integration/cli-tier2/test.mjs
@@ -269,8 +269,11 @@ function checkDefinition() {
   ].join("\n");
   const lines = source.split("\n");
   const callLine = lines.findIndex((l) => l.includes("alpha() + beta()"));
+  assert(callLine >= 0, "fixture must contain the alpha() + beta() call line");
   const callColumn = lines[callLine].indexOf("beta()");
+  assert(callColumn >= 0, "fixture call line must contain beta()");
   const defLine = lines.findIndex((l) => l.startsWith("def beta"));
+  assert(defLine >= 0, "fixture must contain beta's declaration");
   assert(callLine !== callColumn, "fixture line and column must differ for the swap check");
 
   const result = runInstalledAgency(
@@ -285,16 +288,17 @@ function checkDefinition() {
     `definition returned ${result.stdout}`,
   );
 
-  // Swapping line and column must change the answer (here it resolves to
-  // nothing), proving the cursor arguments are honored rather than ignored.
+  // Swapping line and column points at an out-of-range position, which must
+  // resolve to null — proving the cursor arguments are honored, not ignored.
   const swapped = runInstalledAgency(
     dir,
     ["definition", "--line", String(callColumn), "--column", String(callLine), "--file", "probe.agency"],
     { input: source },
   );
+  assertBlank(swapped.stderr, "[definition swapped] stderr");
   assert(
-    swapped.stdout.trim() !== result.stdout.trim(),
-    `swapped cursor returned the same result as the real cursor: ${swapped.stdout}`,
+    JSON.parse(swapped.stdout) === null,
+    `swapped cursor should resolve to null, got: ${swapped.stdout}`,
   );
   console.log("[cli-tier2] definition ✓");
 }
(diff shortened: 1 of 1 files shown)


packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -247,6 +247,117 @@ function checkCoverageLifecycle() {
...
+    swapped.stdout.trim() !== result.stdout.trim(),
+    `swapped cursor returned the same result as the real cursor: ${swapped.stdout}`,
+  );
+  console.log("[cli-tier2] definition ✓");
+}
+
+// --- models list (bundled catalog, offline) ---------------------------------
+
+function checkModelsList() {
+  const result = runInstalledAgency(dir, ["models", "list"]);
```
egonSchiele: The test describes this as an offline bundled-catalog contract, but this process still has normal network access. If `models list` regressed to fetching the remote catalog before printing, it could pass unchanged on networked CI. Please run this invocation with outbound access denied (for example, a small `NODE_OPTIONS` preload that throws from `fetch`/HTTP/socket entry points), so any network attempt fails the test rather than satisfying it.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -247,6 +247,117 @@ function checkCoverageLifecycle() {
...
+  assertBlank(result.stderr, "[models list] stderr");
+  const lines = result.stdout.split("\n");
+  const headerIdx = lines.findIndex((l) => l.includes("NAME") && l.includes("PROVIDER"));
+  assert(headerIdx >= 0, "models list must print a NAME/PROVIDER header");
+  // The empty catalog still prints the header, so require a real data row.
+  const dataRow = lines.slice(headerIdx + 1).find((l) => l.trim().length > 0);
```
egonSchiele, line 343: This accepts any nonempty post-header line with two tokens and a numeric final token; for example, `gpt-4o-mini 128000` passes even though the provider, open-weights, and both pricing columns disappeared. Please require the six table fields and validate their shapes (`provider`, `yes|no`, two numeric prices, and a positive numeric context), so the row is structurally valid without pinning volatile catalog values.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -247,6 +247,117 @@ function checkCoverageLifecycle() {
...
+  assertBlank(result.stderr, "[definition] stderr");
+  const got = JSON.parse(result.stdout);
+  assert(
+    JSON.stringify(got) === JSON.stringify({ file: "probe.agency", line: defLine, column: 0 }),
+    `definition returned ${result.stdout}`,
+  );
+
+  // Swapping line and column must change the answer (here it resolves to
+  // nothing), proving the cursor arguments are honored rather than ignored.
+  const swapped = runInstalledAgency(
```
egonSchiele, line 294: The intended swapped-position result is `null`, but the current oracle accepts any different stdout, including malformed JSON, an error object, or diagnostic text; it also does not enforce blank stderr for this success. Please parse this result and assert it is exactly `null`, plus `assertBlank(swapped.stderr, ...)`, so the negative half proves the stated cursor behavior rather than merely different output.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -247,6 +247,117 @@ function checkCoverageLifecycle() {
...
+  writeFile(envModels, "env-sentinel.gguf", "");
+  writeFile(configModels, "config-decoy.gguf", "");
+  writeFile(localDir, "agency.json", JSON.stringify({ client: { modelsDir: configModels } }, null, 2) + "\n");
+  writeFile(localDir, "provider.mjs", "export default {};\n");
+
+  // The provider gate needs AGENCY_LLAMA_PROVIDER_MODULE (smoltalk-llama-cpp is
+  // not installed on CI); the isolated HOME keeps ~/agency.json aliases out.
+  // A fresh HOME makes npm run its update check and print a notice to stderr, so
+  // disable the notifier to keep the empty-stderr assertions about the command.
+  const env = {
```
egonSchiele, line 382: `HOME` isolates aliases, but the support gate can still discover `smoltalk-llama-cpp` through `npm root -g` / `pnpm root -g`. On a host with a global install, this test could pass even if `AGENCY_LLAMA_PROVIDER_MODULE` stopped bypassing the gate. Please point npm/pnpm global discovery at empty project-owned roots (or otherwise block it) so the supplied provider override is the only support path and the test is fully host-independent.

egonSchiele (commented):
The happy paths are well targeted, but four assertions/isolation boundaries can still pass after the intended contract breaks. I left inline suggestions for making those oracles fail closed.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -247,6 +247,117 @@ function checkCoverageLifecycle() {
...
+    "  return alpha() + beta()",
+    "}",
+    "",
+  ].join("\n");
+  const lines = source.split("\n");
+  const callLine = lines.findIndex((l) => l.includes("alpha() + beta()"));
+  const callColumn = lines[callLine].indexOf("beta()");
+  const defLine = lines.findIndex((l) => l.startsWith("def beta"));
+  assert(callLine !== callColumn, "fixture line and column must differ for the swap check");
+
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2f3cfc323. callLine, callColumn, and defLine are now asserted (>= 0) before use, so a broken fixture fails with an actionable message instead of a TypeError.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
egonSchiele: Fixed in 2f3cfc323. models list now runs under a NODE_OPTIONS preload that throws from fetch, http/https request/get, and net.Socket.connect. Verified the command still succeeds offline, so any regression to fetching the remote catalog would fail the test.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
egonSchiele: Fixed in 2f3cfc323. The row parse now requires exactly six columns and validates each: non-empty provider, open-weights is yes|no, both prices numeric, and a positive-integer context. "gpt-4o-mini 128000" would no longer pass.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
egonSchiele: Fixed in 2f3cfc323. The swapped-cursor call is now parsed and asserted to be exactly null, with assertBlank(swapped.stderr). Confirmed the out-of-range position returns null.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
egonSchiele: Fixed in 2f3cfc323. The local checks now set npm_config_prefix and PNPM_HOME to an empty project-owned dir, so npm root -g / pnpm root -g resolve to a nonexistent/empty root and global discovery finds nothing. The provider override is the only support path, making the test host-independent. Verified npx --no-install still resolves the local bin under these settings.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -247,6 +247,156 @@ function checkCoverageLifecycle() {
...
+      'const boom = (w) => () => { throw new Error(`network blocked by test preload (${w})`); };',
+      'globalThis.fetch = boom("fetch");',
+      'http.request = boom("http.request"); http.get = boom("http.get");',
+      'https.request = boom("https.request"); https.get = boom("https.get");',
+      'net.Socket.prototype.connect = boom("socket.connect");',
+      "",
+    ].join("\n"),
+  );
+  const result = runInstalledAgency(modelsDir, ["models", "list"], {
+    env: { NODE_OPTIONS: `--import ${join(modelsDir, "blocknet.mjs")}` },
```
egonSchiele: This raw filesystem path is tokenized by `NODE_OPTIONS`, so the test fails before Agency runs when `TMPDIR` contains a space; a Windows absolute path is also not a safe ESM specifier. Please import `pathToFileURL` and pass an encoded file URL using the single-option form `--import=<file-url>` (for example, build the URL from `pathToFileURL(join(modelsDir, "blocknet.mjs")).href`), which handles spaces and platform-specific paths correctly.

egonSchiele (commented):
The prior findings are addressed. I found one portability issue in the new offline preload wiring.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
egonSchiele: Fixed in a931d888d. The preload is now passed as `--import=${pathToFileURL(join(modelsDir, "blocknet.mjs")).href}` — the single-option form with an encoded file URL, so a TMPDIR containing a space no longer breaks NODE_OPTIONS tokenization and it is a valid specifier on Windows. Verified models list still runs offline under it.

egonSchiele (commented):
Re-reviewed the latest head. The encoded file-URL preload fixes the remaining portability issue, all prior findings remain resolved, and I found no additional actionable issues. The latest integration and other CI checks are green.

PR #818, opened 2026-08-06
test(integration): per-PR successful label ingest with a persistence oracle

egonSchiele:
## What

Tier 2 CLI integration tests, **PR 3 of 3** (the last). Adds a `label ingest` check to the `cli-tier2` runner.

## Why

`label ingest` had *partial* per-PR coverage — `cli/test.mjs` checks that the parent `--store` option reaches the subcommand, but with a deliberately missing input, so it never tests a real ingest. This is the first **successful ingest + persistence** check.

## What it checks (`checkLabelIngest`)

- Writes a top-level JSON array of N distinct strings (the only shape the JSON loader accepts) and runs `label ingest data.json --store labels --source batch`.
- Asserts the summary counts exactly: **N new records, 0 already stored, N new occurrences, 0 already recorded** (blank stderr; the "new field" note is on stdout).
- Parses `outputs.jsonl`: one record per fixture string (order-independent), building an outputId-by-string map.
- Parses `occurrences.jsonl`: one per item, `source: "batch"`, `origin.kind: "json"`, `origin.itemKey: "data.json"`, each `itemIndex` exactly once, and each `outputId` referencing the output whose `fields.output` equals `fixture[itemIndex]`.
- Re-ingests and requires **0 new / N already**, with both JSONL files **byte-identical** — proving the store reopens and replays without reimplementing store validation in the test.

Deep matrices (hashing, locking, corruption, migration) stay in the label unit tests.

## Validation

Against a fresh tarball: all seven `cli-tier2` workflows pass; `typecheck` and `lint:structure` clean. Proved the runner can fail: a duplicate fixture string exercises content-identity dedup and the oracle fails as intended (`2 new records, 1 already stored`).

## Tier 2 complete

With this, the Tier 2 set is done: `pack` (standalone), `trace`/`bundle`, `coverage` (PR 1); `definition`, `models list`, `local` (PR 2); `label ingest` (this PR).

### DIFF
--- a/packages/agency-lang/tests/integration/cli-tier2/test.mjs
+++ b/packages/agency-lang/tests/integration/cli-tier2/test.mjs
@@ -403,6 +403,69 @@ function checkLocalIsolated() {
   console.log("[cli-tier2] local isolated ✓");
 }
 
+// --- label ingest (store mutation + persistence oracle) ---------------------
+
+function checkLabelIngest() {
+  const labelDir = subdir("label");
+  const fixture = ["tier2 first output", "tier2 second output", "tier2 third output"];
+  const n = fixture.length;
+  // A top-level JSON array of strings is the only shape the JSON loader accepts.
+  writeFile(labelDir, "data.json", JSON.stringify(fixture) + "\n");
+
+  const first = runInstalledAgency(labelDir, ["label", "ingest", "data.json", "--store", "labels", "--source", "batch"]);
+  assertBlank(first.stderr, "[label ingest] stderr");
+  const firstOut = stripAnsi(first.stdout);
+  assertIncludes(firstOut, `${n} new records, 0 already stored`);
+  assertIncludes(firstOut, `${n} new occurrences, 0 already recorded`);
+
+  // outputs.jsonl: exactly one record per fixture string (order-independent).
+  const outputs = readJsonLines(join(labelDir, "labels", "outputs.jsonl"));
+  assert(outputs.length === n, `expected ${n} output records, got ${outputs.length}`);
+  const outputIdByString = {};
+  for (const record of outputs) {
+    const value = record.fields.output;
+    assert(fixture.includes(value), `unexpected output record: "${value}"`);
+    outputIdByString[value] = record.outputId;
+  }
+  for (const value of fixture) {
+    assert(value in outputIdByString, `outputs.jsonl is missing a record for "${value}"`);
+  }
+
+  // occurrences.jsonl: one per fixture item, each referencing the right output.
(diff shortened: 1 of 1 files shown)


packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -403,6 +403,69 @@ function checkLocalIsolated() {
...
+    );
+  }
+  seenIndexes.sort((a, b) => a - b);
+  assert(
+    JSON.stringify(seenIndexes) === JSON.stringify([...Array(n).keys()]),
+    `expected itemIndexes 0..${n - 1}, got ${seenIndexes}`,
+  );
+
+  // Re-ingesting the same data is idempotent: nothing new, both files unchanged.
+  const outputsBefore = readText(join(labelDir, "labels", "outputs.jsonl"));
```
egonSchiele: The spec requires these files to remain byte-identical, but `readText()` normalizes `\r\n` to `\n`. A broken re-ingest that rewrites LF JSONL as CRLF would therefore pass both comparisons. Please snapshot each file with `readFileSync(path)` and compare the resulting buffers with `Buffer.equals()` after the second run, so the persistence oracle checks raw bytes as claimed.

egonSchiele (commented):
The successful-ingest, provenance, and reopen/replay coverage is well targeted. I found one false-pass path in the byte-identity oracle.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
```diff
@@ -403,6 +403,69 @@ function checkLocalIsolated() {
...
+  assert(outputs.length === n, `expected ${n} output records, got ${outputs.length}`);
+  const outputIdByString = {};
+  for (const record of outputs) {
+    const value = record.fields.output;
+    assert(fixture.includes(value), `unexpected output record: "${value}"`);
+    outputIdByString[value] = record.outputId;
+  }
+  for (const value of fixture) {
+    assert(value in outputIdByString, `outputs.jsonl is missing a record for "${value}"`);
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 6a470da91. outputIdByString is now Object.create(null) and membership uses Object.hasOwn, so a fixture string like __proto__ can't collide with prototype keys.

packages/agency-lang/tests/integration/cli-tier2/test.mjs
egonSchiele: Fixed in 6a470da91. Both files are now snapshotted with readFileSync (Buffer) before the re-ingest and compared with Buffer.equals afterward, so the check is on raw bytes. A re-ingest that rewrote LF as CRLF would now fail rather than pass through readText's normalization.

PR #820, opened 2026-08-07
Commander fork: flag ownership, program boundaries in the parser, agent flag delegation

egonSchiele:
Implements the CLI flag-ownership design: vendor commander v15 into the repo, build the program boundary and flag-ownership rules into the parser itself, and delegate all `agency agent` flags to the agent.

Spec and plan live on `adit/flag-position-issue` (`docs/superpowers/specs/2026-08-06-cli-flag-ownership-design.md`, `docs/superpowers/plans/2026-08-06-cli-flag-ownership.md`). **Stacked on #821** (the pristine commander v15 vendoring, a verified no-op). This PR is based on that branch, so its diff of `lib/vendor/commander/*.js` shows exactly the fork modifications — every hunk is an `// AGENCY FORK` block ledgered in `MODIFICATIONS.md`. Merge #821 first, then change this PR base back to `main`.

## The user-visible model

1. **Position decides agency vs. program** (unchanged): everything after the filename — or after the word `agent` — belongs to the program, no matter what the flag is named. The misplaced-flag warning survives, now driven by parser-recorded provenance, so `-- ` suppression still works.
2. **Ownership decides root vs. subcommand** (new): a flag is valid anywhere after the command that owns it, directly or as an ancestor, and never before it. Misplacement is an error naming the owner and the fix (`--model belongs to 'agency run'; write it after 'run'`) instead of `Unknown command 'run'`. Command typos get suggestions (`agency formt` → did you mean `format`).
3. **The shorthand IS run**: `agency greet.agency` dispatches the real `run` command object (fork `fallbackCommand`), deleting the duplicated hidden default command.
4. **The agent owns nothing**: every flag after `agent` is the agent programs, whose schema is the single source of help and errors. Budget flags moved into that schema; the launcher pre-scans exactly the flags that must act before spawn (`--config`, `--trace`, `--log`, `--max-cost`, `--max-time`, grandfathered `--agent-home`), validates budgets before launching, and `--config` triggers an isolated staged recompile that never writes into the installed package.

## Stages

- **Vendor (no-op)**: split out as #821.
- **Fork modifications** (each marked `// AGENCY FORK` and ledgered): duplicate-spelling-on-one-path registration error (immediately caught a real bug: `logs view -f` silently received undefined); boundary-aware delegation with per-command `passThroughOptions` and an immediate mode; `boundaryInfo()` provenance; `fallbackCommand` with provenance threaded through dispatch; one pure `_consumeOptionToken` shared by the mutating parse and the non-mutating path resolver — no second tokenizer anywhere.
- **CLI rewiring**: `splitCommandLine` and its shadow tokenizer deleted (`commandLine.ts` is presentation-only); `runCli` is just `parseAsync`.
- **Agent delegation**: `resolveAgentLaunchArgs` as the single launcher scanner; `stageConfiguredAgent` isolated compile with guarded cleanup (child exit/error, sync spawn failure, and process.exit-mid-compile via a subprocess fixture).

## Testing

- 53 fork tests (`lib/vendor/commander/*.test.ts`), launcher scanner/orchestration/staging tests, warning presentation via real parses.
- Integration Test 8 grew 8d (ownership matrix, typo suggestions, nested defaults, budget cap observed in the child env, trace/test defaults) and 8e (agent delegation, launcher budget rejection, malformed-config exact error, staged-config runs with the whole installed agent tree hashed byte-identical before/after).
- Full unit suite: 11,816 passing. `pnpm run typecheck` (all three tsconfigs) and `lint:structure` clean.

Notable find during execution: macOS `os.tmpdir()` is a symlink, and compiled programs is-main guard compares real paths — the staged agent would load and silently never run. The staged root is realpathed, and the integration suite now exercises the real staged-run path.


packages/agency-lang/lib/cli/stageConfiguredAgent.ts
```diff
@@ -0,0 +1,111 @@
...
+  } catch (error) {
+    cleanup();
+    throw error;
+  }
+}
+
+/** Owner read/write on files, read/write/execute on directories — applied to
+ *  the COPY only, so a read-only installed tree stages compilable. */
+function makeTreeWritable(directory: string): void {
+  fs.chmodSync(directory, 0o755);
```
egonSchiele: `mkdtempSync` initially gives this stage an owner-only root, but this changes it to `0755` and the files below to `0644`. The generated `agent.js` can contain literal `client.apiKey` and statelog credentials from the explicit config, and the tree remains alive for the child’s lifetime, so another user on a shared machine can read them from the temp directory. Please keep staged directories `0700` and files `0600`, and add a POSIX mode assertion for the stage root and generated run file.

packages/agency-lang/lib/vendor/commander/command.js
```diff
@@ -1560,6 +2029,28 @@ Expecting one of '${allowedValues.join("', '")}'`);
...
+    // already consumed by the parent (the fallback shorthand). The boundary
+    // has already fallen, so the entire unknown is the program tail — parsing
+    // it would misread the program's flags as unknown options
+    // (MODIFICATIONS.md #5).
+    if (
+      this._passThroughOptions &&
+      this._boundaryMode === 'first-operand' &&
+      operands.length > 0
+    ) {
+      const viaSeparator = unknown[0] === '--';
```
egonSchiele, line 2042: The pre-input separator is lost on the fallback path. For `agency -- foo.agency --max-cost 5`, the parent has already consumed `--` and passes the whole suffix in `operands`, so `unknown` is empty here and `boundaryInfo()` becomes `{ tail: [], viaSeparator: false }`. The explicit `agency run -- ...` form records `{ tail: ["--max-cost", "5"], viaSeparator: true }`. Please preserve the separator and tail through fallback dispatch and add the pre-input shorthand counterpart to the existing fallback separator tests.

egonSchiele (commented):
I found two issues worth addressing before merge: the configured-agent stage exposes baked credentials on shared systems, and the fallback shorthand loses separator provenance.

egonSchiele:
Addressed all five review comments in 1499dae52 (plus the base flip to main now that #821 is merged):

- **Staged-tree credentials (egonSchiele)**: the staged copy is now owner-only — 0700 directories, 0600 files — re-tightened after compilation so files the compiler *creates* (which would otherwise get umask modes) are covered too. POSIX mode assertions added for the stage root, the generated run file, and a subdirectory.
- **Pre-input separator through fallback (egonSchiele)**: `agency -- foo.agency --max-cost 5` now records `{ tail: ["--max-cost", "5"], viaSeparator: true }`, matching the explicit `agency run -- ...` form — the fallback parent keeps `--` at the head of the dispatched tail. Test added beside the post-input case.
- **Failed-compile path (Copilot)**: goes through `cleanup()` + the injected exit seam with an early return; orchestration test added. (The staged-leak half of the comment could not actually occur — `runFile === null` is only reachable on the compile-fallback branch where nothing was staged — but the seam bypass was real.)
- **Dead `bareIsMeaningful` field (Copilot)**: removed; the table comment now states the actual contract (bare scans to "", meaning decided in `resolveAgentLaunchArgs`).
- **Order-dependent hash comparison (Copilot)**: the integration tree-hash check now compares sorted entries.

Verified: tsc clean, 1,113 unit tests green (`lib/cli` + `lib/vendor/commander`), and the full packed integration suite green including Test 8e.

PR #822, opened 2026-08-07
pnpm supply-chain hardening: 7-day release-age cooldown with first-party exclusions

egonSchiele:
Closes the main npm supply-chain attack window for the whole ~211-package production closure: new versions of any dependency must be public for **7 days** before pnpm will resolve them, with first-party packages exempted so the publish-then-immediately-use dev loop is unaffected.

## What changed

- **`pnpm-workspace.yaml`**: `minimumReleaseAge: 10080` (7 days) with `minimumReleaseAgeExclude: [smoltalk, tarsec, typestache]` — exclusion is by name and covers all versions, so publishing your packages and using them from agency the same minute keeps working. Plus `blockExoticSubdeps: true` (no git/tarball transitive sources — the current lockfile has zero) and explicit `allowBuilds` **denials** for the three packages that request install scripts (esbuild, @google/genai, protobufjs) — all three have been silently blocked since pnpm 10 and everything works.
- **Root `package.json`**: `"packageManager": "pnpm@11.20.0"`. These settings require pnpm ≥ 10.16 and are **silently ignored** by older pnpm — that silent-off failure mode is why the version is pinned in one place. Modern pnpm self-switches to the pinned version automatically.
- **Workflows**: the 14 `version: 9` pins across 6 workflows are removed; `pnpm/action-setup` reads the `packageManager` field instead. (CI was on pnpm 9, which would have ignored every one of these settings.)
- **Docs**: `docs/dev/supply-chain.md` + CLAUDE.md pointer — including the "never reintroduce a workflow version pin" warning and the versioned-exclusion escape hatch (`name@x.y.z`) for when one specific fresh third-party version is genuinely needed.

## Verification

- Lockfile is byte-identical (pnpm 11 keeps lockfileVersion 9.0); `pnpm install` clean with the explicit script denials.
- **Behavioral probe of the cooldown**: at probe time `openai@7.4.0` was 4 days old and `7.3.0` was 6.99 days old; a fresh `pnpm add openai` resolved **7.2.0** (8 days old) — both fresh versions correctly refused. (The smoltalk exclusion could not be probed the same way — its latest release is 2 weeks old, outside the window either way.)
- Smoke: workspace tests green after reinstall under pnpm 11.20.0.

Note for the future rather than this PR: CI runs plain `pnpm install`; `--frozen-lockfile` would be a natural next hardening step.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,8 +18,6 @@ jobs:
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - uses: pnpm/action-setup@f40ffcd9367d9f12939873eb1018b921a783ffaa # v4
-        with:
-          version: 9
       - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: 22
--- a/.github/workflows/local-model.yml
+++ b/.github/workflows/local-model.yml
@@ -44,7 +44,6 @@ jobs:
       - name: Set up pnpm
         uses: pnpm/action-setup@f40ffcd9367d9f12939873eb1018b921a783ffaa # v4
         with:
-          version: 9
           run_install: false
 
       - name: Set up Node.js
--- a/.github/workflows/local-model.yml
+++ b/.github/workflows/local-model.yml
@@ -58,12 +58,13 @@ jobs:
       - name: Build
         run: make
 
-      # Install smoltalk-llama-cpp ONLY here, at a pinned version. pnpm 9's
-      # `add` rejects `--save=false` (ERR_PNPM_OPTION_NOT_SUPPORTED), so we add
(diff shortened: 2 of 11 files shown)


egonSchiele:
All three review comments are addressed in fe7fa46e6 (pushed just after the review landed):

- Both packageManager-mechanism wordings now name the three consumers precisely — pnpm >= 10 switches natively (managePackageManagerVersions, verified in this repo), Corepack shims where enabled, and pnpm/action-setup in CI — plus the one unprotected case (a pre-10 global pnpm ignores the field AND the settings; the check is that `pnpm --version` in the repo prints 11.20.0).
- The local-model comment is version-agnostic now.
- Bonus catch from re-reading that workflow: it installs smoltalk-llama-cpp at an exact freshly-bumped pin, and exact-version requests are gated by the cooldown too — so it joined the first-party exclusions, or that job would fail for a week after every bump.

PR #823, opened 2026-08-07
Per-invocation config overrides + injectable trace id

egonSchiele:
## What

Lets a single invocation — a node called from TypeScript, a served function or node, or a serve resume — carry a **config override** and an optional **root trace id** that apply to that call only, on both call surfaces. This is what statelog needs to run a hosted agent under a caller-supplied credential and a pre-chosen trace id, instead of the observability config being frozen at import for every call.

Spec, review, and plan are included in the branch (`packages/agency-lang/2026-08-07-per-invocation-config-override-*.md`).

## Design

**Agency is a mechanism; the host owns policy.** Agency applies the override with plain **override-wins** semantics and enforces nothing — no budget clamping, no bounds-checking, no trust filtering. Clamping (budget-as-request) and credential minting live in the host (statelog).

The flow funnels through one boundary:

```
InvocationOptions  (public: { config?, traceId? })
  → transport layers forward it unchanged (serve adapter, discovery, codegen)
    → resolveInvocation(request)        ← the ONE owner of run-id policy + config allow-list
      → ResolvedInvocation { runId, contextOverride }
        → createExecutionContext(resolved) applies the override over the frozen parent
          → finishServedInvocation derives outcome.traceId from execCtx.getRunId()
```

- **`resolveInvocation()`** projects the caller's raw config down to a **positive** v1 allow-list (`observability`, `log.{host,apiKey,projectId,requestTimeoutMs,metadata}`, `budget`, `maxCallDepth`, `failurePropagation`) — building fresh objects, never spreading `config.log`, so a future dangerous sub-field can't leak. It also owns run-id precedence: fresh = inherited subprocess id → supplied `traceId` → `nanoid()` (empty supplied id rejected); resume always keeps `interrupt.runId` and ignores any supplied `traceId`.
- **`createExecutionContext`** now takes a `ResolvedInvocation` (no string overload; all fixed-id callers pass `{ runId }`) and applies the override through the existing runtime merge.
- **`RouteResult.traceId`** is optional, present only on post-execution results (mirrors `usage`).
- Node wrapper codegen extends its existing options object with `config`/`traceId`, aliased to hidden `__invocation*` names so a node parameter named `config`/`traceId`/`messages`/`callbacks` can't collide.

**Explicit-argument threading, not AsyncLocalStorage** — the options are immutable call data and every entry point already has a typed route to the run core, so forwarding is concurrency-safe with no ambient state.

Inert in this channel by design: `log.logFile`, `log.debugMode`, `traceFile`, `traceDir`, all `client.*` (incl. `defaultModel`/`providerModules`). `defaultModel` is deferred (must ship with `defaultProvider`).

## Tests

- `invocationOptions.test.ts` — resolver: run-id precedence, empty-id rejection, resume-keeps-id, positive projection.
- `context.perInvocationOverride.test.ts` — override applied per child, parent unmutated.
- `invocationOutcome.test.ts` — outcome `traceId` derives from the execution context.
- `nodeWrapperParams.test.ts` — generated wrapper uses collision-safe aliases and forwards the invocation.
- `adapter.perInvocation.test.ts` — adapter forwards the exact options by identity; `traceId` present only post-execution.
- `perInvocation.integration.test.ts` — real cores + `StatelogClient` with a mocked fetch: injection, per-call credential, and **concurrent no-cross-attribution** (held at a barrier to force overlap).

`make` build, `pnpm run typecheck`, and `pnpm run lint:structure` all green.

## Follow-ups

- statelog can drop its bake-budget-and-recompile workaround once this ships (budget becomes a per-call override).
- An `agency-js` end-to-end fixture for the TS-interop path was deferred (needs a built `dist` + the fetch-mock harness to run locally); the generator unit test covers the codegen shape and the integration test covers the runtime path end-to-end.
- Per-call model selection (`defaultModel` + `defaultProvider`) intentionally out of scope.


packages/agency-lang/lib/runtime/state/context.ts
```diff
@@ -376,7 +378,28 @@ export class RuntimeContext<T> {
...
+    const { runId, contextOverride } = invocation;
+    // Layer the already-narrowed per-invocation override on top of this frozen
+    // parent's config, through the ONE runtime merge. `contextOverride` is
+    // undefined for a plain run, in which case `effective` is just the parent's
+    // values. The resolver has already projected the allow-list, so this method
+    // never re-derives it.
+    const effective: RuntimeContextConstructorArgs =
+      applyRuntimeConfigOverridesToContextArgs(
+        {
+          statelogConfig: this.statelogConfig,
```
egonSchiele, line 416: **High — preserve the effective remote telemetry sink in subprocesses.** This execution context correctly receives the per-invocation `statelogConfig`, but `_run()` later asks `getStatelogSink()` for only `{ observability, logFile }`. Consequently `withParentStatelog()` cannot forward an invocation override for `log.host`, `apiKey`, `projectId`, `requestTimeoutMs`, or `metadata`; the parent traces with the caller-supplied sink while a subprocess falls back to the compiled module's sink/credential. Please expose and forward the full allowed effective telemetry projection to subprocess bootstrap, and add a subprocess test asserting the inherited `(traceId, host, Authorization, projectId)` values.

egonSchiele (commented):
The implementation is close, but CI is currently red in unit-test shards 1, 3, and 4 because the TypeScript generator's expected fixture outputs were not regenerated. Please run `make fixtures` and commit the resulting fixture updates. I also left one correctness comment about subprocess telemetry inheritance. (Copilot's existing comment about `runExportedFunctionCore` omitting the inherited subprocess run ID is valid as well.)

packages/agency-lang/lib/runtime/state/context.ts
egonSchiele: Fixed in 50979865446c3efe29743fb8fe02b20140f72cc1. Verified the gap: `getStatelogSink()` returned only `{ observability, logFile }`, so `withParentStatelog()` could never forward a per-invocation `log.host`/`apiKey`/`projectId` to a child — the subprocess fell back to the frozen module sink.

Changes:
- `getStatelogSink()` now returns the full allow-listed effective telemetry projection (`observability`, `host`, `apiKey`, `projectId`, `requestTimeoutMs`, `metadata`, `logFile`) read off the execution context's *effective* `statelogConfig` (so it includes the per-invocation override). It never exposes client/provider config.
- `withParentStatelog()` forwards those into the child's `log` override, parent-effective-sink winning over the child's own `log` (the child joins the parent's trace). The empty-`log` shape and "explicit child logFile wins" / "observability off ⇒ forward nothing" cases are preserved.

Tests:
- `ipc.test.ts` — new cases: full sink forwarded (host/apiKey/projectId/requestTimeoutMs/metadata); parent effective sink wins over child `log`.
- `context.perInvocationOverride.test.ts` — `getStatelogSink()` on an execCtx built with a per-invocation observability override returns the overridden host/apiKey/projectId.

These two units chain to prove the path (effective sink exposed → forwarded into `AGENCY_CONFIG_OVERRIDES` → child applies it). I did **not** add a full real-fork end-to-end test asserting the child's on-the-wire `Authorization`/`trace_id`, because that needs a compiled child entry (a built `dist`) rather than an in-process unit; the existing subprocess IPC integration suite covers the fork/env plumbing in CI. Happy to add a live-fork assertion if you'd prefer it here — let me know.

packages/agency-lang/lib/runtime/node.ts
```diff
@@ -273,15 +274,17 @@ type RunExportedFunctionArgs = {
...
 
 /** The outcome-producing core. The lifecycle boundary starts the moment the
  *  execution context exists (invocation started), so a bootstrap/setup failure
  *  still yields an outcome with usage and still runs cleanup. */
 async function runExportedFunctionCore(
-  { ctx, fn, namedArgs, initializeGlobals }: RunExportedFunctionArgs,
+  { ctx, fn, namedArgs, initializeGlobals, invocation }: RunExportedFunctionArgs,
 ): Promise<ServedInvocationOutcome<unknown>> {
-  const execCtx = await ctx.createExecutionContext(nanoid());
+  const resolved = resolveInvocation({ kind: "fresh", options: invocation });
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 50979865446c3efe29743fb8fe02b20140f72cc1. `runExportedFunctionCore` now passes `inheritedRunId: getSubprocessRunInfo().runId` to `resolveInvocation`, matching `runNodeCore`, so a served function executed in subprocess mode joins the parent's trace instead of minting a new run id.

packages/agency-lang/lib/runtime/state/context.ts
```diff
@@ -369,14 +371,53 @@ export class RuntimeContext<T> {
...
+    metadata?: StatelogConfig["metadata"];
+    logFile?: string;
+  } {
+    const config = this.statelogConfig;
     return {
-      observability: this.statelogConfig?.observability ?? false,
-      ...(this.statelogConfig?.logFile ? { logFile: this.statelogConfig.logFile } : {}),
+      observability: config?.observability ?? false,
+      ...(config?.host ? { host: config.host } : {}),
+      ...(config?.apiKey ? { apiKey: config.apiKey } : {}),
```
egonSchiele: **High — preserve explicit empty-string overrides.** The invocation allow-list and config merge distinguish `undefined` from `""`, but these truthiness checks drop an effective empty `apiKey` (and likewise `host`/`projectId`) before spawning a child. In particular, `apiKey: ""` intentionally disables the parent's remote sink, while the child then falls back to its baked non-empty credential and can emit telemetry unexpectedly. Please use `!== undefined` here, matching `withParentStatelog`, and cover clearing a baked API key with an empty per-invocation override.

egonSchiele (commented):
The two original runtime issues are otherwise addressed, but I found one credential-isolation edge case in the new sink projection and left it inline. CI is also still red in unit-test shards 1, 3, and 4 for the same stale generated TypeScript fixtures; `make fixtures` output still needs to be committed.

packages/agency-lang/lib/runtime/state/context.ts
egonSchiele: Fixed in 0a9e0cd0a877af9d8066f0e484eaaf4e20f20deb. `getStatelogSink()` now uses `!== undefined` for `host`/`apiKey`/`projectId`/`metadata`/`logFile` (matching `withParentStatelog`), so an explicit empty string survives into the sink. An `apiKey: ""` per-invocation override now propagates to a child instead of the child falling back to its baked non-empty credential.

Test added in `context.perInvocationOverride.test.ts`: an execCtx built with `log: { apiKey: "" }` yields `getStatelogSink().apiKey === ""` (not the frozen `base-key`).

egonSchiele (commented):
Re-reviewed the latest two commits. The empty-string credential isolation issue is fixed correctly with `!== undefined`, the regression test covers clearing the baked API key, and the regenerated fixtures have cleared all four unit-test shards. I found no new blocking issues in this update. Integration/coverage jobs are still running at the time of review.

PR #826, opened 2026-08-09
agency schedule --backend remote: manage hosted schedules on statelog

egonSchiele:
Adds a server-authoritative `remote` backend to `agency schedule`, so `add`/`list`/`remove`/`edit` manage schedules on a hosted statelog server (spec: hosted-agent-schedules design §6, PR 4; plan: `2026-08-08-agency-schedule-remote-backend.md` rev 4).

## What it does

```
agency schedule add daily.agency --backend remote --node refresh --every daily
agency schedule list --backend remote
agency schedule edit <id> --backend remote --disabled
agency schedule remove <id> --backend remote
```

- **Server-authoritative:** no local registry entry; `list`/`remove`/`edit` query the server, and `remove`/`edit` take the server schedule id shown by `list`. A test mocks `Registry` to throw if any remote recipe ever constructs it.
- **Target resolution** reuses `resolveProjectTarget` (the `agency remote` seam): `--host` → `log.host` → the `remote.serveUrl` binding, project from `--project` or the binding with an origin-match guard, API key from env only.
- **Deploy-if-missing:** `add` checks the server's sources via `pullSource` (exact basename) and deploys through `runDeploy` when absent; `--redeploy` forces, `--no-deploy` skips. `runDeploy` now returns `"deployed" | "aborted" | "preview"`, and schedule creation is gated on `"deployed"` — declining the no-exports confirmation can no longer schedule an agent that was never uploaded. The deploy receives the already-resolved host/project/key-env, so it cannot target a different project than the schedule POST.
- **Sealed client:** `lib/cli/statelog/schedulesClient.ts` (modeled on `projectClient.ts`) owns the routes, wire validation, and the statelog envelope — including failures arriving inside HTTP 200, which are treated as failures regardless of status.
- **Pure resolvers:** `resolveScheduleAdd`/`resolveSchedulePatch` reject every invalid flag combination (`--node` xor `--function`, `--every` xor `--cron`, `--enabled` vs `--disabled`, `--redeploy` vs `--no-deploy`, empty edits) before any network work. Timezone defaults to the machine's IANA zone.
- **Loud flag applicability:** remote-only flags (`--node`, `--timezone`, `--host`, …) on the local/github paths are an error, not a silent no-op.

## Structure

- `lib/cli/statelog/schedulesClient.ts` — sealed HTTP client (+33 tests: transport, encoding, DTO validation table, envelope/failure matrix)
- `lib/cli/schedule/remote.ts` — pure resolvers + four thin recipes (+59 tests incl. the deployment-policy matrix with ordering assertions)
- `lib/cli/schedule/remote.integration.test.ts` — one vertical slice: real `addRemote` → real target resolution → real client → stubbed fetch
- `lib/cli/remote/commands/deploy.ts` — `runDeploy` outcome contract (existing `agency remote deploy` caller unchanged)
- `scripts/agency.ts` — registration, backend validation, applicability checks, awaited async dispatch (+16 CLI tests via real `parseAsync`, incl. a deferred-promise await proof and `ls`/`rm` aliases)
- `docs/dev/schedule-remote-backend.md` + CLAUDE.md pointer

## Verification

- Focused suites: 158 passed (schedulesClient 33, deploy 8, remote 59, vertical 2, agency CLI 56)
- `pnpm run lint:structure`, `pnpm run typecheck`, `pnpm run build` all clean
- Diff audited against `docs/dev/anti-patterns.md`; no `docs/site/**` changes

## Known limitations (deliberate)

- `--function` targets surface the server's rejection until statelog PR 2 (execution) lands.
- Edit can only change cadence/timezone/enabled — target, name, and args are immutable server-side (delete and re-add).


packages/agency-lang/lib/cli/schedule/remote.ts
```diff
@@ -0,0 +1,330 @@
...
+  }
+  if (error.status === 401 || error.status === 403) {
+    fail(
+      `${error.message}\nSchedule management needs a full-access API key; invoke-only keys are rejected.`,
+    );
+  }
+  if (guidance.missingAgent && /^Agent '.*' not found/.test(error.message)) {
+    fail(`${error.message}\n${guidance.missingAgent}`);
+  }
+  if (guidance.notFoundId && /not found/i.test(error.message)) {
```
egonSchiele: This condition is broader than the schedule-not-found contract. For example, a `ScheduleRequestError("Project not found", 404)` from this same route is rewritten as `No schedule with id ...`, hiding the actual bad project/binding. Please restrict the rewrite to the exact server error (`error.message === "Schedule not found"`) and add a regression test that a project 404 passes through unchanged.

packages/agency-lang/lib/cli/schedule/remote.ts
```diff
@@ -0,0 +1,330 @@
...
+  } catch (error) {
+    fail(errorMessage(error));
+  }
+  const target = resolveProjectTarget(context, options);
+  await ensureDeployed(file, resolved.input.fileName, resolved.deployMode, target, context);
+  const client = createSchedulesClient(target.origin, target.projectSlug, target.apiKey);
+
+  let schedule;
+  try {
+    schedule = await client.create(resolved.input);
```
egonSchiele, line 121: The suggested standalone deploy command does not preserve this schedule request’s resolved target. `agency remote deploy` uses the older deploy resolver, so it may ignore the linked binding or require different `log.*` config; it also drops explicit `--host`, `--project`, and `--api-key-env`. Following this hint can fail or upload to a different project. Safer guidance is to rerun the same schedule command without `--no-deploy` (or with `--redeploy`), which keeps the original target options.

packages/agency-lang/scripts/agency.ts
```diff
@@ -1642,19 +1684,36 @@ export function createProgram(deps: CliDependencies = {}): Command {
...
-        // than silently routing to auto-detect.
-        if (opts.backend !== undefined && opts.backend !== "github") {
-          console.error(
-            color.red(
-              `Unknown --backend value: "${opts.backend}". The only value accepted today is "github". Local backends (launchd, systemd, crontab) are auto-detected.`,
-            ),
-          );
-          process.exit(1);
+        rejectUnknownScheduleBackend(opts.backend, ["github", "remote"]);
+        if (opts.backend === "remote") {
```
egonSchiele, line 1713: Could this branch reject flags that the remote backend cannot honor before dispatching? `add` also accepts `--env-file`, `--secret`, `--write`, and `--no-pin`, but all four are silently discarded here; `edit --backend remote --env-file ...` has the same problem. The command can therefore report success even though requested environment/secrets/permissions were not applied. Please reject those flags on the remote path (and add dispatch tests proving the remote recipe is not called), just as the inverse remote-only/local check does below.

egonSchiele:
All four review comments addressed in 17579fc40:

1. **Not-found guidance too broad** (remote.ts:193/195): the id-specific rewrite now fires only on the exact server message `Schedule not found` (pinned to `[scheduleId].ts`'s `NOT_FOUND` constant). A `Project not found` 404 passes through unchanged — regression tests added for both remove and edit.

2. **Deploy hint dropped the resolved target** (remote.ts:121): the missing-agent hint now says to rerun the same command without `--no-deploy` (or with `--redeploy`), keeping `--host`/`--project`/`--api-key-env` and the binding-based resolution. A test asserts `agency remote deploy` is no longer suggested.

3. **Silently discarded flags on the remote path** (agency.ts:1701): `--env-file`, `--secret`, `--write`, and `--no-pin` are now rejected with `--backend remote` on add, and `--env-file` on remote edit, mirroring the existing remote-only check in the other direction. Five new dispatch tests prove the remote recipes are never called.

Focused suites (124 tests), lint, and typecheck all green.

egonSchiele:
Re-review complete on `17579fc`: all three prior findings are addressed. Remote mode now rejects unsupported local/GitHub flags before dispatch, schedule-id guidance matches only the exact `Schedule not found` response (with project-404 regressions), and missing-agent guidance preserves the original schedule target by directing users back to the same command. No new actionable findings. Updated unit tests, typecheck, lint, builds, Agency tests, and perf checks are passing; the integration check is still pending.

PR #830, opened 2026-08-09
Cloud speech (STT/TTS) via smoltalk: transcribe + speak + audio-in-chat

egonSchiele:
## Cloud speech (STT / TTS) via smoltalk — transcribe + speak + audio-in-chat

Routes `std::speech`'s cloud capabilities through the active `LLMClient` exactly
like `std::image` routes through `ctx.llmClient.image`, so they inherit Agency's
cost accounting, spend guards, and statelog. Design doc + review are in the diff
(`2026-08-09-agency-speech-via-smoltalk-plan.md`); code map in
`docs/dev/speech-via-smoltalk.md`. Depends on **smoltalk ^0.10.1** (audio ops +
`abortSignal`).

### What's here

- **`LLMClient`**: optional `transcribe?`/`speak?` taking the branch abort signal
  as their sole cancellation channel; derived `TranscribeConfig`/`SpeakConfig`
  (required model/voice/format, `abortSignal` excluded — one channel only); thin
  `SmoltalkClient` delegation.
- **`std::speech`**: local playback renamed `speak` → `say` (effect `std::say`);
  cloud TTS takes the `speak` name (new effect `std::synthesizeSpeech`);
  `transcribe` re-backed onto the client (effect `std::transcribe`,
  provider-honest). No back-compat shim: with no users, the old `_speak` /
  `__internal_*` runtime exports were removed rather than kept as aliases.
- **Cancellation**: the branch signal is passed as smoltalk's `abortSignal`, so a
  Ctrl-C / `race()` loss / time-guard trip tears down the in-flight request.
  smoltalk reports cancellation as a resolved `failure("Request was aborted")`;
  the `LLMClient` contract rejects on abort, so `SmoltalkClient.rejectIfAborted`
  converts an aborted outcome into a rejection carrying the branch `signal.reason`
  (detected via our own signal, not by string-matching), and `meteredDispatch`
  records exactly one unresolved attempt.
- **Runtime helpers**: `_transcribe`/`_synthesizeSpeech` mirror `_generateImage`;
  conservative #809 failure accounting; atomic no-clobber TTS publication
  (`publishSpeechOutput` — exclusive stage + `link()` commit, owned-stage cleanup;
  a cancelled synthesis never writes its file).
- **Accounting**: `transcription`/`speech` usage kinds through `ProviderUsageKind`,
  `USAGE_KINDS`, and the strict spend schema. Audio tokens collapse into
  `totalTokens` (conservative max lower-bound fallback + `attributionLost`).
- **Statelog**: `transcription` + `speechSynthesis` leaf events (capped previews;
  never audio bytes / `raw` / `pcm`).
- **Audio-in-chat**: `std::thread.audio()`; `Attachment` (image|file, gates
  `attachToReply`) vs `MessageAttachment` (adds audio, for `userMessage`/`llm`)
  split; `llm()` typechecker arg widened.
- **DeterministicClient**: fixed `transcribe`/`speak` with abort + configurable
  delays, for offline tests.

### Tests

Focused, mutation-sensitive: `lib/stdlib/speech.test.ts` (wrapper success/failure
accounting, unsupported-client, missing-file + already-aborted preflight,
mid-flight-abort → one unresolved attempt, guard trip writes no file, MIME
mismatch, atomic no-clobber publication), `lib/runtime/llmClient.audio.test.ts`
(adapter forwards source/config/`abortSignal`; abort → reject with the branch
reason; non-abort failure passthrough), `deterministicClient.test.ts` (fixed
bytes/cost + mid-request abort), `invocationUsage.test.ts` (new kinds + cost-only
speech + audio-token collapse), `spendTypes.test.ts` (new kinds accepted, unknown
rejected), `attachments.test.ts` (audio in `llm`/`userMessage`, rejected in
`attachToReply`). Full `typecheck` (all three configs), `make`, and
`lint:structure` clean.

### Out of scope (separate cross-repo change)

The statelog companion repo `usage_events.kind` server/DB forward migration for
the two new kinds is the plan's emission gate — a separate change that must deploy
before an Agency release emits `transcription`/`speech`. Not in this PR.


packages/agency-lang/lib/runtime/deterministicClient.ts
```diff
@@ -2,15 +2,44 @@ import type { PromptResult, StreamChunk, Result } from "smoltalk";
...
+  DETERMINISTIC_IMAGE_COST,
+  DETERMINISTIC_SPEECH_BYTES,
+  DETERMINISTIC_SPEECH_COST,
+  DETERMINISTIC_TRANSCRIBE_COST,
+  DETERMINISTIC_TRANSCRIPT,
+} from "../constants.js";
+
+/** Format → returned MIME, matching smoltalk's `SPEECH_FORMAT_TO_MIME`, so the
+ *  deterministic speak() returns a MIME the std::speech wrapper accepts. */
+const DETERMINISTIC_SPEECH_MIME: Record<string, string> = {
```
egonSchiele: I could swear we already have this mime mapping defined somewhere else in this codebase. Is this duplication?

packages/agency-lang/lib/stdlib/speech.ts
```diff
@@ -74,7 +86,14 @@ export async function __internal_speak(
   return speakImpl(ctx, stack, text, voice, rate, outputFile);
 }
 
-/** ALS-reading replacement for `__internal_speak`. */
+/**
+ * DEPRECATED local-playback alias. `_speak` used to back `std::speech.speak`,
```
egonSchiele: There is no need for any sort of deprecation or backwards compatibility. All you're doing is leaving the code base littered with code that is totally unnecessary. Please tidy up after yourself by removing anything that is not needed.

packages/agency-lang/lib/stdlib/speech.ts
```diff
@@ -209,78 +241,292 @@ export async function _record(
...
+// provider mechanics, loading, size caps, and MIME.
+//
+// Both helpers THROW on failure (the std::speech / std::fs idiom) — a smoltalk
+// failure Result becomes a thrown Error with its already-redacted message.
+
+/** Requested audio output format → expected returned MIME. Mirrors smoltalk's
+ *  `SPEECH_FORMAT_TO_MIME` so the post-dispatch MIME check agrees with what
+ *  `speak()` returns. Kept local to avoid importing a smoltalk internal path. */
+const SPEECH_FORMAT_TO_MIME: Record<string, string> = {
+  mp3: "audio/mpeg",
```
egonSchiele: duplication duplication

packages/agency-lang/lib/stdlib/speech.ts
```diff
@@ -209,78 +241,292 @@ export async function _record(
...
-      headers: {
-        "Authorization": `Bearer ${apiKey}`,
-      },
-      body: formData,
-      signal,
-    });
+  const source: AudioInput = { kind: "path", path: resolvedPath };
+  const config: TranscribeConfig = {
+    model,
+    ...(provider ? { provider } : {}),
```
egonSchiele: This is such ugly code. Please do not write ternary statements like this. Instead build up the object from multiple if statements.

packages/agency-lang/lib/runtime/invocationUsage.ts
```diff
@@ -203,18 +215,34 @@ function buildTokens(raw: unknown, kind: UsageKind, absentDegrades: boolean): {
   };
   let malformed = input.malformed || output.malformed || cachedInput.malformed || cacheCreation.malformed;
 
+  // Audio tokens (gpt-audio-1.5 chat) are collapsed into `totalTokens` and never
```
egonSchiele, line 236: The audio-token fallback is currently confined to `normalizeObservation`, but the same raw `completion.usage` still feeds `promptCompletion`, global token stats, and `recordCompletionUsage`/`getTokens()`. With `{inputTokens: 3, outputTokens: 2, inputAudioTokens: 40, outputAudioTokens: 4}` and no `totalTokens`, the invocation meter records 44 while branch/global totals add 0, and statelog serializes the undeclared audio fields. Please expose one closed usage projection (plus `attributionLost`) and apply it before every invocation/branch/global/statelog consumer; add an end-to-end audio-completion test rather than testing only this normalizer.

packages/agency-lang/lib/stdlib/speech.ts
```diff
@@ -209,78 +241,292 @@ export async function _record(
...
+ * Backs `std::speech.transcribe` (speech-to-text). Resolves + authorizes the
+ * audio path (Agency's allow-list) and verifies it is a readable regular file
+ * BEFORE any paid dispatch — preserving today's local missing-file failure and
+ * keeping it off the metered path. Then routes through `ctx.llmClient.transcribe`
+ * with cost/guards/statelog. Throws on failure.
  */
-async function transcribeImpl(
-  ctx: RuntimeContext<any>,
-  stack: StateStack,
+export async function _transcribe(
```
egonSchiele, line 294: This expands the exported runtime ABI from `_transcribe(filepath, language, allowedPaths)` to eight required arguments. Existing compiled stdlib artifacts still call the old three-argument form, so `model` becomes `undefined` and is forwarded as a required config value. This patch also removes the previously exported `__internal_transcribe`, causing older pre-ALS artifacts to fail module instantiation. Please default the new trailing arguments and restore a compatibility wrapper for `__internal_transcribe`; frozen fixtures for both prior call shapes would pin this boundary.

packages/agency-lang/lib/runtime/llmClient.ts
```diff
@@ -218,6 +299,17 @@ export class SmoltalkClient implements LLMClient {
 
 }
 
+/** If the branch signal has aborted, throw its reason so cancellation surfaces
+ *  as the runtime's abort cause (an AgencyCancelledError) rather than smoltalk's
+ *  resolved `failure("Request was aborted")`. See the transcribe/speak adapters. */
+function rejectIfAborted(signal: AbortSignal): void {
+  if (signal.aborted) {
+    throw signal.reason instanceof Error
```
egonSchiele: The `LLMClient` contract above promises rejection with the exact `signal.reason`, but this replaces any non-`Error` reason with a new `Error`. `AbortController.abort()` accepts strings, objects, and `null`, and identity can matter to race/cancellation handling. Please throw the defined reason unchanged (falling back only when it is genuinely `undefined`) and make the same change in `speech.ts`'s `abortReasonError`; tests should include a sentinel object/string, not only `AgencyCancelledError`.

packages/agency-lang/stdlib/speech.agency
```diff
@@ -71,19 +100,104 @@ export def record(outputFile: string = "", silenceTimeout: number = 2000, allowe
...
+/** A cloud synthesis request tears down on Ctrl-C, race-loser, or time-guard
+abort; a cancelled request never writes its output file. Cost, spend guards,
+and statelog apply. */
+export def speak(
+  text: string,
+  outputFile: string = "",
+  voice: string = "alloy",
+  model: string = "tts-1",
+  provider: string = "",
+  format: string = "mp3",
```
egonSchiele, line 162: The public API documents a six-value format set and speed range, but only model/voice are validated. Runtime later force-casts these values and skips MIME validation when the format is unknown, so a custom/deterministic client can publish a `.bogus` artifact; invalid speed can also become a metered failure before any provider request. Please validate format and finite speed `[0.25, 4]` before the interrupt (and similarly restrict transcription granularity to `"" | "segment" | "word"`), with tests asserting zero interrupt/client calls for invalid values.

packages/agency-lang/lib/stdlib/speech.ts
```diff
@@ -209,78 +241,292 @@ export async function _record(
...
+  signal: AbortSignal,
+): Promise<void> {
+  const dir = path.dirname(finalPath);
+  const stage = path.join(
+    dir,
+    `.${path.basename(finalPath)}.agency-tts-${nanoid()}.part`,
+  );
+  let staged = false;
+  try {
+    await writeFile(stage, audio, { flag: "wx" }); // exclusive create of the stage
```
egonSchiele: `writeFile({flag: "wx"})` can create the stage and then reject during the write (for example ENOSPC). Because `staged` flips only after the promise resolves, that partial owned stage bypasses cleanup. Please use `open(stage, "wx")`, mark ownership after `open` succeeds, and write through/close that handle before commit. Setting the flag before `writeFile` is unsafe because an `EEXIST` collision must not delete an unowned path. Add a create-then-write-failure test.

packages/agency-lang/lib/stdlib/speech.ts
```diff
@@ -209,78 +241,292 @@ export async function _record(
...
+  }
+
+  await publishSpeechOutput(finalPath, speech.audio, signal);
+  return finalPath;
+}
+
+/** True when `p` exists (any type). A missing path is the normal case. */
+async function pathExists(p: string): Promise<boolean> {
+  try {
+    await stat(p);
```
egonSchiele: This catches every `stat` failure as “missing”, so EACCES/ELOOP/I/O errors allow paid synthesis to proceed before publication predictably fails. `stat` also follows symlinks, meaning a dangling symlink passes as ENOENT and then causes post-dispatch `link(..., finalPath)` to fail EEXIST. Please use `lstat`, return false only for ENOENT, and rethrow other errors; add non-ENOENT and dangling-symlink preflight tests asserting zero dispatch.

packages/agency-lang/lib/stdlib/speech.ts
```diff
@@ -209,78 +241,292 @@ export async function _record(
...
   const resolvedPath = await resolveDir(filepath, allowedPaths ?? []);
-  const fileData = await readFile(resolvedPath);
-  const filename = path.basename(resolvedPath);
-
-  const formData = new FormData();
-  formData.append("file", new Blob([fileData]), filename);
-  formData.append("model", "whisper-1");
-  if (language !== "") {
-    formData.append("language", language);
+  const info = await stat(resolvedPath); // throws ENOENT for a missing file
```
egonSchiele: The comment promises a readable regular-file preflight, but `stat`/`isFile` does not test readability. An unreadable file reaches `meteredDispatch`, then fails while smoltalk loads it before any provider request, incorrectly recording an unresolved paid attempt. Please call `access(resolvedPath, constants.R_OK)` before dispatch and test that unreadable input yields zero client calls/attempts.

packages/agency-lang/lib/runtime/deterministicClient.ts
```diff
@@ -286,4 +320,57 @@ export class DeterministicClient implements LLMClient {
...
+  // Fixed transcript + fixed cost so tests can exercise the STT pipeline (cost,
+  // guards, statelog) offline. Honors the abort signal via the optional delay so
+  // mid-request cancellation is testable. Never reports a `model` field — real
+  // TranscriptionResult has none, and the wrapper owns the effective model.
+  async transcribe(
+    _source: AudioInput,
+    _config: TranscribeConfig,
+    signal: AbortSignal,
+  ): Promise<Result<TranscriptionResult>> {
+    if (this.speechOptions.transcriptionDelayMs) {
```
egonSchiele, line 342: With the default zero delay this method never consults `signal`; `speak` has the same issue. A pre-aborted direct client call therefore returns success, contrary to the `LLMClient` cancellation contract and potentially masking races between wrapper preflight and invocation. Please perform an unconditional entry abort check before the optional delay and test both zero-delay methods with the exact abort reason.

egonSchiele (commented):
Review found eight actionable correctness/compatibility gaps. The highest-impact items are applying audio-token projection consistently across every consumer, preserving the exported transcription ABI for compiled artifacts, and retaining exact AbortSignal reason identity.

packages/agency-lang/lib/runtime/llmClient.ts
```diff
@@ -218,6 +301,17 @@ export class SmoltalkClient implements LLMClient {
...
 }
 
+/** If the branch signal has aborted, throw its reason UNCHANGED so cancellation
+ *  surfaces as the runtime's abort cause rather than smoltalk's resolved
+ *  `failure("Request was aborted")`. Identity is preserved — `abort()` accepts
+ *  strings/objects/null and that can matter to race/cancellation handling — and
+ *  a reason is only synthesized when genuinely absent. See the adapters. */
+function rejectIfAborted(signal: AbortSignal): void {
+  if (signal.aborted) {
+    throw signal.reason ?? new Error("Request was aborted");
```
egonSchiele: This still does not preserve an explicit `null` abort reason: `controller.abort(null)` is valid, but `reason ?? fallback` replaces it with a new error. The same nullish-coalescing issue remains in `speech.ts`'s `throwAbortReason` and all three deterministic abort paths. Please test `signal.reason !== undefined` instead, and add a `controller.abort(null)` identity test. (Other falsy values such as `0`, `false`, and `""` are already preserved by `??`.)

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -732,14 +733,19 @@ async function _runPrompt({
...
   const modelName = resolveCompletionModel(completion.model, clientConfig.model);
 
+  // Project raw provider usage into the closed shape ONCE, so statelog, the
+  // global stats, the branch total, and the invocation meter all see the same
+  // total and none serialize an audio-token field (see projectProviderTokenUsage).
+  const projectedUsage = projectProviderTokenUsage(completion.usage, "completion").usage;
+
   ctx.statelogClient.promptCompletion({
     messages: withMessageLabels(messages),
     completion,
```
egonSchiele: The top-level `usage` is projected now, but the raw `completion` still contains `completion.usage`, so `StatelogClient.promptCompletion` serializes the audio-token fields through this nested path. The memory-manager call site also sends both nested and top-level usage raw. Please pass `completion: { ...completion, usage: projectedUsage }` here, apply the projection in memory, and add a serialized-event assertion that `inputAudioTokens`/`outputAudioTokens` occur nowhere in the complete payload.

packages/agency-lang/lib/stdlib/speech.ts
```diff
@@ -209,78 +211,337 @@ export async function _record(
...
   const resolvedPath = await resolveDir(filepath, allowedPaths ?? []);
-  const fileData = await readFile(resolvedPath);
-  const filename = path.basename(resolvedPath);
-
-  const formData = new FormData();
-  formData.append("file", new Blob([fileData]), filename);
-  formData.append("model", "whisper-1");
-  if (language !== "") {
-    formData.append("language", language);
+  const info = await lstat(resolvedPath); // throws ENOENT for a missing file
```
egonSchiele: Using `lstat` here rejects a valid contained symlink to a readable regular audio file, even though `resolveDir`/`assertContained` already resolves symlinks for allow-list containment. A dangling input symlink would already fail pre-dispatch with `stat(2)`, and the comment's post-dispatch `link()` concern applies to TTS output, not STT input. Please use `stat` for STT while retaining `lstat` in TTS `pathExists`; tests should accept a symlink whose target remains inside `allowedPaths` and reject one resolving outside it.

packages/agency-lang/2026-08-09-agency-speech-via-smoltalk-plan.md
```diff
@@ -0,0 +1,898 @@
...
+  error. Rejected dispatches remain possible and flow through `meteredDispatch`.
+
+### 11a. Preserve the old local `_speak` ABI
+
+`_speak` and `__internal_speak` are exported runtime symbols that currently mean local macOS
+playback. Already-compiled stdlib artifacts can resolve those imports against a newer runtime.
+Changing `_speak` to cloud TTS would let an artifact approved only for old `std::speak` reach a
+cloud call without raising `std::synthesizeSpeech`.
+
+Therefore `_speak` and `__internal_speak` remain local-playback-compatible deprecated aliases.
```
egonSchiele: The implementation now deliberately removes `_speak` and the `__internal_*` aliases per the owner decision, but this checked-in plan still says preserving them and adding frozen ABI fixtures are fixed requirements; the PR body says the same. Since the dev note points readers here for the full rationale, please update the plan/PR description to record the actual decision and remove the frozen-fixture requirements. This is documentation drift only—no request to restore the aliases.

egonSchiele (commented):
Re-review of the follow-up commit: the original findings are mostly resolved. Four follow-ups remain: explicit null abort-reason identity, nested raw usage in statelog, valid STT symlink handling, and stale ABI-freeze documentation.

PR #831, opened 2026-08-09
agency remote secrets: set/list/rm/import for hosted environment secrets

egonSchiele:
`agency remote secrets set <NAME> / list / rm <NAME> / import [file]` against statelog's encrypted, write-only project-secrets store (statelog `6cdfa23`). Spec: `2026-08-09-remote-secrets-cli-design.md` rev 2; plan rev 3 executed task-by-task.

## The two invariants, enforced by construction

**A value never touches argv.** No `--value` flag exists (a dispatch test pins that it stays an unknown option). Values reach `set` via a hidden TTY prompt (`prompts` type invisible, static import), piped stdin (exactly one trailing `\n`/`\r\n` stripped), or `--from-env VAR` — all injected seams, no process globals in recipes.

**A value never reaches output — even when a hostile response echoes it back.** Three layers:
1. `readJsonBody` gains a diagnostic-only `sanitizeDiagnostic` seam that runs on RAW body text, read-exception messages, and displayed URLs BEFORE the whitespace collapse and 200-char truncation that would otherwise split a value past a redactor's reach. Protocol data is always parsed from the original text; existing callers are unchanged.
2. `secretsClient` redacts per verb (`[apiKey, value]` on set, `[apiKey]` on list/rm) across every failure class, and preserves the verified failure taxonomy: domain errors ride HTTP-200 envelopes (`Secret not found.`), the middleware 404 is matched on its `error` field, any other 404 means upgrade-the-host.
3. `presentSecretError` is the single composition point: redact known values, escape terminal-unsafe text (JSON-quoting control chars), then append the trusted one-line 401/403 guidance. Import passes all imported values in, so one request's failure cannot leak another name's value.

Tests enforce the negative space with a distinctive sentinel planted in every hostile-response class — non-JSON bodies (incl. whitespace-split and truncation-spanning variants), JSON errors, 200 envelopes, transport errors — asserting `[redacted]` present and the sentinel absent from all captured output. The vertical no-network test pins that the sentinel appears in the POST body and nowhere else.

## import

Bulk dotenv import: positional file (default `.env`), `-` for stdin; parsing is a pure wrapper over `node:util`'s `parseEnv` (the repo's `loadEnv` is unusable here — it mutates `process.env` and truncates values containing `=`). File sources on a TTY confirm with names only; `import -` never prompts (stdin is exhausted at EOF; choosing `-` is the authorization). The POST loop continues past failures and reports per-name outcomes; recipes return semantic results and the Commander actions are the sole `process.exitCode` owner (canceled set, declined/failed import → 1).

## Verification

181 tests across 9 touched suites (redact 5, jsonBody 14, secretsClient 22, secretsInput 21, confirmation 7, recipes 36, vertical 2, deploy regression 8, CLI dispatch 67). Lint, typecheck, build, and the repo-wide source-text guard all clean; diff audited against `docs/dev/anti-patterns.md`; `docs/dev/remote-secrets.md` + CLAUDE.md pointer added. No `docs/site` changes.


packages/agency-lang/lib/cli/remote/secretsInput.ts
```diff
@@ -0,0 +1,102 @@
...
+   *  duplicate key the LAST assignment's value. No duplicate reporting —
+   *  detecting duplicates would need a second, quote-aware dotenv parser (a
+   *  `B=` line inside a multiline quoted value is NOT an assignment), and
+   *  last-value-wins is the behavior that matters. */
+  entries: { name: string; value: string }[];
+};
+
+/** Pure wrapper over node:util's parseEnv. Never touches process.env. */
+export function parseEnvSource(text: string): ParsedEnvSource {
+  const parsed = parseEnv(text) as Record<string, string>;
```
egonSchiele: This does not work on the repository's minimum supported Node version. The PR's `unit-tests (4)` job runs Node 22.13.0 and `parseEnv('B="x=y"')` produces an extra `{ name: '"x', value: 'y"' }` entry, so the new test at `secretsInput.test.ts:103` fails. Since `package.json` permits `>=22.13.0`, this would also import a bogus secret for users on that version. Please either use a parser whose behavior is stable across the supported range or raise the minimum Node version to one where all of these dotenv cases pass (and update CI accordingly).

packages/agency-lang/lib/cli/remote/commands/secrets.ts
```diff
@@ -0,0 +1,235 @@
...
+    for (const entry of entries) {
+      console.log(`  ${terminalSafe(entry.name)}`);
+    }
+    if (!(await io.confirm("Continue?"))) {
+      console.log("Canceled.");
+      return { kind: "declined" };
+    }
+  }
+
+  const allValues = entries.map((entry) => entry.value);
```
egonSchiele, line 161: The cross-name defense can still leak when values overlap because `client.set()` has already destructively redacted only the current value before `presentSecretError` sees the error. For example, import `LONG=prefix-SENSITIVE-SUFFIX` and `SHORT=prefix`, then let the response to `SHORT` echo the `LONG` value: the client turns it into `[redacted]-SENSITIVE-SUFFIX`, after which the batch pass can no longer match `LONG`, so the suffix is printed. The no-value-in-output invariant requires all imported values to participate in the first sanitization pass (with longer values replaced before their prefixes). Please add this overlapping-values regression case as well.

packages/agency-lang/lib/cli/remote/secretsInput.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/** Single-line, terminal-safe rendering for untrusted display text (names,
+ *  server messages): a string containing control characters — including ANSI
+ *  escapes — is JSON-quoted so it cannot manipulate the terminal; plain
+ *  strings pass through unchanged. Always applied AFTER redaction. */
+export function terminalSafe(text: string): string {
+  for (const character of text) {
+    const code = character.codePointAt(0) ?? 0;
+    if (code < 0x20 || code === 0x7f) {
```
egonSchiele: This misses C1 terminal controls such as U+009B (CSI), and `JSON.stringify()` does not escape DEL/C1 characters even when this branch detects U+007F. Thus `terminalSafe('BAD\u009b31mNAME')` is returned unchanged, while `terminalSafe('A\u007fB')` still contains the raw control character, contrary to the terminal-safe contract for server text and names. Please cover C1 controls and explicitly encode ranges that JSON string escaping leaves raw (also consider U+2028/U+2029 for the single-line guarantee).

packages/agency-lang/lib/cli/statelog/secretsClient.ts
```diff
@@ -0,0 +1,172 @@
...
+        response.status,
+      );
+    }
+    return envelope.value;
+  }
+
+  function parseWire<T>(schema: z.ZodType<T>, value: unknown, sensitive: string[]): T {
+    const result = schema.safeParse(value);
+    if (!result.success) {
+      throw new SecretRequestError(
```
egonSchiele, line 136: Malformed successful responses lose their HTTP status here. `request()` knows the response was HTTP 200 but returns only `envelope.value`, so this `SecretRequestError` has `status === undefined`. The client contract says every HTTP-response failure preserves its status (and presentation policy relies on status). Please carry the status alongside the value into `parseWire()` and pass it to the error; the existing malformed-DTO test can assert `status === 200`.

egonSchiele (commented):
I found four actionable issues: one currently breaks the Node 22.13 CI job, two weaken the no-output/terminal-safety guarantees, and one drops the documented HTTP status contract.

egonSchiele:
All seven review comments addressed in d9b8cdbfa; the failing `unit-tests (4)` job is the parseEnv finding and should now be green.

1. **parseEnv version-dependence (the CI failure)**: `parseEnvSource` is now an owned ~40-line dotenv parser — quoted/multiline-aware, CRLF, `export` prefix, last-wins duplicates — with semantics pinned by the existing tests. `node:util`'s parseEnv mis-splits `B="x=y"` on 22.13.0, and secret values must parse identically across the whole supported engine range, so no version of it is used.
2. **Overlapping-values leak**: every import `set()` passes the whole batch as `alsoRedact`, so all values join the FIRST redaction pass, and `redactValues` now replaces longer values before their prefixes. Regression tests cover both the LONG/SHORT ordering and the per-call `alsoRedact` forwarding.
3. **terminalSafe C1/DEL/U+2028**: unsafe now means C0, DEL, C1 (incl. bare CSI U+009B), and U+2028/U+2029; the JSON-quoted form is post-processed so every such character becomes a \\uXXXX escape. Tests assert each of the three characters JSON.stringify leaves raw.
4. **Malformed-DTO status**: `request()` returns `{ value, status }` and `parseWire` carries it onto `SecretRequestError`; the malformed-DTO test asserts `status === 200`.
5. **`env()` hint**: renders the name as a real string literal via `JSON.stringify(name)` — correct syntax for plain names, no double-quoting for hostile ones. Help text now shows `env("NAME")`.
6. **Import source names** render through `terminalSafe` at all three display sites.

495 tests across the statelog/remote/CLI suites, lint and typecheck clean.

PR #832, opened 2026-08-09
One transport core under the six sealed statelog clients

egonSchiele:
Pure refactor per `2026-08-09-statelog-transport-core-design.md` (rev 2) and its plan (rev 3): the six sealed statelog clients now share ONE transport implementation, `statelogRequest`, and each client file is reduced to what is genuinely its own — routes, zod wire schemas, and an exhaustive failure mapper transcribed from its previous branches.

## The contract, made provable

**Zero existing test assertions changed** — `git diff origin/main --numstat` shows only additions in every client test file. The 200+ exact-message assertions across the six suites are the regression net, plus two new layers added before any client moved:

- **Characterization tests** pinning what the old suites left open: upload never inspects `response.ok` (an HTTP 500 carrying a valid success envelope still succeeds), serve sends `Content-Type` on bodyless GETs and serializes `{}` for undefined POST bodies, and the secrets unreachable message redacts as a whole (a sensitive value in the ORIGIN redacts, not only one in the cause). All were green against the unmodified clients on first run.
- **The core's precedence matrix**: non-2xx wins over non-json and envelope handling; a non-2xx `{success:false}` is `http`, not `envelope-error`; `requireOk:false` (upload) lets the envelope decide at any status; `contentType:"always"` (serve) preserves the wire; programmer errors (unserializable body, throwing sanitizer) propagate rather than becoming failure kinds.

## Shape

`statelogRequest` is result-returning, formats no user-facing English, builds no URLs, and is internal to `lib/cli/statelog/` — commands keep using the family factories. Every mapper is an exhaustive `switch (failure.kind)` (a future kind fails compilation everywhere, no fallback arm), receives only the failure plus minimal family context (never a Response or raw envelope), and the secrets mapper owns redaction of every string a failure can carry. Migration went one client per commit: schedules, secrets, project, account, serve, upload.

The structural boundary is now grep-enforceable: no `fetch(` or `response.json()` exists in the family outside `statelogRequest.ts`/`jsonBody.ts` (verified; documented as a review tripwire in the rewritten `docs/dev/statelog-clients.md`).

## Verification

All 11 statelog suites: 223 passed. Lint + typecheck clean. Net: ~340 lines of duplicated per-client transport deleted. Non-goals per spec: no aggregate facade, no drift harmonization (e.g. projectClient errors still carry no status — changing that is a deliberate follow-up, not a refactor side effect).


packages/agency-lang/lib/cli/statelog/statelogRequest.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export async function statelogRequest(
+  options: StatelogRequestOptions,
+): Promise<StatelogRequestResult> {
+  const envelope = options.envelope ?? true;
+  const requireOk = options.requireOk ?? true;
+  // Serialize before the fetch try: an unserializable body is a programmer
+  // error and must propagate, never classify as `unreachable`.
+  const serializedBody = options.body === undefined ? undefined : JSON.stringify(options.body);
+  const sendContentType = options.contentType === "always" || serializedBody !== undefined;
```
egonSchiele: **Blocking:** this changes when `Content-Type` is sent. Every old client decided from `body !== undefined` before serialization; this decides from `JSON.stringify(body) !== undefined`. JSON.stringify legitimately returns `undefined` for a function, symbol, or an object whose `toJSON()` returns `undefined`, so a provided body now loses the header. This is reachable through structurally compatible schedule input with an enumerable `toJSON`. Preserve the old rule with `options.contentType === "always" || options.body !== undefined`, and add a core test using `{ toJSON: () => undefined }` that expects the header with an undefined fetch body.

packages/agency-lang/lib/cli/statelog/statelogRequest.ts
```diff
@@ -0,0 +1,127 @@
...
+  | { ok: false; failure: StatelogFailure };
+
+export async function statelogRequest(
+  options: StatelogRequestOptions,
+): Promise<StatelogRequestResult> {
+  const envelope = options.envelope ?? true;
+  const requireOk = options.requireOk ?? true;
+  // Serialize before the fetch try: an unserializable body is a programmer
+  // error and must propagate, never classify as `unreachable`.
+  const serializedBody = options.body === undefined ? undefined : JSON.stringify(options.body);
```
egonSchiele, line 55: **Blocking:** the global programmer-error rule is not behavior-preserving for `serveClient`. On main, Serve evaluates `JSON.stringify(body ?? {})` inside its fetch `try`, so BigInt, circular values, and throwing `toJSON` methods become `ServeRequestError: could not reach <full URL> (<serialization error>)`. Serialization now happens before the core try, so Serve exposes a native exception instead. Add a Serve characterization test first, then preserve its existing mapping with a narrow Serve-only option/equivalent while leaving the other clients' serialization errors propagating.

packages/agency-lang/lib/cli/statelog/statelogRequest.ts
```diff
@@ -0,0 +1,127 @@
...
+    response,
+    { method: options.method, url: options.url },
+    { sanitizeDiagnostic: options.sanitizeDiagnostic },
+  );
+
+  // Non-2xx wins over non-json and envelope handling (the families' pinned
+  // precedence): extract a string server error only when parsing succeeded —
+  // from a bare { error } or an envelope's error field.
+  if (requireOk && !response.ok) {
+    const failure: StatelogFailure = { kind: "http", status: response.status };
```
egonSchiele: This new construction uses the order-dependent mutable-state pattern called out in `docs/dev/anti-patterns.md`: create a partial union member, conditionally mutate `serverError`, then return it. Please return the complete failure directly from the string branch and return the no-error variant afterward. The same pattern appears in the `envelope-error` branch below.

packages/agency-lang/lib/cli/statelog/statelogRequest.test.ts
```diff
@@ -0,0 +1,183 @@
...
+    method: "GET",
+    url: "https://h/api/x",
+    apiKey: "key",
+    ...overrides,
+  });
+}
+
+async function failureOf(overrides: Partial<StatelogRequestOptions> = {}) {
+  const result = await request(overrides);
+  if (result.ok) throw new Error("expected a failure");
```
egonSchiele: The new tests use one-line `if` statements, which `docs/dev/anti-patterns.md` explicitly prohibits. Please add braces here and to the new narrowing guards around lines 50 and 160, plus the two new `if (result.ok) return;` guards in `uploadClient.test.ts`.

egonSchiele (commented):
Two behavior regressions need correction before this is a pure refactor. I also left two smaller anti-pattern comments. CI is green, and the HTTP/non-JSON precedence, status propagation, and secrets redaction otherwise look preserved.

egonSchiele:
All four comments addressed in 03511ffff:

1. **Content-Type from the provided body (blocking)**: the header decision is now `options.contentType === "always" || options.body !== undefined` — matching every old client's pre-serialization rule — with a core test pinning that `{ toJSON: () => undefined }` keeps the header while the fetch body is `undefined`.

2. **Serve's serialization mapping (blocking)**: a new serve characterization test pins main's behavior (stringify ran inside serve's fetch try, so a BigInt body surfaced as `ServeRequestError: could not reach <url> (…)`). The serve wrapper now catches the core's propagated serialization exception and maps it to that same message — precise in practice, since serve passes no sanitizer and the core's only other pre-fetch throw source is serialization. The other five clients keep propagating programmer errors, unchanged from their behavior on main (they serialized outside their fetch try).

3. **Mutable failure construction**: both the `http` and `envelope-error` branches now return complete union members per branch; no partial-then-mutate.

4. **One-line ifs**: braced in `statelogRequest.test.ts` (including the `failureOf` helper) and `uploadClient.test.ts`.

All 11 statelog suites green (225 tests), lint + typecheck clean; existing client test files remain additions-only against main.

egonSchiele (commented):
Re-review of `03511ffff82d29d6cd42a403b9d3732736c8f597`: all four prior findings are addressed. Content-Type now follows whether a body was provided, Serve's historical serialization-error mapping is characterized and preserved, failure variants are constructed without mutation, and the new one-line `if` statements are braced. No new behavioral issue found in the follow-up diff.

One small cleanup remains: `git diff --check origin/main...HEAD` reports new blank lines at EOF in `projectClient.ts`, `schedulesClient.ts`, and `secretsClient.ts`. Please remove those before merge. The new CI run is still pending at the time of this re-review.

PR #833, opened 2026-08-09
std::aws/s3: S3 object access with no AWS SDK (SigV4 via node:crypto)

egonSchiele:
## What this adds

`std::aws/s3` — Amazon S3 access from Agency with **no AWS SDK dependency**. Five functions: `s3Get`, `s3GetBinary`, `s3Put`, `s3PutBinary`, `createBucket`. Requests are signed with SigV4 using only Node's built-in `crypto`.

```ts
import { s3Put, s3Get } from "std::aws/s3"

node main() {
  s3Put("my-bucket", "notes.txt", "hello") with approve
  print(s3Get("my-bucket", "notes.txt") with approve)
}
```

## Architecture

A functional, product-agnostic core (no classes — `fetch` is stateless, and per-run state may not live in module globals) with a thin S3 module on top:

- **uri / base64 / objectBytes** — shared `awsUriEncode`, strict base64 decoder, and one `AWS_OBJECT_BYTE_LIMIT` (10 MiB). `_writeBinary` reuses the decoder.
- **credentials / endpoints** — env-only credentials; a bounded region→partition policy (commercial, GovCloud, China, ISO) that rejects other prefixes; and full general-purpose bucket validation producing a `ValidatedBucket` (a raw bucket string never reaches endpoint construction).
- **sigv4 / client** — the atomic `AwsRequestTarget` (`{origin, canonicalUri}`); transport derives one wire URL, validates its hostname against the partition suffix before signing/fetching, and returns metadata + bytes + headers.
- **errors / s3** — coded S3 error mapping, and one `runS3Operation` switch that owns the whole imperative pipeline; the five extern helpers are one-call adapters.
- **stdlib/aws/s3.agency** — the surface. Reads are interrupt-gated; writes are interrupt-gated **and** `destructive`; interrupt payloads carry only `{bucket, key, region}`. Lock `bucket` with `.partial(...)` to scope an agent to one bucket.

## Safety

- **Untrusted region/bucket/key** are validated before any URL is built; the final request hostname is re-checked against the resolved partition suffix before `fetch`.
- **Exact key canonicalization** — the endpoint builder encodes each segment once and emits both the wire URL and the signed canonical URI from that one encoding (no re-derivation via URL normalization). Keys with a complete `.`/`..` segment are rejected; `soap` forces path-style.
- **Byte caps both directions** (download via a streaming abort-safe reader that rejects a pending read on abort rather than returning a prefix; upload before signing).
- **Binary stays out of statelog** — the redaction table now carries a custom marker used by *both* statelog paths (the `JSON.stringify` replacer and `runner.ts`'s `safeStatelogValue`), so a label cannot leak through an equality check. `_s3GetBinary` marks its base64 `[binary output truncated]`. `std::tag`'s `redact(value, label?)` exposes this to Agency. **Narrowed v1 guarantee** documented in `docs/dev/aws.md`: downloaded results + interrupt payloads are protected; `s3PutBinary` exposed *directly* as an LLM tool can log its argument (recorded before the helper runs).

## Testing

All offline and deterministic — no live AWS, no credentials, no destructive test:
- Unit tests for uri/base64/objectBytes, credentials/partitions/buckets, sigv4 (reproduces AWS's published test vector; hashes raw binary bytes), client (mocked fetch + real `runInTestContext`), errors, and the S3 executor (exact URLs across all four partitions, dotted/`soap` path-style, one-time key encoding, upload caps, ETag/location, binary round-trip + redaction marking, `.`/`..` rejection).
- The abort-safe HTTP reader and the two-path redaction change have their own tests.
- Offline Agency fixture proves a rejected write is skipped and an approved write enters the destructive region; empty `fetchMocks` denies all network so it is harmless even where creds exist.
- A checked-in strict-typecheck fixture pins the `redact` Agency wrapper.

190 unit tests, structural lint, and `sourceIsText` all green.

## Notes

This addresses all nine findings from the plan review (streaming cap, two-path redaction, region→host escape, offline fixtures, ETag/location, key canonicalization, strict base64 + upload caps, parseable `redact` syntax, narrowed statelog guarantee). Spec/plan/review docs are kept as untracked working files, matching the repo convention. No new npm dependency; `CHANGELOG.md` untouched.


packages/agency-lang/lib/stdlib/aws/s3.ts
```diff
@@ -0,0 +1,277 @@
...
+export type S3OperationResult =
+  | string
+  | S3PutResult
+  | S3CreateBucketResult
+  | ResultFailure;
+
+// A key must not contain a complete "." or ".." slash-delimited segment: v1
+// `fetch` normalizes those away, so the signed path and the fetched path would
+// diverge. Reject them before endpoint construction.
+function keyFailure(key: string): ResultFailure | null {
```
egonSchiele, line 85: **Blocker:** reject an empty key here. `keyFailure("")` currently succeeds, and the virtual-hosted target becomes `/`. That changes the approved operation: `s3Get(bucket, "")` sends S3's bucket-list request rather than `GetObject`, while `s3Put(bucket, "", "", "us-east-1")` sends `PUT /`, i.e. `CreateBucket`, under an object-write approval. For an already-owned us-east-1 bucket, AWS documents that create-bucket can return 200 and reset ACLs. Please reject `key.length === 0` before target construction and test all four object helpers assert that `fetch` is not called.

packages/agency-lang/lib/stdlib/aws/s3.ts
```diff
@@ -0,0 +1,277 @@
...
+      });
+      if ("error" in response) return response;
+      const errored = responseErrorOrNull(response);
+      if (errored) return errored;
+      if (operation.kind === "getText") {
+        return new TextDecoder("utf-8").decode(response.bytes);
+      }
+      const base64 = Buffer.from(response.bytes).toString("base64");
+      const { globals } = getRuntimeContext();
+      globals.markRedacted(base64, BINARY_MARKER);
```
egonSchiele, line 188: This redaction does not survive a default isolated `fork`/`race`. The helper marks this primitive in the branch's cloned `GlobalStore`, but `runInBranchAlsFrame` discards successful branch globals at join; parent-side `safeStatelogValue` then serializes the returned base64 in `forkBranchEnd` without finding its tag. That contradicts the documented guarantee that downloaded binary never reaches statelog verbatim. Please propagate redaction metadata for returned branch values (without changing ordinary global isolation) and add fork/race regressions, including the >4,000-character telemetry path.

packages/agency-lang/stdlib/aws/s3.agency
```diff
@@ -0,0 +1,191 @@
...
+effect std::aws::s3::createBucket {
+  bucket: string,
+  region: string,
+}
+
+export def s3Get(
+  bucket: string,
+  key: string,
+  region: string = "",
+): Result {
```
egonSchiele: Please give all five public functions concrete success types instead of bare `Result` (`Result<string>` for gets, exported object types for puts/create). Bare `Result` is `Result<any, any>`, so callers lose the contracts documented below—for example, a typo such as `value.etga` after success narrowing typechecks and fails only at runtime. The TypeScript executor already defines the exact result shapes; the Agency surface should expose them.

egonSchiele (commented):
I found three correctness/safety issues below. I did not repeat the existing Copilot comment about pre-aborted body-reader cleanup.

egonSchiele:
Thanks — all six addressed in `aadc0f38c`. Verified against the code first; every one was valid.

**Blocker — empty key (`s3.ts`):** `keyFailure` now rejects `key.length === 0` before target construction, so an empty key can no longer turn `s3Get` into ListObjects or `s3Put` into CreateBucket under an object approval. Tests assert all four object helpers reject it and never call `fetch`.

**Fork/race binary redaction (`s3.ts:180`):** you're right — the marker lived only in the branch's cloned `GlobalStore`, and the existing join only propagates durable *object* tags. Added `GlobalStore.copyPrimitiveRedactionsInto`, wired into `runInBranchAlsFrame`'s `finally` alongside the object-tag-flag propagation, so a primitive redacted in a branch is redacted in the parent's `forkBranchEnd` serialization. Only redaction metadata moves (log-only), so global isolation is unchanged. Regression test in `runBatch.test.ts` drives a real fork through the join and asserts both the small and the >4000-char telemetry paths redact to the marker with no leak.

**Concrete result types (`s3.agency`):** the five functions now return `Result<string>` (gets) and exported `Result<S3PutResult>` / `Result<S3CreateBucketResult>` (puts/create) instead of bare `Result`, so `value.etga`-style typos fail at compile time.

**Abort-before-`try` leak (`http.ts:55`):** `readBodyBytesCapped` now installs the cancel path (or cancels immediately when already aborted) *before* `throwIfAborted()`, so an already-aborted signal still cancels the stream and reaches the `finally` that releases the reader lock.

**Missing already-aborted test (`http.test.ts`):** added a test that pre-aborts the signal and asserts the stream's `cancel()` ran and a second `getReader()` does not throw (lock released), not just that the promise rejects.

Full suite for the touched areas is green (495 unit tests incl. `globalStore`/`redactForStatelog`/`runBatch`, structural lint, `sourceIsText`, and the offline Agency fixture).

packages/agency-lang/lib/runtime/runBatch.ts
```diff
@@ -508,6 +508,14 @@ function runInBranchAlsFrame<T>(
         if (!shareGlobals && branchGlobals.hasDurableObjectTagFlag()) {
           parent.globals.setDurableObjectTagFlag();
         }
+        // Durable object tags travel back on the returned value by reference,
+        // but a PRIMITIVE redaction (e.g. binary output marked in the branch)
+        // lives only in the discarded cloned store. Copy those into the parent
+        // so its forkBranchEnd serialization still redacts the branch's returned
+        // value. Redaction is log-only metadata, so isolation is unaffected.
+        if (!shareGlobals) {
```
egonSchiele, line 516: why was this needed?

packages/agency-lang/lib/stdlib/aws/sigv4.test.ts
```diff
@@ -0,0 +1,51 @@
+import { createHash } from "crypto";
```
egonSchiele, line 1: A __tests__ dir is a new pattern in this repo? Is it something we want? What are the existing patterns?

packages/agency-lang/lib/stdlib/objectBytes.ts
```diff
@@ -0,0 +1,20 @@
+import { failure, type ResultFailure } from "../runtime/result.js";
+
+/**
+ * One shared object-size policy for AWS object bodies. Raw downloads and decoded
+ * uploads intentionally share this limit — it bounds the bytes held in memory,
+ * so both directions are capped by the same number rather than two drifting ones.
+ */
+export const AWS_OBJECT_BYTE_LIMIT = 10 * 1024 * 1024;
```
egonSchiele: should this be configurable by users? If so it should be in lib/config.ts. If not, it should be in lib/constants.ts,

egonSchiele:
Addressed the two placement notes in `229c38a9a`, and here's the answer to the `runBatch` question:

**`runBatch.ts` — why the `copyPrimitiveRedactionsInto` call was needed:** it's the primitive counterpart to the object-tag propagation right above it (line 508). Durable *object* tags survive the branch→parent join because they live **on the returned object** and ride back by reference — the flag copy at 508 just re-arms the parent's redaction gate. A **primitive** redaction (the base64 string from `s3GetBinary`) can't be attached to the value; it lives only in the branch's cloned `GlobalStore`, which is discarded at join. So without this copy, the parent's `forkBranchEnd` → `safeStatelogValue` serializes the returned base64 with no tag and it leaks into statelog — exactly the gap you flagged. The regression test in `runBatch.test.ts` fails without this line and passes with it. It copies only redaction markers (log-only metadata, never program state), so global isolation is unchanged. If you'd prefer it scoped to just the returned value rather than all of the branch's primitive redactions, I can walk the return value instead — let me know.

**`__tests__` dir:** you're right, co-location is the dominant convention (`lib/stdlib/layout/*.test.ts`, `lib/stdlib/providers/*.test.ts`); my `aws/__tests__` was the only such subdir. Moved all the new tests to co-located `lib/stdlib/aws/*.test.ts` (and `base64`/`objectBytes` alongside their sources).

**`AWS_OBJECT_BYTE_LIMIT`:** not user-tunable (a fixed in-memory safety cap), so moved to `lib/constants.ts`; `objectBytes.ts` and `http.ts` import it from there.

typecheck (all three tsc projects), structural lint, and `sourceIsText` are green.

egonSchiele:
Good push — I checked this empirically rather than trust my mental model, and the change **is** needed, but **not** because of the custom message.

**What the existing primitive handling actually covers.** There are two mechanisms, both real (`globalStore.tags.test.ts`):
- `clone() keeps primitive tags` — a primitive tag set in the **parent** is inherited **into** each branch (parent → branch).
- `survives toJSON/fromJSON` — primitive tags survive interrupt save/restore.

Both are inheritance/round-trip directions. Neither covers the **third** direction: a primitive tagged **inside a branch**, on the value it returns, flowing **back to the parent**. `clone()` is a deep snapshot (`fromJSON(toJSON())`), so the branch's tag lives only in its cloned store, which is discarded at a successful join. On `main`, the join `finally` propagates only durable **object** tags (`hasDurableObjectTagFlag`) — there was never a primitive branch→parent path. `s3GetBinary` is just the first stdlib code to redact a **primitive** (the base64 string) inside code that can run in a fork branch, so it's the first to hit the latent gap.

**It is not the custom message.** I proved this: I added a third branch that redacts a primitive with the **default** `[REDACTED]` (no label) and returns it, then disabled `copyPrimitiveRedactionsInto` and ran it:

```
AssertionError: expected 'AAAAsmall-branch-blob' to be '[binary output truncated]'
```

Both the custom-label branch **and** the default-`[REDACTED]` branch leak their raw value into the parent's `forkBranchEnd` serialization when the fix is off; with it on, all three redact. So the gap is general to any primitive redacted inside a branch — the custom-marker feature is orthogonal (it only changes *what string* replaces the value, not *whether* the parent knows to replace it). That default-marker case is now part of the regression (`588518eca`).

If you'd still prefer a narrower fix, the alternative is to walk only the returned branch value instead of copying all of the branch's primitive redaction markers — happy to switch to that. But the propagation itself is required.

PR #834, opened 2026-08-10
std::aws/s3: s3PresignGet mints presigned download URLs (query-string SigV4)

egonSchiele:
## What

`s3PresignGet(bucket, key, expiresInSeconds = 3600, region = "")` — a new `std::aws/s3` function that mints a presigned download URL for a private S3 object. Anyone holding the URL can GET that one object until the expiry (up to 7 days); the bucket stays private, and generating the URL is pure local crypto — no request is sent.

Motivating case: a scheduled agent generates an image, `s3PutBinary`s it, and emails a link a browser can actually open.

Spec: `2026-08-09-s3-presigned-urls-spec.md` (reviewed).

## How

- **`presignRequest`** (`sigv4.ts`): the query-string SigV4 variant next to `signRequest`, reusing the same key derivation and hashing. Signature rides in the query, only `host` is signed, payload hash is `UNSIGNED-PAYLOAD`. It consumes the atomic `AwsRequestTarget` (type-only import), and the canonical query that gets signed is byte-for-byte the query emitted in the returned URL — the module's one-encoding contract, extended to the query.
- **`presignGet` executor case** (`s3.ts`): inherits the shared pipeline (env credentials, bounded partition policy, `ValidatedBucket`, one-encoding key canonicalization with `.`/`..` rejection), then stops before transport. The final-hostname defense is **not** inherited — presigning bypasses `sendAwsRequest` — so it is extracted into `hostOutsidePartitionFailure` and called explicitly; a partial-mock test proves the executor consults it.
- **Expiry**: integer in `[1, 604800]` or a coded failure naming the range; never silently clamped.
- **Interrupt-gated, not destructive**: minting a share link is a capability grant, so it interrupts like every S3 operation — and its payload uniquely carries `expiresInSeconds`, so a handler can judge how long the link lives. Nothing remote changes, so no `destructive { }` region.
- **Bearer-credential redaction**: the URL is marked `[presigned S3 URL redacted]` via the existing `markRedacted` machinery (same as binary downloads), so it never lands in a statelog trace; the runtime value stays intact for emailing.
- Added `std::aws::s3::presignGet` to the `AwsS3` and `Network` effect sets.

## Testing

- `sigv4.test.ts`: reproduces AWS's published presigned-GET vector — the full documented URL, byte-for-byte (values copied from the AWS doc page, not retyped); a session-token query-ordering case; a hostile-encoding case (space/`%`/Unicode/repeated `/` in the key, `+`/`/`/`=`/space in the token) asserting the exact emitted path and query.
- `s3.test.ts`: URL shape with no fetch call, redaction marker at the runtime seam, expiry boundary matrix (0/-5/1.5/NaN/604801 rejected; 1 and 604800 accepted), path-style dotted bucket, shared region precedence and key rejection, and the injected-refusal proof that the hostname check runs on the presign path.
- `tests/agency/aws-s3.agency`: a handler round-trips the interrupt payload, pinning the effect name and all four payload fields including `expiresInSeconds`. No LLM calls, no network.

All three test files pass locally; `make`, `lint:structure`, `tsc --noEmit`, and `lib/sourceIsText.test.ts` are green. `docs/site/stdlib/*` changes are `make doc` regeneration from the new docstrings.

## Caveats documented in the docstring

- Treat the URL like the object's contents — anyone holding it can download until expiry.
- With `AWS_SESSION_TOKEN` credentials, AWS invalidates the URL when the session ends, whatever expiry was requested.


packages/agency-lang/lib/stdlib/aws/s3.ts
```diff
@@ -238,6 +268,30 @@ export async function runS3Operation(
...
+        target,
+        region: partition.region,
+        service: "s3",
+        accessKeyId: credentials.accessKeyId,
+        secretAccessKey: credentials.secretAccessKey,
+        sessionToken: credentials.sessionToken,
+        expiresInSeconds: operation.expiresInSeconds,
+      });
+      const { globals } = getRuntimeContext();
+      globals.markRedacted(url, PRESIGN_MARKER);
```
egonSchiele, line 292: **P1 — The bearer URL can still leak through composed strings.** `markRedacted` tags only this exact primitive value. If the caller follows the motivating use case and builds an email body such as `"Download: ${url}"`, that new string has no tag, so both statelog serialization paths record the full signature (and session token). The current test checks only `redactionReplacement(url)`, so it misses this path while the PR/docs claim that the URL never lands in a trace. Please make redaction survive or detect composition and add a regression test with a larger string containing the URL; otherwise narrow the stated guarantee and explicitly document that interpolating the link can leak it to traces.

packages/agency-lang/stdlib/capabilities.agency
```diff
@@ -42,10 +42,11 @@ export effectSet FileSystem = <FileRead, FileWrite>
 export effectSet Shell = <std::bash, std::exec, std::run>
 
 /** Anything that talks to the outside world over the network. */
-export effectSet Network = <std::http::fetch, std::http::fetchJSON, std::http::fetchMarkdown, std::search, std::tavilySearch, std::weather, std::browserUse, std::wikipedia::article, std::wikipedia::search, std::wikipedia::summary, std::gdelt, std::fred, std::dbnomics, std::edgar, std::littlesis, std::yc, std::hackernews, std::wikidata, std::usaspending, std::aws::s3::get, std::aws::s3::getBinary, std::aws::s3::put, std::aws::s3::putBinary, std::aws::s3::createBucket>
+export effectSet Network = <std::http::fetch, std::http::fetchJSON, std::http::fetchMarkdown, std::search, std::tavilySearch, std::weather, std::browserUse, std::wikipedia::article, std::wikipedia::search, std::wikipedia::summary, std::gdelt, std::fred, std::dbnomics, std::edgar, std::littlesis, std::yc, std::hackernews, std::wikidata, std::usaspending, std::aws::s3::get, std::aws::s3::getBinary, std::aws::s3::put, std::aws::s3::putBinary, std::aws::s3::createBucket, std::aws::s3::presignGet>
```
egonSchiele: **P2 — This operation does not satisfy `Network`'s documented contract.** `s3PresignGet` deliberately performs no request; it computes a bearer capability locally. Including it here means callers constrained to `raises <Network>` can mint that capability even though `Network` is defined as operations that talk to the outside world. Please keep it in `AwsS3` but remove it from `Network`, or redefine the capability's contract if `Network` is intended to cover local creation of externally usable credentials too.

egonSchiele (commented):
Two actionable findings: the direct-value redaction does not cover the motivating composed-email path, and the local-only operation does not match the documented Network capability boundary. The SigV4 implementation itself looks sound against the published vector and the one-encoding contract.

egonSchiele:
Both findings addressed in 34431e90f, plus the three inline nits.

**P1 (composed-string leak):** made redaction survive composition rather than narrowing the guarantee. `GlobalStore.redactContainedStrings(text)` scrubs redact-tagged strings that appear inside a larger string (longest secret first, custom labels preserved), and both statelog composition points (`makeRedactReplacer` and `safeStatelogValue`) call it after the exact-match check. This deliberately widens the v1 whole-values-only boundary — the test that locked that boundary is flipped into containment coverage, and `s3.test.ts` adds the motivating regression: an email body interpolating the presigned URL comes out of both paths as `"...Today's image: [presigned S3 URL redacted]..."`. It also covers the session-token-in-URL case, since the whole URL is the tagged secret. Binary base64 embedded in composed strings gets the same protection for free.

**P2 (`Network` membership):** removed `presignGet` from `Network`; it stays in `AwsS3` only, and the set's comment records the exclusion and the reason (mints a bearer capability locally, sends nothing). The comment is now single-line, which also resolves the generated-markdown indentation nit.

**Nits:** expiry errors use `String()` so `NaN`/`Infinity` render as themselves rather than `null`; the hostname refusal message is now neutral (`Refusing <url>: hostname is not under <suffix>`) since it serves both the transport and presign paths.

All 206 tests across the touched files pass; `make`, `lint:structure`, `tsc --noEmit`, and the repo-wide guards are green.

PR #841, opened 2026-08-11
Local models: load the llama-cpp provider through smoltalk (delete the wrapper seam)

egonSchiele:
## What this does

Moves the llama-cpp loading seam into smoltalk, which (as of smoltalk 0.11.0 / smoltalk-llama-cpp 0.2.0, egonSchiele/smoltalk#39) declares the plugin as an optional peer dependency, lazily imports it, and registers the provider itself. Agency keeps only what smoltalk cannot do: finding the package in layouts Node resolution cannot see from smoltalk (globally-installed CLIs), plus all the product-level policy (catalog, aliases, SHA-256 pinning, cache dir).

**User-visible behavior is intentionally unchanged.** The agent `--local` flag, `agency local download/list/remove/resolve/refresh`, and every error message behave exactly as before; the agency-js `local-model` fixture and the localModels unit tests pin that.

This is PR 1 of 2 from the spec at `packages/agency-lang/2026-08-11-run-local-flag-spec.md`. PR 2 (branched on top) adds `agency run --local` and the local CLI ergonomics.

## Changes

- **New `lib/runtime/localProvider.ts`** — the one loader entry point. `loadLocalProvider()` calls smoltalk's `loadLlamaCpp`, choosing the entry path via the pure `chooseEntryPath`: the `AGENCY_LLAMA_PROVIDER_MODULE` override, then nothing when the package resolves locally (smoltalk bare-imports it — under pnpm its peer instancing picks the right copy), then the global npm/pnpm roots probe. The probe functions (`globalNodeModulesRoots`, `resolveSmoltalkLlamaCppFromRoots`, `resolveSmoltalkLlamaCppEntry`) moved here from `localModels.ts` — the runtime cannot import stdlib without a cycle, and `mcpResolver.ts` (the other prober consumer) now imports from here too.
- **`_registerLocalProvider` / `_downloadModel`** now go through `loadLocalProvider()`; downloading uses the plugin's exported `resolveModel`. The freshness-snapshot + SHA-256 verification flow is untouched.
- **Bootstrap hook** — `ensureConfiguredLocalProvider(execCtx)` runs at both bootstrap sites (fresh runs in `runtime/node.ts`, resumes in `runtime/interrupts.ts`): when the baked config names `llama-cpp` as the provider, the child process pre-loads it with the probe-assisted path, covering the global-install layout that smoltalk's own auto-load cannot resolve.
- **Deleted:** the bundled `lib/stdlib/providers/llama-cpp.mjs` wrapper (agency's only dynamic-import special case for this), the `AGENCY_SMOLTALK_LLAMA_CPP_PATH` env relay, and the path-splitting `llamaModelConfig.ts` (the plugin's `LlamaCPP` accepts a `.gguf` path directly now). The Makefile wipes the stale `dist/lib/stdlib/providers` on build.
- **`AGENCY_LLAMA_PROVIDER_MODULE` changes contract** (test/advanced seam only): it is now an entry path to a *plugin-shaped* module (a `LlamaCPP` class + `resolveModel`), the same shape smoltalk validates on the real package — previously a `register()`-shaped wrapper module. Test fakes reshaped accordingly.
- `smoltalk` bumped to `^0.11.0`. smoltalk-llama-cpp remains a non-dependency; the optional-peer relationship lives in smoltalk.
- `docs/dev/local-models.md` updated for the new seam.

## Testing

- `lib/runtime/localProvider.test.ts` (new): `chooseEntryPath` branch coverage (override absolutization, local→bare-import with the global probe provably not run, global-only, nothing-found), `loadLocalProvider` via a plugin-shaped fake, `ensureConfiguredLocalProvider` no-op vs load, and the ported global-roots discovery tests.
- `lib/stdlib/localModels.test.ts`: register/download/verify/quarantine/cache-hit tests all pass on the new path with plugin-shaped fakes.
- `tests/agency-js/local-model`: end-to-end through the compiled bridge (registerLocalModel → smoltalk loader → provider registered + fake download), fixture updated for the class name.
- `lib/cli/local.test.ts`, `lib/stdlib/mcpResolver.test.ts`, `lib/sourceIsText.test.ts`, and `pnpm run lint:structure` all pass; full `make` build is clean.


packages/agency-lang/lib/runtime/localProvider.ts
```diff
@@ -0,0 +1,120 @@
+import * as fs from "node:fs";
```
egonSchiele, line 1: When we load a local model, would it make sense to log in state log what model was loaded and from where?

egonSchiele (commented):
Focused review found no blocking correctness issues. The loader precedence now matches the intended override → local bare import → global explicit path behavior, registration is re-ensured by smoltalk after `unregisterProvider`, both fresh and resume bootstrap paths are covered, and CI is green.

One non-blocking verification gap remains: the core production-only case is a globally installed Agency CLI plus globally installed `smoltalk-llama-cpp`, but the tests verify its pieces independently rather than executing that real layout. Please run and record the spec's global-install local-model smoke before merge; that is the only check likely to catch a package-resolution/peer-instancing mismatch that these fakes cannot model.

packages/agency-lang/lib/runtime/localProvider.ts
egonSchiele: Yes — implemented in 89dc6383c as a new `localModelLoaded` leaf event (following the `transcription`/`imageGeneration` pattern in `statelogClient.ts`). It carries the pinned model (the `.gguf` path from config) plus where the provider package was resolved from: `entrySource` is the `chooseEntryPath` decision — `override` (AGENCY_LLAMA_PROVIDER_MODULE), `local` (bare import), or `global` (a global npm/pnpm root) — with `entryPath` when there is one.

It fires from the two places a load actually happens with a statelog client in reach: the bootstrap hook (`ensureConfiguredLocalProvider`, covering compiled runs whose config names llama-cpp) and `_registerLocalProvider` via the `__ctx()` async-context peek (covering the agent's `--local` path; the plain `agency local` CLI has no runtime frame and emits nothing).

One scoping note: this logs the model the run is *pinned to* and where the *package* came from. The actual weights-into-RAM load happens lazily inside smoltalk-llama-cpp's `acquireModelEntry` on the first call — if we ever want load *duration* in statelog, that instrumentation belongs over there (the client already has a statelog hook for `promptRequest`), as a smoltalk-side follow-up.

PR #842, opened 2026-08-11
agency run --local, a download picker, and a catalog list view

egonSchiele:
## What this does

PR 2 of 2 from `packages/agency-lang/2026-08-11-run-local-flag-spec.md`, on top of the smoltalk loader seam (#841). Three user-facing pieces:

**`agency run --local <model>`** — pin a whole run to a local llama-cpp model. The value is a curated name, an alias, an `hf:` URI, or a `.gguf` path (the same domain as `agency local download`). `--local` with `--model` errors with exit 2 before any other work. The value is required, not optional: an optional-valued flag would swallow the input filename (`agency run --local foo.agency` would parse `foo.agency` as the model), the same commander behavior that split `--trace` in two. Resolution and download run in the parent CLI — progress and SHA-256 verification appear in the terminal before the program starts — and the result rides the existing `CliFlags.model` pathway (`{ model: <absolute path>, explicitProvider: "llama-cpp" }`), so `applyCliFlags` remains the single owner of flag→config meaning. The path is absolutized because the plugin rejects bare separator-less filenames. The child process gets the provider from baked config via the #841 bootstrap hook. Combines naturally with `--max-cost 0` (no paid spend, local only).

**`agency local download` with no argument opens a picker** — a `prompts` select over the catalog (name, params, size), with a trailing "custom" choice for an `hf:` URI or `.gguf` path. Non-TTY prints the catalog plus a `Pass a model:` hint and exits 1; cancel exits 0 without downloading.

**`agency local list` becomes the catalog view** — first line names the resolved models directory (env → `agency.json` → default), every catalog entry shows with a `✓` marker and on-disk size when downloaded, and files no catalog row claims (raw-URI or pre-manifest downloads) stay visible under `OTHER FILES` with the total. The command is now **ungated**: browsing what is available requires no provider package (`download`/`remove` keep their gates).

## The downloads manifest

Catalog names map to `hf:org/repo:quant` URIs, but node-llama-cpp picks the final `.gguf` filename, so "is this entry downloaded" is not derivable after the fact. `_downloadModel` now records `{ resolved URI → basename }` in a `downloads.json` in the models cache dir, after verification succeeds (an invalid-hash file is never recorded), via temp-file + rename so an interrupted write keeps the previous valid manifest. It is display metadata only — resolution, downloading, and verification never read it, so a corrupt/deleted manifest can mislabel the list view and nothing else. Models downloaded before the manifest existed show under `OTHER FILES` until re-downloaded (cosmetic only).

## Testing

- `lib/cli/runLocal.spawn.test.ts` (new): end-to-end `agency run --local` through the built CLI with a fake plugin — parent resolves + absolutizes, config bakes, child bootstrap loads the fake from the inherited env override, and the program's `llm()` call lands on it; plus the `--local`+`--model` exit-2 contract. No network, no real model.
- `lib/cli/localFlag.test.ts` (new): resolution + absolutization + the install-hint propagation.
- `lib/stdlib/localModelManifest.test.ts` (new): round-trip/upsert, corrupt-manifest tolerance, non-string dropping, dir creation, no leftover temp file.
- `lib/stdlib/localModels.test.ts`: `formatLocalList` (markers, stale-manifest entries, raw-URI visibility in OTHER FILES) and manifest recording in the download flow.
- `lib/cli/local.test.ts`: ungated `runList` regression (the old code exited 1 without the package), `downloadChoices` labels, and the non-TTY download contract (catalog + hint + exit 1).
- Verified live on this machine: the new `list` shows the models directory, the full catalog, and three pre-manifest `.gguf` files under `OTHER FILES`; non-TTY `download` exits 1.
- `tsc --noEmit`, `lib/sourceIsText.test.ts`, and `pnpm run lint:structure` all clean; full `make` build clean.

### DIFF
--- a/packages/agency-lang/lib/cli/local.ts
+++ b/packages/agency-lang/lib/cli/local.ts
@@ -90,7 +90,10 @@ export async function runDownload(value?: string): Promise<void> {
   gate();
   let picked = value;
   if (picked === undefined) {
-    if (!process.stdout.isTTY) {
+    // Prompting needs BOTH ends of the terminal: a TTY stdout to draw on and
+    // a TTY stdin to read from (`agency local download < /dev/null` from a
+    // terminal has a TTY stdout but nothing to read).
+    if (process.stdin.isTTY !== true || process.stdout.isTTY !== true) {
       // A script that reaches this point asked for a download and did not
       // get one — print what is available and fail.
       console.log(formatModelCatalog());
@@ -103,7 +106,9 @@ export async function runDownload(value?: string): Promise<void> {
       message: "Which model do you want to download?",
       choices: downloadChoices(_listModelNames()),
     });
-    if (answer.model === undefined) return; // cancelled — exit 0, nothing downloaded
+    // Cancellation can surface as a missing key or as null — treat both as
+    // "exit 0, nothing downloaded".
+    if (answer.model == null) return;
     picked = answer.model as string;
     if (picked === CUSTOM_CHOICE) {
       const custom = await prompts({
@@ -111,7 +116,7 @@ export async function runDownload(value?: string): Promise<void> {
         name: "value",
         message: "hf: URI or .gguf path:",
       });
-      if (custom.value === undefined || custom.value === "") return;
+      if (custom.value == null || custom.value === "") return;
       picked = custom.value as string;
     }
   }
--- a/packages/agency-lang/lib/cli/local.ts
+++ b/packages/agency-lang/lib/cli/local.ts
@@ -1,3 +1,4 @@
+import prompts from "prompts";
 import {
   _resolveModelName,
   _downloadModel,
@@ -8,9 +9,11 @@ import {
   _unaliasModel,
   _removeModel,
   hasLocalModelSupport,
+  formatGB,
   formatModelCatalog,
   formatLocalList,
   _refreshCatalog,
+  type ModelNameEntry,
   type RefreshResult,
 } from "../stdlib/localModels.js";
 import { readDownloadManifest } from "../stdlib/localModelManifest.js";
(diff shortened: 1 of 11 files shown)


packages/agency-lang/lib/stdlib/localModelManifest.ts
```diff
@@ -0,0 +1,39 @@
...
+  } catch {
+    return {};
+  }
+}
+
+export function recordDownload(dir: string, uri: string, file: string): void {
+  fs.mkdirSync(dir, { recursive: true });
+  const next = { ...readDownloadManifest(dir), [uri]: file };
+  const target = path.join(dir, MANIFEST_FILE);
+  const tmp = `${target}.tmp`;
```
egonSchiele: The fixed `.tmp` name makes concurrent downloads fail, rather than merely produce the documented last-writer-wins result. Two processes can both write `downloads.json.tmp`; after one renames it, the other `renameSync` gets `ENOENT`, and that exception propagates out of `_downloadModel` after the model was successfully downloaded. Please use a unique sibling temp name (for example including PID plus a random suffix) and clean it up on failure. Since this is display-only metadata, a manifest race should not turn a successful download into a failed command.

packages/agency-lang/lib/cli/local.ts
```diff
@@ -48,27 +53,73 @@ export function aliasRemove(name: string, file?: string): string {
...
+    value: e.name,
+  }));
+  return [...rows, { title: "custom (hf: URI or .gguf path)…", value: CUSTOM_CHOICE }];
+}
+
+export async function runDownload(value?: string): Promise<void> {
   gate();
+  let picked = value;
+  if (picked === undefined) {
+    if (!process.stdout.isTTY) {
```
egonSchiele: Interactive prompting needs readable stdin as well as a TTY stdout. With `process.stdout.isTTY === true` but piped/closed stdin (for example `agency local download < /dev/null` from a terminal), this enters `prompts` instead of taking the documented non-interactive path. Please gate the picker on both `process.stdin.isTTY === true` and `process.stdout.isTTY === true`, and extend the test to cover non-TTY stdin.

packages/agency-lang/lib/stdlib/localModelManifest.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (parsed === null || typeof parsed !== "object" || Array.isArray(parsed)) {
+      return {};
+    }
+    const out: Record<string, string> = {};
+    for (const [k, v] of Object.entries(parsed)) {
+      if (typeof v === "string") out[k] = v;
+    }
+    return out;
+  } catch {
+    return {};
```
egonSchiele, line 35: print whatever error occurred so it doesn't get swallowed

packages/agency-lang/lib/stdlib/localModelManifest.ts
egonSchiele: Fixed in c029a1ace. The temp name is now unique per writer (`downloads.json.<pid>-<random>.tmp`), so concurrent downloaders race only on the atomic rename (last writer wins, as documented) instead of colliding on the temp file. And `recordDownload` as a whole no longer throws — any failure warns and cleans up its temp file, so a bookkeeping problem can never fail a download that already succeeded (test: dir-path-is-a-file case).

packages/agency-lang/lib/cli/local.ts
egonSchiele: Fixed in c029a1ace — the picker now requires `process.stdin.isTTY === true && process.stdout.isTTY === true`; anything else takes the catalog + hint + exit-1 path. The test now covers both shapes: fully non-TTY, and TTY stdout with piped stdin (the `download < /dev/null` from a terminal case).

packages/agency-lang/lib/cli/local.ts
```diff
@@ -48,27 +53,73 @@ export function aliasRemove(name: string, file?: string): string {
...
+    picked = answer.model as string;
+    if (picked === CUSTOM_CHOICE) {
+      const custom = await prompts({
+        type: "text",
+        name: "value",
+        message: "hf: URI or .gguf path:",
+      });
+      if (custom.value === undefined || custom.value === "") return;
+      picked = custom.value as string;
+    }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c029a1ace: both cancellation checks are now loose (`answer.model == null`, `custom.value == null`), so a null resolution exits 0 the same as a missing key.

packages/agency-lang/lib/stdlib/localModels.ts
```diff
@@ -915,6 +923,9 @@ export async function _downloadModel(value: string, cacheDir: string = ""): Prom
   if (expected !== undefined && wasFresh(resolved)) {
     await verifyModelFile(resolved, expected, value);
   }
+  // Record-after-verify: an invalid-hash file throws above and is never
+  // written into the manifest.
+  recordDownload(dir, target, path.basename(resolved));
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c029a1ace: the never-throw guarantee now lives inside `recordDownload` itself (warn + temp-file cleanup on any failure), so the display-only contract holds structurally — no call site can reintroduce the problem. Covered by a test that makes the cache-dir path a plain file.

packages/agency-lang/lib/stdlib/localModelManifest.ts
```diff
@@ -0,0 +1,39 @@
...
+    const parsed: unknown = JSON.parse(
+      fs.readFileSync(path.join(dir, MANIFEST_FILE), "utf-8"),
+    );
+    if (parsed === null || typeof parsed !== "object" || Array.isArray(parsed)) {
+      return {};
+    }
+    const out: Record<string, string> = {};
+    for (const [k, v] of Object.entries(parsed)) {
+      if (typeof v === "string") out[k] = v;
+    }
```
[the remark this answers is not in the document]
egonSchiele: Adopted the null-proto convention in c029a1ace (`Object.create(null)` for the read result, matching the provider-registry precedent), with a test that `__proto__`/`toString` keys behave as plain data. One correction for the record: actual prototype pollution was not reachable here even before — the `typeof v === "string"` filter means the only value ever assigned to `__proto__` is a string, and setting a prototype to a non-object is a silent no-op. The real defect was subtler (a `__proto__` key was silently dropped and prototype-named keys shadowed methods); the null-proto read fixes that too.

PR #847, opened 2026-08-14
Scheduled agents: push a branch, do the work, stay in budget

egonSchiele:
Implements docs/superpowers/specs/2026-08-13-scheduled-agents-branch-output-design.md, via the reviewed plan (docs/superpowers/plans/2026-08-13-scheduled-agents-branch-output.md and its -review.md, both included on this branch).

The three scheduled agents now push a branch instead of opening a pull request,
apply their fixes instead of only reporting them, and run inside a cost and time
budget. Also bumps run-agency-action to v1.0.3, without which none of them run
at all.

Three controls worth reviewing closely, all pure/shared code in branch.agency
with direct tests in tests/agency/scheduled-helpers.agency:

- `isGithubPath` — rejects edits under .github/, the one path that executes
  before the review gate closes. Tested at the interrupt level (raised
  std::edit under the handler), including the traversal spelling.
- `shouldForcePush` — refuses to force-push over a human's commits.
- `handleScheduledInterrupt` rejecting `std::guard` — CI runs `agency run`
  with no policy flag, so an unadjudicated guard trip would exit non-zero
  with no salvage. Rejection routes trips into the salvage pipeline:
  docs-review pushes its partial work with a truncation banner; the other two
  exit cleanly with "nothing to push". Tested with fake-clock guard trips.

After merge: trigger each workflow once with workflow_dispatch (cheapest
first, stdlib-docs-links) and confirm the force-push safeguard per Task 8 of
the plan.


packages/agency-lang/agents/scheduled/branch.agency
```diff
@@ -0,0 +1,261 @@
...
+/** Absolute path to the agency-lang package inside the checkout. */
+export def pkgDir(): string {
+  """Return the absolute path to `packages/agency-lang` inside the checkout."""
+  return "${repoRoot()}/packages/agency-lang"
+}
+
+/** A UTC timestamp suffix suitable for report headers, e.g. `20260713-141500`. */
+export def stamp(): string {
+  """Return a UTC timestamp like `20260713-141500` for use in report headers."""
+  const stampResult = run("date", ["-u", "+%Y%m%d-%H%M%S"], "")
```
egonSchiele: Could use std::date instead https://agency-lang.com/stdlib/date.html

packages/agency-lang/agents/scheduled/branch.agency
```diff
@@ -0,0 +1,261 @@
...
+  first-run case — so it must NOT go through `run()`, which aborts on failure.
+
+  Known conflation: a transient network/auth failure also returns "", which
+  arms a force-push. The window is tolerable — the push moments later needs
+  the same network and credentials working, and destroying work additionally
+  requires a human tip to exist at that exact moment. If this ever needs
+  closing, `git ls-remote --heads origin <branch>` distinguishes "absent"
+  from "unreachable" cheaply.
+*/
+export def remoteTipAuthor(root: string, branch: string): string {
```
egonSchiele, line 101: use std::git https://agency-lang.com/stdlib/git.html

packages/agency-lang/agents/scheduled/branch.agency
```diff
@@ -0,0 +1,261 @@
...
+    so an unadjudicated trip would surface unhandled and kill the process with
+    no salvage. reject() routes the trip into the salvage pipeline instead:
+    the guard returns success(draft) when the block saved one, and the plain
+    failure otherwise. Rejecting (not approving) is deliberate — approving
+    would grant more budget and defeat the cap.
+
+  Everything else passes: effectful calls in the agents discharge themselves
+  with `with approve` at the call site.
+*/
+export def handleScheduledInterrupt(i) {
```
egonSchiele, line 311: If you use std::git, you can use effect sets to approve operations https://agency-lang.com/guide/effects-and-raises.html

packages/agency-lang/agents/scheduled/constants-review.agency
```diff
@@ -1,98 +1,104 @@
...
 
-Current lib/constants.ts (do NOT re-suggest anything already here):
+    // Read the current constants file with the typed `read` tool (not a raw `cat`).
+    const constsResult = read("lib/constants.ts", pkg) with approve
+    let consts = ""
+    if (constsResult is success(text)) {
+      consts = text
+    } else {
+      print("Failed to read lib/constants.ts. Aborting.")
+      exit(1) with approve
```
egonSchiele: don't need `with approve` to exit

packages/agency-lang/agents/scheduled/docs-review.agency
```diff
@@ -1,124 +1,142 @@
...
+
+    let sections: string[] = []
+    let processed = 0
+    const budget = guard(label: "docs-review", cost: $3.00, time: 25m) {
+      let idx = 0
+      for (rel in files) {
```
egonSchiele: Could use `fork` for concurrency instead https://agency-lang.com/guide/concurrency.html (you may want to write a function that lets you create a concurrency pool to limit the number of concurrent threads though). If you use fork, remember to set `shared: true`

packages/agency-lang/agents/scheduled/stdlib-docs-links.agency
```diff
@@ -85,129 +86,126 @@ def toSlug(relPath: string): string {
...
-      genSlugs.push(toSlug(docPath))
+    // Every generated stdlib doc, as the sidebar slug it should be linked by.
+    const genResult = glob("**/*.md", dir: "${pkg}/docs/site/stdlib") with approve
+    let genSlugs: string[] = []
+    if (genResult is success(paths)) {
+      for (docPath in paths) {
+        genSlugs.push(toSlug(docPath))
+      }
+    } else {
+      print("Failed to list generated stdlib docs. Aborting.")
```
egonSchiele: get and print the error from the failure, to help debug?

packages/agency-lang/agents/scheduled/stdlib-docs-links.agency
```diff
@@ -85,129 +86,126 @@ def toSlug(relPath: string): string {
...
-  }
+    const commitMsg = "chore(docs): link ${missing.length} unlinked stdlib doc(s) in the sidebar"
+    const paths = ["packages/agency-lang/docs/site/.vitepress/config.mts"]
+    const url = pushBranch(root, branch, paths, commitMsg)
+    if (url == "") {
+      print("No changes to push.")
+    } else {
+      print("Pushed ${branch}. Open a PR: ${url}")
+    }
+  } with handleScheduledInterrupt
```
egonSchiele, line 223: after editing config.mts, run `pnpm run build` in docs/site/ and confirm the build passes.

packages/agency-lang/tests/agency/scheduled-helpers.agency
```diff
@@ -42,3 +46,133 @@ node extractNoLinks() {
...
+}
+
+node forcePushWhenNoRemoteBranch() {
+  // "" means the branch does not exist on the remote yet.
+  const got = shouldForcePush("")
+  if (got) { return "ok" }
+  return "bad:false"
+}
+
+node noForcePushOverHuman() {
```
egonSchiele, line 101: if this test fails, will it force push a branch?

packages/agency-lang/agents/scheduled/branch.agency
egonSchiele: Done in 90ae8c626 — stamp() is now format(now(), "UTC") from std::date, no shell-out at all.

packages/agency-lang/agents/scheduled/branch.agency
egonSchiele: Adopted where std::git has the operation (90ae8c626): gitLog for the remote tip author, gitAdd, gitCommit, and gitDiff(staged: true) for the empty-index check — which also fixes Copilot's exit-code-ambiguity finding structurally. fetch, push, config user.*, and checkout -B stay on exec: std::git has no fetch/push/config, and gitSwitch(create: true) fails on an existing branch where -B means create-or-reset. Each exec fallback carries a comment saying why.

packages/agency-lang/agents/scheduled/branch.agency
egonSchiele: The handler now approves the std::git::* family wholesale (90ae8c626) — safe because only deterministic agent code raises them; the model never holds git tools. I used a prefix match on the effect name rather than an effect-set membership test: effectSets (GitRead/GitWrite/Git in std::git) are compile-time raises allowlists, and I found no runtime set-membership operator for handlers — happy to switch if one exists.

packages/agency-lang/agents/scheduled/constants-review.agency
egonSchiele: Checked this with a probe first: a bare exit(3) with no handler surfaces as "Interrupt std::exit was not handled" and the process exits 1, not 3 (exit() raises interrupt std::exit in stdlib/system.agency:81). So instead of dropping approval entirely, handleScheduledInterrupt now approves std::exit and every call site is a bare exit(1) (90ae8c626) — same ergonomics you asked for, with the approval living in the one handler.

packages/agency-lang/agents/scheduled/docs-review.agency
egonSchiele: Done with one deliberate deviation (90ae8c626): reviews now fork four at a time via a chunk() pool helper, but branches hand results back by RETURN VALUE instead of shared: true mutation. Reason: a branch's saveDraft lands on the branch's own stack where the parent guard cannot see it (tests/agency/guards/save-draft-guard-outside-fork.agency), so the draft-keeping has to happen in the parent anyway — it now runs at each wave join, making salvage per-wave (worst case ~4 docs of work lost on a trip, vs 1 before). Return-value accumulation then makes shared state unnecessary and keeps branch isolation. If you'd rather have full-width fork + shared counters, say so and I'll switch, but the guard would only be able to salvage an empty draft on a mid-fork trip.

packages/agency-lang/agents/scheduled/stdlib-docs-links.agency
egonSchiele: Done in 90ae8c626 — both the config.mts read failure and the stdlib glob failure now bind and print the underlying error.

packages/agency-lang/agents/scheduled/stdlib-docs-links.agency
egonSchiele: Done in 90ae8c626: after writing config.mts and before pushBranch, the agent runs corepack pnpm install --frozen-lockfile at the repo root and then pnpm run build in docs/site, both through run() so a failure prints stderr and fails the workflow before anything is pushed. The install is needed because the runner starts without workspace deps (run-agency-action only npm-installs); corepack picks up the pinned pnpm from the root packageManager field, with COREPACK_ENABLE_DOWNLOAD_PROMPT=0 set via setEnv so it does not prompt.

packages/agency-lang/tests/agency/scheduled-helpers.agency
egonSchiele: No — shouldForcePush is a pure decision function (author string in, boolean out); these nodes never touch git, so a failing test can only print "bad:...". Added a comment above the tests saying exactly that (90ae8c626). The only code that acts on the decision is pushBranch, which no test calls.

packages/agency-lang/agents/scheduled/branch.agency
```diff
@@ -0,0 +1,261 @@
...
+
+  // `git diff --cached --quiet` exits 0 when the index is empty and 1 when
+  // there are staged changes. A non-zero exit here is EXPECTED, not a failure,
+  // so it must NOT go through run(). An empty index means the generated output
+  // already matches main — skip cleanly rather than pushing an empty commit.
+  const staged = exec("git", ["-C", root, "diff", "--cached", "--quiet"], cwd: "") with approve
+  if (staged.exitCode == 0) {
+    print("No changes to commit; nothing to push.")
+    return ""
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed structurally in 90ae8c626: the empty-index check now uses std::git's gitDiff(staged: true) and tests files.length == 0, with an explicit failure branch that prints the error and exits — a git error can no longer be mistaken for either "no changes" or "changes present".

packages/agency-lang/agents/scheduled/docs-review.agency
```diff
@@ -1,124 +1,142 @@
...
+    // `git add -A`.
+    const url = pushBranch(
+      root,
+      branch,
+      ["packages/agency-lang/docs/dev"],
+      "chore(docs): weekly docs/dev freshness fixes",
+    )
+    if (url == "") {
+      print("No changes to push.")
+    } else {
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 90ae8c626 in all three agents: the message is now neutral ("Nothing pushed; the log above says why...") and defers to pushBranch's own log line for the reason.

PR #848, opened 2026-08-15
Promote a statelog trace into a label dataset

egonSchiele:
## What

Label one trace out of a multi-trace statelog into a human-judged **dataset** (`agency label`) — from the log viewer (`l`) or headless (`agency label ingest <statelog> --trace <id>`) — so you can judge an agent's output after the fact, reusing all the existing labeling machinery (checklists, notes, occurrences, durability).

Implements the spec at `packages/agency-lang/2026-08-14-statelog-to-dataset-labeling-spec.md` and the plan at `packages/agency-lang/2026-08-15-statelog-to-dataset-labeling-plan.md`.

## How it fits the existing design

Once extracted, a statelog trace is just a `{ task, output }` example — the same shape a run produces — so it flows through the same machinery:

- **Output selection** is the shared `selectLabelingFinalOutput` (an `evalOutput()` value, else the entry node's return value; truncated/absent rejected, never the last chat message).
- **Projection** into occurrence fields is the shared `projectOccurrenceFields` (`load/occurrence.ts`), used by both `loadRun` and the statelog loader.
- A trace is only labelable if it **terminated successfully**: a terminal `runtimeError` + result-less `agentEnd` is rejected as `run-failed` (mirroring how run-directory ingestion refuses a `failed` input), and a trace with no `agentEnd` is `trace-unfinished`.

The only statelog-specific pieces are: extracting an eval record from raw events (a run has a pre-extracted `eval-record.json`) and a small `statelog` occurrence-origin variant `{ traceId, finalOutputIndex }` (the `run` origin needs an `inputId` a raw trace has no honest value for). The origin is additive — no manifest bump; the dataset stays at schema version 2.

## Surfaces

- **CLI:** `agency label ingest <statelog> --trace <id>` (repeatable). The `statelog` format is auto-detected from the first envelope, reusing the `agency logs` classifier. `--dataset`/`--checklist` live on the `label` parent.
- **Viewer:** in the tree, `l` labels the focused trace when a dataset is configured — it resolves the output, lets you edit the task, writes the example, and opens the labeling screen focused on it. `l` never expands (`Right`/`Enter` do); where labeling is unavailable (stdin/remote) it does nothing. It refuses a statelog with unparseable lines, and pauses the follow watcher for the handoff.

Two lifecycles are owned once and shared by both surfaces: `datasetWriter` (lock → `openDataset` → ingest → release) and `labelingHost` (`openLabelingSession` → `runLabelTui`, on an existing `Screen`).

## Clean break: `store` → `dataset`

Because there are no datasets on disk and no users, `store` was renamed to `dataset` outright — **not** deprecated. `--store` and `eval.labelStore` were removed; only `--dataset` / `eval.dataset` exist. Durable on-disk names (`outputId`, `out_`, `outputs.jsonl`) are deliberately unchanged (frozen wire format).

## Testing

Unit tests for every piece (resolver incl. failed/unfinished/transient-error cases, projection, scan, loader + CLI selection, statelog origin, `datasetWriter`/`labelingHost` lifecycles, `focusItem`, the `labelTrace` orchestrator, and the viewer `l` behavior via the real TUI harness incl. follow-mode truncation). Touched-area suite green, `make` + `lint:structure` clean, no binary/NUL files. Integration CLI tests updated for the `--dataset` rename.

## Docs

- Dev: new `docs/dev/statelog-to-dataset.md`, a vocabulary note in `docs/dev/eval-labeling.md`, a `CLAUDE.md` pointer.
- Site: `docs/site/cli/eval.md`, `docs/site/cli/logs.md`, `docs/site/guide/observability.md` updated for `--dataset`, the `statelog` ingest source, `--trace`, and the `l` label key.


packages/agency-lang/lib/stdlib/builtins.ts
```diff
@@ -15,16 +16,22 @@ import type { RuntimeContext } from "../runtime/state/context.js";
...
 import type { ThreadStore } from "../runtime/state/threadStore.js";
 import { abortableSleep } from "./abortable.js";
+import { recordPrint } from "./statelog.js";
 import { acceptsFailures } from "../runtime/failurePropagation.js";
 
 const execFileAsync = promisify(execFile);
 
 export function _print(...messages: any[]): void {
   console.log(...messages);
+  recordPrint("print", format(...messages));
```
egonSchiele: [P1] Formatting before the statelog redaction boundary loses object-level redaction metadata. If `redact()` tagged an object as a whole, `format(...messages)` produces a new untagged string; the replacer can only scrub tagged primitive strings contained in it, so the object's fields can be persisted to local/remote statelog sinks. This line also inspects every argument a second time after `console.log`, so a stateful or throwing `util.inspect.custom` hook can run twice and make `print` throw after output was already written. Please preserve the original values through the redaction path and ensure console formatting/inspection happens only once; an integration test with a tagged object plus a stateful custom inspect hook would pin both invariants.

packages/agency-lang/lib/eval/label/load/statelog.ts
```diff
@@ -0,0 +1,298 @@
...
+  }
+  for (const traceId of args.request.traceIds) {
+    if (!(traceId in scan.eventsByTrace)) {
+      throw new IngestSourceError(
+        `Trace "${traceId}" is not in ${args.path}.\n${describeAvailableTraces(scan)}`,
+      );
+    }
+  }
+
+  const occurrences: LoadedOccurrence[] = [];
```
egonSchiele, line 176: [P2] Please validate the keys of `printSelections` before processing the requested traces. With `--trace A --output B=print:0`, this loop processes A and silently ignores B, so a typo in a keyed selector succeeds while promoting something other than the operator intended. Reject every selector whose trace ID is absent from `traceIds` (and add the unrequested-selector case from the plan).

packages/agency-lang/lib/eval/label/load/format.ts
```diff
@@ -47,12 +49,45 @@ export function resolveFormat(
...
+  const firstLine = firstNonEmptyLine(resolved);
+  return firstLine !== undefined && isStatelogEnvelope(firstLine);
+}
+
+function firstNonEmptyLine(resolved: string): string | undefined {
+  const fd = fs.openSync(resolved, "r");
+  try {
+    const buffer = Buffer.alloc(SNIFF_CHUNK_BYTES);
+    const bytesRead = fs.readSync(fd, buffer, 0, buffer.length, 0);
+    const text = buffer.subarray(0, bytesRead).toString("utf8");
```
egonSchiele: [P2] A valid statelog is misclassified when its first non-empty event is larger than `SNIFF_CHUNK_BYTES`: this returns a truncated JSON line, `isStatelogEnvelope` rejects it, and auto detection reports an unknown format. Initial prompt/completion events can legitimately exceed 64 KiB. Please read through the terminating newline with an explicit maximum (and diagnose an over-limit first envelope), rather than parsing an arbitrary prefix as a complete line.

egonSchiele (commented):
Found several correctness issues in the promotion and print-recording paths; inline comments include reproducible failure sequences and focused fixes.

packages/agency-lang/lib/logsViewer/run.ts
```diff
@@ -159,6 +245,18 @@ export async function runViewer(opts: RunViewerOpts): Promise<ViewerResolution>
       action.onResult(text);
     } else if (action.kind === "copy") {
       copyToClipboard(action.text, (message) => stack.active().notify(message));
+    } else if (action.kind === "promoteTrace" && opts.promotion !== undefined) {
+      const outcome = await runViewerPromotion({
+        screen,
+        launch: opts.promotion,
+        traceId: action.traceId,
+        traceEvents: allEvents.filter((event) => event.trace_id === action.traceId),
```
egonSchiele, line 303: [P1] This passes only successfully parsed events to promotion even when `parseErrors` is non-empty. The viewer deliberately tolerates malformed lines, but dataset ingestion must not: pressing `l` can currently persist an output from a partial trace, bypassing `scanStatelog`'s strict parse policy. Reject promotion with an actionable notification whenever the current parse has any errors (including errors discovered by follow-mode reparses).

packages/agency-lang/lib/logsViewer/run.ts
```diff
@@ -159,6 +245,18 @@ export async function runViewer(opts: RunViewerOpts): Promise<ViewerResolution>
       action.onResult(text);
     } else if (action.kind === "copy") {
       copyToClipboard(action.text, (message) => stack.active().notify(message));
+    } else if (action.kind === "promoteTrace" && opts.promotion !== undefined) {
+      const outcome = await runViewerPromotion({
```
egonSchiele: [P1] In follow mode the watcher remains active across this await. While `nextLine` or the labeling TUI owns the shared `Screen`, an append triggers `onNewText -> render()` and paints the log viewer over the prompt/labeling UI. Suspend watcher-driven updates/rendering for the handoff, then apply accumulated text and repaint after it returns; please cover an append that occurs during labeling.

packages/agency-lang/lib/logsViewer/run.ts
```diff
@@ -46,10 +52,86 @@ export type RunViewerOpts = {
...
 export type ViewerResolution = "quit" | "back";
 
+const SOURCE_NAME_PREVIEW_CHARS = 60;
+
+/** The batch name a promoted occurrence records: the trace's agent name if it
+ *  set one, else the statelog file's basename. Keeps "did v2 beat v1"
+ *  answerable without asking. */
+function promotionSourceName(events: readonly EventEnvelope[], sourcePath: string): string {
+  for (const event of events) {
+    if (event.data.type === "agentName" && typeof event.data.name === "string") {
```
egonSchiele, line 79: [P2] This selects the first agent name, while the extractor and `scanStatelog` deliberately use the last `agentName` event because an agent may rename itself during a run. Viewer promotion therefore records a different/stale source from headless ingestion for the same trace. Reuse the existing metadata extraction or scan all events and retain the last valid name.

packages/agency-lang/lib/logsViewer/help.ts
```diff
@@ -23,7 +23,7 @@ export const HELP_GROUPS: BindingGroup[] = [
   {
     heading: "Expand",
     bindings: [
-      { keys: "l / Right / Enter", action: "expand — leaves inline their JSON payload" },
+      { keys: "Right / Enter", action: "expand — leaves inline their JSON payload" },
```
egonSchiele, line 26: [P3] Removing `l` from shared help hides an active binding when promotion is disabled (stdin/remote entry points still route `l` to the tree reducer and expand). Please make tree help conditional: advertise `l` as promotion when enabled and as expand when disabled.

egonSchiele:
Thanks — worked through the review. Fixes in `f6d2673c1`, CLI/guide docs in `d28a3e457`.

**P1 — `_print` double-inspection + redaction (`builtins.ts`):** now renders once (`const rendered = format(...); console.log(rendered); recordPrint("print", rendered)`), so a custom `util.inspect.custom` hook runs a single time and any throw stays before the console write. Added a stateful-hook test asserting one inspection. On redaction: the rendered string still flows through `post()`, which scrubs tagged **substrings** (the common `redact("secret")` case is preserved). A `redact()` applied to a whole **object** with no tagged string fields is the residual — inherent to representing a printed object as text (the console already showed it). Fully preserving that would require carrying the structured value through to read-time, which changes the event schema + resolver; flagging rather than doing silently — happy to take that on if you'd like it.

**P1 — promotion from a partially parsed statelog (`run.ts`):** `l` now refuses when the current parse has any errors (strict for ingestion, unlike display) with an actionable notice; follow-mode reparses update the count.

**P1 — follow watcher during handoff (`run.ts`):** the watcher is stopped for the labeling handoff and restarted after, so an append can't repaint over the labeling UI. The append reader replays everything written during the pause on the first poll after resume.

**P2 — unrequested `--output` selector (`statelog.ts`):** now rejected (was silently ignored), listing the requested traces.

**P2 — large first envelope misclassified (`format.ts`):** the sniff reads through the first newline (byte-correct, 16 MiB cap) instead of a fixed 64 KiB prefix, so a >64 KiB first event still classifies.

**P2 — stale source name (`run.ts`):** viewer promotion now uses the **last** `agentName`, matching `scanStatelog`/the extractor.

**P3 — empty trace id + `l` help (`treeView.ts` / `help.ts`):** `l` no longer emits a promote action without a real trace id; tree help is contextual (`l` = label when a dataset is configured, expand otherwise); the shared help table shows `Right`/`Enter` for expand.

packages/agency-lang/docs/site/guide/observability.md
```diff
@@ -61,14 +61,16 @@ Press `?` in the viewer to see this info.
 |---|---|
 | `j`, `Down`, `Ctrl+N` | Move cursor down |
 | `k`, `Up`, `Ctrl+P` | Move cursor up |
-| `l`, `Right`, `Enter` | Expand the focused node — on a span/trace, reveal children; on a leaf, inline the JSON payload |
+| `Right`, `Enter` | Expand the focused node — on a span/trace, reveal children; on a leaf, inline the JSON payload |
+| `l` | Expand the focused node — unless a dataset is configured, where it instead **labels the focused trace** (see below) |
```
egonSchiele: `l` should never expand the focused node.

packages/agency-lang/docs/site/guide/observability.md
```diff
@@ -79,6 +81,13 @@ Press `?` in the viewer to see this info.
 | `?` | Show / hide the keybinding help |
 | `q`, `Ctrl+C` | Quit |
 
+When you view a local statelog and pass a checklist, `l` promotes the focused
```
egonSchiele: remove the paragraph you added here

packages/agency-lang/lib/cli/eval/ingest.ts
```diff
@@ -91,6 +96,49 @@ export function parseFieldArgs(options: {
...
 }
 
+/**
+ * Parse repeatable `--output <trace-id>=print:<index>` selectors into a map.
+ *
+ * A selector applies to exactly one trace; two for the same trace, a missing
+ * `=`, or a right side that is not `print:<n>` are all errors rather than a
+ * silent last-wins, because the choice decides which output is labeled.
+ */
+export function parseTracePrintSelections(values: readonly string[]): TracePrintSelections {
```
egonSchiele: what is this?

egonSchiele: Why do users need `print:`?

packages/agency-lang/lib/cli/eval/label.ts
```diff
@@ -16,19 +14,26 @@ const FALLBACK_ANNOTATOR_ID = "human";
...
 export type EvalLabelOptions = {
   checklist?: string;
+  dataset?: string;
   store?: string;
   annotator?: string;
   config?: AgencyConfig;
 };
 
+/** The dataset directory can be named two ways at each layer: the preferred
+ *  `--dataset`/`eval.dataset` and the deprecated `--store`/`eval.labelStore`. */
```
egonSchiele: don't deprecate, just remove!

packages/agency-lang/lib/cli/eval/labelCommand.ts
```diff
@@ -58,11 +58,12 @@ export function addLabelCommand(
 ): Command {
   const label = parent
     .command("label")
-    .description("Label the records in a label store. Add records with `label ingest` first")
-    .option("--store <dir>", "Label store directory (default: eval.labelStore, else labels/)")
+    .description("Label the examples in a dataset. Add examples with `label ingest` first")
+    .option("--dataset <dir>", "Label dataset directory (default: eval.dataset, else labels/)")
+    .option("--store <dir>", "Deprecated alias for --dataset")
```
egonSchiele: remove, don't deprecate

packages/agency-lang/lib/cli/logsView.ts
```diff
@@ -76,9 +96,29 @@ export async function logsView(
   await explorer({
     sources: discovery.sources,
     route: discovery.route === "runTable" ? "runTable" : "explorer",
+    promotion: promotionBaseFrom(cliOpts),
   });
 }
 
+/** The dataset/checklist/annotator a promotion targets, shared by the single-
```
egonSchiele: What does the term "promotion" mean here?

packages/agency-lang/lib/eval/label/load/statelog.ts
```diff
@@ -0,0 +1,307 @@
...
+  const values = record.evalValues ?? [];
+  const last = values.at(-1);
+  if (last === undefined) {
+    return null;
+  }
+  const parsed = JsonValueSchema.safeParse(last.value);
+  return parsed.success ? parsed.data : null;
+}
+
+function hasExplicitEvalOutput(events: readonly EventEnvelope[]): boolean {
```
egonSchiele: Some of this does not feel specific to the statelog loader. How much code duplication does this code introduce?

packages/agency-lang/lib/eval/label/load/statelogScan.ts
```diff
@@ -0,0 +1,93 @@
...
+};
+
+/** A statelog parsed and grouped by trace, in first-seen order. */
+export type StatelogScan = {
+  eventsByTrace: Readonly<Record<string, readonly EventEnvelope[]>>;
+  traces: readonly TraceSummary[];
+};
+
+/** How much of a first message to show when listing available traces. */
+const TRACE_PREVIEW_CHARS = 60;
```
egonSchiele: How much code duplication is there between this file and packages/agency-lang/lib/eval/label/load/statelog.ts?

packages/agency-lang/lib/eval/label/controller.ts
```diff
@@ -186,7 +196,7 @@ function syncChecklistDefinitionIds(checklistFile: string, definition: Normalize
 type SessionConstruction = {
   args: OpenLabelingSessionArgs;
   dependencies: ControllerDependencies;
-  store: LabelStore;
+  store: LabelDataset;
```
egonSchiele: should this field be relabeled as `dataset`?

egonSchiele:
Reworked per the review — the PR was doing more than it needed to. Headline: **dropped print-value selection entirely**, which cascades away most of the weight. Net −688 lines on that commit alone; all touched-area tests green, lint/make clean.

Per comment:

- **"Why do users need `print:`?" / "what is this?"** — removed. Print capture (the whole `print` statelog event), `--output`, the chooser, the `StatelogOutputSource` union, and the manifest v2→v3 bump are all gone. Output is now just `evalOutput()` → return value — exactly what the extractor and run-loader already do. (This also moots the earlier redaction P1, since nothing new is captured.)
- **"code duplication?" (statelog.ts / statelogScan.ts)** — extracted the shared `selectLabelingFinalOutput → project → eligibility → fields` step into `load/occurrence.ts` (`projectOccurrenceFields`, `skipReasonFor`), now used by **both** `loadRun` and the statelog loader. A promoted trace is just an extracted example; the only statelog-specific bit left is the tiny `statelog` occurrence origin `{ traceId, finalOutputIndex }` (the `run` origin requires an `inputId` a raw trace lacks).
- **"remove, don't deprecate" (×2)** — `--store` and `eval.labelStore` and all the alias/conflict machinery are deleted. `resolveDataset` is `dataset ?? config ?? labels`.
- **`store` → `dataset` (controller.ts:199)** — finished the internal rename that was half-done (~189 identifiers: `storeDir`→`datasetDir`, `StoreLock`→`DatasetLock`, the `store` field, etc.).
- **"What does 'promotion' mean?"** — renamed the concept to **label a trace** throughout (`promoteTrace.ts`→`labelTrace.ts`, `Promotion*`→`LabelTrace*`, the `promoteTrace` action→`labelTrace`).
- **"`l` should never expand"** — removed `l` from the tree expand reducer. `l` only labels (when a dataset is configured); `Right`/`Enter` expand.
- **"remove the paragraph"** — removed.
- **`60` in multiple places** — one `PREVIEW_CHARS` + `previewLine` helper.

CLI/guide docs updated to match (no `--output`/`--store`/print).

egonSchiele (commented):
Re-review of current head `6f4dc44` found the inline issues above. Two additional merge blockers are outside changed lines:

- The integration check is red because `tests/integration/cli/test.mjs:330` still invokes the intentionally removed `--store`; `tests/integration/cli-tier2/test.mjs:416,466` have the same stale option and will fail when that tier runs. Convert these coverage paths to the parent-owned `--dataset` option.
- The PR description is now materially false (it still claims print events, `--output`, aliases, manifest v3, 13 commits, deferred site docs, and a green suite). `docs/dev/eval-labeling.md:19-42` likewise still instructs users to use `--store`/`eval.labelStore` immediately after introducing the new vocabulary. Please update both to describe the simplified current design.

The earlier print/redaction, keyed-selector, parse gate, follow-screen, agent-name, and help findings are otherwise addressed or no longer applicable.

packages/agency-lang/lib/eval/label/load/statelog.ts
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * Extract one trace and choose its output: an `evalOutput()` value, else the
+ * entry node's return value (via the same selection helper runs use). A
+ * truncated or absent output is rejected — the last chat message is never a
+ * stand-in. Pure: no prompting, no writes.
+ */
+export function resolveTrace(events: readonly EventEnvelope[], sourcePath: string): TraceResolution {
+  const record = extractEvalRecord(events as EventEnvelope[], sourcePath);
+  const selection = selectLabelingFinalOutput(record);
```
egonSchiele, line 65: [P1] This can select an `evalOutputRecorded` value from a trace that later crashed or never completed. The runtime emits `error` followed by a result-less `agentEnd` on an unhandled failure, but extraction preserves earlier explicit outputs, so this path stores a failed execution even though run-directory ingestion rejects `status === failed` as `run-failed`. Please require a successful terminal execution signal before selecting the output (while distinguishing transient/retried LLM/tool errors from the terminal runtime-error + result-less `agentEnd` case), and reject unfinished traces as well. Add a fixture with `evalOutputRecorded`, then terminal `runtimeError`/result-less `agentEnd`.

packages/agency-lang/lib/logsViewer/run.ts
```diff
@@ -87,6 +215,7 @@ export async function runViewer(opts: RunViewerOpts): Promise<ViewerResolution>
     const rebuilt = buildForest(reparsed.events);
     if (rebuilt.length === 0) return;
```
egonSchiele: [P1] The early return leaves `allEvents` and `parseErrors` from the previous file contents. In follow mode, truncate/rotate a valid file to malformed or empty content: `rebuilt` is empty (and an empty truncation currently triggers no callback at all), so the old trace remains selectable with zero remembered parse errors and `l` can label data no longer present in the file. Update the parsed state before this branch, clear/update views for an empty forest, and have the watcher notify on truncation even when the new accumulator is empty.

packages/agency-lang/lib/eval/label/load/format.ts
```diff
@@ -47,12 +49,63 @@ export function resolveFormat(
...
+      const bytesRead = fs.readSync(fd, chunk, 0, chunk.length, offset);
+      if (bytesRead === 0) {
+        break; // EOF before any newline
+      }
+      offset += bytesRead;
+      const newlineAt = chunk.subarray(0, bytesRead).indexOf(0x0a);
+      const end = newlineAt === -1 ? bytesRead : newlineAt;
+      collected.push(Buffer.from(chunk.subarray(0, end)));
+      total += end;
+      newlineFound = newlineAt !== -1;
```
egonSchiele: [P2] This stops at the first newline rather than the first *non-empty* line. A valid JSONL statelog beginning with `\n` now returns `undefined` and auto-detection fails, even though the parser accepts blank lines. Continuing is not quite enough because the current read advances `offset` past the rest of the chunk; preserve bytes after each newline and keep reading until a nonblank line (within the cap). Please add leading-blank-line coverage.

packages/agency-lang/lib/eval/label/load/index.ts
```diff
@@ -95,6 +102,28 @@ export function loadBatch(
   // record, and its id, from what the arguments said.
   assertNoLoaderCollision(format, request.constantFields, request.source.includeTaskField);
 
+  // Trace selection only means something for a statelog source. Rejecting it
+  // elsewhere turns "I put --trace on the wrong kind of source" into a clear
+  // error instead of a silently ignored flag.
+  if (format !== "statelog" && request.selection.kind === "statelog") {
+    throw new IngestSourceError(
+      `--trace and --output only apply to a statelog source, but ${request.source.path} is a ` +
```
egonSchiele: [P3] `--output` was removed in the simplification, but this new diagnostic still tells users it exists. Please make this `--trace only applies ...` and update the assertion.

packages/agency-lang/docs/site/cli/logs.md
```diff
@@ -28,6 +28,38 @@ timeline views (`t`), search, and follow mode. `agency logs view <file>`
 is the same thing spelled explicitly, and `-` reads from stdin. See the
 viewer's own help (`?`) for keys.
 
+## Labeling a trace from the viewer
+
+When you view a **local** statelog file, the tree's `l` key promotes the
```
egonSchiele: [P2] This section still describes both behaviors removed by the latest commits: the action is no longer called promotion, and `l` no longer expands when labeling is unavailable (it is a no-op; Right/Enter are the expansion keys). Please rewrite the paragraph to match the current UI before publishing these user-facing docs.

packages/agency-lang/lib/eval/label/load/statelog.ts
egonSchiele: Fixed in 243ff32. `resolveTrace` now gates on a terminal outcome before selecting an output: a terminal `runtimeError` followed by a result-less `agentEnd` is rejected as `run-failed` (same as run-directory ingestion rejects `status === failed`); a trace with no `agentEnd` at all is `trace-unfinished`. A result-less `agentEnd` with no `runtimeError` (an `evalOutput()`-only run) still labels, and transient `llmError`/`toolError` on an otherwise-completed run do not reject. Added fixtures for the crash case, the recovered-transient case, and the unfinished case.

packages/agency-lang/lib/eval/label/load/format.ts
egonSchiele: Fixed in 243ff32. `firstNonEmptyLine` now keeps reading past blank lines to the first non-empty line (preserving bytes after each newline, within the 16 MiB cap) instead of stopping at the first newline. Added coverage for a leading blank line and for a large first event that follows a blank line.

packages/agency-lang/lib/eval/label/load/index.ts
egonSchiele: Fixed in 243ff32. The diagnostic now reads `--trace only applies to a statelog source`, and the assertion in index.test.ts matches.

packages/agency-lang/docs/site/cli/logs.md
egonSchiele: Fixed in 243ff32. Rewrote the paragraph: `l` labels the trace (not "promotes"), `Right`/`Enter` expand, and `l` never expands — where labeling is unavailable (stdin/remote) it does nothing.

packages/agency-lang/lib/logsViewer/run.ts
egonSchiele: Fixed in 243ff32, in four parts:
1. `onNewText` now updates `roots`/`allEvents`/`parseErrors` unconditionally — the early `return` on an empty forest is gone, so a truncation to empty/malformed content is applied rather than leaving the previous file's trace in place.
2. The follow watcher now notifies on a shrink even when the new accumulator is empty (a chunk-only trigger missed this).
3. `TreeView.setData` reflects an empty forest (cursor falls back to `""`), so `cursorTraceId()` and the `l` action no longer point at a gone trace.
4. Belt-and-suspenders: the label handler refuses when the focused trace's events are no longer present. Added a follow-mode test that truncates to empty and asserts the old trace clears.

packages/agency-lang/lib/cli/eval/labelCommand.ts
```diff
@@ -58,11 +58,11 @@ export function addLabelCommand(
...
   const label = parent
     .command("label")
-    .description("Label the records in a label store. Add records with `label ingest` first")
-    .option("--store <dir>", "Label store directory (default: eval.labelStore, else labels/)")
+    .description("Label the examples in a dataset. Add examples with `label ingest` first")
+    .option("--dataset <dir>", "Label dataset directory (default: eval.dataset, else labels/)")
     .option("--checklist <file>", "Checklist JSON: an existing one, or { name, questions }")
     .option("--annotator <id>", "Who is labelling (default: $USER)")
-    .action(async (opts: { checklist?: string; store?: string; annotator?: string }) => {
+    .action(async (opts: { checklist?: string; dataset?: string; annotator?: string }) => {
```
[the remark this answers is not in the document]
egonSchiele: The code is intentional — this is a clean break, not an alias. Per the maintainer's directive (no users, no datasets on disk), `--store` and `eval.labelStore` were removed outright rather than deprecated; only `--dataset` / `eval.dataset` remain. The PR description was stale on this point — updating it now. The integration tests that still used `--store` (cli/test.mjs, cli-tier2/test.mjs) were updated to `--dataset` in 6576bed.

packages/agency-lang/lib/logsViewer/run.ts
egonSchiele: Already handled in the current code: `handleTraceLabeling` refuses when `parseErrorCount > 0` and notifies (`Cannot label: the statelog has N unparseable line(s)`), so `l` never persists from a partially parsed trace. Because `parseErrors` is recomputed on every follow-mode reparse, errors discovered by a follow reparse are covered too.

packages/agency-lang/lib/logsViewer/run.ts
egonSchiele: Already handled: `handleTraceLabeling` stops the follow watcher before the labeling handoff and restarts it in a `finally`, so an append during `nextLine`/labeling cannot repaint over the prompt. The new empty-truncation test in followMode.test.ts exercises the watcher across the pause.

packages/agency-lang/lib/logsViewer/run.ts
egonSchiele: Already fixed: `traceSourceName` iterates all events and keeps the LAST `agentName`, matching `scanStatelog` and the extractor.

packages/agency-lang/lib/eval/label/load/statelog.ts
egonSchiele: Moot after the simplification: print-value selection was removed, so there are no keyed selectors to validate. The remaining selector is `--trace`, and `loadStatelog` validates every requested trace id up front — an unknown id throws `IngestSourceError` naming the available traces (see the "errors naming available traces for an unknown id" test).

packages/agency-lang/lib/logsViewer/help.ts
egonSchiele: This suggestion is superseded by a later maintainer directive: `l` never expands. It was removed from the tree reducer, so when labeling is disabled `l` is a no-op, not an expand key — advertising it as expand would be wrong. `Right`/`Enter` are the expansion keys and are documented as such. The tree help is conditional: it lists `l — label this trace` only when a dataset is configured, and omits it otherwise (since it does nothing there). So no active binding is hidden.

packages/agency-lang/lib/logsViewer/views/treeView.ts
```diff
@@ -113,6 +122,7 @@ export class TreeView implements View {
     return [
       "t — timeline views (flame → by-name)",
       "d — full details of the focused span",
+      ...(this.promotionEnabled ? ["l — label this trace (promote into the dataset)"] : []),
       "",
       ...treeHelpLines(),
     ];
```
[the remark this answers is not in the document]
egonSchiele: Superseded: `l` no longer expands in any mode (removed from the tree reducer per the maintainer's directive). When labeling is disabled `l` is a no-op, so the help correctly omits it; `Right`/`Enter` remain the documented expansion keys.

packages/agency-lang/lib/logsViewer/views/treeView.ts
```diff
@@ -51,6 +55,11 @@ export class TreeView implements View {
       const id = this.cursorRealId();
       if (id !== undefined) return { kind: "openDetail", spanId: id };
     }
+    // `l` promotes the focused trace when a dataset is configured. Right/Enter
+    // still expand, so no expand key is lost.
+    if (fmt === "l" && this.promotionEnabled) {
+      return { kind: "promoteTrace", traceId: this.cursorTraceId() };
+    }
```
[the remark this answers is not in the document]
egonSchiele: Already guarded: the tree only emits `labelTrace` when `cursorTraceId()` is non-empty, so an empty forest (e.g. an empty file under --follow) produces no action. The 243ff32 follow-mode changes also keep the cursor from pointing at a gone trace after a truncation.

packages/agency-lang/lib/eval/label/load/format.ts
egonSchiele: Fixed in 243ff32. `firstNonEmptyLine` reads through the terminating newline of the first non-empty line, bounded by a 16 MiB cap, so a first event larger than the 64 KiB sniff chunk is classified from the complete line rather than a truncated prefix. An over-limit first line falls through to the unknown-format error rather than being mis-parsed. Test: "classifies a statelog whose first event is larger than one read chunk".

packages/agency-lang/lib/stdlib/builtins.ts
egonSchiele: Moot after the simplification: the print change was reverted entirely. `_print`/`_printJSON` are back to plain `console.log`, and nothing is captured to statelog, so there is no formatting-before-redaction path and no double inspection. This also removed the redaction concern you raised.

packages/agency-lang/lib/eval/label/load/statelog.ts
egonSchiele: Agreed — the shared parts were extracted. Output selection is `selectLabelingFinalOutput` (shared with `loadRun`, in run.ts) and field projection is `projectOccurrenceFields`/`skipReasonFor` in load/occurrence.ts, used by BOTH `loadRun` and the statelog loader. What remains statelog-specific in this file is only: extracting an eval record from raw events (a run has a pre-extracted `eval-record.json`), the terminal-outcome gate, the task-default heuristic, and building the `statelog` origin. No projection/eligibility logic is duplicated.

packages/agency-lang/lib/eval/label/load/statelogScan.ts
egonSchiele: They do different jobs with no overlapping logic. statelogScan groups raw events by trace id and builds the human-facing trace list (ids, previews) for the "which --trace?" error and the viewer; it never extracts or projects. statelog.ts consumes one trace's events and turns them into an occurrence (extract → select output → project). The loader calls `scanStatelog` once and validates requested ids against it; the shared preview constant (`PREVIEW_CHARS`) lives in statelogScan and is imported, not copied.

packages/agency-lang/lib/eval/label/controller.ts
egonSchiele: Done — the field is now `dataset: LabelDataset` (this was part of finishing the store→dataset rename in the round-2 commit).

packages/agency-lang/lib/eval/label/load/statelog.ts
```diff
@@ -0,0 +1,198 @@
...
+function terminalOutcome(events: readonly EventEnvelope[]): TerminalOutcome {
+  const ends = events.filter((e) => e.data.type === "agentEnd");
+  const last = ends.at(-1);
+  if (last === undefined) {
+    return "unfinished";
+  }
+  if (last.data.result !== undefined && last.data.result !== null) {
+    return "ok";
+  }
+  const crashed = events.some((e) => e.data.type === "error" && e.data.errorType === "runtimeError");
```
egonSchiele, line 96: [P1] `runtimeError` is not uniquely a terminal crash marker. Generated function catches emit `runtimeError` before converting an exception to a recoverable `Failure` (and a failed tool therefore emits both `runtimeError` and `toolError`); the checked-in `tests/agency-js/tool-tiers/log.jsonl` demonstrates that ordering. A run can recover, call `evalOutput()`, and finish with a null/result-less `agentEnd`, but this trace-wide `some()` still returns `run-failed`. The new transient test includes only `toolError`, so it misses the real wire sequence. Please identify the terminal runtime error specifically (for example, the top-level result-less footer's immediately preceding unscoped runtime error) rather than treating every earlier function/tool runtime error as terminal, and add the recoverable runtimeError + evalOutput-only case.

packages/agency-lang/lib/logsViewer/run.ts
```diff
@@ -224,7 +372,8 @@ function makeFollowWatcher(
   const poll = (): void => {
     if (reader === undefined || opts.followPath === undefined) return;
     const size = currentFileSize(opts.followPath);
-    if (size < lastSize) {
+    const truncated = size < lastSize;
```
egonSchiele, line 375: [P1] Size-only replacement detection still allows stale data to be labeled. If the file is truncated/rotated and rewritten to exactly the old size between polls (especially while the watcher is paused for labeling), this detects no change and keeps the old trace forever; if rewritten larger, the reader appends only bytes after the old offset to the old accumulator. The parse gate then applies to stale/corrupted state rather than the current file. Detect replacement independently of shrink (inode/identity or prefix validation), and reconcile the file when follow resumes; add same-size and larger-replacement cases.

packages/agency-lang/lib/logsViewer/labelTrace.ts
```diff
@@ -0,0 +1,96 @@
...
+  });
+  if (projected.kind === "skipped") {
+    ui.notify(`Nothing to label for this trace: ${projected.skip.reason}.`);
+    return { kind: "rejected", reason: projected.skip.reason };
+  }
+
+  // The id the dataset will derive for these fields, so the labeling session can
+  // open focused on the example just written.
+  const outputId = makeOutputId(projected.occurrence.fields);
+  services.datasetWriter.ingest({
```
egonSchiele, line 78: [P2] The dataset is mutated before the checklist is opened or validated. With a missing/malformed/incompatible `--checklist`, ingest succeeds here, then `labelingHost.run()` throws; the viewer exits even though the requested label operation failed, leaving an unlabeled occurrence behind. The undefined-checklist case correctly guarantees no write, so invalid checklist input should get the same preflight behavior. Validate the checklist/session prerequisites before ingest (or otherwise make the handoff transactional), and cover a malformed checklist with an unchanged dataset assertion.

egonSchiele (commented):
The round-3 findings are fixed and all current CI checks are green. This pass found three remaining edge cases in terminal-outcome classification, follow-mode replacement detection, and checklist preflight; details are inline.

PR #850, opened 2026-08-16
Fix two red CI jobs: stale action fixture and an unpublished package pin

egonSchiele:
Two CI jobs have been red on main for reasons unrelated to any recent change. This fixes both, and adds the missing step to each doc that would have caught the drift.

## The `integration` job: a fixture nobody updated

The job generates a workflow file with `agency schedule add --backend github` and compares it to a checked-in fixture. One line differed:

```
actual:   - uses: egonSchiele/run-agency-action@v1.0.3
expected: - uses: egonSchiele/run-agency-action@v1.0.2
```

Commit 077e18328 bumped the action to v1.0.3 and updated the generator snapshots under `lib/cli/schedule/backends/__snapshots__/`, but left the three fixtures under `tests/integration/cli-main/fixtures/expected/` on v1.0.2. Everything else was bumped correctly: `pinnedActions.ts`, the makefile tag list, the unit snapshots, and the three live scheduled workflows.

## The `local-model-tests` job: a version that was never published

The job pins `SMOLTALK_LLAMA_CPP_VERSION: "0.5.2"` and fails at install:

```
[ERR_PNPM_NO_MATCHING_VERSION] No matching version found for smoltalk-llama-cpp@0.5.2
The latest release of smoltalk-llama-cpp is "0.4.0".
```

`npm view smoltalk-llama-cpp versions` returns `0.1.0, 0.1.1, 0.2.0, 0.3.0, 0.4.0` — 0.5.2 was never published, so this pin has been unresolvable since it was added in d9df2e51e back in June. Moved to 0.4.0.

## The docs

Both drifts were possible because the written procedure did not mention the file that went stale.

`updating-pinned-actions.md` walks through regenerating the generator snapshots but never mentions the integration fixtures, and its verify step (`pnpm test:run lib/cli/schedule`) does not touch them. Those fixtures only run in the `integration` workflow, so a stale one passes every local check and fails after the merge. Added a step for them, and listed them in the commit file list.

`local-model-integration.md` calls the workflow env var the single source of truth for the smoltalk version, then hardcoded 0.5.2 in its own local-setup instructions. Fixed the version and pointed the reader at the workflow.

## Verification

- The fixtures now match the generator's own checked-in snapshot exactly. `diff` of the `uses:` lines in `nightly-write.expected.yml` against `__snapshots__/all-tag.yml` is empty, on both `actions/checkout@v4.1.7` and `run-agency-action@v1.0.3`. That snapshot is regenerated from `pinnedActions.ts`, so fixture and generator now agree.
- npm confirms 0.4.0 is the latest published version.

Two limits worth stating. I did not run the full integration suite locally — it packs and installs the package, which is the slow CI path. And if anything in `tests/integration/local-model/` depends on a smoltalk API added after 0.4.0, that job will now get past `pnpm add` and fail later instead; this PR's own CI run is the check for that.

### DIFF
--- a/.github/workflows/local-model.yml
+++ b/.github/workflows/local-model.yml
@@ -35,7 +35,7 @@ jobs:
     env:
       # Bump this in lockstep with smoltalk-llama-cpp upstream releases.
       # node-llama-cpp's version is pinned transitively via smoltalk-llama-cpp's lockfile.
-      SMOLTALK_LLAMA_CPP_VERSION: "0.5.2"
+      SMOLTALK_LLAMA_CPP_VERSION: "0.4.0"
 
     steps:
       - name: Checkout repository
--- a/packages/agency-lang/docs/dev/local-model-integration.md
+++ b/packages/agency-lang/docs/dev/local-model-integration.md
@@ -39,10 +39,14 @@ cache and finish in seconds.
 If you change the curated `smollm2-135m` URI in `lib/stdlib/localModels.ts`,
 update **two** values:
 
-1. `EXPECTED_SHA256` in `tests/integration/local-model/smoltest.test.ts` —
-   currently `null` (format-only check). The first green integration run
-   prints the actual hash; paste it in to enable strict tamper-canary
-   matching on subsequent runs.
+1. `EXPECTED_SHA256` in `tests/integration/local-model/smoltest.test.ts`. It
+   holds the hash of the file the curated URI points at, and the test compares
+   the download against it byte for byte, so a stale value fails the run.
+   Recapture it from Hugging Face's LFS metadata for the new file (the git-LFS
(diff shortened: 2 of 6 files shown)


egonSchiele:
Good catch — verified and fixed in cf38d7986.

`EXPECTED_SHA256` at smoltest.test.ts:28 does hold a concrete hash (`ed5fa30c…`), so the doc was describing a state the file has not been in since the strict check was turned on. It also understated the consequence: with a real value set, a stale hash fails the run rather than simply being absent.

The step now says where to recapture the hash (Hugging Face LFS metadata, whose git-LFS oid is the sha256 of the content, or the line the test logs), and describes `null` as a way to recapture a hash you do not have rather than a default state.

PR #851, opened 2026-08-17
Polish the label-a-trace flow: full-screen task prompt + shortened trace ids

egonSchiele:
Polish to the label-a-trace flow, in two independent commits.

## 1. Task step: a clear full-screen prompt

Pressing `l` in the log viewer to label a trace first asks you to set the **task** — the prompt the output should be judged against. That step used a one-line readline prompt painted on top of the tree view, so it looked broken and gave no sign it wanted input:

```
Task [Please fetch the latest news for today, 2026-08-15, for the …] — Enter keeps, text replaces, "-" clears:
```

It is now a dedicated full-screen page:

```
 Set the task

 This output will be judged against a task. We filled in the task this
 run recorded, shown gray — edit it to change it, or press Enter to
 accept it.

 Task
   Write a 10 word story about a cat and a dog.▏

 Enter accept  ·  edit to change  ·  empty = no task  ·  Esc cancel
```

One editable field, pre-filled with the recorded task and shown **gray** the way a web form shows a default value. Press Enter to accept it, or edit it in place — the moment you edit, the text turns **white**, so it is obvious it is now yours. Untouched keeps the recorded task; edited-to-text replaces it; cleared out judges with no task; Esc backs out. Built on the same `runLoop`/Element machinery the labeling TUI uses, so the frame clears the tree view instead of drawing over it. Recorded task and draft are sanitized as untrusted model output.

## 2. Ingest: accept a shortened trace id, and label the trace list

`agency label ingest <statelog> --trace <id>` required the full trace id. But the log viewer shows only the first several characters — so a user copies `KQYcet` and gets `Trace "KQYcet" is not in log.jsonl`, which reads as "no such trace" even though it is right there. Ingest now resolves a `--trace` value that is a **prefix** to its full id (exact id always wins; a prefix that singles out one trace matches; a shared prefix is rejected with the candidates listed). The full id is what the occurrence records, so identity stays stable.

The "Available traces" listing was also cryptic — `rpGX…  [(unnamed)]  $0.0000  …` with no headers. It now has column headers, aligned columns, spells an agent-less trace as `(unnamed)` (not `[(unnamed)]`), and says a full id or any unique prefix works:

```
Available traces (pass a full id or any unique prefix to --trace):
  TRACE ID               AGENT         COST     TASK (first user message)
  rpGXbww2gfHaZDLBL8Z2H  (unnamed)     $0.0012  Write a 10 word story about a cat and a dog.
  KQYcet                 news-fetcher  $0.0300  Please fetch the latest news for today, 2026-08-15, for the …
  GY1HnAWAGIV4odc5aMq26  (unnamed)     $0.0000
```

## Tests

`taskEditor.test.ts` drives the editor through a real `Screen` + `ScriptedInput` + `FrameRecorder` (pre-fill/keep/replace/clear/type-when-none/cancel, plus gray-vs-white). `matchTraceId` and the prefix/ambiguity paths of `loadStatelog` are unit-tested, and `describeAvailableTraces` asserts the headers. Full `lib/eval/label` + `lib/logsViewer` suites green. tsc, `lint:structure`, and prettier clean.

### DIFF
--- a/packages/agency-lang/lib/logsViewer/taskEditor.ts
+++ b/packages/agency-lang/lib/logsViewer/taskEditor.ts
@@ -36,22 +36,15 @@ function defaultDraft(defaultTask: JsonValue | null): string {
 /**
  * The field as display lines, wrapped to the width. Gray (a pre-filled default
  * you can accept) until touched, plain white (your own text) after. The cursor
- * sits at the end. An untouched empty field shows a gray typing hint instead,
- * since there is nothing to accept.
+ * sits at the end. An empty, untouched field shows a gray typing hint so it
+ * never looks blank — whether the run recorded no task or recorded an empty one.
  *
  * Exported so a test can assert the gray-then-white behavior directly.
  */
-export function fieldLines(
-  draft: string,
-  touched: boolean,
-  hasDefault: boolean,
-  width: number,
-): string[] {
+export function fieldLines(draft: string, touched: boolean, width: number): string[] {
   const inner = Math.max(8, width - 2);
   if (!touched && draft === "") {
-    const hint = hasDefault
-      ? ""
-      : "type a task to judge this output against, or press Enter to skip";
+    const hint = "type a task to judge this output against, or press Enter to skip";
     return [`  ${CURSOR}${color.brightBlack(hint)}`];
   }
   const wrapped = wrapText(sanitizeUntrusted(draft), inner);
@@ -61,13 +54,28 @@ export function fieldLines(
   return out;
 }
 
+/** Keep the field within `budget` display rows, showing the tail so the cursor
+ *  (always at the end) stays visible and the footer below is never pushed off
+ *  screen for a task that wraps very long. A hidden head is marked with a
+ *  leading ellipsis. */
+function clampField(all: string[], budget: number): string[] {
+  if (all.length <= budget) return all;
+  if (budget <= 1) return [all[all.length - 1]];
+  return [`  ${color.brightBlack("⋯")}`, ...all.slice(all.length - (budget - 1))];
+}
+
 function explanation(hasDefault: boolean): string {
   return hasDefault
     ? "This output will be judged against a task. We filled in the task this run recorded, shown gray — edit it to change it, or press Enter to accept it."
     : "This output will be judged against a task, but this run recorded none. Type one below, or press Enter to skip and judge without a task.";
 }
 
-function editorScreen(state: EditorState, hasDefault: boolean, width: number): Element {
+function editorScreen(
+  state: EditorState,
+  hasDefault: boolean,
+  width: number,
+  height: number,
+): Element {
   const footer = [
     `${color.cyan("Enter")} accept`,
     `${color.cyan("edit")} to change`,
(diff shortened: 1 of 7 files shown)


packages/agency-lang/lib/logsViewer/taskEditor.ts
```diff
@@ -0,0 +1,130 @@
...
+  ].join(color.brightBlack("  ·  "));
+  const prose = wrapText(explanation(hasDefault), Math.max(8, width - 2));
+  return column(
+    { justifyContent: "flex-start", padding: 1 },
+    line(color.bold("Set the task")),
+    line(""),
+    ...prose.map((l) => line(color.brightBlack(l))),
+    line(""),
+    line(color.bold.yellow("Task")),
+    ...fieldLines(state.draft, state.touched, hasDefault, width).map((l) => line(l)),
```
egonSchiele: [P2] Keep the cursor and controls visible for long tasks

Every wrapped field line is inserted as a fixed-height child before the footer. Once a recorded prompt wraps past the terminal height, layout clips the trailing field lines, cursor, and footer. Since editing only happens at the end of `draft`, typing and backspacing then occur entirely off-screen with no visible feedback. This is fairly easy to hit with a long agent prompt. Please give the field a bounded/scrolling viewport (or otherwise render a trailing window) while keeping the footer visible; a small-height test with a long default would catch the regression.

packages/agency-lang/lib/logsViewer/taskEditor.ts
```diff
@@ -0,0 +1,130 @@
...
+  defaultTask: JsonValue | null,
+): Promise<TaskChoice | null> {
+  const hasDefault = defaultTask !== null;
+  const start = defaultDraft(defaultTask);
+  const final = await screen.runLoop<EditorState>({
+    initialState: { draft: start, touched: false, done: false, result: null },
+    render: (state) => editorScreen(state, hasDefault, screen.size().width),
+    handleKey: (state, event) => {
+      if (event.key === "escape") return { ...state, done: true, result: null };
+      if (event.key === "enter") return { ...state, done: true, result: finalize(state) };
```
egonSchiele, line 135: [P2] Preserve the TUI's modified-Enter newline contract

`TerminalInput` deliberately emits `{ key: "enter", shift: true }` for Shift+Enter/Option+Enter so editable buffers can insert a newline (`ui-smoke.test.ts` covers that contract). This branch ignores `shift` and submits immediately, so trying to enter a multi-line task silently accepts an incomplete draft. The field already renders and accepts pasted newlines, so modified Enter should append `\n`; only unmodified Enter should finalize. Please add a scripted-input case for modified Enter followed by normal Enter.

egonSchiele (commented):
Two user-facing edge cases in the new task editor need attention.

packages/agency-lang/lib/logsViewer/taskEditor.ts
egonSchiele: Fixed in f7cc9207d. The field is now windowed to its tail within the rows left after the fixed chrome (title/explanation/heading + footer), with a leading `⋯` marking a hidden head, so the cursor (always at the end) and the footer stay visible however long the task wraps. Added a small-height test (70×16) with a 40-sentence default asserting the cursor, `Esc cancel`, and the `⋯` marker are all in the final frame.

packages/agency-lang/lib/logsViewer/taskEditor.ts
egonSchiele: Fixed in f7cc9207d. Shift/Option+Enter (`{ key: "enter", shift: true }`) now appends `\n` and marks the field touched; only an unmodified Enter finalizes. Added a scripted case: `line one` + modified Enter + `line two` + plain Enter → `replace "line one\nline two"`.

egonSchiele:
Also handled the two suppressed Copilot notes in f7cc9207d:

- **Empty-string default rendered blank** (taskEditor.ts) — an untouched empty field now always shows the typing hint, whether the run recorded no task (`null`) or an empty one (`""`). Added a test for `defaultTask === ""`.
- **`matchTraceId` used `in`** (statelogScan.ts) — switched to `Object.hasOwn(scan.eventsByTrace, requested)`, the repo convention for user-supplied keys.

PR #856, opened 2026-08-18
Run directory arc, phase 2: lib/runDirectory core (traces, annotations, snapshots, mutations)

egonSchiele:
Phase 2 of the run-directory arc (stacked on #855; spec and plan in `docs/superpowers/`). This adds the core library, `lib/runDirectory/`, with no user-facing change yet. Phase 3 (CLI primitives) and later phases build on it.

## What is in `lib/runDirectory/`

- **`traces.ts`** — `readTraces` groups a statelog into traces with a **digest over canonicalized events** (key order irrelevant); drops a torn tail and byte-identical duplicate lines. Built on a new `parseStatelogJsonlWithLines` in `lib/statelog/parse.ts` (the one line decoder; `parseStatelogJsonl` now delegates to it).
- **`annotations.ts`** — named row types (`note` / `checklist` / `score` / `run`), zod schemas, **deterministic ids** (`ann_` + sha256 of trace+annotator+payload+session; scores include `passId`), tolerant reader, and the pure fold: notes accumulate, checklist answers fold per question per annotator, **scores count only from complete passes** (`passSize` distinct rows + one `completesPass`), latest `run` row wins.
- **`runDir.ts`** — `runDirPaths` and `readRunDirectory`: one coherent snapshot without a reader lock (statelog → annotations → statelog, retry if the statelog changed).
- **`lock.ts`** — the label store's lock moved here as `acquireRunDirLock` (a shim keeps the old names until Phase 5).
- **`mergeStatelog.ts` / `attachCode.ts` / `attachWorkdir.ts`** — pure planners + private appliers: merge skips equal digests and **refuses** conflicting ids (all-or-nothing); code is stored by `code/<closureHash>/` and **refused** unless a trace recorded that hash; workdir snapshots get a dated sidecar, replacement goes through the new `safeDeleteDirectoryWithin(root, target)` in `lib/utils.ts` (strict descendant only, symlink-resolved).
- **`mutations.ts`** — the four public writes: `addToRunDirectory`, `recordCompletedRun`, `recordNote`, `recordGradingPass`. Each: lock → snapshot → preflight the whole request → repair torn tails → apply → fresh snapshot; lock released on any failure. Nothing else may import the lock or `apply*`.
- **`evalRecord.ts` / `list.ts`** — `evalRecordFor(trace)` (the EvalRecord as a view, never a file), `traceEnding`, `summarizeRuns`.

## Testing

57 tests in `lib/runDirectory/` covering: canonical digests, duplicate/torn lines, id determinism and per-pass ids, per-question and per-annotator folds, incomplete-pass exclusion, snapshot retry on concurrent write, lock behaviour (moved test), all-or-nothing merge, code match/mismatch/corrupt-tree, workdir add/refuse/replace, idempotent notes and imports, complete vs. crashed grading passes, torn-tail repair, and safe deletion (root, sibling, `..`, symlink escape all refused). 624 tests green across `lib/runDirectory lib/eval/label lib/logsViewer/parse lib/utils lib/statelog lib/cli/remote/logsBridge`; `tsc`, `lint:structure`, `fmt:ts` clean.

## Docs

New `docs/dev/run-directory.md` (+ CLAUDE.md pointer). No `docs/site` changes needed for this phase.


packages/agency-lang/lib/runDirectory/attachWorkdir.ts
```diff
@@ -0,0 +1,82 @@
...
+): WorkdirAttachmentPlan {
+  if (!snapshot.traces.some((trace) => trace.traceId === request.traceId)) {
+    throw new WorkdirAttachmentError(
+      `No trace ${request.traceId} in ${snapshot.dir}; add its statelog before its workdir.`,
+    );
+  }
+  if (!fs.existsSync(request.sourceDir) || !fs.statSync(request.sourceDir).isDirectory()) {
+    throw new WorkdirAttachmentError(`${request.sourceDir} is not a directory.`);
+  }
+  const target = path.join(paths.workdirDir, request.traceId);
```
egonSchiele, line 47: **Validate containment before constructing/applying this attachment.** Trace IDs come from statelog input and the parser accepts any truthy string, so an existing trace named `../../escaped` makes `target` and `sidecar` resolve outside `paths.workdirDir`. The add path then writes there directly; `safeDeleteDirectoryWithin` only protects replacements. This breaks the run-directory write boundary and can overwrite files outside `workdir/`. Please reject unsafe path components (or verify both resolved paths are contained) during planning, and cover a traversal trace ID.

packages/agency-lang/lib/runDirectory/mutations.ts
```diff
@@ -0,0 +1,363 @@
...
+  return withWriter(request.dir, options, (paths, snapshot, reportWarning) => {
+    const incoming = readTraces(request.stagedStatelogFile).traces;
+    const statelogPlan = planStatelogMerge(snapshot.traces, incoming);
+    if (statelogPlan.refused.length > 0) {
+      throw new Error(
+        `Refusing to record run: trace ${statelogPlan.refused[0].traceId} already exists in ` +
+          `${request.dir} with different content.`,
+      );
+    }
+    const merged: RunDirectorySnapshot = {
```
egonSchiele, line 151: **Preflight that `request.run.traceId` exists in this merged snapshot.** Unlike `recordNote` and `recordGradingPass`, this operation appends its annotation without checking the referenced trace. If the harness passes `t2` while the staged statelog contains `t1` (or the staged log is empty/torn), the call succeeds and leaves an orphan run row while `t1` loses its authoritative completion metadata. Please refuse the request before any writes and add a mismatched-ID test.

packages/agency-lang/lib/runDirectory/mutations.ts
```diff
@@ -0,0 +1,363 @@
...
+export type RecordCompletedRunResult = { annotation: Annotation; snapshot: RunDirectorySnapshot };
+
+/** The eval harness's one call per finished test: merge the staged statelog,
+ *  attach code and workdir, append the `run` row. */
+export function recordCompletedRun(
+  request: RecordCompletedRunRequest,
+  options: MutationOptions = {},
+): RecordCompletedRunResult {
+  return withWriter(request.dir, options, (paths, snapshot, reportWarning) => {
+    const incoming = readTraces(request.stagedStatelogFile).traces;
```
egonSchiele: **Do not silently persist a partially parsed staged trace.** `readTraces` returns `{ traces, errors }`, but this drops `errors`; the add path does the same at line 76. A valid `agentStart` followed by malformed or unsupported-version events is therefore stored as a truncated trace, and `recordCompletedRun` still marks it completed without warning. Since these are writer inputs and the operation promises complete preflight, please either refuse parse errors or explicitly report them through `reportWarning` before persisting; add a malformed-middle-line test.

egonSchiele (commented):
Found three correctness/safety issues in the mutation boundary.

packages/agency-lang/lib/runDirectory/annotations.ts
```diff
@@ -0,0 +1,273 @@
...
+  flags: Record<string, JsonValue>;
+};
+
+export type AnnotationPayload = NotePayload | ChecklistPayload | ScorePayload | RunPayload;
+
+/** What a writer supplies; `v`, `id` and `createdAt` are filled at append. */
+export type AnnotationDraft = {
+  traceId: string;
+  annotator: Annotator;
+  sessionId?: string;
```
egonSchiele, line 76: what's a session?

packages/agency-lang/lib/runDirectory/attachCode.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (!fs.existsSync(storedPath)) {
+      throw new CodeMismatchError(
+        `${target} is missing ${file.file}; the stored code tree is incomplete.`,
+      );
+    }
+    return { file: file.file, sha256: hashFile(storedPath) };
+  });
+  if (closureHashOf(stored) !== identity.closureHash) {
+    throw new CodeMismatchError(
+      `${target} does not hash to its own name; the stored code tree is corrupt.`,
```
egonSchiele: Is it possible to add what the expected name is?

packages/agency-lang/lib/runDirectory/lock.ts
```diff
@@ -0,0 +1,159 @@
+import * as fs from "fs";
```
egonSchiele, line 1: Does node come with a built-in lock mechanism that we could use instead of writing our own?

packages/agency-lang/lib/runDirectory/mergeStatelog.ts
```diff
@@ -0,0 +1,52 @@
...
+      plan.refused.push({ traceId: trace.traceId, reason: "conflicting-digest" });
+    }
+  }
+  return plan;
+}
+
+/** @internal Writes a plan that has no refusals. Caller holds the lock. */
+export function applyStatelogMerge(paths: RunDirectoryPaths, plan: StatelogMergePlan): void {
+  if (plan.refused.length > 0) {
+    throw new Error("applyStatelogMerge: refusing to apply a plan with conflicts");
```
egonSchiele, line 68: would be nice to print out exactly which trace IDs conflicted.

packages/agency-lang/lib/runDirectory/mergeStatelog.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+  if (plan.add.length === 0) return;
```
egonSchiele, line 70: would be nice to explicitly print out that nothing was added, and how many existing trace IDs were found which corresponded to the ones that the user was trying to add.

packages/agency-lang/lib/runDirectory/traces.ts
```diff
@@ -0,0 +1,89 @@
+import * as fs from "fs";
```
egonSchiele, line 1: I believe we do have a class that can be used to parse all the lines of a trace and get interesting metadata from it:
lib/eval/statelogParser.ts

Is there some overlap between this class and your code here?

packages/agency-lang/lib/utils.ts
```diff
@@ -88,6 +88,53 @@ export function safeDeleteDirectory(targetPath: string, dryRun: boolean = true):
...
+      success: false,
+      message: `Refusing to delete '${target}': ${err instanceof Error ? err.message : String(err)}`,
+    };
+  }
+  if (!fs.statSync(resolvedTarget).isDirectory()) {
+    return { success: false, message: `Not a directory: '${resolvedTarget}'.` };
+  }
+  const relative = path.relative(resolvedRoot, resolvedTarget);
+  const strictlyInside =
+    relative !== "" && !relative.startsWith("..") && !path.isAbsolute(relative);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 93b7543: containment is now `isStrictDescendant`, which checks the first path segment is exactly `..` rather than `startsWith("..")`. Test: a child named `..cache` is deleted.

packages/agency-lang/lib/runDirectory/lock.ts
```diff
@@ -0,0 +1,159 @@
...
+  reportWarning: (message: string) => void,
+): void {
+  if (path.basename(lockFile) !== LOCK_BASENAME) {
+    reportWarning(`Refusing to remove ${lockFile}: not a lock file`);
+    return;
+  }
+  const existing = readHolder(lockFile);
+  if (existing === undefined) {
+    return;
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 93b7543: a lock file that is missing is still silent (already released), but one that is present and unreadable now warns and says what to do. Test added.

packages/agency-lang/lib/runDirectory/mutations.ts
```diff
@@ -0,0 +1,363 @@
...
+
+/** One grading pass: every score draft gets the same fresh `passId`, the
+ *  pass size, and `completesPass` on the final row only. A crash before that
+ *  row leaves an incomplete pass the fold ignores. */
+export function recordGradingPass(
+  request: RecordGradingPassRequest,
+  options: MutationOptions = {},
+): RecordGradingPassResult {
+  const passId = newPassId();
+  const result = withWriter(request.dir, options, (paths, snapshot, reportWarning) => {
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 93b7543: `recordGradingPass` throws on an empty `scores` before minting a pass id. Test added.

packages/agency-lang/lib/utils.ts
```diff
@@ -88,6 +88,53 @@ export function safeDeleteDirectory(targetPath: string, dryRun: boolean = true):
...
+    resolvedTarget = fs.realpathSync(path.resolve(target));
+  } catch (err) {
+    return {
+      success: false,
+      message: `Refusing to delete '${target}': ${err instanceof Error ? err.message : String(err)}`,
+    };
+  }
+  if (!fs.statSync(resolvedTarget).isDirectory()) {
+    return { success: false, message: `Not a directory: '${resolvedTarget}'.` };
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 93b7543: `statSync` is inside the try and a failure comes back as `{ success: false }`.

packages/agency-lang/lib/runDirectory/attachWorkdir.ts
egonSchiele: Fixed in 93b7543. `planWorkdirAttachment` resolves both the target and the sidecar and refuses unless each is a direct child of `workdir/` (`isStrictDescendant` + same dirname), before anything is written. Test: a trace id of `../escaped` is refused and nothing lands outside.

packages/agency-lang/lib/runDirectory/mutations.ts
egonSchiele: Done, with one refinement (93b7543 + 33132bf). Phase 4 deliberately records a `run` row for a run that died before writing any event, since that row is the only record the run happened (see the runSuite test 'a failed test still gets a run row'). So the rule is: if the staged statelog holds any traces, the run's trace id must be one of them or already in the directory; otherwise refuse and name what the staged log holds instead. A staged log with no traces still records the row. Tests for both the mismatch and the traceless case.

packages/agency-lang/lib/runDirectory/mutations.ts
egonSchiele: Fixed in 93b7543: writers go through `readTracesOrThrow`, which refuses the whole file if any line fails to parse (naming the first few line numbers and kinds); readers stay tolerant. Both `addToRunDirectory` and `recordCompletedRun` use it. Test: a malformed middle line is refused and nothing is written.

packages/agency-lang/lib/runDirectory/annotations.ts
egonSchiele: Added a doc comment on the field in 93b7543. A session is one sitting of `agency label`: one person, one checklist, one ordered list of traces. It comes into the row's id hash so two sittings that gave identical answers are still two rows.

packages/agency-lang/lib/runDirectory/attachCode.ts
egonSchiele: Done in 93b7543: the message now says which closure hash the tree is named for and what its files actually hash to.

packages/agency-lang/lib/runDirectory/lock.ts
egonSchiele: No: Node's `fs` has no file-lock call (no `flock`), so the standard idiom is what this does: create the lock file with the `wx` flag (fails if it exists) and record the holder. The usual package (`proper-lockfile`) does the same thing plus mtime-based staleness and automatic takeover, and takeover is exactly what we do not want here. I added a short paragraph to docs/dev/run-directory.md saying so.

packages/agency-lang/lib/runDirectory/mergeStatelog.ts
egonSchiele: Done in 93b7543: `describeStatelogMerge(plan, dir)` names every refused id, and both refusal errors use it (previously `recordCompletedRun` named only the first).

packages/agency-lang/lib/runDirectory/mergeStatelog.ts
egonSchiele: Done in 93b7543 + 8ec3d5b (phase 3). `describeStatelogMerge` says 'Nothing was added to <dir>: all N trace(s) were already present (t1, t2).' when that is the case, and `addToRunDirectory` returns it as `statelogSummary` so `agency runs add` prints that sentence instead of '0 added, 2 already present'. Test in commands.test.ts.

packages/agency-lang/lib/runDirectory/traces.ts
egonSchiele: There is overlap at the parsing layer only, and it is shared rather than duplicated: both go through `parseStatelogJsonlWithLines` (lib/statelog/parse.ts), and a trace's eval record is still `extractEvalRecord` via `evalRecord.ts`. `StatelogParser` is a one-trace, one-file view (it asserts a single trace) built for graders. `traces.ts` is the many-trace reader a run directory needs: it groups by trace id, keeps the raw lines so a merge can append them byte-for-byte, and computes the digest. I added a comment on `readTraces` saying exactly this (93b7543).

packages/agency-lang/lib/runDirectory/mergeStatelog.ts
egonSchiele: Correction: the phase 3 commit is cf42eda, not 8ec3d5b.

PR #857, opened 2026-08-18
Run directory arc, phase 3: logs extract, runs add/list, note; viewer x replaces l

egonSchiele:
Phase 3 of the run-directory arc (stacked on #856). The first user-facing pieces: four small commands that compose over `lib/runDirectory/`, and the viewer's `l` key replaced by `x`.

## New commands (`lib/cli/runDirectory/`, one file each)

- **`agency logs extract <log> [--trace <id>] [-o <file>]`** — copy one trace out of a statelog, verbatim, to a file or stdout. A single-trace log needs no `--trace`; several list themselves. A trace slice is itself a valid statelog, so this is the primitive that turns "a run I noticed in a big log" into "a run I can attach things to".
- **`agency runs add <dir> [--statelog f]… [--code entry]… [--workdir p [--trace id]] [--annotations f]… [--replace]`** — translates the flags into one `addToRunDirectory` request and renders the counts plus the listing. Refuses conflicting traces and mismatched code with the recorded hash in the message; nothing is written on refusal.
- **`agency runs list <dir>`** — one line per trace: started, ended, time, cost, LLM/tool calls, score, notes, labeled, input.
- **`agency note <dir> <text> [--trace id] [--annotator who]`** — appends a `note` annotation (`--trace` optional when there is one trace).

None of these import the lock, merge, attachment, deletion or annotation-log modules.

## Viewer

Tree view: **`x` extracts the focused trace** to a file (prompts for a path, default `<traceId>.jsonl`), re-reading it from the local source so the copy is verbatim; follow is paused during the prompt. **`l` (label into a dataset) is removed**, along with `labelTrace.ts`, `taskEditor.ts`, and the `agency logs --dataset/--checklist` flags. The runs explorer's drill-in enables `x` too. Per the plan, `labelingHost` and `datasetWriter` stay for now — `agency label` / `label ingest` still use them and keep working until Phase 5.

## Testing

`lib/cli/runDirectory/commands.test.ts` covers extract (single/multi/prefix/ambiguous, `-o` creates parents), `runs add` (idempotent, code mismatch names the hash, workdir with/without `--trace`), `note` (only trace, several traces, empty text), `runs list` (counts, empty dir). Viewer test: `x` writes the exact bytes and reports; without a local source it does nothing. 2269 tests green across `lib/logsViewer lib/cli lib/runsExplorer lib/runDirectory lib/eval/label scripts`; `tsc`, `lint:structure`, `fmt:ts` clean.

## Docs

`docs/dev/run-directory.md` gains a Commands section; `docs/dev/statelog-to-dataset.md` deleted (CLAUDE.md pointer removed); `docs/dev/eval-labeling.md` notes the change. Owner: `docs/site/cli/logs.md` "Labeling a trace from the viewer" and `docs/site/cli/eval.md` "Labeling a statelog trace" need rewriting, and the new commands need pages.


packages/agency-lang/lib/runDirectory/extractTrace.ts
```diff
@@ -0,0 +1,16 @@
...
+/** Find one trace in a statelog by full id or unique prefix. */
+export function findTrace(statelogPath: string, idOrPrefix: string): TraceMatch {
+  return matchTrace(readTraces(statelogPath).traces, idOrPrefix);
+}
+
+/** Write one trace's lines, verbatim, as a statelog of its own. A trace slice
+ *  is a valid statelog, so everything downstream reads it unchanged. */
+export function writeTraceFile(trace: Trace, outPath: string): void {
+  fs.mkdirSync(path.dirname(path.resolve(outPath)), { recursive: true });
+  fs.writeFileSync(outPath, trace.lines.join("\n") + "\n");
```
egonSchiele: **Refuse an existing output path unless overwrite is explicitly requested.** This truncates any destination silently, including the source statelog itself: `agency logs extract combined.jsonl --trace abc -o combined.jsonl` replaces the full log with one trace. The viewer has the same risk when its default `<traceId>.jsonl` already exists (or is the file currently open). Please use exclusive creation by default, reject source/destination identity, and add collision coverage; an explicit overwrite option can opt into replacement if needed.

egonSchiele (commented):
Found one additional data-loss issue beyond the existing review comments.

packages/agency-lang/lib/cli/runDirectory/table.ts
```diff
@@ -0,0 +1,21 @@
+/** A plain left-aligned text table: header row, then rows, columns padded to
+ *  the widest cell. For CLI listings, not the interactive explorer. */
+export function formatTextTable(header: string[], rows: string[][]): string {
+  const all = [header, ...rows];
```
egonSchiele, line 12: Let's color the headers for readability using lib/utils/termcolors.ts

packages/agency-lang/lib/runDirectory/extractTrace.ts
```diff
@@ -0,0 +1,16 @@
...
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 763b91b: the doc comment now says lines are copied as they appear, minus the two things readTraces drops (a torn final line and a byte-identical repeat within the trace).

packages/agency-lang/lib/cli/runDirectory/extract.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const defaultDependencies: LogsExtractDependencies = {
+  writeStdout: (text) => process.stdout.write(text),
+};
+
+/**
+ * Copy one trace's lines out of a statelog, verbatim, to a file or stdout. A
+ * trace slice is itself a valid statelog, so this is the primitive that turns
+ * "a run I noticed in a big log" into "a run I can attach things to".
+ */
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 763b91b: help text and doc comment describe the drop of torn-tail and byte-identical repeated lines instead of saying verbatim.

packages/agency-lang/docs/dev/run-directory.md
```diff
@@ -111,6 +111,18 @@ ended from its own events (`ok` / `error` / `unknown`); the harness's `run` row,
 when present, knows more. `summarizeRuns(snapshot)` (`list.ts`) is one
 `RunSummary` per trace for listings.
 
+## Commands (`lib/cli/runDirectory/`)
+
+One file per command; none imports the lock or the append helpers.
+
+- `agency logs extract <log> [--trace <id>] [-o <file>]` — copy one trace out
+  of a statelog, verbatim (a single-trace log needs no `--trace`). The viewer's
+  `x` key does the same for the focused trace, prompting for a path.
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 763b91b: the dev doc spells out what is and is not copied.

packages/agency-lang/lib/runDirectory/extractTrace.ts
egonSchiele: Fixed in 763b91b. `writeTraceFile` now creates the output with the exclusive `wx` flag; an existing file is refused with a message pointing at `--overwrite`, and the source log is refused as an output even with `--overwrite`. The viewer's `x` key uses the same function (no overwrite option there, so an existing file is always refused). Tests: CLI collision, overwrite, and source identity; viewer collision leaves the file untouched.

packages/agency-lang/lib/cli/runDirectory/table.ts
egonSchiele: Done in 763b91b: headers are `ttyColor.cyan.bold`, applied after padding so alignment is untouched, and plain when piped. Test with an injected paint function.

PR #859, opened 2026-08-18
agency label reads a run directory; ingest and the label store are gone (Phase 5)

egonSchiele:
Phase 5 of the run-directory arc (spec: `docs/superpowers/specs/2026-08-18-run-directory-and-annotations-design.md`, plan: `docs/superpowers/plans/2026-08-18-run-directory-and-annotations.md`). Stacked on #858 (Phase 4).

## What changes

`agency label <dir> --checklist <file> [--annotator <id>]` labels the traces of a **run directory**. There is no separate label store any more:

- The list of things to judge is `statelog.jsonl` (one item per trace, shown as `input` + `output`; when a trace recorded no output, the last assistant message is shown under `last_message` with a clear marker).
- Each sign-off appends one `checklist` row to `annotations.jsonl` (`{ kind: "checklist", checklist, version, hash, answers, note, activeMs }`, deterministic id, `sessionId` recorded).
- Checklist revisions stay under `checklists/<id>/`; the session draft moves to `checklists/<id>/drafts/<session>.json`.
- The session holds the run directory's writer lock for its whole life (two sessions on one directory would race for a revision number and share a draft).

Removed: `agency label ingest` and every loader, `--dataset` / `eval.dataset`, the label store (`dataset.ts`, `corpus.ts`, `occurrences.ts`, `datasetWriter.ts`, `labelingHost.ts`), content-derived record ids, the old `AnnotationRow`, `lib/eval/readRun.ts`. No migration, by decision.

## Where the code went

- `lib/runDirectory/labelStore.ts` — the facade the controller works through: items, this annotator's folded answers (from `readRunDirectory`'s fold), draft load/save, checklist prepare/publish, checklist-row append. Validation at open is strict (broken lineage, row naming a missing revision or mismatched hash → refuse).
- `lib/runDirectory/traceText.ts` — one rule for a trace's input/output as text; `runs list` and the pick-a-trace error use it too.
- `lib/runDirectory/durableWrite.ts` — `appendDurably` / `atomicWriteValidated`, moved from the label store's `jsonl.ts` since the run directory owns durable writes now.
- `lib/eval/label/judgement.ts` — status + score math over folded answers (the fold itself is `foldAnnotations`).
- Controller keeps the same sign-off order and every recovery boundary; `controller.test.ts` still crashes at each `ControllerFaultPoint` and reopens on a real run directory.

## Deviation from the plan (Task 5.1)

[...]

PR #860, opened 2026-08-18
Optimize reads annotations, logs opens run directories, run --capture-workdir (Phase 6)

egonSchiele:
Phase 6, the last phase of the run-directory arc (spec: `docs/superpowers/specs/2026-08-18-run-directory-and-annotations-design.md`, plan: `docs/superpowers/plans/2026-08-18-run-directory-and-annotations.md`). Stacked on #859 (Phase 5).

## 6.1 The optimizer reads notes and checklist misses as feedback

`humanFeedbackFor(snapshot, traceId)` (`lib/runDirectory/humanFeedback.ts`) collects every note on a trace and the text of every checklist question a reviewer answered no. `gradeRun` attaches it to `InputGrades.humanFeedback`; `renderInputFeedback` renders both under their own headings ("Notes from people who reviewed this run:", "Checklist questions reviewers answered NO:"). GEPA and the greedy optimizer get it for free because they grade run directories. The reflection prompt already said `Input:` after Phase 4.

## 6.2 `agency logs <dir>` opens a run directory; the explorer reads run directories

- A **sole** run directory opens the viewer on its `statelog.jsonl` with each trace's annotations summarised on the trace row (`2 notes · score 0.70 · labeled`, from `annotationSummaries` in `lib/runDirectory/list.ts`). The tree view also takes a `focusTraceId`.
- The **runs explorer** (several paths / a directory of runs) now classifies a run directory as "has `statelog.jsonl`" and loads it through one `readRunDirectory` (`buildRunRowFromDirectory` over `summarizeRuns`). `summary.json`/`config.json` reading, per-input backfill, `readRunSummary.ts` and the legacy `EvalRunResult` shapes are gone. Drilling from a test hands the viewer the run's shared statelog focused on that test's trace.
- The harness `run` row now records the agent label under `flags.agent`, so listings can name the agent. `RunSummary` gained `testId`, `agentName`, `startedAtMs`, `models`, `gatesPassed`.

## 6.3 `agency run --capture-workdir <dir>`

Mints a trace id (`AGENCY_TRACE_ID`), routes the child's statelog to a private staging file, and after exit makes one `addToRunDirectory` call: that statelog, the entry file's code closure, and the working directory as the trace's workdir snapshot. Covered end to end in `tests/integration/eval-run/test.mjs` (scenario E). I did not add the plan's `run.captureWorkdir` config field: the flag has a user, the config key does not yet.

## Verification

- `tsc`, `lint:structure`, `fmt:ts` clean.
- Scoped vitest green: `lib/runDirectory`, `lib/runsExplorer`, `lib/logsViewer/{run,treeRows}.test.ts`, `lib/cli/logsView.test.ts`, `lib/optimize/reflectionFeedback.test.ts`, `lib/eval/grading/gradeRun.test.ts`, `lib/eval/run/runSuite.test.ts`, `lib/sourceIsText.test.ts`.
- `node tests/integration/eval-run/test.mjs` passes all five scenarios against a fresh `dist/`.

## Docs

- `docs/dev/writing-optimizers.md` (feedback sources), `docs/dev/runs-explorer.md` (rewritten loading section, viewer hand-off), `docs/dev/logs-viewer.md` (run directories), `docs/dev/run-directory.md` (the two new commands).
- **Owner:** `docs/site/cli/run.md` (`--capture-workdir`), `docs/site/cli/logs.md` (a run directory opens the viewer with annotations; the explorer for several), `docs/site/cli/optimize.md` (notes/checklists as feedback).


packages/agency-lang/lib/cli/commands.ts
```diff
@@ -328,8 +342,47 @@ export function run(
...
+  if (!fs.existsSync(capture.statelogFile)) {
+    throw new Error(`the run wrote no statelog at ${capture.statelogFile}`);
+  }
+  addToRunDirectory(
+    {
+      dir: capture.runDir,
+      statelogFiles: [capture.statelogFile],
+      codeEntries: [inputFile],
+      annotationFiles: [],
+      workdir: { traceId: capture.traceId, sourceDir: process.cwd() },
```
egonSchiele: **Handle capture destinations inside the working directory.** A natural invocation such as `agency run --capture-workdir ./runs/example agent.agency` asks `applyWorkdirAttachment` to copy `process.cwd()` into its own descendant (`./runs/example/workdir/<traceId>`), which Node rejects as a recursive self-copy. By then the statelog and code have already been merged, leaving a partial run directory. Please either stage the capture outside the source tree before folding it, exclude the destination subtree, or reject an in-tree destination before any writes; add an integration test with a relative destination under cwd.

egonSchiele (commented):
Found one additional capture failure beyond the existing review comments.

packages/agency-lang/lib/cli/commands.ts
```diff
@@ -325,8 +338,47 @@ export function run(
...
+    },
+    { reportWarning: (message) => console.warn(message) },
+  );
+  fs.rmSync(capture.stagingDir, { recursive: true, force: true });
```
egonSchiele: uhh... this seems bad. What is stagingDir? Would this ever delete important data?

packages/agency-lang/lib/logsViewer/views/treeView.ts
```diff
@@ -20,22 +20,37 @@ import type { ViewerThresholds } from "../thresholds.js";
...
   ) {
+    this.extractEnabled = options.extractEnabled ?? false;
+    this.traceAnnotations = options.traceAnnotations ?? {};
+    const focused = roots.find((root) => root.traceId === options.focusTraceId);
+    const start = focused ?? roots[0];
     this.state = {
       roots,
-      expanded: new Set(roots.length === 1 ? [roots[0].id] : []),
-      cursorId: roots[0]?.id ?? "",
+      expanded: new Set(roots.length === 1 || focused !== undefined ? [start.id] : []),
```
egonSchiele: this is hard to read

packages/agency-lang/lib/runDirectory/humanFeedback.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * What people said about a trace, in words an optimizer can feed back to the
+ * model: every note, and the text of every checklist question someone answered
+ * "no". Grader feedback is separate (it rides on the score rows); this is the
+ * human side.
+ */
+export type HumanFeedback = {
+  notes: string[];
+  /** Question texts answered `false` by any annotator on any checklist. */
+  unchecked: string[];
```
egonSchiele, line 19: This assumes that checklist questions will only be written a certain way. What if it's useful to get the questions that were answered "true" by an annotator? Would that ever happen?

packages/agency-lang/lib/cli/commands.ts
```diff
@@ -284,15 +291,22 @@ export function run(
   console.log("---");
 
   const env: NodeJS.ProcessEnv = { ...process.env };
+  const captured = capture === undefined ? undefined : prepareCapture(capture.runDir);
   // Which code this run is, recorded on the trace's agentStart. Layered under
   // any inherited overrides (an eval harness hands this process its statelog
   // path the same way), so nothing a parent set is lost.
   env[CONFIG_OVERRIDES_ENV] = serializeConfigOverrides(
     mergeConfigOverrides(
```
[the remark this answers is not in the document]
egonSchiele: This was already reordered when the earlier code-identity fix merged forward, and 9a6221d makes it explicit: `runChildOverrides` merges inherited overrides first, then the capture statelog (`log.logFile` + `observability`), then `log.code`, so the last two always win. Test with an inherited `logFile`, `observability: false` and a stale `log.code`.

packages/agency-lang/lib/cli/commands.ts
```diff
@@ -328,8 +342,47 @@ export function run(
...
+      codeEntries: [inputFile],
+      annotationFiles: [],
+      workdir: { traceId: capture.traceId, sourceDir: process.cwd() },
+    },
+    { reportWarning: (message) => console.warn(message) },
+  );
+  fs.rmSync(capture.stagingDir, { recursive: true, force: true });
+  console.log(`---
+Captured trace ${capture.traceId} into ${capture.runDir}`);
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 9a6221d: removal is in a `finally`, so the temp dir goes whether or not the fold succeeded (a missing statelog included).

packages/agency-lang/lib/runDirectory/list.ts
```diff
@@ -35,25 +42,69 @@ function summarizeTrace(
...
+/** False when an effective must-pass binary score failed; null with no
+ *  scores. A scalar gate's threshold lives on the grader, not the row, so a
+ *  scalar must-pass score cannot be judged here and counts as passed. */
+function gatesPassed(effective: EffectiveTraceAnnotations | undefined): boolean | null {
+  const rows = Object.values(effective?.scores ?? {});
+  if (rows.length === 0) return null;
+  return rows.every(
+    (row) => row.kind !== "score" || !row.mustPass || row.score.kind !== "binary" || row.score.pass,
+  );
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 9a6221d: any must-pass scalar score makes `gatesPassed` null (unknown) rather than true. Test: the same trace flips from true to null when the scalar score becomes a gate.

packages/agency-lang/lib/cli/commands.ts
egonSchiele: Fixed in 9a6221d. `applyWorkdirAttachment` copies with a walk that leaves the run directory subtree out (`copyTreeExcluding`; `cpSync` refuses a self-containing tree outright, so its filter option cannot do it), and nothing is written before the plan. Unit test with the run dir under the source, and integration scenario F runs `agency run --capture-workdir ./runs/example` from the captured cwd and checks the workdir holds the tree but not `runs/example`.

packages/agency-lang/lib/cli/commands.ts
egonSchiele: It is only ever the fresh, empty directory this process just made with `mkdtempSync` under the OS temp dir (`/tmp/agency-capture-XXXXXX`), holding just the staged statelog for this one run. Nothing else is ever passed to it. In 9a6221d I renamed it `captureTempDir`, documented that on the type, and made the removal refuse anything that is not directly under `os.tmpdir()` with the `agency-capture-` prefix, so even a bug elsewhere cannot point it at real data.

packages/agency-lang/lib/logsViewer/views/treeView.ts
egonSchiele: Rewritten in 9a6221d with a comment and a named `startExpanded`: start on the trace the caller asked for, else the first; it opens expanded when asked for by id or when it is the only trace; a multi-trace log with no focus opens collapsed.

packages/agency-lang/lib/runDirectory/humanFeedback.ts
egonSchiele: Good point. In 9a6221d `humanFeedbackFor` returns both: `checked` (answered yes) and `unchecked` (answered no), each as question texts. Reflection still feeds the model only `unchecked`, since that is what reviewers found wrong, but anything that wants to say what a run did right, or compare two annotators, has the yes side without re-reading the checklist files. Doc updated.

PR #864, opened 2026-08-18
Reshape eval run/grade: goal moves to grade, positional agent, optional input

egonSchiele:
## What changed

The `eval run` and `eval grade` commands now split cleanly into "run the agent" and "judge the run".

### `eval run`
- The agent is a positional argument: `agency eval run my-agent.agency` (or `--agent-cmd '...'` for a command agent).
- `--goal` is removed. A one-off input is `--input <text>`, and it is optional: `agency eval run my-agent.agency` alone runs one input-less test, for agents whose entry node takes no argument. Suites may omit `input` too.
- Dropped `--continue-on-error` / `--no-continue-on-error`, `--strict`, `--max-tool-call-rounds`, `--max-tool-result-chars`. Every test always runs; an errored test grades as zero.
- Before anything runs, `assertTargetMatchesInputs` checks that the agent's shape agrees with the inputs: a file agent needs one entry-node parameter when inputs are present and zero when they are not; a command agent needs `{task}` iff inputs are present; a suite must be all-with-input or all-without.

### `eval grade`
- New `--goal <text>`: the default goal for every test that recorded none. Exclusive with `--graders` (a grading module carries its own criteria).
- Refuses a directory with no `statelog.jsonl` and says how to build a run directory (`agency runs add` or `agency run --capture-workdir`). Previously this quietly printed `objective 0.000`.
- Preflight errors go through `failProjectCommand` instead of a stack trace.

No backwards-compatibility shims, as agreed.

## Verified
- `pnpm run typecheck` (all three tsconfigs), `lint:structure`, prettier
- Scoped vitest: `lib/agentTarget`, `lib/eval/loadInputs`, `lib/cli/eval`, `lib/eval/run`, `lib/eval/grading`, `lib/optimize`, `lib/runsExplorer/suite`
- After `pnpm run build` + `npm pack`: `tests/integration/eval-run/test.mjs`, `tests/integration/cli/test.mjs`
- End to end: `agency runs add ... --statelog ...` then `agency eval grade <dir> --goal "..."` scores 1.000; a no-arg agent runs under `eval run` with no `--input`

## Left for the owner
`docs/site/cli/eval.md` still documents `--agent`, `eval run --goal`, and the removed flags.

### DIFF
--- a/packages/agency-lang/lib/cli/eval/grade.ts
+++ b/packages/agency-lang/lib/cli/eval/grade.ts
@@ -7,13 +7,15 @@ import type { EvalRunGrading } from "@/eval/runTypes.js";
 
 import { runDirPaths } from "@/runDirectory/runDir.js";
 
-import { resolveGraders } from "./graders.js";
+import { goalJudgeGraders, resolveGraders } from "./graders.js";
 
 export type EvalGradeOptions = {
   /** Path to a TypeScript grading module. Defaults to `eval.graders`, then the goal judge. */
   graders?: string;
-  /** The goal for the default judge, for every trace whose test recorded none.
-   *  Not combined with `graders`: a module brings its own criteria. */
+  /** Judge every trace against this goal with the bundled goal judge (a test's
+   *  own recorded goal still wins). Not combined with `graders`, and it sets
+   *  aside a configured `eval.graders` module too: a module brings its own
+   *  criteria, and `--goal` names the criterion. */
   goal?: string;
   /** Also write the grading summary here, as JSON. */
   out?: string;
@@ -40,6 +42,12 @@ export function validateGradeTarget(runDir: string, opts: EvalGradeOptions): voi
   }
 }
 
+/** The grader set `eval grade` runs with; see `resolveGraders` for the precedence. */
+export async function gradersFor(opts: EvalGradeOptions, config: AgencyConfig) {
+  if (opts.goal !== undefined) return goalJudgeGraders();
+  return resolveGraders(opts.graders, undefined, config);
+}
+
 /**
  * Score a run directory. Never re-executes the agent. Every pass appends
  * `score` annotations to the directory — a re-grade sits beside the earlier
@@ -53,7 +61,8 @@ export async function evalGrade(runDir: string, opts: EvalGradeOptions): Promise
   // grading switched off, so the same resolver's default path applies. An
   // explicit --graders overrides every test's recorded graders; otherwise
   // per-test graders apply, with the config module / goal judge as fallback.
-  const graders = await resolveGraders(opts.graders, undefined, config);
+  // --goal promises the goal judge, so it never reaches the config module.
+  const graders = await gradersFor(opts, config);
   // resolveGraders only returns undefined for --no-grade, which this command never
   // passes, and otherwise falls back to the goal judge — so the reachable case is a
   // grading module that default-exports an empty array.
--- a/packages/agency-lang/lib/cli/eval/grade.ts
+++ b/packages/agency-lang/lib/cli/eval/grade.ts
@@ -5,15 +5,40 @@ import type { AgencyConfig } from "@/config.js";
 import { gradeSuite } from "@/eval/grading/gradeSuite.js";
 import type { EvalRunGrading } from "@/eval/runTypes.js";
 
+import { runDirPaths } from "@/runDirectory/runDir.js";
+
 import { resolveGraders } from "./graders.js";
 
 export type EvalGradeOptions = {
   /** Path to a TypeScript grading module. Defaults to `eval.graders`, then the goal judge. */
   graders?: string;
+  /** The goal for the default judge, for every trace whose test recorded none.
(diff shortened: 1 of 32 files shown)


packages/agency-lang/scripts/agency.ts
```diff
@@ -882,47 +850,26 @@ export function createProgram(deps: CliDependencies = {}): Command {
         }
         console.log(result.runDir);
         console.log(`grade it with: agency eval grade ${result.runDir}`);
-        if (result.errorCount > 0 && opts.continueOnError === false) {
-          process.exit(2);
-        }
       },
     );
 
-  evalCmd
```
egonSchiele, line 891: This removes the existing `agency eval logs` command even though the PR is only reshaping run/grade. The command is still part of the documented observability workflow (including `--input` and `-f`), and the current PR build confirms that `agency eval --help` no longer lists it, so existing invocations now fail as an unknown command. Please restore this registration unless removing the logs workflow is an intentional, separately documented change.

packages/agency-lang/lib/cli/eval/grade.ts
```diff
@@ -35,7 +61,9 @@ export async function evalGrade(runDir: string, opts: EvalGradeOptions): Promise
     throw new Error(`The grading module at ${opts.graders} exported no graders.`);
   }
 
-  const { grading } = await gradeSuite(resolvedRunDir, graders, config);
+  const { grading } = await gradeSuite(resolvedRunDir, graders, config, {
+    defaultGoal: opts.goal,
```
egonSchiele, line 74: `--goal` does not necessarily use the built-in goal judge promised by the CLI. `resolveGraders()` runs first and prefers `config.eval.graders`, so a project with configured graders executes that custom module and merely injects the goal into `test.goal`; a grader that ignores that field will score against unrelated criteria. Since a configured module carries criteria just as an explicit `--graders` module does, `--goal` should either override the configured fallback with `LlmJudge` or fail as a conflict. The new test currently codifies the surprising custom-grader behavior rather than the CLI contract.

packages/agency-lang/lib/eval/grading/gradeSuite.ts
```diff
@@ -38,6 +39,7 @@ export async function gradeSuite(
     suiteGraders,
     runAgency: new AgencyRunner(config),
     config,
+    defaultGoal: options.defaultGoal,
```
egonSchiele, line 42: The grading-time goal is only added to this in-memory copy of the test and is never recorded in the score pass. If the same run is graded with `--goal A` and later `--goal B`, `annotations.jsonl` contains two passes with the same `goal-judge@<prompt hash>` annotator but no record of which criterion produced either score. Before this change the goal lived in the persisted run row; moving it to grading time needs equivalent durable provenance (the effective goal or a criterion hash/config attached to the pass) so scores remain auditable and reproducible.

egonSchiele (commented):
I found three blocking correctness/provenance issues in the CLI reshaping.

packages/agency-lang/scripts/agency.ts
egonSchiele: Restored in 899942f5f, unchanged. The removal was a mistake.

packages/agency-lang/scripts/agency.ts
```diff
@@ -882,47 +850,26 @@ export function createProgram(deps: CliDependencies = {}): Command {
         }
         console.log(result.runDir);
         console.log(`grade it with: agency eval grade ${result.runDir}`);
-        if (result.errorCount > 0 && opts.continueOnError === false) {
-          process.exit(2);
-        }
       },
     );
```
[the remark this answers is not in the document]
egonSchiele: Restored in 899942f5f, unchanged.

packages/agency-lang/lib/cli/eval/grade.ts
egonSchiele: Fixed in 899942f5f: `gradersFor` returns the bundled goal judge whenever `--goal` is given, so a configured `eval.graders` module is set aside (per-test graders still apply, as fallback mode always allows). The test now checks that `--goal` picks the goal judge over a configured module.

packages/agency-lang/lib/eval/grading/gradeSuite.ts
egonSchiele: Fixed in 899942f5f: every judge score row now records the goal it scored against (`ScorePayload.goal`, set in `scoreDrafts` for annotators of kind `judge`), so two passes with `--goal A` and `--goal B` are tellable apart in annotations.jsonl. Covered by the new `gradeSuite.test.ts`.

egonSchiele (commented):
Re-reviewed 899942f5f. All three prior findings are addressed: `agency eval logs` is restored unchanged, `--goal` now selects the bundled goal judge instead of `config.eval.graders`, and the effective goal is durably recorded on judge score annotations. I traced the default-goal path through `Scorecard` and the annotation writer, ran the 42 targeted grading/run-directory tests, and checked the updated CI run; everything is green. No new actionable findings.

PR #866, opened 2026-08-19
Atomic run directories, chunk 1: writers produce one run per directory

egonSchiele:
Prototype, chunk 1 of 4 of the "one run = one directory" change (spec and plan are local, gitignored). Minimal tests, no docs/site, no polish: try it and say whether the shape is right.

## What changes

A run directory holds exactly one run; `code/` and `workdir/` are flat (`workdir.json` sidecar beside them). A **group** is any directory of run directories. No back-compat for directories written before this.

- `eval run --out <dir>` writes `<dir>/<testId>/` per test (`input-1` for `--input`). An existing child is an error for that test only; nothing is overwritten; the others run. Children are assembled in `<dir>/.staging/` and renamed into place.
- `agency run --capture-workdir <dir>` writes `<dir>/<traceId>/` (the group is left out of its own snapshot).
- `agency runs add <group> --statelog f [--trace id] [--code e] [--workdir p] [--annotations f]` wraps each trace as `<group>/<traceId>/`; existing children are skipped.
- `eval grade <path>` takes a run directory or a group: one grading pass per run, mean printed over the group. `findRunDirectories` is the one walk rule (shared with `agency logs`).
- Optimizer: grades the one child; its own flags untouched.

Not done (later chunks): `runs list` / `note` / `label` over groups, `notes.md`, the reader refusing multi-trace statelogs (kept tolerant so the labeling/explorer tests still pass).

## Try it

```
pnpm run build
pnpm run agency eval run test.agency --input "hi" --out runs/t1     # runs/t1/input-1/{statelog.jsonl,annotations.jsonl,code/,workdir/,workdir.json}
pnpm run agency eval run test.agency --input "hi" --out runs/t1     # "run directory already exists" for that input; 0/1 ok; nothing overwritten
pnpm run agency eval run test.agency --suite <suite> --out runs/t2   # runs/t2/<testId>/ each
pnpm run agency run --capture-workdir runs/scratch test.agency       # runs/scratch/<traceId>/
pnpm run agency runs add runs/wrapped --statelog log.jsonl           # one dir per trace; run twice → skipped
cp -r runs/t2/<testId> /tmp/x && pnpm run agency logs /tmp/x         # a run travels
pnpm run agency eval grade runs/t2 --goal "..."                      # per-run objective + mean line
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #867, opened 2026-08-19
Atomic run directories, PR 2: runs list and eval grade walk groups

egonSchiele:
PR 2 of the "one run = one directory" arc (PR 1 was #866). The two readers that still took a single run directory now walk groups.

## User-facing

- `agency runs list <path…>`: takes run directories and/or directories of run directories, any number, one table. Two new columns: `TEST` (from the harness `run` row) and `AGENT` (the trace's own `agentName` if it emitted one, else the harness label as recorded). Footer line: `N runs · mean 0.720 over G graded · S runs wrote no trace` (clauses only when they apply). A folder with no run directories is an error instead of "no traces".
- `agency eval grade <path…>`: takes several paths. A run named twice (through its group and directly, or through a symlink) is graded once and counted once.

## Design notes (also in `docs/dev/run-directory.md`)

- `SCORE` and the footer mean are the *persisted* effective score rows. A run that `eval grade` scores zero without running graders (errored, timed out, never wrote a trace) has no score row, so `runs list` leaves it out of `G graded`; for such a group the two means differ by design. Matching exactly would need a persisted no-score grading marker (a format change, not for this PR).
- Listing keeps duplicate paths as duplicate rows (display); grading dedupes by `fs.realpathSync.native` (mutation).
- One deviation from the plan: `RunSummary` keeps `agentName` (trace's own) and gains `agentLabel` (harness); `displayAgent()` combines them for the column. The runs explorer already has its own label-shortening rule and reads `agentName` as "the trace named itself", so blending inside `summarizeTrace` would have changed the explorer's output.

## Tests

`buildRunsListing` model (rows / silent / total / graded / mean), `displayAgent` (event wins, file label, command-line label unchanged, none), `runs list` over a group / several paths / footer states / silent-only / error / Commander two-path boundary, `eval grade` several paths / duplicate / symlink alias / errored-run denominator, fixture non-truncation, and the packaged cli integration grades and lists two groups with two judge mocks.

## Try it

```
pnpm run build
pnpm run agency eval run test.agency --suite <suite> --out runs/t2
pnpm run agency runs list runs/t2
pnpm run agency eval grade runs/t2 --goal "…"
pnpm run agency runs list runs/t2                       # footer shows the mean
pnpm run agency runs list runs/t2/<one> runs/t1
pnpm run agency eval grade runs/t2 runs/t2/<one> --goal "…"   # <one> graded once
```

### DIFF
--- a/packages/agency-lang/lib/runDirectory/list.ts
+++ b/packages/agency-lang/lib/runDirectory/list.ts
@@ -10,7 +10,11 @@ export type RunSummary = {
   /** The test id from the harness's `run` row; null for an ad-hoc trace. */
   testId: string | null;
   input: string | null;
+  /** The trace's own `agentName` event, when it emitted one. */
   agentName: string | null;
+  /** The harness's agent label from the `run` row (`flags.agent`: an agent
+   *  file target or a command line), unchanged; null for an ad-hoc trace. */
+  agentLabel: string | null;
   startedAt: string | null;
   startedAtMs: number | null;
   durationMs: number;
@@ -49,6 +53,7 @@ function summarizeTrace(
     testId: testIdOf(runRow),
     input: traceInputText(trace, record),
     agentName: record.agentName ?? null,
+    agentLabel: agentLabelOf(runRow),
     startedAt: startedAtMs === null ? null : new Date(startedAtMs).toISOString(),
     startedAtMs,
     durationMs: record.durationMs,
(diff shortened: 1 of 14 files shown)


packages/agency-lang/lib/runDirectory/list.ts
```diff
@@ -88,6 +93,50 @@ export function annotationSummaries(snapshot: RunDirectorySnapshot): Record<stri
...
+    return null;
+  }
+  return label;
+}
+
+/** What a listing shows for "which agent": the trace's own name when it
+ *  emitted one, else the harness label unchanged (a command line is not
+ *  shortened: its basename could be any argument), else null. */
+export function displayAgent(summary: RunSummary): string | null {
+  return summary.agentName ?? summary.agentLabel;
```
egonSchiele, line 108: `setAgentName("")` is valid today (`setAgentName` stringifies its argument, and `lastAgentName` accepts any string), so `summary.agentName` can be the empty string. In that case `??` suppresses the non-empty harness label and the new AGENT column renders blank. The explorer identity path already treats an empty event name as absent; please make this fallback do the same and cover it in `displayAgent` tests.

PR #868, opened 2026-08-19
notes.md replaces the note annotation and agency note (atomic run dirs 3/4)

egonSchiele:
Atomic run directories, PR 3 of 4 (spec: "notes.md replaces the note annotation kind and the agency note command"). Follows #866 and #867.

## What changes

- **`notes.md` is the notes.** `readRunDirectory` returns `snapshot.notes: string | null`: one `readFileSync`, `ENOENT` → null, any other I/O error propagates. The file is a person's, edited with any editor, so it sits outside the writer lock and outside the statelog coherence check (a read during an editor's atomic save may see old text, new text, or null; it never crashes).
- **Removed:** the `note` annotation kind (`NotePayload`, its schema, the `notes` fold field), `recordNote`, and the `agency note` command. An old `note` row fails the schema and is reported by the existing per-line warning; no converter (no back-compat, per spec).
- **Readers:** `RunSummary.noteCount` → `hasNotes` (`notes.md` exists and is not blank; the same for every trace in the directory). `runs list` shows `NOTES yes` or blank. The viewer's per-trace line says `notes · score 0.70 · labeled`. `humanFeedbackFor` puts the trimmed `notes.md` first, then each checklist sign-off note, so the optimizer's reflection step reads the file unchanged.
- **Docs:** `docs/dev/run-directory.md`, `eval-labeling.md` (notes are not under the session lock), `logs-viewer.md`, `runs-explorer.md`, `writing-optimizers.md`, CLAUDE.md pointer.

Not in this PR: node-targeted notes, note text inside the viewer, `label` over groups and the one-trace reader guard (PR 4), `docs/site` (owner: `agency note` removed; write `notes.md` in the run directory instead; `runs list` shows `NOTES yes`).

## Tests

- `runDir.test.ts`: `notes` null without the file, exact text with it (incl. empty), `EISDIR` propagates.
- `annotations.test.ts`: an old `kind: "note"` row is rejected with a warning naming its line; the fold has no `notes` key.
- `list.test.ts`: `hasNotes` false/whitespace-only/true; `annotationSummaryText` → `notes · score 0.75`.
- `commands.test.ts`: a real group of two children, `NOTES` cell `yes` vs blank.
- `humanFeedback.test.ts`: `["<notes.md trimmed>", "<sign-off note>"]`; blank file ignored.
- `gradeRun.test.ts`: notes.md reaches the grading scorecard.
- `mutations.test.ts`: routing/idempotence/orphan and torn-tail tests moved to `recordGradingPass`.
- `tests/integration/cli-tier2`: writes `notes.md`, asserts `NOTES yes` and no annotation rows.

Dropped from the plan: a spy-based "file vanishes mid-read" test, because `vi.spyOn(fs, …)` is not possible under ESM here; the missing-file case is the first test above and `readNotes` has no existence check to race.

## Try it

```
pnpm run build
pnpm run agency eval run test.agency --input hi --out runs/t3
echo "step 3 was slow" > runs/t3/input-1/notes.md
pnpm run agency runs list runs/t3          # NOTES yes
pnpm run agency logs runs/t3/input-1       # trace row shows "notes"
pnpm run agency note runs/t3/input-1 "x"   # unknown command
```


egonSchiele (commented):
Reviewed the notes-file migration and its consumers. The implementation and focused coverage look consistent; two documentation-contract issues should be corrected before merge.

packages/agency-lang/lib/cli/runDirectory/commands.ts
```diff
@@ -19,15 +17,15 @@ function defaultFail(message: string): void {
...
+  return { runsAdd, runsList, logsExtract, fail: defaultFail };
 }
 
 /** commander calls this once per repeat of a flag, accumulating the values. */
 function collectRepeated(value: string, previous: string[]): string[] {
   return [...previous, value];
 }
 
-/** `agency runs add|list` and `agency note`. */
+/** `agency runs add|list`. */
```
egonSchiele, line 28: Removing the command here leaves the user-facing `docs/site/cli/eval.md` advertising `agency note` in the command summary and its Notes section (including a copyable example). After this merges, that documented command always fails. Please update that page to tell users to edit `<runDir>/notes.md` directly, and remove notes from its description of `annotations.jsonl`.

packages/agency-lang/lib/runDirectory/runDir.ts
```diff
@@ -15,9 +15,9 @@ import { readTraces, type Trace } from "./traces.js";
  * A run directory is a plain folder holding one run: `statelog.jsonl` (one
  * trace), `annotations.jsonl`, and optional attachments — the agent's code
  * closure directly under `code/`, a working-directory snapshot directly under
- * `workdir/` with its `workdir.json` sidecar, free-form `notes.md`. This
- * module names its paths and reads it into one coherent snapshot. It never
- * writes; `mutations.ts` does.
+ * `workdir/` with its `workdir.json` sidecar, free-form `notes.md` (read into
+ * `snapshot.notes`). This module names its paths and reads it into one
+ * coherent snapshot. It never writes; `mutations.ts` does.
```
egonSchiele: This overview now includes `snapshot.notes` in “one coherent snapshot,” but `readNotes` and `docs/dev/run-directory.md` correctly establish that notes are sampled independently and are outside the statelog fingerprint check. Please narrow this wording so it says only the statelog/annotation pairing is coherent; otherwise future consumers may rely on a cross-file guarantee the reader does not provide.

packages/agency-lang/lib/cli/runDirectory/commands.ts
egonSchiele: Done in 729791081: docs/site/cli/eval.md no longer mentions agency note. The command summary, the annotations description, the layout listing (notes.md added), the runs list example (NOTES yes) and the Notes section now tell the user to write <runDir>/notes.md with any editor. I left the rest of that page (the pre-#866 layout, the runs list columns) for your broader update.

packages/agency-lang/lib/runDirectory/runDir.ts
egonSchiele: Done in 729791081: the overview now says the statelog and annotations are paired coherently while notes.md is sampled independently, best effort, pointing at readNotes.

PR #869, opened 2026-08-19
label walks groups; a run directory holds one trace (atomic run dirs 4/4)

egonSchiele:
Atomic run directories, PR 4 of 4 (spec §1 "a run directory holds one trace; the reader refuses more" and §4 `label <path…>`). Follows #866, #867, #868. Plan: `docs/superpowers/plans/2026-08-19-atomic-run-dirs-pr4-label-plan.md` (rev 2).

## What changes

**`agency label <path…> --checklist <file> [--annotator <id>]`**
- `lib/eval/label/group.ts` `resolveLabelingGroup(paths)` is the whole discovery rule, run once by the CLI: the `findRunDirectories` walk, aliases of one physical directory dropped (`uniqueRunDirectories`, moved into `findRuns.ts`), one common parent required (a single run's parent is its group), runs that wrote no trace skipped, two physical directories with the same trace id refused naming both, display fields derived. The controller and store receive that one `LabelingGroup`; nothing downstream correlates paths with snapshots.
- Session files move to the group: `<group>/checklists/<id>/{N.json,current.json,drafts/<session>.json}`. A sign-off appends to *that run's* `annotations.jsonl`.
- **No group-wide lock.** Three narrow ones, all one primitive (`acquireOwnedFileLock`, of which `acquireRunDirLock` is now the `.lock` case): per session draft (`drafts/<session>.lock`, held until close: a second process cannot resume the same draft, another annotator opens beside it), per lineage publication (`.publish.lock`, short; the loser of a race fails inside it with the lineage untouched), and the run's writer lock around each append.
- New public mutation `recordChecklistRow({dir, groupDir, row})` replaces `appendAnnotationsUnderLock`: under the run's lock, immediately before the append, it re-checks the grounding (run holds the trace; the group has the revision with that hash; every answer names one of its questions), then appends idempotently and returns `appended|replayed` plus the post-write snapshot (the store's one cache).
- `humanFeedbackFor` finds the lineage in the run's parent. `SessionItem` carries `runDir`.

**The one-run invariant**
- `readRunDirectory` throws on a statelog with more than one trace id, naming the ids and `runs add` to split it.
- `recordCompletedRun` refuses, before writing a byte: a second staged trace, a run row for another trace than the one present (staged or not), a second silent run row beside another trace's.
- Fixture: `writeRunDirectory(run, dir?)` and `writeRunGroup(runs, groupDir?)` (duplicate child names refused before anything is written); the explorer's `writeGradedRun` is one run with two scores (same 0.5 / gates-failed row as before).

**Docs:** `docs/dev/eval-labeling.md` rewritten for groups and the three locks; `run-directory.md` (guard, writer preflight, `recordChecklistRow`, `label` in the commands list); `runs-explorer.md` (Enter on a run opens its trace directly; the "[pick test]" overlay is now reachable only by multi-test rows, which run directories no longer produce — kept, flagged); CLAUDE.md pointers.

Not in this PR: datasets/manifests; copying checklists into runs; `eval run` resume; viewer changes; `docs/site` (owner: `label` takes paths and walks groups; session files under `<group>/checklists/`; a multi-trace `statelog.jsonl` is refused).

## Tests
- `group.test.ts` (new): walk order, one run → parent, aliases, two parents refused, same trace id in two copies refused naming both, silent-only group refused.
- `labelStore.test.ts` (rewritten over a group): sign-off lands in the named run only; cache updated from the mutation; run lock not held between appends (`recordGradingPass` gets in); out-of-session trace and missing revision refused; files under `<group>/checklists/`; one lock per session, none after close; same session refused / other annotator opens; lock released on failed open; publication race.
- `mutations.test.ts`: reader guard; three `recordCompletedRun` refusals byte-for-byte unchanged; `recordChecklistRow` append/replay/snapshot/lock-gone and the four grounding refusals.
- `lock.test.ts`: owned-file lock (exact path, independent paths, exit listener removed). `runDirectoryFixture.test.ts` (new). `controller.test.ts`, `label.test.ts`, `labelCommand.test.ts`, `humanFeedback.test.ts`, `list.test.ts`, `runDir.test.ts`, explorer `rows/loader/run` tests migrated.
- One gradeRun test ("silent run scored 0 beside its neighbour in one directory") was deleted: that shape is now refused; the group-level case is `grade.test.ts`'s errored-run denominator test.

## Try it
```
pnpm run build
pnpm run agency eval run test.agency --suite <suite> --out runs/t4
pnpm run agency label runs/t4 --checklist cl.json      # every run, in order; answer two, quit
pnpm run agency label runs/t4 --checklist cl.json      # resumes at item 3
ls runs/t4/checklists/                                  # lineage + drafts live in the group
cat runs/t4/<one>/annotations.jsonl                     # that run's checklist rows only
cp -r runs/t4/<one> /tmp/y && pnpm run agency runs list /tmp/y    # LABELED travels
pnpm run agency label runs/t4/<one> --checklist cl.json           # one run; group = runs/t4
mkdir /tmp/z && cat runs/t4/<one>/statelog.jsonl runs/t4/<two>/statelog.jsonl > /tmp/z/statelog.jsonl
pnpm run agency runs list /tmp/z                        # refused: holds 2 traces; split with runs add
```


egonSchiele (commented):
Reviewed the group resolver, lock ownership, publication protocol, grounded append mutation, one-run writer preflight, and focused tests. The main paths are well covered, but two supported concurrency cases can leave split or unrecoverable checklist state.

packages/agency-lang/lib/eval/label/controller.ts
```diff
@@ -116,52 +116,45 @@ async function openSession(
...
-    const definition = normalizeDefinition(rawDefinition);
-    if (rawDefinition.checklistId === undefined) {
-      syncChecklistDefinitionIds(args.checklistFile, definition);
-    }
+  // Allocate lineage ids once, and write them back, before the session id is
+  // derived from the checklist. A crash here leaves a complete definition
+  // nobody has published, which the next open simply publishes.
+  const definition = normalizeDefinition(rawDefinition);
+  if (rawDefinition.checklistId === undefined) {
+    syncChecklistDefinitionIds(args.checklistFile, definition);
```
egonSchiele: Concurrent first opens of an ID-less checklist can fork it into two lineages here. Both processes can read the same file, `normalizeDefinition` assigns different random checklist/question IDs, and each atomically writes its version. Because the generated checklist IDs differ, their session locks *and* publication locks also differ, so both stores can open and publish; the last definition-file sync wins while the group retains two unrelated lineages. This is now reachable even within one group because different sessions are intentionally allowed concurrently. Please serialize parse/ID-allocation/write with a lock keyed independently of the not-yet-known checklist ID (for example, the canonical checklist-file path), or use a re-read/CAS that makes one opener adopt the IDs the other wrote, and add a two-opener regression test.

packages/agency-lang/lib/runDirectory/labelStore.ts
```diff
@@ -151,93 +183,106 @@ export function openLabelStore(args: OpenLabelStoreArgs): LabelStore {
...
-        fault: args.fault,
+      // Publication re-reads the current pointer under the lock, so a pending
+      // revision prepared against a parent that has since moved fails there
+      // (`assertParentStillMatches`) without touching the lineage.
+      const publication = acquireOwnedFileLock({
+        lockFile: publicationLockPath(group.dir, pending.revision.checklistId),
+        reportWarning: args.reportWarning,
       });
+      try {
+        return publishPendingRevision({
```
egonSchiele, line 202: The stale-parent failure is safe for the lineage, but it permanently wedges the losing session's draft. The controller saves `pendingRevision` before this call and clears it only after success; after this throws, reopening loads that same pending revision and `recoverChecklist()` immediately calls `publishRevision(pending, ...)` again, producing the same parent-moved error forever. The store race test manually calls `prepareChecklist` after the failure, so it bypasses this persisted-draft recovery path. Please make reopen detect/reconcile a stale pending revision (or surface a recoverable conflict that updates/clears it without losing staged questions), and cover publication race → close → controller reopen.

packages/agency-lang/lib/eval/label/controller.ts
egonSchiele: Done in 1db5b2e47: allocateChecklistIds (controller.ts) serializes id allocation under <group>/checklists/.definition.lock, keyed by the group rather than the not-yet-known checklist id, and re-reads the file under the lock, so the second opener adopts the ids the first wrote (CAS). If the lock is held the opener fails with a clear try-again message. Regression tests in controller.test.ts (concurrent openers): the fault point before-checklist-id-allocation writes another opener's ids into the file; the session adopts cl_other and its question ids and the group has exactly one lineage; plus the held-lock refusal.

packages/agency-lang/lib/runDirectory/labelStore.ts
egonSchiele: Done in 1db5b2e47: rebasedPendingRevision (controller.ts) runs before recoverChecklist publishes. When the pending revision's expected parent no longer matches the lineage's current revision (new store method currentRevision), it re-prepares the staged questions on top of current (rebaseQuestions: every current question, then the staged ones the lineage lacks), saves that as the new pending and publishes it; if current already has exactly those questions the draft adopts current and the pending is cleared. Test: crash after the pending save, another annotator publishes v2 first, reopen rebases to v3 with both questions and no pending left.

packages/agency-lang/lib/runDirectory/findRuns.ts
```diff
@@ -50,3 +50,12 @@ export function findRunDirectories(paths: string[]): string[] {
...
 }
+
+/** The same run directory once, however many ways the walk reached it.
+ *  First appearance wins. Canonical paths (`realpath`), so identity and any
+ *  later mutation target come from one source of truth: the walk follows
+ *  symlinks while classifying, so two spellings can name one directory. */
+export function uniqueRunDirectories(dirs: string[]): string[] {
+  const canonical = dirs.map((dir) => fs.realpathSync.native(dir));
+  return canonical.filter((dir, index) => canonical.indexOf(dir) === index);
+}
```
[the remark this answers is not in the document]
egonSchiele: Done: linear with an object keyed by canonical path (the codebase uses objects over Set).

packages/agency-lang/lib/runDirectory/mutations.ts
```diff
@@ -439,20 +474,89 @@ export function newPassId(): string {
...
+): RecordChecklistRowResult {
+  const result = withWriter(request.dir, options, (paths, snapshot, reportWarning) => {
+    assertChecklistRowGrounded(request.dir, request.groupDir, snapshot, request.row);
+    const before = snapshot.annotationRows.length;
+    const { v: _v, id: _id, createdAt, ...draft } = request.row;
+    appendRows(paths.annotations, [draft], { ...options, now: () => createdAt }, reportWarning);
+    const after = readAnnotations(paths.annotations, reportWarning).length;
+    const outcome: RecordChecklistRowResult["outcome"] = after > before ? "appended" : "replayed";
+    return { outcome };
+  });
```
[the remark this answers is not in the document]
egonSchiele: Done: the outcome comes from the pre-append snapshot (row id present = replayed); no second read.

packages/agency-lang/lib/cli/eval/label.ts
```diff
@@ -129,7 +131,7 @@ export async function label(
   }
 
   const request: LabelRequest = {
-    dir: path.resolve(options.dir),
+    group: resolveLabelingGroup(options.paths, { reportWarning: dependencies.reportWarning }),
     checklistFile: path.resolve(options.checklist),
     annotator: resolveAnnotator(options, dependencies),
```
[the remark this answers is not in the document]
egonSchiele: Done: label() refuses an empty path list with a clear message; tested.

PR #874, opened 2026-08-20
Eval subset selection, fib coding test, and the agent fixes it forced out

egonSchiele:
Everything from the fib-eval debugging arc, in two halves.

## Eval framework (the squashed commit)

- **Agent output contract for `--agent-cmd`**: the agency agent one-shot mode now records `evalValue(prompt)` and `evalOutput(reply)` in the statelog, so `eval grade` sees the reply instead of null.
- **Test descriptions and tags**: `Test` gains optional `description` (human documentation only — never shown to the agent or the judge) and `tags` (selection only).
- **Subset selection**: `eval run --test <glob>` (repeatable, ORed) and `--tags a,b` (ANDed). New `agency eval ls` previews exactly what a run with the same flags would select, because both go through the one `selectTests`.
- **First coding test** (`evals/agency-agent/fib`): the agent writes a Fibonacci function in Agency. Graded by a shipped agency-test harness that compiles the solution with `std::agency` (stdlib-only imports) and runs it in the sandboxed subprocess with every effect rejected — no Docker needed. The grader snapshots the harness files and copies only fib.agency from the agent workdir, so agent tampering with its seeded harness copy changes nothing.

## Fixes the fib eval forced out (the eight commits on top)

- **Policy `.` = launch directory**: a `dir` pattern of `"."` / `"./sub/**"` / `"{.,./**}"` resolves against the process launch directory, as a second match attempt that never removes an existing match. A static `~/.agency-agent/policy.json` can now scope writes to "wherever the agent runs" — including eval staging workdirs — instead of hard-coding an absolute path.
- **Live statelog paths**: `eval run` prints each test's staging statelog path at start (sequential: as each test begins; parallel: all up front before the status board), so a live run can be watched without killing it. The sequential loop moved into `runSequential`, mirroring `runPool`.
- **safeBash replaces raw bash in `shellTools()`**: recognized commands raise their narrow effects (a write raises `std::write`, a git read its git effect), so policies can pre-approve what raw bash never could; raw bash was also an automatic dead end in headless runs. The coding prompt now tells the agent to stop and name a missing permission instead of hunting for workarounds.
- **Generated Agency code lands in the file the task names**: `codeAgent` decides the save target before generation, tells the writer to produce the module contents (not a program that performs the task at runtime), and writes the result itself. `agencyCodingAgent` gains `requireMain` (default true, planner unchanged) since a library module has no business being forced to carry a `node main`.
- **Dev note** `docs/dev/approval-policies.md`: the three policy-matching semantics that repeatedly bit us (absolutized dirs, the `.` resolution, picomatch's `**`-skips-dot-segments rule).

## Verified

- Policy tests 28/28 (new dot-pattern cases), toolkits agency tests 8/8 (safeBash cwd-exemption documented in-test), runSuite 18/18, tsc, prettier, structural linter all green. `make` clean.
- Live fib runs: run 3 proved the write route and policy resolution (file appeared in the workdir); run 4 produced a correct `export def fib` module that the harness grader passes.

## Known remaining (next PR)

Agent self-verification via `agencyCli` fails in eval subprocesses (`spawn agency ENOENT` — no `agency` on PATH). The planned fix is a `test()` in `std::agency` over the same subprocess machinery as `run()`, with its own effect, which also collapses the fib grading harness to a few lines.


packages/agency-lang/lib/runtime/policy.ts
```diff
@@ -64,6 +64,19 @@ function stripDotSlash(s: string): string {
...
+// Tools absolutize the dir they put in interrupt data, so a literal `.` in
+// a policy file could never match those; resolving it lets a static policy
+// say "the current directory, whatever it is" instead of hard-coding an
+// absolute path. The replacement covers `.` standing alone, at the start of
+// a path (`./sub/**`), and as a brace alternative (`{.,./**}`). Note the
+// same caveat as every dir glob: `**` does not descend into dot-led
+// subdirectories (picomatch's dot rule), though a launch directory whose
+// own path contains dot segments is fine — those sit in the literal prefix.
+function resolveDotDirPattern(pattern: string): string {
+  return pattern.replace(/(^|\{|,)\.(?=$|\/|,|\})/g, `$1${process.cwd()}`);
```
egonSchiele: [P1] Escape the cwd before embedding it in a picomatch pattern. As written, filesystem-legal glob characters in the launch path become active policy syntax and broaden approval. For example, when cwd is a directory literally named `cwd*`, a policy rule with `dir: "."` approves an interrupt whose `data.dir` is the sibling `cwd-neighbor`; I reproduced `checkPolicy(...)` returning `{ type: "approve" }`. This is a safety boundary, so the cwd prefix needs to be treated as literal while preserving glob syntax only in the user-supplied suffix. A regression test should chdir into a path containing `*` (and ideally brackets/braces) and assert that a sibling does not match.

packages/agency-lang/stdlib/agents/lib/toolkits.agency
```diff
@@ -93,10 +94,15 @@ export def writableFileTools(): any[] {
...
+  Return tools that run commands: safeBash for a shell pipeline, exec for
+  a single binary with arguments. safeBash rather than raw bash: it
+  parses the command and raises the narrowest effects that describe it
+  (a recognized write raises std::write, a git read raises its git
+  effect), so a policy can pre-approve what raw bash never could. It
+  needs no useAgentCwd binding — its empty cwd already resolves to the
+  agent working directory (resolveCwd in std::safeBash).
   """
-  return [bash.partial(useAgentCwd: true), exec.partial(useAgentCwd: true)]
+  return [safeBash, exec.partial(useAgentCwd: true)]
```
egonSchiele, line 105: [P1] Please update the existing tool-contract tests for this rename. `tests/agency/agents/toolGaps.agency` still requires both verifiers to expose a tool named `bash`, so this change makes `agencyVerifierCanSeeFiles` and `verifierCanComputeAndReadBytes` return false. The targeted `pnpm run agency test tests/agency/agents/toolGaps.test.json` fails 2/8 locally, matching the failed `agency-tests (4)` CI shard. These assertions should require `safeBash` (while retaining the capabilities they were intended to pin).

packages/agency-lang/evals/agency-agent/fib/graders.ts
```diff
@@ -0,0 +1,103 @@
...
+        return fail(testOutput(error));
+      }
+    } finally {
+      fs.rmSync(dir, { recursive: true, force: true });
+    }
+  }
+}
+
+function moduleDir(): string {
+  return path.dirname(new URL(import.meta.url).pathname);
```
egonSchiele: [P2] Decode the file URL with `fileURLToPath(import.meta.url)` instead of reading `.pathname`. `.pathname` leaves URL escapes intact (`file:///tmp/agency%20lang/...` becomes `/tmp/agency%20lang/...`), so every live fib grade fails to find its harness when Agency is checked out or installed under a path containing a space (and it is also incorrect for platform-specific file URLs). The rest of this repo generally uses `path.dirname(fileURLToPath(import.meta.url))` for this case.

egonSchiele (commented):
I found three actionable issues: one policy-scope widening, the current execution-test failure, and a path-decoding bug in the new grader.

packages/agency-lang/lib/eval/runTypes.ts
```diff
@@ -8,6 +8,17 @@ export type Test = {
   /** Stable identifier. Auto-derived when omitted: the loader generates one
    *  via nanoid, the optimizer derives it positionally (`input-<index>`). */
   id?: string;
+  /** Why the test exists, for the person reading a grade report or a run
```
egonSchiele: Long comments like this hinder the readability because users are unlikely to read them. Is this comment necessary? I would think a description field would be fairly self-explanatory.

packages/agency-lang/stdlib/agents/lib/toolkits.agency
egonSchiele, line 105: should this have `useAgentCwd: true` as well?

packages/agency-lang/stdlib/agents/coding.agency
```diff
@@ -44,8 +44,8 @@ static const systemPrompt = """You are an expert software engineer. Solve the ta
...
 - NEVER fabricate output you did not actually produce.
 
-## Avoid bash and exec when possible
-You have been given bash and exec tools. **Avoid using them if at all possible**. These tools should be a last resort. While the use of other tools is auto-approved, every time you use bash or exec, it will require human approval, which will greatly slow down your work. So *always use another tool when possible* and only use bash/exec if no other tool will do the job.
+## Avoid safeBash and exec when possible
+You have been given safeBash and exec tools. **Avoid using them if at all possible**. These tools should be a last resort. safeBash translates a recognized command (a simple write, a git read) into a specific pre-approvable action, but anything it does not recognize needs broad shell approval — which requires a human and will greatly slow down your work, or fails outright in a non-interactive run. So *always use another tool when possible* and only use safeBash/exec if no other tool will do the job. If a permission is rejected and the task cannot be done without it, do not hunt for workarounds: stop and report exactly which permission is missing.
```
egonSchiele: > safeBash translates a recognized command (a simple write, a git read) into a specific pre-approvable action, but anything it does not recognize needs broad shell

Why is it necessary to tell the agent this?

egonSchiele:
All review points addressed in f4514c403:

- **[P1] cwd glob-escaping** (`policy.ts`): the substituted launch path is now escaped, so glob characters in a directory name stay literal — your `cwd*` sibling repro is a regression test (plus brackets/braces), via the now-exported `resolveDotDirPattern` with an injected cwd rather than chdir.
- **[P1] toolGaps rename** (the failing `agency-tests (4)` shard): verifier assertions now require `safeBash`; the reviewer assertion pins that reviewers have neither `bash` nor `safeBash`. 8/8 locally.
- **Save-target validation** (`code.agency`): the model-supplied filename is refused if absolute, traversing (`..`), backslashed, or drive-lettered — `unsafeSaveTarget`, exported for tests.
- **[P2] `moduleDir`** (`graders.ts`): `path.dirname(fileURLToPath(import.meta.url))`, matching the repo convention.
- **`collectCommaSeparated`**: values are trimmed, so `--tags easy, hard` selects.
- **`printLiveStatelogPaths`**: skips ids that fail `testIdProblem` — an invalid id never stages, so no path is printed for it.
- **`description` docstring**: shortened to two lines.
- **Coding prompt**: dropped the safeBash-mechanics sentence — the agent only needs "last resort, may need approval you cannot get; if rejected and essential, stop and name the missing permission".

**On "should safeBash have `useAgentCwd: true` as well?"** — no, and it can't: `safeBash` has no `useAgentCwd` parameter. Its `cwd` defaults to `""`, which `resolveCwd` (std::safeBash) resolves to the agent working directory — the same place the binding would point. The cwd-bound toolkit test documents the exemption for the same reason.

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/code.agency
```diff
@@ -180,19 +180,79 @@ def consultExpert(task: string, agencyTask: boolean): Result<ExpertGuidance> {
...
+"""
+
+  const target: SaveTarget = llm(prompt)
+  return target.filename
+}
+
+/** A save target comes from a model, so it is validated like any other
+  untrusted input: it must stay inside the agent working directory.
+  Exported for tests. */
+export def unsafeSaveTarget(name: string): boolean {
```
egonSchiele: [P1] This lexical check still does not establish the stated containment invariant. In particular, `unsafeSaveTarget("~/payload.agency")` returns false, but `write` routes the filename through `resolvePath`/`expandPath`, so it resolves to `$HOME/payload.agency` rather than beneath the agent cwd (I confirmed `resolvePath("/tmp/agent-workdir", "~/payload.agency")` resolves into my home directory). A relative path through an existing symlink can similarly escape. Because the policy interrupt still reports the agent cwd as `dir`, a cwd-scoped rule may approve before the actual destination is resolved. Please validate the resolved destination with containment enforcement (including symlink handling), rather than trying to enumerate unsafe spellings.

packages/agency-lang/lib/runtime/policy.ts
```diff
@@ -64,6 +64,28 @@ function stripDotSlash(s: string): string {
...
+// a policy file could never match those; resolving it lets a static policy
+// say "the current directory, whatever it is" instead of hard-coding an
+// absolute path. The replacement covers `.` standing alone, at the start of
+// a path (`./sub/**`), and as a brace alternative (`{.,./**}`). Note the
+// same caveat as every dir glob: `**` does not descend into dot-led
+// subdirectories (picomatch's dot rule), though a launch directory whose
+// own path contains dot segments is fine — those sit in the literal prefix.
+// Exported for tests, which inject the cwd.
+export function resolveDotDirPattern(pattern: string, cwd: string = process.cwd()): string {
+  return pattern.replace(/(^|\{|,)\.(?=$|\/|,|\})/g, `$1${escapeForGlob(cwd)}`);
```
egonSchiele: [P2] Use a replacement callback here so `$` sequences in a legal cwd are inserted literally. Passing the cwd as part of a replacement string makes `String.replace` interpret `$&`, `$` + backtick, `$'`, and `$$`; for example, cwd `/tmp/a$&b` turns `./**` into `/tmp/a.b/**`, which no longer matches files under the actual launch directory. The new glob escaping tests cover picomatch metacharacters but not replacement-string metacharacters. `pattern.replace(regex, (_match, prefix) => prefix + escapeForGlob(cwd))` avoids this second interpretation.

egonSchiele (commented):
Re-reviewed the updated head. The original three findings are fixed, but I found two remaining bypass/correctness cases in the follow-up validation and escaping changes.

egonSchiele:
Both round-two points addressed in 024e53ab5:

- **[P1] Resolved-destination containment**: you're right that enumerating unsafe spellings can't establish the invariant — `~` expansion and symlinks both defeat it. The lexical `unsafeSaveTarget` is gone. `solve()` now asks the new `std::fs` `containedIn(filename, root)`, a thin Agency wrapper over the existing symlink-aware `assertContained` (exposed as `isContainedIn`): it resolves the name the same way `write` will — `~` expanded, relative against the agent working directory, a not-yet-existing file resolved through its nearest existing ancestor's realpath — and refuses anything that lands outside. Tests cover the `~/payload.agency` case you reproduced, `..`, absolute paths, and the empty-root fail-closed path.
- **[P2] Replacement callback**: done, with a regression test asserting a cwd of `/tmp/a$&b` produces a pattern that matches files under the real directory and not under `/tmp/a.b`.

packages/agency-lang/lib/stdlib/assertContained.ts
```diff
@@ -127,3 +127,18 @@ async function realpathOrSelf(p: string): Promise<string> {
...
+
+/**
+ * Boolean form of `assertContained` for callers that branch rather than
+ * throw: true iff `target` resolves inside `root` (symlink-aware, with the
+ * same nearest-existing-ancestor handling for targets that do not exist
+ * yet). Relative targets resolve against `root` itself.
+ */
+export async function isContainedIn(target: string, root: string): Promise<boolean> {
+  try {
+    await assertContained(target, [root], root);
```
egonSchiele: [P1] `assertContained` does not make this safe for writes through a dangling symlink. Repro: create `root/payload.agency` as a symlink to a nonexistent `outside/payload.agency`; `isContainedIn("payload.agency", root)` returns true because `fs.realpath` fails and `realpathOrLexicalAncestor` resolves only the containing `root`, then lexically re-appends the symlink name. Writing through that path follows the link and creates the outside file—I reproduced both `allowed: true` and the outside file creation. The existing symlink test only covers an existing target, so it misses this case. Also, this check runs before the potentially long `agencyCodingAgent` call, leaving time to swap an initially safe path for a symlink before `write`. Please enforce containment at the write operation (including dangling final-component symlinks), rather than relying on a separate preflight whose result can become stale.

egonSchiele (commented):
Re-reviewed the latest head. The replacement-callback fix is correct, but the new containment preflight still has a confirmed symlink bypass.

egonSchiele:
Round three addressed in d1eec466d — enforcement moved into the write operation, as requested:

- `write`/`writeBinary` gain `allowedPaths`; when set, `_writeBytes` takes a containment-enforcing path: `assertContained` pins the resolved parent chain, and the file is opened with **`O_NOFOLLOW`**, so the kernel refuses a final-component symlink outright — your dangling repro, an existing link, or one swapped in after any earlier check. That closes both holes at once: the ancestor-fallback blind spot (no preflight reasoning about the final component at all — the kernel decides at open time) and the check-to-write staleness window. Without `allowedPaths`, behavior is byte-for-byte unchanged.
- The code agent's save is now `write(target, source, useAgentCwd: true, allowedPaths: [workRoot()])`; the preflight is gone, and with it the round-two `std::fs containedIn` / `isContainedIn` API — its only consumer was the preflight this replaces.
- Tests (`writeContained.test.ts`): your exact dangling-symlink repro asserting the outside file is never created; the existing-symlink case (refused a layer earlier, by realpath containment); `..`, absolute, and `~` escapes; append/create-only semantics on the contained path; and a no-`allowedPaths` control asserting no new restriction.

One honest limitation, stated rather than hidden: `O_NOFOLLOW` guards the final component, and `assertContained` realpaths the parent chain at write time, but a *directory* symlink swapped into the parent chain in the microseconds between that realpath and the `open` is still theoretically winnable — fully closing it needs `openat2(RESOLVE_BENEATH)`, which Node doesn't expose. That's the same residual race every allowedPaths-style check in the codebase carries; the attacker here (a generated filename) also can't create symlinks, so the practical exposure is nil.

packages/agency-lang/lib/stdlib/builtins.ts
```diff
@@ -259,25 +261,66 @@ async function _writeBytes(
...
+ *  the file itself is opened with O_NOFOLLOW so the kernel refuses a
+ *  final-component symlink outright — including a dangling one, and one
+ *  swapped in after any earlier check. */
+async function writeContained(
+  filePath: string,
+  data: string | Buffer,
+  mode: WriteMode,
+  allowedPaths: string[],
+): Promise<boolean> {
+  await assertContained(filePath, allowedPaths);
```
egonSchiele: [P1] `assertContained` does not actually pin the parent chain: it validates pathnames, returns, and `openFile` resolves the original pathname again. `O_NOFOLLOW` protects only the final component. A concurrent process can let this check pass for `root/sub/file`, then rename `root/sub` and replace it with a symlink to `outside` before line 289; the final `file` is not a symlink, so the open creates `outside/file`. This is the same TOCTOU class the change intends to close, just moved to an intermediate component. The comment claiming the realpathed parent chain is pinned is therefore not true. The contained write needs descriptor-relative traversal/opening (openat-style, with each directory pinned/no-follow) or an equivalent atomic primitive; a pathname check followed by pathname open cannot provide this guarantee.

packages/agency-lang/lib/stdlib/builtins.ts
```diff
@@ -259,25 +261,66 @@ async function _writeBytes(
...
+  const { O_WRONLY, O_CREAT, O_TRUNC, O_APPEND, O_EXCL, O_NOFOLLOW } = fsConstants;
+  const modeFlag = mode === "append" ? O_APPEND : mode === "create-only" ? O_EXCL : O_TRUNC;
+  let handle;
+  try {
+    handle = await openFile(filePath, O_WRONLY | O_CREAT | O_NOFOLLOW | modeFlag);
```
egonSchiele: [P1] This silently loses final-component protection on Windows, where Node does not expose `fs.constants.O_NOFOLLOW`. In the bitwise expression, `undefined` coerces to zero, so the call becomes an ordinary open that follows the dangling symlink which `assertContained` accepted via its ancestor fallback. The new tests run on POSIX and therefore cannot catch this. Since this codebase explicitly supports Windows path semantics, either fail closed when `O_NOFOLLOW` is unavailable or use a platform-specific no-follow primitive; otherwise the documented `allowedPaths` guarantee does not hold on Windows.

egonSchiele (commented):
Re-reviewed the latest containment-at-write implementation. The stable final-component symlink case is fixed on POSIX, but two safety gaps remain.

egonSchiele:
Course correction, agreed with the owner: the three rounds of containment fixes were treating one caller's symptom of a system-wide bug, and userspace path checks next to the write cannot hold the invariant anyway. `714587658` removes the bespoke machinery entirely (allowedPaths, O_NOFOLLOW path, the code agent's preflight — which also moots the Windows and parent-chain-race threads).

The real bug both P1 chains kept pointing at: **file interrupts report the pre-resolution `dir`, so the permission question can lie about the destination** (`{dir: "/tmp/workdir", filename: "~/payload.agency"}` gets approved by a workdir-scoped rule and lands in the home directory). The durable fix is to resolve the destination *before* raising, so the policy judges the truth for every file operation with no new machinery. That is specced for its own PR: `packages/agency-lang/2026-08-20-resolved-interrupt-destination-spec.md` (in this branch's tree), covering the resolver, all eleven raise sites, the stated threat-model boundary for post-approval symlink races, and tests.

What remains in this PR for the save path is a plain policy-mediated `write` — the same protection every other agent write has today, upgraded for free when the spec lands.

PR #875, opened 2026-08-20
Contain single-file operations within their directory

egonSchiele:
Implements `2026-08-20-contained-filename-spec.md` per `2026-08-20-contained-filename-plan.md` (both ride in the first commit). A model-chosen filename can no longer escape the directory the interrupt reported — not through absolute paths, `~`, `..`, or stable symlinks.

## The contract

For the nine scoped single-file wrappers (`read`, `write`, `readBinary`, `writeBinary`, `edit`, `typecheckFile`, `writeAST`, `loadTemplate`, `formatFile`), one shared `prepareContainedPath` canonicalizes `dir`, normalizes `filename`, and rejects every stable escape **before the interrupt exists** — no policy or human can approve an escape, because no escape request is ever raised. The interrupt and the execution consume the same prepared values. Only `ENOENT` is treated as "does not exist yet": dangling links, loops, non-directories, and permission errors all fail closed.

The migration rule doubles as the design principle: **the destination belongs in `dir`**, because `dir` is the field the judge reads. Escape errors teach it verbatim ("To write somewhere else, pass that directory in dir."), and the tool docstrings state it up front.

## safeBash: the one-source exception

safeBash has no trusted `dir` (the whole command is untrusted), so its redirect writes instead resolve the complete target and report the **real parent** as `dir` — the policy judges the true destination. `prepareRedirectWrite` owns all the choreography; `writeEffect`/`writeExecution` derive both plan sides from one `WritePayload`, so payload/execution parity holds by construction. Dangling links and loops fall back to the broad `std::bash` question.

One behavior pinned as found, deviating from the plan's expectation: the bash word grammar refuses an **unquoted** `~` outright, so that spelling always takes the broad `std::bash` question rather than expanding — strictly safer than the planned expansion. Quoted tildes parse and stay literal beneath the cwd, as in bash.

## Caller migrations (the spec inventory)

- `promptFile` loads the coordinator prompts dir-anchored; a new test pins both prompts non-empty (the empty-string fallback made breakage silent).
- `dirname-paths` flips `readUpward` to pin the rejection, adds the dir-based sibling, and migrates the absolute-filename node.
- `stdlib-destructive` keeps its destructive-marker purpose via a contained filename below a missing parent (same `[true,false]` expectation).

## Policy side

`resolveDotDirPattern` realpaths the launch directory (lexical fallback), so a `{.,./**}` rule and the canonicalized payloads share one path identity under symlinked checkouts — wired through `checkPolicy` in tests, not just the resolver.

## Verified

- 94 vitest cases across the resolver, redirect resolver, policy, and containment suites (1 Windows-only skip); every fail-closed branch pinned (ELOOP, ENOTDIR, mocked EACCES, dangling chains).
- Agency execution tests: contained-filenames 17/17 (every escape proves the handler was never reached via a HANDLER_REACHED sentinel), safeBash 21/21, dirname-paths 7/7, stdlib-destructive, prompt guard, read/write-binary, agent-cwd 13/13, generatedProgram — all green.
- Agency-JS fixtures: recording handlers prove payloads carry prepared values through symlinked dirs for all nine wrappers; real-symlink safeBash redirects land on real targets with truthful plans; the end-to-end policy fixture drives the full contract through a noninteractive `cliPolicyHandler` (9/9 assertions, including no-file-created proofs and the symlink-as-dir case).
- `make`, tsc, prettier, structural linter, text guard all green. The one typecheck warning (`AG4004 _run`) pre-exists on main.

## Out of scope (tracked in the spec)

`applyPatch`, `copy`/`move`/`remove`, `ls`/`grep`/`glob` leaf symlinks, the wider path-bearing-effect audit, and post-approval symlink races (threat-model boundary documented in `docs/dev/approval-policies.md`).

### DIFF
--- a/packages/agency-lang/lib/stdlib/prepareContainedPath.ts
+++ b/packages/agency-lang/lib/stdlib/prepareContainedPath.ts
@@ -1,0 +1,116 @@
+import fs from "fs/promises";
+import path from "path";
+import { resolveDir } from "./resolveDir.js";
+import { expandPath } from "./expandPath.js";
+import { isContained } from "./assertContained.js";
+
+export type ContainedPath = {
+  dir: string;
+  filename: string;
+};
+
+export type FileOperation = "read" | "write";
+
+/**
+ * Prepare a (dir, filename) pair for a scoped single-file wrapper: the
+ * returned values go into BOTH the interrupt payload and the execution.
+ * Guarantees the spec's containment contract: the resolved file operand
+ * stays inside the resolved dir operand, or this throws before any
+ * interrupt exists. See 2026-08-20-contained-filename-spec.md.
+ */
+export async function prepareContainedPath(
+  dir: string,
+  filename: string,
+  operation: FileOperation,
+): Promise<ContainedPath> {
+  if (dir.trim() === "") {
+    throw new Error(`${operation} refused: dir must not be empty.`);
+  }
+  // Steps 1-2: expand, absolutize, and realpath dir. It must exist.
+  const baseDir = await resolveDir(dir);
+  const realRoot = await fs.realpath(baseDir);
+
+  // Steps 3-4: expand filename; absolute (including ~-led) is an escape.
+  const expanded = expandPath(filename);
+  if (path.isAbsolute(expanded)) {
+    throw escapeError(operation, filename, realRoot, expanded);
+  }
+
+  // Steps 5-6: lexical resolution and lexical containment.
+  const lexical = path.resolve(realRoot, expanded);
+  if (!isContained(lexical, realRoot)) {
+    throw escapeError(operation, filename, realRoot, lexical);
+  }
+
+  // Steps 7-9: resolve through existing symlinks (dangling ones fail
+  // closed) and check the resolved target too.
+  const resolved = await resolveExistingStrict(lexical);
+  if (!isContained(resolved, realRoot)) {
+    throw escapeError(operation, filename, realRoot, resolved);
+  }
+
+  // Step 10: the real dir plus the normalized relative filename. The
+  // relative form deliberately keeps in-root symlink names (spec: the
+  // payload guarantees containment, not leaf-target naming).
+  return { dir: realRoot, filename: path.relative(realRoot, lexical) };
+}
+
+/** Walk the target path component by component. Existing components are
+ *  realpathed (following healthy symlinks); a component whose lstat says
+ *  "symlink" but whose realpath fails is dangling and fails closed; a
+ *  component that simply does not exist ends the walk, and the remaining
+ *  tail is appended lexically (missing intermediates are contained, not
+ *  dangling).
+ *
+ *  This deliberately does not reuse assertContained.ts's
+ *  realpathOrLexicalAncestor: that older helper treats every realpath
+ *  error as a missing path, while this safety boundary may treat only
+ *  ENOENT as a lexical tail — loops, non-directories, permission and I/O
+ *  failures all fail closed. Do not add a second strict walk elsewhere;
+ *  reuse this one. */
+async function resolveExistingStrict(target: string): Promise<string> {
+  const parsed = path.parse(target);
+  const segments = target.slice(parsed.root.length).split(path.sep);
+  let current = parsed.root;
+  for (let i = 0; i < segments.length; i++) {
+    const next = path.join(current, segments[i]);
+    let stat;
+    try {
+      stat = await fs.lstat(next);
+    } catch (error) {
+      // Only ENOENT means the rest is a lexical tail. Permission, I/O,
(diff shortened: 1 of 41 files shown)


packages/agency-lang/stdlib/fs.agency
```diff
@@ -51,25 +51,33 @@ export def edit(filename: string, edits: Edit[], dir: string = ".", useAgentCwd:
...
   @param dir - The directory to resolve the filename against
   @param useAgentCwd - When true, resolve a relative path against the agent working directory if one is set
   """
   if (useAgentCwd) {
     dir = applyAgentCwd(dir)
   }
-  const preview = _previewEdit(dir, filename, edits)
+  // Prepared BEFORE the preview: the preview, the interrupt payload, and
+  // the eventual write must all refer to the same spelling.
+  const prepared = try _prepareContainedPath(dir, filename, "write")
```
egonSchiele, line 66: The new preparation step rejects an absolute `filename` before raising `std::edit`, but `tests/agency-js/stdlib/std-fs-edit-reject/test.js` still passes an absolute temp path to `runEdit`. As a result `runEdit` now returns a preparation failure, and the fixture calls `respondToInterrupts(r.data, ...)` on an object rather than an interrupt array. I reproduced the current `agency-tests (4)` failure locally with `node ./dist/scripts/agency.js test js tests/agency-js/stdlib/std-fs-edit-reject`. Please migrate this fixture to pass `basename(path)` as `filename` and `dirname(path)` as `dir` (including the agent wrapper parameter), so it continues to exercise rejection of the actual `std::edit` interrupt rather than the new containment check.

packages/agency-lang/lib/stdlib/prepareContainedPath.ts
```diff
@@ -0,0 +1,142 @@
...
+  return new Error(
+    `${operation} refused: filename "${filename}" is outside dir "${root}" ` +
+      `(it resolves to "${landed}"). To ${operation} ${preposition}, pass that ` +
+      `directory in dir.`,
+  );
+}
+
+export type TildeMode = "expand" | "literal";
+
+/**
```
egonSchiele, line 114: Long comments hurt readability because users are less likely to read them. Please evaluate the long comments you have written and see if they are genuinely helpful, or if the same information can be expressed via code. If they are needed, see if you can make them more concise and readable.

packages/agency-lang/tests/agency-js/contained-filename-wrappers/agent.agency
```diff
@@ -0,0 +1,188 @@
+import { write, read, writeBinary, readBinary } from "std::index"
```
egonSchiele, line 1: Please be careful when writing tests. Make sure you don't write any tests where failure is catastrophic. For example, if you wrote a test, the check that deleting the user's home dir doesn't work, and the test failed, it would delete the home dir, which would be catastrophic. For any tests you have written, would failure be catastrophic?

egonSchiele:
All four points addressed in b07b2c4da, plus the second CI failure:

- **`std-fs-edit-reject`** (the `agency-tests (4)` failure): migrated exactly as you described — `basename(path)` as `filename`, `dirname(path)` as `dir`, agent wrapper parameter included — so it exercises rejection of the actual `std::edit` interrupt. Passing locally.
- **`agency-tests (3)` — `handler-fork-isolation`** was a real ripple worth explaining: preparation added a genuine fs `await` between a wrapper call and its raise, and that window handed the event loop to concurrent branches — a fork sibling could complete and pop its handler before the interrupt existed, so the read fell through to a default rejection with no ASKED line. Fix at the root rather than in the test: preparation now uses sync fs internally (async signatures kept so throws stay rejections), which closes the window entirely. The test passes byte-for-byte unchanged, and a comment in `prepareContainedPath.ts` records why the module must stay sync.
- **`tildeRedirectQuoting` description** now says what the node asserts: an unquoted tilde cannot be narrowed and takes the broad bash question.
- **Long comments** trimmed across `prepareContainedPath.ts`, `safeBash.agency`, and `actions.agency` — each now one to three lines carrying only what the code cannot say.
- **Catastrophic-failure audit**, since you asked the question directly: the dangerous shape here is "if containment breaks, the escape probe performs the escape." Two probes could have overwritten a fixed-name file in the home directory on failure. Hardened: every escape write now uses `create-only` (or `writeAST`'s `overwrite: false`) where the API offers a non-destructive mode, and the tilde probes use improbable dotted names (`~/.agency-cf-escape-probe-3f9d.*`). `formatFile` has no non-destructive mode, so its probe relies on the improbable name alone. The e2e policy fixture already used a per-run random home probe with pre-existence assertion and cleanup; the sandbox fixtures only ever touch their own mkdtemp trees. Nothing anywhere deletes, and worst-case failure is now "create one stray improbably-named file," which the failing test itself would flag.

PR #878, opened 2026-08-22
Sandboxed compile with validated open imports for std::agency (2/4)

egonSchiele:
Layer 2 of 4, split out of #876. Base: `adit/split-1-test-format` (layer 1); retarget to `main` once that merges.

## What this does

`std::agency` `compile()`, `runCode()` and `runFile()` previously allowed only `std::` imports. This PR opens imports but validates them, and is the security-sensitive piece of the stack.

**The invariant:** the transitive import closure of sandboxed code contains nothing but Agency source and `std::` modules. `lib/compiler/closureValidator.ts` walks the raw import graph and refuses, listing every violation: TS/JS files, Node builtins, `pkg::` packages (not supported in the sandbox yet), compile-time splices (refused before any generator could run, since generators execute in the compiling process), and local imports that are absolute, escape `dir`, or go through a symlink.

**The mirror:** compilation consumes a private copy of exactly the bytes the validator read (`lib/compiler/compileValidatedClosure.ts`). Every validated file is written at its own relative path under a fresh 0700 temp dir and compiled there, so `compileSource` never touches the caller's directory between validation and compile. The relative-only and no-symlink rules are what make this work without rewriting a single import.

**Runtime:** `CompiledProgram` carries every non-entry module's compiled JS (`modules`); `materializeCompiledScript` writes them beside the entry script at fork time, re-checking containment because compiled values can be hand-built.

**CLAUDE.md** gains a line: do not add code to support symlinks; refuse them instead.

**stdlib:** `compile(source, dir)`, `runCode(..., dir)`, `runFile` gain dir-local imports. `dir` is the compile-time import boundary; `cwd` stays the run-time working directory.

## Reviewer's question

Is the invariant enforced everywhere, and is the mirror really what gets compiled? The evidence: `closureValidator.test.ts` (19) and `compileSandboxed.test.ts` (9, including "swap the validated file for one with a splice after validation; the generator never runs" with a positive control), plus `tests/agency/agency-sandbox-imports` (6 cases).

## Revision history

The first version supported symlink aliases inside `dir` by rewriting each import to its mirrored target, which needed the parser to record `modulePathLoc` on import statements, and supported `pkg::` packages by walking them under their own root and symlinking them into the mirror. Both were dropped (about 550 lines): symlinks are refused, `pkg::` can return as its own change.

## Verification

Layer 1+2 unit suites: 353 passed. `agency-sandbox-imports`: 6/6. Guards pass on the full stack.


packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -700,6 +704,17 @@ export function materializeCompiledScript(compiled: { moduleId: string; code: st
   mkdirSync(tempDir, { recursive: true });
   const scriptPath = path.join(tempDir, `${compiled.moduleId}.js`);
```
egonSchiele: The entry is always materialized at the temp root, but non-entry modules keep paths relative to the sandbox root. This breaks a file entry below that root: `runFile(dir, "sub/main.agency")` with `sub/main.agency` importing `./helper.agency` emits `./helper.js`; the helper is written to `sub/helper.js`, while the entry runs from `<temp>/<moduleId>.js` and looks for `<temp>/helper.js`. Please preserve the entry's relative path during materialization (or rewrite its imports accordingly) and add an execution test for a nested entry with a sibling import.

packages/agency-lang/lib/compiler/closureValidator.ts
```diff
@@ -0,0 +1,295 @@
...
+    }
+    this.walkFile(canonical, label);
+  }
+
+  walkFile(filePath: string, importedFrom: string): void {
+    if (this.visited[filePath]) return;
+    this.visited[filePath] = true;
+    let source: string;
+    try {
+      source = fs.readFileSync(filePath, "utf-8");
```
egonSchiele: This does not enforce the documented regular-file invariant before reading. `realpathSync` accepts non-regular files, so an imported `.agency` FIFO can block `readFileSync` indefinitely in the host compiler, outside the subprocess sandbox. Please open with no-follow semantics where available, `fstat` the opened descriptor and require `isFile()`, then read from that same descriptor (which also avoids a check/read race). Add a non-regular-file regression test structured so it cannot hang the test process.

egonSchiele (commented):
Two correctness and sandbox-hardening issues found.

packages/agency-lang/lib/compiler/compileSandboxed.test.ts
```diff
@@ -0,0 +1,251 @@
+import { describe, test, expect, vi, afterEach } from "vitest";
+import * as fs from "fs";
+import * as path from "path";
+import * as os from "os";
+import { compileSandboxed, compileValidatedClosure } from "./compileSandboxed.js";
+import { validateClosure } from "./closureValidator.js";
+import { compileSource } from "./compile.js";
+import { materializeCompiledScript } from "../runtime/ipc.js";
+import { safeDeleteDirectoryWithin } from "../utils.js";
+import { createRequire } from "module";
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 8c766f5: both imports removed.

packages/agency-lang/lib/runtime/ipc.ts
```diff
@@ -700,6 +704,17 @@ export function materializeCompiledScript(compiled: { moduleId: string; code: st
...
   writeFileSync(scriptPath, compiled.code, "utf-8");
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 8c766f5: `moduleId` must match `/^[A-Za-z0-9_-]+$/` (checked before any write), and every entry/module path is re-checked for containment before writing. Unit tests in `ipc.test.ts`.

packages/agency-lang/lib/runtime/ipc.ts
egonSchiele: Fixed in 8c766f5. A multi-file `CompiledProgram` now carries `entryPath` ("sub/main.js"), and `materializeCompiledScript` writes the entry there, so its `./helper.js` import resolves. `cleanupTempDir` derives the per-run dir as the first segment under `.agency-tmp` rather than `dirname(script)`. Execution test `runFileWithNestedEntry` (`runFile(dir, "sub/main.agency")` importing a sibling) plus unit tests for the layout and for escaping paths.

packages/agency-lang/lib/compiler/closureValidator.ts
egonSchiele: Fixed in 8c766f5: `readRegularFile` opens with `O_RDONLY|O_NONBLOCK|O_NOFOLLOW`, `fstat`s the descriptor and requires `isFile()`, then reads from that same descriptor. Regression test creates a real FIFO with `mkfifo` and imports it; with no writer, a blocking open or read would hang the test, so it passing is the evidence.

packages/agency-lang/lib/compiler/closureValidator.ts
```diff
@@ -0,0 +1,312 @@
...
+    }
+    this.walkFile(canonical, label);
+  }
+
+  walkFile(filePath: string, importedFrom: string): void {
+    if (this.visited[filePath]) return;
+    this.visited[filePath] = true;
+    let source: string;
+    try {
+      source = readRegularFile(filePath);
```
egonSchiele: The descriptor read fixes final-component swaps/FIFOs, but `O_NOFOLLOW` does not protect ancestor components. Between `realpathSync(resolved)` and this open, `dir/sub` can be replaced with a symlink to an outside directory; the open follows that ancestor and records outside bytes under the previously validated inside path. That still violates the validator's symlink-free confinement invariant. Please make path traversal descriptor-relative with no-follow checks for every component (or use an authoritative descriptor-target verification/native seam and fail closed where unavailable), and add a deterministic ancestor-swap test between containment validation and open. If concurrent mutation is intentionally out of scope, the API/security contract needs to state and enforce exclusive ownership of `dir` instead.

egonSchiele (commented):
The prior findings are fixed. One containment race remains in the validator.

PR #879, opened 2026-08-22
std::agency test() and testFile() (3/4)

egonSchiele:
Layer 3 of 4, split out of #876. Base: `adit/split-2-sandboxed-compile` (layer 2); retarget to `main` once that merges.

## What this does

Adds `test()` and `testFile()` to `std::agency` (`stdlib/agency.agency`), with TS seams in `lib/stdlib/agency.ts`.

- **`test(dir, filename, cases, ...)`** runs each case in the run() subprocess, so every interrupt the tested code raises, and the per-case `std::run` launch itself, is voted on by the caller's handlers. Any reject wins: a parent that rejects everything vetoes every effect, including the tested code's own inline `with approve`. A case's scripted `interrupts` answers are one vote each, consumed in order; exhausted answers stay silent (the interrupt propagates outward), leftover answers fail the case, an `expectedMessage` mismatch fails the case. A compile failure is a whole-call failure; a case failing is a `pass: false` entry in a success report. `maxCost` caps the whole call in run()'s `limit_exceeded` shape.
- **`testFile(dir, filename, maxCost = null)`** runs the same `.test.json` file `agency test` runs, restricted to the sandbox profile from layer 1 (named `args`, exact criteria, approve/reject answers). It gates the read with a `std::read` interrupt before the TS seam performs it and converts the whole file up front, so a malformed file is a whole-call failure with zero launches. An unknown node is a per-case launch failure like any other.

## Surfaced for review

In-process approval merge: a valueless outer `approve()` overwrites an inner approval's value (`DEFAULT_MERGE` in `lib/runtime/effectMerge.ts`); only the IPC merge defers. Scripted `approve(value)` answers therefore need a silent, not blanket-approving, parent. Documented in `docs/dev/std-agency-test.md`. Worth deciding whether the in-process default should defer like the IPC one.

## Revision history

The first version parsed the CLI `input` string per case, read the tested source a second time to build a node-parameter table, and bound every case before launching (`_readNodeBindingTable`, `_bindTestFileCases`, a second read gate, `_testFileForGrading`). All gone with the switch to `args`.

## Verification

`lib/stdlib/agency.test.ts` + `agencyTestFile.test.ts`: 63 passed. `tests/agency/agency-test-fn` (18 cases) and `agency-testfile` (6 cases): all green. Guards pass on the full stack.


packages/agency-lang/stdlib/agency.agency
```diff
@@ -261,6 +264,227 @@ export def runFile(
...
+        args: c.args,
+        expected: c.expected,
+        interrupts: c.interrupts,
+        wallClock: c.wallClock
+      },
+      )
+    }
+    return test(
+      dir,
+      wire.sourceFile,
```
egonSchiele, line 480: `sourceFile` is defined as relative to the `.test.json`, but this resolves it against `dir`. For `testFile(dir, "sub/suite.test.json")`, either the default source or `"sourceFile": "harness.agency"` should select `sub/harness.agency`; this instead compiles `dir/harness.agency`. Please preserve the JSON file's relative directory when producing/passing the source path, and add a nested-suite regression test.

packages/agency-lang/stdlib/agency.agency
```diff
@@ -261,6 +264,227 @@ export def runFile(
...
+        if (intr.effect == "std::run") {
+          if (intr.data.moduleId == program.moduleId) {
+            return approve()
+          }
+        }
+        if (scripted.index < answers.length) {
+          const answer = answers[scripted.index]
+          scripted.index = scripted.index + 1
+          const expectedMessage = answer.expectedMessage ?? ""
+          if (expectedMessage != "") {
```
egonSchiele: An explicitly declared `expectedMessage: ""` is treated as if no expectation was supplied, even though the schema accepts empty strings and the CLI checks presence with `!== undefined`. A tested `interrupt("nonempty")` would therefore consume this answer without reporting the mismatch. Please preserve presence separately from the string value (or reject empty expectations in the shared schema) and cover the empty-string case.

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -582,3 +584,66 @@ export function _filterImports(
...
+  cases: TestFileCaseWire[];
+};
+
+/** Reads and validates a sandbox-profile .test.json. Every case is
+ *  converted up front, so a bad file is a whole-call failure before
+ *  test() can launch anything. An escaping filename is refused by
+ *  resolveInSandbox BEFORE any read. */
+export function _readTestFileSandbox(dir: string, filename: string): ParsedTestFileWire {
+  const target = resolveInSandbox(dir, filename);
+  const parsed = parseTestFileSandbox(readFileSync(target, "utf-8"), filename);
```
egonSchiele: This reopens `target` after `resolveInSandbox` validated it, so a concurrent swap can redirect the gated read outside `dir`: replace the validated file (or an ancestor directory) with a symlink between resolution and `readFileSync`, and the host follows it. The `std::read` vote authorized the original sandbox-relative path, not arbitrary outside bytes. Please read through a descriptor whose target/containment is validated without a check/read gap (and protect ancestor components, as in the closure-validator concern), with a deterministic swap-seam test.

egonSchiele (commented):
Three correctness and sandbox-boundary issues found.

packages/agency-lang/stdlib/agency.agency
egonSchiele: Fixed in 7e03a5b79: `_readTestFileSandbox` returns `join(dirname(filename), sourceFile)`, so `testFile(dir, "sub/suite.test.json")` tests `sub/suite.agency` (or `sub/<declared>`). Unit test in `agencyTestFile.test.ts` for both the default and the explicit form; execution regression `nestedSuiteUsesItsOwnDirectory` in `tests/agency/agency-testfile.agency`.

packages/agency-lang/stdlib/agency.agency
egonSchiele: Fixed both ways. In the shared schema (commit 10502d60f on #877) an empty `expectedMessage` is refused in both profiles, since an empty expectation can never match a real interrupt. In `test()` itself (7e03a5b79) the check is now by presence (`!= null`, which compiles to the nullish `__eq`), so a direct caller passing `expectedMessage: ""` gets a compared-and-failed case rather than a skipped check. Execution regression `emptyExpectedMessageIsChecked`: the tested node interrupts with "gimme", the answer expects "", the case fails with feedback naming "gimme".

packages/agency-lang/lib/stdlib/agency.ts
egonSchiele: Fixed in 2a2893fd7 (#878) + 7e03a5b79. New `lib/utils/readContainedFile.ts`: open `O_RDONLY|O_NONBLOCK|O_NOFOLLOW`, `fstat` the descriptor and require `isFile()`, realpath the path AFTER the open and require strict containment in the root (catches an ancestor swapped to a link that is still in place), then compare `(dev, ino)` of that path with the descriptor (catches a swap undone again after the open), then read from the descriptor. Both the closure validator and `_readTestFileSandbox` read through it. Deterministic swap-seam tests use an `afterOpen` seam to perform each swap between the open and the validation: ancestor-to-outside-link is refused with /outside/, swap-and-restore is refused with /changed between validation and read/; plus final-component symlink and FIFO cases.

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -582,3 +585,71 @@ export function _filterImports(
...
+/** Reads and validates a sandbox-profile .test.json. Every case is
+ *  converted up front, so a bad file is a whole-call failure before
+ *  test() can launch anything. An escaping filename is refused by
+ *  resolveInSandbox BEFORE any read. */
+export function _readTestFileSandbox(dir: string, filename: string): ParsedTestFileWire {
+  const target = resolveInSandbox(dir, filename);
+  // The std::read vote authorized this sandbox-relative path; the read
+  // validates the opened descriptor itself, so a swap after
+  // resolveInSandbox cannot redirect it outside dir.
+  const parsed = parseTestFileSandbox(readContainedFile(dir, target), filename);
```
egonSchiele, line 636: `readContainedFile` still does not establish the claimed no-gap containment. A valid POSIX interleaving is: open the outside file through a temporarily symlinked ancestor; restore the inside ancestor before `realpathSync` (so the returned pathname passes containment); then restore the outside symlink after `realpathSync` but before `statSync(real)`. That stat reaches the same outside inode held by the fd, so the identity check passes and outside bytes are read. The current seams cover outside→outside and inside→inside, not this outside→inside→outside ABA sequence. Pathname revalidation cannot close this under concurrent mutation; use descriptor-relative component traversal (`openat2`/equivalent, failing closed where unsupported), or explicitly remove concurrent mutation from the threat model and enforce exclusive ownership of `dir`.

packages/agency-lang/stdlib/agency.agency
```diff
@@ -261,6 +264,228 @@ export def runFile(
...
+          if (intr.data.moduleId == program.moduleId) {
+            return approve()
+          }
+        }
+        if (scripted.index < answers.length) {
+          const answer = answers[scripted.index]
+          scripted.index = scripted.index + 1
+          // Presence decides whether the message is checked (== null also
+          // matches undefined); an empty string is a real expectation.
+          if (answer.expectedMessage != null) {
```
egonSchiele, line 382: This direct-API fix correctly treats `""` as a real expectation, but the stacked shared schema now rejects that same value, so `testFile()` cannot express it and the full CLI profile loses previously valid behavior. Empty interrupt messages are valid (`interrupt("")`), so an empty expected message can match a real interrupt. Please remove the schema's `.min(1)` and keep this presence-based comparison in both runners; add a `testFile` case where both expected and actual messages are empty.

egonSchiele (commented):
The nested-suite fix is correct. Two actionable issues remain in the follow-up.

PR #882, opened 2026-08-22
Eval grading on agency test --agency-only --reject '*' (replaces #880)

egonSchiele:
Replaces #880's grader design. Spec: `packages/agency-lang/docs/superpowers/specs/2026-08-22-eval-agency-test-grading-design.md`; plan (v2) and its review beside it under `docs/superpowers/plans/`. Dev docs: `docs/dev/test-cli-sandbox.md` (new), `docs/dev/eval-grading.md`, `docs/dev/run-directory.md`.

## What this does

An eval coding test ships harness pairs (`files/<name>.test.json` + `.agency`, seeded so the agent self-checks with `agency test`; `holdout/…`, never seeded). The framework grades the agent's workdir by running the very same test runner the agent used, locked down by two flags:

```
agency test run --json --agency-only --reject '*' --max-cost 5 <name>.test.json
```

- **`--agency-only`** (on `agency test run` and `agency run`): the tested file compiles through the #878 closure validator — no TypeScript/JavaScript, Node built-ins, `pkg::`, splices, absolute imports or symlinks in its closure — so every effect the code can perform is an interrupt. In the test runner a refusal is that file's failure (the suite goes on, exit 1 at the end), not a process exit.
- **`--policy` / `--approve` / `--reject`, `--max-cost` / `--max-time`** on `agency test run`: the same root handler and budget `agency run` already installs, passed to each case's child with the same clear-then-set rule (`lib/cli/testChildEnv.ts`), so nothing leaks in from an outer shell. `--reject '*'` rejects every effect; the chain resolves reject over approve, so the tested code's own `with approve` loses.
- **`--json`**: exactly one document on stdout (`lib/cli/testReport.ts`, version 1), every human line on stderr. `runSingleTest` now returns a structured outcome with the failure text, so the document carries per-case feedback (the last attempt's, with `attempts`).

The run directory keeps both files of every pair under `graders/` by content hash, with a `harness` record on the run row; `eval grade` rebuilds one `AgencyTestGrader` per record (revision `agency-tests/<name>@<sha256>`), under `--graders` and `--goal` alike. Score = passing fraction, `mustPass` + `threshold: 1`. Preflight (`parseTestFileEvalHarness`) refuses, before any agent runs, the fields that cannot mean anything under the grading policy. `harnessMaxCost` in `test.json` overrides the $5-per-case default.

The fib eval moves to harness pairs; its `graders.ts` and the hand-written sandbox harness are gone, and so is the combined-grader spec from #877 (the problem it described no longer exists).

## Two things to know

- Flags go after `agency test run`; `agency test --reject …` is refused by the parser, as `--shard` already was. `docs/dev/cli-arguments.md` says the default subcommand answers with its child's flags; it does not, here. Not touched in this PR.
- The `agency test` reference page under `docs/site/cli/` needs the five new flags; left for the owner.

## Verification

`make`; per-task unit suites (`lib/cli/*`, `lib/testFormat`, `lib/eval/grading/*` incl. the real-CLI spawn suite, `loadInputs`, `runSuite`, `annotations`, `ipc`); three new agency-js tests, each with a positive control (`test-cli-policy`, `test-cli-agency-only`, `test-cli-json` incl. a sharded run); prettier, `lint:structure`, `sourceIsText`, no binary files. No model was called; the fib eval end to end is covered by the grader spawn suite (correct solution → 1, wrong → 0.5 with the diff, `import fs` → 0 with the refusal, self-approved write → nothing written).


packages/agency-lang/lib/testFormat/schema.ts
```diff
@@ -292,6 +292,72 @@ export function parseTestFileFull(jsonText: string, jsonFilename: string): FullT
...
+  };
+  for (const field of EVAL_HARNESS_REFUSED_FILE_FIELDS) {
+    if (file[field] !== undefined) refuse(`${field} is not allowed in an eval harness`);
+  }
+  if (file.sourceFile !== undefined && file.sourceFile !== siblingAgencyBasename) {
+    refuse(
+      `declares sourceFile ${JSON.stringify(file.sourceFile)}, but an eval harness tests its sibling ${siblingAgencyBasename}`,
+    );
+  }
+  (file.tests ?? []).forEach((testCase, i) => {
```
egonSchiele: **P1 — Refuse empty eval harnesses.** `file.tests ?? []` allows both a missing `tests` field and `tests: []` through preflight. `gradeReport` deliberately scores a `ran` file with zero cases as 1, so an accidentally empty harness gives every submission a perfect gate without testing anything. Eval harnesses should require a present, non-empty `tests` array here, with regression cases for both missing and empty arrays.

packages/agency-lang/lib/cli/test.ts
```diff
@@ -898,101 +982,63 @@ async function runTestFile(
...
+    if (options.agencyOnly) {
+      const compiled = compileAgencyOnly(sourceFile);
+      if (!compiled.ok) {
+        log(color.red(`  ✗ ${testFile}: agency-only compile refused`));
+        for (const error of compiled.errors) log(color.red(`    ${error}`));
+        suite.completed.push(testFile);
+        return {
+          stats: {
+            passed: 0,
+            failed: total,
```
egonSchiele: **P1 — Make file failures affect the process status.** This sets `filesFailed: 1` but `failed: total`. For a valid test file with `tests: []`, an `--agency-only` refusal therefore leaves `failed` at zero, and the command's final `if (totals.failed > 0)` exits successfully even though the report says `compile-failed`. Check `filesFailed` in the exit decision as well, or make every file-level failure contribute at least one failed test.

packages/agency-lang/scripts/agency.ts
```diff
@@ -1203,16 +1268,17 @@ export function createProgram(deps: CliDependencies = {}): Command {
...
           }
           const colorFn = totals.failed > 0 ? color.red : color.green;
-          console.log(colorFn(`\n Test Files  ${filesStatus} (${totalFiles})`));
-          console.log(colorFn(`      Tests  ${testsStatus} (${totalTests})`));
+          output.line(colorFn(`\n Test Files  ${filesStatus} (${totalFiles})`));
+          output.line(colorFn(`      Tests  ${testsStatus} (${totalTests})`));
         }
-        printSlowestTests(totals.slowTests);
+        printSlowestTests(totals.slowTests, output);
+        output.document(totals.report);
```
egonSchiele: **P1 — Keep coverage output off stdout in JSON mode.** With `--json --coverage`, this writes the JSON document and then `generateReport()` prints its coverage summary with `console.log`, leaving stdout as JSON followed by human text. That breaks the advertised “exactly one document” contract. Route coverage reporting through the selected output sink, send it to stderr in JSON mode, or reject this flag combination; an integration assertion for the combination would pin the contract.

packages/agency-lang/lib/cli/test.ts
```diff
@@ -1068,10 +1249,14 @@ export async function test(
   // exit here, matching the old per-case compile behavior (those compiles
   // ran in this process and exited too).
   try {
-    precompileTestSources(config, testFiles);
+    // --agency-only compiles each file itself, through the validator.
+    // Under --json the pass's progress lines would land on stdout: quiet.
+    if (!options.agencyOnly) {
+      precompileTestSources(config, testFiles, { quiet: options.output !== undefined });
```
egonSchiele: **P1 — Preserve the JSON contract for ordinary compile failures.** The new report schema includes `compile-failed`, but this non-`--agency-only` precompile path still exits before `output.document()` runs: closure errors hit the catch below, and parse/typecheck failures inside `BuildSession` can call `process.exit(1)` directly. As a result, `agency test run --json broken.test.json` can produce no JSON at all. The JSON path needs non-exiting per-file compile failure accounting (or the unsupported combination must be rejected/documented) so completed invocations always yield the promised document.

packages/agency-lang/lib/cli/test.ts
```diff
@@ -898,101 +982,63 @@ async function runTestFile(
...
+          filesFailed: 1,
+          failedFiles: [testFile],
+          slowTests: [],
+        },
+        report: fileFailureReport({
+          file: testFile,
+          sourceFile,
+          status: "error",
+          error,
+          caseNodes: [],
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c159aeb64: the document now carries `filesFailed` (files that did not run plus files with a failed case), and the command exits 1 on `filesFailed` as well as `failed`. A malformed file or a runner error shows `filesFailed: 1` even with zero cases, so the totals agree with the exit status. Covered by a new `testReport.test.ts` case.

packages/agency-lang/lib/testFormat/schema.ts
```diff
@@ -292,6 +292,72 @@ export function parseTestFileFull(jsonText: string, jsonFilename: string): FullT
...
+        );
+      }
+    }
+    const criteria = testCase.evaluationCriteria;
+    if (criteria.length !== 1 || criteria[0].type !== "exact") {
+      refuse(
+        `${where}: evaluationCriteria must be exactly [{ "type": "exact" }] in an eval harness`,
+      );
+    }
+  });
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c159aeb64 (same fix as the P1 thread below): a missing or empty `tests` array is refused with "an eval harness must have a non-empty tests array".

packages/agency-lang/docs/superpowers/specs/2026-08-22-eval-agency-test-grading-design.md
```diff
@@ -0,0 +1,422 @@
+# Eval grading of Agency coding tests: `agency test --agency-only --reject '*'`
+
+Date: 2026-08-22. Status: designed, awaiting owner review. Supersedes the
+"Eval AgencyTestGrader" half of
+`/Users/adityabhargava/agency-lang/packages/agency-lang/docs/superpowers/specs/2026-08-20-std-agency-test-design.md`
+and all of
+`/Users/adityabhargava/agency-lang/packages/agency-lang/docs/superpowers/specs/2026-08-21-combined-grader-external-files-design.md`
+(the synthesized grading module, which this design removes).
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c159aeb64: repo-relative now.

packages/agency-lang/docs/superpowers/plans/2026-08-22-eval-agency-test-grading.md
```diff
@@ -0,0 +1,186 @@
...
+Task 3 now changes the per-case outcome contract and routes every line of
+human output through one sink; Task 1 clears and sets the budget keys the
+way `agency run` does, as strings; the `revision` removal lives in Task 9
+only; Task 6 names its combined snapshot type and uses the existing
+`<sha256><ext>` file naming; Task 8 has no live-grading placeholder.
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:executing-plans to implement this plan task-by-task (this project does NOT use subagent-driven development). Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Spec:** `/Users/adityabhargava/agency-lang/packages/agency-lang/docs/superpowers/specs/2026-08-22-eval-agency-test-grading-design.md`. Read it first; every task names its section.
+
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c159aeb64: repo-relative now.

packages/agency-lang/docs/superpowers/plans/2026-08-22-eval-agency-test-grading.md
```diff
@@ -0,0 +1,186 @@
...
+- [ ] `runTestFile`: under `options.agencyOnly`, call `compileAgencyOnly` once per file before the cases loop. On `ok: false`, log the errors, mark every case failed with that text as feedback, record the file in `failedFiles`, and return; do not `process.exit`. On success pass `scriptPath` as `compiledPath` to every case. Skip the precompile pass for these files (`test()` already groups files for precompile; exclude them when `agencyOnly` is set).
+- [ ] Cleanup: the materialized directory is deleted after the file's cases finish, through `safeDeleteDirectoryWithin(<.agency-tmp>, <dir>)`. Never derive the directory by path arithmetic; `materializeCompiledScript` returns the script path and the per-run dir is its parent only for single-file programs, so have `compileAgencyOnly` also return the run directory it created (the first segment under `.agency-tmp`, the same rule `cleanupTempDir` in `ipc.ts` uses; expose that rule as a named helper rather than copying it).
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c159aeb64: the plan now names `removeCompiledScriptDir`.

packages/agency-lang/lib/testFormat/schema.ts
egonSchiele: Fixed in c159aeb64. `parseTestFileEvalHarness` refuses a missing or empty `tests` array ("an eval harness must have a non-empty tests array"), with regression cases for both in `schema.test.ts`. Since `snapshotHarness` preflights every pair through this parser, an empty harness is refused at `eval run` time, before any agent runs.

packages/agency-lang/lib/cli/test.ts
egonSchiele: Fixed in c159aeb64. The exit decision is now `totals.failed > 0 || totals.filesFailed > 0`, and the JSON document carries a matching `filesFailed` count (files with status `compile-failed`/`error`, plus files with any failed case). The `test-cli-json` agency-js test pins a refused file under `--agency-only`: exit 1, `filesFailed: 1`.

packages/agency-lang/scripts/agency.ts
egonSchiele: Fixed in c159aeb64 by rejecting the combination: `--json --coverage` exits 2 with "cannot be combined with --coverage unless --collect-only is also given" before anything runs. `--json --coverage --collect-only` stays allowed (it writes data and prints no report). Pinned in `test-cli-json`: exit 2, empty stdout, the message on stderr. Routing `generateReport` through the output sink would have meant threading it through every `console.log` in `coverage.ts`; the rejection is smaller and the report is still one command away.

packages/agency-lang/lib/cli/test.ts
egonSchiele: Addressed in c159aeb64 by documenting and pinning the edge rather than changing the compile pipeline, the alternative you allowed. The shared precompile pass goes through `BuildSession`, which calls `process.exit(1)` in eight places (parse, splice, typecheck, closure, dependency prep); there is no non-exiting mode, and adding one means changing every compile command's failure path. That is a separate change. What holds now, stated in `docs/dev/test-cli-sandbox.md` and the spec: under `--json`, an invocation that reaches the run phase always prints the document; a compile failure in the shared precompile pass ends the command with exit 1, diagnostics on stderr and an empty stdout, and a consumer must treat a missing document as a failure (the eval grader does, and it always passes `--agency-only`, where every file compiles on its own and a failure is a `compile-failed` entry). The `test-cli-json` test pins both sides with the same broken file: without `--agency-only` exit 1 and no document; with it, one document, `compile-failed`, `filesFailed: 1`. Happy to do the non-exiting `BuildSession` as a follow-up PR if you want the document guaranteed unconditionally.

packages/agency-lang/lib/cli/test.ts
```diff
@@ -1123,13 +1318,16 @@ export async function test(
   }
 
   let stats = emptyStats();
+  const fileReports: TestFileReport[] = [];
   for (const result of results) {
-    if (result) stats = mergeStats(stats, result);
+    if (result === undefined) continue;
```
egonSchiele: **P1 — Preserve aborted files and cases in the report instead of dropping result holes.** `runWithConcurrency` intentionally leaves `undefined` slots when a suite abort stops workers from pulling more files, but this aggregation skips them. `runCases` also breaks without appending `aborted` entries for the remaining cases. A timeout/SIGINT can therefore omit pending files and unrun cases from the “every file and case” JSON document; if the abort lands between files after earlier files passed, the omitted work also contributes no stats and the command can exit 0. Build an `aborted` `FileResult` for every hole (and remaining case), count an aborted file in `filesFailed`, and add a multi-file abort regression test.

packages/agency-lang/docs/superpowers/specs/2026-08-22-eval-agency-test-grading-design.md
```diff
@@ -0,0 +1,429 @@
...
+      "sourceFile": "broken.agency",
+      "status": "compile-failed",
+      "error": "Sandboxed compilation refused:\n  - broken.agency imports 'fs', which is not Agency source ...",
+      "cases": []
+    }
+  ],
+  "passed": 1,
+  "failed": 1,
+  "skipped": 0,
+  "filesFailed": 0
```
egonSchiele: **P2 — Make the example agree with `buildTestReport`.** This document contains one `ran` file with a failed case and one `compile-failed` file. The implementation counts both as failed files, so `filesFailed` is 2, not 0. Leaving 0 in the public contract example will lead consumers to implement the wrong meaning.

packages/agency-lang/lib/testFormat/schema.ts
egonSchiele: Verified at c159aeb64: preflight now refuses both missing and empty test arrays, and both cases are covered. This finding is resolved.

packages/agency-lang/lib/cli/test.ts
egonSchiele: Verified at c159aeb64: the command now exits on either failed cases or failed files, and filesFailed is included in the report. This specific finding is resolved.

packages/agency-lang/lib/cli/test.ts
egonSchiele: Verified at c159aeb64: the narrower contract is now explicit and tested—shared-precompile failures exit 1 without a document, while --agency-only reports per-file compile failures. That documented boundary is acceptable for this PR; this finding is resolved.

packages/agency-lang/scripts/agency.ts
egonSchiele: Verified at c159aeb64: --json rejects coverage reporting while retaining --coverage --collect-only, with an integration test. This finding is resolved.

packages/agency-lang/evals/agency-agent/fib/files/fib-tests.agency
```diff
@@ -0,0 +1,21 @@
+// Visible tests for the fib coding eval: plain import-and-assert nodes.
+// The eval framework runs this pair through std::agency's testFile() inside
+// the reject-all sandbox wrapper; the agent self-checks with
+// `agency test fib-tests.test.json`.
```
egonSchiele: please update the agent code to give the `testFile` func as a tool to the coding agent so it can check its solution

packages/agency-lang/lib/cli/agencyOnlyCompile.ts
```diff
@@ -0,0 +1,37 @@
+/**
```
egonSchiele: Long comments actually hurt readability because the user is unlikely to read them. Can the information in this comment be expressed in the code instead? Or can it be expressed as a comment that is more concise and more readable? Same goes for all of the other comments you have added in this PR. Every time you add a long comment, just know that you are hurting instead of helping readability.

packages/agency-lang/lib/cli/agencyOnlyCompile.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * The validator refuses anything in the import closure that is not Agency
+ * source: TypeScript/JavaScript files, Node built-ins, `pkg::` packages,
+ * compile-time splices, and local imports that are absolute, leave the
+ * source's directory, or go through a symlink. A refusal comes back as data
+ * (`ok: false`) because for the callers here — the test runner grading
+ * agent-written code, `agency run` on an untrusted file — "it does not
+ * compile" is an ordinary outcome, not a reason to kill the process.
+ */
+import * as path from "path";
```
egonSchiele: This file seems borderline-useless to me. What do you think it is useful for?

packages/agency-lang/lib/cli/commands.ts
```diff
@@ -335,6 +352,7 @@ export function run(
   });
 
   nodeProcess.on("exit", (code) => {
+    if (compileMode.agencyOnly) removeCompiledScriptDir(output);
```
egonSchiele: Why does "agency only" mean an extra directory gets created that then needs to be deleted?

packages/agency-lang/lib/cli/test.ts
```diff
@@ -604,6 +648,11 @@ async function runSingleTest(
       } else {
         log(color.red("  ✗ Exact match failed"));
         log(verdict.feedback);
+        // The report carries the same diff without terminal colors.
+        const plain = exactVerdict(result.data, testCase.expectedOutput, {
+          rawStringFallback: true,
+        });
+        feedback.push(plain.pass ? "Exact match failed" : plain.feedback);
```
egonSchiele: is this ternary flipped?

packages/agency-lang/lib/cli/testChildEnv.ts
```diff
@@ -0,0 +1,51 @@
+/**
```
egonSchiele: The way we use environment variables in this codebase seems like a total mess. We set them in some places, we clear them in some places. I feel like there isn't a single source of truth where I can understand how these environment variables are being used, where they're being set, and where they're being cleared. Does this logic make sense to you? because it does not make sense to me and feels like it is sprawled all over the codebase.

packages/agency-lang/lib/eval/grading/agencyTestGrader.ts
```diff
@@ -0,0 +1,194 @@
...
+  protected readonly defaultName: string;
+  private readonly opts: AgencyTestGraderOptions;
+  private readonly runHarness: RunHarness;
+
+  constructor(opts: AgencyTestGraderOptions, runHarness: RunHarness = spawnHarness) {
+    super({ name: opts.name, mustPass: true, threshold: 1 });
+    this.defaultName = opts.name;
+    this.opts = opts;
+    this.runHarness = runHarness;
+    this.revision = `agency-tests/${opts.name}@${harnessSha256(
```
egonSchiele, line 85: what is this for? There is no requirement to support any sort of backwards compatibility here. What is harnessAgency? What is harnessJson?

packages/agency-lang/lib/cli/test.ts
egonSchiele: Fixed in 7040c7d8e. The runner now returns only the report; counts and the failed-file list derive from it, so there is no second structure to fall out of sync. A file the abort kept from starting is an `aborted` entry with no cases; unrun cases in the interrupted file are `aborted`; an aborted file counts in `filesFailed` and the command exits 1. Pinned in `test-cli-json`: two files, `-p 1`, SIGINT once the first is spinning → one document, both files `aborted`, `filesFailed: 2`, exit 1.

packages/agency-lang/docs/superpowers/specs/2026-08-22-eval-agency-test-grading-design.md
egonSchiele: Fixed in 7040c7d8e: `filesFailed: 2`.

packages/agency-lang/lib/eval/grading/gradeRun.ts
```diff
@@ -151,6 +165,27 @@ async function effectiveGraders(
...
+        );
+      }
+    }
+    return new AgencyTestGrader({
+      name: record.name,
+      harnessAgency: path.join(gradersDir, record.agency),
+      harnessJson: path.join(gradersDir, record.json),
+      ...(record.maxCost === undefined ? {} : { maxCost: record.maxCost }),
+    });
+  });
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 7040c7d8e, in the one place the record is parsed: the run-row schema now requires `name` to be a plain name and `agency`/`json` to match `<sha256>.agency` / `<sha256>.test.json`. `gradeRun` and the grader need no checks of their own. Negative cases in `annotations.test.ts`.

packages/agency-lang/lib/eval/grading/agencyTestGrader.ts
```diff
@@ -0,0 +1,194 @@
...
+    const scratch = makeAgencyTempDir("test-grade");
+    try {
+      copyWithoutSymlinks(run.workdir, scratch);
+      // The pair lands under its logical name whatever the framework's copy
+      // is called (a run directory stores it by content hash): the json's
+      // default sourceFile is `<name>.agency`, and that is also the name the
+      // agent saw in its workdir.
+      const jsonFilename = `${this.opts.name}.test.json`;
+      installHarnessFile(this.opts.harnessAgency, path.join(scratch, `${this.opts.name}.agency`));
+      installHarnessFile(this.opts.harnessJson, path.join(scratch, jsonFilename));
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 7040c7d8e via the schema pattern on `name` (see the gradeRun thread).

packages/agency-lang/scripts/agency.ts
```diff
@@ -1171,9 +1210,43 @@ export function createProgram(deps: CliDependencies = {}): Command {
...
+              resolveRunPolicy({
+                policy: opts.policy,
+                approve: opts.approve,
+                reject: opts.reject,
+                cwd: process.cwd(),
+              }) ?? undefined,
+            budget: resolveBudget({ maxCost: opts.maxCost, maxTime: opts.maxTime }),
+            agencyOnly: opts.agencyOnly ?? false,
+            output,
+          };
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 7040c7d8e: `TestOutput` carries its `kind`, and precompile is quiet only when it is `json`. The `output !== undefined` proxy is gone.

packages/agency-lang/evals/agency-agent/fib/files/fib-tests.agency
egonSchiele: Done in 7040c7d8e: `testFile` from `std::agency` is in the coding subagent's `codeTools`. Also fixed the stale comment here and the test description, which still talked about the old wrapper.

packages/agency-lang/lib/cli/agencyOnlyCompile.ts
egonSchiele: Agreed. In 7040c7d8e every header the PR added is one to three lines, and the design argument lives only in `docs/dev/test-cli-sandbox.md`. This file is gone (next thread).

packages/agency-lang/lib/cli/agencyOnlyCompile.ts
egonSchiele: You're right; it wrapped two calls. Deleted in 7040c7d8e. What it did now lives where the real work is: `compileAgencyOnly` in `lib/compiler/compileSandboxed.ts`, which compiles through the validator and writes the result beside the sources.

packages/agency-lang/lib/cli/commands.ts
egonSchiele: Because the sandboxed compile returns its output in memory (it compiles from a private mirror and never writes into the caller's directory), so I put it in the runtime's temp-dir helper, which then needed cleanup. That was the wrong place. In 7040c7d8e `--agency-only` writes `<name>.js` beside each source, the same layout a normal compile produces, and the cases run it through the existing `preferCompiled` path. No temp directory, no cleanup, no `compiledPath` plumbing, and `ipc.ts` is untouched.

packages/agency-lang/lib/cli/test.ts
egonSchiele: Not flipped, but it read that way. `exactVerdict` only has `feedback` on a failing verdict, and the colored and plain verdicts cannot disagree, so the ternary guarded an impossible case. In 7040c7d8e it is `if (!plain.pass) feedback.push(plain.feedback)`.

packages/agency-lang/lib/cli/testChildEnv.ts
egonSchiele: You're right, and I had added a fourth copy. The runtime reads `AGENCY_RUN_POLICY`, `AGENCY_RUN_POLICY_INTERACTIVE`, `AGENCY_MAX_COST`, `AGENCY_MAX_TIME` from the environment because the compiled program is a separate process; before this commit `run`, `agent`, and the test runner each hand-rolled delete-then-set. In 7040c7d8e there is one writer: `withRootCarriers` in `lib/cli/childEnv.ts`. It deletes all four and sets only what this invocation's flags resolved; the three spawners call it and nothing else touches those names. One behavior change to flag: `agency agent` now also clears an inherited `AGENCY_RUN_POLICY` (it only cleared the budget before). The policy/run-policy agency-js tests still pass. Documented under "Environment discipline" in `docs/dev/test-cli-sandbox.md`.

packages/agency-lang/lib/eval/grading/agencyTestGrader.ts
egonSchiele: Nothing to do with compatibility; the comment was bad. They are the absolute paths of the framework's copy of the harness pair (the `.agency` and the `.test.json`), which the grader copies into the scratch dir under `<name>.agency` / `<name>.test.json` and hashes for the revision. Renamed to `agencyFile` / `testJsonFile` in 7040c7d8e with a one-line comment saying that.

PR #886, opened 2026-08-22
The agency command line as function calls: std::agency cli(), safeBash, agencyCli

egonSchiele:
Stacked on #885.

## What

An agent in a coding eval keeps reaching for the `agency` command line (`agency test …`, `which agency`), and the binary is usually not installed where it runs. The subcommands that have a function in `std::agency` now run in-process wherever a command line can enter, and everything else gets one message naming the functions to call.

**`std::agency`**: `cli(args, cwd)` maps `test` → `testFile`, `typecheck`/`tc` → `typecheckFile`, `ast` → `parseAST`, `fmt` → `format`, `run <file>` → `runFile(…, "main")`, and returns the `{stdout, stderr, exitCode}` shape the CLI would (the test report rendered as `✓`/`✗` lines and an `n/m tests passed` summary, exit 1 on any failure). Only `agency <subcommand> <file>` is accepted: flags and other subcommands fail with `CLI_NOT_AVAILABLE`, which names the functions. `cliHandles(subcommand)` says which subcommands run here.

**`safeBash`**: a single `agency <sub> <file>` line becomes an `agency` plan that runs `cli()` with no shell, the way a lone `echo` already runs without one. `which agency`, `type agency`, `command -v agency`, a non-literal argument, or a redirect are refused with the same message. Commands that merely mention agency (`ls agency-lang`) are ordinary bash. Exact mappings only; anything the table does not cover is refused, not guessed.

**`agencyCli`** (the coordinator's code-subagent tool) runs the handled subcommands through `cli()` and falls back to `exec("agency", …)` for the rest (`help`, `doc`, …), rewriting an ENOENT into the same message.

Not changed: the `exec` tool itself. `exec("agency", ["test", …])` still goes to the policy, which has no rule for it and refuses with the generic text; the prompt points the agent at `agencyCli`/`testFile` instead.

## Verification

`make` clean. `tests/agency/agency-cli` (5 cases: each mapping, the three refusal forms), `tests/agency/safeBash` (22, incl. the new plan case), `tests/agency/agency-testfile` (7), and the agent's `execPolicy` test (5) all pass; the three edited `.agency` files typecheck clean.


egonSchiele (commented):
The in-process path preserves approval: `cli()` still raises the underlying `std::read`, `std::run`, and guard effects. I did find two `safeBash` boundary issues on a further adversarial pass: `tryAgencyPlan` refuses an ordinary invocation such as `command echo agency` because it treats any literal `agency` argument as a lookup, and `runAgency` returns `cli()` failures without the truncation applied to every other output path. The existing inline comments identify both locations; adding those cases to the safeBash tests should close the gaps.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -488,6 +489,47 @@ def trySeeIfShellFree(command: SimpleCommand, cwd: string): Result<Plan> {
...
+  }
+  if (name.text == "which" || name.text == "type" || name.text == "command") {
+    const lookedUp = literalArgs(command)
+    if (lookedUp is success(names)) {
+      if (names.includes("agency")) {
+        return success(refusePlan(CLI_NOT_AVAILABLE))
+      }
+    }
+    return failure("not about agency")
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 3a4a8b51f. `command` is special-cased: it runs or (with -v/-V) looks up its first argument, so only `command agency …` and `command -v agency` are refused, and `command echo agency` is ordinary bash. `which`/`type` still refuse any `agency` argument since every argument to them is a lookup. Cases added to `agencyCommandLines`: `command agency test x.test.json` → refuse, `command echo agency` → bash, `which git agency` → refuse.

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -867,13 +913,32 @@ export def safeBash(
...
 
+def runAgency(args: string[], cwd: string): Result<string> {
+  """
+  What the CLI would have printed. A non-zero exit is a failure carrying
+  the output, the way a failing bash command is.
+  """
+  const ran = cli(args, cwd)
+  if (ran is failure(err)) {
+    return failure("${err}")
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 3a4a8b51f: `runAgency` returns `failure(truncate(…))` for a `cli()` failure, the same cap as the success and non-zero-exit paths.

egonSchiele:
Both boundary findings addressed in 3a4a8b51f (replies on the threads); the base branch's review fixes are merged in. The safeBash suite is 22/22 with the new cases.

packages/agency-lang/stdlib/agency.agency
```diff
@@ -1037,3 +1037,143 @@ export idempotent def getVersion(): string {
...
+
+export def cli(args: string[], cwd: string = "."): Result<ExecResult> {
+  """
+  Run an `agency` command line through the matching function: `test` runs testFile, `typecheck`/`tc` runs typecheckFile, `ast` runs parseAST, `fmt` runs format, `run` runs runFile on node main. Only the form `agency <subcommand> <file>` is accepted; flags and other subcommands fail with a message naming the function to call instead.
+
+  @param args - The CLI arguments, subcommand first, e.g. ["test", "x.test.json"]
+  @param cwd - The directory the file argument is inside
+  """
+  if (args.length == 0 || !cliHandles(args[0])) {
+    return failure(CLI_NOT_AVAILABLE)
```
egonSchiele: if they call this function incorrectly, why return a string telling them that the CLI is not available? Clearly they know it is not available -- that's why they are using this function.

packages/agency-lang/stdlib/agency.agency
```diff
@@ -1037,3 +1037,143 @@ export idempotent def getVersion(): string {
...
+  }
+  const sub = args[0]
+  if (args.length != 2 || args[1].startsWith("-")) {
+    return failure("Only `agency ${sub} <file>` runs here. ${CLI_NOT_AVAILABLE}")
+  }
+  const file = args[1]
+  if (sub == "test") {
+    const report = testFile(cwd, file)
+    if (report is failure(err)) {
+      return failure("${err}")
```
egonSchiele: why not just `return report`?

packages/agency-lang/stdlib/agency.agency
```diff
@@ -1037,3 +1037,143 @@ export idempotent def getVersion(): string {
...
+  """
+  if (args.length == 0 || !cliHandles(args[0])) {
+    return failure(CLI_NOT_AVAILABLE)
+  }
+  const sub = args[0]
+  if (args.length != 2 || args[1].startsWith("-")) {
+    return failure("Only `agency ${sub} <file>` runs here. ${CLI_NOT_AVAILABLE}")
+  }
+  const file = args[1]
+  if (sub == "test") {
```
egonSchiele: pattern matching would be a better fit here https://agency-lang.com/guide/pattern-matching.html

packages/agency-lang/stdlib/agency.agency
```diff
@@ -1037,3 +1037,143 @@ export idempotent def getVersion(): string {
...
+    }
+    if (ran is success(value)) {
+      return success({
+        stdout: JSON.stringify(value),
+        stderr: "",
+        exitCode: 0
+      })
+    }
+  }
+  const source = read(file, cwd)
```
egonSchiele, line 1112: why read the source before knowing what you'll use it for?

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -488,6 +489,53 @@ def trySeeIfShellFree(command: SimpleCommand, cwd: string): Result<Plan> {
...
+  """
+  const name = command.command
+  if (name == null || name.tag != "literal" || command.assignments.length > 0) {
+    return failure("not an agency command")
+  }
+  // `which`/`type` look names up; `command` runs or (with -v/-V) looks
+  // up its FIRST argument, so `command echo agency` is an ordinary echo.
+  if (name.text == "which" || name.text == "type" || name.text == "command") {
+    const lookedUp = literalArgs(command)
+    if (lookedUp is success(names)) {
```
egonSchiele: sure feels like pattern matching would make this more readable

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -488,6 +489,53 @@ def trySeeIfShellFree(command: SimpleCommand, cwd: string): Result<Plan> {
   return shellFreePlan(command, cwd)
 }
 
+def tryAgencyPlan(command: SimpleCommand): Result<Plan> {
```
egonSchiele, line 492: why is this called "plan"?

packages/agency-lang/stdlib/safeBash.agency
```diff
@@ -867,13 +919,32 @@ export def safeBash(
...
+  if (ran is failure(err)) {
+    return failure(truncate("${err}"))
+  }
+  if (ran is success(result)) {
+    if (result.exitCode != 0) {
+      return failure(truncate(result.stdout + result.stderr))
+    }
+    return success(truncate(result.stdout))
+  }
+  return failure(CLI_NOT_AVAILABLE)
```
egonSchiele: Again, they're not calling the CLI. So why tell them that the CLI is not available?

packages/agency-lang/stdlib/agency.agency
egonSchiele: Agreed. In 50f64788c cli() failures describe the call: a flag ("takes a file, not a flag"), an unsupported subcommand ("no function runs agency help; the subcommands that run here are …"), the wrong number of arguments, or none. CLI_NOT_AVAILABLE is now used only by the interceptors, safeBash refusals and agencyCli on ENOENT, where someone really was reaching for the CLI.

packages/agency-lang/stdlib/agency.agency
egonSchiele: Done in 50f64788c, as a match: success(report) => success(renderTestReport(report)); failure(err) => failure(err).

packages/agency-lang/stdlib/agency.agency
egonSchiele: Done in 50f64788c. cli() is one match over the argument list (["test", file], ["tc", file], …, with guards for a flag and for the wrong arity), and each subcommand helper is a match over its Result. One parser limit I hit: a match cannot be an arm's body, so ast/fmt read first and match on the parse/format result.

packages/agency-lang/stdlib/agency.agency
egonSchiele: Fixed in 50f64788c: the read happens inside the ast and fmt helpers only; test, typecheck, and run never touch the file themselves.

packages/agency-lang/stdlib/safeBash.agency
egonSchiele: Done in 50f64788c: tryAgencyPlan is a match on the command word (agency / which / type / command / _), with the lookup rule expressed as one boolean per arm.

packages/agency-lang/stdlib/safeBash.agency
egonSchiele: Plan is this file's name for the decision safeBash makes before anything runs: which effects to raise and what to execute (the Plan type in safeBash/actions.agency; planFor returns one, and echoPlan/bashPlan/refusePlan build them). tryAgencyPlan follows trySeeIfShellFree: a Result<Plan> whose failure means "not this path". The docstring now says so (50f64788c).

packages/agency-lang/stdlib/safeBash.agency
egonSchiele: Fixed in 50f64788c: runAgency passes cli()'s own failure through (truncated), and that failure now describes the call rather than pointing at the CLI.

PR #889, opened 2026-08-22
recommended policy: reads only in the launch directory; verifier gets the Agency test tools

egonSchiele:
## What

**Scoped reads in the built-in `recommended` policy** (`lib/runtime/builtinPolicies.ts`). The five read-only file effects (`std::read`, `std::readBinary`, `std::ls`, `std::glob`, `std::grep`) are approved in two places:

- the launch directory and below, written as `{.,./**}`. The matcher resolves `.` against the process cwd at match time, so the policy file `getPolicyForAgent` writes on a fresh machine keeps meaning "wherever the agent runs".
- the agency install's own `stdlib/` and `dist/` trees. `agencyGuide` and the other docs tools are `read` partially applied to `stdlib/docs/<section>`, and bundled skills are read the same way; without this rule those tools return rejections in a headless run.

No trailing `reject`: a read elsewhere is undecided, which prompts in an interactive session and auto-rejects under `-p`. `with-writes` spreads `recommended`, so it inherits the scoping. The agent's own reads of `~/.agency-agent` bypass the policy (`_internalIo`), so they need no rule.

**The verifier can test Agency code** (`stdlib/agents/verifier.agency`). Its toolkit now includes `agencyCodeTools()` (`typecheck`, `parseAST`, `testFile`) and its prompt says the `agency` command line is not available and names `testFile`.

**The Agency writer no longer carries `testFile`** (`stdlib/agents/agency/coding.agency`). `agencyCodingAgent` returns source and `code.agency` saves it afterwards, so a harness run from inside the writer always hits a file that does not exist yet. In the second fib run it did exactly that (ENOENT), then made 45 identical successful `typecheck` calls over four minutes trying to satisfy "verify with testFile" before returning. It now has `typecheck` and `parseAST` only, and its prompt says the caller saves the file and runs the tests. The verifier (which runs after the save) is where `testFile` belongs.

**Two tool-loop guards** (`lib/runtime/prompt.ts`, note in `docs/dev/tool-loop-guards.md`). Both refuse a call before it runs and tell the model why in a tool message; neither fires hooks or counts as a failure.
- *Repeated calls*: once the same tool with the same arguments (canonical JSON, so order does not matter) has returned the same result `maxRepeatedToolCalls` times in a row (default 3), the next identical call is refused: "this is call 4 to typecheck with exactly these arguments, and the previous 3 all returned the same result…". A different result restarts the count, so tools that legitimately return something new never trip it; applies to every tool regardless of markers. `llm(..., { maxRepeatedToolCalls })` / `setLlmOptions`; `0` disables.
- *Markup arguments*: a string argument starting with `</antml` or containing `<parameter name=` is the model's own call syntax leaking into an optional string it meant to leave empty (all 12 garbled calls in two eval runs were `stdin` or `flags` with this shape). The call is refused with the argument named.
- Tests: pure helpers in `prompt.test.ts`; real wiring in `tests/agency-js/repeated-tool-calls` (five identical calls → 3 runs + 2 refusals; a garbled call never runs; `maxRepeatedToolCalls: 0` runs all five).

**`codeAgent` returns the verifier's findings** (`code.agency`, `withVerification`). On the unattended paths the findings were only used to decide on a fix round and dropped on success, so the reply to "write fib.agency and report the test results" was the file contents alone. In run `190426` the coordinator, which had asked for results, then made four `codeAgent` rounds, an oracle call, and two reviews for work that had passed at minute one, until the run was interrupted. The reply now ends with `Verified:` and the findings.

**`eval grade` on a harness-graded test** (`gradeRun.ts`, `moduleGraders`). The suite fallback (config module, else the goal judge) was appended even when the test's graders are its harness pairs, so grading the fib run failed with "no goal; an LLM judge needs a goal" before the harness ran. Harness records now count as the test's own graders and the fallback is skipped (`--graders` override still applies). Test: fallback mode + harness → grades are `["h"]` only; fails without the fix.

## Why

In the fib eval, the verifier spent four minutes and ~35 tool calls looking for an `agency` executable: `ls ~`, `ls ~/agency-lang`, reading the repo's `package.json`, globbing `stdlib/`. All of it was approved because `recommended` approved every read everywhere, and none of it could work because the verifier had no `testFile`.

## Tests

- `lib/runtime/builtinPolicies.test.ts`: the read rules' shape, and picomatch checks that the install pattern covers `stdlib/docs`, `stdlib/agents/skills`, and `dist/lib/agents` but not the package root, its parent, or a home directory.
- `lib/agents/agency-agent/tests/readPolicy.agency`: through the real `checkPolicy`, reads in cwd and below approve; reads at `/`, a home directory, `/etc`, and `cwd/..` propagate.
- Verifier typechecks; `docs/site/stdlib/agents/verifier.md` regenerated by `make`.

Dev note: new section in `docs/dev/approval-policies.md`.


packages/agency-lang/lib/runtime/builtinPolicies.ts
```diff
@@ -39,11 +57,11 @@ export const recommendedAutoApprovePolicy: Policy = {
   // A sandboxed subprocess: the code it runs raises its own effects back
   // through this same policy, so approving the launch grants nothing more.
   "std::run": approve,
-  "std::read": approve,
-  "std::readBinary": approve,
-  "std::ls": approve,
-  "std::glob": approve,
-  "std::grep": approve,
+  "std::read": readScopeRules(),
```
egonSchiele, line 62: [P1] This changes the in-memory built-in, but the agent persists the selected/default policy as expanded JSON in `~/.agency-agent/policy.json` and reuses that file unchanged on later launches (`getPolicyForAgent`). As a result, every existing user whose saved recommended policy contains the old catch-all read rules keeps unrestricted reads after upgrading, so the safety change does not reach them. For newly written policies, this second rule also pins the current package installation path; after a Node/package-manager upgrade moves Agency, bundled docs and skills are read from a new path and start prompting or auto-rejecting while the obsolete path remains approved. Please add a migration/dynamic expansion mechanism rather than persisting these version/location-dependent built-in rules verbatim.

packages/agency-lang/lib/runtime/builtinPolicies.test.ts
```diff
@@ -1,20 +1,43 @@
...
-  it("resolves 'recommended' with reads approved and no write rule", () => {
+  it("resolves 'recommended' with reads scoped and no write rule", () => {
     const p = builtinPolicy("recommended", "/tmp/base");
     expect(p).not.toBeNull();
-    expect(p!["std::read"]).toEqual([{ action: "approve" }]);
+    // The matcher resolves `.` against the launch directory; the second
+    // rule is the agency install, which the docs and skills tools read.
+    for (const effect of ["std::read", "std::readBinary", "std::ls", "std::glob", "std::grep"]) {
+      expect(p![effect][0]).toEqual({ match: { dir: "{.,./**}" }, action: "approve" });
+      expect(p![effect]).toHaveLength(2);
```
egonSchiele: [P1] Please update the other consumers that assert the recommended rule shape. The current PR is red because `lib/cli/runPolicy.test.ts` still expects `std::read` to equal `[{ action: "approve" }]` in three tests (lines 15, 66, and 100); CI reports all three failures. Those tests should keep checking built-in resolution and inline-rule layering, but use the new two scoped base rules.

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -173,6 +173,70 @@ function toolErrorMessage(error: any): string {
...
+
+function repeatKey(toolName: string, args: Record<string, unknown>): string {
+  return `${toolName}\u0000${canonicalJson(args)}`;
+}
+
+type RepeatRecord = { count: number; result: string };
+
+/** Record one completed call and return how many times in a row this exact
+ *  call has now produced this exact result. A different result starts over. */
+function noteRepeat(records: Record<string, RepeatRecord>, key: string, result: string): number {
```
egonSchiele: [P1] This does not count calls "in a row": each key keeps its count while any number of different calls run. For example, `readStatus()` can return `pending`, then `advanceJob()` mutates the job, and after three such cycles the next `readStatus()` is refused because the old per-key count was never reset. More fundamentally, once a polling/read tool returns the same value three times, it is never run again, so an external state change cannot produce the new result that would reset the count. This default applies to every tool, including stateful status/queue/clock tools. Please either track an actual consecutive global sequence and reset it on any intervening call, or limit this breaker to tools whose contract makes repeated execution provably redundant; add interleaving and delayed-state-change coverage.

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -173,6 +173,70 @@ function toolErrorMessage(error: any): string {
...
+/** The name of a string argument that holds tool-call markup instead of a
+ *  value, or null. Claude sometimes emits the closing tag of its own call
+ *  syntax where an optional string it meant to leave empty belongs, and the
+ *  next parameter's text leaks in: `stdin: "</antml name=\"stdin\">\n<parameter
+ *  name=\"allowedExecutables\">[]"`. Running that call wastes a round at best
+ *  (`grep` rejects the regex flags) and at worst executes with garbage. */
+function markupArgument(args: Record<string, unknown>): string | null {
+  for (const [name, value] of Object.entries(args)) {
+    if (typeof value !== "string") continue;
+    if (value.startsWith("</antml") || value.includes("<parameter name=")) return name;
```
egonSchiele: [P1] This heuristic rejects valid string data, not just malformed calls. A source-analysis/XML tool may legitimately receive `"<parameter name=..."`, and a tool inspecting a captured Anthropic transcript may legitimately receive a string beginning with `</antml`; both are now impossible to invoke, with no opt-out. The motivating twelve cases had a narrower shape: an optional string began with its own `</antml name="<argument>">` closing tag, optionally followed by leaked parameter markup. Please match that observed full-value/argument-name shape (and ideally only optional string parameters, which requires the handler schema) rather than banning these substrings globally. Add positive tests for tools intentionally processing these literals.

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/code.agency
```diff
@@ -440,8 +440,22 @@ Fix each problem exactly.
       summary = fixed.value
       saveDraft(summary)
     }
+    return summary
   }
-  return summary
+  return withVerification(summary, checked)
```
egonSchiele: [P1] `summary` can be a solve failure here. Lines 413–415 turn a failed `supervisedSolve` into `"Coding agent stopped: ..."`, but the verifier still runs against whatever is already on disk, and this new return appends its successful findings under `Verified:`. If a stale artifact from an earlier attempt satisfies the task, the response now says the coding agent stopped and then reports that failed work as verified. Return immediately when `solved` fails, or only call `withVerification` from the successful solve branch.

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/code.agency
```diff
@@ -440,8 +440,22 @@ Fix each problem exactly.
       summary = fixed.value
       saveDraft(summary)
     }
+    return summary
```
egonSchiele: [P2] The new verification evidence is still dropped on every successful repair path. When the first check finds an error and `fixed` succeeds, this branch returns only the new source; the unattended branch does the same at line 491 even though its prompt promises "your work will be re-checked." Thus the exact request motivating this change (return the file and test results) regresses whenever one fix round is needed, and the coordinator can resume the same repeated-agent loop. Re-run the verifier on `fixed.value`/the saved artifact and append those final findings; cover both clean-first-pass and repaired-success responses.

packages/agency-lang/lib/runtime/builtinPolicies.ts
```diff
@@ -25,6 +27,22 @@ function agencyExecApproveRules(): PolicyRule[] {
...
+// still means "wherever the agent is running" on the next launch. Reads
+// anywhere else fall through: a prompt in an interactive session, an
+// automatic rejection in a headless one.
+function readScopeRules(): PolicyRule[] {
+  const install = escapeGlob(path.dirname(getStdlibDir()));
+  return [
+    { match: { dir: "{.,./**}" }, action: "approve" },
+    { match: { dir: `{${install}/stdlib/**,${install}/dist/**}` }, action: "approve" },
+  ];
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed in the latest commit. The install rule is now a token, `<agency>`, resolved by the matcher at match time (`resolveInstallDirPattern` in `lib/runtime/policy.ts`), so nothing calls `getPackageRoot` at module load. If the root cannot be found, the token stays unresolved and the rule never matches; nothing throws. Test: "leaves the token unresolved, and nothing thrown, when the root cannot be found".

packages/agency-lang/lib/agents/agency-agent/lib/policy.agency
```diff
@@ -41,7 +41,7 @@ export def givePolicyChoice() {
   },
     {
     key: "recommended",
-    label: "recommended default policy, allow reading files and browsing the web, no writes"
+    label: "recommended default policy, allow reading files in the current directory and browsing the web, no writes"
```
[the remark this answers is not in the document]
egonSchiele: Label now reads "allow reading files in the current directory (and agency's own docs and skills) and browsing the web, no writes".

packages/agency-lang/lib/runtime/builtinPolicies.ts
egonSchiele: Both parts addressed. (1) Dynamic expansion: the built-in no longer bakes a path. The second read rule is `{<agency>/stdlib/**,<agency>/dist/**}`, and the matcher resolves `<agency>` against `getPackageRoot()` at match time, the same way `.` resolves against the cwd, so a saved file keeps working after the package moves (`resolveInstallDirPattern`, tests in `policy.test.ts`). (2) Migration: `migrateCatchAllReads` (`builtinPolicies.ts`) runs in `getPolicyForAgent` on the saved file. A read effect whose rules are still exactly the old catch-all `[{action:"approve"}]` gets the scoped rules; the agent prints one line naming the effects it changed and saves. A read rule the user edited is left as written (test: the hand-scoped `std::glob` survives). Written up in `docs/dev/approval-policies.md`.

packages/agency-lang/lib/runtime/builtinPolicies.test.ts
egonSchiele: Updated the three `runPolicy.test.ts` expectations to the new two-rule base (`readScopeRules()` is exported for this); they still check built-in resolution and inline layering (`[reject, ...readScopeRules()]`). 13/13 locally.

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: Reworked to a real consecutive sequence. The loop keeps ONE streak record (key, result, count) instead of a per-key map: `noteRepeat` extends it only when both the key and the result match the immediately previous call, and anything else starts a new streak of one, so `readStatus → advanceJob → readStatus` never accumulates (test: five such cycles, count stays 1). On the second point, a refusal now resets the streak, so the call is interrupted every N identical runs rather than banned: a poll that is genuinely waiting on the world runs again after the model answers, and a changed result starts fresh (test: "a refusal restarts the count, so the call is interrupted, not banned"; the agency-js test now expects 5 identical calls → 4 runs + 1 refusal). I kept it on all tools rather than marker-gated, since with the reset it is an interruption rather than a ban; happy to gate on `idempotent` instead if you would rather.

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: Narrowed to the observed shape. `markupArgument` now takes the handler's `params` and looks only at parameters with a default (`hasDefault`), and only at a value that IS a closing `</antml…>` tag: named for that argument, or followed by leaked `<parameter` markup. Positive tests added: a required `command` given a captured transcript beginning with `</antml name="stdin">`, an XML tool given `<parameter name="x">` mid-document, and a `</antml-other>g` value all pass through. One of the twelve observed cases (`</antml name flallags">50`, a garbled tag with nothing leaked after it) is deliberately not matched and fails the old way; noted in the dev doc.

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/code.agency
egonSchiele: Fixed: `escalateToExpert` tracks whether the solve completed and attaches `Verified:` only in that case; after a failed solve the "Coding agent stopped" summary is returned as-is.

packages/agency-lang/lib/agents/agency-agent/lib/policy.agency
```diff
@@ -41,7 +42,7 @@ export def givePolicyChoice() {
   },
     {
     key: "recommended",
-    label: "recommended default policy, allow reading files and browsing the web, no writes"
+    label: "recommended default policy, allow reading files in the current directory (and agency's own docs and skills) and browsing the web, no writes"
```
egonSchiele: undo this change please. We want to keep these labels short and the change you have added is not strictly needed

packages/agency-lang/lib/agents/agency-agent/lib/policy.agency
```diff
@@ -78,7 +79,18 @@ export def getPolicyForAgent(
         process.exit(1)
       }
     } else {
-      setPolicy(POLICY_PATH, policy.value)
+      // A saved copy of an older recommended policy approved every read
```
egonSchiele: No need for backwards compatibility. Remove this, please.

packages/agency-lang/lib/eval/grading/gradeRun.ts
```diff
@@ -158,6 +160,12 @@ async function moduleGraders(
   if (entry.test.graders !== undefined) {
     return cache(entry.test.graders);
   }
+  // A harness-graded test carries its own graders, so the fallback (a
```
egonSchiele: What does harness graded mean? Does that mean that it has agency tests to run? If so, I don't get that meaning from the harness graded or hasHarness terminology.

packages/agency-lang/lib/runtime/policy.ts
```diff
@@ -103,6 +104,29 @@ export function resolveDotDirPattern(pattern: string, cwd: string = process.cwd(
   );
 }
 
+/** The token a `dir` pattern uses for the agency install itself, so a rule
+ *  approving reads of the shipped docs and skills can be saved to a policy
+ *  file without pinning the install path of one machine or one version. */
+export const INSTALL_TOKEN = "<agency>";
```
egonSchiele: Again, the name `INSTALL_TOKEN` does not make it obvious to me what this is.

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -173,6 +173,105 @@ function toolErrorMessage(error: any): string {
...
+    if (tag[0].includes(param.name) || rest.trimStart().startsWith("<parameter")) {
+      return param.name;
+    }
+  }
+  return null;
+}
+
+/** JSON with object keys sorted at every level, so two calls that pass the
+ *  same arguments in a different order get the same key. */
+function canonicalJson(value: unknown): string {
```
egonSchiele: Surely these functions can go in some sort of util file instead of bloating a file that is already quite large.

egonSchiele: Also, is there a performance concern with any of this? What if the value is binary data or what if the value is very large?

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/code.agency
egonSchiele: Done on both paths. After a successful repair, `escalateToExpert` and `solveUnattended` run the verifier again and return its findings with the repaired result. `withVerification` heads them "Verified:" when clean and "Re-check found these problems:" when errors remain, so a partial repair is never reported as verified.

packages/agency-lang/lib/agents/agency-agent/lib/policy.agency
egonSchiele: Reverted to the original label.

packages/agency-lang/lib/agents/agency-agent/lib/policy.agency
egonSchiele: Removed: `migrateCatchAllReads`, its bridge into `std::policy`, the call in `getPolicyForAgent`, and its test. The dev note now says a file saved before this change keeps its old rules and can be deleted or edited.

packages/agency-lang/lib/eval/grading/gradeRun.ts
egonSchiele: Yes, exactly that: the test has Agency tests (`agencyTests` in its spec, stored on the run row as harness records) to run against the agent's output. Renamed the parameter to `hasAgencyTests` and rewrote the comment and the doc line in those words; the test is now "a test with Agency tests gets no fallback grader: those tests are its graders".

packages/agency-lang/lib/runtime/policy.ts
egonSchiele: Renamed to `AGENCY_INSTALL_DIR_PLACEHOLDER` ("in a dir pattern, `<agency>` stands for the directory the agency package is installed in") and the function to `expandAgencyInstallDir`.

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: Moved all of it into `lib/runtime/toolLoopGuards.ts` (with its own test file); `prompt.ts` now only imports and calls the helpers. On performance: the streak now stores SHA-256 digests of the canonical arguments and of the stringified result, never the text, so a tool that takes a whole source file or returns megabytes (or a Buffer, which stringifies as JSON like any other result) puts 64 hex chars on the frame, not the payload. The remaining cost is one hash per call, linear in size, over a string the loop already builds to send the result to the model.

PR #890, opened 2026-08-23
Eval tracking PR 1: trials, batch statistics, silent-run summaries, agency eval upload

egonSchiele:
PR 1 of 4 for tracking the agency agent's eval scores over time on statelog. This is the agency-lang half: run a suite several times, get honest statistics, and upload the run directories. Spec: `docs/superpowers/specs/2026-08-22-eval-tracking-pipeline-design.md`; plan: `docs/superpowers/plans/2026-08-22-eval-tracking-agency-lang.md`; dev note: `docs/dev/eval-tracking.md`.

## What changes

- **`eval run --trials k`**: each test runs `k` times, trial-major (`a/1, b/1, a/2, b/2`), each at `<out>/<test>/<trial>/` (one trial keeps the flat layout). Every harness `run` row now records `batch` (the group directory name), `trial`, and `flags.trials`. `findRunDirectories` looks two levels down and never enters `.staging`.
- **Batch statistics** (`lib/eval/batchStatistics.ts`): per-test mean ± SE and batch accuracy with an SE over per-trial batch means, paired by trial index. Refuses uneven grids, duplicate/gapped trial indices, and mixed batches. `eval grade` prints one block per multi-trial batch, grouped by batch id so `eval grade a b` never merges unrelated batches.
- **One run summary for traced and silent runs** (`lib/runDirectory/list.ts`): a run that died before its first event now summarizes (status `failed`, zero metrics) instead of vanishing. The summary gains `status`, `endedAt`, `eventCount`, `suiteSource`/`suiteSha`, `batch`, `trial`, and `score` (grading's rule: a run that did not finish counts as 0; grading writes no score row for it, so without this a batch mean would skip crashes). `summarizeEvalRun(input)` takes canonical rows so statelog never rebuilds a directory snapshot.
- **`agency eval upload <paths…>`**: uploads run directories to the linked project. Per run it asks the server for a proven trace state (`missing` / `empty` / `live` / `bulk-prefix` / `invalid`), then uploads all, resumes a bulk prefix, skips, or refuses (partial live and invalid traces are never touched). Events go in 500-event chunks with a 0-based `sequence`, then annotations upsert by deterministic id. A silent run creates its trace by posting `events: []`. Idempotent. No host/project/key flags by design.
- **Agent names are URL path segments**: `setAgentName` now rejects whitespace, punctuation, and empty/`.`/`..` segments (they fold out of a URL even when encoded). The agency agent names itself `agency-agent/<brain>`.
- **`agency-lang/eval` exports** the summary, statistics, annotation schema/id, and the agent-name rule; deliberately not the run-directory reader, trace parser, or annotation fold.

## Behavior changes to know about

- `agency runs list` now lists a silent run as a row (test id, `error`), not just a footer count.
- `findRunDirectories` on a directory of flat groups (e.g. `runs/`) now finds those groups' runs (two levels), where before it found nothing.
- An existing `setAgentName("name with spaces")` call throws at runtime now.

## Deviations from the spec (also in the dev note)

- Upload state is a discriminated union, not `{ count, maxSequence }`: the latter cannot tell a partly live-streamed trace from a resumable bulk prefix.
- The summary gained more than `batch`/`trial` (status, timestamps, score, silent runs) so statelog does not derive any of it.

## What PR 2 (statelog) must provide

`logs.sequence` + unique `(trace_id, sequence)`; `POST /api/projects/:slug/logs/bulk` that creates a missing trace and accepts `events: []`; `GET /api/projects/:slug/traces/:traceId/upload-state`; `POST /api/projects/:slug/annotations` that recomputes `annotationId` before writing; a per-trace summary via `summarizeEvalRun`. Listed in `docs/dev/eval-tracking.md`.

## Tests

New/updated unit tests for the schema, summaries (incl. silent runs and the score rule), discovery, `--trials` layout and validation, batch statistics, grade output, agent-name rule and stdlib guard, the sealed client, `eventPlan` and the upload orchestration, CLI shape, and the public surface. One Agency test (`agentName.test.json`). `make`, `lint:structure`, `fmt:ts`, typecheck, and the text/binary guards pass.


packages/agency-lang/lib/cli/eval/upload.ts
```diff
@@ -0,0 +1,285 @@
...
+  targets: string[],
+  target: EvalUploadTarget,
+  dependencies: EvalUploadDependencies = {},
+): Promise<EvalUploadResult> {
+  const client =
+    dependencies.client ?? createEvalUploadClient(target.origin, target.projectSlug, target.apiKey);
+  const reportWarning = dependencies.reportWarning ?? ((message) => console.warn(message));
+  const dirs = uniqueRunDirectories(findRunDirectories(targets));
+  const records: RunRecord[] = [];
+  for (const dir of dirs) {
```
egonSchiele: You could do this as a map and await all the promises to have everything upload in parallel.

packages/agency-lang/lib/eval/run/runSuite.ts
```diff
@@ -130,36 +138,44 @@ export async function runSuite(
...
-    const runDir = path.join(groupDir, testId);
-    const idProblem = testIdProblem(testId, groupDir, runDir);
+    const testDir = path.join(groupDir, testId);
+    const runDir = runDirFor(testDir, trial, trials);
+    const idProblem = testIdProblem(testId, groupDir, testDir);
     if (idProblem !== undefined) {
-      return { testId, traceId, runDir, status: "error", errorMessage: idProblem };
+      return { testId, trial, traceId, runDir, status: "error", errorMessage: idProblem };
     }
     if (fs.existsSync(runDir)) {
```
egonSchiele: **Important:** This only checks the nested trial directory. If `<out>/<test>/statelog.jsonl` already exists from a prior one-trial run, a later `--trials 2` invocation sees `<out>/<test>/1` as free and writes both trials beneath the existing run. `findRunDirectories` then classifies `<out>/<test>` as a run and never descends, so the newly executed trials are hidden from grade/upload. Please reject a multi-trial job when `testDir` is already a run directory (and add a flat-run-then-`--trials` regression test).

packages/agency-lang/lib/eval/run/runSuite.ts
```diff
@@ -130,36 +138,44 @@ export async function runSuite(
   const flags: Record<string, string | number | boolean> = {
     parallel,
     agent: target.label,
+    trials,
   };
   const harness = { kind: "harness" as const, id: `agency-eval@${harnessVersion()}` };
+  // The batch is the group directory's name: one suite invocation, so every
+  // trial of every test in it carries the same id wherever the rows travel.
+  const batch = path.basename(groupDir);
```
egonSchiele: **Important:** The basename is not a unique suite-invocation id. For example, `/tmp/team-a/nightly` and `/tmp/team-b/nightly` both persist `batch: "nightly"`; grading both paths merges them in `batchStatisticsByBatch`, and uploading both to one project merges them under `(project_id, batch)` as well. That contradicts the stated invariant that unrelated batches never merge. Please mint a durable unique batch id per invocation rather than deriving identity solely from a user-chosen basename (the basename can remain a display label if needed).

packages/agency-lang/lib/cli/statelog/evalUploadClient.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+const remoteTraceStateSchema: z.ZodType<RemoteTraceState> = z.discriminatedUnion("kind", [
+  z.object({ kind: z.literal("missing") }),
+  z.object({ kind: z.literal("empty") }),
+  z.object({ kind: z.literal("live"), eventCount: z.number().int().nonnegative() }),
+  z.object({
+    kind: z.literal("bulk-prefix"),
+    eventCount: z.number().int().nonnegative(),
+    nextSequence: z.number().int().nonnegative(),
```
egonSchiele, line 60: **Important:** The documented `bulk-prefix` invariant requires `eventCount === nextSequence`, but the response schema accepts inconsistent values. A response such as `{ eventCount: 2, nextSequence: 100 }` makes `eventPlan` resume at 100, upload no remaining events, and then post annotations for an incomplete trace. Since this parser is the sealed wire boundary, please refine the schema to reject inconsistent bulk-prefix states (and test the malformed response).

packages/agency-lang/lib/runDirectory/findRuns.ts
```diff
@@ -20,11 +21,32 @@ export function isRunDirectory(dir: string): boolean {
 }
 
 export function childRunDirectories(dir: string): string[] {
+  const found: string[] = [];
+  for (const child of childDirectories(dir)) {
+    if (isRunDirectory(child)) {
+      found.push(child);
+    } else if (path.basename(child) !== STAGING_DIR) {
+      found.push(...childDirectories(child).filter(isRunDirectory));
```
egonSchiele, line 29: This intended two-level discovery change currently leaves the PR's `integration` check red. `tests/integration/cli/test.mjs` still expects `agency eval grade eval-runs` to report `not a run directory`; it now discovers `smoke/input-1` and reaches grading, producing the missing-goal error instead. Please update that integration assertion to cover the new discovery behavior (for example, grade with an explicit goal) so CI passes.

egonSchiele (commented):
I found three correctness issues and one CI regression that should be addressed before merge.

packages/agency-lang/lib/statelog/agentName.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+export const AGENT_NAME_MAX_LENGTH = 200;
+
+/** Letters, digits, `.`, `_`, `-`, and `/` between segments. */
+export const AGENT_NAME_PATTERN = /^[A-Za-z0-9._/-]+$/;
+
+/** Why `name` is not a valid agent name, or null when it is. */
+export function agentNameProblem(name: string): string | null {
+  if (name === "") {
+    return "an agent name must not be empty";
```
egonSchiele, line 18: We should add all these checks in the standard library function that lets you set the agent name.

packages/agency-lang/lib/cli/eval/upload.ts
egonSchiele: Done in 7b6b98a9d: `evalUpload` now runs `uploadRun` over every directory with `Promise.all`; results keep input order.

packages/agency-lang/lib/eval/run/runSuite.ts
egonSchiele: Done in 7b6b98a9d: a multi-trial job now errors when `<out>/<test>` is already a run directory ("is already a run directory (a single-trial run); use another --out"), and nothing is written beneath it. Regression test: "--trials over a group that already holds a flat run for the test is an error, not hidden trials".

packages/agency-lang/lib/eval/run/runSuite.ts
egonSchiele: Done in 7b6b98a9d: the batch id is now `<group name>-<8-char nanoid>` minted once per `runSuite` call, so `team-a/nightly` and `team-b/nightly` get different ids; the group name stays as the readable prefix. Test: "two invocations of the same --out name never share a batch id".

packages/agency-lang/lib/cli/statelog/evalUploadClient.ts
egonSchiele: Done in 7b6b98a9d: the client rejects a `bulk-prefix` whose `eventCount` and `nextSequence` disagree (`consistentState`, applied right after the zod parse), with a test for `{ eventCount: 2, nextSequence: 100 }`. Kept as a post-parse check because zod discriminated unions only take plain objects, not refined ones.

packages/agency-lang/lib/runDirectory/findRuns.ts
egonSchiele: Done in 7b6b98a9d: the integration test now grades `eval-runs` with `--goal` and expects both nested runs ("mean 1.000 over 2 runs"), and checks the not-a-run-directory refusal on a fresh `not-runs/` folder instead.

packages/agency-lang/lib/statelog/agentName.ts
egonSchiele: The standard library function already does these checks: `setAgentName` in `stdlib/statelog.agency` calls `_setAgentName` (`lib/stdlib/statelog.ts`), which runs `agentNameProblem` first and throws `setAgentName: <reason>` before it looks for a frame, so a bad name fails at the call site inside or outside an Agency run. Covered by "setAgentName rejects a name that cannot be a URL segment, before looking for a frame" in `lib/stdlib/statelog.test.ts`, and the docstring spells out the rule. Happy to move the check into the `.agency` wrapper itself if you would rather see it there.

Issue #891, opened 2026-08-23
agency doc: a @hidden marker so eval entry nodes stay out of the stdlib reference


egonSchiele:
## What

A way to keep a declaration out of the `agency doc` output, such as a `@hidden` tag in its doc comment (or a rule that nodes are skipped by default).

## Why

`stdlib/agents/agency/review.agency` now carries an eval entry node, `evalMain(input: ReviewEvalInput): Feedback[]`, so an eval suite can score `agencyReviewAgent` with `agency eval run stdlib/agents/agency/review.agency:evalMain --suite ...` and no wrapper file. A node in a library module never runs when the module is imported, so it is harmless to users of the module. But `agency doc` renders it as part of the module's public API: `docs/site/stdlib/agents/agency/review.md` now has a `### evalMain` section next to `agencyReviewAgent`, and `### ReviewEvalInput` next to the real types.

The same will apply to every stdlib agent that gains an eval entry node, which is the plan for the review and coding agents at least.

## Options

1. A `@hidden` tag in a doc comment (`/** @hidden ... */`) that `agency doc` honors for any declaration.
2. Skip nodes in library modules by default. Simpler, but it hides a node someone documents on purpose.
3. A naming rule (`eval*` nodes are skipped). Least machinery, most magic.

Option 1 seems right: explicit, and it also covers helper types like `ReviewEvalInput` that exist only for the node.

## Where

`agency doc` lives under `lib/cli/doc*` and `lib/doc/`; the conventions are in `docs/site/cli/doc.md`.

egonSchiele:
Fixed in #905 (squash-merged as 30d8b4397). Noting what shipped, since the merge auto-closed this.

Option 1 from the issue: a `@hidden` tag on the declaration. No language change was needed — bare `@hidden` already parsed, typechecked, and survived `agency fmt`, and tag attachment already reached the declarations `agency doc` renders.

Applied to `ReviewEvalInput` and `CodingEvalInput`. Both `evalMain` nodes are gone from the reference too, via a separate fix below.

Beyond what the issue asked for:

- **`std::agency`'s `describe()` applies the same rule**, including through re-export chains. Its docstring already claimed to share the doc visibility rule and it reports exported types, so otherwise the two tools would have disagreed about these exact modules.
- **Nodes are now filtered on `export`**, like functions, types, and constants always were. That inconsistency is why a non-exported `evalMain` rendered at all. Across the whole stdlib it removed exactly those two nodes and nothing else.
[...]

PR #892, opened 2026-08-23
agency-review eval suite, evalMain on the review agent, hosted search by route, graders belong to tests

egonSchiele:
Evals for one stdlib agent at a time, starting with the Agency code reviewer. Four things travel together because each one was found by the one before it.

## 1. `evals/agency-review`: a suite for Agency code reviewers

The suite describes the job and never names an agent. Input is `{ task, source }`; output is the stdlib's `Feedback[]` (`[{ error, feedback }]`). Each test directory carries its own `graders.ts` beside `test.json`. Three tests, every planted source typechecking clean, because those are the bugs a reviewer that only runs the typechecker cannot see:

- `fib-off-by-one` (bug): `fib(0)` returns 1. Gate: rejects. Judge: names the base case.
- `fib-correct` (clean): gate: no error finding.
- `ungated-delete` (bug, Agency-specific): the task says leave the deletion to the caller; the code does `remove(file) with approve`. Gate: rejects. Judges: names the `with approve`, and every error finding is real (the first run rejected for three invented reasons, including `for (file of logs)`).

The gate is deterministic, so the headline number needs no model; the judges run only where they mean something.

```
pnpm run agency eval run stdlib/agents/agency/review.agency:evalMain --suite evals/agency-review --out runs/agency-review
pnpm run agency eval grade runs/agency-review
```

Three runs of the stdlib reviewer scored 0.92, 0.82, 0.42: it caught the `with approve` once in three, flagged the correct fib once, and invented JavaScript syntax complaints. That is the variance `--trials` exists to measure, and a clear reading that the reviewer's one-sentence prompt needs Agency-specific guidance.

## 2. `agencyReviewAgent` carries its eval entry node

`stdlib/agents/agency/review.agency` gains `export type ReviewEvalInput` and `node evalMain(input: ReviewEvalInput): Feedback[]`. A node in a library module never runs when the module is imported (checked: the entry file's `main` runs, the library's does not, no name collision), so the suite scores the reviewer with `review.agency:evalMain` and no wrapper file. Any other reviewer supplies a node with the same signature. `agency doc` renders the node in the reference for now; #891 asks for a `@hidden` marker.

## 3. Hosted search follows the route (a bug the first run found)

`hostedSearchTools` requested `web_search` for `gpt-4o-mini` by provider *family*: `"openai"` prefix-matched the `"openai-responses"` hosted entry. Under this repo's `agency.json` (base `openai` provider) every stdlib agent therefore asked a client for a capability it lacks; the call failed, `failOpenFeedback` swallowed it, and `agencyReviewAgent(source, task)` returned **no findings** with only a log line to show for it. Inside the agency agent this was masked by provider promotion; used on its own, the reviewer was silently broken.

`_hostedSearchTools` (`lib/stdlib/llm.ts`) now resolves the model and provider the call will actually use (explicit > branch `setLlmOptions` > baked `agency.json`) and asks smoltalk's `getHostedTools` about that exact pair, the same check that later rejects an unsupported request. `hostedSearchTools(model, provider)` gains the provider and all eight stdlib agents pass their own through (each had the parameter and was dropping it). Unknown models and an empty model with no default still err open. Tests: `llm.hostedSearchTools.test.ts` (the route split fails on the old code), `tests/agency/agents/searchTools.agency`.

## 4. Graders belong to tests: `eval.graders`, `--graders`, `inputScope` removed; `eval grade --suite` added

Every suite in the repo writes a `graders.ts` per test, and the three ways of grading from somewhere else only muddied the precedence (`inputScope` also read tags from `metadata.tags` while `--tags` reads `tags`). All three are gone. `GradingContext.graders` is a `GraderSource`:

- `snapshot` (default): the copy the run directory stored when the agent ran.
- `suite` (`eval grade <runs> --suite <dir>`): each run graded by its test's *current* `graders.ts` and harness pairs, matched by test id, so an improved grader re-scores old runs without re-running the agent. A run whose test is not in the suite is an error naming both.
- `override`: one set for every test. The optimizer's objective; never a CLI option.

A test with no graders is scored by the bundled goal judge against its `goal`. `--goal` and `--suite` are exclusive, as `--goal` and `--graders` were.

Not changed: `eval optimize --graders` and `eval.optimize.graders`. The optimizer's single grader set is the objective it optimizes (`baseOptimizer.ts:277`), a different thing; if it should move to per-test graders too, that is its own change.

Docs: `docs/site/cli/eval.md` (also: the agent is a positional argument, and a `Grade`'s score carries its kind), `docs/site/guide/graders.md`, `docs/misc/config.md`, `docs/dev/eval-grading.md`, `docs/dev/run-directory.md`.

## 5. `eval run` exits 1 when any test errored

Found while iterating: with an existing `--out` directory every test was refused ("run directory already exists") and the command still exited 0 with "0/3 tests ok". The run still completes and writes everything; the exit code now reports that not every test ran to an end, the way `eval grade` exits 2 on a failed gate. The options list in `eval.md` also described flags `eval run` no longer has (`--goal`, `--run-id`, `--runs-dir`, `--no-continue-on-error`, `--strict`) and missed `--input`, `--test`, `--tags`, `--out`, `--trials`; rewritten from `--help`.

## Verification

`make` clean; `lib/cli/eval`, `lib/eval`, `lib/optimize`, `lib/config.test.ts`, `lib/runDirectory` (94 files, 990 tests) pass; `lint:structure`, prettier, the text/binary guard, and `tsc --noEmit` pass. `eval grade runs/x --suite evals/agency-review` re-grades the real run; `--suite evals/smoke` refuses it by name. The review suite ran three times through the node with no hosted-search errors.

https://claude.ai/code/session_01B1f3ZTxh92pD7N4DDbDA3h


packages/agency-lang/lib/stdlib/llm.ts
```diff
@@ -115,47 +115,51 @@ export function _hostedModelInfo(name: string): HostedModelInfo | null {
...
+ *  overrides ("" = none): branch defaults first, then the baked config. Reads
+ *  `agencyStore` directly rather than `getRuntimeContext()`, which throws
+ *  outside an execution frame; no frame just means no defaults. */
+function resolveLlmRoute(model: string, provider: string): { model: string; provider: string } {
+  const store = agencyStore.getStore();
+  const branch = (store?.stack?.other.llmDefaults ?? {}) as { model?: string; provider?: string };
+  const baked = (store?.ctx?.smoltalkDefaults ?? {}) as { model?: string; provider?: string };
+  return {
+    model: model || branch.model || baked.model || "",
+    provider: provider || branch.provider || baked.provider || "",
```
egonSchiele, line 169: This resolves the model and provider independently, but the actual `llm()` route does not inherit an ambient provider when an explicit model is supplied with `provider: ""`. `llmOptions` emits that empty provider, the per-call config overwrites branch/baked defaults, and smoltalk then auto-resolves the provider from the explicit model. For example, with ambient `openai-responses`, `hostedSearchTools("gemini-2.5-flash", "")` checks Gemini against `openai-responses` and returns no search tool, while the subsequent LLM call routes through Gemini's catalog provider and does support search. Please resolve model/provider as one route: when a model override is present without a provider override, use that model's catalog provider rather than the ambient provider, and add this mixed-precedence case to the test.

packages/agency-lang/docs/site/cli/eval.md
```diff
@@ -47,26 +47,27 @@ A test says nothing about the agent. Tests describe the work; whoever runs the e
 For a single ad-hoc run, use `--goal` instead of `--suite`:
 
 ```bash
-agency eval run --agent agent.agency --goal "Answer with a concise summary"
+agency eval run agent.agency --goal "Answer with a concise summary"
```
egonSchiele: This example still uses the removed `eval run --goal` option; `eval run` now exposes `--input`, so following it produces an unknown-option error. The synopsis on line 11 has the same stale `--goal`, and the suite example on line 27 still uses removed `--run-id` instead of `--out`. Also, line 43 still says `input` is required and entry nodes always take one parameter, while the runner supports no-input suites with zero-parameter nodes. Could you update this section from the current `eval run --help` contract?

egonSchiele (commented):
Two actionable issues found. Separately, the current CI run also fails formatting (`baseGrader.ts`, `gradeRun.ts`) and typechecking (`run.test.ts:368-369` still passes `graders` to an inferred narrower test type).

packages/agency-lang/stdlib/agents/agency/review.agency
```diff
@@ -141,3 +142,24 @@ export def agencyReviewAgent(
...
+}
+
+/** Eval entry point: `agency eval run stdlib/agents/agency/review.agency:evalMain
+  --suite <dir>`. A node in a library module never runs when the module is
+  imported; it only exists so a suite can score this reviewer without a
+  wrapper file. Any other reviewer scored on the same suite supplies a node
+  with this signature. */
+node evalMain(input: ReviewEvalInput): Feedback[] {
+  evalValue(input)
+  const reviewed = agencyReviewAgent(source: input.source, task: input.task)
```
egonSchiele: `agencyReviewAgent` raises `std::guard`, but this new root node has no handler. The stdlib fixture check now fails with AG3009 at this call, and more importantly a budget trip escapes the eval entry point as an interrupt instead of deterministically producing the agent’s salvaged/failure result. Please handle this explicitly at the entry boundary (likely `with reject`, so the eval’s `$10`/`10m` cap remains a cap rather than being auto-approved) and verify the stdlib typecheck fixture is warning-free.

packages/agency-lang/evals/agency-review/fib-off-by-one/graders.ts
```diff
@@ -0,0 +1,31 @@
...
+const findings = (output: unknown): Feedback[] => (Array.isArray(output) ? output : []);
+const errors = (output: unknown) => findings(output).filter((item) => item?.error === true);
+
+export default [
+  grader(
+    ({ output }) => ({
+      score: { kind: "binary", pass: errors(output).length > 0 },
+      feedback: `${errors(output).length} error finding(s), ${findings(output).length} total`,
+    }),
+    { name: "rejects", mustPass: true },
```
egonSchiele: I think this is the wrong place to add a gate. This means that if the grader did not identify the correct number of issues, we don't even bother to check what it has found. This means that it will be really hard to see incremental scores and to see if the agent is improving over time, because even if it does everything else correctly, as long as it continues to return the wrong number of findings, it will keep getting the same score.

packages/agency-lang/evals/agency-review/ungated-delete/graders.ts
```diff
@@ -0,0 +1,48 @@
...
+const findings = (output: unknown): Feedback[] => (Array.isArray(output) ? output : []);
+const errors = (output: unknown) => findings(output).filter((item) => item?.error === true);
+
+export default [
+  grader(
+    ({ output }) => ({
+      score: { kind: "binary", pass: errors(output).length > 0 },
+      feedback: `${errors(output).length} error finding(s), ${findings(output).length} total`,
+    }),
+    { name: "rejects", mustPass: true },
```
egonSchiele: same comment, wrong place to gate

packages/agency-lang/evals/agency-review/ungated-delete/test.json
```diff
@@ -0,0 +1,11 @@
+{
+  "description": "Agency-specific: the task says deletion must wait for the user's approval, but the code approves its own destructive effect with `with approve`. Typechecks cleanly.",
+  "tags": [
+    "bug",
+    "handlers"
+  ],
+  "input": {
+    "task": "Write an Agency program with a node main that deletes every file in ./tmp whose name ends in .log. The deletion is destructive, so it must be left for the caller to approve or reject: do not approve it inside the program.",
```
egonSchiele: I'm not sure I understand this. Why is the source being put in the JSON instead of in a separate file? Why does the task say to write an agency program instead of review an agency program?

packages/agency-lang/lib/stdlib/llm.ts
egonSchiele: Fixed in 335bd8e48. Verified the mechanism you described: llmOptions sets provider (possibly "") whenever a model is named, the per-call spread overwrites the branch/baked pair, and smoltalk resolveProvider falls back to the catalog on a falsy provider. resolveLlmRoute now returns {model, provider} untouched whenever a model is named, and only a call with no model override falls back to the ambient pair. Added the mixed-precedence tests inside a fake ALS frame: ambient openai-responses + explicit gemini-2.5-flash -> web_search (failed before the fix), explicit provider still wins, and the ambient pair decides when no model is named.

packages/agency-lang/stdlib/agents/agency/review.agency
egonSchiele: Fixed in 335bd8e48. evalMain now calls the reviewer `with reject`, so the $10/10m caps stay caps: a guard trip comes back as the fail-open Feedback[] result instead of an interrupt escaping the entry node. The seeded input file read is `with approve` at the same boundary. `agency typecheck stdlib/agents/agency/review.agency` is warning-free and make is clean.

packages/agency-lang/evals/agency-review/fib-off-by-one/graders.ts
egonSchiele: Removed the gate in 335bd8e48 — here, in ungated-delete, and the no-false-positive gate in fib-correct, which had the same problem. Every grader now costs only its own share, so a reviewer that starts naming the bug or cutting noise moves the score while its verdict is still wrong. On the re-run, fib-correct went from pinning at 0 (when it false-positived) to 0.5, which is the granularity you described.

packages/agency-lang/evals/agency-review/ungated-delete/graders.ts
egonSchiele: Same fix, 335bd8e48 — gates removed across the suite.

packages/agency-lang/evals/agency-review/ungated-delete/test.json
egonSchiele: Two answers, and a change for the first. (1) The source is now a file: each test carries files/solution.agency, seeded into the workdir, and the input became { task, sourceFile } (335bd8e48). Inline JSON was left over from the first prototype; a real .agency file is easier to read and can be typechecked directly. (2) The task field is the assignment the code under review was answering — the reviewer judges the code against it, so it reads "Write an Agency program that…" the way the original task did. The instruction to review lives in the review agent itself. The suite README now says this explicitly; if the field name still reads wrong I can rename it, e.g. to something like `assignment`.

packages/agency-lang/docs/site/cli/eval.md
egonSchiele: Fixed in 335bd8e48: synopsis and the one-off example use --input/--out, the --run-id example is gone, and the input paragraph now says input is optional and zero-parameter entry nodes run no-input suites (verified against Test.input in runTypes.ts). CI formatting (baseGrader.ts, gradeRun.ts) and the run.test.ts type error are also fixed in that commit — pnpm run typecheck (all three projects) passes now; my earlier check only ran tsc -p . and missed tsconfig.tests.json.

packages/agency-lang/lib/stdlib/llm.ts
```diff
@@ -115,47 +115,51 @@ export function _hostedModelInfo(name: string): HostedModelInfo | null {
...
-    effective = defaults.model ?? "";
-  }
-  if (effective === "") {
+ *  The answer depends on the ROUTE, not the model family: gpt-4o-mini has
+ *  hosted search through "openai-responses" and none through the base
+ *  "openai" client. So this resolves the model and provider the way the
+ *  call will (explicit argument > branch `setLlmOptions` default > baked
+ *  agency.json default > smoltalk's catalog provider for the model) and
+ *  asks smoltalk's `getHostedTools` about exactly that pair, which is the
+ *  same check that later rejects an unsupported request.
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 335bd8e48: the comment moved onto resolveLlmRoute and now describes the actual rule (named model -> named provider, else catalog, never ambient; ambient pair only when no model is named).

packages/agency-lang/docs/site/cli/eval.md
```diff
@@ -8,8 +8,8 @@ description: How to run an Agency agent against a test suite, score it with grad
 `agency eval` runs, grades and compares agent runs. Everything it produces lives in a **run directory**: a folder with the runs' statelog and an append-only file of annotations (grades, checklist answers), plus a free-form `notes.md` you write yourself. The main commands are:
 
 ```
-agency eval run (--agent <file>[:<node>] | --agent-cmd '<command with {input}>') (--suite <file|dir|git-url> | --goal <text>) [-n <count>]
-agency eval grade <runDir> [--graders <file>]
+agency eval run (<file>[:<node>] | --agent-cmd '<command with {input}>') (--suite <file|dir|git-url> | --goal <text>) [-n <count>]
+agency eval grade <runDir> [--suite <file|dir|git-url>] [--goal <text>]
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 335bd8e48 (synopsis uses --input; --goal remains only on grade/optimize).

packages/agency-lang/docs/site/cli/eval.md
```diff
@@ -24,7 +24,7 @@ agency runs list <runDir>
 `agency eval run` runs an Agency agent against every test in a suite and writes a run directory:
 
 ```bash
-agency eval run --agent agent.agency:evalMain --suite suite.json --run-id smoke
+agency eval run agent.agency:evalMain --suite suite.json --run-id smoke
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 335bd8e48 (--out).

packages/agency-lang/evals/agency-review/ungated-delete/test.json
egonSchiele: Renamed to `assignment` (type, evalMain, the three test.json inputs, README) in the follow-up commit. agencyReviewAgent(task:) itself is unchanged; evalMain maps assignment -> task.

PR #898, opened 2026-08-23
agency-coding eval suite: harness-graded tests for the Agency writer, with evalMain

egonSchiele:
## What this is

The companion suite to `evals/agency-review`: that one scores an agent that reads Agency code, this one scores the agent that writes it (`agencyCodingAgent`). Two tests to get it going, per our discussion. The two suites feed each other: a program this agent writes badly is a candidate reviewer test with its oracle already attached.

## How grading works

There is no `graders.ts` anywhere in this suite. Each test carries a hidden harness pair in `holdout/` (an `.agency` file importing the saved solution plus a `.test.json` of expected node outputs). The eval framework already discovers these pairs and grades the run's working directory with `agency test --agency-only --reject '*'` as a must-pass `AgencyTestGrader` — the score is the passing fraction, so partially correct programs earn partial credit. `holdout/` is never seeded into the workdir, so the writer cannot code to the oracle.

Both oracles are verified at authoring time: my reference solutions pass 5/5, and an off-by-one wrong solution fails 4/5.

## The entry node

`coding.agency` gains `CodingEvalInput` (`{assignment, outFile, requireMain}`) and `node evalMain`, mirroring the reviewer's entry node. The writer returns source without saving it ("the caller saves the code you return"), so evalMain is that caller: it saves into the workdir (`with approve`), rejects budget trips so caps stay caps, and saves even a failed run's last draft so the harness can score what partial credit it earns.

## The tests

- **sum-multiples** — sum the multiples of 3 or 5 below n; the simplest module deliverable, one exported def and a loop.
- **reverse-digits** — reverse an integer's digits. Needs a while loop and hand-built integer division (Agency has no C-style `for` and no `floor` in the stdlib), so JS reflexes do not transfer directly.

## First real run: the suite already found something

```
pnpm run agency eval run stdlib/agents/agency/coding.agency:evalMain --suite evals/agency-coding --out runs/agency-coding-smoke
# 2/2 tests ok, total LLM cost: $0.01   <- nonzero: the smoltalk 0.12.0 cost fix is confirmed live
pnpm run agency eval grade runs/agency-coding-smoke -n 2
# both tests: score 0.000 — "Function 'sumMultiples' in './solution.agency' is not exported"
```

The agent solved both tasks with correct logic but omitted `export` on the defs, despite the assignment saying "exports def …". Root cause is visible in the agent: `checkSource` validates review findings, compile, and `node main`, but never checks exports, and every example in the system prompt is a program, never a module. I left the agent untouched here — the suite's job is to surface this, and the fix is its own change (likely: teach `checkSource` to require the exports the task names, or add a module-shaped example to the prompt).

The smoke run is in my local `runs/agency-coding-smoke` (gitignored) if you want to inspect it.

https://claude.ai/code/session_01B1f3ZTxh92pD7N4DDbDA3h

### DIFF
--- a/packages/agency-lang/docs/site/stdlib/agents/agency/coding.md
+++ b/packages/agency-lang/docs/site/stdlib/agents/agency/coding.md
@@ -28,7 +28,28 @@ export type WriteFailure = {
 }
 ```
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/agents/agency/coding.agency#L26))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/agents/agency/coding.agency#L27))
+
+### CodingEvalInput
+
+What an eval hands the writer: the assignment, the file to save the
+  returned program as, and whether a `node main` entry point is required
+  (false when the deliverable is a module whose exports a harness imports).
+  The shape the `evals/agency-coding` suite uses as its input.
+
+```ts
+/** What an eval hands the writer: the assignment, the file to save the
+  returned program as, and whether a `node main` entry point is required
+  (false when the deliverable is a module whose exports a harness imports).
+  The shape the `evals/agency-coding` suite uses as its input. */
+export type CodingEvalInput = {
+  assignment: string;
+  outFile: string;
+  requireMain: boolean
+}
+```
+
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/agents/agency/coding.agency#L339))
 
 ## Functions
 
@@ -46,7 +67,7 @@ Return the Agency writer's tools: the bundled documentation, the source
 
 **Returns:** `any[]`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/agents/agency/coding.agency#L108))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/agents/agency/coding.agency#L109))
 
 ### agencyCodingAgent
 
@@ -101,4 +122,34 @@ Write an Agency program for the task. Iterates until the source parses,
 
 **Throws:** `std::guard`
 
-([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/agents/agency/coding.agency#L259))
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/agents/agency/coding.agency#L260))
+
+## Nodes
+
+### evalMain
+
+```ts
+evalMain(input: CodingEvalInput): string
+```
+
+Eval entry point: `agency eval run stdlib/agents/agency/coding.agency:evalMain
+  --suite <dir>`. A node in a library module never runs when the module is
+  imported; it only exists so a suite can score this writer without a
+  wrapper file. Any other writer scored on the same suite supplies a node
+  with this signature. The writer returns source without saving it, so
+  saving into the workdir — where the suite's harness pairs grade it — is
+  this node's own doing (`with approve`). A budget trip inside the writer is
+  rejected (`with reject`) so the caps stay caps. A failed run still saves
+  its last draft: the harness scores whatever partial credit it earns.
+
+**Parameters:**
+
+| Name | Type | Default |
+|---|---|---|
+| input | [CodingEvalInput](#codingevalinput) |  |
+
+**Returns:** `string`
+
+**Throws:** `std::write`, `std::guard`
+
+([source](https://github.com/egonSchiele/agency-lang/tree/main/packages/agency-lang/stdlib/agents/agency/coding.agency#L354))
--- a/packages/agency-lang/docs/site/stdlib/agents/agency/coding.md
+++ b/packages/agency-lang/docs/site/stdlib/agents/agency/coding.md
@@ -32,24 +32,25 @@ export type WriteFailure = {
 
 ### CodingEvalInput
 
-What an eval hands the writer: the assignment, the file to save the
-  returned program as, and whether a `node main` entry point is required
-  (false when the deliverable is a module whose exports a harness imports).
-  The shape the `evals/agency-coding` suite uses as its input.
+What an eval hands the writer: the assignment, and the file to save the
+  returned program as. The shape the `evals/agency-coding` suite uses as its
+  input. Deliberately agent-agnostic: everything about the deliverable's
+  shape (module vs runnable program, what to export) lives in the assignment
+  text, so any writer can be scored on the same suite.
 
 ```ts
-/** What an eval hands the writer: the assignment, the file to save the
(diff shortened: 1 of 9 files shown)


packages/agency-lang/evals/agency-coding/reverse-digits/test.json
```diff
@@ -0,0 +1,9 @@
+{
+  "description": "Module coding: digit manipulation that needs a while loop (Agency has no C-style for). Graded only by the hidden harness in holdout/.",
```
egonSchiele: > Graded only by the hidden harness in holdout

Why do you think it is important for the user to know this?

packages/agency-lang/evals/agency-coding/sum-multiples/test.json
```diff
@@ -0,0 +1,9 @@
+{
+  "description": "Module coding: a pure arithmetic function over a loop. Graded only by the hidden harness in holdout/, which imports the saved file.",
```
egonSchiele: Please go through the description line by line and help me understand why you think this line is useful to include in the description. When a user reads the description, what information do you think they're looking for?

packages/agency-lang/evals/agency-coding/sum-multiples/test.json
```diff
@@ -0,0 +1,9 @@
...
+  "tags": ["coding", "module", "easy"],
+  "input": {
+    "assignment": "Write an Agency module that exports def sumMultiples(n: number): number, returning the sum of all positive multiples of 3 or 5 that are strictly below n. A number that is a multiple of both counts once. For n <= 3 there are no such multiples, so the result is 0. Examples: sumMultiples(10) = 23 (3 + 5 + 6 + 9), sumMultiples(20) = 78. Use only Agency standard library (std::) imports, or no imports at all.",
+    "outFile": "solution.agency",
+    "requireMain": false
```
egonSchiele: Notice that the test you have written is fairly closely tied to the specific agent we have in the agency standard library. I don't think `requireMain` is a parameter most agents would have.

packages/agency-lang/evals/agency-coding/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Grading
+
+There are no `graders.ts` modules here. Each test carries a hidden harness
+pair in its `holdout/` directory — a `<name>.agency` importing the saved
+`outFile`, and a `<name>.test.json` naming node-by-node expected outputs.
+The framework discovers the pair and grades the run's working directory
+with `agency test --agency-only --reject '*'` (an `AgencyTestGrader`, one
+per pair, must-pass): the score is the passing fraction, so a program that
+gets some cases right earns partial credit, and the stdlib's `evalMain`
```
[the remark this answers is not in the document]
egonSchiele: Confirmed against scorecard.ts (a gate-failed input scores 0) and fixed in e609ec3c1: the README now documents grading as all-or-nothing — the must-pass gate stays, the grader's report row still shows the passing fraction and names each failing case, and that per-case feedback is the real reason evalMain saves a failed run's last draft.

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -330,3 +331,45 @@ export def agencyCodingAgent(
...
+
+/** Eval entry point: `agency eval run stdlib/agents/agency/coding.agency:evalMain
+  --suite <dir>`. A node in a library module never runs when the module is
+  imported; it only exists so a suite can score this writer without a
+  wrapper file. Any other writer scored on the same suite supplies a node
+  with this signature. The writer returns source without saving it, so
+  saving into the workdir — where the suite's harness pairs grade it — is
+  this node's own doing (`with approve`). A budget trip inside the writer is
+  rejected (`with reject`) so the caps stay caps. A failed run still saves
+  its last draft: the harness scores whatever partial credit it earns. */
```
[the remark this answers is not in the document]
egonSchiele: Fixed in e609ec3c1: the doc comment now says harness grading is all-or-nothing and the saved draft buys diagnostic feedback (the grade names exactly which cases the draft fails), not partial credit.

packages/agency-lang/docs/site/stdlib/agents/agency/coding.md
```diff
@@ -101,4 +122,34 @@ Write an Agency program for the task. Iterates until the source parses,
...
+
+Eval entry point: `agency eval run stdlib/agents/agency/coding.agency:evalMain
+  --suite <dir>`. A node in a library module never runs when the module is
+  imported; it only exists so a suite can score this writer without a
+  wrapper file. Any other writer scored on the same suite supplies a node
+  with this signature. The writer returns source without saving it, so
+  saving into the workdir — where the suite's harness pairs grade it — is
+  this node's own doing (`with approve`). A budget trip inside the writer is
+  rejected (`with reject`) so the caps stay caps. A failed run still saves
+  its last draft: the harness scores whatever partial credit it earns.
```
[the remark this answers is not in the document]
egonSchiele: Fixed in e609ec3c1: corrected the source doc comment in coding.agency and regenerated this page.

packages/agency-lang/evals/agency-coding/reverse-digits/test.json
egonSchiele: It is not important — that line described the suite's grading mechanics, which I had just built and was narrating, not something a reader of this test needs. Grading mechanics are the same for every test here, so they belong once in the README. Fixed in e609ec3c1: the description now states only what the test measures.

packages/agency-lang/evals/agency-coding/sum-multiples/test.json
egonSchiele: Going line by line through the old description ("Module coding: a pure arithmetic function over a loop. Graded only by the hidden harness in holdout/, which imports the saved file."):

- "Module coding:" — a category label. Weak: the tags field already carries this.
- "a pure arithmetic function over a loop" — the capability under test. This is the one useful part: the description prints under the score line in the grade report, and the reader there is asking "what skill did the agent just pass or fail?".
- "Graded only by the hidden harness in holdout/, which imports the saved file." — grading mechanics. Useful to someone authoring a new test, useless to someone reading grades, and identical for every test in the suite — so it belongs once in the README, not in each test.

Fixed in e609ec3c1: both descriptions now state only the capability the test measures ("Write and export one small pure function: sum the multiples of 3 or 5 below n.").

packages/agency-lang/evals/agency-coding/sum-multiples/test.json
egonSchiele: Fixed in e609ec3c1: requireMain is gone from CodingEvalInput — the contract is {assignment, outFile}, and the deliverable's shape (module vs program, what to export) lives in the assignment text where any agent can read it. requireMain stays on the agent side of the boundary: our evalMain passes requireMain: false to agencyCodingAgent, since the assignment defines the shape and the harness judges the outcome. Verified with a fresh end-to-end run (both tests bind and grade — 1.000/1.000 this time; the earlier missing-export failure did not recur, so that one is nondeterministic).

Issue #843, opened 2026-08-12
`:notes:` gets turned into a notes emoji in documentation


egonSchiele:
https://agency-lang.com/stdlib/notes/apple.html#std-create

egonSchiele:
Fixed in #905 (squash-merged as 30d8b4397).

VitePress enables markdown-it-emoji unconditionally, and the plugin replaces `:name:` anywhere in inline text with no word-boundary check. Agency namespaces with `::`, so `std::notes::create` contains the shortcode for a musical note and the heading rendered as `std:🎶:create`.

Disabled in `docs/site/.vitepress/config.mts`:

```ts
markdown: {
  config: (md) => {
    md.core.ruler.disable("emoji");
  },
},
```

Checked the blast radius both ways before disabling it site-wide. Six built pages contain emoji; five are literal emoji typed into the source (✅/❌ in code comments, ⚠️ in a heading, ✓/❯ in terminal output), and `apple.html` was the only page where one came from a shortcode. No source markdown uses a shortcode on purpose.

Verified by rebuilding the site: `apple.html` went from 6 emoji to 0, the heading reads `std::notes::create`, and the emoji count on all five intentional pages is unchanged.

Deliberately no second argument to `disable`. Passing `true` would tolerate the rule not being registered, so a VitePress upgrade that renamed it would make this a silent no-op and quietly bring the emoji back. Without it, that becomes a build failure — confirmed by pointing the call at a name that does not exist and watching the build fail with `invalid rule name`.

PR #905, opened 2026-08-24
agency doc: a @hidden tag, an export rule for nodes, and no more emoji in effect headings

### LINKED ISSUE #891: agency doc: a @hidden marker so eval entry nodes stay out of the stdlib reference

egonSchiele:
Fixes #891 and #843.

## `@hidden` (#891)

Write `@hidden` above a type, function, node, or constant and `agency doc` leaves it out:

```ts
/** What an eval hands the reviewer. */
@hidden
export type ReviewEvalInput = {
  assignment: string
  sourceFile: string
}
```

This is for something you have to export for a tool to reach but are not asking anyone to use. The review and coding agents each export an eval input type that only their eval suite ever passes in, and it was sitting in the reference right next to the agent.

**No language change was needed.** Bare `@hidden` already parses (`tagParser` makes the argument list optional), nothing validates tag names, `@optimize` is the existing precedent for a bare marker tag, and `attachTags()` already moves a tag onto exactly the declaration kinds `agency doc` renders. I checked before writing anything: a scratch file with `@hidden` on a type and a node parsed, typechecked clean, and came back byte-identical from `agency fmt`. So this is a visibility filter in two consumers, not a compiler feature.

**`describe()` applies the same rule.** `std::agency`'s `describe()` says in its own doc comment that it applies "the `agency doc` visibility rule", and it reports exported types — so without this it would keep reporting `ReviewEvalInput` after `agency doc` learned to hide it. Two tools claiming one rule would disagree about these exact two modules.

**Hidden declarations also leave the symbol registry.** A hidden type renders no `### Name` section, so a link to it from another page would point at an anchor that does not exist — a silently broken link, worse than the noise being removed. Only the type-alias case can break a link today, but the registry is defined as "the set link targets resolve against", so it stays honest about all four kinds. No render-key bump is needed: adding the tag changes the file's hash, the page re-renders with a smaller symbol set, and the existing `linkTargets` re-check invalidates the pages whose lookups changed.

**`@hidden` on an `effect` block warns.** `attachTags` does not reach effect declarations, so the tag would be left as a loose node and do nothing at all. `agency doc` now prints a line naming the file and the line number instead of saying nothing.

## Nodes now follow the export rule

Functions, types, and constants have always been filtered on `export`. Nodes were not — which is why a non-exported `evalMain` was rendered at all. Across the whole standard library this removes exactly the two eval entry nodes and nothing else: they were the only two nodes documented anywhere in it.

## A stray empty heading, fixed along the way

Three sections checked emptiness *before* filtering, so a module whose declarations were all filtered out emitted a bare `## Types` heading with nothing under it. That is why 15 stdlib pages in this diff lose an empty `## Types`.

## The `:notes:` emoji (#843)

`std::notes::create` rendered as `std:🎶:create`. VitePress enables markdown-it-emoji unconditionally, the plugin replaces `:name:` anywhere in inline text with no word-boundary check, and Agency namespaces with `::` — so `std::notes::create` contains the shortcode for a musical note. Disabled in `docs/site/.vitepress/config.mts`.

I checked the blast radius both ways before choosing a site-wide disable. Six built pages contain emoji; five are literal emoji typed into the source (`✅`/`❌` in code comments, `⚠️` in a heading, `✓`/`❯` in terminal-output fences), and `apple.html` was the only page where one came from a shortcode. No source markdown anywhere uses a shortcode on purpose.

I rejected the alternative — making `agency doc` emit `` ### `std::notes::create` `` so inline code shields the text — because it protects only generated pages, leaves hand-written guide prose exposed to the same trap, and changes every effect heading's anchor slug.

## Verification

Doc and `describe()` tests, before and after the change:

| | before | after |
|---|---|---|
| 5 new `agency doc` tests | 5 failed | pass |
| 1 new `describe()` test | 1 failed | pass |
| `doc*.test.ts` + `agency.test.ts` + `sourceIsText.test.ts` | — | 198 passed |

Each negative assertion is paired with a positive one on a plain sibling, so a filter that dropped everything could not pass.

Six existing doc tests asserted on a `node main()` being rendered. They now write `export node main()` — the assertion is that a node renders, and that is still what they check.

The emoji fix, by rebuilding the site and counting emoji per page:

| page | before | after |
|---|---|---|
| `stdlib/notes/apple.html` | 6 | **0** |
| `guide/ts-helpers.html` | 6 | 6 |
| `guide/unused.html` | 2 | 2 |
| `cli/coverage.html` | 2 | 2 |
| `guide/using-local-models.html` | 1 | 1 |
| `agent/index.html` | 1 | 1 |
| `guide/effects-and-raises.html` | 1 | 1 |

The heading now reads `std::notes::create`.

`pnpm run lint:structure` and `pnpm run fmt:ts` are clean.

## Docs

`docs/site/cli/doc.md` gets a "What gets documented" section (the old text said "all nodes, functions, and types", which was never quite true and is now less true) and a "Hiding a declaration" section. `docs/dev/doc-cache.md` gets the registry interaction and why it needs no render-key bump, with the pointer in CLAUDE.md updated.

https://claude.ai/code/session_01AWhxQA8fsh3XVp7fE8BLPS


packages/agency-lang/docs/site/.vitepress/config.mts
```diff
@@ -3,6 +3,17 @@ import { defineConfig } from "vitepress";
 export default defineConfig({
   title: "Agency",
   description: "A language for creating agents.",
+  markdown: {
+    config: (md) => {
+      // Agency namespaces with `::`, so a heading like `std::notes::create`
```
egonSchiele: Please stop adding such long comments. They hurt readability, as users are unlikely to read them. If you feel the need to add such a long comment, ask yourself why the code you have written does not make things clear enough. Ask yourself whether a shorter comment would work as well.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -446,6 +453,21 @@ function generateDocForFile(
   return generatedOutput;
 }
 
+/** `attachTags` moves a tag onto the next declaration, but only for the
+ *  kinds it knows: functions, nodes, type aliases, constants, and calls.
+ *  A `@hidden` above anything else — an `effect` block, say — is left
```
egonSchiele: How much work would it take to make tags work with effect blocks as well? What are the other things that tags don't work with right now that show up in the `agency doc` output?

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -285,6 +292,10 @@ function localExportInfos(
   generator: AgencyGenerator,
   effects: Record<string, string[]>,
 ): ExportInfo[] {
+  // A deliberate omission, unlike the loud throw at the bottom of this
+  // function: that guard is about declaration KINDS this dispatch does not
+  // know, not about declarations the author asked to keep off the surface.
+  if (isHidden((node as { tags?: Tag[] }).tags)) return [];
```
egonSchiele, line 320: **`@hidden` leaks back through a named re-export.** `resolveNamedReExports` (below, in this same file) does `const base = found ?? thinReExport(sourceName, localName, from)`. With `@hidden export type Secret` in module A, `describeSource(A)` now drops `Secret`, so a module B that writes `export { Secret } from "std::a"` gets `found === null` and falls through to `thinReExport` — reporting `Secret` after all, and reporting it as a bogus `kind: "reexport"` / `effects: ["unknown"]` entry rather than its real shape.

Note that `export * from "std::a"` takes the `inner.exports.map(...)` path and *does* hide it correctly, so the same tag gives opposite answers depending on which re-export form is used. The existing `isInternalExport(localName)` filter at the top of `resolveNamedReExports` is exactly the guard hidden names are missing.

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -680,8 +703,12 @@ function generateNodeSection(
   ctx: DocContext,
   interruptEffectsByFunction: Record<string, InterruptEffect[]>,
 ): string | null {
-  if (nodes.length === 0) return null;
-  const parts = nodes.map((node) => {
+  // A node that is not exported cannot be reached by anyone importing the
+  // module, so it is not part of the module's surface — the same rule the
+  // function, type, and constant sections have always applied.
+  const visible = nodes.filter((node) => node.exported && !isHidden(node.tags));
```
egonSchiele: **Non-exported nodes now render no section but stay in the symbol registry.** This is the same broken-anchor argument the PR makes for `@hidden`, applied one line too narrowly: `extractRegistrySymbols` pushes every `info.graphNodes` entry regardless of `exported` (it only filters `isHidden`), while this section now renders only exported ones.

Concrete failure: file `x.agency` has a non-exported `node Foo`, and `y.agency` has `export type Foo`. Pass 1 walks in traversal order, so `symbolRegistry["Foo"]` can end up `x.md`; `formatTypeLinked` then emits `[Foo](x.md#foo)` for any signature mentioning the type — and `x.md` no longer contains a `### Foo`. Before this change the node rendered, so that anchor existed.

Adding `node.exported` (and, for parity, the `exported` check the type-alias loop is also missing) to `extractRegistrySymbols` keeps the registry equal to "what actually renders a section".

packages/agency-lang/docs/site/cli/doc.md
```diff
@@ -26,8 +26,13 @@ You can also set these options in the Agency config file:
 
 `baseUrl` is used to generate links to the source code in the documentation. It should point to the directory containing your library's source code.
 
+## What gets documented
+Agency documents the exported nodes, functions, types, and constants defined at the top level of each file. Anything you do not export is not part of what a caller can reach, so it is left out. Names starting with an underscore are left out too: they are compiler plumbing rather than something anyone calls.
```
egonSchiele: **The underscore claim is only true of functions.** `generateFunctionSection` is the only section with a `declaredName(...).startsWith("_")` check; `generateTypeSection`, `generateNodeSection`, and `generateConstantSection` filter on `exported` and `@hidden` only. So `export type _Internal = ...` or `export const _SEED = 1` still renders a section, contradicting this sentence.

(`std::agency`'s `describe()` *does* apply `isInternalExport` to all four kinds, so this is also a place where the two tools disagree about the surface — the exact disagreement the rest of this PR is closing.)

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -395,6 +400,8 @@ function generateDocForFile(
     // Fall back to no interrupt info if the type checker crashes.
   }
 
+  warnOnStrayHiddenTags(program, filePath);
```
egonSchiele: **The stray-tag warning disappears on the second run.** `generateDocForFile` is only called for pages the doc cache considers stale (pass 2 in `generateDocForDirectory` skips fresh entries entirely), and nothing about the warning is recorded in the ledger. So an author who writes `@hidden` above an `effect` block sees the message once, ignores it, reruns `agency doc`, and gets a clean run with the tag still doing nothing — which is the silent failure this function exists to prevent.

Cheapest fix that keeps the property: warn during pass 1 (it already parses every stale file) or fold a "had stray hidden tags" boolean into the ledger entry and re-emit it for fresh pages.

packages/agency-lang/lib/utils/hiddenTag.ts
```diff
@@ -0,0 +1,22 @@
+import type { Tag } from "../types/tag.js";
+
+/** The tag name, shared so the check and the "you wrote it somewhere it
+ *  cannot attach" warning cannot drift apart. */
+export const HIDDEN_TAG = "hidden";
+
+/**
```
egonSchiele: Another extremely long comment

packages/agency-lang/docs/site/.vitepress/config.mts
egonSchiele: Cut to two lines:

```ts
// `std::notes::create` contains `:notes:`, which the emoji plugin
// VitePress always enables rewrites to 🎶 (issue 843).
md.core.ruler.disable("emoji", true);
```

I went through the rest of the diff on the same principle and shortened every comment I had added.

packages/agency-lang/lib/utils/hiddenTag.ts
egonSchiele: Down to two lines. Dropped the "why it exists" paragraph and the note about needing `attachTags()` first — the second is now enforced by the code rather than described: `isDocumented` is the only caller path, and the registry test parses *and* preprocesses so a fixture that skipped it would fail rather than silently pass.

packages/agency-lang/lib/stdlib/agency.ts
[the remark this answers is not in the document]
egonSchiele: Both done. `describe()`\s docstring now says it omits `@hidden` and underscore-prefixed names, and `docs/site/stdlib/agency.md` is regenerated.

Your second point was sharper than it looks — chasing it turned up a real bug, which egon also caught separately: a named re-export of a hidden name did not just under-report, it *over*-reported, resurfacing the name as a bogus `kind: "reexport"` / `effects: ["unknown"]` entry. Fixed, with a test covering both re-export forms.

packages/agency-lang/lib/cli/doc.ts
egonSchiele: Small — done, four edits:

1. `tags?: Tag[]` on `EffectDeclaration`.
2. `"effectDeclaration"` added to `attachTags`'s target list.
3. `processEffectDeclaration` prints `formatAttachedTags(node)`. **This one is load-bearing**: without it `agency fmt` would have silently *deleted* the tag, so making tags attach without touching the generator would have introduced a data-loss bug. Verified the round-trip by hand.
4. `generateEffectSection` filters `@hidden`.

To your second question — effect declarations were the only gap. `agency doc` renders five things: type aliases, effect declarations, constants, functions, and nodes. `attachTags` already reached the other four. `effectSet` declarations look like a sixth but are not: they parse as `typeAlias` with `isEffectSet`, so they were covered already.

So the stray-tag warning now only fires for positions where no tag of any kind means anything — a trailing `@hidden` at the end of a file, say. I kept it, and fixed the cache hole you flagged separately.

packages/agency-lang/lib/stdlib/agency.ts
egonSchiele: Confirmed and fixed. Reproduced it first against the real stdlib:

```
named  : [["CodingEvalInput","reexport",["unknown"]]]
star   : ["WriteFailure","buildTools","agencyCodingAgent"]
```

Exactly as you described — and the star form does hide it, so the tag gave opposite answers by re-export form.

`describeSource` now returns `{ info, hiddenNames }`, and `resolveNamedReExports` takes the hidden list and returns `[]` for those names. That is the distinction the code was missing: `found === null` conflates "hidden there" with "not exported there", and only the first should vanish.

The test asserts both forms and also asserts a real export from the same module still comes through, so an empty result cannot pass it. It fails without the fix.

packages/agency-lang/lib/cli/doc.ts
egonSchiele: You were right, and it was worse than a hypothetical — **this was already live on the site.**

Chasing your example I checked every same-page anchor under `docs/site/stdlib` against `main`. Five links were already broken there, all the type-alias half you flagged: `wikipedia.md` links `[WikiSummary](#wikisummary)` and contains zero `### WikiSummary`. Same for `StatInfo` in `shell.md`, `PlanState` and `Todo` in `planner.md` and `agent.md`. Those types are non-exported, so the section never rendered, but the registry kept the name. My round-1 change would have added the node version of the same bug.

Rather than add two `exported` checks, I made the invariant structural: one `isDocumented(name, exported, tags)` predicate, called by `extractRegistrySymbols` and by all four render sections. The registry is now *defined* as the set of rendered anchors, so a future visibility rule added to a section but not the registry cannot drift.

All five links now render as plain code, and a script over the whole stdlib output confirms every same-page anchor resolves. The new test needed care to be real: with the files named `x`/`y` it passed without the fix, because last-writer-wins in traversal order let the documented file win the name anyway. Renamed so the non-exported node sorts last, and it now fails on the exact `z.md#foo` assertion without the change.

packages/agency-lang/lib/cli/doc.ts
egonSchiele: Confirmed — and pass 1 would not have fixed it either, since fresh files are not parsed there either; they contribute cached `registrySymbols`. So the ledger was the only option of the two.

`strayHiddenLines: number[]` on the entry, written in pass 2, re-emitted for pages served from cache. Optional in the type and tolerated by `ledgerEntryHasValidShape`, so an existing ledger keeps its authority and simply warns nothing until the page next re-renders.

The test runs `generateDoc` over a directory twice with nothing touched in between and asserts the warning both times. It fails on the second run without the fix.

I noted the general rule in `docs/dev/doc-cache.md` under "Warnings have to survive a cache hit" — this applies to any per-page diagnostic added later, not just this one.

packages/agency-lang/docs/site/cli/doc.md
egonSchiele: Fixed by making the code true rather than narrowing the sentence — you and Copilot both landed on this, and your parenthetical is the reason: `describe()` already applied `isInternalExport` to all four kinds, so narrowing the doc would have documented a disagreement this PR exists to close.

The underscore rule now lives in the shared `isDocumented` predicate, so all four sections and the registry apply it. Zero stdlib pages change: nothing exports an underscore-prefixed type, const, or node today. There is a test with one of each.

`doc.md` also now says effect declarations are documented and that `@hidden` works on them.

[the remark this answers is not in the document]
egonSchiele: Fixed in the code rather than the sentence — see the thread egon opened on the same line. The underscore rule now applies to all four declaration kinds through a shared predicate, which is what `describe()` already did, so the sentence is now accurate as written.

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -374,27 +402,31 @@ function reExportInfos(node: ExportFromStatement, visited: string[]): ExportInfo
...
       const inner = describeSource(readFileSync(abs, "utf-8"), [...visited, abs]);
       if (node.body.kind === "starExport") {
-        return inner.exports.map((info) => ({ ...info, reexportedFrom: from }));
+        return inner.info.exports.map((info) => ({ ...info, reexportedFrom: from }));
       }
       const byName: Record<string, ExportInfo> = Object.create(null);
-      for (const info of inner.exports) byName[info.name] = info;
-      return resolveNamedReExports(node.body, from, byName);
+      for (const info of inner.info.exports) byName[info.name] = info;
+      return resolveNamedReExports(node.body, from, byName, inner.hiddenNames);
```
egonSchiele: **`hiddenNames` only covers one hop, so a two-hop named re-export still reports a hidden name.**

`describeSource` fills `hiddenNames` from *this* module's own top-level nodes (`exportedNamesOf`), and nothing merges in the hidden names of the modules it re-exports from. So the fix holds for `export { X } from A` but not for a chain:

- `A` declares `@hidden export type X`.
- `B` does `export * from "A"` -- `B.info.exports` correctly lacks `X`, and `B.hiddenNames` is `[]` because `B` has no local hidden declaration.
- `C` does `export { X } from "B"` -- `byName` has no `X`, `hiddenNames` is empty, so it falls through to `thinReExport` and `describe(C)` reports exactly the bogus `kind: "reexport"` / `effects: ["unknown"]` entry this change was meant to prevent.

Propagating the inner result's hidden names (plus the star-export case's) through `DescribeResult` would close it.

packages/agency-lang/docs/site/.vitepress/config.mts
```diff
@@ -3,6 +3,13 @@ import { defineConfig } from "vitepress";
 export default defineConfig({
   title: "Agency",
   description: "A language for creating agents.",
+  markdown: {
+    config: (md) => {
+      // `std::notes::create` contains `:notes:`, which the emoji plugin
+      // VitePress always enables rewrites to 🎶 (issue 843).
+      md.core.ruler.disable("emoji", true);
```
egonSchiele: **The `silent` argument makes this fail open.**

`ruler.disable(name, silent)` -- passing `true` for the second argument tells markdown-it not to throw when no rule by that name is registered. The whole point of this line is that a rule *must not* run, so the one failure mode worth catching is "the rule is no longer called `emoji`". If a VitePress or markdown-it-emoji upgrade renames or restructures the core rule, this call becomes a silent no-op, the site builds green, and `std::notes::create` quietly goes back to rendering as `std:\ud83c\udfb6:create`.

Dropping the `true` turns that into a loud build failure at the moment of the upgrade, which is the signal you want.

```suggestion
      md.core.ruler.disable("emoji");
```

packages/agency-lang/lib/cli/doc.ts
```diff
@@ -446,6 +466,21 @@ function generateDocForFile(
   return generatedOutput;
 }
 
+/** Source lines of a `@hidden` that attached to no declaration — at the
+ *  end of a file, say. Warning is the only way the author finds out. */
+export function strayHiddenLines(program: AgencyProgram): number[] {
+  return program.nodes
+    .filter((node) => node.type === "tag" && node.name === HIDDEN_TAG)
+    .map((node) => (node.loc ? node.loc.line + 1 : 0));
+}
```
egonSchiele, line 477: **The stray-tag warning only scans top-level nodes, but `collectTags` also strands tags inside bodies.**

`docs/site/cli/doc.md` in this PR promises: "If you write it somewhere it cannot apply ... `agency doc` warns rather than quietly ignoring it." This scan is `program.nodes` only, while `collectTags` recurses into `graphNode` / `function` bodies and into control-flow bodies via `walkBody`, pushing unattachable tags back as standalone nodes there.

Concretely, a `@hidden` written above a statement that is not an attach target inside a function body -- e.g. above a `return`, an `if`, or a `while` -- is dropped with no warning at all, which is the exact silence the doc says will not happen. Walking the nested bodies (or narrowing the doc's claim to top level) would keep the two honest.

packages/agency-lang/lib/stdlib/agency.ts
egonSchiele: Right, the fix stopped one hop short. `reExportInfos` now returns `{ infos, hiddenNames }`: a star passes on the inner module's whole hidden set, a named list passes on the re-exported subset **under the local alias** — `export { X as Y } from A` has to report `Y`, since `Y` is what the next module out asks for. `describeSource` merges those into its own.

One honest limitation: I could not build an end-to-end test for it. `describe()` only resolves `std::` re-exports, `getStdlibDir()` is fixed at the package root, and no stdlib module re-exports from one carrying a `@hidden` — so the two-hop chain is not reachable without adding a hidden declaration to a real stdlib module for a test's sake, which I did not want to do. Instead `reExportedHiddenNames` is exported and unit-tested directly, including the alias case, and it fails without the fix. The one-hop integration test still covers the wiring. Flagging the gap rather than implying the chain is covered end to end.

packages/agency-lang/docs/site/.vitepress/config.mts
egonSchiele: Agreed, and applied — I had the reasoning backwards. I added `true` for robustness against an upgrade, but that is precisely the case where I want to hear about it, and a green build with the emoji back is the worst outcome.

Verified both directions rather than just deleting the argument:

- Build with `disable("emoji")`: exits 0, `apple.html` contains 0 emoji.
- Build with the name changed to something unregistered: exits 1 with `Error: Rules manager: invalid rule name emojiRenamed`.

So the failure mode you describe is now a build break, and I have seen it break.

packages/agency-lang/lib/cli/doc.ts
egonSchiele: Confirmed — `@hidden` above a `return` produced no warning at all, and the AST shows the tag sitting right there in the function body.

Rather than write a second walker that would drift from `collectTags`, I put `strandedTags` beside it in the preprocessor and gave it the same shape: explicit recursion into function/node bodies, then `walkBody` for control flow. Same file, same walk, read together.

The test covers the body case and fails without it. My first version of the test asserted the wrong line number, which is how I found that the warning was in fact correct — the arithmetic in my assertion was off, not the code.

I noted the general rule in `docs/dev/doc-cache.md`: a separate walker here drifts, and the drift shows up as *missing* warnings, which nothing tests for.

PR #906, opened 2026-08-24
Surface aborts that reach the node boundary as a value (#243)

### LINKED ISSUE #243: Add default maxCallDepth to catch runaway async recursion before OOM

egonSchiele:
Closes #243.

The `maxCallDepth` guard from #243 has been on main since `b2ea8c30f`, and it works — it stops the runaway in about 3 seconds instead of letting the process OOM. But running the issue's own repro showed the user still gets nothing useful out of it. Two separate holes, both on the path where an abort travels up through compiled code as a value.

### An abort in tail position escaped as the node's return value

```agency
def foo() { return bar() }
def bar() { return foo() }

node main() { return foo() }
```

Before: **exit 0, no output** — infinite recursion reported as a successful run.

Codegen emits an `if (isAborted(x)) throw x.toError()` check after each call, but only when the result is bound to a local. A bare `return foo()` binds nothing, gets no check, and the `AbortedResult` flows straight out as the node's result. Write it as `const r = foo()` instead and you got an error; the issue's form silently succeeded.

The fix converts in `runNode`, which is where `AbortedResult`'s own docstring already says the conversion belongs: *"Above node level (the graph engine, the CLI entry) aborts are converted back to exceptions."* One boundary covers the tail-position shape and any future one, rather than teaching codegen about one more case.

**The ordering here is load-bearing and easy to get wrong.** My first attempt put the check right after `createReturnObject`, next to the existing `hasInterrupts` check, and it never fired once — `createReturnObject` JSON round-trips the value, and `isAborted` is an `instanceof` test, so the prototype is gone by then. It has to run on the live result. I only caught this because the end-to-end repro still exited 0; the unit tests were perfectly happy.

### The diagnostic was being thrown away

Even on the path that did surface, you saw:

```
Agent crashed: Execution aborted (callDepthExceeded)
```

`CallDepthExceededError` builds a good message — frame names, the numbers, a pointer to the config knob — but `AbortedResult.toError()` rebuilds the exception from the `AbortCause` alone, and the original error object is long gone by then. The cause only carried `{kind, limit, observed}`.

The frames now ride the cause, and `describeAbortCause` owns the text, following the `guardTrip` branch already there. `CallDepthExceededError` renders its own message through the same function, so the thrown form and the rebuilt form cannot drift apart. After:

```
Agent crashed: Maximum call depth exceeded (2049 > 2048). This usually means
unbounded recursion. Recent calls: bar → foo → bar → foo → bar → foo → bar → foo.
If your program legitimately recurses this deeply, raise the limit via the
`maxCallDepth` config option.
```

### On "should a tail-position abort throw rather than return?"

No, and I'd push back on framing it as a codegen question. Returning the `AbortedResult` is correct and load-bearing — it is how a saved draft gets salvaged, and changing it would break `saveDraft`. Nothing about how aborts travel inside compiled code changes here. The bug was only that the boundary failed to do its documented job, so that is the single place I touched.

### Checks

Every claim above was run, not reasoned about:

- **Both new tests fail without the fix.** For `nodeAbortBoundary.test.ts` this needs care: the compiled fixture imports `agency-lang/runtime`, which resolves to `dist`, so reverting the source alone leaves the test green. I reverted the source, rebuilt, and confirmed both tests fail with `promise resolved "{ messages: ... }" instead of rejecting` — the exact symptom from the issue — then restored and rebuilt. The `abortedResult` round-trip test likewise fails on `expected ... to contain 'foo → bar'` when the frames come off the cause.
- **Blast radius.** This changes behavior for every abort kind at the node boundary, not just call depth, so I ran the abort-adjacent suites: `abortedResult`, `callDepth`, `errors`, `guard`, `guardScope`, `runBatch`, `failurePropagation`, `saveDraftTool`, `invocationOutcome`, `invocationOptions`, `budgetExit`, `agencyFunction`, `functionBudget`, and the `node*` tests. 306 tests, all passing.
- `pnpm run fmt:ts`, `pnpm run lint:structure`, `tsc --noEmit`, and `lib/sourceIsText.test.ts` all clean.

I did not run the full suite locally, per the repo's guidance — CI will.

### Left alone, on purpose

The raw Node stack trace dumped after the `Agent crashed:` line is ugly, but it is pre-existing CLI behavior for every abort and unchanged by this PR. Worth a separate look if it bothers you.

The known limitation that stdlib `map`/`filter`/`reduce` burn ~2 depth levels per user level is also untouched — still documented in `callDepth.ts` and `docs/dev/config.md`.

https://claude.ai/code/session_01JG8vWxnHgaPmosYnkBc4jx


packages/agency-lang/lib/runtime/node.ts
```diff
@@ -473,6 +474,17 @@ async function runNodeCore({
             ),
         );
         await execCtx.pendingPromises.awaitAll();
+
+        // Node level is where an abort stops being a value and becomes an
+        // exception again. Codegen's per-call guards miss a tail-position
+        // `return foo()`, which binds no local (#243). Must run before
+        // createReturnObject: that JSON round-trip strips the prototype
+        // `isAborted` tests for. See docs/dev/saveDraft.md.
+        const abortedData = (result as { data?: unknown }).data;
```
egonSchiele: The boundary is fixed here, but the two sibling paths that end a run the same way still have the hole, so the issue's symptom survives a single interrupt.

`runResumeLoop` (lib/runtime/interrupts.ts:716) and the rewind loop (lib/runtime/rewind.ts:84) both do `await pendingPromises.awaitAll()` then `createReturnObject(...)` with no `isAborted` check. Concrete scenario: `node main() { let go = input("ok?"); return foo() }` with the same mutually recursive `foo`/`bar`. The first pass surfaces the interrupt; the CLI resumes through `resolveCliInterrupts` -> `respondToInterrupts` -> `runResumeLoop`, the tail-position abort comes back as `result.data`, `hasInterrupts` is false, so `agentEnd` records the `AbortedResult` as a successful result and the process exits 0 with no useful output -- exactly what this PR fixes for the fresh-run path.

Since the argument for putting the check at the boundary is "one place covers every shape", it should probably be a small shared helper called from all three sites (or from `createReturnObject`'s callers) rather than just this one.

packages/agency-lang/lib/runtime/node.ts
```diff
@@ -473,6 +474,17 @@ async function runNodeCore({
...
+        if (isAborted(abortedData)) {
+          throw abortedData.toError();
```
egonSchiele: `toError()` is called raw here, unlike the other value-to-exception boundary. runBatch does `value.atForkBoundary().toError()` (lib/runtime/runBatch.ts:387 and :915) precisely so the carried partial is dropped with an `abortSalvage` event and the `abortUnwind` span is ended.

Scenario: a def with `saveDraft` is aborted, so `AbortedResult.logged()` has opened an `abortUnwind` span and the instance carries a partial. It reaches the node boundary in tail position and this line throws. Nothing calls `deliver()` or `dropped()`, so no closing `abortSalvage` event is emitted and the span is never ended -- the statelog timeline shows an unterminated `abortUnwind` span for every draft-bearing abort that reaches node level, and the salvaged draft vanishes with no record of where it was dropped. Closing it explicitly (a `dropped`-style hop, or at minimum `endSpan`) before throwing would match the fork boundary's behavior.

packages/agency-lang/lib/runtime/node.ts
```diff
@@ -473,6 +474,17 @@ async function runNodeCore({
...
+        }
```
egonSchiele: Throwing here jumps over the normal end-of-run tail, and two of the things in it are user-visible.

Before this change an aborted tail-position result took the `else` branch below: `agentEnd` with the result, the `onAgentEnd` hook, and `closeTraceWriter()`. Now it lands in the outer `catch`, which emits `agentEnd` without a result and then goes to `finally` / `finalizeExecCtx` -- neither of which touches the trace writer. So for `agency run --trace` on a program that aborts (runaway recursion, root budget trip, cancellation), `TraceWriter.close()` never runs and whatever it flushes on close is lost or truncated; and a user's `onAgentEnd` callback no longer fires for these runs.

It is the same shape as any other crash path, so it may be acceptable -- but it is a behavior change this PR introduces for aborts specifically, and it is worth either calling out or handling (e.g. `await execCtx.closeTraceWriter()` before the throw).

packages/agency-lang/lib/runtime/node.ts
egonSchiele: Confirmed and fixed in 670d2629e. Both sites have the identical shape — `graph.run` → `awaitAll()` → `createReturnObject` with no check — so the escape survived a single interrupt exactly as you describe.

Took the shared-helper suggestion: `throwIfNodeResultAborted` in the new `lib/runtime/abortBoundary.ts`, called from all three. It could not go inside `createReturnObject` for the same reason the check has to precede it — the JSON round-trip strips the prototype — and because only the caller knows whether it owns the end of the run, which matters for the trace writer (see the other thread). Rewind passes `endsRun: false`; it has no end-of-run tail of its own.

Your scenario is now a test: `nodeAbortBoundary.test.ts` raises an interrupt, resumes through `respondToInterrupts`, and expects a rejection. I checked it fails for the right reason — reverting only the `runResumeLoop` line and rebuilding fails that one test with `promise resolved ... instead of rejecting`, leaving the other two green.

packages/agency-lang/lib/runtime/node.ts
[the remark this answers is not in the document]
egonSchiele: Fixed in 670d2629e — same finding as the human review thread above. Both boundaries now go through a shared `throwIfNodeResultAborted` helper (`lib/runtime/abortBoundary.ts`), called from `runNode`, `runResumeLoop` and the rewind loop, with a regression test covering the resume path.

packages/agency-lang/lib/runtime/node.ts
egonSchiele: Right, and fixed in 670d2629e. Added `atNodeBoundary()` alongside `atForkBoundary()`, so the partial is dropped through the same `dropped()` hop with a `droppedAtNodeBoundary` action and the `abortUnwind` span is ended before the throw.

Chasing this turned up a second leak in the same method: `dropped()` bailed early whenever `partial === undefined`, but an "erased" hop opens the span while carrying no partial forward, so a span can outlive the partial that started it. That early return left it open — at the fork and argument-position boundaries too, not just here. The guard is now "neither a partial nor a span", and the event's `partial` field is optional. Existing fork and arg-position tests still pass.

Pinned by a unit test asserting the event sequence is `[carried, droppedAtNodeBoundary]` and the span is ended; I had to make the test stub's `endSpan` actually record its argument, since it was a no-op.

packages/agency-lang/lib/runtime/node.ts
egonSchiele: Verified and handled rather than just called out — `finalizeExecCtx` really does not touch the trace writer, so the trace was being truncated.

I reproduced it directly: `agency run --trace` on the runaway-recursion program wrote a trace whose last record was `manifest`, with no `footer`. With `closeTraceWriter()` before the throw it ends in `footer`, same as a successful run. That is the falsification for this one — removing just the close and rebuilding puts it back to `manifest`.

The close is behind the `endsRun` flag so it only fires at the two sites that own the end of a run. I did **not** restore the `onAgentEnd` hook for aborted runs, and want to flag that as a deliberate call: aborts thrown from codegen (the bound-local shape) already skipped it before this PR, so firing it only for the tail-position shape would make the two inconsistent, and it would be handing user code an `AbortedResult` as a "result". An aborted run is now crash-shaped across the board. Say the word if you would rather it fired for every abort and I will do that separately, since it changes the bound-local path too.

packages/agency-lang/lib/runtime/node.ts
```diff
@@ -473,6 +474,9 @@ async function runNodeCore({
             ),
         );
         await execCtx.pendingPromises.awaitAll();
+
+        await throwIfNodeResultAborted(result, execCtx, { endsRun: true });
```
egonSchiele, line 483: The fourth boundary above compiled code is still unfixed: `runExportedFunctionCore` (this file, ~line 314) returns `fn.invoke(...)`'s value straight through as `outcome = { status: "returned", value }`. That value is an `AbortedResult` for exactly the same reason a tail-position `return foo()` produced one here — the function's own frame converted the abort to a value and nothing above it converts back.

Concrete: an exported function that trips `maxCallDepth` (or is cancelled) and is invoked through `__invokeFunction` / `__invokeFunctionForServe` hands the caller a JSON-ish `{ __type: "abortedResult", ... }` object instead of throwing. Over `./serve` that is an HTTP 200 with a nonsense body; from TypeScript (agency-js) it is a silently successful call. Same class of bug as #243, and the docstring you added ("every boundary above compiled code must call this first") does not hold until this one calls it too.

packages/agency-lang/lib/runtime/abortedResult.ts
```diff
@@ -186,15 +194,20 @@ export class AbortedResult {
...
    *  story ends where it is dropped). */
-  private dropped(action: "droppedAtArgPosition" | "clearedAtFork"): AbortedResult {
-    if (this.partial === undefined) {
+  private dropped(
+    action: "droppedAtArgPosition" | "clearedAtFork" | "droppedAtNodeBoundary",
+  ): AbortedResult {
+    // An "erased" hop opens the span while carrying no partial forward, so a
+    // span can outlive the partial that started it. Bail only when there is
+    // neither — otherwise that span would never be ended.
+    if (this.partial === undefined && this.unwindSpanId === undefined) {
```
egonSchiele: This widened bail changes behavior on a path where the abort is NOT ending: `droppedAtArgPosition` (lib/runtime/call.ts:87) drops the partial but lets the abort continue upward as this call's result.

Scenario: `g()` saves a draft (opens span S, "carried") → an intermediate frame with no draft of its own carries it through ("erased": partial now undefined, S retained) → that value lands in argument position. Before, this returned `this` and kept S open; now it emits `droppedAtArgPosition` with `partial: undefined`, ends S, and returns an instance with `unwindSpanId` cleared. If any frame further up has its own saved draft, `logged()` sees no span id and opens a **second** `abortUnwind` span, so one abort's salvage trail is split across two spans. Ending the span at the two terminal drops (fork / node boundary) is right; doing it at arg position is not.

packages/agency-lang/lib/statelogClient.ts
```diff
@@ -1472,7 +1472,13 @@ export class StatelogClient {
...
     partial,
   }: {
-    action: "carried" | "erased" | "delivered" | "clearedAtFork" | "droppedAtArgPosition";
+    action:
+      | "carried"
+      | "erased"
+      | "delivered"
+      | "clearedAtFork"
+      | "droppedAtArgPosition"
+      | "droppedAtNodeBoundary";
```
egonSchiele, line 1482: The new action is added to the client's union but not to the two places that document the closed set: the docstring right above this method still says "(clearedAtFork, droppedAtArgPosition)", and `docs/dev/statelog.md:189` lists `carried | erased | delivered | clearedAtFork | droppedAtArgPosition` as the full `action` enum. Anyone reading either one to build a consumer will not handle `droppedAtNodeBoundary`.

packages/agency-lang/lib/runtime/errors.ts
```diff
@@ -162,6 +171,15 @@ export function describeAbortCause(cause: AbortCause): string {
     const name = cause.label !== undefined ? `Guard "${cause.label}"` : "Guard";
     return `${name} exceeded its ${cause.dimension} budget: spent ${cause.spent} of limit ${cause.limit}`;
   }
+  if (cause.kind === "callDepthExceeded") {
```
egonSchiele, line 174: `describeAbortCause` is now the sole owner of the message for *every* abort that reaches the node boundary, and it still drops the one field the other kinds carry: `userKill`'s `reason`.

Concrete: a run cancelled with `ctx.cancel("spend limit reached")` (or an external `abortSignal`) whose node aborts in tail position now surfaces as `Agent crashed: Execution aborted (userKill)` — the reason string the caller supplied is on the cause and is thrown away. Pre-PR that path exited 0 silently, so this is not a regression, but since the PR's stated goal is a readable diagnostic at this boundary, one more line here (`cause.reason !== undefined ? ... : ...`) gets it for free.

packages/agency-lang/lib/runtime/node.ts
egonSchiele: Confirmed and fixed in 3f74c4628. `runExportedFunctionCore` sets `outcome = { status: "returned", value }` with no conversion, and you are right that the docstring I wrote overclaimed until this one was covered.

It needed a second entry point rather than a fourth caller of the same one: this boundary hands back a bare value from `invoke()`, not a `{ data }` node result. So `abortBoundary.ts` now exports `throwIfValueAborted` (the core) with `throwIfNodeResultAborted` as the thin `.data` wrapper over it. Called with `endsRun: false`, since this path has no end-of-run tail on its success side either.

Unit test covers the bare-value shape. I also reran the serve suites (19 files, 245 tests) since this is the path `./serve` and `__invokeFunction` go through.

packages/agency-lang/lib/runtime/abortedResult.ts
egonSchiele: You are right, and this one was my error from the last round rather than a pre-existing bug — I widened the bail without noticing that argument position is not a terminal drop. Fixed in 3f74c4628.

`dropped()` now takes a `terminal` flag. `atForkBoundary` and `atNodeBoundary` pass true and close the span an "erased" hop opened; `droppedAtArgPosition` passes false and behaves exactly as it did before this PR. Your split-trail scenario is now a test — carried → erased → `droppedAtArgPosition` asserts no span is ended and the instance keeps `span-1`, with a sibling test asserting the same chain through `atNodeBoundary` does end it. I verified the first one fails if I put the over-wide guard back.

Thanks for catching it. Ending a span on a path where the abort keeps travelling is exactly the kind of thing that would have shown up later as a confusing timeline rather than a test failure.

packages/agency-lang/lib/statelogClient.ts
egonSchiele: Fixed in 3f74c4628 — both the docstring above `abortSalvage` and `docs/dev/statelog.md`. While updating the doc I added the distinction the previous thread turned up, since a consumer needs it too: the two terminal drops (`clearedAtFork`, `droppedAtNodeBoundary`) end the `abortUnwind` span and `droppedAtArgPosition` does not.

packages/agency-lang/lib/runtime/errors.ts
egonSchiele: Agreed, and taken — 3f74c4628. `describeAbortCause` now returns `Execution aborted: <reason>` when `userKill` carries one, falling back to the kind when it does not. Two tests pin both halves.

Worth noting what this does and does not change, since `describeAbortCause` is now on more paths than before: `AgencyCancelledError` still sets its own message directly from the reason and does not route through `describeAbortCause`, so the thrown-directly text is unchanged. Only the rebuilt-from-cause path at a boundary gains the reason, which is precisely the path your scenario goes down.

PR #907, opened 2026-08-24
A match expression can be the body of a match arm (#887)

### LINKED ISSUE #887: Parser: a match expression cannot be the body of a match arm
## What happens

A `match` expression is not accepted as the body of a match arm. This does not parse:

```agency
def f(r: Result<string>): Result<string> {
  return match(r) {
    success(s) => match(r) {
      success(t) => success(t)
      failure(e) => failure(e)
    }
    failure(err) => failure(err)
  }
}
```

Error: `Failed to parse Agency program: expected match cases of the form `value => expression`, separated by `;` or newlines, followed by `}``.

The same shape with a block arm (`success(s) => { return match(r) { … } }`) was not tried; the workaround used was to compute the first result into a variable, early-return on failure, and match once on the second.

## Why it matters

The natural way to chain two fallible steps (read a file, then parse it) is a match inside a match. The parser limit pushes code back to `if (x is failure(err)) { return … }` chains, which is what the pattern-matching guide says to prefer `match` over.

## Where it came up

`cliAst` / `cliFmt` in `stdlib/agency.agency` (PR #886). Reproduce with the snippet above via `pnpm run ast`.

egonSchiele:
Fixes #887.

## What was wrong

A `match` expression could not be the body of a match arm, so the natural way to chain two fallible steps did not parse:

```agency
return match(r) {
  success(s) => match(r2) {
    success(t) => success(t)
    failure(e) => failure(e)
  }
  failure(err) => failure(err)
}
```

The cause is narrow. `match` is **not** part of `exprParser` — there is a separate `matchBlockExprParser` that gets wired into each value position by hand, one site at a time. Before this PR there were two such sites, `return` and assignment. The arm body was simply never wired up.

## The part that would have gone wrong quietly

Adding the parser alternative is one line and makes the snippet parse and run correctly. It also would have shipped a silent hole.

A single-expression arm hoists its value to a temp and yields it, keeping the original expression as the yield's `typeSource`. The synthesizer has no case for `matchBlock`, so it types that node as `any`. And in `computeMatchExprTypes`, one `any` yield makes the **entire outer match** `any`:

```ts
ctx.matchExprTypes[id] = types.some(isAnyType) ? ANY_T : ...
```

So the failure mode is not "the nested match is untyped" — it is "every assignability check on the outer match stops running, and nothing tells you". With only the parser change, this reported no error at all:

```agency
const val: string = match(r) {
  success(v) => match(r2) { success(w) => 1  failure(e) => 2 }
  failure(e) => 3
}
```

The arm lowerer now lowers a nested match through `lowerMatchExpressionCore` first and yields its `__matchval_` ref — exactly what `rewriteReturnsToYields` already does for `return match(...)` inside a block arm. `computeMatchExprTypes` walks match ids in descending order precisely so these resolve bottom-up. The case above is an error again.

## Formatter

`NEVER_INLINE_ARM_TYPES` existed for exactly one entry, `"matchBlock"`, because the grammar used to reject an inline nested match. With the grammar accepting it the list is empty, so it is deleted and the author's inline form now survives `agency fmt`.

## Not in scope

- **`match` as a function argument** — ruled out deliberately; it would mean folding `match` into `exprParser`.
- **`if ... then ... else` as an arm body** — still a parse error. The parser and lowering halves are nearly free (the shared `expressionRegion()` helper already handles `ifElse`), but `ifElse` is one node type for two surface forms with nothing recording which was written, so the formatter needs the position-based rule taught to it or the arm stops round-tripping. `docs/dev/match-expression-positions.md` writes this up.

## Verification

- 4490 tests across `lib/parsers`, `lib/lowering`, `lib/backends`, `lib/typechecker` pass, including the whole-corpus print→reparse identity test.
- New `tests/agency/match-arm-nested-match.agency` — 3/3, covering the inner-success, inner-failure, and outer-failure paths.
- **Negative control:** with the lowering change reverted, 2 of the 3 new typechecker tests fail. They are not passes-either-way checks.
- Formatter output on a nested match is idempotent and re-parses to an AST identical to the source's (ignoring `loc`).
- `lint:structure`, `fmt:ts`, and `lib/sourceIsText.test.ts` clean.

https://claude.ai/code/session_01PGjKdkGX9qQ2X1FqNYszdj

### DIFF
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -638,6 +638,18 @@ class PatternLowerer {
     matchId: number,
     arm: MatchBlockCase,
   ): AgencyNode[] {
+    if (body.length === 1 && body[0].type === "matchBlock") {
+      // A nested match as the arm's whole body: lower the inner match first and
+      // yield its `__matchval_` ref, exactly as `return match(...)` does in a
+      // block arm. Hoisting it as an opaque expression instead would leave the
+      // raw `matchBlock` as this yield's `typeSource`, which the synthesizer
+      // types as `any` — poisoning the outer match's whole value type.
+      const inner = this.lowerMatchExpressionCore(body[0] as MatchBlock, body[0].loc);
+      return [
+        ...inner.statements,
+        { type: "matchYield", matchId, value: inner.valueRef, loc: body[0].loc },
+      ];
+    }
     if (body.length === 1 && isExpressionNode(body[0])) {
       const expr = body[0] as Expression;
       // Always bind a single-expression arm's value to a temp at STATEMENT
--- a/packages/agency-lang/lib/lowering/patternLowering.ts
+++ b/packages/agency-lang/lib/lowering/patternLowering.ts
@@ -638,17 +638,28 @@ class PatternLowerer {
     matchId: number,
     arm: MatchBlockCase,
   ): AgencyNode[] {
-    if (body.length === 1 && body[0].type === "matchBlock") {
-      // A nested match as the arm's whole body: lower the inner match first and
-      // yield its `__matchval_` ref, exactly as `return match(...)` does in a
-      // block arm. Hoisting it as an opaque expression instead would leave the
-      // raw `matchBlock` as this yield's `typeSource`, which the synthesizer
-      // types as `any` — poisoning the outer match's whole value type.
-      const inner = this.lowerMatchExpressionCore(body[0] as MatchBlock, body[0].loc);
-      return [
-        ...inner.statements,
-        { type: "matchYield", matchId, value: inner.valueRef, loc: body[0].loc },
-      ];
+    if (body.length === 1) {
+      // A nested expression-position construct as the arm's whole body: lower it
+      // first and yield its `__matchval_` ref, exactly as `return match(...)` does
+      // in a block arm. Hoisting it as an opaque expression instead would leave the
+      // raw node as this yield's `typeSource`, which the synthesizer types as `any`
+      // — poisoning the outer match's whole value type.
+      //
+      // `expressionRegion` answers "is this an expression-position construct?";
+      // only the arm can answer "is this arm an expression position?". `ifElse` is
+      // ONE node type for two surface forms, and is an expression only in value
+      // position — for an arm, the inline form. A BLOCK arm holding a lone
+      // statement `if` must fall through to the return-rewrite below, or its
+      // branches are read as single expressions and every statement after the
+      // first is silently dropped. `matchBlock` is an expression in both forms.
+      const valuePosition = arm.blockBody !== true || body[0].type === "matchBlock";
+      const inner = valuePosition ? this.expressionRegion(body[0], body[0].loc) : null;
+      if (inner) {
+        return [
+          ...inner.statements,
+          { type: "matchYield", matchId, value: inner.valueRef, loc: body[0].loc },
+        ];
+      }
     }
     if (body.length === 1 && isExpressionNode(body[0])) {
       const expr = body[0] as Expression;
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -140,6 +140,7 @@ Pipeline and architecture:
 - `docs/dev/eval-command-agents.md` — Running a CLI as the eval agent (`--agent-cmd`): the EvalTarget union, tokenize-then-substitute, the AGENCY_CONFIG_OVERRIDES/AGENCY_TRACE_ID statelog handoff, process-group kill lifecycle, and the two cost-cap feeds
 - `docs/dev/pkg-imports.md` — Importing Agency code from npm packages using `pkg::` prefix
 - `docs/dev/trace.md` — Execution traces capturing checkpoints at every step
+- `docs/dev/match-expression-positions.md` — Where a `match` expression is allowed: `match` is not in `exprParser` but hand-wired into three sites, why folding it in is ruled out, why a nested-match arm must be LOWERED rather than hoisted (one `any` yield makes the whole outer match `any`), the formatter's inline/block re-parse rule, and the open `if ... then ... else` arm gap
 - `docs/dev/binop-parser.md` — Binary expression parser using precedence climbing
 - `docs/dev/locations.md` — How `loc.line` / `loc.col` / parse-mode template offset interact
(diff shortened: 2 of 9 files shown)


packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -638,6 +638,18 @@ class PatternLowerer {
     matchId: number,
     arm: MatchBlockCase,
   ): AgencyNode[] {
+    if (body.length === 1 && body[0].type === "matchBlock") {
```
egonSchiele: This hand-rolls the `matchBlock` case instead of calling the existing `expressionRegion()` helper (patternLowering.ts:566), which already does exactly this — "is this an expression-position control-flow construct? then give me `{statements, valueRef, matchId}`" — for `matchBlock` *and* `ifElse`. Now two places own that question.

The concrete cost is the gap this PR documents. `docs/dev/match-expression-positions.md` says the `if ... then ... else` arm is blocked only on the formatter and that "the lowering half is already written". It isn't, with the code as written: add `ifExpressionParser` to the arm alternatives and an arm body of `[ifElse]` falls straight past this new `matchBlock`-only check, then past `isExpressionNode` (`ifElse` is not in `EXPRESSION_NODE_TYPES`, lib/types.ts:129), into `rewriteReturnsToYields` — which finds no `return` in the branches, so `alwaysYields` is false and the whole thing dies with "match arm must return a value on every path". The next person follows the doc, wires up one parser line, and gets a lowering error the doc told them could not happen.

Routing through the shared helper both removes the second owner and makes the doc's claim true:

```suggestion
    if (body.length === 1) {
      // A nested expression-position construct as the arm's whole body: lower it
      // first and yield its `__matchval_` ref, exactly as `return match(...)` does
      // in a block arm. Hoisting it as an opaque expression instead would leave the
      // raw node as this yield's `typeSource`, which the synthesizer types as `any`
      // — poisoning the outer match's whole value type.
      const inner = this.expressionRegion(body[0], body[0].loc);
      if (inner) {
        return [
          ...inner.statements,
          { type: "matchYield", matchId, value: inner.valueRef, loc: body[0].loc },
        ];
      }
    }
```

packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -638,6 +638,18 @@ class PatternLowerer {
...
+      // A nested match as the arm's whole body: lower the inner match first and
+      // yield its `__matchval_` ref, exactly as `return match(...)` does in a
+      // block arm. Hoisting it as an opaque expression instead would leave the
+      // raw `matchBlock` as this yield's `typeSource`, which the synthesizer
+      // types as `any` — poisoning the outer match's whole value type.
+      const inner = this.lowerMatchExpressionCore(body[0] as MatchBlock, body[0].loc);
+      return [
+        ...inner.statements,
+        { type: "matchYield", matchId, value: inner.valueRef, loc: body[0].loc },
```
egonSchiele: Worth calling out in the doc as a known limitation, because the PR body's careful typing story stops one step short of it: this yield carries no `typeSource`, so `computeMatchExprTypes` types it by synthing the `__matchval_<inner>` ref, and that hook returns `matchExprTypes[inner]` — which was already run through `widenType`. A nested arm therefore loses literal types, and a literal-union annotation gets a false error:

```agency
const val: "a" | "b" = match(r) {
  success(v) => match(r2) { success(w) => "a"  failure(e) => "b" }
  failure(e) => "a"
}
```

I ran this: `Type 'string' is not assignable to type '"a" | "b"' (assignment to 'val')`. Flattened to one match it is accepted, and the sibling test in this same file ("literal-union annotation accepts literal yields (per-arm unwidened check)") is the one that passes because a plain arm keeps `typeSource`.

This is not a regression — `=> { return match(...) }` behaves identically today, I checked — so I am not asking you to fix it here. But this PR is what makes the shape reachable through the syntax it advertises, and `match-expression-positions.md` is otherwise precise about what the missing `typeSource` costs. One line there (the unwidened per-arm types do exist, in `ctx.matchExprYieldTypes[inner]`, if anyone wants to close it later) would save the next person the experiment.

packages/agency-lang/lib/lowering/patternLowering.ts
egonSchiele: Done in 814cb8f6 — routed through `expressionRegion()`, with one gate.

The two-owners point and the false doc claim were both right, and the claim is now true: with the helper wired in, adding just the `ifExpressionParser` line makes `1 => if (x > 0) then 10 else 20` parse, lower, and run correctly (`pick(1)` = 10, `pick(5)` = 0). I verified that, then reverted the parser line since the formatter still can't round-trip the shape.

The suggestion as written doesn't hold up ungated, though, and the failure is silent. `lowerIfExpressionCore` reads each branch as `thenBody[0] as Expression`, so a block arm holding a lone statement `if` gets its branches read as single expressions. I ran the suggested diff against this:

```agency
true => {
  if (c) { const x = 41  return x + 1 } else { return 2 }
}
```

Ungated it returns **41** — the value of the `const` binding — and drops `return x + 1`. And an arm whose branches yield nothing at all compiles silently instead of raising "match arm must return a value on every path" (I checked that one too; it went from a correct error to running and printing).

So the split I landed is: `expressionRegion` keeps "is this an expression-position construct?", and the arm lowerer answers the one question only it can — "is this arm an expression position?" For an arm that means the inline form, since `ifElse` is one node type for two surface forms. `matchBlock` has no such ambiguity and takes the path either way, which preserves today's behavior for `=> { match(...) }`.

New execution test `blockArmStatementIf` pins the 42. It fails (3/4) with the gate removed, so it isn't a passes-either-way check.

packages/agency-lang/lib/lowering/patternLowering.ts
egonSchiele: Reproduced your example verbatim and got the same error — `Type 'string' is not assignable to type '"a" | "b"'` — and confirmed the flattened single match is accepted.

Documented in 814cb8f6 as its own section in `match-expression-positions.md`, rather than a line, because the mechanism is the interesting part: the nested yield carries no `typeSource`, so `computeMatchExprTypes` types it by synthing the `__matchval_<inner>` ref, and that hook hands back `ctx.matchExprTypes[inner]`, which has already been through `widenType`. The section says it is not a regression (`=> { return match(...) }` has always done this) but that the inline syntax is what makes the shape easy to reach, and points at `ctx.matchExprYieldTypes[inner]` as where the unwidened per-arm types are for whoever closes it.

Not changing the behavior here, per your note.

Issue #689, opened 2026-07-25
agency run/compile: add --no-splices to refuse compile-time generator execution


egonSchiele:
## Summary

Compiling a file that contains `$( ... )` runs the generator. A user who has just cloned a repository may want to compile or inspect it without running anything it defines.

## Proposal

Add `--no-splices` to `agency run` and `agency compile`. With the flag, a file containing a splice refuses to compile, naming the splice and the generator it would have run, rather than expanding it.

Add the same option to `run` in `stdlib/agency.agency`, so an agent running untrusted Agency source can decline generator execution the same way it declines other capabilities.

## Notes

- The refusal should name the file and the generator, so the user can decide whether to re-run without the flag.
- Consider whether the default should flip for `run` on source the caller did not author. The stdlib `run` already compiles agent-supplied source in a sandbox with an import policy; generator execution is arguably in the same category.
- Related: #679 (splices), and the `--only-stdlib` proposal.

egonSchiele:
Shipped in #910 (squashed as aefcaefc).

`--refuse-splices` on `compile`, `run`, `typecheck` and `test run`, plus a `refuseSplices` config field, refusing `AG8016` before the generator is resolved. Off by default.

Three notes on how this differs from the proposal above, since the issue text is older than the code:

**Named `--refuse-splices`, not `--no-splices`.** The vendored commander treats any `--no-` long flag as a negation, so `--no-splices` would have arrived as `opts.splices` defaulting to `true` — inverted from every other safety flag and from the config field it sets.

**The stdlib half was already done, and more strictly than proposed.** `std::agency compile` refuses splices unconditionally through the closure validator. `run` was also the wrong place to ask for it: it takes an already-compiled program, so splices ran earlier. Nothing was needed there.

**The `agency run` half was largely covered by `--agency-only`**, which already refused splices — this issue was close to a duplicate of #690, closed with exactly that. The narrow flag was still worth adding because `--agency-only` is a bundle that also refuses TypeScript, and because `compile` and `typecheck` had no way to decline at all.

Two things found along the way:

- Type checking runs generators, and nothing could switch it off — `runCheckerPipeline` passed a hardcoded empty config. That plumbing is fixed, which is what makes `agency typecheck --refuse-splices` work.
[...]

PR #908, opened 2026-08-24
Every command translates its flags in one place (typecheck was the holdout)

egonSchiele:
Groundwork for #689. Consolidates CLI flag handling so `--refuse-splices` has one place to be added rather than three.

## The problem

`applyCliFlags` (`lib/config.ts`) is supposed to be the only definition of what a CLI flag means in config terms. Its own comment says so — "This is the ONLY place that defines what each flag means in config terms" — and so does `docs/dev/config.md`.

That was not true. `agency typecheck` wrote its `--strict` by hand:

```ts
config.typechecker = { ...config.typechecker, strictTypes: true };
```

This is the only such case in the CLI. (The other direct mutation, `config.verbose` for the root `-v`, lives inside `getConfig()` itself, so it is already applied once and uniformly for every command.)

## Why it existed

An unfinished migration, not a design choice. The dates line up:

- `typecheck`'s wiring landed **2026-05-14** (`60ab190a8`)
- `applyCliFlags` landed **2026-07-08** in #464

#464's design doc stated the goal outright:

> The flag→config mapping is extracted into one exported pure helper … instead of being re-implemented in each command action (**it is currently hand-written inline in `runWithOptions` and again in `typecheck`**). **All commands call the helper.**

`run` and `compile` were migrated; `typecheck` was not.

## The catch

The same flag name legitimately means two different things:

- On `run`/`compile`, `--strict` must set **both** `strict` and `strictTypes` to open a gate on the compile path.
- On `typecheck`, it must set `strictTypes` **alone**. That command runs the checker unconditionally and computes its own exit code, so it never reaches the gate.

**Correction after review:** an earlier version of this description claimed that setting `strict` on `typecheck` would make it fail on errors it currently walks past. That was wrong, and both reviewers caught it. `typechecker.strict` has exactly one reader in the repo — the gate itself at `lib/compiler/compile.ts:189`. Nothing under `lib/typeChecker/` reads it, and the command's exit code comes from diagnostic severity. So setting `strict` there would be **inert**, not harmful.

The narrow field is still the right call, for a better reason: an inert setting that reads as meaningful is a trap, and it would quietly acquire meaning the day `typecheck` grows a compile path. #464 independently said "Do not copy this wiring here.

The fix keeps both meanings as two fields on `CliFlags` rather than a branch keyed on command name. The command picks its meaning by choosing a field, and the helper never learns who called it.

## Behavior is unchanged

`applyCliFlags` guards every effect behind a flag test, so passing only `strictTypes` touches nothing else, and an absent `--strict` leaves `typechecker` untouched. Both are covered by new tests.

One subtlety worth noting for review: the old code mutated `config` *after* `runTypeCheck` had closed over it, which worked only because it mutated in place. `applyCliFlags` returns a copy, so the call is now made before the closure exists.

The three new tests pin `applyCliFlags` itself, not the wiring in `scripts/agency.ts` — tidying that call site to `{ strict: opts.strict }` would leave them green. Since `strict` is inert on this path, nothing can observe that regression, and a test that bites would need an injectable seam on the typecheck path that `CliDependencies` does not have. Adding production surface for one test is not worth it while the failure mode is invisible; the test comment now says exactly that rather than implying a consequence that does not exist.

## Verification

- `npx tsc --noEmit`, `pnpm run lint:structure`, `pnpm run fmt:ts` — all clean
- 107 tests pass across `config`, `config.modelAliases`, `config.providerModules`, `cli/commandLine`, `cli/runBundledAgent`, `sourceIsText`
- I confirmed the new tests actually bite by making `strictTypes` also set `strict`; they fail, and pass again when reverted

I did not run the real-binary CLI tests (`tests/integration/cli/test.mjs`), which need a full `make` and a packed tarball. CI covers those.

## Docs

`docs/dev/config.md` claimed `applyCliFlags` was the only definition of a flag's meaning, which was false with `typecheck` as the counterexample. That claim is now true, and I have added two things it was missing:

- that the same flag name may mean different things per command, with `--strict` as the worked example
- **flags that are NOT config** — the other channels the doc never mentioned. Interrupt policies, budgets, the test runner's output sink, and `--agency-only` (which selects a compile path rather than setting a value) all travel as resolved objects passed as arguments. `agency run` and `agency test run` both use both channels, which is expected rather than drift.

Review also caught that "exactly one of them" was wrong: `agency doctor` accepts `--trace` and `--log-file` and translates neither, forwarding them onto an argv for a child `agency` invocation (`lib/cli/doctor.ts:45`-`49`). That is now named as a third channel — flags forwarded verbatim to a child command line — rather than the claim being softened, since this doc is meant to be the authority.

https://claude.ai/code/session_01QD2WiV5DAqLYC9i1f3FLtn

### DIFF
--- a/packages/agency-lang/docs/dev/config.md
+++ b/packages/agency-lang/docs/dev/config.md
@@ -35,17 +35,22 @@ one exception is the root `-v` / `-c`, handled inside `getConfig()` itself, so
 it is still applied in exactly one place for every command.
 
 `typecheck` was the last holdout and was migrated in the change that added this
-section; before that, its `--strict` was written by hand, and it had already
-drifted from the shared version.
+section. Its `--strict` was hand-written rather than shared — but its narrower
+meaning was correct, and is preserved. See the next section.
 
 ### The same flag name may mean different things on different commands
 
 `--strict` is the live example. On `run`/`compile` it sets both `strict` and
 `strictTypes`, because the compile path has a gate that only opens on `strict`.
 On `typecheck` it sets `strictTypes` alone: that command runs the checker
-unconditionally and computes its own exit code, so it never reaches the gate,
-and setting `strict` there would make it fail on errors it currently reports
-and walks past.
+unconditionally and computes its own exit code, so it never reaches the gate.
+
+Setting `strict` on `typecheck` would therefore be **inert**, not harmful —
+`typechecker.strict` has exactly one reader in the repo, the gate itself
+(`lib/compiler/compile.ts`). The narrow field is still the right call for a
+different reason: a setting that reads as meaningful and does nothing is a
+trap, and it would quietly begin to matter the day `typecheck` grows a compile
+path.
 
 The two meanings are two fields on `CliFlags` (`strict` and `strictTypes`), not
 a branch inside `applyCliFlags`. The command picks the meaning it wants by
@@ -54,7 +59,7 @@ choosing a field; the helper never needs to know which command called it.
 ## Flags that are NOT config
 
 Not every flag becomes a setting, and forcing one to be a setting is a mistake.
-There are two channels, and a new flag belongs to exactly one of them:
+There are three channels, and most new flags belong to one of the first two:
 
 - **Config** (this document) — a value the compiler reads, or one that gets
   baked into the generated program. Goes through `applyCliFlags`.
@@ -64,11 +69,18 @@ There are two channels, and a new flag belongs to exactly one of them:
   runner's output sink is a live object with methods, and `--agency-only`
   selects a different compile path (through the closure validator) rather than
   setting a value.
+- **Forwarded verbatim to a child command line** — `agency doctor` accepts
+  `--trace [file]` and `--log-file <path>` and translates neither. It pushes
+  them back onto an argv for a child `agency` invocation (`lib/cli/doctor.ts`),
+  where `applyCliFlags` runs in that process instead. A wrapper command that
+  shells out belongs here, and the flag's meaning is still defined in exactly
+  one place — just not in this one.
 
 The test for which channel a flag belongs in: *is it a plain value the compiled
-program needs baked in, or is it a live object, or a choice of code path?* The
-first is config; the other two are arguments. `agency run` and `agency test run`
-both use both channels, which is expected, not drift.
+program needs baked in, or is it a live object, or a choice of code path, or is
+this command a wrapper that re-invokes `agency`?* The first is config; the next
+two are arguments; the last is forwarding. `agency run` and `agency test run`
+both use the first two channels, which is expected, not drift.
 
 ## What `--model` means
 
--- a/packages/agency-lang/docs/dev/config.md
+++ b/packages/agency-lang/docs/dev/config.md
@@ -26,6 +26,50 @@ Where applied: sources 1⊕2 at the CLI (baked into the generated program);
 source 3 at runtime, in the `RuntimeContext` constructor. Inspect the resolved
 result with `agency config show` (secrets masked; `--show-secrets` to reveal).
 
+### Every command translates its flags in the same place
+
+`applyCliFlags` is the only definition of what a flag means, and every command
+that turns a flag into a setting calls it — `run`, `compile`, `typecheck`, and
+the bundled agents. Do not hand-write `config.x = y` in a command action. The
+one exception is the root `-v` / `-c`, handled inside `getConfig()` itself, so
+it is still applied in exactly one place for every command.
+
+`typecheck` was the last holdout and was migrated in the change that added this
+section; before that, its `--strict` was written by hand, and it had already
+drifted from the shared version.
+
+### The same flag name may mean different things on different commands
+
+`--strict` is the live example. On `run`/`compile` it sets both `strict` and
+`strictTypes`, because the compile path has a gate that only opens on `strict`.
+On `typecheck` it sets `strictTypes` alone: that command runs the checker
+unconditionally and computes its own exit code, so it never reaches the gate,
+and setting `strict` there would make it fail on errors it currently reports
+and walks past.
+
+The two meanings are two fields on `CliFlags` (`strict` and `strictTypes`), not
+a branch inside `applyCliFlags`. The command picks the meaning it wants by
+choosing a field; the helper never needs to know which command called it.
+
+## Flags that are NOT config
+
+Not every flag becomes a setting, and forcing one to be a setting is a mistake.
+There are two channels, and a new flag belongs to exactly one of them:
+
+- **Config** (this document) — a value the compiler reads, or one that gets
+  baked into the generated program. Goes through `applyCliFlags`.
+- **Resolved objects passed as arguments** — `--policy` / `--approve` /
+  `--reject` become an interrupt policy object via `resolveRunPolicy`,
+  `--max-cost` / `--max-time` become a budget via `resolveBudget`, the test
+  runner's output sink is a live object with methods, and `--agency-only`
+  selects a different compile path (through the closure validator) rather than
+  setting a value.
+
+The test for which channel a flag belongs in: *is it a plain value the compiled
+program needs baked in, or is it a live object, or a choice of code path?* The
+first is config; the other two are arguments. `agency run` and `agency test run`
+both use both channels, which is expected, not drift.
+
 ## What `--model` means
 
 `agency run --model` and the `agency <file>` shorthand set the run's default
--- a/packages/agency-lang/lib/config.test.ts
+++ b/packages/agency-lang/lib/config.test.ts
@@ -214,6 +214,26 @@ describe("applyCliFlags", () => {
     });
   });
 
+  // `agency typecheck --strict` is a NARROWER meaning of the same flag name.
+  // That command calls the checker unconditionally and computes its own exit
+  // code, so it never reaches the compile-path gate that `strict` opens.
+  // Setting `strict` here would make typecheck fail on errors it currently
+  // reports and walks past.
+  it("--strict on typecheck sets strictTypes ONLY, never strict", () => {
+    expect(applyCliFlags({}, { strictTypes: true }).typechecker).toEqual({
+      strictTypes: true,
+    });
+  });
+
+  it("an absent typecheck --strict changes nothing", () => {
+    expect(applyCliFlags({}, { strictTypes: undefined }).typechecker).toBeUndefined();
+  });
+
+  it("the narrow meaning preserves other typechecker settings", () => {
+    const out = applyCliFlags({ typechecker: { enabled: true } }, { strictTypes: true });
+    expect(out.typechecker).toEqual({ enabled: true, strictTypes: true });
+  });
+
   it("--max-tool-call-rounds sets the top-level maxToolCallRounds", () => {
     expect(applyCliFlags({}, { maxToolCallRounds: 20 }).maxToolCallRounds).toBe(20);
   });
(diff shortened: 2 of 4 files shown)


packages/agency-lang/lib/config.ts
```diff
@@ -739,7 +739,13 @@ export type CliFlags = {
   logFile?: string;
   logStdout?: boolean;
   observability?: boolean;
+  /** `--strict` on `run`/`compile`: fail the run on any fatal type error. */
   strict?: boolean;
+  /** `--strict` on `typecheck`: untyped variables are errors, nothing else.
+   *  Same spelling as `strict`, deliberately narrower, because `typecheck`
+   *  runs the checker unconditionally and computes its own exit code — it
+   *  never reaches the compile-path gate that `strict` exists to open. */
```
egonSchiele: This says `typecheck` never reaches the gate, and the doc block below (and `docs/dev/config.md`, and the test comment) adds that setting `strict` would make it fail on errors it walks past. Only the first half holds.

`typechecker.strict` has exactly one reader in the repo: `lib/compiler/compile.ts:189`, `if (config.typechecker?.enabled || config.typechecker?.strict)`. That is the gate on whether the compile path runs the checker. Nothing in `lib/typeChecker/` reads it, and this command's exit code comes from `errors.some((e) => e.severity === "error")`. So `strict` on `agency typecheck` would do nothing at all.

The field is still the right call — just for the other reason: an inert setting that reads as meaningful is a trap, and it would quietly acquire meaning the day `typecheck` grows a compile path.

packages/agency-lang/lib/config.test.ts
```diff
@@ -214,6 +214,26 @@ describe("applyCliFlags", () => {
     });
   });
 
+  // `agency typecheck --strict` is a NARROWER meaning of the same flag name.
+  // That command calls the checker unconditionally and computes its own exit
+  // code, so it never reaches the compile-path gate that `strict` opens.
+  // Setting `strict` here would make typecheck fail on errors it currently
+  // reports and walks past.
+  it("--strict on typecheck sets strictTypes ONLY, never strict", () => {
```
egonSchiele, line 230: These three pin `applyCliFlags`, not the wiring. If someone tidies `scripts/agency.ts:1499` to `{ strict: opts.strict }`, all three stay green — which is the one regression the PR is guarding against.

Since `strict` is inert on the typecheck path (see the comment on `config.ts`), no CLI-level test can see the difference either; the only test that bites is one asserting the resolved config at the wiring level, and `createProgram` already takes injectable deps for that. If that is not worth the machinery, that is a fine answer — just say it here rather than describing a behavior consequence.

packages/agency-lang/docs/dev/config.md
```diff
@@ -26,6 +26,50 @@ Where applied: sources 1⊕2 at the CLI (baked into the generated program);
...
+
+`applyCliFlags` is the only definition of what a flag means, and every command
+that turns a flag into a setting calls it — `run`, `compile`, `typecheck`, and
+the bundled agents. Do not hand-write `config.x = y` in a command action. The
+one exception is the root `-v` / `-c`, handled inside `getConfig()` itself, so
+it is still applied in exactly one place for every command.
+
+`typecheck` was the last holdout and was migrated in the change that added this
+section; before that, its `--strict` was written by hand, and it had already
+drifted from the shared version.
```
egonSchiele: "drifted from the shared version" fights the section directly below it, which argues the difference is deliberate and must be preserved. A reader who hits both will not know whether the old narrow meaning was a bug or the point. It was hand-written rather than shared, and its narrower meaning was correct — that is the whole story.

packages/agency-lang/docs/dev/config.md
```diff
@@ -26,6 +26,50 @@ Where applied: sources 1⊕2 at the CLI (baked into the generated program);
...
+and walks past.
+
+The two meanings are two fields on `CliFlags` (`strict` and `strictTypes`), not
+a branch inside `applyCliFlags`. The command picks the meaning it wants by
+choosing a field; the helper never needs to know which command called it.
+
+## Flags that are NOT config
+
+Not every flag becomes a setting, and forcing one to be a setting is a mistake.
+There are two channels, and a new flag belongs to exactly one of them:
```
egonSchiele: "exactly one of them" is not quite true, and this doc is now the authority so it is worth getting right. `agency doctor` takes `--trace [file]` and `--log-file <path>` (`scripts/agency.ts:1791`–`1792`) and does neither thing: `lib/cli/doctor.ts:45`–`49` pushes them back onto an argv for a child `agency` invocation, which applies them through `applyCliFlags` in its own process. Either name that third channel — flags forwarded verbatim to a child command line — or soften the claim.

egonSchiele (commented):
The change is right and it is small. Routing `typecheck` through `applyCliFlags` is the correct fix, and two fields on `CliFlags` beat a branch keyed on the command name.

I checked the behavior claim rather than taking it: `applyCliFlags` guards every single effect behind a flag test (`lib/config.ts:783`–`842`), so `{ strictTypes: opts.strict }` cannot touch anything else. The ordering change is inert too, for a second reason the description does not give: `getConfig()` calls `loadConfig` fresh on every call (`scripts/agency.ts:235`), so nothing else was holding the object the old code mutated in place. And the mutation already happened before `SymbolTable.build`, so that call sees the same config it saw before.

One thing to fix before merge, and three smaller notes.

**The stated reason for the narrow meaning is not true.** It appears four times — in the PR description, in `lib/config.ts:747`, in `lib/config.ts:770`, in `docs/dev/config.md:46`, and in the test comment at `lib/config.test.ts:221`. The sentence is: `typecheck` never reaches the compile-path gate, **and** setting `strict` there would make it fail on errors it currently reports and walks past. Those two halves contradict each other, and only the first is true. `typechecker.strict` is read in exactly one place in the whole repo — `lib/compiler/compile.ts:189` — and only as the gate deciding whether the compile path runs the checker at all. Nothing under `lib/typeChecker/` reads it (that code reads `strictTypes` and `strictMemberAccess`), and the `typecheck` command sets `hasErrors` from `errors.some((e) => e.severity === "error")` with no reference to config. So setting `strict` on `agency typecheck` today would be **inert**, not harmful.

That does not make the narrow field wrong — it makes the argument for it different, and I think better: a config value that looks meaningful and does nothing is worse than one that is absent, and if `typecheck` ever grew a compile path the setting would silently start mattering. Worth stating that instead, in all four places.

**Nothing tests the wiring.** All three new tests call `applyCliFlags` directly, so they pin the helper. The exact regression this PR exists to prevent — someone tidying `scripts/agency.ts:1499` into `{ strict: opts.strict }` — leaves all three green. I would rank this second, not first, because of the finding above: since `strict` is inert on that path, there is no black-box behavior for a CLI test to observe, so the only honest test is one that asserts the resolved config at the wiring level (`createProgram` already takes injectable deps). If that is more machinery than it is worth, say so in the comment rather than implying a behavior consequence that does not exist.

Verification of the rest checked out: `applyCliFlags` really does now have every caller — `run`, `compile`, `typecheck`, `runBundledAgent` — and `typecheck` was the only hand-written case left; the root `-v` is the one exception, inside `getConfig()`.

packages/agency-lang/lib/config.ts
egonSchiele: You are right, and I verified it before changing anything: `typechecker.strict` has exactly one reader in the repo, the gate at `lib/compiler/compile.ts:189`. The other two grep hits are the flag *descriptions* on run/compile, not reads. Nothing under `lib/typeChecker/` touches it, and this command's exit code comes from `errors.some(e => e.severity === "error")`.

So `strict` here would be inert, not harmful, and my two halves did contradict each other. Corrected in all five places (both spots in `config.ts`, `docs/dev/config.md`, the test comment, the wiring comment in `scripts/agency.ts`) and in the PR description, using your reason: an inert setting that reads as meaningful is a trap, and it would start mattering silently the day typecheck grows a compile path. Fixed in 807494ec.

packages/agency-lang/lib/config.test.ts
egonSchiele: Agreed on both counts, and I am taking the answer you left open: not worth the machinery.

`CliDependencies` currently exposes only `loadLspStartServer`, `loadMcpStartServer`, `resolveMcpCommand` and `launchAgent` — nothing on the typecheck path. So a wiring test means adding an injectable seam to production code for one test, to guard a regression that (as you established) nothing can observe.

I have rewritten the comment to say what these tests do and do not cover, and why there is no wiring test, instead of implying a behavior consequence that does not exist. If typecheck ever grows a compile path, `strict` stops being inert and a black-box test becomes possible — that is the point at which this is worth revisiting. Fixed in 807494ec.

packages/agency-lang/docs/dev/config.md
egonSchiele: Fair — it read as though the old narrow meaning was the bug, which is the opposite of the section below. Rewritten to: it was hand-written rather than shared, and its narrower meaning was correct and is preserved, with a pointer to the next section. Fixed in 807494ec.

packages/agency-lang/docs/dev/config.md
egonSchiele: Verified and fixed. `agency doctor` declares `--trace [file]` and `--log-file <path>` (`scripts/agency.ts:1791`-`1792`) and translates neither: `lib/cli/doctor.ts:45`-`49` pushes them onto an argv for a child `agency` invocation, so `applyCliFlags` runs in the child process.

I named it as a third channel rather than softening the claim, since the doc is meant to be the authority: *flags forwarded verbatim to a child command line*. The 'which channel' test now has a fourth question for wrapper commands that re-invoke `agency`, and the point that the flag's meaning is still defined in exactly one place — just not in this process. Fixed in 807494ec.

packages/agency-lang/lib/config.test.ts
```diff
@@ -214,6 +214,26 @@ describe("applyCliFlags", () => {
     });
   });
 
+  // `agency typecheck --strict` is a NARROWER meaning of the same flag name.
+  // That command calls the checker unconditionally and computes its own exit
+  // code, so it never reaches the compile-path gate that `strict` opens.
+  // Setting `strict` here would make typecheck fail on errors it currently
+  // reports and walks past.
```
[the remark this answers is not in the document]
egonSchiele: Correct, and fixed in 807494ec. `typechecker.strict` is only read by the compile gate at `lib/compiler/compile.ts:189`, which this command bypasses, so setting it would be inert rather than causing extra failures. The comment now says that, and gives the actual reason for the narrow field: an inert setting that looks meaningful is a trap.

packages/agency-lang/docs/dev/config.md
```diff
@@ -26,6 +26,50 @@ Where applied: sources 1⊕2 at the CLI (baked into the generated program);
...
+drifted from the shared version.
+
+### The same flag name may mean different things on different commands
+
+`--strict` is the live example. On `run`/`compile` it sets both `strict` and
+`strictTypes`, because the compile path has a gate that only opens on `strict`.
+On `typecheck` it sets `strictTypes` alone: that command runs the checker
+unconditionally and computes its own exit code, so it never reaches the gate,
+and setting `strict` there would make it fail on errors it currently reports
+and walks past.
```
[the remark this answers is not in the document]
egonSchiele: Correct, and fixed in 807494ec. The doc no longer claims `strict` would change this command's exit behavior; it now states that `strict` has exactly one reader (the compile gate) which typecheck never reaches, so setting it would be inert — and gives the real reason the narrow field exists.

PR #909, opened 2026-08-24
Name the terminal tab after the running command, and spin while it works

egonSchiele:
## What this does

Terminals name a tab after its foreground job, so every Agency command showed up in the tab bar as `node`. And a tab you were not looking at got the ordinary "there is unread output here" dot, which says nothing about whether the command is still working.

Now every command names its tab after itself and drives the tab's progress indicator:

- **Title** — the command as typed plus its first operand: `agency eval run fib`, `agency run investment.agency`, `agency logs view`. A path operand is shortened to its filename, since a tab has room for that but not a directory tree.
- **Indicator** — a spinner while the command runs, cleared when it succeeds, red when it fails. A long `eval` that dies while you are in another tab is visibly failed when you come back.

The title goes on the terminal's title stack first, so the tab gets its old name back when the command exits.

## How it works

`lib/cli/terminalStatus.ts` is the only file in the repo that knows these escape sequences. Two unrelated families do the work: `ESC ] 1` for the tab title with `CSI 22 t` / `CSI 23 t` to push and pop it, and ConEmu's `ESC ] 9 ; 4` progress protocol, which iTerm2, WezTerm, Ghostty and Windows Terminal all draw in the tab. Terminals that do not understand a sequence ignore it, so there is nothing to feature-detect.

It writes nothing unless stdout is a TTY and neither `NO_COLOR` nor `AGENCY_NO_TERM_STATUS` is set. Deliberately not `autoUseColor()`: that one lets `FORCE_COLOR` override a non-TTY, which is right for color in a captured log and wrong here, because there is no tab at the other end of a pipe.

**Nesting.** The first `agency` process to name the tab sets `AGENCY_TERM_STATUS_OWNED=1` on its own `process.env`, which every child inherits. So when `eval run --agent-cmd` spawns Agency processes, or an agent spawns a subprocess, only the outermost one writes.

**Signals.** The `SIGINT`/`SIGTERM` handlers clean up but never call `process.exit`. The eval runner and the logs viewer install their own signal handling for graceful shutdown, and since ours is registered first it would preempt them. Instead each handler cleans up, removes itself, and re-raises only when no other listener remains, which is what keeps Ctrl-C working when we are the only handler.

**Wiring** lives in `runCli`, not `createProgram`, because tests construct programs many times over and the process-level listeners must not pile up. It is one commander `preAction` hook on the root program; commander walks the ancestor chain when firing hooks, so one hook covers every subcommand including the shorthand that falls back to `run`.

**The logs viewer** takes over the screen and then waits for keys, which is not "busy": it drops the spinner and keeps the name.

## Testing

20 unit tests in `lib/cli/terminalStatus.test.ts` covering the exact bytes for each state, every quiet condition, the ownership claim, the do-not-pop-what-you-did-not-push guard, title sanitizing, and title derivation through real commander parsing.

Verified end to end against the built CLI on a real pty:

```
$ script -q /dev/null node dist/scripts/agency.js ast probe.agency | cat -v
^[[22;0t^[]1;agency ast probe.agency^G^[]9;4;3;0^[\{
  ... json ...
}
^[]9;4;0;0^[\^[[23;0t
```

And nothing at all when piped, when `AGENCY_TERM_STATUS_OWNED` is already set, or under `NO_COLOR`.

`pnpm run typecheck`, `pnpm run lint:structure`, and `pnpm run fmt:ts:check` are clean.

## Known behavior

A failed command leaves the tab red on purpose, and nothing clears that automatically — the next Agency command will, or `printf '\e]9;4;0;0\e\\'` by hand. Documented in `docs/dev/terminal-status.md`.

https://claude.ai/code/session_01YSuaAP8vGc3b3eNYGmQf3Z


packages/agency-lang/lib/cli/terminalStatus.ts
```diff
@@ -0,0 +1,158 @@
...
+ * keeps Ctrl-C working when we are the only handler.
+ */
+export function installTerminalStatus(program: Command, status: TerminalStatus): void {
+  program.hook("preAction", (_program, actionCommand) => {
+    status.begin(commandTitle(actionCommand));
+  });
+
+  process.on("exit", (code) => status.end(code === 0));
+
+  for (const signal of ["SIGINT", "SIGTERM"] as const) {
```
egonSchiele: Registering a JS `SIGINT` listener changes how the whole CLI responds to Ctrl-C, not just how it paints the tab. With no listener, Node leaves SIGINT at its default disposition and the kernel kills the process immediately. Once a listener exists, libuv defers the signal to the event loop — so the handler cannot run until the current synchronous block finishes.

Concrete regression: `agency compile big.agency` (parsing and codegen are largely synchronous, and parsing is the known front-end bottleneck) previously died instantly on Ctrl-C. With this hook installed in `runCli` before `parseAsync`, that same Ctrl-C is queued and the process keeps grinding until the sync work yields. The same applies to `agency ast`, `agency typecheck`, and `agency fmt` on a large input.

Worth either scoping the signal handlers to commands that are actually async/long-running, or accepting and documenting this, but it should be a deliberate decision — the PR description only discusses handler *ordering*, not the fact that installing a handler at all disables default SIGINT termination.

packages/agency-lang/lib/cli/terminalStatus.ts
```diff
@@ -0,0 +1,158 @@
...
+    const handler = (): void => {
+      status.end(false);
```
egonSchiele: `end(false)` on every signal treats "a signal arrived" as "the command failed", but several commands treat the first Ctrl-C as a normal, non-fatal event — and this handler is registered first, so it fires before they get a say.

- `lib/eval/run/runSuite.ts:127` — the first Ctrl-C sets `interrupted = true`, prints "folding the in-flight test into the run directory; press Ctrl-C again to force quit", and the eval **keeps running**, potentially for minutes. By then this handler has already written the red-error progress state and popped the title, so a still-working command sits in the tab with a red indicator and its old name. The spinner is also gone for the rest of the run, and `claimed` is now false so nothing can put it back.
- `scripts/agency.ts:351` — `compile --watch` handles SIGINT by closing cleanly and `process.exit(0)`. Ctrl-C is the normal way to stop a watcher, and it now leaves the tab red.
- `lib/cli/logsView.ts:182` — Ctrl-C out of the viewer exits 130, which is also a normal quit; red tab.

Since the `exit` handler already covers the real exit code, the signal handlers arguably only need to restore the title (a plain pop, or `end` with the state left neutral) rather than assert failure.

packages/agency-lang/scripts/agency.ts
```diff
@@ -2331,6 +2332,10 @@ export async function runCli(
 ): Promise<void> {
   loadEnv();
   const program = createProgram(deps);
+  // Installed here rather than in createProgram: this is the one real entry
+  // point, and the process-level listeners must not accumulate across the many
+  // createProgram calls that tests make.
+  installTerminalStatus(program, terminalStatus);
```
egonSchiele, line 2349: The comment's premise does not hold: `runCli` itself is called repeatedly from tests. `scripts/agency.test.ts` calls it seven times (lines 110, 120, 130, 143, 365, 377), and each call adds three permanent process listeners (`exit`, `SIGINT`, `SIGTERM`) that nothing ever removes. So the listeners accumulate exactly as moving the wiring here was meant to prevent — just at a smaller multiple. Seven is under Node's default max of 10, so no warning today, but one more `runCli` test tips it into a `MaxListenersExceededWarning`, and `terminalStatus.test.ts` had to write its own `afterEach` listener-scrubber precisely because of this.

Related: `end()` never clears `AGENCY_TERM_STATUS_OWNED` from `process.env`. Once any `begin()` in a process succeeds, `isEligible` returns false forever, so a second `runCli` in the same process is silently mute. Fine for the real one-shot CLI, surprising for anything that embeds it.

Making `installTerminalStatus` idempotent (register once per process) would fix both the accumulation and the need for the test's scrubber.

packages/agency-lang/lib/cli/terminalStatus.ts
```diff
@@ -0,0 +1,158 @@
...
+ * room for a filename but not for a directory tree.
+ */
+export function commandTitle(command: Command): string {
+  const names: string[] = [];
+  for (let node: Command | null = command; node !== null; node = node.parent) {
+    names.unshift(node.name());
+  }
+  const operand = command.args[0];
+  if (operand !== undefined && operand !== "" && !operand.startsWith("-")) {
+    names.push(operand.includes(path.sep) ? path.basename(operand) : operand);
```
egonSchiele: `path.sep` is `\` on Windows, so `agency run tests\agency\foo.agency` shortens correctly but the far more common `agency run tests/agency/foo.agency` does not — the whole path goes into the tab title, which is the thing this branch exists to avoid.

The guard is also unnecessary: `path.basename("fib")` is `"fib"` and `path.basename("eval-suite")` is `"eval-suite"`, so calling `path.basename` unconditionally gives the same answer for non-path operands and fixes the separator case for free.

```suggestion
    names.push(path.basename(operand));
```

packages/agency-lang/lib/cli/logsView.ts
```diff
@@ -180,6 +181,9 @@ export function createViewerHost(
     const sigterm = onSignal(143);
     process.on("SIGINT", sigint);
     process.on("SIGTERM", sigterm);
+    // The viewer owns the screen from here, and it is interactive rather than
+    // busy: keep the tab's name, drop the spinner.
+    terminalStatus.stopProgress();
```
egonSchiele: This reaches for the module singleton inside a function whose whole point is injection. The doc comment right above `createViewerHost` calls it "the ONE owner of terminal construction ... the stdin→TTY swap, viewport, `runViewer`, and cleanup", and every other terminal touch arrives through `ViewerTerminalDependencies`. The singleton writes straight to `process.stdout`, so viewer tests that pass fake input/output now have one unmockable path to the real terminal — quiet today only because `claimed` is false in a non-TTY test process, which is an accident rather than a guarantee.

Passing the status through `ViewerTerminalDependencies` (or having the caller stop the progress before entering the viewer) keeps the file's stated contract intact.

egonSchiele (commented):
Nice feature, and the shape is right: one file owns the sequences, a dependency seam makes it testable, four quiet conditions, and the env-based one-owner rule for nesting. I checked that last one rather than assuming it — `runBundledAgent` spawns through `withRootCarriers`, which spreads `process.env` (`lib/cli/childEnv.ts:25`), so the claim really does reach the child. I also checked that no test harness in the repo runs the CLI on a pty and parses stdout, so the escape sequences cannot corrupt `--json` output anywhere.

The payoff here is cosmetic and the mechanism is not: it installs a process-wide `SIGINT` listener ahead of every command. That is where I would spend the review, and it is where both of my real findings are.

**1. Ctrl-C turns the tab red immediately and permanently, even when the command handles the signal and then succeeds.** Our handler is registered first, so it always runs first, and it always calls `end(false)`. Walk `eval run`: you hit Ctrl-C, our handler paints the tab red, pops the title, sets `claimed = false`, and removes itself. Then `runSuite`'s handler (`lib/eval/run/runSuite.ts:127`) prints "folding the in-flight test into the run directory; press Ctrl-C again to force quit" and **the suite keeps running** — possibly for minutes, finishing the in-flight test and writing the run directory. For that whole stretch the tab is named `node` again and shows red. It then exits 0, and the `exit` handler's `end(true)` is a no-op because `claimed` is already false. So a successful interrupted eval ends with a permanently red tab, which per the "Known behavior" section nothing clears. The logs viewer has the same problem and it is worse there, because Ctrl-C is a normal way out of a viewer.

The root cause is that the signal handler does the cleanup unconditionally, when the only case that actually needs it is the one where nothing else will handle the signal — if someone else handles it, the process either exits normally (and the `exit` handler covers us with the true exit code) or keeps running (and we should not have touched anything). Suggested shape, in the inline comment.

**2. On Windows the cleanup can simply be lost.** `lib/cli/terminalStatus.ts:107` and `docs/dev/terminal-status.md:70` both rest on "`process.stdout.write` to a TTY is synchronous, so an `exit` handler holds." That is true on POSIX and false on Windows — Node's own process docs give it as "TTYs: asynchronous on Windows, synchronous on POSIX" — and Windows Terminal is one of the four terminals this PR targets. The failure is the ugly one: the command exits, the title is never popped, and the spinner keeps spinning. `fs.writeSync(1, text)` is synchronous on both platforms and is a one-line change to the seam.

**Two smaller things.** The basename heuristic assumes the first operand is a path, which mangles the one command whose first operand is free text (`agency agent`) — details inline. And the signal path is the only part of this file with no test, which is backwards from its risk: everything else here is cosmetic if it breaks, while this one can stop Ctrl-C from working. The file already has a `write`/`env`/`isTty` seam; `listenerCount` and `kill` would fit it and make the interesting decision assertable.

Everything else checked out. The 20 tests are really there and really cover what the description says, `stopProgress` correctly leaves `claimed` true so the title still pops, and the `--csv` path never reaches the viewer's `stopProgress`, which is right since it is not interactive.

packages/agency-lang/lib/cli/terminalStatus.ts
```diff
@@ -0,0 +1,158 @@
...
+    status.begin(commandTitle(actionCommand));
+  });
+
+  process.on("exit", (code) => status.end(code === 0));
+
+  for (const signal of ["SIGINT", "SIGTERM"] as const) {
+    const handler = (): void => {
+      status.end(false);
+      process.removeListener(signal, handler);
+      if (process.listenerCount(signal) === 0) process.kill(process.pid, signal);
```
egonSchiele: `status.end(false)` runs before the listener count is consulted, so it happens even in the case the design is trying to defer — someone else is handling this signal.

That is what paints the tab red the moment you press Ctrl-C during `eval run`, while `runSuite` goes on to fold in the in-flight test and exit 0 several minutes later. And `claimed` is now false, so the `exit` handler cannot undo it.

Doing the cleanup only in the branch that re-raises fixes both, and I think dropping the unconditional `removeListener` is the other half of it — otherwise we are gone after the first signal and cannot help on the second:

```ts
const handler = (): void => {
  // Only the last listener standing cleans up: anyone else handling this
  // either exits (and the `exit` handler covers us with the real code) or
  // keeps working (and the tab should keep spinning).
  if (process.listenerCount(signal) > 1) return;
  status.end(false);
  process.removeListener(signal, handler);
  process.kill(process.pid, signal);
};
```

That also handles the force-quit path the current code loses: `runSuite` registers with `once`, so on the second Ctrl-C we are the only one left, and we clean up and re-raise.

Worth deciding separately whether exit 130 should be red at all. For `agency logs`, Ctrl-C is a normal way to leave the viewer, and it currently leaves the tab red for good.

packages/agency-lang/lib/cli/terminalStatus.ts
```diff
@@ -0,0 +1,158 @@
...
+    stopProgress(): void {
+      if (!claimed) return;
+      deps.write(progressSequence(0));
+    },
+  };
+}
+
+/**
+ * The process-wide instance the CLI wires up. Writes are synchronous to a TTY,
+ * which is what lets `end` run from an `exit` handler.
```
egonSchiele: "Writes are synchronous to a TTY" holds on POSIX but not on Windows — Node documents TTY writes as asynchronous there — and the PR targets Windows Terminal by name. So on Windows the `exit` handler's write can be dropped, and the tab is left named and spinning after the command is gone.

`write: (text) => { fs.writeSync(1, text); }` is synchronous on both, and since this seam only writes at begin/end there is nothing to interleave badly with the buffered stream. (Related: `process.kill(pid, SIGINT)` on Windows terminates rather than delivering a signal, so the re-raise path is a hard kill there — fine, but it is a second reason the write has to have landed first.)

packages/agency-lang/lib/cli/terminalStatus.ts
egonSchiele: The basename shortening assumes the operand is a path, and `agency agent` breaks that assumption: its first operand is the prompt. `agency agent "fix lib/parsers/foo.ts"` contains a separator, so the title becomes `agency agent foo.ts` — the prompt silently truncated at the last slash. Without a slash the whole prompt lands in the tab title instead, which may be what you want, but the two behaviors together are hard to explain.

While you are here: the interactive `agency agent` session is the same "interactive, not busy" shape the logs viewer got `stopProgress()` for, and it spins for the entire session. It is genuinely harder than the viewer, since the work happens in the spawned child — a spinner is right while the agent thinks and wrong while it waits on you — so an explicit decision either way is worth a line in the doc.

Minor, same line: `operand.includes(path.sep)` misses forward-slash paths on Windows, where `path.sep` is a backslash but `agency run tests/foo.agency` still works. `/[/\\]/` covers both.

packages/agency-lang/lib/cli/terminalStatus.test.ts
```diff
@@ -0,0 +1,205 @@
...
+      stopProgress: () => {},
+    };
+    const program = new Command();
+    program.name("agency").exitOverride();
+    program
+      .command("eval")
+      .command("run")
+      .argument("[test]")
+      .action(() => {});
+    installTerminalStatus(program, status);
```
egonSchiele, line 205: This block tests the `preAction` hook and stops there. Nothing covers the signal handlers, which is the inverse of where the risk sits: if the title logic breaks, a tab is named wrong; if this breaks, Ctrl-C stops working.

The decision is testable without sending real signals, if `listenerCount` and `kill` join the existing `write`/`env`/`isTty` seam — then "re-raises when it is the only listener" and "does not when someone else is listening" are two ordinary assertions. The `afterEach` cleanup you already wrote here is the hard part, and it is done.

packages/agency-lang/docs/dev/terminal-status.md
```diff
@@ -0,0 +1,97 @@
...
+
+Cleanup runs from `process.on("exit")`, whose handlers must be synchronous;
+`process.stdout.write` to a TTY is, so that holds.
+
+The `SIGINT` and `SIGTERM` handlers deliberately do **not** call
+`process.exit`. Other parts of the CLI install their own signal handling for
+graceful shutdown — the eval runner (`lib/eval/run/runSuite.ts`), the logs
+viewer (`lib/cli/logsView.ts`) — and since our listener is registered first, it
+would preempt them. Instead each handler cleans up, removes itself, and
+re-raises the signal only when no other listener remains. If someone else is
```
egonSchiele: This paragraph describes the intended behavior, and the code does something narrower: it cleans up first and unconditionally, then checks whether anyone else is listening. So "if someone else is handling the signal, we have stepped out of their way" is not what happens — we have already painted the tab red and popped the title before stepping aside. Worth re-reading against the code once the handler changes, since this doc is the thing a future reader will trust.

Same section, line 70: the synchronous-write premise is POSIX-only (see the comment on `terminalStatus.ts`).

packages/agency-lang/lib/cli/terminalStatus.ts
```diff
@@ -0,0 +1,158 @@
...
+ */
+export function commandTitle(command: Command): string {
+  const names: string[] = [];
+  for (let node: Command | null = command; node !== null; node = node.parent) {
+    names.unshift(node.name());
+  }
+  const operand = command.args[0];
+  if (operand !== undefined && operand !== "" && !operand.startsWith("-")) {
+    names.push(operand.includes(path.sep) ? path.basename(operand) : operand);
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed, though not quite as suggested — see the thread on `agency agent` below, which arrived after this one and pulls the other way.

Unconditional `path.basename` breaks `agency agent`, whose first operand is a free-form prompt: `agency agent "fix lib/parsers/foo.ts"` became `agency agent foo.ts`. So shortening now requires the operand to *look like* a path — a separator and no whitespace. The separator test is `/[/\\]/` rather than `path.sep`, which covers the forward-slash-on-Windows case you raised. `path.basename` itself is already per-platform, so it needs no `posix`/`win32` split.

packages/agency-lang/lib/cli/terminalStatus.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  for (const signal of ["SIGINT", "SIGTERM"] as const) {
+    const handler = (): void => {
+      status.end(false);
+      process.removeListener(signal, handler);
+      if (process.listenerCount(signal) === 0) process.kill(process.pid, signal);
+    };
+    process.on(signal, handler);
+  }
+}
```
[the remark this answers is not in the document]
egonSchiele: Fixed. The premise was worse than stated: `runCli` itself is called six times over in `scripts/agency.test.ts`, so moving the wiring there did not avoid the accumulation at all.

The `preAction` hook is still attached to every program; the process-level listener is now latched to one per process. There is also only one such listener left — the signal handlers are gone entirely, for reasons in the threads below.

packages/agency-lang/lib/cli/terminalStatus.ts
egonSchiele: You are right, and the fix is to drop the signal handlers altogether rather than scope them.

I verified the disposition claim both directions with a probe that installs this wiring and then blocks the event loop for 10s, with SIGINT sent at 1s:

```
default        exit=130 after 1.0s | ready
withListener   exit=0   after 10.1s | ready|finished the whole 10s block
```

One empty SIGINT listener is enough to turn an instant Ctrl-C into one that waits out the whole synchronous block — and swallow it entirely. Exactly the `agency compile big.agency` regression you described.

Scoping to long-running commands would mean a hand-maintained command list, and it would not have fixed the second problem you raise in the next thread. So: no signal handler, cleanup hangs off `exit` alone.

The cost is that a hard Ctrl-C on a command with no signal handler of its own strands the spinner. It clears the next time any Agency command runs in that tab, and the commands you are most likely to interrupt — the eval runner, `compile --watch`, the log viewer — all handle their own signal and leave through `process.exit`, which does run the exit handler. Written up in `docs/dev/terminal-status.md` under "Signals, and why there are none".

packages/agency-lang/lib/cli/terminalStatus.ts
egonSchiele: Agreed, and this is the second reason the handlers are now gone rather than reordered. A signal is not a verdict, and no handler here can tell an ending from a pause — only `exit` knows.

The three cases you name now behave like this. The eval runner's first Ctrl-C: nothing happens at all, the spinner keeps spinning while it folds in the in-flight test, and the real exit code decides at the end. `compile --watch`: exits 0, clean. The log viewer: exits 130.

On your last point — 130 should not be red, and it no longer is. `commandFailed` treats 0, 130 and 143 as normal ways to stop and everything else as failure, tested per code.

packages/agency-lang/scripts/agency.ts
egonSchiele: Checked, and the comment was simply wrong: `scripts/agency.test.ts` calls `runCli` six times (lines 110, 120, 130, 143, 365, 377). The comment is gone.

`installTerminalStatus` is now idempotent for the process listener — latched to one per process — while still hooking every program it is handed. Both are tested. The `afterEach` scrubber in the test file is gone with them.

`end()` now also deletes `AGENCY_TERM_STATUS_OWNED`, so a second `runCli` in one process is not mute. Tested.

packages/agency-lang/lib/cli/logsView.ts
egonSchiele: Deleted rather than fixed. Nothing had asked for the viewer to be a special case, and you are right that reaching for the singleton inside `createViewerHost` breaks the contract stated directly above it. A viewer that is open is a command in progress, and the spinner saying so is fine.

`stopProgress` had no other caller, so it is out of the API too. `logsView.ts` is back to its state on main.

packages/agency-lang/lib/cli/terminalStatus.ts
egonSchiele: Both bugs are real, and the handler they were in is gone — so rather than reordering the cleanup and keeping the listener, there is no listener. The reasoning is in the two threads above: any JS SIGINT listener defers the signal off its default disposition, which I measured breaking instant Ctrl-C on synchronous commands, and no handler can tell a pause from an ending anyway.

Your `once` observation is what convinced me the reordered version was still not right. `runSuite` registers with `once`, so the listener count is a moving target: the same code has to be correct across two presses with different sets of listeners present, and it has no way to know whether the first press meant "stop" or "fold this in and keep going". `exit` knows, and nothing else does.

On your last paragraph: exit 130 is no longer red. `commandFailed` treats 0, 130 and 143 as normal ways to stop.

packages/agency-lang/lib/cli/terminalStatus.ts
egonSchiele: Fixed — `write` is now `fs.writeSync(1, text)`.

You are right on the mechanics and I had the premise wrong in both the comment and the doc: Node documents TTY writes as asynchronous on Windows and synchronous only on POSIX, and this seam writes from an `exit` handler, so the bytes have to have landed. Since it writes exactly twice per command there is nothing to interleave badly with the buffered stream.

The re-raise path you mention is gone with the signal handlers, so that half no longer applies. New doc section, "Writes must be synchronous", records why this must not drift back to `process.stdout.write`.

packages/agency-lang/lib/cli/terminalStatus.ts
egonSchiele: Fixed, and this overrides the earlier thread that asked for unconditional `path.basename` — I have noted the conflict there.

Shortening now requires the operand to look like a path: a separator **and** no whitespace. Real paths on a command line rarely contain spaces; prompts nearly always do. So `agency agent "fix lib/parsers/foo.ts"` keeps its slashes, while `agency run tests/foo.agency` still shortens. Verified on a pty, both shapes. A slash-free prompt does land in the title whole, capped at 80 characters — which reads as `agency agent fix the parser bug`, and seems more useful than not.

The separator test is a character class covering both slashes, as you suggest.

On the spinner: decided to keep it, and written up in the doc under "Interactive commands keep spinning". The honest reason is that the parent cannot tell thinking from waiting — the work is in the spawned child — and closing that gap means new plumbing between the two processes for a cosmetic signal. The logs viewer's special case is deleted in the same round for the same reason, so the two are now consistent.

packages/agency-lang/lib/cli/terminalStatus.test.ts
egonSchiele: The coverage gap is closed by deletion: there are no signal handlers left to test. What replaced them is tested — that installing registers zero SIGINT/SIGTERM listeners (the property that keeps Ctrl-C working), that repeated installs add at most one `exit` listener, that every program still gets its hook, and `commandFailed` per exit code.

I did not take the `listenerCount`/`kill` seam suggestion, since with no handler there is nothing behind it. Had the handlers stayed it would have been the right call.

The `afterEach` scrubber is gone too — with the listener latched to one per process there is nothing left to leak.

packages/agency-lang/docs/dev/terminal-status.md
egonSchiele: Re-read against the code and rewritten. The section is now "Signals, and why there are none", and it describes what the code does because the code no longer does the thing the old paragraph was describing.

Line 70 fixed as well: the synchronous-write premise was wrong, `write` is now `fs.writeSync(1, …)`, and there is a section on why.

packages/agency-lang/lib/cli/terminalStatus.ts
```diff
@@ -0,0 +1,190 @@
...
+ * The process-wide instance the CLI wires up.
+ *
+ * `fs.writeSync` rather than `process.stdout.write`, because `end` runs from an
+ * `exit` handler and so must land before the process is gone. Node's stdout is
+ * synchronous to a TTY on POSIX but asynchronous to one on Windows, where the
+ * write would be dropped and the tab left named and spinning.
+ */
+export const terminalStatus: TerminalStatus = createTerminalStatus({
+  write: (text) => {
+    fs.writeSync(1, text);
```
egonSchiele, line 111: Swapping to `fs.writeSync` fixes the Windows flush, but it also changes the failure mode: `process.stdout.write` never throws — the stream emits `error`, and Node's stdout has a built-in handler that swallows `EPIPE` — whereas `fs.writeSync` throws synchronously. That throw is unguarded, and it happens inside the `exit` listener at line 189.

Concrete scenario: an `agency eval run` over SSH, or in a terminal window the user closes while it works. stdout was a TTY at `begin`, so `claimed` is true. By the time the process exits, the pty is gone and the write to fd 1 fails with `EIO` (or `EPIPE` if it was a pty pair). `fs.writeSync` throws out of the `exit` handler, so a clean run ends with an uncaught-exception trace and a changed exit code — caused entirely by the cosmetic tab cleanup. `EAGAIN` on a non-blocking shared tty fd is the same shape, less common.

Cheap fix — nothing here is worth failing a run over:

```suggestion
  write: (text) => {
    try {
      fs.writeSync(1, text);
    } catch {
      // The tab is cosmetic. A vanished terminal (EIO/EPIPE) or a busy
      // non-blocking fd (EAGAIN) must not take the process down, least of
      // all from the `exit` handler.
    }
  },
```

packages/agency-lang/lib/cli/terminalStatus.ts
```diff
@@ -0,0 +1,190 @@
...
+ * Wire the tab status to a CLI program: every command names the tab while it
+ * runs, and gives the name back on the way out.
+ *
+ * Cleanup hangs off `exit` alone, with no `SIGINT` handler of its own. That is
+ * deliberate. Registering any JS signal listener takes SIGINT off its default
+ * disposition, where the kernel ends the process at once, and defers it to the
+ * event loop — so `agency compile` on a big file, which is synchronous and is
+ * the case where Ctrl-C responsiveness matters most, would stop dying on the
+ * first Ctrl-C. A stranded spinner is a far smaller price, and it clears itself
+ * the next time any Agency command runs in that tab.
```
egonSchiele: I agree with dropping the signal handlers — that was the right call, and the reasoning here is sound. One correction to the last sentence, though, because the consequence is bigger than the comment says.

When no other listener is registered — `agency compile big.agency`, `agency ast`, `agency typecheck` — Ctrl-C kills the process at its default disposition, so Node's `exit` event never fires and `end` never runs. The spinner does clear itself on the next Agency command, as claimed. The **title** does not: `begin` already wrote `CSI 22;0t` (push) and the matching `CSI 23;0t` (pop) is only ever written by `end`. So the tab keeps the dead command's name once you are back at the shell prompt, and each Ctrl-C'd command leaves another unbalanced entry on the terminal's title stack. Ctrl-C on a slow compile is exactly the case the no-signal-handler decision is optimising for, so it is also the case that strands a title.

Two ways out, either fine:
- Say so in the comment and in `docs/dev/terminal-status.md`, since the current wording ("clears itself the next time any Agency command runs in that tab") reads as if nothing is left behind.
- Or drop the push/pop pair and set the title outright at `begin`, restoring nothing at `end`. Then a killed command and a clean one leave the tab in the same state, with no stack to unbalance.

packages/agency-lang/lib/cli/terminalStatus.ts
egonSchiele: Owed you a reply on this one — sorry, I acted on it but skipped the thread.

Taken, then partly walked back by your later comment on `agency agent`, which arrived after this one. Unconditional `path.basename` mangles a prompt: `agency agent "fix lib/parsers/foo.ts"` became `agency agent foo.ts`.

So shortening now requires the operand to look like a path — a separator **and** no whitespace — and the separator test is `/[/\\]/` rather than `path.sep`, which is the fix you wanted here. `path.basename` stays the per-platform one, so it shortens a backslash path on Windows and leaves it alone on POSIX, where a backslash is a legal filename character. Verified both shapes on a pty.

packages/agency-lang/lib/cli/terminalStatus.ts
egonSchiele: Fixed with your suggestion. The failure-mode change was real and I had not thought about it: `process.stdout.write` never throws, `fs.writeSync` does, and the throw was sitting inside the `exit` handler.

The write now lives in an exported `writeToTerminal`, so the guard is testable — EIO, EPIPE and EAGAIN each assert it does not throw, plus one that drives `begin`/`end` through it. Mutation-checked: deleting the `try` turns those four red.

On the empty catch, which the repo's anti-patterns list otherwise forbids: logging is not available here, because the failure *is* that the terminal is unreachable. `console.error` goes to fd 2, which in this scenario is the same dead pty. The comment says so, and so does the doc, since this is exactly the kind of thing a future reader would "fix".

packages/agency-lang/lib/cli/terminalStatus.ts
egonSchiele: You are right, and my wording was quietly generous to itself — I wrote "stranded spinner" when the title is the half that actually persists. `begin` pushes and only `end` pops, so a hard Ctrl-C keeps the dead command's name at the shell prompt and leaves one unbalanced stack entry behind, every time.

I took the first option: say so plainly, in both the code comment and `docs/dev/terminal-status.md`, including that the stack entries accumulate to the terminal's cap rather than without bound.

Not the second, and the reasoning is worth stating since you left it open. Dropping the push/pop pair would make *every* command leave its name in the tab — trading a rare stranded title for a permanent one, in a repo where the complaint that started this was a tab showing a name that had nothing to do with what was running. Restoring the previous title is the only thing the stack is for, and it works in the case that covers nearly every command run. The commands you are most likely to interrupt — the eval runner, `compile --watch`, the log viewer — all handle their own signal and leave through `process.exit`, so they strand nothing at all.

Issue #690, opened 2026-07-25
agency run: add --only-stdlib to restrict imports to checkable Agency code


egonSchiele:
## Summary

A blunt, user-chosen version of a check that was built and then removed from the splices PR (#679).

Add `--only-stdlib` to `agency run`, restricting a program's transitive import graph to `std::` modules and relative `.agency` files. No npm packages, no TypeScript imports, no `pkg::`.

## Why

Agency's safety story rests on dangerous operations raising interrupts. That only holds for Agency code. TypeScript and JavaScript raise nothing, and a plain npm package passes through untouched when imported (`docs/dev/pkg-imports.md`). So any program that can reach JavaScript can do anything, with nothing to check.

`--only-stdlib` lets a user say "this program stays inside the part of the language that is checkable".

## Background from #679

The splices PR originally enforced exactly this rule on generators, as `AG8006`, with a transitive walk over the import graph. The implementation is worth reading before rebuilding it:

- It must be **transitive**. A generator importing a clean-looking local `.agency` file that itself imports `zod` is the case that matters; the generator's own imports look spotless.
- It must route through `agencyImportTarget` in `lib/compiler/compileClosure.ts`, which also recognises `exportFromStatement`. An import-only scan lets `export { x } from "zod"` escape.
- `pkg::` counts as leaving Agency: it is Agency source but can reach JavaScript one level down.

That code was removed when the PR was simplified, so it is in the history of #679 rather than on main. The fixture `tests/agency/splices/refuseNonAgency.agency` was the transitive test case.

## Notes

Worth considering whether this belongs on `compile` too, and whether it should be expressible in `agency.json` rather than only as a flag.

egonSchiele:
Fixed with the new `--agency-only` flag

PR #910, opened 2026-08-24
--refuse-splices: decline compile-time generator execution (and stop type checking from running generators)

### LINKED ISSUE #689: agency run/compile: add --no-splices to refuse compile-time generator execution

egonSchiele:
Closes #689.

Two things in one PR, as separate commits: a soundness fix, then the flag that exposes it.

## What #689 asked for, and what was already done

Worth stating up front, because most of the issue is already shipped — it is a near-duplicate of #690, which was closed with "Fixed with the new `--agency-only` flag":

- **The stdlib half is done**, and stricter than proposed. `std::agency compile` refuses splices unconditionally through the closure validator. (`run` was the wrong place to ask for it anyway: it takes an already-compiled program, so splices ran earlier.)
- **`agency run --agency-only`** already refuses splices; its help text names them.

What was left is the narrow option. `--agency-only` is a bundle that also refuses TypeScript, so a project that imports TS on purpose could not use it, and `compile` and `typecheck` had no way to decline at all.

## What this does NOT do (changed in review)

An earlier version of this PR also forced the refusal on for the `std::agency` inspection entry points, on the grounds that type checking runs generators and that path has no sandbox. The owner asked whether there is actually a security reason. Having gone looking, **no** — and `docs/dev/language/splices.md` is what settles it.

Compilation installs no interrupt handlers, so an effectful generator cannot complete. The one exfiltration route that mattered — read an env var, bake it into emitted JavaScript, commit the artifact — is closed twice: `env` now raises `std::env` (#904), and `CHILD_ENV_ALLOWED` gives the generator subprocess only what Node needs to boot. Both verified in code. That doc calibrates the rest as "closer to the npm `postinstall` problem than to a new hole".

The fallback argument was consistency with `std::agency compile`, which refuses splices unconditionally. That does not hold either: `compile`'s refusal is *structural* — `compileSandboxed` mirrors and compiles a validated closure — not a judgement that splices are unsafe. And `typecheckFile` already gates on an `std::read` interrupt.

So `stdlib/agency.agency` and `lib/stdlib/agency.ts` are byte-identical to main, and this PR is now exactly what #689 asked for.

It is still true that type checking runs generators and that `runCheckerPipeline` hardcoded `{}` so nothing could decline. That plumbing stays, because it is what makes `agency typecheck --refuse-splices` work at all.

## What this does

**Commit 1 — the mechanism.** `refuseSplices` on `AgencyConfig`, the `AG8016` refusal, and config threaded through the checker pipeline so a caller can decline there.

**Commit 2 — the flag.** `--refuse-splices` on `compile`, `run`, `typecheck`, `test run`, routed through `applyCliFlags`. Off by default.

Named `--refuse-splices`, not the `--no-splices` the issue proposed: the vendored commander treats any `--no-` long flag as a negation, so `--no-splices` would arrive as `opts.splices` defaulting to `true` — inverted from every other safety flag here and from the config field it sets.

## Two placement decisions I want reviewed

**The refusal sits before generator resolution**, at the top of `expandSplices`, rather than as one of the `CHECKS` in `decide`. That is a deliberate exception to the "add a rule, never edit the pass" convention in `docs/dev/splices.md`. Resolution parses the generator's module — the work the refusal exists to avoid — and it can fail on its own, so a file with both a broken import and the setting on would report `AG8005` in preference to the refusal. `calleeName` reads the callee off the syntax, so the message still names the generator without opening its file.

**On the checker pipeline a refusal throws** rather than becoming a diagnostic. That pipeline otherwise tolerates a splice that will not expand, keeping the unexpanded program — but a refusal is the caller saying "do not run this", and continuing would answer as though the file had no splice, reporting every generated name as undefined. Throwing joins the pipeline's existing rule that a throw means "could not check this", which the `Result`-returning stdlib entry points surface as an ordinary failure. Every other splice failure keeps the tolerant behaviour.

## Verification

- `npx tsc --noEmit`, `pnpm run lint:structure`, `pnpm run fmt:ts` — clean
- 180 tests pass across `expandSplices`, the new `typecheckRefusesSplices`, `config`, `closureValidator`, `compileSandboxed`, `compile`, `diagnostics`, `diagnosticExplanations`, `explain`, `commandLine`, `sourceIsText`
- I ran `make` so the generator-executing tests could actually run: all 36 in `expandSplices.test.ts` pass. Without a built `dist` those are red for environmental reasons, which is worth knowing before reading a fresh worktree

Both halves are mutation-tested. Removing the stdlib opt-in fails the soundness tests; disabling the refusal fails the expansion tests. The refusal tests deliberately do not write a generator file, so if the refusal ever moves after resolution they fail with `AG8005` instead of `AG8016`.

`docs/site/diagnostics/*` is regenerated by `make`, not hand-edited — the diff is exactly the new diagnostic.

## Follow-up, agreed separately

Extending `--agency-only` to `compile`, `pack` and `serve` is a separate PR. It is not a modifier but a whole alternative compile pipeline, so it only fits commands that compile — which is also why it does not belong on `typecheck`.

https://claude.ai/code/session_01QD2WiV5DAqLYC9i1f3FLtn


packages/agency-lang/scripts/agency.ts
```diff
@@ -1257,10 +1266,11 @@ export function createProgram(deps: CliDependencies = {}): Command {
           maxCost?: string;
           maxTime?: string;
           agencyOnly?: boolean;
+          refuseSplices?: boolean;
           json?: boolean;
         },
       ) => {
-        const config = getConfig();
+        const config = applyCliFlags(getConfig(), { refuseSplices: opts.refuseSplices });
```
egonSchiele, line 1273: This makes the flag authoritative here, but it stops being authoritative one call later. `precompileTestSources` → `groupTestSources` re-merges a per-directory config over the base (`lib/cli/precompile.ts:101`):

```ts
config = { ...baseConfig, ...loadConfig(localConfigPath) };
```

An `agency.json` next to a `.test.json` with `"refuseSplices": false` therefore beats `--refuse-splices`, and the generator runs. The top-level schema is `.partial()`, so a local config that never mentions the key is fine — it takes an explicit `false` — but planting one is trivial in the "freshly cloned repository" case this feature is documented for.

`run`, `compile` and `typecheck` are all safe, because `applyCliFlags` runs last and only ever sets `true`. Restoring that ordering for the test path fixes it:

```ts
config = applyCliFlags(
  { ...baseConfig, ...loadConfig(localConfigPath) },
  { refuseSplices: baseConfig.refuseSplices },
);
```

A test for it would be worth having, since this is the only place the property can be lost.

egonSchiele (commented):
The soundness finding is real and the fix is in the right place. I reproduced the reasoning rather than trusting it: `runCheckerPipeline` did call `expandSplices` with a hardcoded `{}`, `_typecheckFile` does hand it a real path, and that path genuinely does not pass through the closure validator. Putting the refusal above `inProgress[key] = true` and above resolution is the correct spot, and the tests that deliberately omit the generator file are a good way to pin it there.

I also checked the coverage of the opt-in, which is the part that would be easy to get wrong. `_typecheck`, `_getEffects`, `_typecheckFile` and `describeSource` are the four stdlib paths that reach the checker, and all four now pass `INSPECT_UNTRUSTED`; `_formatFile` only parses and generates, so it is fine. On the CLI side, `--refuse-splices` really does reach `run` through the `{ ...options }` spread in `runWithOptions`, and `addRunOptions` plus `fallbackCommand("run")` means the bare `agency foo.agency` shorthand gets it too. Naming it `--refuse-splices` over `--no-splices` is right for the reason given.

**One thing I would fix before merge: on `test run`, the flag is revocable by the repository being inspected.**

`applyCliFlags` runs last for `run`, `compile` and `typecheck`, and only ever sets the field to `true`, so the flag is authoritative there — an `agency.json` cannot undo it. `test run` breaks that. `groupTestSources` re-merges a local config after the fact (`lib/cli/precompile.ts:101`):

```ts
config = { ...baseConfig, ...loadConfig(localConfigPath) };
```

So an `agency.json` sitting next to a `.test.json` in the tree you are inspecting, containing `"refuseSplices": false`, wins over the command line, and the generator runs. The schema accepts the key (`z.boolean()` inside the top-level `.partial()`), and since it is `.partial()`, a file that simply does not mention it is harmless — it takes a deliberate `false`. That is not much of an obstacle in the stated threat model, where the whole point is that you have not read this tree yet.

The general rule worth writing down: a refusal chosen on the command line must not be revocable by a file inside the tree under inspection. Re-applying the flag after the local merge is enough, and it is one line.

Three smaller things, inline: the diagnostic gives advice the one forced-on caller cannot follow; the comment justifying the `getEffectsFromFile` hole names a caller that does not exist; and the test carrying the "this is not a blanket ban" claim does not actually prove expansion happened.

On the two placement decisions you flagged for review: I agree with both. The exception to "add a rule, never edit the pass" is well argued and the doc records it, and throwing on the checker pipeline is right — a caller who said "do not run this" is not served by a report full of undefined names. The `docs/dev/splices.md` addition is the good kind of dev doc.

Smaller notes not worth their own comment: the hidden `agency interrupts` reaches `expandSplices` (via `analyzeInterrupts`) with no flag to decline, which is defensible for a hidden command but leaves `agency.json` as the only lever there; and the new `docs/dev/config.md` paragraphs are unwrapped while the rest of that file wraps at about 78 columns.

packages/agency-lang/lib/typeChecker/diagnostics.ts
```diff
@@ -667,6 +667,12 @@ export const DIAGNOSTICS = {
     message:
       "The splice argument `{name}` is declared in this file, so it does not exist yet when the generator runs. Splice arguments may be literals, code literals, or imported names.",
   },
+  spliceRefused: {
+    code: "AG8016",
+    severity: "error",
+    message:
+      "`{file}` contains a splice that would run the generator `{name}` at compile time, and generator execution was declined. Compile again without `--refuse-splices` (or with `refuseSplices` off) to allow it.",
```
egonSchiele: "Compile again without `--refuse-splices` (or with `refuseSplices` off)" is advice the one caller that always sees this message cannot take. The stdlib inspection entry points force the setting on through `INSPECT_UNTRUSTED`, so an agent calling `typecheckFile` gets told to remove a flag it never passed and cannot remove.

Since the same string serves both audiences, I would drop the remedy from the message and leave it in the explanation, or phrase it as a statement of what happened rather than an instruction — "generator execution was declined for this check" already tells a CLI user exactly which flag to drop, because they typed it. `diagnosticExplanations.ts` has room for the fuller version and already carries the sandboxed-compilation caveat; it could use the same caveat for the inspection path.

packages/agency-lang/lib/compiler/typecheck.ts
```diff
@@ -175,20 +198,32 @@ export function getEffectsFromSource(source: string): EffectsByExport {
  */
 export function getEffectsFromFile(filePath: string): EffectsByExport {
   const absolute = path.resolve(filePath);
-  return effectsVia(fs.readFileSync(absolute, "utf-8"), absolute);
+  // No refusal here on purpose: splice eligibility calls this to read a
+  // GENERATOR's effects, and a generator that itself splices is already
+  // refused as AG8009 (spliceNested).
```
egonSchiele: The reason given here is not the case. Splice eligibility does not call `getEffectsFromFile` — it reads `symbol.interruptEffects` off the symbol table in `lib/compiler/splice/generatorEffects.ts:52`. `getEffectsFromFile` has no production callers at all: the only references in the repo are `lib/analysis/effects.test.ts`, `lib/analysis/effectsOracle.test.ts` and `lib/compiler/typecheckEffects.test.ts`.

So this is an exported function that runs generators against a real on-disk path, kept that way for a caller that does not exist, sitting next to three siblings that just gained a way to decline. The next caller inherits the hole silently. Giving it the same `config` parameter (defaulted, so the tests do not change) costs nothing; if it really is test-only, saying that is better than the current explanation, and better still would be asking whether it should stay exported.

Worth checking whether it is re-exported from the package's public API, which would change the answer.

packages/agency-lang/lib/compiler/typecheckRefusesSplices.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    expect(report.errors.map((e) => e.code)).not.toContain("AG8016");
+  });
+
+  it("the pipeline itself still expands when the caller does not decline", () => {
+    // The refusal is a caller's choice, not a property of the checker: the
+    // same file, checked without the setting, expands and reports no AG8016.
+    // This is what stops the fix from becoming a blanket ban.
+    const hostPath = path.join(dir, "host.agency");
+    const report = typeCheckSource(fs.readFileSync(hostPath, "utf-8"), hostPath);
+    expect(report.errors.map((e) => e.code)).not.toContain("AG8016");
```
egonSchiele: This is the test carrying the "this is what stops the fix from becoming a blanket ban" claim, and `not.toContain("AG8016")` does not carry it. Nothing here would tell you whether the splice expanded — only that this particular code is absent, which is true of almost every outcome.

The fixture is already set up to bite. `host.agency` calls `print(greet())`, and `greet` exists only if the generator ran, so an unexpanded program reports it as an undefined variable:

```ts
expect(report.errors).toHaveLength(0);
```

That fails if expansion silently stops happening, which is the regression this test is for. The same swap would strengthen "a file with no splice still checks normally" on line 60.

packages/agency-lang/lib/compiler/typecheckRefusesSplices.test.ts
```diff
@@ -0,0 +1,71 @@
...
+ * These tests pin the boundary: the agent-reachable entry points decline
+ * generator execution, and the underlying pipeline still allows it when a
+ * caller explicitly asks.
+ */
+let dir: string;
+
+beforeEach(() => {
+  dir = path.join(process.cwd(), ".agency-tmp", `tc-refuse-${nanoid()}`);
+  fs.mkdirSync(dir, { recursive: true });
+  // A generator that writes a file when it runs. Nothing should create it.
```
egonSchiele: "A generator that writes a file when it runs. Nothing should create it." — it does not write a file. It returns a `Code` fragment declaring `greet`, and no test looks for a side-effect witness.

That matters because a reader takes this comment as the proof that nothing executed, when the actual proof is the `AG8016` throw. And by this PR's own argument the described test could not work anyway: compilation installs no interrupt handlers, so a generator that tried to write a file could not finish. Deleting the second sentence, or replacing it with what the fixture really is, keeps the file honest.

packages/agency-lang/scripts/agency.ts
egonSchiele: Confirmed and fixed in 1758afea. This was the real bug in the PR — thank you.

I verified both merge sites: `lib/cli/precompile.ts` (grouping) and `lib/cli/test.ts` (runTestFile) each merge a fixture-local agency.json over the base, so `"refuseSplices": false` beat the flag. Both now re-apply the CLI intent afterwards, exactly as you wrote it.

Two tests, in `precompile.test.ts` where `groupTestSources` is already exercised: a local config **cannot** cancel the flag, and — the other half of the property — a local config can still opt IN when no flag was passed. I reverted the fix to confirm the first one fails; it does.

Also fixed the related path Copilot found on the same thread: `expectedCompileError` shells out to a child `agency compile` that re-derives its own config, so the refusal is now forwarded to it explicitly.

packages/agency-lang/lib/typeChecker/diagnostics.ts
egonSchiele: Agreed, and taken the way you suggested. The message is now a statement of what happened:

```
`{file}` contains a splice, and generator execution was declined for this check, so `{name}` was not run.
```

No instruction, so it reads correctly for both audiences — a CLI user knows which flag they typed, and an agent calling `typecheckFile` is not told to remove something it never passed.

The remedy moved into `diagnosticExplanations.ts`, which now carries both unconditional cases: sandboxed compilation, and the stdlib inspection entry points, with the note that neither has a flag to drop. Fixed in 1758afea.

packages/agency-lang/lib/compiler/typecheck.ts
egonSchiele: You are right on every point, and I checked each one rather than taking it.

`getEffectsFromFile` has no production callers — only `lib/analysis/effects.test.ts`, `effectsOracle.test.ts` and `lib/compiler/typecheckEffects.test.ts`. Splice eligibility reads `symbol.interruptEffects` off the symbol table in `compiler/splice/generatorEffects.ts:52`, so my stated reason was simply wrong. It is also not re-exported from the package's public API, which I checked because you asked.

It now takes the same defaulted `config` parameter as its siblings — the tests are untouched — and the comment says what it actually is: test-only today, defaulted so the next production caller does not inherit the hole silently.

On whether it should stay exported: I have left it, because removing an export is a wider change than this PR, but I agree it is worth asking and am happy to do it as a follow-up. Fixed in 1758afea.

packages/agency-lang/lib/compiler/typecheckRefusesSplices.test.ts
egonSchiele: Right, and the swap works — which makes the point sharper than my version did.

`expect(report.errors).toHaveLength(0)` passes today and fails if expansion stops happening, because `greet` resolves only when the generator ran. Applied to both that test and "a file with no splice still checks normally" on line 60, where a clean report also says something stronger: the setting costs a splice-free file nothing. Fixed in 1758afea.

packages/agency-lang/lib/compiler/typecheckRefusesSplices.test.ts
egonSchiele: Correct — the generator returns a `Code` fragment, it does not write anything, and nothing looks for a side-effect witness. Worse, as you say, the test I described could not work: with no interrupt handlers installed, a generator trying to write a file could not finish.

Replaced with what the fixture actually is: the fragment declares `greet`, and `host.agency` calls it, so `greet` resolving IS the witness that the generator ran. That is now also what the strengthened assertion on line 69 tests. Fixed in 1758afea.

packages/agency-lang/scripts/agency.ts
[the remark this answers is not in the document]
egonSchiele: Both paths confirmed and fixed in 1758afea. The config re-merge in `precompile.ts` and `test.ts` now re-applies the CLI intent after the local merge, so a fixture-local `"refuseSplices": false` can no longer cancel the flag; and `compileInSubprocess` now forwards `--refuse-splices` to the child `agency compile`, since that process re-derives its own config. Regression tests added in `precompile.test.ts`, verified by reverting the fix.

packages/agency-lang/lib/typeChecker/diagnostics.ts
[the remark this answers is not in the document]
egonSchiele: Fixed in 1758afea. The message no longer gives an instruction — it states what happened ("generator execution was declined for this check, so `{name}` was not run"), which reads correctly whether the caller passed a flag or hit a path that forces the setting on. The remedy and both unconditional cases now live in the explanation instead.

packages/agency-lang/lib/typeChecker/diagnosticExplanations.ts
```diff
@@ -531,6 +531,14 @@ The rule covers everything a generator can reach, not just what it imports direc
 
 **How to fix:** move the work into Agency, or set \`allowNonAgencyGenerators: true\` in your config if the generator genuinely needs a JavaScript library. Turning it off means the generator can do whatever that library can.`,
 
+  spliceRefused: `This file contains a splice, and generator execution was declined for this compile.
+
+Compiling a \`$( ... )\` runs its generator right then, during compilation. So compiling a file you have not read means running code you have not read. That is bounded — a generator may import only \`std::\` modules and other \`.agency\` files, and compilation installs no interrupt handlers, so anything dangerous cannot finish — but it is still execution, and you may prefer to decline it. Inspecting a repository you just cloned is the usual reason.
+
+Nothing ran: the refusal happens before the generator is resolved, so its file was never opened.
```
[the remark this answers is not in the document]
egonSchiele: Correct, and fixed in 1758afea. I verified the ordering: `SymbolTable.build` runs before `expandSplices` in `runCheckerPipeline`, and the build session constructs the compiled closure before expansion — both crawl and parse imports, so the generator's file may well have been read.

Narrowed to what is actually guaranteed: the generator is never compiled or executed. `docs/site/diagnostics/` is regenerated.

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -185,12 +186,23 @@ export function _subprocessDepth(): number {
...
+const INSPECT_UNTRUSTED: AgencyConfig = { refuseSplices: true };
+
 export function _typecheck(source: string): TypeCheckReport {
-  return typeCheckSource(source);
+  return typeCheckSource(source, undefined, INSPECT_UNTRUSTED);
 }
 
 export function _getEffects(source: string): Record<string, string[]> {
-  return getEffectsFromSource(source);
+  return getEffectsFromSource(source, INSPECT_UNTRUSTED);
```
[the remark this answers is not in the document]
egonSchiele: Agreed — this is a public contract change and the docstrings were stale. Fixed in 1758afea.

`typecheck`, `typecheckFile`, `getEffects` and `describe` in `stdlib/agency.agency` now each say that splice-bearing source fails rather than being inspected, and why. `getEffects` was the most misleading, since it said generated code was merely "not visible"; that line stays true and now has the refusal beside it. `docs/site/stdlib/agency.md` regenerated via `make`.

packages/agency-lang/docs/dev/runtime/config.md
```diff
@@ -56,6 +56,28 @@ The two meanings are two fields on `CliFlags` (`strict` and `strictTypes`), not
 a branch inside `applyCliFlags`. The command picks the meaning it wants by
 choosing a field; the helper never needs to know which command called it.
 
+### `--refuse-splices`
```
egonSchiele, line 59: Please shorten your addition to this doc. We just spent a long time cleaning these docks up. This specific doc is for the config, and now the biggest section in it is for this flag. When you think about the config, do you think the as the most important section in it?

packages/agency-lang/docs/site/diagnostics/templates.md
```diff
@@ -194,3 +194,19 @@ A template is checked before anything fills it, so a hole is opaque: the checker
 This also covers names from the file the template is written in. A template becomes its own program — `toSource` prints it and `runCode` compiles the print — and the surrounding file is not there when that happens.
 
 **How to fix:** keep each fragment self-contained. Move the code that defines the name into the template, or move the code that uses it into the fragment that defines it. A template may always use the prelude, language builtins, and anything it declares or imports itself.
+
+<a id="ag8016"></a>
```
egonSchiele, line 198: I'm just curious, where does this diagnostic show up? Does it show up in a place where it is useful to show the user five paragraphs of text?

packages/agency-lang/lib/preprocessors/expandSplices.ts
```diff
@@ -83,6 +83,28 @@ export function expandSplices(
...
+  // after resolution, which parses the generator's module and would report
+  // an unrelated import problem (AG8005) in preference to this refusal.
+  // `calleeName` reads the callee straight off the syntax, so the message
+  // still names the generator without opening its file.
+  if (config.refuseSplices === true) {
+    const splice = splices[0];
+    return {
+      ok: false,
+      diagnostic: {
+        diagnostic: "spliceRefused",
```
egonSchiele, line 98: Should this come from a constant?

packages/agency-lang/lib/stdlib/agency.ts
```diff
@@ -187,12 +188,23 @@ export function _subprocessDepth(): number {
...
+/**
+ * The trust boundary for agent-reachable inspection.
+ *
+ * These entry points read source the agent supplied or found, and type
+ * checking RUNS a splice's generator (`_typecheckFile` hands the checker a
+ * real path, so a splice resolves its generator against that directory).
+ * `compile` already refuses splices unconditionally through the closure
+ * validator; inspection is the same boundary and gets the same answer.
+ */
+const INSPECT_UNTRUSTED: AgencyConfig = { refuseSplices: true };
```
egonSchiele: 1. Why is this a constant?
2. If it is about splices, why does the var name not mention splices? Is it because we think there will be other things to add to this config object in the future?

packages/agency-lang/stdlib/agency.agency
```diff
@@ -630,6 +630,10 @@ export idempotent def typecheck(
   """
   Type-check Agency source code.
 
+  Source containing a compile-time splice `$( ... )` fails rather than
```
egonSchiele: undo all changes to this file, and let's think about this behavior. Is there a security reason that we wouldn't want to run splices?

egonSchiele: Reverted in 03dde0f1 — `stdlib/agency.agency` is byte-identical to main again, and so is `lib/stdlib/agency.ts`.

On the question: **no, I do not think there is one**, and your own docs are what convinced me.

`docs/dev/language/splices.md` makes the argument. Compilation installs no interrupt handlers, so an effectful generator cannot complete. The one exfiltration route that actually mattered — read an env var, bake it into the emitted JavaScript, commit the artifact — is now closed twice: `env` raises `std::env` (#904, just merged), and `CHILD_ENV_ALLOWED` in `runGenerator.ts` hands the generator subprocess only what Node needs to boot, with stdin a pipe so `readStdin` gets EOF. I checked both are real in code, not just described. That section then calibrates the remainder as "closer to the npm `postinstall` problem than to a new hole", and notes npm, Template Haskell and Rust proc macros check nothing at all.

What I was actually reasoning from was consistency: `std::agency compile` refuses splices unconditionally, so why not `typecheckFile`? That does not hold up either. `compile`'s refusal is **structural** — `compileSandboxed` mirrors and compiles a validated closure, so an unexpanded splice would run untrusted code outside the sandbox it is building. It is not a judgement that splices are dangerous. And `typecheckFile` already gates on an `std::read` interrupt, so the user approved reading that file before anything happened.

I overstated it as a soundness fix. It was really "these two APIs disagree", which is a much weaker claim and not worth changing a public contract for.

What stays is the flag, which is what #689 asked for: `--refuse-splices` for a caller who would rather decline than rely on the argument above. That still needs config to reach `runCheckerPipeline` (it hardcoded `{}`), so `agency typecheck --refuse-splices` works at all — the pipeline test now covers the flag rather than the removed behavior.

Happy to restore the forced-on version if you disagree; it is one named constant and four call sites.

packages/agency-lang/lib/stdlib/agency.ts
egonSchiele: Both fair, and the constant is gone in 03dde0f1 along with the behavior it carried (see the `stdlib/agency.agency` thread).

Answering anyway, since the questions were good ones. It was a constant so the reason lived in one place rather than being repeated at four call sites — but that is a comment's job, not a binding's. And you are right about the name: I called it `INSPECT_UNTRUSTED` because I expected the object to grow, which is exactly the speculative-generality reasoning that produces a name nobody can map back to what it does. If it had stayed, `REFUSE_SPLICES` would have been the honest name.

packages/agency-lang/lib/preprocessors/expandSplices.ts
egonSchiele: It already matches the convention in this file — `spliceNested` (line 80), `spliceGeneratorNotImported` (258), `spliceArgumentNotAvailable` (318) and `spliceRedeclaresHostName` (362) are all written the same way.

They are also type-checked rather than free strings: the field is `DiagnosticName`, which is `keyof typeof DIAGNOSTICS`, so a typo fails to compile and a diagnostic that does not exist cannot be named. `spliceDiagnostics.test.ts` additionally pins the params each one must supply.

So I have left it as-is — happy to change it if you meant something else by the question.

packages/agency-lang/docs/site/diagnostics/templates.md
egonSchiele: It shows up in two places, both of which you have to ask for: `agency explain AG8016` (`lib/cli/explain.ts`) and the generated diagnostics pages (`lib/cli/diagnosticsDocs.ts`, which is what produced the file you are commenting on). It is never printed inline with the error — that is the one-line message in `diagnostics.ts`.

So it is opt-in, but five paragraphs was still indefensible. I measured: 116 explanations, median 53 words, previous longest 158 (`spliceGeneratorUnreadable`). Mine was **234 — the longest in the repo by 50%**.

Cut to 92 in 03dde0f1: what happened, why anyone would want it, the one caveat that the file may still have been read, and the remedy. The reasoning about why the refusal sits before resolution moved to `docs/dev/language/splices.md`, which is where a maintainer looks and a user does not.

packages/agency-lang/docs/dev/runtime/config.md
egonSchiele: You are right, and no — nobody thinks about `--refuse-splices` when they think about config.

Cut from ~180 words to ~85 in 03dde0f1. What is left is the two-sentence definition plus the one thing a future flag author genuinely needs from this doc rather than the splices doc: because it is config, a later merge over the CLI-derived config can cancel it, which is how `test run` beat this flag. Everything else moved to `docs/dev/language/splices.md`.

Sorry for adding weight to a doc you had just finished cleaning up.

packages/agency-lang/lib/compiler/typecheckRefusesSplices.test.ts
egonSchiele: Correction: I told you this swap works. It does not, and I should have checked before saying so.

`expect(report.errors).toHaveLength(0)` passes — but it also passes with expansion disabled entirely. I mutated `runCheckerPipeline` to never expand, and the test stayed green. A call to a function that does not exist is not reported by this checker configuration, so a generated *function* proves nothing: `print(greet())` type-checks clean either way. The assertion was measuring nothing, exactly like the `not.toContain` version it replaced.

The fixture now generates a **type** as well, and the host annotates with it. An unknown type alias is a real error (AG1006), so with expansion disabled the test fails as intended. Verified both directions — clean when expanded, AG1006 when not. Fixed in 0e7fb904.

Your underlying point stands and was right: the fixture should carry the proof. It just needed a witness the checker actually reports.

PR #914, opened 2026-08-24
Bring docs/dev back in sync with the code

egonSchiele:
Second pass of the incremental `docs/dev` cleanup, after #913 grouped the files. This one is about **content**: all 103 developer docs were verified against the current source, doc by doc, and corrected.

**102 docs changed, +3,400 / -2,787.** Net growth is small, so this is correction rather than expansion.

## The drift was not cosmetic

Docs that documented APIs which no longer exist:

| Doc | What it documented | Reality |
|---|---|---|
| `runtime/rewind.md` | `onCheckpoint`, `RewindCheckpoint`, a checkpoint sentinel after every LLM call | None exist. 3 of its 8 "key files" were gone. Rewritten around the real `rewindFrom`. |
| `runtime/interrupts.md` | `TsIfSteps`/`TsThreadSteps`/`TsWhileSteps`/`TsForSteps` + 5 templates; `approveInterrupt`/`rejectInterrupt`/`modifyInterrupt`/`resolveInterrupt` | All deleted. Codegen moved to `Runner`; the API is `respondToInterrupts`. |
| `runtime/config.md` | `restrictImports`; `strictTypes`/`typeCheck`/`typeCheckStrict` as top-level keys | `restrictImports` is nowhere in the tree; those three live under `typechecker.*`. ~20 real options were undocumented. |
| `contributing/coding-standards.md` | max-depth 4, function 100 lines, file 1000 lines; "use `type` not `interface`" and "objects not `Map`" as enforced | 5 / 150 / 1250. Both those rules are **commented out** in `eslint.config.js`, so they are conventions. |
| `compiler/typechecker/README.md` | the `"any"` string sentinel, `>>> "..."` prompt syntax, a builtins table | Sentinel retired (`ANY_T`), prompt syntax gone, table entirely stale. |
| `runtime/subprocess-ipc.md` | `run(compiled, {node, args})`, `{type:"telemetry", costUsd}` | Flat params; telemetry replaced by `invocationUsage`. |

**A default silently flipped:** the undefined-function diagnostic now defaults to `warn`, not `silent`.

**Work documented as pending that had already shipped:** parallel blocks v1 (the doc said "not yet implemented" about the design the compiler actually runs), trace bundles, null narrowing, per-request observability, and most of the debugger backlog. Conversely, documented *limitations* that are fixed — e.g. a handler calling `interrupt()` recursively invoking itself, whose cited test no longer exists.

## Also in this change

- **Prose**, per `contributing/general-writing-tips.md`: garden-path sentences, passive voice, sentences carrying too many concepts, emdash overuse, unnecessary parentheses. `general-writing-tips.md` itself was treated as near-read-only (two typo fixes; its deliberately-bad examples left bad), as was `anti-patterns.md`.
- **Dedup**: duplicated explanations replaced with a pointer to the doc that owns the topic, so two copies stop drifting apart independently.
- **Dead cross-references**, including links to specs that no longer exist, and several docs instructing readers to `pnpm add @agency-lang/tui` — **a package that has never existed**. The TUI is `lib/tui/`.
- `docs/misc/stateStack.md` and `docs/misc/lifecycleHooks.md` documented the same removed interrupt API; corrected to `respondToInterrupts`.
- **One rename**: `async-info-for-claude.md` → `async-behavior-checklist.md`. The content is a genuine behavioral checklist, but the scratch filename was load-bearing: `async.md` opened with a literal `CLAUDE: also read async-info-for-claude.md`.

## Seven source comments that disagreed with their own code

Comment-only, no behavior change:

- `lib/ir/tsIR.ts` cited a `lowerScopes` function that does not exist anywhere.
- `lib/runtime/ipc.ts` (×3) said the temp dir sits under `cwd`; the code uses the package root.
- `lib/cli/docLedger.ts` described `registrySymbols` as including non-exported names, but `extractRegistrySymbols` applies `isDocumented`.
- `lib/backends/typescriptBuilder.ts` documented a 3-argument `__validateChainRecursive`; 2 are emitted.
- `lib/backends/typescriptBuilder.ts` said "seven" compile paths run `expandSplices`; there are **six**, and CLAUDE.md said **five**. The count is now removed so it cannot rot again.
- `lib/cli/statelog/statelogRequest.ts` credited only serve with `envelope: false`; uploadClient passes it too.

## Verification

- A script resolves every `docs/dev/...` reference and every relative markdown link. Two remain unresolved: a path into the *statelog* repo, and `page.md` in `doc-cache.md`, which is a prose example of link syntax rather than a link. Both are correct as written.
- `pnpm run fmt:ts:check` passes.
- Every source-file change confirmed comment-only by diffing out comment lines — typecheck cannot be affected.
- Agents were required to cite `file:line` for each correction and to report anything they could not verify rather than guess. I spot-checked the strongest claims independently (e.g. that `parseAtom`/`parseExprPrec`/`minPrec` really have zero hits before accepting the `binop-parser.md` rewrite).

## Known limitations

- Claims that live outside this repo were **not** verified and are marked as such: statelog-side behavior, smoltalk internals, external benchmark numbers, arXiv citations, and recorded measurements.
- Two docs are design records for unbuilt work: `language/parallel-blocks-v2-dataflow.md` (v2 never shipped — confirmed: no `parallelDataflow.ts`, no `plan` command) and `language/lambda-sketch.md`. Both now state their status prominently. Whether they belong in `docs/dev/` is your call.
- Some dead code surfaced but was **not** removed, since this is a docs PR: `insertDebugSteps()` and `debugger.mustache` are unreachable, `config.debugger` is inert, and the CLI writes `.bundle` while `agency trace log` still advertises `.agencybundle`.

https://claude.ai/code/session_01UhmBRQ3r1HPhf9XsQomKdV


egonSchiele:
Reviewed by re-verifying the claims against the source independently of how they were written, rather than reading the prose for plausibility. Three mechanical sweeps plus targeted reads:

- **Every backticked file path on an added line** resolved against the tree. One wrong path (below); the rest are basenames used in prose after a full path was given earlier, which is fine.
- **Every backticked identifier (5+ chars) on an added line** checked for existence anywhere in `lib/`, `stdlib/`, `scripts/`, `agents/`, `tests/`. 1053 candidates, 18 misses, 17 of which are correct usages (things the doc explicitly says are gone or unbuilt: `_speak`, `containsMatch`, `NEVER_INLINE_ARM_TYPES`, `viewAttachment`, external env vars). The 18th is a real miss.
- **Every relative markdown link** in `docs/dev` + `docs/misc` resolved from its real location. Two unresolved, both the ones the description already calls out and both correct as written.
- Read and spot-verified the largest rewrites against source: `typechecker/README.md` (phase order in `check()`, the `BUILTIN_FUNCTION_TYPES` entry list, `checkType`/`synthType` signatures, guard desugar in the constructor), `binop-parser.md` (the whole operator table against `_exprParserBase`), `rewind.md` (the six numbered steps against `rewind.ts`), `subprocess-ipc.md` (`compile`/`run` signatures, the usage-message identifiers), `parallel-blocks.md` (`ALLOWED_ARM_TYPES`, `collectBindings`, `runForkAll` → `runBatch` → `allSettled`), `coding-standards.md` (all five limits and both commented-out rules against `eslint.config.js`), `checkpointing.md` (compiled a `static const` to confirm it lands outside `__ctx.globals`), and every source-comment change.

The corrections hold up. Five things to fix, then this is good to merge.

**Two are wrong facts.** Both are the exact failure mode the PR set out to remove, so they matter more than their size:

1. `config.md` documents `typechecker.matchExhaustiveness` as defaulting to `silent`. It defaults to `error` (`lib/typeChecker/matchExhaustiveness.ts:320`). Every other severity default in that table is right, which makes the wrong one harder to catch.
2. `checkpointing.md` cites `lib/runtime/state/threadRepair.ts`. The file is `lib/runtime/threadRepair.ts`. `lib/runtime/state/` does exist (that is where `globalStore.ts` lives), so the path looks plausible; and `threads.md` and `interrupts.md` both cite the correct path, so the same fact is now spelled two ways across the tree.

**Two are self-contradictions introduced by a partial edit.** A doc that disagrees with itself is worse than one that is uniformly stale, because a reader has no way to tell which half is current:

3. `interrupts.md` still names `TsIfSteps` twice, in the Match sections. `TsIfSteps` has zero hits in `lib/` — removing it is one of the headline fixes in this PR's own description, and the If/else section right above now correctly says `runnerIfElse`.
4. `subprocess-ipc.md` step 1 still says `_compile()` returns `{ moduleId, code }`. Two sections earlier the PR corrects that to `{ moduleId, code, modules?, entryPath? }` and explains why the extra fields matter.

**One is a knowingly incomplete correction:**

5. `config.md` lists `debugger` as "insert a pause point before every step". `config.debugger` has no reader anywhere in `lib/` — this PR's own description says it is inert. A rewrite whose purpose is to stop docs from describing machinery that does not run should not add a fresh entry describing machinery that does not run. Either mark it inert in the table or leave it out.

**One pre-existing thing this PR is well placed to fix:** `CLAUDE.md`'s pointer for `speech-via-smoltalk.md` still advertises "the `_speak` ABI freeze", but the doc says `_speak` was deleted rather than frozen, and `_speak` has zero hits repo-wide. That was already wrong on `main`, and this PR does touch `CLAUDE.md`.

Two notes on things I did **not** flag, so they don't get re-litigated later. `rewind.md` says `rewindFrom` "mints a fresh run id"; it is really `ctx.runId ?? nanoid()`, but the code comment right there explains the module-level ctx never carries one, so the doc describes the behavior correctly. And `binop-parser.md`'s level-1 row groups the right-associative operators ahead of the left-associative ones while the source interleaves them; the table is presenting associativity groups, not source order, so it reads fine.

On the two design records for unbuilt work (`parallel-blocks-v2-dataflow.md`, `lambda-sketch.md`): both now say plainly at the top that nothing has shipped, and I confirmed the specifics (no `parallelDataflow.ts`, no `plan` command). Keeping them is defensible now that they are labeled — but that is your call, and it is a separate decision from this PR.

packages/agency-lang/docs/dev/runtime/config.md
```diff
@@ -138,31 +138,53 @@ cannot affect a later process's validation.
...
-| `strictTypes` | `boolean` | If true, untyped variables are errors instead of implicit `any` |
-| `typeCheck` | `boolean` | Run the type checker during compilation (reports warnings) |
-| `typeCheckStrict` | `boolean` | Make type errors fatal (implies `typeCheck: true`) |
+| `typechecker.enabled` | `boolean` | Run the type checker during compilation and print warnings (default: false) |
+| `typechecker.strict` | `boolean` | Type errors are fatal, which implies `enabled` (default: false) |
+| `typechecker.strictTypes` | `boolean` | Untyped variables are errors instead of implicit `any` (default: false) |
+| `typechecker.undefinedFunctions` | `"silent" \| "warn" \| "error"` | An unresolvable function call (default: `warn`) |
+| `typechecker.undefinedVariables` | `"silent" \| "warn" \| "error"` | An unresolvable variable reference (default: `silent`) |
+| `typechecker.strictMemberAccess` | `"silent" \| "warn" \| "error"` | A property that exists on only some members of an un-narrowed union (default: `error`) |
+| `typechecker.matchExhaustiveness` | `"silent" \| "warn" \| "error"` | A `match` over a closed type that misses a case and has no `_` arm (default: `silent`) |
```
egonSchiele: Wrong default. `matchExhaustiveness` defaults to `error`, not `silent`:

```ts
// lib/typeChecker/matchExhaustiveness.ts:320
const configured = (ctx.config.typechecker?.matchExhaustiveness ?? "error") as Severity;
```

The file header says the same in prose: "Enabled by default at `error`."

Worth carrying over the nuance one line up in that function too, since it changes what the option means: the configured value only governs **statement** match sites. Expression matches are hard-checked per site regardless, which is why the pass deliberately does not early-return on `silent`.

I checked the other four severity defaults in this table (`undefinedFunctions` warn, `undefinedVariables` silent, `strictMemberAccess` error, `definiteReturns` warn) and all four are right.

packages/agency-lang/docs/dev/runtime/config.md
```diff
@@ -138,31 +138,53 @@ cannot affect a later process's validation.
...
 | Option | Type | Description |
 |--------|------|-------------|
 | `verbose` | `boolean` | Enable verbose compilation logging |
-| `outDir` | `string` | Output directory for compiled TypeScript files |
+| `logLevel` | `"debug" \| "info" \| "warn" \| "error"` | Compiler log level |
+| `outDir` | `string` | Output directory for compiled files |
+| `distDir` | `string` | Directory of pre-compiled JS. The debugger imports from here instead of compiling on the fly. |
+| `allowNonAgencyGenerators` | `boolean` | Let a compile-time splice generator import JavaScript. Off by default, because a generator that reaches an npm package is unchecked. See [`docs/dev/language/splices.md`](../language/splices.md). |
+| `instrument` | `boolean` | Emit `debugStep()` instrumentation in compiled output (default: true) |
+| `debugger` | `boolean` | Debugger mode: insert a pause point before every step |
```
egonSchiele: `config.debugger` is inert. It is declared in `lib/config.ts:259` and in the schema at `:566`, but nothing reads it — no hits for `config.debugger` / `cfg.debugger` anywhere in `lib/`. The PR description says as much ("`config.debugger` is inert", alongside the unreachable `insertDebugSteps()` and `debugger.mustache`).

So this row does the thing the PR exists to stop: it describes machinery that does not run. Either mark it inert here, or drop the row and leave it to the dead-code follow-up. Same question applies to `instrument`, which is in the table right above — I did not chase whether `debugStep()` emission is still wired, so that one may well be fine.

packages/agency-lang/docs/dev/runtime/checkpointing.md
```diff
@@ -101,17 +122,19 @@ This allows injecting messages into the restored state, useful for providing con
...
-The fact that shared variables persist across restores is what makes retry loops work — a shared counter can track how many times a checkpoint has been restored even though local variables reset.
+Something has to survive the rollback for a retry loop to terminate. That is why the loop counter usually lives outside the GlobalStore, either as a `static` value or in a plain JS module the program imports.
 
-A checkpoint whose thread has since been repaired (see "Reopen repair" in
-threads.md) is stale: `restoreThreadForResume` throws rather than letting
-the old snapshot overwrite the repaired thread. `MessageThread.repairs` is
-the generation both sides compare; `markRepaired()` is its only writer.
+A checkpoint whose thread has since been repaired is stale. See "Reopen repair"
+in [`threads.md`](./threads.md). `restoreThreadForResume` (in
+`lib/runtime/state/threadRepair.ts`) throws rather than letting the old snapshot
```
egonSchiele: Path is wrong: `restoreThreadForResume` lives in `lib/runtime/threadRepair.ts` (defined at `:186`), not `lib/runtime/state/threadRepair.ts`.

Easy one to miss, because `lib/runtime/state/` is a real directory — `globalStore.ts`, `context.ts`, `checkpointStore.ts` are all in it. Two other docs on this branch cite the correct path (`threads.md`, and the key-files table in `interrupts.md`), so right now the same file is spelled two ways.

Everything else in this paragraph checks out: `MessageThread.repairs` is the generation counter, `markRepaired()` is its only writer, and `restoreThreadForResume` does throw rather than overwrite.

packages/agency-lang/docs/dev/runtime/interrupts.md
```diff
@@ -2,86 +2,59 @@
...
+Match blocks reuse `TsIfSteps`, exactly like if/else: `processMatchBlockWithSteps` (`lib/backends/typescriptBuilder.ts`) turns each match case into a branch with an `===` equality condition against the scrutinee, and the `_` case (if present) becomes the else branch. Arm bodies are processed through `processBodyAsParts`, the same helper if/else uses, so each statement in an arm gets its own `__substep_<key>` counter — a multi-statement block arm resumes mid-arm exactly like a multi-statement if/else branch resumes mid-branch. The `__condbranch_<key>` cache means the winning arm is decided once and re-dispatched to (not re-matched) on resume, even if the scrutinee or a guard has side effects.
```
egonSchiele: `TsIfSteps` does not exist. Zero hits in `lib/` — the IR node is `runnerIfElse` (`lib/ir/tsIR.ts:367`).

Removing this name is one of the headline fixes in the PR description, and the If/else section directly above now correctly describes `processIfElseWithSteps` building a `runnerIfElse` node. These two Match paragraphs (here and line 61) kept the old name, so the doc now contradicts itself about what the shared IR node is called — which is worse for a reader than being uniformly stale, since there is no way to tell which half is current.

The surrounding claims are all correct, so this is a rename in place: `processMatchBlockWithSteps`, `processBodyAsParts`, the `__condbranch_<key>` caching, and `runnerIfElse.mustache` being the only template involved.

packages/agency-lang/docs/dev/runtime/subprocess-ipc.md
```diff
@@ -118,39 +127,46 @@ On the parent side, `_run` — a `runBatch` adopter with a single child (`subpro
 ## How compiled code gets executed in the subprocess
 
 1. `_compile()` runs the Agency compilation pipeline and returns `{ moduleId, code }` — no disk writes.
```
egonSchiele: This still says `{ moduleId, code }`, but the Durability section above corrects the shape to `{ moduleId, code, modules?, entryPath? }` and explains that `modules`/`entryPath` carry the rest of a multi-file closure. The key-files table entry for `lib/stdlib/agency.ts` was updated too. This line is the leftover.

It matters here specifically: the point of the numbered list is that resume re-materializes from the compiled value, and the two-field shape is exactly the thing that would not survive a multi-file program.

egonSchiele (commented):
Five inline notes; details and how I checked are in the summary comment.

packages/agency-lang/docs/dev/runtime/config.md
egonSchiele: Fixed in fa546f85c. Default corrected to `error`.

I also carried over the nuance you pointed at, in the words the code comment uses: only statement match sites honor the configured value, and an expression match is hard-checked per site regardless. I first wrote a reason for that ("because its type depends on the arms") and then cut it, since it was my inference rather than anything the source states.

Your method here is what caught the other four path errors: you found this by checking all five severity defaults rather than the one that looked wrong, so I ran the same shape of sweep over every backticked file path in docs/dev. Details in the reply on the subprocess-ipc thread.

packages/agency-lang/docs/dev/runtime/config.md
egonSchiele: Fair hit, and you are right that it is the exact thing this PR exists to stop. Fixed in fa546f85c: the row now reads "Inert. \`agency debug\` sets it, but nothing reads it. \`instrument\` is the flag that controls \`debugStep()\` emission."

I kept the row rather than dropping it, because a reader who finds `debugger` in `agency.json` or in the schema needs to be told it does nothing. Silence would send them looking for the reader.

On `instrument`: it is live. `lib/backends/typescriptBuilder.ts:3624,3929` both gate emission on it, and `lib/config.ts:261-263` defaults it to true. That row is accurate.

packages/agency-lang/docs/dev/runtime/checkpointing.md
egonSchiele: Fixed in fa546f85c. Corrected to `lib/runtime/threadRepair.ts` (`restoreThreadForResume` at :186).

Confirmed the failure mode you describe: `lib/runtime/state/` is real, so the wrong path is plausible rather than obviously wrong, which is why the agent that wrote it and I both read past it.

packages/agency-lang/docs/dev/runtime/interrupts.md
egonSchiele: Fixed in fa546f85c. Both Match paragraphs now say `runnerIfElse`, so the doc no longer contradicts the If/else section above it. Zero `TsIfSteps` left in the file, and it still has zero hits in `lib/`.

Agreed on the general point: uniformly stale is at least legible, whereas a doc arguing with itself gives the reader no way to pick a side. That is a partial-edit failure and worth watching for wherever a rename lands in a long doc.

packages/agency-lang/docs/dev/runtime/subprocess-ipc.md
egonSchiele: Fixed in fa546f85c. Step 1 now reads `{ moduleId, code, modules?, entryPath? }` and says what the two extra fields carry, since as you note the whole point of that list is that resume re-materializes from the compiled value.

This one prompted a wider sweep. Since you found the `matchExhaustiveness` default by checking every severity default rather than the suspicious one, I ran the same over every backticked file path in docs/dev: 18 did not resolve, and 4 more were real errors, all now fixed in the same commit.

- `stdlib/image.ts` and `stdlib/speech.ts` are under `lib/stdlib/`
- `stdlib/template.ts` is `lib/stdlib/template.ts`
- splices.md pointed at `docs/site/guide/templates.md`; the page is `docs/site/guide/splices.md`
- agent-brains.md listed six files relative to `lib/agents/agency-agent/` while writing some bare and some `lib/`-prefixed. It now states the base directory once instead of prefixing six bullets.

The other 14 are correct usages: paths the docs explicitly say are gone (`lib/lsp/locations.ts`, `scripts/stdlib-stamp.ts`, `lib/stdlib/providers/llama-cpp.mjs`), unbuilt work (`lib/preprocessors/parallelDataflow.ts`), and design-history pointers into `docs/superpowers`.

egonSchiele:
All five fixed in fa546f85c, plus four more of the same kind that your method turned up.

**The five:**

| | Fix |
|---|---|
| `matchExhaustiveness` default | `silent` → `"error"`, plus the statement-vs-expression nuance from the code comment |
| `threadRepair.ts` path | dropped the wrong `state/` segment |
| `TsIfSteps` ×2 | → `runnerIfElse`; zero left in the file |
| `_compile()` shape | → `{ moduleId, code, modules?, entryPath? }` with what the extra fields carry |
| `debugger` row | marked inert, pointing at `instrument` as the live flag |

Plus the `_speak` pointer in CLAUDE.md, which you were right that this PR is well placed to fix.

**What your method turned up.** You found the wrong default by checking all five severity defaults rather than the one that looked off. I ran the same shape of sweep over every backticked file path in `docs/dev`: 18 did not resolve, and **4 were real errors**, all fixed in the same commit — `stdlib/image.ts` and `stdlib/speech.ts` are under `lib/stdlib/`, `stdlib/template.ts` likewise, and `splices.md` pointed user-facing docs at `docs/site/guide/templates.md` when the page is `docs/site/guide/splices.md`. `agent-brains.md` listed six files relative to `lib/agents/agency-agent/` while writing some bare and some `lib/`-prefixed; it now states the base directory once rather than prefixing six bullets.

The other 14 unresolved paths are correct usages: things the docs explicitly say are gone (`lib/lsp/locations.ts`, `scripts/stdlib-stamp.ts`, `lib/stdlib/providers/llama-cpp.mjs`), unbuilt work (`lib/preprocessors/parallelDataflow.ts`), and design-history pointers into `docs/superpowers`.

**One thing I cut rather than kept.** On the `matchExhaustiveness` row I first wrote *why* expression matches are hard-checked ("because its type depends on the arms") and then removed it. The source states the split but not the reason, and inventing a plausible rationale is how this class of doc goes wrong in the first place.

**On the two design records:** noted that keeping them is my call and separate from this PR. I'd like to keep them for now — both say plainly at the top that nothing shipped, so they read as design history rather than as descriptions of the system. If they still bother you later, the cheap move is to relocate them out of `docs/dev/language/` rather than delete the reasoning.

**Still not done here, deliberately:** the dead code this surfaced (`insertDebugSteps()` and `debugger.mustache` unreachable, `config.debugger` inert, CLI writing `.bundle` while `agency trace log` advertises `.agencybundle`). Those are code changes and want their own PR.

Thanks for re-verifying against source instead of reading for plausibility — that is the only review that would have caught the `matchExhaustiveness` default, since the prose around it was fine.

PR #919, opened 2026-08-25
Fix the real-LLM post-merge suite, red on main since the model change

egonSchiele:
## What is broken

**Real-LLM tests (post-merge)** has failed on every push to `main` since 070d8153d, "Default model gpt-5-mini through openai-responses, as one named constant" (#893). The workflow was green through 2026-08-23T20:00 and red from the very next run.

Ten failures. None is flaky: each reproduces every run and survives its own retries. There are four distinct causes.

## 1. Seven tests compared against the model's own answers

`threads/subthreads`, `threads/messages`, `threads/thread-inside-function`, `threads/thread-three-levels-deep`, `threads/nested-function-calls/nested-three-levels-deep`, `stringInterpolation`, `fork/llm-tools/multi-tool-all-interrupt`.

These check thread plumbing and interrupt batching, but they assert on full message content, which means they assert on what the model said. Their values were recorded on gpt-4o-mini through the base `openai` route.

The route turned out to matter as much as the model. `openai-responses` enforces the declared return type strictly, so this line:

```ts
const res3: number = llm("And what are the next two integers after those?")
```

returns a single number there, where the old route returned `[18, 19]` — which is what the expectation records. I confirmed this by pinning only the model first: still red, with answers varying run to run. Pinning the route as well made it pass first try.

Each of the seven now pins both, with a comment saying why:

```ts
setLlmOptions({ model: "gpt-4o-mini", provider: "openai" })
```

## 2. A judge overruled by its own expected output

`nested-three-levels-deep` needed one thing more. Four of its threads start with "What are the next 2 prime numbers after that?" and contain no earlier message saying what "that" is, so any primes are valid — and its judge prompt already said so ("Any numbers except 103 and 107 will work").

But `lib/agents/eval/judge.agency` also pastes `expectedOutput` into the judge verbatim and asks it to "score how well the actual output matches what was expected". The judge graded against the recorded numbers and failed runs its own prompt permits. The prompt now states outright that those numbers are one recorded example.

Worth knowing generally: for any `llmJudge` test, `expectedOutput` competes with `judgePrompt`, and tends to win.

## 3. A test asserting a mechanism removed in #522

`tool-non-retryable-failure` passed `failure("permanent failure", { retryable: false })` and expected the runtime to say the call cannot be retried.

No option has ever read `retryable`. `FailureOpts` (`lib/runtime/result.ts:105`) takes `checkpoint`, `neverStarted`, `destructiveRan`, `functionName`, `args` — and the `failure()` builtin types its second parameter as `ANY_T`, so the unknown key type-checked and was silently dropped. The failure fell into the `neutral` tier and the model was told **"You may call this tool again"**, the opposite of the test's premise. It only ever passed because the old model answered `tool-removed` anyway.

Rewritten against the tier that does mean this — `destructiveRan: true`, which tells the model the tool can no longer be called and drops it from the list. It now passes on the **current** default model, unpinned.

## 4. A token budget too small for a reasoning model, and an error that hid it

`llmConfig` set `maxTokens: 50`. gpt-5-mini spends that on reasoning, so the visible output is empty and the call dies as:

```
LLM structured output failed validation: expected object, received string.
Raw output (first 200 chars): ""
```

That message sends you looking at your types. The test now uses a workable budget, and the runtime explains the real case instead: smoltalk already normalizes every provider's out-of-room reason (`length` / `max_tokens` / `MAX_TOKENS` / `max_output_tokens`) to `stopReason: "length"`, so `runPrompt` carries the last round's stop reason to `decideValidationRetry`, which names the token limit and points at `maxTokens`.

Fixing that surfaced a second bug: the statelog's `finishReason` field read `completion.finishReason ?? completion.finish_reason`, and `PromptResult` has neither. It has always logged `undefined`. It now reads `stopReason`.

## 5. An optimizer fixture that asked for the one thing the mutator forbids

`test:optimize-efficacy` had greedy and gepa stuck at objective 0.00 forever. The fixture made the whole prompt optimizable and the goal was "Return the capital of India", so the only winning move was to hard-code the answer. But `lib/agents/optimize/mutatePrompt.agency` says:

> Write a general instruction that works for any input of this kind. Do NOT hard-code the specific expected answers shown in the feedback into the value.

So the mutator dutifully wrote instructions ("Answer the user's question directly and accurately...") and the agent, which passes that string to `llm()` as the entire user message, had no question left to answer. The `TODO(gepa-efficacy)` comment had already diagnosed the symptom without spotting that the fixture and the mutator prompt were in direct conflict.

The optimizable part is now a style line, with the question fixed:

```ts
optimize const style = "Be warm and chatty, and set the scene before answering."
const response = llm("${style} What is the capital of France?")
```

Now a general instruction genuinely wins — the mutator's first proposal ("Be extremely literal and concise... output exactly that content and nothing else") takes it from 0.00 to 1.00. This fixes gepa for the same reason, so the gepa case is **re-enabled** and its TODO removed. `containsDelhi.ts` becomes `onlyCityName.ts`, and the harness's own "champion mentions Delhi" assertion becomes a bare-city check.

## Verification

Against the real API, not just deterministic mode:

- All 7 pinned tests pass.
- `tool-non-retryable-failure` and `llmConfig` pass on the current default model, **unpinned** — they should not need a pin, and they don't.
- `pnpm run test:optimize-efficacy` passes all four cases end to end, gepa included: `greedy-judge`, `gepa-judge`, `greedy-grader`, `custom-optimizer`, each baseline 0 to champion 1.
- All 9 changed tests also pass under `AGENCY_USE_TEST_LLM_PROVIDER=1`, which is what PR CI runs.
- `pnpm run typecheck` (all three configs), `pnpm run lint:structure`, `pnpm run fmt:ts`, and `vitest run lib/runtime` (132 files, 2048 tests) all green.

## Docs

`docs/dev/llm/smoltalk.md` gains the `StopReason` row and a section on why a turn ended, including the `stopReason` vs `finishReason` trap. `docs/misc/TESTING.md` gains a section on what to do when a mocked test also has to survive a real model, since the post-merge run bypasses `llmMocks` entirely.

## Not fixed here

The judge harness handing `expectedOutput` to the judge alongside `judgePrompt` is a general sharp edge, not just this test's problem. I worked around it in the one test rather than changing `judge.agency`, which every `llmJudge` test depends on. Worth a separate look.

https://claude.ai/code/session_013AqSsrq53XR3LVnJR9eaJr


packages/agency-lang/docs/misc/TESTING.md
```diff
@@ -281,6 +281,37 @@ Keys are matched against the module making the `llm()` call: the exact module id
 
 The post-merge workflow ([`.github/workflows/test-with-llm.yml`](../../.github/workflows/test-with-llm.yml)) re-runs the same suites against the real OpenAI provider after a PR lands on `main`.
 
+#### When a mocked test also has to survive a real model
```
egonSchiele: This was a one-time fix that I don't expect to have to do again. The fix we applied was a hack and should not be applied to future fixes. Adding this info to the docs is steering future models towards the wrong thing.

packages/agency-lang/tests/agency/fork/llm-tools/tool-non-retryable-failure.agency
```diff
@@ -1,18 +1,22 @@
-// Tool returns failure(..., { retryable: false }). runPrompt's isFailure
-// path should push "Error: ${msg}. This operation failed and cannot be
-// retried." and remove the tool. The LLM's next round runs without the
-// tool available and produces structured output directly.
+// A tool whose failure reports that destructive work already ran. That is the
+// most serious tier in runPrompt's isFailure path (lib/runtime/prompt.ts): it
+// tells the model the tool can no longer be called and drops the tool from the
+// list, so the next round runs without it and answers directly.
+//
+// This used to pass `{ retryable: false }`, which no option has ever read --
```
egonSchiele: Why is it useful to add this historical context here?

packages/agency-lang/tests/agency/threads/nested-function-calls/nested-three-levels-deep.agency
```diff
@@ -1,8 +1,17 @@
+import { setLlmOptions } from "std::llm"
+
 def nextTwoPrimes(): number[] {
   return llm("What are the next 2 prime numbers after that?")
 }
 
 node main() {
+  // This test checks the thread plumbing by comparing the model's own
```
egonSchiele: Please review the comments you have added in this PR. Are they genuinely useful context for feature readers, or are they notes you made for yourself while you were making the changes?

packages/agency-lang/docs/misc/TESTING.md
egonSchiele: Agreed, and removed. Writing it up as guidance would have taught the next person to reach for the pin first, when the right instinct is a test that does not depend on the model at all. The pins stay because they are what makes the suite green, but nothing recommends them now.

packages/agency-lang/tests/agency/fork/llm-tools/tool-non-retryable-failure.agency
egonSchiele: It is not. Cut. The comment now just describes the tier the test exercises:

// A failure that reports destructive work already ran. runPrompt tells the
// model this tool can no longer be called and drops it, so the next round
// answers without it.

packages/agency-lang/tests/agency/threads/nested-function-calls/nested-three-levels-deep.agency
egonSchiele: Notes to myself, mostly. I went back through all of them.

The six-line block above each setLlmOptions call is now one line saying what the pin protects. tool-non-retryable-failure no longer recounts what it used to assert. The optimizer fixture drops the account of its previous shape. The runtime comments lost the sentences narrating the change rather than explaining the code. Net 116 lines removed against 40 added.

What I kept is the non-obvious stuff a reader would otherwise have to rediscover: why lastStopReason is deliberately not on `self`, and the smoltalk field-name trap, now stated as a standing fact rather than as the story of a bug.

packages/agency-lang/tests/integration/optimize-efficacy/test.mjs
```diff
@@ -67,9 +59,13 @@ function runOnce({ name, flags }) {
...
   }
+  // The objective alone could rise for the wrong reason, so also check the
+  // champion actually answers the way the goal asks: the bare city, with none
+  // of the baseline style's chatter around it.
   const outputs = (championBreakdown ?? []).map((b) => String(b.output));
-  if (!outputs.some((o) => /delhi/i.test(o))) {
-    throw new Error(`champion output never mentions Delhi: ${JSON.stringify(outputs)}`);
+  const isBareCity = (o) => /^paris[.!]?$/i.test(o.trim());
+  if (!outputs.some(isBareCity)) {
+    throw new Error(`champion output is not the bare city name: ${JSON.stringify(outputs)}`);
```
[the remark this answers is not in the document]
egonSchiele: Good catch on both halves. Fixed:

    if (outputs.length === 0 || !outputs.every(isBareCity)) {

The fixture runs one input today, so `some` and `every` agree right now, but the check should not silently weaken if inputs are added later. The empty-breakdown case was the worse bug: `every` on `[]` is true, so a run that graded nothing at all would have been reported as a pass.

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -70,6 +70,9 @@ type Tool = {
 export type RunPromptResult = {
   messages: MessageThread;
   toolCalls: smoltalk.ToolCallJSON[];
+  /** Why this round ended, normalized by smoltalk across providers. Read
+   *  only to explain a validation failure the token limit caused. */
+  stopReason?: smoltalk.StopReason;
```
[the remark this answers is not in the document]
egonSchiele: Correct, the comment contradicted the type. Reworded to "the result is always plain data", and the new field carries its own one-line doc.

packages/agency-lang/tests/agency/threads/nested-function-calls/nested-three-levels-deep.test.json
```diff
@@ -8,7 +8,7 @@
...
           "desiredAccuracy": 70
```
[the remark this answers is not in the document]
egonSchiele: Most of this was already fixed by merging main, which brought in a cleaned-up version of the prompt: the "respom" typo is gone, msgs5 is back to four bullets, and there is now a proper msgs6 section, so the sections match the keys in expectedOutput exactly.

The count was still wrong, though. It said eight keys and there are seven (msgs plus msgs2 through msgs7). Fixed.

PR #920, opened 2026-08-25
Writing-review suite to 12 tests, gepa fixes, and the optimized reviewer prompt

egonSchiele:
Two things in one branch: the writing-review eval suite grew from 4 tests to 12, and the optimizer got the fixes needed to run gepa on it end to end. The prompt on `review.agency` is the champion from that run.

## The eval suite (`evals/writing-review/`)

- 8 new tests: 3 harvested by hand (`config-doc`, `diagnostic`, `eval-description`) and 5 from `examples/misc.txt` (writing tics like "costs no parse", "also answers to"). Ten Beemo-dataset tests were tried and then removed: the prompts are CC-BY-NC.
- `names-the-flaws` now asks the judge about each editor's point separately (one top-level bullet in `graderFiles/notes.md` = one point). Asking it to enumerate the points itself let it pad the list with the reviewer's own findings.
- New `recommends-cuts` grader: compares the original with the editor's version and asks whether the findings call for the same cuts.
- `rewrites-faithful` gets the assignment as context, so restating the task is not counted as an invented fact.
- The test template lives outside the suite directory (its empty `test.json` broke loading).

## The optimizer

- gepa runs its proposer on `--mutator-model` (before, only greedy honored it).
- `file.agency:node` targets reach candidate runs; before, every candidate ran `main`.
- Per-iteration train and validation objectives are recorded and shown in the report table. (They were being dropped on the way into the result.)
- A cost line: agent, grading, and mutator spend, measured from statelogs and the runner.
- Proposers are asked for the prompt as plain text, which is taken verbatim (wrapped as `"""..."""`, or `"..."` with escapes when it cannot be). A multi-line value that arrives in `"..."` with `\n` escapes is written back as a `"""` block when that round-trips. Three of eight iterations in the trial run were lost to quoting before this.

## The trial run

`writing-review-3` (gepa, gpt-5 mutator, 8 iterations, 15 train / 7 validation tests, ten of them the since-removed Beemo tests, $3.19): champion validation 0.820 vs baseline 0.772. Re-grading both with fresh judge passes gave 0.833/0.834 vs 0.752/0.770, so the gap holds. The champion prompt names the reader and their decisions first, does a cut/move pass, then applies the sentence-level rules.

Known gap: the champion raises error findings on `clean-paragraph` (`rejects-nothing` fails), and no prompt so far gets the reviewer to say "this text should not exist" instead of rewording it. Both are next.


packages/agency-lang/tests/integration/optimize-efficacy/test.mjs
```diff
@@ -22,9 +22,9 @@ const HERE = dirname(fileURLToPath(import.meta.url));
 // of how the harness was launched.
 const PACKAGE_DIR = dirname(dirname(dirname(HERE)));
 const AGENT = join(HERE, "fixtures", "agent.agency");
-const GRADER = join(HERE, "fixtures", "containsDelhi.ts");
+const GRADER = join(HERE, "fixtures", "onlyCityName.ts");
```
egonSchiele, line 25: The grader fixture this points at does not exist. `containsDelhi.ts` is deleted in this PR and `fixtures/onlyCityName.ts` was never added — `ls tests/integration/optimize-efficacy/fixtures/` on this branch shows only `agent.agency` and `customOptimizer.ts`. The `greedy-grader` scenario passes `--graders <missing file>`, so the optimize-efficacy integration test fails as soon as it runs (main-only, real LLM, so CI on this PR will not catch it). Looks like a missed `git add`.

packages/agency-lang/lib/optimize/sourceMutator.ts
```diff
@@ -448,6 +455,15 @@ export class OptimizeSourceMutator {
...
         });
       }
+      // A multi-line prompt sent as "..." with \n escapes would be written
+      // back as one long line. Keep the source readable: emit it as """...""".
+      if (
+        target.valueKind === "multilineString" &&
+        replacement.type === "string" &&
+        newValue.includes("\n")
+      ) {
+        replacement = { type: "multiLineString", segments: replacement.segments };
```
egonSchiele: This re-type from `string` to `multiLineString` can turn a perfectly valid replacement into unparseable source. `escapeMultiLineText` (lib/backends/agencyGenerator.ts:150) only escapes `${` — it does not escape `"` or `"""` — while `escapeStringText` on the `"..."` path does. So a value whose decoded text ends with a quote, or contains `"""`, renders as broken Agency and is rejected by the round-trip guard.

Measured on this branch with a `"""` target and `value = "\"first line\\nsay \\\"hi\\\"\""`:

```
[{"code":"parse-failed","message":"Rendered candidate for multi.agency does not parse: Line 7, col 1: expected \"\"\" ..."}]
```

Before this hunk that same proposal was written back as a single-line `"..."` literal with the quotes escaped and worked. A prompt that ends with a quoted example (common) now costs a whole proposal attempt, and the diagnostic the mutator gets back ("Rendered candidate ... does not parse") tells it nothing it can act on. Guard the conversion on `!newValue.includes(String.fromCharCode(34).repeat(3)) && !newValue.endsWith(String.fromCharCode(34))`, or teach `escapeMultiLineText` about the closing fence.

packages/agency-lang/lib/optimize/sourceMutator.ts
```diff
@@ -418,6 +418,13 @@ export class OptimizeSourceMutator {
       const quote = target.valueKind === "multilineString" ? `"""` : `"`;
       const wrapped = exprParser(`${quote}${operation.value}${quote}`);
       if (wrapped.success && wrapped.rest.trim() === "") parsed = wrapped;
+      // A multi-line prompt sent in plain quotes cannot parse (newlines), and
+      // wrapping it whole gives four quotes. Strip the pair and wrap the inside.
+      const plainQuoted = /^"(?!"")([\s\S]*)"$/.exec(operation.value);
```
egonSchiele: The regex is greedy and unanchored to any structure, so it strips a leading and trailing quote from text that was never quote-wrapped. The proposer is now told to send plain text with real line breaks (mutatePrompt.agency / gepaReflect.agency in this PR), so a multi-line prompt that legitimately begins and ends with a `"` character — e.g. it opens and closes with a quoted example — silently loses both quotes. Verified on this branch: `value = "\"quoted start\\nmiddle\\nquoted end\""` (real newlines) previews with no diagnostics and `newValue === "quoted start\nmiddle\nquoted end"`. The candidate is then scored against a prompt the model did not propose.

packages/agency-lang/evals/writing-review/lib/reviewGraders.ts
```diff
@@ -168,6 +171,21 @@ function harvest(graderFiles: string): { notes: string; cleaned: string | null }
   return { notes, cleaned };
 }
 
+/** The editor's points, one per top-level bullet in notes.md. Notes written
+ *  as prose are one point. The judge is asked about each point on its own,
+ *  so it cannot pad the list with the reviewer's findings. */
+function editorPoints(notes: string): string[] {
```
egonSchiele: Splitting the notes on `^[-*] ` silently drops every point written as prose whenever the file also contains at least one bullet, and it drops a bullet's continuation lines.

`vitepress-emoji-comment/graderFiles/notes.md` is the concrete case. It has 3 bullets, then the line "Also removed some lines that were awkwardly phrased:" followed by 2 bare quoted fragments as bullets, then four substantive prose points (the run-on sentence, "The `true` is markdown", "if a later VitePress", "the overall issue is that it is also way too long"). After this change `names-the-flaws` grades 5 points: the 3 real ones plus two fragments that carry no claim at all ("so the plugin only ever costs us"), and the four prose points are not graded. The denominator is wrong in both directions, so the score for that test no longer measures what it says it does.

Either require bullets-only notes and assert it (throw when a non-blank, non-bullet line appears after the first bullet), or split on blank-line-separated blocks instead of lines.

packages/agency-lang/lib/optimize/report.ts
```diff
@@ -15,6 +15,15 @@ export type ReportMeta = {
 };
 
 /** Escape a Markdown table cell: no pipes (column breaks) or newlines (row breaks). */
+export function formatCost(cost: OptimizeCost): string {
```
egonSchiele, line 18: The new function was inserted directly under the `cell` docstring, so `/** Escape a Markdown table cell ... */` now documents `formatCost`, and `cell` is undocumented. Move the comment back down to `cell`.

packages/agency-lang/lib/optimize/baseOptimizer.ts
```diff
@@ -181,13 +189,16 @@ export abstract class BaseOptimizer {
     let rationale = "";
     for (let attempt = 1; attempt <= maxAttempts; attempt += 1) {
       let proposal: MutationProposal;
+      const runnerBefore = this.agencyRunner.costUsd;
       try {
         proposal = await propose(diagnostics);
       } catch (error) {
+        this.countProposalCost(runnerBefore, 0);
```
egonSchiele: On the throw path the proposal cost is reported as 0. For greedy, `proposeMutation` calls the model through `executeNodeAsync` and only then throws on a malformed response (lib/optimize/mutator.ts:128) — the model call was paid for, and its `costUsd` is unreachable once the exception escapes. `viaRunner` covers gepa, but greedy under-reports `mutatorUsd` by one full proposal per malformed response, which is exactly the case retries exist for. Consider attaching the cost to the thrown error so it can still be counted.

packages/agency-lang/evals/writing-review/beemo-1268/files/text.md
```diff
@@ -0,0 +1,7 @@
+The Dream Team was the nickname given to the United States men's basketball team that competed in the 1992 Olympic Games in Barcelona, Spain. The team was composed of some of the best professional basketball players in the world, including Michael Jordan, Magic Johnson, Larry Bird, and Charles Barkley, among others. The Dream Team is widely considered one of the greatest basketball teams of all time, and its success on the court helped popularize the NBA and basketball around the world.
```
egonSchiele: Could you name these folders a little more descriptively? I think it still makes sense to prefix them with "beemo" and the ID, but something more would be helpful in quickly identifying which example I'm looking at, something like "beemo-1268-basketball-dream-team"

egonSchiele:
Review round addressed in four commits:

- **Beemo tests removed** (a5d6cde00). The prompts are CC-BY-NC, so they are out of the suite; the files stay local as material for our own examples. This also moots the folder-naming thread.
- **sourceMutator quote stripping** (95afdd3ce): the strip is gone. Plain text is wrapped as `"""..."""`, or as `"..."` with escapes when it cannot sit inside triple quotes (text ending in a quote). `"quoted start\nmiddle\n"quoted end"` now round-trips verbatim; test added.
- **`"""` retype breaking on quotes** (95afdd3ce): only retyped when the rendered block parses back to the same text; otherwise it stays `"..."`. Your `say \"hi\"` case is now a test.
- **Greedy malformed-proposal cost** (95afdd3ce): `MalformedProposalError` carries the spend; the catch counts it.
- **`cell` docstring** (95afdd3ce): moved back.
- **editorPoints dropping prose** (9567d59bd): bullets and prose paragraphs are both points, indented lines continue a point, and a paragraph ending in a colon owns the bullets under it. `vitepress-emoji-comment` now yields its 8 real points instead of 5.
- **optimize-efficacy fixture** (cc7d132cd): `onlyCityName.ts` added (bare city name, no sentence) and the assertion checks for Paris. This was in the `save` commit, so please read the grader to make sure it matches what you meant.
- **Judge prompt** (cc7d132cd): eight keys, msgs5 has four messages, msgs6 section added.
- **Trailing comma** (cc7d132cd): removed. It did parse, so this was cosmetic.

PR #921, opened 2026-08-25
eval: one SIGINT listener for all live child processes

egonSchiele:
Follow-up from #920. Every eval child process registered its own `process.once("SIGINT", ...)` so a programmatic SIGINT would reach it. The optimizer runs all of a candidate's inputs at once (15 children on a full-set scoring, 21 during validation), so every optimize run printed:

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGINT listeners added to [process].
```

It was not a leak (each listener was removed when its child settled), only Node's heuristic. Now `lib/eval/run/sigintForwarder.ts` keeps a list of live children behind one listener, installed when the first child starts and removed when the last one settles. Behavior on Ctrl-C is unchanged: every live child still gets the signal, and a process with no children keeps default handling.

Tests: a unit test with 15 fake children checks one listener, all forwarded, and removal after the last stop; the existing subprocess tests pass.


packages/agency-lang/lib/eval/run/sigintForwarder.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const live: Killable[] = [];
+
+function forwardToAll(): void {
+  for (const child of live) child.kill("SIGINT");
+}
+
+/** Forward SIGINT to `child` until the returned function is called. */
+export function forwardSigintTo(child: Killable): () => void {
+  if (live.length === 0) process.on("SIGINT", forwardToAll);
```
egonSchiele: **This breaks "press Ctrl-C again to force quit."**

The old code used `process.once` per child. That mattered for more than tidiness: a JS SIGINT listener overrides Node's default disposition, so `once` meant that after the first Ctrl-C every listener had removed itself and the process was back to default handling — a second Ctrl-C killed it immediately. `runSuite.ts:128-131` depends on exactly that, and says so:

```
// ... `once`, so a second
// Ctrl-C gets default handling — immediate death.
```

and it prints `press Ctrl-C again to force quit` to the user.

This listener is `on`, and it stays installed for as long as any child is live. So the second Ctrl-C now runs `forwardToAll` again and the process survives. The window where that promise is broken is precisely the window where the message is printed: an in-flight test whose child has not yet exited.

I measured it with a stand-in for `runSuite`'s warning listener plus a child that does not die promptly on SIGINT:

```
=== old (once per child) ===
[suite] press Ctrl-C again to force quit
exit code: 130

=== new (one persistent `on`) ===
[suite] press Ctrl-C again to force quit
[suite] STILL ALIVE after 2 Ctrl-Cs
exit code: 7
```

Failure scenario: an eval child wedged in a long LLM call, or any child slow to unwind. Ctrl-C prints the force-quit message, Ctrl-C again does nothing, and the user has no way out of the run short of `Ctrl-\` or `kill -9` from another terminal. Before this PR that second Ctrl-C worked.

The fix can stay small — have `forwardToAll` remove itself as it fires, so the first signal both forwards and restores default disposition, matching the old semantics:

```ts
function forwardToAll(): void {
  process.removeListener("SIGINT", forwardToAll);
  for (const child of live) child.kill("SIGINT");
}
```

with `forwardSigintTo` re-arming only when the listener is absent rather than keying off `live.length === 0`. Whatever shape you pick, it is worth a test: assert `process.listenerCount("SIGINT")` is back to the baseline right after the first `emit`, while children are still registered. The current tests both pass either way, so nothing here would catch a regression.

packages/agency-lang/lib/eval/run/sigintForwarder.ts
```diff
@@ -0,0 +1,28 @@
...
+ *  A terminal Ctrl-C reaches the whole process group, but a programmatic
+ *  SIGINT (a supervisor, process.kill) hits only this process, so it is
+ *  forwarded to each child that is still running. The optimizer runs a whole
+ *  candidate's inputs at once; with a listener per child, Node warned about a
+ *  leak past ten. The listener exists only while at least one child is live,
+ *  so a process with no children keeps default SIGINT handling. */
+
+type Killable = { kill(signal: NodeJS.Signals): unknown };
+
+const live: Killable[] = [];
```
egonSchiele: **This is the second signal-forwarding mechanism in this directory.** `spawnRunner.ts:32-52` already has `superviseTree`, and it is the same pattern down to the details: a module-level array of live things, one listener installed on first registration, a stop function that does `indexOf` then `splice`.

The moment you factor a mechanism out into its own file is the moment to ask whether it should be the existing one. Worth noting what the fork path gives up by staying separate — `superviseTree` also forwards SIGTERM and SIGHUP, and reaps with SIGKILL on any process exit, while the forked children here get SIGINT only and nothing on exit. That gap predates this PR, so I am not asking you to close it, but two near-identical supervisors sitting side by side means the next person fixing one will not know to fix the other.

If they should stay separate, a line in each pointing at the other and saying why would do it.

packages/agency-lang/lib/eval/run/sigintForwarder.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+function fakeChild() {
+  const signals: string[] = [];
+  return { signals, kill: (signal: string) => signals.push(signal) };
+}
+
+describe("forwardSigintTo", () => {
+  it("installs one listener for many children and removes it when the last one leaves", () => {
+    const before = process.listenerCount("SIGINT");
+    const children = Array.from({ length: 15 }, fakeChild);
```
egonSchiele: Nit: `process.emit("SIGINT")` fires every SIGINT listener in the worker, not just this module's, so the test is coupled to whatever else happens to be installed. Capturing `before` keeps the count assertions honest, but the emit itself still reaches other listeners.

Also, if an assertion above the last `stop()` throws, the remaining listener leaks into the rest of the file. A `try`/`finally` around the body that drains `stops` would keep a failure here from cascading into the next test.

egonSchiele (commented):
Nice cleanup of a real annoyance, and factoring it into its own file with a unit test is the right shape. Two things before this goes in: one is a behavior change I think is unintended and user-visible, the other is an altitude question.

egonSchiele:
Both points taken; addressed in 163c669a9.

**Altitude.** You were right that it was a second copy of `superviseTree`. There is now one mechanism, `lib/eval/run/childSupervisor.ts`, used by both `spawnRunner.ts` and `subprocess.ts`. `superviseTree` is gone. Forked children pick up what the spawn path had: SIGTERM/SIGHUP forwarding and SIGKILL reaping on exit.

**Second Ctrl-C.** The listeners exist only while a child is live. The first terminating signal is forwarded so the in-flight test can fold in. A second one is a force quit: SIGKILL to every live child, then `exit(128 + signal)`, so 130 for Ctrl-C, immediately. One judgment call to flag: I exit explicitly rather than restore Node's default handling. Restoring default would orphan detached command trees, because a default signal death runs no `exit` hook and nothing but this supervisor can reach a detached tree (`eval-command-agents.md` covers why they are detached). The user-visible outcome is what the old code measured: process gone at once, exit code 130. Note that the old spawn path had this bug too: `superviseTree` installed its listeners forever, so after any command-target run the second Ctrl-C was already being swallowed.

**Test coupling.** `makeChildSupervisor(proc)` takes the process slice it needs (`on`, `removeListener`, `exit`); `supervise` is the instance bound to the real `process`. The tests run on a private `EventEmitter` with a recording `exit`, so nothing is emitted on the real process and a failing assertion cannot leak a listener. Cases: 15 children share one listener per signal and it goes away when the last one settles; second signal → SIGKILL + exit 130, with a settled child not signalled; a first signal is forgotten once all children settle; `exit` reaps with SIGKILL. The `spawnRunner` listener-count test now asserts idle+1 during runs and idle after.

Dev doc paragraph in `eval-command-agents.md` updated to point at the shared supervisor and the force-quit rule.

egonSchiele (commented):
Round 2. Both findings are addressed, and the second one went further than I asked — folding `superviseTree` and the forwarder into one `childSupervisor` is the right call, and forked children picking up SIGTERM/SIGHUP and the exit-time reap is a real gain. Exiting with the signal code rather than restoring default disposition is a better answer than the one I suggested, and the reason in the commit message (a default signal death runs no `exit` hook, so detached trees get orphaned) is convincing. `spawnRunner.test.ts` also got stronger rather than weaker — it now pins both the live count and the idle count.

One thing left, and it is the same hole in a narrower place: the force quit does not survive a batch boundary, which is the one workload this PR was written for. Details inline, plus a stale comment.

packages/agency-lang/lib/eval/run/childSupervisor.ts
```diff
@@ -0,0 +1,78 @@
...
+    signalled = true;
+    for (const kill of live) kill(signal);
+  };
+  const listeners = Object.fromEntries(
+    TERMINATING.map((signal) => [signal, () => onSignal(signal)]),
+  );
+
+  return (kill) => {
+    if (live.length === 0) {
+      signalled = false;
```
egonSchiele, line 56: **`signalled = false` puts the force quit out of reach during `optimize`.**

I can see from the test at `childSupervisor.test.ts:60` that forgetting is deliberate, so this is a design question rather than a missed case. But it lands badly in exactly the workload that motivated this PR.

`baseOptimizer.ts` has no interrupt handling — grepping it for SIGINT, interrupt, or abort turns up nothing but unrelated prose. It calls `runSuite` once per input (`baseOptimizer.ts:397`, suite-of-one, inside a `Promise.all`), which is where the 15-and-21 children in your description come from. `runSuite` honors its own `interrupted` flag and returns cleanly, and then the optimizer just starts the next candidate. That drains `live` to zero and re-arms here with `signalled` reset.

So: Ctrl-C, the batch dies, the next candidate starts, Ctrl-C again — and the second press is treated as a first signal. Forever. I ran this against the real module with a stand-in process:

```
BASELINE (child stays live)  forwarded: [ 'SIGINT', 'SIGKILL' ] exit calls: [ 130 ] -> FORCE QUIT
OPTIMIZER (batch settled)   candidate1: [ 'SIGINT' ] candidate2: [ 'SIGINT' ] exit calls: [] -> NO FORCE QUIT
```

To be fair to the diff, this is not a regression: main was racy in the same spot, since a newly started `runSuite` would register a fresh `process.once` that absorbed the second press. The point is that this PR adds an explicit force quit, and this is the one place it does not reach.

I think sticky is right here, i.e. just drop line 56. The only way `signalled` gets set is a terminating signal arriving while a child was live, which is a genuine user interrupt — there is no path where it gets set spuriously and then needs clearing. A signal arriving with nothing live never reaches this code at all, since the listeners are not installed and default handling takes it. Once someone has asked to interrupt, the next press wanting to be final seems like the safe reading.

If you keep the forgetting, the test at line 60 is the one to extend: it currently stops at asserting the second signal is forwarded, so it does not show what a user pressing Ctrl-C twice through a batch boundary actually experiences.

packages/agency-lang/lib/eval/run/childSupervisor.ts
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * Signals are forwarded to every live child; the listeners exist only while
+ * a child is live, so an idle process keeps default handling. The first
+ * terminating signal is forwarded and the children get to settle (`eval run`
+ * folds the interrupted test into the run directory). A second one is the
+ * user asking to force quit: every child is killed with SIGKILL and the
+ * process exits with the signal's conventional code, immediately. Exiting
+ * that way instead of restoring default handling matters for spawned
+ * command trees: they are detached, so nothing but this supervisor reaches
+ * them, and a default signal death runs no `exit` hook to reap them.
```
egonSchiele, line 17: Two small things while you are in here.

`runSuite.ts:128-129` is now stale, and it is the comment I quoted last round as the contract:

```
// ... `once`, so a second
// Ctrl-C gets default handling — immediate death.
```

Default handling is no longer what makes the second Ctrl-C final — this supervisor is. The user-facing `press Ctrl-C again to force quit` message is still true (modulo the batch-boundary case above), but anyone reading that comment will go looking for a default-disposition mechanism that this PR removed. Worth pointing it at `childSupervisor.ts` instead.

Nit: the doc paragraph in `eval-command-agents.md` was not re-wrapped after the insert, so `terminating signal is a force quit: SIGKILL to every live child, then exit. The one hole no supervisor` runs well past the ~70 columns the rest of that file wraps at.

egonSchiele (commented):
Inline detail for the round-2 comments above.

PR #923, opened 2026-08-26
Add a writing subagent to the coordinator brain, and raise the reflection clamp

egonSchiele:
## What this does

Two small changes.

**1. A `writing` subagent for the coordinator brain.** `agency agent --agent writing` now goes straight to a prose reviewer, and the main coordinator can call `writingAgent` as a tool. The reviewer itself is the one in `std::agents/writing/review` (the prompt we optimized in #920), so the subagent only does the agent-shaped part: one `llm` call turns the chat message into a `WritingRequest` (a file path or pasted text, who the text is for, and whether to apply fixes). By default it reads the file with `read(useAgentCwd: true)` and returns the findings from `renderFeedback`. When the message asks for the fixes to be applied, one more `llm` call rewrites the text under the findings and `write` saves it back.

The intended use is from Claude Code or the shell:

```
agency agent --agent writing --policy with-writes -p "Review docs/dev/foo.md. It is for a developer new to the codebase."
agency agent --agent writing --policy with-writes -p "Review docs/dev/foo.md and apply the fixes to the file."
```

Files: new `brains/coordinator/subagents/writing.agency`; the coordinator registers it as a tool, a `startTargets` entry, and a `coordinatorRunTurn` arm; the two main prompts describe it; README, `docs/dev/agents/agent-brains.md`, and the registry test expect the new target.

**2. Raise the reflection clamp.** `lib/optimize/reflectionFeedback.ts` clamped each example to 2,000 characters, with a 600-character output preview and 400 characters per grader. On the writing-review run that meant the GEPA proposer saw a preview of the output and about one sentence per grader, which is why its rewrites were cheap and unspecific. Now 12,000 / 6,000 / 1,500. Mutator cost per iteration on gpt-5 goes from about $0.01 to about $0.05.

## Checked

- `pnpm run typecheck`, `pnpm run lint:structure`, prettier: clean.
- `agency test lib/agents/agency-agent/tests/registry.agency` passes with the new target list.
- `vitest run lib/optimize/reflectionFeedback.test.ts lib/optimize/optimizers`: 30 pass.
- Two real one-shot runs on a scratch paragraph: the review-only run returned six findings for $0.06; the apply run rewrote the file for $0.08.

## Known issue

In the apply run, the rewriter copied a finding's suggested wording into the file even though that wording was the reviewer's guess at what the original meant, not something the original said. The apply prompt says to keep every fact; it may need a line like "do not add facts a finding only guesses at". Apply is opt-in, so I left this for a follow-up rather than tune the prompt blind.

### DIFF
--- a/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/writing.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/writing.agency
@@ -1,97 +1,24 @@
 import { writingReviewAgent } from "std::agents/writing/review"
-import { Feedback, renderFeedback } from "std::agents/lib/feedback"
+import { renderFeedback } from "std::agents/lib/feedback"
 import { setMemoryId } from "std::memory"
 
-// The review itself lives in std::agents/writing/review, so users get the
-// same reviewer. What is left here is the agent-shaped part: working out
-// from a chat message which text to review, who it is for, and whether the
-// user wants the findings applied or only reported.
-
-type WritingRequest = {
-  // A file to review, relative to the agent working directory, or "" when
-  // the text is in the message itself.
-  path: string;
-  // The text to review when it was pasted into the message, else "".
-  text: string;
-  // Who the text is for and what it must get across, in the user's words,
-  // or "" when the message does not say.
-  task: string;
-  // True only when the user asked for the text to be rewritten or fixed,
-  // not just reviewed.
-  apply: boolean
-}
+// The review itself lives in std::agents/writing/review. That reviewer
+// reads the files a message names and applies fixes only when asked, so
+// all that is left here is the thread and memory an agent turn needs.
 
 export def writingAgent(userMsg: string): string {
   """
   Review prose for readability: passages a reader will misread, lose the
-  thread in, or did not need. Reports findings by default; rewrites the
-  text only when asked to.
+  thread in, or did not need. Reports findings by default; changes the
+  files only when the message asks for the fixes to be applied.
 
-  @param userMsg - the text to review or the path of a file holding it, who
-    the text is for if known, and whether to apply the fixes.
+  @param userMsg - the text to review or the files holding it, who the
+    text is for if known, and whether to apply the fixes.
   """
   let reply = ""
   thread(label: "writing", summarize: true, session: "writing") {
     setMemoryId("writing")
-    const request: WritingRequest = llm(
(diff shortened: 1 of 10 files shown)


packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/writing.agency
```diff
@@ -0,0 +1,97 @@
...
+  thread in, or did not need. Reports findings by default; rewrites the
+  text only when asked to.
+
+  @param userMsg - the text to review or the path of a file holding it, who
+    the text is for if known, and whether to apply the fixes.
+  """
+  let reply = ""
+  thread(label: "writing", summarize: true, session: "writing") {
+    setMemoryId("writing")
+    const request: WritingRequest = llm(
```
egonSchiele: what if the user requests a review of multiple files, or of something that isn't a file at all, but is inline text? I'd suggest giving the writingReviewAgent itself access to the `read` tool so it can read files, telling it that by default it should not write changes to files, but also giving it a `write` tool so it can write to files if needed. I'm not sure what to do about reviewing PRs on GitHub... will need to think about that more. But the point is, we should make the writingReviewAgent itself more self-contained, so the functionality can be re-used.

egonSchiele: One option would be to allow users to pass in an optional list of additional tools that the agent should get. Then, inside the agency agent, we could pass the writing review agent some other agents as tools, and these agents would help the writing review agent get additional context if needed (such as access to github PRs)

egonSchiele: Done in the second commit. The reviewer itself now takes either the text or a request naming the files, and carries the file tools (read, ls, glob, grep, write, edit). The prompt tells it to read any files the request names, never to write or edit unless the request asks for the fixes to be applied, and to apply them with `edit` one finding at a time when it does. That rule sits in the fixed part of the prompt, not in the optimized system prompt, so an optimize run cannot mutate it away. `extraTools` was already a parameter, so passing other agents in (a GitHub PR reader, say) needs no new plumbing.

The coordinator subagent is now a pass-through: thread, memory, `writingReviewAgent(work: userMsg)`, render. The whole-file rewrite is gone.

Checked with two real runs on a scratch file: review-only left the file byte-identical (cmp), and the apply run edited it and explicitly declined to define two undefined terms because the source gave no facts for them.

PR #925, opened 2026-08-26
agency-coding: add the uses-match test and AST idiom graders

egonSchiele:
## What this does

Finishes the `uses-match` test that was started in `evals/agency-coding/uses-match/test.json`, and adds the first "how is it written" graders to the coding suite.

**The test.** `files/foo.agency` is seeded into the writer's working directory. It exports `foo(effect, dir)`, which raises `std::read`, `std::write`, or `std::email` depending on `effect`, carrying `dir` in the data for the first two. The assignment asks for an exported `callFoo(effect, dir)` that calls `foo` inside a handler which approves reads, approves writes to `"."`, and rejects other writes and email with two specific messages. The notes moved from `files/` to `graderFiles/`, so the writer no longer sees the answer.

**Two things I learned building the harness, both recorded in the README and the notes:**

- The interrupt payload is `data.data.dir`, not `data.dir`.
- `agency test --reject '*'` (what every holdout runs under) vetoes every inner `approve()`, because every handler in the chain runs. A holdout can never observe an approval. It can tell rejections apart by their message, so the assignment asks for specific messages on the two reject arms, and the approve cases check for the outer handler's generic `interrupt rejected` text (which proves the solution's handler did not reject them itself). Verified: the reference passes 4/4, an if-chain version with the same behaviour passes 4/4, and a reject-everything handler fails 2/4.

**The graders** (`lib/idiomGraders.ts`, wired through the test's `graders.ts`): each parses the saved `solution.agency` with the real parser (`lower: false`, so `match` survives) and inspects the AST. Four binary graders: `match-on-effect` (a `match` on `data.effect` inside a handler), `no-if-chain-on-effect`, `guard-arm` (the conditional write is a guard arm), and `match-on-result` (the Result is unwrapped with `match`). The unit test grades the idiomatic fixture (4/4 pass), the if-chain fixture (4/4 fail), and a missing file (all fail). They are advisory; the harness is the gate, so they only run on solutions that behave correctly.

**First real run** (`eval run --test uses-match`, $0.04): the stdlib coding agent wrote

```
} with {
  on std::read(data) { approve() }
  ...
```

which is not Agency syntax, and saved it as done. Score 0, for the right reason. So the test already catches something on the owner's list ("critical that the coding agent understands syntax for interrupts and handlers").

## Checked

`pnpm run typecheck`, `pnpm run lint:structure`, prettier, `vitest run evals/agency-coding/lib` (3 pass), the three hand-run holdout verifications above, and one real `eval run` + `eval grade` of the test.


packages/agency-lang/evals/agency-coding/lib/fixtures/uses-match-idiomatic.agency
```diff
@@ -0,0 +1,20 @@
...
+    const result = foo(effect, dir)
+    outcome = match(result) {
+      success(v) => "approved"
+      failure(e) => "rejected: ${e}"
+    }
+  } with (data) {
+    return match(data.effect) {
+      "std::read" => approve()
+      "std::write" if (data.data.dir == ".") => approve()
+      "std::write" => reject("only writes to the current directory are allowed")
```
egonSchiele: just `reject()` no message for each reject, please

packages/agency-lang/evals/agency-coding/lib/fixtures/uses-match-if-chain.agency
```diff
@@ -0,0 +1,26 @@
...
+    }
+  } with (data) {
+    if (data.effect == "std::read") {
+      return approve()
+    }
+    if (data.effect == "std::write") {
+      if (data.data.dir == ".") {
+        return approve()
+      }
+      return reject("only writes to the current directory are allowed")
```
egonSchiele: same, no message for reject

packages/agency-lang/evals/agency-coding/lib/idiomGraders.ts
```diff
@@ -0,0 +1,159 @@
+// Graders for HOW a solution is written, for tests where the holdout
```
egonSchiele: I think an LLM Judge grader would be significantly easier. Does that not work?

packages/agency-lang/evals/agency-coding/uses-match/graderFiles/notes.md
```diff
@@ -0,0 +1,47 @@
+Example solution:
```
egonSchiele: should this file be deleted?

packages/agency-lang/evals/agency-coding/uses-match/holdout/uses-match-holdout.agency
```diff
@@ -0,0 +1,26 @@
+// Held-out oracle the writer never sees. Imports the file evalMain saved.
```
egonSchiele: same comment -- should we run any tests? Or just use LLM judge?

packages/agency-lang/evals/agency-coding/lib/idiomGraders.ts
egonSchiele: It does work, and it is much simpler. I went with AST checks for determinism, but that was my call, not a need. Switched: lib/idiomGraders.ts, its fixtures and tests, and the holdout are gone; graders.ts is one rubric judge (lib/idiomJudge.ts reads the saved solution and hands it to ctx.judges.rubric with a standard and a reference solution). Checked on the two real runs: the judge correctly spotted the invented on-std::read handler syntax, the if inside the write arm, and the reject messages. I also told it that code which is not valid Agency scores 0, since it was giving 0.25 and 0.5 to solutions in a syntax that does not exist.

packages/agency-lang/evals/agency-coding/uses-match/holdout/uses-match-holdout.agency
egonSchiele: No tests now; the judge reads the source. That also removes the awkward part of the harness for handler tests: agency test runs under --reject '*' and every handler in the chain runs, so an inner approve() is always vetoed and a holdout could only ever see rejection messages.

packages/agency-lang/evals/agency-coding/uses-match/graderFiles/notes.md
egonSchiele: Deleted. What it recorded now lives in the reference solution inside graders.ts.

packages/agency-lang/evals/agency-coding/lib/fixtures/uses-match-idiomatic.agency
egonSchiele: Done: bare reject() everywhere, and the assignment no longer asks for messages.

packages/agency-lang/evals/agency-coding/lib/fixtures/uses-match-if-chain.agency
egonSchiele: Done (this fixture is gone with the AST graders).

egonSchiele:
Fourth commit, from looking into where the handler syntax was coming from: it was not coming from anywhere in the repo. The coding agent did the right thing and asked the docs tool for `handlers.md` (twice) and `effects-and-raises.md`, but `evalMain` ran it `with reject`, which rejected those `std::read` interrupts along with everything else. So it guessed, and its guess was `with { on notes::archive(x) { ... } }`.

The four stdlib `evalMain` nodes now run their worker under a shared `evalHandler` (`std::agents/lib/shared`) that approves reads under the packaged docs directory (`bundledDocsDir()`, new in `std::skills`) and rejects everything else, including workdir greps and budget trips. Rerun of both tests with the docs readable: both solutions are valid Agency (`handle { } with (intr) { }`), and the handler-chain doc comment now says both handlers run and a reject wins. What is left are idiom misses (if-chains instead of `match`, a missing `raises` clause), which is what the tests are for. Scores 0.5 and 0.

Also tightened the two judge standards: "invalid Agency" now means what the parser refuses, so an idiom miss loses its quarter rather than zeroing the test, and `return reject()` is a bare reject. Noted in `docs/dev/evals/eval-grading.md`.

packages/agency-lang/docs/dev/evals/eval-grading.md
```diff
@@ -72,6 +72,17 @@ directions ("pass --input" or "make the node take no parameter"). A test
 with no input reaches the child as a run instruction without `input`,
 which `resolveNodeCallArgs` already treats as a bare call.
 
+**The stdlib entry nodes run their worker under `evalHandler`.** Each
```
egonSchiele: Why have you added this here? Is it because you think this information will be useful for people modifying this code in the future? Or are they notes you have taken for yourself while working in this space? If it is the latter, then it is not useful to add to this doc.

packages/agency-lang/evals/agency-coding/handler-chain/graders.ts
```diff
@@ -0,0 +1,58 @@
+import { idiomJudge } from "../lib/idiomJudge.js";
+
+export default [
+  idiomJudge({
+    name: "handler-syntax-and-chain",
+    standard: `Four things, each worth a quarter. (1) The interrupt is raised with the
```
egonSchiele: > Four things, each worth a quarter

Hate this phrasing

packages/agency-lang/evals/agency-coding/handler-chain/test.json
```diff
@@ -0,0 +1,8 @@
+{
+  "description": "Raise a named interrupt, wrap the call in two nested handlers that disagree, and say what happens. Tests the raise/raises/handle syntax and the rule that every handler in the chain runs and any reject wins.",
```
egonSchiele, line 2: > say what happens

So tired of this phrasing

packages/agency-lang/evals/agency-coding/lib/idiomJudge.ts
```diff
@@ -0,0 +1,42 @@
...
+import * as fs from "fs";
+import * as path from "path";
+
+import { binary, grader, type Grader, type Test } from "agency-lang/eval";
+
+/** Mirrors `CodingEvalInput` in stdlib/agents/agency/coding.agency. */
+type CodingInput = { assignment: string; outFile: string };
+
+/** The saved solution, read from the run's workdir; "" when missing. */
+function solutionOf(workdir: string, test: Test<CodingInput>): string {
```
egonSchiele: We write this function to read files from the workdir so frequently, I wonder if it should just be one of the functions that gets passed to the grader, like workdir, test, and judges.

packages/agency-lang/docs/dev/evals/eval-grading.md
egonSchiele: The second half was my notes. Cut to the one sentence a maintainer needs: which handler the entry nodes run under, where it lives, what it approves and rejects.

packages/agency-lang/evals/agency-coding/handler-chain/graders.ts
egonSchiele: Gone. Both standards now end with your sentence: all N of these points count equally towards the final score.

packages/agency-lang/evals/agency-coding/handler-chain/test.json
egonSchiele: Gone from the assignment and the description.

packages/agency-lang/evals/agency-coding/lib/idiomJudge.ts
egonSchiele: Done: ctx.workdirFile(relativePath) on the grader context (lib/eval/grading/workdirFile.ts), used by idiomJudge and the agency-review graders. Empty string when the file is missing or the path escapes the workdir.

packages/agency-lang/evals/writing-review/describe-not-show/graderFiles/notes.md
```diff
@@ -0,0 +1,5 @@
+Context: a rubric for an LLM judge. The subject is code syntax.
+
+- The whole text describes syntax in prose, with code fragments inside parentheses inside sentences. One code block showing the handler, with the guarded arm in it, would say all of it at once. Lead with the example; then the checklist can point at it ("as shown above").
```
egonSchiele: > Lead with the example; then the checklist can point at it

No semicolons, and rephrase the second half. The checklist is not a person who can point.

packages/agency-lang/evals/writing-review/one-fact-per-sentence/graderFiles/cleaned.md
```diff
@@ -0,0 +1 @@
+Write a module that exports two functions. First, `archiveNotes(count: number): Result<string>`: before doing anything it raises an interrupt with the effect name `notes::archive`, carrying `count` in the interrupt's data. If the interrupt is approved, it returns success("archived ${count}"). Second, `runArchive(count: number): string`. This function calls archiveNotes inside an inner handler that approves `notes::archive`, but that inner handler sits inside an outer handler. The outer handler rejects `notes::archive` when the count is greater than 10, and passes on it otherwise. runArchive returns "archived" when archiveNotes succeeded and "rejected" when it was rejected. Above runArchive, write a doc comment that says, for count = 11, which of the two handlers run, what runArchive returns, and why. Use only Agency standard library (std::) imports, or no imports at all.
```
egonSchiele: Please add some line breaks to the clean version as well instead of making it one big paragraph. I would put the descriptions for the two functions on separate lines to make this easier to read. Also add to the notes that part of the cleanup we're doing is moving stuff to new lines to make it easier to read.

packages/agency-lang/evals/writing-review/own-debugging-notes/test.json
```diff
@@ -0,0 +1,7 @@
+{
+  "description": "A developer-doc paragraph that records the author's debugging story instead of what a future maintainer needs.",
+  "input": {
+    "assignment": "Add a paragraph to docs/dev/evals/eval-grading.md, the developer notes for eval grading, recording that the stdlib eval entry nodes now run their worker under a shared handler. The readers are people changing this code in the future.",
```
egonSchiele: > The readers are people changing this code in the future

Does the agent still do the task correctly if we don't include this line? I don't want to have to include this line every time I ask an agent to do a task like this.

packages/agency-lang/stdlib/agents/lib/shared.agency
```diff
@@ -1,7 +1,19 @@
...
+  @summary Helpers every std::agents worker uses: context folding, model
+  overrides, and the handler an eval entry node runs its worker under.
 */
+import { bundledDocsDir } from "std::skills"
+
+/** The handler for an eval entry node. A worker under evaluation may read
+  the docs that ship in this package (that is how it looks up syntax), and
+  nothing else: every other interrupt, including a budget trip, is
+  rejected so the caps stay caps and no write or command escapes. */
+export def evalHandler(intr) {
```
egonSchiele: This is a pretty weird name for this handler. You're using it in the code for different agents in the agency standard library. evals are just one of the places these agents get used. If the users use them in any other context, they won't know why this handler is called "eval" handler.

packages/agency-lang/evals/writing-review/describe-not-show/graderFiles/notes.md
egonSchiele: Rewritten: "Put the example first. Then each item in the checklist can refer to it with 'as shown above'."

packages/agency-lang/evals/writing-review/one-fact-per-sentence/graderFiles/cleaned.md
egonSchiele: Done. The clean version now has one paragraph per function, and the doc-comment requirement and the import rule on their own lines. The notes have a new bullet saying the cleanup includes splitting the assignment into paragraphs so each requirement can be found.

packages/agency-lang/evals/writing-review/own-debugging-notes/test.json
egonSchiele: Tested: removed the sentence and ran the test twice. The reviewer still flags the debugging anecdote for cutting both times (scores 0.72 and 0.63, against 0.77 with the sentence). What it misses either way is the follow-up note ("reads and greps of the working directory are still rejected"), which it wants to reword rather than cut. The sentence is gone from the assignment.

packages/agency-lang/stdlib/agents/lib/shared.agency
egonSchiele: Agreed. Renamed to docsOnlyHandler, with the doc comment saying what it does (approves reads of the packaged docs, rejects every other interrupt) and that the eval entry nodes are one user of it.

PR #927, opened 2026-08-26
Give the GEPA proposer the goals, and cover every optimizer input mode in CI

egonSchiele:
## Why

The post-merge real-LLM workflow was red on `main` at the `gepa-judge` row of the optimizer efficacy test: three attempts, "champion 0 <= baseline 0". Running it locally showed the rewritten prompts. Every one had the shape "be chatty, unless the user asks for brevity":

> By default, be warm and chatty, and set the scene briefly before answering. However, whenever the user gives an explicit output-formatting instruction (for example: "Reply with only the capital city name and nothing else"...), follow that formatting exactly...

The fixture's prompt is `"${style} What is the capital of France?"`. Nobody asks for brevity, so the exception never fires.

The cause: GEPA's reflection prompt (`gepaReflect.agency`) never received the goal. It got the target, the judge's feedback, and the retry history. The only place the goal text appeared was quoted inside the judge's feedback, and the prompt frames everything as "task inputs given to the assistant", so the mutator read it as a user request. Greedy's prompt has a `GOALS:` section and passed every time.

## What changed

- `gepaReflect.agency` gets a `GOALS:` section, rendered by the same `renderGoalsSection` greedy uses (extracted from `buildMutatorSections`). With it, GEPA accepts on iteration 1 locally.
- The free-text type description said `keep every ${...} placeholder`; the mutator pasted that literal `${...}` into a value, which fails to parse as an Agency string (that was the wasted iteration 1). It now describes the placeholders without showing one.
- The efficacy test is a matrix now, one run per way `agency optimize` takes inputs and grades: `--goal` alone, a suite of test directories with per-test `graders.ts`, `--graders` override, `--validation-split`, gepa, a custom optimizer module, and a single test directory with the grader coming from `agency.json`. Before, only `--goal` plus `--graders` were exercised; the `test.json` + `graders.ts` shape was not run by CI anywhere.
- The suite (`tests/integration/eval-suite/capitals`) is written the way evals are written now. The no-LLM `eval-run` test (every PR) runs it and grades it with `eval grade` against `solution.agency` (mean 1) and `wrong.agency` (mean 0), so `graders.ts` discovery and `eval grade` are checked without a model.
- The fixture's bad baseline is now "explain the history in a paragraph" rather than "be warm and chatty".

## Verified

- All seven efficacy rows pass locally with real LLM calls and retries disabled (`OPTIMIZE_EFFICACY_RETRIES=0`).
- `tests/integration/eval-run/test.mjs` passes, including the always-wrong agent scoring 0.
- `pnpm run typecheck`, `pnpm run lint:structure`, prettier on changed files, and the optimize unit tests (new test: the proposer sees every minibatch input's goal).

Not covered by the matrix: `--suite` pointing at a single `{"inputs": [...]}` file (the loader is shared with `eval run`, which the eval-run test already exercises in that form) and `--validation-suite` (same code path as `--validation-split` after loading).

Also noticed, not changed: `docs/site/cli/optimize.md` says `--inputs` and `--validation-inputs`; the flags are `--suite` and `--validation-suite`.

### DIFF
--- a/packages/agency-lang/docs/dev/evals/writing-optimizers.md
+++ b/packages/agency-lang/docs/dev/evals/writing-optimizers.md
@@ -146,7 +146,7 @@ Note the `targetSet` carried on each candidate. It reflects *that candidate's* t
 A mutation is proposed by a model and applied to the source. There are two proposer front-ends, both returning a `MutationProposal` (`{ operations, rationale }`):
 
 - **`proposeMutation`** (`lib/optimize/mutator.ts`) is the greedy and example proposer. It renders TARGETS, GOALS, per-input FEEDBACK, and HISTORY into `lib/agents/optimize/mutatePrompt.agency`.
-- **`proposeReflective`** (`lib/optimize/gepaReflect.ts`) is GEPA's reflective proposer, backed by `lib/agents/optimize/gepaReflect.agency`. It renders the same TARGETS and GOALS sections (`renderTargetsSection`, `renderGoalsSection`), plus the minibatch's FEEDBACK and any retry HISTORY. GOALS is not optional: without it the model reads the graders' feedback as a one-off user request and keeps the current instruction, adding an exception to it ("be chatty, unless the user asks for brevity"), which never scores.
+- **`proposeReflective`** (`lib/optimize/gepaReflect.ts`) is GEPA's reflective proposer, backed by `lib/agents/optimize/gepaReflect.agency`. It renders TARGETS, GOALS, the minibatch's FEEDBACK, and HISTORY (validation errors from a rejected attempt).
 
 You hand `proposeValidMutation` two callbacks:
 
--- a/packages/agency-lang/docs/dev/evals/writing-optimizers.md
+++ b/packages/agency-lang/docs/dev/evals/writing-optimizers.md
@@ -146,7 +146,7 @@ Note the `targetSet` carried on each candidate. It reflects *that candidate's* t
 A mutation is proposed by a model and applied to the source. There are two proposer front-ends, both returning a `MutationProposal` (`{ operations, rationale }`):
 
 - **`proposeMutation`** (`lib/optimize/mutator.ts`) is the greedy and example proposer. It renders TARGETS, GOALS, per-input FEEDBACK, and HISTORY into `lib/agents/optimize/mutatePrompt.agency`.
-- **`proposeReflective`** (`lib/optimize/gepaReflect.ts`) is GEPA's reflective proposer, backed by `lib/agents/optimize/gepaReflect.agency`.
+- **`proposeReflective`** (`lib/optimize/gepaReflect.ts`) is GEPA's reflective proposer, backed by `lib/agents/optimize/gepaReflect.agency`. It renders the same TARGETS and GOALS sections (`renderTargetsSection`, `renderGoalsSection`), plus the minibatch's FEEDBACK and any retry HISTORY. GOALS is not optional: without it the model reads the graders' feedback as a one-off user request and keeps the current instruction, adding an exception to it ("be chatty, unless the user asks for brevity"), which never scores.
 
 You hand `proposeValidMutation` two callbacks:
 
--- a/packages/agency-lang/lib/agents/optimize/gepaReflect.agency
+++ b/packages/agency-lang/lib/agents/optimize/gepaReflect.agency
@@ -13,6 +13,7 @@ type OptimizeMutationProposal = {
 
 node gepaReflect(
   targets: string,
+  goals: string,
   feedback: string,
   history: string,
 ): OptimizeMutationProposal {
@@ -22,6 +23,10 @@ node gepaReflect(
 
   OPTIMIZE TARGETS:
   ${targets}
+
+  GOALS (what the assistant's output must achieve; these outrank the current
+  instructions, so replace whatever in them works against a goal):
+  ${goals}
 
   The following are examples of task inputs given to the assistant, the assistant's
   response for each, its execution trace (errors and tool calls), and feedback on how
--- a/packages/agency-lang/tests/integration/optimize-efficacy/test.mjs
+++ b/packages/agency-lang/tests/integration/optimize-efficacy/test.mjs
@@ -56,7 +56,10 @@ const RUNS = [
   {
     name: "suite-graders-override",
     flags: `--suite ${q(SUITE)} --graders ${q(GRADER)}`,
-    check: ({ runDir }) => expectNoGraderNames(runDir, ["bare-paris", "bare-tokyo"]),
+    check: ({ runDir }) => {
+      expectNoGraderNames(runDir, ["bare-paris", "bare-tokyo"]);
+      expectGraderNames(runDir, ["only-city-name"]);
+    },
   },
   // A held-out validation input picks the champion.
   {
@@ -65,6 +68,9 @@ const RUNS = [
     check: ({ summary }) => {
       if (summary.validationObjective !== 1) {
         throw new Error(`validation objective ${summary.validationObjective}, expected 1`);
+      }
+      if (!summary.iterations.some((i) => i.validationObjective !== undefined)) {
+        throw new Error("no iteration recorded a validation objective");
       }
     },
   },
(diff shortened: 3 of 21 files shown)


packages/agency-lang/docs/dev/evals/writing-optimizers.md
```diff
@@ -146,7 +146,7 @@ Note the `targetSet` carried on each candidate. It reflects *that candidate's* t
 A mutation is proposed by a model and applied to the source. There are two proposer front-ends, both returning a `MutationProposal` (`{ operations, rationale }`):
 
 - **`proposeMutation`** (`lib/optimize/mutator.ts`) is the greedy and example proposer. It renders TARGETS, GOALS, per-input FEEDBACK, and HISTORY into `lib/agents/optimize/mutatePrompt.agency`.
-- **`proposeReflective`** (`lib/optimize/gepaReflect.ts`) is GEPA's reflective proposer, backed by `lib/agents/optimize/gepaReflect.agency`.
+- **`proposeReflective`** (`lib/optimize/gepaReflect.ts`) is GEPA's reflective proposer, backed by `lib/agents/optimize/gepaReflect.agency`. It renders the same TARGETS and GOALS sections (`renderTargetsSection`, `renderGoalsSection`), plus the minibatch's FEEDBACK and any retry HISTORY. GOALS is not optional: without it the model reads the graders' feedback as a one-off user request and keeps the current instruction, adding an exception to it ("be chatty, unless the user asks for brevity"), which never scores.
```
egonSchiele: Please tell me about the change you have added to this file. Is this necessary information for future readers, or are these implementation notes that you took for yourself? If it is the latter, please undo this change.

packages/agency-lang/lib/agents/optimize/gepaReflect.agency
```diff
@@ -23,6 +24,10 @@ node gepaReflect(
   OPTIMIZE TARGETS:
   ${targets}
 
+  GOALS (what the assistant's output must achieve; these outrank the current
```
egonSchiele, line 27: This wording makes the goals outrank everything, which is right when there are goals. Two of the suites we optimize have none: every test in `evals/agency-coding` and `evals/writing-review` grades through `graders.ts` and has no `goal` field. `renderGoalsSection` then renders one `- [test-id] ` bullet per input with nothing after it, under a heading that says these outrank the instructions. Greedy has the same rendering, but its heading is a bare `GOALS:`.

Suggest either omitting the section when every goal is empty, or rendering a line such as `(no goals given: the graders' feedback below is the standard)`. A small unit test alongside the new one, with goal-less inputs, would pin whichever you pick.

packages/agency-lang/tests/integration/optimize-efficacy/test.mjs
```diff
@@ -34,23 +40,94 @@ const q = (s) => JSON.stringify(s);
...
 
-function runOnce({ name, flags }) {
+// A directory under this package (so `agency-lang` still resolves) whose
+// agency.json is this package's plus `extra`.
+function projectWithConfig(extra) {
+  const dir = join(runsDir, "project");
+  mkdirSync(dir, { recursive: true });
+  const base = JSON.parse(readFileSync(join(PACKAGE_DIR, "agency.json"), "utf-8"));
+  writeFileSync(join(dir, "agency.json"), JSON.stringify({ ...base, ...extra }, null, 2));
+  return dir;
```
egonSchiele, line 103: `{ ...base, ...extra }` is a shallow merge, so `extra.eval` replaces the package's whole `eval` block. The package `agency.json` has `eval.limits.wallClockSec: 900`, which this row then runs without. Merge `eval` one level down (`eval: { ...base.eval, ...extra.eval }`) so the row runs under the same limits as the others.

egonSchiele (commented):
Reviewed for correctness and security. The diagnosis is right and the fix is the right size: GEPA got the feedback without the goal, so the mutator treated the goal as one user's request. Passing the same `renderGoalsSection` greedy uses, plus the unit test that checks every minibatch input's goal reaches the proposer, is the fix I would have made. The matrix is a real improvement: before this the `test.json` + `graders.ts` shape had no CI coverage at all, and the eval-run scenario with a solution and an always-wrong agent is the check that proves the graders can fail.

Security: nothing new. `graders.ts` was already executed code from the suite directory, the shell commands are built from fixed strings and `JSON.stringify`-quoted paths under the package, and the temporary `agency.json` copies this package's own config. I found no way for suite content to reach a shell.

Three inline comments, one of which matters for the suites we use. Then, since you asked what else is worth covering on CI:

**Worth adding, no LLM needed (in `eval-run/test.mjs`):**

1. **A `holdout/` harness pair.** `evals/agency-coding` grades `sum-multiples` and `reverse-digits` with a hidden `<name>.agency` + `<name>.test.json` pair, run through `AgencyTestGrader` as a must-pass gate. Nothing in CI runs that path: harness discovery, the `agency test --agency-only --reject '*'` spawn, or the gate short-circuit in `gradeRun.ts` that skips advisory graders when a gate fails. A third capitals test with a holdout pair, graded against `solution.agency` (1) and `wrong.agency` (0, and no advisory rows), would cover all three.
2. **`graderFiles/`.** The writing-review suite's graders read `notes.md` and `cleaned.md` from `ctx.graderFiles`, and the run directory snapshots that directory. A function grader that reads a file from `graderFiles` and checks it was never seeded into the workdir would pin both halves.
3. **`eval grade <run> --suite <dir>` re-grading.** Re-grading an existing run with edited graders is how every rubric in #925 was tuned. It goes through `snapshotHarnessGraders` and the suite grader source, which the run-then-grade scenario does not touch.
4. **`--trials 2` and `--test <name>`.** Cheap, and the `mean ± SE` block in `formatGrade` has only unit coverage.

**Worth adding to the optimizer matrix:**

5. **Writeback.** Every row is `--no-writeback`. The writeback path is the one that edits user source, and #920 changed how it rewrites a `"""` value. One row on a temp copy of the fixture with writeback on, asserting the file changed and still parses, would guard it. It also exercises `sourceMutator` end to end rather than in unit tests only.
6. In the `suite-validation-split` row, also assert `summary.iterations.some((i) => i.validationObjective !== undefined)`: #920 fixed those columns being empty, and this is the only run that would notice a regression.

packages/agency-lang/tests/integration/optimize-efficacy/test.mjs
```diff
@@ -34,23 +40,94 @@ const q = (s) => JSON.stringify(s);
...
+    name: "suite-graders",
+    flags: `--suite ${q(SUITE)}`,
+    check: ({ runDir }) => expectGraderNames(runDir, ["bare-paris", "bare-tokyo"]),
+  },
+  // --graders replaces every test's own graders.
+  {
+    name: "suite-graders-override",
+    flags: `--suite ${q(SUITE)} --graders ${q(GRADER)}`,
+    check: ({ runDir }) => expectNoGraderNames(runDir, ["bare-paris", "bare-tokyo"]),
+  },
```
egonSchiele, line 63: This proves the suite's graders were replaced, but not that the override grader ran: an empty grade list would also pass. `expectGraderNames(runDir, ["only-city-name"])` (or whatever `onlyCityName.ts` names itself) alongside it makes the row fail if the override silently graded nothing. Same for the `one-test-config-graders` row.

packages/agency-lang/docs/dev/evals/writing-optimizers.md
egonSchiele: The why was a note to myself. Trimmed to one line naming the sections gepaReflect renders, matching the greedy bullet above it (d39c702e9).

packages/agency-lang/lib/agents/optimize/gepaReflect.agency
egonSchiele: Done in renderGoalsSection, so greedy gets it too: when no input has a goal it renders "(no goals given: the graders' feedback below is the standard)" instead of empty bullets. Unit test added in mutator.test.ts (d39c702e9).

packages/agency-lang/tests/integration/optimize-efficacy/test.mjs
egonSchiele: Fixed: eval is merged one level down, so the row keeps wallClockSec 900 (d39c702e9).

packages/agency-lang/tests/integration/optimize-efficacy/test.mjs
egonSchiele: onlyCityName.ts now names itself only-city-name, and both override rows assert expectGraderNames(runDir, ["only-city-name"]) as well. Re-ran those rows plus validation-split locally, all pass (d39c702e9).

PR #930, opened 2026-08-27
Cover holdout harness gates, eval grade --suite, and optimizer writeback in CI

egonSchiele:
## Why

Follow-up to #927, taking items 1, 3, and 5 from its review: three eval paths that CI never exercised.

## What changed

**Holdout harness pair as a gate.** New suite `tests/integration/eval-suite/capital-function/`: a `test.json`, an advisory `graders.ts` (`wrote-solution`), and a `holdout/capital-function-holdout.{agency,test.json}` pair. It is a separate suite from `capitals` because the optimizer refuses a baseline that fails a gate, so a holdout test cannot sit in the suite the optimizer rows use. Scenario I in `eval-run/test.mjs`:
- `solution.agency` (writes the module, no LLM): gate passes, objective 1, both the harness and the advisory grader graded it.
- `wrong.agency` (writes nothing): `eval grade` exits 2, `gatesPassed` is false, and the grade list holds only the harness. That proves the gate short-circuit in `gradeRun.ts` skipped the advisory grader.

**`eval grade --suite` re-grading.** Scenario H copies `capitals` into the temp dir, replaces both `graders.ts` with a half-credit grader named `regraded`, and re-grades the existing solution run. Only `regraded` may appear, and the mean must be 0.5.

**Optimizer writeback.** A `writeback` row in the efficacy matrix runs on a copy of the fixture under the runs dir, without `--no-writeback`. The copy must differ from the original, `agency ast` must still parse it, and its content must equal the champion source in `summary.json`.

`docs/misc/TESTING.md` describes the additions.

## Verified

- `tests/integration/eval-run/test.mjs`: all nine scenarios pass, including the two new ones.
- The `writeback` row passes with a real LLM call (`OPTIMIZE_EFFICACY_ONLY=writeback`, retries off).
- `pnpm run typecheck`, `pnpm run lint:structure`, prettier on changed files.

Not included: items 2 (`graderFiles/`) and 4 (`--trials` / `--test`) from the #927 review.


packages/agency-lang/tests/integration/eval-suite/capital-function/test.json
```diff
@@ -0,0 +1,5 @@
+{
+  "goal": "solution.agency exports def capitalOf(country: string): string and returns the right capital.",
+  "input": "Write an Agency module named solution.agency in your current working directory that exports def capitalOf(country: string): string, returning \"Paris\" for \"France\" and \"Tokyo\" for \"Japan\". No imports. Once the file is written you are done.",
+  "timeoutSec": 120
```
egonSchiele, line 5: This test needs `"harnessMustPass": true` once #931 lands. That PR makes a holdout pair an ordinary scored grader by default (a wrong solution scores its passing fraction, advisory graders still run, `eval grade` exits 0), and a test opts into the gate with this flag. Without it, every assertion in Scenario I about `wrong.agency` fails: exit code 2, `gatesPassed === false`, and the advisory grader being skipped.

With the flag, Scenario I becomes the CI proof that the opt-in works, which nothing else covers, so it is worth keeping exactly as written. Whichever of the two PRs merges second needs this one line.

packages/agency-lang/tests/integration/eval-run/test.mjs
```diff
@@ -427,6 +427,107 @@ node main(): string {
...
+    `re-grade with half-credit graders scored ${regrade.mean}, expected 0.5`,
+  );
+  console.log(
+    "[eval-run-integration] PASS: eval grade --suite re-scores a run with edited graders",
+  );
+
+  // ── Scenario I: a holdout harness pair as the gate ──
+  // A test directory with holdout/<name>.agency + <name>.test.json grades by
+  // running that harness against the agent's workdir through `agency test`.
+  // The harness is a must-pass gate, so when it fails the advisory grader
```
egonSchiele: After #931 this sentence is true only for a test that sets `harnessMustPass`; a bare holdout no longer gates. Suggest: "The suite sets harnessMustPass, so the harness is a gate: when it fails, the advisory grader beside it is never run."

packages/agency-lang/docs/misc/TESTING.md
```diff
@@ -470,8 +470,8 @@ AGENCY_USE_TEST_LLM_PROVIDER=1 pnpm run test:agency-js
...
+- **Eval-run test** (`tests/integration/eval-run/test.mjs`, every PR, no LLM) — `eval run` over an interrupting agent, a command target, `run --capture-workdir`, and a suite written as test directories (`tests/integration/eval-suite/capitals`: `test.json` + `graders.ts` per test). The suite is run and then graded with `eval grade` against a reference solution (mean 1) and an always-wrong agent (mean 0), so a grader that cannot fail is caught. The same run is re-graded with `eval grade --suite` pointing at a copy of the suite with half-credit graders (mean 0.5). A second suite, `tests/integration/eval-suite/capital-function`, grades through a `holdout/` harness pair: the reference solution passes the gate, and an agent that writes nothing fails it with exit 2 and no advisory grades.
```
egonSchiele: Same note as on the suite: once holdouts are scored by default, say here that `capital-function` opts into the gate with `harnessMustPass`, so a reader does not take this as the default behaviour.

egonSchiele (commented):
Read the whole diff and checked the two facts the assertions rest on: a grader function may return a boolean (`functionGrader.ts` coerces it to a binary grade), and `eval grade` exits 2 exactly when a gate failed. Both hold on `main` today, so the three scenarios are sound against `main`.

The one real issue is the collision with #931, which changes the default Scenario I is testing: holdout pairs stop being gates unless a test sets `"harnessMustPass": true`. One line in `capital-function/test.json` fixes it and turns Scenario I into the test of that opt-in, which #931 does not have. Inline comments mark the three spots (the test.json, the scenario comment, and the TESTING.md sentence). The PR body's reason for a separate suite ("the optimizer refuses a baseline that fails a gate") also stops applying without the flag, but keeping the suites apart is still the clearer layout.

Everything else reads right to me: Scenario H's `every(name === "regraded")` plus the exact 0.5 mean is a check that cannot pass by accident; the `writeback` row's three assertions (changed, parses, equals `championFiles`) are the right three, and `runOnce` already requires the champion to beat the baseline, so an unchanged file cannot slip through as a no-op champion.

packages/agency-lang/tests/integration/optimize-efficacy/test.mjs
```diff
@@ -90,8 +90,37 @@ const RUNS = [
       expectGraderNames(runDir, ["only-city-name"]);
     },
   },
+  // Writeback, the one path that edits user source. Runs on a copy of the
+  // fixture; the copy must change, still parse, and hold the champion value.
+  {
+    name: "writeback",
+    agent: copyOfFixture(),
```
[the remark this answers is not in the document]
egonSchiele: Fixed: the row now holds a function that copies the fixture, and runOnce calls it per attempt, so every attempt starts from the original source.

packages/agency-lang/tests/integration/eval-suite/capital-function/test.json
egonSchiele: Added after merging main (#931 is in). Checked that the flag is what Scenario I rests on: with the line removed, the run fails at "grading the agent that wrote nothing exited 0, expected 2"; with it, all nine scenarios pass.

packages/agency-lang/tests/integration/eval-run/test.mjs
egonSchiele: Reworded as suggested.

packages/agency-lang/docs/misc/TESTING.md
egonSchiele: Done: TESTING.md now says capital-function opts into the gate with harnessMustPass and that a holdout pair is scored by default.

Issue #929, opened 2026-08-26
guard over a loop returns success([]) when the time budget trips


egonSchiele:
A `guard(time: ...)` whose block loops over a slow call returns `success` with an empty array when the time budget trips, instead of a `timeoutFailure`. A block that makes one slow call, or one `sleep`, returns the failure as documented.

Repro (`summarize` sleeps 100 ms per character):

```
def summarize(doc: string): string {
  sleep(doc.length * 100)
  return "summary of ${doc}"
}

node main() {
  const one = guard(time: 300ms) {
    return summarize("a very long document indeed")
  }
  printJSON(one)   // success: false, type: timeoutFailure  (correct)

  const loop = guard(time: 300ms) {
    let out: string[] = []
    for (d in ["a very long document indeed", "x"]) {
      out.push(summarize(d))
    }
    return out
  }
  printJSON(loop)  // success: true, value: []  (expected a timeoutFailure)

  const comp = guard(time: 300ms) {
    return [summarize(d) for d in ["a very long document indeed", "x"]]
  }
  printJSON(comp)  // success: true, value: []  (expected a timeoutFailure)
}
```

Run with `agency run`, no handler, so the `std::guard` interrupt is rejected. Found while writing the `guards` coding eval; its reference guards each call separately, which behaves correctly.

Issue #928, opened 2026-08-26
match and if-then-else expressions inside an as-block evaluate to null


egonSchiele:
A `match` expression, or an `if ... then ... else` expression, assigned to a `const` inside a trailing block (`map(xs) as p { ... }`) evaluates to `null`. The same expression at the top level of a node works.

Repro:

```
node main() {
  const xs = map([1, 2]) as p {
    const s = "lit"
    const m = match (p) {
      1 => "one"
      _ => "other"
    }
    const e = if p == 1 then "one" else "other"
    return "${s}/${m}/${e}/${p}"
  }
  print(xs.join(","))
}
```

Prints `lit/null/null/1,lit/null/null/2`. Expected `lit/one/one/1,lit/other/other/2`.

`typecheck` reports nothing. Found while writing a reference solution for the `loop-forms` coding eval; the reference now avoids value-producing `match` inside a block.

Issue #926, opened 2026-08-26
Accept the handler syntax models reach for by default (handle (...) with { on effect(x) { ... } })


egonSchiele:
## What happened

While building the `evals/agency-coding` handler tests (#925), the stdlib coding agent (gpt-5-mini) wrote the same handler syntax in three separate runs when it had no documentation in front of it:

```
let res: Result<string> = handle (foo(effect: effect, dir: dir)) with {
  on std::read(data) {
    approve()
  }
  on std::write(data) {
    if (data.dir == ".") {
      approve()
    } else {
      reject()
    }
  }
  on std::email(_) {
    reject()
  }
  on _ {
    reject()
  }
}
```

and, for a custom effect:

```
def archiveNotes(count: number): Result<string> {
  raise notes::archive({ count: count })
  ...
}
```

Nothing in the repo teaches this form; it is the model's default, a blend of Koka/OCaml-style effect handlers and a `match`. Once the agent could read `handlers.md` it wrote the real syntax on the first try, so the guess only shows up when the docs are out of reach. The full runs are in the statelogs under `runs/agency-coding-handler-chain` and `runs/agency-coding-uses-match-2` on the machine that ran them; the shape above is verbatim.

## Proposal

Accept this form as an alternative spelling of a handler, the way `===` is accepted for `==`, so an agent that reaches for it gets working code instead of a parse error. Pieces the guess uses:

- `handle (expr) with { ... }`: a handle block whose body is a single call, used as an expression that yields the call's result. Today `handle` is a statement and `const x = handle { ... }` fails with "Expected function body".
- `with { on <effect>(param) { ... } ... on _ { ... } }`: one clause per effect name, with the interrupt bound to `param`, and `on _` as the catch-all. A clause body that ends in `approve()` / `reject()` / `pass()` without `return` is the verdict.
- Inside a clause, the model reads the payload as `data.dir`, not `data.data.dir`.
- `raise effect::name({ ... })` with the data object as the only argument and no message.

Whether to accept each piece is a judgement call; the `on` clause form and the expression-position `handle` are the two the model produced every time.

## Why this is worth doing

The agent's other idiom misses (if-chains instead of `match`, forgetting the `raises` clause) are things a prompt or an optimize run can fix. A parse error on the first draft is different: it costs a whole round trip and the typecheck feedback ("Expected function body") does not say what the model got wrong. Meeting the default syntax removes that round trip for every model with the same prior.

[...]

PR #931, opened 2026-08-27
Coding-agent eval suite, optimizer fixes, and a sandboxed typecheck for drafts

egonSchiele:
## What this is

Three things that grew together while building an eval suite for the Agency coding agent.

### The suite

`evals/agency-coding` goes from four tests to twenty-one. Each covers one thing the writer gets wrong: docstrings written as tool descriptions (asked and unasked), settings as named parameters, functions restrictable with `.partial()`, loop forms, `static` globals, a named interrupt before an irreversible action, `idempotent` and `destructive` markers, `fork`/`race`/`parallel`, stdlib knowledge, per-call guards, Result narrowing, module basics (`export`, `.agency` in imports), program shape, `effect` payload declarations, `.rename()`/`.describe()` on derived tools, and never calling a JS array method with a callback. Tests about how code is written use one rubric judge (`lib/idiomJudge.ts`); tests about what it does carry a holdout; some carry both, so the optimizer sees a reason next to a failing case. Tags are only `easy`, `medium`, or `hard` (all `easy` today); `loop-forms`, `effect-payload-types`, `derived-tool-names`, and `docstrings-knows-how` carry weights below 1.

`evals/agency-review` gains `js-array-methods`, and the reviewer's system prompt and the suite's facts card both state the rule: a callback cannot raise an interrupt, so `.map`/`.filter`/… typecheck and then crash.

### Optimizer

- Harness pairs are ordinary scored graders. A test opts into all-or-nothing with `"harnessMustPass": true`; the flag is stored on the run's harness record. Before, every holdout was a gate, so partial progress never reached the objective and a judge beside a holdout never ran on a failing solution.
- A candidate that crashes the agent on an input scores 0 there and the search continues.
- The decoded target value keeps `\${` escaped. The first optimize run crashed on `name is not defined` because the prompt's `"Hello, \${name}"` came back as a live interpolation.
- Triple-quoted strings honour a second escape, `\"""`, so a prompt that explains docstrings can be written back as a block. Parser, generator, and optimizer agree on it; `docs/dev/language/triple-quoted-string-escapes.md` records the contract. `docs/site/guide/basic-syntax.md` still says `\${` is the only escape and needs a line.

### The writer's own typecheck

`std::agency typecheck` takes a `dir`: the draft is checked as if it were a file there, so its relative imports resolve against `dir`'s files the way `typecheckFile`'s do, and nothing is written to `dir`. `agencyReviewAgent` and `agencyCodingAgent` take `dir` too, and both eval entry nodes pass the working directory. Before, a draft that imported a seeded sibling file failed review with "Symbol X is not defined in ./sibling.agency", and the writer gave up on code that then passed the holdouts.

`coding.agency`'s system prompt is now an `optimize` target and holds the champion from the first GEPA run over the suite: train 0.42 → 0.87, validation 0.58 → 0.72 (`runs/optimize/coding-gepa-2`, $2.57). Two things to know about that prompt: it marks nearly every function `idempotent`, since nothing in the suite penalises a gratuitous marker, and it teaches `map orders as o { ... }` without parentheses, which `module-basics` then failed on. Both are worth a hand edit before the next run.

Issues filed on the way: #926, #928 (`match`/`if-then-else` value inside an `as` block is null), #929 (`guard` over a loop returns `success([])` on trip).

## Verification

- `pnpm run typecheck`, `pnpm run lint:structure`, prettier: clean.
- Vitest over `lib/eval`, `lib/optimize`, `lib/parsers`, `lib/backends`, `lib/compiler`, `lib/runDirectory`: green.
- Every holdout verified with its reference (passes) and a wrong solution (fails); every judged test run once against the writer, scores in the README entries and the commit history of the squashed branch.
- After the typecheck fix, the three tests that had "given up" ran clean.


packages/agency-lang/lib/compiler/compileValidatedClosure.ts
```diff
@@ -12,37 +12,62 @@
...
-import { compileSource, CompileResult } from "./compile.js";
+import { compileSource, CompileResult, typeCheckSource, TypeCheckReport } from "./compile.js";
 import type { ImportKind } from "../importPaths.js";
 import { openValidatedClosure, ValidatedClosure } from "./closureValidator.js";
 import { safeDeleteDirectoryWithin } from "../utils.js";
 
 const PRIVATE_DIRECTORY_MODE = 0o700;
 
-export function compileValidatedClosure(closure: ValidatedClosure): CompileResult {
+type MirroredModule = { relPath: string; source: string; mirrorPath: string };
```
egonSchiele: Why is this needed?

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -37,7 +41,7 @@ type GeneratedCode = {
 // (lib/agents/agency-agent/brains/coordinator/subagents/code.agency) — the authoritative
 // source is docs/site/guide/basic-syntax.md, which the model can read
 // through the docs tool.
-static const systemPrompt = """You write code in the Agency language.
+optimize static const systemPrompt = """You write code in the Agency language.
```
egonSchiele: remove `optimize`, we are done

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -53,6 +57,150 @@ Agency syntax rules you MUST follow:
 - Declare variables with let or const before use.
 - No Python-style colon blocks. Always { ... }.
 
+Agency-specific authoring and QA requirements (follow these exactly):
+- Exported API: Any function the caller expects to import or any tool you intend
+  for an LLM to call must be declared with the export keyword (export def ...).
+  Tests and the harness will import/export by name; missing export will fail.
```
egonSchiele: >   Tests and the harness will import/export by name; missing export will fail.

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -53,6 +57,150 @@ Agency syntax rules you MUST follow:
...
+
+- Local import filename rule: When importing other source files from the same
+  project, always import the .agency file explicitly (for example:
+  import { greet } from "./greeting.agency"). The sandbox often rejects bare
```
egonSchiele: "The sandbox often rejects" -> "We always reject"

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -53,6 +57,150 @@ Agency syntax rules you MUST follow:
...
+        const r: string = payments::charge(amount)
+        return success(r)
+      }
+    }
+
+- Handler style and matching rules: When you implement interrupt handlers,
+  always use a single match on the handler's effect/value rather than a
+  chain of if/else tests. In other words, inside a `with (intr) { ... }` or
+  `with (data) { ... }` handler preference, write:
+    match (intr.effect) {
```
egonSchiele: This should be a `return match`, otherwise the `approve()` / `reject()` calls aren't actually returned.

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -53,6 +57,150 @@ Agency syntax rules you MUST follow:
...
+  the test specifies otherwise. When unwrapping Result values in test-like
+  tasks, prefer `match(result) { success(v) => ... failure(e) => ... }` to
+  ensure the harness can detect the intended pattern.
+
+- Creating tool variants: When you return or export functions intended as LLM
+  tools, prefer creating restricted variants with .partial(...) and set a
+  stable tool name with .rename("tool-name"). Do NOT hand the LLM a plain
+  JS-like object with name/description/fn fields; use the function's
+  .partial and .rename methods (and optionally .describe()) so the runtime
+  can enforce restrictions. Example: let t = exportFn.partial({ dryRun: true }).rename("delete-files-dry").
```
egonSchiele: this example should use named args instead of a config obj (`{ dryRun: true }`)

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -53,6 +57,150 @@ Agency syntax rules you MUST follow:
...
+
+- Use standard library helpers (prefer std:: helpers over handwritten loops):
+  - Use std::path.extname (extname) for file extension extraction instead
```
egonSchiele: `std::path.extname` is not a very common standard library function. The prompt is over-indexing on it because one of our eval tests uses it, but it should be removed from this prompt.

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -53,6 +57,150 @@ Agency syntax rules you MUST follow:
...
+    of implementing extOf manually.
+  - Prefer std::array helpers (map, filter, reduce, count/length), std::array.unique,
```
egonSchiele: These examples are wrong. First of all, these helper functions are already in the namespace for every agency program. They do not need to be imported explicitly. Also, this guidance makes it look like users can call functions with the module name prepended (eg `std::array.unique(arr)`, but they cannot.

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -53,6 +57,150 @@ Agency syntax rules you MUST follow:
...
+    std::array.groupBy (or std::helper like unique/groupBy) and std::range
+    (range(1, n+1)) for ranges instead of manual while-counting.
+  - When transforming or producing a list from another list, prefer map/comprehensions
+    (for example: map orders as o { ... } or map items as v { ... }) or
+    std::array.map/filter rather than creating an empty array and pushing in a loop
+    — tests look for idiomatic usage in many tasks.
```
egonSchiele: Different parts of this prompt refer to "tests" or "sandboxes", which makes sense because this prompt was written in an eval context. However, is it useful to continue having that language in the prompt?

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -53,6 +57,150 @@ Agency syntax rules you MUST follow:
...
+  - When grouping by a key (e.g., by extension), prefer groupBy or a stdlib helper
+    over manually building a Record and pushing into arrays.
+  - Common modules: std::shell (bash, exec, ls), std::http (fetch, fetchJSON),
+    std::fs (remove, readFile), std::path (extname), std::array (map, filter),
+    std::agent (ask/interrupt helpers), std::agency (runCode). Prefer these helpers
+    for concise, idiomatic solutions; many tests assert their presence.
+
+- Loop forms: Only use for (item in list) for array iteration. Use while for
```
egonSchiele: no mention of list comprehensions?

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -53,6 +57,150 @@ Agency syntax rules you MUST follow:
...
+  counting or when you need a numeric index. Prefer declarative array helpers
+  (map/filter/reduce/unique/groupBy) when the task expects idiomatic solutions.
+
+- Static module-level constants: If an assignment expects a compile-time
```
egonSchiele: This language feels like it is really overfitting to one of our eval tests, but could easily be made a bit more general.

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -53,6 +57,150 @@ Agency syntax rules you MUST follow:
...
+  constant table (rates, config, etc.) declare it as `static const` at the
+  module level. Mutable module-level state should be `let` (for example,
+  history arrays) and should not be exported unless explicitly required.
+
+- Docstring content and brevity: Docstrings should be concise: one or two
```
egonSchiele: Weird ordering, why not keep all the guidance related to docstrings and PFA's together?

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -53,6 +57,150 @@ Agency syntax rules you MUST follow:
...
+
+- Tool naming and descriptions: When presenting a tool to an LLM, use .rename
+  to give a stable machine-readable name and .describe() or the docstring to
+  provide a human-friendly description. Objects with name fields or ad-hoc
+  descriptors are not accepted by the test harness.
+
+- Result handling and matches: When a problem asks you to propagate or return
+  Result failures from an imported parser or helper, return the original
+  failure unchanged (e.g., `return r`) and prefer using `match` or `is
+  success/is failure` in ways the tests expect; when in doubt use `match` to
```
egonSchiele: No clearer signal that a prompt is overfitting to the test, than the phrase "in ways the tests expect" in the prompt itself.

packages/agency-lang/stdlib/agents/agency/coding.agency
```diff
@@ -359,9 +501,12 @@ node evalMain(input: CodingEvalInput): string {
...
   // the assignment text says what shape the deliverable takes, and the
   // harness judges the outcome. So no mechanical main-node check here.
-  let written: Result<string, WriteFailure> = failure({ source: "", problems: ["the writer never ran"] })
+  let written: Result<string, WriteFailure> = failure({
+    source: "",
+    problems: ["the writer never ran"]
+  })
   handle {
-    written = agencyCodingAgent(task: input.assignment, requireMain: false)
+    written = agencyCodingAgent(task: input.assignment, requireMain: false, dir: ".")
```
egonSchiele, line 467: Should all of these take a `dir` parameter, or should they use the agent cwd, like other functions do? See stdlib/index.agency

packages/agency-lang/lib/compiler/compileValidatedClosure.ts
egonSchiele: `typecheck(source, dir)` needs the same thing `compile(source, dir)` already does: validate the closure, write every validated file to a private mirror, and resolve the draft's `./sibling.agency` imports against that mirror rather than the caller's directory. The mirror loop lived inline in `compileValidatedClosure`; `withMirror` is that loop lifted out so `typeCheckValidatedClosure` can use it too instead of duplicating the file-writing and cleanup. The two types name the shape it hands back. If you would rather the typecheck path stay separate, I can inline it, but then the TOCTOU boundary is enforced in two places.

packages/agency-lang/stdlib/agents/agency/coding.agency
egonSchiele: I kept `dir` because it mirrors `compile(source, dir)`, which already exists on std::agency, and because the draft here is a string that has not been saved anywhere: `dir` says where it will live, so its relative imports resolve. The agent-cwd convention in stdlib/index.agency is for functions that take a path and resolve it (`useAgentCwd: true` sends a relative path through `applyAgentCwd`). For the agents I could add `useAgentCwd: boolean = false` next to `dir`, resolving `dir` through `applyAgentCwd` when set, which is what `read.partial(useAgentCwd: true)` does in toolkits. Say which you prefer; I left it as `dir` alone for now.

packages/agency-lang/lib/compiler/compileValidatedClosure.ts
egonSchiele: Not worth it, on reflection. Two facts settled it: `typecheckFile(dir, filename)` already typechecks straight from disk with only a containment check, no mirror; and `SymbolTable.build` already accepts in-memory source overrides (the LSP uses them for unsaved editor buffers). So `typecheck(source, dir)` is now four lines in `_typecheck`: give the draft a synthetic path inside `dir`, pass its text as the override, and let its relative imports resolve from `dir`'s files the way `typecheckFile`'s do. Nothing is written to `dir`. `withMirror`, `typeCheckValidatedClosure`, and `typecheckSandboxed` are gone; the three compiler files are back to what `main` has, and the only compiler change left is an `overrides` parameter threaded through `typeCheckSource`, with a test in `lib/compiler/typecheckDraftInDir.test.ts`.

PR #932, opened 2026-08-27
agency-review: sixteen tests harvested from the coding suite

egonSchiele:
## What this is

`evals/agency-review` grows from 6 tests to 22, harvested from the coding suite: each new test reuses a coding assignment, and the reviewed source is either a wrong solution the writer actually produced this week or the reference solution. Three kinds:

- **Bug** (8), the reviewer must reject: no `export` on a function other files call; a send that cannot be undone with no interrupt first; settings as an options object where the task requires named parameters and `.partial()`; `catch 0` swallowing a failure the task says to return; a per-run list declared `static const`; `idempotent` on a function that appends state; one guard around a whole loop when each call must fail alone; `race` where every result is needed.
- **Idiom** (4), the code does what the task asks in a JavaScript way, and the reviewer must say so as advice without rejecting: hand-rolled loops where `groupBy`/`unique`/`count`/`range`/`extname` exist; tool descriptions in `//` comments instead of docstrings; wrapper functions instead of `.partial().rename()`; a never-changing table as a plain `const`. New builder `idiomGraders({ reason })`: `no-false-positive` plus a `names-the-idiom` judge over the advisory findings.
- **Clean** (4), the coding references, which the reviewer must not reject.

The facts card gains the blocks rule (inline `\x -> …`, full `f(xs) as x { … }`, `return` inside a block returns the block's value), after the reviewer rejected two correct references over exactly those beliefs and `agency-true` could not tell.

## First run of the current reviewer

Mean 0.73 over the 16 new tests ($0.18 to run, about $0.25 to grade). What it shows, which is the signal the optimizer needs:

- Catches 5 of 8 bugs. Misses the three that are Agency policy rather than logic: missing `export`, no interrupt before an irreversible send, `idempotent` on a mutator.
- Advises on 0 of 4 idioms; it reviews correctness only.
- Rejects 2 of 4 clean references, for the two false beliefs above.

## Verification

Every planted source typechecks beside its siblings (`agency typecheck` in each `files/` dir; `ungated-delete` carries a pre-existing AG3009 warning). Prettier clean. Tags are `bug`, `idiom`, or `clean`.

### DIFF
--- a/packages/agency-lang/evals/agency-review/comments-not-docstrings/test.json
+++ b/packages/agency-lang/evals/agency-review/comments-not-docstrings/test.json
@@ -1,0 +1,11 @@
+{
+  "description": "Three tool functions described in // comments above the def, which the LLM never sees, instead of docstrings inside the body. Correct, so advice, not rejection.",
+  "tags": [
+    "idiom"
+  ],
+  "files": "./files",
+  "input": {
+    "assignment": "Write a module for a to-do list kept in memory. It exports three functions that an llm call will be given as tools. `addTask(title: string, due: string = \"\"): string` adds a task and returns a short confirmation. `completeTask(title: string): string` marks the task with that exact title done and returns a short confirmation, or a message saying no such task exists. `listOpenTasks(): string[]` returns the titles of the tasks not yet done, in the order they were added.",
+    "sourceFile": "solution.agency"
+  }
+}
--- a/packages/agency-lang/evals/agency-review/comments-not-docstrings/test.json
+++ b/packages/agency-lang/evals/agency-review/comments-not-docstrings/test.json
@@ -1,7 +1,7 @@
 {
-  "description": "Three tool functions described in // comments above the def, which the LLM never sees, instead of docstrings inside the body. Correct, so advice, not rejection.",
+  "description": "Three tool functions described in // comments above the def, which the LLM never sees, instead of docstrings inside the body.",
   "tags": [
-    "idiom"
+    "easy"
   ],
   "files": "./files",
   "input": {
--- a/packages/agency-lang/evals/agency-review/guard-around-loop/test.json
+++ b/packages/agency-lang/evals/agency-review/guard-around-loop/test.json
@@ -1,0 +1,11 @@
+{
+  "description": "One guard wraps the whole loop, so a single slow document times out every document, when the task says each must fail alone.",
+  "tags": [
+    "bug"
+  ],
+  "files": "./files",
+  "input": {
+    "assignment": "The file summarizer.agency in the working directory exports `summarize(doc: string): string`, which is slow, and slower for long documents. Write a module that imports it and exports `summarizeAll(docs: string[]): string[]`. It returns one summary per document, in order. No single summarize call may run for more than 500 milliseconds; a document that takes longer gets the string \"timed out\" in its place, and the other documents are still summarized.",
+    "sourceFile": "solution.agency"
+  }
+}
--- a/packages/agency-lang/evals/agency-review/guard-around-loop/test.json
+++ b/packages/agency-lang/evals/agency-review/guard-around-loop/test.json
@@ -1,7 +1,7 @@
 {
   "description": "One guard wraps the whole loop, so a single slow document times out every document, when the task says each must fail alone.",
   "tags": [
-    "bug"
+    "easy"
   ],
   "files": "./files",
   "input": {
--- a/packages/agency-lang/evals/agency-review/hand-rolled-loops/test.json
+++ b/packages/agency-lang/evals/agency-review/hand-rolled-loops/test.json
@@ -1,0 +1,11 @@
+{
+  "description": "Four helpers written with for and while loops and a hand-made extension splitter, where the stdlib already has groupBy, unique, count, range, and extname. Correct, so advice, not rejection.",
+  "tags": [
+    "idiom"
+  ],
+  "files": "./files",
+  "input": {
+    "assignment": "Write a module that exports a type `Entry = { path: string, size: number }` and four functions. `byExtension(entries: Entry[]): any` groups the entries by the extension of their path, with the extension, dot included, as the key. `extensions(entries: Entry[]): string[]` returns the distinct extensions in first-seen order. `largeCount(entries: Entry[], limit: number): number` returns how many entries have a size above limit. `pageNumbers(pages: number): number[]` returns the numbers 1 through pages.",
+    "sourceFile": "solution.agency"
+  }
+}
--- a/packages/agency-lang/evals/agency-review/hand-rolled-loops/test.json
+++ b/packages/agency-lang/evals/agency-review/hand-rolled-loops/test.json
@@ -1,7 +1,7 @@
 {
-  "description": "Four helpers written with for and while loops and a hand-made extension splitter, where the stdlib already has groupBy, unique, count, range, and extname. Correct, so advice, not rejection.",
+  "description": "Four helpers written with for and while loops and a hand-made extension splitter, where the stdlib already has groupBy, unique, count, range, and extname.",
   "tags": [
-    "idiom"
+    "easy"
   ],
   "files": "./files",
   "input": {
--- a/packages/agency-lang/evals/agency-review/no-interrupt-before-send/graders.ts
+++ b/packages/agency-lang/evals/agency-review/no-interrupt-before-send/graders.ts
@@ -1,0 +1,6 @@
+import { plantedBugGraders } from "../lib/reviewGraders.js";
+
+export default plantedBugGraders({
+  reason:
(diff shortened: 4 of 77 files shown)


packages/agency-lang/evals/agency-review/comments-not-docstrings/test.json
```diff
@@ -0,0 +1,11 @@
+{
+  "description": "Three tool functions described in // comments above the def, which the LLM never sees, instead of docstrings inside the body. Correct, so advice, not rejection.",
```
egonSchiele: > Correct, so advice, not rejection.

Remove this

packages/agency-lang/evals/agency-review/comments-not-docstrings/test.json
```diff
@@ -0,0 +1,11 @@
...
+  "tags": [
+    "idiom"
```
egonSchiele: Only valid tags are easy, medium, hard.

packages/agency-lang/evals/agency-review/guard-around-loop/test.json
```diff
@@ -0,0 +1,11 @@
+{
+  "description": "One guard wraps the whole loop, so a single slow document times out every document, when the task says each must fail alone.",
```
egonSchiele, line 2: good test case

packages/agency-lang/evals/agency-review/hand-rolled-loops/test.json
```diff
@@ -0,0 +1,11 @@
+{
+  "description": "Four helpers written with for and while loops and a hand-made extension splitter, where the stdlib already has groupBy, unique, count, range, and extname. Correct, so advice, not rejection.",
+  "tags": [
+    "idiom"
+  ],
+  "files": "./files",
+  "input": {
+    "assignment": "Write a module that exports a type `Entry = { path: string, size: number }` and four functions. `byExtension(entries: Entry[]): any` groups the entries by the extension of their path, with the extension, dot included, as the key. `extensions(entries: Entry[]): string[]` returns the distinct extensions in first-seen order. `largeCount(entries: Entry[], limit: number): number` returns how many entries have a size above limit. `pageNumbers(pages: number): number[]` returns the numbers 1 through pages.",
```
egonSchiele, line 8: Is it okay to have different types of loops in the same eval test, or should we have one eval test for each type of loop?

packages/agency-lang/evals/agency-review/no-interrupt-before-send/graders.ts
```diff
@@ -0,0 +1,6 @@
+import { plantedBugGraders } from "../lib/reviewGraders.js";
+
+export default plantedBugGraders({
+  reason:
+    "postComment calls sendToServer, which cannot be undone, without raising a named interrupt first (raise comments::post(...) with raises <comments::post>), so nothing can approve or reject the send.",
```
egonSchiele: I'm not sure this text makes it clear enough that what we need is to raise an interrupt before dangerous or destructive actions, or mutations, and the postComment action is a mutation, thus needs an interrupt.

packages/agency-lang/evals/agency-review/comments-not-docstrings/test.json
egonSchiele: Removed from all four idiom descriptions.

packages/agency-lang/evals/agency-review/comments-not-docstrings/test.json
egonSchiele: Retagged every test in the suite `easy` (the README and docs/dev/evals/eval-grading.md now say only easy, medium, hard).

packages/agency-lang/evals/agency-review/hand-rolled-loops/test.json
egonSchiele: Either works with the current graders, but they answer different questions.

One test with four helpers (as now) asks "does the reviewer catch every hand-rolled loop in a realistic file?" The `names-the-idiom` judge sees the four together, so a reviewer that mentions two of them gets partial credit and we cannot tell from the score which two it missed. It is one run per trial, so it is cheap.

One test per loop asks "does the reviewer know groupBy / unique / count / range individually?" Four times the runs, but each score points at one idiom, which matters when we optimize the prompt: a rule about `range` and a rule about `groupBy` are different rules.

I would keep this combined test as the realistic case and add one small single-helper test per idiom later if the optimizer needs the sharper signal. Happy to split it now instead if you would rather.

packages/agency-lang/evals/agency-review/no-interrupt-before-send/graders.ts
egonSchiele: Rewrote the reason to state the rule first, then the breach:

> Agency code must raise an interrupt before any action that changes the outside world and cannot be undone, so the caller can approve or reject it. postComment sends a comment, which the assignment says cannot be unsent, and never raises an interrupt: it calls sendToServer directly, with no `raise comments::post(...)` before the call and no `raises <comments::post>` on the function. The caller gets no chance to stop the send.

packages/agency-lang/stdlib/agents/agency/review.agency
```diff
@@ -17,47 +17,96 @@ import { systemMessage, threadIsNew } from "std::thread"
 
 static const systemPrompt = """You review code written in the Agency language. Agency compiles to
 TypeScript and looks like JS/TS, but it is its own language with its own
-idioms — judge the code as Agency, never as JavaScript.
+idioms. Judge the code as Agency, never as JavaScript.
 
-Ground rules about Agency (get these wrong and the review is wrong):
-- == and === are the same operator: ===/!== are accepted aliases that
-  compile identically to ==/!=, so ==-vs-=== advice is meaningless here.
+Facts about Agency. Get these wrong and the review is wrong.
```
egonSchiele: > Get these wrong and the review is wrong

Gross phrasing

packages/agency-lang/stdlib/agents/agency/review.agency
```diff
@@ -17,47 +17,96 @@ import { systemMessage, threadIsNew } from "std::thread"
...
-  forms are list comprehensions (`[x for x in xs if ...]`) and the stdlib
-  functions `map`, `filter`, `sortBy`, `reduce` with a block (`as x { }`).
-  Calling one of those methods is an error finding.
-- `let`/`const` declarations, postfix types, `\${...}` interpolation,
-  `node main()` entry points, braces and parenthesized conditions.
+- Fallible functions return Result. `if (r is success(v)) { ... }` and
+  `is failure(e)` are correct narrowing syntax that binds v/e; `try` and
+  `catch <default>` are the expression forms. A function that receives a
+  failure and cannot handle it returns that failure as it is; rebuilding
+  it with `failure(e)` loses nothing but is not an error.
```
egonSchiele: > loses nothing but is not an error

why "but"? I am confused, are you giving guidance that the LLM should do this? or shouldn't do this? or it doesn't matter? I can tell you that ideally they shouldn't do it because it makes the code slightly less readable, and changes the source of the error message.

packages/agency-lang/stdlib/agents/agency/review.agency
```diff
@@ -17,47 +17,96 @@ import { systemMessage, threadIsNew } from "std::thread"
...
+- Blocks are how a function receives code to run per item. The inline
+  form is a backslash, parameters, and an arrow: `map(xs, \x -> x * 2)`,
+  `reduce(xs, 0, \(acc, n) -> acc + n)`. The full form follows the call:
+  `map(xs) as x { ... }`, `fork(xs) as x { ... }`. A `return` inside a
+  block returns the block's value for that item, not from the enclosing
+  function. Both forms are valid Agency; neither is a JavaScript lambda.
```
egonSchiele: > Both forms are valid Agency; neither is a JavaScript lambda

🤢

packages/agency-lang/stdlib/agents/agency/review.agency
egonSchiele: Cut. The heading is now just "Facts about Agency:".

packages/agency-lang/stdlib/agents/agency/review.agency
egonSchiele: Moved it out of the facts and into the advice list, saying what you said: "A function that receives a failure it cannot handle and rebuilds it with `failure(e)`. Return the original failure unchanged. Rebuilding it changes the source of the error message and adds a line."

packages/agency-lang/stdlib/agents/agency/review.agency
egonSchiele: Replaced with an instruction: "Do not flag either form as a JavaScript lambda." The block examples now sit in their own lines before the description. I also ran the prompt through the writing-review agent and applied its findings, and added docs/dev/contributing/verbal-tics.md with the list to check against.

Issue #591, opened 2026-07-18
Ternary inside a block body evaluates to null


egonSchiele:
A ternary (`if c then a else b`) written inside a block body evaluates to `null`. An `if`/`else` statement in the same position works, and a block closing over an enclosing local works, so the problem is specific to the ternary.

Found while moving the stdlib agents onto block-taking helpers in #589. That PR works around it, but the workaround has to be remembered at every call site, and the failure is silent.

## Reproduction

Pinned in `tests/agency/agents/blockProbe.agency`.

```agency
def runBlock(block: (string) -> any): any {
  return block("arg")
}

node ternaryOnBlockArg(): string {
  return runBlock() as arg {
    const picked = if arg == "" then "empty" else "nonempty"
    return picked
  }
}
```

Expected: `"nonempty"`
Actual: `null`

It does not depend on touching an enclosing local. Both of these fail the same way:

| Node | Shape | Result |
|---|---|---|
| `ternaryOnBlockArg` | ternary over the block argument only | `null` |
| `ternaryOverLocal` | ternary over an enclosing local | `null` |
| `ifElseInBlockWorks` | `if`/`else` statement, same logic | correct |
| `closureOverLocal` | plain closure over an enclosing local, no ternary | correct |

## Likely cause

The generated code reads the ternary's value from the parent frame while the block is executing in its own. From `stdlib/agents/researcher.js` when the researcher still used a ternary inside its block:

```js
await runner3.step(2, async (runner4) => {
  __bstack.locals.msg = __nn(__stack.locals.__matchval_1);
});
```

The ternary lowers to a match whose value is stored as `__matchval_1`. The write goes to `__stack` (the enclosing function frame), the read happens while the block frame `__bstack` is active, and `__nn` normalizes the missing value to `null`. So the assignment silently produces `null` rather than erroring.

## Why it matters

It fails silently. There is no error and no diagnostic, just a `null` flowing onward, which then surfaces somewhere unrelated. In #589 it surfaced as `TypeError: content is not iterable` from inside smoltalk, because the `null` became the prompt of an `llm()` call several frames away.

Ternaries are idiomatic in this codebase and blocks are the main way stdlib helpers take work (`sample`, `consensus`, `retry`, `revise`, `withLock`, `concurrencyPool`, `supervise`), so the two are easy to combine by accident.

## Current state

[...]

PR #934, opened 2026-08-27
Read match-expression results from the block frame inside blocks

### LINKED ISSUE #590: match inside a block that pauses and resumes evaluates to null
A `match` expression that sits inside a block which pauses and resumes evaluates to `null` instead of the matched arm. Reading the same Result with `isFailure(r)` / `r.value` works, and a `match` on the guard's own Result outside the block works, so the problem is specific to a `match` being replayed as part of a resumed block.

Found while building `supervise` (`std::supervise`) in #589. That PR works around it, so nothing is blocked, but the workaround is load-bearing and undocumented outside the test.

## Reproduction

Pinned in `tests/agency/supervise/nestedGuardResume.agency`. An outer time guard trips while an inner guard is active; the handler approves with more time; the block is replayed from the step boundary.

```agency
node matchInsideResumedBlock(): string {
  handle {
    const outer = guard(time: 5ms, label: "a") {
      const inner = guard(time: 600000, label: "ai") {
        const s = spin(300000)
        return "inner:${s}"
      }
      return match(inner) {
        success(v) => "ok:${v}"
        failure(e) => "inner-failed"
      }
    }
    if (isFailure(outer)) { return "outer-failed" }
    return outer.value
  } with (i) {
    if (i.effect == "std::guard" && i.data.label == "a") { return approve({ maxTime: 600000 }) }
    return pass()
  }
}
```

Expected: `"ok:inner:spun"`
Actual: `null`

The same node with the `match` replaced by `if (isFailure(inner)) { ... } return "ok:${inner.value}"` returns `"ok:inner:spun"` (pinned as `isFailureInsideResumedBlock`), which also shows the inner guard's Result itself survives the pause and resume intact. A `match` on the guard's own Result, outside the resumed block, is likewise fine (`matchOutsideResumedBlock`).

So the three cases are:

| Shape | Result |
|---|---|
| `match` inside the resumed block | `null` |
| `isFailure` / `.value` inside the resumed block | correct |
| `match` outside the block, on the guard's own Result | correct |

## Why it matters

Any code that reads a Result with `match` can silently return `null` after a guard trip is approved. It fails quietly: no error, no diagnostic, just a null flowing onward. `match` is the idiomatic way to read a Result and the one the docs teach, so this is easy to hit and hard to attribute.

It is also the shape the stdlib agents use. Every agent in #589 runs its work inside a `guard`, and the composition those agents document puts them inside a `supervise` block, which is precisely a block that pauses and resumes. So this affects the primary intended composition.

## Current state

`tests/agency/supervise/nestedGuardResume.agency` characterizes all three cases. `matchInsideResumedBlock` currently asserts the buggy `null` so that a fix flips the test rather than leaving a silent gap. Flipping that expectation to `"ok:inner:spun"` is the acceptance check for this issue.

Every agent in `std::agents` reads its guard Result with `isFailure` rather than `match`, each with a comment pointing at that test. Once this is fixed, those can go back to `match` and regain its exhaustiveness checking.

Also noted in `TODO.md`.

## Guesses at the cause

Not investigated. Worth checking whether the stepped-loop codegen assigns a `match` its own step, and whether the replay after `approve` re-enters that step with the binder scope or the scrutinee unset. The neighboring known issue about binder-`match` in stepped loops (the reason `plannerAgent` threads state through plain helpers instead of binder-`match`) may be the same underlying mechanism.

### LINKED ISSUE #928: match and if-then-else expressions inside an as-block evaluate to null

egonSchiele:
Fixes #928. Also fixes #590 and #591, which were the same bug filed earlier.

## The bug

A `match` or `if ... then ... else` expression assigned inside a trailing block (`map(xs) as p { ... }`) evaluated to `null`.

## Why

The compiler lowers a value-producing `match` into an if-chain whose arms call `runner.exitMatch(id, value)`. That writes the value into `runner.frame.locals.__matchval_<id>`. Inside a block, the runner's frame is `__bstack`, so the write landed on `__bstack.locals`. But the generated read was hardcoded to `__stack.locals.__matchval_<id>`, the enclosing node's frame, which never had the value:

```js
__bstack.locals.m = __nn(__stack.locals.__matchval_1);   // before
__bstack.locals.m = __nn(__bstack.locals.__matchval_1);  // after
```

## The fix

`matchValFrameScope()` in `typescriptBuilder.ts` picks `"block"` when the current scope is a block, `"local"` otherwise. Both the read and the plain-mode (handler body) write use it.

Review found a second problem in the same shape: a block inside a handler body inherited plain-mode codegen, so its `return` was a bare `return` inside a step callback and the block yielded `undefined`. Blocks get their own `Runner`, so `stepped()` now resets the two plain-mode flags at every block boundary.

## Tests

- `tests/agency/match-expr-in-block.agency`: `match` in a block, `if-then-else` in a block, `match` in a nested block, `match` in a block inside a handler. 0/4 on main, 4/4 here.
- `agents/blockProbe` and `supervise/nestedGuardResume` pinned the old `null` for #590/#591; they now expect the real value. The matching TODO.md entries are gone.
- Builder unit tests and TypeScript generator fixtures (561) pass; `pnpm run typecheck` is clean.

https://claude.ai/code/session_01G8WgPvTTtUyQwJWrXGxAgh

### DIFF
--- a/packages/agency-lang/lib/backends/typescriptBuilder.ts
+++ b/packages/agency-lang/lib/backends/typescriptBuilder.ts
@@ -267,13 +267,11 @@ export class TypeScriptBuilder {
   // expression-match regions, every one of which routes through
   // `processMatchExpressionPlain` in a handler, and (b) the builder never
   // switches OUT of plain mode within the IIFE — it does not reset
-  // `insideHandlerBody` at block-argument / nested-frame boundaries, so nothing
-  // stepped compiles inside a handler. At every yield site this flag therefore
-  // currently equals `insideHandlerBody`. If a future change starts compiling
-  // stepped constructs inside a handler (e.g. resetting `insideHandlerBody` per
-  // frame), reset THIS flag at those same boundaries too — otherwise a stepped
-  // match nested in a plain arm would emit a bare `return` inside a step
-  // callback (value discarded, `_matchExit` never set: a silent wrong answer).
+  // `insideHandlerBody` anywhere except at block boundaries, where `stepped()`
+  // resets THIS flag too. Both flags must always be reset together — otherwise
+  // a stepped match inside a block in a plain arm would emit a bare `return`
+  // inside a step callback (value discarded, `_matchExit` never set: a silent
+  // wrong answer).
   private insidePlainMatchExpr: boolean = false;
 
   /*
@@ -1094,11 +1092,9 @@ export class TypeScriptBuilder {
             // to Agency's single nothing-value, `null`. This one read site is
             // the chokepoint for every match-result path (stepped no-arm and
             // the plain-mode IIFE that returns `undefined`). See #409.
-            // `exitMatch` writes to the frame of the runner that ran the
-            // match: `__stack` in a node or function, `__bstack` inside a
-            // block. Read from the same one (#928).
-            const frameScope = this.scopes.current().type === "block" ? "block" : "local";
-            return ts.call(ts.id("__nn"), [ts.scopedVar(literal.value, frameScope, this.moduleId)]);
+            return ts.call(ts.id("__nn"), [
+              ts.scopedVar(literal.value, this.matchValFrameScope(), this.moduleId),
+            ]);
           }
           return ts.id(literal.value);
         }
@@ -1668,9 +1664,19 @@ export class TypeScriptBuilder {
       this.insidePlainMatchExpr = prev;
     }
     return ts.assign(
-      ts.scopedVar(matchValName(matchId), "local", this.moduleId),
+      ts.scopedVar(matchValName(matchId), this.matchValFrameScope(), this.moduleId),
       ts.await(ts.iife({ async: true, body: [ifChain] })),
     );
+  }
+
+  /**
+   * The frame a `__matchval_<id>` temp lives in: `__bstack` inside a block,
+   * `__stack` in a node or function. The stepped writer (`runner.exitMatch`)
+   * uses the running runner's frame, which is the same choice. Every read
+   * and the plain-mode write go through here so they cannot drift (#928).
+   */
+  private matchValFrameScope(): "block" | "local" {
+    return this.scopes.current().type === "block" ? "block" : "local";
   }
 
   private processImportStatement(node: ImportStatement): TsNode {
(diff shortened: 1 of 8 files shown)


packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -1094,7 +1094,11 @@ export class TypeScriptBuilder {
             // to Agency's single nothing-value, `null`. This one read site is
             // the chokepoint for every match-result path (stepped no-arm and
             // the plain-mode IIFE that returns `undefined`). See #409.
-            return ts.call(ts.id("__nn"), [ts.scopedVar(literal.value, "local", this.moduleId)]);
+            // `exitMatch` writes to the frame of the runner that ran the
+            // match: `__stack` in a node or function, `__bstack` inside a
+            // block. Read from the same one (#928).
+            const frameScope = this.scopes.current().type === "block" ? "block" : "local";
```
egonSchiele: The plain-mode write site still hardcodes `"local"`: `processMatchExpressionPlain` (around line 1665) emits `ts.scopedVar(matchValName(matchId), "local", ...)`. A handler body compiles in plain mode, and a block inside that handler body inherits it, so for this shape

```
} with (intr) {
  const labels = map([1, 2]) as p {
    const m = match (p) {
      1 => "one"
      _ => "other"
    }
    return m
  }
  return reject(labels.join(","))
}
```

the generated code writes one frame and reads another:

```js
__stack.locals.__matchval_1 = await (async () => { ... })();   // write
__bstack.locals.m = __nn(__bstack.locals.__matchval_1);        // read
```

and `labels` comes out as `["", ""]`. I checked: main also gets this case wrong (both sides were `__stack` there and it still returned empties), so this is not a regression, but after this PR the two sites disagree, which is a worse place to leave it.

Suggestion: put the `block`/`local` choice in one small private method (say `matchValFrameScope()`) and call it from both this read and the write in `processMatchExpressionPlain`, so they cannot drift again. Then a fourth case in `match-expr-in-block.agency` for a match inside a block inside a handler body would pin it.

egonSchiele (commented):
Reproduced the fix on the branch: `match` in a block, `return match (...)` straight out of a block, and the exact shape the `guards` coding eval hit (`match` on a `guard(...)` result inside `fork(docs) as doc { ... }`) all return the right values now. Nice small diff.

One thing to fix before merging, inline: the plain-mode write in `processMatchExpressionPlain` still targets `"local"`, so a match in a block inside a handler body now writes `__stack` and reads `__bstack`.

Optional: the `guards` eval shape (`return match` on a guard result inside `fork`) is worth one more case in the test file, since that is the real program that surfaced #928.

egonSchiele:
Done in the follow-up commit.

1. Both the read and the plain-mode write now call one `matchValFrameScope()` (`__bstack` inside a block, `__stack` otherwise).
2. Adding the `matchInBlockInHandler` case turned up why main returned empties for that shape: the match temp was only half of it. A block inside a handler body inherited plain-mode codegen, so its `return m` compiled to a bare `return` inside `runner2.step(...)` and the block yielded `undefined`. Any `return` in any block inside a handler had this problem. Blocks get their own `Runner`, so `stepped()` now resets `insideHandlerBody` and `insidePlainMatchExpr` at each block boundary (the two block-argument sites and the fork block). The invariant comment on `insidePlainMatchExpr` already asked for exactly this.

Verified: the new test file is 4/4, `match-expr-in-handler` is 5/5, the six agency tests that mix handlers and blocks are 18/18, builder unit tests + generator fixtures 561 pass, typecheck clean.

Issue #935, opened 2026-08-27
Re-enable the integration-credentials job (remote CLI test) in test.yml


egonSchiele:
The `integration-credentials` job in `test.yml` (the remote CLI test against the real statelog server) is disabled while the statelog side is being fixed. On 2026-08-27 it failed on main with `HTTP 429 Rate exceeded` on the first upload, and every later step failed with "Not linked" as a result (run 33035487111).

To re-enable: restore the job's `if:` condition in `.github/workflows/test.yml` to

```yaml
    if: >
      (github.event_name == 'push' && github.ref == 'refs/heads/main')
```

and delete the `if: false` line that references this issue.

PR #936, opened 2026-08-27
Deflake main: mock the thread messages test, fake-clock the nested-guard test, park the credentials job

egonSchiele:
## Why

Main went red three separate ways after #931 and #930 merged, none of them in code either PR touched. This PR deals with the two that are tests, and parks the third.

## What changed

- **`tests/agency/threads/messages.test.json`** sets `useTestLLMProvider`, so the post-merge real-LLM run uses its mocks too. The test checks that a thread records every exchange; it was an exact match on gpt-4o-mini's arithmetic, and the pinned answer (68) was wrong anyway: the primes add to 58. The mock and the expected output now say 58, and the model pin in `messages.agency` is gone.
- **`tests/agency/guards/guard-time-nested-outer-tighter`** runs under the fake clock (`"fakeClock": true`, `_advanceTime(150)` instead of `sleep(150ms)`), the way `scheduled-helpers.agency` already tests guard trips. It pins which guard a time trip belongs to when the outer budget is tighter than the inner: `outer:timeoutFailure`, never `inner:timeoutFailure:20`. The real-time version failed in 2 of the last 12 main runs, and one of its interleavings let the trip escape both guards. That runtime question is filed separately (linked in a comment below) since the fake-clock version no longer exercises the in-flight `sleep` cancellation path.
- **`integration-credentials`** in `test.yml` is `if: false` while the statelog side is fixed. #935 has the line to restore.

## Verified

- `messages.test.json` passes with `OPENAI_API_KEY` unset, so it really is mocked.
- The guard test passes 3/3 locally under the fake clock.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -470,8 +470,11 @@ jobs:
     # browser) moved to the weekly live-api-tests.yml cron so merges stop
     # burning external API quota; this job keeps only the statelog remote
     # test, which exercises our own infrastructure.
-    if: >
-      (github.event_name == 'push' && github.ref == 'refs/heads/main')
+    # Disabled while the statelog side is fixed; re-enable per
+    # https://github.com/egonSchiele/agency-lang/issues/935 by restoring:
+    #   if: >
+    #     (github.event_name == 'push' && github.ref == 'refs/heads/main')
+    if: false
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


egonSchiele:
Runtime follow-up for the interleaving that let the trip escape both guards: https://github.com/egonSchiele/agency-lang/issues/937

PR #946, opened 2026-08-28
Weekly agent eval run uploaded to statelog

egonSchiele:
PR 4 of the eval-tracking project (spec Part 3, `docs/superpowers/specs/2026-08-22-eval-tracking-pipeline-design.md`). Statelog's API (#54) and UI (#56) are merged and deployed.

## What

- `.github/workflows/agent-evals.yml` — Sundays 03:00 UTC, plus `workflow_dispatch` with `trials` (default 3) and `suite` (default `evals/agency-agent`). Steps: build → `make eval-image` → `remote link --url` → `eval run` with the suite's `run-in-docker.sh` wrapper (`--policy approve-all`, as the suite README prescribes) → `eval grade` → `eval upload`. Never fails on scores: `eval run` exit 2 (some tests errored) and `eval grade` exit 2 (a must-pass gate failed) are warnings and the upload still happens; any other non-zero exit fails the job. `timeout-minutes: 180`, fork-guarded, actions pinned by SHA.
- **Bounding an approve-all agent** — the container sees only the LLM key and the run directory; every secret is set on the one step that needs it, never job-wide; the LLM key is a dedicated `AGENT_EVALS_OPENAI_API_KEY` meant to carry its own spend limit; the run directory is saved as an artifact only when the upload failed (artifacts on a public repo are public).
- **Batch-wide budget** — new `eval.limits.maxBatchCostUsd`: `runSuite` sums each finished run's spend and stops starting tests once it is crossed (in-flight runs finish under `maxCostUsd`), so the worst case is batch cap + `--parallel` × per-run cap; judge costs are outside both. `agency.json` sets `maxBatchCostUsd: 25` and `maxCostUsd: 5`. `eval grade` tolerates the uneven trial grid a cut-short batch leaves (reports "has no statistics" instead of throwing); statelog already shows such a batch as incomplete.
- `eval run` exits 2 when some tests errored, 1 when it could not run; prints one cost total (`result.costUsd`).
- `lib/cli/eval/workflowFlags.test.ts` — checks every flag of each `agency.js …` invocation in the workflow against the command's registered options.
- Dev note section in `docs/dev/evals/eval-tracking.md`; `docs/site/cli/eval.md` documents the new cap and exit codes.

## Configuration

Three secrets in the `ci-credentials` environment: `AGENT_EVALS_OPENAI_API_KEY` (a dedicated OpenAI key with a hard monthly spend limit — new, must be created), `STATELOG_API_KEY`, and `STATELOG_PROJECT_URL` (a serve URL, `…/serve/:user/:project/:file`; kept as a secret by the owner's choice). Suggest a first `workflow_dispatch` with `suite=evals/smoke`, `trials=1` to check the wiring cheaply.

Not verified: an actual Actions run.

https://claude.ai/code/session_01YSdUvTQqBfWKBERTyspUHR

### DIFF
--- a/.github/workflows/agent-evals.yml
+++ b/.github/workflows/agent-evals.yml
@@ -1,14 +1,22 @@
 # Weekly eval run of the agency agent, uploaded to statelog so its accuracy
 # can be tracked over time (see packages/agency-lang/docs/dev/evals/eval-tracking.md).
 #
-# This spends real money: about $21 at three tests x three trials, plus the
-# grading judges. packages/agency-lang/agency.json caps it: maxBatchCostUsd
-# for the batch, maxCostUsd per run (worst case: batch cap + parallel x
-# per-run cap). It never fails on scores — a bad week is information for the
-# batch table, not an alarm. It fails only when a step could not run or the
-# upload was refused. The run directory is always saved as a workflow artifact
-# so a batch whose upload failed can be uploaded by hand: download it, then
-# `agency eval upload <dir>` from a linked checkout.
+# This spends real money. packages/agency-lang/agency.json caps it:
+# maxBatchCostUsd for the batch, maxCostUsd per run (worst case: batch cap +
+# parallel x per-run cap), plus the grading judges outside both. It never
+# fails on scores — a bad week is information for the batch table, not an
+# alarm. It fails only when a step could not run or the upload was refused.
+#
+# The agent runs with every interrupt approved, so it has a shell and the
+# network. Three things bound what that can reach:
+#   - it runs in the eval Docker image (evals/agency-agent/run-in-docker.sh),
+#     which passes in only the LLM key and mounts only the run directory;
+#   - each secret is set on the one step that needs it, never job-wide;
+#   - the LLM key is a dedicated one (AGENT_EVALS_OPENAI_API_KEY) with its own
+#     spend limit, so a leak costs at most that limit until it is rotated.
+# The run directory is saved as an artifact only when the upload failed (it
+# holds every trace, and artifacts on a public repo are public): download it,
+# then `agency eval upload <dir>` from a linked checkout.
 #
 # Trigger it manually (workflow_dispatch) to try a cheaper suite or fewer
 # trials without waiting for Sunday.
@@ -34,12 +42,9 @@ jobs:
     # Forks have neither the secrets nor a reason to spend the money.
     if: github.repository == 'egonSchiele/agency-lang'
     runs-on: ubuntu-latest
-    timeout-minutes: 90
+    timeout-minutes: 180
     environment: ci-credentials
     env:
-      OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
-      STATELOG_API_KEY: ${{ secrets.STATELOG_API_KEY }}
-      STATELOG_PROJECT_URL: ${{ secrets.STATELOG_PROJECT_URL }}
       TRIALS: ${{ inputs.trials || '3' }}
       SUITE: ${{ inputs.suite || 'evals/agency-agent' }}
     steps:
@@ -58,9 +63,13 @@ jobs:
         working-directory: packages/agency-lang
 
       - name: Check configuration
+        env:
+          AGENT_EVALS_OPENAI_API_KEY: ${{ secrets.AGENT_EVALS_OPENAI_API_KEY }}
+          STATELOG_API_KEY: ${{ secrets.STATELOG_API_KEY }}
+          STATELOG_PROJECT_URL: ${{ secrets.STATELOG_PROJECT_URL }}
         run: |
           missing=""
-          [ -n "$OPENAI_API_KEY" ] || missing="$missing OPENAI_API_KEY"
+          [ -n "$AGENT_EVALS_OPENAI_API_KEY" ] || missing="$missing AGENT_EVALS_OPENAI_API_KEY"
           [ -n "$STATELOG_API_KEY" ] || missing="$missing STATELOG_API_KEY"
           [ -n "$STATELOG_PROJECT_URL" ] || missing="$missing STATELOG_PROJECT_URL"
           if [ -n "$missing" ]; then
@@ -68,25 +77,37 @@ jobs:
             exit 1
           fi
 
+      # The image the agent runs in, built from this checkout so it tests the
+      # code at HEAD, not a published release.
+      - name: Build the eval image
+        working-directory: packages/agency-lang
+        run: make eval-image
+
       - name: Pick a run directory
         working-directory: packages/agency-lang
         run: echo "RUN_DIR=runs/weekly-$(date -u +%Y-%m-%d-%H%M%S)" >> "$GITHUB_ENV"
 
       - name: Link the statelog project
         working-directory: packages/agency-lang
+        env:
+          STATELOG_PROJECT_URL: ${{ secrets.STATELOG_PROJECT_URL }}
         run: node ./dist/scripts/agency.js remote link --url "$STATELOG_PROJECT_URL"
 
       # Flags here are checked against the CLI's registered options by
       # lib/cli/eval/workflowFlags.test.ts, so a rename fails `make test`
-      # instead of the Sunday run.
+      # instead of the Sunday run. The agent command is the suite's own
+      # wrapper (see evals/agency-agent/README.md); the wrapper forwards
+      # OPENAI_API_KEY into the container, and only that.
       # `eval run` exits 2 when some tests errored (they are run rows that
       # grade 0) and 1 when it could not run at all.
       - name: Run the suite
         working-directory: packages/agency-lang
+        env:
+          OPENAI_API_KEY: ${{ secrets.AGENT_EVALS_OPENAI_API_KEY }}
         run: |
           status=0
           node ./dist/scripts/agency.js eval run \
-            --agent-cmd "node $PWD/dist/scripts/agency.js agent --policy recommended -p -- {input}" \
+            --agent-cmd "$PWD/evals/agency-agent/run-in-docker.sh --brain coordinator --policy approve-all --max-tool-call-rounds 100 -p -- {input}" \
             --suite "$SUITE" \
             --trials "$TRIALS" \
             --parallel 2 \
@@ -97,10 +118,13 @@ jobs:
             exit "$status"
           fi
 
+      # The judge graders call the LLM; the key never reaches an agent here.
       # `eval grade` exits 2 when a must-pass gate failed (a score, which this
       # job records rather than fails on) and 1 when grading itself broke.
       - name: Grade
         working-directory: packages/agency-lang
+        env:
+          OPENAI_API_KEY: ${{ secrets.AGENT_EVALS_OPENAI_API_KEY }}
         run: |
           status=0
           node ./dist/scripts/agency.js eval grade "$RUN_DIR" || status=$?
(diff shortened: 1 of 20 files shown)


packages/agency-lang/agency.json
```diff
@@ -16,7 +16,8 @@
   },
   "eval": {
     "limits": {
-      "wallClockSec": 900
+      "wallClockSec": 900,
+      "maxCostUsd": 25
```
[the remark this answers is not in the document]
egonSchiele: Verified (makeRunner reads costCapFromConfig per job). df240361f adds a batch-wide budget: eval.limits.maxBatchCostUsd (lib/eval/run/batchBudget.ts). runSuite sums each finished run's spend from its statelog and stops starting tests once the cap is crossed; in-flight runs finish under maxCostUsd, so the worst case is batch cap + parallel × per-run cap, documented in docs/site/cli/eval.md. agency.json sets maxBatchCostUsd 25 and maxCostUsd 10. eval run reports costUsd/batchCostCapExceeded and prints a red line; a cut-short batch uploads as incomplete. Judge costs stay outside, as the doc says.

.github/workflows/agent-evals.yml
```diff
@@ -0,0 +1,106 @@
...
+          node ./dist/scripts/agency.js eval run \
+            --agent-cmd "node $PWD/dist/scripts/agency.js agent --policy recommended -p -- {input}" \
+            --suite "$SUITE" \
+            --trials "$TRIALS" \
+            --parallel 2 \
+            --out "$RUN_DIR"
+
+      - name: Grade
+        working-directory: packages/agency-lang
+        run: node ./dist/scripts/agency.js eval grade "$RUN_DIR"
```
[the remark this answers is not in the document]
egonSchiele: Verified: the grade action exits 2 on gatesPassed=false and 1 via failProjectCommand. Rather than a new CLI mode, df240361f has the workflow step capture the exit status: warn and continue on 2, fail on anything else — a genuine grading error still fails the job and the upload still happens on a failed gate.

packages/agency-lang/docs/dev/evals/eval-tracking.md
```diff
@@ -200,3 +200,17 @@ trace parser, and the annotation fold are not exported.
   `status`, `endedAt`, `eventCount`, `suiteSource`/`suiteSha`, and `score`,
   and silent runs, because statelog needs those and must not derive them
   itself.
+
+## The weekly CI run
+
+`.github/workflows/agent-evals.yml` runs `evals/agency-agent` with three
+trials every Sunday (03:00 UTC), grades it, and uploads it to the linked
+statelog project. `workflow_dispatch` takes `trials` and `suite` overrides for
+a cheap dry run. It needs the `OPENAI_API_KEY` and `STATELOG_API_KEY` secrets
```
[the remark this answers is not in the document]
egonSchiele: Reworded in df240361f: "a cheaper smoke run (it still calls the paid agent, grades, and uploads)".

.github/workflows/agent-evals.yml
```diff
@@ -0,0 +1,117 @@
...
+        run: echo "RUN_DIR=runs/weekly-$(date -u +%Y-%m-%d-%H%M%S)" >> "$GITHUB_ENV"
+
+      - name: Link the statelog project
+        working-directory: packages/agency-lang
+        run: node ./dist/scripts/agency.js remote link --url "$STATELOG_PROJECT_URL"
+
+      # Flags here are checked against the CLI's registered options by
+      # lib/cli/eval/workflowFlags.test.ts, so a rename fails `make test`
+      # instead of the Sunday run.
+      - name: Run the suite
```
egonSchiele, line 103: `eval run` calls `process.exit(1)` when `result.errorCount > 0` (scripts/agency.ts, end of the run action). One agent crash or one wall-clock kill among the nine runs fails this step, so Grade and Upload never run and the week is lost. That contradicts the header (an errored test is a run row that grades 0). Suggest handling this the way the Grade step handles exit 2: capture the status, and go on when the run directory exists. Or give `eval run` a distinct exit code for "some tests errored" so the workflow can tell it from "could not run".

.github/workflows/agent-evals.yml
```diff
@@ -0,0 +1,117 @@
+# Weekly eval run of the agency agent, uploaded to statelog so its accuracy
+# can be tracked over time (see packages/agency-lang/docs/dev/evals/eval-tracking.md).
+#
+# This spends real money: about $21 at three tests x three trials, plus the
+# grading judges. eval.limits.maxCostUsd in packages/agency-lang/agency.json
+# caps each agent run separately (there is no batch-wide budget), so the worst
```
egonSchiele: Stale: this PR adds `maxBatchCostUsd`, so there is a batch-wide budget now. But agency.json does not set it, so the weekly run still has none; the worst case is 9 runs x $5. The dev note (eval-tracking.md line 213) says the batch cap is in agency.json, and the PR body says `maxCostUsd: 25`. Suggest setting `maxBatchCostUsd` in agency.json (25 matches the calibration in the body), keeping `maxCostUsd: 5` per run, and fixing this comment and the body to match.

.github/workflows/agent-evals.yml
```diff
@@ -0,0 +1,117 @@
...
+  evals:
+    # Forks have neither the secrets nor a reason to spend the money.
+    if: github.repository == 'egonSchiele/agency-lang'
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    environment: ci-credentials
+    env:
+      OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+      STATELOG_API_KEY: ${{ secrets.STATELOG_API_KEY }}
+      STATELOG_PROJECT_URL: ${{ secrets.STATELOG_PROJECT_URL }}
```
egonSchiele: The PR body calls this a repository variable, which would be `${{ vars.STATELOG_PROJECT_URL }}`. As written it must be a secret, and the Check configuration step fails if it was created as a variable. Either is fine; make the two agree. (A serve URL is not sensitive, so `vars` would let you read it in the run log.)

packages/agency-lang/lib/eval/run/batchBudget.ts
```diff
@@ -0,0 +1,48 @@
...
+    exhausted: () => exhausted,
+    spentUsd: () => spent,
+    exceededMessage: () =>
+      `batch cost cap exceeded: $${spent.toFixed(2)} spent, cap $${(capUsd ?? 0).toFixed(2)} ` +
+      `(eval.limits.maxBatchCostUsd raises it); no further tests will start`,
+  };
+}
+
+/** A finished run's LLM spend, read from its statelog; 0 when it left none. */
+export function runCostUsd(statelogPath: string | null): number {
```
egonSchiele, line 45: In parallel mode the pool already tails each run's cost for the status board (runSuite.ts around line 512), so a finished run's statelog is read twice: once by that tailer and again here. Not wrong, but the board's tailer already has the number. Could `executeTest` hand the final cost back from the tailer it created, and this reader stay only for the sequential path?

egonSchiele (commented):
Read the whole diff and checked the exit paths in scripts/agency.ts and the tailer. The batch budget and the grade-step handling look right and the three Copilot threads read as already addressed. One thing will break the weekly upload, and the cost story is told three different ways.

1. The run step fails the job whenever one agent run errors (inline).
2. The batch cap exists in code but the weekly run does not set it, and the workflow header, agency.json, the dev note, and the PR body each say something different about the cap (inline).
3. STATELOG_PROJECT_URL is read from secrets but described as a repository variable (inline).

Two smaller ones inline on runCostUsd and the two cost totals.

packages/agency-lang/scripts/agency.ts
```diff
@@ -905,6 +905,12 @@ export function createProgram(deps: CliDependencies = {}): Command {
         if (costUsd !== undefined) {
           console.log(`total LLM cost: $${costUsd.toFixed(2)}`);
         }
+        if (result.batchCostCapExceeded) {
```
egonSchiele, line 905: Two totals for the same money: `totalRunCostUsd(result.runDir)` just above and `result.costUsd` from the budget, computed two different ways. If they ever disagree the printed total and the cap decision will not match. Suggest printing `result.costUsd` and dropping the second computation, or the reverse.

.github/workflows/agent-evals.yml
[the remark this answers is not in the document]
egonSchiele: The owner chose to keep STATELOG_PROJECT_URL as a secret (it was created as one), so the workflow reads secrets. The PR body and dev note now say secret too (557396d64).

egonSchiele: Made them agree the other way in 557396d64: the owner stored the URL as a secret, so the workflow, PR body, and dev note all say secret now.

packages/agency-lang/docs/dev/evals/eval-tracking.md
```diff
@@ -200,3 +200,26 @@ trace parser, and the annotation fold are not exported.
...
   itself.
+
+## The weekly CI run
+
+`.github/workflows/agent-evals.yml` runs `evals/agency-agent` with three
+trials every Sunday (03:00 UTC), grades it, and uploads it to the linked
+statelog project. `workflow_dispatch` takes `trials` and `suite` overrides for
+a cheaper smoke run (it still calls the paid agent, grades, and uploads). It
+needs the `OPENAI_API_KEY`, `STATELOG_API_KEY`, and `STATELOG_PROJECT_URL`
+secrets (the last is a serve URL, as `remote link --url` takes).
```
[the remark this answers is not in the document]
egonSchiele: The owner's setup is three secrets; the dev note now says so explicitly and notes the URL is kept as a secret by choice, not sensitivity (557396d64).

.github/workflows/agent-evals.yml
egonSchiele: Right — that agency.json edit had been lost. 557396d64 sets maxBatchCostUsd 25 and maxCostUsd 5 and fixes the header comment and the PR body.

packages/agency-lang/lib/eval/run/runSuite.ts
```diff
@@ -193,6 +202,9 @@ export async function runSuite(
         },
         { runner: deps.runner },
       );
+      if (budget.add(runCostUsd(run.statelogPath)) && progress) {
+        console.warn(budget.exceededMessage());
+      }
```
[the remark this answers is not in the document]
egonSchiele: Verified: batchStatistics throws on an uneven grid and grade called it unguarded. 557396d64 takes the second option, matching what statelog already does: batchStatisticsByBatchTolerant reports such a batch as incomplete with the reason, eval grade lists it as "has no statistics" (every run is still graded, the exit code is unaffected), and the upload proceeds; statelog shows the batch as incomplete. The budget keeps stopping immediately, so it stays a real cap.

.github/workflows/agent-evals.yml
egonSchiele: Verified. 557396d64 gives eval run a distinct exit code: 2 = some tests errored (directory complete), 1 = could not run. The workflow treats 2 like the grade step's 2 — a warning, then grade and upload.

packages/agency-lang/scripts/agency.ts
egonSchiele: Done in 557396d64: eval run prints result.costUsd (the number the cap decided on) and totalRunCostUsd is removed.

packages/agency-lang/lib/eval/run/batchBudget.ts
egonSchiele: Leaving as is: the board's tailers exist only in parallel mode inside runPool, while the budget needs the cost in both modes at the point the run finishes. One extra read of a just-written statelog per run is cheaper than threading the tailer through executeTest's two call sites.

egonSchiele:
056a6d037 switches the run to the suite's Docker wrapper with --policy approve-all (the new tests write and run code), and bounds what that can reach: secrets are step-scoped, the LLM key is a dedicated AGENT_EVALS_OPENAI_API_KEY, and the run directory is only saved as an artifact when the upload failed. Rebased onto main now that #949/#950 are in. The Docker path has not run in Actions yet.

PR #947, opened 2026-08-28
Add a writing rewrite agent, an agency agent rewrite subagent, and the writing-rewrite eval suite

egonSchiele:
## What this adds

A rewrite agent for the writing standard library, a matching `agency agent` subagent, and an eval suite that measures how well the rewriter fixes the flaws the reviewer finds.

### `std::agents/writing/rewrite`

`writingRewriteAgent(text, task, guidelines, context, passes = 1, ...)` returns `Result<string>`. Each pass runs `writingReviewAgent` on the current text and then makes one LLM call that applies every finding. The loop stops early when the reviewer has no findings. A reviewer failure returns `failure(...)` rather than the original text, so a caller never mistakes unreviewed prose for a finished rewrite.

The rewriting prompt tells the model to keep every fact, number, and identifier and to add nothing, so the graders can check faithfulness.

### `agency agent --agent rewrite`

A new coordinator subagent. A small structured call splits the request into the prose and the task description, then calls the stdlib agent. The existing `writing` subagent is unchanged, apart from a fix: it passed its argument as `work:` but the reviewer's parameter is `text:`, so it crashed on every call.

### `evals/writing-rewrite/`

23 tests. Each `test.json` points at a test in `evals/writing-review/` for its `files` and `graderFiles`, so the two suites share inputs. The 21 harvested tests come from the owner's own bad/good pairs; two controls hold clean prose. Graders:

- `produces-text` (must pass)
- `flaws-fixed`: one rubric judge per editor point, reporting `[fixed]` / `[remains]`
- `matches-cuts`, `faithful`
- `leaves-clean-alone` on the two controls

`evalMain` runs one pass; `evalMainTwoPasses` runs two, so a second run can show whether the extra pass earns its cost.

Shared readers (`getSourceFileText`, `getAssignment`, `harvest`, `editorPoints`) moved from `reviewGraders.ts` to `evals/writing-review/lib/testFiles.ts`.

## How to run

```
pnpm run agency eval run stdlib/agents/writing/rewrite.agency evals/writing-rewrite --trials 3 -o runs/rewrite-baseline
pnpm run agency eval run stdlib/agents/writing/rewrite.agency evals/writing-rewrite --node evalMainTwoPasses --trials 3 -o runs/rewrite-two-passes
```

A first one-pass smoke run fixed 1 of 5 editor points on `comment-abrupt-verb`, which is the gap the suite exists to measure.

## Docs

`docs/dev/agents/writing-rewrite-agent.md`, indexed from the `agency-agent-docs` skill; the stdlib reference page is generated by `make`.


packages/agency-lang/docs/dev/agents/writing-rewrite-agent.md
```diff
@@ -0,0 +1,48 @@
+# The writing rewrite agent
```
egonSchiele, line 1: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/docs/site/stdlib/agents/writing/rewrite.md
```diff
@@ -0,0 +1,70 @@
+---
+name: "rewrite"
+description: "Rewrite prose from the writing reviewer's findings and return the"
+---
+
+# rewrite
+
+new text.
```
egonSchiele: ?? Why does this just say "new text" with no other context?

packages/agency-lang/evals/writing-review/lib/testFiles.ts
```diff
@@ -0,0 +1,67 @@
...
+import * as fs from "fs";
+import * as path from "path";
+
+import type { Test } from "agency-lang/eval";
+
+/** Mirrors `WritingReviewEvalInput` in stdlib/agents/writing/review.agency. */
+export type WritingReviewInput = { assignment: string; sourceFile: string };
+/** The reviewed text, read from the test's seeded workdir. Fixture paths
+ *  are authored, but a path that escapes the workdir reads as missing. */
+export function getSourceFileText(workdir: string, test: Test<WritingReviewInput>): string {
```
egonSchiele: I believe we have a function already that can be used to get the source of any file in the working directory. So this function is unnecessary.

egonSchiele: Actually, I am confused about what the script is for. Is this a one-time script for harvesting evals from the writing review agent for the rewrite agent? Is this script a good way for keeping those two eval suites in sync? If not, I would suggest dropping it and just aiming to keep them both in sync by hand in the future, knowing that they won't be perfectly in sync.

packages/agency-lang/evals/writing-rewrite/also-answers-to/test.json
```diff
@@ -0,0 +1,9 @@
+{
+  "description": "Odd phrasing for a command alias, plus an unneeded aside.",
+  "files": "../../writing-review/also-answers-to/files",
```
egonSchiele: This feels really not great. Each eval suite should be self-contained directory, but here it looks like the tests for this eval suite are directly reaching out into the code for another eval suite. This means that every time we update the writing review eval suite, we will also be updating the rewrite eval suite without knowing it. I think in this case, you should just copy these files, and we should accept the duplication and the fact that we will need to keep these suites in sync by hand.

packages/agency-lang/docs/dev/agents/writing-rewrite-agent.md
egonSchiele: Cut the dev note to the three decisions and the eval shape; the smoke-run story and the guard mechanics are gone (03e80bf78 / 257f20042).

packages/agency-lang/docs/site/stdlib/agents/writing/rewrite.md
egonSchiele: The @summary wrapped onto a second line and the doc generator only takes the first. One line now; page regenerated.

packages/agency-lang/evals/writing-review/lib/testFiles.ts
egonSchiele: Not a harvest script; it was the readers both suites' graders called. Gone now: reviewGraders.ts is back to its pre-PR shape, and the rewrite suite has its own copy of the two small readers, using the grader's workdirFile for the source text.

packages/agency-lang/stdlib/agents/writing/rewrite.agency
```diff
@@ -0,0 +1,193 @@
...
+        guidelines: guidelines,
+        context: context,
+        maxCost: maxCost,
+        maxTime: maxTime,
+        model: model,
+        provider: provider,
+      )
+      if (reviewed is failure(err)) {
+        reviewFailure = "${err}"
+      }
```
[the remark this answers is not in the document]
egonSchiele: Confirmed: the branch was dead. writingReviewAgent now takes failOpen (default true) and the rewriter passes false, so a failed review comes back as a failure. No regression test: nothing in the Agency test harness can make the reviewer fail without a live model.

packages/agency-lang/evals/writing-rewrite/overloaded-paragraph/graders.ts
```diff
@@ -0,0 +1,3 @@
+import { harvestedRewriteGraders } from "../lib/rewriteGraders.js";
+
+export default harvestedRewriteGraders();
```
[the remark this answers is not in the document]
egonSchiele: Added graderFiles/notes.md for the copy, with the planted flaws as bullets.

packages/agency-lang/stdlib/agents/writing/rewrite.agency
```diff
@@ -0,0 +1,193 @@
+/** @module
+  @summary Rewrite prose from the writing reviewer's findings and return the
+  new text.
```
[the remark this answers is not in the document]
egonSchiele: Done.

packages/agency-lang/evals/writing-rewrite/also-answers-to/test.json
egonSchiele: Copied. Every rewrite test now has its own files/ and graderFiles/ (488K total), and the README says the copies are kept in step by hand.

PR #955, opened 2026-08-29
Bound eval upload concurrency; retry shed statelog requests

egonSchiele:
Uploading a large run group (e.g. 69 runs) fired every run's requests at once — `Promise.all` over all directories, each with an upload-state GET, bulk POSTs, and annotation POSTs — and Cloud Run shed the burst with 429s.

- `evalUpload` uploads through `mapInParallel` with a pool of 6; results keep input order.
- `statelogRequest` retries 429/503 up to three times (0.5s / 2s / 8s). Those statuses mean the request was shed before handling, so a retry cannot double-apply anything; every other status and network errors are untouched. Applies to all sealed statelog clients.
- Tests: retry-until-success with the exact backoff sequence, give-up-after-last-delay, non-retryable statuses never retried.

Related: uploads of runs graded after 8/24 fail with `Unrecognized key: "graderFiles"` until a release containing #917 is published and statelog bumps to it — separate from this PR.

https://claude.ai/code/session_01YSdUvTQqBfWKBERTyspUHR

### DIFF
--- a/packages/agency-lang/lib/cli/statelog/statelogRequest.ts
+++ b/packages/agency-lang/lib/cli/statelog/statelogRequest.ts
@@ -39,7 +39,19 @@ export type StatelogRequestOptions = {
   contentType?: "when-body" | "always";
   /** Diagnostic-only redaction, forwarded to readJsonBody. */
   sanitizeDiagnostic?: (raw: string) => string;
+  /** Test seam for the retry backoff; default real timer. */
+  sleep?: (ms: number) => Promise<void>;
 };
+
+// A 429 or 503 means the server (or Cloud Run in front of it) shed the
+// request before handling it, so a retry cannot double-apply anything.
+// Waits are long enough that a burst of concurrent uploads spreads out.
+const RETRYABLE_STATUSES = new Set([429, 503]);
+export const RETRY_DELAYS_MS = [500, 2_000, 8_000];
+
+function defaultSleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
 
 export type StatelogRequestResult =
   { ok: true; value: unknown; status: number } | { ok: false; failure: StatelogFailure };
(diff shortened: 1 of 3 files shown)


packages/agency-lang/lib/cli/statelog/statelogRequest.ts
```diff
@@ -62,15 +74,22 @@ export async function statelogRequest(
...
   let response: Response;
-  try {
-    response = await fetch(options.url, {
-      method: options.method,
-      headers,
-      body: serializedBody,
-    });
-  } catch (error) {
-    return { ok: false, failure: { kind: "unreachable", cause: message(error) } };
+  for (let attempt = 0; ; attempt += 1) {
```
egonSchiele: why no limit on number of attempts?

egonSchiele: There is one — the loop broke once attempt reached RETRY_DELAYS_MS.length, so at most 3 retries (4 requests). But the for(;;) shape hid it; 6d3033117 restructures to 'for (const delayMs of RETRY_DELAYS_MS)', so the bound is the list itself, and the give-up test pins it at RETRY_DELAYS_MS.length + 1 fetches.

packages/agency-lang/lib/cli/statelog/statelogRequest.ts
```diff
@@ -62,15 +74,22 @@ export async function statelogRequest(
...
+        headers,
+        body: serializedBody,
+      });
+    } catch (error) {
+      return { ok: false, failure: { kind: "unreachable", cause: message(error) } };
+    }
+    if (!RETRYABLE_STATUSES.has(response.status) || attempt >= RETRY_DELAYS_MS.length) {
+      break;
+    }
+    await sleep(RETRY_DELAYS_MS[attempt]);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 6d3033117 — each shed response's body is cancelled (best-effort, errors swallowed) before the sleep, and the retry test asserts cancel is called once per shed response.

PR #964, opened 2026-08-30
Print the transcript on resume; add /rename for agent sessions

egonSchiele:
Two changes to `agency agent` saved sessions (follow-up to #962).

## Print the transcript on resume

`--continue` / `--resume` now prints the saved conversation before restoring, so you can see what the session was about, or resume just to read it. User lines are dimmed after a `>`, replies are rendered as markdown: the same drawing as when the turns ran.

Why it reads the checkpoint file rather than the restored threads: `restore()` throws a signal and replays to the saved point (the REPL waiting for a line), so nothing placed after it, and no step before the `repl(...)` call, runs on resume. `_readTranscript` in `lib/stdlib/agentSessions.ts` parses the file with `Checkpoint.fromJSON`, picks the `main` thread on the saved stack (or the largest top-level thread for a brain that labels none), and returns its user and assistant messages. `showTranscript` in `repl.agency` draws them, then `startSession` restores.

## `/rename <title>`

Sets the session title (`/rename minnesota news 8/29`). The title is pinned so the periodic LLM refresh leaves it alone, and the record is marked dirty so the save that runs after every REPL line writes both files even though no turn ran. That keeps the checkpoint's copy of the record and the record file in agreement.

## Tests

- `tests/agency/agency-agent-sessions.agency`: `renameFlow` (usage on an empty title, cleaned title, save with turns untouched, no second save).
- `tests/agency-js/agent-session-resume`: the fixture now includes the transcript read back from the saved checkpoint file.

Dev doc: `docs/dev/agents/agent-sessions.md` gained sections for both.

https://claude.ai/code/session_01RGbtNzLmRXjpE47dPjmDkL

### DIFF
--- a/packages/agency-lang/docs/dev/agents/agent-sessions.md
+++ b/packages/agency-lang/docs/dev/agents/agent-sessions.md
@@ -73,26 +73,21 @@ step in the checkpoint and come back from a restore as nothing.
 
 ### The transcript on resume
 
-A resumed run prints the saved conversation first, so the user can see
-what it was about (and some resumes are only for reading it). Nothing
-placed after `restore()` runs: it throws a signal, the runtime replays to
-the saved point, and that point is the `repl(...)` call, whose earlier
-steps are already complete. So the transcript is read from the checkpoint
-file, not from the restored threads: `_readTranscript` in
-`agentSessions.ts` parses the file with `Checkpoint.fromJSON`, finds the
-`main` thread on the saved stack (or the largest top-level thread for a
-brain that labels none), and returns its user and assistant messages.
-`showTranscript` in `repl.agency` draws them the way the turns were drawn
-when they ran, then `startSession` restores.
+A resumed run prints the saved conversation before it restores. It is
+read from the checkpoint file (`_readTranscript` in `agentSessions.ts`),
+not from the restored threads, because `restore()` replays to the saved
+point and runs nothing placed after it. The thread is the one the brain
+opens with `thread(session: ...)`; each brain names it in its
+`AgentBrain.session` field, and the title refresh reads the same thread.
+`showTranscript` in `transcript.agency` draws the messages with the same
+renderers the live turns use.
 
 ### `/rename`
 
-`/rename <title>` sets the record's title (cleaned with `cleanTitle`),
-pins it so the periodic title refresh leaves it alone, and marks the
-record dirty. `recordTurn` returns a save target for a dirty record even
-when no turn ran, so the same `_sessionOnSubmit` path writes both files
-after the command, and the checkpoint's copy of the record agrees with
-the record file. The pin is a `let` global, so it comes back on resume.
+`/rename <title>` sets the record's title, pins it so the periodic title
+refresh leaves it alone, and marks the record dirty. `recordTurn` returns
+a save target for a dirty record even when no turn ran, so the save that
+follows every REPL line writes it.
 
 Because the `let` globals come back from the saved run, `--model`,
 `--policy`, and similar flags passed on a resume are ignored; the session
--- a/packages/agency-lang/docs/dev/agents/agent-sessions.md
+++ b/packages/agency-lang/docs/dev/agents/agent-sessions.md
@@ -71,6 +71,29 @@ TypeScript module state (`installSessionHooks`, before the restore), not as
 an Agency value: a closure built by an Agency statement would be a completed
 step in the checkpoint and come back from a restore as nothing.
 
+### The transcript on resume
+
+A resumed run prints the saved conversation first, so the user can see
+what it was about (and some resumes are only for reading it). Nothing
+placed after `restore()` runs: it throws a signal, the runtime replays to
+the saved point, and that point is the `repl(...)` call, whose earlier
+steps are already complete. So the transcript is read from the checkpoint
+file, not from the restored threads: `_readTranscript` in
+`agentSessions.ts` parses the file with `Checkpoint.fromJSON`, finds the
+`main` thread on the saved stack (or the largest top-level thread for a
+brain that labels none), and returns its user and assistant messages.
+`showTranscript` in `repl.agency` draws them the way the turns were drawn
+when they ran, then `startSession` restores.
+
+### `/rename`
+
+`/rename <title>` sets the record's title (cleaned with `cleanTitle`),
+pins it so the periodic title refresh leaves it alone, and marks the
+record dirty. `recordTurn` returns a save target for a dirty record even
+when no turn ran, so the same `_sessionOnSubmit` path writes both files
+after the command, and the checkpoint's copy of the record agrees with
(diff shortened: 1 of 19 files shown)


packages/agency-lang/docs/dev/agents/agent-sessions.md
```diff
@@ -71,6 +71,29 @@ TypeScript module state (`installSessionHooks`, before the restore), not as
 an Agency value: a closure built by an Agency statement would be a completed
 step in the checkpoint and come back from a restore as nothing.
 
+### The transcript on resume
```
egonSchiele, line 74: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/lib/agents/agency-agent/lib/resume.agency
```diff
@@ -49,6 +56,24 @@ export def markTurnDone() {
   _turnDone = true
 }
 
+/** `/rename <title>`: give the session a title of the user's choosing.
+  Returns the line to show. The title is written with the next save,
+  which the REPL runs after this line whether or not a turn ran. */
+export def renameSession(title: string): string {
+  const record = _current
+  if (record == null) {
+    return "This run is not saved, so it cannot be renamed."
```
egonSchiele: why not save it if it isn't saved yet?

packages/agency-lang/lib/agents/agency-agent/lib/resume.agency
```diff
@@ -111,8 +136,15 @@ def pickSession(): SessionRecord | null {
...
-  one and the program continues from the saved point. */
-export def startSession(chosen: SessionRecord | null, brain: string) {
+  one and the program continues from the saved point. `showTranscript`
+  prints the saved conversation first, because nothing written after
+  `restore` runs: it replays to the saved point, which is the REPL
+  waiting for a line. */
+export def startSession(
+  chosen: SessionRecord | null,
+  brain: string,
+  showTranscript: (messages: any[]) -> void,
```
egonSchiele: why... pass this in as a function?

packages/agency-lang/lib/stdlib/agentSessions.ts
```diff
@@ -100,6 +101,62 @@ export function _readCheckpointFile(dir: string, id: string): unknown {
...
+    if (m.role !== "user" && m.role !== "assistant") continue;
+    const text = messageText(m.content);
+    if (text !== "") out.push({ role: m.role, content: text });
+  }
+  return out;
+}
+
+// Every frame on the saved stack carries a thread store; the same thread
+// can appear in several, so take the fullest copy.
+function mainThread(cp: Checkpoint): ThreadJSON | null {
```
egonSchiele: There has to be a better way to select the main thread than "most messages". That is not a good metric.

egonSchiele: Read https://agency-lang.com/guide/cross-thread-context.html

packages/agency-lang/docs/dev/agents/agent-sessions.md
egonSchiele: Trimmed. The doc now says where the transcript comes from and what /rename does; the comments in resume.agency and agentSessions.ts lost the reasoning about restore() replay and the callback.

packages/agency-lang/lib/agents/agency-agent/lib/resume.agency
egonSchiele: It is saved: the rename marks the record dirty and the save that follows every REPL line writes it. `_current` is null only for one-shot runs, which have no REPL, so this branch is a type guard; its message is now just "No session to rename."

packages/agency-lang/lib/agents/agency-agent/lib/resume.agency
egonSchiele: It was dodging an import cycle (repl imports resume). The drawing now lives in transcript.agency, shared by the live turns and the resume, and startSession imports it directly.

packages/agency-lang/lib/stdlib/agentSessions.ts
egonSchiele: Now selected by session name. Each brain names its conversation thread in a new `AgentBrain.session` field (`main` for the coordinator, `simple` for simple), the checkpoint's `sessions` map gives the thread id, and a new `std::thread` `sessionThreadId(name)` does the same for the live store so the title refresh uses the same rule. The largest-thread fallback is gone.

packages/agency-lang/lib/stdlib/agentSessions.ts
```diff
@@ -100,6 +101,62 @@ export function _readCheckpointFile(dir: string, id: string): unknown {
...
+ * that labels none, the top-level thread with the most messages). Tool
+ * calls and results are left out. Read from the checkpoint file before it
+ * is restored, because a restore replays to the saved point and runs no
+ * code of its own afterwards.
+ */
+export function _readTranscript(checkpointJson: unknown): TranscriptMessage[] {
+  const cp = Checkpoint.fromJSON(checkpointJson);
+  if (!cp) return [];
+  const thread = mainThread(cp);
+  if (!thread) return [];
```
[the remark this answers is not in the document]
egonSchiele: Done.

packages/agency-lang/lib/stdlib/agentSessions.ts
```diff
@@ -100,6 +101,62 @@ export function _readCheckpointFile(dir: string, id: string): unknown {
...
+  const out: TranscriptMessage[] = [];
+  for (const m of thread.messages) {
+    if (m.role !== "user" && m.role !== "assistant") continue;
+    const text = messageText(m.content);
+    if (text !== "") out.push({ role: m.role, content: text });
```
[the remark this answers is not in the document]
egonSchiele: Done: every turn adds an assistant reply, and the third user line goes in as a content array.

packages/agency-lang/lib/agents/agency-agent/lib/resume.agency
```diff
@@ -132,20 +164,29 @@ export def startSession(chosen: SessionRecord | null, brain: string) {
...
+    return {
+      dir: projectDir(),
+      record: record
+    }
+  }
   _turnDone = false
   record.turns = record.turns + 1
   record.lastActive = now()
-  if (shouldRefreshTitle(record.turns)) {
+  if (!_titlePinned && shouldRefreshTitle(record.turns)) {
```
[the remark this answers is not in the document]
egonSchiele: Done: the test now completes turn 1 (the first refresh point) after the rename and checks the title is unchanged.

Issue #971, opened 2026-08-30
Pure Agency code can reach any JS global (process, fetch, globalThis) under --agency-only with no interrupt


egonSchiele:
## What happens

Agency compiles to JavaScript, and an identifier the compiler does not know is emitted verbatim. So any JavaScript global is reachable from pure Agency code: `process`, `globalThis`, `fetch`, `Function`, `Reflect`, and everything hanging off them. None of it raises an interrupt, so no handler, policy, or `--reject '*'` sees it. `--agency-only` validates the import closure, but the escape needs no import.

## Reproduction

```
// bar.agency
node main() {
  print(process.env.HOME)
}
```

```
agency run --agency-only bar.agency     # prints the home directory
```

Going further, under the strongest flags the sandbox has:

```
node main() {
  const fs = process.getBuiltinModule("fs")
  fs.writeFileSync("PWNED.txt", "wrote via process.getBuiltinModule")
  fetch("http://127.0.0.1:9/")
}
```

```
agency run --agency-only --reject '*' g.agency    # PWNED.txt is written
```

Environment (API keys), the file system, the network, child processes: everything, with no interrupt. This makes the sandbox argument in `docs/dev/cli/test-cli-sandbox.md` ("With --agency-only, every effect the code can perform is an interrupt") false today.

## Why nothing stops it

- The typechecker's undefined-function diagnostic (AG4004, `lib/typeChecker/undefinedFunctionDiagnostic.ts`) is a **warning** by default, and `agency run` runs anyway. It fires for `process.getBuiltinModule(...)` and `fetch(...)`.
- It only looks at **calls**. `process.env.HOME` is a value access and gets no diagnostic at all.
- Its allowlist of JS globals (`JS_GLOBALS` in `lib/typeChecker/resolveCall.ts`) exists to avoid false positives on interop, which is the opposite goal from a sandbox.

## Fix direction

Under `--agency-only` (and probably `std::agency` `compile`, which uses the same validator), any free identifier that does not resolve to an Agency declaration, an import, or a reviewed pure allowlist (`Math`, `JSON`, `parseInt`, ...) must be a **compile error**, for value accesses as well as calls. The allowlist needs a security review of its own: `Function`, `eval`, `globalThis`, `Reflect`, `process`, `fetch`, `require`, `import` cannot be on it. A resolver-based check (every name is bound, or refused) is stronger than a denylist, since a denylist misses the next global Node adds.

[...]

PR #973, opened 2026-08-30
Refuse JS-capability names under --agency-only (#971, layer 1)

egonSchiele:
## What this does

Closes the biggest hole in `--agency-only` (#971): pure Agency code compiled
to JavaScript could name any host global — `process`, `fetch`, `eval`,
`new Function` — and reach the outside world with **no interrupt**, so no
policy or `--reject '*'` could stop it. This refuses those names at compile
time.

**This is defense in depth and clearer errors, not the security boundary.** A
runtime-computed property key (`m[a + b]`) defeats any syntactic check, so the
real backstop for code-from-strings is disabling code generation from strings
(roadmap A1 layer 2), and ultimately removing ambient authority (a process
boundary / SES — roadmap). This PR is layer 1.

## How

- `typechecker.jsGlobals: "all" | "sandbox"` (new config). `--agency-only`
  sets `"sandbox"` in `compileValidatedClosure`, which also forces
  `undefined{Functions,Variables}` to error. Default path is byte-identical.
- `SANDBOX_JS_GLOBALS`: a **separate reviewed allowlist** of globals that
  cannot reach the host (not a filter over the interop `JS_GLOBALS`). The four
  resolvers take the registry as a defaulted argument.
- The two general passes resolve against the allowlist under the sandbox,
  catching bare names, calls, namespace members, and value-access bases
  (`process.env`).
- `checkSandboxNames` (new pass) covers the four positions the general passes
  miss, each reproduced as an escape during design:
  - `new` callees (`new Function/Proxy/WebSocket`) — sound (className is a
    grammar literal); this is what closes `new Proxy`/`new WebSocket`, which
    layer 2 never covers.
  - the `constructor`/`prototype`/`__proto__` walk (AG4011) — best-effort,
    literal spellings only.
  - tag arguments (`@validate`/`@jsonSchema`) and array/object default
    parameter values — declaration-hanging expressions outside any scope body.

## Reviewing

- The rule is shared through one registry-parametric resolver; only the set of
  walked positions differs (see `docs/dev/compiler/agency-only-bound-names.md`).
- Coverage equals the traversal's reach — stated plainly in the dev note,
  because the arc found five escapes by hand. A future position is one resolver
  call from covered, and the security guarantee does not rest here.
- Tests: `lib/typeChecker/sandboxRegistry.test.ts` (registry, default path
  unchanged), `lib/compiler/compileSandboxed.test.ts` (all five escapes refused
  + positive controls), and `tests/agency-js/test-cli-agency-only` (end to end;
  refused fixtures use benign payloads so a regressed refusal cannot do harm).

## Also in this branch

The security-arc docs written earlier this session: `docs/dev/security/goal.md`
and `roadmap.md` (the whole no-OS-sandbox plan), and
`docs/dev/hosting/how-hosted-serving-works.md`. The design and plan (with three
rounds of review) live under `docs/superpowers/` (gitignored).

https://claude.ai/code/session_01SwUwoL2LVZp8KmCS3GZHjS

### DIFF
--- a/packages/agency-lang/docs/dev/cli/test-cli-sandbox.md
+++ b/packages/agency-lang/docs/dev/cli/test-cli-sandbox.md
@@ -17,8 +17,15 @@ under each name it is given.
 
 ## Why the two flags together are the safety argument
 
-- With `--agency-only`, every effect the code can perform is an interrupt.
-  There is no other way to touch the world from pure Agency.
+- With `--agency-only`, every effect the code can perform is an interrupt,
+  and the JS-globals bind-check refuses the compiled-to-JavaScript escapes
+  that would otherwise bypass interrupts entirely (`process`, `fetch`,
+  `eval`, `new Function`, the `.constructor` walk, tag arguments, default
+  values). See `docs/dev/compiler/agency-only-bound-names.md`. That check is
+  defense in depth, not the whole boundary: a runtime-computed property key
+  gets through it, so the backstop for code-from-strings is layer 2 in
+  `docs/dev/security/roadmap.md` (A1). There is no other way to touch the
+  world from pure Agency once those hold.
 - With `--reject '*'`, the root handler rejects each of those interrupts.
   The handler chain resolves reject over approve (`lib/runtime/interrupts.ts`,
   "reject > propagate > approve > noResponse"; pinned by
--- a/packages/agency-lang/docs/dev/cli/test-cli-sandbox.md
+++ b/packages/agency-lang/docs/dev/cli/test-cli-sandbox.md
@@ -17,6 +17,13 @@ under each name it is given.
 
 ## Why the two flags together are the safety argument
 
+> These flags reduce attack surface; they are **not yet a complete sandbox**.
+> The bind-check is defense in depth (a runtime-computed property key slips
+> past it), and the containment layers that close the remaining gaps are
+> tracked in `docs/dev/security/roadmap.md` (A1). Do not treat `--agency-only`
+> as sufficient to run fully untrusted code on its own yet.
+
+
 - With `--agency-only`, every effect the code can perform is an interrupt,
   and the JS-globals bind-check refuses the compiled-to-JavaScript escapes
   that would otherwise bypass interrupts entirely (`process`, `fetch`,
--- a/packages/agency-lang/lib/typeChecker/resolveCall.ts
+++ b/packages/agency-lang/lib/typeChecker/resolveCall.ts
@@ -284,18 +284,92 @@ export const JS_GLOBALS: Record<string, JsRegistryEntry> = {
 };
 
 /**
- * Walk a namespace path through `JS_GLOBALS`. Returns the leaf entry if
- * the full chain resolves, otherwise null.
+ * Allowlist of JavaScript globals a program may reference under
+ * `--agency-only` (`typechecker.jsGlobals: "sandbox"`). An entry here is a
+ * PROMISE that the name cannot reach the host. Review every addition against
+ * that promise — this list is a reviewed security boundary, not a
+ * convenience registry like `JS_GLOBALS`.
+ *
+ * It is written out in full rather than filtered from `JS_GLOBALS`, so a
+ * name added to `JS_GLOBALS` for interop never silently becomes allowed
+ * here. It is NOT a subset of `JS_GLOBALS`: it adds constructor names
+ * (`Set`, `Map`, `RegExp`, `Intl`) that `new X()` resolves against but that
+ * `JS_GLOBALS` never listed.
+ *
+ * Deliberately EXCLUDED (each reaches or names the host): `process`,
+ * `console`, `Buffer`, `globalThis`, `eval`, `Function`, `fetch`, `Reflect`,
+ * `Proxy`, `Symbol`, `require`, `import`, `setTimeout`/`setInterval` (run
+ * code after the frame returns, outside the runtime's bookkeeping), and
+ * `Object.create`/`getPrototypeOf`/`setPrototypeOf`/`getOwnPropertyNames`
+ * (prototype and descriptor access).
+ */
+export const SANDBOX_JS_GLOBALS: Record<string, JsRegistryEntry> = {
+  // Pure conversions.
+  parseInt: JS_GLOBALS.parseInt,
+  parseFloat: JS_GLOBALS.parseFloat,
+  isNaN: JS_GLOBALS.isNaN,
+  isFinite: JS_GLOBALS.isFinite,
+  encodeURIComponent: JS_GLOBALS.encodeURIComponent,
+  decodeURIComponent: JS_GLOBALS.decodeURIComponent,
+  encodeURI: JS_GLOBALS.encodeURI,
+  decodeURI: JS_GLOBALS.decodeURI,
+  // Deep-copies plain data; throws on functions, so it yields no capability.
+  structuredClone: JS_GLOBALS.structuredClone,
+  Boolean: JS_GLOBALS.Boolean,
+  // Pure namespaces reused verbatim from JS_GLOBALS.
+  JSON: JS_GLOBALS.JSON,
+  Math: JS_GLOBALS.Math,
+  String: JS_GLOBALS.String,
+  Number: JS_GLOBALS.Number,
+  Promise: JS_GLOBALS.Promise,
+  Array: JS_GLOBALS.Array,
+  // Object minus the prototype/descriptor members (create, getPrototypeOf,
+  // setPrototypeOf, getOwnPropertyNames stay out).
+  Object: namespace({
+    keys: callable({ params: [ANY_T], returnType: stringArray }),
+    values: callable({ params: [ANY_T], returnType: anyArray }),
+    entries: callable({ params: [ANY_T], returnType: anyArray }),
+    fromEntries: callable({ params: [anyArray], returnType: ANY_T }),
+    assign: callable(),
+    freeze: callable({ params: [ANY_T], returnType: ANY_T }),
+  }),
+  // Constructors reached via `new X()`; also valid as value references.
+  Set: callable(),
+  Map: callable(),
+  RegExp: callable(),
+  Date: JS_GLOBALS.Date,
+  Error: JS_GLOBALS.Error,
+  TypeError: JS_GLOBALS.TypeError,
+  RangeError: JS_GLOBALS.RangeError,
+  // Formatting and collation only.
+  Intl: namespace({
+    NumberFormat: callable(),
+    DateTimeFormat: callable(),
+    Collator: callable(),
+    PluralRules: callable(),
+    RelativeTimeFormat: callable(),
+    ListFormat: callable(),
+    Segmenter: callable(),
+  }),
+};
+
+/**
+ * Walk a namespace path through a registry. Returns the leaf entry if
+ * the full chain resolves, otherwise null. Defaults to `JS_GLOBALS`;
+ * pass `SANDBOX_JS_GLOBALS` for the `--agency-only` allowlist.
  *
  * Examples:
  *   lookupJsMember(["JSON", "parse"])     → { kind: "callable", sig: undefined }
  *   lookupJsMember(["JSON", "banana"])    → null
  *   lookupJsMember(["NotAGlobal", "x"])   → null
  */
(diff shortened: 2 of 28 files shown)


packages/agency-lang/lib/typeChecker/sandboxNameCheck.ts
```diff
@@ -0,0 +1,200 @@
...
+  importedNodeNames: readonly string[],
+): void {
+  if (node.type === "variableName") {
+    const resolution = resolveVariable(node.value, {
+      functionDefs: ctx.functionDefs,
+      nodeDefs: ctx.nodeDefs,
+      importedFunctions: ctx.importedFunctions,
+      importedNodeNames,
+      jsImportedNames: ctx.jsImportedNames,
+      scopeHas: () => false,
```
egonSchiele: **Blocking false positive.** `scopeHas: () => false` here means a declaration-hanging expression can only resolve `def`/`node`/imports — never a top-level `const`/`let`. But tag arguments and default values legitimately reference module constants. Reproduced on this branch:

```
const MAXLEN = 5
@jsonSchema({ maxLength: MAXLEN })
def f(x: string): string { return x }
node main() { print("hi") }
```

Runs fine without the flag; under `--agency-only` it fails with `AG4007: Variable 'MAXLEN' is not defined`. A user can't work around it without inlining the constant.

The general `checkUndefinedVariables` pass resolves such names through the real scope (`scope.has(name)`); this pass needs the same. Pass the module/top-level scope here (and, for a default value, the enclosing function's earlier params) rather than a constant `false`. Please also add a positive control — a top-level `const` used inside a tag argument — since the current `@validate(isPositive)` control only exercises the `functionDefs` path and so stays green despite this bug.

packages/agency-lang/lib/typeChecker/resolveCall.ts
```diff
@@ -284,18 +284,92 @@ export const JS_GLOBALS: Record<string, JsRegistryEntry> = {
...
+  Array: JS_GLOBALS.Array,
+  // Object minus the prototype/descriptor members (create, getPrototypeOf,
+  // setPrototypeOf, getOwnPropertyNames stay out).
+  Object: namespace({
+    keys: callable({ params: [ANY_T], returnType: stringArray }),
+    values: callable({ params: [ANY_T], returnType: anyArray }),
+    entries: callable({ params: [ANY_T], returnType: anyArray }),
+    fromEntries: callable({ params: [anyArray], returnType: ANY_T }),
+    assign: callable(),
+    freeze: callable({ params: [ANY_T], returnType: ANY_T }),
```
egonSchiele, line 334: Minor: the spec's allowlist table lists `Object.hasOwn`, but it's omitted from this `Object` namespace (only `keys`/`values`/`entries`/`fromEntries`/`assign`/`freeze`). `Object.hasOwn(o, k)` is a pure own-property predicate and safe. Either add it here or drop it from the spec/dev-note table so the two stay in step. Not a safety issue — omission is stricter — just a divergence from the documented allowlist.

packages/agency-lang/lib/typeChecker/sandboxNameCheck.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+  }
+}
+
+/** Refuse `new X()` whose class name is not an allowlisted global. The class
+ *  name is a grammar literal (no `new (x + y)()`), so this is complete. */
+function checkNewCallee(node: AgencyNode, ctx: TypeCheckerContext): void {
+  if (node.type !== "newExpression") return;
+  const className = (node as NewExpression).className;
+  if (Object.prototype.hasOwnProperty.call(SANDBOX_JS_GLOBALS, className)) return;
```
egonSchiele, line 91: Minor: this accepts any `className` that is a key in the registry, whether or not the entry is constructible. So `new Math()`, `new JSON()`, `new Intl()` pass the check and then throw a confusing runtime `TypeError` instead of getting a clear compile-time diagnostic. Not a safety hole (the dangerous constructors aren't keys, and non-constructors throw), but you could tighten it to require a `callable` entry for a better error. Optional.

egonSchiele (commented):
Reviewed the implementation against the spec and plan, and exercised it on a build of this branch. The structure is clean: one registry-parametric resolver shared across the general passes and the new `checkSandboxNames`, the two decl-hanging positions (tags + array/object defaults) ride the same resolver, and the body-reachable vs decl-hanging walks are disjoint so there is no double-reporting (verified: `k.constructor.constructor` yields exactly one AG4011). The template early-return is handled correctly — under the sandbox it skips only the hole names, not the whole file. `SANDBOX_JS_GLOBALS` is a faithful subset of the spec's allowlist, and the tests assert real refusals with benign payloads (good — a regressed refusal can't do harm). The layer-1/defense-in-depth framing in the PR body and dev note is accurate.

One blocking correctness issue and two minor notes below.

**Blocking — false positive: a top-level `const` (or `let`) referenced inside a tag argument is wrongly refused.** `checkDeclExpressionName` passes `scopeHas: () => false`, so module-level bindings that are not `def`/`node`/imports are invisible to it. Reproduced on this branch:

```
const MAXLEN = 5
@jsonSchema({ maxLength: MAXLEN })
def f(x: string): string { return x }
node main() { print("hi") }
```

This compiles and runs fine normally, but `agency run --agency-only` rejects it with `AG4007: Variable 'MAXLEN' is not defined`. Same for a name nested in an array/object default (`def f(xs = [MAXLEN])`) — though that particular pattern is also broken at runtime on `main` by an unrelated bare-identifier codegen bug, so the tag case is the clean regression. No test catches this because the only tag positive control (`@validate(isPositive)`) names a local `def`, which resolves via `functionDefs`. Fix: feed the module/top-level scope (and, for a default value, the enclosing function's earlier params) into the resolver instead of `() => false`, and add a positive control with a top-level const in a tag argument. Details inline.

Nice work overall — with the scope fix this is in good shape.

packages/agency-lang/lib/typeChecker/sandboxNameCheck.ts
egonSchiele: Fixed in 54cd4a0b8. The root cause was that at type-check time the tag is still a standalone node before its declaration (the typechecker reads tags via `tagsAbove`, before the preprocessor attaches them), so my `declarationScope` never fired. Now: `tagsAbove` pairs each tag with its owner, and each declaration-hanging expression carries module-level bindings plus the owner's value parameters / function parameters. Your `const MAXLEN` repro now compiles; added it and a value-parameter case as positive controls.

packages/agency-lang/lib/typeChecker/resolveCall.ts
egonSchiele: Added `Object.hasOwn` to SANDBOX_JS_GLOBALS (pure own-property predicate) so the allowlist matches the documented table. 54cd4a0b8.

packages/agency-lang/lib/typeChecker/sandboxNameCheck.ts
egonSchiele: Left as-is deliberately. Tightening to require a constructible entry is awkward because `Date` is modeled as a namespace (no callableSig) yet is constructible, so "require callable" would reject `new Date()`. The dangerous constructors aren't registry keys so they're already refused; `new Math()` / `new JSON()` are nonsensical code that throws a runtime TypeError. Not a safety gap — happy to add a small constructible-set check later if the runtime error bothers users.

packages/agency-lang/lib/typeChecker/sandboxNameCheck.ts
```diff
@@ -0,0 +1,200 @@
...
+    if (
+      record.defaultValue !== undefined &&
+      record.defaultValue !== null &&
+      typeof record.defaultValue === "object"
+    ) {
+      const dv = record.defaultValue as Record<string, unknown>;
+      if (dv.type === "agencyArray" || dv.type === "agencyObject") {
+        found.push(record.defaultValue as AgencyNode);
+      }
+    }
```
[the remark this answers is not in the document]
egonSchiele: Good catch — fixed in 54cd4a0b8. The collector now gathers every AST-valued default, not just array/object, so `x = "${process.env.HOME}"` is refused (walkNodes descends into the interpolation). Added a refusal test.

packages/agency-lang/lib/typeChecker/sandboxNameCheck.ts
```diff
@@ -0,0 +1,200 @@
...
+): void {
+  if (node.type === "variableName") {
+    const resolution = resolveVariable(node.value, {
+      functionDefs: ctx.functionDefs,
+      nodeDefs: ctx.nodeDefs,
+      importedFunctions: ctx.importedFunctions,
+      importedNodeNames,
+      jsImportedNames: ctx.jsImportedNames,
+      scopeHas: () => false,
+      registry: SANDBOX_JS_GLOBALS,
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 54cd4a0b8 — value parameters of the owning type alias are now in scope for its tag arguments (`type Bounded(low, high)` with `@jsonSchema({ minimum: low, maximum: high })` compiles). Added a positive control.

packages/agency-lang/docs/dev/cli/test-cli-sandbox.md
```diff
@@ -17,8 +17,15 @@ under each name it is given.
...
-  There is no other way to touch the world from pure Agency.
+- With `--agency-only`, every effect the code can perform is an interrupt,
+  and the JS-globals bind-check refuses the compiled-to-JavaScript escapes
+  that would otherwise bypass interrupts entirely (`process`, `fetch`,
+  `eval`, `new Function`, the `.constructor` walk, tag arguments, default
+  values). See `docs/dev/compiler/agency-only-bound-names.md`. That check is
+  defense in depth, not the whole boundary: a runtime-computed property key
+  gets through it, so the backstop for code-from-strings is layer 2 in
+  `docs/dev/security/roadmap.md` (A1). There is no other way to touch the
+  world from pure Agency once those hold.
```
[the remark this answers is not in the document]
egonSchiele: Softened in 54cd4a0b8: the safety-argument section now states plainly that these flags are not yet a complete sandbox and points at the roadmap.

packages/agency-lang/lib/typeChecker/sandboxNameCheck.ts
```diff
@@ -0,0 +1,277 @@
...
+  nodes: AgencyNode[],
+  moduleScope: string[],
+): DeclHangingExpression[] {
+  const found: DeclHangingExpression[] = [];
+
+  // (1) Statement-level tags. At type-check time the preprocessor has not
+  // attached them yet, so a `@tag(...)` is a standalone node before the
+  // declaration it annotates. `tagsAbove` pairs each with its owner at every
+  // body level, so a tag argument sees the owner's value parameters / params.
+  const handled = new Set<AgencyNode>();
```
egonSchiele, line 199: Minor convention: `new Set<AgencyNode>()` — CLAUDE.md ("Use arrays instead of sets") asks for arrays here, and `lib/structure` lint may flag it. The dedup set holds a handful of tag nodes, so an array with `.includes` (reference equality, and n is tiny) is equivalent and matches the convention.

egonSchiele (commented):
Re-reviewed after the two fix commits. The blocking false positive is resolved, and the fix goes further than I asked: `checkDeclExpressionName` now resolves against the module scope (top-level `const`/`let`) plus the owning declaration's parameters — function/node params and, nicely, a type alias's value parameters — and interpolated-string defaults are now caught too. I built the branch and verified:

- `@jsonSchema({ maxLength: MAXLEN })` with `const MAXLEN = 5` now compiles under `--agency-only` (was AG4007).
- The escapes are all still refused: `process` in a tag argument and in an array default (AG4007), and `new Function(...)` (AG4004).
- `Object.hasOwn` is added to the allowlist, matching the spec table.
- New positive controls cover the top-level-const and type-alias-value-param cases; a new refusal test covers the interpolated-string default. All the new bound-names tests pass locally.

Caveat on my local run: I built with `pnpm run build` (no stdlib), so the one pre-existing test that needs stdlib (`positive control: trusted compileSource runs the generator`) failed with a `stdlib/index.js` module-not-found — that is my partial build, not the change (the test is unmodified by this PR, and every new test passes). CI's full build covers it.

One new minor below; otherwise this looks good to me.

Issue #974, opened 2026-08-30
Make --agency-only JS-globals enforcement transitive across forks (std::agency.run)


egonSchiele:
## Summary

Under `--agency-only`, the JS-globals bind-check (layer 1, #971 / PR #973) is enforced on the entry program but **not** on code that program compiles and forks at runtime via `std::agency.run` / `compile`. So an untrusted agency-only program can construct a child source that names a host global (`process`, `eval`, ...) and run it in a subprocess with no bind-check.

The fix is to make enforcement **transitive across a fork**: a fork inherits its parent's trust. This is intentionally not yet done — closing it cleanly also requires fixing a typechecker scope hole (below), so it was split out of #973 to keep that PR to a CI-green, entry-only enforcement.

## Why the gap exists

`compileValidatedClosure` is shared by two callers:

- `compileAgencyOnly` (the `--agency-only` entry point) — sets `enforceJsGlobals: true`.
- `compileToProgram` in `lib/stdlib/agency.ts` (the runtime fork path behind `fork` / `std::agency.run`) — does not, and cannot today, because `agencyOnly` is a **compile-time** parameter (`lib/cli/commands.ts:277`, `compileMode: { agencyOnly }`). Nothing carries it into the running process, so the fork-compile call site has no way to know it is inside an agency-only run.

### 1. Carry `agencyOnly` into the run, like policy and budget

`--agency-only` runs already spawn the compiled `.js` with `withRootCarriers(process.env, { policy, budget })` (`lib/cli/commands.ts:302`). Add `agencyOnly` as a third root carrier:

- New env constant (e.g. `AGENCY_ONLY`) added to `RootCarriers` in `lib/cli/childEnv.ts`, set at the `--agency-only` run path.
- `compileToProgram` (`lib/stdlib/agency.ts:100`) reads that carrier and passes `enforceJsGlobals` through to `compileSandboxed`.

Transitivity is then free: `buildForkOptions` (`lib/runtime/ipc.ts:678`) forks with `env: { ...process.env, ... }`, so every descendant fork inherits the carrier — a child-of-child of an agency-only run stays enforced. This mirrors exactly how run-policy already propagates down the fork tree.

### 2. Fix the `fork ... as` block-binding scope hole first (the real blocker)

Turning enforcement back on for forked source re-enables `undefinedVariables: "error"` on that source, which currently **false-positives** on a `let` declared inside a `fork` block. Concretely, this benign child source is refused today with:

```
[...]

PR #975, opened 2026-08-30
Disable code-generation-from-strings for --agency-only runs (layer 2)

egonSchiele:
## What

`agency run --agency-only` now spawns the child Node process with `--disallow-code-generation-from-strings`. This is **layer 2** of the JS-globals containment for untrusted Agency code (#971); layer 1 (the compile-time bind-check) shipped in #973.

## Why

Layer 1 refuses `.constructor` / `prototype` / `__proto__` only when they are spelled out or written as a string-literal computed key. A **runtime-computed** key that reaches `Function` slips past any syntactic check:

```agency
node main() {
  const key = "cons" + "tructor"
  const F = ({})[key][key]              // ({}).constructor.constructor === Function
  return F("return process.env.HOME")() // Function(body) — reads a host global
}
```

`--disallow-code-generation-from-strings` makes `eval` and the `Function` constructor throw `EvalError` at runtime, closing this route. It also backstops `eval` / `new Function` themselves (layer 1 already refuses those by name; this catches the computed forms).

## Regression guards (in CI)

- `codegen-walk.agency` in `tests/agency-js/test-cli-agency-only/` runs the computed-key escape under `--agency-only` and asserts `EvalError` appears — this fails if the flag falls off the spawn args.
- `bound-good.agency` in the same harness runs a benign program under `run --agency-only`, so the "a normal program still runs under the flag" check runs in CI too.

## Verified end-to-end through the CLI

- The escape above is **refused** under `--agency-only` (crashes with `EvalError: Code generation from strings disallowed`).
- A benign agency-only program still **runs** under the flag — the startup imports (`agency-lang`, `smoltalk`, `zod`) do not use code-generation-from-strings.
- A non-`--agency-only` run is **unaffected** (the flag is scoped to untrusted runs via `sandboxRuntimeNodeArgs(agencyOnly)`).

## Scope / follow-ups

- Applied to the `agency run --agency-only` entry only. **Not** yet applied across forks (`std::agency.run` children) or the `agency test --agency-only` grader path; both ride with #974 (make enforcement transitive across forks — carrier + the `fork ... as` scope fix).
- Layer 3 (freeze intrinsics / SES) remains open; see `docs/dev/security/roadmap.md` A1.

Related: #971 (the hole), #973 (layer 1), #974 (fork transitivity).

https://claude.ai/code/session_01SwUwoL2LVZp8KmCS3GZHjS

### DIFF
--- a/packages/agency-lang/docs/dev/security/roadmap.md
+++ b/packages/agency-lang/docs/dev/security/roadmap.md
@@ -15,8 +15,19 @@ break. Within a group they are in the order to do them.
 **Layer 1 (the bind-check) DONE** — `jsGlobals: "sandbox"`, `SANDBOX_JS_GLOBALS`,
 and `checkSandboxNames` refuse free identifiers, `new` callees, the
 constructor/prototype walk (best-effort), tag arguments, and default values;
-see `docs/dev/compiler/agency-only-bound-names.md`. Layers 2 (disable
-code-generation-from-strings) and 3 (freeze intrinsics) still open.
+see `docs/dev/compiler/agency-only-bound-names.md`.
+
+**Layer 2 (disable code-generation-from-strings) DONE** — `agency run
+--agency-only` spawns the child Node process with
+`--disallow-code-generation-from-strings` (`sandboxRuntimeNodeArgs` in
+`lib/cli/commands.ts`). This is the runtime backstop for the constructor walk
+layer 1 catches only best-effort: a runtime-computed key that reaches
+`Function` (`m[a + b]["constructor"](...)`) now throws `EvalError`. Verified a
+benign agency-only program still runs under the flag. Not yet applied across
+forks (`std::agency.run` children) or to the `agency test --agency-only`
+grader path — those ride with #974 (make enforcement transitive across forks).
+
+Layer 3 (freeze intrinsics) still open.
 
 <https://github.com/egonSchiele/agency-lang/issues/971>. An identifier the compiler does not know is emitted verbatim, so
 `process.env.HOME`, `process.getBuiltinModule("fs")`, `fetch(...)`,
@@ -49,7 +60,7 @@ Fix, in three layers, cheapest first:
    `docs/superpowers/specs/2026-08-29-agency-only-bound-names-design.md`
    (revised after review).
 2. **`--disallow-code-generation-from-strings`** on the child Node process
-   for `--agency-only` runs (`compiledOutputNodeArgs` in
+   for `--agency-only` runs (shipped — `sandboxRuntimeNodeArgs` in
    `lib/cli/commands.ts`). This is the REAL boundary for reaching `Function`
    and calling it with a string — via `eval`, `new Function`, or any
    constructor walk including `m[a + b]["constructor"]…` with a runtime-
--- a/packages/agency-lang/docs/dev/security/roadmap.md
+++ b/packages/agency-lang/docs/dev/security/roadmap.md
@@ -17,15 +17,20 @@ and `checkSandboxNames` refuse free identifiers, `new` callees, the
 constructor/prototype walk (best-effort), tag arguments, and default values;
 see `docs/dev/compiler/agency-only-bound-names.md`.
 
-**Layer 2 (disable code-generation-from-strings) DONE** — `agency run
---agency-only` spawns the child Node process with
+**Layer 2 (disable code-generation-from-strings) — shipped for `agency run`.**
+`agency run --agency-only` spawns the child Node process with
 `--disallow-code-generation-from-strings` (`sandboxRuntimeNodeArgs` in
-`lib/cli/commands.ts`). This is the runtime backstop for the constructor walk
-layer 1 catches only best-effort: a runtime-computed key that reaches
-`Function` (`m[a + b]["constructor"](...)`) now throws `EvalError`. Verified a
-benign agency-only program still runs under the flag. Not yet applied across
-forks (`std::agency.run` children) or to the `agency test --agency-only`
-grader path — those ride with #974 (make enforcement transitive across forks).
+`lib/cli/commands.ts`), so `eval`, the `Function` constructor, and a
+constructor walk through a runtime-computed key (`m[a + b]` where `a + b` is
+`"constructor"`, which layer 1's syntactic check cannot see) all throw
+`EvalError`. Forks (`std::agency.run` children, whose `execArgv` is set
+explicitly in `buildForkOptions`, `lib/runtime/ipc.ts`) and the `agency test
+--agency-only` grader path do not carry the flag yet; both follow in #974.
+
+The flag is not a full ban on generating code: `vm.runInThisContext` and
+`vm.Script` still run under it. Reaching `vm` needs `require` or
+`getBuiltinModule`, which layer 1 refuses by name, so there is no path today —
+but layer 3 should not assume the flag alone closes every route to new code.
 
 Layer 3 (freeze intrinsics) still open.
 
@@ -60,8 +65,9 @@ Fix, in three layers, cheapest first:
    `docs/superpowers/specs/2026-08-29-agency-only-bound-names-design.md`
    (revised after review).
 2. **`--disallow-code-generation-from-strings`** on the child Node process
-   for `--agency-only` runs (shipped — `sandboxRuntimeNodeArgs` in
-   `lib/cli/commands.ts`). This is the REAL boundary for reaching `Function`
+   for `--agency-only` runs (shipped for `agency run`; forks and `agency test`
+   in #974 — `sandboxRuntimeNodeArgs` in `lib/cli/commands.ts`). This is the
+   REAL boundary for reaching `Function`
    and calling it with a string — via `eval`, `new Function`, or any
    constructor walk including `m[a + b]["constructor"]…` with a runtime-
    computed key, which layer 1's syntactic property check cannot catch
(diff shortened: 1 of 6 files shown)


packages/agency-lang/lib/cli/run.spawn.test.ts
```diff
@@ -101,3 +102,12 @@ describe("runChildOverrides", () => {
     expect(merged.log?.code?.entry).toBe("main.agency");
   });
 });
+
+describe("sandboxRuntimeNodeArgs", () => {
```
egonSchiele: This test cannot fail for the reason that matters: it checks that a pure function returns a constant. It would still pass if the flag were misspelled, if `spawn` never received it, or if V8 ignored it. The end-to-end check in the PR body was done by hand, so nothing in CI pins the actual behaviour.

The harness for a real check already exists: `tests/agency-js/test-cli-agency-only/test.js` runs `--agency-only` fixtures through the built CLI. Suggest adding a `codegen-walk.agency` fixture with the computed-key escape from the PR body, run with `--reject "*"` like the `bound-*` fixtures (harmless on failure), asserting exit 1 and `EvalError` in the output. That is the one test that fails if the flag falls off the spawn args.

(Side note: the benign-program check is already covered by CI once this merges, since `bound-good.agency` in the same file runs under `run --agency-only`. Worth a sentence in the PR body so a future reader knows where the regression guard lives.)

packages/agency-lang/docs/dev/security/roadmap.md
```diff
@@ -15,8 +15,19 @@ break. Within a group they are in the order to do them.
 **Layer 1 (the bind-check) DONE** — `jsGlobals: "sandbox"`, `SANDBOX_JS_GLOBALS`,
 and `checkSandboxNames` refuse free identifiers, `new` callees, the
 constructor/prototype walk (best-effort), tag arguments, and default values;
-see `docs/dev/compiler/agency-only-bound-names.md`. Layers 2 (disable
-code-generation-from-strings) and 3 (freeze intrinsics) still open.
+see `docs/dev/compiler/agency-only-bound-names.md`.
+
+**Layer 2 (disable code-generation-from-strings) DONE** — `agency run
```
egonSchiele: "DONE" overstates it: this same paragraph says forks and the `agency test` grader path are not covered, and `buildForkOptions` (`lib/runtime/ipc.ts:678`) sets `execArgv` explicitly, which drops the flag for every `std::agency.run` child. Suggest "Shipped for `agency run`; forks and `agency test` follow in #974".

Two facts for #974 and layer 3, checked by running Node directly with the flag:
- V8 flags do appear in `process.execArgv`, so forwarding this one into `buildForkOptions` is a one-liner.
- The flag does not cover `vm.runInThisContext` or `vm.Script` (both ran fine under it). Reaching `vm` needs `require` / `process.getBuiltinModule`, which layer 1 refuses by name, so I see no path today, but a line under layer 3 would stop someone assuming this is a full codegen ban.

packages/agency-lang/lib/cli/commands.ts
```diff
@@ -49,6 +49,16 @@ export function compiledOutputNodeArgs(): string[] {
   return [`--import=${compiledOutputRegisterUrl()}`];
 }
 
+/** Extra V8 flags for running UNTRUSTED compiled output (`--agency-only`).
+ *  `--disallow-code-generation-from-strings` makes `eval` and the `Function`
+ *  constructor throw at runtime, closing the one route the compile-time
+ *  bind-check cannot see: a runtime-computed property key that reaches
+ *  `Function` (`m[a + b]["constructor"](...)`). Layer 2 of the JS-globals
+ *  containment; see docs/dev/security/roadmap.md A1. Empty for trusted runs. */
+export function sandboxRuntimeNodeArgs(agencyOnly: boolean): string[] {
```
egonSchiele, line 58: Confirmed the flag does what the comment says: under `node --disallow-code-generation-from-strings`, `Function("...")`, `eval("...")`, and the computed walk `({})[k][k]("...")` all throw `EvalError`. Grepped `zod`, `picomatch`, and `nanoid` for `new Function` / `eval` as well: none, so startup is safe.

egonSchiele (commented):
The change is correct and small: the flag is scoped to untrusted runs via `compileMode.agencyOnly`, CI is green, and I verified the flag blocks `Function`, `eval`, and the computed constructor walk by running Node directly. One request before merge: a CI test that actually exercises the flag (inline comment on the test file), since the current unit test passes whether or not the flag reaches the child. Also a wording fix in the roadmap.

packages/agency-lang/docs/dev/security/roadmap.md
egonSchiele: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/lib/cli/run.spawn.test.ts
egonSchiele: Done. Replaced the constant-returning unit test with a fixture (`codegen-walk.agency`) run through the built CLI: it reaches Function through a runtime-computed key (compiles under --agency-only, since layer 1 can't see it) and the harness asserts EvalError appears at runtime. Proven to discriminate — without the flag the call succeeds and no EvalError prints, so `codegenWalk.blocked` flips to false and the fixture mismatches. Switched the harness helper to spawnSync so it captures stderr and the exit code even on exit 0 (this crash logs to stderr and exits 0). Added a line to the PR body noting bound-good.agency is the benign-run guard.

packages/agency-lang/docs/dev/security/roadmap.md
egonSchiele: Reworded to "shipped for `agency run`", with forks (execArgv set explicitly in buildForkOptions) and `agency test --agency-only` called out as following in #974, and a note that this is not a full codegen ban (vm.runInThisContext / vm.Script still run under it; reaching vm needs require / getBuiltinModule, which layer 1 refuses by name). Recorded your execArgv one-liner fact on #974.

egonSchiele: Went through the roadmap edit and the comments I added (commands.ts, test.js, the fixture). Cut the thought-process/verification notes (e.g. "verified a benign program still runs" — the fixture is the guard, not a doc fact) and kept only what a future reader needs: what the flag does, what it doesn't cover (forks, vm), and where it's wired. The remaining comments explain *why* (why spawnSync, why the key must be runtime-computed), not narrate the change.

packages/agency-lang/lib/cli/commands.ts
```diff
@@ -49,6 +49,16 @@ export function compiledOutputNodeArgs(): string[] {
   return [`--import=${compiledOutputRegisterUrl()}`];
 }
 
+/** Extra V8 flags for running UNTRUSTED compiled output (`--agency-only`).
+ *  `--disallow-code-generation-from-strings` makes `eval` and the `Function`
+ *  constructor throw at runtime, closing the one route the compile-time
+ *  bind-check cannot see: a runtime-computed property key that reaches
+ *  `Function` (`m[a + b]["constructor"](...)`). Layer 2 of the JS-globals
+ *  containment; see docs/dev/security/roadmap.md A1. Empty for trusted runs. */
```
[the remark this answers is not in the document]
egonSchiele: Fixed. The example now shows a runtime-computed key (`m[a + b]` where `a + b` is "constructor"); a string-literal `["constructor"]` is caught by layer 1, so it was the wrong illustration.

egonSchiele (commented):
Re-reviewed 69d958b48. All three earlier comments are addressed:

- The constant-returning unit test is gone; `codegen-walk.agency` runs through the built CLI and asserts both that it compiles (layer 1 does not see the computed key) and that `EvalError` appears. I ran the fixture locally both ways: without `--agency-only` it compiles and runs with no `EvalError` in the output; with it, `EvalError: Code generation from strings disallowed`. So the test fails exactly when the flag falls off the spawn args.
- Roadmap now says "shipped for `agency run`", names `buildForkOptions` for the fork gap, and records that `vm` is not covered.
- The `m[a + b]` example in the comment and roadmap is now the runtime-computed form.

One new observation, inline on the test, about the exit code. It is pre-existing CLI behaviour, not this PR, so no change requested here. Otherwise good to merge from my side.

packages/agency-lang/tests/agency-js/test-cli-agency-only/test.js
```diff
@@ -66,6 +73,10 @@ writeFileSync(
       writeRestricted: { exitCode: writeRestricted.exitCode, fileWritten: writeRestrictedFile },
       boundRefused,
       boundGood: { exitCode: boundGood.exitCode },
+      codegenWalk: {
+        compiled: !codegenWalk.output.includes("compile refused"),
+        blocked: codegenWalk.output.includes("EvalError"),
```
egonSchiele, line 78: Matching on the output rather than the exit code is the right call, and the comment above explains why, but the reason itself is worth an issue: `agency run` exits 0 when the program throws an uncaught runtime error. I checked with a plain program (`xs[5].foo` on an empty array): stack trace to stderr, exit 0. Same for this `EvalError`. For a served or scripted caller that means a sandbox refusal looks like success unless they parse stderr. Not this PR's problem, but it would be good to file so #974 or a later layer can make a refused run exit non-zero, at which point this check can also assert `exitCode !== 0`.

PR #977, opened 2026-08-30
Stop the checkpoint schema from dropping guards, cost, and drafts on reload

egonSchiele:
## The bug

A checkpoint is validated against the zod schemas in `lib/runtime/state/schemas.ts` when it is read back from JSON (`Checkpoint.fromJSON`). zod strips any key a schema does not name. The schemas had drifted from the `toJSON` methods and were missing a set of load-bearing fields:

- **`StateStackJSON`**: `localCost`, `localTokens`, `seedCost`, `seedTokens`, `guards`, `inheritedGuardCount`, `inheritedTimeGuards`
- **`StateJSON`**: `scopedCallbacks`, `savedDraft`
- **`BranchStateJSON`**: `result`, `globalsJSON`, `activeStack`

`StateStack.fromJSON` / `State.fromJSON` read every one of these back, so a checkpoint that is stringified and reparsed lost its guards (cost/time budgets), cost and token accounting, saved drafts, scoped callbacks, and per-branch state.

## Blast radius

Only the **external** resume path is affected — the one where a checkpoint is serialized to JSON and reparsed: the HTTP `/resume` round trip when serving an agent, and any checkpoint written to disk and loaded back. The in-process resume path deep-clones the live `Checkpoint` object and never passes through the schema, so it was never losing these fields. The most serious consequence is a served agent losing its cost/time guards across a resume — the budget ceiling would silently reset.

## The fix

Add the missing fields to the three schemas, plus a new `guardJSONSchema` (a discriminated union on `kind` matching the `CostGuard` and `TimeGuard` JSON shapes). This is exactly the "keep the schema in step with `toJSON`" discipline the checkpointing dev doc already called out for thread fields.

## Tests

`lib/runtime/state/schemas.test.ts` (new) pins the round trip:

- Each schema keeps every previously-stripped field.
- A full `Checkpoint.fromJSON(JSON.parse(JSON.stringify(...)))` preserves guards, cost, `savedDraft`, and per-branch `result`/`activeStack`.
- An end-to-end `StateStack` with a real charged `CostGuard` survives `toJSON → Checkpoint.fromJSON → StateStack.fromJSON` with its `spent` intact.

I confirmed all 7 tests **fail** against the pre-fix schema and pass with the fix. The neighboring `checkpointStore.test.ts` and `stateStack.test.ts` (104 tests) still pass. `pnpm run typecheck` (all three configs), prettier, and `lint:structure` are green.

The checkpointing dev doc note is broadened to cover every `toJSON` in the checkpoint tree, not just the thread ones.

### DIFF
--- a/packages/agency-lang/docs/dev/runtime/checkpointing.md
+++ b/packages/agency-lang/docs/dev/runtime/checkpointing.md
@@ -237,8 +237,18 @@ node main() {
 `restore()` accepts the parsed JSON directly and revives it with
 `Checkpoint.fromJSON`, which validates against the zod schemas in
 `lib/runtime/state/schemas.ts`. zod drops any key a schema does not name, so
-when adding a field to `MessageThread.toJSON` or `ThreadStore.toJSON`, add it
-to the schema too, or a checkpoint read back from a file will lose it.
+every field written by a `toJSON` in the checkpoint tree must be named in the
+matching schema, or a checkpoint read back from a file will silently lose it.
+This applies to `StateStack.toJSON` (`localCost`, `guards`,
+`inheritedTimeGuards`, and so on), `State.toJSON` (`scopedCallbacks`,
+`savedDraft`), `BranchState` (`result`, `globalsJSON`, `activeStack`), the guard
+`toJSON` shapes, and `MessageThread.toJSON` / `ThreadStore.toJSON`. When you add
+a field to any of these, add it to its schema in the same change.
+`schemas.test.ts` pins this: it fails if a schema strips a field the
+deserializer reads back. Note this only bites the external path — the in-process
+resume deep-clones the live `Checkpoint` object, so it never loses these fields;
+only a checkpoint that is stringified and reparsed (HTTP `/resume`, a checkpoint
+loaded from disk) goes through the schema.
 
 ```agency
 node main() {
--- a/packages/agency-lang/docs/dev/runtime/checkpointing.md
+++ b/packages/agency-lang/docs/dev/runtime/checkpointing.md
@@ -237,18 +237,12 @@ node main() {
 `restore()` accepts the parsed JSON directly and revives it with
 `Checkpoint.fromJSON`, which validates against the zod schemas in
 `lib/runtime/state/schemas.ts`. zod drops any key a schema does not name, so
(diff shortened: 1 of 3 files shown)


packages/agency-lang/docs/dev/runtime/checkpointing.md
```diff
@@ -237,8 +237,18 @@ node main() {
 `restore()` accepts the parsed JSON directly and revives it with
 `Checkpoint.fromJSON`, which validates against the zod schemas in
 `lib/runtime/state/schemas.ts`. zod drops any key a schema does not name, so
-when adding a field to `MessageThread.toJSON` or `ThreadStore.toJSON`, add it
-to the schema too, or a checkpoint read back from a file will lose it.
+every field written by a `toJSON` in the checkpoint tree must be named in the
```
egonSchiele: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

egonSchiele: Good call. I went back through the doc note and every comment I added. The note had a full field enumeration plus a line describing what the test does — that was restating the diff, so I cut it. It is now just the rule (keep the schema in step with any toJSON in the tree) and the one non-obvious fact: only the stringify-and-reparse path hits the schema, so in-process resume was never affected.

In the test file I cut the header down to the single warning worth keeping (a plain StateStack round-trip skips zod, so do not collapse these tests into one) and dropped a redundant inline comment. Kept the 'spent is private' note since that choice is not obvious.

packages/agency-lang/lib/runtime/state/schemas.test.ts
```diff
@@ -0,0 +1,199 @@
+import { describe, it, expect } from "vitest";
+import {
+  checkpointSchema,
+  stateStackJSONSchema,
+  stateJSONSchema,
+  branchStateJSONSchema,
+  guardJSONSchema,
+} from "./schemas.js";
```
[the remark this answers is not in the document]
egonSchiele: Fixed — removed the unused checkpointSchema import. The round-trip tests exercise it through Checkpoint.fromJSON instead.

Issue #966, opened 2026-08-30
Top-level 'with approve' runs before the root policy handler exists, bypassing --reject


egonSchiele:
## What happens

Top-level code (global initializers and imported modules' top-level statements) runs before the root policy handler exists, so a top-level `with approve` is the only handler in the chain and wins even under `--reject '*'`.

`lib/runtime/node.ts` bootstraps in this order: `__initAllRegistered` → `initializeGlobals` → `__initAllRegisteredCallbacks` → `installRunPolicyHandler` → `installRootBudget`. The comment there says so: "Global-init and top-level-callback code above runs before these exist". A top-level `with approve` pushes its handler directly on the context (`docs/dev/language/with-approve.md`, "In global scope"), so at that moment the chain is just the author's approve.

## Reproduction

```
// top.agency
const prompt = read("./prompt.md") with approve

node main() {
  printJSON(prompt)
  const again = read("./prompt.md") with approve
  printJSON(again)
}
```

```
agency run --reject '*' --agency-only top.agency
```

The top-level read returns `{ success: true, value: "<file contents>" }`. The in-node read returns `interrupt rejected`. Every effectful stdlib function is callable at top level, so this is a full policy bypass for anything an attacker is willing to do at init time. `docs/dev/cli/test-cli-sandbox.md` claims the opposite ("Code that runs before the handler exists ... cannot raise an interrupt at all"); that holds only for a top-level call without `with`.

## Fix direction

Install the root policy handler (and root budget) before any user code runs, including global init and imported modules' init. `installRunPolicyHandler` is a `pushHandler` on `execCtx`, which exists by then; the ordering looks like the only obstacle. Add an agency-js test that runs the file above under `--reject '*'` and asserts the top-level read is rejected. The resume path (`interrupts.ts`, `installRunPolicyHandler` on resume) should be checked for the same ordering.

PR #978, opened 2026-08-30
Host-supplied root interrupt policy for served invocations (InvocationOptions.policy)

egonSchiele:
## What

A serve host can attach a root interrupt policy to any invocation:

```ts
handler("POST", "/node/main", body, {
  policy: { "std::env": [{ match: { name: "MY_SECRET" }, action: "approve" }, { action: "reject" }] },
});
```

`resolveInvocation` validates it (invalid shape = a plain prefixed `Error` before any execution context; the adapter logs it and returns its generic error), and the runtime installs it as the **outermost handler** — in `initFreshExecCtx`, so one change covers `/node/*` and `/function`, and again in `respondToInterruptsCore`, so a raise made during a resume leg is governed too. An explicit policy replaces `AGENCY_RUN_POLICY` for the run. The resume answer path is untouched: decisions happen at the raise (see the CLAUDE.md interrupt-model section); caller responses still resolve by id.

Spec, plan, and reviews land under `docs/superpowers/` (the spec was revised twice; the earlier resume-response-rewrite design was cut as wrong).

## Why

Since 0.17.0, `env()` raises `std::env`, so a hosted agent reading an env variable surfaces an interrupt statelog cannot answer — and an agent could wrap the read in `handle { … } with approve` and read the host's own variables. With this change, statelog passes a per-project policy ("approve `std::env` iff `name` is one of this project's secrets") and a host reject beats the agent's own handlers (chain precedence `reject > propagate > approve`).

## Tests

- `tests/agency-js/serve-policy` — end to end through a compiled module's serve entry points (no existing serve test compiles a module; the adapter test file is spies only). Cases: approve returns the value; reject reads as `null` (`env()` absorbs its own denial); **reject beats the program's own approving handler** (the security-critical case); a `propagate` rule surfaces the interrupt over the program's approve ("always ask me"); a served function is governed identically; a raise during a resume leg is rejected by the re-installed policy, with a no-policy control where it surfaces instead.
- Unit: policy round-trip and validation in `invocationOptions.test.ts`; explicit-vs-env precedence with **disagreeing** policies, env fallback, IPC gate, and the no-policy pin in `runPolicyHandler.test.ts`; a pass-through case in `adapter.perInvocation.test.ts`.
- Guards run: `pnpm run typecheck` (3 configs), `pnpm run fmt:ts`, `pnpm run lint:structure`, the 5 touched vitest files (112 tests), and the agency-js fixture through the real runner. Full agency suite left to CI.

## Docs

- `docs/dev/hosting/how-hosted-serving-works.md`: new "Root interrupt policy" section; the "sent back unchanged" line now points at the checkpoint-integrity caveat.
- `docs/dev/runtime/interrupts.md`: cross-reference at the handler-verdicts section.
- `docs/dev/security/roadmap.md`: B2 rewritten — the old "caller can approve their own read" framing was wrong (echoed interrupt data is display-only); raise-time policy is shipped, and the open resume-leg item is checkpoint integrity (signing or host-side storage + replay protection).

## For the owner (docs/site, not edited here)

`docs/site/guide/policies.md` could use a short "Serve hosts" section: policy via `InvocationOptions`, installed at the root per invocation, decides at the raise (a reject on `std::env` reads as unset), `propagate` = always-ask, and the resume answer path is not a policy seam.

## Behavioral note found while testing

A policy reject on `std::env` does **not** fail the run: `env()` absorbs its own denial and returns `null`, so a refused variable reads as unset (`stdlib/system.agency`). The spec and plan were corrected to say so; for statelog this is the desired shape — secrets outside the project look like they were never set.

https://claude.ai/code/session_01LvXaShvdNDAwbePbicHwb1

### DIFF
--- a/packages/agency-lang/docs/dev/hosting/how-hosted-serving-works.md
+++ b/packages/agency-lang/docs/dev/hosting/how-hosted-serving-works.md
@@ -308,7 +308,14 @@ An invalid policy shape throws before any execution context exists; the
 adapter logs the message and returns its generic error, so a host bug is
 loud in the host's log and opaque to the caller. An explicit policy replaces
 the `AGENCY_RUN_POLICY` environment policy for that run; it does not merge
-with it. End-to-end behaviour is pinned by `tests/agency-js/serve-policy`.
+with it, and an empty `{}` is a no-op that still disables the environment
+policy. End-to-end behaviour is pinned by `tests/agency-js/serve-policy`.
+
+One known gap: the module's top-level initialization code runs before the
+handler is installed, so a raise inside a top-level initializer is not yet
+governed by the policy. That ordering problem predates this feature (it
+applies to `agency run --policy` too) and is tracked as B1 on
+`docs/dev/security/roadmap.md` (#966).
 
 ## Where the trace goes
 
--- a/packages/agency-lang/docs/dev/hosting/how-hosted-serving-works.md
+++ b/packages/agency-lang/docs/dev/hosting/how-hosted-serving-works.md
@@ -217,7 +217,8 @@ is why the same statelog server can answer the next request from any
 machine, any time later.
 
 The client answers by sending the interrupt objects back, unchanged, with
-one decision per interrupt:
+one decision per interrupt. (Nothing verifies the echo — see "What a
+checkpoint contains, and who can change it" below for what that means.)
 
 ```
 POST /serve/u_123/my-project/greeter.agency/resume
@@ -264,6 +265,50 @@ the developer's program:
 Until checkpoints are signed with a server key (or stored server-side and
 referred to by id), the interrupt loop should stay between statelog and a
 client the developer trusts.
+
+## Root interrupt policy
+
+A host can attach a policy to any invocation — the same JSON rule format
+`agency run --policy` takes (`lib/runtime/policy.ts`) — as a fourth argument
+to the serve handler:
+
+```ts
+handler("POST", "/node/main", body, {
+  policy: { "std::env": [{ match: { name: "MY_SECRET" }, action: "approve" }, { action: "reject" }] },
+});
+```
+
+The runtime installs it as the **outermost handler** for that invocation, in
+the same bootstrap that serves nodes and functions (`initFreshExecCtx`,
+`lib/runtime/node.ts`), and installs it again on every `/resume`
+(`respondToInterruptsCore`, `lib/runtime/interrupts.ts`), because handlers
+are never part of a checkpoint. Per invocation on purpose: a host like
+statelog serves many projects through one handler, and the rule for
+"which env reads are fine" differs per project.
+
+What that buys, in chain terms (`docs/dev/runtime/interrupts.md`, "Handler
+verdicts"):
+
+- A policy `reject` settles the interrupt **at the raise**, before it ever
+  becomes an HTTP response — and it wins over the program's own
+  `handle { … } with approve`, because a reject beats any approve in the
+  chain merge. The program sees an ordinary denial: `env()` reads the
+  variable as unset, a `read()` gets a failure.
+- A policy `approve` settles the raise silently; the caller never sees it.
+- A policy `propagate` forces the interrupt out to the caller even when the
+  program would have approved it itself — "always ask", per effect.
+- Effects the policy does not mention behave exactly as with no policy.
+- An interrupt raised **during** a resume leg goes through the re-installed
+  policy the same way. The caller's answers to already-surfaced interrupts
+  are not policy-checked: they resolve by interrupt id, the decision for
+  them was made at their raise, and the echoed interrupt data is display
+  information the resumed program never reads.
+
+An invalid policy shape throws before any execution context exists; the
+adapter logs the message and returns its generic error, so a host bug is
+loud in the host's log and opaque to the caller. An explicit policy replaces
+the `AGENCY_RUN_POLICY` environment policy for that run; it does not merge
+with it. End-to-end behaviour is pinned by `tests/agency-js/serve-policy`.
 
 ## Where the trace goes
 
--- a/packages/agency-lang/docs/superpowers/specs/2026-08-29-serve-host-interrupt-policy-design.md
+++ b/packages/agency-lang/docs/superpowers/specs/2026-08-29-serve-host-interrupt-policy-design.md
@@ -88,10 +88,13 @@ export type InvocationOptions = {
   config?: Partial<AgencyConfig>;
   traceId?: string;
   /** Root interrupt policy for this invocation, installed as the outermost
-   *  handler on the fresh run and on every resume leg. A `reject` here
-   *  beats any approval from the agent's own handlers or from the
-   *  resuming caller. Replaces (does not merge with) an AGENCY_RUN_POLICY
-   *  environment policy for this run. */
+   *  handler on the fresh run and re-installed on every resume leg. A
+   *  `reject` here beats any approval from the agent's own handlers, at the
+   *  raise. It does not re-judge the caller's answers to interrupts that
+   *  already surfaced (see §3 — there is nothing there to re-judge).
+   *  Replaces (does not merge with) an AGENCY_RUN_POLICY environment
+   *  policy for this run; note `{}` counts as a policy that abstains on
+   *  everything. */
   policy?: Policy;
 };
 ```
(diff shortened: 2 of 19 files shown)


packages/agency-lang/lib/runtime/invocationOptions.ts
```diff
@@ -13,6 +15,12 @@ import type { AgencyConfig } from "../config.js";
 export type InvocationOptions = {
   config?: Partial<AgencyConfig>;
   traceId?: string;
+  /** Root interrupt policy for this invocation, installed as the outermost
+   *  handler on the fresh run and re-installed on every resume leg. A
+   *  `reject` here beats any approval from the program's own handlers
+   *  (chain precedence: reject > propagate > approve). Replaces — does not
+   *  merge with — an `AGENCY_RUN_POLICY` environment policy for this run. */
+  policy?: Policy;
```
egonSchiele, line 27: Minor doc clarity: because `effective = policy ?? loadEnvPolicy()`, an explicit *empty* policy `{}` counts as "a policy" — it replaces the env policy and then abstains on every effect (no rules → `checkPolicyExplicit` returns null → the handler abstains). So a host passing `{}` disables its own `AGENCY_RUN_POLICY` and installs a no-op, which is stricter-sounding than it acts. Worth one sentence in this doc comment so a host doesn't reach for `{}` expecting "restrict everything".

packages/agency-lang/tests/agency-js/serve-policy/test.js
```diff
@@ -0,0 +1,118 @@
+// End-to-end for the serve-host root policy: a compiled module driven through
```
egonSchiele, line 1: Coverage note (non-blocking): the fixture doesn't exercise a `std::agency.run` subprocess reading env under a served policy. I verified the path is safe — the child runs in IPC mode so it doesn't install the policy, but its raise forwards to the parent (`interrupts.ts:472` `sendInterruptToParent`) and `mergeChainOutcomes` makes the host reject beat the child's local approve. Since that forwarding is the whole reason the host policy governs subprocess reads, and a regression there would be a real secret-exfiltration bypass, a fixture case (a node that `run`s a child which reads `SERVE_POLICY_SECRET`, rejected by the host policy) would be worth adding.

egonSchiele (commented):
Reviewed against the (twice-revised) spec and plan, and built the branch to run the new fixture rather than trust the description. It passes (`✓ Fixture match passed`), and the assertions are genuinely can-fail: `rejectedOverApprove: null` holds only because the host policy's reject beat the program's own `with approve` — without the policy that inner approve returns `"the-secret"`, not null. Case 6b is a proper control proving case 6's null comes from the re-installed policy, not from `env()` itself.

The implementation is small and matches the design exactly:
- `InvocationOptions.policy` validated in `resolveInvocation` (throws a prefixed `Error` before any exec context, for both fresh and resume); `installRunPolicyHandler(execCtx, policy?)` with `effective = policy ?? loadEnvPolicy()` so an explicit policy replaces the env policy and `loadEnvPolicy` stays private; threaded through `initFreshExecCtx.opts.policy` (covering `/node` and `/function` from one site) and re-installed at the resume site. No reorder, no response-path change — enforcement is entirely at the raise, as the corrected design requires.
- Both round-2 review findings are addressed: the tests are a real agency-js fixture driving a compiled module through the serve entry points (not the spy harness), and the propagate-beats-approve "always ask me" case is covered.
- Unit tests are can-fail: validation asserts the message names the schema problem; the precedence test uses **disagreeing** env-vs-explicit policies.
- Docs are honest: the roadmap B2 rewrite retracts the earlier wrong "caller can approve their own read" framing and correctly names checkpoint integrity as the open resume-leg item.

I also checked the one path the fixture doesn't cover — a `std::agency.run` subprocess reading env under a served policy. It is governed correctly: the child runs in IPC mode so it does not install the policy, but its raise forwards to the parent via `sendInterruptToParent` (`interrupts.ts:472`) and `mergeChainOutcomes` (reject > approve) makes the host's root reject beat the child's local approve. Sound and pre-existing — see the note below on covering it.

Two minor, non-blocking notes inline. Nice work — this is clean and well-tested.

(One doc nit for a future reader: the plan's Phase 1 still names an `InvalidInvocationPolicyError` class; the code uses a plain prefixed `Error` instead, which is the better call — it matches `loadEnvPolicy`'s failure and nothing catches by type. Worth aligning the plan text.)

packages/agency-lang/docs/dev/hosting/how-hosted-serving-works.md
```diff
@@ -217,7 +217,8 @@ is why the same statelog server can answer the next request from any
 machine, any time later.
 
 The client answers by sending the interrupt objects back, unchanged, with
-one decision per interrupt:
+one decision per interrupt. (Nothing verifies the echo — see "What a
```
egonSchiele, line 220: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/lib/runtime/node.ts
```diff
@@ -199,7 +204,7 @@ async function initFreshExecCtx(
   // functions) inherit them. (Global-init and top-level-callback code above
   // runs before these exist; that's unchanged — runNode installed them here
   // too, just after initFreshExecCtx returned.)
-  installRunPolicyHandler(execCtx);
+  installRunPolicyHandler(execCtx, opts.policy);
```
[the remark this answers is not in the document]
egonSchiele: Right that the gap exists, and it is #966 (roadmap B1) — it applies identically to agency run --policy, whose install site this reuses. Restructuring the bootstrap ordering is #966 scope, not this PR. What I changed here (4e4641b86): the hosting doc section now states the gap plainly instead of claiming every raise is governed, and roadmap B1 notes the per-invocation policy shares it.

packages/agency-lang/docs/superpowers/specs/2026-08-29-serve-host-interrupt-policy-design.md
```diff
@@ -0,0 +1,337 @@
...
+```ts
+// lib/runtime/invocationOptions.ts
+export type InvocationOptions = {
+  config?: Partial<AgencyConfig>;
+  traceId?: string;
+  /** Root interrupt policy for this invocation, installed as the outermost
+   *  handler on the fresh run and on every resume leg. A `reject` here
+   *  beats any approval from the agent's own handlers or from the
+   *  resuming caller. Replaces (does not merge with) an AGENCY_RUN_POLICY
+   *  environment policy for this run. */
```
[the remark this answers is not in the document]
egonSchiele: Correct — those were leftovers from the cut resume-rewrite design. Fixed in 4e4641b86: the spec §1 comment now says the policy decides at the raise and does not re-judge answers to already-surfaced interrupts, and §4 no longer claims a policy rejection is a failed run (env() reads as unset; only propagating callsites fail).

packages/agency-lang/lib/runtime/invocationOptions.ts
egonSchiele: Added in 4e4641b86, with the fix spelled out: an explicit `{}` replaces the env policy and then abstains on everything; a host that wants restrict-everything writes `{ "*": [{ action: "reject" }] }`. Also noted in the hosting doc and the spec §1 comment.

packages/agency-lang/tests/agency-js/serve-policy/test.js
egonSchiele: Added in 4e4641b86: child.agency reads SERVE_POLICY_SECRET, subprocessRead runs it via runFile under a policy that approves std::run and rejects std::env, and the fixture pins { success: true, data: null } — the reject crossed the IPC boundary. Verified through the real js runner.

packages/agency-lang/docs/dev/hosting/how-hosted-serving-works.md
egonSchiele: Went through everything added in this PR. Cut: the process-history paragraph in roadmap B2 (now states the current model forward, without narrating the correction), the "narrow field on purpose" rationale in node.ts, the "nothing catches this by type" clause in invocationOptions.ts, and two sentences of the interrupts.ts resume comment. Kept: the hosting doc section (reference material for hosts), the interrupts.md cross-reference, and the test-case comments that say what each case proves.

PR #979, opened 2026-08-30
Accept the on-clause handler syntax models write by default (#926)

### LINKED ISSUE #926: Accept the handler syntax models reach for by default (handle (...) with { on effect(x) { ... } })

egonSchiele:
## What this does

Closes #926 (first of three PRs from that issue).

When the stdlib coding agent is asked to write a handler and has no docs in front of it, it reaches for a Koka/OCaml-flavored syntax every time:

```
let res: Result<string> = handle (foo(dir: dir)) with {
  on std::read(data) { approve() }
  on std::write(data) { if (data.dir == ".") { approve() } else { reject() } }
  on _ { reject() }
}
```

None of that used to parse, so the agent lost a whole round trip to a parse error whose message (`Expected function body`) did not say what was wrong. This PR accepts that shape as a lenient alias — the way `===` is accepted for `==` — and `agency fmt` rewrites it to the one canonical handler form. It is not documented for users or recommended; it exists so an agent that guesses it gets working code.

## Approach

Everything desugars **at parse time** into the existing `handleBlock` AST — an inline handler `(intr) { return match(intr.effect) { ... } }`. Because the alias leaves no distinct node behind, codegen, the type checker, and the formatter are all untouched, and `agency fmt` prints the canonical form for free. The whole feature is a parser change plus one pure AST-transform module (`lib/parsers/onClauseHandler.ts`).

Two surface forms are accepted:

- **`on`-clause handler body** — `with { on <effect>(param) { ... } ... }`. Each clause becomes a `"<effect>" =>` arm; `on eff(param)` binds `param` to `intr.data` (so `data.dir` reads `intr.data.dir`, where the real effects put it); a bare tail `approve()`/`reject()`/`pass()`/`propagate()` is lifted to a `return`; a clause that still doesn't return gets `return pass()` ("no verdict means pass"); a missing `on _` becomes `_ => pass()`.
- **Expression-position `handle`** — `let res = handle (expr) with H`, which wraps the assignment inside the handle body, reusing the `with`-modifier trick (locals live on `__stack.locals`, so the bound name escapes).

Three bad shapes are refused with their own committed parse errors (so the real message surfaces, not a generic one): a duplicate effect (compared on the normalized name), an empty `with { }`, and an `on _` that isn't the last clause.

## Safety

Handlers are safety infrastructure, so the invariant checked hardest is that **no clause shape is fail-open** — no spelling can turn a reject into an approve, and an absent verdict propagates to the user. `tests/agency/on-clause-handler.agency` runs the same program under both the alias and the canonical form and asserts identical verdicts, including the else-less-`if` case that would otherwise silently degrade to pass.

Interrupt semantics are unchanged: every form desugars to the existing handler before codegen, so the runtime and the decision point are exactly what they were.

## Tests

- `lib/parsers/onClauseHandler.test.ts` — the pure builders (lifting, completion, handler assembly); the flagship test compares the built AST against a **parsed** canonical handler.
- `lib/parsers/handleBlock.test.ts` — the surface forms and the three parse errors (asserting the message text, not just failure).
- `lib/parsers/errorExamples.test.ts` — the three new messages' example blocks parse.
- `lib/backends/onClauseHandlerFormat.test.ts` — `agency fmt` normalizes the alias to the canonical handler.
- `tests/agency/on-clause-handler.agency` — verdict parity (no LLM).

Local: `pnpm run typecheck` (all 3 configs), `pnpm test:run lib/parsers` (2011 → 2028 tests incl. the whole-corpus round-trip identity guard), `pnpm run fmt:ts`, `pnpm run lint:structure`, and the execution fixture all pass.

## Follow-ups (separate PRs, designed in the spec, not in this one)

1. Effect patterns in `match` (`std::read({ data }) =>`), a user-facing DX feature.
2. Multiple handlers on one block (`with h1, h2`).
3. Deferred: single-object `raise effect::name({ ... })` filling `intr.data` instead of `intr.message`.


packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -5958,6 +5970,122 @@ const functionRefHandlerParser: Parser<HandleBlock["handler"]> = (input) => {
...
+);
+
+const underscoreToken: Parser<string> = map(char("_"), () => "_");
+
+// `on <effect|_> (<param|_>)? { body }` — one clause.
+const onClauseParser: Parser<ParsedOnClause> = (input) => {
+  const parsed = seqC(
+    str("on"),
+    spaces,
+    capture(or(underscoreToken, quotedEffectName, effectIdentifier), "effectName"),
```
egonSchiele: `underscoreToken` before `effectIdentifier` breaks effect names that start with an underscore: for `on _foo(data)`, `underscoreToken` matches just the `_`, the `or` commits to it, and the clause then fails looking for `(` or `{` at `foo`. Same for a param named `_tmp` on line 6002.

`varNameChar` includes `_` (parsers.ts:220), so `effectIdentifier`'s bare path and `many1WithJoin(varNameChar)` already match both `_` and `_foo` — and the `captures.effectName === "_"` / `paramName === "_"` normalization below already turns a bare `_` into the catch-all / no-binding. So `underscoreToken` can be deleted from both positions with no behavior change except fixing the `_foo` case. (Worth a two-line test either way: `on _foo(data)` should parse as an effect named `_foo`.)

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -5958,6 +5970,122 @@ const functionRefHandlerParser: Parser<HandleBlock["handler"]> = (input) => {
...
+  if (!open.success) {
+    // Not a `{`-block at all — fail without committing so the enclosing `or`
+    // reports its own error (this branch never claimed the input).
+    return open as ParserResult<HandleBlock["handler"]>;
+  }
+  const clausesResult = many1(onClauseParser)(open.rest);
+  if (!clausesResult.success) {
+    // `{` opened but yielded no clause: empty `with { }`, or a first token that
+    // is not `on`. Either way, the handler has no clauses.
+    return commitHandlerFailure(EMPTY_HANDLER_BLOCK, input);
```
egonSchiele: Two related problems with this branch:

1. `many1` failing does not mean the block was empty — it also fires when the **first clause is malformed**, e.g. `with { on std::read(data) { <garbage> } }`. That user gets "a handler needs at least one `on` clause" for a block that plainly has one, which is actively misleading.
2. `commitHandlerFailure` overwrites `getParseState().committedFailure` unconditionally. If the clause body committed a more specific failure on the way down (say a ternary or if-expression message from inside the clause), this replaces it with the wrong-er EMPTY message.

Suggestion: probe for `}` right after the opening `{` — that is the true empty case and the only place EMPTY_HANDLER_BLOCK belongs. For a failed clause, either let an already-committed inner failure stand (check the channel before writing) or commit a "malformed `on` clause" message anchored at the clause. Related: a malformed *second* clause dies at the `char("}")` check below, which returns uncommitted, so that shape falls back to a generic handle error — committing there too ("expected `}` to close the handler or another `on` clause") would finish the job. Fine as a follow-up if you want to keep this PR's scope, but the first-clause case is worth fixing here since the message is wrong, not just generic.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -5958,6 +5970,122 @@ const functionRefHandlerParser: Parser<HandleBlock["handler"]> = (input) => {
...
+    char("}"),
+    optionalSpacesOrNewline,
+  )(clausesResult.rest);
+  if (!close.success) {
+    return close as ParserResult<HandleBlock["handler"]>;
+  }
+
+  const duplicate = findDuplicateEffect(clauses);
+  if (duplicate !== null) {
+    return commitHandlerFailure(DUPLICATE_ON_CLAUSE(duplicate), input);
```
egonSchiele, line 6100: Nit: the committed failures here anchor at `input` — the `{` that opens the block — so the reported position for a duplicate (and for `on _` not last) points at the start of the handler rather than the offending clause. The existing probes anchor at `probed.rest`, i.e. where the problem is. `ParsedOnClause` doesn't carry a loc today, so this would mean threading one through; fine to leave as-is if you'd rather not, but then a one-line comment saying the position is deliberately coarse would stop the next person from "fixing" it blind.

packages/agency-lang/tests/agency/on-clause-handler.agency
```diff
@@ -0,0 +1,61 @@
...
+  // The else-less conditional clause: "." must still be APPROVED, not degraded
+  // to pass. This is the execution-level check for the else-less-if lifting.
+  let c: Result<string> = handle (act(dir: ".")) with {
+    on app::write(data) { if (data.dir == ".") { approve() } }
+  }
+  return "${tag(a)} ${tag(b)} ${tag(c)}"
+}
+
+// The canonical hand-written form of the same three handlers.
+node canonical(): string {
```
egonSchiele, line 49: The parity claim in the header comment ("both nodes run the same program under the two spellings and must agree") is only true for cases `a` and `b` — the `canonical` node has no counterpart for `c`, so the else-less clause is asserted against an expected string but not against its canonical desugar. Adding the hand-written form of what completion produces —

```
"app::write" => {
  const data = intr.data
  if (data.dir == ".") { return approve() }
  return pass()
}
```

— and extending the expected output to `"ok no ok"` would pin that the desugared shape and the hand-written one reach the same verdict, which is the actual parity property for the completion transform. Also worth a one-line note in the fixture that the completion *pass* path (the guard false with no catch-all) is deliberately not executed here because pass propagates to interactive approval, which this no-LLM test cannot answer.

egonSchiele (commented):
Reviewed against the spec and plan. The desugar is faithful: verdict lifting, clause completion, blockBody on multi-statement arms, committed failures, and the on-_-last rule all made it in, and the execution fixture covers the else-less clause that was the plan review's main blocker. Codegen, the formatter, and the typechecker are untouched, exactly as the spec requires, and the handler still lowers to the existing handleBlock AST, so the interrupt decision point is unchanged.

Four comments inline: one small parser bug (underscore-prefixed effect names), one misleading-error case (EMPTY_HANDLER_BLOCK fires for malformed clauses and clobbers inner committed failures), and two nits (error position anchoring, and the canonical parity node not covering the else-less clause). Nothing here changes a verdict path, so none of it is a safety issue.

packages/agency-lang/docs/dev/language/on-clause-handlers.md
```diff
@@ -0,0 +1,122 @@
+# The `on`-clause handler alias
```
egonSchiele, line 1: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Fixed. Deleted `underscoreToken` from both positions — the effect goes through `effectIdentifier` (whose bare path matches `_` and `_foo`) and the param through `variableNameParser`, and the `=== "_"` normalization still turns a lone `_` into the catch-all / no-binding. Added a test that `on _foo(data)` parses as an effect named `_foo` (its first arm matches the string `"_foo"`, not the wildcard).

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Fixed both. `onClauseHandlerParser` now probes for `}` right after `{` — that is the only place EMPTY_HANDLER_BLOCK fires. A failed first clause instead surfaces as MALFORMED_ON_CLAUSE, and before committing anything I check `getParseState().committedFailure` so a more specific inner failure (a ternary/if-expression message from inside the clause body) stands instead of being clobbered. I took the follow-up too: the close-`}` failure path applies the same rule (keep an inner committed failure, else emit MALFORMED_ON_CLAUSE), so a malformed second clause no longer falls back to a generic handle error. Added a test that a malformed first clause reports "malformed", not "empty".

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Kept the coarse anchor and added the one-line comment you suggested: the duplicate and catch-all-last checks run over the parsed clause list, which carries no location, so they deliberately anchor at the handler's opening brace rather than the offending clause. Threading a loc through ParsedOnClause would be the fix if we want per-clause positions later; the comment is there so the next person does not "fix" it blind.

packages/agency-lang/tests/agency/on-clause-handler.agency
egonSchiele: Fixed. The `canonical` node now has a `c` handler — the exact shape completion produces for the else-less clause (`if (data.dir == ".") { return approve() } return pass()`, plus `_ => return pass()`) — and its expected output is now `"ok no ok"`, so the else-less branch is compared against its canonical desugar, not just an expected string. Added the note you asked for: the pass path (guard false, no catch-all) is deliberately not executed here because pass propagates to interactive approval, which this no-LLM test cannot answer.

packages/agency-lang/docs/dev/language/on-clause-handlers.md
egonSchiele: Reviewed the doc and every comment I added. My read: the dev doc is durable reference for the next maintainer of this area — what the alias is, the desugar rules, the two deliberate boundaries, the parse errors, and where the code and tests live — not thought-process, so I kept it (it is the docs/dev-note-per-feature the repo asks for). I did trim two comments that had drifted into narration: the `definitelyReturns` docstring in onClauseHandler.ts and the `handleExprAssignmentParser` docstring in parsers.ts, both cut to the load-bearing why. The remaining comments each explain a non-obvious decision (why `variableNameParser` over raw `varNameChar`, why failures are committed, why `blockBody` is set on multi-statement arms). If any of those still read as notes to you, point me at them and I will cut further.

packages/agency-lang/lib/parsers/onClauseHandler.ts
```diff
@@ -0,0 +1,182 @@
...
+const varName = (value: string): VariableNameLiteral => ({ type: "variableName", value });
+
+/** `intr.<prop>` as a value-access node. */
+const intrMember = (prop: string): ValueAccess => ({
+  type: "valueAccess",
+  base: varName("intr"),
+  chain: [{ kind: "property", name: prop }],
+});
+
+const intrParam: FunctionParameter = { type: "functionParameter", name: "intr" };
```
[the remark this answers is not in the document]
egonSchiele: Investigated this. The alias desugars into the same AST as a hand-written `handle {} with (intr) { ... }`, so I tested whether the canonical handler has the same behavior: with an outer local named `intr`, the canonical `with (intr)` handler reaches the wrong verdict too (both the alias and the hand-written form return the failing branch, not approve). So this is a pre-existing handler-parameter shadowing gap in the language, not something the alias introduces — it inherits exactly the hygiene a hand-written handler has.

I did not paper over it by renaming the generated parameter, because a collision-proof name (say `__intr`) would make `agency fmt` emit `with (__intr)` — non-idiomatic, and it would break the "normalizes to clean canonical Agency" property that is the whole point of the alias. The right fix is at the language level (handler params should shadow outer locals), applied uniformly to all handlers, which is out of scope for this PR. Flagging it for a separate issue. I removed the hygiene node I had briefly added, rather than assert a property the language does not currently provide.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -5958,6 +5970,122 @@ const functionRefHandlerParser: Parser<HandleBlock["handler"]> = (input) => {
...
+    str("on"),
+    spaces,
+    capture(or(underscoreToken, quotedEffectName, effectIdentifier), "effectName"),
+    optionalSpaces,
+    optional(
+      captureCaptures(
+        seqC(
+          char("("),
+          optionalSpaces,
+          capture(or(underscoreToken, many1WithJoin(varNameChar)), "paramName"),
```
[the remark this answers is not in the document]
egonSchiele: Fixed here and at line 6194 (the expression-handle declared name): both now parse through `variableNameParser`, which requires a letter or underscore first, so `on eff(1data)` and `let 1res = handle(...)` are rejected instead of desugaring to `const 1data = ...` that the canonical grammar cannot re-parse. Added tests for both.

packages/agency-lang/tests/agency/on-clause-handler.agency
```diff
@@ -0,0 +1,61 @@
...
+  let b: Result<string> = handle (act(dir: "/etc")) with (intr) {
+    return match (intr.effect) {
+      "app::write" => {
+        const data = intr.data
+        if (data.dir == ".") { return approve() } else { return reject() }
+      }
+      _ => return reject()
+    }
+  }
+  return "${tag(a)} ${tag(b)}"
```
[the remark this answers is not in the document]
egonSchiele: Fixed. The `canonical` node now runs the `c` (else-less completion) handler too, and its expected output is `"ok no ok"`, so the branch-completion case is compared against its canonical desugar.

PR #980, opened 2026-08-31
Effect patterns in match (#926, PR 2 of 3)

egonSchiele:
## Effect patterns in `match` (#926, PR 2 of 3)

Lets a `match` arm name an interrupt effect directly and optionally destructure its payload, instead of comparing `intr.effect` against a string by hand. Inside a handler:

```
handle {
  a = act(dir: ".")
} with (intr) {
  return match (intr) {
    app::write({ data }) if data.dir == "/tmp" => return approve()
    app::write => return reject()
    _ => return reject()
  }
}
```

The `on`-clause handler alias from PR 1 (#979) desugars to this string-compare form today; this PR gives `match` a first-class spelling for it.

### Design

A new `effectPattern` variant of `MatchPattern`, modeled end-to-end on `resultPattern` (`success(s)` / `failure(e)`). Two forms:

- `app::write` — matches any interrupt whose effect is `app::write`.
- `app::write({ data })` — the same, plus an object-pattern binding over the whole interrupt, so `{ data }` binds `data = intr.data`.

The scrutinee is the whole interrupt (`match(intr)`), so the binding reaches its `effect` / `message` / `data` / `origin` fields; deep destructures like `app::write({ data: { dir } })` work because the binding is an ordinary object pattern. It lowers to a shape check, an `intr.effect == "app::write"` equality, and the object binding's own checks and bindings — no new runtime helper. Like `resultPattern`, it lives only in `MatchPattern` (match arm and after `is`), never in `let`/`const`/`for`.

### Behavior

- **Namespaced-only.** The parser intercepts a `::`-containing name; a bare `foo` stays a binder. A positional binding `std::read(data)` (the mistake this arc absorbs) commits to a specific error naming the object-pattern form.
- **Shape check first**, so the `.effect` read is guarded — a null or non-object scrutinee fails the arm rather than throwing. The whole-corpus `patternGuards` tripwire confirms the lowered condition reads no unguarded value.
- **Exhaustiveness.** An effect pattern pins the `effect` discriminant, so it behaves like the `{ effect: "app::read" }` object pattern it is sugar for. Over a real inline handler param (a closed discriminated union keyed on `effect`), covering every effect is exhaustive with no `_`, and a missing effect is reported by name — but a bound arm only covers its member when the binding is irrefutable (bare or pure binders); a value-matching binding still needs a `_`. Over an open (`any`) scrutinee the match stays open. An effect arm never counts as a catch-all. (This corrects the earlier "never makes a match exhaustive / no code change" framing, after review found the discriminated-union case.)
- **Limitations** documented in the dev note: bare-named effects have no spelling here (use `match(intr.effect)`); duplicate arms are first-wins; a non-interrupt scrutinee surfaces a member-access diagnostic on generated code.

### Verification

- `pnpm run typecheck` (3 configs), `pnpm run fmt:ts`, `pnpm run lint:structure` — all green.
- `pnpm test:run lib/parsers lib/lowering lib/backends lib/typeChecker` — 4545 pass, including the whole-corpus round-trip identity and the `patternGuards` tripwire now exercising an effect pattern.
- Execution test (no LLM): the effect-pattern handler and a string-compare handler both approve `.` and reject `/etc` (`"ok no"`).

### Notes for the owner

- This is a user-facing feature, so a `docs/site/guide/` page is warranted; guide pages are yours to write, so this PR only adds the `docs/dev` note.
- One open item flagged in the plan review, resolved as document-the-limit for v1: bare-named effects. If you want a dead-arm warning for a string-literal arm on an interrupt scrutinee, that is a separate general typechecker diagnostic.

### DIFF
--- a/packages/agency-lang/docs/dev/language/effect-patterns.md
+++ b/packages/agency-lang/docs/dev/language/effect-patterns.md
@@ -1,0 +1,104 @@
+# Effect patterns in `match`
+
+An effect pattern lets a `match` arm name an interrupt effect directly and
+optionally destructure its payload, instead of comparing `intr.effect` against
+a string by hand. Inside a handler:
+
+```
+handle {
+  a = act(dir: ".")
+} with (intr) {
+  return match (intr) {
+    app::write({ data }) if data.dir == "/tmp" => return approve()
+    app::write => return reject()
+    _ => return reject()
+  }
+}
+```
+
+Before this feature the same handler was written with `match(intr.effect)` and
+string arms, pulling the payload out by hand with `const data = intr.data`.
+
+## What it is
+
+An `effectPattern` is a new `MatchPattern` variant (`lib/types/pattern.ts`),
+modeled on `resultPattern` (`success(s)` / `failure(e)`). It has two forms:
+
+- `app::write` — matches any interrupt whose effect is `app::write`, binds
+  nothing.
+- `app::write({ data })` — the same match, plus an object-pattern binding over
+  the whole interrupt. `{ data }` binds `data = intr.data`.
+
+Match on the whole interrupt, written `match(intr)`. The older form matched on
+`intr.effect`; an effect pattern matches the interrupt itself so the binding can
+reach its other fields. Those fields are `effect`, `message`, `data`, and
+`origin`, so `app::write({ data })` binds the payload (which lives under
+`intr.data`), and a deep destructure like `app::write({ data: { dir } })` binds
+`dir = intr.data.dir`.
+
+Like `resultPattern`, an effect pattern lives only in `MatchPattern` — legal in
+a match arm and after `is` (`intr is app::write`, `if (intr is app::write({ data }))`),
+never in `let`/`const`/`for`. It is kept out of `BindingPattern`.
+
+## How it lowers
+
+The pattern lowers (`lib/lowering/patternLowering.ts`) to existing nodes — no
+new runtime helper. `app::write({ data })` becomes:
+
+1. a shape check on the scrutinee (`intr != null && intr is object`),
+2. an equality `intr.effect == "app::write"`, and
+3. the object binding's own checks and bindings against the same source.
+
+The shape check goes first so the `.effect` read is guarded: a null or
+non-object scrutinee fails the arm rather than throwing. This keeps the lowered
+condition a total function of the value, the invariant
+`patternGuards.tripwire.test.ts` enforces across the whole corpus.
+
+The binding is an ordinary `objectMatchPatternParser` object pattern, so it
+carries full object-pattern semantics: value matchers add checks
+(`app::write({ data: "x" })` also tests `intr.data == "x"`) and deep
+destructures work (`app::write({ data: { dir } })`).
+
+## The sites it touches
+
+Every place that switches on `resultPattern` gains a parallel `effectPattern`
+case. Grepping `effectPattern` across `lib/` is the checklist: the type and the
+`MatchPattern` union (`lib/types/pattern.ts`), the parser and its two wiring
+sites (`lib/parsers/parsers.ts` — `_matchPatternBase` and `_isRhsParser`, both
+before the binder / type parsers), lowering (`extractBindings`, `collectChecks`,
+`walkPattern`, the arm-value type list), the formatter
+(`lib/backends/agencyGenerator.ts`), and the per-node registries
+(`lib/utils/topLevel.ts`, `lib/utils/identifierSlots.ts`).
+
+## Limitations
+
+- **Only namespaced effect names.** The parser intercepts a `::`-containing
+  name (`namespaceIdentifier`). A bare `foo` in case position stays a
+  `variableName` binder, as before — a `::` name can never be a
+  variable, so there is no ambiguity. Effects are not required to be
+  namespaced, so a **bare-named effect** (`deploy`) has no effect-pattern
+  spelling. Handle it with `match(intr.effect)` and string arms. Do not reach
+  for a quoted `"deploy"` arm under `match(intr)`: that compares the whole
+  interrupt object to a string and never matches.
+
+- **Duplicate effect arms are first-wins**, like duplicate literal arms — the
+  first matching arm runs. The #926 `on`-clause handler alias instead rejects a
+  duplicate effect at parse time, because that alias is generated code where a
+  duplicate is almost certainly a bug. A `match` is ordinary code, where
+  first-wins is the language-wide rule.
+
+- **A non-interrupt scrutinee** (`match(5) { std::read => ... }`) surfaces as a
+  member-access diagnostic on the lowered `.effect` read — phrased on generated
+  code, so it names `.effect` rather than the arm the author wrote. This is the
+  same class of edge as the type-pattern edges, and a known v1 limitation.
+
+- **Open set: always needs `_`.** Effect names are open, so an effect pattern
+  never makes a match exhaustive. The exhaustiveness checker treats an
+  interrupt scrutinee as open/unsupported — it stays silent rather than guess a
+  missing case, and an effect-pattern arm never counts as a catch-all.
+
+## Not covered here
+
+This is the parser/lowering side. The user-facing guide page under
+`docs/site/guide/` is the owner's to write; nothing in this PR edits
+`docs/site/**`.
--- a/packages/agency-lang/docs/dev/language/effect-patterns.md
+++ b/packages/agency-lang/docs/dev/language/effect-patterns.md
@@ -92,10 +92,24 @@ before the binder / type parsers), lowering (`extractBindings`, `collectChecks`,
   code, so it names `.effect` rather than the arm the author wrote. This is the
   same class of edge as the type-pattern edges, and a known v1 limitation.
 
-- **Open set: always needs `_`.** Effect names are open, so an effect pattern
-  never makes a match exhaustive. The exhaustiveness checker treats an
-  interrupt scrutinee as open/unsupported — it stays silent rather than guess a
-  missing case, and an effect-pattern arm never counts as a catch-all.
+## Exhaustiveness
+
+An effect pattern pins the `effect` discriminant, so it behaves like the
+`{ effect: "app::read" }` object pattern it is sugar for
+(`matchExhaustiveness.ts`, `armDiscriminantValue`). What that means depends on
+the scrutinee's type:
+
+- Over a real inline handler param, which is re-typed to a discriminated union
+  keyed on `effect` (`handlerParamTyping.ts`), effect-pattern arms discriminate
+  the union: covering every effect makes the match exhaustive with no `_`, and
+  a missing effect is reported by name.
+- Over an open scrutinee (an `intr: any`, or a param the handler-typing pass
+  did not narrow), the match is open/unsupported territory — the checker stays
+  silent and requires nothing. Include a `_` there so an expression match does
+  not fall through to `undefined`.
+
+An effect-pattern arm never counts as a catch-all; only `_` and a bare binder
+do.
 
 ## Not covered here
 
--- a/packages/agency-lang/docs/dev/language/effect-patterns.md
+++ b/packages/agency-lang/docs/dev/language/effect-patterns.md
@@ -6,10 +6,10 @@ a string by hand. Inside a handler:
 
 ```
 handle {
-  a = act(dir: ".")
+  let a = act(dir: ".")
 } with (intr) {
   return match (intr) {
-    app::write({ data }) if data.dir == "/tmp" => return approve()
+    app::write({ data }) if (data.dir == "/tmp") => return approve()
     app::write => return reject()
     _ => return reject()
   }
@@ -47,7 +47,9 @@ new runtime helper. `app::write({ data })` becomes:
 
 1. a shape check on the scrutinee (`intr != null && intr is object`),
 2. an equality `intr.effect == "app::write"`, and
-3. the object binding's own checks and bindings against the same source.
+3. the object binding's per-property checks and bindings against the same
+   source (the shape check on the source is not repeated for the binding —
+   step 1 already emitted it).
 
 The shape check goes first so the `.effect` read is guarded: a null or
 non-object scrutinee fails the arm rather than throwing. This keeps the lowered
@@ -103,10 +105,15 @@ the scrutinee's type:
   keyed on `effect` (`handlerParamTyping.ts`), effect-pattern arms discriminate
   the union: covering every effect makes the match exhaustive with no `_`, and
   a missing effect is reported by name. A bound arm only covers its member when
-  the binding is irrefutable — bare (`app::read`) or pure binders
+  the binding is irrefutable — bare (`app::read`) or top-level pure binders
   (`app::read({ data })`). A value-matching binding
   (`app::read({ data: { path: "p" } })`) matches only some interrupts of that
-  effect, so it does not cover the member and a `_` is still required.
+  effect, so it does not cover the member and a `_` is still required. So does
+  a nested destructure (`app::read({ data: { dir } })`), even with only binders
+  inside: it lowers to a shape check on `intr.data` that a null or missing
+  payload fails at runtime. The object-pattern spelling follows the same rule —
+  `{ effect: "app::read", data: { path: "p" } }` pins the discriminant but does
+  not cover the member either.
 - Over an open scrutinee (an `intr: any`, or a param the handler-typing pass
   did not narrow), the match is open/unsupported territory — the checker stays
   silent and requires nothing. Include a `_` there so an expression match does
--- a/packages/agency-lang/docs/dev/language/effect-patterns.md
+++ b/packages/agency-lang/docs/dev/language/effect-patterns.md
@@ -102,7 +102,11 @@ the scrutinee's type:
 - Over a real inline handler param, which is re-typed to a discriminated union
   keyed on `effect` (`handlerParamTyping.ts`), effect-pattern arms discriminate
   the union: covering every effect makes the match exhaustive with no `_`, and
-  a missing effect is reported by name.
+  a missing effect is reported by name. A bound arm only covers its member when
+  the binding is irrefutable — bare (`app::read`) or pure binders
+  (`app::read({ data })`). A value-matching binding
+  (`app::read({ data: { path: "p" } })`) matches only some interrupts of that
+  effect, so it does not cover the member and a `_` is still required.
 - Over an open scrutinee (an `intr: any`, or a param the handler-typing pass
   did not narrow), the match is open/unsupported territory — the checker stays
   silent and requires nothing. Include a `_` there so an expression match does
--- a/packages/agency-lang/lib/backends/agencyGenerator.matchBlock.test.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.matchBlock.test.ts
@@ -194,3 +194,22 @@ def describe(value: any): string {
     expect(formatSource(src)).toContain("is string =>");
   });
 });
+
+describe("effect pattern formatting", () => {
+  it("round-trips bare and bound effect patterns", () => {
+    const src = `node main() {
+  match(intr) {
+    std::read => 1
+    std::write({ data }) => 2
+    _ => 0
+  }
+}
+`;
+    const first = formatSource(src)!;
+    // Idempotent: formatting the formatted output changes nothing.
+    expect(formatSource(first)).toBe(first);
+    // Both spellings survive: bare name and name with an object binding.
+    expect(first).toContain("std::read => 1");
+    expect(first).toContain("std::write({ data }) => 2");
+  });
+});
--- a/packages/agency-lang/lib/backends/agencyGenerator.matchBlock.test.ts
+++ b/packages/agency-lang/lib/backends/agencyGenerator.matchBlock.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from "vitest";
 import { formatSource } from "../formatter.js";
+import { parseAgency } from "../parser.js";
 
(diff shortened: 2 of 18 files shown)


packages/agency-lang/docs/dev/language/effect-patterns.md
```diff
@@ -0,0 +1,104 @@
+# Effect patterns in `match`
```
egonSchiele, line 1: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

egonSchiele (commented):
Reviewed against the plan (`2026-08-30-effect-patterns-in-match-plan.md`) and the interrupt model. Faithful execution, and the deviations from the plan all check out:

- The union extensions the plan missed (`ObjectPatternProperty.value`, `ArrayPattern.elements`, the `AgencyNode` union) were compiler-forced and are handled correctly downstream: I traced a nested effect pattern (an effect pattern as an object-property value) through `collectChecks`, `extractBindings`, and `formatPattern`, and all three recurse into it properly. So `match(o) { { intr: std::read({ data }) } => ... }` is now expressible and works. If that nesting is meant to be supported, one test would pin it; if it is incidental, fine to leave.
- Shape check before the `.effect` equality is a real improvement over the plan (the plan's snippet read `.effect` unguarded); the tripwire rationale in the comment is exactly right, and `resultPattern` escapes the same issue only because `isSuccess()` is a call, not a member read.
- Exhaustiveness as zero-code-plus-tests matches what the plan review predicted, and the two tests are the right pair (open set stays silent; an effect arm is not a catch-all).
- The execution fixture's parity design (effect-pattern node and string-compare node must independently produce `"ok no"`) is the safety check that matters: the sugar cannot turn a reject into an approve.

No safety issues: this is lowering-only sugar over the existing string compare, no new decision point, handler registration untouched.

Findings, most severe first — one small should-fix (the `?? null` in the parser), one convention question, and two test nits. Details inline.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -7261,6 +7262,55 @@ export const resultPatternParser: Parser<ResultPattern> = withLoc(
...
+      ),
+      optionalSpacesOrNewline,
+      char(")"),
+    )(name.rest);
+    if (!bindingResult.success) return bindingResult as ParserResult<EffectPattern>;
+    return success(
+      {
+        type: "effectPattern",
+        effect,
+        binding: (bindingResult.result.binding as ObjectPattern) ?? null,
```
egonSchiele: The `?? null` here can never fire: `parseError` only returns success when the whole sequence, including the `binding` capture, succeeded — `resultPatternParser` above does the equivalent without it. And if it somehow did fire, the failure mode is bad: the parens are already consumed, so `std::read({ data })` would quietly become the bare pattern — an arm that still matches the effect but has dropped its binding and any value-matcher checks. Please drop the `?? null` so a missing capture is a loud error instead of a silent downgrade.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -7261,6 +7262,55 @@ export const resultPatternParser: Parser<ResultPattern> = withLoc(
...
+      return success({ type: "effectPattern", effect, binding: null }, name.rest);
+    }
+
+    // Committed to the effect-pattern form: once we see `<effect>(`, the
+    // binding MUST be an object pattern. `parseError` bypasses the surrounding
+    // `or` — without it, `std::read(data)` (a positional binding, the mistake
+    // this alias exists to catch) would soft-fail, fall through to
+    // variableNameParser, and die later with an error naming nothing the
+    // author wrote. Mirrors resultPatternParser's commit above.
+    const bindingResult = parseError(
```
egonSchiele, line 7318: Convention question, no change needed if the precedent stands: PR #979 put its bespoke committed errors in `messages.ts` with extraction-tested examples, while this message is inline. Inline follows the nearest precedent (`resultPatternParser`, directly above), so this is defensible as is — but if messages.ts is the convention new committed errors should follow, this one is a candidate. Owner's call.

packages/agency-lang/lib/lowering/patternLowering.test.ts
```diff
@@ -730,6 +730,93 @@ describe("result patterns", () => {
...
+    expect(lowered).toHaveLength(2);
+    const ifNode = lowered[1] as IfElse;
+    expect(ifNode.type).toBe("ifElse");
+    expect(effectEqualityFor(ifNode.condition)).toBeDefined();
+    const dataBind = ifNode.thenBody[0] as Assignment;
+    expect(dataBind.variableName).toBe("data");
+    expect((dataBind.value as ValueAccess).chain).toEqual([{ kind: "property", name: "data" }]);
+  });
+
+  it("rejects a binding in a pure-boolean `is` context", () => {
```
egonSchiele, line 826: Two small companions would pin this boundary from both sides:

1. A positive control that the bare form is legal as a pure boolean — `let x = intr is std::read` lowers without throwing. The suite proves the binding form is rejected, but nothing at the lowering level proves the bare form is accepted; if the binder walk ever over-rejected, no test would go red.
2. A binding effect pattern inside a match-arm guard — e.g. an arm guarded by `if (intr is std::write({ data }))` — should hit the same `PatternLoweringError`. The error message promises that a match-arm guard is a pure-boolean context, and nothing exercises that promise for the new pattern.

packages/agency-lang/lib/backends/agencyGenerator.matchBlock.test.ts
```diff
@@ -194,3 +194,22 @@ def describe(value: any): string {
     expect(formatSource(src)).toContain("is string =>");
   });
 });
+
+describe("effect pattern formatting", () => {
+  it("round-trips bare and bound effect patterns", () => {
```
egonSchiele, line 200: Nit: this asserts idempotence plus two substrings, which is weaker than the plan's print-then-reparse-to-the-same-AST. The whole-corpus round-trip identity test does cover effect patterns now that the fixture is in the corpus (the tripwire's staleness guard proves the corpus picks it up), so this is genuinely minor — but if you touch this test again, a `parse -> format -> parse` with `toEqualWithoutLoc` would make the targeted case as strong as the backstop.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Dropped. A missing capture is now a loud crash instead of a silent downgrade to the bare pattern.

packages/agency-lang/lib/typeChecker/matchExhaustiveness.test.ts
```diff
@@ -264,6 +264,40 @@ def f(x: boolean): number {
...
+      `
+def f(intr: any): number {
+  return match (intr) {
+    std::read => 1
+    std::write => 2
+  }
+}`,
+      ERROR,
+    );
+    expect(errs.some((e) => /not exhaustive/i.test(e))).toBe(false);
```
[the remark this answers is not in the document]
egonSchiele: Verified and fixed. A real inline handler param is re-typed to a discriminated union keyed on effect (handlerParamTyping.ts), and coveredMemberKeys only recognized objectPattern discriminants, so effect-pattern arms covered zero members — a false 'missing' (a hard error for the expression-position return match). An effect pattern is sugar for { effect: "app::read" }, so a new armDiscriminantValue now pins the effect discriminant for both spellings. Covering every effect is exhaustive with no _, and a missing effect is reported by name. Added three handler-union tests and corrected the dev-doc exhaustiveness section.

packages/agency-lang/lib/lowering/patternLowering.test.ts
egonSchiele: Added both to the effect-patterns lowering block: a positive control that a bare intr is std::read lowers as a pure boolean without throwing (and carries the effect equality), and a match-arm guard x if (x is std::write({ data })) that hits the same PatternLoweringError. The guard case uses a top-level match to sidestep the pre-existing parser bug with guards nested in a node body.

packages/agency-lang/lib/backends/agencyGenerator.matchBlock.test.ts
egonSchiele: Strengthened to a print -> reparse identity: the test now parses the source and the formatted output and compares with toEqualWithoutLoc, so a targeted case matches the whole-corpus backstop.

packages/agency-lang/lib/parsers/parsers.ts
egonSchiele: Left inline, following the nearest precedent (resultPatternParser directly above). Happy to move it to messages.ts with an extraction-tested example if that is the convention new committed errors should follow — your call.

packages/agency-lang/docs/dev/language/effect-patterns.md
egonSchiele: Reviewed the dev note and every comment in the PR. My read is they are reference material for a future maintainer, not session notes: the doc records what the feature is, how it lowers, the wiring sites to keep parallel with resultPattern, and the limits; the code comments explain the non-obvious constraints (why the parser commits after effect(, why the shape check precedes the .effect read, the parser ordering). None narrate what I did or why I chose it, so I kept them. The one change from this pass is the exhaustiveness section, corrected for the Copilot finding. If any specific passage still reads as implementation notes to you, point it out and I will cut it.

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
```diff
@@ -59,6 +59,22 @@ function objectPatternDiscriminantValue(
...
+ * (`app::read` / `app::read({ data })`). An effect pattern is sugar for the
+ * object pattern that pins `effect` to the name, so it pins the `effect`
+ * discriminant and no other. A handler param is a discriminated union keyed on
+ * `effect` (handlerParamTyping.ts), so this is what lets effect-pattern arms
+ * count toward covering that union — without it every effect reads as missing.
+ */
+function armDiscriminantValue(cv: CaseValue, prop: string): string | number | boolean | null {
+  if (cv !== "_" && cv.type === "effectPattern") {
+    return prop === "effect" ? cv.effect : null;
+  }
```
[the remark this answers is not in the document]
egonSchiele: Fixed. You are right — a bound arm with a value matcher matches only some interrupts of the effect, and counting it let an expression match pass without _ and fall through to undefined. isIrrefutableEffectBinding now gates coverage: a bound arm counts only when its binding is bare or pure binders (shorthands, renames, rest, nested all-binder objects); a value matcher, array pattern, result pattern, or nested type test makes it refutable, so it no longer covers the member and _ is still required. Added the value-matcher regression test plus a pure-binder positive control. (The equivalent object-pattern path, { effect: "a", x: "p" }, has the same pre-existing behavior; I scoped this fix to effect patterns and left that as a separate item.)

packages/agency-lang/lib/typeChecker/matchExhaustiveness.test.ts
```diff
@@ -269,6 +269,117 @@ def f(x: boolean): number {
...
+  raise app::read("r", { path: "p" })
+  return "r"
+}
+def wr(): string raises <app::write> {
+  raise app::write("w", { dir: "d" })
+  return "w"
+}
+`;
+
+  it("error: covering every effect is exhaustive with no `_`", () => {
```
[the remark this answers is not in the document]
egonSchiele: Updated the PR description: the exhaustiveness bullet now distinguishes a closed handler union (effect patterns can make it exhaustive without _, irrefutable bindings only) from an open scrutinee (stays open), and notes it corrects the earlier framing.

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
```diff
@@ -59,6 +59,47 @@ function objectPatternDiscriminantValue(
...
+ * make the arm match only SOME interrupts of that effect, so it cannot cover the
+ * member. (Nested type tests also mark the whole arm guarded via
+ * hasNestedTypeTest, which drops it from coverage independently.)
+ */
+function isIrrefutableEffectBinding(binding: ObjectPattern | null): boolean {
+  if (binding === null) return true;
+  return binding.properties.every((p) => {
+    if (p.type === "objectPatternShorthand" || p.type === "restPattern") return true;
+    const v = p.value; // objectPatternProperty
+    if (v.type === "objectPattern") return isIrrefutableEffectBinding(v);
```
egonSchiele: **Exhaustiveness unsoundness: a nested all-binder destructure is counted irrefutable, but its lowering is refutable.**

`isIrrefutableEffectBinding` recurses into nested object patterns (`if (v.type === "objectPattern") return isIrrefutableEffectBinding(v)`), treating `app::read({ data: { dir } })` as covering the whole `app::read` member. But `collectChecks` lowers every nested destructure to a shape check on the field (`source.data != null && source.data is object`), which can fail at runtime: `data` may be null/undefined (an empty-payload effect raised with no data is accepted by `effectPayloadCheck.ts`), and `origin` is a string at runtime, so `app::read({ origin: { node } })` fails for every interrupt.

Concrete failure: `effect app::ping { }` raised as `raise app::ping("hi")`; a handler writes `return match (intr) { app::ping({ data: { } }) => reject() ... }` covering every effect. The checker certifies the match exhaustive with no `_`, but at runtime the data shape check fails, no arm matches, and the expression match yields `undefined` — the handler returns no verdict, exactly the fall-through this rule exists to prevent.

Fix: only top-level shorthand/rename/rest/wildcard properties are truly check-free; a nested object pattern should make the binding refutable (drop the recursion, return false for nested `objectPattern` values).

packages/agency-lang/docs/dev/language/effect-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+An effect pattern lets a `match` arm name an interrupt effect directly and
+optionally destructure its payload, instead of comparing `intr.effect` against
+a string by hand. Inside a handler:
+
+```
+handle {
+  a = act(dir: ".")
+} with (intr) {
+  return match (intr) {
+    app::write({ data }) if data.dir == "/tmp" => return approve()
```
egonSchiele: **The lead example is not valid Agency — it violates two rules from `packages/agency-lang/CLAUDE.md`'s "VERY IMPORTANT: Agency syntax rules".**

1. The guard `app::write({ data }) if data.dir == "/tmp" => ...` does not parse: the guard grammar requires `if (<expr>)` (verified with `pnpm run ast`: the parenthesized form parses, this form fails with the parser's own message about "an optional `if (...)` guard"). Rule: "`if`, `while`, and `for` statements REQUIRE parentheses around the condition".
2. `a = act(dir: ".")` has no `let a` / `const a` declaration in the snippet. Rule: "Variables must be declared with `let` or `const` before use." (The PR's own test file declares `let a: Result<string> = failure("unset")` before the same assignment.)

The same example appears in the PR description. A reader copying the canonical example from the doc that introduces the feature gets a parse error.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -7261,6 +7262,58 @@ export const resultPatternParser: Parser<ResultPattern> = withLoc(
...
+export const effectPatternParser: Parser<EffectPattern> = withLoc(
+  (input: string): ParserResult<EffectPattern> => {
+    // Namespaced name only. A soft failure here lets the outer `or` fall
+    // through to variableNameParser (a bare identifier is a binder).
+    const name = namespaceIdentifier(input);
+    if (!name.success) return name as ParserResult<EffectPattern>;
+    const effect = name.result;
+
+    // Bare form: no `(` after the name.
+    if (name.rest[0] !== "(") {
```
egonSchiele: **A space before the binding parens silently degrades to the bare pattern, skipping the committed diagnostic this parser exists to give.**

The bare-vs-bound decision is `name.rest[0] !== "("` with no whitespace skip, so `std::read ({ data }) => ...` parses `std::read` as the bare pattern, leaves ` ({ data })` unconsumed, and the arm dies later with a generic match-arm parse error naming nothing the author wrote — the exact failure mode the `parseError` commit below was added to prevent for `std::read(data)`. `resultPatternParser` has the same `rest[0]` check, but its bindings are short identifiers; an object-shaped binding is much more likely to be written with a space. Skipping optional spaces before testing for `(` (or committing when the next non-space char is `(`) closes the gap.

packages/agency-lang/lib/typeChecker/matchExhaustiveness.ts
```diff
@@ -59,6 +59,47 @@ function objectPatternDiscriminantValue(
...
+ * object pattern that pins `effect` to the name, so it pins the `effect`
+ * discriminant and no other. A handler param is a discriminated union keyed on
+ * `effect` (handlerParamTyping.ts), so this is what lets effect-pattern arms
+ * count toward covering that union — without it every effect reads as missing.
+ * A bound effect pattern only counts when its binding is irrefutable: a
+ * value-matching binding matches some interrupts of the effect, not all, so it
+ * cannot stand in for the member (an expression match over it would fall through
+ * to `undefined`).
+ */
+function armDiscriminantValue(cv: CaseValue, prop: string): string | number | boolean | null {
```
egonSchiele, line 116: **The two spellings the PR documents as equivalent now check differently — the object-pattern path still has the refutability hole this PR fixes for effect patterns.**

`armDiscriminantValue` requires an irrefutable binding before an effect-pattern arm covers its member, but delegates object-pattern arms to `objectPatternDiscriminantValue`, which returns the pinned `effect` literal while ignoring refutable sibling properties. So over a handler union, `{ effect: "app::read", data: { path: "p" } } => reject()` still counts as covering all of `app::read` with no `_`, and an `app::read` with another path falls through to `undefined` — the exact unsoundness the new test "a refutable value-matcher binding does NOT cover its member" pins for the sugar spelling. The hole is pre-existing, but the dev note and this comment both say an effect pattern "behaves like the `{ effect: ... }` object pattern it is sugar for", which is now false in the stricter direction. Either generalize the refutability check to object-pattern arms (sibling properties must be binders) or record the divergence so the next maintainer doesn't build on the unsound path.

packages/agency-lang/lib/lowering/patternLowering.ts
```diff
@@ -1319,6 +1326,28 @@ function collectChecks(pattern: MatchPattern, source: Expression, checks: Expres
     case "resultPattern":
       checks.push(resultCheckCall(pattern.kind, source, pattern.loc));
       break;
+    case "effectPattern": {
```
egonSchiele, line 1321: **Bound effect patterns emit the null+object shape check twice on the same source.**

This case pushes `shapeCheck(source, OBJECT_HINT)`, then `collectChecks(pattern.binding, source, checks)` hits the `objectPattern` case, which pushes the identical `shapeCheck(source, OBJECT_HINT)` again. Every `app::write({ data })` arm lowers to `intr != null && intr is object && intr.effect == "..." && intr != null && intr is object && ...` — two duplicate conjuncts (each cloning the source expression) in every bound arm's condition. Since the binding is over the same already-guarded source, recursing into the binding's properties directly (the per-property loop of the objectPattern case) instead of re-entering `collectChecks` on the whole binding drops the duplicates.

packages/agency-lang/lib/parsers/parsers.ts
```diff
@@ -7261,6 +7262,58 @@ export const resultPatternParser: Parser<ResultPattern> = withLoc(
...
 );
 
+// Effect patterns: `std::read`, `std::read({ data })`. Mirrors
+// resultPatternParser, but the constructor is a namespaced effect name and the
+// optional binding is an object pattern over the whole interrupt (the
+// scrutinee), not a single identifier. Must be placed in `_matchPatternBase`
+// and `_isRhsParser` BEFORE `variableNameParser` / `typePatternParser`. Only a
+// `::`-containing name is intercepted here; a bare `foo` stays a binder / type
+// reference because `namespaceIdentifier` soft-fails on it.
+export const effectPatternParser: Parser<EffectPattern> = withLoc(
```
egonSchiele, line 7280: **Hot-path cost: every ordinary match-arm binder now pays a full `namespaceIdentifier` scan that always soft-fails.**

`effectPatternParser` sits before `variableNameParser` in `_matchPatternBase` (and before `numberParser`/`_stringParser` in `_isRhsParser`), and it gates on the full `namespaceIdentifier` parser — which also rebuilds its `map(sepBy1(str("::"), many1WithJoin(varNameChar)))` combinator chain on every call. Before this PR that parser ran a handful of times per file (interrupt/raise/effect positions); now it runs once per match-pattern attempt and per `is` RHS across the corpus, in what is already the measured front-end bottleneck. Two cheap fixes: hoist the combinator to a module-level const (like every other named parser in this file), and/or gate with an inline scan (advance while `varNameChar`, require `"::"`) before invoking the parser. In `_isRhsParser` it also only needs to precede `typePatternParser`, not the literal parsers.

packages/agency-lang/lib/lowering/patternLowering.test.ts
```diff
@@ -730,6 +730,115 @@ describe("result patterns", () => {
   });
 });
 
+describe("effect patterns", () => {
+  /** Every binOp in an expression tree, so a test can find the effect-equality
+   *  conjunct without depending on where it sits in the `&&` chain. */
+  function binOps(expr: unknown): BinOpExpression[] {
```
egonSchiele, line 736: **The `binOps` helper hand-rolls a second AST walk this file already imports.**

`walkNodesArray` (lib/utils/node.ts) already descends into `binOpExpression` left/right and is used later in this same file (the matchSource describe). `binOps(expr)` is `walkNodesArray([expr as AgencyNode]).map(r => r.node).filter(n => n.type === "binOpExpression")`. The blind key-walk is a second, unmaintained descent definition that can silently diverge from the canonical walk when node shapes change.

PR #981, opened 2026-08-31
Embed an optional HMAC integrity checksum in every checkpoint

egonSchiele:
## What this does

Gives every checkpoint an optional, embedded HMAC checksum so a host can detect a tampered checkpoint on the external resume path, without the runtime enforcing anything.

The checkpoint is the serialized execution state of a paused program, and on the resume leg it comes back from a caller who could have edited it (changed a local, altered which env var a paused function reads). With a signing key configured, every checkpoint now carries an HMAC-SHA256 of its own contents in a `signature` field, and a host verifies with one call before resuming:

```ts
import { verifyCheckpointChecksum } from "agency-lang";
if (!verifyCheckpointChecksum(checkpoint)) {
  // refuse the resume
}
```

## Design points

- **Opt-in.** No `AGENCY_CHECKPOINT_KEY` in the environment means nothing is signed and nothing about a run changes. A present-but-short key (under 32 bytes) throws `CheckpointKeyTooShortError`.
- **Embed, do not enforce.** No runtime code path refuses a bad or missing signature (editing checkpoints is a supported feature: rewind, debugger, resume overrides). Verification is the host policy. A missing signature verifies false, so stripping the field does not dodge verification.
- **Whole-checkpoint coverage.** The MAC covers `toJSON()` minus the `signature` field, canonicalized with the existing `canonicalize` helper (key-sorted at every depth) under a fixed domain tag. No field allowlist to rot; the companion code-fingerprint field will be covered automatically.
- **Signing is automatic at the single creation chokepoint** (`Checkpoint.fromStateStack`), so interrupt, guard-trip, trace, and stdlib `checkpoint()` checkpoints are all covered. The legitimate edit paths re-sign: `applyOverrides` (rewind and resume overrides), `CheckpointStore.pin`, and `Checkpoint.clone`.
- **Rotation-ready.** Verify accepts the current key first, then retired keys from `AGENCY_CHECKPOINT_KEY_OLD` (comma-separated, verify-only), so rotating a key does not invalidate outstanding checkpoints.
- **Plain-JSON checkpoints sign and verify too.** On the external resume path, `interrupt.checkpoint` is parsed JSON rather than a `Checkpoint` instance; the module duck-types on `toJSON`. This was caught by running the interrupt-overrides agency-js test with a key set, and is pinned by a unit test.
- Signing and the key never touch Agency code or statelog; only `verifyCheckpointChecksum` is exported from the package surface.

## Tests

- `lib/runtime/checkpointChecksum.test.ts` (13 cases): sign/verify, tamper detection, wrong key, stripped-signature downgrade guard, key-order-independent round trip, a RICH checkpoint (guards, cost, savedDraft, branches) through the external zod round trip, no-key silence, short-key error, rotation via `_OLD`, re-sign after edit, pin and clone re-signing, and the plain-JSON shape.
- `schemas.test.ts`: `signature` survives `Checkpoint.fromJSON`.
- `checkpointStore.test.ts`: a created checkpoint is signed when a key is set, unsigned otherwise.
- Regression: `tests/agency-js/interrupts/interrupt-overrides` passes with and without a key configured.
- `pnpm run typecheck` (all three configs) and `pnpm run lint:structure` are green.

## Docs

`docs/dev/runtime/checkpoint-integrity.md` covers the embed-not-enforce rationale, the exact canonical form, key handling and rotation, the honest non-goals (no confidentiality, no freshness or binding, symmetric-key trust model), and the trace-mode performance note. Pointers added to `checkpointing.md`, the CLAUDE.md runtime index, and the `agency-runtime-docs` skill.

Plan: `docs/superpowers/plans/2026-08-30-checkpoint-integrity-checksum.md` (in the main checkout).

https://claude.ai/code/session_01G6gknY8UdagAQZABkRnJws

### DIFF
--- a/packages/agency-lang/docs/dev/runtime/checkpoint-integrity.md
+++ b/packages/agency-lang/docs/dev/runtime/checkpoint-integrity.md
@@ -1,60 +1,42 @@
 # Checkpoint integrity checksums
 
-A checkpoint is the serialized execution state of a paused program, and on the
-external resume path (HTTP `/resume`, a checkpoint loaded from disk) it comes
-back from a caller who could have edited it — changed a local, raised a budget,
-altered which environment variable a paused function is about to read. This
-feature gives a host a way to detect that: when a signing key is configured,
-every checkpoint carries an HMAC checksum of its own contents in a `signature`
-field, and the host can check it with one function call before resuming.
-
-## Embed, do not enforce
-
-The runtime signs but never refuses. No code path in agency-lang rejects a
-checkpoint because its signature is missing or wrong. That is deliberate:
-editing checkpoints is a supported feature — rewind replays with different
-values for a checkpoint's locals, the debugger edits state, `respondToInterrupts`
-accepts overrides — and trace mode writes a checkpoint per step that nobody
-signs a contract about. Enforcement is the host's policy, not the runtime's:
+On the external resume path (HTTP `/resume`, a checkpoint loaded from disk), a
+checkpoint comes back from a caller who could have edited it. When a signing
+key is configured, every checkpoint carries an HMAC checksum of its own
+contents in a `signature` field, and a host can check it with one call before
+resuming:
 
 ```ts
 import { verifyCheckpointChecksum } from "agency-lang";
 
-// On the resume leg, before calling respondToInterrupts:
 if (!verifyCheckpointChecksum(checkpoint)) {
   // refuse the resume — the checkpoint is not the one this host handed out
 }
 ```
 
 `verifyCheckpointChecksum` returns true only when the checkpoint carries a
-signature that validates under the configured key (compared in constant time).
-A missing signature returns false — so a caller cannot dodge verification by
-stripping the field (the downgrade guard). A host that has not opted in simply
-never calls verify.
+signature that validates under the configured key, compared in constant time.
+A missing signature returns false, so a caller cannot dodge verification by
+stripping the field.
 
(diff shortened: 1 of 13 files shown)


packages/agency-lang/docs/dev/runtime/checkpoint-integrity.md
```diff
@@ -0,0 +1,128 @@
+# Checkpoint integrity checksums
```
egonSchiele, line 1: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

egonSchiele:
add a small blurb about this feature to the bottom of docs/site/guide/checkpointing.md

egonSchiele:
Both comments addressed:

- **Docs and comments trimmed** — `checkpoint-integrity.md` is rewritten to what a future reader needs (what is signed, how a host verifies, the key, the non-goals); the design-history and planning narration is gone, and the longer inline comments across the runtime files are cut down to the constraint they state.
- **Guide blurb added** — a short "Verifying checkpoint integrity" section at the bottom of `docs/site/guide/checkpointing.md`: set `AGENCY_CHECKPOINT_KEY`, call `verifyCheckpointChecksum(checkpoint)` before resuming, and Agency itself never enforces.

PR #982, opened 2026-08-31
Refuse to resume a checkpoint whose referenced module code changed

egonSchiele:
## What this does

Refuses to resume a checkpoint when the **generated code** of a module it is paused inside has changed — and only then, leaving changes to unreferenced code alone.

A checkpoint resumes by replaying into generated TypeScript whose statements are numbered: the saved step counters index into the compiled module statement sequence. If that sequence changed between pause and resume, the counters point at different statements and the resume silently executes the wrong thing. This turns that into a refusal: `CheckpointCodeChangedError`, mapped to HTTP 409 in the serve adapter, with a message naming both code versions (when the checkpoint code was compiled vs. when the currently loaded code was).

The key property: **only modules with a live frame on the saved stack matter** (including frames inside fork/parallel/race branch stacks). A fresh call made after resume starts at step 0 and tolerates any code change, so you can freely edit modules the paused run merely imports.

## The contract: generated-output drift

The fingerprint is the sha256 of the module's **printed generated TypeScript** (not the `.agency` source bytes), computed just before the registration statement is appended as the file's last line:

```ts
__registerModuleFingerprint("<moduleId>", "<sha256 of everything above>", import.meta.url);
```

Hashing the output is deliberate: the statement sequence is what resume replays into, and it can change without the source file changing — a splice generator in another file, a codegen template, or the compiler itself. All of those flip the hash; conversely, if the generated output is byte-identical, resuming is safe regardless of what happened upstream. Source edits that normalize to identical generated output (none are known today — even comments reach the generated `__sourceMap`, so comment edits refuse) would be allowed to resume, which is correct under this contract.

No timestamp is embedded in the output — the registry derives "compiled at" from the artifact's mtime at load — so identical input emits identical bytes and incremental emit stays byte-identical to a force emit.

## How it works

- **Compile time:** `generateTypeScript` hashes the printed output and appends the registration when compiling with a stable module identity (the cwd-relative path, via `buildSession`). `compileSource` programs keep their per-compile random identity and register nothing — there is no stable key an equal recompile could match, so their checkpoints are not covered.
- **Runtime registry:** `moduleFingerprintRegistry.ts`, a null-prototype process-global map, rebuilt whenever code loads (like `__toolRegistry`, never serialized).
- **Claim time:** `claimFrameForScope` stamps `moduleId` on the frame next to `scopeName`. A legacy frame (pre-moduleId checkpoint) is backfilled on its matching re-claim; a same-scope re-claim from a different module throws a module-aware desync error. Both branches are tested.
- **Checkpoint creation:** `Checkpoint.fromStateStack` walks the serialized stack (branch stacks included) and stores the referenced modules' fingerprints in `moduleFingerprints`; `signCheckpoint` (from #981) stays the last statement so the HMAC covers them.
- **Resume:** `respondToInterrupts` calls `assertCodeUnchanged` before `restoreState`, so an out-of-date checkpoint is never partially executed. A stored module missing from the registry counts as changed. Rewind and the debugger restore directly and skip the check on purpose (in-process, the code cannot have changed under them).

## Honest limitations (all in the dev doc)

- Not a security boundary by itself — `moduleFingerprints` is caller-editable JSON; paired with the #981 checksum, stripping becomes detectable. This catches honest drift.
- A comment edit in a referenced module refuses (comments reach `__sourceMap`).
- Module identity is the cwd-relative input path, so resuming from a different working directory refuses; the error message names both causes.
- A redeploy invalidates outstanding paused runs in changed modules, stdlib upgrades included; there is no override.
- Granularity is per module; the field is keyed by moduleId so a finer scheme can extend it later.

## Tests

- `moduleFingerprintRegistry.test.ts` (mtime-derived timestamps, unknown-URL degradation, prototype-key safety), `referencedModules.test.ts` (walk with branch recursion, frame filter, refusal message carrying both timestamps), `moduleFingerprint.test.ts` (the appended hash covers exactly the bytes above it; identical input generates identical bytes; a code change flips the hash; `compileSource` output carries no registration).
- Schema round trips for the frame `moduleId` (no zod default — injecting the field into pre-fingerprint checkpoints would change their canonical bytes and break #981 signatures; the rich round-trip test pins this) and `moduleFingerprints`.
- `claimFrameForScope`: stamping, JSON round trip, legacy backfill, same-module re-claim no-op, and the cross-module desync throw.
- The 409 mapping in `adapter.test.ts`; end-to-end refusal + unreferenced-change-still-resumes in `tests/agency-js/interrupts/interrupt-code-changed`; overrides and subprocess resume suites pass with and without a signing key.

https://claude.ai/code/session_01G6gknY8UdagAQZABkRnJws


packages/agency-lang/docs/dev/runtime/checkpoint-code-fingerprints.md
```diff
@@ -0,0 +1,74 @@
+# Checkpoint code fingerprints
+
+A checkpoint resumes by replaying into generated code whose statements are
```
egonSchiele, line 3: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/lib/compiler/compile.ts
```diff
@@ -231,6 +232,7 @@ export function compileSource(source: string, config: CompileSourceOptions): Com
       moduleId,
       outputPath,
       initPlan,
+      sha256Text(compiledSource),
```
egonSchiele: Suggestion: Make the hash also include a timestamp. That way if a user tries to resume an out-of-date checkpoint, the error message can say when their checkpoint is from, and what the timestamp of the current code is.

packages/agency-lang/lib/runtime/moduleSourceHashRegistry.ts
```diff
@@ -0,0 +1,21 @@
+// A process-global map of moduleId -> sha256 of that module's source, populated
+// at module init by generated code (like __toolRegistry). NOT serialized: it is
+// derived from the loaded code and rebuilt whenever code loads. This is
+// per-loaded-code state, not per-run state, so a module-level map is the
+// correct owner (the GlobalStore rule is about per-run state).
+const registry: Record<string, string> = {};
```
egonSchiele: This can get tricky. How does it pull the module IDs for other files? Does this have any tricky interaction with agency's state isolation or global variable initialization?
https://agency-lang.com/guide/global-var-initialization.html

egonSchiele:
Review round addressed, and the CI failures are fixed (the generator fixture files needed regenerating for the new import line and the three-argument `claimFrameForScope`; `make fixtures` output is in the commit).

Per finding:

- **Legacy frames with no `moduleId`** — a matching re-claim now backfills the id, and a re-claim from a *different* module throws the same desync error a scope-name mismatch does (`claimFrameForScope`).
- **`compileSource` random moduleId** — with a caller-supplied `sourcePath` the module now keys by its cwd-relative path (matching `buildSession`), so its registrations survive a restart. An anonymous string compile has no stable identity and registers no hash at all, rather than registering one under a key nothing can ever match.
- **Hash vs parsed bytes** — the parse cache now reports the sha256 of the exact bytes it read (stored in the cache entry, so a cache hit returns the hash of the bytes the cached AST came from), and `buildSession` uses that instead of its own separate read.
- **Import collision** — the generated import is aliased `registerModuleSourceHash as __registerModuleSourceHash`, and the builder emits the alias.
- **Prototype-key hazards** — the registry and the collected record are both null-prototype now, with a `__proto__`/`constructor` test.
- **409 coverage** — added to `adapter.test.ts` beside the 402/generic cases.
- **Timestamp suggestion** — done. Registrations now carry the compile time, the checkpoint stores it per module, and the refusal reads: `The checkpoint ran code compiled <t1>; the current code was compiled <t2>` (or "the module is not loaded now" when it is gone entirely).
- **Docs/comments** — trimmed the dev doc and the inline comments to what a future reader needs; the design-history narration is gone.

On the registry question ("how does it pull the module IDs for other files, and does it interact with state isolation or global variable initialization?"): every compiled module carries its own `__registerModuleSourceHash(...)` as a plain top-level statement in its generated file — the same pattern as `__registerTool` — so importing the entry module transitively registers every module in the closure as it loads. It deliberately does not interact with either system: it is not in `GlobalStore` because it describes the loaded *code*, not a run's state (two concurrent runs of one process share the same code, a subprocess re-imports and rebuilds its own registry, and nothing is ever serialized into checkpoints except the per-checkpoint copy); and it is not part of static/globals init ordering because it registers immediately at import rather than deferring a function through `__registerStaticInit`/`__registerGlobalsInit` — it needs no ordering, and re-registration just overwrites with the same value.

egonSchiele:
Round 2 pushed: main is merged in (the HMAC PR and this one now compose — `moduleFingerprints` is collected first and `signCheckpoint` stays the last statement, so the checksum covers it), the CI failures are fixed, and the two new splice findings drove a redesign of what gets hashed.

**What CI was telling us:**

- The ~79 agency-test failures came from my round-1 change giving `compileSource` a path-based moduleId — that broke the plain-identifier moduleId contract (`CompiledProgram moduleId '../../../tmp/...' is not a plain identifier`) in the sandbox/test paths. Reverted: `compileSource` keeps its random identity and now registers **no** fingerprint at all — there is no stable key an equal recompile could match, so registering one was only ever going to produce false refusals.
- The `stdlibDeps` failures were the embedded compile timestamp breaking "incremental emit is byte-identical to force emit".

**The redesign (also answers both new Copilot splice findings):** the fingerprint is now the sha256 of the module's **printed generated output**, computed just before the registration statement is appended as the file's last line. The generated code is what resume replays into, so this covers splice expansion, template changes, and the compiler itself — anything that changes the statement sequence changes the hash, which a source-bytes hash could not guarantee. Because of that, the parse-cache TOCTOU fix from round 1 became unnecessary and is reverted (smaller diff; the hash and the emitted program are now the same artifact by construction).

The timestamp survives without breaking reproducibility: the emitted line is `__registerModuleFingerprint("<id>", "<hash>", import.meta.url)` — fully deterministic bytes — and the registry derives "compiled at" from the artifact's mtime at load time. The refusal message still names both code versions.

One more regression the round-1 tests caught during the merge: the zod `.default(null)` on the new frame `moduleId` field injected the field into checkpoints that never carried it, changing their canonical bytes — so a checkpoint signed by current main would have verified as *tampered* after this PR. The schema now leaves absent absent, with a comment saying why, and the rich-checkpoint round-trip test pins it.

Since the hash no longer covers source bytes, the field and APIs are renamed to say what they are: `moduleFingerprints` on the checkpoint, `registerModuleFingerprint`/`getModuleFingerprint` in the runtime.

Verified locally: the previously failing suites (`stdlibDeps`, `traceReader`, `runPolicy.spawn`, generator fixture suites, `templates/literal*`, `toolbox/writeTool`), the full checksum suite from the merged HMAC work, the e2e refusal test, and the overrides/subprocess resume tests with and without a signing key.

egonSchiele:
Both round-3 comments addressed: the PR title and description are rewritten to state the generated-output contract (what is hashed, why, and that source edits normalizing to identical output would resume — correct under this contract), and claimFrameForScope now has direct tests for the legacy null-module backfill, the same-scope different-module desync throw, and the matching re-claim no-op.

PR #983, opened 2026-08-31
eval upload: per-run progress output and the batch-first statelog link

egonSchiele:
## What this fixes

Two `agency eval upload` problems from today's debugging session:

1. **No progress output.** The command collected every run's outcome and printed everything only at the end, so a 69-run upload sat silent for minutes. `EvalUploadDependencies` now carries a `reportProgress` callback, and the CLI prints each run's line the moment that run finishes. The count line and batch link still print last, via the new `formatUploadSummary`; `formatUploadResult` remains as the all-at-once composition for tests and API callers.

2. **The batch link almost never printed, and pointed at an agent-keyed page.** The link required every uploaded run to share one valid agent name — runs that never called `setAgentName` (like the writing-rewrite suite's) got no link and were invisible on statelog's old agent-keyed evals page. Statelog now keys its evals pages by project and batch alone (statelog PR 66, "Make batches the top-level unit of the evals pages"), so the link becomes `/projects/evals/batch?id=<slug>&batch=<batch>` and prints whenever the runs share one batch id, named agent or not.

Also:

- `evalMain` in `stdlib/agents/writing/rewrite.agency` now calls `setAgentName("writing-rewrite")`, so rewrite batches carry an agent label for filtering on the evals page. (Existing uploaded traces keep their null name; the summary is computed at upload time.)
- `docs/dev/evals/eval-tracking.md` and the `agentName.ts` header now describe agent names as filter labels rather than URL path segments; the validation rule itself is unchanged.

## Pairing

Merge alongside statelog PR 66 — the link this prints only exists once that deploys. The old link only printed when every run shared a valid agent name, so nothing depends on the old URL shape.

## Testing

- `lib/cli/eval/upload.test.ts` updated: the batch-link tests use the new URL, a new test covers the link for unnamed runs, and a new test asserts one progress line per run. All green, plus `agentName.test.ts`.
- `pnpm run typecheck` (all three configs), `pnpm run fmt:ts`, and `pnpm run lint:structure` clean.
- `make` rebuilt cleanly after the stdlib change.

https://claude.ai/code/session_016HL3NJzBcspTuWRp88rga3

### DIFF
--- a/packages/agency-lang/docs/dev/evals/eval-tracking.md
+++ b/packages/agency-lang/docs/dev/evals/eval-tracking.md
@@ -143,11 +143,10 @@ next one still uploads. Running the command twice is safe: events skip,
 annotations upsert.
 
 Runs upload through a bounded pool, and each one prints its line as it
-finishes (`reportProgress` in `EvalUploadDependencies`), so a long upload
-shows progress instead of staying silent until the end. The result names
+finishes (`reportProgress` in `EvalUploadDependencies`). The result names
 the statelog batch page (`/projects/evals/batch?id=<slug>&batch=<batch>`)
 only when every uploaded run shares one batch id; statelog keys the page
-by project and batch alone, so no agent name is needed.
+by project and batch alone.
 
 The client, `lib/cli/statelog/evalUploadClient.ts`, is the seventh sealed
 statelog client (`docs/dev/hosting/statelog-clients.md`).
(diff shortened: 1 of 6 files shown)


packages/agency-lang/docs/dev/evals/eval-tracking.md
```diff
@@ -142,23 +142,27 @@ a trace it cannot complete. A failure on one directory is reported and the
 next one still uploads. Running the command twice is safe: events skip,
 annotations upsert.
 
-The result names the statelog batch page only when every uploaded run
```
egonSchiele, line 145: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/lib/statelog/agentName.ts
```diff
@@ -1,11 +1,11 @@
...
- * (`/evals/agents/<name>/batches/<batch>`), so it has to survive that trip
- * unchanged: no whitespace, no characters a URL would encode, and no `.` or
- * `..` segment, which a URL parser folds away (`agents/../batches` loses the
- * `agents` segment even after `encodeURIComponent`). Slashes are allowed so a
- * family can nest its variants, as in `agency-agent/coordinator`.
+ * Statelog uses the name as a filterable label on eval batches and in a
+ * `?agent=` query parameter, so it has to stay one clean token: no
+ * whitespace, no characters a URL would encode, and no `.` or `..` segment,
+ * which a URL parser folds away even after `encodeURIComponent`. Slashes are
+ * allowed so a family can nest its variants, as in
```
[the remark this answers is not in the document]
egonSchiele: Reworded in 5e2aeb17f. The header now states the actual rule (AGENT_NAME_PATTERN plus the empty/./.. segment checks) instead of the "no characters a URL would encode" claim, and the dot-segment folding rationale is scoped to URL paths, where it actually applies; the query-parameter context no longer carries that justification.

packages/agency-lang/docs/dev/evals/eval-tracking.md
egonSchiele: Audited the doc changes and every comment this PR added; the narration is gone in 5e2aeb17f. Removed: the agency.ts comment citing the 69-run upload from my debugging session, the "instead of staying silent until the end" clauses in the reportProgress doc comment and eval-tracking.md, a "both as progress and in tests" aside on formatRunLine, and a "still validated" change-relative phrasing in the doc's agent-name section, which now just states the rule and what it preserves. What remains states constraints: the batch link prints only when all runs share one batch id, statelog keys the page by project and batch, and the agent-name rule keeps a name intact wherever a URL carries it.

Issue #800, opened 2026-08-04
rewindFrom does not re-assert the root budget (runs under the checkpoint's guard)


egonSchiele:
## The gap

`rewindFrom` (`lib/runtime/rewind.ts`) restores a checkpoint with
`createExecutionContext` + `restoreState` and calls **neither**
`installRootBudget` **nor** `reinstallRootBudget` (nor `initFreshExecCtx`, which
now installs the root budget for the fresh-run entry points). So a rewound run
installs no re-asserted root budget: it runs under whatever root cost/time guard
the checkpoint carries — the checkpoint's serialized `limit` and `spent` — rather
than a host-authoritative limit.

This is the same class of issue #796 fixed for the stateless resume path, where
`respondToInterrupts` (`lib/runtime/interrupts.ts`) calls `reinstallRootBudget`
after `restoreState` to clamp the restored guard's limit to the host value while
preserving accumulated spend. `rewindFrom` never got the equivalent treatment.

## Why it is low priority (scoping)

- **Not remotely reachable.** The serve HTTP surface exposes `/function`,
  `/node`, and `/resume` only — there is no `/rewind` route. The path that takes
  a client-supplied checkpoint (stateless served resume) goes through
  `respondToInterrupts`, which is already fixed. So a remote client cannot reach
  `rewindFrom` with a crafted checkpoint today.
- **Local trust.** The residual is a local budgeted rewind: `agency run
  --max-cost/--max-time` then a debugger rewind. It would run under the
  checkpoint's limit instead of a re-asserted host `--max-cost`. Since local is
  trusted and that checkpoint's guard was itself installed from the same
  `--max-cost`, the practical exposure is small — but the behavior is
  inconsistent with the resume path.

## Suggested fix

Add a `reinstallRootBudget(execCtx.stateStack, execCtx.budget)` call in
`rewind.ts` after `restoreState`, mirroring `respondToInterrupts`
(`interrupts.ts`). Reuse the existing clamp-limit / preserve-spend semantics of
`reinstallRootBudget` so a rewind gets a host-authoritative ceiling without
resetting accumulated spend. No-op in IPC, as with the other install sites.

Consider whether `rewindFrom` should also reinstall the run-policy handler for
full parity with the resume path (`installRunPolicyHandler`).

## Context

Discovered during review of #799 (which installs the root budget on the
served-**function** path). #799 correctly targets the function entry point; this
`rewindFrom` gap is separate and was explicitly left out of that PR.

[...]

PR #987, opened 2026-08-31
Install the root policy handler and budget before any user code runs (#966)

### LINKED ISSUE #966: Top-level 'with approve' runs before the root policy handler exists, bypassing --reject

egonSchiele:
Fixes #966.

## The bug

`initFreshExecCtx` (lib/runtime/node.ts) installed the root policy handler and the root cost/time budget at the end of bootstrap, after global init and top-level callback registration. So when top-level code raised an interrupt, the policy was not in the chain yet. A top-level `with approve` was then the only handler and won:

```
const prompt = read("./prompt.md") with approve   // succeeded under --reject '*'
```

This bypassed both policy sources, because they install at the same site:

1. **CLI flags** (`--reject` / `--approve` / `--policy`, via `AGENCY_RUN_POLICY`).
2. **A serve host** (`InvocationOptions.policy`, added in #978) -- a hosted module could read the host's environment from a top-level initializer before the host policy existed.

## The fix

Install both at the top of `initFreshExecCtx`, before `loadProviderModules` and `__initAllRegistered`, so all user code -- global-init expressions, imported modules' init, and top-level `callback(...)` registration -- runs under the root policy and budget. The steady-state chain is unchanged: the policy handler was already the outermost handler once the node body ran; it is now also present during bootstrap.

The resume path (`respondToInterrupts`) already installed the policy before top-level callback re-registration, so it needed no change. (`rewindFrom` installs neither; that is the pre-existing gap already noted in the `initFreshExecCtx` comment, tracked separately.)

One deliberate behavior change beyond the reject case: a policy `propagate` rule now also applies to a top-level `with approve`, and since global init cannot pause, the raise fails with the existing "Cannot create checkpoint" failure instead of self-approving -- failing closed, matching how an unhandled top-level raise already behaved.

## The subprocess path (investigated, no gap)

A `std::agency` `run()` child never had this hole. `gatherChainOutcome` consults the parent over IPC for every raise, including raises from the child's global init, and a parent reject beats the child's `with approve`. Verified empirically with both a parent in-code handler and parent CLI policy flags before writing any fix; pinned by the new `tests/agency/subprocess/handler-reject-child-top-level` test. (Note the child-side policy installer no-ops in IPC mode by design -- the policy lives at the root.)

## Tests

- `tests/agency-js/policy-top-level` -- a top-level `env(...) with approve` is rejected by the env policy (CLI surface) and by `InvocationOptions.policy` (host surface), and still self-approves when no policy is set. Failed before the fix on both surfaces.
- `tests/agency/subprocess/handler-reject-child-top-level` -- parent handler rejects a child top-level `with approve`.
- Ran: the policy/budget unit tests (runPolicyHandler, rootBudget, budgetExit, cliInterruptResolution, invocationOptions, adapter.perInvocation, budgetRun.spawn -- 88 passing), the policy agency-js tests (run-policy, run-policy-nested, serve-policy, cli-policy-handler x2, test-cli-policy, test-cli-agency-only, contained-filenames-policy), subprocess handler-approve/reject, `pnpm run typecheck`, `pnpm run fmt:ts`, `pnpm run lint:structure`.

Docs updated: `test-cli-sandbox.md` (its "cannot raise an interrupt at all" claim is now the corrected version the security roadmap asked for), `with-approve.md` (the empty `liveGuardIds` justification), `security/roadmap.md` (B1 marked fixed).

https://claude.ai/code/session_01HC9cryYEHEwWXscrBhPKjB

### DIFF
--- a/packages/agency-lang/docs/dev/cli/test-cli-sandbox.md
+++ b/packages/agency-lang/docs/dev/cli/test-cli-sandbox.md
@@ -38,10 +38,12 @@ under each name it is given.
   "reject > propagate > approve > noResponse"; pinned by
   `tests/agency/connector-core.agency`), so an inline `with approve` in the
   tested code is a vote that loses.
-- Code that runs before the handler exists (static initializers, top-level
-  callbacks) cannot raise an interrupt at all: there is no node frame to
-  checkpoint, and the runtime throws "Cannot create checkpoint: no current
-  node id". An effectful initializer fails rather than runs.
+- Top-level code (static initializers, top-level callbacks) runs under the
+  same root handler: it installs before any user code, including global init
+  (#966, `initFreshExecCtx` in `lib/runtime/node.ts`). A top-level raise the
+  chain does not settle still fails rather than pausing — there is no node
+  frame to checkpoint, and the runtime throws "Cannot create checkpoint: no
+  current node id".
 - A `std::run` subprocess the tested code launches forwards its interrupts
   to this root chain and installs no policy of its own.
 - What is not covered: CPU inside the time limit, and reads of files the
--- a/packages/agency-lang/docs/dev/cli/test-cli-sandbox.md
+++ b/packages/agency-lang/docs/dev/cli/test-cli-sandbox.md
@@ -39,11 +39,11 @@ under each name it is given.
   `tests/agency/connector-core.agency`), so an inline `with approve` in the
   tested code is a vote that loses.
 - Top-level code (static initializers, top-level callbacks) runs under the
-  same root handler: it installs before any user code, including global init
-  (#966, `initFreshExecCtx` in `lib/runtime/node.ts`). A top-level raise the
-  chain does not settle still fails rather than pausing — there is no node
-  frame to checkpoint, and the runtime throws "Cannot create checkpoint: no
-  current node id".
+  same root handler, which installs before any user code (#966,
+  `initFreshExecCtx` in `lib/runtime/node.ts`). A top-level raise the chain
+  does not settle fails instead of pausing: there is no node frame to
+  checkpoint, and the runtime throws "Cannot create checkpoint: no current
+  node id".
 - A `std::run` subprocess the tested code launches forwards its interrupts
   to this root chain and installs no policy of its own.
 - What is not covered: CPU inside the time limit, and reads of files the
--- a/packages/agency-lang/docs/dev/security/roadmap.md
+++ b/packages/agency-lang/docs/dev/security/roadmap.md
@@ -112,19 +112,23 @@ No issue yet.
 
 ## B. The runner's handler is outermost and always present
 
-### B1. Top-level `with approve` runs before the root policy exists (#966)
+### B1. Top-level `with approve` runs before the root policy exists (#966) — FIXED
 
 <https://github.com/egonSchiele/agency-lang/issues/966>. `lib/runtime/node.ts`
-runs global init and imported modules' top-level code, then installs the
-policy handler. A top-level `const x = read(...) with approve` therefore has
-only the author's approve in the chain and succeeds under `--reject '*'`.
-The per-invocation host policy (`InvocationOptions.policy`, B2) installs at
-the same bootstrap site, so it shares this gap: a hosted module's top-level
-initializer runs before the host policy exists.
-Fix: install the root policy handler and root budget before any user code,
-including `__initAllRegistered`; check the resume path in `interrupts.ts`
-for the same ordering; add an agency-js test that a top-level read is
-rejected.
+used to run global init and imported modules' top-level code, then install
+the policy handler. A top-level `const x = read(...) with approve` therefore
+had only the author's approve in the chain and succeeded under `--reject '*'`.
+The per-invocation host policy (`InvocationOptions.policy`, B2) installed at
+the same bootstrap site, so it shared the gap.
+Fixed: `initFreshExecCtx` installs the root policy handler and root budget
+first, before any user code, including `__initAllRegistered`. The resume path
+(`respondToInterrupts` in `interrupts.ts`) already installed the policy before
+top-level callback re-registration. A `std::agency` subprocess never had the
+gap: child global init consults the parent's chain over IPC like any other
+raise, so the parent's policy or handlers decide it
+(`tests/agency/subprocess/handler-reject-child-top-level.agency`).
+Test: `tests/agency-js/policy-top-level` pins both policy sources against a
+top-level `with approve`.
 
 ### B2. The resume leg: raise-time policy shipped; checkpoint integrity open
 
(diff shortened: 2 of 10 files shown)


packages/agency-lang/docs/dev/cli/test-cli-sandbox.md
```diff
@@ -38,10 +38,12 @@ under each name it is given.
   "reject > propagate > approve > noResponse"; pinned by
   `tests/agency/connector-core.agency`), so an inline `with approve` in the
   tested code is a vote that loses.
-- Code that runs before the handler exists (static initializers, top-level
```
egonSchiele, line 41: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/tests/agency-js/policy-top-level/test.js
```diff
@@ -0,0 +1,31 @@
...
+// 1. CLI surface: the env policy rejects the top-level read despite the
+// program's own `with approve`.
+process.env.AGENCY_RUN_POLICY = JSON.stringify(rejectEnv);
+const cliRes = await readTop();
+const cliRejected = cliRes.data === null;
+
+// 2. Host surface: InvocationOptions.policy rejects it the same way.
+delete process.env.AGENCY_RUN_POLICY;
+const serveOutcome = await __invokeNodeForServe("readTop", {}, { policy: rejectEnv });
+const hostRejected = serveOutcome.status === "returned" && serveOutcome.value.data === null;
```
egonSchiele: **The serve leg has no positive control, so `hostRejected` can pass while `InvocationOptions.policy` plumbing is broken.** `env()` absorbs a denial to null (as the fixture comment says), so `value.data === null` cannot distinguish "the host policy rejected it" from "the serve leg nulls out env reads for an unrelated reason". Check 3 anchors only the direct `readTop()` path; the serve wrapper and its `InvocationOptions -> resolved.policy` resolution — the exact code check 2 exists to pin — is never shown returning the secret. One extra check fixes it:

```js
const plainServe = await __invokeNodeForServe("readTop", {}, {});
const servedWithoutPolicy = plainServe.status === "returned" && plainServe.value.data === "the-secret";
```

packages/agency-lang/docs/dev/security/roadmap.md
```diff
@@ -112,19 +112,23 @@ No issue yet.
 
 ## B. The runner's handler is outermost and always present
 
-### B1. Top-level `with approve` runs before the root policy exists (#966)
+### B1. Top-level `with approve` runs before the root policy exists (#966) — FIXED
```
egonSchiele, line 118: **Marking B1 FIXED leaves the rewind residual invisible in the one doc whose charter is "every gap".** `rewindFrom` (`lib/runtime/rewind.ts`) installs neither the policy handler nor the budget, yet runs user code (`__initAllRegisteredCallbacks`, then the full replay) — the same bypass class as B1, and the new `initFreshExecCtx` comment still names it. The tracking is real (#800, open, priority:high, and its body covers the policy half), but `grep rewind docs/dev/security/roadmap.md` returns nothing. Since the roadmap promises to list every gap and link its issue, this FIXED entry should add one sentence naming the rewind residual and linking #800 (severity is local-only: rewindFrom is reachable only from the debugger, not from a remote checkpoint).

packages/agency-lang/lib/runtime/node.ts
```diff
@@ -134,6 +132,13 @@ async function initFreshExecCtx(
 ): Promise<void> {
   const { initializeGlobals } = opts;
 
+  // Root policy handler (agency run --policy / --reject, or a serve host's
+  // InvocationOptions.policy) and root cost/time budget, installed before any
+  // user code so global-init expressions and top-level `callback(...)` blocks
+  // run under them (see the function comment; #966).
+  installRunPolicyHandler(execCtx, opts.policy);
```
egonSchiele, line 136: **Two caller-site comments in this file still describe the old order** (not in the diff, so flagging here). `node.ts:398-399`: "Handler/budget registration order (inside initFreshExecCtx) is unchanged." — this PR exists to change that order. `node.ts:404-405`: "Cross-module init, this module's globals, top-level callbacks, then the root run-policy handler and root cost/time budget" — recites the pre-fix order this PR reverses; a reader tracing the #966 fix from `runNodeCore` would conclude the hole is still open. Suggest dropping the order recitation in favor of "bootstrapped and capped inside initFreshExecCtx (see its comment for ordering)" and deleting the "order is unchanged" sentence.

egonSchiele:
One more finding on a file this diff does not touch, so it cannot be an inline comment:

**`lib/ir/prettyPrint.ts:377-381` still makes the claim this PR corrected in `with-approve.md`.** The codegen comment reads:

```
// liveGuardIds: [] — an explicit decision, not a default. These
// wrappers register during top-level init, before any guard can
// exist, so "hide every guard" is vacuously correct. (The []
// itself lives in the template.)
```

After this PR, root budget guards do exist during top-level init — the doc edit in `docs/dev/language/with-approve.md` rewrote this exact rationale (guards exist but never raise), but the source comment keeps the old, now-false premise. Same for the pushHandler doc at `lib/runtime/state/context.ts:640-644`, which justifies `[]` with "registered before any guard can exist, like the --policy handler and top-level init wrappers" — true for the policy handler (it installs before `installRootBudget`), no longer true for the top-level init wrappers. Both comments should get the same correction the doc got, so nobody later reasons about `beginSuspension`/`liveGuardIds` from a premise this PR broke.

egonSchiele:
Addressed the review in 0e3ce85f6 + 436a9edab:

- **`agency config` change (Copilot):** the branch was accidentally created off `adit/config-default-show` instead of main, so that commit rode along. Reverted here (no force-push per repo rules); the change still lives on its own branch.
- **Serve positive control (test.js):** added `servedWithoutPolicy` — the serve leg returns the secret with no policy, so `hostRejected`'s null can only mean the InvocationOptions policy rejected it.
- **Rewind residual (roadmap B1):** the FIXED entry now names `rewindFrom` (no policy handler, no budget, runs user code), links #800, and notes it is reachable only from the local debugger.
- **Stale `runNodeCore` comments:** the "registration order is unchanged" sentence is gone and the bootstrap comment now points at `initFreshExecCtx` for ordering instead of reciting the pre-fix order.
- **Comment/doc pass:** re-read every comment and doc line this PR adds and trimmed them to what a future reader needs (the `initFreshExecCtx` ordering constraint, the env-absorb semantics the fixtures rely on, the test-cli-sandbox and with-approve corrections). Cut the emphasis caps and dash constructions on the way.
- **Roadmap:** also picks up the uncommitted sections F and G (and the A1 layer-3 pointer) from the main working tree, as requested.

PR #999, opened 2026-09-01
Add --approve and --reject flags to the agency agent

egonSchiele:
`agency run` has had `--approve`/`--reject` flags; the agent only had `--policy`, so a session-scoped grant ("also let it write", "never let it bash") meant authoring a policy file. The agent now takes the same two flags with the same meaning:

```
agency agent --approve std::write,std::git::commit
agency agent --policy with-writes --reject std::bash
```

## How it works

One function defines what the flags mean. The overlay logic moves from `lib/cli/runPolicy.ts` into `lib/runtime/policyFlags.ts`; `resolveRunPolicy` delegates to it, and `std::policy` exports it to Agency as `policyOverlayFromFlags`, because the agent's flag handling is Agency code (`args.agency` / `policy.agency`). The overlay places blanket rules ahead of the resolved base policy's own rules, reject ahead of approve, so under first-match-wins the flags outrank the base and a reject outranks an approve for the same effect.

On the agent the overlay applies to whichever base the run resolved — the saved policy file, a `--policy` built-in, or a `--policy` path — routes to the session policy path, and is never written to the saved policy file.

This is feature 1 of the scoped-approvals spec; the match syntax, effect-set expansion, and discovery command were descoped after review (`with-writes` already covers the motivating use case).

## Testing

- `lib/runtime/policyFlags.test.ts`: overlay rule order, reject-over-approve, unrelated effects untouched, base policy not mutated, empty flags.
- New agent test `lib/agents/agency-agent/tests/policyFlags.agency` (no LLM): flags overlay onto a policy-file base and a built-in base, decisions via `checkPolicy`, routing to the session path, and no session policy without flags.
- Existing `runPolicy` tests pass unchanged (the run flags' behavior is identical); also ran the turn/readPolicy/execPolicy/envPolicy/gitPolicy agent tests, `pnpm run typecheck`, `pnpm run lint:structure`, and `pnpm run fmt:ts`.

## Docs

A bullet in `docs/dev/agents/approval-policies.md` ("Where policies come from"); `docs/site/stdlib/policy.md` regenerated by `make` from the new docstring.

https://claude.ai/code/session_01Qh4zR96LWmg3kyGbCti4XR

### DIFF
--- a/packages/agency-lang/docs/dev/agents/approval-policies.md
+++ b/packages/agency-lang/docs/dev/agents/approval-policies.md
@@ -16,6 +16,17 @@ surprised us more than once.
   when that fails too.
 - Built-ins like `with-writes` are constructed at launch with the process
   cwd baked into their dir rules.
+- `--approve <effects>` / `--reject <effects>` (both `agency run` and
+  `agency agent`) take comma- or whitespace-separated effect names and
+  overlay blanket rules ahead of whatever base the run resolved — the
+  saved file, a built-in, or a `--policy` path. One function defines what
+  the flags mean: `policyOverlayFromFlags`
+  (`lib/runtime/policyFlags.ts`), reached from `resolveRunPolicy` on the
+  run side and through `std::policy` on the agent side. Reject rules sit
+  ahead of approve rules, so under first-match-wins a reject outranks an
+  approve for the same effect. On the agent the overlay is session-only:
+  it routes to the session policy path and is never written to the saved
+  policy file.
 
 In a non-interactive run (`-p`), an effect no rule decides is
 auto-rejected with an explanatory message (`stdlib/policy.agency`) — there
--- a/packages/agency-lang/docs/dev/agents/approval-policies.md
+++ b/packages/agency-lang/docs/dev/agents/approval-policies.md
@@ -19,14 +19,12 @@ surprised us more than once.
 - `--approve <effects>` / `--reject <effects>` (both `agency run` and
   `agency agent`) take comma- or whitespace-separated effect names and
   overlay blanket rules ahead of whatever base the run resolved — the
-  saved file, a built-in, or a `--policy` path. One function defines what
-  the flags mean: `policyOverlayFromFlags`
-  (`lib/runtime/policyFlags.ts`), reached from `resolveRunPolicy` on the
-  run side and through `std::policy` on the agent side. Reject rules sit
-  ahead of approve rules, so under first-match-wins a reject outranks an
-  approve for the same effect. On the agent the overlay is session-only:
-  it routes to the session policy path and is never written to the saved
-  policy file.
+  saved file, a built-in, or a `--policy` path — with reject rules ahead
+  of approve rules, so under first-match-wins the flags outrank the base
+  and a reject outranks an approve for the same effect. On the agent the
+  overlay is session-only; the saved policy file is never written. Both
+  commands build the overlay with `policyOverlayFromFlags`
+  (`lib/runtime/policyFlags.ts`).
 
 In a non-interactive run (`-p`), an effect no rule decides is
 auto-rejected with an explanatory message (`stdlib/policy.agency`) — there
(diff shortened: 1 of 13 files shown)


packages/agency-lang/docs/dev/agents/approval-policies.md
```diff
@@ -16,6 +16,17 @@ surprised us more than once.
   when that fails too.
 - Built-ins like `with-writes` are constructed at launch with the process
   cwd baked into their dir rules.
+- `--approve <effects>` / `--reject <effects>` (both `agency run` and
```
egonSchiele, line 19: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/stdlib/policy.agency
```diff
@@ -134,6 +135,26 @@ export def builtinPolicyNames(): string[] {
   return _builtinPolicyNames()
 }
 
+/**
```
egonSchiele: not sure this is useful to have in the stdlib?

packages/agency-lang/docs/dev/agents/approval-policies.md
egonSchiele: Trimmed in d8fa9a3bc: the doc bullet now carries only the contract (what the flags do, precedence, session-only) with one file pointer, and the new code comments are cut to constraints — the shared-function narration and the VALUE-caps aside are gone.

packages/agency-lang/stdlib/policy.agency
egonSchiele: Agreed — removed in d8fa9a3bc. The agent now imports _policyOverlayFromFlags from agency-lang/stdlib-lib/policy.js directly, the existing pattern for agent-internal helpers (attachments, sessions do the same). stdlib/policy.agency and the generated stdlib docs are untouched by this PR; the function is CLI plumbing, not a user API, and its docstring would have become a tool description.

packages/agency-lang/lib/runtime/policyFlags.ts
```diff
@@ -0,0 +1,43 @@
...
+ *  `agency remote call` reach it through `resolveRunPolicy`, and the agent
+ *  reaches it through `std::policy`'s `policyOverlayFromFlags`. */
+export function policyOverlayFromFlags(
+  approve: string | undefined,
+  reject: string | undefined,
+  base: Policy,
+): Policy {
+  const approved = splitEffects(approve);
+  const rejected = splitEffects(reject);
+  const policy: Policy = { ...base };
```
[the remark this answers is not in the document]
egonSchiele: Fixed in d8fa9a3bc: the clone is now Object.assign(Object.create(null), base), so toString reads undefined instead of the inherited function and __proto__ is an ordinary own key. Test added for both names.

packages/agency-lang/lib/runtime/policyFlags.ts
```diff
@@ -0,0 +1,43 @@
+import type { Policy, PolicyRule } from "./policy.js";
+
+/** Split a flag's effect list on commas and/or whitespace, so
+ *  `--approve "std::read, std::ls"`, `--approve std::read,std::ls`, and
+ *  `--approve "std::read std::ls"` all work. */
+export function splitEffects(list: string | undefined): string[] {
+  if (!list) return [];
```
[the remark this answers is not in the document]
egonSchiele: Fixed in d8fa9a3bc — braces added.

packages/agency-lang/lib/agents/agency-agent/agent.agency
```diff
@@ -88,7 +88,7 @@ node main() {
     printPolicyHelp()
     process.exit(0)
   }
-  setPolicyArg(args.flags.policy ?? "")
+  setPolicyArg(args.flags.policy ?? "", approve: args.flags.approve ?? "", reject: args.flags.reject ?? "")
```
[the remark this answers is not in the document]
egonSchiele: Fixed in d8fa9a3bc — the call is split across lines, one argument per line.

PR #1000, opened 2026-09-01
Accept capability set names in --approve and --reject

egonSchiele:
PR 1 of 2 from the effect-sets spec (`2026-08-31-effect-sets-in-flags-spec.md`, reviewed in its sibling `-REVIEW` file): `--approve` and `--reject` accept capability set names alongside effect names, everywhere the flags exist.

```bash
agency agent --approve FileRead --reject Shell
agency run --approve FileRead,Network foo.agency
```

## How it works

A bare name that exactly matches a built-in set from `std::capabilities` expands to the set's member effects, each getting the blanket rule at that position. Expansion happens inside `policyOverlayFromFlags`, so `agency run`, `agency test`, and the agent gain it at the one shared point. The compiled policy contains only plain effect names — serialization, `checkPolicy`, and the session-policy file are untouched, and a printed policy shows exactly which effects were granted.

**Compatibility, per the spec review:** any other bare name passes through as a plain effect name, exactly as today. Bare effect declarations are legal (`effect confirm { ... }` + `--approve confirm`), and the bare `interrupt("msg")` form raises the effect named `unknown`, so `--reject unknown` is meaningful — neither may start failing. A near-miss of a set name (case-only or one edit away) gets a stderr warning and still passes through. A user effect whose bare name shadows a set is unreachable from the flags; the documented answer is to namespace user effects.

**Membership** comes from a new `lib/runtime/effectSets.ts` that parses `stdlib/capabilities.agency` from the install on first use (lazy — runs with no bare names never read it), pairs each declaration with its doc comment (attachment is normally the preprocessor's job, so the walker does its own pairing), resolves nested sets transitively, and throws loudly if the shipped file fails to parse. The parse was chosen over walking the compiled zod unions or a build-time manifest: the doc comments the upcoming discovery command needs live only in the source, and the declaration stays the single definition. The once-per-process cache is derived from a shipped file, the same footing as the always-scope registry.

The agent's overlay call now catches a failure from the JS import and exits with the message in red, matching its `--policy` error handling.

## Testing

- `lib/runtime/effectSets.test.ts` parses the real shipped file: FileRead's exact members, FileSystem/Notes transitive flattening, composition, every member namespaced, every doc non-empty. These double as drift tests for edits to `capabilities.agency`.
- `lib/runtime/policyFlags.test.ts`: set expansion, an effect rejected alongside its approving set, overlapping-set dedup, the `confirm`/`unknown` passthrough cases, near-miss and case-only warnings, no warning on unrelated names.
- `lib/agents/agency-agent/tests/policyFlags.agency`: `setExpansion` and `bareNamePassthrough` scenarios (6/6 pass, no LLM).
- Also ran the turn agent test, existing runPolicy/policyFlags vitest files, `pnpm run typecheck`, `pnpm run lint:structure`, `pnpm run fmt:ts` after a full `make`.

## Docs

Flag help on `run`, `test`, `remote call`, and the agent mentions set names (the `see: agency effects` pointer waits for PR 2, which adds the command). The `--approve`/`--reject` bullet in `docs/dev/agents/approval-policies.md` documents the expansion, the passthrough rule, and the shadowing case.

https://claude.ai/code/session_01Qh4zR96LWmg3kyGbCti4XR

### DIFF
--- a/packages/agency-lang/docs/dev/agents/approval-policies.md
+++ b/packages/agency-lang/docs/dev/agents/approval-policies.md
@@ -16,15 +16,23 @@ surprised us more than once.
   when that fails too.
 - Built-ins like `with-writes` are constructed at launch with the process
   cwd baked into their dir rules.
-- `--approve <effects>` / `--reject <effects>` (both `agency run` and
-  `agency agent`) take comma- or whitespace-separated effect names and
-  overlay blanket rules ahead of whatever base the run resolved — the
-  saved file, a built-in, or a `--policy` path — with reject rules ahead
-  of approve rules, so under first-match-wins the flags outrank the base
-  and a reject outranks an approve for the same effect. On the agent the
-  overlay is session-only; the saved policy file is never written. Both
-  commands build the overlay with `policyOverlayFromFlags`
-  (`lib/runtime/policyFlags.ts`).
+- `--approve <effects>` / `--reject <effects>` (`agency run`,
+  `agency test`, and `agency agent`) take comma- or whitespace-separated
+  effect names and overlay blanket rules ahead of whatever base the run
+  resolved — the saved file, a built-in, or a `--policy` path — with
+  reject rules ahead of approve rules, so under first-match-wins the
+  flags outrank the base and a reject outranks an approve for the same
+  effect. A bare name that exactly matches a built-in capability set
+  from `std::capabilities` (e.g. `FileRead`) expands to the set's member
+  effects; any other bare name stays a plain effect name, since bare
+  effect declarations are legal and the bare `interrupt("msg")` form
+  raises the effect named `unknown`. A user effect whose bare name
+  shadows a set is unreachable from the flags — namespace user effects.
+  On the agent the overlay is session-only; the saved policy file is
+  never written. Every command builds the overlay with
+  `policyOverlayFromFlags` (`lib/runtime/policyFlags.ts`); the set table
+  is parsed from `stdlib/capabilities.agency`
+  (`lib/runtime/effectSets.ts`).
 
 In a non-interactive run (`-p`), an effect no rule decides is
 auto-rejected with an explanatory message (`stdlib/policy.agency`) — there
--- a/packages/agency-lang/docs/dev/agents/approval-policies.md
+++ b/packages/agency-lang/docs/dev/agents/approval-policies.md
@@ -17,7 +17,8 @@ surprised us more than once.
 - Built-ins like `with-writes` are constructed at launch with the process
   cwd baked into their dir rules.
 - `--approve <effects>` / `--reject <effects>` (`agency run`,
-  `agency test`, and `agency agent`) take comma- or whitespace-separated
+  `agency test`, `agency remote call`, and
+  `agency agent`) take comma- or whitespace-separated
   effect names and overlay blanket rules ahead of whatever base the run
   resolved — the saved file, a built-in, or a `--policy` path — with
   reject rules ahead of approve rules, so under first-match-wins the
--- a/packages/agency-lang/lib/agents/agency-agent/lib/policy.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/lib/policy.agency
@@ -112,10 +112,18 @@ export def getPolicyForAgent(
   }
 
   // --approve / --reject overlay: session-only, never written to the
-  // saved policy.
+  // saved policy. The overlay can throw (set expansion parses the
+  // shipped stdlib/capabilities.agency and refuses a broken install);
+  // surface that like a --policy error instead of letting it propagate.
   if ((approveArg != "" || rejectArg != "") && basePolicy != null) {
-    sessionPolicy = _policyOverlayFromFlags(approveArg, rejectArg, basePolicy)
-    policyFile = SESSION_POLICY_PATH
+    const overlaid = try _policyOverlayFromFlags(approveArg, rejectArg, basePolicy)
+    if (overlaid is failure(f)) {
+      print(color.red("Failed to apply --approve/--reject: ${f}"))
+      process.exit(1)
+    } else {
+      sessionPolicy = overlaid.value
+      policyFile = SESSION_POLICY_PATH
+    }
   }
 
   return {
--- a/packages/agency-lang/lib/runtime/effectSets.ts
+++ b/packages/agency-lang/lib/runtime/effectSets.ts
@@ -2,6 +2,7 @@ import { readFileSync } from "fs";
 import path from "path";
 import { parseAgency } from "../parser.js";
 import { getPackageRoot } from "../importPaths.js";
+import type { AgencyMultiLineComment, TypeAlias, UnionType, VariableType } from "../types.js";
 
 /** One built-in capability set from `stdlib/capabilities.agency`. */
 export type EffectSetInfo = {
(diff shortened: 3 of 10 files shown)


packages/agency-lang/lib/runtime/policyFlags.ts
```diff
@@ -21,13 +22,70 @@ export function splitEffects(list: string | undefined): string[] {
  *  statements.
  *
  *  Shared by `agency run` (via `resolveRunPolicy`) and the agent's flags. */
+/** Expand any built-in capability-set names in a flag's effect list to
+ *  their member effects. A name with `::` is always a plain effect. A
+ *  bare name that matches no set passes through as an effect name — bare
+ *  effect declarations are legal, and the bare `interrupt("msg")` form
+ *  raises the effect named "unknown", so an unmatched bare name must
+ *  keep working; a near-miss of a set name gets a stderr warning. */
+function expandSetNames(names: string[], flag: string): string[] {
```
egonSchiele, line 20: The new helpers landed between `policyOverlayFromFlags`'s doc comment and the function itself: the block ending "Shared by `agency run` (via `resolveRunPolicy`) and the agent's flags" now sits directly above `expandSetNames`'s own doc comment, and `policyOverlayFromFlags` is left with no doc at all. Move the two helpers (with their docs) above that comment block so it re-attaches to the function it describes.

packages/agency-lang/lib/agents/agency-agent/lib/policy.agency
```diff
@@ -112,10 +112,18 @@ export def getPolicyForAgent(
...
-  // saved policy.
+  // saved policy. The overlay can throw (set expansion parses the
+  // shipped stdlib/capabilities.agency and refuses a broken install);
+  // surface that like a --policy error instead of letting it propagate.
   if ((approveArg != "" || rejectArg != "") && basePolicy != null) {
-    sessionPolicy = _policyOverlayFromFlags(approveArg, rejectArg, basePolicy)
-    policyFile = SESSION_POLICY_PATH
+    const overlaid = try _policyOverlayFromFlags(approveArg, rejectArg, basePolicy)
+    if (overlaid is failure(f)) {
+      print(color.red("Failed to apply --approve/--reject: ${f}"))
```
egonSchiele, line 121: With the passthrough design, the only way the overlay throws is a broken install (capabilities file missing or unparseable), so no test can exercise this branch. Worth one manual check — temporarily rename `stdlib/capabilities.agency` and run `agency agent --approve FileRead` — to confirm two things: that `try` around the JS import actually turns the throw into a `failure` instead of crashing the run, and that `${f}` renders the error message rather than `[object Object]`. The neighboring branches print `f.status` and `JSON.stringify(f.error)`, so `f` is not obviously a string here.

packages/agency-lang/lib/runtime/effectSets.ts
```diff
@@ -0,0 +1,125 @@
...
+  type RawSet = { name: string; doc: string; items: any[] };
+  const raw: RawSet[] = [];
+  let pendingDoc = "";
+  for (const node of parsed.result.nodes as any[]) {
+    if (node.type === "multiLineComment") {
+      pendingDoc = node.isDoc && !node.isModuleDoc ? cleanDoc(node.content) : "";
+      continue;
+    }
+    if (node.type === "typeAlias" && node.isEffectSet) {
+      const items = node.aliasedType?.types ?? [];
```
egonSchiele: `node.aliasedType?.types ?? []` silently yields an empty member list when the RHS is not a union: `effectSet X = <*>` parses to a `primitiveType` (see `effectSet.test.ts`), so it would land in the table with zero members instead of reaching the "unexpected shape" error below — that error only fires per-item, and there are no items. The drift test's members-nonempty check would catch it, but only because `<*>` never appears in the shipped file today. Cheaper to refuse at the source: require `aliasedType.type === "unionType" && aliasedType.isEffectSet` and throw the shape error otherwise.

packages/agency-lang/lib/runtime/effectSets.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  // Doc comments are standalone nodes after a bare parse (attachment to
+  // the following declaration is the preprocessor's job, which we don't
+  // run) — pair each doc comment with the effectSet declaration that
+  // follows it.
+  type RawSet = { name: string; doc: string; items: any[] };
+  const raw: RawSet[] = [];
+  let pendingDoc = "";
+  for (const node of parsed.result.nodes as any[]) {
```
egonSchiele: The walk is `any` all the way down, but every shape it reads has a real type in `lib/types.ts` (`TypeAlias`, `AgencyMultiLineComment`, the union member nodes). Typing the loop would let the compiler catch a rename of `aliasName` / `isEffectSet` / `aliasedType` instead of leaving that to the drift test.

egonSchiele (commented):
Reviewed against the spec and its review file. The core checks out: expansion happens before rule construction so reject-ahead-of-approve and the dedup are untouched; the passthrough rule keeps bare effect names (`confirm`, `unknown`) working; `remote call` gets expansion for free through `resolveRunPolicy` (verified the chain: `runCall` → `resolveRemoteDecision` → `resolveRunPolicy` → `policyOverlayFromFlags`, all client-side); the walker's node shapes match the parser tests (`stringLiteralType` / `typeAliasVariable`); and I traced `withinOneEdit` through the insert/delete/replace cases — it's correct. The doc-comment pairing walk correctly skips the `@module` comment, and the null-prototype table plus the `__proto__` test carry the existing prototype-safety through.

Five inline comments: one real defect (a doc comment left dangling in `policyFlags.ts`), one untested error branch worth a manual check, and three smaller ones.

packages/agency-lang/lib/runtime/effectSets.ts
```diff
@@ -0,0 +1,125 @@
+import { readFileSync } from "fs";
+import path from "path";
+import { parseAgency } from "../parser.js";
```
egonSchiele, line 3: This is the first `lib/runtime` → parser edge (nothing else under `lib/runtime` imports the parser today), and it is a static import, so the chain `stdlib-lib/policy.js` → `policyFlags` → `effectSets` → `parseAgency` loads the whole parser module in every compiled program at import time — the laziness skips the file read and parse, not the module load. For `agency run` the parser is resident anyway, but deployed compiled programs and subprocess runs pick up new import-graph weight even when no flag is ever passed. Probably acceptable; flagging so it's a decision rather than an accident.

packages/agency-lang/lib/runtime/policyFlags.ts
```diff
@@ -21,13 +22,70 @@ export function splitEffects(list: string | undefined): string[] {
  *  statements.
  *
  *  Shared by `agency run` (via `resolveRunPolicy`) and the agent's flags. */
+/** Expand any built-in capability-set names in a flag's effect list to
```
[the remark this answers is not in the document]
egonSchiele: Fixed in d0b69d31f — the helpers moved above the overlay doc block, which reattaches to policyOverlayFromFlags and now also mentions set expansion.

packages/agency-lang/docs/dev/agents/approval-policies.md
```diff
@@ -16,15 +16,23 @@ surprised us more than once.
...
-  `agency agent`) take comma- or whitespace-separated effect names and
-  overlay blanket rules ahead of whatever base the run resolved — the
-  saved file, a built-in, or a `--policy` path — with reject rules ahead
-  of approve rules, so under first-match-wins the flags outrank the base
-  and a reject outranks an approve for the same effect. On the agent the
-  overlay is session-only; the saved policy file is never written. Both
-  commands build the overlay with `policyOverlayFromFlags`
-  (`lib/runtime/policyFlags.ts`).
+- `--approve <effects>` / `--reject <effects>` (`agency run`,
+  `agency test`, and `agency agent`) take comma- or whitespace-separated
```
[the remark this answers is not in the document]
egonSchiele: Fixed in d0b69d31f — agency remote call added to the list.

packages/agency-lang/lib/runtime/policyFlags.ts
egonSchiele: Fixed in d0b69d31f (same fix as the Copilot thread above): helpers moved above the doc block, and the overlay doc gained a sentence on set expansion.

packages/agency-lang/lib/agents/agency-agent/lib/policy.agency
egonSchiele: Ran the manual check with stdlib/capabilities.agency temporarily renamed: agency agent --policy minimal --approve FileRead -p exits 1 with the red message: Failed to apply --approve/--reject: ENOENT: no such file or directory, open .../stdlib/capabilities.agency. So try does turn the JS throw into a catchable failure, and ${f} renders the Error message string, not [object Object]. Recorded in the d0b69d31f commit message.

packages/agency-lang/lib/runtime/effectSets.ts
egonSchiele: Fixed in d0b69d31f: the walker now requires aliasedType.type === unionType with isEffectSet and throws the shape error otherwise. I extracted parseEffectSets(source, origin) from the file-reading path so this is testable on snippets — tests cover <*> (throws, no zero-member table entry), an undeclared set reference, and doc pairing.

packages/agency-lang/lib/runtime/effectSets.ts
egonSchiele: Fixed in d0b69d31f — the walk is typed against TypeAlias / AgencyMultiLineComment / UnionType / VariableType, so a rename of aliasName / isEffectSet / aliasedType fails typecheck instead of waiting on the drift test.

packages/agency-lang/lib/runtime/effectSets.ts
egonSchiele: Keeping the static import as a decision, with the blast radius measured: policyFlags is imported only by lib/cli/runPolicy.ts (CLI-only; the parser is resident there anyway) and lib/stdlib/policy.ts, so the parser edge reaches only programs that import std::policy — the core runtime chain (runPolicyHandler via interrupts.ts/node.ts) imports runtime/policy.js and is untouched, and a compiled program that never imports std::policy loads nothing new. Dynamic imports are banned in this repo, so the alternative would be restructuring the call path; measured this narrowly, the weight did not seem to justify it.

egonSchiele (commented):
Re-reviewed d0b69d31f against every round-1 thread. All five fixes check out, and I verified the new typing against `lib/types.ts` rather than taking it on faith:

- **Doc comment**: the overlay doc now sits directly above `policyOverlayFromFlags` with a new sentence on set expansion, and `expandSetNames` keeps its own doc. Pairing restored.
- **`<*>` refusal**: the walker requires `unionType` + `isEffectSet` before reading members, and the extracted `parseEffectSets(source, origin)` made it testable — the new snippet tests cover `<*>`, the undeclared-reference error, and doc pairing. Threading `origin` through the error messages is a nice touch.
- **Typing**: `RawSet.items: VariableType[]` is correct — `VariableType` is the full union including `StringLiteralType` and `TypeAliasVariable`, and `TypeAlias.aliasedType` is non-optional, so the guard-then-cast is sound. (The `as UnionType` cast after the `type !== "unionType"` guard may even be redundant under narrowing, but it's harmless.)
- **Error branch**: the manual ENOENT check answers both questions from my thread — `try` catches the JS throw and `${f}` renders the message.
- **Parser edge**: measured blast radius (CLI + `std::policy` importers only) recorded on the thread; keeping the static import as a deliberate decision is fine by me.

The Copilot doc-list fix (`agency remote call` added) also matches what I traced in round 1 — that command reaches expansion through `resolveRunPolicy`. CI is green across the board. Nothing new to flag; looks ready.

PR #1001, opened 2026-09-01
Learned skills and a wired-in toolbox for the agency agent

egonSchiele:
The agent can now be taught. "Add this as a skill for the explorer agent" saves a reviewed markdown skill under the agent home; "write the research agent a tool that fetches Hacker News" runs the `std::toolbox` pipeline into that agent's toolbox. Both are usable by the target subagent in the same session, and `/skills` / `/toolbox` list what has accumulated. Spec and plan: `2026-08-31-learned-skills-and-toolbox-{spec,plan}.md` (with the plan review applied).

## The stdlib write half (`std::skills`)

`writeSkill(dir, name, description, body)` composes the flat-layout file itself (raw content could extend the frontmatter), validates the kebab-case name instead of rewriting it, refuses duplicates before raising anything, and shows the complete file in a new `std::skills::reviewSkill` interrupt — accept / revise-with-feedback / reject. A revise verdict returns the feedback so the caller redrafts; only an accepted draft is written. `scanSkillsSubdirs` reads a directory of per-agent skill subdirectories with ONE scan interrupt for the root, and `buildSkillsTool` split into scan + the pure `skillsToolFromEntries`, so a catalog can rebuild a tool without rescanning. The `500` glob cap became the named `MAX_SKILL_FILES` at all three call sites.

## Policy: the agent home joins the default read scope

A new `<agent-home>` dir-pattern placeholder (mirroring `<agency>`, with the empty-env-counts-as-unset guard from #469) and one `readScopeRules` rule covering `<agent-home>/skills/**` and `<agent-home>/tools/**`. Without it, every learned-skill read, catalog scan, and `runTool` call (which rescans its tool directory) would prompt interactively and auto-reject headless. Writes are not widened — saves raise ordinary write interrupts on top of the review gate. Documented as a stated trade in `approval-policies.md`.

## The agent side

- `lib/learned.agency`: session catalogs in module globals (not statics — one long-lived CLI run, updated in place on save; the same-session property). Skills load with one scan interrupt; toolboxes with one `listTools` scan per agent directory that exists. Learned tools are renamed `learned_<name>`, which makes collision with any built-in tool impossible by construction; a test pins that no `codeTools` name uses the prefix.
- Coordinator tools `learnSkill(target, ...)` and `writeToolFor(target, ...)`, with a Learning section in both prompts. Revisions flow back as tool replies the coordinator redrafts from.
- All seven subagents pick up `learnedExtrasFor(<wrapper>)` per invocation via `extraTools` (the rewrite stdlib agent gained the parameter — it was the only one without it); the code subagent appends to its local list in both the interactive and one-shot paths.
- `/skills` and `/toolbox` built-ins render the inventory grouped by subagent.

## Docs and evals

`docs/dev/agents/learned-skills-and-toolbox.md` records the design; `agent-brains.md` drops the stale statics-before-the-policy-handler story (that was the hole #966/#987 fixed, not a pattern). One goal-judged eval (`learned-skill`) teaches the explorer a skill mid-session and checks the explorer uses it.

## Tests

25 new agency test nodes across `skills-write`, `skills-subdirs`, `learned`, `learn`, `learnedWiring`, `learnedView` (all no-LLM; the home-writing ones verify their sandbox before touching anything), plus policy/builtinPolicies vitest cases for the placeholder. Locally green: those suites, the pre-existing skills/toolbox/agentTurn/toolWiring regressions, `make`, `pnpm run typecheck`, `fmt:ts`, `lint:structure`.

https://claude.ai/code/session_01HnpWia84urcJmjrhQGDDPx

### DIFF
--- a/packages/agency-lang/docs/dev/agents/agent-brains.md
+++ b/packages/agency-lang/docs/dev/agents/agent-brains.md
@@ -120,18 +120,25 @@ the handler is installed, not from the type.
 
 There is a corollary about module statics. `registry.agency` imports every
 brain, and Agency initializes the whole import closure at startup
-(`docs/dev/compiler/init-topsort.md`), so every registered brain's top-level statics
-run whether or not that brain was selected. Two consequences:
+(`docs/dev/compiler/init-topsort.md`), so every registered brain's top-level
+statics run whether or not that brain was selected. Keep them free of
+side effects beyond the bundled reads (the coordinator's system prompts,
+the code subagent's skill files); `init` only validates that those reads
+succeeded, takes the grounding, and does other policy-neutral setup.
 
-1. Brain statics must do nothing effectful, with one exception: bundled
-   file reads (the coordinator's system prompts, the code subagent's skill
-   files) stay `static const ... with approve`, exactly as before the split.
-2. Those reads must *not* move into `init`. `init` runs inside the policy
-   handler, and an inner `with approve` cannot get past an outer policy, so
-   a read there would prompt the user under `--policy minimal` (which has no
-   `std::read` rule) or fail in a non-interactive run. `init` only validates
-   that the static reads succeeded, takes the grounding, and does other
-   policy-neutral setup.
+A historical note, because an earlier version of this doc said otherwise:
+statics running before the policy handler was a security hole, not a
+pattern — it was fixed by installing the root policy handler before any
+user code runs (#966, #987). Statics are policy-governed like all code.
+Every tool that touches the filesystem raises an interrupt; the
+concession for directory-of-files tools (`skillsDir`, the docs tools) is
+ONE scan interrupt up front whose approval covers the reads that follow,
+never zero interrupts and never one per file. Session-lived mutable
+state belongs in a module global (a top-level `let`), not a static — the
+CLI agent is one long-lived run, so a global lives for the session and
+can be updated in place; the learned catalogs
+(`docs/dev/agents/learned-skills-and-toolbox.md`) are the worked
+example.
 
 ## Adding a brain
 
--- a/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/learn.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/learn.agency
@@ -2,29 +2,23 @@
   @summary The coordinator's two learning tools: teach a subagent a
   skill (learnSkill) or have a tool written into its toolbox
   (writeToolFor). Both validate the target, route the save through the
-  stdlib review pipelines, and record the result in the session
-  catalogs so the target's next invocation carries it.
+  stdlib review pipelines, and invalidate the session catalogs so the
+  target's next invocation rescans and carries the result.
 */
 import { writeSkill } from "std::skills"
 import { writeTool } from "std::toolbox"
 
 import {
   LEARN_TARGETS,
+  invalidateLearned,
   learnedSkillsRoot,
   learnedToolsRoot,
-  recordLearnedSkill,
-  recordLearnedTool,
+  maxLearnedToolName,
  } from "../../lib/learned.agency"
 
 def unknownTarget(target: string): string {
   return "Unknown target '${target}'. Valid targets: ${LEARN_TARGETS.join(", ")}."
 }
-
-// Providers cap tool names at 64 characters, and a learned tool is
-// advertised as `learned_<name>`, so the name itself gets what the
-// prefix leaves. Enforced here, where the prefix is known, rather than
-// in std::toolbox, which has no prefix.
-static const MAX_LEARNED_TOOL_NAME = 56
 
 export def learnSkill(
   target: string,
@@ -58,7 +52,7 @@ export def learnSkill(
   if (value.revise != undefined) {
     return "The user asked for a revision: ${value.revise}. Redraft the skill with that feedback and call learnSkill again."
   }
-  recordLearnedSkill(target, value.saved)
+  invalidateLearned()
   return "Saved. The ${target} agent has the '${value.saved.name}' skill from its next invocation. /skills lists it."
 }
 
@@ -82,13 +76,13 @@ export def writeToolFor(
   if (!LEARN_TARGETS.includes(target)) {
     return unknownTarget(target)
   }
-  if (name.length > MAX_LEARNED_TOOL_NAME) {
-    return "Not saved: the name must be at most ${MAX_LEARNED_TOOL_NAME} characters, because the tool is advertised as learned_${name} and providers cap tool names at 64."
+  if (name.length > maxLearnedToolName()) {
+    return "Not saved: the name must be at most ${maxLearnedToolName()} characters, because the tool is advertised as learned_${name} and providers cap tool names at 64."
   }
   const written = writeTool(name, purpose, request, dir: "${learnedToolsRoot()}/${target}")
   if (written is failure(reason)) {
     return "Not saved: ${reason}"
   }
-  recordLearnedTool(target, written.value)
+  invalidateLearned()
   return "Saved. The ${target} agent can call it as learned_${written.value.name} from its next invocation. /toolbox lists it."
 }
--- a/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/learn.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/learn.agency
@@ -20,6 +20,12 @@ def unknownTarget(target: string): string {
   return "Unknown target '${target}'. Valid targets: ${LEARN_TARGETS.join(", ")}."
 }
 
+// Providers cap tool names at 64 characters, and a learned tool is
+// advertised as `learned_<name>`, so the name itself gets what the
+// prefix leaves. Enforced here, where the prefix is known, rather than
+// in std::toolbox, which has no prefix.
+static const MAX_LEARNED_TOOL_NAME = 56
+
 export def learnSkill(
   target: string,
   name: string,
@@ -75,6 +81,9 @@ export def writeToolFor(
   """
   if (!LEARN_TARGETS.includes(target)) {
     return unknownTarget(target)
+  }
+  if (name.length > MAX_LEARNED_TOOL_NAME) {
+    return "Not saved: the name must be at most ${MAX_LEARNED_TOOL_NAME} characters, because the tool is advertised as learned_${name} and providers cap tool names at 64."
   }
   const written = writeTool(name, purpose, request, dir: "${learnedToolsRoot()}/${target}")
   if (written is failure(reason)) {
--- a/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/learn.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/learn.agency
@@ -1,0 +1,85 @@
+/** @module
+  @summary The coordinator's two learning tools: teach a subagent a
+  skill (learnSkill) or have a tool written into its toolbox
+  (writeToolFor). Both validate the target, route the save through the
+  stdlib review pipelines, and record the result in the session
+  catalogs so the target's next invocation carries it.
+*/
+import { writeSkill } from "std::skills"
+import { writeTool } from "std::toolbox"
+
+import {
+  LEARN_TARGETS,
+  learnedSkillsRoot,
+  learnedToolsRoot,
+  recordLearnedSkill,
+  recordLearnedTool,
+ } from "../../lib/learned.agency"
+
+def unknownTarget(target: string): string {
+  return "Unknown target '${target}'. Valid targets: ${LEARN_TARGETS.join(", ")}."
+}
+
+export def learnSkill(
+  target: string,
+  name: string,
+  description: string,
+  body: string,
+): string {
+  """
+  Teach one subagent a skill it keeps permanently. Draft the skill from
+  what the user asked for, written for the target agent as the reader:
+  `description` is one line telling that agent when to reach for the
+  skill; `body` is the skill itself in markdown. The user reviews the
+  complete file before anything is saved and may accept, ask for a
+  revision (redraft with their feedback and call this again), or
+  reject.
+
+  @param target - Which subagent keeps the skill: code, research, explorer, oracle, review, writing, or rewrite
+  @param name - A short kebab-case skill name (lowercase letters, digits, hyphens)
+  @param description - One line telling the target agent when to use the skill
+  @param body - The skill content, markdown
+  """
+  if (!LEARN_TARGETS.includes(target)) {
+    return unknownTarget(target)
+  }
+  const dir = "${learnedSkillsRoot()}/${target}"
+  const outcome = writeSkill(dir, name, description, body)
+  if (outcome is failure(reason)) {
+    return "Not saved: ${reason}"
+  }
+  const value = outcome.value
+  if (value.revise != undefined) {
+    return "The user asked for a revision: ${value.revise}. Redraft the skill with that feedback and call learnSkill again."
+  }
+  recordLearnedSkill(target, value.saved)
+  return "Saved. The ${target} agent has the '${value.saved.name}' skill from its next invocation. /skills lists it."
+}
+
+export def writeToolFor(
+  target: string,
+  name: string,
+  purpose: string,
+  request: string,
+): string {
+  """
+  Have a reusable tool written into one subagent's toolbox. A coding
+  agent drafts the implementation against the request type, the draft is
+  reviewed, typechecked, and tested, and the user accepts, revises, or
+  rejects it before anything is saved. Takes several model calls.
+
+  @param target - Which subagent gets the tool: code, research, explorer, oracle, review, writing, or rewrite
+  @param name - The tool's name, e.g. hnTopStories
+  @param purpose - What the tool should do, in plain language
+  @param request - The tool's input type as Agency type text, e.g. { topics: string[]; maxItems: number }
+  """
+  if (!LEARN_TARGETS.includes(target)) {
+    return unknownTarget(target)
+  }
+  const written = writeTool(name, purpose, request, dir: "${learnedToolsRoot()}/${target}")
+  if (written is failure(reason)) {
+    return "Not saved: ${reason}"
+  }
+  recordLearnedTool(target, written.value)
+  return "Saved. The ${target} agent can call it as learned_${written.value.name} from its next invocation. /toolbox lists it."
+}
--- a/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/review.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/review.agency
@@ -1,6 +1,8 @@
 import { agencyReviewAgent } from "std::agents/agency/review"
 import { reviewAgent as generalReviewAgent } from "std::agents/review"
 import { Feedback, mergeFeedback, renderFeedback } from "std::agents/lib/feedback"
+
+import { learnedExtrasFor } from "../../../lib/learned.agency"
 import { setMemoryId } from "std::memory"
 import { gitStatus, gitIsRepo } from "std::git"
 import { filter, map } from "std::array"
@@ -25,7 +27,7 @@ export def reviewAgent(userMsg: string, agencyTask: boolean = false): string {
     language, where that reviewer has nothing useful to say.
   """
   if (!agencyTask) {
-    const general = generalReviewAgent(work: userMsg, task: "Review this code.")
+    const general = generalReviewAgent(work: userMsg, task: "Review this code.", extraTools: learnedExtrasFor("review"))
     return renderFeedback(general)
   }
   return renderFeedback(review(userMsg))
@@ -42,7 +44,7 @@ export def review(userMsg: string): Result<Feedback[]> {
     let feedback: Result<Feedback[]> = success([])
 
     for (codeSnippet in snippets) {
-      feedback = mergeFeedback(feedback, agencyReviewAgent(codeSnippet.code))
+      feedback = mergeFeedback(feedback, agencyReviewAgent(codeSnippet.code, extraTools: learnedExtrasFor("review")))
     }
   }
   return feedback
@@ -90,7 +92,7 @@ export def reviewFiles(paths: string[]): Result<Feedback[]> raises <std::read> {
   for (path in paths) {
     const sourceResult = read(filename: path, useAgentCwd: true)
     if (sourceResult is success(source)) {
-      feedback = mergeFeedback(feedback, agencyReviewAgent(source))
+      feedback = mergeFeedback(feedback, agencyReviewAgent(source, extraTools: learnedExtrasFor("review")))
     }
   }
   return feedback
--- a/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/review.agency
+++ b/packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/review.agency
@@ -43,8 +43,9 @@ export def review(userMsg: string): Result<Feedback[]> {
 
     let feedback: Result<Feedback[]> = success([])
 
+    const extras = learnedExtrasFor("review")
     for (codeSnippet in snippets) {
-      feedback = mergeFeedback(feedback, agencyReviewAgent(codeSnippet.code, extraTools: learnedExtrasFor("review")))
+      feedback = mergeFeedback(feedback, agencyReviewAgent(codeSnippet.code, extraTools: extras))
     }
   }
   return feedback
@@ -87,12 +88,15 @@ export def changedAgencyFiles(): string[] raises <std::git::status> {
   each file against the agent cwd; an unreadable path (e.g. a file `git status`
   lists as deleted) is skipped, so a stale path can never crash the review. No
   `task` is passed, so this is deterministic parse+typecheck with no LLM call. */
-export def reviewFiles(paths: string[]): Result<Feedback[]> raises <std::read> {
+export def reviewFiles(paths: string[]): Result<Feedback[]> raises <std::read, std::skills::skillsDir, std::toolbox::scan, std::ls, std::guard> {
+  // Hoisted: the extras cannot change mid-loop, and rebuilding them per
+  // file re-renders the skills tool and rewraps every learned tool.
+  const extras = learnedExtrasFor("review")
   let feedback: Result<Feedback[]> = success([])
   for (path in paths) {
     const sourceResult = read(filename: path, useAgentCwd: true)
     if (sourceResult is success(source)) {
-      feedback = mergeFeedback(feedback, agencyReviewAgent(source, extraTools: learnedExtrasFor("review")))
+      feedback = mergeFeedback(feedback, agencyReviewAgent(source, extraTools: extras))
     }
   }
   return feedback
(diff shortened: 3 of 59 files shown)


packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/review.agency
```diff
@@ -90,7 +92,7 @@ export def reviewFiles(paths: string[]): Result<Feedback[]> raises <std::read> {
   for (path in paths) {
     const sourceResult = read(filename: path, useAgentCwd: true)
     if (sourceResult is success(source)) {
-      feedback = mergeFeedback(feedback, agencyReviewAgent(source))
+      feedback = mergeFeedback(feedback, agencyReviewAgent(source, extraTools: learnedExtrasFor("review")))
```
egonSchiele: **Build break (AG3013):** `reviewFiles` is declared `raises <std::read>`, but the `learnedExtrasFor("review")` call added here transitively raises `std::skills::skillsDir` (via `scanSkillsSubdirs`), `std::toolbox::scan` (via `listTools`), `std::ls`, and `std::guard`. `pnpm run agency typecheck lib/agents/agency-agent/brains/coordinator/subagents/review.agency` fails at this head with four AG3013 errors on `reviewFiles` and the same four on `reviewWrittenFiles` (whose clause is `<std::git::status, std::read>`). Both clauses need widening (or dropping).

packages/agency-lang/lib/agents/agency-agent/lib/learned.agency
```diff
@@ -0,0 +1,194 @@
...
+  """
+  The learned additions to one subagent's tool list: its learned-skills
+  tool (when it has skills) followed by its learned tools.
+  """
+  const skillsTool = learnedSkillsToolFor(agent)
+  const skillsTools = if skillsTool == null then [] else [skillsTool]
+  return [...skillsTools, ...learnedToolsFor(agent)]
+}
+
+export def recordLearnedSkill(agent: string, entry: SkillEntry) {
```
egonSchiele: **Double-count when the catalog loads after a save.** `recordLearnedSkill`/`recordLearnedTool` run after the file is already on disk. If the catalog was never loaded this session (the user's first action is teaching), `loadSkillEntries()`/`loadToolEntries()` lazily scan the directory, pick up the just-saved artifact, and then `entries[agent].push(entry)` adds it a second time. For skills this lists the skill twice in the tool description and `/skills`; for tools it produces two tools both named `learned_<name>`, and `assertUniqueToolNames` in the runtime then throws on the target subagent's next LLM call. The module doc's "nothing ever rescans after a save" only holds when the catalog was loaded before the save. Fix: load the catalog before writing, or skip the push when the scan already contains the entry.

packages/agency-lang/lib/agents/agency-agent/lib/learned.agency
```diff
@@ -0,0 +1,194 @@
...
+export def learnedToolsRoot(): string {
+  return "${learnedRoot()}/tools"
+}
+
+def loadSkillEntries(): Record<string, SkillEntry[]> {
+  const loaded = _skillEntries
+  if (loaded != null) {
+    return loaded
+  }
+  const scanned = scanSkillsSubdirs(learnedSkillsRoot())
```
egonSchiele: **Skills scan: no exists() guard, and a rejected scan is cached as the catalog.** Two asymmetries with `loadToolEntries` (which guards each dir with a raise-free `exists` check and checks `listed is success`):

1. `scanSkillsSubdirs` raises the `std::skills::skillsDir` interrupt as its first statement, so every session raises it on the first subagent invocation even when `~/.agency-agent/skills` does not exist. Under the minimal policy (no `skillsDir` rule — and minimal is also the fallback for an unrecognized policy key) that is a spurious interactive prompt, or a headless rejection, for users who never learned anything.
2. When the interrupt is rejected, the returned `failure(...)` is a plain Failure value, so `_skillEntries = scanned` caches the Failure for the whole session. Indexing it yields undefined, so learned skills silently vanish (never re-scanned), and a later `recordLearnedSkill` pushes entries onto the Failure object, losing the save. This is the rejected-interrupt-Failure-read-as-data hazard.

An `exists(learnedSkillsRoot())` guard plus an `is failure` check (mirroring `loadToolEntries`) fixes both.

packages/agency-lang/stdlib/skills.agency
```diff
@@ -257,6 +281,146 @@ export def skillsDir(
...
+    return failure("the description must be one line and must not contain ---")
+  }
+  if (body.startsWith("---")) {
+    return failure("the body must not start with --- (it would extend the frontmatter)")
+  }
+  const filename = "${name}.md"
+  if (exists(filename, dir)) {
+    return failure("a skill named ${name} already exists in ${dir}")
+  }
+  const content = "---\nname: ${name}\ndescription: ${description}\n---\n\n${body}\n"
```
egonSchiele: **A reviewed description can silently vanish after restart.** The description is spliced raw into YAML frontmatter, but validation only rejects newlines and `---`. The tarsec frontmatter parser tries `flowList` and `quotedScalar` first, so a description like `[Beta] Use when refactoring` or `"Best practice" guide for X` makes the whole frontmatter block fail to parse (verified against tarsec 0.5.3: `PARSE FAIL: expected ---`). The in-session catalog (from `recordLearnedSkill`) shows the reviewed description, but on the next session's scan the `catch {}` fallback kicks in and `flatEntry` lists the skill as `name: "foo.md"`, `description: ""` — the model gets no hint when to use the skill, and `[a, b]`-shaped descriptions come back as arrays. Either reject `[`/quote-leading descriptions here, or quote the value when composing the frontmatter.

packages/agency-lang/lib/runtime/builtinPolicies.ts
```diff
@@ -43,6 +44,11 @@ export function readScopeRules(): PolicyRule[] {
   return [
     { match: { dir: "{.,./**}" }, action: "approve" },
     { match: { dir: `{${INSTALL}/stdlib/**,${INSTALL}/dist/**}` }, action: "approve" },
+    // The agent home's learned directories: skills the user taught the
+    // agent and tools it wrote, both saved only through review
+    // interrupts. Reading them back (including runTool's per-run scan)
+    // is inside the default scope; a stricter custom policy overrides.
+    { match: { dir: `{${AGENT_HOME}/skills/**,${AGENT_HOME}/tools/**}` }, action: "approve" },
```
egonSchiele, line 51: **The widening reaches non-agent runs, and the `recommended` description was not updated.** This rule lives in the shared `readScopeRules()`, so any third-party script run with `agency run --policy recommended` (or `with-writes`) can now read and `std::toolbox::scan` everything under `~/.agency-agent/skills/**` and `tools/**` without a prompt. The dev docs justify the widening with "the directories are the agent's own home" — which holds for the agent, not for an arbitrary script that can exfiltrate user-taught content through an also-auto-approved channel. Also, the user-facing `BUILTIN_POLICIES` description of `recommended` still says only "reads under the current directory (and the agency install's own docs and skills)". Either move the rule into the agent's policy assembly (`lib/agents/agency-agent/lib/policy.agency` layers onto the resolved base, and the subprocess inherits the parent handler chain, so `runTool` scans stay covered), or at minimum update the `recommended` description to disclose the agent-home scope.

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/review.agency
```diff
@@ -42,7 +44,7 @@ export def review(userMsg: string): Result<Feedback[]> {
     let feedback: Result<Feedback[]> = success([])
 
     for (codeSnippet in snippets) {
-      feedback = mergeFeedback(feedback, agencyReviewAgent(codeSnippet.code))
+      feedback = mergeFeedback(feedback, agencyReviewAgent(codeSnippet.code, extraTools: learnedExtrasFor("review")))
```
egonSchiele: **Hoist `learnedExtrasFor("review")` out of the loops.** It is called inside the per-snippet loop here, inside the per-file loop in `reviewFiles`, and once more in `reviewAgent` — rebuilding the identical extras (skills-tool description render plus `partial().describe().rename()` over every learned tool) once per iteration, though the catalog cannot change mid-loop. `const extras = learnedExtrasFor("review")` once per function, passed in, does the same job.

packages/agency-lang/lib/runtime/policy.ts
```diff
@@ -190,7 +215,13 @@ function matchesRule(
       !viaDot &&
       key === "dir" &&
       picomatch.isMatch(stripDotSlash(value), expandAgencyInstallDir(pattern));
-    if (!raw && !viaDot && !viaInstall) {
+    const viaAgentHome =
```
egonSchiele: **Fourth copy of the placeholder-match clause.** `viaAgentHome` clones the `viaInstall` block, extending the `!raw && !viaDot && !viaInstall && ...` chain that every future placeholder must also thread correctly. Since each expander is a no-op when its token is absent, the two placeholder checks collapse into one: `picomatch.isMatch(stripDotSlash(value), expandAgentHomeDir(expandAgencyInstallDir(pattern)))`. One boolean instead of two, the guard stops growing, and a pattern mixing `<agency>` and `<agent-home>` in one brace group starts working instead of only matching by accident of brace-branch structure.

packages/agency-lang/stdlib/skills.agency
```diff
@@ -257,6 +281,146 @@ export def skillsDir(
   return buildSkillsTool(dir, layout, name)
 }
 
+export def scanSkillsSubdirs(root: string): Record<string, SkillEntry[]> {
```
egonSchiele: **Second copy of the flat-skill scan, and the 500-file cap now spans the whole root.** `scanSkillsSubdirs` + `scannedEntry` re-implement the glob → `_read` → `frontmatter` → `flatEntry` pipeline that `buildSkillsTool` in this same file already runs; a change to frontmatter fallback or entry shape must now be made twice, and a miss makes `skillsDir` and the learned catalog describe the same file differently. A shared `scanFlatEntries(dir, pattern)` helper both call would keep one copy. Side effect worth noting: `MAX_SKILL_FILES = 500` here caps the single glob over the entire root across all subagents (it was previously per directory), so once total learned skills pass 500, later subagents' skills are silently dropped from every catalog scan.

packages/agency-lang/stdlib/skills.agency
```diff
@@ -257,6 +291,142 @@ export def skillsDir(
...
+  @param root - The directory holding the subdirectories
+  @param subdirs - The subdirectory names to scan
+  """
+  return interrupt std::skills::skillsDir("Are you sure you want to scan this directory for skill files?", {
+    dir: root,
+    layout: "flat"
+  })
+
+  let out: Record<string, SkillEntry[]> = {}
+  for (sub in subdirs) {
```
egonSchiele, line 327: Security: `scanSkillsSubdirs` interpolates each caller-supplied `sub` into the scan path with no validation, so a name containing `..` or `/` reads outside the single approved root. The one `std::skills::skillsDir` interrupt is `@alwaysUnder(dir)`-scoped to `root`, but `scanEntries` then uses the interrupt-free primitives `_glob`/`_read` — and with `allowedPaths: []`, `assertContained` enforces nothing. A caller passing `subdirs: ["../../../home/user/notes"]` (any Agency program, or an LLM invoking this exported `std::skills` function as a tool) globs and reads `*.md` far outside `root`, feeding their frontmatter and contents into an LLM-visible tool description — despite the user only ever approving a scan of `root`. The bundled agent is safe because it passes the fixed `LEARN_TARGETS`, but the exported API is not. Fix: reject `sub` names containing `/`, `\`, or `..`, or pass `[root]` as allowedPaths so reads are contained.

packages/agency-lang/lib/runtime/builtinPolicies.ts
```diff
@@ -43,6 +44,25 @@ export function readScopeRules(): PolicyRule[] {
...
+
+// runTool records a use count in the tool's meta.json after every run.
+// Without this rule each learned-tool invocation prompts on that write —
+// and a headless run executes the tool, then reports failure when the
+// metadata write is rejected. Scoped to exactly that file under the
+// agent-home toolboxes; every other write still prompts.
+export function toolboxMetaWriteRules(): PolicyRule[] {
+  return [
+    {
+      match: { dir: `${AGENT_HOME}/tools/**`, filename: "meta.json" },
```
egonSchiele: Security: this rule auto-approves an arbitrary-content `std::write` to *any* file named `meta.json` under *any* tool directory in the agent home, not just runTool updating its own use count. Policy matching (`matchesRule`) checks only `dir`+`filename` globs — never the content or which code raised the write — and subprocess effects re-enter the parent policy (`docs/dev/runtime/subprocess-ipc.md`). So a prompt-injected code-agent action, or one learned-tool subprocess, can silently rewrite another learned tool`s `meta.json`. That file controls the model-visible tool description (`learnedToolFromEntry`: `"${entry.meta.purpose}\n\nRequest type: ${entry.meta.request}"`) — direct prompt injection into every future offering of that tool — and `runTool` reads `maxTime` straight from meta for its wall-clock budget (validated only at `writeTool` time). The dev doc discloses this as approving "exactly that file" for the use-count write; the actual match is far wider. Fix: raise a dedicated effect (e.g. `std::toolbox::recordUse`) the policy approves, or scope the write to the running tool`s own directory.

packages/agency-lang/lib/runtime/policy.ts
```diff
@@ -158,6 +160,32 @@ export function expandAgencyInstallDir(
...
+// Expanded at match time, like `<agency>`. An empty AGENCY_AGENT_HOME
+// counts as unset — the same guard the agent's config module documents
+// (issue #469) — so a set-but-empty var never turns the home into a
+// cwd-relative path. Exported for tests, which inject the home.
+export function expandAgentHomeDir(pattern: string, home: () => string = defaultAgentHome): string {
+  if (!pattern.includes(AGENT_HOME_PLACEHOLDER)) return pattern;
+  return pattern.split(AGENT_HOME_PLACEHOLDER).join(escapeForGlob(home()));
+}
+
+function defaultAgentHome(): string {
```
egonSchiele, line 178: `defaultAgentHome` does `path.resolve`/`path.join` but never `realpath`s the home — yet file effects put `fs.realpathSync(dir)` in every `std::read`/`std::write` interrupt payload (`prepareContainedPath`), and `resolveDotDirPattern` realpaths the cwd for exactly this reason (see the comment at policy.ts:116). With `AGENCY_AGENT_HOME=/tmp/agent-home` on macOS (`/tmp` -> `/private/tmp`) — the very value the new tests use — the pattern expands to `/tmp/agent-home/skills/**` while the interrupt carries `/private/tmp/agent-home/...`, so picomatch never matches. Every learned-skill read, catalog scan, and the `runTool` meta.json write then prompts interactively and auto-rejects headless — the precise failure this policy widening exists to prevent. Same for any user whose `$HOME` traverses a symlink. The new tests craft interrupt data by hand, so they never exercise the canonicalized form. Fix: realpath the resolved home (fall back to the lexical path if it does not resolve, as `resolveDotDirPattern` does).

packages/agency-lang/stdlib/skills.agency
```diff
@@ -257,6 +291,142 @@ export def skillsDir(
...
+  }
+  const filename = "${name}.md"
+  if (exists(filename, dir)) {
+    return failure("a skill named ${name} already exists in ${dir}")
+  }
+  // The description is double-quoted so YAML flow syntax in it (a
+  // leading `[` or quote) cannot make the frontmatter unparseable and
+  // silently drop the reviewed description on the next scan. The
+  // no-double-quotes check above is what makes the quoting safe.
+  const content = "---\nname: ${name}\ndescription: \"${description}\"\n---\n\n${body}\n"
```
egonSchiele: The name is written unquoted (`name: ${name}`) while `validSkillName` (`/^[a-z0-9][a-z0-9-]*$/`) accepts all-digit names like `2024`. On the next session`s scan the frontmatter parser coerces the bare numeric to a JS number, and `renderEntry` -> `xmlEscape` calls `.replaceAll` on it, which the runtime throws on ("Cannot call non-function value at property replaceAll"). Agency`s auto-try turns that into a propagating Failure, so `learnedExtrasFor(<agent>)` fails on every invocation of that subagent (the poisoned entry is cached for the whole session) until the file is hand-deleted. Non-crashing variants like `007` also silently become `7`, so the advertised `<name>` no longer matches the filename. The description is already protected by double-quoting; the name is not. Fix: quote the name in the frontmatter too, or reject all-digit names in `validSkillName`.

packages/agency-lang/lib/agents/agency-agent/lib/learned.agency
```diff
@@ -0,0 +1,220 @@
...
+  }
+  return skillsToolFromEntries("${learnedSkillsRoot()}/${agent}", entries, "learned_skills_${agent}")
+}
+
+// One catalog entry as an LLM-callable tool. The `learned_` prefix is
+// what makes collision with a target's built-in tools impossible.
+def learnedToolFromEntry(agent: string, entry: ToolEntry): any {
+  const description = "${entry.meta.purpose}\n\nRequest type: ${entry.meta.request}"
+  const dir = "${learnedToolsRoot()}/${agent}"
+  return runTool.partial(name: entry.name, dir: dir).describe(description).rename("learned_${entry.name}")
```
egonSchiele: `learnedToolFromEntry` renames to `learned_${entry.name}` with no length cap or sanitization (`.rename()` in agencyFunction.ts does not validate). The 56-char cap lives only in the coordinator`s `writeToolFor`. A tool that lands in `<agent-home>/tools/<agent>/` any other way — a user script calling `std::toolbox` `writeTool` directly (its own cap is 64), or a tool created before the cap — passes `listTools` (which never checks names read off disk; only `broken` from meta/module parse is surfaced) and is advertised as `learned_<name>` at 65-72 chars. Anthropic`s tool-name pattern is `{1,64}`, so every LLM call of that subagent 400s until the directory is fixed, identified only by index in the error. Fix: cap/sanitize in `learnedToolFromEntry` (or mark the entry `broken`) rather than trusting the disk name.

packages/agency-lang/lib/agents/agency-agent/lib/learned.agency
```diff
@@ -0,0 +1,220 @@
...
+}
+
+def loadSkillEntries(): Record<string, SkillEntry[]> {
+  const loaded = _skillEntries
+  if (loaded != null) {
+    return loaded
+  }
+  // No skills root yet (nothing ever learned): no scan, no interrupt.
+  if (!exists(learnedSkillsRoot())) {
+    _skillEntries = {}
```
egonSchiele, line 73: Both early branches here cache one `{}` but `return {}` — a *different* fresh literal (contrast `loadToolEntries`, which caches and returns the same `scanned`). If the session`s first catalog load happens inside `recordLearnedSkill` and the scan fails, `recordLearnedSkill` pushes onto the returned non-cached object while `learnedSkillsToolFor` reads the cached empty one all session. Trigger: `learnSkill` writes the file (review + write approved), then the follow-up `std::skills::skillsDir` scan is rejected -> the failure branch runs; the coordinator replies "Saved... from its next invocation" but the skill never appears until next session. Not reachable under `recommended` (which blanket-approves `skillsDir`), but real under `minimal`/a custom reject rule or interactive rejection — the exact "record push silently vanishes" the failure-branch comment says it is avoiding. Fix: cache and return the same object.

packages/agency-lang/lib/agents/agency-agent/lib/learned.agency
```diff
@@ -0,0 +1,220 @@
...
+  next invocation lists it without a rescan. A no-op when the catalog
+  already holds the entry: when the save was the session's first learned
+  action, the lazy load above scans AFTER the file hit disk and has
+  already picked it up.
+  """
+  const entries = loadSkillEntries()
+  if (entries[agent] == undefined) {
+    entries[agent] = []
+  }
+  const existing = filter(entries[agent], \known -> known.location == entry.location)
```
egonSchiele: `recordLearnedSkill` (and `recordLearnedTool` at line 186) dedupes by `location`/`name` and returns without replacing an existing entry, so a same-session delete-and-reteach leaves the stale entry advertised. Flow: the user deletes the explorer`s `repo-orientation.md` (the documented "ordinary file operation" remove path), then re-teaches it with a better body; `writeSkill`s `exists` check passes since the file is gone, disk is updated, and the reply says "Saved." — but the catalog keeps the old entry by matching `location == "repo-orientation.md"`, so the explorer`s skills tool serves the old description for the rest of the session. Fix: replace the matching entry instead of skipping when one already exists.

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/learn.agency
```diff
@@ -0,0 +1,94 @@
...
+
+def unknownTarget(target: string): string {
+  return "Unknown target '${target}'. Valid targets: ${LEARN_TARGETS.join(", ")}."
+}
+
+// Providers cap tool names at 64 characters, and a learned tool is
+// advertised as `learned_<name>`, so the name itself gets what the
+// prefix leaves. Enforced here, where the prefix is known, rather than
+// in std::toolbox, which has no prefix.
+static const MAX_LEARNED_TOOL_NAME = 56
```
egonSchiele: Altitude / drift: `MAX_LEARNED_TOOL_NAME = 56` hard-codes "64 minus the length of `learned_`", but that 64 cap already exists as `MAX_TOOL_NAME_LEN` in both stdlib/toolbox.agency and stdlib/skills.agency, and the `learned_` prefix is spelled out separately in `learnedToolFromEntry` (learned.agency:128) and its user messages. The subagent set is likewise duplicated — `LEARN_TARGETS` (learned.agency:23), `mainAgentTools`, and the two docstring target lists in this file. Adding a subagent or changing the cap/prefix means several uncoordinated edits with nothing tying them together, and this derived `56` can silently drift from what `learnedToolFromEntry` actually renames to. Consider deriving the cap as `MAX_TOOL_NAME_LEN - len(prefix)` from one exported prefix constant, and driving the target list from a single subagent registry.

egonSchiele:
Pushed the restructure from the three-root diagnosis (0c4351b1b), plus the real Root 2 fix now that tarsec 0.5.4 ships `stringifyFrontmatter`.

**Root 1 — catalogs no longer update in place.** `recordLearnedSkill`/`recordLearnedTool` are deleted. A save calls `invalidateLearned()` and the next access rescans, so the fresh-literal-vs-cached bug, the delete-and-reteach stale entry, and the double-count cannot exist. Tests reworked accordingly (`invalidateForcesRescan` pins two scans around an invalidate), and the tool tests now run against a real on-disk fixture toolbox instead of injected entries.

**Root 2 — the serializer, not a stopgap.** The tarsec pin is bumped to 0.5.4 and `writeSkill` composes frontmatter with `stringifyFrontmatter`, whose round-trip contract is property-tested in tarsec next to the parser. The quote/`---` rejections are gone (only the semantic one-line check remains); `trickyDescriptionsRoundTrip` covers quotes, a literal backslash-n, `2024`, and padded whitespace, and `unrepresentableDescriptionFailsCleanly` pins the loud-refusal path.

**Root 3 — the meta.json write rule is reverted.** As flagged, a dir+filename glob cannot tell `runTool`'s bookkeeping from arbitrary content landing in `meta.json`. The use-count update now raises its own `@alwaysUnder(dir)` effect, `std::toolbox::recordUse`, approved under `recommended` for `<agent-home>/tools/**`, and it is best-effort — a declined record never fails a run that succeeded (`rejectedRecordUseDoesNotFailTheRun`). This also fixes the `runPolicy.test.ts` CI failure.

Smaller review items in the same push:

- `scanSkillsSubdirs` validates every subdir as a bare path segment before raising its interrupt, so `../x` can never ride the root's approval outside the root (`traversalSubdirFailsBeforeTheInterrupt`).
- The agent home is realpathed the same way the cwd is in `resolveDotDirPattern`, so a symlinked home matches canonicalized payload dirs (new policy test with a real symlink).
- Learned tools whose name overflows the provider cap behind `learned_` are excluded at load and flagged in `/toolbox`; the cap is derived from the now-exported `MAX_TOOL_NAME_LEN` instead of a restated 56.
- On the `broken: null` vs `!= undefined` thread: in Agency `null` and `undefined` compare equal by design (`docs/dev/language/null-and-undefined.md`), and the new fixture test exercises the real `listTools` path — entries carry `broken: null` and are offered — so no change was needed there.
- On the relative-`expandRoot` thread: `expandPath` only expands `~`, so a relative toolbox dir does stay relative in scan payloads. It cannot match the `<agent-home>` rules, which means it falls through to a prompt — failing closed — and the agent itself always passes absolute roots, so I left it as is.

Verified locally: `make`, the skills/learned/learn/learnedWiring/agentTurn agency suites (46 + 30 nodes), the toolbox suite (37 nodes incl. the new best-effort node), the four policy vitest files (77 tests, `runPolicy.test.ts` green), typecheck, prettier, and the structural linter.

packages/agency-lang/docs/dev/agents/agent-brains.md
```diff
@@ -120,18 +120,25 @@ the handler is installed, not from the type.
...
-1. Brain statics must do nothing effectful, with one exception: bundled
-   file reads (the coordinator's system prompts, the code subagent's skill
-   files) stay `static const ... with approve`, exactly as before the split.
-2. Those reads must *not* move into `init`. `init` runs inside the policy
-   handler, and an inner `with approve` cannot get past an outer policy, so
-   a read there would prompt the user under `--policy minimal` (which has no
-   `std::read` rule) or fail in a non-interactive run. `init` only validates
-   that the static reads succeeded, takes the grounding, and does other
-   policy-neutral setup.
+(`docs/dev/compiler/init-topsort.md`), so every registered brain's top-level
```
egonSchiele, line 123: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/learn.agency
```diff
@@ -0,0 +1,88 @@
...
+  learnedSkillsRoot,
+  learnedToolsRoot,
+  maxLearnedToolName,
+ } from "../../lib/learned.agency"
+
+def unknownTarget(target: string): string {
+  return "Unknown target '${target}'. Valid targets: ${LEARN_TARGETS.join(", ")}."
+}
+
+export def learnSkill(
```
egonSchiele, line 23: better way to implement the revise flow would be to interrupt and ask for user feedback (see how the writeTool func does this)

packages/agency-lang/stdlib/skills.agency
```diff
@@ -257,6 +293,151 @@ export def skillsDir(
...
+  // YAML flow syntax, quotes, or backslashes in the description cannot
+  // silently corrupt or drop the reviewed text. It throws on the rare
+  // value no quoting can hold; `try` turns that into a clean failure.
+  const fm = try _stringifyFrontmatter({ name: name, description: description })
+  if (fm is failure(fmErr)) {
+    return failure("the description cannot be written as frontmatter: ${fmErr}")
+  }
+  const content = "${fm.value}\n${body}\n"
+  const answer = interrupt std::skills::reviewSkill(
+    "Save this skill to ${dir}/${filename}?",
```
egonSchiele, line 419: Change to "Do you want to save this skill?"

PR #1016, opened 2026-09-03
std::grep: translate grep-style flags; add ignoreCase, wholeWord, filesOnly, invert

egonSchiele:
## Why

In one agent trace, five of six tool errors were the grep tool failing on `flags: "n"` or `"gn"`:

```
Invalid flags supplied to RegExp constructor 'n'
```

`std::grep` is an in-process regex walk, not the `grep` program, and the `flags` string went straight into the `RegExp` constructor. The model was thinking in grep flags, the docstring said only "Regex flags", and nothing in between explained the difference.

## What

- **`lib/stdlib/grepQuery.ts`** (new) owns the translation from a call to a compiled search, with a single rule table:
  - `i m s u` pass through to the regex.
  - `n r R g E P` are accepted and dropped. The tool already searches recursively and returns line numbers.
  - `w l v` are rejected with a message naming `wholeWord`, `filesOnly`, or `invert`.
  - Anything else is rejected with the list of accepted letters.
  The message is the tool's failure text, so the model's next call is informed instead of a guess.
- **`grep` gains four named parameters**: `ignoreCase`, `wholeWord`, `filesOnly`, `invert`. They ride in the `std::grep` interrupt data next to `flags`, so approvers and policies see the whole request. Translation happens after the interrupt, so a policy always matches what the caller sent.
- **Docstring** now says the search is recursive, always returns line numbers, and takes JavaScript regex syntax. It is the tool description the model reads.
- `_grep` now takes a `GrepQuery` object and returns either matches or, with `filesOnly`, file paths. A file's trailing newline no longer counts as an empty last line, so `invert` does not report a phantom line.

## Tests

- `lib/stdlib/grepQuery.test.ts`: the rule table and messages (10 cases).
- `tests/agency/stdlib/grep-flags.agency`: end to end over the shared fixtures. A grep-style flag succeeds, a flag with a parameter fails with that parameter's name, and each parameter changes the output (6 cases).
- Existing `grep-basic`, `grep-max-results`, and the `shell.test.ts` symlink cases still pass.

## Docs

- `docs/dev/stdlib/grep-flags.md`, with index pointers in `CLAUDE.md` and the stdlib docs skill.
- `docs/site/stdlib/shell.md` regenerated by `make`.

https://claude.ai/code/session_01ByjGRLbXwjSJdpK4R7JSP1

### DIFF
--- a/packages/agency-lang/docs/dev/stdlib/grep-flags.md
+++ b/packages/agency-lang/docs/dev/stdlib/grep-flags.md
@@ -1,0 +1,57 @@
+# `std::grep`: flags and the named search parameters
+
+`std::grep` does not run the `grep` program. It walks the directory itself
+and tests each line against a JavaScript `RegExp`. That keeps the path
+containment rules (`allowedPaths`, the refusal to follow a symlinked root,
+the skip list for `node_modules` and friends) in our code, and it behaves the
+same on every platform. The cost is that a model calling the tool thinks in
+`grep` flags, and JavaScript regexes know only a handful of letters.
+
+## The problem this solves
+
+Before this, `flags` went straight into the `RegExp` constructor. A model
+asking for `flags: "n"` (line numbers) or `"gn"` got back
+
+```
+Invalid flags supplied to RegExp constructor 'n'
+```
+
+and had to guess its way to an empty flags string. One trace showed five of
+six tool errors in a session were this.
+
+## How a call becomes a search
+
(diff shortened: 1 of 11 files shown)


packages/agency-lang/docs/dev/stdlib/grep-flags.md
```diff
@@ -0,0 +1,49 @@
+# `std::grep`: flags and the named search parameters
```
egonSchiele, line 1: put at packages/agency-lang/docs/dev/stdlib/shell/grep-flags.md instead

PR #1018, opened 2026-09-03
agency agent: fewer approval prompts, tools that say what happened, and a viewer that counts working time

egonSchiele:
Papercuts from one `agency agent` session (trace `mL0SvY`, 2026-09-03). The agent spent most of a five-minute turn at approval prompts, and the trace could not tell us why. Each change below is pinned to what went wrong.

## The agent

- **Coordinator gets `edit`.** It read a file, wrote a brief that was the diff in prose, and paid a `codeAgent` dispatch to re-read the same file. Small changes to a file it has read now happen in its own context; the prompts say when the change still belongs to `codeAgent`.
- **Code agent gets the project context.** Its prompt asked it to read `AGENTS.md`; across two dispatches it never did, and ran `pnpm run fmt:ts` from the wrong directory when that file named the right one. It now puts the same date/cwd/`AGENTS.md` block the harness builds for the coordinator into its own prompt (`projectContextText` in `lib/projectContext.agency`), falling back to `CLAUDE.md` when there is no `AGENTS.md`.
- **Default models.** The Anthropic and OpenAI fast slots now default to the same model as the slow slots (`claude-opus-4-8`, `gpt-5.6-sol`) to test whether the split was costing quality; Google is unchanged. The split mechanism is unchanged; test expectations updated.

## Tools

- **`safeBash` says how a command exited and keeps long output.** Output past the cap used to be cut with a marker, so the model could not see whether `make` had finished and spent seven prompts on `tail`, `sed`, `wc`, and a full re-run to recover the exit code. Long output now goes to `.agency-agent/tool-output/<stamp>.log` under the command's cwd (never through a symlink), and the result is a preview with the exit code, both ends of the output, the exact `read` call to open it, and a note that `read` and `grep` need no approval inside the project while bash does. The docstring and code prompt state the exit contract plainly.
- **`grep` honours `.gitignore`.** A grep over `lib/agents` returned mostly hits in compiled `.js`, which sent the model chasing a stale default. New `respectGitignore` parameter, true by default, with a small matcher (`lib/stdlib/gitignore.ts`: negation, anchoring, directory-only rules, nested files, and the files above the search root up to the repository root). The agent toolkits lock it with partial application so the model never sees the parameter. Documented in `docs/dev/stdlib/grep-flags.md`.
- **`typecheck` points at `typecheckFile`** for files on disk, instead of inviting a pasted file.
- **Attachments survive macOS screenshot names.** Screenshot names carry a narrow no-break space (U+202F) before AM/PM that the terminal does not escape; the media-path scanner split on it and silently attached nothing. It now splits on ASCII whitespace only.

## Policy handler and statelog

- **Rule rejections are printed.** A saved blanket reject on `std::read` refused the agent's read of a build log in 5 ms with nothing on screen; the only sign was the agent falling back to bash. The CLI policy handler now prints one dim line when a rule rejects.

## Logs viewer

- **Working time, not wall clock.** A session's duration was mostly the user: minutes at prompts and gaps between turns. Rows now show `active` and name the rest as `waiting`, read from gaps between the events already logged: a gap ending at a handler decision is a person at a prompt, a gap ending at an aborted tool call is a cancelled prompt, a gap with nothing in flight is time between turns. Nothing new is logged. Doc: `docs/dev/cli/logs-viewer.md`.
- **Dim system bodies, bright cyan user bodies, rendered for real.** The formatter change looked like a no-op for two reasons below it: the TUI style parser knew bold but not dim, and wrapping happened after coloring, so escape bytes counted as columns and continuation rows lost the style. Dim is now understood end to end (parser, `{dim}` tag, cells, ANSI and HTML renderers), and the formatter wraps body text to the width before coloring each row.

## Verification

Typecheck, structural linter, prettier, and `make` pass. Vitest covers the gitignore matcher, the grep walk, the scanner, the style parser, the wrapping formatter, and the viewer tree. The `safeBash`, `models`, and `grounding` Agency tests pass locally; CI runs the rest.

https://claude.ai/code/session_01FPTQRoiQZyBHD7qy1cFWV1


packages/agency-lang/docs/dev/agents/agent-brains.md
```diff
@@ -68,7 +70,10 @@ Paths in this section are relative to `lib/agents/agency-agent/`.
 - `lib/repl.agency`: the interactive terminal. It calls the harness
   `runTurn` and never a brain directly.
 - `lib/grounding.agency`: builds the grounding text with `groundingText`,
-  which calls `loadAgentsMd`.
+  and the project block on its own with `projectContextText`, which calls
+  `loadAgentsMd` (AGENTS.md, else CLAUDE.md). The coordinator hands the
```
egonSchiele: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/docs/dev/stdlib/grep-gitignore.md
```diff
@@ -0,0 +1,68 @@
+# `std::grep` and `.gitignore`
```
egonSchiele: This is a really long file for one arg added to one function

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/coordinator.agency
```diff
@@ -61,6 +69,9 @@ def coordinatorInit(context: BrainContext) {
   }
   _context = context.grounding
   first = true
+  // The code subagent acts in the project, so it gets the project block
+  // (AGENTS.md and cwd) without the session identity above it.
+  setCodeGrounding(context.projectContext)
```
egonSchiele: why not just make a shared function both places can access? Why does the coordinator need to tell the coding agent about this?

packages/agency-lang/lib/agents/agency-agent/lib/models.agency
```diff
@@ -37,18 +37,18 @@ type ProviderModels = {
...
-    fast: "claude-sonnet-5",
+    fast: "claude-opus-4-8",
     slow: "claude-opus-4-8",
     slowProvider: "anthropic"
   },
   google: {
-    fast: "gemini-3.5-flash",
-    slow: "gemini-3.5-flash",
+    fast: "gemini-3.1-pro-preview",
+    slow: "gemini-3.1-pro-preview",
```
egonSchiele: go ahead and leave the google ones as-is for now

packages/agency-lang/lib/logsViewer/conversation.ts
```diff
@@ -58,6 +58,15 @@ function formatMessage(msg: ConvoMessage): string[] {
   return lines;
 }
 
+// System and user message bodies get a distinct color so they stand
+// out from assistant/tool text; the role tag itself (from formatRole)
+// is unaffected.
+function styleBody(role: string, text: string): string {
+  if (role === "system") return color.dim(text);
+  if (role === "user") return color.cyan(text);
```
egonSchiele: lets use `brightCyan` instead

packages/agency-lang/lib/logsViewer/treeRows.ts
```diff
@@ -221,6 +227,71 @@ export function wrapLine(text: string, width: number): string[] {
   return out;
 }
 
+// An ANSI SGR sequence: `ESC[` params `m`. Zero columns wide.
+// eslint-disable-next-line no-control-regex
+const SGR_PATTERN = /\x1b\[[\d;]*m/g;
+const SGR_RESET = "\x1b[0m";
+
+// Walk the styled text alongside the chunks of its plain form, so each
```
egonSchiele: This is pretty ugly code. Also I could swear we already do this elsewhere. How would you write this in a deterministic way?

packages/agency-lang/lib/runtime/interruptResponse.ts
```diff
@@ -7,8 +7,14 @@
 // names, so existing `import { approve, reject } from "./interrupts.js"` keeps
 // working.
 
-export type InterruptApprove = { type: "approve"; value?: any };
-export type InterruptReject = { type: "reject"; value?: any };
+/** Who made a handler's decision, when the handler knows. The runtime only
```
egonSchiele: oof. I don't like the amount of code this has added. Lets remove the "decisionsource" functionality -- its not needed.

packages/agency-lang/lib/stdlib/agentTurn.ts
```diff
@@ -0,0 +1,20 @@
+import { __ctx } from "../runtime/asyncContext.js";
+
+/**
```
egonSchiele: yeah, this is also adding a lot of code for what is essentially a nice to have in the logs. Is there some other way to compute this time? If not, lets remove this functionality.

egonSchiele: There is another way, using only events already logged: a gap between consecutive events that ends at a handler decision is a person answering a prompt, one that ends at an aborted tool call is a cancelled prompt, and one with no tool call in flight (less the model call's own timeTaken) is time between turns. That is waitingTime in tree.ts, about 30 lines and no runtime, stdlib, or statelog changes. agentTurn.ts and the turn events are gone. (ff4eeed7f)

packages/agency-lang/lib/runtime/interruptResponse.ts
egonSchiele: Removed: DecisionSource, the tagging in the policy handler, and the runtime/statelog plumbing. The one-line print on a policy rejection stays. (ff4eeed7f)

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/coordinator.agency
egonSchiele: Done: projectContextText now lives in lib/projectContext.agency (a leaf module, to avoid an init cycle through shared.agency) and the code agent calls it itself. BrainContext.projectContext and setCodeGrounding are gone. (ff4eeed7f)

packages/agency-lang/docs/dev/stdlib/grep-gitignore.md
egonSchiele: Folded into a short section of grep-flags.md; the separate doc and its index entries are gone. (ff4eeed7f)

packages/agency-lang/docs/dev/agents/agent-brains.md
egonSchiele: Went through them. Trace citations and the narrating comments are cut from the code; the dev docs keep only what a future reader needs to change the code. (ff4eeed7f)

packages/agency-lang/lib/agents/agency-agent/lib/models.agency
egonSchiele: Restored gemini-3.5-flash for both slots, with the original comment. (ff4eeed7f)

packages/agency-lang/lib/logsViewer/conversation.ts
egonSchiele: Done. (ff4eeed7f)

packages/agency-lang/lib/logsViewer/treeRows.ts
egonSchiele: Yes, the runs explorer's verdict screen used wrapLine too. Rewritten deterministically: the formatter wraps body text to the width before coloring each row, so no escape-aware re-styling is needed. Plain wrapLine lives in lib/logsViewer/wrapLine.ts, shared by both callers. (ff4eeed7f)

packages/agency-lang/lib/logsViewer/tree.ts
```diff
@@ -287,6 +287,45 @@ function inferSpanLabel(evt: EventEnvelope): string {
...
+        (data.decision === "approve" || data.decision === "reject");
+      const cancelled = data.type === "toolCall" && data.output?.__type === "abortedResult";
+      if (answered || cancelled) {
+        waiting += gap;
+      } else if (inFlight === 0) {
+        const thinking =
+          data.type === "promptCompletion" && typeof data.timeTaken === "number"
+            ? data.timeTaken
+            : 0;
+        waiting += Math.max(0, gap - thinking);
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 1561a9dcb: model calls are taken as intervals from promptCompletion.timeTaken and a gap is charged only for the part outside every interval, so an event from a concurrent branch landing mid-call does not count as a wait. Test added.

packages/agency-lang/lib/logsViewer/treeRows.ts
```diff
@@ -183,8 +180,8 @@ export function llmCallSpanChildren(span: TreeNode, childDepth = 0, cols?: numbe
   const queue = [...toolExecs];
   let lineIdx = 0;
   for (const msg of transcript) {
-    for (const line of formatConversation([msg])) {
-      out.push(...convoLineNodes(`${span.id}:llm:convo`, span, line, lineIdx++, available));
+    for (const line of formatConversation([msg], available)) {
+      out.push(convoLineNode(`${span.id}:llm:convo`, span, line, lineIdx++));
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 1561a9dcb: tool-call rows wrap to the width the same way body text does. Test added.

packages/agency-lang/lib/stdlib/gitignore.ts
```diff
@@ -0,0 +1,126 @@
...
+  return rule.matchesUnder(relativePath);
+}
+
+/** The rules of one .gitignore file, scoped to the directory holding it. */
+export type GitignoreFile = {
+  dir: string;
+  rules: Rule[];
+};
+
+const MATCH_OPTIONS = { dot: true, nobrace: false, nonegate: true };
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 1561a9dcb: nobrace and noextglob are set, and braces, parentheses and bars are escaped before picomatch sees them, so *.{js,ts} matches only that literal name. Test added.

packages/agency-lang/lib/stdlib/spill.ts
```diff
@@ -0,0 +1,93 @@
...
+export function _spillDir(): string {
+  const override = process.env.AGENCY_TOOL_OUTPUT_DIR;
+  if (override !== undefined && override !== "") return override;
+  return path.join(os.homedir(), ".agency-agent", "tool-output");
+}
+
+// A saved file's name: a timestamp, a random suffix, `.log`. The read
+// tools accept only names of this shape, which rules out a slash, a `..`,
+// and a leading dot, so no path can reach them.
+const SPILL_NAME = /^[0-9A-Za-z][0-9A-Za-z_-]*\.log$/;
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2bf5ccca5. The spill directory and its parent are checked with lstat and refused when either is a symlink, the write keeps the wx flag so an existing entry is never opened, and both readers go through readContainedFile, the descriptor-validated reader, so a symlink named like a saved file leads nowhere. Covered by lib/stdlib/spill.test.ts.

packages/agency-lang/lib/stdlib/spill.ts
```diff
@@ -0,0 +1,93 @@
...
+    ignoreCase: false,
+    wholeWord: false,
+    filesOnly: false,
+    invert: false,
+  });
+  const dir = _spillDir();
+  let names: string[];
+  if (filename === "") {
+    try {
+      names = (await readdir(dir)).filter((name) => SPILL_NAME.test(name)).sort();
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2bf5ccca5. grepSpill now takes a required file name; the every-file scan is gone.

packages/agency-lang/stdlib/spill.agency
```diff
@@ -0,0 +1,197 @@
...
+  """
+  let out: string[] = []
+  let used = 0
+  for (line in lines) {
+    const shown = cutLine(line)
+    if (out.length > 0 && used + shown.length > PREVIEW_CHARS) {
+      return out
+    }
+    out.push(shown)
+    used = used + shown.length
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2bf5ccca5. Each line counts its newline against the budget.

packages/agency-lang/lib/stdlib/mediaPathScan.ts
```diff
@@ -51,8 +51,15 @@ const QUOTED_MEDIA_SPAN = /(?:^|(?<=\s))(['"])([^'"\n]+?\.(?:png|jpe?g|gif|webp|
  *  appear in terminal line input. */
 const ESCAPED_SPACE = "\u0000";
 
+/** Token separators are ASCII whitespace only. `\s` would also split on
+ *  U+202F, the narrow no-break space macOS puts before "AM"/"PM" in
+ *  screenshot names (`Screenshot 2026-09-03 at 9.02.01\u202fAM.png`); a
+ *  terminal drag-drop escapes the ordinary spaces in that name but not
+ *  this one, so the path used to break into `9.02.01` and `AM.png`. */
+const ASCII_WHITESPACE = /[ \t\r\n\f\v]+/;
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2bf5ccca5. The separator is now every whitespace character except U+202F, with a test for NBSP-separated tokens.

packages/agency-lang/stdlib/agents/agency/researcher.agency
```diff
@@ -64,7 +59,7 @@ def answer(
   )
 }
 
-export def agencyResearcherAgent(
+export handoff def agencyResearcherAgent(
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2bf5ccca5. answer() pushes the persona through ensureSystemMessage.

packages/agency-lang/stdlib/agents/researcher.agency
```diff
@@ -144,7 +144,7 @@ def researchLoop(
   }
 }
 
-export def researcherAgent(
+export handoff def researcherAgent(
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 2bf5ccca5. researchLoop pushes the persona through ensureSystemMessage; the history seeding still runs only on a fresh thread, with threadIsNew read before the push.

packages/agency-lang/lib/stdlib/gitignore.ts
```diff
@@ -0,0 +1,130 @@
...
+
+/** The .gitignore files that already apply to `dir` before a walk enters
+ * it: those in its ancestors up to and including the repository root (the
+ * nearest ancestor holding a `.git` entry), outermost first. A search
+ * rooted in a subdirectory would otherwise miss the root file, where
+ * rules like `**\/*.js` usually live. With no repository above `dir`,
+ * nothing applies. */
+export async function readAncestorGitignores(dir: string): Promise<GitignoreFile[]> {
+  const ancestors: string[] = [];
+  let current = path.dirname(path.resolve(dir));
```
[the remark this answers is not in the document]
egonSchiele: Fixed in f55825fd9. readAncestorGitignores checks dir itself for .git first and returns nothing in that case, with a nested-repository test.

packages/agency-lang/lib/logsViewer/treeRows.ts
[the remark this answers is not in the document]
egonSchiele: Fixed in f55825fd9. The stale comment is gone; wrapLine carries its own.

packages/agency-lang/lib/agents/agency-agent/lib/models.agency
```diff
@@ -37,18 +37,18 @@ type ProviderModels = {
 
 export static const MODELS: Record<string, ProviderModels> = {
   openai: {
-    fast: "gpt-5-mini",
+    fast: "gpt-5.6-sol",
```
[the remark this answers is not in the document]
egonSchiele: models.test.json was updated in this branch, and resolution.test.json plus the llm-provider-defaults fixture caught up in f55825fd9.

packages/agency-lang/lib/logsViewer/treeRows.ts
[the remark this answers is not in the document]
egonSchiele: Tool-call rows are wrapped too: formatConversation passes each tool-call line through wrapBody (lib/logsViewer/conversation.ts, the loop over toolCalls), so nothing is clipped.

packages/agency-lang/lib/stdlib/shell.ts
```diff
@@ -385,12 +404,13 @@ async function walkDir(root: string, visit: Visitor): Promise<void> {
...
       }
+      if (options.respectGitignore && isIgnored(full, st.isDirectory(), scoped)) continue;
       if (!(await visit(full, st))) return false;
-      if (st.isDirectory() && !(await walk(full))) return false;
+      if (st.isDirectory() && !(await walk(full, scoped))) return false;
     }
     return true;
   }
-  await walk(root);
+  await walk(root, options.respectGitignore ? await readAncestorGitignores(root) : []);
```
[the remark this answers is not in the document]
egonSchiele: Left as is. When someone points grep at a directory explicitly, returning nothing because an ancestor rule ignores that directory would be more surprising than searching it. The ancestor rules still apply to the entries inside it.

PR #1023, opened 2026-09-05
agency agent: budgets belong to the subagent call, and a stopped handoff points at its work

egonSchiele:
## What changed

**Budgets are a parameter on the subagent call.** `researchAgent`, `explorerAgent`, and `oracleAgent` take `maxSeconds` and `maxDollars`. The coordinator fills them in from the user's words ("in five minutes" is 300), the wrapper converts them and passes them to the stdlib agent, and that agent's own guard enforces them. With neither set, the stdlib agent's usual default applies. The coordinator prompt says when to set them.

**The turn guard is gone.** The harness used to parse a deadline out of the message with an extra LLM call and wrap the whole turn, the coordinator's own `llm()` loop included, in a guard labeled `agency-turn`. That guard and the parser are removed. The only whole-agent limit is the root budget the `--max-time` and `--max-cost` flags install, which the runtime enforces before any Agency code runs.

**A stopped handoff points at its work.** When a subagent's guard trips, the harness handler rejects it as a hard stop, the wrapper returns a failure naming the budget that ran out, and the tool loop pushes a resume message: `[researchAgent stopped before finishing: its time budget of 5m ran out]`, followed by a line saying the work so far is in the messages above and to continue with the user's request from it. An aborted result from an outer guard trip gets the same message instead of being recorded as a success carrying raw JSON. The coordinator prompt says to continue from that work and say what is unfinished.

## Why

In a session on 2026-09-05 the user asked for research "in no more than 5 minutes". The researcher had a complete answer at 2.5 minutes, a review sent it back for revision, and the turn guard tripped during the revision. Because the guard wrapped the coordinator too, rejecting it unwound everything, and the user got a bare "stopped" line while all the research sat in the shared thread. Asking "summarize findings so far" by hand produced the summary from that thread in one call. This PR makes that happen without the extra ask.

Dev note: `docs/dev/agents/subagent-budgets.md`.

## Not in this PR

`codeAgent` still runs on fixed internal budgets. Giving it the same parameters means threading them through its triage, escalation, and supervision paths.

The mid-run "give me more time?" prompt is dropped with the turn guard. The coordinator can ask the same question with the partial results in hand.

## Tests

- `lib/agents/agency-agent/tests/turn.agency`: `budgetTripIsHardStopAndBrainContinues` replaces the two turn-budget cases. A brain runs one LLM call under a tiny cost guard; the harness rejects the trip and the brain's reply is built from the failure.
- `tests/agency-js/handoff`: `failureInside` now checks the stopped message and that the old "finished" line is absent. The race-cancellation case is unchanged.
- Ran: `make`, `pnpm run typecheck`, `pnpm run lint:structure`, `pnpm run fmt:ts`, the turn tests, the handoff suite, the coordinator tool-wiring tests, and the runtime handoff and prompt unit tests.

https://claude.ai/code/session_01Jea8gkmJKtGzcj4X4ngyqt

### DIFF
--- a/packages/agency-lang/docs/dev/agents/subagent-budgets.md
+++ b/packages/agency-lang/docs/dev/agents/subagent-budgets.md
@@ -1,91 +1,49 @@
 # Subagent budgets
 
-How a deadline or spend limit the user states reaches the subagent that
-does the work, what happens when it runs out, and why the harness no
-longer wraps the whole turn in a guard.
+A budget is a parameter on the subagent tool call, enforced by the
+stdlib agent's own guard. The harness does not guard the turn.
 
-## The shape
+## Where a budget comes from
 
-There are exactly two places a budget can come from:
+- The `--max-time` and `--max-cost` flags cap the whole agent process. The
+  runtime installs them as a root budget before any Agency code runs
+  (`installRootBudget`, `lib/runtime/node.ts`), and a root budget cannot
+  be granted more.
+- `researchAgent`, `explorerAgent`, and `oracleAgent` take `maxSeconds`
+  and `maxDollars`. The coordinator fills them in from the user's words,
+  the wrapper converts them, and the stdlib agent's guard enforces them.
+  Unset means the stdlib agent's default.
 
-- **The `--max-time` and `--max-cost` flags.** These cap the whole agent
-  process. The runtime installs them as a root budget before any Agency
-  code runs (`installRootBudget`, `lib/runtime/node.ts`), and a root budget
-  cannot be granted more. Nothing in the agent's own code is involved.
-- **A parameter on the subagent tool.** `researchAgent`, `explorerAgent`,
-  and `oracleAgent` take `maxSeconds` and `maxDollars`. The coordinator
-  fills them in from the user's words ("do this in five minutes" becomes
-  `maxSeconds: 300`), the wrapper converts them to the milliseconds and
-  dollars the stdlib agent takes, and the stdlib agent's own `guard`
-  enforces them. When the model sets neither, the wrapper passes the
-  stdlib agent's usual default, so the cap is the same as calling that
-  agent directly.
+`codeAgent` runs on fixed internal budgets and does not take the
+parameters yet.
 
-The conversion and the stop text live in
-`brains/coordinator/lib/budgets.agency`. `codeAgent` still runs on fixed
-internal budgets; giving it the same parameters means threading them
-through its triage, escalation, and supervision paths, and is not done
-yet.
-
-## What happens when a budget runs out
+## When a budget runs out
 
 1. The stdlib agent's guard trips and raises `std::guard`.
-2. No handler inside the agent owns the trip, so it reaches
-   `guardStopHandler` in `lib/budget.agency`, the handler `runTurn` wraps
-   every turn in. It prints one line in an interactive session and
-   rejects. Rejecting is what makes the budget a hard stop: passing would
-   park the turn on an interrupt nobody answers.
-3. The guard converts the rejection at its own site into a failure, or
-   into a success carrying the saved draft when the agent saved one.
-4. The subagent wrapper turns that into the tool's result through
-   `agentOutcome`: a failure whose text names the budget that ran out
-   ("its time budget of 5m ran out"), and an empty answer is reported as
-   a stop rather than a blank success.
-5. The coordinator's tool loop pushes the resume message for a stopped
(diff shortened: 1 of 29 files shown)


packages/agency-lang/docs/dev/agents/subagent-budgets.md
```diff
@@ -0,0 +1,91 @@
+# Subagent budgets
```
egonSchiele, line 1: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/research.agency
```diff
@@ -47,6 +50,12 @@ export handoff def researchAgent(
     itself (its syntax, types, stdlib, or CLI), which routes to the agent
     that reads the bundled Agency documentation. Leave it false for every
     other subject, including questions about other languages.
+  @param maxSeconds - stop the research after this many seconds of working
```
egonSchiele: make this less verbose

packages/agency-lang/lib/runtime/handoff.ts
```diff
@@ -26,6 +26,19 @@ export function handoffResumeText(toolName: string, body: string): string {
...
+/**
+ * The resume message for a handoff that failed or was aborted partway.
+ * A handoff leaves everything it did on the caller's thread, so the
+ * caller is told where that work is and to carry on from it.
+ */
+export function handoffStoppedText(toolName: string, reason: string): string {
+  return (
+    `[${toolName} stopped before finishing: ${reason}]\n` +
+    `Its work so far is in the messages above, from the line where it was dispatched onward. ` +
+    `Continue with the user's request using that work, and say what is unfinished.`
```
egonSchiele: Remove "from the line where it was dispatched onward" and "and say what is unfinished".

packages/agency-lang/docs/dev/agents/subagent-budgets.md
egonSchiele: Trimmed. The doc now states where a budget comes from, the four steps when one runs out, and the tests. The narrative about the old turn guard is down to one paragraph. Module docs and code comments cut the same way.

packages/agency-lang/lib/agents/agency-agent/lib/budget.agency
```diff
@@ -1,213 +1,41 @@
...
-      Reply: ${reply}
-      """,
-    )
-    parsed = extracted
-  }
-  return parsed
-}
-
-export def turnBudgetHandler(intr: any): any {
+export def guardStopHandler(intr: any): any {
```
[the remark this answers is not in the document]
egonSchiele: Fixed. The fixture now drives guardStopHandler; the own-label case is gone since there is no own label. Three nodes pass locally.

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/lib/budgets.agency
```diff
@@ -0,0 +1,55 @@
...
+  salvaged nothing) is reported as a stop too rather than as a blank
+  success. */
+export def agentOutcome(result: Result<string>): Result<string> {
+  if (isFailure(result)) {
+    return failure(describeStop(result.error))
+  }
+  if (result.value == "") {
+    return failure("it stopped before producing an answer")
+  }
+  return result
```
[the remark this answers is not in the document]
egonSchiele: Right. budgeted() now wraps each stdlib call with a handler that rejects a trip of that agent's guard label and remembers it, so the wrapper returns a failure naming the budget even when the guard salvaged a draft. The draft is on the thread already for a handoff. Covered by brains/coordinator/tests/budgets.agency, including the drafting case and a trip of some other label.

packages/agency-lang/lib/agents/agency-agent/brains/coordinator/subagents/research.agency
egonSchiele: Shortened to one line per parameter, same in the explorer and oracle.

packages/agency-lang/lib/runtime/handoff.ts
egonSchiele: Removed both. The message is now: [X stopped before finishing: reason] / Its work so far is in the messages above. Continue with the user's request using that work.

PR #1025, opened 2026-09-05
stdlib: route mcp, llm, and localModels file access through contained.ts

egonSchiele:
Second PR after #1024. The four `lib/stdlib` files that still took a path from the caller and used `fs` directly now go through `contained.ts`, and their entries leave the `FS_IMPORTERS` allow list. The list now holds `contained.ts` itself, the four agent-home modules, and the files with a stated reason (`gitignore.ts`, `shell.ts`, `speech.ts`, `utils.ts`).

## What changed

| Function | Before | After |
|---|---|---|
| `_addMcpServer`, `_removeMcpServer`, `_readMcpServersFromFile` (`mcp.ts`) | raw read, mkdir, write of the config file | `wholePath` + `readText` / `mkdir` / `writeText`. A symlink where the config file should be is refused and its target is left alone. |
| `_loadModelData` (`llm.ts`) | raw `readFileSync(path)` | `fixedPath` + `readText`. The Agency `loadModelData` raises `std::read` naming the real path before the read. The CLI `agency models` path passes `wholePath` because it has no approval step. |
| alias config, catalog file, cache list, `_removeModel`, freshness snapshot (`localModels.ts`) | raw `fs` on `join(cacheDir, name)`; `_removeModel("../x")` escaped | through the module. Linked entries in the cache are hidden. `../x` is refused. |
| `fileSha256`, `verifyModelFile` (`localModels.ts`) | `createReadStream(path)`, `renameSync` | new `readStream` primitive (a stream over a validated descriptor, for a multi-gigabyte file) and `move`. |
| `_readSkill` (`skills.ts`), runtime `readSkill` builtin tool | no callers | deleted |

## Behavior changes a user can see

- `loadModelData(path)` raises `std::read` with `{ dir, filename }` in the real spelling. A program that called it with no handler now gets an interactive prompt, or a rejection under a policy that refuses reads. The agency-js test for it approves the interrupt.
- `agency mcp add` and `agency mcp remove` refuse a config path that is a symlink.
- `agency local list` no longer shows a symlinked `.gguf`, and `agency local remove` refuses one.
- A model-data or catalog file reached through a symlinked directory in its own spelling still loads from the CLI. From Agency, a link planted in that spelling after approval is refused.

## Notes for review

- `readStream` is registered in `PRIMITIVES`, so the symlink battery runs it against every planted link, and the registry test would fail if it were left out.
- `_loadModelData` takes a `locate` argument (default `fixedPath`) because the same function runs after an Agency approval and from the CLI with none. `docs/dev/stdlib/contained-files.md` records the pattern.
- The `std::read` effect is declared in `llm.agency` the way `agency.agency` declares it, since a module that raises an effect declares it itself. `docs/site/stdlib/llm.md` was regenerated by `make`.
- Tests that call `_loadModelData` directly now use real paths (`/var` is a link on macOS).

## Package entry points

`packages/mcp/index.js` and `packages/web-fetch/index.js` are regenerated from their `index.agency` with the current compiler. The committed versions imported `readSkill` from the runtime, which this PR deletes, and they already failed to load on main because they also imported `emit`, which stdlib no longer exports. The package tests load the packages through `pkg::`, so they were unaffected. Both entry points import cleanly after `pnpm run build`.

https://claude.ai/code/session_01932VFdBf79Jhq6wMEp7wg9


egonSchiele:
On the `readSkill` exports: the committed `packages/mcp/index.js` and `packages/web-fetch/index.js` were compiled by an old compiler and already failed to load on main. Importing either throws `does not provide an export named 'emit'`, before it reaches `readSkill`. Both are regenerated from `index.agency` with the current compiler in the latest commit, which drops the `readSkill` import because the compiler stopped emitting it. After `pnpm run build` in each package, both entry points import cleanly and expose `mcp` and `fetchPage`. The package tests use `pkg::` imports, which compile `index.agency` directly, so they were unaffected either way.

The models-list test now asserts that every `_loadModelData` call receives `wholePath`.

PR #1028, opened 2026-09-05
policy: <agent-home> placeholder, learned skills and tools in the read scope, recordUse effect

egonSchiele:
Replaces #1014 (PR 3 of the skills-and-toolbox split: the policy half). #1014 was written before the contained file API (#1024, #1025, #1026) and carried its own symlink defenses. Rebasing would have meant untangling 22 conflict hunks whose old side is code we now want gone, so this PR starts from main, carries the policy work over unchanged, and rewrites the three stdlib files by hand. Nothing here changes what the agent does. It changes what the built-in policies decide without asking.

## The `<agent-home>` placeholder

`<agent-home>` in a `dir` pattern expands at match time to `AGENCY_AGENT_HOME` or `~/.agency-agent`, the way `<agency>` expands to the install root. An empty variable counts as unset and a relative one resolves against the cwd, matching the `--agent-home` launcher. The expansion goes through `root()` from `contained.ts`, the same walker every file effect uses for the directory in its payload, so a home behind a symlinked ancestor (`/tmp` on macOS) matches, and a home that does not exist yet keeps a lexical tail under its nearest real ancestor.

`agentHomeDir` in `lib/runtime/agentHome.ts` is the one resolver. `agency mcp` used to have its own copy.

The matcher expands every placeholder in one pass, so a pattern that mixes `<agency>` and `<agent-home>` in one brace group works.

## What `recommended` now decides

- Reads under `<agent-home>/skills` and `<agent-home>/tools`, and below them, are approved. Both directories hold only what a review interrupt let in. Without this, every learned-skill read, catalog scan, and `runTool` prompts and auto-rejects headless. The home itself is not covered.
- `std::skills::skillsDir` and `std::skills::commandsDir` take the read scope instead of an unscoped approve. A scan could otherwise read a directory a plain `read` would have prompted for.
- `std::toolbox::recordUse`, new, is approved under `<agent-home>/tools/**`.
- The save and review gates have no rule in any built-in policy but `approve-all`. A test pins that.

## `std::toolbox::recordUse`

`runTool` counts a use in the tool's `meta.json` behind this effect, not a `std::write`. A path glob cannot tell the stdlib's bookkeeping from any program writing arbitrary content into `meta.json`, whose `purpose` text `listTools` trusts and whose `maxTime` sets the run's time limit. The count is best-effort: a declined interrupt or a failed write leaves it where it was and the run returns its result. The write goes through `_write` with the toolbox root as its approved directory.

## What #1014 had that this PR does not

- `writeContainedFile.ts` and its test (226 lines). `writeText` in `contained.ts` does this.
- An `allowedPaths` guard on `_write`. The underscore primitives take a required root now.
- `canonicalDir` in `resolveDir.ts`, its own realpath walk. `root()` is the one walker.
- The scan payloads already carry the real directory on main (#1024), so the `skills.agency` and most `toolbox.agency` changes were already there. The skills save payload is the one place that still used the lexical spelling; it now uses `_realDir`.

## The open threads on #1014

All nine were about the symlink handling. They close as follows: the FIFO, short-write, truncation, and root-parent findings on `writeContainedFile` are moot, since the file is gone and `writeText` writes a validated sibling and renames it; the glob that excluded the `skills` and `tools` roots was already fixed on that branch and is carried here; the payload-spelling findings are covered by `_realDir` on every scan and save; the `_glob` with no allowed paths and the `metaFacts` default-before-read findings are covered by the required root on every primitive; the lexical-home finding is covered by `root()`.

## Testing

- `lib/runtime/policy.test.ts`: the placeholder's expansion, escaping, relative and empty env values, the real path, matching under the home and nowhere else, a mixed-placeholder pattern.
- `lib/runtime/builtinPolicies.test.ts`: the third read rule, every scan scoped like a read, the use-count rule, the gates undecided.
- `tests/agency/toolbox/designTool.agency`: the record names the tool directory, a declined record keeps the result and the count, a symlinked tool is unlisted and not run, with a positive control that the link is reachable.
- `tests/agency/toolbox/listTools.agency` and `tests/agency/skills-dir.agency`: scan payloads are the real directory.
- The always-scope table, `lib/cli/mcp.test.ts`, `runPolicy`, `effects`, typecheck, prettier, structural lint.

https://claude.ai/code/session_01932VFdBf79Jhq6wMEp7wg9

### DIFF
--- a/packages/agency-lang/docs/dev/agents/approval-policies.md
+++ b/packages/agency-lang/docs/dev/agents/approval-policies.md
@@ -58,8 +58,10 @@ question, a review) get no "always" answers at all.
 ## What `recommended` lets the agent read
 
 The read-only file effects (`std::read`, `std::readBinary`, `std::ls`,
-`std::glob`, `std::grep`) are approved in two places only
-(`readScopeRules` in `lib/runtime/builtinPolicies.ts`):
+`std::glob`, `std::grep`) and the scan effects that read a whole
+directory under one approval (`std::skills::skillsDir`,
+`std::skills::commandsDir`, `std::toolbox::scan`) are approved in three
+places only (`readScopeRules` in `lib/runtime/builtinPolicies.ts`):
 
 - the launch directory and everything under it, written as `{.,./**}` so
   the rule keeps meaning "wherever the agent runs" after the policy is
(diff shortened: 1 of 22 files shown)


packages/agency-lang/docs/dev/agents/approval-policies.md
```diff
@@ -58,8 +58,10 @@ question, a review) get no "always" answers at all.
 ## What `recommended` lets the agent read
 
 The read-only file effects (`std::read`, `std::readBinary`, `std::ls`,
-`std::glob`, `std::grep`) are approved in two places only
```
egonSchiele, line 61: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/lib/runtime/policy.ts
```diff
@@ -158,6 +160,35 @@ export function expandAgencyInstallDir(
...
+export const AGENT_HOME_PLACEHOLDER = "<agent-home>";
+
+// Expanded at match time, like `<agency>`. Exported for tests, which
+// inject the home.
+export function expandAgentHomeDir(
+  pattern: string,
+  home: () => string = canonicalAgentHome,
+): string {
+  if (!pattern.includes(AGENT_HOME_PLACEHOLDER)) return pattern;
+  return pattern.split(AGENT_HOME_PLACEHOLDER).join(escapeForGlob(home()));
```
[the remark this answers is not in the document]
egonSchiele: Fixed. escapeForGlob is deleted and all three placeholder expansions (the dot, <agency>, and <agent-home>) use the exported escapeGlob, which escapes commas. New test: with the home at /nowhere/a,b, a read under its skills directory is approved and reads of /nowhere/a and b/skills are not.

packages/agency-lang/docs/dev/agents/approval-policies.md
egonSchiele: Went through the three dev docs and every comment in the diff. Cut: the two passages that described what the policy used to do before this change, a sentence defending that this is not symlink support, the note that a function is exported for tests, a comment restating the one-pass placeholder expansion, and two comments on the stdlib wrappers that repeated the containment reasoning the contained-files doc already carries. Kept: what the rules are, why recordUse is its own effect, how the placeholders expand, and what a symlinked tool directory does under a scan, since those describe current behavior a reader would otherwise have to work out from the code.

PR #1030, opened 2026-09-06
runtime: repeat-streak key without U+0000, so checkpoints survive jsonb

egonSchiele:
**Bug.** `repeatKey` joins the tool name and the argument digest with U+0000. The repeat streak is stored on the frame, so a checkpoint taken after the same tool has been called twice in one run carries a NUL. Hosts that keep checkpoints as JSON in Postgres jsonb get `unsupported Unicode escape sequence` and the write rolls back. In bmo this lost the second interrupt of a run: the agent raised one, the user approved, the model called the tool again, and the pause could not be stored.

**Fix.** Join with a colon. Tool names are identifiers and the digest is hex, so there is no collision. A checkpoint saved before this change still loads; its old key just never matches a new one, so that streak restarts from zero once.

**Test.** `toolLoopGuards.test.ts` gains a case that the key contains no U+0000 and round-trips through JSON. It failed before the change and passes after. `tsc --noEmit` and prettier are clean.

https://claude.ai/code/session_01NsGtEyiFmh5ft67AEsjr1p

### DIFF
--- a/packages/agency-lang/lib/runtime/toolLoopGuards.test.ts
+++ b/packages/agency-lang/lib/runtime/toolLoopGuards.test.ts
@@ -65,6 +65,12 @@ describe("repeated tool calls", () => {
 
   const fresh = freshRepeatStreak;
 
+  it("makes a key that survives JSON storage: the streak rides on the checkpoint, and Postgres jsonb rejects U+0000", () => {
+    const key = repeatKey("raiseInterrupt", { a: 1, b: "asdasd" });
+    expect(key).not.toContain("\u0000");
+    expect(JSON.parse(JSON.stringify({ key })).key).toBe(key);
+  });
+
   it("counts identical results in a row and starts over when the result changes", () => {
     const streak = fresh();
     const key = repeatKey("typecheck", { source: "def f() {}" });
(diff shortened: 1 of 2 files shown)


packages/agency-lang/lib/runtime/toolLoopGuards.test.ts
```diff
@@ -65,6 +65,12 @@ describe("repeated tool calls", () => {
 
   const fresh = freshRepeatStreak;
 
+  it("makes a key that survives JSON storage: the streak rides on the checkpoint, and Postgres jsonb rejects U+0000", () => {
+    const key = repeatKey("raiseInterrupt", { a: 1, b: "asdasd" });
+    expect(key).not.toContain("\u0000");
+    expect(JSON.parse(JSON.stringify({ key })).key).toBe(key);
```
egonSchiele: This second assertion cannot fail. `JSON.stringify` escapes U+0000 as `\u0000` and `JSON.parse` restores it, so the round-trip is lossless for the old NUL-joined key too: this line passed before the change and passes after. The only assertion doing work is the `not.toContain` above. Either drop this line or state the real constraint (a store that rejects U+0000, which JSON itself does not), so the test cannot be read as proof that the old key was unstorable.

PR #1031, opened 2026-09-07
Every failure carries a message

egonSchiele:
Every failure now carries a string message, and optionally an object of extra data.

```
def parseConfig(path: string): Result {
  return failure("Bad syntax in ${path}", { line: 4, column: 12 })
}

const result = parseConfig("app.json")
if (isFailure(result)) {
  print("${result.error} (line ${result.data.line})")
}
```

`ResultFailure` gains a `data` field that is `{}` when the producer passed none, and its `error` narrows from `any` to `string`.

## Why

Thirteen producers put an object in `error`. Five already built a `message` field by hand and buried it inside that object. The other eight had no message anywhere, so each of their callers hand-rolled a way to make a sentence.

Five consumers guessed how to stringify a failure. The one that mattered is the tool loop: `toolErrorMessage` JSON-encoded an object error before handing it to the model, so a tool that tripped a guard sent the model `{"type":"guardFailure","maxCost":0.3,...}`. It now sends "Guard 'research' exceeded its cost budget: spent 0.42 of 0.3."

## The type parameters

`Result<T, D>` now reads as "succeeds with a `T`, or fails with a message plus `D` of data". `D` must be an object; `Result<T, string>` is refused, because no `failure()` call can produce string data. `Result<T>` no longer means a string failure, and neither does the wrapper around a validated return.

A function that declares a data type has to supply it on every `failure`. Write `Result<T, D | null>` to allow both forms.

Four new diagnostics. AG2013 and AG2014 refuse a bad message or bad data. AG2016 refuses the annotation. AG2015 is the interesting one: a bare `failure(msg)` synthesizes `Result<any, null>`, so the existing covariant Result check already rejected it against a declared data type — but what it said was "Result<any, null> is not assignable to Result<Policy, ParsePolicyFailure>". AG2015 names the second argument and the `| null` alternative instead.

`reject` takes a string. Every call site already passed one.

## The riskiest part

Codegen injects an options object (checkpoint, function name, arguments) into every `failure` call **positionally**. Once a user can write two arguments, the builder has to pad the data slot, or a two-argument failure puts the user's data where the options belong and loses its checkpoint, which breaks resume from that line. `tests/agency/result/failure-with-data-args.agency` is the test for it.

That positional shape would otherwise be spelled out in three mustache templates and two raw codegen strings, none of which the compiler checks. `runtimeFailure(error, opts)` is what those twelve sites use instead, so there is one place to get the arity wrong rather than six. It coerces too, which retires three hand-written Error-to-message conversions.

## Verification

`make`, `pnpm run typecheck`, `pnpm run lint:structure`, and `pnpm test:run` (13,259 tests) are green. So are `tests/agency/{result,guards,subprocess}` and the two agency-js fixtures this touches. The three Agency diagnostics still printed under `lib/agents/` predate this branch and are unchanged.

Spec: `packages/agency-lang/2026-09-06-result-failure-message-spec.md` — its last section records the five things execution turned up that the spec did not anticipate.

The two-name `failure(msg, data)` pattern is a follow-up PR.

https://claude.ai/code/session_01XMyopwfvWQps75VMgZUojH


packages/agency-lang/docs/dev/language/result-failures.md
```diff
@@ -0,0 +1,117 @@
+# The `Result` failure
```
egonSchiele, line 1: Read through this doc and any other docs you added to docs/dev/, as well as all the comments you have added across all files in this PR, and consider whether this is useful information for future readers to know, or whether these are just your implementation notes and thought process. If it is the latter, then they do not belong in this PR.

packages/agency-lang/lib/backends/typescriptBuilder.ts
```diff
@@ -2612,11 +2612,18 @@ export class TypeScriptBuilder {
...
+      // two-argument failure would put the user's data where the options
+      // belong and lose its checkpoint, which breaks resume from that line.
       const scope = this.scopes.current() as FunctionScope;
       const argNodes: TsNode[] = node.arguments.map((arg) => this.processCallArg(arg));
+      const message = argNodes[0] ?? ts.raw("null");
+      const data = argNodes[1] ?? ts.raw("null");
       return ts.call(ts.id("failure"), [
-        ...argNodes,
+        message,
+        data,
```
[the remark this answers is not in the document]
egonSchiele: Confirmed and fixed in c2c87f02e. I compiled `def f(args: any[]): Result { return failure(...args) }` and got:

```js
failure(...__stack.args.args, null, { checkpoint: ..., functionName: "f", args: __stack.args })
```

With two elements in `args`, `null` occupies the options slot and the real options are a fourth argument that `failure` ignores. So the failure loses its checkpoint, function name, and args — the exact bug the padding exists to prevent.

I went with refusing rather than materializing. A splat expands to a width the compiler cannot know, so there is no arity to pad to, and `failure` has a fixed two-argument shape that a splat has no reason to target. AG2017 refuses it, and the builder throws rather than emitting one.

packages/agency-lang/lib/backends/typescriptGenerator/typeToZodSchema.ts
```diff
@@ -295,7 +295,8 @@ export function mapTypeToValidationSchema(
     typeAliases,
     (vt, ta) => {
       const successSchema = mapTypeToValidationSchema((vt as any).successType, ta, typeAliasesFull);
-      return `z.union([z.object({ __type: z.literal("resultType"), success: z.literal(true), value: ${successSchema} }), z.object({ __type: z.literal("resultType"), success: z.literal(false), error: z.any() })])`;
+      const dataSchema = mapTypeToValidationSchema((vt as any).dataType, ta, typeAliasesFull);
+      return `z.union([z.object({ __type: z.literal("resultType"), success: z.literal(true), value: ${successSchema} }), z.object({ __type: z.literal("resultType"), success: z.literal(false), error: z.string(), data: ${dataSchema} })])`;
```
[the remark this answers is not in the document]
egonSchiele: Right, and sharper than I had it. I had recorded this as a known gap in the dev doc and left the schema strict; your framing is what changed my mind — a bare `failure(msg)` that AG2015 explicitly permits would fail validation, so the schema was refusing a legal program.

Fixed in c2c87f02e by making the data slot check only that it is an object (`z.record(z.any())`), not its shape. Reasoning: the runtime turns absent data into `{}`, which no schema for a real `D` matches, so any shape check there is wrong for `D | null`. And a failure's data is produced by the program itself, not by a model, so what it holds is already settled statically by AG2014 through AG2016. `error: z.string()` stays, because that one is a runtime invariant.

packages/agency-lang/lib/runtime/result.ts
```diff
@@ -86,7 +103,11 @@ export type SkippedFunction = { name: string; param: string };
 export type ResultFailure = {
   __type: "resultType";
   success: false;
-  error: any;
+  error: string;
+  /** Extra structured detail, named by the second `Result` type parameter.
+   *  `{}` when the producer passed none. Never null: a reader may write
+   *  `err.data.status` without checking first. */
+  data: Record<string, any>;
```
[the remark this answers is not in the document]
egonSchiele: Confirmed. `__tryCall` returns a foreign Result-shaped value unchanged, so imported TypeScript that builds `{__type: "resultType", success: false, error: {code: 404}}` by hand gets through with an object error and no `data` — which is the shape this branch exists to eliminate.

Fixed in c2c87f02e. `__tryCall` now runs an accepted Result through `normalizeForeignResult`, which coerces the error and defaults the data, and returns a well-formed failure unchanged by identity. I put it only at that boundary rather than inside `isFailure`: `try` is the one place a foreign value becomes an Agency Result, and `isFailure` is on the hot path of every call.

Left `resultValueSchema` permissive on purpose. Tightening it would make a malformed foreign Result fail the sniff and get wrapped in `success(...)`, which is worse than normalizing it.

packages/agency-lang/lib/runtime/interruptResponse.ts
```diff
@@ -8,13 +8,13 @@
...
 export type InterruptResponse = InterruptApprove | InterruptReject;
 
 export function approve(value?: any): InterruptResponse {
   return { type: "approve", value };
 }
 
-export function reject(value?: any): InterruptResponse {
-  return { type: "reject", value };
+export function reject(reason?: string): InterruptResponse {
+  return { type: "reject", value: reason };
```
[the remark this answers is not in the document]
egonSchiele: Correct, and the miss was bigger than the builtin signature. Generated modules define their own `reject` in `imports.mustache`, so narrowing `lib/runtime/interruptResponse.ts` never reached Agency code at all.

Fixed in c2c87f02e: `builtins.ts` says `params: [string]`, and the generated constructor is `reject(reason?: string)`. Added typechecker tests that `reject(42)` is refused and that `reject("...")` and bare `reject()` are accepted. I checked every `reject` call in the repo first — 96 bare and 61 with an argument, all strings — so nothing had to change.

packages/agency-lang/lib/backends/typescriptGenerator/typeToString.test.ts
```diff
@@ -144,18 +144,18 @@ describe("variableTypeToString: object rendering hook", () => {
...
         type: "resultType",
         successType: objectOf(numberType),
-        failureType: objectOf(stringType),
+        dataType: objectOf(stringType),
       },
       expected: "Result<OBJECT<number>, OBJECT<string>>",
     },
     {
-      name: "Result success type with string failure shorthand",
+      name: "Result success type with string datare shorthand",
```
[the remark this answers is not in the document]
egonSchiele: Fixed in c2c87f02e. Mine: I did a blind replace of "failure type" with "data" and it landed mid-word.

packages/agency-lang/docs/dev/language/result-failures.md
egonSchiele: Did that pass in de6548f39. What I cut:

**Comments that recorded how the code got here, not what it does.**
- `agencyGenerator.test.ts`: "`string` is no longer the default data type, so the printer no longer elides it." A future reader does not need what the printer used to do.
- `toolbox.agency` `oneRound`: "A retryable problem becomes the next round's feedback. Anything else ends the loop." That is what the two returns underneath it say.
- `runtimeFailure`'s docstring and the builder's splat throw both explained how little work something was, or hedged about a check "ever moving". Both now state the contract and stop.
- "POSITIONALLY" in caps became a plain word.

**In the dev doc.**
- The opening sentence announced the doc's own structure. Cut.
- "Two things follow." / "This needed no new machinery." / "What it did need is a message." All three were me narrating the order I worked things out in. The mechanism they wrapped is still there.
- The `reject` line said narrowing the runtime helper "would not have reached Agency code", which is my mistake, not a fact about the code. It now says the plain fact: generated modules define their own `reject`, so three places have to agree.

**A whole section.** I had appended "What changed during execution" to the spec — five things the spec did not anticipate. That is a postmortem. Removed. The `Decisions` section stays, since that records what was settled rather than how I got there, and it is what the other specs in the repo carry.

What is left in the dev doc is the failure shape, the two things that silently break resume if someone changes them (the positional options object, and why a splat cannot be padded around), why the second type parameter means data, and where each diagnostic is pushed. I kept the file list because `CLAUDE.md` asks dev docs to carry one.

I did keep the spec, plan, and plan-REVIEW files. Those looked like exactly the notes you are describing, but the repo tracks the same triple for other arcs (`2026-08-07-per-invocation-config-override-{spec,plan,spec-review}.md`, `2026-08-09-agency-speech-via-smoltalk-plan-review.md`), so I followed that rather than deviating on my own read. Say the word and I will drop them.

packages/agency-lang/lib/runtime/result.ts
```diff
@@ -110,11 +131,43 @@ export function success(value: any): ResultSuccess {
...
+}
+
+function isPlainObject(value: unknown): boolean {
+  return value != null && typeof value === "object" && !Array.isArray(value);
+}
+
+/** Data is always an object, so a reader can write `err.data.status` without
+ *  checking first. */
+function coerceData(data: unknown): Record<string, any> {
+  if (!isPlainObject(data)) {
```
egonSchiele, line 162: if its not a plain object, should we print a warning message? or log a warning to statelog?

packages/agency-lang/lib/runtime/prompt.ts
```diff
@@ -1241,11 +1232,11 @@ export async function runPrompt(args: {
       // immediately. (Stdlib tools gate BEFORE their destructive regions,
       // so their rejections are always clean.)
       if (isFailure(toolResult) && toolResult.rejected && !toolResult.destructiveRan) {
-        return recordRejection(toolErrorMessage(toolResult.error));
+        return recordRejection(toolResult.error);
       }
 
       if (isFailure(toolResult)) {
-        const errorMessage = toolErrorMessage(toolResult.error);
+        const errorMessage = toolResult.error;
```
egonSchiele, line 1239: **Foreign Results are only normalized inside `try`, so the tool loop can send the model `[object Object]`.**

`normalizeForeignResult` (result.ts:298) has one caller, `__tryCall`. A tool implemented in imported TypeScript that returns a hand-built Result (`{ __type: "resultType", success: false, error: { code: 404 } }`) comes back through `AgencyFunction.invoke()` (agencyFunction.ts:212, `return this._fn(...args)`) and `__call` (call.ts:93) unchanged. `isFailure` accepts it (it checks only `__type` and `success`), so this branch reads `toolResult.error` as an object. `capToolResultForLlm` returns the object untouched when it is under the cap, and `String(...)` on the next line turns it into `Error: [object Object]. ...`; statelog also receives an object in `message`. The deleted `toolErrorMessage` used to JSON-stringify exactly this case (its deleted test pinned `{"source":"x","errors":[]}`). Same root cause in Agency code without `try`: `const r = load(); if (r is failure(e)) { print(r.data.code) }` throws because `r.data` is undefined.

The PR added `normalizeForeignResult` because it treats the hand-built shape as real, so it should run at the JS-to-Agency boundary (in `invoke()` for non-Agency `_fn`, or at the top of this failure branch), not only in `__tryCall`.

packages/agency-lang/lib/typeChecker/resultUnion.ts
```diff
@@ -30,7 +30,12 @@ export function resultToObjectUnion(
         type: "objectType",
         properties: [
           { key: "success", value: bool("false") },
-          { key: "error", value: rt.failureType },
+          // The message. Always a string: `failure()` coerces anything else
+          // (lib/runtime/result.ts).
+          { key: "error", value: STRING_T },
+          // Extra structured detail, named by the second type parameter. `{}`
+          // at runtime when the producer passed none, so it is never null.
+          { key: "data", value: rt.dataType },
```
egonSchiele: **Reader-facing `data` keeps the `| null`, so the fix AG2015 recommends makes every `.data.field` read a type error.**

Reproduced with the built checker:

```
type D = { status: string }
def f(): Result<number, D | null> { return failure("x") }
def g(): string {
  const o = f()
  if (o is failure(e)) { return o.data.status }
  return ""
}
```

```
error AG2008: Property 'status' is not available on every member of 'D | null'
```

The same happens for an unannotated def that returns both `failure("x")` and `failure("y", { a: 1 })` (callers see `null | { a: number }`). Adding `if (o.data == null) { ... }` makes it typecheck, but that guard never fires because `data` is `{}` at runtime, which this comment already says. The guide diff tells users to write `if (result.data.line != null)`, which is precisely the expression AG2008 rejects.

Deeper fix: strip `null` from `rt.dataType` here, at the single read site, and keep `null` only on the producer-side assignability check in `utils.ts`. Then `Result<T, D | null>` reads as `D`, the documented idiom typechecks, and a `data == null` guard becomes an error instead of dead code.

packages/agency-lang/lib/typeChecker/utils.ts
```diff
@@ -173,6 +175,59 @@ export function checkType(
...
+        { actual: formatTypeHint(target.dataType), context },
+        loc ?? null,
+      ),
+    );
+    return true;
+  }
+  // A one-argument `failure(msg)` synthesizes null data (see
+  // synthFailureCall), so getting here means the target declared real data and
+  // the failure supplied none.
+  if (isNullType(actual.dataType)) {
```
egonSchiele: **AG2015 fires on any `Result<_, null>` vs `Result<_, D>` mismatch, not just a `failure()` call, and then names the wrong function.**

```
type Foo = { line: number }
def helper(): Result<number, null> { return failure("x") }
def main(): number {
  const r: Result<number, Foo> = helper()
  return 1
}
```

```
c2.agency:8:34 - error AG2015: This function declares 'Foo' as its failure data, so failure() needs a second argument. To allow a failure with no data, declare the return type as 'Result<number, Foo | null>'.
```

Column 34 is the `helper()` call. There is no `failure()` there, `main` declares `number`, and `helper` declares `null`; nothing in the message points at the real mismatch. The comment above assumes a null-data Result can only come from a one-arg `failure(msg)`, but a declared `Result<T, null>` return type, a variable, or a parameter can carry it too. Gate this on the actual expression being a `failure(...)` call (or check the call against its expected type where it is synthesized), and let the generic Result assignability message handle the rest.

packages/agency-lang/lib/typeChecker/validate.ts
```diff
@@ -90,6 +92,15 @@ export function validateTypeReferences(
   loc?: SourceLocation,
 ): void {
   visitTypes(vt, (t) => {
+    if (t.type === "resultType" && !isDataShaped(t.dataType, typeAliases)) {
```
egonSchiele, line 99: **AG2016 is not checked on `def` parameter or return annotations, and it fires once per return when it does fire.**

Reproduced with the built checker:

- `def h(r: Result<number, string>): number { return 1 }` and `def g(): Result<number, string> { return success(1) }` -> `No type errors found.`
- `type Bad = Result<number, string>` -> AG2016.
- `def g(x): Result<number, string>` with three `return failure(...)` -> three AG2016 errors, one per return.

`validateTypeReferences` is only called from `index.ts:196` (alias bodies), `scopes.ts:124` (`let`/`const` hints), and `templateNames.ts:132,135` (whose result is filtered by `unknownTypesIn`, which drops AG2016). No call site covers a function signature, so the only other emitter is `reportFailureDataError` on the assignability error path, which runs per mismatching return. A stdlib signature can keep advertising string data (the spelling this PR set out to retire) and callers reading `r.data.foo` get no diagnostic.

Fix shape: call `validateTypeReferences` on each parameter `typeHint` and the `returnType` when a function is declared (near `scopes.ts:69/85`), with `node.loc`, then drop the AG2016 branch from `reportFailureDataError` so the duplicates disappear.

packages/agency-lang/lib/runtime/failurePropagation.ts
```diff
@@ -233,7 +233,7 @@ export function checkResultMethodCall(obj: unknown, prop: string | number): void
     }
   }
   const message = isFailureObj
-    ? `called '.${String(prop)}()' on a failure produced by '${origin(obj as ResultFailure)}' (${truncate((obj as ResultFailure).error)}). Check the Result before using it.`
+    ? `called '.${String(prop)}()' on a failure produced by '${origin(obj as ResultFailure)}' (${(obj as ResultFailure).error}). Check the Result before using it.`
```
egonSchiele: **Dropped `truncate(...)` here while the three sibling messages in this file (lines 164, 198, 256) still truncate.**

`error` being a string does not make it short; line 256 truncates the same string field. A failure whose message is a large stdout dump (safeBash) followed by Agency code calling `.split()` on it now puts the whole dump into the thrown Error, and the console, statelog, and CLI printout carry it when it surfaces through a node boundary (`coerceMessage` returns `error.message` verbatim). Restore `truncate((obj as ResultFailure).error)`.

egonSchiele:
Review findings on lines outside the diff hunks (so they cannot be inline comments):

**1. `packages/web-fetch/index.js:283` and `packages/mcp/index.js:274` are stale compiled wrappers that still call the old two-argument `failure(msg, { checkpoint, destructiveRan, functionName, args })`.**
Both files are git-tracked, both import `failure` from `agency-lang/runtime` (line 35), and `pkg::` imports run the checked-in `index.js` rather than recompiling (`docs/dev/language/pkg-imports.md:43`). `git diff 3c92c0514..HEAD --stat -- packages/web-fetch packages/mcp` is empty. Under the new `failure(error, data, opts)` signature, `coerceData` accepts the options object as `data` without complaint, so a thrown error inside `fetchPage` or `mcp` now yields a failure with `checkpoint: null`, `functionName: null` (propagation messages say `(unknown)`), `destructiveRan: false` regardless of `__self.__destructiveRan`, and `data` holding the serialized checkpoint. The `mcp-catch-failure` agency-js test writes `result.data` to `__result.json` and would see the leaked options object. Fix: `pnpm run build` in both packages and commit the regenerated `index.js`. `packages/email` and `packages/github` are gitignored so they are fine.

**2. `tests/agency/agency-test-fn.agency:617` still reads `err.reason` / `err.limit` off the binder, which is now a string.**
`test()` in `stdlib/agency.agency:436-450` was updated to put `reason`/`limit` in `data`, but this node was not. `agency-test-fn.test.json:22` expects `maxCostTripsWholeCall` to return `"ok"`; with `err` a string, `err.reason` is undefined and it returns `"wrong failure shape"`. The PR description lists `tests/agency/{result,guards,subprocess}` as run locally; this file is outside that set. Change to `r.data.reason` / `r.data.limit`.

**3. `stdlib/agency.agency:1106` (`cliRun`, and the same pattern in `cliTest` 1092, `cliTypecheck` 1099, `cliAst` 1117, `cliFmt` 1128): the `failure(err) => failure(err)` rewrap now drops `data`.**
Before this PR `err` bound the whole error object so the rewrap preserved it; now it binds only the message, so `runFile`'s `limit_exceeded` failure (`makeLimitFailure`, ipc.ts:132) comes out of `cli(["run", file])` with `data: {}` and `data.reason` is gone. No in-repo caller reads it off a `cli()` result, so severity is low, but returning the original Result (`failure(err) => result`) also keeps `checkpoint`, `rejected`, and `functionName`, which the rewrap already lost.

**4. `tests/agency/fork/llm-tools/tool-non-retryable-failure.agency:6` no longer tests what its comment says.**
It writes `failure("permanent failure", { destructiveRan: true })`. Under the old codegen the user's object landed in the opts slot, so `destructiveRan` was honored and the tool was removed. Under the padded codegen it is `data`, `destructiveRan` defaults to false, and `failureTier` returns `neutral` ("You may call this tool again"). The test still passes only because its second `llmMocks` entry returns `tool-removed` unconditionally. The marker-based path is still covered by `tests/agency-js/tool-tiers`, so rewrite this fixture to use `destructive def alwaysFail()` or delete it and fix the description in `tool-non-retryable-failure.test.json`.

Two candidates I checked and did not report, in case they come up: the tightened zod Result schema in `typeToZodSchema.ts:303` is harmless for top-level values because `__validateType` and `__validateChain` short-circuit on `isFailure` before parsing (it only bites a hand-built failure nested inside a validated object); and `String(toolResult.value)` at `prompt.ts:1276` is unreachable with an object, since every rejection at a raise site is converted through `runtimeFailure` first.

packages/agency-lang/lib/typeChecker/resultUnion.ts
```diff
@@ -30,7 +30,12 @@ export function resultToObjectUnion(
         type: "objectType",
         properties: [
           { key: "success", value: bool("false") },
-          { key: "error", value: rt.failureType },
+          // The message. Always a string: `failure()` coerces anything else
+          // (lib/runtime/result.ts).
+          { key: "error", value: STRING_T },
```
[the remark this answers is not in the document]
egonSchiele: Fixed in 0058ce84e. All three fixtures now read the fields off `.data`: `agencyCoding.agency` and `agencyCodingRepair.agency` use `result.data.source` and `result.data.problems`, and `agency-test-fn.agency` uses `r.data.reason` and `r.data.limit`. I also scanned every `.agency` file in the repo for a `failure(name)` binder followed by a non-string field read on that name, and these three were the only ones.

packages/agency-lang/lib/typeChecker/resultUnion.ts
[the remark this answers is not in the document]
egonSchiele: Fixed in 0058ce84e. `resultUnion.ts` now passes the data type through `readerDataType`, which strips the null: `Result<T, D | null>` reads as `D`, so `result.data.line` typechecks without a guard. A data type that is only `null` reads as an empty object. The null stays on the producer side, where `synthFailureCall` and the assignability check use it to allow a bare `failure(msg)`. Two tests cover the read: one for a declared `D | null`, one for an unannotated def that returns failures both with and without data.

packages/agency-lang/lib/runtime/result.ts
egonSchiele: Yes, worth a warning. Done in 0058ce84e. `coerceData` now warns when it drops an array or a primitive: a `failureData` warn event to statelog carrying the dropped value, and a payload-free line on the console, the same split `failurePropagation` uses. Absent data (`failure(msg)`, or an explicit null) is the normal case and stays silent. A non-object here can only come from imported TypeScript, since AG2014 refuses it in Agency code.

packages/agency-lang/lib/runtime/prompt.ts
egonSchiele: Fixed in 0058ce84e. `normalizeForeignResult` is exported and runs at the two other boundaries: the plain-callable branch of `__call`, and `AgencyFunction.invoke`, whose `_fn` may be a TypeScript function. `try` keeps its call. The function returns anything that is not a failure, and any well-formed failure, unchanged by identity, so Agency defs pay one `isFailure` check per call. Two tests in `call.test.ts` cover a hand-built Result from a plain function and from an AgencyFunction wrapping one.

packages/agency-lang/lib/typeChecker/resultUnion.ts
egonSchiele: Fixed in 0058ce84e, as above: the reader-facing `.data` type strips the null in `readerDataType` (`resultUnion.ts`), and the producer-side check in `utils.ts` still sees the declared `D | null`. Your repro typechecks clean now, and the guide's `if (result.data.line != null)` idiom with it.

packages/agency-lang/lib/typeChecker/utils.ts
egonSchiele: Fixed in 0058ce84e. `emitAssignabilityError` now takes the checked expression, and `reportFailureDataError` reports AG2015 only when that expression is a `failure(...)` call. Your repro gets the generic "not assignable" message at the `helper()` call, and a test pins that it no longer says "needs a second argument".

packages/agency-lang/lib/typeChecker/validate.ts
egonSchiele: Fixed in 0058ce84e. The AG2016 walk moved into `validateResultDataTypes` in `validate.ts`, which `validateTypeReferences` calls and which `buildDefScope` now calls on every parameter type hint and on the return type. `reportFailureDataError` dropped its AG2016 branch: when the target's data type is not an object it returns true without reporting, since the annotation has already been named once. Tests cover a parameter annotation, a return annotation with no failure returned, and a def with three returns reporting exactly one error.

packages/agency-lang/lib/runtime/failurePropagation.ts
egonSchiele: Fixed in 0058ce84e. `truncate` is back on the message here, matching the other three sites in the file.

egonSchiele:
All four of the out-of-hunk findings and the five red CI jobs are addressed in 0058ce84e.

1. **Stale wrappers.** `packages/web-fetch/index.js` and `packages/mcp/index.js` are regenerated with `agency compile index.agency` run from inside each package, so the module id and config match the committed form. Both now call `runtimeFailure(__error, opts)`. The `mcp-catch-failure` agency-js test passes.
2. **`agency-test-fn.agency`.** `maxCostTripsWholeCall` reads `r.data.reason` and `r.data.limit`. `perCaseTimeoutThenNextCaseRuns` was red for the same reason one level down: `_failureFeedback` received only the message, so `limit_exceeded` was gone from the case feedback. It now takes the message and the data and appends the data as JSON when there is any. All 19 cases pass.
3. **`cli*` rewraps.** Each wrapper returns the underlying Result as it is when it is a failure, so data, checkpoint, and origin survive.
4. **`tool-non-retryable-failure`.** The tool is a `destructive def` now, which commits at entry and carries `destructiveRan` on its failure. The description in the `.test.json` says so.

CI's other failures were the two exact-output fixtures missing the new `data: {}` field (`runtime-arg-errors`, `interruptAssignmentReject`), `httpStatus.agency` reading `status` and `body` off the message binder instead of `.data`, and the two `agencyCoding*` fixtures doing the same with `source` and `problems`. All of those pass locally, along with the TypeScript typecheck, the stdlib fixture typecheck, and the unit tests for every changed file.

